PR #167, opened 2016-01-22
Fix 1569 winning hero with no troops shall retreat

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=1569
The following assumptions were made:
1) Winning hero with no army still gains XP, new levels, etc.
2) Raised dead with Necromancery skill are added before the check, so hero remains alive in such cases

PR #162, opened 2016-01-20
Fix 2377 "Seer Empty Text"

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=2377

PR #175, opened 2016-01-27
Feature/switch retreat no troops

vmarkovtsev:
This is related to https://github.com/vcmi/vcmi/pull/167
Since this new switch breaks the save format, I send it as a separate PR


lib/CModHandler.h
```diff
@@ -253,17 +253,20 @@ class DLL_LINKAGE CModHandler
...
 		int MAX_HEROES_ON_MAP_PER_PLAYER;
+		bool WINNING_HERO_WITH_NO_TROOPS_RETREATS;
 
 		template <typename Handler> void serialize(Handler &h, const int version)
 		{
 			h & data & CREEP_SIZE & WEEKLY_GROWTH & NEUTRAL_STACK_EXP & MAX_BUILDING_PER_TURN;
-			h & DWELLINGS_ACCUMULATE_CREATURES & ALL_CREATURES_GET_DOUBLE_MONTHS & MAX_HEROES_AVAILABLE_PER_PLAYER & MAX_HEROES_ON_MAP_PER_PLAYER;
+			h & DWELLINGS_ACCUMULATE_CREATURES & ALL_CREATURES_GET_DOUBLE_MONTHS &
+			MAX_HEROES_AVAILABLE_PER_PLAYER & MAX_HEROES_ON_MAP_PER_PLAYER &
+			WINNING_HERO_WITH_NO_TROOPS_RETREATS;
```
alexvins: No need to break saves. just make if false for lower version.

vmarkovtsev: Can you give an example? Where is the version, how to use it in serialization?..

alexvins: "const int version" parameter is current used version. 

in Connection.h there are definitions of supported versions:

```cpp
const ui32 version = 755;
const ui32 minSupportedVersion = 753;
```

change it to 

```cpp
const ui32 version = 756;
const ui32 minSupportedVersion = 753;
```

and in serialization

```cpp
if(version >= 756) //save format backward compatibility
{
	h & WINNING_HERO_WITH_NO_TROOPS_RETREATS;
}
else if(!h.saving)
{
	WINNING_HERO_WITH_NO_TROOPS_RETREATS = false;
}
```

vmarkovtsev: Thanks!

ArseniyShestakov:
You can as well do same thing with #162 if you want it merged before save format breakage.

lib/filesystem/AdapterLoaders.h
```diff
@@ -59,14 +59,14 @@ class DLL_LINKAGE CFilesystemList : public ISimpleResourceLoader
...
-        //class is not copyable 
-        return *this; 
+	CFilesystemList(CFilesystemList &)
+    {
+		//class is not copyable
+    }
+    CFilesystemList &operator=(CFilesystemList &)
+    {
+        //class is not copyable
+        return *this;
```
ArseniyShestakov: Some mix of tabs and spaced there.

vmarkovtsev:
Added new format version and worked on @ArseniyShestakov hint

PR #176, opened 2016-01-27
Fix 2160 dismissing a VIP hero

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=2160

### DIFF
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -23,6 +23,7 @@
 #include "../CCreatureHandler.h"
 #include "../BattleState.h"
 #include "../CTownHandler.h"
+#include "../mapping/CMap.h"
 #include "CGTownInstance.h"
 
 ///helpers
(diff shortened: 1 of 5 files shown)


lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1454,3 +1455,27 @@ bool CGHeroInstance::hasVisions(const CGObjectInstance * target, const int subty
 
 	return (distance < visionsRange) && (target->pos.z == pos.z);
 }
+
+bool CGHeroInstance::isMissionCritical(const CMap* map) const
```
alexvins: Map objects should not use map pointers from outside. For now they should use IObjectInterface::cb for game state access.

alexvins: As for triggeredEvents it may be IObjectInterface::cb->getMapHeader()->triggeredEvents

vmarkovtsev:
@alexvins did what you said, API got better anyway +1

PR #173, opened 2016-01-26
Fix 981 reset hero on hiring after retreat/surrender

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=981

PR #178, opened 2016-01-28
Fix 2406 reset potentially available heroes

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=2406
As noted at https://github.com/vcmi/vcmi/pull/173#issuecomment-175847431
I am resetting every eligible hero in the pool - on next week, retreated coward can be chosen again randomly.

### DIFF
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -1080,6 +1080,20 @@ si32 CGHeroInstance::manaRegain() const
 	return 1 + valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, 8) + valOfBonuses(Bonus::MANA_REGENERATION); //1 + Mysticism level
 }
 
+si32 CGHeroInstance::getManaNewTurn() const
+{
+	if(visitedTown && visitedTown->hasBuilt(BuildingID::MAGES_GUILD_1))
+	{
+		//if hero starts turn in town with mage guild - restore all mana
+		return std::max(mana, manaLimit());
+	}
+	si32 res = mana + manaRegain();
+	res = std::min(res, manaLimit());
+	res = std::max(res, mana);
+	res = std::max(res, 0);
+	return res;
+}
+
 // /**
 //  * Places an artifact in hero's backpack. If it's a big artifact equips it
 //  * or discards it if it cannot be equipped.
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -1337,7 +1337,7 @@ void CGameHandler::newTurn()
 				NewTurn::Hero hth;
 				hth.id = hero->id;
 				hth.move = maxmove;
-				hth.mana = std::max((si32)(0), std::max(hero->mana, std::min((si32)(hero->mana + hero->manaRegain()), hero->manaLimit())));
+				hth.mana = hero->getManaNewTurn();
 				n.heroes.insert(hth);
 			}
 		}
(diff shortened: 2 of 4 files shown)


server/CGameHandler.cpp
```diff
@@ -1324,6 +1324,25 @@ void CGameHandler::newTurn()
...
+		{
+			// reset retreated or surrendered heroes
+			auto maxmove = hero->maxMovePoints(true);
+			// if movement is greater than maxmove, we should decrease it
+			if(hero->movement != maxmove || hero->mana < hero->manaLimit())
+			{
+				NewTurn::Hero hth;
+				hth.id = hero->id;
+				hth.move = maxmove;
+				hth.mana = std::max((si32)(0), std::max(hero->mana, std::min((si32)(hero->mana + hero->manaRegain()), hero->manaLimit())));
```
alexvins: This line is unreadable: too much "()". Also it is better to use std::max<si32>(), std::min<si32>().

vmarkovtsev: This line is the same as 1396, which I copy-pasted. Let's refactor these std:max(std::min()) in a separate PR - I don't want to regress here beforehand.

vmarkovtsev:
OK, I've had time to refactor this.

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1080,6 +1080,20 @@ si32 CGHeroInstance::manaRegain() const
 	return 1 + valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, 8) + valOfBonuses(Bonus::MANA_REGENERATION); //1 + Mysticism level
 }
 
+si32 CGHeroInstance::getManaTomorrow() const
```
ArseniyShestakov: Here is small note why I against having any function that refer certain dates like "tomorrow".

In our bonus system any of bonuses may expire on any turn. That not require immediate fixing because vanilla Heroes 3 mechanics unlikely have anything that alter mana regen for certain time, but you must keep it in mind as anything possible in mods.

Currently I have helper called TurnInfo that currently live in pathfinder as I not yet refactored it, but it's already used in newTurn() to give hero appropriate movement points. So you may want to reuse it there for now.

ArseniyShestakov:
So add to what Warmonger posted. It's better to make max mana function similar to what already implemented for movement points. Later on I'll try to find some way to merge TurnInfo functionality into bonus system so you don't have to work with complex selectors just to check few bonuses on certain turn.

vmarkovtsev:
So I am renaming getManaTomorrow to getManaNewTurn, right?

As for the logic, it was:
```
std::max((si32)(0), std::max(hero->mana, std::min((si32)(hero->mana + hero->manaRegain()), hero->manaLimit())))
```
or with indent
```
std::max(
    (si32)(0),
    std::max(
        hero->mana,
        std::min(
            (si32)(hero->mana + hero->manaRegain()),
            hero->manaLimit()
        )
    )
)
```
if the old code is wrong, I will fix it. Otherwise, I just made it more clear for understanding.

ArseniyShestakov:
> So I am renaming getManaTomorrow to getManaNewTurn, right?
Yep that would be better. Though what would be good is to make sure that e.g manaRegain() check bonuses for appropriate turn.

vmarkovtsev:
Renamed. Bonuses seem out of scope of this PR - but I of course agree that it is necessary.

ArseniyShestakov:
Ok then looks like it working well at moment.

ArseniyShestakov:
So one more interested note. I didn't know that retreated hero don't get full mana points on next turn in tavern, but I tested H3 and it's currently exactly same behaviour. I think code is good now, merging.

PR #179, opened 2016-01-29
Fix 2405 quest timeout mechanics

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=2405

Shifted by one day to conform with OH3


vmarkovtsev:
Line 535 is the only really changed

alexvins:
@vmarkovtsev please do formatting in separate commit.

PR #180, opened 2016-01-29
Refactored CFileInfo & ResourceID

Zyx-2000:
When I worked on windows unicode filename support, I noticed several problems with the class CFileInfo:

- It does not support unicode file paths on windows
- The meaning of its methods does not match those of boost::filesystem::path
- It does not make sense to use the same class for VCMI's virtual file system and the OS filesystem
- It requires unnecessary string copying (and therefore potentially expensive memory allocations)

Therefore I decided to replace CFileInfo, with a set of functions in a new namespace called FileInfo. These functions use boost::string_ref to avoid string copies and memory allocations. They also use the same name meanings as boost's path methods.

I also changed ResourceID; before, creating an instance of it, assuming move semantics were used whenever possible, required 2-3 string copies/memory allocations, while it now requires only one. I also disabled unused set methods, and the default constructor. They can be re-enabled again if needed, but I think it is safer to not have ResourceID objects in uninitialized/invalid states.


client/gui/CAnimation.cpp
```diff
@@ -75,29 +75,26 @@ class CFileCache
 	static const int cacheSize = 50; //Max number of cached files
 	struct FileData
 	{
-		ResourceID name;
-		size_t size;
-		ui8 * data;
+		ResourceID				name;
```
alexvins: Do not use TAB for aligning, align with spaces. Tabs are for indentation.

ArseniyShestakov:
This comment isn't directly related to pull request, but as you work on filesystem code I have two requests / suggestions:

* VCMI currently don't update filesystem when it's changed. E.g if you added new save/map into appropriate directory and client was already running those won't become visible. So currently you must restart it each time if you added new map.

* We currently use SDL_mixer for music playback which does support streaming, but due to our resource manager limitations we only have option to load full track into memory. This make it much harder to implement some music playback behaviour that H3 have ([issue 1634](http://bugs.vcmi.eu/view.php?id=1634)). This one is super low priority, but that would be handy if we can somehow keep using resource manager for music, but let SDL_mixer stream it.

Thanks!

Zyx-2000:
Actually, I don't think any changes to filesystem code are needed to implement music streaming. All that is needed is to write an [SDL_RWops](https://wiki.libsdl.org/SDL_RWops) wrapper around CInputStream.

ArseniyShestakov:
Would be great if you could implement it. Then I can work on actual music playback changes.

ArseniyShestakov:
So I tested game with this branch more include with mods and everything looks fine. Merging.

PR #181, opened 2016-01-29
Feature/drawbridge mechanics

ArseniyShestakov:
Very basic code is ready, but proper implementation will include some rework of both movements and obstacle code. Currently I'm not planning to implement any client-side visualization, but if you have any ideas how to make future client-side programming easier please let me know.

More details about H3 drawbridge behaviour and implementation:

* All walking stacks request lowering of drawbridge from hexes before 95 and 96.
* All walking stacks request lowering from hex before 94, but only in case their destination at least hex 95. 
Exception: Fortress have moat on hex 94 so bridge will be lowered even if 94 is destination hex.
* Bridge is automatically raised when no creature standing on tiles 95 and 96. 
Exception: Fortress bridge is not raised if there creature on hex 94 as well.
* If bridge raised any stack (alive or dead) on hex 94 wont let lower it. 
Exception: destroyed gate tower make bridge constantly lowered.


lib/CBattleCallback.cpp
```diff
@@ -1129,9 +1138,20 @@ AccessibilityInfo CBattleInfoCallback::getAccesibility() const
...
+		{
+		case EDrawbridgeState::RAISED:
+			accessability = EAccessibility::GATE;
+			break;
+
+		case EDrawbridgeState::RAISED_BLOCKED:
+			accessability = EAccessibility::UNAVAILABLE;
+			break;
+		}
+		ret[95] = ret[96] = accessability;
```
alexvins: Use ESiegeHexes here

lib/GameConstants.h
```diff
@@ -510,6 +510,15 @@ namespace EWallState
 	};
 }
 
+enum class EDrawbridgeState
```
alexvins: Just in case define underlying type here.

lib/NetPacksLib.cpp
```diff
@@ -1208,6 +1208,11 @@ DLL_LINKAGE void BattleObstaclePlaced::applyGs( CGameState *gs )
 	gs->curB->obstacles.push_back(obstacle);
 }
 
+DLL_LINKAGE void BattleDrawbridgeStateChanged::applyGs(CGameState *gs)
+{
+	gs->curB->si.drawbridgeState = state;
```
alexvins: Should be check for curB. State change on the end of battle in theory may lead to crash.

ArseniyShestakov:
So basic implementation is feature complete, but with two "visual" glitches related to double-wide stacks:

* When moving from outside to outer bridge hex 94 stack will stop one two one tile before bridge hex even if moving from top or down hexes. This is intended behavior when stack moving from hex 93 to 95, but not when you moving from 59 to 95.
* Bridge raising isn't handled properly for double wide stacks. So bridge will only be raised after stack movement is finished while for single-hex stacks it's raised once it's leave gate / bridge tile.

Though both problems don't affect gameplay at all while first glitch appear due to some dirty hack that fix actual game mechanics issue.

ArseniyShestakov:
My internet is lagging like hell, so I get bored and decide to implement client visual for drawbridge.
So it's working with sound effect, but without animation. Code is bad, sorry...

ArseniyShestakov:
So this one is imperfect and I may still want some changes like moving more hex IDs into enum all around client code. Though to actually improve drawbridge in server code I'll need to rework obstacle handling and probably movement too and I don't want to put these changes into this branch.

client/battle/CBattleInterface.h
```diff
@@ -198,16 +198,32 @@ class CBattleInterface : public CIntObject
...
+			BACKGROUND = 0,
+			BACKGROUND_WALL = 1,
+			KEEP,
+			BOTTOM_TOWER,
+			BOTTOM_WALL,
+			WALL_BELLOW_GATE,
+			WALL_OVER_GATE,
+			UPPER_WALL,
+			UPPER_TOWER,
+			GATE = 9,
```
alexvins: Is there a reason to explicitly define this?

ArseniyShestakov: Yes. Enum like that is clearly much better than magic numbers with comments everywhere in code.

alexvins: You misunderstood me. Why  "GATE = 9" and everything else is auto?

ArseniyShestakov: Absolutely no reason, just forgot to remove it.

PR #182, opened 2016-01-30
Fix 1916 water wheel reward

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=1916

### DIFF
--- a/lib/mapObjects/CRewardableObject.cpp
+++ b/lib/mapObjects/CRewardableObject.cpp
@@ -422,7 +422,7 @@ void CRewardableObject::setPropertyDer(ui8 what, ui32 val)
 
 void CRewardableObject::newTurn() const
 {
-	if (resetDuration != 0 && cb->getDate(Date::DAY) > 1 && cb->getDate(Date::DAY) % (resetDuration) == 1)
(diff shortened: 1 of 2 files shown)


lib/mapObjects/CRewardableObject.cpp
```diff
@@ -1003,16 +1003,30 @@ void CGVisitableOPW::initObj()
...
 		break;
 	}
 }
 
+void CGVisitableOPW::setPropertyDer(ui8 what, ui32 val)
+{
+	if(ID == Obj::WATER_WHEEL && what == ObjProperty::REWARD_RESET)
+	{
+		auto& reward = info[0].reward.resources[Res::GOLD];
+		if(info[0].numOfGrants == 0)
```
DjWarmonger: Interesting solution. I think this check is not needed since this funtion won't be called on day 1 anyway.

vmarkovtsev: This check is actually needed: if there was no visits on the previous week (every Monday numOfGrants is set to 0, on each visit it ++), we increase the reward to 1000 on this week.

DjWarmonger: Ok got it :)

PR #183, opened 2016-01-30
Fix 1989 cavaliers -> champions in stables

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=1989

### DIFF
--- a/lib/mapObjects/CRewardableObject.cpp
+++ b/lib/mapObjects/CRewardableObject.cpp
@@ -23,13 +23,13 @@
 
 bool CRewardLimiter::heroAllowed(const CGHeroInstance * hero) const
 {
-	if (dayOfWeek != 0)
+	if(dayOfWeek != 0)
 	{
 		if (IObjectInterface::cb->getDate(Date::DAY_OF_WEEK) != dayOfWeek)
 			return false;
 	}
 
-	for (auto & reqStack : creatures)
+	for(auto & reqStack : creatures)
 	{
 		size_t count = 0;
 		for (auto slot : hero->Slots())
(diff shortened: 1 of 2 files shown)


ArseniyShestakov:
Small comment about H3 behaviour: InfoWindow must appear as well even if it 2nd+ visit of stables and you just get an upgrade without any movement bonus.

vmarkovtsev:
Damn, it will be much harder to display 2 InfoWindows - it is hardcoded to be a single window per CGBonusingObject, so there are going to be quite a few ugly hacks.
Besides, I need to find out the corresponding text ID. Can you post a screenshot of that window from OH3 please?

ArseniyShestakov:
It's uses same text ID in H3 so there is no special message that only include information about upgrade.

As you note there was object refactoring, but unfortunately Ivan didn't find way to make objects fully configurable back then. So if adding infowindow on upgrade will require dirty hacks in code then just leave it as and post bug about unfinished UI feature on bug tracker.

vmarkovtsev:
Am I right that if we visit stables for the first time and have cavaliers, the message says "... blah-blah you got champions", and if we visit stables for the second time and have cavaliers, exactly the same message must be shown (not the usual "you have already visited this")? Or should we concatenate "you have already visited this" with "you got champions" text and picture? The third variant is first display "you have already visited this" and then display  "... blah-blah you get champions" (the bad one - hard to implement).

ArseniyShestakov:
Just show exactly same message as it's what H3 do.

vmarkovtsev:
Got it, it's easy then

vmarkovtsev:
Done

PR #184, opened 2016-01-30
support for music streaming

Zyx-2000:
I have implemented music streaming, as discussed [here](https://github.com/vcmi/vcmi/pull/180). Implemented through a wrapper for SDL around CInputStream. It can be used for inputs to SDL other than music, if needed in the future.

I also fixed a memory leak (and a bug) in MusicEntry.

### DIFF
--- a/client/SDLRWwrapper.cpp
+++ b/client/SDLRWwrapper.cpp
@@ -1,3 +1,13 @@
+/*
+ * SDLRWwrapper.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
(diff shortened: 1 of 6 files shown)


ArseniyShestakov:
Awesome! Going to test it soon.

client/SDLRWwrapper.cpp
```diff
@@ -0,0 +1,85 @@
+#include "StdInc.h"
```
alexvins: New files missing copyright banner.

ArseniyShestakov:
Thanks again. I hoped to have time implement features based on streaming earlier, but didn't have time last few days. For now I'll merge this one since it's worked well for me.

PR #185, opened 2016-01-31
Fix 2264 learning stone & Learning skill

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=2264

### DIFF
--- a/lib/mapObjects/CRewardableObject.h
+++ b/lib/mapObjects/CRewardableObject.h
@@ -104,9 +104,10 @@ public:
 	/// if set to true, object will be removed after granting reward
 	bool removeObject;
 
-	/// Generates list of components that describes reward
-	virtual void loadComponents(std::vector<Component> & comps) const;
-	Component getDisplayedComponent() const;
+	/// Generates list of components that describes reward for a specific hero
+	virtual void loadComponents(std::vector<Component> & comps,
+	                            const CGHeroInstance * h) const;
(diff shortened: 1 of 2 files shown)


lib/mapObjects/CRewardableObject.h
```diff
@@ -104,9 +104,10 @@ class DLL_LINKAGE CRewardInfo
 	/// if set to true, object will be removed after granting reward
 	bool removeObject;
 
-	/// Generates list of components that describes reward
-	virtual void loadComponents(std::vector<Component> & comps) const;
-	Component getDisplayedComponent() const;
+	/// Generates list of components that describes reward for a specific hero
+	virtual void loadComponents(std::vector<Component> & comps,
+	                            const CGHeroInstance * h) const;
```
DjWarmonger: You really don't need o break lines like that in the age of HD monitors ;)

vmarkovtsev: Have you ever done a 3-way merge with 120-char lines?

vmarkovtsev: To be more precise, here is the motivation: https://google.github.io/styleguide/cppguide.html#Line_Length

DjWarmonger: For me it's quite the opposite with one large code window in Visual Studio. Once I had to work at split-screen and it was a nightmare.

PR #186, opened 2016-01-31
Rename radious -> radius

vmarkovtsev:
There is no such English word, "radious"
Some files get new lines in the end - BSD sed bug + best practices

Sorry, it just pissed me off :)


ArseniyShestakov:
You need to update "artifacts.json" as well. Compatibility for mods also probably required, but no idea if there any mods currently that use "SIGHT_RADIOUS" and I not sure what happen for nonexist bonus in configuration file.

DjWarmonger:
Yes, many do. Cover heroes and H4 artifacts use this bonus, probably other mods as well.

It is not possible to rename bonus itself, we need to keep backward compatibility with existing mods. Sorry.

vmarkovtsev:
I see. So leaving RADIOUS but renaming everything else will do, right?

ArseniyShestakov:
@DjWarmonger is there actually no way to keep mods compatibility while renaming bonus?

vmarkovtsev:
OK, so I left SIGHT_RADIOUS but renamed everything else.

Gaerzi:
It's always possible to keep radious as an alias for backward compatibility but use radius otherwise.

ArseniyShestakov:
@Gaerzi this is tricky, but it's would be perfect if we could just add alias on mods loading. Sadly it's may require hacks in mods loading code. Probably it's just better to have old version until we eventually decide to break mod compatability.

ArseniyShestakov:
As @vmarkovtsev requested merging it as is.

PR #187, opened 2016-01-31
Fix 2362 cast spells in town garrison

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=2362

### DIFF
--- a/client/windows/CSpellWindow.cpp
+++ b/client/windows/CSpellWindow.cpp
@@ -6,6 +6,7 @@
 #include "CAdvmapInterface.h"
 #include "GUIClasses.h"
 #include "InfoWindows.h"
+#include "CCastleInterface.h"
 
 #include "../CBitmapHandler.h"
 #include "../CDefHandler.h"
@@ -109,7 +110,7 @@ CSpellWindow::CSpellWindow(const SDL_Rect &, const CGHeroInstance * _myHero, CPl
 		Uint8 *sitesPerOurTab = s.combatSpell ? sitesPerTabBattle : sitesPerTabAdv;
 
 		++sitesPerOurTab[4];
-		
+
 		s.forEachSchool([&sitesPerOurTab](const SpellSchoolInfo & school, bool & stop)
 		{
 			++sitesPerOurTab[(ui8)school.id];
@@ -156,9 +157,9 @@ CSpellWindow::CSpellWindow(const SDL_Rect &, const CGHeroInstance * _myHero, CPl
 
 	leftCorner = BitmapHandler::loadBitmap("SpelTrnL.bmp", true);
 	rightCorner = BitmapHandler::loadBitmap("SpelTrnR.bmp", true);
-	
+
 	spells = new CAnimation("Spells.def");
-		
+
 	spellTab = CDefHandler::giveDef("SpelTab.def");
 	schools = CDefHandler::giveDef("Schools.def");
 	schoolBorders[0] = CDefHandler::giveDef("SplevA.def");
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
... and please merge https://github.com/vcmi/vcmi/pull/186 - it triggers full 10 mins recompilation of everything every time I switch branches :)

lib/mapping/CDrawRoadsOperation.cpp
```diff
@@ -140,45 +140,47 @@ static bool ruleIsSomething(const std::string & rule)
 	return rule == "+";
 }
 
+/*
 static bool ruleIsAny(const std::string & rule)
 {
 	return rule == "?";
 }
+*/
```
ArseniyShestakov: Can you clarify why exactly you commented out that function?

ArseniyShestakov:
Disabling drawing of spell icons while garrisoned is really bad solution. Instead just patch "CSpellWindow::SpellArea::clickLeft" and add additional check here for adventure spells in case of hero is in garrison.

lib/mapping/CDrawRoadsOperation.cpp
vmarkovtsev: It is unused - generates a warning.

vmarkovtsev:
The problem is a bit deeper: it does not matter if hero is visiting town or stands in garrison, if town interface is active, any adventure spell is displayed wrong. We should check if town iface is active then.

vmarkovtsev:
So here is what I've done. In client, I did what @ArseniyShestakov wrote if castle interface is active. There is no ability to find out if castle iface is active on server side, so I only disabled casting spells if hero is in town garrison.

lib/mapping/CDrawRoadsOperation.cpp
ArseniyShestakov: It is used in assert unless you build without them as release. Also I think formatting fixes in related code are reasonable, but please at least don't include unrelated files into commits. That RMG-related file clearly has nothing to do with adventure map spells.

vmarkovtsev:
I wrapped ruleIsAny with NDEBUG. Didn't know it is used in assertions, sorry

ArseniyShestakov:
Sorry, but I think change in this file just must be dropped as completely unrelated to fix.

vmarkovtsev:
Of course, maaaster :)

ArseniyShestakov:
BTW note for everyone else. When first looked problem I think to why not just check if hero garrisoned on client too, but there is actually multiple bugs and crashes possible if you cast with CCastleInterface active so it's reasonable limitation.

ArseniyShestakov:
One more thing.

owner->myInt->castleInt wouldn't work there?

client/windows/CSpellWindow.cpp
```diff
@@ -622,10 +623,11 @@ void CSpellWindow::SpellArea::clickLeft(tribool down, bool previousState)
 			owner->myInt->showInfoDialog(std::string(msgBuf));
 			return;
 		}
-
+		auto intiter = GH.listInt.end();
+		--intiter; --intiter;
 		//battle spell on adv map or adventure map spell during combat => display infowindow, not cast
 		if((sp->isCombatSpell() && !owner->myInt->battleInt)
-			|| (sp->isAdventureSpell() && owner->myInt->battleInt))
+		   || (sp->isAdventureSpell() && (owner->myInt->battleInt || dynamic_cast<CCastleInterface*>(*intiter))))
```
ArseniyShestakov: owner->myInt->castleInt check for nullptr wouldn't work there?

vmarkovtsev:
Changed to owner->myInt->castleInt - it works.

BTW, I know how to fix those Travis failures and why they appear.
Ubuntu's repo is set to some mirror which is updated regularly but sometimes metadata becomes to be in a inconsistent state (apt-get upgrade runs in the middle of mirror's sync). This is fixed by setting the official Ubuntu package server (packages.ubuntu.com) in /etc/apt/sources.list
I've successfully cured it this way on my production servers.

PR #136, opened 2015-11-07
issue/1372

alexvins:
Second attempt to fix http://bugs.vcmi.eu/view.php?id=1372
(#124 has serious regressions, merge has been reverted.)

PR #188, opened 2016-02-01
Fix 1376 turrets damage can decrease

vmarkovtsev:
Try to fix http://bugs.vcmi.eu/view.php?id=1376

I've chosen way #3 (hardcode) as the most simple one.

### DIFF
--- a/config/creatures/special.json
+++ b/config/creatures/special.json
@@ -112,7 +112,11 @@
 		"index": 149,
 		"level": 0,
 		"faction": "neutral",
-		"abilities": { "shooter" : { "type" : "SHOOTER" } },
+		"abilities":
+		{
+			"shooter" : { "type" : "SHOOTER" },
+			"ignoreDefence" : { "type" : "ENEMY_DEFENCE_REDUCTION", "val" : 100 }
+		},
 		"graphics" :
 		{
 			"animation": "CLCBOW.DEF" // needed to pass validation, never used


lib/CBattleCallback.cpp
```diff
@@ -879,37 +885,50 @@ TDmgRange CBattleInfoCallback::calculateDmgRange(const BattleAttackInfo &info) c
...
+	{
+		double multAttackReduction =
+			(100 - battleBonusValue(info.attackerBonuses,
+			                        Selector::type(Bonus::GENERAL_ATTACK_REDUCTION)))
+			/ 100.0;
+		attackDefenceDifference +=
+			battleBonusValue(info.attackerBonuses,
+			                 Selector::typeSubtype(Bonus::PRIMARY_SKILL,
+			                                       PrimarySkill::ATTACK))
+			* multAttackReduction;
```
DjWarmonger: Here we can see that 80-char limit leads to multi-line zig-zags :/

vmarkovtsev: Matter of habit. All PEP8-conformant (the majority) Python code looks the same. Actually, my editor font is 12-14 (I see bad) and on 13" macbook it was the only way to view everything (no splitscreens).

DjWarmonger: At least aligne them to left ot right so we could see it's one piece of code. Now every line has different alignment :v

ArseniyShestakov:
Hardcoding is reasonable, but not in case when we already have appropriate way to do things. AVS suggested to solve this by adding ENEMY_DEFENCE_REDUCTION bonus to arrow towers (#136). I there any reason this can't be done that way?

lib/CBattleCallback.cpp
vmarkovtsev: It is standard to format code this way (align parameters by function boundary), but I don't insist - if you suggest another formatting style to fit into 80 - OK. Could you post an example?

DjWarmonger:
ENEMY_DEFENCE_REDUCTION set to 100 means that effective creature defense will be always zero, but turret attack can go high and result in damage bonus. Is that desired?

ArseniyShestakov:
As far as I know that exactly what we want in case of arrow towers.

vmarkovtsev:
I'll see what I can do with this bonus

vmarkovtsev:
Wow, that was easy :)

config/creatures/special.json
```diff
@@ -112,7 +112,11 @@
 		"index": 149,
 		"level": 0,
 		"faction": "neutral",
-		"abilities": { "shooter" : { "type" : "SHOOTER" } },
+		"abilities":
+		{
+			"shooter" : { "type" : "SHOOTER" },
+			"reduceDefence" : { "type" : "ENEMY_DEFENCE_REDUCTION", "val" : 100 }
```
alexvins: "ingoreDefence" is better name for this ability

DjWarmonger:
We've desined awesome mod system for a purpose ;)

PR #190, opened 2016-02-02
Fix unused function warning in release mode

vmarkovtsev:
As discussed at https://github.com/vcmi/vcmi/pull/187#discussion-diff-51508439

### DIFF
--- a/lib/mapping/CDrawRoadsOperation.cpp
+++ b/lib/mapping/CDrawRoadsOperation.cpp
@@ -12,7 +12,7 @@
 #include "CDrawRoadsOperation.h"
 #include "CMap.h"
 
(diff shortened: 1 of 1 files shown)


alexvins:
NDEBUG should be explicitly defined I think.

vmarkovtsev:
No. NDEBUG is defined by msvc automatically (by gcc, luckily not)  and not defined otherwise according to *the standard*.  And again by the standard assert is a macro which does not generate function call with NDEBUG defined. We are already safe, and if someone needs this function in non-asserting code, he must just remove ifndef away. Pease note that my approach is standard and is used in Clang, Google and Samsung projects.

alexvins:
OK.

PR #191, opened 2016-02-02
Implement psychic elementals vs mind immune units

vmarkovtsev:
As mentioned in TODO.


ArseniyShestakov:
It's already merged, but I think better way to implement it would be to add bonus like MIND_DAMAGE or something to creature instead of hardcoding it with creature id.

DjWarmonger:
Better yet, "SPECIAL_DAMAGE" with subID equal to bonus type that gives protection.

ArseniyShestakov:
Oh something like SPECIAL_DAMAGE is also great idea!

PR #192, opened 2016-02-02
Implement half damage to petrified units

vmarkovtsev:
Another TODO

### DIFF
--- a/config/spells/ability.json
+++ b/config/spells/ability.json
@@ -1,10 +1,10 @@
 {
 	"stoneGaze" : {
 		"index" : 70,
-		"targetType": "NO_TARGET",		
+		"targetType": "NO_TARGET",
 		"animation":{
 			//need special animation
-		},		
+		},
 		"sounds": {
 			"cast": "PARALYZE"
 		},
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
BTW, how about giving greater basilisks 25% stone gaze chance? They are greater, you know :)

vmarkovtsev:
... and does "blinded unit retaliates" mean that currently blinded units strike back?

ArseniyShestakov:
Yes, it's 50% damage on basic and 25% on advanced level. On expert blinded stack don't strikes back [http://heroes.thelazy.net/wiki/Blind]([1]).

ArseniyShestakov:
About altering mechanics: you can always make it moddable later on :-)

alexvins:
It should not be hardcoded. It should be additional spell effect.
Smth like:
```javascript
"defenceBoost":{			"type" : "PRIMARY_SKILL",
						"subtype" : "primSkill.defence",
						"val" : 50,
						"valueType" : "PERCENT_TO_ALL",
						"duration" : "N_TURNS"}
```

lib/CBattleCallback.cpp
```diff
@@ -1020,8 +1020,12 @@ TDmgRange CBattleInfoCallback::calculateDmgRange(const BattleAttackInfo &info) c
 		multBonus *= 0.5;
 	}
 
-	// TODO attack on petrified unit 50%
-	// blinded unit retaliates
+	// attack on petrified unit 50%
+	if(info.defenderBonuses->getBonus(Selector::source(Bonus::SPELL_EFFECT, SpellID::STONE_GAZE)))
```
alexvins: This is wrong on it own: stoneGaze has 2 effects with different duration.

vmarkovtsev:
Did as @alexvins wrote

PR #193, opened 2016-02-02
Implement blinded units attack penalties

vmarkovtsev:
Last TODO

### DIFF
--- a/config/spells/ability.json
+++ b/config/spells/ability.json
@@ -1,10 +1,10 @@
 {
 	"stoneGaze" : {
 		"index" : 70,
-		"targetType": "NO_TARGET",		
+		"targetType": "NO_TARGET",
 		"animation":{
 			//need special animation
-		},		
+		},
 		"sounds": {
 			"cast": "PARALYZE"
 		},
@@ -26,8 +26,18 @@
 						"val" : 0,
 						"type" : "NO_RETALIATION",
 						"duration" : "UNITL_BEING_ATTACKED"
+					},
+					"defenceBoost" : {
+						"type" : "PRIMARY_SKILL",
+						"subtype" : "primSkill.defence",
+						"val" : 50,
+						"valueType" : "PERCENT_TO_ALL",
+						"duration" : [
+						    "UNITL_BEING_ATTACKED",
+						    "N_TURNS"
+						]}
 					}
-				}				
+				}
 			}
 		},
 		"flags" : {
@@ -37,10 +47,10 @@
 	"poison" : {
 		"index" : 71,
 		"targetType": "NO_TARGET",
-		
+
 		"animation":{
 			"affect":["SP11_"]
-		},		
+		},
 		"sounds": {
 			"cast": "POISON"
 		},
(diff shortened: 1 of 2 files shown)


lib/CBattleCallback.cpp
```diff
@@ -1020,8 +1020,28 @@ TDmgRange CBattleInfoCallback::calculateDmgRange(const BattleAttackInfo &info) c
 		multBonus *= 0.5;
 	}
 
-	// TODO attack on petrified unit 50%
 	// blinded unit retaliates
+	auto blinded = info.attackerBonuses->getBonus(Selector::source(Bonus::SPELL_EFFECT, SpellID::BLIND));
```
alexvins: blind spell has 3 effects = 3 bonuses (2 of them with 0 val). Which bonus will you get?

lib/CBattleCallback.cpp
```diff
@@ -1020,8 +1020,28 @@ TDmgRange CBattleInfoCallback::calculateDmgRange(const BattleAttackInfo &info) c
...
 	}
 
-	// TODO attack on petrified unit 50%
 	// blinded unit retaliates
+	auto blinded = info.attackerBonuses->getBonus(Selector::source(Bonus::SPELL_EFFECT, SpellID::BLIND));
+	if(blinded)
+	{
+		switch(blinded->val)
+		{
+			case SecSkillLevel::BASIC:
```
alexvins: Spell effect bonus "val" can`be SecSkillLevel. It is explicitly defined number (defaults to "power" spell attribute).

vmarkovtsev: I looked at
```
auto isAdvancedAirShield = [](const Bonus *bonus)
	{
		return bonus->source == Bonus::SPELL_EFFECT
			&& bonus->sid == SpellID::AIR_SHIELD
			&& bonus->val >= SecSkillLevel::ADVANCED;
	};
```
below. Is that code wrong too?

lib/CBattleCallback.cpp
vmarkovtsev: So the TODO is outdated and everything works as-is, right?

lib/CBattleCallback.cpp
DjWarmonger: Yes it's wrong altogether.

lib/CBattleCallback.cpp
alexvins: Need testing. Have not tested this exact case.

lib/CBattleCallback.cpp
DjWarmonger: Air shield gives GENERAL_DAMAGE_REDUCTION bonus.
This piece of code effectively does nothing and is outdated.

On the other hand, the code indicates that distance penalty and air shield penalty should not be cumulative. Strategija reads:
"Attacker is a shooter with range, wall or melee penalty, or Advanced Air Shield is cast"

Though I'm not sure it's correct or logical.

vmarkovtsev:
Closing as this is already implemented

PR #195, opened 2016-02-10
Configurable ambient, visit and removal sounds

ArseniyShestakov:
More details about H3 ambient sound implementation available on mantis:
https://bugs.vcmi.eu/view.php?id=2368
Other issues fixed:
https://bugs.vcmi.eu/view.php?id=814
https://bugs.vcmi.eu/view.php?id=1630


client/CPlayerInterface.cpp
```diff
@@ -2813,3 +2814,83 @@ void CPlayerInterface::showWorldViewEx(const std::vector<ObjectPosInfo>& objectP
...
+				currentObjects.insert(std::make_pair(id, volume));
+		};
+
+		auto tt = cb->getTile(tile);
+		if(tt->isWater())
+			updateObjects(ObjectInstanceID());
+
+		auto obj = tt->topVisitableObj(pos == tile);
+		if(obj && obj->ID != Obj::HERO)
+			updateObjects(obj->id);
```
DjWarmonger: id is unique identifier of objetc instance. Shouldn't sounds be differentiated by <ID, subID> pair?

ArseniyShestakov: More likely I'll differentiate them by sound filename later on as I don't think we want multiple channels of same sound effect playing simultaneously even if multiple objects with same ambient present.

ArseniyShestakov:
So I'm finished checking map objects and made full list:
https://gist.github.com/ArseniyShestakov/616dd31231d5ac5942ad

config/objects/dwellings.json
```diff
@@ -2,6 +2,7 @@
 	"creatureGeneratorCommon" : {
 		"index" :17,
 		"handler": "dwelling",
+		"ambientSound" : "LOOPSWOR",
```
alexvins: As you adding sounds on sybtype level, do not add it on top level. Add default sound in base struct below.

config/objects/moddables.json
```diff
@@ -183,19 +183,20 @@
...
-			"goldMine" :      { "index" : 6, "rmg" : { "value" : 7000 } },
-			"abandoned" :	  { "index" : 7}
+			"sawmill" :       { "index" : 0, "rmg" : { "value" : 1500 }, "ambientSound" : "LOOPLUMB" },
+			"alchemistLab" :  { "index" : 1, "rmg" : { "value" : 3500 }, "ambientSound" : "LOOPSTAR" },
+			"orePit" :        { "index" : 2, "rmg" : { "value" : 1500 }, "ambientSound" : "LOOPSULF" },
+			"sulfurDune" :    { "index" : 3, "rmg" : { "value" : 3500 }, "ambientSound" : "LOOPSULF" },
+			"crystalCavern" : { "index" : 4, "rmg" : { "value" : 3500 }, "ambientSound" : "LOOPCRYS" },
+			"gemPond" :       { "index" : 5, "rmg" : { "value" : 3500 }, "ambientSound" : "LOOPGEMP" },
+			"goldMine" :      { "index" : 6, "rmg" : { "value" : 7000 }, "ambientSound" : "LOOPMINE" },
+			"abandoned" :	  { "index" : 7,}
```
alexvins: redundant comma

lib/IGameCallback.cpp
```diff
@@ -65,7 +65,9 @@ void CPrivilagedInfoCallback::getTilesInRange( std::unordered_set<int3, ShashInt
 			{
 				int3 tilePos(xd,yd,pos.z);
 				double distance;
-				if(patrolDistance)
+				if(distanceFormula == 2) // ambient soud distance
```
alexvins: distanceFormula shouldbe of some enum class type

alexvins: Also extract calculation with formula selection to int3 method

lib/IGameCallback.cpp
```diff
@@ -47,11 +47,11 @@ void CPrivilagedInfoCallback::getFreeTiles (std::vector<int3> &tiles) const
 	}
 }
 
-void CPrivilagedInfoCallback::getTilesInRange( std::unordered_set<int3, ShashInt3> &tiles, int3 pos, int radious, boost::optional<PlayerColor> player/*=uninit*/, int mode/*=0*/, bool patrolDistance/*=false*/) const
+void CPrivilagedInfoCallback::getTilesInRange(std::unordered_set<int3, ShashInt3> &tiles, int3 pos, int radious, boost::optional<PlayerColor> player/*=uninit*/, int mode/*=0*/, int distanceFormula/*=0*/) const
```
alexvins: I understand why we have unordered here (FoW calculation), but in most cases vector<int3> is enough and much faster. It would be good to have getTilesInRange with vector container too.

ArseniyShestakov:
Rebased on top of develop. Going to work on this now.

road debug
```diff
@@ -0,0 +1,291 @@
+OOOO  OOO  O OOOOOOOOOOOOOOOOO OOOOOOOOOOOOOOOO OOOOOOOOOOOOOOOOOO  OOOOOOOOO OO OO OOOOOOOOOO    OOOOOOOOOOOOOOOOOOOOOOOOOOOOOO  OOOOOOOOOOOOOO
```
alexvins: this file should not be commied.

ArseniyShestakov: Yeah accidentally added it while in rebase. Fixed.

lib/mapObjects/CObjectHandler.h
```diff
@@ -183,6 +186,7 @@ class DLL_LINKAGE CGObjectInstance : public IObjectInterface
 		}
 
 		h & pos & ID & subID & id & tempOwner & blockVisit & appearance;
+		h & ambientSound;
```
alexvins: version check missing

lib/mapObjects/CObjectHandler.cpp
```diff
@@ -215,6 +216,8 @@ void CGObjectInstance::initObj(CRandomGenerator & rand)
 		blockVisit = true;
 		break;
 	}
+
+	VLC->objtypeh->getHandlerFor(ID, subID)->setAmbient(this);
```
alexvins: too much code duplication, if it too difficult to ensure CGObjectInstance::initObj call for all descendants move this initialization to CGameState::initMapObjects()

ArseniyShestakov:
I had no internet for about 12 hours so implemented almost everything I wanted.
Also tested probably every map object and added all missing visit sounds.
Still need to rework playback and channel management code.

lib/mapObjects/CObjectHandler.cpp
ArseniyShestakov: Yeah I find out that doing that just after initObj call is way better since we initialize objects just in few places. Though there is problem that for whatever reason "CObjectClassesHandler::getHandlerFor" can't find handler for some objects and it's fail on runtime error.

lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -202,6 +203,12 @@ CObjectClassesHandler::ObjectContainter * CObjectClassesHandler::loadFromJson(co
 	{
 		loadObjectEntry(entry.first, entry.second, obj);
 	}
+
+	if(!json["sounds"]["ambient"].String().empty())
+		obj->sounds.ambient = json["sounds"]["ambient"].String();
+	if(!json["sounds"]["visit"].String().empty())
+		obj->sounds.visit = json["sounds"]["visit"].String();
+
```
alexvins: No need to check for empty string

lib/mapObjects/MiscObjects.cpp
```diff
@@ -713,7 +714,7 @@ void CGMine::initObj(CRandomGenerator & rand)
 			tempOwner = PlayerColor::NEUTRAL;
 	}
 
-	producedQuantity = defaultResProduction();
+	producedQuantity = defaultResProduction();;
```
alexvins: typo

ArseniyShestakov:
So this one now rebased almost finished. Now sound IDs only stored in appropriate handlers and not in object instances since I guess this is what @DjWarmonger suggested two years ago. Also moved distance formula to ```int3```, implemented pickup sounds into schema, fixed few sounds that were still missing, etc.

Not perfect code, but works. Only problem is some glitches in sounds after battle is finished, but that might occur because some other code doesn't stop playback properly so I need to investigate.

lib/int3.h
```diff
@@ -92,6 +92,31 @@ class int3
...
 
+	enum EDistanceFormula
+	{
+		DIST_2D = 0,
+		DIST_MANHATTAN, // patrol distance
+		DIST_CHEBYSHEV, // ambient sound distance
+		DIST_2DSQ
+	};
+
+	ui32 dist(const int3 & o, EDistanceFormula formula) const
```
DjWarmonger: Is this formula going to be used elsewhere? If in RMG or AI, I'd already implement it as a template + inlines to avoid additional overhead.
But probably it's not necessary to reuse it, anyway.

client/CPlayerInterface.cpp
```diff
@@ -446,6 +448,15 @@ void CPlayerInterface::heroKilled(const CGHeroInstance* hero)
 		adventureInt->selection = nullptr;
 }
 
+void CPlayerInterface::heroVisit(const CGHeroInstance * visitor, const CGObjectInstance * visitedObj, bool start)
+{
+	EVENT_HANDLER_CALLED_BY_CLIENT;
+	if(start && visitedObj->getVisitSound())
+	{
+		CCS->soundh->playSound(boost::str(boost::format("%s.wav") % visitedObj->getVisitSound().get()));
```
alexvins: I do not like this. Make get[...]Sound return proper value, or pre-process inside `playSound`.

client/CPlayerInterface.cpp
```diff
@@ -2905,3 +2918,103 @@ void CPlayerInterface::showWorldViewEx(const std::vector<ObjectPosInfo>& objectP
...
+		{
+			for(const CGObjectInstance * obj : tt->blockingObjects)
+			{
+				if(obj->getAmbientSound())
+					updateObjects(obj->getAmbientSound().get());
+			}
+		}
+		// for nos steal a bit from CGameState::battleGetBattlefieldType which is super slow
+		// Also sound is for every special terrain on tile and not just one on top.
+		for(auto &obj : CGI->mh->map->objects)
```
alexvins: Use CGI->mh->ttiles

lib/int3.h
ArseniyShestakov: I don't think it's will be heavily used since almost everything only need ```dist2d``` / ```dist2dSQ```. Manhattan distance used for patrol only and chebyshev only for ambient sounds that are not really performance critical, but it's still nice to explicitly specify what formula is used especially considering that we might need some other weird formulas in future.

client/CPlayerInterface.cpp
ArseniyShestakov: I also want to investigate how this would work with mods. Actually I think our filesystem able to play identical music even if it's has different extension / container, but I will need to test it.

lib/mapObjects/CRewardableObject.h
```diff
@@ -271,7 +269,11 @@ class DLL_LINKAGE CRewardableObject : public CArmedInstance
 		h & onVisited;
 		h & onEmpty;
 		h & visitMode;
-		h & soundID;
+		if(version < 778 && !h.saving)
```
alexvins: Remove `&& !h.saving` condition. Initlialize soundID.

lib/mapObjects/CObjectClassesHandler.h
```diff
@@ -172,6 +190,10 @@ class DLL_LINKAGE AObjectTypeHandler : public boost::noncopyable
 			h & typeName;
 			h & subTypeName;
 		}
+		if(version >= 778)
```
alexvins: The problem is that for any old save nearly all sounds will not work. You removed all hardcoded sounds, but do not provide defaults for old saves.

ArseniyShestakov: That's true, but there no easy way to partially reload JSON...

lib/int3.h
```diff
@@ -109,6 +134,11 @@ class int3
 	{
 		return abs(o.x - x) + abs(o.y - y);
 	}
+	//chebyshev distance used for ambient sounds (z coord is not used)
+	double chebdist2d(const int3 & o) const
+	{
+		return abs(std::max(abs(o.x) - abs(x), abs(o.y) - abs(y)));
```
henningkoehlernz: This should be return std::max(abs(o.x - x), abs(o.y - y));

ArseniyShestakov:
So I improved the code, moved most to sound handler and also fixed not so obvious bug.
Now I just need to find out what's going wrong when hero surrender / flee.

lib/mapObjects/CRewardableObject.h
ArseniyShestakov: Yeah I get ``` !h.saving``` likely make no sense, but we do have it in some other code. Should we get rid of it too?

ArseniyShestakov:
So just fixed problem with sound by proper locking.
Now everything almost ready except for compatibility.

client/CPlayerInterface.cpp
```diff
@@ -2905,3 +2914,52 @@ void CPlayerInterface::showWorldViewEx(const std::vector<ObjectPosInfo>& objectP
...
+	cb->getVisibleTilesInRange(tiles, pos, CCS->soundh->ambientGetRange(), int3::DIST_CHEBYSHEV);
+	for(int3 tile : tiles)
+	{
+		int dist = pos.dist(tile, int3::DIST_CHEBYSHEV);
+		// We want sound for every special terrain on tile and not just one on top
+		for(auto & ttObj : CGI->mh->ttiles[tile.x][tile.y][tile.z].objects)
+		{
+			auto obj = ttObj.obj;
+			if(!obj || !obj->getAmbientSound())
+				continue;
```
ArseniyShestakov: This one I especially dislike and this check doesn't help. Problem is that map handler information might sill have object that already removed.

I not sure how exactly to avoid that.

lib/mapObjects/CRewardableObject.h
alexvins: Not in this branch

client/CPlayerInterface.cpp
alexvins: Copy sound-related data to `TerrainTileObject`.

ArseniyShestakov:
So now everything is ready except for the fact that loaded games won't have visit sounds.

DjWarmonger:
Okay, but lack of sounds for old saves is not a regression in any way. If that was merged few months back, nobody would even remember by now :P

dydzio0614:
I think the deal are saves from bug reports, though I guess it does not affect anybody else than people who check bugs. I personally do not mind lacking visit sounds when checking bug reports.

alexvins:
@DjWarmonger please fix commit history

DjWarmonger:
That's exactly what I was trying to do this morning :v

DjWarmonger:
How can I do that?

alexvins:
1) DO NOT do `git pull`, it will merge remote changes to yours. (but you can do `git pull --ff-only`)
2) do `fetch`, then `rebase`

alexvins:
* now you need to forced rebase thus branch over develop. and (re)push

alexvins:
The only use of `git pull` in our workflow is 'git pull --ff-only' for develop branch is there **are no local changes** and **current local branch is also develop**

DjWarmonger:
But I always get this message:

```
git.exe push --progress "origin" feature/ambientSounds:feature/ambientSounds

To https://github.com/vcmi/vcmi/
! [rejected]          feature/ambientSounds -> feature/ambientSounds (non-fast-forward)
error: failed to push some refs to 'https://github.com/vcmi/vcmi/'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```

alexvins:
should be `git.exe push --force-with-lease ...`. `force-with-lease` allows to rewrite commit history

alexvins:
In Tortoisegit it is `Force: may discard known changes`

DjWarmonger:
I think it works! Thanks.

alexvins:
* By default `push` do not overwrite commit history
* if you want do to that add f.e. `--force-with-lease`
* for protected branches (`develop` and `master` in out repo) history overwrite is forbidden entirely.

PR #196, opened 2016-02-21
Feature/armies merging

ArseniyShestakov:
At first I'm want to make merging working as in H3 which is ready as well as few fixes for siege / battle handling. There is bigger plan to implement more merging options since H3 algorithm don't merge stacks of same type and you may end up with smaller army if you split stacks too much.

### DIFF
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -802,6 +802,14 @@ void CGTownInstance::removeCapitols (PlayerColor owner) const
 	}
 }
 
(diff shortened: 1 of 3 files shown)


ArseniyShestakov:
So I didn't found appropriate way to make sure battleFinished is called for CGHeroInstance since it's map object removed in endBattle. However everything else must work properly now and behaviour is replicate H3.

ArseniyShestakov:
So I decide to test this one more and then merge without adding extra bloat. Advanced options will be in separate pull request since they require refactoring of current army merging code.

PR #197, opened 2016-02-27
Issues/2304/2418

ArseniyShestakov:
Since it's easier to put feedback via pull requests here we go:
http://bugs.vcmi.eu/view.php?id=2304
http://bugs.vcmi.eu/view.php?id=2418

### DIFF
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -1018,9 +1018,9 @@ void CBattleInterface::stackRemoved(int stackID)
 		}
 	}
 
-	delete creAnims[stackID];
-	creAnims.erase(stackID);
-	creDir.erase(stackID);
+//	delete creAnims[stackID];
+//	creAnims.erase(stackID);
+//	creDir.erase(stackID);
 	redrawBackgroundWithHexes(activeStack);
 	queue->update();
 }
@@ -3408,8 +3408,13 @@ BattleObjectsByHex CBattleInterface::sortObjectsByHex()
 
 	BattleObjectsByHex sorted;
 
+	auto stacks = curInt->cb->battleGetStacksIf([](const CStack * s)
+	{
+		return !s->isTurret();
+	});
+
 	// Sort creatures
-	for (auto & stack : curInt->cb->battleGetAllStacks())
+	for (auto & stack : stacks)
 	{
 		if(creAnims.find(stack->ID) == creAnims.end()) //e.g. for summoned but not yet handled stacks
 			continue;
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -1018,6 +1018,8 @@ void CBattleInterface::stackRemoved(int stackID)
 		}
 	}
 
+	//todo: ensure that ghost stack animation has fadeout effect
+
 	redrawBackgroundWithHexes(activeStack);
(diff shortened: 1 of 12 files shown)


ArseniyShestakov:
So first problem is that we can't handle summoned creatures exactly as clones because it's breaking one-elemental limit check. Since stack disappear we can no longer check if there was creature summoned or no. Related issue:
http://bugs.vcmi.eu/view.php?id=2293

There also related not implemented feature. H3 had special limit of how many stacks could ever fight for one side on battlefield which is 20 stacks. So it's another reason to make some storage for all stacks.

DjWarmonger:
20 stack limit actually made WoG crash in some scenarios and in general was troublesome. I don't think we need to impose it, it has no impact in OH3 - but could give additional possibilities in mods.

alexvins:
>that we can't handle summoned creatures exactly as clones

We can, but just do not delete them. Will do this way.

ArseniyShestakov:
@DjWarmonger I didn't mean that we must enforce it by default, but we still need to have options that make VCMI behaviour as close to H3 as possible. And some limit is totally reasonable from balance standpoint anyway.

alexvins:
@ArseniyShestakov it is even possible with current mechanics to have more than 20 stacks on one side without mods?

ArseniyShestakov:
Idea is that 20 is total limit of stacks, not just alive one. 7 stacks is your army, up to 3 war machines.
So you can summon creatures up to 10 times, but after that point you won't be able to summon anything, even demons.

alexvins:
7 normal + 7 clones + 2 war machines + 3 arrow towers + 1 elemental + 1 daemon = 20. It is just a mechanics limitation. (+1 commander = crash in WoG mentioned bu Warmonger).

dydzio0614:
Maybe mechanical limit makes sense in cases like the one here: [https://www.youtube.com/watch?v=tuy_ncoBhGM](url)

alexvins:
@dydzio0614 @ArseniyShestakov Anyway, 20 stack limit is off-topic here.

ArseniyShestakov:
So at moment elemental summon is broken if you already had clone. Also probably we need to keep in mind that you may first cast clone on "real" elemental and then summon one.

ArseniyShestakov:
Another problem with current code: clones remain visible after original stack is killed. You can't control them though so they look like inactive bitmap.

alexvins:
Feature and re-factoring freeze.

ArseniyShestakov:
So now clone stacks disappear only when round is ended so you can able to do weird things with corpse before they disappear. E.g resurrect stack with 0 creatures in it.

ArseniyShestakov:
So it's works now. Though there is small text bug:
pr_197_bugtext

Repro:

* Make clone of stack.
* Summon first kind of elemental.
* Attempt to summon different elemental.

So similar to issue [2293](http://bugs.vcmi.eu/view.php?id=2293).

ArseniyShestakov:
Everything else looks fine so we can merge it if text bug can be fixed later on together with one I mentioned.

alexvins:
Yes, "Gnoll" glitch is same code as 2293, but different bug. However "Gnoll" glitch is a regression so I`ll fix it and may be also 2293 here.

ArseniyShestakov:
So I think it's fixed! May be it's time to merge it.

PR #199, opened 2016-03-08
Fixed segmentation fault when right-clicking on artefact on the "Black Market" window.

argarak:
In file `client/windows/CTradeWindow.cpp`, which initialises the "Black Market" window, in the function `clickRight`, the pointer "hlp" was null. Later on in your code you referenced the function `getEffectiveDescription`, and inside it you assigned a *null* pointer to a string and caused the segmentation fault. (I think...)

Therefore, I looked into your code and found a way of initialising the message using the variable `CGI`.

Eventhough it way seem like I changed a lot, I only really added this code (inside if statement):

```
if(id >= 0)  {
	    //adventureInt->handleRightClick(hlp->getEffectiveDescription(), down);
	
	    const CArtifact &artifact = *CGI->arth->artifacts[id];
	    std::string desc = artifact.Description();
	    
	    adventureInt->handleRightClick(desc, down);
}
```

The rest of the code is Emacs "fixing" indentation, but source code always ends up looking weird after you indent it in Emacs. (To fix this just indent everything in your favourite editor)

And since there was already a variable named `id`, I can use this method. If you would like to make this even smaller, remove the comment and the variable `desc` and just put in into `handleRightClick`.

After this modification, right-clicking on the artefact worked perfectly well.


ArseniyShestakov:
Can you please rebase on top of latest develop?
Better if you could fast-forward your develop and apply that fix in some new branch.

argarak:
I made a new branch in argarak/vcmi called `market-fix`. I ran the command `git rebase -i HEAD~5` which gave the output: `Successfully rebased and updated refs/heads/market-fix.`

Is this what you were looking for? If not, could you please tell me more specifically what commands I have to run? I'm a newbie with Git.

ArseniyShestakov:
What I mean is to use "git rebase --interactive" and keep only needed changes in one file dropping all other commits. This is of course if you want us to merge your pull request and not just apply fix. Of course if you want your pull request merged it's shouldn't change code formatting as well.

If you work on Linux I strongly recommend to use Qt Creator if you don't want to spend much time on configuration. But if you only want one single fix applied I may commit it separately.

argarak:
Alright, instead of me wasting my time attempting to mess this up even further, create a commit and replace the code at CTradeWindow.cpp:272 with:

```
const CArtifact &artifact = *CGI->arth->artifacts[id];
adventureInt->handleRightClick(artifact.Description(), down);
```

And then you may close this pull request.

ArseniyShestakov:
No problem. I'll look into that one a bit more carefully. Thanks for report.

DjWarmonger:
Just create new branch, pull latest "develop", change these two lines, then push.
This PR needs to be closed sooner or later.

ArseniyShestakov:
We need to use CArtifactInstance::getEffectiveDescription anyway not just CArtifact::Description. I'll commit fix soon, just a bit busy atm.

ArseniyShestakov:
End up it's not that easy to do what I wanted since CArtifactInstance doesn't used for market. E.g market only store IDs of artifacts that available for sale, but not actual instances. So just implemented it your way for now.

https://github.com/vcmi/vcmi/commit/faae305ec0dad5203ce15e3ee1d89f88e4104327

PR #202, opened 2016-07-29
Noncopyable bonus system node2

alexvins:
Now without serialization changes.

### DIFF
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -1075,13 +1075,14 @@ void CGameState::initPlayerStates()
 	logGlobal->debug("\tCreating player entries in gs");
 	for(auto & elem : scenarioOps->playerInfos)
 	{
-		std::pair<PlayerColor, PlayerState> ins(elem.first,PlayerState());
-		ins.second.color=ins.first;
-		ins.second.human = elem.second.playerID;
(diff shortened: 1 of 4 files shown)


ArseniyShestakov:
This one is good so far. I tested plenty of old save files and everything seems fine.
Though can't say yet how good these changes at preventing save corruption.

alexvins:
Complete

ArseniyShestakov:
Update on branch. I tested it quite a bit and Alex fixed it further. So now it's actually let you load every previously "corrupted" save linked in in [#2314](http://bugs.vcmi.eu/view.php?id=2314), but there is still issue that cause crashes in bonus system that might be related changes.

In same time there is plenty of lockup and crashes within AI that prevent me from automated testing so I'll have to fix them first.

DjWarmonger:
Is there a reason not to merge this branch now?

ArseniyShestakov:
I still wish to do some testing after I finish fixing desyncs since I want to make sure all problems with bonuses actually fixed. It's was hard to test it properly because of few freezes and crashes within VCAI.

Though if you think we should merge it now then I could try to prioritize testing instead of desyncs.

BTW @DjWarmonger can you please check #203 with MSVC?

DjWarmonger:
Oh I see. Take your time then :)

PR #203, opened 2016-08-12
boost::format support for CLogger

alexvins:
We are tend to use boost::format for logging, so here is new log API extension to replace CLoggerStream

- also it is a first step to organize vstd namespace. New include directory introduced for such library headers.
- this PR includes changes in all build systems so I`m waiting for feedback at least about (un)successful compilation.


include/vstd/CLoggerBase.h
```diff
@@ -0,0 +1,114 @@
...
+namespace vstd
+{
+	class DLL_LINKAGE CLoggerBase
+	{
+	public:
+		virtual ~CLoggerBase(){};
+
+		virtual void log(ELogLevel::ELogLevel level, const std::string & message) const = 0;
+
+		template<typename T, typename ... Args>
```
DjWarmonger: Won't all these templates significantly increase compiled code size?

alexvins: I don`t think so. Templates itself should be completely optimized out.

ArseniyShestakov:
Just want to confirm this compile fine with Clang on Linux

alexvins:
@DjWarmonger Any objections on merging this?

DjWarmonger:
I need to find a time to check this. In my calendar ;)

DjWarmonger:
This compiles just fine under MSVS. Ready for merge.

PR #204, opened 2016-08-29
Do not use atexit

krnowak:
Just call these functions directly in the handleQuit function.

std::atexit generally does not play nice with static objects in other
translation units/libraries - there are no guarantees about the order
of destruction of static objects and atexit functions coming from
different translation units/libraries. Removing calls to atexit fixes
crashes when quitting the game. The crash happened during the dispose
function somewhere inside boost's locale library that is used
indirectly by the logger - some locale function tried to lock a static
mutex that was already destroyed.

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -290,7 +290,6 @@ int main(int argc, char** argv)
 	console = new CConsoleHandler;
 	*console->cb = processCommand;
 	console->start();
-	atexit(dispose);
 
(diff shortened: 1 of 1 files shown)


client/CMT.cpp
```diff
@@ -1278,6 +1276,8 @@ void handleQuit(bool ask/* = true*/)
 			SDL_Quit();
 
 		std::cout << "Ending...\n";
+		dispose();
+		SDL_Quit();
```
alexvins: wrong and duplicated call, see line above

alexvins:
I`ll apply modified fix. Thank you.

PR #205, opened 2016-08-30
Allow choosing the display for the game window

krnowak:
This adds a new setting to video settings that allows the user to choose the display for the game. Also adds the relevant UI in the launcher.

### DIFF
--- a/launcher/CMakeLists.txt
+++ b/launcher/CMakeLists.txt
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 2.8.7)
 
 include_directories(${CMAKE_HOME_DIRECTORY} ${CMAKE_HOME_DIRECTORY}/include ${CMAKE_CURRENT_SOURCE_DIR})
 include_directories(${ZLIB_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${Qt5Widgets_INCLUDE_DIRS} ${Qt5Network_INCLUDE_DIRS})
+include_directories(${SDL_INCLUDE_DIR})
 
 set(launcher_modmanager_SRCS
 	modManager/cdownloadmanager_moc.cpp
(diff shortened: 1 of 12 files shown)


launcher/main.cpp
```diff
@@ -1,11 +1,13 @@
...
 #include "mainwindow_moc.h"
 #include <QApplication>
+#include "sdldisplayquery.h"
 
 int main(int argc, char *argv[])
 {
+	auto displayList {getDisplays()};
 	QApplication a(argc, argv);
-	MainWindow w;
+	MainWindow w {displayList};
```
alexvins: TravisCI reports error with clang here.

launcher/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ cmake_minimum_required(VERSION 2.8.7)
 
 include_directories(${CMAKE_HOME_DIRECTORY} ${CMAKE_HOME_DIRECTORY}/include ${CMAKE_CURRENT_SOURCE_DIR})
 include_directories(${ZLIB_INCLUDE_DIR} ${Boost_INCLUDE_DIRS} ${Qt5Widgets_INCLUDE_DIRS} ${Qt5Network_INCLUDE_DIRS})
+include_directories(${SDL_INCLUDE_DIR})
```
alexvins: Linkage with SDL missing?

launcher/main.cpp
krnowak: Yeah, the usual mistake with `auto foo {bar}` resulting in `std::initializer_list<…>`.

launcher/CMakeLists.txt
krnowak: Oops, yeah. Strange that it worked for me though…

krnowak:
Updated.

krnowak:
Another idea: instead of the `displayIndex` number, we could probably store something like:
```
"display": {
    "width": 1600,
    "height": 1200,
    "topleftx": 1920,
    "toplefty": 0
}
```

This basically tells the game that I want to use the 1600x1200 display that is to the right of the 1920x1200 display. This setup would probably be immune to the differences in display numbering in Wayland and X11. But not sure if this is really important.

alexvins:
questionable idea, not for this branch, please file a feature request.

PR #209, opened 2016-09-07
Fix save desyncs

ArseniyShestakov:
Many small related fixes already made it into develop. These are remaining large changes that far from perfect, but so far they fix all major desyncs. In case you can find any desync in newly created game please post backtrace from within CLoadIntegrityValidator::read.

I would happily accept suggestions on:

- Code style / formatting / etc.
- ~~Thoughts on better place for RNG since CCallbackBase is questionable.~~
- Where to move CGameState::battleGetBattlefieldType since it's not really related to battle itself. Probably we need different class / file exclusively for adventure map mechanics.

More information on changes available on [mantis#2459](http://bugs.vcmi.eu/view.php?id=2459). There is still some potencial desyncs that will affect multiplayer game, but so far I was only work on single client problems only.

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -653,7 +653,7 @@ AttackPossibility AttackPossibility::evaluate(const BattleAttackInfo &AttackInfo
 	for(int i  = 0; i < totalAttacks; i++)
 	{
 		std::pair<ui32, ui32> retaliation(0,0);
-		auto attackDmg = cbc->battleEstimateDamage(curBai, &retaliation);
+		auto attackDmg = cbc->battleEstimateDamage(CRandomGenerator::getDefault(), curBai, &retaliation);
 		ap.damageDealt = (attackDmg.first + attackDmg.second) / 2;
 		ap.damageReceived = (retaliation.first + retaliation.second) / 2;
 
@@ -737,7 +737,7 @@ int PotentialTargets::bestActionValue() const
 
 void EnemyInfo::calcDmg(const CStack * ourStack)
 {
-	TDmgRange retal, dmg = cbc->battleEstimateDamage(ourStack, s, &retal);
+	TDmgRange retal, dmg = cbc->battleEstimateDamage(CRandomGenerator::getDefault(), ourStack, s, &retal);
 	adi = (dmg.first + dmg.second) / 2;
 	adr = (retal.first + retal.second) / 2;
 }
--- a/AI/StupidAI/StupidAI.cpp
+++ b/AI/StupidAI/StupidAI.cpp
@@ -44,7 +44,7 @@ struct EnemyInfo
 	{}
 	void calcDmg(const CStack * ourStack)
 	{
-		TDmgRange retal, dmg = cbc->battleEstimateDamage(ourStack, s, &retal);
+		TDmgRange retal, dmg = cbc->battleEstimateDamage(CRandomGenerator::getDefault(), ourStack, s, &retal);
 		adi = (dmg.first + dmg.second) / 2;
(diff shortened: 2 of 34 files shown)


ArseniyShestakov:
Please avoid commenting code itself yet since I messed rebase a bi and want to fix it.

ArseniyShestakov:
Now I'm going to land massive commit that in same change don't really change game behaviour. When I started initial desync bugfixes I was thinking about preventing this kind of issues in future and one of solutions for that is to make sure RNG from GS not used outside.

So with suggestions from @alexvins this is how RNG must be used now:

* RNG will be explicitly passed on map object initialization.
* Only GS and netpacks should use RNG from GS.
* All server-side code include map objects should be using CRandomGenerator::getDefault() which going to be serialized on server-side one way or another.
* Client-side that not affect reproducibility going to use same CRandomGenerator::getDefault() which of course not going to be serialized.

TODO:

1. There so far one case in CBank where hacky getter from gamestate used. Probably we need proper netpack for that kind of object reset.
2. AIs use RNG in few cases, but for reproducibility we need to serialize it and make sure player actions in GUI not affect it.
3. There is several usages of random_shuffle and probably some other functions that make game unreproducible which have to be converted to use proper RNG.

Some of this might fixed or not fixed within this pull request.

alexvins:
Too much useless changes just renaming rng -> rand, please revert them

lib/mapObjects/CRewardableConstructor.cpp
```diff
@@ -44,7 +44,7 @@ void CRandomRewardObjectInfo::init(const JsonNode & objectConfig)
 	parameters = objectConfig;
 }
 
-void CRandomRewardObjectInfo::configureObject(CRewardableObject * object, CRandomGenerator & rng) const
+void CRandomRewardObjectInfo::configureObject(CRewardableObject * object, CRandomGenerator & rand) const
```
DjWarmonger: Agree with AVS, these do not change anything.

ArseniyShestakov:
Sorry, I just initially used "rng" for my own code and then decide to switch it to "rand" since it's more readoble so it's end up easier to change it everywhere. I'll revert if you feel it's important.

alexvins:
It makes more difficult to review.

DjWarmonger:
There are other ways to do that. Firstly, this function takes rng as argument and it doesn't matter how it's called internally - neither needs to be changed.
Alternatively you coudl create rng as singleton (static variable?) and not pass it around at all, but it looks like overkill. All we need it to unify rng calls once.

Would it help to detec bugs if you made rng non-copyable?

ArseniyShestakov:
> There are other ways to do that. Firstly, this function takes rng as argument and it doesn't matter how it's called internally - neither needs to be changed.

As I said I rebased my own changes and accidentally patched other code that I shouldn't touch.

> Alternatively you coudl create rng as singleton (static variable?) and not pass it around at all, but it looks like overkill. All we need it to unify rng calls once.

Actually idea is quite opposite: we need to make sure that nobody accidentally use wrong RNG ever. This is needed since unfortunately a lot of server-side, map objects and battlle code is a mess that need massive refactoring. For instance:

* Some map object functions only used from within CGameState only.
* Some used on server-side only.
* Some used both within gamestate and server-side code.
* Something probably even used by both client and server.

Problem that gamestate RNG should be only used from code that apply changes for all clients and server-side RNG state should be secret. It's would be even better if we could simply limit any access to  GS RNG the way so you simply can't use it wrong way, but with codebase we have now this is not possible.

Also anything static is super evil...

ArseniyShestakov:
Now rebased without useless changes.

alexvins:
Thanks. Reviewed successfully.

PR #211, opened 2016-09-10
Check if building's dependencies are allowed

Chocimier:
Fix for bug 2321

### DIFF
--- a/lib/CGameInfoCallback.cpp
+++ b/lib/CGameInfoCallback.cpp
@@ -520,10 +520,12 @@ EBuildingState::EBuildingState CGameInfoCallback::canBuildStructure( const CGTow
 	if(vstd::contains(t->forbiddenBuildings, ID))
 		return EBuildingState::FORBIDDEN; //forbidden
 
-	std::function<bool(BuildingID id)> allowedTest;
-	std::function<bool(BuildingID id)> possiblyNotBuiltTest;
+	auto possiblyNotBuiltTest = [&](BuildingID id) -> bool
+	{
+		return ((id == BuildingID::CAPITOL) ? true : !t->hasBuilt(id));
+	};
 
-	allowedTest = [&](BuildingID id) -> bool
+	std::function<bool(BuildingID id)> allowedTest = [&](BuildingID id) -> bool
 	{
 		if (vstd::contains(t->forbiddenBuildings, id))
 		{
@@ -531,12 +533,6 @@ EBuildingState::EBuildingState CGameInfoCallback::canBuildStructure( const CGTow
 		}
 
 		return t->genBuildingRequirements(id, true).satisfiable(allowedTest, possiblyNotBuiltTest);
-	};
-
-	possiblyNotBuiltTest = [&](BuildingID id) -> bool
-	{
-		//TODO consider destroing
-		return !t->hasBuilt(id);
 	};
 
 	if (!t->genBuildingRequirements(ID, true).satisfiable(allowedTest, possiblyNotBuiltTest))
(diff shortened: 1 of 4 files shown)


lib/CGameInfoCallback.cpp
```diff
@@ -520,6 +520,28 @@ EBuildingState::EBuildingState CGameInfoCallback::canBuildStructure( const CGTow
 	if(vstd::contains(t->forbiddenBuildings, ID))
 		return EBuildingState::FORBIDDEN; //forbidden
 
+	std::function<bool(BuildingID id)> allowedTest;
+	std::function<bool(BuildingID id)> possiblyNotBuiltTest;
+
+	allowedTest = [&](BuildingID id) -> bool
```
alexvins: We declare lambdas as "auto"

lib/CGameInfoCallback.cpp
```diff
@@ -520,6 +520,28 @@ EBuildingState::EBuildingState CGameInfoCallback::canBuildStructure( const CGTow
...
+
+		return t->genBuildingRequirements(id, true).satisfiable(allowedTest, possiblyNotBuiltTest);
+	};
+
+	possiblyNotBuiltTest = [&](BuildingID id) -> bool
+	{
+		//TODO consider destroing
+		return !t->hasBuilt(id);
+	};
+
```
alexvins: Define this before allowedTest

lib/CGameInfoCallback.cpp
```diff
@@ -520,6 +520,28 @@ EBuildingState::EBuildingState CGameInfoCallback::canBuildStructure( const CGTow
...
+		{
+			return false;
+		}
+
+		return t->genBuildingRequirements(id, true).satisfiable(allowedTest, possiblyNotBuiltTest);
+	};
+
+	possiblyNotBuiltTest = [&](BuildingID id) -> bool
+	{
+		//TODO consider destroing
```
alexvins: Please explain this TODO item

lib/LogicalExpression.h
```diff
@@ -99,6 +99,123 @@ namespace LogicalExpressionDetail
...
+	class FalsifiabilityVisitor;
+
+	/// Visitor to test whether expression's value can be true
+	template <typename ContainedClass>
+	class SatisfiabilityVisitor : public boost::static_visitor<bool>
+	{
+		typedef ExpressionBase<ContainedClass> Base;
+
+		std::function<bool(const typename Base::Value &)> satisfiabilityTest;
+		FalsifiabilityVisitor<ContainedClass> *falsifiabilityVisitor;
```
alexvins: falsifiability[Visitor, Test], satisfiability[Visitor, Test], countSatisfiable, countFalsifiable should be extracted to common base class template.

smth like:

```
template <typename ContainedClass>
	class BaseVisitor : public boost::static_visitor<bool> ...

template <typename ContainedClass>
	class SatisfiabilityVisitor : public BaseVisitorContainedClass>
```

lib/CGameInfoCallback.cpp
Chocimier: This is about building A which need building B not to be built; can be expressed by genBuildingRequirements(A) = NoneOf(B).

If VCMI allow to destroy building, possiblyNotBuiltTest should rather return (!t->hasBuilt(id) || t->canBeDestroyed(id)) or just (true) if any building can be destroyed.

Assuming that is ever planned.

DjWarmonger: - There are WoG scripts which allow to destroy building. Not a priority, but...
- In original game, Capitol is destroyed if player conquering a town already has one.

PR #194, opened 2016-02-04
Refactoring/serialization

DjWarmonger:
Testing.

PR #212, opened 2016-09-11
Refactoring/serialization2

ArseniyShestakov:
Rebased code for @IvanSavenko serializer refactoring from PR #194.
It's still have old issue with YourTurn netpack, but works.

Old serializer files are still there since I still trying to make sure all other changes are applied properly.


ArseniyShestakov:
So Alex suggested try to get this one ready to merge and here it is. I ported all changes exclude one within CLoadIntegrityValidator since errors there are client / server synchronization problems most of time and not actual save game mismatch.

Also in past this code wasn't able to load my test map and probably some saves, but looks like some of changes done to bonuses / pointers serialization fixed these problems. Anyway I suggest to merge this one and let others test it.

ArseniyShestakov:
So this one going to break OS X build if merged.

alexvins:
Looks like Apple`s clang fails with "std::vector<bool>" it do not use `std::enable_if < std::is_same<T, std::vector<bool> >::value, int  >` overload but uses std::vector one

ArseniyShestakov:
So Ivan fixed problem that caused issues in multiplayer.
Now one of things left is OS X / Xcode bug.

Another one is find out why save loading fail in MXE builds for Windows.

ArseniyShestakov:
So everything is fine now except Povelitel's problem with saves.
Somehow I managed to break non-English paths support for save games.

lib/serializer/BinaryDeserializer.h
```diff
@@ -0,0 +1,535 @@
...
+	}
+	template <typename T, size_t N>
+	void load(std::array<T, N> &data)
+	{
+		for(ui32 i = 0; i < N; i++)
+			load( data[i] );
+	}
+	template <typename T>
+	void load(std::set<T> &data)
+	{
```
alexvins: Code duplication: sequence container (set, unordered_set, list) overloads are actually identical

lib/serializer/BinaryDeserializer.h
```diff
@@ -0,0 +1,535 @@
...
+	template <typename T1, typename T2>
+	void load(std::pair<T1,T2> &data)
+	{
+		load(data.first);
+		load(data.second);
+	}
+
+	template <typename T1, typename T2>
+	void load(std::map<T1,T2> &data)
+	{
```
alexvins: same as for sequence containers

lib/serializer/BinarySerializer.h
```diff
@@ -0,0 +1,373 @@
...
+			save(data[i]);
+	}
+	template <typename T, size_t N>
+	void save(const std::array<T, N> &data)
+	{
+		for(ui32 i=0; i < N; i++)
+			save(data[i]);
+	}
+	template <typename T>
+	void save(const std::set<T> &data)
```
alexvins: same as in deserializer Code duplication: sequence container (set, unordered_set, list) overloads and map overloads are actually identical

lib/serializer/BinaryDeserializer.h
```diff
@@ -0,0 +1,535 @@
...
+
+	int write(const void * data, unsigned size);
+
+public:
+	bool reverseEndianess; //if source has different endianness than us, we reverse bytes
+	si32 fileVersion;
+
+	std::map<ui32, void*> loadedPointers;
+	std::map<ui32, const std::type_info*> loadedPointersTypes;
+	std::map<const void*, boost::any> loadedSharedPointers;
```
alexvins: Possible memory leak in case of shared_ptr serialised in network connection. If it is easy to fix it should be fixed before merge.

lib/serializer/BinaryDeserializer.h
IvanSavenko: True, but there is no way to specialize template for multiple types. SFINAE with enable_if would work here but it would turn this trivial code into template metaprogramming jungles.

lib/serializer/BinaryDeserializer.h
IvanSavenko: Details? This map is used to find already created shared_ptr for newly loaded pointer. I don't see how it may cause memory leak. If there is an element in this map then there is a loaded shared_ptr with same pointer.

alexvins: This map holds all transferred shared pointers for all connection lifetime.

lib/serializer/BinaryDeserializer.h
alexvins: Redirection to private template method (smth. like 
`template <typename Containter, typename T>
void load(Containter & data)` or even `template <typename Containter>` using typedefs inside container  ) should also work

IvanSavenko: True but 
1) this is already a trivial function, less then 10 lines in size
2) this would add one more entry to call stack - precisely what I wanted to minimize in this refactoring.  Originally there were 10-15 internal calls between two subsequent calls of T::serialize() making serialization debugging almost impossible.

alexvins: It may be a macro, but I do not like them.
So we have 3 options:
1) code duplication with no changes
2) longer callstack with redirection
3) less readable with macro

lib/serializer/BinaryDeserializer.h
IvanSavenko: Actually it seems it is not just shared pointers... loadedPointers also stores pointers for object lifetime. And since those are raw pointers this means possible memory corruption if serializer will try to load already deleted object. Also this map can grow a lot during game session since it is never cleared.

Still all of this code was present in old serializer as well so merge won't bring any new bugs and I don't see any easy way to fix these.

lib/serializer/BinaryDeserializer.h
IvanSavenko: or
4) implement template to detect containers with one check via SFINAE. I already made is_serializeable template to detect any class with ::serialize() method. Same approach may work for container.

lib/serializer/BinaryDeserializer.h
alexvins: loadedPointers is less of a problem. We are not transfering much of plain pointers and I have a plan to switch all pointers used in netpacks to shared (if there are still any). As for loadedSharedPointers I was thinking to save then as weak_ptr and do garbage collection in Connection lets say each game day.

lib/serializer/BinaryDeserializer.h
alexvins: sounds interesting, but should be separate PR - enough black magic here.

ArseniyShestakov:
I just finished testing of plenty old saves again and everything looks fine.

Would be cool if we can avoid memory leaks, but this can be always done outside of this pull request too. So if anyone want to land some changes please do now since I'll prefer to merge this one today / tomorrow.

ArseniyShestakov:
@DjWarmonger if you wish to release stable soon can you please look at this one?
I really wish to merge it before release since it's fix MP.

PR #213, opened 2016-09-13
Check if hero knows spell even without wisdom

Chocimier:


ArseniyShestakov:
BTW I sent you PM on our forums so if you have time please check it.

PR #214, opened 2016-09-16
Properly check if corpse was visited

Chocimier:
Fixes two issues:
- corpse without loot couldn't be marked as visited
- taking loot from once visitable objects marked them as visited for all players

Fixes bugs 1975 and 1043.

### DIFF
--- a/lib/mapObjects/CRewardableObject.cpp
+++ b/lib/mapObjects/CRewardableObject.cpp
@@ -317,11 +317,17 @@ bool CRewardableObject::wasVisited(PlayerColor player) const
 		case VISIT_UNLIMITED:
 		case VISIT_BONUS:
 			return false;
-		case VISIT_ONCE:
-		case VISIT_PLAYER:
+		case VISIT_ONCE: // FIXME: hide this info deeper and return same as player?
+//			for (auto & visit : info)
+//			{
+//				if (visit.numOfGrants != 0)
+//					return true;
+//			}
 			return vstd::contains(cb->getPlayer(player)->visitedObjects, ObjectInstanceID(id));
 		case VISIT_HERO:
 			return false;
+		case VISIT_PLAYER:
+			return vstd::contains(cb->getPlayer(player)->visitedObjects, ObjectInstanceID(id));
 		default:
 			return false;
 	}
--- a/lib/mapObjects/CRewardableObject.cpp
+++ b/lib/mapObjects/CRewardableObject.cpp
@@ -317,16 +317,11 @@ bool CRewardableObject::wasVisited(PlayerColor player) const
 		case VISIT_UNLIMITED:
 		case VISIT_BONUS:
 			return false;
-		case VISIT_ONCE: // FIXME: hide this info deeper and return same as player?
-			for (auto & visit : info)
-			{
-				if (visit.numOfGrants != 0)
-					return true;
-			}
-		case VISIT_HERO:
-			return false;
+		case VISIT_ONCE:
 		case VISIT_PLAYER:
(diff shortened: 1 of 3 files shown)


lib/mapObjects/CRewardableObject.cpp
```diff
@@ -317,16 +317,11 @@ bool CRewardableObject::wasVisited(PlayerColor player) const
 		case VISIT_UNLIMITED:
 		case VISIT_BONUS:
 			return false;
-		case VISIT_ONCE: // FIXME: hide this info deeper and return same as player?
```
DjWarmonger: This might fix the Corpse for now, but the code was meant for configurable object support (unfinished). Please leave it commented out at least.

ArseniyShestakov:
I'll need to check how this affect other objects that using onceVisitable.

> taking loot from once visitable objects marked them as visited for all players
I not entirely sure this is most desired behaviour.

It's would be perfect if all players that could "see" tile when visit occur would get it market as visitable.

DjWarmonger:
Agree with Aresniy. Of course object should be market as visitable for all players, but they can't know it until they visit that object on their own.

lib/mapObjects/CRewardableObject.cpp
```diff
@@ -826,12 +827,16 @@ void CGOnceVisitable::initObj(CRandomGenerator & rand)
...
+			info.resize(1);
 			if(rand.nextInt(99) < 20)
 			{
-				info.resize(1);
 				loadRandomArtifact(rand, info[0], 10, 10, 10, 0);
 				info[0].message.addTxt(MetaString::ADVOB_TXT, 37);
 			}
+			else
+			{
+				info[0].message.addTxt(MetaString::ADVOB_TXT, 38);
```
alexvins: What is the point of adding explicit empty reward?

Chocimier: It is meant to call [grantReward](https://github.com/Chocimier/vcmi/blob/iss1975/lib/mapObjects/CRewardableObject.cpp#L112) which marks object visited.

Chocimier:
> taking loot from once visitable objects marked them as visited for all players

It is what I removed, not introduced.

ArseniyShestakov:
So if there zero grants remain it's must be marked as visited team-wide.
If there still grants then marked as visited just for player who visited it.

That would be fine.

Chocimier:
Ok, you are right. Are there such multiple-reward objects with which I  
could test?

ArseniyShestakov:
Tested. This is exactly what we needed. Thanks!

ArseniyShestakov:
BTW if you going to contribute more what account do you want to use for bug tracker?
I suppose it's one with "Chocimier" in name?

lib/mapObjects/CRewardableObject.cpp
```diff
@@ -161,9 +161,15 @@ void CRewardableObject::onHeroVisit(const CGHeroInstance *h) const
 						grantRewardWithMessage(rewards[CRandomGenerator::getDefault().nextInt(rewards.size()-1)]);
 						break;
 				}
-				return;
+				break;
 			}
 		}
+
+		if (getAvailableRewards(h).size() == 0)
```
ArseniyShestakov: Uh... missed that this is going to cause crash for pickable objects.
Going to fix this soon.

ArseniyShestakov:
Hotfixed crash:
https://github.com/vcmi/vcmi/commit/bb05db90e53183bd67524ca54d5c317e1f5aa4fb

But if you have idea how to do that better feel free to suggest changes.

ArseniyShestakov:
Oh. Sadly I feel this one broken weekly revisitable objects too...

PR #215, opened 2016-09-16
Fix for compile error in VS 2015

dydzio0614:
Disabling Boost variadic templates to make VCMI compilable under Visual Studio 2015. Apparently this is MSVS compiler bug.

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -146,6 +146,10 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #define BOOST_THREAD_DONT_PROVIDE_THREAD_DESTRUCTOR_CALLS_TERMINATE_IF_JOINABLE 1
 #define BOOST_BIND_NO_PLACEHOLDERS
 
+#if defined(_MSC_VER) && (_MSC_VER == 1900)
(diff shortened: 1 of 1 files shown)


ArseniyShestakov:
Would be cool if you provide some gist there on what exactly failed to compile with variadit templates enabled. It's interesting since we wanted to use them to replace some macros.

dydzio0614:
In boost library header files in utility folder there is base_from_member.hpp header. At line 135 I get "_syntax error: identifier 'T'_".
Here is template that fails to compile: http://pastebin.com/DmU3aCEZ (error at line 13 in link provided)

ArseniyShestakov:
Thanks.

DjWarmonger:
Pull accepted. Do you have any more plans for it?

dydzio0614:
Plans for what?

DjWarmonger:
I asked Areseniy.

ArseniyShestakov:
For not this pull is okay since there nothing I needed from Boost. I just had some ideas how to replace exist macros with something better sometimes in future and now I'll have to be more careful about MSVC compatability.

PR #219, opened 2016-09-19
Update secondary skill handling. Fix issue 2307

dydzio0614:
Also making secondary skill icon display behavior same as in H3 during popup message.

### DIFF
--- a/lib/mapObjects/CGPandoraBox.cpp
+++ b/lib/mapObjects/CGPandoraBox.cpp
@@ -68,32 +68,51 @@ void CGPandoraBox::giveContentsUpToExp(const CGHeroInstance *h) const
 
 	if(gainedExp || changesPrimSkill || abilities.size())
 	{
+		std::vector<SecondarySkill> unpossessedAbilities;
+		std::vector<int> unpossessedAbilityLevels;
+		int abilitiesRequiringSlot = 0;
+
+		//filter out unnecessary secondary skills
+		for (int i = 0; i < abilities.size(); i++)
+		{
+			int curLev = h->getSecSkillLevel(abilities[i]);
+			bool abilityCanUseSlot = !curLev && ((h->secSkills.size() + abilitiesRequiringSlot) < GameConstants::SKILL_PER_HERO); //limit new abilities to number of slots
+
+			if (abilityCanUseSlot)
+				abilitiesRequiringSlot++;
+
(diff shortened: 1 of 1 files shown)


lib/mapObjects/CGPandoraBox.cpp
```diff
@@ -68,32 +68,51 @@ void CGPandoraBox::giveContentsUpToExp(const CGHeroInstance *h) const
 
 	if(gainedExp || changesPrimSkill || abilities.size())
 	{
+		std::vector<SecondarySkill> unpossessedAbilities;
```
DjWarmonger: I suggest std::pair<Secondary Skill, int>

lib/mapObjects/CGPandoraBox.cpp
```diff
@@ -68,32 +68,51 @@ void CGPandoraBox::giveContentsUpToExp(const CGHeroInstance *h) const
...
+		std::vector<int> unpossessedAbilityLevels;
+		int abilitiesRequiringSlot = 0;
+
+		//filter out unnecessary secondary skills
+		for (int i = 0; i < abilities.size(); i++)
```
DjWarmonger: use auto

lib/mapObjects/CGPandoraBox.cpp
```diff
@@ -68,32 +68,51 @@ void CGPandoraBox::giveContentsUpToExp(const CGHeroInstance *h) const
...
+			
 		for(int i=0; i<primskills.size(); i++)
 			if(primskills[i])
 				iw.components.push_back(Component(Component::PRIM_SKILL,i,primskills[i],0));
 
-		for(int i=0; i<abilities.size(); i++)
-			iw.components.push_back(Component(Component::SEC_SKILL,abilities[i],abilityLevels[i],0));
-
-		cb->showInfoDialog(&iw);
+		for(int i=0; i<unpossessedAbilities.size(); i++)
```
DjWarmonger: same - auto will give shorter code

ArseniyShestakov:
@DjWarmonger can you please clarify how exactly you expect to use auto in that case?

He actually use "i" to access multiple different vectors and this looks like reasonable approach to me if we don't want to change object serialization.

ArseniyShestakov:
Ok I got it. It's actually possible to use auto some of loops if he won't duplicate object properites structure there and just use std::pair. Agree on it.

DjWarmonger:
Thank you very much :)

PR #220, opened 2016-09-19
Fix memory problems with BonusList

vmarkovtsev:
Bonus * -> std::shared_ptr<Bonus>

This cures the following problems:

1) Memory corruption at exit. Some Bonus-es were deleted twice (mods?).
2) Memory leaks. Some Bonuses were not deleted.
3) Reduce the number of "Orphaned child" messages.

Valgrind reports 0 leaked memory and no invalid reads/writes now.

Warning: although I tested it, it should be tested by someone else.

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -128,10 +128,8 @@ int BonusList::totalValue() const
 	int indepMin = 0;
 	bool hasIndepMin = false;
 
-	for (auto & elem : bonuses)
+	for (auto& b : bonuses)
 	{
-		Bonus *b = elem;
-
 		switch(b->valType)
 		{
 		case Bonus::BASE_NUMBER:
@@ -179,7 +177,7 @@ int BonusList::totalValue() const
 	if(hasIndepMin && hasIndepMax)
 		assert(indepMin < indepMax);
 
-	const int notIndepBonuses = boost::count_if(bonuses, [](const Bonus *b)
+	const int notIndepBonuses = boost::count_if(bonuses, [](const std::shared_ptr<Bonus>& b)
 	{
 		return b->valType != Bonus::INDEPENDENT_MAX && b->valType != Bonus::INDEPENDENT_MIN;
 	});
@@ -201,24 +199,23 @@ int BonusList::totalValue() const
 
 	return valFirst;
 }
-const Bonus * BonusList::getFirst(const CSelector &selector) const
+
+std::shared_ptr<Bonus> BonusList::getFirst(const CSelector &select)
 {
-	for (auto & elem : bonuses)
+	for (auto & b : bonuses)
 	{
-		const Bonus *b = elem;
-		if(selector(b))
-			return &*b;
+		if(select(b.get()))
+			return b;
 	}
 	return nullptr;
 }
 
-Bonus * BonusList::getFirst(const CSelector &select)
+const std::shared_ptr<Bonus> BonusList::getFirst(const CSelector &selector) const
 {
-	for (auto & elem : bonuses)
+	for (auto & b : bonuses)
 	{
-		Bonus *b = elem;
-		if(select(b))
-			return &*b;
+		if(selector(b.get()))
+			return b;
 	}
 	return nullptr;
 }
@@ -230,19 +227,17 @@ void BonusList::getBonuses(BonusList & out, const CSelector &selector) const
 
 void BonusList::getBonuses(BonusList & out, const CSelector &selector, const CSelector &limit) const
 {
-	for (auto & elem : bonuses)
+	for (auto & b : bonuses)
 	{
-		Bonus *b = elem;
-
 		//add matching bonuses that matches limit predicate or have NO_LIMIT if no given predicate
-		if(selector(b) && ((!limit && b->effectRange == Bonus::NO_LIMIT) || ((bool)limit && limit(b))))
+		if(selector(b.get()) && ((!limit && b->effectRange == Bonus::NO_LIMIT) || ((bool)limit && limit(b.get()))))
 			out.push_back(b);
 	}
 }
 
 void BonusList::getAllBonuses(BonusList &out) const
 {
-	for(Bonus *b : bonuses)
+	for(auto & b : bonuses)
 		out.push_back(b);
 }
 
@@ -267,13 +262,13 @@ void BonusList::eliminateDuplicates()
 	bonuses.erase( unique( bonuses.begin(), bonuses.end() ), bonuses.end() );
 }
 
-void BonusList::push_back(Bonus* const &x)
+void BonusList::push_back(std::shared_ptr<Bonus> x)
 {
 	bonuses.push_back(x);
 	changed();
 }
 
-std::vector<Bonus*>::iterator BonusList::erase(const int position)
+BonusList::TInternalContainer::iterator BonusList::erase(const int position)
 {
 	changed();
 	return bonuses.erase(bonuses.begin() + position);
@@ -285,7 +280,7 @@ void BonusList::clear()
 	changed();
 }
 
-std::vector<BonusList*>::size_type BonusList::operator-=(Bonus* const &i)
+std::vector<BonusList*>::size_type BonusList::operator-=(std::shared_ptr<Bonus> const &i)
 {
 	auto itr = std::find(bonuses.begin(), bonuses.end(), i);
 	if(itr == bonuses.end())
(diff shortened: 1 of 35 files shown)


vmarkovtsev:
It looks compatible... Anyway, I loaded saves from the past and they worked.

20 сент. 2016 г. 2:02 пользователь "Alexander Shishkin" <
notifications@github.com> написал:

lib/NetPacksLib.cpp
```diff
@@ -106,10 +106,10 @@ DLL_LINKAGE void SetCommanderProperty::applyGs(CGameState *gs)
 	switch (which)
 	{
 		case BONUS:
-			commander->accumulateBonus (accumulatedBonus);
+			commander->accumulateBonus (std::make_shared<Bonus>(accumulatedBonus));
```
alexvins: You should change accumulatedBonus declaration instead. NetPacks changes are freely allowed (for now).

lib/HeroBonus.cpp
```diff
@@ -823,23 +815,22 @@ void CBonusSystemNode::addNewBonus(Bonus *b)
...
+void CBonusSystemNode::accumulateBonus(std::shared_ptr<Bonus> b)
 {
-	Bonus *bonus = exportedBonuses.getFirst(Selector::typeSubtype(b.type, b.subtype)); //only local bonuses are interesting //TODO: what about value type?
+	auto bonus = exportedBonuses.getFirst(Selector::typeSubtype(b->type, b->subtype)); //only local bonuses are interesting //TODO: what about value type?
 	if(bonus)
-		bonus->val += b.val;
+		bonus->val += b->val;
 	else
-		addNewBonus(new Bonus(b)); //duplicate needed, original may get destroyed
+		addNewBonus(b); //duplicate needed, original may get destroyed
```
alexvins: Duplicate is still needed here. It may work now because called with unique object.

lib/HeroBonus.cpp
```diff
@@ -809,7 +801,7 @@ void CBonusSystemNode::updateBonuses(const CSelector &s)
 		child->updateBonuses(s);
 }
 
-void CBonusSystemNode::addNewBonus(Bonus *b)
+void CBonusSystemNode::addNewBonus(std::shared_ptr<Bonus> b)
```
alexvins: This may even unique_ptr or at least unique() check with warning.

vmarkovtsev:
Fixed accumulatedBonus and duplicate. Uniqueness check fails every time so I think it will be the next step.
Besides, I commented the orphaned check message because shared_ptr guarantees that all the memory is freed in the end (and it really is) and this warning is not useful now.

ArseniyShestakov:
> Valgrind reports 0 leaked memory and no invalid reads/writes now.
This is unrelated to this branch, but we still have massive memory leak that visible with massif: [mantis#2349](http://bugs.vcmi.eu/view.php?id=2349). Even after many changes it's still there and memcheck don't provide useful info on it.

vmarkovtsev:
@ArseniyShestakov this is fun, will be able to drill it on the weekend, thx for pointing out.

lib/JsonNode.cpp
```diff
@@ -560,7 +560,7 @@ Key reverseMapFirst(const Val & val, const std::map<Key, Val> & map)
 	return "";
 }
 
-void JsonUtils::unparseBonus( JsonNode &node, const Bonus * bonus )
+void JsonUtils::unparseBonus( JsonNode &node, const std::shared_ptr<Bonus> bonus )
```
alexvins: Should be reference

lib/NetPacksLib.cpp
```diff
@@ -269,10 +269,10 @@ DLL_LINKAGE void GiveBonus::applyGs( CGameState *gs )
 
 	assert(cbsn);
 
-	if(Bonus::OneWeek(&bonus))
+	if(Bonus::OneWeek(std::make_shared<Bonus>(bonus)))
```
alexvins: Same as for accumulateBonus. bonus in GiveBonus should be shared_ptr.

lib/HeroBonus.cpp
```diff
@@ -809,8 +802,16 @@ void CBonusSystemNode::updateBonuses(const CSelector &s)
 		child->updateBonuses(s);
 }
 
-void CBonusSystemNode::addNewBonus(Bonus *b)
+void CBonusSystemNode::addNewBonus(const std::shared_ptr<Bonus>& b)
```
alexvins: Parameter should be just std::shared_ptr<Bonus>b (unlike accumulateBonus)

vmarkovtsev: If we do this, there will be a pointer's copy for sure and unique() will always return false. I am fine with it so I will remove the commented out code in the implementation then.

alexvins:
Serializer makes smart and plain pointers save-compatible, but AFAIK mixing smart and plain pointers for same objects can lead to problems. Changes should be made very carefully.

lib/HeroBonus.cpp
alexvins: Can we enforce moving here?

lib/NetPacks.h
```diff
@@ -465,7 +465,7 @@ struct SetCommanderProperty : public CPackForClient //120
 	ECommanderProperty which;
 	TExpType amount; //0 for dead, >0 for alive
 	si32 additionalInfo; //for secondary skills choice
-	Bonus accumulatedBonus;
+	std::shared_ptr<Bonus> accumulatedBonus;
```
alexvins: I`m afraid this is also wrong (this object is not constructed anywhere). Also I`m not sure that shared_ptr works correctly with netpacks.

lib/HeroBonus.cpp
vmarkovtsev: This blows up the brain :) Rvalues seem to open new cool bugs at this point.

lib/NetPacks.h
vmarkovtsev: It is initialized at CGameHandler::levelUpCommander line 258.

vmarkovtsev: OK I need to test the commander level up I guess.

lib/JsonNode.cpp
vmarkovtsev: Fixed

lib/NetPacks.h
alexvins: Sorry, missed that line 258 ...
But there are still problems with shared pointers in netpacks - 100% memory leak and valgring will not help there. Each shared_ptr is stored in loadedSharedPointers inside serializer (untill end of game in case of netpacks).

lib/HeroBonus.cpp
vmarkovtsev: We can do an explicit copy inside the method, what do you think?

vmarkovtsev: Okay, I tried making a copy and vcmiserver started to segfault in Template dctor at game end. Bad idea. Damn, it is so fragile!

lib/NetPacks.h
vmarkovtsev: Fixed. Reverted to Bonus.

lib/HeroBonus.cpp
alexvins: No it must be same Bonus objects, there are some places where Bonus is modified after addNewBonus call.
As for smart ptr itself I prefer addNewBonus declaration consistent with removeBonus, and do not care about unique check.

lib/HeroBonus.h
```diff
@@ -27,9 +27,9 @@ typedef std::set<CBonusSystemNode*> TNodes;
 typedef std::set<const CBonusSystemNode*> TCNodes;
 typedef std::vector<CBonusSystemNode *> TNodesVector;
 
-class CSelector : std::function<bool(const Bonus*)>
+class CSelector : std::function<bool(const std::shared_ptr<Bonus>)>
```
alexvins: shared_ptr here is pointless. Selectors has nothing to do with object ownership.

lib/HeroBonus.h
```diff
@@ -652,43 +654,43 @@ class DLL_LINKAGE CBonusSystemNode : public IBonusBearer, public boost::noncopya
...
 	void getRedDescendants(TNodes &out);
-	Bonus *getBonusLocalFirst(const CSelector &selector);
+	std::shared_ptr<Bonus> getBonusLocalFirst(const CSelector &selector);
 
 	void attachTo(CBonusSystemNode *parent);
 	void detachFrom(CBonusSystemNode *parent);
 	void detachFromAll();
-	virtual void addNewBonus(Bonus *b); //b will be deleted with destruction of node
-	void accumulateBonus(Bonus &b); //add value of bonus with same type/subtype or create new
+	virtual void addNewBonus(const std::shared_ptr<Bonus>& b); //b will be deleted with destruction of node
```
alexvins: Comment is not relevant any more.

vmarkovtsev:
@alexvins could you please add all your review suggestions at once or it will never end otherwise.

lib/HeroBonus.h
vmarkovtsev: Fixed. Reverted to pointers.

lib/HeroBonus.h
vmarkovtsev: Removed.

alexvins:
Sorry @vmarkovtsev but bonus system is vital part of vcmi so changes have to reviewed several times.

vmarkovtsev:
I beg someone please merge this, the second time I fix the conflicts

DjWarmonger:
No objections from me.

alexvins:
Ok i`ll merge this manually in several minutes

ArseniyShestakov:
Another thing I wanted to talk about is giving repository permissions to @vmarkovtsev. Of course I'll still prefer that any large changes or those require testing go through pull requests, but I pretty sure he can fix smaller things just fine just like we do.

A least he'll able to add himself to AUTHORS file since we really should ask all contributors to do that.

PR #221, opened 2016-09-20
[WIP] Experimental FoW refactoring

ArseniyShestakov:
FoW code we have now is bad and solution I see for this is to have sight map for each team. So tiles within sight radius of owned objects would never ever covered by darkness and we don't need to re-check owned objects every time unless object themself changed.

How this should be implemented is another story. First I thought to store all tiles with sight in vector with duplicates and just add / remove them. Then I find out that current fogOfWarMap can be also used for storage since it's unlikely that we'll ever get to situation where ui8 limit wouldn't be enough.

### DIFF
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -746,6 +746,7 @@ void CGameState::init(StartInfo * si, bool allowSavingRandomMap)
 	buildBonusSystemTree();
 	initVisitingAndGarrisonedHeroes();
 	initFogOfWar();
+	initSightMap();
 
 	// Explicitly initialize static variables
 	for(auto & elem : players)
@@ -1512,6 +1513,55 @@ void CGameState::giveCampaignBonusToHero(CGHeroInstance * hero)
 	}
 }
 
+void CGameState::removeSightnObj(const CGObjectInstance * obj)
+{
+	addSightObj(obj, false);
+}
+
+void CGameState::addSightObj(const CGObjectInstance * obj, bool add)
+{
+	if(!vstd::contains(players, obj->tempOwner))
+		return;
+
+	auto p = getPlayer(obj->tempOwner);
+	if(p->status != EPlayerStatus::INGAME)
+		return;
+
+	addSightObj(p->team, obj, add);
+}
+
+void CGameState::addSightObj(TeamID team, const CGObjectInstance * obj, bool add)
+{
+	auto ts = getTeam(team);
+	std::unordered_set<int3, ShashInt3> tiles;
+	getTilesInRange(tiles, obj->getSightCenter(), obj->getSightRadius());
+	for(int3 t : tiles)
+	{
+		if(add)
+		{
+			if(ts->fogOfWarMap[t.x][t.y][t.z] == 0)
+				ts->fogOfWarMap[t.x][t.y][t.z] = 2;
+			else
+				ts->fogOfWarMap[t.x][t.y][t.z]++;
+		}
+		else
+		{
+			ts->fogOfWarMap[t.x][t.y][t.z]--;
+		}
+	}
+}
+
+void CGameState::initSightMap()
+{
+	for(CGObjectInstance * obj : map->objects)
+	{
+		if(!obj)
+			continue; //not a flagged object
+
+		addSightObj(obj);
+	}
+}
+
 void CGameState::initFogOfWar()
 {
 	logGlobal->debug("\tFog of war"); //FIXME: should be initialized after all bonuses are set
@@ -1529,18 +1579,6 @@ void CGameState::initFogOfWar()
 			for(int h=0; h<map->height; ++h)
 				for(int v = 0; v < (map->twoLevel ? 2 : 1); ++v)
 					elem.second.fogOfWarMap[g][h][v] = 0;
-
-		for(CGObjectInstance *obj : map->objects)
-		{
-			if(!obj || !vstd::contains(elem.second.players, obj->tempOwner)) continue; //not a flagged object
-
-			std::unordered_set<int3, ShashInt3> tiles;
-			getTilesInRange(tiles, obj->getSightCenter(), obj->getSightRadius(), obj->tempOwner, 1);
-			for(int3 tile : tiles)
-			{
-				elem.second.fogOfWarMap[tile.x][tile.y][tile.z] = 1;
-			}
-		}
 	}
 }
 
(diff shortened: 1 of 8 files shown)


lib/CGameState.cpp
```diff
@@ -1608,6 +1610,64 @@ void CGameState::giveCampaignBonusToHero(CGHeroInstance * hero)
...
+	auto ts = getTeam(team);
+	std::unordered_set<int3, ShashInt3> tiles;
+	getTilesInRange(tiles, obj->getSightCenter(), obj->getSightRadius());
+	for(int3 t : tiles)
+	{
+		auto old = ts->fogOfWarMap[t.x][t.y][t.z];
+		if(add)
+		{
+			if(ts->fogOfWarMap[t.x][t.y][t.z] == 0)
+				ts->fogOfWarMap[t.x][t.y][t.z] = 2;
```
alexvins: Please explain why "0" become "2"

lib/CGameState.cpp
```diff
@@ -1608,6 +1610,64 @@ void CGameState::giveCampaignBonusToHero(CGHeroInstance * hero)
...
+			else
+				ts->fogOfWarMap[t.x][t.y][t.z]++;
```
alexvins: Each element is ui8, you should check overflow somehow.

lib/CGameState.cpp
ArseniyShestakov: 0 mean that tile in hidden from team.
1 mean that it's visible.
Anything more than 1 mean there is N objects with sight over that tile so it's can't be covered by darkness

lib/CGameState.cpp
ArseniyShestakov: Yep, but possibility that we'll have more than 255 objects that have sight over one tile is very unlikely. Isn't it?

alexvins: Unlikely but possible

lib/CGameState.cpp
alexvins: This should be documented. Also please add constants for 0, 1 & 2

lib/CGameState.cpp
ArseniyShestakov: If it's happen that would mean FoW going to break completely.
So I can just add assert so it's indicate if something going really wrong.

ArseniyShestakov:
So here is a bit cleaner version of code, but there still a lot of work to do:

* I'll rework map objects code related to sight range. At least where bonus system can be used.
* All situations where hero bonuses change must be handled properly too since both creatures and artifacts could increase sight radius using mods.

ArseniyShestakov:
And now sadly I stuck a bit. I need proper way to detect if hero (and in future probably other object) bonuses changed so I can update sight map on the fly as well as sent some notification about that on client, but so far it's looks like there no easy way to do that. Of course addding code to every single netpack and GH function is not an option.

In past I had similar problem with movement-related bonuses since they can be affected by artifacts too.

lib/IGameCallback.cpp
```diff
@@ -54,7 +54,7 @@ void CPrivilagedInfoCallback::getTilesInRange( std::unordered_set<int3, ShashInt
 		logGlobal->errorStream() << "Illegal call to getTilesInRange!";
 		return;
 	}
-	if (radious == -1) //reveal entire map
+	if (radious > 5000) //reveal entire map
```
alexvins: Should be constant.

ArseniyShestakov:
So little update on this branch and why it's not going into next release. My idea with FoW map looks good and give more flexibility for the future multiplayer modes (like simultaneous turns), but it's have one important requirement: on change of any FoW-affecting bonuses gamestate must be update immidiately.

Otherwise there might be problem like that:

* Hero stopped on tile, FoW map altered in default 5 tiles around.  
* Player give hero sight radius altering artefact or hero got scouting skill.
* Then FoW map won't be updated of course.
* Now if hero is killed before he moved FoW map altered for more than 5 tiles around and corrupted.

We discussed this with Alex little bit before and end up we need proper events for bonus system so code like one that manages FoW can be automatically notified if any bonuses changed so it's can update player FoW map.

So for the next update I'll just add more plumbing for exist FoW code.

alexvins:
Just ignore heroes when cashing FoW, apply hero sight radius OTF.

ArseniyShestakov:
OTF? 

I can't "ignore" anything at all since fog of war map will be the only storage on FoW information and every other check will be removed.

alexvins:
You should decide what is more difficult to implement: capture all bonus changes or do some OTF checks

ArseniyShestakov:
In this case I want proper code with no compomises. After all old FoW code work mostly so no reason to replace one crutch other.

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -325,14 +325,7 @@ void CGDwelling::serializeJsonOptions(JsonSerializeFormat & handler)
...
 {
-	if (subID == ETownType::TOWER)
-	{
-		if (hasBuilt(BuildingID::GRAIL)) //skyship
-			return -1; //entire map
-		if (hasBuilt(BuildingID::LOOKOUT_TOWER)) //lookout tower
-			return 20;
-	}
-	return 5;
+	return 5 + valOfBonuses(Bonus::SIGHT_RADIOUS);
```
alexvins: Add this "5" as permanent bonus

alexvins:
Set of tiles visible to hero is actually a volatile variable. When you try to cache it you will get problems.

ArseniyShestakov:
The problem is that I don't do any "caching" really. There no info who exactly see the tile in the map and all idea is just know that at least once object can "see" it so we never cover it by fog. Yeah it's possible to implement special hack to not deal with heroes or towns that way, but then there is no point in refactoring really since the idea is to simplify the FoW code.

alexvins:
No, it is still a cache: does not matter that only count of uncovering object is stored. As I understand the problem, without cache we need to check all objects on map to see it tile can be hidden, reducing check to only heroes and may be towns is huge improvement anyway.

ArseniyShestakov:
Ok I'll consider if it's possible to merge it with exception for heroes / towns.

ArseniyShestakov:
So it's doesn't look that my idea would work after all. Problem is that our bonus system only make it possible to track when _something_ inside bonus tree has changed. So when it's changed I wouldn't be able to tell what bonus nodes are affected and even if it's was possible to track just single node it's possible that hero could for instance have artifact that affect not just his sight, but will also propagate it's bonus to the whole team or every owned object of player, etc.

So efficiently with every bonus tree change I'll need to rebuild FoW map one way or another.

PR #222, opened 2016-09-27
Implementing hero battle window

dydzio0614:
Right click on hero during battle works like in original H3 now, though cursor doesn't change properly when hovering mouse over hero sprite.

### DIFF
--- a/lib/CGameInfoCallback.cpp
+++ b/lib/CGameInfoCallback.cpp
@@ -273,6 +273,16 @@ bool CGameInfoCallback::getHeroInfo(const CGObjectInstance * hero, InfoAboutHero
 
 	bool accessFlag = hasAccess(h->tempOwner);
 
+	if (!accessFlag && gs->curB) //if it's battle we can get enemy hero full data
+	{
+		ui8 playerSide = gs->curB->playerToSide(*player);
+		if (playerSide >= 0)
+		{
+			if (gs->curB->sides[!playerSide].hero == h)
+				accessFlag = true;
+		}
+	}
+
 	if(!accessFlag && nullptr != selectedObject)
 	{
 		const CGHeroInstance * selectedHero = dynamic_cast<const CGHeroInstance *>(selectedObject);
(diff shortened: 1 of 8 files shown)


ArseniyShestakov:
Btw I agree with @DjWarmonger that we don't want to leak maximal mana outside of battle.
As far as I understand on adventure map there is no information about that even with expert visions spell.

lib/CGameInfoCallback.cpp
```diff
@@ -273,6 +273,16 @@ bool CGameInfoCallback::getHeroInfo(const CGObjectInstance * hero, InfoAboutHero
 
 	bool accessFlag = hasAccess(h->tempOwner);
 
+	if (!accessFlag && gs->curB) //if it's battle we can get enemy hero full data
```
ArseniyShestakov: Fact that some battle is going is not enough to provide info like that.

What I think would even better is to move this check into CBattleInfoEssentials callback and name it like "CBattleInfoEssentials::playerHasAccessToHeroInfo".

Then in this case it's would be:
`if (!accessFlag && gs->curB)
    accessFlag = gs->curB->playerHasAccessToHeroInfo();`

lib/CGameInfoCallback.cpp
```diff
@@ -282,6 +292,9 @@ bool CGameInfoCallback::getHeroInfo(const CGObjectInstance * hero, InfoAboutHero
 
 	dest.initFromHero(h, accessFlag);
 
+	if (accessFlag && !gs->curB)
+		dest.details->manaLimit = -1; //we do not want to leak max mana info outside battle so set to meaningless value
```
ArseniyShestakov: This is of course not how something like that have to be implemented.
Instead you need extend InfoAboutHero's "detailed" bool and probably replace it with enum like:
EInfoLevel: BASIC, DETAILED, INBATTLE

This would let us extend info level if sometimes in future we'll have more things depend on it.

dydzio0614:
Changed as requested, though now there might be some code related to getting hero info that does not have to be called during battles. Disguise bonus check is an example.

PR #226, opened 2016-10-01
Add myself to AUTHORS

vmarkovtsev:
Suggested in #220.

Spelling Mac OS X as "macOS" because it was renamed in 2016: https://en.wikipedia.org/wiki/Mac_OS

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,7 +5,7 @@ Michał Urbańczyk aka Tow,        <impono@gmail.com>
 maintenance, reverse engineering, general support.
 
 Mateusz B. aka Tow dragon,                <matcio1@gmail.com>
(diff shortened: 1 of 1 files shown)


ArseniyShestakov:
@DjWarmonger @alexvins 
So guys can we finally give commit access to Vadim?
Posted about that in #220, but no one answered.

ArseniyShestakov:
@vmarkovtsev
If you have some time to do so you might join our still unofficial IRC: #vcmi at freenode.net

I wish we could have next release as stable as possible, but there few tricky crashes and not all of them we managed to fix with @alexvins. So that will be cool if you could look at them.

Also would be cool to add automated macOS builds with Travis so we can disciss that too.

PR #228, opened 2016-10-01
Fix launcher's UI for Ubuntu's scaled view on Retina displays

vmarkovtsev:
I had to write some dark layout magic if the text "Start the game" is longer than 60px of the button.


ArseniyShestakov:
I suppose this will only appear on High-DPI monitor, right?

vmarkovtsev:
Yes exactly. Hires fonts occupy much more width. The layout on other displays should remain the same.

vmarkovtsev:
Actually, I'd like this to be reviewed by @IvanSavenko - it is his code.

PR #229, opened 2016-10-02
Fix freshly recruited hero phantom in castle entrance

vmarkovtsev:
Currently if you hire a hero and go out from the castle, there will be a phantom hero in the entrance which copies the state and movement of the real one. This is because of the double printing. heroInGarrisonChange() prints the new hero for the second time, so we should check if heroInGarrisonChange() is called and not print in applyCl() then.


ArseniyShestakov:
I'll check it in multiplayer / hotseat and merge if it's okay.

PR #231, opened 2016-10-02
Fix cmake warnings

vmarkovtsev:


ArseniyShestakov:
I have one request regarding CMake. If you still build on Ubuntu that would be really cool if you have an idea how to make option that will force cmake to use built-in or system fuzzylite. Currently on Ubuntu 16.04 it's use system fuzzylite and work just fine and built-in version fail to built with latest Clang anyway, but I suspect we might sometimes want to forcefully use FL from our git.

PR #234, opened 2016-10-08
Bug fix for berserk:

dreamer10:
Fixed the issue that the berserk affected stack won't attack the nearest live stack. (The distance should be calculated as the distance between affected stack's position and the hex position reachable for this stack)


ArseniyShestakov:
Tested. Looks like more valid behaviour that we have so merged.
Thanks for contribution!

PR #238, opened 2016-10-12
Fix segfault/assertion failure during battle

vmarkovtsev:
The cause of the assertion failure at BattleStackAttacked::applyGs "at"
was shooting and casting some spell in handleAttackBeforeCasting() and
eventually killing the whole stack. Fix: filter dead stacks in the end of
handleAttackBeforeCasting().

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -3892,7 +3892,7 @@ bool CGameHandler::makeBattleAction(BattleAction &ba)
 					BattleAttack bat;
 					prepareAttack(bat, stack, destinationStack, (i ? 0 : distance),  ba.additionalInfo); //no distance travelled on second attack
 					//prepareAttack(bat, stack, stackAtEnd, 0, ba.additionalInfo);
-					handleAttackBeforeCasting(bat); //only before first attack
+					handleAttackBeforeCasting(&bat); //only before first attack
 					sendAndApply(&bat);
 					handleAfterAttackCasting(bat);
 				}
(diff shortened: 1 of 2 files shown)


server/CGameHandler.cpp
```diff
@@ -5209,16 +5209,23 @@ void CGameHandler::attackCasting(const BattleAttack & bat, Bonus::BonusType atta
...
 {
-	const CStack * attacker = gs->curB->battleGetStackByID(bat.stackAttacking);
-	attackCasting(bat, Bonus::SPELL_BEFORE_ATTACK, attacker); //no death stare / acid breath needed?
+	const CStack * attacker = gs->curB->battleGetStackByID(bat->stackAttacking);
+	attackCasting(*bat, Bonus::SPELL_BEFORE_ATTACK, attacker); //no death stare / acid breath needed?
+	// filter possibly dead stacks
+	bat->bsa.erase(std::remove_if(bat->bsa.begin(), bat->bsa.end(),
+	               [this](const BattleStackAttacked &bsa)
+	               {
+	                 return gs->curB->getStack(bsa.stackAttacked) == nullptr;
```
alexvins: Use const accessor battleGetStackByID in gamehandler. getStack should be used only in applyGs

server/CGameHandler.h
```diff
@@ -279,7 +279,7 @@ class CGameHandler : public IGameCallback, CBattleInfoCallback
 
 	void run(bool resume);
 	void newTurn();
-	void handleAttackBeforeCasting (const BattleAttack & bat);
+	void handleAttackBeforeCasting(BattleAttack *bat);
```
alexvins: Why not just remove const here?

vmarkovtsev: Habit from https://google.github.io/styleguide/cppguide.html#Reference_Arguments

server/CGameHandler.cpp
vmarkovtsev: Fixed

PR #241, opened 2016-10-16
Fix issue #2068 - maps list update

vmarkovtsev:
http://bugs.vcmi.eu/view.php?id=2068#c6839

I have no idea why the maps list was updated on Windows - I see no way it could. Added explicit rescan with filtering. It allows updating ZIP archive contents on the fly (not implemented).

### DIFF
--- a/lib/filesystem/CFilesystemLoader.h
+++ b/lib/filesystem/CFilesystemLoader.h
@@ -38,6 +38,7 @@ public:
 	std::string getMountPoint() const override;
 	bool createResource(std::string filename, bool update = false) override;
 	boost::optional<boost::filesystem::path> getResourceName(const ResourceID & resourceName) const override;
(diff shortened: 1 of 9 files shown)


lib/filesystem/CFilesystemLoader.h
```diff
@@ -50,7 +51,7 @@ class DLL_LINKAGE CFilesystemLoader : public ISimpleResourceLoader
 	 * key = ResourceID for resource loader
 	 * value = name that can be used to access file
 	*/
-	std::unordered_map<ResourceID, boost::filesystem::path> fileList;
+	mutable std::unordered_map<ResourceID, boost::filesystem::path> fileList;
```
vmarkovtsev: Adding "mutable" here because fileList is merely the cache and not the real instance state (over which we do not have the control - it is volatile and belongs to OS).

PR #242, opened 2016-10-17
Implement adventure map scroll cursor change

dydzio0614:
Also, cursor changes back to default when stop of scrolling detected. Fixes issue #2399.

### DIFF
--- a/client/windows/CAdvmapInterface.cpp
+++ b/client/windows/CAdvmapInterface.cpp
@@ -59,6 +59,31 @@ using namespace CSDL_Ext;
 
 CAdvMapInt *adventureInt;
 
+static void setScrollingCursor(ui8 direction)
+{
+	if(direction & CAdvMapInt::RIGHT)
+	{
+		if(direction & CAdvMapInt::UP)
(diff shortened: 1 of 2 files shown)


client/windows/CAdvmapInterface.cpp
```diff
@@ -59,6 +59,31 @@ using namespace CSDL_Ext;
 
 CAdvMapInt *adventureInt;
 
+static void SetScrollingCursor(ui8 direction)
```
dydzio0614: Maybe this procedure could be optimized

client/windows/CAdvmapInterface.cpp
```diff
@@ -956,11 +981,18 @@ void CAdvMapInt::show(SDL_Surface * to)
 
 		if(scrollingDir)
 		{
+			SetScrollingCursor(scrollingDir);
+			ScrollingState = true;
```
ArseniyShestakov: I think all codebase have more or less consistent style for variables and function names. So please rebase it without capital letters in the beginning.

PR #245, opened 2016-10-29
Add commander resurrection info window upon entering town

vmarkovtsev:
Do you know the correct info window text? I dumped allTexts to no avail.

### DIFF
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -562,15 +562,25 @@ void CGTownInstance::onHeroVisit(const CGHeroInstance * h) const
 	}
 	else if(h->visitablePos() == visitablePos())
 	{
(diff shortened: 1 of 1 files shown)


alexvins:
this a WoG feature, so the text should be somewhere in ZMESS00.TXT and we heed to handle situation when wog mod is not present.

alexvins:
One solution would be put text to config\translate.json or combine with wog text - use translate.json as a fall-back.

vmarkovtsev:
The correct messages appear in ZNPC00.txt. I am unable to play original WoG at the moment, but it seems to me that commanders shall be recovered after paying money. Do we have those mechanics defined in some document? @ArseniyShestakov you may know this...

ArseniyShestakov:
Sorry, but in this case I can't do much except to test it later since I don't play with WoG usually.

vmarkovtsev:
Merging this because the problem appeared to be much deeper than I thought. We have to implement commander both hiring and reviving.

PR #246, opened 2016-10-30
Fix large obstacles drawing in front of the stacks

vmarkovtsev:
Some big background obstacles have placement position lower than their top hex, thus they are drawn in front of the stacks. Since it is hard to fix every obstacle which is incorrectly configured in DEF files, we can offload them to beforeAll after initial sorting by hex.
Test: [save](https://mega.nz/#!dF5RWRZL!5A0zCeKD8ekHXZl3A08asl7a0cu2XbpFVDk04QBp5o8). Attack the hero on the left and look at the devil commander. Before fix: overlapped, no head. After fix: everything is normal.

### DIFF
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -3234,7 +3234,7 @@ void CBattleInterface::showBattlefieldObjects(SDL_Surface *to)
 
 	// dead stacks should be blit first
 	showStacks(to, objects.beforeAll.dead);
-	for ( auto & data : objects.hex )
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
Shit I didn't want to merge this. Can you please test that rendering battles is OK?

ArseniyShestakov:
I tested 10-15 battles include some sieges and at least it's seems you didn't break anything.

ArseniyShestakov:
BTW example of what was fixed:

p246_after
p246_before

PR #248, opened 2016-10-31
issue 2584

alexvins:


ArseniyShestakov:
Now resurrection spell (of hero) can't be casted on alive stack.
Is it intended if hero only have basic resurrection?

ArseniyShestakov:
And yet archangels still can't cast resurrection on dead stacks.

alexvins:
1) should hero cast resurection if he cant heal one full creature?
2) archangels can cast if press 'F', but now they can also move to that hex.

ArseniyShestakov:
1) As far as I see from my tests - no. If you can't resurrect one creature then you can't cast it no matter if it's alive stack or corpse. I sadly not sure if alive stack health make sense or no.
2) I completely forgot about "F" change. So it's does work properly, but this is really confusing change that probably should be included in changelog.

alexvins:
'F' hotkey is another ticket http://bugs.vcmi.eu/view.php?id=1286

ArseniyShestakov:
So as far as I see it's works better now.

Though resurrection itself need more testing and work since behaviour of basic spell is really inconsistent. E.g if you revive corpse of stack with archangels, then revive it with basic resurrection and don't touch it until end of battle amount of creatures in the end going to be not what you expect really.

alexvins:
> E.g if you revive corpse of stack with archangels, then revive it with basic resurrection and don't touch it until end of battle amount of creatures in the end going to be not what you expect really.

Not reproducible

ArseniyShestakov:
Should we look to merge this one?

alexvins:
I think yes

ArseniyShestakov:
Ok it's weird, but with this branch even 29 archangels can't resurrect stack of one.

alexvins:
moved to #323

PR #250, opened 2016-11-02
Cumulative spell effects

alexvins:


DjWarmonger:
For accumulated values there's CBonusSystemNode::accumulateBonus() function.

The one thing it doesn't allow is to dispell (or wear off) partial effects independently - but it's not the case of Distrupting Ray.

alexvins:
accumulateBonus has different selector for presence check, and it may not work for spells with effect on f.e. primary skill

client/battle/CBattleInterface.cpp
```diff
@@ -1359,7 +1359,7 @@ void CBattleInterface::spellCast(const BattleSpellCast *sc)
 
 void CBattleInterface::battleStacksEffectsSet(const SetStackEffect & sse)
 {
-	if (sse.effect.back().sid == -1 && sse.stacks.size() == 1 && sse.effect.size() == 2)
+	if (sse.stacks.size() == 1 && sse.effect.size() == 2 && sse.effect.back().sid == -1)
```
ArseniyShestakov: Formatting.

lib/spells/CDefaultSpellMechanics.cpp
```diff
@@ -445,118 +445,128 @@ void DefaultSpellMechanics::battleLogDefault(std::vector<MetaString> & logLines,
...
+		sse.stacks.push_back(affected->ID);
 
-			//Apply hero specials - peculiar enchants
-			const ui8 tier = std::max((ui8)1, affected->getCreature()->level); //don't divide by 0 for certain creatures (commanders, war machines)
-			if(bonus)
+		//Apply hero specials - peculiar enchants
+		const ui8 tier = std::max((ui8)1, affected->getCreature()->level); //don't divide by 0 for certain creatures (commanders, war machines)
+		if(bonus)
+		{
+			switch(bonus->additionalInfo)
```
ArseniyShestakov: This switch also had formatting unlike all other in spells code.
Might be it's time to fix it if you change it anyway.

server/CGameHandler.cpp
```diff
@@ -4507,12 +4507,14 @@ void CGameHandler::stackAppearTrigger(const CStack *st)
...
-		st->stackEffectToFeature(sse.effect, pseudoBonus);
-		if (sse.effect.size())
+		const CSpell * sp = SpellID(b->subtype).toSpell();
+		const int level = ((val > 3) ?  (val - 3) : val);
+
+		sp->getEffects(sse.effect, level, false, 50);
+		//this makes effect accumulate for at most 50 turns by default, but effect may be permanent and last till the end of battle
+		sp->getEffects(sse.cumulativeEffects, level, true, 50);
+
+		if (!sse.effect.empty() || !sse.cumulativeEffects.empty())
```
ArseniyShestakov: And here.

ArseniyShestakov:
This is weird, but misiokles has reported it's doesn't work now and It's looks like that's the case.
I tried to debug it and I see stackEnchantedTrigger is called, but effect doesn't stack as expected.

ArseniyShestakov:
Ok I'm about to abuse your branch to test my deploy code.
If I break something I have backup of code.

Also added note to changelog and updated wiki.

PR #253, opened 2016-11-03
[WIP] Multiplayer: graceful player loss and disconnection

ArseniyShestakov:
With following changes client won't ask server to close unless he's host and server will gracefully remove player from the game if he's disconnected.

More things I'm possible going to do:

- [ ] Make proper netpack for gs change and rename "enteredLosingCheatCode" to something else.
- [ ] Option to pass control of disconnected player to AI instead to removal.
- [ ] Let host if he lost keep client open without active PlayerInterface. Should be similar with AI control hack.
- [ ] Pass ability to host to another client if host disconnected. AI state would be lost, but no reason why this shouldn't be possible.


DjWarmonger:
Probably the best option to have would be leave player state as it is without control and let client reconnect, gaining control back.

ArseniyShestakov:
No doubt option to reconnect would be better. For simple connection loss (ISP IP change, etc) we can just give N minutes to client to reconnect. Though for actual crashed client GS transfer can be quite complex task.

ArseniyShestakov:
No doubt option to reconnect would be better. For simple connection loss (ISP IP change, etc) we can just give N minutes to client to reconnect. Though for actual crashed client GS transfer can be quite complex task.

sigboe:
#Suggestion:

 ##Situation 1:
User disconnects by using menu options. Settings on the server side will either flag this as a loss condition, or change over to AI.

 ##Situation 2:
User disconnects for unknown reason. Game continues as normal, when it is the disconnected users turn, an option for the host/other players will be available for AI to take over his account or give the user a loss condition.

ArseniyShestakov:
Yeah I can make client sent "I left the game" message to the server when disconnected gracefully.

Other option would require some kind of voting so I not sure how this should be handled. Anyway not going to work on it until I make something for host player loss.

sigboe:
@ArseniyShestakov Maybe intercept the dialog that shows its the other players turn. When its a client that left without saying goodbye players can vote wait (will be able to change vote later), or vote for the leaver to get a loss condition or be taken over by AI.

ArseniyShestakov:
Voting sounds overengineering for now when anyone who going to attempt play VCMI multiplayer clearly communicating over other channels. In future if we're going to have some kind of lobby for multiplayer we can back to it.

So for now I'll attempt to make dialog for the host.

sigboe:
@ArseniyShestakov Great that was my first suggestion anyway 😃

ArseniyShestakov:
So I'll revive this branch shortly with minimal amount of changes. I'm not going to implement any advanced disconnection handling this time because it's end up not easy to pass whole game state through network. Of course I can implement savefile transfer, but that would require a lot of work on loading code so it's only take useful parts.

So for now my goal is to just merge small QoL changes since they require for my other branch.

ArseniyShestakov:
Ok I started to refactoring it and current name and description make no sense now.
I'll just create new PR once it's ready.

PR #254, opened 2016-11-05
[WIP] Map handler refactoring

alexvins:
Map object animation converted for use of CAnimation. This probably already enables 32bit graphics for adventure map objects, but I have nothing to test with.

FoW also converted.

Terrain is different task - it requires rotations.


ArseniyShestakov:
I suppose you might now want to drop CDefHandler includes from AdventureMapClasses.cpp and CAdvMapInterface.cpp

alexvins:
@ArseniyShestakov done

PR #255, opened 2016-11-05
Travis: add Windows build via MXE

ArseniyShestakov:
Still have to find out why return code is wrong, but I think it's works fine.

In future I want to to publish artifacts from Travis too so daily builds not depend on my own PC.

### DIFF
--- a/.travis.mxe
+++ b/.travis.mxe
@@ -1,0 +1,27 @@
+#!/bin/sh
+
+# Add MXE repository and key
+echo "deb http://pkg.mxe.cc/repos/apt/debian wheezy main" \
+    | sudo tee /etc/apt/sources.list.d/mxeapt.list
+
+sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys D43A795B73B16ABE9643FE1AFD8FFF16DB45C6AB
+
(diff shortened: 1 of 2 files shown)


viciious:
It'd be nice to have Appveyod MSVC builds too

ArseniyShestakov:
@viciious yeah I would absolutely love to have CI for MSVC so we don't break compatibility, but sadly we have no one using VS together with CMake as generator and I don't really want to host project files in repository. Also I have no idea where to look for prebuilt libraries for Windows and not really want to trying to make it work myself.

ArseniyShestakov:
Yeah for some reason vcmitest fail to build with MXE and I not sure if I should disable it for that build or no.

viciious:
@ArseniyShestakov yeah, pre-built libraries for msvc is definitely something worth having as a downloadable bundle. Pulling it all together is an exercise in patience you don't want to repeat too often. Maybe I can do it some time later.

ArseniyShestakov:
@viciious just in case if you ever decide to participate I dropped you invite to our slack.

I suppose dydzio might have all libraries we need, but his VS setup is kind a tricky.

PR #256, opened 2016-11-06
Refactor CArtPlace to create CCommanderArtPlace + extend commander artifact support

dydzio0614:
Second commit probably makes CCommanderArtPlace more ugly and responsible for more things than it should, but it allows to give commander artifacts back to hero via left click on artifact.

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -41,13 +41,13 @@ void CHeroArtPlace::createImage()
 {
 	OBJ_CONSTRUCTION_CAPTURING_ALL;
 
-	int graphic = 0;
+	int imageIndex = 0;
 	if (ourArt)
-		graphic = ourArt->artType->iconIndex;
+		imageIndex = ourArt->artType->iconIndex;
 	if (locked)
-		graphic = ArtifactID::ART_LOCK;
+		imageIndex = ArtifactID::ART_LOCK;
 
-	image = new CAnimImage("artifact", graphic);
+	image = new CAnimImage("artifact", imageIndex);
 	if (!ourArt)
 		image->disable();
 
@@ -969,8 +969,7 @@ CCommanderArtPlace::CCommanderArtPlace(Point position, const CGHeroInstance * co
 void CCommanderArtPlace::clickLeft(tribool down, bool previousState)
 {
 	if (down && ourArt && text.size())
-		LOCPLINT->showYesNoDialog("Do you want to give this artifact back to hero?", [this] { returnArtToHeroCallback(); }, [] {});
-		//CArtPlace::clickLeft(down, previousState);	
+		LOCPLINT->showYesNoDialog(CGI->generaltexth->localizedTexts["commanderWindow"]["artifactMessage"].String(), [this] { returnArtToHeroCallback(); }, [] {});
 }
 
 void CCommanderArtPlace::clickRight(tribool down, bool previousState)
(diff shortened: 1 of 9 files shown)


client/widgets/CArtifactHolder.cpp
```diff
@@ -945,3 +943,102 @@ bool CArtifactsOfHero::SCommonPart::Artpos::valid()
...
+	ArtifactPosition freeSlot = ourArt->firstBackpackSlot(commanderOwner);
+
+	ArtifactLocation src(commanderOwner->commander.get(), artifactPos);
+	ArtifactLocation dst(commanderOwner, freeSlot);
+
+	if (ourArt->canBePutAt(dst, true))
+	{
+		LOCPLINT->cb->swapArtifacts(src, dst);
+		setArtifact(nullptr);
+		parent->redraw();
```
dydzio0614: This line is probably a code smell, but I did not find better solution with current design.

client/widgets/CArtifactHolder.cpp
```diff
@@ -945,3 +943,102 @@ bool CArtifactsOfHero::SCommonPart::Artpos::valid()
...
+	}
+}
+
+void CCommanderArtPlace::setArtifact(const CArtifactInstance * art)
```
dydzio0614: This method is similar to CHeroArtPlace::setArtifact - code duplication?

client/widgets/CArtifactHolder.h
```diff
@@ -41,13 +41,44 @@ class CWindowWithArtifacts : public CArtifactHolder
...
+	void clickRight(tribool down, bool previousState) override;
+
+	virtual void setArtifact(const CArtifactInstance *art)=0;
+};
+
+class CCommanderArtPlace : public CArtPlace
+{
+protected:
+	const CGHeroInstance * commanderOwner;
+	ArtifactPosition commanderSlotID;
```
dydzio0614: I decided to separate this from CHeroArtPlace::slotID, as the latter is managed magically by CArtifactsOfHero, and I did not want to set it up in CArtPlace constructor (which is now a base class)

client/widgets/CArtifactHolder.cpp
```diff
@@ -945,3 +943,102 @@ bool CArtifactsOfHero::SCommonPart::Artpos::valid()
...
+{
+	LRClickableAreaWTextComp::clickLeft(down, previousState);
+}
+
+void CArtPlace::clickRight(tribool down, bool previousState)
+{
+	LRClickableAreaWTextComp::clickRight(down, previousState);
+}
+
+CCommanderArtPlace::CCommanderArtPlace(Point position, const CGHeroInstance * commanderOwner, ArtifactPosition artSlot, const CArtifactInstance * Art) : CArtPlace(position, Art), commanderOwner(commanderOwner), commanderSlotID(artSlot.num)
```
dydzio0614: Yeah, this class constructor takes may arguments, but I guess it is ok, considering it does everything without relying on CArtifactsOfHero magic

client/widgets/CArtifactHolder.cpp
```diff
@@ -945,3 +943,102 @@ bool CArtifactsOfHero::SCommonPart::Artpos::valid()
...
+{
+	createImage();
+	setArtifact(Art);
+}
+
+void CCommanderArtPlace::clickLeft(tribool down, bool previousState)
+{
+	if (down && ourArt && text.size())
+		LOCPLINT->showYesNoDialog("Do you want to give this artifact back to hero?", [this] { returnArtToHeroCallback(); }, [] {});
```
alexvins: Missing localization. Use config/traslate.json

PR #262, opened 2016-12-10
Further fix for mantis #2599

dydzio0614:
Remove visibility check for hero info during battle, causes crash when enemy attacks from monolith.


ArseniyShestakov:
InfoAboutHero struct simply doesn't contain information about artifacts.

PR #264, opened 2016-12-21
Added few files and folders to .gitignore which should not be part of…

NikolayIT:
… the repository when compiling with Visual Studio 2015

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,17 @@ doc/*
 VCMI_VS11.sdf
 *.ipch
 VCMI_VS11.opensdf
+
+# Visual Studio
+*.suo
+*.user
+/AI/*/RD
+/client/RD
+/launcher/RD
+/lib/RD
+/lib/minizip/RD
+/scripting/erm/RD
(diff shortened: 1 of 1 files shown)


ArseniyShestakov:
Might be add all *.opendb and *.lib files?

.gitignore
```diff
@@ -36,3 +36,17 @@ doc/*
 VCMI_VS11.sdf
 *.ipch
 VCMI_VS11.opensdf
+
+# Visual Studio
+*.suo
+*.user
+/AI/*/RD
```
alexvins: RD is a release build with debug info? OK, but where are other build types?

NikolayIT: When compiling in Debug configuration, no new files are created.

DjWarmonger:
Of course these files shouldn't be a part of repository, but I don't see they ever were in first place. But as you wish.

NikolayIT:
@DjWarmonger they wasn't in the repository. They are created by Visual Studio 2015 when I compile the source.

PR #268, opened 2017-01-21
Small update for VS2015

krs0:
You will need 3 VC global Macros like:

BOOSTDIR=c:\Home\Krs\Work\VCMI\boost
BOOSTLIBDIR=$(BOOSTDIR)\lib32-msvc-14.0
QTCORE=c:\Home\Krs\Work\VCMI\Qt\5.7\msvc2015


krs0:
The latest addition was not really intended to land here. (But it did because of a Git mistake I made). I will play a little more with the project configurations and signal when I'm ready.

krs0:
Ok These should be the latest changes to VS 2015 configurations.

ArseniyShestakov:
Nobody who using VS want to check it so closing it.
Also unlikely we'll officially support VS solution in future.

New guide for easy build with CMake-generated solution soon will be there:
https://wiki.vcmi.eu/index.php?title=How_to_build_VCMI_(Windows/Vcpkg)

PR #269, opened 2017-01-26
Add few creature abilities similar to WoG missing ones, fix bug with summoned shooter stack init

dydzio0614:
Implemented abilities:
WoG ghost raising count from kills
Werewolf transform on attack
Santa gremlin summon guardians at battle start

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -5537,8 +5537,105 @@ void CGameHandler::runBattle()
 	}
 
 	//initial stacks appearance triggers, e.g. built-in bonus spells
-	for (auto stack : gs->curB->stacks)
+	auto initialStacks = gs->curB->stacks;
+
+	for (CStack * stack : initialStacks)
 	{
+		if (stack->hasBonusOfType(Bonus::SUMMON_GUARDIANS))
+		{
+			const std::shared_ptr<Bonus> summonInfo = stack->getBonus(Selector::type(Bonus::SUMMON_GUARDIANS));
+			auto accessibility = getAccesibility();
+			CreatureID creatureData = CreatureID(summonInfo->subtype);
+			std::vector<BattleHex> targetHexes;
+			bool targetIsBig = stack->getCreature()->isDoubleWide(); //target = creature to guard
+
+			if (!creatureData.toCreature()->isDoubleWide())
+				targetHexes = stack->getSurroundingHexes();
+			else
+			{
+				if(stack->attackerOwned) //handle front guardians, TODO: should we handle situation when units start battle near opposite side of the battlefield? Cannot happen in normal H3...
+					BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::RIGHT, false).moveInDir(BattleHex::EDir::RIGHT, false), targetHexes);
+				else
+					BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::LEFT, false).moveInDir(BattleHex::EDir::LEFT, false), targetHexes);
+
+				//guardian spawn locations for four default position cases for attacker and defender, non-default starting location for att and def is handled in first two if's
+				if (stack->attackerOwned && ( (stack->position.getY() % 2 == 0) || (stack->position.getX() > 1) ))
+				{
+					if (targetIsBig && (stack->position.getY() % 2 == 1) && (stack->position.getX() == 2)) //handle exceptional case
+					{
+						BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::TOP_RIGHT, false), targetHexes);
+						BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::BOTTOM_RIGHT, false), targetHexes);
+					}
+					else
+					{	//add back-side guardians for two-hex target, side guardians for one-hex
+						BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(targetIsBig ? BattleHex::EDir::TOP_LEFT : BattleHex::EDir::TOP_RIGHT, false), targetHexes);
+						BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(targetIsBig ? BattleHex::EDir::BOTTOM_LEFT : BattleHex::EDir::BOTTOM_RIGHT, false), targetHexes);
+
+						if (!targetIsBig && stack->position.getX() > 2) //back guard for one-hex
+							BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::LEFT, false), targetHexes);
+						else if (targetIsBig)//front-side guardians for two-hex target
+						{
+							BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::RIGHT, false).moveInDir(BattleHex::EDir::TOP_RIGHT, false), targetHexes);
+							BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::RIGHT, false).moveInDir(BattleHex::EDir::BOTTOM_RIGHT, false), targetHexes);
+							if (stack->position.getX() > 3) //back guard for two-hex
+								BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::LEFT, false).moveInDir(BattleHex::EDir::LEFT, false), targetHexes);
+						}
+					}
+					
+				}
+
+				else if (!stack->attackerOwned && ((stack->position.getY() % 2 == 1) || (stack->position.getX() < GameConstants::BFIELD_WIDTH - 2)))
+				{
+					if (targetIsBig && (stack->position.getY() % 2 == 0) && (stack->position.getX() == GameConstants::BFIELD_WIDTH - 3 )) //handle exceptional case... equivalent for above for defender side
+					{
+						BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::TOP_LEFT, false), targetHexes);
+						BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(BattleHex::EDir::BOTTOM_LEFT, false), targetHexes);
+					}
+					else
+					{
+						BattleHex::checkAndPush(BattleHex(stack->position).moveInDir(targetIsBig ? BattleHex::EDir::TOP_RIGHT : BattleHex::EDir::TOP_LEFT, false), targetHexes);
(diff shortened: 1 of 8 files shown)


dydzio0614:
Lost summoned units are added to killed units, just like elementals from summon spell. Is that wrong? About shortening the code, I am not sure how to do it further. I merged two "default starting position hex calculation" cases with initializing in the middle of the battlefield after moving unit in tactics phase to first two "if's", to share some code.

BattleHex(stack->position) calls are used because moveInDir function changes *this rather than returning new value so I would have to reset variable after every use.

server/CGameHandler.cpp
```diff
@@ -5225,6 +5318,44 @@ void CGameHandler::handleAfterAttackCasting(const BattleAttack & bat)
...
+		else if (attacker->hasBonusOfType((Bonus::TRANSMUTATION), 1))
+			resurrectInfo.amount = defender->count;
+		else
+			return; //wrong subtype
+
+		BattleStacksRemoved victimInfo;
+		victimInfo.stackIDs.insert(bat.bsa.at(0).stackAttacked);
+
+		sendAndApply(&victimInfo);
+		sendAndApply(&resurrectInfo);
```
alexvins: Should tansmuted stack inherit effects (f.e. active spells, bonuses from stack artifacts)?

server/CGameHandler.cpp
```diff
@@ -5476,8 +5607,38 @@ void CGameHandler::runBattle()
 	}
 
 	//initial stacks appearance triggers, e.g. built-in bonus spells
-	for (auto stack : gs->curB->stacks)
+	auto initialStacks = gs->curB->stacks;
```
alexvins: Should be comment here about required copy of stacks. IMO it is not obvious.

server/CGameHandler.cpp
```diff
@@ -5476,8 +5607,38 @@ void CGameHandler::runBattle()
...
 	{
+		if (stack->hasBonusOfType(Bonus::SUMMON_GUARDIANS))
+		{
+			const std::shared_ptr<Bonus> summonInfo = stack->getBonus(Selector::type(Bonus::SUMMON_GUARDIANS));
+			auto accessibility = getAccesibility();
+			CreatureID creatureData = CreatureID(summonInfo->subtype);
+			std::vector<BattleHex> targetHexes;
+			bool targetIsBig = stack->getCreature()->isDoubleWide(); //target = creature to guard
+
+			if (!creatureData.toCreature()->isDoubleWide())
```
alexvins: Should be documented here how guardian positions look for big creature.

server/CGameHandler.cpp
```diff
@@ -6222,6 +6383,12 @@ CasualtiesAfterBattle::CasualtiesAfterBattle(const CArmedInstance * _army, Battl
 				StackLocation sl(army, st->slot);
 				newStackCounts.push_back(TStackAndItsNewCount(sl, st->count));
 			}
+			else if (st->count > army->getStackCount(st->slot) && st->hasBonusOfType(Bonus::SOUL_STEAL, 0)) //allow WoG ghost amount grow from battles
+			{
+				logGlobal->debug("Stack gained %d units.", army->getStackCount(st->slot) - st->count);
```
alexvins: Should be `st->count - army->getStackCount(st->slot)`

server/CGameHandler.cpp
```diff
@@ -6222,6 +6383,12 @@ CasualtiesAfterBattle::CasualtiesAfterBattle(const CArmedInstance * _army, Battl
 				StackLocation sl(army, st->slot);
 				newStackCounts.push_back(TStackAndItsNewCount(sl, st->count));
 			}
+			else if (st->count > army->getStackCount(st->slot) && st->hasBonusOfType(Bonus::SOUL_STEAL, 0)) //allow WoG ghost amount grow from battles
```
alexvins: No need to check for SOUL_STEAL here. Let stack to grow with any reason.

DjWarmonger:
@dydzio0614 , Someone already handled that ;):

	BattleHex movedInDir(EDir dir, bool hasToBeValid = true) const
	{
		BattleHex result(*this);
		result.moveInDir(dir, hasToBeValid);
		return result;
	}

The same operation could be used for
stack->position.getY(); getX()

server/CGameHandler.cpp
dydzio0614: I do not know how is that handled for WoG werewolves, though I would keep things as they are. Being 100% WoG compliant would require some hardcoding and this ability is good foothold anyway.

server/CGameHandler.cpp
```diff
@@ -137,6 +137,76 @@ static void giveExp(BattleResult &r)
 	}
 }
 
+static void SummonGuardiansHelper(std::vector<BattleHex> & output, BattleHex targetPosition, bool targetIsAttacker, bool targetIsTwoHex) //return hexes for summoning two hex monsters in output, target = unit to guard
```
dydzio0614: Maybe targetPosition should not be passed by value?

alexvins: It should be `const`. Then you can use method poitned by Warmonger.

server/CGameHandler.cpp
```diff
@@ -137,6 +137,79 @@ static void giveExp(BattleResult &r)
...
 
+static void SummonGuardiansHelper(std::vector<BattleHex> & output, const BattleHex & targetPosition, bool targetIsAttacker, bool targetIsTwoHex) //return hexes for summoning two hex monsters in output, target = unit to guard
+{
+	int x = targetPosition.getX();
+	int y = targetPosition.getY();
+
+	if (targetIsAttacker) //handle front guardians, TODO: should we handle situation when units start battle near opposite side of the battlefield? Cannot happen in normal H3...
+		BattleHex::checkAndPush(targetPosition.movedInDir(BattleHex::EDir::RIGHT, false).movedInDir(BattleHex::EDir::RIGHT, false), output);
+	else
+		BattleHex::checkAndPush(BattleHex(targetPosition).movedInDir(BattleHex::EDir::LEFT, false).movedInDir(BattleHex::EDir::LEFT, false), output);
```
alexvins: This line still have excess BattleHex copying.

server/CGameHandler.cpp
```diff
@@ -6222,6 +6391,12 @@ CasualtiesAfterBattle::CasualtiesAfterBattle(const CArmedInstance * _army, Battl
 				StackLocation sl(army, st->slot);
 				newStackCounts.push_back(TStackAndItsNewCount(sl, st->count));
 			}
+			else if (st->count > army->getStackCount(st->slot) && !st->hasBonusOfType(Bonus::SOUL_STEAL, 1))
```
alexvins: What does `!st->hasBonusOfType(Bonus::SOUL_STEAL, 1)` actually check?

dydzio0614: From bonus description: BONUS_NAME(SOUL_STEAL) /*val - number of units gained per enemy killed, subtype = 0 - gained units survive after battle, 1 - they do not*/ \

It is additional option for the bonus to allow stack grow only for the battle time.

alexvins: Sorry, missed that comment.
Then you should use "resurrected" field of CStack like low level resurrection do (more over you designed SOUL_STEAL as "overheal"). No checks here.

dydzio0614: Ok, I will amend last commit when done

server/CGameHandler.cpp
```diff
@@ -993,7 +993,7 @@ void CGameHandler::applyBattleEffects(BattleAttack &bat, const CStack *att, cons
 		StacksHealedOrResurrected::HealInfo hi;
 		hi.stackID = att->ID;
 		hi.healedHP = bsa.killedAmount * att->valOfBonuses(Bonus::SOUL_STEAL) * att->MaxHealth(); //TODO: Should unit be additionally healed after life drain?
-		hi.lowLevelResurrection = false;
+		hi.lowLevelResurrection = att->hasBonusOfType(Bonus::SOUL_STEAL, 1);
```
alexvins: What if creature both subtype 0 and subtype 1 bonuses?

dydzio0614: You mean default subtype or? I do not know how creature can have two subtypes at once

alexvins: Any entity may have several bonuses of same type.

alexvins: F.e. In this case creature may have 25% soul steal permanently + 25% till the end of battle.

dydzio0614: I think its acceptable way of working, in other case it would not make sense to implement two same bonuses on single unit unless somebody tries to break something badly

DjWarmonger:
You check same condition multiple times, which is convoluted. How about:
```

		if (att->hasBonusOfType(Bonus::SOUL_STEAL, 0))
 	{
 		StacksHealedOrResurrected::HealInfo hi0;
 		hi0.stackID = att->ID;
 		hi0.healedHP = bsa.killedAmount * att->valOfBonuses(Bonus::SOUL_STEAL, 0) * att->MaxHealth();
 		shi.healedStacks.push_back(hi0);
 		hi0.lowLevelResurrection = false;
}
		if (att->hasBonusOfType(Bonus::SOUL_STEAL, 1))
 {
 		StacksHealedOrResurrected::HealInfo hi1;
 		hi1.stackID = att->ID;
 		hi1.healedHP = bsa.killedAmount * att->valOfBonuses(Bonus::SOUL_STEAL, 1) * att->MaxHealth();
 		hi1.lowLevelResurrection = true;

 		shi.healedStacks.push_back(hi1);
}
 		if (hi0.healedHP > 0 || hi1.healedHP > 0)
 			bsa.healedStacks.push_back(shi);
```
Last condition needs separate handling, but code is much shorter still. The above example could be still be rolled into a loop or lambda. Duplicated code is never good.

server/CGameHandler.cpp
```diff
@@ -906,7 +979,34 @@ void CGameHandler::applyBattleEffects(BattleAttack &bat, const CStack *att, cons
...
+			{
+				StacksHealedOrResurrected::HealInfo hi;
+				hi.stackID = att->ID;
+				hi.healedHP = bsa.killedAmount * att->valOfBonuses(Bonus::SOUL_STEAL, i) * att->MaxHealth();
+				hi.lowLevelResurrection = (bool)i;
+				shi.healedStacks.push_back(hi);
+			}
+		}
+		if (std::any_of(shi.healedStacks.begin(), shi.healedStacks.end(), [](StacksHealedOrResurrected::HealInfo healInfo) { return healInfo.healedHP > 0; }))
+		{
```
DjWarmonger: For future reference: http://www.boost.org/doc/libs/1_54_0/libs/algorithm/doc/html/the_boost_algorithm_library/CXX11/any_of.html

PR #270, opened 2017-01-29
 Making unit tests for BattleHex and some estetic improvements.

FeniksFire:
Making unit tests for BattleHex.
Delete unnecessary range, templates.

### DIFF
--- a/test/CVcmiTestConfig.cpp
+++ b/test/CVcmiTestConfig.cpp
@@ -33,10 +33,15 @@ CVcmiTestConfig::CVcmiTestConfig()
 	loadDLLClasses();
 	logGlobal->info("Initialized global test setup.");
 
-// These lines make a "Test setup error:"
-//	const std::string TEST_DATA_DIR = "test/";
-//	auto loader = new CFilesystemLoader("test/", TEST_DATA_DIR);
-//	dynamic_cast<CFilesystemList*>(CResourceHandler::get())->addLoader(loader, false);
+	/* TEST_DATA_DIR may be wrong, if yes below test don't run,
+	find your test data folder in your build and change TEST_DATA_DIR for it*/
+	const std::string TEST_DATA_DIR = "test/";
+	auto path = boost::filesystem::current_path();
+	path+= "/" + TEST_DATA_DIR;
+	if(boost::filesystem::exists(path)){
+		auto loader = new CFilesystemLoader("test/", TEST_DATA_DIR);
+		dynamic_cast<CFilesystemList*>(CResourceHandler::get())->addLoader(loader, false);
+	}
 }
 
 CVcmiTestConfig::~CVcmiTestConfig()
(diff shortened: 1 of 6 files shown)


test/CVcmiTestConfig.cpp
```diff
@@ -33,10 +33,10 @@ CVcmiTestConfig::CVcmiTestConfig()
 	loadDLLClasses();
 	logGlobal->info("Initialized global test setup.");
 
-	const std::string TEST_DATA_DIR = "test/";
-
-	auto loader = new CFilesystemLoader("test/", TEST_DATA_DIR);
-	dynamic_cast<CFilesystemList*>(CResourceHandler::get())->addLoader(loader, false);
+// These lines make a "Test setup error:"
+//	const std::string TEST_DATA_DIR = "test/";
```
alexvins: May be this path is incorrect in your build.

FeniksFire: If so, more people have that same problem, and this is wrong so should be comment out, I can or someone else can uncommend it when look in closer for it  and fix it ;)

alexvins: Where is tests executable in your build?

alexvins: Try const std::string TEST_DATA_DIR = "testdata/";

FeniksFire: Not working. Here you have somes screens:
http://imgur.com/a/KmAqp

DjWarmonger:
@FeniksFire Oh I didn't know that. If they are supported by standard now, we can definitelly drop boost here.

test/CVcmiTestConfig.cpp
alexvins: Try `const std::string TEST_DATA_DIR = ".";`

In your build "terrainViewMappings.json" & "TerrainViewTest.h3m" are in same directory with vcmitest.

FeniksFire: Huh, It's works. Thanks for tips. These differences seem to stem from differences in project file, but I'm not sure. I will therefore have to do something about it.

alexvins: Sorry, I`ve done this on purpose about a year ago. The idea was to make tests work built in-place, I`ll do smth more smart later.

alexvins: Its better to enclose `auto loader = new CFilesystemLoader("test/", TEST_DATA_DIR);` with ifdef for now.

FeniksFire: well ok, I enclose it and I give short info about these lines, about the path :)

FeniksFire: I do it without ifdef although a similar effect, simple check if directory exist.

FeniksFire:
there is anything I should correct in this pull? why don't merge it?

DjWarmonger:
@alexvins , it's up to you.

PR #272, opened 2017-02-01
Extend creature siege support. Fix bug #2600

dydzio0614:
Make ballistics skill apply only to catapult, add new skill to handle multi attack on siege walls for creatures, fix cyclop kings

### DIFF
--- a/config/bonuses_texts.json
+++ b/config/bonuses_texts.json
@@ -17,6 +17,12 @@
 		"name": "Additional retaliations",
 		"description": "May Retaliate ${val} extra times"
 	},
+	
+	"ADVANCED_CATAPULT":
+	{
+		"name": "Advanced wall breaker",
+		"description": "Can hit siege walls ${val} extra times per attack"
+	},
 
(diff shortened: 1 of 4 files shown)


DjWarmonger:
It would be simpler just to add "val" parameter to CATAPULT which contains the number of extra shots.

dydzio0614:
I thought about that, though cyclop kings take their "catapult" bonus from crtraits.txt so I thought this may be problematic to give them bonus value. AVS also said on slack: "separate bonus design is consistent with shooting ans spellcasting", so I decided to go that way.

DjWarmonger:
Maybe at least give it clear name, such as "CATAPULT_SHOTS" or something.

Alucard648:
What about "EXPERT_CATAPULT" bonus to work like Expert Ballistics hero secondary skill?

DjWarmonger:
It shoudl be more general. We might want to add more catapult shots in the future. Also more shots for Ballista - for instance Ring of Machine Affinity from ToE does that.

PR #275, opened 2017-02-04
Implement ranged counterattack feature

dydzio0614:


DjWarmonger:
Looks okay, but how about some creatures to test all these new abilities?

dydzio0614:
I changed archers and marksmen on my local VCMI copy for testing purpose and everything seems to work alright. There are no creatures in h3 / wog / hota that use that feature, though it got requested by one modder.

PR #277, opened 2017-02-09
Restart button hacky fix

dydzio0614:
On some machines restart scenario button closes VCMI. This happens probably because server is not fully closed when code assumes it is (game not completely ended when starting it again). Added 1 second delay on restart, which should cause the button to work correctly most of the time.

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -1179,7 +1179,8 @@ static void handleEvent(SDL_Event & ev)
 			{
 				StartInfo si = *client->getStartInfo(true);
 				endGame();
-				startGame(&si);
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
There actually used to be a delay like 300ms. I think I removed it and wanted to debug this shit because adding delays on exit is a very dirty hack. I didn't have time to fully trace it but the real problem is surely with thread dependency hell and races. I highly encourage you to properly debug the problem instead of hacking delays.

dydzio0614:
I managed to get proper solution that just works, using one boost::mutex. Before pulling it I will try to learn if something else than mutex can be better / more intuitive for code readers in this case as boost offers many interesting stuff for concurrent programming.

dydzio0614:
Proper fix: https://github.com/vcmi/vcmi/pull/288

PR #278, opened 2017-02-09
Small bug fix.

FeniksFire:
During the unit tests I saw that hex can take from occupiedHex() signed values and then he get stupid value. So changing from ui16 to si16.

### DIFF
--- a/lib/CBattleCallback.cpp
+++ b/lib/CBattleCallback.cpp
@@ -1479,11 +1479,11 @@ AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes (const CStack
 	RETURN_IF_NOT_BATTLE(at);
 
 	const int WN = GameConstants::BFIELD_WIDTH;
-	si16 hex = (attackerPos != BattleHex::INVALID) ? attackerPos.hex : attacker->position.hex; //real or hypothetical (cursor) position
+	BattleHex hex = (attackerPos != BattleHex::INVALID) ? attackerPos.hex : attacker->position.hex; //real or hypothetical (cursor) position
 
 	//FIXME: dragons or cerbers can rotate before attack, making their base hex different (#1124)
 	bool reverse = isToReverse (hex, destinationTile, isAttacker, attacker->doubleWide(), isAttacker);
-	if (reverse)
+	if (reverse && attacker->doubleWide())
 	{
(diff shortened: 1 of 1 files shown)


DjWarmonger:
Well, then implementation is incorrect. Battlefield hex should never take negative values, as in the picture: http://bugs.vcmi.eu/file_download.php?file_id=797&type=bug&show_inline=1&file_show_inline_token=20170210ef80ea5043de4e2037b0e2a371348ae656f07220

dydzio0614:
Picture link does not work

DjWarmonger:
Maybe now: http://bugs.vcmi.eu/view.php?id=792

FeniksFire:
@DjWarmonger BattleHex can at this moment take negative values. Like INVALID(-1). See the contributors and check yourself for this mistake. At this moment this make for me BIG CRASH. If nobody change it in the future, and you don't merge it, error will be still here(and then what?). No my fault that implementation is bad, if in the future someone try repair it I he should replace all fragments like "si16 hex" to "ui16 hex".

DjWarmonger:
Ah I see. Then better use special typedef (THex?) to ensure correct handling all over the code.

FeniksFire:
good idea, I try do it ;)

lib/CBattleCallback.cpp
```diff
@@ -1479,11 +1479,11 @@ AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes (const CStack
 	RETURN_IF_NOT_BATTLE(at);
 
 	const int WN = GameConstants::BFIELD_WIDTH;
-	ui16 hex = (attackerPos != BattleHex::INVALID) ? attackerPos.hex : attacker->position.hex; //real or hypothetical (cursor) position
+	BattleHex hex = (attackerPos != BattleHex::INVALID) ? attackerPos.hex : attacker->position.hex; //real or hypothetical (cursor) position
 
 	//FIXME: dragons or cerbers can rotate before attack, making their base hex different (#1124)
 	bool reverse = isToReverse (hex, destinationTile, isAttacker, attacker->doubleWide(), isAttacker);
-	if (reverse)
+	if (reverse && attacker->doubleWide())
```
FeniksFire: Well, I think a little more time about it. It can be even ui16 type if in this place we will have the condition for doubleWide(). But for make sure, without typedef I change ui16 for BattleHex. I think it should be okay now.

PR #279, opened 2017-02-09
Counterattack correction. Fixes bug #2501

dydzio0614:
Creatures should not be able to counterattack multiple times during one opponent attack under any circumstances.


Alucard648:
Speaking for multi-counterattack: What about to implement a creature bonus type that allows monster to execute a counterattack for each attack done by the same opponent in the same turn (I.E. a Crusader attacks twice and high-leveled Royal Griffin retaliates twice) as long, as creature has counterattacks in the current round.

PR #281, opened 2017-02-17
Fix compilation with clang 4

janisozaur:
See https://www.securecoding.cert.org/confluence/display/cplusplus/EXP58-CPP.+Pass+an+object+of+the+correct+type+to+va_start
for details


janisozaur:
You could fix your travis configuration, it's been failing for days.

PR #282, opened 2017-02-17
Avoid scroollbar in resolution combobox

drizt:


drizt:
Without patch:

vcmi launcher_790

With patch:

vcmi launcher_789

PR #284, opened 2017-02-18
Make AI able to buy extra buildings. Fix bug #2640

dydzio0614:


AI/VCAI/VCAI.cpp
```diff
@@ -1404,6 +1404,16 @@ void VCAI::buildStructure(const CGTownInstance * t)
 		return;
 	if (tryBuildAnyStructure(t, std::vector<BuildingID>(extra, extra + ARRAY_COUNT(extra))))
 		return;
+
+	//at the end, try to get and build any extra dwelling upgrades (for example HotA 3rd level dwelling)
+	std::vector<BuildingID> extraDwellingUpgrades;
+
```
DjWarmonger: In the end I guess there could be more extra buildings in VCMI than just dwellings, so you may want to change name of this vector.

PR #285, opened 2017-02-20
Improve .spec file

drizt:
Now rpm can be build for Fedora and openSUSE.

### DIFF
--- a/rpm/vcmi.spec
+++ b/rpm/vcmi.spec
@@ -1,52 +1,76 @@
-Summary:			VCMI is an open-source project aiming to reimplement HoMM3 game engine, giving it new and extended possibilities.
-Name:				vcmi
-Version:			0.99
-Release:			1%{?dist}
-License:			GPLv2+
-Group:				Amusements/Games
-
-%define _unpackaged_files_terminate_build 0
-
-# The source for this package was pulled from upstream's vcs.  Use the
-# following commands to generate the tarball:
-#  wget https://github.com/vcmi/vcmi/archive/0.98.tar.gz
-#  tar -xzf 0.98.tar.gz vcmi-0.98-1
-Source:				vcmi-0.98-1.tar.gz
-
-URL:				http://forum.vcmi.eu/portal.php
-
-BuildRequires:		cmake
-BuildRequires:		gcc-c++ >= 4.7.2
-BuildRequires:		SDL2-devel
-BuildRequires:		SDL2_image-devel
-BuildRequires:		SDL2_ttf-devel
-BuildRequires:		SDL2_mixer-devel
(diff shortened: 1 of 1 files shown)


drizt:
Any prorgress here?

ArseniyShestakov:
To make it more clear why I just didn't merged it right away. We have a mess with packaging on Linux and even our Ubuntu PPA contain broken packages since Debian rules outdated as well. There also some magic in VCMI code itself to make it all work together somehow that was setup long ago and really need to be improved. That ```CMAKE_SKIP_RPATH``` I believe could be completely useless too.

Sadly I never really had time to seat and find if code we have now is doing right thing and how to improve it.

Another thing is that I really want all VCMI descriptions everywhere to be consistent: on website, on wiki, in packages, etc. So I really wish to update Debian rules as well.

ArseniyShestakov:
And one more note. I not sure whatever our main repository should actually include Debian / RPM configuration at all. Might be best investment of time would be to get Snap or Flatpak working properly.

drizt:
Ok. It's not a problem. Some note: vcmi RPM package can't be added to official Fedora repository due dependencies from ffmpeg package from RPMFussion repository.

ArseniyShestakov:
We only need ffmpeg for video playback that basically only needed for campaigns. Might be we could find better way to handle it in future since we can allow to convert video from Bink before playing and then use something other than ffmpeg for playback.

lemenkov:
@ArseniyShestakov this would be helpful indeed. Is it possible to convert original Bink-videos to some well-known format (VP8, VP9) and use these instead?

ArseniyShestakov:
@lemenkov main problem is that we can only provide utility or instruction for converting, but not files themself. So again Fedora users going to be on their own since I doubt anything other than ffmpeg support bink.

hroncok:
BTW I'm working on putting vcmi into rpmfusion. https://bugzilla.rpmfusion.org/show_bug.cgi?id=4815

I doubt it could be in Fedora proper (even without ffmpeg) as per Packages_which_are_not_useful_without_external_bits

ArseniyShestakov:
@hroncok I'll look into adding compatibility for newer Boost.

hroncok:
Thanks.

hroncok:
@ArseniyShestakov You might want to see this [Absolutely disgusting and untested patch for compatibility with 1.66](https://bugzilla.rpmfusion.org/show_bug.cgi?id=4815#c5).

nullkiller:
I expect this thing is outdated. If somebody plan to fix it fill free to reopen this PR or better create a new one. Btw ffmpeg dependency is optional now.

PR #288, opened 2017-03-16
Restart button fix

dydzio0614:
Added bool variable that prevents game from starting if server is running. That guarantees server to close in time.

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -90,6 +90,7 @@ std::queue<SDL_Event> events;
 boost::mutex eventsM;
 
 bool gNoGUI = false;
+bool serverNotDead = false;
 static po::variables_map vm;
 
 //static bool setResolution = false; //set by event handling thread after resolution is adjusted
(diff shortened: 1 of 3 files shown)


client/CMT.cpp
```diff
@@ -90,6 +90,7 @@ std::queue<SDL_Event> events;
 boost::mutex eventsM;
 
 bool gNoGUI = false;
+bool serverNotDead = false;
```
alexvins: Ensure synchronized access to this variable. (smth from std::atomic f.e.)

client/CMT.cpp
```diff
@@ -1248,6 +1249,10 @@ static void mainLoop()
 
 void startGame(StartInfo * options, CConnection *serv/* = nullptr*/)
 {
+	while (serverNotDead == true)
```
alexvins: write just `while (serverNotDead)`

client/CMT.cpp
alexvins: May be serverAlive is better name.

dydzio0614: What is the reason though?

mwu-tow:
Check the `CondSh<bool>`  — it serves this exact idiom (flag with synchronized access that can be waited upon). With that you can get rid of those ugly sleep calls.

dydzio0614:
Will do

PR #291, opened 2017-03-23
Upgraded to fuzzylite 6.0

jcrada:
The new version of fuzzylite 6.0 provides important performance improvements, bug fixes, new features, and a new license: GNU General Public License 3.

Commit 250b585 deletes the bundled version of fuzzylite. Compiles and builds fine assuming that the fuzzylite libraries are installed in the system.

Commit 9b7b7dc adds the fuzzylite library (without examples) to the `AI/FuzzyLite` directory. Compiles and builds fine when defining `-DFORCE_BUNDLED_FL=ON`.


alexvins:
Should FuzziLite be now a submodule as we have no local changes now?

jcrada:
I think having fuzzylite as a submodule might be a better idea. I can create the submodule in the next 48 hours.

mwu-tow:
Good idea, using git submodule should be preferable over directly including the third-party library sources code in the repo.

alexvins:
Commits are huge, please remove excess  9b7b7dc and 0fc4ed6

ArseniyShestakov:
Thanks! I'll try to make sure this doesn't break AI ASAP.
Then we can merge it.

ArseniyShestakov:
@DjWarmonger can you please look into this branch?
I really wish to merge it since there no reason to keep FL inside our repository.

dydzio0614:
When checking this i would keep in mind some potential "bugs" - this looks suspicious: https://github.com/vcmi/vcmi/blob/develop/AI/VCAI/Fuzzy.cpp#L383

jcrada:
If you can export the engine to the FuzzyLite Language, I can take a look at that using QtFuzzyLite. Just send me the result of: `engine->toString();`

ArseniyShestakov:
@DjWarmonger ok I merged it all once again on latest develop: 
https://github.com/vcmi/vcmi/tree/jcrada-develop
If you can verify that everything is okay that will be very helpful.

@jcrada thanks, but you're not to worry about that. Сomment that Dydzio pointed to is outdated as we find out in Slack. Hopefully in future we'll have automated test to verify our AI is behave properly with new version.

ArseniyShestakov:
Ok just in case @dydzio0614 also checked that AI behavior seems fine with branch merged.
So I think we better to merge it and find out if something actually breaks.

Though I might still need to look how it's compile let's say on Ubuntu 14.04 and see if we need any compatibility code.

ArseniyShestakov:
Ok I merged it here: 82d107b934e8dd6a5235f4f7e2c1f2e1611805f1

Mainly because I want to merge my Vcpkg branch soon and I want to test it with newer FuzzyLite already.

PR #292, opened 2017-03-23
Fix highlight in city.

FeniksFire:
Fixed refreshing highlight in the city. I'm not sure of this solution, but I don't see any disadvantages of it. For now I think it's safe.


ArseniyShestakov:
Sorry, but can you please attach screenshot or at least explain what's exactly this one fixing?

FeniksFire:
Sure, but I think that short movie is better in that case.
So here you have:
Before:
https://www.dropbox.com/s/kdvtxh6td67x2pq/Before.avi?dl=0
After:
https://www.dropbox.com/s/si1hk6gv8zpxe6z/After.avi?dl=0

ArseniyShestakov:
This is even better, thanks. I wasn't aware such issue exist.

I'm not entirely sure if this right way to fix it, but since it's improve behaviour I think it's fine to merge it.

PR #293, opened 2017-03-26
Fix bug #2476

dydzio0614:
Finally fixed this annoying bug. Tested on Windows 7, I do not see any graphical difference after changes.


dydzio0614:
After investigating the issue further, I think this is not a good fix. Supposedly better one will be created in separate PR

PR #294, opened 2017-03-27
Fix bug #2476 (2nd solution)

dydzio0614:
This commit allows lower position bounds to raise mouseMoved events for interface objects (there aren't many objects which use that). This fixes bug #2476. It even improves battle hex behavior when mouse is between two of them.


alexvins:
What about other uses of `isItIn`?

dydzio0614:
I found only 2 other objects than CAdvMapInt "listening" to it during adventure map (+ battlehexes during battle) - game seems to work fine.

PR #299, opened 2017-05-17
Android support

Fayth:
Contains VCMI codebase changes that allow compilation of the android app (using launcher project https://github.com/Fayth/vcmi-android to compile).

Lets hope it doesn't break the dekstop version. ;)


server/CVCMIServer.cpp
```diff
@@ -615,9 +625,24 @@ int main(int argc, char** argv)
...
 	CResourceHandler::clear();
 
   return 0;
 }
+
+#ifdef VCMI_ANDROID
+
+void CVCMIServer::create()
+{
+	boost::this_thread::sleep(boost::posix_time::milliseconds(1000));
```
alexvins: What is the point of this 1 second timeout?

lib/VCMIDirs.cpp
```diff
@@ -535,23 +545,50 @@ std::string VCMIDirsXDG::libraryName(const std::string& basename) const { return
...
 
-// on Android HOME will be set to something like /sdcard/data/Android/is.xyz.vcmi/files/
-bfs::path VCMIDirsAndroid::userDataPath() const { return getenv("HOME"); }
+bfs::path VCMIDirsAndroid::libraryPath() const { return nativePath; }
+bfs::path VCMIDirsAndroid::userDataPath() const { return basePath; }
 bfs::path VCMIDirsAndroid::userCachePath() const { return userDataPath() / "cache"; }
 bfs::path VCMIDirsAndroid::userConfigPath() const { return userDataPath() / "config"; }
 
+boost::filesystem::path VCMIDirsAndroid::fullLibraryPath(const std::string &desiredFolder,
+														 const std::string &baseLibName) const {
```
alexvins: Put `{` on new line, do not indent arguments too far.

server/CVCMIServer.cpp
ArseniyShestakov: This I can answer. There is no boost::interprocess on Android so this is used instead. In fact we could actually benefit from removing interprocess everywhere as well since it's the only thing we using it for. I think we discussed that.

ArseniyShestakov: What we really need for both desktop and Android is probably just add N attempts to connect to server within N seconds. So if server port not appear to be opened in that time then we consider it's failed to start.

client/windows/CAdvmapInterface.cpp
```diff
@@ -470,6 +530,9 @@ CAdvMapInt::CAdvMapInt():
   spellBeingCasted(nullptr), position(int3(0, 0, 0)), selection(nullptr),
   updateScreen(false), anim(0), animValHitCount(0), heroAnim(0), heroAnimValHitCount(0),
 	activeMapPanel(nullptr), duringAITurn(false), scrollingDir(0), scrollingState(false)
+#ifdef VCMI_ANDROID
+	, swipeEnabled(settings["general"]["swipe"].Bool()), swipeMovementRequested(false), swipeTargetPosition(int3(-1,-1,-1))
```
Fayth: Okay, good point. Actually it is added in the config, but I didn't apply it to the repo (in android project, config folder is embedded inside the app).

client/windows/CAdvmapInterface.h
```diff
@@ -56,6 +56,16 @@ class CTerrainRect
...
 	EMapAnimRedrawStatus lastRedrawStatus;
 	CFadeAnimation * fadeAnim;
+
+	int3 swipeInitialMapPos;
+	int3 swipeInitialRealPos;
+	bool isSwiping;
+	static constexpr float SwipeTouchSlop = 16.0f;
+
+	void handleHover(const SDL_MouseMotionEvent &sEvent);
+#ifdef VCMI_ANDROID
```
Fayth: I thought about doing it for desktop later. Preferably using right or middle mouse button.

server/CVCMIServer.cpp
Fayth: I'll need to check if that sleep is actually needed. It looks like it might be a leftover (because client doesn't really do anything worthwhile at this time -- after requesting server start, it just sleeps waiting for server response), but I can't check at the moment.

Fayth: About the whole design of interprocess communication then yeah, I can't use boost's interprocess. Currently client and server communicate through android Messanger (in java), which basically sends messages between processes. So basically client requests that the server should be started and when server finishes initializing, it sends the message that it's ready.

lib/AndroidVMHelper.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+static JavaVM *vmCache = nullptr;
+
+/// cached java classloader so that we can find our classes from other threads
+static jobject vcmiClassLoader;
+static jmethodID vcmiFindClassMethod;
+
+void AndroidVMHelper::cacheVM(JNIEnv *env)
+{
+    env->GetJavaVM(&vmCache);
```
Fayth: I think I'll fiiiinally have to configure my IDE to correctly format code for vcmi.

server/CVCMIServer.cpp
ArseniyShestakov: Do we really need use any of it at all? Interprocess or Messager both.
Is there reason why we can't simply retry connection until it's successful?

Fayth:
(I'll try to fix the issues during the weekend)

server/CVCMIServer.cpp
Fayth: I'm not sure how can we start the server other than sending the message to it... There might be a way, but currently I can't think of any solution.

ArseniyShestakov: I have no ideas how Android work at all and on desktop we can just start server process without any additional communication channel with it since we'll connect to it over network anyway. So interprocess only add complexity and cause issues on Windows.

ArseniyShestakov:
You should be now able to transfer vcmi-android repository too.

Global.h
```diff
@@ -698,3 +698,16 @@ namespace vstd
 }
 using vstd::operator-=;
 using vstd::make_unique;
+
+#if defined VCMI_ANDROID && defined NO_STD_TOSTRING
```
Fayth: Currently I define it in cmake (unconditionally). Android uses gnustl by default and it looks like there's no to_string (it is possible to use libc++, but it's not fully supported).
If there's a better way to detect this, I'm open to suggestions.

Fayth: Also, I added VCMI_ANDROID guard there (and most of other places), because, for now, I tried to change desktop's version logic as little as possible.

Fayth:
One more thing about code style -- is there any guideline about CRLF vs LF line endings? It seems it's not consistent now.

lib/CGameInterface.cpp
```diff
@@ -22,256 +37,245 @@
...
+#if THIS_COULD_BE_USED_WHEN_ANDROID_FIXES_RTTI_SUPPORT
+    logGlobal->debugStream() << "Opening AI from " << libpath.string();
+    void *dll = dlopen(libpath.string().c_str(), RTLD_LOCAL | RTLD_LAZY);
+    if (dll)
+    {
+        getName = (TGetNameFun) dlsym(dll, "GetAiName");
+        getAI = (TGetAIFun) dlsym(dll, methodName.c_str());
+    }
+    else
+        logGlobal->errorStream() << "Error: " << dlerror();
```
Fayth: I generally like to have braces everywhere, but coding guideline on wiki states that we should "Avoid using unnecessary open/close braces".

lib/VCMIDirs.cpp
Fayth: How do you suggest to align the arguments in this case? According to wiki "(...) please align your arguments to match the opening brace". Should I use normal continuation indent instead?

alexvins:
Line endings should be LF, but nobody actually cares about it.

Global.h
Fayth: I don't exactly know how that works. Generally there are several runtimes that can be used to build native code (see: https://developer.android.com/ndk/guides/cpp-support.html#hr ). The last time I tried to experiment with these, the full libc++ didn't work correctly (compiled, but there were errors during runtime). I don't know why gnustl (which is, according to the page I linked "This runtime is the GNU Standard C++ Library, (libstdc++-v3)") doesn't have that method -- it seems it is possible to circumvent the missing methods in some way (example here http://stackoverflow.com/questions/17950814/how-to-use-stdstoul-and-stdstoull-in-android/18124627#18124627 ), but I didn't try that.

lib/CGameInterface.cpp
alexvins: But there is `}` on next line.

Global.h
alexvins: Oh I`ve forgotten about -std=gnu++11. I use that from the very beginning for MinGW.

lib/CGameInterface.cpp
Fayth: Whoops, true.
I think I should remove that whole "fixed rtti for android" block. I don't think it can be used in near future and it would require to change much more in the build process anyway.

ArseniyShestakov:
Not that I don't care about line endings, but git solve that for us: it's just locally convert them to proper version. And I doubt it's worth it to mess with code changes history just to get LF everywhere.

Global.h
Fayth: It looks like std::to_string works when I add both -std=gnu++11 and _GLIBCXX_USE_C99 (which is internal define, so that might be a bad idea...)

server/CVCMIServer.cpp
Fayth: I removed that thread sleep line.

I think that researching&redesigning client-server communication model should be a long-term task. Current implementation isn't broken and feels rather stable.

Global.h
alexvins: OK. Let this this hack-fix be. But if we get more problems with STL on Android NDK, I will insists on proper solution.

P.S. some more explanation:

> Root Cause: The root cause seems to be that the C99 functionality usage has been disabled in the GNU stdlibc++ on the armeabi-v7a platform due to the fact the the Bionic libc does not support complex math (the standard C library on Android is Bionic).

http://stackoverflow.com/questions/17950814/how-to-use-stdstoul-and-stdstoull-in-android#18124627

Fayth: Well, I can add this define, but that is a hack as well. 
But yeah, as long as it's only to_string, it's not a big problem to have our impl. As far as I can see, the only methods guarded by this in basic_string header are to_string and various string->number conversions.

Fayth: Okay, I can confirm that _GLIBCXX_USE_C99 breaks armeabi-v7a compilation.

lib/VCMIDirs.cpp
alexvins: Replace `boost::filesystem::path` with `bfs::path` and put on one line

Global.h
alexvins: I suggest put smth like
`#ifdef VCMI_ANDROID` 
`#define VCMI_NO_STD_TOSTRING`
`#endif`
with short comment in macros section of global.h and finish this discussion.

PR #301, opened 2017-05-26
More war machines

alexvins:
Added an ability to make new war machines. Any artifact in conjunction with creature can become war machine.
Tested with Factory mod.

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -161,16 +161,18 @@ void CHeroArtPlace::clickLeft(tribool down, bool previousState)
 				{
 					const CArtifact * const cur = ourOwner->commonInfo->src.art->artType;
 
-					switch(cur->id)
+					if(cur->id == ArtifactID::CATAPULT)
 					{
-					case ArtifactID::CATAPULT:
 						//should not happen, catapult cannot be selected
-						assert(cur->id != ArtifactID::CATAPULT);
-						break;
-					case ArtifactID::BALLISTA: case ArtifactID::AMMO_CART: case ArtifactID::FIRST_AID_TENT: //war machines cannot go to backpack
+						logGlobal->error("Attempt to move Catapult");
+					}
+					else if (cur->isBig())
+					{
+						//war machines cannot go to backpack
 						LOCPLINT->showInfoDialog(boost::str(boost::format(CGI->generaltexth->allTexts[153]) % cur->Name()));
-						break;
-					default:
+					}
+					else
+					{
 						setMeAsDest();
 						vstd::amin(ourOwner->commonInfo->dst.slotID, ArtifactPosition(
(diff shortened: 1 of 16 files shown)


server/CGameHandler.cpp
```diff
@@ -5997,6 +5992,18 @@ void CGameHandler::putArtifact(const ArtifactLocation &al, const CArtifactInstan
 	sendAndApply(&pa);
 }
 
+bool CGameHandler::giveHeroNewArtifact(const CGHeroInstance* h, const CArtifact* art)
```
ArseniyShestakov: Since we agree on formatting of pointers that would be good to keep them consistent in new code.

server/CGameHandler.h
```diff
@@ -143,6 +143,7 @@ class CGameHandler : public IGameCallback, CBattleInfoCallback
 
 	void removeAfterVisit(const CGObjectInstance *object) override;
 
+	bool giveHeroNewArtifact(const CGHeroInstance *h, const CArtifact *art);
```
ArseniyShestakov: Formatting here too.

config/artifacts.json
```diff
@@ -17,22 +17,26 @@
...
-		"type" : ["HERO"]
+		"type" : ["HERO"],
+		"warMachine":"ammoCart"
 	},
 	"firstAidTent":
 	{
 		"index" : 6,
-		"type" : ["HERO"]
+		"type" : ["HERO"],
+		"warMachine":"firstAidTent"
```
ArseniyShestakov: Formatting.

client/widgets/CArtifactHolder.cpp
```diff
@@ -159,16 +159,18 @@ void CHeroArtPlace::clickLeft(tribool down, bool previousState)
...
+					if(cur->id == ArtifactID::CATAPULT)
 					{
-					case ArtifactID::CATAPULT:
 						//should not happen, catapult cannot be selected
-						assert(cur->id != ArtifactID::CATAPULT);
-						break;
-					case ArtifactID::BALLISTA: case ArtifactID::AMMO_CART: case ArtifactID::FIRST_AID_TENT: //war machines cannot go to backpack
+						logGlobal->error("Attempt to move Catapult");
+					}
+					else if (cur->isBig())
```
ArseniyShestakov: Formatting.

ArseniyShestakov:
So I tested this one with assets zeryss provided (you can find it in chat, mod.json have extra space but config is correct) and it's works. Though his custom cannon doesn't shot on it's own and I guess it's also hardcoded somewhere.

Other than that only problem I noticed it's crash within CCreatureAnimation::setType (second assert), but that might be due to wrong mod config.

ArseniyShestakov:
Like I posted before I not sure how correct config available for me is so not sure if there additional work required to make it shoot automatically. That's would be nice to implement, but if you want to work on something else this can be merged now.

PR #303, opened 2017-05-28
Aesthetic changes in BattleHex.

FeniksFire:
I just wanted to remove a couple of annoying comments.

### DIFF
--- a/lib/BattleHex.cpp
+++ b/lib/BattleHex.cpp
@@ -126,29 +126,16 @@ BattleHex BattleHex::operator+(BattleHex::EDir dir) const
 std::vector<BattleHex> BattleHex::neighbouringTiles() const
 {
 	std::vector<BattleHex> ret;
-	checkAndPush(cloneInDirection(TOP_LEFT, true), ret);
-	checkAndPush(cloneInDirection(TOP_RIGHT, true), ret);
-	checkAndPush(cloneInDirection(RIGHT, true), ret);
-	checkAndPush(cloneInDirection(BOTTOM_RIGHT, true), ret);
-	checkAndPush(cloneInDirection(BOTTOM_LEFT, true), ret);
-	checkAndPush(cloneInDirection(LEFT, true), ret);
+	for(EDir dir = EDir(0); dir <= EDir(5); dir = EDir(dir+1))
+		checkAndPush(cloneInDirection(dir, true), ret);
 	return ret;
 }
 
 signed char BattleHex::mutualPosition(BattleHex hex1, BattleHex hex2)
 {
-	if(hex2 == hex1.cloneInDirection(TOP_LEFT,true))
-		return TOP_LEFT;
-	if(hex2 == hex1.cloneInDirection(TOP_RIGHT,true))
-		return TOP_RIGHT;
-	if(hex2 == hex1.cloneInDirection(RIGHT,true))
-		return RIGHT;
-	if(hex2 == hex1.cloneInDirection(BOTTOM_RIGHT,true))
-		return BOTTOM_RIGHT;
-	if(hex2 == hex1.cloneInDirection(BOTTOM_LEFT,true))
-		return BOTTOM_LEFT;
-	if(hex2 == hex1.cloneInDirection(LEFT,true))
-		return LEFT;
+	for(EDir dir = EDir(0); dir <= EDir(5); dir = EDir(dir+1))
+		if(hex2 == hex1.cloneInDirection(dir,true))
+			return dir;
 	return INVALID;
 }
 
--- a/lib/BattleHex.cpp
+++ b/lib/BattleHex.cpp
@@ -127,14 +127,14 @@ std::vector<BattleHex> BattleHex::neighbouringTiles() const
 {
 	std::vector<BattleHex> ret;
 	for(EDir dir = EDir(0); dir <= EDir(5); dir = EDir(dir+1))
-		checkAndPush(cloneInDirection(dir, true), ret);
+		checkAndPush(cloneInDirection(dir, false), ret);
 	return ret;
 }
 
 signed char BattleHex::mutualPosition(BattleHex hex1, BattleHex hex2)
 {
 	for(EDir dir = EDir(0); dir <= EDir(5); dir = EDir(dir+1))
-		if(hex2 == hex1.cloneInDirection(dir,true))
+		if(hex2 == hex1.cloneInDirection(dir,false))
 			return dir;
 	return INVALID;
 }
--- a/lib/BattleHex.cpp
+++ b/lib/BattleHex.cpp
@@ -76,7 +76,7 @@ std::pair<si16, si16> BattleHex::getXY() const
 	return std::make_pair(getX(), getY());
 }
 
-BattleHex& BattleHex::moveInDir(EDir dir, bool hasToBeValid)
+BattleHex& BattleHex::moveInDirection(EDir dir, bool hasToBeValid)
 {
 	si16 x(getX()), y(getY());
 	switch(dir)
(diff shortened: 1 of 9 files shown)


lib/BattleHex.h
```diff
@@ -31,17 +31,13 @@ struct DLL_LINKAGE BattleHex //TODO: decide if this should be changed to class f
...
+	BattleHex& moveInDirection(EDir dir, bool hasToBeValid = true);
+	BattleHex& operator+=(EDir dir);
+	BattleHex generateHexMovedInDirection(EDir dir, bool hasToBeValid = true) const;
 	BattleHex operator+(EDir dir) const;
 	std::vector<BattleHex> neighbouringTiles() const;
-	//returns info about mutual position of given hexes (-1 - they're distant, 0 - left top, 1 - right top, 2 - right, 3 - right bottom, 4 - left bottom, 5 - left)
 	static signed char mutualPosition(BattleHex hex1, BattleHex hex2);
-	//returns distance between given hexes
-	static char getDistance(BattleHex hex1, BattleHex hex2);
+	static char getDistanceBetweenHexes(BattleHex hex1, BattleHex hex2);
```
alexvins: this is obvious even without "BetweenHexes"

lib/BattleHex.h
```diff
@@ -31,17 +31,13 @@ struct DLL_LINKAGE BattleHex //TODO: decide if this should be changed to class f
...
-	BattleHex& moveInDir(EDir dir, bool hasToBeValid = true);
-	BattleHex& operator+=(EDir dir); //sugar for above
-	//generates new BattleHex moved by given dir
-	BattleHex movedInDir(EDir dir, bool hasToBeValid = true) const;
+	BattleHex& moveInDirection(EDir dir, bool hasToBeValid = true);
+	BattleHex& operator+=(EDir dir);
+	BattleHex generateHexMovedInDirection(EDir dir, bool hasToBeValid = true) const;
 	BattleHex operator+(EDir dir) const;
 	std::vector<BattleHex> neighbouringTiles() const;
-	//returns info about mutual position of given hexes (-1 - they're distant, 0 - left top, 1 - right top, 2 - right, 3 - right bottom, 4 - left bottom, 5 - left)
```
alexvins: This requires documentation anyway.

lib/BattleHex.h
```diff
@@ -31,17 +31,13 @@ struct DLL_LINKAGE BattleHex //TODO: decide if this should be changed to class f
...
 	si16 getY() const;
 	std::pair<si16, si16> getXY() const;
-	//moving to direction
-	BattleHex& moveInDir(EDir dir, bool hasToBeValid = true);
-	BattleHex& operator+=(EDir dir); //sugar for above
-	//generates new BattleHex moved by given dir
-	BattleHex movedInDir(EDir dir, bool hasToBeValid = true) const;
+	BattleHex& moveInDirection(EDir dir, bool hasToBeValid = true);
+	BattleHex& operator+=(EDir dir);
+	BattleHex generateHexMovedInDirection(EDir dir, bool hasToBeValid = true) const;
```
alexvins: I suggest `cloneInDirection`

DjWarmonger: How about `cloneAndMove?`

lib/BattleHex.h
FeniksFire: You can have different distances, name should explain all things so that in the worst case everything is clear.

lib/BattleHex.h
FeniksFire: I don't agree, only library need documentation for me, comments make trash in the other code. In the definition you have everything beautifully explained, I have changed it to be so.

lib/BattleHex.h
alexvins: Name should explain distance calculation formula but not the fact that parameters are hexes

FeniksFire: you convinced me

lib/BattleHex.h
alexvins: If you suggest removing nearly all comments from class declarations, this requires discussion out of PR scope.

lib/BattleHex.cpp
```diff
@@ -102,58 +102,54 @@ BattleHex& BattleHex::moveInDir(EDir dir, bool hasToBeValid)
...
-	// H3 order : TR, R, BR, BL, L, TL (T = top, B = bottom ...)
-
-	checkAndPush(hex - ( (hex/WN)%2 ? WN+1 : WN ), ret); // 1
-	checkAndPush(hex + 1, ret); // 2
-	checkAndPush(hex + ( (hex/WN)%2 ? WN : WN+1 ), ret); // 3
-	checkAndPush(hex + ( (hex/WN)%2 ? WN-1 : WN ), ret); // 4
-	checkAndPush(hex - 1, ret); // 5
-	checkAndPush(hex - ( (hex/WN)%2 ? WN : WN-1 ), ret); // 6
-
+	checkAndPush(cloneInDirection(TOP_LEFT, true), ret);
```
alexvins: Didn`t you changed the order of returning hexes? It is important for game mechanics.

lib/BattleHex.cpp
```diff
@@ -102,58 +102,54 @@ BattleHex& BattleHex::moveInDir(EDir dir, bool hasToBeValid)
...
-	if(hex2 == hex1 + 1 && hex1%17 != 16) //right
-		return 2;
-	if(hex2 == hex1 + ( (hex1/17)%2 ? 17 : 18 )) //bottom right
-		return 3;
-	if(hex2 == hex1 + ( (hex1/17)%2 ? 16 : 17 )) //bottom left
-		return 4;
-	if(hex2 == hex1 - 1 && hex1%17 != 0) //left
-		return 5;
-	return -1;
+	if(hex2 == hex1.cloneInDirection(TOP_LEFT,true))
```
alexvins: This should be cycle but not 6 identical statements.

lib/BattleHex.h
```diff
@@ -15,7 +15,7 @@ struct DLL_LINKAGE BattleHex //TODO: decide if this should be changed to class f
 {
 	si16 hex;
 	static const si16 INVALID = -1;
-	enum EDir { RIGHT, BOTTOM_RIGHT, BOTTOM_LEFT, LEFT, TOP_LEFT, TOP_RIGHT };
+	enum EDir { TOP_LEFT, TOP_RIGHT, RIGHT, BOTTOM_RIGHT, BOTTOM_LEFT, LEFT};
```
alexvins: Any reason to reorder?

test/Battlefield.cpp
```diff
@@ -37,12 +36,12 @@ BOOST_AUTO_TEST_CASE(getNeighbouringTiles)
 	BOOST_TEST(neighbouringTiles.size()==6);
 
 	BOOST_REQUIRE(neighbouringTiles.size()==6 && mainHex==93);
-	BOOST_TEST(neighbouringTiles.at(0)==75);
-	BOOST_TEST(neighbouringTiles.at(1)==94);
-	BOOST_TEST(neighbouringTiles.at(2)==110);
-	BOOST_TEST(neighbouringTiles.at(3)==109);
-	BOOST_TEST(neighbouringTiles.at(4)==92);
-	BOOST_TEST(neighbouringTiles.at(5)==76);
+	BOOST_CHECK(std::find(neighbouringTiles.begin(), neighbouringTiles.end(), 75) != neighbouringTiles.end());
```
alexvins: std::find is too long, use vstd::contains. And order is important, so better to revert or check order in another test

lib/BattleHex.cpp
FeniksFire: It's fine. I check it.

lib/BattleHex.cpp
FeniksFire: It's good look to the header file, TOP_LEFT->0, TOP_RIGHT->1, RIGHT->2.... I check everything and it's work. You want use enum so I use it.

lib/BattleHex.h
FeniksFire: See, reply in: BattleHex::mutualPosition

lib/BattleHex.cpp
alexvins: // H3 order : TR, R, BR, BL, L, TL (T = top, B = bottom ...)
TOP_LEFT should be last

FeniksFire: This really don't have any meaning in practice, but I will change it.

lib/BattleHex.cpp
alexvins: This is another question:
you have same statement
`	if(hex2 == hex1.cloneInDirection(DIR,true))
		return DIR;`
6 times in a row.

FeniksFire: ok so I can reduce it to one. and It will be fine.

lib/BattleHex.cpp
FeniksFire: Omg, it's fine ehh. Look at the old code;
zrzut ekranu z 2017-05-28 22-11-49

alexvins: I`m talking not about enum values, but about order of values in returned vector.

FeniksFire: yhm, yeah and it's fine since I change order in EDir

FeniksFire:
@ArseniyShestakov Errors appear after recent updates @alexvinsa not after this. I see it when I update my code to the last vcmi/origin version. For me all works fine.

ArseniyShestakov:
Please make sure you actually built with asserts enabled.
Latest develop works fine for me, but your code fail even if not merged.

FeniksFire:
Ok my bad, I build it in debug mode now and you have right. I will fix it in a few minutes.

ArseniyShestakov:
I don't want to look into the battle code myself for long so I'll appreciate if you could explain what that check is was about and why it's need to be just disabled / not removed.

FeniksFire:
You have this function in BattleHex::neighbouringTiles :
`checkAndPush(cloneInDirection(dir, false), ret);`
1) `cloneInDirection` -> get all fields even if they aren't valid
2) ```
void BattleHex::checkAndPush(BattleHex tile, std::vector<BattleHex> & ret)
{
	if(tile.isAvailable())
		ret.push_back(tile);
}
``` check if field is available, and if so add available hexes to vector. In other way when I get cloneInDirection(dir, true), asserts make error because I add something like invalid hexes.

Now next fuction BattleHex::mutualPosition. In the original version didn't check if is valid. so I change it to be like in original.

ArseniyShestakov:
Makes sense then. Thanks.

PR #304, opened 2017-05-29
Do not show whether Witch Hut is visited by hero, if was not visited …

Chocimier:


ArseniyShestakov:
I think this kind of information should be shared and not even within player state, but for the whole team. While we don't store information for team I think it's could have callback that would make it possible to check whatever anyone in team even visited object.

I not sure if H3 behavior is different, but there certainly no point in applying artificial limitations like that.

Chocimier:
> have callback that would make it possible to check whatever anyone in team even visited object

Good point, that could be also used at banks where known is something else that visited - if you enter and not fight, player can see how many creatures are in.

So do we need something like CArmedInstance::knownToPlayer for CBank  and CPlayersVisited::knownForPlayer for CGWitchHut and similar?

ArseniyShestakov:
Yep we need something like that and it's should take in mind if one of players in team also visited it. Since there can be different kind of objects that only visitable per: whole team, player or just one hero.

PlayerState struct already have visitedObjects and ChangeObjectVisitors netpack can already apply it who the whole team. So there just need to be proper checking for that.

Chocimier:
OK, so that seems to need a little bit greater patch that current. Can you please merge what we have now to remove obvious bug? I'll improve more objects to share information inside team as next task.

ArseniyShestakov:
Of course. I just noted that if you work on that code anyway it's worth to actually improve it all around.
Thanks!

ArseniyShestakov:
And yeah very please don't forget to assign bugs on yourself on bug tracker if you working on something:
https://bugs.vcmi.eu/view.php?id=2604

PR #305, opened 2017-06-01
Network and multiplayer improvements

ArseniyShestakov:
Based off #253.

There is bunch of improvements for network code without any significant changes. Game still going to crash for guests if host disconnected, but it's should be more stable overall. Any big feature like reconnecting / loading from server game state would likely require massive refactoring since a lot of code is just bad.


server/CGameHandler.cpp
```diff
@@ -1031,6 +1031,21 @@ void CGameHandler::handleConnection(std::set<PlayerColor> players, CConnection &
 {
 	setThreadName("CGameHandler::handleConnection");
 
+	auto handleDisconnection = [&](const std::exception e)
```
alexvins: Should be reference

server/CGameHandler.cpp
```diff
@@ -1031,6 +1031,21 @@ void CGameHandler::handleConnection(std::set<PlayerColor> players, CConnection &
...
+	{
+		assert(!c.connected); //make sure that connection has been marked as broken
+		logGlobal->error(e.what());
+		conns -= &c;
+		for(auto playerConn : connections)
+		{
+			if(playerConn.second == &c)
+			{
+				gs->getPlayer(playerConn.first)->enteredLosingCheatCode = 1;
```
alexvins: You should use NetPack for GameState modification

PR #306, opened 2017-06-03
Debug features: auto tests, headless and spectator modes

ArseniyShestakov:
Spectator code have a lot of room for improvements and clearly there have to be CSpectatorInterface with it's own UI and priveleged callback access (it's kind a there, but not really implemented), but for now we need at least some solution for VCAI / BattleAI debugging and testing since working on them blind is very hard.

So far biggest things I dislike that still feasible to fix in reasonable timeshare are: lack of PlayerState so many things like FoW and netpacks need to be plumbed, also workarounds related to UI colors. Also I kind a dislike that I have to use "session settings" everywhere, but I not yet find better way to determine that player is in spectator mode.

Things I wish to implement before merging:

- [x] Hotkey to pause spectator "playback" on adventure map or in battle.
- [x] Also stopping of "playback" when menus / popups active since currently they tend to lose focus.
- [x] Hotkey to skip particular battle.
- [x] Graceful exit on end for "testing mode".

### DIFF
--- a/client/CMessage.cpp
+++ b/client/CMessage.cpp
@@ -296,6 +296,8 @@ void CMessage::drawIWindow(CInfoWindow * ret, std::string text, PlayerColor play
 
 void CMessage::drawBorder(PlayerColor playerColor, SDL_Surface * ret, int w, int h, int x, int y)
 {
+	if(playerColor.isSpectator())
+		playerColor = PlayerColor(1);
 	std::vector<const IImage*> &box = piecesOfBox.at(playerColor.getNum());
 
 	// Note: this code assumes that the corner dimensions are all the same.
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -245,6 +245,9 @@ void CPlayerInterface::heroMoved(const TryMoveHero & details)
 	if (LOCPLINT != this)
 		return;
 
+	if(settings["session"]["spectate"].Bool() && settings["session"]["spectate-ignore-hero"].Bool())
+		return;
+
 	const CGHeroInstance * hero = cb->getHero(details.id); //object representing this hero
 	int3 hp = details.start;
 
@@ -321,7 +324,12 @@ void CPlayerInterface::heroMoved(const TryMoveHero & details)
 	}
 
 	ui32 speed;
-	if (makingTurn) // our turn, our hero moves
+	if(settings["session"]["spectate"].Bool())
+	{
+		if(!settings["session"]["spectate-hero-speed"].isNull())
+			speed = settings["session"]["spectate-hero-speed"].Integer();
+	}
+	else if (makingTurn) // our turn, our hero moves
 		speed = settings["adventure"]["heroSpeed"].Float();
 	else
 		speed = settings["adventure"]["enemySpeed"].Float();
@@ -333,7 +341,6 @@ void CPlayerInterface::heroMoved(const TryMoveHero & details)
 		CGI->mh->printObject(hero);
 		return; // no animation
 	}
-
 
 	adventureInt->centerOn(hero); //actualizing screen pos
 	adventureInt->minimap.redraw();
@@ -471,6 +478,14 @@ int3 CPlayerInterface::repairScreenPos(int3 pos)
 		pos.y = CGI->mh->sizes.y - adventureInt->terrain.tilesh + CGI->mh->frameH;
 	return pos;
 }
+
+void CPlayerInterface::activateForSpectator()
+{
+	adventureInt->state = CAdvMapInt::INGAME;
+	adventureInt->activate();
+	adventureInt->minimap.activate();
+}
+
 void CPlayerInterface::heroPrimarySkillChanged(const CGHeroInstance * hero, int which, si64 val)
 {
 	EVENT_HANDLER_CALLED_BY_CLIENT;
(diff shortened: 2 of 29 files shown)


client/battle/CBattleInterface.cpp
```diff
@@ -105,12 +105,17 @@ CBattleInterface::CBattleInterface(const CCreatureSet *army1, const CCreatureSet
 {
 	OBJ_CONSTRUCTION;
 
-	if (!curInt)
+	if(spec)
```
ArseniyShestakov: Oh. This is certainly something useless. Probably remaining of my duel mode fixing attempts.

lib/CGameInfoCallback.cpp
```diff
@@ -689,6 +689,24 @@ CGameInfoCallback::CGameInfoCallback(CGameState *GS, boost::optional<PlayerColor
 const std::vector< std::vector< std::vector<ui8> > > & CPlayerSpecificInfoCallback::getVisibilityMap() const
 {
 	//boost::shared_lock<boost::shared_mutex> lock(*gs->mx);
+	if(player == PlayerColor::SPECTATOR) // TODO: need to be replaced somehow
+	{
+		static std::vector<std::vector<std::vector<ui8> > > map;
```
alexvins: having static object does not make sense if you rebuilding it on each call, also it is not thread-safe.

lib/NetPacksLib.cpp
```diff
@@ -1842,6 +1842,9 @@ DLL_LINKAGE void BattleSetStackProperty::applyGs(CGameState *gs)
 
 DLL_LINKAGE void PlayerCheated::applyGs(CGameState *gs)
 {
+	if(player == PlayerColor::SPECTATOR)
```
alexvins: Here should be more generic check for valid player (check for PLAYER_LIMIT_I)

client/CPlayerInterface.cpp
```diff
@@ -1642,7 +1662,8 @@ void CPlayerInterface::update()
 	}
 
 	//in some conditions we may receive calls before selection is initialized - we must ignore them
-	if (adventureInt && !adventureInt->selection && GH.topInt() == adventureInt)
+	if (adventureInt && GH.topInt() == adventureInt
+		&& (!adventureInt->selection && LOCPLINT->playerID != PlayerColor::SPECTATOR))
```
alexvins: this method is inside CPlayerInterface try to do not use LOCPLINT. Also you already have CPlayerInterface::isSpectator()

client/CMessage.cpp
```diff
@@ -296,6 +296,8 @@ void CMessage::drawIWindow(CInfoWindow * ret, std::string text, PlayerColor play
 
 void CMessage::drawBorder(PlayerColor playerColor, SDL_Surface * ret, int w, int h, int x, int y)
 {
+	if(playerColor == PlayerColor::SPECTATOR)
+		playerColor = PlayerColor(1);
```
alexvins: This chunk is duplicated many times. Consider adding a function|method like
 `PlayerColor playerColorToColor(const PlayerColor & color)` 
somewhere

client/windows/CWindowObject.cpp
```diff
@@ -225,9 +225,13 @@ void CWindowObject::setShadow(bool on)
 
 void CWindowObject::showAll(SDL_Surface *to)
 {
+	auto color = LOCPLINT ? LOCPLINT->playerID : PlayerColor(1);
```
alexvins: Move this line to new static method of CPlayerInterface.

client/CPlayerInterface.cpp
ArseniyShestakov: CPlayerInterface::isSpectator is bad solution sadly since I don't want even more complex check everywhere and LOCPLINT might be not present when run headless. For now I just decide to check settings "session" everywhere since it's globally visible, but can't really say I'd like it.

ArseniyShestakov:
I tried to add sever shutdown sync with std::atomic so I hope it's supported well enough in compilers we use. If needed I'll implement it some other way.

ArseniyShestakov:
I implemented all large features I wanted. Now only minor fixes.

client/CPlayerInterface.cpp
alexvins: in headless mode CPlayerInterface::update() should not be called

ArseniyShestakov: Yep, but I still wish to have one centralized check and CPlayerInterface isn't place for it.

ArseniyShestakov:
@Fayth I'll appreciate if you could check in background if changes in this branch break something on Android. I tried to not mess it with interprocess, but I suppose it's might fail still. So if it's wouldn't take long please test it so I can rework it before it's merged.

Fayth:
Anything specific that I should check?
It compiles correctly and the game starts fine.

ArseniyShestakov:
@Fayth nothing really. I just worried about std::atomic and my changes to server startup.
Thank you very much!

client/gui/CGuiHandler.cpp
```diff
@@ -191,6 +192,52 @@ void CGuiHandler::handleEvent(SDL_Event *sEvent)
...
+				s["session"]["spectate-skip-battle"].Bool() = !s["session"]["spectate-skip-battle"].Bool();
+				break;
+
+			case SDLK_F8:
+				s["session"]["spectate-skip-battle-result"].Bool() = !s["session"]["spectate-skip-battle-result"].Bool();
+				break;
+
+			case SDLK_F9:
+				//not working yet since CClient::run remain locked after CBattleInterface removal
+				boost::unique_lock<boost::recursive_mutex> un(*LOCPLINT->pim);
```
alexvins: do not use lock here

ArseniyShestakov:
So now this one is ready. I worry that few of added hacks and "settings" check could affect performance badly, but probably it's won't be big deal.

@alexvins I didn't change those lines related to PlayerColor because they're also related to the old duel mode code that I still want to purge in separate PR. Also I'd rather not to try "improve" it since it's won't change anything. Actual proper solution for the spectator code in this case is to give it's own CSpectatorInterface class and make adjustments for other UI code, but that's not something I about to do right now.

PR #307, opened 2017-06-04
Added option for saving generated maps on client side

alexvins:
* new configuration option 'general.saveRandomMaps'
* maps being saved to 'userCachePath/RandomMaps'
* no deletion of old random maps
* map filename generated based on template name and random seed

### DIFF
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -408,10 +408,9 @@ void CClient::newGame( CConnection *con, StartInfo *si )
 
 	// Initialize game state
 	gs = new CGameState();
-	logNetwork->infoStream() <<"\tCreating gamestate: "<<tmh.getDiff();
+	logNetwork->info("\tCreating gamestate: %i",tmh.getDiff());
 
-	gs->scenarioOps = si;
-	gs->init(si);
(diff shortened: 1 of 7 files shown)


ArseniyShestakov:
Can't start game on saved maps: all player slots marked as computer-only.

alexvins:
> 2017-Jun-04 20:23:01.018554 INFO global [1b68] - Final player config: 8 total, 8 cpu-only

This is how RMG works

ArseniyShestakov:
Then no reason not to merge. I'll try to fix UI after then.

alexvins:
The thing is that player options are partly ignored for random map - StartInfo is used instead that why this bug never appears before.

ArseniyShestakov:
Yeah both UI and RMG need some work. When this one merged I'll try to implement proper restart so saved map can be used. E.g in H3 client doesn't re-generate map, but change seed and restart it.

ArseniyShestakov:
Yeah both UI and RMG need some work. When this one merged I'll try to implement proper restart so saved map can be used. E.g in H3 client doesn't re-generate map, but change seed and restart it.

PR #310, opened 2017-06-07
Adventure map swipe support for non-android platforms

Fayth:
This PR adds support for swiping adventure map, partially reusing recent android implementation.
Currently "desktop" version of swipe uses middle mouse button (default "left" button on android) and can be enabled simultaneously with normal map scrolling (these modes are currently exclusive in android version).

First commit is a minor refactor of handling mouse button events in CGuiHandler (removes some duplicated code) + middle mouse button handling.
Second is actual swipe logic (updated version tested on windows and on android).

The setting that controls swipe (general -> swipe) can't be modified in-game so I changed it to be enabled by default (just for now?).

### DIFF
--- a/client/gui/CGuiHandler.h
+++ b/client/gui/CGuiHandler.h
@@ -10,6 +10,7 @@ class CIntObject;
 class IUpdateable;
 class IShowActivatable;
 class IShowable;
+enum class EIntObjMouseBtnType;
 template <typename T> struct CondSh;
 
 /*
@@ -53,6 +54,7 @@ private:
 	//active GUI elements (listening for events
 	CIntObjectList lclickable,
 				   rclickable,
+				   mclickable,
 				   hoverable,
 				   keyinterested,
 				   motioninterested,
@@ -62,6 +64,7 @@ private:
 	               textInterested;
 
 
+	void handleMouseButtonClick(CIntObjectList &interestedObjs, EIntObjMouseBtnType btn, bool isPressed);
 	void processLists(const ui16 activityFlag, std::function<void (std::list<CIntObject*> *)> cb);
 public:
 	void handleElementActivate(CIntObject * elem, ui16 activityFlag);
--- a/client/gui/CGuiHandler.h
+++ b/client/gui/CGuiHandler.h
@@ -64,7 +64,7 @@ private:
 	               textInterested;
 
 
-	void handleMouseButtonClick(CIntObjectList &interestedObjs, EIntObjMouseBtnType btn, bool isPressed);
+	void handleMouseButtonClick(CIntObjectList & interestedObjs, EIntObjMouseBtnType btn, bool isPressed);
 	void processLists(const ui16 activityFlag, std::function<void (std::list<CIntObject*> *)> cb);
 public:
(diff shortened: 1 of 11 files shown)


DjWarmonger:
Interesting :). However, the trackball I use has no middle mouse button so I can't really test it.

client/gui/CGuiHandler.h
```diff
@@ -62,6 +64,7 @@ class CGuiHandler
 	               textInterested;
 
 
+	void handleMouseButtonClick(CIntObjectList &interestedObjs, EIntObjMouseBtnType btn, bool isPressed);
```
ArseniyShestakov: Add space after & there.

client/widgets/AdventureMapClasses.cpp
```diff
@@ -587,7 +587,7 @@ void CMinimap::hover(bool on)
 
 void CMinimap::mouseMoved(const SDL_MouseMotionEvent & sEvent)
 {
-	if (pressedL)
+	if (mouseState(EIntObjMouseBtnType::LEFT))
```
ArseniyShestakov: Remove that extra space here.

client/gui/CIntObject.cpp
```diff
@@ -340,16 +357,9 @@ void CKeyShortcut::keyPressed(const SDL_KeyboardEvent & key)
...
-		{
-			pressedL = true;
-			clickLeft(true, prev);
-		}
-		else
-		{
-			pressedL = false;
-			clickLeft(false, prev);
-		}
+		bool prev = mouseState(EIntObjMouseBtnType::LEFT);
```
ArseniyShestakov: I suppose "prev" can now be removed, just pass result to clickLeft.

client/gui/CGuiHandler.h
Fayth: Yeaaah, sorry. I was trying to set this in visual studio formatter and then kinda forgot (and, apparently, visual studio doesn't have the option to change that...). Will fix.

I'm not really up-to-date with c development -- isn't there some portable way for a project to have correct code style used by IDEs by default? Clang format?

Fayth:
@DjWarmonger well, technically it could be made configurable (to change bool in the config to int/enum defining required mouse button). The only problem I can see is that using left/right could somehow disrupt normal clicking... it might be okay, but keep in mind that swipe requires "normal" button actions to be resolved in "release"/"up" state instead of "press"/"down" (as it is now in adventure map).

client/gui/CGuiHandler.h
ArseniyShestakov: To be fair it's something I like to know too since it's would be much better if we can add style checking into Travis once and for all. Qt Creator have Beautifier plugin that support 3 different formatters include clang-format, but they all of course require custom configurations since VCMI on VCMI we have somewhat mixed style.

And I never had time to configure formatter.

Fayth: I can later check if clang-format + some quick formatting config would work (separate from this PR).

ArseniyShestakov: If you can please do. Thanks.

ArseniyShestakov:
One more small request: keep git subject line below 50 characters when possible. Thanks.

PR #312, opened 2017-06-14
Remove code related to duel mode

ArseniyShestakov:
It's broken since forever and need cleaner reimplementation anyway.
New implementation will be much cleaner due to our map format and spectator mode.

@FeniksFire is working on battle code refactoring I'll wait for his changes if they conflict with this branch.
So please do not merge.


DjWarmonger:
I think we should disable it, but not remove completely. Even if it stays only for reference.

Why is unused code an issue, anyway?

alexvins:
@DjWarmonger if someone need a reference, he can see commit history

ArseniyShestakov:
To be fair I was one who wanted to keep it until I really tried to fix and use it. Problem isn't that code unused, but fact that it's was added long before VCMI had mods support and map format so it's add bunch of hacks all around the code.

Today all you need to make AI attack some army is set it to Event or just on top of exit monolith in case of heroes battles or just create some DuelAI that attack everything it's sees. And if we ever need some configuration tool Alex could probably add some "duel editor" panel in few hours of work, but that's not really required.

DjWarmonger:
I can see that game handler / AI config is obsolete, but what's wrong with json parser?

alexvins:
Duel JSON format is too old: uses only numeric id`s - so no mods support.

ArseniyShestakov:
What point to have separate duel config parser if we have map format?

Even customized creatures can be added via Mods. Only thing that can't be handled with mods is obstacle, but let's say fair is there any possible reason why someone would want to test some special obstacle placement?

DjWarmonger:
You convinced me.

But yes, I can absolutely see why someone might want to test obstacle placement - such as new obstacles or summon spells / abilities.

PR #314, opened 2017-06-21
Share visited status of CPlayersVisited objects across team

Chocimier:
It makes sense for all such objects - cartographer, obelisk, shrine and witch hut. 
Should it be renamed to CTeamVisited?

### DIFF
--- a/lib/mapObjects/MiscObjects.cpp
+++ b/lib/mapObjects/MiscObjects.cpp
@@ -60,18 +60,18 @@ static std::string & visitedTxt(const bool visited)
 	return VLC->generaltexth->allTexts[id];
 }
 
-void CPlayersVisited::setPropertyDer( ui8 what, ui32 val )
+void CTeamVisited::setPropertyDer( ui8 what, ui32 val )
 {
-	if(what == CPlayersVisited::OBJPROP_VISITED)
+	if(what == CTeamVisited::OBJPROP_VISITED)
 		players.insert(PlayerColor(val));
 }
 
-bool CPlayersVisited::wasVisited( PlayerColor player ) const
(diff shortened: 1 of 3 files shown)


DjWarmonger:
Probably yes. Now I see wasVisited() calls wasVisited() :v

ArseniyShestakov:
Actually I expect we could have objects that visiting status might be not shared across team. Thoughts?

Chocimier:
I see that CPlayersVisited does nearly the same what is done with CRewardableObject::VISIT_PLAYER.
So we can move CRewardableObject::wasVisited to CGObjectInstance, add there VISIT_TEAM and remove CPlayersVisited.

Another issue: what it means for bank to be visited: look into and know what guards are there or defeat?

ArseniyShestakov:
We cannot just "move" them for two reasons:

- There is save serialization breakage involved.
- Those objects remained separate exactly because map object refactoring wasn't completed.

So unless you want to do some heavy refactoring and complete what Ivan started long ago there is no reason to toss code around at all. Might be we could rename them then and leave everything as is if CRewardable object can already handle object that visited once per player OR team.

Chocimier:
Yes, CRewardable can handle per player visit and also can handle per team visit.

Sorry, but I do not understand - what am I supposed to do now?

ArseniyShestakov:
Sorry for not answering straight away. If all objects still using CPlayersVisited should be "team visited" then you can rename it and please add comment that it's legacy class and it's not to be used.

As long as it's not break anything that should be fine.

ArseniyShestakov:
I didn't want to bother you to fix formatting, so I just squashed / merged it manually.
Next time please fix formatting along the way. Thanks.

PR #317, opened 2017-07-01
Refactored CStack::attackerOwned to CStack::side

alexvins:
_not tested_

### DIFF
--- a/lib/CStack.h
+++ b/lib/CStack.h
@@ -24,7 +24,7 @@ public:
 	ui32 firstHPleft; //HP of first creature in stack
 	PlayerColor owner; //owner - player colour (255 for neutrals)
 	SlotID slot;  //slot - position in garrison (may be 255 for neutrals/called creatures)
-	bool attackerOwned; //if true, this stack is owned by attakcer (this one from left hand side of battle)
+	ui8 side;
 	BattleHex position; //position on battlefield; -2 - keep, -3 - lower tower, -4 - upper tower
 	///how many times this stack has been counterattacked this round
 	ui8 counterAttacksPerformed;
@@ -39,8 +39,8 @@ public:
 	//overrides
 	const CCreature* getCreature() const {return type;}
 
-	CStack(const CStackInstance *base, PlayerColor O, int I, bool AO, SlotID S); //c-tor
-	CStack(const CStackBasicDescriptor *stack, PlayerColor O, int I, bool AO, SlotID S = SlotID(255)); //c-tor
+	CStack(const CStackInstance *base, PlayerColor O, int I, ui8 Side, SlotID S); //c-tor
+	CStack(const CStackBasicDescriptor *stack, PlayerColor O, int I, ui8 Side, SlotID S = SlotID(255)); //c-tor
 	CStack(); //c-tor
 	~CStack();
 	std::string nodeName() const override;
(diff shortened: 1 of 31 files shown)


lib/battle/ReachabilityInfo.h
```diff
@@ -27,7 +27,7 @@ struct DLL_LINKAGE ReachabilityInfo
 	{
 		const CStack * stack; //stack for which calculation is mage => not required (kept for debugging mostly), following variables are enough
 
-		bool attackerOwned;
+		ui8 side;
```
ArseniyShestakov: Might be we should use strongly typed ENUM instead or something?
Since replacing true/false with 0/1 doesn't make it easier to understand.

I suppose H3 always put attacker on left side and defenders on the right so...
Though I wish to know what @FeniksFire thinking about it since he was working on battle code refactoring.

FeniksFire: @alexvins btw, any reason to change bool to ui8?  Enum is nice solution but not here, we should use struct called SideInBattle. Maybe you can make enum there instead of `std::array<SideInBattle, 2> sides;` This will for sure improve legibility of the code.

FeniksFire:
Can you replace `ui8 side`, to `BattleSide side`?

lib/battle/BattleHex.h
```diff
@@ -8,7 +8,15 @@
  *
  */
 #pragma once
-#include "../GameConstants.h"
+
+namespace BattleSide
+{
+	enum
```
ArseniyShestakov: You can actually make it ui8 too.

FeniksFire: ok, np, when you merge it I will change it to BattleSide. Need to change one line I think ;)

lib/CStack.h
```diff
@@ -77,6 +77,8 @@ class DLL_LINKAGE CStack : public CBonusSystemNode, public CStackBasicDescriptor
 	bool coversPos(BattleHex position) const; //checks also if unit is double-wide
 	std::vector<BattleHex> getSurroundingHexes(BattleHex attackerPos = BattleHex::INVALID) const; // get six or 8 surrounding hexes depending on creature size
 
+    BattleHex::EDir destShiftDir() const;
```
ArseniyShestakov: Tabs

lib/battle/BattleHex.h
alexvins: It can be changed to `enum class BattleSide : ui8` everywhere, but then it will require casts to access BattleInfo::sides

ArseniyShestakov: Then it's not important. I just noted in case you might missed it.

lib/battle/ReachabilityInfo.h
alexvins: next commit adds usage of enum

ArseniyShestakov:
Branch itself seems okay, so please squash it for now.

Though I found some regression and I not yet sure why it's appear, probably in one of other merges.
Hotseat savegames tend to crash on load and other games also crashing...

alexvins:
rebased

ArseniyShestakov:
Not sure if you'll find it useful but here is backtrace:

`5  BattleHex::setXY                   BattleHex.cpp        56   0x7ffff6df1bfa 
6  BattleHex::moveInDirection         BattleHex.cpp        92   0x7ffff6df1e04 
7  BattleHex::cloneInDirection        BattleHex.cpp        120  0x7ffff6df1f6c 
8  BattleHex::operator+               BattleHex.cpp        126  0x7ffff6df1fab 
9  CBattleInterface::canStackMoveHere CBattleInterface.cpp 2039 0x5555557cb563 
10 CBattleInterface::handleHex        CBattleInterface.cpp 2109 0x5555557cc30d 
11 CBattleInterface::mouseMoved       CBattleInterface.cpp 585  0x5555557c1f69 
12 CGuiHandler::handleMoveInterested  CGuiHandler.cpp      407  0x55555581891c 
13 CGuiHandler::handleMouseMotion     CGuiHandler.cpp      387  0x555555818741 
14 CGuiHandler::handleEvent           CGuiHandler.cpp      262  0x555555817919 
15 CGuiHandler::handleEvents          CGuiHandler.cpp      184  0x555555816d2b 
16 CPlayerInterface::update           CPlayerInterface.cpp 1660 0x555555b43162 
17 CGuiHandler::renderFrame           CGuiHandler.cpp      446  0x555555818b3f 
18 mainLoop                           CMT.cpp              1306 0x555555af6e10 
19 main                               CMT.cpp              561  0x555555af064c`

PR #319, opened 2017-07-01
Improve VS build filters file + precompiled header path fix

dydzio0614:
Add "battle" virtual folder to match new physical "battle" folder in vcmi/lib

EDIT: also need to fix paths to let visual studio compile project after recent changes.


dydzio0614:
Could we have this PR merged? VCMI does not even compile without 2nd commit using settings from vcxproj files that are in repo

PR #320, opened 2017-07-02
AppVeyor integration

viciious:
This MR integrates vcmi with AppVeyor. The needed dependencies are available for download here: https://dependencies.vcmi.download/msvc/vcmi_depends_x86.zip although I recommend uploading them elsewhere (say, github releases). Only x86 is supported at the moment.

The produced binaries along with runtime dependencies are available as build artifacts. The bundled binaries have been verified to run on a Windows 7 system.
The installer produced by cpack on tag releases is not included.

Following the merge someone will probably need to setup the vcmi project AppVeyor and maybe tweak a few things around.


ArseniyShestakov:
I'll mess with CI scripts here a bit and then I'll manually merge this one.

ArseniyShestakov:
Ok I merged it, but someone still need to enable webhooks because I don't have permissions.

PR #321, opened 2017-07-02
Fix NodeComparer::operator() to provide strict weak ordering

viciious:
This fixes annoying assertion failures in MSVC builds


viciious:
A small clarification of what the change does: it tweaks the comparator to return false when it's passed with two identical nodes, otherwise the strict weak ordering breaks.

PR #322, opened 2017-07-02
Fix attack cursor on guarded visitable object

dydzio0614:
There is good chance that options.originalMovementRules check should not exist in first place. After using fly spell in OH3 you get attacked immediately when visiting guarded object.


alexvins:
Changes were cherry picked.

PR #323, opened 2017-07-04
CStack tweaks

alexvins:
* Unified CStack ammo, casts and counterattacks
* it is possible now to add casts and shoots OTF (f.e. with spell bonus)
* https://bugs.vcmi.eu/view.php?id=2486
* https://bugs.vcmi.eu/view.php?id=2584

### DIFF
--- a/test/battle/CHealthTest.cpp
+++ b/test/battle/CHealthTest.cpp
@@ -20,11 +20,16 @@ class CUnitHealthInfoMock : public IUnitHealthInfo
 public:
 	CUnitHealthInfoMock():
 		maxHealth(UNIT_HEALTH),
-		baseAmount(UNIT_AMOUNT)
-	{}
+		baseAmount(UNIT_AMOUNT),
+		health(this)
+	{
+		health.init();
+	}
 
 	int32_t maxHealth;
 	int32_t baseAmount;
+
+	CHealth health;
 
 	int32_t unitMaxHealth() const override
 	{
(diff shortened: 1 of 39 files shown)


test/battle/CHealthTest.cpp
```diff
@@ -0,0 +1,221 @@
...
+	health.takeResurrected();
+	checkEmptyHealth(health, uhi);
+}
+
+BOOST_AUTO_TEST_CASE(resurrectPermanent)
+{
+	CUnitHealthInfoMock uhi;
+	CHealth health(&uhi);
+	health.init();
+
```
FeniksFire: For me this is code duplication. You should make fixture for handling it. You can even change CUnitHealthInfoMock class for that.

alexvins: converted to fixture

ArseniyShestakov:
Povelitel has reposted that in this branch Archangels stack can cast resurrect any number of times within battle.

ArseniyShestakov:
Ok there another bug. After battle all stacks that were killed are removed from army, but all damaged stacks restored in full strength like they wasn't in battle at all.

- Go to battle with any weak monsters or scattered. E.g I had 21 centaurs against 7 stacks of 2 vampires (14 in total).
- Win it by casting Armageddon, but only 3 of 18 centaurs remained.
- Now after battle I have 21 of them again.

ArseniyShestakov:
I don't find any more obvious regressions that might affect playability.
Should be sufficient to merge it now and test it.

PR #325, opened 2017-07-04
macOS: use HTTPS for appcast 

dunn:
Otherwise a warning pops up at each launch (until http://vcmi.eu/sparkle_appcast.xml is upgraded to HTTPS).

See https://sparkle-project.org/documentation/app-transport-security/

### DIFF
--- a/client/Info.plist
+++ b/client/Info.plist
@@ -9,7 +9,7 @@
 	<key>CFBundleIdentifier</key>
 	<string>com.vcmi.vcmiclient</string>
 	<key>SUFeedURL</key>
-	<string>http://vcmi.eu/sparkle_appcast.xml</string>
+	<string>https://vcmi.eu/sparkle_appcast.xml</string>
 	<key>SUPublicDSAKeyFile</key>
 	<string>vcmi_dsa_public.pem</string>
 </dict>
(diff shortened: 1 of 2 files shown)


dunn:
Looks like I need to add `NSAllowsArbitraryLoads` as well:
screen shot 2017-07-04 at 4 34 21 pm

dunn:
`NSAllowsArbitraryLoads` needs to be added to `vcmibuilder.app`'s Info.plist, so I'll make that a separate PR.

dunn:
(Unless the WoG download can be upgraded to HTTPS.)

ArseniyShestakov:
At moment vcmi.eu is all behind CloudFlare, but since we don't really ship any updates for macOS builds at all this is fine. There are no other problems due to fact that we actually missing sparkle_appcast.xml completely?

dunn:
Yeah, as far as I know it just sees `http://` and issues the warning.  So instead I could just change those strings to `https://`.

Or, if Sparkle isn't currently being used, could VCMI build without it by default?

ArseniyShestakov:
Someone just need to implement such option then. Few people on HN said they might try to help us, but whatever they going to appear I have no idea. As long as fix in this PR sufficient to make VCMI startup properly I could merge this.

When we going to have working DMGs with all dependencies included then we can consider fixing Sparkle too.

dunn:
OK, tomorrow I'll double-check that changing the appcast URL to HTTPS works, and we can do that instead.  Unfortunately I don't know C++ so I don't think I can deal with the Sparkle option.

ArseniyShestakov:
I managed to make it build in VM with Sierra and it's actually working for me when vcmiclient binary launched directly from terminal. So errors are certainly not critical at least.

Just in case you can avoid using "vcmibuilder" by creating "~/Library/Application Support/vcmi/game_data_prepared" file and manually moving Data, MP3 and other directories into that directory.

PR #326, opened 2017-07-05
macOS: use HTTPS for downloads

dunn:
This fixes the ATS errors I saw in vcmibuilder here: https://github.com/vcmi/vcmi/pull/325#issuecomment-312969419


ArseniyShestakov:
I really unsure whatever this one as well as #325 really needed since VCMI have all required assets included into repository Mods directory and WoG is no longer required, but can be downloaded as standalone mod via launcher instead.

dunn:
So we can just remove both download steps instead?  I'll try to figure out how to do that, but in the meantime this lets the application launch.

ArseniyShestakov:
Then let's just merge it if it's essential to make it work for now, but please don't waste your time on fixing it because we certainly need cross-platform solution and not just one that only work on Mac.

PR #329, opened 2017-07-07
Change output path for cmake builds

viciious:
All binary and libraries files will now go into the same directory, suitable for runtime linking.


viciious:
A small note: the reason didn't use the vcmi_set_output_dir macro in minizip is because I wanted to avoid introducing the reverse-dependency of minizip on vcmi.

PR #330, opened 2017-07-07
Tweak battle creatures amount box hiding

dydzio0614:
While new implementation is not perfect, it follows OH3 behavior more closely. Should also be more player-friendly.

Box should not be invisible for whole attack + counterattack action. Also does not get displayed when amount is 0. In most cases box number should go invisible before changing now (though the timing when it changes is different from OH3 yet).

### DIFF
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -3275,19 +3275,31 @@ void CBattleInterface::showAliveStacks(SDL_Surface *to, std::vector<const CStack
 {
 	auto isAmountBoxVisible = [&](const CStack *stack) -> bool
 	{
-		if (stack->hasBonusOfType(Bonus::SIEGE_WEAPON)) // siege weapons are always singular
+		if(stack->hasBonusOfType(Bonus::SIEGE_WEAPON)) // siege weapons are always singular
 			return false;
 
-		if (curInt->curAction)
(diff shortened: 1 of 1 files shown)


client/battle/CBattleInterface.cpp
```diff
@@ -3275,19 +3275,31 @@ void CBattleInterface::showAliveStacks(SDL_Surface *to, std::vector<const CStack
 {
 	auto isAmountBoxVisible = [&](const CStack *stack) -> bool
 	{
-		if (stack->hasBonusOfType(Bonus::SIEGE_WEAPON)) // siege weapons are always singular
+		if(stack->hasBonusOfType(Bonus::SIEGE_WEAPON)) // siege weapons are always singular
```
alexvins: this is not exactly true, battle code somewhat support stacked war machines.

dydzio0614: Well, I guess we can change it in future if needed. For example with future scripting support we may try (or let others try) to reimplement some original WoG war machine scripts that handle stacked war machines support.

Or maybe you have any idea what to do with this code now? We can add "&& stack->count ==1" to condition to hide box only when there is 1 war machine in stack.

alexvins: Yes add `&& stack->count ==1`

PR #338, opened 2017-07-13
[WIP, Waiting] Block some hexes on "ship to ship" battlefield

dydzio0614:
TODO:
Change serialization version, backwards compatibility

Waiting for FeniksFire obstacle branch and Alex CStack branch before continuing

### DIFF
--- a/lib/battle/CObstacleInstance.cpp
+++ b/lib/battle/CObstacleInstance.cpp
@@ -88,7 +88,7 @@ bool CObstacleInstance::stopsMovement() const
 
 bool CObstacleInstance::blocksTiles() const
 {
-	return obstacleType == USUAL || obstacleType == ABSOLUTE_OBSTACLE || obstacleType == FORCE_FIELD;
+	return obstacleType == USUAL || obstacleType == ABSOLUTE_OBSTACLE || obstacleType == FORCE_FIELD || obstacleType == BATTLEFIELD_LIMITER;
 }
 
 SpellCreatedObstacle::SpellCreatedObstacle()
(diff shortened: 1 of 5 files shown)


FeniksFire:
As for me it's be nice to remake everything for better concept than now but if someone don't want do it the best thing it's stick to the original concept without mixing things. 
```
//Accessibility is property of hex in battle. It doesn't depend on stack, side's perspective and so on.
enum class EAccessibility
{
...
    UNAVAILABLE, //indestructible wall parts, special battlefields (like boat-to-boat)
...
};
```

lib/battle/CObstacleInstance.cpp
```diff
@@ -142,6 +142,19 @@ void SpellCreatedObstacle::battleTurnPassed()
 		turnsRemaining--;
 }
 
+std::vector<BattleHex> BattlefieldLimiterObstacle::getAffectedTiles() const
+{
+	std::vector<BattleHex> impassableTiles;
```
alexvins: This method does not make sense. You should make BattlefieldLimiterObstacle 1-tile and put your obstacles to positions during battle setup.

lib/registerTypes/RegisterTypes.h
```diff
@@ -201,6 +201,7 @@ void registerTypesMapObjects2(Serializer &s)
 	//s.template registerType<CObstacleInstance>();
 		s.template registerType<CObstacleInstance, MoatObstacle>();
 		s.template registerType<CObstacleInstance, SpellCreatedObstacle>();
+		s.template registerType<CObstacleInstance, BattlefieldLimiterObstacle>();
```
alexvins: this line completely breaks save compatibility.

dydzio0614: Yep - what is the current solution for such things? No register new classes at all?

alexvins:
[Not a part of review] IMO, it is better to alter AccessibilityInfo calculation rather than defining new obstacle type.

dydzio0614:
Yep, I will probably remake this branch from scratch - though I am still interested in answer to how to handle serialization when planning to register new type.

alexvins:
New entity types must be added last (after others, before netpacks).

lib/registerTypes/RegisterTypes.h
alexvins: you should register between registerTypesMapObjectTypes and registerTypesClientPacks1, may add new method for current additions and use it for new types until we decide to break save compatibility.

PR #340, opened 2017-07-13
Add proper handling for ship-to-ship battlefield

dydzio0614:
* When this battlefield type is detected proper hexes are marked as inaccessible
* Do not allow obstacle placing on inaccessible tiles

### DIFF
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -857,9 +857,7 @@ AccessibilityInfo CBattleInfoCallback::getAccesibility() const
 
 	//special battlefields with logically unavailable tiles
 	std::vector<BattleHex> impassableHexes;
-	switch(battleGetBattlefieldType().num)
-	{
-	case BFieldType::SHIP_TO_SHIP:
+	if(battleGetBattlefieldType().num == BFieldType::SHIP_TO_SHIP)
 		impassableHexes = 
 		{ 
 			6, 7, 8, 9,
@@ -872,8 +870,6 @@ AccessibilityInfo CBattleInfoCallback::getAccesibility() const
 			159, 160, 161, 162, 163,
 			176, 177, 178, 179, 180
 		};
-		break;
-	}
 	for(auto hex : impassableHexes)
 		ret[hex] = EAccessibility::UNAVAILABLE;
 
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -860,7 +860,18 @@ AccessibilityInfo CBattleInfoCallback::getAccesibility() const
 	switch(battleGetBattlefieldType().num)
 	{
 	case BFieldType::SHIP_TO_SHIP:
-		impassableHexes = { 6, 7, 8, 9, 24, 25, 26, 58, 59, 60, 75, 76, 77, 92, 93, 94, 109, 110, 111, 126, 127, 128, 159, 160, 161, 162, 163, 176, 177, 178, 179, 180 };
+		impassableHexes = 
+		{ 
+			6, 7, 8, 9,
+			24, 25, 26, 
+			58, 59, 60,
+			75, 76, 77,
+			92, 93, 94,
+			109, 110, 111,
+			126, 127, 128,
+			159, 160, 161, 162, 163,
+			176, 177, 178, 179, 180
+		};
 		break;
 	}
 	for(auto hex : impassableHexes)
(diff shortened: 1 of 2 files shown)


lib/battle/CBattleInfoCallback.cpp
```diff
@@ -855,6 +855,17 @@ AccessibilityInfo CBattleInfoCallback::getAccesibility() const
 		ret[BattleHex(0, y)] = EAccessibility::SIDE_COLUMN;
 	}
 
+	//special battlefields with logically unavailable tiles
+	std::vector<BattleHex> impassableHexes;
+	switch(battleGetBattlefieldType().num)
+	{
+	case BFieldType::SHIP_TO_SHIP:
+		impassableHexes = { 6, 7, 8, 9, 24, 25, 26, 58, 59, 60, 75, 76, 77, 92, 93, 94, 109, 110, 111, 126, 127, 128, 159, 160, 161, 162, 163, 176, 177, 178, 179, 180 };
```
alexvins: too long line, also `{` should be on new line

dydzio0614: so how to fix this?

alexvins: put `{` on new line and split list into several lines

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -855,6 +855,28 @@ AccessibilityInfo CBattleInfoCallback::getAccesibility() const
 		ret[BattleHex(0, y)] = EAccessibility::SIDE_COLUMN;
 	}
 
+	//special battlefields with logically unavailable tiles
+	std::vector<BattleHex> impassableHexes;
+	switch(battleGetBattlefieldType().num)
+	{
```
FeniksFire: This look weird for me, simple if() won't be enough instead of switch?

dydzio0614: It would, but switch will better reflect the purpose in case of adding more battlefields. Will be visible at first glance that whole control statement is about handling different battlefield types.

FeniksFire: So this should be throw into some function that will provide "visibility" for programmer, and then you can change switch to if().

dydzio0614: After all... I think it is pretty unlikely we will need to handle more cases. Will change to "if" soon

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -855,6 +855,24 @@ AccessibilityInfo CBattleInfoCallback::getAccesibility() const
 		ret[BattleHex(0, y)] = EAccessibility::SIDE_COLUMN;
 	}
 
+	//special battlefields with logically unavailable tiles
+	std::vector<BattleHex> impassableHexes;
+	if(battleGetBattlefieldType().num == BFieldType::SHIP_TO_SHIP)
+		impassableHexes = 
+		{
```
ArseniyShestakov: I think we shouldn't allow multi-line code in else-if without brackets. E.g this is bad: 
```
if(a)
	//comment
	dosomething();
```
So I want brackets around:
```
if(a)
{
	var =
	{
		1, 2, 3, 4,
		5, 6, 7, 8
	}
}
```
This is not request for actual changes yet. What everyone else thinks?

alexvins: 1)If there are brackets inside, outside brackets required. 
2) Your first example is questionable.

dydzio0614: What about:
`for(auto group : groups)
	for(auto item : group)
		doSth();`
Or:
if(a)
	if(b)
		doSth();

I got used to skip { } in cases of multi conditional expressions, with exception when there is "else" so it will be "sealed" by curly brackets to which "if" it belongs.

ArseniyShestakov: Yep my example is just bad.

ArseniyShestakov: @dydzio0614
I think this is totally fine to avoid brackets if body is a single-line expression and especially so if it's not really followed by any code because this is look worse:
```
if(a)
	for(auto item : group)
		doSth();
else if
{
	code();
}
else
{
	code();
}
```
I sadly can't find any good example of single, but multi-line expressions in our code. Most of them are just bad code so I would rather not put them as example for anyone.

alexvins:
```
if(a)
	if(b)
		if(c)
			doSth();
```
->
```
if(a)
{
	if(b)
		if(c)
			doSth();
}
```
If whole (inner) contraption is more than 1 line then outer brackets required

PR #341, opened 2017-07-13
Add option to toggle monthly artifact change in black market

dydzio0614:
EDIT: Implemented as "hardcoded feature".
Fixes mantis #2714.


DjWarmonger:
Well, this is  a feature borrowed from H5 (similiar to dwellings accumulate creatures).

ArseniyShestakov:
As Warmonger noted this kind of feature of VCMI.
You might want to instead make it option in "hardcodedFeatures".

ArseniyShestakov:
And then it's need to be added on Wiki too:
https://wiki.vcmi.eu/index.php?title=Game_mechanics

lib/CModHandler.h
```diff
@@ -268,20 +268,22 @@ class DLL_LINKAGE CModHandler
...
 		bool WINNING_HERO_WITH_NO_TROOPS_RETREATS;
+		bool BLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE;
 
 		template <typename Handler> void serialize(Handler &h, const int version)
 		{
 			h & data & CREEP_SIZE & WEEKLY_GROWTH & NEUTRAL_STACK_EXP & MAX_BUILDING_PER_TURN;
 			h & DWELLINGS_ACCUMULATE_CREATURES & ALL_CREATURES_GET_DOUBLE_MONTHS &
 			MAX_HEROES_AVAILABLE_PER_PLAYER & MAX_HEROES_ON_MAP_PER_PLAYER;
 			if(version >= 756)
-			{
```
ArseniyShestakov: If you didn't do that yourself please disable this automatic curly brackets removal. In many cases it's makes sense to avoid or remove them, but serialization compatibility code is not one of those places.

lib/mapObjects/CGMarket.cpp
```diff
@@ -284,13 +285,13 @@ std::vector<int> CGBlackMarket::availableItemsIds(EMarketMode::EMarketMode mode)
...
 void CGBlackMarket::newTurn(CRandomGenerator & rand) const
 {
-	if(cb->getDate(Date::DAY_OF_MONTH) != 1) //new month
-		return;
-
-	SetAvailableArtifacts saa;
-	saa.id = id.getNum();
-	cb->pickAllowedArtsSet(saa.arts, rand);
-	cb->sendAndApply(&saa);
+	if(VLC->modh->settings.BLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE && cb->getDate(Date::DAY_OF_MONTH) == 1) //new month and feature enabled
```
ArseniyShestakov: I would suggest to use dedicated if instead:
```
if(!VLC->modh->settings.BLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE)
	return;
```
So it's easier to understand is some very special hardcoded option and not just one of parameters.

lib/CModHandler.h
dydzio0614: Oh, I removed these on purpose as they stick to my code where I used "no bracket style"

lib/mapObjects/CGMarket.cpp
dydzio0614: Okay, though apart from this pull request - isn't it better practice to use "if" for main code logic and leave the rest for automatic return instead of explicitly quitting from function at beginning?

ArseniyShestakov: I personally don't have any strong opinion on that topic, but I think it's depend on what is your code doing. If it's some kind of rules checking or input validation there nothing wrong with multiple checks instead of large one.

lib/CModHandler.h
ArseniyShestakov: I think we should always use brackets in serialization code, but it's all about it.

dydzio0614: So should I change to brackets?

ArseniyShestakov: Yes. I'll probably have to write something about that on wiki.
I not sure if we have any article about compatibility code yet.

ArseniyShestakov:
Should be merged after #323.

PR #343, opened 2017-07-15
Fix color change of highlighted hex borders

dydzio0614:
Behavior is exactly same as in OH3 now except for behavior non existing in H3 (showing enemy unit range on mouse hover).
Battlefield hex green border color doesn't get darker when highlighted with this change. Visually, hex frames consist of borders of "this pixel and neighbour pixels". In OH3 when highlighting pixel with mouse, border gets thinner as part of it belonging to "current hex" gets replaced by hex highlight graphics. Hexes highlighted not by mouse should have full border width.


dydzio0614:
Now it should work as in OH3 also for showing enemy unit range and casting range spells.

PR #345, opened 2017-07-16
Code style: add empty parameter list to all lambda expressions

ArseniyShestakov:
For great consistency since in other half of codebase all lambda expressions have parameter list.
This obviously also make it easier for Uncrustify to detect lambdas.

Added it to our coding guidelines:
https://wiki.vcmi.eu/index.php?title=Coding_guidelines#Lambda_expressions

### DIFF
--- a/AI/VCAI/VCAI.cpp
+++ b/AI/VCAI/VCAI.cpp
@@ -606,7 +606,7 @@ void VCAI::heroGotLevel(const CGHeroInstance *hero, PrimarySkill::PrimarySkill p
 	LOG_TRACE_PARAMS(logAi, "queryID '%i'", queryID);
 	NET_EVENT_HANDLER;
 	status.addQuery(queryID, boost::str(boost::format("Hero %s got level %d") % hero->name % hero->level));
-	requestActionASAP([=]{ answerQuery(queryID, 0); });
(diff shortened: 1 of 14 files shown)


dydzio0614:
I support this change very much. I did not even know you can define lambdas skipping ()... what a mess... If I remember correctly in c++ (but not in C) it is also legit to init some lists of elements like these: "int a[] {5,3,1};", "std::vector<int> {5,2,1};"

mwu-tow:
For the record — I think that empty arguments list just add visual clutter. I don't think that we should worry that these lambdas look too much as an array variable definition. 

Is there no other feasible way for Uncrustify to disambiguate?

ArseniyShestakov:
There could be some other way to make Uncrustify detect them, but I don't know one.

Also personally I care more about code style consistency and having ```()``` everywhere doesn't sound like a bad thing to me. Like we use empty ```()``` for heap allocations already.

mwu-tow:
Heap allocations have a different meaning for () — it is value intialziation of an allocated variable — and I'd consider missing it to be a style bug (ie leaving variable value not intialized).

As for this — I like more skipping () for lambdas whenever possible but it is just a matter of personal preference (and preference is usually just being accustomed to). If you prefer it other way, then don't hesitate to push forward with this PR.  (same goes for my comment in the other PR)

PR #346, opened 2017-07-16
Code style: use multi-line comments for commented out code

ArseniyShestakov:
I think this should also be enforced in our coding guidelines.

I think it's totally fine to comment single line of code like that:
```
//bool var = something();
```
But two+ lines multi-line comments must be used.

### DIFF
--- a/AI/BattleAI/BattleAI.h
+++ b/AI/BattleAI/BattleAI.h
@@ -79,22 +79,24 @@ public:
 	void print(const std::string &text) const;
 	BattleAction useCatapult(const CStack *stack);
 	void battleStart(const CCreatureSet * army1, const CCreatureSet * army2, int3 tile, const CGHeroInstance * hero1, const CGHeroInstance * hero2, bool Side) override;
-	//void actionFinished(const BattleAction &action) override;//occurs AFTER every action taken by any stack or by the hero
-	//void actionStarted(const BattleAction &action) override;//occurs BEFORE every action taken by any stack or by the hero
-	//void battleAttack(const BattleAttack *ba) override; //called when stack is performing attack
-	//void battleStacksAttacked(const std::vector<BattleStackAttacked> & bsa) override; //called when stack receives damage (after battleAttack())
-	//void battleEnd(const BattleResult *br) override;
-	//void battleResultsApplied() override; //called when all effects of last battle are applied
-	//void battleNewRoundFirst(int round) override; //called at the beginning of each turn before changes are applied;
-	//void battleNewRound(int round) override; //called at the beginning of each turn, round=-1 is the tactic phase, round=0 is the first "normal" turn
-	//void battleStackMoved(const CStack * stack, std::vector<BattleHex> dest, int distance) override;
-	//void battleSpellCast(const BattleSpellCast *sc) override;
-	//void battleStacksEffectsSet(const SetStackEffect & sse) override;//called when a specific effect is set to stacks
-	//void battleTriggerEffect(const BattleTriggerEffect & bte) override;
-	//void battleStart(const CCreatureSet *army1, const CCreatureSet *army2, int3 tile, const CGHeroInstance *hero1, const CGHeroInstance *hero2, bool side) override; //called by engine when battle starts; side=0 - left, side=1 - right
-	//void battleStacksHealedRes(const std::vector<std::pair<ui32, ui32> > & healedStacks, bool lifeDrain, bool tentHeal, si32 lifeDrainFrom) override; //called when stacks are healed / resurrected first element of pair - stack id, second - healed hp
-	//void battleNewStackAppeared(const CStack * stack) override; //not called at the beginning of a battle or by resurrection; called eg. when elemental is summoned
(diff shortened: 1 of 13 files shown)


dydzio0614:
I personally think stuff like
```
//int var1 = something;
//int var2 = something;
```
that consist of not many lines could optionally remain, but if a logical block is involved or larger amount of lines then /* */ would be better, especially in cases when typing multiple // would be tedious job. So to sum up - change is good but bugging programmers about this when margin cases are found may be not worth it.

ArseniyShestakov:
@dydzio0614 this isn't going to be enforced via formatter so it's okay to have some blocks of code commented out. We can always point this out to our new contributors in PR preview.

mwu-tow:
What is the rationale for the change?

I am used to having `//` everywhere, but is is mostly a matter of IDE that just make it easy to comment/uncomment the selected lines. With `/**/` it just becomes more difficult to uncomment only a single line from set plus one needs to be careful of nesting comments. Also, I like when Git Blame for a line shows a commit that disabled the line.

ArseniyShestakov:
@mwu-tow few motivations behind that:

1. Not every IDE let you comment/uncomment them easily.
2. At least in Qt Creator it's much easier to just toggle multi-line comment and there no such feature for multi-line one started with //. This is same in almost any editor I know.
3. I really wish that // were used mainly for comments.

DjWarmonger:
I agree with Tow (probably because we both use MVS :P). Many tools have automatic // insertion and it's impossible to enforce other style of coding.

BTW,  I'm observing that you focus too much on code formatting and too little on actual features that any player could care about.

AI/BattleAI/BattleAI.h
```diff
@@ -79,22 +79,24 @@ class CBattleAI : public CBattleGameInterface
...
+	void battleTriggerEffect(const BattleTriggerEffect & bte) override;
+	void battleStart(const CCreatureSet *army1, const CCreatureSet *army2, int3 tile, const CGHeroInstance *hero1, const CGHeroInstance *hero2, bool side) override; //called by engine when battle starts; side=0 - left, side=1 - right
+	void battleStacksHealedRes(const std::vector<std::pair<ui32, ui32> > & healedStacks, bool lifeDrain, bool tentHeal, si32 lifeDrainFrom) override; //called when stacks are healed / resurrected first element of pair - stack id, second - healed hp
+	void battleNewStackAppeared(const CStack * stack) override; //not called at the beginning of a battle or by resurrection; called eg. when elemental is summoned
+	void battleObstaclesRemoved(const std::set<si32> & removedObstacles) override; //called when a certain set  of obstacles is removed from batlefield; IDs of them are given
+	void battleCatapultAttacked(const CatapultAttack & ca) override; //called when catapult makes an attack
+	void battleStacksRemoved(const BattleStacksRemoved & bsr) override; //called when certain stack is completely removed from battlefield
+*/
```
DjWarmonger: Now, assuming this code will ever be used, how are you going to uncomment only ONE function in the middle? It's multiplying the work.

dydzio0614: double/triple click line to select, cut, paste? :P
I think keeping original order of functions doesnt matter

ArseniyShestakov:
@DjWarmonger I temporary focused on formatting for one reason: right now we have 6-7 developers who more or less actively contribute. I really wish we had more developers and there actually are interested people who might start to contribute eventually. 

In this situation it's make sense to at least ask everyone to properly format new code and for long time both me and @alexvins simply reviewed all PRs asking to fix thing or two every single time, but there are problems:

1. There is no standard we all agreed on. So we can't even sent people on wiki because it's outdated.
2. VCMI codebase is full of inconsistent formatting. So it's easy to add more of badly formatted code.
3. When every PR contain formatting changes on top of code changes it's make it harder to review.

So it's annoying and waste of time for both developer reviewing changes and one who have to waste time on fixing formatting when it's something automated tool should do. This is why I want to offload formatting verification on CI. Then when commit or PR contain changes that break formatting standard we know that in instant.

As bonus we'll get updated version of our coding guidelines on wiki so we can link it to all new contributors. Also Dydzio already have good formatting configuration for both MSVS and ReSharper and I'll provide one for Qt Creator and Alex I guess can give us one for Code::Blocks. So when we finally decide on everything we'll just share them any no one would need to configure it on their own.

DjWarmonger:
Well yes, this certainly makes more sense to me.

I just would like to point out that formatting with `/*` or `//` is irrelevant. C++ allows both for different purposes.

ArseniyShestakov:
@DjWarmonger that is fine if you don't like this change since listening everyones opinion is reason why it's all in pull requests. It's important to make sure everyone aware of suggested changes and have chance to contribute their vision before we enforce them.

mwu-tow:
I believe that QtCreator also uses `//` for commenting/uncomementing by default. At least it is what happened to me when I pressed Ctrl+/ while having a few lines selected. What can be more easy than using default IDE settings? 

Side note — having huge blocks of commented-out code is not a good thing anyway. I think that practically all of the code in this PR should be either removed or receive an additional comment stating why it is relevant today. For history we have git.

ArseniyShestakov:
Thanks for pointing out!

I never actually needed to comment huge blocks of code so I didn't knew about Ctrl + /.

PR #347, opened 2017-07-16
[WIP] Code style: preview of Uncrustify formatting

ArseniyShestakov:
This is preview only! 

There some things I personally still want to talk about:

- Switching to no space after * and &.
- Namespace indention. Some parts of our code look especially ugly when it's disabled.

Roadblocks:

- [ ] Problem with formatting of ```**``` and ```***``` @FeniksFire pointed to:
```
tiles = new CTileInfo * *[width];
int main(int argc, char * * argv)
```
- [x] Problem with detection of lambdas with omitted parameters list. Should be fixed by #345.
- [ ] Indention of lambda expressions. I hack-fixed them with "indent_continue = -1", but that not a solution.
- [ ] Multi-line if-else conditions. I'll talk about them in comments.
- [ ] Preprocessor directives formatting. I didn't look a them at all.
- [ ] Changes in formatting multi-line commented-out code. I don't want Uncrustify to touch it.
- [ ] And really multi-line everything since "indent_continue = -1" tend to cause weird results.

What would need to be fixed manually after initial formatting:

- In switch break should be moved inside curly brackets. Can be done with Uncrustify option "mod_move_case_break", but It's better to be enabled only once so it's not break something in future.
- Some tabs will need to be replaced with spaces. Formatter doesn't replace tabs with spaces when it's for instance move colons in class initialization.

Please comment on code changes! Then when it's updated we won't have too many outdated comments.
I'll force push this branch every time with two commits: uncrustify config file and preview.


ArseniyShestakov:
Before comment don't forget to check out our coding guidelines:
https://wiki.vcmi.eu/index.php?title=Coding_guidelines
They partially outdated, but we really need to keep them up to date now. 

Feel free to leave any feedback on what you like or dislike about formatted code or if something is missing.

client/gui/CGuiHandler.h
```diff
@@ -48,33 +49,34 @@ class CGuiHandler
...
-				   rclickable,
-				   mclickable,
-				   hoverable,
-				   keyinterested,
-				   motioninterested,
-	               timeinterested,
-	               wheelInterested,
-	               doubleClickInterested,
-	               textInterested;
+	rclickable,
```
mwu-tow: I'd say that when declaring multiple names, they should be indented, so it is clear that they belong to the type given in the initial declaration line.

client/gui/Fonts.cpp
```diff
@@ -352,27 +355,25 @@ void CBitmapHanFont::renderText(SDL_Surface * surface, const std::string & data,
...
 	}
 	SDL_UnlockSurface(surface);
 }
 
-CBitmapHanFont::CBitmapHanFont(const JsonNode &config):
-    fallback(new CBitmapFont(config["fallback"].String())),
-    data(CResourceHandler::get()->load(ResourceID("data/" + config["name"].String(), EResType::OTHER))->readAll()),
-    size(config["size"].Float())
+CBitmapHanFont::CBitmapHanFont(const JsonNode & config)
+	: fallback(new CBitmapFont(config["fallback"].String())), data(CResourceHandler::get()->load(ResourceID("data/" + config["name"].String(), EResType::OTHER))->readAll()), size(config["size"].Float())
```
mwu-tow: I'm not sure if it is an issue we want to tackle right now, but such long lines are hard to visually parse.
Would it be possible to have a rule like "if initialization list fits within 120 chars then keep it single line, otherwise break it down to single member initializer per line.

Similar sentiment also applies to signatures of function taking many arguments and their call sites.

client/windows/CHeroWindow.cpp
```diff
@@ -103,33 +102,37 @@ CHeroWindow::CHeroWindow(const CGHeroInstance *hero):
 	//artifs = new CArtifactsOfHero(pos.topLeft(), true);
 	ourBar = new CGStatusBar(7, 559, "ADROLLVR.bmp", 660); // new CStatusBar(pos.x+72, pos.y+567, "ADROLLVR.bmp", 660);
 
-	quitButton = new CButton(Point(609, 516), "hsbtns.def", CButton::tooltip(heroscrn[17]), [&]{ close(); }, SDLK_RETURN);
+	quitButton = new CButton(Point(609, 516), "hsbtns.def", CButton::tooltip(heroscrn[17]), [&] { close();
+	}, SDLK_RETURN);
```
mwu-tow: Such lambda formatting feels wrong. Either the code block should be fully expanded or everything should be kept inline.

client/windows/CHeroWindow.cpp
```diff
@@ -300,8 +310,12 @@ void CHeroWindow::update(const CGHeroInstance * hero, bool redrawNeeded)
 
 void CHeroWindow::dismissCurrent()
 {
-	CFunctionList<void()> ony = [=] { close(); };
-	ony += [=] { LOCPLINT->cb->dismissHero(curHero); };
+	CFunctionList<void()> ony = [ = ] {
```
mwu-tow: why did default capture specifier get spaces around it? Elsewhere I see `[=]`

test/StdInc.h
```diff
@@ -10,7 +10,5 @@
 	#define BOOST_TEST_MODULE VcmiTest
 	#include <boost/test/unit_test.hpp>
 	#include "CVcmiTestConfig.h"
-	BOOST_GLOBAL_FIXTURE(CVcmiTestConfig);
+BOOST_GLOBAL_FIXTURE(CVcmiTestConfig);
```
mwu-tow: I think the intention was to keep whole conditionally compiled block indented.

AI/StupidAI/StupidAI.cpp
```diff
@@ -94,20 +96,23 @@ bool isCloser(const EnemyInfo & ei1, const EnemyInfo & ei2, const ReachabilityIn
 
 }
 
-static bool willSecondHexBlockMoreEnemyShooters(const BattleHex &h1, const BattleHex &h2)
+static bool willSecondHexBlockMoreEnemyShooters(const BattleHex & h1, const BattleHex & h2)
 {
-	int shooters[2] = {0}; //count of shooters on hexes
+	int shooters[2] =
+	{
+		0
```
dydzio0614: Hmm I am not a fan of expanding simple initializers like that

AI/BattleAI/AttackPossibility.cpp
```diff
@@ -24,21 +24,24 @@ int AttackPossibility::attackValue() const
...
 
 	const int remainingCounterAttacks = getValOr(state.counterAttacksLeft, enemy, enemy->counterAttacks.available());
 	const bool counterAttacksBlocked = attacker->hasBonusOfType(Bonus::BLOCKS_RETALIATION) || enemy->hasBonusOfType(Bonus::NO_RETALIATION);
-	const int totalAttacks = 1 + AttackInfo.attackerBonuses->getBonuses(Selector::type(Bonus::ADDITIONAL_ATTACK), (Selector::effectRange (Bonus::NO_LIMIT).Or(Selector::effectRange(Bonus::ONLY_MELEE_FIGHT))))->totalValue();
+	const int totalAttacks = 1 + AttackInfo.attackerBonuses->getBonuses(Selector::type(Bonus::ADDITIONAL_ATTACK), (Selector::effectRange(Bonus::NO_LIMIT).Or(Selector::effectRange(Bonus::ONLY_MELEE_FIGHT))))->totalValue();
 
-	AttackPossibility ap = {enemy, hex, AttackInfo, 0, 0, 0};
+	AttackPossibility ap =
+	{
+		enemy, hex, AttackInfo, 0, 0, 0
```
dydzio0614: Also here - not sure if "expand" change like this is actually improving code quality at all

server/NetPacksServer.cpp
```diff
@@ -22,145 +22,145 @@
...
-
-#define WRONG_PLAYER_MSG(expectedplayer) do {std::ostringstream oss;\
-			oss << "You were identified as player " << gh->getPlayerAt(c) << " while expecting " << expectedplayer;\
-			logNetwork->error(oss.str()); \
-			if(c) { SystemMessage temp_message(oss.str()); boost::unique_lock<boost::mutex> lock(*c->wmx); *c << &temp_message; } } while(0)
-
-#define ERROR_IF_NOT_OWNS(id)	do{if(!PLAYER_OWNS(id)){WRONG_PLAYER_MSG(gh->getOwner(id)); ERROR_AND_RETURN; }}while(0)
-#define ERROR_IF_NOT(player)	do{if(player != gh->getPlayerAt(c)){WRONG_PLAYER_MSG(player); ERROR_AND_RETURN; }}while(0)
+#define PLAYER_OWNS(id) (gh->getPlayerAt(c) == gh->getOwner(id))
+#define ERROR_AND_RETURN                                                                                                \
```
mwu-tow: It would be great if uncrustify could format macro body that:
a) use rules as for formatting code
b) align trailing `\` on the same column

lib/mapping/CMap.h
```diff
@@ -90,12 +89,9 @@ struct DLL_LINKAGE PlayerInfo
...
 	ui8 powerPlaceholders;
 
-	template <typename Handler>
-	void serialize(Handler & h, const int version)
+	template<typename Handler> void serialize(Handler & h, const int version)
 	{
-		h & p7 & hasRandomHero & mainCustomHeroId & canHumanPlay & canComputerPlay & aiTactic & allowedFactions & isFactionRandom &
-				mainCustomHeroPortrait & mainCustomHeroName & heroesNames & hasMainTown & generateHeroAtMainTown &
-				posOfMainTown & team & generateHero;
+		h & p7 & hasRandomHero & mainCustomHeroId & canHumanPlay & canComputerPlay & aiTactic & allowedFactions & isFactionRandom & mainCustomHeroPortrait & mainCustomHeroName & heroesNames & hasMainTown & generateHeroAtMainTown & posOfMainTown & team & generateHero;
```
mwu-tow: Another case of "it would be great if overlong lines could be split".

AI/StupidAI/StupidAI.cpp
```diff
@@ -49,7 +49,8 @@ struct EnemyInfo
 	const CStack * s;
 	int adi, adr;
 	std::vector<BattleHex> attackFrom; //for melee fight
-	EnemyInfo(const CStack * _s) : s(_s), adi(0), adr(0)
+	EnemyInfo(const CStack * _s)
+		: s(_s), adi(0), adr(0)
```
dydzio0614: Any advantages of this formatting change except saving some space for really many parameters? For me it's little bit easier to track code when { } belong to something that is written in one line above {, but maybe I freak too much because I saw too many ugly multi-line if's and elses

AI/StupidAI/StupidAI.cpp
```diff
@@ -117,7 +122,10 @@ BattleAction CStupidAI::activeStack( const CStack * stack )
 	if(stack->type->idNumber == CreatureID::CATAPULT)
 	{
 		BattleAction attack;
-		static const std::vector<int> wallHexes = {50, 183, 182, 130, 78, 29, 12, 95};
+		static const std::vector<int> wallHexes =
+		{
+			50, 183, 182, 130, 78, 29, 12, 95
```
dydzio0614: So here we have bit longer "list initializer"... When looking at it I do not find it to be particularly easier to read, and if the benefit is saving space then it will not help for really long value lists or when they are not short numbers.

AI/VCAI/VCAI.cpp
```diff
@@ -3325,30 +3338,30 @@ this functions returns one target tile or invalid tile. We will use it to poll p
...
+	int3 ret(-1, -1, -1);
 
 	int sourceSector = retreiveTile(h->visitablePos()),
-		destinationSector = retreiveTile(dst);
+	destinationSector = retreiveTile(dst);
 
-	const Sector *src = &infoOnSectors[sourceSector],
-		*dest = &infoOnSectors[destinationSector];
+	const Sector * src = &infoOnSectors[sourceSector],
+	* dest = &infoOnSectors[destinationSector];
```
dydzio0614: Ugh, is this "* dest" thing intended?

CCallback.h
```diff
@@ -34,61 +34,61 @@ class IBattleCallback
...
-
-	virtual void trade(const CGObjectInstance *market, EMarketMode::EMarketMode mode, int id1, int id2, int val1, const CGHeroInstance *hero = nullptr)=0; //mode==0: sell val1 units of id1 resource for id2 resiurce
-
-	virtual int selectionMade(int selection, QueryID queryID) =0;
-	virtual int sendQueryReply(const JsonNode & reply, QueryID queryID) =0;
-	virtual int swapCreatures(const CArmedInstance *s1, const CArmedInstance *s2, SlotID p1, SlotID p2)=0;//swaps creatures between two possibly different garrisons // TODO: AI-unsafe code - fix it!
-	virtual int mergeStacks(const CArmedInstance *s1, const CArmedInstance *s2, SlotID p1, SlotID p2)=0;//joins first stack to the second (creatures must be same type)
-	virtual int mergeOrSwapStacks(const CArmedInstance *s1, const CArmedInstance *s2, SlotID p1, SlotID p2) =0; //first goes to the second
-	virtual int splitStack(const CArmedInstance *s1, const CArmedInstance *s2, SlotID p1, SlotID p2, int val)=0;//split creatures from the first stack
+	virtual void recruitHero(const CGObjectInstance * townOrTavern, const CGHeroInstance * hero) = 0;
```
dydzio0614: Maybe it makes sense to debate if leaving pure virtual function "0's" without spacebars could be useful for marking this particular "0" usage and potentially faster code understanding. Not worth it though in my opinion.

Global.h
```diff
@@ -279,442 +279,397 @@ void inline handleException()
...
+}
 
-	//checks if a is within b and c
-	template <typename t1, typename t2, typename t3>
-	bool iswithin(const t1 &value, const t2 &min, const t3 &max)
-	{
-		return value >= min && value <= max;
-	}
+//checks if a is between b and c
+template<typename t1, typename t2, typename t3>	bool isbetween(const t1 & value, const t2 & min, const t3 & max)
```
dydzio0614: One more example of template "one lining" - this is not helpful at all imo. Moreover we go for two-liners with constructor's initializer lists etc so why opposite approach.

Global.h
```diff
@@ -279,442 +279,397 @@ void inline handleException()
...
+}
+template<typename T, typename Arg1, typename Arg2> std::unique_ptr<T> make_unique(Arg1 && arg1, Arg2 && arg2)
+{
+	return std::unique_ptr<T>(new T(std::forward<Arg1>(arg1), std::forward<Arg2>(arg2)));
+}
+template<typename T, typename Arg1, typename Arg2, typename Arg3> std::unique_ptr<T> make_unique(Arg1 && arg1, Arg2 && arg2, Arg3 && arg3)
+{
+	return std::unique_ptr<T>(new T(std::forward<Arg1>(arg1), std::forward<Arg2>(arg2), std::forward<Arg3>(arg3)));
+}
+template<typename T, typename Arg1, typename Arg2, typename Arg3, typename Arg4> std::unique_ptr<T> make_unique(Arg1 && arg1, Arg2 && arg2, Arg3 && arg3, Arg4 && arg4)
```
dydzio0614: I definitely dont want to write code like this huge line

client/CBitmapHandler.cpp
```diff
@@ -18,17 +18,17 @@
...
 {
-	SDL_Surface * loadH3PCX(ui8 * data, size_t size);
+SDL_Surface * loadH3PCX(ui8 * data, size_t size);
 
-	SDL_Surface * loadBitmapFromDir(std::string path, std::string fname, bool setKey=true);
+SDL_Surface * loadBitmapFromDir(std::string path, std::string fname, bool setKey = true);
 }
 
-bool isPCX(const ui8 *header)//check whether file can be PCX according to header
+bool isPCX(const ui8 * header) //check whether file can be PCX according to header
```
dydzio0614: Spacebar before comment - interesting change, i think code looks bit more tidy with it.

client/CBitmapHandler.cpp
```diff
@@ -42,33 +42,36 @@ SDL_Surface * BitmapHandler::loadH3PCX(ui8 * pcx, size_t size)
 	SDL_Surface * ret;
 
 	Epcxformat format;
-	int it=0;
-
-	ui32 fSize = read_le_u32(pcx + it); it+=4;
```
dydzio0614: What is current opinion on "never keep multiple statements in one line?" I wonder if these "multiple statements" dont actually improve code readability when used smartly.

client/CBitmapHandler.cpp
```diff
@@ -109,43 +112,43 @@ SDL_Surface * BitmapHandler::loadBitmapFromDir(std::string path, std::string fna
...
 
-	SDL_Surface * ret=nullptr;
+	SDL_Surface * ret = nullptr;
 
 	auto readFile = CResourceHandler::get()->load(ResourceID(path + fname, EResType::IMAGE))->readAll();
 
-	if (isPCX(readFile.first.get()))
-	{//H3-style PCX
+	if(isPCX(readFile.first.get()))
+	{ //H3-style PCX
```
dydzio0614: Is it decided yet where are we allowed to place comments? For example Microsoft's C# guidelines discourage using comments at the end of lines with code, in favor of comments one line above code. I do not that too much though.

client/CBitmapHandler.cpp
```diff
@@ -109,43 +112,43 @@ SDL_Surface * BitmapHandler::loadBitmapFromDir(std::string path, std::string fna
...
 	}
 	else
 	{ //loading via SDL_Image
 		ret = IMG_Load_RW(
-		          //create SDL_RW with our data (will be deleted by SDL)
-		          SDL_RWFromConstMem((void*)readFile.first.get(), readFile.second),
-		          1); // mark it for auto-deleting
-		if (ret)
+			//create SDL_RW with our data (will be deleted by SDL)
+			SDL_RWFromConstMem((void *)readFile.first.get(), readFile.second),
```
dydzio0614: Not sure what to think about spacebar in casting, Is this side-effect work of uncrustify?

client/CMessage.h
```diff
@@ -21,10 +21,10 @@ class CMessage
 {
 public:
 	//Function usd only in CMessage.cpp
-	static std::pair<int,int> getMaxSizes(std::vector<std::vector<SDL_Surface*> > * txtg, int fontHeight);
+	static std::pair<int, int> getMaxSizes(std::vector<std::vector<SDL_Surface *>> * txtg, int fontHeight);
```
dydzio0614: I think I reported this case already, but this change looks really dumb imo with that pointer-spacebar mangling.

client/battle/CBattleInterfaceClasses.h
```diff
@@ -51,9 +52,10 @@ class CBattleConsole : public CIntObject
 class CBattleHero : public CIntObject
 {
 	void switchToNextPhase();
+
 public:
 	bool flip; //false if it's attacking hero, true otherwise
-	CDefHandler *dh, *flag; //animation and flag
+	CDefHandler * dh, * flag; //animation and flag
```
dydzio0614: A bit unintuitive pointer init formatting

dydzio0614 (commented):
My impressions on various changes after formatting...

dydzio0614:
Now comment that actually answers SXX questions:

*Switching to no space after * and &.

-I think I will stay neutral here after all. That means, I think no space formatting makes sense as it is more common style and visually differentiates pointers from multiply operators in code (int c = (int)a * b; for example). I do not want to get in the way of more productive VCMI developers who like existing style more as I know I can adapt. This may fix some weird stuff from coverity like std::vector<Cstack *> - who needs that spacebar inside template parameters ❓ 

*Namespace indention. Some parts of our code look especially ugly when it's disabled.

-No personal opinion here, though what Tow said on slack sounds kinda good "at first glance" - no indenting for huge namespaces to make code within it feel naturally formatted. Unfortunately coverity is "all or nothing" tool.

client/gui/CGuiHandler.h
ArseniyShestakov: This not really intended. It's problem caused by "indent_continue = -1". But other than that we talked with @alexvins and I come to conclusion it's better if we won't declare using this comma syntax.

Basically it's would be better if they were declared on their own lines.
```
	CIntObjectList lclickable;
	CIntObjectList rclickable;
	CIntObjectList mclickable;
	CIntObjectList hoverable;
	CIntObjectList keyinterested;
	CIntObjectList motioninterested;
	CIntObjectList timeinterested;
	CIntObjectList wheelInterested;
	CIntObjectList doubleClickInterested;
	CIntObjectList textInterested;
```
It's of course will be more code, but it's not as error-prone as declaration with comma especially when it's come to working with pointers.

client/windows/CHeroWindow.cpp
ArseniyShestakov: @mwu-tow this will be fixed once #345 merged. Sorry that I didn't mention it earlier.
Basically Uncrustify have a hard time to parse lambdas with omitted parameter list.

lib/mapping/CMap.h
ArseniyShestakov: Thanks for pointing out. I think for the sake of desync debugging we should really put all of serializer statements on own lines:
```
h & p7;
h & hasRandomHero;
h & mainCustomHeroId;
...
```
Probably should submit PR for that.

AI/BattleAI/BattleAI.cpp
```diff
@@ -53,7 +53,7 @@ BattleAction CBattleAI::activeStack( const CStack * stack )
 		if(stack->hasBonusOfType(Bonus::SIEGE_WEAPON) && stack->hasBonusOfType(Bonus::HEALER))
 		{
 			auto healingTargets = cb->battleGetStacks(CBattleInfoEssentials::ONLY_MINE);
-			std::map<int, const CStack*> woundHpToStack;
+			std::map<int, const CStack *> woundHpToStack;
```
ArseniyShestakov: This is totally intended and I don't see anything wrong with that since ```*>``` on it's own have no special meaning in C++ as far as I aware. Space after the ptr star is questionable, but we'll have space before it for sure.

client/windows/CHeroWindow.cpp
ArseniyShestakov: Bug as well and depend on #345.

AI/BattleAI/common.h
```diff
@@ -11,8 +11,7 @@
 
 class CBattleCallback;
 
-template<typename Key, typename Val, typename Val2>
-const Val getValOr(const std::map<Key, Val> &Map, const Key &key, const Val2 defaultValue)
+template<typename Key, typename Val, typename Val2> const Val getValOr(const std::map<Key, Val> & Map, const Key & key, const Val2 defaultValue)
```
ArseniyShestakov: Depend on what other think we might not change it at all or force new line instead.
Problem is that half of our code use new line and other half is doesn't.

This is Uncrustify option:
```nl_template_class               = remove   # ignore/add/remove/force```

File: AI/BattleAI/BattleAI.cpp [hunk 24]
```diff
@@ -286,7 +301,7 @@ void CBattleAI::attemptCastingSpell()
 
 	for(PossibleSpellcast & psc : possibleCasts)
 		psc.value = evaluateSpellcast(psc);
-	auto pscValue = [] (const PossibleSpellcast &ps) -> int
+	auto pscValue = [](const PossibleSpellcast & ps) -> int
```
ArseniyShestakov: I don't think Uncrustify provide any control over this, at least I didn't manage to find it.

CCallback.h
ArseniyShestakov: Again I don't think I can control this with options.
I'll make a list of thing I supposedly couldn't change via configuration in PR description.

AI/BattleAI/BattleAI.cpp
dydzio0614: Just in case - resharper allows it if needed.

FeniksFire:
I can't find it anywhere so, what will be the format of `int **a`? This will not be something like `int * * a`?

ArseniyShestakov:
@FeniksFire yep it's a mess. Certainly the problem. Add it to roadblocks.

dydzio0614:
I think target syntax should be int ** a; if we stay with current spacebar rules.
But anyway, using double pointers at this point will most likemy be some code smell... we do not write C and C++ has its own syntax for many different stuff.

server/CVCMIServer.cpp
```diff
@@ -607,11 +608,11 @@ void handleLinuxSignal(int sig)
 }
 #endif
 
-int main(int argc, char** argv)
+int main(int argc, char * * argv)
```
ArseniyShestakov: Let's comment on problem with ```***``` ```**``` here.

ArseniyShestakov: @dydzio0614 for something like ```***``` we should probably use boost::multi_array, but we really only have 13 matches for ``` * * ``` now.

@FeniksFire do you think we should still care about them?

dydzio0614:
There is option that if we decide to unify some stuff uncrustify cant handle, we might use resharper and then decide what to do with these cases for future maintenance etc. Even if we dont find automaton solution it may be better to unify some stuff (like spacebars around lambda's ->) than leave "as is" because "uncrustify cannot handle"

server/CVCMIServer.cpp
dydzio0614: To think that problem would be gone if we swapped to "no spacebar" style :D

FeniksFire: Ehh, well sometimes it's nice to have double pointer, but if we have only 13 matches for them, then we can change it manually. In any case we should think about such things :).

ArseniyShestakov:
@dydzio0614 we can only use one solution for formatting because it's enough burden to maintain one. If Resharper will ever be useful on CI as command-line tool we can switch to it since I don't really care if utility is proprietary as long as it's get job done.

Uncrustify provide option to "ignore" some constructions that it doesn't handle well, but it's not always possible because some things like pure virtual functions for instance likely since they tied assignment in general.

dydzio0614:
@SXX I mean to use resharper one time to unify some additional formatting stuff that we will not care about later because our standard tools will not be able check them. So we will end up with more consistent codebase for existing code. Lack of checks for these "stuff" for new code is another topic.

server/CVCMIServer.cpp
ArseniyShestakov: @FeniksFire yeah we certainly should keep it in mind and this is why this PR is here.
It's simply to hard to keep it all in mind and check it all for just one person.

Might be you can find some other weird bits of code here.

client/windows/CSpellWindow.cpp
```diff
@@ -449,9 +444,13 @@ void CSpellWindow::keyPressed(const SDL_KeyboardEvent & key)
...
 			bool down = key.keysym.sym == SDLK_DOWN;
-			static const int schoolsOrder[] = { 0, 3, 1, 2, 4 };
+			static const int schoolsOrder[] =
+			{
+				0, 3, 1, 2, 4
+			};
 			int index = -1;
-			while(schoolsOrder[++index] != selectedTab);
+			while(schoolsOrder[++index] != selectedTab)
+				;
```
ArseniyShestakov: This is something uncrustify doesn't handle well, but this probably need to be rewritten anyway.

AI/StupidAI/StupidAI.cpp
IvanSavenko: For "really may parameters" one member per line would look much better. Extra bonus - keep commas before member name, e.g.
```
EnemyInfo(const CStack * _s)
: s(_s)
, adi(0)
, adr(0)
```
Or even add vertical alignment (although I doubt that auto-formatters can do this)
```
EnemyInfo (const CStack * _s)
: s       (_s)
, adi     (0)
, adr     (0)
```
Such format helps a lot in readability for classes with 10+ members with names of different length

server/CVCMIServer.cpp
mwu-tow: In this particular case — `argv` type should be actually `char *argv[]`.

client/windows/CSpellWindow.cpp
mwu-tow: Actually I like it that way — it is more explicit that there is an intended empty loop;

AI/StupidAI/StupidAI.cpp
dydzio0614: I think we are limited to "binary" solution with uncrustify - either "new line" everything, even for 1,2,3 members or one line 10+ members...

client/CBitmapHandler.cpp
IvanSavenko: IMO this indicates different problem - `it+=4` should be part of `read_le_u32` - e.g. wrapped in some memory buffer class instead of manually advancing pointer for correct length each time you call this method.

mwu-tow:
To answer the questions from the OP:
* I am supportive of "no space after * and &" change
* if we are to either have all namespaces indented or not indented, then I'd say "indent" (though it is depending on the fact that we have few namespaces, and hardly any nested ones)

AI/StupidAI/StupidAI.cpp
IvanSavenko: New line for everything sounds like a good option for me. And normally you don't need a struct with one member so you always end up with 2+ members

AI/VCAI/VCAI.cpp
dydzio0614: Nevermind, it thought there is ; at end of 1st line

client/CPreGame.cpp
```diff
@@ -181,77 +181,77 @@ void updateStartInfo(std::string filename, StartInfo & sInfo, const std::unique_
...
+	void applyOnPG(CSelectionScreen * selScr, void * pack) const override
 	{
-		T *ptr = static_cast<T*>(pack);
+		T * ptr = static_cast<T *>(pack);
 		ptr->apply(selScr);
 	}
 };
 
-template <> class CApplyOnPG<CPack> : public CBaseForPGApply
+template<> class CApplyOnPG<CPack>: public CBaseForPGApply
```
alexvins: Unwanted change. There should be spaces around `:` even after `>`

client/CVideoHandler.cpp
```diff
@@ -185,48 +185,62 @@ bool CVideoPlayer::open(std::string fname, bool loop, bool useOverlay, bool scal
...
-	if (texture)
+	if(texture)
 	{ // Convert the image into YUV format that SDL uses
 		sws = sws_getContext(codecContext->width, codecContext->height, codecContext->pix_fmt,
-							 pos.w, pos.h,
-							 AV_PIX_FMT_YUV420P,
-							 SWS_BICUBIC, nullptr, nullptr, nullptr);
+		pos.w, pos.h,
+		AV_PIX_FMT_YUV420P,
+		SWS_BICUBIC, nullptr, nullptr, nullptr);
```
IvanSavenko: Why indention has been removed here? Now you don't see that these 3 lines are actually parameters from previous method.

AI/VCAI/VCAI.cpp
ArseniyShestakov: Just in case this one is inteded because there is no difference for uncrustify config between:
```
const Sector * src = &infoOnSectors[sourceSector],
* dest = &infoOnSectors[destinationSector];
```
Or:
```
const Sector * src = &infoOnSectors[sourceSector];
const Sector * dest = &infoOnSectors[destinationSector];
```

client/CPreGame.cpp
```diff
@@ -1325,7 +1356,8 @@ SelectionTab::SelectionTab(CMenuScreen::EState Type, const std::function<void(CM
 		select(0);
 		break;
 	case CMenuScreen::loadGame:
-	case CMenuScreen::saveGame:;
+	case CMenuScreen::saveGame:
+		;
```
alexvins: to be removed manually

client/CBitmapHandler.cpp
dydzio0614: This is example of general case when sometimes multi statements may have their use, but yep...
I guess i am getting too paranoid about theoretical cases while our codebase has almost 0 legit usage of "multi statements in one line"

client/CBitmapHandler.cpp
ArseniyShestakov: With config we have now this is intended, but I not sure if this possible to adjust space before the pointer just for casting.

client/battle/CBattleInterfaceClasses.h
IvanSavenko: Don't declare multiple members or variables in a single line and this won't be a problem.

dydzio0614: @ArseniyShestakov so maybe we should add some rules to coding guidelines about multi elements intantiation with comma?

client/CVideoHandler.cpp
ArseniyShestakov: This is not really intended and there must be single tab indention. Probably I should really leave all lambdas completely broken for now because  "indent_continue = -1" confuse everyone.

client/battle/CCreatureAnimation.cpp
```diff
@@ -425,28 +427,22 @@ inline void CCreatureAnimation::putPixel(ui8 * dest, const SDL_Color & color, si
...
-	    || getType() == CCreatureAnim::MOUSEON;
+	return getType() == CCreatureAnim::HOLDING || getType() == CCreatureAnim::MOUSEON;
 }
 
 bool CCreatureAnimation::isMoving() const
 {
-	return getType() == CCreatureAnim::MOVE_START
-	    || getType() == CCreatureAnim::MOVING
-	    || getType() == CCreatureAnim::MOVE_END;
+	return getType() == CCreatureAnim::MOVE_START || getType() == CCreatureAnim::MOVING || getType() == CCreatureAnim::MOVE_END;
```
IvanSavenko: Why this was merged into a single line?
Not sure how common large ` A || B || C && D` statements in code but having each element on a separate line usually looks better.

AI/StupidAI/StupidAI.cpp
ArseniyShestakov: Yes. We can also don't touch it at all.

@IvanSavenko I'm really against any alignment because it's become complete mess as soon as you use different IDE settings from author of the code. And even if formatter can do some alignment it's will be mess for one developer or another.

AI/BattleAI/common.h
dydzio0614: I would say... always use new line. Will be easier to parse in head then as well.

alexvins (dismissed):

AI/StupidAI/StupidAI.cpp
dydzio0614: https://stackoverflow.com/questions/3171540/long-constructor-initialization-lists

I guess new line for every parameter is fair enough.

AI/StupidAI/StupidAI.cpp
dydzio0614: @alexvins because if you initialize less elements than exist, rest is zero initialized. This syntax can initialize array of any size with zeros.

client/gui/Fonts.cpp
dydzio0614: One more reason why it may be good to have one element of initializer list per line.

client/battle/CCreatureAnimation.cpp
ArseniyShestakov: Basically for the reason mentioned in PR description: there is serious issue that Uncrustify try to indent multi-line conditions and align them with spaces. This is controlled following options:
```
# The position of boolean operators in wrapped expressions
pos_bool                        = join     # ignore/join/lead/lead_break/lead_force/trail/trail_break/trail_force
```

Other than that almost all VCMI code that using multi-line condition is just bad or totally horrible: especially CGameHandler with complain and comments inside if-else conditions. It's so bad so in some cases something like that is preferable:
```
bool boolOptionNameOne = somelongcodition == somelongtype;
bool boolOptionNameTwo = somelongcodition == somelongtype && 9000 == 9000;
if(boolOptionNameOne && boolOptionNameTwo)
```
To what we have how:
```
			//ranged counterattack
			if (destinationStack->hasBonusOfType(Bonus::RANGED_RETALIATION)
				&& !stack->hasBonusOfType(Bonus::BLOCKS_RANGED_RETALIATION)
				&& destinationStack->ableToRetaliate()
				&& stack->alive()) //attacker may have died (fire shield)
			{
```

client/battle/CBattleInterfaceClasses.h
ArseniyShestakov: We'll certainly have them. We talked about that with @alexvins privately and there is certainly very few cases where you have legit reason to use comma in declaration.

I commented on it to Tow: https://github.com/vcmi/vcmi/pull/347#discussion_r127613322

AI/StupidAI/StupidAI.cpp
```diff
@@ -94,20 +96,23 @@ bool isCloser(const EnemyInfo & ei1, const EnemyInfo & ei2, const ReachabilityIn
...
+static bool willSecondHexBlockMoreEnemyShooters(const BattleHex & h1, const BattleHex & h2)
 {
-	int shooters[2] = {0}; //count of shooters on hexes
+	int shooters[2] =
+	{
+		0
+	}; //count of shooters on hexes
 
 	for(int i = 0; i < 2; i++)
-		for (auto & neighbour : (i ? h2 : h1).neighbouringTiles())
```
ArseniyShestakov: This is something need separate fixing.

AI/VCAI/AIUtility.cpp
```diff
@@ -126,7 +128,7 @@ const CGHeroInstance * HeroPtr::operator*() const
 	return get();
 }
 
-void foreach_tile_pos(std::function<void(const int3& pos)> foo)
+void foreach_tile_pos(std::function<void(const int3 &pos)> foo)
```
ArseniyShestakov: Looks like a bug. It's all formatted automatically.

AI/VCAI/VCAI.cpp
```diff
@@ -703,7 +703,8 @@ void VCAI::showGarrisonDialog(const CArmedInstance *up, const CGHeroInstance *do
 void VCAI::showMapObjectSelectDialog(QueryID askID, const Component & icon, const MetaString & title, const MetaString & description, const std::vector<ObjectInstanceID> & objects)
 {
 	status.addQuery(askID, "Map object select query");
-	requestActionASAP([=]{ answerQuery(askID, 0); });
+	requestActionASAP([ = ] { answerQuery(askID, 0);
+	});
```
ArseniyShestakov: Yes.

client/CDefHandler.cpp
```diff
@@ -152,102 +153,104 @@ SDL_Surface * CDefHandler::getSprite (int SIndex, const ui8 * FDef, const SDL_Co
...
+	SDL_FreePalette(p);
 
-	int ftcp=0;
+	int ftcp = 0;
 
 	// If there's a margin anywhere, just blank out the whole surface.
-	if (TopMargin > 0 || BottomMargin > 0 || LeftMargin > 0 || RightMargin > 0) {
-		memset( reinterpret_cast<char*>(ret->pixels), 0, FullHeight*FullWidth);
+	if(TopMargin > 0 || BottomMargin > 0 || LeftMargin > 0 || RightMargin > 0)
+	{
```
ArseniyShestakov: It did not add {}. All code modification options are disabled.

Global.h
```diff
@@ -279,442 +279,397 @@ void inline handleException()
...
+}
 
-	//assigns smaller of (a, b) to a and returns minimum of (a, b)
-	template <typename t1, typename t2>
-	t1 &amin(t1 &a, const t2 &b)
+//assigns greater of (a, b) to a and returns maximum of (a, b)
+template<typename t1, typename t2> t1 & amax(t1 & a, const t2 & b)
+{
+	if(a >= b)
+		return a;
```
ArseniyShestakov: Don't think there way to enforce something like that, but I'll add it to wiki.

client/battle/CCreatureAnimation.cpp
dydzio0614: I think this particular code could be refactored to switch(getType())... for the sake of readability after reformatting

AI/VCAI/AIUtility.cpp
dydzio0614: maybe the digit "3" is source of bug. However with bugs like this... that source file will get bugged here everytime its processed by uncrustify

AI/VCAI/Goals.cpp
```diff
@@ -883,43 +879,43 @@ TSubgoal GatherTroops::whatToDoToAchieve()
 				continue;
 
 			BuildingID bid(BuildingID::DWELL_FIRST + creature->level - 1 + upgradeNumber * GameConstants::CREATURES_PER_TOWN);
-			if (t->hasBuilt(bid)) //this assumes only creatures with dwellings are assigned to faction
+			if(t->hasBuilt(bid)) //this assumes only creatures with dwellings are assigned to faction
 			{
```
dydzio0614: this change only fixed spacebar - excessive { } is separate problem. This commit only changed whitespaces.

client/CPreGame.cpp
dydzio0614: I personally like this change

ArseniyShestakov: @dydzio0614 Alex mean that extra semicolon ";"

dydzio0614: ah, "extra". Okay

client/CMT.cpp
```diff
@@ -1295,9 +1292,12 @@ static void handleEvent(SDL_Event & ev)
...
 {
 	SettingsListener resChanged = settings.listen["video"]["fullscreen"];
-	resChanged([](const JsonNode &newState){  CGuiHandler::pushSDLEvent(SDL_USEREVENT, FULLSCREEN_TOGGLED); });
+	resChanged([](const JsonNode & newState)
+	{
+		CGuiHandler::pushSDLEvent(SDL_USEREVENT, FULLSCREEN_TOGGLED);
+	});
 
-	inGuiThread.reset(new bool(true));
+	inGuiThread.reset(new bool (true));
```
ArseniyShestakov: I think uncrustify would do the same for any of standard types. I suppose it's might be a bug.

Global.h
```diff
@@ -279,442 +279,397 @@ void inline handleException()
...
+}
 
-	template <typename V, typename Item, typename Item2>
-	bool contains(const std::unordered_map<Item,V> & c, const Item2 &i)
-	{
-		return c.find(i)!=c.end();
-	}
+//returns true if unordered set c contains item i
+template<typename Item>	bool contains(const std::unordered_set<Item> & c, const Item & i)
+{
```
ArseniyShestakov: This is problem with all of "nl_" options and ```nl_template_class``` in particular. Basically when newline removed it's keep that extra tab untouched even in case of colons and such. Se we'll need to manually replace some tabs with spaces once.

Though we already decide we don't want template to be on same line everywhere.

ArseniyShestakov:
I just pushed new preview with following changes:

```pos_bool = ignore```
- Joining of bool operators '&&' into one-line is disabled. This expose problem with indent of multi-line if-else conditions.
```
nl_cpp_lambda_leave_one_liners = true
nl_assign_leave_one_liners = true
```
- Formatting of single-line lambdas and assignments wouldn't be touched now, but we added [new rules regarding lambdas usage on wiki](https://wiki.vcmi.eu/index.php?title=Coding_guidelines#Lambda_expressions).
```nl_template_class = ignore```
- New line presence after template<> will be ignored. It's up to developer to add one or keep it same line.
```align_keep_tabs = false```
- Will remove some of useless spaces remaining after new line removal, e.g one going after comma in lists in place of space.
```indent_else_if = true```
- Will convert ambiguous if-else indent that don't have braces. E.g I added braces around two of such statements in cdd9dbc7f1228f62b872e800cca301e0ee38fb19. Also rule about that added to wiki that if indention is required then braces should be present.

**Lambdas are still broken and multi-line conditions have wrong indent!**

File: lib/mapping/CMapEditManager.h [hunk 39]
```diff
@@ -58,8 +58,9 @@ template<typename T>
 class DLL_LINKAGE CMapSelection
 {
 public:
-	explicit CMapSelection(CMap * map) : map(map) { }
-	virtual ~CMapSelection() { };
+	explicit CMapSelection(CMap * map)
+		: map(map) {}
```
ArseniyShestakov: Some imperfection here, but might be we need no constructors with initializers lists in headers?

AI/VCAI/Fuzzy.cpp
```diff
@@ -145,14 +145,14 @@ void FuzzyHelper::initTacticalAdvantage()
...
 		ta.engine.addInputVariable(ta.castleWalls);
 		{
-			fl::Rectangle* none = new fl::Rectangle("NONE", CGTownInstance::NONE, CGTownInstance::NONE + (CGTownInstance::FORT - CGTownInstance::NONE) * 0.5f);
+			fl::Rectangle * none = new fl::Rectangle("NONE", CGTownInstance::NONE, CGTownInstance::NONE + (CGTownInstance::FORT - CGTownInstance::NONE) * 0.5f);
 			ta.castleWalls->addTerm(none);
 
-			fl::Trapezoid* medium = new fl::Trapezoid("MEDIUM", (CGTownInstance::FORT - CGTownInstance::NONE) * 0.5f, CGTownInstance::FORT,
-				CGTownInstance::CITADEL, CGTownInstance::CITADEL + (CGTownInstance::CASTLE - CGTownInstance::CITADEL) * 0.5f);
+			fl::Trapezoid * medium = new fl::Trapezoid("MEDIUM", (CGTownInstance::FORT - CGTownInstance::NONE) * 0.5f, CGTownInstance::FORT,
+								   CGTownInstance::CITADEL, CGTownInstance::CITADEL + (CGTownInstance::CASTLE - CGTownInstance::CITADEL) * 0.5f);
```
alexvins: I don`t like such aligning, even if it would be accepted align should be done with spaces

lib/mapping/CMapEditManager.h
alexvins: Impossible for templates, problematic for netpacks

AI/VCAI/Fuzzy.cpp
ArseniyShestakov: I don't like any alignment at all especially with spaces, but I still not find a way to control it in every case.

AI/StupidAI/StupidAI.cpp
DjWarmonger: It's not that simple, as some constructors have dozens of parameters and take up a lot of vertical space.

client/CBitmapHandler.cpp
DjWarmonger: Opening brace is a perfect place for a comment - it carries no other message, but lets us save some vertical space.

client/gui/Fonts.cpp
DjWarmonger: In the end I think it's up to coder to write a line so it's visually clear, using spaces and newlines. There might be simply no "best" set of rules which works for every case imaginable, I think this attempt is just futile.

ArseniyShestakov: @DjWarmonger even if we don't want to enforce it we still need to decide on sane default for source we have now since most old code is badly formatted.

PR #354, opened 2017-07-23
Android: better CTextInput handling

Fayth:
(not sure how we should handle features that need to modify both vcmi and vcmi-android projects -- in any case, I pushed vcmi-android's counterpart to https://github.com/Fayth/vcmi-android/tree/better-textinput )

The (main) problem with input handling in vcmi+android is that when SDL opens phone's software keyboard, the screen is partially hidden and, in most cases, the keyboard hides the CTextInput. In this scenario, the user doesn't see what is the content of the textbox until he closes the keyboard.

This PR (along with vcmi-android's part) allows vcmi to send CTextInput text changed events to android java code. Then we display an overlay showing what is being typed into the textbox (example: http://i.imgur.com/XOi7Sbd.png ). The overlay is visible while the soft keyboard is opened.

Additionally, I noticed 2 more problems with the text input (not really in scope of this PR):
- pressing enter on the soft keyboard might execute a hotkey action (e.g. in hotseat players list, pressing enter changes the displayed window) but it never unfocuses the textbox so the soft keyboard remains opened; should be relatively easy to fix -- manually unfocusing the control in the callback should work
- closing the soft keyboard (via phone's back button) doesn't unfocus the CTextInput, which prevents from opening the keyboard again until the field gets unfocused due to some other action; this might be hard to fix -- from what I checked, SDL eats this event and doesn't pass it to vcmi; possible solution is to create jni callback and notify about closing the keyboard manually

### DIFF
--- a/client/widgets/TextControls.cpp
+++ b/client/widgets/TextControls.cpp
@@ -18,6 +18,10 @@
 
 #include "../../lib/CGeneralTextHandler.h" //for Unicode related stuff
 
+#ifdef VCMI_ANDROID
+#include "lib/CAndroidVMHelper.h"
+#endif
+
 std::string CLabel::visibleText()
 {
 	return text;
@@ -427,6 +431,9 @@ CTextInput::CTextInput(const Rect &Pos, SDL_Surface *srf)
 void CTextInput::focusGot()
 {
 	CSDL_Ext::startTextInput(&pos);
+#ifdef VCMI_ANDROID
+	notifyAndroidTextInputChanged(text);
+#endif
 }
 
 void CTextInput::focusLost()
@@ -485,6 +492,9 @@ void CTextInput::keyPressed( const SDL_KeyboardEvent & key )
 	{
 		redraw();
 		cb(text);
+#ifdef VCMI_ANDROID
+		notifyAndroidTextInputChanged(text);
+#endif
 	}
 }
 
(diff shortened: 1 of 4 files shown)


ArseniyShestakov:
I know a there some apps on Android that simply take you into full-screen input window when you click on certain input fields. Can't we do the same for VCMI or there some problems with it?

Fayth:
Yeah, that's android's default behaviour in landscape mode. Not sure why SDL disables it. 
If you think it would be okay, I can try to do it this way.

ArseniyShestakov:
Only problem I could think about it's difficulty in determining if player actually want to use input. From what I understand it's will be better if you can actually click on input field to trigger fullscreen input and that's not problem for us since we only have input in few cases: player name, server ip and chat.

Though I understand that most of games wouldn't want fullscreen input since it's break immersion and most of mobile games actually have own code to display input data within their UI using their own style and fonts..

ArseniyShestakov:
Anyway let me know if we could merge this code anyway now since it's make input a bit better already. Especially if you have limited tine and fullscreen ennoblement might take some time.

Fayth:
Short version: I think this PR can be merged.

In general, user wants to use the input when he focuses it (usually by pressing the textbox). The problem is that vcmi CTextInput automatically tries to grab focus in ctor ( https://github.com/vcmi/vcmi/blob/develop/client/widgets/TextControls.cpp#L393 ) so it causes the soft keyboard to appear immediately when the textbox is created.

This would be even bigger problem with fullscreen edit, because currently it would work like this:
- user opens the window with textbox (e.g. new game -> multiplayer), 
- CTextInput grabs focus and starts SDL's SDL_StartTextInput,
- user is presented with a fullscreen input box and no clue what it's for.
Because of this, I think we can't use fullscreen edit as is (assuming it's possible with SDL, I haven't checked).

ArseniyShestakov:
Ok then let's merge it as is and if you find time to improve it then you can always do. Thanks.

PR #357, opened 2017-07-28
Code style: serialize everything on own lines for easier debugging

ArseniyShestakov:
This is helpful change for debugging both save breakages and sync problems between different compilers.

New rules added to wiki as well:
https://wiki.vcmi.eu/index.php?title=Coding_guidelines#Serialization

Some of used regexps:
```
(\s+)(h & [^ &]+) ([^;]+);
(\s+)(h & [^ &>\)]+) ([^;>\)]+);
\1\2;\n\1h \3;
```

### DIFF
--- a/AI/VCAI/AIUtility.h
+++ b/AI/VCAI/AIUtility.h
@@ -62,7 +62,9 @@ public:
 
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
-		h & this->h & hid & name;
+		h & this->h;
+		h & hid;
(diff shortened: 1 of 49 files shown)


ArseniyShestakov:
Just in case please don't waste your time looking at diff itself.
I checked every single line of it and it's fine.

PR mainly to see if anyone have any suggestions on how else serialize template could be improved.

alexvins:
If you see serialize template is too long then class probably have too many fields.

PR #358, opened 2017-07-29
Fixed bugs in Altar

Chocimier:


client/windows/CTradeWindow.cpp
```diff
@@ -1314,10 +1316,28 @@ void CAltarWindow::selectionChanged(bool side)
...
 }
 
+void CAltarWindow::selectOppositeItem(bool side)
+{
+	bool oppositeSide = !side;
+	int pos = vstd::find_pos(items[side], side ? hLeft : hRight);
+	int oppositePos = vstd::find_pos(items[oppositeSide], oppositeSide ? hLeft : hRight);
+
+	if (pos >= 0 && pos != oppositePos)
+	{
```
alexvins: remove space after `if`

ArseniyShestakov:
Basically to fix Altar you need to do one of two things: change TradeOnMarketplace netpack so it's can handle altrar transations better, e.g allow to "sell" more than one thing at a time.

Other option is implement different callback, netpack and server-side code for Altar so you might avoid touching other trading code.

ArseniyShestakov:
And yeah might be you meant that second commit fixed some other bug and not 2607?
Then fix commit description and I'll merge this change separately.

ArseniyShestakov:
Yeah to be clear what problem is there with multiple stack / artifact sacrifices. You could use ```vcmiainur``` cheat or just grant hero plenty of artifacts and sacrifice them at once. Problem occur when server send query about hero levelup to client before sacrifice is complete or there simply too many requests to server from client at once.

Chocimier:
Thanks for tips. Second patch makes creatures selected to sacrifice appear in right panel. I didn't test case when one stack give new level, so it seemed to be fixed.

PR #360, opened 2017-08-01
Use boost::make_optional for Boost 1.64 compatibility

ArseniyShestakov:
I certainly unsure this it the right way to fix this issue. Shortly in Boost 1.63 there is breaking change in Boost::optional supposedly related [to this fix](https://svn.boost.org/trac10/ticket/12652). So when VCMI compiled with Boost 1.64 and newer in MSVS 2015 following error occur:
```
D:\DEV\vcmi\lib\mapObjects\CGHeroInstance.cpp(1449): error C2593: 'operator =' is ambiguous [D:\DEV\vcmi2\lib\vcmi.vcxproj]
d:\dev\vcmi\lib\registertypes\../serializer/BinaryDeserializer.h(501): error C2593: 'operator =' is ambiguous [D:\DEV\vcmi2\lib\vcmi.vcxproj]
```
They're not limited to following case, but affect every single assignment to boost::optional in our code.

If anyone have idea for the better way handle this then it's would be great.

### DIFF
--- a/lib/serializer/BinaryDeserializer.h
+++ b/lib/serializer/BinaryDeserializer.h
@@ -496,9 +496,10 @@ public:
 		load( present );
 		if(present)
 		{
+			//TODO: replace with emplace once we start request Boost 1.56+, see PR360
 			T t;
 			load(t);
-			data = std::move(t);
+			data = boost::make_optional(std::move(t));
 		}
 		else
 		{
--- a/client/CPreGame.cpp
+++ b/client/CPreGame.cpp
@@ -3697,7 +3697,7 @@ void CBonusSelection::updateBonusSelection()
 		anim->setCustom(picName, 0);
 		bonusButton->setImage(anim);
 		const SDL_Color brightYellow = { 242, 226, 110, 0 };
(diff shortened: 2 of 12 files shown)


ArseniyShestakov:
Supposedly @edeksumo had problem with it too.

lib/serializer/BinaryDeserializer.h
```diff
@@ -498,7 +498,7 @@ class DLL_LINKAGE BinaryDeserializer : public CLoaderBase
 		{
 			T t;
 			load(t);
-			data = std::move(t);
+			data = boost::make_optional(std::move(t));
```
ArseniyShestakov: Tow suggested to use emplace here:
```
			data.emplace();
			load(*data);
```
It's seems to work well, but in other cases I not sure whatever we should use it since ```boost::make_optional``` seems to be easier to understand it's boost::optional value is assigned here.

ArseniyShestakov:
My bad. Our Travis build use Boost 1.54 and "emplace" only implemented in 1.56.

PR #359, opened 2017-08-01
Spells refactoring9

alexvins:
**Ready for review, testing and merging**

- This is redesign, do not expect many new features
- This is redesign, do not expect fixing bugs in this PR but if you fixing something related to spells, please make pull request to _this_ branch
- This is just part 9 of redesign, do not expect unit and even integration tests, but it close to it.
- New feature: custom spell effects aka Spell Format v2
- Save format changed, mods and saves compatibility retained

PR #363, opened 2017-08-11
Obstacle system refreshing

Illithidek:
I want redesign/add some obstacles functionality. Like any wip, it's far away from end.
Anyway, I put it here for all to know that such thing exists.
Pls, don't comment about code formatting, serialization corruption with old saves, some unused code etc. In few words for me there are things that I will do during work on it.

EDIT:
TODO for now:

- [x] Primary obstacles

- [x] Basic modding for static obstacles 

- [ ] Basic modding for magic obstacles

- [x] Suggested by _Andruids_: Adding dissapear and intereaction animation


ArseniyShestakov:
Please consider fixing this as well:
https://bugs.vcmi.eu/view.php?id=2786

FeniksFire:
Fixed: https://bugs.vcmi.eu/view.php?id=2786

lib/battle/obstacle/ObstacleJson.h
```diff
@@ -10,21 +10,37 @@
...
 	virtual int32_t getOffsetGraphicsInX() const;
 	virtual bool getCanBeRemovedBySpell() const;
-	virtual int32_t getPosition() const;
-	virtual std::string getGraphicsName() const;
+	virtual int16_t getPosition() const;
+	virtual std::vector<std::string> getPlace() const;
+	virtual int32_t getDamage() const;
+
+	virtual int8_t getSpellLevel() const;
+	virtual int8_t getSpellPower() const;
```
alexvins: Spell power should be 32 bit

lib/battle/obstacle/ObstacleJson.h
```diff
@@ -10,21 +10,37 @@
...
+	virtual int8_t getBattleSide() const;
+	virtual int8_t isVisibleForAnotherSide() const;
+	virtual int8_t getTurnsRemaining() const;
```
alexvins: Turns remaining should be 16 bit at least

FeniksFire:
Serialization for old saves should working now.

lib/NetPacks.h
```diff
@@ -1738,6 +1738,11 @@ struct ObstaclesRemoved : public CPackForClient
 
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
+		if(version < 777 && !h.saving)
```
alexvins: We do not use backward compatibility for NetPacks, server and client must have same version.

lib/CTownHandler.h
```diff
@@ -253,6 +253,21 @@ class DLL_LINKAGE CTown
...
+			h & moatDamage;
+		}
+		if(version >= 758 && version < 777)
+		{
+			std::vector<BattleHex> moatHexes;
+			h & moatHexes;
+		}
+		else if(!h.saving)
+		{
+			std::vector<BattleHex> moatHexes = {11, 28, 44, 61, 77, 111, 129, 146, 164, 181};
```
alexvins: I do not understand this line

FeniksFire: hahah, just copy old code and don't see it :D

lib/CTownHandler.h
```diff
@@ -253,6 +253,21 @@ class DLL_LINKAGE CTown
 
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
+		if(version < 777 && !h.saving)
```
alexvins: You should not check for `h.saving` in this case. Feel free to raise exception if saving with old version but do not just ignore. `h.saving` check in serialization should be used to initialize newly added fields when loading old save.

lib/CTownHandler.h
```diff
@@ -268,15 +279,6 @@ class DLL_LINKAGE CTown
 		h & primaryRes;
 		h & warMachine;
 		h & clientInfo;
-		h & moatDamage;
```
alexvins: Serialization order is important. Move backward compatibility code here

lib/NetPacks.h
```diff
@@ -1733,11 +1734,16 @@ struct ObstaclesRemoved : public CPackForClient
...
 
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
-		h & obstacles;
+		if(version < 777 && !h.saving)
+		{
+			std::set<si32> obstacles; //uniqueIDs of removed obstacles
+			h & obstacles;
+		}
+		h & id;
```
alexvins: this line must be conditional (only for newer version).

lib/battle/RangeGenerator.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * RangeGenerator.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#include "StdInc.h"
+#include "RangeGenerator.h".h"
```
alexvins: Copy-past error

lib/battle/obstacle/StaticObstacle.cpp
```diff
@@ -0,0 +1,59 @@
...
+bool StaticObstacle::canRemovedBySpell() const
+{
+	return canBeRemovedBySpell;
+}
+
+StaticObstacle::StaticObstacle()
+{
+}
+
+StaticObstacle::StaticObstacle(ObstacleJson info, int16_t position)
```
alexvins: Should be `const ObstacleJson & info`

lib/battle/obstacle/StaticObstacle.cpp
```diff
@@ -0,0 +1,59 @@
...
+{
+	if(!info.randomPosition())
+		position = info.getPosition();
+	setArea(ObstacleArea(info.getArea(),position));
+	canBeRemovedBySpell = info.canBeRemovedBySpell();
+	setGraphicsInfo(info.getGraphicsInfo());
+}
+
+StaticObstacle::StaticObstacle(ObstacleJson * info, int16_t position)
```
alexvins: Should be const ObstacleJson * info

lib/battle/obstacle/SpellCreatedObstacle.cpp
```diff
@@ -0,0 +1,84 @@
...
+	casterSide = info.getBattleSide();
+	setGraphicsInfo(info.getGraphicsInfo());
+}
+
+bool SpellCreatedObstacle::canRemovedBySpell() const
+{
+	switch (getType())
+	{
+	case ObstacleType::FIRE_WALL:
+		if(spellLevel >= 2)
```
alexvins: This is completely wrong. Result depends on remove obstacle spell level but not the level obstacle itself was created with.
It you want this logic be here, pass spellLevel from RemoveObstacleMechanics

lib/battle/obstacle/SpellCreatedObstacle.h
```diff
@@ -0,0 +1,44 @@
...
+#pragma once
+#include "Obstacle.h"
+#include "ObstacleJson.h"
+
+class DLL_LINKAGE SpellCreatedObstacle : public Obstacle
+{
+public:
+	ObstacleType obstacleType;
+	si8 turnsRemaining;
+	si8 casterSpellPower;
```
alexvins: spell power and turns remainingshould be at least 32 bit,

lib/battle/obstacle/SpellCreatedObstacle.h
```diff
@@ -0,0 +1,44 @@
...
+	si8 spellLevel;
+	si8 casterSide;
+	si8 visibleForAnotherSide;
```
alexvins: this should be `bool`

lib/battle/obstacle/StaticObstacle.h
```diff
@@ -0,0 +1,35 @@
...
+ */
+#pragma once
+#include "../BattleHex.h"
+#include "ObstacleJson.h"
+#include "Obstacle.h"
+
+class DLL_LINKAGE StaticObstacle : public Obstacle
+{
+public:
+	bool canBeRemovedBySpell = 0;
```
alexvins: Please do not initialize boolean with a number.

lib/battle/obstacle/ObstacleRandomGenerator.h
```diff
@@ -0,0 +1,22 @@
...
+ * ObstacleRandomGenerator.h, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#pragma once
+#include "battle/RangeGenerator.h".h"
```
alexvins: Copy-paste error

lib/battle/obstacle/ObstacleJson.cpp
```diff
@@ -0,0 +1,125 @@
...
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#include "StdInc.h"
+#include "ObstacleJson.h"
+
+ObstacleType ObstacleJson::typeConvertFromString(std::string type) const
```
alexvins: should be `const std::string &`

lib/battle/obstacle/ObstacleArea.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#pragma once
+#include "../BattleHex.h"
+#include "boost/algorithm/searching/boyer_moore.hpp"
```
alexvins: this include is not used in any of declarations, should be moved to .cpp

lib/battle/obstacle/Obstacle.h
```diff
@@ -0,0 +1,56 @@
...
+			h & uniqueID;
+			h & obstacleType;
+			h & pos;
+		}
+		h & ID;
+		h & area;
+		h & graphicsInfo;
+	}
+private:
+	boost::uuids::uuid generateID();
```
alexvins: this method should be static

lib/battle/BattleInfo.cpp
```diff
@@ -646,10 +461,174 @@ BattleInfo * BattleInfo::setupBattle(int3 tile, ETerrainType terrain, BFieldType
...
+					int x = blocked.getX();
+					if(x <= 2 || x >= 14)
+						return false;
+				}
+				return true;
+			};
+			if(info.randomPosition())
+			{
+				std::vector<int> fields;
+				for(int i = 18; i<=168; i++)
```
alexvins: this should be defined as constants, also it need some explanation

config/obstacles.json
```diff
@@ -3,1233 +3,1206 @@
...
+		"battlefieldSurface" : [16],
+		"offsetGraphicsInX" : 300,
+		"offsetGraphicsInY" : 214,
 		"blockedTiles" :  [59, 60, 74, 75, 76, 93, 94, 95, 111, 112],
-		"defname" : "ObMCL00.pcx"
+		"graphics" : "ObMCL00.pcx",
+		"canBeRemoved" : false,
+	}
+],
+"InherentObstacles" : [
```
alexvins: use camelCase for JSON field names

config/obstacles.json
```diff
@@ -3,1233 +3,1206 @@
...
+	{
+		"type" : "moat",
+		"blockedTiles" : [11, 28, 44, 61, 77, 111, 129, 146, 164, 181],
+		"damage" : 70,
+		"offsetGraphicsInX" : 410,
+		"offsetGraphicsInY" : 90,
+		"graphics" : "SgCsMoat.pcx",
+		"place" : ["Castle"]
```
alexvins: Where should modded castles moat configuration should be?

FeniksFire: what you mean? By default moat should be in InherentObstacles since it shouldn't be random I think.

alexvins: Each new town added in mod has own moat graphics

alexvins: You`ve removed  moat graphics handling from `getSiegeName`, but recreated only original towns configuration in obstacles.json. What to do with modded towns?

FeniksFire: Hmm, I will try with Cove but should work if you write ""place" : ["Cove"]" I suppose.

FeniksFire: It's working, you can also change the graphics if you want.
zrzut ekranu z 2017-10-14 07-57-16

FeniksFire: @alexvins ok, I even extract graphics and update a hota mod for future
zrzut ekranu z 2017-10-14 09-07-17

config/obstacles.json
FeniksFire: I suppose you mean lower camel case, since I used upper camel case.

config/obstacles.json
alexvins: The problem is that you can not change `config/obstacles.json` from two mods at the same time (only one override will be actually used), each mod must have own configuration. We have moat graphics format in town options (see "config/schemas/townSiege.json").

FeniksFire: Old format... I deleted things that use him, since it was totally bad, it's not even about using, it's about design. I know about conflicts, there are many ways to bypass them, and back to old format is the worst. Still it's wip since there is needs to repair that conflict and random obstacle generator(now it's only generating static obstacles).

alexvins: You do not have to update all the mods yourself, you "just" need to retain backward compatibility.

FeniksFire: It's terrible idea for me, backward serialization is enought, I prefer update all mods, then keep that old compatibility.

alexvins: Backward compatibility for mods is even more important than save format's one.

FeniksFire: If all projects have such systems then nothing would develop, I can care for backwards mods compatibility but I can also update them. I can restore old code, or care and make new one, better one. Anyway I won't back to these solutions.

alexvins: Lets move discussion to forum https://forum.vcmi.eu/t/poll-about-backward-compatibility/4335

lib/battle/handler/BattlefieldHandler.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void BattlefieldHandler::loadObject(std::string scope, std::string name, const JsonNode & data, size_t index)
+{
+	auto object = std::make_shared<ObstacleJson>(data);
+	obstacleConfigs[index] = object;
+}
+
+std::vector<JsonNode> BattlefieldHandler::loadLegacyData(size_t dataSize)
+{
+	std::vector<JsonNode> data;
```
alexvins: * Do not use legacy data if you actually do have it.
* even if you use legacy data API, `loadLegacyData` must return vector of individual entities, but not 1-element vector with entire configuration

lib/battle/handler/BattlefieldHandler.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "StdInc.h"
+#include "BattlefieldHandler.h"
+#include "CModHandler.h"
+#include "battle/obstacle/ObstacleJson.h"
+
+BattlefieldHandler::BattlefieldHandler()
+{
+	const JsonNode obstacleConfig(JsonNode(ResourceID("config/obstacles.json")));
+
```
alexvins: * why handler for obstacle types called BattlefieldHandler? Battlefield type is another thing (see `BFieldType`).
* Handler should not access .json files directly. `config/obstacles.json` link should be in `gameConfig.json` (aka mod.json for Core mod)

lib/battle/handler/BattlefieldHandler.cpp
```diff
@@ -0,0 +1,61 @@
...
+	std::vector<JsonNode> data;
+	data.push_back(JsonNode(ResourceID("config/obstacles.json")));
+	return data;
+}
+
+std::vector<bool> BattlefieldHandler::getDefaultAllowed() const
+{
+	std::vector<bool> allowed;
+	for(auto i : obstacleConfigs)
+		allowed.push_back(true);
```
alexvins: Use `allowed.resize(obstacleConfigs.size(), true)`

lib/battle/handler/BattlefieldHandler.cpp
```diff
@@ -0,0 +1,61 @@
...
+	return allowed;
+}
+
+std::vector<std::shared_ptr<ObstacleJson> > BattlefieldHandler::getObstacleConfigs() const
+{
+	return obstacleConfigs;
```
alexvins: You`re should not copy entire vector here.

lib/battle/handler/BattlefieldHandler.h
```diff
@@ -0,0 +1,33 @@
...
+#pragma once
+
+#include "GameConstants.h"
+#include "IHandlerBase.h"
+#include "../obstacle/Obstacle.h"
+#include "../obstacle/ObstacleJson.h"
+
+class ObstacleJson;
+class DLL_LINKAGE BattlefieldHandler : public IHandlerBase
+{
```
alexvins: Handler should be serialized as part of LibClasses.

lib/battle/obstacle/ObstacleJson.cpp
```diff
@@ -114,6 +114,13 @@ bool ObstacleJson::randomPosition() const
 	return config["position"].isNull();
 }
 
+bool ObstacleJson::isInherent()
+{
+	if(config["inherent"].isNull())
+		return false;
+	return config["inherent"].Bool();
```
alexvins: This line is enough. `Bool()` returns `false` for null node.

config/obstacles.json
```diff
@@ -20,1189 +20,1346 @@
...
-		"visibleForAnotherSide" : 0,
-		"spellLevel" : 3,
-		"spellPower" : 5,
-		"offsetGraphicsInX" : 0,
-		"offsetGraphicsInY" : 0,
-		"graphics" : "C09SPF1.def",
-		"place" : ["Tower", "Crypt"]
-	}
-]
+"Obstacle001" :
```
alexvins: Object identifiers should also be in lowerCamelCase.

lib/battle/handler/BattlefieldHandler.cpp
FeniksFire: When I name it I hope that everything that is for arrangement of things and graphics on the battlefield, will be handled by this in future: siege, placement of units, graphics for battlefield etc. will go here. It's needs now a little refactor that seperate loading obstacles to another class, at all I was more worried about understand how quickly force newly handler to work to not waste too much time.

alexvins: The problem is that ModHandler do not designed to support one handler for different entity types. If you plan to make Battlefield type also configurable it will also need separate handler class.

FeniksFire: Maybe it was too far-reaching plan, BFieldType everything complicate, I will think about it a little more, I always change that name. However for future writing in mod.json 20 different jsons supported by 20 handlers in mod.json that set the one battle doesn't sound too good. It would be better to have one json that can set whole battle.

alexvins (dismissed):
This one is postponed until #359 merged

lib/battle/obstacle/ObstacleGraphicsInfo.h
```diff
@@ -12,31 +12,35 @@
...
 	
 	void serializeJson(JsonSerializeFormat & handler);
 	
-	
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
 		h & offsetGraphicsInX;
 		h & offsetGraphicsInY;
-		h & graphicsName;
+		h & graphics[0];
```
alexvins: Serializer can handle arrays by itself

client/battle/CBattleInterface.cpp
```diff
@@ -3401,15 +3290,29 @@ void CBattleInterface::showStacks(SDL_Surface *to, std::vector<const CStack *> s
...
+			auto img = anim.getImage(frameIndex);
+			if(obstacle->getArea().getPosition() != 0)
+				p = getObstaclePosition(img->height(), *obstacle);
+			img->draw(to, p.x + obstacle->getGraphicsInfo().getOffsetGraphicsInX(), p.y + obstacle->getGraphicsInfo().getOffsetGraphicsInY());
+		}
+		else if(resID.getType() == EResType::IMAGE)
 		{
-			Point p = getObstaclePosition(img, *obstacle);
-			img->draw(to, p.x, p.y);
+			auto bitmap = BitmapHandler::loadBitmap(obstacle->getGraphicsInfo().getGraphics(obstacle->state));
```
alexvins: Same here, try to cache obstacle bitmap.

client/battle/CBattleInterface.cpp
```diff
@@ -3752,3 +3589,30 @@ void CBattleInterface::showPiecesOfWall(SDL_Surface *to, std::vector<int> pieces
 		}
 	}
 }
+
+void CBattleInterface::obstacleAnimEffect(const Obstacle &oi, ObstacleState state, bool waitForAnimations)
```
alexvins: - Missing space after `&`. 
- Is ObstacleState copying intended?

client/battle/CBattleInterface.cpp
```diff
@@ -3752,3 +3589,30 @@ void CBattleInterface::showPiecesOfWall(SDL_Surface *to, std::vector<int> pieces
...
+	std::string graphicsname = oi.getGraphicsInfo().getGraphics(state);
+	
+	ResourceID resID(graphicsname);
+	
+	if(resID.getType() == EResType::ANIMATION)
+	{
+		CAnimation anim(graphicsname);
+		anim.preload();
+		int frameIndex = ((animCount + 1) * 25 / getAnimSpeed()) % anim.size();
+		imageHeight = anim.getImage(frameIndex)->height();
```
alexvins: You should find more effective way of obtaining height.

client/battle/CBattleInterface.h
```diff
@@ -1,407 +1,404 @@
-/*
```
alexvins: Whole file changed, line ending problem?

client/widgets/TextControls.h
```diff
@@ -42,7 +42,7 @@ class CLabel : public CTextContainer
 
 	std::shared_ptr<CPicture> background;
 public:
-
+
```
alexvins: just revert this

lib/CHeroHandler.h
```diff
@@ -329,7 +320,12 @@ class DLL_LINKAGE CHeroHandler : public IHandlerBase
 		h & expPerLevel;
 		h & ballistics;
 		h & terrCosts;
+		if(version < 777 && !h.saving)
+		{
+			std::map<int, CObstacleInfo> obstacles;
+			std::map<int, CObstacleInfo> absoluteObstacles;
 		h & obstacles;
```
alexvins: Fix indentation here.

lib/CMakeLists.txt
```diff
@@ -346,10 +377,11 @@ set(lib_HEADERS
 		StartInfo.h
 		StringConstants.h
 		UnlockGuard.h
+		UUID.h
 		VCMIDirs.h
 		vcmi_endian.h
 		VCMI_Lib.h
-)
+        battle/obstacle/ObstacleState.h)
```
alexvins: Put this line near other battle/obstacle headers

lib/CModHandler.cpp
```diff
@@ -17,6 +17,7 @@
 #include "CCreatureHandler.h"
 #include "CArtHandler.h"
 #include "CTownHandler.h"
+#include "battle/handler/BattlefieldHandler.h"
```
alexvins: use relative path

lib/NetPacksLib.cpp
```diff
@@ -1,1622 +1,1625 @@
-/*
```
alexvins: Whole file changed, problem with line endings?

lib/VCMI_Lib.h
```diff
@@ -66,6 +68,7 @@ class DLL_LINKAGE LibClasses
 		h & arth;
 		h & creh;
 		h & townh;
+		h & battlefieldHandler;
```
alexvins: Backward compatibility code is missing.

lib/battle/BattleInfo.cpp
```diff
@@ -1,1093 +1,1035 @@
-/*
```
alexvins: Whole file changed, problem with line endings?

lib/battle/BattleInfo.h
```diff
@@ -1,151 +1,154 @@
-/*
```
alexvins: Whole file changed, problem with line endings?

lib/battle/obstacle/Obstacle.h
```diff
@@ -0,0 +1,58 @@
...
+			BattleHex pos;
+			ui8 obstacleType;
+			si32 uniqueID;
+			h & id;
+			h & uniqueID;
+			h & obstacleType;
+			h & state;
+			h & pos;
+		}
+		h & ID;
```
alexvins: This should not be called in case of `version < 777`

lib/registerTypes/RegisterTypes.h
```diff
@@ -208,9 +209,9 @@ void registerTypesMapObjects2(Serializer &s)
 	s.template registerType<CBonusSystemNode, CArtifactInstance>();
 	s.template registerType<CArtifactInstance, CCombinedArtifactInstance>();
 
-	//s.template registerType<CObstacleInstance>();
-		s.template registerType<CObstacleInstance, MoatObstacle>();
-		s.template registerType<CObstacleInstance, SpellCreatedObstacle>();
+	s.template registerType<Obstacle, StaticObstacle>();
```
alexvins: This is a problem. You registering 4 classes where were 3. This completely breaks saves compatibilty.

alexvins (changes requested):

lib/battle/BattleInfo.cpp
FeniksFire: I made small test, by using push -f I checked for LF, CRLF, CR, but it's not resolve the problem, any ideas what it can be?

lib/battle/BattleInfo.h
FeniksFire: I made small test, by using push -f I checked for LF, CRLF, CR, but it's not resolve the problem, any ideas what it can be?

lib/NetPacksLib.cpp
FeniksFire: I made small test, by using push -f I checked for LF, CRLF, CR, but it's not resolve the problem, any ideas what it can be?

client/battle/CBattleInterface.h
FeniksFire: I made small test, by using push -f I checked for LF, CRLF, CR, but it's not resolve the problem, any ideas what it can be?

lib/battle/BattleInfo.h
alexvins: set 
``` 
[core]
	safecrlf = warn
```
in .gitconfig and recommit ff3a2148

FeniksFire: Thanks! Will check tommorow.

FeniksFire: Well with `safectrlf = warn` each time when I want  commit I get `fatal: bad numeric config value 'warn' for 'core.autocrlf': invalid unit`
Are you sure there must be `warn`?

alexvins: It should work, see https://git-scm.com/docs/git-config#git-config-coresafecrlf

FeniksFire: Huh, there were 2 problems. 
1) CLion seems to have some bug since it doesn't change line endings.
2)You was close but instead of `git config --global core.autocrlf warn` change to `git config --global core.autocrlf false` and this help me a bit.

lib/CModHandler.cpp
FeniksFire: What is wrong with it?

alexvins: Now it looks OK )

lib/UUID.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "UUID.h"
+
+boost::uuids::uuid UUID::getID() const
+{
+	return ID;
+}
+
+boost::uuids::uuid UUID::generateID()
+{
+	static boost::uuids::random_generator_mt19937 gen;
```
alexvins: Build fails here. No such symbol `random_generator_mt19937`. Boost version problem?

FeniksFire: You're right, CI works with `boost 1.64` when `boost 1.67` provide for  convenience 
` random_generator_mt19937` that is definition of previous version  `basic_random_generator<mt19937>`

lib/CHeroHandler.h
```diff
@@ -250,6 +222,29 @@ class DLL_LINKAGE CHeroClassHandler : public IHandlerBase
 	}
 };
 
+struct CObstacleInfo // for backward compatibility
```
alexvins: Put this where old CObstacleInfo was

FeniksFire: Ahh, ok.

PR #364, opened 2017-08-11
New log api

alexvins:
Just to test build

### DIFF
--- a/lib/CGameInfoCallback.cpp
+++ b/lib/CGameInfoCallback.cpp
@@ -22,9 +22,9 @@
 #include "CPlayerState.h"
 
 //TODO make clean
-#define ERROR_VERBOSE_OR_NOT_RET_VAL_IF(cond, verbose, txt, retVal) do {if(cond){if(verbose)logGlobal->errorStream() << BOOST_CURRENT_FUNCTION << ": " << txt; return retVal;}} while(0)
-#define ERROR_RET_IF(cond, txt) do {if(cond){logGlobal->errorStream() << BOOST_CURRENT_FUNCTION << ": " << txt; return;}} while(0)
-#define ERROR_RET_VAL_IF(cond, txt, retVal) do {if(cond){logGlobal->errorStream() << BOOST_CURRENT_FUNCTION << ": " << txt; return retVal;}} while(0)
+#define ERROR_VERBOSE_OR_NOT_RET_VAL_IF(cond, verbose, txt, retVal) do {if(cond){if(verbose)logGlobal->error("%s: %s",BOOST_CURRENT_FUNCTION, txt); return retVal;}} while(0)
+#define ERROR_RET_IF(cond, txt) do {if(cond){logGlobal->error("%s: %s", BOOST_CURRENT_FUNCTION, txt); return;}} while(0)
+#define ERROR_RET_VAL_IF(cond, txt, retVal) do {if(cond){logGlobal->error("%s: %s", BOOST_CURRENT_FUNCTION, txt); return retVal;}} while(0)
 
 PlayerColor CGameInfoCallback::getOwner(ObjectInstanceID heroID) const
 {
@@ -120,7 +120,7 @@ const CGObjectInstance* CGameInfoCallback::getObj(ObjectInstanceID objid, bool v
 	if(oid < 0  ||  oid >= gs->map->objects.size())
 	{
 		if(verbose)
-			logGlobal->errorStream() << "Cannot get object with id " << oid;
+			logGlobal->error("Cannot get object with id %d", oid);
 		return nullptr;
 	}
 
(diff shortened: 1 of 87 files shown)


lib/serializer/BinaryDeserializer.cpp
```diff
@@ -61,11 +62,11 @@ void CLoadFile::openNextFile(const boost::filesystem::path & fname, int minimalV
 
 			auto versionptr = (char*)&serializer.fileVersion;
 			std::reverse(versionptr, versionptr + 4);
-			logGlobal->warnStream() << "Version number reversed is " << serializer.fileVersion << ", checking...";
+			logGlobal->warn("Version number reversed is %x, checking...", serializer.fileVersion);
```
ArseniyShestakov: It's certainly was bugged before i suppose? Since fileVersion is just ```si32```.

lib/serializer/BinaryDeserializer.cpp
```diff
@@ -79,13 +80,11 @@ void CLoadFile::openNextFile(const boost::filesystem::path & fname, int minimalV
...
 
-void CLoadFile::reportState(CLogger * out)
+void CLoadFile::reportState(vstd::CLoggerBase * out)
 {
 	out->debug("CLoadFile");
 	if(!!sfile && *sfile)
-	{
-		out->debugStream() << "\tOpened " << fName << "\n\tPosition: " << sfile->tellg();
-	}
+		out->debug("\tOpened %s \tPosition: %d", fName, sfile->tellg());
```
ArseniyShestakov: Lost newline, but might be intended.

server/CGameHandler.cpp
```diff
@@ -6082,7 +6082,7 @@ void CGameHandler::spawnWanderingMonsters(CreatureID creatureID)
 	for (int i = 0; i < amount; ++i)
 	{
 		tile = tiles.begin();
-		logGlobal->trace("\tSpawning monster at %s",(*tile)());
+		logGlobal->trace("\tSpawning monster at %s",tile->toString());
```
ArseniyShestakov: space missing

lib/CGameInfoCallback.cpp
```diff
@@ -22,9 +22,9 @@
 #include "CPlayerState.h"
 
 //TODO make clean
-#define ERROR_VERBOSE_OR_NOT_RET_VAL_IF(cond, verbose, txt, retVal) do {if(cond){if(verbose)logGlobal->errorStream() << BOOST_CURRENT_FUNCTION << ": " << txt; return retVal;}} while(0)
-#define ERROR_RET_IF(cond, txt) do {if(cond){logGlobal->errorStream() << BOOST_CURRENT_FUNCTION << ": " << txt; return;}} while(0)
-#define ERROR_RET_VAL_IF(cond, txt, retVal) do {if(cond){logGlobal->errorStream() << BOOST_CURRENT_FUNCTION << ": " << txt; return retVal;}} while(0)
+#define ERROR_VERBOSE_OR_NOT_RET_VAL_IF(cond, verbose, txt, retVal) do {if(cond){if(verbose)logGlobal->error("%s: %s",BOOST_CURRENT_FUNCTION, txt); return retVal;}} while(0)
+#define ERROR_RET_IF(cond, txt) do {if(cond){logGlobal->error("%s: %s", BOOST_CURRENT_FUNCTION, txt); return;}} while(0)
```
alexvins: We using both, randomly. We should select one of them.

lib/serializer/BinaryDeserializer.cpp
alexvins: Yes, but I should remove \t also.

lib/CGameInfoCallback.cpp
ArseniyShestakov: I just supposed there were some reason why ```BOOST_CURRENT_FUNCTION``` used in Macro, but if there are none we could better stick with ```__FUNCTION__``` since it's should be supported well enough.

lib/serializer/BinaryDeserializer.cpp
ArseniyShestakov: @alexvins as you wish. No idea of extra tab is important to make this part of log look nicer.

lib/CGameInfoCallback.cpp
alexvins: BOOST_CURRENT_FUNCTION expands to more descriptive macro than __FUNCTION__ (f.e.__PRETTY_FUNCTION__ for GCC and __FUNCSIG__ for MSVS), but more descriptive is also much longer text.

lib/CGameState.cpp
```diff
@@ -718,7 +718,7 @@ void CGameState::init(StartInfo * si, bool allowSavingRandomMap)
 		initCampaign();
 		break;
 	default:
-		logGlobal->errorStream() << "Wrong mode: " << (int)scenarioOps->mode;
+		logGlobal->error("Wrong mode: %d", (int)scenarioOps->mode);
```
dydzio0614: Resharper suggests using static_cast for conversions though

lib/NetPacksLib.cpp
```diff
@@ -1730,14 +1730,18 @@ DLL_LINKAGE std::string CatapultAttack::AttackInfo::toString() const
...
+	std::ostringstream out;
+	out << "[";
+	for(auto it = attackedParts.begin(); it != attackedParts.end(); ++it)
+	{
+		out << it->toString();
+		if(std::prev(attackedParts.end()) != it) out << ", ";
+	}
+	out << "]";
+
+	return boost::str(boost::format("{CatapultAttack: attackedParts '%s', attacker '%d'}") % out.str() % attacker);
```
alexvins: this need more refactoring, for now it was moved from global.h 
`template<typename T>
std::ostream & operator<<(std::ostream & out, const std::vector<T> & container)`

I do not like default stream operation to be debug printing.

lib/CGameState.cpp
alexvins: will change to static_cast<int>

lib/serializer/BinaryDeserializer.cpp
alexvins: It should be HEX format because here we dealing with endianness.

PR #366, opened 2017-08-18
Fix CMake MSVS build problem

dydzio0614:
Same errors as here when building vcmi client and server: https://stackoverflow.com/questions/27196312/boost-1-57-0-program-options-msvs-2013-linker-error
Fix based on: https://groups.google.com/forum/#!topic/boost-list/Lhqvoet8N18


ArseniyShestakov:
Just in case problem with variadic templates in MSVC documented here:
https://developercommunity.visualstudio.com/content/problem/69009/compilation-error-on-msvc2015-on-boostbase-from-me.html

PR #368, opened 2017-08-19
Visual Studio files tweaks

dydzio0614:
This PR contains major improvements of hosted Visual Studio files.
*remove /zm compiler flag when building, allow compiler to use multi CPU cores by default
*better default library / include path, shared for all projects: /deps/include and /deps/libs - subfolders for x86 and 64 bit versions of libraries not added in favor of simplicity
*launcher building simplified - dependant on macro QTDIR containing Qt path
*Macro for desired VCMI output directory VCMI_Out - AI files will land in AI subfolder
*Setting paths for QTDIR and VCMI_Out in VCMI_Global_user.props, moving includes to deps/include (boost and ffmpeg libraries in appropiate subfolders - "boost", "libavcodec"...) and .lib files (all libs placed directly without subfolders) to deps/libs is enough to make VCMI build

To remove VCMI_Global_user.props from appearing in unstaged files run git command "git update-index --assume-unchanged VCMI_Global_user.props" and create backup to replace when needed as some git commands will restore it to the hosted version ("git checkout ." for example)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ VCMI_VS11.opensdf
 /test/RD
 /VCMI_VS11.VC.opendb
 /AI/FuzzyLite.lib
+/deps
--- a/AI/BattleAI/BattleAI.vcxproj
+++ b/AI/BattleAI/BattleAI.vcxproj
@@ -113,6 +113,7 @@
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>StdInc.h</PrecompiledHeaderFile>
       <AdditionalOptions>/Zm159 %(AdditionalOptions)</AdditionalOptions>
+      <MultiProcessorCompilation>true</MultiProcessorCompilation>
     </ClCompile>
     <Link>
       <AdditionalDependencies>VCMI_lib.lib;%(AdditionalDependencies)</AdditionalDependencies>
--- a/AI/BattleAI/BattleAI.vcxproj
+++ b/AI/BattleAI/BattleAI.vcxproj
@@ -21,6 +21,7 @@
   <PropertyGroup Label="Globals">
     <ProjectGuid>{C0300513-E845-43B4-9A4F-E8817EAEF57C}</ProjectGuid>
     <RootNamespace>BattleAI</RootNamespace>
+    <WindowsTargetPlatformVersion>8.1</WindowsTargetPlatformVersion>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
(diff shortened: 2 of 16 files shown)


adanski:
@ArseniyShestakov wouldn't it be easier to just completely drop VS-specific project files? VS2017 supports CMake just fine.

dydzio0614:
I personally support the opinion of Tow in this matter - quote:

If I were to suggest anything stick to following:
* stick to CMake as a primary, official way to build VCMI on any supported platform + tutorial how to do that
* provide any other build systems (like MS Build) as an semi-official (as in: stored on repo but not required to be always up-to-date) alternative for more advanced users that have a build system of their preferring; maintained and kept on repo as long as there is a VCMI developer using it

dydzio0614:
So few more points that are my opinion on this:
1. Dropping VS files support does not make things "easier" - they are "officially unsupported" way of building VCMI. CMake is primary and "easiest to setup without errors".
2. VS projects are alternative for people like me - with some VS experience who like working natively with the platform without external tools. Some maintenance knowledge may be required, though when libraries dont change it is limited to adding new files to compile. It's easier to grasp in 100% how non-generated project files work, changing build settings permanently is easy, especially with GUI usage.

adanski:
IMHO, having semi-official and unsupported things in the repo might make more harm than good and confuse people. I'd suggest to think about either officially supporting VS projects and always keep them up-to-date or deprecate them and eventually remove when VS2017 is more widespread.

dydzio0614:
I will keep them up to date as far as I work on VCMI as I do not plan to switch to cmake. Though it doesn't just depend of me what we give "official support" to. We also have code::blocks files in repo used by other active contributor.

I personally think there are people beside me in the world who would prefer to use "pure VS" workflow eventually for flexibility. Forcing them to reinvent the wheel by making own vcxprojs could not be the best idea.

adanski:
AFAIK, one doesn't have to generate VS project. VS2017 should natively support CMake project files.

https://blogs.msdn.microsoft.com/vcblog/2016/10/05/cmake-support-in-visual-studio/

dydzio0614:
Yep I know - though customizing build settings is easier without using cmake projects for experienced VS users. We are not at MSBuild support level yet.

ArseniyShestakov:
@adanski if VCMI was just my own project I would just drop everything except for CMake. The deal is: we also have Code::Blocks configuration inside our project files and it's used by our other lead developer. He has his own workflow and person who doing way more work on code than me certainly can use whatever build system he prefers. So why not VS as well?

I'm just made VCMI build reliable on all OS with CMake it's and it's exactly what we would recommend in our documentation. So I invested about 3 weeks of my work into making it easier for new contributors from any platform to start working on code, but I sure people who are going to work on C++ codebase are smart enough to read the wiki, or ask on forums or on Slack and use CMake if they want trouble-free build.

ArseniyShestakov:
And yeah I spent a lot of time in both versions of Visual Studio we support (2015 and 2017). While CMake support VS2017 has now is better than what you usually get with plugins or generated solution it's still nowhere near to be complete or bug-free experience.

PR #372, opened 2017-09-05
Gui cleanup2

alexvins:
Continued work on replacing CDefHandler with CAnimation


alexvins:
The only remain usage of CDefHandler is obstacles and put of scope of this PR

ArseniyShestakov:
Ok there currently one small bug with morale drawing position.
Other than that branch seems good and I can't see anything.

ArseniyShestakov:
This one now fixed. Another reproducible bug found: bodies are moving left in some situation for whatever reason.

PR #373, opened 2017-09-05
RandGen change type

FeniksFire:
Since map created by editor can have dimensions 1024x1024, "seed" can go out for the
range. So this can give us after radical counting
srand(int3(1024,1024,0)); rand(); number like 60961129830000 when
int32_t can have maximum 2147483647. In that case we need use int64_t
whose maximal value is 9223372036854775807.

### DIFF
--- a/client/windows/QuickRecruitmentWindow.cpp
+++ b/client/windows/QuickRecruitmentWindow.cpp
@@ -73,7 +73,7 @@ void QuickRecruitmentWindow::initWindow(Rect startupPosition)
 		pos.x -= 55 * (creaturesAmount - 3);
 	}
 	backgroundTexture = std::make_shared<CFilledTexture>("DIBOXBCK.pcx", Rect(0, 0, pos.w, pos.h));
-	costBackground = std::make_shared<CPicture>("QuickRecruitmentWindow/costBackground.png", pos.w/2-113, pos.y+290);
+	costBackground = std::make_shared<CPicture>("QuickRecruitmentWindow/costBackground.png", pos.w/2-113, 335);
 }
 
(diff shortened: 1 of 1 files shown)


DjWarmonger:
A simpler solution would be to take seed modulo 256. It won't change anything about existing maps, but will prevent overflow.

FeniksFire:
@DjWarmonger, I'm not quite sure what you mean so you can expand your idea? Did you mean that input in `pos.x, pos.y` should be modulo from 256?

FeniksFire:
@DjWarmonger good idea but not for 256, maximal seed for 256x256 is 15253376976000 so still can be out of range.

DjWarmonger:
Now that's an issue, as 252 x 252 maps already exist :/ Not sure if we need to support them exactly, though.

As far as I remember, this formula generates repeating 36 x 36 patterns.

FeniksFire:
Umm, just people can't see this since when seed is out of range is UB, and then seed can be equal to any number, after that is generated thanks for that seed some obstacle that is picked to battlefield... And then h3 behaviour and vcmi are different.

ArseniyShestakov:
Before this merged we should make sure that's change don't add more of incompatibility between different compilers.

alexvins:
@ArseniyShestakov this code is server-only. Battle is prepared on server side and then being sent to players.

ArseniyShestakov:
Oh, then we're good. Thanks for notice.

DjWarmonger:
Okay, then we should do both:
* Take int3 parameters modulo 252 (7*36) to make sure the value is capped.
* Use type of platform-independent int64_t everywhere. Can stay like now, though it doesn't get negative as far as I can see.

Also, please add comments that the magic numbers are used here for backward compatibility with OH3 obstacle placement.

FeniksFire:
closed for now, these things will be moved to obstacle branch

PR #376, opened 2017-09-10
New animations

alexvins:
New features
- support margins in .json animation
- new creatures animation types: distinct 2-hex attack and cast

### DIFF
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -1069,12 +1069,27 @@ void CBattleInterface::stacksAreAttacked(std::vector<StackAttackedInfo> attacked
 		}
 	}
 	waitForAnims();
-	int targets = 0, killed = 0, damage = 0;
-	for (auto & attackedInfo : attackedInfos)
+
+	std::array<int, 2> killedBySide = {0, 0};
+
+	int targets = 0, damage = 0;
+	for(const StackAttackedInfo & attackedInfo : attackedInfos)
 	{
 		++targets;
-		killed += attackedInfo.amountKilled;
 		damage += attackedInfo.dmg;
+
+		ui8 side = attackedInfo.defender->side;
+		killedBySide.at(side) += attackedInfo.amountKilled;
+	}
+
+	int killed = killedBySide[0] + killedBySide[1];
+
+	for(ui8 side = 0; side < 2; side++)
+	{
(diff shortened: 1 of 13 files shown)


edeksumo:
For me creatures still don't play cast animation. And as I remember in OH, when creature don't have casting animation it will play coresponding attack anim.

edeksumo:
Now it's working. But now that I'm looking on this, some creatures shouldn't play this animation, like wyverns and it's looking odd when them strike twice. I thought that mayby it shoulld be ranged animation used than meelee but archangels don't have it and as they resurect they play meelee animation.

edeksumo:
So I look't into archangel def and he has anim for casting so when they don't have anim for casting they use ranged attack animation.

edeksumo:
Now it's working allmost like in OH, but ogre magi don't play ani when casting (have animation), and for example basilsk (?) have animation to cast after attack but don't work.  And alternative death ani is working but is VERY lagy like in slow motion.

alexvins:
There is no generic way to determine should animation be played after attack or not, I just blocked it.

edeksumo:
There is no way to check that creature have animation to casting spell and play it? And what about ogre magi?

alexvins:
In VCMI any creature can have any ability, so I just blocked animation for after and before attack cast. I`ll check ogre magi in a few miniutes

misiokles:
I think that not all animations are displayed in battle hero defs also. For example - shuttle part of such defs. When battle is on and player doesn't do any click for some time, hero's horse start to 'breath' and 'shuttle' legs.

alexvins:
@misiokles you`re right, only standing and cast actually used.

client/battle/CBattleInterface.cpp
```diff
@@ -1416,9 +1416,11 @@ CBattleInterface::PossibleActions CBattleInterface::getCasterAction(const CSpell
 void CBattleInterface::setHeroAnimation(ui8 side, int phase)
 {
 	if(side == BattleSide::ATTACKER)
-		attackingHero->setPhase(phase);
+		if(attackingHero)
+			attackingHero->setPhase(phase);
 	else
-		defendingHero->setPhase(phase);
+		if(defendingHero)
+			defendingHero->setPhase(phase);
```
ArseniyShestakov: I think we decide explicitly that something like that would need braces.

alexvins: fixed

PR #379, opened 2017-09-14
Hero specialty scaling

henningkoehlernz:
The current way hero specialties can scale is very restricted. This change introduces updaters to bonuses, which perform updates (currently val recalculation) based on bearer attributes, and simplifies the json specialties format. It also converts the old

	[ { "type":13, "val": 1, "subtype": 1, "info": 5 } ]
and

	[ { "growsWithLevel" : true, "bonuses" : [...] } ]
formats to new format.

Notes:
* Updates to bonuses are performed during inheritance. So e.g. an artifact could grant a scaling bonus to centaurs, which would then be updated when inherited by the hero wearing it. This updated bonus is only used during bonus calculations, and never stored at Bonus System nodes - so using "bonuses" command-line command will show un-updated bonuses, and multiple heroes wearing the same artifact won't interfere with each other.
* The following updaters are currently available:
  * GrowsWithLevelUpdater ("GROWS_WITH_LEVEL"). Parameters are valPer20 and stepSize, and it updates val to ceil(valPer20 * floor(level / stepSize) / 20). StepSize is optional and defaults to 1.
  * TimesHeroLevelUpdater ("TIMES_HERO_LEVEL"). Multiplies val with the level of the hero.
  * TimesStackLevelUpdater ("TIMES_STACK_LEVEL"). Multiplies val with the level of the stack. Works for stacks without instance (summons, war machines), but war machines currently have level 0.
* Set logging of domain "mod" to "trace" to see result of old format conversion.

New format preview:

	"specialty" : {
		"bonuses" : {
			"archery" : {
				"subtype" : "skill.archery",
				"type" : "SECONDARY_SKILL_PREMY",
				"updater" : "TIMES_HERO_LEVEL",
				"val" : 5,
				"valueType" : "PERCENT_TO_BASE"
			}
		}
	}

	"specialty" : {
		"base" : {
			"limiters" : [
				{
					"parameters" : [ "archer", true ],
					"type" : "CREATURE_TYPE_LIMITER"
				}
			]
		},
		"bonuses" : {
			"attack" : {
				"subtype" : "primSkill.attack",
				"type" : "PRIMARY_SKILL",
				"updater" : {
					"parameters" : [ 6, 2 ],
					"type" : "GROWS_WITH_LEVEL"
				}
			},
			"defence" : {
				"subtype" : "primSkill.defence",
				"type" : "PRIMARY_SKILL",
				"updater" : {
					"parameters" : [ 3, 2 ],
					"type" : "GROWS_WITH_LEVEL"
				}
			},
			"speed" : {
				"type" : "STACKS_SPEED",
				"val" : 1
			}
		}
	}

Alternatively the latter specialty can be written short as follows, which will be converted internally using creature stats to follow the HMM3 default formula:

	"specialty" : {
		"creature" : "archer"
	}

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -376,8 +376,14 @@ namespace vstd
 		for(auto iter = map.cbegin(); iter != map.cend(); iter++)
 		{
 			if(iter->second == value)
+			{
+				if(found)
+					*found = true;
 				return iter->first;
+			}
 		}
+		if(found)
+			*found = false;
 		return Key();
 	}
 
--- a/Global.h
+++ b/Global.h
@@ -369,6 +369,18 @@ namespace vstd
 		return std::find(c.begin(),c.end(),i);
 	}
 
+	//returns first key that maps to given value if present, returns success via found if provided
+	template <typename Key, typename T>
+	Key findKey(const std::map<Key, T> & map, const T & value, bool * found = nullptr)
+	{
+		for(auto iter = map.cbegin(); iter != map.cend(); iter++)
+		{
+			if(iter->second == value)
+				return iter->first;
+		}
+		return Key();
+	}
+
 	//removes element i from container c, returns false if c does not contain i
 	template <typename Container, typename Item>
 	typename Container::size_type operator-=(Container &c, const Item &i)
--- a/config/heroes/necropolis.json
+++ b/config/heroes/necropolis.json
@@ -234,10 +234,8 @@
 				"necromancy" : {
 					"subtype" : "skill.necromancy",
 					"type" : "SECONDARY_SKILL_PREMY",
-					"updater" : {
-						"parameters" : [ 100 ],
-						"type" : "GROWS_WITH_LEVEL"
-					},
+					"updater" : "TIMES_HERO_LEVEL",
+					"val" : 5,
 					"valueType" : "PERCENT_TO_BASE"
 				}
 			}
@@ -366,10 +364,8 @@
 				"sorcery" : {
 					"subtype" : "skill.sorcery",
 					"type" : "SECONDARY_SKILL_PREMY",
-					"updater" : {
-						"parameters" : [ 100 ],
-						"type" : "GROWS_WITH_LEVEL"
-					},
+					"updater" : "TIMES_HERO_LEVEL",
+					"val" : 5,
 					"valueType" : "PERCENT_TO_BASE"
 				}
 			}
@@ -391,10 +387,8 @@
 				"eagleEye" : {
 					"subtype" : "skill.eagleEye",
 					"type" : "SECONDARY_SKILL_PREMY",
-					"updater" : {
-						"parameters" : [ 100 ],
-						"type" : "GROWS_WITH_LEVEL"
-					},
+					"updater" : "TIMES_HERO_LEVEL",
+					"val" : 5,
 					"valueType" : "PERCENT_TO_BASE"
 				}
 			}
@@ -457,10 +451,8 @@
 				"necromancy" : {
 					"subtype" : "skill.necromancy",
 					"type" : "SECONDARY_SKILL_PREMY",
-					"updater" : {
-						"parameters" : [ 100 ],
-						"type" : "GROWS_WITH_LEVEL"
-					},
+					"updater" : "TIMES_HERO_LEVEL",
+					"val" : 5,
 					"valueType" : "PERCENT_TO_BASE"
 				}
 			}
--- a/config/heroes/necropolis.json
+++ b/config/heroes/necropolis.json
@@ -148,7 +148,7 @@
 				"deathRipple" : {
 					"subtype" : "spell.deathRipple",
 					"type" : "SPECIAL_SPELL_LEV",
-                    "updater" : "TIMES_HERO_LEVEL",
+					"updater" : "TIMES_HERO_LEVEL",
 					"val" : 3
 				}
 			}
@@ -170,7 +170,7 @@
 				"meteorShower" : {
 					"subtype" : "spell.meteorShower",
 					"type" : "SPECIAL_SPELL_LEV",
-                    "updater" : "TIMES_HERO_LEVEL",
+					"updater" : "TIMES_HERO_LEVEL",
 					"val" : 3
 				}
 			}
@@ -238,7 +238,7 @@
 				"animateDead" : {
 					"subtype" : "spell.animateDead",
 					"type" : "SPECIAL_SPELL_LEV",
-                    "updater" : "TIMES_HERO_LEVEL",
+					"updater" : "TIMES_HERO_LEVEL",
 					"val" : 3
 				}
 			}
--- a/config/heroes/necropolis.json
+++ b/config/heroes/necropolis.json
@@ -11,36 +11,7 @@
 			{ "skill" : "resistance", "level": "basic" }
 		],
 		"specialty" : {
-			"base" : {
-				"limiters" : [
-					{
-						"parameters" : [ "walkingDead", true ],
-						"type" : "CREATURE_TYPE_LIMITER"
-					}
-				]
-			},
-			"bonuses" : {
-				"attack" : {
-					"subtype" : "primSkill.attack",
-					"type" : "PRIMARY_SKILL",
-					"updater" : {
-						"parameters" : [ 5, 2 ],
-						"type" : "GROWS_WITH_LEVEL"
-					}
-				},
-				"defence" : {
-					"subtype" : "primSkill.defence",
-					"type" : "PRIMARY_SKILL",
-					"updater" : {
-						"parameters" : [ 5, 2 ],
-						"type" : "GROWS_WITH_LEVEL"
-					}
-				},
-				"speed" : {
-					"type" : "STACKS_SPEED",
-					"val" : 1
-				}
-			}
+			"creature" : "walkingDead"
 		}
 	},
 	"vokial":
@@ -55,36 +26,7 @@
 			{ "skill" : "artillery", "level": "basic" }
 		],
 		"specialty" : {
-			"base" : {
-				"limiters" : [
-					{
-						"parameters" : [ "vampire", true ],
-						"type" : "CREATURE_TYPE_LIMITER"
-					}
-				]
-			},
-			"bonuses" : {
-				"attack" : {
-					"subtype" : "primSkill.attack",
-					"type" : "PRIMARY_SKILL",
-					"updater" : {
-						"parameters" : [ 10, 4 ],
-						"type" : "GROWS_WITH_LEVEL"
-					}
-				},
-				"defence" : {
-					"subtype" : "primSkill.defence",
-					"type" : "PRIMARY_SKILL",
-					"updater" : {
-						"parameters" : [ 9, 4 ],
-						"type" : "GROWS_WITH_LEVEL"
-					}
-				},
-				"speed" : {
-					"type" : "STACKS_SPEED",
-					"val" : 1
-				}
-			}
+			"creature" : "vampire"
 		}
 	},
 	"moandor":
@@ -99,36 +41,7 @@
 			{ "skill" : "learning", "level": "basic" }
 		],
 		"specialty" : {
-			"base" : {
-				"limiters" : [
-					{
-						"parameters" : [ "lich", true ],
-						"type" : "CREATURE_TYPE_LIMITER"
-					}
-				]
-			},
-			"bonuses" : {
-				"attack" : {
-					"subtype" : "primSkill.attack",
-					"type" : "PRIMARY_SKILL",
-					"updater" : {
-						"parameters" : [ 13, 5 ],
-						"type" : "GROWS_WITH_LEVEL"
-					}
-				},
-				"defence" : {
-					"subtype" : "primSkill.defence",
-					"type" : "PRIMARY_SKILL",
-					"updater" : {
-						"parameters" : [ 10, 5 ],
-						"type" : "GROWS_WITH_LEVEL"
-					}
-				},
-				"speed" : {
-					"type" : "STACKS_SPEED",
-					"val" : 1
-				}
-			}
+			"creature" : "lich"
 		}
 	},
 	"charna":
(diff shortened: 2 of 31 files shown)


lib/HeroBonus.cpp
```diff
@@ -1496,3 +1609,89 @@ void LimiterList::add( TLimiterPtr limiter )
...
+
+	root["type"].String() = "GROWS_WITH_LEVEL";
+	root["parameters"].Vector().push_back(JsonUtils::intNode(valPer20));
+	if(stepSize > 1)
+		root["parameters"].Vector().push_back(JsonUtils::intNode(stepSize));
+
+	return root;
+}
+
+std::string nameForBonus(const Bonus & bonus)
```
alexvins: Why this function is not part of Bonus class?

lib/HeroBonus.h
```diff
@@ -986,3 +1001,42 @@ void BonusList::insert(const int position, InputIterator first, InputIterator la
...
+	virtual bool update(Bonus & b, const CBonusSystemNode & context) const = 0;
+	virtual std::string toString() const;
+	virtual JsonNode toJsonNode() const = 0;
+
+	template <typename Handler> void serialize(Handler &h, const int version)
+	{
+	}
+};
+
+struct DLL_LINKAGE ScalingUpdater : public IUpdater
```
alexvins: As long as `IUpdater` declared with `class` descendants should be consistent.

lib/JsonNode.cpp
```diff
@@ -214,6 +214,49 @@ bool JsonNode::isNumber() const
 	return type == DATA_INTEGER || type == DATA_FLOAT;
 }
 
+bool JsonNode::isEmpty() const
+{
+	switch(type)
+	{
+		case DATA_NULL:
+			return true;
+		case DATA_STRUCT:
```
alexvins: This should handle vector type node too.

lib/mapObjects/CGHeroInstance.h
```diff
@@ -108,7 +108,8 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
 		}
 	};
 
-	std::vector<HeroSpecial*> specialty;
+	std::vector<HeroSpecial*> specialtyDeprecated;
+	CBonusSystemNode specialty;
```
alexvins: Is this field used?

lib/HeroBonus.cpp
```diff
@@ -1095,6 +1117,60 @@ std::string Bonus::Description() const
 	return str.str();
 }
 
+JsonNode subtypeToJson(Bonus::BonusType type, int subtype)
+{
+	switch(type)
```
alexvins: We should find more generic solution for encoding sybtype and addInfo

lib/HeroBonus.cpp
henningkoehlernz: My only excuse is that I must have been sleepier than I thought. Fixed.

lib/JsonNode.cpp
henningkoehlernz: I don't consider a vector with no elements to be "empty" on purpose here. The reason is that the intersection of two json nodes is the empty vector (rather than null) only if both are the empty vector. So when constructing a "base" object as intersection of nodes we want to preserve [ ] in the intersection, but not { } which (typically) will be the result of intersecting two distinct structs.
The name "isEmpty" may not be ideal here.

lib/HeroBonus.cpp
henningkoehlernz: Agreed. But right now it's only used for converting legacy format (and still subject to manual copy-paste), so didn't want to cram a bigger change into this branch.

lib/HeroBonus.h
henningkoehlernz: fixed.

lib/mapObjects/CGHeroInstance.h
henningkoehlernz: no, removed (both specialtyDeprecated and specialty).

lib/HeroBonus.h
```diff
@@ -986,3 +1000,40 @@ void BonusList::insert(const int position, InputIterator first, InputIterator la
...
+class DLL_LINKAGE IUpdater
+{
+public:
+	virtual ~IUpdater();
+
+	virtual const std::shared_ptr<Bonus> update(const std::shared_ptr<Bonus> b, const CBonusSystemNode & context) const = 0;
+	virtual std::string toString() const;
+	virtual JsonNode toJsonNode() const = 0;
+
+	template <typename Handler> void serialize(Handler &h, const int version)
```
alexvins: missing space after `&`.

lib/HeroBonus.h
```diff
@@ -986,3 +1000,40 @@ void BonusList::insert(const int position, InputIterator first, InputIterator la
...
+class DLL_LINKAGE ScalingUpdater : public IUpdater
+{
+public:
+	int valPer20;
+	int stepSize;
+
+	ScalingUpdater();
+	ScalingUpdater(int valPer20, int stepSize = 1);
+
+	template <typename Handler> void serialize(Handler &h, const int version)
```
alexvins: missing space after `&`.

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -765,6 +539,23 @@ void CGHeroInstance::recreateSecondarySkillsBonuses()
 			updateSkill(SecondarySkill(skill_info.first), level);
 }
 
+void CGHeroInstance::recreateSpecialtyBonuses(std::vector<HeroSpecial*> & specialtyDeprecated)
```
alexvins: Missing space before `*`

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -765,6 +539,23 @@ void CGHeroInstance::recreateSecondarySkillsBonuses()
...
+{
+	auto HeroSpecialToSpecialtyBonus = [](HeroSpecial & hs) -> SSpecialtyBonus
+	{
+		SSpecialtyBonus sb;
+		sb.growsWithLevel = hs.growsWithLevel;
+		sb.bonuses = hs.getBonusList();
```
alexvins: Should be explicit copying.

lib/CHeroHandler.cpp
```diff
@@ -379,31 +379,267 @@ void CHeroHandler::loadHeroSkills(CHero * hero, const JsonNode & node)
...
+		{
+			const auto &creatures = VLC->creh->creatures;
+			bonus->type = Bonus::SPECIAL_UPGRADE;
+			bonus->subtype = spec.subtype; //base id
+			bonus->additionalInfo = spec.additionalinfo; //target id
+			result.push_back(bonus);
+			//propagate for regular upgrades of base creature
+			for(auto cre_id : creatures[spec.subtype]->upgrades)
+			{
+				bonus = std::make_shared<Bonus>(*bonus);
```
alexvins: Copy construction variable from itself is confusing.

lib/CHeroHandler.cpp
```diff
@@ -379,31 +379,267 @@ void CHeroHandler::loadHeroSkills(CHero * hero, const JsonNode & node)
...
+
+	//deprecated, used only for original specialties
+	const JsonNode & specialtiesNode = node["specialties"];
+	if (!specialtiesNode.isNull())
+	{
+		logMod->warn("Hero %s has deprecated specialties format.", hero->identifier);
+		for(const JsonNode &specialty : specialtiesNode.Vector())
+		{
+			SSpecialtyInfo spec;
+			spec.type = specialty["type"].Float();
```
alexvins: Should be `.Integer()`

lib/JsonNode.cpp
```diff
@@ -603,6 +647,23 @@ bool JsonUtils::parseBonus(const JsonNode &ability, Bonus *b)
 	if (!value->isNull())
 		b->propagator = parseByMap(bonusPropagatorMap, value, "propagator type ");
 
+	value = &ability["updater"];
+	if (!value->isNull())
```
alexvins: Remove space after `if`

config/heroes/necropolis.json
```diff
@@ -201,10 +411,15 @@
 			{ "skill" : "necromancy", "level": "basic" },
 			{ "skill" : "mysticism", "level": "basic" }
 		],
-		"specialties":
-		[
-			{ "type":3, "val": 39, "subtype": 0, "info": 3 }
-		]
+		"specialty" : {
+			"bonuses" : {
+				"summonBoat" : {
```
alexvins: This is wrong. Should be `animateDead` +3% per level

config/heroes/special.json
```diff
@@ -162,11 +287,23 @@
...
+						"type" : "HAS_ANOTHER_BONUS_LIMITER"
+					}
+				],
+				"subtype" : "primSkill.attack",
+				"type" : "PRIMARY_SKILL",
+				"val" : 1
+			},
+			"bonuses" : {
+				"attack" : {  },
+				"attack2" : {  }
```
alexvins: What is this? Two identical bonuses?

alexvins: Should be +5 attack and +5 defence

FeniksFire:
I don't like when someone write senseless comments, so can you try remove although the part of them?

FeniksFire:
I don't like when someone write senseless comments, so can you try remove although the part of them?

DjWarmonger:
```
The following updaters are currently available:
GrowsWithLevelUpdater ("GROWS_WITH_LEVEL"). Parameters are valPer20 and stepSize, and it updates val to valPer20 * ceil(level / stepSize) / 20. StepSize is optional and defaults to 1.
```

Unfortunately, this is incorrect for creatures which have different stats in base and upgraded form. Plus, its superflouous and overcomplicated while the old code calculated the value on the fly. In particular, for creatures introduced in mods.

ArseniyShestakov:
@henningkoehlernz am I getting it right that you implemented what Warmonger posted about?

I wasn't very much into VCMI until January, but since I now back and you rebased it I guess we can finally finish merge this branch as well. Any ideas if this will in any way conflict with SpellsRefactoring9?

henningkoehlernz:
Yes, latest change should address the concerns Warmonger had. Will check for merge conficts with SpellsRefactoring.

DjWarmonger:
Okay, I will just merge it after I do some tests.

henningkoehlernz:
W.r.t. the merge conflict in NetPacksLib.cpp, note that I renamed CBonusSystemNode.updateBonuses into CbonusSystemNode.reduceBonusDurations, as the name updateBonuses became very misleading with the introduction of bonus updaters.

Global.h
```diff
@@ -369,6 +369,18 @@ namespace vstd
 		return std::find(c.begin(),c.end(),i);
 	}
 
+	//returns first key that maps to given value if present, returns success via found if provided
+	template <typename Key, typename T>
+	Key findKey(const std::map<Key, T> & map, const T & value, bool * found = nullptr)
```
alexvins: `found` not used

henningkoehlernz: fixed.

config/heroes/special.json
```diff
@@ -25,11 +29,9 @@
...
 		[
 			{ "skill" : "wisdom", "level": "basic" },
-			{ "skill" : "fireMagic", "level": "expert" } ],
-		"specialties":
-		[
-			{ "type":11, "val": 14, "subtype": 0, "info": 3 }
-		]
+			{ "skill" : "fireMagic", "level": "expert" }
+		],
+		"specialty" : { "bonuses" : {  } }
```
alexvins: Something missing here

henningkoehlernz: No, that's correct. Adrienne's "specialty" is that she starts with expert fire magic.
http://heroes.thelazy.net/wiki/Adrienne

config/heroes/tower.json
```diff
@@ -136,10 +142,15 @@
...
-		"specialties":
-		[
-			{ "type":3, "val": 3, "subtype": 60, "info": 0 }
-		]
+		"specialty" : {
+			"bonuses" : {
+				"hypnotize" : {
+					"subtype" : "spell.hypnotize",
+					"type" : "SPECIAL_SPELL_LEV",
+					"val" : 3
```
alexvins: This bonus is used in `CGHeroInstance::getSpellBonus` and have kind of TIMES_HERO_LEVEL updater there,
```
valOfBonuses(Bonus::SPECIAL_SPELL_LEV, spell->getIndex()) * level
```
 could it be converted?

config/heroes/special.json
alexvins: OK, then comment is missing :)

henningkoehlernz: I agree, but I don't think the Json parser permits comments?

alexvins: `//`-style comments allowed

config/heroes/tower.json
henningkoehlernz: Yes, could move the level out of that formula and add TIMES_HERO_LEVEL updater, but is there a need for such a bonus that doesn't scale with level?

config/heroes/special.json
henningkoehlernz: Right, just noticed there are comments like this right at the beginning of the file. Will add.

config/heroes/tower.json
alexvins: It might be useful for modders. I think mostly about unification now. Other bonuses near in code do not have explicit scaling.

alexvins: Same applies to SPECIAL_BLESS_DAMAGE.

lib/HeroBonus.cpp
```diff
@@ -1543,3 +1705,138 @@ void LimiterList::add( TLimiterPtr limiter )
...
+		return newBonus;
+	}
+	return b;
+}
+
+std::string GrowsWithLevelUpdater::toString() const
+{
+	char buf[100];
+	sprintf(buf, "GrowsWithLevelUpdater(valPer20=%d, stepSize=%d)", valPer20, stepSize);
+	return std::string(buf);
```
ArseniyShestakov:
```
return boost::str(boost::format("GrowsWithLevelUpdater(valPer20=%d, stepSize=%d)") % valPer20 % stepSize);
```
Something like that should work? We use Boost anyway.

lib/HeroBonus.h
```diff
@@ -362,6 +366,8 @@ struct DLL_LINKAGE Bonus : public std::enable_shared_from_this<Bonus>
 		h & effectRange;
 		h & limiter;
 		h & propagator;
+		if(version >= 781)
+			h & updater;
```
ArseniyShestakov: We come up with idea that it's could be good for serialization code to always have brackets, even for single line.

https://wiki.vcmi.eu/Coding_guidelines#Serialization

config/heroes/tower.json
henningkoehlernz: Done.

DjWarmonger:
Adela's bonus value is correct, but affects only max damage and not all damage. 
adela dmg bonus 2
adela dmg bonus

DjWarmonger:
This seems to be an issue with develop, tho.

ArseniyShestakov:
Finally we can merge this one!

PR #382, opened 2017-09-28
Fixing bugs for external dwelling bonus of Golem Factory.

bigapple10027:
[Problem]
  The occupied Golem Factory on adventure map doesn't give
  the player corresponding weekly growth bonus.

[Solution]
  1. Add the dwelling (of multiple generators) to a player's dewelling set.
  2. Created helper function to calculate external dwelling bonus
  points.

[Test]
  Manual testing on the map, works fine.

https://bugs.vcmi.eu/view.php?id=2789

### DIFF
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -157,7 +157,7 @@ void CGDwelling::setPropertyDer(ui8 what, ui32 val)
 	switch (what)
 	{
 		case ObjProperty::OWNER: //change owner
-			if (ID == Obj::CREATURE_GENERATOR1) //single generators
+			if (ID == Obj::CREATURE_GENERATOR1 || ID == Obj::CREATURE_GENERATOR4)
 			{
 				if (tempOwner != PlayerColor::NEUTRAL)
 				{
@@ -546,9 +546,7 @@ GrowthInfo CGTownInstance::getGrowthInfo(int level) const
 	int dwellingBonus = 0;
 	if(const PlayerState *p = cb->getPlayer(tempOwner, false))
 	{
-		for(const CGDwelling *dwelling : p->dwellings)
-			if(vstd::contains(creatures[level].second, dwelling->creatures[0].second[0]))
-				dwellingBonus++;
+		dwellingBonus = getDwellingBonus(creatures[level].second, p->dwellings);
 	}
 
 	if(dwellingBonus)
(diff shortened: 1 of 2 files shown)


lib/mapObjects/CGTownInstance.cpp
```diff
@@ -570,6 +568,17 @@ GrowthInfo CGTownInstance::getGrowthInfo(int level) const
 	return ret;
 }
 
+int CGTownInstance::getDwellingBonus(const std::vector<CreatureID>& creatureIds, const std::vector<ConstTransitivePtr<CGDwelling> >& dwellings) const
+{
+	return std::accumulate(dwellings.cbegin(), dwellings.cend(), 0,
+			[&creatureIds] (int count, decltype(dwellings[0])& dwelling) -> int {
```
alexvins: 1) too long statement, please declare at least outer lambda as variable
2) fix formatting https://wiki.vcmi.eu/Coding_guidelines

bigapple10027: 1. I simplified the code by replacing the lambda with `for` loops. Let's make the code more readable.
2.  I read the coding style guidelines on the wiki, however, I don't follow why putting space in-between `if` (or `for`) and `(` is a bad style? Both Mono-project (which is the main source the wiki refers to) and [Google C++ coding style](https://google.github.io/styleguide/cppguide.html#Conditionals) list it as good.
3. Can we add some automatic code style checking script upon pull request? I'd be happy to do some investigation on this.

alexvins: 1. Thanks
2. We have our guidelines. We open for discussion of improving them.
3. Planned, see #347

ArseniyShestakov:
Shouldn't bonus be given for  ```CREATURE_GENERATOR2``` and ```CREATURE_GENERATOR3```?

bigapple10027:
Hi @ArseniyShestakov ,

I'm glad you brought this up. Could you tell me what's the difference for each of them? I assume it's dwelling that can generate 1, 2, 3, 4 different types of creatures? I'd like to add some comments for these constants in this pull request :)

ArseniyShestakov:
I thinks so, but to be fair after checking the code I not even sure if this is actually needed.
```CREATURE_GENERATOR3``` likely only used for war machine factory and what  ```CREATURE_GENERATOR2```  used for I have no idea.

Mods probably just use dwellings handler and not use these IDs, but this need to be checked.

bigapple10027:
Yeah, I searched both `CREATURE_GENERATOR3` and `CREATURE_GENERATOR2` in this repository and they're actually only used [here](https://github.com/vcmi/vcmi/blob/5067e73c30158cee392413c492bb7f626a90d7ae/lib/mapping/MapFormatH3M.cpp#L1247).

ArseniyShestakov:
I think we can safely just add them to condition just in case regardless of how useful it is.
Might be someone will make a mod that let you build war machines and they'll have growth.

bigapple10027:
Cool, will publish another revision later today. Thanks for the comment :)

PR #383, opened 2017-10-05
Text fixes 

godric3:
Fixes some small text bugs like missing artifact name or bonus description

https://bugs.vcmi.eu/view.php?id=2763
https://bugs.vcmi.eu/view.php?id=2740
https://bugs.vcmi.eu/view.php?id=2737
https://bugs.vcmi.eu/view.php?id=2731


ArseniyShestakov:
I tested 2763 and 2740, but code is seems to be correct for another two.
Thanks for contributing!

I gave you proper permissions on bug tracker so if you want to work on something else feel free to assign issues to yourself.

PR #384, opened 2017-10-09
fix for banks(crypt, ships) not giving morale penalty

godric3:
also texts should be closer to original now: on visit we can decide if we want to search bank, if already cleared some of them give morale penalty

fix https://bugs.vcmi.eu/view.php?id=2756

### DIFF
--- a/lib/mapObjects/CBank.cpp
+++ b/lib/mapObjects/CBank.cpp
@@ -95,63 +95,39 @@ bool CBank::wasVisited (PlayerColor player) const
 	return !bc; //FIXME: player A should not know about visit done by player B
 }
 
-void CBank::onHeroVisit (const CGHeroInstance * h) const
+void CBank::onHeroVisit(const CGHeroInstance * h) const
 {
-	if (bc)
+	int banktext = 0;
+	ui16 soundID = soundBase::ROGUE;
+	switch (ID)
 	{
-		int banktext = 0;
-		ui16 soundID = soundBase::ROGUE;
-		switch (ID)
-		{
-		case Obj::CREATURE_BANK:
-			banktext = 32;
-			break;
-		case Obj::DERELICT_SHIP:
-			banktext = 41;
-			break;
-		case Obj::DRAGON_UTOPIA:
-			banktext = 47;
-			break;
-		case Obj::CRYPT:
-			banktext = 119;
-			break;
-		case Obj::SHIPWRECK:
-			banktext = 122;
-			break;
-		case Obj::PYRAMID:
-			soundID = soundBase::MYSTERY;
-			banktext = 105;
-			break;
-		}
-		BlockingDialog bd (true, false);
-		bd.player = h->getOwner();
-		bd.soundID = soundID;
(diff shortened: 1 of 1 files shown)


lib/mapObjects/CBank.cpp
```diff
@@ -262,32 +265,35 @@ void CBank::doVisit(const CGHeroInstance * hero) const
 			std::set<SpellID> spells;
 
 			bool noWisdom = false;
+			if(textID == 106){
```
alexvins: Put `{` on new line

lib/mapObjects/CBank.cpp
```diff
@@ -95,124 +95,127 @@ bool CBank::wasVisited (PlayerColor player) const
...
-		break;
-	case Obj::PYRAMID:
-		textID = 106;
+		case Obj::CREATURE_BANK:
+		case Obj::DRAGON_UTOPIA:
+		default:
+			iw.text << VLC->generaltexth->advobtxt[33];// This was X, now is completely empty
+			iw.text.addReplacement(getObjectName());
+		}
+		if (textID != -1) {
```
alexvins: Put `{` on new line

lib/mapObjects/CBank.cpp
```diff
@@ -95,124 +95,127 @@ bool CBank::wasVisited (PlayerColor player) const
...
-		if (bc)
-			textID = 124;
-		else
-			textID = 123;
-		break;
-	case Obj::PYRAMID:
-		textID = 106;
+		case Obj::CREATURE_BANK:
+		case Obj::DRAGON_UTOPIA:
+		default:
```
alexvins: What is the point of such case label? Should it better to be just `default`?

godric3: Just readability, searching for creature bank, or utopia code will be easier with such case.
Not sure if default is needed, can it occur?

alexvins: (Modded) Creature bank can have any ID, it defined by `handler` field.

godric3: So shouldn't `default` be added to two other switch statements, with some generic text? lines 102 and 142

alexvins: I think it should

godric3: fixed braces, and` default`

lib/mapObjects/CBank.cpp
```diff
@@ -95,124 +95,130 @@ bool CBank::wasVisited (PlayerColor player) const
...
+	default:
+		banktext = 32;
+		break;
+	}
+	BlockingDialog bd(true, false);
+	bd.player = h->getOwner();
+	bd.soundID = soundID;
+	bd.text.addTxt(MetaString::ADVOB_TXT, banktext);
+	if (ID == Obj::CREATURE_BANK)
+		bd.text.addReplacement(getObjectName());
```
alexvins: default case also use text 32 and should use object name replacement

godric3: fixed

PR #386, opened 2017-10-26
Add keyboard shortcuts for splitting army from HD mod.

FeniksFire:
Add keyboard shortcuts for quick army management from HD mod.

Just want add fast some really nice shortcuts from HD mod.
1) [LShift] + LClick – splits a half units from the selected stack into
an empty slot. WARNING: The second behaviour ("smart split, ignores
single creatures") of this HD mod option is not done.
2) [LCtrl] + LClick – splits a single unit from the selected stack into
an empty slot.
3) [LCtrl] + [LShift] + LClick – split single units from the selected
stack into all empty hero/garrison slots

### DIFF
--- a/client/widgets/CGarrisonInt.cpp
+++ b/client/widgets/CGarrisonInt.cpp
@@ -281,7 +281,10 @@ void CGarrisonSlot::clickLeft(tribool down, bool previousState)
 		bool refr = false;
 		const CGarrisonSlot * selection = owner->getSelection();
 		if(!selection)
+		{
 			refr = highlightOrDropArtifact();
+			handleSplittingShortcuts();
+		}
 		else if(selection == this)
 			refr = viewInfo();
 		// Re-highlight if troops aren't removable or not ours.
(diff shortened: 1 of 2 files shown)


FeniksFire:
If someone need, here is short video showing how it's work: https://www.dropbox.com/s/u54o6j5ddajv69h/SplittingShortcuts.mp4?dl=0

dydzio0614:
What about OH3 shift+click shortcut? As far as I remember it opens stack split window with selected slot

FeniksFire:
@dydzio quote from [hd mod site](https://sites.google.com/site/heroes3hd/eng/description/extended-ui)

> This option does not disable the default creature split: to do the default creature split with [Shift], you need to first select the stack, and then press shift .

Uhh and on the ocassion I see that I make a little bad behavior for shift shortcut, originally it should take into consideration all stack in all slots ;d, when I make that it only split one stack into half amount each time. I will correct it soon ;d.

alexvins:
It crashes on split. Select stack, press SHIFT select another position. Stack split 50/50 instantly and split window also opened, after split closing window crash

FeniksFire:
@alexvins thanks for report, should be fine now.

PR #392, opened 2017-11-17
Siege tower handling improvements

dydzio0614:
Fixed bugs #2581 and #2495 + somewhat improved damage calculation


server/CGameHandler.cpp
```diff
@@ -854,7 +854,7 @@ void CGameHandler::prepareAttack(BattleAttack &bat, const CStack *att, const CSt
 {
 	bat.bsa.clear();
 	bat.stackAttacking = att->ID;
-	const int attackerLuck = att->LuckVal();
+	const int attackerLuck = (att->getCreature()->idNumber != CreatureID::ARROW_TOWERS) ? att->LuckVal() : 0;
```
alexvins: You should add NO_LUCK bonus to arrowTower instead

PR #393, opened 2017-11-26
Change JsonType to enum class

FeniksFire:
Change `enum JsonType` to `enum class JsonType`


alexvins:
I see no benefits from this change. It might be better to pull JsonType to global namespace but still no profit. JsonType is the only enumeration inside JsonNode. nothing to mix up with.

FeniksFire:
And the benefits are quite a lot, that why C++11 introduced them. We probably should have only class enums and get rid of the old ones.

alexvins:
OK, accepted as quality improvement.

PR #394, opened 2017-12-20
Quick Recruitment Window upgrade

FeniksFire:
Now is possible to switch by button between creatures upgrades.


client/windows/CreaturePurhaseCard.cpp
```diff
@@ -15,21 +15,43 @@
...
+}
+
+void CreaturePurhaseCard::switchCreatureLevel()
+{
+	OBJECT_CONSTRUCTION_CAPTURING(ACTIVATE + DEACTIVATE + UPDATE + SHOWALL);
+	auto id = vstd::find(upgradesID, creatureOnTheCard->idNumber);
+	if(creatureOnTheCard->idNumber == upgradesID.back())
+		id = upgradesID.begin();
+	else
+		id++;
```
alexvins: What does `++` mean for creature id? [Upgrade id] =[Downgrade id] + 1 is true only for OH3 creatures by accident. It may not work with mods, (especially alternative upgrades)

client/windows/CreaturePurhaseCard.cpp
```diff
@@ -15,21 +15,43 @@
...
+	picture = std::make_shared<CCreaturePic>(pos.x, pos.y, VLC->creh->creatures[*id]);
+	picture->moveTo(Point(pos.x + parent->pos.x, pos.y + parent->pos.y));
+	creatureOnTheCard = VLC->creh->creatures[*id];
```
alexvins: do not use `VLC->creh->creatures` directly, use `CreatureID::toCreature()` instead.

client/windows/CreaturePurhaseCard.cpp
```diff
@@ -15,21 +15,43 @@
...
 {
 	initMaxButton();
 	initMinButton();
+	initCreatureSwitcherButton();
 }
 
 void CreaturePurhaseCard::initMaxButton()
 {
-	maxButton = std::make_shared<CButton>(Point(pos.x + 52, pos.y + 178), "iam014.def", CButton::tooltip(), std::bind(&CSlider::moveToMax,slider), SDLK_m);
+	maxButton = std::make_shared<CButton>(Point(pos.x + 52, pos.y + 178), "iam014.def", CButton::tooltip(), std::bind(&CSlider::moveToMax,slider));
```
alexvins: `initMaxButton` and `initMaxButton` looks like 99% code duplication, please consider refactoring as you already tweaking them.

client/windows/CreaturePurhaseCard.cpp
```diff
@@ -15,21 +15,43 @@
...
+void CreaturePurhaseCard::switchCreatureLevel()
+{
+	OBJECT_CONSTRUCTION_CAPTURING(ACTIVATE + DEACTIVATE + UPDATE + SHOWALL);
+	auto id = vstd::find(upgradesID, creatureOnTheCard->idNumber);
+	if(creatureOnTheCard->idNumber == upgradesID.back())
+		id = upgradesID.begin();
+	else
+		id++;
+	picture = std::make_shared<CCreaturePic>(pos.x, pos.y, VLC->creh->creatures[*id]);
+	picture->moveTo(Point(pos.x + parent->pos.x, pos.y + parent->pos.y));
```
alexvins: This should be handled by inheriting parent position aka capture SHARE_POS

client/windows/CreaturePurhaseCard.cpp
FeniksFire: It's works for all mods, cove, forge, rampart from alternative upgrades also checked. So let look closer to it. First I copy into every "CreaturePurhaseCard" upgradesID that will store all id that the creature can be improved. Next, here I create iterator that have name "id". For it, ++ means only that I get next element from vector, in that case from "upgradesID". So nope, I don't get id+1. I get next id from upgradesID.

client/windows/CreaturePurhaseCard.cpp
FeniksFire: Huh, just want delete SDLK_m shortcut at end :). By mistake, I put it once time and now want remove.

client/windows/CreaturePurhaseCard.cpp
alexvins: I missed that `id` is an iterator.

alexvins: Will it more clear to do like that:
` auto index = vstd::find_pos(upgradesID, creatureOnTheCard->idNumber)`
`auto nextCreatureId = vstd::circular_at(upgradesID, ++index)`
?

client/windows/CreaturePurhaseCard.cpp
alexvins: BTW, could we add proper shortcuts for that, f.e. {CTRL|SHIFT] + <number> to min|max card <number>?

client/windows/CreaturePurhaseCard.cpp
FeniksFire: Did you mean vstd::circularAt? Yeah it's seems to be nicer solution.

alexvins: circularAt

client/windows/CreaturePurhaseCard.cpp
FeniksFire: seems nice, done!

client/windows/QuickRecruitmentWindow.cpp
```diff
@@ -53,8 +53,10 @@ void QuickRecruitmentWindow::setCreaturePurhaseCards()
 	{
 		if(!town->town->creatures.at(i).empty() && !town->creatures.at(i).second.empty() && town->creatures[i].first)
 		{
-			CreatureID crid = town->creatures[i].second[town->creatures[i].second.size() - 1];
-			cards.push_back(std::make_shared<CreaturePurhaseCard>(VLC->creh->creatures[crid], position, town->creatures[i].first, this));
+			std::vector<CreatureID> crids;
+			for(auto j : town->creatures[i].second)
+				crids.push_back(j);
+			cards.push_back(std::make_shared<CreaturePurhaseCard>(crids, position, town->creatures[i].first, this));
```
alexvins: No need to copy `town->creatures[i].second`, pass constant reference to CreaturePurhaseCard ctor.

FeniksFire: heheh, what a stupid mistake, fixed

client/windows/CreaturePurchaseCard.h
```diff
@@ -17,22 +17,23 @@ class CButton;
...
 {
 public:
 	const CCreature * creatureOnTheCard;
 	std::shared_ptr<CSlider> slider;
 	QuickRecruitmentWindow * parent;
 	int maxAmount;
 	void sliderMoved(int to);
-	CreaturePurhaseCard(const CCreature * creature, Point position, int creaturesMaxAmount, QuickRecruitmentWindow * parents);
-
+	CreaturePurchaseCard(std::vector<CreatureID> creaturesID, Point position, int creaturesMaxAmount, QuickRecruitmentWindow * parents);
```
alexvins: this need to be `const std::vector<CreatureID> &`

FeniksFire: fixed

PR #395, opened 2017-12-22
Adding simple background for quick recruitment window

FeniksFire:
Some people says that want improve graphics for creature block in quick recruitment window. Just want give them such option, not sure if this is good way/how should be graphics added to vcmi. So if this is bad way, give me a guideline how to do it.


alexvins:
Graphic assets should have sources too, we store them in vcmi-assets repo.

FeniksFire:
Thanks, already maked pr for vcmi-assets!

alexvins:
New picture is OK, but it reveals that many gui elements are not centred.
1) upgrade switch is off to right by 2 pixels
2) available amount label is off to right by about 4 pixels
3) amount labels should be vertically aligned
4) cost widget is off to left by 2 pixels

FeniksFire:
@alexvins I do it myself and don't have xcf, but anyway, the update will come soon since someone more skilled do it today. He just need possibility of replacing graphics to looking for results.

PR #401, opened 2018-01-04
Fix line endings

janisozaur:
This is purely a whitespace change to fix all the line endings stored in the repository.


janisozaur:
Why not?

dydzio0614:
@DjWarmonger Same. For example generated visual studio project files also got changed here ^^

janisozaur:
@dydzio0614 yes, it was indeed, intentionally.

dydzio0614:
Did you change line endings from Carriage Return + LF to LF or?

janisozaur:
@dydzio0614 yes, CRLF->LF

alexvins:
We should make line endings consistent, but do not for all files at once.

janisozaur:
@alexvins what do you suggest then?

DjWarmonger:
LF is a standard for Linux and MacOS, CRLF is a standard for Windows. So this effort is futile and should be handled by Git.

DjWarmonger:
In particular converting MSVS files to Linux format is very wrong.

janisozaur:
@DjWarmonger 

> this effort is futile and should be handled by Git.

Yes, it would be, had you stored the files with native encoding (LF) in the first place. Since you haven't, upon cloning the files checked out into working tree is created with wrong line ending on _every platform but Windows_.

> In particular converting MSVS files to Linux format is very wrong.

It is not. Like you pointed out -- and then immediately forgot about in next comment -- line conversion should be handled by git. MSVC handles LF files just fine, but it doesn't matter as they will be checked out as CRLF anyways (these are the defaults in the most git clients I've seen). This is much more portable way of working than storing CRLF in the repository, especially if there is an occasional need to modify MSVC project files outside of Windows.

Please see the documentation: https://git-scm.com/docs/gitattributes#_end_of_line_conversion

alexvins:
@janisozaur split changes to several PR
@DjWarmonger git handles line endings fine. With `autocrlf=true` I even forgot that we have such a mess in repository :)

DjWarmonger:
I just checked and my client *does* have autocrlf turned on by default.

janisozaur:
@alexvins that would just waste all the CI time\* and I don't see the benefits it would provide. It's easier to skip over one such commit than over several ones when doing just about anything related to the history.

If you still insist, I can split it into several chunks, but please provide some guidance on how to split it. Per directory? Per file type? Something else?

\* There is `[ci skip]`, but for the initial version it's good to prove everything does indeed work. And the other point still holds.

FeniksFire:
I don't understand what is benefit from it.

janisozaur:
@FeniksFire to let git handle line ending conversion so that the files are checked out with proper endings on each platform.

The Travis CI build failed because one of the jobs timed out, can you please restart it?

FeniksFire:
So still is question that same, what is benefit from it, it was never a problem that git don't handle line ending conversion.

ArseniyShestakov:
What I wonder about is the same problem as with my formatting experiment. Is there reliable way of avoiding all these changes to appear in let's say "git blame" by default and without rewriting history?

If no then I'm strongly against merging anything like that. And yeah I'm Linux user, but I never actually had real issues with all this CRLF mess and I certainly don't see why it's more important than wasting my time when I want to look back into history.

At least we can eventually do something like that alongside with formatting changes.

janisozaur:
> Is there reliable way of avoiding all these changes to appear in let's say "git blame"

`-w` explicitly skips over whitespace changes.

> [`Ignore whitespace when comparing the parent’s version and the child’s to find where the lines came from.`](https://git-scm.com/docs/git-blame#git-blame--w)

Most other git utils follow.

ArseniyShestakov:
Problem is that Git integration into IDE isn't that flexible usually and I doubt let's say Qt Creator will let me adjust it. And I do use it a lot since of code in VCMI is far from self-explanatory.

So I simply don't see any reason why changing line endings actually worth it. Unlike formatting it's absolutely not affect work flow for most of developers.

janisozaur:
@ArseniyShestakov I don't usually use such utilities as they are overconfusing simple things, but I checked and Qt Creator has the option to ignore whitespace and it was even on by default in my case.

This is quite an essential feature and I would consider lack of support for it in a given client a bug which should be reported.

ArseniyShestakov:
If everyone else would confirms tools they use won't conflict with this change and this won't cause any issues with merges then this can be done. Though I would suggest to let this one stay open for a while simply because it's very low priority problem.

alexvins:
Merge tools can ignore whitespace, but I doubt that git itself can. So all active branches will get conflicts from this changes.

janisozaur:
Sadly I don't know whether or not this will invalidate the PRs, but fixing the possibly ensuing conflicts would be trivial. There are no non-whitespace changes here. I see there's another PR that's been brewing for a while and looks to be on the verge of getting merged: https://github.com/vcmi/vcmi/pull/359 if you want, I can redo this once this is in to reduce chance of conflicts.

And to answer the question raised by @FeniksFire and @ArseniyShestakov:
It will allow git to store the native line endings and fill the @DjWarmonger's [request of having git handle the conversion](https://github.com/vcmi/vcmi/pull/401/#issuecomment-355399488). Some tools on each of the platforms work less reliably or with degraded experience when encountering foreign line endings. E.g. the way vim presents CRLFs: https://i.stack.imgur.com/gv4Dd.png. There is not much more to it.

FeniksFire:
@janisozaur can you descibe what exactly do you mean by "less reliably or with degraded experience"?

ArseniyShestakov:
Ok. I perfectly understand reasoning behind this vast change even if it's very small. I don't like that ```^M``` either (even if I guess vim has option to ignore CRLF), but as I said above I'm not convinced amount of time we'll waste fixing PRs going to worth it.

Though if we eventually going to enforce some formatting on all code we can then as well get rid of CRLF.

@FeniksFire
I'll certainly ask any new contributor who going to push **new files** with CRLF to fix it.

FeniksFire:
The thing is, such requires they are terrible since everytime someone will forget about them he must fixing things. And not everyone have time for that.

PR #402, opened 2018-01-06
Reworked pregame for multiplayer

ArseniyShestakov:
Altered H3 behavior:
- Player is able to choose hero for the AI. H3 don't let you do that, but I see this as pointless limitation.

Some experimental / fun commands for the lobby chat:

- ```!passhost <connectionId>```
This will pass "host" privilege to another client. Server also automatically pass host privilege if client holding privilege is disconnected.
- ```!forcep <connectedPlayerId> <playerColorId>```
This will forcefully assign client player to in-game player of certain color.
Implemented for proof-of-concept cooperative gameplay since multiple clients will be able to control the same player colors. Though any changes in options will override this change since lobby code is not aware of possibility. So setup everything, leave one connectedPlayer unused, assign it to color you want with command and start the game.
Of course in-game UI will require significant amount of work to make it actually fun to play. For now camera centered moving hero, popups appear for both players and something will likely cause crashes, but you can still have some fun with it!


client/CPreGame.cpp
```diff
@@ -3167,37 +3151,35 @@ CMultiMode::CMultiMode()
...
-		GH.pushInt(new CSimpleJoinScreen(CMenuScreen::MULTI_NETWORK_HOST));
-	else
-		GH.pushInt(new CSelectionScreen(CMenuScreen::newGame, CMenuScreen::MULTI_NETWORK_HOST));
+	GH.pushInt(new CMultiPlayers(settings["general"]["playerName"].String(), CMenuScreen::MULTI_NETWORK_HOST, state));
 }
 
 void CMultiMode::joinTCP()
+{
+	GH.popIntTotally(this);
+	GH.pushInt(new CMultiPlayers(settings["general"]["playerName"].String(), CMenuScreen::MULTI_NETWORK_GUEST, state));
```
alexvins: How this even works? You accessing `state` field after destruction.

ArseniyShestakov: Oops, but somehow it does work for sure.

ArseniyShestakov:
@alexvins 
Don't worry this is just a random number from my head so I didn't checked what other branches using.
There still tons of work require to make it all work so I'll have tons of time to change it.

Not going to be merged before you branch anyway.

ArseniyShestakov:
So today I come to the point where I'm more or less happy changes on how client and server communicate. Code quality is still low, but I'll be happy for any suggestions on ```CServerHandler```, ```CVCMIServer``` and new lobby netpacks so if you have time take a look at them.

As of now I'll start to rework broken features for new code: campaigns, restarting, etc.

client/CMT.cpp
```diff
@@ -1330,50 +1337,26 @@ static void mainLoop()
...
-		{
-			elem.second.playerID = PlayerSettings::PLAYER_AI;
-			if(i < ais.size())
-				elem.second.name = ais[i++];
-		}
-	}
-
-    client = new CClient();
+	client = new CClient();
+	CSH->client = client;
```
alexvins: 1) startGame() should be moved to CServerHandler
2) no need to have static `client` pointer. Isn`t CSH->client enough?

client/pregame/CLobbyScreen.cpp
```diff
@@ -0,0 +1,173 @@
...
+	{
+
+	}
+}
+
+void CLobbyScreen::toggleMode(bool host)
+{
+	auto getColor = [=]()
+	{
+		return host ? Colors::WHITE : Colors::ORANGE;
```
alexvins: remove lambda, just save colour in local variable

client/pregame/CSavingScreen.cpp
```diff
@@ -0,0 +1,107 @@
...
+	curTab = tabSel;
+
+	tabSel->onSelect = std::bind(&CSavingScreen::changeSelection, this, _1);
+	buttonStart = new CButton(Point(411, 535), "SCNRSAV.DEF", CGI->generaltexth->zelp[103], std::bind(&CSavingScreen::saveGame, this), SDLK_s);
+	buttonStart->assignedKeys.insert(SDLK_RETURN);
+}
+
+CSavingScreen::~CSavingScreen()
+{
+	vstd::clear_pointer(localSi);
```
alexvins: does CSavingScreen actually owns localSi?

lib/NetPacksBase.h
```diff
@@ -17,13 +17,18 @@ class CArmedInstance;
...
+class CConnection;
 
 #include "ConstTransitivePtr.h"
 #include "GameConstants.h"
 
 struct DLL_LINKAGE CPack
 {
-	CPack() {};
+	// Not serialized. Set in CConnection::retreivePack
+	ui16 type;
```
alexvins: remove this

lib/mapping/CMapInfo.cpp
```diff
@@ -45,7 +45,7 @@ CMapInfo::CMapInfo() : scenarioOpts(nullptr), playerAmnt(0), humanPlayers(0),
 
 #define STEAL(x) x = std::move(tmp.x)
 
-CMapInfo::CMapInfo(CMapInfo && tmp):
+CMapInfo::CMapInfo(CMapInfo & tmp):
 	scenarioOpts(nullptr), playerAmnt(0), humanPlayers(0),
```
alexvins: Why you broken move constructor?

lib/serializer/Connection.h
```diff
@@ -58,27 +58,28 @@ class DLL_LINKAGE CConnection
...
 	bool isOpen() const;
-	bool isHost() const;
 	template<class T>
 	CConnection &operator&(const T&);
 	virtual ~CConnection();
 
-	CPack *retreivePack(); //gets from server next pack (allocates it with new)
-	void sendPackToServer(const CPack &pack, PlayerColor player, ui32 requestID);
+	// MPTODO: we need to assign shared_ptr to CConnection on CPack, but how to do this properly?
+	// Using weak_ptr inside CConnetion itself seems like even worse idea
```
alexvins: remove this comment

server/CGameHandler.cpp
```diff
@@ -93,29 +92,36 @@ class CBaseForGHApply
...
+		{
+			return ptr->applyGh(gh);
+		}
+		catch(notAllowedActionException & e)
+		{
+			return false;
+		}
+		catch(...)
+		{
+			throw;
```
alexvins: You should catch everything here.

server/CGameHandler.cpp
```diff
@@ -93,29 +92,36 @@ class CBaseForGHApply
...
+		}
 	}
 };
 
 template <>
 class CApplyOnGH<CPack> : public CBaseForGHApply
 {
 public:
-	bool applyOnGH(CGameHandler *gh, CConnection *c, void *pack, PlayerColor player) const override
+	bool applyOnGH(CGameHandler * gh, void * pack) const override
```
alexvins: Use `static_assert` instead of `assert(0);`

server/CGameHandler.cpp
```diff
@@ -2438,12 +2391,12 @@ void CGameHandler::changeSpells(const CGHeroInstance * hero, bool give, const st
...
 
-void CGameHandler::sendMessageTo(CConnection &c, const std::string &message)
+void CGameHandler::sendMessageTo(std::shared_ptr<CConnection> c, const std::string &message)
 {
 	SystemMessage sm;
 	sm.text = message;
-	boost::unique_lock<boost::mutex> lock(*c.wmx);
-	c << &sm;
+	boost::unique_lock<boost::mutex> lock(*c->wmx);
+	*(c.get()) << &sm;
```
alexvins: This line is unreadable.

client/pregame/CSavingScreen.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "../CGameInfo.h"
+#include "../../lib/CGeneralTextHandler.h"
+#include "../../lib/filesystem/Filesystem.h"
+#include "../../lib/CConfigHandler.h"
+
+CSavingScreen::CSavingScreen()
+	: CSelectionBase(CMenuScreen::saveGame)
+{
+	OBJ_CONSTRUCTION_CAPTURING_ALL;
+	bordered = false;
```
FeniksFire: move to the list initialization

client/pregame/CSavingScreen.cpp
```diff
@@ -0,0 +1,107 @@
...
+	center(pos);
+	// MPTODO: I doubt saving screen need any of it
+	localSi = LOCPLINT->cb->getStartInfo();
+	localMi = new CMapInfo();
+	localMi->mapHeader = std::unique_ptr<CMapHeader>(new CMapHeader(*LOCPLINT->cb->getMapHeader()));
+
+	tabSel = new SelectionTab(screenType); //scenario selection tab
+	tabSel->recActions = DISPOSE; // MPTODO
+	tabSel->recActions = 255;
```
FeniksFire: This looks strange, first you equal recActions to DISPOSE and then you make it = 255. Is it intentional?

client/pregame/CPreGame.h
```diff
@@ -0,0 +1,206 @@
...
+
+	CPicture * background;
+	std::vector<CPicture *> images;
+
+	CIntObject * createTab(size_t index);
+
+public:
+	std::vector<std::string> menuNameToEntry;
+
+	enum EState { //where are we?
```
FeniksFire: useless comment

client/CMT.cpp
```diff
@@ -453,15 +458,17 @@ int main(int argc, char * argv[])
...
-	CCS->soundh->init();
-	CCS->soundh->setVolume(settings["general"]["sound"].Float());
-	CCS->musich = new CMusicHandler();
-	CCS->musich->init();
-	CCS->musich->setVolume(settings["general"]["music"].Float());
-	logGlobal->info("Initializing screen and sound handling: %d ms", pomtime.getDiff());
-
+	if(!settings["session"]["headless"].Bool())
+	{
+		//initializing audio
```
FeniksFire: useless comment

client/pregame/SelectionTab.h
```diff
@@ -0,0 +1,74 @@
...
+	bool operator()(const std::shared_ptr<CMapInfo> aaa, const std::shared_ptr<CMapInfo> bbb);
+	mapSorter(ESortBy es) : sortBy(es){};
+};
+
+/// The selection tab which is shown at the map selection screen
+class SelectionTab : public CIntObject
+{
+public:
+	int positions; //how many entries (games/maps) can be shown
+	CPicture * bg; //general bg image
```
FeniksFire: useless comment and variable name, change it to background

lib/StartInfo.h
```diff
@@ -127,3 +152,98 @@ struct StartInfo
...
+			}
+		}
+		if(isClientHost(clientId))
+			players.insert(PlayerColor::NEUTRAL);
+
+		return players;
+	}
+
+	std::vector<ui8> getConnectedPlayerIdsForClient(int clientId) const
+	{
```
FeniksFire: create cpp and move these body function there

lib/serializer/Connection.cpp
```diff
@@ -216,19 +209,26 @@ void CConnection::reportState(vstd::CLoggerBase * out)
...
 	boost::unique_lock<boost::mutex> lock(*rmx);
-	logNetwork->trace("Listening... ");
-	iser & ret;
-	logNetwork->trace("\treceived server message of type %s", (ret? typeid(*ret).name() : "nullptr"));
-	return ret;
+	iser & pack;
+	logNetwork->trace("\treceived CPack of type %s", (pack ? typeid(*pack).name() : "nullptr"));
+	if(pack == nullptr)
+	{
+		logNetwork->error("Received a nullptr CPack! You should check whether client and server ABI matches.");
```
FeniksFire: you can delete that brackets that "if" and "else" have

server/CGameHandler.cpp
```diff
@@ -1835,19 +1788,19 @@ void CGameHandler::run(bool resume)
...
+	for (auto elem : conns)
 	{
 		std::set<PlayerColor> pom;
 		for (auto j = connections.cbegin(); j!=connections.cend();j++)
-			if (j->second == elem)
+			if(vstd::contains(j->second, elem))
 				pom.insert(j->first);
 
-		boost::thread(std::bind(&CGameHandler::handleConnection,this,pom,std::ref(*elem)));
+// 		boost::thread(std::bind(&CGameHandler::handleConnection,this,pom,elem));
```
FeniksFire: useless comment

client/CServerHandler.h
```diff
@@ -0,0 +1,112 @@
...
+class CServerHandler : public IServerAPI, public LobbyInfo
+{
+	SharedMemory * shm;
+
+	CApplier<CBaseForLobbyApply> * applier;
+
+	boost::recursive_mutex * mx;
+	std::list<CPackForLobby *> incomingPacks; //protected by mx
+
+	CStopWatch th;
```
FeniksFire: naming by shortcuts is not the best idea, it's good but for small ranges, I would change these names

ArseniyShestakov:
@FeniksFire I'm was hoping for more feedback on code structure itself rather than style / formatting / comments. E.g I unsure how I can further split ```CServerHandler``` into smaller components, etc.

But thanks anyway. I'll clean all of thing mentioned above along the way as I work on campaigns.

client/CMT.cpp
ArseniyShestakov: This one will still require some work since there is CMT with debug console and I don't know yet how to improve it.

lib/mapping/CMapInfo.cpp
ArseniyShestakov: It's was conflicting with some of my shared_ptr code during refactoring and the only thing commit description on it mentioned is that it was implemented to workaround some VS2013 bug.

Is it really needed?

lib/NetPacksBase.h
ArseniyShestakov: Removed code using it, but totally forgot to remove it from CPack. Thanks.

client/CServerHandler.h
ArseniyShestakov: I generally agree with this, but since code is still quite a mess and there no proper getters I feel like things such as ```si```, ```mi``` that used very often is reasonable trade off until another refactoring.

client/pregame/CSavingScreen.cpp
ArseniyShestakov: Nope, just some remaining from refactoring.

client/pregame/CSavingScreen.cpp
ArseniyShestakov: This one not owned by CSavingScreen.

server/CGameHandler.cpp
ArseniyShestakov: This one wouldn't work.

lib/mapping/CMapInfo.cpp
alexvins: `+CMapInfo::CMapInfo(CMapInfo & tmp)` is a copy constructor now, but it still works like move constructor.

client/pregame/CSavingScreen.cpp
ArseniyShestakov: Just checked - yes it is. At least for now it's not unified.

dydzio0614:
Looks like there is missing #include "StdInc.h" in CServerHandler.cpp

ArseniyShestakov:
Ok I'll postpone re-enabling of ```--headless``` because I'll need another refactoring pass to make lobby code completely independent from UI. And this option not critical enough to be a problem.

ArseniyShestakov:
So I got all basic stuff in. Half of it probably don't work properly and code is horrible, but in case I wouldn't be available next month it's could be merged and hotfixed relatively easy.

So here is list of things still need to be fixed:

- [ ] Campaigns: not all crossover options work properly.
- [ ] Campaigns: it's possible to get duplicate hero in ```Armageddon's Blade -> Playing With``` fire.
- [ ] Campaigns: if you win scenario / all campaign and then go to main menu and try to start another one client might crash.
- [ ] Campaigns: restart button from adventure map options crash. One on "scenario information" screen does work, but only implemented for single player.
- [ ] Lobby: when starting game with random map loading screen will only appear after server generated it. Start sequence needs refactoring.
- [ ] Lobby: when loading saves player isn't placed to his old color, but to first human-playable color like if it's was new game.
- [ ] Lobby: when loading it's would be nice to show randomized information in OptionsTab like on ```CScenarioInfoScreen```.
- [ ] In-game: restart button doesn't work properly if it's random map.
- [ ] In-game: when loading saves changes are not applied to CGameState. So ```PlayerState``` and ```StartInfo``` contain old player names. This is important for CGEvent since it's need to know if player human or AI.
- [ ] In-game: for loaded campaign view of "Scenario Information" on bonus selection screen will cause crash since CServerHandler doesn't contain enough of state.

There also two related bugs that likely are in develop:
- [ ] After loading some saves ```PlayerState::towns``` is empty. Town is owned by player, but gamestate don't know about that.
- [ ] Some normal and custom campaigns crash on attempt to playback prologue voice file at least on Linux.

client/CServerHandler.cpp
```diff
@@ -0,0 +1,690 @@
...
+		catch(...)
+		{
+			logNetwork->error("\nCannot establish connection! Retrying within 1 second");
+			boost::this_thread::sleep(boost::posix_time::seconds(1));
+		}
+	}
+	if(state == EClientState::CONNECTION_CANCELLED)
+	{
+		logNetwork->info("Connection aborted by player!");
+		return;
```
alexvins: remove this `return` and add `else` clause

client/CServerHandler.cpp
```diff
@@ -0,0 +1,690 @@
...
+	{
+	case StartInfo::NEW_GAME:
+		client->newGame();
+		break;
+	case StartInfo::CAMPAIGN:
+		client->newGame();
+		break;
+	case StartInfo::LOAD_GAME:
+		client->loadGame();
+		break;
```
alexvins: should be default clause here, just in case

lib/CGameState.cpp
```diff
@@ -840,16 +847,7 @@ void CGameState::initNewGame(const IMapService * mapService, bool allowSavingRan
 void CGameState::initCampaign(const IMapService * mapService)
 {
```
alexvins: remove parameter as you not use it anymore

lib/NetPacksLobby.h
```diff
@@ -0,0 +1,311 @@
...
+		h & bonusId;
+	}
+};
+
+struct LobbyChangePlayerOption : public CLobbyPackToServer
+{
+	enum EWhat : ui8 {UNKNOWN, TOWN, HERO, BONUS};
+	ui8 what;
+	si8 direction; //-1 or +1
+	PlayerColor color;
```
alexvins: Passing "each click" to server is not the best idea, pass selected value (with possible delay so if client change value fast only last is propogated).

lib/NetPacksLobby.h
```diff
@@ -0,0 +1,311 @@
...
+	{
+		h & what;
+		h & direction;
+		h & color;
+	}
+};
+
+struct LobbySetPlayer : public CLobbyPackToServer
+{
+	PlayerColor clickedColor;
```
alexvins: rename to smth like `selectedColor`

server/CVCMIServer.cpp
```diff
@@ -1,679 +1,962 @@
-/*
```
alexvins: Whole file changed. Is something wrong with line endings?

server/CGameHandler.cpp
```diff
@@ -1190,114 +1185,64 @@ void CGameHandler::applyBattleEffects(BattleAttack & bat, std::shared_ptr<battle
...
+		sendPackageResponse(true);
 	}
-	catch(...)
+	else
 	{
-		serverShuttingDown = true;
-		handleException();
-		throw;
+		logGlobal->error("Message cannot be applied, cannot find applier (unregistered type)!");
+		sendPackageResponse(false);
```
alexvins: I suspect this will never be called, Pack with unregistered type can not be received and should result in connection shutdown.

server/CVCMIServer.cpp
ArseniyShestakov: I actually changed 95% of code in it so I seen to reason to preserve any history. Usually I only make sure line endings are fine in new files, but I think it's make sense to do this and formatting when I actually checked all that code myself.

This also apply to CClient.

lib/NetPacksLobby.h
ArseniyShestakov: I have some ideas how to improve this by verifying state on "host" player so changes for that player UI can be immediately applied for best responsiveness while server would still actually control everything.

But it's minor thing that I probably won't work on yet. After all with local server it's applied almost immediately and for multiplayer we better to get some actually working lobby rather than delay landing this code.

lib/NetPacksLobby.h
ArseniyShestakov: I will intentionally keep this one like that until I actually fix all mess of how players positions actually managed by lobby backend code. Since I just moved code that was applied on client to the server it's messy, but since it's working I want to work on it after everything else.

client/CServerHandler.h
```diff
@@ -0,0 +1,149 @@
...
+	virtual void setPlayerOption(ui8 what, ui8 dir, PlayerColor player) const = 0;
+	virtual void setDifficulty(int to) const = 0;
+	virtual void setTurnLength(int npos) const = 0;
+	virtual void sendMessage(const std::string & txt) const = 0;
+	virtual void sendGuiAction(ui8 action) const = 0; // TODO: possibly get rid of it?
+	virtual void sendStartGame(bool allowOnlyAI = false) const = 0;
+};
+
+/// structure to handle running server and connecting to it
+class CServerHandler : public IServerAPI, public LobbyInfo
```
FeniksFire: One thing that made me wonder, why CServerHandler inherits from LobbyInfo.

server/CVCMIServer.h
```diff
@@ -1,112 +1,105 @@
...
+class CBaseForGHApply;
+
+enum class EServerState : ui8
+{
+	LOBBY,
+	GAMEPLAY,
+	SHUTDOWN
+};
+
+class CVCMIServer : public LobbyInfo
```
FeniksFire: same thing here, is this inheritance is necessary?

client/lobby/RandomMapTab.cpp
```diff
@@ -145,39 +137,29 @@ RandomMapTab::RandomMapTab()
...
-	// Initialize map info object
 	updateMapInfoByHost();
 }
 
 void RandomMapTab::showAll(SDL_Surface * to)
 {
 	CIntObject::showAll(to);
-
 	// Headline
 	printAtMiddleLoc(CGI->generaltexth->allTexts[738], 222, 36, FONT_BIG, Colors::YELLOW, to);
```
alexvins: All this should be replaced with CLabel

client/CServerHandler.h
ArseniyShestakov: This is done for convenience. Efficiently I might only inherit ```LobbyState``` and put everything here, but right I can update it this way on client:
```
bool LobbyUpdateState::applyOnLobbyHandler(CServerHandler * handler)
{
	hostChanged = state.hostClientId != handler->hostClientId;
	static_cast<LobbyState &>(*handler) = state;
	return true;
}
```
It's just saved me time on editing that code compared to if I just applied each field separately. So I had UUID out since to ```LobbyInfo``` so let's say UUID not changes when new state applied on client.

Also you might argue I should just put whole ```LobbyState``` into separate variable on both server and client, but right now it's will only make half of lines of code longer without obvious benefit. Since I choose to iterate over existing code rather than try to rewrite everything from scratch there still plenty of other more important improvements needed than deciding on nice looking API for shared lobby state.

client/CServerHandler.cpp
```diff
@@ -0,0 +1,676 @@
...
+	logNetwork->trace("Setting up thread calling server: %d ms", th->getDiff());
+
+	th->update();
+
+#ifndef VCMI_ANDROID
+	if(shm)
+		shm->sr->waitTillReady();
+#else
+	logNetwork->info("waiting for server");
+	while (!androidTestServerReadyFlag.load())
```
alexvins: Remove space after `while`

client/lobby/RandomMapTab.cpp
ArseniyShestakov: I wonder if we can do something about ```IFont::renderTextLeft``` usage too, though it used to render on top of ```SDL_Surface```.

client/CServerHandler.h
ArseniyShestakov: Yeah it's would be the best way, but before that code quality need to be on the point where it's going to have somewhat clear API...

client/lobby/CBonusSelection.cpp
```diff
@@ -553,41 +553,36 @@ CBonusSelection::CRegion::CRegion(CBonusSelection * _owner, bool _accessible, bo
...
 void CBonusSelection::CRegion::show(SDL_Surface * to)
 {
-	//const SCampPositions::SRegionDesc & desc = owner->campDescriptions[owner->ourCampaign->camp->header.mapVersion].regions[myNumber];
 	if(!accessible)
 	{
-		// Show as striped
-		blitAtLoc(graphics[2], 0, 0, to);
+		graphicsNotSelected->disable();
+		graphicsSelected->disable();
+		graphicsStriped->enable();
```
alexvins: You should not call enable/disable during rendering. It should be done on condition change. It that is too difficult, do not capture SHOWALL and call `child->show()` here

ArseniyShestakov: I'm already in process of cleanup of ```show``` and ```showAll``` .
Just not quite finished it yet.

ArseniyShestakov:
So I almost get rid of explicit blitting except in ```SelectionTab::printMaps``` that might need some special care and few other places. So no more tons of code in ```::showAll```. YAY!

lib/CGameState.cpp
```diff
@@ -673,9 +671,9 @@ int CGameState::getDate(Date::EDateType mode) const
 CGameState::CGameState()
 {
 	gs = this;
-	applierGs = new CApplier<CBaseForGSApply>();
-	registerTypesClientPacks1(*applierGs);
-	registerTypesClientPacks2(*applierGs);
+	applier = new CApplier<CBaseForGSApply>();
```
FeniksFire: any way for smart pointer here?

lib/NetPacksLobby.h
```diff
@@ -0,0 +1,311 @@
...
+class CCampaignState;
+class CLobbyScreen;
+class CServerHandler;
+class CMapInfo;
+struct StartInfo;
+class CMapGenOptions;
+struct ClientPlayer;
+class CVCMIServer;
+
+struct CPackForLobby : public CPack
```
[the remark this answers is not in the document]
ArseniyShestakov: @FeniksFire I think @alexvins can explain in details, but there a reasons why there can't be pure virtual stuff in netpacks.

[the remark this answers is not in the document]
alexvins: It may have (pure) virtual methods but they must be used only in LIB.

ArseniyShestakov:
So I end up finding that campaigns doesn't work very well in ```develop``` either so they'll need fixing anyway. While I'll probably land fix or two in this branch it's likely complete.

I'll continue improvements, but we really must merge this so work on UI improvements can continue.

PR #408, opened 2018-01-22
add vcmibuilder version check for innoextract

paulsetchko:
add vcmibuilder check for the old innoextract is unable to unpack the latest GOG package

### DIFF
--- a/vcmibuilder
+++ b/vcmibuilder
@@ -62,7 +62,7 @@ then
 	exit 0
 fi
 
-# test if a program version ($1) is lesser than the required one ($2)
+# test if a program version ($1) is less than the required one ($2)
 verlt() {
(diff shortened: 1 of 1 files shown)


ArseniyShestakov:
I don't get why this ```printf``` needed, but really you should at least check if version is greater.

paulsetchko:
fixed the commented issue; do you know why is that travis takes so long to start the mac build?

ArseniyShestakov:
Travis usually takes long on Mac builds mainly because we commit a lot and they have limited Mac instances pool. Regardless it's not required for this to be merged since we don't really expose it on macOS.

ArseniyShestakov:
Sorry for taking long to merge this, had to check that innoextract on 14.04(1.4) and 16.04 (1.5) in fact use same version output. Thanks for contributing!

PR #409, opened 2018-01-25
Refactoring: get rid of macro in server-side request validation code

ArseniyShestakov:
This is very minimal refactoring of ```NetPacksServer.cpp``` that I did simply because it's annoy me enough today during bigger netcode rework. It's unlikely that I'll be able to split my other branch in meaningful commits, but this change can for sure go standalone.

I'll prefer to merge this ASAP unless it's will cause any conflicts with other branches.

### DIFF
--- a/server/CGameHandler.h
+++ b/server/CGameHandler.h
@@ -310,4 +310,9 @@ class clientDisconnectedException : public std::exception
 
 };
 
+class ExceptionNotAllowedAction : public std::exception
+{
+
+};
+
 void makeStackDoNothing();
--- a/lib/NetPacks.h
+++ b/lib/NetPacks.h
@@ -65,6 +65,16 @@ struct CPackForServer : public CPack
 		logGlobal->error("Should not happen... applying plain CPackForServer");
 		return false;
 	}
+
+protected:
+	void throwNotAllowedAction();
+	void throwOnWrongOwner(CGameHandler * gh, ObjectInstanceID id);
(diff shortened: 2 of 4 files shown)


server/CGameHandler.h
```diff
@@ -310,4 +310,9 @@ class clientDisconnectedException : public std::exception
 
 };
 
+class notAllowedActionException : public std::exception
```
FeniksFire: Someone also coded such thing a bit higher:
```
class clientDisconnectedException : public std::exception
{

};
```
Anyway I think that it's don't have too much sense create such future-oriented things if they are empty.

alexvins: Class name should start with capital letter.

ArseniyShestakov: It's my code as well. I choose to throw another object just for the sake of readability and might be to put logging in there as well instead of spreading it around the code.

@FeniksFire any suggestion on whatever I should throw for just temporary things?

alexvins: @FeniksFire what is in general wrong with declaring new exception class just to filter it on catch?

FeniksFire: @ArseniyShestakov think that it's mostly depends on situation, so there is no one rule.
@alexvins ehh, such empty things for me makes code harder to read, and maintain the proper quality of the code. Of course there is nothing wrong with that code, just there is no need to code in this way and want point that such things can make trash in future.

PR #410, opened 2018-02-03
Refactoring: replace macro with variadic templates on client-side

ArseniyShestakov:
There still few netpacks that don't work with copying so I left some macro.
Also I'll need to work on naming it all properly.

Though it's works and I need to find out if it's going to compile with anything except my latest GCC.

PS: Big thanks to @mwu-tow for helping me with this.

### DIFF
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -485,8 +485,8 @@ void CClient::newGame( CConnection *con, StartInfo *si )
 // 	for(FileInfo &m : scriptModules)
 // 	{
 // 		CScriptingModule * nm = CDynLibHandler::getNewScriptingModule(m.name);
-// 		privilagedGameEventReceivers.push_back(nm);
-// 		privilagedBattleEventReceivers.push_back(nm);
+// 		privilegedGameEventReceivers.push_back(nm);
+// 		privilegedBattleEventReceivers.push_back(nm);
 // 		nm->giveActionCB(this);
 // 		nm->giveInfoCB(this);
 // 		nm->init();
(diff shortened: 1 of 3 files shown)


ArseniyShestakov:
Pushed version without copying and without any macros since there no issues with copying.

Though we still have to get rid of mess in ```SetStackType``` / ```InsertNewStack``` since it's pass ```CCreature``` pointer over network.

DjWarmonger:
Does forwarding arguments make the code possibly more efficient?

For sure it look better now.

ArseniyShestakov:
@DjWarmonger of course, but netpacks usually contain little ehough data inside so difference very negligible. Reason for first commit being without forwarding was that I forgot about differences ```std::shared_ptr``` introduce and had fun time trying to make it work. So I tried several iterations until I find out what is actually wrong.

mwu-tow:
@DjWarmonger Forwarding can be faster — as it can mean either passing by reference or to passing by move. Some types (like string) are relatively slow to copy, though in our case the difference will be negligible. 
I'd say it is not so much about performance as for correct semantics. The purpose of the `INTERFACE_CALL_IF_PRESENT` is to forward its arguments from a call site to the actual function — not to force them into being passed/received in any particular way. Whether they are passed by value or reference, copied or moved — it will be decided by the called method signature and what is given at the call site. In that matter taking arguments by T&& and std::forward-ing should give the same behaviour as we had with macros before.

After all, it wouldn't make sense to have method in interface taking const reference and call that method with a helper that forces copying, effectively undoing the "optimization".

@ArseniyShestakov Seeing all that `cl` passing makes me wonder if `INTERFACE_CALL_IF_PRESENT` (&friends) shouldn't be a member of `CClient`?

ArseniyShestakov:
@mwu-tow Good idea, but before that I'll have to make proper spectator implementation that actually use privileged callback and own UI class. I don't want to put in CClient header what we have now.

ArseniyShestakov:
Ok I'll merge this and then rebase Alex's branch myself.

PR #411, opened 2018-02-09
Hide "this statement may fall through" warnings

drizt:
Hide false-positive warnings like

```
/home/taurus/develop/vcmi/lib/battle/BattleInfo.cpp: In static member function ‘static BattleInfo* BattleInfo::setupBattle(int3, ETerrainType, BFieldType, const CArmedInstance**, const CGHeroInstance**, bool, const CGTownInstance*)’:
/home/taurus/develop/vcmi/lib/battle/BattleInfo.cpp:463:17: warning: this statement may fall through [-Wimplicit-fallthrough=]
    bonusSubtype = 0;
    ~~~~~~~~~~~~~^~~
/home/taurus/develop/vcmi/lib/battle/BattleInfo.cpp:465:2: note: here
  case BFieldType::FIERY_FIELDS:
  ^~~~

```

### DIFF
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -2090,6 +2090,7 @@ void CGameState::updateRumor()
 			}
 			else
 				rumor.type = RumorState::TYPE_RAND;
+			FALLTHROUGH
 
 		case RumorState::TYPE_RAND:
 			do
--- a/Global.h
+++ b/Global.h
@@ -86,6 +86,21 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #  error "iOS system isn't yet supported."
 #endif
 
+// Each compiler uses own way to supress fall through warning. Try to find it.
+#ifdef __has_cpp_attribute
+#  if __has_cpp_attribute(fallthrough)
+#    define FALLTHROUGH [[fallthrough]];
+#  elif __has_cpp_attribute(gnu::fallthrough)
+#    define FALLTHROUGH [[gnu::fallthrough]];
+#  elif __has_cpp_attribute(clang::fallthrough)
+#    define FALLTHROUGH [[clang::fallthrough]];
+#  else
+#    define FALLTHROUGH
+#  endif
+#else
+#  define FALLTHROUGH
+#endif
+
 /* ---------------------------------------------------------------------------- */
 /* Commonly used C++, Boost headers */
 /* ---------------------------------------------------------------------------- */
--- a/lib/CPathfinder.cpp
+++ b/lib/CPathfinder.cpp
@@ -527,6 +527,7 @@ CGPathNode::ENodeAction CPathfinder::getDestAction() const
 		}
 
 		/// don't break - next case shared for both land and sail layers
+		FALLTHROUGH
 
 	case ELayer::SAIL:
 		if(isDestVisitableObj())
--- a/lib/battle/BattleInfo.cpp
+++ b/lib/battle/BattleInfo.cpp
@@ -462,18 +462,22 @@ BattleInfo * BattleInfo::setupBattle(int3 tile, ETerrainType terrain, BFieldType
 		{
 			bonusSubtype = 0;
 		}
+		FALLTHROUGH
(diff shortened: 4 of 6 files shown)


lib/CGameState.cpp
```diff
@@ -2090,6 +2090,7 @@ void CGameState::updateRumor()
 			}
 			else
 				rumor.type = RumorState::TYPE_RAND;
+			//-fallthrough
```
alexvins: Is there any documentation for such syntax?

drizt: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
-Wimplicit-fallthrough=3 case sensitively matches one of the following regular expressions:
    -fallthrough

ArseniyShestakov:
Is Clang also following same syntax? Should we add something to our CMake configuration?

alexvins:
>C++17 provides a standard way to suppress the -Wimplicit-fallthrough warning using [[fallthrough]]; instead of the GNU attribute. In C++11 or C++14 users can use [[gnu::fallthrough]];, which is a GNU extension.

alexvins:
If we find an answer for clang I`d suggest using attribute instead of comment

alexvins:
https://stackoverflow.com/questions/27965722/c-force-compile-time-error-warning-on-implicit-fall-through-in-switch

alexvins:
@ArseniyShestakov clang has attribute, but usage breaks compilation with MSVS. Comment seems to be gcc-only.

ArseniyShestakov:
I guess we might for now go with some macro, right?

dydzio0614:
If we drop VS 2015 support then it would work

alexvins:
Yes we could use a macro with  `[[gnu::fallthrough]];` for GCC  '[[clang::fallthrough]];' for clang and empty for others

ArseniyShestakov:
We'll need to know behavior of much older Clang / GCC versions so we can only enable it for supported compilers.

@dydzio0614 you help by suggest what to use for MSVC and what version support it.

dydzio0614:
Visual Studio supports [[fallthrough]] since first version of VS 2017 - macro check: if defined(_MSC_VER) /*MSVC compiler detected*/ && (_MSC_VER >= 1910) /*version of that compiler is at least 19.1 (first iteration of VS 2017)*/

drizt:
[[fallthrough]] and [[clang::fallthrough]] works with my g++ 
```

$ g++ --version
g++ (GCC) 7.3.1 20180130 (Red Hat 7.3.1-2)
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

```

alexvins:
We support gcc from at least 4.8 (4.7.2+ strict condition)

drizt:
Hm project use C++11. But I don't see where it explicity definded in the code.

alexvins:
See `Global.h`

alexvins:
standard version -std=c++0x set in root cmakelists.txt

drizt:
CMake now allows to explicity set C++ standard.

```
# Enable C++14
set(CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_STANDARD 14)

```

drizt:
Is there any reasons to use `-std=c++0x` instead of `-std=c++11`? https://stackoverflow.com/a/24589681/2808783 . I grep sources with `c++11` and found nothing.

alexvins:
-std=c++0x was added quite long ago, when c++11 was not so widely supported. I see no reason to change it now. Eventually we`ll switch to newer standard and change it accordingly.

drizt:
Add `FALLTHROUGH` macro.

ArseniyShestakov:
Looks like it will work well enough. Merged!

jcfr:
As a best practice, I would suggest to use macros with a prefix specific to this project. Instead of `FALLTHROUGH`, you could use `VCMI_FALLTHROUGH`.

Generally speaking, prefixing macros avoid clashes with other dependencies.

PR #412, opened 2018-02-09
Update version on each compiling

drizt:


ArseniyShestakov:
Is the old version of code didn't worked for you?
I was pretty sure it's updated with every little commit just fine regardless of any other changes.

drizt:
It works for me. And yes current way is pretty good. But in TODO _So for now commit hash only updated when CMake reconfigured._ This patch allows to avoid extra CMake reconfiguration.

ArseniyShestakov:
In case this code should actually work for all generators you should remove that TODO item from main CMakeLists.txt.
Now I can recall it's was problem with Visual Studio where it's only re-generated Version.cpp on reconfiguration.

drizt:
@dydzio0614, can you test this with Visual Studio?

drizt:
@ArseniyShestakov, should I move **Version.cmake** to **cmake_modules** folder?

ArseniyShestakov:
Yeah it's good idea to place it into ```cmake_modules```.
I'll check it with MSVS.

drizt:
Also with this patch it is possible to add extra data to Version.cpp. Compilation date and time for example.

ArseniyShestakov:
This code doesn't seems to work when you building outside of source directory.

drizt:
Change `CMAKE_CURRENT_SOURCE_DIR` to `CMAKE_CURRENT_LIST_DIR` in **GetGitRevisionDescription.cmake**.

ArseniyShestakov:
Thanks for contribution! Tested with MSVS2015/2017 and Xcode 8.3 and everything seems to work well.

I cherry picked and edited it a bit here:
https://github.com/vcmi/vcmi/commit/3c1c72319b53b6d7c531a33dc5f5e500a9572e3d
Closing.

PR #414, opened 2018-02-09
Fixed lobby player assignment - wrong AI identify

dydzio0614:
Fixed undefined behavior - zero should be set instead of empty collection iterator content


ArseniyShestakov:
Thanks! Good catch!

PR #418, opened 2018-02-23
Fix mantis bugs: #2884. #2885, #2886

dydzio0614:
Fixed mummy, rust dragon, halflings creature abilities

### DIFF
--- a/config/creatures/neutral.json
+++ b/config/creatures/neutral.json
@@ -256,6 +256,10 @@
 				"val" : 25,
 				"addInfo" : 20
 			},
+			"fireBreath" :
+			{
+				"type" : "TWO_HEX_ATTACK_BREATH"
+			},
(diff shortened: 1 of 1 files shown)


ArseniyShestakov:
Might be worth to check some interesting details about Mummy ability:
http://heroes.thelazy.net/wiki/Mummy

No idea how correct are they since I can't find them in russian fizmig, but worth to check.

alexvins:
Fiery field aspect is part of another issue https://bugs.vcmi.eu/view.php?id=2888, and it _is_ described in fizmig. I do understand what does this  

> This is due to a fact that the ability is not affected by Fiery Fields terrain, and creatures with immunity to fire magic (e.g. Fire Elementals) are not affected by it.

say about fire immunity. Is ability not affected by immunity or immune creatures are not affected by spell?

PR #422, opened 2018-02-28
secondary skill bonuses get updated via replacement (issue 2796)

henningkoehlernz:
bugfix, see https://bugs.vcmi.eu/view.php?id=2796

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -850,6 +850,14 @@ void CBonusSystemNode::removeBonus(const std::shared_ptr<Bonus>& b)
 	CBonusSystemNode::treeHasChanged();
 }
 
+void CBonusSystemNode::removeBonuses(const CSelector &selector)
+{
+	BonusList toRemove;
+	exportedBonuses.getBonuses(toRemove, selector);
+	for(auto bonus : toRemove)
+		removeBonus(bonus);
+}
+
 bool CBonusSystemNode::actsAsBonusSourceOnly() const
 {
 	switch(nodeType)
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -791,7 +791,7 @@ void CBonusSystemNode::detachFrom(CBonusSystemNode *parent)
 	CBonusSystemNode::treeHasChanged();
 }
 
-void CBonusSystemNode::popBonuses(const CSelector &s)
+void CBonusSystemNode::removeBonusesRecursive(const CSelector &s)
 {
 	BonusList bl;
 	exportedBonuses.getBonuses(bl, s, Selector::all);
@@ -799,7 +799,7 @@ void CBonusSystemNode::popBonuses(const CSelector &s)
 		removeBonus(b);
 
 	for(CBonusSystemNode *child : children)
-		child->popBonuses(s);
+		child->removeBonusesRecursive(s);
 }
 
 void CBonusSystemNode::reduceBonusDurations(const CSelector &s)
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -791,14 +791,10 @@ void CBonusSystemNode::detachFrom(CBonusSystemNode *parent)
 	CBonusSystemNode::treeHasChanged();
 }
 
-void CBonusSystemNode::removeBonusesRecursive(const CSelector &s)
+void CBonusSystemNode::removeBonusesRecursive(const CSelector & s)
 {
-	BonusList bl;
-	exportedBonuses.getBonuses(bl, s, Selector::all);
-	for(auto b : bl)
-		removeBonus(b);
-
-	for(CBonusSystemNode *child : children)
+	removeBonuses(s);
+	for(CBonusSystemNode * child : children)
 		child->removeBonusesRecursive(s);
 }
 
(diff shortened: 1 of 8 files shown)


lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -559,21 +559,10 @@ void CGHeroInstance::recreateSpecialtyBonuses(std::vector<HeroSpecial *> & speci
 
 void CGHeroInstance::updateSkill(SecondarySkill which, int val)
 {
+	removeBonuses(Selector::source(Bonus::SECONDARY_SKILL, which));
 	auto skillBonus = (*VLC->skillh)[which]->getBonus(val);
 	for (auto b : skillBonus)
-	{
-		// bonuses provided by different levels of a secondary skill are aggregated via max (not + as usual)
-		// different secondary skills providing the same bonus (e.g. ballistics might improve archery as well) are kept separate
```
ArseniyShestakov: In past code added each level of secondary skills as separate bonus. So each level of skill might have completely different effects and all of them would be assigned to hero separately. Now on update you remove all assigned bonuses for the skill of hero, but only adding effects of single level.

So to make sure, there is no legacy data depended on old behavior? I suppose old saves loaded from legacy data are either converted by your compatibility code or simply didn't depend on it anyway?

lib/HeroBonus.cpp
```diff
@@ -850,6 +850,14 @@ void CBonusSystemNode::removeBonus(const std::shared_ptr<Bonus>& b)
 	CBonusSystemNode::treeHasChanged();
 }
 
+void CBonusSystemNode::removeBonuses(const CSelector &selector)
```
ArseniyShestakov: And one more request. We already had ```CBonusSystemNode::popBonuses``` that doing almost (?) the same thing except it's also pop bonuses from children. I suppose secondary skills not affect children nodes so it's should be okay.

Though I think we need to rename ```popBonuses``` now to make it easier to grasp where you should use it or ```removeBonuses```. At least it's seems like something important for me.

ArseniyShestakov:
@alexvins I would appreciate your opinion on methods naming here.

lib/mapObjects/CGHeroInstance.cpp
henningkoehlernz: All secondary skill definitions specified the full bonus for each level anyways. The only potential issue is with mods that would assign different effects for each level (and assume them to be commulative) - but given that secondary skill modding is still fairly new, I doubt there are any.

lib/HeroBonus.cpp
henningkoehlernz: Hadn't spotted that function. So I fully agree, popBonuses is not ideal, especially since we have a method removeBonus for single bonus. My suggestion would be `removeBonusesRecursive` or similar.

lib/mapObjects/CGHeroInstance.cpp
ArseniyShestakov: I guess fact that they're not cumulative give system even more flexibility so this is fine.

lib/HeroBonus.cpp
ArseniyShestakov: Makes sense to me. Go for ```removeBonusesRecursive``` then.

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -559,21 +559,10 @@ void CGHeroInstance::recreateSpecialtyBonuses(std::vector<HeroSpecial *> & speci
 
 void CGHeroInstance::updateSkill(SecondarySkill which, int val)
```
ArseniyShestakov: Would be also nice to rename ```updateSkill``` to ```updateSkillBonus``` or something.

lib/HeroBonus.cpp
```diff
@@ -791,15 +791,15 @@ void CBonusSystemNode::detachFrom(CBonusSystemNode *parent)
 	CBonusSystemNode::treeHasChanged();
 }
 
-void CBonusSystemNode::popBonuses(const CSelector &s)
+void CBonusSystemNode::removeBonusesRecursive(const CSelector &s)
```
alexvins: This duplicates `removeBonuses` it should call it

lib/HeroBonus.cpp
```diff
@@ -850,6 +850,14 @@ void CBonusSystemNode::removeBonus(const std::shared_ptr<Bonus>& b)
 	CBonusSystemNode::treeHasChanged();
 }
 
+void CBonusSystemNode::removeBonuses(const CSelector &selector)
+{
+	BonusList toRemove;
+	exportedBonuses.getBonuses(toRemove, selector);
```
alexvins: Do not forget about second selector, Set it to `Selector::all` if unsure. It you omit it, function will only return `b->effectRange == Bonus::NO_LIMIT`

lib/HeroBonus.h
```diff
@@ -709,13 +709,13 @@ class DLL_LINKAGE CBonusSystemNode : public virtual IBonusBearer, public boost::
 	void propagateBonus(std::shared_ptr<Bonus> b);
 	void unpropagateBonus(std::shared_ptr<Bonus> b);
 	void removeBonus(const std::shared_ptr<Bonus>& b);
+	void removeBonuses(const CSelector &selector);
+	void removeBonusesRecursive(const CSelector &s);
```
alexvins: put space around `&`. Whole file has bad formatting, but new additions should follow guidelines.

lib/HeroBonus.cpp
henningkoehlernz: Fixed. Though seems counter-intuitive that this is default behaviour.

alexvins: This is kind of legacy, good solution is to completely remove this function. (There is much to clean up in bonus API)

PR #423, opened 2018-03-01
Configure JSON merge behaviour via #override flag

henningkoehlernz:
By default, JSON structs from mods are always merged with an existing struct. This is troublesome when the new data is not intended to be an extension of the old data. E.g.

	"valeska":
	{
		...
		"specialty" : {
			"creature" : "archer"
		}
	},

patched with

	"core:valeska":
	{
		"specialty" : {
			"base" : { ... },
			"bonuses" : { ... }
		}
	},

will result in duplicate bonuses:

	"valeska":
	{
		...
		"specialty" : {
			"base" : { ... },
			"bonuses" : { ... },
			"creature" : "archer"
		}
	},

With this change it is now possible to specify the patch as

	"core:valeska":
	{
		"specialty#override" : {
			"base" : { ... },
			"bonuses" : { ... }
		}
	},

which completely replaces (overrides) the specialty field:

	"valeska":
	{
		...
		"specialty" : {
			"base" : { ... },
			"bonuses" : { ... }
		}
	},

A side-effect is that the # character can no longer be used in naming fields.

### DIFF
--- a/lib/JsonNode.cpp
+++ b/lib/JsonNode.cpp
@@ -70,7 +70,8 @@ JsonNode::JsonNode(ResourceID && fileURI, bool &isValidSyntax):
 
 JsonNode::JsonNode(const JsonNode &copy):
 	type(JsonType::DATA_NULL),
-	meta(copy.meta)
+	meta(copy.meta),
+	flags(copy.flags)
 {
 	setType(copy.getType());
 	switch(type)
@@ -96,6 +97,7 @@ void JsonNode::swap(JsonNode &b)
 	swap(meta, b.meta);
 	swap(data, b.data);
 	swap(type, b.type);
+	swap(flags, b.flags);
 }
 
 JsonNode & JsonNode::operator =(JsonNode node)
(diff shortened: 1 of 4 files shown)


lib/JsonNode.h
```diff
@@ -45,8 +45,10 @@ class DLL_LINKAGE JsonNode
 	JsonData data;
 
 public:
-	/// free to use metadata field
+	/// free to use metadata fields
 	std::string meta;
+	// meta-flags like override
+	std::vector<std::string> flags;
```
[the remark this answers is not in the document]
henningkoehlernz: Thanks, fixed.

lib/JsonNode.cpp
```diff
@@ -872,23 +874,42 @@ void JsonUtils::merge(JsonNode & dest, JsonNode & source)
...
+				for(std::string flag : source.flags)
+				{
+					if(flag.compare("override") == 0)
+					{
+						override = true;
+						break;
+					}
+				}
+			}
+			if(override)
```
ArseniyShestakov: Construction above seems like overkill. This should work:
```
if(!noOverride && vstd::contains(source.flags, "override"))
```
I understand that you want to make code more explicit, but I really doubt it's required there.

ArseniyShestakov:
BTW in future please add your branches to main repository.
Then I'll be able to immediately fix, squash it and merge myself. Thanks.

lib/JsonNode.cpp
henningkoehlernz: Just been too long that I coded much for VCMI, forgot we had that handy function. Fixed.

PR #425, opened 2018-03-02
Catapult rework

dydzio0614:
Fix for catapult/walls mechanics to match H3.

Done:
- Fixed cyclops wall destruction to match H3 completely
- Repaired arrow towers buggy state while half-destroyed
- Visual fix for arrow tower destruction - before changes it was very ugly from player's point of view (shooters disappearing before stone hit etc.)
- Check hitchance/target randomization
- Fix Catapult spell effect

### DIFF
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -2905,9 +2905,13 @@ std::string CBattleInterface::SiegeHelper::getSiegeName(ui16 what, int state) co
 		switch (state)
 		{
 		case EWallState::INTACT : return 1;
-		case EWallState::DAMAGED : return 2;
+		case EWallState::DAMAGED :
+			if(what == 2 || what == 3 || what == 8) // towers don't have separate image here - INTACT and DAMAGED is 1, DESTROYED is 2
+				return 1;
+			else
+				return 2;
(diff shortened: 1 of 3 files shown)


alexvins:
Are any fixes required in Catapult spell effect too? I suppose at least "Visual fix" is.

dydzio0614:
Yep, there are some changes required - send BattleUnitsChanged netpack after CatapultAttack netpack, remove shooters from towers only if tower is completely destroyed by attack, not just damaged.

alexvins:
Will you do it yourself, or I should push that changes here?

dydzio0614:
Okay, so this branch is pretty much done for now. Hit chance management is fair enough, though behavior of catapult without basic ballistics is noticeably different from H3. Too hard to say how it should work without knowing exact mechanics. It needs to wait its time, what is already done is good enough to be merged.

ArseniyShestakov:
Great visual improvement! Thanks!

There still one visual bug present at least:

* Something damage gate.
* Now Cyclops Kings attack bridge first and destroy gates.
* No changes are visible.
* Now Cyclops Kings making second hit.
* Now after second hit and end of turn destroyed gates appear.

I guess it's might be related to my own code for draw bridge and state probably only changed after turn is completed.

PR #426, opened 2018-03-03
redefined lctrl to lgui for macs

jerryskye:
The reason for my PR is Macs use ctrl + click shortcut for popping up a context menu. It can't be turned off.
I changed the code so that the cool shortcuts for splitting armies use the command key, rather than control.

Keep up the good work guys!
Regards


ArseniyShestakov:
Thanks for contribution!

PR #427, opened 2018-03-07
Extend Bonus.addInfo to integer vector

henningkoehlernz:
Some bonus types (SPELL_BEFORE_ATTACK, SPELL_AFTER_ATTACK) are already running out of fields to store data, new or extended bonus types (in particular IMPROVED_NECROMANCY) may need more storage as well.

To address this, the addInfo field (currently an integer) is extended to an array of integers. When specified as integer in JSON, this is converted into array of length one (for ease of use and downwards compatibility).

### DIFF
--- a/config/creatures/dungeon.json
+++ b/config/creatures/dungeon.json
@@ -183,7 +183,7 @@
 				"type" : "SPELL_AFTER_ATTACK",
 				"subtype" : "spell.stoneGaze",
 				"val" : 20,
-				"addInfo" : 2000 // FIXME: replace with range field?
+				"addInfo" : [0,2]
 			}
 		},
 		"upgrades": ["medusaQueen"],
@@ -217,7 +217,7 @@
 				"type" : "SPELL_AFTER_ATTACK",
 				"subtype" : "spell.stoneGaze",
 				"val" : 20,
-				"addInfo" : 2000 // FIXME: replace with range?
+				"addInfo" : [0,2]
 			}
 		},
 		"graphics" :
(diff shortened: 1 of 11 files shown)


DjWarmonger:
Hmhmm, this is getting complex. I propose to handle IMPROVED_NECROMANCY as separate bonuses (necromancy level -> creature ID). We can then store necromancy level as addInfo and creature id as a subtype, or the other way round.

henningkoehlernz:
This change doesn't touch IMPROVED_NECROMANCY yet (just prepares things for it). The issue for IMPROVED_NECROMANCY was that I wanted another field to make creature raised also depend on level of creature killed (something that's not part of H3), and val seemed like a poor choice.
Once we have the functionality of this pull request, I'm sure there'll be other uses for it.

alexvins:
I think we should do something with `addInfo` anyway. If extending it is too complex when we should remove it and always use separate bonus instead.

henningkoehlernz:
I don't think it's very complex - just requires a lot of small code changes to switch from int to array internally since it's used as int in a bunch of places.
Secondly, removing addInfo and somehow linking multiple bonuses together to store the required information for one bonus seems much more complex and cumbersome to me.

ArseniyShestakov:
I don't have strong opinion on how bonus system should work in this particular case, but code seems reasonable and I don't see any reason it's might break something. So if someone tested this manually feel free to merge it.

Otherwise I might merge it later myself, but only if I able to test it properly.

ArseniyShestakov:
@henningkoehlernz so if you have test mods already using this feel free to merge it yourself.

henningkoehlernz:
Tested with medusa (which makes use of the new format), but don't seem to have permission to merge.

ArseniyShestakov:
Oops, wasn't aware that collaborators permissions don't let you merge.
Will keep it that in mind.

PR #428, opened 2018-03-10
Reflect changes in boost::asio released in Boost 1.66.

Raphnalor:
The service template parameters, and the corresponding classes,
are disabled by default for now. For example,
basic_socket<Protocol, SocketService> => basic_socket<Protocol>

### DIFF
--- a/lib/serializer/Connection.cpp
+++ b/lib/serializer/Connection.cpp
@@ -14,6 +14,9 @@
 #include "../mapping/CMap.h"
 #include "../CGameState.h"
 
+#if BOOST_VERSION >= 106600
+#define BOOST_ASIO_ENABLE_OLD_SERVICES
+#endif
 #include <boost/asio.hpp>
 
 using namespace boost;
--- a/lib/serializer/Connection.h
+++ b/lib/serializer/Connection.h
@@ -22,7 +22,13 @@ namespace boost
 		{
 			class tcp;
 		}
+
+#if BOOST_VERSION >= 106600  // Boost version >= 1.66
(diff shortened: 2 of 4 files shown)


FeniksFire:
I think that few boost versions should be supported backwards, if there is not much difference. Changing `io_service` to `io_context` is unnecesary and bad for backward compatibility. Check that [typedef](http://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/io_service.html). So `io_service` can stay. Only thing that you need to change are forward declarations in `CVCMIServer.h` and `Connection.h`

ArseniyShestakov:
There is already code suggested for this in PR #285 with compatibility for older Boost.
I'm little bit short of time recently, but as soon as I can I'll merge one or another.

ArseniyShestakov:
@Raphnalor so if you can test it and put it in this PR I'll be happy to accept it.
Later I can improve the code a bit once I finalize #402.

Raphnalor:
I've fixed compatibility with older Boost versions.

MarcinWieczorek:
Does this fix *all* compilation issues with boost 1.66?

Raphnalor:
I was able to build VCMI with boost 1.66 on Windows without errors using this fix.

MarcinWieczorek:
I'm compiling it now.

MarcinWieczorek:
It works.

ArseniyShestakov:
@Raphnalor merge button is just for team members regardless of "approved" state.
Thanks for your contribution!

PR #431, opened 2018-03-16
Fixed Issue #2918

henningkoehlernz:
See https://bugs.vcmi.eu/view.php?id=2918


lib/HeroBonus.cpp
```diff
@@ -178,6 +178,13 @@ bool CAddInfo::operator!=(si32 value) const
 	return !operator==(value);
 }
 
+si32 & CAddInfo::operator[](size_type pos)
+{
+	if (pos <= size())
```
alexvins: no space after `if`, please.

PR #432, opened 2018-03-17
Fix highlighting of a selected bonus for campaign.

Raphnalor:
The problem is that a selected bonus isn't highlighted:
https://drive.google.com/file/d/16QcvXwAlxwpQI1nlsljNY1flylFf0QK1/view

My PR fixes it:
https://drive.google.com/file/d/1cKmpqYmMeLized9jogW418r-nHO6qr3_/view


ArseniyShestakov:
@Raphnalor thanks for the fix!

If you decide to fix anything else in pregame please check it's state in #402 since I changed almost all that code. This particular fix though isn't something I had time for yet.

PR #433, opened 2018-03-18
Make bonus stacking configurable + fix duplicate propagation/inheritance

henningkoehlernz:
Addresses several related problems:
- Propagation / unpropagation of duplicate bonuses is inconsistent, causing bugs
- Duplicate bonuses never stack, which is not always intended behaviour (e.g. multiple copies of resource generating artifacts)
- Different bonuses always stack, which is not always intended behaviour (e.g. Angel + Archangel morale bonuses)

This is addressed as follows:
- Duplicate bonuses are never eliminated during propagation/inheritance.
- Unpropagation eliminates only a single copy of duplicated bonus
- Bonus receives a new field `stacking` that determines stacking behaviour:
-- empty string = no stacking with duplicates (default)
-- `"ALWAYS"` = stacks with duplicates & everything else
-- some other value = no stacking with bonuses with same stacking value

Example usage:
```
"angel" : {
    ...
    "const_raises_morale" : {
        "propagator" : "HERO",
        "type" : "MORALE",
        "val" : 1,
        "stacking" : "angel"
    },
    ...
},
"archAngel" : {
    ...
    "const_raises_morale" : {
        "propagator" : "HERO",
        "type" : "MORALE",
        "val" : 1,
        "stacking" : "angel"
    },
    ...
}
```
```
"endlessSackOfGold" : {
    ...
    {
        "subtype" : "resource.gold",
        "type" : "GENERATE_RESOURCE",
        "val" : 1000,
        "valueType" : "BASE_NUMBER",
        "stacking" : "ALWAYS"
    }
    ...
}
```

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -245,39 +245,43 @@ void BonusList::changed()
 		CBonusSystemNode::treeHasChanged();
 }
 
-TBonusListPtr BonusList::stackingBonuses() const
+void BonusList::stackBonuses()
 {
-	TBonusListPtr result(new BonusList(*this));
-	boost::sort(result->bonuses, [](std::shared_ptr<Bonus> b1, std::shared_ptr<Bonus> b2) -> bool
+	boost::sort(bonuses, [](std::shared_ptr<Bonus> b1, std::shared_ptr<Bonus> b2) -> bool
 	{
 		if(b1 == b2)
 			return false;
-		if(b1->stacking != b2->stacking)
-			return b1->stacking < b2->stacking;
-		if(b1->valType != b2->valType)
-			return b1->valType < b2->valType;
+#define COMPARE_ATT(ATT) if(b1->ATT != b2->ATT) return b1->ATT < b2->ATT
+		COMPARE_ATT(stacking);
+		COMPARE_ATT(type);
+		COMPARE_ATT(subtype);
+		COMPARE_ATT(valType);
+#undef COMPARE_ATT
 		return b1->val > b2->val;
 	});
 	// remove non-stacking
 	int next = 1;
-	while(next < result->size())
+	while(next < bonuses.size())
 	{
 		bool remove;
-		const std::shared_ptr<Bonus> last = result->bonuses[next-1];
-		const std::shared_ptr<Bonus> current = result->bonuses[next];
+		const std::shared_ptr<Bonus> last = bonuses[next-1];
+		const std::shared_ptr<Bonus> current = bonuses[next];
+
 		if(current->stacking.empty())
 			remove = current == last;
 		else if(current->stacking == "ALWAYS")
 			remove = false;
 		else
 			remove = current->stacking == last->stacking
+				&& current->type == last->type
+				&& current->subtype == last->subtype
 				&& current->valType == last->valType;
+
 		if(remove)
-			result->bonuses.erase(result->bonuses.begin() + next);
+			bonuses.erase(bonuses.begin() + next);
 		else
 			next++;
 	}
-	return result;
 }
 
 int BonusList::totalValue() const
@@ -291,8 +295,7 @@ int BonusList::totalValue() const
 	int indepMin = 0;
 	bool hasIndepMin = false;
 
-	TBonusListPtr afterStacking = stackingBonuses();
-	for(std::shared_ptr<Bonus> & b : *afterStacking)
+	for(std::shared_ptr<Bonus> b : bonuses)
 	{
 		switch(b->valType)
 		{
@@ -717,6 +720,7 @@ const TBonusListPtr CBonusSystemNode::getAllBonuses(const CSelector &selector, c
 			BonusList allBonuses;
 			getAllBonusesRec(allBonuses);
 			limitBonuses(allBonuses, cachedBonuses);
+			cachedBonuses.stackBonuses();
 
 			cachedLast = treeChanged;
 		}
@@ -761,7 +765,6 @@ const TBonusListPtr CBonusSystemNode::getAllBonusesWithoutCaching(const CSelecto
 	if(!root || root == this)
 	{
 		limitBonuses(beforeLimiting, afterLimiting);
-		afterLimiting.getBonuses(*ret, selector, limit);
 	}
 	else if(root)
 	{
@@ -779,8 +782,9 @@ const TBonusListPtr CBonusSystemNode::getAllBonusesWithoutCaching(const CSelecto
 			if(vstd::contains(limitedRootBonuses, b))
 				afterLimiting.push_back(b);
 
-		afterLimiting.getBonuses(*ret, selector, limit);
 	}
+	afterLimiting.getBonuses(*ret, selector, limit);
+	ret->stackBonuses();
 	return ret;
 }
 
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -245,6 +245,39 @@ void BonusList::changed()
 		CBonusSystemNode::treeHasChanged();
 }
 
+BonusList BonusList::stackingBonuses() const
+{
+	BonusList result(*this);
+	boost::sort(result.bonuses, [](std::shared_ptr<Bonus> b1, std::shared_ptr<Bonus> b2) -> bool
+	{
+		if(b1 == b2)
+			return false;
+		if(b1->stacking != b2->stacking)
+			return b1->stacking < b2->stacking;
+		if(b1->valType != b2->valType)
+			return b1->valType < b2->valType;
+		return b1->val > b2->val;
+	});
+	// remove non-stacking
+	int next = 1;
+	while(next < result.size())
+	{
+		bool remove;
+		if (result[next]->stacking.empty())
+			remove = result[next] == result[next-1];
+		else if (result[next]->stacking == "ALWAYS")
+			remove = false;
+		else
+			remove = result[next]->stacking == result[next-1]->stacking
+				&& result[next]->valType == result[next-1]->valType;
+		if(remove)
+			result.bonuses.erase(result.begin() + next);
+		else
+			next++;
+	}
+	return result;
+}
+
 int BonusList::totalValue() const
 {
 	int base = 0;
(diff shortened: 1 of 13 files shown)


alexvins:
It is confusing for me that  `"stacking" : "creature.angel"` means **no** stacking with other "creature.angel' marked bonus.

lib/HeroBonus.h
```diff
@@ -345,6 +345,12 @@ struct DLL_LINKAGE Bonus : public std::enable_shared_from_this<Bonus>
 #undef BONUS_VALUE
 	};
 
+	enum Stacking
+	{
+		STACKING_OTHER = -1,
+		STACKING_ALWAYS = -2
+	};
```
alexvins: I suggest using enum class
```
enum class Stacking : si32 
{
OTHER = -1,
ALWAYS = -2
}
```

DjWarmonger:
Why the dotted syntax? `creature.angel` In this case it doesn't seem the items separated by dot refer to any identifier in terms of mechanics, so can be replaced by simple string value `creatureAngel` or, for instance `stablesBonus`.

The Stables movement bonus is handled explicitely in the code BTW, would be nice if it could be unified though.

henningkoehlernz:
You're right, that might be more elegant.

lib/HeroBonus.cpp
```diff
@@ -375,16 +407,9 @@ int BonusList::valOfBonuses(const CSelector &select) const
 	BonusList ret;
 	CSelector limit = nullptr;
 	getBonuses(ret, select, limit);
-	ret.eliminateDuplicates();
 	return ret.totalValue();
 }
 
-void BonusList::eliminateDuplicates()
```
alexvins: You should put stacking code here, not to `totalValue()`
* cached BonusList`s will be preprocessed
* no additional checks will be needed where `getAllBonuses` used directly

henningkoehlernz: I think conceptually stacking is part of total value calculation, whereas getAllBonuses should really return all bonuses, not just some. Also, `IBonusBearer::valOfBonuses` uses caching as well, so I doubt it makes much of a difference performance-wise.

alexvins: You put stacking after cache request but it was before (f.e. in CBonusSystemNode::getAllBonuses).
And it should be no additional stacking processing inside BonusList. Stacking is IBonusBearer responsibility.

henningkoehlernz: Just because duplicates were eliminated as part of getAllBonuses before doesn't mean that's a good approach. And I don't see why stacking should be the responsibility of IBonusBearer - totalValue calculation is part of BonusList because all you need is a list of bonuses, and the same argument applies to stacking.

lib/HeroBonus.cpp
```diff
@@ -1198,27 +1222,36 @@ std::string Bonus::Description() const
 	std::ostringstream str;
 
 	if(description.empty())
-		switch(source)
+	{
+		if(stacking.empty() || stacking == "ALWAYS")
```
alexvins: Does this change hint that `stacking` is actually new special kind of bonus source?

henningkoehlernz: Yes, the stacking string now shows up as the source:
Morale Info

alexvins: I mean can we reuse bonus source/ source id field for that and do not add stacking field?

henningkoehlernz: No, bonus source is used for other things, e.g. we might have different secondary skills granting non-stacking bonuses, but setting them to same source id means we can't remove them easily when upgrading to the next level. Also bonus source is set automatically throughout the code, and having it identify the actual source is helpful for debugging. And finally we may want bonuses from different source types to not stack.

henningkoehlernz:
There are a few bonuses that are not using the new system, but I don't plan to address those in this updates for the following reasons:
- Stables: There are two types of stables (castle structure and dwelling) which check that the player does not already have a bonus before adding one. This is sensible behaviour though, as we wouldn't want to add a new bonus each time a hero visits anyways. What could be changed is that the two structures add different (non-stacking) bonuses, but as they are currently hard-coded there's currently little point to this.
- Mages/Pegasi: The calculation of spell cost bonuses/penalties is currently using special code. The reason for this is that propagation of bonuses to the enemy army is currently a mess of work-arounds, which will need to be addressed first.

lib/HeroBonus.cpp
```diff
@@ -246,6 +245,45 @@ void BonusList::changed()
...
+#define COMPARE_ATT(ATT) if(b1->ATT != b2->ATT) return b1->ATT < b2->ATT
+		COMPARE_ATT(stacking);
+		COMPARE_ATT(type);
+		COMPARE_ATT(subtype);
+		COMPARE_ATT(valType);
+#undef COMPARE_ATT
+		return b1->val > b2->val;
+	});
+	// remove non-stacking
+	int next = 1;
```
alexvins: next should be `size_t` or even better vector iterator

lib/HeroBonus.cpp
```diff
@@ -1198,27 +1222,36 @@ std::string Bonus::Description() const
...
+				str << VLC->skillh->skillName(sid);
+				break;
+			default:
+				//todo: handle all possible sources
+				str << "Unknown";
+				break;
+			}
 		}
+		else
+			str << stacking;
```
alexvins: This is OK for now, but we should have separate field for stacked bonus description, so it could be localized without affecting mechanics, server and client can have different locales.

lib/HeroBonus.cpp
```diff
@@ -1283,6 +1316,8 @@ JsonNode Bonus::toJsonNode() const
 		root["val"].Integer() = val;
 	if(valType != ADDITIVE_VALUE)
 		root["valueType"].String() = vstd::findKey(bonusValueMap, valType);
+	if(stacking != "")
+		root["stacking"].String() = stacking;
```
alexvins: Some fields are missing here (f.e. source and source id). Is that intentional?

lib/JsonNode.cpp
```diff
@@ -598,6 +598,10 @@ bool JsonUtils::parseBonus(const JsonNode &ability, Bonus *b)
 	if (!value->isNull())
 		b->valType = static_cast<Bonus::ValueType>(parseByMap(bonusValueMap, value, "value type "));
 
+	value = &ability["stacking"];
+	if (!value->isNull())
+		b->stacking = value->String();
```
alexvins: No need to check for `isNull()` here, `String()` return `""` for null node.

test/mock/mock_BonusBearer.cpp
```diff
@@ -28,10 +28,7 @@ void BonusBearerMock::addNewBonus(const std::shared_ptr<Bonus> & b)
 const TBonusListPtr BonusBearerMock::getAllBonuses(const CSelector & selector, const CSelector & limit, const CBonusSystemNode * root, const std::string & cachingStr) const
 {
 	if(cachedLast != treeVersion)
-	{
-		bonuses.eliminateDuplicates();
 		cachedLast = treeVersion;
```
alexvins: Put stacking here.

lib/HeroBonus.cpp
henningkoehlernz: Yes - this function is being used to convert bonuses to a format you would store in a config file - here source and source_id wouldn't make an appearance here.

lib/HeroBonus.cpp
henningkoehlernz: Couldn't we just use the existing description field for this?

alexvins: We could.

lib/HeroBonus.cpp
alexvins: But what about `duration`, `turnsRemain`, `description`?

henningkoehlernz: Not `turnsRemain` I think, as that is dynamic, but `duration` and `description` would make sense. Added this functions when converting old bonus formats into new, currently it's just used for creating log outputs that can be copy-pasted if needed.

henningkoehlernz: Actually, it looks like there is duplicated functionality with `JsonUtils::unparseBonus`, though it seems that the latter is never used anywhere.

alexvins: turnsRemain is dynamic but configuration value is used as initial value for temporary bonuses (f.e. for spells it by default set to spellpower, but could be explicitly defined too)

henningkoehlernz: fixed / cleaned up

lib/HeroBonus.cpp
```diff
@@ -1318,10 +1309,20 @@ JsonNode Bonus::toJsonNode() const
 		root["valueType"].String() = vstd::findKey(bonusValueMap, valType);
 	if(stacking != "")
 		root["stacking"].String() = stacking;
+	if(description != "")
+		root["description"].String() = description;
+	if(effectRange != NO_LIMIT)
+		root["effectRange"].String() = vstd::findKey(bonusLimitEffect, effectRange);
+	if(duration != PERMANENT)
+		root["duration"].String() = vstd::findKey(bonusDurationMap, duration);
```
alexvins: This is wrong, duration is a bitmask

henningkoehlernz: Out of interest - if the duration set describes "when bonus is automatically removed", then what is the purpose of having `PERMANENT = 1` instead of `PERMANENT = 0`? If there is any ...

alexvins: duration is bitmask it can not have 0 value

henningkoehlernz: What I'm saying is that 0 = no bit set = not removal case since no bit set, which is logically the same as permanent duration.

alexvins: May be because it was not bitmask initially

ArseniyShestakov:
So I felt there can be something bad going and I managed to break this one.
Reproduction:

1. Start any map.
2. Get one stack of Angels and one of Archangels. There must be 2 or more creatures in stack.
3. Now go into hero window and start swap position of stacks with each other and then splitting them.
4. After split there will be two bonuses from either Angels or Archangels and you will get actually +2 to morale from them and it's will be visible on client-side even in battle. I didn't checked if there any actual effect, but at least client see +2 morale and who appropriate icons.

So it's not exactly 100% reproduction, but I sure you'll manage.

PR #435, opened 2018-03-21
Added uncapped by att/def damage feature

dydzio0614:
Added hardcoded feature that may in some cases be fun and useful for custom maps designed to use it in connection with VCMI primary stats not capping at 99. 
With this feature enabled damage can go down to 1 when attacking enemy stack with huge amount of defense points, and damage increase for huge amount of attack points is uncapped as well.


lib/battle/CBattleInfoCallback.cpp
```diff
@@ -707,12 +708,12 @@ TDmgRange CBattleInfoCallback::calculateDmgRange(const BattleAttackInfo & info)
 	//bonus from attack/defense skills
 	if(attackDefenceDifference < 0) //decreasing dmg
 	{
-		const double dec = std::min(0.025 * (-attackDefenceDifference), 0.7);
+		const double dec = !VLC->modh->settings.NO_ATT_AND_DEF_DAMAGE_INFLUENCE_LIMIT ? std::min(0.025 * (-attackDefenceDifference), 0.7) : 0.025 * (-attackDefenceDifference);
```
alexvins: It is better to make `0.7` and `4.0` constants configurable

dydzio0614: How that should be implemented for json configuration? 2 separate hardcoded features with int parameter?

alexvins: 3 hardcoded features with float parameter (0.025 might be configurable too)

lib/CModHandler.cpp
```diff
@@ -663,6 +663,14 @@ void CModHandler::loadConfigFromFile (std::string name)
 	logMod->debug("\tWINNING_HERO_WITH_NO_TROOPS_RETREATS\t%d", static_cast<int>(settings.WINNING_HERO_WITH_NO_TROOPS_RETREATS));
 	settings.BLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE = hardcodedFeatures["BLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE"].Bool();
 	logMod->debug("\tBLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE\t%d", static_cast<int>(settings.BLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE));
+	settings.ATTACK_POINT_DMG_MULTIPLIER = hardcodedFeatures["ATTACK_POINT_DMG_MULTIPLIER"].Float();
+	logMod->debug("\tATTACK_POINT_DMG_MULTIPLIER\t%d", static_cast<float>(settings.ATTACK_POINT_DMG_MULTIPLIER));
```
alexvins: Use %f for floats

lib/CModHandler.h
```diff
@@ -313,6 +318,21 @@ class DLL_LINKAGE CModHandler
 			{
 				BLACK_MARKET_MONTHLY_ARTIFACTS_CHANGE = true;
 			}
+			
+			if(version >= 782)
```
alexvins: Wrong version, yours is 788

dydzio0614: Version problem git fixed in merge commit

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -705,15 +706,15 @@ TDmgRange CBattleInfoCallback::calculateDmgRange(const BattleAttackInfo & info)
...
-		const double dec = std::min(0.025 * (-attackDefenceDifference), 0.7);
-		multBonus *= 1.0 - dec;
-	}
-	else //increasing dmg
-	{
-		const double inc = std::min(0.05 * attackDefenceDifference, 4.0);
-		additiveBonus += inc;
+	if(attackDefenceDifference < 0) //decreasing dmg
+	{
+		const double dec = std::min(VLC->modh->settings.DEFENSE_POINT_DMG_MULTIPLIER * (-attackDefenceDifference), VLC->modh->settings.DEFENSE_POINTS_DMG_MULTIPLIER_CAP);
```
alexvins: Too long lines, copy settings to local constants first.

lib/CModHandler.h
alexvins: But it looks like its not.

lib/CModHandler.cpp
dydzio0614: I cast as I follow another positions in hardcoded features, and did not find support for double so casted it to float, relying on automatic float->double conversion after doing that

alexvins: `static_cast<int>` is here for boolean fields. Doubles should print fine with `%f`

dydzio0614: Ah yes, I thought you mean 
hardcodedFeatures["ATTACK_POINTS_DMG_MULTIPLIER_CAP"].Float()

alexvins:
@dydzio0614 please rebase this branch.

dydzio0614:
In my VCMI fork the branch is rebased, here in pull request it is not refreshed... what next?

nullkiller:
This might be useful but a bit outdated thing. @dydzio0614 any chance to rebase it on top of develop?

dydzio0614:
I will look at this after I buy new PC, should happen before version 1.1 stabilization starts
Should not be hard to remake from scratch though

PR #437, opened 2018-03-22
Fix Issue 2214

henningkoehlernz:
See https://bugs.vcmi.eu/view.php?id=2214


ArseniyShestakov:
How is this going to work with taverns on adventure map?

ArseniyShestakov:
I'm sorry this was a dumb question and this is why I removed it. :-)

PR #438, opened 2018-03-24
Allow the creation of new secondary skills

henningkoehlernz:
Skills gain the following new attributes (required for new skills, auto-populated for existing):
* Images used for basic, advanced and expert level
* Default probabilities to gain them for might/magic heroes (heroClass can override)

Mod for testing is available at https://github.com/henningkoehlernz/vcmi-mods (new_skills)

### DIFF
--- a/client/Graphics.cpp
+++ b/client/Graphics.cpp
@@ -24,6 +24,7 @@
 #include "../lib/CGeneralTextHandler.h"
 #include "../lib/CCreatureHandler.h"
 #include "CBitmapHandler.h"
+#include "../lib/CSkillHandler.h"
 #include "../lib/spells/CSpellHandler.h"
 #include "../lib/CGameState.h"
 #include "../lib/JsonNode.h"
@@ -442,5 +443,17 @@ void Graphics::initializeImageLists()
 		addImageListEntry(spell->id+1, "SPELLINT", spell->iconEffect);
 		addImageListEntry(spell->id, "SPELLBON", spell->iconScenarioBonus);
 		addImageListEntry(spell->id, "SPELLSCR", spell->iconScroll);
+	}
+
+	for(const CSkill * skill : CGI->skillh->objects)
+	{
+		for(int level = 1; level <= 3; level++)
+		{
+			int frame = 2 + level + 3 * skill->id;
+			const CSkill::LevelInfo & skillAtLevel = skill->at(level);
+			addImageListEntry(frame, "SECSK32", skillAtLevel.iconSmall);
+			addImageListEntry(frame, "SECSKILL", skillAtLevel.iconMedium);
+			addImageListEntry(frame, "SECSK82", skillAtLevel.iconLarge);
+		}
 	}
 }
--- a/lib/CHeroHandler.cpp
+++ b/lib/CHeroHandler.cpp
@@ -20,6 +20,7 @@
 #include "CModHandler.h"
 #include "CTownHandler.h"
 #include "mapObjects/CObjectHandler.h" //for hero specialty
+#include "CSkillHandler.h"
 #include <math.h>
 
 #include "mapObjects/CObjectClassesHandler.h"
@@ -117,9 +118,15 @@ CHeroClass * CHeroClassHandler::loadFromJson(const JsonNode & node, const std::s
 		heroClass->primarySkillHighLevel.push_back(node["highLevelChance"][pSkill].Float());
 	}
 
-	for(const std::string & secSkill : NSecondarySkill::names)
+	for(auto skillPair : node["secondarySkills"].Struct())
 	{
-		heroClass->secSkillProbability.push_back(node["secondarySkills"][secSkill].Float());
+		int probability = skillPair.second.Integer();
+		VLC->modh->identifiers.requestIdentifier(skillPair.second.meta, "skill", skillPair.first, [heroClass, probability](si32 skillID)
+		{
+			if(heroClass->secSkillProbability.size() <= skillID)
+				heroClass->secSkillProbability.resize(skillID + 1, -1); // -1 = override with default later
+			heroClass->secSkillProbability[skillID] = probability;
+		});
 	}
 
 	VLC->modh->identifiers.requestIdentifier ("creature", node["commander"],
@@ -241,6 +248,17 @@ void CHeroClassHandler::afterLoadFinalization()
 			float chance = heroClass->defaultTavernChance * faction->town->defaultTavernChance;
 			heroClass->selectionProbability[faction->index] = static_cast<int>(sqrt(chance) + 0.5); //FIXME: replace with std::round once MVS supports it
 		}
+		// set default probabilities for gaining secondary skills where not loaded previously
+		heroClass->secSkillProbability.resize(VLC->skillh->size(), -1);
+		for(int skillID = 0; skillID < VLC->skillh->size(); skillID++)
+		{
+			if(heroClass->secSkillProbability[skillID] < 0)
+			{
+				const CSkill * skill = (*VLC->skillh)[SecondarySkill(skillID)];
+				logMod->trace("%s: no probability for %s, using default", heroClass->identifier, skill->identifier);
+				heroClass->secSkillProbability[skillID] = skill->gainChance[heroClass->affinity];
+			}
+		}
 	}
 
 	for (CHeroClass * hc : heroClasses)
@@ -277,11 +295,6 @@ CHeroHandler::CHeroHandler()
 {
 	VLC->heroh = this;
 
-	for (int i = 0; i < GameConstants::SKILL_QUANTITY; ++i)
-	{
-		VLC->modh->identifiers.registerObject("core", "skill", NSecondarySkill::names[i], i);
-		VLC->modh->identifiers.registerObject("core", "secondarySkill", NSecondarySkill::names[i], i);
-	}
 	loadObstacles();
 	loadTerrains();
 	for (int i = 0; i < GameConstants::TERRAIN_TYPES; ++i)
@@ -943,13 +956,6 @@ std::vector<bool> CHeroHandler::getDefaultAllowed() const
 	return allowedHeroes;
 }
 
-std::vector<bool> CHeroHandler::getDefaultAllowedAbilities() const
-{
-	std::vector<bool> allowedAbilities;
-	allowedAbilities.resize(GameConstants::SKILL_QUANTITY, true);
-	return allowedAbilities;
-}
-
 si32 CHeroHandler::decodeHero(const std::string & identifier)
 {
 	auto rawId = VLC->modh->identifiers.getIdentifier("core", "hero", identifier);
@@ -962,18 +968,4 @@ si32 CHeroHandler::decodeHero(const std::string & identifier)
 std::string CHeroHandler::encodeHero(const si32 index)
 {
 	return VLC->heroh->heroes.at(index)->identifier;
-}
-
-si32 CHeroHandler::decodeSkill(const std::string & identifier)
-{
-	auto rawId = VLC->modh->identifiers.getIdentifier("core", "skill", identifier);
-	if(rawId)
-		return rawId.get();
-	else
-		return -1;
-}
-
-std::string CHeroHandler::encodeSkill(const si32 index)
-{
-	return NSecondarySkill::names[index];
 }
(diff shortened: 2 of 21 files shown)


Alucard648:
Will it be possible in future to have more than 3 mastery levels of the same secondary skill?

henningkoehlernz:
@Alucard648 Not with this PR, but planning to look into it.

lib/CGameInfoCallback.cpp
```diff
@@ -55,7 +55,11 @@ bool CGameInfoCallback::isAllowed( int type, int id )
 	case 1:
 		return gs->map->allowedArtifact[id];
 	case 2:
-		return gs->map->allowedAbilities[id];
+		// new skills are always allowed
+		if(id >= gs->map->allowedAbilities.size())
```
ArseniyShestakov: Our editor and VCMI map format will allow new skills disablement once this branch is landed so you need to support that. I 99% sure this already work for custom spells so you'll likely just need to copy-paste a bit of code used for ```allowedSpell```.

lib/CHeroHandler.h
```diff
@@ -134,6 +134,9 @@ std::vector<std::shared_ptr<Bonus>> SpecialtyBonusToBonuses(const SSpecialtyBonu
 
 class DLL_LINKAGE CHeroClass
 {
+private:
+	// used to store mapping until secondary skills have been loaded
+	std::map<std::string, int> secSkillProbabilityMap;
```
ArseniyShestakov: Is this needed because hero handler loaded before everything else?
I don't really see if SkillHandler depend on anything else so might be we should just initialize it first?

henningkoehlernz: Yes, that's the reason, and you're right - prescribing a load order for handlers would get around the need for this workaround. However, the way I understand it this isn't a good fit with the way things currently work - load order is essentially random (determined by iterator over a map), with anything that requires other things to be loaded first being handled in `afterLoadFinalization`. This lets us deal with cyclic dependencies. There aren't any for SkillHandler, but I think changing the system would require some more discussion.

alexvins: There is special API for such cases `CIdentifierStorage::requestIdentifier` to call functor just after identifier registered
Example from spell handler:
```
	for(const auto & node : json["gainChance"].Struct())
	{
		const int chance = node.second.Integer();

		VLC->modh->identifiers.requestIdentifier(node.second.meta, "faction",node.first, [=](si32 factionID)
		{
			spell->probabilities[factionID] = chance;
		});
	}
```

lib/CGameInfoCallback.cpp
alexvins: This just need to be reverted `allowedAbilities.size()` shall equal actual number of skills. And getDefaultAllowedAbilities() should be moved to SkillHandler

lib/CSkillHandler.h
```diff
@@ -19,6 +19,7 @@ class DLL_LINKAGE CSkill // secondary skill
 	struct LevelInfo
 	{
 		std::string description; //descriptions of spell for skill level
+		std::vector<std::string> images; //icons used to display skill - [0]=small, [1]=medium, [2]=large
```
alexvins: Make this 3 fields f.e. iconSmall, iconMedium, iconLarge

lib/CSkillHandler.h
```diff
@@ -27,31 +28,42 @@ class DLL_LINKAGE CSkill // secondary skill
...
+	void setImages(const std::vector<std::string> & images, int level);
 	const std::vector<std::shared_ptr<Bonus>> & getBonus(int level) const;
 	const std::string & getDescription(int level) const;
+	const std::vector<std::string> & getImages(int level) const;
 	std::string toString() const;
 
 	SecondarySkill id;
 	std::string identifier;
 	std::string name; //as displayed in GUI
+	std::vector<int> gainChance; // gainChance[0/1] = default gain chance on level-up for might/magic heroes
```
alexvins: This should be two scalar fields, or std::array<int32_t, 2>.Also please note that fixed size type highly recommended for serializable fields.

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1107,7 +1107,7 @@ std::vector<SecondarySkill> CGHeroInstance::getLevelUpProposedSecondarySkills()
 	std::vector<SecondarySkill> skills;
 	//picking sec. skills for choice
 	std::set<SecondarySkill> basicAndAdv, expert, none;
-	for(int i=0;i<GameConstants::SKILL_QUANTITY;i++)
+	for(int i = 0; i < VLC->skillh->size(); i++)
```
alexvins: `i` should be `size_t`

lib/mapObjects/MiscObjects.cpp
```diff
@@ -1439,7 +1439,7 @@ void CGWitchHut::initObj(CRandomGenerator & rand)
 {
 	if (allowedAbilities.empty()) //this can happen for RMG. regular maps load abilities from map file
 	{
-		for (int i = 0; i < GameConstants::SKILL_QUANTITY; i++)
+		for (int i = 0; i < VLC->skillh->size(); i++)
```
alexvins: Please fix formatting here. No space after `for`

lib/CHeroHandler.h
henningkoehlernz: The issue with this approach is that I don't just need the skill id, but also the default chance. And the assumption that once the skill id is available the skill will have loaded is dangerous.

alexvins: This assumption come from `CHandlerBase::loadObject` where registerObject called last and used in almost all handlers

henningkoehlernz: What about this (and there might be other similar surprises):
```CHeroHandler::CHeroHandler()
{
	VLC->heroh = this;

	for (int i = 0; i < GameConstants::SKILL_QUANTITY; ++i)
	{
		VLC->modh->identifiers.registerObject("core", "skill", NSecondarySkill::names[i], i);
		VLC->modh->identifiers.registerObject("core", "secondarySkill", NSecondarySkill::names[i], i);
	}
	loadObstacles();
	loadTerrains();
	for (int i = 0; i < GameConstants::TERRAIN_TYPES; ++i)
	{
		VLC->modh->identifiers.registerObject("core", "terrain", GameConstants::TERRAIN_NAMES[i], i);
	}
	loadBallistics();
	loadExperience();
}```

Comment (alexvins) [reply lib/CHeroHandler.h]: You`re right this is yet another legacy

File: lib/CSkillHandler.h [hunk 7]
```diff
@@ -30,30 +32,31 @@ class DLL_LINKAGE CSkill // secondary skill
...
+	const LevelInfo & at(int level) const;
+	LevelInfo & at(int level);
+
 	std::string toString() const;
 
 	SecondarySkill id;
 	std::string identifier;
 	std::string name; //as displayed in GUI
-	std::vector<int> gainChance; // gainChance[0/1] = default gain chance on level-up for might/magic heroes
+	std::array<int, 2> gainChance; // gainChance[0/1] = default gain chance on level-up for might/magic heroes
```
alexvins: use explicit size (int32_t or int64_t) if field to be serialized

lib/CModHandler.cpp
```diff
@@ -84,6 +84,15 @@ static std::pair<std::string, std::string> splitString(std::string input, char s
 	return ret;
 }
 
+std::string CIdentifierStorage::fullName(const std::string & name, const std::string & type)
```
alexvins: There is `CModHandler::makeFullIdentifier` already

lib/CHeroHandler.cpp
```diff
@@ -277,11 +295,6 @@ CHeroHandler::CHeroHandler()
 {
 	VLC->heroh = this;
 
-	for (int i = 0; i < GameConstants::SKILL_QUANTITY; ++i)
-	{
-		VLC->modh->identifiers.registerObject("core", "skill", NSecondarySkill::names[i], i);
-		VLC->modh->identifiers.registerObject("core", "secondarySkill", NSecondarySkill::names[i], i);
```
alexvins: `secondarySkill` registration is missing now?

client/Graphics.cpp
```diff
@@ -443,4 +444,15 @@ void Graphics::initializeImageLists()
 		addImageListEntry(spell->id, "SPELLBON", spell->iconScenarioBonus);
 		addImageListEntry(spell->id, "SPELLSCR", spell->iconScroll);
 	}
+
+	for(const CSkill * skill : CGI->skillh->objects)
+	{
+		for(int level = 1; level <= 3; level++)
+		{
+			int frame = 2 + level + 3 * skill->id;
+			addImageListEntry(frame, "SECSK32", skill->at(level).iconSmall);
```
alexvins: Just a bit to much duplication, please extract `skill->at(level)` to local variable.

lib/CHeroHandler.cpp
```diff
@@ -117,9 +118,15 @@ CHeroClass * CHeroClassHandler::loadFromJson(const JsonNode & node, const std::s
...
-	for(const std::string & secSkill : NSecondarySkill::names)
+	for(auto skillPair : node["secondarySkills"].Struct())
 	{
-		heroClass->secSkillProbability.push_back(node["secondarySkills"][secSkill].Float());
+		int probability = skillPair.second.Integer();
+		VLC->modh->identifiers.requestIdentifier(skillPair.second.meta, "skill", skillPair.first, [heroClass, probability](si32 skillID)
+		{
+			if(heroClass->secSkillProbability.size() <= skillID)
+				heroClass->secSkillProbability.resize(skillID + 1, -1); // -1 = override with default later
+			heroClass->secSkillProbability[skillID] = probability;
```
alexvins: Far out of scope of this PR, but seems like heroclass has problems with `int` in persistent data. And `probability` here should be `si32`

lib/CSkillHandler.cpp
```diff
@@ -31,16 +31,9 @@ CSkill::LevelInfo::~LevelInfo()
...
 {
-	if(id == SecondarySkill::DEFAULT)
-		identifier = "default";
-	else
-		identifier = NSecondarySkill::names[id];
-	// init levels
-	LevelInfo emptyLevel;
-	for(int level = 1; level < NSecondarySkill::levels.size(); level++)
-		levels.push_back(emptyLevel);
+	levels.resize(3);
```
alexvins: `3` should be global constant for now, more over higher skill level feature has been requested.

lib/CSkillHandler.cpp
```diff
@@ -31,16 +31,9 @@ CSkill::LevelInfo::~LevelInfo()
 {
 }
 
-CSkill::CSkill(SecondarySkill id) : id(id)
+CSkill::CSkill(SecondarySkill id, std::string identifier) : id(id), identifier(identifier)
```
alexvins: Put initialization on next line

lib/CSkillHandler.cpp
```diff
@@ -56,19 +49,14 @@ void CSkill::addNewBonus(const std::shared_ptr<Bonus> & b, int level)
 	levels[level-1].effects.push_back(b);
 }
 
-void CSkill::setDescription(const std::string & desc, int level)
+const CSkill::LevelInfo & CSkill::at(int level) const
 {
-	levels[level-1].description = desc;
+	return levels[level - 1];
```
alexvins: valid level should be checked

lib/CSkillHandler.cpp
```diff
@@ -182,10 +168,15 @@ CSkill * CSkillHandler::loadFromJson(const JsonNode & json, const std::string &
 		for(auto b : levelNode["effects"].Struct())
 		{
 			auto bonus = JsonUtils::parseBonus(b.second);
-			bonus->sid = skill->id;
 			skill->addNewBonus(bonus, level);
 		}
-		skill->setDescription(levelNode["description"].String(), level);
+		skill->at(level).description = levelNode["description"].String();
+		if(!levelNode["images"].isNull())
```
alexvins: This check not needed, you can call `levelNode["images"]["small"].String()` on Null node. But you could change it to check for `JsonType::DATA_STRUCT`

lib/CHeroHandler.cpp
henningkoehlernz: fixed.

lib/CHeroHandler.h
```diff
@@ -192,6 +192,8 @@ class DLL_LINKAGE CHeroClass
 		h & imageMapFemale;
 	}
 	EAlignment::EAlignment getAlignment() const;
+
+	friend class CHeroClassHandler;
```
ArseniyShestakov: I suppose this one no longer needed, right?

PR #439, opened 2018-03-29
Enable Limiter nesting with AllOf, AnyOf, NoneOf

henningkoehlernz:
Samples of new format (do not inherit to undead or devils):

	"limiters" : [
		"noneOf",
		"IS_UNDEAD",
		{
			"type" : "CREATURE_TYPE_LIMITER",
			"parameters" : [ "devil", true ]
		}
	]

Or the logical equivalent (to show off nesting)

	"limiters" : [
		"allOf",
		[
			"noneOf",
			"IS_UNDEAD"
		],
		[
			"noneOf",
			{
				"type" : "CREATURE_TYPE_LIMITER",
				"parameters" : [ "devil", true ]
			}
		]
	]

Or in compact format (allOf prefix is optional)

	"limiters" : [
		[ "noneOf", "IS_UNDEAD" ],
		[ "noneOf",
			{
				"type" : "CREATURE_TYPE_LIMITER",
				"parameters" : [ "devil", true ]
			}
		]
	]

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -1171,7 +1171,7 @@ void CBonusSystemNode::limitBonuses(const BonusList &allBonuses, BonusList &out)
 		for(int i = 0; i < undecided.size(); i++)
 		{
 			auto b = undecided[i];
-			BonusLimitationContext context = {b, *this, out};
+			BonusLimitationContext context = {b, *this, out, undecided};
 			int decision = b->limiter ? b->limiter->limit(context) : ILimiter::ACCEPT; //bonuses without limiters will be accepted by default
 			if(decision == ILimiter::DISCARD)
 			{
@@ -1660,6 +1660,10 @@ int HasAnotherBonusLimiter::limit(const BonusLimitationContext &context) const
 	//if we have a bonus of required type accepted, limiter should accept also this bonus
 	if(context.alreadyAccepted.getFirst(mySelector))
 		return ACCEPT;
+
+	//if there are no matching bonuses pending, we can (and must) reject right away
+	if(!context.stillUndecided.getFirst(mySelector))
+		return DISCARD;
 
 	//do not accept for now but it may change if more bonuses gets included
 	return NOT_SURE;
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -1347,7 +1347,7 @@ JsonNode Bonus::toJsonNode() const
 	if(turnsRemain)
 		root["turns"].Integer() = turnsRemain;
 	if(limiter)
-		root["limiters"].Vector().push_back(limiter->toJsonNode());
+		root["limiters"] = limiter->toJsonNode();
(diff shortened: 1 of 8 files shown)


lib/HeroBonus.cpp
```diff
@@ -1840,7 +1842,49 @@ int LimiterList::limit( const BonusLimitationContext &context ) const
 	return wasntSure ? ILimiter::NOT_SURE : ILimiter::ACCEPT;
 }
 
-void LimiterList::add( TLimiterPtr limiter )
+void AllOfLimiter::add(TLimiterPtr limiter)
+{
+	limiters.push_back(limiter);
+}
+
+int AnyOfLimiter::limit(const BonusLimitationContext & context) const
```
alexvins: All that `limit` functions duplicate each other at 99% and (important) differences are not that easy to see. You should merge "logical" limiters into one class

lib/JsonNode.cpp
```diff
@@ -567,6 +567,96 @@ void JsonUtils::resolveIdentifier(const JsonNode &node, si32 &var)
...
+	case JsonNode::JsonType::DATA_STRING: //pre-defined limiters
+		return parseByMap(bonusLimiterMap, &limiter, "limiter type ");
+		break;
+	case JsonNode::JsonType::DATA_STRUCT: //customizable limiters
+		{
+			std::string limiterType = limiter["type"].String();
+			const JsonVector & parameters = limiter["parameters"].Vector();
+			if(limiterType == "CREATURE_TYPE_LIMITER")
+			{
+				std::shared_ptr<CCreatureTypeLimiter> creatureLimiter = std::make_shared<CCreatureTypeLimiter>();
```
ArseniyShestakov: I remember that you dislike using ```auto``` and I agree it make sense in some parts of code where readability is more important. Yet I don't think object creation like this is one of these cases.

henningkoehlernz: I agree it's not needed here, but I don't see it causing any harm either.

ArseniyShestakov:
Yes please add some note into change log. This change is important enough to mention it there.

lib/JsonNode.cpp
ArseniyShestakov: It's not really important. I just seen there is ```auto``` in other methods so I just noticed this one.

lib/registerTypes/RegisterTypes.h
```diff
@@ -139,6 +139,9 @@ void registerTypesMapObjectTypes(Serializer &s)
 	s.template registerType<IUpdater, GrowsWithLevelUpdater>();
 	s.template registerType<IUpdater, TimesHeroLevelUpdater>();
 	s.template registerType<IUpdater, TimesStackLevelUpdater>();
+
+	s.template registerType<ILimiter, AnyOfLimiter>();
+	s.template registerType<ILimiter, NoneOfLimiter>();
```
alexvins: You should adjust serialize format version when changing type registration

PR #440, opened 2018-03-31
Gui cleanup3

alexvins:
* IImage is now being stored and used as shared_ptr
* Spell window has been reworked
* Changed most gui classes to use smart pointers

### DIFF
--- a/AI/VCAI/VCAI.cpp
+++ b/AI/VCAI/VCAI.cpp
@@ -447,7 +447,7 @@ void VCAI::advmapSpellCast(const CGHeroInstance * caster, int spellID)
 	NET_EVENT_HANDLER;
 }
 
-void VCAI::showInfoDialog(const std::string &text, const std::vector<Component*> &components, int soundID)
+void VCAI::showInfoDialog(const std::string &text, const std::vector<Component> &components, int soundID)
 {
 	LOG_TRACE_PARAMS(logAi, "soundID '%i'", soundID);
 	NET_EVENT_HANDLER;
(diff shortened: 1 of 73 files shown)


alexvins:
Enough refactoring for now. Next part to be done in separate PR after #402 merge.

ArseniyShestakov:
Ok first bug found: town selection is broken. Hero selection works, but town selection doesn't.

ArseniyShestakov:
Next hero button also broken.

ArseniyShestakov:
Next: "Scroll hero list" upper buttons always active and don't have right click text.

ArseniyShestakov:
Sleep / Wake also broken, should change icon and switch for next hero which doesn't happen.

ArseniyShestakov:
In hero levelup window label telling you ```%s is not level %d %s``` is missing.

screenshot_20180405_225245

alexvins:
"Sleep / Wake" is not a regression, selection and levelup fixed.

alexvins:
"Sleep / Wake" is not a regression, selection and levelup fixed.

ArseniyShestakov:
Ok problems above that are not regressions in develop are fixed.

ArseniyShestakov:
Another problem: right click on skills icons in levelup window doesn't bring popup.

alexvins:
fixed both

ArseniyShestakov:
One more regression, it's feels like main menu object not being fully destroyed when gameplay start. This in fact could be fault of my own branch, that's possible. But seems like this branch exposed problem even more.

Yet in this branch when you open "Artifact Merchant" menu and then left click on bottom middle artifact in the right I get "high scores menu not implemented" error and seems like if I click down there I also attempt to "exit game" via main menu and get dialog about it. Also once I got same thing appeared in battle so it's seems like universal problem.

PR #441, opened 2018-04-03
Configure session parameters in settings.json

henningkoehlernz:
Allows specification of session parameters (in particular `savefrequency`) in settings.json.

- command-line arguments, if present, override settings in settings.json
- changes to session parameters via command-line arguments do not change settings.json
- setting savefrequence to a value <= 0 disables autosaves

Note: currently `savefrequency` is the only parameters to be configured in settings.json (can easily add more though by adding them to config/schema/settings.json)

### DIFF
--- a/include/vstd/StringUtils.h
+++ b/include/vstd/StringUtils.h
@@ -1,0 +1,6 @@
+namespace vstd
+{
+
+	DLL_LINKAGE std::vector<std::string> split(std::string s, std::string separators);
+
+}
--- a/lib/CConfigHandler.cpp
+++ b/lib/CConfigHandler.cpp
@@ -42,7 +42,7 @@ SettingsStorage::NodeAccessor<Accessor>::operator Accessor() const
 }
 
 template<typename Accessor>
-SettingsStorage::NodeAccessor<Accessor> SettingsStorage::NodeAccessor<Accessor>::operator () (std::vector<std::string> _path)
+SettingsStorage::NodeAccessor<Accessor> SettingsStorage::NodeAccessor<Accessor>::operator () (std::vector<std::string> _path) const
 {
 	std::vector<std::string> newPath = path;
 	newPath.insert( newPath.end(), _path.begin(), _path.end());
@@ -51,11 +51,12 @@ SettingsStorage::NodeAccessor<Accessor> SettingsStorage::NodeAccessor<Accessor>:
 
 SettingsStorage::SettingsStorage():
 	write(NodeAccessor<Settings>(*this, std::vector<std::string>() )),
-	listen(NodeAccessor<SettingsListener>(*this, std::vector<std::string>() ))
+	listen(NodeAccessor<SettingsListener>(*this, std::vector<std::string>() )),
+	autoSaveConfig(false)
 {
 }
 
-void SettingsStorage::init()
+void SettingsStorage::init(bool autoSave)
 {
 	std::string confName = "config/settings.json";
 
@@ -67,6 +68,7 @@ void SettingsStorage::init()
 
 	JsonUtils::maximize(config, "vcmi:settings");
 	JsonUtils::validate(config, "vcmi:settings", "settings");
+	autoSaveConfig = autoSave;
 }
 
 void SettingsStorage::invalidateNode(const std::vector<std::string> &changedPath)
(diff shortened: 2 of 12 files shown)


ArseniyShestakov:
Sorry, but all point of "session" parameters is to make sure they are not saved.
Why would one want to save them?

PS: I for instance already moved last save / map into other part of storage in my branch since it's the only data from ```session``` that needs saving.

henningkoehlernz:
The benefit of this approach is that you can set the options both ways, either via config file (if you want them to be persistent) or via command line (if you don't).

ArseniyShestakov:
I just look at the code very fast, but is it right that current code will save any options that I passed via command line to settings.json? If that's the case then it's absolutely not acceptable.

Basically all options other than ```savefrequency``` should never ever be saved after used through command line because their only intention was to let you run testing with 5-20 client instances in background while your normal client you working with is not affected.

ArseniyShestakov:
Ok this way it's better and can be merged. I'll wait on other opinion though (if there will be any).

lib/CConfigHandler.cpp
```diff
@@ -67,6 +67,7 @@ void SettingsStorage::init()
 
 	JsonUtils::maximize(config, "vcmi:settings");
 	JsonUtils::validate(config, "vcmi:settings", "settings");
+	originalConfig = config;
```
alexvins: Should we only save "session" here?

henningkoehlernz: Would be sufficient - I figured we might need whole config at some point in the future, but that's perhaps premature.

alexvins: No session parameters should be in settings at all, but some more settings could be added to CLI.

henningkoehlernz: I haven't used any of the session parameters myself yet (apart from `savefrequency`), but I figured that they were added to sessions because there was a need to configure them on the fly (e.g. to disable autosaves for faster mass-testing).
Do you mean launcher UI rather than CLI? Or to modify non-session parameters from CLI (without saving)?

henningkoehlernz:
Ok, what it does now is the following:
* Client no longer auto-saves settings, only launcher does.
* Client command-line options can override arbitrary settings (no longer restricted to session category)
* Moved `savefrequency` to general, and renamed to `saveFrequency` (in settings.json only, command-line capitalization is unchanged)
* Command-line parameters only override existing config when present.
* A value of `saveFrequency < 1` means no auto-saves of games.

TODO: Add saveFrequency to launcher UI.

Global.cpp
```diff
@@ -0,0 +1,12 @@
+#include "Global.h"
```
alexvins: do not do this. Global.h must be header only

Global.h
```diff
@@ -721,6 +721,8 @@ namespace vstd
 		return false;
 	}
 
+	std::vector<std::string> split(std::string s, std::string separators);
```
alexvins: Add another .h to `include/vstd` dir. Put .cpp to `lib/vstd`

alexvins: And please do not include yet another helper to `Global.h`. I should only contain global compiler configuration and heavy 3rd party includes for PCH.

lib/CMakeLists.txt
```diff
@@ -124,6 +124,8 @@ set(lib_SRCS
 		spells/effects/RemoveObstacle.cpp
 		spells/effects/Sacrifice.cpp
 
+		${CMAKE_HOME_DIRECTORY}/include/vstd/StringUtils.cpp
```
alexvins: should be `vstd/StringUtils.cpp`. Only header should be in include dir.

include/vstd/StringUtils.cpp
```diff
@@ -0,0 +1,12 @@
+#include "StringUtils.h"
+
+namespace vstd {
```
alexvins: put `{` on new line and do not indent namespace contents. We indent namespace if and only if it contains forward declarations

ArseniyShestakov: I don't really like such inconsistency. Will make work on automated formatting much harder...

henningkoehlernz: Agreed regarding inconsistency. Also, whatever we decide should probably go into https://wiki.vcmi.eu/Coding_guideline.

include/vstd/StringUtils.h
```diff
@@ -0,0 +1,5 @@
+namespace vstd {
```
alexvins: put `{` on new line

client/CMakeLists.txt
```diff
@@ -57,6 +57,8 @@ set(client_SRCS
 		lobby/RandomMapTab.cpp
 		lobby/SelectionTab.cpp
 
+		vstd/StringUtils.cpp
```
alexvins: remove this

lib/vstd/StringUtils.cpp
```diff
@@ -0,0 +1,13 @@
+#include <vstd/StringUtils.h>
```
alexvins: `#include "StdInc.h"` must be first line

ArseniyShestakov:
Sorry that I broken it, I'll fix it.

@henningkoehlernz.
1. To rebase into single commit use interactive rebase against e.g develop:
```git rebase -i develop```
Then you just put commits you want to combine inti previous one by ```s``` (squash).

With interactive mode you also able to change order of your commits (as long as there are no conflicts or you resolve them), or edit commit contents or description or do pretty much any history rewrite as you wish.

3. All you need is to do:
```
git fetch REMOTE_NAME
git reset --hard REMOTE_NAME/BRANCH_NAME
```
Other ugly way is to just switch branches, delete force-pushed branch and checkout it again so version from remote is used. Obviously in both cases you'll lose local changes.

lib/vstd/StringUtils.cpp
henningkoehlernz: fixed.

PR #442, opened 2018-04-03
[WIP] AI: Retreat and Surrender

### LINKED ISSUE #976: Battle AI should try to save his powerful heroes
I think currently AI never tries to run away/surrender, while oh3 AI loves running away after throwing Magic Arrow/Lightning Bolt :) Shackles of War is a must there.

In some cases like when AI's quite powerful hero (high stats and/or good artifacts) is losing, he should try to run/surrender and save him.

przecze:
Added `SURRENDER` and `RETREAT` battle action, for now, the enemy simply retreat in the first turn.


ArseniyShestakov:
Just keep in mind that decision for what to do should be passed to actual player interface, be it VCAI or human player. I'll try to prepare some PoC to make communication between BattleAI and player interfaces possible.

przecze:
Ok, so now the enemy chooses to retreat after predicting his failure.
No consultations with the player interface are implemented yet, but it's ready to be merged.

AI/BattleAI/BattleAI.cpp
```diff
@@ -554,12 +558,127 @@ void CBattleAI::print(const std::string &text) const
...
+		{
+			if (!firstRound)
+			{
+				state->nextRound(0);//todo: set actual value?
+				firstRound = false;
+			}
+			for(auto unit : round)
+			{
+				LOGL("---------------> consider turn of " + unit->unitType()->namePl);
+				if (!state->getForUpdate(unit->unitId())->alive())
```
alexvins: Just `(!unit->alive())` should be anough

przecze: well, it's not. If unit is alive now, but is killed in the predicted scenario, `unit->alive()` remains true in contrast to what is done here. But I'll make sure, thanks.

alexvins: I see now, you need to call battleGetTurnOrder on HypotheticBattle  but not `Callback`

ArseniyShestakov:
Ok here is PoC on BattleAI -> VCAI or player interface communication:
https://github.com/ArseniyShestakov/vcmi/commit/53cd5a0dc8e463226ece0ad6a4522673d14876a9

ArseniyShestakov:
I'll also try to get some PoC code that to set goal for VCAI to buy specific hero.
So you'll have something to play with.

ArseniyShestakov:
So here is my thoughs on how this algorithm can be improved.

Just in case here is my suggestion on VCAI decision making:
1. First you'll need to pass current army state within battle to the VCAI one way or another since VCMI only apply changes to actual hero on battle completion. Of course it's possible to access data in ```GameState::curB``` inside VCAI, but I wouldn't recommend that.
2. Then you need to measure current VCAI army strength: just use ```CCreatureSet::getArmyStrength``` for that and measure strength of army for all owned towns and heroes.
3. Then you check what % of player army strength current losing hero's army represent and if it's let's say more than 25% you start consider surrendering instead of retreat.
4. No you check what ```%``` of gold surrender will cost compared to all gold AI player currently has and if it's less than ```25%``` then you go for surrender.

...

Another suggestion is to add more complex loss prediction for the cases where surrender is possible and hero have significant value for VCAI (represent large % of player strength). Instead of checking that you'll definitely lose battle in next **N** turns you can check whatever % of army strength you will lose compared to your enemy.

So in case where VCAI will prefer to to keep the army alive you go for surrender if in next turn you lose let's say 50% army strength while enemy lose less than 10%.

PS: All values can obviously be adjusted or replaced with fuzzy logic n future, but you should get eneral idea.

AI/BattleAI/BattleAI.cpp
przecze: I double checked this thing. There has to be `state->getForUpdate(..)->alive`. Otherwise, it can return alive even if we predict the unit will die.
it also has to be fixed in spellcasting prediction

AI/BattleAI/BattleAI.cpp
```diff
@@ -552,12 +556,139 @@ void CBattleAI::print(const std::string &text) const
...
+					swb->removeUnitBonus(Bonus::UntilAttack);
+				if(ap.damageReceived > 0)
+					swb->removeUnitBonus(Bonus::UntilBeingAttacked);
+
+				LOGL(unit->unitType()->namePl + " will attack:");
+				for(auto affected : ap.affectedUnits)
+				{
+					LOGL("->"+affected->unitType()->namePl);
+					swb = state->getForUpdate(affected->unitId());
+					*swb = *affected;
```
alexvins: There is more generic problem with algorithm (same for spells): after this line queue is actually invalid and should be rebuilt. `battleGetTurnOrder` should be called for each unit and from `state`

przecze:
Some simple logic on the VCAI side is implemented.

BattleAI part needs some fixes, as @alexvins noticed. Will come to that next.

PR #446, opened 2018-04-07
Gui cleanup4

alexvins:


karliss:
Aren't shadow parts https://github.com/vcmi/vcmi/blob/5c09f751b343fc3d46dd5c3f4f08b0614cf3522f/client/windows/CWindowObject.cpp#L223  that were changed in GUI cleanup 3 leaking memory  due to `shadow.defActions` not having DISPOSE flag?

ArseniyShestakov:
@karliss I'm sorry if I misunderstand your point, but all these new branches you see have end goal to get rid of ```DISPOSE``` completely. Though since we change them step-by-step there will be some leaks until refactoring of all UI is complete.

karliss:
No problem, I wasn't sure how clean you you are trying to have the develop branch during this refactoring.

ArseniyShestakov:
@karliss
Regardless thanks for feedback. We usually trying to keep develop as stable as possible so comments like that are really appreciated!

Unfortunately in this particular case we decide to land code as soon as possible since refactoring of UI require huge amount of work and it's easier to test each group of changes separately rather than trying to rewrite it all within single branch.

ArseniyShestakov:
Ok there is regression: when on Linux window minimized and then opened again renderer is recreated.
After that cursor become invisible.

alexvins:
@karliss it should not leak

there is `OBJECT_CONSTRUCTION_CUSTOM_CAPTURING(255);` just before that line

karliss:
@alexvins CIntObject.cpp destructor checks DISPOSE flag in parents defActions and childeren recActions. Having it only for children is insufficient.

alexvins:
@karliss shadow is default constructed `CIntObject` it should have defActions = 255

karliss:
Just confirmed it with a debugger `shadow->defactions == 239` most likely due to https://github.com/vcmi/vcmi/blob/develop/client/gui/CIntObject.cpp#L32 and https://github.com/vcmi/vcmi/blob/develop/client/windows/CWindowObject.cpp#L216 .

alexvins:
OK, fixed

client/battle/CBattleInterface.h
```diff
@@ -121,14 +121,26 @@ class CBattleInterface : public CIntObject
 private:
 	SDL_Surface *background, *menu, *amountNormal, *amountNegative, *amountPositive, *amountEffNeutral, *cellBorders, *backgroundWithHexes;
 
-	CButton *bOptions, *bSurrender, *bFlee, *bAutofight, *bSpell,
-		* bWait, *bDefence, *bConsoleUp, *bConsoleDown, *btactNext, *btactEnd;
-	CBattleConsole *console;
-	CBattleHero *attackingHero, *defendingHero; //fighting heroes
-	CStackQueue *queue;
+	std::shared_ptr<CButton> bOptions;
+	std::shared_ptr<CButton> bSurrender;
```
FeniksFire: these buttons, shouldn't be unique_ptr?

ArseniyShestakov: @FeniksFire why?

Usage of ```unique_ptr``` apply certain requirements on how headers are organized too and ours aren't very good. You know...

FeniksFire: @ArseniyShestakov and why you need shared_ptr? From first point of view something like "button" shouldn't be shared, and even modified. So sole owner for buttons sounds good.

ArseniyShestakov: It does make sense, but I suppose at least for now we slap ```shared``` on everything so we can keep internals of GH working until we actually rewrite the whole thing. After all GH itself not yet use smart pointers I suppose.

alexvins: All widgets must be `shared_ptr`, later child links in CIntObjects to be changed to shared_ptr and parent link to weak_ptr

alexvins: @FeniksFire  What is incorrectly written?

ArseniyShestakov:
Ok so I finally able to get my hands on branch. Regression found:

- If you start game with multiple towns and switch between them inside town screen then when you attempt to leave screen you'll have to click on button multiple times since all screens remain alive in background. This also cause crashes.

dydzio0614:
Little testing revealed 1 bug so far - cursor type does not change in some cases where it should. When interacting with adventure map object to make dialog window appear, the cursor does not change to arrow, but stays the same type as when clicking on object to pick it. Same with starting battle by attacking and sword icon being active until actual battle starts, etc.

Also some crash happened that may be unrelated to the branch. At townscreen crash happened. At 1st I was clicking various buildings, including marketplace and mage guild, then I was moving cursor over empty creature slots. At that moment crash occured, probably when cursor was at border between upper and lower slots

ArseniyShestakov:
Ok I tested it again today and it's seems alright to merge.
Hopefully I didn't break something. :-)

PR #447, opened 2018-04-08
Unittest improvements

karliss:
Added the testdata copying as suggested by @FeniksFire  and made some changes for better ctest integration.
TODO:
- [x] investigate how QT Creator test discovery works
- [x] decide what to do with ctest running tests one by one (slow)

Cmake 3.10 added automatic google test discovery.  It is optional but results in nicer report and allows completing the testing even when some of the test cases crash . Downside is that it runs each test case in separate executable which is quite slow due to SetUp and TearDown taking most of the time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,6 +302,7 @@ if(ENABLE_LAUNCHER)
 	add_subdirectory(launcher)
 endif()
 if(ENABLE_TEST)
+	enable_testing()
 	add_subdirectory(test)
 endif()
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,5 +1,3 @@
-enable_testing()
-
 set(googleTest_Dir ${CMAKE_CURRENT_SOURCE_DIR}/googletest)
 if(EXISTS ${googleTest_Dir})
 	set(GTestSrc ${googleTest_Dir}/googletest)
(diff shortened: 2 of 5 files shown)


alexvins:
@karliss if you goal is to run tests by CI, then you should start from extracting tests, not depending on Content into separate executable.

karliss:
@alexvins my primary goal is getting the tests working as nicely as possible for local running. For running in CI splitting in two executables is one possible approach. It is also worth looking in to test filtering/tagging functionality provided by gtest and fixing catastrophic failures so that test suite can be completed without restarting it or running each test case in separate executable.

karliss:
Had some progress investigating why qtCreator sometimes finds the tests only in CMemoryBufferTest.cpp.  One of the checks weather source file could contain unit tests is if it includes gtest/gtest.h. Sometimes it decides that test files include StdInc.h from client folder instead of StdInc.h from test folder. Inspecting unopened test files in "Tools/C++ Code Model Inspector" show the same result.

ArseniyShestakov:
I don't think Code Model actually able to handle our PCH as of now and I didn't managed to get it work in any version of Qt Creator. Even if I increase scanned file limit too 200MB since our precompiled headers are usually more than 100MB in size they seems to be completely ignored.

karliss:
Header size doesn't seem to be the problem this time. QtCreator has header lookup cache at CppSourceProcessor::resolveFile which gets reused between all the targets during initial source indexing after opening a project. It maps the headers as used in source to their absolute paths. Before doing proper lookup based on current target's include paths it checks the cache. Since the cache ignores differences in header paths it causes using wrong header in case of multiple headers with the same name. Disabling this cache resulted in correct header lookup and all the tests to be found. The precise conditions for repeating it are a bit more complicated than I described.

Now it's clear that this isn't caused by vcmi doing something bad in Cmake configuration.

It is possible to workaround this problem by including gtest/gtest.h directly in each test or using test/StdInc.h instead of StdInc.h but I don't think it's worth it. Both would be easy to do incorrectly as using wrong from wouldn't cause a major problem and I guess the second might cause problems with precompiled headers during compilation.

I will make a bug report for QtCreator once I find the time to do it properly.

karliss:
In case someone is curios here is the QtCreator bug https://bugreports.qt.io/browse/QTCREATORBUG-20276 . 

In the end I did less than I expected, but at least it the process of running it should be slightly smoother.  This pull request is now up for review.

PR #449, opened 2018-04-13
Exposed existing limiters for modders

godric3:
Exposed limiters:
-  `CREATURE_NATIVE_TERRAIN` - predefined, works if creature is on its native terrain
- `CREATURE_FACTION_LIMITER` - customizable, creature is from specified faction
- `CREATURE_ALIGNMENT_LIMITER` - customizable, creature is from specified alignment
- `CREATURE_TERRAIN_LIMITER`  - customizable, creature is on specified terrain, if not specified then works like `CREATURE_NATIVE_TERRAIN`

### DIFF
--- a/lib/CStack.cpp
+++ b/lib/CStack.cpp
@@ -282,6 +282,13 @@ bool CStack::canBeHealed() const
 		   && !hasBonusOfType(Bonus::SIEGE_WEAPON);
 }
 
+bool CStack::isOnNativeTerrain() const
+{
+	if(base)
+		return type->isItNativeTerrain(base->armyObj->battle->terrainType);
+	return false;
+}
+
 const CCreature * CStack::unitType() const
 {
 	return type;
--- a/lib/CStack.cpp
+++ b/lib/CStack.cpp
@@ -284,9 +284,12 @@ bool CStack::canBeHealed() const
 
 bool CStack::isOnNativeTerrain() const
 {
-	if(base)
-		return type->isItNativeTerrain(base->armyObj->battle->terrainType);
-	return false;
+	return type->isItNativeTerrain(battle->getTerrainType());
+}
+
+bool CStack::isOnTerrain(int terrain) const
+{
+	return battle->getTerrainType() == terrain;
 }
 
 const CCreature * CStack::unitType() const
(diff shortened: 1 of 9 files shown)


ArseniyShestakov:
BTW might be I overestimate the problem and since you use string names already it's shouldn't make that much difference. After all we can just patch mods with prefixes once terrain modding is actually there.

So if @alexvins have different opinion I'll revoke my request to change it within this PR.

ArseniyShestakov:
It's could be nice to add some to modding section of ```ChangeLog```.

Also I suppose we really should tell it to all contributors, but since you already made few PRs before it's make even more sense: add yourself to ```AUTHORS```. Thanks!

godric3:
If I get you right, you want terrains be accessible just like, for example, resources? 
So registering them https://github.com/vcmi/vcmi/blob/develop/lib/CModHandler.cpp#L619
And then requesting identifiers like with factions, creatures, etc.?

ArseniyShestakov:
Exactly. It's unlikely that we'll allow to add alignments anytime soon since gameplay-wise it's make little sense, but terrains it's a feature that many modders asked about for years and sooner or later we'll have it.

So better to do it properly from very beginning.

lib/HeroBonus.cpp
```diff
@@ -1721,24 +1724,49 @@ bool CPropagatorNodeType::shouldBeAttached(CBonusSystemNode *dest)
...
+		return stack->base->armyObj->battle->terrainType != terrainType;
+	}
+	return true;
 	//TODO neutral creatues
 }
 
+std::string CreatureTerrainLimiter::toString() const
+{
+	char buf[100];
+	sprintf(buf, "CreatureTerrainLimiter(terrainType=%s)",
```
alexvins: If you really want to use `printf` instead of boost::format, use more safer `snprintf` and make 100 a named constant

lib/CStack.cpp
```diff
@@ -282,6 +282,13 @@ bool CStack::canBeHealed() const
 		   && !hasBonusOfType(Bonus::SIEGE_WEAPON);
 }
 
+bool CStack::isOnNativeTerrain() const
+{
+	if(base)
+		return type->isItNativeTerrain(base->armyObj->battle->terrainType);
```
alexvins: There is `this->battle` for that. Also why stack without `base` (f.e. summoned) can not have native terrain?

PR #450, opened 2018-04-22
Add server GAMEPLAY_STARTING state.

karliss:
Fixes [2962](https://bugs.vcmi.eu/view.php?id=2962) .

### DIFF
--- a/server/CVCMIServer.h
+++ b/server/CVCMIServer.h
@@ -32,6 +32,7 @@ class CBaseForGHApply;
 enum class EServerState : ui8
 {
 	LOBBY,
+	GAMEPLAY_STARTING,
 	GAMEPLAY,
 	GAMEPLAY_ENDED,
 	SHUTDOWN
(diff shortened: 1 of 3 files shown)


server/CVCMIServer.h
```diff
@@ -9,6 +9,8 @@
  */
 #pragma once
 
+#include "Global.h"
```
ArseniyShestakov: This should not be required I suppose.

server/NetPacksLobbyServer.cpp
```diff
@@ -115,6 +115,9 @@ bool LobbyChatMessage::checkClientPermissions(CVCMIServer * srv) const
 
 bool LobbySetMap::applyOnServer(CVCMIServer * srv)
 {
+	if (srv->state != EServerState::LOBBY) {
+		return false;
+	}
```
ArseniyShestakov: Should be formatted according to our guidelines.

server/CVCMIServer.h
karliss: Removed it, lets fix it globally.

server/NetPacksLobbyServer.cpp
karliss: Done

ArseniyShestakov:
Ok I cherry-picked it again so not to waste your time:
https://github.com/vcmi/vcmi/commit/965b33d16a5e98d334e401e558fde7be9d56acf9

PR #452, opened 2018-04-30
Fix memory and other problems found by address and UB sanitizers.

karliss:
There are slight changes in experience amounts  after 151th level most likely due to double having less significant digits than ui64.

Among other things should fix https://bugs.vcmi.eu/view.php?id=2954.


FeniksFire:
For me it will be great if you can use in future `uint32_t` instead  `ui32` etc. Typedefs for these types in new code is no longer needed.

karliss:
@FeniksFire Sure, in that case would be good to  add a comment in header declaring them or style wiki indicating what utilities are deprecated.

PR #454, opened 2018-06-05
[WIP] AI rework

nullkiller:


nullkiller:
The idea of rework is to make code more simple and reduce duplication.
First of all I decided to split abstract and elementar goals onto two separate things: goals and tasks.
Goals can only analyze and generate tasks and should not modify gamestate.
Tasks should perform actual actions like moving to tile, buy army or something.
I left Win as main goal and just ported its logic.
Conquer now is responsible for general AI activity in order to increase its own power and destroy enemy.
Build is a subgoal of Conquer, its logic was rewriten to try enforce some balance between army development and gold production.
I use more strict prioritization, in general goal should not assign task with priority grater than 1 or lover than 0. Typically it is values 0, 0.1, 0.3, 0.8. When tasks are forwarded to parent goal their priority is modified according to goal's own priority.
FuzzyLite is mostly eliminated from AI logic. I am not going to use it for now but maybe I will decide to use it later. Sometimes it might be good to prioritize using more than one parameter (now usually it is only distance to object).
CaptureObjects was added as a main instrument to visit objects.
A lot of code was removed and a lot have to be removed still. Also some refactoring is still required.

AI/VCAI/Goals.cpp
```diff
@@ -1117,11 +1149,39 @@ TSubgoal GatherTroops::whatToDoToAchieve()
...
+
+		addTasks(heroTasks, sptr(CaptureObjects().ofType(Obj::TOWN).sethero(heroPtr)), 1);
+		addTasks(heroTasks, sptr(CaptureObjects().ofType(Obj::HERO).sethero(heroPtr)), 0.95);
+		addTasks(heroTasks, sptr(CaptureObjects().ofType(Obj::MINE).sethero(heroPtr)), 0.7);
+		addTasks(heroTasks, sptr(CaptureObjects().sethero(HeroPtr(heroPtr))), 0.5);
+
+		auto strongestHero = vstd::maxElementByFun(heroes, [](const CGHeroInstance* h) -> bool { return h->getArmyStrength(); });
+
+		if (cb->getDate(Date::MONTH) > 1 && nextHero.get() == strongestHero[0]) {
+			addTasks(heroTasks, sptr(Explore().sethero(HeroPtr(heroPtr))), 0.7);
```
DjWarmonger: What has month to do with exploration?

AI/VCAI/AIUtility.cpp
```diff
@@ -375,33 +375,6 @@ int3 whereToExplore(HeroPtr h)
...
-						nearbyVisitableObjs.push_back(obj);
-				}
-			}
-		}
-	}
-	vstd::removeDuplicates(nearbyVisitableObjs); //one object may occupy multiple tiles
-	boost::sort(nearbyVisitableObjs, CDistanceSorter(h.get()));
-	if(nearbyVisitableObjs.size())
-		return nearbyVisitableObjs.back()->visitablePos();
-
```
DjWarmonger: What happened to this code?

AI/VCAI/Goals/Build.cpp
```diff
@@ -304,13 +310,48 @@ Tasks::TaskList Build::getTasks()
...
+
+		std::vector<const CGObjectInstance*> interestingObjects;
 
+		vstd::copy_if(ai->visitableObjs, std::back_inserter(interestingObjects), [&](const CGObjectInstance* obj) -> bool {
+			auto id = obj->ID;
+			auto mineOrResource = id == Obj::MINE || id == Obj::ABANDONED_MINE || id == Obj::RESOURCE;
+
+			return mineOrResource && obj->subID == resType;
+		});
+
```
DjWarmonger: I believe there should still be intermediate step "gather resources" between "build" and "capture mine". Resources can also be generated via trading, and this functionality is already handled.

AI/VCAI/Goals/Build.cpp
```diff
@@ -327,7 +327,7 @@ Tasks::TaskList Build::getTasks()
 		}
 
 		double daysToGetResource = requiredResources[resType] / (dailyIncome[resType] + 0.001); // avoid zero divide
-		double priority = resType == std::max(1.0, daysToGetResource / 10);
+		double priority = resType == std::min(1.0, daysToGetResource / 10);
```
DjWarmonger: This is incorrect.

If daysToGetResource is 0, we should get max priority (now 1) and realize the task immediately. If daysToGetResource is higher, the task should get lower priority. Now we get highest priority it it takes 10 or more days to collect resources.

AI/VCAI/Goals/Conquer.cpp
```diff
@@ -58,8 +59,14 @@ Tasks::TaskList Conquer::getTasks() {
 	// lets process heroes according their army strength in descending order
 	std::sort(heroes.begin(), heroes.end(), isLevelHigher);
 
-	addTasks(tasks, sptr(CaptureObjects().ofType(Obj::TOWN)), 1);
-	addTasks(tasks, sptr(CaptureObjects().ofType(Obj::HERO)), 0.95);
+	addTasks(tasks, sptr(Defence()), 1);
```
DjWarmonger: Making Defense part of "Conquer" is questionable. In some scenarios AI might be able to win the game immediately without defending or passing turn.

AI/VCAI/Goals/Build.cpp
nullkiller: Exactly. As more days we need to get resource as more we are blocked by it and can't develop town further and need to do something to reduce this blocking. daysToGetResource are calculated as a ratio required/income. 0 means that we do not need it at all.

AI/VCAI/Goals/Conquer.cpp
nullkiller: Win goal should take care of it. Priorities are local to current goal.

AI/VCAI/AIUtility.cpp
nullkiller: It is a part of Conquer now. Explore just explores. It moves only 1 tile so if there is something to get we will receive a task to get it. If it has enough priority we will do it instead of exploring. Kind of single responsibility. Sure it eats more resources especially CPU.

AI/VCAI/VCAI.cpp
```diff
@@ -606,13 +575,17 @@ void VCAI::showBlockingDialog(const std::string & text, const std::vector<Compon
...
 	if(selection) //select from multiple components -> take the last one (they're indexed [1-size])
 		sel = components.size();
 
 	if(!selection && cancel) //yes&no -> always answer yes, we are a brave AI :)
 		sel = 1;
 
+	// TODO: Find better way to understand it is Chest of Treasures
+	if (components.size() == 2 && components.front().id == Component::RESOURCE) {
+		sel = 1; // for now lets pick gold from a chest.
+	}
```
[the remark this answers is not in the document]
nullkiller: By default it selects the last option. I just overrode this behavior for a particular query type.

AI/VCAI/AIUtility.cpp
```diff
@@ -333,24 +333,13 @@ bool compareDanger(const CGObjectInstance * lhs, const CGObjectInstance * rhs)
...
-		{
-			return false;
-		}
-	}
+	si64 heroStrength = h->getTotalStrength();
+	si64 dangerStrength = evaluateDanger(tile, *h);
+	si64 dif = dangerStrength - heroStrength / SAFE_ATTACK_CONSTANT;
 
-	return true; //there's no danger
+	return std::max((si64)0, dif);
```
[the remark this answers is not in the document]
nullkiller: Should I use ?: std:max is much slower?

AI/VCAI/Goals/Build.cpp
DjWarmonger: But pro players are NOT blocked by town development. Developing towns over many days is suboptimal strategy to begin with.

By making AI fixed on the goals which are hardest to achieve, you will never make it progress.

AI/VCAI/AIUtility.cpp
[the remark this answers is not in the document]
ArseniyShestakov: You just don't make function call where simple ```>``` is sufficient.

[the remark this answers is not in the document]
nullkiller: Why std:max exists than?  Can't compiler inline its code?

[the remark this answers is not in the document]
DjWarmonger: @nullkiller This is not assembler. You can write clear code explicitely ;)
@alexvins The more we should stay within unsigned numbers.

PR #455, opened 2018-06-21
AI - improve exploration, cancel deadends

nullkiller:
Fixed revealed tiles estimation. Diagonal move opens more tiles but spends more movement points.
Improved dead-end cancellation. Now it understands that it is in dead end faster.

### DIFF
--- a/AI/VCAI/AIUtility.cpp
+++ b/AI/VCAI/AIUtility.cpp
@@ -430,12 +430,12 @@ bool isBlockVisitObj(const int3 & pos)
 	return false;
 }
 
-bool hasReachableNeighbor(const int3 &pos, const CPathsInfo* paths, CCallback * cbp)
+bool hasReachableNeighbor(const int3 &pos, HeroPtr hero, CCallback * cbp)
 {
-	for (crint3 dir : int3::getDirs())
+	for(crint3 dir : int3::getDirs())
 	{
 		int3 tile = pos + dir;
-		if (cbp->isInTheMap(tile) && paths->getPathInfo(tile)->reachable())
+		if(cbp->isInTheMap(tile) && cbp->getPathsInfo(hero.get())->getPathInfo(tile)->reachable())
 		{
 			return true;
 		}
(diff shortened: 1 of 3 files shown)


AI/VCAI/AIUtility.cpp
```diff
@@ -431,18 +430,31 @@ bool isBlockVisitObj(const int3 & pos)
 	return false;
 }
 
-int howManyTilesWillBeDiscovered(const int3 & pos, int radious, CCallback * cbp)
+bool hasReachableNeighbor(const int3 &pos, const CPathsInfo* paths, CCallback * cbp)
+{
+	for (crint3 dir : int3::getDirs())
```
alexvins: No space after `for`, `if` please (and around `(` `)` in general )

AI/VCAI/VCAI.cpp
```diff
@@ -2668,6 +2667,7 @@ int3 VCAI::explorationNewPoint(HeroPtr h)
 	int radius = h->getSightRadius();
 	CCallback * cbp = cb.get();
 	const CGHeroInstance * hero = h.get();
+	const CPathsInfo * pathsInfo = cbp->getPathsInfo(hero);
```
alexvins: You should put `getPathsInfo` even upper on call stack, where `h` is set.

AI/VCAI/AIUtility.cpp
```diff
@@ -431,18 +430,31 @@ bool isBlockVisitObj(const int3 & pos)
 	return false;
 }
 
-int howManyTilesWillBeDiscovered(const int3 & pos, int radious, CCallback * cbp)
+bool hasReachableNeighbor(const int3 &pos, const CPathsInfo* paths, CCallback * cbp)
```
alexvins: I do not like mutable CCallback  pointer here.

AI/VCAI/VCAI.cpp
alexvins: We should not cache it - it invalidates very often (almost on any gamestate change). So you should put it to upper most function but do not store between AI API calls. (IOW put it so it transferred in a function parameter but do not stored in class field)

AI/VCAI/AIUtility.cpp
nullkiller: Sorry, now I call non-const method CCallback::getPathsInfo(hero)

AI/VCAI/VCAI.cpp
```diff
@@ -2625,22 +2625,20 @@ void VCAI::buildArmyIn(const CGTownInstance * t)
...
 			{
-				if(isSafeToVisit(h, tile) && isAccessibleForHero(tile, h))
-				{
-					if(isBlockVisitObj(tile))
-						continue;
-					else
-						dstToRevealedTiles[tile] = howManyTilesWillBeDiscovered(radius, hpos, dir);
-				}
+				auto distance = hpos.dist2d(tile); // diagonal movement opens more tiles but spends more mp
+				dstToRevealedTiles[tile] = howManyTilesWillBeDiscovered(tile, radius, cb.get(), h) / distance;
```
dydzio0614: Maybe it would make sense to change dstToRevealedTiles to contain "double" value now?

dydzio0614:
Tested this branch a little bit and nothing wrong found in game stability and AI behavior.

AI/VCAI/VCAI.cpp
nullkiller: Yes, I expect it was double at some point of time.

nullkiller: I also considered to get real cost in movement points instead of just dist2d but decided that it is some implicit logic which will force AI exploring along the roads first because it is cheaper. Not sure it is always desired behavior so decided to leave it with dist2d

DjWarmonger:
Probably cleaner solution than mine. Too bad I can't properly test it atm since RMG is broken.

PR #457, opened 2018-06-25
Map-of-proposed-towns

GuardianRM:
This file contains a fairly complete map of the cities created or proposed, collected based on data and interpretations from various resources. It can serve as an auxiliary guide.


ArseniyShestakov:
Nice compilation, but what does it have to do with our code? :-)

Post it on forum somewhere:
https://forum.vcmi.eu/

alexvins:
Empty city idea is worth a try, but again not here. It could sepearate repository in https://github.com/vcmi-mods organization

GuardianRM:
Ok, think it can be closed

ArseniyShestakov:
Just feel free to post on forum if you wish to make some town overview for VCMI :-)

GuardianRM:
I think there is no point in that. The idea is that, in principle, on any platform, there is not enough "all in one" version. 9-10 cities are not that many. Heroes 3 could be, for example, with 30 or so cities - as far as I understand usually the intuition, suffices for 1 maximum of 2 cities. The idea is to fix everything in one release. In principle, it can be done on a standard template without animation units changing only their card type images and parameters and without new buildings on the game map for additional cities, then it will not take too much time. It could also be a start for the "finished version", but it looks like an "idea of ​​fix", since it is quite long-term and difficult to animate units and buildings for heroes without sufficient artistic skills, but it would be interesting to look at queuers 3 with 30 cities in principle and without animation.

PR #461, opened 2018-07-18
Fix counterattack wrong mechanics (Re-fix bug 2501)

dydzio0614:


server/CGameHandler.cpp
```diff
@@ -4074,10 +4074,10 @@ bool CGameHandler::makeBattleAction(BattleAction &ba)
 				//we check retaliation twice, so if it unblocked during attack it will work only on next attack
 				if(stack->alive()
 					&& !stack->hasBonusOfType(Bonus::BLOCKS_RETALIATION)
-					&& (i > 0 || !firstStrike)
+					&& (i == 0 && !firstStrike)
```
alexvins: Why this changed?

dydzio0614: Counterattack should happen only after 1st attack

PR #463, opened 2018-07-18
Fix bone/ghost dragon handling in skeleton transformer

dydzio0614:
Fixes mantis bug 2932

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -3668,7 +3668,8 @@ bool CGameHandler::transformInUndead(const IMarket *market, const CGHeroInstance
 	//resulting creature - bone dragons or skeletons
 	CreatureID resCreature = CreatureID::SKELETON;
 
-	if (s.hasBonusOfType(Bonus::DRAGON_NATURE)
+	if ((s.hasBonusOfType(Bonus::DRAGON_NATURE)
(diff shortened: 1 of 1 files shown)


server/CGameHandler.cpp
```diff
@@ -3668,7 +3668,9 @@ bool CGameHandler::transformInUndead(const IMarket *market, const CGHeroInstance
 	//resulting creature - bone dragons or skeletons
 	CreatureID resCreature = CreatureID::SKELETON;
 
-	if (s.hasBonusOfType(Bonus::DRAGON_NATURE)
+	if ((s.hasBonusOfType(Bonus::DRAGON_NATURE)
+			&& (s.getCreatureID() != CreatureID::BONE_DRAGON)
+			&& (s.getCreatureID() != CreatureID::GHOST_DRAGON))
```
alexvins: This should be more generic solution f.e. `!hasBonusOfType(Bonus::UNDEAD)`

DjWarmonger: We already have Dracolich for that matter

dydzio0614: I do not see how is it much better - 1 less line of code? Current conditions make it fairly clear why they are here in the first place.

alexvins: Condition should be: any dragon|hydra to bone dragon but undead dragon to skeleton.

PR #464, opened 2018-07-19
Moved some existing logic to new ResourceManager.

DjWarmonger:
Not for merge, just trying new concepts. Possible tasks:

- [x] Define interface for ResourceManager

- [ ] Nicer solution for Capitol reserve

- [ ] Make VCAI request gathering resources when it can buy noting

- [ ] Make VCAI set up reserve in Resource Manager

- [ ] Make VCAI release reserve when stuff is bough / constructed

- [x] Make Resource Manager suggest Goals (CollectRes)

- [ ] Serialization for ResourceManager

- [x] Unit tests


AI/VCAI/VCAI.h
```diff
@@ -406,7 +402,6 @@ class VCAI : public CAdventureAI
 		h & visitableObjs;
 		h & alreadyVisited;
 		h & reservedObjs;
-		h & saving;
```
alexvins: do not forget to put version check/backward compatibility here, this should be done from very beginning

DjWarmonger: As far as I see, old properties can be dropped with no harm. Either way, not sure how to handle it.

DjWarmonger: Just realized I have no old savegame to test it, anyway.

alexvins: Nothing could be changed in serialization without harm.

DjWarmonger:
Nvm, got it.

AI/VCAI/AIUtility.h
```diff
@@ -35,7 +35,7 @@ extern const int GOLD_RESERVE;
 //provisional class for AI to store a reference to an owned hero object
 //checks if it's valid on access, should be used in place of const CGHeroInstance*
 
-struct HeroPtr
+struct DLL_EXPORT HeroPtr
```
alexvins: All this DLL_EXPORT will not work, because you need it to be DLL_IMPORT inside tests. You need other macro, that work same way as DLL_LINKAGE for VCMILib

alexvins: You`re lucky :P DLL_LINKAGE macro is there on purpose ;-)

DjWarmonger: DLL_LINKAGE didn't work, since in VCAI it's set to import

alexvins: Please, read my message again "You need other macro, that work same way ..."

AI/VCAI/ResourceManager.cpp
```diff
@@ -0,0 +1,304 @@
...
+{
+	ai = AI;
+}
+
+bool ResourceManager::canAfford(const TResources & cost) const
+{
+	return freeResources().canAfford(cost);
+}
+
+TResources ResourceManager::estimateIncome() const
```
ArseniyShestakov: I do get this might be beyond what you want to touch during this refactoring, but it's could be great to create unified income calculation code within libvcmi since current one is a mess: code that handle thieves guild is separate and one that handle actual income is separate as well.

DjWarmonger:
What I've got right now is pretty nice and stable. One big thing I'd like to add is to make AI collect resources for Gather Army and buy army only on request and not just at the end of turn. Need to think it through, it's a lot of work.

DjWarmonger:
Hmmhmm, rebase fail again?

alexvins:
You`ve pushed all this shit to upstream

alexvins:
I have to rewrite upstream to fix this.

PR #465, opened 2018-07-20
Map object rating for VCAI

dydzio0614:
Initial version of this feature is finished.

Initial prototype AI list of objects and their values is done. It uses object config files to grab values from, current version uses RMG values, tweaked for some objects. Added first usage of this feature to VCAI::wander().

TODO:
- [x] Polish evaluator access design
- [x] Polish data structure
- [x] Use object evaluation in fuzzy logic system
- [x] How to treat objects without value?

OPTIONAL:
- [x] AI object value config
- [ ] Unit tests (not in this PR)

Implemented as singleton for ease of use and no unintuitive "middleman" by getting access via ai-> etc. For now object database has shared state across all AI players.

### DIFF
--- a/AI/VCAI/Fuzzy.cpp
+++ b/AI/VCAI/Fuzzy.cpp
@@ -295,9 +295,9 @@ float FuzzyHelper::getWanderTargetObjectValue(const CGHeroInstance & h, const Ob
 	boost::optional<int> objValueKnownByAI = MapObjectsEvaluator::getInstance().getObjectValue(obj->ID, obj->subID);
 	int objValue = 0;
 
-	if(objValueKnownByAI.is_initialized()) //consider adding value manipulation based on object instances on map
+	if(objValueKnownByAI != boost::none) //consider adding value manipulation based on object instances on map
 	{
-		objValue = std::min(std::max(objValueKnownByAI.value(), 0), 20000);
+		objValue = std::min(std::max(objValueKnownByAI.get(), 0), 20000);
 	}
 	else
 	{
--- a/AI/VCAI/Fuzzy.cpp
+++ b/AI/VCAI/Fuzzy.cpp
@@ -490,7 +490,7 @@ void FuzzyHelper::initWanderTarget()
 
 		wanderTarget.visitGain->addTerm(new fl::Ramp("LOW", 5, 0)); 
 		wanderTarget.visitGain->addTerm(new fl::Triangle("MEDIUM", 4, 6));
-		wanderTarget.visitGain->addTerm(new fl::Ramp("LOW", 5, 10));
+		wanderTarget.visitGain->addTerm(new fl::Ramp("HIGH", 5, 10));
 		wanderTarget.visitGain->setRange(0, 10);
 
 		wanderTarget.addRule("if distance is LONG and objectValue is HIGH then visitGain is MEDIUM");
--- a/AI/VCAI/Fuzzy.cpp
+++ b/AI/VCAI/Fuzzy.cpp
@@ -199,6 +199,20 @@ void FuzzyHelper::initTacticalAdvantage()
 	}
 }
 
+float FuzzyHelper::calculateTurnDistanceInputValue(const CGHeroInstance * h, int3 tile) const
+{
+	float turns = 0.0f;
+	float distance = CPathfinderHelper::getMovementCost(h, tile);
+	if(distance)
+	{
+		if(distance < h->movement) //we can move there within one turn
+			turns = (fl::scalar)distance / h->movement;
+		else
+			turns = 1 + (fl::scalar)(distance - h->movement) / h->maxMovePoints(true); //bool on land?
+	}
+	return turns;
+}
+
 ui64 FuzzyHelper::estimateBankDanger(const CBank * bank)
 {
 	//this one is not fuzzy anymore, just calculate weighted average
@@ -420,20 +434,7 @@ float FuzzyHelper::evaluate(Goals::VisitTile & g)
 		return 0;
 
 	//assert(cb->isInTheMap(g.tile));
-	float turns = 0;
-	float distance = CPathfinderHelper::getMovementCost(g.hero.h, g.tile);
-	if(!distance) //we stand on that tile
-	{
-		turns = 0;
-	}
-	else
-	{
-		if(distance < g.hero->movement) //we can move there within one turn
-			turns = (fl::scalar)distance / g.hero->movement;
-		else
-			turns = 1 + (fl::scalar)(distance - g.hero->movement) / g.hero->maxMovePoints(true); //bool on land?
-	}
-
+	float turns = calculateTurnDistanceInputValue(g.hero.h, g.tile);
 	float missionImportance = 0;
 	if(vstd::contains(ai->lockedHeroes, g.hero))
 		missionImportance = ai->lockedHeroes[g.hero]->priority;
--- a/AI/VCAI/Fuzzy.cpp
+++ b/AI/VCAI/Fuzzy.cpp
@@ -19,6 +19,7 @@
 #include "../../lib/VCMI_Lib.h"
 #include "../../CCallback.h"
 #include "VCAI.h"
+#include "MapObjectsEvaluator.h"
 
 #define MIN_AI_STRENGHT (0.5f) //lower when combat AI gets smarter
 #define UNGUARDED_OBJECT (100.0f) //we consider unguarded objects 100 times weaker than us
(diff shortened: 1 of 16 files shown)


AI/VCAI/MapObjectsEvaluator.cpp
```diff
@@ -0,0 +1,40 @@
...
+				newObject.second.insert(std::pair<int,int>(secondaryID, handler->getRMGInfo().value));
+			}
+		}
+		objectDatabase.insert(newObject);
+	}
+}
+
+boost::optional<int> MapObjectsEvaluator::getObjectValue(int primaryID, int secondaryID)
+{
+	auto object = objectDatabase.find(primaryID);
```
nullkiller: I just wander why it is impossible to use .at here? Not strong at stdlib types

dydzio0614:
Is it properly fixed now? :P

AI/VCAI/MapObjectsEvaluator.h
```diff
@@ -0,0 +1,48 @@
...
+*
+* Authors: listed in file AUTHORS in main folder
+*
+* License: GNU General Public License v2.0 or later
+* Full text of license available in license.txt file, in main folder
+*
+*/
+#pragma once
+
+struct AiMapObjectID
```
alexvins: This struct would be useful globally as part of CObjectClassesHandler

dydzio0614: What to do with it? Move to CObjectClassesHandler.h and what then?

alexvins: new getHandlerFor overload should accept it

lib/mapObjects/CObjectClassesHandler.h
```diff
@@ -65,6 +65,30 @@ struct DLL_LINKAGE RandomMapInfo
 	}
 };
 
+struct DLL_LINKAGE AiMapObjectID
```
alexvins: As long as you moved here, please select different name

lib/mapObjects/CObjectClassesHandler.h
```diff
@@ -65,6 +65,30 @@ struct DLL_LINKAGE RandomMapInfo
...
+		else
+			return this->secondaryID < other.secondaryID;
+	}
+
+	bool operator==(const AiMapObjectID& other)
+	{
+		if(this->primaryID == other.primaryID)
+			return this->secondaryID == other.secondaryID;
+
+		return false;
```
alexvins: should this be just ` return this->primaryID == other.primaryID && this->secondaryID == other.secondaryID`?

lib/mapObjects/CObjectClassesHandler.h
```diff
@@ -65,6 +65,30 @@ struct DLL_LINKAGE RandomMapInfo
...
 };
 
+struct DLL_LINKAGE AiMapObjectID
+{
+	si32 primaryID;
+	si32 secondaryID;
+
+	AiMapObjectID(si32 primID, si32 secID) : primaryID(primID), secondaryID(secID) {};
+
+	bool operator<(const AiMapObjectID& other)
```
alexvins: `const` here

lib/mapObjects/CObjectClassesHandler.h
```diff
@@ -65,6 +65,30 @@ struct DLL_LINKAGE RandomMapInfo
...
+	{
+		if(this->primaryID != other.primaryID)
+			return this->primaryID < other.primaryID;
+		else
+			return this->secondaryID < other.secondaryID;
+	}
+
+	bool operator==(const AiMapObjectID& other)
```
alexvins: `const` here

DjWarmonger:
IMO most formulas (except object value itself) shoudl be copied / inheried from FuzzyHelper::initVisitTile() to avoid code duplication.

dydzio0614:
What for example can be copied from FuzzyHelper::initVisitTile() ?

dydzio0614:
Finally travis builds without errors. PR is ready for merge.

AI/VCAI/Fuzzy.cpp
```diff
@@ -199,6 +202,20 @@ void FuzzyHelper::initTacticalAdvantage()
 	}
 }
 
+float FuzzyHelper::calculateTurnDistanceInputValue(const CGHeroInstance * h, int3 tile) const
+{
+	float turns = 0.0f;
+	float distance = CPathfinderHelper::getMovementCost(h, tile);
```
nullkiller: Please fix it, it is wrong. The only correct way I know is
```
ui32 distanceToTile(const CGHeroInstance* hero, int3 pos) {
	auto pathsInfo = cb->getPathsInfo(hero);

	return distanceToTile(pathsInfo, pos);
}

ui32 distanceToTile(const CPathsInfo* pathsInfo, int3 pos) {
	auto pathInfo = pathsInfo->getPathInfo(pos);
	auto hero = pathsInfo->hero;
	int result = (int)(pathInfo->turns * hero->maxMovePoints(true) + hero->movement) - (int)pathInfo->moveRemains;

	return std::max(result, 0);
}
```
But maybe std:max should be replaced somehow.

dydzio0614: this is copypaste of thing that was initially in evaluate(VisitTile ...). I have no clue how it works and I would deal with it after merge.

config/objects/moddables.json
```diff
@@ -63,21 +64,22 @@
...
-			"mithril" : { "index" : 7 } // TODO: move to WoG?
+			"wood" :    { "index" : 0, "aiValue" : 1400, "rmg" : { "value" : 1400, "rarity" : 300 }, "templates" : { "res" : { "animation" : "AVTwood0.def" } } },
+			"mercury" : { "index" : 1, "aiValue" : 2000, "rmg" : { "value" : 2000, "rarity" : 300 }, "templates" : { "res" : { "animation" : "AVTmerc0.def" } } },
+			"ore" :     { "index" : 2, "aiValue" : 1400, "rmg" : { "value" : 1400, "rarity" : 300 }, "templates" : { "res" : { "animation" : "AVTore0.def"  } } },
+			"sulfur" :  { "index" : 3, "aiValue" : 2000, "rmg" : { "value" : 2000, "rarity" : 300 }, "templates" : { "res" : { "animation" : "AVTsulf0.def" } } },
+			"crystal" : { "index" : 4, "aiValue" : 2000, "rmg" : { "value" : 2000, "rarity" : 300 }, "templates" : { "res" : { "animation" : "AVTcrys0.def" } } },
+			"gems" :    { "index" : 5, "aiValue" : 2000, "rmg" : { "value" : 2000, "rarity" : 300 }, "templates" : { "res" : { "animation" : "AVTgems0.def" } } },
+			"gold" :    { "index" : 6, "aiValue" : 750, "rmg" : { "value" : 750,  "rarity" : 300 }, "templates" : { "res" : { "animation" : "AVTgold0.def" } } },
```
nullkiller: Why gold is so small?

dydzio0614: well, I did not see your suggestions added here: https://etherpad.net/p/VCMI_ADVENTURE_MAP_AI

Initial values used are random map generator values. Though I  can see how 700 gold has less impact than 6 sulfur for example when building buildings.

nullkiller: Gold supports everything. It is why gold mine has grater priority. It would be great so if your case is handled by ResourceManager so that we gather resource when we see its direct need. Just my opinion so.

AI/VCAI/Fuzzy.cpp
DjWarmonger: Hmm, we can fix both now or wait for Pathfinder rewrite. Either way, there are at least 2 other places in VCAI where Pathfinder doesn't perform as expected.

nullkiller: Any more details about these places?

DjWarmonger: One causes this bug: https://bugs.vcmi.eu/view.php?id=2980

The other is `int3 SectorMap::firstTileToGet()` calling `cb->buildBoat()` in a middle of logic

DjWarmonger: So, conclusion -> wait until pathfinder gets rewrite, covering SectorMap functionality

DjWarmonger: Unless current CPathfinderHelper::getMovementCost() call is gamebreaking?

nullkiller: I think boat logic will be covered also. Pathfinder will tell that boat is needed and payh execution logic in vcai will hire a boat

DjWarmonger: So, anyway, what is wrong with Dydzio's code and why it should be changed?

PR #468, opened 2018-07-27
[WIP] Gui refactoring

Illithidek:


client/widgets/Buttons.cpp
```diff
@@ -145,14 +145,20 @@ void CButton::onButtonClicked()
...
+	auto parent = getParent();
 	std::string prefix = "Parent is";
-	while (parent)
+	while (parent != nullptr)
 	{
+		logAnim->error("crash1");
+		
 		logAnim->trace("%s%s at %dx%d", prefix, typeid(*parent).name(), parent->pos.x, parent->pos.y);
-		parent = parent->parent;
+		parent = getParent()->getParent();
```
alexvins: should be `parent->getParent()`

FeniksFire: Thanks for noticing!

client/gui/view/View.h
```diff
@@ -67,7 +67,10 @@ class View : public IShowActivatable
 public:
 	View(ui16 used=0, Point offset=Point());
 	virtual ~View();
-
+	
+	View *getParent() const;
+	void setParent(View *parent);
```
alexvins: This should be removed, or made protected it you really want to have setter anyway.
You should use `addChild` `removeChild` from outside

FeniksFire: I will. Give me some time, I just want rework all these things, so I will made some things temporary :).

PR #470, opened 2018-07-29
AI: add some pathfinder extensibility and replace SectorMap.

nullkiller:
The goal is refactoring CPathfinder in a way so we can extend or change its rules to be able to find a path through monsters and other objects.
- [x] Extract neighbour nodes finder
- [x] Extract a set of IPathfinderRule
- [x] Add new API to client to allow other rules and modules for pathfinder
- [x] Build a pathfinder for AI in order to replace most of SectorMap functionality.
- [x] Patch Explore, Conquer, CollectRes
- _Support building new boats_ (Will be done in the next PR)
- [x] Patch wander
- _Cleanup VisitTile and VisitObj_ (Need to consider it later, not a simple task)

### DIFF
--- a/AI/VCAI/AIUtility.h
+++ b/AI/VCAI/AIUtility.h
@@ -59,6 +59,10 @@ public:
 	const CGHeroInstance * operator->() const;
 	const CGHeroInstance * operator*() const; //not that consistent with -> but all interfaces use CGHeroInstance*, so it's convenient
 	bool operator==(const HeroPtr & rhs) const;
+	bool operator!=(const HeroPtr & rhs) const
+	{
+		return !(*this == rhs);
+	}
 
 	const CGHeroInstance * get(bool doWeExpectNull = false) const;
 	bool validAndSet() const;
@@ -88,6 +92,7 @@ struct ObjectIdRef
 
 	const CGObjectInstance * operator->() const;
 	operator const CGObjectInstance *() const;
+	operator bool() const;
 
 	ObjectIdRef(ObjectInstanceID _id);
 	ObjectIdRef(const CGObjectInstance * obj);
@@ -166,8 +171,9 @@ bool shouldVisit(HeroPtr h, const CGObjectInstance * obj);
 
 ui64 evaluateDanger(const CGObjectInstance * obj);
 ui64 evaluateDanger(crint3 tile, const CGHeroInstance * visitor);
-bool isSafeToVisit(HeroPtr h, crint3 tile);
 bool isObjectRemovable(const CGObjectInstance * obj); //FIXME FIXME: move logic to object property!
+bool isSafeToVisit(HeroPtr h, uint64_t dangerStrength);
+bool isSafeToVisit(HeroPtr h, crint3 tile);
 
 bool compareMovement(HeroPtr lhs, HeroPtr rhs);
 bool compareHeroStrength(HeroPtr h1, HeroPtr h2);
--- a/AI/VCAI/AIUtility.h
+++ b/AI/VCAI/AIUtility.h
@@ -176,6 +176,7 @@ bool compareArtifacts(const CArtifactInstance * a1, const CArtifactInstance * a2
 ui64 howManyReinforcementsCanBuy(HeroPtr h, const CGTownInstance * t);
 ui64 howManyReinforcementsCanGet(HeroPtr h, const CGTownInstance * t);
 int3 whereToExplore(HeroPtr h);
+uint32_t distanceToTile(const CGHeroInstance * hero, int3 pos);
 
 class CDistanceSorter
 {
--- a/AI/VCAI/AIUtility.h
+++ b/AI/VCAI/AIUtility.h
@@ -179,7 +179,7 @@ bool compareMovement(HeroPtr lhs, HeroPtr rhs);
 bool compareHeroStrength(HeroPtr h1, HeroPtr h2);
 bool compareArmyStrength(const CArmedInstance * a1, const CArmedInstance * a2);
 bool compareArtifacts(const CArtifactInstance * a1, const CArtifactInstance * a2);
-ui64 howManyReinforcementsCanBuy(HeroPtr h, const CGTownInstance * t);
+ui64 howManyReinforcementsCanBuy(HeroPtr h, const CGDwelling * t);
 ui64 howManyReinforcementsCanGet(HeroPtr h, const CGTownInstance * t);
 int3 whereToExplore(HeroPtr h);
 uint32_t distanceToTile(const CGHeroInstance * hero, int3 pos);
--- a/AI/VCAI/AIhelper.h
+++ b/AI/VCAI/AIhelper.h
@@ -16,25 +16,26 @@
 
 #include "ResourceManager.h"
 #include "BuildingManager.h"
+#include "Pathfinding/CPathfindingManager.h"
 
 class ResourceManager;
 class BuildingManager;
 
 
 //indirection interface for various modules
-class DLL_EXPORT AIhelper : public IResourceManager, public IBuildingManager
+class DLL_EXPORT AIhelper : public IResourceManager, public IBuildingManager, public IPathfindingManager
 {
 	friend class VCAI;
 	friend struct SetGlobalState; //mess?
 
 	std::shared_ptr<ResourceManager> resourceManager;
 	std::shared_ptr<BuildingManager> buildingManager;
+	std::shared_ptr<CPathfindingManager> pathfindingManager;
 	//TODO: vector<IAbstractManager>
 public:
 	AIhelper();
 	~AIhelper();
 
-	//from ResourceManager
 	bool canAfford(const TResources & cost) const;
 	TResources reservedResources() const override;
 	TResources freeResources() const override;
@@ -43,21 +44,26 @@ public:
 	TResource allGold() const override;
 
 	Goals::TSubgoal whatToDo(TResources &res, Goals::TSubgoal goal) override;
-	Goals::TSubgoal whatToDo() const override; //peek highest-priority goal
+	Goals::TSubgoal whatToDo() const override;
 	bool containsObjective(Goals::TSubgoal goal) const;
 	bool hasTasksLeft() const override;
+
+	bool getBuildingOptions(const CGTownInstance * t) override;
+	boost::optional<PotentialBuilding> immediateBuilding() const override;
+	boost::optional<PotentialBuilding> expensiveBuilding() const override;
+	boost::optional<BuildingID> canBuildAnyStructure(const CGTownInstance * t, const std::vector<BuildingID> & buildList, unsigned int maxDays = 7) const override;
+
+	Goals::TGoalVec howToVisitTile(HeroPtr hero, int3 tile, bool allowGatherArmy = true) override;
+	Goals::TGoalVec howToVisitObj(HeroPtr hero, ObjectIdRef obj, bool allowGatherArmy = true) override;
+	Goals::TGoalVec howToVisitTile(int3 tile) override;
+	Goals::TGoalVec howToVisitObj(ObjectIdRef obj) override;
+	std::vector<AIPath> getPathsToTile(HeroPtr hero, int3 tile) override;
+	void resetPaths() override;
 
 private:
 	bool notifyGoalCompleted(Goals::TSubgoal goal);
 
 	void setCB(CPlayerSpecificInfoCallback * CB) override;
 	void setAI(VCAI * AI) override;
-
-	//from BuildingManager
-public:
-	bool getBuildingOptions(const CGTownInstance * t) override;
-	boost::optional<PotentialBuilding> immediateBuilding() const override;
-	boost::optional<PotentialBuilding> expensiveBuilding() const override;
-	boost::optional<BuildingID> canBuildAnyStructure(const CGTownInstance * t, const std::vector<BuildingID> & buildList, unsigned int maxDays = 7) const override;
 };
 
--- a/AI/VCAI/AIhelper.h
+++ b/AI/VCAI/AIhelper.h
@@ -16,7 +16,7 @@
 
 #include "ResourceManager.h"
 #include "BuildingManager.h"
-#include "Pathfinding/CPathfindingManager.h"
+#include "Pathfinding/PathfindingManager.h"
 
 class ResourceManager;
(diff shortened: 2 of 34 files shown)


lib/CPathfinder.cpp
```diff
@@ -19,6 +19,86 @@
 #include "CConfigHandler.h"
 #include "../lib/CPlayerState.h"
 
+bool canSeeObj(const CGObjectInstance * obj)
```
nullkiller: This method is a bit awkward. Is not it a responsibility of CGObjectInstance or something around it?

lib/CPathfinder.cpp
```diff
@@ -19,6 +19,86 @@
...
+{
+	/// Pathfinder should ignore placed events
+	return obj != nullptr && obj->ID != Obj::EVENT;
+}
+
+CNeighbourFinder::CNeighbourFinder(CPathfinder * pathfinder)
+	:pathfinder(pathfinder), neighbours(), neighbourTiles(), accessibleNeighbourTiles()
+{
```
nullkiller: Circular dependency to pathfinder is also awkward. But we actually need CGameState and CpathNodeInfo here I think. Will think how to refactor it.

lib/CPathfinder.cpp
```diff
@@ -98,154 +190,149 @@ void CPathfinder::calculatePaths()
...
-		for(auto & neighbour : neighbours)
+		auto neighbourNodes = neighbourFinder->calculateNeighbours();
+		for(auto & neighbour : neighbourNodes)
 		{
-			if(!isPatrolMovementAllowed(neighbour))
-				continue;
-
-			dt = &gs->map->getTile(neighbour);
-			dtObj = dt->topVisitableObj();
-			for(ELayer i = ELayer::LAND; i <= ELayer::AIR; i.advance(1))
```
nullkiller: This loop is inside calculateNeighbours now

lib/CPathfinder.cpp
```diff
@@ -98,154 +190,149 @@ void CPathfinder::calculatePaths()
...
-				/// Check transition without tile accessability rules
-				if(cp->layer != i && !isLayerTransitionPossible(i))
-					continue;
+			if(destination.node->locked)
+				continue;
 
-				dp = out.getNode(neighbour, i);
-				if(dp->locked)
-					continue;
+			if(!isPatrolMovementAllowed(destination.node->coord))
```
nullkiller: All these can be separate IPathfindingRule objects

AI/VCAI/Pathfinding/AIPathfinder.cpp
```diff
@@ -12,30 +12,50 @@
...
+		}
+		else
+		{
+			nodeStorage = std::make_shared<AINodeStorage>(cb->getMapSize());
+			storagePool.push_back(nodeStorage);
+		}
+
+		storageMap[hero] = nodeStorage;
+
 		boost::unique_lock<boost::mutex> pathLock(nodeStorage->getMutex());
```
alexvins: Where is a lock for storagePool?

nullkiller: Good question. I also asked it myself. Do you think having a separate lock for storagePool and map is wise? Or should I use a single lock for both pool and all paths?

alexvins: You should use separate lock if it is possible to release it earlier than nodeStorage one.

nullkiller: Does not look like this is wise. Storage in storageMap should either be already calculated or not present at all

AI/VCAI/AIUtility.h
```diff
@@ -166,16 +171,18 @@ bool shouldVisit(HeroPtr h, const CGObjectInstance * obj);
 
 ui64 evaluateDanger(const CGObjectInstance * obj);
 ui64 evaluateDanger(crint3 tile, const CGHeroInstance * visitor);
-bool isSafeToVisit(HeroPtr h, crint3 tile);
 bool isObjectRemovable(const CGObjectInstance * obj); //FIXME FIXME: move logic to object property!
+bool isSafeToVisit(HeroPtr h, uint64_t dangerStrength);
+bool isSafeToVisit(HeroPtr h, crint3 tile);
```
FeniksFire: think it would be better if it will be HeroPtr member method and then marked as const

AI/VCAI/AIUtility.h
```diff
@@ -166,16 +171,18 @@ bool shouldVisit(HeroPtr h, const CGObjectInstance * obj);
...
 bool isObjectRemovable(const CGObjectInstance * obj); //FIXME FIXME: move logic to object property!
+bool isSafeToVisit(HeroPtr h, uint64_t dangerStrength);
+bool isSafeToVisit(HeroPtr h, crint3 tile);
 
 bool compareMovement(HeroPtr lhs, HeroPtr rhs);
 bool compareHeroStrength(HeroPtr h1, HeroPtr h2);
 bool compareArmyStrength(const CArmedInstance * a1, const CArmedInstance * a2);
 bool compareArtifacts(const CArtifactInstance * a1, const CArtifactInstance * a2);
-ui64 howManyReinforcementsCanBuy(HeroPtr h, const CGTownInstance * t);
+ui64 howManyReinforcementsCanBuy(HeroPtr h, const CGDwelling * t);
```
FeniksFire: same here

AI/VCAI/VCAI.h
```diff
@@ -216,14 +214,14 @@ class DLL_EXPORT VCAI : public CAdventureAI
 	Goals::TSubgoal questToGoal(const QuestInfo & q);
 
 	void recruitHero(const CGTownInstance * t, bool throwing = false);
-	bool isGoodForVisit(const CGObjectInstance * obj, HeroPtr h, SectorMap & sm);
+	bool isGoodForVisit(const CGObjectInstance * obj, HeroPtr h, boost::optional<uint32_t> movementCostLimit = boost::none);
+	bool isGoodForVisit(const CGObjectInstance * obj, HeroPtr h, AIPath & path);
```
FeniksFire: mark this as const

FeniksFire: also `isGoodForVisit` need change `VCAI::isTileNotReserved `to be const, to make it you need f.e make there 2 addidionals getters so it will be look like that:
```
bool VCAI::isTileNotReserved(const CGHeroInstance * h, int3 t) const
{
...
		if(obj && vstd::contains(ai->getReservedObjs(), obj) && !vstd::contains(ai->getReservedHeroesMap().at(h), obj))
...
}
```

AI/VCAI/AIhelper.h
```diff
@@ -16,19 +16,21 @@
...
 #include "BuildingManager.h"
+#include "Pathfinding/CPathfindingManager.h"
 
 class ResourceManager;
 class BuildingManager;
 
 
 //indirection interface for various modules
-class DLL_EXPORT AIhelper : public IResourceManager, public IBuildingManager
+class DLL_EXPORT AIhelper : public IResourceManager, public IBuildingManager, public IPathfindingManager
```
FeniksFire: What is point of this inheritance? If you will delete it, it will be work as well.

AI/VCAI/Pathfinding/AINodeStorage.h
```diff
@@ -0,0 +1,97 @@
...
+	bool hasBetterChain(CPathNodeInfo & source, CDestinationNodeInfo & destination);
+	AIPathNode * getNode(const int3 & coord, const EPathfindingLayer layer, int chainNumber);
+	std::vector<AIPath> getChainInfo(int3 pos);
+
+	void setHero(HeroPtr heroPtr)
+	{
+		hero = heroPtr.get();
+	}
+
+	const CGHeroInstance * getHero()
```
FeniksFire: change it to const

AI/VCAI/Pathfinding/AINodeStorage.h
```diff
@@ -0,0 +1,97 @@
...
+};
+
+struct AIPath
+{
+	std::vector<AIPathNodeInfo> nodes;
+
+	AIPath();
+
+	/// Gets danger of path excluding danger of visiting the target object like creature bank
+	uint64_t getPathDanger();
```
FeniksFire: change to be const

AI/VCAI/Pathfinding/AINodeStorage.h
```diff
@@ -0,0 +1,97 @@
...
+
+	/// Gets danger of path including danger of visiting the target object like creature bank
+	uint64_t getTotalDanger(HeroPtr hero);
+
+	int3 firstTileToGet();
```
FeniksFire: change to be const

AI/VCAI/Pathfinding/AINodeStorage.h
```diff
@@ -0,0 +1,97 @@
...
+{
+	std::vector<AIPathNodeInfo> nodes;
+
+	AIPath();
+
+	/// Gets danger of path excluding danger of visiting the target object like creature bank
+	uint64_t getPathDanger();
+
+	/// Gets danger of path including danger of visiting the target object like creature bank
+	uint64_t getTotalDanger(HeroPtr hero);
```
FeniksFire: change to be const

AI/VCAI/Pathfinding/AINodeStorage.h
```diff
@@ -0,0 +1,97 @@
...
+
+	int3 firstTileToGet();
+
+	uint32_t movementCost();
```
FeniksFire: change to be const

AI/VCAI/AIhelper.h
```diff
@@ -59,5 +61,14 @@ class DLL_EXPORT AIhelper : public IResourceManager, public IBuildingManager
 	boost::optional<PotentialBuilding> immediateBuilding() const override;
 	boost::optional<PotentialBuilding> expensiveBuilding() const override;
 	boost::optional<BuildingID> canBuildAnyStructure(const CGTownInstance * t, const std::vector<BuildingID> & buildList, unsigned int maxDays = 7) const override;
+
+	//from PathfindingManager
```
FeniksFire: Not very useful comment, try get rid of some of them.

AI/VCAI/Pathfinding/AINodeStorage.h
```diff
@@ -0,0 +1,97 @@
...
+};
+
+class AINodeStorage : public INodeStorage
+{
+private:
+	mutable boost::mutex pathMx;
+	int3 sizes;
+	boost::multi_array<AIPathNode, 5> nodes; //[w][h][level][layer][chains]
```
FeniksFire: Here, that comment is also not too usefull, I even can't understand what means `[w][h]`?

AI/VCAI/AIUtility.h
nullkiller: As far as I understood HeroPtr is kind of smart pointer? If we will start adding some methods there we will mix it a bit. Especially when calling CGHeroInstance we will use -> and when calling isSafeToVisit it will be just a dot. It looks weird. It might be a responsibility of some BattleManager or something. Do you want to get rid of AIUtility or do you want to have some AIHeroInstance with additional behavior and functionality?

AI/VCAI/AIhelper.h
nullkiller: Not mine. //indirection interface for various modules
I just did the same way as it was already done

AI/VCAI/AIUtility.h
FeniksFire: huh, I just don't like these globals there, any way to get rid of them?

AI/VCAI/AIUtility.h
FeniksFire: Better that than beign global.

AI/VCAI/AIhelper.h
FeniksFire: huh ok, can stay for now, let's hope that someone will change it in seperate PR

AI/VCAI/Pathfinding/CPathfindingManager.h
```diff
@@ -0,0 +1,64 @@
...
+
+public:
+	CPathfindingManager() = default;
+	CPathfindingManager(CPlayerSpecificInfoCallback * CB, VCAI * AI = nullptr); //for tests only
+
+	Goals::TGoalVec howToVisitTile(HeroPtr hero, int3 tile, bool allowGatherArmy = true) override;
+	Goals::TGoalVec howToVisitObj(HeroPtr hero, ObjectIdRef obj, bool allowGatherArmy = true) override;
+	Goals::TGoalVec howToVisitTile(int3 tile) override;
+	Goals::TGoalVec howToVisitObj(ObjectIdRef obj) override;
+	std::vector<AIPath> getPathsToTile(HeroPtr hero, int3 tile) override;
```
FeniksFire: same here

AI/VCAI/Pathfinding/CPathfindingManager.h
```diff
@@ -0,0 +1,64 @@
...
+	void resetPaths() override;
+
+private:
+	void setCB(CPlayerSpecificInfoCallback * CB) override;
+	void setAI(VCAI * AI) override;
+
+	Goals::TGoalVec findPath(
```
FeniksFire: also const

AI/VCAI/Pathfinding/CPathfindingManager.h
```diff
@@ -0,0 +1,64 @@
...
+	virtual ~IPathfindingManager() = default;
+	virtual void setCB(CPlayerSpecificInfoCallback * CB) = 0;
+	virtual void setAI(VCAI * AI) = 0;
+
+	virtual void resetPaths() = 0;
+	virtual Goals::TGoalVec howToVisitTile(HeroPtr hero, int3 tile, bool allowGatherArmy = true) = 0;
+	virtual Goals::TGoalVec howToVisitObj(HeroPtr hero, ObjectIdRef obj, bool allowGatherArmy = true) = 0;
+	virtual Goals::TGoalVec howToVisitTile(int3 tile) = 0;
+	virtual Goals::TGoalVec howToVisitObj(ObjectIdRef obj) = 0;
+	virtual std::vector<AIPath> getPathsToTile(HeroPtr hero, int3 tile) = 0;
```
FeniksFire: const?

PR #471, opened 2018-08-03
Building manager

DjWarmonger:
- Moved build-related logic to new class BuildingManager
- Separated logic from actual build action


DjWarmonger:
I did not change any logic, it's pure refactoring.

dydzio0614:
Ok. Also make sure you do not lose handling of buildings with non-standard ID, for example 3rd creature upgrades.

PR #472, opened 2018-08-04
Yog is not able to buy a spellbook in his campaign anymore, as intended

yupinov:
Reported at https://bugs.vcmi.eu/view.php?id=2652 and partially at https://bugs.vcmi.eu/view.php?id=1635

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -37,6 +37,8 @@
 #include "../../lib/spells/CSpellHandler.h"
 #include "../../lib/CTownHandler.h"
 #include "../../lib/GameConstants.h"
+#include "../../lib/StartInfo.h"
+#include "../../lib/mapping/CCampaignHandler.h"
 #include "../../lib/mapObjects/CGHeroInstance.h"
 #include "../../lib/mapObjects/CGTownInstance.h"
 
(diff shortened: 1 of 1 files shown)


Alucard648:
I think, this campaign/map specific restriction should not be hardcoded into engine. It should be allowed in map editor to disable acquiring spellbooks for specific heroes.

yupinov:
I assume it was hard-coded and not a part of h3c/h3m originally, given that the hero's name is in the message. The intent was only to reproduce this behavior :-) It also doesn't prevent anyone from implementing the better thing.

DjWarmonger:
I agree with @Alucard648  that this code violates all OOP rules, however can't think of a better way to implement that one-time feature.

alexvins:
We have huge file `config\mapOverrides.json` for other one-time campaign features, but they are almost all win-loss conditions.

client/windows/CCastleInterface.cpp
```diff
@@ -874,11 +876,19 @@ void CCastleBuildings::enterMagesGuild()
 
 	if(hero && !hero->hasSpellbook()) //hero doesn't have spellbok
 	{
+		const StartInfo *si = LOCPLINT->cb->getStartInfo();
 		if(LOCPLINT->cb->getResourceAmount(Res::GOLD) < 500) //not enough gold to buy spellbook
 		{
 			openMagesGuild();
 			LOCPLINT->showInfoDialog(CGI->generaltexth->allTexts[213]);
 		}
```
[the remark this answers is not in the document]
yupinov: Ah, well spotted! Done.

yupinov:
Left a comment about mapOverrides.json.

client/windows/CCastleInterface.cpp
```diff
@@ -874,7 +876,16 @@ void CCastleBuildings::enterMagesGuild()
 
 	if(hero && !hero->hasSpellbook()) //hero doesn't have spellbok
 	{
-		if(LOCPLINT->cb->getResourceAmount(Res::GOLD) < 500) //not enough gold to buy spellbook
+		const StartInfo *si = LOCPLINT->cb->getStartInfo();
+		// it would be nice to find a way to move this hack to config/mapOverrides.json
+		if (si && si->campState && si->campState->camp &&               // We're in campaign,
+			(si->campState->camp->header.filename == "DATA/YOG.H3C") && // which is "Birth of a Barbarian",
+			(hero->subID == 45) && (hero->name == "Yog"))               // and the hero is Yog (based on Solmyr)
```
alexvins: You should not check hero name, it may be localized

yupinov: Riiight. I was only unsure whether the player can buy an actual Solmyr, but now I remember heroes can't be duplicate.

yupinov:
Alright, so can this be merged meanwhile? If I find time/energy to work at mapOverrides, then I will first start with Gem in a separate branch, and if that succeeds, try to apply it to Yog as well.

alexvins:
Merged, thank you for contribution.

PR #474, opened 2018-08-04
FuzzyHelper refactoring + Ai Map Object Evaluation improvements

dydzio0614:
- [x] Split FuzzyHelper logic into extendable goal engine classes
- [x] Use smart pointers for FuzzyLite input variables (ABANDONED - should not be done)
- [x] Create elementar version of GetObj goal
- [x] Decide how to handle creature dwelling AI values and implement solution
- [x] Split Fuzzy.h/cpp files


AI/VCAI/Fuzzy.h
```diff
@@ -27,42 +27,52 @@ class engineBase
...
+};
+
+class EvalVisitTile : public HeroMovementGoalEngine
+{
+public:
+	~EvalVisitTile();
+};
+
+class EvalWanderTargetObject : public HeroMovementGoalEngine //designed for use with VCAI::wander()
+{
```
DjWarmonger: That's right!

AI/VCAI/Fuzzy.cpp
```diff
@@ -373,142 +373,93 @@ float FuzzyHelper::evaluate(Goals::RecruitHero & g)
...
+		heroStrength->addTerm(new fl::Ramp("LOW", 0.2, 0));
+		heroStrength->addTerm(new fl::Triangle("MEDIUM", 0.2, 0.8));
+		heroStrength->addTerm(new fl::Ramp("HIGH", 0.5, 1));
+		heroStrength->setRange(0.0, 1.0);
 
-		vt.turnDistance->addTerm(new fl::Ramp("SMALL", 0.5, 0));
-		vt.turnDistance->addTerm(new fl::Triangle("MEDIUM", 0.1, 0.8));
-		vt.turnDistance->addTerm(new fl::Ramp("LONG", 0.5, 3));
-		vt.turnDistance->setRange(0.0, 3.0);
+		turnDistance->addTerm(new fl::Ramp("SMALL", 0.5, 0));
```
alexvins: Should be `SHORT`

alexvins: Also if we use MEDIUM term here, why not use LOW and HIGH too? And input variable configuration Ramp-Triangle-Ramp should factored out to function or even to subclass of `fl::InputVariable`

AI/VCAI/Fuzzy.cpp
```diff
@@ -373,142 +373,93 @@ float FuzzyHelper::evaluate(Goals::RecruitHero & g)
...
+		strengthRatio->addTerm(new fl::Ramp("HIGH", SAFE_ATTACK_CONSTANT, SAFE_ATTACK_CONSTANT * 3));
+		strengthRatio->setRange(0, SAFE_ATTACK_CONSTANT * 3);
 
 		//strength compared to our main hero
-		vt.heroStrength->addTerm(new fl::Ramp("LOW", 0.2, 0));
-		vt.heroStrength->addTerm(new fl::Triangle("MEDIUM", 0.2, 0.8));
-		vt.heroStrength->addTerm(new fl::Ramp("HIGH", 0.5, 1));
-		vt.heroStrength->setRange(0.0, 1.0);
+		heroStrength->addTerm(new fl::Ramp("LOW", 0.2, 0));
+		heroStrength->addTerm(new fl::Triangle("MEDIUM", 0.2, 0.8));
```
alexvins: If input is exactly 0.2, no term will be actually active

DjWarmonger: Terms must cover entire range of values, and at least one rule must trigger for every combination.

alexvins: Terms of input variables must overlap, it this case at 0.2 no rule will be triggered (membership function is zero for all of them)

AI/VCAI/Fuzzy.cpp
```diff
@@ -373,142 +373,93 @@ float FuzzyHelper::evaluate(Goals::RecruitHero & g)
...
+		heroStrength->setRange(0.0, 1.0);
 
-		vt.turnDistance->addTerm(new fl::Ramp("SMALL", 0.5, 0));
-		vt.turnDistance->addTerm(new fl::Triangle("MEDIUM", 0.1, 0.8));
-		vt.turnDistance->addTerm(new fl::Ramp("LONG", 0.5, 3));
-		vt.turnDistance->setRange(0.0, 3.0);
+		turnDistance->addTerm(new fl::Ramp("SMALL", 0.5, 0));
+		turnDistance->addTerm(new fl::Triangle("MEDIUM", 0.1, 0.8));
+		turnDistance->addTerm(new fl::Ramp("LONG", 0.5, 3));
+		turnDistance->setRange(0.0, 3.0);
```
alexvins: All terms have default height (1.0) but variables have custom ranges, I doubt that this is intended.

dydzio0614:
About dwellings... I tried, but with current objects code I was unable to calculate dwelling values. Creature data for particular dwelling type is apparently initialized at object instantiation and I found no way to handle it properly via VLC->objtypeh or VLC->creh

DjWarmonger:
But what is the problem? For every dwelling object we can determine the creature type(s) it gives. AI can already buy creatures from dwellings, so it can also know their parameters.

dydzio0614:
We cannot determine the creature it gives for mods. We can read json node for creature name, but creature handler cannot resolve creature by name alone, because it also needs creature namespace as 2nd parameter - for non-mod creatures it is "core".

DjWarmonger:
Don't read any json here lol, just use actual dwelling properties from *obj.
Like here: https://github.com/vcmi/vcmi/blob/develop/AI/VCAI/Goals.cpp#L1545

dydzio0614:
The code you provided works on CGObjectInstance, map objects evaluator works on TObjectTypeHandler and is not tied to objects on current map.

DjWarmonger:
So it's bad design on your side :P VCAI always evaluates concrete objects of CGObjectInstance * it sees in a map.

dydzio0614:
Debatable, I find this logic pretty straightforward, human player also knows objects that do not exist on particular map :P

DjWarmonger:
Well yes, but VCAI has no use of objects that do not exist on a map. Especially not for fuzzy evaluation. You are probably confusing FindObj with GetObj, which could be renamed to VisitObj and for all purposes is equal to VisitTile.

dydzio0614:
Until 2 days ago I thought it is possible to get all object data without creating CGObjectInstance and this is the annoying design part IMO

DjWarmonger:
Uhh, many comments in VCAI say "TODO" for that matter :P

DjWarmonger:
Either way, I don't see why MapObjectsEvaluator couldn't convert CGObjectInstance to ID and subID, this is trivial to implement.

dydzio0614:
MapObjectEvaluator state is persistent between games as far as I remember and constructor adding objects to database is not called again when starting new map with new objects

dydzio0614:
I may try to create mock objectinstances for all dwellings but I do not guarantee good performance of that solution

DjWarmonger:
Can stay like that if you wish to, but then again I made this graphs for a purpose: https://wiki.vcmi.eu/File:VCAI_Target.png

DjWarmonger:
What on earth are you talking about, just add a function that takes CGObjectInstance * as an argument

dydzio0614:
Okay, this branch is done. Needs to be tested before merge.

DjWarmonger:
Error from console:
```
FindWanderTarget: [syntax error] consequent expected output variable, but found <value>
{at \ai\fuzzylite\fuzzylite\src\rule\consequent.cpp::fl::Consequent::load() [line:182]}
```

DjWarmonger:
Also getting crash every time, in different situations:
`Cannot get object with id 5135. Object was removed`

DjWarmonger:
Okay I think we can merge this one. I already have a pack of bugfixes based on this branch.

dydzio0614:
Yeah, I agree.

PR #477, opened 2018-08-05
Vcai rework

DjWarmonger:
Reworked VCAI core, but more work is needed.
* VCAI exhaustively decomposes all possible goals
* abstract goals are added to the pool
* goals who failed decomposition are removed from pool
* AI is very slow, needs improvement

- [x] Remove striveToGoal() altogether
- [x] Remove makeTurnInternal()

### DIFF
--- a/AI/VCAI/ResourceManager.cpp
+++ b/AI/VCAI/ResourceManager.cpp
@@ -217,8 +217,8 @@ bool ResourceManager::notifyGoalCompleted(Goals::TSubgoal goal)
 		});
 		if (it != queue.end()) //removed at least one
 		{
+			logAi->debug("Removing goal %s from ResourceManager.", it->goal->name());
 			queue.erase(queue.s_handle_from_iterator(it));
-			logAi->debug("Removed goal %s from ResourceManager.", it->goal->name());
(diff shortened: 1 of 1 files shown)


AI/VCAI/VCAI.cpp
```diff
@@ -781,129 +781,212 @@ void VCAI::makeTurn()
...
 		}
+	}
+}
 
-		//now try to win
-		/*below line performs goal decomposition, result of the function is ONE goal for ONE hero to realize.*/
-		striveToGoal(sptr(Goals::Win()));
+void VCAI::mainLoop()
+{
+	std::vector<Goals::TSubgoal> basicGoals, elementarGoals; //no duplicates allowed (operator ==)
```
alexvins: Why not use std::set then?

DjWarmonger: std::set needs < operator... which I just implemented afterwards :P

DjWarmonger: Ahh yes, now I remember. I want to remove duplicates using == operator, which is stronger (removes also non-identical goals).

PR #480, opened 2018-08-21
attribute for battle image added on hero class, allowing on hero sche…

fiorin:
Added support for unique battle defs for heroes configurable via .json with format:

"images" : {"large" : "portrait.bmp", "small" : "portrait.bmp","specialtyLarge" : "specialty.bmp","specialtySmall" : "specialty.bmp","battleImage" : "custom.def"}

That feature allow create semi-neutral heroes in pre-existante classes with different graphicals.

The map adventure object is avaiable with filters om 'map object' attribute via json

### DIFF
--- a/client/Graphics.cpp
+++ b/client/Graphics.cpp
@@ -449,7 +449,6 @@ void Graphics::initializeImageLists()
 		addImageListEntry(hero->imageIndex, "UN44", hero->iconSpecLarge);
 		addImageListEntry(hero->imageIndex, "PORTRAITSLARGE", hero->portraitLarge);
 		addImageListEntry(hero->imageIndex, "PORTRAITSSMALL", hero->portraitSmall);
-		addImageListEntry(hero->imageIndex, "BATTLEIMAGE", hero->battleImage);
 	}
 
 	for(const CArtifact * art : CGI->arth->artifacts)
--- a/client/Graphics.cpp
+++ b/client/Graphics.cpp
@@ -449,6 +449,7 @@ void Graphics::initializeImageLists()
 		addImageListEntry(hero->imageIndex, "UN44", hero->iconSpecLarge);
 		addImageListEntry(hero->imageIndex, "PORTRAITSLARGE", hero->portraitLarge);
 		addImageListEntry(hero->imageIndex, "PORTRAITSSMALL", hero->portraitSmall);
+		addImageListEntry(hero->imageIndex, "BATTLEIMAGE", hero->battleImage);
 	}
 
 	for(const CArtifact * art : CGI->arth->artifacts)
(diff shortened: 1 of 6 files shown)


lib/CHeroHandler.h
```diff
@@ -121,6 +122,7 @@ class DLL_LINKAGE CHero
 		h & iconSpecLarge;
 		h & portraitSmall;
 		h & portraitLarge;
+		h & battleImage;
```
alexvins: Must be new format version (like in lines below)

client/Graphics.cpp
```diff
@@ -449,6 +449,7 @@ void Graphics::initializeImageLists()
 		addImageListEntry(hero->imageIndex, "UN44", hero->iconSpecLarge);
 		addImageListEntry(hero->imageIndex, "PORTRAITSLARGE", hero->portraitLarge);
 		addImageListEntry(hero->imageIndex, "PORTRAITSSMALL", hero->portraitSmall);
+		addImageListEntry(hero->imageIndex, "BATTLEIMAGE", hero->battleImage);
```
alexvins: Just remove this line, it makes no sense.

lib/CHeroHandler.h
fiorin: Lines below ? 
"std::vector<std::shared_ptr<Bonus>> SpecialtyInfoToBonuses(const SSpecialtyInfo & spec, int sid);"

That format ?

alexvins: No

alexvins:
```
if(version >= 759)
{
	h & identifier;
}
```

alexvins: Also change SERIALIZATION_VERSION in CSerializer.h

lib/CHeroHandler.h
```diff
@@ -122,9 +122,9 @@ class DLL_LINKAGE CHero
 		h & iconSpecLarge;
 		h & portraitSmall;
 		h & portraitLarge;
-		h & battleImage;
 		if(version >= 759)
 		{
+			h & battleImage;
```
alexvins: Correct is:
```
	if(version >= 790)
	{
		h & battleImage;
	}
```

after old code

fiorin: Thought the 'new handle format' was implemented after 759. But know i understand, we dont have a massive change on format, just a ponctual ones based on version. Thanks for that clarify

ArseniyShestakov:
Thanks for your contribution!

I just updated wiki as well:
https://wiki.vcmi.eu/Hero_Format
Please add some info on animation dimensions here if you can. Thanks!

fiorin:
Maybe im gonna need change that structure because i couldnt override the 'images' object structure from core heroes. Im trying some things. Soon i came back with a patch. And i provide some info too

PR #484, opened 2018-08-23
Create artifact AI value handling

dydzio0614:


AI/VCAI/MapObjectsEvaluator.cpp
```diff
@@ -68,8 +70,47 @@ boost::optional<int> MapObjectsEvaluator::getObjectValue(const CGObjectInstance
 		}
 		return aiValue;
 	}
-	else
-		return getObjectValue(obj->ID, obj->subID);
+	else if(obj->ID == Obj::ARTIFACT)
```
alexvins: A function start to grow. `getObjectValue` should be moved to `CGObjectInstance` as virtual method

PR #488, opened 2018-08-25
Stabilize vcmi

KasyanDiGris:
This PR solve following issues:

- Segfault in town because of using already removed statusbar
- Server segfault on last client disconnected

### DIFF
--- a/server/CVCMIServer.cpp
+++ b/server/CVCMIServer.cpp
@@ -445,7 +445,8 @@ void CVCMIServer::clientDisconnected(std::shared_ptr<CConnection> c)
 	connections -= c;
 	for(auto it = playerNames.begin(); it != playerNames.end();)
 	{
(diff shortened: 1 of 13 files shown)


server/CVCMIServer.cpp
```diff
@@ -443,14 +443,16 @@ void CVCMIServer::clientConnected(std::shared_ptr<CConnection> c, std::vector<st
 void CVCMIServer::clientDisconnected(std::shared_ptr<CConnection> c)
 {
 	connections -= c;
-	for(auto & pair : playerNames)
+	for(auto it = playerNames.begin(); it != playerNames.end();)
 	{
-		if(pair.second.connection != c->connectionID)
+		if(it->second.connection != c->connectionID) {
```
alexvins: put `{` on new line

PR #499, opened 2018-10-21
Client pack araives before client is initialized.

nullkiller:
Client pack arrives before client is initialized. The map can be downloaded from http://www.maps4heroes.com/forum/opinions.php?map_id=791&game=3
If somebody wants to investigate it deeper you are welcome. I am not strong at client-server interaction.


ArseniyShestakov:
Ok I actually started to debug this right now.

ArseniyShestakov:
Ok I created bug report here and I closing this one.
This is bug that don't need to be hotfixed like that.

https://bugs.vcmi.eu/view.php?id=3020

PR #500, opened 2018-10-27
Fix crash when generating random map with random number of players

godric3:
Changes  in CMapGenOptions are required, because without them generated map have only AI players(it will start, but after saving you couldn't load it)

### DIFF
--- a/lib/rmg/CMapGenOptions.cpp
+++ b/lib/rmg/CMapGenOptions.cpp
@@ -71,8 +71,13 @@ void CMapGenOptions::setPlayerCount(si8 value)
 	if (compOnlyPlayerCount > possibleCompPlayersCount)
 		setCompOnlyPlayerCount(possibleCompPlayersCount);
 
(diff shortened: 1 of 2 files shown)


lib/rmg/CMapGenOptions.cpp
```diff
@@ -167,7 +172,10 @@ void CMapGenOptions::resetPlayersMap()
 		auto pc = PlayerColor(color);
 		player.setColor(pc);
 		auto playerType = EPlayerType::AI;
-		if ((getPlayerCount() != RANDOM_SIZE && color >= realPlayersCnt)
+		if (getPlayerCount() != RANDOM_SIZE && color < realPlayersCnt)
+		{
+			playerType = EPlayerType::HUMAN;
+		}else if ((getPlayerCount() != RANDOM_SIZE && color >= realPlayersCnt)
```
alexvins: please fix formatting
```suggestion
		}
		else if((getPlayerCount() != RANDOM_SIZE && color >= realPlayersCnt)
```

PR #508, opened 2018-10-29
[WIP]Trying to debug creature town recruitment crash

dydzio0614:
Did not plan to make PR, but Travis did not put automated build for branch. Hopefully this time it will work.


Toneyisnow:
I'm also encountering this bug, but why do you push some Logging codes into main branch? I though it would handle some mal-formed config and make sure it's not crashing.

nullkiller:
The thing was fixed by MXE update

PR #509, opened 2018-10-29
Skill and hero specialty bonus description

godric3:
Fix for https://bugs.vcmi.eu/view.php?id=3011

### DIFF
--- a/lib/CHeroHandler.h
+++ b/lib/CHeroHandler.h
@@ -134,8 +134,8 @@ public:
 };
 
 // convert deprecated format
-std::vector<std::shared_ptr<Bonus>> SpecialtyInfoToBonuses(const SSpecialtyInfo & spec, int sid);
-std::vector<std::shared_ptr<Bonus>> SpecialtyBonusToBonuses(const SSpecialtyBonus & spec, int sid);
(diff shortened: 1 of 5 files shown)


lib/CHeroHandler.h
```diff
@@ -135,7 +135,7 @@ class DLL_LINKAGE CHero
 
 // convert deprecated format
 std::vector<std::shared_ptr<Bonus>> SpecialtyInfoToBonuses(const SSpecialtyInfo & spec, int sid);
-std::vector<std::shared_ptr<Bonus>> SpecialtyBonusToBonuses(const SSpecialtyBonus & spec);
+std::vector<std::shared_ptr<Bonus>> SpecialtyBonusToBonuses(const SSpecialtyBonus & spec, int sid);
```
DjWarmonger: How about default sid value to make this argument optional?

dydzio0614: optional argument... maybe boost::optional will be useful for this?

DjWarmonger: Nope, I meant just default `sid = 0` so coder doesn't need to write anything to skip this new feature.

godric3: I could add it, but is there reason to? This function is used for converting deprecated format, so it's not likely to be used more

PR #512, opened 2018-10-31
Support for using relative file path in json schema references

godric3:
This will enable the IDE to understand the references. In this way, IDE can be more helpful when writing jsons

### DIFF
--- a/lib/JsonNode.cpp
+++ b/lib/JsonNode.cpp
@@ -904,7 +904,7 @@ const JsonNode & getSchemaByName(std::string name)
 	if (vstd::contains(loadedSchemas, name))
 		return loadedSchemas[name];
 
-	std::string filename = "config/schemas/" + name + ".json";
+	std::string filename = "config/schemas/" + name;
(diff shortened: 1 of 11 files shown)


lib/JsonNode.cpp
```diff
@@ -904,7 +904,7 @@ const JsonNode & getSchemaByName(std::string name)
 	if (vstd::contains(loadedSchemas, name))
 		return loadedSchemas[name];
 
-	std::string filename = "config/schemas/" + name + ".json";
```
alexvins: Do not change this (and all "$ref" ref fields in schemas), use
`ResourceID(std::string name, EResType::Type type)` constructor in following lines instead. It will work both with and without extension.

godric3: My point was to change `$ref` fields, so any external app(IDE) will be able to validate mod's jsons against schema. With `vcmi:schemaName` it's not possible, because `vcmi` is not defined protocol.

PR #515, opened 2018-11-03
Erase boat object goals on hero death

dydzio0614:
This code may fix bug #3021 - unsure.


ArseniyShestakov:
Are you absolutely sure it's not handled by ```VCAI::objectRemoved```?

DjWarmonger:
I don't quite get it either. If hero is in a boat, he's still a hero. If there's boat without hero, it's just ordinary object. Even if there's a bug, this shouldn't need a special treatment.

dydzio0614:
Yeah, now this code looks invalid for me. The problem is that for some reason AI does not reliably remove goals related to boarded hero when hero on boat dies.

DjWarmonger:
IIRC there was a bug also where normal land hero is lost., is it any different?

PR #517, opened 2018-11-04
Hero try to execute a quest to kill himself

nullkiller:
There is an issue reported by povelitel that game hangs during AI turn. It was found out that the issue is in a quest to kill a hero

### DIFF
--- a/AI/VCAI/VCAI.cpp
+++ b/AI/VCAI/VCAI.cpp
@@ -2469,7 +2469,6 @@ Goals::TSubgoal VCAI::questToGoal(const QuestInfo & q)
 			}
 			return sptr(Goals::FindObj(Obj::PRISON)); //rule of a thumb - quest heroes usually are locked in prisons
 															 //BNLOG ("Don't know how to recruit hero with id %d\n", q.quest->m13489val);
-			break;
 		}
 		case CQuest::MISSION_ARMY:
 		{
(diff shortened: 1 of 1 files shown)


AI/VCAI/VCAI.cpp
```diff
@@ -2459,6 +2459,25 @@ Goals::TSubgoal VCAI::questToGoal(const QuestInfo & q)
 		}
 		case CQuest::MISSION_HERO:
 		{
+			if(q.obj->tempOwner == playerID)
+			{
+				if(q.obj->ID == Obj::HERO)
+				{
+					auto heroToProtect = cb->getHero(q.obj->id);
+
+					return sptr(Goals::GatherArmy().sethero(heroToProtect));
```
alexvins: I think you should return invalid goal here and setup protect goal some different way, because you can not complete this quest.

AI/VCAI/VCAI.cpp
```diff
@@ -2459,6 +2459,25 @@ Goals::TSubgoal VCAI::questToGoal(const QuestInfo & q)
 		}
 		case CQuest::MISSION_HERO:
 		{
+			if(q.obj->tempOwner == playerID)
```
alexvins: MISSION_HERO is not a kill hero quest, see MISSION_KILL_HERO below

nullkiller: Ok. Looks like this requires some additional investigation and proper testing.

nullkiller: Now it is ok.

PR #520, opened 2018-11-11
AI: GATHER_TROOPS improvements and fixes

nullkiller:
There is a freeze in GATHER_TROOPS. It counted on hero buying army in town while visiting it but now we need to request BuyArmy goal in order to do this. The hero only picks army on visiting and does not buy it.

### DIFF
--- a/AI/VCAI/Goals.cpp
+++ b/AI/VCAI/Goals.cpp
@@ -137,7 +137,7 @@ bool Goals::GatherArmy::operator==(const GatherArmy & other) const
 
 bool Goals::BuyArmy::operator==(const BuyArmy & other) const
 {
-	return town == other.town;
+	return town == other.town && objid == other.objid;
 }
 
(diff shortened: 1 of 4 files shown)


AI/VCAI/Goals.cpp
```diff
@@ -1219,11 +1219,58 @@ bool CollectRes::fulfillsMe(TSubgoal goal)
 	return false;
 }
 
+int GatherTroops::getCreaturesCount(const CArmedInstance * army)
```
dydzio0614: Is this function supposed to be used to determine value of army to gather? If yes, use creature AI Value in all relevant places so 3 pikemen won't be treated as better than 2 angels etc.

nullkiller: No, it is used to check if there is anything available at all. GATHER_TROOPS is for quests and it is bound to gather exact amount of something not army of certain power. For the last we have GatherArmy

PR #521, opened 2018-11-14
Fix creature animation callback circular reference

dydzio0614:
Probably best way to fix problem mentioned today on Slack. Use weak_ptr for safe circular reference when design relies on it.

### DIFF
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -47,26 +47,30 @@
 CondSh<bool> CBattleInterface::animsAreDisplayed(false);
 CondSh<BattleAction *> CBattleInterface::givenCommand(nullptr);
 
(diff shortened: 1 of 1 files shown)


client/battle/CBattleInterface.cpp
```diff
@@ -47,26 +47,26 @@
 CondSh<bool> CBattleInterface::animsAreDisplayed(false);
 CondSh<BattleAction *> CBattleInterface::givenCommand(nullptr);
 
-static void onAnimationFinished(const CStack *stack, std::shared_ptr<CCreatureAnimation> anim)
+static void onAnimationFinished(const CStack *stack, std::weak_ptr<CCreatureAnimation> anim)
 {
-	if (anim->isIdle())
+	if (anim.lock()->isIdle())
```
alexvins: lock once to local `shared_ptr` variable, also lock result should be checked

PR #523, opened 2018-11-17
Build boat in allied shipyards is allowed now

nullkiller:
One more freeze. AI wanted to build boat in allied shipyard. It was not allowed on server

### DIFF
--- a/server/NetPacksServer.cpp
+++ b/server/NetPacksServer.cpp
@@ -272,7 +272,9 @@ bool HireHero::applyGh(CGameHandler * gh)
 
 bool BuildBoat::applyGh(CGameHandler * gh)
 {
-	throwOnWrongOwner(gh, objid);
+	if(gh->getPlayerRelations(gh->getOwner(objid), gh->getPlayerAt(c)) == PlayerRelations::ENEMIES)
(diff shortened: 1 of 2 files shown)


server/NetPacksServer.cpp
```diff
@@ -272,7 +272,9 @@ bool HireHero::applyGh(CGameHandler * gh)
 
 bool BuildBoat::applyGh(CGameHandler * gh)
 {
-	throwOnWrongOwner(gh, objid);
+	if(gh->getPlayerRelations(gh->getOwner(objid), gh->getPlayerAt(c)) == PlayerRelations::ENEMIES)
+		throwAndCompain(gh, "Can't build boat at enemy shipyard");
```
DjWarmonger: Isn't this a typo?

nullkiller: Sure it is. Wanna me to fix it?

nullkiller: I also think I can extract a method throwWnenOwnerIsEnemy and use it for recruit hero

nullkiller: I also think I can extract a method throwWnenOwnerIsEnemy and use it for recruit hero

PR #524, opened 2018-11-18
AI goals refactoring

nullkiller:
Wanna split Goals.cpp onto small files


dydzio0614:
While you work on it, you may want on Goals namespace if it is really needed and dealwiththe mess where some goal class methods are in goal namespace andsome are not,  sofor example in visual studio you need to check members of Goals::VisitObj to find constructor and VisitObj without goals to get rest of methods. Some goals may be worth renaming to reflect their purpose clearly - for example GatherArmy and GatherTroops.

AI/VCAI/Goals/AbstractGoal.h
```diff
@@ -0,0 +1,262 @@
...
+			value = 0;
+			aid = -1;
+			objid = -1;
+			resID = -1;
+			tile = int3(-1, -1, -1);
+			town = nullptr;
+		}
+
+		OSETTER(bool, isElementar)
+			OSETTER(bool, isAbstract)
```
alexvins: wrong indentation

AI/VCAI/Goals/Build.cpp
```diff
@@ -0,0 +1,93 @@
...
+using namespace Goals;
+
+TGoalVec Goals::Build::getAllPossibleSubgoals()
+{
+	TGoalVec ret;
+
+	for(const CGTownInstance * t : cb->getTownsInfo())
+	{
+		//start fresh with every town
+		ai->ah->getBuildingOptions(t);
```
alexvins: Is return value ignored intentionally?

AI/VCAI/Goals/CollectRes.h
```diff
@@ -0,0 +1,40 @@
...
+		CollectRes()
+			: CGoal(Goals::COLLECT_RES)
+		{
+		}
+		CollectRes(int rid, int val)
+			: CGoal(Goals::COLLECT_RES)
+		{
+			resID = rid;
+			value = val;
+			priority = 2;
```
alexvins: It should be on common header with all that priority constants.

AI/VCAI/Goals/Build.cpp
nullkiller: I have not changed any logic. Just mechanical split of one big file.

AI/VCAI/Goals/CollectRes.h
nullkiller: Can put them in abstractgoal.h or in fuzzy helper (it is actually responsible for priority). technically they should not be hardcoded

AI/VCAI/Goals/Build.cpp
nullkiller: I was right. It fetches to internal state. It does not look like Get*

nullkiller:
I just want to split the big file. I do not plan any deep refactoring.

AI/VCAI/Goals/CollectRes.cpp
```diff
@@ -0,0 +1,208 @@
+/*
+ * Goals.cpp, part of VCMI engine
```
ArseniyShestakov: Please fix all filenames in headers so we don't have to do it after.

nullkiller: Ok, will do

ArseniyShestakov:
Tested and this seems to be working fine for me though it's expected. Merged.

PR #526, opened 2018-12-09
AI: CompleteQuest goal and summon boat spell support

nullkiller:
Extract quest related code from VCAI to a separate goal and implement support of Summon Boat spell.

### DIFF
--- a/AI/VCAI/Goals/AdventureSpellCast.cpp
+++ b/AI/VCAI/Goals/AdventureSpellCast.cpp
@@ -38,10 +38,10 @@ TSubgoal AdventureSpellCast::whatToDoToAchieve()
 	if(!spell->isAdventureSpell())
 		throw cannotFulfillGoalException(spell->name + " is not an adventure spell.");
 
-	if(!vstd::contains(hero->spells, spellID))
-		throw cannotFulfillGoalException("Hero has no " + spell->name);
+	if(!hero->canCastThisSpell(spell))
+		throw cannotFulfillGoalException("Hero can not cast " + spell->name);
 
-	if(hero->mana < hero->getSpellCost(spellID.toSpell()))
+	if(hero->mana < hero->getSpellCost(spell))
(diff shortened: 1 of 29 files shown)


ArseniyShestakov:
I'll just do little testing and then merge it since I just curious to test AI on my test maps a little.

nullkiller:
I have not checked it so. I am not much interested in fixing prioritization and touching mainLoop or decomposeGoal. The idea is to put some sign in front of town and see if AI will try to bypass sign and capture town or it will chose to visit some resource instead because sign priority is very low.

AI/VCAI/Pathfinding/AIPathfinderConfig.cpp
```diff
@@ -142,6 +182,84 @@ namespace AIPathfinding
 					logAi->debug("Virtual boat added at %s", boatLocation.toString());
 				}
 			}
+
+			auto hero = nodeStorage->getHero();
+
+			if(vstd::contains(hero->spells, SpellID::SUMMON_BOAT))
```
ArseniyShestakov: So ```hero->spells``` should never be used because it's not include spells added by bones system.

AI/VCAI/Goals/AdventureSpellCast.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+	logAi->trace("Decomposing adventure spell cast of %s for hero %s", spell->name, hero->name);
+
+	if(!spell->isAdventureSpell())
+		throw cannotFulfillGoalException(spell->name + " is not an adventure spell.");
+
+	if(!vstd::contains(hero->spells, spellID))
+		throw cannotFulfillGoalException("Hero has no " + spell->name);
+
+	if(hero->mana < hero->getSpellCost(spellID.toSpell()))
```
ArseniyShestakov: And yeah reuse that ```spell``` here.

PR #527, opened 2018-12-25
AI: fix freeze while endlesly trying to recruit hero

nullkiller:
Conquer goal sometimes want to recruit a hero. If it is possible but is declined by ResourceManager because it has something more important or wants to save money it stores RecruitHero goal in its own cache and waits for a better time. When the better time happens we just want to execute RecruitHero elementar version as it was stored in elementar form in ResourceManager. The last more or less blindly tries to recruit hero and fails with cannotCompleteGoal exception. But the one path which checks presence of any tavern does not. It just does nothing and exits. So we try to recruit hero again and again and goal does nothing. It is not completed because it did not throw goalCompletedException nor it canceled because it did not threw cannotCompleteGoalException.

### DIFF
--- a/AI/VCAI/VCAI.cpp
+++ b/AI/VCAI/VCAI.cpp
@@ -823,6 +823,7 @@ void VCAI::makeTurn()
 	}
 	markHeroAbleToExplore(primaryHero());
 	visitedHeroes.clear();
+	ai->ah->resetPaths();
 
 	try
 	{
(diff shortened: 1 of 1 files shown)


AI/VCAI/VCAI.cpp
```diff
@@ -1007,6 +1008,9 @@ void VCAI::mainLoop()
 				for (auto basicGoal : ultimateGoalsFromBasic[goalToRealize])
 					goalsToRemove.push_back(basicGoal);
 
+				// sometimes resource manager contains an elementar goal which is not able to execute anymore and just fails each turn.
+				ai->ah->notifyGoalCompleted(goalToRealize);
```
nullkiller: Just an optimization hack. Otherwise it will try to recruit hero once each turn until is completed.

AI/VCAI/VCAI.cpp
```diff
@@ -2056,6 +2060,10 @@ void VCAI::tryRealize(Goals::RecruitHero & g)
 		//TODO try to free way to blocked town
 		//TODO: adventure map tavern or prison?
 	}
+	else
+	{
+		throw cannotFulfillGoalException("No town to recruit hero!");
```
nullkiller: Main fix.

PR #528, opened 2018-12-25
AI: tweak explore to work with new pathfinding.

nullkiller:
Since the new pathfinding was implemented explorationDesperate function has been broken. I removed it because it duplicates logic of explorationNewPoint except of using sector map. Now the new pathfinding already contains all required info and I do not see any point of having two separate functions.

### DIFF
--- a/AI/VCAI/Goals/Explore.cpp
+++ b/AI/VCAI/Goals/Explore.cpp
@@ -28,7 +28,7 @@ using namespace Goals;
 
 bool Explore::operator==(const Explore & other) const
 {
-	return other.hero.h == hero.h;
+	return other.hero.h == hero.h && other.allowGatherArmy == allowGatherArmy;
 }
 
 std::string Explore::completeMessage() const
@@ -127,20 +127,33 @@ TGoalVec Explore::getAllPossibleSubgoals()
 	{
 		for(auto obj : objs) //double loop, performance risk?
 		{
-			auto waysToVisitObj = ai->ah->howToVisitObj(h, obj);
+			auto waysToVisitObj = ai->ah->howToVisitObj(h, obj, allowGatherArmy);
 
 			vstd::concatenate(ret, waysToVisitObj);
 		}
 
-		TSubgoal goal = howToExplore(h);
+		TSubgoal goal = exploreNearestNeighbour(h);
 
-		if(goal->invalid())
-		{
-			ai->markHeroUnableToExplore(h); //there is no freely accessible tile, do not poll this hero anymore
-		}
-		else
+		if(!goal->invalid())
 		{
 			ret.push_back(goal);
+		}
+	}
+
+	if(ret.empty())
+	{
+		for(auto h : heroes)
+		{
+			TSubgoal goal = explorationNewPoint(h);
+
+			if(goal->invalid())
+			{
+				ai->markHeroUnableToExplore(h); //there is no freely accessible tile, do not poll this hero anymore
+			}
+			else
+			{
+				ret.push_back(goal);
+			}
 		}
 	}
 
@@ -152,7 +165,6 @@ TGoalVec Explore::getAllPossibleSubgoals()
 	{
 		throw goalFulfilledException(sptr(Explore().sethero(hero)));
 	}
-	//throw cannotFulfillGoalException("Cannot explore - no possible ways found!");
 
 	return ret;
 }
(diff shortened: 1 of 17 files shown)


AI/VCAI/Goals/Explore.cpp
```diff
@@ -179,3 +167,220 @@ bool Explore::fulfillsMe(TSubgoal goal)
...
+
+	return false;
+}
+
+int Explore::howManyTilesWillBeDiscovered(
+	const int3 & pos, 
+	int radious, 
+	CCallback * cbp, 
+	HeroPtr hero, 
+	std::function<bool (const int3 &)> filter) const
```
nullkiller: Moved and added a filter. explorationNewPoint does not use the filter anymore. explorationBestNeighbor uses hasVisibleNeighbors as the filter

AI/VCAI/Goals/Explore.cpp
```diff
@@ -179,3 +167,220 @@ bool Explore::fulfillsMe(TSubgoal goal)
...
+			{
+				ret++;
+			}
+		}
+	}
+
+	return ret;
+}
+
+TSubgoal Explore::explorationBestNeighbour(int3 hpos, int radius, HeroPtr h) const
```
nullkiller: Just moved and a bit changed the signature. Now it returns TSubgoal to visit a tile instead of the tile itself.

AI/VCAI/Goals/Explore.cpp
```diff
@@ -179,3 +167,220 @@ bool Explore::fulfillsMe(TSubgoal goal)
...
+{
+	int ret = 0;
+	for(int x = pos.x - radious; x <= pos.x + radious; x++)
+	{
+		for(int y = pos.y - radious; y <= pos.y + radious; y++)
+		{
+			int3 npos = int3(x, y, pos.z);
+			if(cbp->isInTheMap(npos) 
+				&& pos.dist2d(npos) - 0.5 < radious 
+				&& !cbp->isVisible(npos)
```
alexvins: This is well-known bottleneck. As long as the function has been re-factored, I'd like it to be improved too.
The problem is that `isVisible` does searching for Team State and also duplicates `isInTheMap` check. We should obtain TeamState outside of a cycle and access `fogOfWarMap` directly

AI/VCAI/Goals/Explore.cpp
```diff
@@ -179,3 +167,220 @@ bool Explore::fulfillsMe(TSubgoal goal)
...
+TSubgoal Explore::explorationBestNeighbour(int3 hpos, int radius, HeroPtr h) const
+{
+	std::map<int3, int> dstToRevealedTiles;
+	VCAI * aip = ai.get();
+	CCallback * cbp = cb.get();
+
+	for(crint3 dir : int3::getDirs())
+	{
+		int3 tile = hpos + dir;
+		if(cb->isInTheMap(tile))
```
alexvins: Same here

AI/VCAI/Goals/Explore.cpp
```diff
@@ -179,3 +167,220 @@ bool Explore::fulfillsMe(TSubgoal goal)
...
+	return result;
+}
+
+void Explore::getVisibleNeighbours(const std::vector<int3> & tiles, std::vector<int3> & out, CCallback * cbp) const
+{
+	for(const int3 & tile : tiles)
+	{
+		foreach_neighbour(tile, [&](int3 neighbour)
+		{
+			if(cbp->isVisible(neighbour))
```
alexvins: Same here, most important case.

AI/VCAI/Goals/Explore.cpp
```diff
@@ -179,3 +168,222 @@ bool Explore::fulfillsMe(TSubgoal goal)
...
+	CCallback * cbp = cb.get();
+	VCAI *aip = ai.get();
+	const TeamState * ts = cbp->getPlayerTeam(aip->playerID);
+
+	std::vector<std::vector<int3>> tiles; //tiles[distance_to_fow]
+	tiles.resize(radius);
+
+	foreach_tile_pos([&](const int3 & pos)
+	{
+		if(!cbp->isVisible(pos))
```
alexvins: TeamState should be used here too

AI/VCAI/Goals/Explore.cpp
```diff
@@ -179,3 +168,222 @@ bool Explore::fulfillsMe(TSubgoal goal)
...
+			tiles[0].push_back(pos);
```
alexvins: `tiles[0]` would contain all hidden tiles of entire map, you should at least reserve some space in a vector.

nullkiller:
Logger experiments were removed from this PR.

alexvins:
Merged already right before force push. Please start new PR if there is more to do.

PR #535, opened 2019-01-10
VCAI Fixes

alexvins:
Overall game speed (in headless mode, gui performance is out of scope) has been improved. AI pathfinding is still a bottleneck, even inlined version of graph initialization is slow. Pathfinder constructor is now 2 times faster itself called less often but still uses 20% of total CPU time (was 50%). Several smaller mostly bonus related bottlenecks has been fixed too.

### DIFF
--- a/AI/VCAI/FuzzyEngines.cpp
+++ b/AI/VCAI/FuzzyEngines.cpp
@@ -89,14 +89,15 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 
 float HeroMovementGoalEngineBase::calculateTurnDistanceInputValue(const Goals::AbstractGoal & goal) const
 {
-	const float maxMovePoints = (float)goal.hero->maxMovePoints(true);
-
-	if(goal.evaluationContext.movementCost != 0)
+	if(goal.evaluationContext.movementCost > 0)
 	{
-		return goal.evaluationContext.movementCost / maxMovePoints;
+		return goal.evaluationContext.movementCost;
 	}
-
-	return distanceToTile(goal.hero.h, goal.tile) / maxMovePoints;
+	else
+	{
+		auto pathInfo = ai->myCb->getPathsInfo(goal.hero.h)->getPathInfo(goal.tile);
+		return pathInfo->cost;
+	}
 }
 
 TacticalAdvantageEngine::TacticalAdvantageEngine()
--- a/AI/VCAI/FuzzyEngines.cpp
+++ b/AI/VCAI/FuzzyEngines.cpp
@@ -50,15 +50,25 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 	double shootersStrenght = 0;
 	ui32 maxSpeed = 0;
 
+	static const CSelector selectorSHOOTER = Selector::type(Bonus::SHOOTER);
+	static const std::string keySHOOTER = "type_"+std::to_string((int32_t)Bonus::SHOOTER);
+
+	static const CSelector selectorFLYING = Selector::type(Bonus::FLYING);
+	static const std::string keyFLYING = "type_"+std::to_string((int32_t)Bonus::FLYING);
+
+	static const CSelector selectorSTACKS_SPEED = Selector::type(Bonus::STACKS_SPEED);
+	static const std::string keySTACKS_SPEED = "type_"+std::to_string((int32_t)Bonus::STACKS_SPEED);
+
 	for(auto s : army->Slots())
 	{
 		bool walker = true;
-		if(s.second->type->hasBonusOfType(Bonus::SHOOTER))
+		const CCreature * creature = s.second->type;
+		if(creature->hasBonus(selectorSHOOTER, keySHOOTER))
 		{
 			shootersStrenght += s.second->getPower();
 			walker = false;
 		}
-		if(s.second->type->hasBonusOfType(Bonus::FLYING))
+		if(creature->hasBonus(selectorFLYING, keyFLYING))
 		{
 			flyersStrenght += s.second->getPower();
 			walker = false;
@@ -66,7 +76,7 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 		if(walker)
 			walkersStrenght += s.second->getPower();
 
-		vstd::amax(maxSpeed, s.second->type->valOfBonuses(Bonus::STACKS_SPEED));
+		vstd::amax(maxSpeed, creature->valOfBonuses(selectorSTACKS_SPEED, keySTACKS_SPEED));
 	}
 	armyStructure as;
 	as.walkers = walkersStrenght / totalStrenght;
@@ -79,12 +89,14 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 
 float HeroMovementGoalEngineBase::calculateTurnDistanceInputValue(const Goals::AbstractGoal & goal) const
 {
+	const float maxMovePoints = (float)goal.hero->maxMovePoints(true);
+
 	if(goal.evaluationContext.movementCost != 0)
 	{
-		return goal.evaluationContext.movementCost / (float)goal.hero->maxMovePoints(true);
+		return goal.evaluationContext.movementCost / maxMovePoints;
 	}
 
-	return distanceToTile(goal.hero.h, goal.tile) / (float)goal.hero->maxMovePoints(true);
+	return distanceToTile(goal.hero.h, goal.tile) / maxMovePoints;
 }
 
 TacticalAdvantageEngine::TacticalAdvantageEngine()
--- a/AI/VCAI/Goals/Explore.cpp
+++ b/AI/VCAI/Goals/Explore.cpp
@@ -244,7 +244,8 @@ TSubgoal Explore::explorationBestNeighbour(int3 hpos, int radius, HeroPtr h) con
 				auto distance = hpos.dist2d(tile); // diagonal movement opens more tiles but spends more mp
 				int tilesDiscovered = howManyTilesWillBeDiscovered(tile, radius, cbp, ts, aip, h);
 
-				dstToRevealedTiles[tile] = tilesDiscovered / distance;
+				if(tilesDiscovered > 0)
+					dstToRevealedTiles[tile] = tilesDiscovered / distance;
 			}
 		}
 	}
@@ -371,9 +372,9 @@ TSubgoal Explore::exploreNearestNeighbour(HeroPtr h) const
 	int radius = h->getSightRadius();
 	int3 hpos = h->visitablePos();
 
-	//look for nearby objs -> visit them if they're close enouh
+	//look for nearby objs -> visit them if they're close enough
 	const int DIST_LIMIT = 3;
-	const int MP_LIMIT = DIST_LIMIT * 150; // aproximate cost of diagonal movement
+	const int MP_LIMIT = DIST_LIMIT * 150; // approximate cost of diagonal movement
 
 	std::vector<const CGObjectInstance *> nearbyVisitableObjs;
 	for(int x = hpos.x - DIST_LIMIT; x <= hpos.x + DIST_LIMIT; ++x) //get only local objects instead of all possible objects on the map
--- a/AI/VCAI/Goals/Explore.cpp
+++ b/AI/VCAI/Goals/Explore.cpp
@@ -253,11 +253,12 @@ TSubgoal Explore::explorationBestNeighbour(int3 hpos, int radius, HeroPtr h) con
 	if(dstToRevealedTiles.empty()) //yes, it DID happen!
 		return sptr(Invalid());
 
+	auto paths = cb->getPathsInfo(h.get());
+
 	auto best = dstToRevealedTiles.begin();
 	for(auto i = dstToRevealedTiles.begin(); i != dstToRevealedTiles.end(); i++)
 	{
-		const CGPathNode * pn = cb->getPathsInfo(h.get())->getPathInfo(i->first);
-		//const TerrainTile *t = cb->getTile(i->first);
+		const CGPathNode * pn = paths->getPathInfo(i->first);
 		if(best->second < i->second && pn->reachable() && pn->accessible == CGPathNode::ACCESSIBLE)
 			best = i;
 	}
@@ -336,7 +337,7 @@ TSubgoal Explore::explorationScanRange(HeroPtr h, std::vector<int3> & range) con
 		auto waysToVisit = aip->ah->howToVisitTile(h, tile, allowGatherArmy);
 		for(auto goal : waysToVisit)
 		{
-			if(goal->evaluationContext.movementCost == 0) // should not happen
+			if(goal->evaluationContext.movementCost <= 0.0) // should not happen
 				continue;
 
 			float ourValue = (float)tilesDiscovered * tilesDiscovered / goal->evaluationContext.movementCost;
@@ -374,7 +375,7 @@ TSubgoal Explore::exploreNearestNeighbour(HeroPtr h) const
 
 	//look for nearby objs -> visit them if they're close enough
 	const int DIST_LIMIT = 3;
-	const int MP_LIMIT = DIST_LIMIT * 150; // approximate cost of diagonal movement
+	const float COST_LIMIT = .2; //todo: fine tune
 
 	std::vector<const CGObjectInstance *> nearbyVisitableObjs;
 	for(int x = hpos.x - DIST_LIMIT; x <= hpos.x + DIST_LIMIT; ++x) //get only local objects instead of all possible objects on the map
@@ -383,7 +384,7 @@ TSubgoal Explore::exploreNearestNeighbour(HeroPtr h) const
 		{
 			for(auto obj : cb->getVisitableObjs(int3(x, y, hpos.z), false))
 			{
-				if(ai->isGoodForVisit(obj, h, MP_LIMIT))
+				if(ai->isGoodForVisit(obj, h, COST_LIMIT))
 				{
 					nearbyVisitableObjs.push_back(obj);
 				}
--- a/AI/VCAI/Goals/Explore.h
+++ b/AI/VCAI/Goals/Explore.h
@@ -54,9 +54,9 @@ namespace Goals
 		bool hasReachableNeighbor(const int3 &pos, HeroPtr hero, CCallback * cbp, VCAI * vcai) const;
 
 		void getVisibleNeighbours(
-			const std::vector<int3> & tiles, 
-			std::vector<int3> & out, 
-			CCallback * cbp, 
+			const std::vector<int3> & tiles,
+			std::vector<int3> & out,
+			CCallback * cbp,
 			const TeamState * ts) const;
 
 		int howManyTilesWillBeDiscovered(
--- a/AI/VCAI/Pathfinding/AINodeStorage.cpp
+++ b/AI/VCAI/Pathfinding/AINodeStorage.cpp
@@ -120,6 +120,7 @@ CGPathNode * AINodeStorage::getInitialNode()
 	initialNode->turns = 0;
 	initialNode->moveRemains = hero->movement;
 	initialNode->danger = 0;
+	initialNode->cost = 0.0;
 
 	return initialNode;
 }
@@ -146,6 +147,7 @@ void AINodeStorage::commit(CDestinationNodeInfo & destination, const PathNodeInf
 	{
 		dstNode->moveRemains = destination.movementLeft;
 		dstNode->turns = destination.turn;
+		dstNode->cost = destination.cost;
 		dstNode->danger = srcNode->danger;
 		dstNode->action = destination.action;
 		dstNode->theNodeBefore = srcNode->theNodeBefore;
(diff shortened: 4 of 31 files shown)


AI/VCAI/Goals/Explore.cpp
```diff
@@ -213,8 +196,7 @@ int Explore::howManyTilesWillBeDiscovered(
 			int3 npos = int3(x, y, pos.z);
 			if(cbp->isInTheMap(npos)
 				&& pos.dist2d(npos) - 0.5 < radious
-				&& !ts->fogOfWarMap[npos.x][npos.y][npos.z]
-				&& hasReachableNeighbor(npos, h, cbp, aip))
```
nullkiller: Why do you think it is wrong? It is dead end exploring cancellation.

alexvins: This function counts unexplored tiles within sight radius. `hasReachableNeighbor` check does not make sense.

nullkiller: It checks if unexplored tile has reachable neighbor an thus can be visited later. Otherwise there is much lower probability it will be useful. Even if it is not just a rock but a tile behind some border. The only way to reach it in such case is to cast air walking or dimension door. AI can not do both for now.

nullkiller: It might be an option to count potentially visitable tile as 1 and potentially unreachable as 0.1 or something. If it behaves badly somewhere lets look onto particular case.

AI/VCAI/Goals/Explore.h
```diff
@@ -51,12 +51,11 @@ namespace Goals
 		TSubgoal explorationNewPoint(HeroPtr h) const;
 		TSubgoal explorationBestNeighbour(int3 hpos, int radius, HeroPtr h) const;
 		TSubgoal explorationScanRange(HeroPtr h, std::vector<int3> & range) const;
-		bool hasReachableNeighbor(const int3 &pos, HeroPtr hero, CCallback * cbp, VCAI * vcai) const;
 
 		void getVisibleNeighbours(
-			const std::vector<int3> & tiles, 
-			std::vector<int3> & out, 
-			CCallback * cbp, 
+			const std::vector<int3> & tiles,
```
nullkiller: Wrong line endings?

AI/VCAI/Goals/Explore.cpp
alexvins: If tile is not visible yet it is not reachable too. Reachable neighbour check only works for border of FoW.

AI/VCAI/Goals/Explore.h
alexvins: Spaces before line ending.

AI/VCAI/Goals/Explore.cpp
nullkiller: Yes, it is an expected behavior. The idea is that AI should not get close to border because this way it explores behind it and thus explores unreachable area.

nullkiller: Does it hurt performance? We can try to check if there is any visitable tile but do not require it being reachable. This way we will cancel exploration of rocks. But it requires cb->getTile()

ArseniyShestakov: Basically all changes you see in this branch are about what Alex found in Valgrind's Callgrind output.

nullkiller: Just removing code is not always the best action to optimize things. I am afraid AI might start spending more turns on exploring thus being less productive. I have some ideas to optimize Explore logically by introducing the middle stage when we scan an area around hero instead of all map. Having such stage allows to disable this check in full scan. Also Explore can try to acquire kind of tile accessibility matrix which will be allocated and cached in PathfindingManager

AI/VCAI/VCAI.cpp
```diff
@@ -1844,7 +1844,7 @@ bool VCAI::isAccessibleForHero(const int3 & pos, HeroPtr h, bool includeAllies)
 			}
 		}
 	}
-	return cb->getPathsInfo(h.get())->getPathInfo(pos)->reachable();
+	return ah->isTileAccessible(h, pos);
```
nullkiller: This is not equivalent. PathfindingManager will return true if tile is accessible through a battle or any different way and reachable() returns true only if it is directly accessible. But in general it should be replaced.

AI/VCAI/Goals/Explore.cpp
nullkiller: Comment out this function. I will try to find a better solution. We just need to reduce scan area somehow. And I have a few ideas.

alexvins: I can not remove `howManyTilesWillBeDiscovered` entirely. It is used in both `explorationScanRange` and `explorationBestNeighbour`

nullkiller: I mean hasReachableNeighbor

alexvins: done

nullkiller: Thanks! I will try to do something about it in the nearest future.

AI/VCAI/VCAI.cpp
alexvins: Is there an equivalent in AIPathfinder?

AI/VCAI/Goals/Explore.cpp
DjWarmonger: Wait, this piece of code replaced old one, improving both AI performance and speed. I don't quite see how you could remove it entirely.

nullkiller: hasReachableNeighbor is for dead end cancellation. It improves exploration behavior but obviously not performance

AI/VCAI/VCAI.cpp
nullkiller: Btw VisitTile itself is an example of code which will defenitelly break if swithced to 
ah->isTileAccessibleForHero because it expects direct access

nullkiller: Technically it is possible to implement equivalent function. AIPathfinder nodes are tiles where a hero has to do something like battle or spell cast. So they are not neighbors like in normal CPathfinder. In order to move between them we still need to call cb->getPathsInfo() but only at tryRealize(VisitTile) when we actually decided to move a hero.

So if AIPathfinder path has the length of 1 it is directly accessible. We can implement another version of isAccessibleForHero and check any path length to be 1. 

But this is not needed. We just need to avoid using VisitTile/VisitObj directly but use ah->howToVisitTile/Obj instead. If it returns an empty vector of goals the object is not accessible. So we need to replace all direct usages of VistTile/Obj with ah->howToVisitTile/Obj whenever it is used and remove all logic from VisitTile regarding to checking if tile is accessible and invoking ClearWayTo. But this is my opinion and it is a lot of work.

alexvins: The actual problem is lack of paths cache in Client, single hero cache misses a lot (10 times more path recalculations that should be).

nullkiller: Or you can do it yourself. The same way as it is already done but you need to check that any node also has no previousNeighbor and specialAction

AI/VCAI/Goals/Explore.cpp
alexvins: What if there is a quest artefact hidden in the mountains (only accessible by the Door by design), will AI be able to find it with dead end cancellation?

AI/VCAI/VCAI.cpp
alexvins: I will do something anyway, still not decided what exactly, may be even both.

AI/VCAI/Goals/Explore.cpp
nullkiller: Good question. Still I feel like dead end cancellation is needed but at the same time there should be a way to explore all possible tiles as low priority task.

alexvins: Lets add dead end check only to `explorationBestNeighbour`. So hero will efficiently continue exploring but start at any tile.

alexvins:
Main bottlenecks so far:
* bonus caching, building "cache keys" OTF (f.e. `IBonusBearer::hasBonusOfType`). 
*  `CGHeroInstance::maxMovePoints` usage without supplying TurnInfo

AI/VCAI/Goals/Explore.cpp
nullkiller: This does not work 100% nice. Sure it helps at certain extent. AI goes to dead end, cancels it and now he has nothing to explore because in front is dead end and behind is already explored. It enables full scan without dead-end cancellation and finds tiles in front so he goes and explores them because they are very close.

nullkiller: It is very tricky, this exploration. Maybe with middle stage where we scan in range around hero with dead end cancellation and only after we we enable to find a target near we run full scan with disabled dead end we can achieve better behavior.

nullkiller: Also as an option we can cancel exploration for a whole turn if it once yield no results. Also exploration sometimes triggered twice or even more doing exactly the same thing. Conquer, GatherArmy and CollectRes can produce exploration for instance. Explore decomposes to Explore(hero) instead of directly returning found goals. Also as an option we can put Explore to the list of basic goals and remove it from everywhere else thus allowing to decompose it only once per hero move. Also we can move hero not one tile at a time but 3 tiles or half of exploration radius thus reducing amount of decomposition cycles

nullkiller: Also might be an option to check if vcmieagles ai is set and disable explore at all

DjWarmonger: Can we just skip this line for this PR and test non-controversial performance optimizations?

nullkiller: Comment it out, already done I suppose and I will think about Explore optimization later. I expect I have to establish some profiler locally on Windows and will be able to see what is going on myself.

DjWarmonger: There's already heroesUnableToExplore container with logic exactly for that.
However, it's difficult to exclude exploration at all. If AI wants to explore, it means either it didn't explore entire map yet or map contains no necessary objectives at all.

nullkiller: Probably you are right, it should not take much time if there is no visible tiles. The worst case probably is 50-75% open map

lib/PathfinderUtil.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include "mapping/CMapDefines.h"
+#include "CGameState.h"
+
+namespace PathfinderUtil
+{
+	using FoW = std::vector<std::vector<std::vector<ui8> > >;
+	using ELayer = EPathfindingLayer;
+
+	template<EPathfindingLayer::EEPathfindingLayer layer>
```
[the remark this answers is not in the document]
alexvins: It is an optimization. There a `switch` on layer inside the function, because now layer is a constant inside it can be completely optimized out by compiler.

AI/VCAI/Pathfinding/AIPathfinderConfig.cpp
```diff
@@ -76,12 +76,14 @@ namespace AIPathfinding
 
 		bool isAffordableBy(HeroPtr hero, const AIPathNode * source) const
 		{
+#ifdef VCMI_TRACE_PATHFINDER
```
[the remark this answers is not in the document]
ArseniyShestakov: I really dislike this one though...

AI/VCAI/Pathfinding/AINodeStorage.cpp
```diff
@@ -10,15 +10,61 @@
...
+	int3 pos;
+	int3 sizes = gs->getMapSize();
+	const auto & fow = static_cast<const CGameInfoCallback *>(gs)->getPlayerTeam(hero->tempOwner)->fogOfWarMap;
+	const PlayerColor player = hero->tempOwner;
+
+	//make 200% sure that these are loop invariants (also a bit shorter code), let compiler do the rest(loop unswitching)
+	const bool useFlying = options.useFlying;
+	const bool useWaterWalking = options.useWaterWalking;
+
+	for(pos.x=0; pos.x < sizes.x; ++pos.x)
```
FeniksFire: will be nice to overload some operators in int3 and replace 3 loops with something like that `for (int3 pos; pos<sizes; ++pos)`

AI/VCAI/FuzzyEngines.cpp
```diff
@@ -50,23 +50,33 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 	double shootersStrenght = 0;
 	ui32 maxSpeed = 0;
 
+	static const CSelector selectorSHOOTER = Selector::type(Bonus::SHOOTER);
+	static const std::string keySHOOTER = "type_"+std::to_string((si32)Bonus::SHOOTER);
```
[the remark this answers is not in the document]
dydzio0614: I somewhat agree, while we can have our own "vcmi standard library" stuff, there is no reason to introduce custom shortcuts for trivial stuff that force people to learn codebase-specific things for full understanding and make them use it.

lib/CPathfinder.cpp
```diff
@@ -17,14 +17,57 @@
...
 
 bool canSeeObj(const CGObjectInstance * obj)
 {
 	/// Pathfinder should ignore placed events
 	return obj != nullptr && obj->ID != Obj::EVENT;
 }
 
+void NodeStorage::initialize(const PathfinderOptions & options, const CGameState * gs, const CGHeroInstance * hero)
+{
+	int3 pos;
```
FeniksFire: code dublication with `AINodeStorage::initialize`?

lib/logging/CLogger.cpp
```diff
@@ -387,7 +389,8 @@ void CLogConsoleTarget::setColorMapping(const CColorMapping & colorMapping) { th
 CLogFileTarget::CLogFileTarget(boost::filesystem::path filePath, bool append)
 	: file(std::move(filePath), append ? std::ios_base::app : std::ios_base::out)
 {
-	formatter.setPattern("%d %l %n [%t] - %m");
+//	formatter.setPattern("%d %l %n [%t] - %m");
```
FeniksFire: why this change?

alexvins: VCMI write so much logs, so date formatting in too slow

lib/CPathfinder.h
```diff
@@ -82,21 +117,23 @@ struct DLL_LINKAGE CPathsInfo
...
 
-	mutable boost::mutex pathMx;
-
 	const CGHeroInstance * hero;
 	int3 hpos;
 	int3 sizes;
 	boost::multi_array<CGPathNode, 4> nodes; //[w][h][level][layer]
 
-	CPathsInfo(const int3 & Sizes);
+	CPathsInfo(const int3 & Sizes, const CGHeroInstance * hero_);
```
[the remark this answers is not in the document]
alexvins: I do not like to have method/constructor parameters with same name as existing field.

AI/VCAI/Pathfinding/AINodeStorage.cpp
```diff
@@ -10,15 +10,61 @@
...
+	{
+		for(pos.y=0; pos.y < sizes.y; ++pos.y)
+		{
+			for(pos.z=0; pos.z < sizes.z; ++pos.z)
+			{
+				const TerrainTile * tinfo = &gs->map->getTile(pos);
```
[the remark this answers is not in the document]
FeniksFire: yep, far better than `tinfo`

AI/VCAI/Pathfinding/AINodeStorage.cpp
alexvins: I have no idea how to make such custom iterator fast, you have to add not trivial logic to `++` and compare all 3 coordinates each iteration.

AI/VCAI/FuzzyEngines.cpp
```diff
@@ -50,23 +50,33 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
...
+
+	static const CSelector selectorFLYING = Selector::type(Bonus::FLYING);
+	static const std::string keyFLYING = "type_"+std::to_string((si32)Bonus::FLYING);
+
+	static const CSelector selectorSTACKS_SPEED = Selector::type(Bonus::STACKS_SPEED);
+	static const std::string keySTACKS_SPEED = "type_"+std::to_string((si32)Bonus::STACKS_SPEED);
```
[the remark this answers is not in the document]
alexvins: It is possible for  some similar simple cases, but generic solution is impossible (selector can be _any_ functor with `bool(const Bonus*)` signature)

AI/VCAI/FuzzyEngines.cpp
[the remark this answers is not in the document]
alexvins: done

AI/VCAI/Pathfinding/AINodeStorage.cpp
```diff
@@ -10,15 +10,61 @@
...
 
-AINodeStorage::~AINodeStorage()
+AINodeStorage::~AINodeStorage() = default;
+
+void AINodeStorage::initialize(const PathfinderOptions & options, const CGameState * gs, const CGHeroInstance * hero)
 {
+	//TODO: fix this code duplication with NodeStorage::initialize, problem is to keep `resetTile` inline
+
+	int3 pos;
+	int3 sizes = gs->getMapSize();
```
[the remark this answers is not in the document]
alexvins: done

AI/VCAI/Pathfinding/AINodeStorage.cpp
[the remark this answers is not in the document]
alexvins: done

AI/VCAI/Goals/Explore.cpp
DjWarmonger: What is "commented out", the function call is still removed in this PR.

nullkiller: I have a few ideas for myself.
1. add a middle stage where AI scans 20x20 area around hero for a new point
2. scan perimeter of visible area before expanding it by sight radius and if it yields any result - return it. If all perimeter tiles are blocked expand and scan with deadend cancellation disabled.
3. might be worth to split AI pathfinder API onto one updatePaths(std:vector<HeroPtr>) and the rest is const. This way we can drop synchronization using mutex at all I think.

alexvins: Reverted. Any further improvements considered out of scope.

lib/CPathfinder.cpp
alexvins: additional comment added

client/VCMI_client.cbp
```diff
@@ -13,6 +13,7 @@
 				<Option object_output="../obj/Client/Debug/x86" />
 				<Option type="1" />
 				<Option compiler="gcc" />
+				<Option parameters="--testmap=Maps/Reclamation --spectate --spectate-skip-battle --spectate-ignore-hero" />
```
nullkiller: Is it an intended change?

alexvins: Yes it is.

alexvins: I decided it is better to have default run parameters for debug to be automated run for now.

lib/CPathfinder.cpp
```diff
@@ -193,16 +195,24 @@ void MovementCostRule::process(
 	const PathfinderConfig * pathfinderConfig,
 	CPathfinderHelper * pathfinderHelper) const
 {
-	int turnAtNextTile = destination.turn, moveAtNextTile = destination.movementLeft;
+	bool textTurn = false;
```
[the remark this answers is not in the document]
alexvins: fixed

lib/CPathfinder.cpp
```diff
@@ -211,6 +221,12 @@ void MovementCostRule::process(
 		remains = pathfinderHelper->movementPointsAfterEmbark(moveAtNextTile, cost, destination.action - 1);
 	}
 
+	if(textTurn)
+		costAtNextTile = static_cast<float>(turnAtNextTile) +  static_cast<float>(cost) / maxMovePoints;
```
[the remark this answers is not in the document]
alexvins: Should be fixed now.

lib/CPathfinder.cpp
```diff
@@ -193,16 +195,24 @@ void MovementCostRule::process(
 	const PathfinderConfig * pathfinderConfig,
 	CPathfinderHelper * pathfinderHelper) const
 {
-	int turnAtNextTile = destination.turn, moveAtNextTile = destination.movementLeft;
+	bool textTurn = false;
+	float costAtNextTile = destination.cost;
```
nullkiller: Where is destination.cost set?

lib/CPathfinder.cpp
```diff
@@ -297,12 +313,16 @@ void CPathfinder::calculatePaths()
...
+		int movement = source.node->moveRemains;
+		uint8_t turn = source.node->turns;
+		float cost = source.node->cost;
+
 		hlp->updateTurnInfo(turn);
 		if(!movement)
 		{
 			hlp->updateTurnInfo(++turn);
 			movement = hlp->getMaxMovePoints(source.node->layer);
+			cost = static_cast<float>(turn);
```
nullkiller: Do we need this anymore?

alexvins: Already removed

lib/CPathfinder.cpp
alexvins: In `CPathfinder::calculatePaths()` main loop

PR #536, opened 2019-01-10
AI: town portal support

nullkiller:
Town portal support for AI

- [x] Use town portal to access objects in ah->howToVisitTile/Obj
- [x] Check it works with gather army so AI can visit all towns and gather army from them

### DIFF
--- a/AI/VCAI/Pathfinding/AINodeStorage.cpp
+++ b/AI/VCAI/Pathfinding/AINodeStorage.cpp
@@ -9,7 +9,12 @@
 */
 #include "StdInc.h"
 #include "AINodeStorage.h"
+#include "../Goals/Goals.h"
+#include "../../../CCallback.h"
+#include "../../../lib/mapping/CMap.h"
+#include "../../../lib/mapObjects/MapObjects.h"
 
+extern boost::thread_specific_ptr<CCallback> cb;
 
 AINodeStorage::AINodeStorage(const int3 & Sizes)
 	: sizes(Sizes)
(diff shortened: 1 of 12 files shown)


lib/CPathfinder.cpp
```diff
@@ -315,7 +319,7 @@ void CPathfinder::calculatePaths()
 
 		/// For now we disable teleports usage for patrol movement
 		/// VCAI not aware about patrol and may stuck while attempt to use teleport
-		if(!source.isNodeObjectVisitable() || patrolState == PATROL_RADIUS)
```
ArseniyShestakov: What this change have to do with implementation of town portal support?

ArseniyShestakov: Oops ok I see you just moved it to proper place. Sorry.

ArseniyShestakov:
Just like with boat spell I wish you could reuse more of Alex code there.
I understand that it's can really be later in refactoring, but still.

alexvins:
> Just like with boat spell I wish you could reuse more of Alex code there.
> I understand that it's can really be later in refactoring, but still.
Adventure spells to be redesigned the same way battle spells works, after that we could think about generalization of AI support.

AI/VCAI/Pathfinding/AINodeStorage.cpp
```diff
@@ -133,28 +138,120 @@ std::vector<CGPathNode *> AINodeStorage::calculateNeighbours(
...
+	const PathNodeInfo & source,
+	std::vector<CGPathNode *> & neighbours)
+{
+	SpellID spellID = SpellID::TOWN_PORTAL;
+	const CSpell * townPortal = spellID.toSpell();
+	auto srcNode = getAINode(source.node);
+
+	if(hero->canCastThisSpell(townPortal) && hero->mana >= hero->getSpellCost(townPortal))
+	{
+		auto towns = cb->getTownsInfo();
```
ArseniyShestakov: Unfortunately town portal also work include ally towns. Please look at:
```
TownPortalMechanics::getPossibleTowns
```
This really need to be worked around properly.

And just in case please add some comment that it is code duplication of sorts even if Alex still need more refactoring on his code.

viciious:
Shouldn't this also include support for Inferno's town portal building?

ArseniyShestakov:
Inferno's castle gate is better to be implemented like generic teleporter map object so human players could benefit from automatic pathfinding through them too. Though it's bring little bit of pain in ass since our map object code is still mess.

PR #537, opened 2019-01-13
AI: fix visit obj crash

ArseniyShestakov:
So this is very dumb fix, but I need some confirmation from AI guys before I merge it to develop.

UPD: to clarify what exactly I trying to fix. Problem is that VCAI use VisitObj goal to visit hero and when hero become invisible after FoW change it's cause crash since apparently ```VCAI::heroMoved``` used ```validateObject``` for enemy heroes. Apparently ```validateObject``` doesn't remove ```lockedHeroes``` goals when ebemt hero is no longer visible.

### DIFF
--- a/AI/VCAI/FuzzyEngines.cpp
+++ b/AI/VCAI/FuzzyEngines.cpp
@@ -378,7 +378,11 @@ float VisitObjEngine::evaluate(Goals::VisitObj & goal)
 		return 0;
 
 	auto obj = ai->myCb->getObj(ObjectInstanceID(goal.objid));
-
+	if(!obj)
+	{
+		logAi->error("Goals::VisitObj objid " + std::to_string(goal.objid) + " no longer visible, probably goal used for something it's not intended");
+		return -100; // FIXME: Added check when goal was used for hero instead of VisitHero, but crashes are bad anyway
+	}
 
 	boost::optional<int> objValueKnownByAI = MapObjectsEvaluator::getInstance().getObjectValue(obj);
 	int objValue = 0;
--- a/AI/VCAI/Pathfinding/PathfindingManager.cpp
+++ b/AI/VCAI/Pathfinding/PathfindingManager.cpp
@@ -88,7 +88,7 @@ Goals::TGoalVec PathfindingManager::howToVisitObj(HeroPtr hero, ObjectIdRef obj,
 
 	auto result = findPath(hero, dest, allowGatherArmy, [&](int3 firstTileToGet) -> Goals::TSubgoal
 	{
-		if(obj->ID.num == Obj::HERO && obj->getOwner() == hero->getOwner())
+		if(obj->ID == Obj::HERO)
 			return sptr(Goals::VisitHero(obj->id.getNum()).sethero(hero).setisAbstract(true));
 		else
 			return sptr(Goals::VisitObj(obj->id.getNum()).sethero(hero).setisAbstract(true));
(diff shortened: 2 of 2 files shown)


AI/VCAI/Pathfinding/PathfindingManager.cpp
```diff
@@ -88,15 +88,18 @@ Goals::TGoalVec PathfindingManager::howToVisitObj(HeroPtr hero, ObjectIdRef obj,
...
 			return sptr(Goals::VisitObj(obj->id.getNum()).sethero(hero).setisAbstract(true));
 	});
 
 	for(Goals::TSubgoal solution : result)
 	{
-		solution->setparent(sptr(Goals::VisitObj(obj->id.getNum()).sethero(hero).setevaluationContext(solution->evaluationContext)));
+		if(obj->ID == Obj::HERO)
+			solution->setparent(sptr(Goals::VisitHero(obj->id.getNum()).sethero(hero).setevaluationContext(solution->evaluationContext)));
+		else
+			solution->setparent(sptr(Goals::VisitObj(obj->id.getNum()).sethero(hero).setevaluationContext(solution->evaluationContext)));
```
ArseniyShestakov: I honestly have no clue if it's right thing I done here, but I suppose it's make sense to keep abstract goal to visit hero and non-abstract one that can be dismissed if e.g hero go out of range.

ArseniyShestakov:
Added little explanation of what I trying to fix here. Basically problem is that enemy hero that no longer visible remain in goal within ```lockedHeroes``` since ```VCAI::validateObject``` don't cleanup there.

So probably I should also add cleanup code in ```validateObject``` just like one that already used in ```VCAI::objectRemoved```. Problem is that I not certain how much of code need to be used for hero that become invisible.

DjWarmonger:
It is perfectly possible that enemy hero goes out of sight and we can't evaluate goal in such scenario.

But target objects could also dissapear in many different scenarios, including future custom scripts, and we need to handle such case.

As to this "setParent" thing, this is something I didn't see before - anybody care to explain?

ArseniyShestakov:
I'll really need @nullkiller comment here or @dydzio0614.
After all it's their code.

nullkiller:
setParent is mine. It links goals produced by pathfinder with VisitObj goal to visit the object for which pathfinder was activated. For instance BuildBoat itself use hardly useful. We build boat to access water in order to explore/take some object and so on.

AI/VCAI/Pathfinding/PathfindingManager.cpp
dydzio0614: VisitObj should be used to visit enemy heroes

AI/VCAI/FuzzyEngines.cpp
```diff
@@ -378,7 +378,11 @@ float VisitObjEngine::evaluate(Goals::VisitObj & goal)
 		return 0;
 
 	auto obj = ai->myCb->getObj(ObjectInstanceID(goal.objid));
-
+	if(!obj)
```
dydzio0614: While this solution would fix VisitObj crashes - all crashes appearing here indicate failed logic somewhere else.

AI/VCAI/Pathfinding/PathfindingManager.cpp
dydzio0614: VisitObj is intended to be used to visit enemy heroes

nullkiller:
VCAI::heroMoved used validateObject for enemy heroes. Can we just disable/fix this? Goals bound to invisible objects should be cleared from VCAI state and still not added to ai->visitableObjs and so on.

AI/VCAI/FuzzyEngines.cpp
ArseniyShestakov: Yes, but that's what error messages are for. I'd rather not to have crashes in AI even when something fails.

ArseniyShestakov:
@dydzio0614 - VisitHero code have messages that make it clear it's was initially designed to visit enemy heroes. What do you think about it?

dydzio0614:
When reworking GetObj -> VisitObj I made that one intended to be used to visit enemy heroes.

nullkiller:
@dydzio0614 what is the status of this PR. Can we merge it?

dydzio0614:
No idea anymore, but based on comments I think I changed the design so VisitObj became new intended goal for visiting heroes somewhere along the way so the fix stopped becoming relevant

nullkiller:
If we can not merge it we should close it. Or at least try to reproduce and if it still exist - report issue on github and link this PR there

nullkiller:
[run-away.zip](https://github.com/vcmi/vcmi/files/12503681/run-away.zip)
Can not reproduce it any longer.

PR #540, opened 2019-01-19
AI: optimizae explore logically

nullkiller:
Reduce amount of scanned tiles during full map scan for a new exploration point. Some small optimization.

### DIFF
--- a/AI/VCAI/Goals/Explore.cpp
+++ b/AI/VCAI/Goals/Explore.cpp
@@ -26,6 +26,208 @@ extern FuzzyHelper * fh;
 
 using namespace Goals;
 
+namespace Goals
+{
+	struct ExplorationHelper
+	{
+		HeroPtr hero;
+		int sightRadius;
+		float bestValue;
+		TSubgoal bestGoal;
+		VCAI * aip;
+		CCallback * cbp;
+		const TeamState * ts;
+		int3 ourPos;
+		bool allowDeadEndCancellation;
+		bool allowGatherArmy;
+
+		ExplorationHelper(HeroPtr h, bool gatherArmy)
+		{
+			cbp = cb.get();
+			aip = ai.get();
+			hero = h;
+			ts = cbp->getPlayerTeam(ai->playerID);
+			sightRadius = hero->getSightRadius();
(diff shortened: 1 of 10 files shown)


AI/VCAI/Goals/Explore.cpp
```diff
@@ -26,6 +26,208 @@ extern FuzzyHelper * fh;
...
+			ts = cbp->getPlayerTeam(ai->playerID);
+			sightRadius = hero->getSightRadius();
+			bestGoal = sptr(Goals::Invalid());
+			bestValue = 0;
+			ourPos = h->convertPosition(h->pos, false);
+			allowDeadEndCancellation = true;
+			allowGatherArmy = gatherArmy;
+		}
+
+		void scanSector(int radius)
```
alexvins: radius not used

AI/VCAI/Pathfinding/AINodeStorage.cpp
```diff
@@ -342,19 +342,11 @@ bool AINodeStorage::hasBetterChain(const PathNodeInfo & source, CDestinationNode
...
-	auto chains = nodes[pos.x][pos.y][pos.z][layer];
-
-	for(const AIPathNode & node : chains)
-	{
-		if(node.action != CGPathNode::ENodeAction::UNKNOWN)
-		{
-			return true;
-		}
-	}
+	auto node = nodes[pos.x][pos.y][pos.z][layer][0];
```
alexvins: Is this `auto` copy of reference?

nullkiller: Should I use auto & here? And should I use the same structure in other places where I use not all 5 dimensions? For instance here? It will be boost::multi_array::sub_array
auto chains = nodes[pos.x][pos.y][pos.z][layer];

nullkiller: Done

AI/VCAI/Goals/Explore.cpp
nullkiller: Fixed

PR #541, opened 2019-01-24
Fix broken build on libboost 1.69.0

ilmikko:
The build seems to be broken with the new boost library.
This is caused by the implicit conversions of boost::logic::tribool into bool in the files below.
Explicit casting seems to fix the issue; PR to fix the build.

Built on: g++ 8.2.1 20181127
Libboost version: 1.69.0


alexvins:
Thank you for the contribution.

PR #546, opened 2019-02-06
AI pathfinding const api and updatePaths

nullkiller:
The idea is to make howToVisit* methods of PathfindingManager constant and introduce a separate API for paths update. Also I think it should be able to work without synchronization.

### DIFF
--- a/AI/VCAI/FuzzyEngines.cpp
+++ b/AI/VCAI/FuzzyEngines.cpp
@@ -19,6 +19,7 @@
 #define UNGUARDED_OBJECT (100.0f) //we consider unguarded objects 100 times weaker than us
 
 extern boost::thread_specific_ptr<VCAI> ai;
+extern FuzzyHelper * fh;
 
 engineBase::engineBase()
 {
@@ -202,8 +203,6 @@ TacticalAdvantageEngine::TacticalAdvantageEngine()
 
 float TacticalAdvantageEngine::getTacticalAdvantage(const CArmedInstance * we, const CArmedInstance * enemy)
 {
-	boost::unique_lock<boost::mutex> lock(mx);
-
 	float output = 1;
 	try
 	{
@@ -344,7 +343,7 @@ void HeroMovementGoalEngineBase::setSharedFuzzyVariables(Goals::AbstractGoal & g
 	}
 
 	float strengthRatioData = 10.0f; //we are much stronger than enemy
-	ui64 danger = evaluateDanger(goal.tile, goal.hero.h);
+	ui64 danger = fh->evaluateDanger(goal.tile, goal.hero.h);
 	if(danger)
 		strengthRatioData = (fl::scalar)goal.hero.h->getTotalStrength() / danger;
 
--- a/AI/VCAI/FuzzyEngines.cpp
+++ b/AI/VCAI/FuzzyEngines.cpp
@@ -202,6 +202,8 @@ TacticalAdvantageEngine::TacticalAdvantageEngine()
 
 float TacticalAdvantageEngine::getTacticalAdvantage(const CArmedInstance * we, const CArmedInstance * enemy)
 {
+	boost::unique_lock<boost::mutex> lock(mx);
+
 	float output = 1;
 	try
 	{
--- a/AI/VCAI/Pathfinding/AIPathfinder.cpp
+++ b/AI/VCAI/Pathfinding/AIPathfinder.cpp
@@ -53,6 +53,15 @@ void AIPathfinder::updatePaths(std::vector<HeroPtr> heroes)
 {
 	storageMap.clear();
 
+	auto calculatePaths = [&](const CGHeroInstance * hero, std::shared_ptr<AIPathfinding::AIPathfinderConfig> config)
+	{
+		logAi->debug("Recalculate paths for %s", hero->name);
+
+		cb->calculatePaths(config, hero);
+	};
+
+	std::vector<Task> calculationTasks;
+
 	// TODO: go parallel?
 	for(HeroPtr hero : heroes)
 	{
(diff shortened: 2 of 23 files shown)


alexvins:
Please rebase on develop, it should fix build.

AI/VCAI/VCAI.cpp
```diff
@@ -1392,6 +1400,8 @@ void VCAI::wander(HeroPtr h)
 	while(h->movement)
 	{
 		validateVisitableObjs();
+		ah->updatePaths(getMyHeroes());
```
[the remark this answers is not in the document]
nullkiller: Good question. It looks like yes but it is possible that some goal like CompleteQuest may decide to check all heroes. Maybe it is possible to update paths of one hero without clearing storage map. I will think about it.

AI/VCAI/Pathfinding/AIPathfinder.cpp
```diff
@@ -84,13 +73,43 @@ std::shared_ptr<AINodeStorage> AIPathfinder::getOrCreateStorage(const HeroPtr &
...
 	}
+}
+
+void AIPathfinder::updatePaths(const HeroPtr & hero)
+{
+	std::shared_ptr<AINodeStorage> nodeStorage;
+
+	if(!vstd::contains(storageMap, hero))
+	{
+		if(storageMap.size() < storagePool.size())
```
alexvins: This duplicates code in `AIPathfinder::updatePaths(std::vector<HeroPtr>)`.

nullkiller: I know. What do you think about parallel calculation of paths for all heroes? Can you suggest how it can be implemented? I mean is there any thread pooling tool or some libraries?

nullkiller: Can I try this https://en.cppreference.com/w/cpp/thread/async

alexvins: There is `CThreadHelper` in vcmi, it is very simple but should be sufficient for such task.

alexvins: Also there is a thread pool in boost, you should check it out.

nullkiller: How can I determine the amount of threads for CThreadHelper? I expect it should equal to amount of CPU cores?

nullkiller: Or maybe amount of heroes. It should not be more than 8. Or minimum between the first and the second.

alexvins: Use `boost::thread::hardware_concurrency()` for amount of threads. There is a complete usage example in `CBattleAI::attemptCastingSpell()`

alexvins: _minimum between the first and the second_ sounds reasonable

nullkiller: Can logger be accessed concurrently?

alexvins: Yes it can.

AI/VCAI/Pathfinding/AIPathfinder.cpp
```diff
@@ -53,6 +53,15 @@ void AIPathfinder::updatePaths(std::vector<HeroPtr> heroes)
...
+	auto calculatePaths = [&](const CGHeroInstance * hero, std::shared_ptr<AIPathfinding::AIPathfinderConfig> config)
+	{
+		logAi->debug("Recalculate paths for %s", hero->name);
+
+		cb->calculatePaths(config, hero);
+	};
+
+	std::vector<Task> calculationTasks;
+
 	// TODO: go parallel?
```
alexvins: remove this comment

AI/VCAI/Pathfinding/AIPathfinder.cpp
```diff
@@ -69,12 +78,29 @@ void AIPathfinder::updatePaths(std::vector<HeroPtr> heroes)
...
-		logAi->debug("Recalculate paths for %s", hero->name);
-		cb->calculatePaths(config, hero.get());
+		calculationTasks.push_back(std::bind(calculatePaths, hero.get(), config));
+	}
+
+	int threadsCount = std::min(
+		boost::thread::hardware_concurrency(),
+		(uint32_t)calculationTasks.size());
+
+	if(threadsCount == 1)
```
alexvins: `boost::thread::hardware_concurrency()` could be zero in case of error, you should check that.

AI/VCAI/FuzzyEngines.cpp
```diff
@@ -202,6 +202,8 @@ TacticalAdvantageEngine::TacticalAdvantageEngine()
 
 float TacticalAdvantageEngine::getTacticalAdvantage(const CArmedInstance * we, const CArmedInstance * enemy)
 {
+	boost::unique_lock<boost::mutex> lock(mx);
```
alexvins: It should be more clear why `getTacticalAdvantage` is synchronized while other fuzzy engines methods not.

AI/VCAI/Pathfinding/AIPathfinder.cpp
nullkiller: How to handle 0? Retry or do not use parallelism (thread count = 1)

AI/VCAI/FuzzyEngines.cpp
nullkiller: Made pathfinding having its own instance of FuzzyHelper. No need in mutex now

AI/VCAI/Pathfinding/AIPathfinder.cpp
nullkiller: For now I check if tread count is <= 1 and if it is I use synchronous processing

alexvins: Do not use.

alexvins:
With this merged I`ve got reproducible crash on assertion in `BattleAttack::applyGs`. Both in client and server.

PR #551, opened 2019-02-12
Disable tactical advantage engine until rework

dydzio0614:
As discussed on slack it'd be good idea to disable TA engine for now.


AI/VCAI/AIUtility.cpp
```diff
@@ -237,7 +237,7 @@ ui64 evaluateDanger(crint3 tile, const CGHeroInstance * visitor)
 	if(const CGObjectInstance * dangerousObject = vstd::backOrNull(visitableObjects))
 	{
 		objectDanger = evaluateDanger(dangerousObject); //unguarded objects can also be dangerous or unhandled
-		if(objectDanger)
+		/*if(objectDanger) //TODO: uncomment this code block after improving tactical advantage engine
```
alexvins: You should disable code in `getTacticalAdvantage` let it return 1 for now.

PR #553, opened 2019-02-24
ai fix town portal to occupied town

nullkiller:


AI/VCAI/Goals/AdventureSpellCast.cpp
```diff
@@ -44,6 +44,9 @@ TSubgoal AdventureSpellCast::whatToDoToAchieve()
 	if(hero->mana < hero->getSpellCost(spell))
 		throw cannotFulfillGoalException("Hero has not enough mana to cast " + spell->name);
 
+	if(spellID == SpellID::TOWN_PORTAL && town && town->visitingHero)
+		throw cannotFulfillGoalException("The town is already occupied by " + town->visitingHero->name);
```
alexvins: I`ve pushed this already. This check fixes infinite loop and should remain for safety. The problem is that `AINodeStorage::calculateTownPortalTeleportations` should not try to use invalid portal action.

There is other issue with town portal action. `town->visitingHero` is checked in `TownPortalMechanics::beginCast` and `VCAI::showMapObjectSelectDialog` receives only valid towns. AI should handle the case when desired object is not present in the list.

nullkiller: I will check calculateTownPortalTeleportations but looks like the main issue is that wander  calculated paths only for one hero and thus other heroes might think that the town is empty. I will check this issue next week if it is fully fixed or not. Sure the best way is to remove updatePaths for a single hero and recalculate everything. It is the safest way.

nullkiller: Btw as far as I understood castSpell uses supplied tile to select target town and does not issue the dialog in this case. Is there a way to understand if it failed?

alexvins: There no API for that. If you supply position for town portal, server expects it to be 100% valid. And the easiest way is to heck it is add `SystemMessage` callback to `IGameEventsReceiver`. Correct way is check if request succeeded in `VCAI::requestRealized` (but `CGameHandler::handleReceivedPack` should to be fixed too for that)

alexvins: So generic solution is to check for both:
* if any `SystemMessage` packet received during `VCAI::tryRealize` it should throw `cannotFulfillGoalException`
* additionally same for `VCAI::requestRealized` with 0 result.
The question is if it safe to throw there or not

nullkiller: Finally I have understood that we have to recalculate all paths in wander and not only for a particular hero. Any moving hero can block/unblock paths for other heroes and thus their paths might change. And we can any time have a decomposition to ClearWayTo without hero specified or GatherArmy or something new and will have bugs again. So updatePaths for a single hero should be removed. I will check at weekends more accurately.

alexvins: OK. I`m merging this one, but `calculateTownPortalTeleportations` to be refactored later anyway.

nullkiller: for(const CGTownInstance * targetTown : towns)
		{
			if(targetTown->visitingHero)
				continue;

PR #555, opened 2019-03-13
AI: fix hero exchange logic + refactoring

nullkiller:
I encountered some problems while making exchange between two heroes each has 1 stack which creatures are different. I used 1 pikeman and 10 archangels and wanted to exchange them vise versa.
As a result the new logic tried to rebalance archangels and pikeman stacks.


dydzio0614:
I want to test this before merge. Can you explain the problem in details?

nullkiller:
We have two heroes each has 1 stack and stacks are of different creatures. If the strongest stack is with the wrong hero the logic of stack split is activated and tries to balance both stacks and move creatures from one to another. But the creatures of a different type and server reports an error. No actual hero exchange happens

DjWarmonger:
I'd like to check if this fixes the issue with heroes spamming army exchange.

DjWarmonger:
Nah, there's still a flaw in logic which causes heroes to spam each other if they are allowed to visit each other many times every turn. This patch reverts the workaround for that I made some time ago:

[visitedHeroesBug.zip](https://github.com/vcmi/vcmi/files/2962558/visitedHeroesBug.zip)

nullkiller:
Ok I will think about it and will be back with more robust solution. For now I just found out an issue and quick fixed it

nullkiller:
Refactored exchange logic so that it can now split the weakest stack but taking into account creature speed

dydzio0614:
Looks neat, I will make some test maps and check how AI exchanges perform. If new code replaces without bugs old canGetArmy code then it's nice step forward in code quality.

alexvins:
Broken build should be fixed first

PR #556, opened 2019-03-17
Hardcoded feature: no random week/month effects

dydzio0614:
Fairly useful feature that allows disabling week/month effects (except deity of fire graal effect). While for some players this is fun mechanic, for others these special events just ruin game balance.


Alucard648:
I would also want for moddable week/month effects. More variety for random week/month effects ala Heroes 5.

dydzio0614:
I guess flexible week/month effect modding support should be done as part of scripting system.

PR #557, opened 2019-03-17
Fix crystal dragon ability to match H3

dydzio0614:
Existence of at least 1 crystal dragon in any town or hero army should generate 3 crystals on first day of the  week. Hiding all crystal dragons in garrison, mine etc. stops crystal generation. More dragons in stack or dragon stacks do not change amount of crystals generated. I find the ability fairly "arbitrary" and "existing for compatibility with H3 mechanics" so I think it's OK not to make it flexible for modding.

### DIFF
--- a/config/creatures/neutral.json
+++ b/config/creatures/neutral.json
@@ -122,10 +122,7 @@
 			},
 			"crystals" :
 			{
-				"type" : "GENERATE_RESOURCE",
-				"subtype" : "resource.crystal",
-				"val" : 1,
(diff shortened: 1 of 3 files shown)


Alucard648:
I think, bonuses like this should be moddable in more flexible way, like trigger reward every X days. Crystal Dragons are prime example, Another WoG expamples are Diamond Dragons (+ 1 gem every day) and those divine avatars that once per week permanently increase hero`s primary skill by 1.

PR #562, opened 2019-03-22
Battle action start on client after intro sound

dydzio0614:
This PR adds missing delay for combat action to start when entering battle. Delay can be stopped by pressing ESC key. Still some details missing to make it work 100% as in H3, but there is big improvement already.

### DIFF
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -44,8 +44,6 @@
 #include "../../lib/NetPacks.h"
 #include "../../lib/UnlockGuard.h"
 
-int battleIntroSoundChannel; //required as variable for disabling it via ESC key
-
 CondSh<bool> CBattleInterface::animsAreDisplayed(false);
 CondSh<BattleAction *> CBattleInterface::givenCommand(nullptr);
 
--- a/client/battle/CBattleInterface.cpp
+++ b/client/battle/CBattleInterface.cpp
@@ -44,6 +44,8 @@
 #include "../../lib/NetPacks.h"
 #include "../../lib/UnlockGuard.h"
 
+int battleIntroSoundChannel; //required as variable for disabling it via ESC key
+
 CondSh<bool> CBattleInterface::animsAreDisplayed(false);
 CondSh<BattleAction *> CBattleInterface::givenCommand(nullptr);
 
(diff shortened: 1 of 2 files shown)


client/battle/CBattleInterface.cpp
```diff
@@ -44,6 +44,8 @@
 #include "../../lib/NetPacks.h"
 #include "../../lib/UnlockGuard.h"
 
+int battleIntroSoundChannel; //required as variable for disabling it via ESC key
```
FeniksFire: making such globals is bad idea

client/battle/CBattleInterface.cpp
```diff
@@ -104,7 +106,7 @@ CBattleInterface::CBattleInterface(const CCreatureSet *army1, const CCreatureSet
 	currentlyHoveredHex(-1), attackingHex(-1), stackCanCastSpell(false), creatureCasting(false), spellDestSelectMode(false), spellToCast(nullptr), sp(nullptr),
 	creatureSpellToCast(-1),
 	siegeH(nullptr), attackerInt(att), defenderInt(defen), curInt(att), animIDhelper(0),
-	myTurn(false), moveStarted(false), moveSoundHander(-1), bresult(nullptr)
+	myTurn(false), moveStarted(false), moveSoundHander(-1), bresult(nullptr), battleActionsStarted(false)
```
FeniksFire: I think in class initialization will be better one

client/battle/CBattleInterface.cpp
```diff
@@ -1630,6 +1638,9 @@ void CBattleInterface::setHoveredStack(const CStack *stack)
 
 void CBattleInterface::activateStack()
 {
+	if(!battleActionsStarted)
```
FeniksFire: making some getter method for such things will be nice

client/battle/CBattleInterface.cpp
dydzio0614: It probably would, but what about consistency with other variables atm? Some initialized here, some in-class would result in a mess.

client/battle/CBattleInterface.cpp
dydzio0614: But... this variable is private

client/battle/CBattleInterface.cpp
FeniksFire: I thought that constructor is a mess anyway, keeping consistency makes no sense when itself it's broken, but you can do as you wish

client/battle/CBattleInterface.cpp
FeniksFire: yep that makes a thing a bit strange, like before just thought that in the future refactoring this can be helpfull

PR #570, opened 2019-03-28
Simplify statements

FeniksFire:
-Simplify return statements across the code

### DIFF
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -577,11 +577,8 @@ bool CBattleInfoCallback::battleCanAttack(const CStack * stack, const CStack * t
 	auto &id = stack->getCreature()->idNumber;
 	if (id == CreatureID::FIRST_AID_TENT || id == CreatureID::CATAPULT)
 		return false;
-
-	if (!target->alive())
-		return false;
-
-	return true;
+	
+	return target->alive();
 }
 
 bool CBattleInfoCallback::battleCanShoot(const battle::Unit * attacker, BattleHex dest) const
@@ -610,12 +607,11 @@ bool CBattleInfoCallback::battleCanShoot(const battle::Unit * attacker, BattleHe
 	if(attacker->creatureIndex() == CreatureID::CATAPULT && defender) //catapult cannot attack creatures
 		return false;
 
-	if(attacker->canShoot()
+	return attacker->canShoot()
 		&& battleMatchOwner(attacker, defender)
 		&& defender->alive()
-		&& (!battleIsUnitBlocked(attacker) || attacker->hasBonusOfType(Bonus::FREE_SHOOTING)))
-		return true;
-	return false;
+		&& (!battleIsUnitBlocked(attacker)
+		|| attacker->hasBonusOfType(Bonus::FREE_SHOOTING));
 }
 
 TDmgRange CBattleInfoCallback::calculateDmgRange(const BattleAttackInfo & info) const
--- a/lib/spells/TargetCondition.cpp
+++ b/lib/spells/TargetCondition.cpp
@@ -117,9 +117,9 @@ protected:
 
 		TBonusListPtr levelImmunities = target->getBonuses(Selector::type(Bonus::LEVEL_SPELL_IMMUNITY).And(Selector::info(1)), cachingStr.str());
 		
-		return !(levelImmunities->size() > 0
-		&& levelImmunities->totalValue() >= m->getSpellLevel()
-		&& m->getSpellLevel() > 0);
+		return levelImmunities->size() == 0 ||
+		levelImmunities->totalValue() < m->getSpellLevel() ||
+		m->getSpellLevel() <= 0;
 	}
 };
 
@@ -191,9 +191,9 @@ protected:
 	bool check(const Mechanics * m, const battle::Unit * target) const override
 	{
 		TBonusListPtr levelImmunities = target->getBonuses(Selector::type(Bonus::LEVEL_SPELL_IMMUNITY));
-		return !(levelImmunities->size() > 0 &&
-		levelImmunities->totalValue() >= m->getSpellLevel() &&
-		m->getSpellLevel() > 0);
+		return levelImmunities->size() == 0 ||
+		levelImmunities->totalValue() < m->getSpellLevel() ||
+		m->getSpellLevel() <= 0;
 	}
 };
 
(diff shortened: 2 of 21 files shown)


lib/spells/TargetCondition.cpp
```diff
@@ -57,11 +57,7 @@ class TargetConditionItemBase : public TargetConditionItem
 	bool isReceptive(const Mechanics * m, const battle::Unit * target) const override
 	{
 		bool result = check(m, target);
-
-		if(inverted)
-			return !result;
-		else
-			return result;
+		return inverted == !result;
```
dydzio0614: IMO code gets less readable in cases such as this one

nullkiller: Agreed. inverted probably means that result should be inverted. This case inverted ^ result should work?

FeniksFire: Changed for better readebility `return inverted != result; `.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -610,12 +607,11 @@ bool CBattleInfoCallback::battleCanShoot(const battle::Unit * attacker, BattleHe
...
 
-	if(attacker->canShoot()
+	return attacker->canShoot()
 		&& battleMatchOwner(attacker, defender)
 		&& defender->alive()
-		&& (!battleIsUnitBlocked(attacker) || attacker->hasBonusOfType(Bonus::FREE_SHOOTING)))
-		return true;
-	return false;
+		&& (!battleIsUnitBlocked(attacker)
+		|| attacker->hasBonusOfType(Bonus::FREE_SHOOTING));
```
ArseniyShestakov: Let's say honest I'd rather to have more lines here instead of this kind of stuff.
It's not like your change really improved this code readability.

lib/spells/TargetCondition.cpp
```diff
@@ -120,10 +116,10 @@ class AbsoluteLevelCondition : public TargetConditionItemBase
...
 
 		TBonusListPtr levelImmunities = target->getBonuses(Selector::type(Bonus::LEVEL_SPELL_IMMUNITY).And(Selector::info(1)), cachingStr.str());
-
-		if(levelImmunities->size() > 0 && levelImmunities->totalValue() >= m->getSpellLevel() && m->getSpellLevel() > 0)
-			return false;
-		return true;
+		
+		return !(levelImmunities->size() > 0
+		&& levelImmunities->totalValue() >= m->getSpellLevel()
+		&& m->getSpellLevel() > 0);
```
ArseniyShestakov: And again IMHO usage of ```!``` is confusing in situations like this one.

lib/spells/effects/Dispel.cpp
```diff
@@ -102,9 +102,7 @@ bool Dispel::mainSelector(const Bonus * bonus)
 		if(sourceSpell->id == SpellID::CLONE)
 			return false;
 		//stack may have inherited effects
-		if(sourceSpell->isAdventureSpell())
-			return false;
-		return true;
+		return !sourceSpell->isAdventureSpell();
```
ArseniyShestakov: Use ```CSpell::isCombatSpell```.

lib/battle/CBattleInfoCallback.cpp
FeniksFire: Let's say fair.
When you have such code
`if(statement) return true;
else return false;
`
it's nonsense keeping it seperate.  Better write `return statement;`
Such things should make code more simple to read, and if they don't, then something is wrong with code itself. It's good way to show that something need refactor and it's not expressive enought.

lib/spells/effects/Dispel.cpp
FeniksFire: Yep good catch, like I write above such simplification it's good way to make code more expressive.

lib/spells/TargetCondition.cpp
FeniksFire: Agreed with 100%

FeniksFire: The problem is that already something like that is in `NormalLevelCondition`, so it's code dublication and I don't want want to change one without the other. Will stick for now to be certain in this place.

lib/battle/CBattleInfoCallback.cpp
ArseniyShestakov: It's just matter of taste, but for me this is more readable:
```
if(statement1)
    return false;
if(statement2)
    return false;
if(statement3)
    return false;
return true;
```
I just generally dont care about vertical space used, but I certainly not going to pretend like it's really any importance for me. As long as it's are done right (e.g with fixes you already applied) I have nothing against changes like that.

Though I wish to know what @alexvins think since there changes in his code. He can merge it if it's fine by him.

FeniksFire: Hmm, as for me it's also not so important, I just looked around the code and changed it by the way :).

FeniksFire: Though it's always funny to think about the consequences of such a "small" change.

alexvins: I`m OK with this particular change, but I will oppose merging statements above to this condition.

lib/spells/TargetCondition.cpp
```diff
@@ -120,10 +116,10 @@ class AbsoluteLevelCondition : public TargetConditionItemBase
 		cachingStr << "type_" << Bonus::LEVEL_SPELL_IMMUNITY << "addInfo_1";
 
 		TBonusListPtr levelImmunities = target->getBonuses(Selector::type(Bonus::LEVEL_SPELL_IMMUNITY).And(Selector::info(1)), cachingStr.str());
-
-		if(levelImmunities->size() > 0 && levelImmunities->totalValue() >= m->getSpellLevel() && m->getSpellLevel() > 0)
-			return false;
-		return true;
+
```
alexvins: excess tabs on empty line

lib/spells/effects/Sacrifice.cpp
```diff
@@ -110,11 +110,8 @@ bool Sacrifice::applicable(Problem & problem, const Mechanics * m, const EffectT
 		auto victim = target.at(1).unitValue;
 		if(!victim)
 			return false;
-
-		if(!victim->alive() || !getStackFilter(m, false, victim) || !isReceptive(m, victim))
-			return false;
-
-		return true;
+		
+		return !(!victim->alive() || !getStackFilter(m, false, victim) || !isReceptive(m, victim));
```
alexvins: Too many neagtions

lib/battle/CBattleInfoCallback.cpp
FeniksFire: Yep, agree that simplify to many is bad.

PR #571, opened 2019-03-29
Victory crash/freeze fix

dydzio0614:
This PR fixes bugs that cause crashes/freezes on scenario/campaign map win. Default handling for lost connection changed to avoid crash (do not try to close server). Changing server state to GAMEPLAY_ENDED is properly handled, it left server in unstable state before, causing unavoidable assert failure. And last, thread synchronization has been made between closing server and restarting it for next campaign mission. I guess it can be implemented to work for every server close/start case in the code, so client cannot join exiting server, but I did not dig into it.

Tl;dr: Map finish problems fixed, campaigns work again.

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -1265,6 +1265,7 @@ static void handleEvent(SDL_Event & ev)
 			break;
 		case EUserEvent::CAMPAIGN_START_SCENARIO:
 			{
+				CSH->campaignServerRestartLock.set(true);
 				CSH->endGameplay();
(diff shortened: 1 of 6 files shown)


nullkiller:
There is visit town problem at least for me. When game ends by capturing last AI town. It looks different but I quite often used to end campaign games such way.

ArseniyShestakov:
There are also at least two crashes in MP:
- When one of clients wins, everyone else crashes.
- Seems like server crashes currently when just one player lose or might be when he's disconnected...
Yet I can merge this one now since there certainly no regressions.

PR #572, opened 2019-03-30
Fix crash caused by FuzzyLite destructor

dydzio0614:
Crash was happening on map restart request after AI passes its turn. Initiated by AIPathfinder::init() - clearing of vectors holding AINodeStorage variable. AINodeStorage contains std::unique_ptr<FuzzyHelper> dangerEvaluator, which couldn't be cleared properly, and the top frames of call stack pointed to FuzzyLite's destructors. I think I solved problem, the solution I submit is based on analyzing Engine class destructor - https://github.com/fuzzylite/fuzzylite/blob/release/fuzzylite/src/Engine.cpp#L96 (I used FuzzyLite without debug info). Playtesting shows me that there is no ingame crash anymore after this change.

I did not perform memory leak testing though to confirm if the fix is correct.

### DIFF
--- a/AI/VCAI/FuzzyEngines.cpp
+++ b/AI/VCAI/FuzzyEngines.cpp
@@ -23,7 +23,8 @@ extern FuzzyHelper * fh;
 
 engineBase::engineBase()
 {
-	engine.addRuleBlock(&rules);
+	rules = new fl::RuleBlock();
+	engine.addRuleBlock(rules);
 }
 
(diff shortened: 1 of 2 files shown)


nullkiller:
I also encountered such issue and the fix was close to this. FuzzyLite destroys everything it knows about. Btw memory leak can be tested by a small console endlessly creating danger evaluator.

PR #573, opened 2019-04-07
Nullkiller AI

nullkiller:
This is a long term work in-progress pull request for my wandering around AI. Pull request created by request from @dydzio0614 
The goal is to develop an alternative to VCAI. Just because I like doing such things. Doing them my way. I appreciate constructive discussions about what AI should do and why.

The architecture is still evaluating. For now it is implemented as a plugin to VCAI fully overriding AI behavior for the blue AI. In general VCAI will be used as a gateway. There is an engine which operates behaviors and analysers. The analysers (or maybe managers) extract data from gamestate in a form more suitable for decission making. The behaviors correspond to a certain possibility in the game and generate tasks (elementar goals for now). They can use decomposition but only for some specific activities like quest completion. For now there is no decomposition at all. Later I plan to extract things like moving hero to routines.

AIPathfinder is extended to allow basic hero chain.
Own prioritization engine.

Plans:
- Hero manager - splitting heroes onto categories primary/scout. Estimating primary and secondary skills.
- Army manager - concept of power stack
- Hero chain
-- Use heroes from tavern
-- Allow upgrading army in town
- Improve prioritization. Priority changes in time.
- town development
- exploration
-- vector oriented exploration
-- view earth/view air
- postoptimization
- parallel processing
- further evaluation and improvements

For now AI requires open map and fully developed town. It is not able to compete quests or analyse win conditions.


AI/VCAI/Engine/PriorityEvaluator.cpp
```diff
@@ -0,0 +1,354 @@
...
+		addRule("if armyReward is HIGH and turnDistance is SMALL and armyLoss is LOW then Value is HIGHEST");
+		addRule("if goldReward is HIGH and turnDistance is SMALL and armyLoss is LOW then Value is HIGHEST");
+		addRule("if armyReward is MEDIUM and turnDistance is SMALL and armyLoss is LOW then Value is HIGH");
+		addRule("if goldReward is MEDIUM and turnDistance is SMALL and armyLoss is LOW then Value is BITHIGH");
+		addRule("if goldReward is LOW and armyReward is LOW and turnDistance is not SMALL then Value is LOWEST");
+		addRule("if armyLoss is HIGH then Value is LOWEST");
+		addRule("if armyLoss is LOW then Value is MEDIUM");
+		addRule("if armyReward is LOW and turnDistance is LONG then Value is LOWEST");
+		addRule("if danger is NONE and armyStrength is HIGH and armyReward is LOW then Value is LOW");
+		addRule("if danger is NONE and armyStrength is HIGH and armyReward is MEDIUM then Value is BITLOW");
```
dydzio0614: Other fuzzy logic terms for army Strength are unused in rules. That makes output imbalanced as engine does not react at all when armyStrength is at range 0.2 - 0.5. If you do not want to make counter-rules that impact output for LOW / MEDIUM then consider removing rest of terms, then giving HIGH whole range (0-1) and change term shape.

dydzio0614:
Here are some "what VCAI should do but does not" to consider when trying to reduce some AI flaws:
1. Gathering army goal - it's priority should not ignore what parent goal is when deciding what to do
2. Defense - defend towns, do not run near stronger enemy heroes with our heroes without good reason, retreat with our strong heroes that may be eliminated with even stronger heroes.
3. Do not ignore portals for so long when doing exploration.
4. Do not visit trash objects too often (morale shrines etc. - these are good to visit on the way to somewhere else)

AI/VCAI/Engine/PriorityEvaluator.cpp
nullkiller: Looks like it will evaluate in heroType MAIN/SECONDARY/SCOUT. Secondary is under the question. Maybe it will by subtype of main. Scouts will have different rules priority evaluating.

nullkiller:
@dydzio0614 thank you for your feedback. Goals will evaluate to tasks and they will lose all the logic. The behaviors will replace goals but without decomposition as a main tool. They will be more independent. All the shared logic will be moved to analysers/managers and will be represented in form of some precalculated data or direct queries to the gamestate. All the rest is quite applicable to my AI on later stages.
Also one more important concept is various post optimization of task chains. For instance if a hero is engaged in hero chain and has to wait next turn before completing it we can check if he can visit something on the way and still be at waitpoint in time.

AI/VCAI/AIhelper.h
```diff
@@ -17,14 +17,15 @@
...
+#include "HeroManager.h"
 #include "Pathfinding/PathfindingManager.h"
 
 class ResourceManager;
 class BuildingManager;
 
 
 //indirection interface for various modules
-class DLL_EXPORT AIhelper : public IResourceManager, public IBuildingManager, public IPathfindingManager, public IArmyManager
+class DLL_EXPORT AIhelper : public IResourceManager, public IBuildingManager, public IPathfindingManager, public IArmyManager, public IHeroManager
```
alexvins: We should do smth with this, `AIhelper` started inheriting everything. May be just change to public getters

AI/VCAI/Behaviors/RecruitHeroBehavior.cpp
```diff
@@ -33,13 +34,165 @@ Goals::TGoalVec RecruitHeroBehavior::getTasks()
...
+
+	auto heroToCheck = startupTown->garrisonHero ? startupTown->garrisonHero.get() : startupTown->visitingHero.get();
+	auto paths = cb->getPathsInfo(heroToCheck);
+
+	for(auto obj : ai->visitableObjs)
+	{
+		if(obj->ID == Obj::RESOURCE && obj->subID == Res::GOLD
+			|| obj->ID == Obj::TREASURE_CHEST
+			|| obj->ID == Obj::CAMPFIRE
+			|| obj->ID == Obj::WATER_WHEEL)
```
alexvins: This should work for any `CRewardableObject` descendant

AI/VCAI/Engine/PriorityEvaluator.cpp
```diff
@@ -35,131 +37,28 @@ extern boost::thread_specific_ptr<VCAI> ai;
...
-		addRule("if armyLoss is LOW then Value is MEDIUM");
-		addRule("if armyReward is LOW and turnDistance is LONG then Value is LOWEST");
-		addRule("if danger is NONE and armyStrength is HIGH and armyReward is LOW then Value is LOW");
-		addRule("if danger is NONE and armyStrength is HIGH and armyReward is MEDIUM then Value is BITLOW");
-	}
-	catch(fl::Exception & fe)
-	{
-		logAi->error("visitTile: %s", fe.getWhat());
-	}
+	auto file = CResourceHandler::get("initial")->load(ResourceID("config/ai-priorities.txt"))->readAll();
```
alexvins: Is it intended that by using `initial` filesystem ai-priorities is forbidden to be altered by mods?

AI/VCAI/Behaviors/RecruitHeroBehavior.cpp
nullkiller: It is interested in gold only. Maybe if I will be able to extract reward from there I can generalize it

AI/VCAI/AIhelper.h
nullkiller: As soon as I am able to build reliable logic on top of VCAI I will start cutting old architecture. I am not 100% sure but I expect to get rid of this class and use managers directly. Finally I may end up having a separate AI project so that we are able to select AI in game launcher. Now it is more important to build reliable logic which can play average map successfully.

AI/VCAI/Engine/PriorityEvaluator.cpp
nullkiller: I know nothing about modding. First of all I need this file to be copied correctly during build. It is not the case now. If you can guide me how to use vcmi filesystem I will be happy. I expect it has to be added to some cmake? Also I need new resource type *.fll (fuzzy light language)

AI/VCAI/Behaviors/RecruitHeroBehavior.cpp
alexvins: You can extract everything from reward configuration. 
`CRewardableObject` has `std::vector<CVisitInfo> info;` all possible rewards. Then dig into `CVisitInfo`.

AI/VCAI/AIhelper.h
dydzio0614: It's discussable whether it's better to have separate AI or "two AI's in one" with on/off toggle. Both approaches have their cons and pros. Cons for separate AI would be VCAI not benefitting directly from code improvements common for both AI's and your AI not benefitting from VCAI improvements made by other people.

AI/VCAI/Engine/PriorityEvaluator.cpp
alexvins: * config folder being copied completely, your file it should be handled too.
* you should add `fll` extension to `EResType` or use `*.txt` - you feed a string to importer anyway.
* for now you could just rename to txt and load with
```
CResourceHandler::get()->load(ResourceID("config/ai-priorities.txt", EResType::TEXT))->readAll();
```

nullkiller: Fll is automatically opened by fuzzy light client. I can use txt so. Is the place for file correct? Can I put it to a nee folder "ai"?

AI/VCAI/AIhelper.h
nullkiller: Agreed. But we can move towards some shared things. Or I can forever develop in my own branch and pack its result as mod. This way we can exchange commits using cherry pick or something. Anyway for now I do not plan to move to a separate assembly. It is lowest priority task.

AI/VCAI/Engine/PriorityEvaluator.cpp
alexvins: 'config/ai' is a good place

nullkiller: Ho to access it in resource manager? Same config/priorities.txt or config/ai/priorities.txt?

alexvins: config/ai/priorities.txt

AI/VCAI/Behaviors/RecruitHeroBehavior.cpp
nullkiller: It is protected. I can not access it without extending CRewardableObject. Also I have already extended CBankInfo which itself might be unacceptable

alexvins: Just make `grantReward` & `getVisitInfo` public.

nullkiller:
Branch was renamed and pushed to vcmi git under new name nullkiller-ai

PR #575, opened 2019-04-14
AI: implement hero skills evaluator

nullkiller:
Helps in selecting skills for AI heroes

### DIFF
--- a/AI/VCAI/HeroManager.cpp
+++ b/AI/VCAI/HeroManager.cpp
@@ -1,0 +1,237 @@
+/*
+* HeroManager.cpp, part of VCMI engine
+*
+* Authors: listed in file AUTHORS in main folder
+*
+* License: GNU General Public License v2.0 or later
+* Full text of license available in license.txt file, in main folder
+*
+*/
+
+#include "StdInc.h"
+#include "HeroManager.h"
+
+#include "../../CCallback.h"
+#include "../../lib/mapObjects/MapObjects.h"
+#include "../../lib/CHeroHandler.h"
+
+SecondarySkillEvaluator HeroManager::wariorSkillsScores = SecondarySkillEvaluator(
+	{
+		std::make_shared<SecondarySkillScoreMap>(
+			std::map<SecondarySkill, float>
+			{
+				{SecondarySkill::DIPLOMACY, 2},
+				{SecondarySkill::LOGISTICS, 2},
+				{SecondarySkill::EARTH_MAGIC, 2},
+				{SecondarySkill::ARMORER, 2},
+				{SecondarySkill::OFFENCE, 2},
+				{SecondarySkill::AIR_MAGIC, 1},
+				{SecondarySkill::WISDOM, 1},
+				{SecondarySkill::LEADERSHIP, 1},
+				{SecondarySkill::INTELLIGENCE, 1},
+				{SecondarySkill::RESISTANCE, 1},
+				{SecondarySkill::MYSTICISM, -1},
+				{SecondarySkill::SORCERY, -1},
+				{SecondarySkill::ESTATES, -1},
+				{SecondarySkill::FIRST_AID, -1},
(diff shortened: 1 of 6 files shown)


AI/VCAI/HeroManager.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+void HeroManager::setAI(VCAI * AI)
+{
+	ai = AI;
+}
+
+float HeroManager::evaluateSpeciality(const CGHeroInstance * hero) const
+{
+	auto heroSpecial = Selector::source(Bonus::HERO_SPECIAL, hero->type->ID.getNum());
```
nullkiller: Only speciality for secondary skills like Armorer is evaluated here

nullkiller:
Going to merge it to Nullkiller AI because I need it for further AI development

AI/VCAI/HeroManager.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+	return specialityScore;
+}
+
+float HeroManager::evaluateFightingStrength(const CGHeroInstance * hero) const
+{
+	return evaluateSpeciality(hero) + wariorSkillsScores.evaluateSecSkills(hero) + hero->level * 1.5f;
+}
+
+std::map<HeroPtr, HeroRole> HeroManager::getHeroRoles() const
```
dydzio0614: This logic should be unified with VCAI::primaryHero()

nullkiller: Maybe. But this is a very big and complex question which I am not much interested in.

nullkiller: The main difference between primary hero and this logic is that army value is not taken into account because the army can be given to any hero making it primary which does not look correct.

AI/VCAI/HeroManager.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+}
+
+void SecondarySkillScoreMap::evaluateScore(const CGHeroInstance * hero, SecondarySkill skill, float & score) const
+{
+	auto it = scoreMap.find(skill);
+
+	if(it != scoreMap.end())
+	{
+		score = it->second;
```
ArseniyShestakov: You can use our ```vstd``` function in situations like this:
```
	if(vstd::contains(scoreMap, skill))
	{
		score = scoreMap[skill];
```

alexvins: Using iterator in this case is faster, `vstd::contains` should be used if you just check presence and nothing more.

AI/VCAI/HeroManager.h
```diff
@@ -0,0 +1,107 @@
...
+#include "VCAI.h"
+
+enum HeroRole
+{
+	MAIN,
+
+	SCOUT
+};
+
+class DLL_EXPORT IHeroManager //: public: IAbstractManager
```
alexvins: If you export something you should import it somewhere, in case of AI you export classes only to be used in tests.

nullkiller: Yeah, one day I may decide to write some tests. Now this can be removed.

alexvins: But it can not be just removed, `IHeroManager` is a base class of AIHelper that is already exported.

AI/VCAI/HeroManager.cpp
nullkiller: I think I will fix it soon

ArseniyShestakov:
Ok I wanted to make sure it's actually work with VCAI currently so I want to fix "vcmiglorfindel" cheat so it's work not just during AI turn. Then I'll merge this one.

AI/VCAI/HeroManager.cpp
nullkiller: I thought it has to be fixed but it turned out that it was already correct. To avoid double lookup in map I used iterator. So this can be merged.

PR #576, opened 2019-04-15
Add missing header

grantisu:
This fixes the error I got while trying to build from source:
```
client/widgets/AdventureMapClasses.cpp:195:28: error: ‘SDLX_Size’ was not declared in this scope
```

### DIFF
--- a/client/widgets/AdventureMapClasses.cpp
+++ b/client/widgets/AdventureMapClasses.cpp
@@ -23,6 +23,7 @@
 #include "../CMessage.h"
 
 #include "../gui/CGuiHandler.h"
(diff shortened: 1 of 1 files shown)


dydzio0614:
Seems you did something wrong, when building VCMI from source then VCMI client should have client/StdInc.h file included as precompiled header, that file includes gui/SDL_Compat.h for all files.

grantisu:
> Seems you did something wrong

I probably had some stale intermediate files hanging around; starting from an empty build directory works with every combination of cmake options I can find in my bash history. Sorry for the PR noise. ¯\\\_(ツ)\_/¯

PR #581, opened 2019-05-02
Fixed Crypt/Pyramid double visit sound.

Ewilhan:
Removed sound call from CBank::onHeroVisit() to avoid two sounds being launched at same time. 
Gave priority to JSON files, as I suppose this is the prefered way of setting parameters, rather than hard code.


dydzio0614:
Works well for me.

PR #583, opened 2019-05-03
Fix no player interface swap on being attacked

dydzio0614:
Fix for bug https://bugs.vcmi.eu/view.php?id=3053

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -1607,6 +1607,29 @@ void CPlayerInterface::objectRemoved(const CGObjectInstance * obj)
 	}
 }
 
+void CPlayerInterface::playerBlocked(int reason, bool start)
+{
+	if(reason == PlayerBlocked::EReason::UPCOMING_BATTLE)
+	{
+		if(CSH->howManyPlayerInterfaces() > 1 && LOCPLINT != this && LOCPLINT->makingTurn == false) 
+		{ 
+			//one of our players who isn't last in order got attacked not by our another player (happens for example in hotseat mode)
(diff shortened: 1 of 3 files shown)


ArseniyShestakov:
BTW there still two bugs remaining:

* Notifications for player still not arrive. E.g if you stand on cursed ground and cant cast spells there is no notifications.
* When battle starts control over units not given to player, they not highlighted until you click somewhere on battlefield.

I'm fine with merging this since it's improve it greatly already (you can actually control units and cast spells!), but you might want to fix those too.

client/CPlayerInterface.cpp
```diff
@@ -1607,6 +1607,29 @@ void CPlayerInterface::objectRemoved(const CGObjectInstance * obj)
...
+			GH.curInt = this;
+			adventureInt->selection = nullptr;
+			adventureInt->setPlayer(playerID);
+			std::string msg = CGI->generaltexth->localizedTexts["adventureMap"]["playerAttacked"].String();
+			boost::replace_first(msg, "%s", cb->getStartInfo()->playerInfos.find(playerID)->second.name);
+			std::vector<std::shared_ptr<CComponent>> cmp;
+			cmp.push_back(std::make_shared<CComponent>(CComponent::flag, playerID.getNum(), 0));
+			makingTurn = true; //workaround for stiff showInfoDialog implementation
+			showInfoDialog(msg, cmp);
+			makingTurn = false;
```
ArseniyShestakov: With those 3 lines you bypassed problem for this exact particular case, but all other notifications during battle still hidden from attacked player.

ArseniyShestakov:
Ok dydzio have asked this to be merged and it does make sense since it's actually make battle playable.
Added separate issue to BT for other issues:
https://bugs.vcmi.eu/view.php?id=3054

PR #586, opened 2019-05-04
Ewilhan fix neutral creatures growth week

Ewilhan:
Changed the RNG so that Neutral Creatures cannot have Growth Weeks.
Made it so a new monster is generated until a non-neutral monster is chosen.
Also, added a `isNeutral()` function to `CCreature`, in line with already existing `isGood()` and `isEvil()`.

Fixes [#3046](https://bugs.vcmi.eu/view.php?id=3046).

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -1673,8 +1673,11 @@ void CGameHandler::newTurn()
 				if (monthType < 25)
 				{
 					n.specialWeek = NewTurn::BONUS_GROWTH; //+5
-					std::pair<int, CreatureID> newMonster(54, VLC->creh->pickRandomMonster(getRandomGenerator()));
-					//TODO do not pick neutrals
+					std::pair<int, CreatureID> newMonster(54, 0);
+					do
+					{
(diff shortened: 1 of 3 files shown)


server/CGameHandler.cpp
```diff
@@ -1673,8 +1673,11 @@ void CGameHandler::newTurn()
 				if (monthType < 25)
 				{
 					n.specialWeek = NewTurn::BONUS_GROWTH; //+5
-					std::pair<int, CreatureID> newMonster(54, VLC->creh->pickRandomMonster(getRandomGenerator()));
-					//TODO do not pick neutrals
+					std::pair<int, CreatureID> newMonster(54, 0);
```
alexvins:
```suggestion
					std::pair<int, CreatureID> newMonster(54, CreatureID());
```

dydzio0614: I am not sure if this is proper fix for this issue. That's because neutral faction alignment does not have to mean creatures without town. Creatures from conflux, stronghold and fortress towns have neutral alignment, so for example heroes from these factions can sacrifice both creatures and artifacts for experience etc.

PR #591, opened 2019-05-18
Spawn ship placed outside map if it has visitable position inside map

dydzio0614:
Fixes bug #3060, decided to go for max restrictive conditions so logic for other objects becomes unchanged. If we allow spawning objects via scripting then we'll need to revisit this code.

### DIFF
--- a/lib/NetPacksLib.cpp
+++ b/lib/NetPacksLib.cpp
@@ -676,8 +676,22 @@ DLL_LINKAGE void GiveHero::applyGs(CGameState *gs)
 
 DLL_LINKAGE void NewObject::applyGs(CGameState *gs)
 {
-	const TerrainTile &t = gs->map->getTile(pos);
-	ETerrainType terrainType = t.terType;
+	ETerrainType terrainType;
+
+	if(ID == Obj::BOAT && !gs->isInTheMap(pos)) //special handling for bug #3060 - pos outside map but visitablePos is not
+	{
+		CGObjectInstance testObject = CGObjectInstance();
+		testObject.pos = pos;
+		testObject.appearance = VLC->objtypeh->getHandlerFor(ID, subID)->getTemplates(ETerrainType::WATER).front();
+
+		const int3 previousXAxisTile = int3(pos.x - 1, pos.y, pos.z);
+		assert(gs->isInTheMap(previousXAxisTile) && (testObject.visitablePos() == previousXAxisTile)); 
+	}
+	else
+	{
+		const TerrainTile & t = gs->map->getTile(pos);
+		terrainType = t.terType;
+	}
 
 	CGObjectInstance *o = nullptr;
 	switch(ID)


lib/NetPacksLib.cpp
```diff
@@ -676,8 +676,24 @@ DLL_LINKAGE void GiveHero::applyGs(CGameState *gs)
...
+	ETerrainType terrainType;
+
+	if(ID == Obj::BOAT && !gs->isInTheMap(pos)) //special handling for bug #3060 - pos outside map but visitablePos is not
+	{
+		CGObjectInstance testObject = CGObjectInstance();
+		testObject.pos = pos;
+		testObject.appearance = VLC->objtypeh->getHandlerFor(ID, subID)->getTemplates(ETerrainType::WATER).front();
+
+		const int3 previousXAxisTile = int3(pos.x - 1, pos.y, pos.z);
+		bool test1 = gs->isInTheMap(previousXAxisTile);
```
nullkiller: Why do we need variables test1 and test2. Any chance to have more meaningful names?

lib/NetPacksLib.cpp
```diff
@@ -676,8 +676,24 @@ DLL_LINKAGE void GiveHero::applyGs(CGameState *gs)
 
 DLL_LINKAGE void NewObject::applyGs(CGameState *gs)
 {
-	const TerrainTile &t = gs->map->getTile(pos);
-	ETerrainType terrainType = t.terType;
+	ETerrainType terrainType;
+
+	if(ID == Obj::BOAT && !gs->isInTheMap(pos)) //special handling for bug #3060 - pos outside map but visitablePos is not
+	{
+		CGObjectInstance testObject = CGObjectInstance();
```
nullkiller: Consider name testBoat.

lib/NetPacksLib.cpp
```diff
@@ -676,8 +676,24 @@ DLL_LINKAGE void GiveHero::applyGs(CGameState *gs)
...
+	if(ID == Obj::BOAT && !gs->isInTheMap(pos)) //special handling for bug #3060 - pos outside map but visitablePos is not
+	{
+		CGObjectInstance testObject = CGObjectInstance();
+		testObject.pos = pos;
+		testObject.appearance = VLC->objtypeh->getHandlerFor(ID, subID)->getTemplates(ETerrainType::WATER).front();
+
+		const int3 previousXAxisTile = int3(pos.x - 1, pos.y, pos.z);
+		bool test1 = gs->isInTheMap(previousXAxisTile);
+		int3 test2 = testObject.visitablePos();
+		assert(gs->isInTheMap(previousXAxisTile) && (testObject.visitablePos() == previousXAxisTile));
```
nullkiller: What will be the value of terraintype at the end? As far as I understood getTile was invoked to get this variable

lib/NetPacksLib.cpp
dydzio0614: oh crap, I wanted to remove them, will fix soon

lib/NetPacksLib.cpp
nullkiller: Or maybe testObject is fine here.

lib/NetPacksLib.cpp
dydzio0614: few lines below:
```
switch(ID)
	{
	case Obj::BOAT:
		o = new CGBoat();
		terrainType = ETerrainType::WATER;
...
```

lib/NetPacksLib.cpp
nullkiller: Btw are you sure it is not an issue with position conversion to h3m format? I hardly understand what I am talking about.

dydzio0614: I am sure. The bug appears during boat purchase when shipyard is 1 tile away from map end, and there is water at the border of the map.

lib/NetPacksLib.cpp
nullkiller: Technically you say here assert(gs->isInTheMap(testObject.visitablePos())) which looks more logical (objects visitable pos should be in the map) and does not need mystic x - 1.

dydzio0614: Somewhat true, I made stricter check which ensures visitable position is one tile away to make it more logical for this particular issue.

PR #597, opened 2019-05-26
Ensure no freeze on AI interruption

dydzio0614:
1. Force linking boost thread dynamically to avoid https://stackoverflow.com/questions/35978572/boost-thread-interupt-does-not-work-when-crossing-a-dll-boundary

2. Fix theoretical possible freeze on two threads trying to join AI makingTurn thread.

### DIFF
--- a/AI/VCAI/VCAI.cpp
+++ b/AI/VCAI/VCAI.cpp
@@ -2447,6 +2447,8 @@ void VCAI::recruitHero(const CGTownInstance * t, bool throwing)
 
 void VCAI::finish()
 {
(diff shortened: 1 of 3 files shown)


alexvins:
Should we change smth in build system as long as static linking is not supported?

dydzio0614:
Ask @ArseniyShestakov - currently it looks like both travis and appveyor link boost dynamically, unsure about android build.

dydzio0614:
Android build also uses dynamic linking as far as I can see.

alexvins:
OK lets try this

PR #598, opened 2019-05-28
AI: blind fix for 3062

nullkiller:
Blind fix. According to the logs mechanic of the bug is the next
GatherTroops checks towns for required creature
GatherTroops selects a hero to pick creatures
Hero goes to required target for instance town
Town->Hero exchange logic refuses to pick required creatures because it knows nothing about the quest and the stack is weak and there is no free slot
Infinite loop

### DIFF
--- a/AI/VCAI/Goals/GatherTroops.cpp
+++ b/AI/VCAI/Goals/GatherTroops.cpp
@@ -138,7 +138,8 @@ TGoalVec GatherTroops::getAllPossibleSubgoals()
 
 	CreatureID creID = CreatureID(objid);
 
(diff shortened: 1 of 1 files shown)


AI/VCAI/Goals/GatherTroops.cpp
```diff
@@ -135,6 +136,12 @@ TGoalVec GatherTroops::getAllPossibleSubgoals()
 		}
 	}
 
+	CreatureID creID = CreatureID(objid);
+
+	vstd::erase_if(solutions, [&](TSubgoal goal)->bool{
```
alexvins:
```suggestion
	vstd::erase_if(solutions, [&](TSubgoal goal)->bool
	{
```

PR #600, opened 2019-06-01
Remove useless and wrong recruit conditions

dydzio0614:
Fixes ability to recruit more than 8 heroes etc.


ArseniyShestakov:
There is check on server-side for this. If it's doesn't work you need to fix it as well.

PR #602, opened 2019-06-04
Do not build pseudo-buildings explicitly with events

dydzio0614:
Fixes https://bugs.vcmi.eu/view.php?id=3064 and https://bugs.vcmi.eu/view.php?id=2561

### DIFF
--- a/lib/mapping/MapFormatH3M.cpp
+++ b/lib/mapping/MapFormatH3M.cpp
@@ -2104,19 +2104,19 @@ std::set<BuildingID> CMapLoaderH3M::convertBuildings(const std::set<BuildingID> 
 	{
 		//village hall is always present
 		ret.insert(BuildingID::VILLAGE_HALL);
-	}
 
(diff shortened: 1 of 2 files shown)


lib/mapping/MapFormatH3M.cpp
```diff
@@ -2104,19 +2104,19 @@ std::set<BuildingID> CMapLoaderH3M::convertBuildings(const std::set<BuildingID>
...
-		ret.insert(BuildingID::EXTRA_CITY_HALL);
-	}
-	if(ret.find(BuildingID::TOWN_HALL) != ret.end())
-	{
-		ret.insert(BuildingID::EXTRA_TOWN_HALL);
-	}
-	if(ret.find(BuildingID::CAPITOL) != ret.end())
-	{
-		ret.insert(BuildingID::EXTRA_CAPITOL);
+		if(ret.find(BuildingID::CITY_HALL) != ret.end())
```
DjWarmonger: use vstd::contains instead?

PR #605, opened 2019-06-08
Fix all common victory/defeat crashes

dydzio0614:
To be tested on MXE build before merge


dydzio0614:
Tested, works as intended.

PR #608, opened 2019-06-24
Break Visual Studio 2015 support

dydzio0614:
There are various reasons to stop trying to keep Visual Studio 2015 compatibility currently. One of them is uncompilable and "hard to fix" code in incoming scripting branch. This branch swaps appveyor 32 bit build compiler to VS 2017 for relevant use cases.

After this branch gets merged there are two things to do:
1. Update wiki to match removing VS 2015 support in "how to build" instructions
2. Consider upgrading allowed C++ standard, currently we support C++11, though without explicit mentioning of some C++11 features that are still broken in some toolsets such as VS 2015. I created some notes here, which will help in deciding when extending allowed C++ features for VCMI code: https://wiki.vcmi.eu/User:Dydzio/cpp_standard_upgrade

Adding VS 2019 to appveyor is out of scope of this PR, it should be done separately later, also requires updating Qt in dependencies package since VS 2019 finds code error in currently provided Qt version.

This branch is supposed to be merged into develop before handlersAbstraction branch

### DIFF
--- a/CI/appveyor.yml
+++ b/CI/appveyor.yml
@@ -74,9 +74,9 @@ for:
       - coverity_scan
   environment:
     matrix:
-      - NAME: Coverity - MSVS 2015 x86 - Release
-        APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2015
-        VCMI_GENERATOR: Visual Studio 14 2015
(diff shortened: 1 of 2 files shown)


dydzio0614:
Apparently qt needs to be updated also for VS 2017 32 bit...

PR #612, opened 2019-07-13
Object template rework

DjWarmonger:
Work in progress:

- [x] Move from vector of vectors to boost::multi_array
- [x] Unify interface, clean clean up magic numbers
- [x] Precalculate offsets for better performance
- [ ] Backward save compatibility
- [ ] Unit tests for regular and edge cases

### DIFF
--- a/lib/serializer/BinarySerializer.h
+++ b/lib/serializer/BinarySerializer.h
@@ -344,6 +344,30 @@ public:
 			save((ui8)0);
 		}
 	}
+
+	template <typename T>
+	void save(const boost::multi_array<T, 2> & data)
+	{
+		ui32 length = data.num_elements();
+		*this& length;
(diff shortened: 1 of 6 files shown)


lib/serializer/BinarySerializer.h
```diff
@@ -344,6 +344,30 @@ class DLL_LINKAGE BinarySerializer : public CSaverBase
...
 	}
+
+	template <typename T>
+	void save(const boost::multi_array<T, 2> & data)
+	{
+		ui32 length = data.num_elements();
+		*this& length;
+		auto shape = data.shape();
+		ui32 x = shape[0], y = shape[1];
+		*this& x& y;
```
alexvins: Too ugly, use `save` calls

PR #616, opened 2019-09-13
[3085] - MapObjectEvaluator crashes on custom monolith

nullkiller:
MapObjectEvaluator crashes on custom monolith because
VLC->objtypeh->getHandlerFor(primaryID, secondaryID)
returns NULL
Primary ID: 45, Secondary ID: 8

### DIFF
--- a/AI/VCAI/MapObjectsEvaluator.cpp
+++ b/AI/VCAI/MapObjectsEvaluator.cpp
@@ -26,7 +26,7 @@ MapObjectsEvaluator::MapObjectsEvaluator()
 		for(auto secondaryID : VLC->objtypeh->knownSubObjects(primaryID))
 		{
 			auto handler = VLC->objtypeh->getHandlerFor(primaryID, secondaryID);
-			if(!handler->isStaticObject())
+			if(handler && !handler->isStaticObject())
 			{
 				if(handler->getAiValue() != boost::none)
 				{


AI/VCAI/MapObjectsEvaluator.cpp
```diff
@@ -26,7 +26,7 @@ MapObjectsEvaluator::MapObjectsEvaluator()
 		for(auto secondaryID : VLC->objtypeh->knownSubObjects(primaryID))
 		{
 			auto handler = VLC->objtypeh->getHandlerFor(primaryID, secondaryID);
-			if(!handler->isStaticObject())
+			if(handler && !handler->isStaticObject())
```
nullkiller: I am afraid it is completely incorrect so it avoids crash. The custom monoliths probably should have a handler associated with it.

alexvins:
```
"monolithTwoWay" : {
		"index" :45,
		"handler": "monolith",
		"base" : {
			"sounds" : {
				"ambient" : ["LOOPMON2"],
				"visit" : ["TELPTOUT"]
			}
		},
		"types" : {
			"monolith1" : { "index" : 0 },
			"monolith2" : { "index" : 1 },
			"monolith3" : { "index" : 2 },
			"monolith4" : { "index" : 3 },
			"monolith5" : { "index" : 4 },
			"monolith6" : { "index" : 5 },
			"monolith7" : { "index" : 6 },
			"monolith8" : { "index" : 7 }
		}
	},
```
For monolith "Primary ID: 45, Secondary ID: 8" mod author is responsible for adding new `type`. But i`m afraid there also a bug in RMG there monoliths are created OTF without modifying metadata.

PR #623, opened 2020-01-21
Fix bug: LobbyScreen UI has a bug on the Difficulty toggle group.

Toneyisnow:
https://bugs.vcmi.eu/view.php?id=3102

Scenario:
In LoadGame Lobby screen, the difficulty toggle group is showing multiple choices, but actually only one valid is working. 

The reason is, in the Lobby screen Initialization code will set the difficulty = 0, and each time the player changes a map in the selection, the toggle group control is not resetting the difficulty buttons to disabled state.

How fix:
Add a new method to ToggleGroup class: setSelectedOnly, which will disable all other buttons and then set the selected button.

Note:
During the game loading time, the client is loading the map, and send a NetPack to server: LobbySetMap, and send a NetPack to Interface: LobbyUpdateState.
In the LobbyUpdateState it sets the map difficulty.


Toneyisnow:
This change is having some wrong files, I will create again.

PR #624, opened 2020-01-21
Fix bug: LobbyScreen UI has a bug on the Difficulty toggle group.

Toneyisnow:
Bug: https://bugs.vcmi.eu/view.php?id=3102

Scenario: In LoadGame Lobby screen, the difficulty toggle group is showing multiple choices, but actually only one valid is working.

The reason is, in the Lobby screen Initialization code will set the difficulty = 0, and each time the player changes a map in the selection, the toggle group control is not resetting the difficulty buttons to disabled state.

How fix:
Add a new method to ToggleGroup class: setSelectedOnly, which will disable all other buttons and then set the selected button.

Note:
During the game loading time, the client is loading the map, and send a NetPack to server: LobbySetMap, and send a NetPack to Interface: LobbyUpdateState.
In the LobbyUpdateState it sets the map difficulty.

### DIFF
--- a/client/widgets/Buttons.cpp
+++ b/client/widgets/Buttons.cpp
@@ -23,8 +23,6 @@
 #include "../windows/InfoWindows.h"
 #include "../../lib/CConfigHandler.h"
 
-using namespace std;
-
 void CButton::update()
 {
 	if (overlay)
@@ -439,7 +437,7 @@ void CToggleGroup::setSelected(int id)
 
 void CToggleGroup::setSelectedOnly(int id)
 {
-	for (map<int, shared_ptr<CToggleBase> >::iterator it = buttons.begin(); it != buttons.end(); it++)
+	for(auto it = buttons.begin(); it != buttons.end(); it++)
 	{
 		int buttonId = it->first;
 		buttons[buttonId]->setEnabled(buttonId == id);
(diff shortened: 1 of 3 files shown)


client/widgets/Buttons.cpp
```diff
@@ -23,6 +23,8 @@
 #include "../windows/InfoWindows.h"
 #include "../../lib/CConfigHandler.h"
 
+using namespace std;
```
alexvins: Global `using` is not recommended.

client/widgets/Buttons.cpp
```diff
@@ -425,6 +437,17 @@ void CToggleGroup::setSelected(int id)
 	selectionChanged(id);
 }
 
+void CToggleGroup::setSelectedOnly(int id)
+{
+	for (map<int, shared_ptr<CToggleBase> >::iterator it = buttons.begin(); it != buttons.end(); it++)
```
alexvins: No space after `for` please and use `auto` in iterator declaration.

client/widgets/Buttons.cpp
ArseniyShestakov: Tbh I think we had long talks about it on Slack so I strongly against ```using``` for something as common as ```std```.

Toneyisnow: Ok, that's fine.

client/widgets/Buttons.cpp
Toneyisnow: Make sense.

PR #625, opened 2020-01-25
Implement the Clone Effect in battle.

Toneyisnow:
1. Make color shifter for CAnimation, and use that in the clone effect.
2. Update the original position of the cloned object.

### DIFF
--- a/client/battle/CBattleAnimations.cpp
+++ b/client/battle/CBattleAnimations.cpp
@@ -90,7 +90,7 @@ CBattleStackAnimation::CBattleStackAnimation(CBattleInterface * owner, const CSt
 	assert(myAnim);
 }
 
-void CBattleStackAnimation::shiftColor(ColorShifter * shifter)
+void CBattleStackAnimation::shiftColor(const ColorShifter * shifter)
 {
 	assert(myAnim);
 	myAnim->shiftColor(shifter);
--- a/client/battle/CBattleAnimations.cpp
+++ b/client/battle/CBattleAnimations.cpp
@@ -90,6 +90,12 @@ CBattleStackAnimation::CBattleStackAnimation(CBattleInterface * owner, const CSt
 	assert(myAnim);
 }
 
+void CBattleStackAnimation::shiftColor(ColorShifter * shifter)
+{
+	assert(myAnim);
+	myAnim->shiftColor(shifter);
+}
+
 void CAttackAnimation::nextFrame()
 {
 	if(myAnim->getType() != group)
--- a/client/battle/CBattleAnimations.h
+++ b/client/battle/CBattleAnimations.h
@@ -43,6 +43,8 @@ public:
 	const CStack * stack; //id of stack whose animation it is
 
 	CBattleStackAnimation(CBattleInterface * _owner, const CStack * _stack);
+
+	void shiftColor(ColorShifter * shifter);
 };
 
 /// This class is responsible for managing the battle attack animation
--- a/client/battle/CBattleAnimations.h
+++ b/client/battle/CBattleAnimations.h
@@ -44,7 +44,7 @@ public:
 
 	CBattleStackAnimation(CBattleInterface * _owner, const CStack * _stack);
 
-	void shiftColor(ColorShifter * shifter);
+	void shiftColor(const ColorShifter * shifter);
 };
 
 /// This class is responsible for managing the battle attack animation
--- a/client/battle/CCreatureAnimation.cpp
+++ b/client/battle/CCreatureAnimation.cpp
@@ -142,6 +142,19 @@ void CCreatureAnimation::setType(CCreatureAnim::EAnimType type)
 	play();
 }
 
+void CCreatureAnimation::shiftColor(ColorShifter * shifter)
+{
+	if (forward)
+	{
+		forward->shiftColor(shifter);
+	}
+
+	if (reverse)
+	{
+		reverse->shiftColor(shifter);
+	}
+}
+
 CCreatureAnimation::CCreatureAnimation(const std::string & name_, TSpeedController controller)
 	: name(name_),
 	  speed(0.1),
--- a/client/battle/CCreatureAnimation.cpp
+++ b/client/battle/CCreatureAnimation.cpp
@@ -142,17 +142,13 @@ void CCreatureAnimation::setType(CCreatureAnim::EAnimType type)
 	play();
 }
 
-void CCreatureAnimation::shiftColor(ColorShifter * shifter)
+void CCreatureAnimation::shiftColor(const ColorShifter* shifter)
 {
-	if (forward)
-	{
+	if(forward)
 		forward->shiftColor(shifter);
-	}
 
-	if (reverse)
-	{
+	if(reverse)
 		reverse->shiftColor(shifter);
-	}
 }
 
 CCreatureAnimation::CCreatureAnimation(const std::string & name_, TSpeedController controller)
--- a/client/battle/CCreatureAnimation.h
+++ b/client/battle/CCreatureAnimation.h
@@ -105,7 +105,7 @@ public:
 	void setBorderColor(SDL_Color palette);
 
 	// tint color effect
-	void shiftColor(const ColorShifter* shifter);
+	void shiftColor(const ColorShifter * shifter);
 
 	float getCurrentFrame() const; // Gets the current frame ID relative to frame group.
 
--- a/client/battle/CCreatureAnimation.h
+++ b/client/battle/CCreatureAnimation.h
@@ -104,6 +104,9 @@ public:
 	bool incrementFrame(float timePassed);
 	void setBorderColor(SDL_Color palette);
 
+	// tint color effect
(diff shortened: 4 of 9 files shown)


client/gui/CAnimation.cpp
```diff
@@ -753,6 +779,24 @@ void SDLImage::shiftPalette(int from, int howMany)
...
+	SDL_Palette* palette = surf->format->palette;
+	for (int i = 0; i < 255; i++)
+	{
+		palette->colors[i] = shifter->shiftColor(originalPalette->colors[i]);
+	}
+}
+
+void SDLImage::resetPalette()
+{
+	SDL_Palette * pal = new SDL_Palette();
```
alexvins: `pal` in not used and also leak.

client/gui/SDL_Extensions.h
```diff
@@ -152,6 +152,57 @@ struct ColorPutter
...
+public:
+	virtual SDL_Color shiftColor(SDL_Color clr) = 0;
+};
+
+class ColorShifterLightBlue : public ColorShifter
+{
+public:
+	static ColorShifterLightBlue* create()
+	{
+		return new ColorShifterLightBlue();
```
alexvins: In this case ColorShifter objects will always leak. You could f.e. save local instance with unique_ptr and return pointer to it.

client/gui/SDL_Extensions.h
```diff
@@ -152,6 +152,57 @@ struct ColorPutter
...
+	}
+
+	SDL_Color shiftColor(SDL_Color clr) override
+	{
+		clr.b = clr.b + (255 - clr.b) / 2;
```
alexvins: This method should be `const` (and `ColorShifter` params where it used should be `const` too)

client/gui/CAnimation.h
```diff
@@ -44,6 +45,8 @@ class IImage
 
 	//only indexed bitmaps, 16 colors maximum
 	virtual void shiftPalette(int from, int howMany) = 0;
+	virtual void adjustPalette(ColorShifter * shifter) = 0;
```
alexvins: `ColorShifter * shifter` should be `const`

client/battle/CCreatureAnimation.cpp
```diff
@@ -142,6 +142,19 @@ void CCreatureAnimation::setType(CCreatureAnim::EAnimType type)
 	play();
 }
 
+void CCreatureAnimation::shiftColor(ColorShifter * shifter)
+{
+	if (forward)
```
alexvins: 1. No space after `if` please (also after `for` in other files)
2. Do not use `{}` around single statement

client/battle/CCreatureAnimation.h
```diff
@@ -104,6 +104,9 @@ class CCreatureAnimation : public CIntObject
 	bool incrementFrame(float timePassed);
 	void setBorderColor(SDL_Color palette);
 
+	// tint color effect
+	void shiftColor(ColorShifter* shifter);
```
alexvins: Put space around `*`

client/gui/CAnimation.cpp
```diff
@@ -736,6 +755,13 @@ void SDLImage::verticalFlip()
 	surf = flipped;
 }
 
+// Keep the original palette, in order to do color switching operation
+void SDLImage::savePalette()
+{
+	originalPalette = new SDL_Palette();
```
alexvins: Where does `originalPalette` being deleted? What if it has been saved already? Also you should use `SDL_AllocPalette`

client/gui/CAnimation.cpp
```diff
@@ -736,6 +755,13 @@ void SDLImage::verticalFlip()
 	surf = flipped;
 }
 
+// Keep the original palette, in order to do color switching operation
+void SDLImage::savePalette()
+{
+	originalPalette = new SDL_Palette();
+	memcpy(originalPalette, surf->format->palette, sizeof(SDL_Palette));
```
alexvins: Isn`t SDL has a function to copy palettes? I think `SDL_SetPaletteColors` should be used here.

client/gui/CAnimation.cpp
```diff
@@ -753,6 +779,24 @@ void SDLImage::shiftPalette(int from, int howMany)
 	}
 }
 
+void SDLImage::adjustPalette(ColorShifter * shifter)
+{
+	SDL_Palette* palette = surf->format->palette;
+	for (int i = 0; i < 255; i++)
+	{
+		palette->colors[i] = shifter->shiftColor(originalPalette->colors[i]);
```
alexvins: You should not alter whole palette. There are a few special color indexes that should not be changed (see CDefFile::CDefFile for reference).

client/gui/CAnimation.cpp
Toneyisnow: Good point, mem leak is also a problem I was worrying. I think change them into SDL methods and use unique_ptr would resolve the problem.

client/gui/CAnimation.h
Toneyisnow: Updated.

client/gui/SDL_Extensions.h
Toneyisnow: I think this comment is in wrong place, since it modified the color.

client/gui/SDL_Extensions.h
Toneyisnow: Updated with unique_ptr.

client/battle/CCreatureAnimation.cpp
Toneyisnow: Updated.

client/gui/CAnimation.cpp
Toneyisnow: I will check this and update soon.

client/gui/CAnimation.cpp
Toneyisnow: I will update soon.

client/gui/CAnimation.cpp
Toneyisnow: Yeah, I'm using SDL_SetPaletteColors in another function, but I didn't update here.

Toneyisnow: Updated.

client/gui/SDL_Extensions.h
```diff
@@ -152,6 +152,60 @@ struct ColorPutter
...
+public:
+	virtual SDL_Color shiftColor(SDL_Color clr) const = 0;
+};
+
+class ColorShifterLightBlue : public ColorShifter
+{
+public:
+	static ColorShifterLightBlue * create()
+	{
+		std::unique_ptr<ColorShifterLightBlue> shifter(new ColorShifterLightBlue());
```
alexvins: `shifter` should be static

client/gui/SDL_Extensions.h
```diff
@@ -152,6 +152,60 @@ struct ColorPutter
...
+		return clr;
+	}
+};
+
+class ColorShifterDeepBlue : public ColorShifter
+{
+public:
+	static ColorShifterDeepBlue * create()
+	{
+		std::unique_ptr<ColorShifterDeepBlue> shifter(new ColorShifterDeepBlue());
```
alexvins: same here

client/gui/SDL_Extensions.h
```diff
@@ -152,6 +152,60 @@ struct ColorPutter
...
+		return clr;
+	}
+};
+
+class ColorShifterDeepRed : public ColorShifter
+{
+public:
+	static ColorShifterDeepRed * create()
+	{
+		std::unique_ptr<ColorShifterDeepRed> shifter(new ColorShifterDeepRed());
```
alexvins: and here

client/battle/CBattleAnimations.cpp
```diff
@@ -90,6 +90,12 @@ CBattleStackAnimation::CBattleStackAnimation(CBattleInterface * owner, const CSt
 	assert(myAnim);
 }
 
+void CBattleStackAnimation::shiftColor(ColorShifter * shifter)
```
alexvins: `shifter` should be const here too

client/battle/CBattleAnimations.h
```diff
@@ -43,6 +43,8 @@ class CBattleStackAnimation : public CBattleAnimation
 	const CStack * stack; //id of stack whose animation it is
 
 	CBattleStackAnimation(CBattleInterface * _owner, const CStack * _stack);
+
+	void shiftColor(ColorShifter * shifter);
```
alexvins: `shifter` should be const here too

client/gui/SDL_Extensions.h
Toneyisnow: I didn't plan to make the "shifter" to be Singeleton, so here it only to create a new one.

alexvins: If "shifter" is not singleton, then you need to return smart pointer from this function or just remove it.

Toneyisnow: Removed the functions.

PR #628, opened 2020-02-06
Toneyisnow/fixbug land mine trigger

Toneyisnow:
Fix bug: LandMine is not exploding when enemy steps on it.
The reason is,
1) the mine has attribute hidden=true;
2) when enemy unit moves, the code in BattleInfo.cpp MoveUnit()  (line 817) will update the revealed to true;
3) then in the CGameHandler.cpp handleDamageFromObstacle() (line 4846) is checking , and the condition battleIsObstacleVisibleForSide() will return true, so the effect will not be triggerred.

I think there are multiple ways to fix it, but if my understanding is correct, from the comment of line 4845 in CGameHandler.cpp, "hidden obstacle triggers effects until revealed", here we should add a condition should add a "!revealed", which means:

IF (hidden && ! revealed && visitable) THEN NOT Trigger effect.


Toneyisnow:
> Too many commits for one line change. Please rebase.

PR #629, opened 2020-02-07
Fix bug: LandMine is not exploding when enemy steps on it.

Toneyisnow:
The reason is,
1) the mine has attribute hidden=true;
2) when enemy unit moves, the code in BattleInfo.cpp MoveUnit()  (line 817) will update the revealed to true;
3) then in the CGameHandler.cpp handleDamageFromObstacle() (line 4846) is checking , and the condition battleIsObstacleVisibleForSide() will return true, so the effect will not be triggerred.

I think there are multiple ways to fix it, but if my understanding is correct, from the comment of line 4845 in CGameHandler.cpp, "hidden obstacle triggers effects until revealed", here we should add a condition should add a "!revealed", which means:

IF (hidden && ! revealed && visitable) THEN NOT Trigger effect.

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -4843,7 +4843,7 @@ bool CGameHandler::handleDamageFromObstacle(const CStack * curStack, bool stackI
 				const bool oneTimeObstacle = spellObstacle->removeOnTrigger;
 
 				//hidden obstacle triggers effects until revealed
-				if(!(spellObstacle->hidden && gs->curB->battleIsObstacleVisibleForSide(*obstacle, (BattlePerspective::BattlePerspective)side)))
+				if(!(spellObstacle->hidden && !spellObstacle->revealed && gs->curB->battleIsObstacleVisibleForSide(*obstacle, (BattlePerspective::BattlePerspective)side)))
 				{
 					const CGHeroInstance * hero = gs->curB->battleGetFightingHero(spellObstacle->casterSide);
 					spells::ObstacleCasterProxy caster(gs->curB->sides.at(spellObstacle->casterSide).color, hero, spellObstacle);


alexvins:
If obstacle is revealed, then `battleIsObstacleVisibleForSide` will return true, because there is `revealed` check in `SpellCreatedObstacle::visibleForSide`. I think the problem is BattleStackMoved being applied (and revealed state updated) before obstacle damage processed.

Toneyisnow:
@alexvins Thanks, and I also agree that the pack order seems incorrect. Meanwhile, could you give me some insights about the following attributes of SpellCreatedObstable: "hidden", "revealed", "trigger", "trap"?
I'm a little confused about how they should work? I think the applied objects are only:
QUIKSAND, LANDMINE, FORCEFLD, FIREWALL.

If it could be handled correctly, it would also resolve the FireWall bug (hit 2 hex units twice).

alexvins:
> "hidden", "revealed", "trigger", "trap"?

"hidden" and "revealed" are made for LANDMINE ("hidden" is always true for Landmine, "revealed" is a subject of this PR).  
"trap" makes obstacle work like QUIKSAND - stop movement.
"trigger" does that hitting obstacle _triggers_ indirect spell effects (f.e. LANDMINE and FIREWALL have damage effect, but custom spell could do anything).

Toneyisnow:
I have made some more investigation.

The pack sequence is: BattleStackMoved -> handleDamageFromObstacle -> BattleStackMoved (if unit needs to continue moving)
The revealed is set to true in BattleStackMoved pack, then when handleDamageFromObstacle  checks it will find the mines has been revealed so it's not exploding. That's the root cause.
But when I'm trying to fix it, I found the obstacles in handleDamageFromObstacle() are "const", which means I should not update it. But after looking into the callstack i think handleDamageFromObstacle is the best place to make the fix.
For now, we have three options:
1) Remove the "revealed=true" in moveUnit(), and in handleDamageFromObstacle, remove the "const" restrict for obstacle, and then update revealed to true;
2) Just remove the "revealed=true" in moveUnit(). Then everything will be alright, since LANDMINE shall only work for once and then removed after the damage; But for QUIKSAND, the result is it cannot be revealed after triggerred;
3) Keep the current the code fix, add a "!obstacle->revealed" check.

What do you think?

alexvins:
> * Remove the "revealed=true" in moveUnit(), and in handleDamageFromObstacle, remove the "const" restrict for obstacle, and then update revealed to true;
I like the idea of removing "revealed=true" from moveUnit(), but to make it work we need to add new "netpack".

> * Keep the current the code fix, add a "!obstacle->revealed" check.
I still do not understand why it works.

Toneyisnow:
Ok, what new netpack do you want? UpdateObstacleStatus? 
The reason you think the checking logic is weird is, it says “if it’s not VISIBLE to  the unit, then take damage to the unit”, but at the moment the unit stands on it it reveals it ( which is the wrong part) 
What i’m adding is “also need to check whether it’s just revealed, if yes, also take damage”

alexvins:
> Ok, what new netpack do you want? UpdateObstacleStatus?
OK,

Toneyisnow:
So does that mean you agree to add netpack “UpdateObstacleStstus”? And I just found we already have a similar pack “ObstacleChanged”, I think I could reuse that

alexvins:
> 
> 
> So does that mean you agree to add netpack “UpdateObstacleStstus”? And I just found we already have a similar pack “ObstacleChanged”, I think I could reuse that

Reusing BattleObstaclesChanged is better.

Toneyisnow:
I have opened another PR for this. Closing this now.

PR #630, opened 2020-02-11
Fix bug: LandMine is not exploding to enemies.

Toneyisnow:
Fix bug: LandMine is not exploding to enemies.

The reason is,
the mine has attribute hidden=true;
when enemy unit moves, the code in BattleInfo.cpp MoveUnit() (line 817) will update the revealed to true;
then in the CGameHandler.cpp handleDamageFromObstacle() (line 4846) is checking , and the condition battleIsObstacleVisibleForSide() will return true, so the effect will not be triggerred.

Resolution:
1. Remove the "revealed=true" in moveUnit(), and in handleDamageFromObstacle, remove the "const" restrict for obstacle, and then update revealed to true;
2. After the takeDamage function, add a pack "BattleObstaclesChanged" to update the obstacle to be "revealed=true".

### DIFF
--- a/lib/battle/IBattleState.h
+++ b/lib/battle/IBattleState.h
@@ -87,6 +87,6 @@ public:
 	virtual void setWallState(int partOfWall, si8 state) = 0;
 
 	virtual void addObstacle(const ObstacleChanges & changes) = 0;
-	virtual void updateObstacle(const ObstacleChanges& changes) = 0;
+	virtual void updateObstacle(const ObstacleChanges & changes) = 0;
 	virtual void removeObstacle(uint32_t id) = 0;
 };
--- a/lib/battle/IBattleState.h
+++ b/lib/battle/IBattleState.h
@@ -87,5 +87,6 @@ public:
 	virtual void setWallState(int partOfWall, si8 state) = 0;
 
 	virtual void addObstacle(const ObstacleChanges & changes) = 0;
+	virtual void updateObstacle(const ObstacleChanges& changes) = 0;
 	virtual void removeObstacle(uint32_t id) = 0;
 };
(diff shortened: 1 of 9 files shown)


server/CGameHandler.cpp
```diff
@@ -4858,6 +4859,25 @@ bool CGameHandler::handleDamageFromObstacle(const CStack * curStack, bool stackI
 
 					if(oneTimeObstacle)
 						removeObstacle(*obstacle);
```
alexvins: Put `{}` around because else clause in not one line.

lib/battle/IBattleState.h
```diff
@@ -87,5 +87,6 @@ class DLL_LINKAGE IBattleState : public IBattleInfo
 	virtual void setWallState(int partOfWall, si8 state) = 0;
 
 	virtual void addObstacle(const ObstacleChanges & changes) = 0;
+	virtual void updateObstacle(const ObstacleChanges& changes) = 0;
```
alexvins: - This change requires `BattleStateMock` to be updated.
- Put spaces around `&`

Toneyisnow: Updated.

server/CGameHandler.cpp
Toneyisnow: Updated.

server/CGameHandler.cpp
```diff
@@ -51,6 +51,7 @@
 #ifndef _MSC_VER
 #include <boost/thread/xtime.hpp>
 #endif
+#include <serializer\JsonSerializer.h>
```
alexvins:
```suggestion
#include "../lib/serializer/JsonSerializer.h"
```

Toneyisnow: Oh, forgot this one

Toneyisnow: Oh, forgot this one

PR #631, opened 2020-03-08
Toneyisnow/fixbug putback artifact

Toneyisnow:
Fix bug for deselect artifact: when putting artifact back, call treeHasChanged() to update the cached values.
Also update the variables with type TBonusListPtr to auto for better code change in future.

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -312,7 +312,6 @@ void CHeroArtPlace::deselect ()
 {
 	pickSlot(false);
 	selectSlot(false);
-	CBonusSystemNode::treeHasChanged();
 
 	if(ourArt && ourArt->canBeDisassembled()) //combined art returned to its slot -> restore locks
 	{
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -311,6 +311,9 @@ void CHeroArtPlace::select ()
 void CHeroArtPlace::deselect ()
 {
 	pickSlot(false);
+	selectSlot(false);
+	CBonusSystemNode::treeHasChanged();
+
 	if(ourArt && ourArt->canBeDisassembled()) //combined art returned to its slot -> restore locks
 	{
 		for(int i = 0; i < GameConstants::BACKPACK_START; i++)
--- a/client/windows/CHeroWindow.cpp
+++ b/client/windows/CHeroWindow.cpp
@@ -43,7 +43,9 @@
 const TBonusListPtr CHeroWithMaybePickedArtifact::getAllBonuses(const CSelector & selector, const CSelector & limit, const CBonusSystemNode * root, const std::string & cachingStr) const
 {
 	TBonusListPtr out(new BonusList());
-	TBonusListPtr heroBonuses = hero->getAllBonuses(selector, limit, hero, cachingStr);
+
+	// Fix bug: we are ignoring the cachingStr here, since when pick up and put back the artifact we want to refresh the cache
+	TBonusListPtr heroBonuses = hero->getAllBonuses(selector, limit, hero); 
 	TBonusListPtr bonusesFromPickedUpArtifact;
 
 	std::shared_ptr<CArtifactsOfHero::SCommonPart> cp = cww->getCommonPart();
--- a/client/windows/CHeroWindow.cpp
+++ b/client/windows/CHeroWindow.cpp
@@ -45,7 +45,7 @@ const TBonusListPtr CHeroWithMaybePickedArtifact::getAllBonuses(const CSelector 
 	TBonusListPtr out(new BonusList());
 
 	// Fix bug: we are ignoring the cachingStr here, since when pick up and put back the artifact we want to refresh the cache
-	TBonusListPtr heroBonuses = hero->getAllBonuses(selector, limit, hero); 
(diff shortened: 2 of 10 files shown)


client/widgets/CArtifactHolder.cpp
```diff
@@ -311,6 +311,9 @@ void CHeroArtPlace::select ()
 void CHeroArtPlace::deselect ()
 {
 	pickSlot(false);
+	selectSlot(false);
+	CBonusSystemNode::treeHasChanged();
```
alexvins: Do not do that, this is a dirty workaround, you should call `treeHasChanged` only if bonus tree was actually changed and bonus tree should be changed only by applying NetPacks.

Toneyisnow: I think this is the most straight forward change. If not do this, we might need to clean up the cachingStr data in the cache, but there is no function to do that, currently only changing the tree would clean the cache. 
Since the "deselect" is not a very heavily called action, it would be fine to put a special case here. I think this is the simplest way to fix it.
Could you propose any other approaches? I think "deselect" is a special case that the coder didn't consider for the tree change, there will not be other cases like this.

alexvins: The most straightforward fix is to ignore cache in `CHeroWithMaybePickedArtifact::getAllBonuses` just pass empty string to `hero`. 
The most simple but correct fix is to modify `CHeroWithMaybePickedArtifact::getAllBonuses` to not change `heroBonuses` and change `out` instead.

Toneyisnow: I used your first approach, and validated it's working fine, so I pushed that change.
For your second approach, I think the problem is inside of the **heroBonuses**, so **getAllBonuses** cannot get correct **out** if **heroBonuses** is already wrong.

alexvins: The actual bug is that we modify `heroBonuses`, but should modify `out` instead.
```
for(auto b : *bonusesFromPickedUpArtifact)
		*heroBonuses -= b;
```

Toneyisnow: Oh no, this is not the root cause. I know that the heroBonuses should be const, and we should modify the "out". I could update this part, but that would not help resolve this issue.
During my debugging, the real issue is, the "heroBonuses = hero->getAllBonuses(selector, limit, hero);" is already returning the wrong result (because it's reading cached value after it's picked up).

Toneyisnow: I have made the code swap you mentioned, that we should update the "out" not "heroBonuses".

Toneyisnow: I suppose what you want to say is, to update the current logic:
   heroBonuses = getAllBonuses(...)
   for(auto b : *bonusesFromPickedUpArtifact)
	 out -= b

to new logic:
   heroBonuses = getAllBonuses(...)
   if (pickedUp)
      for(auto b : *bonusesFromPickedUpArtifact)
	 out -= b
   else  // Put back
    for(auto b : *bonusesFromPickedUpArtifact)
	 out += b

That would possibly resolve the issue, but I just tried to update like that, and found that in the deselect() function, the pickedUpArtifact is already called "clear()" before call getAllBonuses, so this proposal might still have problem.

alexvins: > I have made the code swap you mentioned, that we should update the "out" not "heroBonuses".

And this should be enough

Toneyisnow: Thanks. so could you help me complete this PR?

client/windows/CHeroWindow.cpp
```diff
@@ -43,19 +43,21 @@
 const TBonusListPtr CHeroWithMaybePickedArtifact::getAllBonuses(const CSelector & selector, const CSelector & limit, const CBonusSystemNode * root, const std::string & cachingStr) const
 {
 	TBonusListPtr out(new BonusList());
-	TBonusListPtr heroBonuses = hero->getAllBonuses(selector, limit, hero, cachingStr);
+
+	// Fix bug: we are ignoring the cachingStr here, since when pick up and put back the artifact we want to refresh the cache
+	const TBonusListPtr heroBonuses = hero->getAllBonuses(selector, limit, hero);
```
alexvins: Revert back to using cachingStr.

Toneyisnow: This line is actually fixing the issue, if we use cachingStr again, the result will be wrong, and the later codes cannot make it correct. So removing the cachingStr is necessary.

Toneyisnow: Hi Alex, 
I believe this is the most straight forward (being simple, with minimum change, without any confusion) way to fix this issue, although it looked a little ugly, and lose efficiency on the cache.
Since the designer of the "cache" system didn't consider this "pick up" scenario, it might take much more effort to fix that with a neat solution, for such a small issue. And I suppose this (Pick up and put the art back) should be a corner case that makes it very special in the hero artifact system, so there should not be more cases like this.
Thanks for the consideration.

alexvins: I will look again on this issue (maybe today) if I will not find better solution I`ll merge this PR as is.

alexvins:
I`ve pushed changes from https://github.com/vcmi/vcmi/pull/631/commits/0a5e163614ab55bc93239668330d110c1a710b41. For me it fixes the issue.

Toneyisnow:
Thanks for the review! I'm closing this PR now.

PR #633, opened 2020-03-19
Add Korean encoding(CP949)

hwiorn:
This patch adds Korean encoding(cp949) to settings of VCMI.
It makes Korean version HoMM3 show properly.
And It needs a CJK or Korean font in VCMI. 
`True type fonts` mod does not work when character set is CJK, so I made a [Korean font mod](https://github.com/hwiorn/vcmi-mod-korean-truetype-fonts) for this.

sc1

sc2


misiokles:
I think this commit should be merged. Maybe some possible korean contributors will arrive here?

alexvins:
@hwiorn thank you for contribution, merged. Your mod also forked to vcmi-mods, so you could maintain it there.

PR #634, opened 2020-03-25
tray notification

nullkiller:
Tray icon and notification when it is your turn for windows only

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -57,6 +57,7 @@
 #include "gui/CAnimation.h"
 #include "../lib/serializer/Connection.h"
 #include "CServerHandler.h"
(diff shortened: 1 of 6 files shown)


ArseniyShestakov:
I'm personally against having platform-specific code like that especially since there are platform-agnostic ways to do it.

nullkiller:
I agree. But it will take an arm and a leg for me to include some 3-party lib to VCMI build process for this really. I've already spent a few days implementing pure win32 version. Maybe one day somebody can implement it in platform agnostic way.

nullkiller:
Btw @ArseniyShestakov if you know some library and had some experience working with it feel free to suggest it.

PR #635, opened 2020-03-31
Fixed build error due to missing FFMpeg include path

ars3niy:
On systems where ffmpeg includes are not directly in under global include path such as /usr/include/, client/CVideoHandler.h needs the FFMpeg include path in order to include libavformat/avformat.h


runlevel5:
I could confirm this issue is reproducible on Fedora 32 of which ffmpeg-devel package installs headers under `/usr/include/ffmpeg` instead of `/usr/include`. I have been working-around this issue by creating symlink to the headers in `/usr/include`. Good to see this PR address the root of it.

PR #640, opened 2020-05-01
3117: Fix crash after AI revisits teleporter

JacobALundgren:
**	Behavioral change: Fix AI heroes sometimes attempting to probe
teleport exits using data from a previously processed hero, causing
nullptr dereferences.

VCAI::moveHeroToTile has special case logic for revisiting tiles. This
logic could cause teleport exits to be stored in
teleportChannelProbingList, without the list subsequently being
cleared. If the processing of that hero ended immediately after that,
the next hero would believe that the list is accurate for them and
attempt to teleport while not standing on a teleporter.


ArseniyShestakov:
So yeah it's obvious enough to just merge it:
https://bugs.vcmi.eu/view.php?id=3117

PR #609, opened 2019-06-28
Improve battle AI

viciious:
Any playtested feedback would be appreciated.

Changes so far:
- the AI will now consider attacking multiple units
- the preferred strategy now is to avoid collateral damage, if possible, rather than maximizing enemy damage only
- the AI will now consider blocking enemy shooters
- some generic cleanup and refactoring

The code is full of temporary hacks and logs and whatnot, please do not review just yet.

PR #590, opened 2019-05-16
Fixed Battle Creature Ordering

Ewilhan:
Fixed the messy creature order. It now follows correct ordering.

Fixes [#2837](https://bugs.vcmi.eu/view.php?id=2837).

(PS : I also removed a TODO that was wrong. Waiting moraled creatures follow the same "Slower first higher slot first" pattern as waiting creatures.)

PR #643, opened 2020-05-11
Fix CMP_stack to follow strict weak ordering of elements

viciious:
Introduced in ab1c598d4e7fd4e89184b0c5d55ff418db8f7d27, spotted here: https://github.com/vcmi/vcmi/pull/609#issuecomment-624036482

image

While working on a separate issue I was hitting an assertion in CBattleInfoCallback.cpp on the line
```
boost::sort(phase[i], CMP_stack(i, actualTurn, lastMoved));
```

Turns out `CMP_stack::operator()` failed to follow the strict weak ordering rule when the side is -1. Now using std:addressof may seem like a hack but it does the trick when comparing two unit stack that are otherwise identical.

Also, the following comments  seem dubious because as the code works, faster units receive their turn first and cases 1, 2, 3 may actually be merged. This needs to be investigated by someone more knowledgeable than myself.

```
	case 2: //fastest last, upper slot first
	case 3: //fastest last, upper slot first
```

### DIFF
--- a/lib/battle/BattleInfo.cpp
+++ b/lib/battle/BattleInfo.cpp
@@ -1071,14 +1071,14 @@ bool CMP_stack::operator()(const battle::Unit * a, const battle::Unit * b)
 			int as = a->getInitiative(turn), bs = b->getInitiative(turn);
 			if(as != bs)
 				return as > bs;
-			else
-			{
-				if(a->unitSide() == b->unitSide())
-					return a->unitSlot() < b->unitSlot();
-				else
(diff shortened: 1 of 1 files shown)


alexvins:
Looks like after #590 `CMP_stack::operator()` does not matter too much for `phase != 0`, may be we should sort only phase 0.

alexvins:
I`m merging this as is, because correct fix for "dubious commnets" is to move some `takeOneUnit` code back to CMP_stack and far out of scope of this PR.

PR #644, opened 2020-06-29
hacks to optimize a few bonus requests

nullkiller:
This optimization is mostly needed by my AI which builds temporary CArmedInstance and asks MoraleVal() from it quite often.


nullkiller:
Will try to use CCheckProxy as @alexvins suggested.

nullkiller:
If I use CCheckProxy I have to add a few instances of it to IBonusBearer which looks like interface and should not contain any fields. Same for CBonusProxy. Also CCheckProxy is in CUnitState.h under battle namesapce so it might be required to place it to near CBonusProxy first. @alexvins any ideas?

I reactivate this PR for now.

alexvins:
> 
> If I use CCheckProxy I have to add a few instances of it to IBonusBearer which looks like interface and should not contain any fields. Same for CBonusProxy. Also CCheckProxy is in CUnitState.h under battle namesapce so it might be required to place it to near CBonusProxy first. @alexvins any ideas?
> 

You could move to CBonusSystemNode, but I suspect it could be not that easy. Feel free to move CCheckProxy & CTotalsProxy out of battle namespace. (Actually CCheckProxy and CTotalsProxy  are more optimized but specialized versions of CBonusProxy)

nullkiller:
> > If I use CCheckProxy I have to add a few instances of it to IBonusBearer which looks like interface and should not contain any fields. Same for CBonusProxy. Also CCheckProxy is in CUnitState.h under battle namesapce so it might be required to place it to near CBonusProxy first. @alexvins any ideas?
> 
> 1. You could move logic to CBonusSystemNode, but I suspect it could be not that easy.
> 2. Feel free to move CCheckProxy & CTotalsProxy out of battle namespace. (Actually CCheckProxy and CTotalsProxy  are more optimized but specialized versions of CBonusProxy)

There are a few classes inheriting IBonusBearer so we can not move MoraleVal to CBonusSystemNode and the only way is to duplicate its logic or add CCheckProxy instances to IBonusBearer.
Or maybe add protected overloaded MoraleVal template function and add a parameter to it which can answer hasBonusOfType() and in CBonusSystemNode we will pass some custom structure there and in all other cases IBonusBearer itself.

alexvins:
> add CCheckProxy instances to IBonusBearer

I think this is best solution for now, even if IBonusBearer looks like as interface it actually is not. It`s Ok to put cache fields there.

nullkiller:
> I think this is best solution for now, even if IBonusBearer looks like as interface it actually is not. It`s Ok to put cache fields there.
Changed. Moved CCheckProxy and CTotalsProxy to CBonusProxy, added getValue to CTotalsProxy for more generic usage.

PR #645, opened 2020-07-07
Fix mingw build and Random Map Generator Option

Adriankhl:
This PR has done 2 things

1. fix mingw build

2. As mentioned in https://forum.vcmi.eu/t/rmg-options-different-from-homm3/5275, I think allowing arbitrary combination of “Computer only players” and “Human or Computer players” while limiting the total number to be less than 8 should be the correct behavior, following the original homm3. With this PR, I can generate 2 human players as a team and 6 cpu players as 3 teams with the "Clash of Dragons" template.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,7 +160,7 @@ if(WIN32)
 	endif(MSVC)
 
 	if(MINGW)
-		set(SYSTEM_LIBS ${SYSTEM_LIBS} ole32 oleaut32 ws2_32 mswsock dbghelp)
+		set(SYSTEM_LIBS ${SYSTEM_LIBS} ole32 oleaut32 ws2_32 mswsock dbghelp bcrypt)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -160,7 +160,7 @@ if(WIN32)
 	endif(MSVC)
 
 	if(MINGW)
-		set(SYSTEM_LIBS ${SYSTEM_LIBS} ole32 oleaut32 ws2_32 mswsock dbghelp)
+		set(SYSTEM_LIBS ${SYSTEM_LIBS} ole32 oleaut32 ws2_32 mswsock dbghelp bcrypt)
```
hgdagon, line 163: Why this way? Can't we just move line 159 outside of the if statement?

alexvins: We can , but one shared line of code is not worth it.

PR #646, opened 2020-07-11
[WIP] Temporary fix Video button crash

Adriankhl:
The crash happens when I hover my mouse on the campaigns button. Windows only, seems like Linux build works find. After my investigation, it is related to the `close()` function in `CVideoHandler.cpp`. It  seems like if it has already loaded and opened a video resource, and `close()` is called, it will crash. The `close()` function call libraries' function like `sws_freeContext(sws)` to release (supposedly) unused resource.

I believe it is related to improper resource management, maybe also thread related, but I cannot find what exactly the problem is. This PR disable the `close()` function to prevent the crash, but it may lead to memory leak.

Hopefully someone will have some ideas on this, or maybe I will eventually figure the proper fix to this issue.

### DIFF
--- a/client/mainmenu/CCampaignScreen.cpp
+++ b/client/mainmenu/CCampaignScreen.cpp
@@ -129,7 +129,8 @@ void CCampaignScreen::CCampaignButton::show(SDL_Surface * to)
 	}
 	else if(CCS->videoh->fname == video) // When you got out of the bounds of the button then close the video
 	{
-		CCS->videoh->close();
+		// Don't close the video to prevent crash, may lead to memory leak
+		// CCS->videoh->close();
 		redraw();
 	}
 }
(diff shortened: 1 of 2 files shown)


client/mainmenu/CCampaignScreen.cpp
```diff
@@ -129,7 +129,8 @@ void CCampaignScreen::CCampaignButton::show(SDL_Surface * to)
 	}
 	else if(CCS->videoh->fname == video) // When you got out of the bounds of the button then close the video
 	{
-		CCS->videoh->close();
+		// Don't close the video to prevent crash, may lead to memory leak
+		// CCS->videoh->close();
```
nullkiller, line 133: Video issue is a known problem. Would be great if you fix it. First of all where exactly it crashes? What thread is it? What is call stack? It crashes on Windows but not on Linux?

goduck777:
I have a similar issue but it is on Linux platform. I need to disable all the videoh operation in the campaign menu to avoid crash.

I use Gentoo Linux and I compiled VCMI myself. I am willing to help test this issue on my machine.

nullkiller:
As a workaraund to this issue you can delete all vide files from homm3 data. VIDEO.VID and other VID files

nullkiller:
Not reproducable any longer.

PR #647, opened 2020-09-11
Fix exception while calculating paths

ShubusCorporation:
I 've found out that nodes which are already in the queue are subject to change after the rules are applied to them and the cost field is recalculated. This will lead to an "invalid heap" error. So, my  conclusion that the boost priority queue does not fit this goal.

### DIFF
--- a/lib/CPathfinder.cpp
+++ b/lib/CPathfinder.cpp
@@ -161,7 +161,7 @@ CGPathNode * NodeStorage::getInitialNode()
 
 	initialNode->turns = 0;
 	initialNode->moveRemains = out.hero->movement;
-	initialNode->cost = 0.0;
+	initialNode->setCost(0.0);
 
 	return initialNode;
 }
@@ -169,7 +169,7 @@ CGPathNode * NodeStorage::getInitialNode()
 void NodeStorage::commit(CDestinationNodeInfo & destination, const PathNodeInfo & source)
 {
 	assert(destination.node != source.node->theNodeBefore); //two tiles can't point to each other
-	destination.node->cost = destination.cost;
+	destination.node->setCost(destination.cost);
 	destination.node->moveRemains = destination.movementLeft;
 	destination.node->turns = destination.turn;
 	destination.node->theNodeBefore = source.node;
@@ -289,6 +289,28 @@ CPathfinder::CPathfinder(
 	initializeGraph();
 }
 
+
+void CPathfinder::push(CGPathNode * node)
+{
+	if(node && !node->inPQ)
+	{
+		node->inPQ = true;
+		node->pq = &this->pq;
+		auto handle = pq.push(node);
+		node->pqHandle = handle;
+	}
+}
+
+CGPathNode * CPathfinder::topAndPop()
+{
+	auto node = pq.top();
+
+	pq.pop();
+	node->inPQ = false;
+	node->pq = nullptr;
+	return node;
+}
+
 void CPathfinder::calculatePaths()
 {
 	//logGlobal->info("Calculating paths for hero %s (adress  %d) of player %d", hero->name, hero , hero->tempOwner);
@@ -305,19 +327,19 @@ void CPathfinder::calculatePaths()
 	if(isHeroPatrolLocked())
 		return;
 
-	pq.push(initialNode);
+	push(initialNode);
+
 	while(!pq.empty())
 	{
-		auto node = pq.top();
+		auto node = topAndPop();
 		auto excludeOurHero = node->coord == initialNode->coord;
 
 		source.setNode(gs, node, excludeOurHero);
-		pq.pop();
 		source.node->locked = true;
 
 		int movement = source.node->moveRemains;
 		uint8_t turn = source.node->turns;
-		float cost = source.node->cost;
+		float cost = source.node->getCost();
 
 		hlp->updateTurnInfo(turn);
 		if(!movement)
@@ -368,7 +390,7 @@ void CPathfinder::calculatePaths()
 			}
 
 			if(!destination.blocked)
-				pq.push(destination.node);
+				push(destination.node);
 
 		} //neighbours loop
 
@@ -403,7 +425,7 @@ void CPathfinder::calculatePaths()
 				config->nodeStorage->commit(destination, source);
 
 				if(destination.node->action == CGPathNode::TELEPORT_NORMAL)
-					pq.push(destination.node);
+					push(destination.node);
 			}
 		}
 	} //queue loop
@@ -1367,7 +1389,7 @@ bool CDestinationNodeInfo::isBetterWay() const
 	if(node->turns == 0xff) //we haven't been here before
 		return true;
 	else
-		return cost < node->cost; //this route is faster
+		return cost < node->getCost(); //this route is faster
 }
 
 bool PathNodeInfo::isNodeObjectVisitable() const
--- a/lib/CPathfinder.h
+++ b/lib/CPathfinder.h
@@ -14,7 +14,8 @@
 #include "HeroBonus.h"
 #include "int3.h"
 
-#include <boost/heap/priority_queue.hpp>
+#include <boost/heap/fibonacci_heap.hpp>
+
 
 class CGHeroInstance;
 class CGObjectInstance;
@@ -25,6 +26,17 @@ class CGWhirlpool;
 class CPathfinderHelper;
 class CPathfinder;
 class PathfinderConfig;
+
+
+template<typename N>
+struct DLL_LINKAGE NodeComparer
+{
+	STRONG_INLINE
+	bool operator()(const N * lhs, const N * rhs) const
+	{
+		return lhs->getCost() > rhs->getCost();
+	}
+};
 
 struct DLL_LINKAGE CGPathNode
 {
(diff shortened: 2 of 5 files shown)


nullkiller:
Hi. That is strange. Have not seen any errors in pathfinder. It should fail quite frequently but it does not. How about performance? It is performance critical thing for AI.

ShubusCorporation:
Hi! There're some screenshots I've captured during investigation:
debug_app_ui
vs_after_build
vs_call_stack
vs_pq_state

ShubusCorporation:
When I experimentally replaced the Node comparator with a stub (return false), exception gone away. Thus, the root cause is appearance of not sorted data in the pq.

nullkiller:
But does it crash when you run it in RelWithDebInfo configuration?

nullkiller:
Also please provide your build details. Was it done with cmake or using sln included into the project. What was the version of boost and so on. You need very strong arguments to use this standalone priority queue instead of boost one. And performance and portability is one of the most tricky questions.
P.S. I am not strong at C++, just touched CPathfinder a bit for AI. Just trying to help you figure it out.

lib/CPathfinder.cpp
```diff
@@ -289,6 +289,41 @@ CPathfinder::CPathfinder(
...
 
+
+bool CPathfinder::PriorityQueue::empty() const
+{
+	return this->pq.empty();
+}
+
+void CPathfinder::PriorityQueue::push(const CGPathNode* node)
+{
+	if (node != NULL && std::find(this->pq.begin(), this->pq.end(), node) == this->pq.end())
```
nullkiller, line 300: Does it mean O(n) complexity for push operation?

lib/CPathfinder.cpp
```diff
@@ -289,6 +289,41 @@ CPathfinder::CPathfinder(
...
+	{
+		this->pq.push_back(node);
+	}
+}
+
+CPathfinder::PriorityQueue::Iter CPathfinder::PriorityQueue::topImpl() const
+{
+	auto result = std::min_element<Iter>(this->pq.cbegin(), this->pq.end(), PriorityQueue::nodeCostComparator);
```
nullkiller, line 308: Does it mean O(n) complexity for pop operation?

ShubusCorporation (commented):
> 
> Also please provide your build details. Was it done with cmake or using sln included into the project. What was the version of boost and so on. 

Firstly, I make it with cmake: `cmake d:\Sources\VCMI\source -DCMAKE_TOOLCHAIN_FILE=d:\Sources\VCMI\vcpkg\scripts\buildsystems\vcpkg.cmake -G "Visual Studio 14 2015 Win64"`
Then I opened  \VCMI\build\VCMI.sln and build it with VS 2015.

Boost was taken from vcpkg from official site.  It seems,  it's  v. 1.64

nullkiller:
boost priority queue gives logarithmic complexity (linear only in worst case) and this thing is performance critical. The size of the queue on a big map might be about 10K.
Bad order will not cause issues because invalid nodes should be already locked (node.locked == true) and will be just skipped and the thing you see is just a debug assertion which will be disabled in release build.

ShubusCorporation:
Boost priority queue gives logarithmic complexity (linear only in worst case).
But "worst case" means  case when some  kind of "inconvenient" data is pushed to the PQ, not the case when PQ is used in inappropriate way. In last case it is not a Priority Queue and "PQ complexity " makes no sense. Unlikely, that Boost authors added that assertion just to be ignored.

Locked flag is used to prevent extracted node to be pushed into PQ again (if(neighbour->locked) continue;), it's not related to the sorting issue. Locked nodes are not "invalid", but nodes to which the rules were applied, nodes which already was in queue and was extracted from there.

The current implementation issues:

1) Priority Queue does not work as  Priority Queue. It works as something trying to be "Priority Queue":
    -  Already pushed nodes are subject to change
    - Result of top / pop is not most priority node with min cost, as expected

2) There is no duplicates check. It's enough to move mouse and new duplicate node will be pushed to the queue from the "tileHover" handler. 10K nodes in queue is not a limit in this way.

3) Subjectively. On some random middle map 4x4, template JC  from a certain moment it starts to hang on each object taking.
I suspect this is related to this node loop. It's unknown, how often this issue is  reproducible.

ShubusCorporation:
inPQ_BrokenPriority

ShubusCorporation:
This is a reason to think about a more optimal solution. Push may be done as O(1) but on the top/pop loss of speed unavoidable because through current implementation it is not known, is queued data sorted or not.

May be there's a sense to make custom queue which will be intercept changing of node cost and perform resorting,
but it's not so easy to determine, how this approach will affect performance.

nullkiller:
> There is no duplicates check. It's enough to move mouse and new duplicate node will be pushed to the queue from the "tileHover" handler. 10K nodes in queue is not a limit in this way.

Do not understand here. CPathfinder calculates paths to all objects on the map at once. Hover event might trigger pathfinding to check if tile is accessible but it should be only once per each move. The size of the queue depends on how much area it is possible to rich so it will be small at the beginning for a player because it is locked by guards. As more area is accessible as bigger will be the queue size each time pathfinding is invoked.

nullkiller:
It might be possible to remove node from queue when we want to change its cost and add it back afterwards. If this operation will also be logarithmic it should be more or less ok. What do you think?

Regarding to locked flag logically if we fetched a node from pq it should be the shortest path already and thus it is not possible that somebody finds more short path and wants to add a node back to pq so I think the flag was introduced to avoid duplicates processing.

Do you plan contributing to the project further? Would be nice to talk in slack

ShubusCorporation:
The duplicates are pointers (more precisely, references to the const pointer) to the same node which is not locked yet. They appears during neighbors processing. Each node is neighbor of each of its neighbors, this is a cause.

Yes, I'm going to become a contributor in and improve this solution. I will provide it asap. 
About Slack, yes I agree, later.

nullkiller:
One more point I think we can go without top and leave only pop. Should save some cpu ticks. And I hardly recommend you to provide some performance values on a XL map for AI comparing your and old version. The order issue can cause some hardly traceable bugs but I do not think they will be noticeable but AI speed depends on CPathfinder quite much and everybody who play VCMI complains about it.

ShubusCorporation:
@nullkiller 
Hello, Andrii, take a look pls.

I found, that the Boost offers built-in approaches to update nodes  using PQ based on the mutability heaps. It seems, the Fibonacci heap has best complexity for such goals.

https://www.geeksforgeeks.org/fibonacci-heap-set-1-introduction/

There was more simple solution for interception of the node cost change, but it seems, that the solution from Microsoft 
to create properties "__declspec( property( get=get_func_name ) ) declarator"  is not common (VCMI has to support at least gcc 4.8 )

I have tried to play on different maps (small and XL) with different heaps (binomial and fibonacci). 
It looks, that it works well.

nullkiller:
@ShubusCorporation
Hi Dmitry.
That looks way different and now I like it. Going to read about this fibonacci heap a bit. Anyway I have no rights to merge anything so lets wait for @alexvins

lib/CPathfinder.h
```diff
@@ -54,20 +69,61 @@ struct DLL_LINKAGE CGPathNode
 		BLOCKED //tile can't be entered nor visited
 	};
 
+	STRONG_INLINE
+	void setCost(float value)
+	{
+		if (value == cost._cost) {
+			return;
```
alexvins, line 76: Please do not put `{}` around one line blocks

lib/CPathfinder.h
```diff
@@ -54,20 +69,61 @@ struct DLL_LINKAGE CGPathNode
...
+		}
+		bool getUpNode = value < cost._cost;
+		cost._cost = value;
+		// If the node is in the heap, update heap.
+		if (inPQ && pq != nullptr)
```
alexvins, line 81: No space after `if` please,
```suggestion
		if(inPQ && pq != nullptr)
```

lib/CPathfinder.h
```diff
@@ -26,6 +30,17 @@ class CPathfinderHelper;
 class CPathfinder;
 class PathfinderConfig;
 
+
+template<typename N>
+struct DLL_LINKAGE NodeComparer
+{
+	STRONG_INLINE
+	bool operator()(const N* lhs, const N* rhs) const
```
alexvins, line 38:
```suggestion
	bool operator()(const N * lhs, const N * rhs) const
```

lib/CPathfinder.h
```diff
@@ -441,6 +513,12 @@ class CPathfinder : private CGameInfoCallback
 
 	void initializePatrol();
 	void initializeGraph();
+
+	STRONG_INLINE
+	void push(CGPathNode* node);
```
alexvins, line 518:
```suggestion
	void push(CGPathNode * node);
```

lib/CPathfinder.h
```diff
@@ -441,6 +513,12 @@ class CPathfinder : private CGameInfoCallback
...
+
+	STRONG_INLINE
+	CGPathNode* topAndPop();
```
alexvins, line 521:
```suggestion
	CGPathNode * topAndPop();
```

lib/CPathfinder.h
```diff
@@ -54,20 +69,61 @@ struct DLL_LINKAGE CGPathNode
...
+
+	class Cost
+	{
+	public:
+		STRONG_INLINE
+		Cost() : _cost(0.0) {}
+		STRONG_INLINE
+		explicit Cost(float value) : _cost(value) {}
+		STRONG_INLINE
+		operator float() const {
```
alexvins, line 102:
```suggestion
		operator float() const 
		{
```

alexvins (changes requested):

alexvins (approved):
Looks fine, thank you.

nullkiller:
Ok, I will merge it to my branch and perform some testing

nullkiller:
Automatic test suit passed, performance degradation on XL map was not noticed.

nullkiller (approved):

PR #649, opened 2020-09-23
Fix: Random map generator crashes when nativeTerrain is not specified in the town mod

ShubusCorporation:
Reproduction steps:
1. Setup "Ruins" town mod 
2. Try to generate some random map with "Ruins" player.

Expected result: 
Random map should be generated.

Actual result:
Server crashes.

RCA: 
When the "nativeTerrain" parameter is not specified in the faction mod, random map generator assumes, that the native terrain for this faction is a Border (-1), then it tries to create game objects on the border and crashes.

Solution: 
Let's assume, that the default faction terrain will be a grass...

### DIFF
--- a/lib/CTownHandler.cpp
+++ b/lib/CTownHandler.cpp
@@ -708,6 +708,22 @@ void CTownHandler::loadPuzzle(CFaction &faction, const JsonNode &source)
 	assert(faction.puzzleMap.size() == GameConstants::PUZZLE_MAP_PIECES);
 }
 
+ETerrainType::EETerrainType CTownHandler::getDefaultTerrainForAlignment(EAlignment::EAlignment alignment) const
+{
+	ETerrainType::EETerrainType terrain = defaultGoodTerrain;
+
+	switch(alignment)
+	{
+	case EAlignment::EAlignment::EVIL:
+		terrain = defaultEvilTerrain;
+		break;
+	case EAlignment::EAlignment::NEUTRAL:
+		terrain = defaultNeutralTerrain;
+		break;
+	}
+	return terrain;
+}
+
 CFaction * CTownHandler::loadFromJson(const JsonNode &source, const std::string & identifier)
 {
 	auto  faction = new CFaction();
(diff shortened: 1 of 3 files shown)


lib/CTownHandler.h
```diff
@@ -298,6 +298,8 @@ class DLL_LINKAGE CTownHandler : public IHandlerBase
 
 	std::map<CTown *, JsonNode> warMachinesToLoad;
 	std::vector<BuildingRequirementsHelper> requirementsToLoad;
+	ETerrainType::EETerrainType defaultFactionTerrain = ETerrainType::EETerrainType::GRASS;
```
alexvins, line 301: This should be static const

lib/CTownHandler.cpp
```diff
@@ -718,9 +718,12 @@ CFaction * CTownHandler::loadFromJson(const JsonNode &source, const std::string
 	faction->creatureBg120 = source["creatureBackground"]["120px"].String();
 	faction->creatureBg130 = source["creatureBackground"]["130px"].String();
 
+	auto nativeTerrain = source["nativeTerrain"];
+
+	faction->nativeTerrain = nativeTerrain.isNull()
+		? ETerrainType(defaultFactionTerrain)
```
alexvins, line 724: Is this cast actually needed?

alexvins (changes requested):

lib/CTownHandler.cpp
ShubusCorporation: `ETerrainType(defaultFactionTerrain)` in this case is same as `defaultFactionTerrain` 

Firstly, `ETerrainType(EETerrainType _num = WRONG) : num(_num)`
constructor will be called, then `ETerrainType& operator=(const ETerrainType& other)`

In the first case it's clear at a glance, that lvalue has  ETerrainType type, not  EETerrainType

ID_LIKE_CLASS_COMMON(ETerrainType, EETerrainType) macros does not
produce operator= which takes EETerrainType as argument.

But, if I add such operator= , one less object will be constructed

alexvins (approved):

PR #660, opened 2020-10-22
Mod system improvement: Patch 2

ShubusCorporation:
- Show correct building names on the popups
- Show actual picture for Magic University
- Fix RMG issues and assertions

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -846,6 +846,23 @@ void CCastleBuildings::enterFountain(const BuildingID & building, BuildingSubID:
 {
 	std::vector<std::shared_ptr<CComponent>> comps(1, std::make_shared<CComponent>(CComponent::building,town->subID, building));
 	std::string descr = town->town->buildings.find(building)->second->Description();
+	std::string hasNotProduced;
+	std::string hasProduced;
+
+	if(this->town->town->faction->index == (TFaction)ETownType::RAMPART)
+	{
+		hasNotProduced = CGI->generaltexth->allTexts[677];
+		hasProduced = CGI->generaltexth->allTexts[678];
+	}
+	else
+	{
+		auto buildingName = town->town->getSpecialBuilding(subID)->Name();
+
+		hasNotProduced = std::string(CGI->generaltexth->localizedTexts["townHall"]["hasNotProduced"].String());
+		hasProduced = std::string(CGI->generaltexth->localizedTexts["townHall"]["hasProduced"].String());
+		boost::algorithm::replace_first(hasNotProduced, "%s", buildingName);
+		boost::algorithm::replace_first(hasProduced, "%s", buildingName);
+	}
 
 	bool isMysticPondOrItsUpgrade = subID == BuildingSubID::MYSTIC_POND 
 		|| (upgrades != BuildingID::NONE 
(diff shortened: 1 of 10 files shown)


client/windows/CCastleInterface.cpp
```diff
@@ -847,6 +847,17 @@ void CCastleBuildings::enterFountain(const BuildingID & building, BuildingSubID:
...
 	std::string descr = town->town->buildings.find(building)->second->Description();
 
+	auto message677 = CGI->generaltexth->allTexts[677];
+	auto message678 = CGI->generaltexth->allTexts[678];
+	const auto pond = "pond";
+
+	if(this->town->town->faction->index != ETownType::RAMPART)
+	{
+		auto buildingName = town->town->getSpecialBuilding(subID)->Name();
+		boost::algorithm::replace_first(message677, pond, buildingName);
```
alexvins, line 857: This will work only with English localization. We need custom messages for custom buildings.

lib/CGeneralTextHandler.cpp
```diff
@@ -342,6 +342,13 @@ CGeneralTextHandler::CGeneralTextHandler()
 			allTexts.push_back(parser.readString());
 		}
 		while (parser.endLine());
+
+		allTexts[579] = "As you near the %s your body is filled with new energy. You have doubled your normal spell points.";
```
alexvins, line 346: Please do not alter `allTexts`, if you need special texts write them in `config\translate.json` and use from there (translate.json is accessible as `localizedTexts` field of `CGeneralTextHandler`).

lib/mapObjects/CGTownInstance.h
```diff
@@ -132,7 +132,17 @@ class DLL_LINKAGE CGTownBuilding : public IObjectInterface
 	BuildingSubID::EBuildingSubID bType;
 };
 
-class DLL_LINKAGE COPWBonus : public CGTownBuilding
+class DLL_LINKAGE CGBaseTownBonus : public CGTownBuilding
```
alexvins, line 135: Do not alter class inheritance of serializable objects, it breaks savegame backward compatibility and is very difficult to work around.

lib/mapObjects/CGTownInstance.h
```diff
@@ -132,7 +132,17 @@ class DLL_LINKAGE CGTownBuilding : public IObjectInterface
...
+{
+public:
+	CGBaseTownBonus(BuildingSubID::EBuildingSubID subId, CGTownInstance * town);
+	CGBaseTownBonus() {};
+
+protected:
+	std::string bonusGreeting;
```
alexvins, line 142: This should be serialized and by doing this you could break saves.

alexvins (changes requested):

lib/CTownHandler.h
```diff
@@ -122,7 +122,7 @@ class DLL_LINKAGE CBuilding
 			h & subId;
 			h & height;
 		}
-		else if(!h.saving)
+		if(!h.saving && version < 793)
```
alexvins, line 125: Do not change version in this case, it is just a fix.
```suggestion
		if(!h.saving && version < 792)

```

alexvins (changes requested):

lib/CTownHandler.h
ShubusCorporation: Maybe <= 792, because of new added visiting bonuses - we should set subID to make them working
But in this case there will be no difference between saves with visiting SubID-s  and without them

alexvins: I still do not understand.
1. Why version has been increased in this PR?
2. Why are you updating subId for version 792?

ShubusCorporation: I added new values for subId in Part II : some visiting bonuses and others

Suppose, user have save after Part I , but before Part II,
so there're no data about buildings, added in the Part II, but the serializer version is same as in the Part I - 792
Then he update VCMI, update mod, adding new supported types,  load save and note, that visiting bonus is not working.

It's because there's "NONE" subId instead of subId-s added in the Part II and update procedure did not work
because serializer version is actual

Thus, I suggest call `update792` and `updateBonusingBuildings` again to fill visiting bonuses with corresponding Ids 
and increase the serializer version to avoid unnecessary calls (when all Ids should  already present in the save)

alexvins (approved):

PR #662, opened 2020-11-07
Fix: Morale bonus should be shown correctly

ShubusCorporation:
It seems, that the Root Cause of "neutral morale"-related bugs is  initialization order of static / global fields.
There are no guarantee when referring to a global/static field from another global / static field than refereed
field is already initialized. The standard solution in such cases is a lazy initialization, achieved through the wrapping.

Also, I suggest minor optimization to get "Luck/Morale" total values without redundant `getAllBonuses` calling.

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -92,82 +92,80 @@ const std::map<std::string, TUpdaterPtr> bonusUpdaterMap =
 
 ///CBonusProxy
 CBonusProxy::CBonusProxy(const IBonusBearer * Target, CSelector Selector)
-	: cachedLast(0),
+	: bonusListCachedLast(0),
 	target(Target),
 	selector(Selector),
-	data()
+	bonusList()
 {
 
 }
 
 CBonusProxy::CBonusProxy(const CBonusProxy & other)
-	: cachedLast(other.cachedLast),
+	: bonusListCachedLast(other.bonusListCachedLast),
 	target(other.target),
 	selector(other.selector),
-	data(other.data)
+	bonusList(other.bonusList)
 {
 
 }
 
 CBonusProxy::CBonusProxy(CBonusProxy && other)
-	: cachedLast(0),
+	: bonusListCachedLast(0),
 	target(other.target),
 	selector(),
-	data()
+	bonusList()
 {
-	std::swap(cachedLast, other.cachedLast);
+	std::swap(bonusListCachedLast, other.bonusListCachedLast);
 	std::swap(selector, other.selector);
-	std::swap(data, other.data);
+	std::swap(bonusList, other.bonusList);
 }
 
 CBonusProxy & CBonusProxy::operator=(const CBonusProxy & other)
 {
-	cachedLast = other.cachedLast;
+	bonusListCachedLast = other.bonusListCachedLast;
 	selector = other.selector;
-	data = other.data;
+	bonusList = other.bonusList;
 	return *this;
 }
 
 CBonusProxy & CBonusProxy::operator=(CBonusProxy && other)
 {
-	std::swap(cachedLast, other.cachedLast);
+	std::swap(bonusListCachedLast, other.bonusListCachedLast);
 	std::swap(selector, other.selector);
-	std::swap(data, other.data);
+	std::swap(bonusList, other.bonusList);
 	return *this;
 }
 
-TConstBonusListPtr CBonusProxy::get() const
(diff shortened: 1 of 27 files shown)


lib/HeroBonus.cpp
```diff
@@ -200,13 +200,19 @@ int CTotalsProxy::getValue() const
 		value = initialValue + bonuses->totalValue();
 		cachedLast = treeVersion;
 	}
+	return value;
+}
 
+int CTotalsProxy::getValueAndList(TConstBonusListPtr & bonusList) const
+{
+	bonusList = target->getBonuses(selector);
+	value = initialValue + bonusList->totalValue();
```
alexvins, line 209: You should update cache version (`cachedLast`) here or do not change the value

alexvins (changes requested):

lib/HeroBonus.cpp
```diff
@@ -200,13 +200,25 @@ int CTotalsProxy::getValue() const
 		value = initialValue + bonuses->totalValue();
 		cachedLast = treeVersion;
 	}
+	return value;
+}
+
+int CTotalsProxy::getValueAndList(TConstBonusListPtr & bonusList) const
+{
+	const auto treeVersion = target->getTreeVersion();
+	bonusList = target->getBonuses(selector);
```
alexvins, line 209: We could cache bonuses list too, and may be even make `CTotalsProxy` a descendant of `CBonusProxy` (but do not merge completely -  `totalValue()` is somewhat costly operation)

lib/HeroBonus.h
```diff
@@ -726,6 +736,15 @@ class DLL_LINKAGE IBonusBearer
 
 	int MoraleVal() const; //range [-3, +3]
 	int LuckVal() const; //range [-3, +3]
+	/**
+	 Returns total value of all morale bonuses and sets bonusList as a pointer to the list of selected bonuses.
+	 Pay attention: May work without caching. It's not intended for high load using.
```
alexvins, line 741: With changes in `getValueAndList` this will use cache too.

alexvins (changes requested):

lib/HeroBonus.cpp
ShubusCorporation: Done.  lib/HeroBonus.h  and lib/HeroBonus.cpp was changed.
But may be it's excessive because `BonusList`  caching is already present in the `CBonusSystemNode::getAllBonuses`

lib/HeroBonus.cpp
```diff
@@ -91,83 +91,83 @@ const std::map<std::string, TUpdaterPtr> bonusUpdaterMap =
...
-TConstBonusListPtr CBonusProxy::get() const
+TConstBonusListPtr CBonusProxy::getBonusList() const
 {
-	if(target->getTreeVersion() != cachedLast || !data)
+	if(target->getTreeVersion() != bonusListCachedLast || !bonusList)
 	{
 		//TODO: support limiters
-		data = target->getAllBonuses(selector, Selector::all);
-		cachedLast = target->getTreeVersion();
+		bonusList = target->getAllBonuses(selector, Selector::all, nullptr, cachingStr);
```
alexvins, line 146: Do not use `cachingStr` in this class.

alexvins (changes requested):

lib/HeroBonus.cpp
alexvins: This class was made because cache in `CBonusSystemNode` is too slow.

lib/HeroBonus.h
```diff
@@ -71,27 +71,27 @@ class CSelector : std::function<bool(const Bonus*)>
...
 	const BonusList * operator->() const;
-private:
-	mutable int64_t cachedLast;
-	const IBonusBearer * target;
+	TConstBonusListPtr getBonusList() const;
+
+protected:
 	CSelector selector;
-	mutable TConstBonusListPtr data;
+	std::string cachingStr; //for bonusList caching
```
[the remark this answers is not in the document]
ShubusCorporation: But without  `cachingStr`  caching does not work for objects of `CHeroWithMaybePickedArtifact` class because they are being created from scratch with all their proxies each time on user click by the game interface handlers. So, for hero window bonuses are taken without caching and for army stack with caching. It's strange, but it seems, it does not matter for UI, hardly anyone will notice the difference...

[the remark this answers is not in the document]
alexvins: Such objects never were a performance bottleneck.

alexvins (approved):

PR #664, opened 2020-11-21
Try fix bypassing destroyed fort drawbridge

nullkiller:
There is an issue with fort gate drawbridge. AI is not able to bypass it because there we have some obstacle. I changed the code to ignore obstacle when the gate is destroyed or the side is defender.
Now AI can enter the fort if drawbridge is destroyed or exit the fort when defending it.
Also fix for mantis bugs: 2504, 2718

### DIFF
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -1161,7 +1161,7 @@ ReachabilityInfo CBattleInfoCallback::makeBFS(const AccessibilityInfo &accessibi
 	if(!params.startPosition.isValid()) //if got call for arrow turrets
 		return ret;
 
-	const std::set<BattleHex> quicksands = getStoppers(params.perspective);
+	const std::set<BattleHex> obstacles = getStoppers(params.perspective);
 
 	std::queue<BattleHex> hexq; //bfs queue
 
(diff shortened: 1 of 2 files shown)


nullkiller:
For now it can just hang when both sides have only walkers.

alexvins (approved):

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1179,16 +1179,26 @@ ReachabilityInfo CBattleInfoCallback::makeBFS(const AccessibilityInfo &accessibi
 		hexq.pop();
 		
 		//walking stack can't step past the obstacles
-		//TODO what if second hex of two-hex creature enters obstacles
-		if(curHex != params.startPosition && vstd::contains(obstacles, curHex))
+		if(curHex != params.startPosition)
 		{
-			if(curHex == ESiegeHex::GATE_BRIDGE)
+			auto occupiedHexes = battle::Unit::getHexes(curHex, params.doubleWide, params.side);
```
nullkiller, line 1184: Lets handle correctly twohex creatures. Now they are able to go around various obstacles without hitting them with their second hex.

alexvins (approved):

PR #665, opened 2020-11-27
2184 - fix battlefield corners unreachable for 2 hex units

nullkiller:
This fixes mantis bug when you can hide from AI in a few corners. Works only using 1 hex unit against 2 hex unit


nullkiller:
It is obviously code duplication. If somebody find a good place for this logic leave a comment. Probably stupid ai can be a part of VCMI_Lib so that we can call it from server-side to perform automatic actions. Or we can have some AI folder to share some AI routines.

alexvins (approved):

alexvins (approved):

PR #666, opened 2020-11-28
Lodestar Grail should work.

ShubusCorporation:
When Lodestar Grail is built:

- Any terrain is native for all creatures of 'Cove' faction. Thus, any hero with only 'Cove' creatures has not terrain movement penalty.
- Any creature of 'Cove' faction receives +1 attack, +1 defence, +1 speed bonuses in the battle.

When Nomad creature is in the hero army: No terrain penalty when hero is on the sand (according to the FizMiG)

### DIFF
--- a/lib/CCreatureHandler.h
+++ b/lib/CCreatureHandler.h
@@ -114,7 +114,14 @@ public:
 
 	ArtifactID warMachine;
 
-	bool isItNativeTerrain(int terrain) const;
+	bool isItNativeTerrain(ETerrainType::EETerrainType terrain) const;
+	/**
+	Returns creature native terrain considering some terrain bonuses.
+	@param considerBonus is used to avoid Dead Lock when this method is called inside getAllBonuses
+	considerBonus = true is called from Pathfinder and fills actual nativeTerrain considering bonus(es).
+	considerBonus = false is called on Battle init and returns already prepared nativeTerrain without Bonus system calling.
+	*/
+	ETerrainType::EETerrainType getNativeTerrain() const;
 	bool isDoubleWide() const; //returns true if unit is double wide on battlefield
 	bool isFlying() const; //returns true if it is a flying unit
 	bool isShooting() const; //returns true if unit can shoot
@@ -241,6 +248,7 @@ public:
 	CreatureID pickRandomMonster(CRandomGenerator & rand, int tier = -1) const; //tier <1 - CREATURES_PER_TOWN> or -1 for any
 	void addBonusForTier(int tier, std::shared_ptr<Bonus> b); //tier must be <1-7>
 	void addBonusForAllCreatures(std::shared_ptr<Bonus> b);
+	void removeBonusesFromAllCreatures();
 
 	CCreatureHandler();
 	~CCreatureHandler();
--- a/lib/GameConstants.h
+++ b/lib/GameConstants.h
@@ -860,8 +860,9 @@ class DLL_LINKAGE ETerrainType
 public:
 	enum EETerrainType
 	{
+		ANY_TERRAIN = -3,
 		WRONG = -2, BORDER = -1, DIRT, SAND, GRASS, SNOW, SWAMP,
-		ROUGH, SUBTERRANEAN, LAVA, WATER, ROCK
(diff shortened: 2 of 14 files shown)


lib/GameConstants.h
```diff
@@ -861,7 +861,7 @@ class DLL_LINKAGE ETerrainType
 	enum EETerrainType
 	{
 		WRONG = -2, BORDER = -1, DIRT, SAND, GRASS, SNOW, SWAMP,
-		ROUGH, SUBTERRANEAN, LAVA, WATER, ROCK
+		ROUGH, SUBTERRANEAN, LAVA, WATER, ROCK, ANY_TERRAIN
```
alexvins, line 864: ANY_TERRAIN should have negative value like WRONG & BORDER

lib/CCreatureHandler.h
```diff
@@ -34,6 +34,7 @@ class DLL_LINKAGE CCreature : public CBonusSystemNode
 
 	CreatureID idNumber;
 	TFaction faction;
+	mutable ETerrainType::EETerrainType nativeTerrain;
```
alexvins, line 37: this field need serialization, and to prevent desync add additional mutable fileld

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -89,43 +89,62 @@ ui32 CGHeroInstance::getTileCost(const TerrainTile & dest, const TerrainTile & f
 			break;
 		}
 	}
-	else if(ti->nativeTerrain != from.terType && !ti->hasBonusOfType(Bonus::NO_TERRAIN_PENALTY, from.terType))
+	else
 	{
-		static const CSelector selectorPATHFINDING = Selector::typeSubtype(Bonus::SECONDARY_SKILL_PREMY, SecondarySkill::PATHFINDING);
-		static const std::string keyPATHFINDING = "type_"+std::to_string((si32)Bonus::SECONDARY_SKILL_PREMY)+"s_"+std::to_string((si32)SecondarySkill::PATHFINDING);
+		bool nomadBonus = from.terType == ETerrainType::SAND && ti->heroHasCreature(CreatureID::NOMAD);
```
alexvins, line 94: Why is this needed? Nomad already has NO_TERRAIN_PENALTY for sand.

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -1232,6 +1232,16 @@ void CGTownInstance::recreateBuildingsBonuses()
 		addBonusIfBuilt(BuildingID::GRAIL, Bonus::PRIMARY_SKILL, +10, PrimarySkill::ATTACK); //grail
 		addBonusIfBuilt(BuildingID::GRAIL, Bonus::PRIMARY_SKILL, +10, PrimarySkill::DEFENSE); //grail
 	}
+	else if(this->town->getFactionName() == "Cove" && hasBuilt(BuildingID::GRAIL))
```
alexvins, line 1235: `this->town->getFactionName() == "Cove"` such hardcoding should be avoided.

alexvins (changes requested):

lib/mapObjects/CGTownInstance.cpp
ShubusCorporation: Will be changed later, with general solution for all grails. There's no better solution now. Cove Grail should work because it's very popular and semi-official town.

ShubusCorporation (commented):
Faced a bug, so the solution was reworked

lib/CCreatureHandler.h
ShubusCorporation: This field is removed, there's another field in the Stack class

ShubusCorporation:
@alexvins

lib/mapObjects/CGTownInstance.cpp
alexvins: You could use identifier, but do not use localized name.

ShubusCorporation: Accepted.

alexvins (approved):

PR #667, opened 2020-12-01
Improve battle ai fort capturing

nullkiller:
Allow AI to use catapult, destroyed wall

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -16,6 +16,7 @@
 #include "EnemyInfo.h"
 #include "../../lib/CStopWatch.h"
 #include "../../lib/CThreadHelper.h"
+#include "../../lib/mapObjects/CGTownInstance.h"
 #include "../../lib/spells/CSpellHandler.h"
 #include "../../lib/spells/ISpellMechanics.h"
 #include "../../lib/CStack.h" // TODO: remove
@@ -59,6 +60,26 @@ SpellTypes spellType(const CSpell * spell)
 		return SpellTypes::BATTLE;
 
 	return SpellTypes::OTHER;
+}
+
+std::vector<BattleHex> CBattleAI::getBrokenWallMoatHexes() const
+{
+	std::vector<BattleHex> result;
+	
+	for(int wallPart = EWallPart::BOTTOM_WALL; wallPart < EWallPart::UPPER_WALL; wallPart++)
+	{
+		auto state = cb->battleGetWallState(wallPart);
+
+		if(state != EWallState::DESTROYED)
+			continue;
+		
+		auto wallHex = cb->wallPartToBattleHex((EWallPart::EWallPart)wallPart);
+		auto moatHex = wallHex.cloneInDirection(BattleHex::LEFT);
+
+		result.push_back(moatHex);
+	}
+
+	return result;
 }
 
 CBattleAI::CBattleAI()
@@ -199,13 +220,28 @@ BattleAction CBattleAI::activeStack( const CStack * stack )
 
 					if(dists.distToNearestNeighbour(stack, *closestEnemy) < GameConstants::BFIELD_SIZE)
 					{
-						return goTowards(stack, *closestEnemy);
+						return goTowardsNearest(stack, (*closestEnemy)->getAttackableHexes(stack));
 					}
 				}
 			}
 			else
 			{
 				return BattleAction::makeWait(stack);
+			}
+		}
+
+		if(!stack->hasBonusOfType(Bonus::FLYING)
+			&& stack->unitSide() == BattleSide::ATTACKER
+			&& cb->battleGetSiegeLevel() >= CGTownInstance::CITADEL)
+		{
+			auto brokenWallMoat = getBrokenWallMoatHexes();
+
+			if(brokenWallMoat.size())
+			{
+				if(stack->doubleWide() && vstd::contains(brokenWallMoat, stack->getPosition()))
+					return BattleAction::makeMove(stack, stack->getPosition().cloneInDirection(BattleHex::RIGHT));
+				else
+					return goTowardsNearest(stack, brokenWallMoat);
 			}
 		}
 	}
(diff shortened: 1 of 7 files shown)


AI/BattleAI/BattleAI.cpp
```diff
@@ -280,7 +325,47 @@ BattleAction CBattleAI::goTowards(const CStack * stack, const battle::Unit * ene
...
+			EWallPart::UPPER_WALL
+		};
+
+		for(auto wallPart : wallParts)
+		{
+			auto wallState = cb->battleGetWallState(wallPart);
+
+			if(wallState != EWallState::DESTROYED)
+			{
+				targetHex = cb->wallPartToBattleHex(wallPart);
```
alexvins, line 353: Should be `break` here

lib/battle/Unit.cpp
```diff
@@ -82,6 +82,43 @@ std::vector<BattleHex> Unit::getSurroundingHexes(BattleHex position, bool twoHex
...
+{
+	auto defenderHexes = battle::Unit::getHexes(
+		getPosition(),
+		doubleWide(),
+		unitSide());
+
+	if(defenderHexes.size() == 2 && BattleHex::getDistance(defenderHexes.front(), defenderHexes.back()) != 1)
+	{
+		defenderHexes.pop_back();
+	}
```
alexvins, line 95:
```suggestion
		defenderHexes.pop_back();
```

lib/battle/Unit.cpp
```diff
@@ -82,6 +82,43 @@ std::vector<BattleHex> Unit::getSurroundingHexes(BattleHex position, bool twoHex
...
+
+		if(hexes.size() == 2 && BattleHex::getDistance(hexes.front(), hexes.back()) != 1)
+		{
+			hexes.pop_back();
+		}
+
+		for(auto hex : hexes)
+		{
+			vstd::concatenate(targetableHexes, hex.neighbouringTiles());
+		}
```
alexvins, line 114:
```suggestion
		if(hexes.size() == 2 && BattleHex::getDistance(hexes.front(), hexes.back()) != 1)
			hexes.pop_back();

		for(auto hex : hexes)
			vstd::concatenate(targetableHexes, hex.neighbouringTiles());
```

AI/BattleAI/BattleAI.cpp
```diff
@@ -280,7 +325,47 @@ BattleAction CBattleAI::goTowards(const CStack * stack, const battle::Unit * ene
...
+			}
+		}
+	}
+
+	if(!targetHex.isValid())
+	{
+		return BattleAction::makeDefend(stack);
+	}
```
alexvins, line 361:
```suggestion
		return BattleAction::makeDefend(stack);

```

AI/BattleAI/BattleAI.cpp
```diff
@@ -208,6 +229,23 @@ BattleAction CBattleAI::activeStack( const CStack * stack )
...
+			&& cb->battleGetSiegeLevel() >= CGTownInstance::CITADEL)
+		{
+			auto brokenWallMoat = getBrokenWallHexes();
+
+			if(brokenWallMoat.size())
+			{
+				if(stack->doubleWide() && vstd::contains(brokenWallMoat, stack->getPosition()))
+				{
+					return BattleAction::makeMove(stack, stack->getPosition().cloneInDirection(BattleHex::RIGHT));
+				}
```
alexvins, line 244:
```suggestion
					return BattleAction::makeMove(stack, stack->getPosition().cloneInDirection(BattleHex::RIGHT));

```

AI/BattleAI/BattleAI.cpp
```diff
@@ -280,7 +325,47 @@ BattleAction CBattleAI::goTowards(const CStack * stack, const battle::Unit * ene
 
 BattleAction CBattleAI::useCatapult(const CStack * stack)
```
alexvins, line 326: Is this function still unused?

alexvins (changes requested):

AI/BattleAI/BattleAI.cpp
nullkiller: Invert if to reduce nesting?

nullkiller: Do not understand how break can be used here. I need all broken wall hexes because goTowardsNearest will select the nearest one

alexvins: You are cycling all wall parts and select last not destroyed, if you really need last one, set `wallParts` in reverse order and break `wallPart : wallParts` cycle after first match.

AI/BattleAI/BattleAI.cpp
nullkiller: Confirmed. Before when AI with balistics the balista was just doing nothing

AI/BattleAI/BattleAI.cpp
nullkiller: Not last but all destroyed parts so I can pass through the one which is nearest.

AI/BattleAI/BattleAI.cpp
```diff
@@ -61,6 +62,26 @@ SpellTypes spellType(const CSpell * spell)
 	return SpellTypes::OTHER;
 }
 
+std::vector<BattleHex> CBattleAI::getBrokenWallMoatHexes() const
```
nullkiller, line 65: I can use Destroyed here to be consistent with enum value DESTROYED

AI/BattleAI/BattleAI.cpp
nullkiller: This is balista. My fault. I thought it was another method.

nullkiller: I do not plan to implement something very clever and arguing what has to be broken first walls or towers. My idea was simple. Break gate and afterwards break towers so that they do not give extra damage. Btw need to test ballista for fort which has no towers

nullkiller: If you have better idea in which order it is better to break walls in any situation I can accept it so. The main idea of all these changes is to reduce probability of freeze because of endless battle.

nullkiller: Break added

alexvins (approved):

PR #670, opened 2020-12-13
found a bug in code where AI can change client gamestate directly

nullkiller:
There is an issue detected in Battle AI code. It can change client gamestate which in turn can crash the whole game.
Made a clone of the bonus to avoid gamestate modification

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -578,6 +578,7 @@ void CBattleAI::attemptCastingSpell()
 		size_t minTurnSpan = ourUnits/3; //todo: tweak this
 
 		std::vector<battle::Units> newTurnOrder;
+
 		state.battleGetTurnOrder(newTurnOrder, amount, 2);
 
 		const bool turnSpanOK = evaluateQueue(newValueOfStack, newTurnOrder, &state, minTurnSpan, nullptr);
--- a/AI/BattleAI/StackWithBonuses.cpp
+++ b/AI/BattleAI/StackWithBonuses.cpp
@@ -14,11 +14,16 @@
 
 void actualizeEffect(TBonusListPtr target, const Bonus & ef)
 {
-	for(auto bonus : *target) //TODO: optimize
+	for(auto & bonus : *target) //TODO: optimize
 	{
 		if(bonus->source == Bonus::SPELL_EFFECT && bonus->type == ef.type && bonus->subtype == ef.subtype)
 		{
-			bonus->turnsRemain = std::max(bonus->turnsRemain, ef.turnsRemain);
+			if(bonus->turnsRemain < ef.turnsRemain)
+			{
+				bonus.reset(new Bonus(*bonus));
+
+				bonus->turnsRemain = ef.turnsRemain;
+			}
 		}
 	}
 }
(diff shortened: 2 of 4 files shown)


AI/BattleAI/BattleAI.cpp
```diff
@@ -578,6 +578,7 @@ void CBattleAI::attemptCastingSpell()
 		size_t minTurnSpan = ourUnits/3; //todo: tweak this
 
 		std::vector<battle::Units> newTurnOrder;
+
 		state.battleGetTurnOrder(newTurnOrder, amount, 2);
```
nullkiller, line 582: This thing runs in parallel and invokes std::sort on a list of stacks with CMP_Stack comparator

AI/BattleAI/BattleAI.cpp
```diff
@@ -621,7 +622,7 @@ void CBattleAI::attemptCastingSpell()
 	for(PossibleSpellcast & psc : possibleCasts)
 		tasks.push_back(std::bind(evaluateSpellcast, &psc));
 
-	uint32_t threadCount = boost::thread::hardware_concurrency();
+	uint32_t threadCount = 30 + boost::thread::hardware_concurrency();
```
nullkiller, line 625: Will revert. Needed to reproduce the issue

AI/BattleAI/StackWithBonuses.cpp
```diff
@@ -18,7 +18,8 @@ void actualizeEffect(TBonusListPtr target, const Bonus & ef)
 	{
 		if(bonus->source == Bonus::SPELL_EFFECT && bonus->type == ef.type && bonus->subtype == ef.subtype)
 		{
-			bonus->turnsRemain = std::max(bonus->turnsRemain, ef.turnsRemain);
+			//TODO: this code changes gamestate
+			//bonus->turnsRemain = std::max(bonus->turnsRemain, ef.turnsRemain);
```
nullkiller, line 22: And here is what causes this issue. Bonuses from TBonusListPtr target can be from real stacks from gamestate.

lib/battle/BattleInfo.cpp
```diff
@@ -1066,33 +1066,21 @@ bool CMP_stack::operator()(const battle::Unit * a, const battle::Unit * b)
 	{
 	case 0: //catapult moves after turrets
 		return a->creatureIndex() > b->creatureIndex(); //catapult is 145 and turrets are 149
-	case 1: //fastest first, upper slot first
```
nullkiller, line 1069: The logic of phase 1 and phase 3 looks absolutely the same

lib/battle/BattleInfo.cpp
```diff
@@ -1066,33 +1066,21 @@ bool CMP_stack::operator()(const battle::Unit * a, const battle::Unit * b)
...
+			
 			if(as != bs)
 				return as > bs;
 
 			if(a->unitSide() == b->unitSide())
 				return a->unitSlot() < b->unitSlot();
-			else if (a->unitSide() == side || b->unitSide() == side)
-				return a->unitSide() != side;
-
-			return std::addressof(a) > std::addressof(b);
```
nullkiller, line 1095: This does not cause issues but still looks inconsistent and was my first guess. this->side can sometimes be 255 or something and it looks like usual expected scenario.

AI/BattleAI/StackWithBonuses.cpp
nullkiller: After  commenting out this code I was not able to reproduce the crash again.

nullkiller: Cloned the bonus from the gamestate and changed the clone

alexvins (approved):

PR #672, opened 2021-01-06
Feature: Mods system improvement, Part III. Bonusing buildings customization.

ShubusCorporation:
New special building parameters are introduced:

- bonuses (list of bonuses), which holds all permanent bonuses which intended to be apllied when building is built
- onVisitBonuses (list of bonuses), which holds all on bonuses for visiting hero when building is built
- overrides (list of overridden buildings): Now it's possible to determine behavior of special building, should it override effects from other special buildings (such as Brotherhood overrides Tavern) or not.

- Necropolis bonusing buildings are switched to the JSON-defined bonus system.

Cove-like Grail, affects all creatures which belong to the current town faction:
`"bonuses" : [ { "type" : "NO_TERRAIN_PENALTY", "propagator" : "ALL_CREATURES", "limiters" : "CREATURE_FACTION" } ]`
This type of propagator will cause effect even if the corresponding town with the Grail is neutral.

List of available propagators now: 

"BATTLE_WIDE"
"VISITED_TOWN_AND_VISITOR",
"PLAYER_PROPAGATOR", 
"HERO", 
"TEAM_PROPAGATOR", 
"GLOBAL_EFFECT",
"ALL_CREATURES"

Limiters are:

"SHOOTER_ONLY",
"DRAGON_NATURE",
"IS_UNDEAD",
"CREATURE_NATIVE_TERRAIN"
"CREATURE_FACTION"

Not all of them are tested in the context of this PR.

Bonuses may be combined. For example, that's Fortress town Grail definition:
`"bonuses": [
		{ "type": "PRIMARY_SKILL", "subtype": "primSkill.attack", "val": 10 },
		{ "type": "PRIMARY_SKILL", "subtype": "primSkill.defence", "val": 10 }
]`

Confession Chapel (increases morale and luck by 1 for troops defending the city) in the Cathedral town:
`"bonuses": [ { "type": "MORALE", "val": 1 }, { "type": "LUCK", "val": 1 } ]`

It's not recommended to mix permanent and time limited bonuses.
Example: Increases luck of any visiting hero for next battle by +1 (such as Nanomedic in the Covenant town):

`"onVisitBonuses": [ { "duration": "ONE_BATTLE", "type": "LUCK", "val": 1 } ]`


server/CGameHandler.cpp
```diff
@@ -3147,33 +3147,40 @@ bool CGameHandler::buildStructure(ObjectInstanceID tid, BuildingID requestedID,
...
+	for(auto builtID : ns.bid)
+	{
 		processAfterBuiltStructure(builtID);
+		auto building = t->town->buildings.at(builtID);
 
+		if(building->overrideBids.empty()) //NewStructures::applyGs handles it on the client side
+			continue;
+		
+		for(auto overrideBid : building->overrideBids)
+			const_cast<CGTownInstance *>(t)->overriddenBuildings.insert(overrideBid);
```
alexvins, line 3166: Should this be done during netpack apply (and not only on server side)?

alexvins (changes requested):

server/CGameHandler.cpp
ShubusCorporation: > 
> 
> Should this be done during netpack apply (and not only on server side)?

Ok, `NewStructures::applyGs` works on on the client and server side.
Also, visited bonuses should be overridden too, so I've add corresponding code.

lib/CTownHandler.cpp
```diff
@@ -93,50 +94,115 @@ void CBuilding::deserializeFix()
...
+	CTownHandler::loadSpecialBuildingBonuses(currentBuilding["bonuses"], buildingBonuses, this);
+	CTownHandler::loadSpecialBuildingBonuses(currentBuilding["onVisitBonuses"], onVisitBonuses, this);
+
+	if(!onVisitBonuses.empty())
+	{
+		if(subId == BuildingSubID::NONE)
+			subId = BuildingSubID::CUSTOM_VISITING_BONUS;
+
+		for(auto & bonus : onVisitBonuses)
+			bonus->sid = (ui32)((ui64)this & 0xFFFFFFFFLL);
```
ShubusCorporation, line 193: It's supposed, that low bytes of address are unique

lib/NetPacksLib.cpp
```diff
@@ -245,15 +245,21 @@ DLL_LINKAGE void GiveBonus::applyGs(CGameState *gs)
...
-		descr = VLC->generaltexth->arraytxt[bonus.val > 0 ? 110 : 109]; //+/-%d Temporary until next battle"
+			if(bonus.source == Bonus::OBJECT)
+				descr = VLC->generaltexth->arraytxt[bonus.val > 0 ? 110 : 109]; //+/-%d Temporary until next battle"
+			else if(bonus.source == Bonus::TOWN_STRUCTURE)
+			{
+				descr = bonus.description;
+				return;
+			}
+			else
+				bdescr.toString(descr);
```
alexvins, line 258:
```suggestion
			if(bonus.source == Bonus::OBJECT)
			{
				descr = VLC->generaltexth->arraytxt[bonus.val > 0 ? 110 : 109]; //+/-%d Temporary until next battle"
			}
			else if(bonus.source == Bonus::TOWN_STRUCTURE)
			{
				descr = bonus.description;
				return;
			}
			else
			{
				bdescr.toString(descr);
			}

```

alexvins (changes requested):
Bonus source id should be more stable than just local pointer (may be smth like town`s object instance id combined with building id).

Global.h
```diff
@@ -745,6 +743,13 @@ namespace vstd
 		}
 		return ret;
 	}
+
+	STRONG_INLINE
+	ui32 getSid32(ui32 high, ui32 low)
+	{
+		return (high << 16) + low;
+	}
+	using boost::math::round;
```
alexvins, line 752: - Please do not change `Global.h` with no good reason. 
- If you really want this as a separate finction, add a method to `Bonus`

alexvins (changes requested):

alexvins (approved):

PR #675, opened 2021-01-20
Fixed: Mods dependency issues leading to hanging & 'Restart Scenario' button functionality

ShubusCorporation:
Fixed:

- There's a hanging each time when mod dependency  is not resolved, it scares new users.
- 'Restart Scenario' button does not work properly since April 2018.


alexvins (approved):

ShubusCorporation:
And what about this PR?

PR #676, opened 2021-02-01
Patch III for Mod system improvement : Negative bonuses: new feature, fixes, town building: Magic Well 

ShubusCorporation:
- [x] New feature: Negative town bonuses for enemy army implementation
- [x] Fix: Negative bonus bearers should not be affected by own negative bonuses. So, devils, archdevils, negative artifacts (shaman's thing and some dead head), negative hero's abilities (such as Amanda from  Cathedral) should work now.
- [x] Fix: Visiting hero should receive town bonuses during town siege when he defends a town as garrison hero. Bonus tree mounting is fixed to provide town bonuses.

- [x] New feature: 'Magic Well' should work in fan towns.
- [x] New feature: Introduced new limiter. The advantage of this solution is the bonus propagation is implemented using a standard propagation mechanic, it needs just owner adjustment instead of whole bonus copying to the node on battle setup

**Samples:**

Magic Well: `"onVisitBonuses" : [ { "type": "TOWN_MAGIC_WELL" } ],`
Magic Absorber: `"bonuses" : [ { "type": "SPELL_DAMAGE_REDUCTION", "val": 50, "subtype": -1 } ]`
-2 morale to enemy army: `"bonuses" : [ { "type": "MORALE", "val": -2, "effectRange":"ONLY_ENEMY_ARMY" } ]`

**Guide to feature applying:**

No limiters or propagators  should be used with  ` "effectRange": "ONLY_ENEMY_ARMY" `

**Testing:**

Precondition: Devils are in the hero's army.

- Attack neutrals with devils outside a town.  
- Make sure, negatives are propagated from hero's army to the neutrals units.
- Make sure, own negatives are not propagated to the negative bonus owner.
- Attack neutral devils.  
- Make sure, negatives are propagated from neutrals units to the hero's army.
- Attack neutral devils with devils. 
- Make sure, enemy negatives are propagated to all units
- Attack neutral town (no devils in the hero's army) with devils in the town garrison. 
- Make sure, enemy negatives are propagated  from garrison to hero's army
- Attack neutral town (devils are the hero's army) with devils in the town garrison. 
- Make sure, enemy negatives are propagated to all units
- Attack town with built negative bonus building
- Make sure, negatives are propagated from town building to the hero's army
- Capture this town and remain weak  visitor (not garrison) hero as defender. Enemy should attack.
- Make sure, defending visitor obtains positive bonuses from town. 
- Make sure, enemy attacker obtains negative bonuses from town
- Win this battle. Move hero to garrison and wait for next attack.
- Make sure, defending garrison hero obtains positive bonuses from town. 
- Lose this battle and wait until enemy remains this town without guardians.
- Capture this town again, wait for enemy attack 
- Make sure, that negative bonuses from town work properly against enemy attacker only
- Get negative artifacts (shaman's thing, dead head and horseshoe), attack enemy neutral
- Make sure, that negative bonuses from artifacts work against neutrals only. Win battle
- Attack enemy hero. Make sure, that negative bonuses from artifacts work against enemy army only. 
- Choose hero with negative ability, attack enemy with this hero.
- Make sure, negative hero ability work properly against neutrals, enemy armies, town garrisons.
- Make sure, opposite bonuses without value (such as NO_MORALE) work properly against enemy in / outside town battles
- Make sure, that negative primary skills artifacts don't cause primary skill value to be less than valid minimum (0 and 1 for SP and Knowledge)
- Make sure Magic Well works properly (one time in day per hero)
- Make sure, negative artifacts work properly in the old saves
- Make sure, creatures with negative bonuses work properly in the old saves
- Make sure, negative hero's abilities work properly in the old saves
- Make sure, town buildings with  negative bonuses work properly in the old saves

### DIFF
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -558,7 +558,6 @@ int CClient::sendRequest(const CPackForServer * request, PlayerColor player)
 
 void CClient::battleStarted(const BattleInfo * info)
 {
-	BattleInfo::adjustOppositeBonuses(const_cast<BattleInfo *>(info));
 	setBattle(info);
 
 	for(auto & battleCb : battleCallbacks)
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -558,7 +558,9 @@ int CClient::sendRequest(const CPackForServer * request, PlayerColor player)
 
 void CClient::battleStarted(const BattleInfo * info)
 {
+	BattleInfo::adjustOppositeBonuses(const_cast<BattleInfo *>(info));
 	setBattle(info);
+
 	for(auto & battleCb : battleCallbacks)
 	{
 		if(vstd::contains_if(info->sides, [&](const SideInBattle& side) {return side.color == battleCb.first; })
--- a/client/widgets/MiscWidgets.cpp
+++ b/client/widgets/MiscWidgets.cpp
@@ -404,18 +404,22 @@ void MoraleLuckBox::set(const IBonusBearer * node)
 		text += "\n" + noLuck->Description();
 		bonusValue = 0;
 	}
-	else if(modifierList->empty())
-		text += CGI->generaltexth->arraytxt[noneTxtId];//no modifiers
 	else
 	{
-		for(auto& elem : *modifierList)
-		{
-			if(elem->val != 0)
-				//no bonuses with value 0
-				text += "\n" + elem->Description();
-		}
-	}
+		bool isListActual = false;
+		std::string addInfo = "";
 
+		for(auto & bonus : * modifierList)
+		{
+			if(bonus->val && bonus->effectRange != Bonus::ONLY_ENEMY_ARMY)
+			{
+				addInfo += "\n" + bonus->Description();
+				isListActual = true;
+			}
+		}
+		text = isListActual ? text + addInfo
+			: text + CGI->generaltexth->arraytxt[noneTxtId];//no modifiers
+	}
 	std::string imageName;
 	if (small)
 		imageName = morale ? "IMRL30": "ILCK30";
--- a/client/widgets/MiscWidgets.cpp
+++ b/client/widgets/MiscWidgets.cpp
@@ -411,7 +411,7 @@ void MoraleLuckBox::set(const IBonusBearer * node)
 
 		for(auto & bonus : * modifierList)
 		{
-			if(bonus->val && bonus->effectRange != Bonus::ONLY_ENEMY_ARMY)
+			if(bonus->val)
 			{
 				addInfo += "\n" + bonus->Description();
 				isListActual = true;
--- a/client/widgets/MiscWidgets.cpp
+++ b/client/widgets/MiscWidgets.cpp
@@ -406,19 +406,15 @@ void MoraleLuckBox::set(const IBonusBearer * node)
 	}
 	else
 	{
-		bool isListActual = false;
 		std::string addInfo = "";
-
 		for(auto & bonus : * modifierList)
 		{
 			if(bonus->val)
-			{
 				addInfo += "\n" + bonus->Description();
-				isListActual = true;
-			}
 		}
-		text = isListActual ? text + addInfo
-			: text + CGI->generaltexth->arraytxt[noneTxtId];//no modifiers
+		text = addInfo.empty() 
+			? text + CGI->generaltexth->arraytxt[noneTxtId] 
+			: text + addInfo;
 	}
 	std::string imageName;
 	if (small)
--- a/lib/CCreatureHandler.cpp
+++ b/lib/CCreatureHandler.cpp
@@ -392,6 +392,20 @@ void CCreature::fillWarMachine()
 	warMachine = ArtifactID::NONE; //this creature is not artifact
 }
 
+void CCreature::updateOppositeBonuses()
+{
+	auto & bonusList = getExportedBonusList();
+	for(auto & bonus : bonusList)
+	{
+		if(bonus->effectRange == Bonus::ONLY_ENEMY_ARMY //Opposite Side bonuses should not be exported from CREATURE node.
+			|| (bonus->propagator && bonus->propagator->getPropagatorType() == CBonusSystemNode::BATTLE))
+		{
+			bonus->effectRange == Bonus::ONLY_ENEMY_ARMY;
+			bonus->propagator.reset();
+		}
+	}
+}
+
 CCreatureHandler::CCreatureHandler()
 	: expAfterUpgrade(0)
 {
--- a/lib/CCreatureHandler.cpp
+++ b/lib/CCreatureHandler.cpp
@@ -497,13 +497,6 @@ void CCreatureHandler::loadBonuses(JsonNode & creature, std::string bonuses)
 		node["propagator"].String() = "HERO";
 		creature["abilities"]["const_raises_morale"] = node;
 	}
-	if(hasAbility("const_lowers_morale"))
-	{
-		JsonNode node = makeBonusNode("MORALE");
-		node["val"].Float() = -1;
-		node["effectRange"].String() = "ONLY_ENEMY_ARMY";
-		creature["abilities"]["const_lowers_morale"] = node;
-	}
 }
 
 std::vector<JsonNode> CCreatureHandler::loadLegacyData(size_t dataSize)
@@ -870,6 +863,10 @@ void CCreatureHandler::loadCreatureJson(CCreature * creature, const JsonNode & c
 			if (!ability.second.isNull())
 			{
 				auto b = JsonUtils::parseBonus(ability.second);
+
+				if(b->effectRange == Bonus::ONLY_ENEMY_ARMY) //Opposite Side bonuses should not be exported from CREATURE node.
+					b->propagator.reset();
+
 				b->source = Bonus::CREATURE_ABILITY;
 				b->sid = creature->getIndex();
 				b->duration = Bonus::PERMANENT;
--- a/lib/CCreatureHandler.cpp
+++ b/lib/CCreatureHandler.cpp
@@ -1322,14 +1322,20 @@ CreatureID CCreatureHandler::pickRandomMonster(CRandomGenerator & rand, int tier
 	return CreatureID(r);
 }
 
-void CCreatureHandler::addBonusForTier(int tier, std::shared_ptr<Bonus> b)
+void CCreatureHandler::addBonusForTier(int tier, const std::shared_ptr<Bonus> & b)
 {
 	assert(vstd::iswithin(tier, 1, 7));
 	creaturesOfLevel[tier].addNewBonus(b);
 }
 
-void CCreatureHandler::addBonusForAllCreatures(std::shared_ptr<Bonus> b)
+void CCreatureHandler::addBonusForAllCreatures(const std::shared_ptr<Bonus> & b)
 {
+	const auto & exportedBonuses = allCreatures.getExportedBonusList();
+	for(const auto & bonus : exportedBonuses)
+	{
+		if(bonus->type == b->type && bonus->subtype == b->subtype)
+			return;
+	}
 	allCreatures.addNewBonus(b);
 }
 
--- a/lib/CCreatureHandler.cpp
+++ b/lib/CCreatureHandler.cpp
@@ -392,20 +392,6 @@ void CCreature::fillWarMachine()
 	warMachine = ArtifactID::NONE; //this creature is not artifact
 }
 
-void CCreature::updateOppositeBonuses()
-{
-	auto & bonusList = getExportedBonusList();
-	for(auto & bonus : bonusList)
-	{
-		if(bonus->effectRange == Bonus::ONLY_ENEMY_ARMY //Opposite Side bonuses should not be exported from CREATURE node.
-			|| (bonus->propagator && bonus->propagator->getPropagatorType() == CBonusSystemNode::BATTLE))
-		{
-			bonus->effectRange == Bonus::ONLY_ENEMY_ARMY;
-			bonus->propagator.reset();
-		}
-	}
-}
-
 CCreatureHandler::CCreatureHandler()
 	: expAfterUpgrade(0)
 {
@@ -877,10 +863,6 @@ void CCreatureHandler::loadCreatureJson(CCreature * creature, const JsonNode & c
 			if (!ability.second.isNull())
 			{
 				auto b = JsonUtils::parseBonus(ability.second);
-
-				if(b->effectRange == Bonus::ONLY_ENEMY_ARMY) //Opposite Side bonuses should not be exported from CREATURE node.
-					b->propagator.reset();
-
 				b->source = Bonus::CREATURE_ABILITY;
 				b->sid = creature->getIndex();
 				b->duration = Bonus::PERMANENT;
--- a/lib/CCreatureSet.cpp
+++ b/lib/CCreatureSet.cpp
@@ -630,38 +630,10 @@ void CStackInstance::setType(CreatureID creID)
 		setType((const CCreature*)nullptr);
 }
 
-
-void CStackInstance::copyOppositeBonusesFromCreature(const CCreature * creature)
-{
-	auto owner = _armyObj ? _armyObj->getOwner() : PlayerColor::NEUTRAL;
-
-	if(owner == PlayerColor::UNFLAGGABLE)
-		owner = PlayerColor::NEUTRAL;
-
-	auto & creatureBonuses = const_cast<CCreature *>(type)->getExportedBonusList();
-
-	for(auto & creatureBonus : creatureBonuses)
-	{
-		if(creatureBonus->effectRange == Bonus::ONLY_ENEMY_ARMY) //it has better prformance than dynamic_cast
-		{
-			auto bCopy = std::make_shared<Bonus>(*creatureBonus);
-			bCopy->propagator.reset(new CPropagatorNodeType(CBonusSystemNode::BATTLE));
-			bCopy->limiter.reset(new OppositeSideLimiter(owner));
-			this->addNewBonus(bCopy);
-		}
-	}
-}
-
-void CStackInstance::removeOppositeBonuses()
-{
-	removeBonuses(Selector::effectRange()(Bonus::ONLY_ENEMY_ARMY));
-}
-
 void CStackInstance::setType(const CCreature *c)
 {
 	if(type)
 	{
-		removeOppositeBonuses();
 		detachFrom(const_cast<CCreature*>(type));
 		if (type->isMyUpgrade(c) && VLC->modh->modules.STACK_EXP)
 			experience = static_cast<TExpType>(experience * VLC->creh->expAfterUpgrade / 100.0);
@@ -670,11 +642,7 @@ void CStackInstance::setType(const CCreature *c)
 	CStackBasicDescriptor::setType(c);
 
 	if(type)
-	{
-		auto creature = const_cast<CCreature*>(type);
-		copyOppositeBonusesFromCreature(creature);
-		attachTo(creature);
-	}
+		attachTo(const_cast<CCreature*>(type));
 }
 std::string CStackInstance::bonusToString(const std::shared_ptr<Bonus>& bonus, bool description) const
 {
@@ -700,28 +668,9 @@ void CStackInstance::setArmyObj(const CArmedInstance * ArmyObj)
 		detachFrom(const_cast<CArmedInstance*>(_armyObj));
 
 	_armyObj = ArmyObj;
+
 	if(ArmyObj)
-	{
-		auto owner = _armyObj->getOwner();
-
-		if(owner == PlayerColor::UNFLAGGABLE)
-			owner = PlayerColor::NEUTRAL;
-
-		auto & bonusList = getExportedBonusList();
-
-		for(auto & bonus : bonusList)
-		{
-			if(bonus->effectRange != Bonus::ONLY_ENEMY_ARMY)
-				continue;
-
-			auto limPtr = bonus->limiter.get();
-			if(limPtr)
-				static_cast<OppositeSideLimiter *>(limPtr)->owner = owner;
-			else
-				logGlobal->error("setArmyObj. Limiter has been lost. Creature is %s", type ? type->nodeShortInfo() : std::string("No creature"));
-		}
 		attachTo(const_cast<CArmedInstance*>(_armyObj));
-	}
 }
 
 std::string CStackInstance::getQuantityTXT(bool capitalized) const
@@ -762,6 +711,11 @@ std::string CStackInstance::nodeName() const
 		oss << "[UNDEFINED TYPE]";
 
 	return oss.str();
+}
+
+PlayerColor CStackInstance::getOwner() const
+{
+	return _armyObj ? _armyObj->getOwner() : PlayerColor::NEUTRAL;
 }
 
 void CStackInstance::deserializationFix()
--- a/lib/CCreatureSet.cpp
+++ b/lib/CCreatureSet.cpp
@@ -630,10 +630,38 @@ void CStackInstance::setType(CreatureID creID)
 		setType((const CCreature*)nullptr);
 }
 
+
+void CStackInstance::copyOppositeBonusesFromCreature(const CCreature * creature)
+{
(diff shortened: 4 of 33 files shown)


lib/HeroBonus.cpp
```diff
@@ -427,6 +427,9 @@ int BonusList::totalValue() const
 
 	for(std::shared_ptr<Bonus> b : bonuses)
 	{
+		if(b->effectRange == Bonus::ONLY_ENEMY_ARMY)
```
alexvins, line 430: This only fixes bonuses with value, but what about things like NO_MORALE

lib/HeroBonus.cpp
```diff
@@ -888,10 +899,12 @@ void CBonusSystemNode::getBonusesRec(BonusList &out, const CSelector &selector,
 void CBonusSystemNode::getAllBonusesRec(BonusList &out) const
 {
 	BonusList beforeUpdate;
-	FOREACH_CPARENT(p)
```
alexvins, line 891: Function above looks obsolete, it should be removed along with FOREACH_CPARENT macro or changed accordingly to this one.

alexvins (changes requested):

lib/HeroBonus.cpp
ShubusCorporation: Done

lib/HeroBonus.cpp
ShubusCorporation: Also, `anaffectedByMoraleSelector` has been fixed.

alexvins: It was just an example to show that you need to ignore not only bonus value calculation but also presence of bonus.

alexvins: You should check `IBonusBearer::hasBonus`, also in `IBonusBearer::hasBonus(const CSelector &selector, const CSelector &limit, const std::string &cachingStr)`  overload second parameter `limit` is specifically for effectRange ,

alexvins (changes requested):

lib/HeroBonus.cpp
ShubusCorporation: As far as I understand, you propose add 'effectRange' processing to the 'hasBonus' code to hide presence of bonuses intended for enemy army.  It would be some kind of hack and it's out of this PR scope. 

'effectRange' processing is related to the game business logic of a higher level, than the bonus propagation and bonus limiters logic. The bonus limiters work at the level of node tree access, and the 'effectRange' works at the level of the game business logic. So, no need to mix high-level game logic and low-level node tree data access logic. 'hasBonus' methods are a kind of data access API. Business logic should have no place in the data access code. 

Additionally, "effectRange = ONLY_ENEMY_ARMY" doesn't make sense for most of the bonuses without value, so problem you propose to solve is not actual at all.

alexvins: "effectRange = ONLY_ENEMY_ARMY" is a hack on its own. It should be implemented with propagation (BATTLE_WIDE propagator and a limiter for opposite side of bonus source - the tricky part nobody managed to implement yet)

ShubusCorporation: I am inclined to conclude, that this does not have a natural solution within the framework of the bonus system. 

The creature bonus affects all players. Therefore, when creating a stack with a given creature, the bonus from the creature must be customized with information about the current owner. At this stage, we should create a copy of the creature bonus and add information about the current owner to this copy to avoid affecting other players. In this implementation, the bonus on the creature is idle and serves only to create a real bonus. It is better if that idle bonus from creature would not have a propagator, otherwise the BATTLE node will receive an extra bonus from creature node with owner=neutral.

Thus, we get the same two-stage propagation scheme. In this case, in JSON, it would still be better to use effectRange with an empty propagator and then create the desired propagator with the desired limiter. Btw, effectRange will have to be supported anyway for backward compatibility reasons.

On the other hand, there is a separate issue related to the  bonus propagation from the city to the visitor or to the defending hero, which is solved in a different way.

Thus, the logic of the solution becomes blurred, the benefits are not visible, no improvements are foreseen for modders (the corresponding record will be the same or even longer).

And the architectural reasons for this are obvious - an attempt to bring the high-level concept of "current owner - player" to the low level of the node tree and the distribution of bonuses.

ShubusCorporation: So, @alexvins , what do you think about it?

lib/HeroBonus.cpp
```diff
@@ -1269,15 +1338,26 @@ void CBonusSystemNode::removedRedDescendant(CBonusSystemNode *descendant)
 		if(b->propagator)
 			descendant->unpropagateBonus(b);
 
-	FOREACH_RED_PARENT(parent)
-		parent->removedRedDescendant(descendant);
+	TNodes redParents;
+	getRedAncestors(redParents); //get all red parents recursively
```
alexvins, line 1342: Why `newRedDescendant` have special case for TOWN but this function not?

lib/battle/BattleInfo.cpp
```diff
@@ -183,6 +183,44 @@ struct RangeGenerator
...
+	auto & bonusList = node->getExportedBonusList();
+	if(bonusList.empty())
+		return;
+
+	for(const auto & bonus : bonusList)
+	{
+		if(bonus->effectRange != Bonus::ONLY_ENEMY_ARMY)
+			continue;
+
+		static_cast<OppositeSideLimiter *>(bonus->limiter.get())->owner = ownerColor;
```
alexvins, line 197: We should not crash in case of no limiter here.

lib/battle/BattleInfo.h
```diff
@@ -131,13 +136,16 @@ class DLL_LINKAGE BattleInfo : public CBonusSystemNode, public CBattleInfoCallba
...
+	static void adjustOppositeBonuses(BattleInfo * curB);
 	static BattleInfo * setupBattle(int3 tile, ETerrainType terrain, BFieldType battlefieldType, const CArmedInstance * armies[2], const CGHeroInstance * heroes[2], bool creatureBank, const CGTownInstance * town);
 
 	ui8 whatSide(PlayerColor player) const;
 
 	static BattlefieldBI::BattlefieldBI battlefieldTypeToBI(BFieldType bfieldType); //converts above to ERM BI format
 	static int battlefieldTypeToTerrain(int bfieldType); //converts above to ERM BI format
+
+private:
+	STRONG_INLINE static void adjustOppositeBonuses(CBonusSystemNode * node, PlayerColor ownerColor);
```
alexvins, line 148: Is this STRONG_INLINE modifier necessary?

lib/registerTypes/RegisterTypes.h
```diff
@@ -388,6 +388,7 @@ void registerTypes(Serializer &s)
 	registerTypesClientPacks2(s);
 	registerTypesServerPacks(s);
 	registerTypesLobbyPacks(s);
+	s.template registerType<ILimiter, OppositeSideLimiter>();
```
alexvins, line 391: This should be before all NetPacks (in `registerTypesMapObjectTypes`)

alexvins (changes requested):

lib/battle/BattleInfo.h
ShubusCorporation: It's proposed to be called in the for() many times

lib/battle/BattleInfo.cpp
ShubusCorporation: Case of no limiter here is a weird error. So, check and log are added

lib/battle/BattleInfo.h
dydzio0614: It's plain wrong - strong inline should only be used if you know very well what you are doing and only if detailed profiling tells you to do so. Using it mean you know better than compiler how to optimize this case. 99% of the time you will want to just use "inline" or no modifier at all

lib/registerTypes/RegisterTypes.h
ShubusCorporation: Ok

lib/HeroBonus.cpp
ShubusCorporation: That case should be only applicable for battles (rewritten), so on. Battle object will be deleted.

lib/CCreatureSet.cpp
```diff
@@ -659,14 +659,33 @@ std::string CStackInstance::bonusToGraphics(const std::shared_ptr<Bonus>& bonus)
...
+			{
+				if(creatureBonus->effectRange == Bonus::ONLY_ENEMY_ARMY) //it has better prformance than dynamic_cast
+				{
+					auto bCopy = std::make_shared<Bonus>(* creatureBonus);
+					bCopy->propagator.reset(new CPropagatorNodeType(CBonusSystemNode::BATTLE));
+					bCopy->limiter.reset(new OppositeSideLimiter(owner));
+					this->addNewBonus(bCopy);
+				}
+			}
+		}
```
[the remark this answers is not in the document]
ShubusCorporation: Fixed. Copied/cleared on setType and then perming owner adjustment on setArmy, because only army has owner.

lib/CCreatureSet.cpp
```diff
@@ -659,14 +659,33 @@ std::string CStackInstance::bonusToGraphics(const std::shared_ptr<Bonus>& bonus)
...
 {
 	if(_armyObj)
 		detachFrom(const_cast<CArmedInstance*>(_armyObj));
 
 	_armyObj = ArmyObj;
 	if(ArmyObj)
 	{
+		if(type)
+		{
+			auto & creatureBonuses = const_cast<CCreature *>(type)->getExportedBonusList();
```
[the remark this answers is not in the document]
ShubusCorporation: Opposite Bonus from creature should not have propagator, corresponding opposite bonus should be exported from Stack Instance because creature bonuses are shared between all players and can't have owner.
How 'attachTo' is  related here - unclear.

lib/battle/BattleInfo.h
ShubusCorporation: What are particular objections? All that code moved to the separate function only for readability and I don't want to waste resources for  some kind of  push / call / pop / ret assembler instructions.

alexvins (approved):

lib/HeroBonus.cpp
alexvins: I do not like the two-stage scheme, but see no more issues preventing it to be merged.

lib/battle/BattleInfo.h
dydzio0614: objection is about using STRONG_INLINE - skip it unless you know what you're doing and did performance profiling compared to build with compiler optimization on

ShubusCorporation: In the case of principal approach, release build needs to be checked for all supported platforms, there are no resources for this

lib/battle/BattleInfo.h
```diff
@@ -131,13 +136,16 @@ class DLL_LINKAGE BattleInfo : public CBonusSystemNode, public CBattleInfoCallba
...
+	static void adjustOppositeBonuses(BattleInfo * curB);
 	static BattleInfo * setupBattle(int3 tile, ETerrainType terrain, BFieldType battlefieldType, const CArmedInstance * armies[2], const CGHeroInstance * heroes[2], bool creatureBank, const CGTownInstance * town);
 
 	ui8 whatSide(PlayerColor player) const;
 
 	static BattlefieldBI::BattlefieldBI battlefieldTypeToBI(BFieldType bfieldType); //converts above to ERM BI format
 	static int battlefieldTypeToTerrain(int bfieldType); //converts above to ERM BI format
+
+private:
+	inline static void adjustOppositeBonuses(CBonusSystemNode * node, PlayerColor ownerColor);
```
ShubusCorporation, line 148: Replaced from STRONG_INLINE to the inline

ShubusCorporation:
Test steps and usage tip are added to the description

ShubusCorporation:
Build is ready.

ShubusCorporation:
So, how about merge it?

lib/CCreatureSet.h
```diff
@@ -95,6 +95,10 @@ class DLL_LINKAGE CStackInstance : public CBonusSystemNode, public CStackBasicDe
 	ArtBearer::ArtBearer bearerType() const override; //from CArtifactSet
 	virtual std::string nodeName() const override; //from CBonusSystemnode
 	void deserializationFix();
+
+private:
+	void copyOppositeBonusesFromCreature(const CCreature * creature);
+	STRONG_INLINE void removeOppositeBonuses();
```
dydzio0614, line 101: strong inline should only be used in extreme cases where you are sure compiler won't optimize better than you and there is real performance impact so maybe it's better to remove it if you are unsure about using it

ShubusCorporation:
So, @dydzio0614 , @alexvins , are there  any objections on  merging it?

lib/HeroBonus.cpp
```diff
@@ -829,7 +830,47 @@ std::shared_ptr<const Bonus> IBonusBearer::getBonus(const CSelector &selector) c
...
+	case CBonusSystemNode::STACK_BATTLE:
+		return static_cast<const CStack*>(node);
+	default:
+		return nullptr;
+	}
+}
+
+const CStackInstance * retrieveStackInstance(const CBonusSystemNode * node)
+{
+	switch (node->getNodeType())
```
alexvins, line 846:
```suggestion
	switch(node->getNodeType())

```

lib/HeroBonus.cpp
```diff
@@ -829,7 +830,47 @@ std::shared_ptr<const Bonus> IBonusBearer::getBonus(const CSelector &selector) c
...
+	if (!node)
+		return PlayerColor::CANNOT_DETERMINE;
+
+	const CStack * stack = retrieveStackBattle(node);
+	if (stack)
+		return stack->owner;
+
+	const CStackInstance * csi = retrieveStackInstance(node);
+	if (csi && csi->armyObj)
+		return csi->armyObj->getOwner();
```
alexvins, line 868:
```suggestion
	if(!node)
		return PlayerColor::CANNOT_DETERMINE;

	const CStack * stack = retrieveStackBattle(node);
	if(stack)
		return stack->owner;

	const CStackInstance * csi = retrieveStackInstance(node);
	if(csi && csi->armyObj)
		return csi->armyObj->getOwner();

```

lib/HeroBonus.cpp
```diff
@@ -1253,14 +1310,37 @@ void CBonusSystemNode::getRedChildren(TNodes &out)
...
+		}
+	}
+	for(auto parent : redParents)
+	{
+		for(auto b : parent->exportedBonuses)
+		{
+			if(b->propagator)
+				descendant->propagateBonus(b);
+		}
+	}
```
alexvins, line 1343:
```suggestion
	if(descendant->nodeType == BATTLE && nodeType == TOWN) //acts as a pure bonus source, but has bonus bearers.
	{
		for(auto child : children)
			for(auto b : child->exportedBonuses)
				if(b->propagator && b->propagator->getPropagatorType() == BATTLE)
					descendant->propagateBonus(b);
	}
	for(auto parent : redParents)
	{
		for(auto b : parent->exportedBonuses)
			if(b->propagator)
				descendant->propagateBonus(b);
	}

```

lib/HeroBonus.cpp
```diff
@@ -1269,15 +1349,26 @@ void CBonusSystemNode::removedRedDescendant(CBonusSystemNode *descendant)
...
-		parent->removedRedDescendant(descendant);
+	TNodes redParents;
+	getRedAncestors(redParents); //get all red parents recursively
+
+	for (auto parent : redParents)
+	{
+		for (auto b : parent->exportedBonuses)
+			if (b->propagator)
+				descendant->unpropagateBonus(b);
+	}
```
alexvins, line 1360:
```suggestion
	for(auto parent : redParents)
	{
		for(auto b : parent->exportedBonuses)
			if(b->propagator)
				descendant->unpropagateBonus(b);
	}

```

alexvins (changes requested):
Just a few formatting issues

lib/HeroBonus.cpp
ShubusCorporation: Changes have been merged to commit fa2649c70215ab81ac5f2c0a6627ae420514ef96

alexvins (approved):

ShubusCorporation:
@dydzio0614, have you any objections on merging it?

nullkiller:
@ShubusCorporation please rebase on top of develop so that we have latest builds run

lib/CCreatureHandler.cpp
```diff
@@ -392,6 +392,20 @@ void CCreature::fillWarMachine()
...
 
+void CCreature::updateOppositeBonuses()
+{
+	auto & bonusList = getExportedBonusList();
+	for(auto & bonus : bonusList)
+	{
+		if(bonus->effectRange == Bonus::ONLY_ENEMY_ARMY //Opposite Side bonuses should not be exported from CREATURE node.
+			|| (bonus->propagator && bonus->propagator->getPropagatorType() == CBonusSystemNode::BATTLE))
+		{
+			bonus->effectRange == Bonus::ONLY_ENEMY_ARMY;
```
nullkiller, line 403: == ?

lib/CCreatureHandler.cpp
```diff
@@ -392,6 +392,20 @@ void CCreature::fillWarMachine()
 	warMachine = ArtifactID::NONE; //this creature is not artifact
 }
 
+void CCreature::updateOppositeBonuses()
+{
+	auto & bonusList = getExportedBonusList();
+	for(auto & bonus : bonusList)
+	{
+		if(bonus->effectRange == Bonus::ONLY_ENEMY_ARMY //Opposite Side bonuses should not be exported from CREATURE node.
+			|| (bonus->propagator && bonus->propagator->getPropagatorType() == CBonusSystemNode::BATTLE))
```
nullkiller, line 401: Could you clarify case when we have BATTLE propagator and not ONLY_ENEMY_ARMY range? And what will be in this case?

lib/CCreatureHandler.cpp
```diff
@@ -1322,14 +1333,20 @@ CreatureID CCreatureHandler::pickRandomMonster(CRandomGenerator & rand, int tier
 	return CreatureID(r);
 }
 
-void CCreatureHandler::addBonusForTier(int tier, std::shared_ptr<Bonus> b)
+void CCreatureHandler::addBonusForTier(int tier, const std::shared_ptr<Bonus> & b)
```
nullkiller, line 1336: Do you want to make const Bonus (std::shared_ptr<const Bonus>) or save some CPU to pass shared_ptr struct by ref instead of by value? Or am I missing something?

nullkiller (changes requested):

lib/CCreatureSet.cpp
```diff
@@ -662,14 +694,32 @@ std::string CStackInstance::bonusToGraphics(const std::shared_ptr<Bonus>& bonus)
...
+			owner = PlayerColor::NEUTRAL;
+
+		auto & bonusList = getExportedBonusList();
+
+		for(auto & bonus : bonusList)
+		{
+			if(bonus->effectRange != Bonus::ONLY_ENEMY_ARMY)
+				continue;
+
+			auto limPtr = bonus->limiter.get();
```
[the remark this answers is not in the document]
ShubusCorporation: Ok, but another limiter type should not be here,  it's  another theoretical error case.

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -604,28 +603,28 @@ TResources CGTownInstance::dailyIncome() const
 				buildingUpgrade = p2.first;
 			}
 		}
-
```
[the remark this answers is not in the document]
ShubusCorporation: on my look , there are enough semi-empty lines here

client/Client.cpp
```diff
@@ -558,7 +558,9 @@ int CClient::sendRequest(const CPackForServer * request, PlayerColor player)
 
 void CClient::battleStarted(const BattleInfo * info)
 {
+	BattleInfo::adjustOppositeBonuses(const_cast<BattleInfo *>(info));
```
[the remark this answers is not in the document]
ShubusCorporation: I think not worth it. At least this is how determined where info can be changed, and where it shouldn't be changed.
More cases when it may be changed increase probability of errors.

lib/CCreatureHandler.cpp
ShubusCorporation: I mean assignment here

lib/CCreatureHandler.cpp
ShubusCorporation: Because it calls 'addNewBonus' which has following signature:
`void CBonusSystemNode::addNewBonus(const std::shared_ptr<Bonus>& b)`

lib/CCreatureHandler.cpp
ShubusCorporation: If you have a look at  config and mods, other effectRange values don't need specific propagator.
In other hand, BATTLE propagator is used only for Opposite Side Bonuses.
Although, it will be work with another limiters, not only with OppositeSideLimiter, but it will be extra overhead.
Thus, BATTLE with not ONLY_ENEMY_ARMY range is nonsense and it will be overwritten here.
In my instruction I suggest modders to use ONLY_ENEMY_ARMY and don't touch BATTLE propagator.

nullkiller: non-leaving creature reducing all living creatures attack on the battlefield. Also how artifacts blocking magic on the battlefield work? Orb of inhibition if I am not mistaken? Also I suggest to at least print an error if we do override some propagator.

lib/CCreatureHandler.cpp
nullkiller: Ok.

client/widgets/MiscWidgets.cpp
```diff
@@ -404,18 +404,22 @@ void MoraleLuckBox::set(const IBonusBearer * node)
 		text += "\n" + noLuck->Description();
 		bonusValue = 0;
 	}
-	else if(modifierList->empty())
-		text += CGI->generaltexth->arraytxt[noneTxtId];//no modifiers
 	else
 	{
-		for(auto& elem : *modifierList)
+		bool isListActual = false;
```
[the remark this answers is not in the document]
nullkiller: Could we check addInfo != "" instead? It will be more natural and we no longer need a variable? or addInfo.empty() if it has one?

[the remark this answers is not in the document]
nullkiller: isListActual -> List links to modifierList and not to addInfo and this misleads a bit.

client/Client.cpp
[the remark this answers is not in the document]
nullkiller: I still think adjustOppositeBonuses can be an instance method and it can const_cast(this) inside to make required changes. This will be kind of incapsulation, BattleInfo decides that such change does not affect battle itself and can be treated as const. But I still want to take a bit more closer look onto all it.

lib/CCreatureHandler.cpp
ShubusCorporation: Obviously, they don't need to use ONLY_ENEMY_ARMY effect range or Opposite limiter and it's not current discussion theme.

nullkiller: You said BATTLE propagator is used only for Opposite Side Bonuses. I ask why cannot we use BATTLE propagator with creature type limiter or other bonus limiter?

nullkiller: Take into account that you use ONLY_ENEMY_ARMY OR BATTLE propagator, not AND. So any BATTLE propagator has to use OppositeSideLimiter, am I correct?

client/Client.cpp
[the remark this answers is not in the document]
ShubusCorporation: It's called from static setupBattle - that's why it's static.
adjustOppositeBonuses receives non-const  argument and should not do const_cast inside!

lib/CCreatureHandler.cpp
ShubusCorporation: Particular part of code, which you are reviewing is used to update creature data from old saves. So, it's written considering current mods data. I know  only one attempt to use this propagator in 3thd party mod and that attempt looks wrong, Thus, it's possible to say, that BATTLE_WIDE prop. is not used for creatures now.
Propagator checking here is for additional insurance.

P. S. Creature node has no owner - that's why it's not suitable to be holder of bonus with particular 'OppositeSide' limiter

nullkiller (approved):

PR #678, opened 2021-02-08
Fixes mantis tickets 2899 and 2984 (bugged hero spell specialties)

MikeLodz:
- stoneskin/haste/prayer/weakness: didnt work because there were 2 bonus objects in buffer and they were filtered out as a possible duplicate (BattleInfo::addOrUpdateUnitBonus). it was fixed by making them a single bonus.
- disrupting-ray and weakness: had opposite effect, because of missing negation
- added a new specialty types: SPECIAL_ADD_VALUE_ENCHANT and SPECIAL_FIXED_VALUE_ENCHANT. this is to make possible specialties like Aenin (fixed specialty value added to spell value) and Melody (fixed value for spell regardless of anything). These specialties can be used in mods with any heroes.
- slayer spell effect calculations was fixed to include hero Coronius-style specialty.
- finally fixed description for Labetha Conflux hero, this is a OH3 bug described here https://heroes.thelazy.net/index.php/Labetha
Changes were tested and work as intended.
commit was made in cooperation with modder Misiokles

### DIFF
--- a/lib/HeroBonus.h
+++ b/lib/HeroBonus.h
@@ -281,8 +281,6 @@ public:
 	BONUS_NAME(SPECIAL_BLESS_DAMAGE) /*val = spell (bless), additionalInfo = value per level in percent*/\
 	BONUS_NAME(MAXED_SPELL) /*val = id*/\
 	BONUS_NAME(SPECIAL_PECULIAR_ENCHANT) /*blesses and curses with id = val dependent on unit's level, subtype = 0 or 1 for Coronius*/\
-	BONUS_NAME(SPECIAL_ADD_VALUE_ENCHANT) /*specialty like Aenin, additionalInfo = value to add*/\
-	BONUS_NAME(SPECIAL_FIXED_VALUE_ENCHANT) /*specialty like Melody constant 3 luck, additionalInfo = value to fix.*/\
 	BONUS_NAME(SPECIAL_UPGRADE) /*subtype = base, additionalInfo = target */\
 	BONUS_NAME(DRAGON_NATURE) \
 	BONUS_NAME(CREATURE_DAMAGE)/*subtype 0 = both, 1 = min, 2 = max*/\
@@ -323,6 +321,8 @@ public:
 	BONUS_NAME(SPECIAL_CRYSTAL_GENERATION) /*crystal dragon crystal generation*/ \
 	BONUS_NAME(NO_SPELLCAST_BY_DEFAULT) /*spellcast will not be default attack option for this creature*/ \
 	BONUS_NAME(GARGOYLE) /* gargoyle is special than NON_LIVING, cannot be rised or healed */ \
+	BONUS_NAME(SPECIAL_ADD_VALUE_ENCHANT) /*specialty spell like Aenin has, increased effect of spell, additionalInfo = value to add*/\
+	BONUS_NAME(SPECIAL_FIXED_VALUE_ENCHANT) /*specialty spell like Melody has, constant spell effect (i.e. 3 luck), additionalInfo = value to fix.*/\
 
 	/* end of list */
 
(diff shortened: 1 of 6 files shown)


lib/spells/effects/Timed.cpp
```diff
@@ -199,23 +206,35 @@ void Timed::prepareEffects(SetStackEffect & sse, const Mechanics * m, const Effe
...
 				power = std::max(5 - tier, 0);
-				Bonus specialBonus(Bonus::N_TURNS, Bonus::PRIMARY_SKILL, Bonus::SPELL_EFFECT, power, m->getSpellIndex(), PrimarySkill::ATTACK);
-				specialBonus.turnsRemain = duration;
-				buffer.push_back(specialBonus);
 				break;
 			}
+			if(m->isNegativeSpell()) {
+				//negative spells like weakness are defined in json with negative numbers, so we need do same here
+				power = -1 * power;
+			}
```
alexvins, line 219:
```suggestion
			if(m->isNegativeSpell()) 
			{
				//negative spells like weakness are defined in json with negative numbers, so we need do same here
				power = -1 * power;
			}
```

lib/HeroBonus.h
```diff
@@ -281,6 +281,8 @@ class DLL_LINKAGE CAddInfo : public std::vector<si32>
 	BONUS_NAME(SPECIAL_BLESS_DAMAGE) /*val = spell (bless), additionalInfo = value per level in percent*/\
 	BONUS_NAME(MAXED_SPELL) /*val = id*/\
 	BONUS_NAME(SPECIAL_PECULIAR_ENCHANT) /*blesses and curses with id = val dependent on unit's level, subtype = 0 or 1 for Coronius*/\
+	BONUS_NAME(SPECIAL_ADD_VALUE_ENCHANT) /*specialty like Aenin, additionalInfo = value to add*/\
+	BONUS_NAME(SPECIAL_FIXED_VALUE_ENCHANT) /*specialty like Melody constant 3 luck, additionalInfo = value to fix.*/\
```
alexvins, line 285: New bonus types must be added at end of list

alexvins (changes requested):

lib/spells/effects/Timed.cpp
```diff
@@ -175,16 +181,17 @@ void Timed::prepareEffects(SetStackEffect & sse, const Mechanics * m, const Effe
 		if(describe)
 			describeEffect(sse.battleLog, m, converted, affected);
 
-		si32 power = 0;
+		const auto tier = std::max(affected->creatureLevel(), 1); //don't divide by 0 for certain creatures (commanders, war machines)
 
 		//Apply hero specials - peculiar enchants
-		const auto tier = std::max(affected->creatureLevel(), 1); //don't divide by 0 for certain creatures (commanders, war machines)
-		if(bonus)
+		if(peculiarBonus)
```
nullkiller, line 187: Does if/elseif structure means that we can have only one bonus either peculiar or additive or fixed? Because now if both peculiar and additive bonuses exist we will have only peculiar applied

lib/spells/effects/Timed.cpp
MikeLodz: fixed in commit #2

lib/spells/effects/Timed.cpp
MikeLodz: At first I didnt test combination of both bonuses, so made an "else" clause to make sure nothing bad happens.
But since you pointed this out, I changed it, tested it and it works: gave Labetha both bonuses in config, and they added to each other.

lib/HeroBonus.h
MikeLodz: No problem, you are the master here, fixed in commit #2.

nullkiller (approved):

alexvins (approved):

PR #636, opened 2020-03-31
Starting to add some UI improvements from HD mod (dead code)

ars3niy:
Display some new buttons (doing nothing) like in the HD mod in hero exchange dialog when certain additional resource file are present. Those resource files could later be added into VCMI essential add-on, but since no one is going to have them now, all this is currently dead code.

PR #680, opened 2021-02-11
Hero exchange dialog improvement

nullkiller:
The idea of this PR is to implement hero exchange dialog from HD mod. Now it contains only army exchange buttons working and I plan to implement artifact exchange buttons also. This thing can be merged now or we can wait for the rest of changes.

- [x] pick changes from @ars3niy (https://github.com/vcmi/vcmi/pull/636)
- [x] acquire resources
- [x] implement hero exchange
- [x] implement artifact exchange

### DIFF
--- a/client/widgets/Images.cpp
+++ b/client/widgets/Images.cpp
@@ -288,10 +288,18 @@ void CAnimImage::showAll(SDL_Surface * to)
 	if(!visible)
 		return;
 
-	if(flags & CShowableAnim::BASE && frame != 0)
-		if(auto img = anim->getImage(0, group))
+	std::vector<size_t> frames = {frame};
+
+	if((flags & CShowableAnim::BASE) && frame != 0)
+	{
+		frames.insert(frames.begin(), 0);
+	}
+
+	for(auto targetFrame : frames)
+	{
+		if(auto img = anim->getImage(targetFrame, group))
 		{
-			if (isScaled())
+			if(isScaled())
 			{
 				auto scaled = img->scaleFast(float(scaledSize.x) / img->width());
 				scaled->draw(to, pos.x, pos.y);
@@ -299,16 +307,6 @@ void CAnimImage::showAll(SDL_Surface * to)
 			else
 				img->draw(to, pos.x, pos.y);
 		}
-
-	if(auto img = anim->getImage(frame, group))
-	{
-		if (isScaled())
-		{
-			auto scaled = img->scaleFast(float(scaledSize.x) / img->width());
-			scaled->draw(to, pos.x, pos.y);
-		}
-		else
-			img->draw(to, pos.x, pos.y);
 	}
 }
 
--- a/client/widgets/Images.cpp
+++ b/client/widgets/Images.cpp
@@ -228,9 +228,43 @@ CAnimImage::CAnimImage(std::shared_ptr<CAnimation> Anim, size_t Frame, size_t Gr
 	init();
 }
 
+CAnimImage::CAnimImage(std::shared_ptr<CAnimation> Anim, size_t Frame, Rect targetPos, size_t Group, ui8 Flags):
+	anim(Anim),
+	frame(Frame),
+	group(Group),
+	player(-1),
+	flags(Flags),
+	scaledSize(targetPos.w, targetPos.h)
+{
+	pos.x += targetPos.x;
+	pos.y += targetPos.y;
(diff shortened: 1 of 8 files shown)


client/windows/GUIClasses.cpp
```diff
@@ -787,9 +787,95 @@ void CTavernWindow::HeroPortrait::hover(bool on)
...
+
+std::function<void()> CExchangeController::onMoveArmyToRight()
+{
+	return [&]() { hdModQuickExchangeArmy(true); };
+}
+
+std::function<void()> CExchangeController::onSwapArmy()
+{
+	return [&]() { 
+		auto cb = LOCPLINT->cb;
```
alexvins, line 816:
```suggestion
	return [&]() 
	{ 
		auto cb = LOCPLINT->cb;

```

client/windows/GUIClasses.cpp
```diff
@@ -787,9 +787,95 @@ void CTavernWindow::HeroPortrait::hover(bool on)
...
+}
+
+void CExchangeController::hdModQuickExchangeArmy(bool leftToRight)
+{
+	auto source = leftToRight ? left : right;
+	auto target = leftToRight ? right : left;
+	auto cb = LOCPLINT->cb;
+
+	boost::thread([=]
+	{
```
alexvins, line 833: Please explain how threading helps here (but not in function above, both do server call for each slot).

alexvins (changes requested):

client/windows/GUIClasses.cpp
nullkiller: cb->waitTilRelaize locks main thread. For easier implementation here I need to see the results of cb invokations so I need to use waitTilRealize and thus I need a separate thread.

client/widgets/Images.cpp
```diff
@@ -259,10 +290,26 @@ void CAnimImage::showAll(SDL_Surface * to)
 
 	if(flags & CShowableAnim::BASE && frame != 0)
 		if(auto img = anim->getImage(0, group))
-			img->draw(to, pos.x, pos.y);
+		{
```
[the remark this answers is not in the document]
nullkiller: Sorry, I even have not seen it. Will take a closer look now

client/windows/GUIClasses.cpp
alexvins: The question is why is a new thread here but in `CExchangeController::onSwapArmy()` for same amount of `cb` calls is not? Also if you spawn a thread you need (at least shared) lock of gs there, but this in turn will still lock UI. And with acquired lock you could use unlockGsWhenWaiting.

nullkiller: Because I do not need the result of those calls in swapArmy

nullkiller: Btw I think you spotted a nice bug if the first stack for swapping is the last one and the same slot of another army is empty

nullkiller: Is it ok now?

client/widgets/Images.cpp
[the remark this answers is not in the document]
nullkiller: Fixed code duplication

client/windows/GUIClasses.cpp
alexvins: Yes it is

alexvins (approved):

alexvins (changes requested):

client/windows/GUIClasses.cpp
```diff
@@ -787,9 +787,363 @@ void CTavernWindow::HeroPortrait::hover(bool on)
...
+		boost::thread(std::bind(&GsThread::run, std::make_shared<GsThreadState>(action)));
+	}
+
+private:
+	static void run(std::shared_ptr<GsThreadState> state)
+	{
+		boost::shared_lock<boost::shared_mutex> gsLock(CGameState::mutex);
+
+		auto originalWaitTillRealize = state->cb->waitTillRealize;
```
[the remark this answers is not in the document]
nullkiller: Makes sense

nullkiller:
Refactored GsThread and now it is only left to update vcmi essentials or create a new separate hdmod. Anyway it works without HD mod resources, just uses old UI instead.

alexvins (approved):

PR #684, opened 2021-02-16
Fixes some tickets for Campaigns: 3095 cloned heroes, 3164 too high stats

MikeLodz:
Fixes some mantis tickets for Campaign game: 
3095 multiple heroes when switching to a new mission,
3164 Bug with primary skills when moving from mission to mission.

details:
CGameState: method removeHeroFromBothLists didnt work as there were different Hero objects and no exact match. Because this method didnt work, we had duplicated heroes on the map. It was replaced with deleting hero by its name instead.
CGHeroInstance: we didnt initialize level on loading hero from crossover, so the game did full levelups later, and bumped stats.
Changes were tested to work on transfer between 1st and 2nd scenario of Gelu SoD campaign.
Please note that this commit doesnt fix existing issues about transfering artifacts. Transfering artifacts is a more complex thing as if I recall correctly, we shouldnt just transfer every artifact, but only quest ones.

### DIFF
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -381,8 +381,22 @@ void CGameState::CrossoverHeroesList::addHeroToBothLists(CGHeroInstance * hero)
 
 void CGameState::CrossoverHeroesList::removeHeroFromBothLists(CGHeroInstance * hero)
 {
-	heroesFromPreviousScenario -= hero;
-	heroesFromAnyPreviousScenarios -= hero;
+	for (auto obj : heroesFromPreviousScenario)
+	{
+		if (obj->name == hero->name)
+		{
+			heroesFromPreviousScenario -= obj;
+			break;
+		}
+	}
+	for (auto obj : heroesFromAnyPreviousScenarios)
+	{
+		if (obj->name == hero->name) 
+		{
+			heroesFromAnyPreviousScenarios -= obj;
+			break;
+		}
+	}
 }
 
 CGameState::CampaignHeroReplacement::CampaignHeroReplacement(CGHeroInstance * hero, ObjectInstanceID heroPlaceholderId) : hero(hero), heroPlaceholderId(heroPlaceholderId)
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -1399,6 +1399,14 @@ void CGHeroInstance::serializeCommonOptions(JsonSerializeFormat & handler)
 {
 	handler.serializeString("biography", biography);
 	handler.serializeInt("experience", exp, 0);
+	if(!handler.saving)
+	{
+		while(gainsLevel())
+		{
+			++level;
+		}
+	}
+
 	handler.serializeString("name", name);
 	handler.serializeBool<ui8>("female", sex, 1, 0, 0xFF);
 


lib/CGameState.cpp
```diff
@@ -381,8 +381,22 @@ void CGameState::CrossoverHeroesList::addHeroToBothLists(CGHeroInstance * hero)
 
 void CGameState::CrossoverHeroesList::removeHeroFromBothLists(CGHeroInstance * hero)
 {
-	heroesFromPreviousScenario -= hero;
-	heroesFromAnyPreviousScenarios -= hero;
+	for (auto obj : heroesFromPreviousScenario)
```
alexvins, line 384: No space after `for` and `if` please.

alexvins (changes requested):

lib/CGameState.cpp
ShubusCorporation, line 384: "obj" is not informative. Could you rename?

lib/CGameState.cpp
```diff
@@ -381,8 +381,22 @@ void CGameState::CrossoverHeroesList::addHeroToBothLists(CGHeroInstance * hero)
...
+	{
+		if (obj->name == hero->name)
```
nullkiller, line 386: That looks a bit strange to use hero name as an identity. Do we have some hero type or hero identity or whatever? Somehow he knows where to take his icon, base stats name and so on.

ShubusCorporation: Also it's interesting, why we have different objects for same hero? Is it normal or some extra object coping somewhere?

MikeLodz: @ShubusCorporation these are hero objects that comes from json deseralization.
When Campaign scenario ends, the game serializes the heroes into json, and recreate Objects from json in next scenario.
Game takes Heroes from two sources: last map and previous maps. This is why we have 2 lists over here, and some overlapping heroes in both lists. It looks fine, we only need a good way to compare Heroes to say that this is a copy of the same hero. 
For me the hero name is simply "good enough" in our scenarios to find a hero, but no problem, we can look for some systemId instead.

ShubusCorporation: So, is `void CGameState::CrossoverHeroesList::addHeroToBothLists(CGHeroInstance * hero)` used anywhere or we should remove it?

lib/CGameState.cpp
```diff
@@ -381,8 +381,22 @@ void CGameState::CrossoverHeroesList::addHeroToBothLists(CGHeroInstance * hero)
...
+		{
+			heroesFromPreviousScenario -= obj;
+			break;
+		}
+	}
+	for (auto obj : heroesFromAnyPreviousScenarios)
+	{
+		if (obj->name == hero->name)
```
ShubusCorporation, line 394: The hero pointer here is a result of the following searching in the `heroesFromAnyPreviousScenarios` container:
`range::find_if(crossoverHeroes.heroesFromAnyPreviousScenarios, [heroPlaceholder](CGHeroInstance * hero)`
So,  is this condition really necessary for this loop?

lib/CGameState.cpp
alexvins: `subID` field is unique for heroes.

alexvins: there is also `getHeroTypeName()` method, it returns unique key for JSON serialization,

lib/CGameState.cpp
MikeLodz: If the method is called "removeFromBothList" it should search and delete from both lists, doing exact thing for both lists.
Maybe currently its callers dont need to search the second list, but method uses can change over time. And it will be difficult to next code readers to understand why we search first list, and dont search the second one.
When methods do their name, its more managable code.

MikeLodz:
> It looks like an answer to my second question:
> 

well, my motivation was simple: find if there is any simple bug to fix here. I saw the method deleteFromBothLists not doing its job, so I fixed it, tested and it works better.
But you are correct this code is strange, and I cannot guarantee it works in all scenarios.
It would be better to revise it and rewrite. Lets return to it at later time, after we deal with top priority tasks.

matejmosko:
> > It looks like an answer to my second question:
> 
> well, my motivation was simple: find if there is any simple bug to fix here. I saw the method deleteFromBothLists not doing its job, so I fixed it, tested and it works better.
> But you are correct this code is strange, and I cannot guarantee it works in all scenarios.
> It would be better to revise it and rewrite. Lets return to it at later time, after we deal with top priority tasks.

It's a pity to hear that. I was waiting for this commit to be merged before I finnish the mission.

PR #685, opened 2021-03-02
fix boost.locale crash on macOS-arm64

kambala-decapitator:
The fix is taken from https://github.com/SuperTux/supertux/pull/1663

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -191,7 +191,7 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #include <boost/format.hpp>
 #include <boost/functional/hash.hpp>
 #include <boost/lexical_cast.hpp>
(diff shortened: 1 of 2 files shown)


kambala-decapitator:
just thought that https://github.com/vcmi/vcmi/blob/develop/client/CMT.cpp#L169 might be related

ArseniyShestakov:
Need to test it on Linux just in case, but I guess for 99% of users UTF-8 is default anyway. Though who knows might be there some corner cases that make it not just Windows quirck.

kambala-decapitator:
if you think it's dangerous, the condition can be changed to exclude only ARM Macs

ArseniyShestakov:
Ok, finally tested it properly just in case.
Thanks for contribution!

PR #686, opened 2021-03-11
enable cotire for EmptyAI

kambala-decapitator:
If one doesn't build all available targets but only e.g. client, then without this `cmake --install` fails because EmptyAI lib doesn't exist.


kambala-decapitator:
please don't merge yet, looks like it's not a complete fix to build EmptyAI

kambala-decapitator:
forget it: `cmake --install` requires to build `install` target. Had to solve my issue within cmake config.

PR #687, opened 2021-03-20
fix server crash on unexpected client disconnect

kambala-decapitator:
Now server doesn't crash if client suddenly disconnects (e.g. because of crash).

Unfortunately, server doesn't accept new connections afterwards.

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -1195,11 +1195,11 @@ void CGameHandler::handleClientDisconnection(std::shared_ptr<CConnection> c)
 {
 	for(auto playerConns : connections)
 	{
-		for(auto conn : playerConns.second)
+		for(auto i = playerConns.second.begin(); i != playerConns.second.end(); )
 		{
-			if(lobby->state != EServerState::SHUTDOWN && conn == c)
+			if(lobby->state != EServerState::SHUTDOWN && *i == c)
 			{
-				vstd::erase_if_present(playerConns.second, conn);
+				i = playerConns.second.erase(i);
 				if(playerConns.second.size())
 					continue;
 				PlayerCheated pc;
@@ -1208,6 +1208,8 @@ void CGameHandler::handleClientDisconnection(std::shared_ptr<CConnection> c)
 				sendAndApply(&pc);
 				checkVictoryLossConditionsForPlayer(playerConns.first);
 			}
+			else
+				++i;
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


alexvins (approved):

server/CGameHandler.cpp
```diff
@@ -1195,11 +1195,11 @@ void CGameHandler::handleClientDisconnection(std::shared_ptr<CConnection> c)
 {
 	for(auto playerConns : connections)
 	{
-		for(auto conn : playerConns.second)
+		for(auto i = playerConns.second.begin(); i != playerConns.second.end(); )
```
ShubusCorporation, line 1198: All these ".first" and ".second" to me are hard to look at.  Would be great to refactor this code giving human readable names.

server/CGameHandler.cpp
```diff
@@ -1195,11 +1195,11 @@ void CGameHandler::handleClientDisconnection(std::shared_ptr<CConnection> c)
...
 		{
-			if(lobby->state != EServerState::SHUTDOWN && conn == c)
+			if(lobby->state != EServerState::SHUTDOWN && *i == c)
```
ShubusCorporation, line 1200: I have some questions here. 

- Do we really should check  `lobby->state` on each iteration?
- `playerConns.second` is a set of connections, so condition `conn == c` may be true only once per player, so, why there's no break? 
- Then we clear cheater status and then check victory condition. Is it logically, at all? May be, victory condition checking should be first, and then clear cheater status?

@alexvins , what do you think about?

nullkiller: Looks like according to https://github.com/vcmi/vcmi/commit/ac66fc7f42f50ad961cf8a0656be0d9b5415180d the initial version of this method was activated only for disconnected connection. I expect PlayerCheated is used to force player to loose using vcmimelkor cheat. And checkVictoryLossConditionsForPlayer was also activated for the lost player only to check if somebody else had won.

```
if(playerConns.second.size())
	continue;
```
looks wrong here. Probably @ArseniyShestakov can help here.

nullkiller: Also it feels like the method body should get activated only once so we can return after it is activated. Also it is interesting why do we need checkVictoryLossConditionsForPlayer at all. I expect the cheat code should activate this logic already somehow.

dydzio0614: there were and still are bugs introduced in pregame rework branch, due to time constraints SXX had and previous code being really bad, leading to "refactoring introducing unexpected bugs"

nullkiller: The idea is to improve it a bit if we are already here. Because now the intention of this code is unclear. At the same time we can say it is out of scope for this ticket because it aims to fix the crash caused by invalid collection usage.

PR #689, opened 2021-04-04
fix bug: may load mod before dependencies

kdmcser:
Commit [db1f9a1 ](https://github.com/vcmi/vcmi/commit/db1f9a15b0f67c59aacd7e0af41ef460f6455baf) disable topological sort，in some case a mod will lead mod loads before dependencies. If the mod replaces some object with the same in dependencies. Things go wrong. 

I just keep topological sort, put all mod after dependencies. If no more new mods to be be added  in a  topological sort, the sort should be ended. All left mod are broken and need to be logged.

### DIFF
--- a/lib/CModHandler.cpp
+++ b/lib/CModHandler.cpp
@@ -716,64 +716,65 @@ bool CModHandler::checkDependencies(const std::vector <TModID> & input) const
 	return true;
 }
 
-std::vector <TModID> CModHandler::resolveDependencies(std::vector <TModID> input) const
+// Returned vector affects the resource loaders call order (see CFilesystemList::load).
+// The loaders call order matters when dependent mod overrides resources in its dependencies.
+std::vector <TModID> CModHandler::validateAndSortDependencies(std::vector <TModID> modsToResolve) const
 {
-	// Topological sort algorithm
-	// May not be the fastest one but VCMI does not needs any speed here
-	// Unless user have dozens of mods with complex dependencies this code should be fine
+	// Topological sort algorithm.
+	// TODO: Investigate possible ways to improve performance.
+	boost::range::sort(modsToResolve); // Sort mods per name
+	std::vector <TModID> sortedValidMods; // Vector keeps order of elements (LIFO) 
+	sortedValidMods.reserve(modsToResolve.size()); // push_back calls won't cause memory reallocation
+	std::set <TModID> resolvedModIDs; // Use a set for validation for performance reason, but set does not keep order of elements
 
-	// first - sort input to have input strictly based on name (and not on hashmap or anything else)
-	boost::range::sort(input);
-
-	std::vector <TModID> output;
-	output.reserve(input.size());
-
-	std::set <TModID> resolvedMods;
-
-	// Check if all mod dependencies are resolved (moved to resolvedMods)
-	auto isResolved = [&](const CModInfo & mod) -> bool
+	// Mod is resolved if it has not dependencies or all its dependencies are already resolved
+	auto isResolved = [&](const CModInfo & mod) -> CModInfo::EValidationStatus
 	{
+		if(mod.dependencies.size() > resolvedModIDs.size())
+			return CModInfo::PENDING;
+
 		for(const TModID & dependency : mod.dependencies)
 		{
-			if (!vstd::contains(resolvedMods, dependency))
-				return false;
+			if(!vstd::contains(resolvedModIDs, dependency))
+				return CModInfo::PENDING;
 		}
-		return true;
+		return CModInfo::PASSED;
 	};
 
-	while (true)
+	while(true)
 	{
-		size_t lastResolvedCount = resolvedMods.size();
-		std::set <TModID> toResolve; // list of mods resolved on this iteration
-
-		for (auto it = input.begin(); it != input.end();)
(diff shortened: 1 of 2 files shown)


ShubusCorporation:
@kdmcser  , could you provide test samples - minimal set of modes, which causes described issue?

kdmcser:
> @kdmcser , could you provide test samples - minimal set of mods, which causes described issue?

1.  apply 3 mods: wog , Tides of War, high res menu
2.  make high res menu depends on  wog and tides of war(edit json, add depend field), it should replace the start menu.
3.  enter the game, it doest not replace two others.

my mod folder: https://drive.google.com/file/d/1GMom4bx0odqtIhvsbA8UvqqA5DGFkBHQ/view?usp=sharing

ShubusCorporation:
The problem exists and should be fixed. Thanks for your contribution. We are happy to welcome new developers.
First of all, let me introduce our code style : [](https://wiki.vcmi.eu/Coding_guidelines) (pay attention, there are no spaces after "if" and "for" and where to put braces). Unified code style is a standard practice in all large projects.

It should be noted that for the successful development and support our project, we need not just working code, but code that could be easily understood and supported by the developers who come after us. It should be clear why this code is needed and why it is written that way. Thus, no one else will remove it without considering all consequences.

Therefore, I suggest not just restoring the old code, but implement some refactoring and add some comments. It would be great if you update your PR in some kind like this:

Some changes in the header:
`	/**
	* 1. Set apart mods with resolved dependecies from mods which have unresolved dependecies
	* 2. Sort resolved mods using topological algorithm
	* 3. Log all problem mods and their unresolved dependencies
	*
	* @param modsToResolve list of valid mod IDs (checkDependencies returned true - TODO: Clarify it.)
	* @return a vector of the topologically sorted resolved mods: child nodes (dependent mods) have greater index than parents
	*/
	std::vector <TModID> validateAndSortDependencies(std::vector<TModID> modsToResolve) const;`

And in the .cpp:

`// Returned vector affects the resource loaders call order (see CFilesystemList::load).
// The loaders call order matters when dependent mod overrides resources in its dependencies.
std::vector <TModID> CModHandler::validateAndSortDependencies(std::vector <TModID> modsToResolve) const
{
	// Topological sort algorithm. 
    // TODO: Investigate possible ways to improve performance.

	boost::range::sort(modsToResolve); // Sort mods per name
	std::vector <TModID> sortedValidMods; // Vector keeps order of elements (LIFO) 
	sortedValidMods.reserve(modsToResolve.size()); // push_back calls won't cause memory reallocation
	std::set <TModID> resolvedModIDs; // Use a set for validation for performance reason, but set does not keep order of elements

	// Mod is resolved if it has not dependencies or all its dependecies are already resolved
	auto isResolved = [&](const CModInfo & mod) -> CModInfo::EValidationStatus
	{
		if(mod.dependencies.size() > resolvedModIDs.size())
			return CModInfo::PENDING;

		for(const TModID & dependency : mod.dependencies)
		{
			if(!vstd::contains(resolvedModIDs, dependency))
				return CModInfo::PENDING;
		}
		return CModInfo::PASSED;
	};

	while(true)
	{
		std::set <TModID> resolvedOnCurrentTreeLevel;
		for(auto it = modsToResolve.begin(); it != modsToResolve.end();) // One iteration - one level of mods tree
		{
			if(isResolved(allMods.at(*it)) == CModInfo::PASSED)
			{
				resolvedOnCurrentTreeLevel.insert(*it); // Not to the resolvedModIDs, so current node childs will be resolved on the next iteration
				sortedValidMods.push_back(*it);
				it = modsToResolve.erase(it);
				continue;
			}
			it++;
		}
		if(resolvedOnCurrentTreeLevel.size())
		{
			resolvedModIDs.insert(resolvedOnCurrentTreeLevel.begin(), resolvedOnCurrentTreeLevel.end());
		}
		else // If there're no valid mods on the current mods tree level, no more mod can be resolved, should be end.
		{
			break;
		}
	}
	// Left mods have unresolved dependencies, output all to log.
	for(const auto & brokenModID : modsToResolve)
	{
		const CModInfo & brokenMod = allMods.at(brokenModID);
		for(const TModID & dependency : brokenMod.dependencies)
		{
			if(!vstd::contains(resolvedModIDs, dependency))
				logMod->error("Mod '%s' will not work: it depends on mod '%s', which is not installed.", brokenMod.name, dependency);
		}
	}
	return sortedValidMods;
}`

Also, if the function will be renamed, call should be changed too. And check that all works.

ShubusCorporation:
Also, I surprised, we have interesting methods `checkDependencies` and `hasCircularDependency` which are not actually used. I guess, we will improve this system...

kdmcser:
@ShubusCorporation Thank you for teaching me so detailly. I have read the code style, and change the code as you suggest. I did small change to pass compile. Then I check the compile binay working well as we expected. 
By the way, I keep checkDependencies and hasCircularDependency because I dont know why they are remained. Maybe it need a further check?

ShubusCorporation (approved):

ShubusCorporation:
I mentioned it just for information. There are still unclear issues with validation. But it's a reason for further investigations and, probably, new PR-s.  On my opinion, the goal of this PR is achieved.

ShubusCorporation:
@alexvins , have a look, please.

PR #691, opened 2021-04-25
Fix: ID-collisions on large mods collections loading

ShubusCorporation:
Fixed:  
- ID-collisions on large mods collections processing
- one annoying assertion while parsing mods is fixed
- some refactoring of old code for better experience

RCA:

Map index for sub-objects, which have fixed obj->id greater than some fixed bound is calculated per algorithm which is intended for objects without fixed id. But the further code supposes, that calculated sub-object map index is  obj->id  of corresponding sub-object.

### DIFF
--- a/lib/mapObjects/CObjectClassesHandler.cpp
+++ b/lib/mapObjects/CObjectClassesHandler.cpp
@@ -148,30 +148,43 @@ std::vector<JsonNode> CObjectClassesHandler::loadLegacyData(size_t dataSize)
 
 /// selects preferred ID (or subID) for new object
 template<typename Map>
-si32 selectNextID(const JsonNode & fixedID, const Map & map, si32 defaultID)
+si32 selectNextID(const JsonNode & fixedID, const Map & map, si32 fixedObjectsBound)
 {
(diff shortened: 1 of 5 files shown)


lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -148,30 +148,43 @@ std::vector<JsonNode> CObjectClassesHandler::loadLegacyData(size_t dataSize)
...
 		return;
 	}
+	const auto convertedId = VLC->modh->normalizeIdentifier(entry.meta, "core", identifier);
+	const auto & entryIndex = entry["index"];
+	bool useSelectNextID = !isSubobject || entryIndex.isNull();
 
-	std::string convertedId = VLC->modh->normalizeIdentifier(entry.meta, "core", identifier);
-
-	si32 id = selectNextID(entry["index"], obj->subObjects, 1000);
+	if(useSelectNextID && isSubobject)
```
ShubusCorporation, line 181: It's theoretical case, according to the current code behavior and debug logs, sub-objects have ID => this branch is not executed

alexvins (approved):

PR #692, opened 2021-04-28
Fix: VCAI should not attempt to move Spell Book

ShubusCorporation:
Minor fix:
vcai_art_err


nullkiller (approved):
I expect it might also have problems trying to equip angelic alliance while another hero has some armor. So the right hand slot of the target hero is empty but the chest slot is not.

nullkiller:
But looks like it will not try to move the angelic alliance if any required slot is busy and no error will appear. So never mind.

alexvins (approved):

PR #693, opened 2021-05-08
client/lobby/SelectionTab.cpp: initialize generalSortingBy before use

trofi:
Noticed use of uninitialized value when built vcmi with -fsanitize=undefined:

  Established connection with
    VCMI 0.99 b310f2e61ece1ab550614cf1b99b04830820700c (server).
    UUID: bab9a90d-7416-4566-8817-e367ffcac7c1
  ../vcmi-9999/client/lobby/SelectionTab.cpp:138:16:
    runtime error: load of value 32717, which is not a valid value for type 'ESortBy'
  /usr/lib/gcc/x86_64-pc-linux-gnu/12.0.0/include/g++-v12/tuple:190:4:
    runtime error: load of value 32717, which is not a valid value for type 'ESortBy'
  /usr/lib/gcc/x86_64-pc-linux-gnu/12.0.0/include/g++-v12/tuple:190:4:
    runtime error: load of value 32717, which is not a valid value for type 'ESortBy'

Here (before the change) `SelectionTab()` used `generalSortingBy` before
first use:

```
    SelectionTab::SelectionTab(ESelectionScreen Type) {
        ...
        if(tabType != ESelectionScreen::campaignList)
        {
            ...
                ESortBy criteria = (ESortBy)i;
                if(criteria == _name)
                    criteria = generalSortingBy;

                buttonsSortBy.push_back(... std::bind(&SelectionTab::sortBy, this, criteria)));
            ...
        }
        ...
        switch(tabType)
        {
        case ESelectionScreen::newGame:
                generalSortingBy = ESortBy::_name;
        ....
```

The change moves `generalSortingBy` initialization before first use.

Signed-off-by: Sergei Trofimovich <slyfox@gentoo.org>

### DIFF
--- a/client/lobby/SelectionTab.cpp
+++ b/client/lobby/SelectionTab.cpp
@@ -111,10 +111,30 @@ bool mapSorter::operator()(const std::shared_ptr<CMapInfo> aaa, const std::share
 	}
 }
 
+// pick sorting order based on selection
+static ESortBy getSortBySelectionScreen(ESelectionScreen Type)
+{
+	switch(Type)
+	{
+	case ESelectionScreen::newGame:
+		return ESortBy::_name;
+	case ESelectionScreen::loadGame:
+	case ESelectionScreen::saveGame:
+		return ESortBy::_fileName;
+	case ESelectionScreen::campaignList:
+		return ESortBy::_name;
+	}
+	// Should not reach here. But let's not crash the game.
+	return ESortBy::_name;
+}
+
 SelectionTab::SelectionTab(ESelectionScreen Type)
 	: CIntObject(LCLICK | WHEEL | KEYBOARD | DOUBLECLICK), callOnSelect(nullptr), tabType(Type), selectionPos(0), sortModeAscending(true)
 {
 	OBJ_CONSTRUCTION;
+
+	generalSortingBy = getSortBySelectionScreen(tabType);
+
 	if(tabType != ESelectionScreen::campaignList)
 	{
 		sortingBy = _format;
@@ -146,20 +166,16 @@ SelectionTab::SelectionTab(ESelectionScreen Type)
 	switch(tabType)
 	{
 	case ESelectionScreen::newGame:
-		generalSortingBy = ESortBy::_name;
 		tabTitle = CGI->generaltexth->arraytxt[229];
 		break;
 	case ESelectionScreen::loadGame:
-		generalSortingBy = ESortBy::_fileName;
 		tabTitle = CGI->generaltexth->arraytxt[230];
 		break;
 	case ESelectionScreen::saveGame:
 		positionsToShow = 16;
-		generalSortingBy = ESortBy::_fileName;
 		tabTitle = CGI->generaltexth->arraytxt[231];
 		break;
 	case ESelectionScreen::campaignList:
-		generalSortingBy = ESortBy::_name;
 		tabTitle = CGI->generaltexth->allTexts[726];
 		type |= REDRAW_PARENT; // we use parent background so we need to make sure it's will be redrawn too
 		pos.w = parent->pos.w;


alexvins (approved):

client/lobby/SelectionTab.cpp
```diff
@@ -115,6 +115,27 @@ SelectionTab::SelectionTab(ESelectionScreen Type)
...
+	{
+	case ESelectionScreen::newGame:
+		generalSortingBy = ESortBy::_name;
+		break;
+	case ESelectionScreen::loadGame:
+		generalSortingBy = ESortBy::_fileName;
+		break;
+	case ESelectionScreen::saveGame:
+		generalSortingBy = ESortBy::_fileName;
+		break;
```
nullkiller, line 130: These two cases do the same thing. We can combine them. Same for other two cases. Also might be worth to extract this thing to a method. The constructor itself is already quite big.

trofi: Sounds good. Moved out to a separate helper.

client/lobby/SelectionTab.cpp
```diff
@@ -111,10 +111,32 @@ bool mapSorter::operator()(const std::shared_ptr<CMapInfo> aaa, const std::share
...
+// pick sorting order based on selection
+static ESortBy SortBySelectionScreen(ESelectionScreen Type)
+{
+	// initialize 'generalSortingBy' early
+	switch(Type)
+	{
+	case ESelectionScreen::newGame:
+		return ESortBy::_name;
+	case ESelectionScreen::loadGame:
+	case ESelectionScreen::saveGame:
```
dydzio0614, line 123: consider using FALLTHROUGH macro (defined in Global.h) - the purpose of fallthrough is explained here: https://en.cppreference.com/w/cpp/language/attributes/fallthrough

client/lobby/SelectionTab.cpp
```diff
@@ -111,10 +111,32 @@ bool mapSorter::operator()(const std::shared_ptr<CMapInfo> aaa, const std::share
...
+		return ESortBy::_fileName;
+	case ESelectionScreen::campaignList:
+		return ESortBy::_name;
+	}
+	// should not reach here
+	assert(0);
```
nullkiller, line 129: I'd rather return _name in all the rest cases and remove assert at all. It is definitely not the situation where you want game to crash.

nullkiller (changes requested):
Please rebase on top of develop so that latest CI builds are passed and I will merge it.

client/lobby/SelectionTab.cpp
nullkiller: It is obvious. Even C# allows such constructs.

nullkiller (approved):

client/lobby/SelectionTab.cpp
```diff
@@ -111,10 +111,32 @@ bool mapSorter::operator()(const std::shared_ptr<CMapInfo> aaa, const std::share
 	}
 }
 
+// pick sorting order based on selection
+static ESortBy SortBySelectionScreen(ESelectionScreen Type)
```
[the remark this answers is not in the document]
trofi: `static void foo(){}; static int bar=1;` should be equivalent to unnamed namespace `namespace { void foo(){}; int bar=1; }`, but is bound to a single declaration.

The main property `static` symbol is module locality. That usually means:
- for developers: compiler will warn when symbol becomes unused due to other code changes
- for compilers: better optimizations are unlocked due to no requirement of keeping symbol externally visible or ABI-compliant (more of dead code elimination, more constant propagation, more local inlining opportunities).
- for linkers: shorter external symbol namespace speeds linking up.

Maybe `vcmi` uses something else to achieve the equivalent?

client/lobby/SelectionTab.cpp
```diff
@@ -111,10 +111,32 @@ bool mapSorter::operator()(const std::shared_ptr<CMapInfo> aaa, const std::share
...
+{
+	// initialize 'generalSortingBy' early
```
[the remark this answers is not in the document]
trofi: Removed.

client/lobby/SelectionTab.cpp
trofi: Sounds good. Changed to return `_name`.

client/lobby/SelectionTab.cpp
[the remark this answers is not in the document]
nullkiller: No, I am not aware of anything else. Just asking because I am a bit dumb at C++. Thank you for clarifications.

nullkiller:
MXE build crashes - it is a known issue, fixed in my nullkiller-ai branch.

client/lobby/SelectionTab.cpp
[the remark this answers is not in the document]
dydzio0614: https://www.geeksforgeeks.org/what-are-static-functions-in-c/ i guess this applies to C++ as well

PR #694, opened 2021-05-27
Fix: nextPrimarySkill crash in case when probability values are incorrect

ShubusCorporation:
Client crashes when server calculates next incremented primary skill as 'Experience' . 
RCA: some mods contain wrong probability data to calculate next incremented primary skill.

### DIFF
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -1213,7 +1213,8 @@ PrimarySkill::PrimarySkill CGHeroInstance::nextPrimarySkill(CRandomGenerator & r
 {
 	assert(gainsLevel());
 	int randomValue = rand.nextInt(99), pom = 0, primarySkill = 0;
-	const auto & skillChances = (level > 9) ? type->heroClass->primarySkillLowLevel : type->heroClass->primarySkillHighLevel;
+	const auto isLowLevelHero = level < GameConstants::HERO_HIGH_LEVEL;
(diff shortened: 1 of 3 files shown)


lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1220,7 +1220,12 @@ PrimarySkill::PrimarySkill CGHeroInstance::nextPrimarySkill(CRandomGenerator & r
 			break;
 		}
 	}
-
+	if(primarySkill >= GameConstants::PRIMARY_SKILLS)
+	{
```
nullkiller, line 1224: I would suggest to either retry the logic above or normalize probabilities right after they are loaded from json

nullkiller: Or automatically select the skill with highest probability

nullkiller: Also we can sum up all probabilities and use this value instead of 99

ShubusCorporation: I've added some error messages to the mod loading routine.

In general case,, it may be zero or negative value. Thus, I don't want to think instead of mods authors at this point and invent some kind of crutches. My goal here is to warn mod developer and provide appropriate game experience for player.

nullkiller: Agreed.

nullkiller (approved):

ShubusCorporation:
@alexvins , take a look, pls.

PR #695, opened 2021-05-30
Implement lua rng

nullkiller:
VR:T and VR:R

- [x] Need to seed VR:R somehow

### DIFF
--- a/scripting/lua/api/ServerCb.cpp
+++ b/scripting/lua/api/ServerCb.cpp
@@ -15,6 +15,8 @@
 
 #include "../LuaStack.h"
 #include "../../../lib/NetPacks.h"
+#include "../../../lib/CRandomGenerator.h"
+
 namespace scripting
 {
 namespace api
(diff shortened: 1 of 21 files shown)


scripts/lib/erm/VR.lua
```diff
@@ -74,5 +74,18 @@ function VR:M6()
...
 
+function VR:T(maxVal)
+	local rng = Rng.getDefault()
+	local r = rng:nextInt(maxVal)
+	
+	return self.v + r
+end
+
+function VR:R(maxVal)
+	local rng = Rng.getDefault()
```
nullkiller, line 85: We also need a way to get rng with some fixed initial seed or rng from gamestate for VR:R. @alexvins, any ideas?

alexvins: We should extend ServerCbProxy API with `ServerCallback::getRNG`.

nullkiller: Thank you. Will try to do it.

nullkiller: Fixed this.

scripting/lua/api/ServerCb.cpp
```diff
@@ -43,9 +45,27 @@ const std::vector<ServerCbProxy::CustomRegType> ServerCbProxy::REGISTER_CUSTOM =
...
+int ServerCbProxy::getRNG(lua_State * L)
+{
+	LuaStack S(L);
+	ServerCallback * obj = nullptr;
+
+	if(!S.tryGet(1, obj))
+		return S.retVoid();
+
+	S.clear();
+	S.push(dynamic_cast<CRandomGenerator *>(obj->getRNG()));
```
alexvins, line 65: it should be no cast here and API for RNG should use vstd::RNG. Also you should use `RNGMock` in tests

nullkiller: I can try to move nextInt and nextDouble from CRandomGenerator to vstd::RNG and use vstd::RNG* in API.

alexvins: Please do not move, just declare as pure virtual

nullkiller: Fixed this. Was not able to set mocked RNG for VR:T.

PR #700, opened 2021-07-25
client/lobby/CBonusSelection.cpp: fix difficulty overflow in bonus UI

trofi:
Before the change campaign bonus selection screen had inconsistent
overflow behaviour for difficulty selection:

```
    0 1 2 3 4 | available buttons: '-' to decrease
      ^       |                    '+' to increase
```

Before the change:

1. If we click '+' 5 times we will end up on difficulty=4 (ok, saturated).
2. If we click '-' 5 times we will end up on difficulty=1 (unexpected, wrap around).

After the change:

1. If we click '+' 5 times we will end up on difficulty=4 (saturated).
2. If we click '-' 5 times we will end up on difficulty=0 (saturated).

The inconsistency happens because `difficulty` variable has `ui8` type
and server uses `difficulty = vstd::abetween(difficulty, 0, 4)` to
implement the saturation.

For large positive values saturation works as expected:
    vstd::abetween(difficulty=5, 0, 4) -> 4
For small values it does not:
    vstd::abetween(difficulty=-1, 0, 4) -> 4

The change makes client to avoid using negative values.

### DIFF
--- a/client/lobby/CBonusSelection.cpp
+++ b/client/lobby/CBonusSelection.cpp
@@ -460,7 +460,9 @@ void CBonusSelection::increaseDifficulty()
 
 void CBonusSelection::decreaseDifficulty()
 {
-	CSH->setDifficulty(CSH->si->difficulty - 1);
+	// avoid negative overflow (0 - 1 = 255)
+	if (CSH->si->difficulty > 0)
+		CSH->setDifficulty(CSH->si->difficulty - 1);
 }
 
 CBonusSelection::CRegion::CRegion(int id, bool accessible, bool selectable, const SCampPositions & campDsc)


client/lobby/CBonusSelection.cpp
```diff
@@ -455,12 +455,22 @@ void CBonusSelection::restartMap()
...
 }
 
 void CBonusSelection::decreaseDifficulty()
 {
-	CSH->setDifficulty(CSH->si->difficulty - 1);
+	// Rely on client to saturate minimum difficuly level
+	// instead of wrap-around:
+	//     ui8 difficulty = 0; // easiest
+	//     max<ui8>(difficulty - 1, 0) -> 255 (overflow)
+	//     max<ui8>(255, hardest) -> hardest (wrap-around)
```
nullkiller, line 471: I personally think that one comment here is enough. Something like
// avoid negative overflow (0 - 1 = 255)
right before if. I personally believe that if code needs comments - something bad with overall design. So I would change si->dificulty to int. But you solution is easier and less impactful. So I suggest to simplify the comment. It will be enough for code reader to understand why if is needed

trofi: Sounds good. Left only `// avoid negative overflow (0 - 1 = 255)` comment.

PR #704, opened 2021-08-02
Linux packaging: Add metainfo file

Alexander-Wilms:
This adds a [metainfo](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html) file, which is used in KDE Discover, Gnome Software, Flathub etc.

The [Open Age Rating Service](https://hughsie.github.io/oars/) can be used to generate the oars tags by answering a few questions, if you think they should be changed.


ArseniyShestakov:
I guess this one is good enough. Merging.

ArseniyShestakov:
And yeah thank you for contribution!

PR #705, opened 2021-08-14
Fix some warnings, Add headers to the solution, Fix MSVS build

ShubusCorporation:
- Fixed some minor warnings.
- There's an attempt to solve lua51.dll issue - it should be copied into build\{BUILD_OPTION} folder where BUILD_OPTION is Release, Debug or RelWithDebInfo. It's not working when  lua51.dll is in the 'bin' folder.

- Fixed some weird build issues. Without it I available to build solution only commenting some problem lines.
Following lines of code:

trigger %= cmdName >> -identifier >> -condition > qi::lit(";");
normalBodyOption = qi::char_("A-Z") > -(exactBodyOptionList);
instruction %= cmdName >> -identifier >> -condition >> body;
receiver %= cmdName >> -identifier >> -condition >> body;
postTrigger %= cmdName >> -identifier >> -condition > qi::lit(";");

cause following errors on build:

` ERMParser.cpp
\VCMI\vcpkg\installed\x64-windows\include\boost/spirit/home/qi/detail/attributes.hpp(87): error C2593: 'operator =' is ambiguous

\VCMI\vcpkg\installed\x64-windows\include\boost/optional/optional.hpp(994): 

'boost::optional<ERM::Tidentifier> & boost::optional<ERM::Tidentifier>::operator =(boost::none_t) noexcept'
'boost::optional<ERM::Tidentifier> & boost::optional<ERM::Tidentifier>::operator =(boost::optional<ERM::Tidentifier> &&) noexcept'
'boost::optional<ERM::Tidentifier> & boost::optional<ERM::Tidentifier>::operator =(const boost::optional<ERM::Tidentifier> &)'
'boost::optional<ERM::Tidentifier> & boost::optional<ERM::Tidentifier>::operator =<std::vector<ERM::TIexp,std::allocator<_Ty>>>(T_ &&)'
`
The only idea I've found to fix them it's get rid of some optionals (So, I suggest to make identifier, body and optionList non optional).

### DIFF
--- a/scripting/erm/ERMInterpreter.cpp
+++ b/scripting/erm/ERMInterpreter.cpp
@@ -1282,7 +1282,7 @@ namespace ERMConverter
 
 			ERM::Tidentifier tid = trig.identifier.get();
 
-			if(tid.size() == 0)
+			if(tid.empty())
 				throw EInterpreterError("Function must have identifier");
 
 			Variable v = boost::apply_visitor(LVL1IexpToVar(), tid[0]);
@@ -1432,7 +1432,7 @@ bool ERMInterpreter::isATrigger( const ERM::TLine & line )
 	case 0: //v-exp
 		{
 			TVExp vexp = boost::get<TVExp>(line);
-			if(vexp.children.size() == 0)
+			if(vexp.children.empty())
 				return false;
 
 			switch (getExpType(vexp.children[0]))
@@ -1614,7 +1614,7 @@ namespace VERMInterpreter
 			if(asSymbol)
 			{
 				children.resize(children.size()+1);
-				for(int i=children.size()-1; i >0; i--)
+				for(auto i=children.size()-1; i >0; i--)
 				{
 					children[i] = children[i-1];
 				}
@@ -1718,7 +1718,7 @@ namespace VERMInterpreter
 	}
 	VOption OptionConverterVisitor::operator()( ERM::TSymbol const& cmd ) const
 	{
-		if(cmd.symModifier.size() == 0)
+		if(cmd.symModifier.empty())
 			return VSymbol(cmd.sym);
 		else
 			return VNode(cmd);
--- a/scripting/erm/ERMParser.cpp
+++ b/scripting/erm/ERMParser.cpp
@@ -187,7 +187,8 @@ std::vector<LineInfo> ERMParser::parseFile(CERMPreprocessor & preproc)
 	}
 	catch (ParseErrorException & e)
 	{
-		logGlobal->error("Stopped parsing file.");
+		logGlobal->error("ERM Parser Error. File: '%s' Line: %d Exception: '%s'"
+			, preproc.getCurFileName(), preproc.getCurLineNo(), e.what());
 		throw;
 	}
 	return ret;
--- a/scripting/erm/ERMParser.h
+++ b/scripting/erm/ERMParser.h
@@ -37,6 +37,11 @@ public:
 	{
 		return lineNo;
 	}
+
+	const std::string& getCurFileName() const
+	{
+		return fname;
+	}
 };
 
 //various classes that represent ERM/VERM AST
--- a/scripting/erm/CMakeLists.txt
+++ b/scripting/erm/CMakeLists.txt
@@ -7,7 +7,14 @@ set(lib_SRCS
 		ERMScriptModule.cpp
 )
 
-add_library(vcmiERM SHARED ${lib_SRCS})
+set(lib_HDRS
+		StdInc.h
+		ERMParser.h
+		ERMInterpreter.h
+		ERMScriptModule.h
+)
+
+add_library(vcmiERM SHARED ${lib_SRCS} ${lib_HDRS})
 target_link_libraries(vcmiERM ${Boost_LIBRARIES} vcmi)
 
 vcmi_set_output_dir(vcmiERM "scripting")
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -40,9 +40,11 @@ set(test_SRCS
 		erm/ERM_MF.cpp
 		erm/ERM_TM_T.cpp
 		erm/ERM_VR.cpp
+		erm/ERM_UN.cpp
 		erm/ERMPersistenceTest.cpp
 		erm/ExamplesTest.cpp
 		erm/interpretter/ERM_VR.cpp
+		erm/interpretter/ERM_UN.cpp
 		erm/interpretter/ErmRunner.cpp
 
 		events/ApplyDamageTest.cpp
--- a/test/erm/interpretter/ERM_UN.cpp
+++ b/test/erm/interpretter/ERM_UN.cpp
@@ -1,0 +1,28 @@
+/*
+ * ERM_UN.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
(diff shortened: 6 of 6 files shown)


scripting/erm/ERMParser.h
```diff
@@ -251,7 +251,7 @@ namespace ERM
 	//script line
 	typedef boost::variant<TVExp, TERMline> TLine;
 
-	template <typename T> class ERM_grammar;
+	template <typename T> struct ERM_grammar;
```
ShubusCorporation, line 254: It causes warning: ERM_grammar is class here and struct in some other place

scripting/lua/CMakeLists.txt
```diff
@@ -52,6 +52,6 @@ install(TARGETS vcmiLua DESTINATION ${SCRIPTING_LIB_DIR})
 #manually copy lua dll from vcpkg folder to build directory on windows since vcpkg deps copy feature has flaws, using hardcoded paths based on vcmi windows deps package 1.1 from github
 if(MSVC)
 	if(EXISTS ${LUA_INCLUDE_DIR}/../../bin/lua51.dll)
-		file(COPY ${LUA_INCLUDE_DIR}/../../bin/lua51.dll DESTINATION ${CMAKE_BINARY_DIR}/bin)
+		file(COPY ${LUA_INCLUDE_DIR}/../../bin/lua51.dll DESTINATION ${CMAKE_BINARY_DIR}/bin/${CMAKE_INSTALL_CONFIG_NAME})
```
ShubusCorporation, line 55: I guess CMAKE_INSTALL_CONFIG_NAME is Release, Debug  or RelWithDebInfo
But it does not works no because LUA_INCLUDE_DIR == ''LUA_INCLUDE_DIR:PATH=D:/Sources/VCMI/vcpkg/installed/x64-windows/include/luajit' and it's not present

scripting/lua/api/ServerCb.cpp
```diff
@@ -26,22 +26,22 @@ const std::vector<ServerCbProxy::CustomRegType> ServerCbProxy::REGISTER_CUSTOM =
 {
 	{
 		"addToBattleLog",
-		&ServerCbProxy::apply<BattleLogMessage>,
+		(lua_CFunction)(&ServerCbProxy::apply<BattleLogMessage>),
```
ShubusCorporation, line 29: Otherwise, it causes strange errors: "Can not cast overloaded_function to lua_CFunction"

scripting/erm/ERMParser.cpp
```diff
@@ -187,7 +187,7 @@ std::vector<LineInfo> ERMParser::parseFile(CERMPreprocessor & preproc)
 	}
 	catch (ParseErrorException & e)
 	{
-		logGlobal->error("Stopped parsing file.");
```
ShubusCorporation, line 190: It causes  warning because 'e' is not used

nullkiller: logGlobal->error("Stopped parsing file. %s", e.what()); Also we can log file name if available.

scripting/erm/ERMParser.h
```diff
@@ -197,17 +197,17 @@ namespace ERM
...
 	};
 
 	struct Treceiver
 	{
 		TCmdName name;
-		boost::optional<Tidentifier> identifier;
+		Tidentifier identifier;
 		boost::optional<Tcondition> condition;
-		boost::optional<Tbody> body;
+		Tbody body;
```
nullkiller, line 210: Will it be able to parse !!IP:R; or some other very short receivers?

scripting/erm/ERMInterpreter.cpp
```diff
@@ -1276,10 +1276,10 @@ namespace ERMConverter
 
 			out << "ERM:addTrigger({" << std::endl;
 
-			if(!trig.identifier.is_initialized())
+			if(!trig.identifier.size())
```
nullkiller, line 1279: Does it have trig.identifier.empty()? Usually collections expose such method and it is a bit faster.

scripting/erm/ERMInterpreter.cpp
```diff
@@ -837,10 +837,10 @@ namespace ERMConverter
 				break;
 			case 'S': //setting variable
 				{
-					if(!trig.params.is_initialized() || trig.params.get().size() != 1)
+					if(!trig.params.size() || trig.params.size() != 1)
```
nullkiller, line 840: not needed anymore. != 1 is enough

nullkiller: And in other places where we check size for exact value

scripting/erm/ERMParser.cpp
ShubusCorporation: Ok

scripting/erm/ERMInterpreter.cpp
ShubusCorporation: Ok, using !empty() instead of size()

scripting/erm/ERMInterpreter.cpp
ShubusCorporation: Ok

scripting/lua/CMakeLists.txt
nullkiller: Could you check if CMAKE_BUILD_TYPE variable does the same? Might be a bit more logical because it is build phase not the installation one

ArseniyShestakov: BTW this really better to be solved properly without sticking to specific DLL names.

Is there any other projects using lua? What solutions are they using?

nullkiller: @ArseniyShestakov, the goal of this PR is to fix build rather than improve it. There might be a better way but it is already merged and as for me it is too late to demand improvements. Thus making the improvement is a question of a separate PR.

ShubusCorporation: @nullkiller , @ArseniyShestakov

But I'm confused by the following code here:

if(NOT DEFINED CMAKE_INSTALL_CONFIG_NAME)
if(BUILD_TYPE)
string(REGEX REPLACE "^[^A-Za-z0-9_]+" ""
CMAKE_INSTALL_CONFIG_NAME "${BUILD_TYPE}")
else()
set(CMAKE_INSTALL_CONFIG_NAME "Release")
endif()
message(STATUS "Install configuration: "${CMAKE_INSTALL_CONFIG_NAME}"")
endif()

Why a special case for 'Release' is here? May BUILD_TYPE contain 'Release' value?

nullkiller: I think it is some kind of hack to ensure that it is not empty. For instance cpack without parameters supposed to build Release but in NSIS logs I see bin//some.dll which should be bin/release/some.dll. Adding --config Release or whatever cpack wants fixes this issue

ShubusCorporation: So, in context of your original question, if BUILD_TYPE may be empty, may be it's more secure to use `${ CMAKE_INSTALL_CONFIG_NAME}` or have I missed something?

nullkiller: Need to read about it a bit.

nullkiller: If it is some standard thing or not.

scripting/erm/ERMParser.h
```diff
@@ -37,6 +37,11 @@ class CERMPreprocessor
 	{
 		return lineNo;
 	}
+
+	const std::string& getCurFileName() const
```
nullkiller, line 41: spaces around &

PR #707, opened 2021-08-21
fix negative tile cost from pathfinding skill

nullkiller:
In some mods pathfinding can be boosted with artifacts to negative tile cost. unsigned overflow causes it to bypass some checks and later it is casted back to signed and become negative.


lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -105,7 +105,7 @@ ui32 CGHeroInstance::getTileCost(const TerrainTile & dest, const TerrainTile & f
 		if(ret < GameConstants::BASE_MOVEMENT_COST)
```
nullkiller, line 105: This line is bypassed if ret is unsigned because of overflow

PR #711, opened 2021-09-23
Load game data from GOG

cxong:


lib/CMakeLists.txt
```diff
@@ -254,6 +254,7 @@ set(lib_HEADERS
 		filesystem/ISimpleResourceLoader.h
 		filesystem/MinizipExtensions.h
 		filesystem/ResourceID.h
+		filesystem/find_steam_game.h
```
cxong, line 257: I included the whole header for convenience; open to changing how this file is included
For example in devilutionx they had a 3rd party CMakeLists.txt which downloaded the header from github https://github.com/diasurgical/devilutionX/commit/6091c3f3c2fbb68df4c463f5b92e57ce0850e62b

lib/VCMIDirs.cpp
```diff
@@ -312,7 +313,19 @@ bfs::path VCMIDirsWIN32::userConfigPath() const { return userDataPath() / "confi
 
 std::vector<bfs::path> VCMIDirsWIN32::dataPaths() const
 {
-	return std::vector<bfs::path>(1, bfs::path("."));
+	auto paths = std::vector<bfs::path>(1, bfs::path("."));
```
cxong, line 316: I think having the `.` path first means it will look here first

lib/VCMIDirs.cpp
```diff
@@ -312,7 +313,19 @@ bfs::path VCMIDirsWIN32::userConfigPath() const { return userDataPath() / "confi
...
+	// Add game store paths
+#ifdef VCMI_WINDOWS
+	char gogPath[_FSG_PATH_MAX];
+	fsg_get_gog_game_path(gogPath, "1207658787");
+	if (strlen(gogPath) > 0)
+	{
+		paths.push_back(bfs::path(gogPath));
+	}
+	// TODO: steam path for Heroes 3 HD Edition
+	// TODO: Ubisoft Connect
```
cxong, line 326: I can look into supporting steam / ubisoft in future PRs if we want to
I don't know if VCMI supports HD edition though

nullkiller: I think we do not support HD edition

lib/VCMIDirs.cpp
```diff
@@ -312,7 +313,19 @@ bfs::path VCMIDirsWIN32::userConfigPath() const { return userDataPath() / "confi
 
 std::vector<bfs::path> VCMIDirsWIN32::dataPaths() const
 {
-	return std::vector<bfs::path>(1, bfs::path("."));
+	auto paths = std::vector<bfs::path>(1, bfs::path("."));
+	// Add game store paths
+#ifdef VCMI_WINDOWS
+	char gogPath[_FSG_PATH_MAX];
+	fsg_get_gog_game_path(gogPath, "1207658787");
```
nullkiller, line 320: Lets avoid magic numbers. Please introduce a constant

lib/VCMIDirs.cpp
nullkiller: Please remove unused commands.

nullkiller (approved):

lib/filesystem/find_steam_game.h
```diff
@@ -0,0 +1,262 @@
...
+
+	_FSG_FUNC
+	bool _fsg_dir_exists(const char *dir)
+	{
+		struct stat info;
+		return stat(dir, &info) == 0 && (info.st_mode & S_IFDIR);
+	}
+
+	_FSG_FUNC
+	void fsg_get_steam_game_path(char *out, const char *name)
```
nullkiller, line 112: If we do not use this method lets not include it. Another option is to use submodule

nullkiller: We actually use submodules for dependencies not available as packages

.gitmodules
```diff
@@ -5,3 +5,6 @@
 	path = AI/FuzzyLite
 	url = https://github.com/fuzzylite/fuzzylite.git
 	branch = release
+[submodule "3rdparty/find_steam_game"]
+	path = 3rdparty/find_steam_game
+	url = https://github.com/cxong/find_steam_game
```
ArseniyShestakov, line 10: So yeah I'm strongly against adding submodules unless it's very essential or large libraries.

ArseniyShestakov (changes requested):

nullkiller:
I thought about it. Both cases have their props and cons. Sorry for misleading you @cxong.

nullkiller:
I made up my mind what I want from this thing. I want it being optional. This can be some setting in json file for instance. Not sure if it can be done easily so because settings.json is not suitable for it. I have not made up my mind where to put such setting. @cxong are you still interested in completing this PR?

cxong:
I think clearer instructions are required, where in settings.json, what key it should use, are there any other changes needed or just add to settings.json? Closing for now

nullkiller:
> I think clearer instructions are required, where in settings.json, what key it should use, are there any other changes needed or just add to settings.json? Closing for now

My proposition:
- add a new settings file startup.json to the game root directory
- add a field there detectGogInstall=true
- before adding GOG path to filesystem check this field is not false. Only question left how to load it. I can take a look onto this as well. So it should be ok to use standard file functions as our filesystem is not initialized yet

PR #712, opened 2021-10-09
Make scripting dependencies required

kambala-decapitator:
If you build only client (or server) with `cmake --build . --target vcmiclient`, then you receive crash on launch:

> Cannot open dynamic library (./scripting/libvcmiERM.dylib). Throwing...

because scripting libs aren't built.

Although it’s natural that **lib** should depend on scripting libs since it loads them, it's not possible in the current state:

```
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  “vcmiERM” of type SHARED_LIBRARY
    depends on “vcmi” (weak)
    depends on “vcmiLua” (weak)
  “vcmiLua” of type SHARED_LIBRARY
    depends on “vcmi” (weak)
    depends on “vcmiERM” (weak)
  “vcmi” of type SHARED_LIBRARY
    depends on “vcmiERM” (weak)
    depends on “vcmiLua” (weak)
At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

As a workaround I added the scripting dependency to **server** instead.

ERM also depends on LUA now as advised by @nullkiller in Slack:

> ERM interpretter converts ERM to LUA and executes lua
> There might be no direct dependency so


nullkiller:
I will take a look onto it a bit later. Not sure what is the best solution.

PR #717, opened 2021-11-08
Fixed: Advanced options window issues

ShubusCorporation:
1) Only allowed factions should be available as town option
2) User should not be able to launch game with comps only

random_arrows

no_town_arrows

### DIFF
--- a/client/lobby/OptionsTab.cpp
+++ b/client/lobby/OptionsTab.cpp
@@ -33,7 +33,7 @@
 #include "../../lib/mapping/CMap.h"
 #include "../../lib/mapping/CMapInfo.h"
 
-OptionsTab::OptionsTab()
+OptionsTab::OptionsTab() : humanPlayers(0)
 {
 	recActions = 0;
 	OBJ_CONSTRUCTION;
@@ -57,11 +57,15 @@ OptionsTab::OptionsTab()
 void OptionsTab::recreate()
 {
 	entries.clear();
+	humanPlayers = 0;
 
 	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
 	for(auto & pInfo : SEL->getStartInfo()->playerInfos)
 	{
-		entries.insert(std::make_pair(pInfo.first, std::make_shared<PlayerOptionsEntry>(pInfo.second)));
+		if(pInfo.second.isControlledByHuman())
+			humanPlayers++;
+
+		entries.insert(std::make_pair(pInfo.first, std::make_shared<PlayerOptionsEntry>(pInfo.second, * this)));
 	}
 
 	if(sliderTurnDuration)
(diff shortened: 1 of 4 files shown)


client/lobby/OptionsTab.cpp
```diff
@@ -493,7 +497,9 @@ OptionsTab::PlayerOptionsEntry::PlayerOptionsEntry(const PlayerSettings & S)
 
 	if(SEL->screenType != ESelectionScreen::scenarioInfo && SEL->getPlayerInfo(s.color.getNum()).canHumanPlay)
 	{
-		flag = std::make_shared<CButton>(Point(-43, 2), flags[s.color.getNum()], CGI->generaltexth->zelp[180], std::bind(&IServerAPI::setPlayer, CSH, s.color));
+		flag = std::make_shared<CButton>(Point(-43, 2), flags[s.color.getNum()], CGI->generaltexth->zelp[180],
```
nullkiller, line 500: Lets have all args each on a new line

client/lobby/OptionsTab.cpp
```diff
@@ -493,7 +497,9 @@ OptionsTab::PlayerOptionsEntry::PlayerOptionsEntry(const PlayerSettings & S)
...
+			std::bind(& OptionsTab::onSetPlayerClicked, & parentTab, s));
```
nullkiller, line 501: As far as I know we do not need space after & when it is used to take address (unary operation, but we need spaces in declarations)

lib/mapping/MapFormatH3M.cpp
```diff
@@ -220,15 +220,24 @@ void CMapLoaderH3M::readPlayerInfo()
...
 
-		for(int fact = 0; fact < totalFactions; ++fact)
+		const bool isFactionRandom = mapHeader->players[i].isFactionRandom = reader.readBool();
+		const ui16 allFactionsMask = (mapHeader->version == EMapFormat::ROE)
+			? 0b1111111
+			: 0b11111111;
+		const bool allFactionsAllowed = mapHeader->version == EMapFormat::VCMI
+			|| (isFactionRandom && ((allowedFactions & allFactionsMask) == allFactionsMask));
+
+		if(!allFactionsAllowed)
```
nullkiller, line 230: Interesting why the previous code did not work. Logically I see two cases
- all factions allowed
- some factions allowed
For each case we can have isFactionRandom set or unset. Are you sure we cover all these? Four cases in total.
For instance randomFaction + not all factions allowed?

client/lobby/OptionsTab.cpp
```diff
@@ -505,6 +511,12 @@ OptionsTab::PlayerOptionsEntry::PlayerOptionsEntry(const PlayerSettings & S)
 	bonus = std::make_shared<SelectedBox>(Point(271, 2), s, BONUS);
 }
 
+void OptionsTab::onSetPlayerClicked(const PlayerSettings & ps) const
+{
+	if(ps.isControlledByAI() || humanPlayers > 1)
```
nullkiller, line 516: Does it mean multiplayer mode?

lib/mapping/MapFormatH3M.cpp
nullkiller: Are you sure we will not get conflux town for ROE when all factions allowed?

client/lobby/OptionsTab.cpp
ShubusCorporation: Yes, it does.

lib/mapping/MapFormatH3M.cpp
ShubusCorporation: 1) Because `mapHeader->players[i].allowedFactions` by default contains all factions include mods. Then they tried to remove only vanila factions (which are not allowed). Thus, custom factions remained here in any way.

2) Then we should determine if all factions are allowed or not. 

For Non-VCMI maps:
- If a faction is not random let's suppose mod factions are not allowed even bit mask is full.
- If a faction is random all factions (include mods) are allowed when we have full bit mask

3) HoTa allows `Conflux` and `Cove` for RoE maps when a faction is random and all RoE factions are allowed
roe_random
`All For One` - one of RoE maps. So, let's use HoTa implementation as an example.

PR #718, opened 2021-11-09
Fixed: Animation errors should not lead to a hanging

ShubusCorporation:
Invalid defs which don't contain some frames and produce `CAnimation` errors may cause a hanging during a battle.

### DIFF
--- a/client/battle/CCreatureAnimation.cpp
+++ b/client/battle/CCreatureAnimation.cpp
@@ -210,14 +210,19 @@ bool CCreatureAnimation::incrementFrame(float timePassed)
 {
 	elapsedTime += timePassed;
 	currentFrame += timePassed * speed;
+	const auto framesNumber = framesInGroup(type);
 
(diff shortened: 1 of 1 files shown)


client/battle/CCreatureAnimation.cpp
```diff
@@ -210,20 +210,26 @@ bool CCreatureAnimation::incrementFrame(float timePassed)
...
-		if (once)
+		if(once)
 			setType(CCreatureAnim::HOLDING);
 
 		endAnimation();
-		return true;
+		ret = true;
 	}
-	return false;
+	return ret;
```
nullkiller, line 232: Do you prefer a single return in a method? I would leave it as is.

ShubusCorporation: Ok

ShubusCorporation: Ok

PR #720, opened 2021-11-26
Feature: Army management key shortcuts should work as in HD+ Mod

ShubusCorporation:
New features are implemented:

[SHIFT] + Click - “smart” stepwise proportional division. Ignores units with a single creature,  if necessary, evens out the number and divides the stacks proportionally. With this operation, you can quickly align by the number of stacks of the same type, divide the stack in half in 1 click, divide by 3 in 2 clicks. 

[ALT] + Click - collects troops of the same type from hero / garrison slots into a slot

[ALT] + [SHIFT] + Click - disbands the squad (the last squad of the hero cannot be fired).

[CTRL] + [ALT] + Click - in the hero meeting window and in the city window, transfers the stack to another hero. You cannot leave a hero without an army, so at least 1 creature remains. This combination also works in the garrison.

[CTRL] + [ALT] + [SHIFT] + Click - transfers the entire army. If an army is transferred from the hero, then at least one creature remains from the clicked squad.

Reworked:

[CTRL] + [SHIFT] + Click - separates one creature from the squad to all empty hero / garrison slots.
(Now works as a bulk operation)

Minor refactoring is implemented for the following feature:

[CTRL] + Click - separates one creature from the squad into a free slot.

___________________________________________________________________________________________________________________________

I've taken measures to minimize the network packets quantity. The set of bulk operations is implemented due to this purpose.  In addition, the key shortcuts are more responsive now. C++, Algebra and a little inspiration came in handy for deriving a smart balancing formula, implementing armies merging algorithm and other things..

Minor status bar text issue has been fixed for heroes exchange window:
garrison

The copyright has been updated to 2022.

### DIFF
--- a/CCallback.h
+++ b/CCallback.h
@@ -78,6 +78,12 @@ public:
 	virtual void save(const std::string &fname) = 0;
 	virtual void sendMessage(const std::string &mess, const CGObjectInstance * currentObject = nullptr) = 0;
 	virtual void buildBoat(const IShipyard *obj) = 0;
+
+	// To implement high-level army management bulk actions
+	virtual int bulkMoveArmy(ObjectInstanceID srcArmy, ObjectInstanceID destArmy, SlotID srcSlot) = 0;
+	virtual int bulkSplitStack(ObjectInstanceID armyId, SlotID srcSlot, int howMany = 1) = 0;
+	virtual int bulkSmartSplitStack(ObjectInstanceID armyId, SlotID srcSlot) = 0;
+	virtual int bulkMergeStacks(ObjectInstanceID armyId, SlotID srcSlot) = 0;
 };
 
 struct CPackForServer;
@@ -99,7 +105,9 @@ public:
 	friend class CClient;
 };
 
-class CCallback : public CPlayerSpecificInfoCallback, public IGameActionCallback, public CBattleCallback
+class CCallback : public CPlayerSpecificInfoCallback,
+	public IGameActionCallback,
+	public CBattleCallback
 {
 public:
 	CCallback(CGameState * GS, boost::optional<PlayerColor> Player, CClient *C);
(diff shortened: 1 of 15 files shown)


server/CGameHandler.cpp
```diff
@@ -2962,6 +2965,262 @@ void CGameHandler::load(const std::string & filename)
...
+
+	BulkSmartRebalanceStacks bulkSRS;
+
+	if(freeSlot != SlotID())
+	{
+		RebalanceStacks rs;
+		rs.srcArmy = rs.dstArmy = army->id;
+		rs.srcSlot = slotSrc;
+		rs.dstSlot = freeSlot;
+		rs.count = 1;
```
nullkiller, line 3194: Why 1? What does it mean?

server/CGameHandler.h
```diff
@@ -353,7 +357,9 @@ class CGameHandler : public IGameCallback, public CBattleInfoCallback, public En
 	void checkVictoryLossConditions(const std::set<PlayerColor> & playerColors);
 	void checkVictoryLossConditionsForAll();
 
-
+	const std::string complainNoCreatures;
+	const std::string complainNotEnoughCreatures;
+	const std::string complainInvalidSlot;
 };
```
nullkiller, line 364: Lets declare them in CPP file in global scope or do you plan to use them somewhere else outside CGameHandler?

server/CGameHandler.cpp
ShubusCorporation: Just fill free slot (if exists) with target creature, then adjust count.

server/CGameHandler.h
ShubusCorporation: These constants are private, so I'm not going to use them somewhere outside CGameHandler.
In other hand, there are some other classes in the cpp file, so...

CCallback.h
```diff
@@ -80,6 +80,15 @@ class IGameActionCallback
 	virtual void buildBoat(const IShipyard *obj) = 0;
 };
 
+class IGameBulkActionCallback // To implement high-level army management actions
```
nullkiller, line 83: What is the reason having it as a separate interface? Do you plan to implement it alone? I feel like it will always be implemented together with IGameActionCallback.

ShubusCorporation: Ok

PR #721, opened 2021-12-17
fix crash due assertion failure

user2589:
In some cases (seems to happen more often with black dragons), vcmiclient crashes with:

```
/AI/BattleAI/AttackPossibility.cpp:121: static AttackPossibility AttackPossibility::evaluate(const BattleAttackInfo&, BattleHex, const HypotheticBattle*): Assertion `u->unitId() != attacker->unitId()' failed.
```

I think the cause is pretty obvious from the change

### DIFF
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -1504,7 +1504,7 @@ std::vector<const battle::Unit*> CBattleInfoCallback::getAttackedBattleUnits(con
 			if (vstd::contains(at.hostileCreaturePositions, hex))
 				return true;
 			if (vstd::contains(at.friendlyCreaturePositions, hex))
-				return true;
+				return false;
 		}
 		return false;
 	});


lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1504,7 +1504,7 @@ std::vector<const battle::Unit*> CBattleInfoCallback::getAttackedBattleUnits(con
 			if (vstd::contains(at.hostileCreaturePositions, hex))
 				return true;
 			if (vstd::contains(at.friendlyCreaturePositions, hex))
-				return true;
+				return false;
```
nullkiller, line 1507: When getAttackedBattleUnits is called? And what it is expected to return? According to its name I would expect it to return creatures affected by attack. Thus black dragons can hit friendly creatures standing behind the attacked stack.

nullkiller: Btw the method below looks very similar to this one. It also includes friendly creatures

user2589: > When is getAttackedBattleUnits called? And what is it expected to return? 

I only have indirect evidence (below), as there is no design doc or even comments. Having those would dramatically reduce the threshold of entry for new contributors.

> According to its name I would expect it to return creatures affected by attack.

I doubt that. Promised indirect evidence: at [AI/BattleAI/AttackPossibility.cpp#L97](https://github.com/vcmi/vcmi/blob/8cae3398ba93d8c7de0879710c529f276d9148b9/AI/BattleAI/AttackPossibility.cpp#L97), it is used to get attackable units. 

Looking at the code of this method, it gets attackable hexes:

```
	if (rangedAttack)
		at = getPotentiallyShootableHexes(attacker, destinationTile, attackerPos);
	else
		at = getPotentiallyAttackableHexes(attacker, destinationTile, attackerPos);
``` 

The lambda in question is then used to filter units whose hexes are attackable - hence, `false` for friendly units makes sense.

Perhaps, the method should be renamed to `getAttackableBattleUnits`

nullkiller: I have an idea. The assertion might trigger if black dragon attacks enemy in a way that the dragon breath targets hex where the dragon stood before. The dragon is at the left, enemy 1 hex. Dragon moves to the right and attacks so the breath hits the tile where dragon's back was initially. And assertion triggers assert(u->unitId() != attacker->unitId());

ShubusCorporation: I was unable to reproduce: 
after2
before2
before_attack
after

user2589: > The assertion might trigger if black dragon attacks enemy in a way that the dragon breath targets hex where the dragon stood before.

I doubt it was the case. I've seen this issue twice, first time attacking a pack of black dragons, the game crashed on the first turn. It did not crash all the time, but the more spread the dragons were (because the hero had less units), the more often it crashed. Another time it was on the AI's turn, and it crashed 100% of the time - this is when I went out to fix it. So far, it works perfectly fine and I did not see any side effects.

> When is getAttackedBattleUnits called?

The only place this function is called is from the BattleAI function linked above. If there are no objections, I'm going to update the PR to rename the function to  `getAttackableBattleUnits`

ShubusCorporation: > false for friendly units makes sense.
This can break basic mechanics - some creatures may cause damage to allied creatures.

user2589: > This can break basic mechanics - some creatures may cause damage to allied creatures.

This function is only used to evaluate potential attacks by AI

ShubusCorporation: And what does it mean? Battle AI should not take into consideration that it can damage own units?? It is necessary to find out why it looks like the attacker  attacks himself.  "To fix" means find out a Root Cause and fix it, not just a symptom.

user2589: >  "To fix" means find out a Root Cause and fix it

feel free to reject the PR

>> This function is only used to evaluate potential attacks by AI

> Battle AI should not take into consideration that it can damage own units?

I am saying that this function is not used to consider friendly damage, just to enumerate possible attacks. You cannot directly attack a friendly unit, and collateral damage is taken care of in a separate place (AttackPossibility L157). This is THE fix.

nullkiller: The easiest way to fix assertion failure is to disable assertions during build. Probably build in Release or RelWithDebInfo. If you want to fix the issue you are expected to investigate and reproduce the root cause. Understand why the assertion was here, how exactly evaluation works and so on. Otherwise you may introduce more bugs then fix. Our daily builds are built with assertions disabled.

nullkiller: getAttackableBattleUnits feels like it enumerates all units on the battlefield which can be attacked but we have destionationTile so feels like we are evaluating a single attack and we want to know units affected by it so I better like the original name.

nullkiller: And no one from active maintainers know this area well, so do not expect much help here. Still I can not accept the PR without understanding it so either you have to investigate and be confident in what you are doing or I have to spend a lot of time to do this for you. Sorry, I hardly have time for this.

ShubusCorporation:
>> In some cases (seems to happen more often with black dragons)

By the way, could you provide corresponding saves?

lib/battle/CBattleInfoCallback.cpp
nullkiller: Ok, I reproduced it.
image

nullkiller: Enemy dragon under Battle AI control wants to fly across the pikeman and attack from opposite side. Thus it will hit the tile where it stands. This causes assertion to trigger because at the moment of attack the dragon will stay in a different place and no collateral damage will happen.

nullkiller: Magogs can damage themselves with their explosion like attack when attacking adjacent enemy under the bow of sharpshooter effect in the current version of vcmi. Not sure how it works in the original game. So the assertion is logically incorrect. But with dragons it should not count as collateral damage.

ShubusCorporation: Have a look at this:

code
json

nullkiller: I mean AI thinks that dragons can damage themselves and counts collateral damage. And here assertions hints correctly that something is wrong because dragons can not hit themselves.
But for magogs it is possible that a stack of magogs can damage itself attacking adjacent enemy under the effect of bow of sharpshooter (it allows shooting adjacent enemy and magogs hit all adjacent tiles with explosion like attack and they are adjacent so they really hit themselves, I checked it). So here assertion is wrong.

nullkiller: So the actual fix is to use real positions for all units except the attacker. For the attacker we need to use its position according to the destinationTile instead. If the stack's attack affects its own position according to its destination tile - count collateral damage. If the attack affects attacker's original position but not the destination one - do not count collateral damage because the attacker will not be there when the actual attack happens.

nullkiller: @user2589, hope you will finish this PR and all the investigations are not just wasted.

user2589:
> By the way, could you provide corresponding saves?

I didn't keep the saves

lib/battle/CBattleInfoCallback.cpp
user2589: @nullkiller I'm afraid I'm not qualified enough to make such a change, at least for now.

nullkiller: Yep. things sometimes are not that ease as they seams to be at the first glance. Especially when the code is old and full of hacks. I am ok to just remove this assertion. We proved it is not 100% correct.

nullkiller:
Assertion will be removed in scope of another PR

PR #724, opened 2022-01-16
Fix: Get rid of 'Terrain tile is not visible' error message

ShubusCorporation:
Get rid of annoying error message 'Terrain tile is not visible'.

Solution:
1) Do not notify vanquished players interfaces
2) Turn off 'verbose' parameter for enemy heroes because enemy src and/or dst may be not visible for the current player.

### DIFF
--- a/lib/CGameInfoCallback.cpp
+++ b/lib/CGameInfoCallback.cpp
@@ -493,10 +493,12 @@ std::vector<const CGHeroInstance *> CGameInfoCallback::getAvailableHeroes(const 
 
 const TerrainTile * CGameInfoCallback::getTile( int3 tile, bool verbose) const
 {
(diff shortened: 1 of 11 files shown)


lib/CGameInfoCallback.cpp
```diff
@@ -493,10 +493,12 @@ std::vector<const CGHeroInstance *> CGameInfoCallback::getAvailableHeroes(const
 
 const TerrainTile * CGameInfoCallback::getTile( int3 tile, bool verbose) const
 {
-	ERROR_VERBOSE_OR_NOT_RET_VAL_IF(!isVisible(tile), verbose, tile.toString() + " is not visible!", nullptr);
+	if(isVisible(tile))
+		return &gs->map->getTile(tile); //boost::shared_lock<boost::shared_mutex> lock(*gs->mx);
```
nullkiller, line 497: Do we need this lock?

PR #726, opened 2022-03-07
Fix: Crash on startup when mod dependency is not found

ShubusCorporation:
Issue description: Some mods such as 'Tartarus' have implicit dependencies in the 'targetCondition' and other sections. If the dependency is not found, the crash is occurred. 

Fixed: Invalid dependencies should be ignored.

### DIFF
--- a/lib/CModHandler.cpp
+++ b/lib/CModHandler.cpp
@@ -200,20 +200,25 @@ void CIdentifierStorage::registerObject(std::string scope, std::string type, std
 std::vector<CIdentifierStorage::ObjectData> CIdentifierStorage::getPossibleIdentifiers(const ObjectCallback & request)
 {
 	std::set<std::string> allowedScopes;
+	bool isValidScope = true;
 
 	if (request.remoteScope.empty())
 	{
 		// normally ID's from all required mods, own mod and virtual "core" mod are allowed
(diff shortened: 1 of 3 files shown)


lib/CModHandler.cpp
```diff
@@ -200,20 +200,25 @@ void CIdentifierStorage::registerObject(std::string scope, std::string type, std
...
 	{
 		// normally ID's from all required mods, own mod and virtual "core" mod are allowed
-		if (request.localScope != "core" && request.localScope != "")
-			allowedScopes = VLC->modh->getModData(request.localScope).dependencies;
+		if(request.localScope != "core" && !request.localScope.empty())
+		{
+			allowedScopes = VLC->modh->getModDependencies(request.localScope, found);
 
+			if(!found)
+				return std::vector<ObjectData>();
```
nullkiller, line 213: I see that allowedScopes always contains request.localScope and core by default. Are you sure we can not proceed with this instead of short returning here?

ShubusCorporation: `request.localScope` is invalid in this case and can not be processed. I've renamed the `found`  variable to `isValidScope`. I don't see sense to processed it for `core` scope in this case, since input data is invalid. We have an exception here now.

PR #727, opened 2022-03-11
Fix 3184

AgostonSzepessy:
Taken from #684. This fixes a bug with primary skills that results in
heroes not being able to interact with objects, or skills being
incorrect when moving from mission to mission.


nullkiller:
I have tried to bypass campaign from SoD with Yog using vcmisilmaril cheat code and got a crash. Will check a bit later.

PR #728, opened 2022-04-05
Fix RMG to place special dwellings like elemental conflux

nullkiller:
RMG was enable to place special dwellings like Golem Factory and Elemental Conflux.
It was because they are mapped to CREATURE_GENERATOR4

### DIFF
--- a/lib/rmg/CRmgTemplateZone.cpp
+++ b/lib/rmg/CRmgTemplateZone.cpp
@@ -1401,8 +1401,11 @@ bool CRmgTemplateZone::createRequiredObjects()
 		auto tilesBlockedByObject = obj.first->getBlockedOffsets();
 
 		bool finished = false;
-		while (!finished)
+		bool attempt = true;
+		while (!finished && attempt)
 		{
+			attempt = false;
+
 			std::vector<int3> tiles(possibleTiles.begin(), possibleTiles.end());
 			//new tiles vector after each object has been placed, OR misplaced area has been sealed off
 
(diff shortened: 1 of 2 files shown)


lib/rmg/CRmgTemplateZone.cpp
```diff
@@ -2243,41 +2247,48 @@ void CRmgTemplateZone::addAllPossibleObjects()
...
 	{
-		auto dwellingHandler = dynamic_cast<const CDwellingInstanceConstructor*>(VLC->objtypeh->getHandlerFor(Obj::CREATURE_GENERATOR1, secondaryID).get());
-		auto creatures = dwellingHandler->getProducedCreatures();
-		if (creatures.empty())
-			continue;
+		auto subObjects = VLC->objtypeh->knownSubObjects(dwellingType);
 
-		auto cre = creatures.front();
-		if (cre->faction == townType)
+		if(dwellingType == Obj::CREATURE_GENERATOR1)
```
ShubusCorporation, line 2256: What do you think about fixing or creating a wrapper over VLC->objtypeh->knownSubObjects to consider if needed ConfluxROE or not? Thus, the code like this vstd::erase_if_present(subObjects, elementalConfluxROE[i]); will be not needed.

nullkiller: I think h3m map can contain old dwellings, doesn't it? So we still need to be able to construct them but we do not want to put them on the random map?

nullkiller: We probably can add a configuration to json like obsolete: true

nullkiller: This is pure RMG logic not needed for any other place. Might be good to introduce some rmg.blocked=true which will block using this object by RMG in json config but this is a complex task and I do not want to do it now.

PR #730, opened 2022-04-11
Fix a few crashes

nullkiller:
- Workaround crash when reading and modifiing the same instance of std::shared_ptr simultaneously. Want to avoid heavy locking.
- Fix when the battle ends with a spellcast during autofight

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -97,7 +97,9 @@ CBonusProxy::CBonusProxy(const IBonusBearer * Target, CSelector Selector)
 	: bonusListCachedLast(0),
 	target(Target),
 	selector(Selector),
-	bonusList()
+	bonusList(),
+	currentBonusListIndex(0),
+	swapGuard()
 {
 
 }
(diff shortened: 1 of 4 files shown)


lib/HeroBonus.cpp
```diff
@@ -135,18 +144,36 @@ CBonusProxy & CBonusProxy::operator=(CBonusProxy && other)
...
 {
-	if(target->getTreeVersion() != bonusListCachedLast || !bonusList)
+	if(target->getTreeVersion() != bonusListCachedLast || !bonusList[currentBonusListIndex])
 	{
+		boost::lock_guard<boost::mutex>lock(swapGuard);
+
 		//TODO: support limiters
-		bonusList = target->getAllBonuses(selector, Selector::all);
-		bonusListCachedLast = target->getTreeVersion();
+		if(target->getTreeVersion() != bonusListCachedLast || !bonusList[currentBonusListIndex])
```
ShubusCorporation, line 169: It's exactly same condition as in 164 line.

lib/HeroBonus.cpp
```diff
@@ -135,18 +144,36 @@ CBonusProxy & CBonusProxy::operator=(CBonusProxy && other)
...
+
 	return *this;
 }
 
+void CBonusProxy::swapBonusList(TConstBonusListPtr other) const
+{
+	// The idea here is to avoid changing active bonusList while it can be read by a different thread.
+	// Because such use of shared ptr is not thread safe
+	// So to avoid this we change the second offline instance and swap active index
+	auto newCurrent = 1 - currentBonusListIndex;
```
ShubusCorporation, line 157: Is `currentBonusListIndex` should be only `0` or `1` ?
So, may be it would be more clear to define  `currentBonusListIndex` as bool  => it would be 
`auto newCurrent = !currentBonusListIndex;`
https://stackoverflow.com/questions/2725044/can-i-assume-booltrue-int1-for-any-c-compiler

lib/HeroBonus.cpp
nullkiller: The best practice for locking in such cases. First if is unlocked and will most likely cause most of the invokations to go else line. And if it turns out that we do need to update bonusList we enter if body, lock and verify the same condition again but under the lock already. And it might get false because another thread already updated bonusList.

lib/HeroBonus.cpp
nullkiller: Do not think so

PR #733, opened 2022-04-26
Small fixes for AI and launcher

val-gaav:
Fix for AI bug in VCAI and Nullkiller AI . 
Bug report : https://bugs.vcmi.eu/view.php?id=3208

Many years ago I have made resolutions for vcmi 1280x720,  1280x768, 1181x664.
https://forum.vcmi.eu/t/resolution-support-1280x720-1280x768-and-1181x664/1001/10
It seems someone pushed them into vcmi-essentials, but with no support from launcher one can set them only in client and when set they make the launcher crash, so when using them we cannot use launcher at all only client. 

 I've added support for them in launcher code.


nullkiller:
Good job.

PR #734, opened 2022-04-27
vcmibuilder: fix data dir not found on GoG extract

rilian-la-te:
If there is no data dir a.k.a 	data_dir="$temp_dir"/app
found, then GoG vcmibuilder fails. WIth this fix it continues and works.

### DIFF
--- a/vcmibuilder
+++ b/vcmibuilder
@@ -182,6 +182,7 @@ then
 	# innoextract always reports error (iconv 84 error). Just test file for presence
 	test -f "$gog_file" || fail "Error: gog.com executable was not found!"
         gog_file="$(cd "$(dirname "$gog_file")"; pwd)/$(basename "$gog_file")"
+	mkdir -p "$data_dir"
 	cd "$data_dir" && innoextract "$gog_file"
 fi
 


vcmibuilder
```diff
@@ -182,6 +182,7 @@ then
 	# innoextract always reports error (iconv 84 error). Just test file for presence
 	test -f "$gog_file" || fail "Error: gog.com executable was not found!"
         gog_file="$(cd "$(dirname "$gog_file")"; pwd)/$(basename "$gog_file")"
+	mkdir -p "$data_dir"
```
nullkiller, line 185: Lets have it right after data_dir is assigned. It looks more logical this way

rilian-la-te: Done

nullkiller: Great!

rilian-la-te: off-topic: what minimal version of ffmpeg should be supported in VCMI? I want to fix ffmpeg5 support, but do not know, which old code I should preserve.

nullkiller: ffmpeg is used to render bink videos for campaigns. That is all I know about it.

rilian-la-te: So, I can safely assume than versions less than 4.0 is unsupported now? (4.0 is even older than in Debian Buster and Ubuntu 18.04).

PR #737, opened 2022-05-10
CVideoHandler: use new ffmpeg

rilian-la-te:
This is a commit to allow building and working with new FFmpeg (mainly 5.0, but any after 4.1 should be supported).

But this will bump ffmpeg requirement to 4.1 AFAIK.

Fix building with new linux distributions like ArchLinux

### DIFF
--- a/client/CVideoHandler.cpp
+++ b/client/CVideoHandler.cpp
@@ -72,11 +72,6 @@ CVideoPlayer::CVideoPlayer()
 	refreshWait = 0;
 	refreshCount = 0;
 	doLoop = false;
-
-	// Register codecs. TODO: May be overkill. Should call a
-	// combination of av_register_input_format() /
-	// av_register_output_format() / av_register_protocol() instead.
-	av_register_all();
 }
 
 bool CVideoPlayer::open(std::string fname, bool scale)
@@ -127,7 +122,7 @@ bool CVideoPlayer::open(std::string fname, bool loop, bool useOverlay, bool scal
 	stream = -1;
 	for(ui32 i=0; i<format->nb_streams; i++)
 	{
-		if (format->streams[i]->codec->codec_type==AVMEDIA_TYPE_VIDEO)
+		if (format->streams[i]->codecpar->codec_type == AVMEDIA_TYPE_VIDEO)
 		{
 			stream = i;
 			break;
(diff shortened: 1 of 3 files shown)


client/CVideoHandler.cpp
```diff
@@ -266,24 +267,28 @@ bool CVideoPlayer::nextFrame()
...
+								  0, codecContext->height, data, linesize);
 
-						SDL_UpdateYUVTexture(texture, NULL, pict.data[0], pict.linesize[0],
-								pict.data[1], pict.linesize[1],
-								pict.data[2], pict.linesize[2]);
-						avpicture_free(&pict);
+						SDL_UpdateYUVTexture(texture, NULL, data[0], linesize[0],
+								data[1], linesize[1],
+								data[2], linesize[2]);
+						av_freep(&data[0]);
```
nullkiller, line 291: Why do we free only data[0] if we also allocated data[1] and data[2] and they are pointers as well?

client/CVideoHandler.cpp
```diff
@@ -138,18 +133,24 @@ bool CVideoPlayer::open(std::string fname, bool loop, bool useOverlay, bool scal
...
 		return false;
 	}
 
+	codecContext = avcodec_alloc_context3(codec);
+	// Get a pointer to the codec context for the video stream
+	int ret = avcodec_parameters_to_context(codecContext, format->streams[stream]->codecpar);
+	if (ret < 0)
+	{
+		//We cannot get codec from parameters
+		return false;
```
nullkiller, line 151: Do we need to close codeContext here?

nullkiller: I see it will be released by CVideoPlayer destructor anyway

rilian-la-te: Yes, because we allocated context for new ffmpeg earlier. Of we do not close this context, we will get a memory leak.

rilian-la-te: I will add it in next commit, thanks.

client/CVideoHandler.cpp
nullkiller: I see it is according to docs

client/CVideoHandler.cpp
nullkiller: Really looks like it is not needed. Next open will close all previously opened things.

rilian-la-te: nullkiller, I added it in next commit, to avoid confusion. I can remove it again.

nullkiller: Ha-ha, now we have another problem. As far as I understood we can use ffmpeg 5 for macos now? Could you change CI\mac\before_install.sh to use ffmpeg@5?

rilian-la-te: Yes, you can, if there is ffmpeg 5 for Mac. It is you cancelled a pipelines?

rilian-la-te: I will change.

nullkiller: No, mac released ffmpeg@5 formulae and @4 is now keg-only as they say. And it is not enough now to just install it. So lets use ffmpeg5 for mac.

nullkiller: I gave you the path of the file where to fix it

rilian-la-te: Done, let's wait for checks.

PR #744, opened 2022-05-24
CMake presets.

rilian-la-te:
This is just CMake and CI modifications without Conan support.


CMakeLists.txt
```diff
@@ -352,9 +377,29 @@ endif()
 
 
 if(WIN32)
-	file(GLOB dep_files
-		${dep_files}
-		"${CMAKE_FIND_ROOT_PATH}/bin/*.dll")
+	#Conan imports enabled
+	if(DEFINED CONAN_CXX_FLAGS)
```
nullkiller, line 381: Conan is already here

.github/workflows/github.yml
```diff
@@ -8,7 +8,7 @@ on:
     pull_request:
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
-  BUILD_TYPE: Release
+  BUILD_TYPE: RelWithDebInfo
```
nullkiller, line 11: Could you make debug symbols artifact so that in case of crash we can take them and crashdump. Without symbols crash dump is useless

nullkiller: Just gather all pdb files to zip archive.

nullkiller:
Zip structure is a bit extra nested. 2378540199-vcmi-PR-744-ddd5d53 - msvc.zip\out\build\windows-msvc-release

nullkiller:
MXE build is 3 times bigger now. Why? Is it because of RelWithDebInfo?

nullkiller:
> MXE build is 3 times bigger now. Why? Is it because of RelWithDebInfo?

I found an answer. With RelWithDebInfo MXE embeds symbols directly into exe/dll files so I was even able to open crashdump.

rilian-la-te:
@nullkiller why MXE is falling?

.github/workflows/github.yml
rilian-la-te: @nullkiller I can. I will do then.

CMakeLists.txt
rilian-la-te: There is no Conan here, so, I can remove this flag.

nullkiller:
> @nullkiller why MXE is falling?

When you type "crash" in console - it fails. Quite expected behavior. But in overall it works.

CMakeLists.txt
rilian-la-te: fixed

.github/workflows/github.yml
nullkiller: Probably we can have it as a separate review

nullkiller: I mean debug files. *.pdb Probably it is not needed now because we have mxe with debug symbols already onboard

nullkiller: Only MXE is published to downloads so only it is used by Windows community

rilian-la-te: @nullkiller maybe. After this is merged, I want to do 2 new CI jobs. Can I?

rilian-la-te: @nullkiller ok. But this is only after this PR will be merged)

PR #747, opened 2022-05-27
Changes to resolution check on window creation

soarqin:
`SDL_RenderSetLogicalSize()` works fine in fullscreen-desktop mode, so we just use it in window mode as well.

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -381,27 +381,6 @@ int main(int argc, char * argv[])
 				logGlobal->info("\t%s", driverName);
 		}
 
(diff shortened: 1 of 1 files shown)


nullkiller:
I do not understand what it will give. VCMI already can run in fullscreen mode.

nullkiller:
I do not understand what it will give. VCMI already can run in fullscreen mode.

soarqin:
> I do not understand what it will give. VCMI already can run in fullscreen mode.

this PR is not for fullscreen mode but windowed mode

soarqin:
I have rebased and force-pushed the new commit, which adds closest rendering resolution matching

PR #751, opened 2022-06-02
Rmg water support

Nordsoft91:
Moved most of RMG balance-related constants to json file.

Is it possible to change this file via mods?
Hota shall change some of these parameters


config/randomMap.json
```diff
@@ -0,0 +1,59 @@
...
+  "terrain" :
+  {
+    "undergroundAllow" : ["lava"], //others to be replaced by subterranena
+    "groundProhibit" : ["subterranean"] //to be replaced by dirt
+  },
+  "waterZone" :
+  {
+    "treasure" :
+    [
+        { "min" : 2000, "max" : 6000, "density" : 5 },
```
DjWarmonger, line 11: You got the values the other way round: 
> Value 100-1000: Frequency of occurence 5.
> Value 2000-6000: Frequency of occurence 1.

lib/rmg/CMapGenerator.cpp
```diff
@@ -62,10 +62,89 @@ CMapGenerator::CMapGenerator(CMapGenOptions& mapGenOptions, int RandomSeed) :
...
+	loadConfig();
 	rand.setSeed(this->randomSeed);
 	mapGenOptions.finalize(rand);
 }
 
+void CMapGenerator::loadConfig()
+{
+	static std::map<std::string, ETerrainType> terrainMap
+	{
+		{"dirt", ETerrainType::DIRT},
```
nullkiller, line 74: For our own map format all terrains have 2 letter codes like dt, gr and so on. VcmiMapFormat class if I am not mistaken. Should we unify somehow the way we represent terrains in json? Or is it ok to have both 2 letter code and long code for the same terrain? Long codes are obviously more readable

config/randomMap.json
Nordsoft91: Does order make any difference?

Nordsoft91: hm, is there transformation or something? I don't see frequency mentions in code and maps look ok as well.
@DjWarmonger , could you please elaborate more?

nullkiller: Value 2000-6000: Frequency of occurence 1.
vs
 { "min" : 2000, "max" : 6000, "density" : 5 }

Nordsoft91: I've got it! A typo... @DjWarmonger , thanks for catching that, really nice one!

lib/rmg/CRmgTemplate.cpp
```diff
@@ -359,7 +360,7 @@ void ZoneOptions::serializeJson(JsonSerializeFormat & handler)
 			rawStrength = static_cast<decltype(rawStrength)>(zoneMonsterStrength);
 			rawStrength++;
 		}
-		handler.serializeEnum("monsters", rawStrength, STRENGTH);
+		handler.serializeEnum("monsters", rawStrength, EMonsterStrength::ZONE_NORMAL + 1, STRENGTH);
```
Nordsoft91, line 363: Bugfix: until that ZONE_WEAK was default value always (assuming it's not intended)

lib/rmg/CMapGenerator.cpp
```diff
@@ -62,10 +62,89 @@ CMapGenerator::CMapGenerator(CMapGenOptions& mapGenOptions, int RandomSeed) :
...
+		{"dirt_road", ERoadType::DIRT_ROAD},
+		{"gravel_road", ERoadType::GRAVEL_ROAD},
+		{"cobblestone_road", ERoadType::COBBLESTONE_ROAD}
+	};
+	static const ResourceID path("config/randomMap.json");
+	JsonNode randomMapJson(path);
+	for(auto& s : randomMapJson["terrain"]["undergroundAllow"].Vector())
+	{
+		if(!s.isNull())
+			config.terrainUndergroundAllowed.push_back(terrainMap[s.String()]);
```
nullkiller, line 104: A bit of code duplication. Small one so and it will be hard to get rid of it probably. Only extending JsonNode with something like VectorOfInts() VectorOfStrings() might save a few lines of code.

lib/rmg/CRmgTemplateZone.cpp
```diff
@@ -1783,22 +1781,26 @@ bool CRmgTemplateZone::placeMines ()
 			createdMines.push_back(mine);
 			
 			if(!i && (res == ERes::WOOD || res == ERes::ORE))
-				addCloseObject(mine, mineValue.at(res)); //only first woor&ore mines are close
+				addCloseObject(mine, gen->getConfig().mineValues.at(res)); //only first woor&ore mines are close
```
nullkiller, line 1784:
```suggestion
				addCloseObject(mine, gen->getConfig().mineValues.at(res)); //only first wood&ore mines are close
```

lib/rmg/CRmgTemplateZone.cpp
```diff
@@ -1742,14 +1742,14 @@ void CRmgTemplateZone::initTerrainType ()
 
 		//TODO: allow new types of terrain?
 		{
-			if (isUnderground())
+			if(isUnderground())
 			{
-				if (terrainType != ETerrainType::LAVA)
+				if(std::find(gen->getConfig().terrainUndergroundAllowed.cbegin(), gen->getConfig().terrainUndergroundAllowed.cend(), terrainType) == gen->getConfig().terrainUndergroundAllowed.cend())
```
nullkiller, line 1747: Does vstd::contains work here?

PR #752, opened 2022-06-05
Fix hero movement animation

nullkiller:
Fixed hero movement animation, avoid double framerateDelay


client/gui/CGuiHandler.cpp
```diff
@@ -609,30 +609,31 @@ void CFramerateManager::init()
...
 	}
 
+	currentTicks = SDL_GetTicks();
+	// recalculate timeElapsed for external calls via getElapsed()
+	// limit it to 1000 ms to avoid breaking animation in case of huge lag (e.g. triggered breakpoint)
+	timeElapsed = std::min<ui32>(currentTicks - lastticks, 1000);
+
+	lastticks = SDL_GetTicks();
+
 	accumulatedTime += timeElapsed;
```
Nordsoft91, line 630: maybe std::max(1, timeElapsed) could be the good way to avoid any issues

nullkiller: timeElapsed can not be 0 because now we count SDL_Delay

nullkiller: So it will be not less than rateticks

PR #755, opened 2022-06-10
New terrain support - part 1

Nordsoft91:
As change is huge, I decided to pass it through multiple reviews to avoid potential merge conflicts.

Current part supports new terrains and new castles.

This is major change, which means that backward compatibility is violated at some places and code may not be compatible with some mods. However, all features shall work as expected.

Please, use attached mod to test new terrain.

[test_map_terrains-mod.zip](https://github.com/vcmi/vcmi/files/8881179/test_map_terrains-mod.zip)


.github/workflows/github.yml
```diff
@@ -130,7 +130,7 @@ jobs:
           ${{github.workspace}}/**/${{ env.VCMI_PACKAGE_FILE_NAME }}.${{ matrix.extension }}
 
     - name: Upload build
-      if: ${{ matrix.pack == 1 && github.ref == 'refs/heads/develop' && matrix.platform != 'msvc' }}
+      if: ${{ matrix.pack == 1 && (github.ref == 'refs/heads/develop' || startsWith(github.ref, 'refs/heads/features/')) && matrix.platform != 'msvc' }}
```
Nordsoft91, line 133: @nullkiller , if we go with same approach (feature branch), we can keep it. Otherwise will be reverted

AI/Nullkiller/Pathfinding/Actors.cpp
```diff
@@ -70,7 +70,7 @@ int ChainActor::maxMovePoints(CGPathNode::ELayer layer)
 {
 #if AI_TRACE_LEVEL > 0
 	if(!hero)
-		throw std::exception("Asking movement points for static actor");
+		throw std::logic_error("Asking movement points for static actor");
```
Nordsoft91, line 73: std::exception doesn't have non-default constructor. It's bug in CL implementation

client/mapHandler.cpp
```diff
@@ -142,7 +146,7 @@ EMapAnimRedrawStatus CMapHandler::drawTerrainRectNew(SDL_Surface * targetSurface
 
 void CMapHandler::initTerrainGraphics()
 {
-	static const std::vector<std::string> TERRAIN_FILES =
+	/*static const std::vector<std::string> TERRAIN_FILES =
```
Nordsoft91, line 149: will be removed

lib/mapping/CMapDefines.h
```diff
@@ -80,7 +80,7 @@ struct DLL_LINKAGE TerrainTile
 	EDiggingStatus getDiggingStatus(const bool excludeTop = true) const;
 	bool hasFavorableWinds() const;
 
-	ETerrainType terType;
+	CTerrainType terType;
```
alexvins, line 83: This line completely breaks backward compatibility for saves. If that is intentional all compatibility handling code should be cleaned up, `SERIALIZATION_VERSION` & `MINIMAL_SERIALIZATION_VERSION` constants should be changed.
(`SERIALIZATION_VERSION` should be changed even if compatibility is retained)

alexvins (changes requested):

.github/workflows/github.yml
nullkiller: A new branch was created and the PR was retargeted

lib/mapping/CMapDefines.h
nullkiller: I think for such a big feature it is ok to completely break backward compatibility. Any objections? We will merge it to develop only after it is completely ready. For now feature builds will be ok.

ShubusCorporation: I think it's not hard to ensure backward compatibility at this point: if version of BS < current => read old ETerrainType and create new CTerrainType instance using that old -style value, otherwise, just read  CTerrainType object

So, please think about it before merge...

nullkiller (changes requested):

lib/mapping/MapFormatJson.cpp
```diff
@@ -960,12 +960,7 @@ void CMapLoaderJson::readTerrainTile(const std::string & src, TerrainTile & tile
 	{//terrain type
 		const std::string typeCode = src.substr(0, 2);
 
-		int rawType = vstd::find_pos(terrainCodes, typeCode);
-
-		if(rawType < 0)
-			throw std::runtime_error("Invalid terrain type code in "+src);
-
-		tile.terType = ETerrainType(rawType);
+		tile.terType = CTerrainType(typeCode);
```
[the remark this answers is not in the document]
Nordsoft91: How can I test it?

lib/mapping/CMapDefines.h
Nordsoft91: @alexvins , thank you, didn't know about it.
@nullkiller , I think for feature branch we can proceed without backward compatibility, but before merging into develop I'll try to implement something similar to what @ShubusCorporation suggested

client/mapHandler.cpp
```diff
@@ -1432,7 +1444,7 @@ void CMapHandler::getTerrainDescr(const int3 & pos, std::string & out, bool isRM
 		}
 	}
 	if(!isTile2Terrain || out.empty())
-		out = CGI->generaltexth->terrainNames[t.terType];
+		out = CGI->generaltexth->terrainNames[t.terType.id()];
```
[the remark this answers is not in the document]
Nordsoft91: You are right, I tried to get rid from id() method, but it's tricky

lib/CTerrainType.h
```diff
@@ -0,0 +1,77 @@
...
+#include "ConstTransitivePtr.h"
+#include "JsonNode.h"
+
+class DLL_LINKAGE CTerrainType
+{
+public:
+	
+	friend class Manager;
+	
+	class DLL_LINKAGE Manager
```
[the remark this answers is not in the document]
Nordsoft91: It works: rmg doesn't use new terrains while mod is disabled, but if you try to load the game with asphalt it crashes:(

lib/CTownHandler.h
```diff
@@ -385,9 +386,9 @@ class DLL_LINKAGE CTownHandler : public CHandlerBase<FactionID, Faction, CFactio
 	std::vector<BuildingRequirementsHelper> requirementsToLoad;
 	std::vector<BuildingRequirementsHelper> overriddenBidsToLoad; //list of buildings, which bonuses should be overridden.
 
-	const static ETerrainType::EETerrainType defaultGoodTerrain = ETerrainType::EETerrainType::GRASS;
-	const static ETerrainType::EETerrainType defaultEvilTerrain = ETerrainType::EETerrainType::LAVA;
-	const static ETerrainType::EETerrainType defaultNeutralTerrain = ETerrainType::EETerrainType::ROUGH;
+	const static CTerrainType defaultGoodTerrain;
```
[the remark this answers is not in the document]
Nordsoft91: CTownHandler.cpp, 29-31

lib/CTerrainType.cpp
```diff
@@ -0,0 +1,137 @@
...
+int CTerrainType::id() const
+{
+	if(type == "ANY") return -3;
+	if(type == "WRONG") return -2;
+	if(type == "BORDER") return -1;
+	
+	auto _terrains = CTerrainType::Manager::terrains();
+	auto iter = std::find(_terrains.begin(), _terrains.end(), *this);
+	return iter - _terrains.begin();
+}
```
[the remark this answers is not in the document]
Nordsoft91: id is temporary method, I'm thinking on how to get rid of it. Initially I used it almost everywhere, now it's only few places.

lib/mapping/MapFormatJson.cpp
```diff
@@ -960,12 +960,7 @@ void CMapLoaderJson::readTerrainTile(const std::string & src, TerrainTile & tile
 	{//terrain type
 		const std::string typeCode = src.substr(0, 2);
 
-		int rawType = vstd::find_pos(terrainCodes, typeCode);
-
-		if(rawType < 0)
-			throw std::runtime_error("Invalid terrain type code in "+src);
-
-		tile.terType = ETerrainType(rawType);
+		tile.terType = Terrain(src);
```
alexvins, line 963: - `typeCode` is defined in  `namespace TerrainDetail` in this file. It should be moved to terrains.json config and handled somehow.
- `src` is not even a terrain name, it 2 letter terrain code + "terrain view" + "terrain flip"

lib/mapping/MapFormatJson.cpp
```diff
@@ -1298,7 +1293,7 @@ std::string CMapSaverJson::writeTerrainTile(const TerrainTile & tile)
 	out.setf(std::ios::dec, std::ios::basefield);
 	out.unsetf(std::ios::showbase);
 
-	out << terrainCodes.at(int(tile.terType)) << (int)tile.terView << flipCodes[tile.extTileFlags % 4];
+	out << tile.terType.toString().substr(0, 2) << (int)tile.terView << flipCodes[tile.extTileFlags % 4];
```
alexvins, line 1296: Terrain code is not the first 2 letters of terrain name. See `TerrainDetail::terrainCodes`

alexvins (changes requested):

lib/mapping/MapFormatJson.cpp
Nordsoft91: @alexvins , I have no way to debug this code. Please, help with obtaining maps in json format

alexvins: Have you run the test suite?

nullkiller: @alexvins in order to support new terrains everywhere we have to patch the redactor as well. Could you help with it in a later phase when we have json format stable? Or should I consider supporting this activity?

nullkiller: Our redactor is only the way to create a new map with new terrains. Otherwise we can relay only on autogenerated maps

alexvins: > Our redactor is only the way to create a new map with new terrains. Otherwise we can relay only on autogenerated maps

I mean test suite covers map generation, saving to json and loading back It should be enough to test/debug this code

nullkiller: Yes but this does not solve the problem that there is no tool to create an authored map with new terrains. Also later new towns might start using new terrains in their configs and so on so the redactor patch is inevitable at the end.

nullkiller: Probably it is too early to consider this activity so. There is a lot of other work and now we can just use the test suit

nullkiller: I will try to build vcmi_editor from source so and consider to publish a new release binaries for windows

Nordsoft91: @alexvins , do you mean unit tests? If so, unfortunately they are broken even on develop branch.
https://h3vcmi.slack.com/archives/C2J9QFUDS/p1652881185132249

Nordsoft91: What is sense to keep functionality which is not used and not supported? If it used, some maps should be available somewhere:)

nullkiller: Might be that modders are not interested in creating custom authored maps for VCMI only?

Nordsoft91: Right, probably, there should be two-way converter

nullkiller: Two-way not possible. H3M can not handle our new objects. One way exists, I've sent you a few maps to slack

lib/mapping/MapFormatJson.cpp
Nordsoft91: fixed

lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -195,7 +195,7 @@ void CObjectClassesHandler::loadObjectEntry(const std::string & identifier, cons
 	else
 		handler->init(entry);
 
-	if (handler->getTemplates().empty())
+	//if (handler->getTemplates().empty())
```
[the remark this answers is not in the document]
Nordsoft91: It was needed to allow adding templates for existing objects, e.g. mines.
I'm not sure how it works exactly, need testing

lib/mapping/CMapDefines.h
Nordsoft91: Investigation shown that probably we should sacrifice backward compatibility for saves

nullkiller: Yes it is highly likely we should. Anyway we can merge this PR anytime you want as it does not target develop. Just ensure you track somehow all important points spotted that will be addressed later so that they are not get lost.

lib/JsonNode.cpp
```diff
@@ -711,7 +720,7 @@ std::shared_ptr<ILimiter> JsonUtils::parseLimiter(const JsonNode & limiter)
 				{
 					VLC->modh->identifiers.requestIdentifier("terrain", parameters[0], [=](si32 terrain)
 					{
-						terrainLimiter->terrainType = terrain;
+						//terrainLimiter->terrainType = terrain;
```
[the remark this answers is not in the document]
Nordsoft91: I put a comment earlier, see line 723

lib/mapping/MapFormatJson.cpp
val-gaav: Custom map forces users to use certain set mods, while people like to select the mods they use themselves. This is probably why many people may prefer random maps with their own rules/mods. Anyway editor and map format is of course needed and is a must have, it is just that the only way for it to be in real use is if VCMI gets some big mod.   Probably people from MDT team had some plans to do custom maps/campaigns, though it was never published or continued as far as I know.

lib/JsonNode.cpp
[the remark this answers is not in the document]
nullkiller: I am afraid it is quite rare case which might be even never used. Some creature bonus which work on certain terrain only. Usually it is native terrain and exact terrain name will not be specified.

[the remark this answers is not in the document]
Alucard648: OH3 Nomads in army allows hero to ignore terrain movement penalty on desert.

PR #753, opened 2022-06-05
Fix warnings/errors on MSVC

soarqin:
1. update googletest to tag release-1.11.0 to make googletest and googlemock compile on MSVC2022.
2. set gtest_force_shared_crt to ON in test cmake project to make tests compile on MSVC.
3. add /wd4251 and /wd4275 to MSVC compile flags to ignore DLL related warnings for class exports.
4. fix some other warnings and errors while compiling on MSVC2022.

PR #756, opened 2022-06-16
Replace cotire with CMake builtin precompiled headers support

soarqin:
1. This requires CMake version >= 3.16), so CMake option `ENABLE_PCH` is disabled while CMake version < 3.16
2. Unity build is not enabled in cotire use, but now you can simply enable it(but you will get build error at present due to some unity-build unfriendly codes) by passing option `CMAKE_UNITY_BUILD=ON` to command line while on CMake version >= 3.16
3. This also fixes build issue caused by cotire after updating gtest to latest release in #753


nullkiller:
https://en.cppreference.com/w/cpp/language/static_assert
Message is not allowed in C++ 17 standard

nullkiller:
> https://en.cppreference.com/w/cpp/language/static_assert Something changed in C++ 17 standard regarding it.

Probably we need to include <assert.h> to Global.h in order to let static_assert to work

soarqin:
I think the build issue on macOS is that StdInc.h is not recognized as a C++ but C header, will fix it

soarqin:
okay, seems like macOS builds passed

nullkiller:
Good job. Will merge it now

PR #757, opened 2022-06-19
Fix movement freeze after battle in teleport

nullkiller:
Not 100% correct but working fix. If battle happens in teleport the movement considered as failed but destinationTeleport and destinationTeleportPos variables are left uncleared causing further hang during next move.
Now it will be possible to continue movement but teleportation is considered as canceled and you have to reenter teleport again. In original HoMM3 however teleportation succeeds.

### DIFF
--- a/lib/serializer/CTypeList.cpp
+++ b/lib/serializer/CTypeList.cpp
@@ -45,6 +45,21 @@ ui16 CTypeList::getTypeID(const std::type_info *type, bool throws) const
 	return descriptor->typeID;
 }
 
+CTypeList::TypeInfoPtr CTypeList::getTypeDescriptor(ui16 typeID) const
+{
+	auto found = std::find_if(typeInfos.begin(), typeInfos.end(), [typeID](const std::pair<const std::type_info *, TypeInfoPtr> & p) -> bool
+		{
+			return p.second->typeID == typeID;
+		});
+
+	if(found != typeInfos.end())
+	{
+		return found->second;
+	}
+
+	return TypeInfoPtr();
+}
+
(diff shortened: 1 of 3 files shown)


lib/serializer/CTypeList.h
```diff
@@ -148,6 +148,8 @@ class DLL_LINKAGE CTypeList: public boost::noncopyable
 		return getTypeID(getTypeInfo(t), throws);
 	}
 
+	TypeInfoPtr getTypeDescriptor(ui16 typeID) const;
```
alexvins, line 151: What is this function for? Why is it in this PR?
(there is `TypeInfoPtr getTypeDescriptor(const std::type_info *type, bool throws = true) const;` already)

nullkiller: Might be useful for debugging. Allows to get type name by its numeric id

nullkiller: I have not found an existing equivalent

nullkiller: I used it to get type name in requestRealized where we have only numeric id and it is hard to understand what request was actually realized. I can leave some logGlobal->trace there so that the function has at least some usage

lib/serializer/CTypeList.cpp
```diff
@@ -45,6 +45,21 @@ ui16 CTypeList::getTypeID(const std::type_info *type, bool throws) const
 	return descriptor->typeID;
 }
 
+CTypeList::TypeInfoPtr CTypeList::getTypeDescriptor(ui16 typeID) const
+{
+	auto found = std::find_if(typeInfos.begin(), typeInfos.end(), [typeID](const std::pair<const std::type_info *, TypeInfoPtr> & p) -> bool
```
DjWarmonger, line 50: `boost::find_if` is shorter

PR #758, opened 2022-06-20
New battlegrounds

Nordsoft91:
Battlegrounds could be imported from mods as well for now


lib/CHeroHandler.cpp
```diff
@@ -817,28 +817,35 @@ void CHeroHandler::loadExperience()
 
 void CHeroHandler::loadObstacles()
 {
-	auto loadObstacles = [](const JsonNode &node, bool absolute, std::map<int, CObstacleInfo> &out)
+	auto loadObstacles = [](const JsonNode & node, bool absolute, std::vector<CObstacleInfo> & out)
```
nullkiller, line 820: I think obstacles and battlefields related code should be moved from CHeroHandler. They relates to terrains or we can even move them to separate services/handlers. But not in scope of this review

PR #759, opened 2022-06-21
Freeze macos build version to ensure compatibility with older macos v…

Nordsoft91:
Freeze macos build version to ensure compatibility with older macos v…


kambala-decapitator (approved):

kambala-decapitator:
note: ideally, release binary should support at least 10.13+, but for that CI must build all deps from source

PR #761, opened 2022-06-26
allow configurable battleground graphics

nullkiller:
- [x] Add custom terrain patches with custom battlegrounds
- [x] Allow custom battlegrounds for any object
- [x] Allow custom bonuses for terrain patches


lib/Terrain.h
```diff
@@ -42,6 +42,11 @@ class DLL_LINKAGE BattleField
 	{
 		h & name;
 	}
+
+	bool isSpecial() const
+	{
+		return name.find('_') >= 0; // hack for special battlefields, move to JSON
```
Nordsoft91, line 48: Are you going to have parameter instead?

nullkiller: yes, already done

client/battle/CBattleInterface.cpp
```diff
@@ -200,15 +201,9 @@ CBattleInterface::CBattleInterface(const CCreatureSet *army1, const CCreatureSet
...
-		auto bfieldType = curInt->cb->battleGetBattlefieldType();
-		if(!vstd::contains(graphics->battleBacks, bfieldType))
-		{
-			logGlobal->error("%s is not valid battlefield type!", static_cast<std::string>(bfieldType));
-		}
-		else
-		{
-			background = BitmapHandler::loadBitmap(graphics->battleBacks[bfieldType], false);
-		}
+		auto bfieldType = curInt->cb->battleGetBattlefieldType().getInfo();
```
Nordsoft91, line 204: I'd suggest to check fieldType for null

lib/BattleFieldHandler.cpp
```diff
@@ -0,0 +1,108 @@
...
+ *
+ */
+#include "StdInc.h"
+
+#include <vcmi/Entity.h>
+#include "BattleFieldHandler.h"
+
+BattleFieldInfo * BattleFieldHandler::loadFromJson(const std::string & scope, const JsonNode & json, const std::string & identifier, size_t index)
+{
+	BattleFieldInfo * info = new BattleFieldInfo(BattleField(index), identifier);
```
Nordsoft91, line 17: who will delete it?

lib/BattleFieldHandler.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+std::vector<JsonNode> BattleFieldHandler::loadLegacyData(size_t dataSize)
+{
+	return std::vector<JsonNode>();
+}
+
+const std::vector<std::string> & BattleFieldHandler::getTypeNames() const
+{
+	static const  std::vector<std::string> types = std::vector<std::string> { "battlefield" };
```
Nordsoft91, line 70: extra space

test/erm/ERM_BU.cpp
```diff
@@ -174,7 +174,7 @@ TEST_F(ERM_BU_G, Get2)
 	source << "!?PI;" << std::endl;
 	source << "!!BU:G?v1;" << std::endl;
 
-	EXPECT_CALL(binfoMock, battleGetBattlefieldType()).WillOnce(Return(BattleField("evil_fog")));
+	EXPECT_CALL(binfoMock, battleGetBattlefieldType()).WillOnce(Return(BattleField(5)));
```
Nordsoft91, line 177: Could you please add a comment what 5 does mean?

test/erm/ERM_BU.cpp
```diff
@@ -155,7 +155,7 @@ TEST_F(ERM_BU_G, Get)
 	source << "!?PI;" << std::endl;
 	source << "!!BU:G?v1;" << std::endl;
 
-	EXPECT_CALL(binfoMock, battleGetBattlefieldType()).WillOnce(Return(BattleField("snow_trees")));
+	EXPECT_CALL(binfoMock, battleGetBattlefieldType()).WillOnce(Return(BattleField(5)));
```
Nordsoft91, line 158: Could you please add a comment what 5 does mean?

nullkiller:
All unit tests have been fixed locally! Btw we do not run them on CI because they are actually not unit but kind of integration ones and require to boot VLC and other static things unfortunately. And thus they need original game data as well

client/battle/CBattleInterface.cpp
nullkiller: It will write an error to log about unresolved id.

Alucard648:
What about having multiple battlefields, chosen at random, for same terrain/object?

nullkiller:
> What about having multiple battlefields, chosen at random, for same terrain/object?

It is more about graphics probably. We can allow battlefield to have multiple graphics and to chose one at random for instance

PR #762, opened 2022-06-27
Random map generator refactoring

Nordsoft91:
Refactoring has been completed and ready to merge. Some issues could be fixed during testing.

All recommendations, including naming, code style, etc are welcomed

ToDo list:

- [x] low performance for treasure placement
- [x] some blocked treasures
- [x] some unblocked treasures
- [x] ugly water maps
- [x] Rmg shall not crash
- [x] Shipyard placement fails sometimes
- [x] Boat placement fails sometimes
- [x] Incorrect treasure placement for water zone
- [x] Sometimes roads placement fails
- [x] Flipped road tiles on underground
- [x] Feature: river generation
- [x] Feature: improve obstacle placement 
- [x] Feature: obelisks
- [x] Corpse passibility problem
- [x] Distance computation for all area used by treasure pile


lib/rmg/TreasurePlacer.cpp
```diff
@@ -0,0 +1,834 @@
...
+	int maxValue = treasureInfo.max;
+	int minValue = treasureInfo.min;
+	
+	const ui32 desiredValue = generator.rand.nextInt(minValue, maxValue);
+	
+	int currentValue = 0;
+	bool hasLargeObject = false;
+	while(currentValue <= (int)desiredValue - 100) //no objects with value below 100 are available
+	{
+		auto oi = getRandomObject(desiredValue, currentValue, !hasLargeObject);
```
DjWarmonger: Not exactly. Desired value is the **minimal** value of a pile. The max value of a pile remains constant.

Otherwise piles with values, let's say 5000-10000 would almost never generate objects with valueexactly 10000.

Nordsoft91: Thanks! Looks like the bug was presented in "old" rmg as well. Fixed

lib/rmg/ObjectManager.cpp
```diff
@@ -0,0 +1,388 @@
...
+	switch(object.instances().front()->object().ID)
+	{
+		case Obj::TOWN:
+		case Obj::RANDOM_TOWN:
+		case Obj::MONOLITH_TWO_WAY:
+		case Obj::MONOLITH_ONE_WAY_ENTRANCE:
+		case Obj::MONOLITH_ONE_WAY_EXIT:
+		case Obj::SUBTERRANEAN_GATE:
+		case Obj::SHIPYARD:
+		case Obj::MINE:
```
DjWarmonger, line 282: Since when are mines connected with roads?

nullkiller: Was not it your idea to connect them with invisible roads to town to ensure their reachability?

lib/rmg/ObjectManager.cpp
```diff
@@ -0,0 +1,388 @@
...
+	{
+		creId = *RandomGeneratorUtil::nextItem(possibleCreatures, generator.rand);
+		amount = strength / VLC->creh->objects[creId]->AIValue;
+		if (amount >= 4)
+			amount = static_cast<int>(amount * generator.rand.nextDouble(0.75, 1.25));
+	}
+	else //just pick any available creature
+	{
+		creId = CreatureID(132); //Azure Dragon
+		amount = strength / VLC->creh->objects[creId]->AIValue;
```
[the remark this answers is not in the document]
nullkiller: I think CreatureID::AZURE_DRAGON will be better. Also @DjWarmonger do you know why it is exactly azure dragon here? I remember people wanted to ban it with ban mod but were not able due to this hardocde

lib/rmg/ObjectManager.cpp
Nordsoft91: Just for debug purposes, will be removed. Algorithms ensures accessibility for all objects

lib/rmg/RmgArea.h
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+#pragma once
+
+#include "../GameConstants.h"
+#include "../int3.h"
+
+namespace rmg
+{
+	struct HashInt3
```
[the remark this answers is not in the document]
Nordsoft91: It was needed to use unordered_set. At the end I decided to use set. Can be removed

lib/rmg/ObjectManager.cpp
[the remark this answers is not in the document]
DjWarmonger: It's because Azure Dragons are the strongest creatures in game by far, and are a fallback in case that guard strength is extremely high.

lib/rmg/ObjectManager.cpp
Alucard648: What about option to actually place lower quality roads for connection between castles and mines?

Nordsoft91: AFAIK, there are options to enable/disable different road types in hot rmg. Does anybody know what's the logic beyond this?

lib/rmg/TreasurePlacer.cpp
```diff
@@ -0,0 +1,795 @@
...
+			if(accessibleArea.empty())
+			{
+				//fail - fallback
+				rmgObject.clear();
+				return rmgObject;
+			}
+			
+			int3 nextPos = *RandomGeneratorUtil::nextItem(accessibleArea.getTiles(), generator.rand);
+			instance.setPosition(nextPos - rmgObject.getPosition());
+
```
DjWarmonger, line 580: Here is the tricky part. You should check the minimal distance for every new object instance placed, not only the treasure pile starting point. At least this is how I estimted the distance values.
Otherwise one treasure pile might grow in the direction of the other and close the distance.

Nordsoft91: It's easily doable, but it can affect performance.

Nordsoft91: Actually distance is computed for all tiles blocked by object (see ObjectManager.cpp, 53).
But yes, for optimisation purposes during finding position, only initial position of treasure pile is checked

lib/rmg/TreasurePlacer.cpp
```diff
@@ -0,0 +1,795 @@
...
+		};
+		oi.setTemplate(Obj::PRISON, 0, zone.getTerrainType());
+		oi.value = generator.getConfig().prisonValues[i];
+		oi.probability = 30;
+		oi.maxPerZone = generator.getPrisonsRemaning() / 5; //probably not perfect, but we can't generate more prisons than hereos.
+		possibleObjects.push_back(oi);
+	}
+	
+	//all following objects are unlimited
+	oi.maxPerZone = std::numeric_limits<ui32>().max();
```
DjWarmonger, line 119: In other places we've got default values of -1 **and** 1, which is risky at least.

lib/rmg/ObjectManager.cpp
```diff
@@ -0,0 +1,416 @@
...
+	for(const auto & tile : searchArea.getTiles())
+	{
+		obj.setPosition(tile);
+		
+		if(!searchArea.contains(obj.getArea()) || !searchArea.overlap(obj.getAccessibleArea()))
+			continue;
+		
+		float weight = weightFunction(tile);
+		if(weight > bestWeight)
+		{
```
DjWarmonger, line 81: Checking weight is cheap, so you could do this first and only check area fit for tiles better than current.

lib/rmg/TreasurePlacer.cpp
```diff
@@ -0,0 +1,795 @@
...
+					if(ti.getNearestObjectDistance() < minDistance)
+						return -1.f;
+					
+					auto guardedArea = rmgObject.instances().back()->getAccessibleArea();
+					auto areaToBlock = rmgObject.getAccessibleArea(true);
+					areaToBlock.subtract(guardedArea);
+					if(areaToBlock.overlap(zone.freePaths()) || areaToBlock.overlap(manager.getVisitableArea()))
+						return -1.f;
+					
+					return 1.f;
```
DjWarmonger, line 732: Naaah, the real object distance should be returned so you actually pick the most distant location.

As it is now, you still iterate through all the tiles but use the first matching one anyway, am I correct?

lib/rmg/RiverPlacer.cpp
```diff
@@ -0,0 +1,403 @@
...
+	}
+	
+	//make grid
+	for(int j = 0; j < map.map().height; j += 2)
+	{
+		for(int i = 0; i < map.map().width; i += 2)
+		{
+			int3 t{i, j, zone.getPos().z};
+			if(zone.area().contains(t))
+				heightMap[t] += 10.f;
```
nullkiller, line 158: Just an idea. What if we add some random noise here to let rivers be not too straight?

lib/rmg/RmgArea.cpp
```diff
@@ -0,0 +1,395 @@
...
+ *
+ */
+
+#include "RmgArea.h"
+#include "CMapGenerator.h"
+
+namespace rmg
+{
+
+void toAbsolute(Tileset & tiles, const int3 & position)
```
nullkiller, line 17: While you might use it to convert relative shifts to actual tiles this actually shifts/moves tileset towards position. I think in mathematics this operation is called translate. Also would be nice to make this method a member of Tileset as it modifies tiles instance anyway.

nullkiller: So toAbsolute / toRelative might improve readability. Probably we can leave them as is. But still making them members of Tileset might be a better option

lib/rmg/RmgArea.cpp
```diff
@@ -0,0 +1,395 @@
...
+				dBorderCache.insert(t + dTotalShiftCache);
+		}
+	}
+	
+	return dBorderCache;
+}
+
+const Tileset & Area::getBorderOutside() const
+{
+	if(!dBorderOutsideCache.empty())
```
nullkiller, line 157: In some cases it might be faster if getBorderOutside calls getBorder if we use both methods for one tileset quite often

lib/rmg/RmgArea.h
```diff
@@ -0,0 +1,83 @@
...
+		void unite(const Area & area);
+		void intersect(const Area & area);
+		void subtract(const Area & area);
+		void translate(const int3 & shift);
+		
+		friend Area operator+ (const Area & l, const int3 & r); //translation
+		friend Area operator- (const Area & l, const int3 & r); //translation
+		friend Area operator+ (const Area & l, const Area & r); //union
+		friend Area operator* (const Area & l, const Area & r); //intersection
+		friend Area operator- (const Area & l, const Area & r); //AreaL reduced by tiles from AreaB
```
nullkiller, line 67:
```suggestion
		friend Area operator- (const Area & l, const Area & r); //AreaL reduced by tiles from AreaR
```

lib/rmg/RmgArea.cpp
nullkiller: Or even better we can calculate both vectors in one pass! Not sure which way is better. Depends on how often we use both border and border-outside for the same Area

lib/rmg/RmgArea.cpp
```diff
@@ -0,0 +1,395 @@
...
+	if(!dBorderCache.empty())
+		return dBorderCache;
+	
+	//compute border cache
+	for(auto & t : dTiles)
+	{
+		for(auto & i : int3::getDirs())
+		{
+			if(!dTiles.count(t + i))
+				dBorderCache.insert(t + dTotalShiftCache);
```
nullkiller, line 148: Technically we can break inner loop here. It might save up to 50% of hits to dTiles.count()

lib/rmg/RmgArea.cpp
```diff
@@ -0,0 +1,395 @@
...
+{
+	if(dTotalShiftCache != int3())
+	{
+		toAbsolute(dTiles, dTotalShiftCache);
+		dTotalShiftCache = int3();
+	}
+	return dTiles;
+}
+
+const std::vector<int3> & Area::getTilesVector() const
```
nullkiller, line 127: If we getTiles while dTotalShiftCache is not int3() we refresh tiles and retrieve their correct values. But when we getTilesVector we do not check dTotalShiftCache  and thus can get old data not shifted. This might be hard to trace back taking into account that getTilesVector can be called from various functions like overlap()

lib/rmg/RmgArea.cpp
```diff
@@ -0,0 +1,395 @@
...
+
+void Area::translate(const int3 & shift)
+{
+	dBorderCache.clear();
+	dBorderOutsideCache.clear();
+	
+	if(dTilesVectorCache.empty())
+	{
+		getTiles();
+		getTilesVector();
```
nullkiller, line 342: This calls getTiles() already when cache is empty

lib/rmg/RiverPlacer.cpp
Alucard648: Same idea goes for zone borders. I don`t like straight zone borders in maps generated by RMGs in Heroes 5 - 7.

nullkiller: Borders are a different question. Not sure even if this simple idea with noise will even work.

lib/rmg/CMapGenerator.cpp
```diff
@@ -128,49 +89,19 @@ const CMapGenerator::Config & CMapGenerator::getConfig() const
 	return config;
 }
 
-void CMapGenerator::initTiles()
+CMapGenerator::~CMapGenerator()
```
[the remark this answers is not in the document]
Nordsoft91: Yes, destructor will be called for all data members by default (same as default constructor will be called for data member no constructors explicitly)

lib/rmg/CMapGenerator.h
```diff
@@ -16,36 +16,17 @@
 #include "../int3.h"
 #include "CRmgTemplate.h"
 
-class CMap;
+#define _BETA
```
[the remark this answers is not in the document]
Nordsoft91: There is some dump code, which must be removed/disabled prior merging into develop

lib/rmg/ConnectionsPlacer.cpp
```diff
@@ -0,0 +1,279 @@
...
+				assert(otherZone->getModificator<ConnectionsPlacer>());
+				otherZone->getModificator<ConnectionsPlacer>()->otherSideConnection(connection);
+				
+				success = true;
+			}
+		}
+	}
+	
+	//2. connect via water
+	bool waterMode = map.getMapGenOptions().getWaterContent() != EWaterContent::NONE;
```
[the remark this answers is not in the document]
Nordsoft91: Yes, otherwise it will block all coast tiles for normal water mode

lib/rmg/ObstaclePlacer.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+void ObstaclePlacer::process()
+{
+	auto * manager = zone.getModificator<ObjectManager>();
+	if(!manager)
+		return;
+	
+	auto * riverManager = zone.getModificator<RiverPlacer>();
+	
+	typedef std::vector<ObjectTemplate> obstacleVector;
```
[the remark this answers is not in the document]
Nordsoft91: Yes, will fix

lib/rmg/ObstaclePlacer.cpp
```diff
@@ -0,0 +1,187 @@
...
+				auto obj = handler->create(temp);
+				allObjects.emplace_back(*obj);
+				rmg::Object * rmgObject = &allObjects.back();
+				for(auto & offset : obj->getBlockedOffsets())
+				{
+					rmgObject->setPosition(tile - offset);
+					if(!map.isOnMap(rmgObject->getPosition()))
+						continue;
+					
+					if(!rmgObject->getArea().getSubarea([this](const int3 & t)
```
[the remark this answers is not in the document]
Nordsoft91: Theoretically you can have part of object to be outside of map. And it could be expected (like you may see only part of large mountain). If not, there could be negative aesthetic effect - small obstacles on the map edge only

lib/rmg/RiverPlacer.cpp
Nordsoft91: I tried to make it but result was strange. Probably, rivers need some efforts to look nice

Nordsoft91: Zone borders could be fractalized - I have plans to do it later

lib/rmg/RmgArea.cpp
Nordsoft91: Computing them together could be interesting option, let me investigate if we could save something

lib/rmg/RmgArea.cpp
Nordsoft91: That was the idea - do not refresh set while translating to avoid multiple set sorting. So getTilesVector should be "life updated" while getTiles - lazy cache mechanism. If you modify area - you invalidate getTilesVector. If you use translation only - you invalidate getTiles. That's the fastest strategy I was able to do. But maybe I understood you point incorrectly

nullkiller: Technically it seams to me that translated set should still be in sorted order without additional sorting. So if we can update std::set values and avoid sorting itself this might be an option as well. Now it might be quite ease to misuse these methods because they look quite the same but have a little difference in behavior.

nullkiller: I understood. Checked how translate() works.

lib/rmg/RmgPath.cpp
```diff
@@ -0,0 +1,196 @@
...
+			while (cameFrom[backTracking].valid())
+			{
+				result.dPath.add(backTracking);
+				backTracking = cameFrom[backTracking];
+			}
+			return result;
+		}
+		else
+		{
+			auto foo = [&open, &closed, &cameFrom, &currentNode, &distances, &moveCostFunction, &result](const int3& pos) -> void
```
nullkiller, line 112: foo - lets find a better name

lib/rmg/RmgPath.cpp
```diff
@@ -0,0 +1,196 @@
...
+			{
+				if(closed.count(pos))
+					return;
+				
+				if(!result.dArea->contains(pos))
+					return;
+				
+				float movementCost = moveCostFunction(currentNode, pos) + currentNode.dist2d(pos);
```
nullkiller, line 120: currentNode.dist2d(pos) will return either 1 or 1.4 because these nodes are neighbours. Described A* algorithm has gScore and fScore where fScore is estimated distance from source to target through pos as far as I understood. But I do not see where is it implemented, am I missing something?
Btw does it search a new Path from current Path to dst or vise versa from dst to the current path?

lib/rmg/RmgPath.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+	Area a(path);
+	dPath.unite(a);
+	//dArea.unite(a);
+}
+
+void Path::connect(const Area & path)
+{
+	dPath.unite(path);
+	//dArea.unite(path);
```
nullkiller, line 184: Do we need this commented code?

lib/rmg/TileInfo.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+bool TileInfo::shouldBeBlocked() const
+{
+	return occupied == ETileType::BLOCKED;
+}
+bool TileInfo::isBlocked() const
+{
+	return occupied == ETileType::BLOCKED || occupied == ETileType::USED;
+}
+bool TileInfo::isPossible() const
```
nullkiller, line 35: What is the difference between free and possible?

lib/rmg/TreasurePlacer.cpp
```diff
@@ -0,0 +1,795 @@
...
+		if(oi->templ.isVisitableFromTop())
+		{
+			objectInfos.push_back(oi);
+		}
+		else
+		{
+			if(hasLargeObject)
+			{
+				//only one large object per treasure pile
+				continue;
```
nullkiller, line 536: Can we get into endless loop here?

lib/rmg/WaterAdopter.cpp
```diff
@@ -0,0 +1,260 @@
...
+}
+
+void WaterAdopter::createWater(EWaterContent::EWaterContent waterContent)
+{
+	if(waterContent == EWaterContent::NONE || zone.isUnderground() || zone.getType() == ETemplateZoneType::WATER)
+		return; //do nothing
+	
+	distanceMap = zone.area().computeDistanceMap(reverseDistanceMap);
+	
+	//add border tiles as water for ISLANDS
```
nullkiller, line 48: Instead of putting comments you can create methods

lib/rmg/WaterRoutes.cpp
```diff
@@ -0,0 +1,115 @@
...
+				if(map.isOnMap(c) && map.getZoneID(c) != zone.getId())
+				{
+					landTiles.push_back(c - t);
+				}
+			});
+			
+			if(landTiles.size() == 2)
+			{
+				int3 o = landTiles[0] + landTiles[1];
+				if(o.x * o.x * o.y * o.y == 1)
```
nullkiller, line 69: Hard to understand this, what does it mean? Why we check exactly 0 and 1?

nullkiller: Is it possible that this expression will be 0 or 4?

lib/rmg/Zone.h
```diff
@@ -0,0 +1,142 @@
...
+
+#include "../GameConstants.h"
+#include "float3.h"
+#include "../int3.h"
+#include "CRmgTemplate.h"
+#include "RmgArea.h"
+#include "RmgPath.h"
+#include "RmgObject.h"
+
+#define RMG_DUMP
```
nullkiller, line 21: Do we need it now? I mean dumping enabled after merge? I usually disable such things before merge to avoid performance impact.

lib/rmg/WaterRoutes.cpp
Nordsoft91: It checks for corner pattern. It could be zero, it water is 1-tile width - we shouldn't block placement in that case

lib/rmg/TreasurePlacer.cpp
Nordsoft91: Good point. If all objects are large we may get into infinite cycle. Will thing how to fix

Nordsoft91: getRandomObject does same check, probably I can get rid of this condition

lib/rmg/TileInfo.cpp
Nordsoft91: free - tile must be passable
possible - tile can be blocked or passable

lib/rmg/RmgPath.cpp
Nordsoft91: fScore is what called distance here and stored in distances map if I understood this.
As for path, it's starting search from dst to current Path and then backtracking result path from Path to dst. Overall it doesn't matter as it stored as std::set - just set of tiles connecting src and dot

lib/rmg/RmgArea.cpp
Nordsoft91: >  translated set should still be in sorted order without additional sorting

I don't know how to do it. Each time when you modify element in set, you will have resorting. In the end you will have same order, but during modifications there will be a lot of extra operations

lib/rmg/ObstaclePlacer.cpp
```diff
@@ -0,0 +1,187 @@
...
+						obstaclesBySize[temp.getBlockedOffsets().size()].push_back(temp);
+				}
+			}
+		}
+	}
+	for(auto o : obstaclesBySize)
+	{
+		possibleObstacles.push_back(o);
+	}
+	boost::sort(possibleObstacles, [](const obstaclePair &p1, const obstaclePair &p2) -> bool
```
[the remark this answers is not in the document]
Nordsoft91: Probably they are, but I'm inclined to keep it for explicit as you mentioned (this part is inherited from original rmg)
Anyway sorting of sorted array is o(n)

lib/rmg/TreasurePlacer.cpp
Nordsoft91: right, placeAndConnectObject has optimizer argument which defines behavior: true is to look for best position (maximise functional), false is to break on first positive value of the functional. I use false here in order for performance reasoning

lib/rmg/ObjectManager.cpp
Nordsoft91: only positive values of weight function are considered as good positions

Nordsoft91: I'm not sure if weightFunction is cheaper rather than area checks

lib/rmg/TreasurePlacer.cpp
Nordsoft91: Do you propose to use -1 instead?

lib/rmg/TreasurePlacer.cpp
DjWarmonger: First positive value leads to non-uniform distribution. Objects are placed from the top to bottom (and from left to right?), which leads to dense top part of the zone and often empty bottom.

Also, notice that placing object at most distant location leads to convergence since distances around objects get reduced quickly.

I already figured out an algorithm to choose distances quickly with a priority queue, but this can wait for another day.

Nordsoft91: That's true, you pointed on that issue with template marathon. But even with enabled optimiser it will create regular structures which may be not so nice. I think maybe adding some random value would be ok.
Do you think it's regression relatively previous algorithm?

DjWarmonger: Certainly it's a regression if zones end up empty. In fact it's the major bug in current build.

The idea was to place large treasures at large distances from each other first, then fill the gaps with smaller treasures and higher density.

Nordsoft91: I found the problem and made a fix. It was related to fractalization algorithm, not the treasure placer.

PR #766, opened 2022-07-21
Fix crash

Nordsoft91:
Crash was rare:
 - fight with some monster, pointing exactly on tile with monster
 - escape from battle at first move
 - observe crash

Root cause: line adventureInt->heroList.update(hero) calls currentPath->size(), but currentPath is invalid because paths.erase already happened.

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -443,14 +443,15 @@ void CPlayerInterface::heroKilled(const CGHeroInstance* hero)
 	}
 
 	wanderingHeroes -= hero;
-	if (vstd::contains(paths, hero))
-		paths.erase(hero);
 
 	adventureInt->heroList.update(hero);
 	if (makingTurn && newSelection)
 		adventureInt->select(newSelection, true);
 	else if (adventureInt->selection == hero)
 		adventureInt->selection = nullptr;
+	
+	if (vstd::contains(paths, hero))
+		paths.erase(hero);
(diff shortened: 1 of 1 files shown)


nullkiller:
Why does adventureInt->heroList.update(hero) want to access currentPath of removed hero, can we handle that case there exactly by checking hero's owner still match the player color or hero is still alive? Now the fix is quite indirect.

Nordsoft91:
Making hero check in rendering function for particular tile for me is not appropriate - it's better to reduce game logic from rendering functions, rather that mix them. Getting access from particular tile to currently selected hero looks much more complicated. The fix is straightforward - do not delete objects which are used by pointers. Actual destructors shall be called at the end.

Nordsoft91:
This happens in void CTerrainRect::showPath(const SDL_Rect * extRect, SDL_Surface * to) (CAdvmapInterface.cpp, 268) - checking hero here is odd

nullkiller:
Can we just clear currentPath if the hero is current selection?

Nordsoft91:
I think it's not doable in that architecture: currentPath is a property of **tile**. It doesn't contain any game information.

Nordsoft91:
In order to do that we have to iterate through all tiles in the map and every tile, which currentPath == path, change to nullptr. And I'm not sure about safety for that solution. Also it will create more dependencies

nullkiller:
Now I agree.

PR #770, opened 2022-08-03
fix error C2338 when building Client on Windows

kambala-decapitator:
Error message comes from winnt.h and is:

> Windows headers require the default packing option. Changing this can lead to memory corruption. This diagnostic can be disabled by building with WINDOWS_IGNORE_PACKING_MISMATCH defined.

I was building on Windows 8.1 32-bit with MSVC 2019 x86 and Windows SDK 10.0.19041.0.


kambala-decapitator:
macOS build failure can be ignored because the change touches only Windows

PR #775, opened 2022-08-13
Improve treasure placement logic

Nordsoft91:
In order to place treasures in optimal way, I keep priority queue containing distance to object.
When placer works, it starts from most distant tiles instead of top-down order.


lib/rmg/ObjectManager.cpp
```diff
@@ -65,33 +77,64 @@ const rmg::Area & ObjectManager::getVisitableArea() const
...
+	if(optimizer & OptimizeType::DISTANCE)
 	{
-		obj.setPosition(tile);
-		
-		if(!searchArea.contains(obj.getArea()) || !searchArea.overlap(obj.getAccessibleArea()))
-			continue;
-		
-		float weight = weightFunction(tile);
-		if(weight > bestWeight)
+		auto open = tilesByDistance;
```
nullkiller, line 87: What does this collection contain?

lib/rmg/TreasurePlacer.h
```diff
@@ -52,7 +52,7 @@ class TreasurePlacer: public Modificator
 	
 	ObjectInfo * getRandomObject(ui32 desiredValue, ui32 currentValue, ui32 maxValue, bool allowLargeObjects);
 	std::vector<ObjectInfo*> prepareTreasurePile(const CTreasureInfo & treasureInfo);
-	rmg::Object constuctTreasurePile(const std::vector<ObjectInfo*> & treasureInfos);
+	rmg::Object constuctTreasurePile(const std::vector<ObjectInfo*> & treasureInfos, bool densePlacement = false);
```
nullkiller, line 55:
```suggestion
	rmg::Object constructTreasurePile(const std::vector<ObjectInfo*> & treasureInfos, bool densePlacement = false);
```

lib/rmg/ObjectManager.cpp
Nordsoft91: It's a copy of queue containing tiles from furthest to nearest. It's precomputed so I store it as class member

nullkiller: Is it always precomputed? Any chance to misuse this method with not up-to-date cache?

nullkiller: I suggest to introduce some USE_CACHE optimizer flag and if it is not set - we always recalculate cache in this method when we need tilesByDistance. This will at least avoid logical errors in future  when you just change optimization type and it does not work as expected

Nordsoft91: It will completely crash the performance. There is method updateDistances() which is called only when new object is placed on the map. If you don't place objects - no need to recompute distance field.

Nordsoft91: this method findPlaceForObject() is marked as const

PR #777, opened 2022-08-21
Fix for #3218

Nordsoft91:
https://bugs.vcmi.eu/view.php?id=3218

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -948,7 +948,7 @@ void CBonusSystemNode::getAllBonusesRec(BonusList &out) const
 	getAllParents(lparents);
 
 	for(auto parent : lparents)
-		parent->bonuses.getAllBonuses(beforeUpdate);
+		parent->getAllBonusesRec(beforeUpdate);
 
 	bonuses.getAllBonuses(beforeUpdate);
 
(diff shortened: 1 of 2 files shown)


lib/registerTypes/RegisterTypes.h
```diff
@@ -140,11 +140,11 @@ void registerTypesMapObjectTypes(Serializer &s)
 	s.template registerType<IUpdater, GrowsWithLevelUpdater>();
 	s.template registerType<IUpdater, TimesHeroLevelUpdater>();
 	s.template registerType<IUpdater, TimesStackLevelUpdater>();
+	s.template registerType<IUpdater, OwnerUpdater>();
```
nullkiller, line 143: Are you sure it will not break serialization?

Nordsoft91: It will. Good catch. I'll move this commit into new-terrains feature branch where we don't have backward compatibility anyway

lib/HeroBonus.cpp
```diff
@@ -948,7 +948,7 @@ void CBonusSystemNode::getAllBonusesRec(BonusList &out) const
 	getAllParents(lparents);
 
 	for(auto parent : lparents)
-		parent->bonuses.getAllBonuses(beforeUpdate);
+		parent->getAllBonusesRec(beforeUpdate);
```
nullkiller, line 951: Does getAllParents return direct parents or grandparents as well? Looks like all of them and parent->getAllBonusesRec will do the same so it looks like grandparent will be iterated twice and this might cause same bonus to be activated a few times which might be wrong for additive bonuses.

Nordsoft91: That's how it worked exactly before the changes from Sep 2021. I had same concern, but looks like duplicated bonuses will be discarded later in limiters as it will be processed for improper parent.
I've tested it a lot and didn't discover issues (but I'm bad in QA:)) do you have scenarios in mind which need to be verified?

Nordsoft91: It's not the ideal but it fixes regression (as for me missing specialities is a blocker for the release)

PR #779, opened 2022-08-22
Fix RMG bug with incorrect town placement

Nordsoft91:
Due to empty template town was placed by algorithm as it occupies only one tile.

I moved tweak from https://github.com/vcmi/vcmi/pull/777 as it breaks backward compatibility. This branch breaks it anyway


Nordsoft91:
I have no idea why is that so many changes, most of them have been already merged in https://github.com/vcmi/vcmi/pull/775

lib/registerTypes/RegisterTypes.h
```diff
@@ -142,11 +142,11 @@ void registerTypesMapObjectTypes(Serializer &s)
 	s.template registerType<IUpdater, GrowsWithLevelUpdater>();
 	s.template registerType<IUpdater, TimesHeroLevelUpdater>();
 	s.template registerType<IUpdater, TimesStackLevelUpdater>();
+	s.template registerType<IUpdater, OwnerUpdater>();
```
Nordsoft91, line 145: Change moved from https://github.com/vcmi/vcmi/pull/777

lib/rmg/TownPlacer.cpp
```diff
@@ -138,11 +138,12 @@ int3 TownPlacer::placeMainTown(ObjectManager & manager, CGTownInstance & town)
 {
 	//towns are big objects and should be centered around visitable position
 	rmg::Object rmgObject(town);
+	rmgObject.setTemplate(zone.getTerrainType());
```
Nordsoft91, line 141: The actual fix for town placement bug

PR #781, opened 2022-08-23
Fix start game button for rmg without templates

Nordsoft91:
1. The story: I tried to install vcmi on the new machine but It was crashed at random moments (but at most cases right after connecting to the server). Investigating that I understood that the problem is in data resources or in its parsing (despite the fact that original game works fine at the same computer).
So I've decided to make small just to have small hint in the log. It creates purely negative impression for the new people who just want to try vcmi and observing crash without any reason.

We can discuss stricter rule - such as throwing an exception or something, but record in the log helps a lot.

2. I fixed exception catching mechanism. Steps to reproduce a problem:
   a. Disable all rmg template mods from launcher
   b. Run client and try to generate any random map
Expected result: uses sees window with error about missing templates
Actual result: nothing happens

### DIFF
--- a/client/lobby/CLobbyScreen.cpp
+++ b/client/lobby/CLobbyScreen.cpp
@@ -131,25 +131,19 @@ void CLobbyScreen::startScenario(bool allowOnlyAI)
 		CSH->sendStartGame(allowOnlyAI);
 		buttonStart->block(true);
 	}
-	catch(ExceptionMapMissing & e)
+	catch(std::exception & e)
 	{
-		(void)e;	// unused
-	}
-	catch(ExceptionNoHuman & e)
-	{
-		(void)e;	// unused
-		// You must position yourself prior to starting the game.
-		CInfoWindow::showInfoDialog(std::ref(CGI->generaltexth->allTexts[530]), CInfoWindow::TCompsInfo(), PlayerColor(1));
-	}
-	catch(ExceptionNoTemplate & e)
-	{
-		(void)e; // unused
-		// Could not create a random map that fits current choices.
-		CInfoWindow::showInfoDialog(std::ref(CGI->generaltexth->allTexts[751]), CInfoWindow::TCompsInfo(), PlayerColor(1));
+		logGlobal->error("Exception during startScenario: %s", e.what());
+		
+		if(std::string(e.what()) == "ExceptionNoHuman")
+			CInfoWindow::showInfoDialog(std::ref(CGI->generaltexth->allTexts[530]), CInfoWindow::TCompsInfo(), PlayerColor(1));
+		
+		if(std::string(e.what()) == "ExceptionNoTemplate")
+			CInfoWindow::showInfoDialog(std::ref(CGI->generaltexth->allTexts[751]), CInfoWindow::TCompsInfo(), PlayerColor(1));
 	}
 	catch(...)
 	{
-
+		logGlobal->error("Unknown exception");
 	}
 }
 
--- a/lib/CGeneralTextHandler.cpp
+++ b/lib/CGeneralTextHandler.cpp
@@ -493,6 +493,28 @@ CGeneralTextHandler::CGeneralTextHandler()
 			logGlobal->warn("WoG file ZNPC00.TXT containing commander texts was not found");
 		}
 	}
+	
+	//sanity check: in most places we get strings using operator[] so it may lead to crash if data is not valid
+	//use highest hardcoded index to check
+	if(allTexts.size() < 741 ||
+	   arraytxt.size() < 232 ||
+	   primarySkillNames.size() < 4 ||
+	   jktexts.size() < 44 ||
+	   heroscrn.size() < 33 ||
+	   overview.size() < 16 ||
+	   colors.size() < 8 ||
+	   capColors.size() < 8 ||
+	   tcommands.size() < 33 ||
+	   hcommands.size() < 9 ||
+	   fcommands.size() < 7 ||
+	   tavernInfo.size() < 8 ||
+	   zelp.size() < 614 ||
+	   advobtxt.size() < 91 ||
+	   restypes.size() < 7 ||
+	   mines.size() < 7)
+	{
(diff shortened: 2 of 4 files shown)


Nordsoft91:
I can replace all calls with .at instead of [] as well but probably now it's not the best time to do it

lib/CGeneralTextHandler.cpp
```diff
@@ -493,6 +493,28 @@ CGeneralTextHandler::CGeneralTextHandler()
...
+	   tcommands.size() < 33 ||
+	   hcommands.size() < 9 ||
+	   fcommands.size() < 7 ||
+	   tavernInfo.size() < 8 ||
+	   zelp.size() < 614 ||
+	   advobtxt.size() < 91 ||
+	   restypes.size() < 7 ||
+	   mines.size() < 7)
+	{
+		logGlobal->error("YOUR ORIGINAL HOMM3 DATA IS CORRUPTED, GAME MAY CRASH");
```
nullkiller, line 516: It might be because of mods as well. They can override original files.

nullkiller: Btw using your approach we can iterate all files of the same name and discard ones with bad number of lines with a warning

Nordsoft91: You are right, I need to change the message.

client/lobby/CLobbyScreen.cpp
```diff
@@ -131,25 +131,19 @@ void CLobbyScreen::startScenario(bool allowOnlyAI)
...
-		CInfoWindow::showInfoDialog(std::ref(CGI->generaltexth->allTexts[530]), CInfoWindow::TCompsInfo(), PlayerColor(1));
-	}
-	catch(ExceptionNoTemplate & e)
-	{
-		(void)e; // unused
-		// Could not create a random map that fits current choices.
-		CInfoWindow::showInfoDialog(std::ref(CGI->generaltexth->allTexts[751]), CInfoWindow::TCompsInfo(), PlayerColor(1));
+		logGlobal->error("Exception during startScenario: %s", e.what());
+		
+		if(std::string(e.what()) == "ExceptionNoHuman")
```
nullkiller, line 138: Might be worth to introduce constants. Some RmgExceptions::NoHuman

lib/StartInfo.cpp
```diff
@@ -64,7 +64,7 @@ std::string StartInfo::getCampaignName() const
 void LobbyInfo::verifyStateBeforeStart(bool ignoreNoHuman) const
```
nullkiller, line 64: Interesting if we can return an error enum instead of throwing it. So that thing we expect we return and things we do not expect we throw

Nordsoft91: there are few places using this function but exception catching performed only in one place. It requires the refactoring, but I would prefer to postpone it and minimize changes for the release.

lib/CGeneralTextHandler.cpp
```diff
@@ -493,6 +493,28 @@ CGeneralTextHandler::CGeneralTextHandler()
...
+	   tcommands.size() < 33 ||
+	   hcommands.size() < 9 ||
+	   fcommands.size() < 7 ||
+	   tavernInfo.size() < 8 ||
+	   zelp.size() < 614 ||
+	   advobtxt.size() < 91 ||
+	   restypes.size() < 7 ||
+	   mines.size() < 7)
+	{
+		logGlobal->error("GeneralTextHandler: GAME DATA IS CORRUPTED, GAME MAY CRASH!");
```
nullkiller, line 516: lets find a way to specify which data is corrupted and how many lines do we expect. It is very close to assert(mines.size() < 7, "Expected to have at least 7 mines");

nullkiller: But assert() itself might be disabled in release builds so we need some other kind of guard. verify() for instance

Nordsoft91: That's obvious thing, I could preform multiple checks with different messages but haven't done it intentionally. It's sanity check used for troubleshooting, it's not a tool for debugging or development.
It's a purely result of bad ideas in many places of the code (I mean operator[] for vector with hardcoded numbers without checking it's size)

Now let's think how it will be used. I believe if can help with support when people come and ask why game doesn't start or crash at random moment. Usually we ask for logs, which are useless in such type of problems. Now we have troubleshooting algorithm: see message in log? -yes-> do you use custom mods? -yes-> disable them to check if problem there. Problem disappear? -no-> you have corrupted h3data.
But I don't understand what will be use case for "expected to have 7 lines of mines". 
We have to identify and highlight problems such those gettext[356] instead of making comfortable to stay with them

client/lobby/CLobbyScreen.cpp
Nordsoft91: I like more the approach you suggested below (with error codes), but again, not for now, it's a refactoring work for a future

lib/CGeneralTextHandler.cpp
nullkiller: You can show generic message + some detailed one. So if you see it you can understand which file you should check

Nordsoft91: I reproduced problem on original data - there is no sense to change something in mod archive. Actually I have doubt that mods can lead to this situation because VLC will merge files, not override them

Nordsoft91: The actual fix will be do not crash at all.

nullkiller: VLC merge only things which are described with content handlers afaik. Or does it do anything special for text files as well? I remember I fixed some translation file which had a few translations missed for instance and it caused crash. But it was some JSON

Nordsoft91: it should merge texts, all texts are stored in few std::vector containers. With merge you can increase size of the container, but not decrease

Nordsoft91: reverted

PR #741, opened 2022-05-18
Alternative CI implementations

rilian-la-te:
1. Conan (conan.io) - for Windows and Mac
2. Pacman - for MXE
3. MSYS - for Windows

WIP for now.

PR #782, opened 2022-08-24
[WIP] use Conan in macOS CI builds

kambala-decapitator:
@rilian-la-te pinging you as you worked on #741


rilian-la-te:
@kambala-decapitator yes, I was worked on it, but it seems than there is no clue to merging it. 

I will look for this patches or maybe just integrate macos conan to #741

kambala-decapitator:
@rilian-la-te please note that I'm using a more advanced approach with python file for a fine-grained control, also I've stripped dependencies as much as possible using their options

rilian-la-te:
@kambala-decapitator can you merge my Windows work into this PR?

Second, I think for Conan separate CMake presets should be used.

kambala-decapitator:
> @kambala-decapitator can you merge my Windows work into this PR?

it should be tested on other OS first :) let's do it in a separate PR.

main goal for macOS is to get rid of brew-installed dependencies to allow deploying to older OS versions. This is getting more important due to https://github.com/actions/runner-images/issues/5583 (but right now I see that they have recently postponed the decision :) )

Also, with new approach we're able to produce native ARM builds.

> Second, I think for Conan separate CMake presets should be used.

it would make sense for "alternative CI implementations" like your PR. macOS will have the only CI alternative - Conan.

rilian-la-te:
@kambala-decapitator for MacOS it is Brew and Conan. Conan should be alternative for all OS, and other tools should be native instuments.

kambala-decapitator:
you're mixing local development and release on CI. yes, brew for local development is perfect and should stay like that. However, for CI it's a bad choice for reasons described above.

or maybe I'm misunderstanding you? Is your point to have the release presets being usable on local machine for w/e reason, e.g. one just wants to produce a release build to play it? If this, then you have a point, will create new preset.

rilian-la-te:
1. VCMI should be perfectly installable wihout a dependencies, as there will be brew, or debian packages, or something other.
2. VCMI should be workable with or without system impact (eg, Conan, or local dependencies - should not matter) for local development
3. CMake presets is both for CI and for local building and development, so, we should not ban brew in presets for this reasons. So, better to add a Conan preset and inherit presets in there (as I did in PR mentioned above)

kambala-decapitator:
now it's fully clear, thanks for the clarification! Sorry for misunderstanding.

kambala-decapitator:
ready for review

.github/workflows/github.yml
```diff
@@ -23,12 +23,22 @@ jobs:
             os: ubuntu-20.04
             test: 0
             preset: linux-gcc-release
-          - platform: mac
-            os: macos-10.15
+          - platform: mac-intel
+            os: macos-12
```
Nordsoft91, line 27: Will resulting builds compatible with catalina?

kambala-decapitator: Sure! They're compatible down to 10.13. You can already grab artifact from the github actions run.

kambala-decapitator:
@ArseniyShestakov do you have any remarks?

Nordsoft91:
Mac build size is a half of what we had before. Do you have understanding why is that?

kambala-decapitator:
@Nordsoft91 the reasons are:

- libraries are configured with various options to strip features we don't need, see `default_options` in `conanfile.py`
- libs that don't need to be shared with other executables/dylibs (client-only mostly - ffmpeg, SDL) are built as static, that's why you don't see them in the app bundle

kambala-decapitator:
important point: someone should take care to produce proper file names for https://download.vcmi.eu/branch/develop/macOS/

kambala-decapitator:
> important point: someone should take care to produce proper file names for https://download.vcmi.eu/branch/develop/macOS/

adapted server script to put files in subdirectory based on the new filename suffix, suffix is added to filename based on `artifact_platform` matrix parameter.

rilian-la-te:
@nullkiller @kambala-decapitator will merge my Conan Windows PR with this changes soon.

tomkidd:
The ARM64 builds aren't running on my M1 Max system, though the latest Intel builds are. When I try and run it, it just reports as damaged and that I should trash it. 

Also with this PR merged is it possible to still use CMake to build/package it like before?

nullkiller:
Yes. Conan also uses CMake. Also it should be still possible to use homebrew AFAIK but not for ARM probably.

kambala-decapitator:
@tomkidd FYI I can't reproduce using the following way:

1. download ARM dmg from a GitHub action with Safari to my Intel Mac, e.g. https://github.com/vcmi/vcmi/actions/runs/3006478000
2. transfer it to ARM Mac with `scp`
3. launch VCMI on the ARM Mac

kambala-decapitator:
oh, have you tried the most straightforward way? Right-click the app - Open - Open. That's a standard way to run unsigned apps.

tomkidd:
> oh, have you tried the most straightforward way? Right-click the app - Open - Open. That's a standard way to run unsigned apps.

Yeah, and it didn't work. This wasn't the message saying the app is unsigned/from an an unidentified developer, this was saying that the app was damaged. 

That said I tried the `xatttr -d` command above and it worked. 

I'm still figuring out all the quirks in making builds but it's as if there's a magic trick you have to perform on these things with regards to Ad Hoc signing to where they don't get quarantined, or something. For example the RVGL project (Re-Volt source port) does Ad Hoc signing on the ARM64 version of their build and we get the expected behavior: having to right-click and Open to run them. 

Anyways, I run Mac Source Ports and I'd love to get a signed/notarized build of this thing going, bonus if it's Universal 2. I almost had it working but the bundling of the Qt@5 frameworks was a hurdle, but I see yours uses dylibs, how did you build those? 

Thanks for the help.

rilian-la-te:
@tomkidd @kambala-decapitator you can use Qt@6 for VCMI Launcher.

kambala-decapitator:
> This wasn't the message saying the app is unsigned/from an an unidentified developer, this was saying that the app was damaged.

now tried to download directly to arm Mac - have the same issue, will check how to overcome that

> does Ad Hoc signing on the ARM64 version of their build and we get the expected behavior

maybe that's the solution really, thanks!

> I'd love to get a signed/notarized build of this thing going

this requires $100/year, not sure if Apple gives free developer accounts to OSS

> bonus if it's Universal

we decided to upload separate Intel and ARM builds to make builds faster (in parallel) and to avoid extra step to `lipo` everything together in the end. Besides, currently our ARM releases don't contain the `scripting` part because Conan's LuaJIT is quite old, having a universal build would require code change(s) to ensure that the scripting libs aren't attempted to load as currently app crashes if the lib exists but can't be loaded. Scripting doesn't really work at this point, so it's fine to exclude it.

> but I see yours uses dylibs, how did you build those?

basically just pass `-no-framework` to the Qt configure script. But I don't think it really matters if you use dylibs or frameworks - `macdeployqt` should take care of everything.

> you can use Qt@6 for VCMI Launcher

What would it solve? Qt 6 raises minimum deployment target (haven't tested if it can still be built for macOS 10.13 / iOS 10) which is especially bad for Windows as Windows 10 is a requirement for Qt 6. For local builds using Qt 6 is fine, of course.

PR #783, opened 2022-08-25
Check for updates

Nordsoft91:
I've implemented a simple solution to help people to be up-to-date with the latest builds.
I think It will be very helpful for the new release. So all people who get new release will be involved more => more help from community.
Yes, it's very primitive but from my standpoint even that is much better rather that without it. 

Key features:
 - checks updates automatically in silent mode (no dialog for no updates)
 - supports different links for different operation systems
 - people who want to participate in alpha/beta tests, need just to change parameter in the config file to "subscribe"
 - different release types: minor, major, critical
 - download link is actually just a hyperlink, which opens default web-browser, so it can point on wiki, site, file, etc.
 - update check relies only on version and it's easy to make rollbacks, for example
 - shows error information if there is problem with server

BTW, it's my first experience in QT development ever :)

Снимок экрана 2022-08-25 в 06 37 45

Снимок экрана 2022-08-25 в 06 36 57

One thing we need to discuss is where to store autoUpdate.json file.
My proposal is to move my test repository to vcmi space and make everybody be able to update it via pull requests.

Example json
Снимок экрана 2022-08-25 в 07 18 53

### DIFF
--- a/config/schemas/settings.json
+++ b/config/schemas/settings.json
@@ -388,7 +388,7 @@
 				},
 				"updateConfigUrl" : {
 					"type" : "string",
-					"default" : "https://raw.githubusercontent.com/Nordsoft91/vcmi-autoupdate/main/autoUpdate.json"
+					"default" : "https://raw.githubusercontent.com/vcmi/vcmi-updates/master/vcmi-updates.json"
 				}
 			}
 		}
--- a/config/schemas/settings.json
+++ b/config/schemas/settings.json
@@ -363,7 +363,7 @@
 			"type" : "object",
 			"default": {},
 			"additionalProperties" : false,
-			"required" : [ "repositoryURL", "enableInstalledMods", "autoCheckRepositories" ],
+			"required" : [ "repositoryURL", "enableInstalledMods", "autoCheckRepositories", "updateOnStartup" ],
 			"properties" : {
 				"repositoryURL" : {
 					"type" : "array",
@@ -379,6 +379,10 @@
 					"default" : true
 				},
 				"autoCheckRepositories" : {
+					"type" : "boolean",
+					"default" : true
+				},
+				"updateOnStartup" : {
 					"type" : "boolean",
 					"default" : true
 				}
--- a/config/schemas/settings.json
+++ b/config/schemas/settings.json
@@ -363,7 +363,7 @@
 			"type" : "object",
 			"default": {},
 			"additionalProperties" : false,
-			"required" : [ "repositoryURL", "enableInstalledMods", "autoCheckRepositories", "updateOnStartup" ],
+			"required" : [ "repositoryURL", "enableInstalledMods", "autoCheckRepositories", "updateOnStartup", "updateConfigUrl" ],
 			"properties" : {
 				"repositoryURL" : {
 					"type" : "array",
@@ -385,6 +385,10 @@
 				"updateOnStartup" : {
 					"type" : "boolean",
 					"default" : true
+				},
+				"updateConfigUrl" : {
+					"type" : "string",
+					"default" : "https://raw.githubusercontent.com/Nordsoft91/vcmi-autoupdate/main/autoUpdate.json"
 				}
 			}
 		}
--- a/launcher/updatedialog_moc.cpp
+++ b/launcher/updatedialog_moc.cpp
@@ -17,7 +17,7 @@
 #include <QNetworkReply>
 #include <QNetworkRequest>
 
-UpdateDialog::UpdateDialog(QWidget *parent, bool calledManually) :
+UpdateDialog::UpdateDialog(bool calledManually, QWidget *parent):
 	QDialog(parent),
 	ui(new Ui::UpdateDialog),
 	calledManually(calledManually)
@@ -79,7 +79,7 @@ UpdateDialog::~UpdateDialog()
 
 void UpdateDialog::showUpdateDialog(bool isManually)
 {
-	UpdateDialog * dialog = new UpdateDialog(nullptr, isManually);
+	UpdateDialog * dialog = new UpdateDialog(isManually);
 	
 	dialog->setAttribute(Qt::WA_DeleteOnClose);
 }
--- a/launcher/updatedialog_moc.cpp
+++ b/launcher/updatedialog_moc.cpp
@@ -19,9 +19,8 @@ UpdateDialog::UpdateDialog(QWidget *parent) :
 	if(settings["launcher"]["updateOnStartup"].Bool() == true)
 		ui->checkOnStartup->setCheckState(Qt::CheckState::Checked);
 	
-	currentVersion = QString::fromStdString(GameConstants::VCMI_VERSION);
-	ui->currentVersion->setText(currentVersion);
-	
+	currentVersion = GameConstants::VCMI_VERSION;
+	setWindowTitle(QString::fromStdString(currentVersion));
 	
 #ifdef VCMI_WINDOWS
 	platformParameter = "windows";
@@ -95,6 +94,11 @@ void UpdateDialog::loadFromJson(const JsonNode & node)
 		return;
 	}
 	
+	//check whether update is needed
+	std::string newVersion = node["version"].String();
+	if(currentVersion == newVersion)
+		return;
+	
 	if(node["updateType"].String() == "minor")
 		ui->versionLabel->setStyleSheet("QLabel { background-color : gray; color : black; }");
 	if(node["updateType"].String() == "major")
@@ -102,7 +106,7 @@ void UpdateDialog::loadFromJson(const JsonNode & node)
 	if(node["updateType"].String() == "critical")
 		ui->versionLabel->setStyleSheet("QLabel { background-color : red; color : black; }");
 	
-	ui->versionLabel->setText(QString::fromStdString(node["version"].String()));
+	ui->versionLabel->setText(QString::fromStdString(newVersion));
 	ui->plainTextEdit->setPlainText(QString::fromStdString(node["changeLog"].String()));
 	
 	QString downloadLink = QString::fromStdString(node["downloadLinks"]["other"].String());
--- a/launcher/updatedialog_moc.cpp
+++ b/launcher/updatedialog_moc.cpp
@@ -57,15 +57,9 @@ UpdateDialog::UpdateDialog(QWidget *parent, bool calledManually) :
 			return;
 		}
 		
-		//auto const contentType = response->header(QNetworkRequest::ContentTypeHeader).toString();
-		//htmlContent = contentType;
 		auto byteArray = response->readAll();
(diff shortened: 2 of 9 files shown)


nullkiller:
Android does not use standard launcher.

Alucard648:
It`s supposed to be used for desktop OS`s.

kambala-decapitator:
> Android does not use standard launcher.

but iOS does ;)

launcher/updatedialog_moc.cpp
```diff
@@ -0,0 +1,143 @@
...
+	
+	if(calledManually)
+	{
+		setWindowModality(Qt::ApplicationModal);
+		show();
+	}
+	
+	connect(ui->closeButton, SIGNAL(clicked()), this, SLOT(close()));
+	
+	if(settings["launcher"]["updateOnStartup"].Bool() == true)
```
kambala-decapitator, line 36: same as above

launcher/updatedialog_moc.cpp
```diff
@@ -0,0 +1,143 @@
...
+	platformParameter = "ios";
+#elif defined(VCMI_ANDROID)
+	platformParameter = "android";
+#elif defined(VCMI_UNIX)
+	platformParameter = "linux";
+#endif
+	
+	QString url = QString::fromStdString(settings["launcher"]["updateConfigUrl"].String());
+	
+	QScopedPointer<QNetworkAccessManager> manager(new QNetworkAccessManager);
```
kambala-decapitator, line 57: I wouldn't call it a standard pattern. Why not simply deleting the manager inside the response handler (which you already do)?

launcher/updatedialog_moc.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+UpdateDialog::~UpdateDialog()
+{
+	delete ui;
+}
+
+void UpdateDialog::showUpdateDialog(bool isManually)
+{
+	UpdateDialog * dialog = new UpdateDialog(nullptr, isManually);
```
kambala-decapitator, line 86: seems you don't like `auto`, but often it helps a lot :)

launcher/updatedialog_moc.cpp
```diff
@@ -0,0 +1,143 @@
...
+	
+	dialog->setAttribute(Qt::WA_DeleteOnClose);
+}
+
+void UpdateDialog::on_checkOnStartup_stateChanged(int state)
+{
+	Settings node = settings.write["launcher"]["updateOnStartup"];
+	node->Bool() = (state == 2 ? true : false);
```
kambala-decapitator, line 94:
```suggestion
	node->Bool() = state == 2;
```

launcher/updatedialog_moc.cpp
```diff
@@ -0,0 +1,143 @@
...
+		setWindowModality(Qt::ApplicationModal);
+		show();
+	}
+	
+	if(node["updateType"].String() == "minor")
+		ui->versionLabel->setStyleSheet("QLabel { background-color : gray; color : black; }");
+	if(node["updateType"].String() == "major")
+		ui->versionLabel->setStyleSheet("QLabel { background-color : orange; color : black; }");
+	if(node["updateType"].String() == "critical")
+		ui->versionLabel->setStyleSheet("QLabel { background-color : red; color : black; }");
```
kambala-decapitator, line 130:
```suggestion
	QLatin1String bgColor;
	const auto updateType = node["updateType"].String();
	if(updateType == "minor")
		bgColor = {"gray"};
	else if(updateType == "major")
		bgColor = {"orange"};
	else if(updateType == "critical")
		bgColor = {"red"};
	ui->versionLabel->setStyleSheet(QLatin1String("QLabel { background-color : %1; color : black; }").arg(bgColor));
```

can there be other `updateType`s?

launcher/updatedialog_moc.cpp
```diff
@@ -0,0 +1,143 @@
...
+	ui->plainTextEdit->setPlainText(QString::fromStdString(node["changeLog"].String()));
+	
+	QString downloadLink = QString::fromStdString(node["downloadLinks"]["other"].String());
+	if(node["downloadLinks"][platformParameter].getType() == JsonNode::JsonType::DATA_STRING)
+		downloadLink = QString::fromStdString(node["downloadLinks"][platformParameter].String());
+	
+	QString downloadHtml("<a href=\"");
+	downloadHtml += downloadLink + "\">link</a>";
+	
+	ui->downloadLink->setText(downloadHtml);
```
kambala-decapitator, line 142:
```suggestion
	ui->downloadLink->setText(QString{"<a href=\"%1""\">link</a>"}.arg(downloadLink));
```

launcher/updatedialog_moc.cpp
```diff
@@ -0,0 +1,143 @@
...
+		if(response->error() != QNetworkReply::NoError)
+		{
+			ui->versionLabel->setStyleSheet("QLabel { background-color : red; color : black; }");
+			ui->versionLabel->setText("Network error");
+			ui->plainTextEdit->setPlainText(response->errorString());
+			return;
+		}
+		
+		auto byteArray = response->readAll();
+		JsonNode node(byteArray.constData(), byteArray.size());
```
kambala-decapitator, line 73: as the final goal is to throw JSON data into Qt objects, it might be better to use `QJsonDocument` to deal with JSON

launcher/updatedialog_moc.h
```diff
@@ -0,0 +1,41 @@
...
+namespace Ui {
+class UpdateDialog;
+}
+
+class UpdateDialog : public QDialog
+{
+	Q_OBJECT
+
+public:
+	explicit UpdateDialog(QWidget *parent = nullptr, bool calledManually = false);
```
kambala-decapitator, line 24: `parent` param should be last by Qt convention

kambala-decapitator (changes requested):

launcher/updatedialog_moc.cpp
Nordsoft91: For me it worse in terms of readability

launcher/updatedialog_moc.h
Nordsoft91: what is QT convention?

launcher/updatedialog_moc.cpp
Nordsoft91: We use JsonNode in other places of vcmilauncher so It make sense to keep one mechanism rather t

launcher/updatedialog_moc.cpp
Nordsoft91: good point, actually it will be green by default

launcher/updatedialog_moc.cpp
kambala-decapitator: "" after %1 isn't needed, made a typo

launcher/updatedialog_moc.cpp
kambala-decapitator: Would it work if the comparison is placed in brackets?

config/schemas/settings.json
```diff
@@ -381,6 +381,14 @@
...
 					"type" : "boolean",
 					"default" : true
+				},
+				"updateOnStartup" : {
+					"type" : "boolean",
+					"default" : true
+				},
+				"updateConfigUrl" : {
+					"type" : "string",
+					"default" : "https://raw.githubusercontent.com/Nordsoft91/vcmi-autoupdate/main/autoUpdate.json"
```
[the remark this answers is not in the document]
nullkiller: At list it is possible to fetch the list of updates from downloads.vcmi.eu But the idea of this is to allow downgrade in case of instability of latest release.

launcher/updatedialog_moc.cpp
kambala-decapitator: I'd say that the simplest and safest way would be making the network manager a private field and not a pointer. Or even simply:

```c++
auto manager = new QNetworkAccessManager{this};
```

and remove `response->manager()->deleteLater();`

launcher/updatedialog_moc.cpp
Nordsoft91: It's part of the future work

launcher/updatedialog_moc.cpp
Nordsoft91: it's a matter of taste, there is no improvement

launcher/updatedialog_moc.cpp
Nordsoft91: btw code is not compilable

kambala-decapitator: > btw code is not compilable

and the error is?..

sorry, I typed it on GitHub directly.

launcher/updatedialog_moc.cpp
```diff
@@ -53,14 +52,12 @@ UpdateDialog::UpdateDialog(QWidget *parent, bool calledManually) :
...
 	
 	QString url = QString::fromStdString(settings["launcher"]["updateConfigUrl"].String());
-	
-	QScopedPointer<QNetworkAccessManager> manager(new QNetworkAccessManager);
-	
-	QNetworkReply *response = manager->get(QNetworkRequest(QUrl(url)));
+		
+	QNetworkReply *response = networkManager.get(QNetworkRequest(QUrl(url)));
 	
 	QObject::connect(response, &QNetworkReply::finished, [&, response]{
```
kambala-decapitator, line 58:
```suggestion
	connect(response, &QNetworkReply::finished, [&, response]{
```

FYI using `QObject::connect` is needed only when you're not inside a QObject-derived class (like `main` function)

launcher/updatedialog_moc.cpp
nullkiller: Btw might be worth to introduce at least a local constant for 2

Nordsoft91: Probably there should be in QT, I’ll check

config/schemas/settings.json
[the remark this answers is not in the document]
Nordsoft91: GitHub api is not an option - it has very limited restrictions for usage without authentication

launcher/updatedialog_moc.cpp
```diff
@@ -103,8 +103,15 @@ void UpdateDialog::loadFromJson(const JsonNode & node)
 	}
 	
 	//check whether update is needed
+	bool isFutureVersion = true;
 	std::string newVersion = node["version"].String();
-	if(currentVersion == newVersion)
+	for(auto & prevVersion : node["history"].Vector())
+	{
+		if(prevVersion.String() == currentVersion)
+			isFutureVersion = false;
```
kambala-decapitator, line 111:
```suggestion
		if(prevVersion.String() == currentVersion)
		{
  			isFutureVersion = false;
  			break;
  		}
```

Nordsoft91:
Everything from this PR is contained here:
https://github.com/vcmi/vcmi/pull/785

PR #784, opened 2022-08-26
Fix problems with water zone

Nordsoft91:
Issue to be fixed is related to some blocked zones on normal water mode.

### DIFF
--- a/lib/rmg/RmgArea.cpp
+++ b/lib/rmg/RmgArea.cpp
@@ -85,8 +85,15 @@ bool Area::connected() const
 	return connected.empty();
 }
 
-std::list<Area> connectedAreas(const Area & area)
+std::list<Area> connectedAreas(const Area & area, bool disableDiagonalConnections)
 {
+	auto allDirs = int3::getDirs();
+	std::vector<int3> dirs;
+	if(disableDiagonalConnections)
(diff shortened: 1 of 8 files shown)


lib/rmg/RmgArea.cpp
```diff
@@ -85,8 +85,15 @@ bool Area::connected() const
...
-std::list<Area> connectedAreas(const Area & area)
-{
+std::list<Area> connectedAreas(const Area & area, bool disableDiagonalConnections)
+{
+	auto allDirs = int3::getDirs();
+	std::vector<int3> dirs;
+	if(disableDiagonalConnections)
+		dirs.insert(dirs.begin(), allDirs.begin(), allDirs.begin() + 4);
+	else
+		dirs.insert(dirs.begin(), allDirs.begin(), allDirs.end());
```
nullkiller, line 95: Strange that we have no such method already. Would be nice to have one. Btw you already have
static const int3 dirs4[] = {int3(0,1,0),int3(0,-1,0),int3(-1,0,0),int3(+1,0,0)};

Nordsoft91: let't minimise amount of changes at late stages. adding such method affects very int3 class which is very low level I don't want to touch now

Issue #793, opened 2022-09-03
upgrade creature hotkey doesn't work


kambala-decapitator:
in the original game, if you open creature info in town and there's button to upgrade creature, you can invoke it with U button

PR #803, opened 2022-09-03
#793 U shortcut for creature stack stats window

nullkiller:
Adds U shortcut for creature stack window to trigger creature upgrade dialog

### DIFF
--- a/client/windows/CCreatureWindow.cpp
+++ b/client/windows/CCreatureWindow.cpp
@@ -324,6 +324,11 @@ CStackWindow::ButtonsSection::ButtonsSection(CStackWindow * owner, int yOffset)
 
 			upgradeBtn->addOverlay(std::make_shared<CAnimImage>("CPRSMALL", VLC->creh->objects[upgradeInfo.info.newID[buttonIndex]]->iconIndex));
 
+			if(buttonsToCreate == 1) // single upgrade avaialbe
+			{
+				upgradeBtn->assignedKeys.insert(SDLK_u);
+			}
+
 			upgrade[buttonIndex] = upgradeBtn;
 		}
 	}


client/windows/CCreatureWindow.cpp
```diff
@@ -324,6 +324,11 @@ CStackWindow::ButtonsSection::ButtonsSection(CStackWindow * owner, int yOffset)
 
 			upgradeBtn->addOverlay(std::make_shared<CAnimImage>("CPRSMALL", VLC->creh->objects[upgradeInfo.info.newID[buttonIndex]]->iconIndex));
 
+			if(buttonIndex == 0)
```
kambala-decapitator, line 327: any chance to get rid of the magic value?

nullkiller: What do you suggest? FIRST_BUTTON_INDEX = 0? LOWEST_UPGRADE_INDEX = 0?

kambala-decapitator: if you give names to all indexes (e.g. constants), then you can iterate through them as an initialiser list (for-in) instead of plain for. Or have an enum with last element named Count or so.

nullkiller: The goal of constant here is to give some hint on what the code actually does (it is onetime usage). Why 0 is there. Have you understood why? If yes - which constant reflects it the best? If it does not help understanding - it is useless here and plain number might be even better

nullkiller: Also there is some danger in this solution that both upgrades might be equivalent and you just can not automatically decide which one to hit when there is more than one. So we can do it another way and allow shortcut only if there is single upgrade available

kambala-decapitator: thanks, now it's clear that there can be multiple upgrades, and 0 simply means "first".

kambala-decapitator (approved):

PR #808, opened 2022-09-03
Mine roads

DjWarmonger:
If zone has only one entry and no towns, it will have no roads.

In such case RMG will create dirt roads to mines if possible.


Alucard648:
Is it possible to create branching roads to mines in regular zones, like player start zones?

DjWarmonger:
> Is it possible to create branching roads to mines in regular zones, like player start zones?

Possible - yes, but more complicted. Also I think it'd be an overkill to have roads all over the zone.

Alucard648:
> Also I think it'd be an overkill to have roads all over the zone.

It should be a template setting. For template editor.

Nordsoft91:
There should be easier solution - take a look at ObjectManager::placeObject. Maybe adding node there + filtering inside roadPlacer would be better

Alucard648:
> There should be easier solution - take a look at ObjectManager::placeObject. Maybe adding node there + filtering inside roadPlacer would be better

Agreed. This would allow template designer to include road connection with other important elements of generated map, such as dragon utopias.

PR #846, opened 2022-09-06
[844] - fix pathfinding for fly towards guarded tile

nullkiller:
I am not sure about the crash origin but observed bad behavior related to fly and angel wings. It is possible to pick removable objects (artifacts) standing on water tile. After this hero becomes unmovable and can escape only using adv spells. This behavior is fixed now according to HOMM3. It is possible to attack guards directly from air.

### DIFF
--- a/lib/CPathfinder.cpp
+++ b/lib/CPathfinder.cpp
@@ -620,10 +620,11 @@ void LayerTransitionRule::process(
 				destination.blocked = true;
 			}
 		}
-		else if(source.node->accessible != CGPathNode::ACCESSIBLE &&	destination.node->accessible != CGPathNode::ACCESSIBLE)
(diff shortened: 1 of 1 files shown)


Nordsoft91 (approved):

Nordsoft91 (approved):

nullkiller:
You did not observe its turn because this time hero approached the shield from the bottom as it should be

nullkiller:
Regarding crash - it is not related to this fix.

kambala-decapitator:
> You did not observe its turn because this time hero approached the shield from the bottom as it should be

I thought I should see it because I can see the hero and the shield partially. I might be wrong, but in the original I'd see how the hero goes into the darkness and then how shield disappears.

PR #850, opened 2022-09-07
Terrain performance

Nordsoft91:
Complexity of id() before (for 10 terrains, string comparisons, worst case):
1. Create std::set - 10 * ln(10)
2. Create std::vector - 10 * ln(10)
3. Search element in vector - 10

Complexity of id after ln(10)

Summary: speedup ~20 times

### DIFF
--- a/lib/Terrain.cpp
+++ b/lib/Terrain.cpp
@@ -154,6 +154,11 @@ Terrain::Manager::Manager()
 			}
 			
 			terrainInfo[terr.first] = info;
+			if(!terrainId.count(terr.first))
+			{
+				terrainId[terr.first] = terrainVault.size();
+				terrainVault.push_back(terr.first);
+			}
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


lib/Terrain.cpp
```diff
@@ -164,12 +169,18 @@ Terrain::Manager & Terrain::Manager::get()
...
+	return Terrain::Manager::get().terrainVault;
+}
+
+int Terrain::Manager::id(const Terrain & terrain)
+{
+	if(terrain.name == "ANY") return -3;
+	if(terrain.name == "WRONG") return -2;
+	if(terrain.name == "BORDER") return -1;
+	
+	return Terrain::Manager::get().terrainId.at(terrain);
```
nullkiller, line 183: can we put these special ids to terrainID map

Nordsoft91: Certainly, but what this is a bit odd for me, but searching in map is longer than comparison (nevertheless I don't know frequency of usage)

What will be benefit? If it's readability only, I can make separate const array or something at the top of file.

nullkiller: We will have one search only. The one at line 183.

nullkiller: So this will be a bit odd I agree.

nullkiller (approved):

Issue #852, opened 2022-09-07
Show submods of downloaded mod without restarting launcher


kambala-decapitator:
Can be tested e.g. on **VCMI extras**

Issue #851, opened 2022-09-07
Submods aren't activated together with parent mod


kambala-decapitator:
### When downloading a pre-defined mod
1. Download e.g. **VCMI extras**, it gets activated automatically
2. Restart launcher to see submods (to be fixed in #852)
3. Submods are disabled

![Screenshot 2022-09-07 at 12 42 58](https://user-images.githubusercontent.com/1557784/188853892-30d1409d-fa4e-4287-807c-0ed98f4bcb6e.png)

### When installing a mod manually
1. Copy mod directory to **Mods** directory
2. start Launcher, the new mod is disabled by default (correct)
3. Activate the mod, submods aren't activated

![Screenshot 2022-09-07 at 13 18 50](https://user-images.githubusercontent.com/1557784/188854582-aedf2f52-b030-43c9-9269-068fc274cf05.png)

Issue #812, opened 2022-09-04
Launcher shows errors during mod reinstallation


Nordsoft91:
Pre-conditions: launched needs to be able to download mods from remote server. Launcher from develop can do it, but there is no server with mods available. The branch https://github.com/Nordsoft91/vcmi/tree/mod-from-github allows to download mods from GitHub.

Steps to reproduce:
1. Open launcher
2. Press "download and refresh repositories"
3. Choose any mod with submods (e.g wow/vcmi extra/hota)
4. Install mod
5. Make sure all submods are enabled
6. uninstall mod by pressing uninstall
7. Install mod again

Expected result: no errors, mod is installed and enabled along with all submods
Actual result: error occurs
<img width="370" alt="Снимок экрана 2022-09-04 в 03 25 17" src="https://user-images.githubusercontent.com/9308612/188310530-37807280-9626-4341-abcd-75c49817df75.png">

Issue #845, opened 2022-09-06
Make essential files permanent


Nordsoft91:
There are two possible solutions:
 - place files from essential to filesystem folders (data/sprites/etc) and remove mod
 - keep as a mod but make impossible to disable

nullkiller:
I think better to make it directly a part of filesystem, a part of core mod or at least a separate mod but built in runtime the same way as core. So it should not be visible in launcher.

PR #860, opened 2022-09-08
Launcher fixes for mod filesystem

Nordsoft91:
Fixes for
https://github.com/vcmi/vcmi/issues/812
https://github.com/vcmi/vcmi/issues/851
https://github.com/vcmi/vcmi/issues/852

Actually, mod system works as natively expected

### DIFF
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -229,7 +229,7 @@ CModEntry CModList::getMod(QString modname) const
 		if(path != rootPath)
 		{
 			conf = getValue(modSettings, rootPath);
-			auto confMap = conf.toMap();
+			const auto confMap = conf.toMap();
 			if(!conf.isNull() && !confMap["active"].toBool())
 			{
 				settings = confMap;
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -152,6 +152,10 @@ QVariantMap CModList::copyField(QVariantMap data, QString from, QString to)
 	return renamed;
 }
 
+void CModList::reloadRepositories()
+{
+}
+
 void CModList::resetRepositories()
 {
 	repositories.clear();
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -218,7 +218,11 @@ CModEntry CModList::getMod(QString modname) const
 	else
 	{
 		if(!conf.toMap().isEmpty())
+		{
 			settings = conf.toMap();
+			if(settings.value("active").isNull())
+				settings["active"] = true; // default
+		}
 		else
 			settings.insert("active", conf);
 	}
@@ -230,7 +234,7 @@ CModEntry CModList::getMod(QString modname) const
 		{
 			conf = getValue(modSettings, rootPath);
 			const auto confMap = conf.toMap();
-			if(!conf.isNull() && !confMap["active"].toBool())
+			if(!conf.isNull() && !confMap["active"].isNull() && !confMap["active"].toBool())
 			{
 				settings = confMap;
 			}
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -176,7 +176,7 @@ void CModList::modChanged(QString modID)
 {
 }
 
-static QVariant getValue(QVariantMap input, QString path)
+static QVariant getValue(QVariant input, QString path)
 {
 	if(path.size() > 1)
 	{
@@ -184,7 +184,7 @@ static QVariant getValue(QVariantMap input, QString path)
 		QString remainder = "/" + path.section('/', 2, -1);
 
 		entryName.remove(0, 1);
-		return getValue(input.value(entryName).toMap(), remainder);
+		return getValue(input.toMap().value(entryName), remainder);
 	}
 	else
 	{
@@ -201,14 +201,14 @@ CModEntry CModList::getMod(QString modname) const
 	QString path = modname;
 	path = "/" + path.replace(".", "/mods/");
 	QVariant conf = getValue(modSettings, path);
-
+	
 	if(conf.isNull())
 	{
 		settings["active"] = true; // default
 	}
 	else
 	{
-		if(conf.canConvert<QVariantMap>())
+		if(!conf.toMap().isEmpty())
 			settings = conf.toMap();
 		else
 			settings.insert("active", conf);
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -205,7 +205,7 @@ CModEntry CModList::getMod(QString modname) const
 	QString path = modname;
 	path = "/" + path.replace(".", "/mods/");
 	QVariant conf = getValue(modSettings, path);
-	
+
 	if(conf.isNull())
 	{
 		settings["active"] = true; // default
@@ -224,9 +224,10 @@ CModEntry CModList::getMod(QString modname) const
 		if(path != rootPath)
 		{
 			conf = getValue(modSettings, rootPath);
-			if(!conf.isNull() && !conf.toMap()["active"].toBool())
+			auto confMap = conf.toMap();
+			if(!conf.isNull() && !confMap["active"].toBool())
 			{
-				settings = conf.toMap();
+				settings = confMap;
 			}
 		}
 	}
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -92,6 +92,11 @@ bool CModEntry::isUpdateable() const
 	return false;
 }
 
+bool CModEntry::isEssential() const
+{
+	return getValue("storedLocaly").toBool();
+}
+
 bool CModEntry::isInstalled() const
 {
 	return !localData.isEmpty();
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -213,6 +213,19 @@ CModEntry CModList::getMod(QString modname) const
 		else
 			settings.insert("active", conf);
 	}
+	
+	if(settings["active"].toBool())
+	{
+		QString rootPath = path.section('/', 0, 1);
+		if(path != rootPath)
+		{
+			conf = getValue(modSettings, rootPath);
+			if(!conf.isNull() && !conf.toMap()["active"].toBool())
+			{
+				settings = conf.toMap();
+			}
+		}
+	}
 
 	for(auto entry : repositories)
 	{
--- a/launcher/modManager/cmodlist.h
+++ b/launcher/modManager/cmodlist.h
@@ -49,6 +49,8 @@ public:
 	bool isUpdateable() const;
 	// installed
 	bool isInstalled() const;
+	// vcmi essential filea
+	bool isEssential() const;
 
 	// see ModStatus enum
 	int getModStatus() const;
--- a/launcher/modManager/cmodlist.h
+++ b/launcher/modManager/cmodlist.h
@@ -74,6 +74,7 @@ class CModList
 
 public:
 	virtual void resetRepositories();
+	virtual void reloadRepositories();
 	virtual void addRepository(QVariantMap data);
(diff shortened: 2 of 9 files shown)


Nordsoft91:
Thanks @DjWarmonger , I borrowed some points from his PR

lib/filesystem/CFilesystemLoader.cpp
```diff
@@ -52,7 +52,7 @@ void CFilesystemLoader::updateFilteredFiles(std::function<bool(const std::string
 {
 	if (filter(mountPoint))
 	{
-		fileList = listFiles(mountPoint, 1, false);
+		fileList = listFiles(mountPoint, 64, false);
```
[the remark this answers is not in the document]
nullkiller: Might be worth to get rid of magic numbers if possible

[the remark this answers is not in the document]
Nordsoft91: Number has been taken from Filesystem.cpp, line 152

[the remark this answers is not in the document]
Nordsoft91: Ok then, looks like have a better way to solve it

launcher/modManager/cmodlist.cpp
```diff
@@ -176,15 +176,15 @@ void CModList::modChanged(QString modID)
 {
 }
 
-static QVariant getValue(QVariantMap input, QString path)
+static QVariant getValue(QVariant input, QString path)
```
kambala-decapitator, line 183: most Qt types are "heavy", prefer passing them by const ref

launcher/modManager/cmodlist.cpp
```diff
@@ -213,6 +213,19 @@ CModEntry CModList::getMod(QString modname) const
...
+	
+	if(settings["active"].toBool())
+	{
+		QString rootPath = path.section('/', 0, 1);
+		if(path != rootPath)
+		{
+			conf = getValue(modSettings, rootPath);
+			if(!conf.isNull() && !conf.toMap()["active"].toBool())
+			{
+				settings = conf.toMap();
```
kambala-decapitator, line 229: please extract `conf.toMap()` into a variable

launcher/modManager/cmodlist.cpp
```diff
@@ -201,18 +205,31 @@ CModEntry CModList::getMod(QString modname) const
 	QString path = modname;
 	path = "/" + path.replace(".", "/mods/");
 	QVariant conf = getValue(modSettings, path);
-
+
```
kambala-decapitator, line 208: please drop

launcher/modManager/cmodmanager.cpp
```diff
@@ -229,7 +230,7 @@ bool CModManager::doEnableMod(QString mod, bool on)
 {
 	QString path = mod;
 	path = "/activeMods/" + path.replace(".", "/mods/") + "/active";
-
+
```
kambala-decapitator, line 233: please drop

launcher/modManager/cmodmanager.cpp
```diff
@@ -288,17 +287,14 @@ bool CModManager::doUninstallMod(QString modname)
...
 		return addError(modname, "Mod is located in protected directory, plase remove it manually:\n" + modFullDir.absolutePath());
 
-	localMods.remove(modname);
-	modList->setLocalModList(localMods);
-	modList->modChanged(modname);
-
+	CResourceHandler::get("initial")->updateFilteredFiles([](const std::string &){ return true; });
+	loadMods();
+	modList->reloadRepositories();
+
```
kambala-decapitator, line 297: please drop

kambala-decapitator (changes requested):

launcher/modManager/cmodlist.cpp
Nordsoft91: Thus I have to refactor across whole launcher

launcher/modManager/cmodlist.cpp
Nordsoft91: why is purpose?

kambala-decapitator: you're doing it twice, might not be cheap if map is big

launcher/modManager/cmodlist.cpp
kambala-decapitator: yeah, let's leave for another PR :)

launcher/modManager/cmodlist.cpp
Nordsoft91: oh, it's returns value, not reference, ok then

launcher/modManager/cmodlist.cpp
```diff
@@ -208,11 +212,25 @@ CModEntry CModList::getMod(QString modname) const
...
 	}
+	
+	if(settings["active"].toBool())
+	{
+		QString rootPath = path.section('/', 0, 1);
+		if(path != rootPath)
+		{
+			conf = getValue(modSettings, rootPath);
+			auto confMap = conf.toMap();
+			if(!conf.isNull() && !confMap["active"].toBool())
```
kambala-decapitator, line 228: First you convert node to map, then you check if the node type is not null - the logic looks incorrect

Nordsoft91:
Added fox for https://github.com/vcmi/vcmi/issues/845

launcher/modManager/cmodlist.cpp
Nordsoft91: confMap will be empty if cons is null

kambala-decapitator: after double-check I agree that this is ok.

But I've just noticed another issue on this line: `operator[]` [inserts value if it doesn't exist](https://doc.qt.io/qt-5/qmap.html#operator-5b-5d). I guess such behavior is not intended/needed and you should use `QMap::value()` method instead. Alternatively, declare `confMap` as `const`, then you can safely use `operator[]` without side effects.

kambala-decapitator (commented):
tested the last commit about Essentials: works fine.

However, "disabling parent disables children" (as well as enabling) doesn't seem fixed: at least children icons don't change.

Nordsoft91:
@kambala-decapitator , remove settings.json from your data folder/config. It will simulate zero-day installation. That scenario must work. However, if you disable some submods, then reinstall the whole mod, it will preserve your previous settings.
I did this intentionally because I don't want hota fonts to be enabled each time when I switch hota mod for instance

launcher/modManager/cmodlist.cpp
Nordsoft91: ok, I'm a have a god-feeling that those changes will break this very fragile fix :)

kambala-decapitator:
ok, here's video of "zero-day". Something's definitely wrong :)

https://user-images.githubusercontent.com/1557784/189421500-2615a834-b72a-42f4-8500-30a3f056ae7b.mov

Nordsoft91:
That was my last commit to launcher

launcher/modManager/cmodlist.cpp
```diff
@@ -208,11 +217,29 @@ CModEntry CModList::getMod(QString modname) const
...
 	{
-		if(conf.canConvert<QVariantMap>())
+		if(!conf.toMap().isEmpty())
+		{
 			settings = conf.toMap();
+			if(settings.value("active").isNull())
+				settings["active"] = true; // default
+		}
 		else
 			settings.insert("active", conf);
```
kambala-decapitator, line 227: not directly related to your PR, but this line is suspicious. `settings["active"]` is treated everywhere as a boolean, but here whatever is in `conf` is inserted. Are there guarantees that it can only be a boolean? I'd rather change `else` to `else if(conf.canConvert<bool>())`, and in a new `else` branch at least print some error message.

launcher/modManager/cmodlist.cpp
```diff
@@ -208,11 +217,29 @@ CModEntry CModList::getMod(QString modname) const
...
 	}
+	
+	if(settings["active"].toBool())
+	{
+		QString rootPath = path.section('/', 0, 1);
+		if(path != rootPath)
+		{
+			conf = getValue(modSettings, rootPath);
+			const auto confMap = conf.toMap();
+			if(!conf.isNull() && !confMap["active"].isNull() && !confMap["active"].toBool())
```
kambala-decapitator, line 237: `confMap["active"]` would return a null QVariant from an empty map, we can drop one check:

```suggestion
			if(!confMap["active"].isNull() && !confMap["active"].toBool())
```

kambala-decapitator:
regarding the video above: I think I start to understand the new behavior. But I can't say that it's a great UX.

- parent-child enabling works only on the top-level mod (based on the icons), but why not in submods?
- if the top-level mod is disabled and user enables one of the submods, I'd expect that the submod gets enabled. Instead, user must enable top-level mod first (which enables all submods)

There's some strange behavior when disabling _Andruids Bonus Icons_: selection disappears. But with _Default templates_ there's no such issue.

https://user-images.githubusercontent.com/1557784/189474577-c6d020d0-dd16-42f6-b87b-ccf7a6721f1b.mov

Nordsoft91:
I don't want to work on this anymore. I spent 3 hours yesterday to figure out what's going on, I can't understand the logic beyond.
So we can proceed as is or someone can continue working on it.

kambala-decapitator (approved):
> I don't want to work on this anymore. I spent 3 hours yesterday to figure out what's going on, I can't understand the logic beyond.
> So we can proceed as is or someone can continue working on it.

sounds like the whole logic needs to be revised in the future

anyway, current state is a great improvement to the existing functionality, I'd use it

nullkiller:
There should be possibility to disable submods. do we have it now or is it broken?

Nordsoft91:
@nullkiller , there is no regression. Everything worked before works now (at least on the level of my testing)

launcher/modManager/cmodlist.cpp
```diff
@@ -92,6 +92,11 @@ bool CModEntry::isUpdateable() const
 	return false;
 }
 
+bool CModEntry::isEssential() const
```
nullkiller, line 95: Could we unify naming? So it is either essential or stored locally?

launcher/modManager/cmodlist.h
```diff
@@ -49,6 +49,8 @@ class CModEntry
 	bool isUpdateable() const;
 	// installed
 	bool isInstalled() const;
+	// vcmi essential filea
```
nullkiller, line 52: useless comment. Lets remove

launcher/modManager/cmodmanager.cpp
```diff
@@ -82,6 +83,13 @@ void CModManager::loadMods()
 		{
 			boost::filesystem::path name = *CResourceHandler::get()->getResourceName(resID);
 			auto mod = JsonUtils::JsonFromFile(pathToQString(name));
+			if(!name.is_absolute())
```
nullkiller, line 86: Some people might decide to store mods locally as well. Lets introduce a parameter to mod json instead

launcher/modManager/cmodlist.cpp
Nordsoft91: The intention is to check mod is it essential or not. I've considered 3 options:
* by name - the problem is that you may have one more mod in mods folder which you probably'd like to disable/delete if we update it from main repo
* by parameter - the problem here is that someone may include this parameter into other mods (intentionally or accidentally) so it'll look as a bug "cannot disable mod"
* by path - that looks like safest way from my pov. Anyway it's prohibited to delete/install mods locally from launcher. If someone store mods there manually, those should probably know what they are doing

launcher/modManager/cmodlist.h
Nordsoft91: comments for method above are about same:) so I just keep existing "style" :)

launcher/modManager/cmodmanager.cpp
Nordsoft91: Sorry, I'm not going to continue work. There is task, @kambala-decapitator created, all further improvements can be described there if somebody decides to work on it (see previous comments)

launcher/modManager/cmodlist.h
kambala-decapitator:
```suggestion
	// vcmi essential files
```

PR #868, opened 2022-09-09
Nullkiller ai tweaks

### LINKED ISSUE #3227: The abilities on an unit screen do not scroll on the touch device
Also scroll buttons doesn't working

Screenshot_20231126_130804

OS: [Android]
 - Version: 1.3.2

DjWarmonger:
Many AI tweaks:

- AI will now pick Pandoras as they are valuable objects
- AI will pick Prisons. At the same time I fixed #3227 https://bugs.vcmi.eu/view.php?id=3227
- Fixed exchange dialogs which in particular stopped AI from joining armies from Banks, Pandoras and Diplomacy when it already had 7 stacks in army.
- AI will attack anemy heroes agressively. This in fact makes it able to win in finite time.
- Various value tweak here and there

### DIFF
--- a/AI/Nullkiller/Engine/PriorityEvaluator.cpp
+++ b/AI/Nullkiller/Engine/PriorityEvaluator.cpp
@@ -255,6 +255,8 @@ uint64_t RewardEvaluator::getArmyReward(
 		return ai->cb->getPlayerRelations(target->tempOwner, ai->playerID) == PlayerRelations::ENEMIES
 			? enemyArmyEliminationRewardRatio * dynamic_cast<const CGHeroInstance *>(target)->getArmyStrength()
 			: 0;
+	case Obj::PANDORAS_BOX:
+		return 5000;
 	default:
 		return 0;
 	}
@@ -421,6 +423,9 @@ float RewardEvaluator::getSkillReward(const CGObjectInstance * target, const CGH
 		return 8;
 	case Obj::WITCH_HUT:
 		return evaluateWitchHutSkillScore(dynamic_cast<const CGWitchHut *>(target), hero, role);
+	case Obj::PANDORAS_BOX:
+		//Can contains experience, spells, or skills (only on custom maps)
+		return 2.5f;
 	case Obj::HERO:
 		return ai->cb->getPlayerRelations(target->tempOwner, ai->playerID) == PlayerRelations::ENEMIES
 			? enemyHeroEliminationSkillRewardRatio * dynamic_cast<const CGHeroInstance *>(target)->level
@@ -500,6 +505,8 @@ int32_t RewardEvaluator::getGoldReward(const CGObjectInstance * target, const CG
 		return 10000;
 	case Obj::SEA_CHEST:
 		return 1500;
+	case Obj::PANDORAS_BOX:
+		return 5000;
 	case Obj::HERO:
 		return ai->cb->getPlayerRelations(target->tempOwner, ai->playerID) == PlayerRelations::ENEMIES
 			? heroEliminationBonus + enemyArmyEliminationGoldRewardRatio * getArmyCost(dynamic_cast<const CGHeroInstance *>(target))
--- a/AI/Nullkiller/Engine/PriorityEvaluator.cpp
+++ b/AI/Nullkiller/Engine/PriorityEvaluator.cpp
@@ -196,7 +196,6 @@ uint64_t evaluateArtifactArmyValue(CArtifactInstance * art)
 		+ 700 * art->getDefense(false)
 		+ 700 * art->valOfBonuses(Bonus::PRIMARY_SKILL, PrimarySkill::KNOWLEDGE)
 		+ 700 * art->valOfBonuses(Bonus::PRIMARY_SKILL, PrimarySkill::SPELL_POWER)
-		+ 700 * art->getDefense(false)
 		+ 500 * art->valOfBonuses(Bonus::LUCK);
 
 	auto classValue = 0;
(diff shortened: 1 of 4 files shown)


AI/Nullkiller/Engine/PriorityEvaluator.cpp
```diff
@@ -273,7 +299,8 @@ float RewardEvaluator::getEnemyHeroStrategicalValue(const CGHeroInstance * enemy
 		vstd::amax(objectValue, getStrategicalValue(obj));
 	}
 
-	return objectValue / 2.0f + enemy->level / 15.0f;
+	//AI should absolutely prioritize killing enemy heroes, even scouts
+	return std::min(1.0f, objectValue * 0.9f + (1.0f - (1.0f / (1 + enemy->level))));
```
nullkiller, line 303: I think 1.0 it is too high. Scout should not be higher than capturing town with fort in current turn. And better add some basic value than use min(1, value). so that 2 lvl hero is more valuable than 1 lvl. Also we can increase this radius getOneTurnAccessibleObjects.
Just a case which I do not want to get broken. When enemy hero approaches a town you buy and put scouts not far from it and hero attacks them loosing movement points. While you bring your main army to protect the town. This is how original AI did afaik.
Mine idea here is simple. AI captures town and all scouts around become a treat for it. Thus they gain additional strategical value equivalent to this town and AI should eliminate them as well. Probably we can express hero value as a separate fuzzy variable to be able to write more flexible rules. If main hero will start pursuing a scout and the last will constantly run away we might end up having main hero just wasting movement points which is not nice as well. Having a separate variable will allow us writing rules like 
if enemy_hero is weak and turn is now then value is bit_high (or high)
if enemy_hero is strong and turn is now then value is highest

server/CGameHandler.cpp
```diff
@@ -5572,13 +5572,21 @@ bool CGameHandler::isAllowedExchange(ObjectInstanceID id1, ObjectInstanceID id2)
...
+		if (dialog)
+		{
+			auto topArmy = dialog->exchangingArmies.at(0);
+			auto bottomArmy = dialog->exchangingArmies.at(1);
+
+			if (topArmy == o1 && bottomArmy == o2)
 				return true;
 
-			if (dialog->exchangingArmies.at(1) == o1 && dialog->exchangingArmies.at(0) == o2)
+			if (bottomArmy == o1 && topArmy == o2)
```
Nordsoft91, line 5589: maybe combining those conditions is better
(top==1 && bot==2) || (top==2 && bot==1)

AI/Nullkiller/Engine/PriorityEvaluator.cpp
Nordsoft91, line 303: could you please add a comment about this formula?

Nordsoft91 (approved):

AI/Nullkiller/Engine/PriorityEvaluator.cpp
DjWarmonger: Well, I tested this in game AI vs AI and it works. AI finally kills the scouts in lategame instead of playing ping-pong, capturing a town and losing it next turn for weeks. Killing 4 scouts in a a turn is a winning strategy and running past them is not.

DjWarmonger: ALso, I think enemy shouldn't know hero level in first place :P

AI/Nullkiller/Engine/PriorityEvaluator.cpp
```diff
@@ -114,10 +116,32 @@ uint64_t getCreatureBankArmyReward(const CGObjectInstance * target, const CGHero
...
-	for(auto c : creatures)
+	const auto& slots = hero->Slots();
+	ui64 weakestStackPower = 0;
+	if (slots.size() >= GameConstants::ARMY_SIZE)
+	{
+		//No free slot, we might discard our weakest stack
+		weakestStackPower = std::numeric_limits<ui64>().max();
+		for (const auto stack : slots)
+		{
+			vstd::amin(weakestStackPower, stack.second->getPower());
```
Nordsoft91, line 127: stack.second could be a nullptr and game will crash

AI/Nullkiller/Engine/PriorityEvaluator.cpp
nullkiller: AI knows a bit more about you than human unfortunately. It knows exact army structure and your stats. And this is true for both  VCAI and Nullkiller afaik. But I agree that it should not. So this more a problem of client/server architecture that this info is available to AI. If AI knows - it is possible to build modified client which will also show all these stats.
Ok, in current circumstances this might be a better solution. But what I actually wanted was "What you say would make more sense if AI actually tried to buy units in a town to defend it". Still I think 1 is to high. Lets try 0.7f + instead of std:min(1, and see if it gives what you want.

AI/Nullkiller/Engine/PriorityEvaluator.cpp
DjWarmonger: Not really. Stack should always have defined creature type - if it doesn't and AI has them in their army, then something is already **very** wrong.

AI/Nullkiller/Engine/PriorityEvaluator.cpp
DjWarmonger: But it means that 1.0f is the upper cap and the value can still be lower for lesser objects.

nullkiller: I misunderstood it, sorry. Yes this way it looks ok.

nullkiller (approved):

PR #873, opened 2022-09-10
fix launching server when Windows user name has non-latin characters

### LINKED ISSUE #865: Game can't be launched if filepath contains non-unicode letters.
It was reported many times, that VCMI doesn't launch if user's comupter has non-unicode letters or folder name contains some of them.
(https://forum.vcmi.eu/t/help-me-plss-thx/5725)

kambala-decapitator:
Boost stores paths in UTF-8 internally, but calling `std::system()` executes command using system code page which is usually OEM866 (iirc). Converting UTF-8 to system code page is a PITA, so I convert it to wide string instead.

fixes #865

### DIFF
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -47,6 +47,10 @@
 
 #include <vcmi/events/EventBus.h>
 
+#ifdef VCMI_WINDOWS
+#include <windows.h>
+#endif
+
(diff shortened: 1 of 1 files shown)


Nordsoft91:
I have no opportunity to test it on windows

nullkiller (approved):

PR #886, opened 2022-09-12
Mods versioning [part 3]

Nordsoft91:
Those changes were developed in parallel with develop, so I had to resolve ~100 conflicts to merge everything.
Basically, change have 3 big topics:
 - VLC for obstacles
 - Mod versioning solution (without Launcher part) - that's required for VLC for obstacles as it breaks not only backward compatibility for saves, but also all mods with new obstacles become incompatible
 - Some features developed by me and @DjWarmonger for further support of editor, in particular
 > progress bar to reflect loading progress in parallel thread (can be used to show loading bar in game during rmg computations)
> undo/redo implementation for map
> filesystem missing functions

Test scenarios to check mod versions
1. run game regular -> save game -> load game
2. run game with mod A -> save game -> load game -> load game from p1
3. run game with mod B -> save game -> load game -> load game from p1, p2
4. run game with next version of A, B -> save game -> load game from p2, p3 (error expected)
5. run game as in p.3 -> load game from p4 (error expected)

PR #888, opened 2022-09-13
Vlc obstacles [part 1]

Nordsoft91:
First part of PR https://github.com/vcmi/vcmi/pull/886 - contains VLC only


lib/battle/BattleInfo.cpp
```diff
@@ -253,11 +254,11 @@ BattleInfo * BattleInfo::setupBattle(const int3 & tile, const Terrain & terrain,
 
 		auto appropriateAbsoluteObstacle = [&](int id)
 		{
-			return VLC->heroh->absoluteObstacles[id].isAppropriate(curB->terrainType, battlefieldType);
+			return VLC->obstacleHandler->absoluteObstacles[id].getInfo()->isAppropriate(curB->terrainType, battlefieldType);
 		};
 		auto appropriateUsualObstacle = [&](int id) -> bool
 		{
-			return VLC->heroh->obstacles[id].isAppropriate(curB->terrainType, battlefieldType);
+			return VLC->obstacleHandler->obstacles[id].getInfo()->isAppropriate(curB->terrainType, battlefieldType);
```
nullkiller, line 261: Will ((Obstacle)id).getInfo() work?

lib/ObstacleHandler.cpp
```diff
@@ -0,0 +1,115 @@
...
+	info->width = json["width"].Integer();
+	info->height = json["height"].Integer();
+	for(auto & t : json["allowedTerrain"].Vector())
+		info->allowedTerrains.emplace_back(t.String());
+	for(auto & t : json["specialBattlefields"].Vector())
+		info->allowedSpecialBfields.emplace_back(t.String());
+	info->blockedTiles = json["blockedTiles"].convertTo<std::vector<si16>>();
+	info->isAbsoluteObstacle = json["absolute"].Bool();
+	if(info->isAbsoluteObstacle)
+		absoluteObstacles.push_back(info->getId());
```
nullkiller, line 93: This might be dangerous. As far as I understood info->getId().getInfo() should always return the same instance once added to obstacles collection but we add to absoluteObstacles instead and Obstacle::getInfo will found nothing. Or am I missing something?

lib/battle/BattleInfo.cpp
```diff
@@ -274,7 +275,7 @@ BattleInfo * BattleInfo::setupBattle(const int3 & tile, const Terrain & terrain,
 
 				for(BattleHex blocked : obstPtr->getBlockedTiles())
```
nullkiller, line 276: Am I right that this obidgen is implemented incorrectly because it relays that absolute obstacle ids are in range 0 - X and now they start after regular obstacles. So it will often give id of regular obstacle and we will use it to lookup in absoluteObstacles?

nullkiller: I begin to understand. Looks like we have two ids here. One is internal ContentLoader assigned index which is in objects collection and one - an index in obstacles/absoluteObstacles arrays. This looks like some mess. And indexes in JSON looks like just string ids and can be replaced with o1, o2, o3, o345 literals for isntance. Or by animation name.

lib/CHeroHandler.h
```diff
@@ -340,8 +310,6 @@ class DLL_LINKAGE CHeroHandler : public CHandlerBase<HeroTypeID, HeroType, CHero
 		h & expPerLevel;
 		h & ballistics;
 		h & terrCosts;
-		h & obstacles;
```
nullkiller, line 343: Need to check if they are restored correctly after game load. For instance if save has less obstacles than current configs

nullkiller:
In general I suggest to get rid of obstacles/absoluteObstacles collections and replace them with method getObstacles(bool absolute). Ensure that this method is either cached or not invoked too often

lib/CHeroHandler.h
Nordsoft91: But it's just serialised in another place (in ObstacleHandler)

lib/ObstacleHandler.cpp
```diff
@@ -0,0 +1,115 @@
...
+	else
+		obstacles.push_back(info->getId());
+	
+	return info;
+}
+
+std::vector<JsonNode> ObstacleHandler::loadLegacyData(size_t dataSize)
+{
+	return std::vector<JsonNode>();
```
kambala-decapitator, line 102:
```suggestion
	return {};
```

lib/ObstacleHandler.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+std::vector<bool> ObstacleHandler::getDefaultAllowed() const
+{
+	return std::vector<bool>();
```
kambala-decapitator, line 107:
```suggestion
	return {};
```

lib/ObstacleHandler.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const std::vector<std::string> & ObstacleHandler::getTypeNames() const
+{
+	static const  std::vector<std::string> types = std::vector<std::string> { "obstacle" };
+
```
kambala-decapitator, line 113:
```suggestion
	static const std::vector<std::string> types = { "obstacle" };
```

kambala-decapitator (commented):
sorry, can't comment on the logic changes because I don't know the engine yet

Nordsoft91:
Ready for merge

PR #890, opened 2022-09-13
[macOS] perform ad-hoc codesigning

kambala-decapitator:
It seems that macOS ARM builds require signature to prevent run error "VCMI is damaged and must be moved to Trash". Although it's not quite clear if any ARM build requires this or only cross-compiled ones.

Some details can be found in https://github.com/golang/go/issues/42684

Full explanation: https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-universal-apps-release-notes

> New in macOS 11 on Macs with Apple silicon, and starting in macOS Big Sur 11 beta 6, the operating system enforces that any executable must be signed before it’s allowed to run.
> This new policy doesn’t apply to translated x86 binaries running under Rosetta 2, nor does it apply to macOS 11 running on Intel-based platforms.

### DIFF
--- a/osx/CMakeLists.txt
+++ b/osx/CMakeLists.txt
@@ -44,6 +44,20 @@ if(APPLE)
 		)
 		file(REMOVE \"${bundleContentsDir}/conan_imports_manifest.txt\")
 	")
+
+	# perform ad-hoc codesigning
+	set(codesignCommand "codesign --verbose=4 --force --options=runtime --timestamp=none --sign -")
+	set(codesignCommandWithEntitlements "${codesignCommand} --entitlements \"${CMAKE_SOURCE_DIR}/osx/entitlements.plist\"")
+	install(CODE "
+		execute_process(COMMAND
+			${codesignCommand} \"${bundleContentsDir}/MacOS/vcmibuilder\"
+		)
+		foreach(executable vcmiclient vcmiserver vcmilauncher)
+			execute_process(COMMAND
+				${codesignCommandWithEntitlements} \"${bundleContentsDir}/MacOS/\${executable}\"
+			)
+		endforeach()
+	")
 endif(APPLE)
 
 # This will likely only work for Vcpkg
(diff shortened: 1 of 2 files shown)


kambala-decapitator:
can't crack it so far. Noticed that the generated Info.plist is pretty broken (no bundle ID, no version strings), but even after fixing it and signing everything manually the right way (_VCMI.app satisfies its Designated Requirement_) it still says "...must be moved to Bin" :(

@tomkidd any idea what might be missing?

kambala-decapitator:
have some progress, hope to fully solve tomorrow

tomkidd:
I know there's an argument against it, but have you tried using the `--deep` flag on the `codesign` utility when signing the app bundle?

kambala-decapitator:
> I know there's an argument against it, but have you tried using the `--deep` flag on the `codesign` utility when signing the app bundle?

yes, unfortunately it didn't help.

solution in short: use "disable library validation" entitlement when signing executables.

Nordsoft91:
Looks good but do we use install action on GitHub?
I'd suggest new workflow to be ran manually to prepare signed builds

kambala-decapitator:
> Looks good but do we use install action on GitHub?
> I'd suggest new workflow to be ran manually to prepare signed builds

yes, `cpack` runs `install` behind the scenes.

signing workflow will be changed once we have a real (paid) signing certificate.

kambala-decapitator:
should be good now.

@tomkidd could you please test on your side? build: https://github.com/vcmi/vcmi/suites/8283024537/artifacts/362951835

tomkidd:
Yes from that build (the dmg, then the app in it) I'm getting the expected behavior from an ad hoc build on my M1 Max Mac - initial rejection followed by working if you right-click and choose Open.

If you're building your own Qt dylibs and signing them it doesn't make sense that it would be necessary to disable library validation in the entitlements but if there's something I've learned about Mac development is that a lot of things don't make sense to me yet.

PR #892, opened 2022-09-14
Bonus and string optimization

DjWarmonger:
Small buf effective improvements to performance-critical bonus list, based on MSVS profiling (Nullkller AI in lategame)

- Generating caching string with `boost::format` turned out to be more costly than actual cache lookup
- `Int3` logging wasted some CPU time too
- Reallocating `BonusList` was very costly, too

`We may also want to replace `boost::format` in favour of fmt library in the future


Nordsoft91:
Need to rebase over last develop, too many files

PR #893, opened 2022-09-14
Lambda decltype on compile time instead of runtime

Nordsoft91:


kambala-decapitator:
- pairs of functions look identical, can't you call one from another? This way copy-paste will be reduced.
- please give a better name to function parameter: `foo` is a traditional name for dummy variables

PR #894, opened 2022-09-14
Bonus and string optimization

DjWarmonger:
Small buf effective improvements to performance-critical bonus list, based on MSVS profiling (Nullkller AI in lategame)

- Generating caching string with `boost::format` turned out to be more costly than actual cache lookup
- `Int3` logging wasted some CPU time too
- Reallocating `BonusList` was very costly, too

We may also want to replace `boost::format` in favour of `fmt` library in the future

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -663,7 +663,7 @@ int IBonusBearer::valOfBonuses(Bonus::BonusType type, int subtype) const
 	//This part is performance-critical
 
 	char cachingStr[20] = {};
-	sprintf(cachingStr, "type_%ds_%d", (int)type, subtype);
+	std::sprintf(cachingStr, "type_%ds_%d", (int)type, subtype);
 
 	CSelector s = Selector::type()(type);
 	if(subtype != -1)
@@ -693,7 +693,7 @@ bool IBonusBearer::hasBonusOfType(Bonus::BonusType type, int subtype) const
 {
 	//This part is performance-ciritcal
 	char cachingStr[20] = {};
-	sprintf(cachingStr, "type_%ds_%d", (int)type, subtype);
+	std::sprintf(cachingStr, "type_%ds_%d", (int)type, subtype);
 
 	CSelector s = Selector::type()(type);
 	if(subtype != -1)
@@ -975,7 +975,7 @@ void CBonusSystemNode::getAllBonusesRec(BonusList &out) const
 	}
 	bonuses.getAllBonuses(beforeUpdate);
 
-	for(auto & b : beforeUpdate)
+	for(const auto & b : beforeUpdate)
 	{
 		auto updated = b->updater 
 			? getUpdatedBonus(b, b->updater) 
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -555,6 +555,7 @@ void BonusList::getBonuses(BonusList & out, const CSelector &selector) const
 
 void BonusList::getBonuses(BonusList & out, const CSelector &selector, const CSelector &limit) const
 {
+	out.reserve(bonuses.size());
 	for (auto & b : bonuses)
 	{
 		//add matching bonuses that matches limit predicate or have NO_LIMIT if no given predicate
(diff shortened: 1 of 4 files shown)


lib/HeroBonus.cpp
```diff
@@ -943,21 +953,34 @@ void CBonusSystemNode::getAllParents(TCNodes & out) const //retrieves list of pa
...
+	}
 
+	for (auto parent : lparents)
+	{
+		parent->getAllBonusesRec(beforeUpdate);
+	}
 	bonuses.getAllBonuses(beforeUpdate);
 
-	for(auto b : beforeUpdate)
+	for(auto & b : beforeUpdate)
```
kambala-decapitator, line 978:
```suggestion
	for(const auto & b : beforeUpdate)

```

kambala-decapitator:
- please prefix sprintf calls with `std::`
- what if you change magic numbers to `std::snprintf` calls with buffer size pre-calculation?

DjWarmonger:
We can precisely estimate how much space will strings take, I expect ids or coordinates not to be longer than 4 digits. There is still some reserve left.

lib/int3.h
```diff
@@ -158,11 +158,11 @@ class int3
...
 	std::string toString() const
 	{
-		std::string result("(");
-		result += boost::lexical_cast<std::string>(x); result += ' ';
-		result += boost::lexical_cast<std::string>(y); result += ' ';
-		result += boost::lexical_cast<std::string>(z); result += ')';
-		return result;
+		//Performance is important here
+		char str[16] = {};
+		sprintf(str, "(%d %d %d)", x, y, z);
```
kambala-decapitator, line 163: if all coordinates have 4 digits, then you'll have no space for the null terminator.

```suggestion
		char str[17] = {};
		sprintf(str, "(%d %d %d)", x, y, z);
```

kambala-decapitator (changes requested):

lib/int3.h
DjWarmonger: z coordinate does have exactly one digit - it's 0 or 1 at most.

kambala-decapitator: didn't know, thanks!

DjWarmonger: I also realistically think that map sizes above 252 x 252 are pointless, certainly no need for 4 digits.

lib/int3.h
```diff
@@ -158,11 +158,11 @@ class int3
 	//returns "(x y z)" string
 	std::string toString() const
 	{
-		std::string result("(");
-		result += boost::lexical_cast<std::string>(x); result += ' ';
-		result += boost::lexical_cast<std::string>(y); result += ' ';
-		result += boost::lexical_cast<std::string>(z); result += ')';
-		return result;
+		//Performance is important here
+		char str[16] = {};
```
Nordsoft91, line 162: as you have conversion to string anyway, it make sense to have std::string, which also has reserve method.
Optimizer of c++11 will automatically apply move semantic for returning value.

DjWarmonger: Okay, but how do I print into that string? We're still at C++14.

kambala-decapitator: use `&str[0]`

DjWarmonger: reserve() is useless here:
https://stackoverflow.com/questions/9521629/stdstringss-capacity-reserve-resize-functions

kambala-decapitator: yeah, indeed. Then you need to call constructor that allows passing count and character (`'\0'`).

DjWarmonger: Or maybe it is https://www.ibm.com/docs/en/i/7.4?topic=instructions-compare-null-terminated-strings-strcmpnull

Hope this doesn't break

kambala-decapitator: `sprintf` adds trailing null, I don't think it's something what we want to add into `std::string`

kambala-decapitator: if the returned string is then compared with `==`, there can be unintended results

DjWarmonger: No, int3 is not compared here fortunatelly.

Adding trailing `\0` won't hurt us as string is alreayd initialized wih `\0`.

Either way, I think we're overthinking it here

DjWarmonger: Okay, this didn't work:
```
TRACE ai [7708] - Cluster Monster (110 134 0)      count: 1
TRACE ai [7708] - Cluster Monster (97 126 0)       count: 11
TRACE ai [7708] - Cluster Monster (125 40 1)       count: 1
TRACE ai [7708] - Cluster Monster (99 78 0)        count: 5
TRACE ai [7708] - Cluster Monster (78 38 1)        count: 2
TRACE ai [7708] - Cluster Monster (86 98 0)        count: 2
TRACE ai [7708] - Cluster Monster (29 8 0)         count: 3
TRACE ai [7708] - Cluster Monster (105 26 1)       count: 5
```

PR #898, opened 2022-09-14
Cache pathfinding

DjWarmonger:
It's simple and effective.

### DIFF
--- a/AI/Nullkiller/Goals/ExecuteHeroChain.cpp
+++ b/AI/Nullkiller/Goals/ExecuteHeroChain.cpp
@@ -107,7 +107,7 @@ void ExecuteHeroChain::accept(AIGateway * ai)
 						|| targetNode->turns != 0)
 					{
 						logAi->error(
-							"Enable to complete chain. Expected hero %s to arive to %s in 0 turns but he can not do this",
+							"Unable to complete chain. Expected hero %s to arive to %s in 0 turns but he can not do this",
 							hero->name,
 							node.coord.toString());
 
--- a/AI/Nullkiller/Goals/ExecuteHeroChain.cpp
+++ b/AI/Nullkiller/Goals/ExecuteHeroChain.cpp
@@ -107,7 +107,7 @@ void ExecuteHeroChain::accept(AIGateway * ai)
 						|| targetNode->turns != 0)
 					{
 						logAi->error(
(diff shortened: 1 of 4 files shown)


lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -103,11 +103,9 @@ ui32 CGHeroInstance::getTileCost(const TerrainTile & dest, const TerrainTile & f
 		&& !ti->hasBonusOfType(Bonus::NO_TERRAIN_PENALTY, from.terType.id()) //no special movement bonus
 		)
 	{
-		static const CSelector selectorPATHFINDING = Selector::typeSubtype(Bonus::SECONDARY_SKILL_PREMY, SecondarySkill::PATHFINDING);
-		static const std::string keyPATHFINDING = "type_"+std::to_string((si32)Bonus::SECONDARY_SKILL_PREMY)+"s_"+std::to_string((si32)SecondarySkill::PATHFINDING);
 
 		ret = VLC->heroh->terrCosts[from.terType];
```
nullkiller, line 107: Really it should be possible to calculate and cache the whole tile cost by terrain type. At least what I have seen so far we only relay on source tile terrain type and separate logic for roads. So taking terrain cost by terrain type from some map should be even faster. So I might be missing something

nullkiller (approved):

lib/mapObjects/CGHeroInstance.cpp
DjWarmonger: Terrain type will get total rewrite from me, actually I already did that.

AI/Nullkiller/Goals/ExecuteHeroChain.cpp
```diff
@@ -107,7 +107,7 @@ void ExecuteHeroChain::accept(AIGateway * ai)
 						|| targetNode->turns != 0)
 					{
 						logAi->error(
-							"Enable to complete chain. Expected hero %s to arive to %s in 0 turns but he can not do this",
+							"Unable to complete chain. Expected hero %s to arive to %s in 0 turns but he can not do this",
```
kambala-decapitator, line 110:
```suggestion
							"Unable to complete chain. Expected hero %s to arrive to %s in 0 turns but he cannot do this",
```

Nordsoft91 (approved):

PR #899, opened 2022-09-14
NK: fix AI speed

nullkiller:
This branch has goal to reduce AI turn time by reducing hero chain usage. It is kind of experimental.
- fast tasks which does not affect (at least significantly) chain because they do not move heroes
- twice reduce storage size
- turn of hero chain at later stages of turn when scouts move mostly

### DIFF
--- a/AI/Nullkiller/Analyzers/ObjectClusterizer.cpp
+++ b/AI/Nullkiller/Analyzers/ObjectClusterizer.cpp
@@ -120,18 +120,23 @@ const CGObjectInstance * ObjectClusterizer::getBlocker(const AIPath & path) cons
 
 		auto blocker = blockers.front();
 
+		if(isObjectPassable(ai, blocker))
+			continue;
+
 		if(blocker->ID == Obj::GARRISON
-			|| blocker->ID == Obj::MONSTER
-			|| blocker->ID == Obj::GARRISON2
-			|| blocker->ID == Obj::BORDERGUARD
-			|| blocker->ID == Obj::BORDER_GATE
-			|| blocker->ID == Obj::SHIPYARD)
+			|| blocker->ID == Obj::GARRISON2)
 		{
-			if(!isObjectPassable(ai, blocker))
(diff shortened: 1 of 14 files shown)


nullkiller:
Looks like this thing acts more or less the same while significantly reducing memory usage. So lets merge it.

AI/Nullkiller/Analyzers/ObjectClusterizer.cpp
```diff
@@ -120,18 +120,23 @@ const CGObjectInstance * ObjectClusterizer::getBlocker(const AIPath & path) cons
...
+			else
 				return blocker;
 		}
 
-		if(blocker->ID == Obj::QUEST_GUARD && node->actionIsBlocked)
+		if(blocker->ID == Obj::MONSTER
+			|| blocker->ID == Obj::BORDERGUARD
+			|| blocker->ID == Obj::BORDER_GATE
+			|| blocker->ID == Obj::SHIPYARD
+			|| blocker->ID == Obj::QUEST_GUARD && node->actionIsBlocked)
```
AgostonSzepessy, line 139: I recommend putting all the `||`s into parentheses because otherwise it might not give the behavior you expect, unless of course, this was intentional.

```suggestion
					if((blocker->ID == Obj::MONSTER
			|| blocker->ID == Obj::BORDERGUARD
			|| blocker->ID == Obj::BORDER_GATE
			|| blocker->ID == Obj::SHIPYARD
			|| blocker->ID == Obj::QUEST_GUARD) && node->actionIsBlocked)
```

AI/Nullkiller/Pathfinding/AINodeStorage.cpp
```diff
@@ -1060,6 +1056,55 @@ struct TowmPortalFinder
 	}
 };
 
+template<class T>
```
AgostonSzepessy, line 1059: Could you please rename the template parameter to `Vector` to make it more explicit that it's a vector?

```suggestion
template<class Vector>
```

PR #900, opened 2022-09-14
Map array rotation

DjWarmonger:
Map coordinates were reordered so now the map has shape of [z][x][y]. This has numerous advantages:

- For the most part, we are interested in only surface or underground. Now the tiles on same plane are located closer in memory, which should improve cache access both for pathfinding and rendering.
- The innermost loop iterating over 1 or 2 indexes is a waste of CPU cycles.
- Some abysmal `vector<vector<vector>>>` was replaced by `multi_array`, too.
- Thanks to `boost::multi_array` we can slice the interesting plane of the map. This already has its use in renderer and will probably be useful in the future, with editor or GUI rewrite.

Additionally, Nullkiller AI got similiar treatment. Now the outermost dimension is layer (ground, fly, waterwalk) which rarely changes and is often unneccessary. The innermost dimension are chain nodes which are processed together in the innermost loop.


AI/VCAI/SectorMap.cpp
```diff
@@ -71,34 +71,27 @@ void SectorMap::update()
...
-	//std::transform(fow.begin(), fow.end(), sector.begin(), [](const ui8 &f) -> unsigned short
-	//{
-	//	return f; //type conversion
-	//});
-	auto width = fow.size();
-	auto height = fow.front().size();
-	auto depth = fow.front().front().size();
-	for (size_t x = 0; x < width; x++)
+
+	//TODO: Template function to automate copying arrays?
```
nullkiller, line 87: If I am not mistaken this thing is no longer used. SectorMap I mean

nullkiller (changes requested):
Rebase required. Commits from old features/new-terrains should not be a part of develop because it was merged by rebase

PR #901, opened 2022-09-14
Add 1.0.0 release to appstream metadata

Alexander-Wilms:


nullkiller:
Would be nice to know what is this xml for?

Alexander-Wilms:
It's used for application descriptions in app stores on Linux, like Flathub, KDE Discover or Gnome Software:

https://beta.flathub.org/de/apps/details/eu.vcmi.VCMI

PR #902, opened 2022-09-14
Support rewardable objects

Nordsoft91:
I realised that we have many objects from hota, which are not supported (warehouses, magic coliseum, etc.)

I found that we already have class and handler to support it, but unfortunately it didn't work from the box.
So here are small changes to make it possible.

How to test: extract files to "HotA decorations" submod and play map attached
[HotA decorations.zip](https://github.com/vcmi/vcmi/files/9569095/HotA.decorations.zip)
[test3.vmap.txt](https://github.com/vcmi/vcmi/files/9569102/test3.vmap.txt)

### DIFF
--- a/lib/mapObjects/CRewardableObject.h
+++ b/lib/mapObjects/CRewardableObject.h
@@ -175,6 +175,12 @@ public:
 	}
 };
 
+namespace Rewardable
+{
+	const std::array<std::string, 3> SelectModeString{"selectFirst", "selectPlayer", "selectRandom"};
+	const std::array<std::string, 5> VisitModeString{"unlimited", "once", "hero", "bonus", "player"};
+}
+
 /// Base class that can handle granting rewards to visiting heroes.
 /// Inherits from CArmedInstance for proper trasfer of armies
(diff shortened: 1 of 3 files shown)


lib/mapObjects/CRewardableObject.h
```diff
@@ -175,6 +175,12 @@ class DLL_LINKAGE CVisitInfo
...
 
+namespace Rewardable
+{
+	const std::array<std::string, 3> SelectModeString{"selectFirst", "selectPlayer", "selectRandom"};
+	const std::array<std::string, 5> VisitModeString{"unlimited", "once", "hero", "bonus", "player"};
+}
+
 /// Base class that can handle granting rewards to visiting heroes.
 /// Inherits from CArmedInstance for proper trasfer of armies
 class DLL_LINKAGE CRewardableObject : public CArmedInstance
```
nullkiller, line 186: Who inherits it?

nullkiller, line 186: This PR affects a lot! CRewardableObject is inherited by a set of other classes. What does these new settings mean?

Issue #840, opened 2022-09-06
client bugs when assembling artefact set


kambala-decapitator:
Basically a duplicate of https://bugs.vcmi.eu/view.php?id=3039

https://user-images.githubusercontent.com/1557784/188576663-34b9d7cd-c799-4214-b4fe-9c80d2f4794d.mov

As clearly seen from the video, when you try to assemble a set (AotD in this case), something gets wrong with mouse cursor: the real cursor and the in-game one no longer have the same coordinates, which results in inability to click with the in-game cursor, and the behavior persists after dismissing the dialog.

Also, after clicking OK, the replaced artefact (Targ of Ogre) doesn't appear in the bag immediately, you have to click somewhere first.

Save game: [Archive.tar.gz](https://github.com/vcmi/vcmi/files/9494429/Archive.tar.gz), hero Daneel (Forge + new terrains).

kambala-decapitator:
happens only when last artefact from a set replaces some other artefact. If you place last artefact in an empty slot or assemble via right click, the issue is not reproducible.

kambala-decapitator:
test map: https://github.com/vcmi/vcmi/files/9689733/seer-2-artifacts.h3m.zip Put on Admiral's Hat, disassemble it, change amulet, put on the Dolphin amulet.

kambala-decapitator:
I believe this strange mouse offset is related to the switched artifact that becomes "invisible".

the issue seems in `CPlayerInterface::artifactMoved()`, probably related to move and assemble events.

PR #907, opened 2022-09-15
Bulk swap artifacts

AgostonSzepessy:
Implements bulk swapping artifacts in a single packet to optimize this operation, like `BulkMoveArmy` does for transferring stacks. Most of the implementation was based on that, and the existing `MoveArtifact` stuff. This tries to reuse as much of the existing `MoveArtifact` stuff as possible, and some refactoring was done to achieve that.

The biggest difference is that a set of the destination of the transferred needs to be kept while building up the list of destinations. This needs to be done because the changes don't register until the clients have been notified of the changes.

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -2582,6 +2582,9 @@ void CPlayerInterface::artifactMoved(const ArtifactLocation &src, const Artifact
 		if (artWin)
 			artWin->artifactMoved(src, dst);
 	}
+
+	GH.totalRedraw();
+
 	askToAssembleArtifact(dst);
 }
 
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -2583,6 +2583,8 @@ void CPlayerInterface::artifactMoved(const ArtifactLocation &src, const Artifact
 			artWin->artifactMoved(src, dst);
 	}
 
+	// This needs to be called otherwise the artifacts don't get displayed until
+	// the window is closed and reopened
 	GH.totalRedraw();
 
 	askToAssembleArtifact(dst);
--- a/lib/CArtHandler.h
+++ b/lib/CArtHandler.h
@@ -362,7 +362,7 @@ private:
 namespace ArtifactUtils
 {
 	// Calculates where an artifact gets placed when it gets transferred from one hero to another.
-	DLL_LINKAGE ArtifactLocation getArtifactDstLocation(const CGHeroInstance * source, const CGHeroInstance * target, ArtifactPosition srcPosition);
+	DLL_LINKAGE ArtifactLocation getArtifactDstLocation(const CGHeroInstance * source, const CGHeroInstance * target, ArtifactPosition srcPosition, std::set<ArtifactPosition::EArtifactPosition> dstOccupiedSlots);
 	DLL_LINKAGE std::vector<ArtifactPosition> unmovablePositions();
 	DLL_LINKAGE bool isArtRemovable(const std::pair<ArtifactPosition, ArtSlotInfo> & slot);
 }
--- a/lib/CArtHandler.h
+++ b/lib/CArtHandler.h
@@ -252,8 +252,6 @@ public:
 	std::vector<CArtifact *> allowedArtifacts;
 	std::set<ArtifactID> growingArtifacts;
 
-	static const std::vector<ArtifactPosition> UNMOVABLE_POSITIONS;
-
 	void addBonuses(CArtifact *art, const JsonNode &bonusList);
 
 	void fillList(std::vector<CArtifact*> &listToBeFilled, CArtifact::EartClass artifactClass); //fills given empty list with allowed artifacts of given class. No side effects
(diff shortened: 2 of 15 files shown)


Nordsoft91:
Isn't it duplicating same functionality, implemented in vcmi-quick-exchange mod?
Could you please describe what is the change or maybe attach screenshots?

AgostonSzepessy:
This refactors that functionality and makes it more efficient. Previously, it was sending a packet for every artifact moved. Now, it just sends one.

lib/CArtHandler.h
```diff
@@ -230,6 +230,20 @@ class DLL_LINKAGE CCombinedArtifactInstance : public CArtifactInstance
 	}
 };
 
+struct DLL_LINKAGE ArtSlotInfo
```
Nordsoft91, line 233: Let's try to avoid such changes

lib/NetPacks.h
```diff
@@ -988,6 +988,19 @@ struct MoveArtifact : CArtifactOperationPack
...
+struct BulkMoveArtifact : CArtifactOperationPack
+{
+	std::vector<MoveArtifact> artifacts;
+
+	void applyCl(CClient * cl);
+	DLL_LINKAGE void applyGs(CGameState * gs);
+
+	template <typename Handler> void serialize(Handler & h, const int version)
+	{
+		h & artifacts;
```
Nordsoft91, line 1000: Why it's needed to be resialized?

lib/NetPacks.h
```diff
@@ -2179,6 +2192,25 @@ struct ExchangeArtifacts : public CPackForServer
...
+
+	BulkExchangeArtifacts() = default;
+
+	BulkExchangeArtifacts(ObjectInstanceID srcHero, ObjectInstanceID dstHero) : srcHero(srcHero), dstHero(dstHero)
+	{}
+
+	bool applyGh(CGameHandler* gh);
+	template <typename Handler> void serialize(Handler& h, const int version)
+	{
+		h & static_cast<CPackForServer&>(*this);
```
Nordsoft91, line 2208: Why it's needed to be serialised?

server/CGameHandler.cpp
```diff
@@ -61,6 +63,7 @@
 #define COMPLAIN_RET_FALSE_IF(cond, txt) do {if (cond){complain(txt); return false;}} while(0)
 #define COMPLAIN_RET(txt) {complain(txt); return false;}
 #define COMPLAIN_RETF(txt, FORMAT) {complain(boost::str(boost::format(txt) % FORMAT)); return false;}
+#define COMPLAIN_RET_OPT_NONE(txt) { complain(txt); return {}; }
```
Nordsoft91, line 66: Is it boost::optional, please return boost::none

Nordsoft91, line 66:
```suggestion
#define COMPLAIN_RET_OPTIONAL_NONE(txt) { complain(txt); return boost::none; }
```

server/CGameHandler.h
```diff
@@ -345,6 +346,9 @@ class CGameHandler : public IGameCallback, public CBattleInfoCallback, public En
 	void checkVictoryLossConditions(const std::set<PlayerColor> & playerColors);
 	void checkVictoryLossConditionsForAll();
 
+	bool moveArtifactImpl(const ArtifactLocation & al1, const ArtifactLocation & al2, std::vector<MoveArtifact> * artifactsToMove);
```
Nordsoft91, line 349: usually Implement postfix means some override over virtual function. If it's not a case, consider another name (maybe moveArtifactInternal and make it private/protected member)

lib/CArtHandler.h
AgostonSzepessy: Yes, forgot to switch that back

server/CGameHandler.cpp
AgostonSzepessy: Done. Didn't know that existed

server/CGameHandler.h
AgostonSzepessy: Done. It was already private

client/CPlayerInterface.cpp
```diff
@@ -2582,6 +2582,9 @@ void CPlayerInterface::artifactMoved(const ArtifactLocation &src, const Artifact
 		if (artWin)
 			artWin->artifactMoved(src, dst);
 	}
+
+	GH.totalRedraw();
```
Nordsoft91, line 2586: what for is it? does it impact performance?

AgostonSzepessy: If that's not there, the transferred artifacts don't get displayed unless the window is closed and reopened.

AgostonSzepessy: I'll add a comment

nullkiller:
First test failed. Try to move Angelic alliance, armor of damned and dragon set with bulk move. Two arts should go to backpack but one get worn instead

nullkiller:
[artifacts-manual.h3m.txt](https://github.com/vcmi/vcmi/files/9583255/artifacts-manual.h3m.txt)

nullkiller (changes requested):
Set review status to Request changes.

AgostonSzepessy:
@nullkiller What do the artifact slots mean that have a lock on them? It looks like they contain artifacts. Is that just dummy information to satisfy the engine? Is the important thing that they have `locked` set on them?

locked-artifacts

kambala-decapitator:
here's a live example of an artefact set: #840

nullkiller:
@AgostonSzepessy converted to draft until it is fixed and ready for review once again

nullkiller:
Closing this in favor to bulk arts swap v2 PR

Issue #913, opened 2022-09-15
Avatars quest artefact assembly crash


nullkiller:
On Avatars quest map there is an area with artifacts - components of Cornucopia or something (+4 sulfur, mercury, crystal and gems). You can assemble 2 Cornucopia from these artifacts. When you assemble the first one and pick mercury vial it asks you to assemble cornucopia once more. If you agree - server crashes

Alucard648:
Try to test with other dupe combo artifacts.

PR #914, opened 2022-09-15
#913 fix artifact assembling check

nullkiller:
Artefact assembly check only checks that hero wears required artifacts but not if they are already a part of another assembly, e.g. locked


Nordsoft91:
I suggest to test sandro campaign (first mission)

PR #929, opened 2022-09-17
Mods versioning [part 3]

Nordsoft91:
Those changes were developed in parallel with develop, so I had to resolve ~100 conflicts to merge everything.
Basically, change have 3 big topics:
 - VLC for obstacles
 - Mod versioning solution (without Launcher part) - that's required for VLC for obstacles as it breaks not only backward compatibility for saves, but also all mods with new obstacles become incompatible
 - Some features developed by me and @DjWarmonger for further support of editor, in particular
 > progress bar to reflect loading progress in parallel thread (can be used to show loading bar in game during rmg computations)
> undo/redo implementation for map
> filesystem missing functions

Test scenarios to check mod versions
1. run game regular -> save game -> load game
2. run game with mod A -> save game -> load game -> load game from p1
3. run game with mod B -> save game -> load game -> load game from p1, p2
4. run game with next version of A, B -> save game -> load game from p2, p3 (error expected)
5. run game as in p.3 -> load game from p4 (error expected)

### DIFF
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -26,7 +26,7 @@ bool isCompatible(const QString & verMin, const QString & verMax)
 	auto versionMin = QVersionNumber::fromString(verMin);
 	auto versionMax = QVersionNumber::fromString(verMax);
 	
-	auto buildVersion = [](QVersionNumber & ver)
+	auto buildVersion = [maxSections](QVersionNumber & ver)
 	{
 		if(ver.segmentCount() < maxSections)
 		{
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -12,11 +12,59 @@
 
 #include "../../lib/JsonNode.h"
 #include "../../lib/filesystem/CFileInputStream.h"
+#include "../../lib/GameConstants.h"
+
+const int maxSections = 3; // versions consist from up to 3 sections, major.minor.patch
+
+bool isCompatible(const QString & verMin, const QString & verMax)
+{
+	QList<int> vcmiVersionList = {GameConstants::VCMI_VERSION_MAJOR,
+								  GameConstants::VCMI_VERSION_MINOR,
+								  GameConstants::VCMI_VERSION_PATCH};
+
+	if(!verMin.isEmpty())
+	{
+		QStringList verMinList = verMin.split(".");
+		assert(verMinList.size() == maxSections);
+		bool compatibleMin = true;
+		for(int i = 0; i < maxSections; i++)
+		{
+			if(verMinList[i].toInt() < vcmiVersionList[i])
+			{
+				break;
+			}
+			if(verMinList[i].toInt() > vcmiVersionList[i])
+			{
+				compatibleMin = false;
+				break;
+			}
+		}
+
+		if(!compatibleMin)
+			return false;
+	}
+
+	if(!verMax.isEmpty())
+	{
+		QStringList verMaxList = verMax.split(".");
+		assert(verMaxList.size() == maxSections);
+		for(int i = 0; i < maxSections; i++)
+		{
+			if(verMaxList[i].toInt() > vcmiVersionList[i])
+			{
+				return true;
+			}
+			if(verMaxList[i].toInt() < vcmiVersionList[i])
+			{
+				return false;
+			}
+		}
+	}
+	return true;
+}
 
 bool CModEntry::compareVersions(QString lesser, QString greater)
 {
-	static const int maxSections = 3; // versions consist from up to 3 sections, major.minor.patch
-
 	QStringList lesserList = lesser.split(".");
 	QStringList greaterList = greater.split(".");
 
@@ -92,6 +140,15 @@ bool CModEntry::isUpdateable() const
 	return false;
 }
 
+bool CModEntry::isCompatible() const
+{
+	if(!isInstalled())
+		return false;
+
+	auto compatibility = localData["compatibility"].toMap();
+	return ::isCompatible(compatibility["min"].toString(), compatibility["max"].toString());
+}
+
 bool CModEntry::isEssential() const
 {
 	return getValue("storedLocaly").toBool();
@@ -100,6 +157,11 @@ bool CModEntry::isEssential() const
 bool CModEntry::isInstalled() const
 {
 	return !localData.isEmpty();
+}
+
+bool CModEntry::isValid() const
+{
+	return !localData.isEmpty() || !repository.isEmpty();
 }
 
 int CModEntry::getModStatus() const
@@ -193,7 +255,11 @@ static QVariant getValue(QVariant input, QString path)
 		QString remainder = "/" + path.section('/', 2, -1);
 
 		entryName.remove(0, 1);
-		return getValue(input.toMap().value(entryName), remainder);
+		QMap<QString, QString> keyNormalize;
+		for(auto & key : input.toMap().keys())
+			keyNormalize[key.toLower()] = key;
+
+		return getValue(input.toMap().value(keyNormalize[entryName]), remainder);
 	}
 	else
 	{
@@ -203,6 +269,7 @@ static QVariant getValue(QVariant input, QString path)
 
 CModEntry CModList::getMod(QString modname) const
 {
+	modname = modname.toLower();
 	QVariantMap repo;
 	QVariantMap local = localModList[modname].toMap();
 	QVariantMap settings;
@@ -246,14 +313,14 @@ CModEntry CModList::getMod(QString modname) const
 		QVariant repoVal = getValue(entry, path);
 		if(repoVal.isValid())
 		{
-			if(repo.empty())
+			auto repoValMap = repoVal.toMap();
+			auto compatibility = repoValMap["compatibility"].toMap();
+			if(isCompatible(compatibility["min"].toString(), compatibility["max"].toString()))
 			{
-				repo = repoVal.toMap();
-			}
-			else
-			{
-				if(CModEntry::compareVersions(repo["version"].toString(), repoVal.toMap()["version"].toString()))
-					repo = repoVal.toMap();
+				if(repo.empty() || CModEntry::compareVersions(repo["version"].toString(), repoValMap["version"].toString()))
+				{
+					repo = repoValMap;
+				}
 			}
 		}
 	}
@@ -297,12 +364,12 @@ QVector<QString> CModList::getModList() const
 	{
 		for(auto it = repo.begin(); it != repo.end(); it++)
 		{
-			knownMods.insert(it.key());
+			knownMods.insert(it.key().toLower());
 		}
 	}
 	for(auto it = localModList.begin(); it != localModList.end(); it++)
 	{
-		knownMods.insert(it.key());
+		knownMods.insert(it.key().toLower());
 	}
 
 	for(auto entry : knownMods)
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -14,75 +14,51 @@
 #include "../../lib/filesystem/CFileInputStream.h"
 #include "../../lib/GameConstants.h"
 
-const int maxSections = 3; // versions consist from up to 3 sections, major.minor.patch
-
+namespace
+{
 bool isCompatible(const QString & verMin, const QString & verMax)
 {
-	QList<int> vcmiVersionList = {GameConstants::VCMI_VERSION_MAJOR,
-								  GameConstants::VCMI_VERSION_MINOR,
-								  GameConstants::VCMI_VERSION_PATCH};
-
-	if(!verMin.isEmpty())
+	const int maxSections = 3; // versions consist from up to 3 sections, major.minor.patch
+	QVersionNumber vcmiVersion(GameConstants::VCMI_VERSION_MAJOR,
+							   GameConstants::VCMI_VERSION_MINOR,
+							   GameConstants::VCMI_VERSION_PATCH);
+	
+	auto versionMin = QVersionNumber::fromString(verMin);
+	auto versionMax = QVersionNumber::fromString(verMax);
+	
+	auto buildVersion = [](QVersionNumber & ver)
 	{
-		QStringList verMinList = verMin.split(".");
-		assert(verMinList.size() == maxSections);
-		bool compatibleMin = true;
-		for(int i = 0; i < maxSections; i++)
+		if(ver.segmentCount() < maxSections)
 		{
-			if(verMinList[i].toInt() < vcmiVersionList[i])
-			{
-				break;
-			}
-			if(verMinList[i].toInt() > vcmiVersionList[i])
-			{
-				compatibleMin = false;
-				break;
-			}
+			auto segments = ver.segments();
+			for(int i = segments.size() - 1; i < maxSections; ++i)
+				segments.append(0);
+			ver = QVersionNumber(segments);
 		}
+	};
 
-		if(!compatibleMin)
+	if(!versionMin.isNull())
+	{
+		buildVersion(versionMin);
+		if(vcmiVersion < versionMin)
 			return false;
 	}
-
-	if(!verMax.isEmpty())
+	
+	if(!versionMax.isNull())
 	{
-		QStringList verMaxList = verMax.split(".");
-		assert(verMaxList.size() == maxSections);
-		for(int i = 0; i < maxSections; i++)
-		{
-			if(verMaxList[i].toInt() > vcmiVersionList[i])
-			{
-				return true;
-			}
-			if(verMaxList[i].toInt() < vcmiVersionList[i])
-			{
-				return false;
-			}
-		}
+		buildVersion(versionMax);
+		if(vcmiVersion > versionMax)
+			return false;
 	}
 	return true;
+}
 }
 
 bool CModEntry::compareVersions(QString lesser, QString greater)
 {
-	QStringList lesserList = lesser.split(".");
-	QStringList greaterList = greater.split(".");
-
-	assert(lesserList.size() <= maxSections);
-	assert(greaterList.size() <= maxSections);
-
-	for(int i = 0; i < maxSections; i++)
-	{
-		if(greaterList.size() <= i) // 1.1.1 > 1.1
-			return false;
-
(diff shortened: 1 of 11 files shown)


launcher/modManager/cmodlist.cpp
```diff
@@ -12,11 +12,59 @@
...
+			if(verMaxList[i].toInt() > vcmiVersionList[i])
+			{
+				return true;
+			}
+			if(verMaxList[i].toInt() < vcmiVersionList[i])
+			{
+				return false;
+			}
+		}
+	}
```
nullkiller, line 62: Code duplication. LEts introduce a method to compare versions

launcher/modManager/cmodlist.cpp
```diff
@@ -92,6 +140,15 @@ bool CModEntry::isUpdateable() const
 	return false;
 }
 
+bool CModEntry::isCompatible() const
+{
+	if(!isInstalled())
+		return false;
```
nullkiller, line 146: Why false? Now it looks like if mod is not installed - it is not compatible which is a bit weird.

lib/CModHandler.cpp
```diff
@@ -532,6 +532,51 @@ JsonNode addMeta(JsonNode config, std::string meta)
...
+}
+
+CModInfo::Version CModInfo::Version::fromString(std::string from)
+{
+	int major = 0, minor = 0, patch = 0;
+	try
+	{
+		auto pointPos = from.find('.');
+		major = std::stoi(from.substr(0, pointPos));
+		if(pointPos != std::string::npos)
```
nullkiller, line 547: We already called stoi and now check pointPos. What stoi will return if pointPos is npos?

lib/CModHandler.h
```diff
@@ -210,18 +240,6 @@ class DLL_LINKAGE CModInfo
 	static std::string getModDir(std::string name);
 	static std::string getModFile(std::string name);
 
-	template <typename Handler> void serialize(Handler &h, const int version)
```
nullkiller, line 213: No longer needed?

lib/CModHandler.h
```diff
@@ -256,6 +274,13 @@ class DLL_LINKAGE CModHandler
 	void loadMods(std::string path, std::string parent, const JsonNode & modSettings, bool enableMods);
 	void loadOneMod(std::string modName, std::string parent, const JsonNode & modSettings, bool enableMods);
 public:
+	
+	class Incompatibility: public std::logic_error
```
nullkiller, line 278: Do we catch this exception? If not - we need to log error before throwing it. Otherwise users might only observe a crash without any clues what caused it.

nullkiller:
What if mod has no compatibility versions? We always treat it as compatible?

launcher/modManager/cmodlist.cpp
Nordsoft91: it's because you can check compatibility for only installed mods.

lib/CModHandler.cpp
Nordsoft91: just a number. For example, you can have version = "159" without any dots

lib/CModHandler.h
Nordsoft91: No, just mod name and version are serialised

lib/CModHandler.h
Nordsoft91: This is caught on server side. Unfortunatelly I don't know how to bypass it to client . The intention is to catch specific type of excepton and show message to user. If you know how to do it please advice

Nordsoft91:
> What if mod has no compatibility versions? We always treat it as compatible?

Yes, because for some mods compatibility just doesn't make any sense

launcher/modManager/cmodlist.cpp
Nordsoft91: Take a look closer, it's different

lib/CModHandler.cpp
```diff
@@ -532,6 +532,51 @@ JsonNode addMeta(JsonNode config, std::string meta)
...
+std::string CModInfo::Version::toString() const
+{
+	return std::to_string(major) + '.' + std::to_string(minor) + '.' + std::to_string(patch);
+}
+
+bool CModInfo::Version::compatible(const Version & other, bool checkMinor, bool checkPatch) const
+{
+	return  (major == other.major &&
+			(!checkMinor || minor >= other.minor) &&
+			(!checkPatch || minor > other.minor || (minor == other.minor && patch >= other.patch)));
```
kambala-decapitator, line 572:
```suggestion
	return (major == other.major
		&& (!checkMinor || minor >= other.minor)
		&& (!checkPatch || minor > other.minor || (minor == other.minor && patch >= other.patch)));

```

lib/CModHandler.cpp
```diff
@@ -601,6 +652,13 @@ void CModInfo::loadLocalData(const JsonNode & data)
 		validated = data["validated"].Bool();
 		checksum  = strtol(data["checksum"].String().c_str(), nullptr, 16);
 	}
+	
+	//check compatibility
+	bool wasEnabled = enabled;
+	enabled &= vcmiCompatibleMin.isNull() || Version::GameVersion().compatible(vcmiCompatibleMin);
+	enabled &= vcmiCompatibleMax.isNull() || vcmiCompatibleMax.compatible(Version::GameVersion());
```
kambala-decapitator, line 659: don't treat boolean as integer:
```suggestion
	enabled = enabled && (vcmiCompatibleMin.isNull() || Version::GameVersion().compatible(vcmiCompatibleMin));
	enabled = enabled && (vcmiCompatibleMax.isNull() || vcmiCompatibleMax.compatible(Version::GameVersion()));

```

lib/CModHandler.h
```diff
@@ -177,13 +177,43 @@ class DLL_LINKAGE CModInfo
...
+	{
+		int major = 0;
+		int minor = 0;
+		int patch = 0;
+		
+		Version() = default;
+		Version(int mj, int mi, int p): major(mj), minor(mi), patch(p) {}
+		
+		static Version GameVersion();
+		static Version fromString(std::string from);
```
kambala-decapitator, line 191:
```suggestion
		static Version fromString(const std::string & from);

```

lib/CModHandler.h
```diff
@@ -177,13 +177,43 @@ class DLL_LINKAGE CModInfo
...
+		std::string toString() const;
+		
+		bool compatible(const Version & other, bool checkMinor = false, bool checkPatch = false) const;
+		bool isNull() const;
+		
+		template <typename Handler> void serialize(Handler &h, const int version)
```
kambala-decapitator, line 197: why is `version` parameter unused?

```suggestion
		template <typename Handler> void serialize(Handler & h, int version)

```

lib/CModHandler.h
```diff
@@ -177,13 +177,43 @@ class DLL_LINKAGE CModInfo
...
 	std::string identifier;
 
 	/// human-readable strings
 	std::string name;
 	std::string description;
+	
+	/// version of the mod
+	Version version;
+	
+	///The  vcmi versions compatible with the mod
```
kambala-decapitator, line 215:
```suggestion
	/// vcmi versions compatible with the mod

```

lib/CModHandler.h
```diff
@@ -336,8 +361,35 @@ class DLL_LINKAGE CModHandler
 
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
-		h & allMods;
-		h & activeMods;
+		if(h.saving)
+		{
+			h & activeMods;
+			for(auto & m : activeMods)
```
kambala-decapitator, line 367:
```suggestion
			for(const auto & m : activeMods)

```

lib/CModHandler.h
```diff
@@ -336,8 +361,35 @@ class DLL_LINKAGE CModHandler
...
+				if(!allMods.count(m))
+					throw Incompatibility(m + " unkown mod");
+				
+				CModInfo::Version mver;
+				h & mver;
+				if(!allMods[m].version.isNull() && !mver.isNull() && !allMods[m].version.compatible(mver))
+				{
+					std::string err = allMods[m].name +
+					": version needed " + mver.toString() +
+					"but you have installed " + allMods[m].version.toString();
```
kambala-decapitator, line 385:
```suggestion
					std::string err = allMods[m].name
						+ ": version needed " + mver.toString()
						+ " but you have installed " + allMods[m].version.toString();

```

lib/rmg/ObstaclePlacer.cpp
```diff
@@ -62,122 +50,172 @@ void ObstaclePlacer::process()
...
+{
+	manager->insertObjects(instances); //insert as one operation - for undo purposes
+}
+
+std::pair<bool, bool> ObstacleProxy::verifyCoverage(const int3 & t) const
+{
+	std::pair<bool, bool> result(false, false);
+	if(blockedArea.contains(t))
+		result.first = true;
+	return result;
```
kambala-decapitator, line 177:
```suggestion
	return {blockedArea.contains(t), false};
```

lib/rmg/ObstaclePlacer.cpp
```diff
@@ -189,3 +227,45 @@ void ObstaclePlacer::init()
...
 }
+
+std::pair<bool, bool> ObstaclePlacer::verifyCoverage(const int3 & t) const
+{
+	std::pair<bool, bool> result(false, false);
+	if(map.shouldBeBlocked(t))
+		result.first = true;
+	if(zone.areaPossible().contains(t))
+		result.second = true;
+	return result;
```
kambala-decapitator, line 238:
```suggestion
	return {map.shouldBeBlocked(t), zone.areaPossible().contains(t)};
```

lib/rmg/ObstaclePlacer.cpp
```diff
@@ -189,3 +227,45 @@ void ObstaclePlacer::init()
...
+}
+
+void ObstaclePlacer::postProcess(const rmg::Object & object)
+{
+	//river processing
+	if(riverManager)
+	{
+		if(object.instances().front()->object().typeName == "mountain")
+			riverManager->riverSource().unite(object.getArea());
+		if(object.instances().front()->object().typeName == "lake")
```
kambala-decapitator, line 253:
```suggestion
		const auto objTypeName = object.instances().front()->object().typeName;
		if(objTypeName == "mountain")
			riverManager->riverSource().unite(object.getArea());
		else if(objTypeName == "lake")
```

kambala-decapitator (changes requested):

launcher/modManager/cmodlist.cpp
```diff
@@ -92,6 +140,15 @@ bool CModEntry::isUpdateable() const
...
+
+	auto compatibility = localData["compatibility"].toMap();
+	return ::isCompatible(compatibility["min"].toString(), compatibility["max"].toString());
```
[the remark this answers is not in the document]
Nordsoft91: It wont't compile

launcher/modManager/cmodlist.cpp
```diff
@@ -193,7 +255,11 @@ static QVariant getValue(QVariant input, QString path)
 		QString remainder = "/" + path.section('/', 2, -1);
 
 		entryName.remove(0, 1);
-		return getValue(input.toMap().value(entryName), remainder);
+		QMap<QString, QString> keyNormalize;
+		for(auto & key : input.toMap().keys())
```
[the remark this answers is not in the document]
Nordsoft91: why should we care on this in launcher? For me readability in launcher is more important rather than performance.

launcher/modManager/cmodlist.cpp
```diff
@@ -193,7 +255,11 @@ static QVariant getValue(QVariant input, QString path)
...
+			keyNormalize[key.toLower()] = key;
```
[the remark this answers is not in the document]
Nordsoft91: That's exact problem with Forge

lib/CModHandler.cpp
```diff
@@ -532,6 +532,51 @@ JsonNode addMeta(JsonNode config, std::string meta)
 	return config;
 }
 
+CModInfo::Version CModInfo::Version::GameVersion()
+{
+	return Version(GameConstants::VCMI_VERSION_MAJOR, GameConstants::VCMI_VERSION_MINOR, GameConstants::VCMI_VERSION_PATCH);
+}
+
+CModInfo::Version CModInfo::Version::fromString(std::string from)
```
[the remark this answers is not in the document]
Nordsoft91: Nope, from is modified

lib/CModHandler.cpp
Nordsoft91: I prefer style with operators in the end of line, because It's easy to see that line is not ended.
As for indends, GitHub uses 8 spaces as one tab, instead of 4

lib/CModHandler.h
Nordsoft91: that's won't compile

lib/CModHandler.h
Nordsoft91: That's declaration of basic serialization function, it's unified across all serializable objects

Nordsoft91: version is needed to work with previous save versions

lib/rmg/ObstaclePlacer.cpp
Nordsoft91: better to avoid usage of else, but ok

launcher/modManager/cmodlist.cpp
```diff
@@ -12,11 +12,59 @@
 
 #include "../../lib/JsonNode.h"
 #include "../../lib/filesystem/CFileInputStream.h"
+#include "../../lib/GameConstants.h"
 
-bool CModEntry::compareVersions(QString lesser, QString greater)
+const int maxSections = 3; // versions consist from up to 3 sections, major.minor.patch
+
+bool isCompatible(const QString & verMin, const QString & verMax)
```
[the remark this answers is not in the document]
Nordsoft91: Thanks! Didn't know about it

launcher/modManager/cmodlist.cpp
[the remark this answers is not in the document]
kambala-decapitator: huh, that's interesting

launcher/modManager/cmodlist.cpp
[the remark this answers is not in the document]
kambala-decapitator: if you care about such things everywhere, it'll make your code better in general ;)

lib/CModHandler.cpp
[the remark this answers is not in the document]
kambala-decapitator: if you need to work with the passed param, copy it inside the method body. This is implementation detail, API user shouldn't care about it. All he knows is that the string passed into the method will stay as is, hence it should be passed by const ref (as string size is larger than a pointer).

lib/CModHandler.cpp
kambala-decapitator: operators are placed in the beginning of a line mainly to make it easier to see which operator is used, which is especially convenient when lines are long. Not insisting on changing.

> As for indends, GitHub uses 8 spaces as one tab, instead of 4

use mouse to see that lines after `return` are indented by 2 tabs instead of 1 ;)

lib/rmg/ObstaclePlacer.cpp
kambala-decapitator: if type is `mountain`, you'll make a useless comparison with `lake`

> better to avoid usage of else

never heard of such guideline tbh :)

launcher/modManager/cmodlist.cpp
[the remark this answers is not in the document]
Nordsoft91: Disagree, that having extra variable makes code better. Better to have an understanding what is the purpose of variable. Here the purpose you suggested is performance-wise. My point is such micro-optimizations are not needed, so extra variable is just an extra line + extra custom name

lib/CModHandler.cpp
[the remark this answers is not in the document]
Nordsoft91: There is no difference between passing by const reference + internal copy and passing by value for user. For implementation too. Such approach of passing by value and modifying it internally doesn't have any disadvantages.

lib/CModHandler.cpp
Nordsoft91: > beginning of a line mainly to make it easier to see which operator is used

For me it's opposite:)

kambala-decapitator (approved):

launcher/modManager/cmodlist.cpp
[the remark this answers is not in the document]
nullkiller: keyNormalize will contain the last key from input.toMap() matching its lowercased version so still depends on keys order. Thus I do not understand exactly how it normalizes things? Is it theoretically possible to have opposite situation to the current issue in order to fix which we will need the first value instead of the last one?

launcher/modManager/cmodlist.cpp
```diff
@@ -297,12 +340,12 @@ QVector<QString> CModList::getModList() const
 	{
 		for(auto it = repo.begin(); it != repo.end(); it++)
 		{
-			knownMods.insert(it.key());
+			knownMods.insert(it.key().toLower());
```
nullkiller, line 343: Looks like this construction is repeated a few times and its goal is to get lowered keys
Probably something like vstd::concatenate(knownMods, getLoweredKeys(repo)) might be possible

nullkiller: And above keyNormalize = getLoweredKeys(input.toMap())

lib/CModHandler.cpp
nullkiller: Regarding indentation our code style clearly insist on using tabs not spaces

launcher/modManager/cmodlist.cpp
Nordsoft91: those are QStrings and it's needed to use QT methods to work with them

launcher/modManager/cmodlist.cpp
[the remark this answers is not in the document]
Nordsoft91: They are used to be the same. The problem I'm trying to solve it that we have same mods with different name (Forge remote and forge locally). Theoretically it's possible to have two different mods remotely but even this situation should be covered (I don't remember the place, but somewhere with version overriding)

lib/CModHandler.cpp
Nordsoft91: there are tabs

PR #930, opened 2022-09-17
[macOS] save logs to the standard Logs directory

### LINKED ISSUE #875: [macOS] save logs to the standard Logs directory by default
Currently logs are written to the user data directory `~/Library/Application Support/vcmi`. Standard location for logs is `~/Library/Logs`.

kambala-decapitator:
Now logs will be stored in `~/Library/Logs/vcmi`.

closes #875

### DIFF
--- a/lib/VCMIDirs.cpp
+++ b/lib/VCMIDirs.cpp
@@ -13,6 +13,8 @@
 
 namespace bfs = boost::filesystem;
 
+bfs::path IVCMIDirs::userLogsPath() const { return userCachePath(); }
+
 bfs::path IVCMIDirs::userSavePath() const { return userDataPath() / "Saves"; }
 
 bfs::path IVCMIDirs::fullLibraryPath(const std::string &desiredFolder, const std::string &baseLibName) const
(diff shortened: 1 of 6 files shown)


lib/VCMIDirs.cpp
```diff
@@ -458,6 +438,14 @@ bfs::path VCMIDirsOSX::userDataPath() const
 bfs::path VCMIDirsOSX::userCachePath() const { return userDataPath(); }
 bfs::path VCMIDirsOSX::userConfigPath() const { return userDataPath() / "config"; }
 
+bfs::path VCMIDirsOSX::userLogsPath() const
+{
+	// TODO: use proper objc code from Foundation framework
+	if(const auto homeDir = std::getenv("HOME"))
+		return bfs::path{homeDir} / "Library" / "Logs" / "vcmi";
```
Nordsoft91, line 445: It make sense for linux/ios/android, but what the reason to change path for mac? Game data anyway is in application support and game creates artifacts such as downloads, settings. I suggest do not change log path for mac

kambala-decapitator: There're standard OS-specific things that should be respected. Logs directory exists for a reason.

PR #944, opened 2022-09-19
Add right click to Quick Recruit Window

### LINKED ISSUE #937: right clicking a creature doesn't work in the Quick Recruit Dialog

AgostonSzepessy:
Adds clickable area on top of creature picture. The clickable area is just a class that wraps an instance of `CIntObject` and creates an area that users can right click to view information about the creature they're buying. The constants were based off the existing ones in the code and were refined by guessing and checking where clicking falls inside/outside the border of the rectangle.

Fixes #937

### DIFF
--- a/client/gui/CIntObject.h
+++ b/client/gui/CIntObject.h
@@ -140,6 +140,7 @@ public:
 	//double click
 	virtual void onDoubleClick(){}
 
+	// These are the arguments that can be used to determine what kind of input the CIntObject will receive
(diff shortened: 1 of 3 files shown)


kambala-decapitator (approved):
looks good to me, but I'm not an expert in the engine internals to see if there's a better way to implement

Nordsoft91:
Looks like CCreaturePic is designed exactly for these purposes. It derives CIntObject class which has currentMouseState map so you can check right click currentMouseState[EIntObjMouseBtnType::RIGHT] and also it contains pointer to CCreature.
Is it possible to modify CCreaturePic instead of creating new class?

kambala-decapitator:
@AgostonSzepessy how is it supposed to work? I can't get info window to appear via right click on a creature in the Quick Recruit dialog. Checked in the latest develop as well as in the build produced by this PR.

Issue #919, opened 2022-09-16
Crash on campaign advance


nullkiller:
Start any campaign (for instance SoD Specter of Power)
Finish current scenario with vcmisilmaril
Observe server crash

>	VCMI_lib.dll!CGObjectInstance::getWidth() Line 155	C++
 	VCMI_lib.dll!CMap::addBlockVisTiles(CGObjectInstance * obj) Line 300	C++
 	VCMI_lib.dll!CGameState::replaceHeroesPlaceholders(const std::vector<CGameState::CampaignHeroReplacement,std::allocator<CGameState::CampaignHeroReplacement>> & campaignHeroReplacements) Line 2874	C++
 	VCMI_lib.dll!CGameState::placeCampaignHeroes() Line 1100	C++
 	VCMI_lib.dll!CGameState::init(const IMapService * mapService, StartInfo * si, bool allowSavingRandomMap) Line 753	

replaceHeroesPlaceholders
    map->addBlockVisTiles(heroToPlace);
          CGObjectInstance::getWidth()
               appearence is empty, crash

Issue #897, opened 2022-09-14
[REGRESSION] Mines from mods are placed incorrectly


Nordsoft91:
Generate random map with hota or forge mod (beta branch)
Observe mines on "classic" terrains

Not reproducible on 1.0.0

<img width="154" alt="Снимок экрана 2022-09-14 в 15 19 25" src="https://user-images.githubusercontent.com/9308612/190142360-c0951eba-191f-4db3-ab47-4705ae95211e.png">
<img width="164" alt="Снимок экрана 2022-09-14 в 15 18 40" src="https://user-images.githubusercontent.com/9308612/190142370-276ba75f-718a-4fcc-abe5-9497fc4eb00d.png">
<img width="215" alt="Снимок экрана 2022-09-14 в 15 18 37" src="https://user-images.githubusercontent.com/9308612/190142372-c8531a00-96a5-4c7b-b486-42092387aa19.png">
<img width="265" alt="Снимок экрана 2022-09-14 в 15 18 32" src="https://user-images.githubusercontent.com/9308612/190142377-ad21effd-cdad-4baa-9bc4-7581672cef1a.png">
<img width="127" alt="Снимок экрана 2022-09-14 в 15 12 22" src="https://user-images.githubusercontent.com/9308612/190142378-795b9258-f684-4bf1-b0e3-35b5bbeb65ee.png">
![Буфер обмена](https://user-images.githubusercontent.com/9308612/190142401-432843fc-8d8a-4df4-9bdc-08dfd2468784.png)

DjWarmonger:
Actually it's pretty simple. ZonePainter depends on  TownPlacer, however TownPlacer also places mines (!) which already defaulted to grass terain.

Not sure why it worked before, but definitelly current design is illogical.

DjWarmonger:
Actually, new mines spawn incorrectly also on Highlands and Wasteland. Always the one and the same template shows.

Nordsoft91:
@DjWarmonger , that's not the case. TownPlacer doesn't place mines, but just inserting them into the list of required objects. Mines are placed later in ObjectManager processing.

Obviously, we what we see is not grass-templates

PR #948, opened 2022-09-19
[897] Fix for mines templates

Nordsoft91:
That's probably not the best solution for https://github.com/vcmi/vcmi/issues/897, but I've created fix in order to unblock our modders

Taking into account that we have https://github.com/vcmi/vcmi/issues/919, it makes sense considering revert template changes

### DIFF
--- a/lib/mapObjects/CObjectClassesHandler.cpp
+++ b/lib/mapObjects/CObjectClassesHandler.cpp
@@ -117,11 +117,11 @@ std::vector<JsonNode> CObjectClassesHandler::loadLegacyData(size_t dataSize)
 
 	for (size_t i=0; i<totalNumber; i++)
 	{
-		auto templ = new ObjectTemplate;
+		std::shared_ptr<ObjectTemplate> templ(new ObjectTemplate);
 		templ->readTxt(parser);
 		parser.endLine();
 		std::pair<si32, si32> key(templ->id.num, templ->subid);
-		legacyTemplates.insert(std::make_pair(key, std::shared_ptr<const ObjectTemplate>(templ)));
+		legacyTemplates.insert(std::make_pair(key, templ));
 	}
 
 	std::vector<JsonNode> ret(dataSize);// create storage for 256 objects
@@ -550,12 +550,8 @@ SObjectSounds AObjectTypeHandler::getSounds() const
 	return sounds;
 }
 
-void AObjectTypeHandler::addTemplate(std::shared_ptr<const ObjectTemplate> templ)
+void AObjectTypeHandler::addTemplate(std::shared_ptr<ObjectTemplate> templ)
 {
-	//Otherwise the template remains constant
-	auto ptr = const_cast<ObjectTemplate*>(templ.get());
-	ptr->id = Obj(type);
-	ptr->subid = subtype;
 	templates.push_back(templ);
 }
 
--- a/lib/mapObjects/CObjectClassesHandler.cpp
+++ b/lib/mapObjects/CObjectClassesHandler.cpp
@@ -115,13 +115,15 @@ std::vector<JsonNode> CObjectClassesHandler::loadLegacyData(size_t dataSize)
 	size_t totalNumber = static_cast<size_t>(parser.readNumber()); // first line contains number of objects to read and nothing else
 	parser.endLine();
 
-	for (size_t i=0; i<totalNumber; i++)
+	for (size_t i = 0; i < totalNumber; i++)
 	{
-		std::shared_ptr<ObjectTemplate> templ(new ObjectTemplate);
-		templ->readTxt(parser);
+		auto tmpl = new ObjectTemplate;
+
+		tmpl->readTxt(parser);
 		parser.endLine();
-		std::pair<si32, si32> key(templ->id.num, templ->subid);
-		legacyTemplates.insert(std::make_pair(key, templ));
+
+		std::pair<si32, si32> key(tmpl->id.num, tmpl->subid);
+		legacyTemplates.insert(std::make_pair(key, std::shared_ptr<const ObjectTemplate>(tmpl)));
 	}
 
 	std::vector<JsonNode> ret(dataSize);// create storage for 256 objects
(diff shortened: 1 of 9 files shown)


lib/mapObjects/CommonConstructors.h
```diff
@@ -30,21 +30,7 @@ class CDefaultObjectTypeHandler : public AObjectTypeHandler
 	{
 		auto obj = new ObjectType();
 		preInitObject(obj);
-
-		if (tmpl)
```
nullkiller, line 34: We still can test shared_ptr for being empty. Or is such case not possible?

Nordsoft91: that's possible but valid

lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -550,12 +550,8 @@ SObjectSounds AObjectTypeHandler::getSounds() const
 	return sounds;
 }
 
-void AObjectTypeHandler::addTemplate(std::shared_ptr<const ObjectTemplate> templ)
+void AObjectTypeHandler::addTemplate(std::shared_ptr<ObjectTemplate> templ)
 {
-	//Otherwise the template remains constant
-	auto ptr = const_cast<ObjectTemplate*>(templ.get());
```
nullkiller, line 556: Wow.  This does not look correct. I agree.

nullkiller:
@DjWarmonger this PR is for you and it contains some investigations. We count on you to fix both issues or even better to refactor RMG a bit. I suggest to avoid cases when objects are created without template and consequent if(!appearance) checks. Controlling logic though nulls and empty values is treated as a bad design and can cause errors later because people tend to forget about such checks. As stated by @Nordsoft91 these issues block modders and we must fix them ASAP. At least the one with mines because modders want to see new mines but RMG puts original ones instead.
I personally do not see this fix being too bad so probably we can accept it. But it shows how fragile this logic with empty templates is. And I do not want to rollback anything from develop as @Nordsoft91 suggested)

DjWarmonger:
Why not `const` pointer for template? We don't expect templates to ever change once they are loaded.

nullkiller:
I agree but this needs refactoring. Creating a shared_ptr from a pointer extracted from another shared_ptr is defenitely not the way it should work. Each shared_ptr should have its own instance of type in memory.

DjWarmonger:
> @ even better to refactor RMG a bit. I suggest to avoid cases when objects are created without template and consequent if(!appearance) checks. Controlling logic though nulls and empty values is treated as a bad design and can cause errors later because people tend to forget about such checks.

Every object should be given suitable template in `ObjectManager::placeObject()` calling `object.finalize()`, so no invalid object should ever end up of the map. There's one method that ensures it.

nullkiller:
And as I stated it is not a fix, it is an investigation forced by a request from modders that the issue with templates blocks them. We just share our findings with you this way, @DjWarmonger.

nullkiller:
As far as I understood one logic sets a template to the first one in list and another expects it to be empty and should set it according to terrain but does nothing because some template is already specified. This does not look nice. Would be better if object is created with correct template at once. So I do not understand why exactly it got broken.

nullkiller:
`else
		{
			auto templates = getTemplates();
			if (templates.empty())
			{
				throw std::runtime_error("No handler for created object");
			}
			obj->appearance = templates.front(); //just any template for now, will be initialized later
		}`
		
		Now clear this section is from your PR about shared_ptr for ObjectTemplate.

Nordsoft91:
Implemented alternative solution

lib/mapping/CMap.cpp
```diff
@@ -424,7 +424,7 @@ bool CMap::checkForVisitableDir(const int3 & src, const TerrainTile *pom, const
 		if(!vstd::contains(pom->blockingObjects, obj)) //this visitable object is not blocking, ignore
 			continue;
 
-		if (!obj->appearance->isVisitableFrom(src.x - dst.x, src.y - dst.y))
+		if(!obj->appearance || !obj->appearance->isVisitableFrom(src.x - dst.x, src.y - dst.y))
```
nullkiller, line 427: The question is left if we need this or could we omit calling checkForVisitableDir if there is no template specified?

lib/rmg/RmgObject.cpp
```diff
@@ -105,14 +105,11 @@ void Object::Instance::setPositionRaw(const int3 & position)
 
 void Object::Instance::setTemplate(const Terrain & terrain)
 {
-	if(dObject.appearance->id == Obj::NO_OBJ)
-	{
-		auto templates = VLC->objtypeh->getHandlerFor(dObject.ID, dObject.subID)->getTemplates(terrain);
-		if(templates.empty())
-			throw rmgException(boost::to_string(boost::format("Did not find graphics for object (%d,%d) at %s") % dObject.ID % dObject.subID % static_cast<std::string>(terrain)));
+	auto templates = VLC->objtypeh->getHandlerFor(dObject.ID, dObject.subID)->getTemplates(terrain);
```
nullkiller, line 108: Non-conditional set of template I like better!

lib/mapping/CMap.cpp
Nordsoft91: It can be reverted actually, because obj->appearances used in other methods too.
So either we should check for nullptr everywhere or can avoid such changes

Nordsoft91:
Such minor changes I believe is a good enough to be considered as a fix

DjWarmonger:
Works well for me.

> As far as I understood one logic sets a template to the first one in list and another expects it to be empty and should set it according to terrain but does nothing because some template is already specified. This does not look nice. Would be better if object is created with correct template at once. 

IMO as it is now, RMG objects should be always created with null template and only get it when they are finally placed at correct terrain.
So quite the opposite, I may remove object template argument from `create()` method as it's never used anyway. And I don't see why a developer would need to set arbitrary templates somewhere in RMG.

DjWarmonger:
Oh come on, I just restored `const ObjectTemplate`. It was completely unneccessary to juggle any constness in first place.

lib/mapObjects/CommonConstructors.h
DjWarmonger: Okay, so now epty pointer will be valid and default, but can be set explicitely. This is needed in some cases.

DjWarmonger (approved):

Nordsoft91:
@DjWarmonger you methods in binary serialiser and deserialiser are PURE copy-paste of 50+ lines of code

Please, get rid of copy-paste methods and in your own PR please

PR #956, opened 2022-09-20
[CI] switch develop to nightly builds

kambala-decapitator:
- from now on we'll have a single nightly build (at 2 AM UTC) from the latest develop instead of building it on each push
- also possible to trigger build manually now
- if a successful build already exists (e.g. already built manually or ran in the past), current run is cancelled, by default an email will be sent about this

Tested the approach in a dummy workflow: https://github.com/kambala-decapitator/workflow-test/actions/workflows/time-based.yml

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -1,11 +1,11 @@
 name: VCMI
 
 on:
-    push:
-      branches:
-        - features/*
-        - develop
-    pull_request:
+  push:
+    branches:
+      - features/*
+      - develop
+  pull_request:
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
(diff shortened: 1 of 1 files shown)


.github/workflows/github.yml
```diff
@@ -1,17 +1,68 @@
...
-        - features/*
-        - develop
-    pull_request:
+  push:
+    branches:
+      - features/*
+  pull_request:
+  schedule:
+    - cron: '0 2 * * *'
+  workflow_dispatch:
```
nullkiller, line 10: Why do we need dispatch here?

kambala-decapitator: > also possible to trigger build manually now

nullkiller: I understand. It is kind of generic question. Who is going to start it manually and why. Ok, let it be.

kambala-decapitator: in case a build is needed asap without waiting till night

PR #959, opened 2022-09-21
fix cursor image on hover when there're multiple creatures nearby

### LINKED ISSUE #800: cursor shows "pick object" instead of "attack"
My steps were:

1. enter monolith, hero has no more turns, plan to attack Basilisks
2. end turn
3. hover Basilisks - cursor shows "pick object" instead of "attack"

Screenshot 2022-09-03 at 11 48 06
Screenshot 2022-09-03 at 11 48 28
Screenshot 2022-09-03 at 11 48 54

kambala-decapitator:
fixes #800, you can find investigation details there

### DIFF
--- a/client/windows/CAdvmapInterface.cpp
+++ b/client/windows/CAdvmapInterface.cpp
@@ -1711,8 +1711,7 @@ void CAdvMapInt::tileHovered(const int3 &mapPos)
 	}
 	else if(const CGHeroInstance * h = curHero())
 	{
-		int3 mapPosCopy = mapPos;
-		const CGPathNode * pnode = LOCPLINT->cb->getPathsInfo(h)->getPathInfo(mapPosCopy);
+		const CGPathNode * pnode = LOCPLINT->cb->getPathsInfo(h)->getPathInfo(mapPos);
 		assert(pnode);
 
 		int turns = pnode->turns;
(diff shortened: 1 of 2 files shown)


client/windows/CAdvmapInterface.cpp
```diff
@@ -1711,8 +1711,7 @@ void CAdvMapInt::tileHovered(const int3 &mapPos)
 	}
 	else if(const CGHeroInstance * h = curHero())
 	{
-		int3 mapPosCopy = mapPos;
-		const CGPathNode * pnode = LOCPLINT->cb->getPathsInfo(h)->getPathInfo(mapPosCopy);
+		const CGPathNode * pnode = LOCPLINT->cb->getPathsInfo(h)->getPathInfo(mapPos);
```
Nordsoft91, line 1714: does getPathInfo receives const int3 &?

lib/CPathfinder.cpp
```diff
@@ -862,7 +862,7 @@ void DestinationActionRule::process(
 				else
 					action = CGPathNode::BLOCKING_VISIT;
 			}
-			else if(destination.isGuardianTile)
+			else if(destination.guarded)
```
Nordsoft91, line 865: guarded will return true for neighbour tiles too?

client/windows/CAdvmapInterface.cpp
kambala-decapitator: yes, otherwise it wouldn't have compiled

lib/CPathfinder.cpp
kambala-decapitator: yes, as far as I understood. You can also see tested cases in #800

Nordsoft91 (approved):

PR #970, opened 2022-09-22
Add simple server healthcheck

Nordsoft91:


client/CServerHandler.cpp
```diff
@@ -161,6 +161,18 @@ void CServerHandler::startLocalServerAndConnect()
 		threadRunLocalServer->join();
 
 	th->update();
+	
+	try
+	{
+		CConnection testConnection(settings["server"]["server"].String(), getDefaultPort(), NAME, uuid);
+		logNetwork->error("Port is busy, kill other vcmiserver processes");
```
nullkiller, line 168: what if we connect and send special kill package so that server gets shut down?

Nordsoft91: we need somehow to check if the server was started by this client. Otherwise we can kill remote server

client/CServerHandler.cpp
```diff
@@ -161,6 +162,19 @@ void CServerHandler::startLocalServerAndConnect()
...
+		CInfoWindow::showInfoDialog("Port is busy, kill other vcmiserver processes", CInfoWindow::TCompsInfo(), PlayerColor(1));
```
Nordsoft91, line 170: would appreciate for ideas to make text more user-friendly

kambala-decapitator: - please move the hardcoded string into `config/translate.json`
- maybe something like: Another vcmiserver process is running, please terminate it first
- code improvement:
```suggestion
		CInfoWindow::showInfoDialog("Port is busy, kill other vcmiserver processes", {});
```

nullkiller: Can we ask user if they want to terminate it and do it for them automatically? I think this way it will be the best. So I can try to do this myself. Is it possible to kill process via boost or whatever?

kambala-decapitator: > Is it possible to kill process via boost or whatever?

probably yes, but I'm pretty sure we'll need to pull in new boost module

nullkiller: Or we can consider using single-process by default. This I like even more

kambala-decapitator (approved):

kambala-decapitator:
now I see a disadvantage of this change: I can no longer run "dedicated" server. Use case: debugging the server, launching client manually.

Needs some check if server process is able to start new game or it's in a broken state.

PR #971, opened 2022-09-22
Fix most project warnings

kambala-decapitator:
Warnings as reported by Apple clang 13.

Some notes:

- I decided not to touch the deprecation warning of `std::random_shuffle` in `AI/Nullkiller/Pathfinding/AINodeStorage.cpp` as I'm not quite sure of the correct replacement (might have side effects, needs testing). The fix itself is to use variant 3 from https://en.cppreference.com/w/cpp/algorithm/random_shuffle e.g. like in the example in the bottom of the page.
- the FL's varargs warning is ok to silence as there's no UB there, see https://github.com/fuzzylite/fuzzylite/issues/77
- after the "fix operator precedence" change I expected to see some difference in the UI, but strangely nothing has changed... See screenshots: first is "before", second is "after".

Screenshot 2022-09-22 at 10 38 12 Screenshot 2022-09-22 at 10 42 59

Screenshot 2022-09-22 at 10 39 08 Screenshot 2022-09-22 at 10 41 04

### DIFF
--- a/AI/Nullkiller/Pathfinding/Actors.cpp
+++ b/AI/Nullkiller/Pathfinding/Actors.cpp
@@ -269,8 +269,6 @@ ExchangeResult HeroExchangeMap::tryExchangeNoLock(const ChainActor * other)
 			return result; // already inserted
 		}
 
-		auto position = inserted.first;
-
 		auto differentMasks = (actor->chainMask & other->chainMask) == 0;
 
 		if(!differentMasks) return result;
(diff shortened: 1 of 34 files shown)


AI/Nullkiller/Pathfinding/Actors.cpp
```diff
@@ -461,15 +459,6 @@ CCreatureSet * DwellingActor::getDwellingCreatures(const CGDwelling * dwelling,
 			continue;
 
 		auto creature = creatureInfo.second.back().toCreature();
-		auto count = creatureInfo.first;
-
-		if(waitForGrowth)
```
nullkiller, line 466: Probably it was discarded at certain stage because of complexity. At the very beginning hero chain was allowed to by army from dwellings. Now only from town.

lib/rmg/Zone.cpp
```diff
@@ -193,11 +193,7 @@ void Zone::fractalize()
 	rmg::Area clearedTiles(dAreaFree);
 	rmg::Area possibleTiles(dAreaPossible);
 	rmg::Area tilesToIgnore; //will be erased in this iteration
-	
-	//the more treasure density, the greater distance between paths. Scaling is experimental.
-	int totalDensity = 0;
```
nullkiller, line 198: Lets wait for @Nordsoft91 if this was intended

nullkiller (approved):

lib/rmg/Zone.cpp
Nordsoft91: It inherited from previous rmg, can be removed safely

kambala-decapitator:
regular merge please

PR #972, opened 2022-09-22
Cpp map editor

Nordsoft91:
Task tracker https://github.com/Nordsoft91/vcmi/issues
Documentation https://github.com/Nordsoft91/vcmi/wiki


config/schemas/settings.json
```diff
@@ -388,7 +388,7 @@
 				},
 				"updateConfigUrl" : {
 					"type" : "string",
-					"default" : "https://raw.githubusercontent.com/vcmi/vcmi-updates/master/vcmi-updates.json"
+					"default" : "https://raw.githubusercontent.com/Nordsoft91/vcmi-autoupdate/main/autoUpdate.json"
```
nullkiller, line 391: Lets fix this

config/schemas/terrain.json
```diff
@@ -0,0 +1,109 @@
+{
```
nullkiller, line 1: Have not got any schema for this so far? That is strange.

mapeditor/Animation.cpp
```diff
@@ -0,0 +1,780 @@
+/*
+ * CAnimation.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#include "StdInc.h"
```
nullkiller, line 10: Is it a new file? Don't we have it for client already? Or is it same animation but slightly modified for QT?

mapeditor/CGameInfo.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+
+void CGameInfo::updateEntity(Metatype metatype, int32_t index, const JsonNode & data)
+{
+	logGlobal->error("CGameInfo::updateEntity call is not expected.");
+}
+
+spells::effects::Registry * CGameInfo::spellEffects()
+{
+	return nullptr;
```
nullkiller, line 104: If it is never called lets throw an exception that thing is not implemented. Can't we return globalServices->spellEffects()? const cast is required?

mapeditor/generatorprogress.cpp
```diff
@@ -0,0 +1,38 @@
+#include "StdInc.h"
+#include "generatorprogress.h"
+#include "ui_generatorprogress.h"
+#include <thread>
+#include <chrono>
+
+GeneratorProgress::GeneratorProgress(Load::Progress & source, QWidget *parent) :
+	QDialog(parent),
+	ui(new Ui::GeneratorProgress),
```
nullkiller, line 9: Can we use std::unique_ptr?

mapeditor/graphics.cpp
```diff
@@ -0,0 +1,378 @@
...
+	{
+		playerColors[i] = colors[i].rgba();
+	}
+	//gray
+	neutralColor = qRgba(0x84, 0x84, 0x84, 0xFF);
+}
+
+Graphics::Graphics()
+{
+#if 0
```
nullkiller, line 97: One more copied object? Need to do something about it later.

mapeditor/inspector/armywidget.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+
+ArmyWidget::ArmyWidget(CArmedInstance & a, QWidget *parent) :
+	QDialog(parent),
+	army(a),
+	ui(new Ui::ArmyWidget)
+{
+	ui->setupUi(this);
+	
+	uiCounts[0] = ui->count0; uiSlots[0] = ui->slot0;
```
nullkiller, line 13: No way to render it with loop?

mapeditor/inspector/armywidget.cpp
```diff
@@ -0,0 +1,142 @@
...
+	uiCounts[1] = ui->count1; uiSlots[1] = ui->slot1;
+	uiCounts[2] = ui->count2; uiSlots[2] = ui->slot2;
+	uiCounts[3] = ui->count3; uiSlots[3] = ui->slot3;
+	uiCounts[4] = ui->count4; uiSlots[4] = ui->slot4;
+	uiCounts[5] = ui->count5; uiSlots[5] = ui->slot5;
+	uiCounts[6] = ui->count6; uiSlots[6] = ui->slot6;
+	
+	for(int i = 0; i < TOTAL_SLOTS; ++i)
+	{
+		uiCounts[i]->setInputMask("d0000");
```
nullkiller, line 23: what d0000 mean? 4 digits?

mapeditor/inspector/inspector.cpp
```diff
@@ -0,0 +1,734 @@
...
+{
+	if(s == "TRUE")
+		return true;
+	//if(s == "FALSE")
+	return false;
+}
+
+void Initializer::initialize(CArmedInstance * o)
+{
+	if(!o) return;
```
nullkiller, line 45: No logic yet?

mapeditor/inspector/inspector.cpp
```diff
@@ -0,0 +1,734 @@
...
+	
+	o->tempOwner = defaultPlayer;
+	
+	switch(o->ID)
+	{
+		case Obj::RANDOM_DWELLING:
+		case Obj::RANDOM_DWELLING_LVL:
+		case Obj::RANDOM_DWELLING_FACTION:
+			o->initRandomObjectInfo();
```
nullkiller, line 72: Will it make some random initialization? The thing is usually displayed as big green circle with text. Also random monsters exist

mapeditor/inspector/inspector.cpp
```diff
@@ -0,0 +1,734 @@
...
+	}
+	
+	if(!o->type)
+		o->type = VLC->heroh->objects.at(o->subID);
+	
+	o->name = o->type->getName();
+	o->sex = o->type->sex;
+	o->biography = o->type->biography;
+	o->portrait = o->type->imageIndex;
+	o->randomizeArmy(o->type->heroClass->faction);
```
nullkiller, line 122: Afaik hero army can change as well upon map start. Will it be implemented later?

mapeditor/inspector/inspector.cpp
```diff
@@ -0,0 +1,734 @@
...
+	addProperty("Message", o->message, false);
+	{
+		auto * delegate = new InspectorDelegate;
+		delegate->options << "COMPLIANT" << "FRIENDLY" << "AGRESSIVE" << "HOSTILE" << "SAVAGE";
+		addProperty<CGCreature::Character>("Character", (CGCreature::Character)o->character, delegate, false);
+	}
+	addProperty("Never flees", o->neverFlees, InspectorDelegate::boolDelegate(), false);
+	addProperty("Not growing", o->notGrowingTeam, InspectorDelegate::boolDelegate(), false);
+	addProperty("Artifact reward", o->gainedArtifact); //TODO: implement in setProperty
+	addProperty("Army", PropertyEditorPlaceholder(), true);
```
nullkiller, line 319: Is it single creature or do we allow mixed army here?

mapeditor/inspector/inspector.cpp
```diff
@@ -0,0 +1,734 @@
...
+	
+	if(key == "Name")
+		o->name = value.toString().toStdString();
+	
+	if(key == "Hero type")
+	{
+		for(auto t : VLC->heroh->objects)
+		{
+			if(t->getName() == value.toString().toStdString())
+				o->type = t.get();
```
nullkiller, line 522: Will it take last? Does it check if hero is already added on map?

mapeditor/inspector/townbulidingswidget.cpp
```diff
@@ -0,0 +1,243 @@
...
+		case BuildingID::DWELL_LVL_7_UP: return "DWELL_LVL_7_UP";
+		default:
+			return "UNKNOWN";
+	}
+}
+
+TownBulidingsWidget::TownBulidingsWidget(CGTownInstance & t, QWidget *parent) :
+	town(t),
+	QDialog(parent),
+	ui(new Ui::TownBulidingsWidget)
```
nullkiller, line 63: Here we also can use unique_ptr so while you remember to dispose them it is ok

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+	ui->objectPreview->setScene(scenePreview);
+
+	//scenes[0]->addPixmap(QPixmap(QString::fromStdString(resPath.native())));
+
+	//loading objects
+	loadObjectsTree();
+	
+	ui->tabWidget->setCurrentIndex(0);
+	
+	for(int i = 0; i < 8; ++i)
```
nullkiller, line 166: I think it should be possible to grab some constants from VCMI_Lib like number of player colors etc

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+	addGroupIntoCatalog("QUESTS", true, false, Obj::SEER_HUT);
+	addGroupIntoCatalog("QUESTS", true, false, Obj::BORDER_GATE);
+	addGroupIntoCatalog("QUESTS", true, false, Obj::QUEST_GUARD);
+	addGroupIntoCatalog("QUESTS", true, false, Obj::HUT_OF_MAGI);
+	addGroupIntoCatalog("QUESTS", true, false, Obj::EYE_OF_MAGI);
+	addGroupIntoCatalog("QUESTS", true, false, Obj::BORDERGUARD);
+	addGroupIntoCatalog("QUESTS", true, false, Obj::KEYMASTER);
+	addGroupIntoCatalog("wog object", true, false, Obj::WOG_OBJECT);
+	addGroupIntoCatalog("OBSTACLES", true);
+	addGroupIntoCatalog("OTHER", false);
```
nullkiller, line 667: Consider adding objects by mod name

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+	addGroupIntoCatalog("DWELLINGS", true, false, Obj::CREATURE_GENERATOR1);
+	addGroupIntoCatalog("DWELLINGS", true, false, Obj::CREATURE_GENERATOR2);
+	addGroupIntoCatalog("DWELLINGS", true, false, Obj::CREATURE_GENERATOR3);
+	addGroupIntoCatalog("DWELLINGS", true, false, Obj::CREATURE_GENERATOR4);
+	addGroupIntoCatalog("DWELLINGS", true, false, Obj::REFUGEE_CAMP);
+	addGroupIntoCatalog("DWELLINGS", false, false, Obj::RANDOM_DWELLING);
+	addGroupIntoCatalog("DWELLINGS", false, false, Obj::RANDOM_DWELLING_LVL);
+	addGroupIntoCatalog("DWELLINGS", false, false, Obj::RANDOM_DWELLING_FACTION);
+	addGroupIntoCatalog("GROUNDS", true, false, Obj::CURSED_GROUND1);
+	addGroupIntoCatalog("GROUNDS", true, false, Obj::MAGIC_PLAINS1);
```
nullkiller, line 624: Here we can add all objects having terrain patch handler.

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+	}
+}
+void MainWindow::displayStatus(const QString& message, int timeout /* = 2000 */)
+{
+	statusBar()->showMessage(message, timeout);
+}
+
+void MainWindow::on_actionValidate_triggered()
+{
+	new Validator(controller.map(), this);
```
nullkiller, line 1033: Who will free it afterwards?

config/schemas/terrain.json
Nordsoft91: yes, missed that:(

mapeditor/Animation.cpp
Nordsoft91: a lot of duplicated code there to render graphics using QT instead of SDL. I tried do not touch that much

mapeditor/CGameInfo.cpp
Nordsoft91: it's a placeholder for things are not used. But it may be possible to use in future (like spell editor)

mapeditor/generatorprogress.cpp
Nordsoft91: no no, it's qt stuff

mapeditor/inspector/armywidget.cpp
Nordsoft91: I've spent some time to create macros but failed with that. It's possible to add widgets with code instead of creator but have doubts that it'll reduce amount of code

mapeditor/inspector/armywidget.cpp
Nordsoft91: it's needed to be removed actually

mapeditor/inspector/inspector.cpp
Nordsoft91: That's initialization of objects' properties - no properties to initialize in that case

mapeditor/inspector/inspector.cpp
Nordsoft91: yes, that's work

mapeditor/inspector/inspector.cpp
Nordsoft91: AFAIK it doesn't impact game itself because those are not serialized

mapeditor/inspector/inspector.cpp
Nordsoft91: It's about mixed army but this property is not modifiable for monsters

mapeditor/inspector/inspector.cpp
```diff
@@ -0,0 +1,734 @@
...
+{
+	if(!o) return;
+	
+	if("Remove after")
+		o->removeAfterVisit = stringToBool(value.toString());
+	
+	if("Human trigger")
+		o->humanActivate = stringToBool(value.toString());
+	
+	if("Cpu trigger")
```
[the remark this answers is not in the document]
Nordsoft91: good catch - a bug actually

mapeditor/inspector/inspector.cpp
Nordsoft91: no, I don't have such feature implemented, so you can have several heroes of the same type

mapeditor/inspector/townbulidingswidget.h
```diff
@@ -0,0 +1,53 @@
+#ifndef TOWNBULIDINGSWIDGET_H
```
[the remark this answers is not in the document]
Nordsoft91: ++

mapeditor/inspector/townbulidingswidget.cpp
Nordsoft91: no no, qt takes care

mapeditor/mainwindow.cpp
Nordsoft91: OTHER are not mod objects. It's remaining objects which are not under any of categories.
Actually, mod objects are distibuted across groups - you can filter them using special text field (like type "hota" to see objects from this particular mod)

mapeditor/mainwindow.cpp
Nordsoft91: make sense to consider opportunity to distinguish objects by handler instead of ID

mapeditor/mainwindow.cpp
Nordsoft91: qt objects are destroyed by themselves (on close on parent close)

mapeditor/mapcontroller.cpp
```diff
@@ -0,0 +1,511 @@
...
+	}
+	if(VLC->spellh->getDefaultAllowed().size() > map()->allowedSpell.size())
+	{
+		for(int i = map()->allowedSpell.size(); i < VLC->spellh->getDefaultAllowed().size(); ++i)
+			map()->allowedSpell.push_back(false);
+	}
+	if(VLC->heroh->getDefaultAllowed().size() > map()->allowedHeroes.size())
+	{
+		for(int i = map()->allowedHeroes.size(); i < VLC->heroh->getDefaultAllowed().size(); ++i)
+			map()->allowedHeroes.push_back(false);
```
nullkiller, line 91: Can't we just resize the vector? map()->allowedHeroes.resize(VLC->heroh->getDefaultAllowed().size())

mapeditor/mapcontroller.cpp
```diff
@@ -0,0 +1,511 @@
...
+		}
+		//fix flags
+		if(obj->getOwner() == PlayerColor::UNFLAGGABLE)
+		{
+			if(dynamic_cast<CGMine*>(obj.get()) ||
+			   dynamic_cast<CGDwelling*>(obj.get()) ||
+			   dynamic_cast<CGTownInstance*>(obj.get()) ||
+			   dynamic_cast<CGGarrison*>(obj.get()) ||
+			   dynamic_cast<CGShipyard*>(obj.get()) ||
+			   dynamic_cast<CGHeroInstance*>(obj.get()))
```
nullkiller, line 112: Lighthouse?

nullkiller: Interesting also if dark cloud is flaggable. Teoretically it has kind of owner because only one player can see through it

nullkiller: Also might be worth to introduce a method which can answer if it is flaggable or not

mapeditor/mapcontroller.cpp
```diff
@@ -0,0 +1,511 @@
...
+		_scenes[level]->selectionObjectsView.selectionMode = 0;
+		_scenes[level]->selectionObjectsView.draw();
+		return true;
+	}
+	return false;
+}
+
+void MapController::createObject(int level, CGObjectInstance * obj) const
+{
+	_scenes[level]->selectionObjectsView.newObject = obj;
```
nullkiller, line 332: Should we check it for nullptr and discard if not?

mapeditor/mapcontroller.cpp
```diff
@@ -0,0 +1,511 @@
...
+	int objCounter = 0;
+	for(auto o : _map->objects)
+	{
+		if(o->ID == newObj->ID && o->subID == newObj->subID)
+		{
+			++objCounter;
+		}
+	}
+	
+	if((rmgInfo.mapLimit && objCounter >= rmgInfo.mapLimit)
```
nullkiller, line 470: Why does it depend on RMG? Editor should be completely free of any limitations I think. Except 8 heroes of same color on map. And sure 1 grail. All the rest should work.

mapeditor/mapcontroller.cpp
```diff
@@ -0,0 +1,511 @@
...
+		return false; //maplimit reached
+	}
+	if(defaultPlayer == PlayerColor::NEUTRAL && (newObj->ID == Obj::HERO || newObj->ID == Obj::RANDOM_HERO))
+	{
+		error = "Hero cannot be created as NEUTRAL";
+		return false;
+	}
+	if(defaultPlayer != PlayerColor::NEUTRAL && newObj->ID == Obj::PRISON)
+	{
+		error = "Prison must be a NEUTRAL";
```
nullkiller, line 485: Can we just make it neutral by default? It should not depend on current player selection.

mapeditor/mapcontroller.cpp
```diff
@@ -0,0 +1,511 @@
...
+		return false;
+	}
+	
+	if(newObj->ID == Obj::ARTIFACT && !_map->allowedArtifact.at(newObj->subID))
```
nullkiller, line 489: Is it possible that map author wants to block artifact to be sure that only one instance of it can be placed on map because it is involved in some important quest?

nullkiller: There is workaround to put it first and block afterwards so. Quite logical I think.

nullkiller: So some prompt like It is blocked in map settings. Are you sure you want to put it anyway? Wwould be better anyway.

mapeditor/maphandler.cpp
```diff
@@ -0,0 +1,532 @@
...
+{
+	if(player == PlayerColor::UNFLAGGABLE)
+		return;
+	if(sur->format() == QImage::Format_Indexed8)
+	{
+		QRgb color = graphics->neutralColor;
+		if(player != PlayerColor::NEUTRAL && player < PlayerColor::PLAYER_LIMIT)
+			color = graphics->playerColors.at(player.getNum());
+
+		sur->setColor(5, color);
```
nullkiller, line 160: Only one color index? no shadows?

mapeditor/maphandler.cpp
```diff
@@ -0,0 +1,532 @@
...
+	if (!obj)
+	{
+		logGlobal->error("Stray map object that isn't fading");
+		return;
+	}
+
+	uint8_t animationFrame = 0;
+
+	auto objData = findObjectBitmap(obj, animationFrame, obj->ID == Obj::HERO ? 2 : 0);
+	std::vector<std::shared_ptr<QImage>> debugFlagImages;
```
nullkiller, line 385: where is it used?

mapeditor/mapview.cpp
```diff
@@ -0,0 +1,448 @@
...
+void MinimapView::setController(MapController * ctrl)
+{
+	controller = ctrl;
+}
+
+void MinimapView::mouseMoveEvent(QMouseEvent *mouseEvent)
+{
+	this->update();
+	
+	auto * sc = static_cast<MinimapScene*>(scene());
```
nullkiller, line 29: when sc can be nullptr? When there is no scene()?

mapeditor/mapview.cpp
```diff
@@ -0,0 +1,448 @@
...
+
+	auto pos = mapToScene(mouseEvent->pos()); //TODO: do we need to check size?
+	int3 tile(pos.x() / 32, pos.y() / 32, sc->level);
+
+	if(tile == tilePrev) //do not redraw
+		return;
+
+	tilePrev = tile;
+
+	//main->setStatusMessage(QString("x: %1 y: %2").arg(QString::number(pos.x()), QString::number(pos.y())));
```
nullkiller, line 84: Not needed? Lets remove

mapeditor/mapview.cpp
```diff
@@ -0,0 +1,448 @@
...
+			if(event->button() == Qt::RightButton)
+				sc->selectionTerrainView.erase(tileStart + e);
+			else if(event->button() == Qt::LeftButton)
+				sc->selectionTerrainView.select(tileStart + e);
+		}
+	}
+		sc->selectionTerrainView.draw();
+		break;
+
+	case MapView::SelectionTool::Brush4:
```
nullkiller, line 219: Looks like these switch/cases differ only with brush size

mapeditor/mapcontroller.cpp
Nordsoft91: won't it override existing value?

mapeditor/mapcontroller.cpp
Nordsoft91: Or extra interface to inherit from

mapeditor/mapcontroller.cpp
Nordsoft91: Ideally yes but in this case those are initialized in constructor (need to use pointers to accomodate QObject requirements)

mapeditor/mapcontroller.cpp
Nordsoft91: agree

mapeditor/mapcontroller.cpp
Nordsoft91: doable

mapeditor/maphandler.cpp
Nordsoft91: for flags? I'm sure it's the one color

mapeditor/maphandler.cpp
Nordsoft91: nowhere, need to cleanup

mapeditor/mapview.cpp
Nordsoft91: yes, on application start

mapeditor/mapview.cpp
Nordsoft91: I'd restore - it's the one way to know coordinates

mapeditor/mapview.cpp
Nordsoft91: do you suggest short lambda?

mapeditor/mapview.h
```diff
@@ -0,0 +1,130 @@
...
+public:
+	MapSceneBase(int lvl);
+	
+	const int level;
+	
+	virtual void updateViews();
+	virtual void initialize(MapController &);
+	
+protected:
+	virtual std::list<AbstractLayer *> getAbstractLayers() = 0;
```
AgostonSzepessy, line 25: Why use `std::list`? It looks like you're just iterating through the list, and not removing items from the middle of it. `std::vector` would offer better performance here; better cache locality.

mapeditor/playersettings.cpp
```diff
@@ -0,0 +1,77 @@
...
+	{
+		if(p.canAnyonePlay())
+		{
+			paramWidgets.push_back(new PlayerParams(controller, players));
+			ui->playersLayout->addWidget(paramWidgets.back());
+			++players;
+		}
+	}
+
+	if(players < 2)
```
AgostonSzepessy, line 26: Could you please replace `2` with a named constant?

mapeditor/validator.h
```diff
@@ -0,0 +1,30 @@
...
+		bool critical;
+		
+		Issue(const QString & m, bool c): message(m), critical(c) {}
+	};
+	
+public:
+	explicit Validator(const CMap * map, QWidget *parent = nullptr);
+	~Validator();
+	
+	static std::list<Issue> validate(const CMap * map);
```
AgostonSzepessy, line 26: Looks like this can be replaced with `std::vector` too.

mapeditor/graphics.h
```diff
@@ -0,0 +1,84 @@
...
+	
+public:
+	//various graphics
+	QVector<QRgb> playerColors; //array [8]
+	QRgb neutralColor;
+	QVector<QRgb> playerColorPalette; //palette to make interface colors good - array of size [256]
+	QVector<QRgb> neutralColorPalette;
+		
+	// [hero class def name]  //added group 10: up - left, 11 - left and 12 - left down // 13 - up-left standing; 14 - left standing; 15 - left down standing
+	std::map< std::string, std::shared_ptr<Animation> > heroAnimations;
```
AgostonSzepessy, line 58: I recommend replacing all `std::map<std::string, ...>` with `std::unordered_map<std::string, ...>` unless the order of the elements matter. `std::unordered_map` has faster lookup generally than `std::map` (amortized O(1) vs O(log n)).

mapeditor/mainwindow.cpp
AgostonSzepessy: This won't get deleted because it's not assigned an owner; it'll just be dangling.

Nordsoft91: @AgostonSzepessy , please get familiar with QT documentation

AgostonSzepessy: Hey @Nordsoft91, in the future, I would appreciate it if you were less passive-aggressive, but instead more polite and helpful. This comment wasn't really either of those.

AgostonSzepessy:
Looking at all the classes that have `.ui` files, could you please rename them so that's it's clear that they're either dialogs, or windows? It's not clear from looking just at the class names as to what they are. E.g., `PlayerSettings -> PlayerSettingsWindow`, etc.

mapeditor/mapcontroller.cpp
nullkiller: As far as I understood from docs it should not

mapeditor/mainwindow.cpp
Nordsoft91: I've already explained above that qt manages memory, you statement was just incorrect so I suggest to get details from docs (qt has very good docs actually). What else is expected? Sorry, nothing offensive anyway, I value any kind of feedback.

mapeditor/mapview.h
Nordsoft91: Actually, not quite, even about performance. Vector has O(NlnN) complexity for N sequential push backs, while list has O(N). But the reason I chose list was not about performance of course (such kind of performance improvements are completely useless here).
I use list if I want implicitly forbid random access for container. As soon as you have random access you have potential risk of access violation (out of range in the best case) so everything which might happen - happens soon or late:)

mapeditor/playersettings.cpp
Nordsoft91: could you please suggest the name?

mapeditor/validator.h
Nordsoft91: same comment. Implicit advise "iterate over it"

Nordsoft91:
> Looking at all the classes that have `.ui` files, could you please rename them so that's it's clear that they're either dialogs, or windows? It's not clear from looking just at the class names as to what they are. E.g., `PlayerSettings -> PlayerSettingsWindow`, etc.

That's make sense

mapeditor/mainwindow.cpp
nullkiller: I agree. There was nothing aggressive. @AgostonSzepessy you did not ask for any advanced clarification but instead insisted on @Nordsoft91 being wrong without any additional information regarding why you think so. Or is there something QT specific under **because it's not assigned an owner**? How should this be done?

AgostonSzepessy: @Nordsoft91 thanks for clarifying, it makes sense now. I just found your comment very abrupt and not very helpful. However, all is good and I have no ill feelings.

@nullkiller I did provide my reasoning, but it's okay now; water under the bridge.
This is something Qt specific. Anything that derives from a QObject (Validator derives from QDialog) will get cleaned up automatically by the Qt framework when the thing they're attached to also get deleted/removed. In this case, it'll get deleted when either Validator is closed or MainWindow is closed.

mapeditor/mapview.h
AgostonSzepessy: That makes sense. In that case, we need to use a std::list.

mapeditor/graphics.h
AgostonSzepessy: Oh I see. In that case, is there any way to better share the code between the two places rather than duplicating it?

mapeditor/playersettings.cpp
AgostonSzepessy: I'm not really sure. Is it 2 because there must be 2 minimum players? In that case, maybe `NUM_MIN_PLAYERS`? I can't fully tell though why 2 is used a lot in this file.

mapeditor/StdInc.h
```diff
@@ -0,0 +1,50 @@
...
+#include <QString>
+#include <QFile>
+
+using NumericPointer = unsigned long long;
+
+template<class Type>
+NumericPointer data_cast(Type * _pointer)
+{
+	static_assert(sizeof(Type *) == sizeof(NumericPointer),
+				  "Compilied for 64 bit arcitecture. Use NumericPointer = unsigned int");
```
kambala-decapitator, line 22:
```suggestion
				  "Compiled for 64 bit arcitecture. Use NumericPointer = unsigned int");

```

mapeditor/StdInc.h
```diff
@@ -0,0 +1,50 @@
...
+	return reinterpret_cast<NumericPointer>(_pointer);
+}
+
+template<class Type>
+Type * data_cast(NumericPointer _numeric)
+{
+	static_assert(sizeof(Type *) == sizeof(NumericPointer),
+				  "Compilied for 64 bit arcitecture. Use NumericPointer = unsigned int");
```
kambala-decapitator, line 30:
```suggestion
				  "Compiled for 64 bit arcitecture. Use NumericPointer = unsigned int");

```

mapeditor/generatorprogress.cpp
kambala-decapitator: well, it's not a problem to use unique_ptr here. It's just that Qt Creator's template creates naked pointer plus deletes it in destructor. I'd keep as is.

mapeditor/graphics.h
```diff
@@ -0,0 +1,84 @@
...
+	void loadHeroAnimations();
+	//loads animation and adds required rotated frames
+	std::shared_ptr<Animation> loadHeroAnimation(const std::string &name);
+	
+	void loadHeroFlagAnimations();
+	
+	//loads animation and adds required rotated frames
+	std::shared_ptr<Animation> loadHeroFlagAnimation(const std::string &name);
+	
+	void loadErmuToPicture();
```
kambala-decapitator, line 45: sorry, what is Ermu? :)

mapeditor/graphics.h
```diff
@@ -0,0 +1,84 @@
...
+		
+	std::map<std::string, JsonNode> imageLists;
+		
+	//functions
+	Graphics();
+	~Graphics();
+	
+	void load();
+	
+	void blueToPlayersAdv(QImage * sur, PlayerColor player); //replaces blue interface colour with a color of player
```
kambala-decapitator, line 77: maybe stupid question, but why not red?

mapeditor/graphics.cpp
```diff
@@ -0,0 +1,343 @@
...
+	}
+	
+	boatAnimations[0] = loadHeroAnimation("AB01_.DEF");
+	boatAnimations[1] = loadHeroAnimation("AB02_.DEF");
+	boatAnimations[2] = loadHeroAnimation("AB03_.DEF");
+	
+	
+	mapObjectAnimations["AB01_.DEF"] = boatAnimations[0];
+	mapObjectAnimations["AB02_.DEF"] = boatAnimations[1];
+	mapObjectAnimations["AB03_.DEF"] = boatAnimations[2];
```
kambala-decapitator, line 142: it should be a loop

mapeditor/inspector/armywidget.h
```diff
@@ -0,0 +1,49 @@
+#ifndef ARMYWIDGET_H
+#define ARMYWIDGET_H
```
kambala-decapitator, line 2: proper header missing

mapeditor/inspector/armywidget.cpp
kambala-decapitator: simple widgets are ok to draw in designer, but something more complex is easier to create (and understand) from code

mapeditor/inspector/armywidget.cpp
```diff
@@ -0,0 +1,141 @@
...
+		QStyledItemDelegate::setEditorData(editor, index);
+	}
+}
+
+void ArmyDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, const QModelIndex &index) const
+{
+	if(auto * ed = qobject_cast<ArmyWidget *>(editor))
+	{
+		auto isArmed = ed->commitChanges();
+		model->setData(index, "dummy");
```
kambala-decapitator, line 131: why is it needed?

mapeditor/inspector/inspector.h
```diff
@@ -0,0 +1,154 @@
+#ifndef INSPECTOR_H
+#define INSPECTOR_H
```
kambala-decapitator, line 2: proper header

kambala-decapitator (commented):
checked approximately half of the changed files so far

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+			QJsonObject data{{"id", QJsonValue(ID)},
+							 {"subid", QJsonValue(secondaryID)},
+							 {"template", QJsonValue(templateId)},
+							 {"animationEditor", QString::fromStdString(templ->editorAnimationFile)},
+							 {"animation", QString::fromStdString(templ->animationFile)},
+							 {"preview", jsonFromPixmap(preview)}};
+			
+			//create object to extract name
+			std::unique_ptr<CGObjectInstance> temporaryObj(factory->create(templ));
+			QString translated = useCustomName ? tr(temporaryObj->getObjectName().c_str()) : subGroupName;
```
kambala-decapitator, line 456: `tr()` call looks strange here. Is it expected that object names will be a part of editor translations? `lupdate` can't get them out of nowhere, you'll have to add them manually to .ts file.

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+	objectBrowser->setDynamicSortFilter(false);
+	objectBrowser->setRecursiveFilteringEnabled(true);
+	ui->treeView->setModel(objectBrowser);
+	ui->treeView->setSelectionBehavior(QAbstractItemView::SelectItems);
+	ui->treeView->setSelectionMode(QAbstractItemView::SingleSelection);
+	connect(ui->treeView->selectionModel(), SIGNAL(currentChanged(const QModelIndex &, const QModelIndex &)), this, SLOT(treeViewSelected(const QModelIndex &, const QModelIndex &)));
+
+
+	//adding objects
+	addGroupIntoCatalog("TOWNS", false, false, Obj::TOWN);
```
kambala-decapitator, line 532: is it possible to move all this config to a file?

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+		connect(b, &QPushButton::clicked, this, [this, river]{ roadOrRiverButtonClicked(river, false); });
+	}
+	//add spacer to keep terrain button on the top
+	ui->riverLayout->addItem(new QSpacerItem(20, 20, QSizePolicy::Minimum, QSizePolicy::Expanding));
+
+	if(objectBrowser)
+		throw std::runtime_error("object browser exists");
+
+	//model
+	objectsModel.setHorizontalHeaderLabels(QStringList() << QStringLiteral("Type"));
```
kambala-decapitator, line 520:
```suggestion
	objectsModel.setHorizontalHeaderLabels(QStringList() << tr("Type"));
```

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+
+void MainWindow::switchDefaultPlayer(const PlayerColor & player)
+{
+	if(controller.defaultPlayer == player)
+		return;
+	
+	ui->actionNeutral->blockSignals(true);
+	ui->actionNeutral->setChecked(PlayerColor::NEUTRAL == player);
+	ui->actionNeutral->blockSignals(false);
+	for(int i = 0; i < 8; ++i)
```
kambala-decapitator, line 975: iirc there's a constant

mapeditor/playerparams.cpp
```diff
@@ -0,0 +1,139 @@
...
+			auto * ctown = town->town;
+			if(!ctown)
+				ctown = VLC->townh->randomTown;
+			if(ctown && town->getOwner().getNum() == playerColor)
+			{
+				if(playerInfo.hasMainTown && playerInfo.posOfMainTown == town->pos)
+					foundMainTown = townIndex;
+				auto name = town->name + ", (random)";
+				if(ctown->faction)
+					name = town->getObjectName();
```
kambala-decapitator, line 59:
```suggestion
				const auto name = ctown->faction ? town->getObjectName() : town->name + ", (random)";
```

mapeditor/playerparams.cpp
```diff
@@ -0,0 +1,139 @@
...
+				ui->mainTown->addItem(tr(name.c_str()), QVariant::fromValue(i));
```
kambala-decapitator, line 60: this strange `tr()` again

mapeditor/playerparams.cpp
```diff
@@ -0,0 +1,139 @@
...
+	}
+	else
+	{
+		ui->generateHero->setEnabled(false);
+		playerInfo.hasMainTown = false;
+		playerInfo.generateHeroAtMainTown = false;
+		playerInfo.posOfMainTown = int3(-1, -1, -1);
+	}
+
+	ui->playerColor->setTitle(QString("PlayerID: %1").arg(playerId));
```
kambala-decapitator, line 78:
```suggestion
	ui->playerColor->setTitle(tr("Player ID: %1").arg(playerId));
```

mapeditor/playersettings.cpp
kambala-decapitator: I don't understand the meaning of 2 either

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+{
+}
+
+void ObjectsLayer::update()
+{
+	if(!map)
+		return;
+	
+	pixmap.reset(new QPixmap(map->width * 32, map->height * 32));
+	pixmap->fill(QColor(0, 0, 0, 0));
```
kambala-decapitator, line 276:
```suggestion
	pixmap->fill(Qt::transparent);
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+
+void ObjectsLayer::draw(bool onlyDirty)
+{
+	if(!pixmap)
+		return;
+	
+	if(!map)
+		return;
+	
+	pixmap->fill(QColor(0, 0, 0, 0));
```
kambala-decapitator, line 288:
```suggestion
	pixmap->fill(Qt::transparent);
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+void SelectionObjectsLayer::update()
+{
+	if(!map)
+		return;
+	
+	selectedObjects.clear();
+	onSelection();
+	shift = QPoint();
+	if(newObject)
+		delete newObject;
```
kambala-decapitator, line 344: deleting nullptr is safe
```suggestion
	delete newObject;
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+	
+	draw();
+}
+
+void SelectionObjectsLayer::draw()
+{
+	if(!pixmap)
+		return;
+	
+	pixmap->fill(QColor(0, 0, 0, 0));
```
kambala-decapitator, line 358:
```suggestion
	pixmap->fill(Qt::transparent);
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+	
+	QPainter painter(pixmap.get());
+	painter.setCompositionMode(QPainter::CompositionMode_Source);
+	painter.setPen(QColor(255, 255, 255));
```
kambala-decapitator, line 362:
```suggestion
	painter.setPen(Qt::white);
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+{
+}
+
+void MinimapViewLayer::update()
+{
+	if(!map)
+		return;
+	
+	pixmap.reset(new QPixmap(map->width, map->height));
+	pixmap->fill(QColor(0, 0, 0, 0));
```
kambala-decapitator, line 533:
```suggestion
	pixmap->fill(Qt::transparent);
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+	
+	QPainter painter(pixmap.get());
+	painter.setPen(QColor(255, 255, 255));
```
kambala-decapitator, line 536:
```suggestion
	painter.setPen(Qt::white);
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+	painter.drawRect(x, y, w, h);
+	
+	redraw();
+}
+
+void MinimapViewLayer::draw()
```
kambala-decapitator, line 542: looks almost identical to `update()`

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+{
+	if(!map)
+		return;
+	
+	pixmap->fill(QColor(0, 0, 0, 0));
```
kambala-decapitator, line 547:
```suggestion
	pixmap->fill(Qt::transparent);
```

mapeditor/scenelayer.cpp
```diff
@@ -0,0 +1,564 @@
...
+	
+	//maybe not optimal but ok
+	QPainter painter(pixmap.get());
+	painter.setPen(QColor(255, 255, 255));
```
kambala-decapitator, line 551:
```suggestion
	painter.setPen(Qt::white);
```

mapeditor/validator.cpp
```diff
@@ -0,0 +1,162 @@
...
+			{
+				if(!map->players[o->getOwner().getNum()].canAnyonePlay())
+					issues.emplace_back(QString("Object %1 is assinged to non-playable player %2").arg(o->instanceName.c_str(), o->getOwner().getStr().c_str()), true);
+			}
+			//checking towns
+			if(auto * ins = dynamic_cast<CGTownInstance*>(o.get()))
+			{
+				bool has = amountOfCastles.count(ins->getOwner().getNum());
+				if(!has && ins->getOwner() != PlayerColor::NEUTRAL)
+					issues.emplace_back(QString("Town %1 has undefined owner %s").arg(ins->instanceName.c_str(), ins->getOwner().getStr().c_str()), true);
```
kambala-decapitator, line 91:
```suggestion
					issues.emplace_back(tr("Town %1 has undefined owner %2").arg(ins->instanceName.c_str(), ins->getOwner().getStr().c_str()), true);
```

mapeditor/windownewmap.cpp
```diff
@@ -0,0 +1,407 @@
...
+	}
+	//TODO: teams when implemented
+
+	auto waterContent = s.value(newMapWaterContent);
+	if (waterContent.isValid())
+	{
+		switch (waterContent.toInt())
+		{
+			case EWaterContent::RANDOM:
+				ui->waterOpt1->setChecked(true); break;
```
kambala-decapitator, line 86: water and monster options variables need a better name

mapeditor/windownewmap.cpp
```diff
@@ -0,0 +1,407 @@
...
+	s.setValue(newMapHeight, ui->heightTxt->text().toInt());
+	s.setValue(newMapTwoLevel, ui->twoLevelCheck->isChecked());
+	s.setValue(newMapGenerateRandom, ui->randomMapCheck->isChecked());
+
+	s.setValue(newMapPlayers,ui->humanCombo->currentIndex());
+	s.setValue(newMapCpuPlayers,ui->cpuCombo->currentIndex());
+	//TODO: teams when implemented
+
+	EWaterContent::EWaterContent water = EWaterContent::RANDOM;
+	if(ui->waterOpt1->isChecked())
```
kambala-decapitator, line 144: maybe create a list of pairs for water and monsters to avoid manual checks when reading and writing

mapeditor/windownewmap.ui
```diff
@@ -0,0 +1,816 @@
...
+    <rect>
+     <x>310</x>
+     <y>20</y>
+     <width>111</width>
+     <height>101</height>
+    </rect>
+   </property>
+   <layout class="QVBoxLayout" name="verticalLayout">
+    <item>
+     <widget class="QPushButton" name="okButtong">
```
kambala-decapitator, line 768:
```suggestion
     <widget class="QPushButton" name="okButton">
```

mapeditor/windownewmap.h
```diff
@@ -0,0 +1,92 @@
...
+	};
+
+public:
+	explicit WindowNewMap(QWidget *parent = nullptr);
+	~WindowNewMap();
+
+private slots:
+	void on_cancelButton_clicked();
+
+	void on_okButtong_clicked();
```
kambala-decapitator, line 60:
```suggestion
	void on_okButton_clicked();
```

mapeditor/windownewmap.cpp
```diff
@@ -0,0 +1,407 @@
...
+	map->height = options.getHeight();
+	map->twoLevel = options.getHasTwoLevels();
+	
+	map->initTerrain();
+	map->getEditManager()->clearTerrain(&CRandomGenerator::getDefault());
+
+	return std::move(map);
+}
+
+void WindowNewMap::on_okButtong_clicked()
```
kambala-decapitator, line 196:
```suggestion
void WindowNewMap::on_okButton_clicked()
```

mapeditor/windownewmap.cpp
```diff
@@ -0,0 +1,407 @@
...
+	}
+	else
+	{		
+		auto f = std::async(std::launch::async, &::generateEmptyMap, std::ref(mapGenOptions));
+		nmap = f.get();
+	}
+	
+
+	static_cast<MainWindow*>(parent())->controller.setMap(std::move(nmap));
+	static_cast<MainWindow*>(parent())->initializeMap(true);
```
kambala-decapitator, line 257: not nice :)

mapeditor/windownewmap.cpp
```diff
@@ -0,0 +1,407 @@
...
+	close();
+}
+
+void WindowNewMap::on_sizeCombo_activated(int index)
+{
+	std::map<int, std::pair<int, int>> sizes
```
kambala-decapitator, line 263: vector of `QSize`s would probably be better

kambala-decapitator:
when you open Player Settings, configuration UI doesn't appear until you switch number of players in the combobox, even if initial value is correct

CMakeLists.txt
```diff
@@ -61,6 +61,8 @@ endif()
 option(ENABLE_ERM "Enable compilation of ERM scripting module" OFF)
 option(ENABLE_LUA "Enable compilation of LUA scripting module" OFF)
 option(ENABLE_LAUNCHER "Enable compilation of launcher" ON)
+option(ENABLE_EDITOR "Enable compilation of map editor" ON)
```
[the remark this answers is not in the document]
nullkiller: Having thing disabled by default everywhere will cause people to forget about it and errors accumulate unattended.

mapeditor/generatorprogress.cpp
nullkiller: Yes sure if it is qt approach

mapeditor/mapview.h
nullkiller: Afaik vector should have amortized O(N) and I have seen an article showing that linked list is way slower than vector of ints because it needs allocation for each new item while vector always allocates 2x capacity each time it has not enough space and just does memory copy. So if the vector item is quite big - this might be a problem. In this case we can store shared pts as an option.
But I agree regarding random access)

mapeditor/objectbrowser.cpp
```diff
@@ -0,0 +1,68 @@
...
+	auto objId = data["id"].toInt();
+	auto objSubId = data["subid"].toInt();
+	auto templateId = data["template"].toInt();
+
+	auto factory = VLC->objtypeh->getHandlerFor(objId, objSubId);
+	auto templ = factory->getTemplates()[templateId];
+
+	result = result & templ->canBePlacedAt(terrain);
+
+	//text filter
```
nullkiller, line 45: Is it TODO? If not I do not understand the comment

mapeditor/playersettings.cpp
nullkiller: single player? There is another approach. auto singlePlayer = players < 2;

nullkiller: So there is a lot of this 2 everywhere so need a constant)

mapeditor/playersettings.cpp
```diff
@@ -0,0 +1,77 @@
...
+		ui->playersCount->setCurrentText("");
+	else
+		ui->playersCount->setCurrentIndex(players - 2);
```
nullkiller, line 29: What - 2 mins here? Why do we need it?

mapeditor/playersettings.cpp
nullkiller: Btw MIN_NUM_PLAYERS = 1 might be possible and use players <= MIN_NUM_PLAYERS

mapeditor/playersettings.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+PlayerSettings::~PlayerSettings()
+{
+	delete ui;
+}
+
+void PlayerSettings::on_playersCount_currentIndexChanged(int index)
+{
+	assert(index + 2 <= controller.map()->players.size());
```
nullkiller, line 41: will it work?

auto selectedPlayerCount = index + 2;

and use it everywhere?

mapeditor/playersettings.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+	for(int i = 0; i < index + 2; ++i)
+	{
+		if(i < paramWidgets.size())
+			continue;
```
nullkiller, line 46: Shouldn't we set p.canComputerPlay = true; when continue triggers?

nullkiller: I see it is needed only for new widgets.

mapeditor/validator.cpp
```diff
@@ -0,0 +1,162 @@
...
+			if(p.canComputerPlay)
+				++cplayers;
+			if(p.canHumanPlay)
+				++hplayers;
+			if(p.allowedFactions.empty())
+				issues.emplace_back(QString("No factions allowed for player %1").arg(i), true);
+		}
+		if(hplayers + cplayers == 0)
+			issues.emplace_back("No players allowed to play this map", true);
+		if(hplayers + cplayers == 1)
```
nullkiller, line 61: It is possible to have a map with only one player. Its goal can be to find grail for instance.

nullkiller: Sure if your editor does not support special win conditions - it is correct that we need at least 2

mapeditor/windownewmap.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+	EMonsterStrength::EMonsterStrength monster = EMonsterStrength::RANDOM;
+	if(ui->monsterOpt1->isChecked())
+		monster = EMonsterStrength::RANDOM;
+	else if(ui->monsterOpt2->isChecked())
+		monster = EMonsterStrength::GLOBAL_WEAK;
+	else if(ui->monsterOpt3->isChecked())
+		monster = EMonsterStrength::GLOBAL_NORMAL;
+	else if(ui->monsterOpt4->isChecked())
+		monster = EMonsterStrength::GLOBAL_STRONG;
```
nullkiller, line 162: @kambala-decapitator  as far as I understood qml can generate a list of controls from collection dynamically? data bindings and so on?

mapeditor/windownewmap.cpp
nullkiller: We can introduce an event functor onMapCreated or something. Maybe qt has some specific mechanics

nullkiller (approved):

nullkiller:
@Nordsoft91, please, check out if you want to address any comments and ping me back regarding merge

mapeditor/windownewmap.cpp
kambala-decapitator: - not sure why you're mentioning QML, here it's "pure" Qt
- Qt doesn't really have "data binding". What's your idea?

Nordsoft91:
I'll return to it later to fix problems. Thanks for feedback

mapeditor/windownewmap.cpp
nullkiller: Yes, you mentioned qml as an option to rewrite launcher. I understand that this is yet another thing. Never mind.

kambala-decapitator: yes, that's about launcher, and here it's editor. Initially there's no plan to build / adapt it for touch devices.

all these similar UI elements like `ui->monsterOpt4` can be created from code for sure, e.g. in a loop.

mapeditor/graphics.h
Nordsoft91: Unfortunately, it's not possible. Frankly speaking , it's possible technically, but if we move rendering code to lib, it would create dependency on SDL and QT which we certainly don't want. The issue here is that vcmiclient uses SDL for rendering while editor uses QT. Having dependency from editor to client is also bad, because in such case editor will depend on SDL.

mapeditor/inspector/armywidget.cpp
Nordsoft91: it's needed to set proper model size

mapeditor/inspector/rewardswidget.cpp
```diff
@@ -0,0 +1,329 @@
...
+	if(l.empty())
+		ui->rewardList->setEnabled(false);
+	
+	for(auto & s : l)
+		ui->rewardList->addItem(s);
+}
+
+void RewardsWidget::obtainData()
+{
+	if(pandora)
```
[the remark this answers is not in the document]
Nordsoft91: it does, because rewards could be not only from pandoras. Since I support quest type, there will be another constructor to build from it

mapeditor/inspector/townbulidingswidget.cpp
```diff
@@ -0,0 +1,243 @@
...
+	checks.back()->setData(bId, Qt::UserRole);
+	
+	if(building->getBase() == buildingId)
+	{
+		model.appendRow(checks);
+	}
+	else
+	{
+		QStandardItem * parent = nullptr;
+		std::vector<QModelIndex> stack;
```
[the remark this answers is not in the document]
Nordsoft91: I know but it's more like a syntax sugar

mapeditor/inspector/rewardswidget.cpp
```diff
@@ -0,0 +1,329 @@
...
+		auto itemCurr = new QTableWidgetItem(getListForType(typeId)[listId]);
+		itemCurr->setData(Qt::UserRole, listId);
+		ui->rewardsTable->setItem(rewards - 1, 1, itemCurr);
+	}
+	
+	QString am = QString::number(amount);
+	switch(ui->rewardType->currentIndex())
+	{
+		case 6:
+			if(amount <= 1)
```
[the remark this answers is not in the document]
Nordsoft91: Should be.Tried to  do same way as it's deon in old editor

mapeditor/mainwindow.cpp
```diff
@@ -0,0 +1,1115 @@
...
+	if(!controller.map())
+		return;
+
+	auto filenameSelect = QFileDialog::getSaveFileName(this, tr("Save map"), "", tr("VCMI maps (*.vmap)"));
+
+	if(filenameSelect.isNull())
+		return;
+
+	if(filenameSelect == filename)
+		return;
```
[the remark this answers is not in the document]
Nordsoft91: otherwise it'll save file each time when you save it. There is task about it https://github.com/Nordsoft91/vcmi/issues/31

mapeditor/mainwindow.cpp
Nordsoft91: for some reason without tr it shows incorrect symbols

mapeditor/mapcontroller.cpp
```diff
@@ -0,0 +1,499 @@
...
+#include "mainwindow.h"
+#include "inspector/inspector.h"
+
+
+MapController::MapController(MainWindow * m): main(m)
+{
+	_scenes[0].reset(new MapScene(0));
+	_scenes[1].reset(new MapScene(1));
+	_miniscenes[0].reset(new MinimapScene(0));
+	_miniscenes[1].reset(new MinimapScene(1));
```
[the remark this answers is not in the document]
Nordsoft91: no need to pass business logic here

mapeditor/mapview.cpp
```diff
@@ -0,0 +1,449 @@
...
+
+	//TODO: cast parent->parent to MainWindow in order to show coordinates or another way to do it?
+	//main->setStatusMessage(QString("x: %1 y: %2").arg(tile.x, tile.y));
+
+	switch(selectionTool)
+	{
+	case MapView::SelectionTool::Brush:
+		if(mouseEvent->buttons() & Qt::RightButton)
+			sc->selectionTerrainView.erase(tile);
+		else if(mouseEvent->buttons() == Qt::LeftButton)
```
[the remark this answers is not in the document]
Nordsoft91: for explicit intention: no other buttons should be press if you cereate something

CMakeLists.txt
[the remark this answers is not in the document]
kambala-decapitator: not sure which errors you're talking about. We have CI that builds for various platforms.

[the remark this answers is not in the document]
nullkiller: If CI will have it ON by default it is ok.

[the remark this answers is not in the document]
kambala-decapitator: CI must build everything that is supported, of course. But most users wouldn't care about the editor.

mapeditor/mainwindow.cpp
kambala-decapitator: sounds like encoding issues. Do you have Russian text? If yes, I guess it's not UTF-8?

mapeditor/mapview.cpp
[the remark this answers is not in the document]
kambala-decapitator: but why are other buttons allowed for RMB? I don't get the logic. basically, if user presses both RMB and LMB at the same time, I'd simply ignore this case as it's unclear what he's trying to do.

mapeditor/windownewmap.cpp
kambala-decapitator: Qt has signal-slot

Nordsoft91:
@nullkiller , I've finished with minimum I wanted to have, fix it for mxe as well but looks like will be tough for me to continue development actively. I suggest to merge and if we decide do not release, it can be done easily by switching cmake option

Nordsoft91:
Regular merge please (in order to keep history)

PR #975, opened 2022-09-23
Incompatibility response

Nordsoft91:
Bypass error message from the server


client/CServerHandler.cpp
```diff
@@ -546,6 +546,11 @@ void CServerHandler::startCampaignScenario(std::shared_ptr<CCampaignState> cs)
 	SDL_PushEvent(&event);
 }
 
+void CServerHandler::showServerError(std::string txt)
```
kambala-decapitator, line 549:
```suggestion
void CServerHandler::showServerError(const std::string & txt)
```

client/CServerHandler.h
```diff
@@ -137,6 +137,7 @@ class CServerHandler : public IServerAPI, public LobbyInfo
 	void startGameplay();
 	void endGameplay(bool closeConnection = true, bool restart = false);
 	void startCampaignScenario(std::shared_ptr<CCampaignState> cs = {});
+	void showServerError(std::string txt);
```
kambala-decapitator, line 140:
```suggestion
	void showServerError(const std::string & txt);
```

config/translate.json
```diff
@@ -31,7 +31,8 @@
 	{
 		"errors" :
 		{
-			"existingProcess" : "Another vcmiserver process is running, please terminate it first"
+			"existingProcess" : "Another vcmiserver process is running, please terminate it first",
+			"modsIncompatibility" : "Mods are required to load game:"
```
kambala-decapitator, line 35:
```suggestion
			"modsIncompatibility" : "Required mods to load game:"
```

lib/CModHandler.h
```diff
@@ -276,11 +276,32 @@ class DLL_LINKAGE CModHandler
...
-		{}
+		using StringPair = std::pair<const std::string, const std::string>;
+		using ModList = std::list<StringPair>;
+		
+		Incompatibility(ModList && _missingMods):
+			missingMods(std::move(_missingMods))
+		{
+			std::ostringstream _ss;
+			for(auto & m : missingMods)
+				_ss << m.first << ' ' << m.second << std::endl;
```
kambala-decapitator, line 290:
```suggestion
			for(const auto & m : missingMods)
				_ss << m.first << ' ' << m.second << '\n';

```

lib/CModHandler.h
```diff
@@ -366,30 +387,29 @@ class DLL_LINKAGE CModHandler
...
 				h & allMods[m].version;
 		}
 		else
 		{
 			loadMods();
 			std::vector<TModID> newActiveMods;
 			h & newActiveMods;
+			
+			Incompatibility::ModList missingMods;
 			for(auto & m : newActiveMods)
```
kambala-decapitator, line 399:
```suggestion
			for(const auto & m : newActiveMods)

```

lib/NetPacksLobby.h
```diff
@@ -309,3 +309,15 @@ struct LobbyForceSetPlayer : public CLobbyPackToServer
...
 	}
 };
+
+struct LobbyShowMessage : public CLobbyPackToPropagate
+{
+	std::string message;
+	
+	void applyOnLobbyScreen(CLobbyScreen * lobby, CServerHandler * handler);
+	
+	template <typename Handler> void serialize(Handler &h, const int version)
```
kambala-decapitator, line 319:
```suggestion
	template <typename Handler> void serialize(Handler & h, const int version)
```

server/CGameHandler.cpp
```diff
@@ -2984,12 +2984,22 @@ void CGameHandler::load(const std::string & filename)
 		}
 		logGlobal->info("Game has been successfully loaded!");
 	}
-	catch(std::exception &e)
+	catch(const CModHandler::Incompatibility & e)
 	{
 		logGlobal->error("Failed to load game: %s", e.what());
+		auto errorMsg = VLC->generaltexth->localizedTexts["server"]["errors"]["modsIncompatibility"].String() + '\n';
+		errorMsg += e.what();
+		lobby->announceMessage(errorMsg);
```
kambala-decapitator, line 2992:
```suggestion
		lobby->announceMessage(std::move(errorMsg));
```

kambala-decapitator (approved):

client/CServerHandler.cpp
Nordsoft91: That's would be an easy way to get a potential crash in multithreading code

lib/CModHandler.h
Nordsoft91: I don't think it's differs somehow for string stream but you need to be cautious with suggested approach when you are working with more than one thread

server/CGameHandler.cpp
Nordsoft91: announceMessage receives const reference

client/CServerHandler.cpp
kambala-decapitator: tbh I don't understand how (not a multithreading expert). Could you provide an example?

lib/CModHandler.h
kambala-decapitator: the caution would probably apply when string stream can be shared between threads? Here it's a local variable. Or maybe I simply don't know something.

client/CServerHandler.cpp
nullkiller: I can only imagine such case with async lambdas which also capture variables by &. But this case does not look like a problem of such kind. Also it is more a problem of that async lambdas. They should capture by value.

Nordsoft91: You may have class with std::string datafield, which you can transfer to the function.
Class will be destroyed earlier that function is finished.

lib/CModHandler.h
Nordsoft91: local for sure

client/CServerHandler.cpp
kambala-decapitator: aha, by multithreading you meant async call basically. And I thought it's about multiple threads calling the method at the same time. The former makes sense.

Nordsoft91: The common approach I prefer is to use copy instead of reference **anytime** when it **may** be asynchronous calls. The cost payed is nothing in comparison with debugging and searching floating multithreading bugs.

nullkiller: This looks kind of generic practice and will work only if put to code style like do not pass vars by const & where performance is not critical, pass by value instead.

nullkiller: Our code style has nothing regarding passing parameters by const refs btw.

nullkiller: Btw we already have void sendMessage(const std::string & txt). Cant we use it?

Nordsoft91: no, this one is for chat message (function name is misleading for sure)

client/NetPacksLobbyClient.cpp
```diff
@@ -137,3 +137,9 @@ void LobbyUpdateState::applyOnLobbyScreen(CLobbyScreen * lobby, CServerHandler *
 	if(hostChanged)
 		lobby->toggleMode(handler->isHost());
 }
+
+void LobbyShowMessage::applyOnLobbyScreen(CLobbyScreen * lobby, CServerHandler * handler)
+{
+	lobby->buttonStart->block(false);
```
nullkiller, line 143: Who will unlock it later?

Nordsoft91: that's unlocking

PR #978, opened 2022-09-24
Hide unsupported resolutions

### LINKED ISSUE #947: hide resolutions larger than the biggest possible
in #941 we discovered that attempt to select a larger resolution than macbook's native (in _points_, not in _pixels_) crashes game immediately. I suggest to simply hide larger resolutions in such case:
- in the launcher
- in the game menu

kambala-decapitator:
fixes #947


kambala-decapitator:
I'll also need to test on my Windows laptop first

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -107,6 +112,33 @@ CSettingsView::CSettingsView(QWidget * parent)
 {
 	ui->setupUi(this);
 
+	// TODO: read available resolutions from mods
```
Nordsoft91, line 115: Will be challenging from launcher because it doesn't initialize VLC.
suggestion: use settings.json instead

kambala-decapitator: sorry, not sure how settings.json can help determine all supported resolutions...

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -107,6 +112,33 @@ CSettingsView::CSettingsView(QWidget * parent)
...
+		{1920, 1080},
+	};
+	// TODO: handle multiple screens
+	const auto screenSize = qGuiApp->primaryScreen()->size();
+	for(const auto & resolution : supportedResolutions)
+	{
+		if(screenSize.width() < resolution.width() || screenSize.height() < resolution.height())
+			break;
+		ui->comboBoxResolution->addItem(sizeToString(resolution));
+	}
```
Nordsoft91, line 140: Let's remove implicit dependency on index as it can create unexpected bugs with missing resolutions. Do something like this instead
```suggestion
	for(int rId = 0; rId < supportedResolutions.size(); ++rId)
	{
		if(screenSize.width() < supportedResolutions[rId].width() || screenSize.height() < supportedResolutions.height())
			continue;
		ui->comboBoxResolution->addItem(sizeToString(supportedResolutions[rId]), rId);
	}
```

Don't commit this as is, because then you need to extract UserData from item

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -19,6 +19,14 @@
 #include "../../lib/CConfigHandler.h"
 #include "../../lib/VCMIDirs.h"
 
+namespace
+{
+QString sizeToString(const QSize & size)
```
Nordsoft91, line 24: resolutionToString maybe?

kambala-decapitator: yes, good point

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: not sure I'm following. When the loop is entered, the combox has only one hardcoded value: 800x600. The list of resolutions was previously hardcoded in the .ui file, now it's in .cpp with the filter.

nullkiller (approved):

kambala-decapitator:
> I'll also need to test on my Windows laptop first

looks good.

@skaurus could you also test on your side? Get binary from https://github.com/vcmi/vcmi/actions/runs/3118015944

launcher/settingsView/csettingsview_moc.cpp
Nordsoft91: you can move hardcoded values there

launcher/settingsView/csettingsview_moc.cpp
Nordsoft91: filter cuts off everything since one dimension higher than resolution. I suggest to use "continue" instead

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: but settings.json is "user-generated" file, it makes no sense to place hardcoded resolutions there.

Or are you talking about some other config file?

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: that's why the hardcoded list is sorted and respective comment is given :)

it's kind of temporary workaround anyway. The proper way is to obtain resolutions from mods.

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: What do you think of the additional quick hack (for the time being, of course): add resolutions to the list only when Extras are installed? In launcher that would be a dead simple file presence check at hardcoded location.

Nordsoft91: For settings.json it can have same logic as for other configs. Some default values can be there + mods overrides + user overrides. Separate can be also an option.

And yeah, I agree with your proposal. That's the best one: we have file with single default option and extras can override those values. The only problem is we need to teach launcher to load mods at some level

launcher/settingsView/csettingsview_moc.cpp
Nordsoft91: I'll give you an example: If I have device with screen 1366x768 you code won't give me option to select proper resolution, because cycle will break on 1280x800 (screenSize.height < resolution.height)

I believe it's a bug which needs to be fixed prior merge.

Nordsoft91 (changes requested):
Looks like there is a regression

skaurus:
@kambala-decapitator yes, I do not see unsupported resolutions with this build

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: Thank you for the example! Now I got it. You're absolutely right.

nullkiller:
@kambala-decapitator, @Nordsoft91 please update this PR status. Are all requested changes implemented, can be merged?

kambala-decapitator:
not yet

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -110,6 +106,50 @@ void CSettingsView::loadSettings()
...
 
+void CSettingsView::fillValidResolutions(int screenIndex)
+{
+	ui->comboBoxResolution->blockSignals(true); // avoid saving wrong resolution after adding first item from the list
+	ui->comboBoxResolution->clear();
+
+	// TODO: read available resolutions from all mods
+	const QLatin1String extrasResolutionsPath{"/vcmi-extras/Mods/extraResolutions/Content/config/resolutions.json"};
+	const auto extrasResolutionsJson = JsonUtils::JsonFromFile(CLauncherDirs::get().modsPath() + extrasResolutionsPath);
+	const auto resolutions = extrasResolutionsJson.toMap().value(QLatin1String{"GUISettings"}).toList();
```
Nordsoft91, line 117: what if mod installed but deactivated?

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -110,6 +106,50 @@ void CSettingsView::loadSettings()
 	ui->comboBoxAutoSave->setCurrentIndex(settings["general"]["saveFrequency"].Integer() > 0 ? 1 : 0);
 }
 
+void CSettingsView::fillValidResolutions(int screenIndex)
+{
+	ui->comboBoxResolution->blockSignals(true); // avoid saving wrong resolution after adding first item from the list
+	ui->comboBoxResolution->clear();
+
+	// TODO: read available resolutions from all mods
+	const QLatin1String extrasResolutionsPath{"/vcmi-extras/Mods/extraResolutions/Content/config/resolutions.json"};
```
Nordsoft91, line 115: Suggest to move this hardcoded string to settings.json (see schemas/settings.json)

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -110,6 +106,50 @@ void CSettingsView::loadSettings()
...
+	const auto currentScreen = screenIndex < screens.size() ? screens[screenIndex] : qGuiApp->primaryScreen();
+	const auto screenSize = currentScreen->size();
+	for(const auto & entry : resolutions)
+	{
+		const auto resolutionMap = entry.toMap().value(QLatin1String{"resolution"}).toMap();
+		if(resolutionMap.isEmpty())
+			continue;
+		const QSize resolution{resolutionMap[QLatin1String{"x"}].toInt(), resolutionMap[QLatin1String{"y"}].toInt()};
+#ifndef VCMI_IOS
+		if(screenSize.width() < resolution.width() || screenSize.height() < resolution.height())
```
Nordsoft91, line 135: if "x" or "y" is missing, there will be 0 dimension
```suggestion
		if(!resolution.isEmpty() && (screenSize.width() < resolution.width() || screenSize.height() < resolution.height()))
```

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: good point. Shouldn't be difficult to check that.

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: that's probably quite an edge case, but I agree with sanity check

launcher/modManager/cmodmanager.cpp
```diff
@@ -18,7 +18,11 @@
 #include "../jsonutils.h"
 #include "../launcherdirs.h"
 
-static QString detectModArchive(QString path, QString modName)
+namespace
+{
+const QLatin1String extraResolutionsMod{"vcmi-extras.extraresolutions"};
```
Nordsoft91, line 23: So if we have another mod, e.g. "android-resolutions" it won't work?

kambala-decapitator: yes, currently it's only about VCMI extras, just like I initially proposed. That's why proper check of all mods is needed.

config/schemas/settings.json
```diff
@@ -378,6 +378,10 @@
 					"type" : "boolean",
 					"default" : true
 				},
+				"extraResolutionsModPath" : {
+					"type" : "string",
+					"default" : "/vcmi-extras/Mods/extraResolutions/Content/config/resolutions.json"
```
Nordsoft91, line 383: maybe a vector of configs to allow resolutions from other mods too?

kambala-decapitator: I agree, good idea

kambala-decapitator: mod name should also be placed here then instead of hardcoding

Nordsoft91 (approved):
It won't support resolutions from multiple mods in launcher (but should work in game) but still much better than if was before. So supporting other mods is up to you

kambala-decapitator:
let's keep without any more changes for now

client/windows/GUIClasses.cpp
```diff
@@ -562,10 +562,21 @@ void CSystemOptionsWindow::selectGameRes()
 	std::vector<std::string> items;
 	const JsonNode & texts = CGI->generaltexth->localizedTexts["systemOptions"]["resolutionMenu"];
 
-	for( config::CConfigHandler::GuiOptionsMap::value_type& value : conf.guiOptions)
+#ifndef VCMI_IOS
+	SDL_Rect displayBounds;
+	SDL_GetDisplayBounds(std::max(0, SDL_GetWindowDisplayIndex(mainWindow)), &displayBounds);
```
nullkiller, line 567: What value displayBounds will have for IOS?

nullkiller:
Does it affect ingame resolution switcher?

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -110,6 +106,66 @@ void CSettingsView::loadSettings()
...
+	if(resolutions.isEmpty())
+	{
+		ui->comboBoxResolution->blockSignals(false);
+		ui->comboBoxResolution->addItem(resolutionToString({800, 600}));
+		return;
+	}
+
+	const auto screens = qGuiApp->screens();
+	const auto currentScreen = screenIndex < screens.size() ? screens[screenIndex] : qGuiApp->primaryScreen();
+	const auto screenSize = currentScreen->size();
```
nullkiller, line 137: Looks like this variable can be also wrapped with
#ifndef VCMI_IOS
?

kambala-decapitator: All the 3 variables can be, but I prefer to keep less ifdefs - I believe in this case compiler would just throw away unused variables.

kambala-decapitator:
> Does it affect ingame resolution switcher?

Yes, unsupported resolutions will be hidden there as well

launcher/settingsView/csettingsview_moc.cpp
Nordsoft91: agree, let's try to keep less platform dependent logic, especially in launcher

Issue #973, opened 2022-09-22
Wrong battlefield on Favorable Winds


DjWarmonger:
![image](https://user-images.githubusercontent.com/2566990/191819969-d7ea89d5-65ef-44d4-8e73-53f72226f083.png)

Link to the map:
https://drive.google.com/file/d/1wlC_v_ADpTakd9mPxXtQq13AhIyL-VHE/view

DjWarmonger:
This should be regular ship-to-ship battlefield. This happens only inside Favourable Winds, on clear water it's ok.

PR #979, opened 2022-09-24
#973 - use ship battleground for favorable winds

nullkiller:
Looks like favorable winds resource was not finished and regular water was used instead in original homm3.

### DIFF
--- a/config/objects/generic.json
+++ b/config/objects/generic.json
@@ -970,7 +970,7 @@
 	"favorableWinds" : {
 		"index" :225,
 		"handler": "terrain",
-		"types" : { "object" : { "index" : 0, "battleground": "favorable_winds" } }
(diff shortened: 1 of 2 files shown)


config/objects/generic.json
```diff
@@ -970,7 +970,7 @@
 	"favorableWinds" : {
 		"index" :225,
 		"handler": "terrain",
-		"types" : { "object" : { "index" : 0, "battleground": "favorable_winds" } }
+		"types" : { "object" : { "index" : 0, "battleground": "ship" } }
```
Nordsoft91, line 973: Can favourableWinds be placed not on the water?

nullkiller: Not in original map editor

PR #981, opened 2022-09-24
Fix Clang Analyzer warnings

kambala-decapitator:
The following warnings were not fixed:

1. should never be the case, as when `waysToVisitObj` is not empty, this guarantees that `closestWay` will be set to a valid pointer.

```
AI/Nullkiller/Behaviors/CaptureObjectsBehavior.cpp:140:6: warning: Called C++ object pointer is null [core.CallAndMessage]
                        = closestWay->movementCost() / way->getPath().movementCost();
                          ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

2. @nullkiller says it's ok to have it uninitialized. There're more uninitialized fields: `manaCost`, `chainOther`. Points to `HeroChainCalculationTask::calculateExchange()` in NK AI.

```
In file included from AI/Nullkiller/Pathfinding/AINodeStorage.cpp:11:
AI/Nullkiller/Pathfinding/../Engine/../Goals/../Analyzers/../Pathfinding/AINodeStorage.h:34:8: warning: Value assigned to field 'danger' in implicit constructor is garbage or undefined [core.uninitialized.Assign]
struct AIPathNode : public CGPathNode
       ^~~~~~~~~~
```

3. this one I simply didn't understand, buried quite deep. Points to `CGameHandler::load()` - `lf >> *this`.

```
In file included from server/CGameHandler.cpp:42:
In file included from server/CVCMIServer.h:12:
In file included from server/../lib/serializer/Connection.h:12:
server/../lib/serializer/BinaryDeserializer.h:206:3: warning: Method called on moved-from object 'value' [cplusplus.Move]
                hlp.serialize(*this,fileVersion);
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/lib/CStack.cpp
+++ b/lib/CStack.cpp
@@ -304,6 +304,7 @@ std::vector<BattleHex> CStack::meleeAttackHexes(const battle::Unit * attacker, c
 			res.push_back(otherDefenderPos);
 		}
 	}
+	UNUSED(mask);
 
 	return res;
 }
--- a/lib/NetPacksLib.cpp
+++ b/lib/NetPacksLib.cpp
@@ -916,10 +916,8 @@ DLL_LINKAGE void SwapStacks::applyGs(CGameState * gs)
 
 DLL_LINKAGE void InsertNewStack::applyGs(CGameState *gs)
 {
-	auto s = new CStackInstance(type, count);
-	auto obj = gs->getArmyInstance(army);
-	if(obj)
-		obj->putStack(slot, s);
+	if(auto obj = gs->getArmyInstance(army))
+		obj->putStack(slot, new CStackInstance(type, count));
 	else
 		logNetwork->error("[CRITICAL] InsertNewStack: invalid army object %d, possible game state corruption.", army.getNum());
 }
--- a/AI/Nullkiller/AIUtility.cpp
+++ b/AI/Nullkiller/AIUtility.cpp
@@ -307,6 +307,9 @@ bool compareArtifacts(const CArtifactInstance * a1, const CArtifactInstance * a2
 
 bool isWeeklyRevisitable(const CGObjectInstance * obj)
 {
+	if(!obj)
+		return false;
+
 	//TODO: allow polling of remaining creatures in dwelling
 	if(dynamic_cast<const CGVisitableOPW *>(obj)) // ensures future compatibility, unlike IDs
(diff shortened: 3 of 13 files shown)


lib/NetPacksLib.cpp
```diff
@@ -921,7 +921,10 @@ DLL_LINKAGE void InsertNewStack::applyGs(CGameState *gs)
 	if(obj)
 		obj->putStack(slot, s);
 	else
+	{
 		logNetwork->error("[CRITICAL] InsertNewStack: invalid army object %d, possible game state corruption.", army.getNum());
+		delete s;
```
Nordsoft91, line 926: For sure, we need to make some refactoring in future to use std::move for methods such as putStack instead of transferring memory ownership

Nordsoft91 (changes requested):
There is only one change about hero checking I highly recommend to make.
All other comments are optional.

lib/CStack.cpp
```diff
@@ -304,6 +304,7 @@ std::vector<BattleHex> CStack::meleeAttackHexes(const battle::Unit * attacker, c
 			res.push_back(otherDefenderPos);
 		}
 	}
+	UNUSED(mask);
```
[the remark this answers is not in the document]
kambala-decapitator: analyser says that `mask` is unused after the last assignment (a few lines above). In case a new condition is added, it'd be very easy to forget to add the assignment back. That's why I mark the variable as unused explicitly.

Nordsoft91 (approved):

lib/NetPacksLib.cpp
nullkiller: Can't we first check obj and if it is not null
auto s = new CStackInstance(type, count);

nullkiller: So I wait for changes here.

nullkiller (changes requested):

lib/NetPacksLib.cpp
kambala-decapitator: fixed it

kambala-decapitator:
P.S. We should also try scanning with PVS-Studio

Nordsoft91:
PVS studio is quite useful, I used it a lot before and it can find real bugs

PR #992, opened 2022-09-25
[MSVC] place AI dependencies into the right location at build time

kambala-decapitator:
Currently when you build project with MSVC, fuzzylite.dll and tbb.dll end up in the `AI` subdirectory which prevents loading adventure AI dll on game start and client crashes, and you have to move those dlls to the `bin` directory manually. This workaround fixes that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,6 @@ project(VCMI)
 #
 # Vckpg:
 # - Improve install code once there is better way to deploy DLLs and Qt plugins
-# - Move Vcpkg install BundleUtilities code from osx/CMakeLists.txt
 #
 # Other:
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -564,7 +563,6 @@ if(WIN32)
 	set(CPACK_NSIS_EXECUTABLES_DIRECTORY ".")
 	# Use BundleUtilities to fix build when Vcpkg is used and disable it for MXE
 	if(NOT (${CMAKE_CROSSCOMPILING}))
-		add_subdirectory(osx)
```
nullkiller, line 567: No need to connect osx here?

nullkiller: I see it is pure if(WIN32) section

PR #998, opened 2022-09-25
[994] fixed

Nordsoft91:


kambala-decapitator (approved):
logically looks correct, let's also test the binary

kambala-decapitator:
tested: works good!

PR #1000, opened 2022-09-26
Multiplayer

Nordsoft91:
Ok, then here is code which actually allows multiplayer among different platforms and architectures.
As side effect we have boost on RMG performance because now map is generated only once on server side and transferred it over network.

Tested scenarios:
 - start scenario (predefined map) (SP/MP)
 - start random map (SP/MP)
 - save, load, restart scenario (SP/MP)
 - campaign (but still cannot test next missions due to another bug)

No confidence:
 - save random map flag (should work for SP, but not sure about MP)
 - headless
 - AI-only

Known MP-related bugs (not related to the change)
 - cannot swap host/client for saved games
 - crash if someone exits from the game

I tried different approaches to manage this, but looks like that's the best. The main problem is that a lot of objects are connected between each other via raw pointers. That's probably a reason why it wasn't done from the beginning. For instance, artifacts have fields constituents and constituentOf, which creates infinite loop during serialisation. Similar story was for heroes and army. This PR is kind of combination of workarounds to pass all of that.
Note, that bugs, related to MP only, should be fixed separately as they are no regression

### DIFF
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -200,6 +200,8 @@ void CClient::newGame(CGameState * initializedGameState)
 void CClient::loadGame(CGameState * initializedGameState)
 {
 	logNetwork->info("Loading procedure started!");
+	
+	std::unique_ptr<CLoadFile> loader;
 
 	if(initializedGameState)
 	{
@@ -208,7 +210,6 @@ void CClient::loadGame(CGameState * initializedGameState)
 	}
 	else
 	{
-		std::unique_ptr<CLoadFile> loader;
 		try
 		{
 			boost::filesystem::path clientSaveName = *CResourceHandler::get("local")->getResourceName(ResourceID(CSH->si->mapname, EResType::CLIENT_SAVEGAME));
@@ -234,8 +235,6 @@ void CClient::loadGame(CGameState * initializedGameState)
 				loadCommonState(checkingLoader);
 				loader = checkingLoader.decay();
 			}
-			
-			serialize(loader->serializer, loader->serializer.fileVersion);
 		}
 		catch(std::exception & e)
 		{
@@ -253,6 +252,9 @@ void CClient::loadGame(CGameState * initializedGameState)
 	reinitScripting();
 
 	initPlayerEnvironments();
+	
+	if(loader)
+		serialize(loader->serializer, loader->serializer.fileVersion);
 
 	initPlayerInterfaces();
 }
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -180,14 +180,15 @@ events::EventBus * CClient::eventBus() const
 	return clientEventBus.get();
 }
 
-void CClient::newGame()
+void CClient::newGame(CGameState * gameState)
 {
 	CSH->th->update();
 	CMapService mapService;
-	gs = new CGameState();
+	gs = gameState ? gameState : new CGameState();
 	gs->preInit(VLC);
 	logNetwork->trace("\tCreating gamestate: %i", CSH->th->getDiff());
-	gs->init(&mapService, CSH->si.get(), settings["general"]["saveRandomMaps"].Bool());
+	if(!gameState)
+		gs->init(&mapService, CSH->si.get(), settings["general"]["saveRandomMaps"].Bool());
 	logNetwork->trace("Initializing GameState (together): %d ms", CSH->th->getDiff());
 
 	initMapHandler();
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -180,14 +180,14 @@ events::EventBus * CClient::eventBus() const
 	return clientEventBus.get();
 }
 
-void CClient::newGame(CGameState * gameState)
+void CClient::newGame(CGameState * initializedGameState)
 {
 	CSH->th->update();
 	CMapService mapService;
-	gs = gameState ? gameState : new CGameState();
+	gs = initializedGameState ? initializedGameState : new CGameState();
 	gs->preInit(VLC);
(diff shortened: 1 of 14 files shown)


Nordsoft91:
3rd day, still failed with that...

kambala-decapitator:
> 3rd day, still failed with that...

Should maybe others take a look with fresh sight?

Nordsoft91:
Ok, it works (don’t ask how)

kambala-decapitator:
can't say it works :( tested with: macOS Intel host, iOS client.

scenario restart crashes client. Also restart of random map re-generates the map, not sure if that's right.
host logs: [Archive.tar.gz](https://github.com/vcmi/vcmi/files/9674970/Archive.tar.gz), client log: [ios-restart.txt](https://github.com/vcmi/vcmi/files/9674983/ios-restart.txt)
crash log:
```
Thread 12 Crashed:
0   vcmiclient                    	       0x100b52e24 CMapHandler::init() + 68
1   vcmiclient                    	       0x100b52e0c CMapHandler::init() + 44
2   vcmiclient                    	       0x100a06af0 CClient::initMapHandler() + 404
3   vcmiclient                    	       0x100a06898 CClient::newGame(LIB_CLIENT::CGameState*) + 596
4   vcmiclient                    	       0x1009db210 CServerHandler::startGameplay(LIB_CLIENT::CGameState*) + 104
5   libboost_thread.dylib         	       0x104255d10 boost::(anonymous namespace)::thread_proxy(void*) + 232
6   libsystem_pthread.dylib       	       0x1dbcca348 _pthread_start + 116
7   libsystem_pthread.dylib       	       0x1dbcc8948 thread_start + 8
```

game load crashes client. client log: [ios-load.txt](https://github.com/vcmi/vcmi/files/9674993/ios-load.txt), crash log:
```
Thread 12 Crashed:
0   libsystem_kernel.dylib        	       0x1bb675bbc __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x1dbcd4854 pthread_kill + 208
2   libsystem_c.dylib             	       0x18c1c46ac abort + 124
3   libc++abi.dylib               	       0x1993f5dd8 abort_message + 128
4   libc++abi.dylib               	       0x1993e6544 demangling_terminate_handler() + 276
5   libobjc.A.dylib               	       0x199303908 _objc_terminate() + 140
6   libc++abi.dylib               	       0x1993f5274 std::__terminate(void (*)()) + 16
7   libc++abi.dylib               	       0x1993f7e10 __cxa_rethrow + 144
8   vcmiclient                    	       0x10051ca28 CClient::loadGame() + 2756
9   vcmiclient                    	       0x1004ef21c CServerHandler::startGameplay(LIB_CLIENT::CGameState*) + 116
10  libboost_thread.dylib         	       0x103e65d10 boost::(anonymous namespace)::thread_proxy(void*) + 232
11  libsystem_pthread.dylib       	       0x1dbcca348 _pthread_start + 116
12  libsystem_pthread.dylib       	       0x1dbcc8948 thread_start + 8
```

Nordsoft91:
Really hard to debug such stuff

Nordsoft91:
I cannot reproduce it on localhost

Nordsoft91:
@kambala-decapitator , thanks for testing, now should work!

kambala-decapitator:
now restart and load game work! great job @Nordsoft91 !

client/Client.cpp
```diff
@@ -180,14 +180,15 @@ events::EventBus * CClient::eventBus() const
 	return clientEventBus.get();
 }
 
-void CClient::newGame()
+void CClient::newGame(CGameState * gameState)
 {
 	CSH->th->update();
 	CMapService mapService;
-	gs = new CGameState();
+	gs = gameState ? gameState : new CGameState();
```
kambala-decapitator, line 187: maybe name `gameState` something like `currentGameState` or `existingGameState`?

kambala-decapitator (approved):
looks fine to me, but someone else should also review

client/Client.cpp
nullkiller: Really it does not add any more meaning if it is that gameState parameter in load(). I can only suggest gameStateToLoad. Still looks weird. Why can not we initiate reloading of the same map?

nullkiller: Naming - gameStateToLoad
Logic - Why we decided to pass gamestate via network? To avoid random generator desync issues?

Nordsoft91: in regards of naming: newGame(gameState) is quite understandable that new game will be started using gamestate from the argument, isn't it?

kambala-decapitator: we're talking about method parameter name here, not about the calling side :)

kambala-decapitator:
when you open Player Settings, configuration UI doesn't appear until you switch number of players in the combobox, even if initial value is correct

nullkiller (changes requested):
Can not combine stacks of the same creature type. Started campaign, saved, loaded, opened hero 
split stack and combine it back - error message appears complaining about different creature

lib/mapping/CMap.h
```diff
@@ -424,6 +424,7 @@ class DLL_LINKAGE CMap : public CMapHeader
 	void serialize(Handler &h, const int formatVersion)
 	{
 		h & static_cast<CMapHeader&>(*this);
+		h & triggeredEvents; //from CMapHeader
```
nullkiller, line 427: Should we increase minimum serialization version?

Nordsoft91: We already do it for 1.1
I don't think we need to increase each build where we break compatibility.

PR #1003, opened 2022-09-26
compile Version.cpp only when ENABLE_GITVERSION is ON

### LINKED ISSUE #883: ENABLE_GITVERSION=OFF doesn't work
In `CMakeLists.txt` setting `ENABLE_GITVERSION=OFF` causes cmake not to generate `Version.cpp` file. But in `lib/CMakeLists.txt` this file is listed in `lib_SRCS` unconditionally.

kambala-decapitator:
fixes #883

applied the patch by @josch from the linked issue

### DIFF
--- a/cmake_modules/VCMI_lib.cmake
+++ b/cmake_modules/VCMI_lib.cmake
@@ -2,7 +2,6 @@ macro(add_main_lib TARGET_NAME LIBRARY_TYPE)
 	set(MAIN_LIB_DIR "${CMAKE_SOURCE_DIR}/lib")
 	set(lib_SRCS
 		${MAIN_LIB_DIR}/StdInc.cpp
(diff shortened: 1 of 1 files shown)


josch:
@kambala-decapitator Thanks for making a pull request out of this! If you like, you can fix my name to:

    Johannes Schauer Marin Rodrigues <josch@mister-muffin.de>

I got married and my last name changed. If you tell me where you found my old name, then it'd give me an opportunity to fix the old data there. Thanks!

kambala-decapitator:
just pushed commit with your name corrected.

> If you tell me where you found my old name, then it'd give me an opportunity to fix the old data there. Thanks!

I didn't see your name in the GitHub profile, so simply took it from vcmi history :)

PR #1005, opened 2022-09-27
raise iOS deployment target to 12.0

kambala-decapitator:
This is done for:

1. Future Conan support, as it allows building Qt 5.15.x only for iOS 12 without custom patches
2. Reducing CI build time, as only 64-bit devices can target iOS 11 and later (currently we build both 32- and 64-bit)
3. Amount of devices capable of running iOS < 12 (that is, iPhones before 6 and iPads before Air / Mini 2, see [chart](https://blakespot.com/ios_device_specifications_grid.html)) is quite tiny. Anyone willing to run on iOS < 12 can build from source, as current dependencies target iOS 10.0. We can also simply make 2 builds in parallel if desired (e.g. only for release builds).
4. This will allow us to produce builds using the latest available iOS toolchain - Xcode 14 - as it is no longer able to produce armv7 (32-bit) code

Also a few unused entities are removed in the PR.


Nordsoft91:
Does it mean that there will be 64 bit support only? If yes, will it be possible to have 64 bit only as default but keep 32bit to build ad-hoc (not in ci, but locally)

Nordsoft91 (approved):

kambala-decapitator:
> Does it mean that there will be 64 bit support only? If yes, will it be possible to have 64 bit only as default but keep 32bit to build ad-hoc (not in ci, but locally)

yes, correct:

> Anyone willing to run on iOS < 12 can build from source, as current dependencies target iOS 10.0

Nordsoft91:
@nullkiller , can we merge this ASAP? That's really needed to debug multiplayer because now I have to wait 1.5 hours to get builds

PR #1006, opened 2022-09-27
Small misc fixes

AgostonSzepessy:
- Replace a 3 element vector doing a comparison with a tuple because those are much lighter, especially for just a single compare.
- Update README.md to point to GitHub issues instead of the old issue tracker

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ VCMI is work-in-progress attempt to recreate engine for Heroes III, giving it ne
  * Homepage:   https://vcmi.eu/
  * Wiki:       https://wiki.vcmi.eu/
  * Forums:     https://forum.vcmi.eu/
- * Bugtracker: https://bugs.vcmi.eu/
+ * Bugtracker: https://github.com/vcmi/vcmi/issues
  * Slack:      https://slack.vcmi.eu/
 
 ## Installation guides
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -840,11 +840,10 @@ CStackBasicDescriptor CGHeroInstance::calculateNecromancy (const BattleResult &b
 				}
 				else
 				{
-					auto quality = [getCreatureID](std::shared_ptr<Bonus> pick) -> std::vector<int>
+					auto quality = [getCreatureID](std::shared_ptr<Bonus> pick) -> std::tuple<int, int, int>
 					{
 						const CCreature * c = VLC->creh->objects[getCreatureID(pick)];
-						std::vector<int> v = {c->level, static_cast<int>(c->cost.marketValue()), -pick->additionalInfo[1]};
-						return v;
+						return std::tuple<int, int, int> {c->level, static_cast<int>(c->cost.marketValue()), -pick->additionalInfo[1]};
 					};
 					if(quality(topPick) < quality(newPick))
 						topPick = newPick;


README.md
```diff
@@ -9,7 +9,7 @@ VCMI is work-in-progress attempt to recreate engine for Heroes III, giving it ne
  * Homepage:   https://vcmi.eu/
  * Wiki:       https://wiki.vcmi.eu/
  * Forums:     https://forum.vcmi.eu/
- * Bugtracker: https://bugs.vcmi.eu/
+ * Bugtracker: https://github.com/vcmi/vcmi/issues
```
Nordsoft91, line 12: There is incorrect link to bugtracker fromm the website too, would be good to fix it too https://github.com/vcmi/VCMI.eu

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -840,11 +840,10 @@ CStackBasicDescriptor CGHeroInstance::calculateNecromancy (const BattleResult &b
...
 				{
-					auto quality = [getCreatureID](std::shared_ptr<Bonus> pick) -> std::vector<int>
+					auto quality = [getCreatureID](std::shared_ptr<Bonus> pick) -> std::tuple<int, int, int>
 					{
 						const CCreature * c = VLC->creh->objects[getCreatureID(pick)];
-						std::vector<int> v = {c->level, static_cast<int>(c->cost.marketValue()), -pick->additionalInfo[1]};
-						return v;
+						return std::tuple<int, int, int> {c->level, static_cast<int>(c->cost.marketValue()), -pick->additionalInfo[1]};
 					};
 					if(quality(topPick) < quality(newPick))
```
Nordsoft91, line 848: I don't understand, how it works? and how does it work now?

nullkiller: Later we compare two tuples, before it was 2 vectors

Nordsoft91: But operator < is not defined for vectors or I don't know something?

nullkiller: https://en.cppreference.com/w/cpp/container/vector/operator_cmp

README.md
AgostonSzepessy: Good point. I'll open a separate PR for that

lib/mapObjects/CGHeroInstance.cpp
Nordsoft91: according to that, it's gone since c++20. It make sense to replace deprecated stuff (maybe not in this pr but still)

Nordsoft91: as for this pr, std::array<int, 3> can be used instead of tuple since there is no type difference

AgostonSzepessy: We can leave it as a std::tuple. https://stackoverflow.com/questions/48445291/stdtuple-faster-than-stdarray I based it off this which says std::tuple has better perf.

Also, std::vector still has comparison operators as of C++20. They're just implemented using the spaceship operator instead.

nullkiller: As far as I understood it is not gone, just replaced, or changed signature. But yes no objection to replace vector with tuple. Btw is tuple size limited somehow? I mean can it contain 20 items?

AgostonSzepessy: Tuple is not size-limited. It's implemented as a variadic template and the size is determined at compile-time so you can add as much stuff as you want.

For vector, the compare operation can still be performed, but the `operator<` has been replaced with `operator<=>`. More info on `<=>` here: https://devblogs.microsoft.com/cppblog/simplify-your-code-with-rocket-science-c20s-spaceship-operator/

Nordsoft91 (approved):

PR #1017, opened 2022-09-29
Fix incorrect templates for mines in HoTA

DjWarmonger:
Long overdue fix for https://github.com/vcmi/vcmi/issues/897

### DIFF
--- a/lib/mapObjects/CommonConstructors.h
+++ b/lib/mapObjects/CommonConstructors.h
@@ -33,18 +33,10 @@ protected:
 		auto obj = new ObjectType();
 		preInitObject(obj);
 
+		//Set custom template or leave null
 		if (tmpl)
 		{
 			obj->appearance = tmpl;
-		}
-		else
-		{
-			auto templates = getTemplates();
-			if (templates.empty())
-			{
-				throw std::runtime_error("No handler for created object");
-			}
-			obj->appearance = templates.front(); //just any template for now, will be initialized later
 		}
 
 		return obj;
--- a/lib/rmg/RmgObject.cpp
+++ b/lib/rmg/RmgObject.cpp
@@ -105,20 +105,26 @@ void Object::Instance::setPositionRaw(const int3 & position)
 	dObject.pos = dPosition + dParent.getPosition();
 }
 
-void Object::Instance::setTemplate(const TerrainId & terrain)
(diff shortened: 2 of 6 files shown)


lib/mapObjects/CommonConstructors.h
```diff
@@ -33,19 +33,11 @@ class CDefaultObjectTypeHandler : public AObjectTypeHandler
 		auto obj = new ObjectType();
 		preInitObject(obj);
 
+		//Set custom template or leave null
 		if (tmpl)
 		{
 			obj->appearance = tmpl;
 		}
```
Nordsoft91, line 40:
```suggestion
		obj->appearance = tmpl;
```

lib/rmg/RmgObject.cpp
```diff
@@ -292,17 +313,6 @@ void Object::Instance::finalize(RmgMap & map)
 			throw rmgException(boost::to_string(boost::format("Tile %s of object %d at %s is outside the map") % tile.toString() % dObject.id % dObject.pos.toString()));
 	}
 	
-	if (dObject.appearance->id == Obj::NO_OBJ)
```
Nordsoft91, line 295: Shouldn't we check for nullptr here?

Nordsoft91 (changes requested):
Please remove huge piece of copy-pasted code from BinaryDeserializer.h (void load(std::shared_ptr<const T> &data) introduced with the same PR as this bug

lib/rmg/RmgObject.cpp
DjWarmonger: I do it in line 292

lib/mapObjects/CommonConstructors.h
DjWarmonger: Actually it would change logic. Now if tmpl is nullptr (default), the original template, if existing, is not modified.

DjWarmonger:
This "duplicated code" is neccessary to serialize const pointers.

nullkiller:
Sure it is needed for whatever reason. Still it is duplicating code. Can we reuse one method from another? Or leave only const version? We should be able to use it in non-const scenario or am I missing something?

nullkiller:
I will take a look onto this issue and try to fix serialization

Nordsoft91 (approved):

PR #1022, opened 2022-09-30
[Apple] set proper app and build versions

### LINKED ISSUE #1002: store build version in Info.plist

kambala-decapitator:
All platforms: build version is displayed in launcher now.

macOS: app version is now visible in bundle details / quick look.

iOS: now it's easier to see which version you have installed on your device.
In system settings:
Screen Shot 2022-09-30 at 12 50 28

In Xcode:
Screenshot 2022-09-30 at 12 31 50

fixes #1002


Nordsoft91:
Do we need date in the version? Should sha be not enough?

kambala-decapitator:
> Do we need date in the version? Should sha be not enough?

it may be useful, e.g. to see immediately how old is your build

lib/Terrain.cpp
```diff
@@ -397,30 +397,6 @@ TerrainType::TerrainType(const std::string& _name):
 	transitionRequired(false)
 {
 }
-
-TerrainType& TerrainType::operator=(const TerrainType & other)
```
nullkiller, line 401: Do not need it anymore? Or is it autogenerated thing?

kambala-decapitator: Yes, compiler will generate exactly the same, as here simply all fields are copied. Besides, classic implementation of = must have protection from assigning an object to itself.

PR #1023, opened 2022-09-30
unlock adventure map UI after closing battle result dialog

### LINKED ISSUE #966: Server errors after auto-battle
1. Start any battle
2. Press auto-battle (or press A)
3. as soon as you win the battle, press Enter

Expected: window was closed
Actual: hero info window appears, it's possible to do strange things there

kambala-decapitator:
fixes #966

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -930,7 +930,6 @@ void CPlayerInterface::battleEnd(const BattleResult *br)
 		isAutoFightOn = false;
 		cb->unregisterBattleInterface(autofightingAI);
 		autofightingAI.reset();
-		adventureInt->quickCombatUnlock();
 
 		if (!battleInt)
 		{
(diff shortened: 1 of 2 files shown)


client/CPlayerInterface.cpp
```diff
@@ -930,21 +930,22 @@ void CPlayerInterface::battleEnd(const BattleResult *br)
 		isAutoFightOn = false;
 		cb->unregisterBattleInterface(autofightingAI);
 		autofightingAI.reset();
-		adventureInt->quickCombatUnlock();
 
 		if (!battleInt)
```
nullkiller, line 934: Do we need to call it on else? Consider regular battle and --headless mode when there is no UI at all

nullkiller: So I expect that in case of headless it will not even rich this place so this can be ignored. Still before we called unblock always and now only if battleInt not present

nullkiller: I see. we call it later as well

PR #1028, opened 2022-10-01
Launcher fixes2

DjWarmonger:
Rebase of #822

- Fixed resizing of toolbar: https://bugs.vcmi.eu/view.php?id=2728
- Fixed icon update when a sub-mod is enabled or disabled: https://bugs.vcmi.eu/view.php?id=3217

### DIFF
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -91,8 +91,6 @@ bool CModEntry::isEnabled() const
 
 bool CModEntry::isDisabled() const
 {
-	if(!isInstalled())
-		return false;
 	return !isEnabled();
 }
 
--- a/launcher/mainwindow_moc.cpp
+++ b/launcher/mainwindow_moc.cpp
@@ -80,9 +80,9 @@ MainWindow::MainWindow(QWidget * parent)
 
 	ui->tabListWidget->setCurrentIndex(0);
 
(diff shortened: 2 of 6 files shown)


kambala-decapitator (approved):

nullkiller (changes requested):
image
red signs should be only for installed disabled mods. This works correct now but is broken in this branch.

launcher/modManager/cmodlist.cpp
```diff
@@ -91,8 +91,6 @@ bool CModEntry::isEnabled() const
 
 bool CModEntry::isDisabled() const
 {
-	if(!isInstalled())
-		return false;
```
kambala-decapitator, line 95: seems this change broke existing behavior

nullkiller:
More over pressing <<Enable>>causes error message

DjWarmonger:
@nullkiller How do I actually get that long list of mods to download?

kambala-decapitator:
> @nullkiller How do I actually get that long list of mods to download?

add https://raw.githubusercontent.com/vcmi/vcmi-mods-repository/develop/beta.json to the repositories list

nullkiller:
new vcpkg is already here with working TLS

DjWarmonger:
Oh, I'll need this too.

Alexander-Wilms:
Both bugs are fixed by now, I think the MR can be closed.

With a screen resolution of 800x600:

image

PR #1033, opened 2022-10-01
don't add Necromancy to Witch Hut on random maps

### LINKED ISSUE #934: By default on random maps Necromancy can't be learned from Witch Huts.
...but!
In vcmi can be learned.
image

kambala-decapitator:
The fix seems to work: tested by starting/loading existing scenarios to ensure that this code path isn't called, also created random map and visited witch hut with inverted condition to ensure that I receive Necromancy. But I saw there's another possibility although seems more complicated: create witch hut "specialisation" by inheriting from `CDefaultObjectTypeHandler` similarly to e.g. `CTownInstanceConstructor`.

fixes #934

### DIFF
--- a/lib/mapObjects/CObjectClassesHandler.cpp
+++ b/lib/mapObjects/CObjectClassesHandler.cpp
@@ -58,7 +58,6 @@ CObjectClassesHandler::CObjectClassesHandler()
 	SET_HANDLER("randomDwelling", CGDwelling);
 
 	SET_HANDLER("generic", CGObjectInstance);
-	SET_HANDLER("market", CGMarket);
 	SET_HANDLER("cartographer", CCartographer);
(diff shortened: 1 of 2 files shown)


lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -58,7 +58,6 @@ CObjectClassesHandler::CObjectClassesHandler()
 	SET_HANDLER("randomDwelling", CGDwelling);
 
 	SET_HANDLER("generic", CGObjectInstance);
-	SET_HANDLER("market", CGMarket);
```
nullkiller, line 61: Why is that?

nullkiller: Duplicate. I see

PR #1034, opened 2022-10-01
restore last used address in Multiplayer join dialog

### LINKED ISSUE #1004: don't suggest localhost address when joining MP
rather have empty string or restore last used address

kambala-decapitator:
- in all other cases (when machine is host) hardcoded localhost address is displayed
- the very first time also localhost address is shown because it's default setting

fixes #1004

### DIFF
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -59,6 +59,8 @@
 
 template<typename T> class CApplyOnLobby;
 
+const std::string CServerHandler::localhostAddress{"127.0.0.1"};
+
 #ifdef VCMI_ANDROID
 extern std::atomic_bool androidTestServerReadyFlag;
 #endif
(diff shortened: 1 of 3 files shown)


client/CServerHandler.cpp
```diff
@@ -243,7 +245,7 @@ void CServerHandler::startLocalServerAndConnect()
 #else
 	const ui16 port = 0;
 #endif
-	justConnectToServer(settings["server"]["server"].String(), port);
+	justConnectToServer(localhostAddress, port);
```
nullkiller, line 248: Why localhost? Where do we read settings["server"]["server"] now?

nullkiller: Local games?

kambala-decapitator: yes, check the method name ;)

PR #1039, opened 2022-10-02
Modinfo from json

Nordsoft91:


launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -228,6 +233,32 @@ QString CModListView::genModInfoText(CModEntry & mod)
...
+	if(mod.isCompatible())
+		result += compatibleString.arg(tr("Compatibility"));
+	else
+	{
+		auto compatibilityInfo = mod.getValue("compatibility").toMap();
+		auto minStr = compatibilityInfo.value("min").toString();
+		auto maxStr = compatibilityInfo.value("max").toString();
+
+		result += incompatibleString.arg(tr("Compatibility"));
+		if(minStr == maxStr)
```
nullkiller, line 246: We will use this section if both min and max are empty, right?

nullkiller: Or is it impossible because we will go if(mod.isCompatible()) way?

kambala-decapitator (approved):

launcher/modManager/cmodlistview_moc.cpp
nullkiller: I see checks on minStr.empty() and maxStr.empty() if both are empty minStr == maxStr. But I expect if both are empty we can not have mod being incompatible because it has no restrictions specified. This is what I want to confirm)

Nordsoft91: Yes, for both empty or other situation if mod is compatible, those checks aren't executed

Issue #1024, opened 2022-09-30
Host crash if client disconnected


Nordsoft91:
1. Start multiplayer host-client
2. Client hero must be in the town but not in harrison!
3. Force end client application

Expected result: host observes message that client player lost
Actual result: host crash

PR #1052, opened 2022-10-04
MP handle disconnection

Nordsoft91:
Desired behavior:
- client intentionally disconnected - AI takes control over the player(s)
- client accidentally disconnected - continue game if turn not under client control. If player turn, suggest host to kick the client from the session

ToDo:
- [ ] verify client by uuid
- [ ] move pop up messages to translate.json
- [ ] get session ip and port from stored settings
- [ ] fix transferring ownership to AI
- [ ] make server notifications about client connected/disconnected
- [ ] check and support multiple players on single client
- [ ] implement kick player suggestion
- [ ] allow game save & exit if client was disconnected on its turn
- [ ] implement notification about refusing to restore session
- [ ] implement timeout or another solution if server isn't available while session recovery
- [ ] implement host transferring (out of scope?)

### DIFF
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -347,11 +347,21 @@ void PlayerEndsGame::applyCl(CClient *cl)
 
 void PlayerReinitInterface::applyCl(CClient * cl)
 {
-	CSH->si->getIthPlayersSettings(player).connectedPlayerIDs.clear();
-	cl->initPlayerEnvironments();
-	cl->initPlayerInterfaces();
-	if(cl->gameState()->currentPlayer == player)
+	auto & plSettings = CSH->si->getIthPlayersSettings(player);
+	if(!playerConnectionId)
+	{
+		plSettings.connectedPlayerIDs.clear();
+		cl->initPlayerEnvironments();
+		cl->initPlayerInterfaces();
+		if(cl->gameState()->currentPlayer == player)
+			callOnlyThatInterface(cl, player, &CGameInterface::yourTurn);
+	}
+	else
+	{
+		plSettings.connectedPlayerIDs.insert(playerConnectionId);
+		callAllInterfaces(cl, &IGameEventsReceiver::playerStartsTurn, player);
 		callOnlyThatInterface(cl, player, &CGameInterface::yourTurn);
+	}
 }
 
 void RemoveBonus::applyCl(CClient *cl)
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -359,7 +359,9 @@ void PlayerReinitInterface::applyCl(CClient * cl)
 	else
 	{
 		plSettings.connectedPlayerIDs.insert(playerConnectionId);
+		cl->initPlayerInterfaces();
 		callAllInterfaces(cl, &IGameEventsReceiver::playerStartsTurn, player);
+		//if(cl->gameState()->currentPlayer == player)
 		callOnlyThatInterface(cl, player, &CGameInterface::yourTurn);
 	}
 }
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -356,13 +356,14 @@ void PlayerReinitInterface::applyCl(CClient * cl)
 		if(cl->gameState()->currentPlayer == player)
 			callOnlyThatInterface(cl, player, &CGameInterface::yourTurn);
 	}
-	else
+	else if(playerConnectionId == CSH->c->connectionID)
 	{
 		plSettings.connectedPlayerIDs.insert(playerConnectionId);
+		cl->playerint.clear();
 		cl->initPlayerInterfaces();
-		callAllInterfaces(cl, &IGameEventsReceiver::playerStartsTurn, player);
-		//if(cl->gameState()->currentPlayer == player)
-		callOnlyThatInterface(cl, player, &CGameInterface::yourTurn);
+		auto currentPlayer = cl->gameState()->currentPlayer;
+		callAllInterfaces(cl, &IGameEventsReceiver::playerStartsTurn, currentPlayer);
+		callOnlyThatInterface(cl, currentPlayer, &CGameInterface::yourTurn);
 	}
 }
 
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -345,6 +345,15 @@ void PlayerEndsGame::applyCl(CClient *cl)
 		handleQuit(settings["session"]["spectate"].Bool()); // if spectator is active ask to close client or not
 }
 
+void PlayerReinitInterface::applyCl(CClient * cl)
+{
+	CSH->si->getIthPlayersSettings(player).connectedPlayerIDs.clear();
+	cl->initPlayerEnvironments();
+	cl->initPlayerInterfaces();
+	if(cl->gameState()->currentPlayer == player)
+		callOnlyThatInterface(cl, player, &CGameInterface::yourTurn);
+}
+
 void RemoveBonus::applyCl(CClient *cl)
 {
 	cl->invalidatePaths();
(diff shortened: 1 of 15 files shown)


Nordsoft91:
@kambala-decapitator , yes, exactly, still have pending features to do, but we need to check for regression. I'll address code review changes and then suggest to merge for active testing by community

client/NetPacksClient.cpp
```diff
@@ -345,6 +345,32 @@ void PlayerEndsGame::applyCl(CClient *cl)
...
+			callOnlyThatInterface(cl, currentPlayer, &CGameInterface::yourTurn);
+		}
+		else if(playerConnectionId == CSH->c->connectionID)
+		{
+			plSettings.connectedPlayerIDs.insert(playerConnectionId);
+			cl->playerint.clear();
+			cl->initPlayerInterfaces();
+			auto currentPlayer = cl->gameState()->currentPlayer;
+			callAllInterfaces(cl, &IGameEventsReceiver::playerStartsTurn, currentPlayer);
+			callOnlyThatInterface(cl, currentPlayer, &CGameInterface::yourTurn);
```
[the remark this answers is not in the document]
Nordsoft91: What would you recommend? Lambda?

server/CGameHandler.cpp
```diff
@@ -5003,12 +5008,61 @@ bool CGameHandler::makeBattleAction(BattleAction &ba)
...
+	if(isHost && cheat.size() >= 2 && cheat[0] == "game")
+	{
+		if(cheat[1] == "exit" || cheat[1] == "quit" || cheat[1] == "end")
+		{
+			SystemMessage temp_message("game was terminated");
+			sendAndApply(&temp_message);
+			lobby->state = EServerState::SHUTDOWN;
+			return;
+		}
+		if(cheat.size() == 3 && cheat[1] == "save")
```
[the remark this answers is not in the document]
Nordsoft91: not sure if it's correct. So last words will be just ignored. For me it's better to ignore whole senates thus

server/CGameHandler.cpp
```diff
@@ -5003,12 +5008,61 @@ bool CGameHandler::makeBattleAction(BattleAction &ba)
...
+			SystemMessage temp_message("game saved as " + cheat[2]);
+			sendAndApply(&temp_message);
+			return;
+		}
+		if(cheat.size() == 3 && cheat[1] == "kick")
+		{
+			auto playername = cheat[2];
+			PlayerColor playerToKick(PlayerColor::CANNOT_DETERMINE);
+			if(std::all_of(playername.begin(), playername.end(), ::isdigit))
+				playerToKick = PlayerColor(std::stoi(playername));
```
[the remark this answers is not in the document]
Nordsoft91: no, search is performed by player color, not by player name. So we have hardcoded consts, which are used here ("red", "blue", etc). Referring to PlayerColor::getStr(bool)

server/CVCMIServer.cpp
```diff
@@ -489,23 +492,77 @@ void CVCMIServer::clientConnected(std::shared_ptr<CConnection> c, std::vector<st
...
 	connections -= c;
+	if(connections.empty())
+		throw std::runtime_error("No more connections. Closing server.");
+	
+	if(hostClient == c)
+	{
+		//TODO: support host transfer role
+		state = EServerState::SHUTDOWN;
+		return;
+	}
```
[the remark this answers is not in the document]
Nordsoft91: I'm suspecting that some bug can be related to that. I'll investigate later, while bug hunting

client/NetPacksClient.cpp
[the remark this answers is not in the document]
kambala-decapitator: yes, or just a private method. Also I see that this piece of code doesn't depend on the loop variable => lambda should be outside of the loop.

server/CGameHandler.cpp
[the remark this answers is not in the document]
kambala-decapitator: I was more referring to the condition. If player's name consists only of digits, then attempt to kick by color will be performed which is incorrect. Example: name `123`.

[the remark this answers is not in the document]
Nordsoft91: it's always kick by color. You can enter kick 0 or kick red and those will be equivalent. There is no kick by player name.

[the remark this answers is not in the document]
kambala-decapitator: well, in the `else` branch player name is searched :)

Or is `playername` variable actually a color name?.. That's quite misleading then.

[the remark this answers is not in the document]
Nordsoft91: It's a bit blurry, agree. The idea behind is to support kick by actual player name in the future but for now it's player colour.

client/NetPacksClient.cpp
```diff
@@ -345,6 +345,34 @@ void PlayerEndsGame::applyCl(CClient *cl)
...
+		cl->initPlayerInterfaces();
+		auto currentPlayer = cl->gameState()->currentPlayer;
+		callAllInterfaces(cl, &IGameEventsReceiver::playerStartsTurn, currentPlayer);
+		callOnlyThatInterface(cl, currentPlayer, &CGameInterface::yourTurn);
+	};
+	
+	for(auto player : players)
+	{
+		auto & plSettings = CSH->si->getIthPlayersSettings(player);
+		if(playerConnectionId == PlayerSettings::PLAYER_AI)
```
nullkiller, line 361: Can the host which started server reconnect?

client/NetPacksLobbyClient.cpp
```diff
@@ -133,6 +141,9 @@ bool LobbyUpdateState::applyOnLobbyHandler(CServerHandler * handler)
 
 void LobbyUpdateState::applyOnLobbyScreen(CLobbyScreen * lobby, CServerHandler * handler)
 {
+	if(!lobby) //stub: ignore message for game mode
```
nullkiller, line 144: Does it mean you want to fix it later or is it kind of hack?

lib/NetPacksLib.cpp
```diff
@@ -365,6 +365,19 @@ DLL_LINKAGE void PlayerEndsGame::applyGs(CGameState *gs)
 	}
 }
 
+DLL_LINKAGE void PlayerReinitInterface::applyGs(CGameState *gs)
+{
+	if(!gs || !gs->scenarioOps)
+		return;
+	
+	//TODO: what does mean if more that one player connected?
```
nullkiller, line 373: As far as I understood each connection can have a few players in hot-seat mode

lib/mapObjects/CGTownInstance.h
```diff
@@ -349,7 +349,7 @@ class DLL_LINKAGE CGTownInstance : public CGDwelling, public IShipyard, public I
 
 	inline bool isBattleOutsideTown(const CGHeroInstance * defendingHero) const
 	{
-		return defendingHero && garrisonHero && defendingHero != garrisonHero;
```
nullkiller, line 352: Are you sure about it? If garrisonHero is nullptr this must return false

nullkiller (approved):

client/NetPacksLobbyClient.cpp
Nordsoft91: It means, that there is no lobby screen during gameplay, so message shall be silently ignored

lib/NetPacksLib.cpp
Nordsoft91: that should work. The scenario described here is when several connections are related to the one player. I can't imagine such scenario but looking through the code it seems possible

lib/mapObjects/CGTownInstance.h
Nordsoft91: I'm not. but otherwise game crashes (https://github.com/vcmi/vcmi/issues/1024). That's sad if it's wrong because will be tough to fix that

nullkiller: The idea is that if two heroes are in the town - garrison and visiting- first battle goes outside town but if not - inside town

nullkiller: Even if visiting hero is not able to enter garrison because of extra units there but there is no hero - anyway battle goes as siege and in some cases these extra units just disappear.

Nordsoft91: reverted

PR #1068, opened 2022-10-14
Nk retreat

nullkiller:
Inspired by: https://github.com/vcmi/vcmi/pull/442
Nullkiller AI can now retreat and even buy hero back. The logic is very simple for now. Only check if our strength is way smaller than enemy one

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -1401,7 +1401,7 @@ void AIGateway::tryRealize(Goals::Trade & g) //trade
 const CGTownInstance * AIGateway::findTownWithTavern() const
 {
 	for(const CGTownInstance * t : cb->getTownsInfo())
-		if(t->hasBuilt(BuildingID::TAVERN) && !t->visitingHero)
+		if(t->hasBuilt(BuildingID::TAVERN) && (!t->visitingHero || !t->garrisonHero))
 			return t;
 
 	return nullptr;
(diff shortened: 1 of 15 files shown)


nullkiller:
> I think, it`s bettern teach AI to surrender - if retreat is the only option and surrender cost + buyback cost is about 10-25% of traesury (depending on hero role in surrent situation), then use surrender to preserve units.

Not everything at once) Sure I will consider surrendering later.

AI/Nullkiller/AIGateway.cpp
```diff
@@ -488,6 +489,24 @@ void AIGateway::showWorldViewEx(const std::vector<ObjectPosInfo> & objectPositio
...
+boost::optional<BattleAction> AIGateway::makeSurrenderRetreatDecision(
+	const BattleStateInfoForRetreat & battleState)
+{
+	LOG_TRACE(logAi);
+	NET_EVENT_HANDLER;
+
+	double fightRatio = battleState.getOurStrength() / (double)battleState.getEnemyStrength();
+
+	// if we have no towns - things are already bad, so retreat is not an option.
+	if(cb->getTownsInfo().size() && fightRatio < 0.3 && battleState.canFlee)
```
Nordsoft91, line 501: 1) 0.3 is a magic const
2) if enemy hero has a 1 fast unit, he still can throw magic arrow. I suggest do make heuristic smarter and check if AI make magic action before surrender

nullkiller: Spellcast is evaluated before retreat so I expect AI should try it first. 

2 - Smarter things are always welcome! You can check the PR from description. It went exactly this path and actually is smarter than mine. It can detect when it is last turn before hero dies. But for whatever reason it was not merged. Guess why? I can - to many things at once. More code - more bugs - more time to fix - higher chance for PR to be abandoned.
AI code is never ideal, always there is a room for improvement, always in some cases decisions will be not optimal. If you try to fix all it at once - it will just never end.

Nordsoft91: I agree that approach "better than now" is what we should target for. 
But are your sure this behavior is not possible?
1) enemy hero attacks player hero
2) enemy hero immediately retreats

nullkiller: For NKAI it is not allowed to attack if power ratio is below safe_attack constant which is higher than 1 as far as I remember.
Also AI prefer attacking when his power is actually higher

nullkiller: I will fix magic constant in battle ai PR

PR #1073, opened 2022-10-16
[Conan] add iOS support, bump dependencies, improvements

kambala-decapitator:
Pending tasks:

- [x] make sure that dylib symlinks aren't created when creating iOS deps
- [x] update iOS/macOS wiki to point to new docs


docs/conan.md
```diff
@@ -0,0 +1,147 @@
...
+1. [Install Conan](https://docs.conan.io/en/latest/installation.html)
+2. Execute in terminal: `conan profile new default --detect`
+
+## Download dependencies
+
+0. If your platform is not on the list of supported ones or you don't want to use our prebuilt binaries, you can still build dependencies from source or try consuming prebuilt binaries from the central Conan repository - [ConanCenter](https://conan.io/center/). In this case skip to the next section directly.
+
+1. Check if your build environment can use the prebuilt binaries: basically, that your compiler version (or Xcode major version) matches the information below. If you're unsure, just skip to next step.
+
+    - macOS: libraries are built with Apple clang 13 (Xcode 13.4.1), should be consumable by Xcode and Xcode CLT 13.x
```
Nordsoft91, line 23: I use Xcode 11, does it mean that I cannot just download necessary reps?

docs/conan.md
```diff
@@ -0,0 +1,147 @@
...
+## Getting started
+
+1. [Install Conan](https://docs.conan.io/en/latest/installation.html)
+2. Execute in terminal: `conan profile new default --detect`
+
+## Download dependencies
+
+0. If your platform is not on the list of supported ones or you don't want to use our prebuilt binaries, you can still build dependencies from source or try consuming prebuilt binaries from the central Conan repository - [ConanCenter](https://conan.io/center/). In this case skip to the next section directly.
+
+1. Check if your build environment can use the prebuilt binaries: basically, that your compiler version (or Xcode major version) matches the information below. If you're unsure, just skip to next step.
```
Nordsoft91, line 21: "If you're unsure, just skip to next step"
Which step should I skip? In that step there is no actions as I get it

docs/conan.md
```diff
@@ -0,0 +1,147 @@
...
+    - [macOS](https://github.com/vcmi/vcmi-deps-macos/releases/latest): pick **intel.txz** if you have Intel Mac, otherwise - **intel-cross-arm.txz**
+    - [iOS](https://github.com/vcmi/vcmi-ios-deps/releases/latest)
+
+3. Only if you have Apple Silicon Mac and trying to build for macOS or iOS: follow [instructions how to build Qt host tools for Apple Silicon](https://github.com/vcmi/vcmi-ios-deps#note-for-arm-macs), on step 3 copy them to `~/.conan/data/qt/5.15.x/_/_/package/SOME_HASH/bin` (`5.15.x` and `SOME_HASH` are placeholders).
+
+## Generate CMake integration
+
+Conan needs to generate CMake toolchain file to make dependencies available to CMake. See `CMakeDeps` and `CMakeToolchain` [in the official documentation](https://docs.conan.io/en/latest/reference/conanfile/tools/cmake.html) for details.
+
+In terminal `cd` to the VCMI source directory and run the following command. If you want to download prebuilt binaries from ConanCenter, also read the next section.
```
Nordsoft91, line 37: "also read the next section"
Which section? In which case should I not read it?

docs/conan.md
```diff
@@ -0,0 +1,147 @@
...
+  --install-folder=<b><i>conan-generated</i></b> \
+  --no-imports \
+  --build=<b><i>never</i></b> \
+  --profile:build=default \
+  --profile:host=<b><i>CI/conan/PROFILE</i></b>
+</pre>
+
+The highlighted parts can be adjusted:
+
+- ***conan-generated***: directory where the generated files will appear. This value is used in CMake presets from VCMI, but you can actually use any directory and override it in your local presets.
```
Nordsoft91, line 50: please, specify is it relative path or absolute? Probably should be some recommendation (assuming we and in VCMI sources repo, does it make sense to advice something like ../vcmi-deps?)

docs/conan.md
kambala-decapitator: Correct. For iOS you can use legacy pre-built deps, for macos - from brew. Or build from source for any platform.
In the future this can change of course.

docs/conan.md
```diff
@@ -0,0 +1,147 @@
...
+- ***never***: use this value to avoid building any dependency from source. You can also use `missing` to build recipes, that are not present in your local cache, from source.
+- ***CI/conan/PROFILE***: if you want to consume our prebuilt binaries, ***PROFILE*** must be replaced with one of filenames from our [Conan profiles directory](../CI/conan) (determining the right file should be straight-forward). Otherwise, either select one of our profiles or replace ***CI/conan/PROFILE*** with `default` (your default profile).
```
Nordsoft91, line 52: "if you want to consume our rebuild binaries"
can you specify the cases why I may want to consume them?

"determining the right file should be straight-forward"
suggest to specify possible options, e.g. macos-arm, macos-intel

docs/conan.md
```diff
@@ -0,0 +1,147 @@
...
+If you also want to build the missing binaries from source, use `--build=missing` instead of `--build=never`.
+
+```
+conan install . \
+  --install-folder=~/my-dir \
+  --no-imports \
+  --build=never \
+  --profile:build=default \
+  --profile:host=default \
+  -o default_options_of_requirements=True
```
Nordsoft91, line 105:
```
ERROR: /Users/admin/Documents/Projects/vcmi/vcmi/conanfile.py: option 'default_options_of_requirements' doesn't exist
Possible options are []
```

docs/conan.md
kambala-decapitator: Next after 1 is 2 :)

Basically it means that you just try to use the pre-built deps and see if Conan complains or not.

docs/conan.md
Nordsoft91: ERROR: Failed requirement 'libiconv/1.17@kambala/apple' from 'boost/1.79.0'
ERROR: Unable to find 'libiconv/1.17@kambala/apple' in remotes

docs/conan.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Use our prebuilt binaries to build for macOS x86_64 with Xcode
+
+```
+conan install . \
+  --install-folder=conan-generated \
+  --no-imports \
+  --build=never \
+  --profile:build=default \
+  --profile:host=CI/conan/macos-intel
```
Nordsoft91, line 88: WARN: boost/1.79.0: requirement libiconv/1.17 overridden by your conanfile to libiconv/1.17@kambala/apple 
libiconv/1.17@kambala/apple: Not found in local cache, looking in remotes...
libiconv/1.17@kambala/apple: Trying with 'conancenter'...
ERROR: Failed requirement 'libiconv/1.17@kambala/apple' from 'boost/1.79.0'
ERROR: Unable to find 'libiconv/1.17@kambala/apple' in remotes

Nordsoft91 (changes requested):
I followed instruction step by step from zero and got the blocking error

docs/conan.md
Nordsoft91: it confuses, 1 doesn't suppose any action from me, not clear what should I skip

docs/conan.md
Nordsoft91: but now there is no source of truth for brew. So If wiki has mistake, it significantly increases entry level for developers

docs/conan.md
kambala-decapitator: If you don't know what "section" in a document means, then it's not a problem of the document.

In which case - it's written on step 0 of "Download dependencies" section. Please try to understand what you're reading.

docs/conan.md
kambala-decapitator: there's a verb "check" - it means action

docs/conan.md
kambala-decapitator: can be relative or absolute, it's up to you. Also see examples: there're both.

docs/conan.md
kambala-decapitator: > can you specify the cases why I may want to consume them?

there's no specific case, it's up to you to decide. For iOS you can't get binaries from ConanCenter, for example.

> suggest to specify possible options, e.g. macos-arm, macos-intel

they're given on wiki and in examples

kambala-decapitator:
> I followed instruction step by step from zero and got the blocking error

you're not using PR's branch. If you want to test on `develop`, then download macOS dependencies from `1.0` tag.

docs/conan.md
Nordsoft91: > For iOS you can't get binaries from ConanCenter, for example.

Don't you want to specify those scenarios?

docs/conan.md
Nordsoft91: Whatever I check/sure or not, I still have to go next step

docs/conan.md
Nordsoft91: ok, maybe it not clear only for stupid-me and other people don't leave comments because they are not

docs/conan.md
Nordsoft91, line 50:
```suggestion
- ***conan-generated***: directory (absolute or relative) where the generated files will appear (for instance `../vcmi-deps`). This value is used in CMake presets from VCMI, but you can actually use any directory and override it in your local presets.
```

docs/conan.md
nullkiller: Does our CI try to build using brew? Not necessary to publish artifacts, just check?

kambala-decapitator: it used brew in the past, but it's not a viable option for long-term: in the beginning of next year GitHub will remove the 10.15 runner which means that only users on macOS 11 and later will be able to use CI's artefact. With Conan we target 10.13 at the moment as a sensible baseline, but I believe dependencies should be buildable to a lower macOS as well.

docs/conan.md
kambala-decapitator: if you checked and are sure you can't use our prebuilt binaries, it obviously makes no sense to continue with these steps - that's the whole point of the check. If you're unsure, you just continue with these steps and then you'll see if you can use the binaries or not (whether `conan install` fails or not).

docs/conan.md
kambala-decapitator: let's add a link to section to make it fool-proof then :)

docs/conan.md
kambala-decapitator: why should I specify them if they're clear if you open supported platforms link given in the beginning of `Using prebuilt binaries from ConanCenter` section?

another [obvious] case of not using binaries (our or from ConanCenter): you have zero trust to binaries published on the web by random people.

our deps binaries have customised options to produce the smallest possible build on CI for end users. Developers can decide what they want.

docs/conan.md
kambala-decapitator: `conan-generated` is already an "instance". I can write it as `./conan-generated` to make it more clear.

kambala-decapitator:
> If you want to test on develop, then download macOS dependencies from 1.0 tag.

@Nordsoft91 just figured out that it won't work, as `conanfile.py` for macOS in `develop` is a bit broken if you want to build from source. The file in the PR is correct, you can simply copy it and test.

Nordsoft91 (approved):

kambala-decapitator:
@Nordsoft91 could you test the iOS artefact via AltStore for double check that I did everything correctly in https://github.com/vcmi/vcmi/pull/1073/commits/409ab51036fc4d16cd2c9b4300c3894ffd69c9e8 ?

Nordsoft91:
Are you sure that you are referring to proper hash? The last action I found is https://github.com/vcmi/vcmi/suites/8838013089/artifacts/402872761

kambala-decapitator:
> Are you sure that you are referring to proper hash? The last action I found is https://github.com/vcmi/vcmi/suites/8838013089/artifacts/402872761

yes, I'm sure. What I posted is commit reference and it's not the branch's HEAD. Also, artefacts have "strange" commit hash in the name because CI merges PR's branch with `develop` before building.

Issue #1069, opened 2022-10-14
Starting hero position in random maps is hardcoded and not taken from town DEF template mask.


Macron1Robot:
[test_castle_entrance-mod.zip](https://github.com/vcmi/vcmi/files/9788462/test_castle_entrance-mod.zip)
![image](https://user-images.githubusercontent.com/7364571/195900442-b99fc325-78b4-408e-b0df-567d39b71d94.png)
To reproduce activate this mod and start random map as Castle faction.

PR #1075, opened 2022-10-18
[1069] Fix bug related to town entrance

Nordsoft91:
Fix for https://github.com/vcmi/vcmi/issues/1069

### DIFF
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -1148,7 +1148,14 @@ void CGameState::placeCampaignHeroes()
 
 void CGameState::placeStartingHero(PlayerColor playerColor, HeroTypeID heroTypeId, int3 townPos)
 {
-	townPos.x -= 2; //FIXME: use actual visitable offset of town
+	for(auto town : map->towns)
+	{
+		if(town->getPosition() == townPos)
+		{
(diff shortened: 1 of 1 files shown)


lib/CGameState.cpp
```diff
@@ -1866,8 +1873,8 @@ void CGameState::initVisitingAndGarrisonedHeroes()
 		{
 			for(CGTownInstance *t : k->second.towns)
 			{
-				int3 vistile = t->pos; vistile.x--; //tile next to the entrance
-				if(vistile == h->pos || h->pos==t->pos)
+				int3 vistile = t->visitablePos(); vistile.x++; //tile next to the entrance
```
nullkiller, line 1876: Why we changed -- to ++? Also might be worth to actually take all tiles around visitablePos and test them for accessibility and put hero on the first accessible tile. This will be the safest way as for me

Nordsoft91: It's a bugfix, don't want to take extra scope here.

Before it was townpos.x - 2. Hero pos was townpos.x -2 + 1, which translated into townpos.x - 1
Now it's visitablePos, hero pos is visitablepos + 1

Why hero pos is x + 1? Don't know

Issue #989, opened 2022-09-25
attempt to open screenshots of Neutral Heroes mod triggers assertion


kambala-decapitator:
Assertion failed: (modModel->hasMod(modName)), function loadScreenshots, file cmodlistview_moc.cpp, line 723.

kambala-decapitator:
got fixed automagically

PR #1079, opened 2022-10-21
Fix screenshots loading

Nordsoft91:
Screenshots now can be loaded either from mod.json or from repository


kambala-decapitator:
does it fix #989 ?

Nordsoft91:
No, but I cannot reproduce this issue

kambala-decapitator (approved):

PR #406, opened 2018-01-14
New victory/loss conditions

alexvins:
Yet another missing part of Map Format. Initial implementation of new condition types, should make maps with such conditions playable.

PR #1085, opened 2022-10-23
New victory loss conditions

nullkiller:
Rebased https://github.com/vcmi/vcmi/pull/406
There was an idea that this might be already working but need to investigate further, and need support from vcmi editor

### DIFF
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -2341,48 +2341,32 @@ bool CGameState::checkForVictory(PlayerColor player, const EventCondition & cond
 			}
 		}
 		case EventCondition::TRANSPORT:
-		{
-			const CGTownInstance *t = static_cast<const CGTownInstance *>(condition.object);
-			if((t->visitingHero && t->visitingHero->hasArt(condition.objectType))
-				|| (t->garrisonHero && t->garrisonHero->hasArt(condition.objectType)))
 			{
-				return true;
+				const CGTownInstance *t = static_cast<const CGTownInstance *>(condition.object);
+				if((t->visitingHero && t->visitingHero->hasArt(condition.objectType))
+					|| (t->garrisonHero && t->garrisonHero->hasArt(condition.objectType)))
+				{
+					return true;
+				}
 			}
 			return false;
-		}
 		case EventCondition::DAYS_PASSED:
-		{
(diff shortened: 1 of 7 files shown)


lib/CGameState.cpp
```diff
@@ -2340,56 +2324,273 @@ bool CGameState::checkForVictory(PlayerColor player, const EventCondition & cond
...
 		}
 		case EventCondition::TRANSPORT:
-		{
-			const CGTownInstance *t = static_cast<const CGTownInstance *>(condition.object);
-			if((t->visitingHero && t->visitingHero->hasArt(condition.objectType))
-				|| (t->garrisonHero && t->garrisonHero->hasArt(condition.objectType)))
 			{
-				return true;
+				const CGTownInstance *t = static_cast<const CGTownInstance *>(condition.object);
+				if((t->visitingHero && t->visitingHero->hasArt(condition.objectType))
```
Nordsoft91, line 2329:
```suggestion
				const auto * t = dynamic_cast<const CGTownInstance *>(condition.object);
				if(t && (t->visitingHero && t->visitingHero->hasArt(condition.objectType))
```

lib/CGameState.cpp
```diff
@@ -2340,56 +2324,273 @@ bool CGameState::checkForVictory(PlayerColor player, const EventCondition & cond
...
+			return team.count(object->getOwner()) != 0;
+		else
+			return object->getOwner() == playerState->color;
+	};
+
+	std::vector<const CGHeroInstance *> heroesToCheck;
+
+	std::vector<const CArmedInstance *> armiesToCheck;
+
+	auto collectHeroes = [&]()
```
Nordsoft91, line 2387: too many empty lines

lib/CGameState.cpp
```diff
@@ -2340,56 +2324,273 @@ bool CGameState::checkForVictory(PlayerColor player, const EventCondition & cond
...
+	{
+		if(condition.object)
+		{
+			auto hero = dynamic_cast<const CGHeroInstance *>(condition.object);
+
+			if(hero)
```
Nordsoft91, line 2393:
```suggestion
			if(const auto * hero = dynamic_cast<const CGHeroInstance *>(condition.object))
```

PR #1086, opened 2022-10-23
Mantis-2934: fix conditions to upgrade town in h3m format parser

nullkiller:
It is not possible to build Village hall because it is what we start with. So the condition was shifted.

https://bugs.vcmi.eu/view.php?id=2934


kambala-decapitator (approved):
looks sane

PR #1091, opened 2022-10-26
Remote server connection and lobby mode

Nordsoft91:
Quick&dirty prototype for lobby and server.

Ability to connect to remote server, chatting, creating/joining games

To play need to enter server address and port (5002) for connecting players. Host player creates game as usual.

Proxy server is not robust and it's needed to restart it after session and if something bad happens.


Nordsoft91:
Do not remove branch

PR #1099, opened 2022-11-03
disable tests by default

kambala-decapitator:
There're frequent issues when building GTest, normal users don't need tests anyway


nullkiller:
Do we build tests on CI? at least one platform?

kambala-decapitator:
> Do we build tests on CI? at least one platform?

no

PR #1103, opened 2022-11-03
Preventing multiple AI from being created

Nordsoft91:
1) Server hangs and crashes if press "A" many times during battle
This PR makes it stable.
2) Fix crash when pressing "V" during AI turn

### DIFF
--- a/client/mapHandler.cpp
+++ b/client/mapHandler.cpp
@@ -536,6 +536,8 @@ void CMapHandler::CMapWorldViewBlitter::drawTileOverlay(SDL_Surface * targetSurf
 	for(auto & object : objects)
 	{
 		const CGObjectInstance * obj = object.obj;
+		if(!obj)
+			continue;
 
 		const bool sameLevel = obj->pos.z == pos.z;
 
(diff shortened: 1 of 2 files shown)


client/mapHandler.cpp
```diff
@@ -536,6 +536,8 @@ void CMapHandler::CMapWorldViewBlitter::drawTileOverlay(SDL_Surface * targetSurf
 	for(auto & object : objects)
 	{
 		const CGObjectInstance * obj = object.obj;
+		if(!obj)
```
nullkiller, line 539: Interesting case. What does it mean?

Nordsoft91: Multithreading problem - attempt to dereference object which was deleted. General rule - all ptrs has to be checked for nullptr. If it's not expected, we should use references

nullkiller: Usually multithreading problems are solved by proper locking. Any additional details on bug mechanics. Should not object be erased first from objects list and deleted afterwards?

Nordsoft91: Why it should be erased from the list? List is valid, but object has an obj field which is a pointer and could be nullptr, there is no restrictions

Nordsoft91: I agree that maybe root cause is in another place, but rendering must not crash the game

Issue #1071, opened 2022-10-15
In latest night builds there are problems with schema obstacle.json and file obstacles.json.


Macron1Robot:
At start of VCMI in console there are errors about obstacles.json from config folder
1) in config/obstacles.json "allowedTerrain", in config/schemas/obstacle.json there is "allowedTerrains"
![image](https://user-images.githubusercontent.com/7364571/196005627-58db80df-250f-4899-8fd6-bcbfbe9e839f.png)

2) in config/obstacles.json
![image](https://user-images.githubusercontent.com/7364571/196005690-4084c547-790d-46c4-a282-42dfdade520b.png)
there are array of string identifiers of special battle fields.
 in config/schemas/obstacle.json is written^
`		"specialBattlefields": {
			"type": "array",
			"description": "Obstacles can be placed on specified specified battlefields",
			"items": { "$ref" : "battlefield.json" }
		},`
But it must be
`		"specialBattlefields": {
			"type": "array",
			"description": "Obstacles can be placed on specified specified battlefields",
			"items": { "type" : "string" }
		},`

3. in mod.json schema section "obstacles" dissapeared:
![image](https://user-images.githubusercontent.com/7364571/196022869-e7048948-0f31-4a0f-b6f2-876854f7dae2.png)
Block
`		"obstacles":{
			"type":"array",
			"description": "List of configuration files for obstacles",
			"items": { "type":"string", "format" : "textFile" }

		},`
should be added

nullkiller:
Draw should be possible so this is a very special case which obviously needs a mod to test it.

Macron1Robot:
Wrong topic?

nullkiller:
Yes

krs0:
I do not have point 3. on my side... could this be caused by a mod you have?

> 3. In mod.json schema section "obstacles" dissapeared:
![image](https://user-images.githubusercontent.com/15194321/200020262-a58eb919-3c1e-4eee-a46f-a3a8c7418996.png)
> ```
> "obstacles":{
> "type":"array",
> "description": "List of configuration files for obstacles",
> "items": { "type":"string", "format" : "textFile" }
>  },
> ```
> should be added

krs0:
@Macron1Robot could No.3 be moved in a separate issue? So that this one can be closed?

Macron1Robot:
added 1122

krs0:
No.3 moved to #1122. This one can be closed.

PR #1105, opened 2022-11-03
Problems with schema obstacle.json and file obstacles.json #1071

krs0:
#1071 - In latest night builds there are problems with schema obstacle.json and file obstacles.json

### DIFF
--- a/lib/ObstacleHandler.cpp
+++ b/lib/ObstacleHandler.cpp
@@ -85,7 +85,7 @@ ObstacleInfo * ObstacleHandler::loadFromJson(const std::string & scope, const Js
 	info->animation = json["animation"].String();
 	info->width = json["width"].Integer();
 	info->height = json["height"].Integer();
-	for(auto & t : json["allowedTerrain"].Vector())
+	for(auto & t : json["allowedTerrains"].Vector())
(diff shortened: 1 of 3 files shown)


lib/ObstacleHandler.cpp
```diff
@@ -85,7 +85,7 @@ ObstacleInfo * ObstacleHandler::loadFromJson(const std::string & scope, const Js
 	info->animation = json["animation"].String();
 	info->width = json["width"].Integer();
 	info->height = json["height"].Integer();
-	for(auto & t : json["allowedTerrain"].Vector())
+	for(auto & t : json["allowedTerrains"].Vector())
```
Nordsoft91, line 88: If you rename it here, you should also rename it in all mods, which already use "allowedTerrain".
AFAIK there are 2 mods: hota and forge, change is needed in obstacles.json for vcmi-1.1 branches

Nordsoft91 (changes requested):

lib/ObstacleHandler.cpp
krs0: Ok, created 2 more pull requests for hota and forge. Directly in vcmi-1.1 branches.

Nordsoft91 (approved):

Nordsoft91:
@nullkiller merge?

PR #1108, opened 2022-11-04
[HDMod features] Quick combat

Nordsoft91:
Enable quick combat option in game settings to test.

Should work only while attacking a neutral creatures.
Should have only one attempt.

Known issues
- [x] Looks like something wrong with experience
- [x] ~~Lose during quick combat, restart, escape. Purchase same hero and attach again - no quick combat~~ - I decided do not proceed with fixing that
- [x] AI sometimes hanging on query reply
- [x] Support draw case
- [x] fix crash when friendly AI loses battle but user wins it after replay
- [x] In hot seat mode second player has replay button after fighting between players
- [x] In hot seat mode game freezes if attacker loses while defender plays auto-fighting
- [x] If friendly AI spends mana during fight, it's not restored for replay
- [x] Find out root cause of BattleAI stuck

https://user-images.githubusercontent.com/9308612/200134465-6127cd1a-12ca-4857-9dea-8865a261e9ac.mp4


nullkiller:
The branch has conflicts with develop

client/CPlayerInterface.cpp
```diff
@@ -707,7 +707,12 @@ void CPlayerInterface::battleStartBefore(const CCreatureSet *army1, const CCreat
 void CPlayerInterface::battleStart(const CCreatureSet *army1, const CCreatureSet *army2, int3 tile, const CGHeroInstance *hero1, const CGHeroInstance *hero2, bool side)
 {
 	EVENT_HANDLER_CALLED_BY_CLIENT;
-	if (settings["adventure"]["quickCombat"].Bool())
+	bool replay = (lastBattleArmies.first == army1 && lastBattleArmies.second == army2);
+	lastBattleArmies.first = army1;
+	lastBattleArmies.second = army2;
+	//quick combat with neutral creatures only
+	auto * army2_object = dynamic_cast<const CGObjectInstance *>(army2);
+	if(!replay && army2_object && army2_object->getOwner() == PlayerColor::UNFLAGGABLE && settings["adventure"]["quickCombat"].Bool())
```
nullkiller, line 715: Will we have an ability to always use autocombat for all types of fight as it was before?

client/Client.cpp
```diff
@@ -581,15 +581,30 @@ void CClient::battleStarted(const BattleInfo * info)
...
+	if(settings["session"]["spectate"].Bool() && !settings["session"]["spectate-skip-battle"].Bool())
+		callBattleStart(PlayerColor::SPECTATOR, 1);
+
+	if(vstd::contains(playerint, leftSide.color) && playerint[leftSide.color]->human)
+		att = std::dynamic_pointer_cast<CPlayerInterface>(playerint[leftSide.color]);
+
+	if(vstd::contains(playerint, rightSide.color) && playerint[rightSide.color]->human)
+		def = std::dynamic_pointer_cast<CPlayerInterface>(playerint[rightSide.color]);
 
 	//If quick combat is not, do not prepare interfaces for battleint
```
nullkiller, line 603: Looks the comment is no longer actual

server/CGameHandler.h
```diff
@@ -309,6 +310,7 @@ class CGameHandler : public IGameCallback, public CBattleInfoCallback, public En
 
 		const CGHeroInstance *winnerHero, *loserHero;
 		PlayerColor victor, loser;
+		ui8 winnerSide;
```
nullkiller, line 313: Just in case sometimes it is possible to have draw when both sides loose

client/CPlayerInterface.cpp
Nordsoft91: hm, no. For me it looks more like "skipBattle" functionality. This one is hola-style to save time (but without multiple replays)

server/CGameHandler.h
Nordsoft91: Action item was added

client/battle/CBattleInterfaceClasses.cpp
```diff
@@ -405,6 +405,12 @@ CBattleResultWindow::CBattleResultWindow(const BattleResult & br, CPlayerInterfa
 
 	exit = std::make_shared<CButton>(Point(384, 505), "iok6432.def", std::make_pair("", ""), [&](){ bExitf();}, SDLK_RETURN);
 	exit->setBorderColor(Colors::METALLIC_GOLD);
+	
+	if(allowReplay)
+	{
+		repeat = std::make_shared<CButton>(Point(24, 505), "icn6432.def", std::make_pair("", ""), [&](){ bRepeatf();}, SDLK_ESCAPE);
```
IvanSavenko, line 411: Perhaps add description?
This is not H3 feature so it is not clear what "cancel" button does at battle result screen.
Proposition: add texts
"Replay battle" for tooltip in status bar
"Discard results of the battle and replay it manually. Only available once per battle" second string, for right-click popup. (feel free to adjust text to your liking)
These new strings can be added into config/translate.json

client/CPlayerInterface.cpp
```diff
@@ -912,7 +920,13 @@ void CPlayerInterface::battleEnd(const BattleResult *br)
 
 		if(!battleInt)
 		{
-			GH.pushIntT<CBattleResultWindow>(*br, *this);
+			bool replay = !settings["adventure"]["alwaysSkipCombat"].Bool(); //do not allow manual replay
```
nullkiller, line 923: Are you sure we do not need to test something else as well here? If we press autobattle in regular fight?

Nordsoft91: Could be different corner cases which are not tested. For single player it's tested, autofight from UI as well

server/CGameHandler.cpp
```diff
@@ -2640,7 +2664,35 @@ void CGameHandler::startBattlePrimary(const CArmedInstance *army1, const CArmedI
...
-	auto battleQuery = std::make_shared<CBattleQuery>(this, gs->curB);
+	//existing battle query for retying auto-combat
+	auto battleQuery = std::dynamic_pointer_cast<CBattleQuery>(queries.topQuery(gs->curB->sides[0].color));
+	if(battleQuery)
+	{
+		for(int i : {0, 1})
+		{
+			if(heroes[i])
+			{
+				SetMana restoreInitialMana;
```
IvanSavenko, line 2675: Is there anything other than mana that we might want to reset?

What about bonuses that last "until next battle"? They are not lost after quick combat?

client/CPlayerInterface.cpp
```diff
@@ -900,7 +909,14 @@ void CPlayerInterface::battleEnd(const BattleResult *br)
 
 		if(!battleInt)
 		{
-			GH.pushIntT<BattleResultWindow>(*br, *this);
+			bool replay = allowBattleReplay && !settings["adventure"]["alwaysSkipCombat"].Bool(); //do not allow manual replay
+			allowBattleReplay = false;
+			auto wnd = std::make_shared<BattleResultWindow>(*br, *this, replay);
```
IvanSavenko, line 914: Why add a comment when it can be part of the code?
```suggestion
			bool allowManualReplay = allowBattleReplay && !settings["adventure"]["alwaysSkipCombat"].Bool();
			allowBattleReplay = false;
			auto wnd = std::make_shared<BattleResultWindow>(*br, *this, allowManualReplay);

```

client/CPlayerInterface.cpp
```diff
@@ -693,7 +693,15 @@ void CPlayerInterface::battleStartBefore(const CCreatureSet *army1, const CCreat
 void CPlayerInterface::battleStart(const CCreatureSet *army1, const CCreatureSet *army2, int3 tile, const CGHeroInstance *hero1, const CGHeroInstance *hero2, bool side)
 {
 	EVENT_HANDLER_CALLED_BY_CLIENT;
-	if (settings["adventure"]["quickCombat"].Bool())
+	bool replay = (lastBattleArmies.first == army1 && lastBattleArmies.second == army2); //will be true if player already refused auto-battle result
```
IvanSavenko, line 696: Edge case, but attack monsters -> retreat -> attack same monsters again
I think will result in inability to replay battle?

Also, perhaps better to rename to more meningful name? e.g. 
```suggestion
	bool autoBattleResultRefused = (lastBattleArmies.first == army1 && lastBattleArmies.second == army2);
```

server/CGameHandler.cpp
Nordsoft91: Good point, I need to test that

Nordsoft91: tested and it works

IvanSavenko (approved):

krs0:
If I try to redo fight after quick combat enemy just stays there....

https://user-images.githubusercontent.com/15194321/230775557-7ca6908f-0490-438f-a8e8-0a64b3c2026a.mp4

Nordsoft91:
@krs0 , could you please give more information what is the issue?

krs0:
I have downloaded your PR, started first map... All for one? Checked quick combat and attacked the gargoyles a little to the top right of castle. Upon battle replay Gargoyles just sit there like ducks.

Nordsoft91:
@krs0 , but this is expected, no? quick battle does auto-combat, replay allows to fight **same** battle manually

IvanSavenko:
> @krs0 , but this is expected, no? quick battle does auto-combat, replay allows to fight **same** battle manually

Errmm... But why *enemy* gargoyles do nothing? Why enemy AI does nothing on manual replay?

Nordsoft91:
yes, I met this behavior too, but it seems like battleAI issue. I don't think it's related to quick combat

Nordsoft91:
I cannot reproduce it

Nordsoft91:
Ok, creatures receives bonuses from native terrain each round.

I won't proceed with fix because it seems to be very close with @rilian-la-te does here https://github.com/vcmi/vcmi/pull/1882

So waiting for this PR

client/CPlayerInterface.cpp
Nordsoft91: BTW, it's done now

rilian-la-te:
@Nordsoft91 It works as in HD mod?

rilian-la-te:
@Nordsoft91 but player have manual attempt after auto anyway? If so, good feature)

PR #1111, opened 2022-11-06
Code refactor and fix condition bug

GermanAizek:
Code improvements both in readability and excess code parts

### DIFF
--- a/client/windows/CAdvmapInterface.cpp
+++ b/client/windows/CAdvmapInterface.cpp
@@ -573,7 +573,7 @@ CAdvMapInt::CAdvMapInt():
 	strongInterest = true; // handle all mouse move events to prevent dead mouse move space in fullscreen mode
 	townList.onSelect = std::bind(&CAdvMapInt::selectionChanged,this);
 	bg = BitmapHandler::loadBitmap(ADVOPT.mainGraphic);
-	if (ADVOPT.worldViewGraphic != "")
+	if(!ADVOPT.worldViewGraphic.empty())
 	{
 		bgWorldView = BitmapHandler::loadBitmap(ADVOPT.worldViewGraphic);
 	}
--- a/lib/CBonusTypeHandler.cpp
+++ b/lib/CBonusTypeHandler.cpp
@@ -94,7 +94,9 @@ std::string MacroString::build(const GetValue & getValue) const
 CBonusType::CBonusType()
 {
 	hidden = true;
-	icon = nameTemplate = descriptionTemplate = "";
+	icon.clear();
+	nameTemplate.clear();
+	descriptionTemplate.clear();
 }
 
 CBonusType::~CBonusType()
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -532,7 +532,7 @@ std::pair<Obj,int> CGameState::pickObject (CGObjectInstance *obj)
 			if (auto info = dynamic_cast<CCreGenAsCastleInfo*>(dwl->info))
 			{
 				faction = getRandomGenerator().nextInt((int)VLC->townh->size() - 1);
-				if(info->asCastle && info->instanceId != "")
+				if(info->asCastle && !info->instanceId.empty())
 				{
 					auto iter = map->instanceNames.find(info->instanceId);
 
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -46,7 +46,7 @@ void CCreGenAsCastleInfo::serializeJson(JsonSerializeFormat & handler)
 
 	if(!handler.saving)
 	{
-		asCastle = (instanceId != "");
+		asCastle = !instanceId.empty();
 		allowedFactions.clear();
 	}
 
--- a/AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
+++ b/AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
@@ -334,7 +334,7 @@ BuildingInfo::BuildingInfo()
 	buildCost = 0;
 	buildCostWithPrerequisits = 0;
 	prerequisitesCount = 0;
-	name = "";
+	name.clear();
 	armyStrength = 0;
 }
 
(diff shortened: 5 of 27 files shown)


client/CServerHandler.cpp
```diff
@@ -489,7 +489,7 @@ void CServerHandler::sendMessage(const std::string & txt) const
 		std::string connectedId, playerColorId;
 		readed >> connectedId;
 		readed >> playerColorId;
-		if(connectedId.length(), playerColorId.length()) // BUG https://bugs.vcmi.eu/view.php?id=3144
+		if(connectedId.length() < playerColorId.length())
```
kambala-decapitator, line 492: reading the following code, I don't find the fix correct. It looks like the intention was to check that strings are not empty.

lib/CBonusTypeHandler.cpp
```diff
@@ -94,7 +94,9 @@ std::string MacroString::build(const GetValue & getValue) const
 CBonusType::CBonusType()
 {
 	hidden = true;
-	icon = nameTemplate = descriptionTemplate = "";
+	icon.clear();
+    nameTemplate.clear();
+    descriptionTemplate.clear();
```
kambala-decapitator, line 99: we use tabs for indentation
```suggestion
	nameTemplate.clear();
	descriptionTemplate.clear();
```

lib/CGameState.cpp
```diff
@@ -726,7 +726,7 @@ void CGameState::init(const IMapService * mapService, StartInfo * si, bool allow
 	logGlobal->info("\tUsing random seed: %d", si->seedToBeUsed);
 	getRandomGenerator().setSeed(si->seedToBeUsed);
 	scenarioOps = CMemorySerializer::deepCopy(*si).release();
-	initialOpts = CMemorySerializer::deepCopy(*si).release();
+	initialOpts = scenarioOps;
```
kambala-decapitator, line 729: are you absolutely sure that this is correct? To me it looks that the original intention was to keep these entities completely separate.

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -46,7 +46,7 @@ void CCreGenAsCastleInfo::serializeJson(JsonSerializeFormat & handler)
 
 	if(!handler.saving)
 	{
-		asCastle = (instanceId != "");
+		asCastle = (!instanceId.empty());
```
kambala-decapitator, line 49:
```suggestion
		asCastle = !instanceId.empty();
```

kambala-decapitator (changes requested):
Please split the first commit into two: one that actually fixes the reported bug, the other with code improvements.

Btw have you used some static analyser to find those issues?

lib/CGameState.cpp
Nordsoft91: This is incorrect and will break restart scenario functionality

Nordsoft91: This is incorrect and will break restart scenario functionality

client/CServerHandler.cpp
Nordsoft91: Exactly, this command assigns specified player color to connectionId. It doesn't make any sense to compare string length. Passing connectionId="5" and playerColorId="1" would be absolutely valid but fails with proposed change

lib/CGameState.cpp
Nordsoft91: this change is incorrect and will break restart scenario functionality

client/CServerHandler.cpp
nullkiller: Still the syntax is weird. Cpp ref does not say it is legal to put , inside if and what does it mean

nullkiller: Looks like it is comma operator and according to it the first statement result will be just discurded

kambala-decapitator: > Looks like it is comma operator and according to it the first statement result will be just discurded

yes, you're absolutely correct

GermanAizek: @kambala-decapitator OK, then I will remove from this PR changes affected `client/CServerHandler.cpp`

lib/mapObjects/CGTownInstance.cpp
GermanAizek: @kambala-decapitator fix it here https://github.com/vcmi/vcmi/pull/1111/commits/2c01e9680c3778d5f4a2732eb6f7d01c42fc2501

lib/CGameState.cpp
GermanAizek: @Nordsoft91 fix it here https://github.com/vcmi/vcmi/pull/1111/commits/2c01e9680c3778d5f4a2732eb6f7d01c42fc2501

lib/CBonusTypeHandler.cpp
GermanAizek: @kambala-decapitator fix it https://github.com/vcmi/vcmi/pull/1111/commits/2c01e9680c3778d5f4a2732eb6f7d01c42fc2501

client/windows/CAdvmapInterface.cpp
```diff
@@ -573,7 +573,7 @@ CAdvMapInt::CAdvMapInt():
 	strongInterest = true; // handle all mouse move events to prevent dead mouse move space in fullscreen mode
 	townList.onSelect = std::bind(&CAdvMapInt::selectionChanged,this);
 	bg = BitmapHandler::loadBitmap(ADVOPT.mainGraphic);
-	if (ADVOPT.worldViewGraphic != "")
+	if (!ADVOPT.worldViewGraphic.empty())
```
Nordsoft91, line 576: We don’t use spaces after for if while pls fix here and in other placs too

nullkiller (approved):

client/windows/CAdvmapInterface.cpp
GermanAizek: @Nordsoft91 fix it here https://github.com/vcmi/vcmi/pull/1111/commits/7fdad4e0f6edc1fae14b1008360f9bc2a9d885ef

Nordsoft91 (approved):

Issue #1118, opened 2022-11-09
VCMI H3 original archives extractor


krs0:
Original H3 files are grouped into archives. They should be extracted to appropriate folders and consumed from there.

_Images will be extracted to separate PNGs
Sprites will remain in .defs
Resource names will not change_

- [x] Run Archives Extractor code from within VCMI_MapEditor as command line options (Intermediary step until VCMI installer)
- [x] Extract all resources to separate folders. Images, Misc, Sound, Sprites, Video
- [x] Split mixed up resources like: CPRSMALL, FLAGPORT, ITPA, ITPT, TWCRPORT, UN32, UN44, into separate files.
- [x] PCX images will be converted to PNG
- [ ] Split other .defs _(Images not multiple frame Sprites)_ into separate image files. (EG: spell icons, artifacts)

_Once everything is up and running, the Archives Extractor functionality will be incorporated into VCMI Installer_

:: ORIGINAL ::
```
<USER_DATA_PATH>/Data/
    H3ab_ahd.snd
    H3ab_ahd.vid
    H3ab_bmp.lod
    H3ab_spr.lod
    H3bitmap.lod
    H3sprite.lod
    Heroes3.snd
    VIDEO.VID
```

:: EXTRACTED ::
```
<USER_DATA_PATH>/extracted/
    Images
    Misc
    Sound
    Sprites
    Video
    Sprites
        \CPRSMALL
        \FLAGPORT
        \ITPA
        \ITPT
        \TWCRPORT
        \UN32
        \UN44
```

_This is an intermediary step for having a full SoD mod. In the final form all extracted data will be moved to SoD mod folder._

Drawbacks of original approach with archives in DATA/:
- Loading only a subset of resources via mods is not possible, 
- modding is harder than it needs to be, because everything is clumped together. 
- Resources for different factions are clumped together in same file.

Drawbacks of this approach:
- Mods will not be 1 to 1 compatible with other Heroes III distributions. (Different file structure)

krs0:
First working draft in #1119

krs0:
After consulting with Nordsoft, I have moved the functionality to MapEditor. (Perhaps final place is its own project dependent on Animation or only QImage, but that beats me how to do it).
I've also changed saving format for icons to PNG (as per Ivan's suggestion)

krs0:
FLAGPORT PNG;s show some artefacts... will investigate.

[...]

PR #1119, opened 2022-11-09
VCMI H3 original archives extractor #1118

krs0:
First working version of the resource extractor, DefSplitter and PCX convertor. (Everything is extracted to UserDataPath/extracted)
Starting a little to late in the initialization chain because of Graphics object dependency.
Not callable as run option for VCMI_Client.exe
Tested only on Windows


misiokles:
What are diffirencies between josch extractor? https://github.com/josch/lodextract

krs0:
Resource extractor "enablers" are now command line options
note: In VCMI_lib the extractArchives will be passed all the way down the call chain! If anyone has a better idea...
(settings cannot be used, because it is not yet initialized at preinitDLL())
In rest seems complete, next step would be to extract all this functionality in a stand alone VCMI installer.

mapeditor/Animation.cpp
```diff
@@ -686,6 +689,44 @@ std::shared_ptr<QImage> Animation::getImage(size_t frame, size_t group, bool ver
...
+void Animation::exportBitmaps(const bfs::path& path, bool prependResourceName) const
+{
+	if (images.empty())
+	{
+		logGlobal->error("Nothing to export, animation is empty");
+		return;
+	}
+
+	boost::filesystem::path actualPath = path / "Sprites" / name;
+	boost::filesystem::create_directories(actualPath);
```
Nordsoft91, line 701: why not to use QT stuff instead of boost?

mapeditor/Animation.cpp
```diff
@@ -686,6 +689,44 @@ std::shared_ptr<QImage> Animation::getImage(size_t frame, size_t group, bool ver
...
+	for (const auto& groupPair : images)
+	{
+		size_t group = groupPair.first;
+
+		for (const auto& imagePair : groupPair.second)
+		{
+			size_t frame = imagePair.first;
+			const auto img = imagePair.second;
+
+			boost::format fmt("%d_%d.png");
```
Nordsoft91, line 714: please use QString - you can avoid type conversion

mapeditor/Animation.h
```diff
@@ -83,6 +83,8 @@ class Animation
 	void load  (size_t frame, size_t group = 0);
 	void unload(size_t frame, size_t group = 0);
 
+	void exportBitmaps(const boost::filesystem::path& path, bool prependResourceName = false) const;
```
Nordsoft91, line 86: it's never used with prependResourceName=False

mapeditor/resourceExtractor/ResourceConverter.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+/*
+ * VCMI_Lib.cpp, part of VCMI engine
```
Nordsoft91, line 4: 1. It's not VCMI_Lib.cpp
2. header should be on the top (before pragma once)

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+		{
+			if (!bfs::is_regular_file(dir_itr->status()))
+				return;
+
+			std::string filePath = dir_itr->path().string();
+			std::string fileStem = dir_itr->path().stem().string();
+			std::string filename = dir_itr->path().filename().string();
+			filename = boost::locale::to_lower(filename);
+
+			if(filename.find(".pcx") != std::string::npos)
```
Nordsoft91, line 35: boost::filesystem::extension(filename)

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+#include "boost/filesystem/path.hpp"
+#include "boost/locale.hpp"
+
+namespace bfs = boost::filesystem;
+
+// converts all pcx files from /Images into PNG
+void ConvertPcxToPng(bool deleteOriginals)
+{
+	bfs::path imagesPath = VCMIDirs::get().userCachePath() / "extracted" / "Images";
```
Nordsoft91, line 20: pathToQString(VCMIDirs::get().userCachePath()) - and then you can work with QString/QFile/QFilename/QDir

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+		}
+		catch ( const std::exception & ex )
+		{
+			logGlobal->info(dir_itr->path().filename().string() + " " + ex.what() + "\n");
+		}
+	}
+}
+
+// splits a def file into individual parts and converts the output to PNG format
+void SplitDefFile(std::string fileName, bfs::path spritesPath, bool deleteOriginals)
```
Nordsoft91, line 53:
```suggestion
void SplitDefFile(const std::string & fileName, const bfs::path & spritesPath, bool deleteOriginals)
```

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+{
+	if (CResourceHandler::get()->existsResource(ResourceID("SPRITES/" + fileName)))
+	{
+		std::string URI = fileName;
```
Nordsoft91, line 57: why copy of string is needed?

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+	SplitDefFile("TwCrPort.def", spritesPath, deleteOriginals);	// split town creature portraits
+	SplitDefFile("CPRSMALL.def", spritesPath, deleteOriginals);	// split hero army creature portraits 
+	SplitDefFile("FlagPort.def", spritesPath, deleteOriginals);	// adventure map dwellings
+	SplitDefFile("ITPA.def", spritesPath, deleteOriginals);		// small town icons
+	SplitDefFile("ITPt.def", spritesPath, deleteOriginals);		// big town icons
+	SplitDefFile("Un32.def", spritesPath, deleteOriginals);		// big town icons
+	SplitDefFile("Un44.def", spritesPath, deleteOriginals);		// big town icons
+}
+
+void ConvertExtractedResourceFiles(bool splitDefs, bool convertPcxToPng, bool deleteOriginals)
```
Nordsoft91, line 84: haven't tested this, but looks like that files from mods will be also extracted, which is not expected I believe

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+	bfs::path extractedPath = VCMIDirs::get().userDataPath() / "extracted";
+	bfs::path spritesPath = extractedPath / "Sprites";
+
+	SplitDefFile("TwCrPort.def", spritesPath, deleteOriginals);	// split town creature portraits
+	SplitDefFile("CPRSMALL.def", spritesPath, deleteOriginals);	// split hero army creature portraits 
+	SplitDefFile("FlagPort.def", spritesPath, deleteOriginals);	// adventure map dwellings
+	SplitDefFile("ITPA.def", spritesPath, deleteOriginals);		// small town icons
+	SplitDefFile("ITPt.def", spritesPath, deleteOriginals);		// big town icons
+	SplitDefFile("Un32.def", spritesPath, deleteOriginals);		// big town icons
+	SplitDefFile("Un44.def", spritesPath, deleteOriginals);		// big town icons
```
Nordsoft91, line 81:
```suggestion
	for(std::string defFilename : {"TwCrPort.def", "CPRSMALL.def", "FlagPort.def", "ITPA.def", "ITPt.def", "Un32.def", "Un44.def"}
		SplitDefFile(defFilename, spritesPath, deleteOriginals);
```
But actually it's bad idea to hardcore those strings

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+	for ( bfs::directory_iterator dir_itr(imagesPath); dir_itr != end_iter; ++dir_itr )
+	{
+		try
+		{
+			if (!bfs::is_regular_file(dir_itr->status()))
+				return;
+
+			std::string filePath = dir_itr->path().string();
+			std::string fileStem = dir_itr->path().stem().string();
+			std::string filename = dir_itr->path().filename().string();
```
Nordsoft91, line 32: filename can be declared outside to try-catch block and reused in catch

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+// splits a def file into individual parts and converts the output to PNG format
+void SplitDefFile(std::string fileName, bfs::path spritesPath, bool deleteOriginals)
+{
+	if (CResourceHandler::get()->existsResource(ResourceID("SPRITES/" + fileName)))
+	{
+		std::string URI = fileName;
+		std::unique_ptr<Animation> anim = make_unique<Animation>(URI);
+		anim->preload();
+		anim->exportBitmaps(VCMIDirs::get().userCachePath() / "extracted", true);
```
Nordsoft91, line 60: is extraction folder hardcoded?

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+		{
+			logGlobal->info(dir_itr->path().filename().string() + " " + ex.what() + "\n");
+		}
+	}
+}
+
+// splits a def file into individual parts and converts the output to PNG format
+void SplitDefFile(std::string fileName, bfs::path spritesPath, bool deleteOriginals)
+{
+	if (CResourceHandler::get()->existsResource(ResourceID("SPRITES/" + fileName)))
```
Nordsoft91, line 55: there is no guarantee that spritesPath is same path as ResourceID("SPRITES/"  +...)
can you work either with Resources or with filesystem directly?

mapeditor/mainwindow.cpp
```diff
@@ -88,6 +89,39 @@ void MainWindow::saveUserSettings()
...
+		extractArchives = true;
+
+	if (parser.isSet("s"))
+		splitDefs = true;
+
+	if (parser.isSet("c"))
+		convertPcxToPng = true;
+
+	if (parser.isSet("d"))
+		deleteOriginals = true;
```
Nordsoft91, line 122:
```suggestion
	extractArchives = parser.isSet("e");
	splitDefs = parser.isSet("s");
	convertPcxToPng = parser.isSet("c");
	deleteOriginals = parser.isSet("d");
```

lib/filesystem/Filesystem.cpp
```diff
@@ -26,10 +26,11 @@ VCMI_LIB_NAMESPACE_BEGIN
 std::map<std::string, ISimpleResourceLoader*> CResourceHandler::knownLoaders = std::map<std::string, ISimpleResourceLoader*>();
 CResourceHandler CResourceHandler::globalResourceHandler;
 
-CFilesystemGenerator::CFilesystemGenerator(std::string prefix):
+CFilesystemGenerator::CFilesystemGenerator(std::string prefix, bool extractArchives):
 	filesystem(new CFilesystemList()),
 	prefix(prefix)
 {
+	this->extractArchives = extractArchives;
```
Nordsoft91, line 33: should be initialised, not assigned

lib/filesystem/Filesystem.cpp
```diff
@@ -200,16 +201,16 @@ ISimpleResourceLoader * CResourceHandler::get(std::string identifier)
...
 	auto fsConfigData = get("initial")->load(ResourceID(fsConfigURI, EResType::TEXT))->readAll();
 
 	const JsonNode fsConfig((char*)fsConfigData.first.get(), fsConfigData.second);
 
-	addFilesystem("data", "core", createFileSystem("", fsConfig["filesystem"]));
+	addFilesystem("data", "core", createFileSystem("", fsConfig["filesystem"], extractArchives));
 }
 
-void CResourceHandler::addFilesystem(const std::string & parent, const std::string & identifier, ISimpleResourceLoader * loader)
+void CResourceHandler::addFilesystem(const std::string & parent, const std::string & identifier, ISimpleResourceLoader * loader, bool extractArchives)
```
Nordsoft91, line 213: extractAcrhives is not used

lib/filesystem/CArchiveLoader.h
```diff
@@ -94,6 +99,12 @@ class DLL_LINKAGE CArchiveLoader : public ISimpleResourceLoader
...
 	/** Holds all entries of the archive file. An entry can be accessed via the entry name. **/
 	std::unordered_map<ResourceID, ArchiveEntry> entries;
+
+	/** Specifies if Original H3 archives should be extracted to a separate folder **/
+	bool extractArchives;
 };
 
+/** Constructs the file path for the extracted file. Creates the subfolder hierarchy aswell **/
+boost::filesystem::path createExtractedFilePath(std::string outputSubFolder, std::string entryName);
+
```
Nordsoft91, line 109: Declaration is excessive

lib/filesystem/CArchiveLoader.cpp
```diff
@@ -182,4 +212,48 @@ std::unordered_set<ResourceID> CArchiveLoader::getFilteredFiles(std::function<bo
...
+
+	// writeToOutputFile
+	std::ofstream out(extractedFilePath.string(), std::ofstream::binary);
+	out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+	out.write(data.get(), entry.fullSize);
+}
+
+bfs::path createExtractedFilePath(std::string outputSubFolder, std::string entryName)
+{
+	bfs::path extractionFolderPath = VCMIDirs::get().userCachePath() / "extracted" / outputSubFolder;
```
Nordsoft91, line 251: hardcoded path is not good

lib/filesystem/CArchiveLoader.cpp
```diff
@@ -182,4 +212,48 @@ std::unordered_set<ResourceID> CArchiveLoader::getFilteredFiles(std::function<bo
...
+
+	bfs::path extractedFilePath = createExtractedFilePath(outputSubFolder, entry.name);
+
+	// writeToOutputFile
+	std::ofstream out(extractedFilePath.string(), std::ofstream::binary);
+	out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+	out.write(data.get(), entry.fullSize);
+}
+
+bfs::path createExtractedFilePath(std::string outputSubFolder, std::string entryName)
```
Nordsoft91, line 249: const std::string & ?

lib/filesystem/CArchiveLoader.cpp
```diff
@@ -182,4 +212,48 @@ std::unordered_set<ResourceID> CArchiveLoader::getFilteredFiles(std::function<bo
...
+
+	// writeToOutputFile
+	std::ofstream out(extractedFilePath.string(), std::ofstream::binary);
+	out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+	out.write(data.get(), entry.fullSize);
+
+	fileStream.seek(currentPosition); // restore filestream position
+}
+
+void CArchiveLoader::extractToFolder( std::string outputSubFolder, const std::string& mountPoint, ArchiveEntry entry)
```
Nordsoft91, line 233: can this method reuse extractToFolder variant receiving CFileInputStream?

lib/filesystem/CArchiveLoader.cpp
```diff
@@ -182,4 +212,48 @@ std::unordered_set<ResourceID> CArchiveLoader::getFilteredFiles(std::function<bo
 	return foundID;
 }
 
+void CArchiveLoader::extractToFolder( std::string outputSubFolder, CFileInputStream& fileStream, ArchiveEntry entry)
+{
+	si64 currentPosition = fileStream.tell(); // save filestream position
+
+	std::unique_ptr<char[]> data = std::unique_ptr<char[]>(new char[entry.fullSize]);
```
Nordsoft91, line 219: you are creating char* but then casting to ui8*. It make sense to use ui8* here

Nordsoft91, line 219:
```suggestion
	std::vector<ui8> data(entry.fullSize);
```
And then data.pointer()

mapeditor/mainwindow.cpp
```diff
@@ -88,6 +89,39 @@ void MainWindow::saveUserSettings()
 	s.setValue(mainWindowPositionSetting, pos());
 }
 
+void MainWindow::parseCommandLine()
```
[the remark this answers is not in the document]
krs0: In the very end the functionality should be in a separate tool/ project by itself, that consumes Animation or parts of it from a common place with client and editor.
Can we agree that for first iteration this can stay behind command line options?

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "BitmapHandler.h"
+#include "Animation.h"
+
+#include "boost/filesystem/path.hpp"
+#include "boost/locale.hpp"
+
+namespace bfs = boost::filesystem;
+
+// converts all pcx files from /Images into PNG
+void ConvertPcxToPng(bool deleteOriginals)
```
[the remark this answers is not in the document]
krs0: Understood and I will comply...
how do you handle stuff like this? where parameter has same name as function?
```
	if (convertPcxToPng)
		convertPcxToPng(deleteOriginals);
```

mapeditor/resourceExtractor/ResourceConverter.cpp
krs0: This function iterates through <user_data>/extracted.
I populate that folder before mods are loaded. So this function should never see any mod files.

lib/filesystem/CArchiveLoader.h
krs0: Can you elaborate please? I do not get it.

lib/filesystem/CArchiveLoader.cpp
krs0: Can we make it a known path in VCMIDirs.h?
Or leave it hardcoded until standalone version is made?

lib/filesystem/CArchiveLoader.cpp
krs0: Tried but cannot see how... one operates on individual files, the other on whole archives.

client/gui/CAnimation.cpp
```diff
@@ -951,15 +951,15 @@ void CAnimation::initFromJson(const JsonNode & config)
...
+void CAnimation::exportBitmaps(const boost::filesystem::path& path, bool prependResourceName) const
 {
 	if(images.empty())
 	{
 		logGlobal->error("Nothing to export, animation is empty");
 		return;
 	}
 
-	boost::filesystem::path actualPath = path / "SPRITES" / name;
+	boost::filesystem::path actualPath = path / "Sprites" / name;
```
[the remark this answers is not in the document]
krs0: changed back to SPRITES

mapeditor/Animation.cpp
krs0: mainly historical reasons... but since we use both libraries can I keep the boost version?

mapeditor/Animation.cpp
krs0: I left it like this because its the same code as in CAnimation.cpp

mapeditor/Animation.h
krs0: Same function as in CAnimation.h At one point they should be the same.

mapeditor/resourceExtractor/ResourceConverter.cpp
krs0: I am not that familiar with QT. Since we also use boost can I leave it like it is now?

mapeditor/resourceExtractor/ResourceConverter.cpp
krs0: I try to split a def file from extracted folder... nevertheless I want to test that the resource exists so I can create an Animation object.
That is why I mix the both... and since I started from Original files they must be the same.
Or... can you suggest more concretely another way?

client/gui/CAnimation.h
```diff
@@ -122,7 +124,7 @@ class CAnimation
 
 	std::shared_ptr<IImage> getImage(size_t frame, size_t group=0, bool verbose=true) const;
 
-	void exportBitmaps(const boost::filesystem::path & path) const;
+	void exportBitmaps(const bfs::path & path, bool prependResourceName = false) const;
```
Nordsoft91, line 127: I suggest do not make any changes in CAnimation.h and CAnimation.cpp as those functions are not used

mapeditor/mainwindow.h
```diff
@@ -142,4 +144,11 @@ public slots:
 	int mapLevel = 0;
 
 	std::set<int> catalog;
+
+	// command line options
+	QString mapFilePath;			// FilePath to the H3 or VCMI map to open
+	bool extractArchives = false;	// extract original H3 archives into a separate folder
+	bool splitDefs = false;			// splits TwCrPort, CPRSMALL, FlagPort, ITPA, ITPt, Un32 and Un44 into individual PNG's
+	bool convertPcxToPng = false;	// converts single Images (found in Images folder) from .pcx to png.
+	bool deleteOriginals = false;	// delete original files, for the ones splitted / converted.
```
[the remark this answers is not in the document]
Nordsoft91: I'd suggest following:
1. Make widget with checks and paths related to resources converted
2. Call this widget from in-app menu
3. Let user specify all those options and then press ok

If you still want to have this as command line options, then it's even easier. All your new code is executed inside MainWindow constructor. You should get rid of all those class members. It could be something like this:
```
struct ExtractionOptions
{
	ExtractionOptions(const QStringList & args);
	bool extractArchieves, splitDefs, convertPcxToPng, deleteOriginals;
};

MainWindow::MainWindow(...)
{
	...
	ExtractionOptions opts(qApp->arguments());
	if(opts.extractArchives)
		convertExtractedResourceFiles(splitDefs, convertPcxToPng, deleteOriginals);
}

```

client/gui/CAnimation.h
```diff
@@ -122,7 +122,7 @@ class CAnimation
 
 	std::shared_ptr<IImage> getImage(size_t frame, size_t group=0, bool verbose=true) const;
 
-	void exportBitmaps(const boost::filesystem::path & path) const;
+	void exportBitmaps(const boost::filesystem::path & path, bool prependResourceName = false) const;
```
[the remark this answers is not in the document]
Nordsoft91: In such case it's better do not make it at all

mapeditor/mainwindow.cpp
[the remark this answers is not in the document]
Nordsoft91: in the end it's up to you. The risk here is "first iteration" can stay for a very long time

mapeditor/resourceExtractor/ResourceConverter.cpp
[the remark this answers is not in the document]
Nordsoft91: just don't call parameters same as functions

mapeditor/resourceExtractor/ResourceConverter.cpp
Nordsoft91: IMO, editor shouldn't have dependency on boost

client/gui/CAnimation.h
krs0: CAnimation.h and .cpp reverted

mapeditor/Animation.h
krs0: Removed parameter

client/gui/CAnimation.h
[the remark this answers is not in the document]
krs0: reverted CAnimation n.h and .cpp

mapeditor/Animation.cpp
krs0: converted to QString

lib/filesystem/CArchiveLoader.cpp
```diff
@@ -182,4 +209,47 @@ std::unordered_set<ResourceID> CArchiveLoader::getFilteredFiles(std::function<bo
...
+	std::vector<ui8> data(entry.fullSize);
+	fileStream.seek(entry.offset);
+	fileStream.read(data.data(), entry.fullSize);
+
+	bfs::path extractedFilePath = createExtractedFilePath(outputSubFolder, entry.name);
+
+	// writeToOutputFile
+	std::ofstream out(extractedFilePath.string(), std::ofstream::binary);
+	out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+	out.write((char*)&data[0], entry.fullSize);
```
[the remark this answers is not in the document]
krs0: compiler does not let me, write expects const char *

lib/filesystem/CArchiveLoader.cpp
```diff
@@ -182,4 +209,47 @@ std::unordered_set<ResourceID> CArchiveLoader::getFilteredFiles(std::function<bo
...
+
+	std::vector<ui8> data(entry.fullSize);
+	inputStream->read(data.data(), entry.fullSize);
+
+	bfs::path extractedFilePath = createExtractedFilePath(outputSubFolder, entry.name);
+
+	// writeToOutputFile
+	std::ofstream out(extractedFilePath.string(), std::ofstream::binary);
+	out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+	out.write((char*)&data[0], entry.fullSize);
```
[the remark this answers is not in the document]
krs0: works... but I had to change param 2 of extractToFolder from CFileInputStream to CInputStream... works but I really do not get why? because is not using the override methods from CFileInputStream.

lib/filesystem/CArchiveLoader.cpp
krs0: IVCMIDirs::userExtractedPath() created

mapeditor/mainwindow.h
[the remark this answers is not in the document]
krs0: Removed from MainWindow Class

mapeditor/mainwindow.cpp
```diff
@@ -139,6 +175,8 @@ MainWindow::MainWindow(QWidget *parent) :
 	
 	graphics = new Graphics(); // should be before curh->init()
 	graphics->load();//must be after Content loading but should be in main thread
+
+	ConvertExtractedResourceFiles(splitDefs, convertPcxToPng, deleteOriginals);
```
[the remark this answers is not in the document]
krs0: ResourceExtractor code moved to a static class.

lib/filesystem/CArchiveLoader.cpp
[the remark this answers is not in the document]
Nordsoft91: so why not cast it in the same way

mapeditor/mainwindow.cpp
```diff
@@ -88,57 +91,85 @@ void MainWindow::saveUserSettings()
...
 	logGlobal->debug("settings = %s", settings.toJsonNode().toJson());
-	
+
 	// Some basic data validation to produce better error messages in cases of incorrect install
 	auto testFile = [](std::string filename, std::string message) -> bool
 	{
 		if (CResourceHandler::get()->existsResource(ResourceID(filename)))
 			return true;
-		
+
```
Nordsoft91, line 152: those empty lines are polluting review and excessive merge conflicts generators

mapeditor/resourceExtractor/ResourceConverter.h
```diff
@@ -0,0 +1,49 @@
...
+	ConversionOptions conversionOptions;
+};
+
+class ResourceConverter
+{
+
+public:
+
+	// Splits def files that are shared between factions and converts pcx to bmp depending on Extraction Options
+	static void convertExtractedResourceFiles(ConversionOptions conversionOptions);
```
Nordsoft91, line 35: Why just not make flags from ConversionOptions as fields of ResourceConverter?
Basically class methods are implicit calls of functions.
```
class Foo
{
    void bar();
};
```
is equivalent to 
```
class Foo {};
void bar(Foo *);

File: mapeditor/mainwindow.cpp [hunk 33]
```diff
@@ -88,57 +91,85 @@ void MainWindow::saveUserSettings()
...
+		});
+
+	parser.process(qApp->arguments());
+
+	const QStringList positionalArgs = parser.positionalArguments();
+
+	if (!positionalArgs.isEmpty())
+		mapFilePath = positionalArgs.at(0);
+
+	extractionOptions = {parser.isSet("e"), {parser.isSet("s"),	 parser.isSet("c"),	 parser.isSet("d")}};
```
Nordsoft91, line 114: Seems like formatting issues, probably you want something like that
```suggestion
	extractionOptions = {
		parser.isSet("e"), {
			parser.isSet("s"),
			parser.isSet("c"),
			parser.isSet("d")}};
```

mapeditor/mainwindow.cpp
```diff
@@ -88,57 +91,85 @@ void MainWindow::saveUserSettings()
...
+		{"s", QCoreApplication::translate("main", "From an extracted archive, it Splits TwCrPort, CPRSMALL, FlagPort, ITPA, ITPt, Un32 and Un44 into individual PNG's.")},
+		{"c", QCoreApplication::translate("main", "From an extracted archive, Converts single Images (found in Images folder) from .pcx to png.")},
+		{"d", QCoreApplication::translate("main", "Delete original files, for the ones splitted / converted.")},
+		});
+
+	parser.process(qApp->arguments());
+
+	const QStringList positionalArgs = parser.positionalArguments();
+
+	if (!positionalArgs.isEmpty())
```
Nordsoft91, line 111:
```suggestion
	if(!positionalArgs.isEmpty())
```

mapeditor/mainwindow.cpp
```diff
@@ -38,9 +38,12 @@
 #include "mapsettings.h"
 #include "playersettings.h"
 #include "validator.h"
+#include "resourceExtractor/ResourceConverter.h"
 
 static CBasicLogConfigurator * logConfig;
 
+ExtractionOptions extractionOptions;
```
Nordsoft91, line 45: please consider alternatives to having global variable, maybe something like that
```
void MainWindow::parseCommandLine(ExtractionOptions & ops)
{
...
}

MainWindow::MainWindow(QWidget * parent): ...
{
    ...
    ExtractionOptions extractionOptions;
    parseCommandLine(extractionOptions);
...
    preInitDLL(::console, false, extractionOptions.extractArchives);
}
```

Nordsoft91 (commented):
Let me know when you think you done with code improvements, I'll remove request changes badge

mapeditor/mainwindow.cpp
krs0: I do not get it :). Can you please elaborate a little?

lib/filesystem/CArchiveLoader.cpp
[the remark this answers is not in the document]
krs0: Done

mapeditor/resourceExtractor/ResourceConverter.h
krs0: I do not see ResourceConverter as a class with the functionality that it has right now. In the end it is just a simple function call. I grouped all functions in a static class and made only convertExtractedResourceFiles() public so that usage intention is clear.
I do not see the benefits of adding the fields to the class.

mapeditor/mainwindow.cpp
krs0: Done

mapeditor/mainwindow.cpp
krs0: Done like suggested.

mapeditor/Animation.cpp
krs0: I propose to leave the current implementation with boost. I would rather skip this step and learn directly the quirks of std::filesystem::path that we will get with C++17 than the QT's.

krs0:
@Nordsoft91 from my side this is it. Please take a look on the remaining open conversations, maybe some items still come up from there.
Thank you very much! for the time you took for the review.

Nordsoft91 (approved):

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+	bfs::path imagesPath = VCMIDirs::get().userExtractedPath() / "IMAGES";
+	bfs::directory_iterator end_iter;
+
+	for(bfs::directory_iterator dir_itr(imagesPath); dir_itr != end_iter; ++dir_itr)
+	{
+		try
+		{
+			if (!bfs::is_regular_file(dir_itr->status()))
+				return;
```
kambala-decapitator, line 46: @krs0 is it correct to use `return` instead of `continue` here?

krs0: Yup, its a mistake. Works anyway because we have only images, but makes no sense. Will change to continue

Issue #1102, opened 2022-11-03
NKAI does nothing


Nordsoft91:
NKAI has few heroes standing on their positions whole game.
See attached save (vcmieagles) 
[NKAI_does_nothing.zip](https://github.com/vcmi/vcmi/files/9931307/NKAI_does_nothing.zip)

Here is possible army to purchase after capturing its town (resourceful mod was enabled, no issues with gold)
<img width="754" alt="Снимок экрана 2022-11-03 в 20 36 20" src="https://user-images.githubusercontent.com/9308612/199781853-a06deb16-d393-46a1-ac52-d4b30f4e5b75.png">

PR #1121, opened 2022-11-10
#1102 - NKAI: one more freeze - no mains with strong army

nullkiller:


AI/Nullkiller/Behaviors/BuyArmyBehavior.cpp
```diff
@@ -57,8 +57,7 @@ Goals::TGoalVec BuyArmyBehavior::decompose() const
 				continue;
 			}
 
-			if(ai->nullkiller->heroManager->getHeroRole(targetHero) == HeroRole::MAIN
-				&& targetHero->getArmyStrength() >= 300)
```
nullkiller, line 61: Removed a hack. A way more complicated logic is actually needed here

Nordsoft91: am I correctly understand that now all army will be consolidated under single hero?

nullkiller: No, but this thing prevented AI from buying army from town. The idea was to prefer types of army suitable for main hero. This needs better logic.

PR #1125, opened 2022-11-12
Fix [3144]

GermanAizek:
Bug in bug tracker: https://bugs.vcmi.eu/view.php?id=3144

### DIFF
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -489,7 +489,7 @@ void CServerHandler::sendMessage(const std::string & txt) const
 		std::string connectedId, playerColorId;
 		readed >> connectedId;
 		readed >> playerColorId;
-		if(connectedId.length(), playerColorId.length()) // BUG https://bugs.vcmi.eu/view.php?id=3144
+		if(connectedId.length() && playerColorId.length())
 		{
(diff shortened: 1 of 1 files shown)


client/CServerHandler.cpp
```diff
@@ -489,7 +489,7 @@ void CServerHandler::sendMessage(const std::string & txt) const
 		std::string connectedId, playerColorId;
 		readed >> connectedId;
 		readed >> playerColorId;
-		if(connectedId.length(), playerColorId.length()) // BUG https://bugs.vcmi.eu/view.php?id=3144
+		if(connectedId.length() < playerColorId.length())
```
Nordsoft91, line 492: Discussed here: https://github.com/vcmi/vcmi/pull/1111
This change is inappropriate

Nordsoft91 (changes requested):

client/CServerHandler.cpp
Nordsoft91, line 492:
```suggestion
		if(connectedId.length() && playerColorId.length())
```

GermanAizek: @Nordsoft91 fix it https://github.com/vcmi/vcmi/pull/1125/commits/79bc82af63ff60cfb7b41a5b944535d8fd5e46c4

Issue #1100, opened 2022-11-03
NKAI ​​is trying to buy creatures from an ally's dwellings.


Povelitelh3:
He also tries to capture mines, but this is less common.

PR #1126, opened 2022-11-13
#1100 - fix visiting allied dwellings

nullkiller:
We can not buy army in ally dwelling

### DIFF
--- a/AI/Nullkiller/AIUtility.cpp
+++ b/AI/Nullkiller/AIUtility.cpp
@@ -237,7 +237,7 @@ bool isObjectPassable(const Nullkiller * ai, const CGObjectInstance * obj)
 
 bool isObjectPassable(const CGObjectInstance * obj)
 {
-	return isObjectPassable(obj, ai->playerID, cb->getPlayerRelations(obj->tempOwner, ai->playerID));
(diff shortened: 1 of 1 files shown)


AI/Nullkiller/AIUtility.cpp
```diff
@@ -344,11 +344,14 @@ uint64_t timeElapsed(std::chrono::time_point<std::chrono::high_resolution_clock>
...
 bool shouldVisit(const Nullkiller * ai, const CGHeroInstance * h, const CGObjectInstance * obj)
 {
+	auto relations = ai->cb->getPlayerRelations(obj->tempOwner, h->tempOwner);
+
 	switch(obj->ID)
 	{
 	case Obj::TOWN:
 	case Obj::HERO: //never visit our heroes at random
-		return obj->tempOwner != h->tempOwner; //do not visit our towns at random
+		return relations == PlayerRelations::ENEMIES; //do not visit our towns at random
```
Nordsoft91, line 353: Hm, but what if AI would like to fulfill mana by staying in town overnight or just learn spells?

Nordsoft91 (approved):

AI/Nullkiller/AIUtility.cpp
nullkiller: It can not for now) Also a bit more sophisticated logic is needed to understand that we can learn some spells in town.

PR #1130, opened 2022-11-14
Fixes #1096 - do not propose banned skills on levelup

### LINKED ISSUE #1096: Ranger heroes being offered to learn Fire Magic when levelling up
Playing with VCMI 1.0.0 – no mods except for VCMI extras – and the Rampart, my Ranger hero has already got Fire Magic twice as a choice when levelling up. In my experience, this never happens in the original game because Rangers don’t have any chance of learning that secondary skill : https://heroes.thelazy.net/index.php/Fire_Magic (However, I don’t see these probabilities in the original manual)

IvanSavenko:
Remove possibility to get banned skill on levelup as "obligatory skill" if all other such skills have been learned before.

May happen under some conditions, e.g. if hero quickly learns other magic schools - in witch hut/university, leading to case where banned skill is the only obligatory skill that can be offered on levelup

Fixes #1096

### DIFF
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -1117,7 +1117,9 @@ std::vector<SecondarySkill> CGHeroInstance::getLevelUpProposedSecondarySkills() 
 	std::vector<SecondarySkill> obligatorySkills; //hero is offered magic school or wisdom if possible
 	if (!skillsInfo.wisdomCounter)
 	{
-		if (cb->isAllowed(2, SecondarySkill::WISDOM) && !getSecSkillLevel(SecondarySkill::WISDOM))
+		if (cb->isAllowed(2, SecondarySkill::WISDOM) &&
+				!getSecSkillLevel(SecondarySkill::WISDOM) &&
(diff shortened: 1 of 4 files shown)


lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1117,7 +1117,9 @@ std::vector<SecondarySkill> CGHeroInstance::getLevelUpProposedSecondarySkills()
 	std::vector<SecondarySkill> obligatorySkills; //hero is offered magic school or wisdom if possible
 	if (!skillsInfo.wisdomCounter)
 	{
-		if (cb->isAllowed(2, SecondarySkill::WISDOM) && !getSecSkillLevel(SecondarySkill::WISDOM))
+		if (cb->isAllowed(2, SecondarySkill::WISDOM) &&
```
nullkiller, line 1120: I see in all 3 cases we use both isAllowed and heroClass together. Just an idea, what if we create an overload for isAllowed and pass hero instance iinside?

IvanSavenko: Yeah, makes sense. Although probably not override but method of Hero class, similar to existing
CGHeroInstance::canLearnSpell

PR #1143, opened 2022-11-15
Fixed bugs and memleaks

GermanAizek:
I'm waiting for code review

### DIFF
--- a/AI/BattleAI/AttackPossibility.cpp
+++ b/AI/BattleAI/AttackPossibility.cpp
@@ -175,7 +175,7 @@ AttackPossibility AttackPossibility::evaluate(const BattleAttackInfo & attackInf
 				if (!attackInfo.shooting && defenderState->ableToRetaliate() && !counterAttacksBlocked)
 				{
 					damageReceived = averageDmg(retaliation);
-					attackerDamageReduce = calculateDamageReduce(defender, attacker, damageReceived, state);
+					attackerDamageReduce = calculateDamageReduce(attacker, defender, damageReceived, state);
 					defenderState->afterAttack(attackInfo.shooting, true);
 				}
 
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -592,7 +592,7 @@ void AIGateway::showBlockingDialog(const std::string & text, const std::vector<C
 
 					logAi->trace("Guarded object query hook: %s by %s danger ratio %f", target.toString(), hero.name, ratio);
 
-					if(text.find("guarded") >= 0 && (dangerUnknown || dangerTooHigh))
+					if(text.find("guarded") && (dangerUnknown || dangerTooHigh))
 						answer = 0; // no
 				}
 			}
--- a/client/mapHandler.cpp
+++ b/client/mapHandler.cpp
@@ -567,7 +567,7 @@ void CMapHandler::CMapWorldViewBlitter::drawOverlayEx(SDL_Surface * targetSurf)
 			continue;
 
 		realPos.x = initPos.x + (iconInfo.pos.x - topTile.x) * tileSize;
-		realPos.y = initPos.x + (iconInfo.pos.y - topTile.y) * tileSize;
+		realPos.y = initPos.y + (iconInfo.pos.y - topTile.y) * tileSize;
 
 		auto wvIcon = this->objectToIcon(iconInfo.id, iconInfo.subId, iconInfo.owner);
 
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -219,8 +219,10 @@ bool CHeroArtPlace::askToAssemble(const CArtifactInstance *art, ArtifactPosition
 			std::bind(&CCallback::assembleArtifacts, LOCPLINT->cb.get(), hero, slot, true, combination->id));
 
 		if(assemblyPossibilities.size() > 2)
+		{
 			logGlobal->warn("More than one possibility of assembling on %s... taking only first", art->artType->getName());
-		return true;
+			return true;
+		}
 	}
 	return false;
 }
--- a/client/windows/GUIClasses.cpp
+++ b/client/windows/GUIClasses.cpp
@@ -1032,7 +1032,7 @@ std::function<void()> CExchangeController::onSwapArmy()
 	{
 		GsThread::run([=]
 		{
-			if(right->tempOwner != cb->getMyColor()
+			if(left->tempOwner != cb->getMyColor()
 				|| right->tempOwner != cb->getMyColor())
 			{
 				return;
--- a/lib/mapping/CMapService.cpp
+++ b/lib/mapping/CMapService.cpp
@@ -44,7 +44,7 @@ std::unique_ptr<CMap> CMapService::loadMap(const ui8 * buffer, int size, const s
 
 	//might be original campaign and require patch
 	getMapPatcher(name)->patchMapHeader(header);
-	header.release();
+	header.reset();
 
 	return map;
 }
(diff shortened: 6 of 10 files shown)


AI/BattleAI/AttackPossibility.cpp
```diff
@@ -175,7 +175,7 @@ AttackPossibility AttackPossibility::evaluate(const BattleAttackInfo & attackInf
 				if (!attackInfo.shooting && defenderState->ableToRetaliate() && !counterAttacksBlocked)
 				{
 					damageReceived = averageDmg(retaliation);
-					attackerDamageReduce = calculateDamageReduce(defender, attacker, damageReceived, state);
```
nullkiller, line 178: Why do you think it is a bug?

client/mapHandler.cpp
```diff
@@ -567,7 +567,7 @@ void CMapHandler::CMapWorldViewBlitter::drawOverlayEx(SDL_Surface * targetSurf)
 			continue;
 
 		realPos.x = initPos.x + (iconInfo.pos.x - topTile.x) * tileSize;
-		realPos.y = initPos.x + (iconInfo.pos.y - topTile.y) * tileSize;
```
nullkiller, line 570: And this is really interesting)

AI/BattleAI/AttackPossibility.cpp
nullkiller: Retalitation is when defender attacks attacker

AI/Nullkiller/AIGateway.cpp
```diff
@@ -592,7 +592,7 @@ void AIGateway::showBlockingDialog(const std::string & text, const std::vector<C
 
 					logAi->trace("Guarded object query hook: %s by %s danger ratio %f", target.toString(), hero.name, ratio);
 
-					if(text.find("guarded") >= 0 && (dangerUnknown || dangerTooHigh))
+					if(text.find("guarded") && (dangerUnknown || dangerTooHigh))
```
Nordsoft91, line 595: This is wrong. Find method returns npos for non-success

client/widgets/CArtifactHolder.cpp
```diff
@@ -219,8 +219,10 @@ bool CHeroArtPlace::askToAssemble(const CArtifactInstance *art, ArtifactPosition
 			std::bind(&CCallback::assembleArtifacts, LOCPLINT->cb.get(), hero, slot, true, combination->id));
 
 		if(assemblyPossibilities.size() > 2)
+		{
 			logGlobal->warn("More than one possibility of assembling on %s... taking only first", art->artType->getName());
-		return true;
+			return true;
+		}
```
Nordsoft91, line 225: Change breaks assembly functionality completely

client/windows/GUIClasses.cpp
```diff
@@ -1032,7 +1032,7 @@ std::function<void()> CExchangeController::onSwapArmy()
 	{
 		GsThread::run([=]
 		{
-			if(right->tempOwner != cb->getMyColor()
+			if(left->tempOwner != cb->getMyColor()
```
Nordsoft91, line 1035: This probably correct

lib/rmg/Zone.h
```diff
@@ -119,7 +119,7 @@ class Zone : public rmg::ZoneOptions
 	template<class T>
 	void addModificator()
 	{
-		modificators.emplace_back(new T(*this, map, generator));
+		modificators.push_back(std::make_unique<T>(*this, map, generator));
```
Nordsoft91, line 122: Why is this problem?

server/CGameHandler.cpp
```diff
@@ -1029,7 +1029,7 @@ void CGameHandler::makeAttack(const CStack * attacker, const CStack * defender,
 		}
 		if (VLC->modh->settings.data["hardcodedFeatures"]["NEGATIVE_LUCK"].Bool()) // negative luck enabled
 		{
-			if (attackerLuck < 0 && getRandomGenerator().nextInt(23) < abs(attackerLuck))
+			if (attackerLuck != INT_MIN && attackerLuck < 0 && getRandomGenerator().nextInt(23) < abs(attackerLuck))
```
Nordsoft91, line 1032: Why luck can be INT_MIN? :)

Nordsoft91 (changes requested):
Thanks for running vcmi through static analyzer, but please do at least minimal testing of impacted functionality. Some changes are obviously wrong

lib/mapping/CMapService.cpp
```diff
@@ -44,7 +44,7 @@ std::unique_ptr<CMap> CMapService::loadMap(const ui8 * buffer, int size, const s
 
 	//might be original campaign and require patch
 	getMapPatcher(name)->patchMapHeader(header);
-	header.release();
+	header.reset();
```
Nordsoft91, line 47: Looks like header will be removed from map. Did you try to run game?

AI/Nullkiller/AIGateway.cpp
IvanSavenko: But npos is a positive number. So both old and new code are wrong

server/CGameHandler.cpp
nullkiller: probably abs(INT_MIN) gives something bad and this is why it triggered a warning.

lib/mapping/CMapService.cpp
nullkiller: Can we use some std::move magic

server/CGameHandler.cpp
kambala-decapitator: `abs(INT_MIN)` would be an overflow as it would try to fit the value in `int`

lib/rmg/Zone.h
kambala-decapitator: I think it's related to the usage of naked `new` which is usually not a good idea in modern C++

AI/Nullkiller/AIGateway.cpp
kambala-decapitator: yep, correct way is `!= std::string::npos`

client/mapHandler.cpp
kambala-decapitator: I believe the right fix is to use `initPos.y` on this line

AI/BattleAI/AttackPossibility.cpp
IvanSavenko: But this is code is for *retaliation*, so defender and attacker are now swapped.

Wrong, no changes needed here

server/CGameHandler.cpp
IvanSavenko: Noise, would never happen:
`const int attackerLuck = attacker->LuckVal();`
And LuckVal can only return values between -3 and +3

lib/rmg/Zone.h
nullkiller: I like to use new because VS intellisence works better with it than with std::make_unique/std::make_shared. So ptr.reset(new Something()) works better for me

kambala-decapitator: > I like to use new because VS intellisence works better with it than with std::make_unique/std::make_shared. So ptr.reset(new Something()) works better for me

well, it's a problem of IDE and is not an excuse to avoid good code. Install Visual Assist, life will become much better ;)

lib/mapping/CMapService.cpp
IvanSavenko: Original code is... somewhat weird - it creates unique_ptr that owns data already owned by another unique_ptr.
Because of that, this unique_ptr needs to be `release`d before going out of scope.

Wrong, no changes needed here

lib/rmg/Zone.h
Nordsoft91: It's not naked new, it emplaces unique_ptr. Change here doesn't fix any bug or memory leak, it just trigger for endless holy war discussion

IvanSavenko:
Since author made no attempt to reply on reviews I would like to close this PR and submit fix for confirmed issues as part of my PR. 
Any objections/better suggestions?

PR #1146, opened 2022-11-15
Implemented ray-like projectiles for shooters

### LINKED ISSUE #834: Make Beholders temporary attack graphics more friendly.
Because VCMI hasn't yet implemented rays (Magi and Beholders attack), we have (temporarily) replacements. And for now, Beholders shoot... ballista belts that looks ridiculous. It's better give them something that doesn't hurt the eye, for example:

```
			"missile" :
			{
				"projectile": "PMAGEX.DEF"
			}
```

BTW. It would be cool to have some (colourful one?) rays in vcmi.

IvanSavenko:
- Added missing support for ray-like projectiles
- Archmages, Evil Eyes and Beholders now use ray for shooting
- New method to draw 1 pixel-wide line with color gradient at arbitrary angle
- fixed incorrect attackClimaxFrame field for Archmages
- closes #834 as no longer actual

Configuration (for modders, to be copied to wiki after merge):
```
	"ray" : // indicates that creature uses ray animation for ranged attacks instead of missile image
	[
		{ // definition of first (top-most) line in the ray
			"start" : [ 160, 192, 0, 255 ], // color (RGBA components) of ray at starting point
			"end" : [ 160, 192, 0,  64 ]  // color (RGBA components) of ray at finishing point
		},
		{}, // definition of second from top line in the ray, identical format
		... // definitions of remaining lines, till desired width of the ray
	]
```


client/battle/CBattleInterface.cpp
```diff
@@ -3206,18 +3213,58 @@ void CBattleInterface::showProjectiles(SDL_Surface *to)
 				continue; // wait...
 		}
 
-		size_t group = it->reverse ? 1 : 0;
-		auto image = idToProjectile[it->creID]->getImage(it->frameNum, group, true);
+		if ( idToProjectile.count(it->creID) != 0)
```
nullkiller, line 3216: We do not use spaces after if, for, while according to style guidelines

```suggestion
		if(idToProjectile.count(it->creID) != 0)

```

client/gui/SDL_Extensions.cpp
```diff
@@ -361,6 +361,74 @@ void CSDL_Ext::update(SDL_Surface * what)
...
+
+		uint8_t r = lerp(color1.r, color2.r, f);
+		uint8_t g = lerp(color1.g, color2.g, f);
+		uint8_t b = lerp(color1.b, color2.b, f);
+		uint8_t a = lerp(color1.a, color2.a, f);
+
+		Uint8 *p = CSDL_Ext::getPxPtr(sur, x, y);
+		ColorPutter<4, 0>::PutColor(p, r,g,b,a);
+	}
+}
```
Nordsoft91, line 402: drawLineX and drawLineY are equivalent, so instead of duplicating you can use one with swapped x y:

drawLine(sur, x1, y1, x2, y2, color1, color2);
drawLine(sur, y1, x1, y2, x2):

config/schemas/creature.json
```diff
@@ -217,14 +217,18 @@
...
 					"properties":{
 						"projectile": {
 							"type":"string",
 							"description": "Path to projectile animation",
 							"format" : "defFile"
 						},
+						"ray": {
+							"type":"array",
+							"description": "Colors of ray projectile animation"
+						},
```
Nordsoft91, line 231: Could you please make schema full? Imean, add "items" and their subitems too

client/gui/SDL_Extensions.cpp
```diff
@@ -361,6 +361,74 @@ void CSDL_Ext::update(SDL_Surface * what)
 	if(0 !=SDL_UpdateTexture(screenTexture, nullptr, what->pixels, what->pitch))
 		logGlobal->error("%s SDL_UpdateTexture %s", __FUNCTION__, SDL_GetError());
 }
+
+uint8_t lerp(uint8_t a, uint8_t b, float f)
+{
+	return a + std::round((b-a)*f);
```
Nordsoft91, line 367: Please add spaces around binary operations 
```suggestion
	return a + std::round((b - a) * f);
```

client/battle/CBattleInterface.cpp
```diff
@@ -3206,18 +3213,58 @@ void CBattleInterface::showProjectiles(SDL_Surface *to)
...
+				dst.h = image->height();
+				dst.w = image->width();
+				dst.x = static_cast<int>(it->x - dst.w / 2);
+				dst.y = static_cast<int>(it->y - dst.h / 2);
 
-		if(image)
+				image->draw(to, &dst, nullptr);
+			}
+		}
+		if (idToRay.count(it->creID) != 0)
```
Nordsoft91, line 3232:
```suggestion
		if(idToRay.count(it->creID))
```

lib/CCreatureHandler.h
```diff
@@ -11,6 +11,7 @@
 
 #include <vcmi/Creature.h>
 #include <vcmi/CreatureService.h>
+#include <int3.h>
```
Nordsoft91, line 14: Why it's needed?

nullkiller (approved):

client/gui/SDL_Extensions.cpp
IvanSavenko: They are not equivalent - CSDL_Ext::getPxPtr needs non-swapped x&y

nullkiller:
Animation is quite quick. I hardly can see the ray growing to full length. What if we wait a bit after it has grown before removing animation so that the end user can see it full?

IvanSavenko:
> Animation is quite quick. I hardly can see the ray growing to full length. What if we wait a bit after it has grown before removing animation so that the end user can see it full?

I get what you're saying but this is behavior is identical to normal projectiles and does not seems to be different to H3.
Will recheck and see if I can improve this somehow.

PR #1151, opened 2022-11-18
Fix loading images which aren't divided by 4

Nordsoft91:
Thanks to @IvanSavenko for finding the fix for problems

### DIFF
--- a/mapeditor/BitmapHandler.cpp
+++ b/mapeditor/BitmapHandler.cpp
@@ -117,7 +117,7 @@ namespace BitmapHandler
 			{
 				logGlobal->error("Failed to open %s as H3 PCX!", fname);
 			}
-			return image;
+			return image.copy(); //copy must be returned here because buffer readFile.first used to build QImage will be cleaned after this line
 		}
 		else
 		{ //loading via QImage
--- a/mapeditor/BitmapHandler.cpp
+++ b/mapeditor/BitmapHandler.cpp
@@ -81,7 +81,7 @@ namespace BitmapHandler
 		}
 		else
 		{
-			QImage image(pcx + it, width, height, width * 4, QImage::Format_RGB32);
+			QImage image(pcx + it, width, height, width * 3, QImage::Format_RGB888);
 			return image;
 		}
 	}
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Later I with krs found two more cases like this, both are in `ImageLoader::init`
wrong:
`memset((void *)image->bits(), 0, fullSize.y() * fullSize.x());`
correct:
`memset((void *)image->bits(), 0, fullSize.y() * image->bytesPerLine());`

wrong:
`lineStart += margins.y() * image->bytesPerLine() + margins.x();`
correct:

krs0:
Also .pcx single images have some issues I suppose...
Look at last coment in 
VCMI H3 original archives extractor #1118

IvanSavenko:
Yes, pretty much identical bug in `BitmapHandler::loadH3PCX` (BitmapHandler.cpp)
Line 66 right now is;
`QImage image(pcx + it, width, height, QImage::Format_Indexed8);`
should use different constructor:
`QImage image(pcx + it, width, height, width, QImage::Format_Indexed8);`

Line 84 is now:
`QImage image(pcx + it, width, height, QImage::Format_RGB32);`
Should also use different constructor:
`QImage image(pcx + it, width, height, width * 3, QImage::Format_RGB32);`

Somebody ( @krs0  ?) please check whether my assumption is correct.

Nordsoft91:
@IvanSavenko , suspecting last should be
`QImage image(pcx + it, width, height, width * 4, QImage::Format_RGB32);`
right? RGB32 is containing alpha

IvanSavenko:
Oh, yeah. But actually not exactly. Both versions are wrong.
Should be:
`QImage image(pcx + it, width, height, width * 3, QImage::Format_RGB888);`
H3 PCX data is store in RGB format, 24 bits per pixel. Qt provides such format as Format_RGB888.

krs0:
and please add also for both ifs `return image.copy();` as per Ivan

mapeditor/BitmapHandler.cpp
```diff
@@ -81,7 +81,7 @@ namespace BitmapHandler
 		}
 		else
 		{
-			QImage image(pcx + it, width, height, QImage::Format_RGB32);
+			QImage image(pcx + it, width, height, width * 3, QImage::Format_RGB888);
```
nullkiller, line 84: 3 is 3 bytes?

nullkiller (approved):
Looks good to merge

mapeditor/BitmapHandler.cpp
Nordsoft91: Yes, that's byter per line as I understand

PR #1153, opened 2022-11-20
Clickable status bar to send messages with on mobile

nullkiller:
It is possible to start typing message by click on status bar. So I am not sure it will work for Android. Will test later.

### DIFF
--- a/client/widgets/TextControls.h
+++ b/client/widgets/TextControls.h
@@ -124,6 +124,11 @@ class CGStatusBar : public CLabel, public std::enable_shared_from_this<CGStatusB
 protected:
 	Point getBorderSize() override;
 
+	void clickLeft(tribool down, bool previousState) override;
+
+private:
+	std::function<void()> onClick;
+
 public:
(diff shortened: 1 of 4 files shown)


kambala-decapitator:
FYI on iOS you can trigger console with pinch gesture.

if this one would work, then the pinch gesture can be removed I think.

nullkiller:
Android works with it. Now you can tap status bar and send a message. Or cheat code. iOS should work as well I think.

client/widgets/TextControls.h
```diff
@@ -138,6 +143,8 @@ class CGStatusBar : public CLabel, public std::enable_shared_from_this<CGStatusB
 	void show(SDL_Surface * to) override; //shows statusbar (with current text)
 
 	void lock(bool shouldLock); //If true, current text cannot be changed until lock(false) is called
+
+	void setOnClick(std::function<void()> handler);
```
Nordsoft91, line 147: In such case I would make onClick as public and get rid of this method

PR #1154, opened 2022-11-21
Minor improvements & fixes to music player

IvanSavenko:
Miscellaneous issues I've noticed with music player while looking into freezes. 
Nothing critical, can be reviewed/merged without hurry

- music will be selected correctly after rapidly leaving and reentering town screen (while old track in fading out)
- music sets are now stored as single std::map instead of unnecessary "map of maps" (which also had bugs in implementation)
- fixed computation of resume time for restarted/looped over music tracks
- updated & clarified some comments
- converted C functions for SDL_Mixer callbacks into lambdas
- formatting fixes
- also moved locale manipulation in VCMIDirs to Windows-only section - this is only necessary under Windows which does not provide utf8-compatible locale by default

### DIFF
--- a/lib/VCMIDirs.cpp
+++ b/lib/VCMIDirs.cpp
@@ -171,6 +171,7 @@ class VCMIDirsWIN32 final : public IVCMIDirs
 void VCMIDirsWIN32::init()
 {
 	std::locale::global(boost::locale::generator().generate("en_US.UTF-8"));
+	boost::filesystem::path::imbue(std::locale());
 
 	// Call base (init dirs)
 	IVCMIDirs::init();
--- a/lib/VCMIDirs.cpp
+++ b/lib/VCMIDirs.cpp
@@ -170,6 +170,8 @@ class VCMIDirsWIN32 final : public IVCMIDirs
 
 void VCMIDirsWIN32::init()
 {
+	std::locale::global(boost::locale::generator().generate("en_US.UTF-8"));
+
 	// Call base (init dirs)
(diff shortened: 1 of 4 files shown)


lib/VCMIDirs.cpp
```diff
@@ -170,6 +170,9 @@ class VCMIDirsWIN32 final : public IVCMIDirs
 
 void VCMIDirsWIN32::init()
 {
+	std::locale::global(boost::locale::generator().generate("en_US.UTF-8"));
+	boost::filesystem::path::imbue(std::locale());
```
Nordsoft91, line 174: Seems like this line was called for all platforms, except windows. Now for win32 only.

nullkiller (approved):

lib/VCMIDirs.cpp
IvanSavenko: This was incorrectly placed ifdef.
If you look up history you'll see that initially this code was added for Windows, and later - first line was gradually removed from other platforms. However it was removed incorrectly - only first line was placed into ifdef

This code is necessary on Windows in order to access files with non-ascii symbols in path (e.g. cyrillic, chinese) using utf-8 since Windows API only allows such access through its utf-16 API (aka wide/unicode API) .
Windows behavior is unique - all other OS's use utf-8 and don't require any special code to allow such access.

SoundSSGood: Does it have something to do with it https://github.com/vcmi/vcmi/pull/873 ?

PR #1156, opened 2022-11-22
Online mode

Nordsoft91:
Please, do commitable suggestions for formatting/code style comments

- [x] New game/Join game should be New Room/Join Room
- [x] "User" should be "Username"
- [x] Address/Port should be merged into one (beholder.vcmi.eu:5002) and renamed to "Server"
- [x] unusable controls should be greyed-out or hidden (New game/join game before connection, address/port/username after connection, potentially - chat window)
- [x] ~~[Low Priority] There should be a way to fix mods mismatch directly in this window, e.g. add button to install selected mod(s).~~ https://github.com/vcmi/vcmi/issues/1170
 - [x] ~~[Low Priority] Also, launcher should reduce size of icons in the left panel if screen is too small to display all of them at once. For desktop, minimal window height should also be locked to avoid scrolling.~~ https://github.com/vcmi/vcmi/issues/1169
 - [x] "Start game" should be locked if player is in lobby room
 - [x] ~~Host should not have "ready" button but "Launch" instead (or remove it entirely - and start game from bottom-left button.~~ updated logic for game starting - all players should confirm their readiness
 - [x] There should be indication, e.g. checkbox/X mark to show readied players
 - [x] ~~[Low Priority] In-game menu should be adapted to game lobby: For hosts, new game / load game should immediately go to scenario selection/game load screens, skipping all screens in between, instead of asking player to use menu in "one right way". For guests, game should go directly to chosen mode (and probably only allow to start game once host made choice between new game and load game)~~ https://github.com/vcmi/vcmi/issues/1171


IvanSavenko:
Code-wise it is okay I guess, but user experience-wise it is rather poor in my opinion:
However lobby functionality itself is... very basic and not user-friendly...

- New game/Join game should be New Room/Join Room, "User" should be "Username", Address/Port should be merged into one (beholder.vcmi.eu:5002) and renamed to "Server"
- unusable controls should be greyed-out or hidden (New game/join game before connection, address/port/username after connection, potentially - chat window)
- There should be a way to fix mods mismatch directly in this window, e.g. add button to install selected mod(s).
- Also, launcher should reduce size of icons in the left panel if screen is too small to display all of them at once. For desktop, minimal window height should also be locked to avoid scrolling.
- "Start game" should be locked if player is in lobby room
- Host should not have "ready" button but "Launch" instead (or remove it entirely - and start game from bottom-left button.
- There should be indication, e.g. checkbox/X mark to show readied players
- In-game menu should be adapted to game lobby: For hosts, new game / load game should immediately go to scenario selection/game load screens, skipping all screens in between, instead of asking player to use menu in "one right way". For guests, game should go directly to chosen mode (and probably only allow to start game once host made choice between new game and load game)

client/CServerHandler.cpp
```diff
@@ -190,9 +193,17 @@ void CServerHandler::startLocalServerAndConnect()
...
+	if(settings["session"]["lobby"].Bool() && settings["session"]["host"].Bool())
+	{
+		args.push_back("--lobby=" + settings["session"]["address"].String());
+		args.push_back("--connections=" + settings["session"]["hostConnections"].String());
+		args.push_back("--lobby-port=" + boost::lexical_cast<std::string>(settings["session"]["port"].Integer()));
+		args.push_back("--lobby-uuid=" + settings["session"]["hostUuid"].String());
+	}
+	threadRunLocalServer = std::make_shared<boost::thread>([&cond, args, this] {
 		setThreadName("CVCMIServer");
-		CVCMIServer::create(&cond, uuid);
```
nullkiller, line 195: What was this uuid about?

launcher/lobby/lobby_moc.cpp
```diff
@@ -0,0 +1,408 @@
...
+		protocolAssert(args.size());
+		chatMessage("System error", args[0], true);
+		if(authentificationStatus == AuthStatus::AUTH_NONE)
+			authentificationStatus = AuthStatus::AUTH_ERROR;
+		break;
+
+	case CREATED:
+		protocolAssert(args.size());
+		hostSession = args[0];
+		session = args[0];
```
nullkiller, line 91: both args[0]? Is it because it is host?

client/CServerHandler.cpp
Nordsoft91: It's command line argument. It's still here

launcher/lobby/lobby_moc.cpp
Nordsoft91: It means that you are hosting session and you are joined to the same session

IvanSavenko: Are command args explained/documented somewhere?
args[0] is not very descriptive name.
Question does for all commands, not just CREATED

Nordsoft91: In the header there is description for server commands and their arguments

nullkiller (approved):

IvanSavenko (approved):

Issue #960, opened 2022-09-21
Fire Shield effect doesn't play sound


kambala-decapitator:

kambala-decapitator:
the sound is explicitly disabled because there're 2 sounds it seems:
https://github.com/vcmi/vcmi/blob/35f2fb01ab96b4bc0382ca9d36a5a9f94543a586/config/spells/timed.json#L74-L76

what would be the right way to solve that?

Issue #862, opened 2022-09-08
no animation of Pit Lords summoning ability


kambala-decapitator:
Summoning itself works

misiokles:
Yeah, some special animations are still missing. Like Bloodlust or turning to stone.

Issue #791, opened 2022-09-03
wrong icons of arrow towers


kambala-decapitator:
Conflux towers have storm els, but in the turn order they have magma els images (the 3x1 stacks): https://imgur.com/a/QgvV8Gb

haven't checked other towns.

Nordsoft91:
Was fixed earlier

IvanSavenko:
Reopening, still not resolved in develop branch.
Cause of bug is lack of icon for creature 149 (CreatureID::ARROW_TOWERS).

- Solution A: Provide new generic icon for this creature (e.g. simple icon of tower suitable for any town)
- Solution B: Use icon from shooter creature, e.g. Lich icon for Necropolis. May be a bit confusing since same icon can be used by actual creature stack.
- Solution C: Create custom icons for every town in game. Best option visually, but most difficult one since this requires editing original H3 graphics. 

Personal opinion: go with solution A. I can make such icon myself once I have free time, but if somebody else wants to fix this instead - feel free to go ahead. Note that this needs two icons - for large and small battle queues, similar to existing creatures.

kambala-decapitator:
there's also "Arrow Tower Icons" mod in VCMI Extras, maybe it should become a part of Essentials

IvanSavenko:
Hmm... That mod contains:
- "kill" sound, for tower destruction - explosion already has sound, so not sure whether it is necessary.
- "shoot" sound - vcmi should use sound from corresponding creature instead (fixed in my branch)
- tower icons - too different art style compared to H3 in my opinion. I'd rather make monochrome icon & color it with that light-gold color commonly used in H3 UI.

So yeah, it should be in core game but probably with slightly different content.

misiokles:
![towerBigOld](https://user-images.githubusercontent.com/29593042/204387210-e7d73fe3-f888-491e-999e-bf2d70e54c13.png)
![towerSmallOld](https://user-images.githubusercontent.com/29593042/204387236-42178886-88c3-4371-a200-31eb380b91ac.png)
(alternative tower icons - ripped from Inferno's siege screen).

IvanSavenko:
![зображення](https://user-images.githubusercontent.com/1576820/204388394-bb9ea6e9-c4d3-47ec-b24f-cced04338f47.png)
Tested how all of these would look in-game:
- topmost: 3 simple monochrome icons, with slightly different background
- 2nd from top: simple monochrome icon, small battle queue
[...]

PR #1172, opened 2022-11-28
[1.2] Refactoring of BattleInterface class

IvanSavenko:
This is attempt to finally clean up & make more manageable class that grew too large over time and was never (re-)designed properly to handle every battle-related mechanics in game.
Long-term goal is to finish remaining missing features & fix remaining bugs, but to avoid this PR from growing even more, most of these changes will be done separately.
Currently this is still WIP, but feel free to comment on anything inside vcmi/client/battle directory, even if this code was present before - I want to get rid of all code that is not clear enough to make it more easy to maintain in future.
Sorry for pull request that is too large for a comfortable review, but this was easiest way to deal with 4000 lines-sized black magic grimoire called "CBattleInterface.cpp"

**Current organization of client-side battle code**
- CBattleInterface - old God Object, now mostly responsible for connecting various "Controllers", remaining functionality will likely remain here unless there will be good suggestions on how to reorganize it nicely.
- CBattleAnimations - still WIP, a lot of obscure or outdated code
- CBattleControlPanel - GUI object that handles functionality of panel at the bottom of combat screen
- CBattleXXXController - set of classes that control one specific part of combat functionality, e.g. obstacles, projectiles, castle walls, etc. Mostly consists of code that was previously part of CBattleInterface
- CCanvas - new small class that acts as abstraction on top of SDL_Surface and provides rendering target to draw on. Currently used only in battle interface, but long-term goal is to use it everywhere in the game instead of raw SDL_Surface.
- CCreatureAnimation - largely unchanged, code quality is good enough
- CBattleInterfaceClasses - largely unchanged, code quality is good enough

**Known regressions/TODO's:**
- [x] code for rendering of battlefield objects is temporary and currently has complexity of O(n*m) instead of O(n+m)
- [x] Check "Animate Dead" spell
- [x] Check "Sacrifice" spell
- [x] Check "Resurrection" spell effect on dead stack
- [x] Check animations of creature-spellcasters
- [x] During set up animation, obstacles should be rendered according to hex ID, not on top

**Upstream issues that are fixed in branch**
- [x] no hover text on buttons from combat panel
- [x] status bar on server connection screen may be broken. Could be upstream bug, will investigate
- [x] Earthquake spell should have overall sound effect + another sound on wall hit
- [x] Trying to cast Remove Obstacles with no obstacles results in "Unknown problem with this spell, no more information available"
- [x] All area-effect spell effects should play on top of the screen


kambala-decapitator:
> tower shooter has no valid icon for battle queue

#791 

> fire shield has no sound

#960 

> bloodlust effect missing
> petrification effect missing

was mentioned in #862

client/lobby/OptionsTab.cpp
```diff
@@ -468,20 +467,20 @@ OptionsTab::PlayerOptionsEntry::PlayerOptionsEntry(const PlayerSettings & S, con
...
 		whoCanPlay = HUMAN;
 
-	static const char * flags[] =
+	static std::string flags[] =
 	{
 		"AOFLGBR.DEF", "AOFLGBB.DEF", "AOFLGBY.DEF", "AOFLGBG.DEF",
 		"AOFLGBO.DEF", "AOFLGBP.DEF", "AOFLGBT.DEF", "AOFLGBS.DEF"
 	};
-	static const char * bgs[] =
+	static std::string bgs[] =
```
Nordsoft91, line 475: suggest to use std::array

misiokles:
Arrow towers icons are now part of vcmi-extras.

IvanSavenko:
@misiokles check #791
Also, vcmi-extra should provide well, *extra*. Not something needed to fix bugs. So things like tower icons should be in base game.

misiokles:
@IvanSavenko yeah, those icons were made by Witchking especially for vcmi, but sxx/alexvins never added them.

Nordsoft91:
Just started review, but first comment is to consider removing C prefix from class names. This naming approach is deprecated.

IvanSavenko:
> Just started review, but first comment is to consider removing C prefix from class names. This naming approach is deprecated.

Got it, will rename everything in battleint.
Is current style documented somewhere?
What about other letter prefixes, namely EEnum/TTypedef/IInterface? Keep letter or drop it?

Nordsoft91:
As I remember, typedefs and enums should be omitted, but nothing can say about interface. For me having IIterface is fine for pure virtual classes as there is no interface definition in C++. As for documentation, I don't think it's documented well, but there was discussion in slack some time ago. @nullkiller knows better

client/gui/CIntObject.h
```diff
@@ -217,3 +215,13 @@ class WindowBase : public CIntObject
 protected:
 	void close();
 };
+
+class IStatusBar
+{
+public:
+	virtual ~IStatusBar();
+	virtual void clear() = 0;
+	virtual void clearMatching(const std::string & Text) = 0;
```
Nordsoft91, line 224: If I look at this interface, it's completely unclear what clearMatching shall do

client/gui/CIntObject.h
```diff
@@ -217,3 +215,13 @@ class WindowBase : public CIntObject
...
 };
+
+class IStatusBar
+{
+public:
+	virtual ~IStatusBar();
+	virtual void clear() = 0;
+	virtual void clearMatching(const std::string & Text) = 0;
+	virtual void write(const std::string & Text) = 0;
+	virtual void lock(bool shouldLock) = 0;
```
Nordsoft91, line 226: what is that for? Is it mutex lock? What shouldLock means? How behavior from other methods should differ if locked? For such interfaces it's good to have comments

client/gui/CCanvas.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include "CCanvas.h"
+
+#include "SDL_Extensions.h"
+#include "Geometries.h"
+#include "CAnimation.h"
+
+#include "../Graphics.h"
+
+CCanvas::CCanvas(SDL_Surface * surface):
+	surface(surface)
```
Nordsoft91, line 20:
```suggestion
CCanvas::CCanvas(SDL_Surface & surface):
	surface(&surface)
```

Nordsoft91 (changes requested):
Reviewed approximately 50%

client/gui/CCanvas.h
```diff
@@ -0,0 +1,53 @@
...
+	~CCanvas();
+
+	/// renders image onto this canvas at specified position
+	void draw(std::shared_ptr<IImage> image, const Point & pos);
+
+	/// renders section of image bounded by sourceRect at specified position
+	void draw(std::shared_ptr<IImage> image, const Point & pos, const Rect & sourceRect);
+
+	/// renders another canvas onto this canvas
+	void draw(std::shared_ptr<CCanvas> image, const Point & pos);
```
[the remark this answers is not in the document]
IvanSavenko: IImage is always used while wrapped in shared_ptr. 
Using shared_ptr here makes more sense for consistency. 
Sure, I can switch it to reference but:
1) this means more boilerplate code in callers, since you'll need to dereference shared_ptr in every call
2) more error prone since it becomes caller responsibility to check for null = more points of failure.

I'd rather add nullptr checks inside CCanvas code, probably with asserts for debug code.

client/gui/CCanvas.cpp
```diff
@@ -0,0 +1,78 @@
...
+}
+
+CCanvas::CCanvas(const Point & size)
+{
+	surface = CSDL_Ext::newSurface(size.x, size.y);
+}
+
+CCanvas::~CCanvas()
+{
+	SDL_FreeSurface(surface);
```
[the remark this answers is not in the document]
IvanSavenko: Can't find where this is documented in SDL docs for some reason, but yes.
SDL_FreeSurface decrements refcount and, once it reaches 0 - deletes the surface.

We already use such code in SDLImage without issues.

client/gui/CCanvas.cpp
```diff
@@ -0,0 +1,78 @@
...
+	{
+	case ETextAlignment::TOPLEFT:      return graphics->fonts[font]->renderTextLinesLeft  (surface, text, colorDest, position);
+	case ETextAlignment::CENTER:       return graphics->fonts[font]->renderTextLinesCenter(surface, text, colorDest, position);
+	case ETextAlignment::BOTTOMRIGHT:  return graphics->fonts[font]->renderTextLinesRight (surface, text, colorDest, position);
+	}
+}
+
+SDL_Surface * CCanvas::getSurface()
+{
+	return surface;
```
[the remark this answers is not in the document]
IvanSavenko: Technically speaking, no point in that - this method is only for use by SDL API, which always accepts pointer form.

But in this case I can do even better - remove this method entirely since by now it should be unused

client/battle/CCreatureAnimation.h
```diff
@@ -98,7 +99,7 @@ class CCreatureAnimation : public CIntObject
 	void setType(CCreatureAnim::EAnimType type); //sets type of animation and cleares framecount
 	CCreatureAnim::EAnimType getType() const; //returns type of animation
 
-	void nextFrame(SDL_Surface * dest, bool attacker);
+	void nextFrame(std::shared_ptr<CCanvas> canvas, bool facingRight);
```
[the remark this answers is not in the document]
IvanSavenko: OK, will do same for all other usages of Canvas class - it can be created on stack or as member, no need for pointer, in any form.

client/gui/SDL_Extensions.h
```diff
@@ -155,37 +161,51 @@ typedef void (*BlitterWithRotationVal)(SDL_Surface *src,SDL_Rect srcRect, SDL_Su
...
+			uint8_t(std::min(255.f, std::round(input.r * float(mul.r) / 255 + add.r))),
+			uint8_t(std::min(255.f, std::round(input.g * float(mul.g) / 255 + add.g))),
+			uint8_t(std::min(255.f, std::round(input.b * float(mul.b) / 255 + add.b))),
+			uint8_t(std::min(255.f, std::round(input.a * float(mul.a) / 255 + add.a))),
+		};
 	}
 };
 
-class ColorShifterDeepRed : public ColorShifter
+class ColorShifterAddMulExcept : public ColorShifterAddMul
```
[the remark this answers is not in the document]
IvanSavenko: "Uncommon" depends on point of view, but OK, will review names of this classes.

client/widgets/TextControls.cpp
```diff
@@ -339,18 +339,24 @@ void CTextBox::setText(const std::string & text)
...
 }
 
-void CGStatusBar::setText(const std::string & Text)
+void CGStatusBar::write(const std::string & Text)
 {
 	if(!textLock)
 		CLabel::setText(Text);
 }
 
+void CGStatusBar::clearMatching(const std::string & Text)
```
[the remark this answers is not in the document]
IvanSavenko: 1. Ok
2. ALL usages of getText() were followed by clear(). There is no code that actually needs to retrieve text from status bar.
So while you are not wrong, I don't see problem behind method that remove pretty much copy-pasted code (which in some cases was done incorrectly).

client/gui/CIntObject.h
IvanSavenko: Strictly speaking, old code - this is part of CGStatusBar interface, but yeah, good point

IvanSavenko:
Applied most of review comments except for changes involving status bar (will do that later, probably tomorrow)

client/gui/CCanvas.h
```diff
@@ -0,0 +1,53 @@
...
+enum EFonts : int;
+
+/// Class that represents surface for drawing on
+class CCanvas
+{
+	SDL_Surface * surface;
+public:
+
+	/// constructs canvas using existing surface. Caller maintains ownership on the surface
+	CCanvas(SDL_Surface * surface);
```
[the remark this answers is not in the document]
IvanSavenko: Probably better to keep it as it - SDL_Surface is part of C API of SDL, which uses pointers only
It is always used in pointer form (for example, right now codebase has 0 occurrences of `SDL_Surface&` or `SDL_Surface &`).

client/battle/BattleRenderer.h
```diff
@@ -0,0 +1,54 @@
...
+	STACKS        = 1, // after corpses, obstacles
+	BATTLEMENTS   = 2, // after stacks
+	STACK_AMOUNTS = 2, // after stacks, obstacles, corpses
+	EFFECTS       = 3, // after obstacles, battlements
+};
+
+class BattleRenderer
+{
+public:
+	using RendererPtr = Canvas &;
```
Nordsoft91, line 32:
```suggestion
	using RendererRef = Canvas &;
```

client/battle/BattleRenderer.h
```diff
@@ -0,0 +1,54 @@
...
+	using RenderFunctor = std::function<void(RendererPtr)>;
```
Nordsoft91, line 33: std::function is the slowest way (and it can really impact performance). Don't know yet usage if RenderFunction, but if it's something in the loop, better to consider alternatives:
 - template 
 - function pointer

client/battle/BattleRenderer.h
```diff
@@ -0,0 +1,54 @@
...
+
+private:
+	BattleInterface * owner;
```
Nordsoft91, line 36: Same - cound be reference

client/battle/BattleRenderer.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+BattleRenderer::BattleRenderer(BattleInterface * owner):
+	owner(owner)
+{
+}
+
+void BattleRenderer::insert(EBattleFieldLayer layer, BattleHex tile, BattleRenderer::RenderFunctor functor)
+{
+	objects.push_back({ functor, layer, tile });
```
Nordsoft91, line 69: emplace_back

client/battle/BattleRenderer.cpp
```diff
@@ -0,0 +1,77 @@
...
+			return -1;
+
+		if ( object.tile == BattleHex::HEX_AFTER_ALL )
+			return GameConstants::BFIELD_HEIGHT;
+
+		if ( object.tile == BattleHex::INVALID )
+			return GameConstants::BFIELD_HEIGHT;
+
+		assert(0);
+		return GameConstants::BFIELD_HEIGHT;
```
Nordsoft91, line 46: Single result at the end, so could be
```suggestion
		return GameConstants::BFIELD_HEIGHT;
```

client/battle/BattleProjectileController.cpp
```diff
@@ -0,0 +1,376 @@
...
+	if(creature->idNumber == CreatureID::ARROW_TOWERS)
+		creature = owner->siegeController->getTurretCreature();
+
+	if(creature->animation.missleFrameAngles.empty())
+	{
+		logAnim->error("Mod error: Creature '%s' on the Archer's tower is not a shooter. Mod should be fixed. Trying to use archer's data instead...", creature->nameSing);
+		creature = CGI->creh->objects[CreatureID::ARCHER];
+	}
+
+	return creature;
```
Nordsoft91, line 166: Seems like never returns nullptr. So why not to make it explicitly by returning const reference?

client/battle/BattleProjectileController.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+bool BattleProjectileController::stackUsesRayProjectile(const CStack * stack)
+{
+	return !getShooter(stack)->animation.projectileRay.empty();
+}
+
+bool BattleProjectileController::stackUsesMissileProjectile(const CStack * stack)
+{
+	return !getShooter(stack)->animation.projectileImageName.empty();
+}
```
Nordsoft91, line 177: Both should be const methods

client/battle/BattleProjectileController.cpp
```diff
@@ -0,0 +1,376 @@
...
+		projectile->animation     = createProjectileImage(animToDisplay);
+		projectile->frameProgress = 0;
+		projectile->frameNum      = 0;
+		projectile->reverse       = from.x > dest.x;
+		projectile->from          = from;
+		projectile->dest          = dest;
+		projectile->shooterID     = shooter ? shooter->ID : -1;
+		projectile->step          = 0;
+		projectile->steps         = computeProjectileFlightTime(from, dest, AnimationControls::getSpellEffectSpeed());
+		projectile->playing       = false;
```
Nordsoft91, line 372: Repeated twice. Maybe some method? Will have to avoid future mistakes if someone modifies one part D forget about another

client/battle/BattleControlPanel.h
```diff
@@ -0,0 +1,76 @@
...
+VCMI_LIB_NAMESPACE_BEGIN
+class CStack;
+
+VCMI_LIB_NAMESPACE_END
+
+class CButton;
+class BattleInterface;
+class BattleConsole;
+
+/// GUI object that handles functionality of panel at the bottom of combat screen
```
Nordsoft91, line 23: I would call it ribbon

client/battle/BattleActionsController.cpp
```diff
@@ -0,0 +1,775 @@
...
+		else
+			illegalAction = PossiblePlayerBattleAction::INVALID; //we should never be here
+	}
+
+	bool isCastingPossible = false;
+	bool secondaryTarget = false;
+
+	if (currentAction > PossiblePlayerBattleAction::INVALID)
+	{
+		switch (currentAction) //display console message, realize selected action
```
Nordsoft91, line 427: Should be possible to make this logic configurable in the future

client/battle/BattleActionsController.cpp
```diff
@@ -0,0 +1,775 @@
...
+	else
+	{
+		possibleActions.clear();
+		possibleActions.push_back (spellSelMode); //only this one action can be performed at the moment
+		GH.fakeMouseMove();//update cursor
+	}
+}
+
+
+void BattleActionsController::handleHex(BattleHex myNumber, int eventType)
```
Nordsoft91, line 234: Too much stuff happened in this function

Nordsoft91 (approved):
I see that most of changes are just moving code.
Thinking about those new classes, it's certainly better but still could be improved. There is a chance to adopt this according to MVC pattern. BattleInterface will be controller in this case, and other (but not all) controllers are actually not controllers but views and sometimes models. All comments I made are optional

client/battle/BattleAnimationClasses.h
```diff
@@ -0,0 +1,350 @@
...
+class CWaitingAnimation : public CBattleAnimation
+{
+protected:
+	CWaitingAnimation(BattleInterface * owner_);
+public:
+	void nextFrame() override;
+};
+
+/// Class that waits till projectile of certain shooter hits a target
+class CWaitingProjectileAnimation : public CWaitingAnimation
```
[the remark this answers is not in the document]
IvanSavenko: I do want to, but later, in 2nd PR. 
I'm currently in process of making better animation ordering logic and fixing all known animation ordering bugs, which involves major changes into content of these files.
Sure, I can push those changes here, but I'd rather avoid making any new changes (apart from fixing stuff)

client/battle/BattleStacksController.cpp
```diff
@@ -0,0 +1,596 @@
...
+
+#include "../../CCallback.h"
+#include "../../lib/battle/BattleHex.h"
+#include "../../lib/CGameState.h"
+#include "../../lib/CStack.h"
+#include "../../lib/CondSh.h"
+
+static void onAnimationFinished(const CStack *stack, std::weak_ptr<CreatureAnimation> anim)
+{
+	std::shared_ptr<CreatureAnimation> animation = anim.lock();
```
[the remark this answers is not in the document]
IvanSavenko: Old code that I haven't noticed. Investigating...

client/battle/BattleStacksController.cpp
```diff
@@ -0,0 +1,596 @@
...
+	if(!animation)
+		return;
+
+	if (animation->isIdle())
+	{
+		const CCreature *creature = stack->getCreature();
```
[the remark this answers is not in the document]
IvanSavenko: Stack or creature? In either case - no, it shouldn't be.

client/battle/BattleActionsController.cpp
IvanSavenko: Noted.
Completely agree with you, but for now will keep as it is. Will fix in future PR's

client/battle/BattleRenderer.h
IvanSavenko: I would rather use interface + virtual function here instead, but that would need some rework of classes that use this RenderFunction.
Overhead is obviously exists, but compared to execution time of functors it should be negligible

client/battle/BattleSiegeController.h
```diff
@@ -0,0 +1,110 @@
...
+
+struct Point;
+class Canvas;
+class BattleInterface;
+class BattleRenderer;
+class IImage;
+
+namespace EWallVisual
+{
+	enum EWallVisual
```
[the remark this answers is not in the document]
IvanSavenko: Noted for future.

IvanSavenko:
Pushed final commits according to Nordsoft review.
Some review comments are still valid - I agree with vast majority of them, but would like to postpone them to next PR.

If there are no objections from @Nordsoft91 or @nullkiller then I'm merging this PR into develop branch later today (assuming that all CI checks will pass)

Shortly after that I'm going to open new PR, this time - one that actually has user-visible changes - bugfixes & support for missing functionality.

>I see that most of changes are just moving code.
Thinking about those new classes, it's certainly better but still could be improved. There is a chance to adopt this according to MVC pattern. BattleInterface will be controller in this case, and other (but not all) controllers are actually not controllers but views and sometimes models. All comments I made are optional

Yes, my initial goal for this PR was not to fix bugs (although I did it whenever I saw clearly wrong code) but rather prepare foundation for future changes. So this PR is mostly moving & rewriting code without functional changes.

MVC - I'll have to think about it, but not sure. MVC means rather tight coupling, while I attempted to do the opposite - create several classes that have weak dependencies on each other.
Now, I do agree that this could be improved - there are things that I haven't considered or wasn't aware of when I started PR, and I am planing on improving that. But I think domain-based encapsulation makes for sense here than MVC.

PR #1175, opened 2022-11-29
Refactoring of artifacts code

SoundSSGood:


IvanSavenko:
FYI: to fix build you need to create specialization of std::hash in std namespace, e.g.:
```
namespace std 
{
	template<> 
	struct hash<ArtifactID>
	{
	size_t operator()(const ArtifactID & aid) const
	{
		return hash<int>{}(aid.num);
	}
 }
```

lib/CArtHandler.h
```diff
@@ -386,7 +386,28 @@ namespace ArtifactUtils
...
+		ArtifactPosition::TORSO,
+		ArtifactPosition::RIGHT_RING,
+		ArtifactPosition::LEFT_RING,
+		ArtifactPosition::FEET,
+		ArtifactPosition::MISC1,
+		ArtifactPosition::MISC2,
+		ArtifactPosition::MISC3,
+		ArtifactPosition::MISC4,
+		ArtifactPosition::MISC5,
+	};
```
IvanSavenko, line 410: Not sure what's better - this const array or function that return reference to const vector (with all details such as size & elements hidden inside)
But this is minor detail, looks OK othervice.

IvanSavenko (approved):

lib/CArtHandler.cpp
```diff
@@ -1369,14 +1370,14 @@ void CArtifactSet::setNewArtSlot(ArtifactPosition slot, CArtifactInstance *art,
 
 void CArtifactSet::eraseArtSlot(ArtifactPosition slot)
 {
-	if(slot < GameConstants::BACKPACK_START)
+	if(ArtifactUtils::isSlotBackpack(slot))
 	{
-		artifactsWorn.erase(slot);
+		slot = ArtifactPosition(slot - GameConstants::BACKPACK_START);
+		artifactsInBackpack.erase(artifactsInBackpack.begin() + slot);
```
Nordsoft91, line 1376: I suggest to add assert(artifactsInBackpack.begin() + slot < artifactsInBackpack.end());

lib/CArtHandler.h
Nordsoft91: I personally would prefer to have such a hardcoded arrays as it for now declared in header without hiding them inside function. But it's just my preference

lib/mapObjects/CQuest.h
```diff
@@ -15,13 +15,25 @@
...
+{
+	// Used in std::unordered_map
+	template<> struct hash<VCMI_LIB_WRAP_NAMESPACE(ArtifactID)>
+	{
+		size_t operator()(const VCMI_LIB_WRAP_NAMESPACE(ArtifactID) & aid) const
+		{
+			return hash<int>{}(aid.num);
+		}
+	};
+}
```
Nordsoft91, line 28: Extending std is usually not what you are supposed to do.
Instead of having hash function in std namespace, make separate functor and pass it to unordered_map as template argument (for me it seems like improvement desired)

Nordsoft91 (changes requested):
Placing request changes because disagree with overloading std::hash instead of passing hash functor to std::unordered_map

lib/CArtHandler.h
nullkiller: changing headers causes recompilation so I prefer keeping implementation details in cpp units as more as it is possible

nullkiller: Also I am not sure if const variables should be uppercased or not

nullkiller: Also this data may be moved to ArtifactUtils. near isBackpackSlot and as far as I understoot it is heroSlots - unmovable slots, right? So these are hero movable slots and such data may be usable for some other code.

nullkiller: Or even put it into ArtifactPosition. ArtifactPosition::isBackpackSlot(pos), ArtifactPosition::movableHeroSlots(), ArtifactPosition::unmovableHeroSlots()

lib/mapObjects/CQuest.h
IvanSavenko: Actually providing template specializations in std namespace is one of explicitly allowed exceptions in the standard.
So this code is completely legal and won't cause any issues with any standard-compliant c++ compiler.

Extending std namespace vs providing functor to unordered_map is more of a code style question.
And current approach allows declaring unordered_map's seamlessly without any additional code.

lib/CArtHandler.h
IvanSavenko: > as far as I understoot it is heroSlots - unmovable slots, right?

This list also excludes all war machines, which are neither "unmovableSlots" nor "constituentWornSlots".
Kind of makes sense - while war machines may be *theoretically* used as constituent art for combined, this is not something that make sense right now and likely to cause issues.

IvanSavenko: in that case - perhaps move it into .cpp? Something like this:
```
const std::vector<ArtifactPosition::EArtifactPosition> & constituentWornSlots()
{
    static const std::vector<ArtifactPosition::EArtifactPosition>slots = { <elements> }

    return slots;
}
```
This way all internal details like elements & their amount is hidden from the interface & function overhead is minimal since it just returns existing vector.

Would also help with reducing header size - less noise if you're looking through API.

nullkiller: Ok, makes sense.

lib/mapObjects/CQuest.h
Nordsoft91, line 28:
```suggestion
// Used in std::unordered_map
template<> struct std::hash<VCMI_LIB_WRAP_NAMESPACE(ArtifactID)>
{
	std::size_t operator()(const VCMI_LIB_WRAP_NAMESPACE(ArtifactID) & aid) const
	{
		return std::hash<int>{}(aid.num);
	}
};
```

Nordsoft91: > Opening namespace std can accidentally introduce undefined behavior, and is not necessary for specializing class templates.

From: https://en.cppreference.com/w/cpp/language/extending_std
I've put a suggestion

lib/CArtHandler.h
SoundSSGood: > Kind of makes sense - while war machines may be _theoretically_ used as constituent art for combined, this is not something that make sense right now and likely to cause issues.

It was assumed that the list could be expanded in the future

lib/mapObjects/CQuest.h
SoundSSGood: I already tried this case. It causes build fail

SoundSSGood: made hash as structure inside ArtifactID

Nordsoft91 (approved):

PR #1176, opened 2022-11-29
Tower shooter icons fix

### LINKED ISSUE #791: wrong icons of arrow towers

IvanSavenko:
Fixes #791 as described in comments.

- added new icons for tower shooters, to be used as fallback in scenario when user plays without vcmi-extra mod
- added functionality to define tower icon per town, allowing icons that represent this town


client/battle/CBattleInterfaceClasses.cpp
```diff
@@ -763,7 +763,13 @@ void CStackQueue::update()
 		stackBoxes[boxIndex]->setUnit(nullptr);
 }
 
-CStackQueue::StackBox::StackBox(CStackQueue * owner)
+int32_t CStackQueue::getSiegeShooterIconID()
+{
+		return owner->siegeH->town->town->faction->index;
```
Nordsoft91, line 768:
```suggestion
	return owner->siegeH->town->town->faction->index;
```

mapeditor/graphics.cpp
```diff
@@ -310,11 +310,13 @@ std::shared_ptr<Animation> Graphics::getAnimation(const std::shared_ptr<const Ob
 	return ret;
 }
 
-void Graphics::addImageListEntry(size_t index, const std::string & listName, const std::string & imageName)
+void Graphics::addImageListEntry(size_t index, size_t group, const std::string & listName, const std::string & imageName)
 {
 	if (!imageName.empty())
 	{
 		JsonNode entry;
+		if ( group != 0)
```
Nordsoft91, line 318:
```suggestion
		if(group != 0)
```

Nordsoft91 (commented):
Where did you get those new assets? Shouldn't they be contained in vcmi-extras instead?

IvanSavenko:
> Where did you get those new assets? Shouldn't they be contained in vcmi-extras instead?
Made by me in gimp. 
Background for larger image is based on public-domain texture that we also have in vcmi-assets repository.

Nordsoft91:
Please bump version of vcmi essential mod

Nordsoft91 (approved):

nullkiller:
Do we already have new icons in extras mod for regular towns? How can I see them?

IvanSavenko:
There is "Arrow Tower Icons" submod in vcmi extra mod.
After this pull request the submod needs to be updated in order to take use of newly added possibility to define icons per-town.
Without an update the submod will be broken. It won't to any errors/crashes but it won't provide any functionality either.

I assume that either @misiokles or somebody from modders will have to make alternative, per-town icons and once those are done - submod will be updated.

misiokles:
Ok, I'll start to look into and prepare first version os such submod in next week.

Issue #1124, opened 2022-11-11
Support vcmi files on ios


Nordsoft91:
Provide raw access to vcmi data from ios device, so files and folders can be visible in "Files" application
![image](https://user-images.githubusercontent.com/9308612/201445558-fae127b4-bea6-4c0a-ae89-c14f8da34d32.jpeg)

PR #1179, opened 2022-11-30
[1124] Implement browsing data

Nordsoft91:
Implements https://github.com/vcmi/vcmi/issues/1124


kambala-decapitator (changes requested):
please keep keys in alphabetical order

tested on iPad iOS 15: works!

PR #1181, opened 2022-12-02
[beta] Version bump of Linux packaging to 1.1.0

IvanSavenko:


launcher/eu.vcmi.VCMI.metainfo.xml
```diff
@@ -38,6 +38,7 @@
 	<url type="bugtracker">https://github.com/vcmi/vcmi/issues</url>
 	<url type="faq">https://vcmi.eu/faq/</url>
 	<releases>
+		<release version="1.1.0" date="2022-12-02" />
```
nullkiller, line 41: How important this date? Previous one is quite precise

IvanSavenko: So far I have yet to see any scenario in which date is used, either this one or the one set in debian/changelog.
However version number itself can be used for version control/package names.
So I'd rather have wrong date but correct version number.

It makes sense to update release date once it is out, but right now having correct version is more important.

PR #1185, opened 2022-12-02
Fix editor crashes [1.1]

Nordsoft91:


mapeditor/windownewmap.cpp
```diff
@@ -37,8 +37,10 @@ WindowNewMap::WindowNewMap(QWidget *parent) :
 	show();
 
 	//setup initial parameters
-	mapGenOptions.setWidth(ui->widthTxt->text().toInt());
-	mapGenOptions.setHeight(ui->heightTxt->text().toInt());
+	int width = ui->widthTxt->text().toInt();
+	int height = ui->heightTxt->text().toInt();
+	mapGenOptions.setWidth(width ? width : 1);
```
nullkiller, line 42: is it kind of default value? Is not 1 too small?

Nordsoft91: There is an assertion that width and size must be bigger 0.
1 is small but I can imagine map something like 1x200 :)

lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -308,8 +308,9 @@ TObjectTypeHandler CObjectClassesHandler::getHandlerFor(si32 type, si32 subtype)
 		if (objects.at(type)->subObjects.count(subtype))
 			return objects.at(type)->subObjects.at(subtype);
 	}
-	logGlobal->error("Failed to find object of type %d:%d", type, subtype);
-	throw std::runtime_error("Object type handler not found");
+	std::string errorString = "Failed to find object of type " + std::to_string(type) + "::" + std::to_string(subtype);
```
nullkiller, line 311: Why do we need std::to_string here and not below

nullkiller (approved):

PR #1194, opened 2022-12-04
Fix hero-related problems in editor [1.1]

Nordsoft91:


mapeditor/mapcontroller.cpp
```diff
@@ -128,10 +128,15 @@ void MapController::repairMap()
...
+			{
 				nih->typeName = "hero";
+				nih->subTypeName = type->heroClass->identifier;
+			}
 			if(obj->ID == Obj::PRISON)
+			{
 				nih->typeName = "prison";
-			nih->subTypeName = type->heroClass->identifier;
+				nih->subTypeName = "prison";
+			}
```
Nordsoft91, line 139: without this fix maps from editor containing heroes/prisons are not playable

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1434,7 +1434,7 @@ void CGHeroInstance::serializeCommonOptions(JsonSerializeFormat & handler)
 	handler.serializeString("biography", biography);
 	handler.serializeInt("experience", exp, 0);
 
-	if (!handler.saving)
+	if(!handler.saving && exp != 0xffffffff) //do not gain levels if experience is not initialized
```
Nordsoft91, line 1437: value is from CGHeroInstance constructor

PR #1203, opened 2022-12-05
Resource Extractor produces files with only zeroes.

krs0:
This bug was introduces with the modification of:
`void CArchiveLoader::extractToFolder(const std::string& outputSubFolder, const std::string& mountPoint, ArchiveEntry entry)`
to internally make use of:
`void CArchiveLoader::extractToFolder(const std::string & outputSubFolder, CInputStream & fileStream, ArchiveEntry entry)`

### DIFF
--- a/lib/filesystem/CArchiveLoader.cpp
+++ b/lib/filesystem/CArchiveLoader.cpp
@@ -231,7 +231,15 @@ void CArchiveLoader::extractToFolder(const std::string & outputSubFolder, const 
 {
 	std::unique_ptr<CInputStream> inputStream = load(ResourceID(mountPoint + entry.name));
 
-	extractToFolder(outputSubFolder, *inputStream, entry);
+	std::vector<ui8> data(entry.fullSize);
+	inputStream->read(data.data(), entry.fullSize);
+
+	bfs::path extractedFilePath = createExtractedFilePath(outputSubFolder, entry.name);
+
+	// writeToOutputFile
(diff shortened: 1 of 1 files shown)


lib/filesystem/CArchiveLoader.cpp
```diff
@@ -231,7 +231,15 @@ void CArchiveLoader::extractToFolder(const std::string & outputSubFolder, const
...
-	extractToFolder(outputSubFolder, *inputStream, entry);
+	std::vector<ui8> data(entry.fullSize);
+	inputStream->read(data.data(), entry.fullSize);
+
+	bfs::path extractedFilePath = createExtractedFilePath(outputSubFolder, entry.name);
+
+	// writeToOutputFile
+	std::ofstream out(extractedFilePath.string(), std::ofstream::binary);
+	out.exceptions(std::ifstream::failbit | std::ifstream::badbit);
+	out.write((char*)data.data(), entry.fullSize);
```
Nordsoft91, line 242: It's a copy-paste of function above, which is very desirable to avoid. You can fix it easier
```suggestion
	entry.offset = 0;
	extractToFolder(outputSubFolder, *inputStream, entry);
```

Nordsoft91 (changes requested):

lib/filesystem/CArchiveLoader.cpp
krs0: Yup entry.offset = 0 was the culprit. Thanks for the suggestion.

Nordsoft91 (approved):

PR #1204, opened 2022-12-05
[1.2] Single process app support for platforms other than ios

IvanSavenko:
Added support for any platform to use "single-process app" mode that was previously only possible when compiling for ios.
Currently this is mostly to run vcmi in single process for ease of debugging, and not intended for release builds. Don't see anything that would prevent such builds though.

Changes:
- BUILD_SINGLE_APP is now called ENABLE_SINGLE_APP_BUILD for consistency with other CMake options
- ENABLE_SINGLE_APP_BUILD should now work on any platform.
- library used by client is now always called "vcmi", e.g. libvcmi.so, even in singe-app mode. 
- map editor can now be built when BUILD_SINGLE_APP
- added COPY_CONFIG_ON_BUILD CMake option to stop vcmi from copying config & mods directories on library build

Known issues:
- msvc fails during linking in single-app mode. Since I don't have Windows dev environment set up I can't investigate this properly.


IvanSavenko:
Marking as ready for review.
Visual Studio build of single-app fails for some reason, but all other platforms including MXE seems to be ok.
If somebody with MSVC setup can investigate linking issues that would be great.

Otherwise if everyone is OK with changes so far I think PR can be merged

mapeditor/jsonutils.h
```diff
@@ -12,6 +12,8 @@
 #include <QVariant>
 #include "../lib/JsonNode.h"
 
+VCMI_LIB_NAMESPACE_BEGIN
```
Nordsoft91, line 15: Why should it be part of lib namespace?

AI/EmptyAI/StdInc.h
```diff
@@ -4,4 +4,6 @@
 
 // This header should be treated as a pre compiled header file(PCH) in the compiler building settings.
 
-// Here you can add specific libraries and macros which are specific to this project.
\ No newline at end of file
+// Here you can add specific libraries and macros which are specific to this project.
+
+VCMI_LIB_USING_NAMESPACE
```
Nordsoft91, line 9: Why it's needed in stdinc? Should be same for other ai libs?

IvanSavenko: It is actually there in all other ai libs, even in develop branch.
Apparently EmptyAI was not tested/included originally, so it ended up without this change

mapeditor/jsonutils.h
IvanSavenko: JsonUtils namespace is already part of vcmi library (lib/JsonUtils,h).
Including any of its extensions into VCMI_LIB namespace is easier. 
Otherwise every usage of JsonUtils namespace in editor will have to specify which namespace to access - global ::JsonUtils or VCMI_LIB::JsonUtils from library.

Another option would be to rename editor version of this namespace to something else, e.g. JsonEditor/JsonQt/etc.

CMakeLists.txt
```diff
@@ -162,8 +168,9 @@ set(CMAKE_XCODE_ATTRIBUTE_MARKETING_VERSION ${APP_SHORT_VERSION})
 set(CMAKE_XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH NO)
 set(CMAKE_XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] YES)
 
-if(BUILD_SINGLE_APP)
-	add_compile_definitions(SINGLE_PROCESS_APP=1)
+if(ENABLE_SINGLE_APP_BUILD)
+	add_definitions(-DSINGLE_PROCESS_APP=1)
+	#add_compile_definitions(SINGLE_PROCESS_APP=1) # requires cmake 3.12
```
nullkiller, line 173: Do we need this line anymore?

IvanSavenko: Which one?
`SINGLE_PROCESS_APP=1` define is used in c++ code
commented-out `add_compile_definitions` is likely to be needed in future - looks like CMake is replacing `add_definitions` with a set of add_X commands, so `add_definitions`  is likely be removed/deprecated at some point

nullkiller: Why can not we use add_compile_definition right now?

nullkiller: I think we already depend on cmake 3.16 because we use its builtin precompiled headers instead of cotire

nullkiller: But if it is still compilable with cmake 3.10 I suggest to remove commented out code.

IvanSavenko: Precompiler headers check for CMake version, so they work even with older cmake.
I guess I remove commented-out code from here, at least for now.

A separate question is ancient MXE build which uses (among other things) CMake 3.10. But this is definitely out of scope of this PR, will write this on Slack

Shatur:
When `ENABLE_SINGLE_APP_BUILD` is enabled I have to put `config` and `Mods` to `vcmi` folder. They can't be located alongside `vcmiclient`. The behavior can be reproduced with and without this PR. Is this expected?
Asking here because it may be a leftover from times when this feature worked for some platforms.

kambala-decapitator:
@Shatur it has always worked like that. Before this PR "single process build" was supported only on iOS.

what you're looking for is "monolithic install": `ENABLE_MONOLITHIC_INSTALL` cmake option.

nullkiller (approved):

kambala-decapitator:
> The issue is when ENABLE_SINGLE_APP_BUILD is enabled the files alongside the executable aren't respected.

check `VCMIDirs.cpp` - it controls expected location of files

IvanSavenko:
There is also so-called "developmentMode" (see VCMIDirs file). It exists only on unix-like systems that use XDG directory specification.
If "development mode" is detected (see VCMIDirs for what triggers it) vcmi will read files from same directory as executable instead of using system-wide XDG directories.

Personally, I find soft-linking more reliable - for development mode CMake copies config & mods directories after building library. But if you've made changes to config's and no changes to library you'll end up with outdated files in build dir.

For me having all H3 files + soft-linked vcmi config in ~/.local/share/vcmi is more convenient setup, perhaps you can use it as well

IvanSavenko:
It was changed in this branch since "vcmiserver" no longer exists and can't be used for queries.

That being said, if you're using this mode for debugging then this is fine. But for "release" mode I'd avoid using it - this is mode specifically for development and something that can be changed at any point.

For more stable approach I suggest using either XDG_DATA_HOME environment variable or its default value of ~/.local/share for vcmi data

Shatur (approved):
Got it, it does makes sense, thanks for the explanations!
Sorry for offtopic.
Can confirm that the PR works as intended on my PC.

PR #1205, opened 2022-12-05
PCX file conversion in Resource Extractor throws exception on IOS (boost::locale stuff)

krs0:
Replaces boostlocale:to_lower() with boost::algorithm::to_lower_copy()


Nordsoft91:
not iOS, macOS

Nordsoft91:
Tested locally

PR #1208, opened 2022-12-06
Absolute threshold for retreat decision

Nordsoft91:


dydzio0614:
I think this is good idea at least for 1.1 patch until retreat logic gets tweaked or improved

PR #1213, opened 2022-12-06
Fix crash on casting clone on modded creature

### LINKED ISSUE #1211: HOTA nymph cloning issue
Looks like Battle AI when evaluating possibility to clone hota.cove:nymph crashes. Looks like deserializing battle::UnitInfo fails to lookup this id which causes the crash.
It is not clear if human clonning also has such issue.

IvanSavenko:
Fixes #1211 - json deserializer would incorrectly resolve identifiers using "core" mod that should only have access to H3 data, not mods

### DIFF
--- a/lib/GameConstants.cpp
+++ b/lib/GameConstants.cpp
@@ -86,7 +86,7 @@ const Artifact * ArtifactID::toArtifact(const ArtifactService * service) const
 
 si32 ArtifactID::decode(const std::string & identifier)
 {
-	auto rawId = VLC->modh->identifiers.getIdentifier("core", "artifact", identifier);
+	auto rawId = VLC->modh->identifiers.getIdentifier("map", "artifact", identifier);
 	if(rawId)
 		return rawId.get();
 	else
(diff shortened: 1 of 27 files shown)


lib/GameConstants.cpp
```diff
@@ -86,7 +86,7 @@ const Artifact * ArtifactID::toArtifact(const ArtifactService * service) const
 
 si32 ArtifactID::decode(const std::string & identifier)
 {
-	auto rawId = VLC->modh->identifiers.getIdentifier("core", "artifact", identifier);
+	auto rawId = VLC->modh->identifiers.getIdentifier("map", "artifact", identifier);
```
nullkiller, line 89: Looks like this needs a constant. SpecialScopes::MAP or something

nullkiller: Also core scope might be put there as well

IvanSavenko: Makes sense. Should I do that as part of this PR?

nullkiller: I mean I do not expect it destabilize beta so we can do it now,

IvanSavenko: Okay. In that case I'll add
- scope for core objects
- scope for in-game requests (e.g. requests fixed in this PR)
- scope for requests from map loader (for now - aliased to in-game scope, later - it should take use of map dependencies)

PR #1214, opened 2022-12-06
Library of Enlightenment fix

IvanSavenko:
Do not offer choice between multiple identical option if hero is qualified to receive multiple of them due to diplomacy skill


Nordsoft91:
Confirm fix

PR #1222, opened 2022-12-10
Nkai pathfinder fix

nullkiller:


AI/BattleAI/BattleExchangeVariant.cpp
```diff
@@ -204,10 +204,14 @@ MoveTarget BattleExchangeEvaluator::findMoveTowardsUnreachable(const battle::Uni
 	if(targets.unreachableEnemies.empty())
 		return result;
 
+	auto speed = activeStack->Speed();
```
nullkiller, line 207: Camera obscura is a warmachine and has speed of 0, it can not be moved.

AI/Nullkiller/Pathfinding/AINodeStorage.cpp
```diff
@@ -823,13 +811,6 @@ ExchangeCandidate HeroChainCalculationTask::calculateExchange(
 	return candidate;
 }
 
-const CGHeroInstance * AINodeStorage::getHero(const CGPathNode * node) const
-{
-	auto aiNode = getAINode(node);
-
-	return aiNode->actor->hero;
```
nullkiller, line 830: Moved them to header to allow inline optimization. For some reason aiNode->actor sometimes appears to be null  while in crashdump memory it is perfectly valid. I can not understand what is going on but inlining seems to help somehow.

Nordsoft91: so you can add inline keyword

IvanSavenko: surprisingly enough, in c++ "inline" does not means that function should be inlined by compiler.
"inline" effectively means that this function is defined in header and as result - can appear in multiple compilation units.
And methods defined in place are implicitly inline'd. So no need to add inline keyword here.

Nordsoft91: Yes, it's recommendation for compiler not obligation, but it would help developers to understand intention of having those methods in header

IvanSavenko: Thing is - it is not recommendation. I remember having linker error because of incorrectly placed inline.
Won't matter in this particular case though.

PR #1225, opened 2022-12-11
[1.2] Interface builder for random map tab

Nordsoft91:
General Idea is to make all interfaces configurable.

Here I configured interface for random map tab. Next step will be implementation of extended functionality for RMG (template selection, giant maps) and creating special mod for it

Feature is ready for review and testing. It's available from latest **vcmi-extras** (extendedRMG submod)


client/lobby/RandomMapTab.h
```diff
@@ -24,7 +25,7 @@ class CToggleButton;
 class CLabel;
 class CLabelGroup;
 
-class RandomMapTab : public CIntObject
+class RandomMapTab : public InterfaceBuilder
```
IvanSavenko, line 28: It is a bit weird to have GUI element inherit from class named InterfaceBuilder.
Personally if I see something named "Builder" I would presume that this is factory-like class that creates objects one way or another.
Proposition:
a) rename "InterfaceBuilder" to... let's say "BuildableWindow" or "BuildableIntObject" (or any other suitable name)
b) if this makes sense, split InterfaceBuilder class into "builder" that processes config and "object" that is created in "build" process.

client/gui/InterfaceBuilder.h
```diff
@@ -0,0 +1,39 @@
...
+	InterfaceBuilder();
+	InterfaceBuilder(const JsonNode & config);
+
+protected:
+	//must be called after adding callbacks
+	void init(const JsonNode & config);
+	
+	void addCallback(const std::string & callbackName, std::function<void(int)> callback);
+	
+	const std::shared_ptr<CIntObject> widget(const std::string &) const;
```
IvanSavenko, line 29: In almost all usages of this method you use dynamic_pointer_cast.
Perhaps it makes sense to add another method, something like:
```
template<typename T>
const std::shared_ptr<T> widgetAs(const std::string & name) const
{
   return dynamic_pointer_cast<T>(widget(name));
}
```
And use it where applicable?

config/windows.json
```diff
@@ -0,0 +1,513 @@
+{
+	"randomMapTab":
+	{
+		"items":
+		[
```
IvanSavenko, line 5: Since widget names must be non-empty, unique string perhaps replace "items" array with struct/object, with widget name acting as key?
```
"background : { "type" : "...", "image" : "..." }
```

client/gui/InterfaceBuilder.cpp
```diff
@@ -0,0 +1,208 @@
...
+	auto type = config["type"].String();
+	
+	int x = 0, y = 0;
+	if(!config["position"].isNull())
+	{
+		x = config["position"]["x"].Integer();
+		y = config["position"]["y"].Integer();
+	}
+	
+	std::string image, text = buildText(config["text"]);
```
IvanSavenko, line 92: Not sure what's current coding style is, but I'd rather avoid declaring multiple variables in a single line.
Especially when you're initializing only one of them

client/lobby/RandomMapTab.cpp
```diff
@@ -27,137 +27,86 @@
...
-	labelMapSize = std::make_shared<CLabel>(104, 97, FONT_SMALL, EAlignment::CENTER, Colors::WHITE, CGI->generaltexth->allTexts[752]);
-	groupMapSize = std::make_shared<CToggleGroup>(0);
-	groupMapSize->pos.y += 81;
-	groupMapSize->pos.x += 158;
-	const std::vector<std::string> mapSizeBtns = {"RANSIZS", "RANSIZM", "RANSIZL", "RANSIZX"};
-	addButtonsToGroup(groupMapSize.get(), mapSizeBtns, 0, 3, 47, 198);
-	groupMapSize->setSelected(1);
-	groupMapSize->addCallback([&](int btnId)
+	
+	const JsonNode config(ResourceID("config/windows.json"));
```
IvanSavenko, line 38: Using one config file for all windows will quickly become unmanageable. 
Config just for single tab of this single window is already 500 lines in size.
Proposition: move config for this tab into "config/windows/pregame.json"
I'd even consider "config/windows/pregame/randomMapTab.json", but this can be done later if needed

client/lobby/RandomMapTab.cpp
```diff
@@ -27,137 +27,86 @@
...
-
-	labelGroupForOptions->add(68, 465, CGI->generaltexth->allTexts[758]);
-	groupMonsterStrength = std::make_shared<CToggleGroup>(0);
-	groupMonsterStrength->pos.y += 485;
-	groupMonsterStrength->pos.x += BTNS_GROUP_LEFT_MARGIN;
-	const std::vector<std::string> monsterStrengthBtns = {"RANWEAK", "RANNORM", "RANSTRG"};
-	addButtonsWithRandToGroup(groupMonsterStrength.get(), monsterStrengthBtns, 2, 4, WIDE_BTN_WIDTH, 248, 251, EMonsterStrength::RANDOM, false);
-	groupMonsterStrength->addCallback([&](int btnId)
+	
+	addCallback("setMonsterStrenght", [&](int btnId)
```
IvanSavenko, line 99:
```suggestion
	addCallback("setMonsterStrength", [&](int btnId)
```

client/lobby/RandomMapTab.cpp
```diff
@@ -216,39 +165,22 @@ void RandomMapTab::updateMapInfoByHost()
...
+		w->setSelected(opts->getPlayerCount());
+	if(auto w = dynamic_pointer_cast<CToggleGroup>(widget("groupMaxTeams")))
+		w->setSelected(opts->getTeamCount());
+	if(auto w = dynamic_pointer_cast<CToggleGroup>(widget("groupCompOnlyPlayers")))
+		w->setSelected(opts->getCompOnlyPlayerCount());
+	if(auto w = dynamic_pointer_cast<CToggleGroup>(widget("groupCompOnlyTeams")))
+		w->setSelected(opts->getCompOnlyTeamCount());
+	if(auto w = dynamic_pointer_cast<CToggleGroup>(widget("groupWaterContent")))
+		w->setSelected(opts->getWaterContent());
+	if(auto w = dynamic_pointer_cast<CToggleGroup>(widget("groupgroupMonsterStrengthMaxTeams")))
```
IvanSavenko, line 182:
```suggestion
	if(auto w = dynamic_pointer_cast<CToggleGroup>(widget("groupMonsterStrength")))
```

client/gui/InterfaceBuilder.h
Nordsoft91: quite good point, actually, only that function will be needed

config/windows.json
Nordsoft91: you may want to control draw order, but only json array guarantees the sequential iterating

client/gui/InterfaceBuilder.cpp
Nordsoft91: agree, looks weird

client/lobby/RandomMapTab.cpp
Nordsoft91: came to the same conclusion

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -0,0 +1,204 @@
...
+	}
+	return config.String();
+}
+
+std::shared_ptr<CIntObject> InterfaceObjectConfigurable::buildWidget(const JsonNode & config)
+{
+	assert(!config.isNull());
+	auto type = config["type"].String();
+	
+	int x = 0, y = 0;
```
IvanSavenko, line 77: Perhaps use `Point` instead of individual x/y?
Also, if you wish - feel free to add constructor to Button/Label/Picture/etc that accepts Point instead of x/y

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -0,0 +1,204 @@
...
+		return "";
+	
+	if(config.isNumber())
+	{
+		return CGI->generaltexth->allTexts[config.Integer()];
+	}
+	return config.String();
+}
+
+std::shared_ptr<CIntObject> InterfaceObjectConfigurable::buildWidget(const JsonNode & config)
```
IvanSavenko, line 72: (optional, feel free to ignore)
Not sure if I like layout of this method.
First, you load various fields that are shared across some controls... but not all of them so you end up with color/font/alignment that are unused in some routes
And then you load some (also shared) fields in individual blocks, like "zelp"

This is especially eye-jarring with texts - you load "text" for labels, but for buttons you ignore previously loaded "text" and instead load text for buttons from "zelp"

If I were to write such method I would probably split it into group of helper methods, that will load individual elements only in block that actually needs this element, e.g. `std::string loadImage(JsonNode&)` that will only load image in CPicture & CButton constructors.

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -0,0 +1,204 @@
...
+	for(const auto & item : config["items"].Vector())
+	{
+		std::string name = item["name"].isNull()
+						? unnamedObjectPrefix + std::to_string(unnamedObjectId++)
+						: item["name"].String();
+		widgets[name] = buildWidget(item);
+	}
+}
+
+std::string InterfaceObjectConfigurable::buildText(const JsonNode & config) const
```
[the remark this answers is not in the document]
Nordsoft91: I just started working on it, and certainly it will not cover all scenarios. I'll go case by case and see which modifications are needed (probably I should make this review as not ready for merge)

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -0,0 +1,386 @@
...
+}
+
+std::shared_ptr<CToggleGroup> InterfaceObjectConfigurable::buildToggleGroup(const JsonNode & config) const
+{
+	auto position = readPosition(config["position"]);
+	auto group = std::make_shared<CToggleGroup>(0);
+	group->pos += position;
+	if(!config["items"].isNull())
+	{
+		SObjectConstruction obj__i(group.get());
```
IvanSavenko, line 205: Perhaps hide this behind another OBJ_CONSTRUCTION macro?
Not huge fan of macro, but this is how `SObjectConstruction` is always used in code

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -0,0 +1,386 @@
...
+{
+	assert(!config.isNull());
+	//overrides from variables
+	for(auto & item : config["overrides"].Struct())
+	{
+		config[item.first] = variables[item.second.String()];
+	}
+	
+	auto type = config["type"].String();
+	if(type == "picture")
```
IvanSavenko, line 336: Not sure about this if-forest.
Perhaps generate static std::map<std::string, std::function> instead?

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -0,0 +1,386 @@
...
+	}
+	if(type == "slider")
+	{
+		return buildSlider(config);
+	}
+	if(type == "custom")
+	{
+		return const_cast<InterfaceObjectConfigurable*>(this)->buildCustomWidget(config);
+	}
+	return std::shared_ptr<CIntObject>(nullptr);
```
IvanSavenko, line 380:
```suggestion
	logGlobal->error("Failed to construct object of type '%s'", type);
	return std::shared_ptr<CIntObject>(nullptr);
```
What if there is a typo in config file?
No warning, no assert, no throw, just silently create dummy object.

client/lobby/RandomMapTab.cpp
```diff
@@ -205,60 +169,153 @@ void RandomMapTab::updateMapInfoByHost()
...
+			deactivateButtonsFrom(*w, {-1});
+	}
+	if(auto w = widget<CToggleGroup>("groupMonsterStrength"))
+		w->setSelected(opts->getMonsterStrength());
+	if(auto w = widget<CButton>("templateButton"))
+	{
+		if(tmpl)
+			w->addTextOverlay(tmpl->getName(), EFonts::FONT_SMALL);
+		else
+			w->addTextOverlay("default", EFonts::FONT_SMALL);
```
IvanSavenko, line 284: Not 100% sure, but shouldn't this string be translatable?
Also, if "default" means "template will be selected automatically" then I think it should say "Random"

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -0,0 +1,386 @@
...
+	}
+	if(type == "labelGroup")
+	{
+		return buildLabelGroup(config);
+	}
+	if(type == "slider")
+	{
+		return buildSlider(config);
+	}
+	if(type == "custom")
```
IvanSavenko, line 376: Perhaps instead of using "custom" type `buildCustomWidget` should be called unconditionally? (in cases when type was not caught by this if-forest)
This way, `buildCustomWidget` will be able to define its own widgets with any name, with less awkward syntax in json.  Useful if UI needs several different custom widgets.

client/lobby/RandomMapTab.h
```diff
@@ -23,42 +25,80 @@ VCMI_LIB_NAMESPACE_END
...
-	std::shared_ptr<CToggleGroup> groupMonsterStrength;
-	std::shared_ptr<CButton> buttonShowRandomMaps;
 	std::shared_ptr<CMapGenOptions> mapGenOptions;
 	std::shared_ptr<CMapInfo> mapInfo;
+	
+	//options allowed - need to store as impact each other
+	std::set<int> playerCountAllowed, playerTeamsAllowed, compCountAllowed, compTeamsAllowed;
+};
+
+class TemplatesDropBox : public InterfaceObjectConfigurable
```
IvanSavenko, line 54: Hmm... Kind of makes sense to have such widget is generic form, so it can be reused in any other new UI, since H3 UI lacks such drop box.
Don't want to end up with bunch of different drop box, with each having its own slightly different functionality.

Not sure whether this can be done in reasonable amount of time, so feel free to ignore.

IvanSavenko (changes requested):

client/gui/InterfaceObjectConfigurable.cpp
IvanSavenko: Or, create method `protected: registerWidget(std::string name, std::function builder)`
that will internally add name & builder functor to std::map and any inheritors can register their own widgets instead of relying on `buildCustomWidget` method

client/gui/InterfaceObjectConfigurable.cpp
Nordsoft91: I copy-pasted this usage from some other ui piece

client/gui/InterfaceObjectConfigurable.cpp
Nordsoft91: Good point, I need to cover this functionality with logging

client/lobby/RandomMapTab.cpp
Nordsoft91: > shouldn't this string be translatable?
Should be
> it should say "Random"
I took all of this from HDMod to have exactly as there. Anyway will make it configurable

client/gui/InterfaceObjectConfigurable.cpp
Nordsoft91: but it's type, not name. Name is separate field, so you can have many custom widgets with their specific names. And making them unconditionally can bring compatibility problems in future (when some version supports some new widgets)

client/lobby/RandomMapTab.h
Nordsoft91: Not sure if I can do this generic because I rely on pop-up window functionality when show drop-box. I thought about it, but not trying to capture all possible scenarios for now.

client/gui/InterfaceObjectConfigurable.cpp
IvanSavenko: Sorry, not "name" but "type".
If you're creating items for dropbox it makes sense for these items to have "dropboxItem" type (instead of "custom").

And some classes (adventure map or battles) you may end up with several custom widgets, e.g. battle queue, battle log.

So it makes sense to allow derived classes to inject their own widgets with separate "type", e.g. `"type" : "battleLog"` even though these types are not defined in basic interface builder class.

client/gui/InterfaceObjectConfigurable.cpp
IvanSavenko: Ah, yeah, I see such usage on an older branch.
Personally, I'd still hide it behind macro - it is more recognizable than name of actual class behind that macro. 
Perhaps something like this?
`#define OBJECT_CONSTRUCTION_TARGETED(target) SObjectConstruction obj__i(target)`
Anyway, this is just a minor suggestion, feel free to ignore.

client/gui/InterfaceObjectConfigurable.cpp
Nordsoft91: There is problem with that - when modders will see into those examples they may think that it's possible to create "dropboxitem" in their new configs. But it won't as processing "custom" objects is implemented in override function

IvanSavenko: Hmm... could be solved by using better names, e.g. "rmgTemplateItem", indicating that this item was designed specifically for rmg templates.
"custom" is even worse using the same logic - modders may try to use parameters from one "custom" element in another.

It also very... nondescript. What exactly is "custom"? How it appears in UI? How can it be configured? Better name will provide hints by itself.

Nordsoft91: But how it limits you? You still have all fields from object. You can define and process as many objects inside custom processing as you want. Custom means that object is not a standard but specific. As for me it doesn't confuse.

IvanSavenko: Personally I prefer concrete/strongly defined terms/forms, Meanwhile, "custom" is very generic. Sure, it is flexible, but not sure if there is need for such flexibility.

Anyway, don't see any point in arguing. This is not something that will be set in stone, for now let's keep it as it.
I'm sure that some design issues will arise later anyway once this UI customization has been be expanded to other UI's.

client/gui/InterfaceObjectConfigurable.cpp
Nordsoft91: Change like this seems overkilling, buildCustomWidget doesn't work in a way designed:
`No viable conversion from 'std::shared_ptr<CIntObject> (TemplatesDropBox::*)(const JsonNode &)' to 'std::function<std::shared_ptr<CIntObject> (const InterfaceObjectConfigurable *, const JsonNode &)>'`

IvanSavenko: Static map won't work here - disregard my first comment, it needs to be non-static member of InterfaceObjectConfigurable
```
using BuilderFunctor = std::function<std::shared_ptr<CIntObject>(const JsonNode &)>;
std::map<std::string, builderFunctor> functors;

void InterfaceObjectConfigurable::registerWidget(std::string name, const BuilderFunctor & builder)
{
    functors["button"] = builder;
}

InterfaceObjectConfigurable::InterfaceObjectConfigurable()
{
    registerWidget("button", [this](const JsonNode & node){
        return buildButton(node);
    }
}

TemplatesDropBox::TemplatesDropBox()
{
    registerWidget("custom", [this](const JsonNode & node){
        return buildCustomWidget(node);
    }
}
```

IvanSavenko: If you still can't make it work - then OK, leave it as it.

Nordsoft91: Unfortunately, for some of those builders JsonNode is not enough - I've added pointer to InterfaceObjectConfigurable (this), because builder may use internal variables from class instance.
Still can be solved with std::bind, but that's my concern - seems like overcomplicating as amount of types are limited

client/gui/InterfaceObjectConfigurable.cpp
Nordsoft91: I took a look at config json and now agree with you - having multiple "custom" objects without any identification is misleading. Fixed

client/gui/InterfaceObjectConfigurable.cpp
IvanSavenko: One option would be to replace if-forest of basic classes with std::map but keep call to buildCustomWIdget as it is.

Anyway, don't see this as blocker - this is minor change that can be done at any point later, taking into account any future changes.

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -45,9 +45,10 @@ void InterfaceObjectConfigurable::addCallback(const std::string & callbackName,
 void InterfaceObjectConfigurable::init(const JsonNode &config)
 {
 	OBJ_CONSTRUCTION;
-	
+	logGlobal->info("Building configurable interface object");
```
IvanSavenko, line 63: Could stay at debug level, no need to show this every time user opens window

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -373,14 +405,13 @@ std::shared_ptr<CIntObject> InterfaceObjectConfigurable::buildWidget(JsonNode co
...
-	}
-	return std::shared_ptr<CIntObject>(nullptr);
+
+	logGlobal->debug("Calling custom widget building function");
+	return const_cast<InterfaceObjectConfigurable*>(this)->buildCustomWidget(config);
 }
 
 std::shared_ptr<CIntObject> InterfaceObjectConfigurable::buildCustomWidget(const JsonNode & config)
 {
+	logGlobal->error("Default custom widget builder called");
```
IvanSavenko: Probably better to write the reason why this is an error instead
```suggestion
	logGlobal->error("Attempting to create custom widget, but this window has no custom widgets defined!");
```

IvanSavenko (approved):

client/gui/InterfaceObjectConfigurable.cpp
Nordsoft91: ok, looks like I found a more or less nice way of doing that

PR #1227, opened 2022-12-13
Port to OpenDingux handhelds

Shatur:
I trying to port the engine to OpenDingux devices (I own Anbernic RG280V). I had to make some changes in order to compile the project.
What I did:
* Nullkiller is enabled only when TBB is available (I can't compile TBB for this architecture).
* "resolution is not available" is no longer a hard error, needed in order to run. My device have 320x480 display and If I modify `settings.json` and `resolutions.json` to use my native resolution, some elements go off screen. So I need to render in a resolution that is higher then my display (I use 800x1200, 2.5x multiplier, it downscales to my). Maybe there is an easy way to render in my native resolution and downscale everything?
* These devices use [uClibc](https://www.uclibc.org/), so some things aren't available, I added the needed defines. This is not something super rare, for example Chromium has [such checks too](https://chromium.googlesource.com/chromium/src/build/+/refs/heads/main/build_config.h#83).

The game runs:
IMG_20221213_040231


nullkiller:
Does gameplay itself work? The device does not look quite powerful.

Shatur:
@nullkiller I can play https://github.com/ihhub/fheroes2 just fine, so I assumed that vcmi should also work. The device handles emulation up to PS1.

But I actually have a problem with starting a battle, here is my log: [gmenu2x.log](https://github.com/vcmi/vcmi/files/10216418/gmenu2x.log).
I just unable to connect to a server. Maybe you have any idea why it happens?
I applied https://github.com/vcmi/vcmi/pull/1204 and use `ENABLE_SINGLE_APP_BUILD`.

nullkiller:
The app still use socket connection even if it is in single app mode. And looks like it can not contact localhost according to the logs.

IvanSavenko:
1. Resolution. Your display is 320px wide and 480px in height? In that case - yeah, downscaling 2.5 times is the only option. 
VCMI does not have any internal means for scaling and it needs at least resolution of at least 800x600 for battles and town screens.

2. You've mentioned Chromium so device is network-capable. Perhaps something is broken in #1204, will recheck.

CMakeLists.txt
```diff
@@ -312,7 +312,7 @@ find_package(SDL2_ttf REQUIRED)
 if(TARGET SDL2_ttf::SDL2_ttf)
 	add_library(SDL2::TTF ALIAS SDL2_ttf::SDL2_ttf)
 endif()
-find_package(TBB REQUIRED)
+find_package(TBB)
```
IvanSavenko, line 315: Instead of silently failing in case of missing TBB (and not building NullkillerAI) can you add option? 
Something like
`option(ENABLE_NULLKILLER_AI "Enable compilation of nullkiller AI library" ON)`
And if this option is off - don't try looking for TBB and don't include nullkiller subdirectory

IvanSavenko (changes requested):

Shatur:
Still investigating, I tried calling `ip addr add 127.0.0.1/8 dev lo` and now 127.0.0.1 appears in `ip addr`, but I still can't connect to it. Maybe something else needs to be done? Also I noticed that I can't ssh to 127.0.0.1. Doesn't make much sense, but it's usually possible on a machine with ssh server running.

Shatur:
A friend helped me figure this out. It turns out loopback is disabled by default on this device. I enabled it with `ip link set lo up`.
Now I can start the game, but it fails with the following error:
```
Opening VCAI
Cannot open dynamic library (/mnt/Files/Code/gcw0-toolchain/vcmi/install/lib/vcmi/AI/libVCAI.so). Throwing...
terminate called after throwing an instance of 'std::runtime_error'
  what():  Cannot open dynamic library
Aborted
```
On these devices apps are bundled in a single squashfs image with all content. But it tries to load AI dll from a path from my PC. Could anyone point me to where I can change it?

IvanSavenko:
Hmmm... I think you want set CMAKE_INSTALL_PREFIX to your mount point.
VCMI generally expects:

`${CMAKE_INSTALL_PREFIX}/bin/` - vcmiclient & vcmiserver binaries (or just vcmiclient in case of single-app build)
`${CMAKE_INSTALL_PREFIX}/lib/vcmi/` - libvcmi.so & AI directory (there is also Scripting directory, but I think it is off by default)
`${CMAKE_INSTALL_PREFIX}/share/vcmi/` - one of the possible locations for H3 data and config & Mods directories from source code

Another option is to set ENABLE_MONOLITHIC_INSTALL=ON
In this case, vcmi will look for libraries and data in the same directory as vcmiclient, e.g.
```
./vcmiclient
./libvcmi.so
./AI/libVCAI.so
```

Shatur:
> Another option is to set ENABLE_MONOLITHIC_INSTALL=ON

Thank you, I think this is just what I need for this platform.
Will write back how the game performs.

Shatur:
It works! But there are some issues:
* Dialogues are displayed in the center, but the playing field is displayed at the top, and the bottom is a shadowing of what was once displayed. I think it has something to do with the resolution of the game :(
* Performance is very low. I think that this is also due to the fact that the game runs at 2.5x resolution.
IMG_20221213_220143

Shatur:
Made the requested changes.
I think this PR could be useful for other handheld devices such as RG300X.

IvanSavenko:
Resolution can be fixed - you'll need to make custom resolution config & edit some UI bitmaps. But that would mean even higher load due to larger part of adventure map visible at once.

But performance... Don't expect any major improvements any time soon I am afraid.
You could try to using (for example) 800x1000 or even default 800x600 resolution instead. This will distort the image, but will reduce area that game needs to render.
Perhaps I (or somebody else) will have time & interest in working on rendering performance (e.g. using hardware acceleration from SDL), but don't expect that soon.

Shatur:
> Resolution can be fixed - you'll need to make custom resolution config & edit some UI bitmaps. But that would mean even higher load due to larger part of adventure map visible at once.

Hm... I edited `settings.json` and `resolutions.json` in order to run in `800x1200` (without this, even dialogues barely fit the screen). But do I understand correctly that I can try to use native resolution and reduce all images manually?

> Perhaps I (or somebody else) will have time & interest in working on rendering performance (e.g. using hardware acceleration from SDL), but don't expect that soon.

Sad to hear, but thanks for letting me know.
Also I forgot to mention loading times. I need to wait about a 5 minutes to enter enter a scenario :D

IvanSavenko (dismissed):

Nordsoft91:
You can add resolution here to avoid downscaling and get performance improvement
https://github.com/vcmi-mods/vcmi-extras

Shatur:
@Nordsoft91 thank you, will try!

I did a typo in the last commit, fixed it.

IvanSavenko:
@Shatur In regards to performance - if you wish you can try applying this patch.
It will produce graphical artifacts if you know where to look for, but may give you some speedup.
This will switch rendering from vcmi methods to more optimized routines from SDL. Should be faster, but don't expect miracles.
```
diff --git a/client/gui/CAnimation.cpp b/client/gui/CAnimation.cpp
index 73f5b4a23..b0983a0b1 100644
--- a/client/gui/CAnimation.cpp
+++ b/client/gui/CAnimation.cpp
@@ -681,14 +681,7 @@ void SDLImage::draw(SDL_Surface* where, const SDL_Rect* dest, const SDL_Rect* sr
 		destRect.y += dest->y;
 	}
 
-	if(surf->format->BitsPerPixel == 8)
-	{
-		CSDL_Ext::blit8bppAlphaTo24bpp(surf, &sourceRect, where, &destRect);
-	}
-	else
-	{
-		SDL_UpperBlit(surf, &sourceRect, where, &destRect);
-	}
+	SDL_UpperBlit(surf, &sourceRect, where, &destRect);
 }
 
 std::shared_ptr<IImage> SDLImage::scaleFast(float scale) const
```

Shatur:
> Another option is to set ENABLE_MONOLITHIC_INSTALL=ON

I tried this option (I just quickly hacked it before), but it results in the following error: `Cannot open dynamic library (/mnt/Files/Code/gcw0-toolchain/vcmi/install/./AI/libVCAI.so). Throwing...`. Is this a bug? I pushed a small change that fixes it.

> if you wish you can try applying this patch.

Thank you! You are right, it runs a little faster. But not much better.
I believe that using a different resolution with [vcmi-extras](https://github.com/vcmi-mods/vcmi-extras) should help, going to try it letter.

Shatur:
Also opened #1231. Needed after latest changes in master. I decided to open a separate PR since it's a very small change.

Shatur:
> I guess you wanted to say "in develop" :)

Right, a habit :)

Hm, my changes related `XX_DIRS` fails on installation step on IOS. But I'm not sure why, not familiar with this platform. Could anyone take a look at the latest commit?

Shatur:
I think I figured it out. The `ENABLE_NULLKILLER_AI` option must be top-level, otherwise the variable will not be propagated.

Shatur:
> I think I figured it out.

Also `install_vcpkg_imported_tgt` should be called after initializing `BIN_DIR`. So I put it inside `AI/CMakeLists.txt` as it was before. Sorry for the mess.

Nordsoft91:
One thing which I don't know how to handle properly: perhaps there will be possible to select NullkillerAI from launcher which can lead to crash (need to test)

Shatur:
> perhaps there will be possible to select NullkillerAI from launcher 

Tested on my PC, NullkillerAI doesn't show up in the list if it's not there.

Shatur:
Thanks! Will try it later.
Let me know if you need anything else to get this PR merged.

IvanSavenko:
You can use bmp/png in place of pcx.
And in vcmi def's can be replaced with json+bmp/png/etc, see [cancelButton.json](https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/Sprites/stackWindow/cancelButton.json) as example

IvanSavenko:
Downscaling is not possible at the moment. And even if you make custom resolution for your screen you still have town & battle screens that are fixed at size at 800x600

CMakeLists.txt
```diff
@@ -388,20 +392,20 @@ else()
 		set(DATA_DIR "." CACHE STRING "Where to install data files")
 	else()
 		if(NOT BIN_DIR)
-			set(BIN_DIR "bin" CACHE STRING "Where to install binaries")
+			set(BIN_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE STRING "Where to install binaries")
```
IvanSavenko, line 395: CMake script later uses BIN_DIR for `install()` command, meaning that BIN_DIR (and others) should stay without CMAKE_INSTALL_PREFIX.
I also don't see how this could have effect on ENABLE_MONOLITHIC_INSTALL since these definitions are explicitly in else() block and can only be reached if ENABLE_MONOLITHIC_INSTALL is NOT set.

(same issue with LIB_DIR & DATA_DIR)

I think what you want to do is to call `add_definitions` separately for monolithic install and for standard install

IvanSavenko (changes requested):

Shatur:
> You can use bmp/png in place of pcx.
> And in vcmi def's can be replaced with json+bmp/png/etc, see [cancelButton.json](https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/Sprites/stackWindow/cancelButton.json) as example

Thanks for the info!

> Downscaling is not possible at the moment. And even if you make custom resolution for your screen you still have town & battle screens that are fixed at size at 800x600

Yes, I understand that in VCMI Extras we basically adapt HUD. I just thought that I can manually downscale other things like main menu with buttons, battle screen, towers, etc. and provide it in a form of mod.
I'm not sure if it's worth it, but maybe I'll try it someday :smile:
I think that on models with a bigger resolution, the game will work better. Mine is the most portable version.

lib/CModHandler.h
```diff
@@ -14,6 +14,12 @@
 #include "VCMI_Lib.h"
 #include "JsonNode.h"
 
+#ifdef __UCLIBC__
+#undef major
+#undef minor
+#undef patch
+#endif
```
Nordsoft91, line 21: Could you please elaborate why it's needed? And where __UCLIBC__ is defined?

Shatur: uClibc is a small standard C library like glibc or musl. And this stdlib have these macro defined :(
This is why I have to undefine them.

kambala-decapitator: wouldn't it be better to rename variables in VCMI code to avoid the clash?

Shatur: I just took the path of least change.
Let me know if you want to use different variable names instead.

IvanSavenko: @kambala-decapitator 
Nah, I think it's fine.
Should we also ban usage of "min" and "max" because windows.h has those defines?

Personally I'd call this bug of uglibc since library exposes its internal defines using names that are not classified as reserved.

kambala-decapitator: @IvanSavenko fair :)

IvanSavenko (approved):

IvanSavenko:
If there are no objections, then I'm merging these changes into develop tomorrow

PR #1231, opened 2022-12-14
Fix SharedMemory declaration

Shatur:
Should be under VCMI_LIB_NAMESPACE_BEGIN as its definition. But this forward declaration is not used, so I removed it. Also my GCC 11 for MIPS fails to compile the project without this change.
Required for #1227. The error was introduced after latest changes in master.

### DIFF
--- a/client/CServerHandler.h
+++ b/client/CServerHandler.h
@@ -25,7 +25,6 @@ class CGameState;
 struct ClientPlayer;
 struct CPack;
 struct CPackForLobby;
-struct SharedMemory;
 
 template<typename T> class CApplier;
 
--- a/client/CServerHandler.h
+++ b/client/CServerHandler.h
@@ -25,12 +25,11 @@ class CGameState;
 struct ClientPlayer;
 struct CPack;
 struct CPackForLobby;
+struct SharedMemory;
 
 template<typename T> class CApplier;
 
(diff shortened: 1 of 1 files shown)


client/CServerHandler.h
```diff
@@ -25,13 +25,12 @@ class CGameState;
 struct ClientPlayer;
 struct CPack;
 struct CPackForLobby;
+struct SharedMemory;
```
Nordsoft91, line 28: SharedMemory struct is defined in vcmi/interprocess.h inside lib namespace, so adding this outside of namespace  is not valid.
Instead, you can remove forward declaration SharedMemory from CServerHandler as it's unused

Nordsoft91 (changes requested):

client/CServerHandler.h
Shatur: > SharedMemory struct is defined in vcmi/interprocess.h inside lib namespace, so adding this outside of namespace is not valid.

Yes, this is why I moved it under lib namespace.

>  Instead, you can remove forward declaration SharedMemory from CServerHandler as it's unused as I can see

Right, makes sense! Maybe it compiled because it wasn't used. But for some reason failed on my compiler.
Removed.

Nordsoft91 (approved):

Issue #1228, opened 2022-12-13
Second AI activation


nullkiller:
There is something strange in this save. Purple AI activates in parallel with green AI. Sometimes it leads to crash. Mods: grove and haven.
[Saves.zip](https://github.com/vcmi/vcmi/files/10217079/Saves.zip)

PR #1232, opened 2022-12-14
#1228 - prevent second AI activation on AI defeat due to wrong EndTur…

nullkiller:
prevent second AI activation on AI defeat due to wrong EndTurn packet. Defeated AI sends EndTurn normally and it actually ends turn for the next AI, This causes two AI to run in parallel.
- Fixed EndTurn hadling, thanks to @IvanSavenko 
- Added lock on shared storage for Nullkiller::makeTurn to avoid its concurrent usage

### DIFF
--- a/server/NetPacksServer.cpp
+++ b/server/NetPacksServer.cpp
@@ -89,7 +89,17 @@ bool SaveGame::applyGh(CGameHandler * gh)
 
 bool EndTurn::applyGh(CGameHandler * gh)
 {
-	PlayerColor player = GS(gh)->currentPlayer;
+	PlayerColor currentPlayer = GS(gh)->currentPlayer;
+	if(player != currentPlayer)
+	{
+		if(gh->getPlayerStatus(player) == EPlayerStatus::INGAME)
+			throwAndComplain(gh, "Player attempted to end turn for another player!");
+
+		logGlobal->debug("Player attempted to end turn after game over. Ignoring this request.");
+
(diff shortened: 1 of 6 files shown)


server/NetPacksServer.cpp
```diff
@@ -89,7 +89,17 @@ bool SaveGame::applyGh(CGameHandler * gh)
 
 bool EndTurn::applyGh(CGameHandler * gh)
 {
-	PlayerColor player = GS(gh)->currentPlayer;
+	PlayerColor currentPlayer = GS(gh)->currentPlayer;
+	if(player != currentPlayer)
+	{
+		if(gh->getPlayerStatus(player) == EPlayerStatus::INGAME)
+			throwAndComplain(gh, "Player attempted to end turn for another player!");
```
Nordsoft91, line 96: will throwAndComplain close the server? Should complain and return false be enough in that case?

IvanSavenko: No, it won't close the server. And this looks to be standard way to abort processing of illegal pack (see rest of the file).

nullkiller: It will show green message. And if the player is already lost we treat it as a more or less regular flow and just log warning.
I tried to solve it from AI side but the thing is too asynchronous by design so only with some thread::sleep(100) I was able to correctly shut down AI thread.

Nordsoft91 (approved):

PR #1235, opened 2022-12-16
Fix cmake for macOS build

Nordsoft91:


CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ option(ENABLE_GITVERSION "Enable Version.cpp with Git commit hash" ON)
 option(ENABLE_DEBUG_CONSOLE "Enable debug console for Windows builds" ON)
 option(ENABLE_MULTI_PROCESS_BUILDS "Enable /MP flag for MSVS solution" ON)
 option(ENABLE_SINGLE_APP_BUILD "Builds client and server as single executable" OFF)
+option(COPY_CONFIG_ON_BUILD "Copies config folder into output directory at building phase" ON)
```
kambala-decapitator, line 64: maybe better to place the option after `ENABLE_TEST` in the "not iOS" branch (line 55 here)?

Nordsoft91: by default for iOS it will be OFF in such case?

kambala-decapitator: iOS should pack everything in the build folder (inside the .app) unconditionally, otherwise app wouldn't work. If I understand the purpose of this setting right.

Nordsoft91: not sure, I didn't testes iOS. This fixes problem on macOS

PR #1240, opened 2022-12-17
lib/CArtHandler.cpp: avoid return of already destroyed objects

trofi:
`gcc-13` detects return of dangling objects as:

    lib/CArtHandler.cpp: In function 'const std::vector<ArtifactPosition::EArtifactPosition>& ArtifactUtils::unmovableSlots()':
    lib/CArtHandler.cpp:1554:9: warning: returning reference to temporary [-Wreturn-local-addr]
     1554 |         };
          |         ^
    lib/CArtHandler.cpp: In function 'const std::vector<ArtifactPosition::EArtifactPosition>& ArtifactUtils::constituentWornSlots()':
    lib/CArtHandler.cpp:1575:9: warning: returning reference to temporary [-Wreturn-local-addr]
     1575 |         };
          |         ^

The change converts local objects to constants and returns references to them to guarantee that they outlive local functions.

### DIFF
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -1547,17 +1547,16 @@ DLL_LINKAGE ArtifactPosition ArtifactUtils::getArtifactDstPosition(	const CArtif
 
 DLL_LINKAGE const std::vector<ArtifactPosition::EArtifactPosition> & ArtifactUtils::unmovableSlots()
 {
-	return
-	{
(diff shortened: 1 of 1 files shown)


IvanSavenko:
Already fixed in beta branch ( #1183 ), fix will be merged into develop once 1.1 releases

trofi:
Ah, nice! Closing.

Issue #1116, opened 2022-11-08
Assembly of admiral hat crashes the game


val-gaav:
Game crashes when hero tries to assemble combination artifact admiral hat .

vcmiclient: /home/avalon/vcmi/lib/NetPacksLib.cpp:1122: void AssembledArtifact::applyGs(CGameState*): Assertion `vstd::contains(transformedArt->assemblyPossibilities(artSet), builtArt)' failed.
Network error receiving a pack. Connection Connection with VCMI 1.1.0.ae1d2c50e2d000df5ce5569550f8722345c8067a (server) (ID: 1 UUID: dffcdec6-c95a-4b99-9728-a2002a4dc255) dies. What happened: read: Connection reset by peer
Aborted (core dumped)

Way to reproduce - load attached save , pick up the artifact with Kanim hero try to assemble admiral hat and trigger the error. 
 
BTW I also tested with a fresh map and with using vcmiforgeofnoldorking cheatcode that gives all artifacts to the hero. Assembly and disassembly of admiral hat works just fine. The crash is for some reason linked to this specific map situation.

My own linux build - develop branch from November 3 2022

Attached saves and logs 
[logs.zip](https://github.com/vcmi/vcmi/files/9960460/logs.zip)
[crash_Saves_2.zip](https://github.com/vcmi/vcmi/files/9960578/crash_Saves_2.zip)

nullkiller:
Could you try exactly same sequence on a fresh map? Check what was worn, do the same and pick artifact which causes crash. Can you try to assemble it with different hero? I can not load save without mods, it is very strict now.

nullkiller:
Looks like to reproduce it we need to save/load new game. The assert triggers because the address of assembled artifact is different from the expected one. We can use comparison by ArtifactID as an option. But on daily builds assert check is disabled.

kambala-decapitator:
FYI plain save with just the artifact can be found in https://github.com/vcmi/vcmi/files/9689733/seer-2-artifacts.h3m.zip (#840)

nullkiller:
I think the issue somehow relates to the fact that on loading save there is some magic in loading VLC from save file. So after save load there are tow different instances of same CArtifact so comparing them by address is not always an option

nullkiller:
The question is how normal is it. Game does not crash with disabled asserts.

PR #1243, opened 2022-12-18
#1116 - fix assertion

nullkiller:
Fix assertion

### DIFF
--- a/lib/NetPacksLib.cpp
+++ b/lib/NetPacksLib.cpp
@@ -1179,14 +1179,6 @@ DLL_LINKAGE void AssembledArtifact::applyGs(CGameState *gs)
 
 	UNUSED(transformedArt);
 
-	if(!vstd::contains_if(transformedArt->assemblyPossibilities(artSet, combineEquipped), [=](const CArtifact * art)->bool
-		{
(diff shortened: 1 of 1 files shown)


lib/NetPacksLib.cpp
```diff
@@ -1171,9 +1171,22 @@ DLL_LINKAGE void AssembledArtifact::applyGs(CGameState *gs)
...
+		}));
+
 	UNUSED(transformedArt);
 
+	if(!vstd::contains_if(transformedArt->assemblyPossibilities(artSet, combineEquipped), [=](const CArtifact * art)->bool
+		{
+			return art->id == builtArt->id;
+		}))
+	{
+		throw std::exception("Bad");
```
Nordsoft91, line 1187:
```suggestion
		throw std::runtime_error("Bad");
```

Nordsoft91 (changes requested):

lib/NetPacksLib.cpp
nullkiller: Ouch, this is extra here)

Nordsoft91 (approved):

PR #1245, opened 2022-12-18
Artifact assembly fix

### LINKED ISSUE #840: client bugs when assembling artefact set

IvanSavenko:
Fixes #840 - Correctly update drag-and-drop cursor if window with artifacts gets deactivated due to pop-up dialog
Fixes artefact-related assertion failures
Fixes Exchange Window not getting updated on assembly/deassembly of an artifact

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -691,10 +691,8 @@ void CArtifactsOfHero::updateParentWindow()
 
 		if(!updateState)
 		{
-			cew->deactivate();
 			cew->updateWidgets();
 			cew->redraw();
-			cew->activate();
 		}
 	}
 }
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -254,6 +254,22 @@ void CHeroArtPlace::clickRight(tribool down, bool previousState)
 	}
 }
 
+void CArtifactsOfHero::activate()
+{
+	if (commonInfo->src.AOH == this && commonInfo->src.art)
+		CCS->curh->dragAndDropCursor(make_unique<CAnimImage>("artifact", commonInfo->src.art->artType->getIconIndex()));
+
+	CIntObject::activate();
+}
+
+void CArtifactsOfHero::deactivate()
+{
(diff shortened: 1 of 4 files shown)


client/widgets/CArtifactHolder.cpp
```diff
@@ -254,6 +254,22 @@ void CHeroArtPlace::clickRight(tribool down, bool previousState)
...
+}
+
+void CArtifactsOfHero::deactivate()
+{
+	if (commonInfo->src.AOH == this && commonInfo->src.art)
+		CCS->curh->dragAndDropCursor(nullptr);
+
+	CIntObject::deactivate();
+}
+
```
Nordsoft91, line 272: Could you please move those methods closer to other CArtifactsOfHero methods? Currently they are in the middle of CHerOArtPlace definitions

lib/CArtHandler.cpp
```diff
@@ -1062,7 +1060,9 @@ void CCombinedArtifactInstance::createConstituents()
 
 void CCombinedArtifactInstance::addAsConstituent(CArtifactInstance *art, ArtifactPosition slot)
 {
-	assert(vstd::contains(*artType->constituents, art->artType.get()));
+	assert(vstd::contains_if(*artType->constituents, [=](const CArtifact * constituent){
```
Nordsoft91, line 1063:
```suggestion
	assert(artType && art && art->artType);
	assert(vstd::contains_if(*artType->constituents, [art](const CArtifact * constituent){
```

PR #1246, opened 2022-12-18
[1.1] Mod validation fix

IvanSavenko:
Fixes for validation errors on startup:

- Added "obstacles" section in mod.json schema
- Added "compatibility" section in mod.json schema
- Added human-readable error reporting for access to invalid terrain in mods
- Moved tower shooter definitions to "vcmi" mods - required for file presence validation checks to pass since tower shooter
icons reside in vcmi mod

### DIFF
--- a/Mods/vcmi/config/vcmi/towerFactions.json
+++ b/Mods/vcmi/config/vcmi/towerFactions.json
@@ -1,0 +1,101 @@
+{
+	"core:castle" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:rampart" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:tower" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:inferno" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:necropolis" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:dungeon" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:stronghold" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:fortress" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
+	"core:conflux" :
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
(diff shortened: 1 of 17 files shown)


config/schemas/mod.json
```diff
@@ -56,6 +56,13 @@
 			"description": "List of mods that can't be enabled in the same time as this one",
 			"items": { "type":"string" }
 		},
+		"compatibility" : {
+			"type":"object",
+			"description": "List of supported version of vcmi engine",
+			"additionalProperties" : {
+				"type" : "string"
+			}
```
Nordsoft91, line 64: My fault for not making schemas for this
```suggestion
			"description": "Supported versions of vcmi engine",
			"additionalProperties" : false,
			"properties" : {
				"min" : {
					"type" : "string",
					"description" : "minimal compatible vcmi engine version in a format major.minor.patch. When specified, earlier versions won't be supported",
					"pattern" : "^\d+\.\d+\.\d+$"
				},
				"max" : {
					"type" : "string",
					"description" : "maximum compatible vcmi engine version in a format major.minor.patch. When specified, later versions won't re supported",
					"pattern" : "^\d+\.\d+\.\d+$"
				}
			}
```

lib/mapObjects/ObjectTemplate.cpp
```diff
@@ -287,7 +287,15 @@ void ObjectTemplate::readJson(const JsonNode &node, const bool withTerrain)
 	if(withTerrain && !node["allowedTerrains"].isNull())
 	{
 		for(auto& entry : node["allowedTerrains"].Vector())
-			allowedTerrains.insert(VLC->terrainTypeHandler->getInfoByName(entry.String())->id);
+		{
+			try {
+				allowedTerrains.insert(VLC->terrainTypeHandler->getInfoByName(entry.String())->id);
```
Nordsoft91, line 292: getInfoByName returns pointer but throws exception instead of returning nullptr which doesn't make much sense
It's needed to be refactored in future in the way of either returning reference or returning nullptr without exception

Mods/vcmi/config/vcmi/towerFactions.json
```diff
@@ -0,0 +1,101 @@
...
+	{
+		"town" :
+		{
+			"siege" :
+			{
+				"towerIconSmall" : "vcmi/creatureIcons/towerSmall",
+				"towerIconLarge" : "vcmi/creatureIcons/towerLarge",
+			}
+		}
+	},
```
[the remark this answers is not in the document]
IvanSavenko: Long-term solution is probably merge our virtual "core" mod and helper "vcmi" mod into single non-virtual mod and move all moddable config files into such mod. (probably name it "core" for compatibility with existing mods).
This would keep all data in a single place & less special cases to handle in our code.

As for overriding - for new factions nothing would change. They don't use this json in any way. Compatibility icons for unupdated mods are set in arrow tower "creature", which is shared across all towns.

vcmi-extras - will recheck tomorrow morning, worst-case scenario is that that mod will need to be marked as depended on "vcmi" mod, to guarantee correct load order.

config/factions/neutral.json
```diff
@@ -2,7 +2,6 @@
 	"neutral" :
 	{
 		"name" : "Neutral",
-		"index" : 9,
```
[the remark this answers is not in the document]
IvanSavenko: Looks to be largely unused. RMG uses it to define zones without towns. No uses other than that.
Should probably add special constant for that RMG case, e.g. NONE=-2 so it won't get allocated to any mod.

Interesting detail - there is nothing that would prevent modders from giving town to this "neutral" faction. So they can break RMG already.

lib/mapObjects/ObjectTemplate.cpp
IvanSavenko: Yup. Problem is, there are like 5 or so usages of that method alone plus unknown number of usages of similar methods for roads/rivers. And haven't seen any obvious try/catch blocks around others.

Don't want to change all of them, so for now I'd settle on "Failed to find terrain %s for object %s" message which is much more useful for modders than current "exception thrown by map::at" or something like that

config/factions/neutral.json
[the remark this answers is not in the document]
Nordsoft91: "there is nothing that would prevent modders from giving town to this "neutral" faction" - yes, but it will be mod problem, not engine problem. Another story is when mod is good but game is broken

lib/mapObjects/ObjectTemplate.cpp
Nordsoft91: Yeah, agree that for release branch no need to make such refactoring, just noting without action needed

nullkiller (approved):

config/factions/neutral.json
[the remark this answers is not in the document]
IvanSavenko: Decided to revert this change - RMG case is more complex than I thought and don't want to play "find all regressions before release" game.

Instead, changed warning rules in regards to "index" field - now, it can only be used by "core" mod. This resolves original reason behind this change

config/schemas/mod.json
IvanSavenko: Done (partially). Looks like our schema validation does not supports checks by pattern, so for now game will only check for presence of min/max fields

config/schemas/mod.json
```diff
@@ -56,6 +56,23 @@
...
 		},
+		"compatibility" : {
+			"type":"object",
+			"description": "Supported versions of vcmi engine",
+			"additionalProperties" : false,
+			"properties" : {
+				"min" : {
+					"type" : "string",
+					"description" : "minimal compatible vcmi engine version in a format major.minor.patch. When specified, earlier versions won't be supported"
+					//"pattern" : "^\\d+\\.\\d+\\.\\d+$" // Not implemented in schema support
```
Nordsoft91, line 67: hm, I though it built-in fields in json-schema

PR #1251, opened 2022-12-19
[HDMod features] Alternative battle actions

Nordsoft91:
Plan is to make a submod for vcmi-extras to enable/disable this feature. To do that as well as to make it not so ugly, I need this PR to be merged: https://github.com/vcmi/vcmi/pull/1225
Снимок экрана 2022-12-20 в 03 30 25
Снимок экрана 2022-12-20 в 03 30 10
Снимок экрана 2022-12-20 в 03 30 04


IvanSavenko:
Hmm.. Feel free to go ahead, but I was actually thinking about a bit more drastic changes, especially after you configurable UI PR.
It should be possible to allow much more customizable UI for battle screen than just reordering of that panel/ribbon/etc at the bottom. For that, we'll need to:
1) Convert BattleControlPanel into BattleWindow (with both StackQueue and Battlefield objects becoming its children)
2) Remove from BattleInterface all GUI-related logic and move it to BattleWindow
3) Make BattleWindow use your configurable UI functionality

With such changes we can make significant redesign taking into account common requests.
Very, very WIP mock-up of such redesign (note that it has some free space that can be allocated to what you're suggesting)
vcmi-battle

Nordsoft91:
That's ok, go ahead, I'll rebase using your latest changes. And yes, configurable interface will give ability to modders to have different UI views. Step 3 I'll do after merging by previously PR

PR #1257, opened 2022-12-21
[1.1] Fixes #1254 - Allow loading of vcmi maps with custom heroes 

### LINKED ISSUE #1254: Regression: custom maps cannot be started
1. Create any custom map in editor, using hero from mods
2. Launch this map

Expected result: map is started ok (as in 1.0)
Actual result: server crashes

Beta branch has this problem from the beggining
Develop branch first appears on fd21f9e

Assuming root cause is #1777

IvanSavenko:


Nordsoft91:
Cannot confirm fix

IvanSavenko:
Pushed another fix, should be OK now

Nordsoft91:
Tested, works well

PR #1264, opened 2022-12-23
Merge beta into develop

Nordsoft91:


IvanSavenko:
I think we actually should merge master->develop instead, so all master commits including merge with beta will be part of develop branch.
Or, since there are merge conflicts:
create branch "master-merge" from develop, merge master into it, submit PR (if you wish I can do this instead)

PR #1265, opened 2022-12-23
Fix android build

Nordsoft91:


Nordsoft91:
There is no sense to wait for build because code is android specific and android is built only from develop and since now for beta and master branches

Issue #938, opened 2022-09-18
create clang-format config file


kambala-decapitator:
This will solve all boring manual code style reviews

PR #1266, opened 2022-12-23
[preview] Clang format

IvanSavenko:
As proposed in #938 , created clang-format settings for vcmi.

This is only preview, I've re-formatted several files to see how it turns out.
So please comment on which part of style you want to change. If you wish to see some other file(s) - let me know.

Once everyone agrees I'll create branch with only clang-format file, without changes to files.
I think better option would be to format files over times as they are changed instead of formatting entire codebase at once, breaking all WIP branches.

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -21,81 +21,81 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 // Based on: http://sourceforge.net/p/predef/wiki/OperatingSystems/
 //	 and on: http://stackoverflow.com/questions/5919996/how-to-detect-reliably-mac-os-x-ios-linux-windows-in-c-preprocessor
 // TODO?: Should be moved to vstd\os_detect.h (and then included by Global.h)
-#ifdef _WIN16			// Defined for 16-bit environments
-#  error "16-bit Windows isn't supported"
-#elif defined(_WIN64)	// Defined for 64-bit environments
-#  define VCMI_WINDOWS
-#  define VCMI_WINDOWS_64
-#elif defined(_WIN32)	// Defined for both 32-bit and 64-bit environments
-#  define VCMI_WINDOWS
-#  define VCMI_WINDOWS_32
+#ifdef _WIN16 // Defined for 16-bit environments
+#	error "16-bit Windows isn't supported"
+#elif defined(_WIN64) // Defined for 64-bit environments
+#	define VCMI_WINDOWS
+#	define VCMI_WINDOWS_64
+#elif defined(_WIN32) // Defined for both 32-bit and 64-bit environments
+#	define VCMI_WINDOWS
+#	define VCMI_WINDOWS_32
 #elif defined(_WIN32_WCE)
-#  error "Windows CE isn't supported"
+#	error "Windows CE isn't supported"
 #elif defined(__linux__) || defined(__gnu_linux__) || defined(linux) || defined(__linux)
-#  define VCMI_UNIX
-#  define VCMI_XDG
-#  ifdef __ANDROID__
-#    define VCMI_ANDROID
-#  endif
+#	define VCMI_UNIX
+#	define VCMI_XDG
+#	ifdef __ANDROID__
+#		define VCMI_ANDROID
+#	endif
 #elif defined(__FreeBSD_kernel__) || defined(__FreeBSD__)
-#  define VCMI_UNIX
-#  define VCMI_XDG
-#  define VCMI_FREEBSD
+#	define VCMI_UNIX
+#	define VCMI_XDG
+#	define VCMI_FREEBSD
 #elif defined(__GNU__) || defined(__gnu_hurd__) || (defined(__MACH__) && !defined(__APPLE__))
-#  define VCMI_UNIX
-#  define VCMI_XDG
-#  define VCMI_HURD
+#	define VCMI_UNIX
+#	define VCMI_XDG
+#	define VCMI_HURD
 #elif defined(__APPLE__) && defined(__MACH__)
-#  define VCMI_UNIX
-#  define VCMI_APPLE
-#  include "TargetConditionals.h"
-#  if TARGET_OS_SIMULATOR || TARGET_IPHONE_SIMULATOR
-#    define VCMI_IOS
-#    define VCMI_IOS_SIM
-#  elif TARGET_OS_IPHONE
-#    define VCMI_IOS
-#  elif TARGET_OS_MAC
-#    define VCMI_MAC
-#  else
+#	define VCMI_UNIX
+#	define VCMI_APPLE
+#	include "TargetConditionals.h"
+#	if TARGET_OS_SIMULATOR || TARGET_IPHONE_SIMULATOR
+#		define VCMI_IOS
+#		define VCMI_IOS_SIM
+#	elif TARGET_OS_IPHONE
+#		define VCMI_IOS
+#	elif TARGET_OS_MAC
+#		define VCMI_MAC
+#	else
 //#  warning "Unknown Apple target."?
-#  endif
+#	endif
 #else
-#  error "VCMI supports only Windows, OSX, Linux and Android targets"
+#	error "VCMI supports only Windows, OSX, Linux and Android targets"
 #endif
 
 // Each compiler uses own way to supress fall through warning. Try to find it.
 // TODO: replace with c++17 [[fallthrough]]
 #ifdef __has_cpp_attribute
-#  if __has_cpp_attribute(fallthrough)
-#    define FALLTHROUGH [[fallthrough]];
-#  elif __has_cpp_attribute(gnu::fallthrough)
-#    define FALLTHROUGH [[gnu::fallthrough]];
-#  elif __has_cpp_attribute(clang::fallthrough)
-#    define FALLTHROUGH [[clang::fallthrough]];
-#  else
-#    define FALLTHROUGH
-#  endif
+#	if __has_cpp_attribute(fallthrough)
+#		define FALLTHROUGH [[fallthrough]];
+#	elif __has_cpp_attribute(gnu::fallthrough)
+#		define FALLTHROUGH [[gnu::fallthrough]];
+#	elif __has_cpp_attribute(clang::fallthrough)
+#		define FALLTHROUGH [[clang::fallthrough]];
+#	else
+#		define FALLTHROUGH
+#	endif
 #else
-#  define FALLTHROUGH
+#	define FALLTHROUGH
 #endif
 
 /* ---------------------------------------------------------------------------- */
 /* Commonly used C++, Boost headers */
 /* ---------------------------------------------------------------------------- */
 #ifdef VCMI_WINDOWS
-#  ifndef WIN32_LEAN_AND_MEAN
-#    define WIN32_LEAN_AND_MEAN		 // Exclude rarely-used stuff from Windows headers - delete this line if something is missing.
-#  endif
-#  ifndef NOMINMAX
-#    define NOMINMAX				 // Exclude min/max macros from <Windows.h>. Use std::[min/max] from <algorithm> instead.
-#  endif
-#  ifndef _NO_W32_PSEUDO_MODIFIERS
-#    define _NO_W32_PSEUDO_MODIFIERS // Exclude more macros for compiling with MinGW on Linux.
-#  endif
+#	ifndef WIN32_LEAN_AND_MEAN
+#		define WIN32_LEAN_AND_MEAN // Exclude rarely-used stuff from Windows headers - delete this line if something is missing.
+#	endif
+#	ifndef NOMINMAX
+#		define NOMINMAX // Exclude min/max macros from <Windows.h>. Use std::[min/max] from <algorithm> instead.
+#	endif
+#	ifndef _NO_W32_PSEUDO_MODIFIERS
+#		define _NO_W32_PSEUDO_MODIFIERS // Exclude more macros for compiling with MinGW on Linux.
+#	endif
 #endif
 
 #ifdef VCMI_ANDROID
-#  define NO_STD_TOSTRING // android runtime (gnustl) currently doesn't support std::to_string, so we provide our impl in this case
+#	define NO_STD_TOSTRING // android runtime (gnustl) currently doesn't support std::to_string, so we provide our impl in this case
 #endif // VCMI_ANDROID
 
 /* ---------------------------------------------------------------------------- */
@@ -103,11 +103,11 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 /* ---------------------------------------------------------------------------- */
 // Compiler (at least MSVC) is not so smart here-> without that displaying is MUCH slower
 #ifdef _MSC_VER
-#  define STRONG_INLINE __forceinline
+#	define STRONG_INLINE __forceinline
 #elif __GNUC__
-#  define STRONG_INLINE inline __attribute__((always_inline))
+#	define STRONG_INLINE inline __attribute__((always_inline))
 #else
-#  define STRONG_INLINE inline
+#	define STRONG_INLINE inline
 #endif
 
 #define _USE_MATH_DEFINES
@@ -117,12 +117,13 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 
 #include <algorithm>
 #include <array>
+#include <atomic>
 #include <cassert>
 #include <climits>
 #include <cmath>
 #include <cstdlib>
-#include <functional>
 #include <fstream>
+#include <functional>
 #include <iomanip>
 #include <iostream>
 #include <map>
@@ -133,18 +134,17 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #include <set>
 #include <sstream>
 #include <string>
-#include <unordered_set>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
-#include <atomic>
 
 //The only available version is 3, as of Boost 1.50
 #include <boost/version.hpp>
 
 #define BOOST_FILESYSTEM_VERSION 3
 #if BOOST_VERSION > 105000
-#  define BOOST_THREAD_VERSION 3
+#	define BOOST_THREAD_VERSION 3
 #endif
 #define BOOST_THREAD_DONT_PROVIDE_THREAD_DESTRUCTOR_CALLS_TERMINATE_IF_JOINABLE 1
 //need to link boost thread dynamically to avoid https://stackoverflow.com/questions/35978572/boost-thread-interupt-does-not-work-when-crossing-a-dll-boundary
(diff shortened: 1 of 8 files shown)


client/gui/CAnimation.h
```diff
@@ -39,113 +39,113 @@ class IImage
...
 
-	//test transparency of specific pixel
+	// test transparency of specific pixel
 	virtual bool isTransparent(const Point & coords) const = 0;
 
 	virtual Point dimensions() const = 0;
-	int width() const;
-	int height() const;
+	int           width() const;
+	int           height() const;
```
kambala-decapitator, line 61: personally I'm against vertical alignment. Maybe allow it only for enums (if possible).

client/gui/CAnimation.h
```diff
@@ -161,17 +161,21 @@ class CFadeAnimation
 public:
 	enum class EMode
 	{
-		NONE, IN, OUT
+		NONE,
+		IN,
+		OUT
```
kambala-decapitator, line 166: I'd add trailing comma after the last element

lib/GameConstants.cpp
```diff
@@ -22,48 +22,48 @@
...
+const PlayerColor PlayerColor::PLAYER_LIMIT     = PlayerColor(PLAYER_LIMIT_I);
+const TeamID      TeamID::NO_TEAM               = TeamID(255);
 
 namespace GameConstants
 {
 #ifdef VCMI_NO_EXTRA_VERSION
 	const std::string VCMI_VERSION = "VCMI " VCMI_VERSION_STRING;
 #else
-	const std::string VCMI_VERSION = "VCMI " VCMI_VERSION_STRING "." + std::string{GIT_SHA1};
+	const std::string VCMI_VERSION = "VCMI " VCMI_VERSION_STRING "." + std::string{ GIT_SHA1 };
```
kambala-decapitator, line 59: personally I don't like wrapping with spaces, but not insisting on either way

lib/GameConstants.cpp
```diff
@@ -22,48 +22,48 @@
...
 #endif
 }
 
 si32 HeroTypeID::decode(const std::string & identifier)
 {
 	auto rawId = VLC->modh->identifiers.getIdentifier(CModHandler::scopeMap(), "hero", identifier);
-	if(rawId)
+	if (rawId)
```
kambala-decapitator, line 66: current code style is to have no space before opening bracket

lib/GameConstants.cpp
```diff
@@ -215,49 +215,43 @@ std::string FactionID::encode(const si32 index)
...
-		{EActionType::SHOOT, "Shoot"},
-		{EActionType::WAIT, "Wait"},
-		{EActionType::CATAPULT, "Catapult"},
-		{EActionType::MONSTER_SPELL, "Monster spell"},
-		{EActionType::BAD_MORALE, "Bad morale"},
-		{EActionType::STACK_HEAL, "Stack heal"},
-		{EActionType::DAEMON_SUMMONING, "Daemon summoning"}
-	};
+	static const std::map<EActionType, std::string> actionTypeToString = { { EActionType::END_TACTIC_PHASE, "End tactic phase" },
+																		   { EActionType::INVALID, "Invalid" },
```
kambala-decapitator, line 219: this is a strange format

lib/GameConstants.h
```diff
@@ -393,85 +432,118 @@ class FactionID : public BaseForID<FactionID, si32>
...
-	STRONG_INLINE
-	bool IsSpecialOrGrail() const
+	BuildingID(EBuildingID _num = NONE)
+		: num(_num)
 	{
-		return num == SPECIAL_1 || num == SPECIAL_2 || num == SPECIAL_3 || num == SPECIAL_4 || num == GRAIL;
 	}
 
+	STRONG_INLINE
+	bool IsSpecialOrGrail() const { return num == SPECIAL_1 || num == SPECIAL_2 || num == SPECIAL_3 || num == SPECIAL_4 || num == GRAIL; }
```
kambala-decapitator, line 545: looks too long for a one-liner

lib/GameConstants.h
```diff
@@ -567,34 +635,53 @@ namespace MappedKeys
...
-	EXPLORER
-};
+	enum EAiTactic
+	{
+		NONE = -1,
+		RANDOM,
+		WARRIOR,
+		BUILDER,
+		EXPLORER
+	};
```
kambala-decapitator, line 645: ok if it matches current code style, although personally I don't like indentation inside namespace

Global.h
```diff
@@ -28,86 +28,86 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
...
-#elif defined(_WIN64)	// Defined for 64-bit environments
-#  define VCMI_WINDOWS
-#  define VCMI_WINDOWS_64
-#elif defined(_WIN32)	// Defined for both 32-bit and 64-bit environments
-#  define VCMI_WINDOWS
-#  define VCMI_WINDOWS_32
+#ifdef _WIN16 // Defined for 16-bit environments
+#error "16-bit Windows isn't supported"
+#elif defined(_WIN64) // Defined for 64-bit environments
+#define VCMI_WINDOWS
```
kambala-decapitator, line 34: some indentation is definitely needed to make it more readable

kambala-decapitator (commented):
thanks a lot for moving it forward!

> I think better option would be to format files over times as they are changed instead of formatting entire codebase at once, breaking all WIP branches.

if there're wip branches, then yeah, better not to format all at once. For the former a nice solution is using `git-clang-format` that formats only changed lines.

lib/GameConstants.cpp
IvanSavenko: Yeah. Somehow fixed itself after second pass...

lib/GameConstants.cpp
IvanSavenko: I don't mind about this particular case, but it looks like same option also applies to all braced lists. From docs:
Cpp11BracedListStyle 
```
true:                                  false:
vector<int> x{1, 2, 3, 4};     vs.     vector<int> x{ 1, 2, 3, 4 };
vector<T> x{{}, {}, {}, {}};           vector<T> x{ {}, {}, {}, {} };
f(MyMap[{composite, key}]);            f(MyMap[{ composite, key }]);
new int[3]{1, 2, 3};                   new int[3]{ 1, 2, 3 };
```
Personally, for actual lists with multiple items I'd rather keep spaces

client/gui/CAnimation.h
IvanSavenko: Personally I prefer it but clang-format is just *too* agressive with it.
Problems is, same option is also used for assignments alignment, e.g:
```
	delta             = animDelta;
	fadingMode        = mode;
	fadingCounter     = initialCounter();
	fading            = true;
	shouldFreeSurface = freeSurfaceAtEnd;
```
Where (IMO) vertical alignment is justified.
Even in methods, I don't think anything wrong with aligning int/bool/floats (1-2 extra spaces).
Not 10 spaces like in this example obviously.

In that case I'll bring this up for discussion. My proposition:
- enable this option by default
- whenever clang-format gets too agressive - add empty line between long and short lines. Example: this file in next commit (will push shortly)

client/gui/CAnimation.h
IvanSavenko: I'd also add them but apparently this thing only works for JavaScript...
At least it seems that it will preserve any pre-existing trailing commas and won't attempt to remove them

client/gui/CAnimation.h
```diff
@@ -9,16 +9,16 @@
  */
 #pragma once
 
+#include "../../lib/GameConstants.h"
 #include "../../lib/vcmi_endian.h"
 #include "Geometries.h"
-#include "../../lib/GameConstants.h"
```
Nordsoft91, line 14: is it also something format-related?

client/gui/CAnimation.h
```diff
@@ -39,113 +39,114 @@ class IImage
...
 	int width() const;
 	int height() const;
 
-	//only indexed bitmaps, 16 colors maximum
-	virtual void shiftPalette(int from, int howMany) = 0;
+	// only indexed bitmaps, 16 colors maximum
+	virtual void shiftPalette(int from, int howMany)         = 0;
 	virtual void adjustPalette(const ColorShifter * shifter) = 0;
-	virtual void resetPalette() = 0;
+	virtual void resetPalette()                              = 0;
```
Nordsoft91, line 67: this one I don't like

client/gui/CAnimation.h
```diff
@@ -161,17 +162,22 @@ class CFadeAnimation
 public:
 	enum class EMode
 	{
-		NONE, IN, OUT
+		NONE,
+		IN,
+		OUT,
```
Nordsoft91, line 167: enumerating is single line is not allowed?

lib/GameConstants.h
Nordsoft91: initial looks better

Nordsoft91 (commented):
I suggest do not apply so many rules. Let's select minimal set of rules related to spaces but not restrict ourselves a lot.

client/gui/CAnimation.h
IvanSavenko: What exactly? Header reordering?
Yes, clang-format can rearrange includes. Not even sure whether I can disable it completely, but current rules are:
- StdInc.h always comes first
- Main include (include of header with same name as cpp name) always comes after StdInc.h)
- After this, clang-format will detect "groups" - continuous includes without empty lines and sort them alphabetically
- All group will be preserved - clang will not try to rearrange them

lib/GameConstants.h
IvanSavenko: Haven't noticed that it turned entire method into one-liner. Will look into it.

client/gui/CAnimation.h
IvanSavenko: Currently - no. Do we need it?

lib/GameConstants.h
dydzio0614: Not sure what to think, personally I am used to c# language where usually everything sits in namespace and namespaced stuff have default intendation

Global.h
```diff
@@ -302,462 +311,456 @@ void inline handleException()
...
-		seed ^= hasher(v) + 0x9e3779b9 + (seed<<6) + (seed>>2);
-	}
-
-	//returns true if container c contains item i
-	template <typename Container, typename Item>
-	bool contains(const Container & c, const Item &i)
-	{
-		return std::find(std::begin(c), std::end(c),i) != std::end(c);
-	}
+// returns true if container c contains item i
```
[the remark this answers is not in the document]
IvanSavenko: Looks like it. Should we remove this rule? Or remove the space?

dydzio0614:
BTW did you take https://wiki.vcmi.eu/Coding_guidelines into account for this preview?

IvanSavenko:
> BTW did you take https://wiki.vcmi.eu/Coding_guidelines into account for this preview?

Huh, technically - no. Only from looking on changes & from my memory. But don't see any mismatches.
However clang-format settings are much more extensive than coding guidelines. A lot of clang settings are simply not defined in coding guidelines

client/gui/CAnimation.h
kambala-decapitator: > Where (IMO) vertical alignment is justified.

there's a simple "counter-example": consider you add one more variable there that becomes the longest one, will you re-format all the other variables just because of that?

> whenever clang-format gets too agressive - add empty line between long and short lines. Example: this file in next commit (will push shortly)

this would work I think

lib/GameConstants.cpp
kambala-decapitator: yes, for multiple items that's fine

Global.h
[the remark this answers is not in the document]
Nordsoft91: I would prefer to leave it up to developer

client/gui/CAnimation.h
Nordsoft91: "And sort them alphabetically" - is it possible to remove this part?
Actually it's bug-prone, as you may include header which has include inside.

client/gui/CAnimation.h
Nordsoft91: Why not allowing? Take a look at ProtocolConsts from lobby.h
It would be awful.
Also, what is practical benefit from it?

client/gui/CAnimation.h
IvanSavenko: Ah, it looks that I can I think.

And how it is bug-prone? The only exception is StdInc.h since it includes commonly used headers.
Other than that - order of headers does not matters (assuming well-written code)

client/gui/CAnimation.h
Nordsoft91: If we do have different personal preferences, let's not apply rule and rely on developer. For instance I personally don't use such alignments, because sometimes I use text search like 'var = value' to find where variable gets assigned

client/gui/CAnimation.h
IvanSavenko: What's wrong with ProtocolConsts? What's so special about it?

Also, what is practical benefit of code formatting?

lib/GameConstants.cpp
Nordsoft91: For me left part looks better

Nordsoft91:
Can we make strict rules only for those points?
- tabs instead of spaces
- spaces around binary operators
- no space after if/for/while
And maybe something about class/methods/variables naming

client/gui/CAnimation.h
Nordsoft91: You may have implicit declaration of class in header, included into other header. It means, that you won't have comprehensive list of headers.

IvanSavenko: Then this is error in code. Header must have complete set of all required forward declarations.

Nordsoft91: This error is hidden, because you will not have any warning or compilation error. "Classic" example
```
#include <vector>
#include "header.h" //vector is not included but used
```
vs
```
#include "header.h" //compilation error
#include <vector>
```

IvanSavenko: Again, this is an error in header.h. 
header must include all its dependencies. If not - this is a bug.

client/gui/CAnimation.h
Nordsoft91: Practical benefit from code formatting is to simplify code reading. 
Simplicity of code reading slightly correlates with amount of text on developer's screen. Long lines - bad, many lines - also bad.
Thus, should be balance between line width and amount of lines but modern IDEs do text wrapping, but they do nothing with long files. IMO, there is nothing wrong in having NONE, INE, OUT in one line

client/gui/CAnimation.h
kambala-decapitator: I like sorted includes

> header must include (or forward-declare if applicable) all its dependencies. If not - this is a bug.

true

client/gui/CAnimation.h
kambala-decapitator: don't see any sense in having enum values on one line

IvanSavenko:
> I suggest do not apply so many rules. Let's select minimal set of rules related to spaces but not restrict ourselves a lot.

It looks like this is not how clang-format works...
Most options are booleans on/off, but this is usually switch between two formatting styles.
Removing an option does not stops formatting, but instead applies default setting.

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -27,112 +27,114 @@ namespace spells
...
-		case 3: //right bottom
-			return std::make_pair((y%2) ? x : x+1, y+1);
-		case 4: //left bottom
-			return std::make_pair((y%2) ? x-1 : x, y+1);
-		case 5: //left
-			return std::make_pair(x-1, y);
-		default:
-			throw std::runtime_error("Disaster: wrong direction in SRSLPraserHelpers::gotoDir!\n");
-		}
+	case 0: //top left
```
dydzio0614, line 62: Wouldn't it be better if case was indented 1 level deeper than opening brace?

IvanSavenko:
Frankly, the more I look into this the less I believe that we should not use it. 
This is attempt to create formatting description that will be suitable for *every* possible situation. Which is... not something that can be done with such strict definitions.

For example, indenting namespace makes sense if you define 1 function or use 1 forward declaration... but only adds noise to vertical width when used on whole file. But clang-format has no option to ignore indentation on namespaces

Vertical alignment helps with readability if you have like 1-2 symbols difference between line legths. But only harms readability if you attempt to align 20+ symbol long variable with 3-symbol long variable. But clang-format has neither heuristics for this, nor any fine-tuning options, nor possibility to ignore such alignment.

This list can be continued for quite a while.

Another problem is clang-format uses "destroy everything and rebuilt it from scratch" approach. I thought about dropping all rules and then slowly building them from scratch, e.g. first fix tabulation until I am happy with it, then fix alignment, then go onto next part...

But this is simply not something that is supported by clang-format. It is all or nothing.

In my opinion, it may be better to attempt to revive #347 
No idea about how Uncrustify does things, but at the very least it will be different from clang-format

client/gui/CAnimation.h
nullkiller: But JSON parser of Java does not like it)

lib/GameConstants.cpp
nullkiller: I would better use vector x = {1, 2, 3, 4}

lib/GameConstants.h
nullkiller: Also I am against heavily indented things)

lib/GameConstants.h
nullkiller: Yes I am. Still I am ok if there will be no indentation. Also afaik latest C# introduced a lot of alternative constructs to reduce indentations as well. Like you just write namespace Lalala; and that is all

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -27,112 +27,114 @@ namespace spells
...
+}
+}
 
-BattleSpellMechanics::BattleSpellMechanics(const IBattleCast * event, std::shared_ptr<effects::Effects> effects_, std::shared_ptr<IReceptiveCheck> targetCondition_)
-	: BaseMechanics(event),
-	effects(effects_),
-	targetCondition(targetCondition_)
-{}
+BattleSpellMechanics::BattleSpellMechanics(const IBattleCast * event,
+										   std::shared_ptr<effects::Effects> effects_,
```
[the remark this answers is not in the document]
nullkiller: it aligned code to opening brace

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -389,12 +392,13 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
...
 	for(auto & p : effectsToApply)
 	{
-		vstd::erase_if(p.second, [&](const Destination & d)
-		{
-			if(!d.unitValue)
-				return false;
-			return vstd::contains(resisted, d.unitValue);
-		});
+		vstd::erase_if(p.second,
+					   [&](const Destination & d)
```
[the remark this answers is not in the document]
nullkiller: +1

client/gui/CAnimation.h
kambala-decapitator: how is C++ header related to Java or JSON?

rilian-la-te:
@IvanSavenko we can just add clang-format file and write than before commit you should run `git clang-format --staged` and do not reformat alredy commited code. What do you think?

kambala-decapitator:
it's much easier to run `git-clang-format` command :)

rilian-la-te:
@kambala-decapitator for new commits - yes) I think we should include sample clang-format file in repo, but do not reformat all code. What do you think?

kambala-decapitator:
> @kambala-decapitator for new commits - yes) I think we should include sample clang-format file in repo, but do not reformat all code. What do you think?

that was my original intention actually - to format only touched lines.

one-time full codebase formatting is also possible, then we should also start a file with a list of commits that should be skipped when executing `git blame`.

IvanSavenko:
First of all it is not finished - check unresolved conversations. At this point I simply gave up on trying to find configuration that will be suitable for our code and at least *reasonably* readable. Its "rebuild from scratch" approach is really annoying IMO.

What I don't like about clang-format is that it is too unconditional.
For example, you can't setup rule like "split lines longer than X symbols, but don't concatenate lines shorted than Y".
This leads to (IMO) bad formatting when function call/conditional that is more readable in multi-line form gets converted into a single line and vice versa. I do get (and agree with) idea of "too long lines are unreadable", but I really don't like enforced rule of "all expressions shorter than X symbols must be in one line"

Similarly, IMO for long math expressions grouping by priority is more readable, e.g. ` x*x + y*y` - no spaces between multiplication, spaces between additions. But there is no way to either setup clang-format for such format or force it to keep existing format.

There were other cases that I don't remember at the moment, but you get an idea.

If you wish to use it - feel free to do so, but IMO before committing clang-format output needs another manual readability check.

PR #1277, opened 2022-12-23
Implemented cut copy paste for editor

Nordsoft91:


mapeditor/mapcontroller.cpp
```diff
@@ -323,6 +324,48 @@ void MapController::commitObjectErase(int level)
...
+
+void MapController::pasteFromClipboard(int level)
+{
+	_scenes[level]->selectionObjectsView.clear();
+	
+	auto shift = int3::getDirs()[_clipboardShiftIndex++];
+	if(_clipboardShiftIndex == int3::getDirs().size())
+		_clipboardShiftIndex = 0;
+	
+	for(auto & objuptr : _clipboard)
```
IvanSavenko, line 347: `objuptr` is a strange name for a variable. We also usually use camelCase, so should be `objPtr`. And what "u" in variable name stands for?

mapeditor/mapcontroller.cpp
```diff
@@ -323,6 +324,48 @@ void MapController::commitObjectErase(int level)
...
+	{
+		auto * obj = CMemorySerializer::deepCopy(*objuptr).release();
+		auto newpos = objuptr->pos + shift;
```
IvanSavenko, line 350: Also not in camelCase

mapeditor/mainwindow.h
```diff
@@ -98,6 +98,12 @@ private slots:
 	
 	void switchDefaultPlayer(const PlayerColor &);
 
+	void on_actionCut_triggered();
```
dydzio0614, line 101: is editor supposed to follow https://wiki.vcmi.eu/Coding_guidelines#Casing ?

mapeditor/mapcontroller.cpp
```diff
@@ -323,6 +324,48 @@ void MapController::commitObjectErase(int level)
 	main->mapChanged();
 }
 
+void MapController::copyToClipboard(int level)
+{
+	_clipboard.clear();
+	_clipboardShiftIndex = 0;
+	auto selectedObjects = _scenes[level]->selectionObjectsView.getSelection();
+	for(auto * obj : selectedObjects)
```
dydzio0614, line 332: Can obj be null? If not, maybe use &

mapeditor/mapcontroller.h
```diff
@@ -64,6 +67,8 @@ class MapController
 	MainWindow * main;
 	mutable std::array<std::unique_ptr<MapScene>, 2> _scenes;
 	mutable std::array<std::unique_ptr<MinimapScene>, 2> _miniscenes;
+	std::vector<std::unique_ptr<CGObjectInstance>> _clipboard;
```
IvanSavenko, line 70: Huh, why is editor uses different naming style? I see that some of these were before, but we generally don't use underscore for class members.

mapeditor/mapview.cpp
```diff
@@ -270,6 +270,7 @@ void MapView::mousePressEvent(QMouseEvent *event)
 			if(event->button() == Qt::LeftButton)
 			{
 				auto * obj = sc->selectionObjectsView.selectObjectAt(tileStart.x, tileStart.y);
+				auto * obj2 = sc->selectionObjectsView.selectObjectAt(tileStart.x, tileStart.y, obj);
```
dydzio0614, line 273: consider better variable name than obj2 to reflect usage context

IvanSavenko, line 273: I may be missing context, but what is obj & obj2? 
What if we have more than 2 objects on a tile?
I think this place deserves a comment to explain why such logic is needed

mapeditor/mainwindow.h
Nordsoft91: Those methods are automatically generated by Qt. Having exception will be easier than changing

mapeditor/mapcontroller.cpp
Nordsoft91: Since it's container of pointers, auto & will be translated into auto &*. There is no way to store references in containers in C++ (there is reference_wrapper which has underlying pointer as well)

mapeditor/mapview.cpp
Nordsoft91: Hm, but this is literally "object two" - second object selected by user. You can have overlapped objects on map

mapeditor/mapcontroller.cpp
Nordsoft91: Unique ptr

dydzio0614: objuptr -> objUniquePtr

mapeditor/mapcontroller.h
Nordsoft91: It's more pythonic way to distinguish private members

mapeditor/mapcontroller.cpp
IvanSavenko: Is that fact important here? You don't do anything specifically different on basis that this is unique ptr.
If you were to change unique_ptr to shared_ptr this loop won't change. So no need to specify this here.
And even if decide to keep it - then what @dydzio0614 said.

mapeditor/mapcontroller.h
IvanSavenko: ...but this is c++, vcmi. We use different style.

Nordsoft91: Qt, fl-static for example use same approach, so it's quite common naming in c++. If you think it's something really important, I can change, but better in another PR to avoid quite large irrelevant change

IvanSavenko: I'm not saying that this is bad style, but this is not style we use in all other projects, including Qt-based Launcher and FL-based AI's

I'll agree with "outside of PR scope" argument, but in long-term I think that it should be switched to one we use everywhere else.

mapeditor/mainwindow.h
kambala-decapitator: FYI you can always connect signals to slots manually, without using the Designer

mapeditor/mapview.cpp
kambala-decapitator: exactly my thoughts

kambala-decapitator (approved):
don't see any serious issues here

IvanSavenko (approved):

mapeditor/mainwindow.h
IvanSavenko: Manually - as with `connect` method or what's its name?
Yes, but why write boilerplate code that adds nothing to functionality when this can be done automatically behind the scenes?

kambala-decapitator: yes, connect. You can also connect to lambdas.

not saying it's a must :)

PR #1283, opened 2022-12-24
Update README.md

DjWarmonger:
Update download counter


dydzio0614:
I would also add total download counter - https://img.shields.io/github/downloads/vcmi/vcmi/total

DjWarmonger:
> I would also add total download counter - https://img.shields.io/github/downloads/vcmi/vcmi/total

Done.

Issue #1286, opened 2022-12-24
Android VCMI 1.1.0: Can't start new game


lje:
I just updated VCMI Android from 1.0.0 to 1.1.0, when trying to start a new single player game, nothing happens after chosing map and then pressing "Begin". The "Begin" button gets grayed out, but nothing more happens.

Android 13 on Samsung Galaxy Tab S8+

musshorn:
I'm experiencing this too. OnePlus 7 running android 11. 

Looking at the end of the server log, it might be trying to initiate some sort of empty object?

```TRACE global [6d70932cc0] - Calling Init for object 2018, shrub, object
TRACE global [6d70932cc0] - Calling Init for object 2019, rock, object
TRACE global [6d70932cc0] - Calling Init for object 2020, shrub, object
TRACE global [6d70932cc0] - Calling Init for object 2021, shrub, object
TRACE global [6d70932cc0] - Calling Init for object 2022, , 
TRACE global [6d70932cc0] - Calling Ini
```
Is the end of my server log file.

nullkiller:
I will make some tests. We did not test android much.

nullkiller:
My android does not start higher than 1024*600. It worked in 1.0 so maybe some games with SDL resolution broke it

kambala-decapitator:
installed 1.1 on top of w/e post-1.0 build I had, SP scenario works fine in 1080p. Xiaomi Redmi Note 9S running android 11.

nullkiller:
I reproduced on my new samsung. It is very long. So have some fix already.

nullkiller:
Not exactly same case so. Btw how long have you waited?

nullkiller:
I see now. XL map does not want to load.

kambala-decapitator:
> I see now. XL map does not want to load.

can't confirm, on my device it loaded after about a minute. Tested oh3 map.

musshorn:
Ok I can confirm it does load, even the XL maps after about a minute or so. I think it's just because it doesn't cut to the load screen when you press begin it looked bugged for me, even though it's loading in the background

nullkiller:
About a minute. Not acceptable. I said it is slow! I believe it is map passing over network

IvanSavenko:
Already fixed, probably even in 1.1.1

PR #1292, opened 2022-12-25
#1286 - use iOS flow for Android resolutions management

nullkiller:


nullkiller:
For unsupported resolutions it use downscale in this case

kambala-decapitator (approved):

PR #1293, opened 2022-12-25
Translations support

IvanSavenko:
Currently very WIP, 

**Functionality**:
- [x] Implement translations for Launcher using Qt translator
- [ ] Implement translations for Map Editor using Qt translator
- [x] Allow mods to define mod information (e.g. name/description) with localization to current language of Launcher
- [ ] Provide mechanism to translate data from H3 files via mods that is better than direct editing of H3 files
- [ ] (optional/needs research) Provide mechanism to translate .h3m and .h3c files.
- [ ] Integrate existing translate.json into H3 translation support
- [ ] Provide mechanism for mods to translate new content
- [ ] Multiplayer support - ensure that client & server can run with different languages

**Languages**:
- [x] English (already available)
- [x] Ukrainian (done by me)
- [ ] Polish
- [ ] Russian
- [ ] Any other languages?

(if there is anyone willing to provide missing/new translation ask me here or on Slack)

**Other changes**:
- Changed UI of the left panel to a different set of components. Haven't found a way to make old one scale correctly based on length of translated text/height of the Launcher
- Added button to start map editor from Launcher (TODO: implement actual click event)
- Added automatic generation of Qt translations during build (TODO: automatic copy of resulting files)
- Merged "fullscreen" and "true fullscreen" GUI options into single one. Internaly in persistent settings storage they will stay separate
- Added "language" field to persistent settings storage. (TODO: add to schema)


Laserlicht:
Here would be a german translation, which could be added (see attachment):

`"de", // german`

`	"translation_de" : {`
`		"name" : "VCMI - grundlegende Dateien",`
`		"description" : "Grundlegende Dateien, die für die korrekte Ausführung von VCMI erforderlich sind",`
`		"author" : "VCMI-Team",`
`		"modType" : "Grafik",`
`	},`

[launcher_de.zip](https://github.com/vcmi/vcmi/files/10300426/launcher_de.zip)

IvanSavenko:
Thanks, I'll add it once I'm on this branch. 
Oh, and there are few more strings that we have here:
https://github.com/IvanSavenko/vcmi/blob/translate_launcher/config/translate.json
These were added long ago and may already be translated somewhere in some mod. Feel free to translate it when you have time.

Laserlicht:
Thank you! 🙂

The other strings are already translated by this mod: https://github.com/vcmi-mods/german-translation/blob/master/content/config/translate.json

IvanSavenko:
Marking as ready for review.
Other parts, e.g. Map Editor will be done separately. 

I probably still need to recheck CI & update its settings to install qt localization tools, but everything else should be ready.
Any unfinished translations can be added/updated at any point later, potentially as separate PR's

IvanSavenko:
Fix this:
зображення

dydzio0614:
[translate.json.txt](https://github.com/vcmi/vcmi/files/10309824/translate.json.txt)
[launcher_pl.ts.txt](https://github.com/vcmi/vcmi/files/10309825/launcher_pl.ts.txt)
Added redundant txt edition to make github accept files

config/schemas/settings.json
```diff
@@ -51,6 +51,10 @@
 					"type":"string",
 					"default" : "Maps/Arrogance"
 				},
+				"language" : {
+					"type":"string",
+					"default" : "english"
```
Nordsoft91, line 56: add enum to support only limited set of languages

launcher/lobby/lobbyroomrequest_moc.cpp
```diff
@@ -27,6 +27,14 @@ LobbyRoomRequest::LobbyRoomRequest(SocketLobby & socket, const QString & room, c
 	show();
 }
 
+void LobbyRoomRequest::changeEvent(QEvent *event)
+{
+	if ( event->type() == QEvent::LanguageChange)
+	{
+		ui->retranslateUi(this);
+	}
```
Nordsoft91, line 35:
```suggestion
	}
	QDialog::changeEvent(event);
```

launcher/lobby/lobby_moc.cpp
```diff
@@ -35,6 +35,14 @@ Lobby::Lobby(QWidget *parent) :
 	ui->kickButton->setVisible(false);
 }
 
+void Lobby::changeEvent(QEvent *event)
+{
+	if ( event->type() == QEvent::LanguageChange)
+	{
+		ui->retranslateUi(this);
+	}
```
Nordsoft91, line 43:
```suggestion
	}
	QWidget::changeEvent(event);
```

launcher/mainwindow_moc.cpp
```diff
@@ -68,45 +69,29 @@ MainWindow::MainWindow(QWidget * parent)
...
 	if(settings["launcher"]["updateOnStartup"].Bool())
 		UpdateDialog::showUpdateDialog(false);
 }
 
+void MainWindow::changeEvent(QEvent *event)
+{
+	if ( event->type() == QEvent::LanguageChange)
+	{
+		ui->retranslateUi(this);
+	}
```
Nordsoft91, line 92:
```suggestion
	}
	QMainWindow::changeEvent(event);
```

launcher/modManager/cmodlist.cpp
```diff
@@ -157,23 +158,34 @@ QString CModEntry::getName() const
...
-		else
-			return localData[value];
+		useRepositoryData = compareVersions(installedVer, availableVer);
 	}
 
-	if(repository.contains(value))
-		return repository[value];
+	auto & storage = useRepositoryData ? repository : localData;
+
+	if (storage.contains(langValue))
```
Nordsoft91, line 180: no space after if here and below

launcher/modManager/cmodlistmodel_moc.cpp
```diff
@@ -155,13 +145,25 @@ Qt::ItemFlags CModListModel::flags(const QModelIndex &) const
...
+
 	if(role == Qt::DisplayRole && orientation == Qt::Horizontal)
-		return ModFields::header[section];
+		return QCoreApplication::translate("ModFields", header[section].toStdString().c_str());
 	return QVariant();
 }
 
 void CModListModel::reloadRepositories()
 {
+	//emit headerDataChanged(Qt::Horizontal, 0, -1 );
```
Nordsoft91, line 166: remove?

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -72,12 +84,10 @@ void CSettingsView::loadSettings()
...
 #ifdef Q_OS_IOS
 	ui->comboBoxFullScreen->setCurrentIndex(true);
-	ui->checkBoxFullScreen->setChecked(false);
-	for (auto widget : std::initializer_list<QWidget *>{ui->comboBoxFullScreen, ui->checkBoxFullScreen})
-		widget->setDisabled(true);
+	ui->comboBoxFullScreen->setDisabled(true);
 #else
 	ui->comboBoxFullScreen->setCurrentIndex(settings["video"]["fullscreen"].Bool());
-	ui->checkBoxFullScreen->setChecked(settings["video"]["realFullscreen"].Bool());
+	//ui->checkBoxFullScreen->setChecked(settings["video"]["realFullscreen"].Bool());
```
Nordsoft91, line 90: remove?

Nordsoft91 (commented):
Reviewed 50%, will test as well on MacOS and iOS later

launcher/mainwindow_moc.ui
```diff
@@ -198,6 +168,153 @@
...
+       </size>
+      </property>
+      <property name="font">
+       <font>
+        <weight>75</weight>
+        <bold>true</bold>
+       </font>
+      </property>
+      <property name="text">
+       <string>Editor</string>
```
Nordsoft91, line 298:
```suggestion
       <string>Map Editor</string>
```

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -192,14 +207,10 @@ void CSettingsView::on_comboBoxResolution_currentTextChanged(const QString & arg
...
-}
-
-void CSettingsView::on_checkBoxFullScreen_stateChanged(int state)
-{
-	Settings node = settings.write["video"]["realFullscreen"];
-	node->Bool() = state;
+	Settings nodeFullscreen     = settings.write["video"]["fullscreen"];
+	Settings nodeRealFullscreen = settings.write["video"]["realFullscreen"];
+	nodeFullscreen->Bool() = index != 0;
+	nodeFullscreen->Bool() = index == 2;
```
Nordsoft91, line 213:
```suggestion
	nodeFullscreen->Bool() = (index != 0);
	nodeFullscreen->Bool() = (index == 2);
```

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -307,3 +318,20 @@ void CSettingsView::on_updatesButton_clicked()
...
+	if ( qApp->activeWindow() && dynamic_cast<MainWindow*>(qApp->activeWindow()) )
+		dynamic_cast<MainWindow*>(qApp->activeWindow())->updateTranslation();
+}
+
+void CSettingsView::changeEvent(QEvent *event)
+{
+	if ( event->type() == QEvent::LanguageChange)
+	{
+		ui->retranslateUi(this);
+	}
```
Nordsoft91, line 336:
```suggestion
	}
	QWidget::changeEvent(event);
```

Nordsoft91 (changes requested):
1) Translations doesn't work for settings tab and lobby tab
2) This vertical line looks ugly (macOS) - but this is my subjective opinion
Снимок экрана 2022-12-28 в 16 20 42
3) on iOS looks also bad. Assuming that there is no scrollbar if buttons aren't fit screen size for smaller devices
IMG_0788

launcher/mainwindow_moc.cpp
```diff
@@ -68,45 +69,29 @@ MainWindow::MainWindow(QWidget * parent)
 		move(position);
 	}
 
-	//set default margins
+#ifdef Q_OS_IOS
```
Nordsoft91, line 72: instead of checking for this flag, will it be possible to pass a parameter ENABLE_EDITOR from cmake?

launcher/modManager/imageviewer_moc.cpp
```diff
@@ -20,6 +20,14 @@ ImageViewer::ImageViewer(QWidget * parent)
 	ui->setupUi(this);
 }
 
+void ImageViewer::changeEvent(QEvent *event)
+{
+	if ( event->type() == QEvent::LanguageChange)
+	{
+		ui->retranslateUi(this);
+	}
```
Nordsoft91, line 28:
```suggestion
	}
	QWidget::changeEvent(event);
```

launcher/settingsView/csettingsview_moc.ui
```diff
@@ -558,31 +514,122 @@
...
+     </item>
+     <item>
+      <property name="text">
+       <string>Русский (Russian)</string>
+      </property>
+     </item>
+     <item>
+      <property name="text">
+       <string>Українська (Ukrainian)</string>
+      </property>
```
Nordsoft91, line 542: I see that languages are met in several places. I suggest to add items into this combobox in constructor instead of having them statically

lib/VCMIDirs.cpp
```diff
@@ -348,6 +349,7 @@ std::vector<bfs::path> VCMIDirsWIN32::dataPaths() const
 }
 
 bfs::path VCMIDirsWIN32::clientPath() const { return binaryPath() / "VCMI_client.exe"; }
+bfs::path VCMIDirsWIN32::editorPath() const { return binaryPath() / "VCMI_editor.exe"; }
```
Nordsoft91, line 352: There will be template editor and campaign editor at some point
```suggestion
bfs::path VCMIDirsWIN32::mapEditorPath() const { return binaryPath() / "VCMI_editor.exe"; }
```

Nordsoft91, line 352: There will be template editor and campaign editor at some point
```suggestion
bfs::path VCMIDirsWIN32::mapEditorPath() const { return binaryPath() / "VCMI_editor.exe"; }
```

IvanSavenko: Will change, although I think that it would be better to have such tools integrated into our map editor.

Nordsoft91: Maybe a good option instead of making many apps. So agree now with your option

IvanSavenko: Well, for now this thing is still only *map* editor, so renames (both here and UI button) make sense. Can be reverted later after editor upgrade

launcher/settingsView/csettingsview_moc.ui
IvanSavenko: This should be the only place with such form, as in - human-readable names of languages to select from.
Won't be possible to define languages in a single place though - mod schema, settings schema, translation files. And this is only what was added so far - there may be others in future (although I'll obviously would try to avoid duplication).

That being said, I do have plans for a second place to select language - some sort of a "welcome screen" for initial setup. So your suggestion makes sense...

IvanSavenko:
> 1. Translations doesn't work for settings tab and lobby tab

Hmm... Unable to reproduce. Works for me. Switching to German, Polish or Ukrainian instantly translates all tabs.

> 2. This vertical line looks ugly (macOS) - but this is my subjective opinion

That's not how it looks for me. Investigating other approaches.

> 3) **on iOS looks bad**. Assuming that there is no scrollbar if buttons aren't fit screen size for smaller devices

Looking into it.

IvanSavenko:
Addressed some of review comments, will fix remaining once a bit later.

IvanSavenko:
@Nordsoft91 can you check how it looks on iphone now?
Hard to fix something when you're doing in blindly.

launcher/settingsView/csettingsview_moc.ui
IvanSavenko: For now - keeping as it.
I agree that this is valid suggestion and will likely implement it a bit later, e.g. when working on "Welcome" window. Not yet sure where and how we'll need list this of languages, so keeping it as it until we do need it.
Mostly to avoid doing job twice if it turns out that I needed different approach here.

Nordsoft91:
Video of non-working translations on macOS. Not sure that I like those extended icons in terms of design also
https://user-images.githubusercontent.com/9308612/209881813-6ca58c90-43a7-4829-a88f-60ba3f25e962.mp4

Nordsoft91:
Icons on iOS look good. Translations don't work too, but i noticed that essentials mod name was translated (on macOS too)
IMG_74E005EC79AF-1

Nordsoft91:
Have a gut feeling that after adding map editor for iOS, it might be broken again

IvanSavenko:
Thanks for testing.
Yeah, not happy with stretchy buttons either, was mostly trying to fix that iphone bug. 

Missing translation - probably faulty search paths for translations, will fix

Nordsoft91:
I wouldn't call it a bug. You cannot fit small screen everything (one more example https://github.com/vcmi/vcmi/pull/1227)
Maybe having scrolled menu with items is not something we have to change.

IvanSavenko:
> Have a gut-feeling that after adding map editor for iOS, it might be broken again

I guess this will be yours problem then? 
Seriously though - yeah, very likely. Potential solutions:
1) Change icon size property of a buttons based on screen height. Can't experiment with this much without device to test on.
2) For all phones (including potential Android support) use different style for buttons - one that shows icons in the same row as text: `toolButtonStyle == ToolButtonTextBesideIcon`. Desktops have a lot of vertical space, but for mobiles with extremely wide screens this may be a better option.
3) Don't show icon. Not the best option, but definitely possible.

IvanSavenko:
(or option 4 - move these buttons into a widget with scroll bar)

IvanSavenko:
Search paths should be fixed, translations should work now.

@Nordsoft91 When you're available, please check the following list:
- macOS & ios - translations should work now.
- macOS - stretchy icons should be very limited (if you still feels them a bit too stretchy - let me know)
- ios - check whether icons still appear correctly, with no clipping text
- ios - not related to PR, but can you make me a screenshot of mods tab with open side panel (one with mod details)?

CI/mxe/before_install.sh
```diff
@@ -1,12 +1,21 @@
...
+
 # Install nsis for installer creation
 sudo add-apt-repository 'deb http://security.ubuntu.com/ubuntu bionic-security main'
 sudo apt-get install -qq nsis ninja-build libssl1.0.0
 
 # MXE repository was too slow for Travis far too often
-wget -nv https://github.com/vcmi/vcmi-deps-mxe/releases/download/2021-02-20/mxe-i686-w64-mingw32.shared-2021-01-22.tar
-tar -xvf mxe-i686-w64-mingw32.shared-2021-01-22.tar
+wget -nv https://github.com/vcmi/vcmi-deps-mxe/releases/download/2022-12-26/mxe-i686-w64-mingw32.shared-2022-12-26.tar
+tar -xvf mxe-i686-w64-mingw32.shared-2022-12-26.tar
```
kambala-decapitator, line 18: I think MXE upgrade should be a separate PR

config/schemas/mod.json
```diff
@@ -78,6 +112,26 @@
 			"type":"boolean",
 			"description": "If set to true, mod will not be enabled automatically on install"
 		},
+		
+		"english" : {
+			"$ref" : "#/definitions/localizeable"
```
kambala-decapitator, line 117: in other places as well
```suggestion
			"$ref" : "#/definitions/localizable"
```

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -307,3 +317,21 @@ void CSettingsView::on_updatesButton_clicked()
...
 }
 
+
+void CSettingsView::on_comboBoxLanguage_currentIndexChanged(int index)
+{
+	Settings node = settings.write["general"]["language"];
+	node->String() = languageTagList[index];
+
+	if ( qApp->activeWindow() && dynamic_cast<MainWindow*>(qApp->activeWindow()) )
+		dynamic_cast<MainWindow*>(qApp->activeWindow())->updateTranslation();
```
kambala-decapitator, line 327:
```suggestion
	if(auto activeWindow = qApp->activeWindow())
		if(auto mainWindow = dynamic_cast<MainWindow *>(activeWindow))
			mainWindow->updateTranslation();
```

CI/mxe/before_install.sh
IvanSavenko: Maybe. But:
1) Changes in MXE packages are really minor, due to their "as stable as fossils" approach. Only several packages got upgrade to newer "patch" version. I think CMake is the only exception, which went from 3.10 to 3.19
2) Since we use MXE as primary Windows build, this PR is largely pointless if MXE and as result - Windows has translations disabled
3) Since we use MXE prebuilt packages repository, I didn't have an option of only adding new package(s) without upgrading everything else.

kambala-decapitator: Toolchain upgrade itself has no relation to translations, that's why I view it as a completely separate change. After that your can simply rebase your translation branches on top of that.
Point 3 I didn't understand tbh.

IvanSavenko: Regarding 3:
Normally, I would have simply added new dependency to the list.
This is what I did with Ubuntu-based CI's
With our MXE setup this is impossible - if you want to add even a single package, you have to upgrade entire dependencies package.

So yeah, I'd also prefer to only add linguist tools to MXE without full upgrade (especially considering how useless it is), but we don't have such option. It's all or nothing.

kambala-decapitator: clear, thanks. That's totally fine.

launcher/settingsView/csettingsview_moc.cpp
kambala-decapitator: I think you can even merge both `if`s: dynamic_cast of nullptr should be fine

IvanSavenko:
@Nordsoft91 any ETA? If not then I am merging as it and if there are still any issues on mac/ios - will fix them later.

kambala-decapitator:
> @Nordsoft91 any ETA? If not then I am merging as it and if there are still any issues on mac/ios - will fix them later.

@IvanSavenko I can check tomorrow

kambala-decapitator:
Screenshot 2023-01-15 at 16 21 50

- icons: the ones on the left look good to me
- translations: nothing happens after changing language / after restarting the launcher

kambala-decapitator:
iOS: same for icons and translations.

dunno if related to the PR, but game crashed on clicking "new scenario", updating VCMI extras didn't solve it (only this mod is installed). [Logs](https://github.com/vcmi/vcmi/files/10420288/Archive.tar.gz)

Screen Shot 2023-01-15 at 16 27 35
Screen Shot 2023-01-15 at 16 28 28

kambala-decapitator:
I guess this explains why translations don't work:

```
> find /Volumes/VCMI/VCMI.app -iname '*.qm'
>
```

they're simply missing from app bundle. (same for iOS)

btw map editor puts source translation files (`.ts`) into app bundle which is useless.

IvanSavenko:
@kambala-decapitator can you suggest a fix? Not familiar with mac file layouts.
I've attempted to copy icons handling but apparently it didn't work.
All relevant code should be in launcher/CMakeLists.txt

kambala-decapitator:
@IvanSavenko please try https://github.com/kambala-decapitator/vcmi/commit/387bff75f0ca6859dc21cad39c1acaa7975bec56

kambala-decapitator:
just built locally: seems my commit is wrong. Need to dive into cmake+Qt translations, never used it before.

IvanSavenko:
@kambala-decapitator 
check `MainWindow::updateTranslation`
All c++ code relevant for translations loading is there. Other than that - only CMakeLists.txt.
Source code contains *.ts files, during CMake run `lupdate` tool generates binary *.qm files ( which can be loaded by executable

I'll also recheck Linux packaging - it works for me in dev environment, but haven't checked resulting packages.

IvanSavenko:
I think I figured out what's wrong with translation install, will fix

kambala-decapitator:
sadly, no change on macOS. still no qm files in the bundle.

but noticed the following (same with older build): if I select German or Ukrainian, then VCMI Essentials info is shown on that language:

Screenshot 2023-01-16 at 11 57 35

doesn't work for Polish/Russian though (stays in English).

Nordsoft91:
sorry, for missing that. Tested on iOS - basically same as @kambala-decapitator has for macOS. Interface wasn't translated. Other than that looks good

IvanSavenko:
Now that's weird.
I can confirm that translations are now compiled & installed correctly on both Windows (msvc artifact) and Linux.

According to logs & content of produced artifacts I see that mac-intel also has translations. But obviously can't test that myself.
In the same time, on mac-arm & ios CMake (presumably) finds required Qt Linguist, but for some reason I don't see neither generation of translations (lines like `Generating translation/english.qm` are missing from log) nor do I see translation files in output build artifacts.

@kambala-decapitator @Nordsoft91 
Can somebody on Mac look into this? Because I am out of ideas at this point.

kambala-decapitator:
I suspect it's related to cross-compilation, will investigate

Nordsoft91 (approved):
Then I suppose this PR can be closed?

IvanSavenko:
> I suspect it's related to cross-compilation, will investigate

ARM & ios are cross-compiled? Then I may known what's wrong - wanted to disable translations for MXE. Will try a fix in a few minutes

kambala-decapitator:
> ARM & ios are cross-compiled?

iOS is always cross-compiled (obviously). ARM is too because currently there're no ARM runners on GitHub CI.

IvanSavenko:
Yeah, I guess I got tired from all those attempts to get it right and managed to miss the obvious...

IvanSavenko:
Build is done.
And now I see compiled translations in both ios and mac-arm artifacts.
So most likely fixed.

Feel free to recheck ios/mac. Notes on current status:
- vcmi mod (VCMI essential files) has translations for English, German and Ukrainian, rest - TODO
- Launcher & Map editor have translations for English, German, Polish and Ukrainian. Russian - TODO

Will probably merge soon

kambala-decapitator:
translations still don't work on iOS - see only this behavior:

> if I select German or Ukrainian, then VCMI Essentials info is shown in that language

IvanSavenko:
Any ideas what's wrong? Check log - it should contain info on translation loading.

kambala-decapitator:
> Any ideas what's wrong? Check log - it should contain info on translation loading.

thanks for the hint! it's looking in the wrong place:

> Searching for translation at '/private/var/containers/Bundle/Application/EE9205B9-2CD1-44E8-821E-4899A1DC1E36/vcmiclient.app/launcher/translation/ukrainian.qm'

the path has extra `launcher` directory

IvanSavenko:
Ah, ios used different directory for icons & I copied that logic for translations. Will fix.

kambala-decapitator:
> dunno if related to the PR, but game crashed on clicking "new scenario", updating VCMI extras didn't solve it (only this mod is installed)

still the case on iOS, crashed thread:

```
Thread 13 name:  CServerHandler::threadHandleConnection
Thread 13 Crashed:
0   libc++.1.dylib                	       0x19beebea4 std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >::basic_string(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 16
1   vcmiclient                    	       0x100ed6d10 InterfaceObjectConfigurable::readText(LIB_CLIENT::JsonNode const&) const + 80
2   vcmiclient                    	       0x100ed6d10 InterfaceObjectConfigurable::readText(LIB_CLIENT::JsonNode const&) const + 80
3   vcmiclient                    	       0x100ed3f28 InterfaceObjectConfigurable::buildLabel(LIB_CLIENT::JsonNode const&) const + 364
4   vcmiclient                    	       0x100ed8878 std::__1::__function::__func<std::__1::__bind<std::__1::shared_ptr<CLabel> (InterfaceObjectConfigurable::*)(LIB_CLIENT::JsonNode const&) const, InterfaceObjectConfigurable*, std::__1::placeholders::__ph<1> const&>, std::__1::allocator<std::__1::__bind<std::__1::shared_ptr<CLabel> (InterfaceObjectConfigurable::*)(LIB_CLIENT::JsonNode const&) const, InterfaceObjectConfigurable*, std::__1::placeholders::__ph<1> const&> >, std::__1::shared_ptr<CIntObject> (LIB_CLIENT::JsonNode const&)>::operator()(LIB_CLIENT::JsonNode const&) + 52
5   vcmiclient                    	       0x100ed6a50 InterfaceObjectConfigurable::buildWidget(LIB_CLIENT::JsonNode) const + 808
6   vcmiclient                    	       0x100ed285c InterfaceObjectConfigurable::build(LIB_CLIENT::JsonNode const&) + 1164
7   vcmiclient                    	       0x100f0a6e4 RandomMapTab::RandomMapTab() + 2144
8   vcmiclient                    	       0x100ef0164 CLobbyScreen::CLobbyScreen(ESelectionScreen) + 840
9   vcmiclient                    	       0x100e5d5d4 void CGuiHandler::pushIntT<CLobbyScreen, ESelectionScreen>(ESelectionScreen&&) + 72
10  vcmiclient                    	       0x100e5d4fc LIB_CLIENT::LobbyClientConnected::applyOnLobbyHandler(CServerHandler*) + 364
11  vcmiclient                    	       0x100d9fcc4 CApplyOnLobby<LIB_CLIENT::LobbyClientConnected>::applyOnLobbyHandler(CServerHandler*, void*) const + 168
12  vcmiclient                    	       0x100d906dc CServerHandler::threadHandleConnection() + 440
13  libboost_thread.dylib         	       0x1049de120 boost::(anonymous namespace)::thread_proxy(void*) + 232
14  libsystem_pthread.dylib       	       0x1de822348 _pthread_start + 116
15  libsystem_pthread.dylib       	       0x1de820948 thread_start + 8
```

launcher/CMakeLists.txt
```diff
@@ -83,18 +90,26 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 if(TARGET Qt6::Core)
 	qt_wrap_ui(launcher_UI_HEADERS ${launcher_FORMS})
+	if(ENABLE_TRANSLATIONS)
+		qt_add_translation( launcher_QM ${launcher_TS} )
+	endif()
 else()
 	qt5_wrap_ui(launcher_UI_HEADERS ${launcher_FORMS})
+	if(ENABLE_TRANSLATIONS)
+		set_source_files_properties(${launcher_TS} PROPERTIES OUTPUT_LOCATION ${CMAKE_CURRENT_BINARY_DIR}/translation)
```
kambala-decapitator, line 99: according to the [docs](https://doc.qt.io/qt-6/qtlinguist-cmake-qt-add-translations.html#options), this line should be used regardless of Qt version

launcher/CMakeLists.txt
```diff
@@ -83,18 +90,26 @@ set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
 if(TARGET Qt6::Core)
 	qt_wrap_ui(launcher_UI_HEADERS ${launcher_FORMS})
+	if(ENABLE_TRANSLATIONS)
+		qt_add_translation( launcher_QM ${launcher_TS} )
```
kambala-decapitator, line 94: consider using `qt_add_translations`. From [qt_add_translation](https://doc.qt.io/qt-6/qtlinguist-cmake-qt-add-translation.html):

> Warning: This function is deprecated. Consider using the target-based functions [qt6_add_lrelease](https://doc.qt.io/qt-6/qtlinguist-cmake-qt-add-lrelease.html#qt6-add-lrelease) or [qt6_add_translations](https://doc.qt.io/qt-6/qtlinguist-cmake-qt-add-translations.html#qt6-add-translations) instead.

launcher/CMakeLists.txt
IvanSavenko: Done

launcher/CMakeLists.txt
IvanSavenko: Added TODO in code.
That function looks to be quite different from the old one. Don't feel confident enough to correctly change code that is not tested by either CI or by me.

kambala-decapitator: agree

PR #1314, opened 2022-12-26
[1.1] Socket buffered write. Write full cpack at once

nullkiller:
Already boosted game load significantly. Will take a look onto read as well.

### DIFF
--- a/lib/serializer/Connection.h
+++ b/lib/serializer/Connection.h
@@ -21,8 +21,6 @@ namespace boost
 {
 	namespace asio
 	{
-		class streambuf;
-
 		namespace ip
 		{
 			class tcp;
@@ -54,6 +52,7 @@ typedef boost::asio::basic_socket_acceptor<boost::asio::ip::tcp, boost::asio::so
 VCMI_LIB_NAMESPACE_BEGIN
 
 struct CPack;
+struct ConnectionBuffers;
 
 /// Main class for network communication
 /// Allows establishing connection and bidirectional read-write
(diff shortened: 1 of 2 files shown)


dydzio0614:
I wonder if there is chance to boost hero movement speed in multiplayer by utilizing this feature

IvanSavenko:
> I wonder if there is chance to boost hero movement speed in multiplayer by utilizing this feature

Don't expect much. There may be some improvement especially on low-end system (read: Android phones), but most of hero movement lag comes from TCP connection. 
Especially when playing over Wi-Fi or mobile networks (3G/4G/etc)

nullkiller:
If somebody see a better solution for this or any caveats please let me know. I hardly understand what I have done.

lib/serializer/Connection.h
```diff
@@ -63,8 +65,16 @@ class DLL_LINKAGE CConnection
...
 	int write(const void * data, unsigned size) override;
 	int read(void * data, unsigned size) override;
+	void flushBuffers();
 
 	std::shared_ptr<boost::asio::io_service> io_service; //can be empty if connection made from socket
+
+	bool enableBufferedWrite;
+	boost::asio::streambuf writeBuffer;
+
+	bool enableBufferedRead;
```
Nordsoft91, line 75:
```suggestion
	bool enableBufferedWrite = false;
	boost::asio::streambuf writeBuffer;

	bool enableBufferedRead = false;
```

Nordsoft91 (commented):
Why not to keep bufferization always enabled?

IvanSavenko:
> Why not to keep bufferization always enabled?

I am not nullkiller, but I would also be cautious.
Currently change is controlled - we know that bufferization is only done during that single call.
If there is some code path that writes data to Connection without generating CPack then enabling bufferization always will stop that code from working (since it won't be sent until next CPack)

PR #1326, opened 2022-12-27
Support for Polish version of Heroes 3 from gog.com

IvanSavenko:
Small changes to make Polish H3 data work better with vcmi:
- fix yet another scenario of gog.com directory layout
- handle slightly different format of sound archives
- for main menu, position buttons using their central point (instead of top-left). This seems to be the way H3 does this, and it works in all versions that I've tested.

(also tested French version from gog. No idea whether it works correctly, but at least I can start a map)


dydzio0614:
I think this should be part of 1.1.1 hotfix, what do you think?

IvanSavenko:
Are we making said hotfix?

In any case - yeah, changes should be safe enough for a hotfix.

Issue #1300, opened 2022-12-25
Translations support 


IvanSavenko:
Meta-issue to organize progress, will be done in series of PR to avoid large amount of changes at once

**Progress**
- PR #1293 - everything related to Launcher

**Functionality**:
- [x] Implement translations for Launcher using Qt translator
- [ ] Implement translations for Map Editor using Qt translator
- [x] Allow mods to define mod information (e.g. name/description) with localization to current language of Launcher
- [ ] Provide mechanism to translate data from H3 files via mods that is better than direct editing of H3 files
- [ ] (optional/needs research) Provide mechanism to translate .h3m and .h3c files.
- [ ] Integrate existing translate.json into H3 translation support
- [ ] Provide mechanism for mods to translate new content
- [ ] Multiplayer support - ensure that client & server can run with different languages

**Languages**:
- [x] English (already available)
- [x] Ukrainian (done by me)
- [x] German (In progress, by [Laserlicht](https://github.com/Laserlicht))
- [ ] Polish (proposed by dydzio)
- [ ] Russian
- [ ] Any other languages?

**Translation assistance**
If there is anyone willing to provide missing/new translation ask me here or on Slack. What I will need is following:

- Translation of existing vcmi strings from [here](https://github.com/IvanSavenko/vcmi/blob/translate_launcher/config/translate.json) (they may already be translated somewhere in some mod)
- Translation of launcher strings.
If you have Qt Creator installed you can open [this](https://gist.github.com/IvanSavenko/9c2ed77c18b0d9c13c914f7c986906bd) file in Qt Linguist for nicer UI. 
Or you can edit file manually, [Ukrainian translated file](https://github.com/IvanSavenko/vcmi/blob/translate_launcher/launcher/translation/launcher_uk.ts) may be more suitable in this case - it has English strings included.
- Translation of Map Editor - not exported yet

heroesiiifan:
Hi @IvanSavenko 

First of all, thank you for your effort of a better language support.

For [german-translation](https://github.com/vcmi-mods/german-translation) I have a few questions:
* how is it thought in the future with the translate.json? Do we maintain the foreign languages directly in the source of VCMI? Have a current file (Can be intigrated into VCMI with pleasure). But with the current state it can't load via a mod (don't know why though).
* What about the bonuses_texts.json? Doesn't it need to be replaced as well?
* In my case I would leave the translated TXT as they are, they should still load without problems (regardless of the new option), right?
[...]

PR #1329, opened 2022-12-27
Translations support - base functionality for main game texts

IvanSavenko:
Yet another subtask of #1300 

Functionality:
- [x] Provide mechanism to translate data from H3 files via mods that is better than direct editing of H3 files
- [x] Integrate existing translate.json into H3 translation support

List of changes:
- all "simple" interface text files from H3 are now using unified identifiers access in form of `"core.{filename}.{string number}"`
- all "complex" interface text files also use unified identifiers, however exact string format depends on file content, e.g. help.txt uses `"core.help.{number}.{'help' or 'hover'}"`
- all vcmi texts from translate.json are now available in form `"vcmi.string_id_from_translate_json"`
- translate.json now uses simple key-value format without nested fields
- to avoid massive edits through all codebase, CGeneralTextHandler provides public members that emulate file-based access from before.

### DIFF
--- a/client/gui/InterfaceObjectConfigurable.cpp
+++ b/client/gui/InterfaceObjectConfigurable.cpp
@@ -92,19 +92,10 @@ std::string InterfaceObjectConfigurable::readText(const JsonNode & config) const
 	
 	const std::string delimiter = "/";
 	std::string s = config.String();
+	boost::replace_all(s, "/", "." );
+
 	logGlobal->debug("Reading text from translations by key: %s", s);
-	JsonNode translated = CGI->generaltexth->localizedTexts;
-	for(size_t p = s.find(delimiter); p != std::string::npos; p = s.find(delimiter))
-	{
-		translated = translated[s.substr(0, p)];
-		s.erase(0, p + delimiter.length());
-	}
-	if(s == config.String())
-	{
-		logGlobal->warn("Reading non-translated text: %s", s);
-		return s;
-	}
-	return translated[s].String();
+	return CGI->generaltexth->translate(s);
 }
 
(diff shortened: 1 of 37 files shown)


client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -92,19 +92,10 @@ std::string InterfaceObjectConfigurable::readText(const JsonNode & config) const
 	
 	const std::string delimiter = "/";
 	std::string s = config.String();
+	boost::replace_all(s, "/", "." );
```
Nordsoft91, line 95: Instead, let's change <widget>.json files

IvanSavenko: Agreed.
I believe right now there is only one mod that uses this code - your rmg template selection window.
To reduce disruptions I suggest:
1) wait for merge of this PR
2) change mod .json(s)
3) remove this code in separate PR

Nordsoft91: No worry about mod. We can update mod later, as it non-production anyway

IvanSavenko: Done. Changed cpp code and existing .json from code to match changes in branch:
- for labels, loading is done via ID in form of "core.genrltxt.XXX" where XXX is old string number
- for buttons, loading is done either in form of "core.help.XXX" (in which vcmi will lookup strings "core.help.XXX.help" and "core.help.XXX.hover")
- or, in form of directly specifying both help & hover strings: `"help" : { "help" : "...", "hover" : "..." }`

PR #1350, opened 2022-12-29
Enable strict compilation by default

IvanSavenko:
With all warnings fixed, I'd like to propose using -Werror by default.
This is primary targeted towards CI, meaning that all new PR's must be compile-able without warnings. 

We already had a few PR's that had actual bugs that were detected by compiler warnings but ignored by author & at least in one case - missed during review and made it to develop.

Extra bonus is that now *all* warning that you see are coming from your own code - you don't need to filter through output noise to find info relevant to your changes.


kambala-decapitator:
> This is primary targeted towards CI

then I'd say define the option in CI file, no need to enforce it on everyone to prevent future messages "I can't compile because of a strange error"

IvanSavenko:
> > This is primary targeted towards CI
> 
> then I'd say define the option in CI file, no need to enforce it on everyone to prevent future messages "I can't compile because of a strange error"

Done

IvanSavenko:
Any objections against merge?
@kambala-decapitator ?

kambala-decapitator (approved):

Issue #1339, opened 2022-12-29
Cartographer's reveal map logic is incorrect


kambala-decapitator:
Cartographer should not work by levels but rather by terrain types. So underground cartographer should reveal all underground tiles (and not underground layer).

From FizMig:

![Screenshot 2022-12-26 at 11 39 28](https://user-images.githubusercontent.com/1557784/209526330-bdbd42ff-9c53-4cff-a005-5a91c572a5f3.png)

initially discovered in #1303

PR #1351, opened 2022-12-30
Change cartographer tile reveal logic to match H3

dydzio0614:
Fixes https://github.com/vcmi/vcmi/issues/1339

### DIFF
--- a/lib/Terrain.cpp
+++ b/lib/Terrain.cpp
@@ -442,12 +442,12 @@ bool TerrainType::isUnderground() const
 
 bool TerrainType::isSurfaceCartographerCompatible() const
 {
-	return isLand() && id != Terrain::SUBTERRANEAN && id != Terrain::ROCK;
+	return isLand() && isPassable() && id != Terrain::SUBTERRANEAN;
 }
 
 bool TerrainType::isUndergroundCartographerCompatible() const
 {
-	return isLand() && id == Terrain::SUBTERRANEAN;
+	return isLand() && isPassable() && id == Terrain::SUBTERRANEAN;
 }
 
 bool TerrainType::isTransitionRequired() const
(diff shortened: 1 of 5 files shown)


lib/Terrain.cpp
```diff
@@ -440,6 +440,16 @@ bool TerrainType::isUnderground() const
...
 }
 
+bool TerrainType::isSurfaceCartographerCompatible() const
+{
+	return isLand() && id != Terrain::SUBTERRANEAN && id != Terrain::ROCK;
+}
+
+bool TerrainType::isUndergroundCartographerCompatible() const
+{
+	return isLand() && id == Terrain::SUBTERRANEAN;
```
IvanSavenko, line 450: 1) What about ROCK? Not 100% sure but I suspect that it should be "revealed" as well by this cartographer - to replace stars from terra incognita with revealed unpassable rocks

2) Don't use terrain types directly, instead use traits from config - we may have new terrains added by mods.
I see isPassable() check that can replace ROCK and isUnderground() that can replace direct check for SUBTERRANEAN

kambala-decapitator (changes requested):
you have a mix of tabs and spaces for indentation, please change to tabs only

dydzio0614:
ad. 1: ROCK is not revealed in original H3

ad. 2: isUnderground() also includes lava in our config

lib/Terrain.cpp
dydzio0614: ad. 1: ROCK is not revealed in original H3

ad. 2: isUnderground() also includes lava in our config

IvanSavenko: Hmm... Interesting, so there is no way to reveal ROCK in H3?
Okay then.

2 - In that case use isSurface()

dydzio0614:
Everything good now?

IvanSavenko:
Sounds OK for me. 
@kambala-decapitator?

kambala-decapitator (approved):

IvanSavenko:
Waiting my usual ~24 hours and merging tomorrow if there are no objections/further comments.

Issue #1341, opened 2022-12-29
set artifact not assembled after capturing missing pieces from enemy


kambala-decapitator:
save: https://github.com/vcmi/vcmi/files/10300591/Archive.tar.gz

after capturing town I'm asked if I want to assemble AotD, I say yes, but in fact it's not assembled in the backpack. I also see the following logs:

> Ranged attack animation is over
> Unrecognized trigger effect type 5
> Unrecognized trigger effect type 5
> assembleArtifacts: there is no such artifact instance!
> Got false in applying 17AssembleArtifacts... that request must have been fishy!
> Stopping music file MUSIC/COMBAT03 at 60.742001
> Stopping music file Music/Win Battle at 0.704000
> Warning: IntObject re-activated with mismatching used and active
> System message: Server encountered a problem: assembleArtifacts: there is no such artifact instance!
> System message: Server encountered a problem: Got false in applying 17AssembleArtifacts... that request must have been fishy!

initially discovered in #1313

SoundSSGood:
Will take a look

SoundSSGood:
@kambala-decapitator Unfortunately, I can't load your save file. VCMI extras mod incompatibility. I have a latest one.
If you update your mod and catch this bug again, please upload more saves.

kambala-decapitator:
@SoundSSGood simply edit `mod.json` of the mods that game complains about and set version to what the game wants in there. For example, you can comment the existing version field and add the "right" one, and revert afterwards.

PR #1359, opened 2023-01-02
end of battle art assembling fix

SoundSSGood:
Fix for https://github.com/vcmi/vcmi/issues/1341
Now vcmi will not ask after the battle to assemble artifacts if they fall into the backpack. (OH3 behaviour)

### DIFF
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -272,15 +272,18 @@ void EraseArtifact::applyCl(CClient *cl)
 	callInterfaceIfPresent(cl, al.owningPlayer(), &IGameEventsReceiver::artifactRemoved, al);
 }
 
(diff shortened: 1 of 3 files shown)


DjWarmonger:
I am pretty sure there's some Chinese map which requires assembling artifact during opponent's turn to win the following siege . But we'll probably have to live with that.

IvanSavenko:
> I am pretty sure there's some Chinese map which requires assembling artifact during opponent's turn to win the following siege . But we'll probably have to live with that.
Shouldn't have effect on such map - this is only about backpack arts.
They have no effect on hero & H3 is not proposing assembly in such cases either.

IvanSavenko (approved):

IvanSavenko:

PR #1373, opened 2023-01-05
Update defaultMods.json

dydzio0614:
Add comments explaining how new attack/defense manipulators work


IvanSavenko:

Issue #1298, opened 2022-12-25
cursor image jumps on click when hero can no longer move


kambala-decapitator:
In the video I click on the pikemen tower and the "horse" cursor suddenly jumps to the right. However, mouse cursor itself doesn't change the position (here the OS black arrow is visible), and slight mouse movement fixes the "horse" cursor position.

https://user-images.githubusercontent.com/1557784/209476461-f5a524d9-efac-4535-8801-da0c9b204011.mov

dydzio0614:
Can you recall if you were using fullscreen without "real fullscreen mode", with non-native resolution set?

kambala-decapitator:
> Can you recall if you were using fullscreen without "real fullscreen mode", with non-native resolution set?

Exactly

dydzio0614:
The deal is I talked to devs that use SDL back in the day and if I remember correctly for aspect ratio other than default one mouse position should be recalculated on every move to adjust its position... and there are some other things to be wary about. The real solution would be disabling ability to run "non-real fullscreen" for resolution other than screen native, as 99% of games do. Also IDK if related but somebody told me that he has alt-tab behavior problem on windows when running VCMI on multiple monitor setup.

IvanSavenko:
Most likely will be fixed by #2076

IvanSavenko:
Closing, presumably fixed

dydzio0614:
definitely fixed if borderless windowed fullscreen doesn't use resolution with aspect ratio different than screen native

PR #1374, opened 2023-01-05
Hardware cursor

IvanSavenko:
Implemented support for hardware cursors using SDL API

Cursor type (hardware/software) can be selected in settings.json, however right now there is no UI for this option. Will create after translation support/launcher is merged into develop, to avoid conflicts. By default, vcmi now uses hardware cursors

Please check whether this works on all systems, I think this is improvement over current cursors & can be set as default.


dydzio0614:
Problem #1: this change doesn't fix bug with cursor position :(
Problem #2: cursors have another color - it is darker, as if intensity of every pixel color of the cursor was higher - maybe 2 cursors are rendered?

IvanSavenko:
> Problem 1: this change doesn't fix bug with cursor position :( Problem 2: cursors have another color - it is darker, as if intensity of every pixel color of the cursor was higher - maybe 2 cursors are rendered?

1. Well, that's unfortunate. No idea what can cause that bug then.
2. Don't think so. H3 cursors don't have semi-transparency of any kind. Is is possible to make a screenshot somehow?

dydzio0614:
I tried but cannot make screenshot so far and i am 100% sure the cursor is darker, didn't test all cursors - main menu, adventure map horse, cursor when hovering over town, attack cursor checked so far

kambala-decapitator:
now cursor size is that of OS => when I run fullscreen with 1080p setting on 1440p monitor cursor's size doesn't scale any more, see video. I'd prefer to have it scaled if possible (old behavior), but that's not critical.

https://user-images.githubusercontent.com/1557784/210939014-b12fd952-2887-46e6-a37b-ac5b01dd23f8.mov

> Problem 1: this change doesn't fix bug with cursor position :(

I can also reproduce #1298, but slightly differently: now cursor stays in place (because now it's HW), but its image still changes as well as the hint in the bottom - as if it's actually in a different place.

IvanSavenko:
Rebased on new vcmi/develop, marking PR as ready for review & merge

IvanSavenko:
> now cursor size is that of OS => when I run fullscreen with 1080p setting on 1440p monitor cursor's size doesn't scale any more, see video. I'd prefer to have it scaled if possible (old behavior), but that's not critical.

Missed this comment. Strictly speaking - not possible, since cursor is controlled by OS. If OS provides option to change cursor size it may affect hardware cursor in-game.

However, it is possible to switch to old cursor via config option (will be added to Launcher UI).

Separate "cursor scaling" option might be possible, but requires implementation. Don't see it as priority at the moment.

dydzio0614:
I will dedicate some time trying to find out where cursor wrong color comes from... the cursor for me doesn't look the way it does on @kambala-decapitator video

IvanSavenko:
> I will dedicate some time trying to find out where cursor wrong color comes from... the cursor for me doesn't look the way it does on @kambala-decapitator video

It is possible that it is affected by system theme/cursor colors? E.g. you're using dark theme for system/cursor and OS attempts to make our cursor match it?

dydzio0614:
Did not look into this yet, there is non-zero chance that system stuff may affect something though.

dydzio0614:
Looks like linux issue. When I run VCMI via gamescope then cursor color is correct.

dydzio0614 (approved):

kambala-decapitator:
just tested on iOS: seems that iOS and android should keep using SW cursor, otherwise there's no visible pointer at all. Clicks/taps still work, but personally I prefer to see the cursor. I think the setting should default to SW on mobile platforms.

PR #1381, opened 2023-01-06
Client commands code improvements + ingame support

dydzio0614:
Command-line client commands now work from ingame console and got moved to separate file


IvanSavenko:
(haven't looked in detail)
I suggest to use some syntax to differentiate chat & commands.
It is quite possible for player to write (for example) "save" in chat to well, ask another player in MP to save game.
But right now I assume this will trigger a command?

Proposition: only attempt to interpret as commands strings that start with '/'.
So, if string starts with '/' then send to commands processor everything sans '/'
Otherwise - process message as game chat

dydzio0614:
what about "gosolo" pseudo-cheat from original H3

IvanSavenko:
Personally I'd use it unconditionally for any command, meaning - also for gosolo

client/ClientCommandManager.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#pragma once
+
+#ifndef VCMI_IOS
+class PlayerColor;
```
IvanSavenko, line 14: Should in theory fix ios compile
```suggestion
VCMI_LIB_NAMESPACE_BEGIN
class PlayerColor;
VCMI_LIB_NAMESPACE_END
```

client/widgets/AdventureMapClasses.cpp
```diff
@@ -1144,15 +1145,22 @@ void CInGameConsole::startEnteringText()
...
 		std::string txt = enteredText.substr(0, enteredText.size()-1);
 		LOCPLINT->cb->sendMessage(txt, LOCPLINT->getSelection());
 		previouslyEntered.push_back(txt);
+
+		if(txt.at(0) == '/')
+		{
+			//some commands like gosolo don't work when executed from GUI thread
+			boost::thread clientCommandThread(ClientCommandManager::processCommand, txt.substr(1), true);
+			clientCommandThread.detach();
+		}
```
IvanSavenko, line 1163: Proposition: move `LOCPLINT->cb->sendMessage` into else block here.
Meaning, don't send console commands as chat messages

client/ClientCommandManager.cpp
```diff
@@ -0,0 +1,548 @@
...
+	}
+	else if(commandName == "controlai")
+	{
+		std::string colorName;
+		readed >> colorName;
+		boost::to_lower(colorName);
+
+		ClientCommandManager::handleControlAi(colorName);
+	}
+	// Check mantis issue 2292 for details
```
IvanSavenko, line 454: Perhaps add "command no recognized" message or something like that?

client/ClientCommandManager.cpp
```diff
@@ -0,0 +1,540 @@
...
+			logGlobal->error(commandMessage);
+			break;
+		default:
+			std::cout << commandMessage;
+			break;
+	}
+
+	if(currentCallFromIngameConsole)
+	{
+		if(LOCPLINT && LOCPLINT->cingconsole)
```
IvanSavenko, line 535: Access to GUI objects from another thread requires lock.
I think this is correct way to lock, but please recheck
```suggestion
		boost::unique_lock<boost::recursive_mutex> un(*CPlayerInterface::pim);
		if(LOCPLINT && LOCPLINT->cingconsole)
```

IvanSavenko (approved):

client/CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ set(client_SRCS
 		NetPacksClient.cpp
 		NetPacksLobbyClient.cpp
 		SDLRWwrapper.cpp
+        ClientCommandManager.cpp
```
Nordsoft91, line 85: Please use tabs

lib/CConsoleHandler.h
```diff
@@ -76,7 +76,7 @@ class DLL_LINKAGE CConsoleHandler
 #endif
 	}
 
-    std::function<void(const std::string &)> *cb; //function to be called when message is received
+    std::function<void(const std::string &, bool)> *cb; //function to be called when message is received
```
Nordsoft91, line 79: Please, add comment about meaning of bool argument

client/ClientCommandManager.h
```diff
@@ -0,0 +1,31 @@
...
+	static void giveTurn(const PlayerColor &color);
+	static void printInfoAboutInterfaceObject(const CIntObject *obj, int level);
+	static void printCommandMessage(const std::string &commandMessage, ELogLevel::ELogLevel messageType = ELogLevel::NOT_SET);
+	static void handleGoSolo();
+	static void handleControlAi(const std::string &colorName);
+
+public:
+	ClientCommandManager() = delete;
+	static void processCommand(const std::string &message, bool calledFromIngameConsole);
+};
```
Nordsoft91, line 31: it doesn't manage something, it's event not a class but set of functions. So why not have set of functions wrapped by namespace?

client/Client.cpp
```diff
@@ -761,6 +761,20 @@ void CClient::reinitScripting()
...
+void CClient::removeGUI()
+{
+	// CClient::endGame
+	GH.curInt = nullptr;
+	if(GH.topInt())
+		GH.topInt()->deactivate();
+	adventureInt = nullptr;
+	GH.listInt.clear();
+	GH.objsToBlit.clear();
+	GH.statusbar = nullptr;
```
Nordsoft91, line 773: are curInt, adventureInt and statusBar shared_ptrs?
If so, please use reset() instead.
If not, isn't it a memory leak?

Nordsoft91 (changes requested):

client/ClientCommandManager.h
IvanSavenko: It does use "private" visibility to hide internal stuff from outside calls.
Perhaps remove 'static' and use it as proper class instead?
As for name - Controller or Executor perhaps?

client/Client.cpp
dydzio0614: This code was moved from different file without alteration

IvanSavenko: Not a memory leak - `operator =` will correctly release object if needed.
In fact, shared_ptr does not have a way to release object, so no memory leaks are possible.
(unique_ptr does have release(), but not shared_ptr)

client/ClientCommandManager.h
dydzio0614: so you recommend creating new instance for each usage?

IvanSavenko: In my opinion - yes, but on stack, without call to new

dydzio0614: Unsure about that yet, since this class is context independent and ability to freely create its object doesn't bring anything interesting to the table. The internal state isn't interesting for caller. Command can be executed at any arbitrary place in code so being static makes sense, from "domain driven design" point of view etc.

IvanSavenko: Personally don't care either way, as long as it stays as "class" - it is weird to have a namespace with random functions that should not be called.
Nordsoft didn't liked this "static class" approach, so I suggested alternative.

dydzio0614: well, in theory it could be singleton class, but does it matter? For example in C# you have Math class with bunch of static methods, Unity engine adds Random with another static methods (C# standard libraries also has Random but it's non static for some reason and often wrapped in some singleton so you don't create new instances over and over for the sake of good RNG seed)

IvanSavenko: Keep in mind that C#
1) Does not have concept of "namespace with functions", all functions must be part of a class, so Nordsoft proposition is something that only exists in C++
2) I assume C# uses singleton in these cases to avoid (relatively) costly object creation call. In C++, creation on stack is extremely cheap operation. Especially for a state-less class.
3) Singleton is often viewed as an anti-pattern. Our VLC would be a great example of such.

By the way, your class is *not* stateless due to `currentCallFromIngameConsole` member. So yeah, I also suggest to change this, but into a proper class, with allocation on stack.

dydzio0614: Done, though the currentCallFromIngameConsole is there for convenience and it is rewritten on every public function call - I just do not want to pass bool as argument everywhere within class

IvanSavenko: Sure, my complain against `currentCallFromIngameConsole` was mostly about having it as static (which is now fixes)
While this is close to impossible with current design, if you somehow do end up with two instances of your class, you could end up with them accessing same variable since it was static.

dydzio0614: would be still technically impossible under normal conditions for one client to type 2 messages from 2 sources at same time unless deliberately trying to break something

client/Client.cpp
dydzio0614: BTW... curInt and LOCPLINT are... not very smart pointers :laughing:

client/CMakeLists.txt
```diff
@@ -168,6 +169,7 @@ set(client_HEADERS
 		mapHandler.h
 		resource.h
 		SDLRWwrapper.h
+        ClientCommandManager.h
```
Nordsoft91, line 172: I assumed that you will fix spaces in both places, but seems like need to leave comment separately :)

lib/CConsoleHandler.h
```diff
@@ -75,8 +75,8 @@ class DLL_LINKAGE CConsoleHandler
 		funlockfile(stdout);
 #endif
 	}
-
-    std::function<void(const std::string &)> *cb; //function to be called when message is received
+	//function to be called when message is received - string: message, bool: whether call was made from in-game console
+    std::function<void(const std::string &, bool)> *cb;
```
Nordsoft91, line 79: Please, replace spaces with tab

Nordsoft91 (approved):

PR #1386, opened 2023-01-07
Do not allow selecting resolutions not supported by display

IvanSavenko:
Client only right now. Should prevent buggy mouse positioning after selecting weird resolutions

But don't see any easy way to do such filter in Launcher.
SDL provides way to query all available resolutions, so in-game filter is done.
Qt doesn't have such method. So I have no way to query which resolutions are supported while in Launcher.

One option would be to load SDL in Launcher, get the info and exit SDL. (SDL_InitSubsystem & SDL_Exit)
But this change is rather significant and not something I am willing to do for a hotfix.

NOTE: PR targets beta, but I no longer think that change is safe enough for beta

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -31,6 +31,7 @@
 	"vcmi.systemOptions.resolutionButton.help"  : "{Select resolution}\n\n Change in-game screen resolution. Game restart required to apply new resolution.",
 	"vcmi.systemOptions.resolutionMenu.hover"   : "Select resolution",
 	"vcmi.systemOptions.resolutionMenu.help"    : "Change in-game screen resolution.",
+	"vcmi.systemOptions.fullscreenFailed"       : "{Fullscreen}\n\n Failed to switch to fullscreen mode! Current resolution is not supported by display!",
 
 	"vcmi.townHall.missingBase"             : "Base building %s must be built first",
 	"vcmi.townHall.noCreaturesToRecruit"    : "There are no creatures to recruit!",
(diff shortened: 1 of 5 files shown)


dydzio0614:
What do you think about implementing instant resoluton change without restart?

IvanSavenko:
Possible, but not now.
Once I'm done with my current tasks I may try looking into applying "configurable UI" by Nordsoft to adventure map.
If everyhing goes well - it will be possible to select any resolution supported by monitor, without creating custom assets as in vcmi-extra-resolutions mod.

After that - yeah, restart-free resolution change may be possible.

dydzio0614:
As partial fix this is good enough, though I think fixing launcher somehow is actually more important than in-game

dydzio0614:
Though another question is... how will this fix act on mobile platforms

IvanSavenko:
From what I see in code - ios supports any resolution, so my filter should not apply to ios.
Not sure about Android though... Good question

IvanSavenko:
> As partial fix this is good enough, though I think fixing launcher somehow is actually more important than in-game

Yeah, I was actually not sure what to do, so basically pushed this as draft so it won't be lost.
Probably safe to assume that this one is on pause for now.

nullkiller:
As far as I know fullscreen has no effect on Android

IvanSavenko:
@nullkiller 
Question is a bit different.
What happens if player with (for example) FullHD screen selects 800x600 resolution instead of his native? Or some other weird one like 1363x786?
On PC this seems to be causing weird bugs if resolution is not reported as "supported" by monitor.

dydzio0614:
android has mobile-specific code for creating game window

IvanSavenko:
Will probably merge in-game only version, at least for now, before this code is lost.
Will think what to do with Launcher later.

PR #1394, opened 2023-01-08
Artifact movement refactoring

SoundSSGood:
This PR brings VCMI behavior closer to OH3
Fixed slot highlighting. Now the backpack slot is not highlighte.
backpack_highlighted
And other highlighting errors.
backpack_highlighted1

Fixed a bug while drag&drop swapping an equipped artifact (part of a collection). VCMI ask's to assemble the artifact if the rest of the parts were in the backpack.
Also now, when drag&drop artifacts swap, the artifact is moved to a special transit place instead of a backpack. This is a step towards the implementation the backpack size limit in the future.

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -670,6 +670,16 @@ CArtifactsOfHero::CArtifactsOfHero(const Point & position, bool createCommonPart
 CArtifactsOfHero::~CArtifactsOfHero()
 {
 	dispose();
+	// Artifact located in artifactsTransitionPos should be returned
+	if(!curHero->artifactsTransitionPos.empty())
+	{
+		auto artPlace = getArtPlace(
+			ArtifactUtils::getArtifactDstPosition(curHero->artifactsTransitionPos.begin()->artifact, curHero, curHero->bearerType()));
+		assert(artPlace);
+		assert(artPlace->ourOwner);
+		artPlace->setMeAsDest();
+		artPlace->ourOwner->realizeCurrentTransaction();
+	}
 }
 
 void CArtifactsOfHero::updateParentWindow()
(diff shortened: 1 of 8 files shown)


client/widgets/CArtifactHolder.cpp
```diff
@@ -278,13 +278,12 @@ void CHeroArtPlace::select ()
 	if (locked)
 		return;
 
-	selectSlot(true);
 	pickSlot(true);
 	if(ourArt->canBeDisassembled() && slotID < GameConstants::BACKPACK_START) //worn combined artifact -> locks have to disappear
 	{
-		for(int i = 0; i < GameConstants::BACKPACK_START; i++)
+		for(auto slot : ArtifactUtils::constituentWornSlots())
```
IvanSavenko, line 284: Change of behavior: old code included war machines, new one does not. Intended?

client/widgets/CArtifactHolder.cpp
```diff
@@ -309,9 +308,9 @@ void CHeroArtPlace::deselect ()
 	pickSlot(false);
 	if(ourArt && ourArt->canBeDisassembled()) //combined art returned to its slot -> restore locks
 	{
-		for(int i = 0; i < GameConstants::BACKPACK_START; i++)
+		for(auto slot : ArtifactUtils::constituentWornSlots())
```
IvanSavenko, line 311: Also here, same change of behavior

client/windows/GUIClasses.cpp
```diff
@@ -1249,8 +1249,6 @@ CExchangeWindow::CExchangeWindow(ObjectInstanceID hero1, ObjectInstanceID hero2,
 
 CExchangeWindow::~CExchangeWindow()
```
IvanSavenko, line 1250: Empty destructor - delete it entirely?

lib/CArtHandler.cpp
```diff
@@ -851,15 +856,15 @@ void CArtifactInstance::putAt(ArtifactLocation al)
 	assert(canBePutAt(al));
 
 	al.getHolderArtSet()->setNewArtSlot(al.slot, this, false);
-	if(!ArtifactUtils::isSlotBackpack(al.slot))
+	if(!ArtifactUtils::isSlotBackpack(al.slot) && (al.slot != ArtifactPosition::TRANSITION_POS))
```
IvanSavenko, line 859: Perhaps add another method similar to isSlotBackpack, e.g. isSlotEquipment (which returns true for all slots that are considered as "equipped" and provide bonuses to hero)

client/widgets/CArtifactHolder.cpp
SoundSSGood: yes, intended

client/widgets/CArtifactHolder.cpp
SoundSSGood: and here too

client/windows/GUIClasses.cpp
SoundSSGood: done

lib/CArtHandler.cpp
SoundSSGood: done

Issue #1133, opened 2022-11-14
Hero movement is very slow when playing online


Nordsoft91:
The reason - request-response per tile when moving

https://user-images.githubusercontent.com/9308612/201747023-3d7182d1-9198-4dac-b762-377eeaffa62c.mp4

IvanSavenko:
Not unexpected, I knew that this would arise once players play over Internet with non-zero ping.
Simple solution would be to send entire path, but players have ability to stop hero movement at any point.

What can be done:
Option a: send next request earlier:
- Send request to move to tile 1
- Wait for reply, start movement to tile 1 and immediately send request to move to tile 2
- Wait for hero to move to tile 1 & for server to allow movement to tile 2
In this scenario as long as ping is not higher than movement time, gameplay will be smooth & player can still stop hero at (almost) any point - hero will still move 1 tile after receiving order to stop.

Option b: move without receiving server confirmation. If servers denies our movement command (e.g. due to event) abort movement and return hero to last confirmed position. Stopping hero would be instant since we just not sending new message in such case. Not the greatest user experience but most likely move cancelling by server will be rare enough to not be too noticeable. 

Option c: send entire path to server. Problem - as it is right now, server will *immediately* apply entire movement making it impossible to cancel. There is also option of asking server "how far can I move without interruption" first, but again, server needs to be aware of how far hero will actually move in order to send correctly messages like FoW Revealed.
In my opinion - too many changes, too many potential problems.

Option d: start an epic quest to reduce this delay because I am 100% sure that ping is not the only factor here and that there are a lot more events that increase this delay further. Strictly speaking, as long as all players are on the same continent, ping should not be noticeable in games like this. Meaning that here we have delays from other sources, e.g. delay before network packet by client was actually send by system, delay before receiver actually read packet after it was received by system, delay before client started rendering new movement, etc. In theory, if all of these are fixed game should be playable even without other changes.

All in all, option A & B should be both doable. Not sure which of them is better or easier to implement though.

Nordsoft91:
Thanks for inputs, also from network standpoint, client-server interaction is performed via TCP sockets, which are not designed for low-latency systems. Real time games usually use UDP protocol but in our case as vcmi is a turn-based game TCP may work fine with a bit different cpacks design

Nordsoft91:
even slower while playing with @misiokles 
https://user-images.githubusercontent.com/9308612/202865502-06684af3-0b57-48d9-a5d7-89832ce1472d.mp4

[...]

PR #1400, opened 2023-01-10
Transfer data over UDP

Nordsoft91:
PR is needed to test have builds for all platforms

- [ ] Fix compilation for all platforms
- [ ] Manage disconnection
- [ ] Manage local multiplayer
- [ ] Manage in-game reconnections
- [ ] Reimplement online game


IvanSavenko:
Thanks for attempting to do this change. Hopefully enet is as good as described.

I know that this is very, very WIP, but perhaps create abstraction layer over enet instead of spreading it all over the code as we have with boost::asio right now? e.g:
- separate CConnection into IConnection interface + EnetConnection implementation class
- extract from CServerHandler class INetworkClient (with ENetClient as implementation)
- extract from CVCMIServerclass INetworkServer (with ENetServer as implementation)

Also potential problem with MXE (surprise). It seems that they don't provide precompiled package with enet. Meaning, we can't easily upgrade our MXE CI.

kambala-decapitator:
> Also potential problem with MXE (surprise). It seems that they don't provide precompiled package with enet. Meaning, we can't easily upgrade our MXE CI.

one more reason to switch to mingw / msvc and use vcpkg / [conan](https://conan.io/center/enet) (both have it)

dydzio0614:
BTW... @Nordsoft91 did you perform some early benchmark of this approach?

And side note on MXE stuff... I think MSVC is better pick for daily builds IMO for next releases anyway, since it better tuned for running resulting build on Windows in modern times while MXE's GCC 5 isn't. I would probably trust VS 2019 and 2022 optimizations more than what MXE can offer.

IvanSavenko:
@dydzio0614 
As somebody who worked on competitive FPS game: don't expect changes to *performance*. The only change would be decrease in *latency* when playing over network with non-zero ping (read - Internet). For example huge latency from #1133 should be reduced to ping, so ~50 ms if both players are in Europe. And any overhead on top of ping can be looked into and fixed which is not possible with TCP.

I do know of network emulator tool (clumsy) that can be used to check gameplay over Internet with like 0 setup time - we used it back in our FPS, but that thing is Windows-only.

on MXE: agreed. Keep MXE for any 1.1.x releases, but for 1.2 we should use msvc & remove MXE. If somebody can setup mingw to replace MXE - great. But that is optional IMO.

rilian-la-te:
@IvanSavenko Actually, Conan with mingw can be used easily and better than MXE. And then we can drop CMake 3.10.

dydzio0614:
I didn't exactly mean performance but reduction of networking overhead where it is known to cause problems

IvanSavenko:
> @IvanSavenko Actually, Conan with mingw can be used easily and better than MXE. And then we can drop CMake 3.10.

As I said, I have no objections against mingw, so if somebody can set it up - great. Not familiar with mingw or conan myself. 
But I do think that msvc should be the version we use for public releases.

BTW - we can already drop cmake 3.10, I had to upgrade MXE to include Qt translator, so MXE in develop now uses cmake 3.19. I think right now cmake 3.16 is our minimum - due to ubuntu-20.04.

Nordsoft91:
Before making refactoring or significant architecture changes, I want to make a prototype with minimal possible efforts to test whether it actually solves problem with latency. 
So current plan is to make MVP with multiplayer

IvanSavenko:
@Nordsoft91 If you still have latency issues with UDP, let me know - I should be able to investigate their source.

CI/linux/before_install.sh
```diff
@@ -8,4 +8,4 @@ sudo apt-get install libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf
 sudo apt-get install qtbase5-dev
 sudo apt-get install ninja-build zlib1g-dev libavformat-dev libswscale-dev libtbb-dev libluajit-5.1-dev
 # Optional dependencies
-sudo apt-get install libminizip-dev libfuzzylite-dev qttools5-dev
+sudo apt-get install libminizip-dev libfuzzylite-dev qttools5-dev enet
```
IvanSavenko, line 11:
```suggestion
sudo apt-get install libminizip-dev libfuzzylite-dev qttools5-dev libenet-dev
```

kambala-decapitator:
@Nordsoft91 conan deps now have enet. Please apply the following patch:

```diff
diff --git a/conanfile.py b/conanfile.py
index a3030c689..1b8dcd41f 100644
--- a/conanfile.py
+++ b/conanfile.py
@@ -10,6 +10,7 @@ class VCMI(ConanFile):
     settings = "os", "compiler", "build_type", "arch"
     requires = [
         "boost/[^1.69]",
+        "enet/[~1.3.17]",
         "minizip/[~1.2.12]",
         "onetbb/[^2021.3]", # Nullkiller AI
         "qt/[~5.15.2]", # launcher
@@ -31,6 +32,7 @@ class VCMI(ConanFile):
         "with_luajit": False,
 
         "boost/*:shared": True,
+        "enet/*:shared": True,
         "minizip/*:shared": True,
         "onetbb/*:shared": True,
     }
```

kambala-decapitator:
@Nordsoft91 please apply the patch exactly as I posted. Also please keep alphabetical order of the deps (like in my patch).

PR #1407, opened 2023-01-10
Revert "fix unaligned memory access when building with clang"

IvanSavenko:
I love MXE

I think due to this PR FPS on Windows got multiplied by zero.
Creating PR to trigger test build


IvanSavenko:
Update: I think I was wrong and this is unrelated. Will recheck after CI is done.

PR #1409, opened 2023-01-10
Refactoring of terrain handlers

IvanSavenko:
Original goal was to add translation support to terrains, but due to multiple issues with current terrain handler decided to go for a refactoring

- Converted TerrainHandler into a proper handler managed by mod system
- Roads and rivers are now managed by separate handlers
- Implemented support for new hero movement sounds 
- Implemented support for offroad hero movement sounds (h3 feature)
- Implemented validation for road & rivers
- Added not used previously validation for terrains
- Added translatable names for roads & rivers, used by map editor
- All terrains/roads/rivers strings are now under translation system
- Removed default native terrains - H3 neutral faction should not have any native terrain

NOTE 1: breaks save compatibility
NOTE 2: breaks mods that add new terrains (I can update broken mods if requested). 

To fix mod without my help: 
- rename json(s) with terrains/roads/rivers to something other, that won't replace vcmi config
- add reference to json's in mod.json
- edit terrains json to pass validation. Specifically:
- If not exists then add field `"type" : ["SURFACE"],`. If field exists, make sure that this field is an array/list and not single string.
- rename `"code"` to `"shortIdentifier"`
- change `"river"` to use full identifier of a river instead of short code
- rename`"horseSoundId"` field to `"horseSound"` and turn it into a string (check vcmi terrains.json or add your own sound)
- add `"horseSoundPenalty"` field


lib/TerrainHandler.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+#include "StdInc.h"
+#include "TerrainHandler.h"
+#include "VCMI_Lib.h"
+#include "CModHandler.h"
+#include "CGeneralTextHandler.h"
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+//regular expression to change id for string at config
```
DjWarmonger, line 19: Is this comment needed now?

DjWarmonger (approved):

lib/TerrainHandler.cpp
IvanSavenko: Not exactly sure what was its purpose even before my changes. I guess I'll remove it then.

Nordsoft91:
> NOTE 1: breaks save compatibility

so why not bump serialisation version?

config/schemas/road.json
```diff
@@ -0,0 +1,37 @@
...
+		{
+			"type": "string",
+			"description": "Human-readable name of the road"
+		},
+		"shortIdentifier":
+		{
+			"type": "string",
+			"description": "Two-letters unique indentifier for this road. Used in map format"
+		},
+		"tilesFilename":
```
nullkiller, line 25: Just a suggestion to name it tilesGraphics. tilesFilename also looks good I am only afraid that we do not work with files directly but with some kind of resources abstraction. So do we use term file in configs anywhere?

lib/CTownHandler.cpp
```diff
@@ -987,11 +967,13 @@ CFaction * CTownHandler::loadFromJson(const std::string & scope, const JsonNode
 	auto preferUndergound = source["preferUndergroundPlacement"];
 	faction->preferUndergroundPlacement = preferUndergound.isNull() ? false : preferUndergound.Bool();
 
-	//Contructor is not called here, but operator=
-	auto nativeTerrain = source["nativeTerrain"];
-	faction->nativeTerrain = nativeTerrain.isNull()
-		? getDefaultTerrainForAlignment(faction->alignment)
-		: VLC->terrainTypeHandler->getInfoByName(nativeTerrain.String())->id;
+	faction->nativeTerrain = ETerrainId::NONE;
```
nullkiller, line 970: No more getDefaultTerrainForAlignment?

config/schemas/road.json
IvanSavenko: From consistency point of view yes, "tilesGraphics" should be more suitable.
Not sure about readability though. "File" clearly states that modder needs file with terrain tile. "Graphics" is more abstract.

lib/CTownHandler.cpp
IvanSavenko: Correct. Problem is, that code would also trigger for faction of neutral creatures from H3 who are not supposed to have any native terrain, as results Nomads/Azure's/Snipers/etc got native terrain they were not supposed to have.

lib/RoadHandler.h
```diff
@@ -0,0 +1,79 @@
...
+
+	std::string tilesFilename;
+	std::string shortIdentifier;
+	ui8 movementCost;
+
+	RoadType();
+
+	template <typename Handler> void serialize(Handler& h, const int version)
+	{
+		h & tilesFilename;
```
nullkiller, line 46: Backward compatibility is broken again? Do not forget to change serialization version if yes

lib/TerrainHandler.cpp
```diff
@@ -0,0 +1,191 @@
...
+ */
+
+#include "StdInc.h"
+#include "TerrainHandler.h"
+#include "CModHandler.h"
+#include "CGeneralTextHandler.h"
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+TerrainType * TerrainTypeHandler::loadFromJson( const std::string & scope, const JsonNode & json, const std::string & identifier, size_t index)
```
nullkiller, line 18: Long line. Would be nice to split it

lib/TerrainHandler.cpp
```diff
@@ -0,0 +1,191 @@
...
+		ui8(blockedVec[2].Float())
+	};
+
+	info->passabilityType = 0;
+
+	for(const auto& node : json["type"].Vector())
+	{
+		//Set bits
+		auto s = node.String();
+		if (s == "LAND") info->passabilityType |= TerrainType::PassabilityType::LAND;
```
nullkiller, line 61: case sensitive comparison?

IvanSavenko: Yes, intended. Typo's or incorrect case should be detected by json validation.

lib/mapObjects/ObjectTemplate.cpp
```diff
@@ -231,21 +224,14 @@ void ObjectTemplate::readMap(CBinaryReader & reader)
 
 	reader.readUInt16();
 	ui16 terrMask = reader.readUInt16();
-	for(size_t i = Terrain::FIRST_REGULAR_TERRAIN; i < Terrain::ROCK; i++)
+	for(TerrainId i = ETerrainId::FIRST_REGULAR_TERRAIN; i < ETerrainId::ORIGINAL_TERRAIN_COUNT; ++i)
```
nullkiller, line 227: ORIGINAL_TERRAIN_COUNT and ROCK are the same?

lib/rmg/ConnectionsPlacer.cpp
```diff
@@ -84,9 +85,8 @@ void ConnectionsPlacer::selfSideDirectConnection(const rmg::ZoneConnection & con
 	
 	//1. Try to make direct connection
 	//Do if it's not prohibited by terrain settings
-	const auto& terrains = VLC->terrainTypeHandler->terrains();
-	bool directProhibited = vstd::contains(terrains[zone.getTerrainType()].prohibitTransitions, otherZone->getTerrainType())
-						 || vstd::contains(terrains[otherZone->getTerrainType()].prohibitTransitions, zone.getTerrainType());
+	bool directProhibited = vstd::contains(VLC->terrainTypeHandler->getById(zone.getTerrainType())->prohibitTransitions, otherZone->getTerrainType())
+						 || vstd::contains(VLC->terrainTypeHandler->getById(otherZone->getTerrainType())->prohibitTransitions, zone.getTerrainType());
```
nullkiller, line 89: Indentation is too big as for me

lib/mapObjects/ObjectTemplate.cpp
IvanSavenko: No, however that bit will be always at 0.
That being said, I'll recheck usages of ROCK & ORIGINAL_TERRAIN_COUNT.
Code is correct, but I find it weird to use ROCK (or any other identifier) as loop end point

misiokles:
"NOTE 2: breaks mods that add new terrains (I can update broken mods if requested)."

Please do, for HotA terrains/special terrains.

Alucard648:
And asphalt for Forge.

IvanSavenko:
@misiokles 
Done. Updated files here.
But please delete file `New Terrains\content\config\terrains.json` first.
Also, it is very likely that HotA has new sounds for movement on these terrains, you might want to check & update mod if necessary. Right now mod (as before) uses H3 sounds for grass & rough
[New Terrains.zip](https://github.com/vcmi/vcmi/files/10445463/New.Terrains.zip)

There is also something wrong with "Hota new objects" mod - RMG fails if that mod is on. Will investigate. Probably some issue on vcmi side.

IvanSavenko:
@Alucard648 
There is an issue with Forge mod - it has terrain in a submod, but terrain is actually used by main mod (as native terrain for forge).
This goes against current design - submods should be optional additions to main mod.
Mod worked before because this check was broken for terrains, but now it has been fixed.

What can be done here:
a) merge asphalt terrain mod into Forge mod
b) Make some other terrain, e.g. Dirt native terrain for Forge and have Asphalt mod change it to Asphalt if this mod is on.
c) move Forge itself into another submod and mark it as dependent on Asphalt submod

Options A&B should be easy to do. If you wish I can try doing it myself.

Updated files for asphalt mod, also need to delete old `content\config\terrains.json` file
[asphalt-terrain-mod.zip](https://github.com/vcmi/vcmi/files/10446115/asphalt-terrain-mod.zip)

misiokles:
I have 'plans' for Ashalt terrain (how to make it compatible with all sci-fi mods) but I must wait for 1.2 and Andruid's Antagarich Burning mod update.

PR #1419, opened 2023-01-11
Translation support for most of game objects

IvanSavenko:
Add usage of translator for most of objects accessible to mods.
All texts of these objects are now stored in GeneralTextHandler.
Changes are mostly renames/encapsulation - use getters methods instead of direct access to class members.

NOTE: breaks save compatibility. No effect on mods.


client/battle/BattleControlPanel.cpp
```diff
@@ -0,0 +1,328 @@
+/*
+ * BattleControlPanel.cpp, part of VCMI engine
```
DjWarmonger, line 2: Is that a part of this PR?

IvanSavenko: No, it isn't. Removed.

IvanSavenko:
I'd like to merge this one tomorrow.
Changes are spread through entire codebase, so very conflict-prone, but unlikely to cause issues.
If anyone want to postpone merge - let mee know

PR #1422, opened 2023-01-12
Creature bonus for modders: limited shooting range

dydzio0614:
Introduce LIMITED_SHOOTING_RANGE bonus, with "val" being required and  interpreted as max distance in hexes (calculated via same distance algorithm as "half damage" shooting), and addInfo is optional new range for half damage mechanics.

Sample usage:

```
"abilities" :
{
	"canShoot":
	{
		"type": "SHOOTER"
	}

	"limitedRange" :
	{
		"type" : "LIMITED_SHOOTING_RANGE",
		"val": 9
	}
}
```

results in creature being unable to shoot targets that are further than 9 hexes

```
"abilities" :
{
	"canShoot":
	{
		"type": "SHOOTER"
	}

	"limitedRange" :
	{
		"type" : "LIMITED_SHOOTING_RANGE",
		"val": 8,
		"addInfo": 4
	}
}
```

results in creature being unable to shoot targets that are further than 8 hexes, and does full damage only within 4 hexes range. Setting val as 99 or other high value is kind of workaround way to create bonus that manipulates half damage range mechanic and nothing else.

### DIFF
--- a/lib/HeroBonus.h
+++ b/lib/HeroBonus.h
@@ -330,7 +330,8 @@ public:
 	BONUS_NAME(GARGOYLE) /* gargoyle is special than NON_LIVING, cannot be rised or healed */ \
 	BONUS_NAME(SPECIAL_ADD_VALUE_ENCHANT) /*specialty spell like Aenin has, increased effect of spell, additionalInfo = value to add*/\
 	BONUS_NAME(SPECIAL_FIXED_VALUE_ENCHANT) /*specialty spell like Melody has, constant spell effect (i.e. 3 luck), additionalInfo = value to fix.*/\
-	BONUS_NAME(TOWN_MAGIC_WELL) /*one-time pseudo-bonus to implement Magic Well in the town*/ \
+	BONUS_NAME(TOWN_MAGIC_WELL) /*one-time pseudo-bonus to implement Magic Well in the town*/\
+    BONUS_NAME(LIMITED_SHOOTING_RANGE) /*limits range of shooting creatures, doesn't adjust any other mechanics (half vs full damage etc). val - range in hexes*/\
 	/* end of list */
 
 
--- a/lib/HeroBonus.h
+++ b/lib/HeroBonus.h
@@ -331,7 +331,7 @@ public:
 	BONUS_NAME(SPECIAL_ADD_VALUE_ENCHANT) /*specialty spell like Aenin has, increased effect of spell, additionalInfo = value to add*/\
 	BONUS_NAME(SPECIAL_FIXED_VALUE_ENCHANT) /*specialty spell like Melody has, constant spell effect (i.e. 3 luck), additionalInfo = value to fix.*/\
 	BONUS_NAME(TOWN_MAGIC_WELL) /*one-time pseudo-bonus to implement Magic Well in the town*/\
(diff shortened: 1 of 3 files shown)


lib/HeroBonus.h
```diff
@@ -330,7 +330,8 @@ class DLL_LINKAGE CAddInfo : public std::vector<si32>
 	BONUS_NAME(GARGOYLE) /* gargoyle is special than NON_LIVING, cannot be rised or healed */ \
 	BONUS_NAME(SPECIAL_ADD_VALUE_ENCHANT) /*specialty spell like Aenin has, increased effect of spell, additionalInfo = value to add*/\
 	BONUS_NAME(SPECIAL_FIXED_VALUE_ENCHANT) /*specialty spell like Melody has, constant spell effect (i.e. 3 luck), additionalInfo = value to fix.*/\
-	BONUS_NAME(TOWN_MAGIC_WELL) /*one-time pseudo-bonus to implement Magic Well in the town*/ \
+	BONUS_NAME(TOWN_MAGIC_WELL) /*one-time pseudo-bonus to implement Magic Well in the town*/\
+    BONUS_NAME(LIMITED_SHOOTING_RANGE) /*limits range of shooting creatures, doesn't adjust any other mechanics (half vs full damage etc). val - range in hexes*/\
```
Nordsoft91, line 334: tabs instead of spaces

IvanSavenko:
By the way, how distance penalty should work with this modifier? Perhaps have it at half range?

dydzio0614:
Both leaving it as is and limiting distance penalty make sense for different cases. For max range 4 it would be pretty dumb to require range 2 to deal full damage etc.

IvanSavenko:
> Both leaving it as is and limiting distance penalty make sense for different cases. For max range 4 it would be pretty dumb to require range 2 to deal full damage etc.

Well, with max range 4 it makes sense to also give "no range penalty" modifier (or what's its name). Meaning, half-range for penalty should still be good enough setting.

Alucard648:
What about doing coustom range damage modifier? For instance, creature has range limit of 8, but does full damage only 2 hexes away, otherwise damage reduced (-50% in OH3) within range limit. Should be modifable by other bonuses, like spells.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1611,6 +1627,16 @@ bool CBattleInfoCallback::battleHasDistancePenalty(const IBonusBearer * shooter,
 	return true;
 }
 
+bool CBattleInfoCallback::isEnemyUnitWithinSpecifiedRange(BattleHex attackerPosition, const battle::Unit * defenderUnit, unsigned int range) const
+{
+	for(auto hex : defenderUnit->getHexes())
+		if(BattleHex::getDistance(attackerPosition, hex) <= range)
+			return true;
+
+	//TODO what about shooting distance calculation for two-hex shooters? Is it correct?
```
IvanSavenko, line 1636: Checked H3 and yes, this is correct.
For double-wide shooters, game uses position of their "head", which matches what this code does

Nordsoft91:
Does Stupid/BattleAI handle it properly?

dydzio0614:
Yep, AI's handle this properly

Nordsoft91 (approved):

Issue #1236, opened 2022-12-16
Sieges: bugs & missing features


IvanSavenko:
Split from #1219 for convenience, planning to do separately due to its relatively large size.
If somebody is willing to do testing for me (all "Check X" tasks) using current develop branch that would be appreciated.

Misc:
- [ ] Check Teleport spell (only Advanced+ should allow teleporting inside walls)
- [ ] Check whether Cyclops wall attacks are functional
- [ ] Check whether Cyclop attack matches Basic Ballistics params (or Advanced for Cyclop Kings)

Castle Gates: https://heroes.thelazy.net/index.php/Wall
- [ ] castle gates should render unused now image ****DrwC.bmp with front part of gate (chains). Needed to correctly render stacks standing on top of the gate's bridge
- [ ] castle gates have no animation to lower/raise bridge
- [ ] if catapult makes multiple shots, gates will be destroyed only after final shot
- [ ] Check whether placing unit on tile to the right of the drawbridge will open it
- [ ] Check whether placing Force Field on moat will stop it from opening (that's a thing, apparently)

Castle Walls: https://heroes.thelazy.net/index.php/Wall
- [ ] Walls should change visual to more damaged state at midpoint of explosion playback instead of after it
- [ ] Walls should get extra hitpoint once Castle is built
- [ ] Check whether walls correctly incur ranged attacks penalty (and correctly update after damage to walls)

Arrow Towers: https://heroes.thelazy.net/index.php/Arrow_Tower
- [ ] Stack from destroyed tower is removed instantly, should be done at midpoint of the explosion
- [ ] Check whether Arrow Towers receive damage bonus depending on number of built structures in town

Moat: https://heroes.thelazy.net/index.php/Moat
- [ ] Check whether creatures in moat receive additional 10% damage from physical attacks
- [ ] Check whether creatures in moat have their defense reduces by 3
- [ ] Check whether Fortress moat is 2-tiles wide
- [ ] Check whether Tower moat (Mines) deals correct damage
- [ ] Check whether Inferno & Fortress moat deals 90 damage
- [ ] Check whether all other towns' moat deals 70 damage

Ballistics: https://heroes.thelazy.net/index.php/Ballistics
- [ ] Implement weighted-chance target selection 
- [ ] Make each attack into independent Pack, for easier animation ordering
- [ ] Check whether probability to hit intended wall matches H3
- [ ] Check whether dealt damage matches H3

rilian-la-te:
[...]

PR #1423, opened 2023-01-13
Siege mechanics fixes

IvanSavenko:
Implemented several tasks from #1236, namely:
- Castle now provides reinforced walls ( +1 hit point )
- Destruction of wall segments will now remove ranged attack penalty
- Arrow towers now have correct attack value (according to both thelazy & fizmig)
- Multiple fixes to ballistics mechanics and targeting
- Force Field cast in front of drawbridge will now block it as in H3


DjWarmonger:
Did you consider the possibility to boost the overall wall health?

At least H5 Fortress has a building which does that.

IvanSavenko:
> Did you consider the possibility to boost the overall wall HP?
> 
> At least H5 Fortress has a building which does that.

Considered. But hard to do with current design where health is represented by enum. 
Replacing current EWallState enum with "proper" HP would allow this. Perhaps I'll do this, but a bit later - don't want yet another massive PR

IvanSavenko:
Also, I'd like to merge this quickly to push fixed damage calculation (which had to be based on this branch to avoid massive conflicts).
Anyone wants to review this? If yes - then I'll wait for review, if not - merging as it & pushing damage fixes.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -164,8 +157,89 @@ ESpellCastProblem::ESpellCastProblem CBattleInfoCallback::battleCanCastSpell(con
 	return ESpellCastProblem::OK;
 }
 
+struct Point
```
Nordsoft91, line 160: can't believe it's not defined somewhere

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -24,44 +24,37 @@ VCMI_LIB_NAMESPACE_BEGIN
...
 	assert(town);
 	assert(turret->getPosition() >= -4 && turret->getPosition() <= -2);
 
-	const float multiplier = (turret->getPosition() == -2) ? 1.0f : 0.5f;
+	// base damage, irregardless of town level
+	static const int baseDamageKeep = 10;
+	static const int baseDamageTower = 6;
+
+	// extra damage, for each building in town
+	static const int extraDamage = 2;
```
Nordsoft91, line 40: can't be properties of CTown, loaded from config?

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -164,8 +157,89 @@ ESpellCastProblem::ESpellCastProblem CBattleInfoCallback::battleCanCastSpell(con
...
 
+struct Point
+{
+	int x,y;
+};
+
+/// Algorithm to test whether line segment between points line1-line2 will intersect with
+/// AABB (Axis-Aligned Bounding Box) defined by points aabb1 & aabb2
+/// Note that in order to avoid floating point rounding errors algorithm uses integers with no divisions
+static bool intersectionTestSegmentAABB(Point line1, Point line2, Point aabb1, Point aabb2)
```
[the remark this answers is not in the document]
IvanSavenko: "Rect" is not necessary axis-aligned and this is specifically boolean *test* since it does not detects intersection position(s)

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -164,8 +157,89 @@ ESpellCastProblem::ESpellCastProblem CBattleInfoCallback::battleCanCastSpell(con
...
+	int blTest = vector.y*aabb1.x - vector.x*aabb2.y + (line2.x*line1.y-line1.x*line2.y);
+	int brTest = vector.y*aabb2.x - vector.x*aabb2.y + (line2.x*line1.y-line1.x*line2.y);
+
+	// if all points are on the left of our line then there is no intersection
+	if ( tlTest > 0 && trTest > 0 && blTest > 0 && brTest > 0 )
+		return false;
+
+	// if all points are on the right of our line then there is no intersection
+	if ( tlTest < 0 && trTest < 0 && blTest < 0 && brTest < 0 )
+		return false;
```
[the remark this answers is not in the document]
IvanSavenko: No. Primary usage is "strike-through",e.g:
AABB = 1 wall tile on battlefield
line start = position of shooter
line end = position of target.

In thins case, both line1 & line2 are outside.

lib/battle/CBattleInfoCallback.cpp
IvanSavenko: It is defined. In client. With dependencies on SDL. But this code needs to be in library class, without SDL dependency.

lib/battle/CBattleInfoCallback.cpp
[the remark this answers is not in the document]
IvanSavenko: Now I am feeling like an idiot. Rect's in VCMI are always axis-aligned anyway. Will rename.

lib/battle/CBattleInfoCallback.cpp
IvanSavenko: Will try to separate our Point & Rect from SDL, but in another PR. Definitely needed, but outside of PR scope

lib/battle/CBattleInfoCallback.cpp
IvanSavenko: Will do, but in PR #1430 to avoid conflict

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -164,8 +157,89 @@ ESpellCastProblem::ESpellCastProblem CBattleInfoCallback::battleCanCastSpell(con
...
+struct Point
+{
+	int x,y;
+};
+
+/// Algorithm to test whether line segment between points line1-line2 will intersect with
+/// AABB (Axis-Aligned Bounding Box) defined by points aabb1 & aabb2
+/// Note that in order to avoid floating point rounding errors algorithm uses integers with no divisions
+static bool intersectionTestSegmentAABB(Point line1, Point line2, Point aabb1, Point aabb2)
+{
```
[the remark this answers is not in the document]
IvanSavenko: For now added assert's, later will move this code into (currently) client-only class Rect & rename to "intersect"

PR #1424, opened 2023-01-13
Initial version of castle hero exchange button

dydzio0614:
Looks like this:

obraz

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -508,9 +508,22 @@ void HeroSlots::update()
 
 void HeroSlots::splitClicked()
 {
-	if(!!town->visitingHero && town->garrisonHero && (visitingHero->isSelected() || garrisonedHero->isSelected()))
+	openHeroExchangeWindow(true);
+}
+
+void HeroSlots::heroMeetClicked()
+{
+	openHeroExchangeWindow(false);
+}
+
+void HeroSlots::openHeroExchangeWindow(bool heroSelectionRequired)
+{
+	if(!!town->visitingHero && town->garrisonHero)
 	{
-		LOCPLINT->heroExchangeStarted(town->visitingHero->id, town->garrisonHero->id, QueryID(-1));
+		if(!heroSelectionRequired || (visitingHero->isSelected() || garrisonedHero->isSelected()))
+		{
+			LOCPLINT->heroExchangeStarted(town->visitingHero->id, town->garrisonHero->id, QueryID(-1));
+		}
 	}
 }
 
@@ -1157,6 +1170,8 @@ CCastleInterface::CCastleInterface(const CGTownInstance * Town, const CGTownInst
 		heroes->splitClicked();
 	});
 	garr->addSplitBtn(split);
+
+	heroMeet = std::make_shared<CButton>(Point(246, 453), "TSBTNS", CButton::tooltip(CGI->generaltexth->tcommands[3]), [&](){heroes->heroMeetClicked();});
 
 	Rect barRect(9, 182, 732, 18);
 	auto statusbarBackground = std::make_shared<CPicture>(panel->getSurface(), barRect, 9, 555, false);
--- a/client/windows/CCastleInterface.h
+++ b/client/windows/CCastleInterface.h
@@ -121,8 +121,11 @@ public:
 	~HeroSlots();
 
 	void splitClicked(); //for hero meeting only (splitting stacks is handled by garrison int)
+	void heroMeetClicked();
 	void update();
 	void swapArmies(); //exchange garrisoned and visiting heroes or move hero to\from garrison
+private:
+	void openHeroExchangeWindow(bool heroSelectionRequired);
 };
 
 /// Class for town screen management (town background and structures)
@@ -214,6 +217,7 @@ class CCastleInterface : public CStatusbarWindow, public CGarrisonHolder
 
 	std::shared_ptr<CButton> exit;
 	std::shared_ptr<CButton> split;
+	std::shared_ptr<CButton> heroMeet;
 	std::shared_ptr<CButton> fastArmyPurchase;
 
 	std::vector<std::shared_ptr<CCreaInfo>> creainfo;//small icons of creatures (bottom-left corner);


IvanSavenko:
Sorry, but please, something less ugly than this.
Also, what's wrong with clicking on heroes?

krs0:
Why not use the 2 horizontal arrows (special cursor) for hero exchange on adventure map? Or something similar?

And if here... 2 buttons should be available: one to open hero exchange and one for hero Swap (hidden Space key). Hero swap could be 2 vertical arrows.

image

IvanSavenko (changes requested):
Why even use the button? Especially when it looks completely out of place in UI which was not designed for having a button here.
You could:
- Always open exchange window instead of hero window if there are two heroes in town
- Add a hotkey
- Use modifier key, e.g. Shift to open exchange
- Use existing "split troops" button
- Redesign town UI, using Nordsoft's configurable UI and delegate this to optional mod
- Any combination of options above
But please, no dropping buttons in random placed in UI. UX design exists for a reason.

krs0:
Hotkeys ad Modifiers are ok if you have kb and mouse. And are nice speedups, but they should not replace GUI functionality. How would that work on Tablets?

I for one play on PC from couch with mouse only! (with buttons assigned for CTRL, SHIFT, Wait, Defense).

Regarding Use existing "split troops" button...
Those buttons allows you to split troops for corresponding hero. And those should do just that.
There is need for another option to open hero meet window (where you could do more than split troops)
There is need for a button to swap position of the 2 heroes (same what you do with SPCAE hotkey)

IvanSavenko:
@krs 
Proposed button is quite small and located right between two other interactive elements. Good luck reliably hitting it on a smartphone.

We also generally don't have changes like this in the main game and delegate them to mods - see quick exchange or rmg tab.

I do agree with your downsides, but they don't apply to:
- Always open exchange window instead of hero window if there are two heroes in town (I think H5 does this?)
- Redesign town UI, using Nordsoft's configurable UI and delegate this to optional mod

> I for one play on PC from couch with mouse only! 
I think we should consider controller support 🤣

dydzio0614:
@IvanSavenko the thing is HD mod has button here and some people ask why it is missing in VCMI

IvanSavenko:
At the very least, HD mod:
- Changed background so button won't obscure decorative elements
- Changed button size, to fit better into available space
- Changed button icon, to better represent its action

Better solution IMO is to open Exchange Window in place of Hero Window, without any extra buttons.

dydzio0614:
In general I think button should be there, anyway current button is placeholder until the one with better fitting size / graphics gets there, for example something like this:
obraz

IvanSavenko:
This second example also does all 3 changes from HD Mod.
Another issue is resources. We should not have ripped assets in vcmi repo. And custom-made button likely will consist from them.

BTW - your code should block button if town does not have both visiting and garrisoned hero

dydzio0614:
Somebody needs to make fitting buttons from scratch

Nordsoft91:
Instead of hardcoding this functionality, I suggest:
1. inherit this window from InterfaceObjectConfigurable.
2. Configure json describing all interface widgets
3. Create separate mod with enabling this humid feature with json

See exmple: https://github.com/vcmi/vcmi/pull/1251

dydzio0614:
@Nordsoft91 I think reworking whole castle interface just for one button is overkill, especially when there are more important things to do.

Nordsoft91:
IMO, making UI/UX different from OH3 should be possible with mods only.
This PR breaks design unconditionally.

dydzio0614:
What about current creature window that is different from H3, or showing threat level for creatures? I think trivial changes can be part of config toggle on/off.

IvanSavenko:
> What about current creature window that is different from H3, or showing threat level for creatures? I think trivial changes can be part of config toggle on/off.

Frankly, even though I am the one behind new creature window, if I were to make same decision now I would have likely made a different one.
IIRC creature window was made before mods support (or at some very early stages of one). Meaning, "delegate to mods" was not an option back then.
Back then we also had WoG as part of vcmi, so this window was mostly necessary for endless bonuses that creatures receive with stack experience.
It also had a toggle, which was at some point removed for some reason.

Nordsoft91:
I agree with @IvanSavenko. This is eventually should be fix either

nullkiller:
@dydzio0614, regarding assets you can ask community and I think people will draw something for you. There is a separate repo for assets also if you need.

nullkiller:
I personally think it is better to have one button for exchange and it is not hard to make 2 clicks instead of one to swap heroes. Small buttons will be pain for touchscreens and for desktops there might be shortcuts

nullkiller:
But we should remember that better to make small step forward than big step never, or am I wrong?

dydzio0614:
regarding giving access, I believe bunch of people won't be satisfied unless many stuff work exactly same as in HD mod

nullkiller:
> Asking community is possible but at least I personally don't have much trust in community where majority of content is based on ripped assets...
It should be a part of vcmi-extras for sure. If icon is present - button is added. Same way hero exchange window works

val-gaav:
If there is going to be a mod in extras that changes town screen, then how about doing it a bit differently then hdmod and fixing the issue that town sidebar has only 3 towns on list (it is hard to navigate because of that ) Example mockups with and without month&week bar, that bar is not really needed IMHO as in most widescreen resolutions it should be visible with adventure map interface anyway . 


 
mockup
mockup2

krs0:
That is an excellent idea. "Combine" town screen with adv map interface on wide screens!

IvanSavenko:
Another option is to abandon 800x600 format and increase size of window, similar to how our battle queue requires resolution higher that 800x600.

For example, move town list to the right of town view & use newly freed area for any new controls that we need

(will require conversion of town screen class to "configurable UI" model. Nothing complex, but should be done by somebody)

val-gaav:
I guess Ivan idea is the best and I really like it. 

Beside hero exchange button what is really missing is army management buttons (alike to the ones from hero exchange window) . so a button to move all army up , a button to move all army down or switch the positions of units (up and down).

Furthermore some extra info about town could be displayed like name city hall / fort caste icons if build etc. 

Here is a quick mockup :
mockup-sidebar

krs0:
As for the transfer army buttons.. can you change the arrows to point down/up?

val-gaav:
Well, it is very functional though as it gives user a detailed info about town before he enters it. 

Queue bar in battle windows is also quite big and I do like it that way (smaller version is not that readable at least for me).


Mockup with arrows pointing up and down : 

mockup-sidebar2

val-gaav:
Smaller sidebar version (still would be cool to at least have town name there)
mockup-sidebar3

krs0:
Looks better, agree with Ivan swap hero aalso should have up down arrows

peterpaulis:
I wouldn’t maybe change the iconic town UI… but if switching castles is a problem, there could be some “castle switcher” available after pressing some hotkeys anywhere in the game… so the original UI experience remains

swats-the-floran:
aw. i just redrew the town screen >_<

swats-the-floran:
just found that i can open exchange screen by middle click and selecting the right button in hexagonal menu. it has few problems:
1. you should know that menu exists (thought it's only for android)
2. you have to make additional cursor positioning and click.
3. hexagonal menu has out of place design.

i suggest the button opening radial menu by hovering it with mouse cursor (and redesign). thus player will have to make two positionings but only one click - a compromise option.

image

top options in opening menu:  
pass all artifacts / pass all stacks to the top hero

bottom options:  
pass all artifacts / pass all stacks to the bottom hero

left option: swap heroes

right option: open exchange screen.

PR #1426, opened 2023-01-14
Update android JNI bindings

nullkiller:


IvanSavenko:
No objections either way, but if you want stable release for 1.1 then I think you should target beta branch instead?

nullkiller:
> No objections either way, but if you want stable release for 1.1 then I think you should target beta branch instead?

You are right

nullkiller:
Some functionality will be lost after this upgrade unfortunately. RMB I am going to reimplement using pure SDL touch handling and relative control should be done this way as well, but not sure when.

PR #1428, opened 2023-01-14
[Conan] update dependencies and toolchain for Apple platforms

kambala-decapitator:
see new lib versions in https://github.com/vcmi/vcmi-deps-macos/releases/tag/1.2

I'd also include this in `beta` since it has SDL version with fixed sound.


IvanSavenko:
If you wish to also get it into `beta` perhaps retarget/rebase it to `beta` and then merge it into `develop` alongside with #1421 ?

PR #1429, opened 2023-01-14
Implement numeric creature descriptions with config toggle on/off

dydzio0614:
Also includes changing QuantityID to enum, as minor refactoring.

To discuss: setting new behavior as default - for now I set it as ON, at least to raise awareness about existence of feature

I think there should be button in settings menu with bunch of VCMI settings like this one.
Another example stuff that could land in those settings: show grid, show battle order queue, more speed settings buttons in battle... etc.


IvanSavenko:
I think you can also get these descriptions by right-clicking on hero/town/garrison (with certain intel level).
Space is rather limited there - does it looks OK with your changes?

Also, perhaps we should bind "threat levels" to same same toggle? (to keep number of toggles at reasonable level)

dydzio0614:
@IvanSavenko Currently original quantities texts get completely replaced with new format so there is no space issue 
obraz

dydzio0614:
Some more screenshots:

obraz
obraz
obraz

IvanSavenko:
> @IvanSavenko Currently original quantities texts get completely replaced with new format so there is no space issue obraz

What about 500-1000?

dydzio0614:
500-999* and there is enough space for this. HotA mod does this anyway.

kambala-decapitator:
> 500-999* and there is enough space for this. HotA mod does this anyway.

and Legion (1000+) as well? :)

lib/CCreatureHandler.cpp
```diff
@@ -171,25 +171,53 @@ int32_t CCreature::getCost(int32_t resIndex) const
...
+		case CCreature::CreatureQuantityId::SWARM:
+			return "250-499";
+		case CCreature::CreatureQuantityId::ZOUNDS:
+			return "500-999";
+		case CCreature::CreatureQuantityId::LEGION:
+			return "1000+";
+		default:
+			return "[ERROR]";
+	}
 }
```
Nordsoft91, line 221: I would suggest to add static std::map<CCreatureQuantityId, std::string> instead

dydzio0614: You mean remove method and make public static map, or make method use private static map?

IvanSavenko: No public members please...

dydzio0614: yeah thats also what I think but the wording about adding static map "instead" got me confused

client/widgets/MiscWidgets.cpp
```diff
@@ -253,7 +254,16 @@ void CArmyTooltip::init(const InfoAboutArmy &army)
...
 			if(slot.second.count)
-				subtitle = CGI->generaltexth->arraytxt[171 + 3*(slot.second.count)];
+			{
+				if(settings["adventure"]["numericStackQuantities"].Bool())
+				{
+					subtitle = CCreature::getQuantityRangeStringForId((CCreature::CreatureQuantityId)slot.second.count);
+				}
+				else
+				{
+					subtitle = CGI->generaltexth->arraytxt[171 + 3*(slot.second.count)];
```
Nordsoft91, line 264: translate?

Nordsoft91, line 264: I see this 171+3* several times in code. What if making map non-const and overriding its values somewhere in the game loading if flag is false? Then you will just to load value from CCreature::getQuantityRangeStringForId whatever is there

dydzio0614: Not yet sure it will work fine, as I plan ingame toggle for this option somewhere in the future, so displayed way will change instantly

And there are bigger problems with related code, like CArmyInfo (or something that uses it?) sends quantityId 0 from server to indicate that count is "unknown"... etc

IvanSavenko: Another option - move all handling of this text into your method to avoid this duplication, e.g. something like
```
CCreature::getQuantityRangeStringForId()
{
    if (setting{"XXX"].Bool())
        return creatureQuantityRanges[count]; // return numeric form, e.g. 10-20
    else
        return CGI->generaltexth->translate("core.arraytxt", 171 + 3*count); // return H3 text for, e.g. "Few"
}
```

dydzio0614: but the base offset is not always 171

Nordsoft91: Changing instantly will work too, because you can override map values when toggle this option

dydzio0614: As I mentioned after, the value varies from 171 to 174, providing extra words that are present in H3, but are not really necessary in concise numerical format.

krs0:
Do we have also Hota format for adventure map monsters? EG: Guarded by Few (1-5) Angels? Threat... blah blah?

IvanSavenko:
> Do we have also Hota format for adventure map monsters? EG: Guarded by Few (1-5) Angels?
> 
> Also the "Threat:" wording only adds useless text to pop-up. It is pretty clear what Slightly weaker refers to.

We could implement nuclear option here and use formatting, e.g.
Status bar tooltip : `"%amountText (%amountRange) of %creatureName Threat: %threatLevel"`
(and similar ones for infobox & popup window)
All formatting options should be optional, and if they are found in string - replaced by code.

This will allow configuring tooltip text via localization system, without hardcoding stuff in code.
Extra bonus is that order of these substitutions might be different in different languages

dydzio0614:
> > Do we have also Hota format for adventure map monsters? EG: Guarded by Few (1-5) Angels?
> > Also the "Threat:" wording only adds useless text to pop-up. It is pretty clear what Slightly weaker refers to.
> 
> We could implement nuclear option here and use formatting, e.g. Status bar tooltip : `"%amountText (%amountRange) of %creatureName Threat: %threatLevel"` (and similar ones for infobox & popup window) All formatting options should be optional, and if they are found in string - replaced by code.
> 
> This will allow configuring tooltip text via localization system, without hardcoding stuff in code. Extra bonus is that order of these substitutions might be different in different languages

There are more use cases, for example info about guardians in banks. They are handled for now, using another varians of localized texts "x guardians / x creatures etc.". And for another option I am not sure about breaking compatibility with languages that we not have translated, like french

IvanSavenko:
> I am not sure about breaking compatibility with languages that we not have translated, like french - at least for now.

French will work same as now. From my example:
`"%amountText (%amountRange) of %creatureName Threat: %threatLevel"`
amountText and creatureName will be taken from H3 translation, as before.
amountRange is numbers-only and generated in code in this PR
threatLevel as well as "Threat" word are not translated anyway since there is no (AFAIK) French translation mod for vcmi strings.

Anyway, my proposition:
- use hota-style format where possible, e.g. "Few (4-9)". For places which don't have enough space, like info box - use number-only version. Not going to insist on such format, but I do think it is better option
- ideally, but not required - move pattern into translatable string, with separate versions for both states of your toggle, both to allow customization and to reduce string-building in code.
- Make it toggleable in UI (launcher or in settings). I am also OK with merging as it, if you're going to implement one for 1.2, e.g. as part of your vcmi settings PR.

dydzio0614:
Last option is good for me, some more reasoning:
About option 1: i prefer consistent format, and that one is not easily achievable for now for example in creature banks description - I do not want to rework how Metastring thing works etc. just for this.
About option 2: Would also involve rework with no much gain to support all cases rather than "creature right-click string only" - (creature banks etc.). And I doubt people are keen to customize their display formats for this thing.
Not mentioning variants of quantity strings present in H3 files - "x guardians", "x enemies" etc.

nullkiller:
Lets decide upon this branch. Any objections for it to be merged? People do ask such feature time to time so we have a mod which just changes strings

IvanSavenko:
> Lets decide upon this branch. Any objections for it to be merged? People do ask such feature time to time so we have a mod which just changes strings

Both this and #1545 need toggle in settings added. Currently there is no way to change this other than editing config file.
Another option is to switch default setting to H3 default, merge as it, and add option later.

PR #1432, opened 2023-01-14
VCMI settings menu

dydzio0614:
I believe as time goes there will be growing need for ingame settings menu.
Launcher isn't perfect target to be cluttered with minor settings like "show adventure map grid". Managing hardcoded features and stuff that change game mechanics sounds more reasonable for me in that case. Players will also want to be able to change various things via game itself, and also this gives less burden on android launcher.
Some settings to be added include: adventure map grid, on/off toggle for numeric creature quantity display, and probably a lot of HD mod features can be added here - for example "single new button" features that could be overkill to be implemented via mod, or things regarding concating extra info to status bar - movement points on adventure map etc.

Current draft version shown in video: https://streamable.com/khjvrg


IvanSavenko:
Proposition: instead of extending original H3 menu, I think it is better to replace it completely. Not because it is "bad", but mostly to avoid having two windows with settings. (technically, 3 - if you also count battle options window).

Also - consider adding tabs (General, Adventure Map, Battles) and integrating battle options into this window

IvanSavenko:
Also - use InterfaceObjectConfigurable & define window on json instead of hardcoding everything in code.

dydzio0614:
I planned to leave old settings as is, and use configurable interface for new stuff

krs0:
I agree with Ivan. VCMI should reorganize all settings in one place. No reason to stick with old H3 options anymore.
Window with multiple tabs is the way to go. A tab could be "Most common settings" so that players do not need to go through all tabs and get confused like in WoG.

One remark... Why do we need a button AND two labels for changing / displaying resolution? **Resolution 800x600**, with **800x600** clickable should be much user friendly.

Nordsoft91:
what is source of those assets?

IvanSavenko:
@Nordsoft91 
Seems to be these: https://github.com/vcmi/vcmi-assets/tree/develop/Interface
Public domain assets I collected for current creature window + my own work (and now dydzio's apparently)

dydzio0614:
> what is source of those assets?

https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/Data/questDialog.png
https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/Sprites/stackWindow/upgrade-normal.png
https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/Sprites/stackWindow/upgrade-pressed.png

dydzio0614:
Ok, so we have few ways how to approach this thing. First question is if we want to extended settings work without any mods - I think it should. And another question is - if we want them to work without any mods, then what about "compatibility with original game" - do we consider unchanged options menu (except resolution part which is pretty minimal change) as part of UI that should have its original game version?

IvanSavenko:
> First question is if we want extended settings to work without any mods - I think it should. 
Yes, Of course, assuming "vcmi" mod is still present since we use for essential files (unlike optional files in vcmi-extra)

> And another question is - if we want them to work without any mods, then what about "compatibility with original game" - do we consider unchanged options menu as part of UI that should have its original game version? 
Probably we can replace it completely with your menu. 
Thing is, we already use it differently. Instead of video quality we have resolution selector and one of checkboxes (not sure which one) has been replaced with fullscreen toggle.
I also quite often see requests for faster movement speed for both battle and adventure map. And we can't add those without changing menu.
I mean, sure we do have option of keeping speed setting in H3 window as before and add more detailed settings in vcmi menu. But this is extremely illogical and not user friendly.

At the moment, I see no other long-term solution that will both satisfy (valid) player requests and will allow us to keep H3 settings windows (both in-battle and adventure map)

Issue #1356, opened 2023-01-02
[DELETE] key doesn't work in map editor


DjWarmonger:
**Describe the bug**
Pressingle [DELETE] in map editor does nothing.

**To Reproduce**
1. Launch Map Editor
2. Generate a random map (M+U in my case)
3. Select one or more objects

**Expected behavior**
Objects get removed from the map.

**Actual behavior**
Nothing happens as far as I see.

**Did it work earlier?**
Certainly, I added the hokeys myself.

**Version**
Windows  10, VCMI 1.1.0 release

PR #1433, opened 2023-01-14
Added pressing and deleting with the [DELETE] key

SergeyG22:
added pressing and deleting with the [DELETE] key (issue #1356 [DELETE] key doesn't work in map editor)

### DIFF
--- a/mapeditor/mainwindow.cpp
+++ b/mapeditor/mainwindow.cpp
@@ -1181,3 +1181,13 @@ void MainWindow::on_actionPaste_triggered()
 	}
 }
 
+void MainWindow::keyPressEvent(QKeyEvent* event)
+{
+	switch (event->key()) 
+	{
+	case Qt::Key_Delete: 
+			on_actionErase_triggered();
+			break;
+	}
+}
+
--- a/mapeditor/mainwindow.cpp
+++ b/mapeditor/mainwindow.cpp
@@ -1186,8 +1186,8 @@ void MainWindow::keyPressEvent(QKeyEvent* event)
 	switch (event->key()) 
 	{
 	case Qt::Key_Delete: 
-			on_actionErase_triggered();
-			break;
+		on_actionErase_triggered();
+		break;
 	}
 }
 
--- a/mapeditor/mainwindow.cpp
+++ b/mapeditor/mainwindow.cpp
@@ -233,6 +233,10 @@ MainWindow::MainWindow(QWidget* parent) :
 	//Load map from command line
 	if(!mapFilePath.isEmpty())
 		openMap(mapFilePath);
+
+	keyBackspace = new QShortcut(this);
+	keyBackspace->setKey(Qt::Key_Backspace);
+	connect(keyBackspace, &QShortcut::activated, this, &MainWindow::backSpaceAction);
 }
 
 MainWindow::~MainWindow()
@@ -1179,6 +1183,11 @@ void MainWindow::on_actionPaste_triggered()
 	{
 		controller.pasteFromClipboard(mapLevel);
 	}
+}
+
+void MainWindow::backSpaceAction()
+{
+	on_actionErase_triggered();
 }
 
 void MainWindow::keyPressEvent(QKeyEvent * event)
--- a/mapeditor/mainwindow.cpp
+++ b/mapeditor/mainwindow.cpp
@@ -1181,7 +1181,7 @@ void MainWindow::on_actionPaste_triggered()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mapeditor/mainwindow.cpp
```diff
@@ -1181,3 +1181,13 @@ void MainWindow::on_actionPaste_triggered()
 	}
 }
 
+void MainWindow::keyPressEvent(QKeyEvent* event)
```
kambala-decapitator, line 1184: according to our current [coding guidelines](https://wiki.vcmi.eu/Coding_guidelines#Where_to_put_spaces) should be:
```suggestion
void MainWindow::keyPressEvent(QKeyEvent * event)
```

mapeditor/mainwindow.h
```diff
@@ -161,4 +161,6 @@ public slots:
 
 	// command line options
 	QString mapFilePath;			// FilePath to the H3 or VCMI map to open
+protected:
+	virtual void keyPressEvent(QKeyEvent* event);
```
kambala-decapitator, line 165:
```suggestion

protected:
	virtual void keyPressEvent(QKeyEvent * event);
```

kambala-decapitator (changes requested):
the approach itself looks good

Nordsoft91:
Seems like delete and backspace are handled differently. Is it possible to handle them in same way?
Also, is it possible to define hotkey for button as it done for all other actions?

SergeyG22:
'delete' and 'backspace' are handled differently because calling 'backspace' in a standard event loop doesn't work properly. Perhaps installing an event filter will solve this. But I do not know how to do it. The easiest way I've found is to use a hotkey.

kambala-decapitator:
> 'delete' and 'backspace' are handled differently because calling 'backspace' in a standard event loop doesn't work properly. 

at least on macOS I don't see any issues with the following test code:

```cpp
#include <QtWidgets>

class MainWindow : public QMainWindow
{
protected:
	virtual void keyPressEvent(QKeyEvent* e) override
	{
		switch (e->key())
		{
		case Qt::Key_Backspace:
			qDebug("Key_Backspace");
			break;
		default:
			QMainWindow::keyPressEvent(e);
			break;
		}
	}
};

int main(int argc, char* argv[])
{
	QApplication a{argc, argv};
	MainWindow w;
	w.show();
	return a.exec();
}
```

> Perhaps installing an event filter will solve this.

`keyPressEvent` method is identical to handling key presses in event filter

> please handle Backspace key on macOS instead

my original intention was to handle Backspace on macOS and Delete on other platforms (`#ifdef` would've been the easiest solution), sorry for not being clear. I'd suggest to just do this removing all the complexity of the last commit.

SergeyG22:
I don't have a macOS distribution. What you have given works well in the test project. Try adding a 'delete' handler to this code there as well. Still working? And now do the same in the map designer. The bundle of delete and backspace does not work on Windows 10. It seems that backspace is in conflict with something. I don't know anything about Mac. It's not just that I took out the backspace in the hotkey.

kambala-decapitator:
I'll test how Backspace behaves in the editor on macOS

kambala-decapitator:
@SergeyG22 have you tested erase functionality? For me it doesn't work at all on macOS - tried from toolbar, from menu, from instrument palette, with the new shortcut... @Nordsoft91 does it work for you?

btw you can attach key sequence directly to the action, no need to add a separate one or handle key event on main window. I also see that current shortcut is "press backspace, then press delete" which is obviously wrong.

kambala-decapitator:
@SergeyG22 so, does erase work for you on your OS?

@Nordsoft91 and what about you on macOS?

SergeyG22:
bandicam 2023-01-18 11-41-57-868

kambala-decapitator:
thanks, I'm stupid: I was trying to erase land instead of objects =)

> btw you can attach key sequence directly to the action, no need to add a separate one or handle key event on main window. I also see that current shortcut is "press backspace, then press delete" which is obviously wrong.

so let's do this:
1. remove all your changes
2. remove shortcut of erase action from the .ui file
3. in code (in constructor) give the action Backspace key sequence on macOS and Delete on other platforms using `#ifdef`

SergeyG22:
I'll do it as I have free time

Nordsoft91 (approved):
We can keep as is

kambala-decapitator:
> We can keep as is

don't see any reason to bloat code with strange workarounds when doing things right is trivial

Nordsoft91:
Fixed in other PR

PR #1434, opened 2023-01-15
Handle 2 fingers gesture as rclick via SDL

nullkiller:
This is for Android but may be suitable for any touch screen. Second finger triggers right mouse click. I swapped it, popup will appear now under second finger so you can rclick a few things at once without releasing the first finger. Long touch is also possible to implement, but some timer is required for this to detect when some time has passed since the touch started.


client/gui/CGuiHandler.cpp
```diff
@@ -338,22 +356,53 @@ void CGuiHandler::handleCurrentEvent()
...
+			case SDL_BUTTON_RIGHT:
+				handleMouseButtonClick(rclickable, EIntObjMouseBtnType::RIGHT, false);
+				break;
+			case SDL_BUTTON_MIDDLE:
+				handleMouseButtonClick(mclickable, EIntObjMouseBtnType::MIDDLE, false);
+				break;
+			}
+		}
+	}
+	else if(current->type == SDL_FINGERDOWN)
```
nullkiller, line 381: Should we allow this section only for ANDROID or is it ok to live it for all devices?

nullkiller: I do not see any harm if iOS will also support 2-finger rclick

Nordsoft91: I can test on iOS

Nordsoft91: on iOS doesn't work properly - impossible to close pop-up windows.
Long touch works well on iOS, so I'd recommend to disable double touch as right click for iOS, but maybe reuse it as "space" behavior

nullkiller: Maybe it is similar to windows simulator issue. Do you remove both fingers at once?

nullkiller: Hope, I managed to fix it. Can you try one more time?

nullkiller: It was even reproducible on Android. You just need to remove fingers quick enough.

kambala-decapitator:
it works on iOS, but I wouldn't say that quite reliably from my tests - long touch is way more convenient

PR #1438, opened 2023-01-15
Adjust newly built building appearance and outline blinking speed

dydzio0614:
Make buildings appearance look OK with new framerate + match timings description from comments

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -154,14 +154,9 @@ SDL_Color multiplyColors(const SDL_Color & b, const SDL_Color & a, double f)
 
 void CBuildingRect::show(SDL_Surface * to)
 {
-	const ui32 stageDelay = 500;
+	uint32_t stageDelay = BUILDING_APPEAR_TIMEPOINT;
 
-	const ui32 S1_TRANSP  = 500; //500 msec building appear 0->100 transparency
-	const ui32 S2_WHITE_B = 1000; //500 msec border glows from white to yellow
-	const ui32 S3_YELLOW_B= 1500; //500 msec border glows from yellow to normal
-	const ui32 BUILDED    = 2500; //1000 msec delay, nothing happens
-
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Why not use time instead of frame count?
Change those constants to milliseconds and instead of incrementing counter by 1 on each update increment it by time delta (from framerate manager)

dydzio0614:
I dont wanna break anything for now, just minor tuning for next potential stream testing

IvanSavenko:
Judging from code - such changes would be small, contained & unlikely to have side effects

IvanSavenko (approved):

client/windows/CCastleInterface.cpp
```diff
@@ -632,9 +632,9 @@ void CCastleBuildings::addBuilding(BuildingID building)
 		{
 			//reset animation
 			if(structures.size() == 1)
-				buildingRect->stateCounter = 0; // transparency -> fully visible stage
+				buildingRect->stateTimeCounter = 0; // transparency -> fully visible stage
 			else
-				buildingRect->stateCounter = 16; // already in fully visible stage
+				buildingRect->stateTimeCounter = 500; // already in fully visible stage
```
Nordsoft91, line 637: 1. Move S1_TRANSP and other consts outside of class method
2. 
```suggestion
				buildingRect->stateTimeCounter = S1_TRANSP; // already in fully visible stage
```

client/windows/CCastleInterface.h
```diff
@@ -42,6 +42,14 @@ class CBuildingRect : public CShowableAnim
 {
 	std::string getSubtitle();
 public:
+	enum EBuildingCreationAnimationPhases : uint32_t
```
Nordsoft91, line 45: minor suggestion - make it private

Nordsoft91 (approved):

client/windows/CCastleInterface.h
dydzio0614: it wont work because another class uses it

PR #1439, opened 2023-01-15
Fix wrong rendering of castle shipyard

dydzio0614:
Two bugs:
1. if shipyard built before citadel - animation is made out of frames that are intended to be used interchargeably for shipyard depending on citadel build state
2. if shipyard is built and citadel is also present, the citadel water stuff cover shipyard

These bugs also affect shipyard with ship


IvanSavenko (approved):
No objections from me, although I wish there was better way that does not involves hardcoding this stuff...

PR #1446, opened 2023-01-16
Implement movement points info on ALT hold + tile mouse hover

dydzio0614:
Show movement points cost for moves when holding alt key and hovering mouse over visitable tile. Not thoroughly tested but looks OK for basic use cases. Reuses H3 strings. Feature present in HD mod or HotA (not sure which), wanted for multiplayer.

### DIFF
--- a/client/windows/CAdvmapInterface.cpp
+++ b/client/windows/CAdvmapInterface.cpp
@@ -1708,7 +1708,7 @@ void CAdvMapInt::tileHovered(const int3 &mapPos)
 		else
 			CCS->curh->set(Cursor::Map::POINTER);
 	}
-	else if(const CGHeroInstance * h = curHero())
+	else if(const CGHeroInstance * hero = curHero())
 	{
 		std::array<Cursor::Map, 4> cursorMove      = { Cursor::Map::T1_MOVE,       Cursor::Map::T2_MOVE,       Cursor::Map::T3_MOVE,       Cursor::Map::T4_MOVE,       };
 		std::array<Cursor::Map, 4> cursorAttack    = { Cursor::Map::T1_ATTACK,     Cursor::Map::T2_ATTACK,     Cursor::Map::T3_ATTACK,     Cursor::Map::T4_ATTACK,     };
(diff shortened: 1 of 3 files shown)


client/windows/CAdvmapInterface.cpp
```diff
@@ -1721,6 +1721,18 @@ void CAdvMapInt::tileHovered(const int3 &mapPos)
...
 		assert(pnode);
 
+		if(LOCPLINT->altPressed() && pnode->reachable()) //overwrite status bar text with movement info
+		{
+			const int maxMovementPointsForDestinationLastTurn = pnode->turns > 0 ? h->maxMovePoints(pnode->layer == EPathfindingLayer::LAND) : h->movement;
+			const int movementPointsLastTurnCost = maxMovementPointsForDestinationLastTurn - pnode->moveRemains;
+			const int remainingPointsAfterMove = pnode->turns == 0 ? pnode->moveRemains : 0;
+
+			const std::string costWord = VLC->generaltexth->allTexts[346];
+			const std::string leftWord = VLC->generaltexth->allTexts[200];
```
Nordsoft91, line 1731: @IvanSavenko , now we should use VLC->generaltexth->translate, right?

IvanSavenko: Yeah, good idea probably.
```suggestion
			const std::string costWord = VLC->generaltexth->translate("core.genrltxt.346");
			const std::string leftWord = VLC->generaltexth->translate("core.genrltxt.200");

```
For H3 files, format is "core.<filename>.<line>".
You can dump all loaded texts via "convert txt" command

IvanSavenko:
@dydzio0614 Can you make a screenshot or write down exact text from status bar? Case for when turns count is above 0.
Also, does releasing Alt correctly restores original text?

dydzio0614:
Releasing alt restores original text on next refresh, that is - when you hover another tile etc.
Exact text example is: Cost: 1T + 1000, Left: 0

dydzio0614:
T = turn, 1000 is number of movement points

IvanSavenko:
I know. Now explain that to every other player who have discovered this feature.

kambala-decapitator:
> T = turn

I don't think I would have figured that out :)

dydzio0614:
When stuff change as you hover different tiles this gets more clear, though this is basic version of feature, so at least it exists for people that know they need it

IvanSavenko:
Why you love "basic features" so much? Even when "full feature" is not that much harder to implement?
We CAN make it better, cost/time of better solution is not much higher, so why shouldn't we go for a better option from the start?

I mean, I understand that with massive features when it is hard to estimate effects of "full" feature or which of potential approaches is better. But in case of single tooltip - why stop at a midpoint?

dydzio0614:
If you do not want to merge it as is, I may look at it in coming days - i also wanna see if I can get any more useful info from pathfinder when I get free time

IvanSavenko:
Well, it is not that I am *against* this feature in either form, but I do think it can be improved easily. And I don't believe we are in hurry to have this merged.

IvanSavenko (approved):

PR #1451, opened 2023-01-16
Resource converter code refactoring

krs0:
2 small fixes and some general refactoring for Resource converter. Tests generated same outputs.

### DIFF
--- a/mapeditor/resourceExtractor/ResourceConverter.cpp
+++ b/mapeditor/resourceExtractor/ResourceConverter.cpp
@@ -39,7 +39,7 @@ void ResourceConverter::doConvertPcxToPng(const bfs::path & sourceFolder, bool d
 {
 	logGlobal->info("Converting .pcx to .png from folder: %s ...\n", sourceFolder);
 
-	for(bfs::directory_entry & directoryEntry : bfs::directory_iterator(sourceFolder))
+	for(const auto & directoryEntry : bfs::directory_iterator(sourceFolder))
 	{
 		const auto filename = directoryEntry.path().filename();
 		try
@@ -47,7 +47,6 @@ void ResourceConverter::doConvertPcxToPng(const bfs::path & sourceFolder, bool d
 			if(!bfs::is_regular_file(directoryEntry))
 				continue;
 
-			std::string filePath = directoryEntry.path().string();
 			std::string fileStem = directoryEntry.path().stem().string();
 			std::string filenameLowerCase = boost::algorithm::to_lower_copy(filename.string());
 
--- a/mapeditor/resourceExtractor/ResourceConverter.cpp
+++ b/mapeditor/resourceExtractor/ResourceConverter.cpp
@@ -29,7 +29,7 @@ void ResourceConverter::convertExtractedResourceFiles(ConversionOptions conversi
 	std::vector<std::string> defFiles = { "TwCrPort.def", "CPRSMALL.def", "FlagPort.def", "ITPA.def", "ITPt.def", "Un32.def", "Un44.def" };
 
 	if(conversionOptions.splitDefs)
-		splitDefFiles(spritesPath, defFiles, conversionOptions.deleteOriginals);
+		splitDefFiles(defFiles, spritesPath, conversionOptions.deleteOriginals);
 
 	if(conversionOptions.convertPcxToPng)
 		doConvertPcxToPng(imagesPath, conversionOptions.deleteOriginals);
@@ -83,10 +83,10 @@ void ResourceConverter::splitDefFile(const std::string & fileName, const bfs::pa
 		logGlobal->error("Def File Split error! " + fileName);
 }
 
-void ResourceConverter::splitDefFiles(const bfs::path & sourceFolder, std::vector<std::string> defFileNames, bool deleteOriginals)
+void ResourceConverter::splitDefFiles(const std::vector<std::string> & defFileNames, const bfs::path & sourceFolder, bool deleteOriginals)
 {
 	logGlobal->info("Splitting Def Files from folder: %s ...\n", sourceFolder);
 
-	for(std::string defFilename : defFileNames)
+	for(const auto & defFilename : defFileNames)
 		splitDefFile(defFilename, sourceFolder, deleteOriginals);
 }
--- a/mapeditor/resourceExtractor/ResourceConverter.cpp
+++ b/mapeditor/resourceExtractor/ResourceConverter.cpp
@@ -24,64 +24,69 @@
 
 void ResourceConverter::convertExtractedResourceFiles(ConversionOptions conversionOptions)
 {
-	if (conversionOptions.splitDefs)
-		splitDefFiles(conversionOptions.deleteOriginals);
+	bfs::path spritesPath = VCMIDirs::get().userExtractedPath() / "SPRITES";
+	bfs::path imagesPath = VCMIDirs::get().userExtractedPath() / "IMAGES";
+	std::vector<std::string> defFiles = { "TwCrPort.def", "CPRSMALL.def", "FlagPort.def", "ITPA.def", "ITPt.def", "Un32.def", "Un44.def" };
 
-	if (conversionOptions.convertPcxToPng)
-		doConvertPcxToPng(conversionOptions.deleteOriginals);
+	if(conversionOptions.splitDefs)
+		splitDefFiles(spritesPath, defFiles, conversionOptions.deleteOriginals);
+
+	if(conversionOptions.convertPcxToPng)
+		doConvertPcxToPng(imagesPath, conversionOptions.deleteOriginals);
 }
 
-void ResourceConverter::doConvertPcxToPng(bool deleteOriginals)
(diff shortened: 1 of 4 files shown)


mapeditor/resourceExtractor/ResourceConverter.h
```diff
@@ -24,26 +24,34 @@ struct ExtractionOptions
...
-	// splits def files (TwCrPort, CPRSMALL, FlagPort, ITPA, ITPt, Un32 and Un44) so that faction resources are independent
-	// (town creature portraits, hero army creature portraits, adventure map dwellings, small town icons, big town icons, 
-	// hero speciality small icons, hero speciality large icons)
-	static void splitDefFiles(bool deleteOriginals);
-};
\ No newline at end of file
+	/** Splits the given .def files into individual images.
+	* For each .def file, the resulting images will be output in the same folder, in a subfolder (named just like the .def file)
+	*/
+	static void splitDefFiles(const bfs::path & sourceFolder, std::vector<std::string> defFileNames, bool deleteOriginals);
```
Nordsoft91, line 56: splitDefFile has arguments order: name, folder
splitDefFiles has:
folder, names
Why not unify order?

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -24,64 +24,69 @@
...
 		if(deleteOriginals)
-			bfs::remove(spritesPath / fileName);
+			bfs::remove(sourceFolder / fileName);
 	}
 	else
 		logGlobal->error("Def File Split error! " + fileName);
 }
 
-void ResourceConverter::splitDefFiles(bool deleteOriginals)
+void ResourceConverter::splitDefFiles(const bfs::path & sourceFolder, std::vector<std::string> defFileNames, bool deleteOriginals)
```
Nordsoft91, line 86:
```suggestion
void ResourceConverter::splitDefFiles(const bfs::path & sourceFolder, const std::vector<std::string> & defFileNames, bool deleteOriginals)
```

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -24,64 +24,69 @@
...
 {
-	bfs::path spritesPath = VCMIDirs::get().userExtractedPath() / "SPRITES";
+	logGlobal->info("Splitting Def Files from folder: %s ...\n", sourceFolder);
 
-	for(std::string defFilename : {"TwCrPort.def", "CPRSMALL.def", "FlagPort.def", "ITPA.def", "ITPt.def", "Un32.def", "Un44.def"})
-		splitDefFile(defFilename, spritesPath, deleteOriginals);
+	for(std::string defFilename : defFileNames)
```
Nordsoft91, line 90:
```suggestion
	for(const auto & defFilename : defFileNames)
```

mapeditor/resourceExtractor/ResourceConverter.h
nullkiller: Agreed. Would be better to make the order the same

mapeditor/resourceExtractor/ResourceConverter.cpp
krs0: Changed

mapeditor/resourceExtractor/ResourceConverter.cpp
krs0: Changed

mapeditor/resourceExtractor/ResourceConverter.h
krs0: Changed

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -24,64 +24,69 @@
...
-		const auto filename = dir_itr->path().filename();
+		const auto filename = directoryEntry.path().filename();
 		try
 		{
-			if (!bfs::is_regular_file(dir_itr->status()))
-				return;
+			if(!bfs::is_regular_file(directoryEntry))
+				continue;
 
+			std::string filePath = directoryEntry.path().string();
```
IvanSavenko, line 50: unused variable?

mapeditor/resourceExtractor/ResourceConverter.h
```diff
@@ -24,26 +24,34 @@ struct ExtractionOptions
...
 
-	// Splits def files that are shared between factions and converts pcx to bmp depending on Extraction Options
+	// Splits def files that are shared between factions and converts pcx to PNG depending on Extraction Options
 	static void convertExtractedResourceFiles(ConversionOptions conversionOptions);
 
 private:
 
-	// converts all pcx files from /Images into PNG
-	static void doConvertPcxToPng(bool deleteOriginals);
+	/** Converts all .pcx from extractedFolder/Images into .png */
```
IvanSavenko, line 47: We generally use single-line comments, e.g. `//` or triple slash `///` for documentation.
I think only file headers use multi-line comments for some reason

mapeditor/resourceExtractor/ResourceConverter.cpp
```diff
@@ -24,64 +24,69 @@
...
 
-void ResourceConverter::doConvertPcxToPng(bool deleteOriginals)
+void ResourceConverter::doConvertPcxToPng(const bfs::path & sourceFolder, bool deleteOriginals)
 {
-	bfs::path imagesPath = VCMIDirs::get().userExtractedPath() / "IMAGES";
-	bfs::directory_iterator end_iter;
+	logGlobal->info("Converting .pcx to .png from folder: %s ...\n", sourceFolder);
 
-	for(bfs::directory_iterator dir_itr(imagesPath); dir_itr != end_iter; ++dir_itr)
+	for(bfs::directory_entry & directoryEntry : bfs::directory_iterator(sourceFolder))
```
IvanSavenko, line 42: Use `const auto &`?

mapeditor/resourceExtractor/ResourceConverter.cpp
krs0: removed unused variable

mapeditor/resourceExtractor/ResourceConverter.h
krs0: got it. Using /// and // for comments

mapeditor/resourceExtractor/ResourceConverter.cpp
krs0: Using const auto &

mapeditor/resourceExtractor/ResourceConverter.h
```diff
@@ -24,26 +24,35 @@ struct ExtractionOptions
...
 
-	// Splits def files that are shared between factions and converts pcx to bmp depending on Extraction Options
+	// Splits def files that are shared between factions and converts pcx to PNG depending on Extraction Options
 	static void convertExtractedResourceFiles(ConversionOptions conversionOptions);
 
 private:
 
-	// converts all pcx files from /Images into PNG
-	static void doConvertPcxToPng(bool deleteOriginals);
+	// Converts all .pcx from extractedFolder/Images into .png
```
IvanSavenko, line 47: You could have used `///` here since this is documentation for a method. Nothing critical though.

IvanSavenko (approved):

mapeditor/resourceExtractor/ResourceConverter.h
krs0: You mean with summary and all ... or just ///?

IvanSavenko:
```suggestion
	/// Converts all .pcx from extractedFolder/Images into .png
```

Generally, preferred syntax for documenting stuff is this:
```
    /// This function does cool stuff
    void makeFoo()
```
- comment that starts with triple-slash (`///`). This comment should describe *what* function does (and not *why* or *how*).
- If necessary, add description for meaning of function parameters. 
- If comment is too long for a single line, split it into multiple line but keep triple-slash on all lines
- immediately after comment, with no empty lines - definition of described function

dydzio0614: shouldn't good code be self-documenting tho?

IvanSavenko: With extremely simple code - yes.
There is largely no need to document for example Point class
```
/// Construct point from two coordinates. Surprised?
Point( int x, int y)
```
But look here:
https://github.com/vcmi/vcmi/blob/develop/lib/spells/effects/Effect.h
It is well-written code, but without good description of *all* involved classes it is hard to understand what exactly the code does. Keep in mind, all comments there were added by me recently, not by original author.


For comments in code (e.g. inside methods) general rule is that code should be self-documenting, meaning, that *what* code does should be obvious without additional comments, but describing *why* code does it this way is still usually a good idea.

IvanSavenko: Also, @krs0 slight clarification:
Just `///`. We do use doxygen-style comments for documentation, but we rarely use doxygen-style *syntax*.

Personally don't see much point behind it especially since we don't use doxygen. 
However doxygen-style comments usually have different syntax highlight rules in IDE's, making them easily distinguishable from non-documentation comments.

PR #1453, opened 2023-01-17
Fix android UI shift when keyboard appears

nullkiller:
Now input will be always visible


kambala-decapitator:
now iOS is broken, seems the ifdef is no longer needed (but still keep the VCMI_APPLE ifdef)

kambala-decapitator:
I think desktop with zoom setting should also be tested

kambala-decapitator:
disabling iOS ifdef doesn't fix it unfortunately. For now please make sure that iOS code runs unmodified.

nullkiller:
@kambala-decapitator I changed code in a way that it should be logically same for iOS but have no iOS device. Could you check it?

kambala-decapitator (approved):
on iOS works good now

PR #1459, opened 2023-01-17
Translation support for mods

IvanSavenko:
Implemented translation support for mods:
- mods can now define list of per-language config files with translation
- vcmiclient will automatically select set of files based on language set in vcmilauncher
- names for bonuses are now integrated into translation
- old translate.json is now part of English translation of "vcmi" mod
- added Polish translation by @dydzio0614 
- added German translation by @heroesiiifan
- added Ukrainian translation by me, based on heavily outdated mod


dydzio0614:
Just a minor thing: launcher shows in console "no translator found" or something like that, when setting language to english

IvanSavenko:
> Just a minor thing: launcher shows in console "no translator found" or something like that, when setting language to english

Yeah, I saw it, came from "launcher translation" PR, so same in develop.
Should be safe to ignore, will remove it later.
English .ts file is basically a placeholder since actual translation is embedded into .ui. Probably Qt complains that it failed to load anything from it.

IvanSavenko:
Will merge tomorrow if there are no objections

PR #1460, opened 2023-01-18
Refactoring of SDL API usage

IvanSavenko:
- Moved Rect and Point classes into vcmi library
- Rect no longer inherits from SDL_Rect
- renamed vcmi functions that used SDL naming format to standard vcmi naming scheme
- added wrapper for some SDL functions in SDL_Extensions
- moved all functions in SDL_Extensions file into CSDL_Ext namespace
- SDL_Rect is now used only by SDL_Extensions, all client code uses Rect class
- changed some functions in SDL_Extensions to use better interface, e.g. reference instead of pointer, SDL_Color instead of int3.
- Recently added methods for segment-rect intersection are now in Rect class instead of BattleCallback

Note: for now don't report minor (e.g. code style) issues in SDL_Extensions file. This file needs quite a lot of work, which I want to do step-by-step, in next PR(s)


IvanSavenko:
I'd like to merge this one tomorrow.
Changes are spread through entire codebase, so very conflict-prone, but unlikely to cause issues.
If anyone want to postpone merge - let me know

PR #1461, opened 2023-01-18
SDL relative pointer for android

nullkiller:
Implement relative mode when you can drag pointer with finger. Kind of on-screen controls but invisible. This feature existed before in Android and was broken after SDL update so reimplemented it using SDL itself.

### DIFF
--- a/client/gui/CGuiHandler.cpp
+++ b/client/gui/CGuiHandler.cpp
@@ -268,12 +268,17 @@ void CGuiHandler::fakeMouseButtonEventRelativeMode(bool down, bool right)
 	sme.x = CCS->curh->xpos;
 	sme.y = CCS->curh->ypos;
 
-	int windowX, windowY;
+	float xScale, yScale;
+	int w, h, rLogicalWidth, rLogicalHeight;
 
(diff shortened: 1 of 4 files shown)


kambala-decapitator:
I think it makes sense to add it to iOS as well to have feature parity

nullkiller:
> I think it makes sense to add it to iOS as well to have feature parity

I moved #ifndef VCMI_IOS in a way that should allow relative mode but not 2-finger rclick. But it also needs support from the launcher.

client/gui/CGuiHandler.cpp
```diff
@@ -221,6 +228,63 @@ void convertTouch(SDL_Event * current)
...
+
+	sme.x = CCS->curh->xpos;
+	sme.y = CCS->curh->ypos;
+
+	float xScale, yScale;
+	int w, h, rLogicalWidth, rLogicalHeight;
+
+	SDL_GetWindowSize(mainWindow, &w, &h);
+    SDL_RenderGetLogicalSize(mainRenderer, &rLogicalWidth, &rLogicalHeight);
+    SDL_RenderGetScale(mainRenderer, &xScale, &yScale);
```
Nordsoft91, line 276: spaces

PR #1466, opened 2023-01-18
Translate adventure map objects

IvanSavenko:
Refactoring of object classes handler, to bring it in line with other handlers & to implement translation.

### DIFF
--- a/AI/VCAI/MapObjectsEvaluator.cpp
+++ b/AI/VCAI/MapObjectsEvaluator.cpp
@@ -32,10 +32,6 @@ MapObjectsEvaluator::MapObjectsEvaluator()
 				{
 					objectDatabase[CompoundMapObjectID(primaryID, secondaryID)] = handler->getAiValue().get();
 				}
-				else if(VLC->objtypeh->getObjGroupAiValue(primaryID) != boost::none) //if value is not initialized - fallback to default value for this object family if it exists
-				{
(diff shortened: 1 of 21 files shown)


AI/VCAI/MapObjectsEvaluator.cpp
```diff
@@ -32,10 +32,6 @@ MapObjectsEvaluator::MapObjectsEvaluator()
 				{
 					objectDatabase[CompoundMapObjectID(primaryID, secondaryID)] = handler->getAiValue().get();
 				}
-				else if(VLC->objtypeh->getObjGroupAiValue(primaryID) != boost::none) //if value is not initialized - fallback to default value for this object family if it exists
```
dydzio0614, line 35: why this got removed

IvanSavenko: AI value is now always in subtype

dydzio0614: but is there now a way to define default value for new subtypes added by mods?

IvanSavenko: Yes. Same way as for any other properties - "base" field.
See example in diff for `config/objects/moddables.json`

dydzio0614: is it something new? i thing this did not exist back in the day when I added ai value stuff

IvanSavenko: Nope. Always been here.

PR #1476, opened 2023-01-19
Do not consider corpses for movement range display

### LINKED ISSUE #1418: unit's movement range not displayed on hover
**Describe the bug**
On rare occasions hovering a unit doesn't display its movement range

**To Reproduce**
No reliable steps. However, based on situations it seems that standing near (or on top of) a corpse is involved.

**Expected behavior**
Unit's movement range is always displayed

**Actual behavior**
Sometimes unit's movement range isn't displayed

**Did it work earlier?**
First noticed while testing #1229, can't say if it occurs in 1.1

**Screenshots**
video of first time I stumbled on it: https://github.com/vcmi/vcmi/pull/1229#issuecomment-1371968280

reproduced once more with today's develop build:

https://user-images.githubusercontent.com/1557784/211804632-b5c486fc-e424-4020-866e-691baf4d21a7.mov

actual range of Behemoths:

Screenshot 2023-01-11 at 13 03 23

**Version**
 - OS: macOS Intel
 - Version: today's develop

**Additional context**
In the first video Swordsmen are standing next to Cavalier's corpse.

In the second one the 43 Giants are standing on a corpse of a Giant. There're also Basilisks behind Begemoths. If you'd like the [save game](https://github.com/vcmi/vcmi/files/10390627/Archive.tar.gz): attack Giants standing near the Library of Enlightenment with the first hero.

IvanSavenko:
Fixes #1418

### DIFF
--- a/client/battle/BattleFieldController.cpp
+++ b/client/battle/BattleFieldController.cpp
@@ -212,8 +212,8 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesStackRange()
 		result.insert(hex);
 
 	// display the movement shadow of the stack at b (i.e. stack under mouse)
-	const CStack * const shere = owner.curInt->cb->battleGetStackByPos(hoveredHex, false);
-	if(shere && shere != owner.stacksController->getActiveStack() && shere->alive())
+	const CStack * const shere = owner.curInt->cb->battleGetStackByPos(hoveredHex, true);
+	if(shere && shere != owner.stacksController->getActiveStack())
 	{
 		std::vector<BattleHex> v = owner.curInt->cb->battleGetAvailableHexes(shere, true, nullptr);
 		for(BattleHex hex : v)
--- a/client/battle/BattleFieldController.cpp
+++ b/client/battle/BattleFieldController.cpp
@@ -211,7 +211,7 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesStackRange()
 	for(BattleHex hex : set)
 		result.insert(hex);
 
-	// display the movement shadow of the stack at b (i.e. stack under mouse)
+	// display the movement shadow of stack under mouse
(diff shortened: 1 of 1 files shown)


client/battle/BattleFieldController.cpp
```diff
@@ -212,8 +212,8 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesStackRange()
 		result.insert(hex);
 
 	// display the movement shadow of the stack at b (i.e. stack under mouse)
-	const CStack * const shere = owner.curInt->cb->battleGetStackByPos(hoveredHex, false);
-	if(shere && shere != owner.stacksController->getActiveStack() && shere->alive())
+	const CStack * const shere = owner.curInt->cb->battleGetStackByPos(hoveredHex, true);
+	if(shere && shere != owner.stacksController->getActiveStack())
```
kambala-decapitator, line 216: what does `shere` mean? "stack here"?

IvanSavenko: Yes, "stack on currently hovered hex".
I usually replace such well-selected variable names, but missed this one somehow
(and that parameter that got changed from false to true is "select only alive anits")

dydzio0614: we should start from variables such as CSH, CCH, LOCPLINT, VLC.... xD

IvanSavenko: What's wrong with those?
CSH - CServerHandler
CCS - CClientState (not CCH by the way)
LOCPLINT - LOCal PLayer INTerface
VLC - VCMI Library Classes 

COMPLETELY obvious!
Seriously though, largest problem with them is not name, but rather their singleton status. Especially for VLC. Nothing good comes from global state. 
We could have easily implement (for example) single app mode without namespace hacks or "view enemy turn" option. But global state makes those much harder than it should be

PR #1478, opened 2023-01-20
limit displayed units number to 4 digits

### LINKED ISSUE #1448: unit numbers over 1000 are displayed in short form in battle
e.g. 1k instead of 1045, creature details shows full number though. I don't think it's wise to do so and it differs from oh3 behavior. Could be a toggleable option.

Screenshot 2023-01-16 at 14 32 06
Screenshot 2023-01-16 at 14 35 12

kambala-decapitator:
counts < 10000 are displayed as is

fixes #1448

test map: [unit-numbers.vmap.zip](https://github.com/vcmi/vcmi/files/10468748/unit-numbers.vmap.zip) (called "vcmi units number")

Screenshot 2023-01-20 at 20 36 58

Screenshot 2023-01-20 at 20 33 07

### DIFF
--- a/client/battle/BattleInterfaceClasses.cpp
+++ b/client/battle/BattleInterfaceClasses.cpp
@@ -850,7 +850,7 @@ void StackQueue::StackBox::setUnit(const battle::Unit * unit, size_t turn)
 		if (unit->unitType()->idNumber == CreatureID::ARROW_TOWERS)
 			icon->setFrame(owner->getSiegeShooterIconID(), 1);
 
-		amount->setText(CSDL_Ext::makeNumberShort(unit->getCount()));
+		amount->setText(CSDL_Ext::makeNumberShort(unit->getCount(), 4));
 
 		if(stateIcon)
 		{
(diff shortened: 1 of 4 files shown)


DjWarmonger:
Nice one :)

dydzio0614 (approved):
I believe currently existing function isn't broken so I will accept without manually testing... :P

Though I hope somebody gonna write unit test for this function at some point

Anyway, B for billions would be very cool addition too

kambala-decapitator:
> Anyway, B for billions would be very cool addition too

currently it uses G as in Gigabytes, there're also letters T (tera), P (peta) and E (exa) for even higher numbers

IvanSavenko:
> Anyway, B for billions would be very cool addition too

That method actually has enough prefixes to handle uint64_t, but it uses SI prefixes, so it is not "B" but "G" (Giga).
I think for creature numbers we use int32, so you'll only be able to see that prefix for experience

Laserlicht:
> > Anyway, B for billions would be very cool addition too
> 
> currently it uses G as in Gigabytes, there're also letters T (tera), P (peta) and E (exa) for even higher numbers

That should be left as it is, I think.
Otherwise we have to deal with the different scales: https://en.wikipedia.org/wiki/Long_and_short_scales

IvanSavenko (approved):

PR #1483, opened 2023-01-21
fix displaying growth bonus from Statue of Legion

### LINKED ISSUE #1410: Statue of Legion growth display issue
should be +50 **%**, but even better is to display flat value, which would be (1+1) * 50% = 1 in this case

Screenshot 2023-01-11 at 04 49 57

kambala-decapitator:
fixes #1410

test map: [legion-statue.h3m.zip](https://github.com/vcmi/vcmi/files/10472174/legion-statue.h3m.zip)

Screenshot 2023-01-21 at 15 30 20

### DIFF
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -553,24 +553,23 @@ GrowthInfo CGTownInstance::getGrowthInfo(int level) const
 		if(hasBuilt(BuildingID::HORDE_2))
 			ret.entries.push_back(GrowthInfo::Entry(subID, BuildingID::HORDE_2, creature->hordeGrowth));
 
-	int dwellingBonus = 0;
-	if(const PlayerState *p = cb->getPlayerState(tempOwner, false))
-	{
-		dwellingBonus = getDwellingBonus(creatures[level].second, p->dwellings);
-	}
-
-	if(dwellingBonus)
-		ret.entries.push_back(GrowthInfo::Entry(VLC->generaltexth->allTexts[591], dwellingBonus));// \nExternal dwellings %+d
+	//statue-of-legion-like bonus: % to base+castle
+	TConstBonusListPtr bonuses2 = getBonuses(Selector::type()(Bonus::CREATURE_GROWTH_PERCENT));
+	for(const auto & b : *bonuses2)
+		ret.entries.push_back(GrowthInfo::Entry(b->val * (base + castleBonus) / 100, b->Description()));
 
 	//other *-of-legion-like bonuses (%d to growth cumulative with grail)
 	TConstBonusListPtr bonuses = getBonuses(Selector::type()(Bonus::CREATURE_GROWTH).And(Selector::subtype()(level)));
 	for(const auto & b : *bonuses)
 		ret.entries.push_back(GrowthInfo::Entry(b->val, b->Description()));
 
-	//statue-of-legion-like bonus: % to base+castle
-	TConstBonusListPtr bonuses2 = getBonuses(Selector::type()(Bonus::CREATURE_GROWTH_PERCENT));
-	for(const auto & b : *bonuses2)
-		ret.entries.push_back(GrowthInfo::Entry(b->val * (base + castleBonus) / 100, b->Description()));
+	int dwellingBonus = 0;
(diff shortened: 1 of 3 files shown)


dydzio0614:
Maybe it would be decent idea to handle special case for bonus stringifying so you can provide custom value parameter, and if bonus is of percent type, it will process the value rather than using percent as value?

kambala-decapitator:
> Maybe it would be decent idea to handle special case in Description() for bonus stringifying so you can provide custom value parameter, and if bonus is of percent type, it will process the value and use that, rather than using percent as value?

to be honest I don't understand your proposal... could you open a separate issue for improvement?

dydzio0614:
Maybe, but I dont see it as any sort of priority

dydzio0614 (approved):

nullkiller:
In your example it does not count this external dwelling +2 when taking 50%?

kambala-decapitator:
> In your example it does not count this external dwelling +2 when taking 50%?

yes, this is correct behavior. Also checked in oh3.

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -570,7 +570,13 @@ GrowthInfo CGTownInstance::getGrowthInfo(int level) const
 	//statue-of-legion-like bonus: % to base+castle
 	TConstBonusListPtr bonuses2 = getBonuses(Selector::type()(Bonus::CREATURE_GROWTH_PERCENT));
 	for(const auto & b : *bonuses2)
-		ret.entries.push_back(GrowthInfo::Entry(b->val * (base + castleBonus) / 100, b->Description()));
+	{
+		// TODO: needs a better way to create string instead of this hack
+		const auto bonusValue = b->val;
```
nullkiller, line 575: I suggest to at least make a copy of bonus just in case. Modifing const reference of such a shared thing like a bonus might yield some parallelism issues

kambala-decapitator:
> Maybe it would be decent idea to handle special case in Description() for bonus stringifying so you can provide custom value parameter, and if bonus is of percent type, it will process the value and use that, rather than using percent as value?

now I got it :)

lib/mapObjects/CGTownInstance.cpp
nullkiller: @dydzio0614 suggests to modify b->Description() to handle this case internally. What if we introduce just another overload Description(int bonusValue) which will ignore original bonus value?

nullkiller: Or make a copy of bonus and modify value there so it is less reusable

kambala-decapitator: implemented per @dydzio0614 suggestion

dydzio0614:
I wonder if renaming argument to boost::optional<si32> flatValueHint and using it if bonus value type is PERCENT_TO_ALL or PERCENT_TO_BASE would be better - and in this case result would be flatValueHint * val / 100

kambala-decapitator:
> I wonder if renaming new Description() argument to boost::optional flatValueHint and using it if bonus value type is PERCENT_TO_ALL or PERCENT_TO_BASE would be better - and in this case result would be flatValueHint * val / 100, without percent calculation by caller

caller still needs to provide proper value to `GrowthInfo::Entry` constructor, as `GrowthInfo::totalGrowth()` uses it. Does it make sense for `Bonus` to provide computed values? Anyway, IMO such refactoring task is for 1.2.

DjWarmonger:
Shouldn't Status of Legion line be displayed **above** external dwellings? After all the Statue does not multiply the amount coming from dwellings, only base + castle growth.

While Grail does multiply the total amount generated by all other bonuses.

kambala-decapitator:
should I also adjust strings order in this PR? This is how it looks in oh3:
Screenshot 2023-01-23 at 14 46 03 Screenshot 2023-01-23 at 14 46 06

kambala-decapitator:
fixed order:
Screenshot 2023-01-24 at 19 42 47 Screenshot 2023-01-24 at 19 42 48

line break is there because original dwelling string has it

IvanSavenko (approved):
Perhaps we should either trim strings from excessive line endings?
Still, looks to be an improvement either way.

PR #1484, opened 2023-01-21
Shipyard dialog fixes

### LINKED ISSUE #1395: impossible to buy boat when a town is selected
**Describe the bug**
In oh3 it's possible to buy boat in a player-owned shipyard when a town is selected, in VCMI it's possible only when hero is selected.

**To Reproduce**
1. Capture shipyard on advmap
2. Select a town
3. Hover the shipyard to receive "boat cursor"
4. Click - nothing happens

**Expected behavior**
Purchase boat dialog opens

**Actual behavior**
nothing happens

**Did it work earlier?**
dunno

**Version**
 - OS: macOS Intel
 - Version: develop

**Additional context**
The dialog opens if you select a hero instead of town.

kambala-decapitator:
- fixes #1395
- Escape hotkey can dismiss the dialog now

### DIFF
--- a/client/gui/CursorHandler.cpp
+++ b/client/gui/CursorHandler.cpp
@@ -16,6 +16,10 @@
 #include "CAnimation.h"
 #include "../../lib/CConfigHandler.h"
 
+#ifdef VCMI_APPLE
+#include <dispatch/dispatch.h>
+#endif
+
 std::unique_ptr<ICursor> CursorHandler::createCursor()
 {
 	if (settings["video"]["cursor"].String() == "auto")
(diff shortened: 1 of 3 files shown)


client/gui/CursorHandler.cpp
```diff
@@ -401,10 +405,16 @@ CursorHardware::~CursorHardware()
 
 void CursorHardware::setVisible(bool on)
 {
+#ifdef VCMI_APPLE
+	dispatch_async(dispatch_get_main_queue(), ^{
```
IvanSavenko, line 409: Not an objection of any sort, but can you explain what is this and why it is needed?

nullkiller: Quite a lot of UI frameworks prohibit to call UI from non-main thread

IvanSavenko: Ah, this is wrapper that moves this call into main thread?
Sure, go ahead with commit, but I think this code can be moved to render() call which is always done in main thread.

nullkiller (approved):

client/gui/CursorHandler.cpp
kambala-decapitator: yes, SDL calls native API under the hood, and that one is supposed to be called from main thread (main "queue" is a synonym basically). And `^{ ... }` is objc "block" expression (= C++ lambda).

IvanSavenko (approved):

IvanSavenko:
@kambala-decapitator 
Regression - double-clicking town without Shipyard opens "This an abandoned shipyard" dialog.

kambala-decapitator:
@IvanSavenko nice find! Fixed in #1508

PR #1486, opened 2023-01-21
Move rewardable objects into config

IvanSavenko:
Documentation:
https://github.com/vcmi/vcmi/wiki/Modding-~-Objects-~-Rewardable

Changes:
- [x] All CGPickable objects are now fully configurable via json
- [x] Removed no longer used CGPickable class
- [x] Removed no longer use xtrainfo.txt
- [x] Allow bonuses as part of reward
- [x] Allow replacing text placeholders (e.g. `%d`) in messages
- [x] Remove CGBonusingObject object
- [x] Remove CGVisitableOPH object
- [x] Remove Tree of Knowledge
- [x] Remove Magic Well
- [x] Remove CGVisitableOPW object
- [x] Remove CGOnceVisitable object
- [x] Remove CGMagicSpring object
- [x] Check Shipwreck Survivor sound - develop & my branch have different values
- [x] Give an option to visit object from tile next to it (instead of on top of it)
- [x] Objects with "can refuse" flag and only 1 potential reward should show message from reward instead of onSelect
- [x] Allow "allOf"/"anyOf"/"noneOf" in limiter
- [x] Allow option to reroll rewards on reset
- [x] Clear visitors list on reset
- [x] Warrior's Tomb: no -3 to morale on revisiting the tomb
- [x] Tree of Knowledge component should say "+1 Level" instead of just "1"
- [x] Magic Spring (and probably Magic Well) should probably display "Mana" component

TODO's for future (will convert into issues):
- Warrior's Tomb: no confirmation to search the tomb (text ID = 161)
- Magic Spring should have separate reward for each visitable location
- Implement validation for input json
- Implement translation support for custom strings

Potential improvements:
- Allow "variables" for objects like Witch Hut (limiter & reward share the same random variable)
- Guardians that should be defeated before accessing rewards
- More options for "costs", e.g. taking artifacts (and probably other objects)
- Better/more flexible way to replace text placeholders


config/objects/rewardable.json
```diff
@@ -193,13 +214,38 @@
...
 				"index" : 0,
 				"aiValue" : 2000,
 				"rmg" : {
 					"value"		: 2000,
 					"rarity"	: 100
-				}
+				},
+				"visitMode" : "unlimited",
+				"selectMode" : "player",
+				"rewards" : [
```
DjWarmonger, line 227: One option is missing: https://heroes.thelazy.net/index.php/Flotsam

IvanSavenko: Ah, I see it, 25% for 5 wood. Will fix.
Current json is based on ancient version of our code, will recheck.

Alucard648:
Will it be possible to implement battles as part of object properties, like custom creature banks?

IvanSavenko:
> Will it be possible to implement battles as part of object properties, like custom creature banks?

Possible, but probably not for 1.2. For now my main goal is to support H3 objects, which don't have battles

IvanSavenko:
> Will it be possible to implement battles as part of object properties, like custom creature banks?

Also, @Alucard648 why do you need that?
I would rather leave these type of objects safe to access and keep "guarded rewards" for creature banks.

If you want such "guarded reward" why not use existing feature and create custom bank instead?

IvanSavenko:
Marking as ready.
Remaining issues are minor & already exist in develop.

lib/mapObjects/CRewardableConstructor.cpp
```diff
@@ -63,17 +63,26 @@ TRewardLimitersList CRandomRewardObjectInfo::configureSublimiters(CRewardableObj
...
 {
+	std::vector<SpellID> spells;
+	for (size_t i=0; i<6; i++)
+		IObjectInterface::cb->getAllowedSpells(spells, static_cast<ui16>(i));
+
+
 	limiter.dayOfWeek = JsonRandom::loadValue(source["dayOfWeek"], rng);
 	limiter.daysPassed = JsonRandom::loadValue(source["daysPassed"], rng);
-	limiter.minLevel = JsonRandom::loadValue(source["minLevel"], rng);
+	limiter.heroExperience = JsonRandom::loadValue(source["heroExperience"], rng);
```
DjWarmonger, line 73: Do we need a limiter for hero experience? Isn't the level enough?

...besides I don't think many people can tell from of their head ie. what level do you get for 100000 experience or vice versa.

IvanSavenko: Don't expect it to be used actively, mostly added for consistency with rewards.
Yes, checking for level should be preferred way.

config/objects/rewardableOncePerWeek.json
```diff
@@ -184,9 +184,14 @@
 				"selectMode" : "selectFirst",
 				"rewards" : [
 					{
+						"limiter" : { "daysPassed" : 8 },
```
DjWarmonger, line 187: Does it grant only first rewards and not both? What's the logic here?

DjWarmonger (approved):

PR #1492, opened 2023-01-22
Add missing entries for new "limited shooting range" bonus

dydzio0614:


krs0 (commented):
Maybe change 

> Cannot shoot targets beyond ${val} hexes away

To

> Cannot shoot targets that are more than ${val} hexes away

PR #1496, opened 2023-01-22
show resurrect message in the battle log

### LINKED ISSUE #957: amount of resurrected creatures isn't written to battle log
tested with Archangels

oh3 message:
Screenshot 2023-01-22 at 16 12 32

kambala-decapitator:
fixes #957

test map: [resurrect.h3m.zip](https://github.com/vcmi/vcmi/files/10475071/resurrect.h3m.zip) Hero has Archangels and Resurrect spell.

Screenshot 2023-01-22 at 20 34 29
Screenshot 2023-01-22 at 20 39 06

### DIFF
--- a/client/battle/BattleConstants.h
+++ b/client/battle/BattleConstants.h
@@ -79,7 +79,7 @@ enum class ECreatureAnimType
 
 	DEAD            = 22, // new group, used to show dead stacks. If empty - last frame from "DEATH" will be copied here
 	DEAD_RANGED     = 23, // new group, used to show dead stacks (if DEATH_RANGED was used). If empty - last frame from "DEATH_RANGED" will be copied here
-	RESURRECTION    = 24, // new group, used for animating resurrection, if empty - reversed "DEATH" animation will be copiend here
+	RESURRECTION    = 24, // new group, used for animating resurrection, if empty - reversed "DEATH" animation will be copied here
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
@IvanSavenko could you review as it touches battle logic?

IvanSavenko (commented):
Not exactly familiar with spell system, but at least Damage effect and Timed handle log differently. They call method `server->describeChanges()` to check whether log entries have to be generated and if it returns false - skip logging.

Also, it is a bit weird to have spell-related information "leak" into UnitState. Perhaps keep changes contained in Heal effect & compute heal amount/resurrected amount locally, by comparing old health value with new one?

nullkiller:
> Not exactly familiar with spell system, but at least Damage effect and Timed handle log differently. They call method `server->describeChanges()` to check whether log entries have to be generated and if it returns false - skip logging.
> 
> Also, it is a bit weird to have spell-related information "leak" into UnitState. Perhaps keep changes contained in Heal effect & compute heal amount/resurrected amount locally, by comparing old health value with new one?

I agree. Makes sense. Or return some data structure from Heal method describing healing result. { ResurectedCreaturesCount, HealthPointsRestored } and format it before adding to battle log.

kambala-decapitator:
thanks! adjusted.

IvanSavenko (approved):

PR #1498, opened 2023-01-22
WIP: Conan for repace MXE

rilian-la-te:
This patchset allows using Conan for building Windows binaries on Linux.

It is WIP due to requirement of building deps from source and requirement of testing this resulting builds.

I think when it will be finished, we can remove old MXE. For now, it builds and runs, but crashes after I try to start any campaign or scenario and I do not know why.

@IvanSavenko how to check this locally.

1. Take ubuntu 22.04 and clone this branch
2. Replace compiler.version in CI/conan/mingw-linux to 10
3. Run ` conan install . --install-folder=conan-generated           --no-imports           --build=missing           --profile:build=default           --profile:host=CI/conan/mingw-linux           --options with_apple_system_libs=True -s build_type=Release` in cloned dir
4. Run `cmake --build --preset windows-mingw-conan-linux --target install` to build and install VCMI
5. You can also run CPack as usually, bundle should be created correctly.


IvanSavenko:
Hi. Thanks for attempting to do that, will check it out in a day or two once I have time.

rilian-la-te:
@IvanSavenko @kambala-decapitator CI seems to working. Can you test my builds on actual Windows and help me to figure crash origin? Wine is not very helpful on this regard(

rilian-la-te:
It seems than I figured a source of crash of Conan VCMI builds. Unfortunately, this is a bug in Conan recipe and will require a static libpng for all Conan builds. 

@kambala-decapitator do you use Conan for your Mac builds of VCMI? Can you try a Conan + Mac VCMI to confirm or drop my hypothesis than sdl_image with dynamic libpng in Conan is buggy?

rilian-la-te:
@IvanSavenko  @kambala-decapitator I will update deps repo tomorrow, and then PR will becomes ready.

For now, it seems playable on ArchLinux Wine, and when I will update my deps repo for Ubuntu, it will be playable on Ubuntu too.

kambala-decapitator:
> @kambala-decapitator do you use Conan for your Mac builds of VCMI? Can you try a Conan + Mac VCMI to confirm or drop my hypothesis than sdl_image with dynamic libpng in Conan is buggy?

sounds like windows issue as dynamic libpng on macOS works fine: https://github.com/vcmi/vcmi/blob/1236ed6d02e083ec74a3de0c3e6c1020f73a56e6/conanfile.py#L40

rilian-la-te:
This PR is ready for test and review, all CI is passing) @IvanSavenko @kambala-decapitator

IvanSavenko:
Unable to start on Windows 10 system.
Only VCMI_client is affected, all other vcmi executables work fine
зображення

Text version:
The procedure entry point ProcessSocketNotifications could not be located in the dynamic link library S:\Games\vcmi\mingw\3994539044-vcmi-PR-1498-dcda698\VCMI_client.exe. 

Probably some missing lib

IvanSavenko:
Also, VCMI_Client.exe is 128 Mb in size? 
For comparison, MXE has 8 Mb .exe, msvc's exe is just 4 Mb.

client/gui/CGuiHandler.cpp
```diff
@@ -192,7 +192,7 @@ void CGuiHandler::handleEvents()
 	while(!SDLEventsQueue.empty())
 	{
 		continueEventHandling = true;
-		SDL_Event ev = SDLEventsQueue.front();
+		SDL_Event& ev = SDLEventsQueue.front();
```
IvanSavenko, line 195: Change of behavior - after `SDLEventsQueue.pop` variable `ev` will now point to different event since it is now a reference.

Yes, I agree that code is at least weird, if not outright broken, but changed version is very likely to be broken as well and probably even worse than before.

If this is necessary for build - can you at least change it to static? Not sure whether this will work though sice I don't know what initial warning was.
```suggestion
		static SDL_Event ev = SDLEventsQueue.front();

```

lib/filesystem/CFilesystemLoader.cpp
```diff
@@ -88,7 +88,7 @@ bool CFilesystemLoader::createResource(std::string filename, bool update)
 
 	if (!update)
 	{
-		if (!FileStream::CreateFile(baseDirectory / filename))
+		if (!FileStream::FileCreate(baseDirectory / filename))
```
IvanSavenko, line 91: Ah, hello, yet another macro from Windows.h breaking our builds.
I would rather rename it to createFile, since we use camelCase for methods, but no big deal

client/CMakeLists.txt
```diff
@@ -283,6 +283,11 @@ target_link_libraries(vcmiclient PRIVATE
 )
 
 if(ffmpeg_LIBRARIES)
+	if(USING_CONAN AND MINGW)
+		target_link_options(vcmiclient PRIVATE
```
IvanSavenko, line 287: Can you explain (in comment in cmake) what is this and why is it needed?

AI/CMakeLists.txt
```diff
@@ -27,6 +32,9 @@ if(NOT fuzzylite_FOUND)
     set(FL_BUILD_BINARY OFF CACHE BOOL "")
     set(FL_BUILD_SHARED OFF CACHE BOOL "")
 	set(FL_BUILD_TESTS OFF CACHE BOOL "")
+	if("x${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "xGNU" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU")
+		add_compile_options(-Wno-error=deprecated-declarations)
```
IvanSavenko, line 36: I would rather fix warnings, but OK, can be fixed later in separate PR

rilian-la-te: We will fix warnings in FuzzyLite? I disabled it because it is FuzzyLite header warnings, not VCMI one.

IvanSavenko: If this is about fuzzylite then better to add comment about that.
But you also have one in library makefile.
Anyway, right now goal should be making final build work on Windows, warning can be dealt with later

client/gui/CGuiHandler.cpp
rilian-la-te: @IvanSavenko done static

rilian-la-te:
>Update: According to msdn, this method is only available on Windows 10 Build 20348 and later. Apparently, I am on older build (and likely - same goes for many of our players). I think you're somehow building against newest Windows SDK & have not configured targeting older systems.

Removed whole-archive, it should help, please, try new executables.

client/CMakeLists.txt
rilian-la-te: Removed this and replaced by another, more correct workaround.

AI/CMakeLists.txt
rilian-la-te: @IvanSavenko It is about FARPROC, and this is explained in commit message.

rilian-la-te:
@IvanSavenko fixed all but size (this is debuginfo executable, hence huge size. Stripped one will nave 27Mb size, which is huge but not that huge).

lib/filesystem/CFilesystemLoader.cpp
rilian-la-te: Fixed.

kambala-decapitator:
re: size

I guess because client-only libs are built as static (all SDL stuff, FFmpeg).

rilian-la-te:
@IvanSavenko confirmed no input. static SDL_event causes this. With SDL_Event all input was correct in Wine. Ok, I will just set Wno-error there.

@kambala-decapitator Yes, I also think than main cause is ffmpeg (libavcodec.a is 100 mb in size)

CMakeLists.txt
```diff
@@ -544,9 +562,19 @@ endif()
 
 
 if(WIN32)
-	file(GLOB dep_files
-		${dep_files}
-		"${CMAKE_FIND_ROOT_PATH}/bin/*.dll")
+    #Conan imports enabled
```
kambala-decapitator, line 565: please fix indentation

conanfile.py
```diff
@@ -165,6 +173,9 @@ def _disableQtOptions(disableFlag, options):
     def requirements(self):
         # TODO: will no longer be needed after merging https://github.com/conan-io/conan-center-index/pull/13399
         self.requires("libpng/[~1.6.38]", override=True) # freetype / Qt
+        if self.settings.os == "Windows":
+            self.requires("zlib/[~1.2.13]", override=True) # minizip / Qt
+            self.requires("libiconv/[~1.17]", override=True) # ffmpeg / sdl
```
kambala-decapitator, line 178: not sure, but aren't those versions already used by default? or something still requires older/newer?

CI/conan/mingw-linux
```diff
@@ -0,0 +1,36 @@
+toolchain=/usr/i686-w64-mingw32
+target_host=i686-w64-mingw32
+
+[settings]
+
+# We are cross-building to Windows
+os=Windows
+arch=x86
```
IvanSavenko, line 8: Also - why x86?
It looks that our mxe build is 32-bit but don't really see any reason to keep using it.

IvanSavenko: MSVC build is 64-bit and seems to be stable, so no reason not to use 64-bit builds for Windows

conanfile.py
rilian-la-te: @kambala-decapitator because there is conflicts - minizip wants 1.2.12 and Qt wants 1.2.13,

for second - sdl wants 1.17 and ffmpeg wants 1.16

CI/conan/mingw-linux
rilian-la-te: @IvanSavenko yes, 64bit is stable, but I think than 32-bit windows is more used in CIS, especially in old PCs (with Windows XP or Windows 7 onboard). So, what if MSVC will remain 64bit, and Conan can be both 32bit and 64 bit?

conanfile.py
kambala-decapitator: clear, thanks. Then please move these 2 lines to `else` branch of `use Apple system libraries instead of external ones`

CI/conan/mingw-linux
IvanSavenko: Not really. Maybe some ancient embedded systems, but all players use x64:
https://store.steampowered.com/hwsurvey
0.11% Windows 7 (32-bit)

We also had plans to use msvc as primary build (and one we distribute as release), which is 64-bit.

kambala-decapitator: 64-bit only sounds fine. If there's a request from community, we can make 32-bit ones as well.

rilian-la-te: @IvanSavenko @kambala-decapitator okay, I will regenerate to 64-bit. And I will add strip flag to CPACK for Release builds. It is okay?

IvanSavenko: OK.
Strip flag - yes, for "Release" build, keep debug info for "RelWithDebugInfo" or what's its name

kambala-decapitator: sounds good. let's still keep 32-bit deps in GitHub release in case someone wants to make a local 32-bit build with mingw.

rilian-la-te:
> Can you revert it and disable whatever warning was in there?

What about make it as
```c++
		auto ev = SDLEventsQueue.front();
```

It will have behaviour change? Warning will be silenced and input works.

CMakeLists.txt
rilian-la-te: done

IvanSavenko:
> It will have behaviour change? Warning will be silenced and input works.

Should work I guess? Strange that it has effect on warning.
Anyway, don't think too much about it. That code (specifically `current`) variable is bad design and should have been removed long ago. Will probably do it once this one is merged.

conanfile.py
rilian-la-te: done

AI/CMakeLists.txt
rilian-la-te: done

rilian-la-te:
> Should work I guess? Strange that it has effect on warning.

Checked in Wine, it works and silences a warning. Anyway, I will polish it and then set 64-bit build.

CI/conan/mingw-linux
rilian-la-te: Done both strip and 64bit.

IvanSavenko:
Build seems to be working fine now.

Packages size - also fine, wasn't aware of statically linked client libs, that explains the difference. Total size is also smaller than mxe and in line with msvc.

Now, I am trying to follow your instructions and to build libraries, however there is some issue with TBB:
https://gist.github.com/IvanSavenko/293b96f4c1ebf0d409dbdd10d8dfe041
Any ideas?

Also:
1) Your instruction skips conan install
2) I also had to install `gcc-mingw-w64-x86-64` and `g++-mingw-w64-x86-64` packages - othervice, conan can't find cross-compiler.

rilian-la-te:
@IvanSavenko you need to do what I did in CI/before-install/mingw-ubuntu.sh

Namely:

```sh
sudo apt-get update
sudo apt-get install ninja-build mingw-w64 nsis
sudo update-alternatives --set x86_64-w64-mingw32-g++ /usr/bin/x86_64-w64-mingw32-g++-posix

curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-common_10.0.0-2_all.deb && sudo dpkg -i mingw-w64-common_10.0.0-2_all.deb;
curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-x86-64-dev_10.0.0-2_all.deb && sudo dpkg -i mingw-w64-x86-64-dev_10.0.0-2_all.deb;
```

It seems you skipped step 3, do it.

rilian-la-te:
@IvanSavenko conan install I did exactly as in CI, namely 

```
        conan install . \
          --install-folder=conan-generated \
          --no-imports \
          --build=missing \
          --profile:build=default \
          --profile:host=CI/conan/mingw-linux
```

rilian-la-te:
Where I should put an instructions?

IvanSavenko:
@rilian-la-te 
I mean installation of conan itself. I don't use it so didn't have it myself.
Nothing hard to figure out, but I'd rather have complete instructions.

Anyway, running commands from mingw-ubuntu.sh, will see how it goes.

> Where I should put instructions?

For now - here. Will move to dependencies repository once I am done.

rilian-la-te:
@IvanSavenko I already generated all dependencies in my repo.

Here is the link: https://github.com/rilian-la-te/vcmi-deps-windows-conan/releases/tag/0.0.1

IvanSavenko:
And TBB passed, continuing build...

IvanSavenko:
> @IvanSavenko I already generated all dependencies in my repo.
> 
> Here is the link: https://github.com/rilian-la-te/vcmi-deps-windows-conan/releases/tag/0.0.1

Yeah, I saw them. Want to build same package myself, to ensure that we have way to update this package in future

IvanSavenko:
Also, I had this message on the first start:
```
************************* WARNING: GCC OLD ABI COMPATIBILITY ***********************
 
Conan detected a GCC version > 5 but has adjusted the 'compiler.libcxx' setting to
'libstdc++' for backwards compatibility.
Your compiler is likely using the new CXX11 ABI by default (libstdc++11).

If you want Conan to use the new ABI for the default profile, run:

    $ conan profile update settings.compiler.libcxx=libstdc++11 default

Or edit '/home/ivan/.conan/profiles/default' and set compiler.libcxx=libstdc++11

************************************************************************************
```
I ran that command myself, but perhaps it should be moved to `CI/conan/mingw-linux`?
(just asking - no idea what approach is correct here)

IvanSavenko:
Oh, it is already there. Not sure why I got that message then

rilian-la-te:
@IvanSavenko because your default **linux** profile does not contain this string. It is not needed for compiling for Windows anyway.

kambala-decapitator:
> Where I should put instructions?

[docs/conan.md](docs/conan.md), the link is in the readme

IvanSavenko:
@rilian-la-te 
1) What do I need to do to create package similar to yours? I see that my ~/.conan/data is more or less same as yours, but I see some extra (temporary?) files. Any proper way to export prebuilds?

2) attempting to build vcmi via CMake leads to:
```
$ cmake --version
cmake version 3.22.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).
$ cmake --build --preset windows-mingw-conan-linux --target install
CMake Error: Could not read presets from /home/ivan/source/vcmi-conan: Unrecognized "version" field
```
Any ideas?

rilian-la-te:
> @rilian-la-te
> 
>     1. What do I need to do to create package similar to yours? I see that my ~/.conan/data is more or less same as yours, but I see some extra (temporary?) files. Any proper way to export prebuilds?
> 
>     2. attempting to build vcmi via CMake leads to:
> 
> 
> ```
> $ cmake --version
> cmake version 3.22.1
> 
> CMake suite maintained and supported by Kitware (kitware.com/cmake).
> $ cmake --build --preset windows-mingw-conan-linux --target install
> CMake Error: Could not read presets from /home/ivan/source/vcmi-conan: Unrecognized "version" field
> ```
> 
> Any ideas?

1. Remove CMakeUserPresets.json
2. Revert CMakePresets.json to version from PR without local modifications
3. Try again and post here if something is not builds. 

How to generate reps tarball - instructions are clearly in repo with my deps. There is a complete instruction how to generate a tarball from pristine data folder. Pristine means than data folder was used only to build deps of VCMI and nothing else.

CI/mingw-ubuntu/before_install.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+sudo apt-get update
+sudo apt-get install ninja-build mingw-w64 nsis
+sudo update-alternatives --set x86_64-w64-mingw32-g++ /usr/bin/x86_64-w64-mingw32-g++-posix
+
+curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-common_10.0.0-2_all.deb && sudo dpkg -i mingw-w64-common_10.0.0-2_all.deb;
+curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-x86-64-dev_10.0.0-2_all.deb && sudo dpkg -i mingw-w64-x86-64-dev_10.0.0-2_all.deb;
```
kambala-decapitator, line 8:
```suggestion
curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-common_10.0.0-2_all.deb \
  && sudo dpkg -i mingw-w64-common_10.0.0-2_all.deb
curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-x86-64-dev_10.0.0-2_all.deb \
  && sudo dpkg -i mingw-w64-x86-64-dev_10.0.0-2_all.deb
```

CMakeLists.txt
```diff
@@ -328,6 +329,24 @@ if(ENABLE_LUA)
...
+if(USING_CONAN AND (MINGW AND CMAKE_HOST_UNIX))
+	# Hack for workaround https://github.com/conan-io/conan-center-index/issues/15405
+	# Remove once it will be fixed
+	execute_process(COMMAND
+		bash -c "grep -rl Mf ${CONAN_INSTALL_FOLDER} | xargs sed -i 's/Mf/mf/g'"
+	)
+	# Hack for workaround ffmpeg broken linking (conan ffmpeg forgots to link to ws2_32)
+	# Remove once it will be fixed
+	execute_process(COMMAND
+		bash -c "grep -rl secur32 ${CONAN_INSTALL_FOLDER} | xargs sed -i 's/secur32/secur32 ws2_32/g'"
```
kambala-decapitator, line 341: I guess this will be executed each time you run cmake configuration? then it will "ruin" Conan config as on second run you'll receive `secur32 ws2_32 ws2_32`. Usually specifying the same library for linking multiple times is not an issue, hope it won't break anything.

conanfile.py
```diff
@@ -2,6 +2,7 @@
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import is_apple_os
 from conan.tools.cmake import CMakeDeps, CMakeToolchain
+from conan.tools.build import cross_building
```
kambala-decapitator, line 5: please move one line above to maintain alphabetical order

conanfile.py
```diff
@@ -184,6 +192,9 @@ def requirements(self):
             ]
             for lib in systemLibsOverrides:
                 self.requires(f"{lib}@vcmi/apple", override=True)
+        elif self.settings.os == "Windows":
```
kambala-decapitator, line 195: no need to specify explicit OS here:
```suggestion
        else:
```

kambala-decapitator (commented):
btw a trick that I use to keep VCMI dependencies separate from my main `~/.conan` directory is `CONAN_USER_HOME` env var which allows redirecting the conan home directory. When `CONAN_USER_HOME` is not specified, it defaults to user's home `~`. So your custom directory should contain `.conan` directory inside with the same layout as the main one, I use symlinks for all files/dirs except the `data` dir. https://docs.conan.io/en/latest/reference/env_vars.html#conan-user-home

```
> ll ~/dev/vcmi/conan-deps-macos/.conan
total 0
lrwxr-xr-x   1 kambala  staff    42B Oct 15 13:03 artifacts.properties@ -> /Users/kambala/.conan/artifacts.properties
lrwxr-xr-x   1 kambala  staff    32B Oct 15 13:03 cacert.pem@ -> /Users/kambala/.conan/cacert.pem
lrwxr-xr-x   1 kambala  staff    32B Oct 15 13:03 conan.conf@ -> /Users/kambala/.conan/conan.conf
lrwxr-xr-x   1 kambala  staff    41B Oct 15 13:03 config_install.json@ -> /Users/kambala/.conan/config_install.json
drwxr-xr-x  30 kambala  staff   960B Jan 17 13:23 data/
lrwxr-xr-x   1 kambala  staff    33B Dec 31 16:46 global.conf@ -> /Users/kambala/.conan/global.conf
lrwxr-xr-x   1 kambala  staff    27B Oct 15 13:03 hooks@ -> /Users/kambala/.conan/hooks
lrwxr-xr-x   1 kambala  staff    30B Oct 15 13:03 profiles@ -> /Users/kambala/.conan/profiles
lrwxr-xr-x   1 kambala  staff    34B Oct 15 13:03 remotes.json@ -> /Users/kambala/.conan/remotes.json
lrwxr-xr-x   1 kambala  staff    34B Oct 15 13:03 settings.yml@ -> /Users/kambala/.conan/settings.yml
lrwxr-xr-x   1 kambala  staff    33B Oct 15 13:03 version.txt@ -> /Users/kambala/.conan/version.txt
```

CMakeLists.txt
rilian-la-te: I tried it 3 times, cmake is smart enough to deduplicate libraries. But I can change sed line to 2 sed lines, which will be run only once.

rilian-la-te: Replace sed line with 2 lines without side effect

conanfile.py
rilian-la-te: done

conanfile.py
rilian-la-te: done

CI/mingw-ubuntu/before_install.sh
rilian-la-te: done

rilian-la-te:
@IvanSavenko @kambala-decapitator what is a status of a PR?

IvanSavenko:
Looks OK to me, if @kambala-decapitator approves then I'm OK with merge. Not familiar with conan to review it.
Will try to finish my own dependencies pack later today, but it is not required for merge.

docs/conan.md
```diff
@@ -50,6 +54,7 @@ The highlighted parts can be adjusted:
 - ***conan-generated***: directory (absolute or relative) where the generated files will appear. This value is used in CMake presets from VCMI, but you can actually use any directory and override it in your local CMake presets.
 - ***never***: use this value to avoid building any dependency from source. You can also use `missing` to build recipes, that are not present in your local cache, from source.
 - ***CI/conan/PROFILE***: if you want to consume our prebuilt binaries, ***PROFILE*** must be replaced with one of filenames from our [Conan profiles directory](../CI/conan) (determining the right file should be straight-forward). Otherwise, either select one of our profiles or replace ***CI/conan/PROFILE*** with `default` (your default profile).
+- ***note for Windows x86***: we do not have prebuilt profile for Windows x86, but if you replace ***arch=x86_64*** to ***arch=x86*** and ***target_host=x86_64-w64-mingw32*** to ***target_host=i686-w64-mingw32*** then it will be working profile for 32-bit windows.
```
kambala-decapitator, line 57: there's no need to edit profile, it's possible to customise settings on the command line: should be `-s arch=x86` if I'm not mistaken. Please check if `target_host` can also be overridden from command line.

alternatively, create a "base" profile with common parts that will be included into dedicated 32- and 64-bit ones, see Apple profiles for an example.

and adjust the example in docs accordingly.

kambala-decapitator (commented):
still some open points:

1. deps should be moved to the main VCMI deps repo
2. should we already sunset MXE in favour of this new approach?
3. I have installed the latest artifact on windows 10: by default, it was offered to install to Program Files **(x86)** - I thought we now build 64-bit version?
4. (not a blocker) I have Windows N version that doesn't include MS Media Foundation (it can still be installed manually). Now I can't launch client due to MFPlat.dll being missing. Is it possible to build FFmpeg in a way that doesn't require this library?

IvanSavenko:
> deps should be moved to the main VCMI deps repo

Will try to do that later today, either with my own build or (if any issues arise) - with existing build from this branch.

> should we already sunset MXE in favour of this new approach?

AFAIK we still use MXE for daily builds that are uploaded on vcmi.eu. Sure, we can switch them to msvc/mingw but this will raise question on what should we do with beta branch.
Personally, I think we should release 1.1.1 with current mxe build to avoid any unexpected issues, and after that - remove mxe.

Meaning, that this PR should not remove mxe.

> I have installed the latest artifact on windows 10: by default, it was offered to install to Program Files (x86) - I thought we now build 64-bit version?

From what I see, mingw artifact is 64-bit exe. Looks like NSIS installer picks incorrect install path.

kambala-decapitator:
> AFAIK we still use MXE for daily builds that are uploaded on vcmi.eu. Sure, we can switch them to msvc/mingw but this will raise question on what should we do with beta branch. Personally, I think we should release 1.1.1 with current mxe build to avoid any unexpected issues, and after that - remove mxe.
> 
> Meaning, that this PR should not remove mxe.

the PR targets develop, we can remove MXE from CI here. 1.1.1 would still have it and produce "old" builds.

but we can also keep both in develop during some kind of "transition period", just switch server uploads to use the new binaries. It would still be possible to download MXE builds from the Actions page.

IvanSavenko:
> the PR targets develop, we can remove MXE from CI here. 1.1.1 would still have it and produce "old" builds.
> 
> but we can also keep both in develop during some kind of "transition period", just switch server uploads to use the new binaries. It would still be possible to download MXE builds from the Actions page.

Not aware how builds get uploaded to vcmi server, but if we can upload msvc builds for new branches and keep mxe for old branches, e.g. beta then sure. In this case removing mxe makes sense.

rilian-la-te:
@IvanSavenko I will fix 32/64 bit decision for NSIS later today.

rilian-la-te:
@IvanSavenko  @kambala-decapitator 
Install directory is fixed.

docs/conan.md
rilian-la-te: @kambala-decapitator
1. I cannot move deps to main repo (no write rights in VCMI repo), if you move it, I will point script and doc to new location.
4. No, it is not possible, ffmpeg unconditionally required mfplat for Windows.

target_host cannot be overriden from command line in current context. Base profile is also cannot be created, because $target_host cannot be passed into included profile. I can try to make a profile as Jinja2 template, but later today.

rilian-la-te: @kambala-decapitator on Conan. I created an issue in CCI for adding option to building FFmpeg without mediafoundation.

IvanSavenko:
@rilian-la-te 
VCMI repository for prebuilts is ready. Can you
1) Check instructions here: https://github.com/vcmi/vcmi-deps-windows-conan/blob/1.0/README.md Ideally, they should work for newly installed system without any dev environment.
2) Try to use my package for dependencies: https://github.com/vcmi/vcmi-deps-windows-conan/releases/tag/1.0

docs/conan.md
rilian-la-te: Done 2 profiles and doc changes accordingly

rilian-la-te:
@IvanSavenko this instructions is for old version.

I will create a PR with correct instructions soon.

IvanSavenko:
Yeah, because your change didn't really fixed anything but silenced that warning somehow. Probably gcc-10 failed to correctly diagnose your code due to `auto`, but gcc-12 correctly reported that code.
Feel free to disable that warning/move it to non-error.

I'll make PR with proper fix soon, either today or tomorrow. Wanted to refactor that place soon anyway.

rilian-la-te:
@IvanSavenko downloaded artifacts from CI, build seems to starting and I can even make 2 turns in Wine.
I think your deps is working, I am deleting my repo.

IvanSavenko:
Great!
MXE can be removed later, in separate PR, so up to you I guess?
@kambala-decapitator anything else that needs to be addressed?

rilian-la-te:
@IvanSavenko yes, I think, MXE time comes later, not in this PR.

kambala-decapitator:
@IvanSavenko please also upload 32-bit deps

@rilian-la-te I've tried to consume the deps on Windows, but some packages are missing:

ERROR: Missing binary: boost/1.81.0:488003ed85412e8b91e9ab687edcd96f29e94252
ERROR: Missing binary: libpng/1.6.39:5f6f7daebf31b4602de474524655005906797f88
ERROR: Missing binary: qt/5.15.8:0cee88fea073ecabb4d8784940f578a76ce70382

any idea why? my profile matches the one used to build the deps (and most deps were found in the cache):

```
Configuration (profile_host):
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=11
compiler.libcxx=libstdc++11
compiler.version=10
os=Windows
os_build=Windows
[options]
[build_requires]
[env]
[conf]
tools.cmake.cmaketoolchain:generator=Ninja

Configuration (profile_build):
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=11
compiler.libcxx=libstdc++11
compiler.version=10
os=Windows
os_build=Windows
[options]
[build_requires]
[env]
[conf]
tools.cmake.cmaketoolchain:generator=Ninja
```

the command I used:

> conan install . --install-folder=conan-mingw --no-imports --build=never --profile:build=mingw --profile:host=mingw

(ignore the boost being missing, for some reason Conan simply deleted the dir with binaries, investigating why)

rilian-la-te:
@kambala-decapitator you are on Windows or on Linux? I tried it only on Linux. Build profile suggests than you are on Windows, and then you cannot deps, (because conanfile checks cross_building(self) and disables OpenGL in cross-build environment), so, you now build Qt with OpenGL. libpng - please, look into folder, it should be static.

kambala-decapitator:
sorry for being dumb - I forgot to apply conanfile changes from the PR :) now libpng is found.

Qt - thanks, although that's a bit unfortunate. Maybe there will be a solution in the future.

kambala-decapitator (approved):
PR looks good to me

rilian-la-te:
@kambala-decapitator there is an issue in Conan recipe, which requires opengl/system to enable OpenGL even on cross environment, so, it need a workaround now.

kambala-decapitator:
then seems that native Windows mingw build is still desirable as using it in WSL sounds a bit exotic to me

also would be interesting to compare build speed of VCMI on native mingw vs cross-compiling

rilian-la-te:
@kambala-decapitator I do not have Windows at all, so, I cannot measure speed and test MinGW outside of CI. We can just request to set -o qt:opengl=no while using our deps in Windows mingw, I do not think than one additional option for native compilation is bad.

For me using Linux is cruical. Actually I using it for playing too)

IvanSavenko:
@kambala-decapitator 
Will make 32-bit build later today.

kambala-decapitator:
> I cannot measure speed and test MinGW outside of CI
that's a task for the future and for CI :)

> We can just request to set -o qt:opengl=no while using our deps in Windows mingw, I do not think than one additional option for native compilation is bad.
I've already tried using that option (by removing `and cross_building(self)`) but unfortunately existing binary package is still considered incompatible. In the future I'd like to have prebuilt binaries fully compatible with native mingw for windows devs that don't wish to use msvc compiler.

btw tested the latest artifact: all good now.

@rilian-la-te do you wish to wait with PR merge until some of your issues reported to CCI are fixed?

rilian-la-te:
@kambala-decapitator I do not think than waiting for merge is reasonable. I think if we merge it ASAP, then we can remove MXE and bump CMake version at least to 3.15 for no-preset build and to 3.25 for preset build. I wish to be able to create RPM and DEB artifacts by CI itself for someone who want VCMI newer than in repo)

rilian-la-te:
@kambala-decapitator I want to merge it ASAP, I included workarounds to all CCI issues. We can just remove it after issues is fixed.

IvanSavenko:
Then I'll merge this PR later today, once I get 32-bit prebuilts package done.
(unless there will be objections)

Issue #806, opened 2022-09-03
[feature] improve marking turn order of creatures


kambala-decapitator:
for example, on this SS it's unclear which Naga is next. Knowing this would let me kill the one whose turn is next and grant next turn to Master Gremlins (who will suffer from alive Naga otherwise, most probably).

for example, highlight creature on mouse hovering its portrait.

![Screenshot 2022-09-03 at 13 51 09](https://user-images.githubusercontent.com/1557784/188280781-0be9bf82-99b6-4909-820a-0bb26e73d939.png)

kambala-decapitator:
> for example, highlight creature on mouse hovering its portrait.

I see it's already implemented like that in the HD mod

rilian-la-te:
@kambala-decapitator is this should be closed as completed via #1499?

kambala-decapitator:
@rilian-la-te yes, thanks for noticing!

PR #1499, opened 2023-01-23
Battle queue improvement

dydzio0614:
Implements #806 

obraz


IvanSavenko:
Will look into it, but a bit later, today or tomorrow.

lib/battle/CBattleInfoEssentials.cpp
```diff
@@ -185,6 +185,21 @@ const CStack* CBattleInfoEssentials::battleGetStackByID(int ID, bool onlyAlive)
 		return stacks[0];
 }
 
+const CStack* CBattleInfoEssentials::battleGetStackByUnitId(int unitId, bool onlyAlive) const
```
IvanSavenko, line 188: Please remove. This method is identical to `CBattleInfoEssentials::battleGetStackByID`.
Yes, it calls `unitId()` instead of accessing ID, but if you look up `CStack::unitId()` you'll see that this method simply returns ID

IvanSavenko:
Ouch, missed this PR when looking for pull that will be broken by client reorganization.
@dydzio0614 can you rebase/merge this one yourself to solve conflicts?

dydzio0614:
yep

PR #1501, opened 2023-01-23
handle double-click in Town Portal dialog

### LINKED ISSUE #977: can't teleport to a town with double-click on town name
Town Portal skill

kambala-decapitator:
fixes #977

test map (really simple): [townportal.h3m.zip](https://github.com/vcmi/vcmi/files/10479799/townportal.h3m.zip)

### DIFF
--- a/client/gui/CGuiHandler.cpp
+++ b/client/gui/CGuiHandler.cpp
@@ -284,6 +284,8 @@ void CGuiHandler::handleCurrentEvent()
 		switch(current->button.button)
 		{
 		case SDL_BUTTON_LEFT:
+		{
+			auto doubleClicked = false;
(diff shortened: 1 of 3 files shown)


client/gui/CGuiHandler.cpp
```diff
@@ -284,6 +284,8 @@ void CGuiHandler::handleCurrentEvent()
 		switch(current->button.button)
 		{
 		case SDL_BUTTON_LEFT:
+		{
+			auto doubleClicked = false;
```
IvanSavenko, line 288: I think that using auto for bool's is a bit unnecessary, but okay, can stay as it.

IvanSavenko (approved):

nullkiller (approved):

PR #1505, opened 2023-01-23
Cheats improvements

dydzio0614:
- Alternative more friendly names for existing cheats
- Cheats for faerie and azure dragons (vcmifaerie, vcmiazure)
- Cheat "vcmiarmy" for custom units by lowercased unit ID with variable quantity depending on needs (example usage: vcmiarmy core:pikeman)
- Cheat "vcmiexp" for receiving set amount of experience on hero

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -5130,7 +5130,14 @@ void CGameHandler::playerMessage(PlayerColor player, const std::string &message,
 		town = hero->visitedTown;
 
 	if (words.size() == 1 || obj)
+	{
 		handleCheatCode(words[0], player, hero, town, cheated);
+	}
+	else if(words[0] == "vcmiarmy" && words.size() > 1)
+	{
+		std::string cheatCodeWithCreatureIdentifier = std::string(words[0]) + " " + words[1];
+		handleCheatCode(cheatCodeWithCreatureIdentifier, player, hero, town, cheated);
+	}
 	else
 	{
 		for (const auto & i : gs->players)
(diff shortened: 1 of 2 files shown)


server/CGameHandler.cpp
```diff
@@ -7082,13 +7090,46 @@ void CGameHandler::handleCheatCode(std::string & cheat, PlayerColor player, cons
...
+		creatures.insert(std::make_pair("vcmiazure", std::make_pair(132, 5000))); //5000 azure dragons
+		creatures.insert(std::make_pair("vcmifaerie", std::make_pair(134, 5000))); //5000 faerie dragons
 
 		const CCreature * creature = VLC->creh->objects.at(creatures[cheat].first);
 		for (int i = 0; i < GameConstants::ARMY_SIZE; i++)
 			if (!hero->hasStackAtSlot(SlotID(i)))
 				insertNewStack(StackLocation(hero, SlotID(i)), creature, creatures[cheat].second);
 	}
-	else if (cheat == "vcminoldor")
+	else if (boost::starts_with(cheat, "vcmiarmy"))
```
IvanSavenko, line 7104: You can use 
`VLC->modh->identifiers.getIdentifier(CModHandler::scopeGame(), "creature", identifier);`
to find creature by ID instead of direct search in creh->objects

server/CGameHandler.cpp
```diff
@@ -7082,13 +7090,46 @@ void CGameHandler::handleCheatCode(std::string & cheat, PlayerColor player, cons
...
+		std::vector<std::string> words;
+		boost::split(words, cheat, boost::is_any_of(" "));
+
+		if(words.size() < 2)
+			return;
+
+		std::string creatureIdentifier = words[1];
+
+		auto creature = vstd::tryFindIf(VLC->creh->objects,
+										[creatureIdentifier](auto x)
```
nullkiller, line 7118: Did we agree to use these weird indentations? I still against them. Just waste of space.

server/CGameHandler.cpp
nullkiller: I agree. Using string identifier {"archangel", 5} is quite self describing and do not need comments

nullkiller: Moreover we can fallback to our new cheat vcmiarmy. vcmiarchangel = vcmiarmy archangel 5. So your new cheat does not accept the amount which might be useful in some situations

dydzio0614: my cheat fills each slot with different amount of creatures so it is fairly easy to adjust the amount you want - does it sound good enough?

server/CGameHandler.cpp
```diff
@@ -7073,22 +7081,57 @@ void CGameHandler::handleCheatCode(std::string & cheat, PlayerColor player, cons
...
+
+		std::vector<std::string> words;
+		boost::split(words, cheat, boost::is_any_of(" "));
+
+		if(words.size() < 2)
+			return;
+
+		std::string creatureIdentifier = words[1];
+
+		auto creature = vstd::tryFindIf(VLC->creh->objects,
```
IvanSavenko, line 7119: Use getIdentifier here as well

dydzio0614: I made that one allow case insensitive name of creatures on purpose, so you do not need to think if the creature is crystalDragon or crystaldragon etc.

IvanSavenko: It is always crystalDragon. All H3 creatures use such format.
Also, I'd like to eventually remove direct access to (handler)->objects, so such code should be avoided

dydzio0614: i intended this to be used with mods, i guess i can go your way and add case-insensitive search later if needed

PR #1508, opened 2023-01-25
attempt to show shipyard dialog only when a hero is selected

kambala-decapitator:
> double-clicking town without Shipyard opens "This an abandoned shipyard" dialog.

fixes regression coming from #1484


IvanSavenko:
Fix confirmed

PR #1509, opened 2023-01-26
Launcher layout changes

IvanSavenko:
Attempt to improve layout of Launcher elements based on accumulated info & user feedback.

### Mod list
Changes:
- Mod info panel is always visible, apparently it was hidden too well.
- Removed broken "Download size" column from Mod List (was always set to 0)
- Removed "Authors", column from Mod List, now it is only shown in mod info panel (since space is now a bit more limited)
- Slight adjustments to overall layout
- Renamed "Home" to "Contact", to better represent usage of this field by mods

### Settings
Mod list now uses single-column layout with scrolling + separate list of categories for navigation.
Reasoning behind changes:
- On mobile platforms (iphone) available space is quite limited, text barely fits as it is. And you can't resize Launcher window there...
- It looks all our translations are longer than English version. Especially true for German.
- Due to current grid-like arrangement, addition of a single option will always affect entire layout. 
- Since number of options is expanding, we will soon run out of place at least on mobile platforms.

### Screenshots with changes
launcher-mods

launcher-settings


krs0:
> Mod info panel always visible

....this change will trunchiate mod names. Especially for nested mods.
If I am searching for some mods to enable, I am more interested in seeing the whole mod name. And only mod name! Having panel always open is too much clutter.
Imo that panel should be displayed only upon specific user request.

IvanSavenko:
@krs0 Thing is, I haven't changed amount of space allocated to mod names, so names will be truncated like before.

That being said, I'll think about it. But probably in some other way, for example - make mod name the only expanding column, so if you resize Launcher to slightly larger size then mod name column will get more space.

Another option is to allow word-wrapping and allow mod name to use two lines, but not sure whether Qt allow this without major changes.

dydzio0614:
One more thought of what should settings have: exposing hardcodedFeatures from https://github.com/vcmi/vcmi/blob/develop/config/defaultMods.json in launcher

IvanSavenko:
@krs0 
Probably leaving as it. Width of name column is same as before, and with 1-2 exceptions I see entire name for all of them (again, same as before). It is also possible to resize columns manually, if you feel they are to narrow, or resize Launcher itself

@dydzio0614 
Since these settings may affect gameplay they are better off in mods imo. At least until they can be reset to default value.

PR #1512, opened 2023-01-26
Converted ChangeLog to use Markdown syntax

IvanSavenko:
Changelog now uses Markdown syntax which has some additional features on Github, namely:
- nicer formatting
- and more interesting one - HTML anchors, allowing us to link to specific section of a file.
For example like this: [Changelog 0.99 -> 1.0](https://github.com/IvanSavenko/vcmi/blob/changelog_markdown/ChangeLog.md#099---100)

All entries were converted, all the way to initial 0.50 release

With this formatting we can remove embedded changelog from our release announcements & instead use link to it. Main reason is that I've seen multiple complains from players that download links are hard to find due to extremely long changelog.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,6 @@
-1.0.0 -> 1.1.0
+# 1.0.0 -> 1.1.0
 
-GENERAL:
+### GENERAL:
 * iOS is supported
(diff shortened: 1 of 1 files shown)


Alucard648:
404

IvanSavenko:
Fixed link. Deleted my branch after merge.

PR #1514, opened 2023-01-26
Minor fixes to rewardable objects

IvanSavenko:
Fixes so far:
- Fix reset of rewardable objects with reset period of 1 day
- Fix Campfire Gold reward amount. Now always one of 400/500/600


DjWarmonger:
From today's daily build:
Campfire gold
Rewards have not quantized values, should be a multiple of 100.

PR #1518, opened 2023-01-28
Skills rework: separate bonus for each skill - part 1

rilian-la-te:
I propose to make separate bonus for each skill and in some cases for each level.

Tasks:
- [x] Add PERCENT_TO_SOURCE and PERCENT_TO_TARGET_TYPE bonus value types to allow secondary skill specialists works without separate bonuses
- [x] Add an ability to load bonuses from defaultMods.json
- [x] Recheck some functionality like mana limit and mana regen
- [x] Write a converter from old SECONDARY_SKILL_PREMY and SECONDARY_SKILL_VAL2 to new bonuses.
- [x] Write a converter for target conditions too
- [x] Remove some deprecated bonuses and replace it by other workable bonuses 

Convert skills:
- [x] "airMagic"
- [x] "archery"
- [x] "armorer"
- [x] "artillery"
- [x] "ballistics"
- [x] "diplomacy"
- [x] "eagleEye"
- [x] "earthMagic"
- [x] "estates"
- [x] "fireMagic"
- [x] "firstAid"
- [x] "intelligence"
- [x] "learning"
- [x] "logistics"
- [x] "mysticism"
- [x] "navigation"
- [x] "necromancy"
- [x] "offence"
- [x] "pathfinding"
- [x] "resistance"
- [x] "scholar"
- [x] "sorcery"
- [x] "tactics"
- [x] "waterMagic"
- [x] "wisdom"

Limitations:
1. Spells and wisdom should work as before (no separate and isLearnable)
2. Mods may be unhappy due to deprecation of SECONDARY_SKILL_PREMY (only for tactics, because converter cannot handle that).
3. Savegame format will be changed

Testing:
1. Globals likely working.
2. Specials also most likely working (testing on Andra and Int spec)
3. Spell-converted actions also seems to working.
4. Deprecated bonus converter also working.

Fixes: 
https://bugs.vcmi.eu/view.php?id=3209
Clear up usage of CATAPULT and CATAPULT_EXTRA_SHOTS bonuses from #1236

PR #1519, opened 2023-01-28
vcmi: remove MAXED_SPELL bonus

rilian-la-te:
It is covered by SPECIAL_FIXED_VALUE_ENCHANT, so, just replace remaining instances and remove bonus altogether.

### DIFF
--- a/lib/HeroBonus.h
+++ b/lib/HeroBonus.h
@@ -285,7 +285,6 @@ public:
 	BONUS_NAME(SPELL_DAMAGE) /*val = value*/\
 	BONUS_NAME(SPECIFIC_SPELL_DAMAGE) /*subtype = id of spell, val = value*/\
 	BONUS_NAME(SPECIAL_BLESS_DAMAGE) /*val = spell (bless), additionalInfo = value per level in percent*/\
-	BONUS_NAME(MAXED_SPELL) /*val = id. deprecated in favour of SPECIAL_FIXED_VALUE_ENCHANT*/\
 	BONUS_NAME(SPECIAL_PECULIAR_ENCHANT) /*blesses and curses with id = val dependent on unit's level, subtype = 0 or 1 for Coronius*/\
 	BONUS_NAME(SPECIAL_UPGRADE) /*subtype = base, additionalInfo = target */\
 	BONUS_NAME(DRAGON_NATURE) \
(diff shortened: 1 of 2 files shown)


IvanSavenko:
No objections against change itself, but perhaps you should also remove enum value from HeroBonus.h?
(although removing enum value will break saves - probably better to rename it to UNUSED_BONUS or something)

Quick search also shows a few more hits for this bonus name than you have in your changes.

rilian-la-te:
@IvanSavenko removed remaining usage. Renamed enum value to UNUSED0.

DjWarmonger:
We can remove it from Bonus enum, but this will break save compatibility. This also requires bump of SERIALIZATION_VERSION.  So this merge can wait for some convenient moment.

rilian-la-te:
@DjWarmonger Ok, bonus is removed. What do you think about splitting this PR, and where JSON part will be merged right away, save compatibility will wait?

DjWarmonger:
> @DjWarmonger Ok, bonus is removed. What do you think about splitting this PR, and where JSON part will be merged right away, save compatibility will wait?
This is actually a good idea, we can do the cleaning sometime later.

Nordsoft91 (approved):

lib/HeroBonus.h
```diff
@@ -285,7 +285,6 @@ class DLL_LINKAGE CAddInfo : public std::vector<si32>
 	BONUS_NAME(SPELL_DAMAGE) /*val = value*/\
 	BONUS_NAME(SPECIFIC_SPELL_DAMAGE) /*subtype = id of spell, val = value*/\
 	BONUS_NAME(SPECIAL_BLESS_DAMAGE) /*val = spell (bless), additionalInfo = value per level in percent*/\
-	BONUS_NAME(MAXED_SPELL) /*val = id. deprecated in favour of SPECIAL_FIXED_VALUE_ENCHANT*/\
```
nullkiller, line 288: Interesting if we break save compatibility here as bonus enum numeration may shift?

nullkiller: If yes I would rename bonus instead as DEPRECATED_MAXED_SPELL or UNUSED_MAXED_SPELL and remove when we break save compatibility once again. (Or do we plan to do this in 1.2?). Or at least increase serialization version.

nullkiller:
So this waits when we decide if we want to break save compatibility in 1.2?

rilian-la-te:
@nullkiller this bonus is already deprecated. If we break save compatibility, I think we can rebase it and remove bonus altogether.

rilian-la-te:
Handled in #1518

PR #1521, opened 2023-01-29
Rework netpacks

nullkiller:
This is an attempt to fix my Android 5.0 device. For some reason it did not like
* presence of main() in libvcmi_server - refactored
* dynamic_cast on CPack returns nullptr - used visitor pattern and typeid instead

### DIFF
--- a/lib/NetPacks.h
+++ b/lib/NetPacks.h
@@ -36,7 +36,7 @@ struct ArtSlotInfo;
 struct QuestInfo;
 class IBattleState;
 
-struct Query : public CPackForClient
+struct DLL_LINKAGE Query : public CPackForClient
 {
 	QueryID queryID; // equals to -1 if it is not an actual query (and should not be answered)
 };
@@ -53,8 +53,8 @@ struct StackLocation
 	{
 	}
 
-	DLL_LINKAGE const CStackInstance *getStack();
-	template <typename Handler> void serialize(Handler &h, const int version)
+	DLL_LINKAGE const CStackInstance * getStack();
+	template <typename Handler> void serialize(Handler & h, const int version)
 	{
 		h & army;
 		h & slot;
@@ -62,23 +62,21 @@ struct StackLocation
 };
 
 /***********************************************************************************************************/
-
-
-struct PackageApplied : public CPackForClient
+struct DLL_LINKAGE PackageApplied : public CPackForClient
 {
 	PackageApplied() = default;
 	PackageApplied(ui8 Result)
 		: result(Result)
 	{
 	}
-	void applyCl(CClient *cl);
+	virtual void visitTyped(ICPackVisitor & visitor) override;
 
 	ui8 result = 0; //0 - something went wrong, request hasn't been realized; 1 - OK
 	ui32 packType = 0; //type id of applied package
 	ui32 requestID = 0; //an ID given by client to the request that was applied
 	PlayerColor player;
 
-	template <typename Handler> void serialize(Handler &h, const int version)
+	template <typename Handler> void serialize(Handler & h, const int version)
 	{
 		h & result;
 		h & packType;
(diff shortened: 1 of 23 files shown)


nullkiller:
I am not sure what to do now. The thing looks completely weird.
Basic theory is that there is two definitions of LobbyClientConnected CPack only for this device. This might be caused by some OS version specifics related to how libraries are loaded.
Another version is that background service somehow loads wrong stl.
Single process may fix it as well.

IvanSavenko:
FYI - there was somewhat similar issue long time ago.
Objects created in library and in client/server had different instances of typeinfo (but with identical data), e.g.
`typeid(clientPtr) != typeid(libraryPtr)`
however
`typeid(clientPtr).name() == typeid(libraryPtr).name()`

Could be related to this one.
By the way, check `dynamic_ptr_cast` in serialized/Cast.h. It has special case for VCMI_APPLE. Perhaps this code should also be used for Android?

nullkiller:
Probably better to get rid of this duplication and put all cpacks to vcmi_lib

nullkiller:
NetPacksLobbyClient.cpp is a part of vcmiclient for instance. How server knows about these definitions? Maybe it just do not use them so they stay unresolved. I do not understand why Android 5.0 is only affected. Maybe newer platforms can somehow overcome this issue

nullkiller:
Same situation for server. It has NetPacksLobbyServer which also contains some method definitions for lobby packs which only server uses

nullkiller:
reworked netpacks to use visitors and all netpacks code in lib to avoid duplicating typeinfo

lib/NetPacks.h
```diff
@@ -69,15 +69,14 @@ struct StackLocation
...
 {
 	PackageApplied()
 		: result(0), packType(0),requestID(0)
 	{}
 	PackageApplied(ui8 Result)
 		: result(Result), packType(0), requestID(0)
-	{}
-	void applyCl(CClient *cl);
+	{}
	virtual void visitTyped(ICPackVisitor & visitor) override;
```
IvanSavenko, line 79: Weird formatting? Should be end of line instead of... whatever is "^M" is
(not just here, search shows 18 hits)

rilian-la-te: @IvanSavenko it is windows EOL, should not be used in UNIX.

nullkiller: Maybe something. A lot of changes were done using some regex replaces) It will be tested and cleaned up. Just first building version

nullkiller:
Brief testing showed that it works. Now will test on Android 5 device and see if there is still an issue with dynamic cast. Fingers crossed)

nullkiller:
A few tests on my old device showed that dynamic_cast works now as expected. But different issues arise. So will continue investigating.

nullkiller:
Release build works. Debug has triggered some assertion during battle. @IvanSavenko if no objections I will merge it to unblock

server/NetPacksServer.cpp
```diff
@@ -4,11 +4,11 @@
  * Authors: listed in file AUTHORS in main folder
  *
  * License: GNU General Public License v2.0 or later
- * Full text of license available in license.txt file, in main folder
+ * Full pack.text of license available in license.txt file, in main folder
```
IvanSavenko, line 7:
```suggestion
 * Full text of license available in license.txt file, in main folder

```

IvanSavenko (approved):
What kind of assertion? Any idea whether it triggered before?

Would be nice to fix it if possible, but not critical I guess. Can look into it myself later.

dydzio0614:
There are already A LOT of different assertions that hit in debug build... including nullkiller AI frequent assertions, assert when waiting a while during post-battle screen, assert when you set speedFactor to 8 in config and try to shoot with crossbowman or lizard... etc etc.

IvanSavenko:
@dydzio0614 haven't seen any of them. Will recheck.

nullkiller:
> What kind of assertion? Any idea whether it triggered before?
> 
> Would be nice to fix it if possible, but not critical I guess. Can look into it myself later

do not remember. I made a shot with marksman and killed enemy stack from first short while marksman shot twice and some assertion triggered in animation.

IvanSavenko:
Yeah, I think I know what's wrong, not related to branch.

nullkiller:
I will merge it if there is no objections than

IvanSavenko:
> I will merge it if there is no objections than

Sure, no objections here

PR #1522, opened 2023-01-29
Russian translation and Qt6 translation update

rilian-la-te:
Russian translation for launcher and mapeditor.
Small changes to CMakeLists.txt to make generating of translations easier on Qt 6.


IvanSavenko:
Do we have any CI that uses Qt6? 
If not - can you confirm that resulting files are installed into same location as with Qt5?

Other than that - looks ok

rilian-la-te:
@IvanSavenko it is same files and will be installed in same locations. I just use Qt6 CMake helper to update it.

IvanSavenko (approved):

PR #1527, opened 2023-01-30
SDL abstraction layer iteration

IvanSavenko:
One more step towards SDL abstraction layer:
- CPicture class now uses IImage class instead of SDL_Surface
- removed all remaining usages of direct access of CBItmapHandler that loads images into SDL_Surface
- removed usage of SDL integer defines (Uint8/Uint32)
- replaced few remaining usage of SDL_Delay with this_thread::sleep
- added ColorRGBA class that can also be used in library.


IvanSavenko:
Unless there are objections, will merge in 2-3 days, before conflicts with other branches arise.

PR #1528, opened 2023-01-31
vcmi: deprecate MAXED_SPELL

rilian-la-te:
It is almost identical to SPECIAL_FIXED_VALUE_ENCHANT.

Replace its usage in JSON and inside deprecated converter to SPECIAL_FIXED_VALUE_ENCHANT. Remove buggy string inside getEffectLevel.

This is a variant of #1519 without removing savegame compatibility.

### DIFF
--- a/config/heroes/tower.json
+++ b/config/heroes/tower.json
@@ -213,8 +213,9 @@
 		"specialty" : {
 			"bonuses" : {
 				"fortune" : {
+					"addInfo" : 3,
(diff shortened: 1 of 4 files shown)


rilian-la-te:
@DjWarmonger separated bonus removal to ensure save compatibility

DjWarmonger (approved):

DjWarmonger:
Yeah sure, we already have this solution for other heroes (Melodia).

PR #1531, opened 2023-01-31
Remove mxe

IvanSavenko:
- Remove MXE action from CI
- Remove non-preset CMake build from CI config
- Removed MXE-related code from CMakeLists.txt
- Bumped CMake version requirement to 3.16
- Updated readme.md


rilian-la-te:
@IvanSavenko this is not all needed changes(

IvanSavenko:
@rilian-la-te 
Yep. Bumping to 3.16 (due to Ubuntu-20.04) and removing workarounds.
Want to check that CI is working before CMake changes

IvanSavenko:
@rilian-la-te CMakeLists updated. Anything else that I've missed?

rilian-la-te:
@IvanSavenko
1. Add my Wiki about conan instead of MXE: https://wiki.vcmi.eu/How_to_build_VCMI_(Linux/Cmake/Conan). 

Other things may be handled by separate PR (like trying Conan for MSVC too).

rilian-la-te:
@IvanSavenko seems normal) Please, review my Russian translation)

rilian-la-te (changes requested):
``` * [On Linux for Windows with Conan and mingw]( https://wiki.vcmi.eu/How_to_build_VCMI_(Linux/Cmake/Conan))```

File: .github/workflows/github.yml [hunk 1]
```diff
@@ -236,7 +203,7 @@ jobs:
           ${{github.workspace}}/**/${{ env.VCMI_PACKAGE_FILE_NAME }}.${{ matrix.extension }}
 
     - name: Upload build
-      if: ${{ matrix.pack == 1 && (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/beta' || github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/features/')) && matrix.platform != 'msvc' }}
+      if: ${{ matrix.pack == 1 && (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/beta' || github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/heads/features/')) && matrix.platform == 'msvc' }}
```
kambala-decapitator, line 206: FYI no artifacts are uploaded any more due to this change

IvanSavenko: Hmm.. I get why it breaks non-Windows platforms but why msvc build fails to upload as well? And with ssh error?

Can remove this check later today

kambala-decapitator: maybe some setting in Actions configuration to not propagate the key for MSVC build? Don't have access to that to check.

IvanSavenko: DEPLOY_RSA is repositoty secret so I think it should be available to every action?
Don't see any way to configure it anyway.

PR #1532, opened 2023-01-31
hide AI cheating message

nullkiller:
Lets not spam player about AI cheating with vcmieagles for now.


kambala-decapitator:
I'd say better target `beta`

dydzio0614:
"cheater" message is not there, but "vcmieagles" message still appears

nullkiller:
Yes, let vcmieagles stay. It does not irritate that much.

Issue #1320, opened 2022-12-27
effect of Boots of Levitation/Angel Wings isn't applied immediately


kambala-decapitator:
My actions were:
1. approached water
2. put Boots of Levitation on
3. couldn't cross water, thought I don't have enough movement points any more
4. switched back to the boots that give extra movement points, waited till next day
5. put Boots of Levitation on, still couldn't cross water

Pretty sure that I had enough points at the beginning of a day: hero had 3 `+` marks and water distance was small enough.

Then I loaded the save again to re-test, but this time all worked correctly...

https://bugs.vcmi.eu/view.php?id=2098

kambala-decapitator:
same issue with Angel Wings: Fly effect appears only on the next day

kambala-decapitator:
Making a step in any direction makes considering Angel Wings again on next route planning

IvanSavenko:
I guess that changing artifact does not triggers pathfinder cache update.
I see there is `CPlayerInterface::heroBonusChanged` which (I assume) should have handled this case, but didn't for some reason.

kambala-decapitator:
existing bug report: https://bugs.vcmi.eu/view.php?id=2098

PR #1534, opened 2023-02-01
fix movement cost with Fly

### LINKED ISSUE #1412: strange flying path choice
This is a Fortress hero, he has Angel Wings. Why not a straight path was chosen?

Screenshot 2023-01-11 at 07 54 03

maybe related to https://bugs.vcmi.eu/view.php?id=2650?

kambala-decapitator:
fixes #1412

made some tests on a save I had (in the issue) and seems correct now. Which other [edge] cases should be tested to make sure that nothing got broken?

Comment in code said

> as server don't know when player flying and when he's not...
> So this one is waiting till server going to use pathfinder rules for path validation.

but I see that this method is actually called by server from `bool MoveHero::applyGh(CGameHandler * gh)`, so it means that actually server already does everything correctly?


DjWarmonger:
Git blame show that this comment is 7 years old. Though I think it's incorrect and against the fundamental game design in first place - every player action is checked on server, always.

dydzio0614:
First thing: Maybe I would target develop with this patch...

Is this commend indicating that logic is used only for client-side calculation or something?

If yes, then regarding Warmonger's answer - maybe code comment is about that server doesn't (or didn't) lower tile cost after fly spell was applied, so the comment indicates spell logic needed to be changed so the clientside-related pathfinder logic doesn't desync with what server allows?

Not sure about corner cases - angel wings + water walking artifacts? Movement point artifacts? Another obscure part of game logic is movement point amount change that starts applying since next turn when lowest speed of creature in army changes, idk if this patch can break it but im not sure what kind of stuff you need.

IvanSavenko:
> as server don't know when player flying and when he's not...

Outdated comment. Server does known about that now. Check `EPathfindingLayer` enum.
So `getMovementCost` method should probably receive pathfinder nodes instead of terrain tiles and apply values from FLYING_MOVEMENT only for `EPathfindingLayer::AIR`.

Favorable Winds check should also probably be layer-specific (SAIL)

IvanSavenko:
And yes - I would also rather target develop and not beta - this is not trivial change & hard to estimate possible side effects.

IvanSavenko:
One more thing that I would recheck - Fly over water.
Judging by bug reports, it seems that in vcmi Fly over water spends 100 MP per tile. That the only explanation I have as to why Fly over swamp would prefer sidetracking into water.

But while thelazy does not mentions that, according Fizmig Fly over water should cost 140/120/100 MP - same as other high-penalty terrains, e.g. Swamp.

kambala-decapitator:
thank you all for the input!

> If yes, then regarding Warmonger's answer - maybe code comment is about that server doesn't (or didn't) lower tile cost after fly spell was applied, so the comment indicates spell logic needed to be changed so the clientside-related pathfinder logic doesn't desync with what server allows?

seems not the case from my tests: after casting Fly current planned path (green arrows) is reset and next path planning takes Fly into account, server doesn't complain that something's wrong on performing move.

> Not sure about corner cases - angel wings + water walking artifacts? Movement point artifacts? Another obscure part of game logic is movement point amount change that starts applying since next turn when lowest speed of creature in army changes, idk if this patch can break it but im not sure what kind of stuff you need.

I haven't touched hero's total movement points in the patch, so shouldn't be an issue, only tile cost changes. Those artifacts have correct bonuses in the config, but they have another bug btw which should be fixed separately: #1320.

> So getMovementCost method should probably receive pathfinder nodes instead of terrain tiles and apply values from FLYING_MOVEMENT only for EPathfindingLayer::AIR.
> Favorable Winds check should also probably be layer-specific (SAIL)

I'll check that and will try to adjust accordingly.

> And yes - I would also rather target develop and not beta - this is not trivial change & hard to estimate possible side effects.

ok, I agree, will retarget. Will close this PR.

> One more thing that I would recheck - Fly over water.
Judging by bug reports, it seems that in vcmi Fly over water spends 100 MP per tile. That the only explanation I have as to why Fly over swamp would prefer sidetracking into water.

yes, that's what I also figured out why such strange path was chosen in the issue, as Fly doesn't cancel terrain penalty. With the change Fly over water continues to work correctly but no longer tries to build such an _interesting_ route.

> But while thelazy does not mentions that, according Fizmig Fly over water should cost 140/120/100 MP - same as other high-penalty terrains, e.g. Swamp.

this is correct: minimum possible terrain penalty is applied, which is Fly in most cases (e.g. 140 on Basic Air vs Swamp's 175), and Water doesn't have any special penalty making it cost 140 on Basic Air.

IvanSavenko:
FYI - you can retarget PR without closing it:
click Edit to the right from PR name - this will make target branch editable

kambala-decapitator:
btw could someone explain what do the last lines of this method check? I don't understand:

> it might be the last tile - if no further move possible we take all move points

comment says it's also broken for Fly

IvanSavenko:
@kambala-decapitator 
H3 has special handling of the last tile.
For example, if you stand in a middle of grass patch and have 100-140 MP left, you can still move in *any* direction, including diagonally.
This rule get more complex with other scenarios - when you take into account penalty & roads, see here for more details:
https://heroes.thelazy.net//index.php/Movement#Diagonal_move_exception

dydzio0614:
"Due to the bug it also doesn't always apply to heroes on a penalty terrain if they lead armies comprised entirely of creatures native to that terrain. In this case the game oddly 'forgets' about the terrain penalty elimination and allows the last diagonal move only if the hero has at least the number of movement points needed to make a straight move calculated including terrain penalty. As a result the cheaper last diagonal move is generally unattainable for armies native to swamp and snow - because straight move with penalty costs 175 and 150 points, respectively, which is greater than at most 140 points remaining with hero in this scenario."

I vote for fixing this movement case

rilian-la-te:
@kambala-decapitator will this or similar PR will be merged to 1.2?

kambala-decapitator:
Yes, plan to come back to it next week. Android stuff was more important.

kambala-decapitator:
first force-push was new changes (had to amend last commit), second was rebase on develop to fix conflict.

besides Fly I also found that road cost is incorrect and added a few code improvements.

regarding the check of the last tile that said:

> TODO: This part need rework in order to work properly with flying and water walking
> Currently it's only work properly for normal movement or sailing

I've checked the implementation of `getNeighbours()` and to me it looks already correct for fly/water walk cases (hm, just thought that probably except the case of "diagonal movement over coast when sailing" which would be valid), and besides last move must end on the land.

> So getMovementCost method should probably receive pathfinder nodes instead of terrain tiles and apply values from FLYING_MOVEMENT only for EPathfindingLayer::AIR.
> Favorable Winds check should also probably be layer-specific (SAIL)

I didn't find an easy way to change the method to work with `PathNodeInfo` directly because it's also called from `CGameHandler::moveHero()` where those objects seem inaccessible. Such refactoring can be done in another PR.

lib/CPathfinder.cpp
```diff
@@ -1217,48 +1218,37 @@ int CPathfinderHelper::getMovementCost(
...
-	/// TODO: by the original game rules hero shouldn't be affected by terrain penalty while flying.
-	/// Also flying movement only has penalty when player moving over blocked tiles.
-	/// So if you only have base flying with 40% penalty you can still ignore terrain penalty while having zero flying penalty.
 	int ret = hero->getTileCost(*dt, *ct, ti);
-	/// Unfortunately this can't be implemented yet as server don't know when player flying and when he's not.
-	/// Difference in cost calculation on client and server is much worse than incorrect cost.
-	/// So this one is waiting till server going to use pathfinder rules for path validation.
-
-	if(dt->blocked && ti->hasBonusOfType(Bonus::FLYING_MOVEMENT))
+    if(hero->boat != nullptr && dt->terType->isWater())
```
IvanSavenko, line 1222: I could be wrong, but this code may work incorrectly in some edge cases.
If hero is on land, but builds a path across favorable winds on water, this condition will evaluate to false.

Might even happen within single turn/move order - if hero has Admiral's Hat and can embark boat & reach favorable winds in a single turn.

If I am correct, can you either fix or add comment about such scenario?

nullkiller: Looks like boat is really extra here. Another option will be to check if layer is SAIL. This will guarantee that the boat must exist.

lib/CPathfinder.cpp
```diff
@@ -1217,48 +1218,37 @@ int CPathfinderHelper::getMovementCost(
...
-		else if(!hero->boat && ti->hasBonusOfType(Bonus::WATER_WALKING))
-		{
-			ret = static_cast<int>(ret * (100.0 + ti->valOfBonuses(Bonus::WATER_WALKING)) / 100.0);
-		}
+		if(ct->hasFavorableWinds() && dt->hasFavorableWinds())
+			ret = static_cast<int>(ret * 2.0 / 3);
 	}
+	else if(ti->hasBonusOfType(Bonus::FLYING_MOVEMENT))
+		vstd::amin(ret, GameConstants::BASE_MOVEMENT_COST + ti->valOfBonuses(Bonus::FLYING_MOVEMENT));
+	else if(hero->boat == nullptr && dt->terType->isWater() && ti->hasBonusOfType(Bonus::WATER_WALKING))
```
nullkiller, line 1229: Same here as it is allowed to automatically embark/disembark checking layer==WATER is better here

lib/CPathfinder.cpp
nullkiller: I believe EPathfindingLayer should be available somewhere upper in call stack

kambala-decapitator: I'll create a test map with AH and see what happens.

In one case the layer is available, in another one it seems not:

> it's also called from CGameHandler::moveHero() where those objects seem inaccessible

kambala-decapitator:
observing an interesting behavior with AH (same with and without this PR), red rectangle marks the hovered tile

Screenshot 2023-03-18 at 12 28 56 Screenshot 2023-03-18 at 12 29 05 Screenshot 2023-03-18 at 12 33 34

IvanSavenko:
Probably bug in land<->sea move point conversion, see
`CGHeroInstance::movementPointsAfterEmbark`
H3 has similar bug by the way:
https://heroes.thelazy.net//index.php/Admiral%27s_Hat

dydzio0614:
pain, I guess we will eventually need both - bugged and unbugged version

IvanSavenko:
Don't think that we should bother implementing this (clearly buggy) behavior

dydzio0614:
in that case I would probably add comment in code that original H3 behavior is buggy in this regard
and be open to implement this if somebody actually complains about SoD incompatibility

IvanSavenko:
> and be open to implement this if somebody actually complains about SoD incompatibility
We already have such behavior! It is called 'vcminahar'

This is clearly an exploit that allows infinite movement points. Not a feature.

kambala-decapitator:
> and be open to implement this if somebody actually complains about missing SoD exploit

fixed =)

HotA doesn't have the bug according to thelazy btw

kambala-decapitator:
my current test map:
[ah-aw.vmap.zip](https://github.com/vcmi/vcmi/files/11008302/ah-aw.vmap.zip)

kambala-decapitator:
changed code to use Layer when possible, there's still fallback to the manual boat check when Layer info is unavailable.

strangely computed MP with AH: that seems correct due to land <-> water MP re-computation. But AH has other issues, created separate issue for that #1723

also fixed tile cost with Favorable Winds and added a few more code improvements.

same map with FW:
[ah-aw.vmap.zip](https://github.com/vcmi/vcmi/files/11011237/ah-aw.vmap.zip)

IvanSavenko (approved):

PR #1535, opened 2023-02-01
Reorganize client source files

IvanSavenko:
In preparation for my future change, I'd like to rearrange files in client subproject to better represent project structure.
Done:
- [x] Move files related to adventure map to a dedicated folder, similar to battle/lobby
- [x] Move files related to rendering to a dedicated folder
- [x] Split CAnimation file into multiple files
- [x] Split AdventureMapClasses into multiple files
- [x] Split CAdvmapInterface into multiple files
- [x] Split Fonts into multiple files
- [x] Split CursorHandler into multiple files
- [x] Update CMakeLists to reflect new paths
- [x] Fix compilation due to change of includes

Will try to use git features as much as possible, in order to preserve history/blame. I also don't see any ongoing PR's targeting affected files, so conflicts with other PR's should be minimal.


rilian-la-te:
@IvanSavenko there is a CI failure.

IvanSavenko:
Done. Will merge tomorrow, provided that CI passes.

Most of changes are moving code around, other than that:
- removed some `using namespace SDL_Ext` & added namespace to calls
- added IImageLoader interface, to act as bridge between CDefFile and SDLImageLoader (was template method before)

rilian-la-te:
@IvanSavenko maybe add and use clang-format for new files (and only for new files)?

PR #1545, opened 2023-02-02
Improve creature numbers handling in castle interface bottom left part

dydzio0614:
Default label now matches H3, also improved support for HD Mod way (disabled by default) - for now accesible via "/set availableCreaturesAsLabel on"

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -45,6 +45,18 @@
 
 #include <SDL_events.h>
 
+static bool useCompactCreatureBox()
+{
+	//TODO: add UI for easier access
+	return settings["session"]["compactCreatureInfo"].isNumber() ? settings["session"]["compactCreatureInfo"].Bool() : false;
+}
+
+static bool useAvailableAmountAsCreatureLabel()
+{
+	//TODO: add UI for easier access
+	return settings["session"]["availableCreaturesAsLabel"].isNumber() ? settings["session"]["availableCreaturesAsLabel"].Bool() : false;
+}
+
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
> for now accesible via "/set availableCreaturesAsLabel on"

I'd rather make it available in json config (and in the launcher)

IvanSavenko:
> > for now accesible via "/set availableCreaturesAsLabel on"
> 
> I'd rather make it available in json config (and in the launcher)

Seconded.

Also, I propose to add new section in our settings json, let's call it guiTweaks (or something like that) and use it for all changes like this that deviate from H3

dydzio0614:
I want to change this after finishing VCMI settings

client/windows/CCastleInterface.cpp
```diff
@@ -1042,10 +1057,12 @@ void CCreaInfo::clickLeft(tribool down, bool previousState)
 	if(previousState && (!down))
 	{
 		int offset = LOCPLINT->castleInt? (-87) : 0;
+
 		auto recruitCb = [=](CreatureID id, int count)
 		{
 			LOCPLINT->cb->recruitCreatures(town, town->getUpperArmy(), id, count, level);
 		};
+
```
Nordsoft91, line 1065:
```suggestion
```

client/windows/CCastleInterface.cpp
```diff
@@ -1042,10 +1057,12 @@ void CCreaInfo::clickLeft(tribool down, bool previousState)
 	if(previousState && (!down))
 	{
 		int offset = LOCPLINT->castleInt? (-87) : 0;
+
```
Nordsoft91, line 1060:
```suggestion
```

PR #1546, opened 2023-02-02
Remove SDL usage from VCMI library

rilian-la-te:
It will allow us to compile VCMI library without SDL headers at all. 

In line with other SDL removal policy.


rilian-la-te:
@IvanSavenko for removing SDL from VCMI Library (now it depends only on boost).

IvanSavenko:
Can you also make same change in client's SDL_PixelAccess.h for consistency?

IvanSavenko:
Oh, and other places that use SDL_BYTEORDER

rilian-la-te:
@IvanSavenko done

CMakeLists.txt
```diff
@@ -173,6 +173,18 @@ set(CMAKE_XCODE_ATTRIBUTE_MARKETING_VERSION ${APP_SHORT_VERSION})
...
 set(CMAKE_XCODE_ATTRIBUTE_ONLY_ACTIVE_ARCH[variant=Debug] YES)
 
+#Check for endian
+if(${CMAKE_VERSION} VERSION_LESS "3.20.0") 
+	include(TestBigEndian)
+	test_big_endian(VCMI_ENDIAN_BIG)
+	if(VCMI_ENDIAN_BIG)
+		add_definitions(-DVCMI_ENDIAN_BIG)
+	endif()
+elseif(${CMAKE_CXX_BYTE_ORDER} EQUAL "BIG_ENDIAN")
```
kambala-decapitator, line 183: nice trick :)

kambala-decapitator (approved):

CMakeLists.txt
rilian-la-te: @kambala-decapitator you dislike this? In cmake less than 3.20 we does not have CMAKE_CXX_BYTE_ORDER, but have builtin script TestBigEndian, which return 1 if endian is big.

kambala-decapitator: no-no, I like it! Didn't know such thing exists.

rilian-la-te: @kambala-decapitator this is thing than can lead us to remove unneeded dependencies from VCMI) I need to try another CMake trick to reduce Windows distro size)

PR #1547, opened 2023-02-02
Hardcoded features for HotA-like starting conditions

### LINKED ISSUE #1520: HotA's system of starting troops.
**Describe your proposal**
In HotA heroes with a starting army **always** have all troops (and a war machine), without the possibility of absence of any of them. 

OH3 mechanic is:

The probability of heroes starting (or appearing in Tavern) with only 1 stack is 9%, with 2 its 69%,
with 3 its 22%.
Level 1 unit specialists will only start with level 1 units.
Level 2 unit specialists will only start with level 2 units.
Level 3 unit specialists may start with level 1 and level 3 units.
Heroes having the possibility to start with a War Machine will not start with level 2 units.

**Additional context**
I think it will be cool to have an opportunity to switch mechanics in defaultMods.json

dydzio0614:
I am not sure about war machines part, but for other stuff it fixes #1520 and it's improvement overall for some people who want to play with this feature (got some questions about it).


IvanSavenko:
Perhaps better to define chances instead?
```
// In H3 hero has 10% chance to start with 1 stack, 70% for 2 stacks, and 20% for 3 stacks
HERO_STARTING_ARMY_STACKS_COUNT = [ 10, 70, 20 ]
// In HotA, hero always starts with 3 stacks
HERO_STARTING_ARMY_STACKS_COUNT = [ 0, 0, 100 ]

// In H3, town with default building set will always have 1st dwelling and 50% chance for second
DEFAULT_BUILDING_SET_DWELLING_CHANCES = [ 100, 50 ]
// In HotA, towns with default building set will always have 1st and 2nd dwelling
DEFAULT_BUILDING_SET_DWELLING_CHANCES = [ 100, 50 ]
```

dydzio0614:
Planned to do it as you say now, but after all I think it's way less intuitive to use and restore default value

IvanSavenko:
I view this mostly from point of removing hardcoded values from engine. There is no point in hardcoding values that can be loaded from config.

As for ease of editing - file in vcmi mod is not supposed to be edited by players, and modders who replace it can always check version from vcmi mod.

dydzio0614:
This file isn't in VCMI mod, it is in config folder. It is pretty mod-independent as it changes game rules that are not useful to be parts of mod features.

IvanSavenko:
Still, our config files are not supposed to be edited by players. So "easier to restore original value" is a bad argument IMO.

And in long-term I think this file should only be edited by mods. "Game balancing" is a mod feature after all.

dydzio0614:
Long term is long term, but in meantime players want some way to be able to access this feature

IvanSavenko:
Make a small mod that has this field correctly set?
I *really* dislike various "temporary workaround". Why bother with workarounds when it can be done properly?

misiokles:
Creating mini-mods that can alter game configs is rather simple way. Some mod eneble negative luck in such way.

dydzio0614:
OK, though I still do not see it as fully proper way - proper way would allow heroes to have all 7 stacks filled by default etc.

IvanSavenko:
> OK, though I still do not see it as fully proper way - proper way would allow heroes to have all 7 stacks filled by default without hardcoding exactly three array values etc.

Then don't hardcode? Use std::vector instead of std::array and load every existing value from file.
So if heroes get more than 3 stacks game will still load correct values

lib/CModHandler.h
```diff
@@ -382,6 +384,8 @@ class DLL_LINKAGE CModHandler
 			h & ATTACK_POINTS_DMG_MULTIPLIER_CAP;
 			h & DEFENSE_POINT_DMG_MULTIPLIER;
 			h & DEFENSE_POINTS_DMG_MULTIPLIER_CAP;
+			h & NEW_HERO_ALWAYS_3_CREATURE_STACKS;
+			h & DEFAULT_TOWN_ALWAYS_2_DWELLINGS;
```
Nordsoft91, line 388: please, bump serialisation version and also you can provide backward compatibility easy here:
```suggestion
		if(version >= xxx)
		{
			h & NEW_HERO_ALWAYS_3_CREATURE_STACKS;
			h & DEFAULT_TOWN_ALWAYS_2_DWELLINGS;
		}
```

Nordsoft91 (changes requested):

lib/CModHandler.cpp
```diff
@@ -767,6 +767,18 @@ void CModHandler::loadConfigFromFile (std::string name)
 	settings.DEFENSE_POINTS_DMG_MULTIPLIER_CAP = hardcodedFeatures["DEFENSE_POINTS_DMG_MULTIPLIER_CAP"].Float();
 	logMod->debug("\tDEFENSE_POINTS_DMG_MULTIPLIER_CAP\t%f", settings.DEFENSE_POINTS_DMG_MULTIPLIER_CAP);
 
+	JsonVector stacksChancesVector = hardcodedFeatures["HERO_STARTING_ARMY_STACKS_COUNT_CHANCES"].Vector();
```
IvanSavenko, line 770: use json.convertTo<std::vector<int>>();

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -348,16 +348,25 @@ void CGHeroInstance::initArmy(CRandomGenerator & rand, IArmyDescriptor * dst)
 	if(!dst)
 		dst = this;
 
-	int howManyStacks = 0; //how many stacks will hero receives <1 - 3>
-	int pom = rand.nextInt(99);
+	int howManyStacks = 7; //how many stacks will hero receives <1 - 7>
```
IvanSavenko, line 351: Move to actual initialization
```suggestion
```

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -348,16 +348,25 @@ void CGHeroInstance::initArmy(CRandomGenerator & rand, IArmyDescriptor * dst)
...
+	if(allowZeroStacksArmy)
+		stacksCountChances.pop_back();
+
+	int stacksCountInitRandomNumber = rand.nextInt(1, 100);
+
+	auto stacksCountElementIndex = vstd::find_pos_if(stacksCountChances, [stacksCountInitRandomNumber](int element){ return stacksCountInitRandomNumber < element; });
+	if(stacksCountElementIndex == -1)
+		stacksCountElementIndex = stacksCountChances.size();
+
+	howManyStacks = stacksCountElementIndex;
```
IvanSavenko, line 367:
```suggestion
	int howManyStacks = stacksCountElementIndex;
```

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -348,16 +348,25 @@ void CGHeroInstance::initArmy(CRandomGenerator & rand, IArmyDescriptor * dst)
...
-	if(pom < 9)
-		howManyStacks = 1;
-	else if(pom < 79)
-		howManyStacks = 2;
-	else
-		howManyStacks = 3;
+	std::vector<int32_t> stacksCountChances = VLC->modh->settings.HERO_STARTING_ARMY_STACKS_COUNT_CHANCES;
+
+	const int zeroStacksAllowingValue = -1;
+	bool allowZeroStacksArmy = !stacksCountChances.empty() && stacksCountChances.back() == zeroStacksAllowingValue;
```
IvanSavenko, line 357: Not sure if this check is needed - I expect vcmi to break on empty armies anyway.

IvanSavenko (approved):

lib/mapObjects/CGHeroInstance.cpp
dydzio0614: empty armies work in general so this feature should stay

IvanSavenko:
@Nordsoft91 
OK to merge?

Nordsoft91 (approved):

PR #1552, opened 2023-02-04
Add polish translations, tiny improvement of english translations

dydzio0614:
As in title


IvanSavenko:
I think you might also want to translate this section (currently copied from English)
https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/mod.json#L16

IvanSavenko (approved):

PR #1558, opened 2023-02-05
vcmi: build lua with STRICT_COMPILATION

rilian-la-te:
This is a fixes for some LUA warnings for building it with strict compilation.

Let's eat modernizing elephant by pieces) 

Also modernize LUA by clang-tidy, as it is harmless AFAIK.

### DIFF
--- a/scripting/lua/LuaCallWrapper.h
+++ b/scripting/lua/LuaCallWrapper.h
@@ -12,6 +12,7 @@
 
 #include "api/Registry.h"
 #include "LuaStack.h"
+#include <type_traits>
 
 VCMI_LIB_NAMESPACE_BEGIN
 
@@ -187,6 +188,7 @@ public:
 template <typename U, typename T, typename R, typename P1, R(T:: * method)(P1)const>
 class LuaMethodWrapper <U, R(T:: *)(P1)const, method>
 {
+	using PM1 = std::remove_cv_t<std::remove_reference_t<P1>>;
 public:
 	static int invoke(lua_State * L)
 	{
@@ -196,7 +198,7 @@ public:
 		if(!S.tryGet(1,obj))
 			return S.retVoid();
 
-		P1 p1;
+		PM1 p1;
 		if(!S.tryGet(2, p1))
 			return S.retVoid();
 
@@ -210,6 +212,7 @@ public:
 template <typename U, typename T, typename R, typename P1, R(T:: * method)(P1)>
 class LuaMethodWrapper <U, R(T:: *)(P1), method>
 {
+	using PM1 = std::remove_cv_t<std::remove_reference_t<P1>>;
 public:
 	static int invoke(lua_State * L)
 	{
@@ -219,7 +222,7 @@ public:
 		if(!S.tryGet(1,obj))
 			return S.retVoid();
 
-		P1 p1;
+		PM1 p1;
 		if(!S.tryGet(2, p1))
 			return S.retVoid();
 
@@ -233,6 +236,7 @@ public:
 template <typename U, typename T, typename P1, void(T:: * method)(P1)const>
 class LuaMethodWrapper <U, void(T:: *)(P1)const, method>
 {
+	using PM1 = std::remove_cv_t<std::remove_reference_t<P1>>;
 public:
 	static int invoke(lua_State * L)
 	{
@@ -242,7 +246,7 @@ public:
 		if(!S.tryGet(1,obj))
 			return S.retVoid();
 
-		P1 p1;
+		PM1 p1;
 		if(!S.tryGet(2, p1))
 			return S.retVoid();
 
@@ -256,6 +260,7 @@ public:
 template <typename U, typename T, typename P1, void(T:: * method)(P1)>
 class LuaMethodWrapper <U, void(T:: *)(P1), method>
 {
+	using PM1 = std::remove_cv_t<std::remove_reference_t<P1>>;
 public:
 	static int invoke(lua_State * L)
 	{
@@ -265,7 +270,7 @@ public:
 		if(!S.tryGet(1,obj))
 			return S.retVoid();
 
-		P1 p1;
+		PM1 p1;
 		if(!S.tryGet(2, p1))
 			return S.retVoid();
 
@@ -279,6 +284,8 @@ public:
 template <typename U, typename T, typename R, typename P1, typename P2, R(T:: * method)(P1, P2)const>
 class LuaMethodWrapper <U, R(T:: *)(P1, P2)const, method>
 {
+	using PM1 = std::remove_cv_t<std::remove_reference_t<P1>>;
+	using PM2 = std::remove_cv_t<std::remove_reference_t<P2>>;
 public:
 	static int invoke(lua_State * L)
 	{
@@ -288,11 +295,11 @@ public:
 		if(!S.tryGet(1, obj))
 			return S.retVoid();
 
-		P1 p1;
+		PM1 p1;
 		if(!S.tryGet(2, p1))
 			return S.retVoid();
 
-		P2 p2;
+		PM2 p2;
 		if(!S.tryGet(3, p2))
 			return S.retVoid();
 
@@ -306,6 +313,8 @@ public:
 template <typename U, typename T, typename P1, typename P2, void(T:: * method)(P1, P2)const>
 class LuaMethodWrapper <U, void(T:: *)(P1, P2)const, method>
 {
+	using PM1 = std::remove_cv_t<std::remove_reference_t<P1>>;
+	using PM2 = std::remove_cv_t<std::remove_reference_t<P2>>;
 public:
 	static int invoke(lua_State * L)
 	{
@@ -315,11 +324,11 @@ public:
 		if(!S.tryGet(1, obj))
 			return S.retVoid();
 
-		P1 p1;
+		PM1 p1;
 		if(!S.tryGet(2, p1))
 			return S.retVoid();
 
-		P2 p2;
+		PM2 p2;
 		if(!S.tryGet(3, p2))
 			return S.retVoid();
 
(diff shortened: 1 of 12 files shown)


scripting/lua/LuaScriptingContext.cpp
```diff
@@ -206,7 +205,7 @@ JsonNode LuaContext::callGlobal(const std::string & name, const JsonNode & param
 
 		S.clear();
 
-		return JsonNode();
+		return {};
```
dydzio0614, line 208: I personally like previous version more

rilian-la-te: @dydzio0614 Will look for check which cause this behaviour.

rilian-la-te: @dydzio0614 removed modernization

rilian-la-te:
@IvanSavenko can you review? @dydzio0614 told me than I should ask you about a review)

scripting/lua/api/netpacks/BattleUnitsChanged.cpp
```diff
@@ -44,7 +44,7 @@ int BattleUnitsChangedProxy::add(lua_State * L)
 	if(!S.tryGet(1, object))
 		return S.retVoid();
 
-	uint32_t id;
+	uint32_t id = 0;
 
 	if(!S.tryGet(2, id))
 		return S.retVoid();
```
kambala-decapitator, line 50: I think this is superfluous. If I read the logic right: if `id` can't be filled from Lua, return immediately.

scripting/lua/api/netpacks/BattleUnitsChanged.cpp
```diff
@@ -70,7 +70,7 @@ int BattleUnitsChangedProxy::update(lua_State * L)
 	if(!S.tryGet(1, object))
 		return S.retVoid();
 
-	uint32_t id;
+	uint32_t id = 0;
```
kambala-decapitator, line 73: same here

kambala-decapitator (approved):

scripting/lua/api/netpacks/BattleUnitsChanged.cpp
rilian-la-te: @kambala-decapitator this is filled because clang-tidy initialized all variables. I can block this with //NOLINT, but do I need to block it?

scripting/lua/api/netpacks/BattleUnitsChanged.cpp
rilian-la-te: same as above.

scripting/lua/LuaCallWrapper.h
```diff
@@ -196,7 +197,7 @@ class LuaMethodWrapper <U, R(T:: *)(P1)const, method>
 		if(!S.tryGet(1,obj))
 			return S.retVoid();
 
-		P1 p1;
+		std::remove_cv_t<std::remove_reference_t<P1>> p1;
```
IvanSavenko, line 200: Why is this needed?

scripting/lua/LuaReference.cpp
```diff
@@ -23,7 +23,7 @@ LuaReference::LuaReference(lua_State * L)
 	key = luaL_ref(l, LUA_REGISTRYINDEX);
 }
 
-LuaReference::LuaReference(LuaReference && other)
+LuaReference::LuaReference(LuaReference && other) noexcept
```
IvanSavenko, line 26: Do we use `noexcept`? Is there any point in using it?
Not against this change but what are advantages of this form?

scripting/lua/LuaScriptModule.cpp
```diff
@@ -17,7 +17,7 @@
 #define strcpy_s(a, b, c) strncpy(a, c, b)
 #endif
 
-const char *g_cszAiName = "Lua interpreter";
+static const char *g_cszAiName = "Lua interpreter";
```
IvanSavenko, line 20: Frankly, the only useful change I noticed so far

scripting/lua/api/BonusSystem.h
```diff
@@ -54,7 +54,7 @@ class BonusListProxy : public SharedWrapper<const BonusList, BonusListProxy>
 
 	static const std::vector<typename Wrapper::CustomRegType> REGISTER_CUSTOM;
 
-	static std::shared_ptr<const Bonus> index(std::shared_ptr<const BonusList> self, int key);
+	static std::shared_ptr<const Bonus> index(std::shared_ptr<const BonusList> self, int key); //NOLINT
```
IvanSavenko, line 57: Again, comments that are not comment but attributes...

scripting/lua/api/netpacks/BattleUnitsChanged.cpp
IvanSavenko: Then don't use it? Or significantly clean up its output?
So far I don't see any significant improvements neither here nor in #1559

IvanSavenko (commented):
Don't see any significant improvements to be honest.
Not *against* these changes, but they are mostly redundant and have zero need to allocate time on.

scripting/lua/LuaCallWrapper.h
rilian-la-te: @kambala-decapitator In library series some types in VCMI library is converted from passing by value to passing by a const reference. If this change is not implemented, then Lua will breaks on const references passed in this templates. 

This change will strip constness and reference from such types, and just generate a new temporary value.

scripting/lua/LuaReference.cpp
rilian-la-te: @IvanSavenko standard library do not call move constructors if they are not marked noexcept, which leads to copy a variable everywhere. Maybe it this case this is not significant.

scripting/lua/LuaScriptModule.cpp
rilian-la-te: This change is not from modernization, I done it by hand, because Lua do not builds without it)

scripting/lua/LuaCallWrapper.h
IvanSavenko: To clarify - this is actual behavior that *is* happening and not yet another "improvement" from clang-tidy?

In that case perhaps better to add using at the very start of the class:
`using Parameter1 = td::remove_cv_t<std::remove_reference_t<P1>>;`
And then declare variables of type `Parameter1`

rilian-la-te: @IvanSavenko This is not another improvement from clang-tidy. But it will noticed only if you pass const reference to this templates.

Will do your suggestion.

scripting/lua/api/netpacks/BattleUnitsChanged.cpp
rilian-la-te: @IvanSavenko some constructors and destructors made empty by clang-tidy itself and got defaulted by next passes. Human would not detect it as is, if he will not review a code in details.

IvanSavenko: Example? Most complex case I saw is `ObstacleSideOptions::ObstacleSideOptions` [here](https://github.com/vcmi/vcmi/pull/1559/files#diff-b371b0903dfc2aa65381f853eef4f160dd6e624a48edc4ca095550124bcc9d27).
Still clearly an empty constructor that only has default-init of member and parent class.

Nice, but not something worth spending time on.

rilian-la-te: @IvanSavenko here is an example of defaulting a huge copy constructor from https://github.com/vcmi/vcmi/pull/1554/commits/0b22bf0f3f51ee0e102c2effaddb21ffe1b96c88#diff-7906420fb6fbc1c2a5f88c4dc055c77fa91e77aadf5b5f56dda376e2e85a2323

This is from first attempt of clang-tidy, but second attempt will do the same.

IvanSavenko: Nice, although seems to be the only case, and poorly written even from c++03 perspective.
Would be useful later, for c++20 default comparisons though.

rilian-la-te: @IvanSavenko I can remove checks you do not like, like init-all-variables and do modernization again)

rilian-la-te: @IvanSavenko cppcoreguidelines-init-all removed

scripting/lua/api/BonusSystem.h
rilian-la-te: Removed temporarily.

scripting/lua/LuaCallWrapper.h
rilian-la-te: @IvanSavenko your suggestion done for using.

scripting/lua/api/netpacks/BattleUnitsChanged.cpp
rilian-la-te: removed

kambala-decapitator:
looks fine to me, let's wait for @IvanSavenko's feedback

scripting/lua/LuaScriptingContext.cpp
```diff
@@ -35,12 +35,8 @@ namespace scripting
 const std::string LuaContext::STATE_FIELD = "DATA";
 
 LuaContext::LuaContext(const Script * source, const Environment * env_)
-	: ContextBase(env_->logger()),
-	script(source),
-	env(env_)
+	: ContextBase(env_->logger()), L(luaL_newstate()), script(source), env(env_)
```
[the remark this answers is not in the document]
rilian-la-te: done

scripting/lua/LuaStack.cpp
```diff
@@ -20,9 +20,8 @@ namespace scripting
 {
 
 LuaStack::LuaStack(lua_State * L_)
-	: L(L_)
+	: L(L_), initialTop(lua_gettop(L))
```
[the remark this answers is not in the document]
rilian-la-te: done

rilian-la-te:
@IvanSavenko waiting for merge this and #1559 )

PR #1559, opened 2023-02-05
Modernize vstd spell effects

rilian-la-te:
This is a second attempt to modernize VCMI to C++14 using clang-tidy.

This is a part 1 - only vstd and spell effects.

### DIFF
--- a/lib/spells/effects/Catapult.cpp
+++ b/lib/spells/effects/Catapult.cpp
@@ -32,17 +32,9 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Catapult, EFFECT_NAME);
 
-Catapult::Catapult()
-	: LocationEffect(),
-	targetsToAttack(0)
-{
-}
-
-Catapult::~Catapult() = default;
-
 bool Catapult::applicable(Problem & problem, const Mechanics * m) const
 {
-	auto town = m->battle()->battleGetDefendedTown();
+	const auto *town = m->battle()->battleGetDefendedTown();
 
 	if(nullptr == town)
 	{
@@ -113,7 +105,7 @@ void Catapult::apply(ServerCallback * server, const Mechanics * m, const EffectT
 
 		if (attackInfo == ca.attackedParts.end()) // new part
 		{
-			CatapultAttack::AttackInfo newInfo;
+			CatapultAttack::AttackInfo newInfo{};
 			newInfo.damageDealt = 1;
 			newInfo.attackedPart = target;
 			newInfo.destinationTile = m->battle()->wallPartToBattleHex(target);
--- a/lib/spells/effects/Clone.cpp
+++ b/lib/spells/effects/Clone.cpp
@@ -28,14 +28,6 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Clone, EFFECT_NAME);
 
-Clone::Clone()
-	: UnitEffect(),
-	maxTier(0)
-{
-}
-
-Clone::~Clone() = default;
-
 void Clone::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const
 {
 	for(const Destination & dest : target)
@@ -80,7 +72,7 @@ void Clone::apply(ServerCallback * server, const Mechanics * m, const EffectTarg
 
 		BattleUnitsChanged cloneFlags;
 
-		auto cloneUnit = m->battle()->battleGetUnitByID(unitId);
+		const auto *cloneUnit = m->battle()->battleGetUnitByID(unitId);
 
 		if(!cloneUnit)
 		{
@@ -105,7 +97,7 @@ void Clone::apply(ServerCallback * server, const Mechanics * m, const EffectTarg
 		lifeTimeMarker.turnsRemain = m->getEffectDuration();
 		std::vector<Bonus> buffer;
 		buffer.push_back(lifeTimeMarker);
-		sse.toAdd.push_back(std::make_pair(unitId, buffer));
+		sse.toAdd.emplace_back(unitId, buffer);
 		server->apply(&sse);
 	}
 }
--- a/lib/spells/effects/Damage.cpp
+++ b/lib/spells/effects/Damage.cpp
@@ -31,15 +31,6 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Damage, EFFECT_NAME);
 
-Damage::Damage()
-	: UnitEffect(),
-	killByPercentage(false),
-	killByCount(false)
-{
-}
-
-Damage::~Damage() = default;
-
 void Damage::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const
 {
 	StacksInjured stacksInjured;
@@ -52,7 +43,7 @@ void Damage::apply(ServerCallback * server, const Mechanics * m, const EffectTar
 	uint32_t killed = 0;
 	bool multiple = false;
 
-	for(auto & t : target)
+	for(const auto & t : target)
 	{
 		const battle::Unit * unit = t.unitValue;
 		if(unit && unit->alive())
--- a/lib/spells/effects/DemonSummon.cpp
+++ b/lib/spells/effects/DemonSummon.cpp
@@ -28,15 +28,6 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(DemonSummon, EFFECT_NAME);
 
-DemonSummon::DemonSummon()
-	: UnitEffect()
-	, creature(0)
-	, permanent(false)
-{
-}
-
-DemonSummon::~DemonSummon() = default;
-
 void DemonSummon::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const
 {
 	BattleUnitsChanged pack;
@@ -60,7 +51,7 @@ void DemonSummon::apply(ServerCallback * server, const Mechanics * m, const Effe
 			break;
 		}
 
-		auto creatureType = creature.toCreature(m->creatures());
+		const auto *creatureType = creature.toCreature(m->creatures());
 
 		int32_t deadCount         = targetStack->unitBaseAmount();
 		int32_t deadTotalHealth   = targetStack->getTotalHealth();
@@ -110,7 +101,7 @@ bool DemonSummon::isValidTarget(const Mechanics * m, const battle::Unit * s) con
 	if (s->isGhost())
 		return false;
 
-	auto creatureType = creature.toCreature(m->creatures());
+	const auto *creatureType = creature.toCreature(m->creatures());
 
 	if (s->getTotalHealth() < creatureType->getMaxHealth())
 		return false;
--- a/lib/spells/effects/Dispel.h
+++ b/lib/spells/effects/Dispel.h
@@ -27,9 +27,6 @@ namespace effects
 class Dispel : public UnitEffect
 {
 public:
-	Dispel();
-	virtual ~Dispel();
-
 	void apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const override;
 
 protected:
--- a/lib/spells/effects/Heal.cpp
+++ b/lib/spells/effects/Heal.cpp
@@ -31,17 +31,6 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Heal, EFFECT_NAME);
 
-Heal::Heal()
-	: UnitEffect(),
-	healLevel(EHealLevel::HEAL),
-	healPower(EHealPower::PERMANENT),
-	minFullUnits(0)
-{
-
-}
-
-Heal::~Heal() = default;
-
 void Heal::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const
 {
 	apply(m->getEffectValue(), server, m, target);
@@ -117,7 +106,7 @@ void Heal::serializeJsonUnitEffect(JsonSerializeFormat & handler)
 
 void Heal::prepareHealEffect(int64_t value, BattleUnitsChanged & pack, BattleLogMessage & logMessage, RNG & rng, const Mechanics * m, const EffectTarget & target) const
 {
-	for(auto & oneTarget : target)
+	for(const auto & oneTarget : target)
 	{
 		const battle::Unit * unit = oneTarget.unitValue;
 
--- a/lib/spells/effects/LocationEffect.h
+++ b/lib/spells/effects/LocationEffect.h
@@ -23,9 +23,6 @@ namespace effects
 class LocationEffect : public Effect
 {
 public:
-	LocationEffect();
-	virtual ~LocationEffect();
-
 	void adjustTargetTypes(std::vector<TargetType> & types) const override;
 
 	void adjustAffectedHexes(std::set<BattleHex> & hexes, const Mechanics * m, const Target & spellTarget) const override;
--- a/lib/spells/effects/Obstacle.cpp
+++ b/lib/spells/effects/Obstacle.cpp
@@ -30,27 +30,9 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Obstacle, EFFECT_NAME);
 
-ObstacleSideOptions::ObstacleSideOptions()
-	: shape(),
-	range()
-{
-}
+using RelativeShape = std::vector<std::vector<BattleHex::EDir>>;
 
-void ObstacleSideOptions::serializeJson(JsonSerializeFormat & handler)
-{
-	serializeRelativeShape(handler, "shape", shape);
-	serializeRelativeShape(handler, "range", range);
-
-	handler.serializeString("appearSound", appearSound);
-	handler.serializeString("appearAnimation", appearAnimation);
-	handler.serializeString("triggerSound", triggerSound);
-	handler.serializeString("triggerAnimation", triggerAnimation);
-	handler.serializeString("animation", animation);
-
-	handler.serializeInt("offsetY", offsetY);
-}
-
-void ObstacleSideOptions::serializeRelativeShape(JsonSerializeFormat & handler, const std::string & fieldName, RelativeShape & value)
+static void serializeRelativeShape(JsonSerializeFormat & handler, const std::string & fieldName, RelativeShape & value)
 {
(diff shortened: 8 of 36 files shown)


lib/spells/effects/Damage.cpp
```diff
@@ -115,7 +114,7 @@ void Damage::serializeJsonUnitEffect(JsonSerializeFormat & handler)
 
 int64_t Damage::damageForTarget(size_t targetIndex, const Mechanics * m, const battle::Unit * target) const
 {
-	int64_t baseDamage;
+	int64_t baseDamage = 0;
```
kambala-decapitator, line 117: not sure why this is needed - the variable is initialised in all code paths

lib/spells/effects/Dispel.h
```diff
@@ -27,8 +27,7 @@ namespace effects
 class Dispel : public UnitEffect
 {
 public:
-	Dispel();
-	virtual ~Dispel();
+	virtual ~Dispel() = default;
```
kambala-decapitator, line 30: 1. isn't destructor already virtual in UnitEffect?
2. I'd remove it completely instead of using `default`

lib/spells/effects/LocationEffect.h
```diff
@@ -23,8 +23,7 @@ namespace effects
 class LocationEffect : public Effect
 {
 public:
-	LocationEffect();
-	virtual ~LocationEffect();
+	virtual ~LocationEffect() override = default;
```
kambala-decapitator, line 26: same points about destructor

lib/spells/effects/RemoveObstacle.h
```diff
@@ -28,7 +28,7 @@ class RemoveObstacle : public LocationEffect
 {
 public:
 	RemoveObstacle();
-	virtual ~RemoveObstacle();
+	virtual ~RemoveObstacle() override = default;
```
kambala-decapitator, line 31: same

lib/spells/effects/Sacrifice.h
```diff
@@ -22,8 +22,7 @@ namespace effects
 class Sacrifice : public Heal
 {
 public:
-	Sacrifice();
-	virtual ~Sacrifice();
+	virtual ~Sacrifice() override = default;
```
kambala-decapitator, line 25: same

lib/spells/effects/Teleport.h
```diff
@@ -24,8 +24,7 @@ namespace effects
 class Teleport : public UnitEffect
 {
 public:
-	Teleport();
-	virtual ~Teleport();
+	virtual ~Teleport() override = default;
```
kambala-decapitator, line 27: same

lib/spells/effects/Dispel.h
rilian-la-te: @kambala-decapitator will look for it. If it is already virtual, it is better to add override instead.

lib/spells/effects/LocationEffect.h
rilian-la-te: @kambala-decapitator about removal - it can break qualifiers. By default destructor is not public AFAIK. But if destructor is here, it remain public. We need test if we need a public destructors of spell effects or not.

kambala-decapitator: missing destructor is equivalent to having public default destructor

> it can break qualifiers

sorry, I'm not following. Which qualifiers?

lib/spells/effects/Catapult.cpp
```diff
@@ -113,7 +112,7 @@ void Catapult::apply(ServerCallback * server, const Mechanics * m, const EffectT
 
 		if (attackInfo == ca.attackedParts.end()) // new part
 		{
-			CatapultAttack::AttackInfo newInfo;
+			CatapultAttack::AttackInfo newInfo; //NOLINT - braced init will not work due to -Wnarrowing
```
IvanSavenko, line 115: Not a big fan of having comments that are not comments but metainfo.

lib/spells/effects/Dispel.h
IvanSavenko: Just remove it. There is not need for virtual default destructors in derived classes.

lib/spells/effects/Obstacle.cpp
```diff
@@ -103,7 +97,7 @@ void ObstacleSideOptions::serializeRelativeShape(JsonSerializeFormat & handler,
 }
 
 Obstacle::Obstacle()
-	: LocationEffect(),
+	:
```
IvanSavenko, line 100: Weird formatting

lib/spells/effects/RemoveObstacle.cpp
```diff
@@ -32,15 +32,12 @@ namespace effects
 VCMI_REGISTER_SPELL_EFFECT(RemoveObstacle, EFFECT_NAME);
 
 RemoveObstacle::RemoveObstacle()
-	: LocationEffect(),
-	removeAbsolute(false),
-	removeUsual(false),
-	removeAllSpells(false)
+	: removeAbsolute(false),
```
IvanSavenko, line 35: Weird formatting. Might need change in clang-format but please use either
```
RemoveObstacle::RemoveObstacle()
	: removeAbsolute(false)
	, removeUsual(false)
```
or
```
RemoveObstacle::RemoveObstacle():
	removeAbsolute(false),
	removeUsual(false)
```

lib/spells/effects/LocationEffect.h
rilian-la-te: @kambala-decapitator if public, you are right. I thougt than removed member of class is protected.

lib/spells/effects/Catapult.cpp
rilian-la-te: @kambala-decapitator this comment will block clang-tidy from adding a bracing empty init.

I tried to add a braced init myself - but there is -Wnarrowing, because battleHex will return 16-bit, but structure accepts 8 bit.

lib/spells/effects/Obstacle.cpp
rilian-la-te: Overlooked, will fix.

lib/spells/effects/RemoveObstacle.cpp
rilian-la-te: Okay. I prefer the latter actually, but this formatting was there before clang-tidy was modernized it.

lib/spells/effects/Damage.cpp
rilian-la-te: It is just default init of clang-tidy. I do not think than it is bad. If clang-tidy wants to add it - why not?

lib/spells/effects/RemoveObstacle.h
IvanSavenko: Destructors can be marked as override? Nice, but pointless since base class has virtual destructor anyway.

IvanSavenko (commented):
Useful changes:
- removal of empty constructors & destructors

Micro-optimizations - not *bad* changes, but likely give zero effect:
- copy -> const reference
- push_back -> emplace_back

Noise:
- zero-init of variables. I think I actually prefer current style, as an indicator that all code paths should initialize it. Extra bonus is that any sane compiler will issue warning if this is not the case. But zero-init removes said warning.
- default constructors & destructors. Should be deleted entirely

lib/spells/effects/Damage.cpp
IvanSavenko: IMO this is bad change.
Old behavior: if damage is not set on some code paths, compiler will issue "uninitialized" warning.
New behavior: if damage is not set on some code path, it will be zero. Not desired behavior.

rilian-la-te: @IvanSavenko okay, will try to do modernization without cppcoreguidelines-init-variables filter.

rilian-la-te: @IvanSavenko removed

lib/spells/effects/LocationEffect.h
rilian-la-te: @kambala-decapitator removed

lib/spells/effects/Sacrifice.h
rilian-la-te: removed

lib/spells/effects/Obstacle.cpp
rilian-la-te: Fixed.

lib/spells/effects/RemoveObstacle.cpp
rilian-la-te: @IvanSavenko did all the latter style

lib/spells/effects/Teleport.h
rilian-la-te: removed

lib/spells/effects/RemoveObstacle.h
rilian-la-te: removed

lib/spells/effects/Dispel.h
rilian-la-te: removed

lib/spells/effects/Clone.cpp
```diff
@@ -28,9 +28,7 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Clone, EFFECT_NAME);
 
-Clone::Clone()
-	: UnitEffect(),
-	maxTier(0)
+Clone::Clone() : maxTier(0)
```
kambala-decapitator, line 31: why did clang-tidy decide not to initialise `maxTier` in declaration? :)

lib/spells/effects/DemonSummon.cpp
```diff
@@ -29,8 +29,7 @@ namespace effects
 VCMI_REGISTER_SPELL_EFFECT(DemonSummon, EFFECT_NAME);
 
 DemonSummon::DemonSummon()
-	: UnitEffect()
-	, creature(0)
+	: creature(0)
 	, permanent(false)
```
kambala-decapitator, line 33: same

lib/spells/effects/Heal.cpp
```diff
@@ -31,8 +31,7 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Heal, EFFECT_NAME);
 
-Heal::Heal()
-	: UnitEffect(),
+Heal::Heal():
 	healLevel(EHealLevel::HEAL),
 	healPower(EHealPower::PERMANENT),
 	minFullUnits(0)
```
kambala-decapitator, line 37: same

lib/spells/effects/Obstacle.cpp
```diff
@@ -102,8 +96,7 @@ void ObstacleSideOptions::serializeRelativeShape(JsonSerializeFormat & handler,
 	}
 }
 
-Obstacle::Obstacle()
-	: LocationEffect(),
+Obstacle::Obstacle():
 	hidden(false),
 	passable(false),
 	trigger(false),
```
kambala-decapitator, line 102: same

lib/spells/effects/RemoveObstacle.cpp
```diff
@@ -31,16 +31,13 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(RemoveObstacle, EFFECT_NAME);
 
-RemoveObstacle::RemoveObstacle()
-	: LocationEffect(),
+RemoveObstacle::RemoveObstacle():
 	removeAbsolute(false),
 	removeUsual(false),
 	removeAllSpells(false)
```
kambala-decapitator, line 37: same

lib/spells/effects/Registry.cpp
```diff
@@ -24,7 +24,7 @@ class RegistryImpl : public Registry
 {
 public:
 	RegistryImpl() = default;
-	~RegistryImpl() = default;
+	~RegistryImpl() override = default;
```
kambala-decapitator, line 27: these can also be removed actually

lib/spells/effects/Sacrifice.h
```diff
@@ -38,7 +35,7 @@ class Sacrifice : public Heal
 	bool isValidTarget(const Mechanics * m, const battle::Unit * unit) const override;
 
 private:
-	int64_t calculateHealEffectValue(const Mechanics * m, const battle::Unit * victim ) const;
+	static int64_t calculateHealEffectValue(const Mechanics * m, const battle::Unit * victim ) ;
```
kambala-decapitator, line 38:
```suggestion
	static int64_t calculateHealEffectValue(const Mechanics * m, const battle::Unit * victim);
```

lib/spells/effects/Summon.cpp
```diff
@@ -34,9 +34,7 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Summon, EFFECT_NAME);
 
-Summon::Summon()
-	: Effect(),
-	creature(),
+Summon::Summon():
 	permanent(false),
 	exclusive(true),
 	summonByHealth(false),
```
kambala-decapitator, line 40: same

lib/spells/effects/Timed.cpp
```diff
@@ -29,10 +29,7 @@ namespace effects
 
 VCMI_REGISTER_SPELL_EFFECT(Timed, EFFECT_NAME);
 
-Timed::Timed()
-	: UnitEffect(),
-	cumulative(false),
-	bonus()
+Timed::Timed() : cumulative(false)
```
kambala-decapitator, line 32: same

lib/spells/effects/Timed.h
```diff
@@ -39,7 +39,7 @@ class Timed : public UnitEffect
 
 private:
 	void convertBonus(const Mechanics * m, int32_t & duration, std::vector<Bonus> & converted) const;
-	void describeEffect(std::vector<MetaString> & log, const Mechanics * m, const std::vector<Bonus> & bonuses, const battle::Unit * target) const;
+	static void describeEffect(std::vector<MetaString> & log, const Mechanics * m, const std::vector<Bonus> & bonuses, const battle::Unit * target) ;
```
kambala-decapitator, line 42:
```suggestion
	static void describeEffect(std::vector<MetaString> & log, const Mechanics * m, const std::vector<Bonus> & bonuses, const battle::Unit * target);
```

kambala-decapitator (approved):
looks good to me, but let's also wait for @IvanSavenko's feedback.

not insisting on changing the stuff I commented on (except code suggestions).

lib/spells/effects/Clone.cpp
rilian-la-te: I do not know. I ran 2 versions of it, and this is actually version 14. Maybe in version 15 it have better algorithms.

lib/spells/effects/Timed.cpp
rilian-la-te: done

lib/spells/effects/Summon.cpp
rilian-la-te: done

lib/spells/effects/Sacrifice.h
rilian-la-te: done

lib/spells/effects/Registry.cpp
rilian-la-te: done

lib/spells/effects/RemoveObstacle.cpp
rilian-la-te: done

lib/spells/effects/Obstacle.cpp
rilian-la-te: done

lib/spells/effects/Heal.cpp
rilian-la-te: done

lib/spells/effects/DemonSummon.cpp
rilian-la-te: done

lib/spells/effects/Clone.cpp
rilian-la-te: Done by hand

lib/spells/effects/Obstacle.h
```diff
@@ -35,22 +35,17 @@ class ObstacleSideOptions
...
-	int offsetY;
-
-	ObstacleSideOptions();
+	int offsetY = 0;
 
 	void serializeJson(JsonSerializeFormat & handler);
 
 private:
-	void serializeRelativeShape(JsonSerializeFormat & handler, const std::string & fieldName, RelativeShape & value);
+	static void serializeRelativeShape(JsonSerializeFormat & handler, const std::string & fieldName, RelativeShape & value);
```
IvanSavenko, line 43: (minor issue, nothing critical)
Don't see any reason to declare member static unless that member *has* to be called without instance.

lib/spells/effects/Timed.h
```diff
@@ -26,20 +26,17 @@ namespace effects
...
-
 	void apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const override;
 
 protected:
 	void serializeJsonUnitEffect(JsonSerializeFormat & handler) override final;
 
 private:
 	void convertBonus(const Mechanics * m, int32_t & duration, std::vector<Bonus> & converted) const;
-	void describeEffect(std::vector<MetaString> & log, const Mechanics * m, const std::vector<Bonus> & bonuses, const battle::Unit * target) const;
+	static void describeEffect(std::vector<MetaString> & log, const Mechanics * m, const std::vector<Bonus> & bonuses, const battle::Unit * target);
```
IvanSavenko, line 39: Same here

IvanSavenko (approved):

lib/spells/effects/Obstacle.h
rilian-la-te: @IvanSavenko I think than if even clang-tidy tries to declare members static, we can consider to move it from a class. A bonus - static register does not require hidden class parameter)

IvanSavenko: In that case move it from a class. Convert it into a non-member static function in .cpp file
(not possible if this method requires access to private members)

lib/spells/effects/Timed.h
rilian-la-te: done

lib/spells/effects/Obstacle.h
rilian-la-te: done

PR #1561, opened 2023-02-08
VCMI: modernize lib/serializer and modernize lib/spells

rilian-la-te:
More modernizations. Splitted to be easy for review

### DIFF
--- a/lib/serializer/BinaryDeserializer.cpp
+++ b/lib/serializer/BinaryDeserializer.cpp
@@ -24,13 +24,12 @@ CLoadFile::CLoadFile(const boost::filesystem::path & fname, int minimalVersion)
 	openNextFile(fname, minimalVersion);
 }
 
-CLoadFile::~CLoadFile()
-{
-}
+//must be instantiated in .cpp file for access to complete types of all member fields
+CLoadFile::~CLoadFile() = default;
 
 int CLoadFile::read(void * data, unsigned size)
 {
-	sfile->read((char*)data,size);
+	sfile->read(reinterpret_cast<char *>(data), size);
 	return size;
 }
 
@@ -51,7 +50,7 @@ void CLoadFile::openNextFile(const boost::filesystem::path & fname, int minimalV
 		//we can read
 		char buffer[4];
 		sfile->read(buffer, 4);
-		if(std::memcmp(buffer,"VCMI",4))
+		if(std::memcmp(buffer, "VCMI", 4) != 0)
 			THROW_FORMAT("Error: not a VCMI file(%s)!", fName);
 
 		serializer & serializer.fileVersion;
@@ -62,7 +61,7 @@ void CLoadFile::openNextFile(const boost::filesystem::path & fname, int minimalV
 		{
 			logGlobal->warn("Warning format version mismatch: found %d when current is %d! (file %s)\n", serializer.fileVersion, SERIALIZATION_VERSION , fName);
 
-			auto versionptr = (char*)&serializer.fileVersion;
+			auto * versionptr = reinterpret_cast<char *>(&serializer.fileVersion);
 			std::reverse(versionptr, versionptr + 4);
 			logGlobal->warn("Version number reversed is %x, checking...", serializer.fileVersion);
 
@@ -99,7 +98,7 @@ void CLoadFile::clear()
 void CLoadFile::checkMagicBytes(const std::string &text)
 {
 	std::string loaded = text;
-	read((void*)loaded.data(), (unsigned int)text.length());
+	read((void *)loaded.data(), static_cast<unsigned int>(text.length()));
 	if(loaded != text)
 		throw std::runtime_error("Magic bytes doesn't match!");
 }
--- a/lib/serializer/BinarySerializer.cpp
+++ b/lib/serializer/BinarySerializer.cpp
@@ -24,9 +24,8 @@ CSaveFile::CSaveFile(const boost::filesystem::path &fname)
 	openNextFile(fname);
 }
 
-CSaveFile::~CSaveFile()
-{
-}
+//must be instantiated in .cpp file for access to complete types of all member fields
+CSaveFile::~CSaveFile() = default;
 
 int CSaveFile::write(const void * data, unsigned size)
 {
@@ -73,7 +72,7 @@ void CSaveFile::clear()
 
 void CSaveFile::putMagicBytes(const std::string &text)
 {
-	write(text.c_str(), (unsigned int)text.length());
+	write(text.c_str(), static_cast<unsigned int>(text.length()));
 }
 
 VCMI_LIB_NAMESPACE_END
--- a/lib/serializer/CMemorySerializer.cpp
+++ b/lib/serializer/CMemorySerializer.cpp
@@ -32,13 +32,11 @@ int CMemorySerializer::write(const void * data, unsigned size)
 	return size;
 }
 
-CMemorySerializer::CMemorySerializer(): iser(this), oser(this)
+CMemorySerializer::CMemorySerializer(): iser(this), oser(this), readPos(0)
 {
-	readPos = 0;
 	registerTypes(iser);
 	registerTypes(oser);
 	iser.fileVersion = SERIALIZATION_VERSION;
 }
-
 
 VCMI_LIB_NAMESPACE_END
--- a/lib/serializer/CSerializer.h
+++ b/lib/serializer/CSerializer.h
@@ -80,10 +80,9 @@ class DLL_LINKAGE CSerializer
 	TTypeVecMap vectors; //entry must be a pointer to vector containing pointers to the objects of key type
 
 public:
-	bool smartVectorMembersSerialization;
-	bool sendStackInstanceByIds;
+	bool smartVectorMembersSerialization = false;
+	bool sendStackInstanceByIds = false;
 
-	CSerializer();
 	~CSerializer();
 
 	virtual void reportState(vstd::CLoggerBase * out){};
--- a/lib/serializer/CTypeList.h
+++ b/lib/serializer/CTypeList.h
@@ -109,12 +109,7 @@ private:
 
 		return ptr;
 	}
-	CTypeList &operator=(CTypeList &)
-	{
-		// As above.
-		assert(0);
-		return *this;
-	}
+	CTypeList & operator=(CTypeList &) = delete;
 
 	TypeInfoPtr getTypeDescriptor(const std::type_info *type, bool throws = true) const; //if not throws, failure returns nullptr
 	TypeInfoPtr registerType(const std::type_info *type);
--- a/lib/serializer/Connection.h
+++ b/lib/serializer/Connection.h
@@ -88,8 +88,8 @@ public:
 	int connectionID;
 	std::shared_ptr<boost::thread> handler;
 
-	CConnection(std::string host, ui16 port, std::string Name, std::string UUID);
-	CConnection(std::shared_ptr<TAcceptor> acceptor, std::shared_ptr<boost::asio::io_service> Io_service, std::string Name, std::string UUID);
+	CConnection(const std::string & host, ui16 port, std::string Name, std::string UUID);
+	CConnection(const std::shared_ptr<TAcceptor> & acceptor, const std::shared_ptr<boost::asio::io_service> & Io_service, std::string Name, std::string UUID);
 	CConnection(std::shared_ptr<TSocket> Socket, std::string Name, std::string UUID); //use immediately after accepting connection into socket
 
 	void close();
--- a/lib/serializer/ILICReader.cpp
+++ b/lib/serializer/ILICReader.cpp
@@ -1,0 +1,49 @@
+/*
+ * JsonTreeSerializer.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#include "StdInc.h"
+#include "ILICReader.h"
+
+#include "../JsonNode.h"
+
+#include <vstd/StringUtils.h>
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+void ILICReader::readLICPart(const JsonNode & part, const JsonSerializeFormat::TDecoder & decoder, bool val, std::vector<bool> & value) const
+{
+	for(const auto & index : part.Vector())
+	{
+		const std::string & identifier = index.String();
+		const std::string type = typeid(decltype(this)).name();
+
+		const si32 rawId = decoder(identifier);
+		if(rawId >= 0)
+		{
+			if(rawId < value.size())
+				value[rawId] = val;
+			else
+				logGlobal->error("%s::serializeLIC: id out of bounds %d", type, rawId);
+		}
+	}
+}
+
+void ILICReader::readLICPart(const JsonNode & part, const JsonSerializeFormat::TDecoder & decoder, std::set<si32> & value) const
+{
+	for(const auto & index : part.Vector())
+	{
+		const std::string & identifier = index.String();
+
+		const si32 rawId = decoder(identifier);
+		if(rawId != -1)
+			value.insert(rawId);
+	}
+}
+
+VCMI_LIB_NAMESPACE_END
--- a/lib/serializer/ILICReader.h
+++ b/lib/serializer/ILICReader.h
@@ -1,0 +1,23 @@
+/*
+ * ILICReader.h, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#pragma once
+
+#include "JsonTreeSerializer.h"
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+class DLL_LINKAGE ILICReader
+{
+protected:
+	void readLICPart(const JsonNode & part, const JsonSerializeFormat::TDecoder & decoder, bool val, std::vector<bool> & value) const;
+	void readLICPart(const JsonNode & part, const JsonSerializeFormat::TDecoder & decoder, std::set<si32> & value) const;
+};
+
+VCMI_LIB_NAMESPACE_END
--- a/lib/serializer/JsonDeserializer.h
+++ b/lib/serializer/JsonDeserializer.h
@@ -9,11 +9,12 @@
  */
 #pragma once
 
-#include  "JsonTreeSerializer.h"
+#include "ILICReader.h"
+#include "JsonTreeSerializer.h"
 
 VCMI_LIB_NAMESPACE_BEGIN
 
-class DLL_LINKAGE JsonDeserializer: public JsonTreeSerializer<const JsonNode *>
+class DLL_LINKAGE JsonDeserializer: public JsonTreeSerializer<const JsonNode *>, public ILICReader
 {
 public:
 	JsonDeserializer(const IInstanceResolver * instanceResolver_, const JsonNode & root_);
@@ -35,10 +36,6 @@ protected:
 
 	void serializeInternal(std::string & value) override;
 	void serializeInternal(int64_t & value) override;
-
-private:
-	void readLICPart(const JsonNode & part, const TDecoder & decoder, const bool val, std::vector<bool> & value);
-	void readLICPart(const JsonNode & part, const TDecoder & decoder, std::set<si32> & value);
 };
 
 VCMI_LIB_NAMESPACE_END
--- a/lib/serializer/JsonSerializeFormat.cpp
+++ b/lib/serializer/JsonSerializeFormat.cpp
@@ -15,9 +15,7 @@
 VCMI_LIB_NAMESPACE_BEGIN
 
 //JsonSerializeHelper
-JsonSerializeHelper::JsonSerializeHelper(JsonSerializeHelper && other):
-	owner(other.owner),
-	restoreState(false)
+JsonSerializeHelper::JsonSerializeHelper(JsonSerializeHelper && other) noexcept: owner(other.owner), restoreState(false)
 {
 	std::swap(restoreState, other.restoreState);
 }
@@ -40,33 +38,17 @@ JsonSerializeHelper::JsonSerializeHelper(JsonSerializeFormat * owner_)
 }
 
 //JsonStructSerializer
-JsonStructSerializer::JsonStructSerializer(JsonStructSerializer && other)
-	: JsonSerializeHelper(std::move(static_cast<JsonSerializeHelper &>(other)))
-{
-
-}
+JsonStructSerializer::JsonStructSerializer(JsonStructSerializer && other) noexcept: JsonSerializeHelper(std::move(static_cast<JsonSerializeHelper &>(other))) {}
 
 JsonStructSerializer::JsonStructSerializer(JsonSerializeFormat * owner_)
 	: JsonSerializeHelper(owner_)
 {
 }
 
-JsonStructSerializer::~JsonStructSerializer()
-{
-}
-
 //JsonArraySerializer
-JsonArraySerializer::JsonArraySerializer(JsonArraySerializer && other)
-	: JsonSerializeHelper(std::move(static_cast<JsonSerializeHelper &>(other)))
-{
+JsonArraySerializer::JsonArraySerializer(JsonArraySerializer && other) noexcept: JsonSerializeHelper(std::move(static_cast<JsonSerializeHelper &>(other))) {}
 
-}
-
-JsonArraySerializer::JsonArraySerializer(JsonSerializeFormat * owner_):
-	JsonSerializeHelper(owner_)
-{
-	thisNode = &owner->getCurrent();
-}
+JsonArraySerializer::JsonArraySerializer(JsonSerializeFormat * owner_): JsonSerializeHelper(owner_), thisNode(&owner->getCurrent()) {}
 
 JsonStructSerializer JsonArraySerializer::enterStruct(const size_t index)
 {
@@ -110,16 +92,20 @@ size_t JsonArraySerializer::size() const
 }
 
 //JsonSerializeFormat::LIC
-JsonSerializeFormat::LIC::LIC(const std::vector<bool> & Standard, const TDecoder Decoder, const TEncoder Encoder):
-	standard(Standard), decoder(Decoder), encoder(Encoder)
+JsonSerializeFormat::LIC::LIC(const std::vector<bool> & Standard, TDecoder Decoder, TEncoder Encoder):
+	standard(Standard),
+	decoder(std::move(Decoder)),
+	encoder(std::move(Encoder))
 {
 	any.resize(standard.size(), false);
 	all.resize(standard.size(), false);
 	none.resize(standard.size(), false);
 }
 
-JsonSerializeFormat::LICSet::LICSet(const std::set<si32>& Standard, const TDecoder Decoder, const TEncoder Encoder):
-	standard(Standard), decoder(Decoder), encoder(Encoder)
+JsonSerializeFormat::LICSet::LICSet(const std::set<si32> & Standard, TDecoder Decoder, TEncoder Encoder):
+	standard(Standard),
+	decoder(std::move(Decoder)),
+	encoder(std::move(Encoder))
 {
 
 }
--- a/lib/serializer/JsonUpdater.cpp
+++ b/lib/serializer/JsonUpdater.cpp
@@ -245,7 +245,7 @@ void JsonUpdater::serializeBonuses(const std::string & fieldName, CBonusSystemNo
 
 	if(toAdd.getType() == JsonNode::JsonType::DATA_VECTOR)
 	{
-		for(auto & item : toAdd.Vector())
+		for(const auto & item : toAdd.Vector())
 		{
 			auto b = JsonUtils::parseBonus(item);
 			value->addNewBonus(b);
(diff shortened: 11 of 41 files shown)


lib/serializer/BinaryDeserializer.cpp
```diff
@@ -24,9 +24,8 @@ CLoadFile::CLoadFile(const boost::filesystem::path & fname, int minimalVersion)
 	openNextFile(fname, minimalVersion);
 }
 
-CLoadFile::~CLoadFile()
-{
-}
+///This destructor should be located here to avoid side effects
```
IvanSavenko, line 27: What side effects? Please describe reason in comment in code

lib/serializer/BinarySerializer.cpp
```diff
@@ -24,9 +24,8 @@ CSaveFile::CSaveFile(const boost::filesystem::path &fname)
 	openNextFile(fname);
 }
 
-CSaveFile::~CSaveFile()
-{
-}
+///This destructor should be located here to avoid side effects
```
IvanSavenko, line 27: Same here

lib/serializer/CMemorySerializer.h
```diff
@@ -20,7 +20,7 @@ class DLL_LINKAGE CMemorySerializer
 {
 	std::vector<ui8> buffer;
 
-	size_t readPos; //index of the next byte to be read
+	size_t readPos = 0; //index of the next byte to be read
```
IvanSavenko, line 23: Probably better to put this in constructor, to have initialization of all members in a single location.

lib/serializer/BinaryDeserializer.cpp
```diff
@@ -62,7 +61,7 @@ void CLoadFile::openNextFile(const boost::filesystem::path & fname, int minimalV
 		{
 			logGlobal->warn("Warning format version mismatch: found %d when current is %d! (file %s)\n", serializer.fileVersion, SERIALIZATION_VERSION , fName);
 
-			auto versionptr = (char*)&serializer.fileVersion;
+			auto *versionptr = (char*)&serializer.fileVersion;
```
IvanSavenko, line 64: I thought that clang-tidy would replace c-style casts with static_cast/reinterpret_cast?

lib/serializer/CTypeList.h
```diff
@@ -85,7 +85,7 @@ class DLL_LINKAGE CTypeList: public boost::noncopyable
 
 	/// Returns sequence of types starting from "from" and ending on "to". Every next type is derived from the previous.
 	/// Throws if there is no link registered.
-	std::vector<TypeInfoPtr> castSequence(TypeInfoPtr from, TypeInfoPtr to) const;
+	static std::vector<TypeInfoPtr> castSequence(TypeInfoPtr from, TypeInfoPtr to);
```
IvanSavenko, line 88: Again, don't see a reason to declare method static unless method *needs* to be called without isntance of object.
Please 
a) convert this into a static function insider .cpp file
b) revert change (probably better IMO)

lib/serializer/Connection.h
```diff
@@ -79,17 +79,17 @@ class DLL_LINKAGE CConnection
 	std::shared_ptr<boost::mutex> mutexRead;
 	std::shared_ptr<boost::mutex> mutexWrite;
 	std::shared_ptr<TSocket> socket;
-	bool connected;
-	bool myEndianess, contactEndianess; //true if little endian, if endianness is different we'll have to revert received multi-byte vars
+	bool connected = false;
```
IvanSavenko, line 82: Please, keep initialization of all members in one place, either in constructor or in-place.
Don't want to play "find where this field is initialized" game.

lib/serializer/JsonDeserializer.h
```diff
@@ -37,8 +37,8 @@ class DLL_LINKAGE JsonDeserializer: public JsonTreeSerializer<const JsonNode *>
 	void serializeInternal(int64_t & value) override;
 
 private:
-	void readLICPart(const JsonNode & part, const TDecoder & decoder, const bool val, std::vector<bool> & value);
-	void readLICPart(const JsonNode & part, const TDecoder & decoder, std::set<si32> & value);
+	static void readLICPart(const JsonNode & part, const TDecoder & decoder, const bool val, std::vector<bool> & value);
```
IvanSavenko, line 40: Same here. I think better to remove this "modernizing" rule entirely

lib/spells/CSpellHandler.cpp
```diff
@@ -84,54 +84,21 @@ static const ESpellSchool SCHOOL_ORDER[4] =
...
-	targetType(spells::AimType::NO_TARGET),
-	mechanics(),
-	adventureMechanics()
 {
 	levels.resize(GameConstants::SPELL_SCHOOL_LEVELS);
 }
 
-CSpell::~CSpell()
-{
+//This destructor should be placed here to avoid side effects
```
IvanSavenko, line 93: Same

lib/spells/CSpellHandler.h
```diff
@@ -191,17 +185,17 @@ class DLL_LINKAGE CSpell : public spells::Spell
...
+	si32 power = 0; //spell's power
 
 	std::map<TFaction, si32> probabilities; //% chance to gain for castles
 
-	bool combat; //is this spell combat (true) or adventure (false)
-	bool creatureAbility; //if true, only creatures can use this spell
-	si8 positiveness; //1 if spell is positive for influenced stacks, 0 if it is indifferent, -1 if it's negative
+	bool combat = false; //is this spell combat (true) or adventure (false)
+	bool creatureAbility = false; //if true, only creatures can use this spell
+	si8 positiveness = ESpellPositiveness::NEUTRAL; //1 if spell is positive for influenced stacks, 0 if it is indifferent, -1 if it's negative
```
IvanSavenko, line 198: I think better to initialize all of these in constructor.
1) This creates additional "noise" if you simply want to check public interface of a class
2) This also requires providing access to enum values in header files. Meaning, additional includes in header that can't be removed, slowing down compilation, especially - rebuilds where you have to change some common header

IMO in-place initialization is nice for small C-like structs, but for classes better to use constructor init

lib/spells/TargetCondition.cpp
```diff
@@ -277,16 +265,11 @@ class ReceptiveFeatureCondition : public TargetConditionItemBase
 
 private:
 	CSelector selector;
-	std::string cachingString;
+	std::string cachingString = "type_RECEPTIVE";
```
IvanSavenko, line 268: Please keep initializers in one place. It is weird to have selector initialized in constructor and cachingString in-place

lib/serializer/BinaryDeserializer.cpp
```diff
@@ -98,7 +97,7 @@ void CLoadFile::clear()
 
 void CLoadFile::checkMagicBytes(const std::string &text)
 {
-	std::string loaded = text;
+	const std::string& loaded = text;
```
IvanSavenko, line 100: Change of behavior - now `read()` will overwrite `text` since they both point to same variable.
Is this clang-tidy change or manual one?

IvanSavenko (commented):
On making methods static:
I think we should remove this rule. 
Yes, this means that function will get `this` as parameters, which, oh no, will use 8 bytes on stack and several nanoseconds of CPU time if compiler fails at optimizing.

But IMO static methods should be reserved to methods that *should* be called without an instance.

lib/serializer/BinaryDeserializer.cpp
rilian-la-te: Not all, but I will try with version 15 too.

lib/serializer/BinaryDeserializer.cpp
rilian-la-te: clang-tidy one.

lib/serializer/BinaryDeserializer.cpp
rilian-la-te: Side effects is always same - something stops compiling or linking, because destructor implicitly requires knowledge about something included only in *.cpp. I need to explain this in each destructor?

lib/serializer/CMemorySerializer.h
rilian-la-te: @IvanSavenko I can manually try put initializations into a constructor when constructor preserved, but in many cases moving init from a constructor to class can lead to removal of entire constructor.

lib/serializer/BinaryDeserializer.cpp
IvanSavenko: `//must be instantiated in .cpp file for access to complete types of all member fields`

IvanSavenko: Problem is "side effect" is too generic. Could be some weird optimization issue on some platform, could be even compiler bug.

lib/serializer/CMemorySerializer.h
IvanSavenko: Perhaps we can in general remove this rule and only use it manually for small structs.
For example, in this PR `ObjectPosInfo` is the only object where in-place initialization is clearly better.

I don't think that we have large number of "dump structs", with netpacks as possible exception. And for netpacks, we can try processing it with clang-tidy separately.

lib/serializer/BinaryDeserializer.cpp
IvanSavenko: If I am reading this correctly, looks to be present, including auto-fix. And available in older versions.
https://releases.llvm.org/12.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/cppcoreguidelines-pro-type-cstyle-cast.html

rilian-la-te: Yes, it is present. But this rule is present too: ```This rule bans (T)expression only when used to perform an unsafe cast.``` Maybe clang-tidy thinks than cast is safe here.

lib/spells/CSpellHandler.h
rilian-la-te: @IvanSavenko is constructor removed here? If this struct becomes trivial - then I think we should initialize all fields in class there. If not - we can again create default constructor with initialized fields.

lib/serializer/BinaryDeserializer.cpp
IvanSavenko: It might be safe here, but here it is definitely NOT safe:
https://github.com/vcmi/vcmi/pull/1561#discussion_r1100218176

Well, I guess we should look into that warning eventually...

rilian-la-te:
Where now-static town portal methods should be located? In CHeroInstance, or maybe in CGameState?

Because it does not needs logically any spell mechanics - it need only position and town list (and in second case - also Player Team and Player State).

IvanSavenko:
Another interesting option is to have Castle Gate execute TownPortalMechanics with custom parameters.

I'd say that what needs refactoring is town buildings in general. Town portal logic looks OK, problem is with town building. And Castle Gate is not the only such building.

Probably better to leave Castle Gate/ Town Portal as it for now, and change only after there is a clear idea as to how refactor town buildings support.

rilian-la-te:
@IvanSavenko I think than we should have some "Action" than can be applied both to adventure spells and town buildings. For example, TRADE action can be applied both to Marketplace and to hypothetical ALCHEMY spell, and to market adventure map object. Also TO_TOWN action with Town Portal and Castle Gate, Visions action to thieves and Vision spell, and so on. If bonuses can be actionable, than it is a good place to have it use bonuses instead.

lib/serializer/CMemorySerializer.h
rilian-la-te: done

lib/serializer/BinarySerializer.cpp
rilian-la-te: done

lib/serializer/BinaryDeserializer.cpp
rilian-la-te: reverted

lib/serializer/CTypeList.h
rilian-la-te: reverted

lib/spells/CSpellHandler.cpp
rilian-la-te: done

lib/spells/CSpellHandler.h
rilian-la-te: moved to constructor again

lib/spells/TargetCondition.cpp
rilian-la-te: all to fields done

lib/serializer/BinaryDeserializer.cpp
rilian-la-te: reverted

lib/serializer/JsonDeserializer.h
rilian-la-te: About LIC - it needed to move out of classes, because it is just duplicated as is.

rilian-la-te: Done as new interface.

IvanSavenko:
By the way, found `-Wold-style-cast` in clang-tidy: `google-readability-casting` that "suggests automated fixes in some cases". Might be worth trying it.
It will also affect casts between arithmetic types, but that's OK IMO.

lib/serializer/JsonUpdater.cpp
```diff
@@ -67,7 +67,7 @@ void JsonUpdater::serializeInternal(const std::string & fieldName, double & valu
 		value = data.Float();
 }
 
-void JsonUpdater::serializeInternal(const std::string & fieldName, si64 & value, const boost::optional<si64> &)
+void JsonUpdater::serializeInternal(const std::string & fieldName, si64 & value, const boost::optional<si64> & /*defaultValue*/)
```
IvanSavenko, line 70: I would rather un-comment this.
"unused variable" warning is disabled anyway

lib/spells/CSpellHandler.h
```diff
@@ -331,7 +325,7 @@ class DLL_LINKAGE CSpell : public spells::Spell
 	void setIsOffensive(const bool val);
 	void setIsRising(const bool val);
 
-	JsonNode convertTargetCondition(const BTVector & immunity, const BTVector & absImmunity, const BTVector & limit, const BTVector & absLimit) const;
+	static JsonNode convertTargetCondition(const BTVector & immunity, const BTVector & absImmunity, const BTVector & limit, const BTVector & absLimit);
```
IvanSavenko, line 328: Another conversion to static

lib/serializer/JsonSerializeFormat.cpp
```diff
@@ -110,15 +106,15 @@ size_t JsonArraySerializer::size() const
 }
 
 //JsonSerializeFormat::LIC
-JsonSerializeFormat::LIC::LIC(const std::vector<bool> & Standard, const TDecoder Decoder, const TEncoder Encoder):
+JsonSerializeFormat::LIC::LIC(const std::vector<bool> & Standard, const TDecoder& Decoder, const TEncoder& Encoder):
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko Will convert it, it was in first pass, but I disabled this check on second pass.

lib/serializer/JsonUpdater.cpp
rilian-la-te: done

lib/spells/CSpellHandler.h
rilian-la-te: Done

lib/spells/ISpellMechanics.cpp
```diff
@@ -141,16 +141,13 @@ class FallbackMechanicsFactory : public CustomMechanicsFactory
...
 	mode(mode_),
-	magicSkillLevel(),
-	effectPower(),
-	effectDuration(),
-	effectValue(),
 	smart(boost::logic::indeterminate),
 	massive(boost::logic::indeterminate)
 {
-	gameCb = IObjectInterface::cb; //FIXME: pass player callback (problem is that BattleAI do not have one)
+	//FIXME: pass player callback (problem is that BattleAI do not have one)
```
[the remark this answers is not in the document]
rilian-la-te: Done

lib/serializer/Connection.h
IvanSavenko: @rilian-la-te I think you missed this comment?

rilian-la-te: done. This was clang-tidy bogus initialization.

IvanSavenko (approved):
Looks OK.
Anyone else who want to check this before merge? @kambala-decapitator ?

lib/serializer/JsonSerializeFormat.cpp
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko added modernize-pass-by-value again, will try to build VCMI with fixes to pass-by-value.

IvanSavenko:
By the way, @rilian-la-te can you show your current clang-tidy settings that were used for this PR?

rilian-la-te:
@IvanSavenko @kambala-decapitator up-to-date clang-tidy settings is in https://github.com/rilian-la-te/vcmi/tree/clang-tidy-v2, and the rest of modernized lib code is there too.

IvanSavenko:
@rilian-la-te on clang-tiny config:
I compared yours to the one I setup for myself (but for IDE highlight and not for auto-conversion).
I think you can:
1) restore currently disabled `-modernize-avoid-bind`
2) disable `readability-convert-member-functions-to-static` if you haven't already.

There are few other checks which are not really useful for us, but they also don't offer auto-conversion so should be OK for your purposes

Also, FYI - clang-tidy config can be used with more readable format, without 500+ symbols in a single line, e.g:
```
Checks: >
  -*,
  modernize-*,
  -modernize-use-trailing-return-type,
  -modernize-return-braced-init-list

```

rilian-la-te:
@IvanSavenko modernize-avoid-bind breaking spells, I do not know how to rewrite it. About static - I remained it enabled and then edit by hand, because it can show us really corner cases like ILICReader)

IvanSavenko:
Understood. Well, I think both rules can be postponed for now then.
Besides, reviewing changes from a single rule might be better than review of multiple rules in one subsystem.

rilian-la-te:
@IvanSavenko generate changes by one rule per line is more difficult than generate changes at once and then post subsystem by subsystem.

rilian-la-te:
@kambala-decapitator waiting for your verdict and merge)

lib/serializer/ILICReader.cpp
```diff
@@ -0,0 +1,49 @@
...
+	{
+		const std::string & identifier = index.String();
+
+		const si32 rawId = decoder(identifier);
+		if(rawId != -1)
+			value.insert(rawId);
+	}
+}
+
+VCMI_LIB_NAMESPACE_END
```
kambala-decapitator, line 49: please insert empty line at EOF

lib/serializer/ILICReader.h
```diff
@@ -0,0 +1,23 @@
...
+#pragma once
+
+#include  "JsonTreeSerializer.h"
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+class DLL_LINKAGE ILICReader
+{
+protected:
+	void readLICPart(const JsonNode & part, const JsonSerializeFormat::TDecoder & decoder, const bool val, std::vector<bool> & value) const;
```
kambala-decapitator, line 19: why `const bool` param?

lib/serializer/CTypeList.h
```diff
@@ -109,12 +109,7 @@ class DLL_LINKAGE CTypeList: public boost::noncopyable
 
 		return ptr;
 	}
-	CTypeList &operator=(CTypeList &)
-	{
-		// As above.
```
kambala-decapitator, line 114: which "above" does it refer to?

lib/serializer/CSerializer.h
```diff
@@ -80,10 +80,11 @@ class DLL_LINKAGE CSerializer
...
 public:
-	bool smartVectorMembersSerialization;
-	bool sendStackInstanceByIds;
+	bool smartVectorMembersSerialization = false;
+	bool sendStackInstanceByIds = false;
 
-	CSerializer();
+	///This constructor is required for linking with server
+	CSerializer() = default;
 	~CSerializer();
```
kambala-decapitator, line 88: so what's our stance on declaring default ctor/dtor explicitly? In the PR I see a mix of approaches.

lib/serializer/CMemorySerializer.cpp
```diff
@@ -32,9 +32,8 @@ int CMemorySerializer::write(const void * data, unsigned size)
 	return size;
 }
 
-CMemorySerializer::CMemorySerializer(): iser(this), oser(this)
+CMemorySerializer::CMemorySerializer(): iser(this), oser(this), readPos(0)
```
kambala-decapitator, line 35: `readPos` value could be moved to declaration I guess

lib/serializer/JsonDeserializer.h
```diff
@@ -10,10 +10,11 @@
 #pragma once
 
 #include  "JsonTreeSerializer.h"
+#include  "ILICReader.h"
```
kambala-decapitator, line 13: please remove double space

lib/serializer/JsonUpdater.h
```diff
@@ -10,12 +10,13 @@
 #pragma once
 
 #include  "JsonTreeSerializer.h"
+#include  "ILICReader.h"
```
kambala-decapitator, line 13: please remove double space

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -448,16 +448,16 @@ void BattleSpellMechanics::doRemoveEffects(ServerCallback * server, const std::v
...
 	SetStackEffect sse;
 
-	for(auto unit : targets)
+	for(const auto *unit : targets)
 	{
 		std::vector<Bonus> buffer;
 		auto bl = unit->getBonuses(selector);
 
-		for(auto item : *bl)
+		for(const auto& item : *bl)
```
kambala-decapitator, line 456: that's a very strange mix of spacing... Any chance to unify? Per current code style `&` and `*` should actually be surrounded by spaces.

kambala-decapitator (changes requested):

lib/serializer/ILICReader.h
rilian-la-te: @kambala-decapitator because it was in both JsonUpdater.cpp + JsonSerializer.cpp. I can remove const if desired.

lib/serializer/CSerializer.h
rilian-la-te: @kambala-decapitator we cannot skip declaration here, because then it will fail to link.

lib/serializer/CMemorySerializer.cpp
rilian-la-te: It can, but as discussed in Slack, @IvanSavenko tells me than he prefers this approach - if we have a non-default constructor, initalize all fields in constructor, not in declaration.

What is better?

lib/serializer/ILICReader.cpp
rilian-la-te: done

lib/serializer/JsonUpdater.h
rilian-la-te: done

lib/serializer/CTypeList.h
rilian-la-te: It seems to be lost in ages. But this operator will always assert, so, I think deleting it instead is better, because it just not compile if requested instead of runtime error.

lib/serializer/ILICReader.h
rilian-la-te: @kambala-decapitator `const bool` parameter changed to `bool`. But it is still `const bool` semantically, because it should not be changed.

lib/spells/BattleSpellMechanics.cpp
rilian-la-te: passed via clang-format all.

lib/serializer/JsonDeserializer.h
rilian-la-te: done

lib/serializer/CTypeList.h
kambala-decapitator: thanks! deleting is fine, I just couldn't understand what this comment means.

lib/serializer/CSerializer.h
kambala-decapitator: I don't think I understand... If default constructor is removed, compiler would synthesise one automatically. How could that lead to a link error? Or is there another ctor in the class?

lib/serializer/CMemorySerializer.cpp
kambala-decapitator: aha, let's keep then

lib/serializer/ILICReader.h
kambala-decapitator: at the calling site it wouldn't change anyway. tbh I've never seen declaring POD type params as `const`.

lib/serializer/CSerializer.h
IvanSavenko: I think we should drop explicitly defaulted constructor/destructor if possible.

In this particular case it is probably not necessary?
Generally, in order for compiler to generate destructor compiler needs access to all class members (NOT just forward-declaration).
Because of this, if (for example) your class have shared_ptr<XXX> as member, XXX can be forward declared in header. However to emit destructor, compiler will need access to actual declaration of XXX. Meaning, destructor for such class must be defined in a place that has XXX declaration visible. Normally, cpp file is such place.

I think same applies for constructor, but never encounter this myself before.

Also, I won't be surprised if there are few other cases related to dll linkage.

lib/serializer/CMemorySerializer.cpp
IvanSavenko: I would rather keep it here. 
Not 100% sure what solution in regards to in-place initialization is better, so I think for now we should avoid too many changes. Let's do parts on which everyone agrees. That's more than enough things to do anyway.

lib/serializer/CSerializer.h
rilian-la-te: @kambala-decapitator @IvanSavenko I am rechecking this now, if link failure will not remain after clang and gcc upgrade, there can be a compiler bug or my build bug (somewhere I can forgot to do clean-build).

rilian-la-te: After upgrade it seems to linking normally without a constructor after clean-build.
If CI passes - I think constructor here can safely be removed. Changes pushed.

rilian-la-te: CI passed, so, constructor removal is fine.

kambala-decapitator: the PR still has a mix of approaches. Can default ctors/dtors be removed from everywhere?

kambala-decapitator:
seems fine now except the last point about default ctors/dtors

lib/serializer/CSerializer.h
rilian-la-te: @kambala-decapitator dtors - certainly not, there is a places when dtor should be explicitly set to default in *.cpp file.

rilian-la-te: And also - when we need to annotate than destructor is virtual. This is done non-consistent now, but should be more consistent. 

About ctors - I thought I removed all defaulted ctors.

IvanSavenko: Yeah, looks to be mostly consistent now.
Searching for word 'default' in PR I see following scenario:
1) Constructor defined in-place as default: required, since class has 2nd constructor, so compiler will normally delete default (0-params) constructor.
2) Destructor defined as default in header: required, since this is virtual destructor of a base class
3) Destructor defined as default in source file: required for compiler to generate destructors for members.

I see only one inconsistency: sometimes destructor is declared in header and sometimes in source. (case 2 & 3 from above). We *can* move all of them in .cpp, but not sure if this is better style-wise. Any opinions, @kambala-decapitator ?

Personally fine either way IMO.

IvanSavenko: > And also - when we need to annotate than destructor is virtual

Well, those can be marked as `override`, (except for base class obviously). Already saw that in other PR from you. Looks OK I think.

rilian-la-te: @IvanSavenko I mean your case 2) If I have not removed virtual or override destructor in derived class - it is my fault.

kambala-decapitator: declaring virtual dtor is clear.

> dtors - certainly not, there is a places when dtor should be explicitly set to default in *.cpp file.

I don't think I get which case is that (also case number 3)... Could you show an example? Here's an example with forward declaration which I thought was related to this, but it works fine either if you place default dtor in header or remove the dtor from code, in both cases the `die` message is printed:

```
// 1.h
struct A {
	~A();
};
```

```
// 1.cpp
#include "1.h"
#include <iostream>
A::~A() { std::cout << "die\n"; }
```

```
// 2.h
struct A;
struct B {
	A& a;
	
	B(A& a_);
	~B() = default;
};
```

```
#include "1.h"
#include "2.h"
B::B(A& a_) : a{a_} {}

int main()
{
	A a;
	B b{a};
	return 0;
}
```

kambala-decapitator: even if I move `main()` to another file and remove `#include "1.h"` from `2.cpp` there's still no compile/link error

rilian-la-te: @kambala-decapitator to get a link error, you need to slightly adjust an example in file 2.h

```c++
// 2.h
struct A;
struct B {
	std::unique_ptr<A> a;
	
	B(A& a_);
	~B() = default;
};
```

Then it will be link error, because to destroy unique_ptr, you need full signature of A.

kambala-decapitator: now it's fully clear, thank you very much!

kambala-decapitator (approved):

PR #1562, opened 2023-02-08
vcmi: modernize lib/NetPacks.h and lib/NetPacksBase.h

rilian-la-te:
@nullkiller please, consider

@IvanSavenko another huge serialize for review

### DIFF
--- a/lib/NetPacks.h
+++ b/lib/NetPacks.h
@@ -39,10 +39,6 @@ class IBattleState;
 struct Query : public CPackForClient
 {
 	QueryID queryID; // equals to -1 if it is not an actual query (and should not be answered)
-
-	Query()
-	{
-	}
 };
 
 struct StackLocation
@@ -50,11 +46,10 @@ struct StackLocation
 	ConstTransitivePtr<CArmedInstance> army;
 	SlotID slot;
 
-	StackLocation()
-	{}
-	StackLocation(const CArmedInstance *Army, SlotID Slot):
-		army(const_cast<CArmedInstance*>(Army)), //we are allowed here to const cast -> change will go through one of our packages... do not abuse!
-		slot(Slot)
+	StackLocation() = default;
+	StackLocation(const CArmedInstance * Army, const SlotID & Slot)
+		: army(const_cast<CArmedInstance *>(Army))  //we are allowed here to const cast -> change will go through one of our packages... do not abuse!
+		, slot(Slot)
 	{
 	}
 
@@ -71,17 +66,16 @@ struct StackLocation
 
 struct PackageApplied : public CPackForClient
 {
-	PackageApplied()
-		: result(0), packType(0),requestID(0)
-	{}
+	PackageApplied() = default;
 	PackageApplied(ui8 Result)
-		: result(Result), packType(0), requestID(0)
-	{}
+		: result(Result)
+	{
+	}
 	void applyCl(CClient *cl);
 
-	ui8 result; //0 - something went wrong, request hasn't been realized; 1 - OK
-	ui32 packType; //type id of applied package
-	ui32 requestID; //an ID given by client to the request that was applied
+	ui8 result = 0; //0 - something went wrong, request hasn't been realized; 1 - OK
+	ui32 packType = 0; //type id of applied package
+	ui32 requestID = 0; //an ID given by client to the request that was applied
 	PlayerColor player;
 
 	template <typename Handler> void serialize(Handler &h, const int version)
@@ -95,8 +89,11 @@ struct PackageApplied : public CPackForClient
 
 struct SystemMessage : public CPackForClient
 {
-	SystemMessage(const std::string & Text) : text(Text){}
-	SystemMessage(){}
+	SystemMessage(std::string Text)
+		: text(std::move(Text))
+	{
+	}
+	SystemMessage() = default;
 	void applyCl(CClient *cl);
 
 	std::string text;
@@ -109,14 +106,13 @@ struct SystemMessage : public CPackForClient
 
 struct PlayerBlocked : public CPackForClient
 {
-	PlayerBlocked() : reason(UPCOMING_BATTLE), startOrEnd(BLOCKADE_STARTED) {}
 	void applyCl(CClient *cl);
 
 	enum EReason { UPCOMING_BATTLE, ONGOING_MOVEMENT };
 	enum EMode { BLOCKADE_STARTED, BLOCKADE_ENDED };
 
-	EReason reason;
-	EMode startOrEnd;
+	EReason reason = UPCOMING_BATTLE;
+	EMode startOrEnd = BLOCKADE_STARTED;
 	PlayerColor player;
 
 	template <typename Handler> void serialize(Handler &h, const int version)
@@ -129,12 +125,11 @@ struct PlayerBlocked : public CPackForClient
 
 struct PlayerCheated : public CPackForClient
 {
-	PlayerCheated() : losingCheatCode(false), winningCheatCode(false) {}
-	DLL_LINKAGE void applyGs(CGameState *gs);
+	DLL_LINKAGE void applyGs(CGameState * gs) const;
 
 	PlayerColor player;
-	bool losingCheatCode;
-	bool winningCheatCode;
+	bool losingCheatCode = false;
+	bool winningCheatCode = false;
 
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
@@ -146,9 +141,8 @@ struct PlayerCheated : public CPackForClient
 
 struct YourTurn : public CPackForClient
 {
-	YourTurn(){}
 	void applyCl(CClient *cl);
-	DLL_LINKAGE void applyGs(CGameState *gs);
+	DLL_LINKAGE void applyGs(CGameState * gs) const;
 
 	PlayerColor player;
 	boost::optional<ui8> daysWithoutCastle;
@@ -164,7 +158,6 @@ struct EntitiesChanged: public CPackForClient
 {
 	std::vector<EntityChanges> changes;
 
-	EntitiesChanged(){};
 	void applyCl(CClient * cl);
 	DLL_LINKAGE void applyGs(CGameState * gs);
 
@@ -176,11 +169,10 @@ struct EntitiesChanged: public CPackForClient
 
 struct SetResources : public CPackForClient
 {
-	SetResources():abs(true){};
 	void applyCl(CClient *cl);
-	DLL_LINKAGE void applyGs(CGameState *gs);
+	DLL_LINKAGE void applyGs(CGameState * gs) const;
 
-	bool abs; //false - changes by value; 1 - sets to value
+	bool abs = true; //false - changes by value; 1 - sets to value
 	PlayerColor player;
 	TResources res; //res[resid] => res amount
 
@@ -194,16 +186,13 @@ struct SetResources : public CPackForClient
 
 struct SetPrimSkill : public CPackForClient
 {
-	SetPrimSkill()
-		: abs(0), which(PrimarySkill::ATTACK), val(0)
-	{}
 	void applyCl(CClient *cl);
-	DLL_LINKAGE void applyGs(CGameState *gs);
+	DLL_LINKAGE void applyGs(CGameState * gs) const;
 
-	ui8 abs; //0 - changes by value; 1 - sets to value
+	ui8 abs = 0; //0 - changes by value; 1 - sets to value
 	ObjectInstanceID id;
-	PrimarySkill::PrimarySkill which;
-	si64 val;
+	PrimarySkill::PrimarySkill which = PrimarySkill::ATTACK;
+	si64 val = 0;
 
 	template <typename Handler> void serialize(Handler &h, const int version)
 	{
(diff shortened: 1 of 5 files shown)


lib/NetPacks.h
```diff
@@ -1210,17 +1137,15 @@ struct ChangeObjectVisitors : public CPackForClient
...
-	DLL_LINKAGE void applyGs(CGameState *gs);
+	DLL_LINKAGE void applyGs(CGameState *gs) const;
 
-	ChangeObjectVisitors()
-		: mode(VISITOR_CLEAR)
-	{}
+	ChangeObjectVisitors() = default;
 
-	ChangeObjectVisitors(ui32 mode, ObjectInstanceID object, ObjectInstanceID heroID = ObjectInstanceID(-1)):
+	ChangeObjectVisitors(ui32 mode, const ObjectInstanceID& object, const ObjectInstanceID& heroID = ObjectInstanceID(-1)):
```
IvanSavenko, line 1148: Formatting

IvanSavenko (commented):
Looks OK other than constructor formatting in A LOT of netpacks (too many to list each of them). Perhaps pass through clang-format?

Things to consider:
- check thorough potentially uninitialized fields. Perhaps even initialize all int's and all enum's in-place. clang-tidy likely has something for that. Might be detected later, e.g. when checking code that uses NetPack's.
- postponing till after nullkiller's PR to reduce conflicts

lib/NetPacks.h
rilian-la-te: @IvanSavenko passed through clang-format.

lib/NetPacks.h
```diff
@@ -40,20 +40,18 @@ struct Query : public CPackForClient
...
 	SlotID slot;
 
-	StackLocation()
-	{}
-	StackLocation(const CArmedInstance *Army, SlotID Slot):
-		army(const_cast<CArmedInstance*>(Army)), //we are allowed here to const cast -> change will go through one of our packages... do not abuse!
+	StackLocation() = default;
+	StackLocation(const CArmedInstance * Army, const SlotID & Slot)
+		: army(const_cast<CArmedInstance *>(Army))
+		, //we are allowed here to const cast -> change will go through one of our packages... do not abuse!
```
IvanSavenko, line 54: What has happened here with formatting?

lib/NetPacks.h
```diff
@@ -571,8 +533,9 @@ struct ChangeFormation : public CPackForClient
 
 struct RemoveObject : public CPackForClient
 {
-	RemoveObject(){}
-	RemoveObject(ObjectInstanceID ID){id = ID;};
+	RemoveObject() = default;
+	RemoveObject(const ObjectInstanceID & ID)
+		: id(ID){};
```
IvanSavenko, line 538: Another weird formatting. `{}` should be on separate line and without semicolon at the end.
And I see similar form in code below as well.

lib/NetPacks.h
```diff
@@ -729,13 +689,12 @@ struct GiveHero : public CPackForClient
...
 struct OpenWindow : public CPackForClient
 {
-	OpenWindow():id1(-1),id2(-1){}
 	void applyCl(CClient *cl);
 
 	enum EWindow {EXCHANGE_WINDOW, RECRUITMENT_FIRST, RECRUITMENT_ALL, SHIPYARD_WINDOW, THIEVES_GUILD,
 	              UNIVERSITY_WINDOW, HILL_FORT_WINDOW, MARKET_WINDOW, PUZZLE_MAP, TAVERN_WINDOW};
 	ui8 window;
-	si32 id1, id2;
+	si32 id1 = -1, id2 = -1;
```
IvanSavenko, line 697: Kind of surprised - I saw readability check in clang-tidy for *local* variables, that removes multiple declarations in a single line. But nothing for same case in classes?

lib/NetPacks.h
```diff
@@ -767,11 +725,11 @@ struct NewObject : public CPackForClient
 
 struct SetAvailableArtifacts : public CPackForClient
 {
-	SetAvailableArtifacts():id(0){}
 	void applyCl(CClient *cl);
-	DLL_LINKAGE void applyGs(CGameState *gs);
+	DLL_LINKAGE void applyGs(CGameState * gs) const;
 
-	si32 id; //two variants: id < 0: set artifact pool for Artifact Merchants in towns; id >= 0: set pool for adv. map Black Market (id is the id of Black Market instance then)
+	si32 id =
```
IvanSavenko, line 731: What has happened here?

IvanSavenko (commented):
Probably better, just a bit?
For some reason clang-format failed to properly format constructors - search for `{};` in resulting file. 
Old code has semicolons at the end of constructors, which is unnecessary but not illegal. And it looks like clang-format failed to process such code for some reason.

As I said in clang-format PR, this tool can be used as a baseline for formatting, but still needs manual review of its output.

lib/NetPacksLib.cpp
```diff
@@ -411,7 +407,7 @@ DLL_LINKAGE void RemoveObject::applyGs(CGameState *gs)
 
 	if(obj->ID == Obj::HERO) //remove beaten hero
 	{
-		CGHeroInstance * beatenHero = static_cast<CGHeroInstance*>(obj);
+		auto * beatenHero = dynamic_cast<CGHeroInstance *>(obj);
```
nullkiller, line 410: Why should we replace static_cast with dynamic one? This is questionable. We are not prepared to nullptr anyway?

lib/NetPacksLib.cpp
```diff
@@ -553,15 +549,15 @@ void TryMoveHero::applyGs(CGameState *gs)
 	{
 		const TerrainTile &tt = gs->map->getTile(h->convertToVisitablePos(end));
 		assert(tt.visitableObjects.size() >= 1  &&  tt.visitableObjects.back()->ID == Obj::BOAT); //the only visitable object at destination is Boat
-		CGBoat *boat = static_cast<CGBoat*>(tt.visitableObjects.back());
+		auto * boat = dynamic_cast<CGBoat *>(tt.visitableObjects.back());
```
nullkiller, line 552: Here also.

rilian-la-te: Answered earlier. You think than is should remain static_cast with possibility of failure?

lib/NetPacks.h
nullkiller: Semicolon was there initially

lib/NetPacksLib.cpp
IvanSavenko: I do agree with change.
Yes, now code *may* result in `beatenHero == nullptr`, but old code was no safe anyway.
We might drop assert or even abort if this happens, but this is still better than old code.

Using static_cast for upcasting is NOT safe, and will only result in crash at some place later. So better to detect this immediately.
```suggestion
		auto * beatenHero = dynamic_cast<CGHeroInstance *>(obj);
		assert(beatenHero != nullptr);
```

lib/NetPacks.h
rilian-la-te: formatting issues resolved AFAIK.

lib/NetPacks.h
rilian-la-te: formatting issues resolved AFAIK.

lib/NetPacks.h
rilian-la-te: formatting issues resolved AFAIK.

lib/NetPacks.h
rilian-la-te: Formatting issues resolved AFAIK

nullkiller (commented):
I have nothing against it except dynamic_cast. So if @IvanSavenko is ok I have no strong objection here. I am not C++ dev, I was not even aware what is the exact difference between static and dynamic cast for quite a long time

lib/NetPacksLib.cpp
nullkiller: This is more about intention. If you use static_cast you are sure what type is while if you use dynamic - you are not sure. As I understand c++ has no way to give some meaningful error in case of invalid static_cast so option with dynamic cast and assert may look better

IvanSavenko:
Yeah, I do think that we should *never* use static_cast for upcasting. Instead such code should use dynamic_cast + some error checking, at the very least - basic assert(). Or refactored to avoid unconditional upcasting, but this is definitely out of scope for auto-upgrade PR

lib/NetPacks.h
IvanSavenko: Yes, semicolon was here before, but it is unnecessary & apparently messed up formatting for some reason. I thought tools like clang-format should remove it.

Anyway, looks to be OK now.

lib/NetPacks.h
```diff
@@ -2403,8 +2318,9 @@ struct QueryReply : public CPackForServer
 
 struct MakeAction : public CPackForServer
 {
-	MakeAction(){};
-	MakeAction(const BattleAction &BA):ba(BA){};
+	MakeAction() = default;
+	MakeAction(BattleAction BA)
+		: ba(std::move(BA)){}
```
IvanSavenko, line 2323: Missed this one I guess?

lib/NetPacks.h
```diff
@@ -2479,8 +2397,9 @@ struct SaveGame : public CPackForServer
 // TODO: Eventually we should re-merge both SaveGame and PlayerMessage
 struct SaveGameClient : public CPackForClient
 {
-	SaveGameClient(){};
-	SaveGameClient(const std::string &Fname) :fname(Fname){};
+	SaveGameClient() = default;
+	SaveGameClient(std::string Fname)
+		: fname(std::move(Fname)){}
```
IvanSavenko, line 2402: And here

lib/NetPacksLib.cpp
```diff
@@ -750,9 +746,9 @@ DLL_LINKAGE void NewObject::applyGs(CGameState *gs)
 		o = new CGCreature();
 		{
 			//CStackInstance hlp;
-			CGCreature *cre = static_cast<CGCreature*>(o);
+			auto * cre = dynamic_cast<CGCreature *>(o);
```
IvanSavenko, line 749: Please add assert here and in other instances of static_cast -> dynamic_cast change:
```suggestion
			auto * cre = dynamic_cast<CGCreature *>(o);
			assert(cre);
```

lib/NetPacksLib.cpp
```diff
@@ -1331,18 +1327,18 @@ DLL_LINKAGE void SetObjectProperty::applyGs(CGameState *gs)
...
 	}
 
-	CArmedInstance *cai = dynamic_cast<CArmedInstance *>(obj);
+	auto * cai = dynamic_cast<CArmedInstance *>(obj);
 	if(what == ObjProperty::OWNER && cai)
 	{
 		if(obj->ID == Obj::TOWN)
 		{
-			CGTownInstance *t = static_cast<CGTownInstance*>(obj);
+			auto * t = dynamic_cast<CGTownInstance *>(obj);
```
IvanSavenko, line 1335:
```suggestion
			auto * t = dynamic_cast<CGTownInstance *>(obj);
			assert(t);
```

lib/NetPacksLib.cpp
IvanSavenko:
```suggestion
		auto * boat = dynamic_cast<CGBoat *>(tt.visitableObjects.back());
		assert(boat);
```

IvanSavenko:
Also, in my own clang-tidy run on develop branch I see a bunch of following warnings from Netpacks:
`1 uninitialized field at the end of the constructor call [clang-analyzer-optin.cplusplus.UninitializedObject]`
Just FYI, can be easily fixed later.

lib/NetPacksLib.cpp
rilian-la-te: @IvanSavenko done all asserts (please, remove suggestions, I squashed a commits)

lib/NetPacksLib.cpp
rilian-la-te: done

lib/NetPacks.h
rilian-la-te: done

lib/NetPacks.h
rilian-la-te: done

lib/NetPacksLib.cpp
rilian-la-te: done assert

lib/NetPacksBase.h
```diff
@@ -34,8 +34,8 @@ struct DLL_LINKAGE CPack
 {
 	std::shared_ptr<CConnection> c; // Pointer to connection that pack received from
 
-	CPack() : c(nullptr) {};
-	virtual ~CPack() {};
+	CPack() : c(nullptr) {}
```
IvanSavenko, line 37: Non-default constructor is unnecessary since shared_ptr is always initialized to nullptr

IvanSavenko (approved):

lib/NetPacksBase.h
rilian-la-te: Constructor is required here, but I can change it to default if desired.

Will post next series of modernization this evening)

rilian-la-te: done

nullkiller:
I feel like there will be conflicts with my branch but they may be ease to solve. At the same time I may spend more time on mine as tests showed there are other problems

kambala-decapitator (approved):

rilian-la-te:
@nullkiller do I need to wait for you or it can be merged already?

nullkiller:
@rilian-la-te, I am ready to merge mine. Do not know what way is better. My branch is mostly manual and in case of severe conflict it may be hard to merge it. So we touched more or less different parts.

PR #1567, opened 2023-02-11
Modernize maps

rilian-la-te:
Next series. Modernize lib/rmg, lib/mapping and lib/mapObjects.

### DIFF
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -410,17 +410,17 @@ CArtifact * CArtHandler::loadFromJson(const std::string & scope, const JsonNode 
 	return art;
 }
 
-ArtifactPosition CArtHandler::stringToSlot(std::string slotName)
+ArtifactPosition::ArtifactPosition(std::string slotName):
+	num(ArtifactPosition::PRE_FIRST)
 {
 #define ART_POS(x) { #x, ArtifactPosition::x },
 	static const std::map<std::string, ArtifactPosition> artifactPositionMap = { ART_POS_LIST };
 #undef ART_POS
 	auto it = artifactPositionMap.find (slotName);
 	if (it != artifactPositionMap.end())
-		return it->second;
+		num = it->second;
 
 	logMod->warn("Warning! Artifact slot %s not recognized!", slotName);
-	return ArtifactPosition::PRE_FIRST;
 }
 
 void CArtHandler::addSlot(CArtifact * art, const std::string & slotID)
@@ -445,7 +445,7 @@ void CArtHandler::addSlot(CArtifact * art, const std::string & slotID)
 	}
 	else
 	{
-		auto slot = stringToSlot(slotID);
+		auto slot = ArtifactPosition(slotID);
 		if (slot != ArtifactPosition::PRE_FIRST)
 			art->possibleSlots[ArtBearer::HERO].push_back (slot);
 	}
--- a/lib/CArtHandler.h
+++ b/lib/CArtHandler.h
@@ -249,8 +249,7 @@ public:
 
 	boost::optional<std::vector<CArtifact*>&> listFromClass(CArtifact::EartClass artifactClass);
 
-	ArtifactPosition stringToSlot(std::string slotName);
-	CArtifact::EartClass stringToClass(std::string className);
+	static CArtifact::EartClass stringToClass(std::string className); //TODO: rework EartClass to make this a constructor
 
 	/// Gets a artifact ID randomly and removes the selected artifact from this handler.
 	ArtifactID pickRandomArtifact(CRandomGenerator & rand, int flags);
--- a/lib/mapObjects/CArmedInstance.cpp
+++ b/lib/mapObjects/CArmedInstance.cpp
@@ -35,7 +35,6 @@ void CArmedInstance::randomizeArmy(int type)
 		assert(elem.second->valid(false));
 		assert(elem.second->armyObj == this);
 	}
-	return;
 }
 
 // Take Angelic Alliance troop-mixing freedom of non-evil units into account.
@@ -46,10 +45,11 @@ CArmedInstance::CArmedInstance()
 {
 }
 
-CArmedInstance::CArmedInstance(bool isHypotetic)
-	:CBonusSystemNode(isHypotetic), nonEvilAlignmentMix(this, nonEvilAlignmentMixSelector)
+CArmedInstance::CArmedInstance(bool isHypotetic):
+	CBonusSystemNode(isHypotetic),
+	nonEvilAlignmentMix(this, nonEvilAlignmentMixSelector),
+	battle(nullptr)
 {
-	battle = nullptr;
 }
 
 void CArmedInstance::updateMoraleBonusFromArmy()
@@ -71,7 +71,7 @@ void CArmedInstance::updateMoraleBonusFromArmy()
 	const std::string undeadCacheKey = "type_UNDEAD";
 	static const CSelector undeadSelector = Selector::type()(Bonus::UNDEAD);
 
-	for(auto slot : Slots())
+	for(const auto & slot : Slots())
 	{
 		const CStackInstance * inst = slot.second;
 		const CCreature * creature  = VLC->creh->objects[inst->getCreatureID()];
@@ -110,7 +110,7 @@ void CArmedInstance::updateMoraleBonusFromArmy()
 	}
 	else if (!factions.empty()) // no bonus from empty garrison
 	{
-	 	b->val = 2 - (si32)factionsInArmy;
+		b->val = 2 - static_cast<si32>(factionsInArmy);
 		description = boost::str(boost::format(VLC->generaltexth->arraytxt[114]) % factionsInArmy % b->val); //Troops of %d alignments %d
 		description = b->description.substr(0, description.size()-2);//trim value
 	}
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -38,7 +38,7 @@ VCMI_LIB_NAMESPACE_BEGIN
 
 
 ///helpers
-static void showInfoDialog(const PlayerColor playerID, const ui32 txtID, const ui16 soundID = 0)
+static void showInfoDialog(const PlayerColor & playerID, const ui32 txtID, const ui16 soundID = 0)
 {
 	InfoWindow iw;
 	iw.soundID = soundID;
@@ -56,7 +56,7 @@ static void showInfoDialog(const CGHeroInstance* h, const ui32 txtID, const ui16
 static int lowestSpeed(const CGHeroInstance * chi)
 {
 	static const CSelector selectorSTACKS_SPEED = Selector::type()(Bonus::STACKS_SPEED);
-	static const std::string keySTACKS_SPEED = "type_" + std::to_string((si32)Bonus::STACKS_SPEED);
+	static const std::string keySTACKS_SPEED = "type_" + std::to_string(static_cast<si32>(Bonus::STACKS_SPEED));
 
 	if(!chi->stacksCount())
 	{
@@ -97,7 +97,7 @@ ui32 CGHeroInstance::getTileCost(const TerrainTile & dest, const TerrainTile & f
 		if(ret < GameConstants::BASE_MOVEMENT_COST)
 			ret = GameConstants::BASE_MOVEMENT_COST;
 	}
-	return (ui32)ret;
+	return static_cast<ui32>(ret);
 }
 
 TerrainId CGHeroInstance::getNativeTerrain() const
@@ -110,7 +110,7 @@ TerrainId CGHeroInstance::getNativeTerrain() const
 
 	TerrainId nativeTerrain = ETerrainId::ANY_TERRAIN;
 
-	for(auto stack : stacks)
+	for(const auto & stack : stacks)
 	{
 		TerrainId stackNativeTerrain = stack.second->type->getNativeTerrain(); //consider terrain bonuses e.g. Lodestar.
 
@@ -130,19 +130,19 @@ BattleField CGHeroInstance::getBattlefield() const
 	return BattleField::NONE;
 }
 
-ui8 CGHeroInstance::getSecSkillLevel(SecondarySkill skill) const
+ui8 CGHeroInstance::getSecSkillLevel(const SecondarySkill & skill) const
 {
-	for(auto & elem : secSkills)
+	for(const auto & elem : secSkills)
 		if(elem.first == skill)
 			return elem.second;
 	return 0;
 }
 
-void CGHeroInstance::setSecSkillLevel(SecondarySkill which, int val, bool abs)
+void CGHeroInstance::setSecSkillLevel(const SecondarySkill & which, int val, bool abs)
 {
 	if(getSecSkillLevel(which) == 0)
 	{
-		secSkills.push_back(std::pair<SecondarySkill,ui8>(which, val));
+		secSkills.emplace_back(which, val);
 		updateSkillBonus(which, val);
 	}
 	else
@@ -182,7 +182,7 @@ bool CGHeroInstance::canLearnSkill() const
 	return secSkills.size() < GameConstants::SKILL_PER_HERO;
 }
 
-bool CGHeroInstance::canLearnSkill(SecondarySkill which) const
+bool CGHeroInstance::canLearnSkill(const SecondarySkill & which) const
 {
 	if ( !canLearnSkill())
 		return false;
@@ -207,12 +207,12 @@ int CGHeroInstance::maxMovePoints(bool onLand) const
 
 int CGHeroInstance::maxMovePointsCached(bool onLand, const TurnInfo * ti) const
 {
-	int base;
+	int base = 0;
 
 	if(onLand)
 	{
 		// used function is f(x) = 66.6x + 1300, rounded to second digit, where x is lowest speed in army
-		static const int baseSpeed = 1300; // base speed from creature with 0 speed
+		static constexpr int baseSpeed = 1300; // base speed from creature with 0 speed
 
 		int armySpeed = lowestSpeed(this) * 20 / 3;
 
@@ -230,28 +230,25 @@ int CGHeroInstance::maxMovePointsCached(bool onLand, const TurnInfo * ti) const
 	const int subtype = onLand ? SecondarySkill::LOGISTICS : SecondarySkill::NAVIGATION;
 	const double modifier = ti->valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, subtype) / 100.0;
 
-	return int(base * (1 + modifier)) + bonus;
+	return static_cast<int>(base * (1 + modifier)) + bonus;
 }
 
-CGHeroInstance::CGHeroInstance()
- : IBoatGenerator(this)
+CGHeroInstance::CGHeroInstance():
+	IBoatGenerator(this),
+	tacticFormationEnabled(false),
+	inTownGarrison(false),
+	isStanding(true),
+	moveDir(4),
+	mana(UNINITIALIZED_MANA),
+	movement(UNINITIALIZED_MOVEMENT),
+	portrait(UNINITIALIZED_PORTRAIT),
+	level(1),
+	exp(std::numeric_limits<TExpType>::max()),
+	sex(std::numeric_limits<ui8>::max())
 {
 	setNodeType(HERO);
 	ID = Obj::HERO;
-	tacticFormationEnabled = inTownGarrison = false;
-	mana = UNINITIALIZED_MANA;
-	movement = UNINITIALIZED_MOVEMENT;
-	portrait = UNINITIALIZED_PORTRAIT;
-	isStanding = true;
-	moveDir = 4;
-	level = 1;
-	exp = 0xffffffff;
-	visitedTown = nullptr;
-	type = nullptr;
-	boat = nullptr;
-	commander = nullptr;
-	sex = 0xff;
-	secSkills.push_back(std::make_pair(SecondarySkill::DEFAULT, -1));
+	secSkills.emplace_back(SecondarySkill::DEFAULT, -1);
 }
 
 PlayerColor CGHeroInstance::getOwner() const
@@ -259,7 +256,7 @@ PlayerColor CGHeroInstance::getOwner() const
 	return tempOwner;
 }
 
-void CGHeroInstance::initHero(CRandomGenerator & rand, HeroTypeID SUBID)
+void CGHeroInstance::initHero(CRandomGenerator & rand, const HeroTypeID & SUBID)
 {
 	subID = SUBID.getNum();
 	initHero(rand);
@@ -286,7 +283,7 @@ void CGHeroInstance::initHero(CRandomGenerator & rand)
 
 	if(!vstd::contains(spells, SpellID::PRESET)) //hero starts with a spell
 	{
-		for(auto spellID : type->spells)
+		for(const auto & spellID : type->spells)
 			spells.insert(spellID);
 	}
 	else //remove placeholder
@@ -484,11 +481,11 @@ ui8 CGHeroInstance::maxlevelsToWisdom() const
 	return type->heroClass->isMagicHero() ? 3 : 6;
 }
 
-CGHeroInstance::SecondarySkillsInfo::SecondarySkillsInfo()
+CGHeroInstance::SecondarySkillsInfo::SecondarySkillsInfo():
+	magicSchoolCounter(1),
+	wisdomCounter(1)
 {
 	rand.setSeed(0);
-	magicSchoolCounter = 1;
-	wisdomCounter = 1;
 }
 
 void CGHeroInstance::SecondarySkillsInfo::resetMagicSchoolCounter()
@@ -520,14 +517,14 @@ void CGHeroInstance::initObj(CRandomGenerator & rand)
 	}
 
 	//copy active (probably growing) bonuses from hero prototype to hero object
-	for(std::shared_ptr<Bonus> b : type->specialty)
+	for(const std::shared_ptr<Bonus> & b : type->specialty)
 		addNewBonus(b);
 	//dito for old-style bonuses -> compatibility for old savegames
 	for(SSpecialtyBonus & sb : type->specialtyDeprecated)
-		for(std::shared_ptr<Bonus> b : sb.bonuses)
+		for(const std::shared_ptr<Bonus> & b : sb.bonuses)
 			addNewBonus(b);
 	for(SSpecialtyInfo & spec : type->specDeprecated)
-		for(std::shared_ptr<Bonus> b : SpecialtyInfoToBonuses(spec, type->getIndex()))
+		for(const std::shared_ptr<Bonus> & b : SpecialtyInfoToBonuses(spec, type->getIndex()))
 			addNewBonus(b);
 
 	//initialize bonuses
@@ -539,19 +536,19 @@ void CGHeroInstance::initObj(CRandomGenerator & rand)
 void CGHeroInstance::recreateSecondarySkillsBonuses()
 {
 	auto secondarySkillsBonuses = getBonuses(Selector::sourceType()(Bonus::SECONDARY_SKILL));
-	for(auto bonus : *secondarySkillsBonuses)
+	for(const auto & bonus : *secondarySkillsBonuses)
 		removeBonus(bonus);
 
-	for(auto skill_info : secSkills)
+	for(const auto & skill_info : secSkills)
 		if(skill_info.second > 0)
 			updateSkillBonus(SecondarySkill(skill_info.first), skill_info.second);
 }
 
-void CGHeroInstance::updateSkillBonus(SecondarySkill which, int val)
+void CGHeroInstance::updateSkillBonus(const SecondarySkill & which, int val)
 {
 	removeBonuses(Selector::source(Bonus::SECONDARY_SKILL, which));
 	auto skillBonus = (*VLC->skillh)[which]->at(val).effects;
-	for (auto b : skillBonus)
+	for(const auto & b : skillBonus)
 		addNewBonus(std::make_shared<Bonus>(*b));
 }
 
@@ -579,12 +576,12 @@ double CGHeroInstance::getHeroStrength() const
 ui64 CGHeroInstance::getTotalStrength() const
 {
 	double ret = getFightingStrength() * getArmyStrength();
-	return (ui64) ret;
+	return static_cast<ui64>(ret);
 }
 
 TExpType CGHeroInstance::calculateXp(TExpType exp) const
 {
-	return (TExpType)(exp * (100 + valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, SecondarySkill::LEARNING))/100.0);
+	return static_cast<TExpType>(exp * (100 + valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, SecondarySkill::LEARNING)) / 100.0);
 }
 
 int32_t CGHeroInstance::getCasterUnitId() const
@@ -600,12 +597,12 @@ int32_t CGHeroInstance::getSpellSchoolLevel(const spells::Spell * spell, int32_t
 	{
 		int32_t thisSchool = std::max<int32_t>(
 			valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, cnf.skill),
-			valOfBonuses(Bonus::MAGIC_SCHOOL_SKILL, 1 << ((ui8)cnf.id))); //FIXME: Bonus shouldn't be additive (Witchking Artifacts : Crown of Skies)
+			valOfBonuses(Bonus::MAGIC_SCHOOL_SKILL, 1 << (static_cast<ui8>(cnf.id)))); //FIXME: Bonus shouldn't be additive (Witchking Artifacts : Crown of Skies)
 		if(thisSchool > skill)
 		{
 			skill = thisSchool;
 			if(outSelectedSchool)
-				*outSelectedSchool = (ui8)cnf.id;
+				*outSelectedSchool = static_cast<ui8>(cnf.id);
 		}
 	});
 
@@ -621,8 +618,8 @@ int64_t CGHeroInstance::getSpellBonus(const spells::Spell * spell, int64_t base,
 {
 	//applying sorcery secondary skill
 
-	base = (int64_t)(base * (100 + valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, SecondarySkill::SORCERY)) / 100.0);
-	base = (int64_t)(base * (100 + valOfBonuses(Bonus::SPELL_DAMAGE) + valOfBonuses(Bonus::SPECIFIC_SPELL_DAMAGE, spell->getIndex())) / 100.0);
+	base = static_cast<int64_t>(base * (100 + valOfBonuses(Bonus::SECONDARY_SKILL_PREMY, SecondarySkill::SORCERY)) / 100.0);
+	base = static_cast<int64_t>(base * (100 + valOfBonuses(Bonus::SPELL_DAMAGE) + valOfBonuses(Bonus::SPECIFIC_SPELL_DAMAGE, spell->getIndex())) / 100.0);
 
 	int maxSchoolBonus = 0;
 
@@ -631,17 +628,17 @@ int64_t CGHeroInstance::getSpellBonus(const spells::Spell * spell, int64_t base,
 		vstd::amax(maxSchoolBonus, valOfBonuses(cnf.damagePremyBonus));
 	});
 
-	base = (int64_t)(base * (100 + maxSchoolBonus) / 100.0);
+	base = static_cast<int64_t>(base * (100 + maxSchoolBonus) / 100.0);
 
 	if(affectedStack && affectedStack->creatureLevel() > 0) //Hero specials like Solmyr, Deemer
-		base = (int64_t)(base * double(100 + valOfBonuses(Bonus::SPECIAL_SPELL_LEV, spell->getIndex()) / affectedStack->creatureLevel()) / 100.0);
+		base = static_cast<int64_t>(base * static_cast<double>(100 + valOfBonuses(Bonus::SPECIAL_SPELL_LEV, spell->getIndex()) / affectedStack->creatureLevel()) / 100.0);
 
 	return base;
 }
 
 int64_t CGHeroInstance::getSpecificSpellBonus(const spells::Spell * spell, int64_t base) const
 {
-	base = (int64_t)(base * (100 + valOfBonuses(Bonus::SPECIFIC_SPELL_DAMAGE, spell->getIndex())) / 100.0);
+	base = static_cast<int64_t>(base * (100 + valOfBonuses(Bonus::SPECIFIC_SPELL_DAMAGE, spell->getIndex())) / 100.0);
 	return base;
 }
 
@@ -798,17 +795,17 @@ CStackBasicDescriptor CGHeroInstance::calculateNecromancy (const BattleResult &b
 		TConstBonusListPtr improvedNecromancy = getBonuses(Selector::type()(Bonus::IMPROVED_NECROMANCY));
 		if(!improvedNecromancy->empty())
 		{
-			auto getCreatureID = [necromancyLevel](std::shared_ptr<Bonus> bonus) -> CreatureID
+			auto getCreatureID = [necromancyLevel](const std::shared_ptr<Bonus> & bonus) -> CreatureID
 			{
 				const CreatureID legacyTypes[] = {CreatureID::SKELETON, CreatureID::WALKING_DEAD, CreatureID::WIGHTS, CreatureID::LICHES};
 				return CreatureID(bonus->subtype >= 0 ? bonus->subtype : legacyTypes[necromancyLevel]);
 			};
 			int maxCasualtyLevel = 1;
-			for(auto & casualty : casualties)
+			for(const auto & casualty : casualties)
 				vstd::amax(maxCasualtyLevel, VLC->creh->objects[casualty.first]->level);
 			// pick best bonus available
 			std::shared_ptr<Bonus> topPick;
-			for(std::shared_ptr<Bonus> newPick : *improvedNecromancy)
+			for(const std::shared_ptr<Bonus> & newPick : *improvedNecromancy)
 			{
 				// addInfo[0] = required necromancy skill, addInfo[1] = required casualty level
 				if(newPick->additionalInfo[0] > necromancyLevel || newPick->additionalInfo[1] > maxCasualtyLevel)
@@ -819,7 +816,7 @@ CStackBasicDescriptor CGHeroInstance::calculateNecromancy (const BattleResult &b
 				}
 				else
 				{
-					auto quality = [getCreatureID](std::shared_ptr<Bonus> pick) -> std::tuple<int, int, int>
+					auto quality = [getCreatureID](const std::shared_ptr<Bonus> & pick) -> std::tuple<int, int, int>
 					{
 						const CCreature * c = VLC->creh->objects[getCreatureID(pick)];
 						return std::tuple<int, int, int> {c->level, static_cast<int>(c->cost.marketValue()), -pick->additionalInfo[1]};
@@ -837,7 +834,7 @@ CStackBasicDescriptor CGHeroInstance::calculateNecromancy (const BattleResult &b
 		// raise upgraded creature (at 2/3 rate) if no space available otherwise
 		if(getSlotFor(creatureTypeRaised) == SlotID())
 		{
-			for(CreatureID upgraded : VLC->creh->objects[creatureTypeRaised]->upgrades)
+			for(const CreatureID & upgraded : VLC->creh->objects[creatureTypeRaised]->upgrades)
 			{
 				if(getSlotFor(upgraded) != SlotID())
 				{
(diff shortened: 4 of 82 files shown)


lib/mapping/CCampaignHandler.cpp
```diff
@@ -98,27 +80,29 @@ std::unique_ptr<CCampaign> CCampaignHandler::getCampaign( const std::string & na
...
 		std::string scenarioName = name.substr(0, name.find('.'));
 		boost::to_lower(scenarioName);
-		scenarioName += ':' + boost::lexical_cast<std::string>(g-1);
+		scenarioName += ':' + std::to_string(g - 1);
 
 		//set map piece appropriately, convert vector to string
 		ret->mapPieces[scenarioID].assign(reinterpret_cast< const char* >(file[g].data()), file[g].size());
 		CMapService mapService;
-		ret->scenarios[scenarioID].scenarioName = mapService.loadMapHeader((const ui8*)ret->mapPieces[scenarioID].c_str(), (int)ret->mapPieces[scenarioID].size(), scenarioName)->name;
+		ret->scenarios[scenarioID].scenarioName = mapService.loadMapHeader(reinterpret_cast<const ui8 *>(ret->mapPieces[scenarioID].c_str()),
```
IvanSavenko, line 88: Something went wrong with formatting here. 
For long methods I think we should use
```
someVeryLongMethodName(
<+1 tab> parameter1,
<+1 tab> parameter2 );
```

lib/mapping/CMapOperation.h
```diff
@@ -36,7 +36,7 @@ class DLL_LINKAGE CMapOperation : public boost::noncopyable
 	static const int FLIP_PATTERN_BOTH = 3;
 
 protected:
-	MapRect extendTileAround(const int3 & centerPos) const;
+	static MapRect extendTileAround(const int3 & centerPos);
```
IvanSavenko, line 39: Any reason to make it static?

lib/mapping/MapEditUtils.cpp
```diff
@@ -97,18 +102,12 @@ CTerrainSelection::CTerrainSelection(CMap * map) : CMapSelection(map)
 
 void CTerrainSelection::selectRange(const MapRect& rect)
 {
-	rect.forEach([this](const int3 pos)
-	{
-		this->select(pos);
-	});
+	rect.forEach([this](const int3 & pos) { this->select(pos); });
```
IvanSavenko, line 105: I think old formatting was better.

lib/mapping/MapEditUtils.cpp
```diff
@@ -97,18 +102,12 @@ CTerrainSelection::CTerrainSelection(CMap * map) : CMapSelection(map)
...
 }
 
 void CTerrainSelection::deselectRange(const MapRect& rect)
 {
-	rect.forEach([this](const int3 pos)
-	{
-		this->deselect(pos);
-	});
+	rect.forEach([this](const int3 & pos) { this->deselect(pos); });
```
IvanSavenko, line 110: Same here

lib/mapping/MapEditUtils.cpp
```diff
@@ -235,8 +238,8 @@ CTerrainViewPatternConfig::CTerrainViewPatternConfig()
 					{
 						std::vector<std::string> range;
 						boost::split(range, mapping, boost::is_any_of("-"));
-						terGroupPattern.mapping.push_back(std::make_pair(boost::lexical_cast<int>(range[0]),
-							boost::lexical_cast<int>(range.size() > 1 ? range[1] : range[0])));
+						terGroupPattern.mapping.emplace_back(boost::lexical_cast<int>(range[0]),
```
IvanSavenko, line 241: 1) surprised that lexical_cast survived. Or is that only works for int -> string conversion? There is `std::stoi` that can be used here.
2) Something happened with formatting?

lib/mapping/MapEditUtils.h
```diff
@@ -215,13 +215,12 @@ class DLL_LINKAGE CTerrainViewPatternConfig : public boost::noncopyable
...
 	CTerrainViewPatternConfig();
-	~CTerrainViewPatternConfig();
 
 	const std::vector<TVPVector> & getTerrainViewPatterns(TerrainId terrain) const;
-	boost::optional<const TerrainViewPattern &> getTerrainViewPatternById(std::string patternId, const std::string & id) const;
+	boost::optional<const TerrainViewPattern &> getTerrainViewPatternById(const std::string & patternId, const std::string & id) const;
 	boost::optional<const TVPVector &> getTerrainViewPatternsById(TerrainId terrain, const std::string & id) const;
 	const TVPVector * getTerrainTypePatternById(const std::string & id) const;
-	void flipPattern(TerrainViewPattern & pattern, int flip) const;
+	static void flipPattern(TerrainViewPattern & pattern, int flip);
```
IvanSavenko, line 223:
```suggestion
	void flipPattern(TerrainViewPattern & pattern, int flip) const;
```

lib/mapping/MapFormatJson.cpp
```diff
@@ -848,8 +850,9 @@ void CMapFormatJson::writeOptions(JsonSerializer & handler)
 }
 
 ///CMapPatcher
-CMapPatcher::CMapPatcher(JsonNode stream):
-	CMapFormatJson(),
+CMapPatcher::CMapPatcher(const JsonNode & stream)
+	:
```
IvanSavenko, line 854: Weird formatting

lib/mapping/MapFormatJson.h
```diff
@@ -95,12 +95,12 @@ class DLL_LINKAGE CMapFormatJson
 	/**
 	 * Reads one of triggered events
 	 */
-	void readTriggeredEvent(TriggeredEvent & event, const JsonNode & source);
+	static void readTriggeredEvent(TriggeredEvent & event, const JsonNode & source);
```
IvanSavenko, line 98:
```suggestion
	void readTriggeredEvent(TriggeredEvent & event, const JsonNode & source) const;

```

lib/mapping/MapFormatJson.h
```diff
@@ -95,12 +95,12 @@ class DLL_LINKAGE CMapFormatJson
...
 	 * Reads one of triggered events
 	 */
-	void readTriggeredEvent(TriggeredEvent & event, const JsonNode & source);
+	static void readTriggeredEvent(TriggeredEvent & event, const JsonNode & source);
 
 	/**
 	 * Writes one of triggered events
 	 */
-	void writeTriggeredEvent(const TriggeredEvent & event, JsonNode & dest);
+	static void writeTriggeredEvent(const TriggeredEvent & event, JsonNode & dest);
```
IvanSavenko, line 103:
```suggestion
	void writeTriggeredEvent(const TriggeredEvent & event, JsonNode & dest) const;

```

lib/rmg/CRmgTemplate.cpp
```diff
@@ -68,10 +68,7 @@ namespace rmg
 class TerrainEncoder
 {
 public:
-	static si32 decode(const std::string & identifier)
-	{
-		return *VLC->modh->identifiers.getIdentifier(VLC->modh->scopeGame(), "terrain", identifier);
-	}
+	static si32 decode(const std::string & identifier) { return *VLC->modh->identifiers.getIdentifier(CModHandler::scopeGame(), "terrain", identifier); }
```
IvanSavenko, line 71: Why is this one-liner now?

lib/rmg/CRmgTemplate.cpp
```diff
@@ -87,10 +84,7 @@ class ZoneEncoder
 		return boost::lexical_cast<si32>(json);
 	}
 
-	static std::string encode(si32 id)
-	{
-		return boost::lexical_cast<std::string>(id);
-	}
+	static std::string encode(si32 id) { return std::to_string(id); }
```
IvanSavenko, line 87: Same here

lib/rmg/CZonePlacer.cpp
```diff
@@ -358,20 +353,21 @@ void CZonePlacer::separateOverlappingZones(TZoneMap &zones, TForceVector &forces
...
 	float totalDistance = 0;
 	float totalOverlap = 0;
-	for (auto zone : distances) //find most misplaced zone
+	for(const auto & zone : distances) //find most misplaced zone
 	{
 		totalDistance += zone.second;
 		float overlap = overlaps[zone.first];
 		totalOverlap += overlap;
-		float ratio = (zone.second + overlap) / (float)totalForces[zone.first].mag(); //if distance to actual movement is long, the zone is misplaced
+		float ratio =
```
IvanSavenko, line 369: Weird formatting

lib/rmg/ObjectManager.h
```diff
@@ -53,10 +53,18 @@ class ObjectManager: public Modificator
 	bool createRequiredObjects();
 
 	int3 findPlaceForObject(const rmg::Area & searchArea, rmg::Object & obj, si32 min_dist, OptimizeType optimizer) const;
-	int3 findPlaceForObject(const rmg::Area & searchArea, rmg::Object & obj, std::function<float(const int3)> weightFunction, OptimizeType optimizer) const;
+	int3 findPlaceForObject(const rmg::Area & searchArea,
```
IvanSavenko, line 56: Questionable formatting. 
@kambala-decapitator any opinion on how should we format declarations in headers?

lib/rmg/TreasurePlacer.cpp
```diff
@@ -119,8 +118,8 @@ void TreasurePlacer::addAllPossibleObjects()
 	}
 	
 	//all following objects are unlimited
-	oi.maxPerZone = std::numeric_limits<ui32>().max();
-	
+	oi.maxPerZone = std::numeric_limits<unsigned int>::max();
```
IvanSavenko, line 121: Weird change.
```suggestion
	oi.maxPerZone = std::numeric_limits<ui32>::max();
```

lib/rmg/TreasurePlacer.cpp
```diff
@@ -653,11 +655,10 @@ ObjectInfo * TreasurePlacer::getRandomObject(ui32 desiredValue, ui32 currentValu
 		int r = generator.rand.nextInt(1, total);
 		
 		//binary search = fastest
-		auto it = std::lower_bound(thresholds.begin(), thresholds.end(), r,
-								   [](const std::pair<ui32, ObjectInfo*> &rhs, const int lhs)->bool
-								   {
-			return (int)rhs.first < lhs;
-		});
+		auto it = std::lower_bound(thresholds.begin(),
```
IvanSavenko, line 658: Weird formatting

IvanSavenko (commented):
Yet another proposition to remove conversion of methods to static rule

lib/CArtHandler.h
```diff
@@ -249,8 +249,8 @@ class DLL_LINKAGE CArtHandler : public CHandlerBase<ArtifactID, Artifact, CArtif
 
 	boost::optional<std::vector<CArtifact*>&> listFromClass(CArtifact::EartClass artifactClass);
 
-	ArtifactPosition stringToSlot(std::string slotName);
-	CArtifact::EartClass stringToClass(std::string className);
+	static ArtifactPosition stringToSlot(std::string slotName); //Static is intentional here, because VLC singletones is not good
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko i think than making this static is good because there is no need to use VLC singleton for just converting string to enum. Maybe there is an std:: method for this.

lib/mapObjects/CQuest.cpp
```diff
@@ -547,8 +541,7 @@ void CQuest::serializeJson(JsonSerializeFormat & handler, const std::string & fi
 
 }
 
-CGSeerHut::CGSeerHut() : IQuestObject(),
-	rewardType(NOTHING), rID(-1), rVal(-1)
+CGSeerHut::CGSeerHut() : rewardType(NOTHING), rID(-1), rVal(-1)
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko what is wrong with formatting here (if constructor initializers in one line)?

lib/mapping/MapEditUtils.cpp
rilian-la-te: 1. Only for int -> string, yes. I will manually change it to std::stoi.

lib/rmg/TreasurePlacer.cpp
rilian-la-te: How this should be done right, it both prev and now formatting is weird?

lib/mapObjects/CQuest.cpp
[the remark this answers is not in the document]
rilian-la-te: Constructor is removed altogether.

lib/mapping/MapEditUtils.cpp
rilian-la-te: reverted.

lib/mapping/MapEditUtils.cpp
rilian-la-te: reverted

lib/mapping/MapEditUtils.cpp
rilian-la-te: Done with both.

lib/mapping/MapEditUtils.h
rilian-la-te: done

lib/mapping/MapFormatJson.cpp
rilian-la-te: changed

lib/mapping/MapFormatJson.h
rilian-la-te: done

lib/mapping/MapFormatJson.h
rilian-la-te: done

lib/rmg/CRmgTemplate.cpp
rilian-la-te: reverted

lib/rmg/CRmgTemplate.cpp
rilian-la-te: reverted

lib/rmg/CZonePlacer.cpp
rilian-la-te: reverted

lib/rmg/ObjectManager.h
rilian-la-te: reverted

lib/rmg/TreasurePlacer.cpp
rilian-la-te: done

lib/rmg/TreasurePlacer.cpp
rilian-la-te: Redone it using auto helper variable, i think with helper variable it looks better.

lib/rmg/ObjectManager.h
rilian-la-te: But we need to establish rule for long declarations in header.

lib/mapObjects/CGHeroInstance.h
```diff
@@ -157,15 +157,15 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
...
+	void addSpellToSpellbook(const SpellID& spell);
+	void removeSpellFromSpellbook(const SpellID& spell);
+	bool spellbookContainsSpell(const SpellID& spell) const;
 	void removeSpellbook();
 	const std::set<SpellID> & getSpellsInSpellbook() const;
 	EAlignment::EAlignment getAlignment() const;
 	bool needsLastStack()const override;
 
-	ui32 getTileCost(const TerrainTile &dest, const TerrainTile &from, const TurnInfo * ti) const; //move cost - applying pathfinding skill, road and terrain modifiers. NOT includes diagonal move penalty, last move levelling
+	static ui32 getTileCost(const TerrainTile &dest, const TerrainTile &from, const TurnInfo * ti) ; //move cost - applying pathfinding skill, road and terrain modifiers. NOT includes diagonal move penalty, last move levelling
```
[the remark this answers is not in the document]
rilian-la-te: I thought than it should use this via get bonuses, but this is not the case strangely( Will convert back, I think method needs rewriting.

lib/mapObjects/CommonConstructors.cpp
```diff
@@ -185,7 +165,7 @@ void CDwellingInstanceConstructor::initTypeData(const JsonNode & input)
 	guards = input["guards"];
 }
 
-bool CDwellingInstanceConstructor::objectFilter(const CGObjectInstance *, std::shared_ptr<const ObjectTemplate>) const
+bool CDwellingInstanceConstructor::objectFilter(const CGObjectInstance * /*unused*/, std::shared_ptr<const ObjectTemplate> /*unused*/) const
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -528,7 +518,7 @@ void AObjectTypeHandler::init(const JsonNode & input)
 	initTypeData(input);
 }
 
-bool AObjectTypeHandler::objectFilter(const CGObjectInstance *, std::shared_ptr<const ObjectTemplate>) const
+bool AObjectTypeHandler::objectFilter(const CGObjectInstance * /*unused*/, std::shared_ptr<const ObjectTemplate> /*unused*/) const
```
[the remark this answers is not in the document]
rilian-la-te: done (but as obj and tmpl)

lib/mapObjects/CGHeroInstance.h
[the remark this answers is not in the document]
rilian-la-te: Replaced by const one.

lib/mapping/CCampaignHandler.cpp
rilian-la-te: Done with help of helper variable and your suggestion.

lib/mapping/CMapOperation.h
rilian-la-te: reverted

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -863,9 +855,9 @@ void CGTownInstance::deleteTownBonus(BuildingID::EBuildingID bid)
 	bonusingBuildings.erase(bonusingBuildings.begin() + i);
 
 	if(isVisitingBonus)
-		delete (CTownBonus *)freeIt;
+		delete dynamic_cast<CTownBonus *>(freeIt);
 	else if(isWeekBonus)
-		delete (COPWBonus *)freeIt;
+		delete dynamic_cast<COPWBonus *>(freeIt);
```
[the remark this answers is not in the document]
rilian-la-te: To correctly delete a pointer, we need to cast it to original type. There was C-Style cast, we can use dynamic_cast, static_cast or reinterpret_cast. What shoul be used here?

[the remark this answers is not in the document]
rilian-la-te: rechecked again: there is no need to cast, objects already has virtual destructor on IObjectInterface, so, just replaced to `delete freeIt;`

lib/rmg/TreasurePlacer.h
```diff
@@ -24,14 +24,12 @@ struct ObjectInfo
 	std::shared_ptr<const ObjectTemplate> templ;
 	ui32 value = 0;
 	ui16 probability = 0;
-	ui32 maxPerZone = -1;
+	ui32 maxPerZone = 1;
```
IvanSavenko, line 27: Now I like in-place initialization even less. Because such cases WILL appear over time making value from header even less useful than before

rilian-la-te: @IvanSavenko but is there any point to make in-place init to -1 and in one and only constructor initialize field to 1?

lib/mapping/MapFormatJson.cpp
```diff
@@ -189,7 +189,7 @@ namespace TriggeredEventsDetail
 			break;
 		}
 
-		return VLC->modh->makeFullIdentifier("", metaclassName, identifier);
+		return CModHandler::makeFullIdentifier("", metaclassName, identifier);
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko so, you prefer to abandon https://clang.llvm.org/extra/clang-tidy/checks/readability/static-accessed-through-instance.html?
There was a case of already static function of VLC, who we should call it by instance?

lib/mapObjects/CQuest.h
```diff
@@ -47,13 +47,13 @@ class DLL_LINKAGE CQuest final
 	static const std::string  & missionName(Emission mission);
 	static const std::string  & missionState(int index);
 
-	si32 qid; //unique quest id for serialization / identification
+	si32 qid = -1; //unique quest id for serialization / identification
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko this made constructor removal possible? Do we need to save a constructor only to hide a few constants?

lib/mapObjects/CObjectHandler.h
```diff
@@ -114,26 +112,25 @@ class DLL_LINKAGE CGObjectInstance : public IObjectInterface
 {
 public:
 	/// Position of bottom-right corner of object on map
-	int3 pos;
+	int3 pos = {-1, -1, -1};
```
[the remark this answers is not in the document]
rilian-la-te: To delete a constructor, as before. Do we need to save one and only constructor, if it defines only placeholders? What approach should be used then?

lib/rmg/TreasurePlacer.h
IvanSavenko: That's the point. Or lack of it.
Since it is now possible to initialize member in two places, we will eventually end up in such scenario where in-place init and in-constructor init set some field to different values, leading to misleading information in header.

rilian-la-te: I think if we have only one constructor without parameters, which initializes fields to primitive types and nullptr, we can use in-place init and trivial constructor. You think than hide implementation details, but define a non-trivial constructor with initializers is better?

lib/mapObjects/CQuest.h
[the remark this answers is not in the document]
IvanSavenko: I can't say that either old version or new one is good code.
This is one of many of vcmi classes that uses factory-like initialization, meaning that values in the constructor are just placeholders.

Actual values are set someplace else, in this case - in map loader.
At the very least, old code was hiding this useless information from public interface.
Now we're exposing it in header without gaining anything else of value.
Yeah, sure we no longer have ~10 lines or so of constructor. But we've just replaced one poor code with another poor code. 

If the goal was to modernize/improve code then I think this change did nothing in regards to improvements.

lib/mapping/MapFormatJson.cpp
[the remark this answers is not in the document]
IvanSavenko: Yeah, probably.
Not against such usage of static methods, so no need to revert existing code, but also see no point in switching to it.

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -437,7 +428,7 @@ void CGHeroInstance::onHeroVisit(const CGHeroInstance * h) const
 	}
 	else if(ID == Obj::PRISON)
 	{
-		int txt_id;
+		int txt_id = 0;
```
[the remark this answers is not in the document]
rilian-la-te: reverted

lib/mapObjects/CQuest.h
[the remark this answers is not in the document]
rilian-la-te: I think than less lines of code is better, because it have better readability and better maintainability and this headers is not public.

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -862,10 +840,8 @@ void CGTownInstance::deleteTownBonus(BuildingID::EBuildingID bid)
 
 	bonusingBuildings.erase(bonusingBuildings.begin() + i);
 
-	if(isVisitingBonus)
-		delete (CTownBonus *)freeIt;
-	else if(isWeekBonus)
-		delete (COPWBonus *)freeIt;
+	//There is no need to cast, we have virtual base destructor
```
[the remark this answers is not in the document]
rilian-la-te: removed

lib/rmg/TreasurePlacer.h
IvanSavenko: In this particular case - yeah. Not even sure how this code ended up like this.

Such approach *is* nice for small c-like structs that have no methods of their own and have definition less than ~10 lines in size.
In such cases (including this code) yeah, I think it is fine to use inplace init. 

But for anything larger and more complex than that - probably better to avoid it, specifically to prevent such cases.

rilian-la-te: I tried to force removal of constructors (it is almost all hand changes) only in cases when struct have `serialize()` method and follow netpacks pattern. I think there is a good change. In big classes - not so much.

lib/mapping/MapFormatJson.cpp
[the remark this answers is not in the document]
IvanSavenko: Realistically speaking - I don't think that it would help with VLC removal.

Currently my (very long-term) ideas regarding VLC is to either make it constant during the game (solving main issues with it) or to make it member of base library classes, e.g. IObjectInterface (allowing different VLC's for different objects).
But outright removal sound like a fantasy.

So in terms of VLC removal this is as good as removing a pebble from a mountain. We likely have thousands of instances of VLC access, if not more. Can't say that this is a bad change, but I don't see it as good change either. 

If you wish to keep it - OK, up to you I guess?

lib/mapObjects/CObjectHandler.h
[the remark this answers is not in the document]
IvanSavenko: Basically same as discussion below:
https://github.com/vcmi/vcmi/pull/1567#discussion_r1103700205
Except that this is about CGObjectInstance, base class for half of library code. Which IMO should be kept as minimal as possible and as clear as possible.

lib/mapObjects/CGHeroInstance.h
```diff
@@ -54,29 +54,29 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
...
+	ui8 sex = 0xff;
 
 	std::string nameCustom;
 	std::string biographyCustom;
 
 	bool inTownGarrison; // if hero is in town garrison
-	ConstTransitivePtr<CGTownInstance> visitedTown; //set if hero is visiting town or in the town garrison
-	ConstTransitivePtr<CCommanderInstance> commander;
-	const CGBoat *boat; //set to CGBoat when sailing
+	ConstTransitivePtr<CGTownInstance> visitedTown = nullptr; //set if hero is visiting town or in the town garrison
```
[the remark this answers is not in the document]
rilian-la-te: done for all `ConstTransitivePtr` instances.

lib/mapping/MapFormatJson.cpp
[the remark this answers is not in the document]
rilian-la-te: I wish to keep it, because in this place it makes sense (makeFullIdentifier is more like library function instead of method)

lib/mapObjects/CObjectHandler.h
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko I think this is better to init int3's default constructor to -1,-1,-1 rather than 0,0,0.
What do you think about it?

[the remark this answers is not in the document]
IvanSavenko: Hmm...
Potential problem is that there may be code that relies on having current (0,0,0) as default value. Might be difficult to track every such case.

As for using (-1,-1,-1) as default constructor - if goal is to have "clearly invalid" value as default then something even less "normal" might be a better idea, e.g. `std::numeric_limits::min/max`.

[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko there is a comment in int3.h:

```c++
	si32 x, y, z;

	//c-tor: x, y, z initialized to 0
	int3() : x(0), y(0), z(0) {} // I think that x, y, z should be left uninitialized.
```

So, we can just declare default initializers for x, y, z as clearly invalid (either -1 or min/max). What value do you prefer for?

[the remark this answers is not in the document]
IvanSavenko: That comment is ancient. I won't rely on it without some checking to make sure that there is no code that uses default values. For example, by removing/making private this constructor and check what fails to compile.

lib/mapObjects/CGHeroInstance.h
```diff
@@ -54,29 +54,29 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
...
+					 //			765
+	mutable ui8 isStanding = true, tacticFormationEnabled;
 
 	//////////////////////////////////////////////////////////////////////////
 
-	ConstTransitivePtr<CHero> type;
-	TExpType exp; //experience points
-	ui32 level; //current level of hero
+	ConstTransitivePtr<CHero> type = nullptr;
+	TExpType exp = 0xffffffff; //experience points
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko what do you think if we define function like this somewhere in headers?
```c++
template <typename T>
static constexpr T placeholder()
{
    static_assert(std::is_integral(T));
    return reinterpret_cast<T>(-1);
}
```
and then initialize all in-place placeholder value with placeholder?

lib/mapObjects/CObjectHandler.h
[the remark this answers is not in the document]
rilian-la-te: int3 default constructor is called anywhere (likely in each file of lib/*). It seems almost impossible to track if code rely on fact than default values is correct, I think.

lib/rmg/TreasurePlacer.h
rilian-la-te: I think in this place I can save this change. @IvanSavenko do you agree?
In-place init will be difficult to read if we have more than 1 default constructor, I agree with you in this point.

lib/mapObjects/CGHeroInstance.h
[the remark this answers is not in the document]
IvanSavenko: I would rather use
- std::numeric_limits::max for signed (and maybe unsigned)
- not-a-number for floating types
- And not sure what to do with enum's

However there is a problem with this approach - "T". Return type can't be deducted by compiler, so you'll have to initalize every member in form `int field = vstd::uninitialized<int>();` which is a bit too wordy. OK for constructor I guess, but definitely not for in-place init.

lib/mapObjects/CObjectHandler.h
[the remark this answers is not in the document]
IvanSavenko: Hmm... perhaps we can try tricking compiler/clang-tidy into telling that? By NOT initializing values in constructor? Should in theory lead to "maybe uninitialized" warning where needed.

Anyway, I am not against better tracking of uninitialized values, but this needs to be well-tested to minimize chance of regressions. If compiler can't do this part it will have to be done manually.

lib/rmg/TreasurePlacer.h
IvanSavenko: Yes, here change is fine, my comment was mostly to highlight potential issue with using in-place init on large classes that are too big for all relevant code to be visible at glance.

lib/mapObjects/CGHeroInstance.h
```diff
@@ -54,29 +54,29 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
...
+	ui32 level = 1; //current level of hero
 	si32 portrait; //may be custom
 	si32 mana; // remaining spell points
 	std::vector<std::pair<SecondarySkill,ui8> > secSkills; //first - ID of skill, second - level of skill (1 - basic, 2 - adv., 3 - expert); if hero has ability (-1, -1) it meansthat it should have default secondary abilities
 	ui32 movement; //remaining movement points
-	ui8 sex;
+	ui8 sex = 0xff;
```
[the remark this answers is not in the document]
rilian-la-te: For all CHeroInstance: moved any non-nullptr initialization into a constructor.

lib/mapObjects/CGHeroInstance.h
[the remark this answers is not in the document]
rilian-la-te: Moved init back to constructor, no uninitialized func for now.

rilian-la-te:
Will fix constexpr build tomorrow (on my local machine builds fine, it is old compiler issue).

lib/mapObjects/CObjectHandler.h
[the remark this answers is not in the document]
rilian-la-te: Reverted change, but set TODO.

lib/mapObjects/CQuest.h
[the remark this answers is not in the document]
rilian-la-te: Reverted change, but set TODO

lib/CArtHandler.h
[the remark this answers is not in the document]
rilian-la-te: StringToSlot became a constructor, stringToClass become static with TODO for reworking EartClass.

lib/mapping/CMap.cpp
```diff
@@ -719,7 +707,7 @@ CMapEditManager * CMap::getEditManager()
 	return editManager.get();
 }
 
-void CMap::resetStaticData()
+void CMap::resetStaticData() const
```
IvanSavenko, line 710: I think methods like this should not have `const` added.
Yes, technically they do quality for const since method makes no changes to object state.
But logically, such methods should not be marked as const - since they should not be used with const object.

In other words, if method has some (signigicant) side effects, it should not be marked const.

lib/rmg/CMapGenerator.cpp
```diff
@@ -275,13 +274,13 @@ void CMapGenerator::genZones()
 	logGlobal->info("Zones generated successfully");
 }
 
-void CMapGenerator::createWaterTreasures()
+void CMapGenerator::createWaterTreasures() const
```
IvanSavenko, line 277: Another questionable const addition.
Yes, state of object itself remains the same, but internal state changes.
Probably should not be marked as const

lib/rmg/CRmgTemplate.h
```diff
@@ -116,7 +116,7 @@ class DLL_LINKAGE ZoneOptions
 
 	ZoneOptions();
 
-	ZoneOptions & operator=(const ZoneOptions & other);
+	ZoneOptions & operator=(const ZoneOptions & other) = default;
```
IvanSavenko, line 119: Can be removed entirely?

lib/rmg/RmgArea.h
```diff
@@ -30,9 +30,9 @@ namespace rmg
 	public:
 		Area() = default;
 		Area(const Area &);
-		Area(const Area &&);
-		Area(const Tileset & tiles);
-		Area(const Tileset & relative, const int3 & position); //create from relative positions
+		Area(const Area &&) noexcept;
```
IvanSavenko, line 33: Is explicit copy constructor necessary here? 

Similarly, is explicit move constructor necessary here? 
Also, shouldn't it be defined as `Area(Area &&)`?

lib/rmg/TreasurePlacer.cpp
```diff
@@ -651,13 +650,10 @@ ObjectInfo * TreasurePlacer::getRandomObject(ui32 desiredValue, ui32 currentValu
 	else
 	{
 		int r = generator.rand.nextInt(1, total);
-		
+		auto sfun = [](const std::pair<ui32, ObjectInfo *> & rhs, const ui32 lhs) -> bool { return static_cast<int>(rhs.first) < lhs; };
```
IvanSavenko, line 653: Formatting - please no one-liner lambda's
Also - what `sfun` means? `sorter`, `sorterFunctor` or simply `functor` is probably better

lib/mapObjects/CObjectClassesHandler.cpp
```diff
@@ -241,7 +241,7 @@ ObjectClass * CObjectClassesHandler::loadFromJson(const std::string & scope, con
 	{
 		if (!subData.second["index"].isNull())
 		{
-			std::string const & subMeta = subData.second["index"].meta;
+			const std::string & subMeta = subData.second["index"].meta;
```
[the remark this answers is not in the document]
rilian-la-te: `git clang-format` maybe. It is a first PR when I tried to apply clang-format to all changes.

lib/rmg/CRmgTemplate.h
rilian-la-te: @IvanSavenko why? We do not need to assign ZoneOptions?

lib/rmg/RmgArea.h
rilian-la-te: It is move constructor. It should not be const at theory, but it was const, clang-tidy just added noexcept here.

lib/rmg/TreasurePlacer.cpp
rilian-la-te: sfun is sorterfunction. I can rename it to sorter, if you wish.

lib/rmg/CRmgTemplate.h
IvanSavenko: Unless I am missing something, assignment operator is one of compiler-generated methods?

lib/mapObjects/CObjectClassesHandler.cpp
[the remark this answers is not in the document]
rilian-la-te: Should I do something here?

lib/rmg/RmgArea.h
IvanSavenko: Yeah, not questioning your change, but rather *really* weird code both before & after.

lib/rmg/TreasurePlacer.cpp
IvanSavenko: Yeah, please minimize abbreviating things

lib/mapObjects/CObjectClassesHandler.cpp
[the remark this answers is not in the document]
IvanSavenko: No, it's fine. Resolving.

lib/rmg/CRmgTemplate.h
rilian-la-te: In this cause you are right, but compiler generates assignments in much less places than constructors.

rilian-la-te: removed.

lib/rmg/CMapGenerator.cpp
rilian-la-te: removed

lib/mapping/CMap.cpp
rilian-la-te: removed

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -626,8 +607,7 @@ bool CGTownInstance::hasCapitol() const
 	return hasBuilt(BuildingID::CAPITOL);
 }
 
-CGTownInstance::CGTownInstance()
-	:CGDwelling(), IShipyard(this), IMarket(this), town(nullptr), builded(0), destroyed(0), identifier(0), alignment(0xff)
+CGTownInstance::CGTownInstance(): IShipyard(this), IMarket(this), town(nullptr), builded(0), destroyed(0), identifier(0), alignment(0xff)
```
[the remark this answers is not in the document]
rilian-la-te: Made this one-per-line

lib/rmg/TreasurePlacer.cpp
rilian-la-te: renamed to sorter and uninlined.

lib/rmg/RmgArea.h
rilian-la-te: removed const, unsure what more should I do manually.

lib/mapObjects/CArmedInstance.cpp
```diff
@@ -47,9 +46,8 @@ CArmedInstance::CArmedInstance()
 }
 
 CArmedInstance::CArmedInstance(bool isHypotetic)
-	:CBonusSystemNode(isHypotetic), nonEvilAlignmentMix(this, nonEvilAlignmentMixSelector)
+	:CBonusSystemNode(isHypotetic), nonEvilAlignmentMix(this, nonEvilAlignmentMixSelector), battle(nullptr)
```
kambala-decapitator, line 49: probably deserves proper formatting?

kambala-decapitator (changes requested):

lib/mapping/MapFormatJson.cpp
```diff
@@ -810,12 +812,12 @@ void CMapFormatJson::serializePredefinedHeroes(JsonSerializeFormat & handler)
...
 			auto predefinedHero = handler.enterStruct(p.first);
 
-			CGHeroInstance * hero = new CGHeroInstance();
+			auto * hero = new CGHeroInstance();
 			hero->ID = Obj::HERO;
             hero->setHeroTypeName(p.first);
             hero->serializeJsonDefinition(handler);
 
-            map->predefinedHeroes.push_back(hero);
+			map->predefinedHeroes.emplace_back(hero);
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/CArtHandler.cpp
```diff
@@ -410,17 +410,17 @@ CArtifact * CArtHandler::loadFromJson(const std::string & scope, const JsonNode
...
+ArtifactPosition::ArtifactPosition(std::string slotName):
+	num(ArtifactPosition::PRE_FIRST)
 {
 #define ART_POS(x) { #x, ArtifactPosition::x },
 	static const std::map<std::string, ArtifactPosition> artifactPositionMap = { ART_POS_LIST };
 #undef ART_POS
 	auto it = artifactPositionMap.find (slotName);
 	if (it != artifactPositionMap.end())
-		return it->second;
+		this->num = it->second;
```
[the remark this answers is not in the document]
rilian-la-te: removed this

lib/mapObjects/CGHeroInstance.h
```diff
@@ -55,9 +55,10 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
 	//////////////////////////////////////////////////////////////////////////
 
 	ui8 moveDir; //format:	123
-					//		8 4
-					//		765
-	mutable ui8 isStanding, tacticFormationEnabled;
+				 //			8 4
+				 //			765
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -1904,16 +1882,15 @@ const std::string CGTownBuilding::getCustomBonusGreeting(const Bonus & bonus) co
...
+	std::string until;
 
 	if(bonus.type == Bonus::MORALE)
 		param = VLC->generaltexth->allTexts[384];
 	else if(bonus.type == Bonus::LUCK)
 		param = VLC->generaltexth->allTexts[385];
 
-	until = bonus.duration == (ui16)Bonus::ONE_BATTLE
-		? VLC->generaltexth->translate("vcmi.townHall.greetingCustomUntil") : ".";
+	until = bonus.duration == static_cast<ui16>(Bonus::ONE_BATTLE) ? VLC->generaltexth->translate("vcmi.townHall.greetingCustomUntil") : ".";
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -1904,16 +1882,17 @@ const std::string CGTownBuilding::getCustomBonusGreeting(const Bonus & bonus) co
...
 	if(bonus.type == Bonus::MORALE)
 		param = VLC->generaltexth->allTexts[384];
 	else if(bonus.type == Bonus::LUCK)
 		param = VLC->generaltexth->allTexts[385];
 
-	until = bonus.duration == (ui16)Bonus::ONE_BATTLE
-		? VLC->generaltexth->translate("vcmi.townHall.greetingCustomUntil") : ".";
+	until = bonus.duration == static_cast<ui16>(Bonus::ONE_BATTLE) 
+		    ? VLC->generaltexth->translate("vcmi.townHall.greetingCustomUntil")
+			: ".";
```
kambala-decapitator, line 1895: broken alignment:
```suggestion
	until = bonus.duration == static_cast<ui16>(Bonus::ONE_BATTLE) 
		? VLC->generaltexth->translate("vcmi.townHall.greetingCustomUntil")
		: ".";
```

lib/mapObjects/CGHeroInstance.h
```diff
@@ -53,11 +53,12 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
...
 
 	//////////////////////////////////////////////////////////////////////////
-
-	ui8 moveDir; //format:	123
-					//		8 4
-					//		765
-	mutable ui8 isStanding, tacticFormationEnabled;
+	//format:	123
+	//			8 4
+	//			765
```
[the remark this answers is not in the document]
rilian-la-te: No, because for me it was not. I have 4 spaces for tab setting, maybe because of it I do not see broken alignment. Reformatted with spaces only.

lib/mapObjects/CGTownInstance.cpp
rilian-la-te: Reformatted with tabs only.

kambala-decapitator (approved):

IvanSavenko (approved):

rilian-la-te:
@IvanSavenko @kambala-decapitator waiting for merge)

PR #1569, opened 2023-02-11
combined arts swap

SoundSSGood:
swap

### DIFF
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -1023,43 +1023,44 @@ bool CArtifactInstance::isPart(const CArtifactInstance *supposedPart) const
 	return supposedPart == this;
 }
 
-bool CCombinedArtifactInstance::canBePutAt(const CArtifactSet *artSet, ArtifactPosition slot, bool assumeDestRemoved) const
+bool CCombinedArtifactInstance::canBePutAt(const CArtifactSet * artSet, ArtifactPosition slot, bool assumeDestRemoved) const
 {
 	if(slot == ArtifactPosition::TRANSITION_POS)
 		return true;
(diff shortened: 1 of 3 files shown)


lib/CArtHandler.cpp
```diff
@@ -1023,43 +1023,44 @@ bool CArtifactInstance::isPart(const CArtifactInstance *supposedPart) const
...
+	fittingSet.artifactsWorn = artSet->artifactsWorn;
+	auto removedArt = fittingSet.getArt(slot);
+	if(assumeDestRemoved && removedArt)
 	{
-		for(auto art = constituentsToBePlaced.begin(); art != constituentsToBePlaced.end(); art++)
+		if(removedArt->canBeDisassembled())
 		{
-			// pos == slot because we can remove already worn artifact only from that slot. That is our main destination
-			if(art->art->canBePutAt(artSet, pos, pos == slot))
+			auto contitutient = dynamic_cast<CCombinedArtifactInstance*>(removedArt);
```
IvanSavenko, line 1040: `contitutient`? Typo?
And from context, `combinedArtToRemove` is probably a better name

IvanSavenko (approved):
Looks OK code-wise.

To clarify, this change allows swapping two combined artifacts directly, which was not allowed before. Right?

lib/CArtHandler.cpp
SoundSSGood: > To clarify, this change allows swapping two combined artifacts directly, which was not allowed before. Right?
Yeah.
contitutient and combined - both of them occur in code. But "combined" better, I think. Will fix.

nullkiller:
Please rebase on top of develop (or merge develop) so that build artifacts contain your other change

nullkiller:
Tested, found no issues. So we still can not swap angelic alliance and armor of damned but swapping two armors works.

PR #1573, opened 2023-02-12
Small optimizations

rilian-la-te:
This is a set of small optimizations for VCMI:

1. make ConstTransitivePtr always init self as nullptr
2. make int3 as constexpr, which allows us to calculate map distances between fixed points at compile time
3. Remove some unused variables
4. Remove int3 overflow from client, which can be noticed only if constexpr.
5. Add missing include to GUIClasses.cpp, which fixes build with clang++ 15 on ArchLinux.

### DIFF
--- a/client/adventureMap/mapHandler.cpp
+++ b/client/adventureMap/mapHandler.cpp
@@ -23,6 +23,7 @@
 #include "../../lib/mapObjects/CGHeroInstance.h"
 #include "../../lib/mapObjects/CObjectClassesHandler.h"
 #include "../../lib/mapping/CMap.h"
+#include "../../lib/Color.h"
 #include "../../lib/CConfigHandler.h"
 #include "../../lib/CGeneralTextHandler.h"
 #include "../../lib/CStopWatch.h"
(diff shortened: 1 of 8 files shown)


client/adventureMap/mapHandler.cpp
```diff
@@ -925,7 +925,7 @@ void CMapHandler::CMapBlitter::blit(SDL_Surface * targetSurf, const MapDrawingIn
 		{
 			for (realPos.y = initPos.y, pos.y = topTile.y; pos.y < topTile.y + tileCount.y; pos.y++, realPos.y += tileSize)
 			{
-				const int3 color(0x555555, 0x555555, 0x555555);
+				const int3 color(0x55, 0x55, 0x55);
```
IvanSavenko, line 928: Please use ColorRGBA instead. This is not position, but color. No need to use int3 here

lib/int3.h
```diff
@@ -130,29 +123,29 @@ class int3
...
 		const si32 dx = (x - o.x);
 		const si32 dy = (y - o.y);
 		return (ui32)(dx*dx) + (ui32)(dy*dy);
 	}
 	//returns distance on Oxy plane (z coord is not used)
-	double dist2d(const int3 & o) const
+	constexpr double dist2d(const int3 & o) const
 	{
-		return std::sqrt((double)dist2dSQ(o));
+		return vstd::ct_sqrt((double)dist2dSQ(o));
```
IvanSavenko, line 135: Is there any need for compile-tile distance calculation?
This code will also execute in runtime. Meaning potential change in behavior or performance.

Especially since old distance computation had double as result but now it uses int.

IvanSavenko (commented):
I think it is better to keep distance calculation non-constexpr, as before, without addition of weird sqrt computations.

lib/int3.h
rilian-la-te: Made dist2d as non-constexpr, as you request.

client/adventureMap/mapHandler.cpp
rilian-la-te: Done.

IvanSavenko (approved):

rilian-la-te:
@IvanSavenko please merge this, without this develop is not building anymore (error in #1561 is an example).

PR #1574, opened 2023-02-12
combined art move

SoundSSGood:
Allows to move combined to slots locked by the same artifact.
move

This PR makes a change in the mechanics of client-server art transfers.
Now, as soon as the user selects a slot with an artifact, "moveArtifact" is immediately sent to the server. Artifact moves to transit location. This frees up the slot. This approach simplifies the code.

Another approach is when we send "move Artifact" when the user puts the Artifact to slot (as it was before). Code becomes too unwieldy.

Third approach: These changes are not significant. We leave code as is.

I prefer the first one (this PR). I think that +1 transver is not a problem.

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -149,19 +149,12 @@ void CHeroArtPlace::clickLeft(tribool down, bool previousState)
 				select();
 			}
 		}
-		else if(ourArt == ourOwner->commonInfo->src.art) //restore previously picked artifact
-		{
-			deselect();
-		}
-		else //perform artifact transition
+		// Perform artifact transition
+		else if(ourArt != ourOwner->commonInfo->src.art)
 		{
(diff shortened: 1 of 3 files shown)


IvanSavenko:
What happens if players closes hero window with picked up artifact? Same for exchange window.

SoundSSGood:
> What happens if players closes hero window with picked up artifact? Same for exchange window.

The same as in the OH3. The artifact will be placed in the first suitable place.

nullkiller:
Tested, it works. One thing what if we make bulk art swap while one art is in transition slot? And close the window afterwards? IF this works I approve this PR. Has nothing against it.

nullkiller (approved):

SoundSSGood:
> placed in the first suitable place

The artifact will not be moved. It will remain in the transit location. And after closing the window it will be placed in the first suitable place of the hero. Hero from which the artifact was taken.

PR #1580, opened 2023-02-14
Modernize next part

rilian-la-te:
Modernize logging, filesystem and events.

### DIFF
--- a/lib/filesystem/CArchiveLoader.h
+++ b/lib/filesystem/CArchiveLoader.h
@@ -68,9 +68,9 @@ public:
 	void updateFilteredFiles(std::function<bool(const std::string &)> filter) const override {}
 	std::unordered_set<ResourceID> getFilteredFiles(std::function<bool(const ResourceID &)> filter) const override;
 	/** Extracts one archive entry to the specified subfolder. Used for Video and Sound */
-	void extractToFolder(const std::string & outputSubFolder, CInputStream & fileStream, ArchiveEntry entry);
+	void extractToFolder(const std::string & outputSubFolder, CInputStream & fileStream, const ArchiveEntry & entry) const;
 	/** Extracts one archive entry to the specified subfolder. Used for Images, Sprites, etc */
-	void extractToFolder(const std::string & outputSubFolder, const std::string & mountPoint, ArchiveEntry entry);
+	void extractToFolder(const std::string & outputSubFolder, const std::string & mountPoint, ArchiveEntry entry) const;
 
 private:
 	/**
--- a/lib/filesystem/CCompressedStream.cpp
+++ b/lib/filesystem/CCompressedStream.cpp
@@ -26,7 +26,7 @@ si64 CBufferedStream::read(ui8 * data, si64 size)
 {
 	ensureSize(position + size);
 
-	auto start = buffer.data() + position;
+	auto * start = buffer.data() + position;
 	si64 toRead = std::min<si64>(size, buffer.size() - position);
 
 	std::copy(start, start + toRead, data);
@@ -62,7 +62,7 @@ si64 CBufferedStream::getSize()
 
 void CBufferedStream::ensureSize(si64 size)
 {
-	while ((si64)buffer.size() < size && !endOfFileReached)
+	while(static_cast<si64>(buffer.size()) < size && !endOfFileReached)
 	{
 		si64 initialSize = buffer.size();
 		si64 currentStep = std::min<si64>(size, buffer.size());
@@ -127,7 +127,7 @@ si64 CCompressedStream::readMore(ui8 *data, si64 size)
 
 	int decompressed = inflateState->total_out;
 
-	inflateState->avail_out = (uInt)size;
+	inflateState->avail_out = static_cast<uInt>(size);
 	inflateState->next_out = data;
 
 	do
@@ -140,7 +140,7 @@ si64 CCompressedStream::readMore(ui8 *data, si64 size)
 			if (availSize != compressedBuffer.size())
 				gzipStream.reset();
 
-			inflateState->avail_in = (uInt)availSize;
+			inflateState->avail_in = static_cast<uInt>(availSize);
 			inflateState->next_in  = compressedBuffer.data();
 		}
 
@@ -162,7 +162,7 @@ si64 CCompressedStream::readMore(ui8 *data, si64 size)
 			break;
 		default:
 			if (inflateState->msg == nullptr)
-				throw std::runtime_error("Decompression error. Return code was " + boost::lexical_cast<std::string>(ret));
+				throw std::runtime_error("Decompression error. Return code was " + std::to_string(ret));
 			else
 				throw std::runtime_error(std::string("Decompression error: ") + inflateState->msg);
 		}
--- a/lib/filesystem/CZipSaver.cpp
+++ b/lib/filesystem/CZipSaver.cpp
@@ -72,7 +72,7 @@ CZipOutputStream::~CZipOutputStream()
 
 si64 CZipOutputStream::write(const ui8 * data, si64 size)
 {
-	int ret = zipWriteInFileInZip(handle, (const void*)data, (unsigned)size);
+	int ret = zipWriteInFileInZip(handle, data, static_cast<unsigned>(size));
 
 	if (ret == ZIP_OK)
 		return size;
@@ -82,12 +82,11 @@ si64 CZipOutputStream::write(const ui8 * data, si64 size)
 
 ///CZipSaver
 CZipSaver::CZipSaver(std::shared_ptr<CIOApi> api, const boost::filesystem::path & path):
-	ioApi(api),
+	ioApi(std::move(api)),
 	zipApi(ioApi->getApiStructure()),
-	handle(nullptr),
+	handle(zipOpen2_64(path.c_str(), APPEND_STATUS_CREATE, nullptr, &zipApi)),
 	activeStream(nullptr)
 {
-	handle = zipOpen2_64((const void *) & path, APPEND_STATUS_CREATE, nullptr, &zipApi);
 
 	if (handle == nullptr)
 		throw std::runtime_error("CZipSaver: Failed to create archive");
--- a/lib/filesystem/FileStream.cpp
+++ b/lib/filesystem/FileStream.cpp
@@ -166,7 +166,7 @@ std::streamoff FileBuf::seek(std::streamoff off, std::ios_base::seekdir way)
 			throw std::ios_base::failure("bad seek direction");
 		}
 	}();
-	if(std::fseek(GETFILE, (long)off, src))
+	if(std::fseek(GETFILE, static_cast<long>(off), src))
 		throw std::ios_base::failure("bad seek offset");
 
 	return static_cast<std::streamsize>(std::ftell(GETFILE));
--- a/lib/filesystem/Filesystem.cpp
+++ b/lib/filesystem/Filesystem.cpp
@@ -29,7 +29,7 @@ CResourceHandler CResourceHandler::globalResourceHandler;
 
 CFilesystemGenerator::CFilesystemGenerator(std::string prefix, bool extractArchives):
 	filesystem(new CFilesystemList()),
-	prefix(prefix),
+	prefix(std::move(prefix)),
 	extractArchives(extractArchives)
 {
 }
@@ -48,9 +48,9 @@ CFilesystemGenerator::TLoadFunctorMap CFilesystemGenerator::genFunctorMap()
 
 void CFilesystemGenerator::loadConfig(const JsonNode & config)
 {
-	for(auto & mountPoint : config.Struct())
+	for(const auto & mountPoint : config.Struct())
 	{
-		for(auto & entry : mountPoint.second.Vector())
+		for(const auto & entry : mountPoint.second.Vector())
 		{
 			CStopWatch timer;
 			logGlobal->trace("\t\tLoading resource at %s%s", prefix, entry["path"].String());
@@ -82,7 +82,7 @@ void CFilesystemGenerator::loadDirectory(const std::string &mountPoint, const Js
 	std::string URI = prefix + config["path"].String();
 	int depth = 16;
 	if (!config["depth"].isNull())
-		depth = (int)config["depth"].Float();
+		depth = static_cast<int>(config["depth"].Float());
 
 	ResourceID resID(URI, EResType::DIRECTORY);
 
@@ -117,7 +117,7 @@ void CFilesystemGenerator::loadJsonMap(const std::string &mountPoint, const Json
 	if (filename)
 	{
 		auto configData = CResourceHandler::get("initial")->load(ResourceID(URI, EResType::TEXT))->readAll();
-		const JsonNode configInitial((char*)configData.first.get(), configData.second);
+		const JsonNode configInitial(reinterpret_cast<char *>(configData.first.get()), configData.second);
 		filesystem->addLoader(new CMappedFileLoader(mountPoint, configInitial), false);
 	}
 }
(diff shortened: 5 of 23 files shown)


lib/filesystem/CCompressedStream.cpp
```diff
@@ -26,7 +26,7 @@ si64 CBufferedStream::read(ui8 * data, si64 size)
 {
 	ensureSize(position + size);
 
-	auto start = buffer.data() + position;
+	auto *start = buffer.data() + position;
```
IvanSavenko, line 29:
```suggestion
	auto * start = buffer.data() + position;
```

lib/filesystem/CZipSaver.cpp
```diff
@@ -82,12 +82,11 @@ si64 CZipOutputStream::write(const ui8 * data, si64 size)
 
 ///CZipSaver
 CZipSaver::CZipSaver(std::shared_ptr<CIOApi> api, const boost::filesystem::path & path):
-	ioApi(api),
+	ioApi(std::move(api)),
 	zipApi(ioApi->getApiStructure()),
-	handle(nullptr),
+	handle(zipOpen2_64((const void *) & path, APPEND_STATUS_CREATE, nullptr, &zipApi)),
```
IvanSavenko, line 87: What is this? I guess we were lucky and this thing somehow worked before?
```suggestion
	handle(zipOpen2_64(path.c_str(), APPEND_STATUS_CREATE, nullptr, &zipApi)),
```

lib/filesystem/FileStream.cpp
```diff
@@ -45,11 +45,11 @@ voidpf call_zopen64 (const zlib_filefunc64_32_def* pfilefunc,const void*filename
 
 long call_zseek64(const zlib_filefunc64_32_def* pfilefunc, voidpf filestream, ZPOS64_T offset, int origin)
 {
-	if (pfilefunc->zfile_func64.zseek64_file != NULL)
+	if (pfilefunc->zfile_func64.zseek64_file != nullptr)
 		return (*(pfilefunc->zfile_func64.zseek64_file)) (pfilefunc->zfile_func64.opaque, filestream, offset, origin);
 	else
 	{
-		uLong offsetTruncated = (uLong)offset;
+		auto offsetTruncated = (uLong)offset;
```
IvanSavenko, line 52: Looks like auto-upgrade have not replaced c-style casts in this file for some reason?
Not the only case.

Or is file too messy for clang somehow?

lib/logging/CBasicLogConfigurator.cpp
```diff
@@ -145,7 +145,7 @@ EConsoleTextColor::EConsoleTextColor CBasicLogConfigurator::getConsoleColor(cons
 
 void CBasicLogConfigurator::deconfigure()
 {
-	auto l = CLogger::getGlobalLogger();
+	auto *l = CLogger::getGlobalLogger();
```
IvanSavenko, line 148:
```suggestion
	auto * l = CLogger::getGlobalLogger();
```

lib/logging/CLogger.h
```diff
@@ -134,13 +134,13 @@ class DLL_LINKAGE CLogFormatter
...
+	CLogFormatter(const CLogFormatter & copy) = default;
+	CLogFormatter(CLogFormatter && move) = default;
 
-	CLogFormatter(const std::string & pattern);
+	CLogFormatter(std::string pattern);
 
-	CLogFormatter & operator=(const CLogFormatter & copy);
-	CLogFormatter & operator=(CLogFormatter && move);
+	CLogFormatter & operator=(const CLogFormatter & copy) = default;
+	CLogFormatter & operator=(CLogFormatter && move) = default;
```
IvanSavenko, line 143: Can be removed I think?
```suggestion
```

lib/logging/CLogger.h
```diff
@@ -134,13 +134,13 @@ class DLL_LINKAGE CLogFormatter
 {
 public:
 	CLogFormatter();
-	CLogFormatter(const CLogFormatter & copy);
-	CLogFormatter(CLogFormatter && move);
+	CLogFormatter(const CLogFormatter & copy) = default;
+	CLogFormatter(CLogFormatter && move) = default;
```
IvanSavenko, line 138: Can be removed I think?
```suggestion
```

lib/logging/CLogger.cpp
```diff
@@ -231,7 +231,8 @@ CLogger * CLogManager::getLogger(const CLoggerDomain & domain)
 std::vector<std::string> CLogManager::getRegisteredDomains() const
 {
 	std::vector<std::string> domains;
-	for (auto& pair : loggers)
+	domains.reserve(loggers.size());
+for (const auto& pair : loggers)
```
IvanSavenko, line 235:
```suggestion
	for (const auto& pair : loggers)
```

lib/filesystem/Filesystem.h
```diff
@@ -72,7 +72,7 @@ class DLL_LINKAGE CResourceHandler
 	 * @return Returns an instance of resource loader.
 	 */
 	static ISimpleResourceLoader * get();
-	static ISimpleResourceLoader * get(std::string identifier);
+	static ISimpleResourceLoader * get(const std::string& identifier);
```
IvanSavenko, line 75:
```suggestion
	static ISimpleResourceLoader * get(const std::string & identifier);
```

lib/filesystem/Filesystem.cpp
```diff
@@ -126,10 +126,10 @@ ISimpleResourceLoader * CResourceHandler::createInitial()
 {
 	//temporary filesystem that will be used to initialize main one.
 	//used to solve several case-sensivity issues like Mp3 vs MP3
-	auto initialLoader = new CFilesystemList();
+	auto *initialLoader = new CFilesystemList();
```
IvanSavenko, line 129: Not the only occurence
```suggestion
	auto * initialLoader = new CFilesystemList();
```

lib/filesystem/FileStream.cpp
```diff
@@ -91,37 +91,34 @@ static voidpf ZCALLBACK fopen64_file_func(voidpf opaque, const void* filename, i
...
-	ret = FTELLO_FUNC((FILE *)stream);
+	ZPOS64_T ret = FTELLO_FUNC((FILE *)stream);
 	return ret;
 }
 
 static long ZCALLBACK fseek64_file_func(voidpf  opaque, voidpf stream, ZPOS64_T offset, int origin)
 {
 	int fseek_origin = 0;
-	long ret;
+	long ret = 0;
```
IvanSavenko, line 121: Can you remove it entirely? It is basically unused anyway.
```
Line #137: ret = -1;   -> return -1;
Line #139: return ret; -> return 0;
```

rilian-la-te:
Will address your changes later today. What is better by formatting - this or #1567 ?

IvanSavenko:
Probably this one? But it is much smaller so less coverage. Way less possible formatting cases than in #1567 

this:
- wrong formatting of pointers/references - `Type* var` or `Type *var` when `Type * var` should be used

#1567 :
- one-liner lambda's instead of multi-line (even though old code was OK)
- one-liner constructor initializers (same as in old code)
- one-liner assignment instead of nicely formatted multi-line (old code was OK)

rilian-la-te:
@IvanSavenko what is wrong with on-line constructor initializers if there are 2-3 short arguments?

IvanSavenko:
> @IvanSavenko what is wrong with on-line constructor initializers if there are 2-3 short arguments?

1) Define "short". Something  like ` Point():x(0), y(o)` is short. But in your previous PR there were several one-liners that were around 50-100 symbols long.
2) Even for short cases, I think one-per-line is better for readability. You at glance see how many and which fields are set, you don't need to read entire line.
3) Consistency. I'd rather have simple & clean rules, without too many conditions.
4) Version control history/blame. With one per line (especially - with leading comma) you can edit/add parameter without "dirtying" other parameters when viewing git blame.

rilian-la-te:
@IvanSavenko short for me = less than one editor line (usually 80 symbols).

4. Agree
2 and 3 is questional. Even clang-format allow one-liners.

lib/filesystem/CCompressedStream.cpp
rilian-la-te: all formatting is passed via clang-format.

lib/filesystem/CZipSaver.cpp
rilian-la-te: done

lib/filesystem/FileStream.cpp
rilian-la-te: I just skipped it because this was pure C, so, just for purity.
Passed this file also.

lib/logging/CBasicLogConfigurator.cpp
rilian-la-te: all formatting is passed via clang-format.

lib/logging/CLogger.h
rilian-la-te: removed both constructors and operators.

lib/logging/CLogger.h
rilian-la-te: done

lib/logging/CLogger.cpp
rilian-la-te: done

lib/filesystem/Filesystem.cpp
rilian-la-te: done

lib/filesystem/FileStream.cpp
rilian-la-te: done

lib/filesystem/Filesystem.h
rilian-la-te: done

rilian-la-te:
@IvanSavenko addressed your changes.

lib/filesystem/CZipSaver.cpp
```diff
@@ -72,7 +72,7 @@ CZipOutputStream::~CZipOutputStream()
 
 si64 CZipOutputStream::write(const ui8 * data, si64 size)
 {
-	int ret = zipWriteInFileInZip(handle, (const void*)data, (unsigned)size);
+	int ret = zipWriteInFileInZip(handle, (const void *)data, static_cast<unsigned>(size));
```
IvanSavenko, line 75: `const ui8 *` should be castable to `const void*` without explicit casts
```suggestion
	int ret = zipWriteInFileInZip(handle, data, static_cast<unsigned>(size));
```

IvanSavenko (approved):

lib/filesystem/CZipSaver.cpp
rilian-la-te: done, let's look to CI.

lib/filesystem/CArchiveLoader.h
```diff
@@ -68,9 +68,9 @@ class DLL_LINKAGE CArchiveLoader : public ISimpleResourceLoader
 	void updateFilteredFiles(std::function<bool(const std::string &)> filter) const override {}
 	std::unordered_set<ResourceID> getFilteredFiles(std::function<bool(const ResourceID &)> filter) const override;
 	/** Extracts one archive entry to the specified subfolder. Used for Video and Sound */
-	void extractToFolder(const std::string & outputSubFolder, CInputStream & fileStream, ArchiveEntry entry);
+	static void extractToFolder(const std::string & outputSubFolder, CInputStream & fileStream, const ArchiveEntry & entry);
```
nullkiller, line 71: as far as I remember there was an idea to not make methods static, especially if there is non-static overload. The idea is that this method still describes instance behavior and may need some object state later, during some refactoring or whatever.

lib/filesystem/FileStream.cpp
```diff
@@ -91,37 +91,34 @@ static voidpf ZCALLBACK fopen64_file_func(voidpf opaque, const void* filename, i
 			if (mode & ZLIB_FILEFUNC_MODE_CREATE)
 				mode_fopen = "wb";
 
-	if ((filename != NULL) && (mode_fopen != NULL))
-		file = FOPEN_FUNC((const char*)filename, mode_fopen);
+	if((filename != nullptr) && (mode_fopen != nullptr))
+		file = FOPEN_FUNC((const char *)filename, mode_fopen);
```
nullkiller, line 95: No static_cast here?

lib/filesystem/Filesystem.cpp
```diff
@@ -117,7 +117,7 @@ void CFilesystemGenerator::loadJsonMap(const std::string &mountPoint, const Json
 	if (filename)
 	{
 		auto configData = CResourceHandler::get("initial")->load(ResourceID(URI, EResType::TEXT))->readAll();
-		const JsonNode configInitial((char*)configData.first.get(), configData.second);
+		const JsonNode configInitial(reinterpret_cast<char *>(configData.first.get()), configData.second);
```
nullkiller, line 120: What is the difference between reinterpret_cast and static_cast?

lib/filesystem/CArchiveLoader.h
IvanSavenko: That's what I keep telling whenever I see addition of `static`.
User code should make no assumptions as to whether method accesses class members in call.

Missed this case probably

lib/filesystem/CZipSaver.cpp
```diff
@@ -82,12 +82,11 @@ si64 CZipOutputStream::write(const ui8 * data, si64 size)
 
 ///CZipSaver
 CZipSaver::CZipSaver(std::shared_ptr<CIOApi> api, const boost::filesystem::path & path):
-	ioApi(api),
+	ioApi(std::move(api)),
 	zipApi(ioApi->getApiStructure()),
-	handle(nullptr),
+	handle(zipOpen2_64(path.c_str(), APPEND_STATUS_CREATE, nullptr, &zipApi)),
```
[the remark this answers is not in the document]
IvanSavenko: Probably equivalent since old code didn't crash.
But this is pure luck.
zipOpen2_64 requires C string (char*), and not boost::filesystem::path.

Old code worked due to omni-powerful C-style cast but was anything but correct

lib/logging/CBasicLogConfigurator.h
```diff
@@ -33,7 +33,7 @@ class DLL_LINKAGE CBasicLogConfigurator
 	void configureDefault();
 
 	/// Removes all targets from the global logger.
-	void deconfigure();
+	static void deconfigure();
```
nullkiller, line 36: HAve not seen where it is used.

nullkiller (approved):

lib/filesystem/Filesystem.cpp
IvanSavenko: static_cast generally converts between convertible types (e.g int ->float, or classes within same type hierarchy)
reinterpret_cast allows well, reinterpreting variable as another, unrelated type.

So, for example static_cast<int>(float) will convert float to integer, using standard rules (drop fractional part).
static_cast'ing int* to float* would fail since these types are not related
Meanwhile, reinterpret_cast<int*>(float*) will give you your pointer to int, however data under that pointer will not be modified and still contain floating point representation in its IEEE form.

rilian-la-te: Here we required to use reinterpret_cast, because compiler does not allows us to use static_cast,

lib/filesystem/FileStream.cpp
rilian-la-te: This code is vanished from develop, so, no need for modernize.

lib/filesystem/CZipSaver.cpp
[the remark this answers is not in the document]
rilian-la-te: Yes, this is more correct version.

lib/filesystem/CArchiveLoader.h
rilian-la-te: removed static

lib/logging/CBasicLogConfigurator.h
rilian-la-te: removed this function

rilian-la-te: it seems it needed on some platforms for cancel logger init. Restored.

IvanSavenko (approved):

Issue #1036, opened 2022-10-01
Smoother scrolling


Laserlicht:
It would be good if a smaller scroll interval was possible. This is especially noticeable when scrolling with the middle mouse button, when the map section jumps.

A smaller, smoother scroll interval is already used for player movements.

Issue #1413, opened 2023-01-11
glitched movement range and path display for flying hero


kambala-decapitator:
Had this in develop / 1.1 / 1.0

![Screenshot 2023-01-11 at 08 20 18](https://user-images.githubusercontent.com/1557784/211724365-61e219ac-824a-4efd-89e6-f4df72e49905.png)

https://bugs.vcmi.eu/view.php?id=3128

IvanSavenko:
But this *is* the most optimal path to reach your destination 🤣

kambala-decapitator:
true :) but green arrow is definitely wrong.

IvanSavenko:
I doubt that H3 has arrows for U-turn. So should probably block pathfinder from revisiting tiles

kambala-decapitator:
more examples:

![Screenshot 2023-01-14 at 11 03 47](https://user-images.githubusercontent.com/1557784/212463520-9b357366-dbab-44fb-a6a8-5e21635f7018.png) ![Screenshot 2023-01-14 at 11 13 12](https://user-images.githubusercontent.com/1557784/212463521-a5b7bfc6-627b-4d5c-84f5-1617312c1a5b.png)

no strange bottom arrow in the second case

kambala-decapitator:
SS from oh3:
![Screenshot 2023-01-15 at 17 12 57](https://user-images.githubusercontent.com/1557784/212546027-e2e69fc0-0988-4149-8ab5-40dfa2884f37.png)

kambala-decapitator:
another example:

![Screenshot 2023-01-16 at 14 18 48](https://user-images.githubusercontent.com/1557784/212700837-254ce2e6-a143-466f-b231-d9cf81d76fb8.png)

Titans block the road, without Fly hero needs to take a detour. Going to the tile to the left of destination would initiate battle, to the right - no arrows visual bug.

Issue #797, opened 2022-09-03
planned hero path isn't restored on loading game


kambala-decapitator:
1. plan some path for a hero (green / black arrows on the ground)
2. end turn
3. load last autosave
4. no green / black arrows on the ground

PR #1581, opened 2023-02-14
Adventure Map - rendering rewrite

IvanSavenko:
Currently, most notable player reports about vcmi are all about adventure map - low FPS during hero movement, freeze on end of hero movement, etc.

Unfortunately, due to ancient code it is hard to do anything with our CPlayerInterface/AdvMapInt/TerrainRect/mapHandler chain of interdependent classes.

So for now I am attempting to redo map rendering part and resolve current architecture issues. Current plan for this PR:
- fix poor performance during smooth map movement (e.g. hero move)
- allow scrolling of adventure map with pixel-level precision (and not per-tile)
- implement groundwork for unconfirmed hero move / move cancellation (necessary to solve MP freeze). Freeze itself will likely have to wait for later though.

Current progress:
- [x] Implement terrain rendering
- [x] Implement objects rendering
- [x] Make adventure animation time-based
- [x] Implement map scrolling
- [x] Make map scrolling pixel-level
- [x] Move hero path rendering to map renderer class
- [x] Implement scaled View World mode
- [x] Implement View World icons 
- [x] Implement View Earth/Air spells
- [x] Implement hero movement - instant/no animation (for code convenience & in case of huge movement speed)
- [x] Implement hero movement - animated hero sprite move ("standard" move form)
- [x] Implement hero movement - hero fading (embarking/disembarking)
- [x] Implement hero movement - terrain fading (teleport spell/monolith visit)
- [x] Implement object fade-out (e.g. picking a resource, defeating monster)
- [x] Recheck timing of hero movement, make sure that they match H3
- [x] Recheck enemy movement functionality
- [x] Recheck timing of enemy movement, make sure that they match H3
- [x] Implement terrain palette shift animation
- [x] Optimize map rendering - do not redraw unmodified tiles
- [x] Optimize map rendering - use SDL's blit copy where possible
- [x] Implement Puzzle Map functionality
- [x] Restore (implement?) embarking/disembarking sound
- [x] Restore (implement?) water movement sound
- [x] Recheck and correct object render priority between H3 and VCMI
- [x] Recheck whether all map options work in new settings window

Known regressions:
- [x] Restore ambient sounds functionality
- [x] Restore hole check for Grail digging functinality, move it to library.
- [x] Embarking/disembarking should focus/center adventure map on new position of hero
- [x] Fastest movement speed should focus/center adventure map 
- [x] Hero dismiss/defeat should focus/center adventure map only after animation is over
- [x] Minimap does not updates radar position on edge scrolling
- [x] Fix right-click terrain tooltip
- [x] Disable interaction with UI during terrain animations
- [x] During enemy turn infobar switches to selected hero

Other issues related to adventure map (may be fixed in this PR)
- #797
- #843
- #1413


rilian-la-te:
@IvanSavenko maybe make point constexpr? It is fundamental enough)

IvanSavenko:
> @IvanSavenko maybe make point constexpr? It is fundamental enough)

Not *against* it, but don't see any benefits from it. Go ahead if you wish to, but I'll probably keep it as it unless I need constexpr Point.

Rect class is in the same situation by the way.

rilian-la-te:
@IvanSavenko rect is also can benefit from being constexpr. I will do it, but after finishing my ideas with bonuses and modernizations.

dydzio0614:
gigachad rework :tada:

dydzio0614:
BTW consider checking hero defeat & hero dismiss as well

Laserlicht:
resolves #1036

Alucard648:
Will it be possible to implement custom icons for map objects in View World and custom scouting spells like View Fire?

IvanSavenko:
@Alucard648 
Not something I have plans for at the moment.

dydzio0614:
consider checking water / rivers animations

IvanSavenko:
> consider checking water / rivers animations

Already did, they work.

rilian-la-te:
@IvanSavenko did you change tiles fetching algo from server?

IvanSavenko:
> @IvanSavenko did you change tiles fetching algo from server?

What tiles fetching are you talking about?

rilian-la-te:
@IvanSavenko I thinking about dimension door fix. If you preload cached tiles from server, maybe we can check jump availability based on this cached tiles?

IvanSavenko:
As of right now, client always have full gamestate synchronized with server.

This has not changed in this PR.

DD issue comes from callback methods which blocks access to hidden tiles.

client/mapRenderer/MapViewController.cpp
```diff
@@ -173,13 +216,15 @@ void MapViewController::onHeroMoved(const CGHeroInstance * obj, const int3 & fro
...
+
+		if (actionVisible)
+			setViewCenter(movingObject->visitablePos());
 	}
 }
 
 void MapViewController::onHeroRotated(const CGHeroInstance * obj, const int3 & from, const int3 & dest)
 {
-	//TODO
+	//TODO. Or no-op?
```
dydzio0614, line 260: reminds me of https://youtu.be/vCtik1q_IsQ?t=13

IvanSavenko: No such animations in H3 files unfortunately.

rilian-la-te: What this method should do?

IvanSavenko: Most likely - nothing. So will remove it once I confirm this.

This is for handling situation where hero need to face object to interact with. For example, hero is moving alongside road and wants to pick resource pile next to the road.
In this case there will be movement event/netpack but hero must only turn to face object (e.g. resource), without actual movement.

But since hero rotation is set by netpack applier and H3 has no animations for turning around there is no need to do anything in map rendering.

config/widgets/settings/adventureOptionsTab.json
```diff
@@ -31,31 +31,31 @@
...
+					"index": 100,
 					"type": "toggleButton",
 					"image": "sysopb3",
 					"help": "core.help.351",
 					"position": {"x": 96, "y": 0}
 				},
 
 				{
-					"index": 16,
+					"index": 0,
```
dydzio0614, line 58: I did not manually test this, but note that in H3 thi speed is not exactly instant - for example try to move from one corner of XL map to another

IvanSavenko: Yes, but it seems to be extremely fast, like 1 tile per ms. Too fast for my video capture measurements.

At this point I would call it instant and any visual delays were likely not intended.
It is not instant in vcmi either, mostly due to server-client communication delays

IvanSavenko:
Finally done, marking as ready for review.

Still need to block interaction with UI during map animations (and maybe some other corner cases) but other than that - ready.

rilian-la-te:
is map moving with mouse works? I do not remember if it is original h3 feature, but maybe nice to have.

IvanSavenko:
Yes. Work also in current develop, but only on per-tile level.
In branch works on pixel-level.
Activated by holding mouse wheel.

rilian-la-te:
@IvanSavenko but what about mouse scrolling by moving to corners? I cannot find it working with #1518(

IvanSavenko:
Definitely works in this branch and in 1.1
It was broken in 1.2 at some point and I do remember fixing it... somewhere? Possibly in this branch, so fix is not in develop right now

rilian-la-te:
@IvanSavenko Mouse scrolling by corners works in current develop, but not by adventure map corners. but by window corners (so, in right side is incorrect). Something is changed?

IvanSavenko:
Works as designed then. Scrolling should work by *window* corners

rilian-la-te:
image
Here as it should work, but not

image
Here as it works now, but should not.

rilian-la-te:
Notice cursor location in both screenshots.

rilian-la-te:
@IvanSavenko it is possible to implement adventure map corners scrolling at option?

IvanSavenko:
Don't see any reason to.
Not how it works in H3, not some sort of a clear improvement, not what user expects - since moving cursor to right panel will also scroll the map

Laserlicht:
For me, the scrolling over the screen edges in this pull request currently does not work. Maybe this is due to the new menu. The scroll speed is not saved either.

For the scrolling with mouse button:
In itself, it would also be nice to make this configurable instead of the middle button. I find the left and right mouse buttons much more comfortable (since the middle button is actually the mouse wheel nowadays).
But the right button has the help menu. Therefore, my favorite is the left button (like google maps).

IvanSavenko:
@Laserlicht 
Fixed scrolling. Thanks for reporting.

As for mouse button - will probably keep as it. Seems like personal preference, so not everyone will agree with you. And mouse wheel press is unused in H3 so we're free to use it as we wish to.

There is actually hidden (for now) way to use left click for scrolling - by enabling swipe logic used by mobile platforms (Android/ios).
If this mode is on, vcmi will allow moving map via finger swipe. And since right now vcmi does not distinguish between finger touch and mouse click, left click can be used as finger swipe.

Will probably add option in settings window for that.

Laserlicht:
@IvanSavenko 
Yes, of course it is a personal preference.

It would be great to have an option to use the swipe logic from the smartphone on the pc with a mouse. Regardless of this, there are also pc-based systems with touch (e.g. steam deck).

rilian-la-te:
New issue with new loader. Hero is covered by +1 knowledge garden.
изображение

PR #1582, opened 2023-02-14
vcmi: modernize lib/battle

rilian-la-te:
Next part, currently only lib/battle.


lib/battle/CBattleInfoCallback.h
```diff
@@ -156,7 +156,7 @@ class DLL_LINKAGE CBattleInfoCallback : public virtual CBattleInfoEssentials
 	AttackableTiles getPotentiallyShootableHexes(const  battle::Unit* attacker, BattleHex destinationTile, BattleHex attackerPos) const;
 	std::vector<const battle::Unit *> getAttackedBattleUnits(const battle::Unit* attacker, BattleHex destinationTile, bool rangedAttack, BattleHex attackerPos = BattleHex::INVALID) const; //calculates range of multi-hex attacks
 	std::set<const CStack*> getAttackedCreatures(const CStack* attacker, BattleHex destinationTile, bool rangedAttack, BattleHex attackerPos = BattleHex::INVALID) const; //calculates range of multi-hex attacks
-	bool isToReverse(const battle::Unit *attacker, const battle::Unit *defender) const; //determines if attacker standing at attackerHex should reverse in order to attack defender
+	static bool isToReverse(const battle::Unit * attacker, const battle::Unit * defender); //determines if attacker standing at attackerHex should reverse in order to attack defender
```
IvanSavenko, line 159: Same here. Only access via instance in callback

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1552,23 +1553,19 @@ SpellID CBattleInfoCallback::getRandomBeneficialSpell(CRandomGenerator & rand, c
 		case SpellID::SHIELD:
 		case SpellID::FIRE_SHIELD: // not if all enemy units are shooters
 		{
-			auto walker = getAliveEnemy([&](const CStack * stack) //look for enemy, non-shooting stack
-			{
-				return !stack->canShoot();
-			});
+				const auto * walker = getAliveEnemy([&](const CStack * stack) //look for enemy, non-shooting stack
+													{ return !stack->canShoot(); });
```
IvanSavenko, line 1557: Formatting. Original code used correct style for lambda's inside method call

Also - alignment of entire block is 2 indents deep compared to brackets

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1552,23 +1553,19 @@ SpellID CBattleInfoCallback::getRandomBeneficialSpell(CRandomGenerator & rand, c
...
 			break;
 		case SpellID::AIR_SHIELD: //only against active shooters
 		{
-			auto shooter = getAliveEnemy([&](const CStack * stack) //look for enemy, non-shooting stack
-			{
-				return stack->canShoot();
-			});
-			if (!shooter)
-				continue;
+				const auto * shooter = getAliveEnemy([&](const CStack * stack) //look for enemy, non-shooting stack
```
IvanSavenko, line 1565: Same here.

lib/battle/CBattleInfoEssentials.h
```diff
@@ -76,15 +76,15 @@ class DLL_LINKAGE CBattleInfoEssentials : public virtual CCallbackBase, public I
...
 	si8 battleGetTacticsSide() const override; //returns which side is in tactics phase, undefined if none (?)
 
-	bool battleCanFlee(PlayerColor player) const;
-	bool battleCanSurrender(PlayerColor player) const;
+	bool battleCanFlee(const PlayerColor & player) const;
+	bool battleCanSurrender(const PlayerColor & player) const;
 
-	ui8 otherSide(ui8 side) const;
-	PlayerColor otherPlayer(PlayerColor player) const;
+	static ui8 otherSide(ui8 side);
```
IvanSavenko, line 82: Please revert. Callbacks should only use access through instance.

lib/battle/AccessibilityInfo.cpp
```diff
@@ -25,15 +25,15 @@ bool AccessibilityInfo::accessible(BattleHex tile, bool doubleWide, ui8 side) co
...
+	if(at(tile) != EAccessibility::ACCESSIBLE && (at(tile) != EAccessibility::GATE || side != BattleSide::DEFENDER))
 		return false;
 
 	if(doubleWide)
 	{
 		auto otherHex = battle::Unit::occupiedHex(tile, doubleWide, side);
 		if(!otherHex.isValid())
 			return false;
-		if(at(otherHex) != EAccessibility::ACCESSIBLE && !(at(otherHex) == EAccessibility::GATE && side == BattleSide::DEFENDER))
+		if(at(otherHex) != EAccessibility::ACCESSIBLE && (at(otherHex) != EAccessibility::GATE || side != BattleSide::DEFENDER))
```
nullkiller, line 36: Any chance we can refactor this code duplication into a method?

nullkiller: Also such complex conditions may be better to split onto s few operators. In a separate method it may be way easier to organize

lib/battle/BattleHex.cpp
```diff
@@ -84,7 +84,8 @@ std::pair<si16, si16> BattleHex::getXY() const
 
 BattleHex& BattleHex::moveInDirection(EDir dir, bool hasToBeValid)
 {
-	si16 x(getX()), y(getY());
+	si16 x(getX());
```
nullkiller, line 87: I would prefer using = operator here.

lib/battle/BattleHex.cpp
```diff
@@ -135,7 +136,7 @@ std::vector<BattleHex> BattleHex::neighbouringTiles() const
 {
 	std::vector<BattleHex> ret;
 	ret.reserve(6);
-	for(EDir dir = EDir(0); dir <= EDir(5); dir = EDir(dir+1))
+	for(EDir dir = static_cast<EDir>(0); dir <= static_cast<EDir>(5); dir = static_cast<EDir>(dir + 1))
```
nullkiller, line 139: Is there an enum value for 5?

IvanSavenko: Yes, all valid EDir's have valid values in enum, including 0 and 5. 
I am currently skipping cases like this during review.

We should probably recheck all our static_cast's. I expect *a lot* of places where static_cast is an indicator of poor code, like this one.
However I think this should go into a separate PR(s), for easier reviewing and minimal regressions

lib/battle/BattleHex.cpp
```diff
@@ -145,28 +146,31 @@ std::vector<BattleHex> BattleHex::allNeighbouringTiles() const
...
 
 uint8_t BattleHex::getDistance(BattleHex hex1, BattleHex hex2)
 {
-	int y1 = hex1.getY(), y2 = hex2.getY();
+	int y1 = hex1.getY();
+	int y2 = hex2.getY();
 
 	// FIXME: Omit floating point arithmetics
-	int x1 = (int)(hex1.getX() + y1 * 0.5), x2 = (int)(hex2.getX() + y2 * 0.5);
+	int x1 = static_cast<int>(hex1.getX() + y1 * 0.5);
```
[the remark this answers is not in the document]
rilian-la-te: @nullkiller maybe implicit conversion to float is significant, because y1 and y2 can be odd.

lib/battle/BattleInfo.cpp
```diff
@@ -101,7 +101,7 @@ namespace CGH
 			std::vector<int> pom;
 			for(const JsonNode &value : level.Vector())
 			{
-				pom.push_back((int)value.Float());
+			pom.push_back(static_cast<int>(value.Float()));
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/battle/CBattleInfoCallback.h
rilian-la-te: Do not understand your static hate, but done(

lib/battle/CBattleInfoCallback.h
```diff
@@ -106,10 +106,10 @@ class DLL_LINKAGE CBattleInfoCallback : public virtual CBattleInfoEssentials
 
 	std::vector<BattleHex> battleGetAvailableHexes(const ReachabilityInfo & cache, const battle::Unit * unit) const;
 
-	int battleGetSurrenderCost(PlayerColor Player) const; //returns cost of surrendering battle, -1 if surrendering is not possible
+	int battleGetSurrenderCost(const PlayerColor & Player) const; //returns cost of surrendering battle, -1 if surrendering is not possible
 	ReachabilityInfo::TDistances battleGetDistances(const battle::Unit * unit, BattleHex assumedPosition) const;
 	std::set<BattleHex> battleGetAttackedHexes(const CStack* attacker, BattleHex destinationTile, BattleHex attackerPos = BattleHex::INVALID) const;
-	bool isEnemyUnitWithinSpecifiedRange(BattleHex attackerPosition, const battle::Unit * defenderUnit, unsigned int range) const;
+	static bool isEnemyUnitWithinSpecifiedRange(BattleHex attackerPosition, const battle::Unit * defenderUnit, unsigned int range);
```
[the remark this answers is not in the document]
rilian-la-te: Do not understand your static hate, but done(

lib/battle/CBattleInfoCallback.cpp
rilian-la-te: done

lib/battle/CBattleInfoEssentials.h
rilian-la-te: Do not understand your static hate, but done(

lib/battle/AccessibilityInfo.cpp
rilian-la-te: Done in separate method.

lib/battle/BattleHex.cpp
rilian-la-te: done

lib/battle/CBattleInfoCallback.h
IvanSavenko: isToReverse is an interesting example.
IIRC the only reason it can be made static now is because of battle refactoring I did back in December. Before it used some other (non-static) methods from callback

Problem is, opposite is also true. It is possible that in future this function will be changed and will require access to `this` again. And any call that is not made through an instance will be broken.

IvanSavenko: And this is not theoretical example. Back during said battle refactoring I did ended up with this particular problem - I had to make static function non-static and had a fun refactoring session in order to provide instance for every such call.

rilian-la-te: In ideal non-static world, we can use Rust-like approach with 
```c++
struct A {
    int a;
    int b;
    int c;
    float d;
}

namespace A_impl {
    auto count_abc(A self) -> int {
    }
    auto some_other(A self) -> float {
    }
}
```

And then only operators, constructors and destructors will live inside classes.

lib/battle/BattleHex.cpp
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko what do you think, is division by 2 will change inner logic or not? I think there is no change, but I am cautious.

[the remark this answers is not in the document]
IvanSavenko: I think this won't change anything.
However I don't understand:
- why this FIXME was added to begin with. It implies that this is not a trivial change
- what exactly this code is supposed to do?

I think this method needs some investigation - check how it used, check how it is *supposed* to work, check how it actually works, and finally - clarify this code.

I guess you could remove float conversion, but it won't improve readability here much. This is still an obscure code that should be looked into, either now or at some point later.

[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko agree. But I think than later is better, and you already wrote a reasons to @nullkiller .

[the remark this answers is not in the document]
rilian-la-te: changed to /2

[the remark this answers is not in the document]
nullkiller: /2 logically equivalent as getX() gives int and adding int to float will not change decimal part and we can have only 0 or 0.5 as decimal part so when casting to int it will be always ignored. Looks like this methods tries to calculate some distance between two points taking into account that the field is hexagonal.

lib/battle/BattleInfo.cpp
```diff
@@ -355,19 +358,19 @@ BattleInfo * BattleInfo::setupBattle(const int3 & tile, TerrainId terrain, const
 
 	for (auto position : config["commanderPositions"]["field"].Vector())
 	{
-		commanderField.push_back ((int)position.Float());
+		commanderField.push_back(static_cast<int>(position.Float()));
```
nullkiller, line 361: What is the difference between static_cast and (int) operator and why static_cast is better?

lib/battle/Unit.cpp
```diff
@@ -59,7 +59,7 @@ std::vector<BattleHex> Unit::getSurroundingHexes(BattleHex position, bool twoHex
 
 		if(side == BattleSide::ATTACKER)
 		{
-			for(BattleHex::EDir dir = BattleHex::EDir(0); dir <= BattleHex::EDir(4); dir = BattleHex::EDir(dir+1))
+			for(auto dir = static_cast<BattleHex::EDir>(0); dir <= static_cast<BattleHex::EDir>(4); dir = static_cast<BattleHex::EDir>(dir + 1))
```
nullkiller, line 62: Interesting why it is 4 now and 5 in different places. 5 I understand: 6 directions starting from 0, but 4?

nullkiller: It is because 2-hex unit so one hex it occupies and we need to skip it. Such logic is better to show explicitly by if(hex == otherHex) continue; as for me. But for sure not in scope of this PR).

nullkiller (approved):

lib/battle/BattleInfo.cpp
rilian-la-te: @nullkiller static_cast contain more compile checks AFAIK, but in this case C-style cast maybe harmless, but I just ran auto-changer for static_cast.

lib/battle/Unit.cpp
rilian-la-te: @nullkiller because it is for double-wide units. So, one direction is skipped

lib/battle/BattleInfo.cpp
IvanSavenko: @nullkiller No (functional) difference.
However C-style cast can do *any* conversion. All c++ casts can be also performed by c cast.

This harms both readability (it is not clear what code does at glance) and might lead to bugs, especially in case of pointer casting. Generally safe in case of arithmetic types though.

Another problem is that they are hard to find. For example, currently it is almost impossible to find all places where integer is casted to EDir enum. But if converted to static_cast this becomes much easier.

Because of this, C-style casts are generally viewed as not recommended in c++ code and should be replaced with one of 4 c++ casts, depending on context. Since they are hard to find, manual replacement is hard, but since we have tool to take care of that I think we should do it.

nullkiller: Thanks for explanation.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1308,34 +1309,34 @@ static bool isHexInFront(BattleHex hex, BattleHex testHex, BattleSide::Type side
...
+bool CBattleInfoCallback::isToReverse(const battle::Unit * attacker, const battle::Unit * defender) const
 {
 	BattleHex attackerHex = attacker->getPosition();
 	BattleHex defenderHex = defender->getPosition();
 
 	if (attackerHex < 0 ) //turret
 		return false;
 
-	if (isHexInFront(attackerHex, defenderHex, BattleSide::Type(attacker->unitSide())))
+	if(isHexInFront(attackerHex, defenderHex, static_cast<BattleSide::Type>(attacker->unitSide())))
```
IvanSavenko, line 1320: @rilian-la-te (no need to change anything here)

@nullkiller here is an example of why static cast is preferable.

Currently, method `attacker->unitSide()` returns uint8_t, however that value is always interpreted as BattleSide::Type. This is indication of poor code. I can search code for `static_cast<BattleSide::Type>` and find all places where this happens (other than `unitSide`) and fix them if necessary.

Yes, previously you could search for `BattleSide::Type`, but it would show every usage of that type - function parameters, variable declarations, member declarations. Finding actual conversion would require checking all of that

IvanSavenko (approved):

lib/battle/BattleHex.cpp
```diff
@@ -12,6 +12,9 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
+//Constexpr defined array with all directions used in battle
+static constexpr BattleHex::EDir battleDirections[] = {BattleHex::TOP_LEFT, BattleHex::TOP_RIGHT, BattleHex::RIGHT, BattleHex::BOTTOM_RIGHT, BattleHex::BOTTOM_LEFT, BattleHex::LEFT};
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko will rename. It is not a global-level, it is unit-level constexpr definition. As class member or static method it will be promoted to other files, and can lead to unsuccessful linking (I actually tried as private static constexpr member firstly, and failed a linking next. I think as static constexpr unit-only definition it is OK.

I will show you errors later.

lib/battle/CBattleInfoCallback.cpp
rilian-la-te: Okay) I will mark it as resolved, if you think than I do not need to do something, okay?

lib/battle/BattleHex.cpp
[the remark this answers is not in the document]
rilian-la-te: In general - constexpr members can lead to unsuccessful compilation, and it is better to make it private to only compilation unit (hence, static variable in .cpp file, for example).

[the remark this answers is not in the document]
IvanSavenko: > As class member or static method it will be promoted to other files, and can lead to unsuccessful linking

Static *member* probably should be *declared* in header but *defined* in header. AFAIK needed even for constexpr, with c++17 (or was it 20?) it is finally possible to mark such member is 'inline' and define in-place

Static *method* should not have these issues and should work either way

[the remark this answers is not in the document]
rilian-la-te: If something is declared constexpr, it should defined AND declared in one place, AFAIK (at least for c++14). Yes, we can do constexpr method, but what is more optimized?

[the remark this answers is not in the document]
IvanSavenko: Probably method then? Constexpr or not. Works on any system and such form already in use in our code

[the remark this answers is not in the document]
rilian-la-te: It should be constexpr to save us from memory burden. I can try later today.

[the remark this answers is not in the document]
rilian-la-te: done as constexpr method, but I do not know if it is optimized out or not.

PR #1583, opened 2023-02-15
fix regressions from modernization

IvanSavenko:
Well, I guess this is partially my fault since I missed this in review, but @rilian-la-te please be more cautious in future.

### DIFF
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -419,8 +419,8 @@ ArtifactPosition::ArtifactPosition(std::string slotName):
 	auto it = artifactPositionMap.find (slotName);
 	if (it != artifactPositionMap.end())
 		num = it->second;
-
-	logMod->warn("Warning! Artifact slot %s not recognized!", slotName);
+	else
+		logMod->warn("Warning! Artifact slot %s not recognized!", slotName);
 }
 
 void CArtHandler::addSlot(CArtifact * art, const std::string & slotID)
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -243,7 +243,7 @@ CGHeroInstance::CGHeroInstance():
 	movement(UNINITIALIZED_MOVEMENT),
 	portrait(UNINITIALIZED_PORTRAIT),
 	level(1),
-	exp(std::numeric_limits<TExpType>::max()),
+	exp(UNINITIALIZED_EXPERIENCE),
 	sex(std::numeric_limits<ui8>::max())
 {
 	setNodeType(HERO);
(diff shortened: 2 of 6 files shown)


rilian-la-te:
@IvanSavenko what regressions there was?

lib/CArtHandler.cpp
```diff
@@ -419,8 +419,8 @@ ArtifactPosition::ArtifactPosition(std::string slotName):
 	auto it = artifactPositionMap.find (slotName);
 	if (it != artifactPositionMap.end())
 		num = it->second;
-
-	logMod->warn("Warning! Artifact slot %s not recognized!", slotName);
+	else
```
IvanSavenko, line 422: @rilian-la-te 
After method was converted to constructor, code
` return it->second`
became
`num = it->second`

As result, method no longer returns & next line executes unconditionally

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -320,7 +320,7 @@ void CGHeroInstance::initHero(CRandomGenerator & rand)
 	if (patrol.patrolling)
 		patrol.initialPos = visitablePos();
 
-	if(exp == 0xffffffff)
+	if(exp == std::numeric_limits<TExpType>::max())
```
IvanSavenko, line 323: @rilian-la-te you changed value of constant `UNINITIALIZED_EXPERIENCE` (or what was its name) from `0xffffffff` to `std::numeric_limits<TExpType>::max()`

However, TExpType is uint64_t, meaning that this was change of behavior.

Good change, since otherwise we could end up with weird behavior if hero experience ends up exactly on `0xffffffff`. However code apparently had few other places where `0xffffffff` constant was used directly.

lib/CArtHandler.cpp
kambala-decapitator: also missed it :(

rilian-la-te: And I missed it too( Sorry(

lib/mapObjects/CGHeroInstance.cpp
kambala-decapitator: wouldn't it make more sense to use the named constant instead of numeric_limits directly?

rilian-la-te: There is no constant name AFAIK, number is used directly.

Maybe there is a reason to add such a constant, and name it exactly what you suggests?

rilian-la-te: @kambala-decapitator +1

IvanSavenko: Yeah, makes sense, will do.

PR #1595, opened 2023-02-18
VCAI: hard limit on iterations per turn

nullkiller:
Simple hack to workaround freezes in VCAI

### DIFF
--- a/AI/VCAI/VCAI.cpp
+++ b/AI/VCAI/VCAI.cpp
@@ -862,7 +862,7 @@ void VCAI::mainLoop()
 
 	invalidPathHeroes.clear();
 
-	while (basicGoals.size())
+	for (int pass = 0; pass< 30 && basicGoals.size(); pass++)
 	{
 		vstd::removeDuplicates(basicGoals); //TODO: container which does this automagically without has would be nice
 		goalsToAdd.clear();


dydzio0614:
Few questions:
Is size 30 enough? Won't it trigger in legit situations a lot when there is no infinite loop?
Maybe add log message if size exceeded? These bugs should be eventually fixed "proper way"?
Add comment that it is workaround that hides potential bugs?

nullkiller:
> Few questions: Is size 30 enough? Won't it trigger in legit situations a lot when there is no infinite loop? Wouldn't 1000 etc. be more appropriate? Maybe add log message if size exceeded as these bugs should be eventually fixed "proper way"? Add comment that it is workaround that hides potential bugs?

Who knows. NKAI has this feature already. And regarding logs I am not sure they are efficient enough. Nobody cares until it crash/hang. Sure log is needed. I will consider adding it maybe.
30 is enough. It also does not touch wandering so 30 is enough for sure. Big value will cause hanging which will irritate people.

DjWarmonger:
Good spot. There used to be a limit to goal decomposition depth, but it is not sufficient with overhauled architeture.,

IvanSavenko:
@nullkiller  Merge? Or is this waiting for something?

IvanSavenko:
@nullkiller another ping. Should this be merged or not?

PR #1602, opened 2023-02-22
force Conan to use v1

kambala-decapitator:
and a few other general fixes that don't fit anywhere


lib/VCMIDirs.cpp
```diff
@@ -15,6 +15,8 @@
 #include "iOS_utils.h"
 #endif
 
+#include <mutex>
```
IvanSavenko, line 18: Move to global.h perhaps?

IvanSavenko (approved):

lib/VCMIDirs.cpp
kambala-decapitator: grep shows that this include isn't used anywhere else. Should it still be moved?

IvanSavenko: For standard library includes - yeah, I think so

PR #1606, opened 2023-02-24
Update ukrainian.json

burunduk:
Fix tag
Spell check


IvanSavenko:
Thanks for the fixes, @burunduk. 
I should've checked that more carefully myself...

PR #1607, opened 2023-02-24
remove copy-pasted ioapi.c functions from FileStream.cpp

kambala-decapitator:
> copied from ioapi.c due to linker issues on MSVS

seems like an issue in the library header (function not exported) which I adjusted. In upstream declaration is still the same, but no bug reports about it...

anonymous namespace, besides sensible improvement, fixes single-process build on android.

Probably whole FileStream class can be moved to a separate library:
- it has no dependencies on VCMI lib
- in single process build both copies of VCMI lib will contain exactly the same thing => can be made a shared lib ("normal" build can have it static)

### DIFF
--- a/lib/filesystem/FileStream.cpp
+++ b/lib/filesystem/FileStream.cpp
@@ -20,6 +20,8 @@
 
 #include <cstdio>
 
+#define GETFILE static_cast<std::FILE*>(filePtr)
+
 #ifdef VCMI_WINDOWS
 	#ifndef _CRT_SECURE_NO_WARNINGS
 		#define _CRT_SECURE_NO_WARNINGS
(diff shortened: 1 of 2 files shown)


IvanSavenko (approved):
... and even if there are linker issue, they should be resolved other way than including third-party code into vcmi.

> in single process build both copies of VCMI lib will contain exactly the same thing 


Personally I would rather look into removing duplicated VCMI lib in single-app build.
I get that this is necessary right now to keep client & server state separate, but this is not a good solution.

kambala-decapitator:
> Personally I would rather look into removing duplicated VCMI lib in single-app build.

for sure, but that seems far from easy. Personally I don't even know how to do it (don't know architecture well enough).

IvanSavenko:
> for sure, but that seems far from easy. Personally I don't even know how to do it (don't know architecture well enough).

Yeah...
Strictly speaking there are "only" two things that need to be done (that I know of):
- Make IObjectInterface::cb non-static and set it per-instance (based on who creates object - server or client)
- Make sure that anything in VLC is constant during the game and extract any mutable parts from it.

Both are far from trivial though...

PR #1609, opened 2023-02-24
Battle order slight refactoring

krs0:
Slight refactoring for Battle Order.
Any suggestions where to put the newly created `BattlePhases `enum? It is common to both `CUnitState `and `CBattleInfoCallback`.

### DIFF
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -355,15 +355,6 @@ battle::Units CBattleInfoCallback::battleAliveUnits(ui8 side) const
 	});
 }
 
-enum BattlePhases
-{
-	SIEGE, // [0] - turrets/catapult,
-	NORMAL, // [1] - normal (unmoved) creatures, other war machines,
-	WAIT_MORALE, // [2] - waited creatures that had morale,
-	WAIT, // [3] - rest of waited creatures
-	MAX_NO_OF_PHASES // [4] - number of phases.
-};
-
 //T is battle::Unit descendant
 template <typename T>
 const T * takeOneUnit(std::vector<const T*> & allUnits, const int turn, int8_t & sideThatLastMoved, int phase)
@@ -387,7 +378,7 @@ const T * takeOneUnit(std::vector<const T*> & allUnits, const int turn, int8_t &
 
 		switch(phase)
 		{
-		case NORMAL: // Faster first, attacker priority, higher slot first
+		case battle::NORMAL: // Faster first, attacker priority, higher slot first
 			if(returnedUnit == nullptr || currentUnitInitiative > returnedUnitInitiative)
 			{
 				returnedUnit = currentUnit;
@@ -409,8 +400,8 @@ const T * takeOneUnit(std::vector<const T*> & allUnits, const int turn, int8_t &
 				}
 			}
 			break;
-		case WAIT_MORALE: // Slower first, higher slot first
-		case WAIT:
+		case battle::WAIT_MORALE: // Slower first, higher slot first
+		case battle::WAIT:
 			if(returnedUnit == nullptr || currentUnitInitiative < returnedUnitInitiative)
 			{
 				returnedUnit = currentUnit;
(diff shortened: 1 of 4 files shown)


lib/battle/CBattleInfoCallback.cpp
```diff
@@ -433,106 +445,105 @@ void CBattleInfoCallback::battleGetTurnOrder(std::vector<battle::Units> & out, c
...
-	//We'll split creatures with remaining movement to 4 buckets
+	// We'll split creatures with remaining movement to 4 buckets
 	// [0] - turrets/catapult,
 	// [1] - normal (unmoved) creatures, other war machines,
-	// [2] - waited cres that had morale,
-	// [3] - rest of waited cres
-	std::array<battle::Units, 4> phase;
+	// [2] - waited creatures that had morale,
+	// [3] - rest of waited creatures
+	std::array<battle::Units, 4> phases;
```
IvanSavenko, line 466:
```suggestion
	std::array<battle::Units, MAX_NO_OF_PHASES> phases;

```

IvanSavenko:
> Any suggestions where to put the newly created `BattlePhases `enum

I would actually place it into Unit.h and change 
`int battleQueuePhase(int turn) const`
into
`BattlePhases battleQueuePhase(int turn) const`

Also, please avoid enum's in with global view.
Bad: ` enum EnumType { A, B, C }` - A, B, C now have global visibility and may conflict with anything else in codebase
Okay: ` namespace EnumType { enum Type { A, B, C } }` - A, B, C now have visibility limited to EnumType
Great: ` enum class EnumType { A, B, C }` - A, B, C have visibility limited to EnumType and implicit conversions to/from integer types are reported by compiler.
(last option is not really suitable here though - would break iteration & std::array)

lib/battle/CBattleInfoCallback.cpp
krs0: Done

lib/battle/Unit.h
```diff
@@ -25,6 +25,16 @@ class JsonSerializeFormat;
...
 {
+
+enum BattlePhases
+{
+	SIEGE, // [0] - turrets/catapult,
+	NORMAL, // [1] - normal (unmoved) creatures, other war machines,
+	WAIT_MORALE, // [2] - waited creatures that had morale,
+	WAIT, // [3] - rest of waited creatures
+	MAX_NO_OF_PHASES // [4] - number of phases.
+};
```
IvanSavenko, line 36: Non-class enum declarations should be put into their own namespace to keep global namespace (or in this case - `battle` namespace) clean.
```suggestion
namespace BattlePhases
{
	enum Type
	{
		SIEGE, // [0] - turrets/catapult,
		NORMAL, // [1] - normal (unmoved) creatures, other war machines,
		WAIT_MORALE, // [2] - waited creatures that had morale,
		WAIT, // [3] - rest of waited creatures
		MAX_NO_OF_PHASES // [4] - number of phases.
	};
}
```

IvanSavenko: Yes, enum class is better but this enum is used for std::array indexing & for iteration. Enum class would need too many casts with this usage scenario IMO.

> also I'd assign values directly instead of keeping them inside comments
I'd rather say that values are unnecessary here & can be removed entirely

This enum is not mapped to H3 indexes or anything like that, so enum values can be added/removed safely, even if this changes their numeric values.

krs0: Now I understood what you wanted with namespace for enum. Thought having it in battle was enough.
Fixed as suggested and deleted enum values from comments.

krs0:
Finished for my side.

lib/battle/Unit.h
IvanSavenko: "Namespace enum" is something often used in VCMI code (was "invented" way before c++11) so kind of expected you to know about it.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -453,11 +455,11 @@ void CBattleInfoCallback::battleGetTurnOrder(std::vector<battle::Units> & turns,
...
 	// We'll split creatures with remaining movement to 4 buckets
-	// [0] SIEGE - turrets/catapult,
-	// [1] NORMAL - normal (unmoved) creatures, other war machines,
-	// [2] WAIT_MORALE - waited creatures that had morale,
-	// [3] WAIT - rest of waited creatures
-	std::array<battle::Units, battle::MAX_NO_OF_PHASES> phases; // Access using BattlePhases enum
+	// SIEGE - turrets/catapult,
+	// NORMAL - normal (unmoved) creatures, other war machines,
+	// WAIT_MORALE - waited creatures that had morale,
+	// WAIT - rest of waited creatures
```
kambala-decapitator, line 461: why is this comment copy-pasted here?

lib/battle/Unit.h
```diff
@@ -26,14 +26,17 @@ class JsonSerializeFormat;
...
-	WAIT, // [3] - rest of waited creatures
-	MAX_NO_OF_PHASES // [4] - number of phases.
-};
+	enum Type
+	{
+		SIEGE, // turrets/catapult,
+		NORMAL, // normal (unmoved) creatures, other war machines,
+		WAIT_MORALE, // waited creatures that had morale,
+		WAIT, // rest of waited creatures
+		NO_OF_PHASES // number of phases.
```
kambala-decapitator, line 37: `NO` is bad. `NUMBER_OF_PHASES ` or `PHASE_COUNT` would be better.

kambala-decapitator (changes requested):
second commit doesn't seem to add much value and can be squashed into the first one. Last commit can also be squashed into its parent.

lib/battle/Unit.h
krs0: Updated as per review.

lib/battle/CBattleInfoCallback.cpp
krs0: removed

kambala-decapitator (approved):

PR #1615, opened 2023-02-26
Optimization and cleanup CArtifactHolder

SoundSSGood:
Removed extra screen redraw operations. This is especially useful for bulk movements of arts. Also some code cleanup.

### DIFF
--- a/client/CPlayerInterface.h
+++ b/client/CPlayerInterface.h
@@ -131,8 +131,9 @@ public:
 	void artifactPut(const ArtifactLocation &al) override;
 	void artifactRemoved(const ArtifactLocation &al) override;
 	void artifactMoved(const ArtifactLocation &src, const ArtifactLocation &dst) override;
+	void bulkArtMovementStart(size_t numOfArts) override;
 	void artifactAssembled(const ArtifactLocation &al) override;
-	void artifactPossibleAssembling(const ArtifactLocation & dst) override;
+	void askToAssembleArtifact(const ArtifactLocation & dst) override;
 	void artifactDisassembled(const ArtifactLocation &al) override;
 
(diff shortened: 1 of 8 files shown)


client/CPlayerInterface.h
```diff
@@ -276,10 +277,10 @@ class CPlayerInterface : public CGameInterface, public IUpdateable
 
 	bool duringMovement;
 	bool ignoreEvents;
+	std::atomic<size_t> numOfMovedArts;
```
IvanSavenko, line 280: Can it actually be accessed from different threads? PlayerInterface always accessed while in lock (`pim`)

SoundSSGood: Just checked it works without atomic. Removed.

IvanSavenko (approved):

client/widgets/CArtifactHolder.cpp
```diff
@@ -686,7 +672,7 @@ void CArtifactsOfHero::realizeCurrentTransaction()
 								ArtifactLocation(commonInfo->dst.AOH->curHero, commonInfo->dst.slotID));
 }
 
-void CArtifactsOfHero::artifactMoved(const ArtifactLocation & src, const ArtifactLocation & dst)
+void CArtifactsOfHero::artifactMoved(const ArtifactLocation & src, const ArtifactLocation & dst, bool withUpdate)
```
nullkiller, line 675: withUIUpdate or withRedraw

SoundSSGood: > withUIUpdate or withRedraw

in function uses updateParentWindow() rather than safeRedraw(). withUIUpdate is better I think.

PR #1616, opened 2023-02-26
move Android code to the main repo

kambala-decapitator:
Android code copied from https://github.com/vcmi/vcmi-android, that repo shouldn't be needed any more after merging this PR.

- currently Android has issues with `dlopen`ing of AI libs (runtime crashes during AI turn), so I kept the existing approach to link AI libs statically into the VCMI lib. Although the original approach was to collect all AI sources and add them to VCMI lib target (3849c37ed83d6a59d08e1af5140ab9bacd732d95), now at runtime with "multiprocess" build (that uses shared VCMI lib) there's crash due to uncaught Boost ASIO exception on closing server connection which is quite strange as we do handle Boost exceptions; changing that to using static AI targets and linking to them solved that. Static AI is a cmake option available to all platforms.
- some fixes to Android launcher strings and improvements to Java code
- all dependencies are built with Conan
- debug and daily builds use different package IDs to be able to install them alongside the Play Store version

Stuff to solve in the future:
- build/submit to Play Store from CI, requires using single package (AAB) with native libs for all archs inside
- implement external storage to be able to consume game data from other places on filesystem / share it between debug/daily/store builds
- replacing Java launcher with the Qt one


AI/CMakeLists.txt
```diff
@@ -45,9 +48,11 @@ endif()
 #######################################
 
 add_subdirectory(BattleAI)
-add_subdirectory(StupidAI)
-add_subdirectory(EmptyAI)
 add_subdirectory(VCAI)
+if(NOT ENABLE_STATIC_AI_LIBS)
+	add_subdirectory(StupidAI)
+	add_subdirectory(EmptyAI)
+endif()
```
IvanSavenko, line 55: Why not allow this libraries for Android as well?

AI/Nullkiller/CMakeLists.txt
```diff
@@ -120,24 +117,24 @@ set(Nullkiller_HEADERS
 		AIGateway.h
 )
 
+if(NOT ENABLE_STATIC_AI_LIBS)
+	list(APPEND Nullkiller_SRCS main.cpp StdInc.cpp)
+endif()
```
IvanSavenko, line 122: Maybe ifdef in code would be better?

CI/android/signing.properties
```diff
@@ -0,0 +1,4 @@
+STORE_FILE=vcmi-travis.jks
```
IvanSavenko, line 1: Is this some sort of private signing key or just some placeholder?

CMakeLists.txt
```diff
@@ -535,14 +561,48 @@ endif()
...
+	install(DIRECTORY Mods DESTINATION ${DATA_DIR})
+endif()
 if(ENABLE_LUA)
 	install(DIRECTORY scripts DESTINATION ${DATA_DIR})
 endif()
 
-# that script is useless for Windows and iOS
-if(NOT WIN32 AND NOT APPLE_IOS)
+# that script is useless for Windows / iOS / Android
+if(NOT WIN32 AND NOT APPLE_IOS AND NOT ANDROID)
```
IvanSavenko, line 605: Maybe check for Unix/Linux instead?
```suggestion
if(UNIX)
```

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Tue Sep 27 21:00:27 EEST 2022
+distributionBase=GRADLE_USER_HOME
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionPath=wrapper/dists
+zipStorePath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
```
IvanSavenko, line 6: Windows line endings?

client/CMT.cpp
```diff
@@ -580,7 +581,7 @@ void playIntro()
 	}
 }
 
-#ifndef VCMI_IOS
+#if !defined(VCMI_IOS) && !defined(VCMI_ANDROID)
```
IvanSavenko, line 584: Maybe we should add something like VCMI_MOBILE / VCMI_PHONE or something that will be defined for both ios and adroid? Code is full of "if android or ios" checks right now.

IvanSavenko (commented):
Just suggestions, feel free to ignore

AI/CMakeLists.txt
kambala-decapitator: because static AI approach hardcodes selecting battle AI, see `CGameInterface.cpp`

AI/Nullkiller/CMakeLists.txt
kambala-decapitator: don't see any advantages, could you explain?

CI/android/signing.properties
kambala-decapitator: it's a real file which is public in the android repo. @nullkiller do you have more info?

CMakeLists.txt
kambala-decapitator: iOS and Android are definitely Unix :) it basically applies to "desktops that are not Windows", I'd keep as is.

android/gradle/wrapper/gradle-wrapper.properties
kambala-decapitator: copied it from the repo as is. Building from Windows is also possible. Can change, probably Gradle doesn't care about line endings.

client/CMT.cpp
kambala-decapitator: indeed, we have a lot of `defined(VCMI_IOS) || defined(VCMI_ANDROID)` and inverted conditions. Will add `VCMI_MOBILE`.

CI/android/signing.properties
nullkiller: This is a keystore, it is password protected. And not the one used to sign releases

nullkiller: I think it is used to sign dev builds from CI. Password is stored in github secret

nullkiller: Interesting now how it works, are our new builds signed?

lib/VCMIDirs.cpp
```diff
@@ -530,7 +527,63 @@ bfs::path VCMIDirsOSX::libraryPath() const { return "."; }
...
 #endif // VCMI_IOS, VCMI_MAC
 #elif defined(VCMI_XDG)
-class VCMIDirsXDG : public IVCMIDirsUNIX
+class VCMIDirsELF : public IVCMIDirsUNIX
+{
+public:
+	std::string libraryName(const std::string & basename) const override { return "lib" + basename + ".so"; }
+};
+
+// android is treated differently to XDG
```
kambala-decapitator, line 536: can you come up with a better name by chance? Didn't want to use "Linux" because BSDs also use XDG.

CI/android/signing.properties
kambala-decapitator: daily builds are signed as before, the process is preserved

nullkiller: Interesting how it extracts the secret

nullkiller: Or maybe it is not protected properly, I never investigated it precisely.

nullkiller: Looks like the password is hardcoded somewhere in CI/gradle scripts or there is none specified at all)

nullkiller: Exactly, in this file. Not sure if we need to do anything about it so.

AI/CMakeLists.txt
IvanSavenko: Why not add StupidAI & EmptyAI in hardcoded list in CGameInterface.cpp?

kambala-decapitator: this would require adjusting the java launcher as well. I wouldn't put effort into the java launcher at all (unless strongly required) as porting Qt launcher is more important.

feel free to open such issue / work on it if you wish :)

lib/VCMIDirs.cpp
IvanSavenko: Personally I would rather get rid of "VCMIDirsELF" class and copy libraryName method as it into both VCMIDirsAndroid and VCMIDirsXDG 

Yes, this would duplicate 1 line of 1 method, but IMO more clear - this file is already hard to understand due to countless ifdef's and inheritance chains

kambala-decapitator: ok, will do. In the future we can split this file into `VCMIDirs_Windows.cpp` etc.

android/gradle/wrapper/gradle-wrapper.properties
IvanSavenko: Probably better to change. It is git who normally handles line endings, not developer (autoCLRF option or what its name).

AI/CMakeLists.txt
IvanSavenko: It won't *require* adjusting java launcher. We can change CGameInterface but keep only NKAI & VCAI available to user.

Mostly want to avoid potential issues in future, e.g. if Android gets Qt Launcher where StupidAI can be selected and we forget about non-functioning StupidAI in Android.

Anyway, not insisting, feel free to ignore.

AI/Nullkiller/CMakeLists.txt
IvanSavenko: Not really, other than simplifying CMakeLists a bit.
Feel free to ignore.

AI/CMakeLists.txt
kambala-decapitator: oh, I got the point, makes sense! User will also be able to change AI by editing the json settings.

kambala-decapitator:
should we enable single process for android by default btw?

kambala-decapitator:
unfortunately, building on Windows doesn't work out of the box, needs investigation.

rilian-la-te:
@kambala-decapitator why we should build android on Windows?

kambala-decapitator:
updated PR description

rilian-la-te (commented):
Is .idea folder required for building android?

kambala-decapitator:
> Is .idea folder required for building android?

it's not, but android studio saves some useful stuff there as I understand

rilian-la-te:
@kambala-decapitator but may kill .idea folder altogether then? I do not think than idea project will be actively maintained.

kambala-decapitator:
> @kambala-decapitator but may kill .idea folder altogether then? I do not think than idea project will be actively maintained.

at least @nullkiller used (uses?) it and it is present in the android repo

android/.idea/.name
```diff
@@ -0,0 +1 @@
+VCMI
```
nullkiller, line 1: Do we need these idea files?

android/vcmi-app/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,53 @@
...
+            android:name=".VcmiSDLActivity"
+            android:configChanges="keyboardHidden|orientation|screenSize"
+            android:label="@string/app_name"
+            android:launchMode="singleTop"
+            android:screenOrientation="sensorLandscape"
+            android:theme="@style/Theme.VCMI.Full" />
+
+        <service
+            android:name=".ServerService"
+            android:process="eu.vcmi.vcmi.srv"
```
nullkiller, line 48: Do we still have a server?

android/.idea/.name
rilian-la-te: @nullkiller maybe kill all .idea folder?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/ActivityAbout.java
```diff
@@ -0,0 +1,105 @@
...
+        final DialogAuthors dialogAuthors = new DialogAuthors();
+        dialogAuthors.show(getSupportFragmentManager(), DIALOG_AUTHORS_TAG);
+    }
+
+    private void onBtnLibsPressed(final View v)
+    {
+        // TODO 3rd party libs view (dialog?)
+    }
+
+    private void onBtnPrivacyPressed(final View v)
```
nullkiller, line 84: This method is it used anywhere?

android/.idea/.name
dydzio0614: this folder should not be versioned, but be wary to backup this folder when pulling changes after this PR gets merged, so your settings do not get eaten... same goes for clion c++ devs

android/vcmi-app/src/main/java/eu/vcmi/vcmi/NativeMethods.java
```diff
@@ -0,0 +1,161 @@
...
+    @SuppressWarnings(Const.JNI_METHOD_SUPPRESS)
+    public static String nativePath()
+    {
+        final Context ctx = SDL.getContext();
+        Log.i("Accessing ndk path: " + ctx.getApplicationInfo().nativeLibraryDir);
+        return ctx.getApplicationInfo().nativeLibraryDir;
+    }
+
+    @SuppressWarnings(Const.JNI_METHOD_SUPPRESS)
+    public static void startServer()
```
nullkiller, line 75: Do we still use this? from C++?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/NativeMethods.java
```diff
@@ -0,0 +1,161 @@
...
+        Intent intent = new Intent(ctx, SDLActivity.class);
+        intent.setAction(VcmiSDLActivity.NATIVE_ACTION_CREATE_SERVER);
+        // I probably do something incorrectly, but sending new intent to the activity "normally" breaks SDL events handling (probably detaches jnienv?)
+        // so instead let's call onNewIntent directly, as out context SHOULD be SDLActivity anyway
+        ((VcmiSDLActivity) ctx).hackCallNewIntentDirectly(intent);
+//        ctx.startActivity(intent);
+    }
+
+    @SuppressWarnings(Const.JNI_METHOD_SUPPRESS)
+    public static void killServer()
```
nullkiller, line 95: And this

android/.idea/.name
rilian-la-te: @dydzio0614 agree.

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/LegacyConfigReader.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.ObjectInputStream;
+
+import eu.vcmi.vcmi.Const;
+
+/**
+ * helper used to retrieve old vcmi config (currently only needed for h3 data path in order to migrate the data to the new location)
+ *
+ * @author F
+ */
+public final class LegacyConfigReader
```
nullkiller, line 18: First time I see this code. Probably something ancient?

nullkiller (approved):

android/.idea/.name
nullkiller: It was present historically, so lets remove.

nullkiller:
> should we enable single process for android by default btw?

Yes, I believe so. There is a bug now that server process can hang and this forces user to unload app from memory. With single-process this should no longer be an issue

android/.idea/.name
kambala-decapitator: ok, will remove. Anyone can still pull it from the old Android repo.

android/vcmi-app/src/main/AndroidManifest.xml
kambala-decapitator: Andrioid can be built as single and multi process, so yes. Single- doesn't use it of course.

kambala-decapitator:
> > should we enable single process for android by default btw?
> 
> Yes, I believe so. There is a bug now that server process can hang and this forces user to unload app from memory. With single-process this should no longer be an issue

ok, will set it as default

android/vcmi-app/src/main/java/eu/vcmi/vcmi/ActivityAbout.java
kambala-decapitator: not sure, do you understand from grep results?

```
> ag --literal btnprivacy

vcmi-app/build/generated/data_binding_base_class_source_out/debug/out/eu/vcmi/vcmi/databinding/ActivityAboutBinding.jav
29:  public final AppCompatTextView aboutBtnPrivacy;
48:      @NonNull AppCompatTextView aboutBtnPrivacy, @NonNull AppCompatTextView aboutLinkPortal,
55:    this.aboutBtnPrivacy = aboutBtnPrivacy;
103:      AppCompatTextView aboutBtnPrivacy = ViewBindings.findChildViewById(rootView, id);
104:      if (aboutBtnPrivacy == null) {
139:          aboutBtnPrivacy, aboutLinkPortal, aboutLinkRepoLauncher, aboutLinkRepoMain,

vcmi-app/src/main/java/eu/vcmi/vcmi/ActivityAbout.java
84:    private void onBtnPrivacyPressed(final View v)
```

kambala-decapitator: oh, it and `onBtnLibsPressed` have empty implementation with TODO, I guess can be removed

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/LegacyConfigReader.java
kambala-decapitator: probably. it's no longer used.

kambala-decapitator:
wrote build guide: https://wiki.vcmi.eu/How_to_build_VCMI_(Android)

kambala-decapitator:
anything left for merge?

nullkiller:
No objection to merge it from my side

kambala-decapitator:
please mark https://github.com/vcmi/vcmi-android as readonly

PR #1620, opened 2023-02-28
Fix battle ai trying to act with killed stack

nullkiller:
There is a crash dump showing that somehow AI managed to kill own stack with spellcast. Theoretically it is prohibited to injure own stacks so this should not happen but it did somehow. So kind of blind fix.

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -117,9 +117,9 @@ BattleAction CBattleAI::activeStack( const CStack * stack )
 
 		attemptCastingSpell();
 
-		if(auto ret = cb->battleIsFinished())
+		if(cb->battleIsFinished() || !stack->alive())
 		{
-			//spellcast may finish battle
+			//spellcast may finish battle or kill active stack
 			//send special preudo-action
 			BattleAction cancel;
 			cancel.actionType = EActionType::CANCEL;


kambala-decapitator:
> kill own stack with spellcast

Sacrifice and Armageddon can do that for example

IvanSavenko:
It is also possible for Magic Mirror to trigger, reflecting your spell

nullkiller:
Heroes did not have chain lightning, Armageddon or magic mirror. And AI has rule to not use such spells when they can injure own units. So the situation itself looks weird but I do not want to investigate it. There is no save file. After changing a bit manually AI logic to allow Armageddon I can reproduce this issue and I can confirm that it is fixed. But how it occurs with unmodified AI code I do not know)

nullkiller:
Looks like it was HYPNOTIZE + METEOR SHOWER. So active stack was actually not ours and this is why AI allowed to kill it.

nullkiller:
[armagedon-1.zip](https://github.com/vcmi/vcmi/files/10854782/armagedon-1.zip)
Reproduced. Wait next day, when attacked use autofight.
AI casts hypnotize on zealots, pegassi is killed and as soon as next turn happens AI casts armageddon and first creature in the next turn is zealot. It is killed by spell and game crashes

Issue #1619, opened 2023-02-28
CArchiveLoader::CArchiveLoader param "extractArchives" is masked by typo


nickhuang99:
Parameter "extractArchives" can NEVER take effect
<pre><code>
CArchiveLoader::CArchiveLoader(std::string mountPoint, bfs::path archive, bool extractArchives = false);
</code></pre>
because the member variable is initialized by itself instead of input parameter.

Currently there is only one used occasion for "extractArchives" which is default value. So, it is now harmless unless in future it is actually playing a role.
A simple patch shows why it is a typo because the "input" parameter is never used to init member CArchiveLoader::extractArchives. 

<pre><code>
diff --git a/lib/filesystem/CArchiveLoader.cpp b/lib/filesystem/CArchiveLoader.cpp
index b58fe2b02..ad1d59a9b 100644
--- a/lib/filesystem/CArchiveLoader.cpp
+++ b/lib/filesystem/CArchiveLoader.cpp
@@ -24,10 +24,10 @@ ArchiveEntry::ArchiveEntry()
 
 }
 
-CArchiveLoader::CArchiveLoader(std::string _mountPoint, bfs::path _archive, bool extractArchives) :
+CArchiveLoader::CArchiveLoader(std::string _mountPoint, bfs::path _archive, bool _extractArchives) :
     archive(std::move(_archive)),
     mountPoint(std::move(_mountPoint)),
-       extractArchives(extractArchives)
+       extractArchives(_extractArchives)
 {
        // Open archive file(.snd, .vid, .lod)
        CFileInputStream fileStream(archive);
<code></pre>

krs0:
Having same parameter name as member variable is legal (at least on windows and linux). Perhaps a matter of taste?
Nevertheless I will change it, to conform with the rest of constructor parameter names.

IvanSavenko:
It is legal, and sometimes used in vcmi code. So I am OK with diff in bug report (but not insisting on it)

nickhuang99:
it is legal, just usually it is supposedly initiated by input parameter. otherwise there is no use of parameter.

Sent via the Samsung Galaxy S22 5G, an AT&T 5G smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

krs0:
Changed in https://github.com/vcmi/vcmi/pull/1622

krs0:
[...]

PR #1622, opened 2023-02-28
CArchiveLoader param "extractArchives" is masked by typo #1619

krs0:


lib/filesystem/CArchiveLoader.cpp
```diff
@@ -24,10 +24,10 @@ ArchiveEntry::ArchiveEntry()
 
 }
 
-CArchiveLoader::CArchiveLoader(std::string _mountPoint, bfs::path _archive, bool extractArchives) :
+CArchiveLoader::CArchiveLoader(std::string _mountPoint, bfs::path _archive, bool _extractArchives) :
     archive(std::move(_archive)),
     mountPoint(std::move(_mountPoint)),
-	extractArchives(extractArchives)
+	extractArchives(_extractArchives)
```
nullkiller, line 30: It is not a typo but valid c++ construct. First name is looked up as a class member and second as a parameter. we often use this feature. Or am I missing something?

nullkiller: Looks like we do not have unified vision on what way is better)

nullkiller: And our coding guidelines say nothing on this matter unfortunately.

IvanSavenko: Huh, I though that initial report was about typo - self-initialization instead of using constructor parameter.

I'd say that change *is* an improvement, if only for consistency (method accepts 3 parameters, but only 2 start with underscore).

As for code style - I think that using same name for constructor parameters is perfectly fine. We should probably add this info into guidelines

krs0: Now is consistent. Case closed :)

PR #1623, opened 2023-03-01
Make tests working

rilian-la-te:
Unit tests now works and passes AFAIK.

There are some changes in non-testing areas, please, check.


rilian-la-te:
Will look for running without H3 data now.

test/entity/CArtifactTest.cpp
```diff
@@ -36,7 +36,10 @@ TEST_F(CArtifactTest, RegistersIcons)
 	subject->image = "Test1";
 	subject->large = "Test2";
 
-	auto cb = std::bind(&CArtifactTest::registarCb, this, _1, _2, _3, _4);
+	auto cb = [this](auto && PH1, auto && PH2, auto && PH3, auto && PH4)
```
dydzio0614, line 39: I can sense automatic conversion into lambda, which is crap and pretty much always adds needless && and std::forward<decltype> for me

IvanSavenko: But with these changes lambda will run 1 nanosecond faster!

rilian-la-te: This auto conversion was one conversion which I actually ran. It is not for modernization, it just crashes there and was made to easier debugging.

rilian-la-te: Yes, but in command line (in gdb too)

rilian-la-te:
Merged via #2070

Issue #857, opened 2022-09-07
[feature] show possible amount of creatures that would perish


kambala-decapitator:
Currently only possible damage is shown when you hover mouse over enemy creature. We could also show how many creatures would die from the hit to remove the calculation burden from the user.

For example:

> 150-200 damage (10-12 creatures/CREATURE_NAME will perish)

krs0:
I propose to implement it all the way H5 style:
Show in status line, all info of interest in this format:
<min>-<max> damage would be dealt (<min>-<max> <creature_name> would perish). Will retaliate! | No retaliation.

dydzio0614:
sounds like good improvement that even hota doesn't have

PR #1626, opened 2023-03-02
New settings window ux improvements

IvanSavenko:
Attempts to improve design of new settings window.

UI design itself is done, remaining todo's:
- [x] update localization
- [x] clean up old files from vcmi mod
- [x] Split "Show available creatures in town" into two radio buttons
- [x] make battle stack queue update immediately on option change
- [x] Wait for PR #1581 to be merged & update hero/map speed settings
- [x] Implement new options available in #1581

vcmi-settings-general

vcmi-settings-adventure

vcmi-settings-battle


IvanSavenko:
Progress so far, still WIP:

vcmi-settings-battle

Mods/vcmi/config/vcmi/english.json
```diff
@@ -50,13 +50,12 @@
 	"vcmi.adventureOptions.forceMovementInfo.help" : "{Always show move info in status bar}\n\n Replaces default status bar info with movement points data without need to hold ALT button.",
 	"vcmi.adventureOptions.showGrid.hover" : "Show grid",
 	"vcmi.adventureOptions.showGrid.help" : "{Show grid}\n\n Shows grid overlay, showing borders between adventure map tiles.",
-	"vcmi.adventureOptions.mapScrollSpeed4.hover": "4",
 	"vcmi.adventureOptions.mapScrollSpeed4.help": "Set map scrolling speed to very fast",
-	"vcmi.adventureOptions.mapScrollSpeed5.hover": "5",
 	"vcmi.adventureOptions.mapScrollSpeed5.help": "Set map scrolling speed to extremely fast",
+	"vcmi.adventureOptions.mapScrollSpeed5.help": "Set map scrolling speed to fast, very fast.",
```
dydzio0614, line 55: duplicated key

IvanSavenko: FIxed, will commit later.
Not bothering with localization at the moment since I want to review these settings anyway - e.g. shift all texts and add "very slow" as 1st option instead of coming up with more synonyms for "fast"

IvanSavenko:
More progress. Any suggestions / comments?

vcmi-settings-general

dydzio0614:
for queue size "off" the hint about taking effect on next battle doesnt apply

and maybe it would be god idea to add such hint for resolution change, at least until it requires restart

SoundSSGood:
Why were the original audio controls replaced?

IvanSavenko:
@SoundSSGood 
1) Better word would be "not recreated". H3 audio controls are pre-rendered on image background, meaning that for new background we need to draw something similar
2) Personal opinion, but those controls were too rigid. For example, I usually have very low volume (~10-20%), so for me settings went like "quiet" "loud" "extremely loud", with nothing in-between since you can't set it to (for example) 15%. Yes, in vcmi this was somewhat fixed, but this was a hidden feature and not immediately obvious.
3) While I initially had concerns that a slider would look worse than H3 selector, current version (IMO) looks at the very least presentable.

IvanSavenko:
Design done, see PR description for final screenshots.

Marking as ready, still few minor tasks to do, but nothing critical.

dydzio0614:
1. A bit low amount of spots for future battle settings - for example https://github.com/vcmi/vcmi/issues/857 (this particular feature might better be left always toggled on like threat level though)

2. Actual sections borders feel too thin / subtle (for me and Andruids)

IvanSavenko:
@dydzio0614 

1. Currently battle tab still has 4 free "slots", so not a problem at the moment.
Plus, a lot of options are not implemented and have... questionable needs. 
For example auto-combat options can get shared "War Machines" entry instead of 3 separate entries, freeing 2 slots in the process.
Creature info block can be converted into set of buttons similar to battle queue, allowing to reduce its size from 3 to 2 "slots".
And if every other option is exhausted we can increase height of the window - currently can be done purely in json, even without image editor. 

2. Yeah, I think you're right... They are definitely more thin visually, but I think they are more pronounced so looks okay-ish. Also since I am using overlay image without custom background I am rather limited in possible effects - all I can use is RGB + transparency.
So probably will keep as it for now? Will keep in mind for future though.

client/windows/settings/SettingsMainWindow.cpp
```diff
@@ -57,7 +56,7 @@ SettingsMainWindow::SettingsMainWindow(BattleInterface * parentBattleUi) : Inter
 	std::shared_ptr<CButton> restartButton = widget<CButton>("restartButton");
 	assert(restartButton);
 
-	if(CSH->isGuest())
+	if(CSH->isGuest() || parentBattleUi)
 	{
 		loadButton->block(true);
```
dydzio0614, line 61: loading game during battle works, no reason to block it - it is wanted feature

krs0:
Can you rename Queue to Battle Queue?

dydzio0614:
> Can you rename Queue to Battle Queue?

Why tho? it is part of "battle settings" already :P

IvanSavenko:
> Can you rename Queue to Battle Queue?

Maybe Turn Order Queue? "Battle" seems to be redundant since this is battle tab, but "turn order" part might not.

krs0:
Show available creatures in town -> 2 Radio buttons:
  Show number of creatures available for hire
  Show weekly growth of creatures

krs0:
> > Can you rename Queue to Battle Queue?
> 
> Maybe Turn Order Queue? "Battle" seems to be redundant since this is battle tab, but "turn order" part might not.

When I talk about Turn I usually talk about Day. Battle Queue seems concise and explicit enough.

krs0:
Speed settings seem excessive. 1 less wouldn't be enough? That means  5 settings instead of 6?

dydzio0614:
> Speed settings seem excessive. 1 less wouldn't be enough? That means 5 settings instead of 6?

won't be enough unless we make 5th as fast as 6th and make 4th something like 4.5...

krs0:
I know Show move path was ok 20 years ago.... but now technology advanced :)
image

dydzio0614:
what if I tell you this setting is taken from original game's files?

krs0:
> en from



> what if I tell you this setting is taken from original game's files?

I know that, that's why the joke :) but now we have more space so that can be changed
Also while we are at it: Show Hero Reminder -> Show Hero Movement Reminder

krs0:
Numeric Creatures Quantities. -> Numeric Creature Quantities sounds more natural.

IvanSavenko:
> When I talk about Turn I usually talk about Day. Battle Queue seems concise and explicit enough. Combat usually proceeds in rounds. .... But then there is the term creatures turn :)

"Battle Queue" does not explains what this is about. "Creature battle turn order queue" is much more descriptive but too long.
"Show Creature Turn Order" maybe? This is from battle window so no need to explain that this is battle related.

> Speed settings seem excessive. 1 less wouldn't be enough? That means 5 settings instead of 6?

Maybe. But MP is played at extremely high speeds. Fastest options are basically for them.

> Show available creatures in town -> 2 Radio buttons:
Show number of creatures available for hire
Show weekly growth of creatures

H3 does not have "radio button" element. But making two checkboxes from which only one can be checked should be OK.

> Numeric Creatures Quantities. -> Numeric Creature Quantities sounds more natural.

Both are not perfect, but OK, will change.

client/windows/settings/SettingsMainWindow.cpp
IvanSavenko: Enabled loading, kept restart & save locked

krs0:
Again on Trun order Bar... it has 4 different names now depending on the place... lets nail it once and for all :)
"Show Creature Turn Order" its mainly creatures trurn order but not only creatures are involved. Warmachines, towers, comanders, maybe heroes. And is not Turn order since it spans for whole combat.
so.... Combat Turn Order Queue? Combat Turn order Bar? Combat Order Queue, Combat Order Bar? I saw also a weird one... Action order Bar?
I vote for Combat Order Queue

IvanSavenko:
Presuming finished & ready for merge.

@krs0 on, ahem, queue-like-thing-in-battle - for now leaving it as it. 
If you wish - feel free to ask people as to how this option should be named, e.g. on Discord. If there will be some sort of consensus on how to call it we can change it in options, I don't mind.

I don't mind using name that has words "battle" "stack" or "queue", but H3 never uses term "stack" (always creature, for example in casualties from spells). 
Also, I am not native English speaker, so I may be wrong, but at the very least direct translations of "Battle queue" or "Stack queue" or even "Combat Order Queue" are completely nonsensical in Ukrainian.

So for now to avoid spending more time on this non-critical issue I am leaving it as it. If there will be some clear way to call it - I'll gladly rename it.

PR #1629, opened 2023-03-04
vcmi: allow loads bonuses info from mods

rilian-la-te:
It is useful for making custom bonus graphics into mods, now mods just replace game config, which is not good.

Please, backport into stable if possible.

@IvanSavenko


rilian-la-te:
@misiokles @DjWarmonger

PR #1635, opened 2023-03-05
warmachine buying fix

SoundSSGood:
Fixed buying a warmachine in the blacksmith. Now the buy button is not activated if the slot is occupied by another warmachine, like a cannon.

Not sure if the class hierarchy is valid. looks like possibleSlots should be available without dynamic_cast

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -804,10 +804,25 @@ void CCastleBuildings::enterBlacksmith(ArtifactID artifactID)
 		LOCPLINT->showInfoDialog(boost::str(boost::format(CGI->generaltexth->allTexts[273]) % town->town->buildings.find(BuildingID::BLACKSMITH)->second->getNameTranslated()));
 		return;
 	}
-	auto art = artifactID.toArtifact(CGI->artifacts());
+	auto art = dynamic_cast<const CArtifact*>(artifactID.toArtifact(CGI->artifacts()));
 
 	int price = art->getPrice();
-	bool possible = LOCPLINT->cb->getResourceAmount(Res::GOLD) >= price && !hero->hasArt(artifactID);
+	bool possible = LOCPLINT->cb->getResourceAmount(Res::GOLD) >= price;
+	if(possible)
+	{
+		for(auto slot : art->possibleSlots.at(ArtBearer::HERO))
+		{
+			if(hero->getArt(slot) == nullptr)
+			{
+				possible = true;
+				break;
+			}
+			else
+			{
+				possible = false;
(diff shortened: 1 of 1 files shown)


client/windows/CCastleInterface.cpp
```diff
@@ -804,10 +804,23 @@ void CCastleBuildings::enterBlacksmith(ArtifactID artifactID)
 		LOCPLINT->showInfoDialog(boost::str(boost::format(CGI->generaltexth->allTexts[273]) % town->town->buildings.find(BuildingID::BLACKSMITH)->second->getNameTranslated()));
 		return;
 	}
-	auto art = artifactID.toArtifact(CGI->artifacts());
+	auto art = dynamic_cast<const CArtifact*>(artifactID.toArtifact(CGI->artifacts()));
 
 	int price = art->getPrice();
-	bool possible = LOCPLINT->cb->getResourceAmount(Res::GOLD) >= price && !hero->hasArt(artifactID);
+	bool possible = LOCPLINT->cb->getResourceAmount(Res::GOLD) >= price;
+	if(possible)
```
IvanSavenko, line 811: Please add `{}` braces around this if. Yes, it contains single statement but this block is 10 lines long

IvanSavenko (approved):

client/windows/CCastleInterface.cpp
```diff
@@ -804,10 +804,25 @@ void CCastleBuildings::enterBlacksmith(ArtifactID artifactID)
...
 	}
-	auto art = artifactID.toArtifact(CGI->artifacts());
+	auto art = dynamic_cast<const CArtifact*>(artifactID.toArtifact(CGI->artifacts()));
 
 	int price = art->getPrice();
-	bool possible = LOCPLINT->cb->getResourceAmount(Res::GOLD) >= price && !hero->hasArt(artifactID);
+	bool possible = LOCPLINT->cb->getResourceAmount(Res::GOLD) >= price;
+	if(possible)
+	{
+		for(auto slot : art->possibleSlots.at(ArtBearer::HERO))
```
Nordsoft91, line 813: what if possibleSlots.at will return empty container?

rilian-la-te: @Nordsoft91 is this not handled in range-based for specification? I think it possibleSlots.begin() == possibleSlots.end() cycle will not run at all.

SoundSSGood: > what if possibleSlots.at will return empty container?

If the artifact has no possible slots at all, then it can only be placed in a backpack. In this case, the purchase will be allowed.

PR #1639, opened 2023-03-05
Fix loading of non-ASCII text files

IvanSavenko:
Fixes loading on localized H3 copies.
@rilian-la-te please recheck fix.


rilian-la-te:
Looks okay: 
изображение

PR #1640, opened 2023-03-05
vcmi: use qt6 for clang linux build

rilian-la-te:
Test building for both Qt6 and Qt5 in our CI.


rilian-la-te:
@IvanSavenko @kambala-decapitator please, merge after CI passes, it will allow us to test Qt6 builds.

PR #1643, opened 2023-03-05
Fix lldb string inspection for debug build

Adriankhl:
I was trying to use CodeLLDB of vscode to debug something, but I couldn't see any `std::string` value because I don't have the debug version of the standard library on my system: https://stackoverflow.com/questions/58578615/cannot-inspect-a-stdstring-variable-in-lldb

Adding the flag `-fstandalone-debug` solved the problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -300,7 +300,7 @@ if(ANDROID)
 	list(APPEND SYSTEM_LIBS log)
 endif()
 
-if(CMAKE_COMPILER_IS_GNUCXX OR NOT WIN32)
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR NOT WIN32)
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wpointer-arith")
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
will everyone benefit from this flag? As you could've passed it via [CXXFLAGS env var](https://cmake.org/cmake/help/latest/envvar/CXXFLAGS.html)

kambala-decapitator:
I think current approach wouldn't work with multiconfig generators. Maybe https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_DEBUG.html would be a better solution?

Adriankhl:
Ok, I have changed it to use `CMAKE_CXX_FLAGS_DEBUG` and it works for me. I don't use multiconfig generators (I believe it is only for Xcode in this case?) so I am not sure if this works for that.

kambala-decapitator (approved):
> I believe it is only for Xcode in this case?

there's also Ninja multi-config

rilian-la-te:
And visual studio for hardcore Windows guys(

Adriankhl:
To be fair, I used to insist on using either `mingw` or `clang-cl` on my past Windows machine :)
But here the flag belongs to the `NOT WIN32` block so it is not going to apply on Windows anyway.

IvanSavenko (approved):

PR #1645, opened 2023-03-06
Allow rewardables to show its rewards in info bar

rilian-la-te:
Self-explanatory.

When content cannot fit into InfoBar, creates next infobar window, which will shown after timeout.


IvanSavenko:
"Status bar" is text-only area at the bottom of most windows.
I think you mean "Info Bar" - bottom-right box on adventure map?

rilian-la-te:
Yes, info bar, like in HD mod. But I want screen from Library of Enlightenment from HD mod, to look how they show more than 2 components in InfoBar.

client/CPlayerInterface.cpp
```diff
@@ -1031,13 +1031,13 @@ void CPlayerInterface::yourTacticPhase(int distance)
 		boost::this_thread::sleep(boost::posix_time::millisec(1));
 }
 
-void CPlayerInterface::showComp(const Component &comp, std::string message)
+void CPlayerInterface::showComps(const std::pair<Component,Component> & comps, const std::pair<bool, bool> & use, std::string message)
```
IvanSavenko, line 1034: Use std::vector instead of std::pair

krs0:
> "Status bar" is text-only area at the bottom of most windows. I think you mean "Info Bar" - bottom-right box on adventure map?

Yup, he means the "Info Window" :)

client/CPlayerInterface.cpp
rilian-la-te: @IvanSavenko why? I cannot show more than 2 objects now(

IvanSavenko: You can show 0, 1 or 2 components now.
Vector would be more suitable instead of two interconnected pairs.

Besides - I don't think that server logic should be aware of how many objects can be shown in UI.
If you wish - drop "assert(size <= 2)" in UI loading.
Also, if somebody were to upgrade this code to support 3+ elements then this means rewriting entire code since you can't just store third element in pair.

If standard library had "vector with specific max capacity" class I would've used it. But it does not.

rilian-la-te: @IvanSavenko will try to rewrite it using CComponentBox, maybe will work for 3+ components.

IvanSavenko: Not sure if CComponentBox would change anything.
And I think we should use vector even with your current version.

Also, if you want to fit more elements - CComponent constructor accepts "size". You can use smaller image if you have over 2 components to display

rilian-la-te: @IvanSavenko I do not want to write 8 constructors for each variant. I will try CComponentBox.

IvanSavenko:
Also - apart from Library there are also Stables (0 components), and revisiting previously visited objects, including Warrior's Tomb (has components on revisiting)

IvanSavenko:
Also, currently this setting should be configured per-object. Sure, you can made simple mod that covers H3 objects, but it won't cover objects added by another mod.

Proposition:
1) Add default value into "config/defaultMods.json"
2) Allow objects to override default behavior - if object config has "showInInfobox" property use it instead of global default

client/CPlayerInterface.cpp
rilian-la-te: @IvanSavenko stables is 0 components only if you do not have Cavaliers.

rilian-la-te: Done with vector and CComponentBox. Is there any method than:
1. Gets only first row from incoming string
2. Gets required CTextBox height for showing string without scrollbar (in given width)?

rilian-la-te:
Maybe more radical - change InfoWindow and ShowInInfobox into one CPack with two bools than force either display? If bools are not found - up to client how display InfoWindows.

Nordsoft91:
cool, don't forget to update wiki after merge https://github.com/vcmi/vcmi/wiki/Modding-~-Objects-~-Rewardable

rilian-la-te:
@IvanSavenko, other objects should set this to AUTO, and then it should be adventure setting "infobar if possible", and then if there is less than 9 components and text can be packed in infobar - it will show in infobar.

IvanSavenko:
> @IvanSavenko, other objects should set this to AUTO, and then it should be adventure setting "infobar if possible", and then if there is less than 9 components and text can be packed in infobar - it will show in infobar.

Care to test every single message in the game in every possible combination and make sure that nothing is broken?

Unless we know *for sure* that these messages can be moved into infobar without issues we should use same form as before

rilian-la-te:
> > @IvanSavenko, other objects should set this to AUTO, and then it should be adventure setting "infobar if possible", and then if there is less than 9 components and text can be packed in infobar - it will show in infobar.
> 
> Care to test every single message in the game in every possible combination and make sure that nothing is broken?
> 
> Unless we know _for sure_ that these messages can be moved into infobar without issues we should use same form as before

I need to write a checker, like "infobar fitting possible", (less than 8 components, and can fit both text and components into infobar) - then AUTO messages with "infobar if possible is ON" should be moved to infobar.

Offtopic: why scholar and witch hut is not Rewardables? What problems they have?

IvanSavenko:
> I need to write a checker, like "infobar fitting possible", (less than 8 components, and can fit both text and components into infobar) - then AUTO messages with "infobar if possible is ON" should be moved to infobar.

Size is not the only issue. In some cases there can be multiple messages after each other. In some cases these messages normally block game till message is closed. And there may be more cases that I am not aware of.

So please, no. 

> Offtopic: why scholar and witch hut is not Rewardables? What problems they have?

1) Unlike other objects they are configurable in h3m and not completely defined in config. 
2) They need logic where multiple values share same random value, e.g. limiter "hero has skill" should have the same value as reward "hero receives skill". Not impossible to implement but not something supported by Rewardable.

rilian-la-te:
At least, almost all object messages should be AUTO, AFAIK, like in HD mod.

IvanSavenko:
> At least, almost all object messages should be AUTO, AFAIK, like in HD mod.

Almost all object messages are already covered by Rewardable.
Few other exceptions (e.g. Witch Hut/Scholar/etc) should be checked manually whether infobar works with them and if yes - then set to auto.

rilian-la-te:
@IvanSavenko mostly done (except for timers, I am not sure, if we should have configurable timer for InfoBar pickups, for GCResource it will match H3, but for other - we can have configurable timer).

client/adventureMap/CInfoBar.h
```diff
@@ -122,13 +133,19 @@ class CInfoBar : public CIntObject
 
 	void playNewDaySound();
 public:
+
+	static constexpr int data_width = CInfoBar::width - 2 * CVisibleInfo::offset_x;
+	static constexpr int data_height = CInfoBar::height - 2 * CVisibleInfo::offset_y + 4; //yes, +4 is required
+
+	static constexpr int component_offset = 8;
+
```
IvanSavenko, line 141: Any need to have them public?

client/adventureMap/CInfoBar.cpp
```diff
@@ -163,16 +163,62 @@ CInfoBar::VisibleGameStatusInfo::VisibleGameStatusInfo()
...
 
-	text = std::make_shared<CTextBox>(message, Rect(10, 4, 160, 50), 0, FONT_SMALL, ETextAlignment::CENTER, Colors::WHITE);
+	if(!message.empty())
+		text = std::make_shared<CTextBox>(message, textRect, 0, font, ETextAlignment::CENTER, Colors::WHITE);
+}
+
+int CInfoBar::getEstimatedComponentHeight(int numComps) const
+{
+	if (numComps > 8) //Bigger than 8 components - return invalid value
+		return std::numeric_limits<int>::max();
```
IvanSavenko, line 214:
```suggestion
	assert(numComps <= 8);
	if (numComps > 8) //Bigger than 8 components - return invalid value
		return std::numeric_limits<int>::max();
```

rilian-la-te:
@IvanSavenko Implemented long info windows, 1.5 times bigger than usual (AUTO will use long info windows, but INFO ones do not). I want final review and maybe merge into 1.2.

lib/mapObjects/CRewardableObject.cpp
```diff
@@ -448,7 +450,7 @@ void CRewardInfo::loadComponents(std::vector<Component> & comps,
 		comps.emplace_back(Component::EXPERIENCE, 1, heroLevel, 0);
 
 	if (manaDiff || manaPercentage >= 0)
-		comps.emplace_back(Component::PRIM_SKILL, 5, manaDiff, 0);
+		comps.emplace_back(Component::PRIM_SKILL, 5, manaDiff ? manaDiff : h->manaLimit() * manaPercentage / 100, 0);
```
IvanSavenko, line 453: This is fix for mana well UI bug? Where it says it gives 0 mana?
Correct solution would be to extract mana amount calculations from `CRewardableObject::grantRewardAfterLevelup` method into a separate one & use it here.

IvanSavenko (commented):
Also - since translation (I presume) complete, please remove "validateTranslation" from "russian" section in mods/vcmi/mod/mod.json and check that there are no reports of missing strings on startup

lib/NetPacks.h
```diff
@@ -1150,6 +1150,10 @@ struct DLL_LINKAGE NewTurn : public CPackForClient
 
 struct DLL_LINKAGE InfoWindow : public CPackForClient //103  - displays simple info window
 {
+	enum {
+		AUTO, MODAL, INFO
+	};
+	int type = MODAL; //enum above
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko I need to extract it from InfoWindow then, because enum class there does not work.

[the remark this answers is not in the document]
IvanSavenko: This is why
> Optionally - move this enum to NetPacksBase.h

(And in that case name it EWindowMode / EInfoWindowMode or something similar)

client/adventureMap/CInfoBar.cpp
rilian-la-te: @IvanSavenko assert is not good here, because modder can set 9 components as reward, and then we just return false from infowindow, and show modal window instead.

client/widgets/CComponent.h
```diff
@@ -109,10 +113,10 @@ class CComponentBox : public CIntObject
 	int selectedIndex();
 
 	/// constructor for non-selectable components
-	CComponentBox(std::vector<std::shared_ptr<CComponent>> components, Rect position);
+	CComponentBox(std::vector<std::shared_ptr<CComponent>> components, Rect position, int betweenImagesMin = 20, int betweenSubtitlesMin = 10, int betweenRows = 22);
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko I do not know how to fit delegated constructors there. I just need these 3 values to come from constructor, not from constant, delegated constructor would not do anything other than setting this 3 values. Or you mean - write one constructor and then rewrite all other constructors to use it?

client/adventureMap/CInfoBar.h
rilian-la-te: made private and method rewritten

client/adventureMap/CAdvMapInt.cpp
```diff
@@ -86,7 +86,7 @@ CAdvMapInt::CAdvMapInt():
 	statusbar(CGStatusBar::create(ADVOPT.statusbarX,ADVOPT.statusbarY,ADVOPT.statusbarG)),
 	heroList(new CHeroList(ADVOPT.hlistSize, Point(ADVOPT.hlistX, ADVOPT.hlistY), ADVOPT.hlistAU, ADVOPT.hlistAD)),
 	townList(new CTownList(ADVOPT.tlistSize, Point(ADVOPT.tlistX, ADVOPT.tlistY), ADVOPT.tlistAU, ADVOPT.tlistAD)),
-	infoBar(new CInfoBar(Rect(ADVOPT.infoboxX, ADVOPT.infoboxY, 192, 192))),
+	infoBar(new CInfoBar(Rect(ADVOPT.infoboxX, ADVOPT.infoboxY, CInfoBar::width, CInfoBar::height))),
```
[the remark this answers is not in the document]
rilian-la-te: done with adding other constructor.

Mods/vcmi/config/vcmi/english.json
```diff
@@ -47,6 +47,8 @@
 	"vcmi.systemOptions.framerateButton.hover"  : "Show FPS",
 	"vcmi.systemOptions.framerateButton.help"   : "{Show FPS}\n\n Toggles visibility of Frames Per Second counter in corner of game window.",
 
+	"vcmi.adventureOptions.infoBarPick.hover" : "Show messages in Info Panel",
+	"vcmi.adventureOptions.infoBarPick.help" : "{Show messages in Info Panel}\n\n If possible for messages, shows messages in Info Panel, if on adventure map (not in town or in any other screen).",
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/NetPacks.h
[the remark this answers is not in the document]
rilian-la-te: done

client/widgets/CComponent.h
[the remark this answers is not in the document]
IvanSavenko: Basically yes - use delegated constructor to hide default parameters.
Currently you have two constructors:
(well, two pairs, but each pair is similar:
```
CComponentBox(std::vector<std::shared_ptr<CComponent>> components, Rect position); 
CComponentBox(std::vector<std::shared_ptr<CComponent>> components, Rect position, int betweenImagesMin, int betweenSubtitlesMin, int betweenRows);
```
Constructor 1 can be converted into delegated constructor and implemented as
```
CComponentBox(std::vector<std::shared_ptr<CComponent>> components, Rect position)
: CComponentBox(components, position, 20, 10, 22)
```
(same goes for second pair of constructors, with selectable components)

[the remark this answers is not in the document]
rilian-la-te: done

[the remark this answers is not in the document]
IvanSavenko: > But why this is better than just 2 constructors?

In this case you basically have 2 modes:
a) create box with default size values
b) create box with custom size values

There is no point (for example) in defining betweenImagesMin but keeping default value for betweenSubtitlesMin/betweenRows.

Also, having multiple default values might lead to issues later. For example if you remove betweenSubtitlesMin code that provided 4 parameters will still work, but will 4th value will now be used for betweenRows value.
And this is not some theoretical case, but something that happened in vcmi codebase before and something that happened in company I worked at before.

lib/mapObjects/CRewardableObject.h
```diff
@@ -343,6 +348,7 @@ class DLL_LINKAGE CRewardableObject : public CArmedInstance
 		h & static_cast<CArmedInstance&>(*this);
 		h & info;
 		h & canRefuse;
+		h & showInInfobox;
```
IvanSavenko, line 351: FYI: generally, for minor changes like this you should:
1) Bump serialization version in CSerializer.h (but keep minimal version intact)
2) Serialize affected values in form
```
if ( h >= <BUMPED_VERSION>)
    h & showInInfobox;
else
    h = false;
```
In this case probably not important - will merge this PR alongside with your secondary skill changes, which bump minimal save format anyway.

IvanSavenko (approved):
Looks OK for me, will merge in a few days

lib/mapObjects/CRewardableObject.h
rilian-la-te: If I do not need to it now, so, let's resolve this conversation? 

But what structures is saved, and what not? I need to know it)

IvanSavenko: > But what structures is saved, and what not? 

Generally - everything from library side gets saved. There are some exceptions, but those are mostly due to inability to serialized them (e.g. std::function's).

Yes, this is extensive, but this approach works fine. In long-term, dropping serialization of handlers and serializing only map itself might be a good idea, but this is definitely not something needed soon.

rilian-la-te: done, serialization version bumped

client/adventureMap/CInfoBar.cpp
```diff
@@ -266,13 +322,144 @@ void CInfoBar::showDate()
 	redraw();
 }
 
-void CInfoBar::showComponent(const Component & comp, std::string message)
+
+void CInfoBar::pushComponents(const std::vector<Component> & components, std::string message, int timer)
+{
+	auto actualPush = [&](const std::vector<Component> & components, std::string message, int timer, size_t max){
+		std::vector<Component> vect = components; //I do not know currently how to avoid copy here
+		while(!vect.empty())
```
IvanSavenko, line 330: 1) Do you need lambda here? It is called once and is not executed in async mode at some point later
2) Personal opinion, but I would rather call it "performPush", "pushImpl", "doPush". For me "actualPush" does not sounds like a good name for some reason...
3) You worry about "optimization" too much. Oh noes, you've copied vector with 5? 10? elements which took 1 mks. But the you need to render this components which can easily take 100 or even 1000 times more than this copy. There is no point in such micro-optimizations. If you want to improve performance the you need to investigate actual timings via profiler.

IvanSavenko (approved):
Looks OK, merging

lib/mapObjects/CObjectHandler.h
```diff
@@ -57,7 +58,13 @@ class DLL_LINKAGE IObjectInterface
 	virtual void garrisonDialogClosed(const CGHeroInstance *hero) const;
 	virtual void heroLevelUpDone(const CGHeroInstance *hero) const;
 
-//unified interface, AI helpers
+	//unified helper to show info dialog for object owner
+	virtual void showInfoDialog(const ui32 txtID, const ui16 soundID = 0, EInfoWindowMode mode = EInfoWindowMode::AUTO) const;
+
+	//unified helper to show a specific window
+	static void openWindow(const EOpenWindowMode type, const int id1, const int id2 = -1);
```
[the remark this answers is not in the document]
rilian-la-te: But there was converted from a bunch of static functions, so, I made it static. I do not think than we should avoid static like plague) I think if we will even need instance access, then we can convert this to non-static.

client/adventureMap/CInfoBar.cpp
rilian-la-te: 1. These lambda is purely for readability. Of course, it can be changed to not using lambda, but what's wrong with it.
2. Maybe. It can be called pushImpl, for sure. Can be renamed later.
3. On huge pandoras this vector can contain around 1000 elements.

lib/mapObjects/CObjectHandler.h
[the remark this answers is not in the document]
IvanSavenko: Static does have usage. For example singleton access.

But other than that - hardly. In my opinion static is indication of poor design.
If you have static method it is likely that this method should be either part of some namespace (and not part of class) or should be part of another class.

Whether method makes any access to class members is (normally) an implementation detail and should not be exposed to user code.

> I think if we will even need instance access, then we can convert this to non-static.

Problem is, if/when we do so, it won't be as simple as removing "static" from method declaration. This also means updating every call that used access without instance. And this part might not be trivial. No need to promote poor code that will have to be rewritten later.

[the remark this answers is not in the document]
rilian-la-te: For now, there is no call of openWindow with class access notation. So, changing it to instance method is zero-cost.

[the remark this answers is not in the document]
nullkiller: static/non-static is more like behavior of instance or class in general. If the method describes instance behavior but still does not use instance state it is still better to leave it instance method. But if it describes something more generic, overall class behavior, some meta-information it may be static. This thing looks like a factory method so it is ok for it to be static. Otherwise we will need some window manager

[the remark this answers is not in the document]
IvanSavenko: This is not factory method. 
It generates and send network pack that asks client to show window.

And internally it uses another static member (that should not be static)

client/adventureMap/CInfoBar.cpp
rilian-la-te: @IvanSavenko should I change anything here?

IvanSavenko: Up to you.
In my opinion lambda is unnecessary, but this is very minor issue, feel free to ignore.

Issue #820, opened 2022-09-04
sleep/wake hero button icon doesn't update immediately after press


kambala-decapitator:
The function itself is applied immediately, only button icon is changed after switching to another hero and back to the current one

PR #1651, opened 2023-03-07
Fix sleep/wake button icon update

Adriankhl:
Fix this issue: https://github.com/vcmi/vcmi/issues/820


kambala-decapitator (approved):
the fix looks valid

PR #1653, opened 2023-03-07
Changelog update: Added changes from PR's made in February

IvanSavenko:


DjWarmonger:
> first item in adventure: "adventure map rendering was entirely rewritten with better, more functional code"

I think it's enough to say that map rendering is more efficient and smoother. And in fact that's already stated in the changelog.

PR #1657, opened 2023-03-08
NKAI: improve build behavior

nullkiller:
Attempt to improve build behavior
- increase non-gold mines priority
- increase priority of dwelling upgrades
- reduce priority of capitol/city hall if we need to save resources to build it - better to have instant gain than delayed one, moreover delayed for unknown period of time as we cannot say for sure when we will be able to accumulate enough resources


dydzio0614:
maybe there should be some setting for building priorities so AI won't try to rush 5th/6th level dwellings and upgrades by default just because they are cheaper than city hall or something

heavy dwelling rush would be better on very rich maps / multiplayer style maps for example

Issue #795, opened 2022-09-03
some numbers shouldn't be formatted according to locale


kambala-decapitator:
Formatting port number according to locale makes no sense. Also noticed the same in resolution values in the game menu.
I have English Windows 8.1 with region set to Russia.

<img width="30%" src="https://user-images.githubusercontent.com/1557784/188262214-a940ebee-3be0-49d3-8c27-9a2def1ba223.jpg" />

dydzio0614:
We use boost::lexical_cast for int->string conversion which takes locale into account. While on my linux with polish locale nothing happens, on Windows comma separator is present. Maybe we should convert ints via std::to_string instead?

GeorgeK1ng:
This issue no longer exist in 1.7

PR #1662, opened 2023-03-09
vcmi: reduce boost::lexical_cast usage

rilian-la-te:
There is only 4 instances of boost::lexical_cast<std::string> survived.


dydzio0614:
related issue: https://github.com/vcmi/vcmi/issues/795

rilian-la-te:
@dydzio0614 it will not help, we should use `std::to_chars` in such places. But this requires bumping minimal C++ version to C++17.

IvanSavenko (approved):
Looks OK.

@rilian-la-te any other similar PR's in plans / in work?

Issue #1383, opened 2023-01-07
sleep hero state isn't restored on loading game


kambala-decapitator:
steps:
1. set a hero as sleeping
2. save game
3. load game
4. the hero is awake

Povelitelh3:
related https://bugs.vcmi.eu/view.php?id=1017

Adriankhl:
I am trying to dig into this issue. In principle, it seems like the functionality should have been implemented - the `CPlayerInterface::serializeTempl` function should serialize and store the `sleepingHeroes` data. When I save a game, `CPlayerInterface::saveGame` is called to serialize something, but the counterpart `CPlayerInterface::loadGame` is not being called when I load the game. That's probably the cause of this issue.

However, I must say the binary serializer and the server-client architecture are somewhat cryptic to me so it is quite hard to narrow down the exact location of the problem. Is there anyone who are more familiar with these pieces of code?

kambala-decapitator:
doesn't work in 1.5

IvanSavenko:
Yeah, that PR got effectively reverted. Same goes to #797

Laserlicht:
@IvanSavenko Would it be possible to fix this for 1.6?

Because this would fix multiple issues at once.

IvanSavenko:
Not too much work, mostly a question of priorities.
Also not 100% about how to implement it. Considering several potential approaches, but not sure which one would be the best option.

Will add to my 1.6 todo list, but no promises for now.

PR #1663, opened 2023-03-09
Fix client data deserialization, such as sleepingHeroes

Adriankhl:
Seems like I managed to:
- fix https://github.com/vcmi/vcmi/issues/1383
- fix https://github.com/vcmi/vcmi/issues/797

There are 3 changes:

1. The `serialize()` function of `ObjectClass` misses the `id` field, so I fixed it and bump `SERIALIZATION_VERSION` and `MINIMAL_SERIALIZATION_VERSION` to 818.
2. The `CClient::loadGame()` function now try to load client data from the local file if it exists to recover data such as `sleepingHeroes`, `paths`, etc. Common game state data is no longer saved in `.vcgm1` anymore.
3. Call `loadGame` after `installNewPlayerInterface` to deserialize `paths` properly.


client/Client.cpp
```diff
@@ -209,41 +209,45 @@ void CClient::loadGame(CGameState * initializedGameState)
...
-	else
+
+	// assignment of the loader, which is needed to deserialze data such as sleepingHeroes
+	try
 	{
-		try
-		{
-			boost::filesystem::path clientSaveName = *CResourceHandler::get("local")->getResourceName(ResourceID(CSH->si->mapname, EResType::CLIENT_SAVEGAME));
-			boost::filesystem::path controlServerSaveName;
+		boost::filesystem::path clientSaveName = *CResourceHandler::get("local")->getResourceName(ResourceID(CSH->si->mapname, EResType::CLIENT_SAVEGAME));
```
nullkiller, line 216: Feels like you want to merge gamestate passed over network (gs) and thing loaded from file. You removed else and now initalizedGamestate is not used. As far as I understand things. So you can try to load from file first and put needed data into initializedGamestate and gs = initializedGamestate at the end.

Anyway it is more theoretical, lets wait somebody more acquainted in this area.

Adriankhl: I have also tried to reorder the data loading but it still crashes with similar complain about `primaryID`.

Currently I am trying to workaround this by reading the data from file to another object and add sleepHeroes back to the game based on some identifiers (`heroId`?), but I still have to understand how the serialization actually works to implement it.

IvanSavenko:
Tricky one.
@Adriankhl thanks for investigation. Looks like as of right now, client savegame is not read at all (but still generated)

PR #1000 seems to be the cause, more details in this comment: https://github.com/vcmi/vcmi/pull/1000#discussion_r984393797

So change *was* reasonable, but broke sleeping status & movement paths. Perhaps what we should do is:
1) check whether this try-catch block can be removed. I think the only way for CClient::loadGame to be called is with valid `initializedGameState` so currently game never reaches this try-catch
2) If this try-catch can be removed - then well, remove it.
3) Change serialization of client savegame - save ONLY client info, without gamestate ( read: remove saveCommonState call)
4) On game loading do the same - load ONLY client info (read: remove loadCommonState call)

This should fix both sleeping heroes and heroes paths, and should not cause any new issues

@Nordsoft91 Can you check this PR when you have a moment?

IvanSavenko:
Also - check whether your PR also fixes #797 
It *should*, since hero path also go into client savegame

Adriankhl:
> Also - check whether your PR also fixes #797 It _should_, since hero path also go into client savegame

Unfortunately no, something wrong here, `owner.cb` is `nullptr`:
https://github.com/vcmi/vcmi/blob/e764a7255d16d001e464d0f0aa4b8e9c92009c91/client/CPlayerInterface.cpp#L194-L203

But things like `spellbookSettings` seems to restore properly.

Adriankhl:
>     1. check whether this try-catch block can be removed. I think the only way for CClient::loadGame to be called is with valid `initializedGameState` so currently game never reaches this try-catch
> 
>     2. If this try-catch can be removed - then well, remove it.
I think the try-catch block should be there to deal with the case where some players might not have the client save file in a multiplayer game. Though the if-check of `initializedGameState` can be removed if it is always valid.

>     3. Change serialization of client savegame - save ONLY client info, without gamestate ( read: remove saveCommonState call)
> 
>     4. On game loading do the same - load ONLY client info (read: remove loadCommonState call)
Thanks for pointing this out, I have removed `saveCommonState` and `loadCommonState` in client and it seems to work properly!

Adriankhl:
> Also - check whether your PR also fixes #797 It _should_, since hero path also go into client savegame

Ah, the path deserialization can be fixed by a simple reorder of function calls.

client/Client.cpp
```diff
@@ -201,62 +201,52 @@ void CClient::newGame(CGameState * initializedGameState)
...
+			controlServerSaveName = boost::filesystem::path(clientSaveName).replace_extension(".vsgm1");
+			CResourceHandler::get("local")->createResource(controlServerSaveName.string(), true);
+		}
+
+		if(clientSaveName.empty())
+			throw std::runtime_error("Cannot open client part of " + CSH->si->mapname);
+		if(controlServerSaveName.empty() || !boost::filesystem::exists(controlServerSaveName))
+			throw std::runtime_error("Cannot open server part of " + CSH->si->mapname);
+
+		CLoadIntegrityValidator checkingLoader(clientSaveName, controlServerSaveName, MINIMAL_SERIALIZATION_VERSION);
```
IvanSavenko, line 239: Can you check whether CLoadIntegrityValidator (and all access to server save game) can be removed from here?
Not much point behind it now since we no longer do any "load integrity validation"

IvanSavenko (approved):

IvanSavenko:
Looks OK, should be ready for merge.
Will probably keep it as PR for a few days, to postpone saves breaking.

rilian-la-te:
@IvanSavenko maybe merge together with my #1518?

PR #1676, opened 2023-03-12
vcmi: fix week animation

rilian-la-te:
There was always day animation running on new weeks. Run week animation properly now.
Timer for week animation is estimation, maybe not accurately match H3.

### DIFF
--- a/client/adventureMap/CInfoBar.cpp
+++ b/client/adventureMap/CInfoBar.cpp
@@ -64,6 +64,7 @@ CInfoBar::VisibleDateInfo::VisibleDateInfo()
 	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
 
 	animation = std::make_shared<CShowableAnim>(1, 0, getNewDayName(), CShowableAnim::PLAY_ONCE, 180);// H3 uses around 175-180 ms per frame
+	animation->setDuration(1500);
 
 	std::string labelText;
 	if(LOCPLINT->cb->getDate(Date::DAY_OF_WEEK) == 1 && LOCPLINT->cb->getDate(Date::DAY) != 1) // monday of any week but first - show new week info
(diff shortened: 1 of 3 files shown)


client/adventureMap/CInfoBar.cpp
```diff
@@ -99,6 +99,12 @@ std::string CInfoBar::VisibleDateInfo::getNewDayName()
 	}
 }
 
+int CInfoBar::VisibleDateInfo::getNewDayLength()
+{
+	// confirmed to match H3 for 3000 for NEWDAY
+	return std::max(3000, animation->getLength() * 2);
```
IvanSavenko, line 105: Why this change?
Quick test shows that at least week2 & week3 are also visible for 3000 ms

rilian-la-te: It is not fully visible, at least week 3 and 4. 1 and 2 - yes, visible, so this is why max.

rilian-la-te: Week 3 and 4 requires more time to show full anim, but I unsure how many more.

IvanSavenko: Ah, it looks like H3 always shows animation for 3 seconds, but instead calculates speed to play it in 1.5 seconds

rilian-la-te: So, how I should change it to match H3?

rilian-la-te: Replace 180 with 90? Or do it only on new weeks?

IvanSavenko: Hmm.. All 5 files have different number of frames, but same duration

Maybe create method 
```
CShowableAnim::setDuration( int durationMilliseconds)
{
    frameTimeTotal  = durationMilliseconds/ (last - first)
}
```

And then use it as 
`animation->setDuration(1500)`

And revert changes in this method, back to "3000" constant

rilian-la-te: done

IvanSavenko (approved):

PR #1680, opened 2023-03-14
Catapult nearest wall

rilian-la-te:
Catapult now to be able to choose nearest wall instead of random.

(fixes part of #1236)

### DIFF
--- a/lib/spells/effects/Catapult.cpp
+++ b/lib/spells/effects/Catapult.cpp
@@ -115,6 +115,39 @@ void Catapult::applyTargeted(ServerCallback * server, const Mechanics * m, const
 	auto destination = target.at(0).hexValue;
 	auto desiredTarget = m->battle()->battleHexToWallPart(destination);
 
+	auto nearestAttackableWall = [&m,&server](EWallPart part) {
+		auto choosePart = [&] (EWallPart left, EWallPart right, int max = 2) {
+			bool bothSidesAttackable = m->battle()->isWallPartAttackable(left)
+										&& m->battle()->isWallPartAttackable(right);
+			if(bothSidesAttackable)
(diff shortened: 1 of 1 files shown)


IvanSavenko:
Hmm... 
Not sure about this.
1) I doubt that this how H3 works here. Such details are not documented anywhere. Have you investigated it and confirmed or you're basing this on my comment?
2) Current code is already close enough - very few players will notice a difference, so not sure if there is any need to add another code that also does not matches H3
3) Really dislike such brute-force approach. Picking 2 closest targets would have been better instead of hardcoding every possible combination

rilian-la-te:
@IvanSavenko I am not sure too. I think old approach was good enough, but you have a TODO inside your meta-issue.

About distance - we currently does not have such method as count wall distance. Should I add this method?

rilian-la-te:
@IvanSavenko FizMig does not says anything about catapult and how it should hit target, if it misses.

dydzio0614:
then do something moir or less similar for now and add TODO?

IvanSavenko:
> About distance - we currently does not have such method as count wall distance. Should I add this method?

Indirectly it already exists - you can check distance between hexes to which walls are bound. However it won't match H3. I only tested few cases, but at the very least with gates catapult would randomly attack wall above or below. But as it is now, distance (in hexes) from these walls to gates would not b equal due to how they are placed.

Personally I would rather leave it as it for now until we find (or investigate in details) H3 behavior here.

> then do something more or less similar for now and add TODO?

That's precisely what I did :)

rilian-la-te:
So, let's close it then? (without merging)?

IvanSavenko:
Probably yes.

Issue #1687, opened 2023-03-15
Make sure RMG doesn't run out of Monoliths pairs


DjWarmonger:
**Describe your proposal**
As of current develop, if RMG runs out of Monolith pairs on complex maps, it simply stops to place them, generating inacessible zones. My idea was to clone object templates for Monoliths, assuring there are at least 100 pairs with unique id. Even though graphics could repeat, internally they would have different id and make independent pairs, ensuring map passability.

Especially noticeable on templates featuring 48 zones, such as 8mm12.

**Additional context**
It never worked properly - Monoliths were not unique (some pairs used same index, thus all the Monoliths were connected together). but at least all the Monoliths were placed where needed.

DjWarmonger:
Fixed in https://github.com/vcmi/vcmi/pull/1694

PR #1694, opened 2023-03-16
Create extra Monoliths for RMG

DjWarmonger:
Reuse existing set of Monolith graphics to create 100 pairs with different IDs, so that each pair can create an independent connection.

- Fixes https://github.com/vcmi/vcmi/issues/1687
- Additionally, RMG won't use Monoliths which can't be placed at any land. Some Monoliths from mods do not meet this condition. Missing template could lead to RMG error (couldn't find template for object 45) and missing connection.

imageimage

### DIFF
--- a/lib/mapObjects/CObjectClassesHandler.cpp
+++ b/lib/mapObjects/CObjectClassesHandler.cpp
@@ -393,12 +393,43 @@ void CObjectClassesHandler::afterLoadFinalization()
 		}
 	}
 
+	generateExtraMonolithsForRMG();
+}
+
+void CObjectClassesHandler::generateExtraMonolithsForRMG()
+{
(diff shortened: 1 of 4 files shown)


Alucard648:
Speaking about terrain agnostic monoliths: What about checking, if there are terrain specific monoliths available for connecting different zones with the same terrain? For instance, WoG snow monolith can connect two snow terrain zones, as originally designed by WoG team.
As for naval monoliths, when generating islands map, place pairs of these monoliths that are at least 50-60 tiles apart for faster ship travel.

DjWarmonger:
I considered this - however, it has limited use. In such case **both** zones need to be snow, which is rare. Also, this would give player a clue that the zone on the other side has certain terrain type - which cmight be desired or not, but is different than original behavior.

PR #1699, opened 2023-03-16
Rewardable: always initialize numbers

rilian-la-te:
This fixed @krs bug about visited rewards, which I also encountered.

### DIFF
--- a/lib/mapObjects/CRewardableObject.h
+++ b/lib/mapObjects/CRewardableObject.h
@@ -73,8 +73,11 @@ public:
 	CRewardLimiter():
 		dayOfWeek(0),
 		daysPassed(0),
+		heroExperience(0),
 		heroLevel(0),
-		primary(4, 0)
+		manaPercentage(0),
+		manaPoints(0),
+		primary(GameConstants::PRIMARY_SKILLS, 0)
 	{}
 
 	bool heroAllowed(const CGHeroInstance * hero) const;


rilian-la-te:
I do not know why, but default constructors is required on MSVC.

SoundSSGood:
What the reason to change fields initialization from constructor to inside class declaration?

rilian-la-te:
@SoundSSGood what is a reason to have one constructor which always initialize values to same default values? Initializing in class gives a less LOC, and more cleaner in this case. We can actually see which variables are initialized, and which are left uninitialized, which can lead to some strange bugs like @krs0 found.

rilian-la-te:
@IvanSavenko please, look for this patch, CI fixed)

IvanSavenko:
> What the reason to change fields initialization from constructor to inside class declaration?

No reason other than @rilian-la-te preference.
Allows exposing internal information to header and lets you play "find initializer" game when both in-place initialization and constructor are present.

lib/mapObjects/CRewardableObject.h
```diff
@@ -29,26 +29,26 @@ class DLL_LINKAGE CRewardLimiter
...
 	/// percentage of mana points that hero needs to have
-	si32 manaPercentage;
+	si32 manaPercentage = 0;
 
 	/// resources player needs to have in order to trigger reward
 	TResources resources;
 
 	/// skills hero needs to have
-	std::vector<si32> primary;
+	std::vector<si32> primary {0, 0, 0, 0};
```
IvanSavenko, line 51: This should have GameConstants::PRIMARY_SKILLS instead of hardcoded 4 elements.
Perhaps switch to std array?

SoundSSGood:
This is different from the rest of the code style.
I don't think it's a good practice for parts of a project to differ in style.
Are there any plans to transform the style of the rest of the code?
Perhaps it should be included in the style rules?

rilian-la-te:
@SoundSSGood entire lib now uses default initialozation when possible.

IvanSavenko:
> @SoundSSGood entire lib now uses default initialozation when possible.

No it is not.
It is a mess right now.
The very first header that I opened, CCreatureHandler:
https://github.com/vcmi/vcmi/blob/develop/lib/CCreatureHandler.h#L40
- faction and level use zero-init in place
- fightValue, AIValue, growth, hordeGrowth are initialized in constructor

Half of the classes use in-place init, half don't
Half of class members use in-place init, half don't 

And so far ALL regressions that I found from your PR's are exclusively due to messy initializations - e.g. uninitialized pointer.

rilian-la-te:
So, even for simple struct with 3 ints you prefer to use default constructor? (like one of structs here) 

About large structs with complex initialization - you are right, it can be a mess.

IvanSavenko:
> simple structs without a constructor

CCreature is NOT a simple struct. Neither are CGHeroInstance, CStack and multiple other classes.

This is a simple struct: https://github.com/vcmi/vcmi/blob/develop/lib/CTownHandler.h#L167
This is not a simple struct: https://github.com/vcmi/vcmi/blob/develop/lib/CCreatureHandler.h#L31

IvanSavenko:
IMO anything more complex than our netpacks should not use in-place init.
If you wish some sort of "rule" then anything marked as "struct" can use in-place init, anything marked as "class" should use constructor init

rilian-la-te:
@IvanSavenko agree with you. But is there a non-simple structs changed by me in this PR? AFAIK, there is all simple structs. For CCreatureHandler and friends maybe worth creating private default constructor and inherit all constructor from it to avoid initializating same variable by same value in ten places.

IvanSavenko:
> @IvanSavenko uninitiated pointers was here for a years.

Not sure which pointer you're referring to, but this one was properly initialized in constructor before:
https://github.com/vcmi/vcmi/commit/8a6c51a9e6ae3e75bbc326c4050c7e8ca3aa44fc

rilian-la-te:
Because what is a point to have a default constructor in header for, for example, `CRewardResetInfo`?

rilian-la-te:
@IvanSavenko I discuss this change only in context of this PR, so, only classes which was Changes by me manually here. Changes in CCreature was overkill, agree with you. And in many other places too.

lib/mapObjects/CRewardableObject.h
rilian-la-te: There I need to rewrite a json load function too, and json load function can return less than 4 elements. So, what do you think should be better?

IvanSavenko: But it can not? It will always return 4 elements?
```

	std::vector<si32> loadPrimary(const JsonNode & value, CRandomGenerator & rng)
	{
		std::vector<si32> ret;
		for(const auto & name : PrimarySkill::names)
		{
			ret.push_back(loadValue(value[name], rng));
		}
		return ret;
	}
```

rilian-la-te: Okay, switching to std::array makes sense there. What do you think about default initializations not at all, but especially in this PR for secondary classes? I did this change only because there was one constructor in header, and I do not see why we should keep it (and when we will add another variable, we should manually keep constructor init + variable in sync).

IvanSavenko:
> @IvanSavenko I discuss this change only in context of this PR, so, only classes which was changed by me manually here. Changes in CCreature was overkill, agree with you. And in many other places too.

And yet you spent tons of your own time on these changes that I was objecting to, took a lot of my time as well, introduced several already found (and how many not found?) regressions.

And this is when the game is barely working as it and we have hundreds of bugs way more important than philosophic discussions on perfect class initialization form.

Let me ask this differently - are changes in this PR necessary to fix this bug?

lib/mapObjects/CRewardableObject.h
IvanSavenko: Keep as before. There is 0 need to change it IMO.

Spending time on. one of... 225 open issue would be way more productive than these "improvements"

rilian-la-te:
Necessary for fix is only changes in CRewardLimiter. If you wish, I can revert other changes (but this will lead to a code mess).

IvanSavenko:
Please revert. 

I am tired of these extremely enlightening discussions and extremely important changes.

SoundSSGood:
> If you wish some sort of "rule" then anything marked as "struct" can use in-place init, anything marked as "class" should use constructor init

I think it's not a bad idea to put the rule on the rules page. To avoid dirty class initialization in the future.

rilian-la-te:
@IvanSavenko as you wish, reverted all but limiter changes. Ignissa bug looks like to be fixed. Let's see if MSVC will build this.

lib/mapObjects/CRewardableObject.h
```diff
@@ -70,13 +70,6 @@ class DLL_LINKAGE CRewardLimiter
 	/// sub-limiters, none should pass for this limiter to pass
 	TRewardLimitersList noneOf;
 
-	CRewardLimiter():
-		dayOfWeek(0),
-		daysPassed(0),
-		heroLevel(0),
-		primary(4, 0)
-	{}
-
```
IvanSavenko, line 79: Was this hard?
I am tired of this arguing and of searching for new regressions.

```
	CRewardLimiter():
		dayOfWeek(0),
		daysPassed(0),
		heroExperience(0),
		heroLevel(0),
		manaPercentage(0),
		manaPoints(0),
		primary(GameConstants::PRIMARY_SKILLS, 0)
	{}
```

rilian-la-te: Okay, I will do this, this code will be reworked anyway after merging of banks and rewardables (in 1.3).

rilian-la-te: Done.

You simply dislike default initialization at all and prefer to write constructors everywhere. Understood, will not touch this in any old code.

IvanSavenko: I guess I am...

Was ambivalent a month before, but was completely unimpressed by how it turned out in vcmi code...

rilian-la-te: I just think than in some rare cases it is good (like in structs which lives entirely on header). For example, here we have clearly 3 structs (and two is almost the same, so, if I will do rewardables after moats, I think merging RewardInfo and RewardLimiter is reasonable). For CCreature, Bonus, and other big structs with multiple constructors - yes, it is reasonable to init everything in a constructor. Anyway, by reason you said in Slack, VCMI have very bad code even by C++11 standard.

rilian-la-te:
@IvanSavenko this is looking mergeable now?

lib/mapObjects/CRewardableObject.h
nullkiller: I would say there is a good engineering practice. Do not touch what works). So this is more about weird instructions which does not always work. C# for instance also introduced inline initialization as a new cool feature and we use it now at work. But nobody change old code just for fun). Only if you touch some file it is allowed to make mistakeless automatic refactoring. Why in C++ it can break just because you replaced on logically equivalent thing with another? Maybe there is something we do not know behind it? What is the order of default initialization and constructor if both defined?

rilian-la-te: @nullkiller there is interesting property for structs without a constructor: https://en.cppreference.com/w/cpp/language/aggregate_initialization.

If we define a constructor, we cannot use aggregate initialization. Default values allow us to specify default values even without an explicit constructor. 

I think for classes that can become aggregates with removal of one default constructor - usage of default members is reasonable.

rilian-la-te: We use c++14, so, we can use default member initializers on aggregates)

rilian-la-te: @IvanSavenko @nullkiller I mainly for converting this classes:
```c++
// aggregate
struct base1 { int b1, b2 = 42; };
 
// non-aggregate
struct base2
{
    base2() : b3(42) {}
 
    int b3;
};
 ```

Comment (IvanSavenko) [reply lib/mapObjects/CRewardableObject.h]: > Why in C++ it can break just because you replaced on logically equivalent thing with another?

It does not. All regressions were from incorrect initialization. For example, initializer was removed from constructor but not added to member declaration. So far with all "modernizations" I found this issue 3 times. Hopefully that's all of them, but I can only hope.

Also, this is personal opinion, but I dislike how it turns out in code, For example our typical library class right now looks like this:
```
class CArtifact
{
	ArtifactID id;
	std::string identifier;
	EartClass aClass = ART_SPECIAL;
	ui32 price = 0;
}
```
1) `id` field is not initialized in place - this may be effectively enum/integer, but since it is wrapped in class, ArtifactID class will perform its initialization instead.
2) `identifier` is not initialized in place either - since this is also a class, which handles its own initialization
3) `aClass` however uses in-place initialization. Because unlike `id` this is enum which has to be explicitly initialized with valid enum value;
4) `price` also uses in-place initialization since this is also primitive type.

As result we end up with this mess where half of members use in-place init, but half members do not. Before using in-place initialization this was not an issue. And these default values often don't make sense.

Another issue is readability - this change increases amount of text in headers, making it harder to get a quick look on class API. Not a problem for smaller classes that are 10 lines in size, but when you have class with definition that is 100-200 lines long I would rather keep as much info as possible external.

Comment (IvanSavenko) [reply lib/mapObjects/CRewardableObject.h]: @rilian-la-te 
If not for our serializer that requires default constructor I would have proposed to get rid of both default constructors and this aggregate initialization and instead give all such small structs constructor that *always* initializes all members, e.g.
```
class Point {
    int x, y;
    Point(int x, int y) : x(x), y(y) {}
}
```
Such form would have allowed us to
1) Always initialize all class members with *meaningful* values instead of zero-init
2) Allow us to detect/update code that does not provides (meaningful) values to some of the fields.

It is already a pain to find all instances in code where certain class (e.g. netpack) is created and update it, e.g. if you add new member. Aggregate init would make it even worse.

rilian-la-te: @IvanSavenko I disagree with you about readability in some cases. When we have constructor in header (or entire class in *.cpp, readability without constructor is better). About second proposal - it is pain in the ass to create a temporary values with deleted default constructor.

PR #1700, opened 2023-03-16
Modernize resourceSet.

rilian-la-te:
This will break saves, be warned.

If needed, I can bump serialization version.

### DIFF
--- a/lib/ResourceSet.cpp
+++ b/lib/ResourceSet.cpp
@@ -22,7 +22,7 @@ VCMI_LIB_NAMESPACE_BEGIN
 Res::ResourceSet::ResourceSet(const JsonNode & node)
 {
 	for(auto i = 0; i < GameConstants::RESOURCE_QUANTITY; i++)
-		at(i) = static_cast<int>(node[GameConstants::RESOURCE_NAMES[i]].Float());
+		this[i] = static_cast<int>(node[GameConstants::RESOURCE_NAMES[i]].Float());
 }
 
 Res::ResourceSet::ResourceSet(TResource wood, TResource mercury, TResource ore, TResource sulfur, TResource crystal,
--- a/lib/ResourceSet.cpp
+++ b/lib/ResourceSet.cpp
@@ -9,6 +9,7 @@
  */
 
 #include "StdInc.h"
+#include "GameConstants.h"
 #include "ResourceSet.h"
 #include "StringConstants.h"
 #include "JsonNode.h"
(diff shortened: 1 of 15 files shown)


lib/ResourceSet.h
```diff
@@ -32,60 +33,57 @@ namespace Res
 	};
 
 	//class to be representing a vector of resource
-	class ResourceSet : public std::vector<int>
+	class ResourceSet : public std::array<int, GameConstants::RESOURCE_QUANTITY>
```
IvanSavenko, line 36: Perhaps replace inheritance with composition?
Instead of inheriting from container include it as member and provide required interface where necessary, e.g begin()/end()/operator[]

I would also keep vector as before - eventually we might want to allow new resources & keeps save format

DjWarmonger: > eventually we might want to allow new resources
I don't see it, we already have Mithril which is unused. Also it may be difficult to handle more resources in GUI.
Overall the argumentation is similiar to adding more than 2 levels to a map.

dydzio0614: more than 2 levels of a map is on my list as important feature that should be eventually added

IvanSavenko: > I don't see it, we already have Mithril which is unused. Also it may be difficult to handle more resources in GUI. Overall the argumentation is similiar to adding more than 2 levels to a map.

Maybe true, but we should not *increase* difficulty in implementing it.

Best approach here would be to hide this implementation detail in internal container and only provide public interface. (meaning - remove inheritance)
External code should not be aware whether number of resources is fixed or not.

And inheriting from standard containers is already rather suspicious code IMO.

rilian-la-te: Done as composition, but as array.

lib/ResourceSet.cpp
```diff
@@ -18,37 +19,27 @@
...
-	d[Res::GOLD] = gold;
-	d[Res::MITHRIL] = mithril;
+	this[Res::WOOD] = wood;
+	this[Res::MERCURY] = mercury;
+	this[Res::ORE] = ore;
+	this[Res::SULFUR] = sulfur;
+	this[Res::CRYSTAL] = crystal;
+	this[Res::GEMS] = gems;
+	this[Res::GOLD] = gold;
+	this[Res::MITHRIL] = mithril;
```
IvanSavenko, line 38: Not sure whether we need to keep mithril. If/when it eventually gets used it'll likely be in a different way.
Might lead to some side effects though...

lib/ResourceSet.h
```diff
@@ -32,66 +33,146 @@ namespace Res
...
-#undef opEqOperator
+#undef twoOperands
+
+		using const_reference = decltype(container)::const_reference;
+		using value_type = decltype(container)::value_type;
+		using const_iterator = decltype(container)::const_iterator;
+		using iterator = decltype(container)::iterator;
+
+		// Array-like interface
+		TResource & at(Res::ERes index)
```
IvanSavenko, line 96: Do we need at() here? Perhaps always use operator[]?

rilian-la-te: @IvanSavenko what is wrong with at?

IvanSavenko: Don't see any need for it. Currently this only leads to more largely identical methods in resource set.

We almost never use at() in code in general, with exception of const std::map that does not have operator[].

lib/ResourceSet.cpp
rilian-la-te: About mithril removal - maybe it will break wog maps? I do not think than mithril placeholder is bad, I vote for keep it, at least for avoid side effects.

DjWarmonger: Well, we shouldn't make implementing it any more difficult than ii is now.

In fact with current configurable objects it should be already possible to give and take Mithril.

rilian-la-te: at removed.

lib/ResourceSet.h
rilian-la-te: at removed.

lib/ResourceSet.cpp
rilian-la-te: Agree with @DjWarmonger

lib/ResourceSet.h
rilian-la-te: Removal of .at breaks compilation in many places. Maybe save at?

IvanSavenko (approved):

rilian-la-te:
@IvanSavenko this should be merged along with #1706 to avoid double save breakage.

PR #1706, opened 2023-03-17
Proper moats: mechanincs

rilian-la-te:
- [x] Add a moat spell effect
- [x] Add a moat spell caster
- [x] Add a moat spells for all variants
- [x] Add a bonus type and bonus handling for bonuses which should work when creature placed on a hex (not needed)
- [x] Add this bonus with -3 defence to moats
- [x] Playtest all moats (8 variants so far)
- [x] Castle + Conflux
- [x] Rampart
- [x] Tower
- [x] Inferno
- [x] Necropolis
- [x] Dungeon
- [x] Stronghold
- [x] Fortress
- [x] Fix disappearing of landmines and fire walls (currently it stays infinitely on battlefield)
- [x] Fix all obstacles double hex bug with passing (moat should be passed as one step from double-wide units)
- [x] Fix double-affection obstacles bug (if we step on fortress moat, we will get its effects twice)
- [x] Fix disappearing animations of land mines and fire walls (along with sound)
- [x] Fix fire wall double hex bug
- [x] Fix same-hex moat effect applying (found this today). 

Includes:
1. #1777 
2. #1778 

These PRs should be merged in both beta and develop before merging this one.


Alucard648:
Will it be possible to cast any debuff on creature that stepped into moat? Like Berserk. And I have idea for moat variant: every time one steps into moat, Armageddon triggers.

rilian-la-te:
It is WIP, but for first likely yes, and for second likely no.

Alucard648:
Another suggestion for moats: creature ability to ignore moat effects (stopping, damage, triggers etc). For Bastion War Elephants.

rilian-la-te:
@Alucard648 creature ability to ignore moat effects is unneeded. Just add all moat spells to immunity for this creature.

rilian-la-te:
All: tower minefield now works correctly in any case (except disappearing animation).

All other moats needs to have finished same hex bonus (and for fortress moat I need to have one moat element per line, not 2, because now fortress moat activates twice for each hex)

Alucard648:
> @Alucard648 creature ability to ignore moat effects is unneeded. Just add all moat spells to immunity for this creature.

But this would require checking compatibility with mod towns on per creature basis. Imagine having 6-7 creatures with this ability and you had to do this test for every mod town you add with unique moat effect and ensure that it does not affect hero and creature spell immunities.

IvanSavenko:
Perhaps those are not "obstacles" but immobile summoned units?
Not sure whether we support selectable locations for summons though.

rilian-la-te:
@IvanSavenko this PR is already very huge, I think we should try to review and merge first part.

I did not added fade-out animation, but all other changes are done.

rilian-la-te:
There is more complex effect in H3 moat which currently does not implemented - even you do not get damage from moat, you get its defence penalty. So, it needs to be handled separately.

Alucard648:
> Not sure whether we support selectable locations for summons though.

Doing so will allow unhardcoding obstacle placement spells (Fire Wall and Force Field).

rilian-la-te:
@IvanSavenko done moat defence as in H3, but hackish way.

lib/spells/effects/Moat.cpp
```diff
@@ -0,0 +1,179 @@
...
+	auto all = m->battle()->battleGetAllObstacles(BattlePerspective::ALL_KNOWING);
+
+	int obstacleIdToGive = 1;
+	for(auto & one : all)
+		if(one->uniqueID >= obstacleIdToGive)
+			obstacleIdToGive = one->uniqueID + 1;
+
+	for(const auto & destination : moatHexes)  //Moat hexes can be different obstacles
+	{
+		SpellCreatedObstacle obstacle;
```
IvanSavenko, line 146: Perhaps obstacle (at least partially) should be loaded from config?

lib/battle/CObstacleInstance.h
```diff
@@ -120,6 +128,12 @@ struct DLL_LINKAGE SpellCreatedObstacle : CObstacleInstance
 		h & trap;
 
 		h & customSize;
+
+		if(version >= 821)
```
IvanSavenko, line 132: You've bumped minimal format. No need to check for it - all saves with this code are at 821 or newer.

server/CGameHandler.cpp
```diff
@@ -1436,10 +1341,14 @@ int CGameHandler::moveStack(int stack, BattleHex dest)
 	ret = path.second;
 
 	int creSpeed = gs->curB->tacticDistance ? GameConstants::BFIELD_SIZE : curStack->Speed(0, true);
+	bool hasLongBridge = vstd::contains_if(battleGetAllObstaclesOnPos(BattleHex(ESiegeHex::GATE_BRIDGE), false), [](const std::shared_ptr<const CObstacleInstance> obst)
```
IvanSavenko, line 1344: Perhaps rename to "hasWideMoat"? 
Same goes to other places with same variable name

server/CGameHandler.cpp
```diff
@@ -5317,11 +5235,10 @@ bool CGameHandler::handleDamageFromObstacle(const CStack * curStack, bool stackI
 {
 	if(!curStack->alive())
 		return false;
-	bool containDamageFromMoat = false;
 	bool movementStopped = false;
 	for(auto & obstacle : getAllAffectedObstaclesByStack(curStack, passed))
 	{
-		if(obstacle->obstacleType == CObstacleInstance::SPELL_CREATED)
+		if(obstacle->obstacleType == CObstacleInstance::SPELL_CREATED || obstacle->obstacleType == CObstacleInstance::MOAT)
```
IvanSavenko, line 5241: Is this check redundant now?
Other than spell created and now moats other obstacles are non-enterable - they block battlefield & can't affect stacks

lib/spells/effects/Moat.h
```diff
@@ -0,0 +1,41 @@
...
+#include "Obstacle.h"
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+namespace spells
+{
+namespace effects
+{
+
+class MoatPlacer : public Obstacle
```
IvanSavenko, line 22: - file named "Moat.h"
- copyright header says "Obstacle.h"
- class named "MoatPlacer"

Can you unify that?

config/spells/moats.json
```diff
@@ -0,0 +1,740 @@
...
+                "battleEffects":{
+                    "moat":{
+                        "type":"core:moat",
+						"hidden" : true,
+						"trap" : false,
+						"trigger" : true,
+						"triggerAbility" : "core:landMineTrigger",
+                        "dispellable" : true,
+						"removeOnTrigger" : true,
+                        "moatDamage" : 150,
```
IvanSavenko, line 223: Will moat damage work correctly if town is guarded by strong hero who can cast Mines with damage greater than 150?
See "Tower's Land Mines" section here: https://heroes.thelazy.net//index.php/Moat

config/gameConfig.json
```diff
@@ -164,7 +165,9 @@
 			// every 1 defense point damage influence in battle when defense points > attack points during creature attack
 			"defensePointDamageFactor": 0.025, 
 			// limit of damage reduction that can be achieved by overpowering defense points
-			"defensePointDamageFactorCap": 0.7
+			"defensePointDamageFactorCap": 0.7,
+			// if moving a double-wide creature one hex forward or backward should trigger obstacle, if it is obstacle which was on starting hex
+			"oneHexTriggersObstacles": false
```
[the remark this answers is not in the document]
rilian-la-te: No, this will have a following effect: if creature are moving to exactly one hex forward or backward, it will trigger obstacles, even it was present under creature before movement is starting. Only one hex, not any movement.

config/factions/castle.json
```diff
@@ -147,8 +147,7 @@
 			"horde" : [ 2, -1 ],
 			"mageGuild" : 4,
 			"warMachine" : "ballista",
-			"moatDamage" : 70,
-			"moatHexes" : [ 11, 28, 44, 61, 77, 111, 129, 146, 164, 181 ],
+			"moatAbility" : "core:spell.castleMoat",
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko we do not need to define such spell. If mod need one standard moat, it can use predefined moat. I changed spells only for having names exactly as H3.

lib/HeroBonus.h
```diff
@@ -11,6 +11,7 @@
 
 #include "GameConstants.h"
 #include "JsonNode.h"
+#include "battle/BattleHex.h"
```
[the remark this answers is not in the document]
rilian-la-te: I thought about this, but not in this PR, this PR is only for moats.

lib/HeroBonus.h
```diff
@@ -684,11 +685,11 @@ struct BonusLimitationContext
 class DLL_LINKAGE ILimiter
 {
 public:
-	enum EDecision {ACCEPT, DISCARD, NOT_SURE};
+	enum class EDecision : uint8_t {ACCEPT, DISCARD, NOT_SURE};
```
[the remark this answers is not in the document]
rilian-la-te: Small optimization.

config/spells/moats.json
```diff
@@ -0,0 +1,740 @@
+{
+    "castleMoatTrigger" :
+	{
+		"targetType" : "CREATURE",
+		"type": "ability",
+		"name": "Moat",
+		"school": {},
+		"level": 3,
+		"power": 10,
```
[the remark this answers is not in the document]
rilian-la-te: Yes, it is damage portion of land mine spell separated. So, it should be level 3 and have 10 damage per spell power. This trigger is used both in Land Mine spell and in tower mines.

config/spells/moats.json
```diff
@@ -0,0 +1,740 @@
+{
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko abilities can either be magical or non-magical. So, maybe, but maybe not in this PR?

server/CGameHandler.cpp
rilian-la-te: @IvanSavenko what about Hota Dunes obstacles? It should not be a spell created AFAIK, but should be passable and should stop movement.

server/CGameHandler.cpp
```diff
@@ -6435,6 +6332,21 @@ void CGameHandler::runBattle()
...
 
+	//Moat should be initialized here, because only here we can use spellcasting
+	if (gs->curB->town && gs->curB->town->fortLevel() >= CGTownInstance::CITADEL)
+	{
+		auto keep = gs->curB->battleGetUnitsIf([=](const battle::Unit * unit)
+		{
+			return unit->getPosition() == BattleHex::CASTLE_CENTRAL_TOWER;
+		});
+		const auto * h = gs->curB->battleGetFightingHero(BattleSide::DEFENDER);
+		const auto * actualCaster = h ? static_cast<const spells::Caster*>(h) : static_cast<const spells::Caster*>(keep.at(0)); //make keep a caster for moat, if no defender is present
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko Tower's landmines should be affected by defender hero, so base for proxy casting a land mines should be a hero. For unit - yes, we can modify MoatCasterProxy to be functional without a unit.

config/spells/moats.json
rilian-la-te: Yes, and it is a purpose of previous check for hero to have a caster. So, if hero is present, mines will have more than 150 damage if caster can have more effect power and effect level. You cannot have less than moatDamage, but can have more. Only one case which I did not check - how sorcery of defender actually affects tower's landmines in OH3.

config/spells/moats.json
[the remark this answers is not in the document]
rilian-la-te: Forr moat triggers - only to pass validation. Only for landmine and firewall it is actually meaningful.

lib/spells/effects/Moat.cpp
rilian-la-te: @IvanSavenko what are you mean? Can you explain more?

config/spells/moats.json
rilian-la-te: In this PR this always affects damage, so, level 1 Solmyr will have 157 mines damage.

config/factions/castle.json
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko you can see, than conflux moat are using castle spell, and mods can do this too.

config/spells/moats.json
rilian-la-te: FizMig says than sorcery is affect mines like spell, so, I think it is safe to say than we work most likely correct.

dydzio0614: test in original H3 with solmyr?

lib/HeroBonus.h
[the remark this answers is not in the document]
IvanSavenko: Yep. Agreed.
Planning to do this yourself or should I take over?

config/spells/moats.json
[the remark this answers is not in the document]
IvanSavenko: But this is Castle moat. Not Tower. Probably better to set to 0 in all objects other than Tower.

config/spells/moats.json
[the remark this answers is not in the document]
IvanSavenko: Maybe. Up to you. Wanted to reduce this file to more readable state.

lib/HeroBonus.h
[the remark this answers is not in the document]
rilian-la-te: I will do it with a next series of spell fixes (there is a couple of mechanical bugs there, Orb of Vulnerability, level 20 Thorgrim to be mentioned).

lib/spells/effects/Moat.cpp
rilian-la-te: @IvanSavenko there is only 3 static properties there. Notable:
```c++
		obstacle.casterSide = BattleSide::DEFENDER; // Moats are always cast by defender
		obstacle.passable = true; //Moats always passable
		obstacle.nativeVisible = false; //Moats is invisible for native terrain
```

All other is set by config, yes. Almost half of properties is shared for each obstacle, but I do not know how effect instances work for each spell and for each caster. Is there an object instance for each such pair - your suggestion possible. If not and effect loads from config each time - in theory then we can place obstacles with foreign properties.

config/gameConfig.json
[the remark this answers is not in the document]
rilian-la-te: So, example. I tested this with Luna both in OH3 and in this branch. So, if water elementals is placed on Fire Wall on their front hex, they will take damage. If they move will move exactly one hex forward, they will take no damage in OH3, even if they stopped on a fire wall. I consider it as OH3 bug and added option exactly for such cases (with option water elementals will burn after stopping of fire wall regardless if they started on same patch of fire wall or not). But if they move 2 or more hexes, each patch will affect them once, as should be.

config/spells/moats.json
rilian-la-te: @dydzio0614 I do not know a cheat to force build all buildings for AI in OH3. Do you know some?

rilian-la-te:
I will do only one force-push, so, I want to mark as done this features which is done locally.

server/CGameHandler.cpp
rilian-la-te: done

lib/battle/CObstacleInstance.h
rilian-la-te: done

lib/GameSettings.h
```diff
@@ -54,6 +54,7 @@ enum class EGameSettings
 	TEXTS_TERRAIN,
 	TOWNS_BUILDINGS_PER_TURN_CAP,
 	TOWNS_STARTING_DWELLING_CHANCES,
+	ONE_HEX_TRIGGERS_OBSTACLES,
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/NetPacks.h
```diff
@@ -424,15 +425,14 @@ struct DLL_LINKAGE PlayerReinitInterface : public CPackForClient
...
+	RemoveBonus(GiveBonus::EType Who = GiveBonus::EType::HERO)
 		:who(Who)
 	{
 	}
 
 	void applyGs(CGameState * gs);
 
-	enum { HERO, PLAYER, TOWN };
-	ui8 who; //who receives bonus, uses enum above
+	GiveBonus::EType who; //who receives bonus, uses enum above
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/NetPacks.h
```diff
@@ -349,15 +349,16 @@ struct DLL_LINKAGE SetAvailableHeroes : public CPackForClient
...
+	GiveBonus(EType Who = EType::HERO)
 		:who(Who)
 	{
 	}
 
 	void applyGs(CGameState * gs);
 
-	enum { HERO, PLAYER, TOWN };
-	ui8 who = 0; //who receives bonus, uses enum above
+	EType who = EType::HERO; //who receives bonus, uses enum above
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/spells/AbilityCaster.cpp
```diff
@@ -32,10 +31,11 @@ AbilityCaster::~AbilityCaster() = default;
 int32_t AbilityCaster::getSpellSchoolLevel(const Spell * spell, int32_t * outSelectedSchool) const
 {
 	auto skill = baseSpellLevel;
+	const auto & unit = dynamic_cast<const battle::Unit*>(actualCaster);
```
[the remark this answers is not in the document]
rilian-la-te: done

config/spells/moats.json
dydzio0614: it's called map editor xD

config/spells/moats.json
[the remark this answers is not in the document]
rilian-la-te: Set all values to 0, done

config/spells/moats.json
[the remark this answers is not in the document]
rilian-la-te: Abilities should be allowed to have same config as spells, so I do not know currently how to do abilities better. I wanted to reduce this file too, but strict validation is probably better than smaller file.

config/spells/moats.json
rilian-la-te: @dydzio0614 in OH3 sorcery of level 1 solmyr also gives mines 157 dmg) so, I did correctly.

server/CGameHandler.cpp
rilian-la-te: @IvanSavenko there is one difference - moats are static, but Hota dunes are random, Moat for now cannot do that.

server/CGameHandler.cpp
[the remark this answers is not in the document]
rilian-la-te: done

server/CGameHandler.cpp
rilian-la-te: Done (but not for Hota dunes).

config/gameConfig.json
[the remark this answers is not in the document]
rilian-la-te: I do not know if it is H3 bug or H3 feature. I consider it a bug, but someone will consider it a feature. Like VCMI spellcaster creatures trigger obstacles, and H3 ones is not trigger obstacles. I can encode this behaviour also in this switch and set it default to true, if desired.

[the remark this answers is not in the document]
IvanSavenko: IMO both are clearly a bug and should use fixed behavior by default.
Not against having an option to toggle old behavior, but not required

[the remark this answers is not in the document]
dydzio0614: @IvanSavenko i would be cautious about this, due to pvp people being used to abuse this kind of stuff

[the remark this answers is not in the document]
dydzio0614: stuff like "standing in front of gate as attacker prevents it from opening" might technically also be a bug as with proper army + destroying own catapult you can shoot/spellcast strong defenders to death, but it is already required for various singleplayer maps etc.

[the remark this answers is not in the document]
IvanSavenko: Blocking gates is clearly an intended feature. One that has potential exploit - yes, maybe. But same can be said about many other fearues.

While issues mentioned here are clearly bugs.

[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko changed to your second string.

lib/spells/effects/Moat.cpp
rilian-la-te: @IvanSavenko you think than creating template obstacle is better than setting properties from effect always?

IvanSavenko: Template obstacle that *will be directly read from json*
Currently you have (effectively) duplicated code:
1) On loading - code that reads json into Moat class members
2) On casting - Code that copies Moat members into Obstacle members

If it is possible (and makes sense logically) to set it up in a way that:
1) On loading - code that reads json into Obstacle that will be stored as member of Moat. This code can be moved to Obstacle class so Obstacle will be responsible for own state.
2) On casting - code that copies existing obstacle and modifies only necessary fields (e.g. position)

That being said, not insisting on this change.

rilian-la-te: @IvanSavenko in theory, I can use serializeJson method of CObstacleInstance to get a template obstacle, but then we will break mod compatibility (because any obstacle placing spell should be rewritten to use template obstacle, and not use effect settings).

rilian-la-te: @IvanSavenko this is ObstacleSideOptions, but used as obstacle. I think maybe someday we will extend moat class to attackers, and then we also will need side options (now ObstacleSideOptions is used only to store animation).

rilian-la-te: To not to become https://github.com/vcmi/vcmi/pull/363/files, I prefer not to completely redesign obstacle class. Maybe later, but not in this PR.

IvanSavenko (approved):

config/factions/conflux.json
```diff
@@ -152,8 +152,7 @@
 			"mageGuild" : 5,
 			"primaryResource" : "mercury",
 			"warMachine" : "ballista",
-			"moatDamage" : 70,
-			"moatHexes" : [ 11, 28, 44, 61, 77, 111, 129, 146, 164, 181 ],
+			"moatAbility" : "core:spell.castleMoat",
```
misiokles, line 155: why not `"moatAbility" : "core:spell.confluxMoat",` only `"moatAbility" : "core:spell.castleMoat",`?

rilian-la-te: Because castle and conflux moats is identical in every aspect, even name. Initially I want to do 4 type of moat - castle, fortress, inferno, and tower. All other are defined purely for names only.

config/factions/dungeon.json
```diff
@@ -148,8 +148,8 @@
 			"mageGuild" : 5,
 			"primaryResource" : "sulfur",
 			"warMachine" : "ballista",
-			"moatDamage" : 90,
-			"moatHexes" : [ 11, 28, 44, 61, 77, 111, 129, 146, 164, 181 ],
+			"moatAbility" : "core:spell.dungeonMoat",
```
IvanSavenko, line 151: Missed this during review, but preferred form here is
`"moatAbility" : "dungeonMoat",`
1) No need to specify mod ID for requests within same mod
2) No need to specify identifier type (spell) for requests that only accept specific type

lib/CTownHandler.cpp
```diff
@@ -894,6 +894,11 @@ void CTownHandler::loadTown(CTown * town, const JsonNode & source)
 		town->namesCount += 1;
 	}
 
+	VLC->modh->identifiers.requestIdentifier(source["moatAbility"], [=](si32 ability)
+	{
+		town->moatAbility = SpellID(ability);
+	});
+
```
IvanSavenko, line 901: Preferred form here is
`VLC->modh->identifiers.requestIdentifier( "spell", source["moatAbility"], [=](si32 ability)`
Code only accepts spells, so you can specifically request for spell identifier

config/factions/dungeon.json
rilian-la-te: Okay, I just thought than more verbose is better for modmakers.
Will do short form later.

IvanSavenko: Will do this myself soon - partially to get rid of weird and illogical error messages when provided ID has unexpected form

rilian-la-te: @IvanSavenko why we cannot accept all form of valid ID?

lib/CTownHandler.cpp
IvanSavenko: I think we don't?
`modh` however is very different from other handlers/services, so might need different access model than services

config/factions/dungeon.json
IvanSavenko: Will fix that part soon, so all valid ID's should always be accepted.

As to why we have it now this way - probably because nobody bothered to fix it?
My initial design was to have minimalistic ID's - e.g. don't specify ID type unless required & don't specify target mod unless required.

Specifying target mod scope was added because some mods had objects with same ID. And this is only scenario when you need to specify target mod to resove ID.
Specifying ID type is only necessary in some generic cases like bonuses. All other usages of ID work without type.

lib/CTownHandler.cpp
IvanSavenko: GeneralText is (luckily) quite easy one - all we need to do is to provide "translator" service that will call translate() method of GeneralText.

Mod Handler however is much more massive and way more integrated into engine.
Perhaps have it provide multiple "services"? E.g. "identifiers" service, mod information services, and maybe others. At least for now can be done without massive refactoring - meaning, make CModHandler implement multiple interfaces for all provided services.

That being said, extracting identifiers handling from ModHandler might not be a bad idea.

krs0:
A test map for OH3 moats.

[A Moat Test.zip](https://github.com/vcmi/vcmi/files/11182819/A.Moat.Test.zip)

PR #1707, opened 2023-03-17
Fix commander jousting

rilian-la-te:
After #1518 commander jousting was broken (I forgot to add 5% to him, so, it was effectively not working).
Fix this.


rilian-la-te:
@IvanSavenko we does not need to wait a CI here, this PR does not change any code (only JSON).

PR #1716, opened 2023-03-18
TotalValue: fix typo

rilian-la-te:
@kambala-decapitator please check if it fixes bug for you.

### DIFF
--- a/lib/HeroBonus.cpp
+++ b/lib/HeroBonus.cpp
@@ -455,7 +455,9 @@ int BonusList::totalValue() const
 		int indepMax = std::numeric_limits<int>::min();
 	};
 
-	auto percent = [](int base, int percent) -> int {return (base * (100 + percent)) / 100;};
+	auto percent = [](int64_t base, int64_t percent) -> int {
+		return static_cast<int>(vstd::clamp((base * (100 + percent)) / 100, std::numeric_limits<int>::min(), std::numeric_limits<int>::max()));
+	};
 	std::array <BonusCollection, Bonus::BonusSource::NUM_BONUS_SOURCE> sources = {};
 	BonusCollection any;
 	bool hasIndepMax = false;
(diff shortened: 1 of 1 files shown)


rilian-la-te:
Now it breaks every thing, will fix this evening

rilian-la-te:
UPD: looks likely everything is OK

DjWarmonger (approved):

lib/HeroBonus.cpp
```diff
@@ -493,45 +495,33 @@ int BonusList::totalValue() const
...
+			vstd::amax(any.indepMax, percent(src.indepMax, src.percentToSource));
 	}
 	any.base = percent(any.base, any.percentToBase);
 	any.base += any.additive;
 	auto valFirst = percent(any.base ,any.percentToAll);
 
-	if(hasIndepMin && hasIndepMax)
-		assert(any.indepMin < any.indepMax);
+	if(hasIndepMin && hasIndepMax && any.indepMin < any.indepMax)
+		any.indepMax = std::min(any.indepMax, any.indepMin);
```
IvanSavenko, line 514: Redundant check?
Since conditional includes `any.indepMin < any.indepMax` then std::min is redundant - you can simply use `any.indepMax = any.indepMin`

Also not sure what this supposed to do... 
If this code should handle case where min > max then it won't work due to condition

rilian-la-te: @IvanSavenko yes, this code will handle such changes, because now INDEPENDENT_MIN means AT_MOST, and INDEPENDENT_MAX means AT_LEAST (and it was be from introduction of this properties and used in all mods).

rilian-la-te: removed std::max

rilian-la-te: Example: recanter's cloack from (master branch)[https://github.com/vcmi/vcmi/blob/master/config/artifacts.json#L1192]
```json
	"recantersCloak":
	{
		"index" : 83,
		"type" : ["HERO"],
		"bonuses": [
			{
				"type" : "BLOCK_MAGIC_ABOVE",
				"val" : 2,
				"valueType" : "INDEPENDENT_MIN",
				"propagator": "BATTLE_WIDE"
			}
		]
	},
```

But we cleanly have blocked spells from level 3, 4, 5, and not 1 and 2.

rilian-la-te: Angel wings and Fly spell:

```json
	"angelWings":
	{
		"bonuses" : [
			{
				"type" : "FLYING_MOVEMENT",
				"val" : 0,
				"valueType" : "INDEPENDENT_MIN"
			}
		],
		"index" : 72,
		"type" : ["HERO"]
	},
```
and
```json
				"effects" : {
					"fly" : {
						"type" : "FLYING_MOVEMENT",
						"duration" : "ONE_DAY",
						"val" : 40,
						"valueType" : "INDEPENDENT_MIN"
					}
				}
```
So, if you cast basic Fly, then angel wings will override it, because FLYING_MOVEMENT can be AT MOST 0, not higher.

IvanSavenko (approved):

kambala-decapitator:
@rilian-la-te yes, now it's correct again, thanks!

Issue #1282, opened 2022-12-24
shipyard can't place boat on the water


kambala-decapitator:
![Screenshot 2022-12-23 at 23 14 45](https://user-images.githubusercontent.com/1557784/209426886-0e2c5570-036d-4b75-816e-d3273be9dc07.png)

Log files: (sorry for huge logs, was playing for a couple of hours since map creation)
[logs.tar.gz](https://github.com/vcmi/vcmi/files/10298204/logs.tar.gz)

Saves contain initial map state as well as the latest one:
[save.tar.gz](https://github.com/vcmi/vcmi/files/10298205/save.tar.gz)

In the latest state my barbarian hero stands right near the shipyard.

My part of map doesn't have any monoliths / dungeons, therefore it's quite unfortunate. (and currently no free boat on the map to summon, that's what my hero was trying to do)

IvanSavenko:
This is RMG map, correct?
I think shipyard in such position would not function in H3? So the issue is caused by RMG shipyard placement?

kambala-decapitator:
> So the issue is caused by RMG shipyard placement?

correct, that's why the issue has respective label

IvanSavenko:
Ah, missed that label somehow, sorry

DjWarmonger:
Actually bug comes from the fact that for some reason RMG thinks shipyard's border is like this:

```
#####
#____#
##_##
_###_
```
The lowermost tile is supposed to touch water, but the real shape of Shipyard is:

```
#####
#___#
#####
```

Not sure where does this come from, does RMG object have any additional tile @Nordsoft?

PR #1718, opened 2023-03-18
Fix shipyard border calculation for RMG

DjWarmonger:
Fixes https://github.com/vcmi/vcmi/issues/1282


rilian-la-te:
@DjWarmonger please, recheck for cities too.

DjWarmonger:
Towns are never placed near the water deliberately, they don't need to have place for a boat.

Nordsoft91 (approved):

Issue #939, opened 2022-09-18
Rmg generates useless shipyard


Nordsoft91:
<img width="297" alt="Снимок экрана 2022-09-18 в 16 15 01" src="https://user-images.githubusercontent.com/9308612/190901587-1d824d2c-d6d5-4404-947c-1c89ea24a5e7.png">

Rmg parameters: Using random seed: 1663503241
INFO global [0x70000e21c000] - RMG map: 144x144, NO underground
INFO global [0x70000e21c000] - RMG settings: players -1, teams -1, computer players -1, computer teams -1, water -1, monsters -2
INFO global [0x70000e21c000] - RMG template name: Midnight Mix 48
TRACE global [0x70000e21c000] - Player config:
TRACE global [0x70000e21c000] - Player 0: human only
TRACE global [0x70000e21c000] - Player 1: human only

kambala-decapitator:
is it useless because boat can't spawn at the correct location?

kambala-decapitator:
seems it's the same as #1282?

DjWarmonger:
It's not the same. Maybe just Shipyard shouldn't be placed next to very small lakes?

This might be the same case, or not.
<img width="108" alt="Shipyard near rocks" src="https://user-images.githubusercontent.com/2566990/226131435-7176358d-a25b-4118-ad1c-0f196b2b60e2.png">

PR #1720, opened 2023-03-18
Don't place shipyard or boats at very small lakes

DjWarmonger:
Fixes cases like this: https://github.com/vcmi/vcmi/issues/939

### DIFF
--- a/lib/rmg/WaterProxy.cpp
+++ b/lib/rmg/WaterProxy.cpp
@@ -142,6 +142,13 @@ RouteInfo WaterProxy::waterRoute(Zone & dst)
 				dst.areaPossible().subtract(lake.neighbourZones[dst.getId()]);
 				continue;
 			}
+
+			//Don't place shipyard or boats on the very small lake
+			if (lake.area.getTiles().size() < 25)
+			{
+				logGlobal->info("Skipping very small lake at zone %d", dst.getId());
+				continue;
+			}
 						
 			int zoneTowns = 0;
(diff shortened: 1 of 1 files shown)


lib/rmg/WaterProxy.cpp
```diff
@@ -142,6 +142,13 @@ RouteInfo WaterProxy::waterRoute(Zone & dst)
 				dst.areaPossible().subtract(lake.neighbourZones[dst.getId()]);
 				continue;
 			}
+
+			//Don't place shipyard or boats on the very small lake
```
nullkiller, line 146: Will it try to put any treasures there?

DjWarmonger: Yep, but it's irrelevant to the gameplay.

Such small lakes are usually placed right next to land guard between the zones.

lib/rmg/WaterProxy.cpp
```diff
@@ -142,6 +142,13 @@ RouteInfo WaterProxy::waterRoute(Zone & dst)
...
+			if (lake.area.getTiles().size() < 25)
```
Nordsoft91, line 147: Are we sure this won't lead to unplayable maps?

DjWarmonger: In an edge case it should fall back to Monolith conection, but I'll double check.

DjWarmonger: Yep, the default route is invalid - so if RMG won't find any other connection, it will just place Monolith pair.

PR #1722, opened 2023-03-19
vcmi: fix first aid regression

rilian-la-te:
First aid specialists was count a percentage not from full value, but from value - 10.
Fix it by removal of bonus from First Aid Tent and adding 10 pt to each skill level.

Checked with specialist and non-specialist, all is working according to description.


rilian-la-te:
@IvanSavenko @nullkiller please, look)

Issue #1367, opened 2023-01-04
Bug with Town Portal spell and Dungeon Town with Battle Scholar Academy


storageer:
**Describe the bug**
The game stops working after a hero Town Portals into a previously unvisited Dungeon Town, with a Battle Scholar Academy built, and gets a level-up from the gained experience.

**Game logs**
[Testing the Town Portál.zip](https://github.com/vcmi/vcmi/files/10347214/Testing.the.Town.Portal.zip)
[server_log.txt](https://github.com/vcmi/vcmi/files/10347202/server_log.txt)
[VCMI_Client_log.txt](https://github.com/vcmi/vcmi/files/10347203/VCMI_Client_log.txt)
[VCMI_Launcher_log.txt](https://github.com/vcmi/vcmi/files/10347206/VCMI_Launcher_log.txt)
[VCMI_Server_log.txt](https://github.com/vcmi/vcmi/files/10347207/VCMI_Server_log.txt)

**To Reproduce**
Needs
- One hero with access to the Town Portal spell, and less than 1000xp needed to reach a level-up
- One Dungeon town with a Battle Scholar Academy built, but not yet visited by said hero

Steps
1. Use the suitable hero to Town Portal into the Dungeon town
2. Receive a level-up prompt, pick any skill of choice (seemingly irrelevant to the bug)
3. Game hangs up when the hero attempts to leave the town, or another hero tries to trade with him/her

**Expected behavior**
The Hero should behave  without triggering an error, like it would happen if it were to visit the town on horseback

**Actual behavior**
Green text error message gets printed out and the game practically stops responding to any user input

**Did it work earlier?**
Unknown

**Screenshots**
![VCMI TP-Dungeon Server problem](https://user-images.githubusercontent.com/121976640/210644725-9729d793-2cbc-4290-94b8-5a1adac3c24a.png)

**Version**
 - OS: Windows 10
 - Version: VCMI 1.1.0 Release version

**Additional context**
I've created a makeshift map to help me better understand the bug I experienced, and decided to attach it in case it turns out to be useful.

kambala-decapitator:
also sound of running horse plays indefinitely

one of the threads is stuck:

```
    frame #0: 0x00007ff802c6f3ea libsystem_kernel.dylib`__psynch_cvwait + 10
    frame #1: 0x00007ff802ca9a6f libsystem_pthread.dylib`_pthread_cond_wait + 1249
[...]

PR #1728, opened 2023-03-20
Fix town portal query when it calls level up query

Adriankhl:
Fixes https://github.com/vcmi/vcmi/issues/1367

@IvanSavenko You mean something like this? Now every `CGenericQuery` callback is called in `onRemoval`. I did consider to make this behavior optional, but I think the class is safer in this form.


IvanSavenko:
@Adriankhl yes.
Provided that this works & does not breaks other adventure map spells (I think this query is only used by them)

Adriankhl:
Actually seems like this is the only place which uses `CGenericQuery`, as long as this works then it should be good.

PR #1729, opened 2023-03-20
Fixes for battle UI regressions

IvanSavenko:
- Fixed delays between animations, leading to flickering stack numbers
- Fixed assertion failure (and potential hang/crash) on playing shooting animations at huge speeds
- Show "press any key to skip" message while battle opening music is playing


rilian-la-te:
@IvanSavenko it is possible to do opening battle music as in oh3?

rilian-la-te:
I will provide a screen cast from h3 (but it may be another HD mod improvement)

PR #1731, opened 2023-03-20
CStack: make unicorns ability multiplicative

### LINKED ISSUE #1730: According to FizMig, Unicorn aura should be multiplicative
**Describe the bug**
self-explanatory

**Game logs**
Not applicable

**To Reproduce**
Fight Rampart army with Dwarves adjacent to unicorns (or with Resistance hero)

**Expected behavior**
Magic resistance and unicorn aura should  be multiplicative

**Actual behavior**
They are additive

**Did it work earlier?**
In 1.1 - not, before I do not know.

**Screenshots**
Not applicable

**Version**
 - OS: Linux
 - Version: develop

rilian-la-te:
Fixes #1730

### DIFF
--- a/lib/CStack.cpp
+++ b/lib/CStack.cpp
@@ -93,7 +93,7 @@ ui32 CStack::level() const
 
 si32 CStack::magicResistance() const
 {
-	si32 magicResistance = IBonusBearer::magicResistance();
+	auto magicResistance = IBonusBearer::magicResistance();
 
 	si32 auraBonus = 0;
 
@@ -102,10 +102,11 @@ si32 CStack::magicResistance() const
 		if(one->unitOwner() == owner)
 			vstd::amax(auraBonus, one->valOfBonuses(Bonus::SPELL_RESISTANCE_AURA)); //max value
 	}
-	magicResistance += auraBonus;
-	vstd::amin(magicResistance, 100);
+	vstd::abetween(auraBonus, 0, 100);
+	vstd::abetween(magicResistance, 0, 100);
+	float castChance = (100 - magicResistance) * (100 - auraBonus)/100.0;
 
-	return magicResistance;
+	return static_cast<si32>(100 - castChance);
 }
 
 BattleHex::EDir CStack::destShiftDir() const


lib/CStack.cpp
```diff
@@ -102,10 +102,10 @@ si32 CStack::magicResistance() const
 		if(one->unitOwner() == owner)
 			vstd::amax(auraBonus, one->valOfBonuses(Bonus::SPELL_RESISTANCE_AURA)); //max value
 	}
-	magicResistance += auraBonus;
+	magicResistance *= 100.0/auraBonus;
 	vstd::amin(magicResistance, 100);
 
-	return magicResistance;
+	return static_cast<si32>(magicResistance);
```
IvanSavenko, line 108: Are you sure that formula is correct?
Assuming magicResistance = 40 and auraBonus = 20
This line will give 40 * 100 / 20 = 200

Additive bonus as before should result in 60
Presumably correct multiplicative bonus should give (100 - (100 - 40) * (100 - 20)) = 52

Also, I would rather keep formula as simple as possible to make code more clear, even if this leads to few extra math operations.

rilian-la-te: @IvanSavenko thanks for catching. Updated formula to 100/(100-auraBonus), and added abetween for auraBonus.
What I need to do to keep math more simple? Separate auraBonus calculations altogether?

rilian-la-te: I will fix it more ASAP.

rilian-la-te: So, it should work like this:
```100 - ((100 - magicResistance) * (100 - auraBonus)/100.0)```

Because this is what FizMig says: 
```
 Шанс срабатывания заклинания и шанс блокирования – не одно и то же.
Например, если рядом с единорогом находится Кентавр, то шанс срабатывания вражеского Проклятья по нему снижается со 100% до 80%. А если рядом с
единорогом находится Гном, то шанс срабатывания снижается с 80% до 80%*0,8=64%, т.е. сопротивление Гнома становится равным 36%.
```

rilian-la-te: Did it as FizMig.

PR #1736, opened 2023-03-21
vcmi: fix target condition regression

rilian-la-te:
And we also need to backport it to develop ASAP.

Fixes: https://forum.vcmi.eu/t/some-graphic-errors/5935/13 (casting bless on undead).


lib/spells/TargetCondition.cpp
```diff
@@ -519,7 +516,7 @@ void TargetCondition::loadConditions(const JsonNode & source, bool exclusive, bo
 
 			CModHandler::parseIdentifier(keyValue.first, scope, type, identifier);
 
-			std::shared_ptr<TargetConditionItem> item = itemFactory->createConfigurable(scope, type, identifier);
+			item = itemFactory->createConfigurable(scope, type, identifier);
```
kambala-decapitator, line 519: why was there no warning about shadowing a variable?

rilian-la-te: @IvanSavenko agree.

PR #1740, opened 2023-03-21
vcmi: russian

rilian-la-te:


rilian-la-te:
@kambala-decapitator this PR should target both, but I do not know how to make a PR than targets both.

kambala-decapitator:
either do in separate PRs or target just beta if not a blocker - beta will be merged in develop after the release

rilian-la-te:
@kambala-decapitator second PR for beta done, feel free to merge into beta.

IvanSavenko:
Ah, missed that this was for develop.
@rilian-la-te depending on need/activity in beta branch I'll merge beta into develop from time to time, so no need for duplicated PR's for both

PR #1753, opened 2023-03-23
Translate new launcher texts

dydzio0614:


IvanSavenko:
Zero objections to merge, but keep in mind that Launcher might still get few string changes soon

Issue #1592, opened 2023-02-18
Battle UI isn't locked during spellcast when it should be


dydzio0614:
**Describe the bug**
Tested on ripple death spell - you can perform action, like move unit while spell effect not fully applied

**To Reproduce**
1. Reproducible at least on debug builds - start battle with hero that has ripple death
2. cast ripple death and then move unit - there is no clientside block for movement action for unit movement even if spell will kill everyone

**Expected behavior**
You cannot move unit if spell will allow everybody to die, and in H3 UI is blocked until spell animation is complete etc.

**Actual behavior**
you can do whatever while spell animation is applying etc.

**Did it work earlier?**
Not sure

**Screenshots**
[video](https://streamable.com/a4tpwx)

**Version**
 - OS: Kubuntu 22.04, debug build
 - Version: current develop

**Additional info**
For some reason delay between spellbook spell click and spell animation start seems to be bigger in H3, but it might be due to debug build stuff

PR #1758, opened 2023-03-24
Deactivate activeStack during spellCast

Adriankhl:
Fixes https://github.com/vcmi/vcmi/issues/1592


IvanSavenko:
Cursor is not update correctly and stays at cast/spellbook form if you don't move mouse.
Probably:
- when spellcast starts cursor should be changed to blocked
- when spellcast is over call BattleInterface::activateStack() instead which should also update cursor

Adriankhl:
@IvanSavenko I think it works now

PR #1759, opened 2023-03-24
Show potential kill amount in attack tooltip

IvanSavenko:
Implements #857 

Only amount of kills is shown, without potential retaliation information - that part is a bit too complex and needs to account for multiple potential cases - e.g. blind that blocks retaliation, stack elimination, ranged retaliation possibility, etc.


rilian-la-te:
@IvanSavenko why replacing pair to struct with 2 elements?

rilian-la-te:
I initially thought than this is required to store more than 2 elements (and because tuples is bad before c++20), but change more than 300 LOC only for readability - .....

Maybe we at least can add something more into this struct?

IvanSavenko:
> but change more than 300 LOC only for readability - .....

IMO readability is the most important metric of code.
Long but more readable code is better than compact code
Slightly slower but more readable code is better than fast code
More readability is better even if it adds a bit more noise into git blame.

> Maybe we at least can add something more into this struct?

Considering its usage - average() method might be useful. Other than that - not really.

IvanSavenko:
Maybe. But not now.
Don't see a good way/location to add that information right now and don't want to untangle all potential chains as to whether stack can retaliate.

Issue #1455, opened 2023-01-17
Allow spellcasters with multiple non-random spells


IvanSavenko:
From discussions on Slack, recent change to Demon Summoning broke some mods.

Agreed solution:
- "SPELLCASTER" bonus should be propagated, e.g. from hero specialty.
- For a creature with multiple "SPELLCASTER" bonuses there should be a way to disable random selection (e.g. Genie/Faerie Dragon)
- If creature with multiple non-random "SPELLCASTER" bonuses attempts to cast spell, vcmi should pick first spell that can be cast at the selected target.

PR #1764, opened 2023-03-24
Implemented support for multi-spell casters

IvanSavenko:
Implements #1455

Now SPELLCASTER bonus spell selection will work as follows:
- if bonus has no "addInfo" field set (or with non-positive value), this spell will be always added as possible for casting
- if bonus has positive "addInfo" field then game will select one spell among all such bonuses, using "addInfo" as weight (e.g. Faerie Dragon)

Unit will be able to cast all possible spells, provided that target is suitable. If multiple spells can be cast on a target, vcmi will pick one of those spells for casting.

So it is now possible to (for example) give to a single creature buff spell that can be cast on allies, damaging spell that can be cast on enemies and rising spell that can be cast on corpses.

### DIFF
--- a/client/battle/BattleActionsController.cpp
+++ b/client/battle/BattleActionsController.cpp
@@ -480,6 +480,9 @@ bool BattleActionsController::actionIsLegal(PossiblePlayerBattleAction action, B
 		case PossiblePlayerBattleAction::SHOOT:
 			return owner.curInt->cb->battleCanShoot(owner.stacksController->getActiveStack(), targetHex);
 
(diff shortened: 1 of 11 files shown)


Alucard648:
Will it work like spellcasters in H4+ games?

IvanSavenko:
> Will it work like spellcasters in H4+ games?

Not exactly.
You still can't select spell if multiple spells are available. There is not "spellbook" to pick spells from for creature manually. Also, all creatures have limited number of "casts" and don't use "mana".

PR #1767, opened 2023-03-25
Properly shutdown battle thread

Adriankhl:
Currently when I "Quite to Desktop" during battle, there is an error message:
```
vcmiserver: /usr/include/boost/thread/pthread/condition_variable_fwd.hpp:81: boost::condition_variable::~condition_variable(): Assertion `!posix::pthread_mutex_destroy(&internal_mutex)' failed.
Error: server failed to close correctly or crashed!
```

This PR ends the battle thread properly and forces the destructor of CGameHandler to wait for the thread to join.

I think this can go to the beta branch?

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -51,6 +51,8 @@
 #include "../lib/serializer/Cast.h"
 #include "../lib/serializer/JsonSerializer.h"
 #include "../lib/ScriptHandler.h"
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Just a heads up - will look into this one a bit later, once critical 1.2 stuff is done.

This may be a crash, but it only happens when user wants to quite the game anyway and fix is not trivial.

Issue #1733, opened 2023-03-21
Stack amount labels should update on hit


IvanSavenko:
**Describe the bug**
Creatures stack amount label updates immediately on animation start.
This leads to bug where number of creatures changes before target is hit. Especially noticeable with ranged attacks

**To Reproduce**
1. Start battle with shooters in army
2. Attack enemy creature with ranged attack
3. Look on enemy creature stack label

**Expected behavior**
Amount of creatures should update at the same time as hit effect plays

**Actual behavior**
Amount of creatures updates when attack starts

**Did it work earlier?**
Worked in 1.1, battle rewrite regression.

PR #1770, opened 2023-03-25
Process attack/attacked animation before dealing damage to units

Adriankhl:
Fixes https://github.com/vcmi/vcmi/issues/1733

### DIFF
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -748,11 +748,14 @@ void ApplyFirstClientNetPackVisitor::visitBattleStackMoved(BattleStackMoved & pa
 void ApplyFirstClientNetPackVisitor::visitBattleAttack(BattleAttack & pack)
 {
 	callBattleInterfaceIfPresentForBothSides(cl, &IBattleEventsReceiver::battleAttack, &pack);
+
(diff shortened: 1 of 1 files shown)


IvanSavenko:
Hmm... I've considered this, but won't such change lead to opposite effect - where stack labels updates way too late?
Can you check:
- Harpies (unupgraded unit) - will they correctly update their label before flying back if harpy dies in counter-attack?
- Vampires - will label update correctly if their life drain resurrects a unit?
- Attacking creatures from behind (forcing them to play rotation animation after hit)

If your change works - then great. Was expecting more difficulties with this one.

Adriankhl:
>     * Harpies (unupgraded unit) - will they correctly update their label before flying back if harpy dies in counter-attack?
> 
>     * Vampires - will label update correctly if their life drain resurrects a unit?

If my eyes are not blind (the time intervals are quite short :smiley: ), they work correctly.

>     * Attacking creatures from behind (forcing them to play rotation animation after hit)

I don't know what is the correct rotation animation, the unit under attacked rotate backward on hit then rotate back to the original orientation.

IvanSavenko:
Great then! Will recheck myself just in case soon and merge.

PR #1777, opened 2023-03-26
Fixes for various issues extracted from moats branch

rilian-la-te:
Needs to be merged into develop and (maybe) beta before #1706 will landed into develop.

@IvanSavenko @nullkiller @kambala-decapitator 

Fixes: https://bugs.vcmi.eu/view.php?id=3106

### DIFF
--- a/lib/battle/CBattleInfoEssentials.cpp
+++ b/lib/battle/CBattleInfoEssentials.cpp
@@ -47,10 +47,7 @@ std::vector<std::shared_ptr<const CObstacleInstance>> CBattleInfoEssentials::bat
 	else
 	{
 		if(!!player && *perspective != battleGetMySide())
-		{
-			logGlobal->error("Unauthorized obstacles access attempt!");
-			return ret;
-		}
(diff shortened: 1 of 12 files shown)


lib/battle/CBattleInfoEssentials.cpp
```diff
@@ -47,10 +47,7 @@ std::vector<std::shared_ptr<const CObstacleInstance>> CBattleInfoEssentials::bat
 	else
 	{
 		if(!!player && *perspective != battleGetMySide())
-		{
-			logGlobal->error("Unauthorized obstacles access attempt!");
-			return ret;
-		}
+			logGlobal->warn("Unauthorized obstacles access attempt, assuming massive spell");
```
IvanSavenko, line 50: In which scenarios could this happen?

rilian-la-te: @IvanSavenko when we ask if we can use expert dispel and there is a landmines placed by enemy and no other spell. Before this fix we will get "this spell will affects nothing" message. With this change mines are dispelled correctly.

IvanSavenko (approved):

PR #1790, opened 2023-03-28
Refactoring of hero specialty loading code

### LINKED ISSUE #1732: Specialty bonuses are not transferred to 3rd upgraded units.
**Describe the bug**
In OHC we don't have 3rd upgrades. But in Hota mod - yes. Pirates can be upgraded to Corsairs, Corsairs can be upgraded to Seadogs. Anabel's specialty are pirates, so for example, pirates should gain +1 to speed. But it works only for Pirates and Corsairs, but not for Sea Dogs. 
```
		"speed" : {
			"limiters" : { "parameters" : [ "hota.cove:pirate", true ] },
			"type" : "STACKS_SPEED",
			"val" : 1
		} 
```

**Game logs**
N/A

**To Reproduce**
1. Enable Hota mod
2. Choose any scenario with Anabel as starting hero.
3. Build Gunpowders Warehouses (3rd pirate upgrade building)

**Expected behavior**
Sea Dogs should gain +1 speed.

**Actual behavior**
Sea Dogs don't gain +1 speed.

**Did it work earlier?**
I guess not.

**Screenshots**
image
image
image

**Version**
 - OS: [ Windows]
 - Version: [daily builds]

**Additional context**
Possible workaround. Edit Anabel's specialty by using new bonus system, like:
```
		"speed" : {
			"limiters" : { "parameters" : [ "hota.cove:pirate", true ] },
			"type" : "STACKS_SPEED",
			"val" : 1
		},
			"speed2" : {
			"limiters" : { "parameters" : [ "hota.cove:seadog", false ] },
			"type" : "STACKS_SPEED",
			"val" : 1
		}
```

IvanSavenko:
Originally intended to fix bug from 1.2 milestone, however decided to cleanup this code and remove a lot of outdated and no longer used routines, so changes are now targeting 1.3 / develop

Changes:
- removed loading of old specialty types (now replaced with bonuses)
- cleaned up loading of "short" form of creature specialties
- added support for 3rd upgrades - vcmi will now track entire chain of upgrades instead of only direct upgrades
- Fixes #1732

### DIFF
--- a/lib/CHeroHandler.cpp
+++ b/lib/CHeroHandler.cpp
@@ -504,182 +504,60 @@ void CHeroHandler::loadHeroSkills(CHero * hero, const JsonNode & node) const
 	}
 }
 
-// add standard creature specialty to result
-void AddSpecialtyForCreature(int creatureID, std::shared_ptr<Bonus> bonus, std::vector<std::shared_ptr<Bonus>> &result)
-{
-	const CCreature &specBaseCreature = *VLC->creh->objects[creatureID]; //base creature in which we have specialty
-
-	bonus->limiter.reset(new CCreatureTypeLimiter(specBaseCreature, true));
-	bonus->type = Bonus::STACKS_SPEED;
-	bonus->valType = Bonus::ADDITIVE_VALUE;
-	bonus->val = 1;
-	result.push_back(bonus);
-
-	// attack and defense may differ for upgraded creatures => separate bonuses
-	std::vector<int> specTargets;
-	specTargets.push_back(creatureID);
-	specTargets.insert(specTargets.end(), specBaseCreature.upgrades.begin(), specBaseCreature.upgrades.end());
-
-	for(int cid : specTargets)
-	{
-		const CCreature &specCreature = *VLC->creh->objects[cid];
-		bonus = std::make_shared<Bonus>(*bonus);
-		bonus->limiter.reset(new CCreatureTypeLimiter(specCreature, false));
-		bonus->type = Bonus::PRIMARY_SKILL;
-		bonus->val = 0;
-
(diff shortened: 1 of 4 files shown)


rilian-la-te:
@IvanSavenko maybe add standard secondary skill and damage spell speciality too? But proper way, without deprecated bonus converter?

IvanSavenko:
In my to-do list, but originally I wanted slightly different format for such specialties, one that will be (somewhat) moddable and not hardcoded in c++.

Will do, but probably separately from this PR.

rilian-la-te:
@IvanSavenko but for anything moddable there is a bonuses)

IvanSavenko:
@rilian-la-te 
Maybe. My point is that I don't want to hardcode parameters.
For example, creature specialties bonuses grow to 100% at level 20. This part is hardcoded (if you use short form).

Short form for secondary specialties would look like `"skill" : "archery" `. And since all H3 specialties grow at 5% / level that part will end up hardcoded as well.

So I want to provide some way to define this hardcoded part in config, for example as game setting, and potentially - provide a way to override specialty parameters on per-hero level (not needed for H3, might be interesting for mods). Don't see much point in making this growth speed in form of a bonus though.

rilian-la-te:
@IvanSavenko but this growth speed is simple bonus updater, so, why we need to make short form for modders, even if they can use long form if they want customization? Short form is good for mods like "I want more heroes", but for more complex mods long form is preferred IMO.

IvanSavenko:
More code/json config -> more chances of typos, more code to maintain, more code to update if we decide to change something in a specialty format.

rilian-la-te:
@IvanSavenko this will be a problems for a modders, not for VCMI itself, because in H3 almost all specialities are short.

PR #1792, opened 2023-03-28
Fixes for beta - graphical artifacts

IvanSavenko:
- [x] Fix hero animation switching to idle for 1 frame between movement animation iterations
- [x] Fix stack amount boxes positioning to match H3
- [x] Fix stack amount boxes visibility during animations
- [x] Fix stack amount boxes visibility between animations (network lag? threading lag?)
- [x] Fix discovered crashes

@dydzio0614 are you happy now? xD

### DIFF
--- a/client/battle/BattleActionsController.cpp
+++ b/client/battle/BattleActionsController.cpp
@@ -689,14 +689,14 @@ void BattleActionsController::actionRealize(PossiblePlayerBattleAction action, B
 				if (action.spell() == SpellID::SACRIFICE)
 				{
 					heroSpellToCast->aimToHex(targetHex);
-					possibleActions.push_back({PossiblePlayerBattleAction::SACRIFICE, SpellID::SACRIFICE});
+					possibleActions.push_back({PossiblePlayerBattleAction::SACRIFICE, action.spell()});
 					owner.stacksController->setSelectedStack(targetStack);
 					return;
 				}
 				if (action.spell() == SpellID::TELEPORT)
 				{
 					heroSpellToCast->aimToUnit(targetStack);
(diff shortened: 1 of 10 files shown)


IvanSavenko:
Remaining issues with stack amount boxes should now be fixed.
Now their visibility is slightly more restricted than in H3, but should eliminate any flicker, for example when stack turn around.

rilian-la-te:
@IvanSavenko what is "more restricted"? Can you explain more?

rilian-la-te:
@IvanSavenko understood.

client/battle/BattleActionsController.cpp
```diff
@@ -689,14 +689,14 @@ void BattleActionsController::actionRealize(PossiblePlayerBattleAction action, B
...
-					possibleActions.push_back(PossiblePlayerBattleAction::SACRIFICE);
+					possibleActions.push_back({PossiblePlayerBattleAction::SACRIFICE, SpellID::SACRIFICE});
 					owner.stacksController->setSelectedStack(targetStack);
 					return;
 				}
 				if (action.spell() == SpellID::TELEPORT)
 				{
 					heroSpellToCast->aimToUnit(targetStack);
-					possibleActions.push_back(PossiblePlayerBattleAction::TELEPORT);
+					possibleActions.push_back({PossiblePlayerBattleAction::TELEPORT, SpellID::TELEPORT});
```
rilian-la-te, line 699: I am looking for modernizing teleport effect (there is a TODO there, and I think we need teleport variant which will be affected by moats etc.), so, any hardcoded spellID is something than I look to avoid.

IvanSavenko: Will remove one added by me (already available in `action` variable)

rilian-la-te: Thanks)

PR #1799, opened 2023-03-30
Several more fixes for beta branch

IvanSavenko:
- Fixed Roland's specialty to use Swordsmen instead of Griffins
- Blocked movement of war machines in tactics phase
- Fix initial value of sound/music labels in settings
- Removed largely unused signal handling, fixes build with musl c library


client/battle/BattleActionsController.cpp
```diff
@@ -525,6 +525,8 @@ bool BattleActionsController::actionIsLegal(PossiblePlayerBattleAction action, B
 	switch (action.get())
 	{
 		case PossiblePlayerBattleAction::CHOOSE_TACTICS_STACK:
+			return (targetStack && targetStackOwned && targetStack->Speed() > 0);
```
dydzio0614, line 528: plz make toggle to allow old behavior :P

IvanSavenko:
> @IvanSavenko what was wrong with musl?

It does not offers `backtrace` apparently.

dydzio0614:
@Adriankhl almost entirety of lua scripting was written by same guy who made bonus system and spent hundreds of hours on spells reworking, so I would be cautious about that

Issue #1632, opened 2023-03-04
ALT shortcut for showing Movement cost does not start to work unless you move the mouse


krs0:
**Describe the bug**
Pressing ALT does not start showing in status window the movement cost and remaining movement points unless mouse is also moved.

**Expected behavior**
Instantly when I press ALT I get the movement info shown in status bar. (like in HotA)

**Version**
 - OS: Windows
 - Version: develop

dydzio0614:
Not exactly bug IMO since it's extra feature and not obliged to work exactly same as in hota (general functionality is maintained anyway). Maybe could be interested "bug in extra feature" since lack of instant reaction is not very intuitive.

But your suggested behavior is small improvement anyway, so if can be fixed with small effort then why not.

PR #1805, opened 2023-03-31
Show movement cost immediately after pressing left alt key

Adriankhl:
Fixes https://github.com/vcmi/vcmi/issues/1632


krs0:
When I play with lefthand I use Right alt key. Can you add that one too please SDLK_RALT?

Issue #1800, opened 2023-03-30
At each start of turn AI player uses vcmieagles


krs0:
**Describe the bug**
At each start of AI player turn, it will do a vcmieagle. Can be seen in chat.

**To Reproduce**
1. Start a battle
2. end turn
3. look in chat

**Screenshots**
![image](https://user-images.githubusercontent.com/15194321/228864675-4e90e76c-e637-4ec3-9f11-1e1285cdc4e3.png)

**Version**
 - OS: windows
 - Version: beta

rilian-la-te:
@DjWarmonger do we can add some setting to NKAI to disable eagles for friendly/enemy?

DjWarmonger:
It's not the problem on AI side, just the way allies work.

dydzio0614:
i think nullkiller ai may crash without full map vibisibility, it doesn;t have exploration at all

Laserlicht:
Shouldn't the PR do just that? https://github.com/vcmi/vcmi/pull/1532

dydzio0614:
no, it was hiding "CHEATER!!!!!!" message only

krs0:
@Laserlicht VCMI developers use Slack for communication, if you want to join...
https://slack.vcmi.eu/

nullkiller:
AI will work without vcmieagles. But it will mostly do nothing because it has no routines to explore the map. Maybe it is time to port this thing from VCAI as well)

PR #1813, opened 2023-03-31
hide messages from AI

Laserlicht:
hides: https://github.com/vcmi/vcmi/issues/1800

Since I am not the expert in vcmi: can we do it this way? This hides all the messages from the AI.
Or are there messages that are relevant to the user (so that we have to filter for vcmieagles beforehand)?


dydzio0614:
I can see potential for some features related to AI using chat as part of interaction, but since these are probably far future and low priority it makes sense for AI chat messages getting muted for foreseeable future

IvanSavenko (approved):

PR #1816, opened 2023-03-31
Added LotR alternatives for vcmiarmy and vcmiexp

krs0:


rilian-la-te:
@krs0 why not the classics?

krs0:
? what classics?

dydzio0614:
IMO LotR cheat places should not be there in the first place and existing are not removed for compatibility

rilian-la-te:
@dydzio0614 why not? LoTR cheats is good and originates from WoG. Classics for @krs0: vcmigandalfwhite for exp and there is no classic for army.

krs0:
@rilian-la-te did not new those are from WoG.
I just invented vcmimithrandir :)
will change to that one.
@dydzio0614 I just made a quick change following what was existing so that it gets into 1.2

dydzio0614:
not against this change, but in general cheats should be intuitive easy to memorize to be useful, easter egg names are not very practical
too many aliases also may make a mess for players in a wiki

IvanSavenko:
Well... Strictly speaking, VCMI cheats are not from LotR but from Silmarillion. So neither Mithrandir nor Gandalf fit

krs0:
Ok, this was intended as a quick fun fix... maybe not my thing :)

All names seem to be invented for VCMI and no connection to other H3 versions. And like Ivan mentioned seem to be using Silmarillion.

In the Silmarillion, Gandalf is referred to as "Olórin"

You can discard the PR completely if not ok. No biggie.

PR #1818, opened 2023-03-31
config: CATAPULT_EXTRA_SHOTS should be hidden

rilian-la-te:
If it will be shown, it leads to annoying extra bonus for all stacks of heroes with Ballistics.


rilian-la-te:
If it will be shown, it leads to annoying extra bonus for all heroes with Ballistics.

PR #1821, opened 2023-04-01
CTownHandler: remove hardcoded Cove grail

rilian-la-te:
We have now exactly same definition in Json.
So, why have hardcode grail from mod?

### DIFF
--- a/lib/CTownHandler.cpp
+++ b/lib/CTownHandler.cpp
@@ -485,15 +485,6 @@ void CTownHandler::addBonusesForVanilaBuilding(CBuilding * building) const
 		{
 			b = createBonus(building, Bonus::MORALE, +1);
 		}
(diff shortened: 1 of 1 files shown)


krs0:
So lodestar now is equivalent to HotA?
_[Lodestar](https://heroes.thelazy.net/index.php/Lodestar) - [Grail](https://heroes.thelazy.net/index.php/Grail) building of Cove. Brings 5000 gold per day, multiplies the growth of creatures by 1.5 per week (rounded down). Makes any terrain native to Cove (movement without penalty, +1 to attack, defense and speed, see and ignore mines and see quicksand). Cove's siege remains on swamp_

rilian-la-te:
Should be. Will try myself.

IvanSavenko:
Ah, should've went to beta branch. Will cherry-pick.

PR #1822, opened 2023-04-01
Bonus refactoring, part3 (save-incompatible)

### LINKED ISSUE #2125: Complex duration broken
**Describe the bug**
After bonus refactor, complex duration broken

**Game logs**
`Error! Wrong bonus duration format.` when using complex duration

**To Reproduce**
1. Start a game

**Expected behavior**
Complex duration should load

**Actual behavior**
Complex duration broken

**Did it work earlier?**
Before #2108

**Version**
 - OS: Linux
 - Version: develop

**Additional context**
Will fix tomorrow

rilian-la-te:
Fixes #2125 

Serialization version is bumped.

Changes:
1. Replaced BonusDuration from enum class to std::bitset (because duration is really flags, and not an enum, so, we need to be able to enable any durations combined)
2. 100% magic resistance and 100% spell damage reduction now should give a spell immunity to creature (to negative and damage spells respectively). Elemental damage 100% reduction should give immunity to any damage effects of this element, including non-magical. For now, works for any effect except maybe fire shield (need check). But fire shield and rebrith needs major rework, which is outside of scope of this PR.
3. Replaced  (SCHOOL)_SPELL_DMG_PREMY with SPELL_DAMAGE bonus (subtype now is school). ANY subtype is used for sorcery skill.
4. Removed bonuses (SCHOOL)_SPELLS - replaced with SPELLS_OF_SCHOOL
5. Removed DIRECT_DAMAGE_IMMUNITY bonus (replaced by 100% spell damage resist in one place when it is used).
6. MAGIC_SCHOOL_SKILL changes subtype to be in line of other spell school bonuses. Now its subtype is school number directly, be school number powered by two.
7. Allow to use spellSchool string identifier as subtype, which was be a motivation to change of 6.
8. Changed unit tests and added new unit tests according to this changes.


client/battle/BattleAnimationClasses.cpp
```diff
@@ -962,13 +962,13 @@ bool EffectAnimation::init()
...
 		{
-			const CStack * destStack = owner.getCurrentPlayerInterface()->cb->battleGetStackByPos(battlehexes[i], false);
+			const auto * destStack = owner.getCurrentPlayerInterface()->cb->battleGetUnitByPos(battlehexes[i], false);
 			Rect tilePos = owner.fieldController->hexPositionLocal(battlehexes[i]);
 
 			be.pos.x = tilePos.x + tilePos.w/2 - first->width()/2;
 
 			if(destStack && destStack->doubleWide()) // Correction for 2-hex creatures.
-				be.pos.x += (destStack->side == BattleSide::ATTACKER ? -1 : 1)*tilePos.w/2;
+				be.pos.x += (destStack->unitSide() == BattleSide::ATTACKER ? -1 : 1)*tilePos.w/2;
```
IvanSavenko, line 971: What's the type of `destStack->unitSide()` and `destStack->side`? Last time I checked it was uint8_t or something. Feel free to change to `BattleSide` (and convert it to enum class if not already)

lib/CTownHandler.h
```diff
@@ -217,6 +217,8 @@ class DLL_LINKAGE CFaction : public Faction
 	std::string getNameTextID() const override;
 
 	bool hasTown() const override;
+	TerrainId getNativeTerrain() const override;
+	EAlignment getAlignment() const override;
```
IvanSavenko, line 221: If class got getter method we can make corresponding members private.

If CTownHandler needs access to these members during loading stage then make CTownHandler friend of CFaction

lib/HeroBonus.cpp
```diff
@@ -2445,7 +2433,7 @@ JsonNode CreatureAlignmentLimiter::toJsonNode() const
 	JsonNode root(JsonNode::JsonType::DATA_STRUCT);
 
 	root["type"].String() = "CREATURE_ALIGNMENT_LIMITER";
-	root["parameters"].Vector().push_back(JsonUtils::stringNode(EAlignment::names[alignment]));
+	root["parameters"].Vector().push_back(JsonUtils::stringNode(GameConstants::ALIGNMENT_NAMES[static_cast<std::underlying_type<EAlignment>::type>(alignment)]));
```
IvanSavenko, line 2436: I get that you want to do it the right way(tm). But readability...

rilian-la-te: It is a draft now, will be split by multiple PRs (native terrain PR will be first, then I will finish teleport).

rilian-la-te: moved to #1876

lib/CTownHandler.h
rilian-la-te: moved to #1882

client/battle/BattleAnimationClasses.cpp
rilian-la-te: @IvanSavenko later, firstly I want to push refactor + native terrain rework + this spell fixes.

Alucard648:
Will it be possible to limit teleporting for open field battles, if all paths to destination hex are blocked by obstacles and creatures?
Also hoping for placeable obstacle placing spells.

rilian-la-te:
@Alucard648 what you mean by placeable obstacle placing spells?

Alucard648:
H5 obscacle creating spells, like Arcane Crystal, Blade Barrier, controllable quicksand/landmine spells etc. And idea for Summon elemental spell - ability to place summoned creature at chosen location.

rilian-la-te:
@Alucard648 this is not obstacles, this is an immobile creatures.

rilian-la-te:
Looks like SPELL_BEFORE_ATTACK/SPELL_AFTER_ATTACK works.

lib/bonuses/BonusEnum.h
```diff
@@ -12,6 +12,8 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
+#include "../JsonNode.h"
```
IvanSavenko, line 13: Forward-declaration should be enough I think?

lib/bonuses/IBonusBearer.cpp
```diff
@@ -44,14 +43,14 @@ bool IBonusBearer::hasBonus(const CSelector &selector, const CSelector &limit, c
 	return getBonuses(selector, limit, cachingStr)->size() > 0;
 }
 
-bool IBonusBearer::hasBonusOfType(BonusType type, int subtype) const
+bool IBonusBearer::hasBonusOfType(BonusType type, std::optional<int> subtype) const
```
IvanSavenko, line 46: Every statement in here has conditional code.

Proposition: split into two methods instead of using default argument:
```c++
hasBonusOfType(BonusType type, int subtype)
hasBonusOfType(BonusType type)
```

lib/bonuses/BonusParams.cpp
```diff
@@ -17,6 +17,34 @@
...
+	"BLOCK_MORALE",
+	"BLOCK_LUCK",
+	"SELF_MORALE",
+	"SELF_LUCK",
+	"DIRECT_DAMAGE_IMMUNITY",
+	"AIR_SPELL_DMG_PREMY",
+	"EARTH_SPELL_DMG_PREMY"
+	"FIRE_SPELL_DMG_PREMY"
+	"WATER_SPELL_DMG_PREMY",
+	"FIRE_SPELLS" \
```
IvanSavenko, line 42: Missing commas?

lib/bonuses/BonusEnum.h
rilian-la-te: No, it is not, because we return a JsonNode.

config/artifacts.json
```diff
@@ -1184,7 +1184,8 @@
 	{
 		"bonuses" : [
 			{
-				"type" : "AIR_SPELL_DMG_PREMY",
+				"type" : "SPELL_DAMAGE",
+				"subtype" : 0,
```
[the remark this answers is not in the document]
rilian-la-te: @IvanSavenko I can introduce another backward-incompatible change - allow school to be claimed from string, if desired. But then MAGIC_SCHOOL_SKILL should be redesigned.

lib/serializer/BinarySerializer.h
```diff
@@ -364,6 +364,12 @@ class DLL_LINKAGE BinarySerializer : public CSaverBase
 		for(ui32 i = 0; i < length; i++)
 			save(data.data()[i]);
 	}
+	template <std::size_t T>
+	void save(const std::bitset<T> &data)
+	{
+		auto writ = static_cast<uint32_t>(data.to_ulong());
```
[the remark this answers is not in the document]
rilian-la-te: We can save it as uint64_t, if desired. But I can just add a static_assert, because I cannot imagine why we can need such large bitset. What is better?

lib/bonuses/IBonusBearer.cpp
```diff
@@ -15,22 +15,21 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
-int IBonusBearer::valOfBonuses(BonusType type, int subtype) const
+int IBonusBearer::valOfBonuses(BonusType type, std::optional<int> subtype) const
```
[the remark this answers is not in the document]
rilian-la-te: I will do, but why you think than 2 methods is better? Method have only 5 LoC, so, pretty readable.

lib/bonuses/BonusParams.cpp
rilian-la-te: Will fix

lib/bonuses/IBonusBearer.cpp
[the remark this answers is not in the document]
IvanSavenko: LoC is very poor code quality metric.

Two simple methods are much better than one complex method, no matter how small it is

lib/serializer/BinarySerializer.h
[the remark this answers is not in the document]
IvanSavenko: Or both?
1) add static_assert to make sure that bad code fails to compile instead of producing weird and hard to find bug.
2) serialize based on bitset size as uint16/uint32/uint64 - we have a lot of bonus instances so saving them with as smallest possible size might have notable effect on savegame size.

config/artifacts.json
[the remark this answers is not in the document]
IvanSavenko: Personally I am fine with keeping this as it *for now*.
But considering recent changes and now - also this change, we really need some way to replace magic numbers with proper strings for 1.3.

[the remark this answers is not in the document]
rilian-la-te: I will add strings for spellSchool and will unify subtypes of spell school bonuses. Agree with you than we need, but for now IDK how.

lib/bonuses/IBonusBearer.cpp
rilian-la-te: done

config/artifacts.json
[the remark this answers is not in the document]
IvanSavenko: One option would be to use fake identifiers, e.g. register identifier "bonusSubtype.seaMovement" with value 0 (or was that 1?) and resolve it as we normally do.
And then add similar identifiers to cover all values that we need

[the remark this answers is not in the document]
rilian-la-te: Sea is zero, land is one. BTW, we does have recurring subtypes like spell school or damage type. It makes sense to register it just as const.

lib/serializer/BinarySerializer.h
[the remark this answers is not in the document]
rilian-la-te: Done with constexpr if

config/artifacts.json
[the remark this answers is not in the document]
IvanSavenko: Spell schools - if we can reuse existing school constants for bonus subtypes then we should do so.
But othervice - yeah, register them separately, using for example "bonusSubtype" type

lib/GameConstants.h
```diff
@@ -1296,14 +1297,17 @@ class Obstacle : public BaseForID<Obstacle, si32>
 	DLL_LINKAGE static Obstacle fromString(const std::string & identifier);
 };
 
-enum class ESpellSchool: ui8
+enum class ESpellSchool: int16_t
```
IvanSavenko, line 1300: Likely not a big deal, but was ui8 before. Might make sense to change to int8_t, not to int16_t

config/battlefields.json
```diff
@@ -13,7 +13,7 @@
 		"bonuses": [
 			{
 				"type" : "MAGIC_SCHOOL_SKILL",
-				"subtype" : 0,
+				"subtype" : "spellSchool.any",
```
IvanSavenko, line 16: Change is OK, but might cause hidden changes in mods, since subtype now use different underlying values. Or will your bonus conversion handle this?

lib/bonuses/BonusEnum.h
IvanSavenko: No, it is.
You can use forward-declared types as return type or as argument of a function declaration.

For function definition in cpp you will need full type definition though.

IvanSavenko (approved):

config/battlefields.json
rilian-la-te: It is not, and I do not know how it should be handled by converter. Yes, this will require change inside mods (surely in WoG, unsure about others).

lib/GameConstants.h
rilian-la-te: Yes, it was ui8, but is int8_t enough to us?

config/battlefields.json
IvanSavenko: Just thinking on how to detect such cases in mods (to help modders with update)
- require subtypes for *all* bonuses to be string, e.g. on validation level (will need more changes, e.g. damage types, movement types, probably others)
- require subtypes for *some* bonuses to be string, including this one. Not sure how to define this though. Perhaps as property of a bonus, inside CBonusType class? Or simply hardcoded list of bonuses that require strings (or opposite - that require integers)
- rename bonus and handle as bonus conversion

lib/GameConstants.h
IvanSavenko: - if goal was to allow -1 as valid value then we should just switch to int8_t
- if goal is to ensure that any potential new elements will fit - then IMO all such identifiers should use int32_t, as majority of our existing identifiers do

That being said, highly doubt that we'll ever have 100+ schools

lib/bonuses/BonusEnum.h
rilian-la-te: Ok, done

config/battlefields.json
rilian-la-te: I do not think than this should be handled in this PR

lib/GameConstants.h
rilian-la-te: done as int8_t

config/battlefields.json
IvanSavenko: okay, as long as it is handled in some way by 1.3

rilian-la-te: I do not know how to handle it properly for now. Maybe we need some lists of bonuses which require a subtype with chosen identifier?

IvanSavenko: Yeah, something like that.
We can add:
- mapping bonus type -> expected identifiers "class" for subtype value, e.g. `{ MAGIC_SCHOOL_SKILL, "spellSchool" }`
- when attempting to resolve identifier during load (e.g. resolveIdentifer/getIdentifier calls) we should provide this mapped bonus type. This will force identifier class check during bonus resolution
- would also be great to detect whether bonus uses subtypes to begin with - to provide modder with warning if he attempts to give subtype to a bonus that does not uses them

Not sure whether this is best approach or whether this will actually map well on our system, so not insisting on this particular form.

lib/bonuses/BonusEnum.h
```diff
@@ -221,18 +215,20 @@ enum class BonusType
...
+	constexpr Type PERMANENT = 1 << 0;
+	constexpr Type ONE_BATTLE = 1 << 1; //at the end of battle
+	constexpr Type ONE_DAY = 1 << 2;   //at the end of day
+	constexpr Type ONE_WEEK = 1 << 3; //at the end of week (bonus lasts till the end of week, thats NOT 7 days
+	constexpr Type N_TURNS = 1 << 4; //used during battles, after battle bonus is always removed
+	constexpr Type N_DAYS = 1 << 5;
+	constexpr Type UNTIL_BEING_ATTACKED = 1 << 6; /*removed after attack and counterattacks are performed*/
+	constexpr Type UNTIL_ATTACK = 1 << 7; /*removed after attack and counterattacks are performed*/
+	constexpr Type STACK_GETS_TURN = 1 << 8; /*removed when stack gets its turn - used for defensive stance*/
+	constexpr Type COMMANDER_KILLED = 1 << 9;
```
IvanSavenko, line 231: Not 100% sure about this change, but not against it so OK I guess?

IvanSavenko (approved):

lib/bonuses/BonusEnum.h
rilian-la-te: If we will not use  bit sets instead of enum class, then we will break complex durations (like Blind spell).

rilian-la-te: So, I think this is better than enum class.

PR #1848, opened 2023-04-02
Fix map editor crash with cutting random town

Nordsoft91:


dydzio0614:
I would probably target beta branch

Issue #1537, opened 2023-02-01
Player settings empty by default in map editor (0 players and claims it's 2 until setting changed)


dydzio0614:
**Describe the bug**
There are 0 player colors by default in map editor configured for new maps. When you open player settings it shows "2 players" but this is false info - to properly get 2 players and be able to place heroes on map you need to select something else and then select 2 players again. Also by default player colors are "CPU only" which is unintuitive.

**To Reproduce**
1. Create new map
2. You cannot place heroes as you cannot select any color by default
3. Open player settings (or whatever is the name of settings where you configure player colors)
4. It says "2 players" even though its factually wrong - you need to select something to trigger actual creation of player colors for map

**Expected behavior**
Sane default amount of colors, that are accessible without extra steps on new map creation (for example as in original H3 editor)

**Actual behavior**
Map is initialized with 0 colors, and player settings spread misinformation about 2 available colors until setting is changed.

**Did it work earlier?**
No

**Version**
 - OS: all
 - Version: all

PR #1849, opened 2023-04-02
Allow single player on map and fix player params initialization

Nordsoft91:
Fixes https://github.com/vcmi/vcmi/issues/1537


krs0:
Will this work now with no warning for victory condition = artifact?
`Map %s has only one player but allows normal victory?`

Nordsoft91:
Should work

PR #1850, opened 2023-04-02
[Conan][docs] improve using prebuilt binaries for Apple Silicon

kambala-decapitator:
Actually tested build instructions on ARM Mac and found the missing bit


kambala-decapitator:
build error has no relation to the PR

PR #1868, opened 2023-04-03
Shut down the thread for tactic phase properly

Adriankhl:
## Issue

1. Use a hero with the tactic skill to enter a battle
2. Load game during the tactic phase
3. Crash

## Fix
The thread for tactic phase keeps running even if the game is ended. This PR shut that down properly.

This might not be the most elegant solution since I need to create a `forceEndTacticPhase()` function. However, I couldn't figure out a good solution given how the thread is implemented. 

Btw, I have a feeling that it is very easy to make mistake with threads in VCMI, I suggest that people who review PR (including mine :smile:) with threading-related code should pay extra attention to ensure that they work properly.

### DIFF
--- a/client/Client.cpp
+++ b/client/Client.cpp
@@ -7,6 +7,7 @@
  * Full text of license available in license.txt file, in main folder
  *
  */
+#include "Global.h"
 #include "StdInc.h"
 #include "Client.h"
 
@@ -31,6 +32,7 @@
 #include "../lib/registerTypes/RegisterTypes.h"
 #include "../lib/serializer/Connection.h"
 
+#include <memory>
 #include <vcmi/events/EventBus.h>
 
 #if SCRIPTING_ENABLED
@@ -369,6 +371,9 @@ void CClient::endGame()
 		logNetwork->info("Deleted mapHandler and gameState.");
 	}
 
+	//threads cleanup has to be after gs cleanup and before battleints cleanup to stop tacticThread
+	cleanThreads();
+
 	playerint.clear();
 	battleints.clear();
 	battleCallbacks.clear();
@@ -593,7 +598,8 @@ void CClient::battleStarted(const BattleInfo * info)
 
 	if(info->tacticDistance && vstd::contains(battleints, info->sides[info->tacticsSide].color))
 	{
-		boost::thread(&CClient::commenceTacticPhaseForInt, this, battleints[info->sides[info->tacticsSide].color]);
+		PlayerColor color = info->sides[info->tacticsSide].color;
+		playerTacticThreads[color] = std::make_unique<boost::thread>(&CClient::commenceTacticPhaseForInt, this, battleints[color]);
 	}
 }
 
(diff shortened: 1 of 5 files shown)


krs0:
@Adriankhl VCMI developers use Slack for communication. If you want to join...
https://slack.vcmi.eu/

client/Client.cpp
```diff
@@ -593,7 +598,7 @@ void CClient::battleStarted(const BattleInfo * info)
 
 	if(info->tacticDistance && vstd::contains(battleints, info->sides[info->tacticsSide].color))
 	{
-		boost::thread(&CClient::commenceTacticPhaseForInt, this, battleints[info->sides[info->tacticsSide].color]);
+		playerTacticThreads[info->sides[info->tacticsSide].color] = std::make_unique<boost::thread>(boost::thread(&CClient::commenceTacticPhaseForInt, this, battleints[info->sides[info->tacticsSide].color]));
```
nullkiller, line 601: Do we need to specify it once again here if we already said it in std::make_unique<boost::thread>? I am not c++ pro, just want to ensure there will be no copy-construction here

nullkiller: std::make_unique<boost::thread>(boost::thread(&CClient::commenceTacticPhaseForInt, this, battleints[info->sides[info->tacticsSide].color]));
```suggestion
		playerTacticThreads[info->sides[info->tacticsSide].color] = std::make_unique<boost::thread>(&CClient::commenceTacticPhaseForInt, this, battleints[info->sides[info->tacticsSide].color]);

```

IvanSavenko: I believe boost::thread is already non-copyable, so copy-construction should be impossible.
However this line is very overloaded. Perhaps split it?
```suggestion
		PlayerColor color = info->sides[info->tacticsSide].color;
		playerTacticThreads[color] = std::make_unique<boost::thread>(&CClient::commenceTacticPhaseForInt, this, battleints[color]);

```

Nordsoft91, line 601: does it make sense to check if playerTacticThreads[info->sides[info->tacticsSide].color] thread is finished before creating a new thread?

Adriankhl: Yes this looks better

Adriankhl: How can I check this? Call join() if the thread exists?

client/Client.cpp
```diff
@@ -593,7 +598,8 @@ void CClient::battleStarted(const BattleInfo * info)
 
 	if(info->tacticDistance && vstd::contains(battleints, info->sides[info->tacticsSide].color))
 	{
-		boost::thread(&CClient::commenceTacticPhaseForInt, this, battleints[info->sides[info->tacticsSide].color]);
+		PlayerColor color = info->sides[info->tacticsSide].color;
```
Nordsoft91, line 601: can be moved out: 
`if(info->tacticDistance && vstd::contains(battleints, info->sides[info->tacticsSide].color))`
replace with
```
PlayerColor color = info->sides[info->tacticsSide].color;
if(info->tacticDistance && vstd::contains(battleints, color))
```

Nordsoft91 (approved):

client/Client.cpp
Adriankhl: There are many `color` inside that function (`battleStarted`), moving the `color` out might confuse other people in the future, I think there is a reason to keep this inside the if bracket.

PR #1874, opened 2023-04-04
Separate movement highlight for hovered units in battle

krs0:
So far highlighting valid movement for hovered unit works. (in tactics phase too).
Highlighted hex needs a better drawing. (not that pixelated and a little less blue)
There are no individual options.
There are no Shortcut keys implemented.
Max possible movement (as if no other units) not implemented yet.


IvanSavenko (commented):
Code-wise looks OK so far.
However:
- please use more readable file names for new images. "CCNSSh2" is not exactly readable.
- perhaps replace them with better files that don't look like made in paint and use proper alpha channel? (see below)
hex-border
hex-fill

krs0:
> * please use more readable file names for new images. "CCNSSh2" is not exactly readable.
I used same names as in HD Mod... will change.

> * perhaps replace them with better files that don't look like made in paint and use proper alpha channel? (see below)
But they were made in paint! 😅 
... thank you for the images.

krs0:
When a creature is hovered., besides movement highlighting, now for max possible movement, a hex in center of map is highlighted.

krs0:
Ok first phase of highlighting movement for hovered stack is done.
New graphics is added,
Behaviour seems ok in tactics mode and in proper combat
New combat option added: Movement Highlight on Hover

config/widgets/settings/battleOptionsTab.json
```diff
@@ -28,7 +28,7 @@
 				},
 				{
 					"text": "core.genrltxt.397", // Creature info
-					"position": {"x": 10, "y": 205}
+					"position": {"x": 10, "y": 240}
```
IvanSavenko, line 31: Other values changes by 30px, but this one by 35px. Intended?

IvanSavenko (approved):

config/widgets/settings/battleOptionsTab.json
krs0: Probably not. Nice catch!
Corrected.

IvanSavenko:
@krs0 
By the way, for ranged damage highlight we can simply show alternative hex highlight image:
- (nothing) for hexes with full damage range
- (broken?) arrow & '1/2' text for hexes with ranged penalty OR obstacle penalty
- (broken?) arrow & '1/4' text for hexes with ranged penalty AND obstacle penalty
- (broken?) arrow & crossed out circle for creatures with range limit

HOWEVER this should be hidden behind some hotkey / disabled by default option.
Should be way more easier than nice border highlight

dydzio0614:
some sort of border for 1/2 would be neat though...

but maybe I would use hex graphics for that marked by some color, rather than using hex borders,  like green star in the middle of hex in range

IvanSavenko:
@dydzio0614 Not saying that you're wrong, but this is (relatively) hard to implement feature since you need detect bordering hexes, then detect sections of hex that should be highlightted, select appropriate graphical file for each section and make sure that this looks presentable even if shooter has movement range of ~10 hexes.

And this can get quite messy if you also want to also show obstacle penalty from walls.

So I am not against it (provided that it looks presentable), but IMO too complex and too many potential issues compared to much simple and faster to implement option of displaying custom hex overlay image for certain hexes.

IvanSavenko:
> @dydzio0614 Not saying that you're wrong, but this is (relatively) hard to implement feature since you need detect bordering hexes, then detect sections of hex that should be highlightted, select appropriate graphical file for each section and make sure that this looks presentable even if shooter has movement range of ~10 hexes.
> 
> And this can get quite messy if you also want to also show obstacle penalty from walls.
> 
> So I am not against it (provided that it looks presentable), but IMO too complex and too many potential issues compared to much simple and faster to implement option of displaying custom hex overlay image for certain hexes.

krs0:
After thinking about how to calculate the border line... I came to similar onclusion.... Highlight whole hex.
But not whole range, only the hexes on the limit,  and show existing cursors for full damage.

krs0:
Should be merged ok now.

Issue #1745, opened 2023-03-22
Something wrong with the settings window on Linux


SoundSSGood:
**Describe the bug**
The settings window looks a bit different than on Windows

**Game logs**
[VCMI_Client_log.txt](https://github.com/vcmi/vcmi/files/11040300/VCMI_Client_log.txt)
[VCMI_Server_log.txt](https://github.com/vcmi/vcmi/files/11040301/VCMI_Server_log.txt)

**To Reproduce**
Open settings window

**Did it work earlier?**
Don't know

**Screenshots**
![settings_linux](https://user-images.githubusercontent.com/87084363/226916717-be6b4c78-f210-45a8-97d5-e09182666cf3.png)

**Version**
 - OS: Kubuntu 22.04
 - Version: [VCMI version] last develop

IvanSavenko:
`"language" : "russian",`
VCMI attempts to use localized texts but fonts from English copy of H3 don't support cyrillic character set.
Workaround 1: install localization mod
Workaround 2: switch language in Launcher to English

Proper solution (for 1.2.1 release perhaps?) would be to reintroduce true type fonts that support most/all needed languages & use them as fallback for when H3 font can't render text.

SoundSSGood:
Checked the second workround. This fixed the issue.
I don’t know if it’s necessary to close the ticket before the full fix in 1.2.1

IvanSavenko:
Added True Type Fonts submod for vcmi-extra, to be used in cases like this.
Nicer / better solution that looks more in line with H3 fonts might be possible, but probably not in near future

PR #1875, opened 2023-04-04
TTF fonts fallback

IvanSavenko:
Implements support for fixing #1745 

Actual fix (true type fonts) will probably go to vcmi-extras for now - while license-wise fonts can go to vcmi mod, but not sure whether we want this behavior by default.
Plus my concern is that if we remove them later some players will still keep them - since our installer right now is not removing old files properly at the moment


rilian-la-te:
@IvanSavenko but we already use version 821)

IvanSavenko:
> @IvanSavenko but we already use version 821)

Yet another example at to why our serialization versioning is a mess...

rilian-la-te:
@IvanSavenko agree, but do you have other ideas how to handle different serialization?

IvanSavenko:
> @IvanSavenko agree, but do you have other ideas how to handle different serialization?

1) Instead of incrementing version use some unique ID. For example, ID of pull request this branch belongs to or issue ID that this branch fixes. This is unique number that will not conflict between different branches.
2) Instead of having min version and current version ID's use single version ID which will be changed whenever there is a breaking change (using unique ID from p.1)
3) For non-breaking changes, introduce "feature list" - list of ID's that will be located in save header after version ID. Serializer should have method hasFeature(int ID) that returns true if save contains selected "feature". Feature ID's should also use same format from p.1.
4) Whenever there is a breaking change, all existing features should be removed and used unconditionally - to prevent pileup of unused features

rilian-la-te:
@IvanSavenko this is also human errors possible. But firstly we need to select what we actually should serialize, and what should not.

IvanSavenko:
> @IvanSavenko this is also human errors possible. But firstly we need to select what we actually should serialize, and what should not.

Maybe, but this is very long-term issue and for now we should serialize everything that can be serialized, with exception of hardcoded values that don't change with mods.

Meanwhile better serialization that is more fool-proof can be done at any point, including in 1.3

> there is also topic about whether we should serialize AI state... AI serialization was dropped in VCAI due to some problems with serialization/deserialization

Local player state (e.g. sleeping heroes, heroes path) also suffer from this problem. IMO player state can be stored on server, in server savegame, and as for AI - it can either use similar approach, e.g. send it to server or perhaps even move AI to server (not something that likely to happen soon though)

> there also comes topic of serializing battle state if we want to support combat save/load as in H4

Should be done only if we actually implement this feature. Not much point in writing code that won't be used for years.

rilian-la-te:
@IvanSavenko server should not know anything about player state, and if player is AI or not. But can you remember me why client serialization is broken?

rilian-la-te:
@IvanSavenko I agree with this approach only partially. I think server save game should be allowed to load on any client, and language/graphic mods/sounds should be irrelevant. But if we will store client state on the server - language and mods will be predefined by server, and not by local game settings.

About AI - server should know single boolean, but not AI state, because I think than we should be able to change between AI implementations even mid-party.

PR #1876, opened 2023-04-04
VCMI Refactoring

rilian-la-te:
1. Change EAlignment to enum class
2. Change Res::ERes to enum class
3. Move CCreature API which has getters to private space.

Unfortunately, this changes cause many LOC change, so, I decide to do a refactoring PR.

@IvanSavenko

### DIFF
--- a/AI/Nullkiller/AIUtility.cpp
+++ b/AI/Nullkiller/AIUtility.cpp
@@ -20,6 +20,8 @@
 
 #include "../../lib/GameSettings.h"
 
+#include <vcmi/CreatureService.h>
+
 namespace NKAI
 {
 
@@ -278,8 +280,8 @@ creInfo infoFromDC(const dwellingContent & dc)
 	ci.creID = dc.second.size() ? dc.second.back() : CreatureID(-1); //should never be accessed
 	if (ci.creID != -1)
 	{
-		ci.cre = VLC->creh->objects[ci.creID].get();
-		ci.level = ci.cre->level; //this is cretaure tier, while tryRealize expects dwelling level. Ignore.
+		ci.cre = VLC->creatures()->getById(ci.creID);
+		ci.level = ci.cre->getLevel(); //this is creature tier, while tryRealize expects dwelling level. Ignore.
 	}
 	else
 	{
@@ -398,7 +400,7 @@ bool shouldVisit(const Nullkiller * ai, const CGHeroInstance * h, const CGObject
 			{
 				if(level.first
 					&& h->getSlotFor(CreatureID(c)) != SlotID()
-					&& ai->cb->getResourceAmount().canAfford(c.toCreature()->cost))
+					&& ai->cb->getResourceAmount().canAfford(c.toCreature()->getFullRecruitCost()))
 				{
 					return true;
 				}
@@ -411,7 +413,7 @@ bool shouldVisit(const Nullkiller * ai, const CGHeroInstance * h, const CGObject
 	{
 		for(auto slot : h->Slots())
 		{
-			if(slot.second->type->upgrades.size())
+			if(slot.second->type->hasUpgrades())
 				return true; //TODO: check price?
 		}
 		return false;
@@ -438,7 +440,7 @@ bool shouldVisit(const Nullkiller * ai, const CGHeroInstance * h, const CGObject
 			return false;
 
 		TResources myRes = ai->getFreeResources();
-		if(myRes[Res::GOLD] < 2000 || myRes[Res::GEMS] < 10)
+		if(myRes[EGameResID::GOLD] < 2000 || myRes[EGameResID::GEMS] < 10)
 			return false;
 		break;
 	}
--- a/AI/Nullkiller/Engine/FuzzyEngines.cpp
+++ b/AI/Nullkiller/Engine/FuzzyEngines.cpp
@@ -65,13 +65,13 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 	for(auto s : army->Slots())
 	{
 		bool walker = true;
-		const CCreature * creature = s.second->type;
-		if(creature->hasBonus(selectorSHOOTER, keySHOOTER))
+		auto bearer = s.second->getType()->getBonusBearer();
+		if(bearer->hasBonus(selectorSHOOTER, keySHOOTER))
 		{
 			shootersStrength += s.second->getPower();
 			walker = false;
 		}
-		if(creature->hasBonus(selectorFLYING, keyFLYING))
+		if(bearer->hasBonus(selectorFLYING, keyFLYING))
 		{
 			flyersStrength += s.second->getPower();
 			walker = false;
@@ -79,7 +79,7 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 		if(walker)
 			walkersStrength += s.second->getPower();
 
-		vstd::amax(maxSpeed, creature->valOfBonuses(selectorSTACKS_SPEED, keySTACKS_SPEED));
+		vstd::amax(maxSpeed, bearer->valOfBonuses(selectorSTACKS_SPEED, keySTACKS_SPEED));
 	}
 	armyStructure as;
 	as.walkers = static_cast<float>(walkersStrength / totalStrength);
--- a/Global.h
+++ b/Global.h
@@ -753,6 +753,12 @@ namespace vstd
 		if(i < 0) return -i;
 		return i;
 	}
+
+	///C++23
+	template< class Enum > constexpr std::underlying_type_t<Enum> to_underlying( Enum e ) noexcept
+	{
+		return static_cast<std::underlying_type_t<Enum>>(e);
+	}
 }
 using vstd::operator-=;
 
--- a/Global.h
+++ b/Global.h
@@ -279,7 +279,7 @@ template<typename T, size_t N> char (&_ArrayCountObj(const T (&)[N]))[N];
 #define VCMI_LIB_NAMESPACE_BEGIN
 #define VCMI_LIB_NAMESPACE_END
 #define VCMI_LIB_USING_NAMESPACE
-#define VCMI_LIB_WRAP_NAMESPACE(x) x
+#define VCMI_LIB_WRAP_NAMESPACE(x) ::x
 #endif
 
 /* ---------------------------------------------------------------------------- */
--- a/client/lobby/OptionsTab.cpp
+++ b/client/lobby/OptionsTab.cpp
@@ -126,25 +126,25 @@ size_t OptionsTab::CPlayerSettingsHelper::getImageIndex()
 			return GOLD;
 		case PlayerSettings::RESOURCE:
 		{
-			switch((*CGI->townh)[factionIndex]->town->primaryRes)
+			switch((*CGI->townh)[factionIndex]->town->primaryRes.toEnum())
 			{
-			case Res::WOOD_AND_ORE:
+			case EGameResID::WOOD_AND_ORE:
 				return WOOD_ORE;
-			case Res::WOOD:
+			case EGameResID::WOOD:
 				return WOOD;
-			case Res::MERCURY:
+			case EGameResID::MERCURY:
 				return MERCURY;
-			case Res::ORE:
+			case EGameResID::ORE:
 				return ORE;
-			case Res::SULFUR:
+			case EGameResID::SULFUR:
 				return SULFUR;
-			case Res::CRYSTAL:
+			case EGameResID::CRYSTAL:
 				return CRYSTAL;
-			case Res::GEMS:
+			case EGameResID::GEMS:
 				return GEM;
-			case Res::GOLD:
+			case EGameResID::GOLD:
 				return GOLD;
-			case Res::MITHRIL:
+			case EGameResID::MITHRIL:
 				return MITHRIL;
 			}
 		}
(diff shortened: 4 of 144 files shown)


lib/GameConstants.h
```diff
@@ -428,10 +428,7 @@ class SecondarySkill
 
 ID_LIKE_OPERATORS(SecondarySkill, SecondarySkill::ESecondarySkill)
 
-namespace EAlignment
-{
-	enum EAlignment { GOOD, EVIL, NEUTRAL };
-}
+enum class EAlignment : uint8_t { GOOD, EVIL, NEUTRAL };
```
IvanSavenko, line 431: EAlignment *might* also benefit from Identifier class, but not sure whether it is necessary at this point.
It also not as widespread through the code as ERes, so changing it in future will be easy enough

include/vcmi/Creature.h
```diff
@@ -50,7 +52,10 @@ class DLL_LINKAGE Creature : public EntityWithBonuses<CreatureID>
 	virtual int32_t getBaseSpeed() const = 0;
 	virtual int32_t getBaseShots() const = 0;
 
-	virtual int32_t getCost(int32_t resIndex) const = 0;
+	virtual int32_t getCost(ERes resIndex) const = 0;
+	virtual ResourceSet fullCost() const = 0;
```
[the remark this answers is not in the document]
rilian-la-te: I actually tried overload, but renamed to `fullCost` for LUA scripting (lua proxy has diffliculties to use overloads). All other suggestions will be done.

lib/CCreatureHandler.cpp
```diff
@@ -606,15 +617,15 @@ CCreature * CCreatureHandler::loadFromJson(const std::string & scope, const Json
 	{
 		doubledCreatures.insert(CreatureID(index));
 	}
-	cre->idNumber = CreatureID(index);
+	cre->getId() = CreatureID(index);
```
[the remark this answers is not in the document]
rilian-la-te: Handler is already friend class, so, will revert (auto-replace is so auto-replacy)

lib/ResourceSet.h
```diff
@@ -19,169 +19,166 @@ using TResourceCap = int64_t; //to avoid overflow when adding integers. Signed v
...
+enum class ERes
+{
+	WOOD = 0, MERCURY, ORE, SULFUR, CRYSTAL, GEMS, GOLD, MITHRIL,
 
-		WOOD_AND_ORE = 127,  // special case for town bonus resource
-		INVALID = -1
-	};
+	WOOD_AND_ORE = 127,  // special case for town bonus resource
+	INVALID = -1
+};
```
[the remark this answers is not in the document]
IvanSavenko: Also - please give ERes (or whatever its name will be) explicit base type, e.g.
`enum class ERes : int32_t` (or int8_t, but that may have some side effects in theory)

lib/GameConstants.h
rilian-la-te: I do not think than this huge PR should be more huge. I think, later is better.

lib/ResourceSet.h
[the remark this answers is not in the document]
rilian-la-te: Done

lib/CCreatureHandler.cpp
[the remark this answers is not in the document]
rilian-la-te: Reverted

include/vcmi/Creature.h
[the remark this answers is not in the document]
rilian-la-te: Done without overload (getRecruitCost + getFullRecruitCost)

lib/mapping/MapFormatH3M.cpp
```diff
@@ -1256,7 +1256,7 @@ void CMapLoaderH3M::readObjects()
 				readMessageAndGuards(res->message, res, objPos);
 
 				res->amount = reader->readUInt32();
-				if(objTempl->subid == Res::GOLD)
+				if(objTempl->subid == vstd::to_underlying(ECurrencyID::GOLD))
```
IvanSavenko, line 1259: I would probably replace all to_underlying with getNum(), if only for consistency with existing code.
```suggestion
				if(objTempl->subid == CurrencyID(ECurrencyID::GOLD).getNum())
```

Other than this - good to go I think?

AI/Nullkiller/AIUtility.cpp
```diff
@@ -278,8 +280,8 @@ creInfo infoFromDC(const dwellingContent & dc)
 	ci.creID = dc.second.size() ? dc.second.back() : CreatureID(-1); //should never be accessed
 	if (ci.creID != -1)
 	{
-		ci.cre = VLC->creh->objects[ci.creID].get();
-		ci.level = ci.cre->level; //this is cretaure tier, while tryRealize expects dwelling level. Ignore.
+		ci.cre = static_cast<const CCreature*>(VLC->creatures()->getById(ci.creID));
```
IvanSavenko, line 283: What prevents changing type of 'ci.cre' to whatever getById returns?

lib/mapping/MapFormatH3M.cpp
rilian-la-te: @IvanSavenko I will, but why? I think using standard library function is good too (and std::to_underlying is part of C++23).

client/lobby/OptionsTab.cpp
```diff
@@ -126,25 +126,25 @@ size_t OptionsTab::CPlayerSettingsHelper::getImageIndex()
 			return GOLD;
 		case PlayerSettings::RESOURCE:
 		{
-			switch((*CGI->townh)[factionIndex]->town->primaryRes)
+			switch(static_cast<ECurrencyID>((*CGI->townh)[factionIndex]->town->primaryRes))
```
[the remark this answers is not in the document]
rilian-la-te: No, it is just int. Will do

AI/Nullkiller/AIUtility.cpp
rilian-la-te: Many functions in both AIs requires CCreature to go, I think passing Creature there will lead to another wall of changes. But will try.

IvanSavenko: Up to you then

lib/mapping/MapFormatH3M.cpp
IvanSavenko: From my point it is not about library vs custom (Identifier will use to_underlying internally anyway).
The way I see it, this code has constant that should not have been hardcoded.
Proper code that we (eventually) should have here is something like
```c++
    CurrencyID resourceType(objTempl->subID);
    if (resourceType->pileMultiplier != 1)
        amount = amount * resourceType->multiplier;
```
Meaning, this code should use CurrencyID, but its value should come from elsewhere instead of being hardcoded to be gold

IvanSavenko: Or, in other words, I see enumeration values embedded in code a necessary evil at this point (but still huge improvement over hardcoded numeric constants we had originally)
Meaning, we shouldn't add code that is specific to enum's (such as to_underlying)

rilian-la-te: Agree with you on anything but resources, because resources will not be able to load from mods AFAIK.

IvanSavenko: Maybe. But I'd rather keep possibility for new resources open. There is no need to have mithril hardcoded & we eventually may want new resources, e.g. for something like H5-like necromancy.

And at the very least, I'd like to have handler class for resources, if only for better organization of resource-specific data (initial resources, market prices, icons, etc) instead of having it scattered across entire codebase.

client/lobby/OptionsTab.cpp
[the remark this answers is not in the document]
rilian-la-te: done

AI/Nullkiller/AIUtility.cpp
rilian-la-te: done

lib/mapping/MapFormatH3M.cpp
rilian-la-te: All to_underlying for ECurrencyID replaced.

krs0:
Why did you changed naming to ECurrencyID? Why not EResourceID?

rilian-la-te:
> Why not EResourceID?

It is already used by filesystem resources, so, @IvanSavenko suggested ECurrencyID to be distinguish.

rilian-la-te:
@DjWarmonger do you want to review this?

AI/Nullkiller/Engine/FuzzyEngines.cpp
```diff
@@ -65,7 +65,7 @@ armyStructure evaluateArmyStructure(const CArmedInstance * army)
 	for(auto s : army->Slots())
 	{
 		bool walker = true;
-		const CCreature * creature = s.second->type;
+		auto creature = s.second->getType()->accessBonuses();
```
DjWarmonger, line 68: Why does `accessBonuses()` return a creature? This is confusing.

Global.h
```diff
@@ -753,6 +753,12 @@ namespace vstd
 		if(i < 0) return -i;
 		return i;
 	}
+
+	///C++23
```
DjWarmonger, line 757: Since when do we support or require C++23?

include/vcmi/Entity.h
```diff
@@ -14,6 +14,12 @@ VCMI_LIB_NAMESPACE_BEGIN
 
 class IBonusBearer;
 
+class DLL_LINKAGE WithBonuses
+{
+public:
+	virtual const IBonusBearer * accessBonuses() const = 0;
```
DjWarmonger, line 20: maybe `toBonusBearer()` would be a better name?

lib/ResourceSet.h
```diff
@@ -19,169 +20,168 @@ using TResourceCap = int64_t; //to avoid overflow when adding integers. Signed v
 class JsonNode;
 class JsonSerializeFormat;
 
-namespace Res
+class ResourceSet;
+
+enum class ECurrencyID : int8_t
```
DjWarmonger, line 25: Please rename it to `EResourceID`, which is the name normally used for resources. We don't mine any currency here.

DjWarmonger (changes requested):
Some names are confusing.

Global.h
rilian-la-te: We does not support C++23, it is just a feature backported from c++23, in global there is many such features, like clamp from c++17.

include/vcmi/Entity.h
rilian-la-te: @DjWarmonger then we should redesign all entities to it. Should I do it now?

lib/ResourceSet.h
rilian-la-te: @IvanSavenko @DjWarmonger  maybe EGameResID? It is less confusing.

AI/Nullkiller/Engine/FuzzyEngines.cpp
rilian-la-te: renamed to bearer

include/vcmi/Entity.h
DjWarmonger: What is actually the purpose of this class? Just the wrapper of `IBonusBearer`?

lib/ResourceSet.h
IvanSavenko: Either will do for me

include/vcmi/Entity.h
rilian-la-te: Yes, and this wrapper is extensively used in Lua, and I think it is good to use it anywhere where we do not need write access to bonus tree.

DjWarmonger: Okay, I still think we shoudl use as few names and conventions as possible. Maybe just `getBonusBearer()`.

AI/Nullkiller/Engine/FuzzyEngines.cpp
rilian-la-te: Will push after all other topics will decided)

include/vcmi/Entity.h
rilian-la-te: There was an EntityWithBonuses class, and I splitted "WithBonuses" part from it.

rilian-la-te: If there is no other scripts than use accessBonuses exists, then I can rename it to getBonusBearer. Anyway, this class is 6 years old)

rilian-la-te: @IvanSavenko @DjWarmonger there is other similar method in BattleInfo, called asBearer. Maybe we should choose a common name for it, if we need to change?

rilian-la-te: And getBattleNode exists too for a same purpose. I think it is a good refactoring to unify it under a same interface.

IvanSavenko: `toBonusBearer` might be slightly better since this is effectively type casting. But not against `getBonusBearer`. Both are better than current `accessBonuses`.

Considering that we have basically zero scripts rename of a method related to scripts it not a problem

IvanSavenko: Or perhaps getBonusBearer after all? If only for consistency.

toBonusBearer might better represent method since this is type conversion and not getter. On the other hand we can view this fact internal info and provide `getBonusBearer` everywhere

rilian-la-te: Maybe use `toBearer`? Same semantic, but less words.

rilian-la-te: @IvanSavenko I agree with you and think than outside classes should not know about is class a bonus bearer or not.

lib/ResourceSet.h
rilian-la-te: Renamed to GameResID/EGameResID.

rilian-la-te: Will push after all topics will be agreed.

include/vcmi/Entity.h
rilian-la-te: @IvanSavenko @DjWarmonger what name is better? `getBonusBearer`, `getBearer`, `toBonusBearer`, `toBearer`?

IvanSavenko: My preference:
1) getBonusBearer
2) toBonusBearer

rilian-la-te: So, longer is better. Will do getBonusBearer then.

rilian-la-te: All changes pushed.

rilian-la-te:
@DjWarmonger all changes done

rilian-la-te:
@DjWarmonger waiting for your verdict)))

IvanSavenko (approved):

DjWarmonger (approved):

rilian-la-te:
@IvanSavenko can you please merge this, to get #1882 going to review?

PR #1882, opened 2023-04-05
Native terrain refactor

### LINKED ISSUE #1840: ALL_CREATURES bonus is not reset on client when we restart a map
**Describe the bug**
ALL_CREATURES bonus is not reset when we restart a map

**Game logs**
There is entry in server log:
```
[2023-Apr-02 02:42:05.270985] TRACE bonus - #$# Lodestar #propagated to# All creatures
[2023-Apr-02 02:44:43.455437] TRACE bonus - #$# Lodestar #is no longer propagated to# All creatures
```
But in client log  we does not have a second entry (with no longer propagated). WTF?

**To Reproduce**
1. Install Hota
2. Build lodestar
3. Observe "any terrain is native"
4. Out to main menu
5. Start new scenario as Cove again on same map

**Expected behavior**
Cove creatures should not preserve effect of lodestar

**Actual behavior**
Cove creatures will preserve effect of lodestar 

**Did it work earlier?**
Do not know

**Screenshots**
Not applicable

**Version**
 - OS: Linux
 - Version: #1835 

**Additional context**
This bonus is only visual: actually Lodestar stops working on server when map restarts (checked by trying to move creatures on speed 9 with native terrain - their movement was stopped 1 tile before, so, their actual speed was 8), but client still thinks than lodestar is active.

rilian-la-te:
Includes #1876 

Fixes #1840 (but for 1.3)


rilian-la-te:
@IvanSavenko this will break saves, so, if it will be accepted, it should be merged along with #1822

lib/HeroBonus.cpp
```diff
@@ -75,9 +75,8 @@ const std::map<std::string, TLimiterPtr> bonusLimiterMap =
 	{"DRAGON_NATURE", std::make_shared<HasAnotherBonusLimiter>(Bonus::DRAGON_NATURE)},
 	{"IS_UNDEAD", std::make_shared<HasAnotherBonusLimiter>(Bonus::UNDEAD)},
 	{"CREATURE_NATIVE_TERRAIN", std::make_shared<CreatureTerrainLimiter>()},
-	{"CREATURE_FACTION", std::make_shared<CreatureFactionLimiter>()},
+	{"CREATURES_ONLY", std::make_shared<CreatureLevelLimiter>()},
```
IvanSavenko, line 78: Why some limiters were removed even though they are still present in code?

lib/mapObjects/CGHeroInstance.h
```diff
@@ -102,20 +102,6 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
 		}
 	} patrol;
 
-	// deprecated - used only for loading of old saves
```
IvanSavenko, line 105: Just letting you know that I also removed this struct in H3M branch, may lead to conflict unless git manages to resolve it

include/vcmi/Entity.h
```diff
@@ -13,13 +13,23 @@
...
+enum class ETerrainId;
+template<typename T> class Identifier;
 
 class DLL_LINKAGE WithBonuses
 {
 public:
 	virtual const IBonusBearer * getBonusBearer() const = 0;
 };
 
+class DLL_LINKAGE WithNativeTerrain
```
IvanSavenko, line 26: IMO this class alongside with `WithBonuses` need better name...
Any ideas?

lib/CCreatureHandler.h
```diff
@@ -335,9 +327,6 @@ class DLL_LINKAGE CCreatureHandler : public CHandlerBase<CreatureID, Creature, C
 		h & skillLevels;
 		h & skillRequirements;
 		h & commanderLevelPremy;
-		h & allCreatures;
-		h & creaturesOfLevel;
-		BONUS_TREE_DESERIALIZATION_FIX
```
Nordsoft91, line 340: 1) please bump serialisation and min_serialization versions
2) please test multiplayer because I definitely remember that bonus_serialization_fix was one of essential things needed to pass map over network

rilian-la-te: I will bump serialization version when all other complains will be resolved. There is no need to de serialization fix for creature handler now, because now it is not have any bonus nodes inside. All other de serialization fixes are in place.

lib/HeroBonus.cpp
rilian-la-te: Because that limiters will not be functional if initialized by one string without a parameters

include/vcmi/Entity.h
rilian-la-te: What this name should be like?

lib/HeroBonus.cpp
IvanSavenko: Ah, so this is code for parameter-less initialization, while code that provides initialization with parameters still accepts these limiters?

OK then.

include/vcmi/Entity.h
IvanSavenko: Not sure... Perhaps
BonusesProvider / BonusBearerProvider
NativeTerrainProvider
?

Might be better names, but can't think of any at the moment

IvanSavenko: Another option
BonusBearerSource
NativeTerrainSource

XXXProvider might be better though

lib/mapObjects/CGHeroInstance.h
rilian-la-te: You did it as separate commit? If so, I would take this commit as part of rebase, then git will  ignore it after a merge for sure.

include/vcmi/Entity.h
rilian-la-te: If XXXProvider is not used anywhere in other parts of code? If not, I can rename. 

But honestly, I more like names like XXXInterface or WithXXX.

IvanSavenko: I just feel that "WithBonuses" is incomplete name.
EntityWithBonuses is fine, but "WithBonuses" does not sounds right.

IEntityWithBonuses / IEntityWithNativeTerrain would also work IMO.

Not against XXXInterface form, but probably not suitable here.

lib/mapObjects/CGHeroInstance.h
rilian-la-te: No, it is not inside separate commit. @IvanSavenko, can you take this commit inside your H3M branch?

include/vcmi/Entity.h
rilian-la-te: Renamed to IXXXProvider.

lib/CCreatureHandler.h
rilian-la-te: Serialization verson bumped. Network game seems to work on one computer with 2 clients.

lib/BasicTypes.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "GameConstants.h"
+#include "HeroBonus.h"
+
+#include <vcmi/Entity.h>
+#include <vcmi/Faction.h>
+#include <vcmi/FactionService.h>
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+bool INativeTerrainProvider::isItNativeTerrain(TerrainId terrain) const
```
Nordsoft91, line 23:
```suggestion
bool INativeTerrainProvider::isTerrainNative(TerrainId terrain) const
```

lib/BasicTypes.cpp
```diff
@@ -0,0 +1,41 @@
+/*
+ * BasicTypes.cpp, part of VCMI engine
```
Nordsoft91, line 2: Filename basicTypes a bit misleading. Maybe do you have plans to extend it with more functionality? can you share you plans?

lib/CCreatureHandler.cpp
```diff
@@ -692,10 +667,24 @@ std::vector<bool> CCreatureHandler::getDefaultAllowed() const
...
 	if (VLC->settings()->getBoolean(EGameSettings::MODULE_STACK_EXPERIENCE)) 	//reading default stack experience bonuses
 	{
+		logGlobal->debug("\tLoading stack experience bonuses");
+		auto addBonusForAllCreatures = [&](std::shared_ptr<Bonus> b) {
+			auto limiter = std::make_shared<CreatureLevelLimiter>();
+			b->addLimiter(limiter);
+			globalEffects.addNewBonus(b);
+		};
+		auto addBonusForTier = [&](int tier, std::shared_ptr<Bonus> b) {
+			assert(vstd::iswithin(tier, 1, 7));
```
Nordsoft91, line 681: does wot have level 8 creatures?

lib/CCreatureSet.cpp
```diff
@@ -912,6 +915,19 @@ void CStackInstance::serializeJson(JsonSerializeFormat & handler)
 	}
 }
 
+FactionID CStackInstance::getFaction() const
+{
+	if(type)
+		return type->getFaction();
+	else
+		return FactionID::NEUTRAL;
```
Nordsoft91, line 923:
```suggestion
	if(type)
		return type->getFaction();
	return FactionID::NEUTRAL;
```

lib/HeroBonus.h
```diff
@@ -522,6 +523,16 @@ struct DLL_LINKAGE Bonus : public std::enable_shared_from_this<Bonus>
...
 	}
 
+	STRONG_INLINE static ui32 getHighFromSid32(ui32 sid)
+	{
+		return sid >> 16;
+	}
+
+	STRONG_INLINE static ui32 getLowFromSid32(ui32 sid)
+	{
+		return sid & 0xFFFF;
```
Nordsoft91, line 533:
```suggestion
		return sid & 0x0000FFFF;
```

lib/HeroBonus.h
```diff
@@ -522,6 +523,16 @@ struct DLL_LINKAGE Bonus : public std::enable_shared_from_this<Bonus>
 		return (high << 16) + low;
 	}
 
+	STRONG_INLINE static ui32 getHighFromSid32(ui32 sid)
+	{
+		return sid >> 16;
```
Nordsoft91, line 528:
```suggestion
		return sid & 0xFFFF0000;
```

lib/HeroBonus.h
```diff
@@ -1106,12 +1120,13 @@ class DLL_LINKAGE CreatureTerrainLimiter : public ILimiter //applies only to cre
...
 
-class DLL_LINKAGE CreatureFactionLimiter : public ILimiter //applies only to creatures of given faction
+class DLL_LINKAGE CreatureLevelLimiter : public ILimiter //applies only to creatures of given faction
 {
 public:
-	TFaction faction;
-	CreatureFactionLimiter();
-	CreatureFactionLimiter(TFaction faction);
+	uint32_t minLevel;
+	uint32_t maxLevel;
```
Nordsoft91, line 1127: I suggest to use ui32 types as they don't depend on platform (?)

lib/JsonNode.cpp
```diff
@@ -749,15 +749,23 @@ std::shared_ptr<ILimiter> JsonUtils::parseLimiter(const JsonNode & limiter)
...
-					factionLimiter->faction = faction;
+					factionLimiter->faction = FactionID(faction);
 				});
 				return factionLimiter;
 			}
+			else if(limiterType == "CREATURE_LEVEL_LIMITER")
+			{
+				auto levelLimiter = std::make_shared<CreatureLevelLimiter>();
+				levelLimiter->minLevel = parameters[0].Integer();
+				if(parameters[1].isNumber())
```
Nordsoft91, line 765: as I see there is no parameters.size() > 1 condition, is access violation possible?

Nordsoft91 (commented):
I'll test multiplayer later today

lib/CGameState.cpp
```diff
@@ -1838,15 +1839,15 @@ void CGameState::initTowns()
 			int sel = -1;
 
 			for(ui32 ps=0;ps<vti->possibleSpells.size();ps++)
-				total += vti->possibleSpells[ps].toSpell()->getProbability(vti->subID);
+				total += vti->possibleSpells[ps].toSpell()->getProbability(FactionID(vti->subID));
```
IvanSavenko, line 1842: CGTownInstance might benefit from `getFaction` method that returns FactionID(subID), if only for clarity.

lib/JsonNode.cpp
```diff
@@ -991,7 +999,16 @@ bool JsonUtils::parseBonus(const JsonNode &ability, Bonus *b)
 
 	value = &ability["propagator"];
 	if (!value->isNull())
-		b->propagator = parseByMap(bonusPropagatorMap, value, "propagator type ");
+	{
+		//ALL_CREATURES old propagator compatibility
+		if(value->String() == "ALL_CREATURES") 
+		{
+			b->addLimiter(std::make_shared<CreatureLevelLimiter>());
```
IvanSavenko, line 1006: perhaps add warning / conversion information for mods?

lib/JsonNode.cpp
```diff
@@ -749,15 +749,23 @@ std::shared_ptr<ILimiter> JsonUtils::parseLimiter(const JsonNode & limiter)
...
-				std::shared_ptr<CreatureFactionLimiter> factionLimiter = std::make_shared<CreatureFactionLimiter>();
+				std::shared_ptr<FactionLimiter> factionLimiter = std::make_shared<FactionLimiter>();
 				VLC->modh->identifiers.requestIdentifier("faction", parameters[0], [=](si32 faction)
 				{
-					factionLimiter->faction = faction;
+					factionLimiter->faction = FactionID(faction);
 				});
 				return factionLimiter;
 			}
+			else if(limiterType == "CREATURE_LEVEL_LIMITER")
```
IvanSavenko, line 761: Do we have documentation on limiters / propagators for modders somewhere, e.g. on wiki?

If not - might be a good idea to add

lib/HeroBonus.h
IvanSavenko: Not really.
ui32 types are old, vcmi-specific define in Global.h
uint32_t are standard c++ types.

In fact, ui32 and such are now simply define for uint32_t and such.
IMO we should prefer uint32_t in new code and slowly remove ui32 types.

Don't see any reason to actively remove them though (e.g. via mass replace)

lib/HeroBonus.h
IvanSavenko: But that's different effect? Bit shift != bit mask?

IvanSavenko (commented):
Looking at alignment changes - feel free to add it to your bonuses todo list:

refactor NONEVIL_ALIGNMENT_MIX bonus to allow 
1) Any combination of alignmet, e.g. only good factions, only evil factions, etc.
2) Allow to completely ignore alignments (potentialy as special case of mixing neutral, good and evil)

Goal is this: https://heroes.thelazy.net/index.php/Temple_of_Loyalty

(up to you when or where to do this or whether to do at all)

lib/BasicTypes.cpp
rilian-la-te: I want to extract more basic types when I will do HeroBonus.cpp split. 
1. something like ICommonFactionMember (which will have damage, attack, defence and so on)
2. move some common methods from C* types to Entity types (if this methods can be implemented purely only via entity interface, not using any additional state). Of course, methods will remain to be virtual.

lib/CCreatureHandler.cpp
rilian-la-te: @Nordsoft91 level 8 and so creatures was handled by level 7, and this code retain such functionality.

lib/HeroBonus.h
rilian-la-te: @IvanSavenko agree.

lib/HeroBonus.h
rilian-la-te: @IvanSavenko agree.

lib/CGameState.cpp
rilian-la-te: @IvanSavenko so, it should inherit from INativeTerrainProvider?

lib/JsonNode.cpp
rilian-la-te: Yes, will add when PR will be merged.

lib/CGameState.cpp
IvanSavenko: Don't know. Maybe?

Or simply use vti->town->faction instead of subID, should have the same effect.

My goal is to minimize usages of subID's for objects. And here vti->town->faction might be actually better than casting subID.

lib/HeroBonus.h
rilian-la-te: @Nordsoft91 we should not to use & here, because `0xdeadbeef & 0xffff0000 = 0xdead0000`, but `0xdeadbeef >> 16 = 0xdead`, and I need second behaviour.

lib/JsonNode.cpp
rilian-la-te: @IvanSavenko for now,limiters and propagators cannot be handled by converter. Will add a warning.

rilian-la-te:
> Looking at alignment changes - feel free to add it to your bonuses todo list:
> 
> refactor NONEVIL_ALIGNMENT_MIX bonus to allow
> 
>     1. Any combination of alignmet, e.g. only good factions, only evil factions, etc.
> 
>     2. Allow to completely ignore alignments (potentialy as special case of mixing neutral, good and evil)
> 
> 
> Goal is this: https://heroes.thelazy.net/index.php/Temple_of_Loyalty
> 
> (up to you when or where to do this or whether to do at all)

@IvanSavenko maybe just allow parse alignment selector directly for Json? For now, there is a parseSelector method. But selectors cannot be stored, sadly(

IvanSavenko:
> @IvanSavenko maybe just allow parse alignment selector directly for Json? For now, there is a parseSelector method. But selectors cannot be stored, sadly(

Not insisting on any particular implementation as long as:
- Angelic Alliance works as intended
- similar bonuses that permit different alignment mixing are possible (e.g. some sort of "Unholy Alliance" artifact)
- we can make bonus for Temple of Loyalty which allows *any* alignment mixing

At this point I think you have better understanding of bonus system than I do so up to you I guess?

lib/CGameState.cpp
rilian-la-te: done

lib/JsonNode.cpp
rilian-la-te: Added a condition

lib/HeroBonus.h
rilian-la-te: done

lib/CCreatureSet.cpp
rilian-la-te: done

lib/BasicTypes.cpp
rilian-la-te: done

lib/JsonNode.cpp
rilian-la-te: warning done

lib/HeroBonus.h
Nordsoft91: truth, I see those types a part of c++ standard

rilian-la-te: @Nordsoft91 this is linux kernel types, not standard ones)

Nordsoft91: found, that's what I talked about - they are platform dependent:
https://en.cppreference.com/w/cpp/types/integer

> (provided if and only if the implementation directly supports the type)

Nordsoft91:
touched build a bit - no issues found

Nordsoft91 (approved):

rilian-la-te:
@IvanSavenko did you like this PR?

IvanSavenko (approved):

include/vcmi/Entity.h
```diff
@@ -13,13 +13,33 @@
...
+class DLL_LINKAGE IConstBonusProvider
 {
 public:
 	virtual const IBonusBearer * getBonusBearer() const = 0;
 };
 
+class DLL_LINKAGE INativeTerrainProvider
+{
+public:
+	virtual Identifier<ETerrainId> getNativeTerrain() const = 0;
```
Nordsoft91, line 29: Don't we have TerrainId anymore? ETerrainId is enum to support legacy terrains only. What are new terrains?

rilian-la-te: @Nordsoft91 TerrainId is Identifier<ETerrainId>, but type alias cannot be froward declared, so, I used full definition here.

PR #1886, opened 2023-04-06
vcmi: fix rmg header levels

### LINKED ISSUE #1884: In Scenario description there is no number of map levels.
**Describe the bug**
In Scenario description there is empty space where current map number of levels must be shown.

**To Reproduce**
1. Start random map
2. Go to Info window or save the game and open load game dialogue
 

**Expected behavior**
there must be "levels 1" or "levels 2" (if map is with dungeon) text (when additional levels will be implemented, there must be actual number of levels (1 for surface + number of underground levels)

**Actual behavior**
"levels " (no number of levels)

**Did it work earlier?**
Dont remember

**Screenshots**
image

rilian-la-te:
It was printed as char, not as int.

Fixes #1884


krs0:
Wouldn' code like this be more easy to read, if a couple of variables would have been used?

rilian-la-te:
@krs0 this is extracted from #1623, and in time when I did it I did not bother for readability, initial it was a hotfix for unit test passing.

Issue #1885, opened 2023-04-06
Can't close template selection list without choosing a template


DjWarmonger:
**Describe the bug**
Once you click on the drop-down template list in extended RMG, it is not possible to hide it without actually sellecting the template.

**To Reproduce**
Did not try.

**Expected behavior**
Clicking anywhere else on the screen would cause combo box to return to its last state.

**Actual behavior**
Nothing but the list can be clicked.

**Screenshots**
![image](https://user-images.githubusercontent.com/2566990/230334047-3a257fd5-fbbb-4464-96df-38afaa1bc607.png)

**Version**
Beta

**Additional context**
Reported by zuuzel

PR #1887, opened 2023-04-06
Fix: close dropbox with left mouse click outside of the dropbox

Adriankhl:
Fixes https://github.com/vcmi/vcmi/issues/1885


dydzio0614:
obraz
also handled? :P

Nordsoft91:
Tested on ios - works well

Nordsoft91 (approved):

Issue #983, opened 2022-09-24
[feature] new UI for melee attack on touch screens


kambala-decapitator:
This is primarily for iOS and Android.

Currently it's rather difficult to attack from the desired angle with a finger.

kambala-decapitator:
discussion from Slack:

btw main problem in ios and android is melee battle, something to solve in a nice way


Danylchenko
  3:53 PM
I think with gestures


kambala
  3:53 PM
like extra dialog where you can conveniently choose the attacking side


Nordsoft
  3:53 PM
agree, not convenent for now, need something like double touch (one to set direction, econd to confirm)


kambala
  3:53 PM
yeah


Nordsoft
  3:53 PM
of some zoom widget


Danylchenko
  3:53 PM
swipe from target tile towards enemy


kambala
  3:54 PM
it can be an advanced way to target, might be difficult to perform the right gesture


Danylchenko
  3:55 PM
put finger on battle hex near and swipe on enemy?


kambala
  3:56 PM
replied to a thread:
put finger on battle hex near and swipe on enemy?
what about 2-hex creatures when you want to attack “in between”? ;)


Danylchenko
  3:57 PM
Swipe on correct hex. Also we can enable preview mod. Like you first swipe and see in battle log estimated damage and only after this you confirm pushing button


kambala
  3:58 PM
that’s a separate UI topic for brainstorming
3:58
I really wouldn’t do gesture-only (edited) 


Nordsoft
  3:58 PM
but definetely action confirmation should be on touch release, not on touch begin


Danylchenko
  3:58 PM
Another way is swipe from enemy to target hex


Nordsoft
  3:58 PM
touch begin should be like mouse move


Danylchenko
  3:59 PM
Yes. This needs to be coded)
4:00
Extra dialog also possible but on phone there is possibility to even miss enemy and move towards it instead of attack)
4:00
Because fingers are quite big)


kambala
  4:00 PM
that’s what confirmation dialog would be for :)


Danylchenko
  4:00 PM
In HD mod UI highlights affected creatures


kambala
  4:00 PM
which should be a toggleable option of course

kambala-decapitator:
example from Heroes 3 HD for iOS: https://www.youtube.com/watch?v=c-KIC6UgIWo

in the beginning of the video there's tutorial, then on 2:17 battle starts and tutorial is displayed again.

- ranged: double tap an enemy
- melee: first, tap on enemy - attacking hexes are highlighted, then tap on one of the hexes to attack

[...]

PR #1889, opened 2023-04-06
Touchscreen mode for battlefield

Nordsoft91:
Implements https://github.com/vcmi/vcmi/issues/983

Probably not juicy as it described, but
1. It solves problem
2. Better than nothing

Appreciate for testing and feedback

### DIFF
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -102,6 +102,7 @@ void FirstLaunchView::on_comboBoxLanguage_currentIndexChanged(int index)
 
 void FirstLaunchView::enterSetup()
 {
+	setupPlatformSettings();
 	Languages::fillLanguages(ui->listWidgetLanguage, false);
 }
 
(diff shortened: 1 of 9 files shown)


launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -158,6 +159,19 @@ void FirstLaunchView::exitSetup()
 		mainWindow->exitSetup();
 }
 
+// Initial platform-dependend settings
+void FirstLaunchView::setupPlatformSettings()
```
IvanSavenko, line 163: Android currently uses separate launcher so this code won't work on Android.

I actually had same issue recently - ended up creating separate options ( "swipe" and "swipeDesktop") for desktop/mobile platforms, with desktop having set it to false by default and on mobile - set to true.

Ugly workaround, sure, but same can be said about this code. Not sure which option is better.

IvanSavenko:
Was planning on doing this myself for 1.3 if nobody else picks it up. And using similar approach as yours.

Better than nothing, sure, although I would've added some UI indication, e.g. semi-transparent unit shadow to show some reaction on 1st click.

launcher/firstLaunch/firstlaunch_moc.cpp
kambala-decapitator: > Android currently uses separate launcher so this code won't work on Android.

looking forward to solving that for 1.3, already have some initial work done

Nordsoft91:
But it will be reaction at 1st click: as you don't have mouse move on touchscreens, first click means hovering hex with visible intention

IvanSavenko:
> But it will be reaction at 1st click: as you don't have mouse move on touchscreens, first click means hovering hex with visible intention

Well, good enough for first iteration I guess, can be improved later. 

Shadow image would still be much better indication, but I guess not critical for now.

launcher/firstLaunch/firstlaunch_moc.cpp
IvanSavenko: We would probably still need possibility to defined default settings per platform - we already have cursor, map swiping/panning, and now UI for combat.

Will add this to my 1.3 todo-list, should be simple enough

IvanSavenko (approved):

Nordsoft91:
tested - works well

PR #1893, opened 2023-04-07
Add missing include headers when building with -DENABLE_PCH=OFF

trofi:
Example build failure is missing `VCMI_LIB_NAMESPACE_BEGIN` declaration:

    [  0%] Building CXX object lib/CMakeFiles/vcmi.dir/spells/ObstacleCasterProxy.cpp.o
    In file included from lib/spells/ProxyCaster.h:13,
                     from lib/spells/ObstacleCasterProxy.h:11,
                     from lib/spells/ObstacleCasterProxy.cpp:11:
    lib/../include/vcmi/spells/Caster.h:13:1: error: 'VCMI_LIB_NAMESPACE_BEGIN' does not name a type
       13 | VCMI_LIB_NAMESPACE_BEGIN
          | ^~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/lib/spells/ObstacleCasterProxy.cpp
+++ b/lib/spells/ObstacleCasterProxy.cpp
@@ -8,6 +8,7 @@
  *
  */
 
+#include "StdInc.h"
 #include "ObstacleCasterProxy.h"
 
 VCMI_LIB_NAMESPACE_BEGIN


rilian-la-te:
@trofi you should not include `Global.h` there.

@IvanSavenko maybe move VCMI_LIB_NAMESPACE_* definitions into include dir?

IvanSavenko (commented):
ObstacleCasterProxy.cpp (and maybe others) is missing `#include "StdInc.h"` as the very first include
Should be fixed in cpp, not in header.

> @IvanSavenko maybe move VCMI_LIB_NAMESPACE_* definitions into include dir?

Why? In this particular case it won't help - compilation would fail on first reference to anything from std or boost, since those files are only included in global.h

trofi:
Aha, sounds good. Added `#include "StdInc.h"` to `ObstacleCasterProxy.cpp`.

rilian-la-te:
@trofi thanks, forgot about that StdInc.h

PR #1899, opened 2023-04-07
Rewardable expand pt1

Nordsoft91:
Random selection for
 - resources
 - primary stats
 - secondary skills

PR allows witchhut to be rewardable

If approved, I'll update wiki description

### DIFF
--- a/lib/mapObjects/JsonRandom.cpp
+++ b/lib/mapObjects/JsonRandom.cpp
@@ -20,6 +20,7 @@
 #include "../CCreatureHandler.h"
 #include "../CCreatureSet.h"
 #include "../spells/CSpellHandler.h"
+#include "../CSkillHandler.h"
 
 VCMI_LIB_NAMESPACE_BEGIN
 
@@ -136,11 +137,40 @@ namespace JsonRandom
 	std::map<SecondarySkill, si32> loadSecondary(const JsonNode & value, CRandomGenerator & rng)
 	{
 		std::map<SecondarySkill, si32> ret;
+		std::vector<SecondarySkill> except;
+		
 		for(const auto & pair : value.Struct())
 		{
-			SecondarySkill id(VLC->modh->identifiers.getIdentifier(pair.second.meta, "skill", pair.first).get());
-			ret[id] = loadValue(pair.second, rng);
+			std::string skillName = pair.first;
+			
+			if(skillName == "random")
+			{
+				if(pair.second.isNumber())
+					skillName = RandomGeneratorUtil::nextItem(VLC->skillh->objects, rng)->get()->getNameTextID();
+				else
+				{
+					auto skill = *RandomGeneratorUtil::nextItem(loadSecondary(pair.second, rng), rng);
+					ret[skill.first] = skill.second;
+				}
+			}
+			else if(skillName == "except")
+			{
+				for(auto & i : pair.second.Vector())
+				{
+					SecondarySkill id(VLC->modh->identifiers.getIdentifier(pair.second.meta, "skill", i.String()).get());
+					except.push_back(id);
+				}
+			}
+			else
+			{
+				SecondarySkill id(VLC->modh->identifiers.getIdentifier(pair.second.meta, "skill", pair.first).get());
+				ret[id] = loadValue(pair.second, rng);
+			}
 		}
+		
+		for(auto & i : except)
+			ret.erase(i);
+		
 		return ret;
 	}
 
--- a/lib/mapObjects/JsonRandom.cpp
+++ b/lib/mapObjects/JsonRandom.cpp
@@ -89,6 +89,26 @@ namespace JsonRandom
 	{
 		std::string resourceName = loadKey(value, rng, "");
 		si32 resourceAmount = loadValue(value, rng, 0);
+		
+		if(!value["simple"].isNull())
+		{
+			si32 index = *RandomGeneratorUtil::nextItem(std::vector<si32>{Res::WOOD, Res::ORE}, rng);
+			resourceName = GameConstants::RESOURCE_NAMES[index];
+		}
+		if(!value["precious"].isNull())
+		{
+			si32 index = *RandomGeneratorUtil::nextItem(std::vector<si32>{Res::MERCURY, Res::GEMS, Res::SULFUR, Res::CRYSTAL}, rng);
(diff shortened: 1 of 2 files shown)


lib/mapObjects/JsonRandom.cpp
```diff
@@ -89,6 +90,26 @@ namespace JsonRandom
 	{
 		std::string resourceName = loadKey(value, rng, "");
 		si32 resourceAmount = loadValue(value, rng, 0);
+		
+		if(!value["simple"].isNull())
```
IvanSavenko, line 94: But this is already supported without hardcoded resources list:
`"list" : [ "wood", "ore" ]`
`"list" : [ "mercury", "gems", "sulfur", "crystal" ]`

IvanSavenko:
> PR allows witchhut to be rewardable
Unfortunately no.
1) Witch hut needs support for loading preset skill from map settings
2) Witch hut needs support for skill name in status bar after "scouting" object
3) Witch hut needs support to show skill name if hero already have it

lib/mapObjects/JsonRandom.cpp
```diff
@@ -104,17 +125,52 @@ namespace JsonRandom
...
 	std::map<SecondarySkill, si32> loadSecondary(const JsonNode & value, CRandomGenerator & rng)
 	{
 		std::map<SecondarySkill, si32> ret;
+		std::vector<SecondarySkill> except;
+		
 		for(const auto & pair : value.Struct())
 		{
-			SecondarySkill id(VLC->modh->identifiers.getIdentifier(pair.second.meta, "skill", pair.first).get());
-			ret[id] = loadValue(pair.second, rng);
+			std::string skillName = pair.first;
```
IvanSavenko, line 144: Perhaps both primary and secondary can be implemented similar to 'loadResources' / 'loadKey' method, which would allow format like:
```js
[
    {    // gives 1 level of Wisdom skill...
        "type" : "wisdom", 
        "value" : 1 
    },
    {    // ... and 1-2 levels of a randomly selected magic school
        "list" : [ "airMagic", "waterMagic", "earthMagic", "fireMagic" ],
        "min" : 1, "max" 2
    }
]
```
I think we should eventually switch to vectors for all reward types where it makes sense.
And we should also account for skills that are potentially banned on the map.

Perhaps:
- if json uses struct / object form then load as before (and maybe consider than form deprecated)
- if json uses vector / list then for each entry in vector one skill will be generated:
- if entry contains 'type' then this skill will be selected
- else, if entry contains 'list' then one of these skills will be selected randomly (taking into account banned skills?)
- else, any non-banned random skill will be selected
- also, if entry contains 'except' entry then these skills should be except from selection

rilian-la-te: @IvanSavenko but why vectors? Maybe struct of stucts to allow override in mods?

IvanSavenko: Well, these vectors are already inside of another, much larger vector, so not much point. And replacing it is not trivial - we have order-dependent logic with it.

In long term - might make sense to replace. But don't see high priority. For now mods will have to replace entire rewards list.

lib/mapObjects/JsonRandom.cpp
Nordsoft91: yes, it's more like syntax sugar. If we don't need this, I'll remove

IvanSavenko: Probably not needed.
I have an idea to introduce "traits" to objects so you can (for example) pick objects based on said trait, for example pick an artifact that has traits "relic" and "rightHand" to pick random relic sword.

In that case resources might get "common" or "rare" traits. But right now this is just an idea with rather low priority. Don't see need for it at the moment

lib/mapObjects/JsonRandom.cpp
Nordsoft91: @IvanSavenko , I like this concept, the only headache I have - how to bypass map restrictions.
I'm thinking on adding global pointer to active map, is there better ideas?

IvanSavenko: On naming - probably better to rename for consistency with other options which use similar naming:
- "list" -> "anyOf" ("oneOf" is also an option, but anyOf probably sounds better)
- "except" -> "noneOf"

Map restrictions - probably RewardableConstructor should pass reference to appropriate `std::vector<bool>` from CMap to JsonRandom functions, without passing entire CMap. Don't see any better way right now. 

There is (almost) suitable method that JsonRandom could have used - getDefaultAllowed() from handlers, but that method returns global settings, not map-specific.

lib/mapObjects/JsonRandom.cpp
```diff
@@ -87,7 +95,9 @@ namespace JsonRandom
 
 	TResources loadResource(const JsonNode & value, CRandomGenerator & rng)
 	{
-		std::string resourceName = loadKey(value, rng, "");
+		std::set<std::string> defaultResources(GameConstants::RESOURCE_NAMES, GameConstants::RESOURCE_NAMES + GameConstants::RESOURCE_QUANTITY - 2); //except mithril
```
IvanSavenko, line 98: You can use std::begin / std::end instead of pointer arithmetics.
Also, I think this should be '-1', not '-2'. Or not?
```suggestion
		std::set<std::string> defaultResources(std::begin(GameConstants::RESOURCE_NAMES), std::end(GameConstants::RESOURCE_NAMES) - 1); //except mithril
```

lib/mapObjects/JsonRandom.cpp
```diff
@@ -104,16 +114,40 @@ namespace JsonRandom
 		{
 			ret.push_back(loadValue(value[name], rng));
 		}
+		
+		if(!value["random"].isNull())
```
IvanSavenko, line 118: Can probably get same treatment as resources/secondaries, for consistency

Nordsoft91: Agree, done

lib/mapObjects/JsonRandom.cpp
Nordsoft91: good catch, fixed

Nordsoft91:
wiki updated, will merge after building

IvanSavenko (approved):

PR #1717, opened 2023-03-18
Introducing backpack size limit functionality

SoundSSGood:
- [x] Backpack volume check. Move.
- [x] Backpack volume check. Bulk move.
- [x] Backpack volume check. Disassemble in the backpack.
- [x]  Backpack volume check. Get new art inst (Pick up, Dig up, Events, Object rewards).
- [x] Backpack volume check. End of battle.
- [x] Backpack volume check. Starting bonus.
- [x]  Backpack volume check. Using cheats.
- [x] Backpack volume check. Buying in market.
- [x]  Cleanup code.

Messages translation is out of scope.
Removal of arts by client (It is possible to do on heroes exchange window in OH3) is out of scope. Next PR.

PR #1908, opened 2023-04-07
Artifact related fixes

SoundSSGood:
- Adds missed in https://github.com/vcmi/vcmi/pull/1717 ability to remove artifacts by client.
- Now "vcmiartifacts" doesn't spam askAssembleArtifact windows

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -654,7 +654,7 @@ CArtifactsOfHero::~CArtifactsOfHero()
 		}
 		else
 		{
-			//TODO remove artifact
+			LOCPLINT->cb->eraseArtifactByClient(ArtifactLocation(curHero, ArtifactPosition::TRANSITION_POS));
 		}
 	}
 }
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -4063,6 +4063,23 @@ bool CGameHandler::assembleArtifacts (ObjectInstanceID heroID, ArtifactPosition 
 	return true;
 }
 
+bool CGameHandler::eraseArtifactByClient(const ArtifactLocation & al)
+{
+	const auto * hero = getHero(al.relatedObj()->id);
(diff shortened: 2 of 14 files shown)


client/widgets/CArtifactHolder.cpp
```diff
@@ -654,7 +654,7 @@ CArtifactsOfHero::~CArtifactsOfHero()
 		}
 		else
 		{
-			//TODO remove artifact
+			LOCPLINT->cb->eraseArtifactByClient(ArtifactLocation(curHero, ArtifactPosition::TRANSITION_POS));
```
IvanSavenko, line 657: To clarify - this happens if we exit hero window with picked up artifact and have no place to put artifact down, e.g. full backpack. Correct?

server/CGameHandler.cpp
```diff
@@ -4063,6 +4063,23 @@ bool CGameHandler::assembleArtifacts (ObjectInstanceID heroID, ArtifactPosition
...
+	const auto * hero = getHero(al.relatedObj()->id);
+	if(hero == nullptr)
+		COMPLAIN_RET("eraseArtifactByClient: wrong hero's ID");
+
+	const auto * art = al.getArt();
+	if(art == nullptr)
+		COMPLAIN_RET("Cannot remove artifact!");
+
+	if(al.getArt()->artType->canBePutAt(hero) || al.slot != ArtifactPosition::TRANSITION_POS)
+		COMPLAIN_RET("Unlegal removing artifact");
```
IvanSavenko, line 4077:
```suggestion
		COMPLAIN_RET("Illegal artifact removal request");
```

client/widgets/CArtifactHolder.cpp
SoundSSGood: Right. Easy to reproduce with war machines. Pick up the war machine on the exchange screen. And do swap on another hero's war machine. Close the window without placing art in the slot.

IvanSavenko (approved):

PR #1913, opened 2023-04-08
Fix initialization of spell effects registry

### LINKED ISSUE #1869: [Android]Failed to use Cure spell, prompt "Unknown problem with this spell no more information available"
Describe the bug
Android, new est daily builds (VCMI brand development e1c3550), Cure spell, Resurrection spell, and all similar spells fail. First Aid Tent cannot work

Game logs
Not applicable

To Reproduce

Using Cure spell，or using Firstaidtent

Actual behavior
Using Cure spell fail.
It prompts"Unknown problem with this spell no more information available"
First Aid Tent will be skipped.

Did it work earlier?
VCMI brand development 6c693f2 and previous versions can work properly

Screenshots
Error_Cure

Version

OS: Android MIUI 14.0.3
Version: VCMI-branch-develop-e1c3550
Additional context

IvanSavenko:
Fixes #1869 

And this is why boys and girls you should not try to do too smart stuff that is not guaranteed by the standard:

> It is implementation-defined whether or not the dynamic initialization (dcl.init, class.static, class.ctor, class.expl.init) of an object of namespace scope is done before the first statement of main. If the initialization is deferred to some point in time after the first statement of main, _it shall occur before the first use of any function or object defined in the same translation unit_ as the object to be initialized.

For reference, damage and timed effects worked because those effects are directly instantiated by server. Meanwhile compiler dropped all other effects because they were not referenced anywhere in the code.

Removed static-initialization idiocity and moved registration of effects to registry construtor.
Fixed bug for me on single-app build, presumably should also fix it on Android

### DIFF
--- a/client/LobbyClientNetPackVisitors.h
+++ b/client/LobbyClientNetPackVisitors.h
@@ -12,7 +12,9 @@
 #include "../lib/NetPackVisitor.h"
 
 class CClient;
+VCMI_LIB_NAMESPACE_BEGIN
 class CGameState;
+VCMI_LIB_NAMESPACE_END
 
 class ApplyOnLobbyHandlerNetPackVisitor : public VCMI_LIB_WRAP_NAMESPACE(ICPackVisitor)
 {
--- a/lib/spells/effects/Catapult.cpp
+++ b/lib/spells/effects/Catapult.cpp
@@ -23,14 +23,10 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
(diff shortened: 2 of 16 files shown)


rilian-la-te:
Where is a moat?

IvanSavenko:
What moat?
Beta branch.

kambala-decapitator:
thanks! I'll test tomorrow on android

kambala-decapitator:
now works on android too!

PR #1915, opened 2023-04-08
VCMI campaigns format

Nordsoft91:
Basically it works, but want to make format more suitable
Advices about format are welcomed.
Work in this PR is limited by parser - minimal changes in existing campaign classes

Example of playable test here (remove txt extension and place to maps folder)
[tstfile.vcmp.txt](https://github.com/vcmi/vcmi/files/11184446/tstfile.vcmp.txt)

- [x] string ids for building bonuses
- [x] string ids for spell scrolls 
- [x] string ids for primary skills
- [x] string ids for secondary skills
- [x] monsters kept by hero
- [x] artifacts kept by hero
- [x] encapsulating maps (zip?) https://github.com/vcmi/vcmi/issues/1921
- [x] (stretch goal) custom music support https://github.com/vcmi/vcmi/issues/1921
- [x] Logging and error handling
- [x] Embed map background/regions into vcmp https://github.com/vcmi/vcmi/issues/1921
- [ ] Carryover heroes between campaigns (blueprint) https://github.com/vcmi/vcmi/issues/1921

Outside of this PR:
 - Campaign editor integrated into map editor
 - custom video support
 - refactoring of existing classes


Nordsoft91:
It has most of functionality and fine for first revision

lib/mapping/CCampaignHandler.h
```diff
@@ -29,8 +29,12 @@ namespace CampaignVersion
 		RoE = 4,
 		AB = 5,
 		SoD = 6,
-		WoG = 6
+		WoG = 6,
+		VCMI = 1
```
IvanSavenko, line 33: Perhaps use larger number than SoD (+some buffer for theoretical hota format)

Or even set vcmi format to '130' to match vcmi version (1.3.0)

lib/mapping/CCampaignHandler.h
```diff
@@ -221,7 +226,13 @@ class DLL_LINKAGE CCampaignHandler
 	std::vector<size_t> scenariosCountPerCampaign;
 
 	static std::string readLocalizedString(CBinaryReader & reader, std::string filename, std::string modName, std::string encoding, std::string identifier);
+	
+	//parsers for VCMI campaigns (*.vcmp)
+	static CCampaignHeader readHeaderFromJson(JsonNode & reader, std::string filename, std::string modName, std::string encoding);
+	static CCampaignScenario readScenarioFromJson(JsonNode & reader, std::string filename, std::string modName, std::string encoding, int version, int mapVersion );
+	static CScenarioTravel readScenarioTravelFromJson(JsonNode & reader, int version);
```
IvanSavenko, line 233: We really need stateful campaign handler/parser instead of this static forestry

(not necessarily in this PR)

lib/mapping/CCampaignHandler.cpp
```diff
@@ -151,6 +180,257 @@ std::string CCampaignHandler::readLocalizedString(CBinaryReader & reader, std::s
...
+
+	ret.version = reader["version"].Integer();
+	if(ret.version < CampaignVersion::VCMI_MIN || ret.version > CampaignVersion::VCMI_MAX)
+	{
+		logGlobal->info("VCMP Loading: Unsupported campaign %s version %d", filename, ret.version);
+		return ret;
+	}
+	
+	ret.version = CampaignVersion::VCMI;
+	ret.mapVersion = reader["campaignId"].Integer();
```
IvanSavenko, line 195: IMO we should eventually use mapVersion only for h3c's

And for vcmi campaigns embed all corresponding data (e.g. all names of images) into vcmp file.

This is needed to support Heroes Chronicles, HotA, and likely to be requested feature by map makers.

lib/mapping/CCampaignHandler.cpp
```diff
@@ -151,6 +180,257 @@ std::string CCampaignHandler::readLocalizedString(CBinaryReader & reader, std::s
...
+	ret.name = reader["name"].String();
+	ret.description = reader["description"].String();
+	ret.difficultyChoosenByPlayer = reader["allowDifficultySelection"].Bool();
+	//skip ret.music because it's unused in vcmi
+	ret.filename = filename;
+	ret.modName = modName;
+	ret.encoding = encoding;
```
IvanSavenko, line 202: OK to keep, but should be unused for vcmp - json's are always encoded as utf-8, no need to support/consider H3 language-specific encodings

lib/mapping/CCampaignHandler.cpp
```diff
@@ -151,6 +180,257 @@ std::string CCampaignHandler::readLocalizedString(CBinaryReader & reader, std::s
...
+CScenarioTravel CCampaignHandler::readScenarioTravelFromJson(JsonNode & reader, int version )
+{
+	CScenarioTravel ret;
+
+	std::map<std::string, ui8> heroKeepsMap = {
+		{"experience", 1},
+		{"primarySkill", 2},
+		{"secondarySkill", 4},
+		{"spells", 8},
+		{"artifacts", 16}
```
IvanSavenko, line 248: Do these constants have enumeration somewhere? If not - they should IMO

(same for other maps below)

dydzio0614:
offtopic but support for more than 3 starting bonuses can be probably implemented via horizontal scrollbar - convenient enough if somebody doesn't come up with idea to add 20 starting bonuses

lib/mapping/CCampaignHandler.cpp
```diff
@@ -151,6 +180,257 @@ std::string CCampaignHandler::readLocalizedString(CBinaryReader & reader, std::s
...
+	
+	for(auto & k : reader["heroKeeps"].Vector())
+		ret.whatHeroKeeps |= heroKeepsMap[k.String()];
+	
+	for(auto & k : reader["keepCreatures"].Vector())
+	{
+		if(auto identifier = VLC->modh->identifiers.getIdentifier(CModHandler::scopeMap(), "creature", k.String()))
+		{
+			int creId = identifier.get();
+			ret.monstersKeptByHero[creId / 8] |= (1 << creId % 8);
```
IvanSavenko, line 305: this is std::array at the moment. Meaning, it will overflow & crash for creatures from mods

(same for artifsKeptByHero below)

IvanSavenko (commented):
Up to you whether to fix them in this PR or later

Nordsoft91:
> offtopic but support for more than 3 starting bonuses can be probably implemented via horizontal scrollbar - convenient enough if somebody doesn't come up with idea to add 20 starting bonuses

It's supported by format. But I didn't change anything in ui

lib/mapping/CCampaignHandler.h
Nordsoft91: This number doesn't impact anything - it just should be unique in this particular build. It event shall not be preserved with releases.
Real numbers from json are compared with CampaignVersion::VCMI_MIN and CampaignVersion::VCMI_MAX

lib/mapping/CCampaignHandler.h
Nordsoft91: We may consider adding this PR into 1.2 because it's save (h3c maps follows same code as earlier)
It's could be useful for Antagarich Burning campaign

For now idea is to freeze a vcmi file format. Internally for now it's transformed into "h3c format".
Later we should do vice versa

lib/mapping/CCampaignHandler.cpp
Nordsoft91: This field is unused for now and here just for consistency

lib/mapping/CCampaignHandler.cpp
Nordsoft91: enumeration for strings?
All those maps are temporary to transform data to internal legacy h3c format.
Take a look at campToERMU methods - h3c has very specific format (bit masks, etc)

IvanSavenko: Not for strings. But for bit masks from H3, e.g.
```c++
enum EHeroKeeps
{
    EXPERIENCE = 1,
    PRIMARY_SKILLS = 2,
    ....
}
```
Probably needs to be wrapped in namespace to avoid leaking into global namespace - since proper enum class won't be suitable here.

lib/mapping/CCampaignHandler.h
IvanSavenko: In that case CampaignVersion::VCMI_MIN and CampaignVersion::VCMI_MAX may get such values.
'Format 130' is more meaningful than simple incrementation whenever it changes.
Don't see anything wrong with having (for example) format 130 followed by 140 or whatever vcmi version will be when this version gets bumped - they don't have to be continuous.

Not insisting on this though, so up to you.

Nordsoft91: I don't get the idea: what should be maximum supported version for VCMI 1.3?

IvanSavenko: VCMI 1.3 would use versions starting from 130 (and up to how many format changes we'll get during 1.3 development)

If we need to make format change in vcmi 1.4 it will immediately get bumped to '140', even if '130' ended up final format for 1.3

lib/mapping/CCampaignHandler.cpp
Nordsoft91: Again, we won't use them later. I don't think we need enums which are used in only one place

dydzio0614:
one nice feature to consider would be carryover of heroes between campaigns (implemented in H4 editor)
obraz

Nordsoft91:
@dydzio0614 , I think it's nice feature. It also can be used as enemy hero. Like in campaign A you play for castle and have very strong hero till the end; then you play campaign B for inferno and you should fight against hero from camp A. But I barely imagine how to implement it

lib/mapping/CCampaignHandler.h
Nordsoft91: so, how vcmi-1.3 will know whether it can support version 140 or not?

IvanSavenko: Same as it is now. vcmi-1.3 will have VCMI_MAX = 130 (or 131/132/... depending on how many times format was changed during 1.3 development)

The only difference is that instead of keeping continious versioning for 1.4 format changes we'll bump format immediately to 140, skipping whatever left in 130-139 range

lib/mapping/CCampaignHandler.cpp
Nordsoft91: this feature is out of this PR, because needs changes outside parser, in vcmiclient.
For now the workaround is to add campaign_regions.json to the mod with overrides

dydzio0614:
I think h4 uses dedicated data file for serialized heroes from all marked campaigns and if you win campaign marked as "can carryover from this campaign" then entry is created on win with key (most likely campaign filename), and if some other campaign has toggled on "carryover from another campaign named X" then if entry with key "X" exists then hero data is loaded from file, otherwise level 1 is used as placeholder, would probably require extension of "hero placeholder" functionality.

Nordsoft91:
capturing ideas and future work here https://github.com/vcmi/vcmi/issues/1921

lib/mapping/CCampaignHandler.h
Nordsoft91: it means, that vcmi-1.3 won't be able to open maps of version 140, which is not what we probably want.
I would thinking about not changing versions or changing it as rare as possible. Which means, that we may introduce couple new fields, but since 1.3 can open it - it still version 1.

Nordsoft91: also, I don't think we want to have one more version we need to change frequently. At some point we may need to break backward compatibility, but hopefully it won't be soon.

client/lobby/SelectionTab.cpp
```diff
@@ -534,6 +534,13 @@ void SelectionTab::parseMaps(const std::unordered_set<ResourceID> & files)
 	{
 		try
 		{
+			const std::string filename_prefix = file.getName().substr(file.getName().find_last_of("/\\") + 1, 2);
+			if(filename_prefix == "__")
```
IvanSavenko, line 538: I presume this is to hide campaign maps?
We can also use different directory for them - only maps from "Maps" directory are visible in selection tab
a) use existing Data directory
b) create separate directory for campaigns, e.g. Campaigns and load campaigns & their maps from there

Or another option - use different extension for campaign scenario so selection tab won't pick them up

dydzio0614: or maybe use maps in zip package for campaigns like map editor does for maps?

rilian-la-te:
@dydzio0614 this will be a nice feature even for Unholy Alliance campaign (to carryover a heroes from other SoD campaigns).

Alucard648:
I am suggesting zip with no or minimal compression for encapsulating maps in campaigns.

client/lobby/SelectionTab.cpp
Nordsoft91: I was thinking that custom campaigns shall be as separate mods.
Campaign mod can contain custom data, music, objects, specific configurations and of course maps.
The only thing is that we don't want to show those maps in scenario list.

Nordsoft91: We already have launcher to as content distributing system, so better to use that, than create something new

dydzio0614: I rather see campaigns requiring mods as something similar to maps requiring mods, with exception that campaigns will appear on custom campaign list rather than on map list

dydzio0614: also what about reusing campaign content in map editor and random map generator etc.

Nordsoft91: Fixed

Nordsoft91:
Merge?

lib/mapping/CCampaignHandler.cpp
```diff
@@ -35,27 +36,73 @@
...
+	rd.infix = node["infix"].String();
+	rd.xpos = static_cast<int>(node["x"].Float());
+	rd.ypos = static_cast<int>(node["y"].Float());
+	return rd;
+}
+
+CampaignRegions CampaignRegions::fromJson(const JsonNode & node)
+{
+	CampaignRegions cr;
+	cr.campPrefix = node["prefix"].String();
```
IvanSavenko, line 51: Personally I would've used full names in config instead of runtime generation, at least for vcmi campaings.
Not critical though

lib/mapping/CCampaignHandler.cpp
```diff
@@ -151,45 +201,292 @@ std::string CCampaignHandler::readLocalizedString(CBinaryReader & reader, std::s
...
+		return ret;
+	}
+	
+	ret.version = CampaignVersion::VCMI;
+	ret.campaignRegions = CampaignRegions::fromJson(reader["regions"]);
+	ret.numberOfScenarios = reader["scenarios"].Vector().size();
+	ret.name = reader["name"].String();
+	ret.description = reader["description"].String();
+	ret.difficultyChoosenByPlayer = reader["allowDifficultySelection"].Bool();
+	//skip ret.music because it's unused in vcmi
```
IvanSavenko, line 221: Perhaps implement it finally?

lib/mapping/CCampaignHandler.cpp
```diff
@@ -151,45 +201,292 @@ std::string CCampaignHandler::readLocalizedString(CBinaryReader & reader, std::s
...
+		if(k.String() == "experience") ret.whatHeroKeeps.experience = true;
+		if(k.String() == "primarySkills") ret.whatHeroKeeps.primarySkills = true;
+		if(k.String() == "secondarySkills") ret.whatHeroKeeps.secondarySkills = true;
+		if(k.String() == "spells") ret.whatHeroKeeps.spells = true;
+		if(k.String() == "artifacts") ret.whatHeroKeeps.artifacts = true;
+	}
+	
+	for(auto & k : reader["keepCreatures"].Vector())
+	{
+		if(auto identifier = VLC->modh->identifiers.getIdentifier(CModHandler::scopeMap(), "creature", k.String()))
```
IvanSavenko, line 323: Keep in mind that map scope is not implemented and aliased to "game" scope.
Meaning, it will resolve identifiers even if they are not from dependent mod.

Not implemented since maps didn't have list of dependent mods in its format.

IvanSavenko (approved):

lib/mapping/CCampaignHandler.cpp
Nordsoft91: I don't know where it should be used, like literally in which screen:)

lib/mapping/CCampaignHandler.cpp
Nordsoft91: I added it into vmap in this PR. Could you please check changes in CMap and CMapService? Maybe I did it not in supposed way

lib/mapping/CCampaignHandler.cpp
IvanSavenko: From what I remember - this is the music that should play while campaigns scenario screen is open (with map & starting bonuses selection), in place of main menu track

lib/mapping/CCampaignHandler.cpp
IvanSavenko: ModHandler still needs to use information provided by CMap / CMapService. This would need following changes:
- CModHandler::scopeMap should return "map" instead of "game"
- CIdentifierStorage::getPossibleIdentifiers should (in one way or another) request.localScope "map" and add to allowedScopes all map dependencies
- Old map versions should somehow still use "game" scope?

Up to you whether to do these changes now or not.

Nordsoft91: I already have too many changes in this branch, I'll merge to proceed with other improvements

PR #1918, opened 2023-04-09
Fix vcmi map loading

### LINKED ISSUE #1917: Server crash on map loading
**Describe the bug**
This map is "canonical" and used for tests for a while. It's prepared by legacy map editor, converted from OH3 ME RMG map.Unpack it to maps folder. It'll have "untitled" name.
[SomeMap.vmap.zip](https://github.com/vcmi/vcmi/files/11184568/SomeMap.vmap.zip)

**Game logs**
[VCMI_Server_log.txt](https://github.com/vcmi/vcmi/files/11184566/VCMI_Server_log.txt)

**To Reproduce**
1. Install map attached
2. Start game on this map

**Expected behavior**
Map loaded

**Actual behavior**
Server crash

**Did it work earlier?**
Perfectly works in 1.1

**Version**
 - Version: 1.2, 1.3

IvanSavenko:
Fixes #1917 

... however we need to decide how to handle map compatibility in long-term. 

Perhaps make sure that 1.X editor can always load 1.X-1 maps & ask map authors to update maps by re-saving them after major release?

### DIFF
--- a/config/objects/rewardableBonusing.json
+++ b/config/objects/rewardableBonusing.json
@@ -18,7 +18,8 @@
 					"value"		: 100,
 					"rarity"	: 100
 				},
-				
+				"compatibilityIdentifiers" : [ "object" ],
+
 				"blockedVisitable" : true,
 				"onVisitedMessage" : 22,
 				"visitMode" : "bonus",
@@ -49,7 +50,8 @@
 					"value"		: 100,
 					"rarity"	: 100
 				},
-				
+				"compatibilityIdentifiers" : [ "object" ],
+
 				"onVisitedMessage" : 30,
 				"visitMode" : "bonus",
 				"selectMode" : "selectFirst",
@@ -81,7 +83,8 @@
 					"value"		: 100,
 					"rarity"	: 100
 				},
-				
+				"compatibilityIdentifiers" : [ "object" ],
+
 				"onVisitedMessage" : 50,
 				"visitMode" : "bonus",
 				"selectMode" : "selectFirst",
@@ -112,7 +115,8 @@
 					"value"		: 100,
 					"rarity"	: 100
 				},
-				
+				"compatibilityIdentifiers" : [ "object" ],
+
 				"onVisitedMessage" : 56,
 				"visitMode" : "bonus",
 				"selectMode" : "selectFirst",
@@ -164,7 +168,8 @@
 					"value"		: 100,
 					"rarity"	: 50
 				},
-				
+				"compatibilityIdentifiers" : [ "object" ],
+
 				"onVisitedMessage" : 58,
 				"visitMode" : "bonus",
 				"selectMode" : "selectFirst",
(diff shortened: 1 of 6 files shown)


Nordsoft91:
> Perhaps make sure that 1.X editor can always load 1.X-1 maps & ask map authors to update maps by re-saving them after major release?

IMO, we must keep maps backward compatibility as much as we can.
Saves are not important, but maps are much more.

IvanSavenko:
> IMO, we must keep maps backward compatibility as much as we can. Saves are not important, but maps are much more.

True. I am mostly about "what we should guarantee". Meaning:
- vcmi 1.3 *must* load vcmi 1.2 maps
- vcmi 1.2 *must* load vcmi 1.1 maps
- vcmi 1.2 *may* load vcmi 1.0 maps

Nordsoft91:
And keep in mind that some of map designers still keep using old map editor

IvanSavenko:
> And keep in mind that some of map designers still keep using old map editor

- Why? Some problem with newer one or simply have yet to update?
- How old are we talking about? 1.1 is expected, since 1.2 is not yet released. Or something older than that?

Nordsoft91:
Not sure I agree with that.
At some point we may decide do not support some old formats yes it will be something like
 - vcmi 1.x must load 1.(x-1) maps
 - vcmi1.(x+1) doesn't support 1.(x-1) maps
But let's have clear understanding why do we want to break compatibility with older maps (e.g. support cost is too high) and  not consider that as something usual we do every 3-6 month release (like for saves).
Instead we may have some internal adopters, which may transform maps internally into new format.

Nordsoft91:
> Why? Some problem with newer one or simply have yet to update?
Mostly a habit, but some features are missing (like flixibiliyt in hero's settings). Some maps are created with old ME then revised with new ME (for new terrains for example)
> How old are we talking about? 1.1 is expected, since 1.2 is not yet released. Or something older than that?
Yes, basically map I attached is supported from far away times (0.99)

IvanSavenko:
Understood.
Well, in that case our Json map loader will need to have proper versioning instead of optimistic approach we have right now.

krs0:
But why not having automatic converter functions for every map editor release.
This means if you are on 1.3 and upgrade from 1.0 it will be converted in 3 steps from 1.0 to 1.1, from 1.1 to 1.2 and 1.2 to 1.2.

Nordsoft91:
@krs0 , here is use-case:
1. I have vcmi-1.1 and play my favorite map.
2. I don't play for a year or two.
3. I update my vcmi after that to vcmi-5.26
4. Realize that I can't play my favorite map anymore, disappoint, rollback to vcmi-1.1 and never trust in updates anymore

How automatic converter helps to solve that?

krs0:
Your were probably prompted if you want to upgrade, maybe a backup was also made.
Map was automatically updated to 5.15 as it was supposed to.
What exactly is the problem?

PR #1928, opened 2023-04-10
Spell cast reward

Nordsoft91:
Introduces "casts" reward for configurable objects, which executes adv map spell.
It makes possible to implement town gate from hota
[hotaTownGate.json.zip](https://github.com/vcmi/vcmi/files/11186988/hotaTownGate.json.zip)

https://user-images.githubusercontent.com/9308612/230810452-bf73bcbe-4f2b-4f9c-be3e-ad2f87b70056.mp4

### DIFF
--- a/lib/mapObjects/CRewardableConstructor.cpp
+++ b/lib/mapObjects/CRewardableConstructor.cpp
@@ -132,6 +132,7 @@ void CRandomRewardObjectInfo::configureReward(CRewardableObject * object, CRando
 	reward.artifacts = JsonRandom::loadArtifacts(source["artifacts"], rng);
 	reward.spells = JsonRandom::loadSpells(source["spells"], rng, spells);
 	reward.creatures = JsonRandom::loadCreatures(source["creatures"], rng);
+	reward.casts = JsonRandom::loadSpells(source["casts"], rng, spells);
 
 	for ( auto node : source["changeCreatures"].Struct() )
 	{
--- a/lib/mapObjects/CRewardableConstructor.cpp
+++ b/lib/mapObjects/CRewardableConstructor.cpp
@@ -132,7 +132,11 @@ void CRandomRewardObjectInfo::configureReward(CRewardableObject * object, CRando
 	reward.artifacts = JsonRandom::loadArtifacts(source["artifacts"], rng);
 	reward.spells = JsonRandom::loadSpells(source["spells"], rng, spells);
 	reward.creatures = JsonRandom::loadCreatures(source["creatures"], rng);
-	reward.casts = JsonRandom::loadSpells(source["casts"], rng, spells);
+	if(!source["spellCast"].isNull() && source["spellCast"].isStruct())
+	{
+		reward.spellCast.first = JsonRandom::loadSpell(source["spellCast"]["spell"], rng);
+		reward.spellCast.second = source["spellCast"]["schoolLevel"].Integer();
+	}
 
 	for ( auto node : source["changeCreatures"].Struct() )
 	{
--- a/lib/mapObjects/CRewardableObject.cpp
+++ b/lib/mapObjects/CRewardableObject.cpp
@@ -363,13 +363,14 @@ void CRewardableObject::grantRewardAfterLevelup(const CRewardVisitInfo & info, c
 		cb->giveCreatures(this, hero, creatures, false);
 	}
 	
-	if(!info.reward.casts.empty())
+	if(info.reward.spellCast.first != SpellID::NONE)
 	{
-		caster = std::make_unique<spells::OuterCaster>(hero, SecSkillLevel::EXPERT);
-		for(const auto & c : info.reward.casts)
-		{
-			cb->castSpell(caster.get(), c, int3{-1, -1, -1});
-		}
+		caster.setActualCaster(hero);
+		caster.setSpellSchoolLevel(info.reward.spellCast.second);
+		cb->castSpell(&caster, info.reward.spellCast.first, int3{-1, -1, -1});
+		
+		if(info.reward.removeObject)
+			logMod->warn("Removal of object with spell casts is not supported!");
 	}
 	else if(info.reward.removeObject) //FIXME: object can't track spell cancel or finish, so removeObject leads to crash
 		cb->removeObject(this);
--- a/lib/mapObjects/CRewardableObject.cpp
+++ b/lib/mapObjects/CRewardableObject.cpp
@@ -17,6 +17,8 @@
 #include "../IGameCallback.h"
 #include "../CGameState.h"
 #include "../CPlayerState.h"
+#include "../spells/CSpellHandler.h"
+#include "../spells/ISpellMechanics.h"
 
 #include "CObjectClassesHandler.h"
 
@@ -358,6 +360,14 @@ void CRewardableObject::grantRewardAfterLevelup(const CRewardVisitInfo & info, c
 			creatures.addToSlot(creatures.getFreeSlot(), new CStackInstance(crea.type, crea.count));
 
 		cb->giveCreatures(this, hero, creatures, false);
+	}
+	
+	if(!info.reward.casts.empty())
+	{
+		for(const auto & c : info.reward.casts)
+		{
+			cb->castSpell(hero, c, int3{-1, -1, -1});
+		}
 	}
 
 	if(info.reward.removeObject)
--- a/lib/mapObjects/CRewardableObject.cpp
+++ b/lib/mapObjects/CRewardableObject.cpp
@@ -365,9 +365,10 @@ void CRewardableObject::grantRewardAfterLevelup(const CRewardVisitInfo & info, c
 	
 	if(!info.reward.casts.empty())
 	{
+		caster = std::make_unique<spells::OuterCaster>(hero, 3);
 		for(const auto & c : info.reward.casts)
 		{
-			cb->castSpell(hero, c, int3{-1, -1, -1});
+			cb->castSpell(caster.get(), c, int3{-1, -1, -1});
 		}
 	}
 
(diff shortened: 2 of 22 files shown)


rilian-la-te:
@Nordsoft91 please, check obstacles, moats and creature caster. Also, check bind.

rilian-la-te:
@Nordsoft91 is this possible to implement without a getHeroCaster method?

lib/mapObjects/CRewardableConstructor.cpp
```diff
@@ -132,6 +132,7 @@ void CRandomRewardObjectInfo::configureReward(CRewardableObject * object, CRando
 	reward.artifacts = JsonRandom::loadArtifacts(source["artifacts"], rng);
 	reward.spells = JsonRandom::loadSpells(source["spells"], rng, spells);
 	reward.creatures = JsonRandom::loadCreatures(source["creatures"], rng);
+	reward.casts = JsonRandom::loadSpells(source["casts"], rng, spells);
```
IvanSavenko, line 135: Perhaps load only single spell (I doubt that TP + DD in the same object will work properly anyway)
And also load mastery level (none/basic/advanced/expert)

IvanSavenko, line 135: Also, perhaps add warning, e.g.
```
	if(!info.reward.casts.empty() && info.reward.removeObject)
        logMod->warn("Removal of object with spell casts is not supported!");
```

Nordsoft91:
@rilian-la-te , the problem is that caster and hero are different in this case. So there should be a way to get information about hero from caster. It's possible to reduce type to Caster* or CGObjectInstance* but I don't see help in that because dynamic_cast will be used anyway inside all spells. BTW, what is bind?

lib/spells/AdventureSpellMechanics.cpp
```diff
@@ -397,15 +417,15 @@ ESpellCastResult TownPortalMechanics::applyAdventureEffects(SpellCastEnvironment
...
+		const auto relations = env->getCb()->getPlayerRelations(destination->tempOwner, parameters.caster->getCasterOwner());
 
 		if(relations == PlayerRelations::ENEMIES)
 		{
 			env->complain("Can't teleport to enemy!");
 			return ESpellCastResult::ERROR;
 		}
 
-		if(static_cast<int>(parameters.caster->movement) < moveCost)
+		if(static_cast<int>(parameters.caster->getHeroCaster()->movement) < moveCost)
```
IvanSavenko, line 428: - Will "Town Gate" work same as in Hota if we're at 0 MP when entering building?
- Does "Town Gate" takes MP to begin with?

lib/mapObjects/CRewardableObject.cpp
```diff
@@ -359,8 +362,16 @@ void CRewardableObject::grantRewardAfterLevelup(const CRewardVisitInfo & info, c
...
 	}
-
-	if(info.reward.removeObject)
+	
+	if(!info.reward.casts.empty())
+	{
+		caster = std::make_unique<spells::OuterCaster>(hero, SecSkillLevel::EXPERT);
+		for(const auto & c : info.reward.casts)
+		{
+			cb->castSpell(caster.get(), c, int3{-1, -1, -1});
```
IvanSavenko, line 371: Perhaps we should add some handling is spellcasting is impossible?
- No owned towns for TP
- All owned towns are blocked by visitor for TP
- No water tiles nearby for Summon Boat
- Movement check for both DD and TP

Will this display message understandable to player?
If not - we might want to have some handling for this scenario.

rilian-la-te:
@Nordsoft91 bind is a Dendroid effect. 

I think we should not change getCasterUnitId to return hero ID, because it will break Bind and First Aid Tent (fortunately, tent only visually). 

I think than it is better to add a methods "getHero" into AdventureSpellMechanics, which will return a hero or null if we surely need a hero. And it will internally work as dynamic_cast.

Nordsoft91:
Obstacles/remove obstacles, moats, creatures (archangels, magogs, fairy dragons, fire-shield ifrits) were tested

rilian-la-te:
@Nordsoft91 dendroids?

lib/spells/AdventureSpellMechanics.cpp
Nordsoft91: 0 points are not allowed (bug?). It doesn't take any movement  points.

lib/mapObjects/CRewardableObject.cpp
Nordsoft91: But all those checks are inside spell mechanics. I don't want to introduce any spell-specific logic here

lib/spells/AdventureSpellMechanics.cpp
dydzio0614: needs fix, this object is needed for pvp

lib/mapObjects/CRewardableObject.cpp
IvanSavenko: Maybe. But visiting "Town Gate" object only to see message like "Failed to cast a spell" is weird and not what player expects.
Even basic
```
if (cb->canCast())
     cb->castSpell()
else
    showMessage("You approach Town Gate, which were supposed to bring you back to your town, but for some reason gates stay closed. ")  // <- message configured in reward definition
```
would be better

Frankly I would check HotA first, at the very least they already faced these issues (how to handle non-functional gates) and might already have some handling for it.

Nordsoft91:
> I think than it is better to add a methods "getHero" into AdventureSpellMechanics, which will return a hero or null if we surely need a hero. And it will internally work as dynamic_cast.

it won't work (or I don't get an idea) because AdventureSpellMechanics doesn't know if hero is a caster or hero is agrregated field

Nordsoft91:
> dendroids? Catapult? First Aid Tent?

All those are ok

Nordsoft91:
> I think we should not change getCasterUnitId to return hero ID, because it will break Bind and First Aid Tent (fortunately, tent only visually).

Works well, I suppose it's because all units have their own implementation of getCasterUnitId. Change is related only to hero caster

lib/mapObjects/CRewardableObject.cpp
Nordsoft91: > Frankly I would check HotA first, at the very least they already faced these issues (how to handle non-functional gates) and might already have some handling for it.

It's not applicable for hota, because in their case town gate works always. 
In our case we may have failures like in example with boat summoning - hota just doesn't have such object.

All this can-spell-checking logic are deeply inside spell mechanics, so I don't want to deal with that (I was very surprised that it's not possible event to cast spell from lib code - only from server)

As for messages - they are quite meaningful for spells. You won't see "Failed to cast spell", most probably "Hero is tired - try again tomorrow"

I don't want to make silver-bullet changes covering all possible use-cases. Yes - for now it works with limitations, but it solves well very specific problem of supporting town gate from hota mod.

rilian-la-te:
@Nordsoft91 Did you tried to fight dendroid and then kill them when there are binded stack exists?

Do casting Cure spell by hero is not produce error messages inside battle log?

lib/mapObjects/CRewardableObject.cpp
IvanSavenko: Well, this is hota with 0 towns.
So our behavior shoud be as good as hota then 😅 
зображення

rilian-la-te:
@Nordsoft91 You will be okay, if this feature will be redesigned later (for example, when town buildings will get a rework)?

lib/mapObjects/CRewardableObject.cpp
dydzio0614: are we able to easily fix taking movement points though?

Nordsoft91:
> Did you tried to fight dendroid and then kill them when there are binded stack exists?

Binding disappears as create gets turn

> Do casting Cure spell by hero is not produce error messages inside battle log?

Green complaints from server? No

> You will be okay, if this feature will be redesigned later (for example, when town buildings will get a rework)?

Totally ok if this won't be done by me :)

rilian-la-te:
Not green complains, but extra message in battle log like from first aid tent.

lib/mapObjects/CRewardableConstructor.cpp
Nordsoft91: agree, also thinking on.

lib/spells/AdventureSpellMechanics.cpp
Nordsoft91: Fixed

lib/spells/OuterCaster.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include "ProxyCaster.h"
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+namespace spells
+{
+
+class DLL_LINKAGE OuterCaster : public ProxyCaster
```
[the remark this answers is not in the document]
Nordsoft91: I wanted to make ProxyCaster, but it's occupied. Looking for better name, maybe ExternalCaster?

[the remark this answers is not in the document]
IvanSavenko: ExternalCaster sounds better IMO.

Comment to describe how this class is different from ProxyCaster would also help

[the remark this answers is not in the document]
rilian-la-te: @Nordsoft91 why ProxyCaster or SilentCaster is not suitable?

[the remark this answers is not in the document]
Nordsoft91: ProxyCaster has own implementations of all methods.
For SilentCaster I would need to override method which can break something (like obstacles).
Better to have dedicated class for adventure map and not mix with battles

PR #1936, opened 2023-04-10
Upgrade c++ standard to c++17

IvanSavenko:
As discussed on Slack, bumping c++ standard to c++17 since it *should* be supported by all platforms with exception of std::filesystem on MacOS

Test PR to make sure that CI works with c++17 and to make sure that all platforms can use c++17.

If everything goes well will probably replace some of our vstd stuff with standard versions.

Boost deprecation is out of scope and not a priority for me.


kambala-decapitator (approved):

rilian-la-te:
@IvanSavenko let it build, then maybe merge #1916 and then get this included maybe?

IvanSavenko:
`include/tbb/../oneapi/tbb/detail/_task.h:216:13: error: aligned deallocation function of type 'void (void *, std::align_val_t) noexcept' is only available on macOS 10.14 or newer
`

Well, that's unfortunate.... Don't see any clear way to skip/disable that.
Perhaps I can replace tbb with native c++, nullkiller was OK with removal of tbb

kambala-decapitator:
> Perhaps I can replace tbb with native c++, nullkiller was OK with removal of tbb

the less dependencies the better

IvanSavenko:
Intel TBB will have to stay for now.
While I can implement parallel_for via pure std this is not the only issue:
1) We also use some thread safe containers from TBB, they don't have any std or boost replacements.
2) Library support... It seems that libc++ (likely what we use on Mac) does not have support for parallel algorithms and libstdc++ (linux/mingw) depends on TBB internally - meaning, we still need TBB as dependency.

However found a way to workaround alignment requirement that caused CI to fail, so TBB no longer blocks c++17

rilian-la-te:
@IvanSavenko libstdc++ is not depend on tbb, AFAIK. But your point for older Mac is valid.

IvanSavenko:
> libstdc++ is not depend on tbb, AFAIK

Nope. It is. Not "base" version, but if you include <execution> and try to use parallel algorithms libstdc++ will use tbb (and require linking to it)

rilian-la-te:
@IvanSavenko it seems than libc++ is also depend on tbb internally to implement C++17 standard parallel features.

IvanSavenko:
Seems to be working for me. 
Will recheck Windows build later, would be great if somebody could test MacOS, specifically - let NKAI play for a couple of days, to make sure that there are no issues due to disabled aligned allocations.

rilian-la-te:
https://reviews.llvm.org/D55889 - commit) But need GitHub login to see.

kambala-decapitator:
tested on macOS 10.13 (the lowest we support) by skipping turns for a week: seems good.

only once had client freeze after AI attacked a garrison (I vaguely remember something like that was already reported). But after I loaded last save game didn't freeze in the same point. Not sure if it matters, but the freeze happened with the map cheat and after loading the game I forgot to enter it immediately (i.e. I didn't see how AI attacked the same garrison). Entered the cheat on next turn, the garrison was already under AI control, and when AI attacked another garrison, game didn't freeze.

Logs from freeze: [Archive.zip](https://github.com/vcmi/vcmi/files/11211042/Archive.zip)

PR #1938, opened 2023-04-10
config: Orb of Vulnerability should block all resistances from work (Resistance skill, dwarves, unicorns and so on.

rilian-la-te:
Just set MAGIC_RESISTANCE and SPELL_RESISTANCE_AURA to 0 by INDEPENDENT_MIN, and then this bonuses will always return 0.

According to FizMig, it is correct behaviour of this artifact.


Nordsoft91:
You can use this map for testing
[3042_ZZZ_orbvulnerability.vmap.zip](https://github.com/vcmi/vcmi/files/11192142/3042_ZZZ_orbvulnerability.vmap.zip)

rilian-la-te:
It seems I forgot a propagator... Will add after come home.

rilian-la-te:
Propagator added, now tested and resistance is negated on all creatures.

PR #1939, opened 2023-04-10
Do not place object visible tiles over the top of the map.

DjWarmonger:
Top visible tile of the object won't be placed over the top edge of teh map (y < 0).

### DIFF
--- a/lib/mapObjects/CObjectHandler.h
+++ b/lib/mapObjects/CObjectHandler.h
@@ -159,6 +159,7 @@ public:
 	bool visitableAt(int x, int y) const; //returns true if object is visitable at location (x, y) (h3m pos)
 	int3 visitablePos() const override;
 	int3 getPosition() const override;
+	int3 getTopVisiblePos() const;
 	bool blockingAt(int x, int y) const; //returns true if object is blocking location (x, y) (h3m pos)
 	bool coveringAt(int x, int y) const; //returns true if object covers with picture location (x, y) (h3m pos)
 	std::set<int3> getBlockedPos() const; //returns set of positions blocked by this object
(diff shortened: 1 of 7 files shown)


lib/mapObjects/CObjectHandler.h
```diff
@@ -159,6 +159,7 @@ class DLL_LINKAGE CGObjectInstance : public IObjectInterface
 	bool visitableAt(int x, int y) const; //returns true if object is visitable at location (x, y) (h3m pos)
 	int3 visitablePos() const override;
 	int3 getPosition() const override;
+	int3 getTopVisiblePos() const;
```
krs0, line 162: I had a hard time with `getTopVisiblePos()` and even harder to come with something more intelligible -> `getPositionForVisibleTop()` or `getPositionWhenTopIsVisible()`

DjWarmonger: This gets the actual coordinates of the top visible tile of an object placed at its current position, and will always return correct result even if the object is completely outside the map.

lib/rmg/ObjectManager.cpp
```diff
@@ -112,6 +112,9 @@ int3 ObjectManager::findPlaceForObject(const rmg::Area & searchArea, rmg::Object
 				continue;
 			
 			obj.setPosition(tile);
+
+			if (obj.getVisibleTop().y < 0)
+				continue;
```
Nordsoft91, line 117: Consider moving this condition inside weight functions. Thus you can control which objects should satisfy this condition and which shouldn't

Nordsoft91 (approved):

lib/rmg/ObjectManager.cpp
DjWarmonger: All of the objects should fit on a screen.

PR #1941, opened 2023-04-11
vcmi: reduce boost a little bit

rilian-la-te:
In topic.

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -217,8 +217,8 @@ typedef int16_t si16; //signed int 16 bits (2 bytes)
 typedef int8_t si8; //signed int 8 bits (1 byte)
 
 // Lock typedefs
-typedef boost::lock_guard<boost::mutex> TLockGuard;
-typedef boost::lock_guard<boost::recursive_mutex> TLockGuardRec;
+using TLockGuard = std::lock_guard<std::mutex>;
+using TLockGuardRec = std::lock_guard<std::recursive_mutex>;
 
 /* ---------------------------------------------------------------------------- */
 /* Macros */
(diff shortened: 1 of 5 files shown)


Global.h
```diff
@@ -693,7 +693,7 @@ namespace vstd
 	template<typename T>
 	void removeDuplicates(std::vector<T> &vec)
 	{
-		boost::sort(vec);
+		std::sort(vec.begin(), vec.end());
```
IvanSavenko, line 696: I think we should keep boost::ranges
1) they are more convenient to use than begin/end
2) they are already coming in std (c++20 I believe?)

IvanSavenko (commented):
Should be playtested for a bit, in case if there are subtle differences between locks behavior, but othervice OK

Global.h
rilian-la-te: Boost::ranges is preserved, I do not touch it, I just replaced it only here. I think we should not use boost in critical functions if std can satisfy. What do you think?

DjWarmonger: I'm with @IvanSavenko . Shorter is better.

IvanSavenko: Feel free to keep it here, since both versions are quite readable. 

But don't see any need to replace it, critical functions or not. We already include boost globally, so effect on compile time is minimal. And ranges are way more readable than begin/end. So readability >> performance.

rilian-la-te: @IvanSavenko I think when transition to C++17 will be done, boost usage should be reduced in a big way and removed from global.h completely.

rilian-la-te: And remained only in specific files when boost is required.

IvanSavenko: Maybe. But let's first remove std equivalents first (thread/date_time/etc) and then think of remaining parts. In this order.

And even then - I would rather focus on things like interprocess, maybe program_options. Things like ranges/algorithms should be kept in use and in global.h. Don't see any point in replacing them.

dydzio0614: there are some thread features used that were introduced in jthread in c++20 or so, like AI thread interruption

rilian-la-te: @dydzio0614 they do not use global.h threads AFAIK.

IvanSavenko: > Why removing std equivalents? Maybe boost parts which have std equivalents?

Yeah, those.

> Removing ranges and algorithm from global will encourage people do not use it)

Don't see any reason to discourage their usage

PR #1944, opened 2023-04-11
Nullkiller -> Standard-AI

Laserlicht:
Followup for android: #1768

Otherwise with two different default values the Nullkiller AI always get overwritten.

Fix is tested and works.


dydzio0614:
naughty android launcher

PR #1946, opened 2023-04-12
Monoliths and Subterranean Gates placement tweaks

DjWarmonger:
Little tweaks:

- Make sure Mines are placed after Monoliths and all the connections.
- Use product instead of distance sum for Subterranean Gate placement. Yields a little bit more natural results.


Nordsoft91:
> Make sure Mines are placed after Monoliths and all the connections.

What does it?

DjWarmonger:
Well, Monoliths are placed at the edges / corners, which generates longer roads in general.

PR #1968, opened 2023-04-12
Fix caster tactics

### LINKED ISSUE #1958: HotA grail not working
**Describe the bug**
HotA grail was not working, should be now fixed by

https://github.com/vcmi/vcmi/pull/1882

Mantis: https://bugs.vcmi.eu/view.php?id=2804

### LINKED ISSUE #1960: UI is not active during siege with tactics
**Describe the bug**
Not sure, what exactly is root cause, but assuming that battle
[A Capitol Test.h3m.zip](https://github.com/vcmi/vcmi/files/11214641/A.Capitol.Test.h3m.zip)

**To Reproduce**
1. Install build from https://github.com/vcmi/vcmi/actions/runs/4569392274
2. Start map attached (called "A capitol test")
3. Move hero to the right and attack rampart castle

**Expected behavior**
Tactics buttons and actions are available

**Actual behavior**
All actions except auto-battle are not available

**Did it work earlier?**
yes, before https://github.com/vcmi/vcmi/pull/1706

**Screenshots**
(https://h3vcmi.slack.com/files/U049Y8UPH4Z/F052H8HQHLP/untitled_video_-_made_with_clipchamp.mp4)

**Version**
Bisect:
e1c3550 - bad
4d6b88f - good
4d6b88f + PR1706

**Additional context**
Theoretically can be broken in one of those PRs, but it's barely possible:
#1854, #1852, #1848, #1853

rilian-la-te:
Fixes #1960 
Fixes #1958

### DIFF
--- a/client/battle/BattleInterface.cpp
+++ b/client/battle/BattleInterface.cpp
@@ -336,11 +336,16 @@ void BattleInterface::battleFinished(const BattleResult& br, QueryID queryID)
 
 void BattleInterface::spellCast(const BattleSpellCast * sc)
 {
-	windowObject->blockUI(true);
+	// Do not deactivate anything in tactics mode
+	// This is battlefield setup spells
+	if(!tacticsMode)
+	{
+		windowObject->blockUI(true);
 
-	// Disable current active stack duing the cast
-	// Store the current activeStack to stackToActivate
-	stacksController->deactivateStack();
+		// Disable current active stack duing the cast
+		// Store the current activeStack to stackToActivate
+		stacksController->deactivateStack();
+	}
 
 	CCS->curh->set(Cursor::Combat::BLOCKED);
 
--- a/lib/JsonNode.cpp
+++ b/lib/JsonNode.cpp
@@ -749,7 +749,7 @@ std::shared_ptr<ILimiter> JsonUtils::parseLimiter(const JsonNode & limiter)
 				else
 					return std::make_shared<CreatureAlignmentLimiter>(static_cast<EAlignment>(alignment));
 			}
-			else if(limiterType == "FACTION_LIMITER")
+			else if(limiterType == "FACTION_LIMITER" || limiterType == "CREATURE_FACTION_LIMITER") //Second name is deprecated, 1.2 compat
 			{
 				std::shared_ptr<FactionLimiter> factionLimiter = std::make_shared<FactionLimiter>();
 				VLC->modh->identifiers.requestIdentifier("faction", parameters[0], [=](si32 faction)


client/battle/BattleInterface.cpp
```diff
@@ -336,11 +336,16 @@ void BattleInterface::battleFinished(const BattleResult& br, QueryID queryID)
 
 void BattleInterface::spellCast(const BattleSpellCast * sc)
 {
-	windowObject->blockUI(true);
+	// Do not deactivate anything in tactics mode
+	// This is battlefield setup spells
+	if(!tacticsMode)
```
IvanSavenko, line 341: Why is this needed? UI should be reactivated once spellcast ends

rilian-la-te: @IvanSavenko this is not the case if there is no spellcaster hero and no spellcaster unit.

rilian-la-te: In such case no EndAction or any other spell action called, so, interface is just locked and not unlocked next (because in tactics mode there is no speed system to switch to next creature after action)/

IvanSavenko: Why?
Server should still send notification that spell ended.

Does moat casting bypasses BattleSpellMechanics::cast?

rilian-la-te: No, it uses BattleSpellMechanics::cast precisely. But it does not use BattleAction, like opening battle spells. But because opening battle spells runs after tactics, this are not suspectible to this bug.

rilian-la-te: ```BattleSpellMechanics::cast``` use ```BattleSpellCast``` package, so, we react in client to it inside ```BattleInterface::spellCast``, in which we try to block interface twice (in blockUI and deactivateStack), and will not unlock it in any way inside this function. Unlocking usually happens in another function, which will not run in tactics mode.

IvanSavenko: > we should can blow up Tower mines by usage of tactics

I'd call that a bug...

rilian-la-te: @IvanSavenko this is controversal. If we move a creature in tactics mode, and its pathfinder will create a path through minefield, then mines should be blown up (according to FizMig). I do not see it as a bug, because a setup of moats and mines should be done before battle, and not inside it.

krs0: I say VCMI should not implement every H3 bug.
If a bug seems to be crucial and needs to be implemented, should be discussed before.

krs0: Tactics mode supports instant teleportig of creatures. Which invalidates this implementation.

IvanSavenko: Interesting, if there were a bug that allows catapult to attack walls during tactics that would be controversial as well?

After all, setup of walls and towers should be done before battle, and no during it?

rilian-la-te: But setup of tower shooters is done before battle, why moat setup should be done after?

IvanSavenko: I would rather say that 
a) pathfinder should avoid moat hexes in tactics
b) moat effects should not trigger during tactics

Defender needs to see tower mines if he has tactics

rilian-la-te: If defender need to see tower mines - then it should be cast before tactics.

PR #1991, opened 2023-04-14
vcmi: rule-based modernization part1

rilian-la-te:
Rule set:

modernize-use-auto
modernize-use-using
modernize-use-emplace
modernize-loop-convert
misc-redundant-expression
performance-for-range-copy
performance-inefficient-vector-operation
performance-inefficient-algorithm
readability-qualified-auto
readability-redundant-*

@IvanSavenko is this modenization okay?

### DIFF
--- a/lib/LogicalExpression.h
+++ b/lib/LogicalExpression.h
@@ -30,19 +30,14 @@ namespace LogicalExpressionDetail
 		};
 		template<EOperations tag> class Element;
 
-		typedef Element<ANY_OF> OperatorAny;
-		typedef Element<ALL_OF> OperatorAll;
-		typedef Element<NONE_OF> OperatorNone;
+		using OperatorAny = Element<ANY_OF>;
+		using OperatorAll = Element<ALL_OF>;
+		using OperatorNone = Element<NONE_OF>;
 
-		typedef ContainedClass Value;
+		using Value = ContainedClass;
 
 		/// Variant that contains all possible elements from logical expression
-		typedef std::variant<
-			OperatorAll,
-			OperatorAny,
-			OperatorNone,
-			Value
-			> Variant;
+		using Variant = std::variant<OperatorAll, OperatorAny, OperatorNone, Value>;
 
 		/// Variant element, contains list of expressions to which operation "tag" should be applied
 		template<EOperations tag>
@@ -73,7 +68,7 @@ namespace LogicalExpressionDetail
 	template<typename ContainedClass>
 	class TestVisitor
 	{
-		typedef ExpressionBase<ContainedClass> Base;
+		using Base = ExpressionBase<ContainedClass>;
 
 		std::function<bool(const typename Base::Value &)> classTest;
 
@@ -119,7 +114,7 @@ namespace LogicalExpressionDetail
 	template<typename ContainedClass>
 	class PossibilityVisitor
 	{
-		typedef ExpressionBase<ContainedClass> Base;
+		using Base = ExpressionBase<ContainedClass>;
 
 	protected:
 		std::function<bool(const typename Base::Value &)> satisfiabilityTest;
@@ -167,7 +162,7 @@ namespace LogicalExpressionDetail
 	template <typename ContainedClass>
 	class SatisfiabilityVisitor : public PossibilityVisitor<ContainedClass>
 	{
-		typedef ExpressionBase<ContainedClass> Base;
+		using Base = ExpressionBase<ContainedClass>;
 
 	public:
 		SatisfiabilityVisitor(std::function<bool (const typename Base::Value &)> satisfiabilityTest,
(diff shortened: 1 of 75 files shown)


IvanSavenko:
If you can split it into 1 commit per rule that would help a lot with review.
Right now I can't find examples for like half of your rules because majority of changes come from modernize-use-auto

Or just split modernize-use-auto into separate commit/PR. Nice change but hard to review anything due to it

rilian-la-te:
@IvanSavenko this changes is mostly from  readability-qualified-auto, not from modernize-use-auto. But I understood you. Will try to split (but most likely headers will be first, and then *.cpp files by rule).

rilian-la-te:
@IvanSavenko please, take a look into a headers first. Cpp files I think better to post via another PR.

client/adventureMap/CAdvMapInt.h
```diff
@@ -110,7 +110,6 @@ class CAdvMapInt : public CIntObject
 
 	std::shared_ptr<MapAudioPlayer> mapAudio;
 
-private:
```
IvanSavenko, line 113: IMO old code is better - it was separating two semantically different sections - private members and private methods

lib/LogicalExpression.h
```diff
@@ -510,37 +505,34 @@ namespace LogicalExpressionDetail
...
-	/// Base constructor
-	LogicalExpression()
-	{}
-
-	/// Constructor from variant or (implicitly) from Operator* types
-	LogicalExpression(const Variant & data):
-		data(data)
-	{
-	}
+		using Base = LogicalExpressionDetail::ExpressionBase<ContainedClass>;
```
IvanSavenko, line 508: Alignment broke?

lib/rmg/ConnectionsPlacer.h
```diff
@@ -30,8 +30,7 @@ class ConnectionsPlacer: public Modificator
 	
 protected:
 	void collectNeighbourZones();
-	
-protected:
+
```
IvanSavenko, line 33: I'd rather disable this rule.

It might be useful if clang-tidy is integrated in IDE & can show potential issues, but see no reason to enforce it since we sometimes use this to separate logical blocks

lib/serializer/CSerializer.h
```diff
@@ -149,42 +149,28 @@ struct is_serializeable
...
-		//else if
-		boost::mpl::eval_if<std::is_same<CGHeroInstance,U>,
-		boost::mpl::identity<HeroTypeID>,
-		//else if
-		boost::mpl::eval_if<std::is_base_of<CGObjectInstance,U>,
-		boost::mpl::identity<ObjectInstanceID>,
-		//else
-		boost::mpl::identity<si32>
-		> > > > > >::type type;
+		boost::mpl::eval_if<std::is_same<CCreature, U>,
```
IvanSavenko, line 163: boost::mpl killed formatting :(

IvanSavenko (commented):
Judging from this list I'd disable "redundant access specifiers" rule, other than that - looks OK I guess?

lib/serializer/CSerializer.h
rilian-la-te: So, for_each in variant is not last usage) Formatting will be fixed.

client/adventureMap/CAdvMapInt.h
rilian-la-te: restored

lib/rmg/ConnectionsPlacer.h
rilian-la-te: restored, but my IDE is showing yellow marks)

lib/serializer/CSerializer.h
rilian-la-te: restored

lib/LogicalExpression.h
rilian-la-te: returned

rilian-la-te:
@IvanSavenko rebased to latest develop.

lib/LogicalExpression.h
```diff
@@ -510,37 +505,34 @@ namespace LogicalExpressionDetail
...
-	/// Base constructor
-	LogicalExpression()
-	{}
-
-	/// Constructor from variant or (implicitly) from Operator* types
-	LogicalExpression(const Variant & data):
-		data(data)
-	{
-	}
+	using Base = LogicalExpressionDetail::ExpressionBase<ContainedClass>;
```
IvanSavenko, line 508: indentation still dead

rilian-la-te: @IvanSavenko where? I do not see any indentation errors AFAIK. Or you mean Variant (I need to split it into multiple lines)?

IvanSavenko: Lines 510-535 here?

rilian-la-te: Move it 1 level backwards?

IvanSavenko: Yes. Class is defined at top-level, at 0 indentation level. All members should be placed at 1 tab, not 2

rilian-la-te: done

IvanSavenko (approved):
Looks OK, will merge tomorrow if there are no other comments.

rilian-la-te:
@IvanSavenko can you merge this after CI passed?

Issue #1993, opened 2023-04-15
HD executable icon for windows


tytannial:
**Describe your proposal**
Draw a HD executable icon for windows :D

**Additional context**
![image](https://user-images.githubusercontent.com/11215623/232187628-114f94b8-542a-483f-b9f7-1d6d4534c7fc.png)

icon not feet the modern system hd icon

IvanSavenko:
We actually have one, but no idea how to attach it to executable on Windows:
https://github.com/vcmi/vcmi/blob/develop/client/icons/vcmiclient.512x512.png

tytannial:
OK...i will have try to make one 😊

Laserlicht:
I've created one (see pull request):

![grafik](https://user-images.githubusercontent.com/13953785/232211591-f9f318a7-8727-4c8b-9d7b-bd2d26542065.png)

krs0:
Am I the only one who dislikes the VCMI logo?

here the VCMI logo
![image](https://user-images.githubusercontent.com/15194321/232307432-11c7d1d8-6fdf-4abc-95e1-df49cffcaed3.png)

Here various H3 shield logos...
![image](https://user-images.githubusercontent.com/15194321/232307376-a3c4a50c-6277-442a-b0a0-ee4ee64b494f.png)

Here an automaker logo
![image](https://user-images.githubusercontent.com/15194321/232307407-ab291020-0e4f-4d9c-9043-306291164290.png)

IvanSavenko:
Can't say that I am huge fan of our current icon, but I do associate it with VCMI, so not sure if we should significantly change it.

If we have someone interested in redesign then go ahead I guess, but IMO icon should stay similar to how it is now - yellow pixelated "VCMI" text on brown shield

Alexander-Wilms:
I've experimented with [craiyon.ai](https://www.craiyon.com/). If one tries a few prompts and gets a nice result, it could then be recreated manually as a SVG.

![image](https://github.com/vcmi/vcmi/assets/3226457/204657eb-0267-47ec-9069-ab06b08dfd3a)

krs0:
To me something like the first one looks great. But I would have VCMI written in the shield in some old font and keep original colors.

DjWarmonger:
I'd like to take this style: 
![image](https://github.com/vcmi/vcmi/assets/2566990/88c4f535-0752-42ef-9884-58b50bda1d92)
Then recolor the shield and add teh font similiar to current VCMI icon.

[...]

PR #1999, opened 2023-04-15
Big/Vector Icons

Laserlicht:
adresses https://github.com/vcmi/vcmi/issues/1993

also contains svg and derives all other logos from it


Laserlicht:
The vcmi.icns for Mac OS is missing. I currently don't now how to create from svg.

IvanSavenko:
Slightly different than what we had before, but I think I actually like the change

Laserlicht:
> Slightly different than what we had before, but I think I actually like the change

Yes - an exact vector rebuild wasn't easy. The biggest change is the removal of the blur for the big icons.

client/icons/generate_icns.py
```diff
@@ -15,4 +15,3 @@
 
 os.system("iconutil -c icns vcmi.iconset")
 shutil.rmtree("vcmi.iconset")
-
```
kambala-decapitator, line 18: please drop this change

kambala-decapitator:
wanted to create icns, but looks like the script is broken (requires py2?)

```
> cd vcmi/osx
> chmod +x ../client/icons/generate_icns.py
> ../client/icons/generate_icns.py ~/Desktop/vcmiclient.1024x1024.png
from: can't read /var/mail/PIL
Version: ImageMagick 7.1.0-61 Q16-HDRI x86_64 20866 https://imagemagick.org
...
import: delegate library support not built-in '' (X11) @ error/import.c/ImportImageCommand/1297.
/Users/kambala/dev/vcmi/vcmi/client/icons/generate_icns.py: line 4: syntax error near unexpected token `('
/Users/kambala/dev/vcmi/vcmi/client/icons/generate_icns.py: line 4: `img = Image.open(sys.argv[1])'
```

Laserlicht:
> wanted to create icns, but looks like the script is broken (requires py2?)

I was able to create this file with the linux package `icnsutils`

and then with `png2icns vcmi.icns vcmiclient.*.png` (images (e.g. vcmiclient.16x16.png) with 16, 32, 48, 128, 256, 512, 1024 in folder)

kambala-decapitator:
thanks!

checked the file: looks good!

IvanSavenko:
By the way, @Laserlicht is it possible to create separate svg's that contain:
1) only shield, without text
2) only styled text, without shield?

Obviously optional, but we had such request some time ago, and since you're at least somewhat familiar with svg perhaps you're willing to do this?

Laserlicht:
@IvanSavenko With SVG it's a simple task.
vcmiclient_shield
vcmiclient_text

IvanSavenko (approved):

Nordsoft91 (approved):

PR #2013, opened 2023-04-16
Split Process console commands code into individual functions

krs0:
Extracted functions for each if branch, added one line doc for each function, rearranged the commands to be grouped more logically.


IvanSavenko (approved):

krs0:
Broke with merge!!! will fix tomorrow.

PR #2015, opened 2023-04-16
vcmi: migrate to C++17

rilian-la-te:
1. use std::any
2. use std::variant
3. use std::optional

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -577,8 +577,8 @@ namespace vstd
 		return i >= 0  &&  i < c.size();
 	}
 
-	template <typename Container, typename Index>
-	boost::optional<typename Container::const_reference> tryAt(const Container &c, Index i)
+	template<typename Container, typename Index>
+	std::optional<typename Container::value_type> tryAt(const Container & c, Index i)
 	{
 		if(isValidIndex(c, i))
 		{
@@ -586,15 +586,15 @@ namespace vstd
 			std::advance(itr, i);
 			return *itr;
 		}
-		return boost::none;
+		return std::nullopt;
 	}
 
-	template <typename Container, typename Pred>
-	static boost::optional<typename Container::const_reference> tryFindIf(const Container &r, const Pred &t)
+	template<typename Container, typename Pred>
+	static std::optional<typename Container::const_reference> tryFindIf(const Container & r, const Pred & t)
 	{
 		auto pos = range::find_if(r, t);
 		if(pos == boost::end(r))
-			return boost::none;
+			return std::nullopt;
 		else
 			return *pos;
 	}
@@ -669,8 +669,8 @@ namespace vstd
 		return false;
 	}
 
-	template <class M, class Key, class F>
-	typename M::mapped_type & getOrCompute(M & m, Key const & k, F f)
+	template<class M, class Key, class F>
+	typename M::mapped_type & getOrCompute(M & m, const Key & k, F f)
 	{
 		typedef typename M::mapped_type V;
 
--- a/Global.h
+++ b/Global.h
@@ -101,6 +101,7 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #define _USE_MATH_DEFINES
 
 #include <algorithm>
+#include <any>
 #include <array>
 #include <atomic>
 #include <bitset>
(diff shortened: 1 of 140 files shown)


Global.h
```diff
@@ -576,24 +577,24 @@ namespace vstd
 		return i >= 0  &&  i < c.size();
 	}
 
-	template <typename Container, typename Index>
-	boost::optional<typename Container::const_reference> tryAt(const Container &c, Index i)
+	template<typename Container, typename Index>
+	std::optional<typename Container::value_type> tryAt(const Container & c, Index i)
```
IvanSavenko, line 581: Why const_ref has been changed to value_type and only here?
Similar methods don't have similar change

lib/CArtHandler.cpp
```diff
@@ -740,15 +740,16 @@ void CArtHandler::erasePickedArt(const ArtifactID & id)
 {
 	CArtifact *art = objects[id];
 
-	if(auto artifactList = listFromClass(art->aClass))
+	if(auto artl = listFromClass(art->aClass))
```
IvanSavenko, line 743: Why the change? Old code was fine with access via dereferences

lib/CArtHandler.cpp
```diff
@@ -758,20 +759,20 @@ void CArtHandler::erasePickedArt(const ArtifactID & id)
 		logMod->warn("Problem: cannot find list for artifact %s, strange class. (special?)", art->getNameTranslated());
 }
 
-boost::optional<std::vector<CArtifact*>&> CArtHandler::listFromClass( CArtifact::EartClass artifactClass )
+std::optional<std::reference_wrapper<std::vector<CArtifact *>>> CArtHandler::listFromClass(CArtifact::EartClass artifactClass)
```
IvanSavenko, line 762: Perhaps remove this method and merge this code into `CArtHandler::erasePickedArt`?
This is the only one usage of this method and removal of optionals might actually clear up this code which got even more complex with this change

lib/serializer/BinaryDeserializer.h
```diff
@@ -43,10 +44,21 @@ class DLL_LINKAGE BinaryDeserializer : public CLoaderBase
...
+		};
+
+		template <class V>
+		using mpl_types = typename mpl_types_impl<V>::type;
+
 		VariantLoaderHelper(Source & source):
 			source(source)
 		{
-			boost::mpl::for_each<typename Variant::types>(std::ref(*this));
+			boost::mpl::for_each<mpl_types<Variant>>(std::ref(*this));
```
IvanSavenko, line 61: Love removing boost?
Perhaps we can replace this class with something like this:
https://stackoverflow.com/a/57643845
(and remove final usage of mpl along the way)

Global.h
rilian-la-te: @IvanSavenko because other similar methods are unused. Const_ref is not allowed in std::optional, and I tried but unsuccessful in using std::reference_wrapper here. Perhaps remove unused methods?

lib/CArtHandler.cpp
rilian-la-te: @IvanSavenko because there is std::reference_wrapper now here. We should get a reference from it somewhere. So, it is split to 2 methods - firstly, get an optional, and second - get actual reference from wrapper, so, I actually split this check to 2.

Global.h
IvanSavenko: In that case:
- remove unused methods
- optionally - remove this method (only 1 usage in code)

Global.h
```diff
@@ -101,6 +101,7 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #define _USE_MATH_DEFINES
 
 #include <algorithm>
+#include <any>
```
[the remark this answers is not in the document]
rilian-la-te: done

Global.h
rilian-la-te: unused functions removed, method converted to lambda and moved inside VCAI

lib/CArtHandler.cpp
rilian-la-te: done

lib/serializer/BinaryDeserializer.h
rilian-la-te: About MPL - i think it is next time, because it is not a final usage.

rilian-la-te: This will require a separate PR, not simple one as this.

rilian-la-te:
Will fix CI tomorrow.

rilian-la-te:
@IvanSavenko CI fixed, please, look again)

IvanSavenko (approved):

PR #2019, opened 2023-04-17
Rule-based modernization part1: cpp

rilian-la-te:
Rules as with headers, but without readability-qualified-auto (will be too big).

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -526,7 +526,7 @@ void CBattleAI::attemptCastingSpell()
 					if(ap.attackerDamageReduce > 0)
 						swb->removeUnitBonus(Bonus::UntilBeingAttacked);
 
-					for(auto affected : ap.affectedUnits)
+					for(const auto & affected : ap.affectedUnits)
 					{
 						swb = state.getForUpdate(affected->unitId());
 						*swb = *affected;
--- a/AI/BattleAI/BattleExchangeVariant.cpp
+++ b/AI/BattleAI/BattleExchangeVariant.cpp
@@ -18,7 +18,6 @@ AttackerValue::AttackerValue()
 }
 
 MoveTarget::MoveTarget()
-	: positions()
 {
 	score = EvaluationResult::INEFFECTIVE_SCORE;
 }
@@ -334,9 +333,9 @@ std::vector<const battle::Unit *> BattleExchangeEvaluator::getExchangeUnits(
 		return exchangeUnits;
 	}
 
-	for(int turn = 0; turn < turnOrder.size(); turn++)
+	for(auto & turn : turnOrder)
 	{
-		for(auto unit : turnOrder[turn])
+		for(auto unit : turn)
 		{
 			if(vstd::contains(allReachableUnits, unit))
 				exchangeUnits.push_back(unit);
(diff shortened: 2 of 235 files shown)


rilian-la-te:
@Nordsoft91 what is wrong with this change?

IvanSavenko:
> @Nordsoft91 what is wrong with this change?

And what is *good* with this change?
- replacing typedef's with using's does makes code more clear
- using range-based loops where possible does make code more clear

Rest of rules? Improvements are doubtful IMO

> @IvanSavenko should I not to do any modernization, even rule-based?

I am not opposed to these changes, but I am very doubtful at their importance/priority. 
- Do these changes reveal/fix potential bugs? No
- Do these changes improve readability? Except for a few rules - no
- Do these changes improve code in other ways, e.g. performance? Maybe, but difference would be like 0.001% or something, so no.
- Have we reached the point where minor code style issues are largest problem with our codebase? No

rilian-la-te:
@IvanSavenko 

Using auto is a preparation to reduce usage of modifable types like CCreature or CStack,  because there are many places where we can use just Creature or Battle::Unit. 
About readability - I think removing redundant functions greatly improves readablitiy (and make code have less LoC).

IvanSavenko:
> Using auto is a preparation to reduce usage of modifable types like CCreature or CStack, because there are many places where we can use just Creature or Battle::Unit. 

A lot of occurences of "CStack" in code are method or member declarations. Not convertible to auto. And I am doubtful at readability improvements of replacing clearly specified type names with auto. 

Sure, IDE can show actual type, but type is no longer visible at glance.

> About readability - I think removing redundant functions greatly improves readablitiy (and make code have less LoC).

Don't see such changes in your PR. If you're referring to removal of constructors then improvement is doubtful in some cases and needs manual check / conversion IMO.

Also, less LoC != more readability

rilian-la-te:
1. In this PR auto is shown inside iterators and when we cast something. I do not see other places where full type is changed to auto.
2. Not only inside constructors. Also there are extern duplicates removed, especially in AI part. Why constructors removal is doubtful?

IvanSavenko:
> 1. In this PR auto is shown inside iterators and when we cast something. I do not see other places where full type is changed to auto.

In that case it has almost 0 effect on amount of occurrences of classes like CStack in code.

And personal opinion, but while this looks OK for dynamic_cast's / make_shared, I am not sure if changes like this one are an improvement:
`auto number = static_cast<int>(floatingPoint);`
IMO there is no point in using auto for primitive types.

>     2. Not only inside constructors. 

Not only? Like where?

> Also there are extern duplicates removed, especially in AI part. 

extern's should've been moved to a single header and included instead.

>Why constructors removal is doubtful?

In-place init vs constructor init, discussion number 9001?

OK for small, primitive structs, but large, proper classes should use constructors instead

rilian-la-te:
@IvanSavenko no, it is not init vs constructor, it is different discussion, illustrated below.

So, why 
```c++
struct A {
      int x;
      std::string y;
      A (int x) : x(x), y() {}
```
is better than
```c++
struct A {
      int x;
      std::string y;
      A (int x) : x(x) {}
```
?
About externs - externs was both in header file and in all cpp. Now CPP ones is removed.

rilian-la-te:
@IvanSavenko most changes here:
1. 1 -> 2 (in some cases almost all constructor init is removed)
2.Where type before cast replaced to auto, or type inside for replaced to auto.
3. Other rule changes like typedef -> using and auto -> const auto &.

I can move this PTR to header too, if desired.  There was also CB, in each of ai file.

Nordsoft91:
@rilian-la-te ,
This change 200+ files. So, the minimal "price" is:
 - full rebuild after pulling changes for everybody (for me 30-40 min)
 - resolve conflict for all PRs (thus, switch branch, resolve conflicts, compile, sanity check - 30min per PR)


However, if we leave value question aside, I strictly disagree with replacement of explicitly specified types with auto.
Also, let's stop touching formatting in automatic way.

PR #2025, opened 2023-04-18
Improved zone placement

DjWarmonger:
- Added construction algorithm: Place zones initially on a grid based on graph distances between them
- Use Fruchterman-Reingold algorithm for zone placement optimization, with zone stiffness growing in every iteration.
- Use trigonometric factor in metric to create curved and fuzzy zone edges.


fraang:
> * Added construction algorithm: Place zones initially on a grid based on graph distances between them
> 
>     * Use Fruchterman-Reingold algorithm for zone placement optimization, with zone stiffness growing in every iteration.
> 
>     * Use trigonometric factor in metric to create curved and fuzzy zone edges.

Out of curiosity, what was the motivation for the change?

DjWarmonger:
Player feedback from PvP community.

Besides, the old solution gave so-so results and some parts of the algorithm didn't even work as planned.

PR #2030, opened 2023-04-18
Boats refactoring

Nordsoft91:


rilian-la-te:
@Nordsoft91 will be possible to load boat animation from mod?

Nordsoft91:
Yes

lib/GameConstants.h
```diff
@@ -1241,6 +1241,27 @@ class BattleField : public BaseForID<BattleField, si32>
 	DLL_LINKAGE static BattleField fromString(const std::string & identifier);
 };
 
+class TransportId
```
rilian-la-te, line 1244: Why this is ID_LIKE and not Identifier?

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -935,18 +935,14 @@ si32 CGHeroInstance::getManaNewTurn() const
...
 	{
-	case EAlignment::GOOD:
-		return 1;
-	case EAlignment::EVIL:
-		return 0;
-	case EAlignment::NEUTRAL:
-		return 2;
-	default:
-		throw std::runtime_error("Wrong alignment!");
+		case EAlignment::EVIL : return TransportId::ETransportId::BOAT_EVIL;
```
IvanSavenko, line 942: Move to hero class config?

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -1085,16 +1085,15 @@ void CGTownInstance::clearArmy() const
...
 
-int CGTownInstance::getBoatType() const
+TransportId CGTownInstance::getTransportType() const
 {
 	switch (town->faction->alignment)
 	{
-	case EAlignment::EVIL : return 0;
-	case EAlignment::GOOD : return 1;
-	case EAlignment::NEUTRAL : return 2;
+		case EAlignment::EVIL : return TransportId::ETransportId::BOAT_EVIL;
```
IvanSavenko, line 1092: Move to config?

lib/mapObjects/CObjectHandler.cpp
```diff
@@ -457,16 +457,16 @@ IBoatGenerator::EGeneratorState IBoatGenerator::shipyardStatus() const
...
 		return GOOD; //OK
-	else if(t->blockingObjects.front()->ID == Obj::BOAT)
+	else if(t->blockingObjects.front()->ID == Obj::TRANSPORT)
 		return BOAT_ALREADY_BUILT; //blocked with boat
 	else
 		return TILE_BLOCKED; //blocked
 }
 
-int IBoatGenerator::getBoatType() const
+TransportId IBoatGenerator::getTransportType() const
```
IvanSavenko, line 466: Perhaps make this method pure virtual and reimplement it in CGShipyard?

All other inheritors already use own implementation anyway.
Also - would be great to be able to define what "transport" is built in shipyard in config.

IvanSavenko (commented):
Nice!

- Have other "transport" types (for layers other than water) have been tested?

- Perhaps transport should define movement point amount and not hero? ~Math is clearly different (fixed 2500 for boats vs based on creature speed for land)~ (EDIT: might we wrong, can't remember) and might be useful for other transport types, or even for "special" boats that are faster/slower than standard.

client/windows/GUIClasses.cpp
```diff
@@ -1089,7 +1089,7 @@ CShipyardWindow::CShipyardWindow(const TResources & cost, int state, int boatTyp
 	std::string boatFilenames[3] = {"AB01_", "AB02_", "AB03_"};
 
 	Point waterCenter = Point(bgWater->pos.x+bgWater->pos.w/2, bgWater->pos.y+bgWater->pos.h/2);
-	bgShip = std::make_shared<CAnimImage>(boatFilenames[boatType], 0, 7, 120, 96, 0);
+	bgShip = std::make_shared<CAnimImage>(boatFilenames[boatType.getNum()], 0, 7, 120, 96, 0);
```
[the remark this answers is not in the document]
Nordsoft91: doesn't make much sense - shipyard will be revised to support new transports

lib/GameConstants.h
Nordsoft91: what do you mean?

lib/mapObjects/CGHeroInstance.cpp
Nordsoft91: out of this PR

lib/mapObjects/CGTownInstance.cpp
Nordsoft91: out of this PR

lib/mapObjects/CObjectHandler.cpp
Nordsoft91: this change is only about boats, not about shipyards

Nordsoft91:
> Have other "transport" types (for layers other than water) have been tested?

Yes, have a mod prototype for that.

> Perhaps transport should define movement point amount and not hero? Math is clearly different (fixed 2500 for boats vs based on creature speed for land) (EDIT: might we wrong, can't remember) and might be useful for other transport types, or even for "special" boats that are faster/slower than standard.

Right, that's could be more useful, but for now it's solved by adding/removing bonuses, related to movement.
I'll investigate how complex would be to change logic, but not everything at once

lib/GameConstants.h
rilian-la-te: For example - in this file you can find ETerrainId and TerrainId, which done by identifiers.

rilian-la-te:
@IvanSavenko now army movement and all you described works on bonuses, so, it can simply be another node, why not. Like silent artifact.

client/windows/GUIClasses.cpp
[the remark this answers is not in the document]
IvanSavenko: Understood. This shipyard revision is in your plans for 1.3, correct?

lib/mapObjects/CGHeroInstance.cpp
Nordsoft91: no, didn't plan to refactor hero unless there will be such request

config/objects/moddables.json
```diff
@@ -115,9 +115,38 @@
 			}
 		},
 		"types" : {
-			"evil" : { "index" : 0,	"rmg" : { "mapLimit" : 64 } },
-			"good" : { "index" : 1, "rmg" : { "mapLimit" : 64 } },
-			"neutral" : { "index" : 2, "rmg" : { "mapLimit" : 64 } },
+			"evil" : 
+			{ 
+				"index" : 0,	
+				"rmg" : { "mapLimit" : 64 },
```
DjWarmonger, line 121: RMG threats Boat as a special object and doesn't really check this limit. Engine doesn't have such limitations either, afaik.

lib/GameConstants.h
```diff
@@ -756,7 +756,7 @@ class Obj
 		ARTIFACT = 5,
 		PANDORAS_BOX = 6,
 		BLACK_MARKET = 7,
-		BOAT = 8,
+		TRANSPORT = 8,
```
DjWarmonger, line 759: But why? Everyone knows what Boat is, but "Transport" is unclear to say at least.

krs0: That one would be Airship or skyship...
So Boat (aka Ship) could be VEHICLE or CRAFT in generic form and Ship in particular cases when you want to refer to water vehicles.

Nordsoft91: Maybe only renaming (without functional changes) could be in separate PR if we agreed on something. It's a lot of stuff to rename, so I'm with @DjWarmonger here

lib/GameConstants.h
Nordsoft91: fixed, thanks!

PR #2043, opened 2023-04-20
Fix for Tactics movement is limited #2017

### LINKED ISSUE #2017: Tactics movement is limited
**Describe the bug**
Seems like creatures movement is limited with their speed during tactic phase

**To Reproduce**
1. Start battle with hero, having tactics skill and some slow creatures
2. Try to move creature to far hex

**Expected behavior**
Possible to move at any cell

**Actual behavior**
Possible to move only to hexes within speed range

**Did it work earlier?**
Yes

**Screenshots**

https://user-images.githubusercontent.com/9308612/232561355-9ff76e1d-e984-4357-9858-f8c2688d76cd.mp4

**Version**
 - Version: 1.3 (develop)

krs0:
During tactics units instead of moving in whole tactics range, moved only as far as they could normally move (speed).
This is corrected now. When obtaining battleGetAvailableHexes movement range is obtained only for hovered unit. For active unit tactics range is returned.

Resolves #2017

### DIFF
--- a/mapeditor/inspector/inspector.cpp
+++ b/mapeditor/inspector/inspector.cpp
@@ -472,6 +472,9 @@ void Inspector::setProperty(CGLighthouse * o, const QString & key, const QVarian
 void Inspector::setProperty(CGPandoraBox * o, const QString & key, const QVariant & value)
 {
 	if(!o) return;
+	
+	if(key == "Message")
+		o->message = value.toString().toStdString();
 }
 
 void Inspector::setProperty(CGEvent * o, const QString & key, const QVariant & value)


AI/BattleAI/BattleAI.cpp
```diff
@@ -282,7 +282,7 @@ BattleAction CBattleAI::activeStack( const CStack * stack )
 BattleAction CBattleAI::goTowardsNearest(const CStack * stack, std::vector<BattleHex> hexes) const
 {
 	auto reachability = cb->getReachability(stack);
-	auto avHexes = cb->battleGetAvailableHexes(reachability, stack, true);
+	auto avHexes = cb->battleGetAvailableHexes(reachability, stack, false);
```
Nordsoft91, line 285: Why battleAI changes are needed?

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -546,7 +546,7 @@ void CBattleInfoCallback::battleGetTurnOrder(std::vector<battle::Units> & turns,
 		battleGetTurnOrder(turns, maxUnits, maxTurns, actualTurn + 1, sideThatLastMoved);
 }
 
-std::vector<BattleHex> CBattleInfoCallback::battleGetAvailableHexes(const battle::Unit * unit, bool isActiveStack) const
+std::vector<BattleHex> CBattleInfoCallback::battleGetAvailableHexes(const battle::Unit * unit, bool getMovementRange) const
```
Nordsoft91, line 549: get prefix is usually for some return value. Maybe give it a name like: useMovementRange or obtainMovementRange or isUseMovementRange?

AI/BattleAI/BattleAI.cpp
krs0: I added a new parameter to getAvailableHexes, so that this function knows about hovered unit. And thi function is used in AI for whatever purposes.

PR #2048, opened 2023-04-22
RMG tweaks

DjWarmonger:
This PR is for misc / arbitrary adjustments that are likely to come up every now and then.

- Make surface zones have less obstacles / more free area


DjWarmonger:
Replaced by https://github.com/vcmi/vcmi/pull/2218

PR #2053, opened 2023-04-23
Parellel RMG implementation

DjWarmonger:
- [x] Remove dependency of quest artifacts between consecutive zones
- [x] Refactor quest artifact counter
- [x] Refactor Prison counter
- [x] Give every Zone private random number generator
- [x] Implement thread pool
- [x] Feed thread pool with Modificators
- [x] Protect every shared object or state with mutex
- [x] Force single-thread implementation with an option


lib/rmg/QuestArtifactPlacer.cpp
```diff
@@ -38,8 +38,7 @@ void QuestArtifactPlacer::addQuestArtZone(std::shared_ptr<Zone> otherZone)
 
 void QuestArtifactPlacer::addQuestArtifact(const ArtifactID& id)
 {
-	logGlobal->info((boost::format("Need to place quest artifact artifact %s")
-		% VLC->arth->getById(id)->getNameTranslated()).str());
+	logGlobal->info("Need to place quest artifact artifact %s", VLC->arth->getById(id)->getNameTranslated());
```
rilian-la-te, line 41: @DjWarmonger why not VLC->artifacts() instead of VLC->arth? I think using entities is better when we can.

lib/rmg/QuestArtifactPlacer.cpp
```diff
@@ -84,10 +83,10 @@ void QuestArtifactPlacer::placeQuestArtifacts(CRandomGenerator * rand)
...
 
 			auto artifactToReplace = *RandomGeneratorUtil::nextItem(artifactsToReplace, *rand);
-			logGlobal->info((boost::format("Replacing %s at %s with the quest artifact %s")
-				% artifactToReplace->getObjectName()
-				% artifactToReplace->getPosition().toString()
-				% VLC->arth->getById(artifactToPlace)->getNameTranslated()).str());
+			logGlobal->info("Replacing %s at %s with the quest artifact %s",
+				artifactToReplace->getObjectName(),
+				artifactToReplace->getPosition().toString(),
+				VLC->arth->getById(artifactToPlace)->getNameTranslated());
```
rilian-la-te, line 89: Same question as above

lib/rmg/CMapGenerator.cpp
```diff
@@ -389,23 +396,41 @@ int CMapGenerator::getNextMonlithIndex()
...
+	{
+		if (map->map().allowedHeroes[j])
+			ret.push_back(HeroTypeID(j));
+	}
+	return ret;
+}
+
 void CMapGenerator::banQuestArt(const ArtifactID & id)
 {
+	//TODO: Protect with mutex
```
IvanSavenko, line 426: If two threads can simultaneously generate two quests this might lead to RMG producing different maps from the same seed/settings.

Although not sure if we need generation of identical maps in RMG via settings same seed, so feel free to ignore.

DjWarmonger: That's why there is this task:

> Give every Zone private random number generator

Each zone will generate exactly the same regardless their order. There is only a risk that two zones may try to place an artifact in third zone at the same time, and then the order will be undefined. But artifacts on map are randomized at game start anyway, so it doesn't change much.

DjWarmonger:
Whoopsie, actually random execution order makes treasures vary even for the same seed. Only zone placement and basic objects stays intact.

rilian-la-te:
Why boost is better?

DjWarmonger:
> Why boost is better?

Because compiler can find it. We use boost::thread also in other places.

lib/mapping/ObstacleProxy.cpp
```diff
@@ -8,21 +8,9 @@
  *
  */
 
-#include "StdInc.h"
+#include "ObstacleProxy.h"
```
IvanSavenko, line 11:
```suggestion
#include "StdInc.h"
#include "ObstacleProxy.h"
```

lib/mapping/ObstacleProxy.h
```diff
@@ -0,0 +1,77 @@
...
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#pragma once
+
+#include "StdInc.h"
```
IvanSavenko, line 13: Stdinc should be in cpp only
```suggestion
```

lib/rmg/CZonePlacer.h
```diff
@@ -22,10 +22,11 @@ class CRandomGenerator;
...
 
-using TZoneVector = std::vector<std::pair<TRmgTemplateZoneId, std::shared_ptr<Zone>>>;
-using TZoneMap = std::map<TRmgTemplateZoneId, std::shared_ptr<Zone>>;
-using TForceVector = std::map<std::shared_ptr<Zone>, float3>;
-using TDistanceVector = std::map<std::shared_ptr<Zone>, float>;
+typedef std::vector<std::pair<TRmgTemplateZoneId, std::shared_ptr<Zone>>> TZoneVector;
+typedef std::map<TRmgTemplateZoneId, std::shared_ptr<Zone>> TZoneMap;
+typedef std::map<std::shared_ptr<Zone>, float3> TForceVector;
+typedef std::map<std::shared_ptr<Zone>, float> TDistanceVector;
+typedef std::map<int, std::map<int, size_t>> TDistanceMap;
```
IvanSavenko, line 29: Why replace using's with typedef?

Nordsoft91:
good feature! Few items to test:
- assuming, that map editor generator works as well
- water maps
- progress bar in map editor
- same maps are generated using same seed (if all settings are same)

DjWarmonger:
> good feature! Few items to test:
> 
> * assuming, that map editor generator works as well
> * water maps
> * progress bar in map editor
> * same maps are generated using same seed (if all settings are same)

Water - yes
Progress bar - yes
Same maps - unfortunatelly not exactly and I don't see a way to fix it at this point.

RMG seed 0 - 1
RMG seed 0 - 2

Treasures start to diverge. Everything that comes after ConnectionsPlacer has undetermined order so it's going to be random.

IvanSavenko:
> Progress bar - yes, actually it's smoother.

Interesting - is there any way to provide progress bar to client? 
Sitting on lobby and waiting for "something" to happen while game generates map in background is not exactly user-friendly :(

Nordsoft91:
> Everything that comes after ConnectionsPlacer has undetermined order so it's going to be random.

Is this under feature flag? Should be possible to disable and use single thread which guarantees deterministic result (for debugging for instance). Also, seed field in map editor loses its sense then.

Could you please also test network mode?

DjWarmonger:
Okay, I think it kinda works. However I realized that player or team settings tend to be random, and depending on that RMG may be used or not, changing its value. I wasn't able to set all player / team settings to fixed values. Best I got was:

image

You know, player generation in RMG is still a mess, but that seems to be separate issue. When "singleThread" flag is enabled, zones are processed in deterministic order, at least.

DjWarmonger:
Also, localhost works fine, noth players get same map as far as I see.

IvanSavenko:
Currently map is loaded (or created, in case of RMG) by server only and then sent to client, so multiplayer will be fine even same seed does not generates the same map.

Nordsoft91 (approved):
Tested a bit, including MP, seems work.
Few issues discovered, but barely possible to be connected with PR, in particular:
- return/escape hotkeys work from second time, first press is ignored (@IvanSavenko FYI)
- E hotkey ends turn without warning (heroes still have movement points) for host player if no action done in the beginning of turn 
- spectator mode crash at battle

UPD: all issues are reproducible on develop

PR #2054, opened 2023-04-23
Fix killing hero with grail

nullkiller:
When killing hero with grail we end up in endless loop


nullkiller:
[kill-hero-with-grail.zip](https://github.com/vcmi/vcmi/files/11303253/kill-hero-with-grail.zip)

Alucard648:
I think, there should be new artifact class called Ultimate Artifact, that encompasses all ultimate artifacts (Grail in H3, overpowered artifacts in H1 & 2 etc.) If hero with these artifacts is defeated, those artifacts are destroyed, not captured, complete with message told to victor.

PR #2055, opened 2023-04-23
Artifact related code refactoring part2 (client CArtifactHolder)

SoundSSGood:
Improved code readability.
Fixed minor highlighting art places errors on the hero exchange screen (if the second hero is an allied player).
The altar now works correctly with the backpack limit.
CArtifactHolder is now prepared to implement backpack window (with all artifacts).

It is better to do the code review for CArtifactHolder from scratch, not through diff.


client/widgets/CArtifactHolder.h
```diff
@@ -55,138 +58,171 @@ class CCommanderArtPlace : public CArtPlace
...
+	const std::vector<Point> slotPos =
+	{
+		Point(509,30),  Point(567,240), Point(509,80),  //0-2
+		Point(383,68),  Point(564,183), Point(509,130), //3-5
+		Point(431,68),  Point(610,183), Point(515,295), //6-8
+		Point(383,143), Point(399,194), Point(415,245), //9-11
+		Point(431,296), Point(564,30),  Point(610,30), //12-14
+		Point(610,76),  Point(610,122), Point(610,310), //15-17
+		Point(381,296) //18
 	};
```
SoundSSGood, line 143: I think the constants should be in a separate file. But let it stay here for now.

client/widgets/CArtifactHolder.h
```diff
@@ -55,138 +58,171 @@ class CCommanderArtPlace : public CArtPlace
...
+	void swapArtifacts(const ArtifactLocation & srcLoc, const ArtifactLocation & dstLoc);
+	void pickUpArtifact(CHeroArtPlace & artPlace);
+};
+
+class CArtifactsOfHeroAltar : public CArtifactsOfHeroBase
+{
+public:
+	std::set<const CArtifactInstance*> artifactsOnAltar;
+	ArtifactPosition pickedArtFromSlot;
+	std::shared_ptr<CArtifactFittingSet> visibleArtSet;
```
SoundSSGood, line 174: Perhaps it would be better to make it not a pointer. But this header file does not contain the necessary includes. Enabling include is not consistent with the VCMI style. Is there a way to make this not a pointer?

IvanSavenko: - Option 1: use unique_ptr if ownership is non-transferrable
- Option 2: use include, but split this file on 1-file-per-class basis & use includes where necessary

And actually splitting this file to 1-per-class may not be a bad idea irregardless of what happens with this member - it already grew quite large (not insisting on it though)

krs0:
Is now possible to make a cheat command like 
Vcmiartifact goldenGoose?

SoundSSGood:
> Vcmiartifact goldenGoose?

Nope. what the issue with Vcmiartifact goldenGoose? haven't heard of this

krs0:
Its a whish for a cheat comand like vcmiarmy imp... it should give you the specified artifact

SoundSSGood:
Either way it requires changes on the server side. Out of scope this PR. Added to my todo list

krs0:
I was not asking you to implement this, only asking if necessary functions for such a change are available.
IIRC a function like findArtifactById was missing.

IvanSavenko:
> I was not asking you to implement this, only asking if necessary functions for such a change are available. IIRC a function like findArtifactById was missing.

Such function exists for any "registrable" object - requestIdentifier / getIdentifier in our modhandler. Creatures, heroes, artifacts, spells - anything goes.

SoundSSGood:
ready for review

IvanSavenko (approved):
Looks OK, although I am not familiar with artifacts code well enough for a detailed review

Nordsoft91:
@SoundSSGood , will you resolve conflicts?

PR #2067, opened 2023-04-24
Fix tests compiling

Nordsoft91:
To be able to run tests, I removed ERM/scripting tests.

Most of tests failed, but it reflects reality about develop branch at most

### DIFF
--- a/lib/CGameState.cpp
+++ b/lib/CGameState.cpp
@@ -722,7 +722,7 @@ void CGameState::preInit(Services * services)
 	this->services = services;
 }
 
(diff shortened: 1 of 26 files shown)


IvanSavenko:
Perhaps we should enable running tests on CI to prevent future regressions?
Assuming that it is possible to run compilation but ignore failing tests in CI.

rilian-la-te:
@Nordsoft91 what is a difference with #1623? 

Is tests running without H3 data?

rilian-la-te:
On #1623 all tests passing.

rilian-la-te:
If your branch is also not working without H3 data, I suggest to merge #1623 without last commit instead, because that branch is more complete.

PR #2076, opened 2023-04-26
Implement scalable adventure map window

IvanSavenko:
Adventure Map:
- [x] Load layout of Adventure Map using configurable UI system
- [x] Scale layout to any user-selected resolution
- [x] Add support for optional elements that only appear / change at specific width/height
- [x] Find a way to fill unused area in resource bar
- [x] Implement overlays - UI elements that should always appear on top of map view (e.g. map corner gems)
- [x] ~Implement advanced overlays - UI elements that appear on top of map view and should intercept mouse events~

Game Resolution:
- [x] Refactor `recreateWindow` method
- [x] Look for a way to implement better UI upscale filter
- [x] Implementing UI scale option
- [x] Consider possibility of resolution / UI scale change without app restart

Settings Window:
- [x] Hide resolution selector on mobile devices
- [ ] Hide fullscreen options on mobile devices
- [ ] Add Borderless Fullscreen option in game settings (opposite to "True" fullscreen)
- [x] Allow player to select UI scaling
- [x] Show all resolutions supported by monitor in game settings window

Launcher:
- [x] Find a way for Launcher to show list of resolutions supported by monitor (initialize SDL in launcher?)
- [x] Hide resolution selector & fullscreen options on mobile devices
- [ ] Allow player to select UI scaling

Mods:
- [ ] Remove extra resolutions submod from vcmi-extras

Shortcuts system #2078:
- [x] Implement callbacks for adventure map buttons
- [x] Implement hiding of UI elements based on conditions (e.g. hide infobox in world view)
- [x] Implement blocking of unusable buttons (e.g. quest log should be disabled when there are no quests)
- [x] Implement blocking of UI when player should not interact with it (e.g. don't react on clicks during enemy turn)

Regressions:
- [x] Recheck castle interface and kingdom overview for possible regressions
- [x] Fix max chat message line length calculation
- [x] Make sure that world view mode works correctly
- [x] Make sure that surface/underground and sleep/wake toggle works
- [x] Fix broken left-clicking on map view


rilian-la-te:
@IvanSavenko infobar will be scaled too?

IvanSavenko:
> @IvanSavenko infobar will be scaled too?

Not against it, but not planning on doing that either.

Implementing larger infobar for higher resolutions should be quite simple after this PR - either infobar that is higher than standard one to fill empty space or completely different one (that requires redesign of adventure map layout).
However right now my goal is to make UI similar-ish to how it was before, but without need for custom assets and with support for any resolution higher than 800x600.

dydzio0614:
Maybe I and wrong, but I think qt would probably totally close path to potential hardware acceleration, shader support and similar stuff

IvanSavenko:
> Ha, told you we should eventually move to Qt for full flexibility.

Don't see much point to be honest.
If we were starting project from scratch? Maybe. But even in such scenario I have doubts whether this is a better option. Might as well switch to javascript & electron because this is a trend apparently.

And at this point there is simply no reason to use Qt IMO. Even here main issues are not scalability but how to *describe* this scalability without editing H3 assets. Adventure map background has a lot of decorative elements that are tricky to upscale.

rilian-la-te:
@DjWarmonger I think SDL is a normal for vcmi, and I see no reason to rewrite entire client into Qt.

dydzio0614:
maybe it's time to embed AI asset upscaler button in VCMI launcher :laughing:

IvanSavenko:
Marking as ready for review. Some minor things still left to do, will probably fix them tomorrow, but othervice should be ready.

### Remaining changes
- Self-review and cleanup of changes
- Fullscreen selector in Launcher - will look into it tomorrow,
- Rename some classes to adjust their name to their actual functionality - will look into it tomorrow
- "true" overlay element (e.g. hero/town lists on top of map view) are postponed - were request on Slack to support nicer UI redesign, not used in H3, might look into it after cleaning up our event input system.
- Disabling fullscreen options on mobile systems - not *required* since this option (as before) is unused on mobile systems, but should not cause any issues either. Will look into it later, need minor UI redesign for settings window.
- UI scale option in Launcher - will look into it later, optional & needs some potentially duplicated code with client.

### Functionality changes
1) VCMI can now use any resolution supported by monitor, technically - even those less than 800x600 via downscaling
2) Resolution of VCMI can be changed at any point via game options. Change is applied immediately, without restart
3) VCMI now has 3 clearly defined screen modes:
- - Windowed: game runs in window, at resolution specified by player
- - Windowed fullscreen: game runs in fullscreen, resolution ALWAYS matches desktop resolution. Player-selected resolution is ignored. This is also the only option available on mobile devices
- - True fullscreen: game runs in fullscreen, with display resolution changed to one specified by player
4) VCMI now has "Interface Scaling" option. Changing this option from default value of 100% will downscale or upscale game window, allowing players to play at FullHD / 2k / 4k monitors with upscaling instead of requiring magnifying glass to play game. Note that upscaling range is limited - e.g. you need 1600x1200 resolution to play at 200% (double of H3 800x600)
5) "Extra resolutions" submod is deprecated, instead layout will be dynamically generated based on resolution. If vertical resolution is greater than 664 then VCMI will use HD-Mod like layout with 8 heroes slots (but no additional buttons since H3 does not have such graphics)

krs0:
About upscaling.... in the end I would like to use it without knowing resolution details or do math.

so if I select 4k resolution and 200% scaling the correct 'things' are chosen for m in the background right?

IvanSavenko:
Things to do / to fix according to discussion with krs on Slack:
- [x] Fix duplicated elements in resolutions list
- [ ] main menu is not centered after returning from game screen
- [ ] infobox dies after screen resize
- [ ] try to fix high DPI-aware manifest on Windows
- [ ] show hero list scroll buttons even at high resolutions
- [ ] draw border around unused area on side panel similar to infobox border

Probably in separate PR after this one:
- [ ] only show relevant controls in settings window
- [ ] block resolution selector for windowed fullscreen
- [ ] add borderless window / windowed fullscreen options to in-game settings
- [ ] add description for "Exclusive fullscreen" mode
- [ ] add description for "Borderless fullscreen windowed" mode

IvanSavenko:
> About upscaling.... in the end I would like to use it without knowing resolution details or do math.
> 
> so if I select 4k resolution and 200% scaling the correct 'things' are chosen for m in the background right?

I think you should be able to use it like that?
Invalid options for scaling are hidden from UI, so player should not end up in broken game state.
If you have 200% scaling available this means that you can use it - all math is done in background by game.

Laserlicht:
Is it also planned to use the possibility of filters with the UI scaling? (>100%)
xBRZ with VCMI would be really nice.

See Heroes III HD mod:
grafik
grafik

IvanSavenko:
Not in near future.
Currently I use filter provided by SDL library. And list of filters available in it is rather small:
- no filtering (aka nearest)
- antialiasing

Implementing custom filters likely not trivial

IvanSavenko:
@krs0 
Right now - full block, since this is how H3 works.

As for relaxing this block - can be looked into later, but not a priority for me at the moment.

IvanSavenko:
Presuming this to be complete, will merge in a few days unless somebody plans to review / test this.

Some minor to-do items still left, will do them separately in subsequent PR's - I'd rather go for merge now, to allow remaining tasks to be done in separate PR's instead of making this PR even larger

Issue #1276, opened 2022-12-23
Remove leftovers from WoG data


IvanSavenko:
Apparently we still have some leftovers from the time WoG was part of the main game.
We should remove remaining pieces and move this data to WoG if necessary:
- WoG objects leading to startup warnings (#1274 )
- WoG artifacts (don't forget to change value in defaultMods.json)
- Access to text files for stack experience

PR #2078, opened 2023-04-27
Basic shortcuts system

IvanSavenko:
Basic / initial version of shortcuts system. Main reason is to simplify changes in #2076 however this is something that we'll need at some point anyway, so don't see any need for workarounds.

- Game now uses EShortcut enumeration instead of SDL key codes
- GuiHandler translates key codes to shortcut values

Note that ShortcutHandler code is basically a placeholder for any future support for configurable / player-editable shortcuts which are not planned in this PR.


Global.h
```diff
@@ -543,10 +543,16 @@ namespace vstd
 		});
 	}
 
+	template<typename T>
+	T advance_r(const T &obj, int change)
```
Nordsoft91, line 547: what does _r mean? not possible to overload advance?

client/adventureMap/CInGameConsole.cpp
```diff
@@ -155,7 +155,7 @@ void CInGameConsole::keyPressed (const SDL_Keycode & key)
 			}
 			break;
 		}
-	case SDLK_UP: //up arrow
+	case EShortcut::SELECT_UP:
```
Nordsoft91, line 158: why is it SELECT_? Maybe MOVE_ or SCROLL_?

client/battle/BattleFieldController.cpp
```diff
@@ -243,7 +243,7 @@ std::set<BattleHex> BattleFieldController::getMovementRangeForHoveredStack()
 	if (!owner.stacksController->getActiveStack())
 		return result;
 
-	if (!settings["battle"]["movementHighlightOnHover"].Bool())
+	if (!settings["battle"]["movementHighlightOnHover"].Bool() && !GH.isKeyboardShiftDown())
```
Nordsoft91, line 246: how to deal with touchscreen?

client/battle/BattleInterfaceClasses.cpp
```diff
@@ -406,12 +407,12 @@ BattleResultWindow::BattleResultWindow(const BattleResult & br, CPlayerInterface
 	background->colorize(owner.playerID);
 	pos = center(background->pos);
 
-	exit = std::make_shared<CButton>(Point(384, 505), "iok6432.def", std::make_pair("", ""), [&](){ bExitf();}, SDLK_RETURN);
+	exit = std::make_shared<CButton>(Point(384, 505), "iok6432.def", std::make_pair("", ""), [&](){ bExitf();}, EShortcut::GLOBAL_CONFIRM);
```
Nordsoft91, line 410: suggest to rename CONFIRM with ACCEPT

client/gui/CGuiHandler.h
```diff
@@ -115,6 +116,8 @@ class CGuiHandler
 	/// returns current position of mouse cursor, relative to vcmi window
 	const Point & getCursorPosition() const;
 
+	ShortcutHandler & getShortcutsHandler();
```
Nordsoft91, line 119: it's non const, then
```suggestion
	ShortcutHandler & shortcutHandler();
```

client/gui/Shortcut.h
```diff
@@ -0,0 +1,140 @@
...
+{
+	NONE,
+
+	GLOBAL_CONFIRM,    // Enter
+	GLOBAL_CANCEL,     // Escape
+	GLOBAL_RETURN,     // Enter, Escape
+	GLOBAL_FULLSCREEN, // F4
+	GLOBAL_OPTIONS,
+	GLOBAL_BACKSPACE,
+	GLOBAL_MOVE_FOCUS,
```
Nordsoft91, line 22: can you leave a small description nearby each const what it stands for?
some of those names are not obvious

client/gui/ShortcutHandler.h
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+enum class EShortcut;
+using SDL_Keycode = int32_t;
+
+class ShortcutHandler
+{
+public:
+	/// returns list of shortcuts assigned to provided SDL keycode
+	std::vector<EShortcut> translateKeycode(SDL_Keycode key);
```
Nordsoft91, line 20: As I can imagine, return value is used to check if necessary shortcut is pressed.
If that's correct, consider `std::set<EShortcut>`

client/gui/ShortcutHandler.h
```diff
@@ -0,0 +1,24 @@
...
+
+	/// attempts to find shortcut by its unique identifier. Returns EShortcut::NONE on failure
+	EShortcut findShortcut(const std::string & identifier );
```
Nordsoft91, line 23: both could be either static or const

config/campaignSets.json
```diff
@@ -47,18 +47,5 @@
...
-		/// wog campaigns, currently has no assigned button in campaign screen and thus unused
-		"images" : [ {"x": 0, "y": 0, "name":"CAMPZALL"} ],
-		"exitbutton" : {"x": 658, "y": 482, "name":"CMPSCAN", "hotkey" : 27},
-		"items":
-		[
-			{ "x":90,  "y":72,  "file":"DATA/ZC1.H3C", "image":"CAMPZ01", "open": true},
-			{ "x":539, "y":72,  "file":"DATA/ZC2.H3C", "image":"CAMPZ02", "open": true},
-			{ "x":43,  "y":245, "file":"DATA/ZC3.H3C", "image":"CAMPZ03", "open": true},
-			{ "x":311, "y":242, "file":"DATA/ZC4.H3C", "image":"CAMPZ04", "open": true}
-		]
```
Nordsoft91, line 62: Please, keep, as it may be useful. Having this in config will help a lot if someone decide to implement

Nordsoft91 (commented):
Will you also create some wiki material with shortcut string consts?

Global.h
IvanSavenko: version that returns resulting value instead of modifying parameter.

Overriding should be possible but would be confusing due to having two functions with similar signature but one of which modifies passed value and another does not.

Personally I'd rather remove existing advance() and always return resulting value.
Only 8 hits in code for this function. Might be better to replace.

client/adventureMap/CInGameConsole.cpp
IvanSavenko: MOVE might be better... Okay, will rename slider-related hotkeys.

client/battle/BattleFieldController.cpp
IvanSavenko: This check existed before, I only moved it in here from keyPressed. 

And don't think that we can display movement highlight on *cursor hovering* on touchscreen anyway. So desktop-only feature both before my changes and after.

client/gui/CGuiHandler.h
IvanSavenko: Might turn it into const - right now this class is completely stateless, since it mostly placeholder code for future implementation.

Or rename method to `shortcutHandler` and member with same name - to `shortcutHandlerInstance`?

client/gui/ShortcutHandler.h
IvanSavenko: True, however right now this is basically placeholder class for future implementation (hopefully for 1.3, unless I end up too busy).
Proper implementation won't be stateless so static is a no. Const makes sense.

client/gui/Shortcut.h
IvanSavenko: Okay, will review naming.

client/gui/ShortcutHandler.h
IvanSavenko: Not exactly, it is used as list - only operation is calling function for each returned value. So no need for set (although it would also work)

config/campaignSets.json
IvanSavenko: Not against keeping it, but might make more sense to have this in wog mod. Highly doubt that wog campaigns would ever be supported or if there will be any interest in them.

Can keep it but in commented-out form.

Global.h
IvanSavenko: - removed vstd::advance - all usages can be replaced with operator ++
- added vstd::next, based on std::next (instead of advance_r)
(will push changes later)

IvanSavenko:
Updated according to review.

> Will you also create some wiki material with shortcut string consts?

Not against it but as it is now shortcuts are not really usable by modders. Probably a bit later once shortcuts get more usability. What is still needs to be done for shortcuts to be actually usable:
- buttons that can be configured in json should call shortcuts instead of having callback assigned. For example, in battle UI shortcut "defend" should be always available as key 'D', even if UI has no "defend" button. Similarly, we should have functionality to call "show queue" shortcut from button (or any other shortcut)
- similar changes should be done to adventure map UI - to provide possibility to add extra buttons (e.g. dig/show puzzle) directly to adventure map interface
- shortcuts & their assigned buttons should be defined in config and not hardcoded in code
- we need some UI where player can view & edit all in-game shortcut (UI itself is trivial - can be done similarly to settings window, but need code support for shortcut assignment)
- probably rather long-term task, but other ways to trigger shortcuts, suitable for other input methods (e.g. touch screen or game controller)

client/gui/ShortcutHandler.cpp
```diff
@@ -0,0 +1,362 @@
...
+
+#include "StdInc.h"
+
+#include "ShortcutHandler.h"
+#include "Shortcut.h"
+#include <SDL_keycode.h>
+
+std::vector<EShortcut> ShortcutHandler::translateKeycode(SDL_Keycode key) const
+{
+	static const std::multimap<SDL_Keycode, EShortcut> keyToShortcut = {
```
Nordsoft91, line 19: As adding/modifying/removing new shortcut requires modifications in 3 such places (EShortcut, keyToShortcut and shortcutNames below) it's better to keep all of them in single file.
Consider moving keyToShortcut and shortcutNames to Shortcut.h

config/campaignSets.json
Nordsoft91: certainly in WOG mod, but as for me removing this config should go together with adding it to WOG. Otherwise risk to lose it forever is high (as we lost WOG translations - old translations removed while new not added)

IvanSavenko: Sorry, but how can you lose something forever with version control systems?

IvanSavenko: Restoring in commented-out form, will be moved to wog later with #1276 

Still don't see any need for this though - not something that will ever be used & we have git history to find old stuff.

client/gui/ShortcutHandler.cpp
IvanSavenko: Not 3 places.
At the very least you also need to add actual usage of shortcut, in UI code of window that will use said shortcut and likely - button with assigned shortcut in widget configuration json.

So even if I do this, you still won't be able to define shortcuts in one place.

Also want to avoid placing unnecessary info in header, to avoid increase in build times & to keep API minimalistic.

And don't see a way to force compile-time check to detect invalid shortcuts - since it is legal for a shortcut to not have a name and assigned key.

PR #2086, opened 2023-04-28
Support customisable markets

Nordsoft91:
New handler added which allows to make new markets.
- [x] Make sure tests work
- [x] Prepare wiki for market handler

Test map:
[test_markets.vmap.zip](https://github.com/vcmi/vcmi/files/11349138/test_markets.vmap.zip)

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -1709,7 +1709,8 @@ void CPlayerInterface::stopMovement()
 void CPlayerInterface::showMarketWindow(const IMarket *market, const CGHeroInstance *visitor)
 {
 	EVENT_HANDLER_CALLED_BY_CLIENT;
-	if (market->o->ID == Obj::ALTAR_OF_SACRIFICE)
+	auto * o = dynamic_cast<const CGObjectInstance *>(market);
+	if(o && o->ID == Obj::ALTAR_OF_SACRIFICE)
 	{
 		//EEMarketMode mode = market->availableModes().front();
 		if (market->allowsTrade(EMarketMode::ARTIFACT_EXP) && visitor->getAlignment() != EAlignment::EVIL)
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -1720,7 +1720,10 @@ void CPlayerInterface::showMarketWindow(const IMarket *market, const CGHeroInsta
 	}
 	else
 	{
-		GH.pushIntT<CMarketplaceWindow>(market, visitor, market->availableModes().front());
+		if(market->allowsTrade(EMarketMode::CREATURE_UNDEAD))
+			GH.pushIntT<CTransformerWindow>(market, visitor);
+		else
+			GH.pushIntT<CMarketplaceWindow>(market, visitor, market->availableModes().front());
 	}
 }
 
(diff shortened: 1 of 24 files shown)


IvanSavenko:
Nice!

What about:
- configuration of markets in towns?
- a way to define skills / number of skills available in university? (including Conflux version)
- a way to define artifacts in black market (e.g. how many artifacts of each class are available), potentially using JsonRandom

Nordsoft91:
>> configuration of markets in towns?
<< town buildings are handled separately and this I want to change. But it'll be probably later
>> a way to define skills / number of skills available in university? (including Conflux version)
<< it's possible, I already made config for water university (from hota) which offers navigation permanently
>> a way to define artifacts in black market (e.g. how many artifacts of each class are available), potentially using JsonRandom
<< this one not done, but I rely mostly on objects from modders. Anyway new instance handler is quite extendable

client/CPlayerInterface.cpp
```diff
@@ -1709,7 +1709,8 @@ void CPlayerInterface::stopMovement()
 void CPlayerInterface::showMarketWindow(const IMarket *market, const CGHeroInstance *visitor)
 {
 	EVENT_HANDLER_CALLED_BY_CLIENT;
-	if (market->o->ID == Obj::ALTAR_OF_SACRIFICE)
+	auto * o = dynamic_cast<const CGObjectInstance *>(market);
+	if(o && o->ID == Obj::ALTAR_OF_SACRIFICE)
```
IvanSavenko, line 1713: Perhaps this object ID check should be removed and we should instead select UI based on allowed trade types?

client/windows/CTradeWindow.cpp
```diff
@@ -696,21 +696,24 @@ CMarketplaceWindow::CMarketplaceWindow(const IMarket * Market, const CGHeroInsta
...
+			case Obj::BLACK_MARKET:
+				title = CGI->generaltexth->allTexts[349];
+				sliderNeeded = false;
+				break;
+			case Obj::TRADING_POST:
+				title = CGI->generaltexth->allTexts[159];
+				break;
+			case Obj::TRADING_POST_SNOW:
+				title = CGI->generaltexth->allTexts[159];
+				break;
```
IvanSavenko, line 712: Perhaps move these hardcoded strings into object config?

lib/mapObjects/CGMarket.h
```diff
@@ -30,19 +28,19 @@ class DLL_LINKAGE IMarket
...
-		h & o;
-	}
 };
 
 class DLL_LINKAGE CGMarket : public CGObjectInstance, public IMarket
 {
 public:
+	
+	std::set<EMarketMode::EMarketMode> marketModes;
+	int marketEfficacy = 5;
```
IvanSavenko, line 38: Any scenarios in which this value is not loaded from config?

lib/mapObjects/CGMarket.h
```diff
@@ -75,7 +74,12 @@ class DLL_LINKAGE CGBlackMarket : public CGMarket
 class DLL_LINKAGE CGUniversity : public CGMarket
 {
 public:
+	int skillsTotal = 4;
```
IvanSavenko, line 77: Perhaps simply use value provided in config?
If config defines 4 skills - then university will have 4 skills. If number is different - then well, university will also have different number of skills.

lib/mapObjects/CGMarket.h
Nordsoft91: it's loaded from config, here is just a default value

lib/mapObjects/CGMarket.h
Nordsoft91: actually, this one is even not needed, I believe since skills is an array

PR #2099, opened 2023-04-29
Fix loading skills from mods

Nordsoft91:
Actually quite important bug fix for modding, that's why in separate PR


IvanSavenko (approved):
Looks okay-ish since it solves a problem, but generally I think we should avoid any strings manipulations and instead use methods like getIdentifier.

Here however this is not trivial, will think how to design this better later.

PR #2100, opened 2023-04-29
Custom markets

Nordsoft91:
Duplicate of this: https://github.com/vcmi/vcmi/pull/2086

Just from another branch


config/objects/generic.json
```diff
@@ -549,7 +559,17 @@
...
+				},
+				"modes" : ["resource-skill"],
+				"title" : "core.genrltxt.602",
+				"speech" : "core.genrltxt.603",
+				"offer": 
+				[
+					{ "noneOf" : ["necromancy"] },
+					{ "noneOf" : ["necromancy"] },
+					{ "noneOf" : ["necromancy"] },
+					{ "noneOf" : ["necromancy"] }
```
IvanSavenko, line 571: I think this is hota change, not H3 mechanics.

Not against this change though - since this is always a dead slot:
- non-Necropolis heroes can't learn Necromancy due to having it banned for their class
- Necropolis heroes can't learn Necromancy here since they already have it as staring skills

lib/mapObjects/CommonConstructors.cpp
```diff
@@ -291,6 +291,69 @@ void BoatInstanceConstructor::configureObject(CGObjectInstance * object, CRandom
...
+	if(!market)
+		market = new CGMarket;
+	
+	preInitObject(market);
+
+	if(tmpl)
+		market->appearance = tmpl;
+	market->marketModes = marketModes;
+	if(marketEfficacy >= 0)
+		market->marketEfficacy = marketEfficacy;
```
IvanSavenko, line 336: Perhaps this should be obligatory member and set unconditionally?

lib/mapObjects/CommonConstructors.cpp
```diff
@@ -291,6 +291,69 @@ void BoatInstanceConstructor::configureObject(CGObjectInstance * object, CRandom
...
+	
+	market->title = market->getObjectName();
+	if(!title.empty())
+		market->title = VLC->generaltexth->translate(title);
```
IvanSavenko, line 340: Same here - make it obligatory?

IvanSavenko (approved):
Looks OK. Things that might be worth to look into:

1) Check any remaining references to market objects and remove them if possible:
- Obj::ALTAR_OF_SACRIFICE
- Obj::BLACK_MARKET
- Obj::TRADING_POST
- Obj::TRADING_POST_SNOW
- Obj::FREELANCERS_GUILD
- Obj::UNIVERSITY

2) Consider merging CGBlackMarket and perhaps even CGUniversity class into CGMarket class. Not sure whether this is needed/doable, so feel free to ignore

lib/mapObjects/CommonConstructors.cpp
Nordsoft91: no, it will be -1 if field is missing. Then 5 default value will be used

lib/mapObjects/CommonConstructors.cpp
Nordsoft91: There is existing logic that if field is missing, object name will be used

lib/mapObjects/CommonConstructors.cpp
IvanSavenko: Any need for this logic?

config/objects/generic.json
Nordsoft91: I took it from FizMig 2.5, but yes it includes hota. I'm not an expert in h3 mechanics, so can remove if so. As you mentioned, it's dead slot otherwise

rilian-la-te: @Nordsoft91 In FizMig hota changes should shown with hota icon. If this is shown with hota icon - it is hota change.

IvanSavenko: As I said - not against this change. If nobody else objects - go ahead.

lib/mapObjects/CommonConstructors.cpp
Nordsoft91: I think this parameter is not supposed to be used (almost). For all known objects value is 5. Then better to allow omitting parameter making it optional

IvanSavenko: But this is not how all our modding system is written.
We *always* specify all parameters. We have almost no conditional branches in objects loading code.
Even if all H3 objects use the same value we have it specified in all our objects

Nordsoft91: it's not true. Most of parameters in every loader are optional. Open any mod - obviously, modders specify only parameters they need (and it would be completely unfriendly to make the specify everything).

So it's obviously and user-oriented that some parameters must be optional. So we provide flexibility for modders to specify some rare property but please let's not make them specify parameter which they don't need. Such approaches remind me about freaky downloading web-portals, where you must watch 60 sec at advertisement before link appears))

IvanSavenko: However we always do that via zero-init - all such parameters are specifically designed to have false/0/empty string as "reasonable default".

Parameters where such zero-init is not a "reasonable default" always have explicit value in config.

Optional values often lead to an opposite effect - where modder forgets to put parameter that is needed for his object or where modder simply not aware of existence of such parameter (since making parameter as required would report it in validation)

Nordsoft91: Exactly, that's why reasonable default is -1 here, not 5

> Optional values often lead to an opposite effect

and that's totally fine if people aren't aware about something. Definitely better that awareness by force.
Can be solved in "soft" way - documentation, examples, and all-defined approach for embedded objects (described in config folder which can be used as starting point for mods)

IvanSavenko: If you insist on keeping this then at the very least please define all of this in one place, with one conditional check, e.g. in MarketConstuctor initializer.
Currently you have:
- marketConstructor initializing this property to -1 if not found in json
- marketConstructor assigning this property if above 0
- market itself having default value set to 5

Why not simply write something like
```c++
if (!json["efficiency"].isNull())
    efficiency = json["efficiency"].Integer()
else
    efficiency = 5;
```
Instead of spreading this through the code? Right now you have 3 different constants in 3 different locations when it coult be 1 constant in 1 location.

> Can be solved in "soft" way - documentation, examples, and all-defined approach for embedded objects (described in config folder which can be used as starting point for mods)

Modders must be aware of these documentation & examples. They are often not.
This documentation and examples must be maintained. They are often not.

I really like our validation system since it basically enforces everything to be up to date - it checks for any error made by us (such as not tracked field), for any error made by modder (such as missing field) and as result - always up to date. This is way more reliable than documentation & examples.

If I eventually get to my "modding tool" idea it will be based on our validator & will act as source of documentation & examples specifically because it is up to date by design.

Nordsoft91: understood, you are talking not about optional/mandatory parameter, but about initialzation out of loader

IvanSavenko: Kind of both.
Current approach is unnecessarily complex, with developer having to be aware of 3 places responsible for loading.
Reducing it to one conditional in one location is an improvement, but still (in my opinion) unnecessary complexity.

It is fine now since this is basically the only parameter that markets can have. But when you have 10-20 parameters having if-else block for each one makes code too hard to understand.

Nordsoft91: I suppose can be solved by overloading cast methods in JsonNode:
```c++
int JsonNode::Integer() const; //mandatory parameter, exception if missing
int JsonNode::Integer(int defaulValue) const; //optional parameter, exception-safe
```

IvanSavenko: Also an option, although right now we also have an advantage that default values are clear to modders - since missing values are always treated as false/0/empty string. 

Such change also means that we now need to show default value in docs & make sure that those default parameters are up to date with actual code.

config/objects/generic.json
```diff
@@ -72,12 +76,15 @@
...
+				},
+				"modes" : ["resource-resource", "resource-player"],
+				"efficiency" : 5,
+				"title" : "core.genrltxt.159"
 			}
 		}
 	},
 	"freelancersGuild" : {
-		"index" :213,
+		"index" :213, //deprecated
```
IvanSavenko, line 87: Not really deprecated since we need to load this object from h3m

Nordsoft91: That’s why I leave it

IvanSavenko: I meant - don't see any need to mark it as deprecated since this is part of h3m loader

Nordsoft91: disagree, but ok can remove

Nordsoft91: to point out - I was thinking to have it as hint for modders that they shouldn't use index fields for markets

IvanSavenko: That should not be a problem - "index" fields set in mods other than "core" should trigger validation warning and likely lead to a crash (since such object would conflict with our object).

So mods can't really use index field even if they attempt to

PR #2101, opened 2023-04-30
Part 2: refactoring for rewardable object

Nordsoft91:
Map instance is isolated from rewardable mechanic.

It unblocks swithing non-instance types to rewardable rails, which will be in 3rd part.

Only pieces of existing code were moved.

**This PR doesn't imply improvements for existing code, only moving parts of it**


lib/mapObjects/Rewardable.h
```diff
@@ -0,0 +1,350 @@
...
+			h & selectMode;
+			h & infoWindowType;
+		}
+	};
+
+	class DLL_LINKAGE Interface
+	{
+	private:
+		
+		Rewardable::Configuration _configuration;
```
IvanSavenko, line 322: 1. We don't use underscores for member names
2. Might be better to use created in consturctor unique_ptr and move interface class into a separate file to minimize includes. This header already includes quite a lot and this file is likely to be included in many other files in the future.

lib/mapObjects/Rewardable.h
```diff
@@ -0,0 +1,350 @@
...
+	template <typename Handler> void serialize(Handler &h, const int version)
+	{
+		h & limiter;
+		h & reward;
+		h & message;
+		h & visitType;
+	}
+};
+
+namespace Rewardable
```
IvanSavenko, line 257: Why only Configuration and Interface are part of namespace while other parts of rewardables (e.g. RewardInfo) are not?

IvanSavenko (approved):
Personally I would've split Rewardable.h into few more files and potentially moved all of them into a separate directory, e.g. lib/Rewardable/*

Also, ideally - includes in headers should be reserved for inheritance and in all other scenarios - use forward declarations & pointers instead of direct members. 
Includes in headers tend to multiply and kill build times since change to one header leads to rebuild of half of the codebase.

lib/mapObjects/Rewardable.h
Nordsoft91: as I understood, we cannot use forward declaration if serialization method is presented

IvanSavenko: We can, but you need to add `#include` in cpp that performs template instantiations (and only that file)

Can't remember which one, check compiler output to find what file triggered build error.

rilian-la-te: @IvanSavenko agree, we can. I splitting HeroBonus.cpp by your request now and faced same issues - `#include` in around 8 places fixed it.

PR #2105, opened 2023-04-30
Split HeroBonus.h to pieces: part1

rilian-la-te:
No functional changes, only refactoring, but huge BTW.


include/vcmi/spells/Caster.h
```diff
@@ -68,9 +69,9 @@ class DLL_LINKAGE Caster
 	virtual void spendMana(ServerCallback * server, const int32_t spellCost) const = 0;
 
 	virtual int32_t manaLimit() const = 0;
-	
-	///used to identify actual hero caster
-	virtual const CGHeroInstance * getHeroCaster() const = 0;
+
+	/// This caster ID should be used only in adventure spells (-1 if no adventure casting possible)
+	virtual int32_t getAdventureCasterID() const = 0;
```
IvanSavenko, line 74: - Why const CGHeroInstance * got replaced with int32_t?
- What this 'int32_t' means? ObjectInstanceID? HeroTypeID? Something else?

IMO preference for object identification should be:
- Pointer to corresponding object
- enumeration with index of corresponding object
- integer type
Unless there is strong reason not to use pointer/enum

lib/bonuses/HeroBonus.h
```diff
@@ -9,10 +9,8 @@
  */
 #pragma once
 
-#include "GameConstants.h"
-#include "JsonNode.h"
-#include "battle/BattleHex.h"
-#include <limits>
+#include "../GameConstants.h"
+#include "../JsonNode.h"
```
IvanSavenko, line 13: Only needed for BonusParams struct
Suggestion: move BonusParams to a separate file, remove include

lib/bonuses/IUpdater.h
IvanSavenko, line 1: Perhaps `Updaters.h`?
This file contains way more than Updater interface.

lib/bonuses/ILimiter.h
IvanSavenko, line 1: Perhaps `Limiters.h`?
This file contains way more than Limiter interface.

lib/bonuses/HeroBonus.h
IvanSavenko, line 1: Personally I would've tried to split this in way more parts:
1) Bonus enumeration. This is what used most often in code so having file with just enumerations might be a good idea. However enumeration is right now part of Bonus class. So perhaps also rename:
- - Bonus enumerations: `enum class Bonus` (to reduce changes in code)
- - Bonus instance class: `class BonusInstance`
2) Bonus class (potentially renaming to BonusInstance
3) IBonusBearer interface
4) CBonusSystemNode class
5) And probably move selectors into a separate file?

IvanSavenko (commented):
(assuming that content of HeroBonus.h/cpp haven't changed much other than moved into a different files)

include/vcmi/FactionMember.h
```diff
@@ -0,0 +1,69 @@
...
+VCMI_LIB_NAMESPACE_BEGIN
+
+class BonusList;
+
+namespace PrimarySkill
+{
+    enum PrimarySkill : int8_t;
+}
+
+class DLL_LINKAGE IFactionMember: public IConstBonusProvider, public INativeTerrainProvider
```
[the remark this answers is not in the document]
rilian-la-te: Faction member is abstract class, how I should call it?

lib/bonuses/HeroBonus.h
rilian-la-te: @IvanSavenko this is a draft, and I want to do 3,4 and 5 myself.

About selectors - not sure, it is used in many places.

include/vcmi/spells/Caster.h
rilian-la-te: @IvanSavenko this is some id, for hero - it is HeroInstanceID, but this a caster interface, and I do not think than it is good idea to use CGHeroInstance * in spells::Caster (which should not know about heroes or units at all).

include/vcmi/FactionMember.h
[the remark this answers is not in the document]
IvanSavenko: FactionMemberBase perhaps?

include/vcmi/spells/Caster.h
IvanSavenko: I get the idea, but for developer who uses caster interface this won't be as clear....

It also drops any type-safety - if somebody accidentally uses wrong type here compliler won't generate an error.

Perhaps add 
using AdventureMapCaster = CGHeroInstance
in caster interface?

Or make CGHeroInstance inherit from some minimalistic interface, e.g. `AdventureMapCaster`?

lib/bonuses/HeroBonus.h
IvanSavenko: Selectors - I'd rather split it, for logical separation in code.

But instead add convenience header that includes all commonly used parts of bonus system and include selectors into it.

include/vcmi/spells/Caster.h
rilian-la-te: Then we need to make BattleCasterID return CreatureID too, and caster will not free of assumption is there units of creatures. I think we should return an abstract ID in both cases and do not assume what ID is.

include/vcmi/FactionMember.h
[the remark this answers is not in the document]
rilian-la-te: IConstBonusProvider and INativeTerrianProvider is not a mixins, it is true interfaces. ICreature is a mixin too.

include/vcmi/spells/Caster.h
IvanSavenko: creating tiny interfaces (or perhaps even empty "tags") and making creature / hero inherit from them might be better approach then:
- caster class still has no knowledge of concrete types
- compiler can perform type checks since you can't dynamic_cast unrelated types

include/vcmi/FactionMember.h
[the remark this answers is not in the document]
IvanSavenko: ICreature - probably better to rename to CreatureBase as well?

include/vcmi/spells/Caster.h
rilian-la-te: @IvanSavenko my idea is we should not assume than caster is some concrete type, even tag. I want to do adventure spell refactoring, but not in this PR.

IvanSavenko: But you already making such assumption by returning ID to *some* object. Meaning, you imply that caster is a type that has some sort of index.

If we don't want to make such assumptions then we should not provide such method to begin with.

rilian-la-te: We can return ObjectInstanceID from getAdventureCasterID, but, I think, not more. And we should make sure than ObjectInstanceID has NONE.

IvanSavenko: Sounds good. And if ObjectInstanceID::NONE does not exists - then feel free to add as '-1'

lib/bonuses/ILimiter.h
rilian-la-te: done

lib/bonuses/IUpdater.h
rilian-la-te: done

lib/bonuses/HeroBonus.h
rilian-la-te: done

lib/bonuses/HeroBonus.h
rilian-la-te: @IvanSavenko about enumerations - you think than have 5 enum classes is better than current approach? If so, I will split.

include/vcmi/spells/Caster.h
rilian-la-te: done

include/vcmi/FactionMember.h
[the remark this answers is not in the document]
rilian-la-te: done

lib/bonuses/HeroBonus.h
IvanSavenko: By 5 enum classes you mean bonus type / duration / source?

Yeah, splitting is probably better than what we have now. I am huge supporter of "wrong code should not compile" approach, so enum classes with more strict checks are definitely better than simple enum's.

In addition to that, bonus type enum is the only part of current Bonus class that is extensively used through code - Bonus class itself and all others enum's are only used in few locations.

Not sure *how* to split it though. Bonus type certainly deserves its own separate file due to its size & active usage.
For other - hard to say. Not against 1 file per enum but might be too excessive here.

rilian-la-te: Yes, all enumeration values of bonuses.

Ok, will try to split.

rilian-la-te: @IvanSavenko can I split it in next PR? Enum split will be as huge as this PR by my first attempt.

IvanSavenko: Sure

rilian-la-te: then, let's split selectors and enumerators in next PR)

include/vcmi/FactionMember.h
```diff
@@ -0,0 +1,69 @@
...
+VCMI_LIB_NAMESPACE_BEGIN
+
+class BonusList;
+
+namespace PrimarySkill
+{
+    enum PrimarySkill : int8_t;
+}
+
+class DLL_LINKAGE FactionMemberBase: public IConstBonusProvider, public INativeTerrainProvider
```
Nordsoft91, line 24: No data in the class, then it's interface. You can name it in similar way IFactionMember

include/vcmi/spells/Caster.h
```diff
@@ -68,9 +69,9 @@ class DLL_LINKAGE Caster
 	virtual void spendMana(ServerCallback * server, const int32_t spellCost) const = 0;
 
 	virtual int32_t manaLimit() const = 0;
-	
-	///used to identify actual hero caster
-	virtual const CGHeroInstance * getHeroCaster() const = 0;
+
+	/// This caster ID should be used only in adventure spells (-1 if no adventure casting possible)
+	virtual ObjectInstanceID getAdventureCasterID() const = 0;
```
Nordsoft91, line 74: but why?
Returning pointer is much more useful and clear than indirect usage with some index with further searching object though map. Those indexes are quite buggy btw.

Also, how did you test that you won't break functionality, which was recently introduced?

Nordsoft91 (changes requested):
This change interferes with recently integrated functionality: https://github.com/vcmi/vcmi/pull/1928

I suggest to reduce interference and do not override fresh changes or at least perform full testing of new functionality at same level as it was done in my PR.

include/vcmi/FactionMember.h
rilian-la-te: @IvanSavenko this is my thought initially, what do you think?

include/vcmi/spells/Caster.h
rilian-la-te: @Nordsoft91 because caster should not know if there is a hero, unit or some other entity.

About testing - what should I test except adventure and battle spell casting of all spells (I did such testing without mods).

include/vcmi/FactionMember.h
IvanSavenko: This is not "interface" since it also provides implementation of its methods

include/vcmi/spells/Caster.h
rilian-la-te: I was not agree to #1938, just did not have rights to set RequestChanges) And this is my disagreement point BTW)

IvanSavenko: Agree with you. See here: https://github.com/vcmi/vcmi/pull/2105#discussion_r1181228702

include/vcmi/FactionMember.h
rilian-la-te: For me, this is not principial - I can name it both way if we all agree.

rilian-la-te: @IvanSavenko @Nordsoft91 which names are you prefer more for this non-data abstract classes? `XXXBase` or `IXXX`?

include/vcmi/spells/Caster.h
rilian-la-te: @Nordsoft91 Will do a testing with all spells with all skill levels if you want.

include/vcmi/FactionMember.h
IvanSavenko: Looking at our library code: `I` prefix is used only for classes that have all or almost all of their method as pure virtual.
Example: IBattleInfo / IBattleState
Abstact classes that have *some* of their methods implemented (and some are pure virtual) usually don't have `I` prefix - example: battle::Unit

Nordsoft91: @IvanSavenko , this approach I would suggest to change. Virtual or stub methods - something important? Class is mostly about data. If no data provided - then it's interface for me.

include/vcmi/spells/Caster.h
rilian-la-te: @Nordsoft91 if you absolutely think than that change should belongs to other PR - i can split and force-push.
But I do not think than such abomination like pointer inside caster interface should survive at least one release.

@IvanSavenko what do you think about getHeroCaster and my opinion strongly against it?

Nordsoft91: This is new functionality - basically, town gate advmap object shall work. See mod attached here https://github.com/vcmi/vcmi/pull/1928
Also, rewardable object instructions on wiki (in case mod is obsolete)

rilian-la-te: @Nordsoft91 will try. Maybe include something like it in unit test (but need think how to include)?

rilian-la-te: @Nordsoft91 do you have a test map for that mod?

include/vcmi/FactionMember.h
IvanSavenko: If we talk about "interface" as in 'interface' keyword in many languages then it should *not* provide implementation. In many languages it is actually illegal to have implementation inside interface.

For virtual/stub classes term "Abstract" is more common (but largely unused in vcmi codebase).

AbstactFactionMember / AFactionMember / FactionMemberBase are all reasonable names for me.

include/vcmi/spells/Caster.h
IvanSavenko: In my opinion preference should be:
1) Pointer to class.
2) if not possible - then object ID in enum form.
3) if all else fails - object ID in integer form.

- If we want to avoid unnecessary "knowledge" then we can use for example "tag" class that potential caster must inherit from.
- Pointers are generally more robust & result in more clear code (since you don't need constant VLC lookups to find referenced object)
- Pointers are also more flexible, especially when using "tag" / interface class - since said interface can be implemented by any other class.

rilian-la-te: @IvanSavenko but caster should not always be a hero, it can be an object (like Shipyard cast CreateBoat even without a hero involved). So, I think pointer to hero is not suitable for caster.

@Nordsoft91 your mod works, town portal can be cast by campfire, but console spams with ```Removal of object with spell casts is not supported!```. It is ok?

rilian-la-te: @Nordsoft91 used this form 
```json
					{
						"spellCast": {
							"spell" : "townPortal",
							"schoolLevel" : 3
						},
						"removeObject": true
					}
```

Nordsoft91: > I was not agree to https://github.com/vcmi/vcmi/pull/1928, just did not have rights to set RequestChanges) And this is my disagreement point BTW)

Sorry, but I didn't understood that you was not agree.
Don't care if feature works and users are happy, but your approach is very questionable


Your approach won't work for "virtual" casters, you obligate using objects from map only. 
Closest example  - town building.

include/vcmi/FactionMember.h
IvanSavenko: @Nordsoft91 
Thinking about it a bit more: this is abstract class. A textbook abstract class. Not an interface.

- Interface class (sometimes - protocol) - describes a set of method signatures, but does not implements them.
- Abstract class - provides incomplete/partial implementation of an interface and can no be instantiated directly.
- Concrete class - final class that provides complete functionality.

Zero objections against using *some* naming scheme for abstract classes, but it should be different from interfaces. Personally XXXBase is most clear from my point of view, but any other form such as `A` prefix (for consistency with interfaces) is OK for me, as long as there is distinction between interfaces naming scheme and abstract classes.

Nordsoft91: in the end, I believe it's up to @rilian-la-te to decide. All options I, A, Base suggested are ok, even initial one. Not really important

include/vcmi/spells/Caster.h
IvanSavenko: > @IvanSavenko but caster should not always be a hero, it can be an object (like Shipyard cast CreateBoat even without a hero involved). So, I think pointer to hero is not suitable for caster.

This is why I said about "tag class". Create an interface class, let's call it CasterBearer and make Shipyard / CGHeroInstance / etc inherit from it. With such class:
- it is impossible to misunderstand object ID / implicitly convert it into other ID class somewhere (we have a LOT of such type conversions)
- compiler will perform all checks for you while object ID can point to another object that can not be a caster
- it is possible to implement caster that won't actually be present on map. Town building is a great example here.
- developer also can quickly find which classes are valid as casters and which are not

rilian-la-te: @Nordsoft91 why? I thought town building will set an ID of parent town or of visiting/garrison hero. 

@IvanSavenko I think any classes which derived from spells::Caster should be able to cast a spell. But actually we can derive something like AdventureCaster, if you wish. 

Anyway, I think IDs is better than "tag classes", because IDs can be passed around as int and much easier to check.

@Nordsoft91 Another case why I was against getHeroCaster, because you reworked all battle checks to use getHeroCaster (example is #1964). I think getCasterUnitID should remain for battle casters. I can preserve getHeroCaster temporarily for adventure casters, if desired.

include/vcmi/FactionMember.h
rilian-la-te: Will use scheme with `AXXX`, will be appeared with next force-push.

include/vcmi/spells/Caster.h
Nordsoft91: >  I thought town building will set an ID of parent town or of visiting/garrison hero.

So town building cannot be casters in your solution. Only towns. 

> IDs can be passed around as int and much easier to check.

what do you mean easier to check? It's like same saying why not using reinterpret_cast<chat*> for everything.
vice-versa, we should avoid using ids, where possible. How you will debug ids? how you will identify whether it's valid or not? I remember several crashes with Ids, they are very painful to debug and fix

>  I think getCasterUnitID should remain for battle casters

What for? Can you give me an example of use case?

rilian-la-te: Adventure casters can be used with getHeroCaster, and there is a purpose of it. But getCasterUnitID was used  to return unitID or -1 (it which case we should assume than cast was perfrormed by hero or by mechanics like moat setup). You interfere this semantics in #1928, when you made CGHeroInstance return its hero ID instead of -1, and it introduced a several bugs which was fixed hackish way in #1964,  introducing getHeroCaster inside battle classes, it is what I want to avoid most (even more than CGHeroInstance* inside spells::Caster). In this PR I want to separate semantics again and make getBattleCasterID (which will work as getCasterUnitID pre-#1928) and other way to get adventure caster only.

nullkiller: Technically going from abstraction to exact implementation shows that initial abstraction leaks. Why do we need to know who casted a spell?

lib/spells/AdventureSpellMechanics.cpp
```diff
@@ -275,15 +280,15 @@ ESpellCastResult DimensionDoorMechanics::applyAdventureEffects(SpellCastEnvironm
 		env->complain("Destination is out of map!");
 		return ESpellCastResult::ERROR;
 	}
-	
-	if(!parameters.caster->getHeroCaster())
+
+	if(parameters.caster->getAdventureCasterID() < 0)
```
nullkiller, line 284: caster->isAdventureCaster() or even caster->canCastAdventureSpells()

nullkiller: When I think more about it I can not see a way to cast dimension door with anything except a hero. So looks like we need a way to retrive CGHeroInstance from a caster. So this check for adventure caster < 0 sounds weird. If we need a hero we should ask if it is a hero. Otherwise we need an abstraction which can answer how many caster have mana, movement points, sight center and so on.

nullkiller: Now it is not possible to make some other adventure entity able to cast dimension door because we expect hero here. And this abstraction with ID will not help

nullkiller: Why shipyard needs to cast an adventure spell? Can't it just build a boat? Shipyard mechanics is a bit different from boat summoning. One needs resources, another needs mana. Also shipyard never summons existing boat no matter if it exists or not. Mixing up such things may be dangerous. So still they have something in common. Some boat placement mechanics.
Agree, this is a complex question if we should use hero instance or some abstract id.

nullkiller: Anyway if you want a strong abstraction you should always talk to caster itself and never cast it to hero, object or whatever. If you do not want strong abstraction and a lot of proxy methods like getCasterMana, consumeCasterMana or whatever you may have a set of methods to get casterAsBattleUnit, casterAsHero or casterAsObject and so on. some kind of dynamic cast. You cast to required instance and work with it. IF it is null and required you report an error.

nullkiller: A very complex but interesting solution, Spell creates a spell mechanic factory and gives it to caster, caster calls createMechanic on this factory and pass this in it. Depending on who caster is factory has a few overloads for hero, object, battle unit. Factory creates mechanic which has all exact objects inside. IF mechanics needs to be able to  work with both hero and adv object it can have two subclasses with virtual hooks. For instance SummonBoatMechanicsFactory can return SummonBoatMechanics if called for hero and BuildBoatMechanics if called for shipyard.
So factory interface should know about all possible casters and in case if some caster is not supported it can crash or return nullptr. In last case methods should be named like tryCreateMechanics.
This will allow mechanics to work with exact types of casters or build local abstractions on them.
So this means local abstractions may get duplicated.

nullkiller: How about composite mechanics? We can decompose mechanics onto smaller parts like ManaConsumptionMechanics, MovementConsumptionMechanics, and so on. This will allow combining mechanics and even constructing composite spells like Meteor Shower with wall destruction. Depending on caster some mechanics can get replaced with others. This will remove duplication of mechanics code for factory solution

lib/spells/effects/Effects.cpp
```diff
@@ -112,7 +112,7 @@ Effects::EffectsToApply Effects::prepare(const Mechanics * m, const Target & aim
 		//todo: find a better way to handle such special cases
 
 		if(m->getSpellIndex() == SpellID::RESURRECTION && e->name == "cure")
-			applyThis = (m->caster->getHeroCaster() == nullptr);
+			applyThis = (m->caster->getBattleCasterID() >= 0);
```
nullkiller, line 115: Can we have both getBattleCasterID and getAdventureCasterID return something?

lib/spells/AdventureSpellMechanics.cpp
nullkiller: Ok, these were just some ideas, they are to complex to implement.

test/spells/effects/HealTest.cpp
```diff
@@ -372,7 +371,7 @@ TEST_P(HealApplyTest, Heals)
 
 	EXPECT_CALL(*battleFake, setUnitState(Eq(unitId), _, Gt(0))).Times(1);
 
-	EXPECT_CALL(actualCaster, getCasterUnitId()).WillRepeatedly(Return(-1));
+	EXPECT_CALL(actualCaster, getBattleCasterID()).WillRepeatedly(Return(unitId));
```
nullkiller, line 374: Interesting change. So this is a test when one stack heals another stack

lib/spells/effects/Timed.cpp
```diff
@@ -226,8 +226,8 @@ void Timed::convertBonus(const Mechanics * m, int32_t & duration, std::vector<Bo
 		if((nb.sid == SpellID::SHIELD || nb.sid == SpellID::AIR_SHIELD) && (nb.type == Bonus::GENERAL_DAMAGE_REDUCTION))
 			nb.val = 100 - nb.val;
 		//we need to know who cast Bind
-		else if(nb.sid == SpellID::BIND && nb.type == Bonus::BIND_EFFECT && m->caster->getHeroCaster() == nullptr)
-			nb.additionalInfo = m->caster->getCasterUnitId();
+		else if(nb.sid == SpellID::BIND && nb.type == Bonus::BIND_EFFECT && m->caster->getBattleCasterID() >= 0)
```
nullkiller, line 229: Is hero both adventure and battle caster? I would answer yes. Hero can cast during battle. But it is not a unit.
Detecting that caster is unit just because m->caster->getHeroCaster() == nullptr is also wrong. The one who is not a hero may still be not a unit. For now artifacts can cast so probably it is made in a way that the hero actually cast these spells. Moat can cast, ground can theoretically cast something in future. Obstacles? Binding trap. You step into it and it binds you for a few turns. So some getCasterType() which return HERO, UNIT, OBSTACLE and so on. And if you checked that caster is UNIT you can ask getCasterUnitID. Quite logical as for me

nullkiller: If we use getCasterType we actually can use getCasterID as we can understand what type this ID is by caster type. getCasterHero is also possible if we want to know whether caster is a hero. Kind of dynamic cast as I said before.

include/vcmi/spells/Caster.h
Nordsoft91: > But getCasterUnitID was used only in battle to return unitID or -1 (it which case we should assume than cast was perfrormed by hero or by mechanics like moat setup)

Literally, algorithm is following
```
if caster is not hero
  get unit id
```

before done implicitly 
```
if(getCasterUnitID() >= 0) //don't understand what is that? checking that caster is hero, obviously
  ...
```

now explicitly
```
if(getHeroCaster() == nullptr)
  ...
```

IvanSavenko: If this method is such a mess then why have it to begin with?
Perhaps remove it / replace with something else completely?

What are exact use-cases for this method?

nullkiller: > if caster is not hero
>  get unit id
Exactly. One has to figure out that there are only two types of caster unit and hero so if it is not hero - it is unit. And it breaks as soon as we have our initial assumption not true anymore, like moats casting or something.
So direct check if caster is unit then get unit id is way better

rilian-la-te: @IvanSavenko @nullkiller caster is not always a hero or unit. There is a case when caster is just a caster.

rilian-la-te: Like most setup)

nullkiller: So getCasterType() should work? I am afraid that people will think that getBattleCasterID() always return unit ID while its name is generic and it should be able to return something else. Also as I said hero is both battle and adventure caster so not clear if we should return something from both methods for hero.

IvanSavenko: > @IvanSavenko @nullkiller caster is not always a hero or unit. There is a case when caster is just a caster.

Again, what is the use-case of this method? Perhaps we can remove it or replace with something else?

For example, `bool isHeroCaster()` or `bool hasHero()` or something like that

rilian-la-te: 1. In some cases we should know exactly what unit cast a spell (eg. Bind) 
2. Check whether is unit casts a spell or a non-unit (mostly for logs). 
3. Get unit ID for logs. 

Maybe more, this method is popular.

IvanSavenko: - was there any need to change this method? If no - then I suggest revert at least until acceptable solution is found.
- is there any actual improvement in using object ID's over pointers? "Abstraction leaking" is a good argument, but replacing pointer with ID is not a fix for abstraction leak, but simply making it less obvious and more bug-prone.

rilian-la-te: @IvanSavenko Ok, I will move this commit into a separate PR.

But I fundamentally disagree with this approach. This is my opinion below:

1. Anything in include/vcmi should not use concrete types, not a single concrete type, except IDs and only when we should not convert an IDs to another IDs. So, no type knowledge at all except types already presented in include/vcmi.
2. #1928 should remain one and only one example of circumventing this principle, all other PRs should be blocked if they circumvent this principle. And after merge, this circumventing was lead to other problems (like #1964 and introduction of getHeroCaster inside battle types).

I want to fix it myself, but I do not understand why all other developers disagree.

rilian-la-te: Your controversal commit is moved outside of this PR.

lib/spells/AdventureSpellMechanics.cpp
rilian-la-te: Your controversal commit is moved outside of this PR.

lib/spells/effects/Timed.cpp
rilian-la-te: Your controversal commit is moved outside of this PR.

test/spells/effects/HealTest.cpp
rilian-la-te: It was always.

Your controversal commit is moved outside of this PR.

rilian-la-te:
@Nordsoft91 getHeroCaster is bad, but let's discuss it inside another PR.

include/vcmi/FactionMember.h
```diff
@@ -0,0 +1,69 @@
...
+	virtual int getDefense(bool ranged) const;
+	/**
+	 Returns primskill of creature or hero.
+	*/
+	int getPrimSkillLevel(PrimarySkill::PrimarySkill id) const;
+	/**
+	 Returns morale or luck of creature or hero.
+	*/
+	int MoraleVal() const; //range [-3, +3]
+	int LuckVal() const; //range [-3, +3]
```
IvanSavenko, line 59: Move "range [-3, +3]" comments into documentation comment above?

Also, perhaps give each method separate comment?

include/vcmi/FactionMember.h
```diff
@@ -0,0 +1,69 @@
...
+{
+    enum PrimarySkill : int8_t;
+}
+
+class DLL_LINKAGE AFactionMember: public IConstBonusProvider, public INativeTerrainProvider
+{
+public:
+	/**
+	 Returns native terrain considering some terrain bonuses.
+	*/
```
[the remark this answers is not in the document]
rilian-la-te: I am for `/* */` for all comments except commented code blocks. If it is principial for you - can change, comment style is not so principial for me. What do you think?

include/vcmi/FactionMember.h
rilian-la-te: It will be almost duplicate of comment from other val. I can, but do not see a reason. BTW, maybe add +5/-5 range setting in another PR?

rilian-la-te: done

lib/CCreatureHandler.cpp
```diff
@@ -71,7 +73,7 @@ const IBonusBearer * CCreature::getBonusBearer() const
 
 uint32_t CCreature::getMaxHealth() const
 {
-	return CBonusSystemNode::MaxHealth();
+	return MaxHealth();
```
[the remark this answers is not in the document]
rilian-la-te: done

lib/bonuses/HeroBonus.h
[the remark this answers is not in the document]
rilian-la-te: done

include/vcmi/FactionMember.h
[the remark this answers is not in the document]
IvanSavenko: Personally prefer `///` for documentation - since with such format comment is always on the line immediately before method (without empty line between method & comment).

Not insisting on change but I guess I wish we could decide on common style.

include/vcmi/FactionMember.h
IvanSavenko: > almost duplicate of comment from other val. 

True. Although not sure whether IDE's generate method tooltips (for autocomplete / hovering) for such shared comments. They definitely can show hints for individual comments though.

> maybe add +5/-5 range setting in another PR?

Why? H3 range is -3 ... +3.
If for purposes like modding - check game settings. This is where luck/morale chances are defined so we can try to connect these two - so if game settings define 5 options for positive morale we can return same range here.

IvanSavenko:
> @IvanSavenko we need to think how to fix getHeroCaster before 1.3. This commit moved outside of this PR.

Not against changes like this, but they should not harm / decrease readability of the codebase. Things to consider:
(keep in mind that I am not extremely familiar with spells system, so my assumptions might be wrong)
- Statement that CGHeroInstance in caster breaks abstraction level is a valid issue
- However replacing CGHeroInstance with ObjectInstanceID and especially - int32_t is only harmful since it decrease readability of code even further & allows potentially invalid state, e.g. if ObjectInstanceID gets ID of non-hero object while code assumes only hero as valid instance.
- If having CGHeroInstance inside Caster class breaks abstraction then this means that we either need another abstraction, e.g. CasterHolder interface implemented by hero, or that this method should be entirely removed from base caster class
- I personally dislike that `getCasterUnitId` return "bare" int32_t. What is this int32_t? Creature ID? StackInstance ID? Something else? If we don't have something StackInstanceID / UnitID then we definitely need one, if only for clarity.
- We could also consider possibility of implementing MapCaster / BattleCaster with current Caster acting as base class and only provide getHero() / getCaster() in such class.
- Or we could simply use dynamic_cast and cast Caster to a concrete class that has getHero() / getCaster()

Not sure which of these ideas make sense or are doable, but still better than replacing pointer with ID's

IvanSavenko (approved):
Looks OK assuming that changes in lib/HeroBonus -> lib/bonus are only about moving code around

include/vcmi/FactionMember.h
rilian-la-te: Range should be configurable btw, I am not sure if you setting made this possible. Ok, will check a setting.

IvanSavenko: It is semi-configurable now, only within -3...+3 range. Negative luck is disabled in based game via 0 range.

There is one issue with range that needs to be resolved somehow - UI.
H3 has def file with images for -3, -2 ... +2 +3 luck/morale.
Changing range to higher range than H3 requires some handling in UI too

rilian-la-te:
Will resolve conflicts later today. 

About unitID - we do not have a such thing as unitId (this is not a stack instance id or something like, it dynamically assigned when unit appeared in battle AFAIK) 

I thought about AdventureCaster/BattleCaster, but then proxy caster will implement both, or we will need 3 or 4 proxy casters to cover all needs. 

Casting directly will not work, because of many proxy casters (for example, AbilityCaster), which should return parent unit ID. 

BTW, we need unit ID only in bind case or for log lines. Maybe bind can be rewritten and then only logs will remain.

include/vcmi/FactionMember.h
rilian-la-te: Just show +3/-3 image. Anyway, it is not in scope of this PR, firstly we need to finish moving bonuses around.

IvanSavenko:
> About unitID - we do not have a such thing as unitId identifier (this is not a stack instance id or something like, it dynamically assigned when unit appeared in battle AFAIK)

While not critically important or urgent, having such UnitID is a good idea IMO. Another option is to eliminate *any* references to stack instance ID and always use pointer.

> BTW, we need unit ID only in bind case or for log lines. Maybe bind can be rewritten and then only logs will remain.

By log lines you mean player-visible combat log? Or our debug logging system? Debug logging can be simply removed. For battle log - perhaps add method like getCasterName() ?

rilian-la-te:
@IvanSavenko we cannot use pointer here, because unit ID is used in netpacks, AFAIK. 

Log lines - player-visible log. And we cannot use such simple thing like getCasterName there. Something we need to build entire lines, if we use unit (or not use unit).

include/vcmi/Creature.h
```diff
@@ -10,15 +10,29 @@
...
 class CreatureID;
 class ResourceSet;
 enum class EGameResID : int8_t;
 
-class DLL_LINKAGE Creature : public EntityWithNativeTerrain<CreatureID>
+/// Base class for creatures and battle stacks
+class DLL_LINKAGE ACreature: public AFactionMember
+{
+public:
+	bool isLiving() const; //non-undead, non-non living or alive
```
Nordsoft91, line 25: comment is complex, still don't get.
if it's about undead, suggest isUndead() with inverse usage.
Otherwise suggest isAlive()

include/vcmi/Creature.h
```diff
@@ -10,15 +10,29 @@
...
+	ui32 Speed(int turn = 0, bool useBind = false) const; //get speed (in moving tiles) of creature with all modificators
```
Nordsoft91, line 26: class methods should start from lower case: https://wiki.vcmi.eu/Coding_guidelines#Casing

> Local variables and methods start with a lowercase letter and use the camel casing

include/vcmi/FactionMember.h
```diff
@@ -0,0 +1,74 @@
...
+	 For now, uses range [-3, +3]
+	*/
+	int LuckVal() const;
+	/**
+	 Returns total value of all morale bonuses and sets bonusList as a pointer to the list of selected bonuses.
+	 @param bonusList is the out param it's list of all selected bonuses
+	 @return total value of all morale in the range [-3, +3] and 0 otherwise
+	*/
+	int MoraleValAndBonusList(std::shared_ptr<const BonusList> & bonusList) const;
+	int LuckValAndBonusList(std::shared_ptr<const BonusList> & bonusList) const;
```
Nordsoft91, line 71: also should start from lower case

include/vcmi/Creature.h
dydzio0614: there are also gargoyles and elementals

IvanSavenko: It is not about undead, but about any creatures that is neither:
- undead
- mechanical (e.g. War Machines)
- not "non-living", e.g. Golems
- and probably not a Gargoyle

In terms of game logic - this is a creature that is affected by spells like Death Ripple / Resurrection and such

include/vcmi/FactionMember.h
rilian-la-te: this was just moved from IBonusBearer, but I can change.

include/vcmi/Creature.h
rilian-la-te: this is also is just moved from IBonusBearer, but I can change.

rilian-la-te: done

include/vcmi/FactionMember.h
rilian-la-te: done

lib/bonuses/Bonus.cpp
```diff
@@ -0,0 +1,418 @@
+/*
+ * HeroBonus.cpp, part of VCMI engine
```
Nordsoft91, line 2:
```suggestion
 * Bonus.cpp, part of VCMI engine
```

lib/bonuses/IBonusBearer.h
```diff
@@ -0,0 +1,43 @@
+
+/*
+ * CBonusSystemNode.h, part of VCMI engine
```
Nordsoft91, line 3:
```suggestion
/*
 * IBonusBearer.h, part of VCMI engine
```

Nordsoft91 (approved):

lib/bonuses/Bonus.cpp
rilian-la-te: done

lib/bonuses/IBonusBearer.h
rilian-la-te: done

PR #2107, opened 2023-04-30
Part 3: Town buildings

Nordsoft91:
Depends on https://github.com/vcmi/vcmi/pull/2101

Final change, allows to use rewardable features in town buildings


config/objects/generic.json
```diff
@@ -72,7 +75,9 @@
 					"zoneLimit"	: 1,
 					"value"		: 100,
 					"rarity"	: 100
-				}
+				},
+				"modes" : ["resource-resource", "resource-player"],
+				"efficacy" : 5
```
rilian-la-te, line 80: Why "efficacy" and not "efficiency"? I think latter is more correct english)

Nordsoft91: it's not about english, both words are correct, but have different meanings.
It's better to explain on synonyms: if for markets you prefer "productivity" then "efficiency".
If "performance" or "impact" then "efficacy".

BTW, it appears in that PR https://github.com/vcmi/vcmi/pull/2100

IvanSavenko: Or use more direct meaning.
pricesFactor / costsFacts / exchangeRate / etc
To clearly indicate what this factor refers to

krs0: BuildingSubID::BROTHERHOOD_OF_SWORD:
		b = createBonus(building, Bonus::MORALE, +2);

This will be in jsons in the future?

krs0: And this part: "dwellingUpLvl1", BuildingID::DWELL_LVL_1_UP },

This is too rigid.

A creature building could be:
Base - normal creature building
Upgraded - upgraded creature building
Alternate - some other creature building to be chosen instead of base one.
Alternate upgrade - upgrade for the above
3rd upgrade - a second upgrade for the building,  HotA archers like.

Nordsoft91: > This will be in jsons in the future?

this change makes possible to configure special town buildings and map objects in the same way.
But I didn't rewrite legacy json

rilian-la-te:
@Nordsoft91 will this be possible for building do not use GiveBonus, but just propagate a bonus as before?

Nordsoft91:
@rilian-la-te , new type of building - yes, but legacy I kept untouched

rilian-la-te:
@Nordsoft91 we need buildings which should be able to propagate bonus to player, team or global effect, but not give its bonus to heroes. For example, lighthouse or some grail buildings. Maybe next time we will convert all resource buildings to bonuses too.

lib/mapObjects/CGTownBuilding.cpp
```diff
@@ -284,5 +285,220 @@ void CTownBonus::applyBonuses(CGHeroInstance * h, const BonusList & bonuses) con
...
+		case Rewardable::VISIT_BONUS:
+			return contextHero->hasBonusFrom(Bonus::TOWN_STRUCTURE, Bonus::getSid32(town->town->faction->getIndex(), bID));
+		case Rewardable::VISIT_HERO:
+			return visitors.find(contextHero->id) != visitors.end();
+		default:
+			return false;
+	}
+}
+
+void CTownRewardableBuilding::onHeroVisit(const CGHeroInstance *h) const
```
IvanSavenko, line 412: We might need some refactoring (either now or later) to minimize duplicated code in places like here.

lib/mapObjects/CGTownBuilding.cpp
```diff
@@ -284,5 +285,220 @@ void CTownBonus::applyBonuses(CGHeroInstance * h, const BonusList & bonuses) con
 		town->addHeroToStructureVisitors(h, indexOnTV);
 }
 
+CTownRewardableBuilding::CTownRewardableBuilding(const BuildingID & index, BuildingSubID::EBuildingSubID subId, CGTownInstance * cgTown, CRandomGenerator & rand)
```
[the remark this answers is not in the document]
rilian-la-te: Agree with second sentence. And we can remove old form altogether then.

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -487,72 +503,73 @@ void CGTownInstance::newTurn(CRandomGenerator & rand) const
...
 		}
-
-		const auto * manaVortex = getBonusingBuilding(BuildingSubID::MANA_VORTEX);
-
-		if (manaVortex != nullptr)
+		
+		for(const auto * manaVortex : getBonusingBuildings(BuildingSubID::MANA_VORTEX))
 			cb->setObjProperty(id, ObjProperty::STRUCTURE_CLEAR_VISITORS, manaVortex->indexOnTV); //reset visitors for Mana Vortex
-
+
```
rilian-la-te, line 509: here

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -487,72 +503,73 @@ void CGTownInstance::newTurn(CRandomGenerator & rand) const
...
 		//get Mana Vortex or Stables bonuses
 		//same code is in the CGameHandler::buildStructure method
 		if (visitingHero != nullptr)
 			cb->visitCastleObjects(this, visitingHero);
-
+		
 		if (garrisonHero != nullptr)
 			cb->visitCastleObjects(this, garrisonHero);
-
+
```
rilian-la-te, line 517: here

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -487,72 +503,73 @@ void CGTownInstance::newTurn(CRandomGenerator & rand) const
...
+					if (!count) // no dwelling
+						count = VLC->creh->objects[c]->getGrowth();
+					
+					{//no lower tiers or above current month
+						
+						if ((n = getSlotFor(c)).validSlot())
+						{
+							StackLocation sl(this, n);
+							if (slotEmpty(n))
+								cb->insertNewStack(sl, VLC->creh->objects[c], count);
```
rilian-la-te, line 561: can we use `VLC->creatures()->getByIndex(c)` here instead?

lib/mapObjects/CGTownBuilding.cpp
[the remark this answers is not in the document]
IvanSavenko: Removal can be postponed to 1.3 -> 1.4 period, to give modders time to adapt mods, but see no need to wait for H3 towns

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -487,72 +503,73 @@ void CGTownInstance::newTurn(CRandomGenerator & rand) const
...
-									cb->insertNewStack(sl, VLC->creh->objects[c], count);
-								else //add to existing
-									cb->changeStackCount(sl, count);
-							}
+					CreatureID c = town->creatures[i][0];
+					SlotID n;
+					
+					TQuantity count = creatureGrowth(i);
+					if (!count) // no dwelling
+						count = VLC->creh->objects[c]->getGrowth();
```
rilian-la-te, line 553: Here we absolutely can use `VLC->creatures()->getByIndex(c)->getGrowth()`. Can you use it instead of direct calling of VLC->creh?

lib/mapObjects/CGTownBuilding.cpp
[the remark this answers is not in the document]
dydzio0614: And to reduce hardcoded mechanics, like magic pond random resource generation stuff on 1st day?
Not sure if dwarven treasury is doable though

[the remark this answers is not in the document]
IvanSavenko: While I am all for unhardcoding stuff I think right now rewardable object (including one inside town) needs hero to visit it.
So Magic Pond is not convertible right now.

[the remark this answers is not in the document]
Nordsoft91: Maybe I can move non-visiting buildings to rewardable rails, but probably in another review.

As for visiting objects, will try to unhardcode them, but they require id and current logic is based that buildings with id cannot be configurable. Will see what can be done

lib/mapObjects/CGTownInstance.cpp
Nordsoft91: ok, but those added by xcode so I have to revert them line by line manually

lib/mapObjects/CGTownInstance.cpp
Nordsoft91: I suppose it's safe to do, but refactoring of CGTownInstance is out of scope. Change here is just dixing indents, see above code is the same

rilian-la-te: It is a new code, so, we do not need to use constructa than need refactoring in new code.

Nordsoft91:
@rilian-la-te , not sure that we should artificially restrict it. For parameter "showInInfobox" : true reward will be shown in infobox

rilian-la-te:
@Nordsoft91 infobox works only if adventure map is top interface. If top interface is a town, we should not use infobox. My InfoWindow redesign does not handle it, you need to handle it here.

lib/mapObjects/CGTownInstance.cpp
rilian-la-te: Okay, but I suggest to change it regardless, because then we will have less work for refactoring.

Nordsoft91:
@rilian-la-te 

https://user-images.githubusercontent.com/9308612/236318738-5ca4e45b-752f-42bd-96d8-7a9e52347a08.mp4

IvanSavenko:
Not insisting, but don't see reason for using infobox in towns. IMO we can disable it entirely and always use popups. Reasoning:
- You need sufficiently large resolution so town screen won't obscure infobox
- infobox is part of adventure map window and not present on town screen - it is out of player focus.
- after my advmap PR I think many players would use smaller resolutions (e.g. 1280x720) with upscaling, similar to how hd mod works. Meaning, there will be even less players who have infobox visible inside towns
- not sure whether it will work properly in case of multiple popups / multiple rewards - since adventure map does not receives update events infobox will keep displaying first popup

rilian-la-te:
@IvanSavenko I think info box should be used in towns if and only if it will support all options as adventure map. For now it is now, so, I agree with you.

Nordsoft91:
By default it's modal window so design decision I'll leave on modders shoulders

IvanSavenko:
> By default it's modal window so design decision I'll leave on modders shoulders

Where this default is set?

For map objects default is auto, meaning it will follow in-game setting, not modal window.
Right now if I have small resolution and toggled that game setting I will also end up with invisible rewards in town (because they are behind town screen.

Nordsoft91:
Ok, will test on small resolutions too

Nordsoft91:
Tested with 800x600 and 1920x1080. "showInInfoBox" parameter is absent.
Modal window appeared in both scenarios

IvanSavenko:
What happens if you:
1) Use 800x600 resolution
2) Have option "Show rewards in status window" (or what's its name?) enabled in settings?

Nordsoft91:
ok, found this option in game settings

Nordsoft91:
Fixes
https://github.com/vcmi/vcmi/issues/2133
https://github.com/vcmi/vcmi/issues/2110

server/CGameHandler.cpp
```diff
@@ -3612,7 +3612,7 @@ bool CGameHandler::buildStructure(ObjectInstanceID tid, BuildingID requestedID,
 	sendAndApply(&fw);
 
 	if(t->visitingHero)
-		visitCastleObjects(t, t->visitingHero);
+		objectVisited(t, t->visitingHero);
```
IvanSavenko, line 3615: Most likely - cause of regression #2150

Issue #1971, opened 2023-04-13
Client hangs after connecting to server on FreeBSD


danfe:
This issue is about [the bug](https://bugs.vcmi.eu/view.php?id=3146) reported earlier, to make it easier to track here on GitHub.

It happens with version 1.1.1 and apparently independent of the Boost version (originally reported for boost-libs 1.72.0, seeing it now with boost-libs 1.81.0).

Backtraces from _gdb(1)_ attached to both processes while it hangs:
```
(gdb@vcmiclient) bt
#0  0x0000000802e12b6c in ?? () from /lib/libthr.so.3
#1  0x0000000802e22f00 in ?? () from /lib/libthr.so.3
#2  0x0000000802e146cf in ?? () from /lib/libthr.so.3
#3  0x00000008007e9f5b in ?? () from /usr/local/lib/libboost_thread.so.1.81.0
#4  0x00000008007e8292 in boost::thread::join_noexcept() ()
   from /usr/local/lib/libboost_thread.so.1.81.0
#5  0x0000000800efc3ca in boost::thread::join (this=0x814ab3f18)
    at /usr/local/include/boost/thread/detail/thread.hpp:762
#6  CConnection::~CConnection (this=0x814f2cc18, vtt=<optimized out>)
    at /tmp/usr/ports/games/vcmi/work/vcmi-1.1.1/lib/serializer/Connection.cpp:238
#7  0x0000000800efc7f5 in CConnection::~CConnection (this=0x814f2cc18)
    at /tmp/usr/ports/games/vcmi/work/vcmi-1.1.1/lib/serializer/Connection.cpp:236
#8  0x00000000005f1621 in std::__1::__shared_count::__release_shared (
    this=0x814f2cc00) at /usr/include/c++/v1/__memory/shared_ptr.h:177
#9  std::__1::__shared_weak_count::__release_shared (this=0x814f2cc00)
    at /usr/include/c++/v1/__memory/shared_ptr.h:219
#10 std::__1::shared_ptr<CConnection>::~shared_ptr (this=<optimized out>)
    at /usr/include/c++/v1/__memory/shared_ptr.h:959
#11 std::__1::shared_ptr<CConnection>::reset (this=0x811107708)
    at /usr/include/c++/v1/__memory/shared_ptr.h:1054
#12 CServerHandler::resetStateForLobby (this=0x811107600, 
    mode=StartInfo::CAMPAIGN, names=0x0)
[...]

PR #2109, opened 2023-05-01
Fix potential nullptr-to-reference conversion

IvanSavenko:
(will cherry-pick to master after merge)

### DIFF
--- a/server/CVCMIServer.cpp
+++ b/server/CVCMIServer.cpp
@@ -375,10 +375,10 @@ class CVCMIServerPackVisitor : public VCMI_LIB_WRAP_NAMESPACE(ICPackVisitor)
 {
 private:
 	CVCMIServer & handler;
-	CGameHandler & gh;
+	std::shared_ptr<CGameHandler> gh;
 
 public:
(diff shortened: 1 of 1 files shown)


Nordsoft91:
@IvanSavenko , to master to beta?

IvanSavenko:
Not sure. Probably master?

There is no 'beta' branch right now and not sure if we're going to release 1.2.2.

Sure, can create beta branch just for this commit, but since this change is critical for some platforms and completely unused by others commit to master without release/tag might be best approach.

IvanSavenko:
@Alexander-Wilms Any preference on where to put this fix? What would be the preferred approach as far as flatpak packaging goes?

Alexander-Wilms:
@IvanSavenko It doesn't really matter, I can just add the patch file to the package.

IvanSavenko:
Understood.
In that case I am keeping this fix in develop branch only, flatpak/archlinux/etc can cherry-pick/copy patch if needed.

lemenkov:
Hello!
I've got the same issue as mentioned in #1971 but on Linux. I can confirm this PR does fixes #1971.

PR #2118, opened 2023-05-04
Use separate json for vcmi-1.3

IvanSavenko:
Updated vcmi-extras to use shortcuts, now battlefield actions submod from extras needs update that is not compatible with 1.2

### DIFF
--- a/config/schemas/settings.json
+++ b/config/schemas/settings.json
@@ -463,7 +463,7 @@
 				"repositoryURL" : {
 					"type" : "array",
 					"default" : [
-						"https://raw.githubusercontent.com/vcmi/vcmi-mods-repository/develop/vcmi-1.2.json"
+						"https://raw.githubusercontent.com/vcmi/vcmi-mods-repository/develop/vcmi-1.3.json"
 					],
 					"items" : {
 						"type" : "string"


Nordsoft91:
Still, my idea was a bit different:
- for legacy versions
  - vcmi-1.0.json
  - vcmi-1.1.json
- for latest release
  - github.json
- for future release/develop
  - beta.json

IvanSavenko:
> because we may want to have multiple repositories: 

But:
- we don't have multiple repositories
- I am not aware of any attempts to create any 2nd repository
- It does not resolves problem of supporting old versions without manual actions by players

IvanSavenko:
As for your proposition:
1) Game MUST "just work" for every player, irregardless of game versions. Players won't try to figure out why mod repo suddenly died and that they now need to switch it from githib.json to vcmi-1.2.json or something just because we made a new release. He will simply drop 1-star review on google play and forget about vcmi.

2) If we're going to provide multiple repositories (e.g. high quality mods / all mods / external mods) then we can keep using such system where "vcmi-1.x" will contain high quality mods and for example, "vcmi-extra-1.x" would contain all mods. Since again, game MUST just work for everyone.
Also, this raises a question of who is going to maintain that and decide on "mod quality" for all our mods?

Issue #2120, opened 2023-05-04
Add a Highlight for Ranged Full Damage Limit


krs0:
**Describe your proposal**
When in combat or tactics phase, is sometimes useful to see for your ranged creatures the range where they will do full damage.

Proposal below:

![image](https://user-images.githubusercontent.com/15194321/236308540-5898bfbc-9e34-4489-bf7b-b4021ae6d64a.png)

Latest screenshot
![image](https://github.com/vcmi/vcmi/assets/15194321/c105405b-fa19-4f11-88d6-44e527410a27)

**Additional context**
Archers max range limit is not covered in this issue but is enabled by it. https://github.com/vcmi/vcmi/issues/2122

PR #2121, opened 2023-05-04
Show Ranged Full Damage Limit Range

krs0:
This resolves https://github.com/vcmi/vcmi/issues/2120.

For ranged creatures I draw a jagged line to highlight the limit for full damage.
I calculate the full damage range
I calculate the full damage range limit
For each hex on limit I calculate where its outside! neighbors are and I draw a line for each of the edges in that direction.

Graphics need to be changed.


client/battle/BattleFieldController.cpp
```diff
@@ -333,6 +358,205 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesForMovementTarget(
...
+		// |      |
+		// 5      2
+		// |      |
+		//  \    /
+		//   4  3
+		//    \/
+		switch (mask)
+		{
+		// single edges
+		case 0b000001: // top left only
```
Nordsoft91, line 491: I would name your images accordingly: hexEdge000001.png, ...000010.png, 001110.png, etc

client/battle/BattleFieldController.cpp
```diff
@@ -333,6 +358,205 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesForMovementTarget(
...
+			highlightImage = fullDamageRangeLimitBottomRightCorner;
+            break;
+		case 0b111000: // bottom left corner
+			highlightImage = fullDamageRangeLimitBottomLeftCorner;
+            break;
+		case 0b100011: // top left corner
+			highlightImage = fullDamageRangeLimitTopLeftCorner;
+            break;
+        default:
+            break;
```
Nordsoft91, line 550: continue?

client/battle/BattleFieldController.cpp
krs0: I have chosen these long names for multiple edges, because I found it too difficult to differentiate only from binary mask :(.
If you are not strongly against I would keep verbose names.

IvanSavenko: Use `std::map<int, std::shared_ptr<IImage>` instead, with mask from here used as key.
Would also remove need for ~20 members in class.

client/battle/BattleFieldController.h
```diff
@@ -25,6 +25,18 @@ class Canvas;
 class IImage;
 class BattleInterface;
 
+enum ERelativePosition
```
IvanSavenko, line 28: Identical to BattleHex::Direction?

client/battle/BattleFieldController.cpp
IvanSavenko: Also - on naming:
IMO human-readable names are OK, but perhaps unify it a bit? You have images named 0.png ... 5.png alongside with human-readable names.

IvanSavenko (commented):
This probably needs some sort of option?
Not sure if we should have it enabled by default.

krs0:
> This probably needs some sort of option? Not sure if we should have it enabled by default.

Yes, just like movement, it will be enabled by an option.

client/battle/BattleFieldController.cpp
krs0: Not applicable, I switched to maps :)

client/battle/BattleFieldController.cpp
krs0: Implemented with maps

client/battle/BattleFieldController.h
krs0: Zombie code I missed on my own review :(. Deleted!

client/battle/BattleFieldController.cpp
```diff
@@ -333,6 +360,129 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesForMovementTarget(
...
+}
+
+// krs: This needs to be moved somewhere else
+unsigned char setBit(unsigned char mask, int position)
+{
+    if (position >= 0 && position < 8)
+        mask |= (1 << position);
+
+    return mask;
+}
```
Nordsoft91, line 462: just use std::bitset

rilian-la-te: @Nordsoft91 to use a bit set in serializable environments, we need to merge #1822

IvanSavenko: This option does not needs serialization, so bitset is OK (and probably makes sense here)

rilian-la-te: If does not need - then it can be used. But I think than #1822 will beerged before this.

krs0: changed to:
```
uint8_t setBit(std::bitset<6> mask, int position)
{
    if (position >= 0 && position < 6)
        mask.set(position);

    return static_cast<uint8_t>(mask.to_ulong());
}
```

IvanSavenko: You don't need this function at all. Use bitset as type of "mask" variable that you use with this function

krs0: but I need a byte here:
`output.push_back(fullDamageRangeLimitImages[mask]);`

krs0: ok changed to this:
```
	for(auto & directions : fullRangedDamageLimitHexesNeighbourDirections)
	{
		std::bitset<6> mask;
		
		// convert directions to mask
		for(auto direction : directions)
			mask.set(direction);

		uint8_t imageKey = static_cast<uint8_t>(mask.to_ulong());
		output.push_back(fullDamageRangeLimitImages[imageKey]);
	}
```

IvanSavenko: What I mean is that in `calculateFullRangedDamageHighlightImages` method you should change
`uint8_t mask = 0;`
to
`std::bitset<6> mask`

And call `mask.set(direction)` instead of current `mask = setBit(mask, direction);`

And once you need numeric value convert bitset to integer via 
`output.push_back(fullDamageRangeLimitImages[mask.to_ulong()]);`

IvanSavenko: Ah, yeah, correct :)

krs0:
Showing ranged full damage limit range is completed.

I do not fully understand the changes for fuzzy but the submodule looks ok.

krs0:
Closed. Duplicate created that solves merge submodules issues. https://github.com/vcmi/vcmi/pull/2156

PR #2123, opened 2023-05-04
Refactor enums and add zone monster strength "none" in rmg

Warzyw647:
- Split enum EMonsterStrength into EMonsterStrength and EZoneMonsterStrength.
- Removed obsolete ZONE_ and GLOBAL_ from enum item names.
- Renamed vector STRENGTH to zoneMonsterStrengths in CRmgTemplate based on the other enum serialization nearby (zone type).
- Simplified serialization of zone monster strength in preparation for adding zone monster strength: none.

Verified the strength of monsters generated on Jebus Cross and 6lm10a after compiling the changes


lib/rmg/CRmgTemplate.h
```diff
@@ -41,17 +41,24 @@ namespace EWaterContent
...
+	{
+		WEAK,
+		NORMAL,
+		STRONG
+	};
+}
+
 namespace EMonsterStrength
 {
 	enum EMonsterStrength
```
DjWarmonger, line 56: To be consistent, I'd rename this to `EGlobalMonsterStrength`.

lib/rmg/CRmgTemplate.h
```diff
@@ -41,17 +41,24 @@ namespace EWaterContent
 	};
 }
 
+namespace EZoneMonsterStrength
```
rilian-la-te, line 44: Why not an enum class?

lib/rmg/CRmgTemplate.h
Warzyw647: That was my first thought too, but it's also used on screens in game and in editor and in those contexts EZoneMonsterStrength doesn't appear, so I thought that maybe it's better to leave the old name. I mean, random map selection screen doesn't know other monster strengths than global, so it's fine with the old name. But I can easily rename all uses of EMonsterStrength to EGlobalMonsterStrength if you think that would look better.

lib/rmg/CRmgTemplate.h
Warzyw647: Because all enums around are also enums-in-namespaces and not enum classes, so I guess for consistency it might be better if they're either all enum classes or all enums-in-namespaces. And we use those enums as integers. For example global monster strength "strong" corresponds to number 4, which goes up by 1 in zones with "strong" monsters and that 5 is then used to fetch the correct coefficients from arrays in guard count calculations. Not sure how that part of code would look with enum classes. I'm referring to this (in \lib\rmg\ObjectManager.cpp, around line 400):
```
CGCreature * ObjectManager::chooseGuard(si32 strength, bool zoneGuard)
{
	//precalculate actual (randomized) monster strength based on this post
	//http://forum.vcmi.eu/viewtopic.php?p=12426#12426
	
	int mapMonsterStrength = map.getMapGenOptions().getMonsterStrength();
	int monsterStrength = (zoneGuard ? 0 : zone.monsterStrength - EZoneMonsterStrength::NORMAL) + mapMonsterStrength - 1; //array index from 0 to 4
	static const std::array<int, 5> value1{2500, 1500, 1000, 500, 0};
	static const std::array<int, 5> value2{7500, 7500, 7500, 5000, 5000};
	static const std::array<float, 5> multiplier1{0.5, 0.75, 1.0, 1.5, 1.5};
	static const std::array<float, 5> multiplier2{0.5, 0.75, 1.0, 1.0, 1.5};
	
	int strength1 = static_cast<int>(std::max(0.f, (strength - value1.at(monsterStrength)) * multiplier1.at(monsterStrength)));
	int strength2 = static_cast<int>(std::max(0.f, (strength - value2.at(monsterStrength)) * multiplier2.at(monsterStrength)));
	
	strength = strength1 + strength2;
```

rilian-la-te: @IvanSavenko what do you think?

IvanSavenko: Don't know?

Yes, enum class is preferable, except for cases when enum is often used as integer type, making enum class too inconvenient.
Whether these use-cases here qualify as "often used" - not sure. Not familiar enough with RMG code.

Warzyw647: These 2 enums appear as numbers in monster generation formula and the global one is also shown to the player as an int in the map description of the generated random map.
EZoneMonsterStrength has 2 uses: parsing the `"monsters" : "normal", //"weak" "strong"` part of the rmg template json written by the template makers and then using that to increase or decrease monster strength in that zone.
However, all the other enums in rmg don't seem to have strong ties to ints, but I will fix that in a different PR as such fix wouldn't match this PR or branch name.

lib/rmg/CRmgTemplate.h
Warzyw647: Ok, the name EMonsterStrength does feel wrong when there's a EZoneMonsterStrength nearby. I'll rename in a moment as you suggested.

lib/rmg/CRmgTemplate.h
Warzyw647: Correction: scenario description in a generated random map in game shows `monsters strong`, not `monsters 4`. 'monsters 4' is what HotA and SoD shows.

lib/rmg/CRmgTemplate.h
Warzyw647: Done.

DjWarmonger (approved):

Nordsoft91:
I'll test today

lib/rmg/CRmgTemplate.h
```diff
@@ -41,17 +41,24 @@ namespace EWaterContent
 	};
 }
 
-namespace EMonsterStrength
+namespace EZoneMonsterStrength // used as int in monster generation procedure
```
Nordsoft91, line 44: can we have only one enum?

Nordsoft91 (changes requested):
Bug discovered while testing (regression in compare with develop):
1. open RMG tab
2. select "random" button in monster strength group
3. observe it's not selectable

lib/rmg/CRmgTemplate.h
Warzyw647: I'll change RANDOM = -1 back to RANDOM = -2 in EGlobalMonsterStrength, update the branch with recent changes from develop and see if that solves the problem. And if not, I'll investigate further.
As for one enum, the thing is that EZoneMonsterStrength::RANDOM wouldn't make sense while EGlobalMonsterStrength::RANDOM does make sense and the upcoming EZoneMonsterStrength::NONE will make sense (Hota rmg has that option) while EGlobalMonsterStrength::NONE wouldn't make sense.
And it would be nice if adding parsing of "monsters": "none" to ZoneOptions::serializeJson in lib/rmg/CRmgTemplate.cpp didn't make the code more complicated.

Nordsoft91: But would be possible to have single enum with both values?
RANDOM, NONE, WEAK, NORMAL, STRONG

Than change formula and that's it?

Warzyw647: I've changed my mind. I do want both zoneMonsterStrength "random" (randomly chosen for each monster) and perhaps even globalMonsterStrength "none" (though not sure if anyone would be using that), so I'll be going back to 1 enum and implementing both none and random functionality for zones and maybe also the none global functionality. Though not sure about the last one.
I'm also wondering if global monster strength: "random" should mean: "pick one strength at random at map generation start" or "pick one at random every time you make a monster on a map". I mean, it used to work like the first one, but not sure if anyone was using it and the second one is an interesting option I think. Hmm.. would be too chaotic with the none options, so I think the "random" options shouldn't pick the "none" options.

krs0: To pick random strength every time you put a monster, this would make random maps even more, luck based. Do not think such an option is needed at the moment. (other opinions?). Random at map start looks ok to me.

Warzyw647:
Not yet, thanks for pointing that out. I expect the fixes to be ready somewhere around weekend.

Warzyw647:
Let's try this again, this time with the target feature "zone monster strength: none". From my limited testing Maps and Saves seem to be compatible with the current state of the "develop" branch. Though I only checked generating a few maps in game and in map editor, loading a few maps/saves from one version to the other, clicking some buttons for rmg options in map editor and in game (such as water content: none, normal, random and islands) and checked that setting monster strength to "none" in some template's zone does generate unguarded treasures in that zone.

DjWarmonger:
Prompt: Please evaluate this merge request and provide feedback. Merge request url: https://github.com/vcmi/vcmi/pull/2123/files

Answer (ChatGP 4 + web browsing):
Based on the diff provided, here are my observations:

1. The namespace `EPlayerType` and `ETemplateZoneType` have been refactored to use `enum class` instead. This is a good change as `enum class` provides better type safety and allows the enumeration to be scoped, which can prevent naming collisions.

2. The function signature for `setPlayerTypeForStandardPlayer` and a few others have been updated to reflect the change in `EPlayerType`. The changes look correct and consistent with the refactor.

3. `monsterStrength` was refactored in `ZoneOptions` in `lib/rmg/CRmgTemplate.cpp` for better clarity and to handle a broader range of strengths. This includes handling for `ZONE_NONE`, `ZONE_WEAK`, `ZONE_NORMAL`, and `ZONE_STRONG`.

4. However, it appears the changes in `lib/rmg/CMapGenerator.cpp` are incomplete as the diff only shows a partial line of code. This might be an error in the diff or the changes might not be complete yet.

Please note that without knowing the specific goals of these changes and the overall context of the project, it's hard to provide specific advice on whether these changes are beneficial or not. For example, the change in handling `monsterStrength` could be very beneficial if the project needed to handle a broader range of monster strengths, or it could be unnecessary if such functionality isn't needed. The same applies to the refactoring of `EPlayerType` and `ETemplateZoneType` to `enum class` - this could be very beneficial for preventing naming collisions and improving type safety, or it could be unnecessary if these benefits aren't needed in the project.

Also, it's important to ensure that these changes are reflected throughout the entire code base. For example, any other code that relies on `EPlayerType` or `ETemplateZoneType` being a namespace would need to be updated to reflect the change to `enum class`. Similarly, any code that uses `monsterStrength` would need to be updated to reflect the new handling of `monsterStrength`.

In terms of coding standards, the changes appear to be well formatted and follow good practices, such as using `enum class` for type safety and scoping, and providing handling for a range of values for `monsterStrength`.

DjWarmonger (approved):

DjWarmonger:
Works for me.

lib/rmg/TreasurePlacer.cpp
```diff
@@ -527,8 +527,8 @@ size_t TreasurePlacer::getPossibleObjectsSize() const
 }
 
 bool TreasurePlacer::isGuardNeededForTreasure(int value)
-{
-	return zone.getType() != ETemplateZoneType::WATER && value > minGuardedValue;
+{// no guard in a zone with "monsters: none", in a water zone and for small treasures
+	return zone.monsterStrength != EMonsterStrength::ZONE_NONE && zone.getType() != ETemplateZoneType::WATER && value > minGuardedValue;
```
Nordsoft91, line 531: since that, can we assign monsterStrength = ZONE_NONE for water zone when it's generated?
It would help to get rid of this hardcoded guard check

lib/rmg/CRmgTemplate.cpp
```diff
@@ -378,24 +378,31 @@ void ZoneOptions::serializeJson(JsonSerializeFormat & handler)
...
-		if(handler.saving)
-		{
-			rawStrength = static_cast<decltype(rawStrength)>(zoneMonsterStrength);
-			rawStrength++;
-		}
-		handler.serializeEnum("monsters", rawStrength, EMonsterStrength::ZONE_NORMAL + 1, STRENGTH);
-		if(!handler.saving)
+		int temporaryZoneMonsterStrengthIndex = monsterStrength == EMonsterStrength::ZONE_NONE ? 0 : monsterStrength - EMonsterStrength::ZONE_WEAK + 1 ; // temporary until serializeEnum starts supporting std::map
+		// temporaryZoneMonsterStrengthIndex = 0, 1, 2 and 3 for monsterStrength = ZONE_NONE, ZONE_WEAK, ZONE_NORMAL and ZONE_STRONG respectively
+		handler.serializeEnum("monsters", temporaryZoneMonsterStrengthIndex, 2, zoneMonsterStrengths); // default is normal monsters
```
Nordsoft91, line 391: am I correctly understand, that before default value was monsters_strong?

Warzyw647: I just checked on develop and no, the default there is also zone_normal.
I checked that by changing monster strength on players' zones in JC: I gave Blue weak monsters, Teal normal, Green strong and for Red I removed the monster strength field from the template.
Green (strong) got guarded Sawmills.
Teal (normal) got unguarded Sawmills and guarded Campfires.
Blue (weak) got unguarded Sawmills and Campfires.
Red (default) got unguarded Sawmills and guarded Campfires, same as Teal, so default was normal.
The logs did contain the generator's complaint about the missing 'monsters' field, but it did generate the maps.

lib/rmg/TreasurePlacer.cpp
Warzyw647: Good point. I'll check the code that generates the extra water zones not explicitly specified in the template json and set monster strength to zone_none there instead of this hardcoded check.
While at it, I'll also try to figure out why the generator crashed when I specified a template zone's type as "water" in the template json. I did that to the central zone on JC and removed the towns from there and it crashed on develop. I mean, there are no actual templates that use zone type: water, but both on develop and on this branch the generator doesn't complain about a wrong value.

DjWarmonger: Its enought that RMG is trying to spawn any objects which don't have a water template. In particular Mines, but also about anything without HoTA.

Warzyw647: Did as you suggested, Nordsoft, and also updated the branch with recent changes from develop.

IvanSavenko:
@Nordsoft91 @Warzyw647 
Ready for merge? 
Or is there anything else that needs to be addressed here?

Warzyw647:
I think it's ready for merge, unless Nordsoft91 finds a bug or wants more time for testing.

I don't encounter that bug with non-selectable random monsters anymore after the last few commits.
When I put "monsters" : "none" in a template's zone, then that zone generates with treasures and without monsters, but the zone guards do generate and other monster settings in zone and monster strength choices do generate zones with stronger or weaker guards according to the settings. Options for water content also result in different water contents as intended.
So, I don't see any unintended behavior and I see the effects of the intended change (adding "monsters" : "none").

DjWarmonger:
Let me check this.

DjWarmonger:
I clicked monster strength -> random, then generated few maps. Scenariom descriptions shows different monster strengths and certainly monsters show up on the map.

Issue #2127, opened 2023-05-05
When playing H3 Complete Mac demo, hasAnimations assertion fails for some enemies


ghost:
**Describe the bug**
I tried to fight some end-game Dungeon mobs and the game crashed.

**Game logs**
```
vcmiclient: /var/tmp/portage/games-strategy/vcmi-1.2.1/work/vcmi-1.2.1/client/battle/BattleInterface.cpp:773: void BattleInterface::checkForAnimations(): Assertion `!hasAnimations()' failed.
```

**To Reproduce**
1. Get the Mac demo working, see https://github.com/vcmi/vcmi/issues/1198#issuecomment-1536257694
2. Install the Mac demo mod
3. Play the demo with intent to actually finish it, but not to quick so that the Dungeon player can purchase end-game units. I'm not sure which one, but one of them causes this crash.
4. Engage the enemy which has the erroneous creature.

**Expected behavior**
The battle should begin normally.

**Actual behavior**
The game crashes.

**Did it work earlier?**
No idea.

**Version**
 - OS: Gentoo Linux
 - Version: 1.2.1

GeorgeK1ng:
VCMI can be used with merged demo files created using the [Heroes of Might and Magic III Demo Builder for VCMI](https://github.com/GeorgeK1ng/heroes3demo), along with the [Heroes III Demo](https://github.com/vcmi-mods/demo-support) and [VCMI Extras](https://github.com/vcmi-mods/vcmi-extras) mods enabled.
 
No bugs or issues – just enjoy limited gameplay with Castle, Dungeon, and Necropolis! There are no further plans for expanding demo support directly to VCMI.

Issue #2126, opened 2023-05-05
Demo version support


ghost:
**Describe your proposal**
I would like to be able to play the demo version of HoMM3 with VCMI.

**Additional context**
The demo can be downloaded from here https://www.gamefront.com/files/heroes-of-might-magic-iii-the-restoration-of-erathia-demo-english/download
The file we're looking for is `h3demo.exe` with size of `98265058` bytes and sha256 sum `83cbb6319386d6deba65273f53a60395fda8cce74cc24177b2ac3560ded9c315`.
To get from the downloaded file to HoMM3 gamedata, 2 extractions need to be made:
1. Extract data1.cab and data1.hdr from the installer: `7z e h3demo.exe /Disk1/data1.cab /Disk1/data1.hdr`
2. Extract gamedata from data1.cab:
```
unshield x data1.cab H3demo.h3m StrongHold.mp3 'Surrender Battle.mp3' 'Win Scenario.mp3' 'Retreat Battle.mp3' UltimateLose.mp3 LoseCombat.mp3 'Defend Castle.mp3' 'Win Battle.mp3' LoseCastle.mp3HEROES3.SND H3BITMAP.LOD H3SPRITE.LOD VIDEO.VID
```
3. Your `./Program_Files` directory is now a suitable demo asset directory. VCMI should be able to pick this up if you rename it to let's say `~/.local/share/vcmi`
Once this is done, the game should start.
However, 1.2.1 does not start and this is what happens:
```
Starting... 
VCMI 1.2.1 (client)
Creating console and configuring logger: 0 ms                                                                                                                                                                                                
The log file will be saved to "/home/kerneliusz/.cache/vcmi/VCMI_Client_log.txt"                                                                                                                                                                   
        Initialization: 2 ms                                                                                                                                                                                                                 
        Data loading: 13 ms                                                                                                                                                                                                                  
        Mod handler: 0 ms                                                                                                                                                                                                                    
        Mod filesystems: 11 ms                                                                                                                                                                                                               
Initialized logging system based on settings successfully.                                                                                                                                                                                   
[log level] ai => not set                                                                                                                                                                                                                    
[log level] animation => not set                                                                                                                                                                                                             
[log level] bonus => not set                                                                                                                                                                                                                 
[log level] global => trace                                                                                                                                                                                                                  
[log level] mod => not set                                                                                                                                                                                                                   
[log level] network => not set                                                                                                                                                                                                               
[...]

PR #1871, opened 2023-04-04
Hota H3M parsing support (1.3)

IvanSavenko:
Implements *parsing* of HotA map format. Support for these options or gameplay support for these objects is out of scope of this PR.

Translation: HotA maps will *not* be playable or editable with map editor yet with this PR.

### Current progress:
- [x] Refactoring of existing H3M reader to make changes less of a pain
- [x] Parsing of HotA map header 
- [x] Parsing of HotA map objects
- [x] Parsing of HotA Seer Hut format
- [x] All HotA maps can be parsed by VCMI
- [x] All HotA campaigns can be parsed by VCMI
- [x] Investigate new mission types in Seer Huts
- [x] Investigate multiple quests / repetitive quests in the seer hut
- [x] Investigate creature banks options

### Outside of PR scope:
- Support for new options added by HotA
- Loading of new objects functionality from VCMI HotA mod and connecting it to h3m map objects
- Support for HotA maps in scenario selection menu
- Support for HotA campaigns

### List of discovered HotA format changes compared to SoD:
- New victory condition: "Eliminate all monsters"
- New victory condition: "Survive for certain time"
- Map option: "Allow monster/plague weeks"
- Map option: "Combat round limit"
- Map Event option: "Allow human player to activate this event" - unconfirmed, but should be supported
- Wandering Monster option: Precise setup for disposition
- Wandering Monster option: Percentage of group that will join
- Wandering Monster option: Monster only joins for money
- Wandering Monster option: Number of stacks in combat
- Wandering Monster option: Presence of upgraded stack in combat
- Creature Banks option: Non-random guards present selection
- Creature Banks option: Presence of upgraded stack in combat
- Creature Banks option: List of artifacts to receive as reward
- Seer Hut new quest type: 'Be one of the following hero classes'
- Seer Hut new quest type: 'Days passed'
- Seer Hut: Support for multiple quests
- Seer Hut: Support for repeatable quests
- Town option: Allow spell research in town

PR #2130, opened 2023-05-05
Support data from RoE/AB/Demo

Nordsoft91:
There will be special mod with necessary tweaks


client/render/CBitmapHandler.cpp
```diff
@@ -103,21 +104,40 @@ SDL_Surface * BitmapHandler::loadH3PCX(ui8 * pcx, size_t size)
 	return ret;
 }
 
+ResourceID loadBitmapResource(const std::string & path, const std::string & fname)
```
IvanSavenko, line 107: Technically we already have such functionality - see CMappedFileLoader class.
Was used in original version of wog mod, probably unused right now

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -236,10 +272,11 @@ void FirstLaunchView::heroesDataDetected()
 	ui->pushButtonDataNext->setEnabled(true);
 
 	heroesLanguageUpdate();
+	problemWithData = false;
 }
 
 // Tab Heroes III Data
-bool FirstLaunchView::heroesDataDetect()
+std::pair<bool, bool> FirstLaunchView::heroesDataDetect()
```
IvanSavenko, line 279: Perhaps use std::tuple<> instead and read via structured binding?
```c++
auto [roeFound, sodFound] = heroesDataDetect()
```

lib/CArtHandler.cpp
```diff
@@ -301,22 +303,42 @@ CArtHandler::~CArtHandler() = default;
...
+		for(auto & c : slotName)
+		{
+			c = std::toupper(c);
+			if(c == ' ')
+				c = '_';
+		}
+		
+		artSlotAllowed.insert(slotName);
+		slotName.erase(std::remove(slotName.begin(), slotName.end(), '_'), slotName.end()); //no spaces and underscores
+		slotName = std::regex_replace(slotName, std::regex("WARMACHINE"), "MACH"); //special case
```
IvanSavenko, line 338: why regex? We always use boost:replace_first or boost::replace_all in such cases

lib/CModHandler.cpp
```diff
@@ -352,7 +352,7 @@ void CIdentifierStorage::finalize()
 		}
 		logMod->error("All known identifiers were dumped into log file");
 	}
-	assert(errorsFound == false);
+
```
IvanSavenko, line 355: Why change?

IvanSavenko (commented):
> There will be special mod with necessary tweaks

Why mod? Perhaps integrate into vcmi directly? Players want working game and not to figure out which mod they need to install for game to work

client/render/CBitmapHandler.cpp
Nordsoft91: for some reason doesn't work.
If I add code below, other files from data folder become invisible
```
"filesystem" :
	{
		"DATA/" :
		[
			{"type" : "map", "path" : "Data/HeroScr4.pix", "initial" : "Data/HeroScr3.pix"},
		]
	}
```

lib/CArtHandler.cpp
Nordsoft91: I prefer std. Actually, I was thinking that we could get rid of boost dependency at one day

IvanSavenko: In that case should be done more consistently instead of spreading multiple different methods through code.

Also have concerns about regex performance compared to simple string replacement

IvanSavenko: Don't see any complains about regex perfomance, so presumably acceptable.

If we're replacing boost::replace_first / boost::replace_all then I suggest to
- create vstd::replace_first / replace_first methods that will internally use std::regex
- replace all instances of boost usages with vstd

But please, our text processing is already a mess with massive mix of std::string operations, boost format, boost replace and who knows what else. No need to add regex to this pile.

lib/CModHandler.cpp
Nordsoft91: failed here if some objects are missing on original data

lib/CArtHandler.cpp
kambala-decapitator: rather create regex overload. For simple text replacement you can use implementation from cppref: https://en.cppreference.com/w/cpp/string/basic_string/replace (in Example)

lib/CModHandler.cpp
IvanSavenko: Then fix?
This is almost guaranteed to result in crash down the line. What the point of broken demo support?

lib/CArtHandler.cpp
IvanSavenko: Yeah, regex-free implementation with only std::string will probably be the best option.

That being said, my main concern is to avoid increasing mess in our already messy text manipulation code. If we want to replace boost then I suggest making small wrapper in vstd to cover all our typical use-cases and replace boost with that wrapper. But not introducing yet another way to manipulate strings.

client/render/CBitmapHandler.cpp
IvanSavenko: Is hero screen the only file that needs such replacement?
In that case I'd rather use "targeted" fix and all check to hero window.

Also, RoE (and maybe AB) don't have 5th MISC slot, which is probably why RoE has different hero screen image. Probably needs to be handled somehow?

lib/CArtHandler.cpp
kambala-decapitator: > I suggest making small wrapper in vstd to cover all our typical use-cases and replace boost with that wrapper

exactly my thoughts

Nordsoft91: guys, it's short, clear and single line of code. it's not bug-prone and doesn't require refactoring. It doesn't use any thirdparty and is part of standard.

Please, explain why alternative solutions suggested are better and so important to use in vcmi to solve substring replacement problem.

IvanSavenko: Consistency?
It is already tiring to figure out how each formatting operation works in vcmi.
"Oh, we use boost format here? Okay, then we need to wrap this in boost::str and use '%' operator to format stuff, and replaced string will be available as return type"
"Oh, but here we use boost replace, so replacement is done in place, and we need to specify string to be replaced"
"Oh, here we use simple concatenation of multiple std::strings, so simply use operator +"
And now also:
"Oh and here we use std regex so ...  where will be result and how to provide parameters?"

Please, VCMI codebase is already a mess and pretty much every area needs refactoring. Please, don't make this even more messy.

Right now we (mostly) use boost::replace_first. So please, keep using it or refactor code to use better routine.

IvanSavenko:
Also, my main question about this PR:

Why mod? Perhaps integrate into vcmi directly? Players want working game and not to figure out which mod they need to install for game to work

client/render/CBitmapHandler.cpp
Nordsoft91: Ideally would be to rewrite heroWindow to make it configurable. But for now I'm trying to **launch** game, not adopt everything

IvanSavenko: In that case maybe keep this as draft PR until demo support is properly integrated with minimal hacks in codebase?

Nordsoft91: but why this is a hack and what "properly" means?

IvanSavenko: Instead of "tricking" vcmi into thinking that data matches sod we should:

- have this integrated into base game. Meaning, vcmi should be usable with demo files as it, without any additional mods or any other manual configuration by player. 

- we should also either disable features added in AB/SoD or allow them to be properly used with RoE data. For example, MISC_5 slot should be either disabled when RoE data is used or visible in all UI elements (e.g. hero screen, exchange screen, etc).

- for things like hero screen background we should use some mechanism that will:
- 1) detect data type provided by player (roe demo, sod demo, full roe, full ab, full sod, full complete)
- 2) provide information specific to data type, for example - which file should be used for hero screen background, how many artifact slots are present, how many entries are in text files, etc.
- 3) this information can be either hardcoded or loaded from text config, but it should be in a base game and not in some mod that players have to find.
(this can be done for example similarly to how map format "feature levels" are done)

lib/CArtHandler.cpp
Nordsoft91: ok, I see there is no a single usage of regex in codebase. Refactoring is definitely out of this review

client/render/CBitmapHandler.cpp
Nordsoft91: I want to say here that I want to make resource handling aside of actual engine.
Ideally, game shouldn't know anything about presence of some picture. Everything should be done on loading side and hacks are allowed in mod itself.
We have similar functionality with jsons for animations, but not for images.

lib/CModHandler.cpp
Nordsoft91: nope, game operates fine. The problem is that if you have some null identifiers, it gives such issue.
e.g try to have line in mod and you will have same error
`"core:vialOfDragonBlood": null,`

client/render/CBitmapHandler.cpp
IvanSavenko: > Ideally, game shouldn't know anything about presence of some picture

True, what we could do is to have something like this in our json's
```json
    "featureLevelROE" : { "heroScreenBackgroundName" : "HeroScr3.pcx", "hasMisc5Slot" : false },
    "featureLevelSOD" : { "heroScreenBackgroundName" : "HeroScr4.pcx", "hasMisc5Slot" : true }
```
And based on game data detection, vcmi should use either "featureLevelROE" or "featureLevelSOD" entry.

> We have similar functionality with jsons for animations, but not for images.

Thing is, I don't see need for such functionality for images. Reasoning:
1) In this case what we need is filesystem-abstraction level (or window configuration level), not image format level. Since we need to adapt to different data layouts, not to different format of image data
2) Reasoning behind json animations was not filesystem abstraction, but an alternative to binary h3 .def format. Since we needed to provide support for images sequences similar to one provided by .def files. Alternative to binary h3 .pcx images are bmp/png's, not json

Nordsoft91:
@IvanSavenko, I borrowed idea from mac-demo-support mod.
Unfortunately, it doesn't work in 1.2 so I decided to investigate why and what is needed to make it work. Then idea was to try on all kind of data (RoE, AB, demo) and disable features which are not present.
Better to make if work form the box, but comparing efforts needed and related risks, mod is ok.

IvanSavenko:
> Then idea was to try on all kind of data (RoE, AB, demo) and disable features which are not present. Better to make if work form the box, but comparing efforts needed and related risks, mod is ok.

IMO we should try to go for integrated version.
Yes, this means more efforts than mod version but:
- it will likely be more stable than mod version since it will be maintained as part of codebase and not as semi-abandoned mod.
- it will likely to be more preferred by players, especially if we find way to automatically install demo via launcher, similar to how fheroes2 does
- we'll likely need more mods for multiple data versions (roe demo, sod demo, roe full). Which means more mods to maintains with likely duplicated content.

lib/CModHandler.cpp
IvanSavenko: What exactly lead to this error?
Normally, this is due to unfulfilled identifier request, meaning that somewhere in code instead of writing vial of dragonblood ID we kept uninitialized field, that will lead to crash when it is accessed. Or is this not the case here?

Nordsoft91:
1) questionable assumption taking into account amount of regression we have in develop and amount of regression we have after almost every refactoring
2) can be done with mods too (and here I'm ready to make necessary development)
3) mods are ready. Amount of testing needed in both options is same

Nordsoft91:
I understand pros and cons and completely fine if someone would reimplement this feature in different way in future

IvanSavenko:
Why not now?
Why we must use approach that both requires bunch of workarounds in code and completely not obvious to players?

And this is when approach that will not rely on hacks and will be completely transparent to player is straightforward?

IvanSavenko:
Sorry for complaining but 10 years of workarounds and "nah, let's make this properly later" thoughts lead to current unmaintainable mess of text called "vcmi codebase" and basically zero active players.

Maybe it is time to start doing things properly? It is not like we have a deadline to release beta support yesterday.

Nordsoft91:
Because I disagree with that. I have my own vision about how features shall look, what priorities are. Those may not match yours. I'm glad to share, but without imposing it. That's I believe open source development is about.

So, beside the lyrics:
 - feature offered as is. I observed demand on it.
 - I'm ok to fix code-related problems, do testing, consider alternative approaches if clear reasoning provided
 - I'm not ok with hollowing scope and directive management "do that, don't do that"
 - if there are no code-related objections, I consider work done here
 - I can admit your rights not accepting this or any PR. Would be nice to have reason, e.g breaks existing features, code duplication, low quality code, etc. - better than any opinion different from yours is incorrect

IvanSavenko:
My view:
1) this is NOT what user expects. This feature *has* to work by default, integrated into base game. Because this is what our one and only target audience (players) expects. Anything other than that simply won't be used. Sorry, this is not 1990 where players are fine with figuring out how hardware and software works. If feature does not works by default it might as well just not exist. We already have vcmi-extras that are not "extras" but effectively essentials. No need to add new "demo-extras".
2) This change offers only partial support even if only H3 features are considered. I don't see any handling of (for example)
- - missing files, e.g. TentColor.txt, 
- - different features (e.g. lack of MISC_5 slot), 
- - missing RMG dialog
(list is very likely incomplete)
3) Same goes to VCMI features, for example launcher - pretty much every mod will not work with demo files, meaning we need to either disable mod repo completely or decide on a way to handle this case better.
4) This *is* poor quality code in my opinion:
- - image json "format" that is not actually image format but filesystem "link". Such "links" should work on filesystem level, like that old (and possibly broken) json mapping FS.
- - removal of identifiers check assertion is not a proper handling of an error. This is developer saying "I don't care about incorrect input"

Nordsoft91:
1. Ok, that's possible to enable feature by default - by adding my configs into vcmi-essentials for example, or adding into extras or somehow else
2. Some points may not be available yet, which is fine. Anyway It would be unfair if you are merging PRs with features implemented partially (examples [1](https://github.com/vcmi/vcmi/pull/2078), [2](https://github.com/vcmi/vcmi/pull/2005), [3](https://github.com/vcmi/vcmi/pull/1871)) while forcing others to do everything at single PR.
3. Works with this set of mods
Снимок экрана 2023-05-06 в 21 57 50
4. Ok, but
  - how it differs from existing one void CAnimation::initFromJson(const JsonNode & config) ?
  - if incorrect input leads to correct behavior, then input is correct. Moreover, you are dissembling by saying that it makes code poor, taking into account that we had few examples with assert violation **merged** into develop or even not compiling. And I'm not saying about regression bugs we dealt right **now** in "modern-cool-refactored" develop for stuff worked well earlier. So we definitely have different understanding what code quality is.
But I can take a look deeply what is root cause if it's the only blocker. Also, would be interesting to check 1.0 and 1.1, maybe some versions back such warnings even wouldn't appear and then it's regression bug too.

IvanSavenko:
1. Then let's do this. We'll also need automatic detection of game data version and disable/enable appropriate mods.
2. Kind of unfair that you've selected just my PR's. I can find bunch of similar PR's by others (that I've accepted). Key difference is that in all cases me / PR author is going to continue work on this feature and split PR's to simplyfy workflow. Meanwhile in this PR you said, quote: "I understand pros and cons and completely fine if someone would reimplement this feature in different way in future", indicating that you have no desire to address these issues.
3. What about mods that alter H3 objects? Define "works fine"? Mods likely reuse some H3 content - e.g. creature sounds, maybe some graphical assets. For example I know that some towns don't bother with siege screen and copy existing one which does not exists in demo. What happens if you disable extras and try RMG?
4. 
- - CAnimation json format exists specifically to replace information stored in .def file header, to allow (for example) 32-bit graphics in animations. This format can NOT be replaced with filesystem link.
- - "if incorrect input leads to correct behavior, then input is correct" - Are you sure? Can you explain what exactly causes that error and what exactly code has requested that identifier? Either you have not triggered code that uses that uninitialized data or error reporting generated false positive and should be fixed in place. Not by removing error checking.
- - Well, if you wish - feel free to test every single PR in detail before merge. If you wish to take over PR reviews and give me more time for development - then I am all for it. But apparently nobody is interested in this. I already spend (IMO) far too much time on PR reviews and feedback. Specifically to reduce chance of various bugs and regressions. If you can do better job - feel free to take over.

Nordsoft91:
> Then let's do this

Agree it's needed ultimately. But as I said - out of scope for this PR.

> Kind of unfair that you've selected just my PR's

Maybe I understand you wrong, but here is what I read: "if you do feature A, please do features B, C and D, because A is less valuable with B, C and D". I selected your PRs to show that you don't follow this approach too. But it doesn't mean that I consider those PRs bad or something (vice versa, I totally agree - let's eat elephant step by step).

> "I understand pros and cons and completely fine if someone would reimplement this feature in different way in future", indicating that you have no desire to address these issues.

I have on desire to rewrite all legacy parsers to make no-mod embedded solution as you suggested. But please don't make conclusion that I'm not going to evolve mods-based solution

> Mods likely reuse some H3 content - e.g. creature sounds, maybe some graphical assets

From which point we have requirement to keep compatibility with all mods? I bet there are bunch of mods broken without my PR. Again, some features may not work with demo. Some maps won't start with demo. And it's fine

> What happens if you disable extras and try RMG?

My mod has strict dependency for extras

> This format can NOT be replaced with filesystem link.

As I said approach suggested by you is simply doesn't work. Since it doesn't work and my works, I need to understand why my approach is SO bad before spending time to fix legacy one.

> Can you explain what exactly causes that error and what exactly code has requested that identifier?

As I said in previous message, "But I can take a look deeply what is root cause if it's the only blocker"

> Well, if you wish - feel free to test every single PR in detail before merge

I'm not blaming you for those bugs or saying you are doing bad something. On the contrary, I appreciate your efforts.

I'm not blaming you or someone else in particular but sometimes I feel like double standards
 - huge risk-prone refactoring, impacting 200 files: "nice! let's merge!"
 - @Nordsoft91 tiny isolated feature: "oh no! your approach of string replacement is disaster!"
(maybe this is only my subjective perception and only my problem and have nothing with reality)

IvanSavenko:
> Agree it's needed ultimately. But as I said - out of scope for this PR.

If you're planing this at some point before 1.3 - then yes, sure. You often say "out of scope" but show no indication as to what you mean by that:
- Will be done in next PR soon?
- May be done at some point whenever you start working on this code again?
- Not something you're planning to do at all?

> I have on desire to rewrite all legacy parsers to make no-mod embedded solution as you suggested. But please don't make conclusion that I'm not going to evolve mods-based solution

Change to parsers is trivial IMO. Full integration might also be better since I am somewhat concerned about addition of submods to our "vcmi" mod - due to how players often update vcmi by simple overwrite of old version with new one, if at any point we remove preinstalled submods, they will remain in installation directory of players.
Meanwhile, similar files in config directory will simply stay unused and won't be automatically picked during directory scan like submods do.

Best approach might be loading our json's conditionally, e.g. only load "config/faction/conflux.json" if sod data has been detected.

> From which point we have requirement to keep compatibility with all mods? I bet there are bunch of mods broken without my PR. Again, some features may not work with demo. Some maps won't start with demo. And it's fine

Fine for who? For you - maybe. For a player who wants to play with demo? Definitely no. Especially since "won't start" means "crash".

Yes, checking every single mod for compatibility with demo is pointless waste of time, but we should provide support for that. For example add field to mod.json to indicate required data set (demo/roe/sod). And by default assume that all mods require sod & hide / block it if player uses demo files.

And if modder incorrectly set "demo" for mod that requires sod - then this is issue with his mod, not vcmi. But we do need to add support for such check.

Maps (as in, premade h3m maps) as well as campaigns should be completely disabled in demo, since demo can't load any "real-world" map except for one it shipped with.

> My mod has strict dependency for extras

So "extras" now not "extras" but a strict dependency. Great...

> As I said approach suggested by you is simply doesn't work. Since it doesn't work and my works, I need to understand why my approach is SO bad before spending time to fix legacy one.

Well, if you don't want to fix that old filesystem (understandable, I was thinking about completely removing it anyway) then what about changing code of hero window to take account of potentially missing file? We likely also need a way to handle 5th slot, so changes in code of hero window will likely be needed anyway.

Still better than this (effectively) workaround that was made specifically for demo and will never be used by anything other than demo.

> I'm not blaming you or someone else in particular but sometimes I feel like double standards
>     * huge risk-prone refactoring, impacting 200 files: "nice! let's merge!"
>     * @Nordsoft91 tiny isolated feature: "oh no! your approach of string replacement is disaster!"
>       (maybe this is only my subjective perception and only my problem and have nothing with reality)

Subjective perception I think? If it feels like that - then sorry.
- If I see changes like that (even minor) I mention it in review in every PR, not just yours. 
- I do tend to skip things that I view as pointless/unnecessary, as long as I don't view them as negative change. 
- Accepted massive PR's are often rather safe since most of changes are simple renames. And often have way more review "conversations" than here.
- Artifacts-related regressions are likely because nobody knows that code (or does not wants to review those PR's) apart from PR's author, so regressions can easily slip since I only do basic checks.

Nordsoft91:
Scope is not about timeline or assignee. It's about problem to address.

PR about data support, then it is not about removing boost stuff or about automatic data downloading or anything else 
Similar to have a comment "we have 50+ more bugs, please fix all of them too" in a single bugfix review. That what I call out-of-scope.

- "won't start" means "crash" - **wrong statement**
- demo can't load any "real-world" map except for one it shipped with - **wrong statement**
- Best approach might be loading our json's conditionally - **disagree**
- Maps (as in, premade h3m maps) as well as campaigns should be completely disabled in demo - **totally disagree**

There is fundamental disagreement in development approach, blocking further work on feature

krs0:
_Best approach might be loading our json's conditionally - disagree_

If I think of SoD as mod, this sounds the right way to do it. Do not load something that is not needed.
Probably this will end up in `If Demo`, in original SoD mod (or original H3 mod)

krs0:
But thinking about the main purpose of this change...
I understand that it is to easily load VCMI with original H3 data.
So what is the real benefit of supporting RoE or AB? How much usage will this get?
So proposal will be to focus on Demo and SoD and RoE and AB to be retrofitted via mods.

Nordsoft91:
btw, there is a bug created https://github.com/vcmi/vcmi/issues/2127

but @IvanSavenko as a product management expert insisting this is negative-value change

> this is NOT what user expects. This feature has to work by default, integrated into base game. Because this is what our one and only target audience (players) expects. Anything other than that simply won't be used. Sorry, this is not 1990 where players are fine with figuring out how hardware and software works. If feature does not works by default it might as well just not exist. We already have vcmi-extras that are not "extras" but effectively essentials. No need to add new "demo-extras".

Nordsoft91:
one more! https://github.com/vcmi/vcmi/issues/2126

krs0:
@Nordsoft91 I understand your objections and your frustration, and it needs to be addressed, but lets do it in a constructive manner.

PR #2142, opened 2023-05-09
Fix broken bonuses of new artifacts

### LINKED ISSUE #1920: [Android]Artifact has no effect!
Describe the bug

Android, new est daily builds ([VCMI-branch-develop-148cc0b.apk)
The Artifact purchased from Artifact Merchants has no actual effect after equipping the hero. (It must happen)
Sometimes, there is a probability that the Artifact obtained will not produce any effect (probabilistic)

Game logs
Not applicable

To Reproduce

Choose Tower, build Artifact Merchants, and then purchase an Artifact that will no longer have any effect.
The starter comes with an Artifact that can work normally

Did it work earlier?
It seems that it worked properly in previous versions

Screenshots
IMG_20230409_210836
IMG_20230409_213529

Version

OS: Android MIUI 14.0.3
Version: VCMI-branch-develop-148cc0b
Additional context

### LINKED ISSUE #2128: Artifacts does not display its bonuses
**Describe the bug**
Self-explanatory

**Game logs**
Not applicable

**To Reproduce**
1. use vcmiforgeofnoldorking
2. observe stats

**Expected behavior**
Stats should be changed

**Actual behavior**
It is not

**Did it work earlier?**
Yes, before 1.3

**Screenshots**
image

**Version**
 - OS: Linux
 - Version: develop

**Additional context**

IvanSavenko:
Simple workaround since I am not familiar with this code, but should work fine.
If @SoundSSGood or somebody else can propose better fix - go ahead.

PutArtifact can only be sent after NewArtifact. Othervice bonus system does not seems to be working correctly for unknown reason.

- Fixes #1920 
- Fixes #2128

### DIFF
--- a/lib/NetPacksLib.cpp
+++ b/lib/NetPacksLib.cpp
@@ -1793,6 +1793,9 @@ void BulkSmartRebalanceStacks::applyGs(CGameState * gs)
 void PutArtifact::applyGs(CGameState *gs)
 {
 	assert(art->canBePutAt(al));
+	// Ensure that artifact has been correctly added via NewArtifact pack
+	assert(vstd::contains(gs->map->artInstances, art));
+	assert(!art->getParentNodes().empty());
 	art->putAt(al);
 	//al.hero->putArtifact(al.slot, art);
 }
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -6793,8 +6793,21 @@ void CGameHandler::putArtifact(const ArtifactLocation &al, const CArtifactInstan
 bool CGameHandler::giveHeroNewArtifact(const CGHeroInstance * h, const CArtifact * artType, ArtifactPosition pos)
 {
 	assert(artType);
-	if(pos != ArtifactPosition::FIRST_AVAILABLE && !ArtifactUtils::isSlotBackpack(pos))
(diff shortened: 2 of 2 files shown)


SoundSSGood:
> If @SoundSSGood or somebody else can propose better fix - go ahead.

Works for me.
In my opinion at the moment, this looks like the best solution that can be proposed.

> Othervice bonus system does not seems to be working correctly for unknown reason.

since I'm not familiar with the bonus system code, I also don't know the reason. Maybe I'll work on the bonus system when I'm done with the artifacts.

rilian-la-te:
It is not bonus system, it is a netpacks system most likely. Maybe CBonusSystemNode interaction of this system.

IvanSavenko:
Yeah, haven't checked that but most likely serializer issue:
NewArtifact writes entire artifact into pack and then adds artifact into global registry
PutArtifact uses that registry to find existing artifact instance & on success - writes only artifact unique instance ID to netpack

But since here game ended up sending PutArtifact with not yet registered object, both packs got separate artifact instances:
- PutArtifact instance, that was added to hero, but not initialized with artifact type
- NewArtifact instance, that was initialized with artifact type, but was not added to the hero.

PR #2151, opened 2023-05-12
Fix regressions from adventure map PR

IvanSavenko:
Fixes so far:
- Fix broken toggle button callbacks in settings window
- Fix "Duplicate widget name" error on opening settings window


krs0:
This should be merged since many options were just not working.

IvanSavenko:
Sure, merging then

PR #2153, opened 2023-05-13
Separate updates from rendering actions

IvanSavenko:
Minor refactoring, goal is to remove time-counting code from show() method (and derived) and move it into a separate (already existing) method tick(uint32_t).
Reasoning:
- makes logical sense - since these two actions (update based on elapsed time and rendering) are not related to each other
- fixes minor bugs where background object would sometime play animation due to time updates in show / showAll
- allows const show/showAll methods, eliminating potential side-effects from this call

Progress:
- [x] move FramerateManager into separate file
- [x] remove access to SDL from FramerateManager and use chrono instead
- [x] move time-counting code to separate method in adventure map
- [x] move time-counting code to separate method in castle interface
- [x] move time-counting code to separate method in battles

Other changes:
- FPS manager now operates using high-resolution clock, potentially allowing more precise timings
- Framerate counting now uses sliding window using last 60 frames instead of once-per-second update

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -16,6 +16,7 @@
 #include "mainmenu/CMainMenu.h"
 #include "mainmenu/CPrologEpilogVideo.h"
 #include "gui/CursorHandler.h"
+#include "gui/FramerateManager.h"
 #include "CPlayerInterface.h"
 #include "CVideoHandler.h"
 #include "CMusicHandler.h"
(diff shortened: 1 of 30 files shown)


rilian-la-te:
Infobar inside dragon utopias works?

IvanSavenko:
Should work.
Infobar already uses tick() method in develop (and probably in 1.2), so not affected by my PR

PR #2155, opened 2023-05-14
Highlight Adventure Map Object command

krs0:
Usage `showObject <objectID>`
EG: `show windmill`
image


krs0:
For the moment it works with objects that have names in `object->typeName`

krs0:
This will not progress from this state, until nice object names support is available. (too hard of a change for me).
If not mergeable, I still need this PR for my personal tests.

PR #2156, opened 2023-05-14
Highlight Ranged Full Damage Limit

krs0:
This resolves https://github.com/vcmi/vcmi/issues/2120.

For ranged creatures I draw a jagged line to highlight the limit for full damage.
I calculate the full damage range
I calculate the full damage range limit
For each hex on limit I calculate where its outside! neighbors are and I draw a line for each of the edges in that direction.

image

Images could be updated later according to feedback.


client/battle/BattleFieldController.cpp
```diff
@@ -338,6 +365,124 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesForMovementTarget(
...
+	if(bonus != nullptr && bonus->additionalInfo != CAddInfo::NONE)
+		rangedFullDamageDistance = bonus->additionalInfo[0];
+
+	// get only battlefield hexes that are in full range damage distance
+	std::set<BattleHex> fullRangeLimit; 
+	for(auto & hex : battleFieldWithoutSideColumns)
+	{
+		if(BattleHex::getDistance(hoveredHex, hex) <= rangedFullDamageDistance)
+			rangedFullDamageHexes.push_back(hex);
+	}
```
IvanSavenko, line 403: You can use `isAvailable` instead of generating `battleFieldWithoutSideColumns` vector
```suggestion
	for(int i = 0; i < GameConstants::BFIELD_SIZE; i++)
	{
		BattleHex hex(i);
		if(hex.isAvailable() && BattleHex::getDistance(hoveredHex, hex) <= rangedFullDamageDistance)
			rangedFullDamageHexes.push_back(hex);
	}

```

client/battle/BattleFieldController.cpp
```diff
@@ -338,6 +365,124 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesForMovementTarget(
...
+	const CStack * hoveredStack = owner.curInt->cb->battleGetStackByPos(hoveredHex, true);
+	if(!(hoveredStack && hoveredStack->isShooter()))
+		return rangedFullDamageLimitHexes;
+
+	auto rangedFullDamageDistance = GameConstants::BATTLE_PENALTY_DISTANCE;
+
+	// overwrite full ranged damage distance from Additional info field of LIMITED_SHOOTING_RANGE bonus
+	auto bonus = hoveredStack->getBonus(Selector::type()(BonusType::LIMITED_SHOOTING_RANGE));
+	if(bonus != nullptr && bonus->additionalInfo != CAddInfo::NONE)
+		rangedFullDamageDistance = bonus->additionalInfo[0];
```
IvanSavenko, line 423: Perhaps create `rangedPenaltyDistance` method in CStack?

client/battle/BattleFieldController.cpp
```diff
@@ -338,6 +365,124 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesForMovementTarget(
...
+	for(auto & hex : battleFieldWithoutSideColumns)
+	{
+		if(BattleHex::getDistance(hoveredHex, hex) <= rangedFullDamageDistance)
+			rangedFullDamageHexes.push_back(hex);
+	}
+
+	return rangedFullDamageHexes;
+}
+
+std::vector<BattleHex> BattleFieldController::getRangedFullDamageLimitHexes(std::vector<BattleHex> rangedFullDamageHexes)
```
IvanSavenko, line 408: Perhaps merge this method with getRangedFullDamageHexes?

Most of this method is copy of getRangedFullDamageHexes. The only diffence is final block.

client/battle/BattleFieldController.cpp
```diff
@@ -338,6 +365,124 @@ std::set<BattleHex> BattleFieldController::getHighlightedHexesForMovementTarget(
...
+	for(auto & hex : rangedFullDamageHexes)
+	{
+		if(BattleHex::getDistance(hoveredHex, hex) == rangedFullDamageDistance)
+			rangedFullDamageLimitHexes.push_back(hex);
+	}
+
+	return rangedFullDamageLimitHexes;
+}
+
+std::vector<std::vector<BattleHex::EDir>> BattleFieldController::getOutsideNeighbourDirectionsForLimitHexes(std::vector<BattleHex> rangedFullDamageHexes, std::vector<BattleHex> rangedFullDamageLimitHexes)
```
IvanSavenko, line 435: Merge `getOutsideNeighbourDirectionsForLimitHexes` method with `calculateFullRangedDamageHighlightImages`?

client/battle/BattleFieldController.cpp
```diff
@@ -47,6 +47,33 @@ BattleFieldController::BattleFieldController(BattleInterface & owner):
 	cellUnitMovementHighlight = IImage::createFromFile("UnitMovementHighlight.PNG", EImageBlitMode::COLORKEY);
 	cellUnitMaxMovementHighlight = IImage::createFromFile("UnitMaxMovementHighlight.PNG", EImageBlitMode::COLORKEY);
 
+	// load single edges
+	fullDamageRangeLimitImages[0b000001] = IImage::createFromFile("rangeHighlights/green/topLeft.PNG", EImageBlitMode::COLORKEY);
```
IvanSavenko, line 51: Perhaps move into json animation config after all?

Mods/vcmi/Data/rangeHighlights/green/bottom.png
IvanSavenko, line 1: What have you decided on better / improved highlights, e.g. gradient?

client/battle/BattleFieldController.cpp
rilian-la-te: But why in CStack and not in battle::CUnitState? It looks like than it can be applied on all units, CStack is implementation detail.

IvanSavenko: If battle::CUnitState has access to bonus system - then OK I guess? No insisting on location

rilian-la-te: even battle::Unit is ACreature, so can use getBonusBearer() to check bonuses. And I think CUnitState is better location due to this factor.

client/battle/BattleFieldController.cpp
krs0: I do not understand this one :(. You mean there is a specific json file where these images can be specified?

IvanSavenko: What I mean is that instead of listing images in source code and loading them into a map you can use:
- CAnimation class in code
- json definition of "animation", like this one: https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/Sprites/PortraitsLarge.json

This will allow you to load all your images in like 2 lines of code without hardcoded list of paths.

Mods/vcmi/Data/rangeHighlights/green/bottom.png
krs0: For the moment nothing :(. Here someone with better graphics skills could change that? The logic being implemented.
Maybe after some usage of current implementation some new ideas will pop up.

What I want to do in my next PR is to show a similar limiter, but red for shooters with limited Range

client/battle/BattleFieldController.cpp
IvanSavenko:
```c++
 // in header
std::unique_ptr<CAnimation> fullDamageRangeLimitImages;

// in c++ on init
fullDamageRangeLimitImages = std::make_unique<CAnimation>("battle/rangeHighlightsGreen");
fullDamageRangeLimitImages->preload();

// in c++ on render
std::shared_ptr<IImage> imageToShow = fullDamageRangeLimitImages->getImage(indexToShow);
```

lib/battle/BattleHex.h
```diff
@@ -87,6 +87,10 @@ struct DLL_LINKAGE BattleHex //TODO: decide if this should be changed to class f
 	/// returns all valid neighbouring tiles
 	std::vector<BattleHex> neighbouringTiles() const;
 
+	/// returns all valid (not first and last columns) neighbouring tiles, with their relative direction
+	std::vector<std::pair<BattleHex::EDir, BattleHex>> neighbouringTilesWithDirection() const;
```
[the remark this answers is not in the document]
krs0: to draw the limit on the edge of the hexes, I need to know the edge (direction) where to draw. allNeighbouringTiles does not give me any direction information.

client/battle/BattleFieldController.cpp
krs0: Good ideea, will change.

lib/battle/BattleHex.h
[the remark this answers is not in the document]
krs0: Was looking at wrong function: neighbouringTiles :(
Ok, looks almost like what I need. I need only tiles that are valid... could I pass a parameter to this function?
```
std::vector<BattleHex> BattleHex::allNeighbouringTiles() const
{
	std::vector<BattleHex> ret;
	ret.resize(6);

	for(auto dir : hexagonalDirections())
		ret[dir] = cloneInDirection(dir, false);

	return ret;
}
```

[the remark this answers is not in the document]
IvanSavenko: Don't think that parameter is needed, you can:
1) Add check inside this method. This method is used once and caller is only interested in valid hexes.
2) Add check inside your method.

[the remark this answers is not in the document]
krs0: Added check in my method. Cannot add check inside because crash.
For reference: There are 2 distinct methods available: `allNeighbouringTiles `and `neighbouringTiles `this one returns only valid ones (but since is not all, direction cannot be inferred).

client/battle/BattleFieldController.cpp
krs0: Done

krs0:
Ok I have moved the image initialization into a .json file in VCMI mod. Still in code I needed a mapping between individual masks and "frame" numbers (aka Image index).
I do not know where to place the following so I kept them in `battleFieldController.cpp`
`enum HexEdgeMasks`
and
`initializeHexEdgeMaskToFrameIndex()`

krs0:
The following needs to be merged before this one.
https://github.com/vcmi/vcmi/pull/2177

client/battle/BattleFieldController.cpp
krs0: I will take a look on compacting these functions once the max range limit PR is done since some code I think it will be shared between the two.

IvanSavenko:
> Ok I have moved the image initialization into a .json file in VCMI mod. Still in code I needed a mapping between individual masks and "frame" numbers (aka Image index). I do not know where to place the following so I kept them in `battleFieldController.cpp` `enum HexEdgeMasks` and `initializeHexEdgeMaskToFrameIndex()`

FYI: you can simply have non-continious frame ID's in json definition.

And a lot of those files seems redundant... Files with multiple sides are simply combined versions of single-side files. And all but two of single-side files can be generated via vertical or horizontal flips.

krs0:
Are you sure frames must not be contiguous? I tried having masks in Json and console started complaining about the 'missing' frames.
About flipping images, do we have in the code any example? I did not knew that was possible.

krs0:
K will look into it.

krs0:
So, this flipping of images change: this means that after I load the Animation from .json, I will start adding frames to that Animation, that are flipped version of already existing frames?
Like top - bottom, left - right, top left - top right?

IvanSavenko:
@krs0 I guess so?
Check how terrain flipping works, see `MapTileStorage::load`

krs0:
Now less source images are used. Flipped versions will be calculated.

krs0:
Solving conflicts is not fun :(. Please make a decision on merging or discarding this.

IvanSavenko:
Sorry about that.

Let's merge to avoid more conflicts, but have it toggled off by default, until better UI is in.

krs0:
Ok is off by default. Can be merged.

krs0:
Once merged I will continue with showing a "RED" limit for MAX range.

PR #2157, opened 2023-05-15
Update in-game settings window to recent changes

IvanSavenko:
Changes to settings UI window that were postponed from adventure map PR to reduce it a bit.

List of changes / progress:
- [x] Implement "layouts" for configurable UI to reduce code duplication and manual math in UI json
- [x] Add "custom types" / "ui types library" for configurable UI to reduce copy-pasting of similar elements
- [x] Add "Borderless windowed fullscreen" and "Exclusive fullscreen" options instead of single "fullscreen" option (existed before, but only available in launcher)
- [x] Hide fullscreen options & resolution selector on mobile devices
- [x] Block resolution selector when borderless fullscreen mode is used
- [x] Resolution selector should show desktop resolution when borderless fullscreen mode is used


krs0:
Could you implement something like a simple `relative_position` that sets an element position, relative to the previous one?
When I want to insert a checkbox I need to manually go over all elements and update the numbers.

With `relative_position` when you insert the element, everything will just fall into place.

config/widgets/settings/generalOptionsTab.json
```diff
@@ -7,153 +9,105 @@
...
+				},
+				{
+					"text": "vcmi.systemOptions.fullscreenExclusive.hover",
+					"created" : "desktop"
+				},
+				{
+					"text": "vcmi.systemOptions.framerateButton.hover"
+				},
+				{
+					"text": "core.genrltxt.577"
```
IvanSavenko, line 54: @krs0 
What are you asking for is handled by "layouts" like this one.
For example this is definition of 6 labels, that:
- share same settings, as defined by "labelDescription" custom type (left-aligned, medium font, white color)
- placed vertically, at 30px intervals - defined by "verticalLayout" custom type
- (TODO) some labels only appear on desktop systems, so you won't have "fullscreen" option on Android and it won't result in any empty space between unavailable options - since only actually created labels will be added to layout.

Note that only container itself has position defined (45,85). Positions for individual labels are automatically generated based on layout settings

IvanSavenko: In other words, if you want to add one more label all you have to do is add one more section like
```json
{
	"text": "vcmi.systemOptions.myNewOption.hover"
},
```
And all positioning will be handled automatically

IvanSavenko:
Finished. Our configurable UI might benefit from few more improvements, but will do those later, in separate PR.

PR #2161, opened 2023-05-16
Separate Gui Handler into components - Window management

### LINKED ISSUE #1881: Incorrect text in status window for combat settings.
**Describe the bug**
When opening combat settings dialog, hovering on items should display something in status bar, but incorrect text is printed.

**To Reproduce**
1. Start a battle
2. Change View Hex Grid Option
3. See status window

**Screenshots**
image

**Version**
 - OS: Windows
 - Version: develop

IvanSavenko:
One more step to trim down & reorganize GuiHandler class.

List of changes:
- all code related to window stack management is now in a separate class WindowHandler
- renamed window-related functions from xxxInt to xxxWindow form
- added encapsulation to window stack state
- status bar is now private member of guihandler, with accessors functions
- Fixes #1881


IvanSavenko:
@rilian-la-te what do you mean by "move InfoWindow to CComponent"?

rilian-la-te:
@IvanSavenko now InfoWindow renders its components via internal instrument, but there is a CSelectableComponent and CComponent which doing exactly same job, but it is now swappable as is.

IvanSavenko:
True. But kind of scared to do anything with it at the moment. This is also one of the last few places in the codebase that uses direct access to SDL surfaces for rendering.

Agree that we need this change, but probably not something in my immediate plans.

PR #2163, opened 2023-05-16
Adventure map zoom

### LINKED ISSUE #1872: Map zooming
Mouse wheel map zooming, like view world but with animations, hero paths etc. and allow playing with such zoom.

IvanSavenko:
Basic version of zoomable adventure map.
Currently, zooming is possible using mouse wheel or with +- keys on numpad.
It is possible to reset zoom to default (100%) via backspace key.

Possible todo's:
- consider addition of several zoom levels (same as in world view?) 
- check release builds performance and whether it can be improved - due to map animations, this zoom is notably slower than unanimated world view mode.
- consider 'pinch' gesture for zooming on mobile systems (probably after GuiHandler refactoring)

Fixed issues:
- Fixes #1872 

Screenshots:
zoom-in
zoom-out


Laserlicht:
Nice. :)

Another feature would be a (linear?) filter for zoom levels < 100%. Nearest neighbor is not so nice here.
But I could imagine that this is quite complex, isn't it? So rather something for later, right?

IvanSavenko:
@Laserlicht 
Not a bad suggestion, but we use SDL which AFAIK does not provides high-quality scaling filters.

In theory, we can make our own, but those would be *way* slower than provided by SDL since SDL has diverse optimizations for almost every possible system. We already have custom rendering method which we use in some places and that thing *is* slow and uses way more CPU time than similar SDL methods.

And by the way - I think SDL does uses something better than nearest-neighbour. I have barely noticed any graphical artifacts on zoom levels close to 100%

Laserlicht:
Somethimes there are some grapical glitches:
grafik

Settings: 1920x1080@180%

Only in the biggest zoom steps

IvanSavenko:
I thought I fixed that?
Understood, will recheck.

IvanSavenko:
Scaling artifacts should be fixed

Laserlicht:
> Scaling artifacts should be fixed

I've tested it: Great work! :)

> And by the way - I think SDL does uses something better than nearest-neighbour.

Yes - but only on the basis of textures. If I understand it correctly, the surface is scaled with `SDL_BlitScaled`. And only after the tiles have already been scaled they are filtered. At 100% this is then an effect comparable to the nearest neighbour.
If we had textures, then `SDL_RenderCopy` should be able to be used, if I understood it correctly. But I think the effort to change that would currently be too high and we should leave it for now with what we already have.

IvanSavenko:
Scaling - yep.
In theory - we're not that far from hardware acceleration (aka - always use SDL_Textures for rendering). And not that far from HD/upscaled data support. I have mostly get rid of code that blocks these features, so in theory - doable in a reasonable time.

But already have way too much stuff to do. Might be possible for 1.4, but no promises.
And HD data can greatly improve upscaling, whether this one on adventure map or one from UI setting option.

PR #2164, opened 2023-05-17
Artifact related code refactoring part3 (ArtifactUtils, CArtHandler)

SoundSSGood:


cmake_modules/VCMI_lib.cmake
```diff
@@ -184,6 +184,7 @@ macro(add_main_lib TARGET_NAME LIBRARY_TYPE)
 
 		${MAIN_LIB_DIR}/vstd/StringUtils.cpp
 
+                ${MAIN_LIB_DIR}/ArtifactUtils.cpp
```
IvanSavenko, line 187: 1) Use tabs, not spaces.
2) Add header in list below this one, to include it into IDE projects

lib/ArtifactUtils.h
```diff
@@ -0,0 +1,45 @@
+/*
+ * CArtHandler.h, part of VCMI engine
```
IvanSavenko, line 2:
```suggestion
 * ArtifactUtils.h, part of VCMI engine
```

lib/ArtifactUtils.h
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#pragma once
+
+#include <vector>
```
IvanSavenko, line 12: Already included in global.h
```suggestion
```

lib/CArtHandler.cpp
```diff
@@ -871,15 +858,16 @@ void CArtifactInstance::putAt(ArtifactLocation al)
 {
 	assert(canBePutAt(al));
 
-	al.getHolderArtSet()->setNewArtSlot(al.slot, this, false);
+	al.getHolderArtSet()->CArtifactSet::putArtifact(al.slot, this);
```
IvanSavenko, line 861: Indended? Why such construct is needed instead of simple `->putArtifact`?

lib/ArtifactUtils.cpp
```diff
@@ -0,0 +1,223 @@
+/*
+ * CArtHandler.cpp, part of VCMI engine
```
IvanSavenko, line 2:
```suggestion
 * ArtifactUtils.cpp, part of VCMI engine
```

lib/ArtifactUtils.cpp
```diff
@@ -0,0 +1,223 @@
...
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#include "StdInc.h"
+#include "CArtHandler.h"
+
+#include "GameSettings.h"
+#include "ArtifactUtils.h"
```
IvanSavenko, line 15: "Main" include should be first (or well, second, after StdInc.h)
```suggestion
#include "ArtifactUtils.h"

#include "CArtHandler.h"
#include "GameSettings.h"
```

lib/CArtHandler.cpp
IvanSavenko: Frankly I'd rather have these 3 lines of duplicated code but more clear code than calling specific override of a virtual method.

IvanSavenko (approved):

PR #2177, opened 2023-05-21
Move battle fake labels into separate group, and small fixes

krs0:


config/widgets/settings/generalOptionsTab.json
```diff
@@ -185,11 +183,22 @@
...
+					"text": "vcmi.otherOptions.compactTownCreatureInfo.hover",
+				}
+			]
+		},
 		{
 			"name": "compactTownCreatureInfoCheckbox",
 			"type": "checkbox",
 			"help": "vcmi.otherOptions.compactTownCreatureInfo",
-			"position": {"x": 10, "y": 383},
+			"position": {"x": 10, "y": 389},
```
IvanSavenko, line 201: Why change? Checkboxes are 30px apart:
- 323; creatureGrowthAsDwelling
- 353: availableCreaturesAsDwelling
- so compactTownCreatureInfo should be at 383, as before

krs0: I have added a line divider to separate the 2 toggle options. Figured out a line is about 6 :)

IvanSavenko: It is actually 0 :)
Lines don't affect layout right now

Issue #2172, opened 2023-05-19
Better touchscreen UI support


IvanSavenko:
Note: all of this refers to "standard" mode, relative pointer mode (aka touchpad emulation) should probably stay as it.

Considering (at the moment) open PR's that:
1) Allow zoomable adventure map that needs some access for mobile systems
2) Refactor of input system that simplifies changes to touch input

We can now, with relatively low difficulty implement better management of touchscreen systems (e.g. Android) using gestures common in mobile UI use. Since SDL does not provides gesture recognition, this code will have to be done on VCMI side. Proposed list of gestures and their in-game usage:

Tap: (already supported)
- method: quick touch and release on approximately same location
- usage: generates left-click down + up events in game

Swipe:
- method: touch and move in horizontal or vertical direction (but not both)
- usage: generates swipe gesture event, acts similar to mouse wheel in most cases - scrolls underlying list
- usage: potentially, can be used in battles - swipe from attack position to enemy position to select attack direction
- usage: potentially, can be used to implement quick army management shortcuts

Pan:
- method: touch and move in *any* direction. Can be merged with swipe?
- usage: generates pan gesture event, acts to move adventure map view

Pinch:
- method: double-finger gesture, move two fingers closer / further away
- usage: generates pinch gesture event, acts to zoom adventure map view

Touch and hold: (iOS uses slightly different behavior)
- method: touch and keep finger at approximately same point for certain time
- usage: generates right-click down event to show popup window
- next tap gesture will be generate double-click up event to hide popup window

Current two finger-tap that is used to emulate right-click on Android will be removed

Nordsoft91:
Currently pinch is used for spacebar action. I'd advice triple touch for spacebar instead.

PR #2185, opened 2023-05-26
Touch input gestures

IvanSavenko:
Attempt to implement better support for touchscreen UI. For now focus is on gestures / touch detection, in-game UI adjustments are not planned at the moment.

Currently only tested on desktop via SDL touch input emulation mode, will test on Android using CI artifacts.

Progress:
- [x] Implement left-click emulation in vcmi instead of SDL
- [x] Implement right-click emulation via long-tap
- [ ] Implement detection of 'panning' gesture
- [ ] Implement detection of 'swipe' gesture, potentially - as part of 'panning' gesture
- [ ] Implement detection of 'pinch' gesture
- [ ] Use swiping gesture in game UI elements, for example - to move sliders similar to how mouse wheel does
- [ ] Use panning gesture to move adventure map view
- [ ] Use pinch gesture to zoom adventure map
- [ ] Recheck "relative cursor mode" for possible regressions

Optionally:
- [ ] Use 'swipe' gesture to scroll town/hero list on adventure map
- [ ] Use 'swipe' gesture for armies management (to replace ctrl/alt/shift shortcuts from desktop)
- [ ] Try to find a gesture suitable for 'spacebar' aka 'revisit current object' action on adventure map. Alternatively - add this option as new button in UI.

Resoled issues:
- Implements #2172


tianmaster:
After not seeing the mouse cursor, it is impossible to know the melee attack direction of the stack

dydzio0614:
@tianmaster there should be setting for mobile (maybe missing UI?) in 1.3 that changes how melee attack direction works

IvanSavenko:
> After not seeing the mouse cursor, it is impossible to know the melee attack direction of the stack

Noted. Will investigate once I start testing Android build.

> @tianmaster there should be setting for mobile (maybe missing UI?) in 1.3 that changes how melee attack direction works

Hard to call that setting a usable  feature - no UI or any other indication as to how it should work.

Nordsoft91:
> After not seeing the mouse cursor, it is impossible to know the melee attack direction of the stack

Enable touchscreen mode from settings (should be enabled by default actually, suspecting you are using 1.2.1)

tianmaster:
> touchscreen mode

No, it's because the long sword shaped cursor used to indicate the attack direction is hidden. I am using the current touch_ Input_ Gestures version. The touchscreen mode can only determine the attack direction through hex shadows, which is not intuitive enough

Nordsoft91:
- idea: swipe for switching selected reward. And same behavior using tab on desktops

>  Try to find a gesture suitable for 'spacebar' aka 'revisit current object' action on adventure map. Alternatively - add this option as new button in UI.

Triple touch. Tapping small buttons is very bad ux

> Try to find gesture suitable to 'cancel spellcast' action in both battlefield and adventure map

Long touch, as it stands for right click - same as in desktop and hard to miss.

> Consider option of giving orders in battles (e.g. move & attack) via gestures

Barely imagine that via gestures, gestures have non-zero fail rate. Best option is radial menu.

IvanSavenko:
> * idea: swipe for switching selected reward. And same behavior using tab on desktops

Not a bad idea. Or perhaps we should directly select & confirm action, without asking player to hit that tiny "OK" button? 

H3 dialogs only have 2 options to choose from, so can be mapped to swipe left / swipe right.
In that case - desktop will need same functionality on double-clicking reward icon.

> Triple touch (e.g 3 fingers together, not 3 sequential taps). Tapping small buttons is very bad ux

My idea was to make those buttons larger, potentially changing UI a bit. 
Not against triple-touch, but needs to be explained somewhere...

> Barely imagine that via gestures, gestures (or people) have non-zero fail rate. Best option is radial menu.

Swipe from position to attack from to enemy unit that you want to attack. Need to check how usable this is, considering small screen size.

Not sure how suitable radial menu would be specifically for selection of attack direction.

tianmaster:
> touch_ Input_ Gestures

I think giving orders in battle is a feasible solution. In fact, adding touch in version 1.3_ Input_ Before Gestures, I had been using a method similar to sliding touch on the mobile end - slowly sliding the cursor and observing the changes in the direction indicated by the cursor to determine the attack direction. Based on my personal experience, I think this method is quite good.

tianmaster:
Identify issues.
1. When attacking, pressing and holding for too long can easily cause accidental touch, triggering the right-click function.
2. If the stack of the attack happens to be within the hex at the bottom of the screen, and the finger moves to the lower sidebar position, the focus is lost when clicked, and the attack cannot be launched from bottom to top

Laserlicht:
Does the scrolling with the left mouse button still work on non-touch devices (since then it was possible via the setting)?

IvanSavenko:
> Does the scrolling with the left mouse button still work on non-touch devices (since then it was possible via the setting)?

Unfortunately, no, since that functionality has been moved to gesture. So now only mouse wheel on PC can be used for that.

Laserlicht:
Can we provide a functionality to change the key mapping for this function? From middle to left? As replacement for the missing setting.

Nowadays I think the left key is much more usual (e.g. Google Maps) than the middle "scrollwheel" button.

IvanSavenko:
In theory I guess we could try to add an option to interpret left clicks as touchscreen input.
This will allow all (single-finger) gestures in any UI that accepts them.

Not impossible but need to handle few cases, e.g. mouse hovering should still be interpreted as mouse input

Laserlicht:
That would also be a good idea.

tianmaster:
Sliding the adventure map immediately after sliding the mini map will cause the adventure map coordinates to be reset to the coordinate origin.
Clicking on the mini map and sliding the adventure map will not cause this to happen

Laserlicht:
> In theory I guess we could try to add an option to interpret left clicks as touchscreen input. This will allow all (single-finger) gestures in any UI that accepts them.
> 
> Not impossible but need to handle few cases, e.g. mouse hovering should still be interpreted as mouse input

Would it not also be possible to adjust the function in the InputSourceMouse.cpp? If I add the same functionality as the middle button to the left button, then it already works in principle.

The only problem here is that actions on the map are executed with Keydown. Keyup would be more practical here. There could be a condition that the mouse should not have moved significantly to difference between Click and Drag.

And another small QoL suggestion: iOS vibrates very briefly after recognizing the "right button" gesture. As haptic feedback. Should also be possible with SDL, right?

kambala-decapitator:
> And another small QoL idea: iOS vibrates very briefly after recognizing the "right button" gesture. As haptic feedback. Should also be possible with SDL, right?

I doubt it would work out of the box because SDL (unlike our current implementation) doesn't use native "gesture recognizer" objects to handle input

Alucard648:
As for revisiting object: what about hitting "Move Hero" when no paths are chosen?

IvanSavenko:
> The only problem here is that actions on the map are executed with Keydown. Keyup would be more practical here. There could be a condition that the mouse should not have moved significantly to difference between Click and Drag.

Yep. However that would be change of behavior compared to H3.
In theory we could simply use mouseMoved (SDL_MOUSEMOTION) event, and specifically for adventure map - always interpret mouse move as panning if left button is pressed. Might lead to some visual artifacts since map might move a bit when clicking. Need to check how such behavior would look & feel like

> And another small QoL idea: iOS vibrates very briefly after recognizing the "right button" gesture. As haptic feedback. Should also be possible with SDL, right?

Not really. As kambala said, we don't use native gesture recognition. Specifically:
- SDL2 does not have any sort of "gesture recognition" built-in. It can only provide basic finger down/moved/up events
- SDL2 does have some sort of haptic feedback support but no idea whether it is supported on iOS since it seems to be controller/gamepad-oriented. I also don't have iphone so no idea how to reach similar effect.
- Future SDL3 *might* provide access to native gesture recognition system, saw some open issues about that on SDL page. However it is only "might" and SDL3 is not even released yet, and it would take time for it to reach all platforms.

> As for revisiting object: what about hitting "Move Hero" when no paths are chosen?

As basic / default option - maybe. However adventure map buttons are too small on mobiles and not exactly obvious unless you know about this functionality

Laserlicht:
> Not really. As kambala said, we don't use native gesture recognition. Specifically:

Yes, I also meant our own gesture system.
The idea was that the vibration motor is active for a few ms after recognition. My Android does this, for example, when touching the home button.

But if sdl has no native way to control the vibration function, then it is also okay. The advantage of a haptic feedback would only be that you can feel when the right click is detected.

IvanSavenko:
Probably done for now, will move remaining points to a separate issue(s), not sure how to handle them at the moment so don't want to delay this PR further.

dydzio0614:
@IvanSavenko spacebar - maybe GUI button in bottom left corner of the screen, something like this: 
obraz - if just above bottom bar, otherwise if we want to place it on bar, assume high resolution and lots of blank space on high mobile resolution then we can do it like HD mod chat button: 
obraz

PR #2186, opened 2023-05-26
Swipe run out

Laserlicht:
Swipe run out smootly. Like for example google maps.

**Todo:**
- [ ] Find better run out function

### DIFF
--- a/client/mapView/MapViewActions.cpp
+++ b/client/mapView/MapViewActions.cpp
@@ -127,7 +127,7 @@ void MapViewActions::handleSwipeMove(const Point & cursorPosition)
 	}
 
 	if(isSwiping)
-	{	    
+	{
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Might be postpone this until after #2185 - since map swiping is heavily redone.

Not a bad idea though

Laserlicht:
> Might be postpone this until after #2185 - since map swiping is heavily redone.

Yes, makes sense.

Or do you want to implement that as part of the revision? Should then work with touch and mouse swipe.

> Not a bad idea though

Thank you. I think the function gives a much smoother experience especially on touchscreens.

IvanSavenko:
> Or do you want to implement that as part of the revision? Should then work with touch and mouse swipe.

Don't know actually. Definitely good idea for 1.3, but not sure if I'll do it as part of this PR.

Issue #2187, opened 2023-05-27
Launcher crash when press hota


Nordsoft91:
**To Reproduce**
1. Open launcher
2. press "download & refresh repositories"
3. Find horn of the abyss mod
4. Select it

**Expected behavior**
Hota mod info is shown

**Actual behavior**
Crash

**Version**
 - OS: MacOs
 - Version: 1.2.1, 1.3

IvanSavenko:
Not reproducible for me

Nordsoft91:
1) make sure hota mod isn't installed previously
2) check repository https://raw.githubusercontent.com/vcmi/vcmi-mods-repository/develop/vcmi-1.3.json

I have same issue even after downloading hota 1.6.0 manually

IvanSavenko:
vcmi-1.2 - everything OK
vcmi-1.3 - hota is missing for some reason (will investigate), but no crashes

IvanSavenko:
Tested both 1.2.1 and 1.3. Hota not installed

Both versions work fine with their corresponding repository (vcmi-1.2.json and vcmi-1.3.json).

Nordsoft91:
Crash happens in CModListView::genChangelogText from cmodlistview_moc.cpp in std::sort somewhere in qt stuff while arrays are fine. 
I was able to fix by replacing std::sort with qSort but it's marked as deprecated

IvanSavenko:
Noted.
Not sure if we even need sorting in there.

Feel free to submit PR, if not - will fix it myself for 1.3

PR #2188, opened 2023-05-27
Strict weak ordering on the data

Nordsoft91:
Fixes https://github.com/vcmi/vcmi/issues/2187

Explanation from chatgpt:

```
There might be an issue with the use of this function in the context of std::sort. Here's why:

Your original lambda function for std::sort negates the result of compareVersions, which might be causing issues with the sort. 
The compareVersions function returns true if versionLesser is smaller than versionGreater, but in the lambda function, you negate this result.
This means that the sort function will interpret true as "lesser is greater than greater", which is incorrect and might cause unexpected results.

The sorting algorithm expects a comparison function that provides a strict weak ordering.
If it doesn't, the algorithm's behavior is undefined, and it might cause an infinite loop, crash, or other unwanted behavior.
```

### DIFF
--- a/launcher/modManager/cmodlistview_moc.cpp
+++ b/launcher/modManager/cmodlistview_moc.cpp
@@ -200,8 +200,9 @@ QString CModListView::genChangelogText(CModEntry & mod)
 
 	std::sort(versions.begin(), versions.end(), [](QString lesser, QString greater)
 	{
-		return !CModEntry::compareVersions(lesser, greater);
+		return CModEntry::compareVersions(lesser, greater);
 	});
+	std::reverse(versions.begin(), versions.end());
 
 	for(auto & version : versions)
 	{


oleg-derevenetz:
> This means that the sort function will interpret true as "lesser is greater than greater", which is incorrect and might cause unexpected results.

LOL. That's not because "lesser is greater than greater" (WTF?), but because `comp` should return `true` if and only if the first argument of `comp` is lesser than the second, but `!(a < b)` is not `b < a`, but `a >= b` (i.e. `b <= a`). ChatGPT is fantasizing as always.

Nordsoft91:
> "lesser is greater than greater" (WTF?)

Arguments' names in lambda called "lesser" and "greater"

oleg-derevenetz:
> Arguments' names in lambda called "lesser" and "greater"

Yes, but ChatGPT's explanation is wrong anyway. `!(lesser < greater)` is not `lesser > greater`. There is nothing wrong with using `>` for comparison in `std::sort`to sort in descending order, but you can't use `>=` or `<=`. Also instead of using `std::reverse` you can just use reverse iterators in `std::sort`.

PR #2202, opened 2023-06-02
Map objects code reorganization

IvanSavenko:
No functionality / code changes, only reorganization of file layout.

Some of our map objects headers are basically included everywhere, as result - even slightest change in these files triggers almost full project rebuild, slowing down change code->build->test loop.

Changes:
- split several large files into set of smaller files, e.g. ObjectsHandler.cpp, ObjectClassesHandler.cpp
- reduced includes in headers in map objects files to only include files that are actually required


rilian-la-te:
@IvanSavenko all unit tests passes?

IvanSavenko:
@rilian-la-te 
They compile in CI.
And since there were no code changes other than file rearrangement they should pass just as before.

PR #2203, opened 2023-06-03
Remove random seed from map description

DjWarmonger:
Separate PR, as this is controvesial.

Rationale was: Players who can know seed can generate the same (or similiar) map in editor and cheat this way.

### DIFF
--- a/lib/rmg/CMapGenerator.cpp
+++ b/lib/rmg/CMapGenerator.cpp
@@ -164,9 +164,9 @@ std::string CMapGenerator::getMapDescription() const
 		throw rmgException("Map template for Random Map Generator is not found. Could not start the game.");
 
     std::stringstream ss;
-    ss << boost::str(boost::format(std::string("Map created by the Random Map Generator.\nTemplate was %s, Random seed was %d, size %dx%d") +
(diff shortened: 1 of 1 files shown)


dydzio0614:
I would maybe also change "the Random Map Generator" to "VCMI Random Map Generator"

BTW can anyone downvoting this issue explain what is his use case where seed is useful?

Alucard648:
I think, the seed can be used to further debug RMG during development process. Also it can be used to balance new map objects and testing as well as  implementation of VCMI RMG template editor. And what is cheating purpose of generating the same map multiple times?

kambala-decapitator:
> And what is cheating purpose of generating the same map multiple times?

I guess that you can play it in sp to learn things and then have advantage in mp game

Alucard648:
> > And what is cheating purpose of generating the same map multiple times?
> 
> I guess that you can play it in sp to learn things and then have advantage in mp game

As far, as I can tell, you could use the seed to generate map in editor, look at it secretly and use the knowledge to gain advantage in multiplayer, all without the system calling you cheater.

dydzio0614:
Adding to usefulness of this method for cheat prevention: since whole map is passed over network and non-host players do not get seed, there is no way for non-host to cheat by generating relevant map in editor or 2nd game instance without knowing seed

IvanSavenko:
Merge?

PR #2209, opened 2023-06-06
Support for json configuration of some map objects

IvanSavenko:
Preparation for hota h3m support. Allow defining properties of some map objects in config.

In theory / long-term we could convert some of these objects into Rewardable's, but does not seems to be easily doable at this point, so for now these objects will keep their hardcoded logic, but expose their parameters to json:
- [x] Spell Shrine: allow specifying allowed spells (using same logic as configurable objects)
- [x] Hill Fort: allow specifying upgrade costs per level
- [x] Shipyard (town building): allow specifying type of created boat
- [x] Shipyard (map object): allow specifying type of created boat

To be done by @DjWarmonger alongside with RMG changes:
- Tavern: allow water placement - new hero must be generated with boat
- Prison: allow water placement - released hero must be generated with boat


DjWarmonger:
What exactly do you want to configure for water prison? If a prison is placed on water, hero should spawn in a boat. period.

Actually I'd like to handle that along with RMG handling for water prison.

DjWarmonger:
Probably could take water tavern too, they will use same code pieces.

PR #2211, opened 2023-06-07
Fixing regressions

SoundSSGood:
Some fixes for issues found in https://github.com/vcmi/vcmi/pull/2055, https://github.com/vcmi/vcmi/pull/2164

### DIFF
--- a/client/widgets/CArtifactsOfHeroBase.cpp
+++ b/client/widgets/CArtifactsOfHeroBase.cpp
@@ -225,8 +225,8 @@ void CArtifactsOfHeroBase::updateWornSlots()
 
 void CArtifactsOfHeroBase::updateBackpackSlots()
 {
-	for(auto artPlace : backpack)
-		updateSlot(artPlace->slot);
+	if(curHero->artifactsInBackpack.size() <= backpack.size() && backpackPos != 0)
+		backpackPos = 0;
 	scrollBackpackForArtSet(0, *curHero);
 }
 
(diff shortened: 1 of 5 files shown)


kambala-decapitator:
any chance to write proper commit message(s) telling what they actually do? This PR is completely unclear regarding what it solves.

SoundSSGood:
The first regression
Place artifact as on picture. Then move one to the backpack, and equip the one in the backpack. Next, pick up the artifact in the backpack.
er00er01er02

The second. Place any artifact into backpack. Place into backpack the second art Angelic alliance. Next, disassemble the Angelic alliance and scroll backpack right by 1. Assemble Angelic alliance. As a result, the backpack will remain scrolled to the right by 1.
er10er11er12

IvanSavenko:
By that way, not sure if this is related to your recent changes or an old issue, but on current develop if starting bonus is set to artifact AND appropriate slot of our starting hero is NOT empty then this will lead to an assertion failure. 

Release build seems to be working fine - new artifact gets sent to the backpack

Test map with hero that has no empty slots for starting artifact:
[SiegeTests.zip](https://github.com/vcmi/vcmi/files/11678832/SiegeTests.zip)

SoundSSGood:
> Test map with hero that has no empty slots for starting artifact:

thanks, will take a look

IvanSavenko:
Very similar/same issue - visit Warrior's Tomb with no free equipment slots
(same test map, a bunch of tombs to the left from staring town)

SoundSSGood:
@IvanSavenko Fixed in this PR.

Nordsoft91:
@SoundSSGood, does it fix https://bugs.vcmi.eu/view.php?id=3253 ?

SoundSSGood:
> @SoundSSGood, does it fix https://bugs.vcmi.eu/view.php?id=3253 ?

does not fix

PR #2232, opened 2023-06-15
Fix crash in VCAI

wb180:
Start any map with VCAI as a map AI and skip some turns, it will often crash in the first 2-3 turns.


kambala-decapitator (approved):

dydzio0614:
Blind approval from me - won't hurt even if passing 1st condition condition should be handled another way since VCAI is pretty much abandoned and has some non-trivial design flaws / action calculation imbalances to fix, with nobody on the horizon willing to tackle the task

PR #2236, opened 2023-06-17
Create only a single seer hut for every artifact we have

DjWarmonger:


lib/rmg/modificators/TreasurePlacer.cpp
```diff
@@ -477,6 +477,7 @@ void TreasurePlacer::addAllPossibleObjects()
 			}
 
 			oi.probability = 10;
+			oi.maxPerZone = 1;
```
Nordsoft91, line 480: What is the logic? I guess it means that there will be no more than 1 seer hut per zone with quest (not reward) for special art. Is it correct?

DjWarmonger: Each artifact can be only used once, so the number of available artifacts is evenly distributed between the zones and Seer Hut variants. Thus, eeach of these Seer Huts should appear at most once.

PR #2237, opened 2023-06-17
Wide, fictive, repulsive connections

DjWarmonger:
- [x] Wide connections - have no border, no guard or road between zones
- [x] Fictive connections - attract zones, but are not real connections
- [x] Repulsive connections - repulse zones, but are not real connections


lib/rmg/modificators/ObjectManager.cpp
```diff
@@ -86,6 +86,18 @@ void ObjectManager::updateDistances(const rmg::Object & obj)
...
+{
+	RecursiveLock lock(externalAccessMutex);
+	tilesByDistance.clear();
+	for (auto tile : zone.areaPossible().getTiles()) //don't need to mark distance for not possible tiles
+	{
+		ui32 d = pos.dist2dSQ(tile); //optimization, only relative distance is interesting
+		map.setNearestObjectDistance(tile, std::min(static_cast<float>(d), map.getNearestObjectDistance(tile)));
+		tilesByDistance.push(std::make_pair(tile, map.getNearestObjectDistance(tile)));
+	}
+}
```
Nordsoft91, line 99: This code is pure copy-paste of method above, with changes in a single line. Is it possible to avoid somehow?

DjWarmonger:
Ready now, works as expected. "repulsive" connections are a new feature not present in OH3 or HoTA, and they do their job. Check edited Brawl:
[brawl.zip](https://github.com/vcmi/vcmi/files/11852411/brawl.zip)

lib/rmg/modificators/ObjectManager.cpp
DjWarmonger: I could wrap the line in lambda, but it would be very artifical, inlexible and wouldn't make the total lines of code any smaller.

PR #2239, opened 2023-06-17
Android Launcher: Remove unused settings + add scaling

Laserlicht:


kambala-decapitator:
resolution used to work, when did it break?

and swipe broke after input changes?

IvanSavenko:
@kambala-decapitator 
Resolution - effectively removed. On Android/iOS VCMI now always uses native resolution. Instead these platforms use UI scaling to control game size. Main reason is that there were too many issues with using resolutions that have different aspect ratio than actual screen. If we can *replace* resolution option with UI scaling that would be great, but not required IMO - scaling selectable in-game without restart anyway.

Swipe - now always active, so we now have two input methods: standard (with gestures - swipe/pinch/long touch) and relative (touchpad-like)

kambala-decapitator:
@IvanSavenko thanks a lot! Haven't tested 1.3 builds yet

kambala-decapitator:
@Laserlicht do you have troubles setting up local development environment for android? Do you need any help?

as I understand, you essentially edit code in notepad or directly on github, then wait for CI to produce build, then test it...

Laserlicht:
> @Laserlicht do you have troubles setting up local development environment for android? Do you need any help?
> 
> as I understand, you essentially edit code in notepad or directly on github, then wait for CI to produce build, then test it...

Yes, I thought it was just a small thing at first. And I would get along without local setup for android. But now it is also only small things and it works basically.

Should you squash when merging, so that the commits are combined.

kambala-decapitator:
Hint: disable (comment) all other platforms in https://github.com/vcmi/vcmi/blob/develop/.github/workflows/github.yml (line 71 and onwards) for such way of testing changes, then drop this commit when done

IvanSavenko (approved):

PR #2245, opened 2023-06-20
Allow defining default values of settings per platform

IvanSavenko:
Allows settings with different default values per platform (e.g. Android / iOS).

This should allow having (for example) UI scaling at value higher than 100% on mobile, high-dpi screens as default without custom handling of every such option.

Currently this feature is largely unused and is mostly added for future use.

### DIFF
--- a/lib/JsonNode.cpp
+++ b/lib/JsonNode.cpp
@@ -1135,73 +1135,76 @@ Key reverseMapFirst(const Val & val, const std::map<Key, Val> & map)
 	return "";
 }
 
-void minimizeNode(JsonNode & node, const JsonNode & schema)
+static JsonNode getDefaultValue(const JsonNode & schema, std::string fieldName)
 {
-	if (schema["type"].String() == "object")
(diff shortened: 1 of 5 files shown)


lib/JsonNode.cpp
```diff
@@ -1135,73 +1135,78 @@ Key reverseMapFirst(const Val & val, const std::map<Key, Val> & map)
...
-			foundEntries.insert(name);
+#ifdef VCMI_IOS
+	if (!fieldProps["defaultIOS"].isNull())
+		return fieldProps["defaultIOS"];
+#endif
+#ifdef VCMI_ANDROID
+	if (!fieldProps["defaultAndroid"].isNull())
+		return fieldProps["defaultAndroid"];
+#endif
+#ifndef VCMI_MOBILE
```
kambala-decapitator, line 1150: since these values are exclusive, I think it's better to use

```
#if defined(VCMI_IOS)
#elif defined(VCMI_ANDROID)
#else
```

PR #2252, opened 2023-06-24
Crash fix. Cosmetic fixes.

SoundSSGood:
Fixed crash when unequip commander artifacts (regression).

I also noticed that CGrowingArtifact is used only for commander artifacts. Perhaps it makes sense add the same "growing" property to the hero's artifacts?

### DIFF
--- a/lib/ArtifactUtils.cpp
+++ b/lib/ArtifactUtils.cpp
@@ -204,7 +204,7 @@ DLL_LINKAGE CArtifactInstance * ArtifactUtils::createArtifact(CMap * map, const 
 			art = ArtifactUtils::createScroll(SpellID(spellID));
 		}
 	}
-	else //TODO: create combined artifact instance for random artifacts, just in case
+	else
 	{
 		art = new CArtifactInstance(); // random, empty
 	}
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -849,12 +849,12 @@ bool CArtifactInstance::canBePutAt(const ArtifactLocation & al, bool assumeDestR
 	return artType->canBePutAt(al.getHolderArtSet(), al.slot, assumeDestRemoved);
 }
 
-void CArtifactInstance::putAt(ArtifactLocation al)
+void CArtifactInstance::putAt(const ArtifactLocation & al)
 {
 	al.getHolderArtSet()->putArtifact(al.slot, this);
 }
 
-void CArtifactInstance::removeFrom(ArtifactLocation al)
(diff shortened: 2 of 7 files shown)


lib/ArtifactUtils.cpp
```diff
@@ -204,7 +204,7 @@ DLL_LINKAGE CArtifactInstance * ArtifactUtils::createArtifact(CMap * map, const
 			art = ArtifactUtils::createScroll(SpellID(spellID));
 		}
 	}
-	else //TODO: create combined artifact instance for random artifacts, just in case
```
SoundSSGood, line 207: Generation of combined artifacts on random maps does not match the behavior of OH3. I suggest removing this TODO.

lib/CArtHandler.cpp
```diff
@@ -819,9 +819,6 @@ void CArtifactInstance::init()
 std::string CArtifactInstance::getDescription() const
 {
 	std::string text = artType->getDescriptionTranslated();
-	if (!vstd::contains(text, '{'))
-		text = '{' + artType->getNameTranslated() + "}\n\n" + text; //workaround for new artifacts with single name, turns it to H3-style
```
SoundSSGood, line 823: I did not find a single case of using this workaround. I suggest removing it.

lib/ArtifactUtils.cpp
IvanSavenko: Perhaps add an assertion then? So if somebody uses this function for combined artifact he won't end up with broken state.

IvanSavenko (approved):

misiokles:
Growing bonus for hero's artifacts is `"type" : "GROWS_WITH_LEVEL" ` and works fine in 1.2.x
https://github.com/vcmi-mods/wake-of-gods/blob/vcmi-1.2/content/config/artifactsWoG.json

How about creature artifacts in daily builds? They can be worn on/off without issues?
https://github.com/vcmi-mods/wake-of-gods/blob/vcmi-1.2/Mods/StackArts/Content/config/warlordBanners.json

SoundSSGood:
> Growing bonus for hero's artifacts is `"type" : "GROWS_WITH_LEVEL" ` and works fine in 1.2.x

GROWS_WITH_LEVEL - It is not growing artifacts. It is hero's level related artifacts. Growing artifacts have their own level, which increases after each battle. I think in the future it will be possible to add other conditions for growth.

lib/ArtifactUtils.cpp
SoundSSGood: Just checked. If allow in the generator, the combined artifacts will appear on the map. So, assertion is not needed.

PR #2268, opened 2023-06-30
CArtifact CArtifactInstance refactoring

SoundSSGood:
CArtifact CArtifactInstance refactored. "Combined", "growing", "scroll" artifact related code structurized and moved to separated classes. CArtifactInstance moved to own file.

Also now it is possible to use "growing" tag in json files for hero's artifact.
```
"growing": {
	"bonusesPerLevel": [
		{
			"level": 6,
			"bonus":  {
				"type" : "PRIMARY_SKILL",
				"subtype" : "primSkill.attack",
				"val" : 1
			}
		}
	]
}
```
This tag can be used simultaneously with "components"


dydzio0614:
BTW @SoundSSGood do you have plan to implement artifact backpack UI from HD mod someday?

SoundSSGood:
> BTW @SoundSSGood do you have plan to implement artifact backpack UI from HD mod someday?

yes

dydzio0614:
maybe should look and work like in HD mod except graphical part should not be done by copying and photoshopping H3 assets, I guess as last resort Ivan can help with graphical part

SoundSSGood:
> maybe should look and work like in HD mod except graphical part

I think yes. It should be 8x8 artifact window, like in the HD mod. But there also needs to be a scrollbar to support the infinite backpack.

krs0:
Is it hard to add support for fused artifacts?
Fused artifacts
Will take only one slot (because while artifact count on maps increased, slots remain the same)
Eg. Fusing the different neck 'protection from xxx'

SoundSSGood:
> Is it hard to add support for fused artifacts?

Never heard of fused artifacts. I think it should be no more difficult than combined.

krs0:
Well it will be new in VCMI :)

IvanSavenko:
> maybe should look and work like in HD mod except graphical part should not be done by copying and photoshopping H3 assets, I guess as last resort Ivan can help with graphical part

Screen is trivial actually:
- for window background we can use FilledTexture with H3 dialog box background (like in settings window)
- for window borders we also can use H3 assets, check settings window, showAll method I think
- existing sliders can be reused, they are not fixed-size
- for artifacts background I can make semi-transpared background. But that would be the only custom asset.

client/widgets/CArtifactsOfHeroBase.cpp
```diff
@@ -257,11 +257,11 @@ void CArtifactsOfHeroBase::setSlotData(ArtPlacePtr artPlace, const ArtifactPosit
 	{
 		artPlace->lockSlot(slotInfo->locked);
 		artPlace->setArtifact(slotInfo->artifact);
-		if(!slotInfo->artifact->canBeDisassembled())
+		if(!slotInfo->artifact->isCombined())
 		{
 			// If the artifact is part of at least one combined artifact, add additional information
 			std::map<const CArtifact*, int> arts;
-			for(const auto combinedArt : slotInfo->artifact->artType->constituentOf)
+			for(const auto combinedArt : slotInfo->artifact->artType->partOf)
```
IvanSavenko, line 264: Perhaps add getter method for `partOf`?

lib/CArtifactInstance.h
```diff
@@ -0,0 +1,96 @@
...
+		template <typename Handler> void serialize(Handler & h, const int version)
+		{
+			h & art;
+			h & slot;
+		}
+		PartInfo(CArtifactInstance * art = nullptr, const ArtifactPosition & slot = ArtifactPosition::PRE_FIRST)
+			: art(art), slot(slot) {};
+	};
+	std::vector<PartInfo> partsInfo;
+	void addArtInstAsPart(CArtifactInstance * art, const ArtifactPosition & slot);
```
IvanSavenko, line 37: Perhaps simply `addPart()`? "ArtifactInstance" part should be clear from method parameters and in line with form used by `isPart` method

Or use full name, `addArtifactInstanceAsPart`, abbreviations are not exactly readable

client/widgets/CArtifactsOfHeroBase.cpp
SoundSSGood: Will do. VCMI often uses direct access to class fields. If we want to bring VCMI to a uniform style, maybe we should add a rule to the Coding guidelines? Because now the coding style is rather obscure.

IvanSavenko:
@SoundSSGood 
> VCMI often uses direct access to class fields. If we want to bring VCMI to a uniform style, maybe we should add a rule to the Coding guidelines? Because now the coding style is rather obscure.

Coding guidelines do need a review, but stuff like direct access to fields vs get'ters is not part of guidelines since it mostly focuses on formatting style. Might review it eventually when I have time.

Personally I think that object should be either "dump c struct" with no methods other than `serialize` or be fully-fledged class with proper encapsulation.

As for my current long-term view of how our handlers classes should work:
(using artifacts as an example, but same is likely true for the rest of our handlers)
- Access from "external" code should be whenever possible done via 'services', e.g. `VLC->artifacts()` instead of direct access to handler via `VLC->arth` which should be deprecated.
- interface of the entity (e.g. Artifact) should be extended whenever needed for use with services
- entities, such as CArtifact should have private properties with get'ters whenever needed.
- CArtifactHandler should be a 'friend' of CArtifact to allow access to private members for object loading

Note that this is just my opinion based on current state of the code, not going to enforce it. Not even 100% sure whether this approach would work in the end without creating massive boilerplate code.

lib/CArtifactInstance.cpp
```diff
@@ -0,0 +1,192 @@
...
+			return true;
+	}
+
+	return false;
+}
+
+std::vector<CCombinedArtifactInstance::PartInfo> & CCombinedArtifactInstance::getPartsInfo()
+{
+	// TODO romove this func. encapsulation violation
+	return partsInfo;
```
SoundSSGood, line 50: Since there is still code that requires write access to std::vector<PartInfo> partsInfo, this will be fixed a bit later.

lib/CArtifactInstance.h
```diff
@@ -0,0 +1,102 @@
...
+	ConstTransitivePtr<CArtifact> artType;
+
+	CArtifactInstance(CArtifact * art);
+	CArtifactInstance();
+	void setType(CArtifact * art);
+	std::string nodeName() const override;
+	std::string getDescription() const;
+	ArtifactID getTypeId() const;
+	ArtifactInstanceID getId() const;
+	void setId(ArtifactInstanceID id);
```
SoundSSGood, line 83: I think the ID should be set somehow differently. Most likely through the constructor. In any case, I will deal with this in another PR.

SoundSSGood:
Presumably done. Although there are still parts of the code with violation of encapsulation. This will be resolved in future PRs.

PR #2271, opened 2023-07-01
Fix flag for MSVS

DjWarmonger:
This drives me nuts.


DjWarmonger:
> not sure why you even enable tests :)

There are separate "gmock" and "getst" targets which are compiled, even if tests themselves are not.

Issue #1244, opened 2022-12-18
RMG map template. "allowedTowns" array - unable to resolve custom towns ids.


Macron1Robot:
RMG map template. "allowedTowns" array - unable to resolve custom towns ids.

There is no way to use custom towns in rmg templates. 

`TRACE global [c78] - loading .\config\schemas\template.json
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
ERROR mod [c78] - Failed to resolve identifier farriery of type faction from mod core
INFO mod [c78] - 		[DONE] Total Assault`

Macron1Robot:
[template_issue.zip](https://github.com/vcmi/vcmi/files/10253804/template_issue.zip)
here is test mods

IvanSavenko:
Not reproducible with current beta build:
- Installed your mods
- Generated random map at default parameters: error
- Looked up your template
- Generated random map at XL size
- Map successfully generated
- Crash on start due to missing music file
- Looked up your mod config, removed missing files
Restarted game, generated map, map seems to be normal, all players use faction from mod

Macron1Robot:
And there is no other towns on map different from Farriery?

IvanSavenko:
Nope

Nordsoft91:
I can reproduce on both beta and develop. @IvanSavenko did you use maybe one of your dev branches with schema fixes?

IvanSavenko:
No. Checked out beta branch, disabled most mods including default templates mod from vcmi-extra.
Will recheck in more detail.

IvanSavenko:
Still unable to reproduce...
```
$ git status
HEAD detached at vcmi/beta
nothing to commit, working tree clean
```
[...]

PR #2278, opened 2023-07-02
Rmg factions

DjWarmonger:
- [x] Fix modded towns not spawning in random zones
- [x] Fix Wasteland not spawning in random zones
- [x] Handle bannedTowns
- [x] Update all the templates with bannedTowns if applicable
- [x] Look into #1244 
- [x] New field - bannedTerrains (will carry over with option "terrainTypeLikeZone")
- [x] New field - bannedMonsters


Mods/vcmi/config/vcmi/rmg/hdmod/frozenDragons.JSON
```diff
@@ -12,7 +12,7 @@
 				"owner" : 1,
 				"monsters" : "normal",
 				"playerTowns" : { "castles" : 1 },
-				"allowedTowns" : [ "rampart", "tower", "inferno", "dungeon", "stronghold", "fortress" ],
+				"bannedTowns" : ["castle", "necropolis", "conflux"],
 				"allowedMonsters" : [ "castle", "tower", "inferno", "stronghold", "fortress", "neutral", "conflux" ],
```
IvanSavenko, line 16: How do we handle checks like this when considering mods?
Will this check also ban all creatures from mods?

In that case we might want to invert it to "bannedMonsters"?

DjWarmonger: Yes, it will not allow creatures from new factions. However, I don't think it has much impact on balance. It's more for a look & feel, or to tell the zones apart.
Surely I could add "banned Monsters", and even  "banned terrains" as well.

IvanSavenko: Understood.

Well, up to you. No idea whether this option is actually needed.

DjWarmonger:
Okay, template now may use terrain from mod set as dependency:
image

lib/GameConstants.cpp
```diff
@@ -215,6 +216,31 @@ std::string FactionID::encode(const si32 index)
 	return VLC->factions()->getByIndex(index)->getJsonKey();
 }
 
+std::string FactionID::scope()
```
IvanSavenko, line 219: Perhaps rename to entityType / entityClass? Or anything else that does not contains "scope" in its name?

We normally use word "scope" for defining mod that generated this object/json field. Using "scope" for similar purpose might lead to mistakes.

Issue #2199, opened 2023-05-31
Armies management shortcuts for touchscreen devices


IvanSavenko:
Currently, we provide keyboard shortcuts for hero armies management. Namely:
- move unit or merge if unit in destination is the same (via simple click on another slot)
- move unit to another army, and merge if such unit already exists in another army (button in extended UI dialog)
- split unit dialog (via shift-click on the same unit or empty slot)
- show creature info dialog (double-click)
- split unit equally into free slot and reorganize existing stacks of same creature to equal size (shift+click)
- split single unit into a free slot (ctrl+click)
- merge same units into single slot (alt+click)

However, on touchscreen systems these shortcuts are inaccessible and buttons in exchange screen are too small to be usable.

We should provide similar shortcuts via touchscreen gesture, for example - by displaying radial wheel with all of these options on swiping.

PR #2292, opened 2023-07-05
Radial wheel for army management

IvanSavenko:
Implements #2199 

Can be used on desktop via mouse wheel press and on mobile via swipe gesture. 

Supported actions:
- show unit info
- merge all units of the same type
- move unit to another army
- split off 1 unit into an empty slot
- split unit into one more slot equally
- enter unit split mode (same as button)

Progress:
- [x] Fix bugs with displaying/hiding radial wheel
- [x] Make sure that radial wheel is always inside screen borders
- [x] Add user feedback - selected action should be indicated, e.g. highlighted
- [x] Add hover text - show description of selected action in status bar
- [x] Handle inexact swiping, e.g. if cursor / finger is over the gap
- [x] Use entire screen for selection - if cursor is outside of hex grid neareast action should be selected
- [x] Block or hide illegal actions
- [x] Handle case of no valid actions (empty slot)
- [x] Review actions position on the wheel
- [x] Add translations to hover texts
- [x] Add option(s) to configure activation mode (without UI?)

To be done separately:
- Allow accessing radial wheel using right-click and not just wheel/swipe
- Try to improve visuals
- (1.3.x?) Use radial wheel for adventure map (revisit object, show info, show path cost, reset zoom)
- (1.4 or later) Use radial wheel for combat (select action, show unit info, something else?)

Screenshot:
radialWheel


tianmaster:
Several suggestions/questions
After clicking on the unit to display the wheel UI, if you click on another space without units, a new wheel UI will appear, but the previous wheel UI does not disappear. At this point, there are two wheel UIs on the screen at the same time.
2. May the position of the wheel UI be low in screen space? Can you consider appearing in the middle of the unit grid UI, but slightly above it
PS: I think the wheel UI operation is very modern. Have you ever considered using this method for battlefield attacks/movement commands? (If you have free time, thank you)

IvanSavenko:
@tianmaster 
1. Noted. Right now this is very quick and hacky implementation to see how such UI element works like. So bugs are to be expected at the moment.
2. Will recheck positioning, including how it works on Android. Idea is to have menu centered around click position (or alternatively - around center of clicked UI element). And if needed - moved to fit into the screen.
3. Yeah, currently I want generic UI element that can be reused in other scenarios - selection of action on battlefield (attack, shoot, move, show unit information) and for adventure map (show move cost, revisit current object, show object information, reset zoom level). However for battles and adventure map I might need to use right click / long tap for radial wheel activation - since swipe gesture is already taken.

Laserlicht:
Feedback from my side:
- Really good idea (for many parts of the game)
- After moving mouse over icon:
   - it should be highlighted as user feedback (e.g. border color, size...)
   - a text should be shown, which describes the function (e.g. Statusbar / above button)
- We need to find a way between this gesture and "right click" gesture. For example two times: after x milliseconds the wheels appears. and after still holding another x milliseconds the wheel disappears and the right click menu appears.

IvanSavenko:
@Laserlicht 
Thanks for feedback, noted.

- selection feedback - yeah, planned, Currently thinking about showing border only for selected element, having different icon color for selected element, or both.
- action description - yeah, planned, in form of status bar. Might need to place it above radial wheel for mobile systems. Will think about it.
- right click - actually considering having it as alternative option to show this radial wheel. Will be necessary for adventure map and battles where swiping is already in use
- inexact swiping - noted. Might not be important if selection feedback is present. But should be trivial to do in any case.
- screen borders - yeah, current plan is to move radial wheel within screen borders

tianmaster:
Test feedback
1.Can we expand the judgment range to six equally divided regions? Currently, as long as you move out of the grid range, you will lose focus and require very precise control.
2. Clicking on an empty slot still causes the wheel to appear
3. Can you keep the logic of double-clicking to directly enter the bioinformatics interface?
thanks

IvanSavenko:
Updated PR description with current progress.

@tianmaster 
1. Yeah, I am leaning to this approach right now.
2. Known, will fix
3. Yes, left click (desktop) / short tap (mobile) behavior won't be changed.

Notes:
1. Yeah, I did noticed that something is wrong with unit movement. Will investigate
2. Will playtest this on Android a bit later. But moving units should be same as before? Short tap should select unit to move and another short tap should move it? Wheel should show up only if you actually drag/swipe screen.

tianmaster:
> 2\. Will playtest this on Android a bit later. But moving units should be same as before? Short tap should select unit to move and another short tap should move it? Wheel should show up only if you actually drag/swipe screen.

_Will playtest this on Android a bit later. But moving units should be same as before? Short tap should select unit to move and another short tap should move it? Wheel should show up only if you actually drag/swipe screen._
I think it can be done this way, which is in line with the operational logic of most current applications. Or do you have any other ideas? At least it can't be like this now. It's completely impossible to use the function of moving units at now, especially when you need to exchange units

IvanSavenko:
@tianmaster 
Will recheck Android build a bit later, but slightly older build I have on phone right now indeed has broken unit movement. Will make sure that it is fixed.

tianmaster:
> 

Okay, my suggestion is to maintain the click and double click logic as before, and the wheel should only appear as an auxiliary operation during long clicks?

IvanSavenko:
Mostly done, not yet sure what to do with icons, but functionality-wise mostly complete

Screenshot:
radialWheel

IvanSavenko:
Will probably merge it soon and do any remaining changes later
Remaining TODO before merge:
- recheck texts and add to localizations
- add option (probably without UI) to disabled this
- possibly consider how to improve UI

PR #2301, opened 2023-07-08
Haptic feedback

Laserlicht:
Haptic feedback for inputs ~~(currently only Android)~~.

Unsure if also clicks should feedback or only long press... What do you think?

### DIFF
--- a/client/eventsSDL/InputSourceTouch.cpp
+++ b/client/eventsSDL/InputSourceTouch.cpp
@@ -108,6 +108,7 @@ void InputSourceTouch::handleEventFingerDown(const SDL_TouchFingerEvent & tfinge
 {
 	// FIXME: better place to update potentially changed settings?
 	params.longTouchTimeMilliseconds = settings["general"]["longTouchTimeMilliseconds"].Float();
+	params.hapticFeedbackEnabled = settings["general"]["hapticFeedback"].Bool();
 
 	lastTapTimeTicks = tfinger.timestamp;
 
@@ -298,7 +299,7 @@ void InputSourceTouch::emitPinchEvent(const SDL_TouchFingerEvent & tfinger)
 }
 
 void InputSourceTouch::hapticFeedback() {
-	if(hapticFeedbackEnabled)
+	if(hapticFeedbackEnabled) {
 #if defined(VCMI_ANDROID)
 		CAndroidVMHelper vmHelper;
 		vmHelper.callStaticVoidMethod(CAndroidVMHelper::NATIVE_METHODS_DEFAULT_CLASS, "hapticFeedback");
--- a/client/eventsSDL/InputSourceTouch.cpp
+++ b/client/eventsSDL/InputSourceTouch.cpp
@@ -174,7 +174,6 @@ void InputSourceTouch::handleEventFingerUp(const SDL_TouchFingerEvent & tfinger)
 			GH.input().setCursorPosition(convertTouchToMouse(tfinger));
 			GH.events().dispatchMouseLeftButtonPressed(convertTouchToMouse(tfinger));
 			GH.events().dispatchMouseLeftButtonReleased(convertTouchToMouse(tfinger));
-			hapticFeedback();
 			state = TouchState::IDLE;
 			break;
 		}
--- a/client/eventsSDL/InputSourceTouch.cpp
+++ b/client/eventsSDL/InputSourceTouch.cpp
@@ -39,6 +39,7 @@ InputSourceTouch::InputSourceTouch()
 	params.useRelativeMode = settings["general"]["userRelativePointer"].Bool();
 	params.relativeModeSpeedFactor = settings["general"]["relativePointerSpeedMultiplier"].Float();
 	params.longTouchTimeMilliseconds = settings["general"]["longTouchTimeMilliseconds"].Float();
+	params.hapticFeedback = settings["general"]["hapticFeedback"].Bool();
 
 	if (params.useRelativeMode)
 		state = TouchState::RELATIVE_MODE;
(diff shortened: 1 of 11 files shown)


IvanSavenko:
Would be nice to have it toggleable or to use OS API to detect whether user has haptic feedback enabled

dydzio0614:
@Laserlicht probably only long press for now to go safe, though I am not android player so I am not very good at imagining this in practice

Laserlicht:
@IvanSavenko Added setting. Have to test it first. But should be correct?

@dydzio0614 Okay - only Right-"Click" for now. @IvanSavenko I think we should also add it for an action in the radial wheel.

I also added iOS. But I cannot test it. Can someone confirm the functionality?

dydzio0614:
will probably take a while to get iOS test results...

Laserlicht:
> will probably take a while to get iOS test results... 

Maybe I can test with the iPhone from someone...

IvanSavenko:
@Laserlicht 
Check InputSourceTouch::handleEventFingerDown
You need to update settings once they change, similar to long touch duration. Kind of workaround solution at the moment, but works.

@dydzio0614 Might get one eventually (iPhone, not Patreon). My current phone seems to be slowly but surely approaching his end.

kambala-decapitator:
Will try to test on iPad in the next few days

client/eventsSDL/InputSourceTouch.cpp
```diff
@@ -22,6 +22,13 @@
 #include "../gui/MouseButton.h"
 #include "../gui/WindowHandler.h"
 
+#ifdef VCMI_ANDROID
+#include "../../lib/CAndroidVMHelper.h"
+#endif
+#ifdef VCMI_IOS
+#include "../ios/utils.h"
+#endif
```
kambala-decapitator, line 30:
```suggestion
#if defined(VCMI_ANDROID)
#include "../../lib/CAndroidVMHelper.h"
#elif defined(VCMI_IOS)
#include "../ios/utils.h"
#endif
```

client/eventsSDL/InputSourceTouch.cpp
```diff
@@ -287,3 +297,15 @@ void InputSourceTouch::emitPinchEvent(const SDL_TouchFingerEvent & tfinger)
...
 		GH.events().dispatchGesturePinch(lastTapPosition, distanceNew / distanceOld);
 }
+
+void InputSourceTouch::hapticFeedback() {
+	if(hapticFeedbackEnabled) {
+#if defined(VCMI_ANDROID)
+		CAndroidVMHelper vmHelper;
+		vmHelper.callStaticVoidMethod(CAndroidVMHelper::NATIVE_METHODS_DEFAULT_CLASS, "hapticFeedback");
+#endif
+#if defined(VCMI_IOS)
```
kambala-decapitator, line 307:
```suggestion
	if(hapticFeedbackEnabled)
	{
#if defined(VCMI_ANDROID)
		CAndroidVMHelper vmHelper;
		vmHelper.callStaticVoidMethod(CAndroidVMHelper::NATIVE_METHODS_DEFAULT_CLASS, "hapticFeedback");
#elif defined(VCMI_IOS)
```

client/ios/utils.mm
```diff
@@ -43,4 +43,11 @@ void hideLoadingIndicator()
 	[indicator removeFromSuperview];
 	indicator = nil;
 }
+
+void hapticFeedback()
+{
+	UIImpactFeedbackGenerator *hapticGen = [[UIImpactFeedbackGenerator alloc] initWithStyle:(UIImpactFeedbackStyleLight)];
+	[hapticGen impactOccurred];
+	hapticGen = NULL;
```
kambala-decapitator, line 51:
```suggestion
	auto hapticGen = [[UIImpactFeedbackGenerator alloc] initWithStyle:UIImpactFeedbackStyleLight];
	[hapticGen impactOccurred];
```

Laserlicht:
Tested on android and iPhone & is working

config/widgets/settings/generalOptionsTab.json
```diff
@@ -106,6 +110,12 @@
 					"help": "vcmi.systemOptions.longTouchButton",
 					"callback": "setLongTouchDuration",
 					"created" : "touchscreen"
+				},
+				{
+					"name": "hapticFeedbackCheckbox",
+					"help": "vcmi.systemOptions.hapticFeedbackButton",
+					"callback": "hapticFeedbackChanged",
+					"created" : "touchscreen"
```
IvanSavenko, line 118: For consistency - use same condition as in label creation
```suggestion
					"created" : "mobile"
```

client/windows/settings/GeneralOptionsTab.cpp
```diff
@@ -190,6 +194,9 @@ GeneralOptionsTab::GeneralOptionsTab()
 	std::shared_ptr<CToggleButton> framerateCheckbox = widget<CToggleButton>("framerateCheckbox");
 	framerateCheckbox->setSelected(settings["video"]["showfps"].Bool());
 
+	std::shared_ptr<CToggleButton> hapticFeedbackCheckbox = widget<CToggleButton>("hapticFeedbackCheckbox");
+	hapticFeedbackCheckbox->setSelected(settings["general"]["hapticFeedback"].Bool());
```
IvanSavenko, line 198: This checkbox might not exist, e.g. on desktop systems
```suggestion
	if (hapticFeedbackCheckbox)
		hapticFeedbackCheckbox->setSelected(settings["general"]["hapticFeedback"].Bool());
```

IvanSavenko (approved):
Please fix these two issues, but looks OK othervice

config/widgets/settings/generalOptionsTab.json
dydzio0614: unrelated but how do we handle stuff like microsoft surface tablets? :P

IvanSavenko: Same as SDL :)

Generally, touch input should work on any system as long as it is supported by SDL. I see some conflicting info on SDL touch input on Windows, so not sure if it is supported.

Linux-based systems (non-Android) should work, e.g. Steam Deck, but obviously not tested.

IvanSavenko:
Also - perhaps we should move autosaves to *end* of turn? Right now I think we do them at turn start, unlike H3

Laserlicht:
@IvanSavenko Done :)

IvanSavenko:
Just noticed one thing - we play haptic feedback not when popup menu appears, but whenever there is a long touch.

What happens if you try to use long touch on areas that don't have popup menu? For example outside of buttons in main menu?

IvanSavenko:
Yeah, probably only when menu opens. So should be moved into `if (GH.windows().isTopWindowPopup())` block few lines below

PR #2305, opened 2023-07-08
Ability to change autosave limit + set autosave prefix

dydzio0614:
Something got screwed up in launcher settings UI though...


config/schemas/settings.json
```diff
@@ -101,6 +104,18 @@
...
+					"type" : "number",
+					"default": 5
+				},
+				"useSavePrefix" : {
+					"type": "boolean",
+					"default": false
+				},
+				"savePrefix" : {
+					"type": "string",
+					"default": ""
```
IvanSavenko, line 118: Not sure if there is any need of fixed save prefix. After all, this is prefix for *all* saves, see 0 possible use-cases here.

Proposition: remove

client/CPlayerInterface.cpp
```diff
@@ -192,24 +193,39 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
...
+		if(autosaveCountLimit > 0)
+		{
+			cb->save("Saves/" + prefix + "Autosave_" + std::to_string(autosaveCount));
+			autosaveCount %= autosaveCountLimit;
+		}
+		else
+		{
+			std::string stringifiedDate = (cb->getDate(Date::MONTH) < 10
+				? std::string("0") + std::to_string(cb->getDate(Date::MONTH))
+				: std::to_string(cb->getDate(Date::MONTH)))
```
IvanSavenko, line 223: Not sure if such conditional is needed. "123" form is easily recognizeable, but "0123" is not.

I mean, I get that this is to make sorting look good once you reach month 10, but same can be said about reaching month 100.

I'd rather consider altering sorting to account for integer numbering

launcher/settingsView/csettingsview_moc.ui
IvanSavenko, line 1: Will recheck this file in Qt a bit later, just in case.

IvanSavenko:
Also - perhaps we should move autosaves to end of turn? Right now I think we do them at turn start, unlike H3

PR #2308, opened 2023-07-09
Remove client savegames

IvanSavenko:
Up to discussion.

This PR completely removes client savegame files and merges any relevant code with server savegames. Reasoning:
- as of right now, client savegames are unused due to being in completely broken state for years
- proper fix of client-side information saving is certainly preferrable, but I don't see any clear way to do one due to how our serializer is designed
- proper fix of client-side information saving would likely involve moving this information into server savegame file, making client savegames unused in any case
- saves that consist from two parts sometimes confuses players leading to questions or incomplete saves in bug reports

(should not conflict with #2305 )


kambala-decapitator:
Is the client part basically write-only atm?

IvanSavenko:
> Is the client part basically write-only atm?

Yep. And data written into it is unusable - leads to duplicated objects in client state.

DjWarmonger:
By current design Client shouldn't store any information related to the state of game. Only GUI-specific settings, like that issue with hero path: https://github.com/vcmi/vcmi/pull/1663

IvanSavenko:
> By current design Client shouldn't store any information related to the state of game. Only GUI-specific settings, like that issue with hero path: #1663

And it does not. However it stores *pointers* to game state data, such as heroes. And our serializer traverses all reachable pointers and as result - serializes game data into client state.

PR #2320, opened 2023-07-11
Fix building in debug mode

IvanSavenko:
- Fixes (yet again) broken compilation in Debug mode
- Attempt to add one more CI run that uses debug mode to prevent these issues in future

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -76,6 +76,10 @@ jobs:
             os: ubuntu-20.04
             test: 0
             preset: linux-gcc-test
+          - platform: linux
+            os: ubuntu-20.04
+            test: 0
+            preset: linux-gcc-debug
           - platform: mac-intel
             os: macos-12
             test: 0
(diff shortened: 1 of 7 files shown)


kambala-decapitator:
from the code change it's absolutely unclear how/why it builds in release mode

IvanSavenko:
@kambala-decapitator 
In release mode, preprocessor completely removes assert's, so compiler won't "see" code inside assert and won't issue any errors.

rilian-la-te:
@IvanSavenko @kambala-decapitator maybe better to use clang and ubuntu-latest? Or you used dated dependencies and operating system with purpose?

IvanSavenko:
@rilian-la-te 
Ideally - I'd rather use both oldest and newest version, if only to make sure that vcmi can still be compiled with these versions. However not sure about adding new CI runs due to usage limits - sure, it's free, but we do sometimes end up with queued builds for hours due to multiple active PR's.

clang vs gcc - we need to check both, so don't see any point in switching to one or another.

kambala-decapitator:
> @kambala-decapitator
> In release mode, preprocessor completely removes assert's, so compiler won't "see" code inside assert and won't issue any errors.

thanks. I've overlooked the fact that all code changes are inside asserts :)

SoundSSGood:
@IvanSavenko thanks for fixing it. I forget to remove NDEBUG define from preprocessor options.

PR #2321, opened 2023-07-11
Fixes & improvements to tavern/hero pool handling

IvanSavenko:
Changes:
- [x] extract hero pool related code from massive classes CGameState / CGameHandler
- [x] Make hero selection logic same as in H3 - first select class, then select hero
- [x] Implement persistent random seed for taverns, as suggested on Discord
- [x] Fix retreating on 7th day during AI turn (https://bugs.vcmi.eu/view.php?id=2603)
- [x] Fix selection of slot to place retreated heroes (should be random AFAIK)
- [x] Fix retreating of multiple heroes (https://bugs.vcmi.eu/view.php?id=1416)

### DIFF
--- a/CCallback.cpp
+++ b/CCallback.cpp
@@ -270,17 +270,10 @@ void CCallback::recruitHero(const CGObjectInstance *townOrTavern, const CGHeroIn
 {
 	assert(townOrTavern);
 	assert(hero);
-	ui8 i=0;
-	for(; i<gs->players[*player].availableHeroes.size(); i++)
-	{
-		if(gs->players[*player].availableHeroes[i] == hero)
-		{
-			HireHero pack(i, townOrTavern->id);
-			pack.player = *player;
-			sendRequest(&pack);
-			return;
-		}
-	}
+
+	HireHero pack(HeroTypeID(hero->subID), townOrTavern->id);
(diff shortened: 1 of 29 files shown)


dydzio0614:
I am not sure if persistent random seed is good idea to be default behavior without gamesettings activation, since more non-deterministic randomization may be better for standard gameplay?

IvanSavenko:
@dydzio0614 
Don't see much difference to be honest.
The only observable difference would be when constantly reloading the game, which is not "standard gameplay"

rilian-la-te:
@IvanSavenko can you do some QoL improvement - garrison hero hiring (if there is a visitor in town)? I know, this is not OH3, but it is convenient)

Warzyw647:
I agree that constantly reloading is not standard gameplay. And non-deterministic doesn't automatically mean 'easily predictable', which would've had some influence on 'standard gameplay'

dydzio0614:
@rilian-la-te while it is nice, if this isn't very quick to do then there are bunch of more impactful QoL fixes awaiting ATM (like removing hero while he is in town - blocked in H3)

IvanSavenko:
Mostly complete, but not yet tested.
Will test & review code tomorrow.

PR #2323, opened 2023-07-11
Revert apparently broken changes from commits 96420bf, 92dcfd1, c4cf2a1

dydzio0614:
Some stuff that maybe make configurable town buildings function more properly, but break town building construction animations, vcmibuild cheat etc.


IvanSavenko:
Checked code - objectVisited also generates visit query (and makes hero visiting unconditionally, even if hero is already visiting or garrisoned)

In theory, this *might* break buildings that use our configurable system and have multiple options to choose from, but since right now we have 0 test cases for this... sigh...

That being said, code before this PR is definitely broken, especially due to different handling of visiting and garrisoned heroes, so I'd rather merge this as it

IvanSavenko (approved):

PR #2325, opened 2023-07-12
Hero backpack window prototyping

SoundSSGood:
This PR does not provide any new graphic resources. Also the hero window needs a design improvement. But this can already be tested.


lib/GameConstants.h
```diff
@@ -53,6 +53,9 @@ namespace GameConstants
 	constexpr int DEFAULT_SCHOOLS = 4;
 	constexpr int CRE_LEVELS = 10; // number of creature experience levels
 
+	constexpr size_t HERO_BACKPACK_WINDOW_SLOT_COLUMNS = 8;
+	constexpr size_t HERO_BACKPACK_WINDOW_SLOT_LINES = 8;
+
```
IvanSavenko, line 58: Should probably be inside hero backpack class. This is not global constant but local constant

client/widgets/CWindowWithArtifacts.cpp
```diff
@@ -338,10 +360,13 @@ std::optional<std::tuple<const CGHeroInstance*, const CArtifactInstance*>> CWind
 	// The state is possible when the hero has placed an artifact in the ArtifactPosition::TRANSITION_POS,
 	// and the previous artifact has not yet removed from the ArtifactPosition::TRANSITION_POS.
 	// This is a transitional state. Then return nullopt.
-	if(pickedCnt > 1)
+	if(std::accumulate(std::begin(pickedCnt), std::end(pickedCnt), 0, [](size_t accum, const std::map<const CGHeroInstance*, size_t>::value_type & value)
```
IvanSavenko, line 363: Haven't checked, so might be wrong, but:
- you can use boost::range::accumulate for simplicity (if it exists)
- you can use `auto` instead of massive `std::map<const CGHeroInstance*, size_t>::value_type`

client/widgets/CArtifactsOfHeroBackpack.cpp
```diff
@@ -0,0 +1,72 @@
...
+		artPlace->rightClickCallback = std::bind(&CArtifactsOfHeroBase::rightClickArtPlace, this, _1);
+		artPlaceIdx++;
+	}
+	if(backpackCap < 0 || visibleCapasityMax < backpackCap)
+	{
+		auto scrollHandler = std::bind(&CArtifactsOfHeroBase::scrollBackpack, this, _1);
+		leftBackpackRoll = std::make_shared<CButton>(Point(-20, 0), "hsbtns3.def", CButton::tooltip(), [scrollHandler]() { scrollHandler(-1); }, EShortcut::MOVE_LEFT);
+		rightBackpackRoll = std::make_shared<CButton>(Point(368, 318), "hsbtns5.def", CButton::tooltip(), [scrollHandler]() { scrollHandler(+1); }, EShortcut::MOVE_RIGHT);
+		leftBackpackRoll->block(true);
+		rightBackpackRoll->block(true);
```
IvanSavenko, line 52: I assume scroll buttons on 8x8 backpack window would scroll backpack by 1 item? Not by 1 row?

IMO we should scroll by 1 row and possibly use slider in place of buttons

client/windows/CHeroBackpackWindow.cpp
```diff
@@ -0,0 +1,23 @@
...
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#include "StdInc.h"
+#include "CHeroBackpackWindow.h"
+
+#include "../gui/CGuiHandler.h"
+
+CHeroBackpackWindow::CHeroBackpackWindow(const CGHeroInstance * hero)
```
IvanSavenko, line 15: While functionality is definitely something that players requested, but what about visuals?
Are you planning to make them relatively soon, e.g. in week or two?
- If you can make them in time for 1.3 then I am OK with accepting this PR as it
- Alternatively, I can look into this myself when I can find time.

In later case would be great to have backpack disabled by default (macro or hidden setting) until backpack will be made presentable to players.

IvanSavenko (commented):
Only checked code, will test build a bit later.

client/widgets/CArtifactsOfHeroBackpack.cpp
SoundSSGood: It is possible to scroll to any number of items. Including on HERO_BACKPACK_WINDOW_SLOT_COLUMNS. But I will also make scrolling a slider.

client/windows/CHeroBackpackWindow.cpp
SoundSSGood: > but what about visuals?

I am not familiar with make visuals. I'm not sure if I can do it right.

> until backpack will be made presentable to players.

Since the button to open the backpack doesn't look presentable and needs to be redesigned, I planned to remove it altogether before the merge.

client/widgets/CWindowWithArtifacts.cpp
SoundSSGood: > * you can use boost::range::accumulate for simplicity (if it exists)

I prefer to use std instead of boost where possible.

> * you can use `auto` instead of massive `std::map<const CGHeroInstance*, size_t>::value_type`

And thanks for the hint.

IvanSavenko:
Checked build:
- row-by row scrolling is OK, but IMO we should probably use slider instead of buttons (and perhaps - only use it we we have too many artifact to display at once)
- "infinite scroll" mode is interesting, but probably not needed - it would need more graphical assets (for finite scroll we can reuse H3 slider) and is only needed for rather rare case where hero has extremely large number of artifacts in backpack
- there is no way to exit backpack window other than clicking on slot. We probably need button to close it.


Also - noticed bug, not sure if this is regression from this branch or old one (probably old one):
- have equipped artifact that is part of a combined artifact AND modifies primary stat (e.g. Titan's Gladius)
- have combined artifact in backpack (e.g. Titan's Thunder)
- now, if you pick up Titan's Thunder from a backpack (either window or standard H3 backpack) primary stats of your hero would be modified as if you have no Titan's Gladius equipped (even though you've picked artifact from backpack and still have Gladius equipped)

SoundSSGood:
> slider instead of buttons

> need button to close it

Done.

> Also - noticed bug

Will take a look.

IvanSavenko:
Looks to be OK.
I guess can be merged in disabled state. Will look into making proper window UI for this a bit later

SoundSSGood:
> I guess can be merged in disabled state

Ok. I'll remove open button.

dydzio0614:
@SoundSSGood if you look for ideas for some other stuff regarding artifacts... https://github.com/orgs/vcmi/projects/2 has some

SoundSSGood:
done

IvanSavenko (approved):

IvanSavenko:
Will try to make UI for this in time for 1.3

PR #2334, opened 2023-07-15
Battleintro: Skip with mouse button

Laserlicht:


client/battle/BattleWindow.cpp
```diff
@@ -204,6 +204,16 @@ void BattleWindow::keyPressed(EShortcut key)
...
 }
 
+void BattleWindow::clickPressed(const Point & cursorPosition)
+{
+	if (owner.openingPlaying())
+	{
+		owner.openingEnd();
+		return;
+	}
+	InterfaceObjectConfigurable::clickPressed(cursorPosition);
```
dydzio0614, line 214: shouldn't order be reversed (move before "if") so single click won't trigger battlefield action after opening ends?

Laserlicht: It is exact the same logic as for "keyPressed".

I don't understand what you mean. After opening ends the function ignores the action with "return".

dydzio0614: at yeah, nevermind

IvanSavenko (approved):
Will also work for touchscreen / mobiles, where you can't just "press any key"

Laserlicht:
> Will also work for touchscreen / mobiles, where you can't just "press any key"

Like "no keyboard present - press any key to continue..." 😀

PR #2338, opened 2023-07-16
Info box army management

dydzio0614:
Allow creature management in info bar/box/panel - wanted HD mod feature

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -572,7 +572,14 @@ void CPlayerInterface::garrisonsChanged(std::vector<const CGObjectInstance *> ob
 		auto * town = dynamic_cast<const CGTownInstance*>(object);
 
 		if (hero)
+		{
 			adventureInt->onHeroChanged(hero);
+
+			if(hero->inTownGarrison)
+			{
+				adventureInt->onTownChanged(hero->visitedTown);
+			}
+		}
 		if (town)
 			adventureInt->onTownChanged(town);
 	}
--- a/client/adventureMap/CInfoBar.cpp
+++ b/client/adventureMap/CInfoBar.cpp
@@ -51,7 +51,11 @@ CInfoBar::VisibleHeroInfo::VisibleHeroInfo(const CGHeroInstance * hero)
 {
 	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
 	background = std::make_shared<CPicture>("ADSTATHR");
-	heroTooltip = std::make_shared<CHeroTooltip>(Point(0,0), hero);
+
+	if(settings["gameTweaks"]["infoBoxCreatureManagement"].Bool())
+		heroTooltip = std::make_shared<CInteractableHeroTooltip>(Point(0,0), hero);
+	else
+		heroTooltip = std::make_shared<CHeroTooltip>(Point(0,0), hero);
 }
 
 CInfoBar::VisibleTownInfo::VisibleTownInfo(const CGTownInstance * town)
@@ -274,7 +278,12 @@ void CInfoBar::tick(uint32_t msPassed)
 void CInfoBar::clickReleased(const Point & cursorPosition)
 {
 	if(state == HERO || state == TOWN)
+	{
+		if(settings["gameTweaks"]["infoBoxCreatureManagement"].Bool())
+		return;
+
 		showGameStatus();
+	}
 	else if(state == GAME)
 		showDate();
 	else
--- a/client/adventureMap/CInfoBar.cpp
+++ b/client/adventureMap/CInfoBar.cpp
@@ -288,7 +288,7 @@ void CInfoBar::clickReleased(const Point & cursorPosition)
 	if(state == HERO || state == TOWN)
 	{
 		if(settings["gameTweaks"]["infoBarCreatureManagement"].Bool())
-		return;
+			return;
 
 		showGameStatus();
 	}
--- a/client/adventureMap/CInfoBar.cpp
+++ b/client/adventureMap/CInfoBar.cpp
@@ -62,7 +62,11 @@ CInfoBar::VisibleTownInfo::VisibleTownInfo(const CGTownInstance * town)
 {
 	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
 	background = std::make_shared<CPicture>("ADSTATCS");
-	townTooltip = std::make_shared<CTownTooltip>(Point(0,0), town);
+
+	if(settings["gameTweaks"]["infoBoxCreatureManagement"].Bool())
+		townTooltip = std::make_shared<CInteractableTownTooltip>(Point(0,0), town);
+	else
+		townTooltip = std::make_shared<CTownTooltip>(Point(0,0), town);
 }
 
 CInfoBar::VisibleDateInfo::VisibleDateInfo()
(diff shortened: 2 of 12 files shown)


client/widgets/CGarrisonInt.h
```diff
@@ -81,6 +81,13 @@ class CGarrisonInt :public CIntObject
 	bool checkSelected(const CGarrisonSlot * selected, TQuantity min = 0) const;
 
 public:
+	enum class EGarrisonIntSlotsLayout
+	{
+		ONE_ROW,
+		TWO_ROWS,
+		REVERSED_TWO_ROWS
+	};
```
IvanSavenko, line 89: Might be better to move this outside of GarrisonInt (or alternatively - rename to ELayout) - to avoid extremely long `CGarrisonInt::EGarrisonIntSlotsLayout::TWO_ROWS` syntax

IvanSavenko: Also, optionally - integrate smallImgs flag into this enum - since large images is basically just another layout (with larger offsets between images)

client/adventureMap/CInfoBar.h
```diff
@@ -66,14 +69,14 @@ class CInfoBar : public CIntObject
 
 	class VisibleHeroInfo : public CVisibleInfo
 	{
-		std::shared_ptr<CHeroTooltip> heroTooltip;
+		std::variant<std::shared_ptr<CHeroTooltip>, std::shared_ptr<CInteractableHeroTooltip>> heroTooltip;
```
IvanSavenko, line 72: Just use `shared_ptr<CIntObject>` instead of variant

client/adventureMap/CInfoBar.cpp
```diff
@@ -273,8 +282,16 @@ void CInfoBar::tick(uint32_t msPassed)
 
 void CInfoBar::clickReleased(const Point & cursorPosition)
 {
+	timerCounter = 0;
+	removeUsedEvents(TIME); //expiration trigger from just clicked element is not valid anymore
+
 	if(state == HERO || state == TOWN)
+	{
+		if(settings["gameTweaks"]["infoBarCreatureManagement"].Bool())
+		return;
```
IvanSavenko, line 291:
```suggestion
			return;
```

client/widgets/MiscWidgets.cpp
```diff
@@ -357,6 +382,55 @@ CTownTooltip::CTownTooltip(Point pos, const CGTownInstance * town)
 	init(InfoAboutTown(town, true));
 }
 
+CInteractableTownTooltip::CInteractableTownTooltip(Point pos, const CGTownInstance * town)
+		: CGarrisonInt(pos.x, pos.y+73, 4, Point(0, 0), town->getUpperArmy(), nullptr, true, true, CGarrisonInt::EGarrisonIntSlotsLayout::REVERSED_TWO_ROWS)
+{
+	init(InfoAboutTown(town, true));
+}
+
+void CInteractableTownTooltip::init(const InfoAboutTown & town)
```
IvanSavenko, line 391: Perhaps try to reuse common code from non-interactive tooltip to avoid yet another copy-paste of town icon / fort / hall / income? Same question about hero tooltip

Not sure if this is doable, so feel free to ignore

IvanSavenko (commented):
And zuuzel hardships will finally come to an end 😄

client/widgets/MiscWidgets.cpp
dydzio0614: Should be doable by multiple inheritance, but I would rather focus on looking into other features currently

client/adventureMap/CInfoBar.h
dydzio0614: OK, though I think variant reflects the intent better

client/widgets/CGarrisonInt.h
dydzio0614: went for ESlotsLayout to make name more descriptive - check if length satisfies you now

rilian-la-te:
@dydzio0614 is infoboz notifications working with this patch?

dydzio0614:
it worked as far as I checked

rilian-la-te:
@dydzio0614 with dragon utopia too?

dydzio0614:
I tested with crypt and was OK, anyway there should be no reason for this change to break components because it mostly blocks clicking through "hero/town" component and changes how this component itself works

dydzio0614:
anyway based on my testing after this was done, this does not behave 100% as in HD mod (click through is possible if you do it on upper half of component, but I do not think anybody will complain since that functionality has barely any use and default behavior exists via setting

client/CPlayerInterface.cpp
```diff
@@ -572,7 +572,14 @@ void CPlayerInterface::garrisonsChanged(std::vector<const CGObjectInstance *> ob
 		auto * town = dynamic_cast<const CGTownInstance*>(object);
 
 		if (hero)
+		{
 			adventureInt->onHeroChanged(hero);
+
+			if(hero->inTownGarrison)
+			{
+				adventureInt->onTownChanged(hero->visitedTown);
+			}
```
IvanSavenko, line 581: Personal opinion, so OK to keep but for 1-line code block brackets are unnecessary

IvanSavenko (approved):

client/CPlayerInterface.cpp
dydzio0614: hmmm I would need to recheck coding guidelines, but at least back in the day there was rule that if code is part of a block (other than function-level block) then internal part should be wrapped

config/schemas/settings.json
```diff
@@ -529,6 +530,10 @@
 				"skipBattleIntroMusic" : {
 					"type" : "boolean",
 					"default" : false
+				},
+				"infoBarCreatureManagement": {
+					"type" : "boolean",
+					"default" : false
```
IvanSavenko, line 536: By the way - should we keep it on or off by default?

dydzio0614: I thought about "off" to keep compatibility with H3, same as you plan to keep mouse map dragging off

IvanSavenko: Mouse dragging has side effect.
Map click (set path/move hero) with mouse dragging is processed on mouse *release*. This leads to visual lag since it looks like game reacts on click way than in H3 since H3 set path / move hero on mouse *press*.

Infobar change also have side effect (can't click through infobar windows), but this is very rare use-case, so probably not critical.

dydzio0614: did you see how HD mod resolved this issue for double click for right mouse button BTW?

dydzio0614: in HD mod you can click through if you target the uninteractable part (upper half), though it is not even certain if it's better to make it as in HD mod or keep it as is, since misclicks that toggle infobar may be undesired as well

IvanSavenko: Yeah, I've seen it.
Don't think that this is significant issue - infobox is used *extremely* rarely, so likely very few players would be against such change.

So I'd rather keep current (easiest) solution for now and check if players are OK with this.

rilian-la-te: @IvanSavenko infobox with infobox notifications can be useful more than in OH3. So, we need to be sure than army management in infobox is not interfere with infobox notifications.

Honestly, I would rather extend infobox dialog functionality than replicate a HD mod, but now I do not have spare time for hobbies(

PR #2350, opened 2023-07-16
Ban objects on maps without water

DjWarmonger:


DjWarmonger:
I actually wonder how to handle this in custom scenarios. Mapmakers may want to allow some heroes deliberately, so we can't force ban them on map load.

IvanSavenko:
Well, if we want "hota style" approach then it looks to be only used on RMG maps based on "water content" option.
If we want to extend this to single scenarios then we can have this autodetection in H3M maps, and explicit flag in VCMI maps (editable by map maker). Or only use flag on VCMI maps and remove autodetection completely.

IvanSavenko (commented):
- Please update json validation schemas
- I don't see handling of Navigation skill ban. Intended?

rilian-la-te:
It should be optional feature, I think. In OH3 there is no ban for Navigation and related heroes. I think HoTA is evil, but I understand than some players want HoTA features. So, I think this should be optional non-default feature, when HoTA mod is disabled.

IvanSavenko:
Merge?

Issue #2319, opened 2023-07-11
Display hero info outside of battlefield area


DjWarmonger:
Now: 
![image](https://github.com/vcmi/vcmi/assets/2566990/486d68f2-1c66-43fe-a7a2-fea032036bcd)

How it should be preferably, if there's enough space:
![image](https://github.com/vcmi/vcmi/assets/2566990/a4f470df-6864-42d3-86d5-fc2f43c9d2e8)xt about the problem here.

misiokles:
This is hd-mod solution. Maybe some other? More unique? Semi-transparent after pointing mouse/finger at hero?

dydzio0614:
@misiokles I do not see reason to reinvent the wheel for something that is supposed to be primarily practical feature, unless somebody else comes with functionally better design

Laserlicht:
@DjWarmonger can be closed...

PR #2358, opened 2023-07-17
Add option for persistent hero info window as in HD mod

dydzio0614:
Implements #2319 with "i" keyboard shortcut toggle. Due to VCMI having bigger window, visually it is not 100% alignment, but position should be more or less same as in HD mod. For resolutions < 1000 px wide it uses same position as in original game.

Also default popup position even on right click is now as in HD mod for resolutions < 1000 px wide


Mods/vcmi/config/vcmi/english.json
```diff
@@ -113,6 +113,8 @@
 	"vcmi.battleOptions.movementHighlightOnHover.help": "{Movement Highlight on Hover}\n\nHighlight unit's movement range when you hover over it.",
 	"vcmi.battleOptions.rangeLimitHighlightOnHover.hover": "Show range limits for shooters",
 	"vcmi.battleOptions.rangeLimitHighlightOnHover.help": "{Show range limits for shooters on Hover}\n\nShow shooter's range limits when you hover over it.",
+	"vcmi.battleOptions.showStickyHeroInfoWindows.hover": "Permanently show heroes statistics windows",
```
IvanSavenko, line 116: Perhaps drop "Permanently" from name?
Text already rather long, might be even longer with translations

client/battle/BattleWindow.cpp
```diff
@@ -170,6 +207,49 @@ void BattleWindow::showQueue()
...
+		return;
+
+	Settings showStickyHeroInfoWIndows = settings.write["battle"]["stickyHeroInfoWindows"];
+	showStickyHeroInfoWIndows->Bool() = true;
+
+//	if(attackerHeroWindow)
+//		attackerHeroWindow->enable();
+//
+//	if(defenderHeroWindow)
+//		defenderHeroWindow->enable();
```
IvanSavenko, line 247:
```suggestion

```

Issue #1323, opened 2022-12-27
Make swap armies instant. (its laggy ATM)


krs0:
When swapping full army between 2 heroes with full slots, the lag is obvious.
Proposal: Update GUI only once for the final result instead of for each moved creature stack?

![image](https://user-images.githubusercontent.com/15194321/209670456-6b3265b2-29b7-4cce-b829-60c5f7c7f239.png)

IvanSavenko:
Largely fixed in my PR.
There is still a small lag, but way better than before.

PR #2359, opened 2023-07-18
Fix crashes from 1.2.1 statistics on Google Play

IvanSavenko:
Attempt to resolve or at least - add some better error checking for several crashes found in Android statistics.

Largest source of crashes seems to be:
- *extremely* common crash on BattleAI taking action, as well as on activation of quick combat.
- NewObject::applyGs (likely already fixed in boat refactoring)
- few more unknown crashes with unavailable stack trace due to throwing exception from catch block on server

Progress so far:
- remove "catch-all" block from server that does not provides any useful info, but successfully clears any stacktrace in case of exception
- attempt to fix crash on NewObject::applyGs
- refactored BattleAI::activeStack() method into smaller ones, to localize crash (Google Play only shows stacktrace, without locals or line numbers)
- removed async execution from ExchangeWindow, to simplify our threading model
- implemented lazy total redraws to prevent multiple totalRedraws on one frame
- fixes (possibly - only partially) #1323 

TODO: 
- fix (newly discovered bug) that makes it impossible to cancel autocombat
- try to simplify threading model of client/netpack handling to fix extremely common crash on AI action (autocombat likely)

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -93,6 +93,162 @@ void CBattleAI::initBattleInterface(std::shared_ptr<Environment> ENV, std::share
 	movesSkippedByDefense = 0;
 }
 
+BattleAction CBattleAI::useHealingTent(const CStack *stack)
+{
+	auto healingTargets = cb->battleGetStacks(CBattleInfoEssentials::ONLY_MINE);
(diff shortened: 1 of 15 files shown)


IvanSavenko:
Should be fixed,
After 1.3 release will check stats in Google Play for any new/unresolved crashes.

PR #2362, opened 2023-07-19
Implemented support for showing amount of guards in banks

IvanSavenko:
- Visiting banks will now show up guards composition of object
- Only visiting is handled at the moment, hover text is unaffected (will look into it for 1.4, alongside with general review of object scouting logic)
- Guards preview are precise (and don't use ranges) - easier to implement and in a lot of cases unnecessary - since you often can guess precise numbers due to guards having few possible settings.
- For now this option is turned on by default, can be moved to mod if requested.


krs0:
I predict lots of discussions because of precise guards :)

PR #2364, opened 2023-07-19
Mapselection - popup

Laserlicht:
another HD thing...

grafik
grafik

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -38,6 +38,9 @@
 	"vcmi.mainMenu.hostTCP" : "Host TCP/IP game",
 	"vcmi.mainMenu.joinTCP" : "Join TCP/IP game",
 	"vcmi.mainMenu.playerName" : "Player",
+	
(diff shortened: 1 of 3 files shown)


dydzio0614:
Looks promising

Laserlicht:
Added some descriptions.

`std::format` has minimum requirement of C++20. Also there is some condition... :(

grafik

DjWarmonger:
Yes, but we've got boost::format for that.

DjWarmonger:
Also, if it's from HD mod, you may want to add it to the list: https://github.com/orgs/vcmi/projects/2

krs0:
Filename:  
And
Creation Date:
Are just noise in that popup. The information can be understood without that extra words.

At minimum Creation Date could be replaced with with Last Save: or Save Date:

IvanSavenko:
@krs0 Perhaps. But it makes this dialog look a little bit organized and not look like randomly placed set of words.

@DjWarmonger Yeah, it is, but I think only people with write access can add issue to project. Added.

IvanSavenko (approved):

PR #2384, opened 2023-07-22
Bugfixing iteration

IvanSavenko:
Issues to investigate:
- [x] battle hangs on start if enemy AI has tactics phase
- [x] Autocombat can be deactivated (missing commit from previous PR)
- [x] Fixed activation of radial wheel when using touchscreen input
- [x] Game crash on ending battle in a draw
- [x] Option "keepDisabled" does not correctly shows up in Launcher
- [x] sometimes enemy hero tile is visited with overlapping (sharing same tile) so both attacker and defender are seen on same tile
- [x] enchanters lose their creature count label when it's their turn, probably related to casting spell before attack

To be done later:
- [ ] creatures walking animation looks choppy when decreasing FPS to 30 -  can be seen by looking at nomad legs during animation in H3 and VCMI
- [ ] clicking between real and borderless fullscreen in options (one, then another etc.) leads to both being turned off at some point 
- [ ] Switching to world view and back breaks adventure map - no date display.
- [ ] Visiting Magi Hut - hero movement sound plays while terrain is being revealed
- [ ] water under ship covering land tile
- [ ] Error on accepting quick combat results or fighting against AI - non-existing query

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -251,7 +251,7 @@ BattleAction CBattleAI::selectStackAction(const CStack * stack)
 
 void CBattleAI::yourTacticPhase(int distance)
 {
-	cb->battleMakeUnitAction(BattleAction::makeEndOFTacticPhase(cb->battleGetTacticsSide()));
+	cb->battleMakeTacticAction(BattleAction::makeEndOFTacticPhase(cb->battleGetTacticsSide()));
 }
 
 void CBattleAI::activeStack( const CStack * stack )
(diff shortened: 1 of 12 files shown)


dydzio0614:
BTW heroes sharing same tile also works for army exchange on adventure map

IvanSavenko:
Presuming ready, most critical issues are fixed, other stuff can wait a bit.

dydzio0614:
I would say green messages bug is rather important, at least for player reception - and may be related to crash after battle in spectator mode

IvanSavenko:
Yes, but not critical for gameplay.
Will look into it (and other issues) before 1.3, but right now I'd rather merge critical stuff than wait for more fixes

PR #2386, opened 2023-07-22
Wrong placement of onlyOnWaterMap in spells schema.

krs0:


DjWarmonger:
You may want to change actual spell config as well.

krs0:
@DjWarmonger I do not get it. In spell config it was in right place, only schema was wrong. Or?

DjWarmonger:
Ahh you're right, my bad. Schema has 1 more level of nesting than actual config.

PR #2390, opened 2023-07-22
Fix for hero info window mana points not getting spent on spellcast - option 1

dydzio0614:
First proposition to approach mana not refreshing in hero info window - poke window on mana used. Does not handle imp mana steam and spells cheat yet, but these are niche anyway. Somewhat resembles approach that would be used in observer pattern, except that battle interface deals with window refresh in its own code

Pick or close this PR, depending on which option is better, compared with #2391

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -476,6 +476,9 @@ void CPlayerInterface::heroManaPointsChanged(const CGHeroInstance * hero)
 	adventureInt->onHeroChanged(hero);
 	if (makingTurn && hero->tempOwner == playerID)
 		adventureInt->onHeroChanged(hero);
+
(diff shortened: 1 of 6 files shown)


IvanSavenko:
Haven't checked it, but you should be able to track changes easily by handling `SetMana` netpack - `CPlayerInterface::heroManaPointsChanged` method

So
1) Add heroManaPointsChanged method to BattleWindow that will call required method in your widget
2) In CPlayerInterface::heroManaPointsChanged call this method using code like this one:
```
	for (auto window : GH.windows().findWindows<BattleWindow>())
		cgh->heroManaPointsChanged();
```

This *should* handle any sort of mana points change, no matter its origin. And if not - then I'd call this a bug

IvanSavenko (approved):

Issue #2389, opened 2023-07-22
The whole map is revealed to the human player when playing in one team with computer player (Nullkiller AI)


pppbb:
As i wrote player(human) in one team with computer player is seeing the whole map after computer's turn either in single player and multiplayer games. Observed only with Nullkiller AI.

1. Start single scenario or multiplayer games choosing Nullkiller AI as Adventure map AI.
2. Place yourself in a team with computer 
3. Run random map.
4. Wait until computer makes his turn.
The whole map should be visible to the team.

**Did it work earlier?**
I don't know.

**Version**
 - OS: Arch Linux, Windows, Android 
 - Version: 1.2 -1.3

IvanSavenko:
Known issue.
Unfortunately Nullkiller AI currently requires map reveal to properly function.

In 1.3 we've added option to set AI of allied players to a different AI, which by default set to VCAI

paddotk:
@IvanSavenko Is that a temporary solution? As far as I know, the VCAI doesn't work well (mostly idle heroes), so it's not an ideal permanent solution.

IvanSavenko:
Depends on time & priorities of our AI developer.

From what he told me, proper support for exploration of NKAI is not trivial and relatively low on priority list.

PR #2405, opened 2023-07-24
#2389 - treat dwelling defenders as neutrals

nullkiller:
Treat dwelling defenders as neutrals always even if the dwelling is owned.


lib/mapObjects/IObjectInterface.cpp
```diff
@@ -107,6 +107,10 @@ int3 IBoatGenerator::bestLocation() const
 IBoatGenerator::EGeneratorState IBoatGenerator::shipyardStatus() const
 {
 	int3 tile = bestLocation();
+
+	if(!tile.valid())
+		return TILE_BLOCKED; //no available water
```
nullkiller, line 112: Fix console error when not possible to put a ship.

nullkiller:
This is not a part of 1.3 milestone so far but the thing is more or less isolated and was tested. I think it can be merged but can wait 1.3.1 or whatever

DjWarmonger:
Nah, sounds more like a bugfix, which should get merged in beta.

PR #2414, opened 2023-07-25
Fix autocombat crash

IvanSavenko:


IvanSavenko:
Fixed!

And it looks that this is the very same crash that we have in huge numbers on Android, so should help a lot with stability.

Issue #2269, opened 2023-06-30
Possibility to define battle graphics during fight inside dwellings and creature banks. 


misiokles:
**Describe your proposal**
There's ERA3 mod that allows to set specific battle graphics not only during fights at given terrain that also during fights inside dwellings or creature banks.
For example: fight at Nomad Tent.
![image](https://github.com/vcmi/vcmi/assets/29593042/fb6bd2d8-eb1f-4838-8fd9-4c1184280451)

**Additional context**
It would be cool to define such graphics as external mod.
Proposal code:
```
{
	"core:nomadTent": {
		"battleGround" : ["nomadTent.png"]
	}
}
```

Macron1Robot:
it is already implemented, i think

misiokles:
Just this?
```
{
	"core:basiliskPit": {
		"battleground" : "fortress/basilisk_pit"
	}
}
```

Macron1Robot:
No, you must first define "battleground" object

misiokles:
Eh, I'm too dump to configure this :/

nullkiller:
Yes it should already work. I can try to create a sample mod for this

nullkiller:
Abandoned mine right now have battleground "subteranian" always. Same way it is possible to set battleground for any object.
It needs to be defined in battlegrounds.json separately

nullkiller:
I can not reference a battlefield from mod in core mod object. I think it is known issue. So do not know why this restriction was introduced if I externally show that I want identifier from another mod. Maybe it is possible to patch code somehow to disable this restriction. Not so good at modding

nullkiller:
So making
```
	"core:creatureBank": {
		"index" :16,
		"handler": "bank",
		"types": {
			"dwarvenTreasury": {
				"index" : 1,
				"battleground": "snow_mountains",
			}
		}
	}
```
is ok but a custom battlefield - can not resolve. not with prefix nor without it

IvanSavenko:
> I can not reference a battlefield from mod in core mod object. I think it is known issue. So do not know why this restriction was introduced if I externally show that I want identifier from another mod. Maybe it is possible to patch code somehow to disable this restriction. Not so good at modding

This is deliberate change and intended. Otherwise game will crash if you run game without referenced mod. If you need to access another mod add it to dependencies. (although not possible with core mod / h3 data)

> is ok but a custom battlefield - can not resolve. not with prefix nor without it

Bug. Can check it out.

[...]

PR #2416, opened 2023-07-25
Fix resolving of battlefield identifiers

IvanSavenko:
Fixes, possibly - partially (have only fixed issue reported in comments) #2269

@nullkiller Your mod works for me with this change

### DIFF
--- a/lib/mapObjectConstructors/AObjectTypeHandler.cpp
+++ b/lib/mapObjectConstructors/AObjectTypeHandler.cpp
@@ -13,6 +13,7 @@
 
 #include "IObjectInfo.h"
 #include "../CGeneralTextHandler.h"
+#include "../CModHandler.h"
 #include "../VCMI_Lib.h"
 #include "../mapObjects/CGObjectInstance.h"
 #include "../mapObjects/ObjectTemplate.h"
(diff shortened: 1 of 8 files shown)


lib/mapObjectConstructors/AObjectTypeHandler.cpp
```diff
@@ -95,10 +96,15 @@ void AObjectTypeHandler::init(const JsonNode & input)
...
-		battlefield = input["battleground"].String();
-	else
-		battlefield = std::nullopt;
+	battlefield = BattleField::NONE;
+
+	if(!input["battleground"].isNull())
+	{
+		VLC->modh->identifiers.requestIdentifier("battlefield", input["battleground"], [this](int32_t identifier)
+		{
+			battlefield = BattleField(identifier);
```
nullkiller, line 105: Is it some kind of delayed operation?

DjWarmonger: Yes, all mod identifiers are resolved only after every mod is loaded. This allows mods to be dependent of others.

IvanSavenko: It is not for mod load order, but for object load order.
Delayed callbacks here allow loading of objects even with circular references. 

For example, hero class has reference to faction it belongs to. And faction has reference to chances of each hero class in tavern.
We can't load hero class without loading faction first. But we also can't load faction without loading hero class first.

With these delayed callbacks all objects can be loaded in any order and all references to other objects will be set once loading is complete.

nullkiller (approved):

PR #2419, opened 2023-07-26
Multiplayer adventure map responsiveness

### LINKED ISSUE #2223: Adventure map scrolling doesn't work during multiplayer opponent turn
Map position is frozen the moment you click end turn and unlocks only when it's your turn again

**Game logs**
Please attach game logs: `VCMI_client.txt`, `VCMI_server.txt` etc.

**To Reproduce**
1. Start multiplayer match
2. End turn
3. Map scrolling does not work

**Expected behavior**
Map scrolling works

**Actual behavior**
Nothing happens when you move mouse towards border

**Did it work earlier?**
Works in 1.2

**Version**
 - OS: Windows 10
 - Version: 1.3 recent daily build

**Additional info**
For AI opponents should work as it currently does, changed behavior should occur only when waiting for human opponents

dydzio0614:
Unblock some basic actions during human opponent turn - in practice works as intended. Code changes do not affect singleplayer gameplay. Minimap interaction still blocked, but it's a major improvement even without that.

Fixes #2223


client/adventureMap/AdventureMapShortcuts.cpp
```diff
@@ -36,6 +36,14 @@
...
 #include "../../lib/pathfinder/CGPathNode.h"
 
+bool isCurrentPlayerHuman()
+{
+	PlayerColor currentPlayer = LOCPLINT->cb->getCurrentPlayer();
+	bool isHuman = LOCPLINT->cb->getStartInfo()->playerInfos.count(currentPlayer)
+				   && LOCPLINT->cb->getStartInfo()->playerInfos.at(currentPlayer).isControlledByHuman();
+	return isHuman;
+}
+
```
IvanSavenko, line 46: FYI - if you really want free function it must be declared as `static` - to have local visibility (this file only)

client/adventureMap/AdventureMapShortcuts.cpp
```diff
@@ -461,7 +469,13 @@ bool AdventureMapShortcuts::optionSidePanelActive()
 	return state == EAdventureState::MAKING_TURN || state == EAdventureState::WORLD_VIEW;
 }
 
+bool AdventureMapShortcuts::optionMapScrollingActive()
+{
+	return state == EAdventureState::MAKING_TURN || state == EAdventureState::WORLD_VIEW || (state == EAdventureState::ENEMY_TURN && isCurrentPlayerHuman());
```
IvanSavenko, line 474: I would rather split ENEMY_TURN into ENEMY_AI_TURN and ENEMY_PLAYER_TURN and check accordingly
(feel free to pick better names - e.g. "enemy" here can actually be allied player, it basically "any other player")

IvanSavenko (commented):
Probably a better idea to split ENEMY_TURN in two different states.
Method that sets this state (enemyTurnStarted or something like that) also receives AI flag, so you don't need your function.

IvanSavenko:
Also, things to check:
- what happens if you try to left-click during turn of another player? Adventure map, town list, hero list.
- what happens if you left-click all of that when another player is making turn AND you can see that movement?

dydzio0614:
nothing happens during left click so far

IvanSavenko (approved):

PR #2432, opened 2023-07-27
(develop) Remove boost::interprocess usage

IvanSavenko:
Another boost library that we use for no reason.


kambala-decapitator:
what was IPC used for? What problem did it solve?

IvanSavenko:
Client used it to determine when server is ready to accept connections. End.
No idea why we couldn't just wait for connection to be accepted. Fun part - apart from scary red message waiting for a connection works even before this PR.

We also had issues with it before:
 - if player runs game as root on Linux and server crashes shared memory will be only accessible to root, breaking non-root game runs
 - shared memory is not supported by all systems so we have multiple ifdefs with alternative solutions for all platforms

dydzio0614:
note that such things may be remnant of past designs or already reworked mechanics

PR #2452, opened 2023-07-31
(develop) Replaced boost concurrency mechanisms with std

IvanSavenko:
This is wip, partially to see what issues this will bring and what edge cases we need to resolve.
Not 100% sure if I will finish it, since I need to find a way to replace all functionality, including pieces not available in std

Progress:
- [x] replace boost::thread with std equivalent
- [x] replace boost::mutex and friends with std equivalents
- [x] replace boost::chrono with std equivalents (since it is mostly used as timer for threads)
- [x] replace boost::to_string with boost::str / std::to_string (was likely included indirectly, via one of removed boost includes)
- [ ] replace timed_join in `CServerHandler::stopServerConnection` (no std equivalent)
- [x] replace boost::thread_specific_ptr with thread_local variables in AI libraries (std equivalent uses different syntax)
- [x] replace boost thread interruption mechanism with something else (no std equivalent, right now only present as placeholders in vstd)
- [ ] Test everything that might be affected


IvanSavenko:
Closing / dropping this one.
Don't see this as critical in any way and don't see any clear way to solve remaining issues:

- boost thread supports interruption points for which there is no clear way to replace them

- we use `BOOST_THREAD_DONT_PROVIDE_THREAD_DESTRUCTOR_CALLS_TERMINATE_IF_JOINABLE` option which does not matches c++11 / c++17 behavior - std thread calls terminate on attempt to call destructor on still joinable object, but we expect destructor to detach thread. Fixable, but needs looking through codebase & extensive testing.

Will instead submit smaller PR to make our code more in line with std API (so theoretical migration will be simple find & replace), but keep using boost

PR #2456, opened 2023-08-01
Added workaround for running game with outdated extras mod from 1.2

IvanSavenko:
Can't say that I'm happy with this change, but don't see any other practical approach. If somebody can offer solution that will prevent these issues in future that would be great.

Problem:
- If player won't update vcmi-extras mod and keep using one from 1.2, buttons on battlefield screen will be broken
- We have no way to block usage of outdated mod without having player update it
-Situation is even worse on Android where there is no notifications of outdated mods & no way to update mod other than reinstall

Simple solution that fixes this issue is to make sure that old mod will have no effect on game. For example - rename widget json to something else, so old mod override it.

Solutions that could've worked in theory, but not doable in practice:
- All mods must be explicitly marked as compatible with 1.3. Best option, but:
- - requires changes in both Qt and Android launchers 
- - for simple mods that are compatible with both game versions this would require pointless update that only changes compatibility flag
- - it is not clear how to handle development process. When should mods be marked as compatible with 1.4? Immediately after 1.3 release? But 1.4 might get more modding changes. Before 1.4 release? But how to use mods in development builds?

- Mods must use existing "compatibility" field:
- - does not works for future versions - if we discover that existing mod version for 1.2 no longer works in 1.3 and mark it as such, player won't be aware of this information until he actually updates a mod
- - marking mod as compatible with only existing versions with explicit max version leads to the same issues as previous solution

- use different game data directory for each major version:
- - will require players to find & reinstall all mods they were using before.
- - will reset all game settings player might have changes
- - some players have H3 data in this directory, including those who use Launcher for data import, forcing them to re-import data on update


SoundSSGood:
> * All mods must be explicitly marked as compatible with 1.3.
> * for simple mods that are compatible with both game versions this would require pointless update that only changes compatibility flag.

What if we specify the minimum required version, instead of compatibility 1.3 flag? Mods with no version requirements (by default) will be compatible with any version. 

> * But how to use mods in development builds?

It looks like developer builds should be somehow handled separately. For example, do not pass a compatibility check.

SoundSSGood:
What happens if the mod is not loaded at all? Perhaps if VCMI 1.3 detects an outdated mod, it should ignore it?

IvanSavenko:
> Perhaps if VCMI 1.3 detects an outdated mod, it should ignore it?
Yes it should. Problem is - how to indicate that mod is outdated *before* we know it?

Players with VCMI 1.2 already have VCMI Extras 3.3.6. When 3.3.6 was released, we didn't knew yet that it will be incompatible with 1.3.

How should be define that VCMI 1.3 requires newer version of VCMI Extras?
We can't release VCMI Extras 3.3.7 with compatibility flag - players might not install it and upgrade to vcmi 1.3 without updating mods.

And even if we start requiring explicit list of compatible vcmi versions in mods - relatively few mods actually requires update to newer versions. Extras usually do, towns/new content mods might require update. But rest of mods don't require updates.
Requiring such explicit compatibility list would only add more work for modders and might lead to mods getting abandoned simply because nobody bothered with updating compatibility list.

SoundSSGood:
> And even if we start requiring explicit list

Not list, just optional field.

IvanSavenko:
> Not list, just optional field.

We have one already. "compatibility".
Problem is, when mod is released for VCMI version 1.X nobody has any idea whether it will be compatible with version 1.X+1

When we released vcmi-extras for 1.2 we have not marked it as incompatible with 1.3 because well, it was compatible back then. But now it is not compatible. But marking vcmi-extras for 1.2 as incompatible with 1.3 *now* will do nothing because players already have version which is not marked as such:
- we release vcmi-extras for 1.2
- player installs vcmi-extras for 1.2
- we notice that vcmi-extras is not compatible with 1.3 and release updated version that is marked as such
- players installs 1.3, does NOT checks out repository, does NOT updates mod and immediately starts the game. His vcmi-extras is not marked as incompatible with 1.3, so vcmi-extras loads and as result leads to bugs in game.

rilian-la-te:
VCMI can be updated by package manager, and even patched out version number, so, I think, we should update mods, not VCMI.

IvanSavenko:
Merging as it, will consider mod update notifications for 1.4

PR #2458, opened 2023-08-01
Fix crash at water tavern

DjWarmonger:
- Fix crash when recruiting hero at water tavern
- Fix issue with ghost boat blocking the tile under hero
- Extra checks and tweaks

### DIFF
--- a/lib/NetPacksLib.cpp
+++ b/lib/NetPacksLib.cpp
@@ -1382,9 +1382,8 @@ void HeroRecruited::applyGs(CGameState * gs) const
 		auto * boat = dynamic_cast<CGBoat *>(obj);
 		if (boat)
 		{
-			h->boat = boat;
-			h->attachTo(*boat);
-			boat->hero = h;
+			gs->map->removeBlockVisTiles(boat);
+			h->attachToBoat(boat);
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


lib/NetPacksLib.cpp
```diff
@@ -1419,9 +1418,8 @@ void GiveHero::applyGs(CGameState * gs) const
 		auto * boat = dynamic_cast<CGBoat *>(obj);
 		if (boat)
 		{
-			h->boat = boat;
-			h->attachTo(*boat);
-			boat->hero = h;
+			gs->map->removeBlockVisTiles(boat);
```
IvanSavenko, line 1421: Is this correct?
Here boat blockvis tiles are *removed* before attaching
But when we add boat on map init (CGameState.cpp change) its blockvis tiles are *added* before attaching

DjWarmonger: If anything, the GameState is not correct as tiles will be double-blocked by hero and his boat.

DjWarmonger: Tested with HoTA map which has a water Prison and it works fine as well.

nullkiller: Looks like the question is if we should remove blockvis tiles of boat or of hero?

nullkiller: Have you tested this hero disembarking and embarking on this boat later? And visiting this hero by another hero?

DjWarmonger: I tried embarking, but not heroes visiting each other on boats, good idea.

DjWarmonger:
All works for me.

PR #2467, opened 2023-08-03
Build fix for beta

IvanSavenko:
See CI results in #2465

Looks like Github runners got an update in a *perfect* time for us.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -279,6 +279,10 @@ if(MINGW OR MSVC)
 	endif(MSVC)
 
 	if(MINGW)
+
+		# Temporary (?) workaround for failing builds on MinGW CI due to bug in TBB
+		set(CMAKE_CXX_EXTENSIONS ON)
+
 		set(SYSTEM_LIBS ${SYSTEM_LIBS} ole32 oleaut32 ws2_32 mswsock dbghelp bcrypt)
 
 		# Check for iconv (may be needed for Boost.Locale)


IvanSavenko:
Seems to be working, so merging.

In case if somebody wants more info:
- error message was: `/include/oneapi/tbb/detail/_machine.h:248:56: error: ‘_MCW_DN’ was not declared in this scope`
- According to what I found, this is because of `__STRICT_ANSI__ ` define is present, so mingw/gcc will not declare `_MWC_DN`
- This define is add when `-std=c++17` is used (as opposed to `-std=gnu++17`)
- CMake seems to have changed behavior of `CMAKE_CXX_EXTENSIONS` flag (that defines c++/gnu++ flag) in 3.27.1
- Github CI runner now uses CMake 3.27.1

In the end this *might* be mingw bug in the end which was fixed in Mingw 8.0.1, but Github CI has 8.0.0, so no luck here.

I guess we can safely remove this workaround once Github runners get mingw upgrade. Will see.

IvanSavenko:
Do we? In that case no idea why it suddenly broke.
I definitely noticed CMake version change (3.27.0 -> 3.27.1) in logs, but that was the only change that could've affected this.

rilian-la-te:
From our script:

```
curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-common_10.0.0-2_all.deb \
  && sudo dpkg -i mingw-w64-common_10.0.0-2_all.deb;
curl -O -L http://mirrors.kernel.org/ubuntu/pool/universe/m/mingw-w64/mingw-w64-x86-64-dev_10.0.0-2_all.deb \
  && sudo dpkg -i mingw-w64-x86-64-dev_10.0.0-2_all.deb;
```

This is mingw 10.0 packages from ubuntu 20.10. Maybe they updated package version and download cannot be completed now?

PR #2468, opened 2023-08-03
Try running CI CMake builds in parallel

IvanSavenko:
According to Github docs, CI runners have 2 cores (or 3 in case of MacOS runners). But we always run cmake in 1 thread.

Checking whether running CMake in multiple threads would improve CI run times


kambala-decapitator:
ninja / Xcode / msbuild build on all cores by default AFAIK, there's no need for this setting

IvanSavenko:
Unfrtunately, @kambala-decapitator is right. No significant improvement of CI build times.

PR #2484, opened 2023-08-05
NKAI: fix freeze on army gathering

nullkiller:
Proper fix for a freeze when visiting and garrisoned heroes want to exchange.

### DIFF
--- a/AI/VCAI/VCAI.cpp
+++ b/AI/VCAI/VCAI.cpp
@@ -1359,7 +1359,7 @@ void VCAI::wander(HeroPtr h)
 
 	TimeCheck tc("looking for wander destination");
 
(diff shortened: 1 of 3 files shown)


AI/VCAI/VCAI.cpp
```diff
@@ -1359,7 +1359,7 @@ void VCAI::wander(HeroPtr h)
 
 	TimeCheck tc("looking for wander destination");
 
-	while(h->movementPointsRemaining())
+	for(int k = 0; k < 10 && h->movementPointsRemaining(); k++)
```
nullkiller, line 1362: One more hack to avoid complete freezes in VCAI

dydzio0614:
maybe this should target beta to be included in 1.3.1?

PR #2486, opened 2023-08-05
Fix ship_to_ship battlefield resolution

nullkiller:
ship_to_ship resolution fails. Battle between to heroes on water crashes


nullkiller:
Already fixed

PR #2497, opened 2023-08-06
Map full name and locale date

### LINKED ISSUE #2365: Datetime format
**Describe the bug**
Datetime format is `Www Mmm dd hh:mm:ss yyyy` independent from locale (currently `asctime` is used).

**Expected behavior**
Date should follow system-format.

**Screenshots**
grafik

Laserlicht:
Shows the full path to the map with file extension.

Also uses system date&time output format now. Fixes #2365

### DIFF
--- a/lib/mapping/CMapInfo.h
+++ b/lib/mapping/CMapInfo.h
@@ -62,7 +62,6 @@ public:
 		h & campaign;
 		h & scenarioOptionsOfSave;
 		h & fileURI;
-		h & fullFileURI;
 		h & date;
 		h & amountOfPlayersOnMap;
 		h & amountOfHumanControllablePlayers;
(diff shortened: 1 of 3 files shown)


lib/mapping/CMapInfo.h
```diff
@@ -61,6 +62,7 @@ class DLL_LINKAGE CMapInfo
 		h & campaign;
 		h & scenarioOptionsOfSave;
 		h & fileURI;
+		h & fullFileURI;
```
Laserlicht, line 65: @IvanSavenko Is it correct to add the var here?

IvanSavenko: Yes, however this affects save format so please also:
- bump SERIALIZATION_VERSION by one (in CSerializer.h)
- add `if (version >= XXX)` check here (where XXX is new value of serialization version)

Laserlicht: The variable does not actually have to be in the savefile. Can I omit it then?

Laserlicht:
included in #2517

PR #2517, opened 2023-08-08
Folder Support

### LINKED ISSUE #2489: Maps within subfolders
**Describe your proposal**
It should be possible to use folders for maps (and custom campaigns). Like the HD mod.

**Additional context**
With this we can avoid a mix up between maps from different map packs from mods (example: [Mighty Heroes III](https://github.com/vcmi-mods/mighty-heroes-iii/tree/vcmi-1.3)). Also the user can sort his maps in folders.

Maybe also for save games (autosave; see HD mod).

Laserlicht:
Currently only Savegame and Newgame.

Needs some optimization. Due lack of C++ Skills I'm open for suggestions.

fixes #2489

**Todo**
- [x] Campaign
- [x] Save File -> Save to current Path
- [x] Ignore sorting for folders (always on the top)
- [x] Check Filerings

-> Autosave is out of scope -> maybe seperate PR?


Nordsoft91:
Crash when go level up:

https://github.com/vcmi/vcmi/assets/9308612/fe6b75cd-1241-49c8-8797-fb24dd9cc36a


Suggestions:
 - add folder icon for folders
 - store autosaves in separate folder (add feature flag somewhere in settings.json)

Laserlicht:
@Nordsoft91 Can you recheck?

Nordsoft91:
@Laserlicht , checked new version, same. Crash even if I select ... with keyboard arrows.

client/lobby/SelectionTab.cpp
```diff
@@ -421,6 +498,29 @@ void SelectionTab::select(int position)
 	else if(position >= listItems.size())
 		slider->scrollBy(position - (int)listItems.size() + 1);
 
+	if(curItems[py]->isFolder) {
+		if(curItems[py]->folderName == "..")
+		{
+			std::vector<std::string> filetree;
+			boost::split(filetree, curFolder, boost::is_any_of("/"));
+			filetree.erase(filetree.end());
+			filetree.erase(filetree.end());
```
Nordsoft91, line 507: end() points at element after last, I think this is a reason of crash. If you want to remove last element from vector, I propose to use pop_back() method.

Laserlicht: I've changed it. Can you do a retest?

I cannot reproduce this crash. Tried it with Ubuntu and Windows. Also tried some special Chars.

Maybe MacOS specific?

Laserlicht: Update: Was able to reproduce it on android. Fix remove issue. Should work now. :)

Nordsoft91:
Tested and now it works well!

client/lobby/SelectionTab.cpp
```diff
@@ -41,8 +41,16 @@
 #include "../../lib/mapping/MapFormat.h"
 #include "../../lib/serializer/Connection.h"
 
-bool mapSorter::operator()(const std::shared_ptr<CMapInfo> aaa, const std::shared_ptr<CMapInfo> bbb)
+bool mapSorter::operator()(const std::shared_ptr<ElementInfo> aaa, const std::shared_ptr<ElementInfo> bbb)
 {
+	if(aaa->isFolder || bbb->isFolder)
+	{
+		if(aaa->isFolder != bbb->isFolder)
+			return (aaa->isFolder > bbb->isFolder);
```
Nordsoft91, line 49:
```suggestion
			return aaa->isFolder;
```

Nordsoft91:
I found a bug: empty folders are not visible, which is maybe ok for game loading, but not ok for game saving

client/lobby/SelectionTab.h
```diff
@@ -95,6 +107,8 @@ class SelectionTab : public CIntObject
 	ESelectionScreen tabType;
 	Rect inputNameRect;
 
+	std::tuple<std::string, std::string, bool, bool> checkSubfolder(std::string path);
```
Nordsoft91, line 110: please add comment, because signature is not clear: what returning tuple contains? Why path is not const reference?

dydzio0614: C++17 got std::string_view - should we start using that?

client/lobby/SelectionTab.cpp
IvanSavenko: Personally original version is better since it clearly states intent

client/lobby/SelectionTab.h
IvanSavenko: Personally I would rather return struct here, with clearly named members instead of a tuple.

string_view - any actual need? IMO unless this is actually needed, e.g. for performance std::string is a better choice. string_view is non-owning, which *might* be faster, but also needs way more attention - since owning string might be destroyed or reallocated, invalidating any string_view's created from it

client/lobby/SelectionTab.h
```diff
@@ -51,14 +61,16 @@ class SelectionTab : public CIntObject
...
+	std::vector<std::shared_ptr<ElementInfo>> curItems;
+	std::string curFolder;
 	size_t selectionPos;
-	std::function<void(std::shared_ptr<CMapInfo>)> callOnSelect;
+	std::function<void(std::shared_ptr<ElementInfo>)> callOnSelect;
 
 	ESortBy sortingBy;
 	ESortBy generalSortingBy;
 	bool sortModeAscending;
+	int curFilterSize = 0;
```
IvanSavenko, line 73: No need to abbreviate variable names
```suggestion
	int currentMapSizeFilter = 0;
```

client/lobby/SelectionTab.cpp
```diff
@@ -313,6 +321,15 @@ void SelectionTab::keyPressed(EShortcut key)
 
 void SelectionTab::clickDouble(const Point & cursorPosition)
 {
+	int position = getLine();
+	int py = position + slider->getValue();
```
IvanSavenko, line 325:
```suggestion
	int itemIndex = position + slider->getValue();
```

client/lobby/SelectionTab.cpp
```diff
@@ -328,29 +345,82 @@ void SelectionTab::showPopupWindow(const Point & cursorPosition)
...
+			auto [folderName, baseFolder, parentExists, fileInFolder] = checkSubfolder(elem->fileURI);
+
+			if(parentExists)
+			{
+				auto folder = std::make_shared<ElementInfo>();
+				folder->isFolder = true;
+				folder->folderName = "..";
+				if (boost::range::find_if(curItems, [](std::shared_ptr<ElementInfo> e) { return e->folderName == ".."; }) == curItems.end()) {
+					curItems.push_back(folder);
+				}
```
IvanSavenko, line 413: Perhaps split into two statements for clarity?
```suggestion
				auto itemIt = boost::range::find_if(curItems, [](std::shared_ptr<ElementInfo> e) { return e->folderName == ".."; })
				if (itemIt == curItems.end()) {
					curItems.push_back(folder);
				}				
```

client/lobby/SelectionTab.cpp
```diff
@@ -328,29 +345,82 @@ void SelectionTab::showPopupWindow(const Point & cursorPosition)
...
+	bool parentExists = (curFolder != "");
+	bool fileInFolder = false;
+
+	std::vector<std::string> filetree;
+	// delete first element (e.g. 'MAPS')
+	boost::split(filetree, path, boost::is_any_of("/"));
+	filetree.erase(filetree.begin());
+	std::string pathWithoutPrefix = boost::algorithm::join(filetree, "/");
+
+	std::string baseFolder = boost::filesystem::path(pathWithoutPrefix).parent_path().string();
```
IvanSavenko, line 370: these paths come from our virtual filesystem, not from actual system filesystem, better not to use boost::filesystem with it.
```suggestion
	filetree.pop_back();
	std::string baseFolder = boost::algorithm::join(filetree, "/");
```

client/lobby/SelectionTab.cpp
```diff
@@ -639,12 +751,13 @@ void SelectionTab::parseCampaigns(const std::unordered_set<ResourceID> & files)
...
 	{
-		auto info = std::make_shared<CMapInfo>();
+		auto info = std::make_shared<ElementInfo>();
 		//allItems[i].date = std::asctime(std::localtime(&files[i].date));
 		info->fileURI = file.getName();
 		info->campaignInit();
-		if(info->campaign)
+		if(info->campaign) {
 			allItems.push_back(info);
+		}
```
IvanSavenko, line 760: 1. We place opening brackets on new line
2. No need for brackets for simple, single-line statements
```suggestion
		if(info->campaign)
			allItems.push_back(info);
```

client/lobby/SelectionTab.cpp
```diff
@@ -584,11 +694,13 @@ void SelectionTab::parseMaps(const std::unordered_set<ResourceID> & files)
...
 		try
 		{
-			auto mapInfo = std::make_shared<CMapInfo>();
+			auto mapInfo = std::make_shared<ElementInfo>();
 			mapInfo->mapInit(file.getName());
 
 			if (isMapSupported(*mapInfo))
+			{
 				allItems.push_back(mapInfo);
+			}
```
IvanSavenko, line 703:
```suggestion
				allItems.push_back(mapInfo);
```

client/lobby/SelectionTab.h
dydzio0614: when threading is not involved and string_view is used as function parameter then there should be no owning string destruction? string_view is easy to use and remember, if we think about new contributors

IvanSavenko: and string is even easier to use since you don't have to correctly pick between string and string_view. Especially if we think about new contributors.

IvanSavenko: Micro-optimization in vast majority of cases. Unless you're operating on strings several Mb in size.

At the very least this code is guaranteed to work. Even if it will take 1mks longer than code with const reference.

IvanSavenko: For example, here we're discussing copying of 1 string, about 20-30 symbols long.

Meanwhile this code uses boost::split which requires construction of a vector and multiple strings inside said vector. As well as erasing front element from that vector, causing shift of all elements in it.

There is also at least one new redraw request which will force rendering of at least 800x600 pixels via CPU.

So effect of this oh so horrible string copy is zero. Sure, replacing it with const reference is not a bad idea, but nothing critically important for PR acceptance.

krs0:
Super nice QoL feature!!!

Tested it for loading and saving. Looks good.
I can save in what folder I want by typing folder name / file name. Which is good!
It is a little confusing in what folder the save will be made. (also in HD mod).
HD mod has more space for folder name: (if not hard to implement)

image

client/lobby/SelectionTab.cpp
Laserlicht: Prefer it also

dydzio0614:
If this is "lack c++ skills" then you can keep lacking xD

client/lobby/SelectionTab.h
Laserlicht: I've changed it to struct. :)

Laserlicht:
> If this is "lack c++ skills" then you can keep showing how much you lack xD

For me it's lack if the googling time is significant over the coding time. :D
And I'm still at war with the topic of pointers. ;)

Laserlicht:
> I found a bug: empty folders are not visible, which is maybe ok for game loading, but not ok for game saving

I know. But hard to implement. As krs0 mentioned you can always write to subfolder by using `/` in the filename. Now also creates folder, if not exists.

dydzio0614:
well... code review will help detecting code flaws? :P
I also dont remember as much about pointers as before, but as long as you do not use "new" anywhere else than smart pointer constructors, and check if unique_ptr is possible instead of shared_ptr for your use case, that is some decent first step

Laserlicht:
grafik

Looks like this now.

krs0:
Hmmm... but why the uppercase?
Folders stand out, because of their different column layout.
And files? They just would not look right if all uppercase.

Laserlicht:
> Hmmm... but why the uppercase? Folders stand out anyways, because of their different column layout. And files? They just would not look right if all uppercase.

Because it is the concept of handling files in VCMI. Has nothing to do with this PR. I think it's to avoid Case problems. HoMM3 files have sometimes different cases. And we have case sensitive and case insensitive platforms.

Interestingly only standard ascii chars are uppercase. If you use for example german umlauts (äöü) they will stay lowercase.

Laserlicht:
I think I found a solution to change this behavior in vcmi.

@IvanSavenko Any side effects of this case change?

krs0:
Great! works with proper case now. It truly outshines HD mod version now :)

It works really well with filters across different folders!
Sorting option is also remembered.

Maybe not in the scope of this PR but I can think only of 1 improvement:

- SoD icon is some pixels too big and looks out of place.

image

Laserlicht:
> Ups! Crash when I want to save a game... no mods. Every single time / Windows10 Autosave is created.

Fixed. Problem was when the last selected savegame was empty (e.g. after new install).

Also fixed the problem, that the last selected file was saved in uppercase.

IvanSavenko:
> I think I found a solution to change this behavior in vcmi.
> 
> @IvanSavenko Any side effects of this case change?

Perhaps store both case-sensitive / original file name and UPPER CASE resource name in that class? Meaning - add member to that class for original file name. And use case-sensitive name when we want to display it to player, e.g. in save screen?

VCMI code generally assumes case-insensitive resource ID's

IvanSavenko:
By the way, feel free to fix https://github.com/vcmi/vcmi/issues/2540
Obviously optionally, but you might already know where to look

Laserlicht:
> > I think I found a solution to change this behavior in vcmi.
> > @IvanSavenko Any side effects of this case change?
> 
> Perhaps store both case-sensitive / original file name and UPPER CASE resource name in that class? Meaning - add member to that class for original file name. And use case-sensitive name when we want to display it to player, e.g. in save screen?
> 
> VCMI code generally assumes case-insensitive resource ID's

changed it. Should work.

#2540 is fixed

also integrated #2497 here (same files...) which fix #2365

krs0 (commented):
Comment deleted

lib/mapping/CMapInfo.cpp
```diff
@@ -63,11 +63,11 @@ void CMapInfo::saveInit(const ResourceID & file)
 	countPlayers();
 	std::time_t time = boost::filesystem::last_write_time(*CResourceHandler::get()->getResourceName(file));
 
-    std::tm tm = *std::localtime(&time);
+	std::tm tm = *std::localtime(&time);
```
krs0, line 66: Could this be extracted in a GetDateFromTime function?

lib/mapping/CMapInfo.cpp
```diff
@@ -55,16 +58,27 @@ void CMapInfo::saveInit(const ResourceID & file)
...
+	fullFileURI = boost::filesystem::canonical(*CResourceHandler::get()->getResourceName(file)).string();
 	countPlayers();
 	std::time_t time = boost::filesystem::last_write_time(*CResourceHandler::get()->getResourceName(file));
-	date = std::asctime(std::localtime(&time));
+
+	std::tm tm = *std::localtime(&time);
+	std::stringstream s;
+	s.imbue(std::locale(""));
+	s << std::put_time(&tm, "%x %X");
+	date = s.str();
```
krs0, line 70: Could this be extracted in a GetDateFromTime function?

client/lobby/SelectionTab.cpp
```diff
@@ -234,6 +242,7 @@ void SelectionTab::toggleMode()
 		case ESelectionScreen::saveGame:
 			parseSaves(getFiles("Saves/", EResType::SAVEGAME));
 			inputName->enable();
+			inputName->activate();
```
IvanSavenko, line 245: Why was this change necessary? Enable should call activate() anyway from what I see

client/lobby/SelectionTab.cpp
```diff
@@ -313,6 +322,15 @@ void SelectionTab::keyPressed(EShortcut key)
...
 void SelectionTab::clickDouble(const Point & cursorPosition)
 {
+	int position = getLine();
+	int itemIndex = position + slider->getValue();
+
+	if(itemIndex >= curItems.size())
+		return;
+
+	if(curItems[itemIndex]->isFolder)
+		return;
```
IvanSavenko, line 332: Why return? Or we use single click for folders?

Also - getLine() can return -1 (judging from code below), so itemIndex might be -1 as well, potentially crashing on access to curItems[itemIndex]

client/lobby/SelectionTab.cpp
```diff
@@ -328,29 +346,97 @@ void SelectionTab::showPopupWindow(const Point & cursorPosition)
...
+
+auto SelectionTab::checkSubfolder(std::string path)
+{
+	struct Ret
+	{
+		std::string folderName;
+		std::string baseFolder;
+		bool parentExists;
+		bool fileInFolder;
+	} ret;
```
IvanSavenko, line 367: Huh, wasn't aware that such syntax is possible (use type defined locally as return type).

Well, a bit weird IMO but if nobody else is against such syntax then OK for me.

client/lobby/SelectionTab.cpp
Laserlicht: This change fixes #2540

client/lobby/SelectionTab.cpp
Laserlicht: Googled C++ ;)

client/lobby/SelectionTab.cpp
Laserlicht: We use single click.

Other thing: fixed

Nordsoft91:
@Laserlicht , will empty folders be supported (be shown in save/load screens)?

krs0:
I thought also that it was weird, but on second thought... not showing empty folders works great with filters! So I would not lose that functionality.

Maybe display as an option  **✓ Show empty folders** somewhere?

Laserlicht:
> @Laserlicht , will empty folders be supported (be shown in save/load screens)?

No. Currently not.

As @krs0 mentioned. It actually only affects the saving of files. Yes, it would be quite nice. But would also require a lot of redesign of the maplist just for that functionality. I don't currently plan to do that.

You can save to subfolders. No matter if they are empty or maybe don't exist at all. By simply specifying a subfolder-tree in the filename with "/".

client/lobby/SelectionTab.cpp
```diff
@@ -397,8 +490,9 @@ void SelectionTab::sort()
 		std::stable_sort(curItems.begin(), curItems.end(), mapSorter(generalSortingBy));
 	std::stable_sort(curItems.begin(), curItems.end(), mapSorter(sortingBy));
 
+	int firstMap = boost::range::find_if(curItems, [](std::shared_ptr<ElementInfo> e) { return !e->isFolder; }) - curItems.begin();
```
IvanSavenko, line 493:
```suggestion
	int firstMapIndex = boost::range::find_if(curItems, [](std::shared_ptr<ElementInfo> e) { return !e->isFolder; }) - curItems.begin();
```

lib/filesystem/Filesystem.cpp
```diff
@@ -147,7 +147,7 @@ ISimpleResourceLoader * CResourceHandler::createInitial()
 	for (auto & path : VCMIDirs::get().dataPaths())
 	{
 		if (boost::filesystem::is_directory(path)) // some of system-provided paths may not exist
-			initialLoader->addLoader(new CFilesystemLoader("", path, 0, true), false);
+			initialLoader->addLoader(new CFilesystemLoader("", path, 1, true), false);
```
IvanSavenko, line 150: No support nested folders? Only one level?

IvanSavenko (approved):

lib/filesystem/Filesystem.cpp
Laserlicht: Interestingly, `1` is fully sufficient here. It still works over several levels.

Laserlicht:
@IvanSavenko If everything all right: Can you merge? I've have another improvements (other thing) on top of this and it would be easier to make a PR on top of this. :)

PR #2524, opened 2023-08-08
[1.3.1] Reward can refuse fix

Nordsoft91:
Simple fix for 1.3.1


Nordsoft91:
@nullkiller , no, it's likewise rewardable objects on map, but for town buildings (specifically for Pavilion update)

PR #2525, opened 2023-08-08
Battle fast targets

nullkiller:
It targets beta for now but not intended to be included as hotfix.
The idea is to reduce amount of canBeCastAt(). Has not significant performance improvement.


nullkiller:
Some tests show that spellcast evaluation time can be reduced by up to 50% when a lot of spells are in spellbook.

DjWarmonger:
Looks good to me, calculating the speed outside the loop is definitely the way.

nullkiller:
Yes this patch is quite simple, but may theoretically affect real spellcasting. So I checked manually and have not found any issues so far.

DjWarmonger:
Okay, for now I can tell it works.

DjWarmonger:
So, what are we waiting for?

nullkiller:
Ok, this can be merged. It is not complex so hardly can break anything

Issue #2394, opened 2023-07-23
Implement tutorial


Laserlicht:
**Describe your proposal**
The output "Sorry, tutorial is not implemented yet" says everything.

Absolutely low priority. But still worth an issue.

PR #2530, opened 2023-08-08
Support tutorial

Nordsoft91:
https://github.com/vcmi/vcmi/issues/2394

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -38,6 +38,7 @@
 	"vcmi.radialWheel.splitUnit" : "Split creature to another slot",
 
 	"vcmi.mainMenu.tutorialNotImplemented" : "Sorry, tutorial is not implemented yet\n",
+	"vcmi.mainMenu.tutorialNotExist" : "Tutorial map is not found\n",
 	"vcmi.mainMenu.highscoresNotImplemented" : "Sorry, high scores menu is not implemented yet\n",
 	"vcmi.mainMenu.serverConnecting" : "Connecting...",
 	"vcmi.mainMenu.serverAddressEnter" : "Enter address:",
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -37,8 +37,6 @@
 	"vcmi.radialWheel.moveUnit" : "Move creatures to another army",
 	"vcmi.radialWheel.splitUnit" : "Split creature to another slot",
 
-	"vcmi.mainMenu.tutorialNotImplemented" : "Sorry, tutorial is not implemented yet\n",
-	"vcmi.mainMenu.tutorialNotExist" : "Tutorial map is not found\n",
 	"vcmi.mainMenu.highscoresNotImplemented" : "Sorry, high scores menu is not implemented yet\n",
 	"vcmi.mainMenu.serverConnecting" : "Connecting...",
 	"vcmi.mainMenu.serverAddressEnter" : "Enter address:",
--- a/client/mainmenu/CMainMenu.cpp
+++ b/client/mainmenu/CMainMenu.cpp
@@ -199,7 +199,7 @@ static std::function<void()> genCommand(CMenuScreen * menu, std::vector<std::str
 				case 2:
 					return std::bind(CMainMenu::openLobby, ESelectionScreen::loadGame, true, nullptr, ELoadMode::CAMPAIGN);
 				case 3:
-					return std::bind(CMainMenu::startTutorial);
+					return std::bind(CMainMenu::openLobby, ESelectionScreen::loadGame, true, nullptr, ELoadMode::TUTORIAL);
 				}
 			}
 			break;
--- a/client/mainmenu/CMainMenu.cpp
+++ b/client/mainmenu/CMainMenu.cpp
@@ -383,32 +383,8 @@ void CMainMenu::startTutorial()
 		
 	auto mapInfo = std::make_shared<CMapInfo>();
 	mapInfo->mapInit(tutorialMap.getName());
-	
+	CSH->startMapAfterConnection(mapInfo);
 	CMainMenu::openLobby(ESelectionScreen::newGame, true, nullptr, ELoadMode::NONE);
-	
-	std::thread waitForConnectionThread([mapInfo](){
-		boost::this_thread::sleep(boost::posix_time::milliseconds(100)); //delay this thread
-		
-		//connecting to server
-		while(CSH->state != EClientState::LOBBY)
-		{
-			if(CSH->state == EClientState::CONNECTION_CANCELLED || CSH->state == EClientState::NONE)
-				return;
-			boost::this_thread::sleep(boost::posix_time::milliseconds(50));
-		}
-		
-		//start game from main thread
-		GH.dispatchMainThread([mapInfo]()
-		{
-			while(!CSH->si || mapInfo->fileURI != CSH->si->mapname)
-			{
-				CSH->setMapInfo(mapInfo);
-				boost::this_thread::sleep(boost::posix_time::milliseconds(50));
-			}
-			CSH->sendStartGame();
-		});
-	});
-	waitForConnectionThread.detach();
 }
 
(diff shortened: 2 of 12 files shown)


Mods/vcmi/config/vcmi/english.json
```diff
@@ -38,6 +38,7 @@
 	"vcmi.radialWheel.splitUnit" : "Split creature to another slot",
 
 	"vcmi.mainMenu.tutorialNotImplemented" : "Sorry, tutorial is not implemented yet\n",
```
nullkiller, line 40: Looksl ike we no longer need "vcmi.mainMenu.tutorialNotImplemented"

Mods/vcmi/config/vcmi/english.json
```diff
@@ -38,6 +38,7 @@
...
+	"vcmi.mainMenu.tutorialNotExist" : "Tutorial map is not found\n",
```
IvanSavenko, line 41: Please use existing string from H3
`	"core.genrltxt.742" : "The tutorial map file \"tutorial.tut\" is missing from the maps directory.  You cannot start a tutorial without this file.",
`

client/mainmenu/CMainMenu.cpp
```diff
@@ -198,7 +199,7 @@ static std::function<void()> genCommand(CMenuScreen * menu, std::vector<std::str
 				case 2:
 					return std::bind(CMainMenu::openLobby, ESelectionScreen::loadGame, true, nullptr, ELoadMode::CAMPAIGN);
 				case 3:
-					return std::bind(CInfoWindow::showInfoDialog, CGI->generaltexth->translate("vcmi.mainMenu.tutorialNotImplemented"), std::vector<std::shared_ptr<CComponent>>(), PlayerColor(1));
+					return std::bind(CMainMenu::startTutorial);
```
IvanSavenko, line 202: This should load tutorial, not start it

lib/filesystem/ResourceID.cpp
```diff
@@ -129,6 +129,7 @@ EResType::Type EResTypeHelper::getTypeFromExtension(std::string extension)
 		{".MSG",   EResType::MASK},
 		{".H3C",   EResType::CAMPAIGN},
 		{".H3M",   EResType::MAP},
+		{".TUT",   EResType::MAP},
```
IvanSavenko, line 132: This will also make tutorial show up in map list.
Not needed and not visible in H3 either

client/mainmenu/CMainMenu.cpp
Nordsoft91: honestly, I'm not sure that I understand how this should work. Does it load some hardcoded savegame?

IvanSavenko: No. It shows only games made in tutorial.

lib/filesystem/ResourceID.cpp
Nordsoft91: I introduced workaround to filter tutorial map from scenario list, but yes it's not the best approach

client/mainmenu/CMainMenu.cpp
```diff
@@ -371,6 +372,34 @@ void CMainMenu::openCampaignScreen(std::string name)
...
+	}
+		
+	auto mapInfo = std::make_shared<CMapInfo>();
+	mapInfo->mapInit(tutorialMap.getName());
+	
+	CMainMenu::openLobby(ESelectionScreen::newGame, true, nullptr, ELoadMode::NONE);
+	
+	GH.dispatchMainThread([mapInfo](){
+		while(!CSH->c || !CSH->c->handler)
+			boost::this_thread::sleep(boost::posix_time::milliseconds(50));
```
IvanSavenko, line 391: This will lead to UI freeze. Any way to avoid?

client/mainmenu/CMainMenu.cpp
```diff
@@ -371,6 +372,45 @@ void CMainMenu::openCampaignScreen(std::string name)
...
+		}
+		
+		//start game from main thread
+		GH.dispatchMainThread([mapInfo]()
+		{
+			while(!CSH->si || mapInfo->fileURI != CSH->si->mapname)
+			{
+				CSH->setMapInfo(mapInfo);
+				boost::this_thread::sleep(boost::posix_time::milliseconds(50));
```
IvanSavenko, line 406: Still sleeps in main thread

IvanSavenko: Why do even need these threads / sleeps?
Normal scenario startup works just fine without all of this

Nordsoft91: scenario startup is performed in 3 steps:
1) connect to server (show connection pop-up), separate thread waits for connection
2) set scenario by sending request to the server - main thread waits for response
3) send game start request - main thread, same as for single scenario

IvanSavenko: Do we actually need this code which is:
- copy-pasted version of debug map loading
- introduces even more threading to our already overcomplicated threading model?

I already lost count how many threading-related bugs I have fixed recently.  Do we really need to add more code to it?

Nordsoft91: what is bad about multithreading code? I suppose if we want tutorial with no ui freeze, than yes multithreading code is needed. As for bugs, functionality is isolated, I don't see regression risk

IvanSavenko: And what's good in spawning dozens of threads and then attempt to synchronize all of them? What's good in copy-pasting code instead of reusing it?

Do we require some CPU heavy operations that we need to run in a separate thread?

Why not set callback that game should call when connection has been established? Why not unify debug code from which you copy-pasted this block and your new code?

> As for bugs, functionality is isolated, I don't see regression risk

Just like battle replay that broke leveling up on winning against attacking enemy?
Just like new boat types that broke existing boat logic due to uninintialized variables?
Just like configurable buildings in towns that broke visiting of towns by heroes, build all cheat and music playback after recruiting new hero?

IvanSavenko: > If you want to wait for local server, it should be almost instaneous (unless there are other problems in the system). Then don't delay it by adding 100 ms sleep at the beginning.

Far from it. 200-300 ms is already "fast". Might take up to 5 seconds on low-end systems, especially when mods are used.

> I also don't see the benefit for unlocking user's GUI while he already decided to launch a tutorial.

At the very least software mouse pointer will freeze and OS will interpret vcmi as hanging app while sleep is ongoing. GUI needs to be unlocked. Input on the other hand might be ignored/blocked if needed.

IvanSavenko: > Starting local server takes couple seconds, but 100ms is needed to ensure that server launching thread if started before waiting for connection thread

Sleep is not a thread synchronization mechanism. If you need to wait for another thread use mutex / conditional variable / atomic / callback / etc, depending on what's appropriate

> I see one new thread, which waits for connection to server in order to not block popup "connecting to server" window, why we are talking about dozens?

One thread here, one thread there and we already have 13 different threads just in game client (not counting vcmi_lib which also spawns own threads as well as threads from SDL which also interact with our code). Each of these threads complicate code and make tracking notoriously hard multithreading issues even more complex.

> This part is toxic and counter-productive. I don't know how am I supposed to react

Problem is, unless you plan to stick around for a longer period of time, any bugs in this code are for me to resolve. Sure, this code is isolated and won't be called unless you click "Tutorial" button but it is not like we can ask players not to click this button in a wrong way.

So please, keep multithreading to minimum to reduce already painful debugging efforts.

IvanSavenko: How should I contact you then?

I've tried pinging you on some iphone issue. No reply until I got reply from kambala instead. I've tried commenting on your PR with discovered regression. Also no reply. We also had issues since May that in the end turned out to be regression from combat replay PR.

Nordsoft91: nothing special, both GitHub and slack work well usually, but we are all humans so I can miss something or be late with response, certainly no intent to ignore. But even in that case follow up is welcomed

IvanSavenko (approved):

PR #2546, opened 2023-08-10
Fix crash at battle end

DjWarmonger:
Both reproductible and random crashes happened here.

### DIFF
--- a/lib/NetPacksLib.cpp
+++ b/lib/NetPacksLib.cpp
@@ -38,6 +38,8 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
+#define THROW_IF_NO_BATTLE if (!gs->curB) throw std::runtime_error("Trying to apply pack when no battle!");
+
 void CPack::visit(ICPackVisitor & visitor)
 {
 	visitBasic(visitor);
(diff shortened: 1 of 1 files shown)


IvanSavenko:
Were crashes reproducible on latest beta?

Can you add error to log in there?
Such situation should not happen in the first place, no need to silently return

IvanSavenko:
I did saw similar crashes but they should be already properly fixed in beta by this commit:
https://github.com/vcmi/vcmi/pull/2505/commits/d7cbe4ecde8ae8fb56350886fb00db70ff8c17b2

DjWarmonger:
I get this in current beta:
image
image

But no crash.

DjWarmonger:
[crash_2546.zip](https://github.com/vcmi/vcmi/files/12311548/crash_2546.zip)
And yes, looks like AI hero already moves on adventure map before making that request.

IvanSavenko:
Yeah, seems to be happening when AI ends battle with a spell.
However request is correctly denied by server so shouldn't cause a crash.

IvanSavenko:
To clarify - I think that this change is not needed because this is applying netpack to *gamestate*. This netpack has been generated by server, applied by server locally and sent to client.

Such packs must always succeed. If they fail for one reason or another then this is issue somewhere on server and should be fixed on server.

DjWarmonger:
If that's the case, assertion is needed. It should never fail, yet it did.

IvanSavenko:
Yeah, assertion or even throw runtime_error with appropriate message.

DjWarmonger:
Done.

IvanSavenko (approved):

PR #2554, opened 2023-08-10
Fixes for crashes from Android crash reporter

IvanSavenko:
- Fixes crash on fake cursor move
- Fixes crash on accessing non-existing building (outdated mods?)
- Fix possible nullptr dereference in AI
- Fix crash on right-clicking player flags area in RMG setup window
- Replaced some usages of boost::format with vcmi MetaString to avoid exceptions on invalid format string in some localizations


IvanSavenko:
Retargeting this to master, for one more Android hotfix. Hopefully final one before 1.3.1

All previous hotfixes were successful in fixing issues they were intended to fix, but since players no longer crash on game start / map loading they start playing more and start triggering crashes that previously were never reached.

PR #2555, opened 2023-08-11
Fix hota offset + 2 possible crashes

DjWarmonger:
- Fix offset of HoTA monster templates. For good this time.
- Fix the risk of place objects outside the map if they don't start at their (0,0) tile
- Added more potential locations for quest artifacts in zone


DjWarmonger:
Okay, generated 50 XL+U maps in a row without a crash. Can't get better than that, I guess.

PR #2566, opened 2023-08-12
Fix crash on video playback (and who knows how many more)

### LINKED ISSUE #2553: Crash on main menu video playback
**Describe the bug**
According to our crash reporter on Android, playing video file in main menu in some cases leads to a crash. Not reproducible on my phone, even after testing all mods with such video.

Unfortunately, stack trace is the only information that is available.
Other than than:
- likely related to mods. Our default main menu config does not have video configured in it since it is based on H3:Complete
- may not be system-specific, but no confirmation whether this happens on platforms other than Android
- not sure whether this is first entry to main menu (and first video to play) or some "interference" from (for example) in-game video and then - returning to main menu
- happens on multiple different phones, so this is unlikely to be corrupted install

**To Reproduce**
Unknown

**Did it work earlier?**
Unknown

**Version**
 - OS: Android
 - Version: 1.3.0

**backtrace**
```
(avcodec_send_packet+208)
(CVideoPlayer::nextFrame()+269)
(CVideoPlayer::update(int, int, SDL_Surface*, bool, bool)+364)
(CMenuScreen::show(Canvas&)+112)
(WindowHandler::simpleRedrawImpl()+136)
(WindowHandler::simpleRedraw()+122)
(CMainMenu::update()+340)
(CGuiHandler::renderFrame()+123)
(SDL_main+444)
```

IvanSavenko:
Yet another crash caused by poor multithreading logic.
In lobby, server handler thread could access GUI state without acquiring PlayerInterface::pim first, leading to potential data races, commonly resulting in crash inside ffmpeg.

- Fixes #2553 

Thanks to Laserlicht for catching this crash in debugger.

### DIFF
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -88,6 +88,8 @@ template<typename T> class CApplyOnLobby : public CBaseForLobbyApply
 public:
 	bool applyOnLobbyHandler(CServerHandler * handler, void * pack) const override
 	{
+		boost::unique_lock<boost::recursive_mutex> un(*CPlayerInterface::pim);
+
 		T * ptr = static_cast<T *>(pack);
 		ApplyOnLobbyHandlerNetPackVisitor visitor(*handler);
 


Laserlicht:
Seems to work. Since added this line no crash anymore.

PR #2570, opened 2023-08-12
Town selection window

Laserlicht:


krs0:
Hmmmm... how do you open this window? Is this an extra window separate from the smaller ones? Dies it replace the HotA windows?

I have to say, that I hope the small windows are still available. There was something nice at focusing on choosing 1 item at a time. First town, then hero and last bonus

krs0:
Maybe not for this PR but keep in mind 2-3 possible modifications in the future.

Am option to select random town - but different from opponents!
Random town but same for all players!
Random town but exclude banned ones.
Random hero from a hero class.

Laserlicht:
See here how it's work:
https://github.com/vcmi/vcmi/assets/13953785/62ebe3b7-4ab2-489b-8175-261a4b5d0c61

krs0:
Could work in both small and big... not a really huge difference between the two.

I find the flow in individual windows case, a little more user friendly.
You select the town... you are presented with info only related to towns.
You select the hero for that town, you are presented with info only related to chosen town and its heroes. You do not care that there are other 15 towns, you only care about your selected town.

But yea I could see it working both ways.

krs0:
In WoG there is one option to rebuild towns. I do not know your implementation details, but maybe it could be relevant for your code.

image

Laserlicht:
I've changed the behaviour a little bit.

@krs0 Better like this?

https://github.com/vcmi/vcmi/assets/13953785/6af0bb8d-ac96-4e6e-8785-70707e09c44c

SoundSSGood:
Why not like in HD mod?

Laserlicht:
> Why not like in HD mod?

Unusable on smartphones... much too small...

SoundSSGood:
but it can be two separate windows with larger icons like yours. No?

Laserlicht:
I'm not sure it's currently possible to open multiple (active) windows in vcmi.

SoundSSGood:
In HD mod, they open in turn. First the city, then the hero. And the starting bonus (in our case)

IvanSavenko:
- maybe use small icons after all, but with some free space between them? Check right side of hero screen for reference. Should allow placing all H3 towns in 1 column without scroll
- if we're going for single screen to select hero & town perhaps try to integrate that hero description screen from HD Mod? One that shows starting army/skills/spell/etc?

In this case potential layout is:
column 1: town selection + scroll
(small interval with separator)
column 2: hero selection (might classes) + scroll
column 3: hero selection (magic classes) + scroll
(small interval with separator)
(optionally) town description on top
hero description

Not sure where to place bonus selection, might be obvious after layout for rest of items is done

SoundSSGood:
> In HD mod, they open in turn.

I made a mistake. they open at the same time.

Laserlicht:
> In HD mod, they open in turn. First the city, then the hero. And the starting bonus (in our case)
Only for the first time. The second time all windows are opened the same time (and maybe I want to change the town selection).

> maybe use small icons after all
I've tested it on mobile and it works great with the big icons -> I think it's not a bad idea to use them.

krs0:
@Laserlicht I like the new behavior better. First you select town, and only then heroes. If it creates no problems I would go with that (maybe others could chime in with their opinion)
(I would also put this demo video on Discord to get feedback). There is a lot of player activity there since the new release.

@IvanSavenko I love! the big icons. They look gorgeous on my 4k screen :D

IvanSavenko:
> I've tested it on mobile and it works great with the big icons -> I think it's not a bad idea to use them.

It does not looks bad, but:
1) If we add input snapping (where nearest control gets activated) small icons will work fine on mobile
2) Current hero screen works fine on mobile - I never had misclick and only few times - hit empty area between icons
3) It might allow more compact / more informative selection screen, something like this:

(although personally I would drop biography & speciality description from here to free space for town preview)
зображення
- towns are in a single column (needs scroll for mods)
- heroes are split by might & magic classes (and scroll if needed)

IvanSavenko:
Hmm... Another option is to use large icons with same layout. 
However would need scroll even for H3-only dataset

Laserlicht:
@IvanSavenko 

Also not a bad idea. Maybe add a okay button (and allow double click). Then show in the right area further information about last clicked (and activated) town/hero. I would use the right area not only to introduce the heroes, also the towns.

Yes we could also use the big icons. (scrolling via swiping is okay).

But personally for me it's to much to implement this. Feel free, if you want to do this. :)

krs0:
@Laserlicht if we go for single window... then maybe the extra window with only town selection is not helping too much? Just show the whole thing from the beginning.

@IvanSavenko The design proposed by Laserlicht looks much cleaner. Hero detailed info you do not really need always visible at hero selection. Same way as you do not need full town info. The right time to display that hero info window, would be on right click.

Hero detailed info will have a lot more info in VCMI. More spells more skills more creatures... it will be crowded enough all by itself.

krs0:
Buy yes hero separation by class or "special" would be nice.

IvanSavenko:
> @IvanSavenko The design proposed by Laserlicht looks much cleaner. 

Kind of dislike "combined" window to be honest.

Towns-only looks great.
Hero-only would probably also look great.
But Hero+Town+Bonus looks too large for amount of content in this window.

dydzio0614:
I like current design

IvanSavenko:
> @IvanSavenko
> But personally for me it's to much to implement this. Feel free, if you want to do this. :)

Noted. Unfortunately not sure if I'll manage to find time for this. MP features, bugfixing, code maintenance - all of it is on me right now :(

krs0:
I say its better from the 2 proposed :)
I said it before, separate windows would make for a much cleaner implementation. (do not know why HD mod put them side by side... selection speed maybe?)
I'm curious of Discord opinions as well :)

Laserlicht:
Currently I also test the simple solution: One simple window per Group.

grafik

IvanSavenko:
> Currently I also test the simple solution: One simple window per Group.

Perhaps we should go with it as simplest solution that does its job and look for user feedback later?

Laserlicht:
> > Currently I also test the simple solution: One simple window per Group.
> 
> Perhaps we should go with it as simplest solution that does its job and look for user feedback later?

Yes maybe.

Do you know how I can query the possible bonus for the selected map?

Laserlicht:
Simplified if much. We should not start with the perfect solution... We can improve it later...

https://github.com/vcmi/vcmi/assets/13953785/e2abcb86-509e-40ab-a193-d5e478c8fa6b

krs0:
In code you use castle in variable names. Elsewhere in code it is named faction. Could you change so its named consistent?

Laserlicht:
Would leave it like this for now. I think changes can always be made. And everyone likes something different. But so it is now relatively simple. But in the end it is quite time consuming to create something like this. Would like to see first how the users get along with it before we still (optically) adjust what. Fits in any case well in the style of HoMM3 pure (in contrast to the first concept).

https://github.com/vcmi/vcmi/assets/13953785/326b0ba8-29ff-4fe4-9713-f10cbb56b9ff

Laserlicht:
> In code you use castle in variable names. Elsewhere in code it is named faction. Could you change so its named consistent?

Yes. It's also inconsistent in VCMI itself.
PlayerSettings (no new stuff) has a parameter castle for example.

Mods/vcmi/Data/DlgBluBk.png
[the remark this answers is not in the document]
Laserlicht: It's from HD mod.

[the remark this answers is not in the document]
Laserlicht: Thanks. Changed...

client/lobby/OptionsTab.cpp
```diff
@@ -414,8 +417,355 @@ void OptionsTab::CPlayerOptionTooltipBox::genBonusWindow()
...
+		if(elem > 0)
+		{
+			elem--;
+			if(elem >= factions.size())
+				return;
+			set.castle = factions[elem];
+		}
+		else
+		{
+			set.castle = -1;
```
Nordsoft91, line 678: Does -1 mean random? Is there const for it?

krs0: set.castle = set.faction?

krs0: -1 to me means undefined :).
A well named constant would help.

client/CServerHandler.cpp
```diff
@@ -454,11 +454,11 @@ void CServerHandler::setPlayer(PlayerColor color) const
 	sendLobbyPack(lsp);
 }
 
-void CServerHandler::setPlayerOption(ui8 what, si8 dir, PlayerColor player) const
+void CServerHandler::setPlayerOption(ui8 what, si16 value, PlayerColor player) const
 {
 	LobbyChangePlayerOption lcpo;
 	lcpo.what = what;
-	lcpo.direction = dir;
+	lcpo.value = value;
```
Nordsoft91, line 461: why not keeping both? Those small arrows are still visible, so could be both ways to control (arrows + clicking at icon)

Laserlicht: This will keep both. But use the same variable for both. `direction` is misleading then.

client/lobby/OptionsTab.cpp
Laserlicht: > set.castle = set.faction?

Will not change here. Has nothing to do with this PR.

https://github.com/vcmi/vcmi/blob/96e820ddac698c1cc4bb6b31e1908f19c607a945/lib/StartInfo.h#L37C8-L37C8

client/lobby/OptionsTab.h
```diff
@@ -94,6 +96,49 @@ class OptionsTab : public CIntObject
 		CPlayerOptionTooltipBox(CPlayerSettingsHelper & helper);
 	};
 
+	class SelectionWindow : public CWindowObject
```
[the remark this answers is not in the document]
Laserlicht: Would not be bad. But it's more effort than copy out the source. :(

I use CPlayerSettingsHelper and SelType which are part of OptionsTab class...

Should be done in separate refactoring.

Nordsoft91 (approved):

IvanSavenko (approved):
- Still think that random town/hero should be in own row, like in initial version
- Some window title, e.g. "Select hero" might be a good idea (also was in initial version)
- New large classes should go to separate file when possible

Nothing critical though

krs0:
Also in favor of separate row with  small icons for random

Nordsoft91:
for me current version looks perfect: it adopted well for touch-screens. Random in separate row will create unnecessary empty space in window. From me the only 2 minor things
 - click sound is missing after selection
 - clicking outside window probably should close the window without change

Laserlicht:
added @Nordsoft91 sugesstions

The other points: Would wait for player feedback first.

Otherwise, the PR is ready to merge so far, I think.

PR #2571, opened 2023-08-12
Check for playing battle animations

wb180:
Don't start new staged animation, if there are animations playing. Otherwise it leads to situations, like in this [video](https://www.youtube.com/watch?v=pk9xFvjSsGc)(second cast)

### DIFF
--- a/client/windows/CTradeWindow.cpp
+++ b/client/windows/CTradeWindow.cpp
@@ -1262,11 +1262,14 @@ void CAltarWindow::SacrificeAll()
 	}
 	else
 	{
-		for(const auto & aw : arts->visibleArtSet.artifactsWorn)
+		std::vector<ConstTransitivePtr<CArtifactInstance>> artsForMove;
+		for(const auto& slotInfo : arts->visibleArtSet.artifactsWorn)
 		{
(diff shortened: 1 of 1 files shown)


IvanSavenko:
Sorry for delay.
Any idea how game ended up in such situation? Actually sounds like waitForAnimations() call is missing someplace else.

wb180:
> Sorry for delay. Any idea how game ended up in such situation? Actually sounds like waitForAnimations() call is missing someplace else.

It happens like that:
1. BattleInterface::spellCast - it stages BEFORE_HIT and HIT animations
2. Another thread on timer update calls BattleStacksController::updateBattleAnimations - BEFORE_HIT animations start here
3. BattleStacksController::stacksAreAttacked - here HIT animations are starting to play, because there is no check

IvanSavenko:
Sounds like step 2. is incorrect. updateBattleAnimations should not initiate animations when none are playing. It should only continue already started animation chain.

wb180:
Unfortunately have no idea how to fix this properly, so if someone wants to fix this - feel free to do it.

dydzio0614:
fixing client battle bugs in a nutshell xD

Issue #1931, opened 2023-04-10
AI cheating


dydzio0614:
Important for making AI as "strong & fun" as H3 without spending tons of effort to make extremely skilled AI that can outweight equivalent cheats in original H3.

Could be made in form of mod with different cheats being submods.
When toggled on this will give bonuses to players controlled by AI when playing VCMI.
"Mod settings" future feature could be used to easily override default boost values, but some pre-tested values could be good too for now.

Some useful examples:
1. More hitpoints (+ optionally other stats) for AI players when fighting neutrals - simulates being favored in quick combat, would counter AI losing a lot of troops  (my initial idea is +50% hitpoints so stronger version of vial of life artifact)
2. Experience points every day - all AI heroes get experience boost
3. More movement points - making base movement points like the ones with advanced logistics would be annoying to some, but  fun to others
4. Extra daily resources for some consistent resources over time which is not guaranteed by resourceful AI mod
5. Increased troop growth while towns are under AI control

Extra examples which probably could require too much clutter in codebase:
1. Ability to build 2 buildings per turn
2. Different values per difficulty level
3. More starting troops/buildings by default
4. Temporary stat boosts when fighting player heroes (like +3 attack,def,power,knowledge)

Extra note: this feature has chance to heavily impact VCMI singleplayer playability

IvanSavenko:
- More movement points - making base movement points like the ones with advanced logistics would be annoying to some, but fun to others
- Ability to build 2 buildings per turn
- More starting troops/buildings by default
- Temporary stat boosts when fighting player heroes (like +3 attack,def,power,knowledge)

Not sure about these ones - since they can be easily observable by human player.

dydzio0614:
can be observed by human but they pose extra challenge for those who seek whatever ways to have stronger AI without workarounds like controlai console command + typing cheats

Nordsoft91:
As I see, one of the biggest weaknesses of AI is battles. AI loses creatures in battles while human fights loseless with same neutrals.
So I suggest to prevent AI from losing creatures during battles with neutrals. Here are couple ideas how to achieve that:
 - "godmode" AI just won any neutrals which AI considers as weak enough to fight with
 - Increase attack and defence +5/+10 during battles with neutrals
 - Analyze after-battle casualties: if they <10% consider battle as lossless
 - Give 3-5 free resurrections to AI hero's per battle with neutrals

dydzio0614:
@Nordsoft91 see very first thing from my list :P

[...]

Issue #1364, opened 2023-01-03
Introduce chess-like time limits


IvanSavenko:
Competitive MP community uses more complex scheme of timers than available in H3 Complete:

**Base timer**
- counts down only when turn timer runs out, 
- never resets
- by default, set to 12:00
- if timer expires, player ends his turn

**Turn timer**
- counts down when player is making his turn on adventure map, 
- resets every turn
- any unspent time should be added to base timer
- by default, set to 6:00
- if timer expires, "base timer" starts counting down

**Battle timer**
- counts down during battles when creature timer runs out
- resets every battle
- any unspent time is lost
- by default, set to 2:00
- timer is paused during creatures' animation playback
- (pvp) if timer expires, currently acting creature will defend
- (pve) if timer expires, "turn timer" starts counting down

**Creature timer**
- counts down when player is choosing action in battle
- only active in pvp battles
- resets after each creature turn
- by default, set to 0:30
- if timer expires, "battle timer" starts counting down

Partially resolved by https://github.com/vcmi/vcmi/pull/2576, https://github.com/vcmi/vcmi/pull/2674, https://github.com/vcmi/vcmi/pull/2698

- [x] UI to specify new timers
- [x] (pve) if timer expires, "turn timer" starts counting down
- [x] creature timer only active in pvp battles
- [x] Timer pausing when turn is passed in HotSeat mode (potential solution - accept turn message or query)
- [x] Define logic for tactics
- [x] FPS counter overlaps with timer
- [ ] Query auto-answer
- [x] Fix potential concurrency issues - CGameHandler::run can run concurrently with incoming pack processing 
- [x] Fix timer drifting in CGameHandler::run - sleep for 100 ms is not guaranteed to be 100 ms and usually will be slightly longer
- [x] Fix hero getting stuck on water / in obstacles if he was moving with Flight when timer runs out
- [x] Fix hero getting stuck on water if he was moving with Water Walk when timer runs out (should be resolved by https://github.com/vcmi/vcmi/issues/2678)

krs0:
Is this slider for timers needed? On PC it has no function.

![image](https://github.com/vcmi/vcmi/assets/15194321/5c679e50-372f-402f-a04d-fbfcb28c7f7e)

krs0:
Base time is not conveying any meaning.
Extra time? Or Reserve time?

Also why by default unspent turn time is added to this?
Time limitations are there to shorten the game. Why would you make it longer by adding unspent time?
[...]

PR #2576, opened 2023-08-13
Implement turn timer feature

Nordsoft91:
Design suggestions are welcomed. Would appreciate for testing as many bugs could be potentially found.
What is covered:
 - timer if frozen during battles
 - timer is not frozen on queries, but turn pass won't happen until all queries resolved

Known problem: timer is not frozen on turn pass in hotseat mode. This needs to be done separately because change and impact of supporting it is quite big.


IvanSavenko:
This is what we actually need right now.
H3-like timers are not used neither in SP nor in MP
https://github.com/vcmi/vcmi/issues/1364

nullkiller:
Does AI adequately react on it?

client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -61,6 +63,9 @@ AdventureMapInterface::AdventureMapInterface():
 	shortcuts->setState(EAdventureState::MAKING_TURN);
 	widget->getMapView()->onViewMapActivated();
 
+	if(LOCPLINT->cb->getStartInfo()->turnTime > 0)
+		watches = std::make_shared<TurnTimerWidget>();
+
```
IvanSavenko, line 68: Should be defined in widget / configurable UI

server/CGameHandler.cpp
```diff
@@ -2075,10 +2083,31 @@ void CGameHandler::run(bool resume)
 			if(playerColor != PlayerColor::CANNOT_DETERMINE)
 			{
 				//wait till turn is done
+				const int waitTime = 100; //ms
+				int turnTimePropagateFrequency = 5000; //do not send updates too frequently
 				boost::unique_lock<boost::mutex> lock(states.mx);
 				while(states.players.at(playerColor).makingTurn && lobby->state == EServerState::GAMEPLAY)
 				{
-					static time_duration p = milliseconds(100);
+					if(gs->getStartInfo()->turnTime > 0 && !gs->curB) //turn timer check
```
IvanSavenko, line 2091: Please don't add any new functionality to CGameHandler, that class is already unmanageable and I want to split it to pieces soon.

Move this to a new class, let's call it TurnTimerProcessor for example.

client/adventureMap/TurnTimerWidget.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "../render/Colors.h"
+#include "../render/EFont.h"
+#include "../gui/CGuiHandler.h"
+#include "../gui/TextAlignment.h"
+#include "../widgets/Images.h"
+#include "../widgets/TextControls.h"
+#include "../../CCallback.h"
+#include "../../lib/CPlayerState.h"
+
+#include <SDL_render.h>
```
IvanSavenko, line 24: Please don't use SDL directly in UI code

IvanSavenko (commented):
- Needs chess timers, not simple countdown
- Needs AI handling 
- Needs human handling (potentially including closing all active dialogs)

dydzio0614:
@Nordsoft91 since this is considered one of most important features for 1.4 release, ensure that you have time to polish it until 1.4 releases, otherwise I would probably prefer Ivan implementing it, including simultaneous turns support

Nordsoft91:
> @Nordsoft91 since this is considered one of most important features for 1.4 release, ensure that you have time to polish it until 1.4 releases, otherwise I would probably prefer Ivan implementing it, including simultaneous turns support for timers. There are bunch of less convoluted features that can still be done, like AI cheating (#1931) or spells usage on/off toggle for autobattle (got a lot of upvotes in discord when mentioned) - at least they are less prone to major changes when simultaneous turns come

quite weird preference, because collaboration and labor division usually speeds up the development, also I don't see interference with simultaneous turns. As for time allocation, if there will be regression surely will do the best to fix it.
Unless there are better/cheaper/faster incoming solutions I'll keep PR

Nordsoft91:
> Does AI adequately react on it?

I tested few times turn expiration for AI and it seems to be fine

> Needs human handling (potentially including closing all active dialogs)

works from the box, except scenario I described in description

IvanSavenko:
> works from the box, except scenario I described in description

- No notification of any kind. When you are time-pressed you won't be looking on countdown. 
- If your hero is moving at the end of your turn you start getting server errors.
- If your hero is flying over obstacles he will stop movement while on top of them.

IvanSavenko:
And another reminder that we need #1364 and not this version

krs0:
On audio notifications. Besides the 10 seconds countdown, would a single beep at 30 sec be helpful?

Nordsoft91:
@krs0 , I think possible to make different sound configuration, do you have sound we may use?

dydzio0614:
maybe https://www.youtube.com/watch?v=aJwZntohTt8 :rofl:

krs0:
@dydzio0614 Haha :)

I am not good at this kind of stuff, but maybe better than hat dydzio provided :)

Default system message (sysmsg I think) maybe would scare me that I am in last 10 seconds.

So maybe Something different like:
https://notificationsounds.com/message-tones/that-was-quick-606
Or
https://notificationsounds.com/message-tones/good-news-387

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -190,18 +190,29 @@ ColorRGBA InterfaceObjectConfigurable::readColor(const JsonNode & config) const
...
-			return Colors::GREEN;
-		if(config.String() == "orange")
-			return Colors::ORANGE;
-		if(config.String() == "bright-yellow")
-			return Colors::BRIGHT_YELLOW;
+		if(config.isString())
+		{
+			if(config.String() == "yellow")
+				return Colors::YELLOW;
+			if(config.String() == "white")
```
krs0, line 197: One of the things I hate about C++: a simple thing like using string values in enum, has no support in language :(

Could this be done separately? Where the enum is defined?

A map of some sorts 
Or a method to enum class...whith the switches inside
FromString(const std:string &enumValue)

IvanSavenko: This is not enum, but a set of static constants. I guess in theory we also add fromString method in there, but so far this is the only place where we use string->color conversion so probably no need.

krs0:
BUG: Restart scenario does not restart timer.

krs0:
At start of turn would'nt it be better to be actual set timer value? 00 jumping to actual timer value is a little confusing.

image

krs0:
In hot seat mode timer runs while turn is not started yet. Should not count down on this screen.
image

krs0:
I can stay forever in a mine :)

image

IvanSavenko:
Tested AI - game client hangs when AI fails to end turn in time. Reproduced twice out of two attempts

IvanSavenko:
Also, perhaps use player color as color for timer text? So red text when red is making turn, blue when blue is making turn, etc.
This is how hota/hd does it. Right now it is a bit weird to see timer during AI turn.

Nordsoft91:
> Tested AI - game client hangs when AI fails to end turn in time. Reproduced twice out of two attempts

Failed to fix, so disabled time expiration for AI players

> Also, perhaps use player color as color for timer text? 

Done

Nordsoft91:
@krs0 , visual bugs should be fixed, but auto-response for queries (hero exchange, garrisons, etc.) is way more complex.
I implemented it in a way, that time continue expiring while you on query dialog but turn cannot be end. Changing this requires significant refactoring of multiple components so I decide to postpone it.

Nordsoft91:
Merge now to avoid further complex conflict resolution. Need more testing, UI and query resolution

server/CGameHandler.cpp
```diff
@@ -1048,10 +1060,15 @@ void CGameHandler::run(bool resume)
...
+				const int waitTime = 100; //ms
 				boost::unique_lock<boost::mutex> lock(states.mx);
 				while(states.players.at(playerColor).makingTurn && lobby->state == EServerState::GAMEPLAY)
 				{
-					static time_duration p = milliseconds(100);
+					turnTimerHandler.onPlayerMakingTurn(gs->players[playerColor], waitTime);
+					if(gs->curB)
+						turnTimerHandler.onBattleLoop(waitTime);
+
+					static time_duration p = milliseconds(waitTime);
```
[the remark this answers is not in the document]
IvanSavenko: Oh. And there is no concurrency protection!
Yes, there is `states.mx` mutex here, but this mutex only protects `states` member.

Game currently processes packs in `CVCMIServer::threadHandleClient`.
But this code runs in main thread

server/CGameHandler.cpp
```diff
@@ -1248,7 +1268,17 @@ bool CGameHandler::moveHero(ObjectInstanceID hid, int3 dst, ui8 teleporting, boo
...
+		{
+			for(auto topQuery = queries->topQuery(h->tempOwner); true; topQuery = queries->topQuery(h->tempOwner))
+			{
+				moveQuery = std::dynamic_pointer_cast<CHeroMovementQuery>(topQuery);
+				if(moveQuery)
+					queries->popIfTop(moveQuery);
+				else
+					break;
+			}
+		}
```
IvanSavenko, line 1281: Likely cause for regression  #2658

PR #2579, opened 2023-08-13
(develop) Refactoring of server-side battles code

IvanSavenko:
For simturns we'll need to support parallel battles for different players. However, right now game can only have single ongoing battle. Another problem is our GameHandler. This is massive god object, with very unclear interdependencies and unmanageable code with 500+ lines methods.

In order to simplify implementation of parallel battles I would like to reorganize server-side battle handling first - move all battle-related code from GameHandler to new, smaller classes (one at the moment, will split into few more soon). Split existing massive methods into set of smaller methods.

Actual support of parallel battles and any significant refactoring of battle code will be done in separate PR

Changes:
- moved battle-related code to several new classes in server/battle directory
- split some of massive functions in server battle code into composite methods, for ease of readability
- split queries file into composite parts
- casting a spell is now handled using same routines as all other battle actions
- after spell is cast, AI will re-receive activeStack() call to perform action for a stack
- Fixes error messages that sometimes appear during AI battles
- Removed battle handling thread from server. Now all battle processing is done immediately on receiving pack, in the same thread. Removes need for thread synchronization on server. Hopefully fixes crashes on Android.


IvanSavenko:
Mostly done, will probably do another testing & review round later.

PR #2584, opened 2023-08-14
Removed WoG artifacts from core. Replaced with unused to avoid crashes.

krs0:
cWog mod also updated in its own repo: branch 1.4. Commit:  Fix for WoG artifacts that were still using core  (6066fed9bd8943c8125bad64b5168187aecbef5a)

### DIFF
--- a/config/artifacts.json
+++ b/config/artifacts.json
@@ -2286,171 +2286,23 @@
 			"eversmokingRingOfSulfur"
 		]
 	},
-	"magicWand":
+	// Misiokles: not having these 3 artifacts could lead to crashes. RMG will try to place them in some randoms maps (and crash the game)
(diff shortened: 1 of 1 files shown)


Nordsoft91:
why replace? I mean, can we remove them at all and have only in wog mod?

krs0:
@misiokles warned heavily that removing those 3 entries will lead to crashes in RMG.

I have not understood the reasoning, but to be safe I followed Ivan's advice and just renamed and marked them as special.

IvanSavenko:
We need to load them because they are defined in H3 data files even if unused.
Check our creatures list - it has even more "unused" entries.
This is how our H3 data parser works at the moment

IvanSavenko (approved):

PR #2601, opened 2023-08-15
(1.3.2) Android SDK version bump

IvanSavenko:
Google Play requires targeting SDK version 33 (Android 13) for any new updates after August, 31.

According to migration guide, we don't use anything that might be affected, so transition should be smooth. However not sure if our build pipeline needs any other changes.


kambala-decapitator (approved):
That should be enough

PR #2605, opened 2023-08-17
Map preview

Laserlicht:
Another QoL. Not even HD has the function. :)

Todo:
- [x] Check performance (on my cpu with debugging version and worst case "The devil is in the detail" it's acceptable) -> especially mobile devices
- [x] Make proper use of canvas (@IvanSavenko How can I change the code correctly to canvas without using createFromSurface?)
- [x] draw town pos (with colors)
- [x] (Maybe) draw (smaller) underground layer???

grafik

### DIFF
--- a/client/lobby/SelectionTab.cpp
+++ b/client/lobby/SelectionTab.cpp
@@ -815,6 +815,9 @@ SelectionTab::CMapInfoTooltipBox::CMapInfoTooltipBox(std::string text, ResourceI
 	if(renderImage)
 		mapLayerImages = createMinimaps(ResourceID(resource.getName(), EResType::MAP), IMAGE_SIZE);
 
+	if(mapLayerImages.size() == 0)
+		renderImage = false;
+
 	pos = Rect(0, 0, 2 * BORDER + IMAGE_SIZE, 2000);
 	if(renderImage && mapLayerImages.size() > 1)
 		pos.w += IMAGE_SIZE + BORDER;
@@ -889,7 +892,15 @@ std::vector<std::shared_ptr<IImage>> SelectionTab::CMapInfoTooltipBox::createMin
 	std::vector<std::shared_ptr<IImage>> ret = std::vector<std::shared_ptr<IImage>>();
 
 	CMapService mapService;
-	std::unique_ptr<CMap> map = mapService.loadMap(resource);
+	std::unique_ptr<CMap> map;
+	try
+	{
+		map = mapService.loadMap(resource);
+	}
+	catch (...)
+	{
+		return ret;
+	}
 
 	for(int i = 0; i < (map->twoLevel ? 2 : 1); i++)
 	{
--- a/client/lobby/SelectionTab.cpp
+++ b/client/lobby/SelectionTab.cpp
@@ -18,6 +18,7 @@
 #include "../CServerHandler.h"
 #include "../gui/CGuiHandler.h"
 #include "../gui/Shortcut.h"
+#include "../gui/WindowHandler.h"
 #include "../widgets/CComponent.h"
 #include "../widgets/Buttons.h"
 #include "../widgets/MiscWidgets.h"
@@ -27,6 +28,7 @@
 #include "../windows/GUIClasses.h"
 #include "../windows/InfoWindows.h"
 #include "../render/CAnimation.h"
+#include "../render/Canvas.h"
 #include "../render/IImage.h"
 
 #include "../../CCallback.h"
@@ -37,9 +39,12 @@
 #include "../../lib/GameSettings.h"
 #include "../../lib/filesystem/Filesystem.h"
 #include "../../lib/campaign/CampaignState.h"
+#include "../../lib/mapping/CMap.h"
+#include "../../lib/mapping/CMapService.h"
 #include "../../lib/mapping/CMapInfo.h"
 #include "../../lib/mapping/CMapHeader.h"
 #include "../../lib/mapping/MapFormat.h"
+#include "../../lib/TerrainHandler.h"
 #include "../../lib/serializer/Connection.h"
 
 bool mapSorter::operator()(const std::shared_ptr<ElementInfo> aaa, const std::shared_ptr<ElementInfo> bbb)
@@ -353,7 +358,7 @@ void SelectionTab::showPopupWindow(const Point & cursorPosition)
 		if(curItems[py]->date != "")
 			text += boost::str(boost::format("\r\n\r\n%1%:\r\n%2%") % CGI->generaltexth->translate("vcmi.lobby.creationDate") % curItems[py]->date);
 
-		CRClickPopup::createAndPush(text);
+		GH.windows().createAndPushWindow<CMapInfoTooltipBox>(text, ResourceID(curItems[py]->fileURI), tabType == ESelectionScreen::newGame);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


IvanSavenko:
Canvas - probably can be kept as it for now. Will think how to handle such cases better once I have some time.

dydzio0614:
one problem with preview without ability to turn it off is that knowing map layout beforehand gives you an advantage

psyhlo:
> one problem with preview without ability to turn it off is that knowing map layout beforehand gives you an advantage

then make it active only for single play

krs0:
Nice feature!

Some maps have underground.

Some more information would be nice:
Size : Extra Large + Underground
Suggested Difficulty? : Expert
Can be Human Players : 2
Total Players : 8
Teams : 4

Laserlicht:
> Nice feature!

😀

> Definitely behind an option. Some maps have underground.

I'm not sure if we need an option here?

Too many options tend to overwhelm the normal user, don't they? Since the preview is "hidden" behind the right mouse button anyway and you don't have to use it (only if you need the file name -> as @dydzio0614 already said).

Sure. We could also render the background layer (if it exists), but isn't that too much?

> Some more information would be nice, because now it is in the back window. H3 expansion? Type: scenario or campaign Size : Extra Large + Underground Suggested Difficulty? : Expert Victory condition Loss condition Can be Human Players : 2 Total Players : 8 Teams : 4 Author? Language? Map Description?

Sure, would be grateful. But the RMB window is actually intended more as an overhang window, for information that is not on the main screen. Not as a complete overview (with redudant information).

Besides, we do not have all the values. Author and language are not provided in the Heroes Map format.

> then make it active only for single play

Yes, that should be possible.

krs0:
> then make it active only for single play
Why not have same feature for multiplayer scenarios as well? Do not see a reason for removing this feature from MP, since the benefits will be the same.

If someone really wants to "cheat" and see map overlay, they can just look on internet.

IvanSavenko:
By the way, what's the size of the preview? Same as minimap? Might be better to use same size - othervice it might look different from in-game minimap.

> then make it active only for single play

- or make it only for maps (not saves)
- or apply FoW from current player for saves

> Some maps have underground.

Show underground then, side-by-side, at the same scale

dydzio0614:
> > then make it active only for single play
> 
> Why not have same feature for multiplayer scenarios as well? Do not see a reason for removing this feature from MP, since the benefits will be the same: find easier a map that suits you.
> 
> If someone really wants to "cheat" and see map overlay, they can just look on internet.

Because host may be map author and some scenarios may assume that he should be the only one who knows map (for example he serves as spectator for 2 other player with premades map generated in hota RMG)

client/lobby/SelectionTab.h
```diff
@@ -11,6 +11,7 @@
 
 #include "CSelectionBase.h"
 #include "../../lib/mapping/CMapInfo.h"
+#include "../widgets/Images.h"
```
IvanSavenko, line 14: Prefer forward declarations to includes in headers
```suggestion
class CPicture;
```

IvanSavenko:
> Because host may be map author and some usage scenarios may assume that he should be the only one who knows map (for example he serves as spectator for 2 other players with premade map generated in hota RMG).
That assumes all works when this feature is used in multiplayer and non-host uses right mouse button

As it is now - this feature will not work in MP when only host has such map.
(although to be frank, the more I dig into multiplayer code the more I think that our MP does not works to begin with)

dydzio0614:
I guess its OK if it works in multiplayer and people can give feedback, though I request that this feature should not crash for multiplayer guests - probably won't since for now only host can see map list IIRC but we may want to change that

Laserlicht:
The feature works atm only for new games. I don't know how I can convert a savegame in a CMap.

For new games it renders playercolors. If we also get this functionality for load game, we could disable the playercolor rendering. The functions are already capable of this.

Mapsize is 169 (same as in "Quests"). The original minimap top right have 144.

Underground is supported now.

krs0:
@Laserlicht any pictures? :)

Laserlicht:
@krs0 

grafik

dydzio0614 (approved):

client/lobby/SelectionTab.h
```diff
@@ -62,6 +65,24 @@ class SelectionTab : public CIntObject
...
+		bool drawPlayerElements;
+		bool renderImage;
+
+		std::shared_ptr<CFilledTexture> backgroundTexture;
+		std::shared_ptr<CTextBox> label;
+		std::shared_ptr<CPicture> image1;
+		std::shared_ptr<CPicture> image2;
+
+		Canvas createMinimap(std::unique_ptr<CMap> & map, int layer);
+		std::vector<std::shared_ptr<IImage>> redrawMinimap(ResourceID resource, int size=144);
```
IvanSavenko, line 82: Also, `redrawMinimap` does not really redraws. Perhaps rename?
```suggestion
		Canvas createMinimapForLayer(std::unique_ptr<CMap> & map, int layer);
		std::vector<std::shared_ptr<IImage>> createMinimaps(ResourceID resource, int size=144);
```

IvanSavenko (approved):

Nordsoft91 (changes requested):
2 issues found:
 - client crash when unsupported map is previewed (e.g. create map in editor with mods, then disable mods, see example attached). You can show something like "preview is unavailable". It seems critical
[crash_test.vmap.zip](https://github.com/vcmi/vcmi/files/12386859/crash_test.vmap.zip)

 - text alignment is broken for maps with underground - this is minor
Снимок экрана 2023-08-20 в 03 23 23

client/lobby/SelectionTab.cpp
```diff
@@ -797,6 +803,107 @@ std::unordered_set<ResourceID> SelectionTab::getFiles(std::string dirURI, int re
...
+		label->resize(label->label->textSize);
+
+	pos.h = BORDER + label->label->textSize.y + BORDER;
+	if(renderImage)
+		pos.h += IMAGE_SIZE + BORDER;
+	backgroundTexture = std::make_shared<CFilledTexture>("DIBOXBCK", pos);
+	updateShadow();
+
+	// TODO: hacky redraw
+	label = std::make_shared<CTextBox>(text, Rect(BORDER, BORDER, pos.w-2*BORDER, 350), 0, FONT_MEDIUM, ETextAlignment::CENTER, Colors::WHITE);
```
Nordsoft91, line 833: spaces around binary operators here and in other places (see https://wiki.vcmi.eu/Coding_guidelines)

client/lobby/SelectionTab.cpp
```diff
@@ -797,6 +803,107 @@ std::unordered_set<ResourceID> SelectionTab::getFiles(std::string dirURI, int re
...
+	if(renderImage)
+		images = redrawMinimap(ResourceID(resource.getName(), EResType::MAP), IMAGE_SIZE);
+
+	pos = Rect(0, 0, 2*BORDER + IMAGE_SIZE, 2000);
+	if(renderImage && images.size() > 1)
+		pos.w += IMAGE_SIZE + BORDER;
+
+	label = std::make_shared<CTextBox>(text, Rect(BORDER, BORDER, pos.w-2*BORDER, 350), 0, FONT_MEDIUM, ETextAlignment::CENTER, Colors::WHITE);
+	if(!label->slider)
+		label->resize(label->label->textSize);
```
Nordsoft91, line 824: you can cover with lambda to avoid duplicating

client/lobby/SelectionTab.cpp
```diff
@@ -353,7 +359,7 @@ void SelectionTab::showPopupWindow(const Point & cursorPosition)
 		if(curItems[py]->date != "")
 			text += boost::str(boost::format("\r\n\r\n%1%:\r\n%2%") % CGI->generaltexth->translate("vcmi.lobby.creationDate") % curItems[py]->date);
 
-		CRClickPopup::createAndPush(text);
+		GH.windows().createAndPushWindow<CMapInfoTooltipBox>(text, ResourceID(curItems[py]->fileURI), tabType);
```
Nordsoft91, line 362: I suggest to add feature flag in game config in case to disable if somebody will have performance issues (for example, maps could be stored somewhere on network path)

IvanSavenko:
>     * client crash when unsupported map is previewed (e.g. create map in editor with mods, then disable mods, see example attached). You can show something like "preview is unavailable". It seems critical

But do we need to show such maps in map list to begin with? If we use H3 as an example then game does not show unsupported maps (e.g. hota/wog maps in sod)

client/lobby/SelectionTab.cpp
IvanSavenko: Not against it, but don't see much point either - we still need to load header for all maps to populate list and even XL maps are quite small, around 100kb

dydzio0614:
absence of valid maps that VCMI generally supports due to requiring sets of mods is confusing - if I download 50 maps pack and see 20 on list - I would rather want game to tell me which mods I lack rather than trying all mods or opening maps in editor

Laserlicht:
Fixed open points.

@Nordsoft91 
I was not able to reproduce text alignment error. Are you testing latest commit?

Nordsoft91:
Yes, I actually found that now it shows full file path instead of map name.

This is from 1.3.1
Снимок экрана 2023-08-20 в 18 12 08

And this is from your branch
Снимок экрана 2023-08-20 в 18 08 26

Map is the same on both screenshots

Laserlicht:
@Nordsoft91 should be fixed. To long text. Now I use the big window (for two layers) also for only one layer. More space for text. Also fixed space between textbox and image if scrollbar is visible (to long text).

Nordsoft91:
still same, but it's minor anyway. Up to you if want to merge
Снимок экрана 2023-08-20 в 23 35 06

Nordsoft91 (approved):

Nordsoft91:
I suggest to rollback to version with adopted size window for underground/no-underground maps and merge. Later I can take a look into this problem since it seems to be mac specific

Laserlicht:
@Nordsoft91 One more try. It also happens on Windows.

Only on my Ubuntu-Debug-Version everything is correct. :(

Laserlicht:
@Nordsoft91 Fixed (at least on Windows). Ready to merge. :)

Nordsoft91:
yea, confirm that alignment is fixed

PR #2617, opened 2023-08-19
Configurable autobattle spells usage

dydzio0614:
As in title - user can decide if next autocombat session should allow AI using spells or not


lib/CGameInterface.h
```diff
@@ -76,6 +77,7 @@ class DLL_LINKAGE CBattleGameInterface : public IBattleEventsReceiver
 
 	virtual ~CBattleGameInterface() {};
 	virtual void initBattleInterface(std::shared_ptr<Environment> ENV, std::shared_ptr<CBattleCallback> CB){};
+	virtual void initBattleInterface(std::shared_ptr<Environment> ENV, std::shared_ptr<CBattleCallback> CB, AutocombatPreferences autocombatPreferences){};
```
nullkiller, line 80: Default implementation should call the other overload

nullkiller (changes requested):

lib/CGameInterface.h
dydzio0614: It would hide missing implementation problem - implemented callback by battleAI - unless somebody has better idea new callback is intended to be used when autobattle is involved, but if somebody doesn't do it then AI assumes using everything is allowed. Wrongly using callback with autobattle preferences for non-autobattle cases will not break anything if we do not change default AutocombatPreferences values

nullkiller: 4 options
- you make initBattleInterface with 2 parameters non virtual and call 3-parameter one with preferences set all to true. All implementations only implement the one with 3 parameters.
- you make both methods pure virtual and each implementation must implement both
- you make initBattleInterface with 3 parameters to call the one with 2 parameters by default
- default implementation for 3 parameters throws exception clearly showing that this method is not supported

nullkiller: Otherwise if I call initBattleInterface with 3 parameters on StupidAI - it will not initialize internal CB and ENV and so on

dydzio0614: see new commit - I implemented it in stupid AI (aka option 2)

client/CPlayerInterface.cpp
```diff
@@ -662,7 +662,11 @@ void CPlayerInterface::battleStart(const CCreatureSet *army1, const CCreatureSet
 	if ((replayAllowed && useQuickCombat) || forceQuickCombat)
 	{
 		autofightingAI = CDynLibHandler::getNewBattleAI(settings["server"]["friendlyAI"].String());
```
nullkiller, line 664: We select battle AI for autocombat depending on settings so stupid and empty AI should also work here so your changes should be backward compatible or you have to extend all these AIs

dydzio0614: emptyAI is adventure map AI unless I miss something?

nullkiller: No, exactly there is a request to make some dummy adventure AI

lib/CGameInterface.h
nullkiller: Ok

nullkiller (approved):

lib/battle/AutocombatPreferences.h
```diff
@@ -0,0 +1,21 @@
...
+#pragma once
+
+struct AutocombatPreferences
+{
+	bool enableSpellsUsage = true;
+	//TODO: below options exist in original H3, consider usefulness of mixed human-AI combat when enabling autocombat inside battle
+	bool enableUnitsUsage = true;
+	bool enableCatapultUsage = true;
+	bool enableBallistaUsage = true;
+	bool enableFirstAidTendUsage = true;
```
IvanSavenko, line 19: Options are not implemented. Correct? I'd rather avoid options that exists but not used - end result is usually confusing...
```suggestion
//	bool enableUnitsUsage = true;
//	bool enableCatapultUsage = true;
//	bool enableBallistaUsage = true;
//	bool enableFirstAidTendUsage = true;
```

config/schemas/settings.json
```diff
@@ -314,6 +314,10 @@
 				"stickyHeroInfoWindows" : {
 					"type" : "boolean",
 					"default" : true
+				},
+				"enableAutocombatSpells" : {
+					"type": "boolean",
+					"default": true
```
IvanSavenko, line 320: I'd keep it off by default - AI is using mana *too* freely right now

dydzio0614: IMO it is natural to keep it on, so you start with no confusing handicaps on AI side, H3 also uses spells by default

lib/battle/AutocombatPreferences.cpp
IvanSavenko, line 1: No need for empty file, please delete

IvanSavenko (approved):

cmake_modules/VCMI_lib.cmake
```diff
@@ -7,6 +7,7 @@ macro(add_main_lib TARGET_NAME LIBRARY_TYPE)
 		${MAIN_LIB_DIR}/StdInc.cpp
 
 		${MAIN_LIB_DIR}/battle/AccessibilityInfo.cpp
+		${MAIN_LIB_DIR}/battle/AutocombatPreferences.cpp
```
IvanSavenko, line 10: CI fails
```suggestion
```

PR #2620, opened 2023-08-19
Rework spell resistance

rilian-la-te:
TODO:
 - [x] Rework XXX_IMMUNITY bonus to two new bonuses
 - [x] Rework all unit test involved these bonuses + add deprecatedTypeStr
 - [x] Rework jsons for these change
 - [x] Add bonus for generic subschool handling (in perspective replacing MIND_IMMUNITY)
 - [x] Add actual functional + replace MIND_IMMUNITY to new bonus
 - [ ]  Add handling of non-100% resist, add damage type for spells.
 - [ ] Add unit tests

### DIFF
--- a/client/battle/BattleStacksController.cpp
+++ b/client/battle/BattleStacksController.cpp
@@ -270,7 +270,16 @@ std::shared_ptr<IImage> BattleStacksController::getStackAmountBox(const CStack *
 	int effectsPositivness = 0;
 
 	for(const auto & spellID : activeSpells)
-		effectsPositivness += CGI->spellh->objects.at(spellID)->positiveness;
+	{
+		auto positiveness = CGI->spells()->getByIndex(spellID)->getPositiveness();
+		if(!boost::logic::indeterminate(positiveness))
+		{
+			if(positiveness)
+				effectsPositivness++;
+			else
+				effectsPositivness--;
+		}
+	}
 
 	if (effectsPositivness > 0)
 		return amountPositive;
--- a/client/windows/CSpellWindow.cpp
+++ b/client/windows/CSpellWindow.cpp
@@ -74,9 +74,9 @@ class SpellbookSpellSorter
 public:
 	bool operator()(const CSpell * A, const CSpell * B)
 	{
-		if(A->level < B->level)
+		if(A->getLevel() < B->getLevel())
 			return true;
-		if(A->level > B->level)
+		if(A->getLevel() > B->getLevel())
 			return false;
 
 
@@ -562,7 +562,7 @@ void CSpellWindow::SpellArea::hover(bool on)
 	if(mySpell)
 	{
 		if(on)
-			owner->statusBar->write(boost::to_string(boost::format("%s (%s)") % mySpell->getNameTranslated() % CGI->generaltexth->allTexts[171+mySpell->level]));
+			owner->statusBar->write(boost::to_string(boost::format("%s (%s)") % mySpell->getNameTranslated() % CGI->generaltexth->allTexts[171+mySpell->getLevel()]));
 		else
 			owner->statusBar->clear();
 	}
@@ -609,12 +609,12 @@ void CSpellWindow::SpellArea::setSpell(const CSpell * spell)
 		if(schoolLevel > 0)
 		{
 			boost::format fmt("%s/%s");
-			fmt % CGI->generaltexth->allTexts[171 + mySpell->level];
+			fmt % CGI->generaltexth->allTexts[171 + mySpell->getLevel()];
 			fmt % CGI->generaltexth->levels[3+(schoolLevel-1)];//lines 4-6
 			level->setText(fmt.str());
 		}
 		else
-			level->setText(CGI->generaltexth->allTexts[171 + mySpell->level]);
+			level->setText(CGI->generaltexth->allTexts[171 + mySpell->getLevel()]);
 
 		cost->color = secondLineColor;
 		boost::format costfmt("%s: %d");
--- a/client/windows/CSpellWindow.cpp
+++ b/client/windows/CSpellWindow.cpp
@@ -122,9 +122,9 @@ CSpellWindow::CSpellWindow(const CGHeroInstance * _myHero, CPlayerInterface * _m
 
 		++sitesPerOurTab[4];
 
-		spell->forEachSchool([&sitesPerOurTab](const spells::SchoolInfo & school, bool & stop)
+		spell->forEachSchool([&sitesPerOurTab](const ESpellSchool & school, bool & stop)
 		{
-			++sitesPerOurTab[(ui8)school.id];
+			++sitesPerOurTab[(ui8)school];
 		});
 	}
 	if(sitesPerTabAdv[4] % 12 == 0)
--- a/config/artifacts.json
+++ b/config/artifacts.json
@@ -1561,15 +1561,9 @@
 	{
 		"bonuses" : [
 			{
-				"subtype" : "spell.lightningBolt",
-				"type" : "SPELL_IMMUNITY",
-				"val" : 0,
-				"valueType" : "BASE_NUMBER"
-			},
-			{
-				"subtype" : "spell.chainLightning",
-				"type" : "SPELL_IMMUNITY",
-				"val" : 0,
+				"subtype" : "damageType.lightning",
+				"type" : "DAMAGE_TYPE_REDUCTION",
+				"val" : 100,
 				"valueType" : "BASE_NUMBER"
 			}
 		],
--- a/config/artifacts.json
+++ b/config/artifacts.json
@@ -774,8 +774,9 @@
 				"valueType" : "BASE_NUMBER"
 			},
 			{
-				"type" : "MIND_IMMUNITY",
-				"val" : 0,
+				"type" : "DAMAGE_TYPE_REDUCTION",
+				"subtype" : "damageType.mind",
+				"val" : 100,
 				"valueType" : "BASE_NUMBER"
 			}
 		],
--- a/config/bonuses.json
+++ b/config/bonuses.json
@@ -349,14 +349,6 @@
 		}
 	},
 
-	"MIND_IMMUNITY":
-	{
-		"graphics":
-		{
-			"icon":  "zvs/Lib1.res/E_MIND"
-		}
-	},
-
 	"NONE":
 	{
 		"hidden": true
--- a/config/creatures/conflux.json
+++ b/config/creatures/conflux.json
@@ -18,28 +18,23 @@
 				"val" : 100,
 				"valueType" : "BASE_NUMBER"
 			},
-			"meteorShowerImmunity" :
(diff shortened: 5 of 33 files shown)


IvanSavenko:
Can you also check
`DamageCalculator::getDefenseMagicFactor`
`DamageCalculator::getDefenseMindFactor`
?

These are last instances of hardcoded creature abilities. And both are magic-related (mind immunity / magic immunity)

rilian-la-te:
@Macron1Robot no. NEGATIVE_EFFECT_IMMUNITY is a replacement of XXX_IMMUNITY subtype 1 (immunity for XXX negative spells).

rilian-la-te:
@IvanSavenko it will be covered by a ATTACK_DAMAGE_TYPE bonus. There is one target of PR, as well as fixing Water Elementals without hardcoding of spells.

dydzio0614:
you can add "update wiki" to TODO list :P

rilian-la-te:
AFAIK, there is no WIKI now, and documentation is inside a new place.

krs0:
Now there are 2 new terms introduced for damage types: balefire and rock :(

Rock could be meteor and will still make me think of meteorShower
As for baleFire... I have no other suggestion than maybe leave it as Armaghedon? Since is the only spell that it affects?

IvanSavenko:
Or name both of them "meteor shower" and "armageddon" and don't try to invent new terms.
Or just use "earth" and "fire". AFAIK only water elementals are vulnerable to fire, and that includes armageddon.

rilian-la-te:
Rock can be renamed to "meteor" if you wish. About balefire: you are right, this is a type especially for armageddon, because some creatures is vulnerable to armageddon, but not for rest fire spells.

IvanSavenko:
Think from it from player point of view.
If I am looking at Water Elementals stats I want to see "Vulnerable to Armageddon" in list of bonuses. Having "balefire" in that list would only make it confusing. Same for meteor's.

Lightning and Ice are OK, since those are basically part of spell name, but not shower or armageddon.

IvanSavenko:
> @IvanSavenko so, you prefer the old way, where spells just added one by one, not splitted by types?

Yeah, probably. This is how H3 does it, this is how players are used to and this behavior does makes sense.

Having to define vulnerability/immunity to Lighting Bolt and to Chain Lighting separately *is* weird, so not against idea, but it should still be clear to player.

Also feels that this list of elements is rather arbitrary. Why Death Ripple does not have "death" element or something? Why Implosion has no element? Why firewall has no fire element? Why this list is hardcoded and modders can't define new element? In the end I think spell ID is most suitable abstraction level here - modders can add new spells and use ID of new spell to define vulnerabilities the way they want to.

Not sure what to do with Magic Elemental ability though. Perhaps we can use HATE bonus to *decrease* damage against specific units. Or to units that satisfy certain criteria (e.g. have mind immunity bonus / magic immunity)

I am also planning some changes to modding system that would make compatibility patches easier to make & use, so if mod A makes spell and mod B has creature that should be vulnerable to such spell, then either mod can add compatibility patch that will work automatically.

rilian-la-te:
@IvanSavenko element is simply an uint8_t, so, modders clearly will can define a new types (sorry, but not identifiers, I do not know how to allow identifiers to be updated by mods).

Firewall has a fire element, just look inside a FireWallTrigger spell)

rilian-la-te:
About showing an element - agree in some extent, but then I think than we should show mind spells somehow too. Because for now all H3 players just learn mind spells by heart.
This problem was with mind spells too, and was in OH3.

About second - I think it is better just list an element than listing a 5+ spells (like now on water elementals).

rilian-la-te:
@IvanSavenko I think elements are more intuitive than listening all spells.

For example, lightning and ice is almost intuitive (but Titan's bolt is not lightning in OH3).
Fire is intuitive too (fireball, inferno, fire wall)
Rock - also intuitive (but only meteor shower exists).
Mind is also almost intuitive.

Problem is only weird behaviour of Armageddon. Okay, we can preserve Armageddon as special spell if desired, and set its element to ordinary fire).

dydzio0614:
IMO anything other than MIND_IMMUNITY bonus is counter intuitive to use for getting titan-like ability...

rilian-la-te:
@dydzio0614 there is also ice immunity and lightning immunity in OH3.

dydzio0614:
wasn't immunity is basically water immunity, as there are no water damaging spells other than ice (and magic arrow)

IvanSavenko:
> @IvanSavenko I think elements are more intuitive than listening all spells. 

Not sure if I agree with this to be honest.
"Meteor shower" is way more clear than "rock". Especially when corresponding school is called "Earth"
"Fire" might sound intuitive but AFAIK water elementals are NOT vulnerable to fire, so fire wall should NOT have fire element? Which is definitely weird.

We also have Implosion, Land Mines, Magic Arrow - damaging spells that don't have any element? Same to Death Ripple / Destroy Undead

dydzio0614:
@IvanSavenko water elementals definitely get double damage from fireball, and I think water elementals get double damage from firewall and inferno as well

IvanSavenko:
> @IvanSavenko water elementals definitely get double damage from fireball, and I think water elementals get double damage from firewall and inferno as well

But not from firewall or landmines, both of which are in Fire school. And while landmines element is unclear (fire? earth? explosion?) firewall clearly does fire damage. And yet not affected by vulnerability.

rilian-la-te:
1. Magic arrow, implosion, land mines, titan's bolt does not have an element in OH3. Why all spells should have an element? 
2. I will recheck water elementals and fire wall. 
3. Magog's fireball is fire in HoTA.

IvanSavenko:
> 1. Magic arrow, implosion, land mines, titan's bolt does not have an element in OH3. Why all spells should have an element?

So Titan's Bolt is not lighting? A creature that is vulnerable to "lighting" is not vulnerable to Titan's Bolt? 
Yeah, sure I agree that this is how H3 works, but replacing vulnerability to Lightning Bolt with vulnerability to "lightning" will not make things more clear.

> Magog's fireball is fire in HoTA.

But VCMI uses SoD as base, not HotA


I get idea behind your changes, but the more I look into how H3 works here the more I think that immunities / vulnerabilities should be spell-based, not damage type based.

rilian-la-te:
Anything than based on spellID/creatureID is bad for mods, because it will add interdependency to all mods which add spells to all mods which add creatures.

rilian-la-te:
So, we need spell grouping by something other than school.

IvanSavenko:
> Anything than based on spellID/creatureID is bad for mods, because it will add interdependency to all mods which add spells to all mods which add creatures.

I am also planning some changes to modding system that would make compatibility patches easier to make & use, so if mod A makes spell and mod B has creature that should be vulnerable to such spell, then either mod can add compatibility patch that will work automatically.

> So, we need spell grouping by something other than school.

If this is something internal to allow automatic generation of list of spells creature is vulnerable to - sure.
But H3 is very inconsistent here. Player should not guess whether Titan's Bolt is lightning spell or whether firewall is fire spell.

Don't see any clear way to do this without explicit list of spells.

rilian-la-te:
Explicit list is normal, but should not be defined inside a creature. It should be defined elsewhere and should be extendable by mods. Element is just a way to define such list without major rework. 

Compatibility patches is not a solution of a problem, because it just easies interdependency handling, and does not remove it altogether. 

But elements will remove such handling (or, if desired, other way to create creature-independent spell lists).

rilian-la-te:
What I mean by interdependency:

If said modder creates Ice Lance spell, he should download all creature mods, check if it has any creature resistant or immune to ice bolt, and write a compatibility patch. Good? 

If we accept elements, then said modder will simple assign a new spell to ice group and all will work automatically.

IvanSavenko:
I get it, but:
- do we actually have non-zero number of mods that add spells which would benefit from elements mechanics?
- do we actually have non-zero number of mods that add creatures which would benefit from elements mechanics?
- will this decrease clarity for H3 creatures / spells relations?

If this feature in the end will (for example) automatically generate list of spells creature vulnerable to (similar to H3 and to VCMI) then OK. Othervice from my point of view this is regression - there is nothing clear in having vulnerability to "fire" when Armageddon and Fire Wall, clearly fire spells are counted as "fire" element.

dydzio0614:
Personally I would keep single spell immunity and assume them to be hardcoded special cases that exist just because H3 did it, and encourge modders to use whole spell school immunities or hardcoding their own spell - should we add new damage type if mod gives one creature "immune to tornado spell" or something?

rilian-la-te:
@dydzio0614 there is a magic arrow, which is a problem for a "whole school immunity" idea, becausesgic arrow belongs to all schools. So, we need something like "water spells minus magic arrow" or "air spells minus destroy undead and magic arrow".

dydzio0614:
I edited my previous message, so check edited content

IvanSavenko:
> @IvanSavenko what you means by "auto-generate spell lists"? Can you give an example?

If water elementals are vulnerable to "fire" element (not to be confused with Fire School, and keep in mind that Fire Wall is not of fire element) then game should generate list of bonuses like "Vulnerable to Fireball" "Vulnerable to Inferno", "Vulnerable for XXX" instead of displaying unclear and misleading "Creature vulnerable to fire element".

Not insisting on any particular implementation, but end result IMO needs to be concrete list of spells in creature info window instead of abstract "elements"

rilian-la-te:
Why bonuses and not a separate list "spell modifiers"? We may check elements, and then generate spell modifiers for that element. I think it should not interfere with bonuses, because it defeat purpose of bonuses as atomic elements.

IvanSavenko:
As I said - don't care about specifics of implementation as long as result is clearly defined list of spells that can be viewed in creature info screen

krs0:
@rilian-la-te I think you are on a wrong path :(. You want to make elementals immune to element type.

In Oh3 its clearly not the case. For whatever reasons, (prob because Magic arrow is such a basic tool) the developers decided that they are immune to those specific spells and not also to magic arrow. And with no confusion possible they just listed those.

In the end in all OH3 immunities: there is no case where creatures are immune to all spells form a school, except for magic arrow (which I also believed). Every single case of mass immunity, includes magic arrow, because it is of all school types.
Now.... if you want to make available blanket school immunity for future mods... that's fine, but do not extend that to OH3 creatures, or rewrite OH3 creatures immunities as well.

So if someone wants to make Oceanids immune to all water damage spells... 

Either:
type: spellSchoolmmunity,
subtype: waterSchool (for all water spells, magic arrow is water spell)

or:

type: spellSchoolmmunity,
subtype: waterSchool 
parameter: 1 or "magicArrow" // to exlude magicArrow from that list

This is both simple and intuitive, as opposed to your proposition :(

Here an overview of all spells.
https://heroes.thelazy.net/index.php/Spell_immunity

rilian-la-te:
@krs0 your proposal is also not good if it comes to modding. For modding we need a dynamic spell lists and managing those lists, not spellIDs.

PR #2622, opened 2023-08-19
Allow empty army as result of quests creature removal

### LINKED ISSUE #2527: Unable to fulfill seer quests which require whole army
**Describe the bug**
One is unable to fulfill seer quests which require whole army to be given. For example there are at least two such seers in "With Sword and Fire" scenario. Example in video.

**Game logs**
N/a

**To Reproduce**
1. See the video

**Expected behavior**
One should be rewarded from the seer

**Actual behavior**
Seer acts like quest was not completed

**Did it work earlier?**
N/a

**Screenshots**
VCMI 1.3.0: 
https://github.com/vcmi/vcmi/assets/37045566/77dba308-e403-4ec8-98f1-198dfc22d64a

HoTA:
https://youtu.be/iPL7pHisn88?t=867

**Version**
 - OS: Android
 - Version: 1.3.0

**Additional context**

dydzio0614:
Fixes #2527 

Needs design of empty army handling - for example currently there is error on every move "hero has empty army". Would be a waste to remove strict ghost army prevention for typical matches.
I suggest new gamesettings entry.

### DIFF
--- a/client/Client.h
+++ b/client/Client.h
@@ -178,7 +178,7 @@ public:
 	void giveCreatures(const CArmedInstance * objid, const CGHeroInstance * h, const CCreatureSet & creatures, bool remove) override {};
 	void takeCreatures(ObjectInstanceID objid, const std::vector<CStackBasicDescriptor> & creatures) override {};
 	bool changeStackType(const StackLocation & sl, const CCreature * c) override {return false;};
-	bool changeStackCount(const StackLocation & sl, TQuantity count, bool absoluteValue = false) override {return false;};
+	bool changeStackCount(const StackLocation & sl, TQuantity count, bool absoluteValue = false, bool allowZeroStacksArmyResult = false) override {return false;};
 	bool insertNewStack(const StackLocation & sl, const CCreature * c, TQuantity count) override {return false;};
 	bool eraseStack(const StackLocation & sl, bool forceRemoval = false) override {return false;};
 	bool swapStacks(const StackLocation & sl1, const StackLocation & sl2) override {return false;}
--- a/lib/IGameCallback.h
+++ b/lib/IGameCallback.h
@@ -103,7 +103,7 @@ public:
 
 	virtual void giveCreatures(const CArmedInstance *objid, const CGHeroInstance * h, const CCreatureSet &creatures, bool remove) =0;
 	virtual void takeCreatures(ObjectInstanceID objid, const std::vector<CStackBasicDescriptor> &creatures) =0;
-	virtual bool changeStackCount(const StackLocation &sl, TQuantity count, bool absoluteValue = false) =0;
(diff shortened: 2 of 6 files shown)


krs0:
I do not get it. Why do we have cases where 0 armies are possible? Is it needed for backwards compatibility?

IvanSavenko:
Personally I would only allow *swapping* armies by default.
So if seer hut has bring troops as quest and offers troops as reward then allow taking all troops - since hero won't end up with empty army in such scenario.

Cases that leave hero without any army should be off by default - this is pretty much a bug in H3, no need to replicate it. Not against game option, but default value should be off.

SparrowJacek:
Using WoG mod, you can easily create a "ghost" hero with no army, if you win a fight having only your Commander left.

In general I find it very strange that a quest guard would not allow me to pass, if I really have the required army, no matter if I have anything else, or not.

dydzio0614:
@IvanSavenko hero without army is often used in crafted maps, #2527 is a light case where you get additional unit as a reward...

nullkiller:
Empty army might be very buggy. What if we just give 1lvl creature isntead?

IvanSavenko:
> @IvanSavenko hero without army is often used in crafted maps, #2527 is a light case where you get additional unit as a reward...

So if you play on such a map - enable this toggle in game config. Othervice this is bug abuse - you can go to any seer that requests army to create ghost hero

dydzio0614:
There is one thing remaining: we should probably not serialize this toggle so when somebody is stuck he can reload save, what do you think?

IvanSavenko:
> There is one thing remaining: we should probably not serialize this toggle so when somebody is stuck he can reload save, what do you think?

Good question...
Don't like making such exceptions so can't say that I like this approach.

We could try making *all* game settings non-serializeable and always pick them from active mods. Need to think how to implement this though - after all, we want client to receive settings from game server, so they needs to be serialized at some point.

nullkiller:
How about cheat code to change game settings on the fly?

IvanSavenko:
Not a bad idea actually.  Will need to make new netpack to propagate these settings, but other than that - should work.

mdindoffer:
> Cases that leave hero without any army should be off by default - this is pretty much a bug in H3, no need to replicate it. Not against game option, but default value should be off.

How bold of you to assume this. I would argue the exact opposite - the fact that the original game does not crash, rather handles these scenarios gracefully is a sign that it is *not* a bug.

Consider a case where you win a battle with only summoned elementals left standing. They then disappear after battle, so you have no army on the hero. Original game then leaves you in a perfectly valid state, with a hero that has maximum speed, who is defeated as soon as it enters another battle. Currently in VCMI your victorious hero just disappears.

Having an arbitrary requirement to always have >0 troops is simply wrong, no matter *how* the player got into this state - be it a quest seer or a battle.

nullkiller:
Homm3 GOG complete, attacked through event - double defeat.
Attacked a monster - double defeat.

PR #2624, opened 2023-08-20
Identifiers system refactoring and cleanup

IvanSavenko:
End goal is to fix/remake support for loading saves with different set of mods (or just with different versions of mods). Was somewhat low-priority for now, but since we want playable MP we should not requires exactly same versions and exactly same set of mods for every player.

Currently this feature exists only in name:
- it is broken due to our json serializer using identifier resolution system incorrectly.
- even if we fix it, only small parts of game (e.g. spell parameters) will correctly reload identifiers on mod change.
- even though server can reload mods using mod list from save, client does not reloads mods nor it rearranges filesystem to match mods from save, so game will fail on accessing to first image/animation from newly activated mod.

Current roadmap:
- (done in this PR) make common base class for identifiers so serialization of identifiers can be implemented in one place
- implement reloading (and re-resolving) of identifiers, e.g. if mod list/order changed or mod got new object / object was removed
- look for a way for client to reload list of active mods on the fly, without restart

Changes in this PR:
- All identifiers use same base class, `IdentifierBase`
- Converted most of "namespace enums" into proper enum classes
- Converted some identifiers with enum inside into identifier with constant values
- Replaced all implicit comparisons of identifiers with integers with explicit ones, via getNum method
- Split GameConstants.h file into several parts to improve readability. File itself remains as convenience include.

TODO for next PR's:
- Add conversion to/from string into identifier class
- Serialize identifiers as strings and re-query identifier ID on load
- Find and remove any bitmasks of identifiers, if any still remain
- Consider removing conversion operator to integer from identifier class


rilian-la-te:
@IvanSavenko did you removed Identifier<Enum> construction?

IvanSavenko:
> @IvanSavenko did you removed Identifier construction?

Yes, all identifiers are now non-template class. They still use template internally (as a base class), but final class is not a template.

rilian-la-te:
But then for each new identifier we need to write full-fledged class now( Not just enum and Identifier specialization.

IvanSavenko:
Yes, however it is quite small, e.g:
```
class CreatureID : public IdentifierWithEnum<CreatureID, CreatureIDBase>
{
public:
	using IdentifierWithEnum<CreatureID, CreatureIDBase>::IdentifierWithEnum;
};
```
Frankly, the only reason why I use class instead of "using" is to allow simple forward-declarations.

rilian-la-te:
So, you only moved boilerplate around. Can you document your changes in your documentation PR?

IvanSavenko:
Not now.
1) Goal for documentation PR is to integrate & update what we have now, not to document everything
2) This PR is likely not the last one related to identifiers, although rest should be more localized

PR #2627, opened 2023-08-20
Threading cleanup

IvanSavenko:
Downscaled version of (unsuccessful) boost -> std concurrency switch attempt.

List of changes:
- Use thread_local specified instead of boost::thread_specific_ptr
- Removed boost::thread_group since std has no such class
- Where possible, use boost API that matches c++ standard: posix_time -> chrono, sleep -> sleep_for
- Removed `terminate_cond` - seems to be unused and only causes freezes on low-end systems
- Give all threads created by VCMI name (should be visible in debug)


client/battle/BattleWindow.cpp
```diff
@@ -407,7 +407,7 @@ void BattleWindow::bFleef()
 		auto txt = boost::format(CGI->generaltexth->allTexts[340]) % heroName; //The Shackles of War are present.  %s can not retreat!
 
 		//printing message
-		owner.curInt->showInfoDialog(boost::to_string(txt), comps);
+		owner.curInt->showInfoDialog(boost::str(txt), comps);
```
kambala-decapitator, line 410: std actually has to_string unlike str

IvanSavenko: However std does not have anything like boost::format.

What happened here:
- I removed "including everything" include for boost thread and only included stuff we actually use
- because of this change, we no longer include boost::to_string (but still include boost::format)
- so I changes boost::to_string to boost::str which is standard way to convert boost::format to string

kambala-decapitator: Yeah, in 17 we have only snprintf, in 20 there's std::format

PR #2632, opened 2023-08-21
Spell resistance rework: part 1

rilian-la-te:
This is only immunity replacement part from draft spell lists PR, just for sake to make PRs smaller.


dydzio0614:
i dont see whats wrong with water elementals being by default immune to water spells from mods TBH

rilian-la-te:
It blocks magic arrow, which is not OH3 behaviour cleanly. So, for part 1 water immunity will be dropped from water elementals.

dydzio0614:
bless, clone, weakness... all work in base H3?

rilian-la-te:
Yes, only Ice Bolt and Frost Ring does not work. Aside normal elemental immunities (NON_LIVING).

rilian-la-te:
Updated to latest develop with battle changes.

rilian-la-te:
@IvanSavenko can you review this part? Psychic and Magic elementals will be in a next part)

lib/CBonusTypeHandler.cpp
```diff
@@ -99,60 +99,66 @@ std::string CBonusTypeHandler::bonusToGraphics(const std::shared_ptr<Bonus> & bo
 		fileName = sp->getIconImmune();
 		break;
 	}
-	case BonusType::FIRE_IMMUNITY:
+	case BonusType::SPELL_DAMAGE_REDUCTION: //Spell damage reduction for all schools
```
IvanSavenko, line 102: And this is why we need way to configure bonus icons at least based on subtype.
Hardcoding everything is not the best idea.

client/battle/BattleStacksController.cpp
```diff
@@ -271,7 +271,16 @@ std::shared_ptr<IImage> BattleStacksController::getStackAmountBox(const CStack *
...
-		effectsPositivness += CGI->spellh->objects.at(spellID)->positiveness;
+	{
+		auto positiveness = CGI->spells()->getByIndex(spellID)->getPositiveness();
+		if(!boost::logic::indeterminate(positiveness))
+		{
+			if(positiveness)
+				effectsPositivness++;
+			else
+				effectsPositivness--;
+		}
```
IvanSavenko, line 282: Might be a good idea to replace this with explicit enum instead of tribool

IvanSavenko (approved):

client/battle/BattleStacksController.cpp
rilian-la-te: There is already explicit enum, but used internally. Externally tribool was used. Maybe will do in next PR.

lib/CBonusTypeHandler.cpp
rilian-la-te: Agree. Selector-based config will be after part 2 (psychic and magic elementals).

rilian-la-te:
Updated to latest changes. Again)

Nordsoft91 (approved):

Issue #2286, opened 2023-07-03
UI is not locked while random map is being generated


IvanSavenko:
**Describe the bug**
While game generates random maps UI remains active and there is no indication that random map is being generated. This is especially noticeable for large maps on weaker systems (e.g. Android).

**To Reproduce**
1. Start a game
2. Attempt to generate random map (especially large ones)
3. Try to use UI while map is generated

**Expected behavior**
Possible alternatives:
a) Game displays map generation progress in a small window (H3 way)
b) Game switches to loading screen immediately once map generation starts

**Actual behavior**
There is no indication that map is being generated, UI is not locked and can be interacted with

**Did it work earlier?**
Unknown, probably not

**Version**
 - VCMI 1.2

DjWarmonger:
Well, we don't have loading progress bar in first place.
Even though RMG offers API for that, whcih already works in the editor.

dydzio0614:
IMO 1st alternative should be used to allow generation cancelling when we get more template features and map does not meet some validation - it is way easier to react to wrong generation furing actual generation (for example because mandatory zone object did not fit) than creating overcomplicated math algorithms to detect potential template flaws in theoretical corner cases of zone placement etc.

DjWarmonger:
Wat? :v

- Maps should be generated in seconds, no reason to interrupt anything
- If generator fails to place some object (or road), it will report the issue in console

dydzio0614:
for competitive reasons it should allow players to abort playing map that didn't generate all necessary objects etc. before they find that out ingame

Issue #2633, opened 2023-08-21
Start scenario doesn't work


Nordsoft91:
**Describe the bug**

**Game logs**
Please attach game logs: `VCMI_client.txt`, `VCMI_server.txt` etc.

**To Reproduce**
1. New game - > scenario
2. Select map incompatible with mods (example attached)
3. Press start and observe error message
4. Select any other map and press start again

[crash_test.vmap.zip](https://github.com/vcmi/vcmi/files/12401700/crash_test.vmap.zip)

**Expected behavior**
Another map can be starter

**Actual behavior**
Button start game is blocked

**Did it work earlier?**
yes

**Version**
 - Version: 1.3.1

**Additional context**
introduced with this commit https://github.com/vcmi/vcmi/commit/c524caee5c9055fda30c95a6b030402a10d4f290 because exception from verifyStateBeforeStart should be handled by upper level

PR #2635, opened 2023-08-21
Loading progress bar

Nordsoft91:
- Loading screen shown immediately now
 - Loading progress is shown for RMG
 - Progress bar is configurable
 - Game/map loading Load::Progress is not implemented (will be later)

https://github.com/vcmi/vcmi/assets/9308612/c94b32ed-ab52-4d30-9209-ad5b1e34f546


IvanSavenko:
Fixes #2286 I presume?

dydzio0614:
What if RMG fails in the middle though? For example try adding some throw in middle of generation where it will be caught and compare result before and after PR

Nordsoft91:
> What if RMG fails in the middle though? For example try adding some throw in middle of generation where it will be caught and compare result before and after PR
> 
> I think it even reproduces with current WoG mod in VCMI 1.3.1 and clash of dragons template... sigh

RMG fails means server crash. You will be dropped back to main menu

DjWarmonger:
> > What if RMG fails in the middle though? For example try adding some throw in middle of generation where it will be caught and compare result before and after PR
> > I think it even reproduces with current WoG mod in VCMI 1.3.1 and clash of dragons template... sigh
> 
> RMG fails means server crash. You will be dropped back to main menu

Yes, however only in game. In the map editor, if server crashes, progress bar will hang forever without any error message. But that's another issue.

client/mainmenu/CMainMenu.cpp
```diff
@@ -582,36 +582,70 @@ void CSimpleJoinScreen::connectThread(const std::string & addr, ui16 port)
...
+	{
+		int status = float(get()) / 255.f * progressBlocks.size();
+		
+		for(int i = 0; i < status; ++i)
+		{
+			progressBlocks.at(i)->activate();
+			progressBlocks.at(i)->visible = true;
+		}
+	}
+
```
IvanSavenko, line 627: Please no actions not related to rendering in show()/showAll(). Use tick() instead.

client/mainmenu/CMainMenu.h
```diff
@@ -9,8 +9,10 @@
  */
 #pragma once
 
+#include "../widgets/Images.h"
```
IvanSavenko, line 12: Prefer forward declarations to includes in headers

server/CVCMIServer.cpp
```diff
@@ -305,6 +323,10 @@ bool CVCMIServer::prepareToStartGame()
 		break;
 	}
 	
+	progressTrackingFinished = true;
+	while(progressTrackingFinished)
+		continue;
```
IvanSavenko, line 328: WTF?
Is this busy wait?
Is this busy waits that use non-atomic variable?

Please use condition_variable instead. Or explicitly join thread instead of detaching it if this is what is needed here.

server/CVCMIServer.cpp
```diff
@@ -279,19 +279,37 @@ void CVCMIServer::prepareToRestart()
 
 bool CVCMIServer::prepareToStartGame()
 {
+	Load::Progress * progressTracking = nullptr;
+	bool progressTrackingFinished = false;
+	std::thread progressTrackingThread([this, &progressTracking, &progressTrackingFinished](){
```
IvanSavenko, line 284: Right now we only use boost thread. Not sure if we should be mixing those two

server/CGameHandler.cpp
```diff
@@ -573,7 +573,7 @@ void CGameHandler::reinitScripting()
 #endif
 }
 
-void CGameHandler::init(StartInfo *si)
+void CGameHandler::init(StartInfo *si, Load::Progress *& progressTracking)
```
IvanSavenko, line 576: Any way to avoid this weird "reference to pointer"?

client/CServerHandler.cpp
```diff
@@ -551,7 +551,7 @@ void CServerHandler::sendStartGame(bool allowOnlyAI) const
 	catch (const std::exception & e)
 	{
 		showServerError( std::string("Unable to start map! Reason: ") + e.what());
-		return;
+		throw e;
```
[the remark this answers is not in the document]
Nordsoft91: I would prefer to remove try/catch because of this https://github.com/vcmi/vcmi/issues/2633 rather than fixing unknown scenarios, but not in this PR. Will remove change here

server/CVCMIServer.cpp
Nordsoft91: fixed

server/CGameHandler.cpp
Nordsoft91: changed

server/CVCMIServer.cpp
```diff
@@ -279,33 +279,54 @@ void CVCMIServer::prepareToRestart()
...
+	auto progressTrackingThread = boost::thread([this, &progressTracking]()
+	{
+		while(!progressTracking.finished())
+		{
+			std::unique_ptr<LobbyLoadProgress> loadProgress(new LobbyLoadProgress);
+			loadProgress->progress = progressTracking.get();
+			addToAnnounceQueue(std::move(loadProgress));
+			boost::this_thread::sleep(boost::posix_time::milliseconds(50));
+		}
+	});
```
IvanSavenko, line 295: Would be great to have an option to install callback that notifies whenever loading progress changes instead of yet another thread

config/mainmenu.json
```diff
@@ -2,7 +2,11 @@
 	//images used in game selection screen
 	"game-select" : ["gamselb0", "gamselb1"],
 
-	"loading" : ["loadbar"],
+	"loading" :
+	{
+		"background" : ["loadbar"],
+		"x": 395, "y": 548, "size": 18, "amount": 20, "name": "loadprog"
+	},
```
IvanSavenko, line 9: We really need to find a way to detect outdated mods without explicit "compatibility" field in every mod...

Nordsoft91: backward compatibility is preserved. If mod is not updated, there will be only loading screen without progress bar

Issue #2561, opened 2023-08-11
Improved autosaves


krs0:
**Describe your proposal**
After folder support is implemented autosaves can be improved.

General ideea is to have settings on when to save and to use <templates> for map names

Saving settings:
- autosave at start of each game (`GAME_START`)
- autosave at end of each turn
- autosave before each battle (in `SAVES/Autosave_#X`)
- max number of autosaves to keep
- templates for autosave name (including subfolders! relative to SAVES/)  
- EG: `Forge/%mapName%_EasyDificulty_%turn%`  `%mapName% %mapName% %turn% %timestamp% %playerNames% `should be supported

Use predefined templates for the following scenarios. (user can overwrite them at any time)

**Single Player scenario**: `%mapName%/%turn%.sav`
- Autosaves using turn name, saved in a folder with same name as map name.
 `EG: 
> A cold Day in Hell/
>     111.sav
>     112.sav
>     113.sav

**Single Player RMG**: `RMG/%mapName% - %timestamp%/%turn%.sav`
All autosaves go in `SAVES/RMG` subfolder.
All autosaves are in a separate folder: <template name> - <date>  
EG: `Jebus Cross - 2019.04.14 11;10`

**Multiplayer Scenario**: `%players%/%mapName% - %timestamp%/%turn%.sav`  
EG: `SAVES/RMG/Player1,Player2,Player3/Jebus Cross - 2019.04.14 11;10/111.sav`

**Multiplayer RMG**: `RMG/%players%/%mapName% - %timestamp%/%turn%.sav`  
EG: `SAVES/RMG/Player1,Player2,Player3/Jebus Cross - 2019.04.14 11;10/111.sav`

**HotSeat:**
Prepend [hotseat] before each save  
EG: `SAVES/RMG/Player1,Player2,Player3/Jebus Cross - 2019.04.14 11;10/[hotseat] 111.sav`

Laserlicht:
A few things that come to mind spontaneously:
- maximum number to keep is difficult to implement. How should the algorithm know which ones to delete in a user-defined schema (with possibly mixed savegames in the folder)?
- timestamp should be a fixed format (e.g. [ISO8601](https://en.wikipedia.org/wiki/ISO_8601) - basic format: `YYYYMMDDThhmmss`) to avoid problems (Windows cannot use some chars used in some time formats `:` or the `/` from the american date format)
- variable for autosave reason should also be added (e.g. `%reason%` will be `beforeBattle`)

[...]

PR #2637, opened 2023-08-22
folder support for autosave

Laserlicht:
With default setting it creates saves with path:
Autosave/ `Mapname` _ `STARTDATE` /Autosave_ `n` .vsgm1

STARTDATE is created each new game (loading gives the same STARTDATE). So you don't overwrite if you play one map multiple times.

Small solution. #2561 is much more complex and we can do this later...


Global.h
```diff
@@ -166,6 +167,9 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #include <boost/range/adaptor/reversed.hpp>
 #include <boost/range/algorithm.hpp>
 #include <boost/thread.hpp>
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_generators.hpp>
+#include <boost/uuid/uuid_io.hpp>
```
Nordsoft91, line 172: if now they are in Global.h, please remove from CServerHandler and CVCMIServer

client/CPlayerInterface.cpp
```diff
@@ -193,6 +193,8 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
 
 void CPlayerInterface::performAutosave()
 {
+	std::string id = cb->getStartInfo()->uuid.substr(0, 8);
```
Nordsoft91, line 196: if you need short uuid to 8 bytes, better to use CRC32, rather than substring to reduce duplication probability. Boost has implementation https://www.boost.org/doc/libs/1_83_0/doc/html/crc.html

lib/StartInfo.h
```diff
@@ -80,6 +80,7 @@ struct DLL_LINKAGE StartInfo
 	ui32 seedToBeUsed; //0 if not sure (client requests server to decide, will be send in reply pack)
 	ui32 seedPostInit; //so we know that game is correctly synced at the start; 0 if not known yet
 	ui32 mapfileChecksum; //0 if not relevant
+	std::string uuid;
```
Nordsoft91, line 83: we use uuid in several places, it's not clear which exact uuid is stored in StartInfo.
I recommend to rename to gameUuid or sessionUuid or something more infromative

Nordsoft91:
Tested, couple suggestions:
 - in folder name, instead of mapname_uuid8, use mask mapname_DDMMYY_hhmm
 - in autosave name, instead of autosave_n use mask autosave_mwd (autosave_111, autosave_112, ...), but it requires automatic cleaning up to avoid disk space consumption
OR
 - add month/week/day into description

Laserlicht:
@Nordsoft91 Thanks for the suggestions.

Yes make sense. But goes in the direction of here: https://github.com/vcmi/vcmi/issues/2561

I wanted to implement a small solution first. The big implementation may then still follow with pleasure.
Especially the cleanup can be complicated (that's why I introduced the UUID to be able to distinguish different games).

By the way, after your change in the Develop, the pipeline no longer runs through. :(

client/CPlayerInterface.cpp
Laserlicht: Yes. But I don't think that here is a real risk of collision?

Laserlicht:
fixed small mistake :)

Laserlicht:
Datetime is in the path now. Uuid is used to avoid overwriting when limit is disabled (playing the same map multiple times).

client/CPlayerInterface.cpp
```diff
@@ -230,7 +232,7 @@ void CPlayerInterface::performAutosave()
 					+ std::to_string(cb->getDate(Date::WEEK))
 					+ std::to_string(cb->getDate(Date::DAY_OF_WEEK));
 
-			cb->save("Saves/" + prefix + "Autosave_" + stringifiedDate);
+			cb->save("Saves/Autosave/" + prefix + "Autosave_" + id + "_" + stringifiedDate);
```
Nordsoft91, line 235: If prefix has folder name which is map+date, do we need id here?
I mean, here we have:
```
Saves/Autosave/mymap_datetime/Autosave_abcd_111
Saves/Autosave/mymap_datetime/Autosave_abcd_112
```
Can we have just `Saves/Autosave/mymap_datetime/Autosave_111` ?
And if so, do we need generated and serialised uuid at all?

Laserlicht: Think that it is just configurable. You can configure it so that it ends up in a folder. Then it is necessary.

Nordsoft91: According to lines 215-217 prefix cannot be empty. I barely imagine how users can use savePrefix without changing its logic. If I put `abc` into savePrefix, we won't have separate folders.
Then it turns into strange tool to disable folders for each game, which could be useful (especially on mobile devices as each save takes 2mb, so if I play 100 times, I will have 20*5 saves which is 200mb) but probably better to have dedicated parameter and remove savePrefix parameter.
But maybe I don't understand something about logic of this parameter

Laserlicht: Check the variable `usePrefix` and maybe check the gui. :)

If you uncheck the checkbox every save file is in one folder (`Autosave`). No problem, if limit > 0 (overwriting active). But if limit = 0 (keeping all) the filename conflicts without id.

Nordsoft91: In current implementation we have overrides with both enabled/disabled prefix with both limit 0 and >0.
I played several times on same map with different configuration and here what I have.

Снимок экрана 2023-08-23 в 04 32 09

Now in your implementation there will be following:
```
usePrefix=true, savePrefix="", autosaveCountLimit=0

Autosaves/mymap_23082023_103120/Autosave_abcd_111
Autosaves/mymap_23082023_103120/Autosave_abcd_112
```

Then I start new map
```
Autosaves/mymap_23082023_121222/Autosave_abcd_111
Autosaves/mymap_23082023_121222/Autosave_abcd_111
```

No overrides because I'd very hard to start map again in exacltly same moment.

The only scenario I see with possible overrides is
```
usePrefix=true, savePrefix="myprefix", autosaveCountLimit=0
```
But this is current behavior and not a regression. That's why intention to have extra uuid in filename is not clear

Laserlicht: Case 1 (overwrite possible):
`usePrefix=true, savePrefix="myprefix", autosaveCountLimit=0`
`usePrefix=false, savePrefix="", autosaveCountLimit=0`

Case 2 (no overwrite possible):
`usePrefix=true, savePrefix="", autosaveCountLimit=0`

Case 3 (overwrite intended):
`autosaveCountLimit>0`

With the current solution case 1 & 2 get the additional id to prevent overwriting. Also (as written) the uuid is a preparation for cleaning up.

Yes this is the current behavior. We can change it completely. But I don't want to invest time for that now.

Nordsoft91: > Case 1 (overwrite possible):

id is not needed

> Case 2 (no overwrite possible):

Covered by date-time, id is also not needed

> Case 3

id is not needed

Laserlicht:
Yes basically it's like loading the first round. ;)
Do you know a solution for this?

krs0:
Folder has date and time already... do we need also the UUID?
This will be used by nonprogrammers, I would say it will be a little off-putting.

Nordsoft91:
Another reason to avoid using uuid, here usePrefix=true, prefix=abc, autosavesCount=0
 - I have to scroll long long list to find my saves
 - saves folder has enormous amount of saves
 - how am I supposed to find a game which I want to play (last for example)? Right click on each of them?

Снимок экрана 2023-08-23 в 16 41 36

Nordsoft91:
> Yes basically it's like loading the first round. ;)
> Do you know a solution for this?

There are two startInfo instances in gameState: scenarioOps and initialOps. cb->getStartInfo(bool) has an argument to obtain first or second. When you restart, scenarioOps is taken from initialOps. Basically, you need to find this place on server side (cpack LobbyStartGame) and reset stringISO for scenarioOps after it's copied from initialOps.

Laserlicht:
> Another reason to avoid using uuid, here usePrefix=true, prefix=abc, autosavesCount=0
> 
>     * I have to scroll long long list to find my saves
> 
>     * saves folder has enormous amount of saves
> 
>     * how am I supposed to find a game which I want to play (last for example)? Right click on each of them?

For me it's exactly the reason for why it makes sense. With the use of the `prefix=abc` you already accepted that all saves are in one folder. With `autosavesCount=0` you say the system you want keep all files (only in this case (!!!) the uid is visible).

Without uid you will overwrite with the new game the first files from the beginning. Without the uid the only possible way to know if it's the old or new game is to look on the file change date. Which is much confusing I find.

For example:
```
Autosave_111
Autosave_112
Autosave_113
Autosave_114
Autosave_115
Autosave_116
```
Which one of these are the last one from the new game? It is clear: Autosave_114 😉

If you want, I can remove it. For me it's okay. 😊
Should I remove it completely? I mean an identifier of the started game (which is keeped over loading) is quite nice and preferable over the date.

But I don't want to rework the entire autosave logic. I think for now this is improvement enough.

Nordsoft91:
> Without uid you will overwrite with the new game the first files from the beginning. Without the uid the only possible way to know if it's the old or new game is to look on the file change date. Which is much confusing I find.

Yes, and this is current behaviour. By introducing uuid you broke scenario, where as user I want to have Autosave_111, 112, 113, 114 without storing all plays and replays. If they are stored in folders, it's fine, but in one folder - disaster. 

Improvement with folders is very good. But breaking existing behavior without folders is not good. IF uuid will be needed in future it can be introduced later, but now it could be harmful.

Laserlicht:
@Nordsoft91 Should be ready now.

client/CPlayerInterface.cpp
```diff
@@ -232,7 +230,7 @@ void CPlayerInterface::performAutosave()
 					+ std::to_string(cb->getDate(Date::WEEK))
 					+ std::to_string(cb->getDate(Date::DAY_OF_WEEK));
 
-			cb->save("Saves/Autosave/" + prefix + "Autosave_" + id + "_" + stringifiedDate);
+			cb->save("Saves/Autosave/" + prefix + "Autosave_" + stringifiedDate);
```
krs0, line 237: "Autosave_" seems redundant, since we are in a folder called /Autosave.

Laserlicht: Yes, what do you suggest? Only counter / day?

krs0: prefix + date should do the same thing.

Nordsoft91 (approved):

client/CPlayerInterface.cpp
```diff
@@ -212,7 +213,15 @@ void CPlayerInterface::performAutosave()
 			prefix = settings["general"]["savePrefix"].String();
 			if(prefix.empty())
 			{
-				prefix = cb->getMapHeader()->name.substr(0, 5) + "_";
+				std::string name = cb->getMapHeader()->name;
+				std::wstring_convert<std::codecvt_utf8<char32_t>, char32_t> conv;
+				int txtlen = conv.from_bytes(name).size();
```
IvanSavenko, line 218: Can you convert these two lines into new function in TextOperations namespaces,
e.g. 
`size_t DLL_LINKAGE getUnicodeCharactersCount(const std::string & text);`

IvanSavenko (approved):

PR #2664, opened 2023-08-24
Fix AI freeze regression

Nordsoft91:


Nordsoft91:
Merging to have fix in next nightly

PR #2668, opened 2023-08-24
Remove "currentPlayer" from gamestate

IvanSavenko:
Another simturns step, replace checking for currently active player with checking whether player is active

- Now gamestate provides isPlayerMakingTurn check instead of one that directly returns active player
- Cleared up validation of netpacks by server, e.g. always check that packet was send by correct player and that player owns affected object

### DIFF
--- a/client/adventureMap/TurnTimerWidget.h
+++ b/client/adventureMap/TurnTimerWidget.h
@@ -39,13 +39,13 @@ private:
 	std::set<int> notifications;
 	
 	std::shared_ptr<DrawRect> buildDrawRect(const JsonNode & config) const;
-	
+
 public:
 
 	void show(Canvas & to) override;
 	void tick(uint32_t msPassed) override;
 	
-	void setTime(int time);
+	void setTime(PlayerColor player, int time);
 
 	TurnTimerWidget();
 };
--- a/server/NetPacksServer.cpp
+++ b/server/NetPacksServer.cpp
@@ -37,76 +37,80 @@ void ApplyGhNetPackVisitor::visitSaveGame(SaveGame & pack)
 
 void ApplyGhNetPackVisitor::visitEndTurn(EndTurn & pack)
 {
-	if (!gh.hasPlayerAt(pack.player, pack.c))
-		gh.throwAndComplain(&pack, "No such pack.player!");
-
+	gh.throwIfWrongPlayer(&pack);
 	result = gh.turnOrder->onPlayerEndsTurn(pack.player);
 }
 
 void ApplyGhNetPackVisitor::visitDismissHero(DismissHero & pack)
 {
-	gh.throwOnWrongOwner(&pack, pack.hid);
+	gh.throwIfWrongOwner(&pack, pack.hid);
 	result = gh.removeObject(gh.getObj(pack.hid));
 }
 
 void ApplyGhNetPackVisitor::visitMoveHero(MoveHero & pack)
 {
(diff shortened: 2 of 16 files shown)


client/adventureMap/TurnTimerWidget.h
```diff
@@ -39,13 +39,13 @@ class TurnTimerWidget : public InterfaceObjectConfigurable
...
 	std::shared_ptr<DrawRect> buildDrawRect(const JsonNode & config) const;
-	
+
 public:
 
 	void show(Canvas & to) override;
 	void tick(uint32_t msPassed) override;
 	
-	void setTime(int time);
+	void setTime(PlayerColor player, int time);
```
Nordsoft91, line 48: could you please take signature from https://github.com/vcmi/vcmi/pull/2674 so we can avoid merge conflicts?

server/NetPacksServer.cpp
```diff
@@ -37,76 +37,80 @@ void ApplyGhNetPackVisitor::visitSaveGame(SaveGame & pack)
...
 {
-	gh.throwOnWrongOwner(&pack, pack.hid);
+	gh.throwIfWrongOwner(&pack, pack.hid);
 	result = gh.removeObject(gh.getObj(pack.hid));
 }
 
 void ApplyGhNetPackVisitor::visitMoveHero(MoveHero & pack)
 {
-	result = gh.moveHero(pack.hid, pack.dest, 0, pack.transit, gh.getPlayerAt(pack.c));
+	gh.throwIfWrongOwner(&pack, pack.hid);
```
Nordsoft91, line 52: removed here, otherwise shows system error if movement is interrupted by timer, because client may receive YourTurn after TryMoveHero https://github.com/vcmi/vcmi/pull/2576/files#diff-a3482c9af9fda544fe0aaddadc25fd201705f93609f5d512899799c8ec879703

IvanSavenko: So? Let's allow players to hack the game and move heroes of another players?

client/adventureMap/TurnTimerWidget.h
IvanSavenko: I can, but shouldn't this be more logical order?

nullkiller: In this case setPlayerTime may be a better name
Btw will we have a separate timer for each player now as later they may have to act in parallel?

IvanSavenko: If we want to do it "hd mod way" then yes.
Although HD mod displays timers for *all* players, including those who are not making turn - it always shows base timer I think.

nullkiller: I do not know HD mod well, but came into same conclusion)

server/NetPacksServer.cpp
Nordsoft91: any evidence that removing this check allows hacking game?

IvanSavenko: Look into code if this is not obvious?

CGameHandler::moveHero does not makes any checks for hero owner. So, malicious client can send MoveHero netpack, with `pack.hid` set to hero of another player and server will happily apply this pack

IvanSavenko: And if this result in error message then something wrong with caller code.
Why CPlayerInterface sends pack with pack.player set to something other than owner of our hero?

Nordsoft91: you are probably right, the fix was in another place

Nordsoft91 (approved):

PR #2673, opened 2023-08-25
Concurrent battles support

IvanSavenko:
Goal is to remove all references to "curB" and other ways to access ongoing battle and replace them with access by ID. General assumptions:
- It should be possible for multiple battles to happen in the same time
- Single player can only be engaged into a single battle
- Except for neutral "player", who can be present in multiple battles (e.g. several battles with neutrals)
- Battles are not serializeable - it is not possible to save game while there is an ongoing battle (can be changed in future, but low-priority task for now)

Current progress:
- [x] multiple battles in GameState
- [x] multiple battles in server logic
- [x] remove/replace CCallbackBase::battle member
- [x] multiple battles in client logic
- [x] ensure that all battle-related netpacks always receive correct BattleID
- [x] test everything


IvanSavenko:
Mostly done, will do another testing round, but basic functionality seems to be working.

rilian-la-te:
@IvanSavenko please, run UnitTests and send failures here.

IvanSavenko:
I can try running tests, but have zero interest in fixing any failures to be honest... Right now this is just too much maintenance pain for too little gain

rilian-la-te:
@IvanSavenko I suggest running and fixing relevant tests to everyone who change core functionality, because even if you do not like tests in this state (requires H3 data, cannot ran in CI), tests are helpful to check some aspects of core functionality like spells. 

I think not adding tests is OK, but breaking already working ones is bad. I think we should not break tests regularly, and then make PR with "fixinig tests again".

wb180:
Tried just to play single player map: on quick battle replay: in debug - assert, in release just freeze:
Screenshot_20230901_205456
Screenshot_20230901_205710

IvanSavenko:
Looks like battle replay. Haven't tested that yet.

PR #2674, opened 2023-08-25
Chess Timer UI and battle timer fixes

Nordsoft91:
Very simple UI to control all our timers. Can do minor improvements, but won't do redesign for now (pop-up window, etc).

https://github.com/vcmi/vcmi/issues/1364

Снимок экрана 2023-08-26 в 03 15 06


krs0:
A screenshot would help :)

IvanSavenko:
- Please add at the very least popup tooltips for timers or some other way to indicate which value corresponds to which timer
- I assume you're placing this in vcmi-extras? Ugh... And what happens if host has extras and uses chess timers while guest does not have extras?
- What about remaining issues with turn timers, e.g. concurrency fix?

krs0:
Looks good.... would be nice if somehow info on each timer could be displayed all the time. Since there is not enough space maybe 1 letter? T for turn , C for creature....
Or a suggestive icon? But have no idea how to symbolize turn so everyone understands.

Another idea depending on how much space is left, add plus minus arrow buttons for each time part? One for minutes, one for seconds

krs0:
Or just display the time and all edit will be done in a popup window

Nordsoft91:
> Please add at the very least popup tooltips for timers or some other way to indicate which value corresponds to which timer

Done

> I assume you're placing this in vcmi-extras? Ugh... And what happens if host has extras and uses chess timers while guest does not have extras?

game still can proceed with timers

> What about remaining issues with turn timers, e.g. concurrency fix?

Basic functionality is ready, broader testing is needed from nightly builds.
I'm capturing known issues in to-do list here, if it something I'm not aware, please describe also there https://github.com/vcmi/vcmi/issues/1364

IvanSavenko:
What will the guest without extras see when host configures chess timers?

Nordsoft91:
> Or a suggestive icon? But have no idea how to symbolize turn so everyone understands.
> Another idea depending on how much space is left, add plus minus arrow buttons for each time part? One for minutes, one for seconds

I was also thinking about it. Those are good, but problem is that somebody needs to draw all those icons/buttons but I'm quite bad designer :) Anyway, design doesn't impact functionality and can be done later, after issues with timers themselves will be fixed

Nordsoft91:
> What will the guest without extras see when host configures chess timers?

OH3 interface. Turn timer change will be visible, other timers will be applied silently

config/schemas/settings.json
```diff
@@ -474,7 +474,7 @@
 				},
 				"defaultRepositoryURL" : {
 					"type" : "string",
-					"default" : "https://raw.githubusercontent.com/vcmi/vcmi-mods-repository/develop/vcmi-1.3.json",
+					"default" : "https://raw.githubusercontent.com/vcmi/vcmi-mods-repository/develop/vcmi-1.4.json",
```
Laserlicht, line 477: Please adjust also https://github.com/vcmi/vcmi/blob/3b06abd0d7d1534a507ebae111ca7003b8d5d189/android/vcmi-app/src/main/java/eu/vcmi/vcmi/ActivityMods.java#L53 if you change the repo.

client/lobby/RandomMapTab.cpp
```diff
@@ -125,6 +121,35 @@ RandomMapTab::RandomMapTab():
 	const JsonNode config(ResourceID("config/widgets/randomMapTab.json"));
 	build(config);
 	
+	//set combo box callbacks
+	if(auto w = widget<ComboBox>("templateList"))
```
IvanSavenko, line 125: What will happen if player with vcmi-extras from 1.3 uses RMG in 1.4?

client/adventureMap/TurnTimerWidget.cpp
```diff
@@ -95,10 +98,22 @@ void TurnTimerWidget::tick(uint32_t msPassed)
 	if(LOCPLINT && LOCPLINT->cb)
 	{
 		auto player = LOCPLINT->cb->getCurrentPlayer();
+		if(LOCPLINT->battleInt)
+		{
+			if(auto * stack = LOCPLINT->battleInt->stacksController->getActiveStack())
+				player = stack->getOwner();
```
IvanSavenko, line 104: Does this method returns correct owner for hypnotized units?

client/lobby/OptionsTab.cpp
```diff
@@ -43,20 +44,147 @@
...
+			tinfo.baseTimer = tpreset.at(0).Integer() * 1000;
+			tinfo.turnTimer = tpreset.at(1).Integer() * 1000;
+			tinfo.battleTimer = tpreset.at(2).Integer() * 1000;
+			tinfo.creatureTimer = tpreset.at(3).Integer() * 1000;
+			CSH->setTurnTimerInfo(tinfo);
+		}
+	});
 	
-	addCallback("setTurnLength", std::bind(&IServerAPI::setTurnLength, CSH, _1));
+	auto parseTimerString = [](const std::string & str){
```
IvanSavenko, line 61: Can you explain what this function does?
Why using `scanf`/`stoi` or boost lexical_cast was not an option?

client/lobby/OptionsTab.cpp
IvanSavenko, line 1: And main question of all - why can't be integrate this into vcmi codebase? Like we did with settings window, adventure map, hero's backpack and multiple other interface improvements?

Asking players to install vcmi-extras for everything is not a good approach.

dydzio0614:
the problem is some people want original UI, potential solution can be classicMode setting introduced in my artifacts backpack PR

IvanSavenko:
@dydzio0614 Option is definitely an option. As long as player can easily find it without searching on forums/discord/etc. Othervice we definitely going to see questions like "timer was set to 4:00 per turn, but it is actually not 4 minutes per turn" Because of course it is not! Host had extras and also set main timer / battle timer!

Nordsoft91:
we don't have such messages for other new RMG options, so assumption about having such complains is based on nothing.
I'm with @dydzio0614 here and believe that vcmi without mods shall give as closer to OH3 as possible UI and experience.
This is how it works now and in the past and I cannot say that agree with approach we did with settings window, adventure map, hero's backpack and multiple other interface improvements.
And not only because of preference, but also because it's extra content, which includes assets inflate our package size.

Let's move conversation out of this thread about whether we shall give OH3 behavior+extas or get away from it because it doesn't related to PR. As for this PR: once if decide to change, it's just moving submod from extras to essentials.

IvanSavenko:
> we don't have such messages for other new RMG options, so assumption about having such complains is based on nothing. 

We had endless questions about this with resolutions other than 800x600, some - about battlefield actions aka option to toggle creature spellcasting and few - about quick hero exchange buttons.

Also, RMG (right now) is purely single-player option. Sure, we have MP, but barely anybody plays it at the moment so we don't end up in situations where host uses option that guest can't see. 

> And not only because of preference, but also because it's extra content, which includes assets inflate our package size.

Total size "inflation" of adventure map is 0 (zero) bytes. Radial menu - 8 Kb. Total inflation of settings window is 50kb. That's 0.037% "inflation" compared to our 130 Mb install on Windows. 

Hero backpack likely to have approximately same size as settings window.

>  it's just moving submod from extras to essentials.

No, it is not. At the very least your mod definitely includes edited H3 assets which we can't simply integrate as it. Like everything else from extras.

IvanSavenko:
Alternative option:

Keep chess timer UI in vcmi-extras. BUT actually enforce our license (GPL / CC-BY-SA) on any new additions to vcmi-extras repository so we can integrate anything from it into vcmi core. Fair?

Nordsoft91:
@IvanSavenko, I understand, that you have your vision and want to force people to follow it by pointing what to do and what not to do, but you make too much pressure and devalue contribution.
Everybody contribute as they can and how they want. Let's try to avoid such conversations and stay productive.

client/battle/BattleWindow.cpp
```diff
@@ -51,7 +51,7 @@ BattleWindow::BattleWindow(BattleInterface & owner):
 
 	REGISTER_BUILDER("battleConsole", &BattleWindow::buildBattleConsole);
 	
-	const JsonNode config(ResourceID("config/widgets/BattleWindow2.json"));
+	const JsonNode config(ResourceID("config/widgets/BattleWindow.json"));
```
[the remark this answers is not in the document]
Nordsoft91: I was thinking to limit vcmi-extras max compatibility by vcmi-1.3, so old mod won't be compatible with new version. But in the end decided to avoid this and limited compatibility only for extendedRmg submod https://github.com/vcmi-mods/vcmi-extras/commit/25dda0a591f93e2f8504f044b472576172407a01

Will revert

IvanSavenko:
@Nordsoft91 
Your quote;
> As for this PR: once if decide to change, it's just moving submod from extras to essentials.

But we can't. Our repository clearly states CC-BY-SA license for all assets. And did so for years. We simply can NOT integrate such submod.

Our mod guidelines by the way also state CC-BY-SA as requirement. So no changes in rules. Right?

client/lobby/RandomMapTab.cpp
Nordsoft91: old mod has compatibility restriction. If it not updated at all, there will be some missing elements

client/adventureMap/TurnTimerWidget.cpp
Nordsoft91: CStack has owner field, getOwner is getter for this field

client/lobby/OptionsTab.cpp
Nordsoft91: will add a comment

Nordsoft91:
Assets are from hota. I don't suggest to integrate such mod into essential files, I suggest to keep them in extras, same as for rmg options

> As for this PR: once if decide to change, it's just moving submod from extras to essentials.
I meant if nobody creates new assets

IvanSavenko:
> I meant if nobody creates new assets

1) Use FilledTexturePlayerColored class as background for timer selection area
2) Use this image (resize if necessary) as background for text input areas
frame-timer

Nordsoft91:
Once the https://github.com/vcmi/vcmi/pull/2687 done, we can proceed with changes you suggested inside the mod.

Nordsoft91:
Merge this part as more changes are coming

PR #2675, opened 2023-08-26
Fix market of time, crashes game

nullkiller:
Sample map. Visiting market of time was crashing game.
[SoD_Dzien_wistaka.zip](https://github.com/vcmi/vcmi/files/12445430/SoD_Dzien_wistaka.zip)

image

### DIFF
--- a/config/objects/generic.json
+++ b/config/objects/generic.json
@@ -870,7 +870,15 @@
 	},
 	"marketOfTime" : { // Unused/not implemented H3 object present on some RoE maps
 		"index" :50,
-		"handler": "generic"
+		"handler": "generic",
+		"types" : {
+			"object" : {
+				"index" : 0,
+				"aiValue" : 0,
+				"rmg" : {
+				}
+			}
+		}
 	},
 	"tavern" : {
 		"index" :95,


dydzio0614:
I would probably target beta with this fix

nullkiller:
Might be so this bug is here for ages)

IvanSavenko:
FYI - there is Market of Time on default map, "Titan's Winter"

As for a bug itself - should probably be fixed in a different way. If vcmi encounters unknown object it shouldn't crash. Some weird user-made maps have H3 objects with weird sub ID (which vcmi is not aware of). 
Wog maps also definitely have objects that vcmi is not aware of

nullkiller:
This one has valid subid of 0. And visible on the map. Because it is a part of original heroes files. Regarding other objects we should either handle them with valid JSON or skip during loading so they no longer present after map starting

nullkiller:
For instance we can allow some way to specify template for all possible subids or whatever. I just got crash, just fixed it. The thing you suggest is a bit bigger.

IvanSavenko:
What exactly caused crash?
Unsupported objects should be created as CGObjectInstance with no-op heroVisit method

nullkiller:
Mostly attempt to log it somewhere) It asks object name and crashes

nullkiller:
Also AI turn crashes with message can not find handler for 50:0

IvanSavenko:
> Mostly attempt to log it somewhere) It asks object name and crashes

We do have some checks for for non-existing handlers. But clearly not enough. Might be better to return some empty/default handler for such cases instead of hunting down every unchecked access.

Not against this PR as a quick fix measure though.

PR #2676, opened 2023-08-26
Flatpak - Use new screenshots

Laserlicht:


Alexander-Wilms:
I think having a screenshot showing the map as default would be nice. So maybe swap the first two images?

PR #2679, opened 2023-08-26
Smooth scroll

### LINKED ISSUE #2238: Smooth scrolling with touch
**Describe your proposal**
Implement smooth scrolling with touch (and also for mouse dragging).

**Additional context**
Example from [Julius](https://github.com/bvschaik/julius) on Android:

https://github.com/vcmi/vcmi/assets/13953785/b6e84f24-7c63-4a5e-a48c-4b5c8c088161

I had done a test before VCMI's input logic was changed: https://github.com/vcmi/vcmi/pull/2186

Laserlicht:
Fixes #2238

See here:

https://github.com/vcmi/vcmi/assets/13953785/9ab5180e-ffa1-4bf7-a016-9d68f9aa1545

### DIFF
--- a/client/mapView/MapView.cpp
+++ b/client/mapView/MapView.cpp
@@ -129,25 +129,24 @@ void MapView::onMapSwiped(const Point & viewPosition)
 }
 
 void MapView::postSwipe(uint32_t msPassed) {
-	if(!actions->dragActive && swipeHistory.size() > 0)
+	if(!actions->dragActive)
 	{
-		Point diff = Point(0, 0);
-		for (auto & x : swipeHistory)
-			diff += x.second;
+		if(swipeHistory.size() > 1)
+		{
+			Point diff = Point(0, 0);
+			for (auto & x : swipeHistory)
+				diff += x.second;
 
-		uint64_t timediff = swipeHistory.rbegin()->first - swipeHistory.begin()->first;
+			uint64_t timediff = swipeHistory.rbegin()->first - swipeHistory.begin()->first;
 
-		postSwipeAngle = diff.angle();
-		postSwipeSpeed = static_cast<double>(diff.length()) / static_cast<double>(timediff); // unit: pixel/millisecond
-
-		std::cout << postSwipeAngle << "   " << postSwipeSpeed << "   " << diff.x << "x" << diff.y << "   " << timediff << "\n";
+			postSwipeAngle = diff.angle();
+			postSwipeSpeed = static_cast<double>(diff.length()) / static_cast<double>(timediff); // unit: pixel/millisecond			
+		}
 		swipeHistory.clear();
 	}
 	if(postSwipeSpeed > 0.1) {
 		double len = postSwipeSpeed * static_cast<double>(msPassed);
 		Point delta = Point(len * cos(postSwipeAngle), len * sin(postSwipeAngle));
-
-		std::cout << len << "   " << delta.x << "x" << delta.y << "\n";
 
 		controller->setViewCenter(model->getMapViewCenter() + delta, model->getLevel());
 
--- a/client/mapView/MapView.cpp
+++ b/client/mapView/MapView.cpp
@@ -137,7 +137,7 @@ void MapView::postSwipe(uint32_t msPassed) {
 			for (auto & x : swipeHistory)
 				diff += x.second;
 
-			uint64_t timediff = swipeHistory.rbegin()->first - swipeHistory.begin()->first;
+			uint32_t timediff = swipeHistory.rbegin()->first - swipeHistory.begin()->first;
 
 			postSwipeAngle = diff.angle();
 			postSwipeSpeed = static_cast<double>(diff.length()) / static_cast<double>(timediff); // unit: pixel/millisecond			
--- a/client/mapView/MapView.cpp
+++ b/client/mapView/MapView.cpp
@@ -134,10 +134,17 @@ void MapView::postSwipe(uint32_t msPassed) {
 		if(swipeHistory.size() > 1)
 		{
 			Point diff = Point(0, 0);
-			for (auto & x : swipeHistory)
-				diff += x.second;
+			std::pair<uint32_t, Point> firstAccepted;
+			uint32_t now = GH.input().getTicks();
+			for (auto & x : swipeHistory) {
+				if(now - x.first < 150) { // only the last 150 ms are catched
+					if(firstAccepted.first == 0)
+						firstAccepted = x;
+					diff += x.second;
+				}
+			}
 
-			uint32_t timediff = swipeHistory.rbegin()->first - swipeHistory.begin()->first;
+			uint32_t timediff = swipeHistory.rbegin()->first - firstAccepted.first;
 
 			postSwipeAngle = diff.angle();
 			postSwipeSpeed = static_cast<double>(diff.length()) / static_cast<double>(timediff); // unit: pixel/millisecond			
--- a/client/mapView/MapView.cpp
+++ b/client/mapView/MapView.cpp
@@ -87,7 +87,7 @@ void MapView::tick(uint32_t msPassed)
 {
 	postSwipe(msPassed);
 
-	controller->tick(msPassed);
+	BasicMapView::tick(msPassed);
 }
 
 void MapView::show(Canvas & to)
@@ -123,7 +123,7 @@ void MapView::onMapScrolled(const Point & distance)
 
 void MapView::onMapSwiped(const Point & viewPosition)
 {
-	swipeHistory[GH.input().getTicks()] = viewPosition;
+	swipeHistory.push_back(std::pair<uint32_t, Point>(GH.input().getTicks(), viewPosition));
 
 	controller->setViewCenter(model->getMapViewCenter() + viewPosition, model->getLevel());
 }
@@ -150,7 +150,8 @@ void MapView::postSwipe(uint32_t msPassed) {
 			postSwipeSpeed = static_cast<double>(diff.length()) / static_cast<double>(timediff); // unit: pixel/millisecond			
 		}
 		swipeHistory.clear();
-	}
+	} else
+		postSwipeSpeed = 0.0;
 	if(postSwipeSpeed > 0.1) {
 		double len = postSwipeSpeed * static_cast<double>(msPassed);
 		Point delta = Point(len * cos(postSwipeAngle), len * sin(postSwipeAngle));
--- a/client/mapView/MapView.cpp
+++ b/client/mapView/MapView.cpp
@@ -163,7 +163,7 @@ void MapView::postSwipe(uint32_t msPassed) {
 
 		controller->setViewCenter(model->getMapViewCenter() + delta, model->getLevel());
 
-		postSwipeSpeed /= 1.1;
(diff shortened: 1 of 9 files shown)


Laserlicht:
If someone has a good substitute for `postSwipeSpeed /= 1.1;`, gladly report.

Works well, but is not frame rate independent. With `msPassed` this should doable.

client/mapView/MapView.cpp
```diff
@@ -82,6 +83,13 @@ void BasicMapView::showAll(Canvas & to)
 	render(to, true);
 }
 
+void MapView::tick(uint32_t msPassed)
+{
+	postSwipe(msPassed);
+
+	controller->tick(msPassed);
```
Nordsoft91, line 90:
```suggestion
	BasicMapView::tick(msPassed);
```

client/mapView/MapView.cpp
```diff
@@ -115,9 +123,37 @@ void MapView::onMapScrolled(const Point & distance)
...
+	{
+		if(swipeHistory.size() > 1)
+		{
+			Point diff = Point(0, 0);
+			for (auto & x : swipeHistory)
+				diff += x.second;
+
+			uint32_t timediff = swipeHistory.rbegin()->first - swipeHistory.begin()->first;
+
+			postSwipeAngle = diff.angle();
```
Nordsoft91, line 149: what if both diff.x and diff.y are zeros? Maybe good to check this case explicitly

client/mapView/MapView.cpp
```diff
@@ -115,9 +123,44 @@ void MapView::onMapScrolled(const Point & distance)
...
+
+			uint32_t timediff = swipeHistory.rbegin()->first - firstAccepted.first;
+
+			postSwipeAngle = diff.angle();
+			postSwipeSpeed = static_cast<double>(diff.length()) / static_cast<double>(timediff); // unit: pixel/millisecond			
+		}
+		swipeHistory.clear();
+	}
+	if(postSwipeSpeed > 0.1) {
+		double len = postSwipeSpeed * static_cast<double>(msPassed);
```
Nordsoft91, line 155: what if FPS is quite low so msPassed is large? For example, what if you have breakpoint in client while moving map?

client/mapView/MapView.h
```diff
@@ -49,7 +49,14 @@ class MapView : public BasicMapView
 {
 	std::shared_ptr<MapViewActions> actions;
 
+	std::map<uint32_t, Point> swipeHistory;
```
Nordsoft91, line 52: not quite optimal to use std::map. If I understand logic here properly, you insertion to the end, iterating over element and access to last element. If so, the best performance would give std::list

```suggestion
	std::list<std::pair<uint32_t, Point>> swipeHistory;
```

client/mapView/MapView.h
```diff
@@ -49,7 +49,14 @@ class MapView : public BasicMapView
 {
 	std::shared_ptr<MapViewActions> actions;
 
+	std::map<uint32_t, Point> swipeHistory;
+	double postSwipeAngle;
+	double postSwipeSpeed;
```
Nordsoft91, line 54:
```suggestion
	double postSwipeAngle = 0.0;
	double postSwipeSpeed = 0.0;
```

client/mapView/MapView.h
IvanSavenko: List is *almost never* optimal. Exceptions do exist but they are extremely rare.
If you want something with better performance then you need vector. Hard to beat its guaranteed memory locality.

But since this is not performance-critical code using most convenient containers is better idea than most optimal one.

client/mapView/MapView.cpp
Laserlicht: Results in 0. And in this case the speed is also 0. So no problem in this case.

client/mapView/MapView.cpp
Laserlicht: Should work as indented (`pixel/millisecond` -> `pixel`). The reason for this is exactly fps independence.

client/mapView/MapView.h
Laserlicht: Changed to vector

Laserlicht:
Tested on Android: really smooth experience. :)

client/mapView/MapView.cpp
IvanSavenko: Specifically to avoid issues with debugger, tick() and as result - this method will never receive value greater than 100 ms

client/mapView/MapView.cpp
```diff
@@ -115,9 +124,49 @@ void MapView::onMapScrolled(const Point & distance)
...
+			uint32_t now = GH.input().getTicks();
+			for (auto & x : swipeHistory) {
+				if(now - x.first < 150) { // only the last 150 ms are catched
+					if(firstAccepted.first == 0)
+						firstAccepted = x;
+					diff += x.second;
+				}
+			}
+
+			uint32_t timediff = swipeHistory.rbegin()->first - firstAccepted.first;
```
IvanSavenko, line 149:
```suggestion
			uint32_t timediff = swipeHistory.back().first - firstAccepted.first;

```

client/mapView/MapView.cpp
```diff
@@ -115,9 +124,49 @@ void MapView::onMapScrolled(const Point & distance)
...
+		if(swipeHistory.size() > 1)
+		{
+			Point diff = Point(0, 0);
+			std::pair<uint32_t, Point> firstAccepted;
+			uint32_t now = GH.input().getTicks();
+			for (auto & x : swipeHistory) {
+				if(now - x.first < 150) { // only the last 150 ms are catched
+					if(firstAccepted.first == 0)
+						firstAccepted = x;
+					diff += x.second;
```
IvanSavenko, line 145: Do we need to accumulate?
Why can't we compare firstAccepted.second and swipeHistory.back().second?

client/mapView/MapView.cpp
```diff
@@ -115,9 +124,49 @@ void MapView::onMapScrolled(const Point & distance)
...
+void MapView::postSwipe(uint32_t msPassed) {
+	if(!actions->dragActive)
+	{
+		if(swipeHistory.size() > 1)
+		{
+			Point diff = Point(0, 0);
+			std::pair<uint32_t, Point> firstAccepted;
+			uint32_t now = GH.input().getTicks();
+			for (auto & x : swipeHistory) {
+				if(now - x.first < 150) { // only the last 150 ms are catched
```
IvanSavenko, line 142: Can you replace '150' with constant with descriptive name, e.g. `catchIntervalMilliseconds`?
Variable names are better documentation than comments

client/mapView/MapView.cpp
```diff
@@ -115,9 +124,49 @@ void MapView::onMapScrolled(const Point & distance)
...
+		swipeHistory.clear();
+	} else
+		postSwipeSpeed = 0.0;
+	if(postSwipeSpeed > 0.1) {
+		double len = postSwipeSpeed * static_cast<double>(msPassed);
+		Point delta = Point(len * cos(postSwipeAngle), len * sin(postSwipeAngle));
+
+		controller->setViewCenter(model->getMapViewCenter() + delta, model->getLevel());
+
+		postSwipeSpeed /= 1 + msPassed * 0.006;
```
IvanSavenko, line 166: Also - please replace `0.006` with constant with descriptive name

IvanSavenko (approved):

client/mapView/MapView.cpp
Laserlicht: Yes it's needed. Because each entry in `swipeHistory` is relative to the entry before. So, if we want to get the correct speed over multiple entries (time range), we have to sum up.

client/mapView/MapView.cpp
```diff
@@ -115,9 +124,49 @@ void MapView::onMapScrolled(const Point & distance)
 
 void MapView::onMapSwiped(const Point & viewPosition)
 {
+	if(settings["adventure"]["smoothDragging"].Bool())
+		swipeHistory.push_back(std::pair<uint32_t, Point>(GH.input().getTicks(), viewPosition));
+
 	controller->setViewCenter(model->getMapViewCenter() + viewPosition, model->getLevel());
 }
 
+void MapView::postSwipe(uint32_t msPassed) {
```
IvanSavenko, line 133:
```suggestion
void MapView::postSwipe(uint32_t msPassed)
{
```

Laserlicht: thanks, overlooked :)

Nordsoft91:
On iOS there are glitches, see video (3-4 seconds and later some)

https://github.com/vcmi/vcmi/assets/9308612/034079bc-e150-4613-a0e7-e0268d1222d5

Laserlicht:
Hmm... Not reproducible on my phone and tablet with android.

Maybe something to do with the low fps?

Nordsoft91:
Set 30 fps - same. To reproduce try to generate bigger map and scroll by swipes several times in a row

https://github.com/vcmi/vcmi/assets/9308612/d0be5e74-3a4c-4976-88ef-d9e3c80c3963

Regarding low FPS, you should expect this. Some people reduce fps intentionally to reduce battery consumption

Laserlicht:
Is it possible also to record touch inputs?

The movement is overwritten. If you touch it extremly short - it doesn't have information about movement. But maybe I can ignore it...

Yes - if you mean this, than I can reproduce it. But come on. At least on my device: Even if I want: It is extremely hard to trigger and looks absolutely stupid.

Does the iphone perhaps have a not so high sampling rate from the touchscreen?

Laserlicht:
can you try it again after compilation?

Nordsoft91:
Still reproducible. On mac as well with mouse drag, no gestures

https://github.com/vcmi/vcmi/assets/9308612/3029533c-b4bb-43c2-b92b-c74ce38865d7

Laserlicht:
I cannot verify it (because I cannot see your clicks/touch).

But in slomo it seems you stopped the mouse/finger (before you released). If it is the case then it's intended. This is necessary for the normal dragging (stop before release).

Laserlicht:
Installed it again. This time on iPhone (and learned again, why I don't like it... Sideloading on iOS is crap...).

I cannot reproduce it also on iOS.

As stated before: It is intended and necessary to stop scrolling if you stop movement before release your finger.

Nordsoft91:
that's how it works with browser

https://github.com/vcmi/vcmi/assets/9308612/e84bc151-a963-41d3-82c7-fc9aa1b51efc

at first glance, I don't see any problem with algorithm, so maybe the postSwipeCatchIntervalMs is too small. Maybe you can increase it to 400 ms or make it configurable so I can play with number and find threshold similar to standard ios/macos behavior?

Laserlicht:
Was able to reproduce. At 60fps no problem (also on iOS). But at 20fps a problem on iOS and Android (not Win-PC).

Because of this I added a dynamic calculation. Hope it works.

Nordsoft91:
Now it seems too long :) I released mouse but map starts fly. Maybe something in the middle?

Also, suggestion: touching anywhere should immediately stop map if it's flying

Laserlicht:
> Now it seems too long :) I released mouse but map starts fly. Maybe something in the middle?

Adjusted

Nordsoft91 (approved):
perfect now

PR #2682, opened 2023-08-27
Dump gamestate to vmap

nullkiller:
I am not sure if I really need it. Still the problem exists of constantly breaking saves because we often make backward incompatible changes. This PR theoretically allows to dump save into vmap and open vmap in newer version. Sure day will get reset to 1 and resources as well and a lot of other things will be in their initial state.


Nordsoft91:
Alternatively what if follow @sxx approach and separate serialization version for gamestate and for network packs?

nullkiller:
I am not sure even I really need it because as @IvanSavenko mentioned too much things will be lost. In this case AI likely may behave differently so you will not be able to reproduce all bugs but only some.
controlai helps investigating save file. But sometimes having map in map editor may be helpful as well. But only if it will not take too much time to implement.

IvanSavenko:
@Nordsoft91 
While separate version for netpacks is not a bad idea it won't have any effects here. Changes to gamestate format are not uncommon, for example - this very PR which affects gamestate and as result - also netpacks.

nullkiller:
This thing can be made backward compatible more or less easily if you wish

PR #2683, opened 2023-08-27
dead area for zooming

### LINKED ISSUE #2594: Adventure map zoom reset on touch devices
**Describe your proposal**
Playing on iPad it's great to pinch zoom in / out the adventure map, but there should be a way to "reset" the zoom... like double tap with one or two fingers, or to use the settings - be able to set the zoom also there - there is no ingame setting currently

[Optional future feature addition] as an addition for consideration, the zoom should be for each hot-seat player individual and possibly saved in the save game?

Laserlicht:
fixes #2594

### DIFF
--- a/client/mapView/MapViewController.cpp
+++ b/client/mapView/MapViewController.cpp
@@ -90,13 +90,13 @@ void MapViewController::modifyTileSize(int stepsChange)
 	// we want to zoom in/out in fixed 10% steps, to allow player to return back to exactly 100% zoom just by scrolling
 	// so, zooming in for 5 steps will put game at 1.1^5 = 1.61 scale
 	// try to determine current zooming level and change it by requested number of steps
-	double currentZoomFactor = model->getSingleTileSize().x / 32.0;
+	double currentZoomFactor = targetTileSize.x / static_cast<double>(defaultTileSize);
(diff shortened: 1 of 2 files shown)


IvanSavenko (approved):
Tested on phone, works quite well. Thanks!

Nordsoft91:
Doesn't work well on macOS

https://github.com/vcmi/vcmi/assets/9308612/e373d757-29a9-40f1-acee-5b4c4c98f79c

Nordsoft91:
works well on iOS, couple suggestions to improve:
 - increase dead area slightly
 - add vibration tick (GH.input().hapticFeedback())

Laserlicht:
> Doesn't work well on macOS
> 2023-08-27.22.49.50.mp4

I think it is a scroll wheel-"emulation"? (pinch on touchpad?)

Nordsoft91:
Yes, it's touchpad. Not a pinch, but two fingers up and down

Laserlicht:
> Yes, it's touchpad. Not a pinch, but two fingers up and down
> 
> 

Okay, some specific gesture. Do you know is there is a solution?

Nordsoft91:
reproduced in 1.3.1, so yes - different bug

Laserlicht:
Ready I think.

PR #2685, opened 2023-08-27
Fix mana drain not refreshing battle hero window mana

dydzio0614:
Battle hero window did not have refreshed mana when MANA_DRAIN creature ability was triggering, as it uses another netpack (BattleTriggerEffect) than usual SetMana


IvanSavenko:
Perhaps we should instead generate ManaChanged netpack in such cases?

dydzio0614:
maybe, unsure where I should do it

PR #2686, opened 2023-08-27
BattleAI: fix freeze

nullkiller:
Freeze on archers attacking some town with towers


dydzio0614:
if this fix is applicable to beta branch then I would target beta

nullkiller:
> if this fix is applicable to beta branch then I would target beta

This is for my battle damage cache PR

PR #2692, opened 2023-08-28
Accept turn query

Nordsoft91:
YourTurn now is a query so player has to answer it to accept the turn.

Needed to stop timer in hot seat game

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -243,7 +243,7 @@ void CPlayerInterface::performAutosave()
 	}
 }
 
-void CPlayerInterface::yourTurn()
+void CPlayerInterface::yourTurn(QueryID queryID)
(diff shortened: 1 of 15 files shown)


client/CPlayerInterface.cpp
```diff
@@ -322,6 +322,9 @@ void CPlayerInterface::acceptTurn()
 		else
 			logGlobal->warn("Player has no towns, but daysWithoutCastle is not set");
 	}
+	
+	JsonNode reply(JsonNode::JsonType::DATA_NULL);
+	cb->sendQueryReply(reply, queryID);
```
IvanSavenko, line 327: Perhaps use selectionMade call? And not this weird version that sends Json for some reason?
```suggestion
	cb->selectionMade(0, queryID);
```

PR #2695, opened 2023-08-28
Add compatibility check with minor and patch versions

Nordsoft91:


IvanSavenko (dismissed):
No.


_No description provided._

Nordsoft91:
Fixes https://github.com/vcmi/vcmi/issues/2696

IvanSavenko:
Also - does your change actually fixes linked issue?
If I use `"max" : "1.3.2"` then I expect "1.3.2" to be compatible, but "1.3.3" not compatible. But according to signature of changed method, "patch" level is still ignored.

IvanSavenko (commented):
- what if mod has "1.3" specified as max, without patch version? Expected behavior is probably ignore patch version then. I think this is also how Qt version comparator works (that we use for Launcher)
- enum CHECK_MAJOR / CHECK_MAJOR_MINOR / CHECK_MAJOR_MINOR_PATCH would probably be more clear than `true, true` form.

In any case, do as you wish.

Nordsoft91:
Yeah, that's a problem - launcher interprets 1.3 as 1.3.0 and shows it as incompatible.
I think, this needs to be changed and way to check versions shall b unified for both launcher and client.
I'd like to make it later, when I will touch launcher for more improvements (https://github.com/vcmi/vcmi/issues/2697)
Снимок экрана 2023-08-28 в 18 40 14

PR #2698, opened 2023-08-28
Turn timer fixes

Nordsoft91:
- Fixed potential concurrency issues
 - Possible to set different timers independently
 - Tactic support
 - Use precision clocks
 - Pause timer when turn is passed in hot-seat mode
 - Fixed overlap between timer and fps counter
 - Timer strings localization


server/CGameHandler.cpp
```diff
@@ -999,18 +999,21 @@ void CGameHandler::run(bool resume)
...
+		clockLast += clockDuration;
 
 		for(PlayerColor player(0); player < PlayerColor::PLAYER_LIMIT; ++player)
 			if(gs->isPlayerMakingTurn(player))
-				turnTimerHandler.onPlayerMakingTurn(player, waitTime);
+				turnTimerHandler.onPlayerMakingTurn(player, timePassed);
 
 		if(gs->curB)
-			turnTimerHandler.onBattleLoop(waitTime);
+			turnTimerHandler.onBattleLoop(timePassed);
```
IvanSavenko, line 1014: Mutex should probably be locked in this method, not inside turnTimeHandler.
Or you need to remove all checks from here - for example, I am not sure if gs->isPlayerMakingTurn can be accessed from multiple threads

IvanSavenko (commented):
Also, please move all localization from mod to vcmi repo, to
Mods/vcmi/config/english.json

server/CGameHandler.cpp
IvanSavenko: Or perhaps - move all timer-related code from here (including all checks) into method inside turnTimeHandler, e.g. `TurnTimerHandler::update()`?

Nordsoft91:
> Also, please move all localization from mod to vcmi repo, to
> Mods/vcmi/config/english.json

https://github.com/vcmi-mods/vcmi-extras/pull/16

Mods/vcmi/config/vcmi/english.json
```diff
@@ -191,6 +191,15 @@
 	"vcmi.randomMapTab.widgets.teamAlignmentsLabel"  : "Team Alignments",
 	"vcmi.randomMapTab.widgets.roadTypesLabel"       : "Road Types",
 
+	"vcmi.optionsTab.widgets.chessFieldBase.help" : "{Base timer}\n\nCounts down when {turn timer} runs out, never resets. If timer expires, player ends turn.",
+	"vcmi.optionsTab.widgets.chessFieldTurn.help" : "{Turn timer}\n\nCounts down when player is making turn on adventure map, resets every turn. If {base timer} is enabled, unspent time will be added to it.",
+	"vcmi.optionsTab.widgets.chessFieldBattle.help" : "{Battle timer}\n\nCounts down during battles when {stack timer} runs out, resets every battle. if timer expires, currently acting stack will defend.",
+	"vcmi.optionsTab.widgets.chessFieldCreature.help" : "{Creature timer}\n\nCounts down when player is choosing action in battle, resets after each stack turn.",
```
IvanSavenko, line 197: @dydzio0614 @krs0 
Any suggestions on how to improve these tooltips?

krs0: I ran them through chatgpt... in the end it is a matter of choice.
@Nordsoft91 What do you say? I think these are somewhat clearer? But slight adjustments still could be made especially for brevity. Maybe `It resets to its initial value` could be just `It resets` like in original text? Although chat gpt strongly preffers the longer version :) and I am afraid to upset it.

```
	"vcmi.optionsTab.widgets.chessFieldBase.help" : "{Extra timer}\n\nStarts counting down when the {turn timer} reaches zero. It is set only once at the beginning of the game. When this timer reaches zero, the player's turn ends."
	"vcmi.optionsTab.widgets.chessFieldTurn.help" : "{Turn timer}\n\nStarts counting down when the player starts their turn on the adventure map. It is reset to its initial value at the start of each turn. If the {extra timer} is in use, any unused turn time will be added to it."
	"vcmi.optionsTab.widgets.chessFieldBattle.help" : "{Battle timer}\n\nCounts down during battles when the {stack timer} reaches zero. It is reset to its initial value at the start of each battle. If the timer reaches zero, the currently active stack will defend."
	"vcmi.optionsTab.widgets.chessFieldCreature.help" : "{Stack timer}\n\nStarts counting down when the player is selecting an action for the current stack during battle. It resets to its initial value after the stack's turn is completed."
```

krs0: I would rename a little the variables chessTimerFieldXXX.help.

PR #2710, opened 2023-08-29
Split artifacts bag button background and icon

dydzio0614:


krs0:
The button still looks missaligned, and if possible color to be more in tone with other buttons. Maybe ask on discord?
And the margins of the button look different from left and right buttons. The white glow.

image

PR #2711, opened 2023-08-29
Allow quick combat vs AI players

### LINKED ISSUE #2671: Quick combat between heroes does not work
I use VCMI 1.3.1 macOS version.
Check Quick Combat in System Options to keep it checked.
The battles between our heroes and neutral creatures are normal and manifest as fast battles.
However, the battle between our heroes and enemy heroes is directly into the battle scene, which is not normal.
This does not appear in the VCMI 1.2.x macOS release.

截屏2023-08-25 21 14 25

dydzio0614:
As in title, partially fixes #2671

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -2133,8 +2133,10 @@ void CGameHandler::setupBattle(int3 tile, const CArmedInstance *armies[2], const
 	engageIntoBattle(bs.info->sides[1].color);
 
 	auto lastBattleQuery = std::dynamic_pointer_cast<CBattleQuery>(queries.topQuery(bs.info->sides[0].color));
-	bool isHumanPlayer = bs.info->sides[1].color.isValidPlayer() && gs->getStartInfo()-> playerInfos.at(bs.info->sides[1].color).isControlledByHuman();
-	bs.info->replayAllowed = lastBattleQuery == nullptr && !isHumanPlayer;
+	bool isDefenderHuman = bs.info->sides[1].color.isValidPlayer() && getPlayerState(bs.info->sides[1].color)->human;
+	bool isAttackerHuman = getPlayerState(bs.info->sides[0].color)->human;
+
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Does not handles case where AI is attacker and human is defender

server/CGameHandler.cpp
```diff
@@ -2133,7 +2133,8 @@ void CGameHandler::setupBattle(int3 tile, const CArmedInstance *armies[2], const
 	engageIntoBattle(bs.info->sides[1].color);
 
 	auto lastBattleQuery = std::dynamic_pointer_cast<CBattleQuery>(queries.topQuery(bs.info->sides[0].color));
-	bs.info->replayAllowed = lastBattleQuery == nullptr && !bs.info->sides[1].color.isValidPlayer();
+	bool isHumanPlayer = bs.info->sides[1].color.isValidPlayer() && gs->getStartInfo()-> playerInfos.at(bs.info->sides[1].color).isControlledByHuman();
```
Nordsoft91, line 2136:
```suggestion
	bool isHumanPlayer = bs.info->sides[1].color.isValidPlayer() && getPlayerState(bs.info->sides[1].color)->human;
```

But if you'd like to enable it even if AI attacks player, you need to check if attacker is AI too

dydzio0614:
Manual replay for battles vs AI player does not work for now - @Nordsoft91 any idea if there is easy fix for this?

Nordsoft91:
Not sure. When battle is finished, sever sends query to both players, AI immediately responds to it. I think this is root cause.
Maybe you can check battle on server side:
 - if AI vs neutrals / AI vs AI - no change
 - if AI vs human - send query to human player only

Nordsoft91:
see BattleResultProcessor::endBattle

IvanSavenko:
This is "beta" so no battleprocessor yet.
@dydzio0614 
See  CGameHadler.cpp:611
```c++
	// in battles against neutrals, 1st player can ask to replay battle manually
	if (!gs->curB->sides[1].color.isValidPlayer())
```
I guess we should replace it to "if one (and only one) player is human, player can ask for battle replay"

PR #2720, opened 2023-08-31
Support for different backpack window capacities

SoundSSGood:
This is what the infinity backpack looks like
infinite
64 items limit.
64_lim
30 items limit.
30_lim

### DIFF
--- a/client/windows/CHeroBackpackWindow.cpp
+++ b/client/windows/CHeroBackpackWindow.cpp
@@ -24,22 +24,26 @@ CHeroBackpackWindow::CHeroBackpackWindow(const CGHeroInstance * hero)
 {
 	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
 
-	stretchedBackground = std::make_shared<CFilledTexture>("DIBOXBCK", Rect(0, 0, 410, 425));
-	pos.w = stretchedBackground->pos.w;
-	pos.h = stretchedBackground->pos.h;
-	center();
+	stretchedBackground = std::make_shared<CFilledTexture>("DIBOXBCK", Rect());
 
-
-	arts = std::make_shared<CArtifactsOfHeroBackpack>(/*Point(-100, -170)*/Point(10, 10));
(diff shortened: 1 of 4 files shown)


DjWarmonger:
Can you resize the window to make use of higher resolutions? Like, 10 x 10 and 12 x 12 variants - if needed.

IvanSavenko:
Frankly filters / sorting would be much more useful than larger window. 
(not insisting on either option though)

SoundSSGood:
> Can you resize the window to make use of higher resolutions? Like, 10 x 10 and 12 x 12 variants - if needed.

Yes, it's easy. If necessary, I can make the API for that

client/windows/CHeroBackpackWindow.cpp
```diff
@@ -24,22 +24,20 @@ CHeroBackpackWindow::CHeroBackpackWindow(const CGHeroInstance * hero)
...
 	pos.w = stretchedBackground->pos.w;
 	pos.h = stretchedBackground->pos.h;
 	center();
-
-
-	arts = std::make_shared<CArtifactsOfHeroBackpack>(/*Point(-100, -170)*/Point(10, 10));
-	arts->setHero(hero);
-	addSet(arts);
-
+	std::iter_swap(GH.createdObj.front()->children.begin(), GH.createdObj.front()->children.begin() + 2);
```
IvanSavenko, line 35: Why is this needed?

SoundSSGood: First the CArtifactsOfHeroBackpack and CButton are created, then CFilledTexture. CFilledTexture is at the beginning of GH.createdObj.front()->children and will be displayed on top.

IvanSavenko: Why can't we swap creation order and create background first?
If you don't know window size you can update it later, e.g:
```cpp
stretchedBackground = std::make_shared<CFilledTexture>( ... )
(create rest of widgets)
stretchedBackground->pos = actualPos
```

SoundSSGood: Right. this approach works

IvanSavenko (approved):

PR #2722, opened 2023-08-31
Launcher mod double click behavior

Nordsoft91:
Use double-click to install/update/enable/disable mods

Also, increate handler width to better control views' sizes


IvanSavenko:
Not sure about "updating' part - player might want to enable/disable mod instead of update. Might be better to keep update separate.

Perhaps also account for clicked column? Click on column with install/update icon will always trigger install/update while clicking on column with enabled status icon will always enable/disable mod.

Issue #2697, opened 2023-08-28
Improve way to check compatibility with mods


Nordsoft91:
Currently it's possible to specify compatibility versions in mod.json, so launcher and client can verify it before activating.
```
"compatibility":
{
  "min": 1.2.0,
  "max": 1.3.2
}
```

However, if mode will omit version part, assuming "any" it will be interpret by both launcher and client differently:
```
"compatibility":
{
  "min": 1.2, //assuming 1.2.x, will be interpret as 1.2.0 but no problem
  "max": 1.3 //assuming 1.3.x, will be interpret as 1.3.0, so mod become incompatible with 1.3.1
}
```

Suggest to unify approaches used for compatibility check in both launcher and client and support behavior described above.

Another suggestion - change function signature. Current:
`bool CModVersion::compatible(const CModVersion & other, bool checkMinor, bool checkPatch) const`
change by `bool CModVersion::compatible(const CModVersion & other, int checkOptions = CHECK_MAJOR | CHECK_MAJOR_MINOR | CHECK_MAJOR_MINOR_PATCH) const`

krs0:
So max 1.3 means now all 1.2.x versions?

PR #2723, opened 2023-09-01
Launcher & lib versions unification

Nordsoft91:
Fixes https://github.com/vcmi/vcmi/issues/2697

### DIFF
--- a/lib/modding/CModVersion.cpp
+++ b/lib/modding/CModVersion.cpp
@@ -61,16 +61,14 @@ std::string CModVersion::toString() const
 
 bool CModVersion::compatible(const CModVersion & other, bool checkMinor, bool checkPatch) const
 {
-	if(minor == Any || other.minor == Any)
-		checkMinor = false;
-	if(patch == Any || other.patch == Any)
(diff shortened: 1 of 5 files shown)


lib/modding/CModVersion.cpp
```diff
@@ -45,19 +45,49 @@ CModVersion CModVersion::fromString(std::string from)
...
+	}
+	return res;
 }
 
 bool CModVersion::compatible(const CModVersion & other, bool checkMinor, bool checkPatch) const
 {
+	if(minor == Any || other.minor == Any)
+		checkMinor = false;
+	if(patch == Any || other.patch == Any)
+		checkPatch = false;
```
IvanSavenko, line 67: Perhaps use separate variables instead of changing function input parameters? Should help a bit with code clarity.
```suggestion
bool doCheckMinor = checkMinor && minor != Any && other.minor != Any;
bool doCheckPatch = checkPatch && patch != Any && other.patch != Any;
```
(and obviously use these two variables below instead of function parameters)

Issue #2642, opened 2023-08-22
Add size for mods


Nordsoft91:
When your on slow network or you use mobile internet, it good to see mods' sizes. Also it's good to see how much size mods keep on device. Currently launcher cannot identify download size because GitHub doesn't provide it

Proposition:
- [ ] Define size for each mod in its json in Mb, fine to have approximation to avoid frequent updates (field is already defined, but missed in mods)
- [ ] Show size in launcher for both installed and downloadable mods (whether as column or in detailed info)

Note: download progress bar will work from the box since mod size is defined

PR #2726, opened 2023-09-01
Support local and download mod size

Nordsoft91:
Fixes https://github.com/vcmi/vcmi/issues/2642


launcher/modManager/cmodmanager.cpp
```diff
@@ -87,15 +87,24 @@ void CModManager::loadMods()
...
 		if(CResourceHandler::get()->existsResource(resID))
 		{
+			//calculate mod size
+			qint64 total = 0;
+			ResourceID resDir(CModInfo::getModDir(modname), EResType::DIRECTORY);
+			if(CResourceHandler::get()->existsResource(resDir))
+			{
+				for(QDirIterator iter(QString::fromStdString(CResourceHandler::get()->getResourceName(resDir)->string()), QDirIterator::Subdirectories); iter.hasNext(); iter.next())
+					total += iter.fileInfo().size();
+				total /= 1024; //to Kb
```
IvanSavenko, line 97: Mod size is stored as kilobytes?

launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -535,7 +538,7 @@ void CModListView::on_updateButton_clicked()
 		auto mod = modModel->getMod(name);
 		// update required mod, install missing (can be new dependency)
 		if(mod.isUpdateable() || !mod.isInstalled())
-			downloadFile(name + ".zip", mod.getValue("download").toString(), "mods");
+			downloadFile(name + ".zip", mod.getValue("download").toString(), "mods", mod.getValue("size").toDouble() * 1024 * 1024);
```
IvanSavenko, line 541: Mod size is stored as megabytes?

IvanSavenko (commented):
1) Can you always store mode size in same units? Bytes for example
2) Are you planning to add mod size for all our ~100 mods?

Nordsoft91:
> Can you always store mode size in same units? Bytes for example

It's not very convenient to specify size in bytes. I changed from Kb to Mb in order simplify size management for modders. It's not supposed to be always up-to-date

> Are you planning to add mod size for all our ~100 mods?

no

IvanSavenko:
> It's not very convenient to specify size in bytes. I changed from Kb to Mb in order simplify size management for modders. It's not supposed to be always up-to-date

From what I see in code, in one place we use kb and in other - mb. Can you change that for consistency

> > Are you planning to add mod size for all our ~100 mods?
> 
> no

Great! Misiokles will be thrilled when he hears that he now need to update all 100 mods!

Please, reconsider. Writing script to batch download and update json should be simple. Don't leave half-done features.

Nordsoft91:
> From what I see in code, in one place we use kb and in other - mb. Can you change that for consistency?

QT uses bytes, but we don't want to show size in bytes to users and we don't want to ask users to specify sizes in bytes.
Thus we need to do convert them. But maybe specifying mod size in Kb is also Ok?

> Great! Misiokles will be thrilled when he hears that he now need to update all 100 mods!

Hears from who? Are you going to assign to-do's for everybody?

> Please, reconsider. Writing script to batch download and update json should be simple. Don't leave half-done features.

Can I live without your orders and management please? Thanks!

IvanSavenko:
> QT uses bytes, but we don't want to show size in bytes to users and we don't want to ask users to specify sizes in bytes. Thus we need to do convert them. But maybe specifying mod size in Kb is also Ok?

From what I see: 
- "localSize": generated by Launcher, uses kilobytes
- "size": loaded from repository, uses megabytes. (by the way, perhaps rename to "downloadSize" for clarity?
- And there is CModEntry::sizeToString method which expects size in kilobytes.

So nothing from Qt itself, we can changes all of these to use the same scale, whether it is kilobytes or megabytes.

> > Great! Misiokles will be thrilled when he hears that he now need to update all 100 mods!
> 
> Hears from who? Are you going to assign to-do's for everybody?

Then who is going to update mods all our mods? Santa? A lot of them don't have active developer at the moment. As result misiokles is the only one who maintains such mods.

Nordsoft91:
So the only change is "size" from Kb to Mb. Remaining are in Kb as it was before. I can change size back from Kb to Mb but not sure how convenient it would be. "localSize" is purely internal thing. I can change it to bytes, but again don't see the benefit. 

> Then who is going to update mods all our mods? Santa? A lot of them don't have active developer at the moment. As result misiokles is the only one who maintains such mods.

Things don't get worse without size field, right? At the same time having size for some mods would be beneficial.
So if you'd like me to share my plans - I'm going to specify size at least for some of mods (expansions and some towns mostly) leaving others as it is.

Nordsoft91:
I personally like Mb more and also want to change progress bar from Kb to Mb, but open for other opinions.
Open questions:
 - which units to use in mod field
 - which units to show on progress bar "Downloading .... (xxx Kb out of xxx KB)"

IvanSavenko:
> So if you'd like me to share my plans - I'm going to specify size at least for some of mods (expansions and some towns mostly) leaving others as it is.

Please also add most popular mods into list: extras, wog, tow, hota

>  which units to use in mod field "size"

Adapt to mod size? This is also what CModEntry::sizeToString does:
- values below 1 Mb are shown as "kB"
- values below 1 Gb are shown as "MB"
etc.

Displaying size of tiny mods as "0.01 MB" is weird.

> which units to show on progress bar "Downloading .... (xxx Kb out of xxx KB)"

Either adapt as well or use "MB" - downloads less than 1 Mb are generally fast enough so no need to show sub-mb values.

Nordsoft91:
> Please also add most popular mods into list: extras, wog, tow, hota

those are already in the "list"

> Adapt to mod size? This is also what CModEntry::sizeToString does:

not sure I've got the idea. If mod xxx has field `"size": 100`, should it be interpret as Mb or in Kb? It's not about displaying, because it works exactly as you described.

> Either adapt as well or use "MB" - downloads less than 1 Mb are generally fast enough so no need to show sub-mb values.

Ok

IvanSavenko:
Ah, in config? Yeah, MB probably makes sense. As long as both localSize and downloadSize use same scale.

Laserlicht:
Some other idea: Maybe we can add filesizes to https://github.com/vcmi/vcmi-mods-repository and update it automatically over ci/cd (download each mod and get filesize)?

Just a idea...

launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -26,6 +26,11 @@
 #include "../../lib/CConfigHandler.h"
 #include "../../lib/Languages.h"
 
+inline double mbToBytes(double mb)
```
IvanSavenko, line 29: - Use static for free functions
- Despite its name, "inline" is not about compiler optimization, but about linking multiple compilation units - no need for "inline" in local functions.
```suggestion
static double mbToBytes(double mb)
```

launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -243,8 +248,11 @@ QString CModListView::genModInfoText(CModEntry & mod)
 	result += replaceIfNotEmpty(mod.getValue("installedVersion"), lineTemplate.arg(tr("Installed version")));
 	result += replaceIfNotEmpty(mod.getValue("latestVersion"), lineTemplate.arg(tr("Latest version")));
 
-	if(mod.getValue("size").isValid())
-		result += replaceIfNotEmpty(CModEntry::sizeToString(mod.getValue("size").toDouble()), lineTemplate.arg(tr("Download size")));
+	if(mod.getValue("localSize").isValid())
+		result += replaceIfNotEmpty(CModEntry::sizeToString(mod.getValue("localSize").toDouble()), lineTemplate.arg(tr("Size")));
+	if((mod.isAvailable() || mod.isUpdateable()) && mod.getValue("size").isValid())
+		result += replaceIfNotEmpty(CModEntry::sizeToString(mbToBytes(mod.getValue("size").toDouble())), lineTemplate.arg(tr("Download size")));
+
```
IvanSavenko, line 255: Perhaps store "localSize" in megabytes as well instead of bytes, for consistency?

launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -243,8 +248,11 @@ QString CModListView::genModInfoText(CModEntry & mod)
 	result += replaceIfNotEmpty(mod.getValue("installedVersion"), lineTemplate.arg(tr("Installed version")));
 	result += replaceIfNotEmpty(mod.getValue("latestVersion"), lineTemplate.arg(tr("Latest version")));
 
-	if(mod.getValue("size").isValid())
-		result += replaceIfNotEmpty(CModEntry::sizeToString(mod.getValue("size").toDouble()), lineTemplate.arg(tr("Download size")));
+	if(mod.getValue("localSize").isValid())
+		result += replaceIfNotEmpty(CModEntry::sizeToString(mod.getValue("localSize").toDouble()), lineTemplate.arg(tr("Size")));
+	if((mod.isAvailable() || mod.isUpdateable()) && mod.getValue("size").isValid())
+		result += replaceIfNotEmpty(CModEntry::sizeToString(mbToBytes(mod.getValue("size").toDouble())), lineTemplate.arg(tr("Download size")));
```
IvanSavenko, line 254: Perhaps rename "size" to "downloadSize" for clarity?

Also - validation schema needs update.

Nordsoft91:
@Laserlicht , that's interesting. I didn't thought about such automation. Not sure how tough ic could be, but at least from my side I can support size from both sources (mod and repository)

launcher/modManager/cmodlistview_moc.cpp
Nordsoft91: it's not stored or serialised but internal field. All internal computations are performed in bytes, including qt stuff. So the only conversion now from Mb to bytes is reading size from mod.json
Changing this value will require two additional conventions in few places (b -> mb and back)

IvanSavenko: In that case please name fields explicitly, e.g.
"localSizeBytes"
"downloadSizeMB"

launcher/modManager/cmodlistview_moc.cpp
Nordsoft91: Since it is existing but unused field, don't want to change it. But let me thing about it

> validation schema needs update.

will do

IvanSavenko: Please consider rename.
"size" is very ambiguous. And unlike local size it needs to be set by modders.

`"size" : 2.5` is very unclear. Is this size of mod? Size of download? Something else? Should it be defined in megabytes? Kilobytes? Bytes?

Laserlicht:
> @Laserlicht , that's interesting. I didn't thought about such automation. Not sure how tough ic could be, but at least from my side I can support size from both sources (mod and repository)

@Nordsoft91 Yes, I just have the concern that the sizes are never maintained. After an adjustment of the mod. And in my opinion, there should therefore be a technical solution.

launcher/modManager/cmodlistview_moc.cpp
Nordsoft91: having downloadSizeMB is even less clear. People may think that if there is downloadSizeMB than could be possible downloadSizeKB. DownloadSize is better, but still size is pretty clear and simple and it present since 0.99 (I removed column when moved mods to GitHub), no need to overcomplicate

IvanSavenko: > People may think that if there is downloadSizeMB than could be possible downloadSizeKB

This will be immediately detected by our validation. This is precisely why we have it.

> but still size is pretty clear and simple 

No it is no. At the very least it can also mean size of unpacked mod.

PR #2738, opened 2023-09-02
Improved autosaves

krs0:
First working version of this proposal: #2561

Code will be rearranged, some placement suggestions (for helper functions in special) would be welcomed.

There are some hardcoded things still.

- [x] I do not know how to find out if game is hotseat

- [x] I do not know how to change from startTimeIso8601 to a more readable format: 20190414T111005 vs 2019-04-14_11-10-05 (there is a class variable for startTimeIso8601 should I just create another one)?
replaced the startTimeIso8601 variable with a more human readable one.

- [x] map name templates should move to launcher as 4 separate settings.
settings created, will not expose them in GUI

If someone could take a look would be great.


client/CPlayerInterface.cpp
```diff
@@ -198,48 +199,200 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
...
+	variables["players"] = playerNames;
+	variables["campaignName"] = campaignName;
+ 
+	if(simpleSaves)
+	{
+		int autosaveCountLimit = settings["general"]["autosaveCountLimit"].Integer(); 
+		// save using a counter, and overwrite depending on autosave count limit
+		if(autosaveCountLimit > 0) 
+		{
+			auto saveNumber = extractNumberAfterHash(lastMapSaveName);
```
Laserlicht, line 329: Wouldn't it better and cleaner to save this information in the save file itself and get it from there? Then you can also make the counter as variable.

client/CPlayerInterface.cpp
```diff
@@ -198,48 +199,184 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
...
+				saveName = settings["general"]["singlePlayerRmgSaveName"].String(); break;
+			case Scenario:
+				saveName = settings["general"]["singlePlayerScenarioSaveName"].String(); break;
+			case Campaign :
+				saveName = settings["general"]["singlePlayerCampaignSaveName"].String(); break;
+			}
 		}
-	}
+
+		saveName = replaceVariables(saveName, variables);
```
IvanSavenko, line 376: You can use `boost::replace_all` instead of this method, e.g.
```cpp
for (auto const & entry : variables)
    boost::replace_all(saveName, "%" + entry.first + "%", entry.second);
```

include/vstd/DateUtils.h
```diff
@@ -4,10 +4,8 @@ VCMI_LIB_NAMESPACE_BEGIN
 
 namespace vstd
 {
-
 	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt);
-	DLL_LINKAGE std::string getDateTimeISO8601Basic(std::time_t dt);
-
+	DLL_LINKAGE std::string getDateTimeFormatted(std::time_t dt);
```
IvanSavenko, line 8: What's the difference between `getFormattedDateTime` and `getDateTimeFormatted`?
At least one of them might need better name and probably comment to explain formatting rules

client/CPlayerInterface.cpp
krs0: is boost not shunned everywhere in favor of std:: ?

client/CPlayerInterface.cpp
krs0: will look how to make that.

client/CPlayerInterface.cpp
IvanSavenko: If there is *identical* functionality in std - maybe.
But here boost provides way more convenient text formatting tools

client/CPlayerInterface.cpp
```diff
@@ -198,48 +199,184 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
...
+
+	// normalize map name
+	//int txtlen = TextOperations::getUnicodeCharactersCount(mapName); 
+	//TextOperations::trimRightUnicode(mapName, std::max(0, txtlen - 15)); // shorten to max 15 chars
+	std::string forbiddenChars("\\/<>:\"|?*\t\n\v\f\r"); 
+	std::replace_if(mapName.begin(), mapName.end(), [&](char c) { return std::string::npos != forbiddenChars.find(c); }, '_' ); // replace forbidden chars with _
+
+	// store variable values in map
+	std::unordered_map<std::string, std::string> variables;
+	variables["mapName"] = mapName;
```
IvanSavenko, line 316: You can add '%' in here and remove string manipulation during replacement:
```suggestion
	variables["%mapName%"] = mapName;

```

krs0:
Autosave limit makes sense only for simple saves or before battle saves.
When we save under the name: Autosave#1... 2,3,4,5, .... and then overwrite 1, then 2... etc.

In complex save naming, it makes no sense. Usually saves will be saved in named folders with turn info as filenames. Lime Broken Alliance/111, 112, 113..... so here limit makes no sense.

IvanSavenko:
If autosave limit does not makes sense in most cases - then do we need such option to begin with?

krs0:
Is someone does not care about complex saving scheme, games will be saved like Autosave#1, 2, 3... etc
This option will come in handy to automatically limit the numbers of autosaves.
I suspect this scheme will be quite handy for android.

IvanSavenko:
And what is used by default?

My point is:
I ran game (current develop) with default settings and I keep getting autosaves in `Autosave/Arrogance_2023XXXXX/1.vsgm1` format.

So I presume that by default we use such complex naming scheme that (effectively) ignores autosave limit. Correct?

include/vstd/DateUtils.h
krs0: I have changed the names to:
getDateTimeLocalized()
getDateTimeFormatted() -> name suggests its a special format, does not give info on the format :( (I would keep it like this until another function arrives)

client/CPlayerInterface.cpp
krs0: Done

client/CPlayerInterface.cpp
krs0: Done

client/CPlayerInterface.cpp
```diff
@@ -198,48 +199,200 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
...
+	std::string mapName = cb->getMapHeader()->name;
+	std::string mapDescription = cb->getMapHeader()->description;
+	std::string campaignName = "NA";
+	std::string timeStamp = cb->getStartInfo()->startTimeIso8601; // ToDo: krs - Get date time in more human format
+	std::string turn = std::to_string(cb->getDate(Date::MONTH))  
+		+ std::to_string(cb->getDate(Date::WEEK))
+		+ std::to_string(cb->getDate(Date::DAY_OF_WEEK)); 
+
+	// get random map info
+	if(mapName == "Random Map")
```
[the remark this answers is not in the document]
krs0: Good catch! ty. Replaced for now by mapGenOptions != nullptr; check.
Problem is the API related to maps campaigns and rmg maps is lacking a lot of functionality, and I do not trust myself to change those APIs.
I was expecting something straightforward like a bool property somwhere isRMGMap :(

client/CPlayerInterface.cpp
krs0: I think save file should not store this information. For now this will work.
Maybe in the future we get rid of it altogether? Or change how it works if Quick Load / Quick Save is implemented?

dydzio0614:
i have just one request: If this PR removes some autosave features, then I want to be able for autosave to work like in original H3 so I can have one AUTOSAVE file without folder

krs0:
@dydzio0614 ok boomer :)

simple autosave - autosaves in SAVES/Autosave#, or SAVES/Autosave_xxx depending on autosave limit

Autosave limit = how many autosaves to keep.

client/CPlayerInterface.cpp
krs0: done

krs0:
Ok works as intended: (maybe I find better place for 1 or 2 variables, but functionality is complete)
If someone can look into changes I did to Launcher (I am not so sure on that editor :( )
Out of scope for this PR:
```
	// ToDo : krs - gather all static info only once at map start / load
	// ToDo : krs - autosaves before each battle
	// ToDo : krs - autosave at beginning of game
	// ToDo : krs - quick autosaves on hotkey (autosave mode Using Counter could be used)
```
Example folder hierarchy from my test run:
[Saves.zip](https://github.com/vcmi/vcmi/files/12579515/Saves.zip)

These are launcher settings:
image
image

krs0:
Long long names are still an issue:
image

krs0:
Ok I have moved some code out of performAutosave and added a limit in GUI to filenames and folder names.

If someone could review, would be great.

I have 2 functions I do not know where to put:

// Generic function to replace forbidden characters with underscores
std::string replaceForbiddenChars(const std::string& input)

// Generic function to concatenate a vector of strings with a specified delimiter
std::string concatenateWithDelimiter(const std::vector<std::string>& strings, const std::string& delimiter)

krs0:
I intend to try! to clean up a little all the places were there is map info (or where these should be :( ). But not in this PR.

client/CPlayerInterface.cpp
```diff
@@ -211,47 +212,189 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
 	}
 }
 
-void CPlayerInterface::performAutosave()
+int extractNumberAfterHash(const std::string& input)
```
IvanSavenko, line 215: 1) Either make this function part of PlayerInterface class or mark it as 'static'
2) Perhaps move all autosave-related logic outside of player interface into a separate class? PlayerInterface already has such similar extracted components

lib/StartInfo.cpp
```diff
@@ -37,13 +37,47 @@ bool PlayerSettings::isControlledByHuman() const
...
+		if(playerInfos.at(player).isControlledByHuman())
+		{
+			playerNames.push_back(playerInfo.name);
+		}
+	}
+
+	return playerNames;
+}
+
+uint8_t StartInfo::getNoOfHumanPlayers() const
```
IvanSavenko, line 58:
```suggestion
uint8_t StartInfo::humanPlayerCount() const
```

krs0: uint8_t getHumanPlayersCount() const;

client/CPlayerInterface.cpp
krs0: extractNumberAfterHash has nothing specific to PlayerInterface so I moved it to TextOperations along with the other string operating functions

krs0:
This will be unmergeable once Launcher is updated. If no other suggestions could this be merged?
Improvements like below are independent, I will list them in the original issue:
```
	// ToDo : krs - gather all static info only once at map start / load
	// ToDo : krs - autosaves before each battle
	// ToDo : krs - autosave at beginning of game
	// ToDo : krs - quick autosaves on hotkey (autosave mode Using Counter could be used)
```

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -40,6 +40,14 @@ static const std::string cursorTypesList[] =
...
 };
 
+static const std::string autosaveModeList[] =
+{
+	"Off",
+	"Using Counter",
+	"Using Turn Info",
+	"Using Schema"
+};
+
```
IvanSavenko, line 50: This strings should be added directly into widget to make them translateable

lib/StartInfo.cpp
```diff
@@ -37,13 +37,47 @@ bool PlayerSettings::isControlledByHuman() const
...
+std::vector<std::string> StartInfo::getPlayerNames() const
+{
+	std::vector<std::string> playerNames;
+	for(PlayerColor player(0); player < PlayerColor::PLAYER_LIMIT; ++player)
+	{
+		if(!playerInfos.count(player))
+			continue;
+
+		auto playerInfo = playerInfos.at(player);
+		if(playerInfos.at(player).isControlledByHuman())
```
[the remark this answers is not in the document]
krs0: Done

client/CPlayerInterface.cpp
```diff
@@ -224,47 +225,135 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
...
+	// ToDo : krs - gather all static info only once at map start / load
+	// ToDo : krs - autosaves before each battle
+	// ToDo : krs - autosave at beginning of game
+	// ToDo : krs - quick autosaves on hotkey (autosave mode Using Counter could be used)
+
+	enum MapType {
+		Campaign,
+		Scenario,
+		RandomMap,
+		Undefined = 255
```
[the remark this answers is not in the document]
krs0: Done

launcher/settingsView/csettingsview_moc.cpp
krs0: using now custom data for controls:

```	ui->comboBoxAutosaveMode->setItemData(0, "Off", Qt::UserRole);
	ui->comboBoxAutosaveMode->setItemData(1, "Using Counter", Qt::UserRole);
	ui->comboBoxAutosaveMode->setItemData(2, "Using Turn Info", Qt::UserRole);
	ui->comboBoxAutosaveMode->setItemData(3, "Using Schema", Qt::UserRole);
```

client/CPlayerInterface.cpp
```diff
@@ -224,47 +225,135 @@ void CPlayerInterface::playerStartsTurn(PlayerColor player)
 	}
 }
 
-void CPlayerInterface::performAutosave()
+void CPlayerInterface::performAutosave()
```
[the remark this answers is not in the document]
krs0: Latest merging was time consuming. If really needed could this be done in separate PR?

lib/TextOperations.cpp
```diff
@@ -210,4 +248,28 @@ std::string TextOperations::escapeString(std::string input)
 	return input;
 }
 
+int TextOperations::extractNumberAfterHash(const std::string & input)
+{
+    size_t hashPos = input.find('#');
```
Nordsoft91, line 253: Use tabs instead of spaces

lib/TextOperations.cpp
```diff
@@ -210,4 +248,28 @@ std::string TextOperations::escapeString(std::string input)
 	return input;
 }
 
+int TextOperations::extractNumberAfterHash(const std::string & input)
+{
+    size_t hashPos = input.find('#');
+	if(hashPos != std::string::npos)
+		return -1;
```
Nordsoft91, line 255: So function extracts only non-negative number, not a number

lib/TextOperations.cpp
```diff
@@ -210,4 +248,28 @@ std::string TextOperations::escapeString(std::string input)
 	return input;
 }
 
+int TextOperations::extractNumberAfterHash(const std::string & input)
+{
+    size_t hashPos = input.find('#');
+	if(hashPos != std::string::npos)
+		return -1;
+
+	std::string numberAsString = input.substr(hashPos + 1); // Get the substring after '#'
```
Nordsoft91, line 257: Only one digit??

lib/TextOperations.cpp
```diff
@@ -199,6 +199,44 @@ size_t TextOperations::getUnicodeCharactersCount(const std::string & text)
...
+    }
+
+    if(!result.empty() && result.size() >= delimiter.size()) {
+        result.erase(result.end() - delimiter.size(), result.end());
+    }
+
+    return result;
+}
+
+std::string TextOperations::replaceForbiddenPathChars(const std::string & input)
```
Nordsoft91, line 226: Suggest to add delimiter as separate argument, instead of hardcoded underscore

lib/TextOperations.cpp
```diff
@@ -199,6 +199,44 @@ size_t TextOperations::getUnicodeCharactersCount(const std::string & text)
...
+{
+    std::string forbiddenChars = "\\/<>:\"|?*\t\n\v\f\r"; // space is not here since folder and file names can have spaces!
+    std::string result = input;
+
+    for (char& c : result)
+    {
+        if(forbiddenChars.find(c) != std::string::npos)
+            c = '_';
+    }
```
Nordsoft91, line 235:
```suggestion
for(char c : forbiddenChars)
std::replace(result.begin(), result.end(), c, delimiter);
```

lib/TextOperations.cpp
```diff
@@ -199,6 +199,44 @@ size_t TextOperations::getUnicodeCharactersCount(const std::string & text)
...
+    
+    for(const std::string& str : strings) {
+        result += str + delimiter;
+    }
+
+    if(!result.empty() && result.size() >= delimiter.size()) {
+        result.erase(result.end() - delimiter.size(), result.end());
+    }
+
+    return result;
```
Nordsoft91, line 223: Something strange: why not iterating over iterators and skip adding delimiter for the last one?

lib/TextOperations.cpp
```diff
@@ -199,6 +199,44 @@ size_t TextOperations::getUnicodeCharactersCount(const std::string & text)
 	return conv.from_bytes(text).size(); 
 }
 
+std::string TextOperations::shortenUnicodeString(const std::string & input, int maxLength)
```
Nordsoft91, line 202: More common name is truncate/truncateUnicodeString

lib/StartInfo.cpp
```diff
@@ -58,13 +58,47 @@ bool PlayerSettings::isControlledByHuman() const
...
+			playerNames.push_back(playerInfo.name);
+		}
+	}
+
+	return playerNames;
+}
+
+uint8_t StartInfo::getHumanPlayersCount() const
+{
+	uint8_t NoOfHumanPlayers = 0;
```
Nordsoft91, line 81: Local variables start with non-capital in vcmi code

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -540,3 +593,17 @@ void CSettingsView::on_spinBoxReservedArea_valueChanged(int arg1)
...
+	ui->comboBoxAutosaveMode->addItem("Off");
+	ui->comboBoxAutosaveMode->setItemData(0, "Off", Qt::UserRole);
+
+	ui->comboBoxAutosaveMode->addItem("Using Counter");
+	ui->comboBoxAutosaveMode->setItemData(1, "Using Counter", Qt::UserRole);
+
+	ui->comboBoxAutosaveMode->addItem("Using Turn Info");
+	ui->comboBoxAutosaveMode->setItemData(2, "Using Turn Info", Qt::UserRole);
+
+	ui->comboBoxAutosaveMode->addItem("Using Schema");
```
Nordsoft91, line 607: Using schema -> using pattern

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -515,23 +569,22 @@ void CSettingsView::on_comboBoxAlliedPlayerAI_currentTextChanged(const QString &
...
 }
 
-void CSettingsView::on_checkBoxAutoSavePrefix_stateChanged(int arg1)
+void CSettingsView::on_spinBoxAutosaveFrequency_valueChanged(int arg1)
 {
-    Settings node = settings.write["general"]["useSavePrefix"];
-    node->Bool() = arg1;
-    ui->lineEditAutoSavePrefix->setEnabled(arg1);
+    Settings node = settings.write["general"]["autosaveFrequency"];
+    node->Float() = arg1;
```
Nordsoft91, line 575: Why float not integer?

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -315,10 +345,34 @@ void CSettingsView::on_comboBoxShowIntro_currentIndexChanged(int index)
 	node->Bool() = index;
 }
 
-void CSettingsView::on_comboBoxAutoSave_currentIndexChanged(int index)
+void CSettingsView::on_comboBoxAutosaveMode_currentIndexChanged(int index)
 {
-	Settings node = settings.write["general"]["saveFrequency"];
-	node->Integer() = index;
+	QVariant itemData = ui->comboBoxAutosaveMode->itemData(index, Qt::UserRole);
+	std::string selectedEntry = itemData.toString().toStdString();
```
Nordsoft91, line 351: Please, don't store such strings in item data. Instead, create enum and store enum value in item data and them compare with enum instead of string comparison

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -40,6 +40,26 @@ static const std::string cursorTypesList[] =
...
+        if (itemData.toString().toStdString() == targetInternalName)
+        {
+            foundIndex = index;
+            break;
+        }
+    }
+
+	assert(foundIndex != -1);
+
+    return foundIndex;
```
Nordsoft91, line 60:
```suggestion
    for (int index = 0; index < comboBox->count(); ++index)
    {
        QVariant itemData = comboBox->itemData(index, Qt::UserRole);
        if (itemData.toString().toStdString() == targetInternalName)
            return index;
    }

	assert(0);

    return -1;
```

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -40,6 +40,26 @@ static const std::string cursorTypesList[] =
 	"software"
 };
 
+int findComboBoxItemByInternalName(QComboBox* comboBox, const std::string& targetInternalName)
```
Nordsoft91, line 43: It doesn't find item, but index

client/lobby/SelectionTab.cpp
```diff
@@ -443,13 +448,17 @@ void SelectionTab::filter(int size, bool selectFirst)
 			std::shared_ptr<ElementInfo> folder = std::make_shared<ElementInfo>();
 			folder->isFolder = true;
 			folder->folderName = folderName;
+			folder->displayName = TextOperations::shortenUnicodeString(folder->folderName, folderNameMaxDisplayLength);
```
Nordsoft91, line 451: Don't you want to add ... in the end of string if truncated?

Nordsoft91 (changes requested):
Formatting is broken - spaces/tabs are messed. We use tabs in vcmi code

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -540,3 +593,17 @@ void CSettingsView::on_spinBoxReservedArea_valueChanged(int arg1)
 	node->Float() = float(arg1) / 100; // percentage -> ratio
 }
 
+void CSettingsView::addItemsToControls()
+{
+	ui->comboBoxAutosaveMode->addItem("Off");
```
Nordsoft91, line 598: Enable translations for those strings using tr()
```suggestion
	ui->comboBoxAutosaveMode->addItem(tr("Off"));
```

lib/TextOperations.cpp
krs0: Done. I made copy past mistake, that was not caught because my VS automatic tools are off.
If not everybody sticks with code formatting guide, VS will change everything and then will let you wonder what are changes and what is whitespaces.

lib/TextOperations.cpp
krs0: /// extracts the number after a # character (useful in autosave filenames)

I think should be obvious without renaming the function. If any suggestion I can rename.

lib/TextOperations.cpp
krs0: Bug!
`std::string numberAsString = input.substr(hashPos + 1, input.length() - (hashPos + 1)); // Get the substring after '#' until the end of the string`

krs0: Bug!
`std::string numberAsString = input.substr(hashPos + 1, input.length() - (hashPos + 1)); // Get the substring after '#' until the end of the string`

lib/TextOperations.cpp
krs0: added delimiter as parameter

lib/TextOperations.cpp
krs0: Done

launcher/settingsView/csettingsview_moc.cpp
krs0: because other controls use the same. Copy paste -> changed to Integet

launcher/settingsView/csettingsview_moc.cpp
krs0: I have started with Enum but Ivan pointed out that there could be problems with translation. I did not understand the remark but in lack of feedback, a quick search on the net showed this itemData as best approach?

I see no problems with it, but if needed  I could revert to previous version that used enums.

launcher/settingsView/csettingsview_moc.cpp
krs0: renamed to findComboBoxItemsIndexByInternalName

client/lobby/SelectionTab.cpp
krs0: done

lib/TextOperations.cpp
krs0: Done

lib/StartInfo.cpp
krs0: renamed to humanPlayersCount

launcher/settingsView/csettingsview_moc.cpp
krs0: Renamed to Using Template. Seems more understandable?

lib/TextOperations.cpp
krs0: changed as suggested.

launcher/settingsView/csettingsview_moc.cpp
krs0: changed as suggested

launcher/settingsView/csettingsview_moc.cpp
krs0: Done.

launcher/settingsView/csettingsview_moc.h
```diff
@@ -68,11 +68,11 @@ private slots:
...
 	void on_comboBoxAlliedPlayerAI_currentTextChanged(const QString &arg1);
 
-    void on_checkBoxAutoSavePrefix_stateChanged(int arg1);
+    void on_spinBoxAutosaveFrequency_valueChanged(int arg1);
 
-    void on_spinBoxAutoSaveLimit_valueChanged(int arg1);
+    void on_spinBoxAutosaveCountLimit_valueChanged(int arg1);
 
-    void on_lineEditAutoSavePrefix_textEdited(const QString &arg1);
+    void on_checkBoxAutosaveBeforeBattles_stateChanged(int arg1);
```
Nordsoft91, line 75: some legacy formatting with spaces

krs0:
Again conflicts on merging :( Will not maintain it further.

krs0:
Requirements are not matching VCMI vision. Dropping this PR.

PR #2740, opened 2023-09-02
[fix] software cursor issue

Laserlicht:


client/render/Canvas.h
```diff
@@ -93,8 +93,8 @@ class Canvas
 	/// renders multiple lines of text with specified parameters
 	void drawText(const Point & position, const EFonts & font, const ColorRGBA & colorDest, ETextAlignment alignment, const std::vector<std::string> & text );
 
-	/// fills selected area with solid color
-	void drawColor(const Rect & target, const ColorRGBA & color);
+	/// fills selected area with solid color, allows replacing or overdrawing
```
Nordsoft91, line 96: What is the difference between replace and no replace?

Laserlicht: replace:
- drawing RGBA 128, 128, 128, 128
- on RGBA 0, 0, 0, 0
- result RGBA  128, 128, 128, 128

overdrawing:
- drawing RGBA 128, 128, 128, 128
- on RGBA 0, 0, 0, 0
- result RGBA  64, 64, 64, 0

Nordsoft91: probably, it can be solved with SDL blending. If this can solve problem, maybe it would be good to take blending mode as an argument instead of unclear bool. Does `Canvas::applyTransparency` do the same?

https://wiki.libsdl.org/SDL2/SDL_BlendMode

Laserlicht: Maybe possible. But I will have no time the next few days to look into it.

client/render/Canvas.h
```diff
@@ -93,8 +93,8 @@ class Canvas
 	/// renders multiple lines of text with specified parameters
 	void drawText(const Point & position, const EFonts & font, const ColorRGBA & colorDest, ETextAlignment alignment, const std::vector<std::string> & text );
 
-	/// fills selected area with solid color
-	void drawColor(const Rect & target, const ColorRGBA & color);
+	/// fills selected area with solid color, allows replacing or overdrawing
+	void drawColor(const Rect & target, const ColorRGBA & color, const bool replace = true);
```
IvanSavenko, line 97: 1) "blendColors" should be more clear than "replace"
2) Declaring method parameters as const (non-reference/pointer) is pointless since this parameter is copied into method anyway
```suggestion
	void drawColor(const Rect & target, const ColorRGBA & color, bool blendColors = false);
```

IvanSavenko, line 97: 1) "blendColors" should be more clear than "replace"
2) Declaring method parameters as const (non-reference/pointer) is pointless since this parameter is copied into method anyway
3) OR actually - might be better to create separate method, e.g. drawColorBlended
```suggestion
	void drawColor(const Rect & target, const ColorRGBA & color)
	void drawColorBlended(const Rect & target, const ColorRGBA & color)
```

client/render/Canvas.h
IvanSavenko: Correct. There is method called `SDL_SetSurfaceBlendMode` to do that.

However here separate method makes more sense - for simple "fill surface with color" just SDL_FillRect is enough.

Blended draw is noticeably slower - rather costly creation of new surface, filling it with semi-transparency, and then - blitting one surface onto another (and I think we use screen-sized surfaces here which are quite large)

IvanSavenko (approved):

PR #2757, opened 2023-09-04
Filesystem refactor - part 1

IvanSavenko:
Initial part of filesystem refactoring. Most of changes in this PR are simple search&replace, with no functionality changes - actual functionality changes will be done in separate PR(s).

Final goal is to completely isolate mod filesystems, to prevent conflicts when two unrelated mods have files with same name and to allow mods to use less complex directory layout without fear of conflicts.

List of changes in this PR:
- Renamed ResourceID class to ResourcePath, to better represent its functionality
- Added several classes that derive from ResourcePath and represent resource of a specific type - ImagePath, JsonPath, etc
- Use ResourcePath instead of std::string to store path to a resource in game entities
- Added explicit creation of ResourcePath for built-in / H3 files
- Added RenderHandler class on client that acts as factory for SDLImage and CAnimation classes - not related to FS changes, but  to avoid another large PR later.

Planned changes (separate PR's):
- resolve several cases where resource name for mod objects are runtime-generated and replace them with explicit name in config to allow tracking mod that added them.
- Allow mods to completely define directory structure, e.g. without config/sounds/sprites/data directories (useful for tiny mods)
- Track any other cases where mod adds resources, such as object icons and json animations.
- embed mod ID that requested resource into ResourceName to load resources from correct mod once needed by client
- when loading resource, access only filesystem that can be used by selected mod - only files from H3 and from mod itself (any additional exceptions?)


krs0:
Will relative paths to resources in config files be possible?

> EG: `animation : "..sprites/attack.def"`

IvanSavenko:
> Will we be able to specify in config files, relative paths to resources ?
> 
> > EG: `animation : "..sprites/attack.def"`

No. Not planned. All filepaths are always relative to the root directory of a mod.

krs0:
bigger mods have tons of config files sometimes 2, 3 deep nested. would be nice to not have to specify the complete path all the way to mod root.

PR #2761, opened 2023-09-05
(1.3.2) Fix Launcher freeze on startup

IvanSavenko:
I am tired of this... Quick fix, without research on how to make this better. Seems to be cause by our repository model where each mod is a separate repository for some reason...


IvanSavenko:
Huh, this fix is actually even better!

It also fixes that large delay after installing large mods! There is still some delay, but it is down to several seconds from several minutes.

PR #2769, opened 2023-09-05
(1.3.2) Implemented tolerance for touch events

### LINKED ISSUE #2439: [Touchscreen] Tolerance for touch
**Describe your proposal**
VCMI should have some tolerance with buttons. Since some controls on smartphones are quite small.

Idea would be that if the user does not hit any element (red), that then the closest button would be pressed (green). But only to a certain tolerance of distance (in pixels).

grafik

IvanSavenko:
When using touch input, events that happen within certain (20px) distance will now trigger nearby control input.

- Only active when touch input is used, for mouse tolerance is set to 0px
- Only clicking and popup display events are affected
- Fixes #2439

### DIFF
--- a/client/eventsSDL/InputSourceTouch.cpp
+++ b/client/eventsSDL/InputSourceTouch.cpp
@@ -121,9 +121,9 @@ void InputSourceTouch::handleEventFingerDown(const SDL_TouchFingerEvent & tfinge
 			if(tfinger.x > 0.5)
 			{
 				if (tfinger.y < 0.5)
-					GH.events().dispatchShowPopup(GH.getCursorPosition());
+					GH.events().dispatchShowPopup(GH.getCursorPosition(), params.touchToleranceDistance);
 				else
-					GH.events().dispatchMouseLeftButtonPressed(GH.getCursorPosition());
+					GH.events().dispatchMouseLeftButtonPressed(GH.getCursorPosition(), params.touchToleranceDistance);
 			}
 			break;
 		}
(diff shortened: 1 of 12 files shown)


client/eventsSDL/InputSourceTouch.h
```diff
@@ -78,6 +78,9 @@ struct TouchInputParameters
 	/// gesture will be qualified as pinch if distance between fingers is at least specified here
 	uint32_t pinchSensitivityThreshold = 10;
 
+	/// touch event will trigger clicking of elements up to X pixels away from actual touch position
+	uint32_t touchToleranceDistance = 20;
```
Laserlicht, line 82: Does it make sense to add this param in config (no GUI)?
Maybe one value is better for tablets and another value is better for small smarphones.

IvanSavenko: I guess I can add some param.

Don't see need for custom value for tablets though - current one seems to be quite good on small phones, and should work just fine even on larger screens

client/eventsSDL/InputSourceTouch.cpp
```diff
@@ -39,6 +39,7 @@ InputSourceTouch::InputSourceTouch()
 	params.relativeModeSpeedFactor = settings["general"]["relativePointerSpeedMultiplier"].Float();
 	params.longTouchTimeMilliseconds = settings["general"]["longTouchTimeMilliseconds"].Float();
 	params.hapticFeedbackEnabled = settings["general"]["hapticFeedback"].Bool();
+	params.touchToleranceDistance = settings["input"]["touchToleranceDistance"].Bool();
```
Laserlicht, line 42: Should be an integer?

PR #2780, opened 2023-09-08
Fix crash on obstacle creation (e.g. moat)

### LINKED ISSUE #2709: Crash in any seige battle
**Describe the bug**
I found this new fatal bug in latest daily branch. 
If in a seige war, if player is defender, it will be crash immediately.
If player is attacker, it will be stuck after arrow tower's action.

**Did it work earlier?**
It occurs after updating new build, I just wanna test backpack GUI.

**Version**
 - OS: Windows
 - Version: 1.4.0 updated this day

IvanSavenko:
Fixes #2709


wb180:
I encountered this crash myself and after I applied latest changes, my game still crashes in void GiveBonus::applyGs(CGameState *gs), because id for battle is also missing in void Moat::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const

IvanSavenko:
@wb180 noted

PR #2782, opened 2023-09-08
Map editor: events, rumors

Nordsoft91:


mapeditor/mapsettings/abstractsettings.h
```diff
@@ -0,0 +1,70 @@
...
+ * abstractsettings.h, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#ifndef ABSTRACTSETTINGS_H
+#define ABSTRACTSETTINGS_H
```
IvanSavenko, line 11: We use pragma once, not include guards

mapeditor/inspector/inspector.cpp
```diff
@@ -540,7 +540,7 @@ void Inspector::setProperty(CGArtifact * o, const QString & key, const QVariant
 	{
 		for(auto spell : VLC->spellh->objects)
 		{
-			if(spell->getJsonKey() == value.toString().toStdString())
+			if(spell->getNameTranslated() == value.toString().toStdString())
```
IvanSavenko, line 543: What if there are two spells with same name?

mapeditor/mapsettings/abstractsettings.h
Nordsoft91: it's automatically generated code by QT

dydzio0614:
will pandora boxes be supported as well?

Nordsoft91:
@dydzio0614 , pandora boxes are supported since 1.2 or even earlier. If it doesn't work, that's one more broken thing in 1.3

dydzio0614:
OK - if you decide to work on hero properties after events then feature-wise editor should be in decent state

Nordsoft91:
that's a plan

mapeditor/inspector/inspector.cpp
Nordsoft91: I'm working on PR with proper way of handling objects via id

dydzio0614:
just to make sure - both map and town timed events are supported in this PR?

Nordsoft91:
no, town events are not supported

PR #2788, opened 2023-09-09
Enable quick combat by default

Nordsoft91:


krs0:
Mixed feelings about this change.
This is a speed hack for players who know the game.
New players should always be presented with a fight until they feel comfortable to enable this option.
New players are a minority, but there still are.

Nordsoft91:
majority of users play with hdmod where it's enabled by default

PR #2795, opened 2023-09-10
Get next visible bound portal feature

david-shiko:
**Summary**
This PR introduces the GetNextVisibleBoundPortalFeature, which enhances the user's navigation experience in the adventure map. Now, right-clicking on any portal will sequentially move the screen to the next exit/entrance, providing a more intuitive way to understand portal connections.

**Future Work**
1. Implement long/short click to distinguish between original behavior (show caption on long press) and show next portal (on short click).
2. Remove focus/selection from hero after viewing next portal exit. This will help to easily return scsreen to the hero (rather than opening hero window).

Multiple exits may necessitate splitting the info window into additional windows.
The info window doesn't display the location on the map and the minimap.
Implementing this feature via the info window may require more time.

### DIFF
--- a/client/adventureMap/AdventureMapInterface.cpp
+++ b/client/adventureMap/AdventureMapInterface.cpp
@@ -45,7 +45,6 @@
 #include "../../lib/mapping/CMapDefines.h"
 #include "../../lib/pathfinder/CGPathNode.h"
 
-
 std::shared_ptr<AdventureMapInterface> adventureInt;
 
 AdventureMapInterface::AdventureMapInterface():
@@ -465,7 +464,6 @@ const CGObjectInstance* AdventureMapInterface::getActiveObject(const int3 &mapPo
 
 void AdventureMapInterface::onTileLeftClicked(const int3 &mapPos)
 {
-
 	if(!shortcuts->optionMapViewActive())
 		return;
 
@@ -726,8 +724,6 @@ void AdventureMapInterface::showMoveDetailsInStatusbar(const CGHeroInstance & he
 
 void AdventureMapInterface::onTileRightClicked(const int3 &mapPos)
 {
-	const CGObjectInstance * obj = getActiveObject(mapPos);
-
 	if(!shortcuts->optionMapViewActive())
 		return;
 
@@ -743,6 +739,7 @@ void AdventureMapInterface::onTileRightClicked(const int3 &mapPos)
 		return;
 	}
 
+	const CGObjectInstance * obj = getActiveObject(mapPos);
 	if(!obj)
 	{
 		// Bare or undiscovered terrain
@@ -760,8 +757,6 @@ void AdventureMapInterface::onTileRightClicked(const int3 &mapPos)
 			std::optional<const CGObjectInstance*> closestExit = portal->getNextVisibleExit(LOCPLINT->playerID);
 			if (closestExit.has_value()) {
 					widget->getMapView()->onCenteredObject(closestExit.value());
-					// Remove to ensure easy hero selection after screen moving (rather than opening hero window)
-					// LOCPLINT->localState->setSelection(nullptr);
 				}
 	}
 
--- a/client/adventureMap/AdventureMapInterface.cpp
+++ b/client/adventureMap/AdventureMapInterface.cpp
@@ -41,8 +41,10 @@
 #include "../../lib/spells/CSpellHandler.h"
 #include "../../lib/mapObjects/CGHeroInstance.h"
 #include "../../lib/mapObjects/CGTownInstance.h"
+#include "../../lib/mapObjects/MiscObjects.h"
 #include "../../lib/mapping/CMapDefines.h"
 #include "../../lib/pathfinder/CGPathNode.h"
+
 
 std::shared_ptr<AdventureMapInterface> adventureInt;
 
@@ -463,6 +465,7 @@ const CGObjectInstance* AdventureMapInterface::getActiveObject(const int3 &mapPo
 
 void AdventureMapInterface::onTileLeftClicked(const int3 &mapPos)
 {
+
 	if(!shortcuts->optionMapViewActive())
 		return;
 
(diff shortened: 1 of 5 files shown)


lib/mapObjects/MiscObjects.cpp
```diff
@@ -359,6 +359,54 @@ std::vector<ObjectInstanceID> CGTeleport::getAllExits(bool excludeCurrent) const
...
+			// isVisible method probably may be merged with getObj but I don't know LOCPLINT to use cb of current player
+			if(obj != nullptr && cb->isVisible(obj, playerColor)) {  // not visible objects returns nullptr (not sure is this check need)
+				return obj;
+			}
+		}
+		return std::nullopt;
+	}
+	catch (const std::exception& e) {
+	logGlobal->error("Exception caught: " + std::string(e.what()));
+	return std::nullopt;
```
Nordsoft91, line 401: adjust indentation

lib/mapObjects/MiscObjects.cpp
```diff
@@ -359,6 +359,54 @@ std::vector<ObjectInstanceID> CGTeleport::getAllExits(bool excludeCurrent) const
...
+			auto portalId = portals[((portalIndex +i) % portalsSize)];  // Get the next one (by index in portals) portal after the current.
+			const CGObjectInstance* obj = cb->getObj(portalId);
+			// isVisible method probably may be merged with getObj but I don't know LOCPLINT to use cb of current player
+			if(obj != nullptr && cb->isVisible(obj, playerColor)) {  // not visible objects returns nullptr (not sure is this check need)
+				return obj;
+			}
+		}
+		return std::nullopt;
+	}
+	catch (const std::exception& e) {
```
Nordsoft91, line 399: which functions throws here?

lib/mapping/MapReaderH3M.cpp
```diff
@@ -106,8 +106,6 @@ int32_t MapReaderH3M::readHeroPortrait()
 
 	if(result.getNum() == features.heroIdentifierInvalid)
 		return int32_t(-1);
-
-	assert(result.getNum() < features.heroesPortraitsCount);
```
Nordsoft91, line 110: why remove?

client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -751,6 +752,14 @@ void AdventureMapInterface::onTileRightClicked(const int3 &mapPos)
 		return;
 	}
 
+	const CGTeleport * portal = dynamic_cast<const CGTeleport*>(obj);
+	if(portal) {
+			std::optional<const CGObjectInstance*> closestExit = portal->getNextVisibleExit(LOCPLINT->playerID);
+			if (closestExit.has_value()) {
```
Nordsoft91, line 758: There are some coding guidelines https://wiki.vcmi.eu/Coding_guidelines.

Please, put braces from next line

Nordsoft91:
> Implement long/short click to distinguish between original behavior (show caption on long press) and show next portal (on short click).

do you plan to make it in scope of this PR?

lib/mapObjects/MiscObjects.cpp
```diff
@@ -359,6 +359,54 @@ std::vector<ObjectInstanceID> CGTeleport::getAllExits(bool excludeCurrent) const
...
+	logGlobal->error("Exception caught: " + std::string(e.what()));
+	return std::nullopt;
+	}
+	catch (...) {
```
IvanSavenko, line 403: Don't use "catch everything" unless there is a strong reason to

IvanSavenko (commented):
I think following approach should be better and more in-line with H3 UX:

- **Left-click**: no change - set path / move hero

- **Ctrl+Left-click**: 
- - if there is a discovered exit - switch map view to next exit (preferably with fade effect, similar to hero teleportation)
- - if there are no known exits - do nothing

- **Right-click**: 
- - If there is a discovered exit: show text "Monolith \n (Ctrl+click to switch to next exit)"
- - If there are no known exits: no change, show text "Monolith"

- **Hovering**: should display same text as right-click

For mobile - not available right now, will try to make radial menu similar to garrisons for adventure map by 1.4 release

lib/mapObjects/MiscObjects.cpp
```diff
@@ -359,6 +359,54 @@ std::vector<ObjectInstanceID> CGTeleport::getAllExits(bool excludeCurrent) const
...
+		std::vector<ObjectInstanceID> portals;
+
+		switch(type) {
+		case CGTeleport::ENTRANCE:
+		case CGTeleport::BOTH:
+			portals = getAllExits(false);
+			break;
+		case CGTeleport::EXIT:
+			portals = getAllEntrances(false);
+			break;
```
IvanSavenko, line 380: What if portal have (for example) 1 exits but multiple entrances? Will player be able to navigate through all entrances?

Nordsoft91 (changes requested):
Tested on map attached: 
[test_monoliths_2.vmap.zip](https://github.com/vcmi/vcmi/files/12578433/test_monoliths_2.vmap.zip)

1. Visit sign -> right click on blue entrance, observe jump on invisible exit
2. Both exits and entrances for one-way portals doesn't work properly
3. For bi-directional portals works well

david-shiko:
> > Implement long/short click to distinguish between original behavior (show caption on long press) and show next portal (on short click).
> 
> do you plan to make it in scope of this PR?

No, I think it should be a separate branch for handling the new input source, which I'll merge into this branch.

IvanSavenko:
> > > Implement long/short click to distinguish between original behavior (show caption on long press) and show next portal (on short click).
> > 
> > 
> > do you plan to make it in scope of this PR?
> 
> No, I think it should be a separate branch for handling the new input source, which I'll merge into this branch.

Please consider approach I described here: https://github.com/vcmi/vcmi/pull/2795#pullrequestreview-1619532001

No need for custom input source. Especially since mobiles already have both short and long tap  - which function similar to left & right click.

PR #2802, opened 2023-09-11
Voice for campaigns

### LINKED ISSUE #1992: Campaign prolog voice output
**Describe the bug**
Currently the voice output for the campaign prolog isn't played.

**To Reproduce**
Start Campaign

**Expected behavior**
In OH3: There is a voice output next to the video.

**Actual behavior**
No voice

Laserlicht:
fixes #1992

### DIFF
--- a/client/CMusicHandler.h
+++ b/client/CMusicHandler.h
@@ -60,6 +60,7 @@ private:
 
 	std::map<AudioPath, int> ambientChannels;
 	std::map<int, int> channelVolumes;
+	std::map<int, bool> channelPlaying;
 
 	void initCallback(int channel, const std::function<void()> & function);
 	void initCallback(int channel);
@@ -78,6 +79,7 @@ public:
 	int playSound(const AudioPath & sound, int repeats=0, bool cache=false);
 	int playSoundFromSet(std::vector<soundBase::soundID> &sound_vec);
 	void stopSound(int handler);
+	bool isSoundPlaying(int handler);
 
 	void setCallback(int channel, std::function<void()> function);
 	void soundFinishedCallback(int channel);
--- a/client/CMusicHandler.h
+++ b/client/CMusicHandler.h
@@ -60,7 +60,6 @@ private:
 
 	std::map<AudioPath, int> ambientChannels;
 	std::map<int, int> channelVolumes;
-	std::map<int, bool> channelPlaying;
 
 	void initCallback(int channel, const std::function<void()> & function);
 	void initCallback(int channel);
@@ -79,7 +78,6 @@ public:
 	int playSound(const AudioPath & sound, int repeats=0, bool cache=false);
 	int playSoundFromSet(std::vector<soundBase::soundID> &sound_vec);
 	void stopSound(int handler);
-	bool isSoundPlaying(int handler);
 
 	void setCallback(int channel, std::function<void()> function);
 	void soundFinishedCallback(int channel);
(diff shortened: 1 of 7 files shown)


lib/campaign/CampaignHandler.cpp
```diff
@@ -403,8 +404,13 @@ CampaignScenario CampaignHandler::readScenarioFromMemory( CBinaryReader & reader
 		ret.hasPrologEpilog = reader.readUInt8();
 		if(ret.hasPrologEpilog)
 		{
-			ret.prologVideo = CampaignHandler::prologVideoName(reader.readUInt8());
+			std::string originalCampaigns[] = { "DATA/GOOD1", "DATA/EVIL1", "DATA/GOOD2", "DATA/NEUTRAL1", "DATA/EVIL2", "DATA/GOOD3", "DATA/SECRET1", "DATA/AB", "DATA/BLOOD", "DATA/SLAYER", "DATA/FESTIVAL", "DATA/FIRE", "DATA/FOOL", "DATA/GEM", "DATA/GELU", "DATA/CRAG", "DATA/SANDRO", "DATA/YOG", "DATA/FINAL", "DATA/SECRET" };
+			bool isOriginalCampaign = std::find(std::begin(originalCampaigns), std::end(originalCampaigns), header.getFilename()) != std::end(originalCampaigns);
```
IvanSavenko, line 408: I think we can assume that all campaigns in "DATA" directory are H3 campaigns instead of such list

client/CMusicHandler.h
```diff
@@ -60,6 +60,7 @@ class CSoundHandler: public CAudioBase
 
 	std::map<AudioPath, int> ambientChannels;
 	std::map<int, int> channelVolumes;
+	std::map<int, bool> channelPlaying;
```
IvanSavenko, line 63: This approach won't work.
SDL can (and will) reuse sound channel ID's.

It works in your case (since no other sounds play) but if somebody attempts to reuse this method he might run into issues.

Instead you should set callback and close window when callback arrives.
See `BattleInterface::playIntroSoundAndUnlockInterface` for example

client/mainmenu/CPrologEpilogVideo.cpp
```diff
@@ -29,8 +29,7 @@ CPrologEpilogVideo::CPrologEpilogVideo(CampaignScenarioPrologEpilog _spe, std::f
 
 	CCS->videoh->open(spe.prologVideo);
 	CCS->musich->playMusic(spe.prologMusic, true, true);
-	// MPTODO: Custom campaign crashing on this?
-//	voiceSoundHandle = CCS->soundh->playSound(CCampaignHandler::prologVoiceName(spe.prologVideo));
+	voiceSoundHandle = CCS->soundh->playSound(spe.prologVoice);
```
IvanSavenko, line 32: Will this work with custom campaigns? No "failed to play sound" errors in console or something like that?

lib/campaign/CampaignHandler.cpp
Laserlicht: What will happen if a mod place a H3C in data instead of maps folder?

IvanSavenko (approved):

lib/campaign/CampaignHandler.cpp
```diff
@@ -168,6 +168,7 @@ CampaignScenario CampaignHandler::readScenarioFromJson(JsonNode & reader)
 		{
 			ret.prologVideo = VideoPath::fromJson(identifier["video"]);
 			ret.prologMusic = AudioPath::fromJson(identifier["music"]);
+			ret.prologVoice = AudioPath::fromJson(identifier["voice"]);
```
Nordsoft91, line 171: Please update documentation:

https://github.com/vcmi/vcmi/wiki/VCMI-Campaign-format#prologepilog

Laserlicht: Done :)

PR #2808, opened 2023-09-11
VCMI credits

Laserlicht:
what do you think about it?


krs0:
Nice to have something like this, but more people should be on that list than only developers.
Modders should be listed?
And Misiokles needs a special category :)

Laserlicht:
I think modders have already a place to get credit: In the launcher at each mod (author field).

Nordsoft91:
How about parsing https://github.com/vcmi/vcmi/blob/develop/AUTHORS ?

Maybe revise a structure of this file to json or something easy to parse. Then no need to do one more manual step prior release

IvanSavenko:
Taking use of AUTHORS file makes sense, but for that we'll need to:
- ensure that this file is installed on all platforms
- update that file - drop horribly outdated "what person does" part, check for missing contributors, add non-coding contributors such as QA (Povelitel, Misiokles, etc). Possibly remove emails?
- move this file or copy it - vcmi does not includes root directory in virtual filesystem, it needs to be in config directory.

Laserlicht:
New approach.

What do you think about it. Could also be displayed by launcher. In Client: Only usernames are shown (because of encoding).
Also the groups are displayed in Client.

IvanSavenko:
Personal opinion on new approach:
- Not sure about removal of AUTHORS file. Personally OK with that but others might disagree
- Probably drop email completely
- For those with both name and nickname we should probably display both - e.g. "Michał Urbańczyk aka Tow"
- Encoding is an issue, but I think can be ignored for now - we likely need to integrate unicode-capable font for 1.4 anyway, for MP chat.

Laserlicht:
@IvanSavenko 

> Probably drop email completely

Okay for me - but if we add authors to launcher too we can also add a "mailto://"-link.

> For those with both name and nickname we should probably display both - e.g. "Michał Urbańczyk aka Tow"

Possible but ingame it's to long (2-3 lines per name; nickname only one line) -> looks not good. But maybe in the launcher?

by003:
VCMI is amazing for many mods, although I am not a developer, I still contributed many mods for it. I reckon modding is one of the attraction aspects for VCMI as well.

IvanSavenko:
> Possible but ingame it's to long (2-3 lines per name; nickname only one line) -> looks not good. But maybe in the launcher?
See video above.

Yeah, watched video only after than message.

Option 1: keep as it is now
Option 2: add line breaks after every name and use format `(real name) \n aka (nickname) \n (empty line)`

Also:
- Give Tow and Tow Dragon special category as project founders?
- Perhaps sort names? Alphabetically probably to be safe. By contributions would also work, but rather arbitrary.

Nordsoft91:
> Perhaps sort names? Alphabetically probably to be safe. By contributions would also work, but rather arbitrary.

Usually it's not sorted alphabetically. Historical order (authors are added sequentially) is more common approach.

> Not sure about removal of AUTHORS file. Personally OK with that but others might disagree

Probably fine to remove file but let's keep short contribution summary which is in the file (without showing it in game)

IvanSavenko:
Might be good idea to add at least these people:
[rilian-la-te](https://github.com/rilian-la-te)
[henningkoehlernz](https://github.com/henningkoehlernz)
[SoundSSGood](https://github.com/SoundSSGood)
[vmarkovtsev](https://github.com/vmarkovtsev)
[krs0](https://github.com/krs0)

They have significant number of commits in stats, but looks to be missing from authors list.

> Probably fine to remove file but let's keep short contribution summary which is in the file (without showing it in game)

That list is horribly incomplete even for no longer active devs

IvanSavenko (approved):

Laserlicht:
> Might be good idea to add at least these people: [rilian-la-te](https://github.com/rilian-la-te) [henningkoehlernz](https://github.com/henningkoehlernz) [SoundSSGood](https://github.com/SoundSSGood) [vmarkovtsev](https://github.com/vmarkovtsev) [krs0](https://github.com/krs0)

done :)

by003:
Can you add me into test? I have already tested many kinds of bugs and they were already fixed.

Laserlicht:
> Can you add me into test? I have already tested many kinds of bugs and they were already fixed.

done :)

by003:
> > Can you add me into test? I have already tested many kinds of bugs and they were already fixed.
> 
> done :)

Thanks, my email address is benyan9110@gmail.com, same in my mods introduction

PR #2812, opened 2023-09-12
Remove dot in english.json for consistency

cewbdex:


cewbdex:
Found that setting help texts are inconsistent too, some have dots and some don't.

Nordsoft91 (approved):

cewbdex:
Added dots everywhere and I'm waiting for your feedback.

IvanSavenko:
> Added dots everywhere and I'm waiting for your feedback.

Looks OK either way, so no objections. H3 seems to be also inconsistent here, so don't see any approach as the "right one"

Issue #2581, opened 2023-08-14
For combined artifacts not all artifacts are counted


krs0:
**Describe the bug**
I have all components for a combined artifact. 1 is equipped on hero, 2 are in the backpack. GUI reports I have only 1 component. (Probably the one on the hero).

![image](https://github.com/vcmi/vcmi/assets/15194321/857a447d-bbfc-487f-96dc-389603e9bda2)

![image](https://github.com/vcmi/vcmi/assets/15194321/97f48674-e1d9-44e2-9d21-935f9e357b24)

**To Reproduce**
1. vcmiartifacts
2. Equiop Vampire's cowl on hero
3. Rightclick any component 
4. 1/3 components are shown as available.

**Expected behavior**
All components should be accounted for.

SoundSSGood:
Any ideas how to fix this?

krs0:
Should list all components that you have, equipped or in backpack.

SoundSSGood:
In this case, a situation may occur when the counter will show 3/3 but the artifact cannot be assembled. Since some items will be in the backpack. And this will be another bugreport

krs0:
If you have all required parts (on you or in backpack) the usual: do you want to combine should appear and artifact will be combined in backpack on first position.

Right now this detection is done separately for equipped and backpack, iirc.

SoundSSGood:
> asks too much

what do you mean?

krs0:
Every time you move a component of a combined artifact you are asked if you want to combine it or not.

Sometimes it is too much asking (at least my opinion)

PR #2814, opened 2023-09-12
Artifacts counter impovements

SoundSSGood:
Trying to fix https://github.com/vcmi/vcmi/issues/2581.
Now the counter in the pop-up window will also counts artifacts that are in the backpack.
A combined artifact can now be assembled even if some of its parts are in the backpack.
In this case, the assembled artifact will be placed at the beginning of the backpack. Or if the assembly is initiated from the backpack, then will be placed in the slot of the artifact that the player clicked on.

It was also noted that the number of annoying assembling requests has increased, because all artifacts are taken into account.
I have reduced the number of such requests. And now the request will be asked only if the artifact is transferred from hero to hero. This is experimental change. (The player can still assemble/disassemble artifacts by clicking the right button.)

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -264,7 +264,7 @@ bool ArtifactUtilsClient::askToAssemble(const CGHeroInstance * hero, const Artif
 	if(hero->tempOwner != LOCPLINT->playerID)
 		return false;
 
-	auto assemblyPossibilities = ArtifactUtils::assemblyPossibilities(hero, art->getTypeId(), ArtifactUtils::isSlotEquipment(slot));
+	auto assemblyPossibilities = ArtifactUtils::assemblyPossibilities(hero, art->getTypeId());
 	if(!assemblyPossibilities.empty())
 	{
 		auto askThread = new boost::thread([hero, art, slot, assemblyPossibilities]() -> void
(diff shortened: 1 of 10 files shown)


SoundSSGood:
> This is experimental change.

This means that feedback on this feature will not hurt.

krs0:
How about: for each combined artifact you are asked only once with popup if you want to assemble it. Once you have all components and open hero window.

Subsequent situations when you again, lose or gather all components should not trigger this popup.

SoundSSGood:
> How about: for each combined artifact you are asked only once

Do you mean once for each hero?

krs0:
Yes. Not per player stored but per hero.

krs0:
Also keep in mind that since there are a lot of new artifacts created, and slots remain the same, probably fused artifacts would be an improvement. (You fuse them and create a new artifact).
But the logic of asking for that will prob be the same as for combined artifacts.

SoundSSGood:
> You fuse them and create a new artifact

Without the possibility of disassembly?

To do this, it is needed to save the assembling request state for each combined artifact. And for each player separately. Because the hero can be owned by different players during one game. And then he can return to the previous overner.
I think this is quite possible to implement. And this will further reduce the number of requests compared to my change.

I'm not sure if such a change is necessary. In any case, I'd like to hear from VCMI members. And also feedback on my change. Perhaps it will be sufficient.

krs0:
Yes fused artifacts cannot be unfused. Think statue of legion.
With Ivan I've discussed something like town artifacts: again statue of legion would be best candidate.

krs0:
https://discord.com/channels/298106089885401090/1151219452407468082

krs0:
No feedback from Discord :(.

Cannot find you on slack or discord, so I will leave this here, since you are improving artifact related code.

Would be nice to have an option to be able to transfer scrolls to spell book.(maybe related to a skill like intelligence)

PR #2820, opened 2023-09-13
vcmi: make "noneOf" selector more restrictive

rilian-la-te:
It should do "or-not" instead of "and-not", because if we write 

```js
"noneOf" : [
     { "type" : "UNDEAD"},
     {"type" : "NON_LIVING"}
]
```

it should not affect both undead and non-living enemies, but not require enemy to be both undead and non-living same time.

Need to be merged after #2821, includes it.


krs0:
I would say it was correct with AND.
You are in fact changing to `notAnyOf`.

When you say "an item meets none of the criteria," you are describing a situation where the item does not satisfy any of the specified criteria or conditions you have set.

rilian-la-te:
Logical then to say noneOf and then And.

PR #2821, opened 2023-09-13
Fix MinGW CI 

rilian-la-te:
@IvanSavenko @Nordsoft91 need to be merged as soon as CI passes in both beta and develop.


Nordsoft91:
I'm useless reviewer, because don't have a way to verify

Issue #2806, opened 2023-09-11
App freezes on Android


Bencedek:
**Describe the bug**
When starting a map, after a short time the UI freezes (sound keeps playing). After switching to something else and switching back, only a black screen shows.

The client log's last two lines are the following:
[2023-Sep-11 18:54:58.811193] DEBUG global - SDL(category 0; priority 3) pixel format wanted SDL_PIXELFORMAT_RGBX8888 (2), got SDL_PIXELFORMAT_RGBX8888 (2)

Looks weird, since the values are the same. I guess it's some kind of rendering error

**Game logs**
[VCMI_Client_log.txt](https://github.com/vcmi/vcmi/files/12578815/VCMI_Client_log.txt)
[VCMI_Server_log.txt](https://github.com/vcmi/vcmi/files/12578816/VCMI_Server_log.txt)

**To Reproduce**
1. Start VCMI
2. New game, single scenario
3. I think the freeze happens after switching between two heroes, but I'm not exactly sure.

**Version**
- Phone: Sony Xperia 10 IV (XQ-CC54)
 - OS: Android 13, build 65.1.A.5.65 release-keys
 - Version: VCMI for Android 1.3.1

IvanSavenko:
How often does this happens? Always after a few minutes of gameplay?

Bencedek:
I've tried it two times, to see if the first one was a temporary error, but the program froze the second time too.
I'll try a few more times to hopefully narrow it down to an exact interaction or time spent ingame.

Bencedek:
What I've found after more tests is that the freeze comes when the hero touches an object that has no dialog (like collecting a resource). The freeze itself is quite inconsistent, but usually occurs when the hero needs to step 2-3 hexes toward the resource pile.
The map I was testing is Dungeon Keeper on 160% difficulty, it has a some resource piles near the beginning, but I could reproduce it on different maps and difficulties as well.

In the logs I see no obvious errors beside music loading ones (which should not freeze the game I guess).

IvanSavenko:
From what I see - your install is missing music. Was that deliberate? Yes, it should not lead to a freeze, but might be related somehow.
Android/data/is.xyz.vcmi/files/vcmi-data/Mp3 is empty.

- Can you try importing data once more, this time with mp3's?
- Can you try playing with maxed-out hero movement speed in settings? It should completely remove all animations from map, to check if freeze is related to them.
[...]

PR #2822, opened 2023-09-13
Attempt to fix freeze during map animations

IvanSavenko:
Not 100% sure whether this is what's causing #2806 but wasn't able to find any other reasons when checking the code


rilian-la-te:
@IvanSavenko please, merge my CI fix, because CI will be red before you do it.

PR #2823, opened 2023-09-13
Fix pathfinder tile cost

### LINKED ISSUE #2619: Incorrect movement calculation from dirt to snow
**Describe the bug**
Reported on discord by Der_Wollschwein

so, here's my current army and movement pobeha
image-6

and here's the info I get when I try to move one tile right
image-7

and here, when I try to move down
image-9

**Expected behavior**
Movement points should depend on starting terrain - dirt - 100 mp
There is also the weird 145 cost, that seems to indicate all movement points are consumed.

Still the bug could be GUI only bug.

dydzio0614:
Fix tile cost, movement points amount should be taken from source tile

Tested also with roads, embark/disembark, basic pathfinding skill. Fixes #2619

### DIFF
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -79,7 +79,7 @@ ui32 CGHeroInstance::getTileMovementCost(const TerrainTile & dest, const Terrain
 			!ti->hasBonusOfType(BonusType::NO_TERRAIN_PENALTY, from.terType->getIndex())) //no special movement bonus
 	{
 
-		ret = VLC->terrainTypeHandler->getById(dest.terType->getId())->moveCost;
+		ret = VLC->terrainTypeHandler->getById(from.terType->getId())->moveCost;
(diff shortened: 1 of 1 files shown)


krs0:
Can you please also check if native armies match the terrain? In the issue linked here there was also a native terrain bug hidden.

IvanSavenko:
@krs0 
what's wrong with native terrains? From what I see - game should use native terrain status of tile hero moves *from*.

dydzio0614:
I see that ALT button tile cost tends to not get refreshed until you move 1 tile, when you are flipping 1 non-native unit from town garrison to hero etc.

krs0:
@IvanSavenko in the provided Screen shot lateral movement should in no way cost 150 unless movement was calculated with snow penalty 150%. But he had all tower army... So maybe there is a hidden penalty bug there.
image

IvanSavenko:
@krs0 
I think this case will also be fixed by this PR

PR #2827, opened 2023-09-13
alt mode to select instead of changing path

Laserlicht:
Ctrl-click on hero will select him instead of changing path of current hero.

@IvanSavenko Where do we document this feature?


IvanSavenko (approved):

IvanSavenko:
> @IvanSavenko Where do we document this feature?

Probably in docs/player/Game_Mechanics.md

Quick army management is described in here, makes sense to place other shortucts in there.

Laserlicht:
Ok, can do it after docs are merged in develop. :)

Laserlicht:
@IvanSavenko Documented :)

PR #2828, opened 2023-09-13
Commander abilities requirements

dydzio0614:
in WoG only master skills are required for commanders, instead of grandmaster to learn special abilities

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -468,8 +468,8 @@ void CGameHandler::levelUpCommander(const CCommanderInstance * c)
 	int i = 100;
 	for (auto specialSkill : VLC->creh->skillRequirements)
 	{
-		if (c->secondarySkills.at(specialSkill.second.first) == ECommander::MAX_SKILL_LEVEL - 1
-			&&  c->secondarySkills.at(specialSkill.second.second) == ECommander::MAX_SKILL_LEVEL - 1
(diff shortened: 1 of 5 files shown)


Warzyw647:
Just to clarify: The requirement in OH3 WoG was "at least master level" not "exactly master level". That is, a Grandmaster skill was also a valid requirement.

server/CGameHandler.cpp
```diff
@@ -468,9 +468,9 @@ void CGameHandler::levelUpCommander(const CCommanderInstance * c)
 	int i = 100;
 	for (auto specialSkill : VLC->creh->skillRequirements)
 	{
-		if (c->secondarySkills.at(specialSkill.second.first) == ECommander::MAX_SKILL_LEVEL
-			&&  c->secondarySkills.at(specialSkill.second.second) == ECommander::MAX_SKILL_LEVEL
-			&&  !vstd::contains (c->specialSKills, i))
+		if (c->secondarySkills.at(specialSkill.second.first) == ECommander::MAX_SKILL_LEVEL - 1
+			&&  c->secondarySkills.at(specialSkill.second.second) == ECommander::MAX_SKILL_LEVEL - 1
+			&&  !vstd::contains (c->specialSkills, i))
```
IvanSavenko, line 473: What if one skill is already at grandmaster and another - at master level?

IvanSavenko:
Postponing to 1.3.3 / 1.4 to be safe

PR #2829, opened 2023-09-13
Easy choosing for a secondary skill (and others)

Laserlicht:
Easy choosing for a secondary skill in Level Up Dialog by double clicking it.

Suggestion from @IvanSavenko

### DIFF
--- a/client/widgets/CComponent.cpp
+++ b/client/widgets/CComponent.cpp
@@ -269,6 +269,12 @@ void CSelectableComponent::clickPressed(const Point & cursorPosition)
 		onSelect();
 }
 
+void CSelectableComponent::clickDouble(const Point & cursorPosition)
+{
+	if(onChoose)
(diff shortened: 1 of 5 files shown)


IvanSavenko:
Can we apply same logic to other selection windows? E.g. treasure chests?

Laserlicht:
> Can we apply same logic to other selection windows? E.g. treasure chests?

done :)

IvanSavenko (approved):
Thanks :)

Nordsoft91:
Very good!

Idea to go broader: what about enabling single touch for touchscreens only?

Laserlicht:
> Very good!

Thanks

> Idea to go broader: what about enabling single touch for touchscreens only?

You could do the same for the desktop and eliminate the okay button.

But I find it nicer this way. Is more like the original.

PR #2830, opened 2023-09-14
Save RNG state and load it on battle replay

wb180:
So a battle replay has same morale, damage, etc as a quick battle, otherwise sometimes you don't even have a chance to perform better than a quick combat. (HD mod replays work the same way).

### DIFF
--- a/server/battles/BattleProcessor.cpp
+++ b/server/battles/BattleProcessor.cpp
@@ -79,6 +79,7 @@ void BattleProcessor::restartBattlePrimary(const BattleID & battleID, const CArm
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


krs0:
HD mod also has infinite replays :).
I still thik this could be made available via a text only option.

Nordsoft91 (approved):

IvanSavenko:
This thing is largely broken:
1) This is not "battle random generator" but global RNG. Definitely will not work once simturns are ready and another player can act at the same time.
2) This is serialization of thread_local variable. And only of a thread local variable. Any calls to RNG that will be done from another thread will not use same RNG seed.

wb180:
Closing: Don't have time atm to work on it.

PR #2835, opened 2023-09-15
Updated release process docs

IvanSavenko:
Updated release process documentation based on how all recent releases were made by me

### DIFF
--- a/docs/maintainers/Release_Process.md
+++ b/docs/maintainers/Release_Process.md
@@ -1,10 +1,15 @@
 < [Documentation](../Readme.md) / Release Process
 
+# Versioning
+For releases VCMI uses version numbering in form "1.X.Y", where:
+- 'X' indicates major release. Different major versions are generally not compatible with each other. Save format is different, network protocol is different, mod format likely different.
+- 'Y' indicates hotfix release. Despite its name this is usually not urgent, but planned release. Different hotfixes for same major version are fully compatible with each other.
+
 # Branches
-Our branching strategy is very similar to GitFlow
-- `master` branch has release commits. One commit - one release. Each release commit should be tagged with version `maj.min.patch`
-- `beta` branch is for stabilization of ongoing release. Only safe changes should be targeted into this branch. Breaking changes (e.g. save format changes) are forbidden in beta.
-- `develop` branch is a main branch for ongoing development. Pull requests with new features should be targeted to this branch, `develop` version is one major release ahead of `beta`
+Our branching strategy is very similar to GitFlow:
+- `master` branch has release commits. One commit - one release. Each release commit should be tagged with version `1.X.Y` when corresponding version is released. State of master branch represents state of latest public release.
+- `beta` branch is for stabilization of ongoing release. Beta branch is created when new major release enters stabilization stage and is used for both major release itself as well as for subsequent hotfixes. Only changes that are safe, have minimal chance of regressions and improve player experience should be targeted into this branch. Breaking changes (e.g. save format changes) are forbidden in beta.
+- `develop` branch is a main branch for ongoing development. Pull requests with new features should be targeted to this branch, `develop` version is one major release ahead of `beta`.
 
 # Release process step-by-step
 
@@ -22,7 +27,7 @@ Should be done 2-4 weeks before planned release date. All major features should 
 - Bump version for Linux on `develop` branch
 - Bump version and build ID for Android on `develop` branch
 
-### Initial release setup (patch releases only)
+### Initial release setup (hotfix releases only)
 
 - Bump vcmi version in CMake on `beta` branch
 - Bump version for Linux on `beta` branch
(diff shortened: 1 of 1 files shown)


docs/maintainers/Release_Process.md
```diff
@@ -1,48 +1,64 @@
...
-- Create draft release page, specify `1.x.0` as tag for `master` after publishing
-- Prepare pull request with release update for web site https://github.com/vcmi/VCMI.eu
+Should be done 1 day before release. At this point beta branch is in full freeze.
+
+- Merge release preparation PR into `beta`
+- Merge `beta` into `master`. This will trigger CI pipeline that will generate release packages
+- Create draft release page, specify `1.x.y` as tag for `master` after publishing
+- Check that artifacts for all platforms have been build by CI `master` branch
+- Attach build artifacts for all platforms to release page
+- Manually extract Windows installer, remove `$PLUGINSDIR` directory and repackage data as .zip archive
```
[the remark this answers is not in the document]
IvanSavenko: Windows installer part? Some players have requested "Installer-free" / portable version in form of zip archive. And according to download stats it does have popularity (accounts for 25% of Windows downloads)

I don't know how to make Windows CI generate multiple artifacts from single run. If we can make it automatically during CI instead of manual actions - that would be great.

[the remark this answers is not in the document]
kambala-decapitator: Sorry, I mean the removal of plugins dir.

As for multiple artifacts, some cmake adjustments would be needed, as cpack generates only the installer. To get zip, there's a built-in cpack generator, just need to figure out how to pass generator type when cpack is called.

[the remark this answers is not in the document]
IvanSavenko: Ah, `$PLUGINSDIR` is internal directory that NSIS installer places into archive (but does not installs). It contains resources used by installer itself.

docs/maintainers/Release_Process.md
```diff
@@ -1,48 +1,64 @@
 < [Documentation](../Readme.md) / Release Process
 
-## Branches
+# Branches
 Our branching strategy is very similar to GitFlow
-* `master` branch has release commits. One commit - one release. Each release commit should be tagged with version `maj.min.patch`
-* `beta` branch is for stabilization of ongoing release. We don't merge new features into `beta` branch - only bug fixes are acceptable. Hotfixes for already released software should be delivered to this branch as well.
-* `develop` branch is a main branch for ongoing development. Pull requests with new features should be targeted to this branch, `develop` version is one step ahead of 'beta`
+- `master` branch has release commits. One commit - one release. Each release commit should be tagged with version `maj.min.patch`
+- `beta` branch is for stabilization of ongoing release. Only safe changes should be targeted into this branch. Breaking changes (e.g. save format changes) are forbidden in beta.
```
Nordsoft91, line 6: In the classic waterfall model, there is feature complete milestone, which means that no new features are supposed to be introduced. Then, sqa can focus on testing for those new features. Making new feature as last minute change is bad pattern because leaves less time for testing => poor quality in the end.
Definition of "safe" is quite unclear also but the main question is why do we have to rush and cannot deliver features into develop with longer stabilization time and more time to fix the regression issues.

docs/maintainers/Release_Process.md
```diff
@@ -1,48 +1,64 @@
...
+- `develop` branch is a main branch for ongoing development. Pull requests with new features should be targeted to this branch, `develop` version is one major release ahead of `beta`
```
Nordsoft91, line 7: Maybe I misunderstood this, does major mean increase major part of version?
We have "major.minor.patch"
Or it's different?

IvanSavenko: Correct. However we never call 1.2.0 -> 1.3.0 updates as "minor".
We have "major releases", e.g. 1.3.0 and "hotfixes", e.g. 1.3.1.

If can think of better phrasing - sure, but neither "step" nor "minor release" are clear.
Perhaps something like "on develop, version is set to 1.)X+1).0 compared to beta"?

Nordsoft91: True, as since bumping actual major version could look confusing (1->2) it maybe make sense to write down our versioning approach in the top, like
1(constant).major.patch unless we/community decide to change it

docs/maintainers/Release_Process.md
IvanSavenko: Is there any need to postpone small, finished features that have minimal chance of regressions AND are frequently asked by players till next release in 3 months?

In the same time, bugfixes for some tricky issues can be actually more dangerous than such minor features. 

Detection of misclick on touch input is technically a feature, but amount of "pain" it causes our players is quite high, so postponing it for 3 months is likely worse option since some controls are barely usable at the moment.

Czech translation is clearly a feature, but chance of regressions is basically zero, so downsides of including it into a hotfix are minimal.

Another problem is that line between "bugfix" and "new feature" can be blurry sometimes.
Should we count "give 500 XP on town capture" as feature or as bug? This can be viewed as bug since this is OH3 behavior. Or as feature since VCMI never had handling of such case.

docs/maintainers/Release_Process.md
IvanSavenko: Makes sense, e.g:
> VCMI releases use 1.X.Y format for versioning, where:
> X - major version
> Y - hotfix version

(but a bit more detailed)

docs/maintainers/Release_Process.md
Nordsoft91: Agree, that bug definition is not very strict since we don't have requirements, but I would say, there are two types of bugs can be (but not must) fixed during stabilization period:
 - something which worked earlier but doesn't work now (regression)
 - newly introduced feature doesn't work as expected (low-quality new functionality)
there are probably one more category, which is game mechanic issue (e.g. behavior is contradicting to requirements, h3 has a quite good game mechanic documentation). Example with 500XP fits this category

The rest are expceptions. And if there are ready features frequently asked by players, why not just simply shift release date?
The main point is do not consider "beta" as "develop" and do not hollow scope and dedicate efforts to improve quality of features we are delivering.

dydzio0614: impact on gameplay matters as well, for example there is artifact backpack in 1.3.2 that is basically isolated functionality that will not break rest of game, and is one of top wanted HD mod features (also it was almost done for 1.3.0)

IvanSavenko: > And if there are ready features frequently asked by players, why not just simply shift release date?

What do you mean? Release 1.4.0 with just these "frequently asked features" like artifact backpack? This is definitely not a major release like our 1.X.0 releases are.

And releasing 1.4.0 "as it" sooner means another save game format breakage, which is already a problem on some platforms.

> The main point is do not consider "beta" as "develop"

But we do not? A lot of risky features went in develop and not it beta.

I also don't remember any regressions introduced during stabilization.

Nordsoft91: no, I meant delay 1.3.0 if there is very desired feature pending, especially taking into consideration "also it was almost done for 1.3.0"

We are talking about documentation, so my suggestion here is only about elaboration for "safe" meaning.

PR #2836, opened 2023-09-15
Prefer symlink word

fulldecent:
For reasons, it is probably preferable that our documentation only says "symlink" here

### DIFF
--- a/docs/players/Installation_Android.md
+++ b/docs/players/Installation_Android.md
@@ -23,7 +23,7 @@ Installation is a two step process, at first you need to install game, then you 
 
 ### I imported game data files, but music in game is missing
 
-**Solution:** Try to run data import again or copy Mp3 folder from Heroes III manually to Android/data/is.xyz.vcmi/files/vcmi-data/Mp3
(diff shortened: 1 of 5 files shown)


IvanSavenko:
Why? Some people may prefer to copy data, e.g. if they extracted installer in some temporary directory.

fulldecent:
Of for sure that is how a lot of people are going to install it.

But when we put the word copy, we are encouraging people to copy copyrighted file (of course, for their own use). In some jurisdictions, this may make this project contributing to bad behavior.

IvanSavenko:
"Move or symlink" then?

dydzio0614:
I think windows OS dudes will often have no idea what symlink is

kambala-decapitator:
What if we simply use word "place"?

@dydzio0614 these are macos instructions

dydzio0614:
My bad, did not check - but the point still stands? MacOS has only simple stuff exposed by default, and users such as artists just need to be able to install application and have access to home folder for files

fulldecent:
Cool, thank you, "place or symlink"

IvanSavenko (approved):

krs0:
I understand the origin of this change, but now it is grammatically incorrect: Place... to?

PR #2841, opened 2023-09-15
Fix Market of Time missing Template

krs0:
Tested and works. (visit but nothing happens). Visitable and blocking seem to work ok. But I am no modder so please doublecheck.

### DIFF
--- a/config/objects/generic.json
+++ b/config/objects/generic.json
@@ -548,7 +548,7 @@
 				"templates" :
 				{
 					"green" : { "animation" : "avxdent.def", "visitableFrom" : [ "---", "+++", "+++" ], "mask" : [ "VV", "BA"], "allowedTerrains":["grass", "swamp", "dirt"] },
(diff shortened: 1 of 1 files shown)


nullkiller:
When I recently fixed market of time it worked. It should fetch template from HOMM3 data?

nullkiller:
Or maybe it depends on data itself and some homm3 data does not have market of time at all?

IvanSavenko:
@nullkiller 
Market of Time has no templates in H3 - this is unused object after all.
And objects without templates generate warning during mod validation.

And map with pre-placed Market of Time does not requires template - H3M has templates embedded into map format.

IvanSavenko (approved):
Looks OK.

Does not fixes original root cause of this change - crash on visiting unknown object, but if goal is to fix warning then probably OK.

krs0:
Yes goal is to fix warning and to shkw no surprise behaviour on visit.

IE should behave like intended on map except nothing happens on visit

Issue #2514, opened 2023-08-07
Double event message


jcialkowski:
**Describe the bug**
An event activation shows a message twice. The first one includes the message with a reward icon, while the second one only displays the message.

**To Reproduce**
1. Step onto event with a reward

**Expected behavior**
only on message should be displayed

**Actual behavior**
the message is shown twice

**Did it work earlier?**
I think so

**Screenshots**
![image](https://github.com/vcmi/vcmi/assets/33762148/577cb002-b486-4223-b564-dca583895958)
![image](https://github.com/vcmi/vcmi/assets/33762148/8012dd2f-aad2-4b84-8549-e2080a91cee5)

**Version**
 - OS: Linux
 - Version: 1.3

Nordsoft91:
@jcialkowski , can you please attach map?

jcialkowski:
sure, this is the map. just move to the right (there are a couple of events placed)
[test.zip](https://github.com/vcmi/vcmi/files/12507588/test.zip)

PR #2844, opened 2023-09-16
New rewardable interface for SeerHuts and pandoras

Nordsoft91:
- seer huts, quest gates, pandora boxes and events now acts as rewardable objects
 - backward compatibility with old maps preserved
 - new rewards widget is available now for all those objects in map editor

Снимок экрана 2023-09-16 в 05 35 48


lib/mapping/MapFormatH3M.cpp
```diff
@@ -1842,77 +1854,89 @@ void CMapLoaderH3M::readSeerHutQuest(CGSeerHut * hut, const int3 & position)
...
 	{
-		auto rewardType = static_cast<CGSeerHut::ERewardType>(reader->readUInt8());
-		hut->rewardType = rewardType;
+		auto rewardType = reader->readUInt8();
+		Rewardable::VisitInfo vinfo;
+		auto & reward = vinfo.reward;
 		switch(rewardType)
 		{
-			case CGSeerHut::EXPERIENCE:
+			case 0: //NOTHING
```
DjWarmonger, line 1862: Please keep the enum for sake of readability. You can move it to `CRewardableObject` or `MapFormaH3M` class.

lib/rewardable/Reward.cpp
```diff
@@ -94,4 +97,97 @@ void Rewardable::Reward::loadComponents(std::vector<Component> & comps,
...
+		a.syncSize(primary);
+		for(int i = 0; i < primary.size(); ++i)
+			a.serializeInt(i, primary[i]);
+	}
+	
+	{
+		auto a = handler.enterArray("secondary");
+		std::vector<std::pair<std::string, std::string>> fieldValue;
+		if(handler.saving)
+		{
```
IvanSavenko, line 125: Copy-pasted code from Limiter

Perhaps unify these massive blocks? Or create helper method to serialize creatures & secondary skills?

lib/mapObjects/CRewardableObject.cpp
```diff
@@ -107,6 +108,10 @@ void CRewardableObject::onHeroVisit(const CGHeroInstance *h) const
 					case Rewardable::SELECT_RANDOM: // give random
 						grantRewardWithMessage(*RandomGeneratorUtil::nextItem(rewards, cb->gameState()->getRandomGenerator()), true);
 						break;
+					case Rewardable::SELECT_ALL: // give all rewards
+						for(auto i : rewards)
+							grantRewardWithMessage(i, i == rewards.size() - 1);
```
IvanSavenko, line 113: Please check whether this mode works correctly if:
- multiple rewards levelup hero
- multiple rewards give creatures and hero has full army

lib/mapping/MapFormatH3M.cpp
```diff
@@ -1019,34 +1019,46 @@ CGObjectInstance * CMapLoaderH3M::readPandora(const int3 & mapPosition)
 void CMapLoaderH3M::readBoxContent(CGPandoraBox * object, const int3 & mapPosition)
 {
 	readMessageAndGuards(object->message, object, mapPosition);
+	Rewardable::VisitInfo vinfo;
+	auto & reward = vinfo.reward;
+	
+	reward.heroExperience = reader->readUInt32();
+	reward.manaDiff = reader->readInt32();
+	reward.bonuses.emplace_back(BonusDuration::ONE_BATTLE, BonusType::MORALE, BonusSource::OBJECT, reader->readUInt8(), object->id);
+	reward.bonuses.emplace_back(BonusDuration::ONE_BATTLE, BonusType::LUCK, BonusSource::OBJECT, reader->readUInt8(), object->id);
```
IvanSavenko, line 1028: `object->id` is not yet initialized at this point

lib/mapping/MapFormatH3M.cpp
IvanSavenko: Yes. Should be an enum in `MapFormatH3M`, similar to `EVictoryConditionType`

lib/mapping/MapFormatH3M.cpp
```diff
@@ -1842,77 +1854,89 @@ void CMapLoaderH3M::readSeerHutQuest(CGSeerHut * hut, const int3 & position)
...
 			}
-			case CGSeerHut::RESOURCES:
+			case 4: //LUCK_BONUS
 			{
-				hut->rID = reader->readUInt8();
-				hut->rVal = reader->readUInt32();
-
-				assert(hut->rID < features.resourcesCount);
-				assert((hut->rVal & 0x00ffffff) == hut->rVal);
+				reward.bonuses.emplace_back(BonusDuration::ONE_BATTLE, BonusType::LUCK, BonusSource::OBJECT, reader->readUInt8(), hut->id);
```
IvanSavenko, line 1884: Likely - `hut->id` is also uninitialized

lib/mapObjects/CGPandoraBox.cpp
```diff
@@ -25,297 +25,119 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
-void CGPandoraBox::initObj(CRandomGenerator & rand)
+void CGPandoraBox::init()
 {
-	blockVisit = (ID==Obj::PANDORAS_BOX); //block only if it's really pandora's box (events also derive from that class)
-	hasGuardians = stacks.size();
+	blockVisit = true;
+	configuration.selectMode = Rewardable::SELECT_ALL;
```
IvanSavenko, line 31: I assume only pandora's from VCMI map can have multiple rewards?

lib/mapObjects/CGPandoraBox.cpp
```diff
@@ -328,117 +150,119 @@ void CGPandoraBox::blockingDialogAnswered(const CGHeroInstance *hero, ui32 answe
...
-
-	resources.serializeJson(handler, "resources");
-
+	
+	if(!handler.saving)
 	{
-		bool haveSkills = false;
-
-		if(handler.saving)
+		//backward compatibility
```
IvanSavenko, line 172:
```suggestion
		//backward compatibility for VCMI maps that use old Pandora Box format
```

lib/mapObjects/CQuest.h
```diff
@@ -138,13 +138,9 @@ class DLL_LINKAGE IQuestObject
 	void afterAddToMapCommon(CMap * map) const;
 };
 
-class DLL_LINKAGE CGSeerHut : public CArmedInstance, public IQuestObject //army is used when giving reward
+class DLL_LINKAGE CGSeerHut : public CRewardableObject, public IQuestObject //army is used when giving reward
```
IvanSavenko, line 141: No longer relevant
```suggestion
class DLL_LINKAGE CGSeerHut : public CRewardableObject, public IQuestObject
```

lib/mapObjects/CQuest.cpp
```diff
@@ -922,163 +793,74 @@ void CGSeerHut::afterAddToMap(CMap* map)
...
-			handler.serializeInt(fullIdentifier, amount);
-		}
-	}
-	else
+	
+	if(!handler.saving)
 	{
-		rewardType = NOTHING;
-
+		//backward compatibility
```
IvanSavenko, line 802:
```suggestion
		//backward compatibility for VCMI maps that use old Seer Hut format
```

IvanSavenko:
> new rewards widget is available now for all those objects in map editor

Interesting - can we add "Import Json" and/or "Export Json" buttons in here?
To convert rewardable object into event and vice versa? However not sure how doable/needed such feature is.

dydzio0614:
Btw HotA allows repeatable quests and mixing one-time quest (triggers first) with repeatable quest in one seer hut - would be cool if such case could be supported without seer hut format remake in future

lib/mapObjects/CGPandoraBox.cpp
dydzio0614: no, just H3 different reward types are in separate tabs, you need to select different dropdown item to pick different reward type, but what you chose in another tab is preserved

IvanSavenko: My question is basically why pandoras use `Rewardable::SELECT_ALL` mode instead of using SELECT_FIRST like all other objects with single reward.

My guess is that in some cases Pandora's might have multiple rewards, but in H3 there is only one (but potentially quite large one). So can't think of anything else other than VCMI map

DjWarmonger: Pandora always could have multiple rewards. It has a single reward only on random maps.

IvanSavenko: You mean like mana and morale?
Those are parts of single "reward" as far as our rewardable object system is concerned.

DjWarmonger: Ahh, alright.

lib/mapping/MapFormatH3M.cpp
Nordsoft91: not sure how to handle it properly

lib/mapObjects/CGPandoraBox.cpp
Nordsoft91: It's more implementation-vise. For pandoras different rewards are shown as different messages. So I try to preserve this OH3 behavior.

IvanSavenko: > For pandoras different rewards are shown as different messages

Are you sure?

зображення

IvanSavenko: And also this may be related: https://github.com/vcmi/vcmi/issues/2514
(not pandora, but event, however they are largely identical)

lib/mapping/MapFormatH3M.cpp
IvanSavenko: ObjectInstanceID is present in method that calls this one (or someplace higher in the call stack)

lib/mapObjects/CGPandoraBox.cpp
Nordsoft91: Try to add creatures into pandora box

IvanSavenko: Looks like H3 has automated splitting of message by type:
- Experience, mana, luck, morale, primary skills, secondary skills go into single window but ONLY if total number of components is below 8
- Artifacts, creatures, resources always go into separate window.

Also, I think we already have message splitting on client - made by rilian-la-te as part of info box notifications and Scholar overflow fix.

So not sure if we need to split pandora reward on server side.

rilian-la-te:
Guards is also handled? Can you check it with info box enabled?

lib/mapObjects/CRewardableObject.cpp
```diff
@@ -27,44 +28,45 @@ static std::string visitedTxt(const bool visited)
...
+void CRewardableObject::selectRewardWthMessage(const CGHeroInstance * contextHero, const std::vector<ui32> & rewardIndeces, const MetaString & dialog) const
+{
+	BlockingDialog sd(configuration.canRefuse, rewardIndeces.size() > 1);
+	sd.player = contextHero->tempOwner;
+	sd.text = dialog;
 
-		if (rewards.size() > 1)
-			for (auto index : rewards)
-				sd.components.push_back(configuration.info.at(index).reward.getDisplayedComponent(h));
+	if (rewardIndeces.size() > 1)
```
IvanSavenko, line 58:
```suggestion
	if (rewardIndices.size() > 1)
```

Nordsoft91:
> Guards is also handled? Can you check it with info box enabled? 

Seems like, but I think we should prohibit info box for events, pandoras and seer huts

rilian-la-te:
@Nordsoft91 why? I think if interaction is not involved, info box is okay.

dydzio0614:
text events are supposed to stop your path etc.... and pandora rewards are messy if there are a lot of them

rilian-la-te:
@dydzio0614 I prefer info box if possible. So, I think it should be option at least (maybe sub option dependent from normal info box).

PR #2845, opened 2023-09-16
Support include path of more recent fuzzylite versions

Alexander-Wilms:


AI/Nullkiller/Engine/FuzzyEngines.h
```diff
@@ -8,7 +8,11 @@
 *
 */
 #pragma once
-#include <fl/Headers.h>
+#if __has_include(<fuzzylite/Headers.h>)
+#  include <fuzzylite/Headers.h>
+#else
+#  include <fl/Headers.h>
+#endif
```
kambala-decapitator, line 15: I'd move it to a separate header to avoid copy-paste

PR #2846, opened 2023-09-16
Fix 2689

nullkiller:
Disallow approaching guards from double guarded tiles


IvanSavenko (approved):

dydzio0614:
i wonder if order of fights is same as in H3 when you delibaretely go into double guardians tile

nullkiller:
Now only one fight happens and hero enters guarded tile without killing second monster

Issue #1912, opened 2023-04-08
Nullkiller AI seem to fall into eternal loop when visiting garrizons.


Macron1Robot:
**Describe the bug**
AI on its turn seems to fall into eternal loop when visiting garrizon objects.
Looks like it cant answer dialog of garrizon objects.

**Game logs**
System message: Server encountered a problem:
| Player "BLUE" has to answer queries before attempting any further actions.
| Top Query: "A query of type '20CGarrisonDialogQuery' and qid = 9258 affecting players BLUE and BLUE"

| Player "BLUE" has to answer queries before attempting any further actions.
| Top Query: "A query of type '20CGarrisonDialogQuery' and qid = 9258 affecting players BLUE and BLUE"

**To Reproduce**
Place garrizons on map (maybe they must have non-zero ai value).

Earlier VCAI acted same way, so maybe its a common problem.

IvanSavenko:
No blocked AI's after playing on "Warmongers" map for a week, so does not seems to be happening in 100% of cases.

nullkiller:
With queries it may go wrong when there are a few of them. For instance 20CGarrisonDialogQuery + level up query.

IvanSavenko:
Logs with stacktraces with this bug can be found here: https://github.com/vcmi/vcmi/pull/1936#issuecomment-1505140726
https://github.com/vcmi/vcmi/files/11211042/Archive.zip

JeffRockatansky:
Having the same issue.

Macron1Robot:
Today it happened again:

System message: Server encountered a problem:
| Player "GREEN" has to answer queries before attempting any further actions.
| Top Query: "A query of type '20CGarrisonDialogQuery' and qid = 2139 affecting players GREEN and GREEN"


| Player "GREEN" has to answer queries before attempting any further actions.
| Top Query: "A query of type '20CGarrisonDialogQuery' and qid = 2139 affecting players GREEN and GREEN"

nullkiller:
If somebody from core team suggest how to deal with it I will fix. Two queries and we answer the wrong one.

nullkiller:
Or we can try to change the protocol in a way that we always know which query is first and which is last. Some server generated timestamp?

IvanSavenko:
> If somebody from core team suggest how to deal with it I will fix. Two queries and we answer the wrong one.

How we end up in this situation to begin with? Any reliable steps to reproduce?

I see that one query is CGarrisonDialogQuery from logs. 
[...]

PR #2847, opened 2023-09-16
#1912 trap, exception on adding duplicating hero

nullkiller:
A trap for #1912. We expect it is duplicating heroes issue so I added a trap for it to trigger earlier and reveal who wants to add duplicating hero. The map with duplicating Orrins attached to #1912 crashes on start now as expected

### DIFF
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -37,6 +37,7 @@
 #include "../modding/ModScope.h"
 #include "../constants/StringConstants.h"
 #include "../battle/Unit.h"
+#include "CConfigHandler.h"
 
 VCMI_LIB_NAMESPACE_BEGIN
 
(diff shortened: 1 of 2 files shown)


Nordsoft91:
Make sense to add check for same heroes into map editor validation so mapmaker will be able to understand what's wrong with the map (if it's not there)

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1489,8 +1489,22 @@ std::string CGHeroInstance::getHeroTypeName() const
...
+	auto existingHero = std::find_if(map->objects.begin(), map->objects.end(), [&](const CGObjectInstance * o) ->bool
+		{
+			return (o->ID == Obj::HERO || o->ID == Obj::PRISON) && o->subID == subID && o != this;
+		});
+
+	if(existingHero != map->objects.end())
+	{
+		logGlobal->error("Hero is already on the map at %s", (*existingHero)->visitablePos().toString());
+
+		throw std::exception("Hero is already on the map");
```
IvanSavenko, line 1501: To fix CI build
```suggestion
		throw std::runtime_error("Hero is already on the map");
```

nullkiller:
Actually will check if this thing can crash map editor

nullkiller:
It crashes map editor. So need to find a way to disable this check there as it has own validation for this.

nullkiller:
Added session variable to tell if we are inside map editor. Now in editor it just issues warning and editor also has validation rule for duplicating heroes

IvanSavenko:
Then I guess let's merge and see if it will detect issues like #3099 ?

PR #2855, opened 2023-09-16
Added explicit "PlayerEndsTurn" netpack

IvanSavenko:
Another small piece towards simturns

- PlayerEndsTurn pack is now sent when player ends turn
- YourTurn pack has been renamed to PlayerStartsTurn for consistency
- PlayerStartsTurn will no longer replace list of acting players, but only appends to list
- PlayerEndsGame and PlayerEndsTurn will remove player from acting players list

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -172,11 +172,17 @@ void CPlayerInterface::initGameInterface(std::shared_ptr<Environment> ENV, std::
 	adventureInt.reset(new AdventureMapInterface());
 }
 
+void CPlayerInterface::playerEndsTurn(PlayerColor player)
+{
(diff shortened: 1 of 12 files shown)


Nordsoft91:
Timers part is missing on server side:
- timer shall stop counting down on playerendsturn event
- when timer is expired, server shall send playerendsturn

IvanSavenko:
No, it is not missing?

TurnTimerHandler::update 
-> calls CGameInfoCallback::isPlayerMakingTurn
-> timer will only count down if this method returned true

TurnTimerHandler::onPlayerMakingTurn
-> calls TurnOrderProcessor::onPlayerEndsTurn
-> calls TurnOrderProcessor::doEndPlayerTurn
-> sends PlayerEndsTurn netpacks

PR #2856, opened 2023-09-17
Hero overview

Laserlicht:
This one:

image

based on HotA 1.7: https://heroes3wog.net/wp-content/uploads/2023/07/2.png

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -30,6 +30,11 @@
 	"vcmi.capitalColors.6" : "Teal",
 	"vcmi.capitalColors.7" : "Pink",
 	
+	"vcmi.heroOverview.startingArmy" : "Starting Armys",
+	"vcmi.heroOverview.warMachine" : "War Machines",
+	"vcmi.heroOverview.secondarySkills" : "Secondary Skills",
+	"vcmi.heroOverview.spells" : "Spells",
+	
 	"vcmi.radialWheel.mergeSameUnit" : "Merge same creatures",
 	"vcmi.radialWheel.fillSingleUnit" : "Fill with single creatures",
 	"vcmi.radialWheel.splitSingleUnit" : "Split off single creature",
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -30,7 +30,7 @@
 	"vcmi.capitalColors.6" : "Teal",
 	"vcmi.capitalColors.7" : "Pink",
 	
-	"vcmi.heroOverview.startingArmy" : "Starting Armys",
+	"vcmi.heroOverview.startingArmy" : "Starting Units",
 	"vcmi.heroOverview.warMachine" : "War Machines",
 	"vcmi.heroOverview.secondarySkills" : "Secondary Skills",
 	"vcmi.heroOverview.spells" : "Spells",
--- a/client/VCMI_client.vcxproj.filters
+++ b/client/VCMI_client.vcxproj.filters
@@ -22,6 +22,9 @@
     <ClCompile Include="windows\CCreatureWindow.cpp">
       <Filter>windows</Filter>
     </ClCompile>
+    <ClCompile Include="windows\CHeroOverview.cpp">
+      <Filter>windows</Filter>
+    </ClCompile>
     <ClCompile Include="windows\CHeroWindow.cpp">
       <Filter>windows</Filter>
     </ClCompile>
@@ -170,6 +173,9 @@
       <Filter>windows</Filter>
     </ClInclude>
     <ClInclude Include="windows\CCreatureWindow.h">
+      <Filter>windows</Filter>
+    </ClInclude>
+    <ClInclude Include="windows\CHeroOverview.h">
       <Filter>windows</Filter>
     </ClInclude>
     <ClInclude Include="windows\CHeroWindow.h">
--- a/client/windows/CHeroOverview.cpp
+++ b/client/windows/CHeroOverview.cpp
@@ -22,6 +22,7 @@
 #include "../widgets/TextControls.h"
 #include "../widgets/MiscWidgets.h"
 
+#include "../../lib/GameSettings.h"
 #include "../../lib/CGeneralTextHandler.h"
 #include "../../lib/CCreatureHandler.h"
 #include "../../lib/CHeroHandler.h"
@@ -120,8 +121,10 @@ void CHeroOverview::genControls()
     labelArmyTitle = std::make_shared<CLabel>(r.x + borderOffset, r.y + borderOffset + 2, FONT_SMALL, ETextAlignment::TOPLEFT, Colors::YELLOW, CGI->generaltexth->translate("vcmi.heroOverview.startingArmy"));
 
     // army numbers
-    r = Rect(302, 3 * borderOffset + yOffset + 62, 292, 20);
+    r = Rect(302, 3 * borderOffset + yOffset + 62, 292, 32);
     backgroundRectangles.push_back(std::make_shared<TransparentFilledRectangle>(r.resize(1), rectangleColor, borderColor));
+
+    auto stacksCountChances = VLC->settings()->getVector(EGameSettings::HEROES_STARTING_STACKS_CHANCES);
 
     // army
     int space = (260 - 7 * 32) / 6;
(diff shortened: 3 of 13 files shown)


misiokles:
Great, but... it's again 1:1 copy of hd-mod feature. Maybe it will be better to rework this window? Make it more panoramical with panels in diffirent positions? For example - biography to the right side? Just loud thinking....

krs0:
Great QoL as always :)

Things to consider:
Hero can start with more than 3 stacks
More than 1 spell
Speciality could be also more than 1 thing
Multiple war machines...
Maybe he can start with multiple artifacts by default?

misiokles:
> Speciality could be also more than 1 thing
Yes, but there's only one icon for specialty (single or gruopped) in hero schema.

> Hero can start with more than 3 stacks
Only in custom maps.

Nordsoft91:
What about making this window based on configurable interface? As @misiokles said, it's 1-1 copy then we could move it to vcmi-extras

Laserlicht:
> What about making this window based on configurable interface? As @misiokles said, it's 1-1 copy then we could move it to vcmi-extras

What sense make it to be configurable in this case? In my opinion it shouldn't be in any mod. It's behind the enhanced ui flag. 

Also as far as I know there is no canvas support in configurable ui. Canvas drawing is heavily used here to avoid to deliver custom graphics.

Laserlicht:
> Great, but... it's again 1:1 copy of hd-mod feature. Maybe it will be better to rework this window? Make it more panoramical with panels in diffirent positions? For example - biography to the right side? Just loud thinking....

Yes. Because it has a very good shape. Feel free to create a good photoshopped draft. Then we can discuss. ;)

Laserlicht:
> Great QoL as always :)
> 
> Things to consider:
> Hero can start with more than 3 stacks
> More than 1 spell
> Speciality could be also more than 1 thing
> Heroes with 3 starting skils are already a thing.
> Multiple war machines...
> Maybe he can start with multiple artifacts by default?

Thanks. 

What of that is really possible from hero side?
I think we should describe the hero itself? Not in custom maps conditions.

For example even if the name is changed the old menu is used currently.

dydzio0614:
@misiokles nope - hero can start more with 3 stacks if you change default gameconfig.json settings and you configure more stacks for heroes

Laserlicht:
Update:
image

Needs a little bit fine tuning. But should okay for our requirements?

misiokles:
Just thinking it's possible to start with all three war machines.

Laserlicht:
Ready:
image

IvanSavenko:
- Why not try to use configurable UI? Yes, you use canvas extensively but I only see two operations: draw vertical line and draw box. You can make tiny class that derives from CIntObject and has showAll() method that draws box and 2nd class - for line.

- Not sure if we need explicit spell book icon - if hero has spell, he always have a spellbook. So probably redundant?

- Unlike hota, in vcmi/h3 starting army and war machine have appear probability. Perhaps show that probability in UI? Not sure if needed...

client/windows/CHeroOverview.h
```diff
@@ -0,0 +1,66 @@
...
+class CMultiLineLabel;
+class CFilledTexture;
+class CAnimImage;
+class CComponentBox;
+class CTextBox;
+
+class CHeroOverview : public CWindowObject
+{
+	const HeroTypeID & hero;
+	int heroIndex;
```
IvanSavenko, line 24: hero & heroIndex are the same thing in the end - number.

Prefer to use HeroTypeID instead of int where possible

client/VCMI_client.vcxproj.filters
```diff
@@ -22,6 +22,9 @@
     <ClCompile Include="windows\CCreatureWindow.cpp">
       <Filter>windows</Filter>
     </ClCompile>
+    <ClCompile Include="windows\CHeroOverview.cpp">
+      <Filter>windows</Filter>
+    </ClCompile>
```
IvanSavenko, line 27: I don't think that anybody uses or updates these project files.

Should probably be removed completely to be honest.

client/windows/CHeroOverview.cpp
```diff
@@ -0,0 +1,256 @@
...
+
+	pos = Rect(0, 0, 600, 485);
+
+	genBackground();
+    genControls();
+
+	center();
+}
+
+void CHeroOverview::genBackground()
```
IvanSavenko, line 45: Better option is to use showAll method and:
1) show background texture
2) draw this overlay
3) call CIntObject::showAll() to draw all controls

misiokles:
> Unlike hota, in vcmi/h3 starting army and war machine have appear probability. Perhaps show that probability in UI? Not sure if needed...

If plan is albo reading starting hero infos directly from custom maps - probability will be 100% always?

client/windows/CHeroOverview.h
Laserlicht: Yes, but `heroIndex` is a little bit shorter than` hero.getNum()`.

client/VCMI_client.vcxproj.filters
Laserlicht: I add it usually to every file I find. :)

Yes. But out of scope here.

client/windows/CHeroOverview.cpp
Laserlicht: I did not understand it exactly. :(

krs0:
For starting army if you can squeeze 1 more portrait all 7 slots could be covered.

6 slots for warmachines is too much... maybe replace 3 with a slider for artifacts?

Laserlicht:
> * Why not try to use configurable UI? Yes, you use canvas extensively but I only see two operations: draw vertical line and draw box. You can make tiny class that derives from CIntObject and has showAll() method that draws box and 2nd class - for line.

Yes, but the effort is too much for me. Has already cost enough time.
And I would first have to see how I can get transparent rectangles drawn.

> * Not sure if we need explicit spell book icon - if hero has spell, he always have a spellbook. So probably redundant?

Yes, redundant under certain circumstances. But also does not disturb!?
Can a hero have an empty spellbook with a mod?

> * Unlike hota, in vcmi/h3 starting army and war machine have appear probability. Perhaps show that probability in UI? Not sure if needed...

Maybe possible.

> For starting army if you can squeeze 1 more portrait all 7 slots could be covered.

Hmm... yes

> 6 slots for warmachines is too much... maybe replace 3 with a slider for artifacts?

6 slots is only for optics. Looks better if both have the same slot amount.

Slider is a bad idea. It's RMB info box...

IvanSavenko:
> And I would first have to see how I can get transparent rectangles drawn.

To be honest this might not be a bad idea even without configurable UI, if only to decrease amount of copy-paste you have now.

All you need is to make class like:
```cpp
class TransparentRectangle : public CIntObject
{
public:
    TransparentRectangle(Rect position, ColorRGBA color ...)
    void showAll(Canvas & to) override;
}
```

Laserlicht:
> ```c++
> class TransparentRectangle : public CIntObject
> {
> public:
>     TransparentRectangle(Rect position, ColorRGBA color ...)
>     void showAll(Canvas & to) override;
> }
> ```

Yes, and it works basically. But it draws with absolute coordinates. How can I change it that it that centering works?

IvanSavenko:
> Yes, and it works basically. But it draws with absolute coordinates. How can I change it that it that centering works?

Can you show your code? Something like this should work:
```cpp
TransparentRectangle::TransparentRectangle(Rect position)
{
    pos = position + pos.topLeft();
}
TransparentRectangle::showAll(Canvas & to)
{
    to.drawRect(pos, color);
}
```
- CIntObject has 'pos' member that should use screen coordinates
- CIntObject initializes its 'pos' to position of parent widget, if any
- CIntObject::center() automatically repositions all child UI elements
- Canvas received in CIntObject::showAll() uses screen coordinates, just like CIntObject::pos, so you should be able to use it directly

Laserlicht:
@IvanSavenko Thanks! I think it works. :)

Laserlicht:
> If plan is albo reading starting hero infos directly from custom maps - probability will be 100% always?

Would leave it without reading from map (I think HotA also don't do this). Reading from map (with all constellations) isn't that easy. But could be done similar to map preview in the future.

@IvanSavenko 
I've reworked it a lot to avoid double code.

Also added probabilities:
image

Mods/vcmi/config/vcmi/english.json
```diff
@@ -30,6 +30,11 @@
 	"vcmi.capitalColors.6" : "Teal",
 	"vcmi.capitalColors.7" : "Pink",
 	
+	"vcmi.heroOverview.startingArmy" : "Starting Armys",
```
IvanSavenko, line 33:
```suggestion
	"vcmi.heroOverview.startingArmy" : "Starting Army",
```

IvanSavenko: Or probably better
```suggestion
	"vcmi.heroOverview.startingArmy" : "Starting Units",
```

IvanSavenko (approved):
Looks OK. Would have been great to make this as configurable widget, but not critical.

PR #2858, opened 2023-09-17
 Use 64px and 32px as icon sizes in launcher sidebar

Alexander-Wilms:
image


IvanSavenko:
60 -> 64px change might be a good idea to reduce scaling artifacts but help & map editor are deliberately smaller than rest - Launcher is used on iOS with very limited vertical space. So less "important" icons were made smaller to avoid overflow

Alexander-Wilms:
Alright, I've changed he smaller ones to 32px:

image

IvanSavenko (approved):

Nordsoft91:
Actually, it introduces a little cosmetic defect on iOS
IMG_4722
IMG_4723

Alexander-Wilms:
Maybe set them all to 48px?

image

Or just leave it as is?

Nordsoft91:
Let's merge as is. There should be software solution for this

PR #2861, opened 2023-09-17
Do not show mod entries without name

### LINKED ISSUE #2849: Empty line in launcher for stardust-terrain mod
**Describe the bug**
image

**To Reproduce**
1. Install 1.3.2
2. See empty line in launcher

**Version**
 - OS: [e.g. Windows,
 - 1.3.2

Nordsoft91:
Fixes #2849

### DIFF
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -136,7 +136,7 @@ bool CModEntry::isInstalled() const
 
 bool CModEntry::isValid() const
 {
-	return (!localData.isEmpty() || !repository.isEmpty()) && !getValue("name").isNull();
+	return !localData.isEmpty() || (!repository.isEmpty() && !repository.contains("mod"));
 }
 
 bool CModEntry::isTranslation() const
--- a/launcher/modManager/cmodlist.cpp
+++ b/launcher/modManager/cmodlist.cpp
@@ -136,7 +136,7 @@ bool CModEntry::isInstalled() const
 
 bool CModEntry::isValid() const
 {
-	return !localData.isEmpty() || !repository.isEmpty();
+	return (!localData.isEmpty() || !repository.isEmpty()) && !getValue("name").isNull();
 }
 
 bool CModEntry::isTranslation() const


krs0:
How can the name be null? Is there a correction needed in the mod as well?

IvanSavenko:
Why are we hiding such mods? Can player (for example) delete such mod if he have one?
In which cases this might happen? Is it possible that this mod will be present in game but hidden in Launcher?

Nordsoft91:
> How can the name be null? Is there a correction needed in the mod as well?

We don't have mod names in repository, and only links

https://github.com/vcmi/vcmi-mods-repository/blob/develop/vcmi-1.4.json

Nordsoft91:
> Why are we hiding such mods? Can player (for example) delete such mod if he have one?
> 
> In which cases this might happen? Is it possible that this mod will be present in game but hidden in Launcher?

There are several possible ways to handle this, hiding is less disruptive, then showing something impossible to install. I will test for local mod, but technically only remote mod can be checked.

krs0:
> > How can the name be null? Is there a correction needed in the mod as well?
> 
> We don't have mod names in repository, only links
> 
> https://github.com/vcmi/vcmi-mods-repository/blob/develop/vcmi-1.4.json

ok but in those links the name property is set correctly to : "name" : "Stardust and Fields of Glory terrain". It is still null for some reason.

Nordsoft91:
Mod is incompatible- that's why it's not shown. But record is there, because mod is contained in repository

krs0:
I would list the mod anyway as disabled.
If I had a mod installed and upg VCMI and mod would disappear from list, I would be surprised :)

Nordsoft91:
It's not supposed to be. It's about downloadable mods

IvanSavenko:
> Mod is incompatible- that's why it's not shown. But record is there, because mod is contained in repository

Why we are hiding mod based on empty name instead of on compatibility status?

Nordsoft91:
> Why we are hiding mod based on empty name instead of on compatibility status?

what do you mean? We have empty empty name because of compatibility status

PR #2864, opened 2023-09-18
support double tap

Laserlicht:
Support double "click" on touch.

### DIFF
--- a/client/eventsSDL/InputSourceTouch.cpp
+++ b/client/eventsSDL/InputSourceTouch.cpp
@@ -182,7 +182,10 @@ void InputSourceTouch::handleEventFingerUp(const SDL_TouchFingerEvent & tfinger)
 		{
 			GH.input().setCursorPosition(convertTouchToMouse(tfinger));
 			if(tfinger.timestamp - lastLeftClickTimeTicks < params.doubleTouchTimeMilliseconds && (convertTouchToMouse(tfinger) - lastLeftClickPosition).length() < params.doubleTouchToleranceDistance)
-				GH.events().dispatchMouseDoubleClick(convertTouchToMouse(tfinger));
+			{
+				GH.events().dispatchMouseDoubleClick(convertTouchToMouse(tfinger), params.touchToleranceDistance);
+				GH.events().dispatchMouseLeftButtonReleased(convertTouchToMouse(tfinger), params.touchToleranceDistance);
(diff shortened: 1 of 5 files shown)


Laserlicht:
tested; works

client/eventsSDL/InputSourceTouch.cpp
```diff
@@ -181,8 +181,15 @@ void InputSourceTouch::handleEventFingerUp(const SDL_TouchFingerEvent & tfinger)
 		case TouchState::TAP_DOWN_SHORT:
 		{
 			GH.input().setCursorPosition(convertTouchToMouse(tfinger));
-			GH.events().dispatchMouseLeftButtonPressed(convertTouchToMouse(tfinger), params.touchToleranceDistance);
-			GH.events().dispatchMouseLeftButtonReleased(convertTouchToMouse(tfinger), params.touchToleranceDistance);
+			if(tfinger.timestamp - lastLeftClickTimeTicks < params.doubleTouchTimeMilliseconds && (convertTouchToMouse(tfinger) - lastLeftClickPosition).length() < params.doubleTouchToleranceDistance)
+				GH.events().dispatchMouseDoubleClick(convertTouchToMouse(tfinger));
```
IvanSavenko, line 185: We may need to add `dispatchMouseLeftButtonReleased` here.
Check double-clicking on any button that can be clicked multiple times, for example "next hero" button.

Also - add touch tolerance to double-click?

```suggestion
				GH.events().dispatchMouseDoubleClick(convertTouchToMouse(tfinger));
				GH.events().dispatchMouseLeftButtonReleased(convertTouchToMouse(tfinger), params.touchToleranceDistance);
```

Nordsoft91:
What about supporting triple touch? Can be used to emulate "space" on touchscreen

IvanSavenko (approved):

Laserlicht:
> What about supporting triple touch? Can be used to emulate "space" on touchscreen

Too much and unintuitive I think.

IvanSavenko:
> What about supporting triple touch? Can be used to emulate "space" on touchscreen

Currently I want to add radial menu similar to one we have for creature management to adventure map.
 
Object revisit is not the only missing action on adventure map, so sounds like a good idea. Wanted to do this for 1.3, but haven't find time for it.

PR #2868, opened 2023-09-18
Simultaneous turns

IvanSavenko:
Very early WIP.

Currently allows AI players to act in the same time as human - for ease of testing.
Basic actions seems to be working, but obviously unstable


Nordsoft91:
Tested a bit using localhost, one issue at this point: end turn button isn't blocked, so it looks like player doesn't know whether still controls turn or not

IvanSavenko:
Huh, that went better than expected.

Most of expected issues turned out to be way easier than expected. What's left is basically:
- Check MP with two clients.
- Add way to display & configure simturns in pregame UI.
- Review UI in game, make sure that UI displays what player expects.

And probably a lot of bugfixing. I fixed what I've found, but obviously can't cover all game functionality.

Nordsoft91:
I would suggest to proceed with merging in early state for broader testing on nightly builds. Even enabled by default seems ok for testing purposes

IvanSavenko:
> I would suggest to proceed with merging in early state for broader testing on nightly builds. Even enabled by default seems ok for testing purposes

For broad testing - yeah, makes sense. In that case:
- Checking basic playability with two clients
- Disabling AI simturns by default (will add some config/command to re-enable them, if only for testing)
- Add some basic way to configure simturns (without UI)
- Have simturns enabled by default for MP

After that - merge, and make UI for pregame in separate PR

Nordsoft91:
> Disabling AI simturns by default (will add some config/command to re-enable them, if only for testing)

Having them enabled would be beneficial for testing purposes. It's possible to make massive emulation tests using autoskip/spectator

client/CPlayerInterface.cpp
```diff
@@ -182,32 +189,14 @@ void CPlayerInterface::playerEndsTurn(PlayerColor player)
...
-
-	// close window from another player
-	if(auto w = GH.windows().topWindow<CInfoWindow>())
-		if(w->ID == QueryID::NONE && player != playerID)
-			w->close();
-	
-	// remove all dialogs that do not expect query answer
-	while (!GH.windows().topWindow<AdventureMapInterface>() && !GH.windows().topWindow<CInfoWindow>())
-		GH.windows().popWindows(1);
-
```
Nordsoft91, line 203: can you please confirm, that different scenarios with turn timer expiring (including messages, hot-seat turn passing, etc) work as expected and don't trigger assertions after your change?

server/processors/HeroPoolProcessor.cpp
```diff
@@ -174,6 +174,9 @@ bool HeroPoolProcessor::hireHero(const ObjectInstanceID & objectID, const HeroTy
 
 	if(mapObject->ID == Obj::TAVERN)
 	{
+		if (!gameHandler->isVisitActiveForPlayer(mapObject, player) && gameHandler->complain("Can't buy hero in tavern you are not visiting!"))
```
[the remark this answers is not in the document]
IvanSavenko: Will check other similar messages and change if they use different form

That being said, we should probably hide either all system messages or all AI messages in our release builds (as in, versions that we release to general public, not just release configuration)

server/CGameHandler.h
```diff
@@ -153,6 +153,9 @@ class CGameHandler : public IGameCallback, public Environment
 	
 	void castSpell(const spells::Caster * caster, SpellID spellID, const int3 &pos) override;
 
+	bool isVisitActiveForAny(const CGObjectInstance *obj);
+	bool isVisitActiveForPlayer(const CGObjectInstance *obj, PlayerColor player);
+	bool isVisitActiveForHero(const CGObjectInstance *obj, const CGHeroInstance *hero);
```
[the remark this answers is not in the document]
IvanSavenko: Actually I think only one method is needed here instead of 3 mostly copy-pasted methods. Will change.

client/CPlayerInterface.cpp
IvanSavenko: Will recheck, although this block was not removed but moved to other method(s)

nullkiller:
For AI simturns are quite complicated. NKAI often relays on fact that state will not change during analytics. This is why it often stores raw pointers. As soon as enemy hero can just disappear on the way it is no longer safe to store raw pointers to enemy heroes in danger hitmap or whatever. VCAI should be a bit safer. First of all because it does not see whole map)

IvanSavenko:
@nullkiller 
Yeah. Although this should not be major problem since simturns are designed to be active only before players make contact.

Another problem is gamestate lock.
AI relatively rarely releases lock on gamestate mutex. Meaning every time player makes action game has to wait for AI to release mutex in order to apply netpack for player.

So AI simturns will likely stay as debug-only option.

nullkiller:
Looks like AI should not lock gamestate at all. Ability to play AI game while there is no contact between AI and player can reduce  time needed to wait for AI turn to finish

IvanSavenko:
No, locking logic is correct. We can't apply gamestate changes when AI is making actions:

- when AI is making turn, gamestate mutex is locked in shared/read-only mode
- when AI is waiting for reply it unlocks gamestate mutex - to let gamepack applier to lock it in unique/write mode.

nullkiller:
> No, locking logic is correct. We can't apply gamestate changes when AI is making actions:
> 
> * when AI is making turn, gamestate mutex is locked in shared/read-only mode
> * when AI is waiting for reply it unlocks gamestate mutex - to let gamepack applier to lock it in unique/write mode.

If player can why AI can not? The most time it takes for AI to figure out what to do, pure analytics. Can two players move at the exact same time right now?

IvanSavenko:
> If player can why AI can not? The most time it takes for AI to figure out what to do, pure analytics. Can two players move at the exact same time right now?

Yes. However they can not make actions when netpack is being applied. Both human and AI have to wait during that moment. For human we achieve that by locking player interface mutex, for AI we achieve that by locking gamestate mutex.

For human player this mutex lock is unnoticeable - player interface mutex gets locked only for short period of time - to render frame or to process user input. So just few milliseconds.

But AI sometimes makes actions way longer - hundreds of milliseconds, blocking network thread from applying netpack with replay on human player actions.

config/widgets/optionsTab.json
```diff
@@ -73,16 +73,41 @@
...
+			"position": {"x": 55, "y": 537},
+			"size": 194,
+			"callback": "setSimturnDuration",
+			"itemsVisible": 1,
+			"itemsTotal": 28,
+			"selected": 0,
+			"style": "blue",
+			"scrollBounds": {"x": 0, "y": 0, "w": 194, "h": 32},
+			"panningStep": 20
+		},
```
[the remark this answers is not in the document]
IvanSavenko: 1) This is quick test UI, will be moved someplace else for 1.4
2) There is not enough free space in UI, either in H3 or in "extended lobby"
3) We should not gate features behind mods

IvanSavenko:
Probably merging this tomorrow to let people test this more extensively.
So far seems to be working fine, with no regressions. Missing pieces like better UI can be done later.

server/processors/TurnOrderProcessor.cpp
```diff
@@ -17,18 +17,134 @@
...
+				for (int x = 0; x < mapSize.x; ++x)
+					if (out.getNode({x,y,z})->reachable())
+						rightReachability[z][x][y] = true;
+	}
+
+	for (int z = 0; z < mapSize.z; ++z)
+		for (int y = 0; y < mapSize.y; ++y)
+			for (int x = 0; x < mapSize.x; ++x)
+				if (leftReachability[z][x][y] && rightReachability[z][x][y])
+					return true;
```
Nordsoft91, line 106: Can be only one pass thought whole map and single multi array `reachabilityMap`
//store array of paths per hero for left heroes and right heroes
std::vector<CPathsInfo> leftPaths, rightPaths;
for(leftHeroes, righHeroes)
{
leftPaths.emplaceBack(mapSize, hero);
...
}
for(z, y, x)
{
for(leftPaths)
reachabilityMap = true;
for(rightPaths)
if(reachabilityMap)
return true;
}
return false

But not actually critical because as I understand called only once per day

IvanSavenko: True. Wanted to keep code as simple as possible for now since we'll likely need to modify it further based on feedback.

Known issues with current approach:
- Paths per player may be computed multiple times - for each player x player pair. Each player results can be cached instead.
- It does not accounts for other ways player can act - town portal, fly, dimension doors.
- It does not accounts for potential zone guards defeat - you can defeat wandering monsters and attempt to attack player on the other side of the connection.
- And if we go into competitive pvp there are more arcane ways to extend movement range - for example prisons give you hero with max move points.

In most cases however this should not be an issue - this code will not be executed if simturns are inactive or if this check is disabled (not possible in current UI though). I also made adjustments to pathfinder to build only path in 1 turn range instead of entire map to reduce CPU usage. 

And even if player does manage to bypass this check and reaches another player during simturns - all interaction is blocked and should not break the game.

For now I am probably leaving it as it - will need to test & discuss this with MP players. Copying HD Mod approach is not something that can be done in vcmi with low difficulty, so will need feedback from players to decide in which direction to go from here.

PR #2869, opened 2023-09-18
disable audio on lost focus

Laserlicht:
HD feature.

For debugging it's helpful (if you have HD and VCMI opened the same time).

Do we need this also in regular codebase?

### DIFF
--- a/client/adventureMap/CInfoBar.cpp
+++ b/client/adventureMap/CInfoBar.cpp
@@ -25,6 +25,7 @@
 #include "../PlayerLocalState.h"
 #include "../gui/CGuiHandler.h"
 #include "../gui/WindowHandler.h"
+#include "../render/IScreenHandler.h"
 
 #include "../../CCallback.h"
 #include "../../lib/CConfigHandler.h"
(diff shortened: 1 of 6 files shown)


Nordsoft91:
Couple ideas:
 - play sound if chat message received even in the background
 - play sound if turn received even in the background

Laserlicht:
@Nordsoft91 Only disable music?

Nordsoft91:
I think it's more tricky. There are environment sounds you may want to not have but want to receive sound notifications when have turn or message.

Laserlicht:
Like this?

Nordsoft91:
Yes. Make sense for new day/new week sounds too

client/adventureMap/CInfoBar.cpp
```diff
@@ -228,14 +228,22 @@ CInfoBar::VisibleComponentInfo::VisibleComponentInfo(const std::vector<Component
...
+		handle = CCS->soundh->playSound(soundBase::newWeek);
 	else if(LOCPLINT->cb->getDate(Date::MONTH) != 1) // not first month
-		CCS->soundh->playSound(soundBase::newMonth);
+		handle = CCS->soundh->playSound(soundBase::newMonth);
 	else
-		CCS->soundh->playSound(soundBase::newDay);
+		handle = CCS->soundh->playSound(soundBase::newDay);
+
+	if(volume == 0)
+		CCS->soundh->setCallback(handle, [&]() { CCS->soundh->setVolume(0); });
```
IvanSavenko, line 246: If player refocuses game when sound is playing he'll end up without sound

IvanSavenko (approved):

PR #2874, opened 2023-09-19
Exchange window update

SoundSSGood:
The bulk movement of artifacts only from the backpack (shift+click) and equipped only(CTRL+click) has been implemented.
Changing icons on buttons is out of scope this PR.
Reworked CExchangeController. And moved to a separate file from the GUIClasses.cpp. Because this class is not GUI.
Added buttons for opening the backpack window to the hero exchange screen.
The buttons for moving the army and artifacts on the side of the allied hero will now be blocked. Just like it's done in HD mod.
Checks for the possibility to bulk move armies and artifacts on the server side have also been fixed.


dydzio0614:
hmmm i isolated crash for "transfer all creatures" button in VCMI 1.3 saved game but not sure if it's worth investigating by me now...

SoundSSGood:
> hmmm i isolated crash for "transfer all creatures" button in VCMI 1.3 saved game but not sure if it's worth investigating by me now...

I didn't fix the crash for "transfer all creatures". I think yes

Issue #2879, opened 2023-09-19
AI boost system proposal


misiokles:
I have and idea how to make AI more dangerous for human player.  But it needs additional parameters in the code/jsons.

For example:
This is Armorer skill code.
```
	"armorer" : {
		"index" : 23,
		"base" : {
			"effects" : {
				"main" : {
					"type" : "GENERAL_DAMAGE_REDUCTION",
					"subtype" : -1,
					"valueType" : "BASE_NUMBER"
				}
			}
		},
		"basic" : {
			"effects" : {
				"main" : { "val" : 5 }
			}
		},
		"advanced" : {
			"effects" : {
				"main" : { "val" : 10 }
			}
		},
		"expert" : {
			"effects" : {
				"main" : { "val" : 15 }
			}
		}
	},
```
Maybe those number could be increased only for enemy heroes? For example 8 for basic, 13 for advanced, 18 for expert? Even more - maybe those numbers should be tied with game difficulty (80%, 100%, 130%, 160%, 200%)?
It would need additional parameters in json, like:
	
"val" : 5,
"bonusOwner" : humanPlayer, enemyAI, alliedAI,
"gameDifficulty" : 80,100,130,160,200

The same system for some hero specialties.
With this AI heroes will be secretly cheating and more difficult to beat.

PR #2882, opened 2023-09-20
Add basic system to give more advantages for ai player

Nordsoft91:
Probably resolves https://github.com/vcmi/vcmi/issues/2879 and https://github.com/vcmi/vcmi/issues/1931

Bonuses themselves are for reference only, will be removed. It's up to modders to find proper balance


Nordsoft91:
> Can we add global/permanent bonuses in addition to per-battle bonuses?

There is similar functionality already, but I'll check

> Can we add bonuses separately for battles vs neutrals and for battles vs other players? To make it easier for AI to clear monsters without obvious cheats in player-vs-player battles.

Do we need bonuses vs other players? I implemented only vs neutrals exactly to hide actual cheating from players

dydzio0614:
Maybe it should be separate difficulty slider, somewhere near ingame difficulty picker or in launcher (off/1/2/3/4/5/based on primary difficulty).

Also not sure how many examples from #1931 can be implemented exclusively via bonus system, i would be OK with hardcoding the rest since most of them exist either in H4 enhancement mod or H5 AI so they feel like pretty standard AI code extensions to me.

IvanSavenko:
> There is similar functionality already, but I'll check

There is similar functionality indeed. But it can't be configured separately for AI's and human players.

> Do we need bonuses vs other players? I implemented only vs neutrals exactly to hide actual cheating from players

Ah, missed check for neutrals. Looks OK in this case.

krs0:
The final solution would be to decouple AI difficulty from starting resources. 2 separate options.
This way even 2 human players could play with same AI but different starting resources. Some kind of handicap system.

krs0:
No need to give attack or defense boost o AI in battle, if AI does well on map.

Lets boost AI on map and then this battle boost is not needed.

dydzio0614:
i think we need to be able to add as many arbitrary boost types as possible

DjWarmonger:
> No need to give attack or defense boost o AI in battle, if AI does well on map.
> 
> Lets boost AI on map and then this battle boost is not needed.

But adventure AI is way better than battle AI, and the reason AI player struggles is that it loses many units in the battle.

dydzio0614:
i agree with @DjWarmonger , though "better" is relative - adventure AI by definition has to be much more complicated due to wider range of responsibilities, and "weak battle AI" such as stupidAI has low chances of ruining game for average players, while bad adventure map AI will just make game non-challenging at all, due to the scope at which these AI's influence the gameplay

luk3ZPL:
"more hero movement per turn etc." it is a AI personality feature in vanilla H3.
Cheating should be optional in my opinion.

Nordsoft91:
let's start with what we have. Boost AI mod is already available. I believe it's challenging enough even on "rook" difficulty

PR #2889, opened 2023-09-21
Proper mod compatibility check logic

Nordsoft91:
Shows missing mods in human-readable format
Change logic of showing missing mods and auto-resolving process
Works for both saves and maps

### DIFF
--- a/lib/modding/CModHandler.cpp
+++ b/lib/modding/CModHandler.cpp
@@ -481,6 +481,10 @@ void CModHandler::trySetActiveMods(const std::vector<std::pair<TModID, CModInfo:
 		return nullptr;
 	};
 	
+	std::vector<TModID> newActiveMods, missingMods, excessiveMods;
+	ModIncompatibility::ModListWithVersion missingModsResult;
+	ModIncompatibility::ModList excessiveModsResult;
+	
 	for(const auto & m : activeMods)
 	{
 		if(searchVerificationInfo(m))
@@ -488,11 +492,11 @@ void CModHandler::trySetActiveMods(const std::vector<std::pair<TModID, CModInfo:
 
 		//TODO: support actual disabling of these mods
 		if(getModInfo(m).checkModGameplayAffecting())
+		{
+			excessiveMods.push_back(m);
 			allMods[m].setEnabled(false);
+		}
 	}
-
-	std::vector<TModID> newActiveMods, missingMods;
-	ModIncompatibility::ModList missingModsResult;
 	
 	for(const auto & infoPair : modList)
 	{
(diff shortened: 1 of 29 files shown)


dydzio0614:
My initial impressions based on skimming through code are good

lib/modding/CModHandler.cpp
```diff
@@ -471,47 +471,63 @@ void CModHandler::afterLoad(bool onlyEssential)
...
 	{
-		if (vstd::contains(saveActiveMods, m))
+		if(hasMod(m))
 			continue;
 
-		auto & modInfo = allMods.at(m);
-		if(modInfo.checkModGameplayAffecting())
-			missingMods.emplace_back(m, modInfo.version.toString());
+		if(getModInfo(m).checkModGameplayAffecting())
+			allMods[m].setEnabled(false);
```
IvanSavenko, line 490: This thing is broken. This is how it was back in 1.2 and we had constant crashes on changing mods

VCMI Client currently can not enable or disable mod in runtime. Client does not updates own VLC when loading save and does not reconfigures filesystem to allow access to mod files

lib/modding/CModHandler.cpp
```diff
@@ -471,47 +471,63 @@ void CModHandler::afterLoad(bool onlyEssential)
...
+		
+		if(modAffectsGameplay || remoteModInfo.impactsGameplay)
+			missingMods.push_back({remoteModInfo.name, remoteModInfo.version.toString()}); //incompatible mod impacts gameplay
 	}
+	
+	if(!missingMods.empty())
+		throw ModIncompatibility(std::move(missingMods));
+	
+	for(auto & m : newActiveMods)
+		allMods[m].setEnabled(true);
```
IvanSavenko, line 530: Same here

lib/modding/CModHandler.cpp
Nordsoft91: well, it's not broken, but unused. We have same in 1.3 too, so I prefer to keep setEnabled/setDisabled methods in order to have complete logic inside this method.

Reloading VLC based on the new mod list is quite big and risky change, so I'm not going to resolve it now.

IvanSavenko (dismissed):
Previous version of mod compatibility check requires *exactly* same set of gameplay-affecting mods.

This is somewhat inconvenient to players but greatly reduces crashes because of our broken VLC loading.

Example:
- Start game with only hota mod (all/almost all submods)
- Start as Cove, save game
- Enable Tides of War
- Load game started as Cove

Expected result:
a) game loads correctly
b) save loading is blocked due to incompatibility (this is also 1.3.2 behavior)

Actual results with this branch (depends on your luck / mod load order):
a) crash
b) completely garbled game objects, e.g. Cove creatures in army got replaced with Tides of War creatures

Nordsoft91:
@IvanSavenko , fixed

lib/modding/CModHandler.cpp
```diff
@@ -471,47 +471,93 @@ void CModHandler::afterLoad(bool onlyEssential)
...
 			continue;
 
-		auto & modInfo = allMods.at(m);
-		if(modInfo.checkModGameplayAffecting())
-			missingMods.emplace_back(m, modInfo.version.toString());
+		//TODO: support actual disabling of these mods
+		if(getModInfo(m).checkModGameplayAffecting())
+		{
+			excessiveMods.push_back(m);
+			allMods[m].setEnabled(false);
```
IvanSavenko, line 497: Correct me if I am wrong but this will mark mod as disabled without actually disabling it.

So
- start game with mod X -> mod marked as enabled
- attempt to start save without mod X -> mod will be marked as disabled, but save will fail to load
- attempt to start save with mod X -> save will fail to load because game thinks that mod is disabled

Actually it seems that such scenario is broken on both 1.3.2 and on your branch - server crash.

Nordsoft91: true, better do not touch mods at all for now

PR #2892, opened 2023-09-22
Highscore Menu & end video

### LINKED ISSUE #2395: Implement high scores
**Describe your proposal**
The output "Sorry, high scores menu is not implemented yet" says everything.

Absolutely low priority. But still worth an issue.

Laserlicht:
Implement Highscore menu and end videos (with highscore input) after game.

Difference to OH3:

- Highscore list is empty by default
- no animated creatures (looks better IMO and is a lot easier)
- RMB shows timestamp of highscore
- you can abort highscore input
- cheater gets 0 points :)
- internally all highscores will saved - if the rank is higher than 11 it will shown at the 11. place with the exact rank number

See here (sorry for german):
image
image
image
image

fixes #2395


Nordsoft91:
- Is it possible to add scrolling to have more records?
- is it possible to absorb data from  external source to show scores from the internet? So we could use it for tournaments
- as two things above, would be nice to have thing through configurable interface, so users can enable different tournament mods

Povelitelh3:
We need to think about cheating. Because there will be people who will add their army or resources to complete the map as quickly as possible and get into the table.
If I understand correctly that the table of records will be available via the Internet for everyone

dydzio0614:
I would sya... tournament scoring would require VCMI dedicated server that host games

IvanSavenko:
Yup. So local-only high scores for now, like in H3

Global high scores are an interesting idea, but needs A LOT of preparation work

Laserlicht:
The scope of this PR is creating a similar highscore than in OH3. To add this missing feature.

Only singleplayer and OH3 implementation for now.

Any additional things can added later. I also think, that we need for an online scoreboard a different screen layout. Maybe it could accessible from this menu ("Online scoreboard" button).

config/highscoreCreatures.json
```diff
@@ -0,0 +1,122 @@
+{
+    "creatures": [
+        { "min" : 1, "max" : 4, "creature": "core:imp" },
```
IvanSavenko, line 3: `core:` prefix is not necessary here

IvanSavenko:
Two more things to note:
- While scrollable window can be considered later, we can store all completed games in json (but only show best 10 in UI)
- We should also check for cheater flag and if set - don't add it to high scores. Just like H3 does. Flag already exists by the way.

Povelitelh3:
a cheat flag won't help, you can create a map with a bunch of cities, dig up the grail and win in one day, getting 500 points

IvanSavenko:
> a cheat flag won't help, you can create a map with a bunch of cities, dig up the grail and win in one day, getting 500 points

Or simply edit json
Or create cheat mod with OP heroes/creature.
Or use software for cheating

Goal is not to work as some sort anti-cheat system but to ignore testing runs. If I completed campaign using solely "win scenario" cheats to look on last map I don't want to see this game at top of high scores.

client/mainmenu/CHighScoreScreen.h
Nordsoft91, line 1: just in case - you have mix of spaces and tabs. In the codebase tabs are always used for indentation

Laserlicht: will fix later :)

Laserlicht:
Basically everything should work now. See PR description.

Some code has to restructured a little bit.

But it should work for scenario and campaign. Can you test, if you find any bugs?

client/CPlayerInterface.cpp
```diff
@@ -1685,16 +1686,39 @@ void CPlayerInterface::showShipyardDialogOrProblemPopup(const IShipyard *obj)
 
 void CPlayerInterface::requestReturningToMainMenu(bool won)
 {
+	HighScoreParameter param;
```
IvanSavenko, line 1689: Perhaps compute this structure on server and send it as part of netpack?

client/CPlayerInterface.cpp
```diff
@@ -1685,16 +1686,39 @@ void CPlayerInterface::showShipyardDialogOrProblemPopup(const IShipyard *obj)
 
 void CPlayerInterface::requestReturningToMainMenu(bool won)
 {
+	HighScoreParameter param;
+	param.difficulty = cb->getStartInfo()->difficulty;
+	param.day = cb->getDate();
+	param.townAmount = cb->howManyTowns();
+	param.usedCheat = cb->getPlayerState(*cb->getPlayerID())->enteredWinningCheatCode;
```
IvanSavenko, line 1693: This only checks for one cheat code. Check PlayerMessageProcessor::cheaters.
(only available on server side)

client/CPlayerInterface.cpp
```diff
@@ -1685,16 +1686,39 @@ void CPlayerInterface::showShipyardDialogOrProblemPopup(const IShipyard *obj)
...
 {
+	HighScoreParameter param;
+	param.difficulty = cb->getStartInfo()->difficulty;
+	param.day = cb->getDate();
+	param.townAmount = cb->howManyTowns();
+	param.usedCheat = cb->getPlayerState(*cb->getPlayerID())->enteredWinningCheatCode;
+	param.hasGrail = false;
+	for(const CGHeroInstance * h : cb->getHeroesInfo())
+		if(h->hasArt(ArtifactID::GRAIL))
+			param.hasGrail = true;
```
IvanSavenko, line 1697: Does not accounts for cases when player builds Grail in town.
Also, not 100% sure, but probably H3 actually check for successful digging action. Meaning, prebuilt grails in towns are not counted, grails acquired in other ways, e.g. picked on map, are not counted either.

client/CServerHandler.h
```diff
@@ -86,6 +89,8 @@ class CServerHandler : public IServerAPI, public LobbyInfo
 
 	std::vector<std::string> myNames;
 
+	std::shared_ptr<HighScoreCalculation> calc;
```
IvanSavenko, line 92: More descriptive name?
```suggestion
	std::shared_ptr<HighScoreCalculation> highScoreCalc;
```

client/CVideoHandler.h
```diff
@@ -31,7 +31,7 @@ class IVideoPlayer : boost::noncopyable
 class IMainVideoPlayer : public IVideoPlayer
 {
 public:
-	virtual void update(int x, int y, SDL_Surface *dst, bool forceRedraw, bool update = true){}
+	virtual void update(int x, int y, SDL_Surface *dst, bool forceRedraw, bool update = true, std::function<void()> restart = 0){}
```
IvanSavenko, line 34:
```suggestion
	virtual void update(int x, int y, SDL_Surface *dst, bool forceRedraw, bool update = true, std::function<void()> onVideoRestarted = nullptr){}

```

client/mainmenu/CHighScoreScreen.h
```diff
@@ -0,0 +1,108 @@
...
+{
+public:
+	int difficulty;
+	int day;
+	int townAmount;
+	bool usedCheat;
+	bool hasGrail;
+	bool allDefeated;
+	std::string campaign;
+	std::string land;
```
IvanSavenko, line 31:
```suggestion
	std::string campaignName;
	std::string scenarioName;
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+CHighScoreInput::CHighScoreInput(std::function<void(std::string text)> readyCB)
+	: CWindowObject(0), ready(readyCB)
+{
+	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
+
+	pos = center(Rect(0, 0, 232, 212));
+	updateShadow();
+
+	background = std::make_shared<CPicture>(ImagePath::builtin("HIGHNAME"));
```
IvanSavenko, line 383: For background you can provide it into CWindowObject constuctor:
`CWindowObject(0, ImagePath::builtin("HIGHNAME")),`
It will automaticallly load this image and update `pos` of your widget using background dimensions

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,401 @@
...
+			texts.push_back(std::make_shared<CLabel>(405, y + i * 50, FONT_MEDIUM, ETextAlignment::CENTER, color, tmp));
+			texts.push_back(std::make_shared<CLabel>(592, y + i * 50, FONT_MEDIUM, ETextAlignment::CENTER, color, std::to_string(curData["points"].Integer())));
+		}
+
+		if(curData["points"].Integer() > 0)
+			images.push_back(std::make_shared<CAnimImage>(AnimationPath::builtin("CPRSMALL"), (*CGI->creh)[HighScoreCalculation::getCreatureForPoints(curData["points"].Integer(), highscorepage == HighScorePage::CAMPAIGN)]->getIconIndex(), 0, 670, y - 15 + i * 50));
+	}
+}
+
+void CHighScoreScreen::buttonCampaginClick()
```
IvanSavenko, line 190:
```suggestion
void CHighScoreScreen::buttonCampaignClick()
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,401 @@
...
+{
+	highscorepage = HighScorePage::CAMPAIGN;
+	addHighScores();
+	addButtons();
+	redraw();
+}
+
+void CHighScoreScreen::buttonStandardClick()
```
IvanSavenko, line 198:
```suggestion
void CHighScoreScreen::buttonScenarioClick()
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,401 @@
...
+		{
+			// move following entries down
+			for (int j = 10; j + 1 >= i; j--)
+			{
+				JsonNode node = persistentStorage["highscore"][calc.isCampaign ? "campaign" : "scenario"][std::to_string(j - 1)];
+				Settings entry = persistentStorage.write["highscore"][calc.isCampaign ? "campaign" : "scenario"][std::to_string(j)];
+				entry->Struct() = node.Struct();
+			}
+
+			Settings entry = persistentStorage.write["highscore"][calc.isCampaign ? "campaign" : "scenario"][std::to_string(i)]["player"];
```
IvanSavenko, line 279: You can acquire necessary node once and use it for every access to reduce copy-pasted access:
```suggestion
			Settings entry = persistentStorage.write["highscore"][calc.isCampaign ? "campaign" : "scenario"][std::to_string(i)];
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,401 @@
...
+	video = won ? "HSANIM.SMK" : "LOSEGAME.SMK";
+}
+
+int CHighScoreInputScreen::addEntry(std::string text) {
+	for (int i = 0; i < 11; i++)
+	{
+		if(calc.calculate().cheater)
+			i = 10;
+
+		JsonNode node = persistentStorage["highscore"][calc.isCampaign ? "campaign" : "scenario"][std::to_string(i)];
```
IvanSavenko, line 267: Perhaps store results as vector?
```suggestion
		JsonNode node = persistentStorage["highscore"][calc.isCampaign ? "campaign" : "scenario"][i];
```

And then for moving entries down - sort using custom functor? E.g.

```cpp
auto sortFunctor = [](const JsonNode & left, const JsonNode & right)
{
    return left["points"].Integer() < right["points"].Integer();
};
boost::range::sort(node.Vector(), sortFunctor);
```

client/CPlayerInterface.cpp
Nordsoft91: Checking digging action is also not precise enough, because grail can be obtained by quest for example. There is combined grail mod, which allows to build grail from multiple artifacts.

IvanSavenko: AFAIK grail from quest will not count as "grail" for H3 score calculation

krs0:
Hidden text like these deserve a short comment.
`>translate("core.genrltxt.75")))`

client/mainmenu/CHighScoreScreen.cpp
Laserlicht: But how? I was not able to do this...
`entry["player"]->String() = text;` don't work.

client/mainmenu/CHighScoreScreen.cpp
Laserlicht: We cannot access  JSON with `[i]`, I already tried it. :(

client/mainmenu/CHighScoreScreen.cpp
IvanSavenko: Check `GeneralOptionsTab::setGameResolution`

client/mainmenu/CHighScoreScreen.cpp
IvanSavenko: Will recheck.
There is definitely Vector() member that gives you vector of JsonNode's.
Not sure about operator []

client/mainmenu/CHighScoreScreen.cpp
Laserlicht: ah... :)

client/CPlayerInterface.cpp
Laserlicht: Added grail in town

client/CPlayerInterface.cpp
Laserlicht: added a generic cheater flag in playerstate.

client/CPlayerInterface.cpp
Laserlicht: I don't want do get in to deep in server/client code. :)

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,399 @@
...
+#include "../../lib/CConfigHandler.h"
+#include "../../lib/CCreatureHandler.h"
+#include "../../lib/constants/EntityIdentifiers.h"
+#include "../../lib/TextOperations.h"
+
+#include "vstd/DateUtils.h"
+
+auto HighScoreCalculation::calculate()
+{
+	struct result
```
Nordsoft91, line 37:
```suggestion
	struct Result
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,399 @@
...
+	{
+		int basic;
+		int total;
+		int sumDays;
+		bool cheater;
```
Nordsoft91, line 42:
```suggestion
		int basic = 0;
		int total = 0;
		int sumDays = 0;
		bool cheater = false;
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,399 @@
...
+	};
+	
+	std::vector<int> scoresBasic;
+	std::vector<int> scoresTotal;
+	double sumBasic = 0;
+	double sumTotal = 0;
+	int sumDays = 0;
+	bool cheater = false;
```
Nordsoft91, line 50:
```suggestion
	std::vector<Result> scores;
	Result summary;
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,399 @@
...
+		if(param.difficulty == 3)
+			tmp *= 1.6;
+		if(param.difficulty == 4)
+			tmp *= 2.0;
+		scoresTotal.push_back(static_cast<int>(tmp));
+		sumTotal += tmp;
+		sumDays += param.day;
+		if(param.usedCheat)
+			cheater = true;
+	}
```
Nordsoft91, line 71:
```suggestion
	const std::array<double, 5> difficultyMultipliers{0.8, 1.0, 1.3, 1.6, 2.0}; 
	for(auto & param : parameters)
	{
		double tmp = 200 - (param.day + 10) / (param.townAmount + 5) + (param.allDefeated ? 25 : 0) + (param.hasGrail ? 25 : 0);
		scores.emplace_back(tmp, tmp * difficultyMultipliers.at(param.difficulty), param.day, param.usedCheat);
		summary.basic += scores.back().basic * 5.0 / parameters.size();
		summary.total+= scores.back().total * 5.0 / parameters.size();
		summary.sumDays += param.day;
		summary.cheater |= param.usedCheat;
	}
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+	if(scoresBasic.size() == 1)
+		return result { scoresBasic[0], scoresTotal[0], sumDays , cheater};
+
+	return result { static_cast<int>((sumBasic / parameters.size()) * 5.0), static_cast<int>((sumTotal / parameters.size()) * 5.0), sumDays, cheater };
```
Nordsoft91, line 76:
```suggestion
	if(scores.size() == 1)
		return scores.front();

	return summary;
```

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,399 @@
...
+	pos = center(Rect(0, 0, 800, 600));
+	updateShadow();
+
+	addHighScores();
+	addButtons();
+}
+
+void CHighScoreScreen::showPopupWindow(const Point & cursorPosition)
+{
+	for (int i = 0; i < 11; i++)
```
Nordsoft91, line 107: please move 11 to separate constant and bind relevant values to it (e.g 10)

client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -0,0 +1,399 @@
...
+
+	background = std::make_shared<TransparentFilledRectangle>(Rect(0, 0, pos.w, pos.h), Colors::BLACK);
+
+	if(won)
+	{
+		int border = 100;
+		int textareaW = ((pos.w - 2 * border) / 4);
+		std::vector<std::string> t = { "438", "439", "440", "441", "676" }; // time, score, difficulty, final score, rank
+		for (int i = 0; i < 5; i++)
+			texts.push_back(std::make_shared<CMultiLineLabel>(Rect(textareaW * i + border - (textareaW / 2), 450, textareaW, 100), FONT_HIGH_SCORE, ETextAlignment::TOPCENTER, Colors::WHITE, CGI->generaltexth->translate("core.genrltxt." + t[i])));
```
Nordsoft91, line 250: why MultiLineLabel?

Nordsoft91 (commented):
Few suggestions on how to improve code. Still believe that should be based on configurable widget => less code + more flexibility

client/mainmenu/CHighScoreScreen.cpp
Nordsoft91: even better: get rid of scores array and have only `Result firstResult`. Then in condition below
```c++
if(parameters.size() == 1)
    return firstResult;
return summary;
```

client/mainmenu/CHighScoreScreen.cpp
Laserlicht: Original strings has a line break in it which has to be supported.

krs0:
Will this work for tutorial?

Laserlicht:
> Will this work for tutorial?

Yes, as in OH3.

lib/CPlayerState.h
```diff
@@ -39,6 +39,7 @@ struct DLL_LINKAGE PlayerState : public CBonusSystemNode, public Player
 	std::vector<QuestInfo> quests; //store info about all received quests
 	std::vector<Bonus> battleBonuses; //additional bonuses to be added during battle with neutrals
 
+	bool cheated;
```
IvanSavenko, line 42: Perhaps remove `PlayerMessageProcessor::cheaters` field since it now duplicates this field?

client/mainmenu/CHighScoreScreen.h
IvanSavenko: Why not rename?

Laserlicht:
@IvanSavenko done :)

IvanSavenko (approved):
Still think that score should be computed on server and passed as part of `PlayerEndsGame` netpack, but I guess not critical.

Laserlicht:
@Nordsoft91 fixed:)
Thanks!

PR #2924, opened 2023-09-23
Datetime android

Laserlicht:
Specific function for android, as current function do not work correctly.

### DIFF
--- a/lib/vstd/DateUtils.cpp
+++ b/lib/vstd/DateUtils.cpp
@@ -1,6 +1,10 @@
 #include "StdInc.h"
 #include <vstd/DateUtils.h>
 
+#if defined(VCMI_ANDROID)
+#include "../CAndroidVMHelper.h"
+#endif
+
 VCMI_LIB_NAMESPACE_BEGIN
 
 namespace vstd
@@ -8,6 +12,11 @@ namespace vstd
 
 	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt)
 	{
+#if defined(VCMI_ANDROID)
+        CAndroidVMHelper vmHelper;
+        return vmHelper.callStaticStringMethod(CAndroidVMHelper::NATIVE_METHODS_DEFAULT_CLASS, "getFormattedDateTime");
+#endif
+
 		std::tm tm = *std::localtime(&dt);
 		std::stringstream s;
 		try
--- a/lib/vstd/DateUtils.cpp
+++ b/lib/vstd/DateUtils.cpp
@@ -1,8 +1,9 @@
 #include "StdInc.h"
 #include <vstd/DateUtils.h>
 
-#include "../CConfigHandler.h"
-#include "../Languages.h"
+#if defined(VCMI_ANDROID)
+#include "../CAndroidVMHelper.h"
+#endif
 
 VCMI_LIB_NAMESPACE_BEGIN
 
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
> current function do not work correctly

could you provide more info?

Laserlicht:
My android devices outputs `09/23/23 21:41:55` which is incorrect for germany. Now it outputs `23.09.23 21:41`. Not a perfect locale, but at least correct. :)

kambala-decapitator:
on iOS I see `Mon Jul 31 18:35:10 2023` which also doesn't match my current locale ru_RU.

could you try using boost.datetime instead? maybe it can also solve your case and at the same time provide improvement for iOS.

Laserlicht:
I think with boost::datetime we would have the same problem.

I tried another approach based on selected language of vcmi.

lib/vstd/DateUtils.cpp
```diff
@@ -12,16 +11,17 @@ namespace vstd
...
 
 		std::tm tm = *std::localtime(&dt);
 		std::stringstream s;
 		try
 		{
-			s.imbue(std::locale(""));
+			if(locale.empty())
+				s.imbue(std::locale(""));
+			else
+				s.imbue(std::locale(locale));
```
IvanSavenko, line 24: Using boost::locale (or replacing all this handling with boost::datetime) might be a good idea.

Problem is, c++ standard does not defines list of supported encodings. 
Whether "de_DE.UTF-8" exists and/or is usable depends on OS, OS language, OS version and phase of the moon.

For reliable, cross-platform behavior boost might be better option

Laserlicht: Do you have an example? I do not find anything fitting our usecase.

Only examples that still using `imbue`.

IvanSavenko: boost locale likely uses libUnicode or some other more reliable locale generation than OS.

So imbuing locale with whatever boost locale provides might be better option. However please check for 'imbue' references in code. I think we already use boost locale somewhere to initialize global locale. But I think only in some cases?

Laserlicht: I don't get it working...

The result (at least for "de_DE.UTF-8" is incorrect.

```
	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt)
	{
		std::string lang = settings["general"]["language"].String();
		std::string localeCode = Languages::getLanguageOptions(lang).locale;
		boost::locale::generator localeGenerator = boost::locale::generator();
		boost::gregorian::date_facet *facet = new boost::gregorian::date_facet{"%x %X"};
		std::locale locale = std::locale(localeCode.empty() ? localeGenerator("") : localeGenerator(localeCode), facet);
		boost::posix_time::ptime tm = boost::posix_time::ptime(boost::gregorian::date(1970, 1, 1)) + boost::posix_time::seconds(static_cast<long>(dt));

		std::stringstream s;
		try
		{
			s.imbue(locale);
		}
		catch(const std::runtime_error & e)
		{
			// locale not be available - keep default / global
		}
		s << tm;
		return s.str();
	}
```

Laserlicht: ICU as backend gives incorrect results. If I switch to std. It's fine on PC. But not android

Laserlicht:
I think currently this android specific case is the only working solution.

Maybe someone with iOS could add a similar solution for this.

Issue #2898, opened 2023-09-22
When creature attacks while it's standing in a moat, sometimes it won't receive damage from moat


HarkUhm:
**Describe the bug**
I can't tell when exactly, but during playing a second  PAH tournament map, creatures very often didn't get damage from moat, when they're already standing in it and attacking an adjacent monster.

**Expected behavior**
Creature standing in a moat should take damage after attacking an adjacent monster and it's counterattack.

**Version**
 - OS: [Windows 10]
 - Version: [VCMI 1.3.2]

dydzio0614:
i think they should receive damage after, not before attacking

krs0:
When testing moats, everything looked fine. 
Please note the correct behaviour of moats:

Moats cause damage to creatures when they enter the moat, or end a turn into one.

If this is note the case, please note what town it was or what creature was not affected.

rilian-la-te:
@krs0 exactly what I want to say)

HarkUhm:
![sc](https://github.com/vcmi/vcmi/assets/141812373/1dc929fe-f021-4e65-8541-99e29bfababb)

rilian-la-te:
@HarkUhm is this occured for vanilla towns?

IvanSavenko:
@rilian-la-te Can't think of any related changes.

rilian-la-te:
@IvanSavenko so, let's wait to see if vanilla towns is affected. If not - certainly a mod issue.

misiokles:
I saw thi gif before and I though it's missing damage hex in moat damage configuration. But Uhm said that this lower Pyro received damage only once, in first turn. Such things seem to be irrevelant to mods.

rilian-la-te:
@misiokles it can be relevant to mods, if pyro just crosses a moat and stand on misconfigured hex. It need to be checked for vanilla towns, if bug will be confirmed - will spend weekend on this instead of BG3)

misiokles:
This is quick check of Forge siege screen and battlefield grid. It seems to hexes 164 ad 181 in Forge's moat config are placed correctly.
![Bez nazwy](https://github.com/vcmi/vcmi/assets/29593042/b4c1e85f-e2c3-4b58-adb7-48788db83780)

rilian-la-te:
@misiokles where are you checking? In moat spell page? Or in town page?

misiokles:
In moat spell page. In vcmi-1.3 is the only place, that such hexes are defined.

[...]

PR #2925, opened 2023-09-23
CGameHandler: explicitly handle obstacles on stand

rilian-la-te:
Start is always added inside "passed" array, which cause obstacle parser to ignore it. So, let's trigger obstacles when just standing still by emptying a passed (as done with one hex trigger)

Fixes: #2898


rilian-la-te:
@IvanSavenko

PR #2934, opened 2023-09-23
Hotseat hide

### LINKED ISSUE #2573: minimap is revealed when turn is passed
**Describe the bug**
Minimap is not hidden when turn is passed to another player in hotseat mode. Thus, player can see opponent's location.

**Did it work earlier?**
IDK

**Screenshots**
Снимок экрана 2023-08-13 в 01 32 32

**Version**
 - Version: 1.3

**Additional context**

Laserlicht:
fixes #2573

What do you think?

Uses dimming to hide complete screen.

image

### DIFF
--- a/client/adventureMap/AdventureMapInterface.cpp
+++ b/client/adventureMap/AdventureMapInterface.cpp
@@ -51,7 +51,7 @@ AdventureMapInterface::AdventureMapInterface():
 	spellBeingCasted(nullptr),
 	scrollingWasActive(false),
 	scrollingWasBlocked(false),
-	isHotseatMessage(false)
+	backgroundDimLevel(settings["adventure"]["backgroundDimLevel"].Integer())
 {
 	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
 	pos.x = pos.y = 0;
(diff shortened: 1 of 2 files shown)


Nordsoft91:
looks like workaround. Shouldn't be FOW instead?

Laserlicht:
Yes, a workaround. Until we have a better solution. :)

client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -172,6 +173,8 @@ void AdventureMapInterface::dim(Canvas & to)
 		if (!casted && !window->isPopupWindow())
 		{
 			int backgroundDimLevel = settings["adventure"]["backgroundDimLevel"].Integer();
+			if(isHotseatMessage)
+				backgroundDimLevel = 255;
```
Nordsoft91, line 177: why not moving backgroundDimLevel to class fields and set it to 255/settings from onHotseatWaitStarted/onPlayerTurnStarted methods?

Laserlicht: Would also be possible. But I think it's more understandable like this? Or not?

Nordsoft91: surely it's personal preference. Extra variable which is used only to control another variable from another method a bit confusing. Controlling dimming directly in onHotseatWaitStarted states clearly, that you want to dim screen on hotseat message

Laserlicht: Okay, convinced. Are fewer lines. :)

IvanSavenko:
> looks like workaround. Shouldn't be FOW instead?

Strictly speaking, fow cover should be trivial - add one check to `MapRendererBaseContext::isVisible`

However it won't cover town list / hero list / minimap / etc. So might not be a bad solution meanwhile.

Laserlicht:
> However it won't cover town list / hero list / minimap / etc. So might not be a bad solution meanwhile.

This was the intention of this PR.

PR #2939, opened 2023-09-24
fix some ukrainian strings in the android launcher

utzuro:
There were few mistakes in the Ukrainian translation for the Android version of the launcher, so I've fixed them.


IvanSavenko:
Дякую :)

Completely forgot that we also have translation in Android launcher.

PR #2947, opened 2023-09-25
enableUiEnhancements button

Laserlicht:
Since I added some features using this flag it would be good to have it on UI. To enable classic experience.

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -90,6 +90,8 @@
 	"vcmi.systemOptions.framerateButton.help"   : "{Show FPS}\n\nToggle the visibility of the Frames Per Second counter in the corner of the game window.",
 	"vcmi.systemOptions.hapticFeedbackButton.hover"  : "Haptic feedback",
 	"vcmi.systemOptions.hapticFeedbackButton.help"   : "{Haptic feedback}\n\nToggle the haptic feedback on touch inputs.",
+	"vcmi.systemOptions.enableUiEnhancementsButton.hover"  : "Interface Enhancements",
+	"vcmi.systemOptions.enableUiEnhancementsButton.help"   : "{Interface Enhancements}\n\nToggle various quality of life interface improvements.",
 
 	"vcmi.adventureOptions.infoBarPick.hover" : "Show Messages in Info Panel",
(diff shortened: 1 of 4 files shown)


Mods/vcmi/config/vcmi/english.json
```diff
@@ -90,6 +90,8 @@
 	"vcmi.systemOptions.framerateButton.help"   : "{Show FPS}\n\nToggle the visibility of the Frames Per Second counter in the corner of the game window.",
 	"vcmi.systemOptions.hapticFeedbackButton.hover"  : "Haptic feedback",
 	"vcmi.systemOptions.hapticFeedbackButton.help"   : "{Haptic feedback}\n\nToggle the haptic feedback on touch inputs.",
+	"vcmi.systemOptions.enableUiEnhancementsButton.hover"  : "Interface Enhancements",
+	"vcmi.systemOptions.enableUiEnhancementsButton.help"   : "{Interface Enhancements}\n\nToggle various quality of life interface improvements.",
```
Nordsoft91, line 94: Not very informative. Maybe enumerate at least most important points?

Laserlicht: This is intentional, because it would have to be constantly maintained. Should rather be in docs I think.

Nordsoft91: I assuming option is enabled by default? If so, you can add something like you have in PR description: "... disable this option if you want classic experience from original game"

IvanSavenko: Possibly describe 2-3 most noticeable changes,e.g. "such as large spell book, spell icons for scrolls, etc"?

And maybe add more complete version somewhere in the docs?

Nordsoft91:
New button overlaps with "Town screen" label

Снимок экрана 2023-09-26 в 20 20 45

Laserlicht:
ah... tested it on android... there is space. :)

Do we have space at the bottom? To move everything one line down?

Nordsoft91:
I think so, but horizontal line probably needs to be adjusted
Снимок экрана 2023-09-26 в 20 40 42

Nordsoft91:
Another option: replace 2 fullscreen toggles with one list/pop-up (same as for resolution selection) with items: windowed, fullscreen (borderless), fullscreen (exclusive)

Laserlicht:
Is fine now. Moved the bottom part to the bottom

PR #2951, opened 2023-09-25
specific spell icon (artifacts)

Laserlicht:
As requested by @dydzio0614 

image

### DIFF
--- a/client/widgets/CArtifactHolder.cpp
+++ b/client/widgets/CArtifactHolder.cpp
@@ -41,6 +41,10 @@ void CArtPlace::setInternals(const CArtifactInstance * artInst)
 		return;
 	}
 
+	image->enable();
+	imageSpell->disable();
+	image->setFrame(artInst->artType->getIconIndex());
+	
 	if(artInst->getTypeId() == ArtifactID::SPELL_SCROLL)
 	{
 		auto spellID = artInst->getScrollSpellID();
@@ -51,13 +55,7 @@ void CArtPlace::setInternals(const CArtifactInstance * artInst)
 			type = spellID;
 			bonusValue = 0;
 
-			if(!settings["general"]["enableUiEnhancements"].Bool())
-			{
-				image->enable();
-				imageSpell->disable();
-				image->setFrame(artInst->artType->getIconIndex());
-			}
-			else
+			if(settings["general"]["enableUiEnhancements"].Bool())
 			{
 				imageSpell->enable();
 				image->disable();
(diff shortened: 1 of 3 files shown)


dydzio0614:
hella good, this is definitely what players want

krs0:
I get keeping OH3 as close as possible... but having extra logic for such a no brainer feature...
I would just enable this and that's it. Same with spell book.

Laserlicht:
> I get keeping OH3 as close as possible... but having extra logic for such a no brainer feature... I would just enable this and that's it. Same with spell book.

For this: Maybe...

For the spellbook: I would say no. The small spellbook has animations as additional feature.

dydzio0614:
I saw somebody few months ago in one H3 discord who hated large spellbook and was disappointed that 3 without HD mod did not run properly on his windows 11 so...

krs0:
Imo this should not turn into default approach for every single modification, just because its possible. It just adds complexity.

dydzio0614:
I think every extra comma in text should be optional if it doesn't belong in H3

Laserlicht:
Yes, I think the same as @dydzio0614 . In my opinion it makes the code better in many cases. Because if you search for the switch-flag you can see cases with different handling than in OH3. Since some features are quite tricky (because not designed by Heroes).

Of course, in the past, many features have been added without an option.

Nordsoft91:
And that's why we have configurable interface and vcmi-extras mod.
HDmod options to he enabled with mod while disabling of it keeps original game in (almost) all UI aspects.
This approach allows to have clear code without multiple if/else conditions

IvanSavenko:
> And that's why we have configurable interface and vcmi-extras mod.

VCMI Extras needs to die:
- Content of questionable copyright / legality
- Not possible to install without restart
- Not possible to enable without restart
- Endless questions from players how to enable corresponding functionality (and who knows how many players who have not asked and dropped VCMI)

> This approach allows to have clear code without multiple if/else conditions

Spellbook would benefit from configurable UI - by providing separate layout file for small & large version. But spell icon for scrolls? Don't see how it can be made better than this.

IvanSavenko (approved):

Laserlicht:
> Spellbook would benefit from configurable UI - by providing separate layout file for small & large version. But spell icon for scrolls? Don't see how it can be made better than this.

Yes, for spellbook it would make sense. For other layouts, etc. But for this it would have to completely rewritten (I don't think that someone will do this task now).

By the way: the "generated" big spellbook should be compatible with the HD one and should theoretically replaceable with this (expect the bookmark icon positions).

PR #2952, opened 2023-09-26
Enable VSync to prevent screen tearing while scrolling across map

### LINKED ISSUE #2935: Enable VSync to prevent screen tearing
**Describe your proposal**
When scrolling across the map, screen tearing occurs.

**Additional context**
I took a screenshot and made a screen recording, but they didn't display the issue.

I aleady tried adding `SDL_SetHint(SDL_HINT_RENDER_VSYNC, "1");` to client/renderSDL/ScreenHandler.cpp, but that didn't help.

Test build with `SDL_HINT_RENDER_VSYNC`: https://github.com/flathub/eu.vcmi.VCMI/pull/143

Alexander-Wilms:
Fixes #2935

With this change, I experience no more screen tearing when moving across the map.

Should this be optional?

### DIFF
--- a/client/gui/FramerateManager.cpp
+++ b/client/gui/FramerateManager.cpp
@@ -11,22 +11,28 @@
 #include "StdInc.h"
 #include "FramerateManager.h"
 
+#include "../../lib/CConfigHandler.h"
+
 FramerateManager::FramerateManager(int targetFrameRate)
 	: targetFrameTime(Duration(boost::chrono::seconds(1)) / targetFrameRate)
 	, lastFrameIndex(0)
 	, lastFrameTimes({})
-	, lastTimePoint (Clock::now())
+	, lastTimePoint(Clock::now())
+	, vsyncEnabled(settings["video"]["vsync"].Bool())
 {
 	boost::range::fill(lastFrameTimes, targetFrameTime);
 }
 
 void FramerateManager::framerateDelay()
 {
-	Duration timeSpentBusy = Clock::now() - lastTimePoint;
+	if(!vsyncEnabled)
+	{
+		Duration timeSpentBusy = Clock::now() - lastTimePoint;
 
-	// FPS is higher than it should be, then wait some time
-	if(timeSpentBusy < targetFrameTime)
-		boost::this_thread::sleep_for(targetFrameTime - timeSpentBusy);
+		// FPS is higher than it should be, then wait some time
+		if(timeSpentBusy < targetFrameTime)
+			boost::this_thread::sleep_for(targetFrameTime - timeSpentBusy);
+	}
 
 	// compute actual timeElapsed taking into account actual sleep interval
 	// limit it to 100 ms to avoid breaking animation in case of huge lag (e.g. triggered breakpoint)
--- a/client/gui/FramerateManager.cpp
+++ b/client/gui/FramerateManager.cpp
@@ -22,7 +22,7 @@ FramerateManager::FramerateManager(int targetFrameRate)
 {
 	if(vsyncEnabled)
 	{
-		static int display_in_use = 0;
+		static int display_in_use = settings["video"]["displayIndex"].Integer();
 		SDL_DisplayMode mode;
 		SDL_GetCurrentDisplayMode(display_in_use, &mode);
 		int displayRefreshRate = mode.refresh_rate;
--- a/client/gui/FramerateManager.cpp
+++ b/client/gui/FramerateManager.cpp
@@ -12,27 +12,36 @@
 #include "FramerateManager.h"
 
 #include "../../lib/CConfigHandler.h"
+#include <SDL_video.h>
 
 FramerateManager::FramerateManager(int targetFrameRate)
-	: targetFrameTime(Duration(boost::chrono::seconds(1)) / targetFrameRate)
-	, lastFrameIndex(0)
+	: lastFrameIndex(0)
 	, lastFrameTimes({})
 	, lastTimePoint(Clock::now())
 	, vsyncEnabled(settings["video"]["vsync"].Bool())
 {
+	if(vsyncEnabled)
+	{
+		static int display_in_use = 0;
+		SDL_DisplayMode mode;
+		SDL_GetCurrentDisplayMode(display_in_use, &mode);
+		int displayRefreshRate = mode.refresh_rate;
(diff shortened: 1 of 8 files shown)


Laserlicht:
What happens with the FPS setting after enabling vsync?

Would it be overwritten (set to monitor fps) and useless?

dydzio0614:
Commercial games make this optional, and I also think it should be
Especially that ability to setup 30 FPS like original H3 is a thing

Alexander-Wilms:
@Laserlicht When I set my monitor to 120 Hz, VCMI's framerate limit to 256 and run VCMI with VSync, the FPS counter shows 120 FPS and I think that's accurate. If I set the limit to 60, the cursor does appear to be rendered at 120 FPS, the rest of the game at 60 FPS.

kambala-decapitator:
If you use hardware cursor (default on desktop), then I think it's expected behavior

IvanSavenko:
Should probably be optional:
- add new option to config/schemas/settings.json (make sure to add it to "required" section as well
- pass flag only if `settings["video"]["vsync"].Bool()` is set to true

Laserlicht:
But... im not sure what's happening in case of lags... with the `getFramerate()`-Function.

Is it possible to get the vsync target fps? And use this one?

IvanSavenko:
> @Alexander-Wilms Since with this approach the target fps can be much more different than the actual fps:

Should be the same? With this change delay should be done by SDL and not but our framerate controller. But should not affect time measurements.

Laserlicht:
> > @Alexander-Wilms Since with this approach the target fps can be much more different than the actual fps:
> 
> Should be the same? With this change delay should be done by SDL and not but our framerate controller. But should not affect time measurements.

Yes, of course. But with Vsync not every fps is possible (except with adaptive sync). Only frame rates that are divisible by the frequency of the monitor.

And in this specific case the setting is used, not the vsync target fps.

Alexander-Wilms:
Like this?


```
float fps;
if(!settings["video"]["vsync"].Bool())
{
    fps = settings["video"]["targetfps"].Float();
} else
{
    // have not yet found an SDL function for getting the VSync rate
    fps = SDL_...
}
postSwipeCatchIntervalMs = std::max(100, static_cast<int>(6.0 * 1000.0 * (1.0 / )));
```

IvanSavenko:
I think this is not needed. Your PR looks fine as it is.

Yes, with vsync *frame time* can only be a multiple of monitor frame time - so 16.6*X for 60-fps display
But *framerate* which is average number of frames over last seconds can have any value, not necessarily 60 or 30.

Laserlicht:
I think we can try it.

@Nordsoft91 Can you try this: https://github.com/vcmi/vcmi/pull/2679
on iOS with low target fps (e.g. 25 fps) if it is still fine with vsync.

Alexander-Wilms:
With this change, the hero freezes a bit when moving.

I'll check if develop doesn't exhibit this.

Nordsoft91:
Actually, I'd prefer to keep target fps, because reducing it helps to save battery time, which is critical for mobile devices

Laserlicht:
> Actually, I'd prefer to keep target fps, because reducing it helps to save battery time, which is critical for mobile devices

As far as I understand it is only disabled when vsync is on. So for the energy saving case you have to disable vsync and set target fps.

Alexander-Wilms:
Maybe 

```
ui32 FramerateManager::getElapsedMilliseconds() const
{
	return lastFrameTimes[lastFrameIndex] / boost::chrono::milliseconds(1);
}
```

should return a float?

client/gui/FramerateManager.cpp
```diff
@@ -11,12 +11,27 @@
...
-	, lastFrameIndex(0)
+	: lastFrameIndex(0)
 	, lastFrameTimes({})
-	, lastTimePoint (Clock::now())
+	, lastTimePoint(Clock::now())
+	, vsyncEnabled(settings["video"]["vsync"].Bool())
 {
+	if(vsyncEnabled)
+	{
+		static int display_in_use = 0;
```
Laserlicht, line 25: Can you use settings["video"]["displayIndex"] instead of hardcoding the display?

Alexander-Wilms:
Should I get rid of the vsyncEnabled attribute, since it's only used in the constructor?

Nordsoft91:
> Should I get rid of the vsyncEnabled attribute, since it's only used in the constructor?

I would remove it

client/gui/FramerateManager.cpp
```diff
@@ -11,12 +11,26 @@
...
+		static int display_in_use = settings["video"]["displayIndex"].Integer();
+		SDL_DisplayMode mode;
+		SDL_GetCurrentDisplayMode(display_in_use, &mode);
+		int displayRefreshRate = mode.refresh_rate;
+		logGlobal->info("Display refresh rate is %d", displayRefreshRate);
+		targetFrameTime = Duration(boost::chrono::seconds(1)) / displayRefreshRate;
+	} else
+	{
+		targetFrameTime = Duration(boost::chrono::seconds(1)) / targetFrameRate;
+	}
```
IvanSavenko, line 33: Why you decided to change the approach?

Previous solution by changing sleep should have been sufficient and more correct.

Alexander-Wilms: The previous approach led to the animation of the moving hero freezing every second or so.

IvanSavenko: That's a separate bug. Your approach was correct.

Try changing CGuiHandler::renderFrame like this:
```cpp
void CGuiHandler::renderFrame()
{
	{
		boost::recursive_mutex::scoped_lock un(*CPlayerInterface::pim);

		if(nullptr != curInt)
			curInt->update();

		if(settings["video"]["showfps"].Bool())
			drawFPSCounter();

		SDL_UpdateTexture(screenTexture, nullptr, screen->pixels, screen->pitch);

		SDL_RenderClear(mainRenderer);
		SDL_RenderCopy(mainRenderer, screenTexture, nullptr, nullptr);

		CCS->curh->render();
		windows().onFrameRendered();
	}
	SDL_RenderPresent(mainRenderer);
	framerate().framerateDelay(); // holds a constant FPS
}

Problem is - game maintains lock on "CPlayerInterface::pim" while wating for vsync, blocking any other actions such as client<->server communication

IvanSavenko (approved):

PR #2957, opened 2023-09-26
Fixed some issues found by pvs

ProjectMutilation:


AI/Nullkiller/AIGateway.cpp
```diff
@@ -710,8 +710,10 @@ void AIGateway::showGarrisonDialog(const CArmedInstance * up, const CGHeroInstan
 	//you can't request action from action-response thread
 	requestActionASAP([=]()
 	{
-		if(removableUnits && up->tempOwner == down->tempOwner)
+		if(removableUnits && up && down && up->tempOwner == down->tempOwner)
```
IvanSavenko, line 713: showGarrisonDialog is always called with non-null up & down. Othervice there is no point in such call - garrison dialog is about exchange between two armies.

I guess incorrect code is above this call:
```cpp
std::string s1 = up ? up->nodeName() : "NONE";
std::string s2 = down ? down->nodeName() : "NONE";
```
Check for nullptr is redundant in there.

ProjectMutilation: @IvanSavenko sorry, when I did rebase, I touched extra commits, but there should be no problems with merging
if it's problem can create PR with clear history

IvanSavenko: Why you can't simply drop them via interactive rebase?

Mods/vcmi/mod.json
Nordsoft91, line 1: unnecessary changes

AI/Nullkiller/AIGateway.cpp
ProjectMutilation: > Why you can't simply drop them via interactive rebase?

fixed, I didn't know about the "drop" directive in rebase, thanks

IvanSavenko (approved):

PR #2960, opened 2023-09-26
Enable ccache for CI

Alexander-Wilms:
This had been removed in the past:
https://github.com/vcmi/vcmi/commit/6e5ad22efc8ea74684980a745236518680b1dafd

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -158,6 +158,15 @@ jobs:
       env:
         VCMI_BUILD_PLATFORM: x64
 
+    - name: ccache
+      if: startsWith(matrix.preset, 'ios') != true
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+          key: ${{ matrix.preset }}
+          # actual cache takes up less space, at most ~1 GB
+          max-size: "5G"
+          verbose: 2
+
     - uses: actions/setup-python@v4
       if: "${{ matrix.conan_profile != '' }}"
       with:
@@ -193,7 +202,13 @@ jobs:
       env:
         PULL_REQUEST: ${{ github.event.pull_request.number }}
 
-    - name: CMake Preset
+    - name: CMake Preset with ccache
+      if: startsWith(matrix.preset, 'ios') != true
+      run: |
+        cmake -DCMAKE_CXX_COMPILER_LAUNCHER=ccache --preset ${{ matrix.preset }}
+
+    - name: CMake Preset without ccache
+      if: startsWith(matrix.preset, 'ios')
       run: |
         cmake --preset ${{ matrix.preset }}
 
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -25,7 +25,8 @@
                 "CMAKE_BUILD_TYPE": "RelWithDebInfo",
                 "ENABLE_TEST": "OFF",
                 "ENABLE_STRICT_COMPILATION": "ON",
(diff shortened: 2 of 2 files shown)


Alexander-Wilms:
Together with https://github.com/hendrikmuhs/ccache-action this should also speed up the CI builds

kambala-decapitator:
no, this must be user's choice. https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_LAUNCHER.html

to enable ccache for CI, adjust cmake invocation. But are you sure that it'd have any effect on CI? AFAIK GitHub provides clean agents on every launch. For this to work, build artifacts must be cached between launches.

Alexander-Wilms:
I couldn't get CMAKE_CXX_COMPILER_LAUNCHER to work reliably, but I'll test it some more and maybe add it to the wiki.

Github actions do offer a cache: https://github.com/hendrikmuhs/ccache-action#how-it-works

Alexander-Wilms:
For the step `Build Preset` of the `linux-gcc-debug`, this reduced the build time from 37 min to 18 min.

kambala-decapitator:
AFAIK ccache needs more work to be enabled for xcode generator (ios platform uses it), hence you might want to disable ccache for ios

Alexander-Wilms:
The ccache steps only add 20 s to the iOS build, so I'd prefer to use the same config for all matrix entries. Once someone adds iOS-specific ccache configuration, it could be adapted. What do you think?

Well, the `hendrikmuhs/ccache-action` does all the heavy lifting.

kambala-decapitator:
> Well, the hendrikmuhs/ccache-action does all the heavy lifting.

Yes, that's it. But haven't tested it yet.

If it adds only a tiny amount of time, then it's fine.

Alexander-Wilms:
Do you know whether precompiled headers can be disabled for `linux-gcc-test`?

rilian-la-te:
@Alexander-Wilms PCH is desablef for linux-gcc-debug target. Why you want to disable it for linux-gcc-test too?

Alexander-Wilms:
So that ccache works with the `linux-gcc-test` preset.

Cf. https://github.com/Alexander-Wilms/vcmi/actions/runs/6341021439/job/17223723579#step:39:13

rilian-la-te:
@Alexander-Wilms so, copy ENABLE_PCH entry from linux-gcc-debug preset inside CMakePresets.json inside linux-gcc-test preset and try)

Alexander-Wilms:
It seems the action doesn't always use the latest cache. Perhaps that is a limitation of Github, but once it's merged, this shouldn't be much of an issue, I hope.

Alexander-Wilms:
`cache hit rate                     89.11 %` and 8:16 min for the `Build Preset` step of `linux-gcc-debug` using a cache size of 2 GB.

rilian-la-te:
@Alexander-Wilms please, return strict compilation, it enables -Werror.

rilian-la-te:
@kambala-decapitator can you fix building iOS without PCH?

Alexander-Wilms:
And should I add `CMAKE_CXX_COMPILER_LAUNCHER=ccache` to the `base-release` CMake preset?

Alexander-Wilms:
I can't get the iOS build to succeed, no matter whether I use ENABLE_PCH= ON or OFF. It doesn't even use ccache.

Alexander-Wilms:
Should be ready to merge. If there are no changes between runs, most matrix entries take around 4 min to complete, only ios and msvc take longer.

PR #2969, opened 2023-09-27
Add query for Tavern, Dwelling and Market visits

IvanSavenko:
Adventure map Tavern and all markets now generate queries for duration of the visit until client closes corresponding window. Taverns / markets in towns do not generate queries

- Fixes broken visitation to these buildings from recent PR.
- Allows proper validation of requests from players by server

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -324,10 +324,13 @@ void AIGateway::heroPrimarySkillChanged(const CGHeroInstance * hero, PrimarySkil
 	NET_EVENT_HANDLER;
 }
 
-void AIGateway::showRecruitmentDialog(const CGDwelling * dwelling, const CArmedInstance * dst, int level)
+void AIGateway::showRecruitmentDialog(const CGDwelling * dwelling, const CArmedInstance * dst, int level, QueryID queryID)
 {
 	LOG_TRACE_PARAMS(logAi, "level '%i'", level);
 	NET_EVENT_HANDLER;
+
+	status.addQuery(queryID, "RecruitmentDialog");
+	requestActionASAP([=](){ answerQuery(queryID, 0); });
 }
 
 void AIGateway::heroMovePointsChanged(const CGHeroInstance * hero)
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -330,7 +330,11 @@ void AIGateway::showRecruitmentDialog(const CGDwelling * dwelling, const CArmedI
 	NET_EVENT_HANDLER;
 
 	status.addQuery(queryID, "RecruitmentDialog");
-	requestActionASAP([=](){ answerQuery(queryID, 0); });
+
+	requestActionASAP([=](){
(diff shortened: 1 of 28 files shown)


server/queries/MapQueries.cpp
```diff
@@ -157,6 +157,50 @@ CBlockingDialogQuery::CBlockingDialogQuery(CGameHandler * owner, const BlockingD
...
+			return false;
+
+		if(dynamic_ptr_cast<AssembleArtifacts>(pack))
+			return false;
+
+		if(dynamic_ptr_cast<EraseArtifactByClient>(pack))
+			return false;
+
+		if(dynamic_ptr_cast<TradeOnMarketplace>(pack) != nullptr)
+			return false;
```
Nordsoft91, line 198:
```suggestion
		if(dynamic_ptr_cast<HireHero>(pack))
			return false;
	}

	if (mode == EOpenWindowMode::UNIVERSITY_WINDOW)
	{
		if(dynamic_ptr_cast<TradeOnMarketplace>(pack))
			return false;
	}

	if (mode == EOpenWindowMode::MARKET_WINDOW)
	{
		if(dynamic_ptr_cast<ExchangeArtifacts>(pack))
			return false;

		if(dynamic_ptr_cast<AssembleArtifacts>(pack))
			return false;

		if(dynamic_ptr_cast<EraseArtifactByClient>(pack))
			return false;

		if(dynamic_ptr_cast<TradeOnMarketplace>(pack))
			return false;
```

Nordsoft91: Can we inherit all those packs from one empty class to have only single cast check? Thus, we can simplify adding of new packs in the future

AI/Nullkiller/AIGateway.cpp
```diff
@@ -154,10 +154,13 @@ void AIGateway::artifactAssembled(const ArtifactLocation & al)
...
 }
 
-void AIGateway::showTavernWindow(const CGObjectInstance * townOrTavern)
+void AIGateway::showTavernWindow(const CGObjectInstance * object, const CGHeroInstance * visitor, QueryID queryID)
 {
 	LOG_TRACE(logAi);
 	NET_EVENT_HANDLER;
+
+	status.addQuery(queryID, "TavernWindow");
+	requestActionASAP([=](){ answerQuery(queryID, 0); });
```
Nordsoft91, line 163: I was thinking NKAI uses advmap taverns to hire heroes too

server/queries/MapQueries.cpp
IvanSavenko: Probably not this time.
We also have similar (and even longer) set of checks for exchange dialog.
We also may want to examine content of netpack to allow only packs that affect objects involved in query so simple empty class won't work.

AI/Nullkiller/AIGateway.cpp
IvanSavenko: Does not looks like it. From what I see NKAI can only recruit heroes from towns.

nullkiller: Yep, I do not remember such functionality so expect that it can not

PR #2990, opened 2023-09-30
Map preview rework

Laserlicht:
- new layout
- supports loadgame for scenario and campaign (only map, without map objects and player colors)
- configurable

grafik


client/windows/CMapOverview.h
```diff
@@ -0,0 +1,59 @@
...
+class IImage;
+class Canvas;
+class TransparentFilledRectangle;
+enum ESelectionScreen : ui8;
+
+class CMapOverview : public CWindowObject
+{
+	class CMapOverviewWidget : public InterfaceObjectConfigurable
+	{
+		CMapOverview& parent;
```
Nordsoft91, line 33: I think it hides parent field from CIntObject

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+	for (int y = 0; y < map->height; ++y)
+		for (int x = 0; x < map->width; ++x)
+		{
+			TerrainTile & tile = map->getTile(int3(x, y, layer));
+
+			ColorRGBA color = tile.terType->minimapUnblocked;
+			if (tile.blocked && (!tile.visitable))
+				color = tile.terType->minimapBlocked;
+
+			if(drawPlayerElements)
```
Nordsoft91, line 79: maybe pass it as an argument?

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+std::vector<std::shared_ptr<IImage>> CMapOverview::CMapOverviewWidget::createMinimaps(std::unique_ptr<CMap> & map, Point size) const
+{
+	std::vector<std::shared_ptr<IImage>> ret = std::vector<std::shared_ptr<IImage>>();
+
+	for(int i = 0; i < (map->twoLevel ? 2 : 1); i++)
+	{
+		Canvas canvas = createMinimapForLayer(map, i);
+		Canvas canvasScaled = Canvas(size);
+		canvasScaled.drawScaled(canvas, Point(0, 0), size);
+		std::shared_ptr<IImage> img = GH.renderHandler().createImage(canvasScaled.getInternalSurface());
```
Nordsoft91, line 129: Why createMinimapForLayer returns Canvas and here you make IImage from it? Maybe createMinimapForLayer can return IImage directly?

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+	auto rect = readRect(config["rect"]);
+	auto color = readColor(config["color"]);
+	if(!config["colorLine"].isNull())
+	{
+		auto colorLine = readColor(config["colorLine"]);
+		return std::make_shared<TransparentFilledRectangle>(rect, color, colorLine);
+	}
+	return std::make_shared<TransparentFilledRectangle>(rect, color);
+}
```
Nordsoft91, line 149: This is common widget - let's move it to InterfaceObjectConfigurable class

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+{
+	logGlobal->debug("Building widget drawPath");
+
+	auto rect = readRect(config["rect"]);
+	auto font = readFont(config["font"]);
+	auto alignment = readTextAlignment(config["alignment"]);
+	auto color = readColor(config["color"]);
+
+	return std::make_shared<CTextBox>(parent.fileName, rect, 0, font, alignment, color);
+}
```
Nordsoft91, line 200: CTextBox is a common widget - deserves to be in InterfaceObjectConfigurable class

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+		{
+			std::time_t time = boost::filesystem::last_write_time(*CResourceHandler::get()->getResourceName(ResourcePath(parent.resource.getName(), EResType::MAP)));
+			text = vstd::getFormattedDateTime(time);
+		}
+		else
+			text = parent.date;
+	}
+	auto position = readPosition(config["position"]);
+	return std::make_shared<CLabel>(position.x, position.y, font, alignment, color, text);
+}
```
Nordsoft91, line 224: This is not a way how it supposed to be. You shouldn't create new object type every time when you need to override one field. Here it's sufficient to build simple CLabel with special name:
```json
...
{
  "type": "label",
  "name": "mapLabel",
  "text": "%mapname%"
}
```

Then in constructor
```c++
build(config);

if(auto w = widget<CLabel>("mapLabel"))
{
   //do special logic for text parsing
}
```

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+	renderImage = settings["lobby"]["mapPreview"].Bool();
+
+	const JsonNode config(JsonPath::builtin("config/widgets/mapOverview.json"));
+
+	REGISTER_BUILDER("drawTransparentRect", &CMapOverview::CMapOverviewWidget::buildDrawTransparentRect);
+	REGISTER_BUILDER("drawMinimap", &CMapOverview::CMapOverviewWidget::buildDrawMinimap);
+	REGISTER_BUILDER("drawPath", &CMapOverview::CMapOverviewWidget::buildDrawPath);
+	REGISTER_BUILDER("drawString", &CMapOverview::CMapOverviewWidget::buildDrawString);
+
+	build(config);
```
Nordsoft91, line 239: you can move all specific logic afterwards
```c++
if(auto w = widget<Type>(name))
{
  //do with widget what you want
}
```

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+	if(!renderImage)
+		return nullptr;
+
+	ResourcePath res = ResourcePath(parent.resource.getName(), EResType::MAP);
+	std::unique_ptr<CMap> campaignMap = nullptr;
+	if(parent.tabType != ESelectionScreen::newGame)
+	{
+		CLoadFile lf(*CResourceHandler::get()->getResourceName(ResourcePath(parent.resource.getName(), EResType::SAVEGAME)), MINIMAL_SERIALIZATION_VERSION);
+		lf.checkMagicBytes(SAVEGAME_MAGIC);
```
Nordsoft91, line 166: could be quite slow sometimes. What about adding variable which will be responsible for checking that?

```c++
if(variables["mapPreviewForSaves"] && parent.tabType != ESelectionScreen::newGame)
```

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+		std::unique_ptr<CMapHeader> mapHeader = std::make_unique<CMapHeader>();
+		StartInfo * startInfo;
+		lf >> *(mapHeader) >> startInfo;
+
+		if(startInfo->campState)
+			campaignMap = startInfo->campState->getMap(*startInfo->campState->currentScenario());
+		res = ResourcePath(startInfo->fileURI, EResType::MAP);
```
Nordsoft91, line 174: what will happen if save game was sent between players and other player doesn't have campaign installed (or removed it locally)?

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	std::vector<std::shared_ptr<IImage>> images;
+	if(!campaignMap)
+		images = createMinimaps(res, Point(rect.w, rect.h));
+	else
+		images = createMinimaps(campaignMap, Point(rect.w, rect.h));
```
Nordsoft91, line 181: so you build 2 minima's, but return only one using "id". What about using createMinimapForLayer instead?

client/windows/CMapOverview.cpp
Laserlicht: Same as if you only have save and no map: no preview.

client/windows/CMapOverview.cpp
Laserlicht: Because `createMinimapForLayer` is only to create the minimap. The (optional) loading and canvas scaling is done by `createMinimaps`. I think it make sense to seperate this.

client/windows/CMapOverview.cpp
Laserlicht: possible, but why? Both layers should have the same option.

client/windows/CMapOverview.cpp
Laserlicht: if I use `w->setText()` I get a warning, that vcmi doesn't find localisation. How to fix?

client/windows/CMapOverview.cpp
Laserlicht: I don't think we need to difference between save and scenario?

client/windows/CMapOverview.cpp
Laserlicht: `createMinimapForLayer` only consist one layer. But here we need both layers.

client/windows/CMapOverview.cpp
Laserlicht: Ah, found the solution text needs to be "".

client/windows/CMapOverview.cpp
Nordsoft91: 1. could be cheating (review map, while should be hidden)
2. could be lower performance

client/windows/CMapOverview.cpp
Nordsoft91: more abstract, doesn't really matter

client/windows/CMapOverview.cpp
Nordsoft91: yes, but then you do `return images[id]`, so why not do `createMinimap(res, Point(rect.x, rect.h), id)`?

client/gui/InterfaceObjectConfigurable.cpp
```diff
@@ -56,6 +56,8 @@ InterfaceObjectConfigurable::InterfaceObjectConfigurable(int used, Point offset)
 	REGISTER_BUILDER("layout", &InterfaceObjectConfigurable::buildLayout);
 	REGISTER_BUILDER("comboBox", &InterfaceObjectConfigurable::buildComboBox);
 	REGISTER_BUILDER("textInput", &InterfaceObjectConfigurable::buildTextInput);
+	REGISTER_BUILDER("transparentFilledRectangle", &InterfaceObjectConfigurable::buildTransparentFilledRectangle);
+	REGISTER_BUILDER("textBox", &InterfaceObjectConfigurable::buildTextBox);
```
Nordsoft91, line 60: there is a doc describing all widgets: https://github.com/vcmi/vcmi/blob/develop/docs/modders/Configurable_Widgets.md#basic-widgets

Laserlicht: I will add it :)

client/windows/CMapOverview.cpp
Laserlicht: 1. not more than newGame (for loadGame there are now playerinfos shown, only the map shape)

Mods/vcmi/config/vcmi/german.json
```diff
@@ -361,4 +365,4 @@
 	"core.bonus.WATER_IMMUNITY.description": "Immun gegen alle Zauber der Wasserschule",
 	"core.bonus.WIDE_BREATH.name": "Breiter Atem",
 	"core.bonus.WIDE_BREATH.description": "Breiter Atem-Angriff (mehrere Felder)"
-}
+}
```
Alexander-Wilms, line 368: Text files should end with a newline: https://stackoverflow.com/a/729795/2278742

Mods/vcmi/config/vcmi/english.json
```diff
@@ -51,6 +51,10 @@
 	
 	"vcmi.lobby.filename" : "Filename",
```
Alexander-Wilms, line 52: Using the label "File path" would be more accurate.

Nordsoft91 (approved):

client/windows/CMapOverview.h
```diff
@@ -0,0 +1,57 @@
...
+class CFilledTexture;
+class CTextBox;
+class IImage;
+class Canvas;
+class TransparentFilledRectangle;
+enum ESelectionScreen : ui8;
+
+class CMapOverview : public CWindowObject
+{
+	class CMapOverviewWidget : public InterfaceObjectConfigurable
```
IvanSavenko, line 31: Personally prefer to avoid embedded classes, if only to avoid more complex declaration form and awkward `CMapOverview::CMapOverviewWidget::CMapOverviewWidget` syntax for methods, but not critical

client/windows/CMapOverview.h
```diff
@@ -0,0 +1,57 @@
...
+	};
+
+	std::shared_ptr<CMapOverviewWidget> widget;
+
+public:
+	const ResourcePath resource;
+	const std::string mapName;
+	const std::string fileName;
+	const std::string date;
+	const ESelectionScreen tabType;
```
IvanSavenko, line 54: Make private?

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,204 @@
...
+CMapOverview::CMapOverview(std::string mapName, std::string fileName, std::string date, ResourcePath resource, ESelectionScreen tabType)
+	: CWindowObject(BORDERED | RCLICK_POPUP), resource(resource), mapName(mapName), fileName(fileName), date(date), tabType(tabType)
+{
+
+	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
+
+	const JsonNode config(JsonPath::builtin("config/widgets/mapOverview.json"));
+	pos = Rect(0, 0, config["items"][0]["rect"]["w"].Integer(), config["items"][0]["rect"]["h"].Integer());
+
+	widget = std::make_shared<CMapOverviewWidget>(*this);
```
IvanSavenko, line 55: Why not something like this?
```suggestion
	widget = std::make_shared<CMapOverviewWidget>(*this);
    pos = widget->pos;
```

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,204 @@
...
+	CMapService mapService;
+	std::unique_ptr<CMap> map;
+	try
+	{
+		map = mapService.loadMap(resource);
+	}
+	catch (...)
+	{
+		return ret;
+	}
```
IvanSavenko, line 115: 1) Don't use "catch everything", catch either std exception or specific exceptions that you want to handler
2) Add warning to log instead of silently ignoring an error.
```suggestion
	catch (const std::exception & e)
	{
		logGlobal->warn("Failed to generate map preview! %s", e.what());
		return ret;
	}
```

client/windows/CMapOverview.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+	const JsonNode config(JsonPath::builtin("config/widgets/mapOverview.json"));
+
+	if(settings["lobby"]["mapPreview"].Bool())
+	{
+		ResourcePath res = ResourcePath(p.resource.getName(), EResType::MAP);
+		std::unique_ptr<CMap> campaignMap = nullptr;
+		if(p.tabType != ESelectionScreen::newGame && config["variables"]["mapPreviewForSaves"].Bool())
+		{
+			CLoadFile lf(*CResourceHandler::get()->getResourceName(ResourcePath(p.resource.getName(), EResType::SAVEGAME)), MINIMAL_SERIALIZATION_VERSION);
```
IvanSavenko, line 160: Is this thing fast enough, especially on low end systems, e.g. mobiles?

Laserlicht: On my 2019 android device it's not noticeable.

client/windows/CMapOverview.cpp
Laserlicht: Tested this. But don't work.
The border then think the window is 0x0px and is totally destroyed.

client/windows/CMapOverview.h
Laserlicht: Then it isn't accessible from Widget class.

client/windows/CMapOverview.cpp
Nordsoft91: Yes, because your need to take imageArea, not pos from FilledTexture

Laserlicht: `imageArea` is not accessable (protected)... But `pos` from texture is working.

Nordsoft91: then maybe introduce getImageArea()

IvanSavenko: Or finally fix Configurable UI and initialize pos as part of UI loading

Laserlicht: I think it should be okay like it is now!?

Nordsoft91: yes, let's merge

Nordsoft91: there is a conflict

Laserlicht: conflict resolved

PR #2995, opened 2023-10-01
[Docs] Update Building_Linux.md

Laserlicht:
Add `clang` and `ninja-build` to support more build setups (we have also a clang config...).


IvanSavenko:
To be fair gcc+make is "reasonable default".
If developer want to use clang and/or ninja he is likely aware of how to install those two.

Meanwhile, people who just want to compile & play are fine without clang & ninja

Laserlicht:
Yes of course. But clang is the first config appears in the list of my IDE.

For my first steps these two packages are missing to successfully start with an IDE.

Issue #1921, opened 2023-04-09
VCMI campaigns feature


Nordsoft91:
To-do list of features we want to see in VCMI campaigns:
- [ ] Refactor campaign with proper handler and individual parsers for h3m and vcmp formats (no bitmaps, special enums, etc)
- [x] Allow monsters and artifacts from mods to be kept (now limited with std::array sizes)
- [ ] Make some mechanism to encapsulate all campaign-related data (maps, maybe music, etc)
- [x] No indexes for campaign backgrounds and regions
- [ ] Introduce saving of campaign state: carryover heroes between campaigns feature
- [ ] Campaign editor as part of map editor

dydzio0614:
not sure if useful, but one thing that comes to my mind is "creature set, artifact set, building set" as starting bonuses, where starting bonus icon shows first item in set or special set icon, and by right-clicking and reading bonus description you can get listed all items

example: creature set starting bonus - 5 basilisks, 10 serpent flies, 20 lizardmen
right click says: "5 basilisks, 10 serpent flies, 20 lizardmen", basilisk icon or some special icon is used

Fabrice-TIERCELIN:
Here is a playable campaign working with the `new-pavilion` mod with several issues to discuss:
[Wars_of_the_desert.vcmp.zip](https://github.com/user-attachments/files/18072031/Wars_of_the_desert.vcmp.zip)

1. What is the size of the images to create for the region?
2. Would it be possible to mention the size of the images to create for the region in the [documentation](https://github.com/vcmi/vcmi/blob/develop/docs/modders/Campaign_Format.md)?
3. With `VCMI-branch-develop-3115894`, the region screen is broken, it shows no image and the previous screen is visible. However, pressing `Enter`, the campaign is still playable. I don't know if it's a bug or a misconfiguration of the campaign.
4. How can we create a `.smk` video?
5. Is there any other format than `.smk` supported of the prolog/epilog video?
6. How can we customize the music of the region screen.
7. During prolog, the music is not played. I don't know if it's a bug or a misconfiguration of the campaign.
8. During prolog, the speech is not played. I don't know if it's a bug or a misconfiguration of the campaign.

Laserlicht:
1 - 800x600 (see CamBk file) - other files than background: up to you
3 - never tested it with vcmp - with beta: at least chronicles is working
4 - would not suggest it (see 5.)
5 - yes WEBM with VP8/VP9 codec - suggested to use this by mods
[...]

PR #3002, opened 2023-10-03
Chronicles support

### LINKED ISSUE #2258: Chronicles campaigns
**Describe your proposal**
VCMI should allow to load the official chronicles campaigns. Currently, these cause VCMI to crash.

**Additional context**
Someone was able to convert these campaigns to Heroes III V6-format (Chronicles uses V7). https://heroes3wog.net/all-heroes-chronicles-campaigns-chapters-for-heroes-3-complete-with-hd-mod/
These files can be loaded without problems.

You can use a HEX-Editor to see the differences.

But we should able to load the original GOG-Versions of the H3C's, instead of the "hacked" ones.

Laserlicht:
fixes #2258

WIP. But basically working.

**Todo:**

- [ ] Overwrite portraits
- [ ] Overwrite voice
- [ ] Adjust coordinates in campaign_regions.json
- [ ] cleanup

grafik


Laserlicht:
@IvanSavenko Is there any possibilities of resource overwriting (Chronicles data has now the prefix "Ch" / "Ch1")?

mapOverrides is not good here... Because we have also to override the voices (part of campaign data).

launcher/aboutProject/aboutproject_moc.cpp
```diff
@@ -15,6 +15,7 @@
 
 #include "../../lib/GameConstants.h"
 #include "../../lib/VCMIDirs.h"
+#include "../../lib/filesystem/CArchiveLoader.h"
```
Alexander-Wilms, line 18: Maybe sort includes by alphabet?

IvanSavenko:
> @IvanSavenko Is there any possibilities of resource overwriting (Chronicles data has now the prefix "Ch" / "Ch1", "Ch2"...)?
> 
> mapOverrides is not good here... Because we have also to override the voices (part of campaign data).

Yes. You can try two approaches:
1) Mounting point. Check config/filesystem.json. It allows you to mount (for example) lod or snd archive from Chronicles into arbitrary location in our virtual filesystem.
2) File mapping. It allows you to map any filename to any other filename in our virtual filesystem. However this thing is unused right now. For example of its usage you can ask misiokles for an old version of wog mod (with .pac archive). Internally in c++ this is managed by `CMappedFileLoader` class

launcher/aboutProject/aboutproject_moc.cpp
```diff
@@ -83,3 +84,141 @@ void AboutProjectView::on_pushButtonBugreport_clicked()
...
 
+
+void AboutProjectView::on_pushInstallAdditional_clicked()
+{
+
+	QFileInfo file = QFileInfo(QFileDialog::getOpenFileName(this, tr("Select exe file of chronicle episode"), "", tr("Application (*.exe)")));
+	QDir sourceRoot = file.absoluteDir();
+
+	if(!sourceRoot.exists())
+		return;
```
Alexander-Wilms, line 95: Maybe use braces?

Cf. Apple's goto bug facilitated by missing braces: https://embeddedgurus.com/barr-code/2014/03/apples-gotofail-ssl-security-bug-was-easily-preventable/

launcher/aboutProject/aboutproject_moc.cpp
```diff
@@ -83,3 +84,141 @@ void AboutProjectView::on_pushButtonBugreport_clicked()
...
+	if(dirData.empty())
+	{
+		QMessageBox::critical(this, "Chronicles data not found!", "Failed to detect valid Chronicles data in chosen directory.\nPlease select directory with installed Chronicles data.");
+		return;
+	}
+
+	QDir sourceData = sourceRoot.filePath(dirData.front());
+	QDir sourceDataParent = sourceRootParent.filePath(sourceRootParent.entryList({"data"}, QDir::Filter::Dirs).front());
+
+	extractFile([chronicleId, sep](ArchiveEntry entry, CArchiveLoader archive, std::string dest){
```
Alexander-Wilms, line 130: In VCMI's code, braces seem to usually be put on the next line

launcher/aboutProject/aboutproject_moc.cpp
IvanSavenko: Personally prefer no braces for simple, single-line blocks.

Clang also provides warning for "misleading indentation", so we'll quickly catch bugs like that goto bug.

lib/campaign/CampaignState.h
```diff
@@ -57,6 +57,7 @@ class DLL_LINKAGE CampaignRegions
 	ImagePath getAvailableName(CampaignScenarioID which, int color) const;
 	ImagePath getSelectedName(CampaignScenarioID which, int color) const;
 	ImagePath getConqueredName(CampaignScenarioID which, int color) const;
+int getRegionCount() const;
```
Nordsoft91, line 60: Indent?

lib/filesystem/CArchiveLoader.cpp
```diff
@@ -197,6 +197,11 @@ std::string CArchiveLoader::getMountPoint() const
 	return mountPoint;
 }
 
+std::unordered_map<ResourcePath, ArchiveEntry> CArchiveLoader::getEntries() const
```
Nordsoft91, line 200:
```suggestion
const std::unordered_map<ResourcePath, ArchiveEntry> & CArchiveLoader::getEntries() const
```

lib/campaign/CampaignState.h
```diff
@@ -90,13 +92,15 @@ class DLL_LINKAGE CampaignHeader : public boost::noncopyable
...
 	bool formatVCMI() const;
+	CampaignVersion formatVersion() const;
 
 	std::string getDescriptionTranslated() const;
 	std::string getNameTranslated() const;
 	std::string getFilename() const;
 	std::string getModName() const;
 	std::string getEncoding() const;
 	AudioPath getMusic() const;
+	int getCampId() const;
```
Nordsoft91, line 103:
```suggestion
	int getCampaignId() const;
```

lib/campaign/CampaignState.cpp
```diff
@@ -81,13 +81,14 @@ ImagePath CampaignRegions::getNameFor(CampaignScenarioID which, int colorIndex,
 {
 	auto const & region = regions[static_cast<int>(which)];
 
-	static const std::string colors[2][8] =
+	static const std::string colors[3][8] =
 	{
+{"", "", "", "", "", "", "", ""},
```
Nordsoft91, line 86: 1. Indent?
2. Ideally should be somewhere in json file

lib/campaign/CampaignState.cpp
```diff
@@ -107,6 +108,11 @@ ImagePath CampaignRegions::getConqueredName(CampaignScenarioID which, int color)
 	return getNameFor(which, color, "Co");
 }
 
+int CampaignRegions::getRegionCount() const
```
Nordsoft91, line 111:
```suggestion
int CampaignRegions::getRegionsCount() const
```

Note: regions count not equivalent to scenarios count

lib/campaign/CampaignHandler.cpp
```diff
@@ -381,7 +381,27 @@ CampaignTravel CampaignHandler::readScenarioTravelFromJson(JsonNode & reader)
...
+}
+	else
+	{
+		std::vector<int> chroniclesCampaigns = {0, 3, 6, 9, 12, 13, 14, 17};
+
+		auto it = std::find(chroniclesCampaigns.begin(), chroniclesCampaigns.end(), campId);
+		if (it != chroniclesCampaigns.end())
+		{
+			auto index = std::distance(chroniclesCampaigns.begin(), it);
+			campId = 20 + index;
```
Nordsoft91, line 397: What is 20?

lib/campaign/CampaignHandler.cpp
```diff
@@ -381,7 +381,27 @@ CampaignTravel CampaignHandler::readScenarioTravelFromJson(JsonNode & reader)
...
 	ret.version = static_cast<CampaignVersion>(reader.readUInt32());
-	ui8 campId = reader.readUInt8() - 1;//change range of it from [1, 20] to [0, 19]
+	ui8 campId = reader.readUInt8();
+	if (ret.version != CampaignVersion::Chr)
+	{
+		campId--; //change range of it from [1, 20] to [0, 19]
+}
+	else
+	{
+		std::vector<int> chroniclesCampaigns = {0, 3, 6, 9, 12, 13, 14, 17};
```
Nordsoft91, line 391: Static?

config/filesystem.json
```diff
@@ -13,7 +13,8 @@
 			{"type" : "lod", "path" : "Data/h3abp_bm.lod"}, // Localized versions only, contains H3:AB patch data
 			{"type" : "lod", "path" : "Data/H3bitmap.lod"}, // Contains H3:SoD data (overrides H3:AB data)
 			{"type" : "lod", "path" : "Data/H3pbitma.lod"}, // Localized versions only, contains H3:SoD patch data
-			{"type" : "dir",  "path" : "Data"}
+			{"type" : "dir",  "path" : "Data"},
+			{"type" : "dir",  "path" : "HcData"} // Chronicles
```
Nordsoft91, line 17: I don't know original structure of chronicles, but will it work if player just want to use only resources from chronicles to play vcmi? It has almost all assets

client/CServerHandler.cpp
```diff
@@ -660,6 +661,9 @@ void CServerHandler::startGameplay(VCMI_LIB_WRAP_NAMESPACE(CGameState) * gameSta
 
 void CServerHandler::endGameplay(bool closeConnection, bool restart)
 {
+if(si->campState->formatVersion() == CampaignVersion::Chr)
+		CResourceHandler::removeFilesystem("data", "Hc" + std::to_string(si->campState->getCampId() - 20) + "Data");
```
Nordsoft91, line 665: Something with indentation

lib/campaign/CampaignState.cpp
Laserlicht: The colorcode with 0 characters :)

The Chronicles campaign symbols don't have colors, so no colorname in files. :)

lib/campaign/CampaignHandler.cpp
Laserlicht: The first free region. In OH3 there are 20 regions. I used here new pseudo regions after 20.

Alexander-Wilms: Maybe use a const variable instead of a magic number?

Laserlicht:
I'm not sure if I make this ready. There are a view points I don't know how to solve it clean.

It's basically working (and loading the correct loading-screen :D). But I'm not 100% happy with this.

config/filesystem.json
Laserlicht: Yes, but:
- I think every chronicles user has also h3 complete
- It's really complicated (some assets are missing; some are replaced)
- We need a renaming or separate namespaces. To allow to load multiple Chronicles (use same assets-names for other assets)

Goal should be a running H3 with the chronicles as extension, not just running one (chronicle/complete) at time.

launcher/aboutProject/aboutproject_moc.cpp
```diff
@@ -83,3 +84,133 @@ void AboutProjectView::on_pushButtonBugreport_clicked()
...
+	if(chronicleId == -1)
+	{
+		QMessageBox::critical(this, "Chronicles data not found!", "Unknown or unsupported Chronicles found.\nPlease select directory with Heroes Chronicles.");
+		return;
+	}
+
+	extractChronicles(chronicleId, sourceRoot);
+}
+
+void AboutProjectView::extractChronicles(int chronicleId, QDir sourceRoot)
```
IvanSavenko, line 113: Can we replace this (rather complex) logic with simple import of "xBitmap.lod" / "xSpriteLog" / etc and using them as it?

client/lobby/CBonusSelection.cpp
```diff
@@ -386,6 +388,12 @@ void CBonusSelection::goBack()
 
 void CBonusSelection::startMap()
 {
+	if(getCampaign()->formatVersion() == CampaignVersion::Chr)
+	{
+		auto * chFilesystem = new CFilesystemLoader("DATA/", VCMIDirs::get().userDataPath() / ("Hc" + std::to_string(getCampaign()->getCampId() - 20) + "Data"), 0);
+		CResourceHandler::addFilesystem("data", "Hc" + std::to_string(getCampaign()->getCampId() - 20) + "Data", chFilesystem);
+	}
+
```
IvanSavenko, line 396: Looks like mod functionality to be honest...
Mod that provides map/campaign and is automatically activated when player selects map/campaign from a mod

IvanSavenko: Kind of similarly to how localization mod activates only if localization matches

launcher/aboutProject/aboutproject_moc.cpp
Laserlicht: Maybe. But instead of this import logic we have to change the region logic. With this import logic it's possible to use regions for the screen (basically the 8 icons are regions).

IvanSavenko: Custom regions for campaigns are good idea for potential modders/map makers.

So changing campaign regions logic does not sounds like a bad idea.

client/lobby/CBonusSelection.cpp
Laserlicht: Yes, that's one of the points which make this unclear. One problem is also the resource priority. If I make a chronicles mod it should also be possible to replace resources of this by mod (for example for voices in other languages).

Laserlicht: But basically I think the clearest way would be a import which creates a mod of the imported files.

Laserlicht:
@IvanSavenko 

Agree with you.

But the mod has to be on campaign level, not on maps level because:
- a custom "region" has to be created (and have to override campaign's region)
- voices has to be replaced
- ideally the loadscreen from chronicles is also used

IvanSavenko:
Yeah, 1 mod = 1 chronicle campaign. So 8 mods in total, probably in form of submods inside a container "Chronicles" submod.

Right now I am just not sure whether we should give such mods already existing mod type "Maps" or add new mod type "Campaign".  Less mod types might be more clear to player & easier to categories, but might lead to confusing situations like this one :)

IvanSavenko:
>     * ideally the loadscreen from chronicles is also used

True, but not sure how to handle it...
1. Chronicles loadscreens don't have all assets needed for main menu - "single scenario" and "multiplayer" buttons are missing for example.
2. It is not clear how should we decide on which style to use. We only know that player selected Chronicles once he clicks in campaign selection and leaves main menu.

So at least for 1st iteration might be better to drop it completely unless there is clear way to use them.

Laserlicht:
@Nordsoft91 Do you plan to implement the missing points of https://github.com/vcmi/vcmi/issues/1921?

I have thought again for a better solution.

I think a good solution would also be to convert it to vcmi campaign:
- Possibility to convert in launcher to mod and vcmp (and override different things) (export of campaign to vcmp not supported yet)
- Region has to recreated in launcher with chronicles schema (already supported; region needs minor enhancement: no color)
- Adding chronicles voices to vcmp (not supported yet)
- Adding new heroes "tarnum" as mod

With this it's possible to have all campaigns at once without deactivation mods etc. And avoids overwriting of different things in gamecode (resources, heroid, etc...)

Issue #2885, opened 2023-09-20
Game crashes in altar of sacrifice


by003:
**Describe the bug**
It crashes while put artifacts from right blanks to hero's backpack.

**Did it work earlier?**
It might not work since 1.1 version.

**Screenshots**
![image](https://github.com/vcmi/vcmi/assets/127420617/fa7c3772-874e-40d1-b68f-072a22102e1f)

If crashed, figures move here as well.
![image](https://github.com/vcmi/vcmi/assets/127420617/a8110858-6dfd-4fdb-949a-bd30d75b74d5)

**Version**
 - OS: Windows
 - Version: 1.3.2

Povelitelh3:
I can confirm it. 
I went into the altar, took the artifact and put it in the next slot in my backpack (not even the altar itself). And immediately the game crashed (client).

```
[2023-Oct-03 16:40:14.772288] TRACE network - 	Made second apply on cl: struct TryMoveHero
[2023-Oct-03 16:40:14.773287] TRACE network - Received CPack of type struct HeroVisit
[2023-Oct-03 16:40:14.777287] TRACE network - 	Made first apply on cl: struct HeroVisit
[2023-Oct-03 16:40:14.777287] TRACE network - 	Applied on gs: struct HeroVisit
[2023-Oct-03 16:40:14.798287] TRACE network - 	Made second apply on cl: struct HeroVisit
[2023-Oct-03 16:40:14.799287] TRACE network - Received CPack of type struct OpenWindow
[2023-Oct-03 16:40:14.806288] TRACE network - 	Made first apply on cl: struct OpenWindow
[2023-Oct-03 16:40:14.806288] TRACE network - 	Applied on gs: struct OpenWindow
[2023-Oct-03 16:40:14.831289] TRACE network - 	Made second apply on cl: struct OpenWindow
[2023-Oct-03 16:40:14.832287] DEBUG global - totalRedraw requested!
[2023-Oct-03 16:40:14.832287] TRACE network - Received CPack of type struct HeroVisit
[2023-Oct-03 16:40:14.848287] TRACE network - 	Made first apply on cl: struct HeroVisit
[2023-Oct-03 16:40:14.848287] TRACE network - 	Applied on gs: struct HeroVisit
[2023-Oct-03 16:40:14.848287] TRACE network - 	Made second apply on cl: struct HeroVisit
[2023-Oct-03 16:40:14.849289] TRACE network - Received CPack of type struct PlayerBlocked
[...]

PR #3010, opened 2023-10-05
Artifacts placement fixed

SoundSSGood:
Fix for combined artifacts. In some cases, the component slot did not set.


Nordsoft91:
does it fix https://github.com/vcmi/vcmi/issues/2885?

SoundSSGood:
> does it fix #2885?

No, the altar needs a lot of reworking.

PR #3021, opened 2023-10-07
video loop fix and missing videos/sounds for battle result

### LINKED ISSUE #2260: Bad loop for the evading video
**Describe the bug**
The deserting video after a battle is not looped as the original version. There was two parts for the video:

1. An introduction played once
2. A loop

So that we have the feeling that the video continues but do not restart.

**Game logs**

**To Reproduce**
1. Start a battle
2. Click on 'Desert'

**Expected behavior**
The animation starts and then loops.

**Actual behavior**
The animation loops entirely.

**Did it work earlier?**

**Screenshots**

**Version**
 - OS: Windows 10
 - Version: 1.2.1

**Additional context**

### LINKED ISSUE #2910: Animation of loosing a battle is cut
sci (2)

**Version**
 - OS: [e.g. Windows 10]
 - Version: [VCMI 1.3.2]

Laserlicht:
Fixes #2260 
Fixes #2910

### DIFF
--- a/client/CVideoHandler.h
+++ b/client/CVideoHandler.h
@@ -23,7 +23,6 @@ public:
 	virtual bool nextFrame()=0;
 	virtual void show(int x, int y, SDL_Surface *dst, bool update = true)=0;
 	virtual void redraw(int x, int y, SDL_Surface *dst, bool update = true)=0; //reblits buffer
-	virtual VideoPath videoName()=0;
 	virtual bool wait()=0;
 	virtual int curFrame() const =0;
 	virtual int frameCount() const =0;
@@ -47,7 +46,6 @@ public:
 	void redraw( int x, int y, SDL_Surface *dst, bool update = true ) override {};
 	void show( int x, int y, SDL_Surface *dst, bool update = true ) override {};
 	bool nextFrame() override {return false;};
-	VideoPath videoName() override {return VideoPath();};
 	void close() override {};
 	bool wait() override {return false;};
 	bool open(const VideoPath & name, bool scale = false) override {return false;};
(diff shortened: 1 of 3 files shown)


Nordsoft91 (approved):

IvanSavenko (commented):
Perhaps instead of retrieving current video form VideoHandler store all required information in BattleResultWindow?

1. Add `getAudioPath` method to BattleResults window that will return audio file for current battle
2. Add `getIntroVIdeoPath` method that will return video path for intro video
3. Add `getLoopedVideoPath` method that will return looped video name. 

Last method should probably be called once, on BattleResultWIndow construction and stored in member - battleResult might be destroyed after construction.

Laserlicht:
Hmm. I'm not sure. The current solution uses the name already saved. Only add a function to get this.

It's more generic and maybe usable in other cases.

Do you think adding new functions would make the code more readable here? I think it's easy to understand.

IvanSavenko:
> Do you think adding new functions (and overhead) would make the code more readable here? I think it's easy to understand.

Yes. Currently constructor does *a lot* of things. over 150 lines of code. 

For readability ~50 lines for a method is probably the limit - enough to fit into screen and look over entire method at once. Anything larger than that likely deserves looking into it and splitting into smaller methods.


Hmm...
Alternatively - perhaps create a separate method that will set both video name and audio path, to separate it from constructor?


In any case please remove video name from video handler - this is global state which is something to be avoided. For example it would prevent potential playback of multiple video files at once.

Laserlicht:
@IvanSavenko Better like this?

Issue #2999, opened 2023-10-02
Antivirus alert


vodkodav:
**Describe the bug**
The antivirus issues a warning if you scan the release VCMI-1.3.2-Windows.exe or zip.
Virtotal scan results https://www.virustotal.com/gui/file/f7e59a542dc09d1c85db19714e1c6a1693881ba8f9545f7c05f9d5c732e3db1d/detection
https://www.virustotal.com/gui/file/4168ed04e01e4dd50256001bf6c4ee5ed41073821cc4cdce20d76e3f34635633

**Game logs**
The problem occurs before installation or startup.

**To Reproduce**
1. Go to https://github.com/vcmi/vcmi/releases/tag/1.3.2
2. Download VCMI-1.3.2-Windows.exe or VCMI-1.3.2-Windows.zip
3. Go to https://www.virustotal.com/
4. Upload VCMI-1.3.2-Windows.exe or VCMI-1.3.2-Windows.zip file
5. Check scanning result

**Expected behavior**
Scanning resullt: No security vendors and no sandboxes flagged this file as malicious

**Actual behavior**
Scanning resullt: 3 security vendors and no sandboxes flagged this file as malicious

**Did it work earlier?**
1.1.1 is ok, in later versions the antivirus issues a warning

**Version**
 - OS: Windows
 - Version: 1.3.2

IvanSavenko:
False positive due to heuristics.
We do a lot of _very scary_ stuff like not having a signature, requiring admin access for install or opening network connections.

In theory we could try contact developers of these antiviruses and report it but most likely too much pain.

psyhlo:
nod32 doesn't recognize it as virus while download and/or install so everything must be ok?

Nordsoft91:
Interesting, that 1.1.1 is ok but later isn't. Multiplayer and other dangerous stuff was since 1.1 or earlier. Could be first launch functionality? Also, iirc, we changed logs directory in 1.2

IvanSavenko:
> Interesting, that 1.1.1 is ok but later isn't. Multiplayer and other dangerous stuff was since 1.1 or earlier. Could be first launch functionality? Also, iirc, we changed logs directory in 1.2

That's when we changed toolchain from MXE+gcc to ubuntu+mingw

Nordsoft91:
No virus detected on develop builds

PR #3026, opened 2023-10-08
install firewall rules with installer

Laserlicht:
no more visible firewall window on windows...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -713,12 +713,22 @@ if(WIN32)
 	endif()
 	if(ENABLE_LAUNCHER)
 		set(CPACK_PACKAGE_EXECUTABLES "VCMI_launcher;VCMI")
-		set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS " CreateShortCut \\\"$DESKTOP\\\\VCMI.lnk\\\" \\\"$INSTDIR\\\\VCMI_launcher.exe\\\"")
+		set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS "
+			CreateShortCut \\\"$DESKTOP\\\\VCMI.lnk\\\" \\\"$INSTDIR\\\\VCMI_launcher.exe\\\"
+			ExecShell '' 'netsh' 'advfirewall firewall add rule name=vcmi_server dir=in action=allow program=\\\"$INSTDIR\\\\vcmi_server.exe\\\" enable=yes profile=public,private' SW_HIDE
+		")
 	else()
 		set(CPACK_PACKAGE_EXECUTABLES "VCMI_client;VCMI")
-		set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS " CreateShortCut \\\"$DESKTOP\\\\VCMI.lnk\\\" \\\"$INSTDIR\\\\VCMI_client.exe\\\"")
+		set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS "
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -713,12 +713,22 @@ if(WIN32)
 	endif()
 	if(ENABLE_LAUNCHER)
 		set(CPACK_PACKAGE_EXECUTABLES "VCMI_launcher;VCMI")
-		set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS " CreateShortCut \\\"$DESKTOP\\\\VCMI.lnk\\\" \\\"$INSTDIR\\\\VCMI_launcher.exe\\\"")
+		set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS "
+			CreateShortCut \\\"$DESKTOP\\\\VCMI.lnk\\\" \\\"$INSTDIR\\\\VCMI_launcher.exe\\\"
```
kambala-decapitator, line 717: make the exe name a param, then you can greatly simplify the if/else

Laserlicht: How to do this in cmake? simple do do a set before and use $variablename?

dydzio0614:
my take: merge as is and do not enforce people unfamiliar with cmake to refactor stuff on their own

Laserlicht:
> my take: merge as is and do not enforce people unfamiliar with cmake to refactor stuff on their own

Yes would properly the be best. :)

Can optimized directly afterwards by a cmake person. :)

CMakeLists.txt
kambala-decapitator: if you don't want to read cmake docs / some tutorial (although I think it's a good opportunity to read some basics), there's a bunch of such examples in our CML. you declare variables with `set(varname value)` and use it inside a string as `${varname}`

kambala-decapitator:
> my take: merge as is and do not enforce people unfamiliar with cmake to refactor stuff on their own

IMO such attitude makes code review senseless

CMakeLists.txt
Laserlicht: You mean like this?

kambala-decapitator: yep, looks right. Well done!

kambala-decapitator:
btw have you tested this solution?

Laserlicht:
Yes. Installing and uninstalling works. Tested with Win11.

kambala-decapitator (approved):

Nordsoft91:
Relevant to https://github.com/vcmi/vcmi/issues/2999 ?

Alexander-Wilms:
Only the MinGW .exe is flagged as malicious:

mingw exe: https://www.virustotal.com/gui/file/15f97d3c88a2d275e82a2745daba31baf9339f8ddefbed541895b6a6033b8fcd?nocache=1
mingw zip: https://www.virustotal.com/gui/file/d7e9de16cffac41ba213d1fe39329eba3db3822af55d4058f1d154324fa83900?nocache=1
msvc exe: https://www.virustotal.com/gui/file/f33ed588beed51ca30142c6e3ce9deefc295c905bfc32c9141f9f33f381b0e54?nocache=1
msvc zip: https://www.virustotal.com/gui/file/7e36c43d824cc9f511e8df43ad49c1931ec5354c75542045e5b9a6b2eff32bc4?nocache=1

Laserlicht:
> Only the MinGW .exe is flagged as malicious:

That's better than 1.3.2.

I think there are many installer modifying the firewall.

PR #3027, opened 2023-10-08
Add game pause

Nordsoft91:
While playing multiplayer it's not comfortable to play with timers without pause. Now timers are freeze if player opens in-game menu

### DIFF
--- a/CCallback.cpp
+++ b/CCallback.cpp
@@ -274,6 +274,19 @@ void CCallback::save( const std::string &fname )
 	cl->save(fname);
 }
 
+void CCallback::gamePause(bool pause)
+{
+	if(pause)
+	{
+		GamePause pack;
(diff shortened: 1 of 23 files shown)


krs0:
If I understood correctly player can force pause by opening game menu?

How about a vote? And/Or a limmited number (3) of short pauses (1-3 min)that could be disabled by opponent after some time. Not to be abused?

Nordsoft91:
> If I understood correctly player can force pause by opening game menu?

Now, it impacts local player only. Other players continue to play.

> How about a vote? And/Or a limmited number (3) of short pauses (1-3 min)that could be disabled by opponent after some time

Too difficult with unclear scenario. Timers are to restrict playing time. For corner case host can kick another player anyway

PR #3028, opened 2023-10-08
Fix campaign advance crash

### LINKED ISSUE #3024: Crash on campaign advance
**Describe the bug**
Crash on campaign advance.

**To Reproduce**
1. Start an campaign
2. Use vcmiwin cheat
3. Continue campaign

**Expected behavior**
Next campaign scenario starts.

**Actual behavior**
Game crashes during sever shutdown because connection is empty. Same happens when you try to exit active game at any stage.

**Did it work earlier?**
Yes. 1.3.2 most likely it worked.

**Version**
 - OS: Windows, msvc
 - Version: latest develop, 1.4

Nordsoft91:
Fix #3024

### DIFF
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -924,7 +924,7 @@ void CServerHandler::threadHandleConnection()
 	try
 	{
 		sendClientConnecting();
-		while(c->connected)
+		while(c && c->connected)
 		{
 			while(state == EClientState::STARTING)
 				boost::this_thread::sleep_for(boost::chrono::milliseconds(10));


client/CServerHandler.cpp
```diff
@@ -924,7 +924,7 @@ void CServerHandler::threadHandleConnection()
 	try
 	{
 		sendClientConnecting();
-		while(c->connected)
+		while(c && c->connected)
```
nullkiller, line 927: Does not it crash later?

nullkiller: void CServerHandler::endGameplay calls connection a few times

Nordsoft91: as I testes, was good. Will download build from PR later

Nordsoft91: tested on build, everything is fine. Did you reproduce crash on PR build?

PR #3037, opened 2023-10-09
show missing resources

### LINKED ISSUE #2614: Highlight missing resources for building a building
**Describe your proposal**
From mav on Discord.

Highlight in red, or somehow else the resources you have not enough of to build a building.

IMG_20230818_140317-1

Laserlicht:
fixes #2614

grafik

### DIFF
--- a/client/widgets/CComponent.cpp
+++ b/client/widgets/CComponent.cpp
@@ -42,7 +42,13 @@
 CComponent::CComponent(Etype Type, int Subtype, int Val, ESize imageSize, EFonts font):
 	perDay(false)
 {
-	init(Type, Subtype, Val, imageSize, font);
+	init(Type, Subtype, Val, imageSize, font, "");
+}
+
+CComponent::CComponent(Etype Type, int Subtype, std::string Val, ESize imageSize, EFonts font):
+	perDay(false)
+{
+	init(Type, Subtype, 0, imageSize, font, Val);
 }
 
 CComponent::CComponent(const Component & c, ESize imageSize, EFonts font)
@@ -54,7 +60,7 @@ CComponent::CComponent(const Component & c, ESize imageSize, EFonts font)
 	init((Etype)c.id, c.subtype, c.val, imageSize, font);
 }
 
-void CComponent::init(Etype Type, int Subtype, int Val, ESize imageSize, EFonts fnt)
+void CComponent::init(Etype Type, int Subtype, int Val, ESize imageSize, EFonts fnt, std::string ValText)
 {
 	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
 
@@ -63,6 +69,7 @@ void CComponent::init(Etype Type, int Subtype, int Val, ESize imageSize, EFonts 
 	compType = Type;
 	subtype = Subtype;
 	val = Val;
+	valText = ValText;
 	size = imageSize;
 	font = fnt;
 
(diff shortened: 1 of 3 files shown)


misiokles:
Just random thought. Maybe here would be better place to show missing resources?
image

Laserlicht:
Yeah, is only a simple solution. But I think it's the best.

Yes you can use the space below. But how? With text? Display the icons (up to 7) a second time? -> Make it more complicated to read...

misiokles:
Yeah, you're right.

krs0:
Is it not enough if you show with read the resources that you do not have enough of?
It will not automatically calculate how many you still need, but that should be simple math.
The proposed solution with red numbers in brackets... raises many questions. Do I have 3 out of 6? Do I need 3 more to reach the needed 6?

Nordsoft91:
Honestly, personally don't like. More numbers with unclear meaning, some braces, bright color... 
Not clear is it difference or how much player has currently? I also wondering, how gold would look like, `20000 (19950)`

Maybe dark gray color do not steal player's attention? As alternative approach - more details about missing resources on right click? Like same layout but with text: `(ore icon) Cost 5, you need 3 more ...`

Alexander-Wilms:
Maybe display resources like this if you have too few of them: 

5      3/6     5       3/6
3/6        3/6         1000

Laserlicht:
It's steals players attention? Then everything is as it should be. :D

I mean it should be immediately obvious what is missing. But, yes you have to learn what the number means. Originally wanted to add a "missing". But that is much too long. Right click is not good here (because this window can also be opened with right click).

Laserlicht:
> Maybe display resources like this if you have too few of them:
> 
> ```
> 5      3/6     5       3/6
> 3/6        3/6         1000
> ```

Could a good idea... I test it. :)

krs0:
Heroes 6 had a nice interface:
Showed missing resources in red. This could be used in VCMI (and detailed info on popup like Nordsoft mentioned as well)
Showed that you were missing resources for a building (the 3 bricks with one in red)
Showed that you could upgrade a building (Up Arrow)

image

Laserlicht:
What about @Alexander-Wilms's idea?

grafik

Honestly I like it much more than the previous approach.

Also changed the color to orange. Or is red better?

Nordsoft91:
this looks better certainly, but still I'd prefer gray

krs0:
But if something is not possible / Resource missing... then why invent new colors? Red should be fine.

misiokles:
Red colour is 'connected' with errors. I vote for orange.

Alexander-Wilms:
I prefer orange as well

krs0:
Yes but you need to be consistent also. Otherwise we end up with xMas tree :). Look at H6 example also using red.

Laserlicht:
I've also tested gray. But it really doesn't look good. Therefore I tested a custom brown:
grafik

I like the second layout. And the color? I think I prefer red (as "error").

misiokles:
It's 'new' feature, so orange is good colour to be origin of such consistencies!

Laserlicht:
Let's vote.

| Emoji        | Color           |
| ------------- |:-------------:|
| 1f389 | red |
| 1f680 | orange | 
| 1f440 | brown  | 

Brown may not look on non brown backgrounds...

kodobi:
How it looks when one have more resources then required?

Laserlicht:
> How it looks when one have more resources then required?

The leading `n /` appears only when resources missing. See wood in the example images

IvanSavenko:
What about only displaying which resources are missing, without current amount? Similar to H6 screenshot?

Laserlicht:
> What about only displaying which resources are missing, without current amount? Similar to H6 screenshot?

Yes, also possible. But I like it that you can see instantly how much is missing, which is a great improvement I think.

PR #3041, opened 2023-10-11
H3INTRO: Use INTRORIM and preserve aspect ratio

Alexander-Wilms:
Also fix glitches while resizing by using SDL_RenderClear()

https://github.com/vcmi/vcmi/assets/3226457/42df4473-7001-48c8-9aa4-053c25c69b3a


IvanSavenko:
Will it work for resolutions that are not 3x4, e.g HD/FullHD?

Alexander-Wilms:
Yes, this is at 1920x1080:

image

It also adapts when resizing:

image

image

Alexander-Wilms:
Not yet

Alexander-Wilms:
When one has selected a widescreen resolution and resizes the window to have a narrower aspect ratio, black bars appear in both axes:

image

I'll try to address that as well.

Alexander-Wilms:
Done

Laserlicht:
Tested, looks good with keeping aspect ratio and rim. :)

Alexander-Wilms:
I noticed if I resize the window a lot, something breaks:

image

Alexander-Wilms:
I tested it some more and noticed that it breaks when one changes the window size by dragging it to the left or right or top, not when one manually changes the window size.

Also, I can already reproduce the issue with the latest develop build (439ce114cbc7322f258bcc804eaf23b7fe50131f).

It seems like SDL doesn't get a usual resize event when one maximizes the window or sets it to half of the screen via the window manager.

This is all the same VCMI instance after resizing it by dragging it to the screen edges:

image

image

image

image

Maybe this issue combined with my code to use the actual window size instead of the resolution selected in the launcher results in the completely broken videos.

Alexander-Wilms:
When I keep resizing the window using latest develop (1192dbff15a2d532cf6678fcf29dcfd10bda6c82) I get the following:

```
[swscaler @ 0x55a4a3881fc0] Warning: data is not aligned! This can lead to a speed loss
Unable to create screen texture
Texture dimensions are limited to 16384x16384
terminate called after throwing an instance of 'std::runtime_error'
  what():  Unable to create screen texture
Abgebrochen (Speicherabzug geschrieben)
```

Alexander-Wilms:
I think all other issues have been resolved.

With this change, the intros now always have the correct aspect ratio and the Catherine intro has a frame.

However, if one has selected a widescreen resolution, there are two pairs of black bars.

Still, I think that looks better than them being stretched. I'm not sure if one would need to create separate renderers for the intros (always using a resolution of 800x600) and the rest of the game to avoid this.

image

image

image

image

client/CVideoHandler.cpp
```diff
@@ -636,26 +648,92 @@ bool CVideoPlayer::playVideo(int x, int y, bool stopOnKey, bool overlay)
...
+		correctedResX = aspectRatio_OH3*resY;
+		correctedResY = resY;
+
+		int offsetX = (resX-correctedResX)/2;
+		int offsetY = (resY-correctedResY)/2;
+		if(showFrame)
+		{
+			backgroundRect = CSDL_Ext::toSDL(Rect(offsetX, offsetY, correctedResX, correctedResY));
+			SDL_RenderCopy(mainRenderer, backgroundTexture, nullptr, &backgroundRect);
+			videoRect = CSDL_Ext::toSDL(Rect(	offsetX+pos.x*correctedResX/originalH3ResX,
```
Alexander-Wilms, line 729: Looks like this with a tab size of 4:

image

Alexander-Wilms:
I'll still need to add interface scaling for the intros.

Alexander-Wilms:
Tested it with various resolutions, aspect ratios and interface scaling factors and it seems to work as expected.

Before merging, I'd remove the debug output, but keep it in case the MR still has to be adapted.

Alexander-Wilms:
There are still issues on Android:

Alexander-Wilms:
Now the intros work fine on desktop and Android:

Laserlicht:
Tested. Looks good. :)

Alexander-Wilms:
Should be ready to review

client/windows/CSpellWindow.cpp
```diff
@@ -534,13 +534,13 @@ void CSpellWindow::setCurrentPage(int value)
 void CSpellWindow::turnPageLeft()
 {
 	if(settings["video"]["spellbookAnimation"].Bool() && !isBigSpellbook)
-		CCS->videoh->playSpellbookAnimation(VideoPath::builtin("PGTRNLFT.SMK"), pos.topLeft() + Point(13, 15));
+		CCS->videoh->playSpellbookAnimation(VideoPath::builtin("PGTRNLFT.SMK"), pos.topLeft() + Point(13, 14), Point(0,0));
```
Alexander-Wilms, line 537: Should `Point(0, 0)` be the default argument?

IvanSavenko (commented):
Ugh, looks like I forgot to press 'finish review' button...

client/renderSDL/ScreenHandler.h
```diff
@@ -75,22 +77,22 @@ class ScreenHandler final : public IScreenHandler
 	ScreenHandler();
 
 	/// Updates and potentially recreates target screen to match selected fullscreen status
-	void onScreenResize() final;
+	void onScreenResize() override;
```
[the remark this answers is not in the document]
Alexander-Wilms: Because SonarCloud reported this issue:

Replace the "final" specifier by "override".

"final" should not be used redundantly cpp:S1990

client/windows/CSpellWindow.cpp
IvanSavenko: What this Point(0,0) signifies and why we need to supply it here?

client/media/CVideoHandler.cpp
```diff
@@ -605,14 +686,24 @@ bool CVideoPlayer::openAndPlayVideoImpl(const VideoPath & name, const Point & po
 		rect.h = instance.dimensions.y;
 
 		if(useOverlay)
-			SDL_RenderFillRect(mainRenderer, &rect);
+		{
+			if(name.getName() == "3DOLOGO" || name.getName() == "NWCLOGO" || name.getName() == "H3INTRO"){
```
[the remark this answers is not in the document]
Alexander-Wilms: For the intros, I need to clear the entire screen instead of overlaying the video

client/media/CVideoHandler.h
```diff
@@ -14,6 +14,8 @@
 #include "../lib/Point.h"
 #include "IVideoPlayer.h"
 
+#include <SDL_render.h>
+
```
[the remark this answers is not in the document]
Alexander-Wilms: If I replace the include with `struct SDL_Rect;`, I get

```
[build] /home/alexander/Technik/Development/vcmi/vcmi-git/client/media/CVideoHandler.cpp:180:24: error: member access into incomplete type 'SDL_Surface'
[build]   180 |                 dimensions.x = screen->w;
[build]       |                                      ^
```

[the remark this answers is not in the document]
IvanSavenko: Forward declaration should be placed in header
While actual include should be moved to .cpp file

client/windows/CSpellWindow.cpp
```diff
@@ -524,13 +524,13 @@ void CSpellWindow::setCurrentPage(int value)
 void CSpellWindow::turnPageLeft()
 {
 	if(settings["video"]["spellbookAnimation"].Bool() && !isBigSpellbook)
-		CCS->videoh->playSpellbookAnimation(VideoPath::builtin("PGTRNLFT.SMK"), pos.topLeft() + Point(13, 15));
+		CCS->videoh->playSpellbookAnimation(VideoPath::builtin("PGTRNLFT.SMK"), pos.topLeft() + Point(13, 14));
```
Alexander-Wilms, line 527: Compared screenshots, and with these coordinate, the image and the video match.

Alexander-Wilms:
Sould be ready to merge I think.

client/windows/CSpellWindow.cpp
Alexander-Wilms: The `preferredLogicalResolution`, which is used by `getVideoAndBackgroundRects()` if an intro is playing.

Issue #2929, opened 2023-09-23
There is no possibility of checking the day and number of towns/players in the box in down right corner


HarkUhm:
**Describe the bug**
OH shows these informations when clicked on castle or hero box in down right corner:
![affasd](https://github.com/vcmi/vcmi/assets/141812373/8816613a-3637-42a3-bbfe-e8dbbe5cbd3e)
![afafas](https://github.com/vcmi/vcmi/assets/141812373/d59b83ae-21b1-44fc-b0f0-35b8cd292020)
VCMI doesn't provide this info
**Version**
 - OS: [Windows 10]
 - Version: [VCMI 1.3.2]

dydzio0614:
it is shown if creature management in this window is disabled, and since i do not think we aim for recreating "hd mod specific stuff" exactly as they are, i leave it to somebody else to decide if this should be treated as "bug" or "enhancement"

rilian-la-te:
@dydzio0614 it should be allow to have a setting to switch quick army management off.

PR #3045, opened 2023-10-12
various interface shortcuts

Laserlicht:
On touch devices it's not possible to use hotkeys.

Because of that. And for people (like me) who search the correct building. Some improvements:

grafik
(additional to existing)
- town icon opens tavern
- gold opens marketplace

grafik
- top -> same as town screen
- bottom -> last setting is now saved (for next menu open; same as HD mod)

grafik
(only if quick army setting active)
- same as town screen
- except: no tavern (we need some space to switch between the different info windows

### DIFF
--- a/client/windows/GUIClasses.cpp
+++ b/client/windows/GUIClasses.cpp
@@ -489,7 +489,7 @@ CTavernWindow::CTavernWindow(const CGObjectInstance * TavernObj, const std::func
 		recruit->addHoverText(CButton::NORMAL, boost::str(boost::format(CGI->generaltexth->tavernInfo[1]) % LOCPLINT->cb->howManyHeroes(false)));
 		recruit->block(true);
 	}
-	else if(LOCPLINT->castleInt && LOCPLINT->castleInt->town->visitingHero)
+	else if((LOCPLINT->castleInt && LOCPLINT->castleInt->town->visitingHero) || (dynamic_cast<const CGTownInstance *>(TavernObj) && dynamic_cast<const CGTownInstance *>(TavernObj)->visitingHero))
 	{
 		recruit->addHoverText(CButton::NORMAL, CGI->generaltexth->tavernInfo[2]); //Cannot recruit. You already have a Hero in this town.
 		recruit->block(true);
@@ -501,6 +501,8 @@ CTavernWindow::CTavernWindow(const CGObjectInstance * TavernObj, const std::func
 	}
 	if(LOCPLINT->castleInt)
 		CCS->videoh->open(LOCPLINT->castleInt->town->town->clientInfo.tavernVideo);
+	else if(dynamic_cast<const CGTownInstance *>(TavernObj))
+		CCS->videoh->open(dynamic_cast<const CGTownInstance *>(TavernObj)->town->clientInfo.tavernVideo);
 	else
 		CCS->videoh->open(VideoPath::builtin("TAVERN.BIK"));
 }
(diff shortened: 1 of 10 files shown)


krs0:
_no tavern (we need some space to switch between the different info windows_

So why not be consequent and allow tavern here as well?

Switching info windows could be done in another way.

Firstly, why would someone want to switch the windows? What extra information does he get?

client/windows/GUIClasses.cpp
```diff
@@ -501,6 +501,8 @@ CTavernWindow::CTavernWindow(const CGObjectInstance * TavernObj, const std::func
 	}
 	if(LOCPLINT->castleInt)
 		CCS->videoh->open(LOCPLINT->castleInt->town->town->clientInfo.tavernVideo);
+	else if(dynamic_cast<const CGTownInstance *>(TavernObj))
+		CCS->videoh->open(dynamic_cast<const CGTownInstance *>(TavernObj)->town->clientInfo.tavernVideo);
```
Nordsoft91, line 505:
```suggestion
	else if(const auto * townObj = dynamic_cast<const CGTownInstance *>(TavernObj))
		CCS->videoh->open(townObj->town->clientInfo.tavernVideo);
```

Laserlicht:
> So why not be consequent and allow tavern here as well?
> 
> Switching info windows could be done in another way.
> 
> Firstly, why would someone want to switch the windows? What extra information does he get?

Because I don't want to break this OH3 functionality: https://github.com/vcmi/vcmi/issues/2929
And if I use the town icon there is (on mobile) basically no space to switch between the screens.

krs0:
But this is not OH3 functionality. Oh3 had the whole square clickable and no InfoPanel like VCMI.
If someone wants OH3 they can disable InfoPanel.
Right now we have a compromise mix of both with reserving a square somwhere to click.

Laserlicht:
@krs0 Added it.

Tested it also on smartphone. It's still possible to switch (click name) -> great! :)

Nordsoft91:
minor bug:
 - open kindom view
 - build fort/townhall or any creatures building
 - expected: icons in kingdom view, gold income, unlocked creatures, etc are updated
 - actual: information and icons are not updated, need to reopen kindom view to see changes
 
 another (maybe even was earlier):
 - "there is no creatures to recruit" message is not painted in player color

Laserlicht:
> minor bug:
 Yes - but no idea how to fix.

Nordsoft91:
interface update happens in CPlayerInterface::buildChanged(const CGTownInstance *town, BuildingID buildingID, int what)

Laserlicht:
fixed

client/windows/GUIClasses.cpp
```diff
@@ -489,7 +489,7 @@ CTavernWindow::CTavernWindow(const CGObjectInstance * TavernObj, const std::func
 		recruit->addHoverText(CButton::NORMAL, boost::str(boost::format(CGI->generaltexth->tavernInfo[1]) % LOCPLINT->cb->howManyHeroes(false)));
 		recruit->block(true);
 	}
-	else if(LOCPLINT->castleInt && LOCPLINT->castleInt->town->visitingHero)
+	else if((LOCPLINT->castleInt && LOCPLINT->castleInt->town->visitingHero) || (dynamic_cast<const CGTownInstance *>(TavernObj) && dynamic_cast<const CGTownInstance *>(TavernObj)->visitingHero))
```
IvanSavenko, line 492: Redundant?
```suggestion
	else if(dynamic_cast<const CGTownInstance *>(TavernObj) && dynamic_cast<const CGTownInstance *>(TavernObj)->visitingHero))
```

IvanSavenko (commented):
Can't say that I am fan of hidden changes that are not obvious to player.
Can we add this information somewhere in player docs at least?

PR #3049, opened 2023-10-13
Fixed morale and luck chance lookup.

Warzyw647:
Good and bad Morale and Luck skipped the first value from the settings. It was an off-by-1 error when changing good/bad Luck/Morale counts (starting from +-1) into array indexes (starting from 0).


IvanSavenko (approved):

Warzyw647:
It's a simple fix that should just work, but I didn't recompile on my machine yet to check because compilation takes a while.

PR #3053, opened 2023-10-14
Improve metainfo file

Alexander-Wilms:
Since the Slack chat seems to focus on development and software centers usually have a separate section to show license info.

E.g.:

image


krs0:
There are some annoying warnings on CI about slack plugin. Would that fall into this cleanup category?

Alexander-Wilms:
I'm not sure what you mean.

I know there are a bunch of warnings because the secret is not available, but would you prefer to remove the Slack integration or fix it?

Also, that's not something I can do.

krs0:
Yeah it was about the missing secret.
@someone else knows how to fix it

IvanSavenko (approved):

Nordsoft91:
@ArseniyShestakov should probably know slack secret

PR #3062, opened 2023-10-15
sort town/hero feature

Laserlicht:
use radial wheel to sort towns

grafik
grafik


Nordsoft91:
I suspect shouldn't impact town portal?

Laserlicht:
Tested: Yes correct - doesn't impact town portal.

But: Because the town portal currently using only text representation, I think it's okay for now.

Something we can do with the extension of the portal window.

client/widgets/RadialMenu.h
```diff
@@ -60,8 +67,10 @@ class RadialMenu : public CIntObject
 	void addItem(const Point & offset, bool enabled, const std::string & path, const std::string & hoverText, const std::function<void()> & callback);
 
 	std::shared_ptr<RadialMenuItem> findNearestItem(const Point & cursorPosition) const;
+
+	bool altLayout;
```
Nordsoft91, line 71:
```suggestion
	bool alternativeLayout;
```

Nordsoft91:
doesn't work on desktop systems?

IvanSavenko:
> doesn't work on desktop systems?

Should work via mouse wheel press + drag

Laserlicht:
Yes with middle mouse and drag

client/adventureMap/CList.h
```diff
@@ -150,8 +150,9 @@ class CTownList	: public CList
 	class CTownItem : public CListItem
 	{
 		std::shared_ptr<CAnimImage> picture;
+		CTownList *parentList;
 	public:
-		const CGTownInstance * const town;
+		int townPos;
```
Nordsoft91, line 155: it's not Pos (usually townPos in other places is town map position, like int3), but index.
The problem with indices is that is's possible to get out-of-range exception (and access violation in your code, as you use operator [] without checking index).

What's wrong with pointer? You can get current position of stored pointer indside array using vstd::find_pos and internally get next/prev town pointer. Less changes -> lower risk then

Laserlicht: The problem with pointer was: That a pointer wasn't valid anymore and I got segmentation fault. With the index the problem no longer exists.

Laserlicht: You can test build before `fix segmentation fault; update townlist on adventuremap`. Move towns in town window and then click onto it to change active town. Sometime it crashes. After the change everything is fine.

Nordsoft91: then maybe use ObjectInstanceId? Or rename to townIndex?

Laserlicht: can rename it

client/adventureMap/CList.cpp
```diff
@@ -348,22 +356,46 @@ void CTownList::CTownItem::update()
...
+
+void CTownList::CTownItem::gesture(bool on, const Point & initialPosition, const Point & finalPosition)
+{
+	if(!on)
+		return;
+
+	if(!LOCPLINT->localState->getOwnedTowns()[townPos])
+		return;
+
+	const std::vector<const CGTownInstance *> towns = LOCPLINT->localState->getOwnedTowns();
```
Nordsoft91, line 380:
```suggestion
	const std::vector<const CGTownInstance *> towns = LOCPLINT->localState->getOwnedTowns();
	
	if(townPos < 0 || townPos > towns.size() - 1 || !towns[townPos])
		return;
```

krs0:
Never thought to try that middle mouse press. Tried left drag, Ctrl, Alt, even right drag... nothing :)

Why not the more convenient Left Mouse click + drag?

Laserlicht:
Because it was the easiest way to implement. And for our biggest player base it is easy to figure out.

And even on PC. I think the middle mouse button as gesture key is not bad.

kodobi:
Middle mouse button is terrible for everything except scrolling.

Laserlicht:
> Middle mouse button is terrible for everything except scrolling. 

I thought so too at first. But I have to say that I find it quite good for gestures now (which you only need now and then).

IvanSavenko:
> Middle mouse button is terrible for everything except scrolling.

Keep in mind that this feature was initially made for Android / touchscreens.
Mouse wheel access was added for ease of testing on PC.

Another problem is that we don't really have other free controls - both left and right clicks are already reserved for other actions.

krs0:
Isnt dragging as 'simple' as looking on click pressed for a mouse move?

From user perspective is harder to use mouse wheel.

Laserlicht:
> From user perspective is harder to use mouse wheel.

Yes maybe. But I like it more than hot keys. The radial wheel is basically implemented as hot key replacement.

As far as I know in HD this feature is implemented with hot keys.

Nordsoft91:
On mac touchpad it's impossible to use this feature

Laserlicht:
> On mac touchpad it's impossible to use this feature

And with mouse? I never used a Mac...

client/widgets/RadialMenu.cpp
```diff
@@ -21,14 +21,14 @@
...
+RadialMenuItem::RadialMenuItem(const std::string & imageName, const std::string & hoverText, const std::function<void()> & callback, bool alternativeLayout)
 	: callback(callback)
 	, hoverText(hoverText)
 {
 	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
 
-	inactiveImage = std::make_shared<CPicture>(ImagePath::builtin("radialMenu/itemInactive"), Point(0, 0));
-	selectedImage = std::make_shared<CPicture>(ImagePath::builtin("radialMenu/itemEmpty"), Point(0, 0));
+	inactiveImage = std::make_shared<CPicture>(ImagePath::builtin(alternativeLayout ? "radialMenu/itemInactiveAlt" : "radialMenu/itemInactive"), Point(0, 0));
+	selectedImage = std::make_shared<CPicture>(ImagePath::builtin(alternativeLayout ? "radialMenu/itemEmptyAlt" : "radialMenu/itemEmpty"), Point(0, 0));
```
IvanSavenko, line 31: Why different layout is needed?

Laserlicht: It's a 90 degrees rotated layout.

With this it's possible to have two vertical buttons one below the other.

IvanSavenko: But it is not really vertical from screenshots:
grafik

IvanSavenko: And does non-rotated version actually looks bad?

IvanSavenko:
Also - why only towns? Perhaps also allow this functionality for heroes?

client/widgets/RadialMenu.cpp
Laserlicht: But the down button is exactly under the up button (without space). That's not possible with non rotated layout.

Laserlicht:
Makes not much sense for heroes I think. Because they are reordered by game itself. If you have hero in town for example.

IvanSavenko:
True, but it often undesired - I might want to keep my main hero on top of the list for convenience.
I don't think that we need to lock heroes ordering, but option to reorder them manually is probably a good idea.

Also - what about adding "move to top" / "move to bottom" options in addition to yours "move by one" options?

Laserlicht:
grafik

client/adventureMap/CList.cpp
```diff
@@ -262,6 +267,43 @@ std::string CHeroList::CHeroItem::getHoverText()
 	return boost::str(boost::format(CGI->generaltexth->allTexts[15]) % hero->getNameTranslated() % hero->type->heroClass->getNameTranslated());
 }
 
+void CHeroList::CHeroItem::gesture(bool on, const Point & initialPosition, const Point & finalPosition)
```
Nordsoft91, line 270: Looks like copy-paste from towns list. Maybe introducing base class common for both with accepting swap function as an argument?

Laserlicht: Yes, may be possible with some work I think.
But I also think it would make sense to keep them separate. We have two slots left for the radial wheel. And we should keep the possibility to add different things for town and heroes.

PR #3063, opened 2023-10-15
Use plural form in identifiers to distinguish them for single case

krs0:
EG: Differentiate between modSettings and modsSettings


IvanSavenko:
Any need for this change?
This will also enable all mods for all players (due to renamed config file).

Also don't see changes in Android launcher

krs0:
Hmmm.... maybe is more complicated than I thought :)

Noted Android launcher needed changes.

The intention for now is to make code (that is well organized to start with) a little more clearer if possible. So future work on this part to be more easy.

As for future work: I will try to make launcher support premade mod configurations as well as multiple mod configurations.
EG: I want to distribute a list of mods (30) and do not want to make the player click them one by one from a list.
EG2: When I play sometimes I want to play something true to OH3... one set of mods active and sometimes to try something whacky like vegetable town => another set of mods active. Would be nice to be able to save the configurations and switch on the fly.

luk3ZPL:
This idea reminds me playlists from foobar - different playlists for different albums. Good idea.

launcher/modManager/cmodmanager.cpp
```diff
@@ -56,17 +56,23 @@ CModManager::CModManager(CModList * modList)
...
+	return modSettingsName;
 }
 
-void CModManager::loadModSettings()
+QString CModManager::getModSettingsPath()
 {
-	modSettings = JsonUtils::JsonFromFile(settingsPath()).toMap();
+	std::string modSettingsFileName = modSettingsName.toStdString() + ".json";
+	return pathToQString(VCMIDirs::get().userConfigPath() / "modSettings" / modSettingsFileName);
+}
```
IvanSavenko, line 71: Why not simply store all mod presets in one json file?

krs0: This is what you have suggested on Discord, but I had more questions, like how to distribute such a list?

So the easiest way I could think of, was to just add it to this folder and discover modSettings in it automatically.

IvanSavenko: This is why I said that distributing presets with mods needs more discussions.
Our modding system *always* references all files instead of scanning folders.

This leads to way more complex code and way more potential fault points.

Like what to do if player drops new preset while Launcher is running? What if player deletes active preset?

krs0: I do not know before hand how VCMI handles things. So I make a prototype. Kindly please take some time and set other directions, if what I am trying to do does not align with the vision for VCMI. (with the caveat that it may well be outside of my powers).

What I have:
For now I can select multiple presets, that the user can add and select via a combobox like in the video below.
All presets are stored in <user_config>/modSettings/ folder.

Future:
modSettings are distributed as part of mod (in root folder?) discovered at launcher start and copied?!? to <user_config>/modSettings/. From this folder they end up in combobox for the user to select.

https://github.com/vcmi/vcmi/assets/15194321/c9369b26-9c17-4668-9f33-118b5f65d6a7

krs0: _Like what to do if player drops new preset while Launcher is running?_

Won't register.

_What if player deletes active preset?_

Nothing will happen. Preset will be overwritten with memory stuff.

_What if player drops preset into read-only location and attempts to edit it in Launcher_

Not possible User_config is not readonly as it is used for other stuff.

IvanSavenko: - Creation of new preset is extremely unclear
- No way to delete mod preset
- This design will not work on iOS with limited horizontal space (it is already too cramped)
- I see literally zero need to spread this across multiple files. Why not store all data in one file and drop all file shuffling?
- From what I see, your change will reset all mod settings for all players

> modSettings are distributed as part of mod (in root folder?) discovered at launcher start and copied?!? 

Please no. Not yet sure how to do this but definitely not like this
1) Literally nobody has ever asked for this
2) VCMI never makes any assumptions of mod file structure
3) VCMI always requires all mod content to be referenced in mod.json

krs0: _Creation of new preset is extremely unclear
No way to delete mod preset
This design will not work on iOS with limited horizontal space (it is already too cramped)_

This with combobox is still first iteration! And for iOs and Android! will look for a better solution. 

_From what I see, your change will reset all mod settings for all players_

It will. But it will be quite easy to copy old mod settings as well if really needed. But then if reselecting mods is a problem, then it speaks in favor of this mods need.

_VCMI never makes any assumptions of mod file structure
VCMI always requires all mod content to be referenced in mod.json_

Then in mod it will be done like that. No problems there.

_Literally nobody has ever asked for this_

I feel this is the biggest issue, a short poll on Discord can shed some light onto it.

From all the reports on broken saves I see that there is a need of some kind of a list of mods from which players should be able to:

- install automatically the mods and needed dependencies
- enable the specified mods
- optionally disable the mods not in list

PR #3064, opened 2023-10-15
Type-safe bonus system

IvanSavenko:
List of changes:
- added VariantIdentifier class that stores one of several possible identifer types in std::variant
- bonus subtype and bonus source now use VariantIdentifier class instead of raw int32_t
- added string identifiers for bonuses like MOVEMENT instead of 0/1/2/etc
- updated all bonuses that were using integer as subtype to use string type
- updated docs to account for chage & clarified some incorrect bonus descriptions

Long-term goal is to save all identifiers as strings, to allow loading saves made with slightly different set of mods / slightly different mod versions / different mod load order. However will be done separately to avoid this PR growing too much. And I am sure that we have other places that still store identifiers as plain numbers which needs to be tracked down first.

Mod compatibility: old mods *should* work, but likely generate significant amount of warnings, since bonus subtype can no longer be a number. Drawback is that we have two bonuses that have creature & spell level as subtypes. To avoid making exception for them, these bonuses will now use subtypes like "creatureLevel2" and "spellLevel5".

TODO: 
- document bonuses with spell/creature level subtypes.


rilian-la-te:
1. Why strings? Maybe some int-based new type, but not string?
2. Can you check unit tests too?
3. Also selector-checking and selector-parsing should be checked too, it can be broken

IvanSavenko:
Strings are only used for loading from json (and in future - for saves/network serialization).
Main change is that internally game now uses `std::variant<SpellID, CreatureID, ... >` instead of plain integer.

Since implicit conversions were banned, I was able to track all sources of bonus subtypes, like limiters, so don't expect any issues with that.

lib/BasicTypes.cpp
```diff
@@ -54,7 +54,7 @@ int AFactionMember::getAttack(bool ranged) const
 {
 	const std::string cachingStr = "type_PRIMARY_SKILLs_ATTACK";
 
-	static const auto selector = Selector::typeSubtype(BonusType::PRIMARY_SKILL, static_cast<int>(PrimarySkill::ATTACK));
+	static const auto selector = Selector::typeSubtype(BonusType::PRIMARY_SKILL, TBonusSubtype(PrimarySkill::ATTACK));
```
Nordsoft91, line 57: `TbonusSubtype` -> `BonusSubtype`?

lib/JsonNode.cpp
```diff
@@ -486,13 +486,235 @@ std::string JsonNode::toJson(bool compact) const
 
 ///JsonUtils
 
-void JsonUtils::parseTypedBonusShort(const JsonVector & source, const std::shared_ptr<Bonus> & dest)
+static void loadBonusSubtype(TBonusSubtype & subtype, BonusType type, const JsonNode & node)
```
Nordsoft91, line 489: maybe move those large-parsing blocks from JsonNode.cpp to somewhere where bonus-related code is located?

lib/campaign/CampaignConstants.h
```diff
@@ -11,6 +11,8 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
+#include "../constants/EntityIdentifiers.h"
```
Nordsoft91, line 14: 1. Why is needed?
2. Perhaps, should be outside of VCMI_LIB namespace?

lib/JsonNode.cpp
IvanSavenko: True, but probably not this time.
Our json parser files contain quite a lot of code largely not related to well, json parsing.
And not sure where such code should go. Bonus class itself? Bonus type handler? Or some separate entity? This code is used not just for bonus itself but also for limiters. There are also weird methods like "load town building bonus" one. Why that even a thing and not part of town handler?

lib/BasicTypes.cpp
IvanSavenko: BonusSubtypeID is probably more in line with our naming scheme. Will rename. Same with TBonusSourceID -> BonusSourceID

PR #3069, opened 2023-10-17
ArtifactLocation now use ID for artHolder identification 

SoundSSGood:
Identification by instance has been replaced by identification by ID. To unify and simplify the codebase.

### DIFF
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -693,7 +693,7 @@ void CArtHandler::makeItCommanderArt(CArtifact * a, bool onlyCommander)
 		a->possibleSlots[ArtBearer::HERO].clear();
 		a->possibleSlots[ArtBearer::CREATURE].clear();
 	}
-	for(auto & slot : ArtifactUtils::commanderSlots())
+	for(const auto & slot : ArtifactUtils::commanderSlots())
 		a->possibleSlots[ArtBearer::COMMANDER].push_back(ArtifactPosition(slot));
 }
 
@@ -1080,7 +1080,7 @@ void CArtifactSet::serializeJsonArtifacts(JsonSerializeFormat & handler, const s
 
 void CArtifactSet::serializeJsonHero(JsonSerializeFormat & handler, CMap * map)
 {
-	for(auto & slot : ArtifactUtils::allWornSlots())
+	for(const auto & slot : ArtifactUtils::allWornSlots())
 	{
 		serializeJsonSlot(handler, slot, map);
 	}
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -693,8 +693,8 @@ void CArtHandler::makeItCommanderArt(CArtifact * a, bool onlyCommander)
 		a->possibleSlots[ArtBearer::HERO].clear();
 		a->possibleSlots[ArtBearer::CREATURE].clear();
 	}
-	for (int i = ArtifactPosition::COMMANDER1; i <= ArtifactPosition::COMMANDER6; ++i)
-		a->possibleSlots[ArtBearer::COMMANDER].push_back(ArtifactPosition(i));
+	for(auto & slot : ArtifactUtils::commanderSlots())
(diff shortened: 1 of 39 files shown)


server/queries/MapQueries.cpp
```diff
@@ -111,12 +111,12 @@ bool CGarrisonDialogQuery::blocksPack(const CPack * pack) const
 
 	if(auto arts = dynamic_ptr_cast<ExchangeArtifacts>(pack))
 	{
-		if(auto id1 = std::visit(GetEngagedHeroIds(), arts->src.artHolder))
-			if(!vstd::contains(ourIds, *id1))
+		if(auto id1 = arts->src.artHolder)
```
IvanSavenko, line 114: I think GetEngagedHeroIds struct is no longer used and can be removed?

server/battles/BattleResultProcessor.cpp
```diff
@@ -393,11 +394,12 @@ void BattleResultProcessor::endBattleConfirm(const CBattleInfoCallback & battle)
 		for (auto armySlot : battle.battleGetArmyObject(loser)->stacks)
 		{
 			auto artifactsWorn = armySlot.second->artifactsWorn;
-			for (auto artSlot : artifactsWorn)
+			for(auto & artSlot : artifactsWorn)
```
IvanSavenko, line 397: (if applicable)
```suggestion
			for(const auto & artSlot : artifactsWorn)
```

lib/NetPacks.h
```diff
@@ -1087,8 +1087,8 @@ struct DLL_LINKAGE BulkMoveArtifacts : CArtifactOperationPack
 
 		LinkedSlots() = default;
 		LinkedSlots(const ArtifactPosition & srcPos, const ArtifactPosition & dstPos)
-			: srcPos(srcPos)
-			, dstPos(dstPos)
+			: srcPos(std::move(srcPos))
+			, dstPos(std::move(dstPos))
```
IvanSavenko, line 1091: I might be wrong, but I think std::move here is pointless since parameters are passed as const reference.
You should use either:
`Foo(const Bar & bar) : bar(bar)`
`Foo(Bar bar) : bar(std::move(bar))`

lib/CArtHandler.cpp
```diff
@@ -1080,9 +1080,9 @@ void CArtifactSet::serializeJsonArtifacts(JsonSerializeFormat & handler, const s
 
 void CArtifactSet::serializeJsonHero(JsonSerializeFormat & handler, CMap * map)
 {
-	for(ArtifactPosition ap = ArtifactPosition::HEAD; ap < ArtifactPosition::AFTER_LAST; ap.advance(1))
+	for(auto & slot : ArtifactUtils::allWornSlots())
```
IvanSavenko, line 1083:
```suggestion
	for(const auto & slot : ArtifactUtils::allWornSlots())
```

lib/CArtHandler.cpp
```diff
@@ -693,8 +693,8 @@ void CArtHandler::makeItCommanderArt(CArtifact * a, bool onlyCommander)
 		a->possibleSlots[ArtBearer::HERO].clear();
 		a->possibleSlots[ArtBearer::CREATURE].clear();
 	}
-	for (int i = ArtifactPosition::COMMANDER1; i <= ArtifactPosition::COMMANDER6; ++i)
-		a->possibleSlots[ArtBearer::COMMANDER].push_back(ArtifactPosition(i));
+	for(auto & slot : ArtifactUtils::commanderSlots())
```
IvanSavenko, line 696:
```suggestion
	for(const auto & slot : ArtifactUtils::commanderSlots())
```

server/queries/MapQueries.cpp
SoundSSGood: sure

lib/NetPacks.h
SoundSSGood: Yes you are right. Removed

IvanSavenko (approved):

PR #3086, opened 2023-10-21
show random maps

### LINKED ISSUE #3048: Add "Show random maps" functionality
**Describe your proposal**
The button Show random maps has no functionality in VCMI.
It should show a list with all random maps generated. In HotA each random map is saved in random_maps folder.

image
image

Laserlicht:
fixes #3048

### DIFF
--- a/lib/gameState/CGameState.cpp
+++ b/lib/gameState/CGameState.cpp
@@ -566,16 +566,19 @@ void CGameState::initNewGame(const IMapService * mapService, bool allowSavingRan
 		{
 			try
 			{
-				auto path = VCMIDirs::get().userCachePath() / "RandomMaps";
+				auto path = VCMIDirs::get().userDataPath() / "Maps" / "RandomMaps";
 				boost::filesystem::create_directories(path);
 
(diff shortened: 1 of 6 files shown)


lib/gameState/CGameState.cpp
```diff
@@ -566,17 +566,20 @@ void CGameState::initNewGame(const IMapService * mapService, bool allowSavingRan
 		{
 			try
 			{
-				auto path = VCMIDirs::get().userCachePath() / "RandomMaps";
+				auto path = VCMIDirs::get().userDataPath() / "Maps" / "RandomMaps";
```
IvanSavenko, line 569: So I guess we now save random maps inside "Maps" directory.
Not against the change, considering that we allow subdirectories now.

IvanSavenko (approved):

lib/gameState/CGameState.cpp
Laserlicht: Yes, exactly. But this folder is hidden and only shown, when using the "show random" button (then all other folders/maps are hidden).

PR #3089, opened 2023-10-21
Fix crash on game load

IvanSavenko:
Looks that due to recent changes game now crashes on save loading.

Cause seems to be JsonNode deserialization. Probably - issue with union usage alongside with copy and/or move constructors.

After looking into it I haven't noticed any obvious bugs, so decided to drop union usage instead and use std::variant instead which is definitely more well-tested than our class.

### DIFF
--- a/lib/CConfigHandler.cpp
+++ b/lib/CConfigHandler.cpp
@@ -61,7 +61,7 @@ void SettingsStorage::init(const std::string & dataFilename, const std::string &
 
 	JsonPath confName = JsonPath::builtin(dataFilename);
 
-	JsonUtils::assembleFromFiles(confName.getOriginalName()).swap(config);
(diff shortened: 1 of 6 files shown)


Laserlicht:
@IvanSavenko 

Is it possible that this PR breaks the persistentStorage?
But interestingly only on android (in my case 14), not on PC. For all newer PR's VCMI crashes when adding a highscore. Or for our new touchscreen help after reading the Bool values out of it.

IvanSavenko:
In theory - possible. But not sure why only on Android...

Perhaps persistent storage file does not exists there? Can you check whether crash happens on PC if you don't have that json present?

Laserlicht:
On PC I never had this problem. Started the touch introduction PR in a clean VM (never seen VCMI). And there was never a crash. But android CI artifact crash.

PR #3092, opened 2023-10-21
Better support for compatibility mods

### LINKED ISSUE #3068: Automatic loading of compatibility mods
**Describe your proposal**
Some mods need submods to fix some compatibility issues. These submods need to be enabled only in certain scenarios EG: Only when both mods are enabled.

@IvanSavenko proposed to add a new type of mod: Compatibility. and enable this mod only when dependencies are resolved for it.

ToW and HotA had conflicts because of chaninging same building in town screen.
_For example, right now we have `Hota balance compatibility patch` submod in `ToW`.
What we need to do is:

- make this mod depend on both tow and `HotA balance` submod
- hide this mod from UI and enable automatically when both tow and `HotA balance` are enabled_

### LINKED ISSUE #3085: New mod type is needed: Maps
Requested on Discord by Laserlicht:

_Can someone add the mod type "Maps"? There are already three of them. 🙂_

IvanSavenko:
- Added mod type "Compatibility". Mods of this type are hidden in Launcher and are always active if they are compatible (e.g. no missing dependencies). Intended to be used for compatibility patches between mods
- Mods of type "Translation" are now hidden except for mods that are installed or mods with same language as currently selected - to reduce our rather long list of mods in repository.
- Added mod of type "Maps" since it was requested and I was changing related code.
- Fixes #3085
- Fixes #3068

### DIFF
--- a/config/schemas/mod.json
+++ b/config/schemas/mod.json
@@ -55,7 +55,7 @@
 		},
 		"modType" : {
 			"type" : "string",
-			"enum" : [ "Translation", "Town", "Test", "Templates", "Spells", "Music", "Sounds", "Skills", "Other", "Objects", "Mechanics", "Interface", "Heroes", "Graphical", "Expansion", "Creatures", "Artifacts", "AI" ],
+			"enum" : [ "Translation", "Town", "Test", "Templates", "Spells", "Music", "Maps", "Sounds", "Skills", "Other", "Objects", "Mechanics", "Interface", "Heroes", "Graphical", "Expansion", "Creatures", "Compatibility", "Artifacts", "AI" ],
 			"description" : "Type of mod, e.g. Town, Artifacts, Graphical."
 		},
 		"author" : {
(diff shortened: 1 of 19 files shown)


Laserlicht:
@IvanSavenko 

For example the german translation have a submod for chronicles translation which is disabled by default. Could be enabled if you download and copy converted campaigns into maps folder.

With this change this is no longer possible. What's the best way here? Remove "Translation"-Type from submod? Shouldn't make any difference (still disabled, when no german, because parent mod is disabled?), right?

IvanSavenko:
@Laserlicht 
Noted. Will recheck this mod tomorrow.

IvanSavenko:
Changed logic for translation mods.
Translation mods will be hidden if they have different base language. However player still can manually enable/disable all of them.
Translation submods will be visible & toggleable by players as before - we don't have many of those, so should not be a problem.

PR #3095, opened 2023-10-22
fix team alignments

### LINKED ISSUE #3091: Random Team alignments is missing borders
**Describe the bug**
Missing borders for team alignments window and some lines that shouldn't there:
grafik

**Did it work earlier?**
Same problem in 1.3.2

**Version**
 - OS: Ubuntu
 - Version: nightly

Laserlicht:
fixes #3091

image

### DIFF
--- a/client/lobby/RandomMapTab.cpp
+++ b/client/lobby/RandomMapTab.cpp
@@ -403,16 +403,9 @@ TeamAlignments::TeamAlignments(RandomMapTab & randomMapTab)
 	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
 
 	widget = std::make_shared<TeamAlignmentsWidget>(randomMapTab);
-
 	pos = widget->pos;
 
-	backgroundTexture = std::make_shared<FilledTexturePlayerColored>(ImagePath::builtin("DiBoxBck"), Rect(0, 0, pos.w, pos.h));
-	backgroundTexture->playerColored(PlayerColor(1));
-
-	std::swap(GH.createdObj.front()->children.end()[-1], GH.createdObj.front()->children.end()[-2]); // widget to top
-	
 	updateShadow();
-
 	center();
 }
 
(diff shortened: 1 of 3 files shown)


client/lobby/RandomMapTab.cpp
```diff
@@ -107,7 +107,7 @@ RandomMapTab::RandomMapTab():
 	//new callbacks available only from mod
 	addCallback("teamAlignments", [&](int)
 	{
-		GH.windows().createAndPushWindow<TeamAlignmentsWidget>(*this);
+		GH.windows().createAndPushWindow<TeamAlignments>(*this);
```
IvanSavenko, line 110: Why we now need two windows for this?
TeamAlignmentsWidget and TeamAlignments?
What's the difference between them?

Laserlicht: One is the Window, the other a Widget inside of this Window. I don't find a nice way to add border without it.

Nordsoft91:
That's something really weird.

There is configurable widget, which is described in json file.
Instead of replacing regular texture with textureplayercolored in json, we are going to have **new object** which is basically adding hardcoded texture and then on the top of it creating old configurable object?

Laserlicht:
> That's something really weird.
> 
> There is configurable widget, which is described in json file. Instead of replacing regular texture with textureplayercolored in json, we are going to have **new object** which is basically adding hardcoded texture and then on the top of it creating old configurable object?

Yes, it's really weird... There is no `config/widgets/randomMapTeamsWidget.json`...
I don't know why it works without json...

Nordsoft91:
@Laserlicht , this file is located in extras. When extras disabled you will see classic ui, not hdmod ui

Laserlicht:
@Nordsoft91 Use the config now.

Adjustment to mod: https://github.com/vcmi-mods/vcmi-extras/pull/19

Nordsoft91 (approved):
1) Ideally, would be good to have frame as configurable widget too, however not sure how many use cases could be
2) cancel button is covered by frame if only 2-3 players are selected, would be good to fix, but not regression

PR #3078, opened 2023-10-19
convert line endings from CRLF (Windows) to LF (Linux/Unix)

josch:
Mixed line endings cause problems when exporting patches with git-format-patch and then trying to "git am" a patch with mixed and non-matching line endings. In such a situation git will fail to apply the patch.

This commit runs the dos2unix tools on the remaining files with CRLF (\r\n) line endings to convert them to line-feeds (\n) only.

Files that are Windows specific like *.vcxproj and *.props files were not converted.

Closes: #3073

PR #3096, opened 2023-10-22
Add CI stage to ensure LF line endings

Alexander-Wilms:
Based on @josch's command:
https://github.com/vcmi/vcmi/pull/3078#issuecomment-1772150744


Alexander-Wilms:
If you prefer, I can drop the second commit and you can do the remaining dos2unix run.

.github/workflows/github.yml
```diff
@@ -145,6 +145,14 @@ jobs:
       with:
         submodules: recursive
 
+    - name: Ensure LF line endings
+      if: ${{ startsWith(matrix.preset, 'linux-clang-test') }}
+      run: |
+        find . -path ./.git -prune -o -path ./AI/FuzzyLite -prune -o -path ./android -prune \
+        -o -path ./osx -prune -o -path ./test/googletest -prune -o -type f \
+        -not -name '*.png' -and -not -name '*.vcxproj*' -and -not -name '*.props' \
+        -not -name '*.wav' -not -name '*.ico' -print0 | ( ! xargs -0 grep -l -z -P '\r\n' )
```
josch, line 154: Parenthesis create a subshell. I do not think you need to spawn a new subshell and grouping with curly braces is sufficient to achieve the effect you want, no?

IvanSavenko (approved):

Issue #1691, opened 2023-03-15
In battle ranged units move 1 tile before hitting in mele


krs0:
**Describe the bug**
In battle ranged units move 1 tile before hitting in mele

**Expected behavior**
Ranged units should stay put before retaliating.

**Actual behavior**
https://user-images.githubusercontent.com/15194321/225360280-95fa5b4c-7d98-4764-be81-7c4bcf3def10.mp4

**Version**
 - OS: Windows
 - Version: develop

IvanSavenko:
That's AI action, right? AI is permitted to move his units...

krs0:
Yes Battle AI action. But very weird and pointless and  above all, deffinetly not OH3 behavior.

dydzio0614:
current battle ai is not trying to mimic H3 AI

rilian-la-te:
@dydzio0614 and I do not think than it should. I think if we can have smarter battle AI, it will be even better.

dydzio0614:
yep, we can have unlimited alternative ais for another purposes

krs0:
One of VCMI battle AI should mimic exactly! H3 battle AI!

It is the core of H3 battles. Maybe vast majority of players will observe only weird things like this useless movement, but more experienced players rely on battle AI doing things in a certain way.

rilian-la-te:
@krs0 only for neutral creatures and optional. AI players' battle AI should not even try to mimic H3 battle AI. You can do such battle AI (mimic H3) as enhancement, if you want)

DjWarmonger:
>  but more experienced players rely on battle AI doing things in a certain way

Well, they'll have to learn how to play again :P

dydzio0614:
the AI which is supposed to mimic H3 is not started by anybody yet

kambala-decapitator:
> more experienced players rely on battle AI doing things in a certain way

which is called "abusing", if we forget a handful of maps where certain AI behaviour is required

krs0:
At least can we agree that this 1 hex moving is stupid?  :)

nullkiller:
> At least can we agree that this 1 hex moving is stupid? :)

[...]

PR #3098, opened 2023-10-22
Battle ai movement fix

nullkiller:
A lot of debugging showed some bugs and flaws in initial design. Should be a bit better now.
AI should see strong stacks (archers) even if guarded by weak stacks (1-units).
AI prefers more slower targets comparing to own stack speed even if they are not reachable this turn. Still it is not hardcoded behavior.


krs0:
Is this affected as well? #1691 
To me from all AI's actions this is one of the strangest, and the one more often encountered.

nullkiller:
> Is this affected as well? #1691 To me from all AI's actions this is one of the strangest, and the one more often encountered.

No. That behavior relates to fact that AI always selects hex with lowest number to attack. Sure we can select move with shortest path in case scores are the same.

In general Battle AI does not mimic OH3 behavior. Vise versa it attempts to overcome some abuses. For OH3 like behavior Stupid AI can be used. It is also not the same but at least it is simpler and will stay like it by design.

krs0:
But why move at all? Ranged unit should stay put and attack. Or move away from attacking stack. This one hex move makes no sense.

kodobi:
If AI will always have to move at least 1-hex before melle attack than it could potentially lead to some exploits.

PR #3101, opened 2023-10-23
Fix data import on Android

IvanSavenko:
Ignore case when checking whether path is white-listed. Fixes data import with H3 data that has different case in directory names.

We had reports where player has "data" or "mp3" directories in imported h3 files, but our Android launcher only copies "Data" or "Mp3".

Blind fix since I don't have java knowledge or Android build set up, will test using build artifacts.


IvanSavenko:
Tested on Android - seems to be working as expected now

kambala-decapitator (approved):

PR #3110, opened 2023-10-26
Mod compatibility fixes

IvanSavenko:
Minor fixes to avoid potential crashes if player has outdated mods:
- Do not crash if handler for map objects was not found, instead fall back onto generic (empty) handler. Discovered since I removed separate "shrine" handler (merged into "configurable"), but hota mod uses "shrine" handler in 1.3.
- Block loading of objects into another mod namespace. Side effect of "torosar " fix - mods that still edit "torosar " now attempt to create new object with mostly invalid/unset properties. With this change such attempts will be ignored.

### DIFF
--- a/lib/mapObjectConstructors/CObjectClassesHandler.cpp
+++ b/lib/mapObjectConstructors/CObjectClassesHandler.cpp
@@ -194,9 +194,9 @@ TObjectTypeHandler CObjectClassesHandler::loadSubObjectFromJson(const std::strin
 	assert(!scope.empty());
 
 	std::string handler = obj->handlerName;
-	if(!handlerConstructors.count(obj->handlerName))
+	if(!handlerConstructors.count(handler))
 	{
-		logMod->error("Handler with name %s was not found!", obj->handlerName);
+		logMod->error("Handler with name %s was not found!", handler);
 		// workaround for potential crash - if handler does not exists, continue with generic handler that is used for objects without any custom logc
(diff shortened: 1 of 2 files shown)


lib/mapObjectConstructors/CObjectClassesHandler.cpp
```diff
@@ -193,13 +193,16 @@ TObjectTypeHandler CObjectClassesHandler::loadSubObjectFromJson(const std::strin
 	assert(identifier.find(':') == std::string::npos);
 	assert(!scope.empty());
 
+	std::string handler = obj->handlerName;
 	if(!handlerConstructors.count(obj->handlerName))
```
Nordsoft91, line 197:
```suggestion
	if(!handlerConstructors.count(handler))
```

lib/mapObjectConstructors/CObjectClassesHandler.cpp
```diff
@@ -193,13 +193,16 @@ TObjectTypeHandler CObjectClassesHandler::loadSubObjectFromJson(const std::strin
...
 	{
-		logGlobal->error("Handler with name %s was not found!", obj->handlerName);
-		return nullptr;
+		logMod->error("Handler with name %s was not found!", obj->handlerName);
```
Nordsoft91, line 199:
```suggestion
		logMod->error("Handler with name %s was not found!", handler);
```

Issue #3112, opened 2023-10-27
Sonarcloud Code Analysis


Alexander-Wilms:
**Describe your proposal**
Setting up Sonarcloud Code Analysis could help improve code quality

I've set it up for my fork and these are the results:
https://sonarcloud.io/summary/overall?id=Alexander-Wilms_vcmi

It tries to detect possible bugs, code smells and security issues:

![image](https://github.com/vcmi/vcmi/assets/3226457/44e5678e-5f69-4016-b27d-e11f1a5e5771)

**Additional context**

Alexander-Wilms:
I've experimented with it a bit.

SonarCloud supports automatic analysis, which only requires one to set up the project on https://sonarcloud.io.
The alternative is analyzing the code using Github Actions, which is what fheroes2 does. But automatic analysis seems a lot simpler for VCMI.

The only things I had to configure were:
1) Give the https://github.com/apps/sonarcloud bot `Read and write access to checks, commit statuses, pull requests, and security events`
2) Create a corresponding project: https://sonarcloud.io/projects/create
3) Exclude 3rd party code from analysis:
![image](https://github.com/vcmi/vcmi/assets/3226457/6dee92a6-d50d-40bb-a833-4c5900313f04)
https://sonarcloud.io/project/settings?category=exclusions&id=vcmi_vcmi#sonar.coverage.exclusions

Then you get the overview where you can browse and filter discovered issues:
![image](https://github.com/vcmi/vcmi/assets/3226457/f36dfc09-041e-4a46-8036-6ef04506af50)

And the bot creates a comment on each PR after the analysis is done, summarizing the newly introduced issues, if any.

I created a test PR adding an unused variable. This also gets caught during CI due to the `-Werror` flag, but other kinds of error are not.

https://github.com/Alexander-Wilms/vcmi/pull/2
![image](https://github.com/vcmi/vcmi/assets/3226457/875de9c0-fc79-43b4-b041-2ae3f141c5db)

![image](https://github.com/vcmi/vcmi/assets/3226457/3ccc1543-b4bd-4855-9d7b-06ab73f268f0)

The bot's comment links here: https://sonarcloud.io/summary/new_code?id=Alexander-Wilms_vcmi&pullRequest=2
![image](https://github.com/vcmi/vcmi/assets/3226457/f2b22cb2-9da7-4dcb-8330-289fb4016246)

If the PR fails to pass the quality gate, that is shown like a failed CI job:
[...]

PR #3113, opened 2023-10-27
Fix small issues

Alexander-Wilms:
Found most of them using the analysis linked in https://github.com/vcmi/vcmi/issues/3112


Alexander-Wilms:
You mean for the SDL calls?

Does that matter, since the calling function is written in C++?

rilian-la-te:
Yes, it is matter. If we will not get any bugs on exotic systems - passing nullptr to pure C is also okay.

client/NetPacksClient.cpp
```diff
@@ -367,19 +367,27 @@ void ApplyClientNetPackVisitor::visitGiveBonus(GiveBonus & pack)
...
-
-	CGI->mh->waitForOngoingAnimations();
+	if(CGI)
+	{
+		if(CGI->mh)
+		{
+			CGI->mh->onObjectFadeOut(obj, pack.initiator);
+			CGI->mh->waitForOngoingAnimations();
+		}
+	}
```
Nordsoft91, line 377:
```suggestion
	if(CGI && CGI->mh)
	{
		CGI->mh->onObjectFadeOut(obj, pack.initiator);
		CGI->mh->waitForOngoingAnimations();
	}
```

client/NetPacksClient.cpp
```diff
@@ -367,19 +367,27 @@ void ApplyClientNetPackVisitor::visitGiveBonus(GiveBonus & pack)
...
-
-	CGI->mh->waitForOngoingAnimations();
+	if(CGI)
+	{
+		if(CGI->mh)
+		{
+			CGI->mh->onObjectFadeIn(obj, pack.initiator);
+			CGI->mh->waitForOngoingAnimations();
+		}
+	}
```
Nordsoft91, line 390:
```suggestion
	if(CGI && CGI->mh)
	{
		CGI->mh->onObjectFadeIn(obj, pack.initiator);
		CGI->mh->waitForOngoingAnimations();
	}
```

lib/BasicTypes.cpp
```diff
@@ -111,7 +111,7 @@ int AFactionMember::moraleValAndBonusList(TConstBonusListPtr & bonusList) const
 	bonusList = getBonusBearer()->getBonuses(moraleSelector, cachingStrMor);
 
 	int32_t maxGoodMorale = VLC->settings()->getVector(EGameSettings::COMBAT_GOOD_MORALE_DICE).size();
-	int32_t maxBadMorale = -VLC->settings()->getVector(EGameSettings::COMBAT_BAD_MORALE_DICE).size();
+	int32_t maxBadMorale = - (int32_t) VLC->settings()->getVector(EGameSettings::COMBAT_BAD_MORALE_DICE).size();
```
Nordsoft91, line 114: why is needed?

lib/int3.h
```diff
@@ -82,19 +82,10 @@ class int3
 
 	constexpr bool operator<(const int3 & i) const
 	{
-		if (z < i.z)
+		if (x < i.x || y < i.y || z < i.z)
```
Nordsoft91, line 85:
```suggestion
		return z < i.z || y < i.y || x < i.x;
```

lib/mapObjects/IMarket.cpp
```diff
@@ -157,7 +157,7 @@ std::vector<int> IMarket::availableItemsIds(EMarketMode mode) const
 const IMarket * IMarket::castFrom(const CGObjectInstance *obj, bool verbose)
 {
 	auto * imarket = dynamic_cast<const IMarket *>(obj);
-	if(verbose && !imarket)
+	if(verbose && !imarket && obj)
```
Nordsoft91, line 160: if obj is nullptr we simply ignore it?

lib/rmg/float3.h
```diff
@@ -117,20 +117,10 @@ class float3
 
 	bool operator<(const float3 & i) const
 	{
-		if (z<i.z)
+		if (x<i.x || y<i.y || z<i.z)
```
Nordsoft91, line 120:
```suggestion
		return z < i.z || y < i.y || x < i.x;
```

server/battles/BattleActionProcessor.cpp
```diff
@@ -1445,8 +1445,11 @@ bool BattleActionProcessor::makePlayerBattleAction(const CBattleInfoCallback & b
 	else
 	{
 		auto active = battle.battleActiveUnit();
-		if(!active && gameHandler->complain("No active unit in battle!"))
+		if(!active)
+		{
+			gameHandler->complain("No active unit in battle!");
```
Nordsoft91, line 1450: change is not required here and in other similar places. It's compact style.

server/battles/BattleFlowProcessor.cpp
```diff
@@ -265,6 +265,10 @@ const CStack * BattleFlowProcessor::getNextStack(const CBattleInfoCallback & bat
 		return nullptr;
 
 	auto next = q.front().front();
+
+	if(next == nullptr)
+		return nullptr;
+
```
Nordsoft91, line 271:
```suggestion
```

Instead, on line 287 replace
```
if(!next->willMove();
```

with
```
if(!next || !next->willMove())
```

server/processors/HeroPoolProcessor.cpp
```diff
@@ -145,11 +145,17 @@ bool HeroPoolProcessor::hireHero(const ObjectInstanceID & objectID, const HeroTy
...
+		gameHandler->complain("Invalid map object!");
 		return false;
+	}
 
-	if (!playerState && gameHandler->complain("Invalid player!"))
+	if (!playerState)
+	{
+		gameHandler->complain("Invalid player!");
 		return false;
+	}
```
Nordsoft91, line 158: unnecessary changes

CI/linux-qt6/validate_json.py
```diff
@@ -12,19 +13,19 @@
...
 
 errors = []
-for path in sorted(Path('.').glob('**/*.json')):
+for path in sorted(Path(".").glob("**/*.json")):
     # because path is an object and not a string
     path_str = str(path)
     try:
-        with open(path_str, 'r') as file:
-            if VALIDATION_TYPE == 'json':
+        with open(path_str, "r") as file:
```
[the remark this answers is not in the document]
Alexander-Wilms: It's the default for [ruff](https://github.com/astral-sh/ruff)

lib/mapObjects/IMarket.cpp
Alexander-Wilms: Oh, right. Should I split the error message in two?

lib/BasicTypes.cpp
Alexander-Wilms: https://rules.sonarsource.com/c/RSPEC-876/

lib/int3.h
```diff
@@ -82,19 +82,7 @@ class int3
...
-		if (y < i.y)
-			return true;
-		if (y > i.y)
-			return false;
-		if (x < i.x)
-			return true;
-		if (x > i.x)
-			return false;
-		return false;
+		return z < i.z || y < i.y || x < i.x;
```
Nordsoft91, line 85: hm, I'm thinking maybe this is not equivalent to initial implementation which can introduce problems:
`A(0, 0, 1) < B(1, 0, 0)`
Before change: false
After change: true

Nordsoft91: equivalent condition looks following:
`return z < i.z || (z == i.z && (y < i.y || (y == i.y && x < i.x)));`

client/widgets/Buttons.h
```diff
@@ -67,7 +67,7 @@ class CButton : public CKeyShortcut
 	void setBorderColor(std::optional<ColorRGBA> borderColor);
 
 	/// adds one more callback to on-click actions
-	void addCallback(std::function<void()> callback);
+	void addCallback(std::function<void()> const & callback);
```
kambala-decapitator, line 70:
```suggestion
	void addCallback(const std::function<void()> & callback);
```

client/widgets/Buttons.cpp
```diff
@@ -66,7 +66,7 @@ void CButton::setBorderColor(std::optional<ColorRGBA> newBorderColor)
 	borderColor = newBorderColor;
 }
 
-void CButton::addCallback(std::function<void()> callback)
+void CButton::addCallback(std::function<void()> const & callback)
```
kambala-decapitator, line 69:
```suggestion
void CButton::addCallback(const std::function<void()> & callback)
```

launcher/main.cpp
```diff
@@ -35,7 +35,7 @@ int main(int argc, char * argv[])
 
 	MainWindow mainWindow;
 	mainWindow.show();
-	result = vcmilauncher.exec();
+	result = QApplication::exec();
```
kambala-decapitator, line 38: now there'll be warning "variable vcmilauncher is unused"

kambala-decapitator:
> @Alexander-Wilms I think if you request C code from C++ functions, using NULL is better than nullptr. C does not know about nullptr.

nullptr is just a helper so that you can't assign null pointer to an int

launcher/main.cpp
Alexander-Wilms: Right, will drop the commit.

AI/BattleAI/BattleAI.cpp
```diff
@@ -197,7 +197,7 @@ BattleAction CBattleAI::useCatapult(const BattleID & battleID, const CStack * st
 	}
 	else
 	{
-		EWallPart wallParts[] = {
+		std::array<EWallPart, 7> wallParts {
```
IvanSavenko, line 200:
```suggestion
		std::array wallParts {
```
(yes, this works)
https://en.cppreference.com/w/cpp/language/class_template_argument_deduction

`std::array<EWallPart, 7>` form is actually somewhat dangerous since it does not checks for initializer size.
If you initialize such array with less than 7 elements it will still compile but won't work as expected

PR #3115, opened 2023-10-27
Allow configuring all objects

### LINKED ISSUE #3087: add vmap extension
**Describe the bug**
If you save map in map editor and don't add the ".vmap" extension by yourself it's missing. File without extension is not shown in VCMI.

**Expected behavior**
Map editor should automatically add ".vmap" is missing.

Also it would be grerat if the "Maps" folder could be the default one in the saving dialog.

**Version**
 - OS: Ubuntu
 - Version: develop

Nordsoft91:
All rewardable objects are possible to customise in map editor.

Also, fixes #3087

### DIFF
--- a/lib/mapObjectConstructors/CRewardableConstructor.cpp
+++ b/lib/mapObjectConstructors/CRewardableConstructor.cpp
@@ -44,7 +44,9 @@ void CRewardableConstructor::configureObject(CGObjectInstance * object, CRandomG
 {
 	if(auto * rewardableObject = dynamic_cast<CRewardableObject*>(object))
 	{
-		objectInfo.configureObject(rewardableObject->configuration, rng);
+		if(rewardableObject->configuration.info.empty())
+			objectInfo.configureObject(rewardableObject->configuration, rng);
+		
 		for(auto & rewardInfo : rewardableObject->configuration.info)
 		{
 			for (auto & bonus : rewardInfo.reward.bonuses)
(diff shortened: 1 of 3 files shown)


mapeditor/mainwindow.cpp
```diff
@@ -77,22 +77,26 @@ void MainWindow::loadUserSettings()
...
 	}
 	auto position = s.value(mainWindowPositionSetting).toPoint();
-	if (!position.isNull())
+	if(!position.isNull())
 	{
 		move(position);
 	}
+	lastSavingDir = s.value(lastDirectorySetting).toString();
+	if(lastSavingDir.isEmpty())
+		lastSavingDir = QString::fromStdString(VCMIDirs::get().userCachePath().make_preferred().string());
```
Laserlicht, line 91:
```suggestion
		lastSavingDir = QString::fromStdString(VCMIDirs::get().userDataPath().make_preferred().string());
```
Shouldn't it DataPath?

IvanSavenko:
Do we embed configuration for all objects into vmap?
So if I use a mod that modifies Witch Hut then these changes won't have effect on vmap's made without such mod?

lib/mapObjectConstructors/CRewardableConstructor.cpp
```diff
@@ -44,7 +44,9 @@ void CRewardableConstructor::configureObject(CGObjectInstance * object, CRandomG
 {
 	if(auto * rewardableObject = dynamic_cast<CRewardableObject*>(object))
 	{
-		objectInfo.configureObject(rewardableObject->configuration, rng);
+		if(rewardableObject->configuration.info.empty())
+			objectInfo.configureObject(rewardableObject->configuration, rng);
+
```
IvanSavenko, line 49: This will break reroll of revisitable objects, e.g. Windmills

IvanSavenko (changes requested):

PR #3120, opened 2023-10-28
Improve ccache hit rate for PRs

Alexander-Wilms:
Ensure the ccache for each PR is separate by appending the PR ID to the key, so they don't interfere with each other.
Fall back to ccache of the vcmi/vcmi repo if no PR-specific ccache is found.

CI for a PR:
image
image

CI for a commit outside a PR (e.g. https://github.com/Alexander-Wilms/vcmi/actions/runs/6680396279/job/18153474337):
image
image

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -166,18 +166,36 @@ jobs:
       env:
         VCMI_BUILD_PLATFORM: x64
 
-    - name: ccache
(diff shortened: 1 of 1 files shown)


Alexander-Wilms:
For some reason, `steps.PR.outputs.pr_found` seems to be inverted and the PR ID thus cannot be used for the cache key. Will need to investigate.

Issue #2993, opened 2023-10-01
Trying to load save during active battle causes a crash


nullkiller:
**Describe the bug**
A crash happens when you try to load save file from active battle window. 

**Game logs**
`BattleInfo::~BattleInfo()
{
	for (auto & elem : stacks)
		delete elem;

	for(int i = 0; i < 2; i++)
		if(auto * _armyObj = battleGetArmyObject(i))
			>>>>_armyObj->battle = nullptr;
}`

**To Reproduce**
1. Start any battle
2. On your turn open Load Game menu
3. Load any game, confirm that you are ok to loose all progress
4. Observe server crash

**Expected behavior**
Should not crash

**Actual behavior**
It crashes

**Did it work earlier?**
I believe so

**Version**
 - OS: Windows, msvc
 - Version: active develop (1.4)

IvanSavenko:
Duplicate of #2953 

Will check BattleInfo destructor. Thanks for the info. Not reproducible on my system for some reason.

Povelitelh3:
Now, in the current development, the game crashes even from the adventure map. And not just in battle.

PR #3122, opened 2023-10-29
#2993 - destruct battle info first when destructing CGameState

nullkiller:


lib/gameState/CGameState.cpp
```diff
@@ -409,6 +409,7 @@ CGameState::CGameState()
 
 CGameState::~CGameState()
 {
+	currentBattles.clear();
```
IvanSavenko, line 412:
```suggestion
	// explicitly delete all ongoing battles first - BattleInfo destructor requires valid CGameState
	currentBattles.clear();
```

PR #3127, opened 2023-10-31
Use the object templates with least terrains allowed

DjWarmonger:
RMG will still choose random template for object appearance, however it will prefer the ones that are terrain-specific (have least number of allowed terrains).

It becomes apparent if you check Subterranean Gates from Antagraich Burning assets.

### DIFF
--- a/lib/rmg/RmgObject.cpp
+++ b/lib/rmg/RmgObject.cpp
@@ -130,6 +130,16 @@ void Object::Instance::setTemplate(TerrainId terrain, CRandomGenerator & rng)
 		auto terrainName = VLC->terrainTypeHandler->getById(terrain)->getNameTranslated();
 		throw rmgException(boost::str(boost::format("Did not find graphics for object (%d,%d) at %s") % dObject.ID % dObject.subID % terrainName));
 	}
+	//Get terrain-specific template if possible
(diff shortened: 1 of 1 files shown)


lib/rmg/RmgObject.cpp
```diff
@@ -130,6 +130,16 @@ void Object::Instance::setTemplate(TerrainId terrain, CRandomGenerator & rng)
...
+	//Get terrain-specific template if possible
+	int leastTerrains = (*boost::min_element(templates, [](const std::shared_ptr<const ObjectTemplate> & tmp1, const std::shared_ptr<const ObjectTemplate> & tmp2)
+	{
+		return tmp1->getAllowedTerrains().size() < tmp2->getAllowedTerrains().size();
+	}))->getAllowedTerrains().size();
+
+	vstd::erase_if(templates, [leastTerrains](const std::shared_ptr<const ObjectTemplate> & tmp)
+	{
+		return tmp->getAllowedTerrains().size() > leastTerrains;
+	});
```
Nordsoft91, line 142: I believe if one template can be placed only on water and another from mod can be placed on water + new water terrain, random template has to be chosen, so this logic is not quite good.
What is the example of selecting improper template?

PR #3128, opened 2023-10-31
CI: Don't append timestamp to ccache keys

Alexander-Wilms:
This will reduce total cache size and should improve cache hit rates further, since PRs don't use multiple caches for each CI matrix entry, so the caches of older PRs are less likely to get evicted

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -176,9 +176,9 @@ jobs:
         restore-keys: |
           ${{ matrix.preset }}-PR-${{ github.event.number }}
           ${{ matrix.preset }}-no-PR
-        # actual cache takes up less space, at most ~1 GB
-        max-size: "5G"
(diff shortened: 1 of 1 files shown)


Alexander-Wilms:
Might not work:

```
Post job cleanup.
ccache stats
Save cache using key "ccache-linux-gcc-test-no-PR-".
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/vcmi/vcmi --files-from manifest.txt --use-compress-program zstdmt
Failed to save: Unable to reserve cache with key ccache-linux-gcc-test-no-PR-, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/develop, Key: ccache-linux-gcc-test-no-PR-, Version: 15c80211763d03468c2c9070680654f9264282cf4daa2c6ceac80f2e3eaeb295
```

Issue #953, opened 2022-09-19
Memory leak


Nordsoft91:
1. Start any scenario
2. Measure memory usage
3. Restart scenario 10x times or more
4. Observe memory leak

<img width="893" alt="Снимок экрана 2022-09-19 в 21 51 50" src="https://user-images.githubusercontent.com/9308612/191087512-e33fe903-a1c0-45e1-9daa-db65cbb0f31c.png">

konopka90:
Hi, I did simple investigation of the problem. My attempt was to use `valgrind` to find and analyse memory leaks. 
Unfortunately, VCMI executed together with valgrind is very slow (around 50x slower on my poor VM) but I can say that I achieved small success.
From the other hand slower code execution is also helpful if want to detect some race conditions. 

Env:
- VCMI commit ID: a7f521cbd9000a22249ae593447a7c34a2625db1
- data files: Heroes of Might and Magic III Complete Edition GOG
- Ubuntu 20.04
- valgrind-3.15.0
- gcc

```
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 9.4.0-1ubuntu1~20.04.1' --with-bugurl=file:///usr/share/doc/gcc-9/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++,gm2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-9 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-9-Av3uEd/gcc-9-9.4.0/debian/tmp-nvptx/usr,hsa --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
[...]

PR #3129, opened 2023-10-31
Fix memleak related to battleResult

smanolloff:
Addresses an issue mentioned in my comment here: https://github.com/vcmi/vcmi/issues/953#issuecomment-1787151606

A `BattleResult` object created in `CGameHandler::setBattleResult`:

```cpp
auto br = new BattleResult();
// ...
battleResult.data = br;
```

leaks in `CGameHandler::setupBattle`:

```cpp
battleResult.set(nullptr);
// effectively battleResult.data = nullptr
```

### DIFF
--- a/lib/CondSh.h
+++ b/lib/CondSh.h
@@ -22,6 +22,13 @@ template <typename T> struct CondSh
 
 	CondSh(T t) : data(t) {}
 
+	void unset()
+	{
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Hi. 
1) we use "master" branch only for releases, for active development we use "develop" branch.
2) This bug should be fixed in develop branch

PR #3131, opened 2023-10-31
More tolerance for Subterranean Gates placement

DjWarmonger:
Sometimes Gates were not placed where possible. Now they can have a little offset to fit easier.


Nordsoft91 (approved):

Nordsoft91:
Now default behavior would be to place gate with one tile shift. Would it be better to place on exact position, and then, if fail, allow tolerance?

DjWarmonger:
Actually the **pair** of Gates is optimized for max distance:
`return path2.valid() ? (dist * otherDist) : -1.f;`
which can give various results depending on other objects in the zone.

PR #3132, opened 2023-10-31
Support for "selectAll" reward

DjWarmonger:
Give all rewards with matching limiter, as in WoG Alm's House.


lib/mapObjects/CRewardableObject.cpp
```diff
@@ -50,6 +50,38 @@ void CRewardableObject::selectRewardWthMessage(const CGHeroInstance * contextHer
 	sd.text = dialog;
 	sd.components = loadComponents(contextHero, rewardIndices);
 	cb->showBlockingDialog(&sd);
+
+}
+
+void CRewardableObject::grantAllRewardsWthMessage(const CGHeroInstance * contextHero, const std::vector<ui32> & rewardIndices, bool markAsVisit) const
```
IvanSavenko, line 56: This is copy-pasted method `CRewardableObject::grantRewardWithMessage` with 1-2 different lines? Perhaps unify code?

IvanSavenko (commented):
This mode will likely be broken if:
- reward 1 gives enough experience to levelup
- reward 2 gives something other than experience / skills / resources

Check `CRewardableObject::grantReward` method

lib/mapObjects/CRewardableObject.cpp
DjWarmonger: Yep, I think I can just call the existing  function for every reward in the loop.

IvanSavenko:
Can you check if this works (for example) if:
- reward 1 gives enough XP to levelup
- reward 2 gives mana points (or artifacts / creatures)

And what if there are two rewards that give levels?
From what I see 2nd reward in these cases will not be granted correctly.

DjWarmonger:
It works fine, as every reward is given in a separate dialog, and game waits for the dialog to be closed. Even multiple rewards of same type.
I might consider merging rewards of same type just to save popup spam.

However, creature rewards seems to be messed up - I get random creature every time.
image

IvanSavenko:
Okay then.

Although random creatures is weird... From what I see, reward can only be reset on new day call.

Oh, this might be due to way we give creatures.
Since hero might have all 7 slots filled, game will:
- put new creatures into garrison of the rewardable object
- start exchange between rewardable object and hero.
- Not 100% sure whether this works correct with multiple rewards.

I also don't see how these creatures are cleared. So if another hero revisits this object later he might be able to access creatures left by previous visits.

DjWarmonger:
The garrison dialog works fine as well, just the creature type does not. In this case I set it to crystalDragon.

IvanSavenko:
Ah, could be randomization issue then.
Potentially - during filtering for possible selection candidates your data was ignored so any non-special creature was deemed applicable.
Feel free to debug `JsonRandom::loadCreature` - possibly `filterKeys` call in there returned all in-game creatures instead of only dragons.

PR #3144, opened 2023-11-03
Remove (some of) implicit conversions of Identifier to integer

IvanSavenko:
Part 1 out of several more.

Currently our Identifier classes have implicit conversion operator to integer. This creates situations where identifier can be silently converted to integer instead of type-safe identifier.

This PR attempt to track down all such cases, however due to numerous cases (~300 just in library) only some of those has been fixed in this PR.

Also made some changes to Identifier classes:
- Identifiers now have `toEntity` method that returns pointer to corresponding Entity interface, e.g. Artifact, Creature (for now - only some classes, will be added to other classes as needed)
- for old code identifiers have methods like toSpell/toCreature that return pointer to a concrete implementation, e.g. CArtifact, CCreature.

### DIFF
--- a/lib/CArtHandler.cpp
+++ b/lib/CArtHandler.cpp
@@ -167,7 +167,7 @@ bool CArtifact::isBig() const
 
 bool CArtifact::isTradable() const
 {
-	switch(id.getNum())
+	switch(id.toEnum())
 	{
 	case ArtifactID::SPELLBOOK:
 	case ArtifactID::GRAIL:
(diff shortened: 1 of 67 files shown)


lib/CArtHandler.cpp
```diff
@@ -167,7 +167,7 @@ bool CArtifact::isBig() const
 
 bool CArtifact::isTradable() const
 {
-	switch(id)
+	switch(id.getNum())
```
SoundSSGood, line 170: id of the same type as the switch-case (ArtifactID). Why do additional type conversion?

IvanSavenko: Ah, should be toEnum(). Will change.

But no. Additional type conversion is needed.

ArtifactID is class. Weird class that sometimes acts like an integer, but still a class.
ArtifactID::Type is enumeration (and this is what should be used for `switch`)

PR #3147, opened 2023-11-04
Serialization refactor - remove VLC from saved games

IvanSavenko:
Rather massive change, in terms of effect. Game will no longer write local VLC state to save files.

This functionality was half-functional for a long time:
- This was only done on server side - client will not receive VLC from save, leading to multiple issues on loading save with different mod list since client and server operate with different VLC with different object ordering.
- Any, even minor changes in mods are likely to break saves even if VLC were to be synchronized on client - since game embeds a lot of data into saves, such as name of files in mods.

Since I don't see a way to fix VLC synchronization - both due to amount of required changes (e.g. reloading active mods on client, including their filesystems) and due to potential unsolvable issues (e.g. since updating mod may change file layout, invalidating data in VLC from save) removing VLC from saves seems to be more stable approach.

List of changes:
- VLC data, such as creature stats or file names are now loaded from current versions of mods, and not from saves. Minor mod updates will take effect immediately, even in existing saves, with exception of data that game randomizes on map start, e.g. content of configurable objects.
- Saving/loading games should be slightly faster, since server no longer runs VLC serialization
- Server will no longer store VLC in save game. Removed serialization support of VLC
- All game entities (heroes, creatures, factions, etc) are now serialized by ID only
- Replaced remaining usages of boost::mpl with std alternatives
- Removed unused CLoadIntegrityValidator class
- Removed no longer used serialize() method from handlers and entities, to avoid accidental use.
- Moved mod compatibility check (ModHandler serialize method) to a new class since ModHandler is no longer serialized
- move artifact randomization from ArtHandler to GameState since ArtHandler is no longer serialized

TODO (separate PR):
- serialize identifiers by string and allow loading games with slightly different set of mods


krs0:
This with loading saves with different mods! is temporary workaround right?
Expected usecase would be to include needed mod names+versions in savefile. And save should be loaded only on 100% match.

krs0:
If loading forces same mod 100%, then pretty much works as intended. As long as a Mod with different CRC does not load, all good :).

IvanSavenko:
> If loading forces same mod 100%, then pretty much works as intended. As long as a Mod with different CRC does not load, all good :).

No, this does not works as intended.

Mod updates SHOULD NOT break saves.

If player updates mod he generally expects to be able to play same saves.
If player install UI-only mod he generally expects to be able to play same saves.

krs0:
Ok there are 2 separate things:

1. No game affecting mods... should not be in saves at all and thus should not influence saves.

2. Other mods... why this lax behaviour? Will create only problems. You have superDuperMegaArchanges on map from mod x. But you allow to load mod x.2 where these were potentially replaced by greenImps.

CRC should be a requirement for loading saves. No you load them on Server or client or from save game... you will have the same thing.

IvanSavenko:
I am player.
I am playing a game
Next time I start Launcher, notice update notification, update mod.

BOOM 

All my save are dead. And no easy way to revert to old versions.
Bad devs.

IvanSavenko:
> But you allow to load mod x.2 where these were potentially replaced by greenImps.

This is completely separate issue and something that we should detect instead of simply any mod update

IvanSavenko:
> We debated this briefly in the past. For git mods, releases for mods would solve those issues.

Too much of a pain and needs significant changes in Launcher, both desktop and Android.

Unless you plan to ask players to install all mods manually?

IvanSavenko:
Huh, while this was not my main intention, but start times are way better than before.
Testing startup of "Myth and Legend" - largest H3 map on release build with no mods:
- New game:
- - current daily build: 6.0s
- - this branch + RNG fix: 3.75s
- Loading game:
- - current daily build 4.5s
- - this branch + RNG fix: 2.2s

Still too long for 20-year old game, but still a huge improvement.

IvanSavenko:
Cleaned up CTypeInfo mess, got some more startup improvements.
End result:
- map start time: 6.0s -> 2.25s
- map save time: 0.15s -> 0.1s
- map load time: 4.5s -> 1.8s

Most of start/load time is not spent on GUI initialization, so outside of this PR scope, might look into it later but likely only after 1.4.

Issue #2848, opened 2023-09-16
Altar of sacrifice: expierence text on artifact selection is off


wb180:
**To Reproduce**
1. Go to Altar of Sacrifice
2. Select an artifact
3. Look for experience text

**Expected behavior**
Text in it's intended place

**Actual behavior**
Text is off on the left

**Screenshots**
![Screenshot_20230916_200824](https://github.com/vcmi/vcmi/assets/109801241/ca431abd-36be-4b14-854c-04624ecd9f16)

**Version**
Latest develop

PR #3152, opened 2023-11-05
Altar rework

SoundSSGood:
- The Creature Altar now works correctly with the "One Unit Left" behavior when clicking the "Sacrifice All" button. OH3 like.
- Creatures altar fixed crash. Open altar window, firstly click on the empty altar slot.
- The button to sacrifice creatures on the altar should be active whenever there are creatures on the altar. Even if the creature slot is not selected.
- Fixed https://github.com/vcmi/vcmi/issues/2848
- Other minor fixes.
- Widgets CAltar, CAltarArtifacts, CAltarCreatures now separated from CTradeWindow.
- A separate CAltarWindow window has been created.

This is just the beginning work. Work in progress. The created classes will be further modified.

### DIFF
--- a/client/widgets/CAltar.cpp
+++ b/client/widgets/CAltar.cpp
@@ -1,0 +1,457 @@
+/*
+ * CAltarWindow.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#include "StdInc.h"
+#include "CAltar.h"
+
+#include "../widgets/CAltar.h"
+#include "../gui/CGuiHandler.h"
+#include "../gui/CursorHandler.h"
+#include "../widgets/Buttons.h"
+#include "../widgets/Slider.h"
+#include "../widgets/TextControls.h"
+
+#include "../CGameInfo.h"
+#include "../CPlayerInterface.h"
+
+#include "../../CCallback.h"
+
+#include "../../lib/networkPacks/ArtifactLocation.h"
+#include "../../lib/CGeneralTextHandler.h"
+#include "../../lib/mapObjects/CGHeroInstance.h"
+#include "../../lib/mapObjects/CGMarket.h"
+
+CAltar::CAltar(const IMarket * market, const CGHeroInstance * hero)
+	: CTradeBase(market, hero)
+{
+	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
+
+	// Experience needed to reach next level
+	texts.emplace_back(std::make_shared<CTextBox>(CGI->generaltexth->allTexts[475], Rect(15, 415, 125, 50), 0, FONT_SMALL, ETextAlignment::CENTER, Colors::YELLOW));
+	// Total experience on the Altar
+	texts.emplace_back(std::make_shared<CTextBox>(CGI->generaltexth->allTexts[476], Rect(15, 495, 125, 40), 0, FONT_SMALL, ETextAlignment::CENTER, Colors::YELLOW));
+	deal = std::make_shared<CButton>(Point(269, 520), AnimationPath::builtin("ALTSACR.DEF"), CGI->generaltexth->zelp[585], std::bind(&CAltar::makeDeal, this));
+	expToLevel = std::make_shared<CLabel>(75, 477, FONT_SMALL, ETextAlignment::CENTER);
+	expForHero = std::make_shared<CLabel>(75, 545, FONT_SMALL, ETextAlignment::CENTER);
+}
+
+void CAltar::deselect()
+{
+	hLeft = hRight = nullptr;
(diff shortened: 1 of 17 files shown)


client/windows/CAltarWindow.cpp
```diff
@@ -0,0 +1,136 @@
...
+	, windowClosedCallback(onWindowClosed)
+{
+	OBJECT_CONSTRUCTION_CUSTOM_CAPTURING(255 - DISPOSE);
+
+	if(mode == EMarketMode::ARTIFACT_EXP)
+		createAltarArtifacts(market, hero);
+	else if (mode == EMarketMode::CREATURE_EXP)
+		createAltarCreatures(market, hero);
+	else
+		close();
```
IvanSavenko, line 38: How this can happen? Perhaps simply throw runtime error if caller created altar window in unsuitable object?

client/widgets/CAltar.cpp
```diff
@@ -0,0 +1,457 @@
...
+	sacrificeBackpackButton->block(hero->artifactsInBackpack.empty());
+
+	arts = std::make_shared<CArtifactsOfHeroAltar>(Point(-365, -11));
+	arts->setHero(hero);
+
+	int slotNum = 0;
+	for(auto & altarSlotPos : posSlotsAltar)
+	{
+		auto altarSlot = std::make_shared<CTradeableItem>(altarSlotPos, EType::ARTIFACT_PLACEHOLDER, -1, false, slotNum++);
+		altarSlot->clickPressedCallback = [this](std::shared_ptr<CTradeableItem> altarSlot) -> void
```
IvanSavenko, line 76: Perhaps move body of this large lambda into a separate method?

client/widgets/CAltar.cpp
```diff
@@ -0,0 +1,457 @@
...
+	OBJECT_CONSTRUCTION_CUSTOM_CAPTURING(255 - DISPOSE);
+
+	labels.emplace_back(std::make_shared<CLabel>(155, 30, FONT_SMALL, ETextAlignment::CENTER, Colors::YELLOW,
+		boost::str(boost::format(CGI->generaltexth->allTexts[272]) % hero->getNameTranslated())));
+	labels.emplace_back(std::make_shared<CLabel>(450, 30, FONT_SMALL, ETextAlignment::CENTER, Colors::YELLOW, CGI->generaltexth->allTexts[479]));
+	texts.emplace_back(std::make_unique<CTextBox>(CGI->generaltexth->allTexts[480], Rect(320, 56, 256, 40), 0, FONT_SMALL, ETextAlignment::CENTER, Colors::YELLOW));
+	lSubtitle = std::make_shared<CLabel>(180, 503, FONT_SMALL, ETextAlignment::CENTER, Colors::WHITE);
+	rSubtitle = std::make_shared<CLabel>(426, 503, FONT_SMALL, ETextAlignment::CENTER, Colors::WHITE);
+
+	unitsSlider = std::make_shared<CSlider>(Point(231, 481), 137, [this](int newVal) -> void
```
IvanSavenko, line 242: Another huge lambda

client/widgets/CAltar.cpp
SoundSSGood: As for me, large lambdas are ok. But I'll move it.

client/windows/CAltarWindow.cpp
SoundSSGood: Oooh, closing won't work from the constructor. Yes, throwing an error is a good idea.

client/widgets/CAltar.cpp
SoundSSGood: @IvanSavenko what do we consider a huge lambda?

IvanSavenko: Hmm... around 5-10 lines probably?

Although here I think what I dislike most is not size of lambda on its own but rather that this is quite large (80 lines) method that does multiple things:
- construct UI
- define button callback methods

SoundSSGood: The constructor seems to be intended for different things. It seemed to me that lambdas would be better than strange methods that would be used in just one place in the code. Perhaps it would be better to divide it into subsections, with the addition of comments?

IvanSavenko: Yeah, but in this case constructor is already quite large, so moving "unrelated" stuff into separate method might make sense.

> strange methods that would be used in just one place in the code

However this is how UI frameworks are usually done. For example in Qt you have "slots" - methods that are called when widget emits a "signal" callback.

So in this case it makes sense to have method "onUnitsSliderMoved" that is called when slider triggers callback. Methods IMO are not just a way to reuse code but also a way to organize it into logical parts.

This is also how most of our UI code is done. Not sure if there is any need to change such approach.

SoundSSGood: ok understood

client/windows/CAltarWindow.cpp
SoundSSGood: replaced by assertion

Issue #3151, opened 2023-11-05
windows build problem - replace symlink with junction link


gamestales:
**Describe the bug**
Building on windows fails with symlink creation

**Build logs**
Error		failed to create symbolic link 'C:/VCMI/build/bin/Release/mapeditor/icons': A required privilege is not held by the client.	

**To Reproduce**
Build from Visual Studio 2019

Build would require administrator privileges. It can be fixed with using junction links as described at:
https://stackoverflow.com/questions/61243174/replacement-of-create-symlink-in-windows
cmd.exe /c mklink /J "C:/VCMI/build/bin/$(Configuration)/launcher/icons" "C:/VCMI/source/launcher/icons"

**If the issue is accepted, I can do it.**

krs0:
I think I have solved the problem by enabling developer mode.

gamestales:
Windows settings developer mode
Just checked and it helps for me as well.

The build works well for me with junction link, so I still would like to do it.

kambala-decapitator:
> The build works well with junction link, so I still would like to do it.

Please go ahead

IvanSavenko:
Yep. Go ahead.
Feel free to also replace `copy_directory` calls from lib for "Mods" and "config" with symlink/junctions.

PR #3154, opened 2023-11-06
#3151-windows-build-symlinks

gamestales:


gamestales:
create link was moved to other file as described here:

https://stackoverflow.com/questions/26024235/how-to-call-a-cmake-function-from-add-custom-target-command/26024835#26024835

IvanSavenko (approved):

kambala-decapitator (approved):

PR #3157, opened 2023-11-06
Use minstd_rand instead of mt19937 to avoid save game bloat

IvanSavenko:
As in title.

Switches RMG that we use in game from mt19937 to minstd_rand.

This might reduce quality of randomization a bit, but:
1) Difference is unlikely to be noticeable
2) Should slightly improve save game loading / game startup times
3) This *greatly* reduces save game size. Size of save game on "All For One" on day 1 went down from 2 194 Kb to 1 213 Kb (This test was done while also including #3147 )

### DIFF
--- a/lib/CRandomGenerator.h
+++ b/lib/CRandomGenerator.h
@@ -14,7 +14,7 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
-using TGenerator = std::mt19937;
+using TGenerator = std::minstd_rand;
 using TIntDist = std::uniform_int_distribution<int>;
 using TInt64Dist = std::uniform_int_distribution<int64_t>;
 using TRealDist = std::uniform_real_distribution<double>;
--- a/lib/CRandomGenerator.h
+++ b/lib/CRandomGenerator.h
@@ -14,6 +14,10 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
+/// Generator to use for all randomization in game
+/// minstd_rand is selected due to following reasons:
+/// 1. Its randomization quality is below mt_19937 however this is unlikely to be noticeable in game
+/// 2. It has very low state size, leading to low overhead in size of saved games (due to large number of random generator instances in game)
 using TGenerator = std::minstd_rand;
 using TIntDist = std::uniform_int_distribution<int>;
 using TInt64Dist = std::uniform_int_distribution<int64_t>;


Alexander-Wilms:
Does that reduce the save game size due to the state size being 32 bits for minstd_rand instead of 19937 bits for mt_19937?

And is the size reduction more than 19902 bits (ignoring the VLC change) because you store the state for multiple instances of the pseudo random number generator?

As someone not very familiar with the codebase I think it would be good if the reasoning behind the choice could be added as a comment or as part of the commit message.

IvanSavenko:
> Does that reduce the save game size due to the state size being 32 bits for minstd_rand instead of 19937 bits for mt_19937?

What you have said is size of *internal state*.
But std random generators can only be serialized as string. For mt_19937 that would be huge array of numbers, in text form. Which is way larger than 19937 bits. Probably around 3-4 times more than that.

> And is the size reduction more than 19905 bits (ignoring the VLC change) because you store the state for multiple instances of the pseudo random number generator?

Yep. We have random generator for each hero (including those not present on map) for so called "skill tree", so 156 rng's as well as few others

> As someone not very familiar with the codebase I think it would be good if the reasoning behind the choice could be added as a comment or as part of the commit message.

Or as comment in code, yeah, will do.

dydzio0614:
what problem does serializing RNG's solve BTW? I would probably serialize value result for next RNG roll and create generator on the fly for future rolls - internal state will not help much in "rolls after next one" since changing time should result in different roll anyway

DjWarmonger:
> internal state will not help much in "rolls after next one" since changing time should result in different roll 

Uhh, no. RMG state is persistent by design, and reloading shouldn't change it. This is to prevent save-scumming.

dydzio0614:
I mean I would personally go by these constraints:
1. When you load save only first next roll for certain generator is guaranteed to be same.
2. If RNG result is time-dependent generate 1 roll ahead and save the result in save, for next rolls just recreate random generator when loading save
2b. If RNG result is time-independent store a seed (in that case if we want to use time-independent generator we should use generator where same seed consisting of one 4/8 bytes number number will always lead to same result)

IvanSavenko:
No. It DOES NOT works that way.
You can't save seed from mt19937 engine in any way other than saving its entire state.
And in case of minstd_rand this is *technically* possible, but still undefined behavior and is not guaranteed to work in every implementation.

PR #3161, opened 2023-11-08
Artifacts swap optimization

SoundSSGood:
The artifact swap operation has been optimized. Now, instead of two net packages, one is used. This will be useful for the quick backpack window. For now the only case where a one artifact movement is used is the end of battle (planned to rework).

Fixed an old regression where the window with a question about assembling an artifact did not appear if the artifact was picked from the map.

giveHeroArtifact and putArtifact in CGameHandler have been merged. There was no reason to have two separate methods.

### DIFF
--- a/lib/networkPacks/PacksForClient.h
+++ b/lib/networkPacks/PacksForClient.h
@@ -975,13 +975,13 @@ struct DLL_LINKAGE CArtifactOperationPack : CPackForClient
 struct DLL_LINKAGE PutArtifact : CArtifactOperationPack
 {
 	PutArtifact() = default;
-	explicit PutArtifact(ArtifactLocation * dst, bool askAssemble = true)
-		: al(*dst), askAssemble(askAssemble)
+	explicit PutArtifact(ArtifactLocation & dst, bool askAssemble = true)
+		: al(dst), askAssemble(askAssemble)
 	{
 	}
 
 	ArtifactLocation al;
-	bool askAssemble = false;
+	bool askAssemble;
 	ConstTransitivePtr<CArtifactInstance> art;
 
 	void applyGs(CGameState * gs);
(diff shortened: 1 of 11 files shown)


server/CGameHandler.cpp
```diff
@@ -2688,55 +2687,54 @@ bool CGameHandler::moveArtifact(const ArtifactLocation & src, const ArtifactLoca
...
-
-		MoveArtifact ma(&srcLoc, &dstLoc);
-		if(srcLoc.artHolder == dstLoc.artHolder)
-			ma.askAssemble = false;
-		sendAndApply(&ma);
+		auto hero = getHero(dst.artHolder);
+		if(ArtifactUtils::checkSpellbookIsNeeded(hero, srcArtifact->artType->getId(), dstSlot))
+			giveHeroNewArtifact(hero, VLC->arth->objects[ArtifactID::SPELLBOOK], ArtifactPosition::SPELLBOOK);
+	}
+	catch(const std::bad_variant_access &)
```
IvanSavenko, line 2729: Not 100% sure, but I think artifact location no longer uses variant?

If yes - then this code might need changes, for handling commander artifacts. Or removed.

lib/networkPacks/PacksForClient.h
```diff
@@ -1064,17 +1064,25 @@ struct DLL_LINKAGE BulkMoveArtifacts : CArtifactOperationPack
 
 	ObjectInstanceID srcArtHolder;
 	ObjectInstanceID dstArtHolder;
+	std::optional<SlotID> srcCreature;
+	std::optional<SlotID> dstCreature;
```
IvanSavenko, line 1068: Alternatively we can add SlotID::NONE constant that will be used for cases like this one, instead of optional. 
Up to you.

server/CGameHandler.cpp
SoundSSGood: fixed

lib/networkPacks/PacksForClient.h
SoundSSGood: In ArtifactLocation is also used std::optional<SlotID>. Therefore, it is more convenient to use the same.
In general, there is no difference. Both are usable.

PR #3163, opened 2023-11-08
WIN32-install-missing-dlls

gamestales:


CMakeLists.txt
```diff
@@ -709,6 +709,19 @@ if(WIN32)
...
+
+	if(MSVC)
+		#parse all binary targets for their dll dependencies, recursively
+		install(TARGETS ${BINARY_TARGETS} RUNTIME_DEPENDENCIES
+			#skip parsing system dlls
+			PRE_EXCLUDE_REGEXES "api-ms-.*" "ext-ms-.*"
+			#include all requeired dlls found in build dir
+			POST_INCLUDE_REGEXES "${CMAKE_BINARY_DIR}.*" 
+			#do not include other dlls (from c:/windows)
+			POST_EXCLUDE_REGEXES ".*"
```
kambala-decapitator, line 721: Not sure how this supposed to work as this regex will match anything

server/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ else()
 	else()
 		add_executable(vcmiserver ${server_SRCS} ${server_HEADERS})
 	endif()
+	set(BINARY_TARGETS ${BINARY_TARGETS} vcmiserver PARENT_SCOPE)
```
kambala-decapitator, line 63: Here you ignore the if branch, but in other cases where static library is also involved you add target to the variable unconditionally. Why?

kambala-decapitator:
Also, which dlls are missing exactly? What issue does this PR solve?

CMakeLists.txt
gamestales: It includes all dlls that are found in build dir (C:/VCMI/build)
and then ignores all found elsewhere - out of the build dir - generally system dlls

server/CMakeLists.txt
gamestales: on the if branch there is SHARED library so it is also a binary

gamestales:
All boost dlls are missing, generally filesystem thread, also sdl2*.
After cpack + and install, every exe was missing something.

server/CMakeLists.txt
kambala-decapitator: I mean the parent `if`, line 53

kambala-decapitator:
> I definitely see both boost and sdl2 in our daily builds, both msvc and mingw builds.
> Those builds are generated as part of our CI using CPack.

exactly, CI builds are correct and working. So not sure what issue you're solving @gamestales

gamestales:
Then it must be my environment. But still on my local build boost and sdl2 dlls are missing.

PR #3165, opened 2023-11-09
Update and edit docs

DjWarmonger:
- Removed old Manual.md.
- Moved content to Game_Mechanics.md.

Quite some more features and shortcuts could be added.

- [x] Update Bonus descriptions to cover all types.


IvanSavenko:
> Update Bonus desciptions to cover all types.

You can search for "TODO" marks in that doc. All other bonuses should be up to date (unless I missed something)

PR #3167, opened 2023-11-10
Docs: Fix CMake command for enabling Ccache

Alexander-Wilms:


Alexander-Wilms:
There's an unrelated issue with macOS: https://github.com/actions/runner-images/issues/7522

Issue #2903, opened 2023-09-22
Wrong description of bad morale for creatures from 3 alignments


HarkUhm:
**Describe the bug**
I had creatures from 3 alignments on tournament map, but description tells just if all creatures fighting in hero's army are from the same town type,:
![image (1)](https://github.com/vcmi/vcmi/assets/141812373/aec1a6fd-fcde-4640-a3af-0d88b99e90a5)

The -1 to morale worked just ok.

**Expected behavior**
Should be info that If creatures fighting in hero's army are from three different town types, their morale is decreased by -1.

**Version**
 - OS: [Windows 10]
 - Version: [VCMI 1.3.2]

Issue #1442, opened 2023-01-16
morale string is cut


kambala-decapitator:
version: develop

"All troops of one **alignm**"

![Screenshot 2023-01-16 at 13 18 12](https://user-images.githubusercontent.com/1557784/212663434-7d4eaf17-8751-40b2-a463-82603a1de5f2.png)

Issue #962, opened 2022-09-21
buggy morale strings


kambala-decapitator:
1. Sepulcher (Crypt) has extra line break
2. Crypt and Warrior Tomb strings contain numbers twice

![Screenshot 2022-09-21 at 12 35 25](https://user-images.githubusercontent.com/1557784/191478186-462f505e-c79e-4262-aa4b-125cb5211301.png)

Also, in oh3 Warrior Tomb has different message:
![Screenshot 2022-09-21 at 15 54 05](https://user-images.githubusercontent.com/1557784/191509289-f491dbed-4e82-4955-881b-83e28d6c2c53.png)

IvanSavenko:
Looks to actually be a bit larger in scale than it appears, probably should be postponed to 1.2 to be safe:

- Extra line breaks are an easy part - see MoraleLuckBox::set method. H3 texts already contain line breaks at the start of strings, should probably trim them to ensure that only one line break gets added.
- H3 strings also already contain morale modifier (e.g. "-1") in string. Meaning that bonus description generation should be somehow made aware of such strings and avoid addition of second suffix.
- "Temporary until next battle" that should be replaced with "Warrior Tomb Visited": currently these objects default to generic description. Bonus class allows custom description but problem is, this is simple std::string that is generated on server. Correct fix would be to turn this in MetaString that can be properly translated by client. However this is definitely major change since it will affect everything that uses Bonus class, meaning - pretty much everything in the game.

PR #3170, opened 2023-11-10
#2903-morale-description-#1442

gamestales:
Partially #962

### DIFF
--- a/lib/mapObjects/CArmedInstance.cpp
+++ b/lib/mapObjects/CArmedInstance.cpp
@@ -111,7 +111,7 @@ void CArmedInstance::updateMoraleBonusFromArmy()
 	{
 		b->val = 2 - static_cast<si32>(factionsInArmy);
 		description = boost::str(boost::format(VLC->generaltexth->arraytxt[114]) % factionsInArmy % b->val); //Troops of %d alignments %d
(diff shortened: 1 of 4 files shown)


lib/mapObjects/CArmedInstance.cpp
```diff
@@ -111,7 +111,7 @@ void CArmedInstance::updateMoraleBonusFromArmy()
 	{
 		b->val = 2 - static_cast<si32>(factionsInArmy);
 		description = boost::str(boost::format(VLC->generaltexth->arraytxt[114]) % factionsInArmy % b->val); //Troops of %d alignments %d
-		description = b->description.substr(0, description.size()-2);//trim value
+		description = description.substr(0, description.size()-3);//trim value
```
IvanSavenko, line 114: What exactly is trimmed here?
Why 3 bytes/characters?

gamestales: It is simmilar in the upper one.
The arraytxt[114] is "\nTroops of %d alignments %d"
Which resolves to "\nTroops of 3 alignments -1"
I trim  " -1".
In the other place there is added " -1".
Maybe one should decide which aproach to use. I would recommend clearing arraytxt at start - while reading, but this would large change.

PR #3182, opened 2023-11-13
Move FPS counter to bottom left, set width of black background to suf…

Alexander-Wilms:
…fice for 3 digit FPS

image

I don't think there's an easy way to avoid the FPS of a previous frame being shown without a solid background.


Laserlicht:
Isn't it possible to use a CLabel instead of drawing directly? Im not sure here...

IvanSavenko:
Might be possible, but more complex.
FPS indicator must be active on any window - main menu, scenario selection, adventure map, battles, etc.
UI elements generally update only with window they belong to.

Issue #3153, opened 2023-11-06
[HDmod] artifact fast selection


Laserlicht:
**Describe your proposal**
Faster selection of artifacts should be implemented in VCMI.

HDmod:
![grafik](https://github.com/vcmi/vcmi/assets/13953785/a76844c1-1375-4391-b6af-c4608d015063)

**Additional context**
It should be implemented platform independent. One idea would be using the radial wheel to select artifacts (perfect on touch; also usable on desktop).

Example:
![grafik](https://github.com/vcmi/vcmi/assets/13953785/ecb5660c-c854-49e5-8662-ad12e37997ac)

krs0:
Radial wheel on desktop looks out of place.

Would go for pop-up window.

Laserlicht:
> Radial wheel on desktop looks out of place.
> 
> Would go for pop-up window.

Yes, also okay. But how to open it on touch devices?

PR #3186, opened 2023-11-13
Quick backpack window

SoundSSGood:
Solution for https://github.com/vcmi/vcmi/issues/3153.
Used mouse wheel press.
Viewing spells on scrolls is also supported.
Not tested on touch devices. I would be happy if someone tested this on iOS/Android.
Artifacts are sorted by ArtifactId. Scrolls by SpellId.

image

### DIFF
--- a/client/widgets/CArtifactsOfHeroBackpack.h
+++ b/client/widgets/CArtifactsOfHeroBackpack.h
@@ -22,18 +22,33 @@ class CListBoxWithCallback;
 class CArtifactsOfHeroBackpack : public CArtifactsOfHeroBase
 {
 public:
+	CArtifactsOfHeroBackpack();
 	CArtifactsOfHeroBackpack(const Point & position);
 	void swapArtifacts(const ArtifactLocation & srcLoc, const ArtifactLocation & dstLoc);
 	void pickUpArtifact(CHeroArtPlace & artPlace);
 	void scrollBackpack(int offset) override;
 	void updateBackpackSlots() override;
 	size_t getActiveSlotRowsNum();
+	size_t getSlotsNum();
 
-private:
+protected:
 	std::shared_ptr<CListBoxWithCallback> backpackListBox;
 	std::vector<std::shared_ptr<CPicture>> backpackSlotsBackgrounds;
-	const size_t HERO_BACKPACK_WINDOW_SLOT_COLUMNS = 8;
-	const size_t HERO_BACKPACK_WINDOW_SLOT_ROWS = 8;
+	const size_t slots_columns_max = 8;
+	const size_t slots_rows_max = 8;
 	const int slotSizeWithMargin = 46;
-	const int sliderPosOffsetX = 10;
+	const int sliderPosOffsetX = 5;
+
+	void initAOHbackpack(size_t slots, bool slider);
+};
+
+class CArtifactsOfHeroQuickBackpack : public CArtifactsOfHeroBackpack
+{
+public:
+	CArtifactsOfHeroQuickBackpack(const Point & position, const ArtifactPosition filterBySlot);
+	void setHero(const CGHeroInstance * hero);
+	ArtifactPosition getFilterSlot();
+
+private:
+	ArtifactPosition filterBySlot;
 };
--- a/client/widgets/CArtifactsOfHeroBackpack.h
+++ b/client/widgets/CArtifactsOfHeroBackpack.h
@@ -25,7 +25,7 @@ public:
 	CArtifactsOfHeroBackpack();
 	CArtifactsOfHeroBackpack(const Point & position);
 	void swapArtifacts(const ArtifactLocation & srcLoc, const ArtifactLocation & dstLoc);
-	void pickUpArtifact(CHeroArtPlace & artPlace);
+	void pickUpArtifact(CArtPlace & artPlace);
 	void scrollBackpack(int offset) override;
 	void updateBackpackSlots() override;
 	size_t getActiveSlotRowsNum();
--- a/client/widgets/CArtifactsOfHeroBackpack.h
+++ b/client/widgets/CArtifactsOfHeroBackpack.h
@@ -22,8 +22,8 @@ class CListBoxWithCallback;
 class CArtifactsOfHeroBackpack : public CArtifactsOfHeroBase
 {
 public:
+	CArtifactsOfHeroBackpack(size_t slotsColumnsMax, size_t slotsRowsMax);
 	CArtifactsOfHeroBackpack();
-	CArtifactsOfHeroBackpack(const Point & position);
 	void swapArtifacts(const ArtifactLocation & srcLoc, const ArtifactLocation & dstLoc);
 	void pickUpArtifact(CArtPlace & artPlace);
 	void scrollBackpack(int offset) override;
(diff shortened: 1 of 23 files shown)


client/windows/CHeroBackpackWindow.cpp
```diff
@@ -19,32 +19,63 @@
...
+
+	init();
+}
+
+void CHeroQuickBackpackWindow::showAll(Canvas & to)
+{
+	if(arts->getSlotsNum() == 0)
+	{
+		close();
+		return;
```
SoundSSGood, line 77: Dirty hack for closing a window.

krs0:
2 tests come to mind.
Does it ignore duplicates? You have 2-3 copies of same artifact... only 1 should show. Including case when artifact is already equipped?
Does it work with composed artifacts? For example when artifact chosen is last piece of a suit? (Do not remember what was latest logic there, could be that no questions should be asked)

SoundSSGood:
> Does it ignore duplicates?

Duplicates are not shown. Also, already equipped artifacts are not displayed.

> Does it work with composed artifacts?

Yes.

Laserlicht:
How about opening the window with a double click (would also work with touch screens)?
And if slot empty maybe a single click?

SoundSSGood:
yeah, it makes sense to try

dydzio0614:
plz not double click - make it exactly same as in HD mod

dydzio0614:
which was either middle mouse button or shift + click

Laserlicht:
> plz not double click - make it exactly same as in HD mod

Why not?

Most peoples than cannot use this feature because no keyboard connected to phone/tablet. 😉

SoundSSGood:
If you have any ideas on how to implement the middle mouse button event, I'll do it. At the moment I only see double click event or shift+click.

dydzio0614:
@Laserlicht I consider adding new double clicking invasive behavior that changes how base game works
and I would rather make this feature PC-exclusive under these circumstances, and it is mostly important for veteran players who need advanced shortcuts - muscle memory compatibility with HD mod is kinda important

Laserlicht:
@dydzio0614 Yes, my intention here is just that many vcmi players are currently finding their way in via mobile platforms. And of course it's effective advertising if everything is available there and works simple. On the PC, the competition with HDmod is simply too great to gain much reach there.

Of course, we can also implement the feature for the PC first. Only on the PC is of course better than not at all. :)

But I would avoid using the middle mouse button. It is currently a pure gesture button.

Better use Ctrl/Shift + Click

dydzio0614:
Well, originally middle mouse button was used and intended for map scrolling but its too inconvenient in many cases for that purpose.
So i do not see why it hurts to add middle mouse button in addition to ctrl/shift + click, and then think how to port it to mobile. If we use radian menu gesture for that popup then it would be consistent with PC where you bring radial menu via middle mouse click. We have bunch of people interested in VCMI on PC, but they are very used to various QoL features (like this one) and they need intuitive way to use it in VCMI.

IvanSavenko:
HD Mod uses mouse wheel press? 

Well, don't think that we need exact implementation to be honest. Majority of our players is on mobiles, and competing with HD Mod / HotA on desktop would be hard, so I won't count on having large number of players who are used to HD Mod way.

Perhaps should be done in the same way as radial wheel? (potentially with different UI instead of wheel - e.g. small backpack-like window instead of wheel)
- player presses mouse wheel / touches screen
- when player moves wheel / finger game shows up temporary menu
- when player releases wheel / finger game selects currently hovered item

@SoundSSGood 
You can check CGarrisonSlot::gesture and RadialMenu class for example.

dydzio0614:
@SoundSSGood can you change target to beta branch instead of develop?

krs0:
So on windows solution is right click? Or?

dydzio0614:
mouse wheel press

good enough for merge into beta IMO, additional shift+mouse shortcut would be cool but can be added later

Laserlicht:
Looks good. Also works basically on android devices. :)

dydzio0614 (approved):

client/widgets/CArtifactsOfHeroBackpack.h
```diff
@@ -22,18 +22,35 @@ class CListBoxWithCallback;
...
+	size_t getSlotsNum();
 
-private:
+protected:
 	std::shared_ptr<CListBoxWithCallback> backpackListBox;
 	std::vector<std::shared_ptr<CPicture>> backpackSlotsBackgrounds;
-	const size_t HERO_BACKPACK_WINDOW_SLOT_COLUMNS = 8;
-	const size_t HERO_BACKPACK_WINDOW_SLOT_ROWS = 8;
+	const size_t slots_columns_max = 8;
+	const size_t slots_rows_max = 8;
```
IvanSavenko, line 38: I don't think that we use such style anywhere?

client/widgets/CWindowWithArtifacts.cpp
```diff
@@ -239,16 +249,43 @@ void CWindowWithArtifacts::rightClickArtPlaceHero(CArtifactsOfHeroBase & artsIns
...
+				std::is_same_v<decltype(artSetWeak), std::weak_ptr<CArtifactsOfHeroMain>> ||
+				std::is_same_v<decltype(artSetWeak), std::weak_ptr<CArtifactsOfHeroKingdom>>)
+			{
+				GH.windows().createAndPushWindow<CHeroQuickBackpackWindow>(artSetPtr->getHero(), artPlace.slot);
+				auto backpackWindow = GH.windows().topWindow<CHeroQuickBackpackWindow>();
+				// Check if backpackWindow is off screen border.
+				auto windowOffset = Point(1, 1);
+				if(auto screenOff = cursorPosition.x + backpackWindow->pos.w + 15 - GH.screenDimensions().x; screenOff > 0)
+					windowOffset.x = screenOff;
+				backpackWindow->moveTo(cursorPosition - windowOffset);
```
IvanSavenko, line 288: There is already CIntObject::fitToScreen method that might work here

IvanSavenko:
Perhaps make number of rows & columns dynamic?
For example,
number of columns = sqrt(number of artifacts in backpack)

Right now if I have 8 artifacts in backpack that fit current slot I get 1 single long row of artifacts (if I get code correctly)
However displaying such case in more tidy 3x3 grid will probably look better.

Back when this window was used only for full backpack fixed number of columns was OK, but for quick UI more compact window would be preferrable

client/widgets/CArtifactsOfHeroBackpack.h
SoundSSGood: What do you have in mind? Could you please clarify

IvanSavenko: I mean code style.
I don't think that we use underscores in variables/members name anywhere. Normally we use camelCase:
```cpp
	const size_t slotsColumnsMax = 8;
	const size_t slotsRowsMax = 8;
```

SoundSSGood: Ahh, probably the snake case

krs0:
Scroll wheel press is meh! 

Ideally scroll wheel will scroll the backpack. In that case good luck pressing the M3 button without scrolling the backpack as well.

SHIFT + Click is the other way HD mod calls this window.

SoundSSGood:
I think that in any case it will be necessary to create a customization panel for hot keys and gestures. So this won't be a big problem. At least in the future.

SoundSSGood:
> number of columns = sqrt(number of artifacts in backpack)

It turned out well.

By the way, now the configuration of some windows is stored in config\widgets\*json. Are there any plans to migrate all in-game windows to json format? Does it make sense to make backpack windows loaded from json?

dydzio0614:
would be cool to move most of UI into json but this is massive amount of work

SoundSSGood:
> would be cool to move most of UI into json but this is massive amount of work

Ok, noted. And most likely I will do this for artifact related widgets.

IvanSavenko:
> By the way, now the configuration of some windows is stored in config\widgets*json. Are there any plans to migrate all in-game windows to json format? Does it make sense to make backpack windows loaded from json?

Hmm... Eventually - probably yes. But not something to actively work on. If you already work on this window - yes. But code rewrite just to move it to json - probably not. I am not against it but feels like waste of time.
Also - this system is currently not suitable for dynamic widgets.

For example moving market-based windows or hero/exchange windows into widgets makes sense.

But moving backpack window is probably not:
1. Will need a lot of copy-paste to describe 8x8 slot grid for "full" window
2. Does not supports dynamic widgets, like this quick backpack window that has different size based on number of artifacts

PR #3190, opened 2023-11-13
Use arraytxt for object bonus description

### LINKED ISSUE #962: buggy morale strings

gamestales:
Fixes #962

### DIFF
--- a/lib/networkPacks/NetPacksLib.cpp
+++ b/lib/networkPacks/NetPacksLib.cpp
@@ -993,29 +993,28 @@ void GiveBonus::applyGs(CGameState *gs)
 
 	std::string &descr = b->description;
 
-	if(bdescr.empty() && (bonus.type == BonusType::LUCK || bonus.type == BonusType::MORALE))
-	{
-		if (bonus.source == BonusSource::OBJECT_TYPE || bonus.source == BonusSource::OBJECT_INSTANCE)
-		{
-			descr = VLC->generaltexth->arraytxt[bonus.val > 0 ? 110 : 109]; //+/-%d Temporary until next battle"
-		}
-		else if(bonus.source == BonusSource::TOWN_STRUCTURE)
-		{
-			descr = bonus.description;
-			return;
-		}
-		else
-		{
-			descr = bdescr.toString();
-		}
-	}
-	else
+	if(!bdescr.empty()) 
 	{
 		descr = bdescr.toString();
 	}
-	// Some of(?) versions of H3 use %s here instead of %d. Try to replace both of them
-	boost::replace_first(descr, "%d", std::to_string(std::abs(bonus.val)));
-	boost::replace_first(descr, "%s", std::to_string(std::abs(bonus.val)));
+	else if(!descr.empty())
+	{
+		//use preseet description
+	}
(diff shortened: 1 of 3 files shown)


gamestales:
screen
So it works but it sets map object description on the fly.

It would be best to fill it during initialization.
Also probably should pass bonus type and object type instead of description.

lib/mapObjects/CRewardableObject.cpp
```diff
@@ -83,8 +83,21 @@ void CRewardableObject::onHeroVisit(const CGHeroInstance *h) const
...
+			for(auto index : rewards)
+			{
+				auto& reword = configuration.info[index].reward;
+				for(auto& bonus : reword.bonuses)
+				{
+					if(bonus.type == BonusType::MORALE && bonus.description.empty())
+						const_cast<std::string&>(bonus.description) = VLC->generaltexth->arraytxt[104]; "\nWarrior Tomb Visited -3";
+				}
+			}
+		}
```
gamestales, line 99: This setup part could be filled somewhere else. I have no idea where.

lib/networkPacks/NetPacksLib.cpp
```diff
@@ -993,25 +993,21 @@ void GiveBonus::applyGs(CGameState *gs)
...
-	if(bdescr.empty() && (bonus.type == BonusType::LUCK || bonus.type == BonusType::MORALE))
+	if(!bdescr.empty()) 
 	{
-		if (bonus.source == BonusSource::OBJECT_TYPE || bonus.source == BonusSource::OBJECT_INSTANCE)
-		{
-			descr = VLC->generaltexth->arraytxt[bonus.val > 0 ? 110 : 109]; //+/-%d Temporary until next battle"
-		}
-		else if(bonus.source == BonusSource::TOWN_STRUCTURE)
-		{
-			descr = bonus.description;
```
gamestales, line 1004: This is actully copied before.

lib/networkPacks/NetPacksLib.cpp
```diff
@@ -993,25 +993,21 @@ void GiveBonus::applyGs(CGameState *gs)
...
-			return;
-		}
-		else
-		{
-			descr = bdescr.toString();
```
gamestales, line 1009: bdescr is empty here

lib/mapObjects/CRewardableObject.cpp
IvanSavenko: Should be in object config, in json. Not hardcoded

lib/networkPacks/NetPacksLib.cpp
```diff
@@ -993,25 +993,21 @@ void GiveBonus::applyGs(CGameState *gs)
...
-		}
+		descr = bdescr.toString();
+	}
+	else if(descr.size()) {} //use preseet description
```
IvanSavenko, line 1000:
```suggestion
	else if(!descr.empty()) 
	{
		//use preset description
	} 
```

PR #3192, opened 2023-11-15
3150 show strength comparison next to threat level

### LINKED ISSUE #3150: Completely wrong threat level displayed
The following battle is more than challenging.
image

Thsi is AI result from battle
image

gamestales:
Fixes #3150


gamestales:
The quick solution by just adding difference of army strength.

If you decide, I could also change description instead of 
Threat Level: Challenging (+10%)

to 
Strength: Stronger/Similar/Weaker (+10%)

IvanSavenko:
Does not solves original issue - 20 wyverns against that army is simply impossible (unless you have wide variety of spells, top pvp player, and use gamey tactics)

IvanSavenko:
Other than that:
- AI value is very imprecise so it won't be useful to average player
- top player can easily guess exact force composition based on memorized AI value's, leading to information reveal

krs0:
There is the use case where units are close to immune to physical damage.
Like Andruids Ozze like creatures. For these cases threat levels will always be off.

gamestales:
army-smimilar
army-stronger
army-weaker

PR #3208, opened 2023-11-20
Bugfixing

### LINKED ISSUE #2963: gold overflow
to reproduce in adequate time: have lots of Ramparts with Treasury

reported in https://forum.vcmi.eu/t/gold-amount-overflow/6091, there's also video demonstrating the bug

Version: 1.3.2

### LINKED ISSUE #3205: AI player starts playing instead of human game
This is the second time I've encountered this error. I chose the red player, started the game and the bot started playing for me. I see his moves and I see his fights. In this case, he took the monks' home (I saw the battle from the inside) and after that the game fell.

All resources were minus 1

VCMI 1.4.0.c0e54b338a0d0be06fa43e24238d274a919a0960, mingw

[server_log.txt](https://github.com/vcmi/vcmi/files/13407031/server_log.txt)
[VCMI_Client_log.txt](https://github.com/vcmi/vcmi/files/13407033/VCMI_Client_log.txt)
[VCMI_Server_log.txt](https://github.com/vcmi/vcmi/files/13407034/VCMI_Server_log.txt)

IvanSavenko:
- Limit resources to 1 000 000 000 to prevent overflow. Fixes #2963
- Fix crash on battlefield deserialization
- Fix crash on artifact movement
- Disable quick combat by default to avoid confusing players
- Disable spectator mode in single player. Likely fixes #3205

### DIFF
--- a/client/lobby/CLobbyScreen.cpp
+++ b/client/lobby/CLobbyScreen.cpp
@@ -76,7 +76,7 @@ CLobbyScreen::CLobbyScreen(ESelectionScreen screenType)
 
 		card->iconDifficulty->addCallback(std::bind(&IServerAPI::setDifficulty, CSH, _1));
 
-		buttonStart = std::make_shared<CButton>(Point(411, 535), AnimationPath::builtin("SCNRBEG.DEF"), CGI->generaltexth->zelp[103], std::bind(&CLobbyScreen::startScenario, this, true), EShortcut::LOBBY_BEGIN_GAME);
+		buttonStart = std::make_shared<CButton>(Point(411, 535), AnimationPath::builtin("SCNRBEG.DEF"), CGI->generaltexth->zelp[103], std::bind(&CLobbyScreen::startScenario, this, false), EShortcut::LOBBY_BEGIN_GAME);
 		initLobby();
 		break;
 	}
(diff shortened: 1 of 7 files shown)


dydzio0614:
I am against disabling spectator mode in singleplayer, in fact this feature was mae with singleplayer in mind in first place

IvanSavenko:
Suggestions then?
This is not first or even second time when player would accidentally enable spectator mode without knowing what happened.

dydzio0614:
I would leave it as is since this is basically straight feature removal which is not justified enough by cases where "problem exists between keyboard and chair" or other cases where player doesn't see that none of flags belong to his name, which is definitely avoidable.

IvanSavenko:
Except that this "feature" causes nothing but issues.

- There is zero indication that spectator mode is active in lobby
- There is zero indication that spectator mode is active in game. AI just moves on its own
- We have multiple reports from players who end up in such situation.
- And I really don't understand what's the use-case for such features? I understand usage of spectator in MP, but in SP? Who and in which scenarios might use this feature?

If these questions are resolved then I am OK with keeping it. But as it is - I would rather remove it for time being.

dydzio0614:
OK do it as you want

DjWarmonger:
By "Spectator mode for singleplayer" you probably mean watching AI play without human intervention? Certainly that'd be useful instead of vcmieagles + autoskip. But bugfixing first.

IvanSavenko:
@dydzio0614 
Sorry about that...

Not against such feature (if somebody finds it useful) but as it is right now it is not really useable.

Can be looked into later, once we have proper spectator support, with both in-game UI and clear way to activate it in lobby. But at the moment we don't really have time/manpower to fix everything.

PR #3216, opened 2023-11-22
Better fix for allowed heroes on h3m maps

### LINKED ISSUE #3215: Cove heroes cannot be chosen in ROE,AB and SOD format maps
I don't know whether it occurs in other's game, it really occurs in my game with latest daily build.
image
Although I chose cove town as my fraction, but heroes from this alignment cannot be selected, and then, a castle hero was given by the system. 

**Version**
 - OS: Windows
 - Version: 1.4.0 daily build in 22nd of November

IvanSavenko:
- Fixes #3215

### DIFF
--- a/lib/IHandlerBase.h
+++ b/lib/IHandlerBase.h
@@ -116,7 +116,7 @@ public:
 		if(index < 0 || index >= objects.size())
 		{
 			logMod->error("%s id %d is invalid", getTypeNames()[0], index);
-			throw std::runtime_error("internal error");
+			throw std::runtime_error("Attempt to access invalid index " + std::to_string(index) + " of type " + getTypeNames().front());
 		}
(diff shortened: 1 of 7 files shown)


krs0:
Will these changes work in case no heroes or no artifacts?

IvanSavenko:
What do you mean? No heroes or no artifacts?

krs0:
Both :). The question comes from SW mods perspective, where no objects are needed.

What I was thinking, if it would be possible when touching code, to accomodate stuff that is not possible atm in VCMI

Not going out of the way to do it... but gradually eliminate some limitations.

IvanSavenko:
Not sure what's going on with SW mod, but this is fix for a regression. So should work as in 1.3

PR #3231, opened 2023-11-26
Split large spellbook from UI enhancements setting

dydzio0614:
As in title - this setting is pretty large change that people will probably want separately from other "ui enhancements changes". Additional thing is that spellbook animation is broken for large spellbook.

### DIFF
--- a/config/schemas/settings.json
+++ b/config/schemas/settings.json
@@ -39,8 +39,7 @@
 				"useSavePrefix",
 				"savePrefix",
 				"startTurnAutosave",
-				"enableUiEnhancements",
-				"enableLargeSpellbook"
+				"enableUiEnhancements"
 			],
 			"properties" : {
 				"playerName" : {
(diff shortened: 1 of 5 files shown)


Laserlicht:
Is it possible to disable spellbook animation checkbox if the large spellbook is enabled?
To make it clear for the user that it isn't working with large spellbook?

dydzio0614:
I do not think we support this particular case yet

Laserlicht:
Okay, then maybe in the future. :)

config/schemas/settings.json
```diff
@@ -131,6 +132,10 @@
 				"enableUiEnhancements" : {
 					"type": "boolean",
 					"default": true
+				},
+				"enableLargeSpellbook" : {
```
IvanSavenko, line 136: Perhaps place this into "gameTweaks" category instead of "general"?

IvanSavenko:
> I do not think we support this particular case yet (blocking option completely)

Checkboxes are buttons, so toggling button off and blocking it should work, e.g:
```cpp
    animationButton->setSelected(false);
    animationButton->block(true);
```

Laserlicht:
Can we merge? :)

I want to update translation. ;)

PR #3250, opened 2023-12-02
update german

Laserlicht:
@IvanSavenko Can we also recreate Launcher and Mapeditor translation? To be able to update these.


Laserlicht:
Thanks. Updated all translations and translated german.

PR #3254, opened 2023-12-03
search spells

### LINKED ISSUE #1467: [feature] search field in spell book
spells literally at your fingertips :)

the text field should gain focus when opening spell book on desktop platforms. The easiest matching would be "starts with", but fuzzy match would also do. when there's only one search result, pressing Enter/Return would cast it.

Laserlicht:
fix #1467

Do we need such a feature?

https://github.com/vcmi/vcmi/assets/13953785/563ab1d0-1e87-4ffb-b009-ce2ed0e24b93

### DIFF
--- a/client/windows/CSpellWindow.cpp
+++ b/client/windows/CSpellWindow.cpp
@@ -124,70 +124,13 @@ CSpellWindow::CSpellWindow(const CGHeroInstance * _myHero, CPlayerInterface * _m
 		offL = offR = offT = offB = offRM = 0;
 		spellsPerPage = 12;
 	}
+
 	pos = background->center(Point(pos.w/2 + pos.x, pos.h/2 + pos.y));
 
-	//initializing castable spells
-	mySpells.reserve(CGI->spellh->objects.size());
-	for(const CSpell * spell : CGI->spellh->objects)
-	{
-		if(!spell->isCreatureAbility() && myHero->canCastThisSpell(spell))
-			mySpells.push_back(spell);
-	}
-	std::sort(mySpells.begin(), mySpells.end(), spellsorter);
+	searchBox = std::make_shared<CTextInput>(Rect(10, isBigSpellbook ? 8 : 20, pos.w-20, 16), FONT_MEDIUM, std::bind(&CSpellWindow::searchInput, this));
+	searchBox->setText(battleSpellsOnly ? myInt->localState->spellbookSettings.spellbookLastFilterBattle : myInt->localState->spellbookSettings.spellbookLastFilterAdvmap);
 
-	//initializing sizes of spellbook's parts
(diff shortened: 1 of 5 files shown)


IvanSavenko:
Not against it I guess?

Better UI would be helpful though:
- add visible text box, e.g. using that semi-transparent rectangle class.
- not sure whether it should be on top or below the spell book
- Possibly add text "Search for spell" / "Type to find spell". Ideally - as gray text in text box that is erased once you start typing (current text input probably does not supports such alt-text, but might be an interesting option to have)

Should be OK for 1.4.X releases.

@misiokles Dedicated button probably not needed as long as there is clearly visible text input field that player can tap on

Laserlicht:
https://github.com/vcmi/vcmi/assets/13953785/8d6bf50a-7e1b-4f82-91d5-768836a04e1e

client/windows/CSpellWindow.cpp
```diff
@@ -124,70 +124,24 @@ CSpellWindow::CSpellWindow(const CGHeroInstance * _myHero, CPlayerInterface * _m
...
-		}
+		Rect r(90, isBigSpellbook ? 480 : 420, isBigSpellbook ? 160 : 110, 16);
+		const ColorRGBA rectangleColor = ColorRGBA(0, 0, 0, 75);
+		const ColorRGBA borderColor = ColorRGBA(128, 100, 75);
+		const ColorRGBA grayedColor = ColorRGBA(158, 130, 105);
+		searchBoxRectangle = std::make_shared<TransparentFilledRectangle>(r.resize(1), rectangleColor, borderColor);
+		searchBoxDescription = std::make_shared<CLabel>(r.center().x, r.center().y, FONT_SMALL, ETextAlignment::CENTER, grayedColor, CGI->generaltexth->translate("vcmi.spellBook.search"));
+
+		searchBox = std::make_shared<CTextInput>(r, FONT_SMALL, std::bind(&CSpellWindow::searchInput, this));
+		searchBox->removeFocus();
```
IvanSavenko, line 140: Why remove focus?
Tested a bit and I feel that it would be way useful if it will be focused by default*
Yes, this will break hotkeys to switch between pages but not sure if those are actually used by players?

* except for mobiles, but that case is already handled by CTextInput class

Laserlicht: Ah, was not aware that the mobile case already handled. That was the reason for me. Will remove. :)

IvanSavenko: Well, even if mobile case was not handled I would have suggested to add it.

Currently does not feels intuitive on desktop

Laserlicht: Yes agree.

Should much better now. :)

IvanSavenko (approved):

PR #3258, opened 2023-12-05
Market rework

SoundSSGood:
- Implemented class composition for market widgets. All subclasses will be located in CTradeBase.cpp. CAltar is now also a subclass (renamed to CExpAltar)
- TradePanels.cpp - implemented panels for the market. (artifacts, resources, creatures...) The market will now show two universal trading panels on the left and right. Instead of constructing it slot-by-slot.
- All widget files that relate to the market are placed in client/widgets/markets
- Now all slot selections in the market window have a thickness of 2 pixels.
- SelectableSlot is my solution for selecting slots. Using CArtPlace as an example you can see how it works. (CTradeableItem has not yet been fully implemented as it should)

TODO (not this PR)
- CTradeWindow will divide into window and widget. The window functionality will probably be combined with CAltarWindow. Since the altar is also part of a unified window of the town market. (Grotto in the Cove)
- CTradeableItem rework.
- CAltarArtifacts related issues.
- Will look into CTransformerWindow.


by003:
Did you think recruiting different war machines can be realized in the city as well?
Only stronghold can recruit two war machines, and nowadays, only ballista can be setted with second war machine in the city.

SoundSSGood:
https://sonarcloud.io/project/issues?pullRequest=3258&open=AYzk6ztNsmsd5FM-BSum&id=vcmi_vcmi
A strange requirement in my opinion. Are there any plans to migrate to C++20?

https://sonarcloud.io/project/issues?pullRequest=3258&open=AYzk6ztNsmsd5FM-BSul&id=vcmi_vcmi
This is not a pure virtual method. Works as intended. Should we leave it as is?

IvanSavenko:
Ah, sorry, should've answered sooner.

At least for now feel free to treat soundcloud as suggestions, and not something that *must* be fixed for merge (will check this PR soon by the way, sorry for delay). 

Feel free to comment on anything that you don't like in sonarcloud suggestions. I did adjusted some of its diagnostics to better match our code style, but there is too much stuff to check at once.

> A strange requirement in my opinion. Are there any plans to migrate to C++20?

Need to find how to tell sonarcloud that we use c++17, not c++20. Other than that - lambdas *might* be preferable over std::bind, but don't see need to enforce it. This rule sounds like a candidate to disable...

> This is not a pure virtual method. Works as intended. Should we leave it as is?

Kind of. Problem is, since this (I presume) is virtual method, so derived classes might override it. Consider this case:
```cpp
class ExtendedAltar : public CAltarArtifacts
{
    void deselect();
};

CAltarArtifacts::CAltarArtifacts()
{
    deselect();
} 

ExtendedAltar altar = new ExtendedAltar();
```
In this scenario you will get call to CAltarArtifacts::deselect inside constructor, even though you have created instance of ExtendedAltar.

I actually remember encountering tricky bug recently, due to calling virtual method in destructor, so will probably keep this rule on, if only to keep an eye on suspicious stuff.

Marking CAltarArtifacts::deselect or entire CAltarArtifacts as `final` should disable this diagnostics. Or you can keep code as it.

client/widgets/markets/TradePanels.h
```diff
@@ -0,0 +1,133 @@
...
+
+	void showPopupWindow(const Point & cursorPosition) override;
+	void hover(bool on) override;
+	void showAll(Canvas & to) override;
+	void clickPressed(const Point & cursorPosition) override;
+	std::string getName(int number = -1) const;
+	CTradeableItem(Point pos, EType Type, int ID, bool Left, int Serial);
+};
+
+struct STradePanel : public CIntObject
```
IvanSavenko, line 59: - should probably be a class, it is strange IMO for struct to inherit from struct (or vice-versa)
- we don't use 'S' or 'C' prefixes for new structs / classes

(same for other entities in this file)
```suggestion
class TradePanel : public CIntObject
```

client/widgets/markets/CTradeBase.h
```diff
@@ -0,0 +1,74 @@
...
+	void removeItems(const std::set<std::shared_ptr<CTradeableItem>> & toRemove);
+	void removeItem(std::shared_ptr<CTradeableItem> item);
+	void getEmptySlots(std::set<std::shared_ptr<CTradeableItem>> & toRemove);
+	virtual void makeDeal() = 0;
+	virtual void deselect();
+	virtual void onSlotClickPressed(const std::shared_ptr<CTradeableItem> & newSlot, std::shared_ptr<CTradeableItem> & hCurSlot);
+};
+
+// Market subclasses
+class CExpAltar : virtual public CTradeBase, virtual public CIntObject
```
IvanSavenko, line 55: Probably no need for abbreviations
```suggestion
class CExperienceAltar : virtual public CTradeBase, virtual public CIntObject
```

client/widgets/markets/CTradeBase.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+#include "../../../lib/CGeneralTextHandler.h"
+#include "../../../lib/mapObjects/CGHeroInstance.h"
+
+CTradeBase::CTradeBase(const IMarket * market, const CGHeroInstance * hero)
+	: market(market)
+	, hero(hero)
+{
+	deal = std::make_shared<CButton>(Point(), AnimationPath::builtin("ALTSACR.DEF"),
+		CGI->generaltexth->zelp[585], std::bind(&CTradeBase::makeDeal, this));
```
IvanSavenko, line 29: This button should actually be different for creature / artifacts altars:
`CGI->generaltexth->zelp[584]` -> creatures altar
`CGI->generaltexth->zelp[585]` -> artifacts altar

Identical for English, but different in some languages. Was reported on Discord recently

client/widgets/markets/TradePanels.cpp
```diff
@@ -0,0 +1,394 @@
...
+		break;
+	}
+
+	if(image)
+	{
+		image->moveTo(pos.topLeft() + posToBitmap);
+		CIntObject::showAll(to);
+	}
+
+	to.drawText(pos.topLeft() + posToSubCenter, FONT_SMALL, Colors::WHITE, ETextAlignment::CENTER, subtitle);
```
IvanSavenko, line 161: Likely old code, so feel free to ignore / postpone, but:
- might be better to move image once, when type/image is changed, not on every redraw
- drawText should be replaced CLabel

client/widgets/markets/TradePanels.h
```diff
@@ -0,0 +1,133 @@
...
+
+class CTradeableItem : public CIntObject, public std::enable_shared_from_this<CTradeableItem>
+{
+public:
+	std::shared_ptr<CAnimImage> image;
+	AnimationPath getFilename();
+	int getIndex();
+	using ClickPressedFunctor = std::function<void(const std::shared_ptr<CTradeableItem>&)>;
+
+	const CArtifactInstance * hlp; //holds ptr to artifact instance id type artifact
```
IvanSavenko, line 32: use better name?
```suggestion
	const CArtifactInstance * artInstance; //holds ptr to artifact instance id type artifact
```

IvanSavenko (approved):
Looks generally OK.
Might be better to replace multiple inheritance with composition, but other than that OK. 

Most of remaining issues likely existed in old code and only showed up in diff because code was moved around.

client/widgets/markets/TradePanels.h
SoundSSGood: > use better name?

CTradeableItem will be reworked. It's just moving the code. I also ignored warnings related to it in sonarcloud

client/widgets/markets/TradePanels.cpp
SoundSSGood: > Likely old code

Yes. I my plan is create a separate CLabel for this. And separate image.

client/widgets/CArtifactHolder.h
```diff
@@ -29,7 +29,7 @@ class CArtifactHolder
 	virtual void artifactAssembled(const ArtifactLocation & artLoc)=0;
 };
 
-class CArtPlace : public LRClickableAreaWTextComp
+class CArtPlace : public LRClickableAreaWTextComp, public SelectableSlot
```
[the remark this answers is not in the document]
SoundSSGood: Do you mean `: virtual public CIntObject` ?
SelectableSlot and LRClickableAreaWTextComp use CIntObject internally. Wouldn't we get two instances of CIntObject in the case without `virtual` ?

client/widgets/markets/TradePanels.h
SoundSSGood: > * we don't use 'S' or 'C' prefixes for new structs / classes

Noted

client/widgets/CArtifactHolder.h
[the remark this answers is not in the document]
IvanSavenko: No, I mean use composition instead of inheriting from SelectableSlot:

```cpp
 class CArtPlace : public LRClickableAreaWTextComp
{
    std::shared_ptr<SelectableSlot> slot;
...
```
In this case you won't need virtual inheritance at all and SelectableSlot can inherit CIntObject without virtual specifier

(or swap LRClickableAreaWTextComp with SelectableSlot and inherit from SelectableSlot and have LRClickableAreaWTextComp as member, but this might be more complex)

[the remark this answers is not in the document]
SoundSSGood: I don't think it would be correct to mix LRClickableAreaWTextComp and SelectableSlot. They have nothing in common except a common parent.

As for the composition: I plan to use SelectableSlot for the market as well. It will be needed a lot of additional wrappers for `selectSlot`, `isSelected` for each related class. Wouldn't this be code bloat?
Or perhaps
```
public:
  std::shared_ptr<SelectableSlot> slot;
```
is ok?

SoundSSGood:
> Marking CAltarArtifacts::deselect or entire CAltarArtifacts as `final` should disable this diagnostics. Or you can keep code as it.

Calling a virtual method with an explicit indication of the class `CTradeBase::deselect();` `CAltarCreatures::deselect();` gives an unambiguous understanding of the code.
Sonarcloud also understands this and removes the warnings.

IvanSavenko:
Looks OK, will merge tomorrow

PR #3259, opened 2023-12-06
Fix corpse

DjWarmonger:
- Moved blockVisit to object properties
- Added "removable" to object properties
- Updated object config accordingly

Actual fix for Corpse (blockVis and not removable):
- Do not place guard next to blockVis tile, if possible (so that some non-blocking object is accessible)
- Do not place two blockVis objects next to each other

Bonus fixes:

- Fixed randomly banning / exceeding cap of objects in zones. This could have severe side effects, too.
- Objects with limited per-map count will now be distributed among zones randomly

TODO:

- [x] Do not route road through Corpse
- [ ] Allow to change blockedVisitable per template (HoTA water object variants)

### DIFF
--- a/lib/mapObjectConstructors/AObjectTypeHandler.cpp
+++ b/lib/mapObjectConstructors/AObjectTypeHandler.cpp
@@ -97,16 +97,8 @@ void AObjectTypeHandler::init(const JsonNode & input)
 		aiValue = static_cast<std::optional<si32>>(input["aiValue"].Integer());
 
 	// TODO: Define properties, move them to actual object instance
-	if(input["blockVisit"].isNull())
-		blockVisit = false;
-	else
-		blockVisit = input["blockVisit"].Bool();
-
-	if(input["removable"].isNull())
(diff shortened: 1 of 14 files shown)


lib/mapObjectConstructors/AObjectTypeHandler.cpp
```diff
@@ -96,6 +96,18 @@ void AObjectTypeHandler::init(const JsonNode & input)
...
+		blockVisit = false;
+	else
+		blockVisit = input["blockVisit"].Bool();
+
+	if(input["removable"].isNull())
+		removable = false;
+	else
+		removable = input["removable"].Bool();
+
+
```
IvanSavenko, line 110: Our JsonNode evaluates bool to false if entry is not present.
And second block is wrong - should have been inverted
```suggestion
	// TODO: Define properties, move them to actual object instance
	blockVisit = input["blockVisit"].Bool();
	removable = input["removable"].Bool();

```

PR #3270, opened 2023-12-09
Update Readme.md

DjWarmonger:
Added some vanilla screenshots

Updated headline


Alexander-Wilms:
Maybe also add one big image, like in the fheroes2 README?

https://github.com/ihhub/fheroes2

DjWarmonger:
I deliberately used clickable miniatures to not bloat informative text with large images. It's just to indicate we have _something_ to show.

PR #3273, opened 2023-12-09
Fix passable hero

DjWarmonger:
Whatever happened, this makes sure it will never happen again no matter the config.


lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -301,6 +307,7 @@ void CGHeroInstance::setHeroType(HeroTypeID heroType)
 
 void CGHeroInstance::initHero(CRandomGenerator & rand, const HeroTypeID & SUBID)
 {
+	blockVisit = true;
```
dydzio0614, line 310: this line already exists in constructor, so maybe problem is somewhere else

DjWarmonger: Ahh yes, didn't mean to commit it

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -150,6 +150,11 @@ bool CGHeroInstance::isCoastVisitable() const
 	return true;
 }
 
+bool CGHeroInstance::isBlockedVisitable() const
```
dydzio0614, line 153: I think this should not be needed if blockvisit is properly assigned in constructor...

PR #3279, opened 2023-12-09
smooth scroll fix and gui option to disable

Laserlicht:


dydzio0614:
did not manually test but idea sounds good

Laserlicht:
Keep in mind, that it only stops the current "animation" (to avoid overlapping).

DjWarmonger:
Zuuzel already complainedit feels unnatural as it is.

But I like it, maybe make it optional feature?

Laserlicht:
> Zuuzel already complainedit feels unnatural as it is.

I think that's very individual. When I used 1.3.2, it felt very unnatural to me. :D

> But I like it, maybe make it optional feature?

The option already existed. I have added it to the GUI. But I leave it active by default because
- we don't differentiate between mobile and desktop (on mobile it should be standard in any case)
- Map dragging is not an OH3 function anyway
- Map dragging (at least with links) is deactivated by default on the desktop anyway

Issue #3295, opened 2023-12-10
Can't launch generated random map


DjWarmonger:
**Describe the bug**
Once generated random map cannot be launched - There's some issue with heroes.

**Game logs**
Hero is already on the map at (75 42 0)

**To Reproduce**
Unpack this map:
https://github.com/vcmi/vcmi/issues/3294

Launch the map with no mods (makes no difference as far as I see)

**Expected behavior**
Map would launch as usual

**Actual behavior**
Instant server crash

**Screenshots**
Looks like Grid sitting in prison at these coordinates is not banned on the map, none of imprisoned heroes are.
![Uploading Gird not banned on random map.png…]()

DjWarmonger:
Some clues from Discord channel: loaded map has multiple Prisons with subID 0, which ich wrong. Every Prison should have unique subID, and 0 is id of Orrin.
[MapWithPrisons.zip](https://github.com/vcmi/vcmi/files/13980833/MapWithPrisons.zip)

Warzyw647:
Another clue from the map editor. Screens from the current version (minus some rmg fixes).
Object inspector for a new hero on a new map. Picking a different name from the list in field Hero type changes SubID:
![newHero](https://github.com/vcmi/vcmi/assets/132299608/d4f6cb7f-da47-4093-a004-2c3cdab4be29)
object inspector for a new prison on a new map:
![newPrison](https://github.com/vcmi/vcmi/assets/132299608/a91dc239-44d7-42bb-8321-a6a3a03bd648)
and object inspector for a prison containing Neela on a SoD map opened in VCMI map editor:
![prisonFromSodMap](https://github.com/vcmi/vcmi/assets/132299608/d175f4d3-b722-4b06-bd15-a239628b7427)

godric3:
Not only prisons cause this problem, I tired to load Antagarich Burning maps. First they don't load due to `'torosar '` hero name that was changes some time ago. After fixing that map crashes with same error (Hero is already on the map ).
Can be reproduced with creating map with two heroes of same class (two knight). Two heroes of different classes won't trigger this crash.
Test map(rename to .vmap)
[HeroOnMap.zip](https://github.com/vcmi/vcmi/files/14077295/HeroOnMap.zip)

Looking into jsons it looks fine:
```
"hero_0" : {
		"l" : 0,
		"options" : {
			"experience" : 9223372036854775807,
			"owner" : "red",
			"type" : "core:edric"
		},
		"subtype" : "knight",
[...]

PR #3298, opened 2023-12-11
Fixes to hero / prison distribution

DjWarmonger:
- New class PrisonHeroDistributor
- Rolling and banning heroes is now protected with Mutex
- Do not iterate through entire hero list every time
- Misplaced or otherwise destroyed Prisons will return their heroes to the pool
- Same is true for Seer Huts which will restore quest artifacts to the pool and unban them on map, globally
- Quest artifacts on random map will correctly remain banned in game


DjWarmonger:
Found an issue: Whenever Prison object is deleted (for instance due to failed treasure placement), hero stays banned. However, Prison pool size is restored so another Prison can spawn.

The same is true for quest artifacts.

DjWarmonger:
Ready for review.
https://github.com/vcmi/vcmi/issues/3295 is still an issue, but this will need to wait for another day,

lib/rmg/CMapGenerator.h
```diff
@@ -65,8 +65,8 @@ class DLL_LINKAGE CMapGenerator: public Load::Progress
 	const std::vector<ArtifactID> & getAllPossibleQuestArtifacts() const;
 	const std::vector<HeroTypeID> getAllPossibleHeroes() const;
 	void banQuestArt(const ArtifactID & id);
-	void banHero(const HeroTypeID& id);
-
+	void unbanQuestArt(const ArtifactID & id);
+
```
Alexander-Wilms, line 69: Trailing whitespace
```suggestion

```

lib/rmg/modificators/PrisonHeroPlacer.h
```diff
@@ -0,0 +1,41 @@
...
+private:
+    void getAllowedHeroes();
+	size_t reservedHeroes;
+
+protected:
+
+    std::vector<HeroTypeID> allowedHeroes;
+};
+
+VCMI_LIB_NAMESPACE_END
```
Alexander-Wilms, line 41: Missing newline
```suggestion
VCMI_LIB_NAMESPACE_END

```

PR #3320, opened 2023-12-15
Fix Windows 7 support

IvanSavenko:
- Fixes issue with oneTBB / libTBB on windows 7 - explicitly define target Windows version when building via conan
- Added workflow to build 32-bit version
- Renamed 'mingw-ubuntu' to 'mingw' to prevent confusion - some players were thinking that this build is for Ubuntu. This is also more in-line with other cross-compiling CI's (ios/Android) which don't use host system in name


kambala-decapitator (approved):

Alexander-Wilms:
It works with Wine.

Laserlicht:
32bit works on Win7-VM (and Nullkiller AI).

grafik

PR #3325, opened 2023-12-16
Crashfixes

IvanSavenko:
- Added check for presence of AB files to detect incomplete data early
- Fixed crash on loading maps if Scholar has been changed to not be a "rewardable" object
- Fixed memory corruption on using member after delete in radial wheel for town/hero sorting
- Fixed crash on randomizing dwelling when no suitable dwelling was found
- Attempt to fix crash on destruction of object classes handler

Merging to master, will upload to Google Play to monitor situation with crashes.

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -69,6 +69,7 @@
 	"vcmi.lobby.noPreview" : "no preview",
 	"vcmi.lobby.noUnderground" : "no underground",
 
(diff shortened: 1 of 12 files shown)


Laserlicht:
>     * Fixed memory corruption on using member after delete in radial wheel for town/hero sorting

This change break the sort town feature. The `updateWidget` is needed. :(

IvanSavenko:
For me that redraw would always happen instantly? Probably when radial wheel is closed. Have not noticed any issues with reordering...

Laserlicht:
https://github.com/vcmi/vcmi/assets/13953785/99f21f0f-fa6c-48f7-9e53-c028dcbeada6

IvanSavenko:
Will recheck.
I was testing mostly with towns, might be different with heroes

IvanSavenko:
Confirmed, will fix

PR #3327, opened 2023-12-16
date time format

### LINKED ISSUE #3311: Same date & time of all autosave games
**Describe the bug**
If you want to load an autosave game, all of them have the same date & time so it is not possible to find out which is the latest one.

**Game logs**
Sorry, no logs, I am on Android.

**To Reproduce**
1. Play game, several rounds
2. Try to load an autosave game
3. All of them have the same date & time

**Expected behavior**
At least time of the saved games should be different.

**Actual behavior**
All the saved games have the same time (and dates).

**Did it work earlier?**
Yes, it worked okay before 1.4 version.

**Screenshots**
No screenshot.

**Version**
 - OS: Android
 - Version: 1.4.0 & 1.4.1

**Additional context**
Nothing.

Laserlicht:
fixes #3311

Datetime format for each language. Since there is no solution to get correct system format on each platform in c++.

### DIFF
--- a/include/vstd/DateUtils.h
+++ b/include/vstd/DateUtils.h
@@ -5,7 +5,7 @@ VCMI_LIB_NAMESPACE_BEGIN
 namespace vstd
 {
 
-	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt, std::string format = "");
+	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt, std::string format);
 	DLL_LINKAGE std::string getDateTimeISO8601Basic(std::time_t dt);
 
 }
--- a/include/vstd/DateUtils.h
+++ b/include/vstd/DateUtils.h
@@ -5,7 +5,7 @@ VCMI_LIB_NAMESPACE_BEGIN
 namespace vstd
 {
 
-	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt);
+	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt, std::string format = "");
 	DLL_LINKAGE std::string getDateTimeISO8601Basic(std::time_t dt);
 
 }
--- a/lib/Languages.h
+++ b/lib/Languages.h
@@ -68,6 +68,9 @@ struct Options
 	/// primary IETF language tag
 	std::string tagIETF;
 
+	/// DateTime format
+	std::string dateTimeFormat;
+
 	/// Ruleset for plural forms in this language
 	EPluralForms pluralForms = EPluralForms::NONE;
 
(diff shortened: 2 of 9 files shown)


lib/Languages.h
```diff
@@ -79,27 +82,27 @@ inline const auto & getLanguageList()
...
+		{ "russian",     "Russian",     "Русский",    "CP1251", "ru", "%F %T", EPluralForms::UK_3, true },
+		{ "spanish",     "Spanish",     "Español",    "CP1252", "es", "%d/%m/%Y %T", EPluralForms::EN_2, true },
+		{ "swedish",     "Swedish",     "Svenska",    "CP1252", "sv", "%F %T", EPluralForms::EN_2, true },
+		{ "turkish",     "Turkish",     "Türkçe",     "CP1254", "tr", "%d.%m.%Y %T", EPluralForms::EN_2, true },
+		{ "ukrainian",   "Ukrainian",   "Українська", "CP1251", "uk", "%d.%m.%Y %T", EPluralForms::UK_3, true },
+		{ "vietnamese",  "Vietnamese",  "Tiếng Việt", "UTF-8",  "vi", "%d/%m/%Y %T", EPluralForms::VI_1, true }, // Fan translation uses special encoding
+
+		{ "other_cp1250", "Other (East European)",   "", "CP1250", "", "", EPluralForms::NONE, false },
+		{ "other_cp1251", "Other (Cyrillic Script)", "", "CP1251", "", "", EPluralForms::NONE, false },
+		{ "other_cp1252", "Other (West European)",   "", "CP1252", "", "", EPluralForms::NONE, false }
```
kambala-decapitator, line 105: where did you get these formats from? Russian and Ukrainian having different formats makes no sense

Laserlicht: https://en.wikipedia.org/wiki/Date_format_by_country (but sometimes there are multiple).

Feel free to make code suggestions for better formats if you know the usual formats. But we should use numeric only I think.

kambala-decapitator: in that list (as well as to my knowledge) both use `dd.mm.yyyy`, so not sure why you picked different formats for those 2 languages

include/vstd/DateUtils.h
```diff
@@ -5,7 +5,7 @@ VCMI_LIB_NAMESPACE_BEGIN
 namespace vstd
 {
 
-	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt);
+	DLL_LINKAGE std::string getFormattedDateTime(std::time_t dt, std::string format = "");
```
kambala-decapitator, line 8: what's the use case of declaring default value and then copy-pasting code to read current format from settings? Either put that expression right here as default value (if possible) or create wrapper that passes it automatically (as I understand, that value can't be fetched inside a vstd function)

Laserlicht: It's not used yet. Could also remove default value if it's better?

I don't want to use settings in vstd. That doesn't make sense for a "std" file, I think.

lib/vstd/DateUtils.cpp
```diff
@@ -1,33 +1,29 @@
...
 		{
-			// locale not be available - keep default / global
+			try
+			{
+				s.imbue(std::locale(""));
+			}
+			catch(const std::runtime_error & e)
+			{
+				// locale not be available - keep default / global
+			}
```
IvanSavenko, line 22:
```suggestion
```
Can be removed since we no longer try to use system locale

include/vstd/DateUtils.h
kambala-decapitator: main point is to prevent copy-paste of reading from settings. You can create wrapper function inside the project (that is, not in vstd)

lib/vstd/DateUtils.cpp
```diff
@@ -1,33 +1,29 @@
...
-#endif
-
 		std::tm tm = *std::localtime(&dt);
 		std::stringstream s;
-		try
-		{
-			s.imbue(std::locale(""));
-		}
-		catch(const std::runtime_error & e)
+		if(format.empty())
```
IvanSavenko, line 13: Perhaps we should always explicitly specify format in caller code and remove this condition?

lib/vstd/DateUtils.cpp
kambala-decapitator: this code path is still used for `other_cp125*` languages

include/vstd/DateUtils.h
IvanSavenko: Move to general text handler then?
Or to related TextOperations namespace?

lib/vstd/DateUtils.cpp
IvanSavenko: other_* languages can't be selected here - those can only be selected for H3 encoding. Not for languages.

include/vstd/DateUtils.h
Laserlicht: Yes, I just didn't think it was necessary at the moment, as it is currently only used three times (settings are already used two times). I also wouldn't know where such a wrapper would fit in well.

lib/vstd/DateUtils.cpp
Laserlicht: Yes its an fallback. We could also remove and fallback to international format.

include/vstd/DateUtils.h
kambala-decapitator: well, 3 > 1 :)

I also don't know, try following Ivan's advice

IvanSavenko: I suggest either TextOperations or general text handler class.
Even if there are "just" 3 usages, all 3 are rather hard to read due to extremely long statements.

So moving those into a dedicated method makes sense

lib/vstd/DateUtils.cpp
kambala-decapitator: considering that other_ languages will never get here, I agree with removing the condition

kambala-decapitator (approved):

IvanSavenko (approved):

PR #3337, opened 2023-12-17
Translate new settings to polish

dydzio0614:
As in title


IvanSavenko (approved):
Will keep it open for now in case if we get new strings, but will merge for 1.4.2

PR #3341, opened 2023-12-17
Fix extraction progress display for mod installation

IvanSavenko:
- Launcher will now actually display progress of mod installation (percentage of extracted files).
- Reduced sleep interval while launcher wait for mod extraction to reduce visible lag


IvanSavenko:
Seems to be (kind of) working.
Installing hota is still slooow process, but at least now it has actual indication of progress.

No idea why it takes so long though...
Either something wrong with our zip extraction code, or with our Windows version of minizip, or there is some issue with Windows that we need to deal with - creation of thousands of files is often slow operation here.

Issue #2777, opened 2023-09-08
Concurrency races in turn timers


IvanSavenko:
**Describe the bug**
Currently, only state of TurnTimer class is protected.
However it is completely possible for:

- Connection thread: receives a pack (e.g. MakeAction)
- GameHandler thread: wakes up from sleep
- Connection thread: starts applying MakeAction received from player
- GameHandler thread: starts applying MakeAction due to timer timeout

This would lead to both threads accessing GameState in same time, leading to a crash.

**Did it work earlier?**
No, previously GameHandler thread would never access GameState

**Version**
 - develop branch

IvanSavenko:
@Nordsoft91

PR #3346, opened 2023-12-19
[1.4.2] Better timers

IvanSavenko:
As reported by Dydzio / pvp players, current version of timers needs improvements.

List of issues that need to be looked into:
- [x] Timers of all players should be visible at same time. Currently only current player timer is visible
- [x] Turn time menu is not really multiplayer-interactive. Most of UI works the way that you see it when another player switches to it and you see settings changes in real time
- [x] Some "vcmi extras" fields from random map tab are also bugged here, at least roads
- [x] it would be good to have separate timer visible for unit timer in pvp battle
- [x] during battle show timers inside battlefield window. Or in separate / external window similar to how hero info popup is positioned
- [x] decide how to handle / warn case where unit timer is set to 0 since this might lead to eternal defend in pvp battles once timer runs out

Changes in this PR:
- Game now shows timers for all human players
- Game now shows timers separately where applicable
- During battle game will now show separate widget with timers for attacker and defender (if applicable)

TODO:
- [x] Recheck everything in pvp with two clients
- [x] Restore notifications sound
- [ ] Try to improve in-battle timers widget

Not reproducible:
- Unit timer does not works for defender in PvP
- When timer runs out for defender there were green messages and some crap happening, eventually game crashed on attempting to move unit

Postponed:
- it would also be nice if playing creature animation did not cause turn timer to freeze and then catch up when animation stops (not trivial due to way our networking works in battles)


IvanSavenko:
Unfortunately unable to reproduce crash on timer running out at least in hotseat.
Will try 2 clients MP later.

However this might be an instance of #2777, so probably not trivial to fix at the moment

dydzio0614:
maybe somewhat related to two players having different creature speeds setup?
maybe it won't trigger that easily after removing 0 seconds of creature timer leading to infinite forced defending

IvanSavenko:
> maybe somewhat related to two players having different creature speeds setup? or maybe it won't trigger that easily after removing 0 seconds of creature timer leading to infinite forced defending

Currently timer for next unit start count down the moment player performs action with previous unit.
So if you have slow creature speed set you will only handicap yourself. Should not lead to crashes and such.

Will check scenario where 1 player has very low speed and another - very high speed, in case if weird stuff happens when player with fast animations makes several actions while player with slow animations still waiting for them to finish

dydzio0614:
why would we need to set battle timer to 0? the table you showed on slack had idea that when you set unit timer to 0 then battle timer would be ignored in pvp battles

IvanSavenko:
Ah, so:
- pve battles have timer (e.g. extra time)
- pvp battles have no timers

OK, will change

dydzio0614:
well, theoretically you can achieve "no pvp battle timer" also by setting unit timer to 99:99 but this is far from intuitive

PR #3356, opened 2023-12-21
[1.4.2?] Possibility to show hero info in levelup window

Laserlicht:
With right click on portrait you can now show hero info in level up window.

Difference to HD mod: Left click isn't possible (currently it's not allowed to change anything in this window while in level up menu).


SoundSSGood:
VCMI already has a CHeroArea class. It used for clickable portraits of heroes. Please consider using it instead of LRClickableArea. A little refactoring may be needed.

PR #3361, opened 2023-12-22
Added more logging to track down freeze on startup

IvanSavenko:
@dydzio0614 please try to reproduce this bug when you have time


dydzio0614:
is debug included in default log level?

IvanSavenko:
For log file - yes

Issue #3353, opened 2023-12-21
Terrain texture is not as clear as in OH3


DjWarmonger:
**Describe the bug**
Terrain texture (generated randomly) has more patches and details, making the map less readable.

**Game logs**
Just generate random map with VCMI and HoTA. Observe snow or sand terrain.

**Expected behavior**
Texture should be clear, with very little patches.

**Actual behavior**
Texture is messy.

**Did it work earlier?**
Doubt it.

**Screenshots**
VCMI: 
![image](https://github.com/vcmi/vcmi/assets/2566990/37ce0833-884a-42d0-81fd-13647560fe42)
HoTA:
![image](https://github.com/vcmi/vcmi/assets/2566990/95ec0071-c249-4cc2-9ece-fdca972f4333)

**Additional context**
I don't know how this terrain is generated, maybe @Nordsoft91 knows.

IvanSavenko:
I think H3 editor even has slider to control "messiness" (called "amount of decorations" or something like this)

Looks like H3 tileset consists from two blocks. In case of snow:
- tiles 49-56: 8 tiles with smooth snow
- tiles 57-72: 16 tiles with "decorations"
However vcmi treats them as single block:
https://github.com/vcmi/vcmi/blob/develop/config/terrainViewPatterns.json#L88

Not sure how to solve this in our terrain painter though.

PR #3362, opened 2023-12-22
Terrain decorations level

IvanSavenko:
Sounded fun so why not? 
Change allows to specify level of terrain decorations used for terrains on RMG maps. Currently set to fixed value of 15% which seems to be similar to example screenshot from #3353. For comparison - previous level was ~66%
If somebody (e.g. Warmonger) want to tweak it or make this option configurable in editor like in H3 - go ahead, but not something that I plan to do myself.

- Implements #3353

Examples:
(note: all terrains except water & rock are affected. I use snow to show difference and because it is most noticeable here)
- Default (15%):
зображення

- No decorations (0%):
зображення

- Only decorations (100%):
зображення

### DIFF
--- a/lib/rmg/threadpool/MapProxy.h
+++ b/lib/rmg/threadpool/MapProxy.h
@@ -28,7 +28,7 @@ public:
     void insertObjects(std::set<CGObjectInstance*>& objects);
     void removeObject(CGObjectInstance* obj);
 
-    void drawTerrain(CRandomGenerator & generator, std::vector<int3> & tiles, TerrainId terrain);
+	void drawTerrain(CRandomGenerator & generator, std::vector<int3> & tiles, TerrainId terrain);
     void drawRivers(CRandomGenerator & generator, std::vector<int3> & tiles, TerrainId terrain);
     void drawRoads(CRandomGenerator & generator, std::vector<int3> & tiles, RoadId roadType);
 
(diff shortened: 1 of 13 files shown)


DjWarmonger:
Works great for me :)

lib/rmg/threadpool/MapProxy.h
```diff
@@ -28,7 +28,7 @@ class MapProxy
     void insertObjects(std::set<CGObjectInstance*>& objects);
     void removeObject(CGObjectInstance* obj);
 
-    void drawTerrain(CRandomGenerator & generator, std::vector<int3> & tiles, TerrainId terrain);
+	void drawTerrain(CRandomGenerator & generator, std::vector<int3> & tiles, TerrainId terrain);
```
DjWarmonger, line 31: Unneccessary tab

DjWarmonger (approved):

IvanSavenko:
Cool. Will merge to 1.4.2 then

> Also, works for new terrains without modifications.

... as long as new terrains use same tiles layout

misiokles:
Highlands and Wastelands have more tiles as far I remember.

IvanSavenko:
@misiokles 
Should work I think?
hota terrains have custom tileset in vcmi - specifically because they have more tiles.
I did adjusted ranges for decorations/plain tiles for hota tileset, so should work (although have not checked in game)

lib/rmg/threadpool/MapProxy.h
IvanSavenko: Actually looks like entire file uses spaces instead of tabs (except for 1 line changed by me)

PR #3363, opened 2023-12-22
Frenzy immunities fixed, less broken genies spellcasting

dydzio0614:
Frenzy immunities are closer to H3 (orb of vulnerability special behavior not handled)

Also a bit better genies spellcasting check until i resigned from continuing

### DIFF
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -1732,6 +1732,12 @@ SpellID CBattleInfoCallback::getRandomBeneficialSpell(CRandomGenerator & rand, c
 				continue;
 		}
 			break;
+
+		case SpellID::FRENZY:
+		case SpellID::MIRTH:
+			if(subject->hasBonusOfType(BonusType::MIND_IMMUNITY))
+				continue;
+			break;
 		}
 		beneficialSpells.push_back(spellID);
 	}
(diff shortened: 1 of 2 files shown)


lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1673,7 +1674,7 @@ SpellID CBattleInfoCallback::getRandomBeneficialSpell(CRandomGenerator & rand, c
 				return !stack->canShoot();
 			});
 
-			if(!walker)
+			if(!walker || subject->hasBonusOfType(BonusType::SPELL_SCHOOL_IMMUNITY, SpellSchool::FIRE))
```
IvanSavenko, line 1677: This block is also used by Shield spell, which is earth spell

IvanSavenko: Strange. Why this block is needed? Code actually checks whether spell can be casted on this target - see `canBeCast` block above.

dydzio0614: the block does not work - if you try to cast genie spell on ballista then half of the attempts fail

IvanSavenko: Ah, we actually check whether spell can be cast *at all*, not whether spell can be cast on this target. Will fix.

IvanSavenko:
Closing, included in #3379

PR #3366, opened 2023-12-23
Battle stacks informations

### LINKED ISSUE #1490: [feature] show current battle losses
could be a separate dialog (if impossible to fit on the battle screen) showing initial and dead count for each unit for both sides, e.g. `5 / 20` meaning 5 dead, initial 20. Could use small unit icons like those on the "battle finished" screen.

### LINKED ISSUE #2961: Spell turns should be shown in creature menu
Like the same function here. It will be more convenient to show the time if in control. 
image

### LINKED ISSUE #3253: Fast unit info in battle
**Describe your proposal**
We should also show a short info for the creature while hovering. Like in HD. Not only the hero.

**Additional context**

Laserlicht:
fixes #3253 
fixes #1490
fixes #2961

Features:
- Show Battle Stacks Info (with showing lost units + remaining health)
- if more than 3 spells active then in the third slot "..." is shown (then you have to right click stack)
- Fix Bug with frozen blue border in order bar
- Show spell duration (if ui enhancement enabled)
- Show luck morale value (if ui enhancement enabled)

Todo:
- [x] add translatable strings
- [x] show active spells
- [x] show spell duration
- [x] find solution for ATB bug (after right clicking unit it keeps highlighted)
- [x] position on mobile

Current state:
grafik

### DIFF
--- a/client/battle/BattleInterfaceClasses.cpp
+++ b/client/battle/BattleInterfaceClasses.cpp
@@ -504,7 +504,7 @@ void StackInfoBasicPanel::initializeData(const CStack * stack)
 	icons.push_back(std::make_shared<CAnimImage>(AnimationPath::builtin("ILCK22"), luck + 3, 0, 47, 143));
 
 	//extra information
-	labels.push_back(std::make_shared<CLabel>(9, 168, EFonts::FONT_TINY, ETextAlignment::TOPLEFT, Colors::WHITE, std::string("Killed") + ":"));
+	labels.push_back(std::make_shared<CLabel>(9, 168, EFonts::FONT_TINY, ETextAlignment::TOPLEFT, Colors::WHITE, VLC->generaltexth->translate("vcmi.battleWindow.killed") + ":"));
 	labels.push_back(std::make_shared<CLabel>(9, 180, EFonts::FONT_TINY, ETextAlignment::TOPLEFT, Colors::WHITE, CGI->generaltexth->allTexts[389] + ":"));
 
 	labels.push_back(std::make_shared<CLabel>(69, 180, EFonts::FONT_TINY, ETextAlignment::BOTTOMRIGHT, Colors::WHITE, std::to_string(killed)));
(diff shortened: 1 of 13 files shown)


krs0:
related to this PR. Turn indicator in ATB would help with keeping track of effects.

IvanSavenko:
Do we need to merge this into 1.4.2? Not sure about relatively large last-minute changes like this one...
Perhaps keep for 1.4.3?

Laserlicht:
> Do we need to merge this into 1.4.2? Not sure about relatively large last-minute changes like this one... Perhaps keep for 1.4.3?

For me 1.4.3 is also okay. Maybe merge directly after 1.4.2 finish. That some people test in in beta "phase".

client/battle/BattleInterfaceClasses.cpp
```diff
@@ -446,6 +446,119 @@ void HeroInfoBasicPanel::show(Canvas & to)
...
+	labelsMultiline.clear();
+
+	initializeData(updatedInfo);
+}
+
+void StackInfoBasicPanel::show(Canvas & to)
+{
+	showAll(to);
+	CIntObject::show(to);
+}
```
IvanSavenko, line 560: Why is this needed?

client/battle/BattleInterfaceClasses.cpp
```diff
@@ -446,6 +446,119 @@ void HeroInfoBasicPanel::show(Canvas & to)
...
+
+StackInfoBasicPanel::StackInfoBasicPanel(const CStack * stack, Point * position, bool initializeBackground)
+	: CIntObject(0)
+{
+	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
+	if (position != nullptr)
+		moveTo(*position);
+
+	if(initializeBackground)
+	{
```
IvanSavenko, line 458: Why is this option needed?

client/battle/BattleInterfaceClasses.cpp
Laserlicht: I don't know. Copied from HeroInfoBasicPanel as base. Do you think we don't need it?

client/battle/BattleInterfaceClasses.cpp
Laserlicht: same as above

dydzio0614: if I got to blind guess then some differences depending on if resolution is 800x600 or bit higher and stuff render bit differently in that case (there was something about rendering outside battle window)

client/battle/BattleInterfaceClasses.cpp
Laserlicht: Tested to remove: It's needed. Without it the rendering is only one frame correct... After that it's overwritten.

by003:
Could VCMI imitate this menu from ERA?
image
Just put ATB in the right of battle screen

Laserlicht:
> Could VCMI imitate this menu from ERA? 

Basically anything is possible if someone program it. But I don't see any great advantage for the time required to do this.

PR #3370, opened 2023-12-23
Ban Armageddon's Blade and Vial of Dragon Blood

DjWarmonger:
Now they need to be explicitely set to `"class" : "SPECIAL"`.


Alucard648:
I think, the special class should be named "class" : "ULTIMATE", same as Grail. And possibility for mods to repace Grail as hidden treasure found via Puzzle Map.

PR #3374, opened 2023-12-23
Default large spellbook

dydzio0614:
To be decided whether to merge or not

Reasoning why it makes sense to be merged:
- we have "enable UI enhancements" by default so large spellbook enhancement would be consistent with this change
- people are used to large spellbook as most of them play with HD mod, because otherwise H3 does not work well on modern windows and it has forced large spellbook with no animation in there
- few upvotes on discord and 0 downvotes while asking if this change should be made

One confusion it may bring:
- some people not reading right-click description and wondering why spellbook animation doesn't work while toggled on (can be eventually mitigated by some sort of graying-out spellbook animation setting when large spellbook is active)

### DIFF
--- a/config/widgets/settings/generalOptionsTab.json
+++ b/config/widgets/settings/generalOptionsTab.json
@@ -72,6 +72,35 @@
 		},
 		{
 			"type" : "verticalLayout",
+			"customType" : "checkboxFake",
+			"position" : {"x": 10, "y": 83},
+			"items" : [
+				{
+					"created" : "desktop"
+				},
+				{},
+				{
+					"created" : "desktop"
+				},
+				{
+					"created" : "desktop"
+				},
+				{},
+				{},
+				{
+					"name": "spellbookAnimationCheckboxPlaceholder"
+				},
+				{
+					"created" : "touchscreen"
+				},
+				{
+					"created" : "mobile"
+				},
+				{}
+			]
+		},
+		{
+			"type" : "verticalLayout",
 			"customType" : "checkbox",
 			"position" : {"x": 10, "y": 83},
 			"items" : [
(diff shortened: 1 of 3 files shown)


misiokles:
I guess 99% of H3 players are used to non-animated large spellbook in hd-mod/HotA.

IvanSavenko:
> people are used to large spellbook as most of them play with HD mod, because otherwise H3 does not work well on modern windows and it has forced large spellbook with no animation in there

Interesting enough is that while HD also enabled quick combat by default, players were still confused when we did same in vcmi. So I wouldn't be so sure that everybody plays with HD mod

dydzio0614:
@IvanSavenko this could be caused by people used to that behavior from previous version of VCMI

IvanSavenko:
possible, but I haven't see anybody saying why game skips combat now / after update. So I think significant part of them are players who have not used HD mod or previous version of vcmi

dydzio0614:
complaints are fairly random, like people did not complain yet that clicking unpassable tile removes hero path (side effect of my fix for VCMI 1.3 for expiring fly spell on turn pass etc.

Laserlicht:
Tested to disable the button: https://github.com/dydzio0614/vcmi/pull/2

Not perfect, but basically works. In the future we should also add a possibility to set CToggleButton disabled (without icon).

Can we do it like this?

dydzio0614:
looks OK for me

config/widgets/settings/generalOptionsTab.json
```diff
@@ -70,6 +70,35 @@
...
+				},
+				{
+					"created" : "touchscreen"
+				},
+				{
+					"created" : "mobile"
+				},
+				{}
+			]
+		},
```
IvanSavenko, line 101: What's this change about?

dydzio0614: change from laserlicht - checkbox becomes fake when large spellbook is toggled on

IvanSavenko: Ah, so we have a layer of fake checkboxes below real ones?

Laserlicht: Exactly. Simplest solution until we have "threestate" checkboxes.

PR #3375, opened 2023-12-23
turn indicator

Laserlicht:
grafik
grafik

### DIFF
--- a/client/battle/BattleInterface.cpp
+++ b/client/battle/BattleInterface.cpp
@@ -58,6 +58,7 @@ BattleInterface::BattleInterface(const BattleID & battleID, const CCreatureSet *
 	, curInt(att)
 	, battleID(battleID)
 	, battleOpeningDelayActive(true)
+	, round(0)
 {
 	if(spectatorInt)
 	{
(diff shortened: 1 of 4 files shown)


dydzio0614:
maybe the small size will not satisfy everybody but good addition overall...

Laserlicht:
> maybe the small size will not satisfy everybody but good addition overall...

Yeah, but disabling for the small is also not good, I think.

The HD way I don't want to choose, because handling the StackQueue is a pain.

IvanSavenko:
Would be nice to add word "turn" somewhere for clarity. If we don't go hd mod approach then perhaps something like this?

Might use a bit more of vertical space though...
зображення

Laserlicht:
I didn't feel it was necessary. Was self-explanatory for me in the game.
The bar has one major disadvantage. It hides the units (the small rectangle also, but not that much).


https://github.com/vcmi/vcmi/assets/13953785/2db951ed-1902-4dce-9cce-234e56298b06

Laserlicht:
Simplified it in small mode:

grafik

I think it's better like this.

krs0:
Its more than enough to work with until a HD mod like will be implemented in the far future.

Thanks for the addition, sometimes I really wanted an indicator when a torn will end.

Laserlicht:
> Is it possible to color the bar separating turns in big version as well? Something more toned down though... like dark grey or something?

Basically it's possible. But I think it doesn't look really good. Tested a little bit in GIMP.

PR #3390, opened 2023-12-25
[Launcher] remove list in settings window

Laserlicht:
Looks much better without it I think:
grafik

And also better for mobile devices.


kambala-decapitator (approved):
I agree

PR #3397, opened 2023-12-27
option for disabling cheats

Laserlicht:
Until we have a better place, it's in turn options:
grafik

@IvanSavenko We should add this to develop and not beta because it breaks savegame, right?

### DIFF
--- a/client/lobby/OptionsTabBase.cpp
+++ b/client/lobby/OptionsTabBase.cpp
@@ -90,6 +90,10 @@ OptionsTabBase::OptionsTabBase(const JsonPath & configPath)
 		CSH->setSimturnsInfo(info);
 	});
 
+	addCallback("setCheatAllowed", [&](int index){
+		CSH->setCheatAllowedInfo(index);
+	});
+
 	addCallback("setTurnTimerAccumulate", [&](int index){
 		TurnTimerInfo info = SEL->getStartInfo()->turnTimerInfo;
 		info.accumulatingTurnTimer = index;
--- a/lib/registerTypes/RegisterTypesLobbyPacks.h
+++ b/lib/registerTypes/RegisterTypesLobbyPacks.h
@@ -54,6 +54,7 @@ void registerTypesLobbyPacks(Serializer &s)
 	s.template registerType<CLobbyPackToServer, LobbySetPlayer>();
 	s.template registerType<CLobbyPackToServer, LobbySetPlayerName>();
 	s.template registerType<CLobbyPackToServer, LobbySetTurnTime>();
+	s.template registerType<CLobbyPackToServer, LobbySetCheatAllowed>();
 	s.template registerType<CLobbyPackToServer, LobbySetSimturns>();
 	s.template registerType<CLobbyPackToServer, LobbySetDifficulty>();
 	s.template registerType<CLobbyPackToServer, LobbyForceSetPlayer>();
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -264,9 +264,11 @@
 	"vcmi.optionsTab.simturnsMin.hover" : "At least for",
 	"vcmi.optionsTab.simturnsMax.hover" : "At most for",
 	"vcmi.optionsTab.simturnsAI.hover" : "(Experimental) Simultaneous AI Turns",
+	"vcmi.optionsTab.cheatAllowed.hover" : "Allow cheats",
 	"vcmi.optionsTab.simturnsMin.help" : "Play simultaneously for specified number of days. Contacts between players during this period are blocked",
 	"vcmi.optionsTab.simturnsMax.help" : "Play simultaneously for specified number of days or until contact with another player",
 	"vcmi.optionsTab.simturnsAI.help" : "{Simultaneous AI Turns}\nExperimental option. Allows AI players to act at the same time as human player when simultaneous turns are enabled.",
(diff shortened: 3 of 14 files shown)


dydzio0614:
i think this is not good GUI place for this option

IvanSavenko:
yeah, probably for develop.
However please bump SERIALIZATION_VERSION constant in CSerializer.h and add version check for your new flag in startinfo.

This way saves from 1.4 can still be loaded in 1.4 (but multiplayer won't be possible)

IvanSavenko:
> i think this is not good GUI place for this option

True, but can't think of any other place for this option at the moment.

client/lobby/OptionsTabBase.cpp
```diff
@@ -90,6 +90,10 @@ OptionsTabBase::OptionsTabBase(const JsonPath & configPath)
 		CSH->setSimturnsInfo(info);
 	});
 
+	addCallback("setCheatAllowed", [&](int index){
+		CSH->setCheatAllowedInfo(index);
+	});
+
```
IvanSavenko, line 96: You also need to set this checkbox to correct value in recreate() method

lib/registerTypes/RegisterTypesLobbyPacks.h
```diff
@@ -54,6 +54,7 @@ void registerTypesLobbyPacks(Serializer &s)
 	s.template registerType<CLobbyPackToServer, LobbySetPlayer>();
 	s.template registerType<CLobbyPackToServer, LobbySetPlayerName>();
 	s.template registerType<CLobbyPackToServer, LobbySetTurnTime>();
+	s.template registerType<CLobbyPackToServer, LobbySetCheatAllowed>();
```
IvanSavenko, line 57: please add to the very end (even if it won't be alphabetical)

IvanSavenko: And if you are curious as to "why":
registerTypesXXX methods generate enumeration of types that can be (de)serialized.

Each type is serialized as (numeric type ID) (serialized type data)
So changing order will lead to a crash if game attempts to load type with wrong type ID

Laserlicht:
> yeah, probably for develop. However please bump SERIALIZATION_VERSION constant in CSerializer.h and add version check for your new flag in startinfo.

Not sure how to check exactly. Haven't found any examples.

IvanSavenko:
> Not sure how to check exactly. Haven't found any examples.

```
if (version >= new value )
    h & newField;
else
    h = defaultValue;
```

dydzio0614:
well, another minor doubts I have:

1. before this feature ai simturns option is isolated and harder to enable by mistake this way
2. there will probably be some simturn behavior flip switch which would be fitting to add somewhere to this screen

Laserlicht:
> well, another minor doubts I have:
> 
>     1. before this feature ai simturns option is isolated and harder to enable by mistake this way
> 
>     2. there will probably be some simturn behavior flip switch which would be fitting to add somewhere to this screen

Yes basically we should have an "other options" screen similar to turn options. But we only have this option yet...

Laserlicht:
Added extra page for additional options:
grafik

Much space for additional improvements. Moved the turn options up.

Also added option for unlimited battle replay (so we have two settings yet).

Missing is a possibility to open this window within bonus selection. Maybe I add this in separate PR later.

IvanSavenko:
I would like to ask our community manager @dydzio0614 whether there were other similar requests before, since this give is more space for potential game options

dydzio0614:
One doubt i have is where we will add button for banning/bidding panel

Another doubt is about what should be serialized or allowed to be changed on the fly the way gameconfig is changeable right now - i think most player requests  would rather fit being parts of gameconfig changeable - hota dimension door mechanics, hota diplomacy nerf, allowing glitch for heroes without army, fixing captain hat unlimited movement points the hota way...

The thing that we need separately anyway is saving last used settings for random map generator, turn options and this menu

Laserlicht:
> One doubt i have is where we will add button for banning/bidding panel

As singleplayer guy: What do you mean? Something like the PvP options from HD mod? I think we can combine it with the Chat button (makes most sense to me). Maybe this one button as threestate switch (Scenario info; Chat; PvP). Or a sub button if chat is enabled.

> Another doubt is about what should be serialized or allowed to be changed on the fly the way gameconfig is changeable right now - i think most player requests would rather fit being parts of gameconfig that is changeable on save reload - hota dimension door mechanics, hota diplomacy nerf, allowing glitch for heroes without army, fixing captain hat unlimited movement points the hota way...

Yeah, but I think some options should be have to predefined and shouldn't be changeable ingame. Especially with multiplayer where the host can set this settings and the player shouldn't able to change it.

dydzio0614:
yeah i mean pvp options from hd mod

dydzio0614:
and i cannot think of anything else than these two options that we have for now - things I listed are probably 2x more important than stuff I will be able to invent for extra options tab...

Laserlicht:
> and i cannot think of anything else than these two options that we have for now - things I listed are probably 2x more important than stuff I will be able to invent for extra options tab...

Yeah, but these two (or one) options are not fitting anywhere else. That was the main problem here. HD uses a popup here for additional options. But I think a page is better.

dydzio0614:
yeah, let it be 2 options for now

Laserlicht:
@IvanSavenko Okay from your side?

IvanSavenko:
Not against it I guess, although have same questions as Dydzio, specifically - that with this we won't have any free space in that screen.

Let me check code once more, but probably looks OK

PR #3399, opened 2023-12-27
ENABLE_STRICT_COMPILATION compilation fixes

Kuxe:
Fixes minor compilation errors when compiling with `ENABLE_STRICT_COMPILATION` using GCC 13.2.1.

### DIFF
--- a/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
+++ b/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
@@ -266,7 +266,7 @@ uint64_t DangerHitMapAnalyzer::enemyCanKillOurHeroesAlongThePath(const AIPath & 
 	int3 tile = path.targetTile();
 	int turn = path.turn();
 
-	auto& info = getTileThreat(tile);
+	const auto& info = getTileThreat(tile);
 	
 	return (info.fastestDanger.turn <= turn && !isSafeToVisit(path.targetHero, path.heroArmy, info.fastestDanger.danger))
 		|| (info.maximumDanger.turn <= turn && !isSafeToVisit(path.targetHero, path.heroArmy, info.maximumDanger.danger));
--- a/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
+++ b/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
@@ -265,11 +265,11 @@ uint64_t DangerHitMapAnalyzer::enemyCanKillOurHeroesAlongThePath(const AIPath & 
 {
 	int3 tile = path.targetTile();
 	int turn = path.turn();
-	const auto & fastestDanger = hitMap[tile.x][tile.y][tile.z].fastestDanger;
-	const auto & maximumDanger = hitMap[tile.x][tile.y][tile.z].maximumDanger;
 
-	return (fastestDanger.turn <= turn && !isSafeToVisit(path.targetHero, path.heroArmy, fastestDanger.danger))
-		|| (maximumDanger.turn <= turn && !isSafeToVisit(path.targetHero, path.heroArmy, maximumDanger.danger));
+	auto& info = getTileThreat(tile);
+	
+	return (info.fastestDanger.turn <= turn && !isSafeToVisit(path.targetHero, path.heroArmy, info.fastestDanger.danger))
+		|| (info.maximumDanger.turn <= turn && !isSafeToVisit(path.targetHero, path.heroArmy, info.maximumDanger.danger));
 }
 
 const HitMapNode & DangerHitMapAnalyzer::getObjectThreat(const CGObjectInstance * obj) const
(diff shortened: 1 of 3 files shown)


AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
```diff
@@ -265,10 +265,11 @@ uint64_t DangerHitMapAnalyzer::enemyCanKillOurHeroesAlongThePath(const AIPath &
 {
 	int3 tile = path.targetTile();
 	int turn = path.turn();
-	const HitMapNode & info = hitMap[tile.x][tile.y][tile.z];
+	const auto & fastestDanger = hitMap[tile.x][tile.y][tile.z].fastestDanger;
+	const auto & maximumDanger = hitMap[tile.x][tile.y][tile.z].fastestDanger;
```
IvanSavenko, line 269: Copy-paste error?
```suggestion
	const auto & maximumDanger = hitMap[tile.x][tile.y][tile.z].maximumDanger;
```

Kuxe: Ugh, embarassing. Yes. Fixed with 61314a36f6b199693ee06ccd237b31fbfe72e31f.

kambala-decapitator:
Tbh I don't understand what the first commit actually fixes. Also, there's a helper method to return threat object, but you're not using it.

The second commit is also not quite clear: if the function always sets the variable, then the fix isn't needed at all. (Although it doesn't hurt)

Kuxe:
> Tbh I don't understand what the first commit actually fixes. Also, there's a helper method to return threat object, but you're not using it. **Edit**: read conversation in the linked issue: it only confirms that this "fix" only silences compiler but doesn't actually fix anything.
> 
> The second commit is also not quite clear: if the function always sets the variable, then the fix isn't needed at all. (Although it doesn't hurt)

First commit just silences compilation warning (or error when compiling with `ENABLE_STRICT_COMPILATION`). Semantically nothing has changed. This is fine if the warning is a false positive. Is it not a false positive? Also I'm on cellphone atm, will update PR later with this helper you mention if it makes sense. 

The second commit actually fixes an error. The uninitialized variable `var` is assigned to some other variable within the serialization-call just below its declaration. 

Just to motivate these changes, they're hard errors with this workflow: `cmake - -preset linux-release-gcc & & cmake - -build - -preset linux-release-gcc`

kambala-decapitator:
1st: will using `getTileThreat` also result in the warning?

2nd: oh yes, you're right. At first I thought that the function below writes to the value, not reads it.

IvanSavenko:
> The second commit actually fixes an error. The uninitialized variable var is assigned to some other variable within the serialization-call just below its declaration.

Not exactly. This code is only used for *de*serialization, so that variable is never read, only written to. However since compiler is not aware of this we got a warning.

Kuxe:
> 1st: will using `getTileThreat` also result in the warning?
I never ran into the warning in `getTileThreat` so chances are that'll do it. I'll update this PR later today or tomorrow.

Kuxe:
> Not exactly. This code is only used for _de_serialization, so that variable is never read, only written to. However since compiler is not aware of this we got a warning.
`var` _is_ read and assigned to another variable whose name is not readily available to me atm. Do you mean that said variable is never used during deserialization? In that case it should be enough to guard the assignment by an `if(!saving)`, just like the bottom assignment in the function.

IvanSavenko:
But it does? There is `if(!handler.saving)` few lines above, line 216.
However since this is not compile-time constant compiler is not aware of it

kambala-decapitator:
Setting the var to 0 with comment saying that this is only to silence warning would be fine anyway

IvanSavenko (approved):

Kuxe:
Using `getTileThreat` seems to quench the warning, commited [88ec4fa](https://github.com/vcmi/vcmi/pull/3399/commits/88ec4fa0e7fd97a9add0bdfa701b0bc9669f0713).

dydzio0614 (approved):

Alexander-Wilms:
Should a CI matrix entry be a added that uses gcc-13?

Available gcc versions:
https://github.com/actions/runner-images/pull/8562/files
Selecting the version to be used:
https://github.com/actions/runner-images/issues/2503

AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
```diff
@@ -265,8 +265,9 @@ uint64_t DangerHitMapAnalyzer::enemyCanKillOurHeroesAlongThePath(const AIPath &
 {
 	int3 tile = path.targetTile();
 	int turn = path.turn();
-	const HitMapNode & info = hitMap[tile.x][tile.y][tile.z];
 
+	auto& info = getTileThreat(tile);
```
kambala-decapitator, line 269:
```suggestion
	const auto& info = getTileThreat(tile);
```

kambala-decapitator (approved):

Kuxe:
> Should a CI matrix entry be a added that uses gcc-13?
> 
> Available gcc versions: https://github.com/actions/runner-images/pull/8562/files Selecting the version to be used: [actions/runner-images#2503](https://github.com/actions/runner-images/issues/2503)

Created https://github.com/vcmi/vcmi/issues/3407 to this end. I think modifying & verifying new CI setup is out of scope of this PR.

PR #3400, opened 2023-12-27
Enable ccache option for linux

Kuxe:
# Enable ccache option for linux
## Rationale
Previously, developers had to know that for linux they have to pass `ccache` manually to `CMAKE_C(XX)_COMPILER_LAUNCHER` but for Windows and XCode it was sufficient to pass `ENABLE_CCACHE`. Now a dev just needs to care about passing `ENABLE_CCACHE` regardless of project generator / platform; which I think is a nice simplification

`ENABLE_CCACHE` is no longer generator / platform dependant option - the option always exists.

## Testing
Tested locally on linux, seems to behave as intended - when option is enabled `ccache` is used and otherwise it is not (verified by zeroing stats via `ccache --zero-stats` and `ccache --show-stats`).

## Documentation
Updated by searching for `ccache` and updating docs where needed

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -244,7 +244,7 @@ jobs:
 
     - name: CMake Preset with ccache
       run: |
-        cmake -DCMAKE_CXX_COMPILER_LAUNCHER=ccache --preset ${{ matrix.preset }}
+        cmake -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }}
 
     - name: Build Preset
       run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,10 +81,7 @@ endif()
 
 option(ENABLE_CCACHE "Speed up recompilation by caching previous compilations" OFF)
 if(ENABLE_CCACHE)
-	find_program(CCACHE ccache)
-	if(CCACHE-NOTFOUND)
-		message(FATAL_ERROR "'ccache' could not be found; install it or set ENABLE_CCACHE=OFF.")
-	endif()
+	find_program(CCACHE ccache REQUIRED)
 endif()
 
 # On Linux, use ccache via CMAKE_CXX_COMPILER_LAUNCHER.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,21 +95,18 @@ if(ENABLE_CCACHE AND LINUX)
 endif()
 
 if(ENABLE_CCACHE AND (CMAKE_GENERATOR STREQUAL "Xcode"))
-	find_program(CCACHE ccache REQUIRED)
-	if(CCACHE)
-		# https://stackoverflow.com/a/36515503/2278742
-		# Set up wrapper scripts
-		configure_file(xcode/launch-c.in   xcode/launch-c)
-		configure_file(xcode/launch-cxx.in xcode/launch-cxx)
-		execute_process(COMMAND chmod a+rx
-												"${CMAKE_BINARY_DIR}/xcode/launch-c"
-												"${CMAKE_BINARY_DIR}/xcode/launch-cxx")
-		# Set Xcode project attributes to route compilation through our scripts
-		set(CMAKE_XCODE_ATTRIBUTE_CC         	"${CMAKE_BINARY_DIR}/xcode/launch-c")
-		set(CMAKE_XCODE_ATTRIBUTE_CXX        	"${CMAKE_BINARY_DIR}/xcode/launch-cxx")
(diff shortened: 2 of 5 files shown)


Kuxe:
Locally I couldn't get this PR to compile without https://github.com/vcmi/vcmi/pull/3399, so let's get that merged first before looking into this PR

.github/workflows/github.yml
```diff
@@ -244,7 +244,7 @@ jobs:
 
     - name: CMake Preset with ccache
       run: |
-        cmake -DCMAKE_CXX_COMPILER_LAUNCHER=ccache --preset ${{ matrix.preset }}
+        cmake -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }}
```
Alexander-Wilms, line 247: I didn't know the `:BOOL` syntax, is there a benefit to it?

Kuxe: I think just using "ON" would work too, but when the type is annotated CMake will barf if user makes a typo when typing ON or OFF.

CMakeLists.txt
```diff
@@ -79,28 +79,34 @@ if(NOT APPLE_IOS AND NOT ANDROID)
...
 
-# On Linux, use -DCMAKE_CXX_COMPILER_LAUNCHER=ccache instead.
-# The XCode and MSVC builds each require some more configuration, which is enabled by the following option:
-if(MSVC OR (CMAKE_GENERATOR STREQUAL "Xcode"))
-  option(ENABLE_CCACHE "Speed up recompilation by caching previous compilations" OFF)
+option(ENABLE_CCACHE "Speed up recompilation by caching previous compilations" OFF)
+if(ENABLE_CCACHE)
+	find_program(CCACHE ccache)
+	if(CCACHE-NOTFOUND)
+		message(FATAL_ERROR "'ccache' could not be found; install it or set ENABLE_CCACHE=OFF.")
```
kambala-decapitator, line 86: If you really want to do it, just add REQUIRED to find_program

kambala-decapitator: in my view, if you're passing ENABLE_CCACHE=ON explicitly, you know what you're doing and what you're expecting to happen :)

Kuxe: You don't necessarily pass it explicitly, you could be using a preset that passes it. Currently all such presets are explicitly suffixed with `_ccache` so my argument is not very strong obviously :), but I don't see any problems with keeping it the way it is - do you?

IvanSavenko: +1

Why reinvent the wheel? What's so special about ccache needs separate handling?

Let's assume that CMake error reporting is bad - why we only produce "nice" error message for ccache? Why we are not using same approach for our library dependencies?

Kuxe: I think any dependency that is enabled by a cmake option should refer to that option in case of error. It's very little risk involved with having an error message with helpful info. I also disagree with the sentiment that it's reinventing the wheel if a custom error message contains more information than the default.

Regardless, I'm new on the project and I value confirming to project standards e.g using REQUIRED in favor of custom error messages, I'll make a new commit with `REQUIRED` instead :)

IvanSavenko: In that case - upstream bug. Please send report to CMake team.

No need to add custom behavior that we need to support. I also don't see any other project similar to vcmi that uses something other than REQUIRED flag to report missing libraries.

kambala-decapitator: > In that case - upstream bug. Please send report to CMake team.

Not sure what you're referring to. @Kuxe means that it's kind of better to show custom message if the find error depends on a project option, I don't see any cmake bug here.

IvanSavenko: What I am saying is that "show human readable message if required library is not found" should not be something we do on our side and is problem with CMake.

Likely not the one that they will even acknowledge, sure, but not something that should be done on our side.

Kuxe: But CMake can't possibly know if a `find_program` is guarded by an option. The best CMake can do is just to tell that the program couldn't be found, but we shouldn't expect CMake to be able to infer what option was used to guard the `find_program` call to begin with.

```
if(ENABLE_<myprogram>)
  // What user-feedback can we provide if  this fails?
  // Could just fallback to telling user that <myprogram> wasn't found (default behaviour and current behavior as of 25be15a)
  // or we could say that <myprogram> wasn't found plus mention that it tried to be found to begin with due to setting ENABLE_<myprogram> to 'on'
  // regardless this is not something that CMake would be able to deduce on it's own
  // unless there was some "find_program_if"...
  find_program(MYPROGRAM myprogram REQUIRED)
endif()

kambala-decapitator (approved):

PR #3404, opened 2023-12-29
Update linux desktop files

cewbdex:
I believe keywords are from [1.1 and above](https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#idm46519897531200). Singlemainwindow needs 1.5 and it is supposed to tell the DE to not launch another windows of the application (game should not be run multiple times, but editor may?). Other changes are translation. If you do not like the changes, I can revert them and send just the translation.


dydzio0614:
why game should not be ran multiple times? what about testing multiplayer lobby on one PC

Alexander-Wilms (commented):
Some suggestions regarding the English strings and German translations

cewbdex:
@dydzio0614 how often does normal user need this? I might be wrong, but I see this as a developer or modder thing. For those DE will not deny multiple instances, it will just not launch another if you click the icon, etc.

@Alexander-Wilms 
Thanks for suggestions, I agree.

dydzio0614:
i did not see any complaints about this option existing currently, and i dont know why developers would need any needless "extra steps" for something that is not broken in first place

cewbdex:
@dydzio0614 
I wanted to revert as you wished, but I didn't have a time before the merge. If this new setting will be problematic for you, you can open PR from the same branch.

PR #3408, opened 2023-12-31
Battle ai fix

nullkiller:
Fix shooters not shooting when broken arrow
Fix casting of fire elementals spell


nullkiller:
Added new fix related to Imps having min damage as 0 because of 50% damage reduce or something. Sometimes happens in WOG. A bit weird case but still.

PR #3409, opened 2023-12-31
Fix headless

nullkiller:
It does not looks like a hotfix so will be retargeted later to develop
Fixes headless mode and a set of fixes for AI. So far:
- attempt to fix parallel write access to hero roles map
- fix counting mana cost of Fly a few times
- added value for pyramid and cyclops stockpile

### DIFF
--- a/AI/Nullkiller/Pathfinding/AINodeStorage.h
+++ b/AI/Nullkiller/Pathfinding/AINodeStorage.h
@@ -41,11 +41,19 @@ namespace AIPathfinding
 	const int CHAIN_MAX_DEPTH = 4;
 }
 
+enum DayFlags : ui8
+{
+	NONE = 0,
+	FLY_CASTED = 1,
+	WATER_WALK_CASTED = 2
+};
+
 struct AIPathNode : public CGPathNode
 {
 	uint64_t danger;
 	uint64_t armyLoss;
-	int32_t manaCost;
+	int16_t manaCost;
+	DayFlags dayFlags;
 	const AIPathNode * chainOther;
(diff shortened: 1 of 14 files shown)


DjWarmonger:
AI should already consider Stockpile for its resources, and I saw it attack Stockpile at least one. Still very rare and possible only when there are no better targets.

Also there is no point to hardcode Pyramid since we already have common interface for Banks.

nullkiller:
> AI should already consider Stockpile for its resources, and I saw it attack Stockpile at least one. Still very rare and possible only when there are no better targets.
> 
> Also there is no point to hardcode Pyramid since we already have common interface for Banks.

Yes it will if there is a need for such resources. In general there is always such need but may be of different strength so there are cases when there is no need and priority will be 0. Now it will have small priority and in case there is no other targets it will still attack stockpile.

nullkiller:
Pyramid is hardcoded because it contains spells. My generic interface towards banks can not handle spells

AI/Nullkiller/Pathfinding/AINodeStorage.h
```diff
@@ -41,11 +41,19 @@ namespace AIPathfinding
 	const int CHAIN_MAX_DEPTH = 4;
 }
 
+enum DayFlags : ui8
+{
+	NONE = 0,
+	FLY_CASTED = 1,
+	WATER_WALK_CASTED = 2
```
Alexander-Wilms, line 48:
```suggestion
	FLY_CAST = 1,
	WATER_WALK_CAST = 2
```

PR #3421, opened 2024-01-02
Add 'windows-mingw-release' CMake preset

Kuxe:
Adds a 'windows-mingw-release'-preset (configure, build and test).

Two commits [8846fba](https://github.com/vcmi/vcmi/pull/3421/commits/8846fba1fcec461c33c8649c1aef3f4e215d9513) and [100664e](https://github.com/vcmi/vcmi/pull/3421/commits/100664e019c37c85401df6744e85ee64aa35c611) fixes compilation errors using said preset.

Full disclosure: [100664e](https://github.com/vcmi/vcmi/pull/3421/commits/100664e019c37c85401df6744e85ee64aa35c611) is _not_ tested locally. I haven't gotten around to actually running locally built binaries yet. Let me know if you want me to test this locally, or if you were able to test it yourself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,9 +84,9 @@ if(ENABLE_CCACHE)
 	find_program(CCACHE ccache REQUIRED)
 endif()
 
-# On Linux, use ccache via CMAKE_CXX_COMPILER_LAUNCHER.
+# On Linux and MinGW, use ccache via CMAKE_CXX_COMPILER_LAUNCHER.
 # The XCode and MSVC builds each require some more configuration further down.
-if(ENABLE_CCACHE AND LINUX)
+if(ENABLE_CCACHE AND (LINUX OR MINGW))
 	set(CMAKE_C_COMPILER_LAUNCHER "ccache")
 	set(CMAKE_CXX_COMPILER_LAUNCHER "ccache")
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,9 +84,8 @@ if(ENABLE_CCACHE)
 	find_program(CCACHE ccache REQUIRED)
 endif()
 
-# On Linux and MinGW, use ccache via CMAKE_CXX_COMPILER_LAUNCHER.
 # The XCode and MSVC builds each require some more configuration further down.
-if(ENABLE_CCACHE AND (LINUX OR MINGW))
+if(ENABLE_CCACHE)
 	set(CMAKE_C_COMPILER_LAUNCHER "ccache")
 	set(CMAKE_CXX_COMPILER_LAUNCHER "ccache")
 endif()
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -110,6 +110,15 @@
             }
         },
         {
+            "name": "windows-mingw-release",
+            "displayName": "Windows x64 MinGW Release",
+            "description": "VCMI Windows Ninja using MinGW",
+            "inherits": "default-release",
+            "cacheVariables": {
+                "CMAKE_BUILD_TYPE": "Release"
+            }
+        },
+        {
             "name": "windows-msvc-release",
             "displayName": "Windows x64 RelWithDebInfo",
             "description": "VCMI RelWithDebInfo build",
@@ -325,6 +334,11 @@
             "inherits": "default-release"
         },
         {
+            "name": "windows-mingw-release",
+            "configurePreset": "windows-mingw-release",
+            "inherits": "default-release"
+        },
+        {
             "name": "windows-msvc-release",
             "configurePreset": "windows-msvc-release",
             "inherits": "default-release"
@@ -408,6 +422,11 @@
         {
             "name": "macos-ninja-release",
             "configurePreset": "macos-ninja-release",
+            "inherits": "default-release"
+        },
+        {
+            "name": "windows-mingw-release",
+            "configurePreset": "windows-mingw-release",
             "inherits": "default-release"
         },
         {
--- a/client/render/Colors.cpp
+++ b/client/render/Colors.cpp
@@ -29,11 +29,11 @@ std::optional<ColorRGBA> Colors::parseColor(std::string text)
 {
 	std::smatch match;
 	std::regex expr("^#([0-9a-fA-F]{6})$");
-	ui8 rgb[3] = {0, 0, 0};
+	std::vector<ui8> rgb;
+	rgb.reserve(3);
 	if(std::regex_search(text, match, expr))
 	{
-		std::string tmp = boost::algorithm::unhex(match[1].str()); 
-		std::copy(tmp.begin(), tmp.end(), rgb);
+		boost::algorithm::unhex(match[1].str(), std::back_inserter(rgb)); 
 		return ColorRGBA(rgb[0], rgb[1], rgb[2]);
 	}
 
@@ -42,8 +42,7 @@ std::optional<ColorRGBA> Colors::parseColor(std::string text)
 	for(auto & color : colors) {
 		if(boost::algorithm::to_lower_copy(color.first) == boost::algorithm::to_lower_copy(text))
 		{
-			std::string tmp = boost::algorithm::unhex(color.second.String()); 
-			std::copy(tmp.begin(), tmp.end(), rgb);
+			boost::algorithm::unhex(color.second.String(), std::back_inserter(rgb)); 
 			return ColorRGBA(rgb[0], rgb[1], rgb[2]);
 		}
 	}
--- a/docs/developers/Building_Windows.md
+++ b/docs/developers/Building_Windows.md
@@ -114,8 +114,10 @@ Extract `ccache` to a folder of your choosing, add the folder to the `PATH` envi
 ## Compile VCMI with MinGW via MSYS2
 - Install MSYS2 from https://www.msys2.org/
 - Start the `MSYS MinGW x64`-shell
-- Install dependencies: `pacman -S mingw-w64-x86_64-SDL2 mingw-w64-x86_64-SDL2_image mingw-w64-x86_64-SDL2_mixer mingw-w64-x86_64-SDL2_ttf mingw-w64-x86_64-boost mingw-w64-x86_64-gcc mingw-w64-x86_64-ninja mingw-w64-x86_64-qt5-base mingw-w64-x86_64-qt5-tools`
+- Install dependencies: `pacman -S mingw-w64-x86_64-SDL2 mingw-w64-x86_64-SDL2_image mingw-w64-x86_64-SDL2_mixer mingw-w64-x86_64-SDL2_ttf mingw-w64-x86_64-boost mingw-w64-x86_64-gcc mingw-w64-x86_64-ninja mingw-w64-x86_64-qt5-static`
 - Generate and build solution from VCMI-root dir: `cmake --preset windows-mingw-release && cmake --build --preset windows-mingw-release`
+
+**NOTE:** This will link Qt5 statically to `VCMI_launcher.exe` and `VCMI_Mapeditor.exe`. See [PR #3421](https://github.com/vcmi/vcmi/pull/3421) for some background.
 
 # Create VCMI installer (This step is not required for just building & development)
 
--- a/docs/developers/Building_Windows.md
+++ b/docs/developers/Building_Windows.md
@@ -111,6 +111,12 @@ Extract `ccache` to a folder of your choosing, add the folder to the `PATH` envi
 - Right click on `BUILD_ALL` project. This `BUILD_ALL` project should be in `CMakePredefinedTargets` tree in Solution Explorer.
 - VCMI will be built in `%VCMI_DIR%/build/bin` folder!
 
+## Compile VCMI with MinGW via MSYS2
+- Install MSYS2 from https://www.msys2.org/
+- Start the `MSYS MinGW x64`-shell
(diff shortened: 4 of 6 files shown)


CMakeLists.txt
```diff
@@ -84,9 +84,9 @@ if(ENABLE_CCACHE)
 	find_program(CCACHE ccache REQUIRED)
 endif()
 
-# On Linux, use ccache via CMAKE_CXX_COMPILER_LAUNCHER.
+# On Linux and MinGW, use ccache via CMAKE_CXX_COMPILER_LAUNCHER.
 # The XCode and MSVC builds each require some more configuration further down.
-if(ENABLE_CCACHE AND LINUX)
+if(ENABLE_CCACHE AND (LINUX OR MINGW))
```
kambala-decapitator, line 89: iirc there's a `GNU` check

CMakePresets.json
```diff
@@ -127,6 +127,15 @@
                 "CMAKE_CXX_COMPILER": "/usr/bin/g++"
             }
         },
+        {
+            "name": "windows-mingw-release",
```
kambala-decapitator, line 131: not sure why you added it if it doesn't actually run on CI. Also it should use Conan to use pre-built binaries.

CMakeLists.txt
Kuxe: I think that'd include installations of GCC on osx, which is not something I've tested

CMakePresets.json
Kuxe: Using CMake presets have some advantages over old-school CMake with flags. Most important advantage is IDE support in VSCode and Visual Studio. So it has value even if CI doesn't run preset.

Not all build configurations use conan, right? Dependencies can easily be installed via MSYS2's `pacman`

CMakeLists.txt
kambala-decapitator: which is not something anyone has ever tested :)

CMakePresets.json
kambala-decapitator: then this deserves a documentation how to use MinGW without Conan (i.e. how to obtain dependencies): point to msys2, list which packages must be installed etc.

with the current PR we'll have a preset that nobody knows how to use

Kuxe: good point, I'll make a new commit that updates documentation on how to build with this preset

CMakeLists.txt
Kuxe: yep, so let's keep `MinGW`

Kuxe:
FYI I also have a `windows-mingw-debug`-preset locally in my `CMakeUserPresets.json`. Let me know if I should carry it over to `CMakePresets.json`

Kuxe:
Running into a crash when starting vcmiserver:

`CObjectClassesHandler.cpp:181`:
```
obj->objects[index] = object;
```
`index`is 20 but size of `obj->objects`is 6. The crash is preceeded by this warning.
```
Object wog:creatureBank.grotto - attempt to load object with preset index! This option is reserved for built-in mod
```

Added to `CObjectClassesHandler::loadSubObject`
```
	if(index > obj->objects.size())
	{
		logMod->error("Object %s:%s.%s - index out of bounds (%i > %i)!", obj->modScope, obj->handlerName, identifier, index, obj->objects.size());
		return;
	}
```
and I seem to be hitting this for several objects, see attached log. 
[log.txt](https://github.com/vcmi/vcmi/files/13814780/log.txt)


Could this be local setup issues? Or is this a known issue? Didn't find any GitHub issue that resembles this.

kambala-decapitator:
> FYI I also have a `windows-mingw-debug`-preset locally in my `CMakeUserPresets.json`. Let me know if I should carry it over to `CMakePresets.json`

and how does it differ from release build?

> I can also start the game now, although I am greeted by a "TLS Initialziation Error" on starting `VCMI_launcher.exe`. I don't get this error when running 1.4.2 release. Investigating...

I believe your Qt version was built without openssl support

CMakeLists.txt
kambala-decapitator: well, what about Makefile or Ninja generator on macOS regardless of the compiler? They should also work by simply specifying the "compiler launcher".

I'd rather just add a check for those generators here (maybe excluding Windows OS, not sure how ccache works there in command-line builds).

CMakePresets.json
kambala-decapitator: > good point, I'll make a new commit that updates documentation on how to build with this preset
> 
> **EDIT:** Added with [a78a3b3](https://github.com/vcmi/vcmi/pull/3421/commits/a78a3b3e4fb505966c07de50e4b2598b32319a40)

thanks!

Kuxe:
> > FYI I also have a `windows-mingw-debug`-preset locally in my `CMakeUserPresets.json`. Let me know if I should carry it over to `CMakePresets.json`
> 
> and how does it differ from release build?

It passes `CMAKE_BUILD_TYPE="debug"` instead of `CMAKE_BUILD_TYPE="release"` 

> > I can also start the game now, although I am greeted by a "TLS Initialziation Error" on starting `VCMI_launcher.exe`. I don't get this error when running 1.4.2 release. Investigating...
> 
> I believe your Qt version was built without openssl support

Thanks! I'll rebuild with openssl and do some more manual smoke testing to see if there's anything obviously broken with MinGW build.

kambala-decapitator:
> It passes `CMAKE_BUILD_TYPE="debug"` instead of `CMAKE_BUILD_TYPE="release"`

such stuff can be passed on the command line, no need to create another preset for it

IvanSavenko:
> Running into a crash when starting vcmiserver:
> 
> `CObjectClassesHandler.cpp:181`:
> 
> ```
> obj->objects[index] = object;
> ```
> 
> `index` is 20 but size of `obj->objects` is 6. The crash is preceded by this warning.

Huh, interesting find. I did have a suspicion that we have memory corruption somewhere in CObjectClassesHandler on some systems - due to reported crashes. Might be caused by such outdated wog mod. 

Noted, will fix. Should not be just a warning in log.

CMakeLists.txt
Kuxe: Not sure what you mean by
>maybe excluding Windows OS

this PR _is_ for Windows, using the `gcc` port [mingw-w64](https://www.mingw-w64.org/) that is used on host-os Windows targeting Windows (not cross-compilation). I know mingw can be used to cross-compile on Linux targeting Windows, but that's unrelated to this PR. I could rewrite it into
```
if(ENABLE_CCACHE AND (LINUX OR (WINDOWS AND MINGW)))
```

so I don't wind up unintentionally supporting ccache for cross-compilation.. but I thought that was a nice side-effect so I just skipped checking for `WINDOWS`. 

> well, what about Makefile or Ninja generator on macOS regardless of the compiler?
It doesn't matter if you use Makefile or Ninja generator. This PR is enabling ccache for `windows-mingw-release`, both Makefile and Ninja generator. 

I just now found out there's no `GNU`-variable in CMake (I also thought it did exist), at least not listed here: https://cmake.org/cmake/help/latest/manual/cmake-variables.7.html

I'm all for supporting other platforms with ccache :), but it's out of scope of this PR

Kuxe:
> > It passes `CMAKE_BUILD_TYPE="debug"` instead of `CMAKE_BUILD_TYPE="release"`
> 
> such stuff can be passed on the command line, no need to create another preset for it

Sure, most things can be done command-line. But presets allow for IDE integrations, which I think is handy :). In gif below I select one configure preset out of all available presets and then start configuring it. Fuzzy-search, visually listing available presets etc. Things you usually don't get in your terminal.
Code_4b8V1AWrdt

For me personally I can have the debug preset in my `CMakeUserPresets.json`. Let's keep it there until the unlikely event comes that someone requests a debug preset for mingw.

Kuxe:
Here are the solutions to the aforementioned OpenSSL problem:

**a)** Use `pacman -S mingw-w64-x86_64-qt5-base` that loads openssl at runtime (not loadtime nor compile time) and copy `libssl-3-x64.dll, libcrypto-3-x64.dll` (from [firadaemons OpenSSL binaries 3.1.4](https://kb.firedaemon.com/support/solutions/articles/4000121705-openssl-3-1-3-0-and-1-1-1-binary-distributions-for-microsoft-windows) or elsewhere) into `.\out\build\windows-mingw-debug-kuxe\bin`. Would probably require some special handling with `cpack`.

**b)** Use `pacman -S mingw-w64-x86_64-qt5-base-static`. This will statically link qt5 into `VCMI_launcher.exe` and `VCMI_mapeditor.exe`. I wanted to check how much this affected binary size:

**Release build**
|file|static|static (stripped)|non-static|non-static (stripped)|
|-----|-----|-----|-----|-----|
|VCMI_launcher.exe|`38669kb`|`21880kb`|`1798kb`|`1222kb`|
|VCMI_mapeditor.exe|`37164kb`|`21253kb`|`2698kb`|`1722kb`|

**Debug build**
|file|static|static (stripped)|non-static|non-static (stripped)|
|-----|-----|-----|-----|-----|
|VCMI_launcher.exe|`74258kb`|`21620kb`|`31114kb`|`1015kb`|
|VCMI_mapeditor.exe|`125748kb`|`20891kb`|`84883kb`|`1345kb`|

**c)** Build Qt5 from source

**d)** Use conan on Windows to get Qt5 with statically linked OpenSSL



I'll go with option `b)` for now, as there are no plans to use `MinGW`-build for shipping so larger binary sizes is OK and it's easier to maintain than **a)**. **c)** and **d)** will take much more effort to get it right.

kambala-decapitator:
> But presets allow for IDE integrations, which I think is handy

right, I always forget about those IDEs that work with CMake directly :)

But isn't it possible to define custom variables right inside the IDE? For example, Qt Creator has such feature iirc

> For me personally I can have the debug preset in my `CMakeUserPresets.json`. Let's keep it there until the unlikely event comes that someone requests a debug preset for mingw.

I agree


> Here are the solutions to the aforementioned OpenSSL problem:

thanks a lot for the investigation! Static libs are fine.

P.S. OpenSSL is needed only for the launcher to make https network calls

CMakeLists.txt
kambala-decapitator: my main remark is that instead of adding every single platform or compiler here we should instead check the CMake generator that's used. I definitely know that ccache works through the standard compiler launcher variables for Makefile and Ninja generators, and definitely neither for Xcode nor for Visual Studio (that's already handled separately).
 
> this PR _is_ for Windows

right, was thinking about msvc :)

kambala-decapitator (approved):

Kuxe:
> But isn't it possible to define custom variables right inside the IDE? For example, Qt Creator has such feature iirc

In VSCode I think that's possible to do via `"cmake.configureArgs": [<args goes here>]` but it's not possible to have separate args for special "types" of configurations / builds, e.g only passing `-g` in some situations. It's possible to just edit the workspace-file over and over again but I think it's more ergonomic to just stuff whatever arguments to cmake you want in a preset and select the right preset

kambala-decapitator:
> I think it's more ergonomic to just stuff whatever arguments to cmake you want in a preset and select the right preset

sure, makes complete sense

CMakeLists.txt
Kuxe: > my main remark is that instead of adding every single platform or compiler here we should instead check the CMake generator that's used.

Ah, now I see what you mean. I'll make a new commit.

Kuxe:
[491dfd8](https://github.com/vcmi/vcmi/pull/3421/commits/491dfd85021752726809860de8759eeedb718f32) removes the platform checks. I don't even check if generator is ninja or make, since CMake won't do anything if other generators are used (see https://cmake.org/cmake/help/latest/prop_tgt/LANG_COMPILER_LAUNCHER.html#prop_tgt:%3CLANG%3E_COMPILER_LAUNCHER) - so checking if generator is ninja or make is redundant.

pppbb:
I don't know it is problem on my side but savegames(in vcmi explorer) don't have time signatures when trying to load them.
Any help?

kambala-decapitator:
> I don't know it is problem on my side but savegames(in vcmi explorer) don't have time signatures when trying to load them.
> 
> Any help?
> 
> 

How is it related to this PR? Please ask on discord for help or open an issue

pppbb:
> How is it related to this PR? Please ask on discord for help or open an issue

I have compiled vcmi with this preset using Msys2. I followed instructions carefully.
@kambala-decapitator  You don't have the issue? OK i'll open the issue.

kambala-decapitator:
> > How is it related to this PR? Please ask on discord for help or open an issue
> 
> 
> 
> I have compiled vcmi with this preset using Msys2. I followed instructions carefully.
> 
> @kambala-decapitator  You don't have the issue? OK i'll open the issue.

Pretty sure that save game "signatures" (probably you mean file extensions?) have nothing to do with mingw. Still unclear about your issue, please open a separate one and provide more details (e.g. screenshot).

pppbb:
> Pretty sure that save game "signatures" (probably you mean file extensions?) have nothing to do with mingw. Still unclear about your issue, please open a separate one and provide more details (e.g. screenshot)

sorry for my english. What im trying to say when i choose a savegame there is no time in the description only date.

kambala-decapitator:
> > Pretty sure that save game "signatures" (probably you mean file extensions?) have nothing to do with mingw. Still unclear about your issue, please open a separate one and provide more details (e.g. screenshot)
> 
> sorry for my english. What im trying to say when i choose a savegame to load there is no time in the description only date.

please open an issue

PR #3422, opened 2024-01-02
Add 'ENABLE_COLORIZED_COMPILER_OUTPUT' CMake option for Clang/GNU

Kuxe:
On my system compiler output was all in shell foreground color before explicitly enabling colors. With
```
-DENABLE_COLORIZED_COMPILER_OUTPUT=on
```
there is now colorized output instead, helps readability.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,7 @@ option(ENABLE_MULTI_PROCESS_BUILDS "Enable /MP flag for MSVS solution" ON)
 option(COPY_CONFIG_ON_BUILD "Copies config folder into output directory at building phase" ON)
 option(ENABLE_STATIC_AI_LIBS "Add AI code into VCMI lib directly" ${staticAI})
 
-option(ENABLE_COLORIZED_COMPILER_OUTPUT "Colorize compilation output (Clang/GNU)." OFF)
+option(ENABLE_COLORIZED_COMPILER_OUTPUT "Colorize compilation output (Clang/GNU)." ON)
 
 if(${ENABLE_COLORIZED_COMPILER_OUTPUT})
 	if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -74,6 +74,16 @@ option(ENABLE_MULTI_PROCESS_BUILDS "Enable /MP flag for MSVS solution" ON)
 option(COPY_CONFIG_ON_BUILD "Copies config folder into output directory at building phase" ON)
 option(ENABLE_STATIC_AI_LIBS "Add AI code into VCMI lib directly" ${staticAI})
 
+option(ENABLE_COLORIZED_COMPILER_OUTPUT "Colorize compilation output (Clang/GNU)." OFF)
+
+if(${ENABLE_COLORIZED_COMPILER_OUTPUT})
+	if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+		add_compile_options(-fcolor-diagnostics)
+	elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
```
kambala-decapitator, line 82:
```suggestion
if(ENABLE_COLORIZED_COMPILER_OUTPUT)
	if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
		add_compile_options(-fcolor-diagnostics)
	elseif(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```

CMakeLists.txt
```diff
@@ -74,6 +74,16 @@ option(ENABLE_MULTI_PROCESS_BUILDS "Enable /MP flag for MSVS solution" ON)
 option(COPY_CONFIG_ON_BUILD "Copies config folder into output directory at building phase" ON)
 option(ENABLE_STATIC_AI_LIBS "Add AI code into VCMI lib directly" ${staticAI})
 
+option(ENABLE_COLORIZED_COMPILER_OUTPUT "Colorize compilation output (Clang/GNU)." OFF)
```
kambala-decapitator, line 77: why not set it to ON by default?

kambala-decapitator (approved):

CMakeLists.txt
Kuxe: Just precaution, I'm not sure if there are any negative implications to enabling this? There could be, I just don't know. I don't mind setting it to ON by default and should someone wind up with problems they can file an issue.

Kuxe: Changed default to `ON` with [4f4298e](https://github.com/vcmi/vcmi/pull/3422/commits/4f4298ece872981fd54e7119991dd87ac113b561)

kambala-decapitator (approved):

PR #3423, opened 2024-01-02
Add 'ENABLE_EXPORT_COMPILE_COMMANDS'

Kuxe:
Add CMake option `ENABLE_EXPORT_COMPILE_COMMANDS`. The generated `compile_commands.json` is also symlinked to `CMAKE_BINARY_DIR`. Symlinking helps tooling, e.g `clangd` (which works out-of-the-box with VSCode as of this PR using `llvm-vs-code-extensions.vscode-clangd`), to find `compile_commands.json`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,6 +106,18 @@ if(ENABLE_CCACHE AND (CMAKE_GENERATOR STREQUAL "Xcode"))
 	set(CMAKE_XCODE_ATTRIBUTE_LDPLUSPLUS 	"${CMAKE_BINARY_DIR}/xcode/launch-cxx")
 endif()
 
+option(ENABLE_EXPORT_COMPILE_COMMANDS "Enable export of 'compile_commands.json', used by C++ tooling such as clangd or cppcheck" ON)
+if(ENABLE_EXPORT_COMPILE_COMMANDS)
+	# NOTE: compile_commands.json is generated during configuration-time
+	#       in the build directory (same directory as CMakeCache.txt)
+	#       Many tools look recursively upwards to find compile_commands.json,
+	#       so put a symlink at CMAKE_BINARY_DIR to help tooling.
+	set(CMAKE_EXPORT_COMPILE_COMMANDS on)
+	execute_process(
+		COMMAND cmake -E create_symlink "${CMAKE_BINARY_DIR}/compile_commands.json" "${CMAKE_CURRENT_SOURCE_DIR}/compile_commands.json"
+	)
+endif()
+
 # Allow to pass package name from Travis CI
 set(PACKAGE_NAME_SUFFIX "" CACHE STRING "Suffix for CPack package name")
 set(PACKAGE_FILE_NAME "" CACHE STRING "Override for CPack package filename")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,18 +106,6 @@ if(ENABLE_CCACHE AND (CMAKE_GENERATOR STREQUAL "Xcode"))
 	set(CMAKE_XCODE_ATTRIBUTE_LDPLUSPLUS 	"${CMAKE_BINARY_DIR}/xcode/launch-cxx")
 endif()
 
-option(ENABLE_EXPORT_COMPILE_COMMANDS "Enable export of 'compile_commands.json', used by C++ tooling such as clangd or cppcheck" ON)
-if(ENABLE_EXPORT_COMPILE_COMMANDS)
-	# NOTE: compile_commands.json is generated during configuration-time
-	#       in the build directory (same directory as CMakeCache.txt)
-	#       Many tools look recursively upwards to find compile_commands.json,
-	#       so put a symlink at CMAKE_BINARY_DIR to help tooling.
-	set(CMAKE_EXPORT_COMPILE_COMMANDS on)
-	execute_process(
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -106,6 +106,18 @@ if(ENABLE_CCACHE AND (CMAKE_GENERATOR STREQUAL "Xcode"))
 	set(CMAKE_XCODE_ATTRIBUTE_LDPLUSPLUS 	"${CMAKE_BINARY_DIR}/xcode/launch-cxx")
 endif()
 
+option(ENABLE_EXPORT_COMPILE_COMMANDS "Enable export of 'compile_commands.json', used by C++ tooling such as clangd or cppcheck" ON)
```
kambala-decapitator, line 109: most devs don't need it, set it to OFF

CMakeLists.txt
```diff
@@ -106,6 +106,18 @@ if(ENABLE_CCACHE AND (CMAKE_GENERATOR STREQUAL "Xcode"))
...
+if(ENABLE_EXPORT_COMPILE_COMMANDS)
+	# NOTE: compile_commands.json is generated during configuration-time
+	#       in the build directory (same directory as CMakeCache.txt)
+	#       Many tools look recursively upwards to find compile_commands.json,
+	#       so put a symlink at CMAKE_BINARY_DIR to help tooling.
+	set(CMAKE_EXPORT_COMPILE_COMMANDS on)
+	execute_process(
+		COMMAND cmake -E create_symlink "${CMAKE_BINARY_DIR}/compile_commands.json" "${CMAKE_CURRENT_SOURCE_DIR}/compile_commands.json"
```
kambala-decapitator, line 117: not sure how wise it is to add to the project. If one needs it, then one would pass CMAKE_EXPORT_COMPILE_COMMANDS explicitly and create symlink manually.

also, pretty sure you want the root source dir, better tell it explicitly
```suggestion
		COMMAND cmake -E create_symlink "${CMAKE_BINARY_DIR}/compile_commands.json" "${CMAKE_SOURCE_DIR}/compile_commands.json"
```

kambala-decapitator (changes requested):

CMakeLists.txt
Kuxe: You sure? I mean any developer that would open up VCMI in any text editor that uses clangd would have to explicitly enable exporting compile_commands.json. Currently with this PR clangd works out of the box without any manual intervention. 

I've never encountered any problems with this export and I think the cost&risk of enabling it by default is low.

IvanSavenko:
Why is this needed? Why can't you point IDE to compile_commands.json in build dir?

IvanSavenko:
clangd docs also state that there is autodetection and suggests to symlink it manually:
https://clangd.llvm.org/installation#project-setup

So all you need to do is to set flag in cmake, no need to change configuration: `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=1`

dydzio0614:
I stay silent because i don't care much about these stuff but in case somebody wants my opinion - we currently have bunch more important things to do than polishing cmake fluff.

Kuxe:
> Why is this needed? Why can't you point IDE to compile_commands.json in build dir?

It can be done! But then I have to do that, and anyone else that wants clangd to work has to do that as well. 

> clangd docs also state that there is autodetection and suggests to symlink it manually: https://clangd.llvm.org/installation#project-setup
> 
> So all you need to do is to set flag in cmake, no need to change configuration: `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=1`

I think they're just suggesting how it can be done, I don't think they mean to recommend against having CMake so this work in configure-time. I'm not sure. 

> I stay silent because i don't care much about these stuff but in case somebody wants my opinion - we currently have bunch of more important things to do than polishing cmake fluff.

I'm not gonna do many more CMake PRs. Most stuff can go into `CMakeUserPresets.json`. But I want to give VCMI project a chance to accept these things into `develop` should anyone else benefit from it.

---

I'll update this PR later to solely modify `.gitignore` so that manually symlinking to root dir is still possible but lists file will be untouched.

IvanSavenko:
> I think they're just suggesting how it can be done, I don't think they mean to recommend against having CMake so this work in configure-time. I'm not sure.

They kind of do:
зображення

- use built-in cmake command to generate file
- use SRC/build as output directory to automatically find it.

And again - don't see anything like this in any other c++ projects similar to vcmi on Github.

Kuxe:
> > I think they're just suggesting how it can be done, I don't think they mean to recommend against having CMake so this work in configure-time. I'm not sure.
> 
> They kind of do: зображення
> 
>     * use built-in cmake command to generate file
> 
>     * use SRC/build as output directory to automatically find it.
> 
> 
> And again - don't see anything like this in any other c++ projects similar to vcmi on Github.
I guess if the 'should' is as in rfc2119 then yes, they recommend doing it manually.

About usage on GitHub, depends on what part you refer to, enabling exporting of `compile_commands.json` is done 20k times on GitHub (https://github.com/search?q=%22set%28CMAKE_EXPORT_COMPILE_COMMANDS+on%29%22&type=code). Using `cmake -E` to either copy or symlink it is much rarer (https://github.com/search?q=%22set%28CMAKE_EXPORT_COMPILE_COMMANDS+on%29%22+%22cmake+-E%22+%22compile_commands.json%22&type=code).

Kuxe:
Discarding this PR in favor of just adding `compile_commands.json` to `.gitignore`, in a new PR.

PR #3438, opened 2024-01-05
Select sdl renderer in Launcher

IvanSavenko:
- Failure to create SDL Renderer will now result show message box with error message before crash (not translatable since crash happens before translations are loaded)
- Added option to select renderer in Launcher from OS-specific list of options (as reported by SDL)

### DIFF
--- a/launcher/settingsView/csettingsview_moc.cpp
+++ b/launcher/settingsView/csettingsview_moc.cpp
@@ -271,8 +271,8 @@ void CSettingsView::fillValidResolutionsForScreen(int screenIndex)
 void CSettingsView::fillValidRenderers()
 {
 	// untested on mobile platforms
-	ui->comboBoxRendererType-hide();
-	ui->labelRendererType-hide();
+	ui->comboBoxRendererType->hide();
+	ui->labelRendererType->hide();
 }
 #endif
 
--- a/launcher/settingsView/csettingsview_moc.cpp
+++ b/launcher/settingsView/csettingsview_moc.cpp
@@ -117,6 +117,7 @@ void CSettingsView::loadSettings()
     ui->lineEditAutoSavePrefix->setEnabled(settings["general"]["useSavePrefix"].Bool());
 
 	Languages::fillLanguages(ui->comboBoxLanguage, false);
+	fillValidRenderers();
 
 	std::string cursorType = settings["video"]["cursor"].String();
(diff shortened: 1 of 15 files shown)


launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -163,6 +164,24 @@ void CSettingsView::fillValidScalingRange()
 
 #ifndef VCMI_MOBILE
 
+static QStringList getAvailableRenderingDrivers()
+{
+	SDL_Init(SDL_INIT_VIDEO);
```
kambala-decapitator, line 169: you must link SDL lib to the launcher target in CMake

kambala-decapitator (changes requested):

launcher/settingsView/csettingsview_moc.cpp
```diff
@@ -228,13 +246,36 @@ void CSettingsView::fillValidResolutionsForScreen(int screenIndex)
 
 	ui->comboBoxResolution->blockSignals(false);
 }
+
+void CSettingsView::fillValidRenderers()
+{
+	ui->comboBoxRendererType->blockSignals(true); // avoid saving wrong resolution after adding first item from the list
```
kambala-decapitator, line 252: use QSignalBlocker

launcher/settingsView/csettingsview_moc.cpp
IvanSavenko: We actually already use SDL in launcher and it works. So I presume already linked with SDL

launcher/settingsView/csettingsview_moc.cpp
IvanSavenko: huh, wasn't aware of it. Replaced where applicable.

launcher/settingsView/csettingsview_moc.cpp
IvanSavenko: yep, already in:
```
if (NOT APPLE_IOS AND NOT ANDROID)
	target_link_libraries(vcmilauncher SDL2::SDL2)
endif()
```

kambala-decapitator: Nice, forgot about that. Probably to get supported resolutions.

kambala-decapitator (approved):

PR #3447, opened 2024-01-07
[WIP] 16-players

Kuxe:
**What?**
Add 16-player support in VCMI-engine. This PR does *not* include new maps for 16-players. This PR will make it possible to create 16-player maps and play them. This PR is _not_ about generalizing player-count to N, I do not intend to make code that enumerates from 1 to 16 more elegant.

16-players-vcmieditor

**Why?**

This has been requested on some occasions:
* http://heroescommunity.com/viewthread.php3?TID=43560
* https://forum.vcmi.eu/t/more-then-8-player/5760

Since 1.4.0 released with simultaneous turns it could possibly be fun to participate in 16-player games.

**How?**
I'm fairly new to VCMI and I don't know the source code. I've mechanically searched for "tan", "purple" etc in source code to identify places where team colors were hard-coded and just extended these places with new colors.

--- 

The added colors are:

|Player (1-indexed)|Name|Hex|
|---------|---------|-----|
| 9 | Maroon | 800000 |
| 10 | Navy | 000080 |
| 11 | Brown | 8b4513 |
| 12 | Lime | 00ff00 |
| 13 | Gold | ffd700 |
| 14 | Khaki | f0e68c |
| 15 | Coral | ff7f50 |
| 16 | Olive | 808000 |

These names and hex-values match that of `./config/textColors.json` just to keep some level of consistency going.

I've tried picking colors from `./config/textColors.json` such that
* All colors are mutually distinguishable
* No color blend with background (at least not worse than "green" does with "swamp terrain")
* Avoid any color on the greyscale since that could be confused with "Neutral" towns.

**DONE**
* Add support in vcmieditor
  - Creating a new map with 16 players is possible
  - In-game flags are rendered with correct color
  - Hotkeys CTRL+SHIFT+[1...8] are used for the new factions
  - Saving + Loading seems to work as intended

**TODO**
* Fix all "TODO 16-players:" that I've added
* ~Make it possibly to launch singleplayer - currently asserts at `MapReaderH3m.cpp:398`~ fixed locally. Will do force push later.
* Random crashing on computer 9 ... 16 turn
* Use new resources - currently fallbacking to player red when needed
* Identify & solve cases in GUI that are designed for 8 players but now needs to be modified to be suitable for 16 players
  - Random map generator 
  - Player starting options
  - Thieves guild (including HotA's Cove town variant?)
  - Marketplace
  - Possibly more
* Make a "real" 16-player map (maybe just add underground to 'Warlords!' by copying overworld?) and try it out to identify in-game issues with 16-players.
* ... possibly more stuff that I haven't encountered yet :)


IvanSavenko (changes requested):
- VCMI by default should have Shadow of Death mechanics. All new changes (especially larger ones, like this one) should be option
- This should be implemented as yet another moddable subsystem, similar to (for example) TownHandler, with new colors added via mods
- This requires numerous changes through game UI: rmg (only 8 buttons right now), player starting options (only 8 slots), thieves guild (only 8 "ranks" possible now), marketplace (sending resources only has 7 slots), and likely many else. I also doubt that 16 flags would fit in player list in scenario information (allies / enemies flags)

Kuxe:
> * VCMI by default should have Shadow of Death mechanics. All new changes (especially larger ones, like this one) should be option

From a user POV it is opt-in because users would have to voluntarily create or download +8 player maps. Users that prefer the vanilla experience doesn't have to do anything, and will not observe any changes at all to GUI elements if playing on maps that are 8 players or less. Any GUI changes in lobby should only take effect iff user selects a +8-player map. So lobby GUI, campaigns and maps from Restoration of Erathia, Armageddon's Blade, Shadow of Death and Horn of the Abyss should behave exactly as before this PR (if not, it's a bug in this PR).

The only observable difference would be within vcmieditor.

Is this OK?
 
>     * This should be implemented as yet another moddable subsystem, similar to (for example) TownHandler, with new colors added via mods

Good suggestion, I'm not familiar with moddable subsystems in VCMI but I'll make that happen before marking this PR as ready for review.

>     * This requires numerous changes through game UI: rmg (only 8 buttons right now), player starting options (only 8 slots), thieves guild (only 8 "ranks" possible now), marketplace (sending resources only has 7 slots), and likely many else. I also doubt that 16 flags would fit in player list in scenario information (allies / enemies flags)

I guess that if I create some new moddable subsystem, all GUI elements that may be affected by player count >8 have to check if player count is indeed above 8 and if so alter GUI. Otherwise do not change the current behavior, meaning that users running regular <=8 maps will not observe any difference.

DjWarmonger:
> From a user POV it is opt-in because users would have to voluntarily create or download +8 player maps

RMG is not an issue, since there are no templates with more than 8 players. Even if there were any, they wouldn't be selectable if there are not enough players set in GUI.

I think this feature can be handled in a similiar way that some WoG features defined in EGameSettings:
```
	MODULE_COMMANDERS,
	MODULE_STACK_ARTIFACT,
	MODULE_STACK_EXPERIENCE
```

For instance, WoG commanders are only created if `MODULE_COMMANDERS` option is true.

Kuxe:
> I think this thread is dead. I would suggest for 16 players colors selection a paging. Make button for navigating page 1 and 2.

I intend to continue working with this PR, sporadically

XCOM-HUB:
> I'm all ears for better choices of colors. Also if someone more artsy than me could do flag designs that'd be nice.

Instead of coral, which looks very similar to orange, choose a peach color, and instead of gold, choose gray.

Olegmods:
> I'm all ears for better choices of colors. Also if someone more artsy than me could do flag designs that'd be nice.
Also the black color could be used somewhere. After all, in the H3 beta it was a player color IIRC.

> Instead of coral, which looks very similar to orange, > choose a peach color, and instead of gold, choose > gray.
Gray is probably a bad idea, since it will be mistaken with neutral flag (which is also gray). Probably replacing gold with yellow should be better.

PR #3467, opened 2024-01-09
Fix HotA creature abilities (add bonuses allowing implementing them)

dydzio0614:
As in title - fixes sea dogs, ayssids, sea witches/sorceresses, nix/nix warriors and haspids
Fangarm is wrong in current version of mod but doable without this PR, needs CREATURE_SPELL_POWER 100 and 100% cast chance

+ move text in creature amount label during battle 1 pixel down to better match H3

### DIFF
--- a/lib/battle/DamageCalculator.cpp
+++ b/lib/battle/DamageCalculator.cpp
@@ -10,8 +10,6 @@
 
 #include "StdInc.h"
 
-#include <cfenv>
-
 #include "DamageCalculator.h"
 #include "CBattleInfoCallback.h"
 #include "Unit.h"
@@ -462,7 +460,8 @@ std::vector<double> DamageCalculator::getAttackFactors() const
 		getAttackJoustingFactor(),
 		getAttackDeathBlowFactor(),
 		getAttackDoubleDamageFactor(),
-		getAttackHateFactor()
+		getAttackHateFactor(),
+		getAttackRevengeFactor()
 	};
 }
 
@@ -532,19 +531,14 @@ DamageEstimation DamageCalculator::calculateDmgRange() const
 	for (auto & factor : defenseFactors)
 	{
 		assert(factor >= 0.0);
-		defenseFactorTotal *= ( 1 - std::min(1.0, factor));
+		defenseFactorTotal *= (1 - std::min(1.0, factor));
 	}
 
-	double resultingFactor = std::min(8.0, attackFactorTotal) * std::max( 0.01, defenseFactorTotal);
-
-	//calculated separately since it bypasses cap on bonus damage
-	double revengeFactor = getAttackRevengeFactor();
-	double revengeAdditionalMinDamage = std::round(damageBase.min * revengeFactor);
-	double revengeAdditionalMaxDamage = std::round(damageBase.max * revengeFactor);
+	double resultingFactor = attackFactorTotal * defenseFactorTotal;
 
 	DamageRange damageDealt {
-		std::max<int64_t>( 1.0, std::floor(damageBase.min * resultingFactor + revengeAdditionalMinDamage)),
-		std::max<int64_t>( 1.0, std::floor(damageBase.max * resultingFactor + revengeAdditionalMaxDamage))
+		std::max<int64_t>( 1.0, std::floor(damageBase.min * resultingFactor)),
+		std::max<int64_t>( 1.0, std::floor(damageBase.max * resultingFactor))
 	};
 
 	DamageRange killsDealt = getCasualties(damageDealt);
--- a/lib/battle/DamageCalculator.cpp
+++ b/lib/battle/DamageCalculator.cpp
@@ -526,11 +526,23 @@ DamageEstimation DamageCalculator::calculateDmgRange() const
 
 	double resultingFactor = std::min(8.0, attackFactorTotal) * std::max( 0.01, defenseFactorTotal);
 
-	info.defender->getTotalHealth();
+	double revengeAdditionalMinDamage = 0.0;
+	double revengeAdditionalMaxDamage = 0.0;
+	if(info.attacker->hasBonusOfType(BonusType::REVENGE)) //HotA Haspid ability
+	{
+		int totalStackCount = info.attacker->unitBaseAmount();
+		int currentStackHealth = info.attacker->getAvailableHealth();
+		int creatureHealth = info.attacker->getMaxHealth();
+
+		double revengeFactor = sqrt(static_cast<double>((totalStackCount + 1) * creatureHealth) / (currentStackHealth + creatureHealth) - 1);
+
+		revengeAdditionalMinDamage = std::round(damageBase.min * revengeFactor);
+		revengeAdditionalMaxDamage = std::round(damageBase.max * revengeFactor);
+	}
 
 	DamageRange damageDealt {
-		std::max<int64_t>( 1.0, std::floor(damageBase.min * resultingFactor)),
-		std::max<int64_t>( 1.0, std::floor(damageBase.max * resultingFactor))
+		std::max<int64_t>( 1.0, std::floor(damageBase.min * resultingFactor + revengeAdditionalMinDamage)),
+		std::max<int64_t>( 1.0, std::floor(damageBase.max * resultingFactor + revengeAdditionalMaxDamage))
 	};
 
 	DamageRange killsDealt = getCasualties(damageDealt);
--- a/lib/battle/DamageCalculator.cpp
+++ b/lib/battle/DamageCalculator.cpp
@@ -9,6 +9,9 @@
  */
 
 #include "StdInc.h"
+
+#include <cfenv>
+
 #include "DamageCalculator.h"
 #include "CBattleInfoCallback.h"
 #include "Unit.h"
@@ -133,7 +136,10 @@ int DamageCalculator::getActorAttackIgnored() const
 
 	if(multAttackReduction > 0)
 	{
-		int reduction = std::round(multAttackReduction * getActorAttackBase());
+		int defaultRoundingMode = std::fegetround();
+		std::fesetround(FE_TOWARDZERO);
+		int reduction = std::nearbyint(multAttackReduction * getActorAttackBase());
+		std::fesetround(defaultRoundingMode);
 		return -std::min(reduction,getActorAttackBase());
 	}
 	return 0;
--- a/lib/battle/DamageCalculator.cpp
+++ b/lib/battle/DamageCalculator.cpp
@@ -281,6 +281,20 @@ double DamageCalculator::getAttackHateFactor() const
 	return allHateEffects->valOfBonuses(Selector::subtype()(BonusSubtypeID(info.defender->creatureId()))) / 100.0;
 }
 
+double DamageCalculator::getAttackRevengeFactor() const
+{
+	if(info.attacker->hasBonusOfType(BonusType::REVENGE)) //HotA Haspid ability
+	{
+		int totalStackCount = info.attacker->unitBaseAmount();
+		int currentStackHealth = info.attacker->getAvailableHealth();
+		int creatureHealth = info.attacker->getMaxHealth();
+
+		return sqrt(static_cast<double>((totalStackCount + 1) * creatureHealth) / (currentStackHealth + creatureHealth) - 1);
+	}
+
+	return 0.0;
+}
+
 double DamageCalculator::getDefenseSkillFactor() const
 {
 	int defenseAdvantage = getTargetDefenseEffective() - getActorAttackEffective();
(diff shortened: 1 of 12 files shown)


lib/battle/DamageCalculator.cpp
```diff
@@ -124,7 +127,22 @@ int DamageCalculator::getActorAttackBase() const
...
+int DamageCalculator::getActorAttackIgnored() const
+{
+	double multAttackReduction = battleBonusValue(info.defender, Selector::type()(BonusType::ENEMY_ATTACK_REDUCTION)) / 100.0;
+
+	if(multAttackReduction > 0)
+	{
+		int defaultRoundingMode = std::fegetround();
+		std::fesetround(FE_TOWARDZERO);
+		int reduction = std::nearbyint(multAttackReduction * getActorAttackBase());
+		std::fesetround(defaultRoundingMode);
```
IvanSavenko, line 142: why not simply floor it?

lib/battle/DamageCalculator.cpp
```diff
@@ -508,11 +526,23 @@ DamageEstimation DamageCalculator::calculateDmgRange() const
...
+	{
+		int totalStackCount = info.attacker->unitBaseAmount();
+		int currentStackHealth = info.attacker->getAvailableHealth();
+		int creatureHealth = info.attacker->getMaxHealth();
+
+		double revengeFactor = sqrt(static_cast<double>((totalStackCount + 1) * creatureHealth) / (currentStackHealth + creatureHealth) - 1);
+
+		revengeAdditionalMinDamage = std::round(damageBase.min * revengeFactor);
+		revengeAdditionalMaxDamage = std::round(damageBase.max * revengeFactor);
+	}
```
IvanSavenko, line 541: Move into a separate method and calculate as part of `getAttackFactors`?

lib/spells/effects/Damage.cpp
```diff
@@ -152,6 +152,22 @@ void Damage::describeEffect(std::vector<MetaString> & log, const Mechanics * m,
...
+		{
+			line.appendTextID("vcmi.battleWindow.accurateShot.resultDescription"); //(number) (unit type) was killed with an accurate shot!
+			line.replaceNumber(kills);
+			firstTarget->addNameReplacement(line, true);
+		}
+		else
+		{
+			line.appendTextID("vcmi.battleWindow.accurateShot.resultDescription.1"); //1 (unit type) were killed by accurate shots!
+			firstTarget->addNameReplacement(line, false);
+		}
```
IvanSavenko, line 168: - Change code to this
- Add translations for these strings:
```
vcmi.battleWindow.accurateShot.resultDescription.0 
vcmi.battleWindow.accurateShot.resultDescription.1
vcmi.battleWindow.accurateShot.resultDescription.2
```
And game will pick correct form even for languages like Polish
(won't work correctly if client and server have different languages at the moment, will look into it later)

```suggestion
		std::string preferredLanguage = VLC->generaltexth->getPreferredLanguage();
		std::string textID = "vcmi.battleWindow.accurateShot.resultDescription";
		line.appendTextID(Languages::getPluralFormTextID( preferredLanguage, kills, text));
		line.replaceNumber(kills);
		firstTarget->addNameReplacement(line, true);
```

server/battles/BattleActionProcessor.cpp
```diff
@@ -1175,6 +1184,62 @@ void BattleActionProcessor::attackCasting(const CBattleInfoCallback & battle, bo
 	}
 }
 
+std::set<SpellID> BattleActionProcessor::getSpellsForAttackCasting(TConstBonusListPtr spells, const CStack *defender)
```
IvanSavenko, line 1187: Still overcomplicated method IMO...

server/battles/BattleActionProcessor.cpp
```diff
@@ -1193,35 +1258,70 @@ void BattleActionProcessor::handleAfterAttackCasting(const CBattleInfoCallback &
...
 
-		double chanceToKill = attacker->valOfBonuses(BonusType::DEATH_STARE, BonusCustomSubtype::deathStareGorgon) / 100.0f;
-		vstd::amin(chanceToKill, 1); //cap at 100%
+		/* mechanics of Accurate Shot as in HotA:
+		 * each creature in an attacking stack has a X% chance of killing a creature in the attacked squad,
+		 * but the total number of killed creatures cannot be more than (number of creatures in an attacking squad) * X/100 (rounded up).
+		 * X = 3 multiplier for shooting without penalty and X = 2 if shooting with penalty. Ability doesn't work if shooting at creatures behind walls.
+		*/
+
+		auto bonus = attacker->getBonus(Selector::type()(BonusType::DEATH_STARE));
```
IvanSavenko, line 1273: please at the very least extract this... code to separate method

dydzio0614: You may also want to check 4th commit and tell me if you prefer split version, though it got bunch of code duplication

lib/battle/DamageCalculator.cpp
IvanSavenko: Only for this particular ability? Would be surprised to be honest.
For all rounding operations? In that case - this is wrong place for such change.

dydzio0614: i dont know if it works that way for this aprticular ability but i assume you checked this kind of stuff when you fixed behemoth damage etc. and I am not eager to check other abilities at the moment

server/battles/BattleActionProcessor.cpp
IvanSavenko: Move this version to a separate method

server/battles/BattleActionProcessor.cpp
```diff
@@ -1193,35 +1258,70 @@ void BattleActionProcessor::handleAfterAttackCasting(const CBattleInfoCallback &
...
+		else //DEATH_STARE
+			singleCreatureKillChancePercent = attacker->valOfBonuses(BonusType::DEATH_STARE, BonusCustomSubtype::deathStareGorgon);
 
+		double chanceToKill = singleCreatureKillChancePercent / 100.0;
+		vstd::amin(chanceToKill, 1); //cap at 100%
 		std::binomial_distribution<> distribution(attacker->getCount(), chanceToKill);
+		int killedCreatures = distribution(gameHandler->getRandomGenerator().getStdGenerator());
 
-		int staredCreatures = distribution(gameHandler->getRandomGenerator().getStdGenerator());
+		if(bonus->type == BonusType::DEATH_STARE)
```
IvanSavenko, line 1295: Are you sure that this branch is needed?

lib/battle/DamageCalculator.cpp
IvanSavenko: Nope. I did not.

Besides - these are float.
They *will* drift and return imprecise results and they *will* work differently on different platforms. So I'd rather keep code simple unless we know that this is correct implementation

IvanSavenko: Or alternatively - use int-only version, e.g.
`int reduction = (getActorAttackBase() * multAttackPercentage + 49) / 100`
So compiler / CPU can not play with floating point rounding / precision

server/battles/BattleActionProcessor.cpp
```diff
@@ -1175,55 +1184,149 @@ void BattleActionProcessor::attackCasting(const CBattleInfoCallback & battle, bo
...
+	else //DEATH_STARE
+		singleCreatureKillChancePercent = attacker->valOfBonuses(BonusType::DEATH_STARE, BonusCustomSubtype::deathStareGorgon);
 
-	if(attacker->hasBonusOfType(BonusType::DEATH_STARE))
+	double chanceToKill = singleCreatureKillChancePercent / 100.0;
+	vstd::amin(chanceToKill, 1); //cap at 100%
+	std::binomial_distribution<> distribution(attacker->getCount(), chanceToKill);
+	int killedCreatures = distribution(gameHandler->getRandomGenerator().getStdGenerator());
+
+	if(bonus->type == BonusType::DEATH_STARE)
```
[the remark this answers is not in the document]
IvanSavenko: I think this is the correct formula for both death stare and accurate shot:
`int maxToKill  = (attacker->getCount() * singleCreatureKillChancePercent + 99) / 100`

- 9 gorgon -> 1 kill
- 10 gorgon -> 1 kill
- 11 gorgon -> 2 kill
- 49 pirate with penalty -> 1 kill
- 50 pirate with penalty -> 1 kill
- 51 pirate with penalty -> 2 kill

rilian-la-te:
Why Accurate Shot is not a Death Stare subtype?

DjWarmonger:
Lol, that was me.

> Why Accurate Shot is not a Death Stare subtype?

Because it has a different animation. There are also some details related to distance penalty.

And yes, when Cove mods was first created, I implemented Accurate Shot as a Death Stare and it had just the same animation.

dydzio0614:
@rilian-la-te because we should not treat hota-based addons like punching bags

inaccurate HotA port is kinda one of top5 complaints of players about VCMI atm

rilian-la-te:
> Lol, that was me.
> 
> > Why Accurate Shot is not a Death Stare subtype?
> 
> Because it has a different animation. There are also some details related to distance penalty.
> 
> And yes, when Cove mods was first created, I implemented Accurate Shot as a Death Stare and it had just the same animation.

Why we attach animation to bonus type after all? It should be different spell to have different animantion, but same bonus. Maybe we can consolidate this bonus with some effort?

>inaccurate HotA port is one of top5 complaints of players about VCMI atm
Mechanically accurate HOTA is good. But we should not have bonus pollution, I wish. I think than we should consolidate bonuses as much as possible.

And I do not understand why hota is so fucking popular, but it is a different topic.

dydzio0614:
i believe that until we have scripting there is no other way than bonus pollution for bunch of custom monster features etc.

rilian-la-te:
Moat immunity can be spell immunity for moat ones. Spellcast on death/on kill is not a pollution. By pollution I means very similar bonuses than can be consolidated into one.

dydzio0614:
Already got different usage of subtype for these bonuses, and both bonuses have potential for expanding in different direction (configurable wall / ranged penalty for accurate shot etc) - out of scope of this PR

IvanSavenko:
@dydzio0614 
Having accurate shot as subtype of death stare makes sense. Because accurate shot is ranged version of death stare with *identical* mechanics / formula. This is the reason I asked to have accurate shot and gorgon stare share code. "Innacurate hota mod" argument is not applicable here because death stare and accurate shot are mostly the same ability.

As far as subtypes go, something like this should work:
- deathStareGorgon: melee-only version, as before
- deathStareCommander: as before, not sure if this should apply on ranged attacks - depending on how wog works I guess?
- deathStareNoRangedPenalty: for ranged shots with no penalty, pirates will get this bonus with value of 3
- deathStareDistancePenalty: for ranged shots with ranged penalty, pirates will get this bonus with value of 2
- optionally, deathStareObstaclePenalty: for ranged shots over walls / obstacles, pirates don't get this bonus but other mods might

rilian-la-te:
@IvanSavenko exactly what I want to say)

rilian-la-te:
@dydzio0614 addInfo)

dydzio0614:
addinfo does not support new typed format which would result in return of stuff serialized as int and kind of reintroduce stuff that Ivan solved with his past PR with converting all subtypes to string

rilian-la-te:
It is a bad reason for a bonus pollution, I think. Maybe introduce type-aware addInfo in another PR? For now, addInfo is used in a small fraction of bonuses.

IvanSavenko:
That's effectively a bug that needs to be fixed.
We already have same issue with SPECIAL_UPGRADE which stores creature ID in addInfo and possibly other bonuses

rilian-la-te:
@IvanSavenko necromancy stores creatureId in Addinfo too.

dydzio0614:
i would leave it as is for now, since i see bonus duplication problem as small and theoretical rather than practical at the moment + change would break mods compatibility (i would rather put current subtype in addinfo if anything), also the way we have it now it is intuitive for modders, also bonus can be easily found by people who pick ability name from hota wiki and try to find it in our bonuses names etc.
not mentioning combat log unintuitive behavior where death stare would give different log message for different addinfos

rilian-la-te:
@dydzio0614 @IvanSavenko I am strongly against bonus pollution and additional bonus just because "it is HOTA". I think we should do it as one bonus. And because it is unmerged - there is no mods who use its bonus name. So, I would rework it to one bonus.

IvanSavenko:
> change would break mods compatibility 

How? Mods don't use accurate shot at the moment and death stare does not uses addInfo right now

> also bonus can be easily found by people who pick ability name from h3 wiki and try to find it in our bonuses names etc.

Add comment in docs that bonus is also used for accurate shot. Description of some bonuses in docs deliberately mentions mapped h3 mechanics for easy search

> not mentioning combat log unintuitive behavior where death stare would give different log message for different addinfos

And some bonuses work completely different based on addInfo. So?

Current accurate shot is not exactly intuitive either.
- why it does not applies to shooting over walls? Ah, hota did it.
- why bonus strength is reduced to 66% when shooting with penalty? Ah, hota did it.

dydzio0614:
the things i totally hate:
1. not having such things told to me on very first code review
2. running 50 battles in hota after every change to ensure everything works correctly

so if you insist i can do it but i would rather look at other stuff

rilian-la-te:
@dydzio0614 @IvanSavenko also about ferocity - maybe it is possible to have it as a subtype of EXTRA_ATTACK bonus? Or it should be completely different one? I am for one bonus, if possible.

rilian-la-te:
@dydzio0614 sorry, but I have a bugs of GitHub at home (IDK why, but it often shows me "destination unreachable"). 

@IvanSavenko about bonus pollution - I think it should be mentioned inside some guidelines. And we should consolidate bonuses, if possible, I think (using my "old bonus" handlers).

rilian-la-te:
@dydzio0614 fixing bonus pollution require way more effort than prevent it before merging.

dydzio0614:
so i leave it up to @IvanSavenko to decide

IvanSavenko:
Merging as it, will merge accurate shot into death stare tomorrow

Laserlicht:
@dydzio0614 

What's the reason for?
> move text in creature amount label during battle 1 pixel down to better match H3

Doesn't look good after this change:
grafik
grafik

HD mod (like before this change):
grafik
grafik

dydzio0614:
@Laserlicht hmmmmmm for me it always looked 1 pixel off - check for number "7"

PR #3473, opened 2024-01-10
[1.5] Remove non-const global variables from library

### LINKED ISSUE #3323: Crashes on random map generation - multithreaded access to bonus system
**Describe the bug**
It looks that we have some crashes in our parallel RMG. Whenever RMG modifies game bonus tree, there is no protection against multi-threaded access, sometimes leading to crashes. Presumably - when two different threads pick same guardian creature or place same artifact.

For example, `CBonusSystemNode::attachTo` adds new element to `parents` and `children` vectors, however these vectors are not protected in any way. There are might be other (less obvious) cases.

Ideally, would be great to remove `children` vector entirely - to make constant part of our bonus system (e.g. CCreature / CArtifact classes) truly constant, but not sure how doable this approach is.

**Version**
 - OS: Android
 - Version: VCMI 1.4.1

**Call Stacks**
```
(LIB_SERVER::CBonusSystemNode::attachTo(LIB_SERVER::CBonusSystemNode&)+286)
(LIB_SERVER::CStackInstance::CStackInstance(LIB_SERVER::CreatureID const&, int, bool)+755)
(LIB_SERVER::ObjectManager::chooseGuard(int, bool)+659)
(LIB_SERVER::ObjectManager::addGuard(LIB_SERVER::rmg::Object&, int, bool)+667)
(LIB_SERVER::TreasurePlacer::createTreasures(LIB_SERVER::ObjectManager&)+857)
(LIB_SERVER::Modificator::run()+87)
(std::__ndk1::__function::__func<LIB_SERVER::CMapGenerator::fillZones()::$_1, std::__ndk1::allocator<LIB_SERVER::CMapGenerator::fillZones()::$_1>, void ()>::operator()()+392)
(boost::detail::task_shared_state<std::__ndk1::function<void ()>, void>::do_run()+1884)
(boost::detail::task_base_shared_state<void>::run()+2997)
(std::__ndk1::__function::__func<LIB_SERVER::ThreadPool::async(std::__ndk1::function<void ()>&&) const::$_0, std::__ndk1::allocator<LIB_SERVER::ThreadPool::async(std::__ndk1::function<void ()>&&) const::$_0>, void ()>::operator()()+3797)
(LIB_SERVER::ThreadPool::spawn()+1884)
```

```
(LIB_SERVER::CBonusSystemNode::attachTo(LIB_SERVER::CBonusSystemNode&)+286)
(LIB_SERVER::CArtifactInstance::CArtifactInstance(LIB_SERVER::CArtifact*)+127)
(LIB_SERVER::ArtifactUtils::createScroll(LIB_SERVER::SpellID const&)+190)
(std::__ndk1::__function::__func<LIB_SERVER::TreasurePlacer::addAllPossibleObjects()::$_3, std::__ndk1::allocator<LIB_SERVER::TreasurePlacer::addAllPossibleObjects()::$_3>, LIB_SERVER::CGObjectInstance* ()>::operator()()+205)
(LIB_SERVER::TreasurePlacer::constructTreasurePile(std::__ndk1::vector<LIB_SERVER::ObjectInfo*, std::__ndk1::allocator<LIB_SERVER::ObjectInfo*>> const&, bool)+1884)
(LIB_SERVER::TreasurePlacer::createTreasures(LIB_SERVER::ObjectManager&)+842)
(LIB_SERVER::Modificator::run()+87)
(std::__ndk1::__function::__func<LIB_SERVER::CMapGenerator::fillZones()::$_1, std::__ndk1::allocator<LIB_SERVER::CMapGenerator::fillZones()::$_1>, void ()>::operator()()+392)
(boost::detail::task_shared_state<std::__ndk1::function<void ()>, void>::do_run()+1884)
(boost::detail::task_base_shared_state<void>::run()+2997)
(std::__ndk1::__function::__func<LIB_SERVER::ThreadPool::async(std::__ndk1::function<void ()>&&) const::$_0, std::__ndk1::allocator<LIB_SERVER::ThreadPool::async(std::__ndk1::function<void ()>&&) const::$_0>, void ()>::operator()()+3797)
(LIB_SERVER::ThreadPool::spawn()+1884)
```

**Additional context**
Reported via Google Play crash statistics

IvanSavenko:
Note: targeting beta since branch is based on beta, will retarget to develop after 1.4.3

With this pull request it should be possible (in theory) for multiple users, such as client and server, to use same VLC instance while keeping their own copies of gamestate with all associated entities. End goal is to allow server to be part of client executable with zero / close to zero server startup time, eliminating a bit too common bugs with server startup, our massive mess of Android-specific and iOS-specific code, and removing annoying delay on starting new game / loading game / campaign transfer.

Approximate roadmap for future PR's (might have to wait for networking / lobby to finish to reduce conflicts):
- remove server/client library separation currently used for single-app build
- split server into "library" and minimalistic app that uses server library
- allow client to start server as a thread in own process using server library
- remove mass of platform-specific code in server startup code
- (not for 1.5) implement intra-process connection type that uses thread-safe container, e.g. boost::lockfree instead of TCP connection

Changes in this PR:
- Removed write access to constant part of bonus system - attaching child to CArtifact or CCreature node no longer modifies this object
- As a side effect - fixes #3323 
- Remove remaining instances of non-const pointers to VLC components or attempts to modify them
- IObjectInterface::cb is no longer static and is instead set on construction, using client or server instance
- Removed remaining non-const variables from VLC. For now - moved to CMap since this is where they are serialized.
- JsonRandom is now a class that contains IGameCallback instead of namespace
- handlers in VLC now use shared_ptr instead of raw ptr's

TODO:
- [x] recheck usage of getBonusLocalFirst method in bonus system
- [x] check scenario transfer in campaigns 
- [x] check RMG
- [x] check obelisks functionality
- [x] check town artifact merchants functionality
- [x] check Conflux university functionality
- [x] check for save compatibility with 1.4


AI/BattleAI/BattleEvaluator.cpp
```diff
@@ -350,10 +350,11 @@ bool BattleEvaluator::attemptCastingSpell(const CStack * activeStack)
 	LOGL("Casting spells sounds like fun. Let's see...");
 	//Get all spells we can cast
 	std::vector<const CSpell*> possibleSpells;
-	vstd::copy_if(VLC->spellh->objects, std::back_inserter(possibleSpells), [hero, this](const CSpell *s) -> bool
-	{
-		return s->canBeCast(cb->getBattle(battleID).get(), spells::Mode::HERO, hero);
-	});
+
+	for (auto const & s : VLC->spellh->objects)
```
kambala-decapitator, line 354: Don't see a reason to change this

lib/CHeroHandler.cpp
```diff
@@ -155,6 +155,13 @@ bool CHeroClass::isMagicHero() const
 	return affinity == MAGIC;
 }
 
+int CHeroClass::tavernProbability(FactionID faction) const
+{
+	if (selectionProbability.count(faction))
+		return selectionProbability.at(faction);
```
kambala-decapitator, line 161: IMO find() would be better

lib/CHeroHandler.cpp
```diff
@@ -394,7 +401,7 @@ void CHeroClassHandler::afterLoadFinalization()
 		}
 	}
 
-	for(CHeroClass * hc : objects)
+	for(auto const & hc : objects)
```
kambala-decapitator, line 404: const auto. Or is that a pointer?

IvanSavenko: Will change. Old habit :)
`const auto` and `auto const` are identical actually. (but not identical to `auto * const` in case of pointers)

AI/BattleAI/BattleEvaluator.cpp
IvanSavenko: Result of my initial attempt to change `VLC->spellh->objects` to shared_ptr.

Both version are actually not good -  `VLC->spellh->objects` should be a private member eventually, with some clear way to access objects / select matching.

rilian-la-te: @IvanSavenko can it be rewritten to use VLC->spells?

lib/IHandlerBase.h
```diff
@@ -44,34 +44,43 @@ class DLL_LINKAGE IHandlerBase
...
 };
 
 template <class _ObjectID, class _ObjectBase, class _Object, class _ServiceBase> class CHandlerBase : public _ServiceBase, public IHandlerBase
 {
+	const _Object * getObjectImpl(const int32_t index) const
+	{
+		if(index < 0 || index >= objects.size())
+		{
+			logMod->error("%s id %d is invalid", getTypeNames()[0], index);
+			throw std::runtime_error("Attempt to access invalid index " + std::to_string(index) + " of type " + getTypeNames().front());
```
kambala-decapitator, line 57: you're calling getTypeNames twice

PR #3479, opened 2024-01-13
Enable unit tests

nullkiller:


Laserlicht:
What happens if I run workflow without `HEROES_3_DATA_PASSWORD`. Is this still possible?
If not: May add a condition that the mission env var disables tests.

For example if I want to run Workflow in my Cloned repository.

Alexander-Wilms:
This should work:

https://stackoverflow.com/a/72926257/2278742

.github/workflows/github.yml
```diff
@@ -206,6 +206,16 @@ jobs:
         max-size: "5G"
         verbose: 2
 
+    - name: Clone Heroes 3 data
```
Alexander-Wilms, line 209:
```suggestion
    - name: Heroes 3 assets
```

PR #3480, opened 2024-01-13
Ban starting heroes in RMG

IvanSavenko:
From what I see, at the moment RMG uses list of allowed heroes from CMap to determine which heroes can be placed on map. 

However if player selects starting hero in game options this hero is not marked as unavailable for RMG and it is possible to get second copy of starting hero somewhere in the prison.

This change propagates list of pre-selected heroes to RMG, so starting hero is never a potential pick for prisons

@DjWarmonger can you check whether my reasoning / fix here is correct? If you want to use different approach here / already have fix in works then I am fine with closing this PR.


lib/rmg/CMapGenerator.cpp
```diff
@@ -491,14 +491,21 @@ const std::vector<HeroTypeID> CMapGenerator::getAllPossibleHeroes() const
 	for (HeroTypeID hero : map->getMap(this).allowedHeroes)
 	{
 		auto * h = dynamic_cast<const CHero*>(VLC->heroTypes()->getById(hero));
-		if ((h->onlyOnWaterMap && !isWaterMap) || (h->onlyOnMapWithoutWater && isWaterMap))
```
DjWarmonger, line 494: This condition needs to stay. There are some heroes that are allowed **only** on water maps or maps without water (Ranloo, Kinkeria in HoTA).

lib/rmg/CMapGenerator.cpp
```diff
@@ -491,14 +491,21 @@ const std::vector<HeroTypeID> CMapGenerator::getAllPossibleHeroes() const
...
-			ret.push_back(hero);
-		}
+
+		if(h->onlyOnMapWithoutWater && isWaterMap)
+			continue;
+
+		bool heroUsedAsStarting = false;
+		for (auto const & player : map->getMapGenOptions().getPlayersSettings())
+			if (player.second.getStartingHero() == hero)
+				heroUsedAsStarting = true;
```
DjWarmonger, line 503: We can break immediately, without checking more players.

lib/rmg/CMapGenerator.cpp
IvanSavenko: Not removed, but split into two. See 2 lines below.

PR #3486, opened 2024-01-13
Replace redundant types with `auto`

Alexander-Wilms:
Fixes some instances of https://sonarcloud.io/organizations/vcmi/rules?open=cpp%3AS5827&rule_key=cpp%3AS5827

First, find lines with redundant types:

```bash
grep -r --include \*.h --include \*.cpp " = " * | grep -v auto | grep "cast<" | grep -v for | grep -v if | grep -Po ".*cast<.*?>" | grep -v googletest | grep -Po ".+ .+ = .+" | tr -d "\t" | grep -v "fuzzylite" > redudant_types.txt
```

Then replace the first one with auto:

```python
import re

with open("redundant_types.txt") as f:
    for line in f:
        print()
        line = line.strip()
        path = line.split(":", 1)[0]
        original_code = line.split(":")[1]
        code = original_code.replace("*", " * ").replace("  ", " ").replace("* >", "*>")
        print(path)
        print(code)
        statement_without_redundant_type_matches = re.findall(r"[a-zA-Z0-9_]+ =.*", code)
        if statement_without_redundant_type_matches:
            statement_without_redundant_type = statement_without_redundant_type_matches[0]
            new_code = f"auto {statement_without_redundant_type}"
            print(new_code)

            with open(path, "r") as f:
                filedata = f.read()

            filedata = filedata.replace(code, new_code)

            with open(path, "w") as f:
                f.write(filedata)
```

### DIFF
--- a/AI/Nullkiller/Pathfinding/Actors.cpp
+++ b/AI/Nullkiller/Pathfinding/Actors.cpp
@@ -327,7 +327,7 @@ ExchangeResult HeroExchangeMap::tryExchangeNoLock(const ChainActor * other)
 			return result;
 		}
 
-		HeroActor * exchanged = new HeroActor(actor, other, newArmy, ai);
+		auto * exchanged = new HeroActor(actor, other, newArmy, ai);
 
 		exchanged->armyCost += newArmy->armyCost;
 		result.actor = exchanged;
@@ -342,7 +342,7 @@ HeroExchangeArmy * HeroExchangeMap::tryUpgrade(
 	const CGObjectInstance * upgrader,
 	TResources resources) const
 {
-	HeroExchangeArmy * target = new HeroExchangeArmy();
+	auto * target = new HeroExchangeArmy();
 	auto upgradeInfo = ai->armyManager->calculateCreaturesUpgrade(army, upgrader, resources);
 
 	if(upgradeInfo.upgradeValue)
@@ -393,7 +393,7 @@ HeroExchangeArmy * HeroExchangeMap::tryUpgrade(
 
 HeroExchangeArmy * HeroExchangeMap::pickBestCreatures(const CCreatureSet * army1, const CCreatureSet * army2) const
 {
-	HeroExchangeArmy * target = new HeroExchangeArmy();
+	auto * target = new HeroExchangeArmy();
 	auto bestArmy = ai->armyManager->getBestArmy(actor->hero, army1, army2);
 
 	for(auto & slotInfo : bestArmy)
@@ -445,7 +445,7 @@ std::string DwellingActor::toString() const
 
 CCreatureSet * DwellingActor::getDwellingCreatures(const CGDwelling * dwelling, bool waitForGrowth)
 {
-	CCreatureSet * dwellingCreatures = new CCreatureSet();
+	auto * dwellingCreatures = new CCreatureSet();
 
 	for(auto & creatureInfo : dwelling->creatures)
 	{
--- a/client/renderSDL/SDLImage.cpp
+++ b/client/renderSDL/SDLImage.cpp
@@ -184,7 +184,7 @@ std::shared_ptr<IImage> SDLImage::scaleFast(const Point & size) const
 	else
 		CSDL_Ext::setDefaultColorKey(scaled);//just in case
 
-	SDLImage * ret = new SDLImage(scaled, EImageBlitMode::ALPHA);
+	auto * ret = new SDLImage(scaled, EImageBlitMode::ALPHA);
 
 	ret->fullSize.x = (int) round((float)fullSize.x * scaleX);
 	ret->fullSize.y = (int) round((float)fullSize.y * scaleY);
--- a/launcher/lobby/lobby_moc.cpp
+++ b/launcher/lobby/lobby_moc.cpp
@@ -139,10 +139,10 @@ void Lobby::serverCommand(const ServerCommand & command) try
 
 			int playersJoined = args[tagPoint++].toInt();
 			int playersTotal = args[tagPoint++].toInt();
-			QTableWidgetItem * sessionPlayerItem = new QTableWidgetItem(QString("%1/%2").arg(playersJoined).arg(playersTotal));
+			auto * sessionPlayerItem = new QTableWidgetItem(QString("%1/%2").arg(playersJoined).arg(playersTotal));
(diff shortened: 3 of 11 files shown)


Alexander-Wilms:
An issue detected by the code analysis:

> Intentionality issue | Not logical
> Value stored to 'ret' is never read
> Unused assignments should be removed [cpp:S1854](https://sonarcloud.io/organizations/vcmi/rules?open=cpp%3AS1854&rule_key=cpp%3AS1854)
> Software qualities impacted:
> 
>     Maintainability

https://github.com/vcmi/vcmi/blob/4b8f8d90f67dde4dacf5ff903e3a4b383c780ec5/server/CVCMIServer.cpp#L1008-L1035

AI/BattleAI/AttackPossibility.cpp
```diff
@@ -117,8 +117,8 @@ float AttackPossibility::attackValue() const
 
 float hpFunction(uint64_t unitHealthStart, uint64_t unitHealthEnd, uint64_t maxHealth)
 {
-	float ratioStart = static_cast<float>(unitHealthStart) / maxHealth;
-	float ratioEnd = static_cast<float>(unitHealthEnd) / maxHealth;
+	auto ratioStart = static_cast<float>(unitHealthStart) / maxHealth;
+	auto ratioEnd = static_cast<float>(unitHealthEnd) / maxHealth;
```
IvanSavenko, line 121: To be honest in case of built-in types explicit type name might be better.

Not against this change though.

client/render/Graphics.cpp
```diff
@@ -113,7 +113,7 @@ void Graphics::initializeBattleGraphics()
 		if(!config["ac_mapping"].isNull())
 		for(const JsonNode &ac : config["ac_mapping"].Vector())
 		{
-			int ACid = static_cast<int>(ac["id"].Float());
+			auto ACid = static_cast<int>(ac["id"].Float());
```
IvanSavenko, line 116:
```suggestion
			int ACid = ac["id"].Integer();
```

client/renderSDL/CTrueTypeFont.cpp
```diff
@@ -29,7 +29,7 @@ std::pair<std::unique_ptr<ui8[]>, ui64> CTrueTypeFont::loadData(const JsonNode &
 
 TTF_Font * CTrueTypeFont::loadFont(const JsonNode &config)
 {
-	int pointSize = static_cast<int>(config["size"].Float());
+	auto pointSize = static_cast<int>(config["size"].Float());
```
IvanSavenko, line 32:
```suggestion
	int pointSize = config["size"].Integer();
```

server/CVCMIServer.cpp
```diff
@@ -1008,7 +1008,7 @@ HeroTypeID CVCMIServer::nextAllowedHero(PlayerColor player, HeroTypeID initial,
 void CVCMIServer::optionNextBonus(PlayerColor player, int dir)
 {
 	PlayerSettings & s = si->playerInfos[player];
-	PlayerStartingBonus & ret = s.bonus = static_cast<PlayerStartingBonus>(static_cast<int>(s.bonus) + dir);
+	auto ret = s.bonus = static_cast<PlayerStartingBonus>(static_cast<int>(s.bonus) + dir);
```
IvanSavenko, line 1011:
```suggestion
	s.bonus = static_cast<PlayerStartingBonus>(static_cast<int>(s.bonus) + dir);
	PlayerStartingBonus & ret = s.bonus;
```

AI/BattleAI/AttackPossibility.cpp
IvanSavenko: Probably OK as it. Unless somebody else prefers old version.
At the very least it should reduce amount of diagnostics at sonarcloud, potentially allowing to find actual issues.

nullkiller: At my company we vise versa prefer exact types instead of auto for better readability. In C++ there are cases when typenames are very big and auto is simpler to write. This is why I believe we exist in such indetermined state regarding this keyword and everyone makes it in a way they prefer. In this exact case I do not see a big difference and have no exact preference so.

lib/pathfinder/PathfindingRules.cpp
```diff
@@ -62,7 +62,7 @@ void MovementCostRule::process(
 
 		const int destinationLayerMaxMovePoints = pathfinderHelper->getMaxMovePoints(destination.node->layer);
 		const float costBeforeConversion = static_cast<float>(destMovePointsLeft) / sourceLayerMaxMovePoints;
-		const float costAfterConversion = static_cast<float>(movementPointsAfterEmbark) / destinationLayerMaxMovePoints;
+		auto costAfterConversion = static_cast<float>(movementPointsAfterEmbark) / destinationLayerMaxMovePoints;
```
nullkiller, line 65: If we change one, why not changing the others? Also Will variable be const after this change? Is it important for it to be const?

server/CVCMIServer.cpp
nullkiller: In general there was an idea that auto & should be always used instead of auto where we expect reference.
Replacing reference with auto may cause bugs because it will no longer be reference but copy instead.

lib/pathfinder/PathfindingRules.cpp
IvanSavenko: No idea why only one line got changed.

Variable is indeed no longer const, but should have no effect on code.

Alexander-Wilms: I'll check why the scripts ignored ``costBeforeConversion`.

Since it's recommended to make any variable a const if possible, I'll go through the changes and re-add it.

IvanSavenko: To be honest might be a good idea to exclude float's/int's/etc from this change after all.
for example this code will turn into
```cpp
const int destinationLayerMaxMovePoints = ...
const auto costBeforeConversion = ...
const auto costAfterConversion = ...
```
So in order to check type of `destinationLayerMaxMovePoints` you need to look left, but in order to check type of other two you need to look at cast. And what's worse is that in this case we have static_cast not on whole expression, but only on one operand.

Alexander-Wilms: I'll split it up in several PRs so it's easier to keep an overview.

I've just replaced the types in initializations using `new`.

PR #3491, opened 2024-01-14
Change default SDL driver name from opengl to empty (autoselection)

IvanSavenko:
We have too many cases where VCMI fails to start on Windows because SDL can't create opengl video driver.

This PR changes video driver on Windows to empty string which our code will interpret as using default / auto-selection.

### DIFF
--- a/config/schemas/settings.json
+++ b/config/schemas/settings.json
@@ -199,6 +199,7 @@
 				},
 				"driver" : {
 					"type" : "string",
+					"defaultWindows" : "",
 					"default" : "opengl",
 					"description" : "preferred graphics backend driver name for SDL2"
 				},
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
why not change to empty string for all platforms? Apple platforms should try Metal, for example (iirc iOS uses hardcoded order Metal -> OpenGL in VCMI)

IvanSavenko:
Don't want to experiment too much.
And we never had any reports of issues with default driver on any platforms other than Windows

dydzio0614:
what about new GUI driver picker - once you select something you cannot get back to "auto"

IvanSavenko:
> what about new GUI driver picker - once you select something you cannot get back to "auto"

It will have empty option as well, so you can select it as well

PR #3510, opened 2024-01-16
Remove trailing semicolons from more macros

Alexander-Wilms:
The analysis also lists this as an error, but I don't know how one could make the trailing semicolon after the macro usage mandatory.

```
RETURN_IF_NOT_BATTLE(battle::Units());
```
https://sonarcloud.io/code?id=vcmi_vcmi&selected=vcmi_vcmi%3Alib%2Fbattle%2FCBattleInfoEssentials.cpp&line=120

The macro is defined as such:

```
#define RETURN_IF_NOT_BATTLE(...) if(!duringBattle()) {logGlobal->error("%s called when no battle!", __FUNCTION__); return __VA_ARGS__; }
```
https://github.com/vcmi/vcmi/blob/develop/lib/battle/IBattleInfoCallback.h#L18

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -111,11 +111,7 @@
 // They all assume that interface mutex is locked.
 #define EVENT_HANDLER_CALLED_BY_CLIENT
 
-#define BATTLE_EVENT_POSSIBLE_RETURN	\
-	if (LOCPLINT != this)				\
-		return;							\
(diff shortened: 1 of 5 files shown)


IvanSavenko:
There is actually kind of "standard" way to wrap various macro:
`do { BODY } while (false)`
- this is a single statement, so code like `if (x) MACRO()` will always work correctly - since macro is simple text replacement, multi-statement macro might break such expressions
- this form requires semicolon at the very end

PR #3513, opened 2024-01-17
hide adventuremap

### LINKED ISSUE #2670: Add support for letterboxing
Currently when playing in a different aspect ratio or scaling level than the original game, vcmi opens stuff with set sizes (e.g. town screen, combat screen) in the center of the screen, but it leaves the adventure map UI on the screen as well

Screenshot_20230825_134716

But in the HD Mod when playing with a vertical resolution smaller than 664 or in the HD Edition,  when viewing such windows, the screen is letterboxed so that only that window is really shown

ss_89f4967a8e42185534d99ebad949b57894bb70fd

I think that the user should be able to toggle between these 2 ways of viewing stuff

Laserlicht:
Setting to hide adventuremap with texture when big 800x600 window is opened.

Fixes #2670

grafik


IvanSavenko:
Perhaps black fill might be better than texture fill? 
Similar to how we hide all UI during hotseat turn wait

dydzio0614:
I like texture fill personally

Laserlicht:
I also prefer texture

1. It's the HD mod way
2. Looks good on mobile (only left and right border)

And we could change this in the future (selectable modes).

PR #3523, opened 2024-01-19
Fix crash on RMG

IvanSavenko:
Apparently there was binding of a stack variable to a lambda by a reference, leading to crashes on RMG generation.

Slightly reworked code to avoid that, but might have reverted/broke some changes from RMG branch.

Will test a bit more and release as 1.4.4 later today.

@DjWarmonger If you have better solution - please let me know.


lib/rmg/modificators/TreasurePlacer.cpp
```diff
@@ -831,7 +837,7 @@ void TreasurePlacer::createTreasures(ObjectManager& manager)
 	{
 		for (auto* oi : treasurePile)
 		{
-			oi->destroyObject();
+			//oi->destroyObject();
```
DjWarmonger, line 840: This one is neccessary, actually likely to happen as RMG fails to find place for treasure pile.

IvanSavenko: Suggestions how to replace? Because at the moment what we have is crash. Which is way worse than too many banned heroes.

Perhaps store created object in `oi`? Can only one object be created using same `oi` instance?

IvanSavenko: Or perhaps that functor should instead be called as part of `Object::clear` / `Object::Instance::clear`?

IvanSavenko:
Added (hopefully) better solution - now that cleanup functor is called in `Object::Instance::clear`, in sync with actual deletion of object from memory

Alexander-Wilms:
The Flatpak Test build of 1.4.3 failed with this error:

```
 using PTY: False
{
    "errors": [
        "desktop-file-failed-validation"
    ],
    "warnings": [
        "appstream-summary-too-long",
        "appstream-screenshot-missing-caption"
    ],
    "desktopfile": [
        "error: value \"yes\" for boolean key \"SingleMainWindow\" in group \"Desktop Entry\" contains invalid characters, boolean values must be \"false\" or \"true\""
    ]
}
```

IvanSavenko:
@Alexander-Wilms will change to 'true'

IvanSavenko:
Seems to be working for me, at the very least - crash is fixed and don't see any obvious bugs in RMG.
Will merge this in ~3 hours if nothing else is found / no better solution for rmg crash

dydzio0614:
does final version solve treasure pile placement?

IvanSavenko:
@dydzio0614 
I think this is correct fix. At the very least - changes make sense to me and I don't have any crashes after dozens of rmg runs on various settings

DjWarmonger:
Looks good to me, certainly not worse than previous implementation. Works well, too.

PR #3530, opened 2024-01-20
Invite hero

Laserlicht:
Inviting hero in tavern (HD mod feature)

Also adds possibility to use double click on portrait to invite.

Extra setting to enable (sorry for german):
grafik

Tavern:
grafik

(simple) GUI:
grafik


dydzio0614:
not sure if this should be extra setting here or game mechanics flip switch in gameconfig.json - what do you think, @IvanSavenko ?

Laserlicht:
The reason for the Extra setting is, that it is basically some kind of game mechanics change. But here you can use it for example also in multi-player games just for fun, if you want.

IvanSavenko:
Thinking about it - how these options differ from what we have in game config json?
Both are for game mechanics changes, Both can be controlled by players, but in a different ways.

Perhaps we should think about integrating them together first? Or try to come up with rules as to which options go where?
I also have *heavy* concerns that these "extra options" will soon turn into wog options level of mess.

dydzio0614:
@IvanSavenko the way they differ is that they persist in saves, and such feature is desired for example for singleplayer tournaments

Laserlicht:
Some other difference is that it can set per game. If you decide in the chat: "Hey, let's play with 'Hero Invite'" the host can set this setting. It's a little bit like in other strategy games lobby.

IvanSavenko:
> @IvanSavenko the way they differ is that they persist in saves, 

Are you sure? Because right now turn timers definitely do not persist even though they use same mechanism as these options. And hard to say whether they *should* persist. Why should we block changes to simturns / timers if players want to change them?

> so players will have harder time to mess with disabling cheats etc. 

Cheating in SP or in SP tournaments is impossible to protect against so poor argument. If player want to cheat they will find a way. And as far as tournament go, they can easily add their own mod with correct game settings, just like PAH tournament did.

> Also these ingame options are easy to toggle which may be useful for stuff like "decide if simultaneous turns pathfinder should ignore monsters or not"

First of all:
I
hate
wog
options

And actually ignoring monsters is not really a mechanics change. Game still plays the same way. Change to how players experience game - sure. But not change in mechanics. Same to battle replay which is effectively easy way to reload game and possibly even cheats toggle.

But invite hero is clear change to mechanics.

IvanSavenko:
> Some other difference is that it can set per game. If you decide in the chat: "Hey, let's play with 'Hero Invite'" the host can set this setting. It's a little bit like in other strategy games lobby.

And why host can't decide in chat "hey, let's play with accumulating dwellings" or "hey, let's play with negative luck" or... well at least half of our game settings?

IvanSavenko:
To summarize my opinion:
- These "extra options" effectively duplicate our game settings logic. Yes, they do provide a way way to set option per-game, but this is still duplication. Why we allow unlimited battle replay per game but don't allow negative luck per map? Why we allow mods to change dwellings accumulation rules but don't allow to enable hero invite via mods?

- I agree that option to change in-game might be convenient, but it can quickly grow into massive mess called "wog options". IMO changes to mechanics should be part of our mod system. VCMI should replicate H3:SoD mechanics. So hero invite toggle should be done as a mod. Possibly same goes to battle replay option, but not sure about this one.

- Things like SP tournaments should be implemented as mod. So options that are required for SP tournament should be accessible to mod system.

Laserlicht:
> And why host can't decide in chat "hey, let's play with accumulating dwellings" or "hey, let's play with negative luck" or... well at least half of our game settings?

That would be cool :D
A drop down box with different gamerule sets. But this would conflict with the modding system... :)

This is a per game rule. Not a global setting for all games. And not changable after game start.

> What makes hero invite special compared to these options?

It's a gui change. And we already have a mess called "vcmi-extra" that works like this...

IvanSavenko:
> This is a per game rule. Not a global setting for all games. And not changable after game start.

Why? Why we can't change this option once the game starts? What makes this option so special compared to others?

> It's a gui change. 

No, it's not.
For example this is mine that gives you 1 of any rare resources (your choice!) for one-time cost of 2500 gold. It makes your selection of starting hero mostly irrelevant since this is the only hero you can get reliably in normal game.

> And we already have a mess called "vcmi-extra" that works with mods...

Not really. vcmi-extras mess consists of UI changes that we can't integrate due to usage of ripped assets.

Negative luck and dwelling accumulation work just fine as part of hota mod. Stack experience works just fine as part of wog mod. Why can't hero invite be a part of similar mod?

Laserlicht:
>     * These "extra options" effectively duplicate our game settings logic. Yes, they do provide a way way to set option per-game, but this is still duplication. Why we allow unlimited battle replay per game but don't allow negative luck per map? Why we allow mods to change dwellings accumulation rules but don't allow to enable hero invite via mods?

Yes, it's a duplication of settings. The main difference is that it is setted per game and that it is saved in the savefile.

>     * I agree that option to change in-game might be convenient, but it can quickly grow into massive mess called "wog options". IMO changes to mechanics should be part of our mod system. VCMI should replicate H3:SoD mechanics. So hero invite toggle should be done as a mod. Possibly same goes to battle replay option, but not sure about this one.

Understand that concern. But the options do little to change that. Currently, every WoG option requires the corresponding code in the game anyway.
Whether this is set by a flag in the mod or by an option does not change anything from the code side.

My point here is only to add options from the HD mod. As this is currently the most common variant. I'm not interested in WoG stuff at all.

dydzio0614:
I would personally move this one option into gameconfig

IvanSavenko:
> Yes, it's a duplication of settings. The main difference is that it is setted per game and that it is saved in the savefile.

Again - what makes this option so special compared to others? Why this option needs to be set per game and others - as part of mod?

> Currently, every (or maybe most) WoG option requires the corresponding code in the game anyway. 

So... Just like this option? Or all other options in this menu?

> Whether this is set by a flag in the mod or by an option does not change anything from the code side.

So let's make this into a mod instead of option?

> My point here is only to add options from the HD mod. As this is currently the most common variant. I'm not interested in WoG stuff at all.

Might be true for you. What should we do when somebody else comes and start adding wog options into the very same menu? IMO wog options is main source of why it is so messy and lack of wog options is why hota is so popular. No need for something similar in VCMI even if it will come with different name.

VCMI implements H3:SoD mechanics. Any extensions to it are domain of mods. Even if these mods have to rely on engine functionality, whether this functionality is generic "mod can add new faction" or more specific "mod can alllow hero invites"

> I would personally move this one option into gameconfig

Seconded...

Laserlicht:
> Seconded

So basically remove it from this menu and make it behind a config flag in gameSettings? And a mod overrides this flag?

IvanSavenko:
Yep. Tiny mod, probably separate from extras, just one more mod into our main repo.

Laserlicht:
The mod to enable:
```
{
	"name" : "Invite hero",
	"description" : "Possibility to invite hero in the tavern",
	"version" : "1.0",
	"author" : "Laserlicht",
	"modType" : "Expansion",
	"compatibility":
	{
		"min" : "1.5.0"
	},
	"changelog":
	{
		"1.0": ["Initial version"]
	},
	"settings" : {
		"heroes" : {
			"invite" : true
		}
	},
	"german" : {
		"name" : "Helden einladen",
		"description" : "Möglichkeit in der Taverne Helden einzuladen"
	}
}
```

Currently for testing min version have to set to "1.4".

@misiokles Can you create a new repo after merged and when we have 1.5 mod list.

IvanSavenko:
Now - on actual topic.
Any reason to have *invite* hero feature instead of *buy* hero feature?
Since the only reason to invite hero is to buy him after inviting then why do it in a roundabout way? Why not simply allow buying any hero directly?

Laserlicht:
- Some kind of balancing (that you have to buy a hero beforehand). Sure - you could also adjust the prices.
- People know the feature like this from the hd mod.
- Feels less like cheating - fits more in tavern mechanics.

server/processors/HeroPoolProcessor.cpp
```diff
@@ -113,6 +114,10 @@ void HeroPoolProcessor::selectNewHeroForSlot(const PlayerColor & color, TavernHe
 
 	CGHeroInstance *newHero = pickHeroFor(needNativeHero, color);
 
+	const auto & heroesPool = gameHandler->gameState()->heroesPool;
+	if(VLC->settings()->getBoolean(EGameSettings::HEROES_INVITE) && heroesPool->unusedHeroesFromPool().count(nextHero) && heroesPool->isHeroAvailableFor(nextHero, color))
+		newHero = heroesPool->unusedHeroesFromPool()[nextHero];
+
```
IvanSavenko, line 120: Checks should probably be done in `HeroPoolProcessor::hireHero`

config/gameConfig.json
```diff
@@ -289,7 +289,9 @@
 			// Chances for a hero with default army to receive corresponding stack out of his predefined starting troops
 			"startingStackChances": [ 100, 88, 25],
 			// number of artifacts that can fit in a backpack. -1 is unlimited.
-			"backpackSize"		: -1
+			"backpackSize"		: -1,
+			// if heroes are invitable in tavern
+			"invite"            : false
```
IvanSavenko, line 294: More clear name.
Same goes to setting enum value
```suggestion
			"tavernInvite"            : false
```

client/windows/GUIClasses.cpp
```diff
@@ -615,6 +655,32 @@ void CTavernWindow::HeroPortrait::hover(bool on)
 		GH.statusbar()->clear();
 }
 
+CTavernWindow::HeroSelector::HeroSelector(std::map<HeroTypeID, CGHeroInstance*> InviteableHeroes, std::function<void(CGHeroInstance*)> OnChoose)
```
IvanSavenko, line 658: Can't reuse starting hero selector window?

client/windows/GUIClasses.cpp
```diff
@@ -507,14 +511,42 @@ CTavernWindow::CTavernWindow(const CGObjectInstance * TavernObj, const std::func
...
 		CCS->videoh->open(VideoPath::builtin("TAVERN.BIK"));
+
+	addInvite();
+}
+
+void CTavernWindow::addInvite()
+{
+	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
+
+	if(VLC->settings()->getBoolean(EGameSettings::HEROES_INVITE))
```
IvanSavenko, line 522: Perhaps use early return to avoid function that consist from single if block?
```suggestion
	if(!VLC->settings()->getBoolean(EGameSettings::HEROES_INVITE))
		return;
```

client/lobby/OptionsTabBase.cpp
```diff
@@ -407,8 +407,14 @@ void OptionsTabBase::recreate()
 	}
 
 	if(auto buttonCheatAllowed = widget<CToggleButton>("buttonCheatAllowed"))
+	{
 		buttonCheatAllowed->setSelectedSilent(SEL->getStartInfo()->extraOptionsInfo.cheatsAllowed);
+		buttonCheatAllowed->setInputEnabled(SEL->screenType != ESelectionScreen::loadGame);
```
IvanSavenko, line 412: Why not block()?

IvanSavenko, line 412: On a similar note - why block at all? Do we need to block this functionality in saves for some reason?

client/windows/GUIClasses.cpp
Laserlicht: Possible. But not that easy. Would need some refactoring, as hero selector window from main menu uses some other internal classes from mainmenu. Don't want to do this now.

client/lobby/OptionsTabBase.cpp
Laserlicht: Because after starting the game the setting is currently not changeable. You have to decide at the map beginning.

I don't want to disable the tab because you may want to look how it's configured. But changing the values after game start doesn't make anything.

May change this behavior in the future.

client/windows/GUIClasses.cpp
```diff
@@ -615,6 +655,32 @@ void CTavernWindow::HeroPortrait::hover(bool on)
...
 
+CTavernWindow::HeroSelector::HeroSelector(std::map<HeroTypeID, CGHeroInstance*> InviteableHeroes, std::function<void(CGHeroInstance*)> OnChoose)
+	: CWindowObject(BORDERED), inviteableHeroes(InviteableHeroes), onChoose(OnChoose)
+{
+	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
+
+	pos = Rect(0, 0, 16 * 48, (inviteableHeroes.size() / 16 + (inviteableHeroes.size() % 16 != 0)) * 32);
+	background = std::make_shared<CFilledTexture>(ImagePath::builtin("DIBOXBCK"), Rect(0, 0, pos.w, pos.h));
+
+	int x = 0, y = 0;
```
IvanSavenko, line 666:
```suggestion
	int x = 0;
	int y = 0;
```

PR #3533, opened 2024-01-20
fix battle text position

Laserlicht:
Not sure if best solution, but works for German/english and polish.

@dydzio0614 can you test?


dydzio0614:
works for me

Laserlicht:
> works for me

can you also test again chinese?

dydzio0614:
looks good as well

PR #3547, opened 2024-01-22
Add second workaround for CPack issue on macOS

Alexander-Wilms:
The `hdiutil: create failed - Resource busy` issue needs a different workaround than the `No child processes` issue.

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -268,6 +268,13 @@ jobs:
       run: |
         ctest --preset ${{matrix.preset}}
 
+    - name: Kill XProtect to work around CPack issue on macOS 
+      if: runner.os == 'macOS'
+      run: |
+        # Cf. https://github.com/actions/runner-images/issues/7522#issuecomment-1556766641
+        echo Killing...; sudo pkill -9 XProtect >/dev/null || true;
+        echo Waiting...; while pgrep XProtect; do sleep 3; done;
+
     - name: Pack
       id: cpack
       if: ${{ matrix.pack == 1 }}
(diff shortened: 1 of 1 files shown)


.github/workflows/github.yml
```diff
@@ -268,6 +268,13 @@ jobs:
       run: |
         ctest --preset ${{matrix.preset}}
 
+    - name: Kill XProtect to work around CPack issue on macOS 
+      if: runner.os == 'macOS'
```
kambala-decapitator, line 272: condition should be "when building for macOS platform", as cpack for iOS doesn't use hdiutil

.github/workflows/github.yml
```diff
@@ -268,6 +268,13 @@ jobs:
...
+      run: |
+        # Cf. https://github.com/actions/runner-images/issues/7522#issuecomment-1556766641
+        echo Killing...; sudo pkill -9 XProtect >/dev/null || true;
+        echo Waiting...; while pgrep XProtect; do sleep 3; done;
```
kambala-decapitator, line 276: if pkill fails, you'll end up in an endless loop

kambala-decapitator (changes requested):

Alexander-Wilms:
It still doesn't work reliably:

```
CPack: Create package
CPack Error: Error executing: /usr/bin/sudo /usr/bin/hdiutil detach "/Volumes/VCMI"
CPack Error: Error detaching temporary disk image.
CPack Error: Problem compressing the directory
CPack Error: Error when generating package: VCMI (PR 3547)
Error: Process completed with exit code 1.
```

kambala-decapitator (approved):

PR #3554, opened 2024-01-24
Fix invalid logic which made AI not attack guards

DjWarmonger:
Default value is set to false, but logic is reversed and thus adventure AI ignores guards and everything behind them.


dydzio0614:
did you test if AI plays correctly again with this change?

DjWarmonger:
> did you test if AI plays correctly again with this change?

Sue I did

dydzio0614 (approved):

IvanSavenko:
Will double-check and add to 1.4.5. Thanks for the fix.

PR #3561, opened 2024-01-25
Try to improve VCAI building build order

MnHebi:
Testing reveals extra structures still get done before intended build order(seen Market being built when Dwellings were expected).

### DIFF
--- a/AI/VCAI/BuildingManager.cpp
+++ b/AI/VCAI/BuildingManager.cpp
@@ -139,7 +139,6 @@ void BuildingManager::setAI(VCAI * AI)
 //Set of buildings for different goals. Does not include any prerequisites.
 static const std::vector<BuildingID> essential = { BuildingID::TAVERN, BuildingID::TOWN_HALL };
 static const std::vector<BuildingID> basicGoldSource = { BuildingID::TOWN_HALL, BuildingID::CITY_HALL };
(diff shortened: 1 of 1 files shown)


AI/VCAI/BuildingManager.cpp
```diff
@@ -172,33 +173,42 @@ bool BuildingManager::getBuildingOptions(const CGTownInstance * t)
...
 
+	if (t->hasBuilt(BuildingID::CASTLE))
+	{
+		if (tryBuildAnyStructure(t, unitGrowth, 4))
+			return true;
+	}
 
+	//make required structures for City Hall
+	if(tryBuildNextStructure(t, cityhallRequirements))
+		return true;
```
IvanSavenko, line 196: Is this necessary? tryBuildNextStructure builds requirements first, so code like `tryBuildThisStructure(t, BuildingID::CITY_HALL)` should build Marketplace/Blacksmith/Mages Guild anyway

MnHebi: Just being thorough in specifying what needs to be built since older code comments indicated that sometimes requirements were not being built when expected.

IvanSavenko: Such comments might be outdated.

And if such a bug exists then it should be investigated.
Sure, you've added fix for City Hall. But what if such bug triggers with Tavern and Town Hall?
Or with any dwelling which might have different requirements, including towns from mods?

PR #3569, opened 2024-01-27
instant finish of battle with autocombat

Laserlicht:


Laserlicht:
@IvanSavenko There is missing a check that this option only can only used against ai. Do you have an idea how to do this? Following code isn't usable in `client/battle/BattleWindow.cpp` because the access to `getPlayerState` is prohibited for the non hosting user on the client side.

```
	auto defender = owner.curInt->cb->getPlayerState(owner.getBattle()->getBattle()->getSidePlayer(BattleSide::DEFENDER));
	auto attacker = owner.curInt->cb->getPlayerState(owner.getBattle()->getBattle()->getSidePlayer(BattleSide::ATTACKER));
	bool isDefenderHuman = defender && defender->isHuman();
	bool isAttackerHuman = attacker && attacker->isHuman();
	onlyOnePlayerHuman = isDefenderHuman != isAttackerHuman;
```

Laserlicht:
Thanks!

IvanSavenko:
> option to replace normal auto-combat with instant finish (if battle against ai, otherwise normal behavior)

Perhaps if this option is on, disable autocombat button in battles vs human? So it will always have one function:
- autocombat if new option is off
- end with quick combat if new option is on

client/battle/BattleWindow.cpp
```diff
@@ -733,6 +749,32 @@ std::optional<uint32_t> BattleWindow::getQueueHoveredUnitId()
 	return queue->getHoveredUnitIdIfAny();
 }
 
+void BattleWindow::endWithAutocombat()
```
IvanSavenko, line 752: Perhaps add confirmation prompt? I think HD mod does have it and generally makes sense since it might have rather serious effects.

client/battle/BattleWindow.cpp
```diff
@@ -52,6 +54,12 @@ BattleWindow::BattleWindow(BattleInterface & owner):
 	pos.h = 600;
 	pos = center();
 
+	PlayerColor defenderColor = owner.getBattle()->getBattle()->getSidePlayer(BattleSide::DEFENDER);
+	PlayerColor attackerColor = owner.getBattle()->getBattle()->getSidePlayer(BattleSide::ATTACKER);
+	bool isDefenderHuman = defenderColor != PlayerColor::NEUTRAL && LOCPLINT->cb->getStartInfo()->playerInfos.at(defenderColor).isControlledByHuman();
+	bool isAttackerHuman = attackerColor != PlayerColor::NEUTRAL && LOCPLINT->cb->getStartInfo()->playerInfos.at(attackerColor).isControlledByHuman();
```
IvanSavenko, line 60: But not sure if this is correct name for that method...
```suggestion
	bool isDefenderHuman = defenderColor.isValidPlayer() && LOCPLINT->cb->getStartInfo()->playerInfos.at(defenderColor).isControlledByHuman();
	bool isAttackerHuman = attackerColor.isValidPlayer() && LOCPLINT->cb->getStartInfo()->playerInfos.at(attackerColor).isControlledByHuman();
```

client/battle/BattleWindow.cpp
```diff
@@ -733,6 +749,32 @@ std::optional<uint32_t> BattleWindow::getQueueHoveredUnitId()
 	return queue->getHoveredUnitIdIfAny();
 }
 
+void BattleWindow::endWithAutocombat() 
+{
+	if(!owner.makingTurn() || owner.tacticsMode)
+		return;
```
IvanSavenko, line 755: Why not allow in tactics? With instant end of tactics phase?

client/battle/BattleWindow.cpp
Laserlicht: Good idea! Done.

client/battle/BattleWindow.cpp
Laserlicht: Didn't want to change that. Normal autobattle also excludes tactic mode.

Issue #2852, opened 2023-09-16
Visual bugs with duplicating artifacts on Altar Of Sacrifice


wb180:
**To Reproduce**
See the gifs:
Case 1:
![1](https://github.com/vcmi/vcmi/assets/109801241/284f1407-01db-4c8b-9c31-7785f762d02f)

Case 2:
![2](https://github.com/vcmi/vcmi/assets/109801241/0042028e-a5fa-486f-9673-551a5f9b8b5f)

**Expected behavior**
Duplication shouldn't happen

**Actual behavior**
A lot of duplication and after you sacrificed actual artifacts, click on duplicated artifacts also gives an error:
![Screenshot_20230916_212247](https://github.com/vcmi/vcmi/assets/109801241/6e759ba4-50e7-4f44-8bdf-bc03825e19af)

**Version**
 - Version: Latest develop

SoundSSGood:
Will fix soon

PR #3574, opened 2024-01-27
Artifacts altar related fixes

SoundSSGood:
Artifacts altar is now artifact holder. Artifacts actually move to the altar. This allows to reuse existing code.

- https://github.com/vcmi/vcmi/issues/2885 - fixed
- https://github.com/vcmi/vcmi/issues/2852 - fixed
- Now assembly/disassembly of artifacts is available on the altar screen.
- War machines can no longer be moved to the altar.
- On the server side there is now a check - cannot sacrifice a non-trade artifact.
- On the server side there is now a check - the evil hero cannot sacrifice artifact.


SoundSSGood:
https://sonarcloud.io/project/issues?pullRequest=3574&open=AY1Mn0nc1uF8CraNJ63O&id=vcmi_vcmi
There are flaws in the code regarding const/non-const instances. Some instances cannot be get as non-const.  I think this needs to be reworked later.

client/widgets/CWindowWithArtifacts.cpp
```diff
@@ -153,26 +138,23 @@ void CWindowWithArtifacts::clickPressedArtPlaceHero(CArtifactsOfHeroBase & artsI
...
 					}
 					if(isTransferAllowed)
-						artSetPtr->swapArtifacts(srcLoc, dstLoc);
+						LOCPLINT->cb->swapArtifacts(srcLoc, dstLoc);
 				}
-				else
+				else if(auto art = artPlace.getArt())
 				{
-					if(artPlace.getArt())
+					if(artSetPtr->getHero()->tempOwner == LOCPLINT->playerID)
```
IvanSavenko, line 145:
```suggestion
					if(artSetPtr->getHero()->getOwner() == LOCPLINT->playerID)
```

lib/CGameInfoCallback.cpp
```diff
@@ -970,17 +971,27 @@ const CGObjectInstance * CGameInfoCallback::getObjInstance( ObjectInstanceID oid
 
 CArtifactSet * CGameInfoCallback::getArtSet(const ArtifactLocation & loc) const
```
IvanSavenko, line 972: I do see that this is old code, but such methods should not return non-const entities.

server/CGameHandler.cpp
```diff
@@ -3432,6 +3434,17 @@ bool CGameHandler::isAllowedExchange(ObjectInstanceID id1, ObjectInstanceID id2)
...
+		{
+			return true;
+		}
+		if(o2->ID == Obj::ALTAR_OF_SACRIFICE)
+		{
+			const auto visitingHero = getVisitingHero(o2);
+			const auto thisHero = static_cast<const CGHeroInstance*>(o1);
+			if(visitingHero == thisHero)
+				return true;
+		}
```
IvanSavenko, line 3447: Instead of checking for object ID you should check that:
1) Object can be casted to IMarket
2) If yes - that object allows trade using EMarketMode::ARTIFACT_EXP mode

lib/CGameInfoCallback.cpp
SoundSSGood: > I do see that this is old code, but such methods should not return non-const entities.

How do you propose to get non-constant instances in this case?

IvanSavenko: CGameInfoCallback should only provide const API.

This method probably should be moved to CGameState and made non-const
And CGameState does provides non-const API for object access: `getObjInstance` or `getHero`

SoundSSGood: Moved

server/CGameHandler.cpp
SoundSSGood: In ApplyGhNetPackVisitor::visitBulkExchangeArtifacts should also be type cast instead of ID comparsion?

SoundSSGood: done

PR #3575, opened 2024-01-27
fix infobox position while disabling

### LINKED ISSUE #2480: Disabling Battle queue during battle should reposition hero portrait display.
**Describe the bug**
During battle with battle queue enabled - hero portrait is shown at left/right side correctly.
image

But when we disable Battle queue by pressing Q button, hero portrait doesn't shift and window looks a little odd.
image

**To Reproduce**
1. Provoke any battle.
2. Press Q during battle
3. Observe glitch.

**Expected behavior**
Portrait should fit to battle window with and without battle queue.

**Version**
 - OS: [Windows]
 - Version: [VCMI 1.3]

Laserlicht:
fixes #2480

### DIFF
--- a/client/battle/BattleWindow.cpp
+++ b/client/battle/BattleWindow.cpp
@@ -130,19 +130,13 @@ void BattleWindow::createStickyHeroInfoWindows()
 	{
 		InfoAboutHero info;
 		info.initFromHero(owner.defendingHeroInstance, InfoAboutHero::EInfoLevel::INBATTLE);
-		Point position = (GH.screenDimensions().x >= 1000)
-				? Point(pos.x + pos.w + 15, pos.y + 60)
-				: Point(pos.x + pos.w -79, pos.y + 195);
-		defenderHeroWindow = std::make_shared<HeroInfoBasicPanel>(info, &position);
+		defenderHeroWindow = std::make_shared<HeroInfoBasicPanel>(info, nullptr);
 	}
 	if(owner.attackingHeroInstance)
 	{
 		InfoAboutHero info;
 		info.initFromHero(owner.attackingHeroInstance, InfoAboutHero::EInfoLevel::INBATTLE);
-		Point position = (GH.screenDimensions().x >= 1000)
-				? Point(pos.x - 93, pos.y + 60)
-				: Point(pos.x + 1, pos.y + 195);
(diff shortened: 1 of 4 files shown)


client/battle/BattleWindow.cpp
```diff
@@ -130,19 +130,13 @@ void BattleWindow::createStickyHeroInfoWindows()
...
 	}
 	if(owner.attackingHeroInstance)
 	{
 		InfoAboutHero info;
 		info.initFromHero(owner.attackingHeroInstance, InfoAboutHero::EInfoLevel::INBATTLE);
-		Point position = (GH.screenDimensions().x >= 1000)
-				? Point(pos.x - 93, pos.y + 60)
-				: Point(pos.x + 1, pos.y + 195);
-		attackerHeroWindow = std::make_shared<HeroInfoBasicPanel>(info, &position);
+		attackerHeroWindow = std::make_shared<HeroInfoBasicPanel>(info, nullptr);
```
IvanSavenko, line 139: I think that 2nd parameter from `HeroInfoBasicPanel` constructor can be removed? Or still used somewhere else?

client/battle/BattleWindow.cpp
```diff
@@ -295,25 +325,21 @@ void BattleWindow::updateStackInfoWindow(const CStack * stack)
...
 	else
 		defenderStackWindow = nullptr;
 	
 	if(stack && stack->unitSide() == BattleSide::ATTACKER)
 	{
-		Point position = (GH.screenDimensions().x >= 1000)
-				? Point(pos.x - 93, attackerHeroWindow ? attackerHeroWindow->pos.y + 210 : pos.y)
-				: Point(pos.x + 1, attackerHeroWindow ? attackerHeroWindow->pos.y : pos.y + 135);
-		attackerStackWindow = std::make_shared<StackInfoBasicPanel>(stack, &position);
+		attackerStackWindow = std::make_shared<StackInfoBasicPanel>(stack, nullptr);
```
IvanSavenko, line 336: Same 2nd parameter here, in `StackInfoBasicPanel`?

client/battle/BattleWindow.cpp
Laserlicht: This one is used for the right click on hero (if the permanent view disabled).

PR #3576, opened 2024-01-28
CVideoHandler: Call SDL_RenderClear() when playing intro, call SDL_Re…

Alexander-Wilms:
…nderFillRect() for spellbook animation

Gets rid of the [glitches during the intro](https://github.com/vcmi/vcmi/pull/3169) and doesn't create a black frame around the spellbook.

I should probably unify `overlayVideo`, `useOverlay` and `overlay` though.

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -430,15 +430,15 @@ void playIntro()
 {
 	auto audioData = CCS->videoh->getAudio(VideoPath::builtin("3DOLOGO.SMK"));
 	int sound = CCS->soundh->playSound(audioData);
-	if(CCS->videoh->openAndPlayVideo(VideoPath::builtin("3DOLOGO.SMK"), 0, 1, true, true, false))
+	if(CCS->videoh->openAndPlayVideo(VideoPath::builtin("3DOLOGO.SMK"), 0, 1, EVideoType::INTRO))
 	{
 		audioData = CCS->videoh->getAudio(VideoPath::builtin("NWCLOGO.SMK"));
 		sound = CCS->soundh->playSound(audioData);
-		if (CCS->videoh->openAndPlayVideo(VideoPath::builtin("NWCLOGO.SMK"), 0, 1, true, true, false))
+		if (CCS->videoh->openAndPlayVideo(VideoPath::builtin("NWCLOGO.SMK"), 0, 1, EVideoType::INTRO))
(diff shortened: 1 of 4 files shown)


client/CMT.cpp
```diff
@@ -430,15 +430,15 @@ void playIntro()
 {
 	auto audioData = CCS->videoh->getAudio(VideoPath::builtin("3DOLOGO.SMK"));
 	int sound = CCS->soundh->playSound(audioData);
-	if(CCS->videoh->openAndPlayVideo(VideoPath::builtin("3DOLOGO.SMK"), 0, 1, true, true))
+	if(CCS->videoh->openAndPlayVideo(VideoPath::builtin("3DOLOGO.SMK"), 0, 1, true, true, false))
```
IvanSavenko, line 433: that `true, true. false` part is anything but readable. Can probably be replaced with single bool (intro mode or page flip mode, don't think that there are other places that use this mode). Or with more clear enum with 2-3 values

IvanSavenko (approved):

client/CMT.cpp
Alexander-Wilms: The campaign screen calls open() and update(), should that use openAndPlayVideo() as well?

IvanSavenko: No. open() + update() is preferable to openAndPlay.

openAndPlay() effectively blocks all game activity other than video playback. Might be useful for intro, but not suitable when there are other elements (e.g mission into subtitles)

Meanwhile open() + update() work in line with our UI system, its only problem is global state - only one video can be played at once.

client/CVideoHandler.cpp
```diff
@@ -672,8 +679,24 @@ bool CVideoPlayer::playVideo(int x, int y, bool stopOnKey)
...
+			overlay = false;
+			break;
+		case EVideoType::SPELLBOOK:
+		default:
+			stopOnKey = false;
+			scale = false;
+			overlay = true;
+	}
 	open(name, false, true, scale);
 	bool ret = playVideo(x, y,  stopOnKey);
```
Alexander-Wilms, line 701: Should overlay be a `CVideoPlayer::playVideo()` argument?

IvanSavenko: Either way would work I think?
Function parameter might be more logical

PR #3583, opened 2024-01-31
town portal dialog adjustments

Laserlicht:
- right click show town info
- click on list element centers map on object
- show icons for towns

### DIFF
--- a/client/windows/GUIClasses.cpp
+++ b/client/windows/GUIClasses.cpp
@@ -39,6 +39,7 @@
 #include "../render/Canvas.h"
 #include "../render/CAnimation.h"
 #include "../render/IRenderHandler.h"
+#include "../render/IImage.h"
 
 #include "../../CCallback.h"
 
@@ -1674,11 +1675,13 @@ CThievesGuildWindow::CThievesGuildWindow(const CGObjectInstance * _owner):
 }
 
 CObjectListWindow::CItem::CItem(CObjectListWindow * _parent, size_t _id, std::string _text)
-	: CIntObject(LCLICK | DOUBLECLICK),
+	: CIntObject(LCLICK | DOUBLECLICK | RCLICK_POPUP),
 	parent(_parent),
 	index(_id)
 {
 	OBJECT_CONSTRUCTION_CAPTURING(255-DISPOSE);
+	if(parent->images.size() > index)
+		icon = std::make_shared<CPicture>(parent->images[index], Point(1, 1));
 	border = std::make_shared<CPicture>(ImagePath::builtin("TPGATES"));
 	pos = border->pos;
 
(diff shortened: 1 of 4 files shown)


IvanSavenko:
Perhaps it should also use same town order as on adventure map?

client/windows/GUIClasses.cpp
```diff
@@ -1724,10 +1740,14 @@ CObjectListWindow::CObjectListWindow(const std::vector<int> & _items, std::share
...
 }
 
-CObjectListWindow::CObjectListWindow(const std::vector<std::string> & _items, std::shared_ptr<CIntObject> titleWidget_, std::string _title, std::string _descr, std::function<void(int)> Callback, size_t initialSelection)
+CObjectListWindow::CObjectListWindow(const std::vector<std::string> & _items, std::shared_ptr<CIntObject> titleWidget_, std::string _title, std::string _descr, std::function<void(int)> Callback, size_t initialSelection, std::vector<std::shared_ptr<IImage>> images)
 	: CWindowObject(PLAYER_COLORED, ImagePath::builtin("TPGATE")),
 	onSelect(Callback),
-	selected(initialSelection)
+	onExit(nullptr),
+	onPopup(nullptr),
+	onClicked(nullptr),
```
IvanSavenko, line 1748: Redundant
```suggestion
```

client/windows/GUIClasses.cpp
```diff
@@ -1701,17 +1704,30 @@ void CObjectListWindow::CItem::select(bool on)
...
+		parent->onPopup(index);
+}
+
+CObjectListWindow::CObjectListWindow(const std::vector<int> & _items, std::shared_ptr<CIntObject> titleWidget_, std::string _title, std::string _descr, std::function<void(int)> Callback, size_t initialSelection, std::vector<std::shared_ptr<IImage>> images)
 	: CWindowObject(PLAYER_COLORED, ImagePath::builtin("TPGATE")),
 	onSelect(Callback),
-	selected(initialSelection)
+	onExit(nullptr),
+	onPopup(nullptr),
+	onClicked(nullptr),
```
IvanSavenko, line 1728: Same here

Laserlicht:
> Perhaps it should also use same town order as on adventure map?

Added ordering functionality and use localstate. Works - maybe we should find a way to push localstate order back to server in future (would also allow to save order).

IvanSavenko:
Yes, this is known bug, which is a bit pain to fix due to how our serializer works. Also affects hero paths and sleep status.

We should probably save local state in json and push it to server as new netpack whenever this local state changes and let server save it as part of save game. This should avoid all issue we had with this before.

If you are interested in it - feel free to try to implement it.

Laserlicht:
> If you are interested in it - feel free to try to implement it.

Not a high prio for me. But maybe later. :)

PR #3591, opened 2024-02-02
Add more patterns to .gitignore

Alexander-Wilms:
These were needed for files created by
`mkdir build && cd build && cmake .. && cmake --build .`


kambala-decapitator:
I'd like to point out that creating build dir inside src dir normally is an incorrect approach

I'd rather just add a single line: `/build`. I see there's `build/` already, but better anchor it to the root and remove the trailing slash.

kambala-decapitator:
based on the lines you added it seems that your build directory was actually the same as src

and cmake invocation without explicit -S and -B options is basically obsolete / legacy, not recommended to use

Alexander-Wilms:
Ah, I missed that. Maybe the reason for the out-of-source build could be mentioned in https://github.com/vcmi/vcmi/blob/develop/docs/developers/Building_Linux.md

kambala-decapitator:
> Ah, I missed that. Maybe the reason for the out-of-source build could be mentioned in https://github.com/vcmi/vcmi/blob/develop/docs/developers/Building_Linux.md and the commands updated to use `-S` and `-B`.

yes, good idea. You can check e.g. in macOS or Android doc.

Alexander-Wilms:
I've now only added the files generated during an out-of-source build.

Laserlicht:
Can you also add:
.vscode/

kambala-decapitator (approved):

Issue #3550, opened 2024-01-23
Neutral army skips the turn while fighting with them by using Battle AI


by003:
**Describe the bug**
While fighting with neutral army, their creatures often do nothing and skip the turn, if using battle AI for them.

**To Reproduce**
1. Set neutral AI as Battle AI.
2. Battle with a neutral army without flying and range creatures
3. See error

**Expected behavior**
They will wait and close to player's army in the end of this turn.

**Actual behavior**
 If AI creatures cannot enter the attack range within one turn, they will skip the turn automatically.

**Did it work earlier?**
It worked well in 1.42.

**Version**
 - OS: Windows
 - Version: 1.44

ninjatao:
In both 1.4.4 & 1.4.5, this also happens in battle with enemy heroes. Compared 1.4.2 and 1.4.5 with the same battle, from scenario 2 of campaign "Play with Fire".

In 1.4.2, all enemy troops move forward in the first turn. 
![3031706155212_ pic](https://github.com/vcmi/vcmi/assets/6846049/45a9e8cf-8ba1-44a3-aa22-a7ee158b498b)

In 1.4.5, many enemy troops didn't do anything in the first several turns. Snapshot in turn 5, when zombies and skeletons didn't move from the beginning of the battle.
![3021706155211_ pic](https://github.com/vcmi/vcmi/assets/6846049/663866e2-d88f-4c28-8612-6bb37839a216)

by003:
Yes it is a fatal problem which need to be fixed in priority I think. If you use auto battle, without ranged or flying army, combat will never ends.

PR #3594, opened 2024-02-03
#3550 - fix case when unit is in reachability map but only next turn

nullkiller:
Fix logic of battle ai when some units can reach enemy in 2 turns only


nullkiller:
What branch should it target?

IvanSavenko:
develop
Will cherry-pick if there is need for a hotfix.

PR #3595, opened 2024-02-03
NKAI Object graph

nullkiller:
The goal of this work is to optimize performance on big maps. For this I want to switch from calculating paths through tile graph to calculating paths through object graph. Combining both things will allow us having robust paths to nearby objects and approximate paths towards far object. The work is still in progress so it is already able to pass a few turns without crash)

Sample of visual logger
image

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -373,7 +373,11 @@ void AIGateway::objectRemoved(const CGObjectInstance * obj, const PlayerColor & 
 		return;
 
 	nullkiller->memory->removeFromMemory(obj);
-	nullkiller->baseGraph->removeObject(obj);
+
+	if(nullkiller->baseGraph && nullkiller->settings->isObjectGraphAllowed())
+	{
+		nullkiller->baseGraph->removeObject(obj);
+	}
 
 	if(obj->ID == Obj::HERO && obj->tempOwner == playerID)
 	{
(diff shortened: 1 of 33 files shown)


DjWarmonger:
Could this potentially let AI operate with fog of war, that is only within the visible area?

nullkiller:
No, vise versa this thing needs open map, because base graph is shared across all AIs. The idea is that the thing is very static and thus does not need to be recalculated each turn.

nullkiller:
So the thing is optional and experimental. By default turned off. Can be merged now more or less safely

DjWarmonger:
So, is there anything to test here?

nullkiller:
I checked a bit, looks alive. In order to enable object graph you need to change or mod nkai-settings.json
`
	"allowObjectGraph" : true,
	"mainHeroTurnDistanceLimit" : 3,
	"scoutHeroTurnDistanceLimit" : 1`

So I do not feel it is ready to be enabled by default and still plan further improvements.
For now fighting with sonar cloud...

DjWarmonger:
I can tell this branch runs flawlessly, without any additional configuration. AI advances quickly and smoothly, no loopholes or other issues detected. Maybe you fixed something along the way?

nullkiller:
object graph is an experimental thing which will substitute normal pathfinding for far objects. But until it is enabled you should not see any difference. You can try to compare execution time on big map with and without configuration. For now I have not gathered much performance metrics. And this thing is very raw so I expect it will get better later.

nullkiller:
Also I reduced size of pathfinder storage significantly and unconditionally) This may improve performance.

nullkiller:
[AI object graph.zip](https://github.com/vcmi/vcmi/files/14554567/AI.object.graph.zip)
Here is a mod to enable object graph btw.

Issue #3604, opened 2024-02-08
switch to json5


Laserlicht:
**Describe your proposal**
Just for discussion: Should we switch our config to json5?

The problem is that we use additional "features" in our json that not supported by the JSON standard. This causes problem in every case, where real json, standard confirm files are expected. For example editors and IDEs which therefore rightly mark the json document as incorrect and show errors.

Basically our used features are supported by json5 and our json files can be validated as json5. Basically they are already json5 but with the incorrect file ending. :)

There are already some big players using json5. E.g. Chromium: https://github.com/chromium/chromium/blob/feb3c9f670515edf9a88f185301cbd7794ee3e52/third_party/blink/renderer/platform/runtime_enabled_features.json5

What do you think?

**Additional context**
We should still able to load `.json`-Files for backward compatibility, but prefer `.json5`-File. And rename jsons in main repo as well as for new / reworked mods.

IvanSavenko:
Might be a good idea, but not a quick one. I think some IDE's (e.g. VS code) have 'jsonc" format (Json with comments) which matches our json parser.

And from what I see, json5 is *way* more than json with comments. Here is their example:
```json5
{
  // comments
  unquoted: 'and you can quote me on that',
  singleQuotes: 'I can use "double quotes" here',
  lineBreaks: "Look, Mom! \
No \\n's!",
  hexadecimal: 0xdecaf,
  leadingDecimalPoint: .8675309, andTrailing: 8675309.,
  positiveSign: +1,
  trailingComma: 'in objects', andIn: ['arrays',],
  "backwardsCompatible": "with JSON",
}
```
Frankly this is not extension of json format but a completely new format.

So while our json files can be viewed as valid json5 format, json5 validator can NOT be used to validate our files. Because if people/modders will try to use any json5 feature other than comments json5 will okay this file but our json parser would fail.

Laserlicht:
> Frankly this is not extension of json format but a completely new format.

Yes, just created a PR to test such cases. :)

> So while our json files can be viewed as valid json5 format, json5 validator can NOT be used to validate our files. Because if people/modders will try to use any json5 feature other than comments json5 will okay this file but our json parser would fail.

Yes, of course our JSON parser supports not every feature. So either a non-standard json parser, or an incomplete json5 parser. :)

By the way: Apple seems to have native JSON5 support: https://developer.apple.com/documentation/foundation/jsondecoder/3766916-allowsjson5

IvanSavenko:
[...]

PR #3605, opened 2024-02-08
Simplify validation of JSON with comments & CI ccache fix for feature branches in forks

Alexander-Wilms:
This simplifies the script and makes the output more readable

Using https://github.com/linjackson78/jstyleson, the [JSON5 example](https://github.com/vcmi/vcmi/issues/3604#issuecomment-1934968493) fails validation:

`❌ json5_example.json: Expecting property name enclosed in double quotes: line 2 column 9 (char 10)`


Alexander-Wilms:
Should be ready to merge

PR #3607, opened 2024-02-10
Remove redundant `virtual` specifiers

Alexander-Wilms:
grep -nr virtual | grep -v googletest | grep override > ../redundant_virtual.txt

```python
import os

with open("../redundant_virtual.txt") as f:
    for line in f:
        print()
        line: str = line.strip()
        print(line)
        tmp = line.split(":")
        file = tmp[0].strip()
        code = tmp[-1].strip()
        print(file)
        print(code)
        new_code = code.replace("virtual ", "", 1)
        # https://superuser.com/a/802490/578501
        command = f"export FIND='{code}' && export REPLACE='{new_code}' && ruby -p -i -e \"gsub(ENV['FIND'], ENV['REPLACE'])\" {file}"
        os.system(command)
```


DjWarmonger:
You could simply do a search & replace with regex placeholder ;)

Alexander-Wilms:
But then I'd have to use RegEx...

IvanSavenko (approved):

PR #3609, opened 2024-02-11
Allow running server as part of client process

### LINKED ISSUE #2560: Improve Start Game speed
**Describe your proposal**
Game start is an order of magnitudes slower than HotA. Screen just lags at connecting to local host even for single player. (which happens to also be confusing by itself)
image

### LINKED ISSUE #3271: Client is unable to connect to the server again
**Describe the bug**
When 1000+ ms passed in another connection attempt and connection is aborted by the user then another connection to the server is not possible.

**Game logs**
Please attach game logs: `VCMI_client.txt`, `VCMI_server.txt` etc.

**To Reproduce**
1. Click Single Player
2. Wait 1000+ ms when Connecting window appear
3. Abort/Cancel connection
4. Try co connect to server again via Single Player or Load the game buttons

**Expected behavior**
Always connect to the server.

**Actual behavior**
Connection to the server is stuck.

**Did it work earlier?**
VCMI 1.3.2 always connected to the server even if aborted (Abort/Cancel didn't worked)

**Screenshots**
-

**Version**
 - OS: W7/W10
 - Version: [VCMI version]

**Additional context**
[vcmi_logs.zip](https://github.com/vcmi/vcmi/files/13623741/vcmi_logs.zip)

IvanSavenko:
With this PR, vcmi will run server as a single thread inside client process. Server and client will share VLC, eliminating double-load and long server startup time. Some delay still exists - UI loading, map list parsing, but it is way lower than before and can be investigated / fixed separately. This mode is used by default on all platforms, including mobile, where it replaces previously existing workarounds.

This also fixes occasional bugs that we had where server will not shutdown correctly for some reason.

Old mode, with server running as separate process is still available on desktop systems, as well as separate vcmiserver executable. However now it uses boost::process API instead of single std::system() call, allowing (for example) forced termination of vcmiserver.

List of changes:
- 'server' / 'vcmiserver' project is now a static library that is used by both client and serverapp. This behavior is now used unconditionally. Old ENABLE_SINGLE_APP_BUILD flag still exists, but only used for client+launcher
- Moved server startup code into a separate project, 'serverapp' (which generates vcmiserver / VCMI_Server.exe app as before). Server app now can only be built on desktop systems, and can be enabled using ENABLE_SERVER flag (on by default)
- Unintended side effect is that server in thread mode no longer creates own log file. Now VCMI_Server_log.txt is only used by server process. Server thread will now write log to same file as client. And in order to distinguish client & server log entries, records in log file will now also contain human-readable thread ID.
- Extracted server management from CServerHandler into separate class/file, with separate implementations of thread and process modes.
- removed a lot of mobile-specific workarounds related to server pseudoprocess startup on these platforms
- reorganized options section of CMakeLists.txt, to clearly specify platform-specific options
- renamed ENABLE_STATIC_AI_LIBS to ENABLE_STATIC_LIBS because as it turns out, this flag also turns libvcmi into a static library. Also fixed compile error when using this mode with gcc
- removed separate server version of library for single-app mode

Fixed bugs:
- Fixes #3271 
- Fixes #2560


IvanSavenko:
Seems to be (mostly) working on Windows, Linux and Android.
Discovered crash on closing client window, will fix.

There is still some delay on opening map list - on my system ~100ms was spent on loading maps and ~200ms for UI loading. Server startup is close to instant now - below 10 ms.

DjWarmonger:
No suprise if every map is stored in an individual file. Streaming could be implemented here to hide the delay.
Same is true for savegames I guess?

IvanSavenko:
Most likely same goes for saved games. However at the moment I have around 200 maps, and less than 10 saved games. So for me saved games open quite fast.

Streaming or multi-threading will help here. But not critical at the moment. While 300 ms is still noticeable amount of time it is still way better than several seconds we had before. Or even more on Androids or with multiple mods.

kambala-decapitator:
only now found time to check the changes - great job! but...

- removing android abilities to run server in separate process and to use dynamic AI libs was a mistake
- merging without testing on iOS also wasn't a great idea

kambala-decapitator:
> Server thread will now write log to same file as client. And in order to distinguish client & server log entries, records in log file will now also contain human-readable thread ID.

maybe I missed it in the changes, but does the log contain a line like "server thread - %u" to quickly figure out which thread server is running on?

IvanSavenko:
> * removing android abilities to run server in separate process 

Can restore if need. But any need for that mode? The way it works right now is that we have annoying several seconds delay when server starts and countless crashes on server shutdown, to a point where I can't find anything other than these crashes on Google Play Console.

> * and to use dynamic AI libs was a mistake

Was that changed? If I did - then this was not intended and trivial to revert.

> * merging without testing on iOS also wasn't a great idea

Any issues on iOS?

> maybe I missed it in the changes, but does the log contain a line like "server thread - %u" to quickly figure out which thread server is running on?

Not exactly. Now every log record will contain human-readable thread name. For example:
`[2024-Feb-14 19:39:12.941873] INFO [MainGUI] global - Initialization of VCMI (together): 2316 ms`
"MainGui" is name of thread that produced this message. And since server now runs in a single thread, all messages logged by server will be logged from context of "runServer" thread.
(except for RMG - which uses thread pool, haven't checked how those threads are named)

kambala-decapitator:
> Can restore if needed. But any need for that mode? The way it works right now is that we have annoying several seconds delay when server starts and countless crashes on server shutdown, to a point where I can't find anything other than these crashes on Google Play Console.

well, it's still a working option. If someone wants to experiment with it, there should be a way.

> Was that changed? If I did - then this was not intended and trivial to revert.
> Apk still contains single vcmiclient.so and CMake has ENABLE_STATIC_LIBS set to ON unconditionally on Android

IIRC Android can be compiled with dynamic AI libs and I tested that a year ago or so. There were runtime issues though, that's why default is to use static. See #1616 for an explanation. But again, if someone wants to experiment with it, there should be a way.

> Any issues on iOS?

Haven't tested yet. But the point is, merging things that have serious impact on a platform ("single process" is the only possible option on iOS) without testing is wrong.

> Now every log record will contain human-readable thread name

nice, thanks!

IvanSavenko:
> well, it's still a working option. If someone wants to experiment with it, there should be a way.

Hmm.. I guess it could be reimplemented as another server mode, in addition to thread and process. Haven't considered that. 

Due to significant changes to server startup code that would require rewrite anyway and have not seen any advantages to keeping that method, considering amount of issues it had.

> IIRC Android can be compiled with dynamic AI libs and I tested that a year ago or so. There were runtime issues though, that's why default is to use static. See #1616 for an explanation. But again, if someone wants to experiment with it, there should be a way.

There is way - change option in CMake from OFF to ON. Would require editing cmake since that option is forced to OFF on Android, but that's all that's needed.

> Haven't tested yet. But the point is, merging things that have serious impact on a platform ("single process" is the only possible option on iOS) without testing is wrong.

True. Problem we don't really have people who test PR's here at the moment. And I had changes that would conflict with changes from here.

I did also tested Android and Windows, And changes on iOS were actually minimal. "Single process" is what iOS already had before - both client and server running as single process, albeit with different library instances (e.g. VLC).

kambala-decapitator:
> Hmm.. I guess it could be reimplemented as another server mode, in addition to thread and process. Haven't considered that.
> Due to significant changes to server startup code that would require rewrite anyway and have not seen any advantages to keeping that method, considering amount of issues it had.

iirc on native side it was basically the same as the server app with the exception that it was loaded as a library. And on JVM side it used that "background service" code that was removed in this PR. Anyway, it's not urgent to bring back and can be tracked as an issue.

> There is way - change option in CMake from OFF to ON. Would require editing cmake since that option is forced to OFF on Android, but that's all that's needed.

yes, that was also my impression based on the PR changes :)

> Problem we don't really have people who test PR's here at the moment

you could just give a call on discord / slack asking for testing or saying "will merge the PR in 24h", I read those channels rather frequently

Laserlicht:
I think you have missed to extend the firewall rule: https://github.com/vcmi/vcmi/blob/8d86216332da1bb6d6d3f24fd0bf247c03c325ca/CMakeLists.txt#L801

At least I got some firewall questions again.

PR #3614, opened 2024-02-13
Sonarcloud fixes 2

IvanSavenko:
Few more fixes for potentially dangerous issues from Sonarcloud

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -689,13 +689,14 @@ namespace vstd
 	bool isAlmostZero(const Floating & value)
 	{
 		constexpr Floating epsilon(0.00001);
-		return std::abs(value) < epsilon;
+		return std::abs(value) <= epsilon;
 	}
 
 	template<typename Floating1, typename Floating2>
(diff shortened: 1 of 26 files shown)


Global.h
```diff
@@ -685,6 +685,19 @@ namespace vstd
 		return a + (b - a) * f;
 	}
 
+	template<typename Floating>
+	bool isAlmostZero(const Floating & value)
+	{
+		constexpr Floating epsilon(0.00001);
```
Alexander-Wilms, line 691: Wouldn't https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon be better?

IvanSavenko: No, that one is too small.
Working with float is *fun*. For example, simple expression like `a*b/c` can be compiled as `a/c*b`, potentially leading to rounding error. And more operations you have - then more these rounding errors can accumulate. In case of 1 operation, value from numeric_limits might be suitable, but when you have potentially large number of operations you want to have much larger epsilon

PR #3615, opened 2024-02-14
Fix SDL2MAIN_LIBRARY CMake warning

Alexander-Wilms:
```
 -- Found SDL2: /usr/lib/x86_64-linux-gnu/libSDL2.so (found version "2.0.20")
CMake Warning (dev) at /usr/local/share/cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:438 (message):
  The package name passed to `find_package_handle_standard_args` (SDL2main)
  does not match the name of the calling package (SDL2).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
  cmake_modules/FindSDL2.cmake:318 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:473 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

This change gets rid of the warning, but I don't know if other references to `SDL2MAIN_LIBRARY` can be removed as well.

### DIFF
--- a/cmake_modules/FindSDL2.cmake
+++ b/cmake_modules/FindSDL2.cmake
@@ -314,13 +314,6 @@ FIND_PACKAGE_HANDLE_STANDARD_ARGS(SDL2
                                   REQUIRED_VARS SDL2_LIBRARY SDL2_INCLUDE_DIR
                                   VERSION_VAR SDL2_VERSION_STRING)
 
-if(SDL2MAIN_LIBRARY)
-  FIND_PACKAGE_HANDLE_STANDARD_ARGS(SDL2main
-                                    REQUIRED_VARS SDL2MAIN_LIBRARY SDL2_INCLUDE_DIR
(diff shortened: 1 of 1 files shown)


rilian-la-te:
SDL2MAIN is needed on Windows, AFAIK. Need to test your patches on Windows.

Alexander-Wilms:
Is the msvc CI build not sufficient?

rilian-la-te:
No, because it will build fine, but can cause runtime failures. So, you need to test runtime failures on Windows or at least in Wine.

Alexander-Wilms:
Both the mingw and the msvc build work fine using Wine.

Alexander-Wilms:
Both builds work fine on Windows 11

kambala-decapitator:
IIUC for other platforms SDL2MAIN is provided by Conan: https://github.com/vcmi/vcmi/blob/develop/conanfile.py#L110

PR #3640, opened 2024-02-25
Fix thread race

IvanSavenko:
Fixes thread race on attempt to send packet from two threads (can happen in AI actions)


IvanSavenko:
Merging. No issues after playing with 6 NKAI + 1 VCAI for 1 month.

PR #3644, opened 2024-02-28
fix castle quick combat

### LINKED ISSUE #3608: Quick combat issue
Using a quick combat while defending the town, if lose in quick battle and reject the result, even win the battle by manual, the town will still become AI’s. 

In 1.50 development build. I think @Laserlicht need to know this because he made the quick combat function.

Laserlicht:
Should fix #3608

@by003 Please test it

### DIFF
--- a/server/queries/BattleQueries.cpp
+++ b/server/queries/BattleQueries.cpp
@@ -19,16 +19,9 @@
 #include "../../lib/battle/SideInBattle.h"
 #include "../../lib/CPlayerState.h"
 #include "../../lib/mapObjects/CGObjectInstance.h"
+#include "../../lib/mapObjects/CGTownInstance.h"
 #include "../../lib/networkPacks/PacksForServer.h"
 #include "../../lib/serializer/Cast.h"
-
-void CBattleQuery::notifyObjectAboutRemoval(const CObjectVisitQuery & objectVisit) const
-{
-	assert(result);
-
-	if(result)
-		objectVisit.visitedObject->battleFinished(objectVisit.visitingHero, *result);
-}
 
 CBattleQuery::CBattleQuery(CGameHandler * owner, const IBattleInfo * bi):
 	CQuery(owner),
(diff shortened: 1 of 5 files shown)


by003:
Is it fixed in develop build?

Laserlicht:
No, it's not merged yet.

You have to download it here: https://github.com/vcmi/vcmi/actions/runs/8087791352

server/queries/BattleQueries.cpp
```diff
@@ -104,6 +98,12 @@ void CBattleDialogQuery::onRemoval(PlayerColor color)
 	}
 	else
 	{
+		auto hero = bi->getSideHero(BattleSide::ATTACKER);
+		auto visitingObj = bi->getDefendedTown() ? bi->getDefendedTown() : gh->getVisitingObject(hero);
```
IvanSavenko, line 102: Why do we need actually "visiting object" instead of defender army? `bi->getSideArmy` I think?

server/queries/BattleQueries.cpp
```diff
@@ -104,6 +98,12 @@ void CBattleDialogQuery::onRemoval(PlayerColor color)
...
+		
 		gh->battles->endBattleConfirm(bi->getBattleID());
+		
+		if(visitingObj)
+			visitingObj->battleFinished(hero, *result);
```
IvanSavenko, line 107: How will the game call `battleFinished` in case of AI-initiated battles? 

From what I see, `CBattleDialogQuery` is only created for AI-vs-human battles. And not created in AI-vs-AI or human-vs-human battles

Laserlicht: Yes, you're right. Changed it... new way only for AI vs human.

by003:
Can you merge it into develop?

by003:
There is a confusion in the version I have downloaded. If you reckon the bug is fixed, juest merge it into develop and I can make a thorough testing.

Laserlicht:
> Can you merge it into develop?

From my side okay.
@IvanSavenko also ok from your side?

PR #3645, opened 2024-02-29
Save rmg options

DjWarmonger:
- [x] Save & load RMG settings in lobby
- [x] Save & load RMG template in Editor
- [x] Save & load game difficulty in lobby
- [x] Add extra map sizes in editor dialog

Additionally, removed some magic numbers and added extra map sizes to Editor drop-down menu.

### DIFF
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -196,10 +196,7 @@ void CServerHandler::startLocalServerAndConnect(bool connectToLobby)
 	auto si = std::make_shared<StartInfo>();
 
 	auto lastDifficulty = settings["general"]["lastDifficulty"];
-	if (lastDifficulty.isNumber())
-	{
-		si->difficulty = lastDifficulty.Integer();
-	}
+	si->difficulty = lastDifficulty.Integer();
 
 	logNetwork->trace("\tStarting local server");
 	serverRunner->start(getLocalPort(), connectToLobby, si);
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -140,6 +140,12 @@ CServerHandler::CServerHandler()
 {
 	uuid = boost::uuids::to_string(boost::uuids::random_generator()());
 	registerTypesLobbyPacks(*applier);
+
+	auto lastDifficulty = settings["general"]["lastDifficulty"];
+	if (lastDifficulty.isNumber())
+	{
+		si->difficulty = lastDifficulty.Integer();
+	}
 }
 
 void CServerHandler::threadRunNetwork()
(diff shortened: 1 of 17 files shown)


client/ServerRunner.h
```diff
@@ -9,12 +9,15 @@
  */
 #pragma once
 
+#include "../lib/StartInfo.h"
+
 class CVCMIServer;
+struct StartInfo;
```
IvanSavenko, line 15: I think you need namespace wrapper here, not include

```suggestion
VCMI_LIB_NAMESPACE_BEGIN
struct StartInfo;
VCMI_LIB_NAMESPACE_END
```

DjWarmonger: KK, got it already. Unusual it was missing froma  file.

client/CServerHandler.cpp
```diff
@@ -193,8 +193,16 @@ void CServerHandler::startLocalServerAndConnect(bool connectToLobby)
 		serverRunner.reset(new ServerThreadRunner());
 #endif
 
+	auto si = std::make_shared<StartInfo>();
+
+	auto lastDifficulty = settings["general"]["lastDifficulty"];
+	if (lastDifficulty.isNumber())
+	{
+		si->difficulty = lastDifficulty.Integer();
+	}
```
IvanSavenko, line 202: Redundant. Settings are validated, so it is not possible to get anything else in here.
```suggestion
	si->difficulty = lastDifficulty.Integer();
```

client/lobby/RandomMapTab.cpp
```diff
@@ -38,6 +38,11 @@
 #include "../../lib/filesystem/Filesystem.h"
 #include "../../lib/RoadHandler.h"
 
+//#include "../../lib/GameSettings.h"
```
IvanSavenko, line 41:
```suggestion
```

mapeditor/jsonutils.cpp
```diff
@@ -69,6 +69,7 @@ QVariant toVariant(const JsonNode & node)
 		return QVariant(node.Bool());
 		break;
 	case JsonNode::JsonType::DATA_FLOAT:
+	case JsonNode::JsonType::DATA_INTEGER:
```
IvanSavenko, line 72: Should probably be like this:
```cpp
	case JsonNode::JsonType::DATA_FLOAT:
		return QVariant(node.Float());
	case JsonNode::JsonType::DATA_INTEGER:
		return QVariant(node.Integer());
```

IvanSavenko:
Also - can you take a look at stuff found by sonarcloud?
https://sonarcloud.io/project/issues?id=vcmi_vcmi&pullRequest=3645&resolved=false&inNewCodePeriod=true

Not asking to fix everything, just those that makes sense to fix from your point of view

client/CServerHandler.cpp
DjWarmonger: But we could get null value if nothing was saved before, am I right?

IvanSavenko: No. If this was simple JsonNode (settings have a bit different logic) then this code would've returned default value for integer (0). But not null.

However since this is a setting, it will *always* have a value. Either one that was set by player before or (if not set) - default value as specified in schema.

PR #3647, opened 2024-02-29
Fix SonarCloud issue: non-portable path to file '<Windows.h>'

Alexander-Wilms:
non-portable path to file '<Windows.h>'; specified path differs in case from file name on disk "#include" paths should be portable cpp:S3806

Cf. https://sonarcloud.io/project/issues?resolved=false&rules=cpp%3AS3806&types=CODE_SMELL&id=vcmi_vcmi&open=AYzk1AQFdKxbEY19qsF2

### DIFF
--- a/client/eventsSDL/NotificationHandler.cpp
+++ b/client/eventsSDL/NotificationHandler.cpp
@@ -18,7 +18,7 @@
 
 #define WIN32_LEAN_AND_MEAN		// Exclude rarely-used stuff from Windows headers
 // Windows Header Files:
-#include <windows.h>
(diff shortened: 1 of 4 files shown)


Alexander-Wilms:
Should `#define WIN32_LEAN_AND_MEAN` be added above each `#include <Windows.h>`?

Alexander-Wilms:
Since the MinGW header files actually have lower-case file names, the rule should probably be deactivated.

IvanSavenko:
Deactivated rule.
Actual issues here will be caught by CI anyway.

PR #3668, opened 2024-03-19
make iOS work again

kambala-decapitator:
- CLauncherDirs global static crashes in `std::call_once`, see stack trace below
- setting socket buffer size throws exception which causes crash on pressing Single Player
- a few general improvements

```
Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Termination Reason: SIGNAL 6 Abort trap: 6
Terminating Process: vcmiclient [602]

Triggered by Thread:  0

Application Specific Information:
abort() called


Thread 0 name:   Dispatch queue: com.apple.main-thread
Thread 0 Crashed:
0   libsystem_kernel.dylib        	       0x1dcb09fbc __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x1ff38f680 pthread_kill + 268
2   libsystem_c.dylib             	       0x19d4b2b90 abort + 180
3   libc++abi.dylib               	       0x1ff2b9ff8 abort_message + 132
4   libc++abi.dylib               	       0x1ff2a9f74 demangling_terminate_handler() + 320
5   libobjc.A.dylib               	       0x18d7e6db4 _objc_terminate() + 160
6   libc++abi.dylib               	       0x1ff2b93bc std::__terminate(void (*)()) + 16
7   libc++abi.dylib               	       0x1ff2bc758 __cxa_rethrow + 204
8   libc++.1.dylib                	       0x1a563b4c0 std::__1::__call_once(unsigned long volatile&, void*, void (*)(void*)) + 292
9   libvcmi.dylib                 	       0x106fd6570 VCMI::VCMIDirs::get() + 72
10  vcmiclient                    	       0x10333995c CLauncherDirs::downloadsPath() + 24
11  vcmiclient                    	       0x103339700 CLauncherDirs::CLauncherDirs() + 52
12  dyld                          	       0x1b819bc04 invocation function for block in dyld4::Loader::findAndRunAllInitializers(dyld4::RuntimeState&) const::$_0::operator()() const + 168
...
```


launcher/launcherdirs.h
```diff
@@ -13,11 +13,13 @@
 class CLauncherDirs
 {
 public:
-	CLauncherDirs();
+	static void prepare();
+	static const CLauncherDirs & get();
```
IvanSavenko, line 17: Another option is to completely remove singleton and return default-constructed copy:
```cpp
static CLauncherDirs get();
```
Class does not have any sort of state so construction cost is zero

IvanSavenko (approved):

launcher/launcherdirs.h
IvanSavenko: Or replace class with a namespace.

kambala-decapitator: even better indeed

kambala-decapitator:
wow, Ubuntu 20 doesn't have access to Qt 5.15...

kambala-decapitator:
@IvanSavenko can be merged?

Issue #1685, opened 2023-03-14
Altar of sacrifice selection glitch


krs0:
**Describe the bug**
Selection square does not match creature icon size.

<img src="https://user-images.githubusercontent.com/15194321/225153699-ab195cbc-e2fe-42b9-b21e-cce416ad614c.png" width=60% height=60%>

**Version**
 - OS: Windows
 - Version: develop

**Additional context**
800x600

krs0:
This is fixed! but.... for good factions the artifact rectangle is also a little off

![image](https://user-images.githubusercontent.com/15194321/225999427-9bab782f-26bc-4c8b-ac4a-24433b2170aa.png)

IvanSavenko:
Fixed in #3669

PR #3669, opened 2024-03-19
Market window and widgets

SoundSSGood:
Should be merged after https://github.com/vcmi/vcmi/pull/3656

Reworked market window.
CTradeWindow  is splitted into widgets for each market type.
Code cleaning and unification.
Many minor bugs fixed: status bar now works on all market slots, selling an artifact does not remove the selection, graphical inaccuracies
https://github.com/vcmi/vcmi/issues/1685 fixed
Freelancers guild now works correctly.

This is still not the final PR for the market. 
Remaining for rework: CTradeableItem, CTransformerWindow, Warlock's Lab  (Not this PR).


dydzio0614:
btw @SoundSSGood do you also plan to work on artifact sets at some point? that is probably last missing artifacts major HD mod feature - using keyboard shortcuts ctrl+number you can create artifact sets for instant changing worn artifacts to saved state

SoundSSGood:
> btw @SoundSSGood do you also plan to work on artifact sets at some point? that is probably last missing artifacts major HD mod feature - using keyboard shortcuts ctrl+number you can create artifact sets for instant changing worn artifacts to saved state

yes I can work on this

client/widgets/markets/CAltarCreatures.cpp
```diff
@@ -26,61 +25,67 @@
...
+		}, bidTradePanel->slots);
+	offerTradePanel->moveTo(pos.topLeft() + Point(334, 110));
+	offerTradePanel->showcaseSlot->moveTo(pos.topLeft() + Point(395, 422));
+	offerTradePanel->showcaseSlot->subtitle->moveBy(Point(0, 3));
+	offerTradePanel->updateSlotsCallback = [this]()
+	{
+		for(const auto & altarSlot : offerTradePanel->slots)
+			updateAltarSlot(altarSlot);
+	};
+	bidTradePanel->deleteSlotsCheck = offerTradePanel->deleteSlotsCheck = std::bind(&CCreaturesSelling::slotDeletingCheck, this, _1);
```
IvanSavenko, line 69: I see that this was before this PR, but why we need 'deleteSlotsCheck'.
Why can't owner (market window) reset all its slots with new information when hero creatures change?

client/widgets/markets/CArtifactsBuying.cpp
```diff
@@ -0,0 +1,124 @@
...
+std::string CArtifactsBuying::getTraderText()
+{
+	if(bidTradePanel->isHighlighted() && offerTradePanel->isHighlighted())
+	{
+		return boost::str(boost::format(
+			CGI->generaltexth->allTexts[267]) %
+			CGI->artifacts()->getByIndex(offerTradePanel->getSelectedItemId())->getNameTranslated() %
+			bidQty %
+			(bidQty == 1 ? CGI->generaltexth->allTexts[161] : CGI->generaltexth->allTexts[160]) %
+			CGI->generaltexth->restypes[bidTradePanel->getSelectedItemId()]);
```
IvanSavenko, line 118: While I am not aware of issues with this particular string, but our MetaString is probably more preferrable to boost::format, at least for new code.

Boost format throws exceptions when format string contains less placeholders ('%s') than provided parameters. But some H3 localization sometimes remove placeholders from translated strings. 

Our MetaString avoids such issue by silently ignoring excessive replacements and might have more readable syntax since it is more integrated with our classes, e.g:
```cpp
MetaString message = MetaString::fromTextID("core.genrltxt.267");
message.replaceName(ArtifactID(offerTradePanel->getSelectedItemId()));
message.replaceNumber(bidQty);
message.replaceTextID( bidQty == 1 ? "core.genrltxt.161" : "core.genrltxt.160")
message.replaceName(GameResID(bidTradePanel->getSelectedItemId())));
return message.toString();
```

IvanSavenko (approved):

IvanSavenko:
And if you are working on fixing remaining market-related issue - you might check this one as well: #3721
Very low-priority, so up to you, not something that we need to fix ASAP.

client/widgets/markets/CAltarCreatures.cpp
SoundSSGood: The removal of the slot also happened in the old version of the market. Removing a slot is easier because it will no longer be available in some loops.

PR #3672, opened 2024-03-23
change BonusList:totalValue to return float (not int as before)

### LINKED ISSUE #3670: Basic intelligence skill adds 20% mana instead of 25%
**Describe the bug**
Limit of mana points with basic intelligence skill is calculated wrong.

**To Reproduce**
1. Create a new game. Choose a hero with basic intelligence skill (e.g. Daremyth from Tower)
2. Daremyth has 36 mana points instead of 37 (3 knowledge points multiplied by 12 instead of 12.5).

**Expected behavior**
Daremyth should have 37 mana points.

**Actual behavior**
Daremyth has 36 mana points.

**Did it work earlier?**
probably not

**Screenshots**
VCMI:
obraz

original Heroes3 game:
obraz

**Version**
 - OS: Linux
 - Version: 1.4.5

**Additional context**
CGHeroInstance::manaLimit() is calculated as PrimarySkillLevel multiplyed by valOfBonuses(BonusType::MANA_PER_KNOWLEDGE) but valOfBonuses returns int so its value is 12 instead of 12.5.
Also, BonusList::totalValue returns int.

Kris-Ja:
Change BonusList:totalValue and BonusList::valOfBonuses to return float (instead of int).

These changes fix #3670. I'm not sure it doesn't break anything.

Probably would be better to separate bonuses which should be calculated using floating point numbers from bonuses like: "MAX_LEARNABLE_SPELL_LEVEL" or make separate functions for different types of bonuses.

I made some type castings from float to int but it probably could be done better (separate function for different cases?).

### DIFF
--- a/lib/BasicTypes.cpp
+++ b/lib/BasicTypes.cpp
@@ -118,7 +118,7 @@ int AFactionMember::moraleValAndBonusList(TConstBonusListPtr & bonusList) const
 	static const std::string cachingStrMor = "type_MORALE";
 	bonusList = getBonusBearer()->getBonuses(moraleSelector, cachingStrMor);
 
-	return std::clamp(bonusList->totalValue(), maxBadMorale, maxGoodMorale);
+	return std::clamp((int)bonusList->totalValue(), maxBadMorale, maxGoodMorale);
 }
 
 int AFactionMember::luckValAndBonusList(TConstBonusListPtr & bonusList) const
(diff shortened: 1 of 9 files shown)


IvanSavenko:
Thanks for attempt, but not 100% sure whether this is good approach multiplayer-wise.

Floating points behave very poorly if you expect identical results. Compilers can (and will) optimize floating point operations differently depending on compiler, platform, optimization level and phase of moon.

This will eventually lead to client and server having different values, leading to hard to debug desync.

Need to think how to handle this better.

There is one related issue by the way - hota has fractional percentage for necromancy artifacts. But vcmi at the moment only allows integer values for percentages.

PR #3673, opened 2024-03-23
change MANA_PER_KNOWLEGDE to percentage

Kris-Ja:
fix for #3670.
This PR breaks saved games (Heroes' manaLimit is divided by 100).

### DIFF
--- a/config/gameConfig.json
+++ b/config/gameConfig.json
@@ -429,8 +429,8 @@
 				},
 				"manaPerKnowledge" :
 				{
-					"type" : "MANA_PER_KNOWLEDGE", //10 mana per knowledge
-					"val" : 10,
+					"type" : "MANA_PER_KNOWLEDGE_PERCENTAGE", //1000% mana per knowledge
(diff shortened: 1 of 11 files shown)


IvanSavenko:
> This PR breaks saved games 

You can fix this:
- add new enum value to ESerializationVersion in lib/serializer/ESerializationVersion.h
- change value of CURRENT to your new enum
- at the end of Bonus::serialize method add check like this:
```cpp
if (h.version < Handler::Version::YOUR_ENUM_VALUE && type == BonusType::MANA_PER_KNOWLEDGE_PERCENTAGE)
{
    if (valType == BonusValueType::ADDITIVE_VALUE || valType == BonusValueType::BASE_NUMBER)
        val *= 100;
}
```

DjWarmonger:
Makes sense, this approach is already used ie. for CREATURE_SPELL_POWER.

Kris-Ja:
> You can fix this:

fixed

PR #3679, opened 2024-03-24
map editor: Allow to customize hero spells

godric3:
I don't use SPELLBOOK_PRESET but I've added it to identifiers just in case 
image


IvanSavenko (approved):

mapeditor/inspector/herospellwidget.cpp
```diff
@@ -0,0 +1,133 @@
...
+		ui->customizeSpells->setChecked(false);
+		ui->tabWidget->setEnabled(false);
+	}
+}
+
+void HeroSpellWidget::initSpellLists()
+{
+	QListWidget * spellLists[] = { ui->spellList1, ui->spellList2, ui->spellList3, ui->spellList4, ui->spellList5 };
+	auto spells = VLC->spellh->objects;
+	for (int i = 0; i < 5; i++)
```
DjWarmonger, line 48: Use `SPELL_LEVELS` constant

PR #3685, opened 2024-03-25
Small fixes

dydzio0614:
Fix for following stuff:

1. Add gold icon to diplomacy window
2. Fix found cases for outdated pathfinder cache

### DIFF
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -228,25 +228,31 @@ void ApplyClientNetPackVisitor::visitSetStackType(SetStackType & pack)
 void ApplyClientNetPackVisitor::visitEraseStack(EraseStack & pack)
 {
 	dispatchGarrisonChange(cl, pack.army, ObjectInstanceID());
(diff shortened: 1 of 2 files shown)


client/NetPacksClient.cpp
```diff
@@ -228,6 +228,7 @@ void ApplyClientNetPackVisitor::visitSetStackType(SetStackType & pack)
 void ApplyClientNetPackVisitor::visitEraseStack(EraseStack & pack)
 {
 	dispatchGarrisonChange(cl, pack.army, ObjectInstanceID());
+    cl.invalidatePaths(); //it is possible to remove last non-native unit for current terrain and lose movement penalty
 }
 
 void ApplyClientNetPackVisitor::visitSwapStacks(SwapStacks & pack)
```
IvanSavenko, line 234: visitSwapStacks and visitInsertNewStack don't need path invalidation?

dydzio0614: May need investigation, did not see them triggering for some use cases I tested yesterday.

PR #3686, opened 2024-03-25
Dimension door changes

dydzio0614:
Fixes for dimension door behavior to be more in line with H3


client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -649,10 +656,10 @@ void AdventureMapInterface::onTileHovered(const int3 &mapPos)
...
 		case SpellID::DIMENSION_DOOR:
 			{
-				const TerrainTile * t = LOCPLINT->cb->getTile(mapPos, false);
-				int3 hpos = LOCPLINT->localState->getCurrentArmy()->getSightCenter();
-				if((!t || t->isClear(LOCPLINT->cb->getTile(hpos))) && isInScreenRange(hpos, mapPos))
-					CCS->curh->set(Cursor::Map::TELEPORT);
+				const TerrainTile * t = LOCPLINT->cb->getTileForDimensionDoor(mapPos, LOCPLINT->localState->getCurrentHero());
+				int3 heroPosition = LOCPLINT->localState->getCurrentArmy()->getSightCenter();
+				if(t && t->isClear(LOCPLINT->cb->getTile(heroPosition))/* && isInScreenRange(hpos, mapPos)*/)
+					CCS->curh->set(Cursor::Map::TELEPORT); //TODO: something wrong with beyond east spell range border cursor on arrogance after TP-ing near underground portal on previous day
```
IvanSavenko, line 662: IMO this entire block should be moved to DimensionDoorMechanics / ScuttleBoatMechanics
for example `bool canCastOnTile(int3 mapTile)` method

This will fix two issues:
- apparently, neither Scuttle Boat nor DD check for screen range on server. Yay, cheating.
- with this check, your CGameInfoCallback::getTileForDimensionDoor method will be unnecessary

lib/constants/NumericConstants.h
```diff
@@ -52,6 +52,7 @@ namespace GameConstants
 	constexpr ui32 BASE_MOVEMENT_COST = 100; //default cost for non-diagonal movement
 	constexpr int64_t PLAYER_RESOURCES_CAP = 1000 * 1000 * 1000;
 	constexpr int ALTAR_ARTIFACTS_SLOTS = 22;
+	constexpr int TOURNAMENT_RULES_DD_MAP_TILES_THRESHOLD = 41472; //map tiles count threshold for 2 dimension door casts with tournament rules
```
IvanSavenko, line 55: What's this magic number? I see this is 144x144x2, perhaps specify it that way?

```suggestion
	constexpr int TOURNAMENT_RULES_DD_MAP_TILES_THRESHOLD = 144*144*2; //map tiles count threshold for 2 dimension door casts with tournament rules
```

config/gameConfig.json
```diff
@@ -313,7 +313,7 @@
 			"goodLuckDice" : [ 24, 12, 8 ],
 			"badLuckDice" : [],
 			
-			// every 1 attack point damage influence in battle when attack points > defense points during creature attack
+			// every 1 attack point damage influence in battle whe	//TODO: test range, visibilityn attack points > defense points during creature attack
```
IvanSavenko, line 316: revert?

client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -957,3 +944,35 @@ void AdventureMapInterface::onScreenResize()
 	if (widgetActive)
 		activate();
 }
+
+bool AdventureMapInterface::isValidAdventureSpellTarget(int3 targetPosition, const CGObjectInstance * topObjectAtTarget, SpellID spellId)
```
IvanSavenko, line 948: Again - why not move this to CSpell or to AdventureMapMechanics class (which is member of CSpell)?

client/adventureMap/AdventureMapInterface.cpp
dydzio0614: what about the fact that this logic is still needed by clientside anyway for determining cursor type etc.?

Issue #3568, opened 2024-01-26
`Pendant of Negativity` doesn't work


drogovozDP:
**Describe the bug**
If you wear `Pendant of Negativity`, enemy heroes are still able to use `lightning bolt` and `chain lightning`.

**Game logs**
Please attach game logs: `VCMI_client.txt`, `VCMI_server.txt` etc.

**To Reproduce**
1. Wear `Pendant of Negativity`
2. Try to fight the enemy hero with these spells: `lightning bolt` and `chain lightning`

**Expected behavior**
The enemy hero won't be able to use `lightning bolt` and `chain lightning`

**Actual behavior**
He is able to use these spells

**Version**
 - OS: macOS ARM
 - Version: 1.4.2

DjWarmonger:
Works for me. Attaching test map:
[TEST Pendant of Negativity.zip](https://github.com/vcmi/vcmi/files/15141073/TEST.Pendant.of.Negativity.zip)

drogovozDP:
> Works for me. Attaching test map: [TEST Pendant of Negativity.zip](https://github.com/vcmi/vcmi/files/15141073/TEST.Pendant.of.Negativity.zip)

Your map worked correctly for me. Can you please check the situation, when you use `Pendant of Negativity` and an enemy uses lightning spells? This is exactly what happened to me

DjWarmonger:
Updated map, added enemy Solmyr. Still he doesn't cast when I equip Pendant.
[TEST Pendant of Negativity.zip](https://github.com/vcmi/vcmi/files/15213782/TEST.Pendant.of.Negativity.zip)

DjWarmonger:
Possibly already fixed, closing. Please reopen if needed.

Issue #3567, opened 2024-01-26
Berserk works on undead


drogovozDP:
**Describe the bug**
The game allows you to cast `berserk` without `sphere vulnerability`. Undead should not be affected by "mind spells"

**To Reproduce**
1. Fight any undead unit
2. Cast berserk on them

**Expected behavior**
"This spell will not affect anyone"

**Actual behavior**
This spell works

**Version**
 - OS: macOS ARM
 - Version: 1.4.2

GrafDeVan:
@GeorgeK1ng berserk works as intended in current develop

PR #3687, opened 2024-03-25
Fix BonusLimitEffect

### LINKED ISSUE #3070: Bloodlust/Precision should not be cumulative
**Describe the bug**
Bloodlust should be casted only once in each creature,now it will be cumulated for many times.
Same as precision.

**To Reproduce**
1. Split Ogre Mages for three teams.
2. Use bloodlust to one creature three times.
3. Attack value for this creature becomes +18, instead of +6.

**Did it work earlier?**
I think it is a regular error. It might never be fixed in previous version. 
ONLY Disrupting Ray is cumulative.

**Version**
 - OS: Windows
 - Version: 1.3.2

Kris-Ja:
Change BonusList::getBonus() to return bonuses with any BonusLimitEffect by default (returned only bonuses with BonusLimitEffect::NO_LIMIT previously)

Fixes #3070.
Also fixes the problem that dispel magic didn't work on bloodlust and precision.

Alternatively, appropriate BonusLimitEffect parameters could be added in a few function calls to fix this issue (if BonusList::getBonuses() should work as before)

### DIFF
--- a/lib/bonuses/Bonus.h
+++ b/lib/bonuses/Bonus.h
@@ -70,7 +70,7 @@ struct DLL_LINKAGE Bonus : public std::enable_shared_from_this<Bonus>
 	std::string stacking; // bonuses with the same stacking value don't stack (e.g. Angel/Archangel morale bonus)
 
 	CAddInfo additionalInfo;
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Looks OK I think.

Not related to #3567 / #3568 by chance?

PR #3696, opened 2024-03-29
Fix: prevent possible int16_t overflow

678098:
Ex. `port` == 40000 -> int16_t assignment overflow


kambala-decapitator (approved):
Do those methods accept unsigned integer?

nullkiller:
void NetworkHandler::connectToRemote(INetworkClientListener & listener, const std::string & host, uint16_t port)

PR #3698, opened 2024-03-29
Fix: include guards, licenses, StdInc

678098:
- Add missing licenses
- Fix file names in licenses
- Move namespace according to code style
- Add missing `#pragma once` in headers
- Make `#include "StdInc.h"` the top level include according to code style
- Removed empty unused `server/battles/ServerBattleCallback.h|.cpp`
- Smaller fixes


IvanSavenko (approved):
Thanks for clean up.
Looks OK, provided that CI passes.

PR #3707, opened 2024-04-01
include innoextract

### LINKED ISSUE #3578: Include innoextract in launcher
**Describe your proposal**
We should think about including innoextract directly into the source.

Following benefits:
- We could add a button in the launcher to extract gog files for easiest install
- This would work platform independent without extra tools (also mobile installation possible)
- Could be a solution for adding chronicles without copyright problems

I think most users use the gog packages as base.

**Additional context**
I'm not sure how easy or hard this task is. Seems we don't need lzma libs for heroes 3 installer (tested to extract with lzma disabled build).

I'm not able to do this because I know not much enough about c++ building systems.

Laserlicht:
I've tried to include innoextract directly in vcmi.

Basically it works. Only change to upstream innoextract is the `cmakelists.txt`.

Locally on Linux with GCC it works great.
grafik

Sadly there is still a problem with clang: https://github.com/Laserlicht/vcmi/actions/runs/8511235163/job/23310781125
And I think there are some hacky things in `cmakelists.txt`-changes.

Do we have some people to help out with the cmake stuff? If you don't have much experience in C++ stuff: its absolutely an nightmare.

fixes #3578

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,9 @@ if(ENABLE_ERM AND NOT ENABLE_LUA)
 	set(ENABLE_LUA ON)
 endif()
 
+include(CMakeDependentOption)
+cmake_dependent_option(ENABLE_INNOEXTRACT "Enable innoextract for GOG file extraction in launcher" ON "ENABLE_LAUNCHER" OFF)
+
 ############################################
 #        Miscellaneous options             #
 ############################################
@@ -188,6 +191,10 @@ if(ENABLE_PCH)
 else()
 	macro(enable_pch ignore)
 	endmacro(enable_pch)
+endif()
+
+if(NOT ENABLE_INNOEXTRACT)
+	add_definitions(-DVCMI_NO_INNOEXTRACT)
 endif()
 
 ############################################
@@ -448,7 +455,11 @@ endif()
 #        Finding packages                  #
 ############################################
 
-find_package(Boost 1.48.0 REQUIRED COMPONENTS date_time filesystem locale program_options system thread)
+set(BOOST_COMPONENTS date_time filesystem locale program_options system thread)
+if(ENABLE_INNOEXTRACT)
+	list(APPEND BOOST_COMPONENTS iostreams)
+endif()
+find_package(Boost 1.48.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
 
 find_package(ZLIB REQUIRED)
 # Conan compatibility
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,10 +193,6 @@ else()
 	endmacro(enable_pch)
 endif()
 
-if(NOT ENABLE_INNOEXTRACT)
-	add_definitions(-DVCMI_NO_INNOEXTRACT)
-endif()
-
 ############################################
 #        Documentation section             #
 ############################################
--- a/docs/players/Installation_iOS.md
+++ b/docs/players/Installation_iOS.md
@@ -26,7 +26,7 @@ Note: if you don't need in-game videos, you can omit downloading/copying file VI
 
 ### Step 2.a: Installing data files with GOG offline installer
 
-If you buyed HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition) you can download the files directly with the browser and install it in launcher. You need to select the exe file first and afterwards the bin file. This process takes a few seconds. Be patient.
+If you bought HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition), you can download the files directly from the browser and install them in the launcher. Select the .exe file first, then the .bin file. This may take a few seconds. Please be patient.
 
 ### Step 2.b: Installing data files with Finder or Windows explorer
 
--- a/docs/players/Installation_iOS.md
+++ b/docs/players/Installation_iOS.md
@@ -24,17 +24,21 @@ To install the signed ipa on your device, you can use Xcode or Apple Configurato
 
 Note: if you don't need in-game videos, you can omit downloading/copying file VIDEO.VID from data folder - it will save your time and space. The same applies to the Mp3 directory.
 
-### Step 2.a: Installing data files with Finder or Windows explorer
+### Step 2.a: Installing data files with GOG offline installer
+
+If you buyed HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition) you can download the files directly with the browser and install it in launcher. You need to select the exe file first and afterwards the bin file. This process takes a few seconds. Be patient.
+
+### Step 2.b: Installing data files with Finder or Windows explorer
 
 To play the game, you need to upload HoMM3 data files - **Data**, **Maps** and **Mp3** directories - to the device. Use Finder (or iTunes, if you're on Windows or your macOS is 10.14 or earlier) for that. You can also add various mods by uploading **Mods** directory. Follow [official Apple guide](https://support.apple.com/en-us/HT210598) and place files into VCMI app. Unfortunately, Finder doesn't display copy progress, give it about 10 minutes to finish.
 
-### Step 2.b: Installing data files using iOS device only
+### Step 2.c: Installing data files using iOS device only
 
 If you have data somewhere on device or in shared folder or you have downloaded it, you can copy it directly on your iPhone/iPad using Files application.
 
 Place **Data**, **Maps** and **Mp3** folders into vcmi application - it will be visible in Files along with other applications' folders.
 
-### Step 2.c: Installing data files with Xcode on macOS
+### Step 2.d: Installing data files with Xcode on macOS
 
 You can also upload files with Xcode. You need to prepare "container" for that.
 
--- a/docs/players/Installation_macOS.md
+++ b/docs/players/Installation_macOS.md
@@ -17,7 +17,7 @@ Please report about gameplay problem on forums: [Help & Bugs](https://forum.vcmi
 
 ### Step 2.a: Installing data files with GOG offline installer
 
-If you buyed HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition) you can download the files directly with the browser and install it in launcher. You need to select the exe file first and afterwards the bin file. This process takes a few seconds. Be patient.
+If you bought HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition), you can download the files directly from the browser and install them in the launcher. Select the .exe file first, then the .bin file. This may take a few seconds. Please be patient.
 
 ### Step 2.b: Installing by the classic way
 
--- a/docs/players/Installation_macOS.md
+++ b/docs/players/Installation_macOS.md
@@ -15,5 +15,11 @@ Please report about gameplay problem on forums: [Help & Bugs](https://forum.vcmi
 
 # Step 2: Installing Heroes III data files
 
+### Step 2.a: Installing data files with GOG offline installer
+
+If you buyed HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition) you can download the files directly with the browser and install it in launcher. You need to select the exe file first and afterwards the bin file. This process takes a few seconds. Be patient.
+
+### Step 2.b: Installing by the classic way
+
 1.  Find a way to unpack Windows Heroes III or GOG installer. For example, use `vcmibuilder` script inside app bundle or install the game with [CrossOver](https://www.codeweavers.com/crossover) or [Wineskin](https://github.com/Gcenx/WineskinServer).
 2.  Place or symlink **Data**, **Maps** and **Mp3** directories from Heroes III to:`~/Library/Application\ Support/vcmi/`
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -21,6 +21,10 @@
 #include "../../lib/filesystem/Filesystem.h"
 #include "../languages.h"
 
+#ifndef VCMI_NO_INNOEXTRACT
+#include "cli/extract.hpp"
+#endif
+
 FirstLaunchView::FirstLaunchView(QWidget * parent)
 	: QWidget(parent)
 	, ui(new Ui::FirstLaunchView)
@@ -32,6 +36,10 @@ FirstLaunchView::FirstLaunchView(QWidget * parent)
 
 	ui->lineEditDataSystem->setText(pathToQString(boost::filesystem::absolute(VCMIDirs::get().dataPaths().front())));
 	ui->lineEditDataUser->setText(pathToQString(boost::filesystem::absolute(VCMIDirs::get().userDataPath())));
+
+#ifdef VCMI_NO_INNOEXTRACT
+	ui->pushButtonGogInstall->hide();
+#endif
 }
 
 void FirstLaunchView::on_buttonTabLanguage_clicked()
@@ -93,6 +101,11 @@ void FirstLaunchView::on_pushButtonDataHelp_clicked()
 {
 	static const QUrl vcmibuilderWiki("https://github.com/vcmi/vcmi/blob/master/docs/players/Installation_Linux.md#install-data-using-vcmibuilder-script");
 	QDesktopServices::openUrl(vcmibuilderWiki);
+}
+
+void FirstLaunchView::on_pushButtonGogInstall_clicked()
+{
+	extractGogData();
 }
 
 void FirstLaunchView::on_comboBoxLanguage_currentIndexChanged(int index)
@@ -218,6 +231,7 @@ void FirstLaunchView::heroesDataDetected()
 
 	ui->labelDataSearch->setVisible(false);
 	ui->labelDataCopy->setVisible(false);
+	ui->pushButtonGogInstall->setVisible(false);
 
 	if(hasVCMIBuilderScript)
 	{
@@ -279,7 +293,69 @@ QString FirstLaunchView::getHeroesInstallDir()
 	return QString{};
 }
 
-void FirstLaunchView::copyHeroesData(const QString & path)
+void FirstLaunchView::extractGogData()
+{
+#ifndef VCMI_NO_INNOEXTRACT
+	QString filterExe = tr("GOG executable") + " (*.exe)";
+	QString fileExe = QFileDialog::getOpenFileName(this, tr("Select a GOG installer (exe) file..."), QDir::homePath(), filterExe);
+	if(fileExe.isEmpty())
+		return;
+
+	QString filterBin = tr("GOG bin file") + " (*.bin)";
+	QString fileBin = QFileDialog::getOpenFileName(this, tr("Select a GOG data (bin) file..."), QFileInfo(fileExe).absolutePath(), filterBin);
+	if(fileBin.isEmpty())
+		return;
+
+	if(QFileInfo(fileExe).completeBaseName() + QString("-1") != QFileInfo(fileBin).completeBaseName())
+	{
+		QMessageBox::critical(this, tr("File name error!"), tr("The filename of exe and bin has to be compatible!"), QMessageBox::Ok, QMessageBox::Ok);
+		return;
+	}
+
+	QTimer::singleShot(100, this, [this, fileExe, fileBin](){ // background to make sure FileDialog is closed...
+		QTemporaryDir dir;
+		if(dir.isValid() && QFile::exists(fileExe) && QFile::exists(fileBin)) {
+			ui->pushButtonGogInstall->setText(tr("Installing... Please wait!"));
+			QPalette pal = ui->pushButtonGogInstall->palette();
+			pal.setColor(QPalette::Button, QColor(Qt::yellow));
+			ui->pushButtonGogInstall->setAutoFillBackground(true);
+			ui->pushButtonGogInstall->setPalette(pal);
+			ui->pushButtonGogInstall->update();
+			ui->pushButtonGogInstall->repaint();
+
+			QString tmpFileExe = dir.filePath(QFileInfo(fileExe).fileName());
+			QFile(fileExe).copy(tmpFileExe);
+			QFile(fileBin).copy(dir.filePath(QFileInfo(fileBin).fileName()));
+
+			::extract_options o;
+			o.extract = true;
+
+			// standard settings
+			o.gog_galaxy = true;
+			o.codepage = (uint32_t)0;
+			o.output_dir = dir.path().toStdString();
+			o.extract_temp = true;
+			o.extract_unknown = true;
+			o.filenames.set_expand(true);
+
+			o.preserve_file_times = true; // also correctly closes file -> without it: on Windows the files are not written completly
+
+			process_file(tmpFileExe.toStdString(), o);
+
+			QStringList dirData = QDir(dir.path()).entryList({"data"}, QDir::Filter::Dirs);
+			if(dirData.empty() || QDir(QDir(dir.path()).filePath(dirData.front())).entryList({"*.lod"}, QDir::Filter::Files).empty())
+			{
+				QMessageBox::critical(this, tr("No Heroes III data!"), tr("Selected files does not contain Heroes III data!"), QMessageBox::Ok, QMessageBox::Ok);
+				ui->pushButtonGogInstall->setText(tr("Install GOG files"));
+				return;
+			}
+			copyHeroesData(dir.path(), true);
+		}
+	});
+#endif
+}
+
+void FirstLaunchView::copyHeroesData(const QString & path, bool move)
 {
 	QDir sourceRoot = QDir(path);
 	
@@ -359,7 +435,10 @@ void FirstLaunchView::copyHeroesData(const QString & path)
 		for(const QString & filename : sourceDir.entryList(QDir::Filter::Files))
 		{
 			QFile sourceFile(sourceDir.filePath(filename));
-			sourceFile.copy(targetDir.filePath(filename));
+			if(move)
+				sourceFile.rename(targetDir.filePath(filename));
+			else
+				sourceFile.copy(targetDir.filePath(filename));
 		}
 	}
 
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -306,12 +306,6 @@ void FirstLaunchView::extractGogData()
 	if(fileBin.isEmpty())
 		return;
 
-	if(QFileInfo(fileExe).completeBaseName() + QString("-1") != QFileInfo(fileBin).completeBaseName())
-	{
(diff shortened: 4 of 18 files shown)


IvanSavenko:
Many thanks for attempt. Would be great addition to 1.5 if we can implement it, especially for mobile platforms

Perhaps instead of embedding innoextract into vcmi repo we should:
- create a fork of innoextract in vcmi org (I can create it or you can use personal repo for test)
- modify CMakeLists as needed in fork
- add vcmi/innoextract repo as git submodule into vcmi/vcmi

This way we'll be able to clearly see all our changes to innoextract and easily upgrade our innoextract version when needed

Laserlicht:
> This way we'll be able to clearly see all our changes to innoextract and easily upgrade our innoextract version when needed

Yes. I had the same idea in mind. But my first goal was to bring it to work.

> Would be great addition to 1.5 if we can implement it, especially for mobile platforms

Would be great and would solve many problems caused by [Layer 8](https://www.ionos.com/digitalguide/server/security/layer-8-problem/). I think for this we would need a second QFileDialog for the bin file and copy both to a temporary folder to avoid sandbox problems.

launcher/ext/innoextract/CMakeLists.txt
```diff
@@ -0,0 +1,538 @@
...
+	include(GNUInstallDirs)
+endif()
+
+
+# Helper scrips
+
+include(CheckSymbolExists)
+
+set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake") # For custom cmake modules
+include(BuildType)
```
IvanSavenko, line 109: This include can make quite a lot of changes to build config.
That clang linkage error might come from here.

Laserlicht: Nice, thanks! Locally it's build now. Let's look what the CI do.

IvanSavenko:
MacOS CI failure - looks like innoextract or vcmilauncher was no linked to boost_iostreams.
Explicitly linking iostreams to vcmilauncher might help. Not sure how boost linking works on macOS/conan, @kambala-decapitator might know more.

Laserlicht:
Yes, I've added iostreams on several places (VCMI wasn't using this part of boost yet).
E.g. in "main" `CMakeLists.txt`.

But I think I've found it... Let's have a look...

Alexander-Wilms:
You could temporarily add `fail-fast: false` to .github/workflows/github.yml so you don't have to fix the builds one after the other:

https://stackoverflow.com/a/65647509/2278742

Laserlicht:
> wouldn't innoextract be possible to get added as git submodule?

Yes. If everything is working. :)

kambala-decapitator:
boost needs to be rebuilt on the Conan platforms

kambala-decapitator (commented):
note that this solution wouldn't work on mobile platforms out of the box because currently there's no ability to select arbitrary directory

kambala-decapitator:
also Conan has innoextract: https://conan.io/center/recipes/innoextract?version=1.9.0

IvanSavenko:
> also Conan has innoextract: https://conan.io/center/recipes/innoextract?version=1.9.0

As a separate executable from what I see. But we mostly need innoextract as part of our single binary for Android / IPhone.

Laserlicht:
> boost needs to be rebuilt on the Conan platforms

Yes. But how?

kambala-decapitator:
> > also Conan has innoextract: https://conan.io/center/recipes/innoextract?version=1.9.0
> 
> As a separate executable from what I see. But we mostly need innoextract as part of our single binary for Android / IPhone.

yeah, indeed, whatever upstream provides

kambala-decapitator:
> > boost needs to be rebuilt on the Conan platforms
> 
> Yes. But how?

currently possible only manually, can do that. So the target is to include this into 1.5?

Laserlicht:
> note that this solution wouldn't work on mobile platforms out of the box because currently there's no ability to select arbitrary directory

Yes the workaround idea would be two file selectors (first exe then bin) and copy it to temp folder (which should be openable?)

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -21,6 +21,8 @@
 #include "../../lib/filesystem/Filesystem.h"
 #include "../languages.h"
 
+#include "cli/extract.hpp"
```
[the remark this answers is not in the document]
Laserlicht: What's the right command to do this on cmake?

Laserlicht:
> > > boost needs to be rebuilt on the Conan platforms
> > 
> > 
> > Yes. But how?
> 
> currently possible only manually, can do that. So the target is to include this into 1.5?

If we get everything working (at least on desktop) until then, then: yes

kambala-decapitator:
btw I think that Windows doesn't need this functionality

Laserlicht:
> btw I think that Windows doesn't need this functionality

You mean the copying of the exe & bin? Yes - I think Windows & Linux doesn't need this. With MacOS I'm not sure.
What's to best way to difference the Platforms? Preprocessor #ifdef?

The two FileSelectors would be also good on these platforms for error detection (and the user knows that he needs two files).

launcher/firstLaunch/firstlaunch_moc.cpp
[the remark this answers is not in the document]
Laserlicht: @kambala-decapitator How can I fix that. I want to avoid changes in innoextract code itself if it isn't necessary.
Can this be done with cmake?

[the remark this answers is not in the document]
kambala-decapitator: > Wrapping only the include isn't working.

too bad, I expected it'd work. I don't think you can solve it purely with cmake. Let's leave as is then.

you could also create a thin wrapper, but not insisting on that.

kambala-decapitator:
> > btw I think that Windows doesn't need this functionality
> 
> You mean the copying of the exe & bin? Yes - I think Windows & Linux doesn't need this. With MacOS I'm not sure. What's to best way to difference the Platforms? Preprocessor #ifdef?
> 
> The two FileSelectors would be also good on these platforms for error detection (and the user knows that he needs two files).

I mean Windows doesn't need innoextract at all as you can install H3 natively. The condition whether to activate innoextract code should be in cmake, something like

```cmake
if(NOT WINDOWS)
  set(USE_INNOEXTRACT ON)
  target_compile_definitions(launcher USE_INNOEXTRACT=1)
endif()
```

(wrote without looking at docs or project, can have mistakes)

Laserlicht:
> I mean Windows doesn't need innoextract at all as you can install H3 natively. The condition whether to activate innoextract code should be in cmake, something like

Ah. Yes Windows don't need it. But even for windows it's quite a nice feature (no need for installing H3) and I would like to keep it.

I placed everything behind a cmake switch. It's now possible to completely disable it (e.g. for specific platforms in CI).

CMakeLists.txt
```diff
@@ -88,6 +88,12 @@ if(ENABLE_ERM AND NOT ENABLE_LUA)
 	set(ENABLE_LUA ON)
 endif()
 
+if(ENABLE_LAUNCHER)
+	option(ENABLE_INNOEXTRACT "Enable innoextract for GOG file extraction in launcher" ON)
+else()
+	set(ENABLE_INNOEXTRACT OFF)
+endif()
+
```
IvanSavenko, line 96: Not sure if CMake provides any better way to handle such cases, but this might be a little inconvenient:
- disable launcher
- run cmake, do stuff
- re-enable launcher, run cmake

You now have innoextract silently disabled even though this was not your intention and you never touched that option

Feel free to ignore if you can't think of better option(s)

kambala-decapitator: maybe dependent option would help https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html#module:CMakeDependentOption

kambala-decapitator:
> How complicated is it to rebuild conan boost? Would like to test the other platforms than Win & Linux.

run command and wait :) Do you have access to macOS / iOS?

Laserlicht:
@kambala-decapitator Can you do that or provide information how to get the CI working? :)

kambala-decapitator:
> @kambala-decapitator Can you do that or provide information how to get the CI working? :)

Can do later.

If you wish to try on your own: simply follow our Conan docs, but pass --build missing instead of --build never

Laserlicht:
> > @kambala-decapitator Can you do that or provide information how to get the CI working? :)
> 
> Can do later.
> 
> If you wish to try on your own: simply follow our Conan docs, but pass --build missing instead of --build never
> 
> After building locally, upload to our Dependencies repo and change URL in the CI script

Would be great if you could do this. :)

I don't have write access to this repositories and no experience in building MacOS stuff.

kambala-decapitator:
@Laserlicht uploaded updated Boost for macOS Intel as depends v1.2.1, please give it a try. I recommend disabling other CI jobs for the time being. Adjust the version here: https://github.com/vcmi/vcmi/blob/develop/CI/mac/before_install.sh#L8

kambala-decapitator:
fixed that error and started CI again

Laserlicht:
@kambala-decapitator Thanks! Works on MacOS. Can you also build iOS? :)

kambala-decapitator:
uploaded iOS 1.2.1: https://github.com/vcmi/vcmi-ios-deps/releases/tag/1.2.1

Laserlicht:
Nice! I've installed VCMI only with IPA, and GOG-files (exe+bin).

Works without problems! Thanks @kambala-decapitator

Now we need only to update MacOS ARM and the two Android's. @kambala-decapitator Is it possible to do that also for me? Then I can finish the PR.

IvanSavenko:
Two Androids would have to wait for kambala's launcher PR.
At the moment Android still uses own java launcher

rilian-la-te:
Can innoextract be optional for "normal" linux?

Laserlicht:
> Can innoextract be optional for "normal" linux?

There is a `ENABLE_INNOEXTRACT` option. If disabled it will completely build without innoextract. But default is `ON`.

Laserlicht:
> Re-enable Android CI but disable innoextract for it. It actually should be disabled already - since Androids have `ENABLE_LAUNCHER` set to OFF, so innoextract option should also be off

Fails in conan setup :(

kambala-decapitator:
> Fails in conan setup :(

either wait for the launcher PR or adjust conanfile.py to use new boost params on non-Android:

    self.options["boost"].without_iostreams = self.settings.os == "Android"

kambala-decapitator:
uploaded macOS ARM

Laserlicht:
> uploaded macOS ARM

Good work! Thanks a lot. :)

docs/players/Installation_iOS.md
```diff
@@ -24,17 +24,21 @@ To install the signed ipa on your device, you can use Xcode or Apple Configurato
 
 Note: if you don't need in-game videos, you can omit downloading/copying file VIDEO.VID from data folder - it will save your time and space. The same applies to the Mp3 directory.
 
-### Step 2.a: Installing data files with Finder or Windows explorer
+### Step 2.a: Installing data files with GOG offline installer
+
+If you buyed HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition) you can download the files directly with the browser and install it in launcher. You need to select the exe file first and afterwards the bin file. This process takes a few seconds. Be patient.
```
IvanSavenko, line 29:
```suggestion
If you bought HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition), you can download the files directly from the browser and install them in the launcher. Select the .exe file first, then the .bin file. This may take a few seconds. Please be patient.
```

docs/players/Installation_macOS.md
```diff
@@ -15,5 +15,11 @@ Please report about gameplay problem on forums: [Help & Bugs](https://forum.vcmi
 
 # Step 2: Installing Heroes III data files
 
+### Step 2.a: Installing data files with GOG offline installer
+
+If you buyed HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition) you can download the files directly with the browser and install it in launcher. You need to select the exe file first and afterwards the bin file. This process takes a few seconds. Be patient.
```
IvanSavenko, line 20:
```suggestion
If you bought HoMM3 on [GOG](https://www.gog.com/de/game/heroes_of_might_and_magic_3_complete_edition), you can download the files directly from the browser and install them in the launcher. Select the .exe file first, then the .bin file. This may take a few seconds. Please be patient.
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,69 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+	QString filterBin = tr("GOG bin file") + " (*.bin)";
+	QString fileBin = QFileDialog::getOpenFileName(this, tr("Select a GOG data (bin) file..."), QFileInfo(fileExe).absolutePath(), filterBin);
+	if(fileBin.isEmpty())
+		return;
+
+	if(QFileInfo(fileExe).completeBaseName() + QString("-1") != QFileInfo(fileBin).completeBaseName())
+	{
+		QMessageBox::critical(this, tr("File name error!"), tr("The filename of exe and bin has to be compatible!"), QMessageBox::Ok, QMessageBox::Ok);
+		return;
+	}
```
IvanSavenko, line 313: Perhaps we can modify innoextract and provide both .bin and .exe paths separately into innoextract, even if they don't match? Or too much of a trouble?

If not - error message needs to be more clear. What is 'compatible'?

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,69 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+			process_file(tmpFileExe.toStdString(), o);
+
+			QStringList dirData = QDir(dir.path()).entryList({"data"}, QDir::Filter::Dirs);
+			if(dirData.empty() || QDir(QDir(dir.path()).filePath(dirData.front())).entryList({"*.lod"}, QDir::Filter::Files).empty())
+			{
+				QMessageBox::critical(this, tr("No Heroes III data!"), tr("Selected files does not contain Heroes III data!"), QMessageBox::Ok, QMessageBox::Ok);
+				ui->pushButtonGogInstall->setText(tr("Install GOG files"));
+				return;
+			}
+			copyHeroesData(dir.path(), true);
```
IvanSavenko, line 352: We might need to clean up afterwards - delete all extracted files other than data/maps/mp3

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,69 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+			o.output_dir = dir.path().toStdString();
+			o.extract_temp = true;
+			o.extract_unknown = true;
+			o.filenames.set_expand(true);
+
+			o.preserve_file_times = true; // also correctly closes file -> without it: on Windows the files are not written completly
+
+			process_file(tmpFileExe.toStdString(), o);
+
+			QStringList dirData = QDir(dir.path()).entryList({"data"}, QDir::Filter::Dirs);
```
IvanSavenko, line 345: I think our current file copy algorithm actually looks up directories to copy in case-insensitive way. Perhaps use same approach here?

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,69 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+	if(QFileInfo(fileExe).completeBaseName() + QString("-1") != QFileInfo(fileBin).completeBaseName())
+	{
+		QMessageBox::critical(this, tr("File name error!"), tr("The filename of exe and bin has to be compatible!"), QMessageBox::Ok, QMessageBox::Ok);
+		return;
+	}
+
+	QTimer::singleShot(100, this, [this, fileExe, fileBin](){ // background to make sure FileDialog is closed...
+		QTemporaryDir dir;
+		if(dir.isValid() && QFile::exists(fileExe) && QFile::exists(fileBin)) {
+			ui->pushButtonGogInstall->setText(tr("Installing... Please wait!"));
```
IvanSavenko, line 318: Perhaps also block button while process is ongoing? Or even block whole UI if Qt allows that.

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Not needed. Temp folder cleans up on destruction of object: https://doc.qt.io/qt-6/qtemporarydir.html

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Yes the UI is already completely blocked. Not the nicest way - async would be great (as the process needs some seconds). But I think okay for now.

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: I've copied the approach from here: https://github.com/vcmi/vcmi/blob/de9dbf0cd78c300dcd1663fc41615b52aadcf0b4/launcher/firstLaunch/firstlaunch_moc.cpp#L380

Basically only to check if we have some H3-like output from innoextract (as you can also feed it with other games for example).

launcher/firstLaunch/firstlaunch_moc.cpp
IvanSavenko: Potential problem is that this temporary directory might be created on a different partition. Meaning:
- free space might be limited
- we won't be able to quickly move files from temporary directory to data directory if it is on different partition

Or it might not be a problem. Other than Linux where /tmp and /home are kind of recommended to be placed on different partitions, on rest of OS's our data directory is usually on same partition as OS and temp directories.

Laserlicht: Yes. Maybe not the perfect solution. But even under Linux it's the usual way to have the temp folder under the main partition. At least for the standard settings of many distributions-installer.

But I would leave it as it is. And maybe change later.

IvanSavenko: While temp folder aka `/tmp` is usually on main partition, `/home` including /home/me/.local/share/vcmi (which is where we save h3 data) may be on a different partition.

But yeah, let's keep as it for now. Unlikely to be an issue. 1 Gb of temp files is low amount nowadays.

IvanSavenko (approved):

kambala-decapitator:
Will review today or tomorrow

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,63 @@ QString FirstLaunchView::getHeroesInstallDir()
 	return QString{};
 }
 
-void FirstLaunchView::copyHeroesData(const QString & path)
+void FirstLaunchView::extractGogData()
+{
+#ifndef VCMI_NO_INNOEXTRACT
+	QString filterExe = tr("GOG executable") + " (*.exe)";
+	QString fileExe = QFileDialog::getOpenFileName(this, tr("Select a GOG installer (exe) file..."), QDir::homePath(), filterExe);
```
kambala-decapitator, line 300: So this works on iOS? Nice!

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,63 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+	if(fileExe.isEmpty())
+		return;
+
+	QString filterBin = tr("GOG bin file") + " (*.bin)";
+	QString fileBin = QFileDialog::getOpenFileName(this, tr("Select a GOG data (bin) file..."), QFileInfo(fileExe).absolutePath(), filterBin);
```
kambala-decapitator, line 305: You could check automatically if bin file with the same name exists and suggest to select it only if not found. But not sure if this would work on mobile platforms.

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,63 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+	if(fileBin.isEmpty())
+		return;
+
+	QTimer::singleShot(100, this, [this, fileExe, fileBin](){ // background to make sure FileDialog is closed...
+		QTemporaryDir dir;
+		if(dir.isValid() && QFile::exists(fileExe) && QFile::exists(fileBin)) {
```
kambala-decapitator, line 311: Not sure what's the point of checking for existence of files that use has just selected

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,63 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+			QString tmpFileExe = dir.filePath("h3_gog.exe");
+			QFile(fileExe).copy(tmpFileExe);
+			QFile(fileBin).copy(dir.filePath("h3_gog-1.bin"));
+
+			::extract_options o;
+			o.extract = true;
+
+			// standard settings
+			o.gog_galaxy = true;
+			o.codepage = (uint32_t)0;
```
kambala-decapitator, line 329:
```suggestion
			o.codepage = 0U;
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,63 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+			o.extract_temp = true;
+			o.extract_unknown = true;
+			o.filenames.set_expand(true);
+
+			o.preserve_file_times = true; // also correctly closes file -> without it: on Windows the files are not written completly
+
+			process_file(tmpFileExe.toStdString(), o);
+
+			QStringList dirData = QDir(dir.path()).entryList({"data"}, QDir::Filter::Dirs);
+			if(dirData.empty() || QDir(QDir(dir.path()).filePath(dirData.front())).entryList({"*.lod"}, QDir::Filter::Files).empty())
```
kambala-decapitator, line 340: Please create the QDir object only once

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,63 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+			{
+				QMessageBox::critical(this, tr("No Heroes III data!"), tr("Selected files does not contain Heroes III data!"), QMessageBox::Ok, QMessageBox::Ok);
```
kambala-decapitator, line 342:
```suggestion
				QMessageBox::critical(this, tr("No Heroes III data!"), tr("Selected files do not contain Heroes III data!"), QMessageBox::Ok, QMessageBox::Ok);
```

CMakeLists.txt
```diff
@@ -190,6 +193,10 @@ else()
 	endmacro(enable_pch)
 endif()
 
+if(NOT ENABLE_INNOEXTRACT)
+	add_definitions(-DVCMI_NO_INNOEXTRACT)
```
kambala-decapitator, line 197: Why not calling it VCMI_INNOEXTRACT? (Without negation)

kambala-decapitator (changes requested):

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,63 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+	QString fileExe = QFileDialog::getOpenFileName(this, tr("Select a GOG installer (exe) file..."), QDir::homePath(), filterExe);
+	if(fileExe.isEmpty())
+		return;
+
+	QString filterBin = tr("GOG bin file") + " (*.bin)";
+	QString fileBin = QFileDialog::getOpenFileName(this, tr("Select a GOG data (bin) file..."), QFileInfo(fileExe).absolutePath(), filterBin);
+	if(fileBin.isEmpty())
+		return;
+
+	QTimer::singleShot(100, this, [this, fileExe, fileBin](){ // background to make sure FileDialog is closed...
```
kambala-decapitator, line 309: Why is this needed?

CMakeLists.txt
IvanSavenko: Editor and launcher use -DENABLE_EDITOR / -DENABLE_EDITOR definition. So by same logic this one should be named -DENABLE_INNOEXTRACT

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Yes, works without problem. :)
Only not so nice thing is that the Filter doesn't work on iOS (I think it's not supported by iOS).

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Yes. But I think I would wait for Android and maybe simplify it later, if possible.

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: It's a kind of workaround. The unpacking process blocks the UI.

Without it: On my testing system it doesn't even close the FileSelector until the process is done. I've tried to force update the UI before extracting, but it doesn't work. :(

With this workaround everything works great. The FileSelector is closed, the button changed name and color.

CMakeLists.txt
Laserlicht: Yes, that's the reason I preferred this.

Laserlicht:
Thanks @kambala-decapitator

launcher/firstLaunch/firstlaunch_moc.cpp
kambala-decapitator: > I've tried to force update the UI 

So calling qApp->processEvents() right before starting extraction doesn't help?

Anyway, extraction should be done on background thread (use QtConcurrent) and ui should show a spinner or indefinite progress bar (if the library doesn't provide progress updates).

CMakeLists.txt
kambala-decapitator: Sorry, I don't get the reason :) 

Also, the define should be added only to the launcher target.

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -279,7 +293,65 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+		QTemporaryDir dir;
+		if(dir.isValid()) {
+			QDir tempDir = QDir(dir.path());
```
kambala-decapitator, line 312: If you will be pushing another commit, let's also improve this

```suggestion
			QDir tempDir{dir.path()};
```

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: No. Tested it and it wouldn't help. :(

Yes. Would also prefer this. But my knowledge to QT is very limited now. The first project with it. So I have no idea how to do it the clean way.

CMakeLists.txt
Laserlicht: Ah, sorry. Misunderstood it.

The reason for this way was `VCMI_NO_EXTRA_VERSION`. Used this as reference.

kambala-decapitator (approved):

kambala-decapitator:
When can it be merged?

PR #3711, opened 2024-04-03
Reindex objects to show interactive objects on top

DjWarmonger:
Some results:

Reindex 1
Reindex 2

**TODO**:

- [ ] Modify Windmill template instead of hardcoding it

### DIFF
--- a/lib/mapping/CMap.cpp
+++ b/lib/mapping/CMap.cpp
@@ -571,8 +571,6 @@ void CMap::removeObject(CGObjectInstance * obj)
 	removeBlockVisTiles(obj);
 	instanceNames.erase(obj->instanceName);
 
-	// FIXME: This is inefficient when batch-deleting objects, as objects are updated one by one.
-
 	//update indeces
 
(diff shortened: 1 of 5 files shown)


IvanSavenko:
Only used for RMG maps. Correct?

DjWarmonger:
I wonder about the editor.

DjWarmonger:
@Ivan is there actually a way to patch original templates? I tried this, and Windmills stopped to spawn altogether:
```
				"templates": {
					"avmwms0" :{
						"zIndex" : 101
					},
					"avmwndd0" : {
						"zIndex" : 101
					}
				}
```

IvanSavenko:
> @ivan is there actually a way to patch original templates? I tried this, and Windmills stopped to spawn altogether:

Might not possible actually. I think H3 templates bypass our modding system at the moment.

Why do you need it by the way?

DjWarmonger:
I'd like to tweak z-index in config instead of hardcoding windmills. But it already works the way it is.

IvanSavenko (approved):

PR #3713, opened 2024-04-04
scalingMode to config

### LINKED ISSUE #3712: Feature Request: Integer scaling (square pixels, not stretched)
**Describe your proposal**
I just installed vcmi on an Android tablet with a screen resolution of 1920x1200, which in fact is perfect for Heroes 3, because that means that I in theory can scale the image 2X since Heroes 3 originally runs at 800x600. The issue however is the current & only scaling method in vcmi, which makes all scaled images look blurry regardless of a device/display being able to do a proper 2X/3X/4X etc scaling of the original res with square, sharp pixels. 

I chose the 200% option in scaling, and after that I was searching in the menus & folders of vcmi on my Android device to change scaling method, without finding any option to do so. I have one install of vcmi from the Play Store (1.4.5) & one vcmi daily (1.5.0) install on my Android tablet.

Laserlicht:
add option to change SDL scaling mode over config file.

fixes #3712

@Gagert Try it with following change (scalingMode) in settings.json
```
	"video" : {
		"resolution" : {
			"height" : 1200,
			"scaling" : 200,
			"width" : 1920
		},
		"scalingMode" : "nearest"
	}
```

Nearest at 200% (see full, unscaled image):
image

### DIFF
--- a/client/renderSDL/ScreenHandler.cpp
+++ b/client/renderSDL/ScreenHandler.cpp
@@ -294,7 +294,7 @@ void ScreenHandler::initializeWindow()
 
 	SDL_RendererInfo info;
 	SDL_GetRendererInfo(mainRenderer, &info);
-	SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "best");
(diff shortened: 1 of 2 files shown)


Gagert:
Thank you, this worked flawlessly on my Android tablet.

Issue #2383, opened 2023-07-22
Biome system for RMG obstacles


DjWarmonger:
**Describe your proposal**
The idea is to allow customisation for obstacle sets placed on a random map. Since many mods now add adventure map decorations, it becomes quite a chaos. Additionally, some of mod content may be suitable only for particular faction or style.

So I suggest to define sets of obstacles of same type and only use a few of randomly selected sets to fill every zone. For instance, use only one type of mountains and trees for every zone, with some extra randomized decorations.

- [x] Enumerate possible obstacle groups (Mountains, trees, lakes, structures etc...)
- [x] Implement format for such obstacle set info
- [x] Write logic which resonably picks random sets of obstacles for every zone
- [ ] Define properties that allow to match obstacle sets to the zone theme: match faction, alignment (good, neutral, evil), surface/underground (for Stardust)
- [ ] Consider a way to forbid non-matching obstacle sets, for instance ice mountains and water lilies (WoG)
- [ ] Add configuration for OH3 obstacles

**Additional context**
Initial idea: https://forum.vcmi.eu/t/placing-graphically-appealing-objects-on-adventure-map/846/11?u=warmonger

PR #3714, opened 2024-04-04
Biome system implementation

DjWarmonger:
Proof of concept of #2383

**EDIT:** By now I configured pretty much all the mods I have. Results:
Biome 06
Biome 09
Biome 17


Alucard648:
Will it be possible to have biome specific battlefield backgrounds and battlefield obstacles?

DjWarmonger:
> Will it be possible to have biome specific battlefield backgrounds and battlefield obstacles?

That's interesting idea, but can be implemented separately. Battlefields are not related to obstacles ATM.

PR #3718, opened 2024-04-05
Few more tweaks for new lobby

IvanSavenko:
- Do not mark channel as unread when receiving chat history, for example when logging in.
- Do not attempt to translate usernames
- Fixed overlapping UI elements on login screen when player has no stored account details (e.g. first launch)
- Blocked creation of accounts with same name but different case (already deployed on server)

### DIFF
--- a/client/globalLobby/GlobalLobbyClient.cpp
+++ b/client/globalLobby/GlobalLobbyClient.cpp
@@ -141,7 +141,7 @@ void GlobalLobbyClient::receiveChatHistory(const JsonNode & json)
 
 		chatHistory[channelKey].push_back(message);
 
-		if(lobbyWindowPtr)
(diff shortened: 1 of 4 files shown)


pppbb:
Do you plan to have passwords for accounts?

IvanSavenko:
@pppbb 
A bit later I think.

Currently we don't use any encryption, so passwords will be sent in a plain text and can be intercepted. So for now roadmap is:
1. Add encryption
2. Add password-protected accounts

However not sure whether it will be done by 1.5 release or later.

And by the way - while this is not explained in game (added to my todo list), but even in current system you are the only one who can use your account. Currently, when you create an account, game will store login details on your system. If somebody else attempts to login into your account or create new one with same name he will get an error.

PR #3730, opened 2024-04-08
Better export command for translations

IvanSavenko:
This PR replaces existing (but inconvenient) `convert txt` command that dumps all strings into log file with more convenient options:
- `/translate` will now export all in-game strings to `<cache directory>/extracted/translation/` directory, with files split on per-mod basis.
- `/translate maps` will do the same, but will only export maps & campaigns. This command separate from "main" command since it needs to load every map, making it rather slow, compared to almost-instant `/translate`.

All files will be exported as valid json that can be used for translations as it, without manual formatting manipulations. If mod has submods, game will put all strings from submods into translation for main mod since this form is one that is generally used for translations

### DIFF
--- a/client/ClientCommandManager.cpp
+++ b/client/ClientCommandManager.cpp
@@ -183,44 +183,108 @@ void ClientCommandManager::handleNotDialogCommand()
 	LOCPLINT->showingDialog->setn(false);
 }
 
-void ClientCommandManager::handleConvertTextCommand()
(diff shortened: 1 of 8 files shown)


IvanSavenko:
It is already effectively filtered by mod - for each mod game will create separate file.

So if you want to translate particular mod all you need is to grab correct file from output directory.

Issue #3732, opened 2024-04-08
crash on AI turn


Povelitelh3:
It seems that this happens when the heroes exchange something with each other. Blue bot.
[VCMI_Client_log.txt](https://github.com/vcmi/vcmi/files/14907423/VCMI_Client_log.txt)
save and dump: https://drive.google.com/file/d/1lmBKc84-Y0P2-Z03t6Nhw7ptw3ToROYi/view?usp=drive_link
1.5.0.0a80c6c27bfa3ede04b5d15cc1878eac2bbc8ae5

> [2024-Apr-08 18:11:43.209617] DEBUG [AIGateway::makeTurn] ai - Moving hero Сэр Мюллич to tile (41 78 1)
> [2024-Apr-08 18:11:43.250617] DEBUG [runNetwork] ai - Adding query 31 - Teleport dialog query with 3 exits. Total queries count: 1
> [2024-Apr-08 18:11:43.251619] DEBUG [AIGateway::requestActionASAP::whatToDo] ai - I'll answer the query 31 giving the choice 1
> [2024-Apr-08 18:11:43.251619] DEBUG [AIGateway::requestActionASAP::whatToDo] ai - Attempted answering query 31 - Teleport dialog query with 3 exits. Request id=33. Waiting for results...
> [2024-Apr-08 18:11:43.536090] DEBUG [runNetwork] ai - Removing query 31 - Teleport dialog query with 3 exits. Total queries count: 0
> [2024-Apr-08 18:11:43.573077] DEBUG [runNetwork] ai - Adding query 45 - Exchange between heroes Сэр Мюллич (1) and ІјКІ-ХтС№ХЯ (1). Total queries count: 1
> [2024-Apr-08 18:11:43.612088] ERROR [runServer] global - Disaster happened.
> [2024-Apr-08 18:11:43.613078] ERROR [runServer] global - Reason: 0xc0000005 - EXCEPTION_ACCESS_VIOLATION at 0033:0,000,7FF,D0C,CD1,568
> [2024-Apr-08 18:11:43.613913] ERROR [runServer] global - Attempt to read from 0x0,000,000,000,000,0F0
> [2024-Apr-08 18:11:43.613913] ERROR [runServer] global - Thread ID: 10116

IvanSavenko:
Likely related to recent PR's from @SoundSSGood 
```
[...]

PR #3739, opened 2024-04-09
Fixes for found regressions

SoundSSGood:
Attempt to fix https://github.com/vcmi/vcmi/issues/3732. 
If the artifact in the source slot is simple, but the artifact in the destination slot is combined, then swapping such artifacts led to errors. Added server side check for moving artifact from destination slot. But now, without AI modification, this can lead to endless loops.  Not sure if I can modify the AI correctly. 
After all the fixes have been done, an attempt to continue the game with a save from the [issue](https://github.com/vcmi/vcmi/issues/3732.) will fail. To test, it is needed to start a new game.

### DIFF
--- a/client/widgets/CWindowWithArtifacts.cpp
+++ b/client/widgets/CWindowWithArtifacts.cpp
@@ -287,13 +287,7 @@ void CWindowWithArtifacts::artifactMoved(const ArtifactLocation & srcLoc, const 
 		// Transition state. Nothing to do here. Just skip. Need to wait for final state.
 		return;
 
-	// When moving one artifact onto another it leads to two art movements: dst->TRANSITION_POS; src->dst
-	// However after first movement we pick the art from TRANSITION_POS and the second movement coming when
-	// we have a different artifact may look surprising... but it's valid.
-
 	auto pickedArtInst = std::get<const CArtifactInstance*>(curState.value());
-	assert(!pickedArtInst || destLoc.artHolder == std::get<const CGHeroInstance*>(curState.value())->id);
-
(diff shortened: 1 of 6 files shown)


client/widgets/CWindowWithArtifacts.cpp
```diff
@@ -287,13 +287,7 @@ void CWindowWithArtifacts::artifactMoved(const ArtifactLocation & srcLoc, const
 		// Transition state. Nothing to do here. Just skip. Need to wait for final state.
 		return;
 
-	// When moving one artifact onto another it leads to two art movements: dst->TRANSITION_POS; src->dst
-	// However after first movement we pick the art from TRANSITION_POS and the second movement coming when
-	// we have a different artifact may look surprising... but it's valid.
-
```
SoundSSGood, line 293: Not relevant at the moment

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -757,6 +757,9 @@ bool CGTownInstance::allowsTrade(EMarketMode mode) const
 
 	case EMarketMode::RESOURCE_SKILL:
 		return hasBuilt(BuildingSubID::MAGIC_UNIVERSITY);
+	case EMarketMode::CREATURE_EXP:
+	case EMarketMode::ARTIFACT_EXP:
+		return false;
```
SoundSSGood, line 762: I'm not sure how to correctly connect the market to the grotto. So for now it's always false.

IvanSavenko:
OK to merge?

SoundSSGood:
> OK to merge?

I think yes. From my side these are all changes

PR #3743, opened 2024-04-10
Allow factions to be special

dydzio0614:
Feature caused by needs of modders - make faction unavailable on standard maps, allows making pseudo-factions mostly for custom maps that are not proper town equivalent for standard gameplay etc. Presence of heroes from this faction in tavern and creatures map presence etc. can be changed independently via their own "special" parameter.

### DIFF
--- a/lib/CTownHandler.cpp
+++ b/lib/CTownHandler.cpp
@@ -1068,8 +1068,7 @@ CFaction * CTownHandler::loadFromJson(const std::string & scope, const JsonNode 
 	
 	auto preferUndergound = source["preferUndergroundPlacement"];
 	faction->preferUndergroundPlacement = preferUndergound.isNull() ? false : preferUndergound.Bool();
-	auto isSpecial = source["special"];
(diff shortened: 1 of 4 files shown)


dydzio0614:
yeah, just the syntax is different as special flag is added to faction itself (no difference from gameplay point of view)

lib/CTownHandler.cpp
```diff
@@ -1068,6 +1068,8 @@ CFaction * CTownHandler::loadFromJson(const std::string & scope, const JsonNode
 	
 	auto preferUndergound = source["preferUndergroundPlacement"];
 	faction->preferUndergroundPlacement = preferUndergound.isNull() ? false : preferUndergound.Bool();
+	auto isSpecial = source["special"];
+	faction->special = isSpecial.isNull() ? false : isSpecial.Bool();
```
IvanSavenko, line 1072: No need to check for null. JsonNode will return default value if key is not preset (so false for bools)
```suggestion
	faction->special = source["special"].Bool();
```

IvanSavenko (approved):

PR #3746, opened 2024-04-11
Fix discovered undefined behavior cases

IvanSavenko:
Attempted to track recent AI and/or artifact-related crashes using gcc sanitizers. Was not able to catch those bugs specifically, but found few others.

Changes:
- enabled gcc 'hardening' flags to detect various memory corruption bugs in runtime. Not sure if they can be used on all platforms, will see what CI says. Linux distros often use these flags for all package builds, so should be safe.
- fixed use-after-free bug in EventDispatcher and in CPrologEpilogVideo
- fixed uninitialized variable in SelectionTab and in CBattleGameInterface

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -393,8 +393,10 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR NOT WIN32)
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error=array-bounds") # false positives in boost::multiarray during release build, keep as warning-only
 	endif()
 
-	# For gcc 14+ we can use -fhardened instead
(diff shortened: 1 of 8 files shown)


SoundSSGood:
> Was not able to catch those bugs specifically

To reproduce the situation with AI, use my [PR](https://github.com/vcmi/vcmi/pull/3739) and [save file](https://drive.google.com/file/d/1lmBKc84-Y0P2-Z03t6Nhw7ptw3ToROYi/view?usp=drive_link).

The AI will try to swap a simple artifact and a combined one. The server will reject this attempt because there are no free slots for the combined. So you should see the infinite loop.

IvanSavenko:
I was actually trying to catch that Android-specific crash on moving to backpack, which sound a lot like undefined behavior somewhere.

I assume you have not managed to reproduce / fix it yet?

PR #3748, opened 2024-04-11
Attempt to fix crash on moving artifact on Android

IvanSavenko:
After testing daily builds on Android I can confirm that crash comes from PR #3656
However I don't see any potentially dangerous changes in it. The only change that could have side-effects that I've found is replacement of std::vector with std::deque. Not a bad change, but perhaps we do some weird stuff with it, like assuming that its storage is continious. Or perhaps there is some bug with Android SDK since deque is very rarely used.


IvanSavenko:
And it works!

No idea what exactly wrong with deque, but I no longer have this crash on my phone.

PR #3751, opened 2024-04-11
Fix assertion failure with drag-and-drop cursor

### LINKED ISSUE #3740: Change cursor graphic assertion fails
**Describe the bug**
Assertion failed: dndObject == nullptr, file vcmi\client\gui\CursorHandler.cpp, line 66

**To Reproduce**
1. Start game
2. Open hero window
3. Pick up artifact
4. Click on backpack button
------------------
1. Start game in window mode
2. Open hero window
3. Pick up artifact
4. Click on close window button

**Expected behavior**
Assertion passes

**Actual behavior**
Assertion fails

**Version**
 - OS: [Windows, Linux]
 - Version: 1.5.0, Develop

IvanSavenko:
Artifact windows will now drop picked artifact on window deactivation instead of on destruction. Removed excessive call to setCursor since it is called before artifact windows can remove their dnd cursor. Cursor on message box still appears normally since it will be reset to pointer when new window is pushed on stack

- Fixes #3740

### DIFF
--- a/client/CMT.cpp
+++ b/client/CMT.cpp
@@ -502,8 +502,6 @@ void handleQuit(bool ask)
 		return;
 	}
 
-	CCS->curh->set(Cursor::Map::POINTER);
-
 	if (LOCPLINT)
 		LOCPLINT->showYesNoDialog(CGI->generaltexth->allTexts[69], quitApplication, nullptr);
 	else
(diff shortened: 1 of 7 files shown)


SoundSSGood:
@IvanSavenko This fix makes it impossible to transfer an artifact from the hero window to the backpack window. This is regression. I don't yet see how this can be done without assertion failure. Perhaps you can suggest other mechanisms on how to do this?

IvanSavenko:
@SoundSSGood 
Perhaps try this approach:
- in window destructor - drop artifact, as before my changes (but don't do anything related to cursor)
- in deactivate() call remove drag-and-drop cursor (but keep artifact in that transition slot)
- in activate() call restore drag-and-drop cursor if there is dragged artifact

Not sure how backpack window works right now - we might need to 'restore' cursor in it as well, on activate call, using artifact that was dragged while player was in hero window.

This should fix backpack bug and correctly restore cursor if some other dialog appears on top of hero window

PR #3759, opened 2024-04-12
creature/town description

### LINKED ISSUE #3735: Creature Biography
Add possybility to write biography for creatures.

**Additional context**
While clicking LMS on creature in creature window, small window would pop up with biography from json.

Laserlicht:
fixes  #3735

But without image for now. And also for towns.

If someone have fun to do this: We could also add biography and description for native creatures towns (can also be done with mod),

### DIFF
--- a/client/widgets/MiscWidgets.cpp
+++ b/client/widgets/MiscWidgets.cpp
@@ -466,6 +466,9 @@ void CInteractableTownTooltip::init(const CGTownInstance * town)
 			if(town->id == townId && town->builtBuildings.count(BuildingID::TAVERN))
 				LOCPLINT->showTavernWindow(town, nullptr, QueryID::NONE);
 		}
+	}, [&]{
+		if(!town->town->faction->getDescriptionTranslated().empty())
+			CRClickPopup::createAndPush(town->town->faction->getDescriptionTranslated());
 	});
 	fastMarket = std::make_shared<LRClickableArea>(Rect(143, 31, 30, 34), []()
 	{
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -1349,6 +1349,9 @@ void CCastleInterface::recreateIcons()
 	{
 		if(town->builtBuildings.count(BuildingID::TAVERN))
 			LOCPLINT->showTavernWindow(town, nullptr, QueryID::NONE);
+	}, [&]{
+		if(!town->town->faction->getDescriptionTranslated().empty())
+			CRClickPopup::createAndPush(town->town->faction->getDescriptionTranslated());
 	});
 
 	creainfo.clear();
--- a/client/windows/CCreatureWindow.cpp
+++ b/client/windows/CCreatureWindow.cpp
@@ -22,6 +22,7 @@
 #include "../widgets/Images.h"
 #include "../widgets/TextControls.h"
 #include "../widgets/ObjectLists.h"
+#include "../windows/InfoWindows.h"
 #include "../gui/CGuiHandler.h"
 #include "../gui/Shortcut.h"
 
@@ -517,6 +518,10 @@ CStackWindow::MainSection::MainSection(CStackWindow * owner, int yOffset, bool s
 	};
 
 	animation = std::make_shared<CCreaturePic>(5, 41, parent->info->creature);
+	animationArea = std::make_shared<LRClickableArea>(Rect(5, 41, 100, 130), nullptr, [&]{
+		if(!parent->info->creature->getBiographyTranslated().empty())
+			CRClickPopup::createAndPush(parent->info->creature->getBiographyTranslated());
+	});
 
 	if(parent->info->stackNode != nullptr && parent->info->commander == nullptr)
 	{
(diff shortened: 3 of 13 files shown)


IvanSavenko:
What the point to be honest?
This is not H3 feature
This is not convenience feature

This is something that should be part of moddable UI and delegated to mods instead of more hacks that we need to support

dydzio0614:
There is one important problem with this feature - it interferes with (not yet implemented) WoG creature experience window which is also triggered by right click on portrait

Laserlicht:
I think the implementation is much cleaner this way than via mods. Also with the thought that everything has to be translatable.

It is very close to how the biography for the heroes is solved in VCMI (OH3 feature). Both in terms of the content of the configs and the in-game presentation. Currently implemented as right click popup as in OH3 with heroes.

And especially with mods, I find it interesting to see what the creator has thought about the creatures and cities. See, for example, the PDF for Ruins town, where there are already corresponding descriptions.

grafik
grafik

Already existing OH3 feature (heroes) for comparison:
grafik

IvanSavenko:
VCMI is open-source engine for Heroes of Might and Magic III.
Not a new game.
There are no creature or faction biographies in H3.

Might look nice, but then next modder asks for creature portraits. Next one - for creature animations on that screen. Next one will ask for actual pre-rendered video.
Same can be said about pretty much any feature that does not exists in H3. Should we implement all of them?

Another problem is that every such optional feature means that other towns/creatures need to handle it somehow.
Okay, Ruins author made such biographies. What about H3 towns? What about every other town made by modders for vcmi? What about every other town ported to vcmi, with Cove is most obvious example.

IMO right now this looks like half-done feature with "and let modders figure out" intent that provides no notable benefits to players. This should be reserved for extended modding functionality, whether it is moddable UI, scripting, or something else.

Laserlicht:
> Might look nice, but then next modder asks for creature portraits.

They already asked for. :D
But that's not really in line with OH3.

> Another problem is that every such optional feature means that other towns/creatures need to handle it somehow. Okay, Ruins author made such biographies. What about H3 towns? What about every other town made by modders for vcmi? What about every other town ported to vcmi, with Cove is most obvious example.

Yes, for not existing it shows: Nothing... Only for those with texts...

> IMO right now this looks like half-done feature with "and let modders figure out" intent that provides no notable benefits to players. This should be reserved for extended modding functionality, whether it is moddable UI, scripting, or something else.

Current implementation is in line with OH3 - so no images, moddable or something else. It's not clear why 3DO implemented it for heroes, but not for towns/creatures.

I'm neutral here. For me it's also okay to drop this. But it was requested and I found it would be a nice feature. Maybe also ask @edeksumo

dydzio0614:
i think this feature makes sense as heries 4, heroes 5 etc. include creature biographies, this is also common feature in many games. Would not hurt to have this kind of stuff implemented to the extent that is common in other games

IvanSavenko:
Implementing stuff because "it was requested" while we have hundreds of known bugs is the reason why we ended up with "buggy modding sandbox" status.

> i think this feature makes sense as heroes 4, heroes 5 etc. include creature biographies as well. 

H4 and H5 also have perks in addition to secondary skills, recruitment in map dwelling without hero present, heroes having more direct role in battles, and many more, not to mention features that are only in H4 or in H5. Should we also implement all of that?

Also - H3 has almost 150 creatures. Who is going to write biographies for them? Who is going to translate this mass of text in languages that we support? Or is this going to be another "let modders figure out" feature?

Laserlicht:
> Also - H3 has almost 150 creatures. Who is going to write biographies for them? Who is going to translate this mass of text in languages that we support? Or is this going to be another "let modders figure out" feature?

There could be a mod "biography and descriptions for OH3 creatures", that uses the descriptions from the manual and add it to the corresponding json elements:
https://heroes3wog.net/download/[Heroes%203]%20Restoration%20of%20Erathia%20Manual.pdf

Town description are good I think. Creatures are rather "functional".

DjWarmonger:
I think Laserlicht got the point - Originally all this info was provided in printed player manual, but nowdays no one has it and players need to learn everything from the game text.

From all the mods, extra text popups seem to be least harmful.

IvanSavenko:
Description from manual are not "biographies" but gameplay guide:
> Small, frail, and armed only with their
natural scratching claws, imps and
familiars are among the weakest units
in the game. They are, however,
produced in great numbers. When
familiars are on the battlefield, they
channel 20% of spell points spent by
enemy spellcasters directly into their
hero’s spell point pool.

"weakest units in the game" Not "weakest creature in Antagarich" or something like this.
They are often extremely short, like 1 sentence-long:
> Both types of demons are capable
hand-to-hand attack units

And are shared between upgrade and downgrade.

Compare that with lore-friendly descriptions of every unit in H5

dydzio0614:
@IvanSavenko check heroes 4 unit biographies :P

edeksumo:
@Laserlicht you know I noone here, but still thinks that is great. Vcmi lacks campaign editor, and such a feature gives modders just a little way to introduce them they fractions and units. And for OH stuff I think that at some point there will be some mad-man that will implement that.

dydzio0614:
@edeksumo campaign JSON format was introduced in VCMI 1.2 or 1.3 (not sure which), so campaigns are totally possible, just same case as with random map templates - you need to configure it via text

IvanSavenko:
> @IvanSavenko check heroes 4 unit biographies :P

Rather short, true, but they are still written in lore-friendly way. Don't see anything like "weakest units in game" or "the only shooter in town".

edeksumo:
@dydzio0614 you know, json is not as user friendly as editor.  And still it's easier to click on portrait rather than play whole campain to just to know that elves have long ears.

DjWarmonger:
I think we should give it a try, code looks rather harmless. At the end of the day it's just another in-game info.

Just rename "biography" to description, as this is confusing as Ivan pointed out.

Laserlicht:
> Just rename "biography" to description, as this is confusing as Ivan pointed out.

Renamed it

PR #3764, opened 2024-04-13
[CI] use macOS 13

kambala-decapitator:


Alexander-Wilms:
Should maybe one CI matrix entry keep using macOS 12, so a bigger range is covered? Like we do it with Ubuntu 20.04 and 22.04.

kambala-decapitator:
OS itself doesn't matter much, and the currently used toolchain (Xcode 14.2) is available on both 12 and 13. With 13 the runner image deprecation becomes more distant :)

PR #3779, opened 2024-04-16
Backpack window cursor fix

SoundSSGood:
Fixed regression of transferring an artifact from the hero window to the backpack window.


SoundSSGood:
https://sonarcloud.io/project/issues?resolved=false&pullRequest=3779&id=vcmi_vcmi&open=AY7ny0BGzciAmhn0nRaJ
Maybe we need something like CIntObject::onDestroy method instead destructor.

IvanSavenko:
No need. Code is fine.
Might be a good idea to make CIntObject non-copyable / non-moveable, but nothing wrong with our destructor usage

PR #3786, opened 2024-04-17
Instant movement

IvanSavenko:
If player uses 'instant' movement speed, game will now send multiple movement requests at once, instead of waiting for server to confirm each of them.
This should help with multiplayer, where server confirmation delay is extremely noticeable due to ping.

As a side effect, it is no longer possible to stop movement with instant speed. 

There is still small delay between each move - around 20 ms in debug build, seems to be caused by UI updates. Might investigate it later, but should not be a problem in release builds

- Should alleviate #1133

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -1216,7 +1216,7 @@ bool AIGateway::moveHeroToTile(int3 dst, HeroPtr h)
 	{
 		//FIXME: this assertion fails also if AI moves onto defeated guarded object
 		//assert(cb->getVisitableObjs(dst).size() > 1); //there's no point in revisiting tile where there is no visitable object
-		cb->moveHero(*h, h->convertFromVisitablePos(dst));
+		cb->moveHero(*h, h->convertFromVisitablePos(dst), false);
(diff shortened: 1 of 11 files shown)


dydzio0614:
is this finished? because for now it will work unconditionally no matter if it is singleplayer game where likelihood of going over event increases a lot etc.

IvanSavenko:
Well, it helps a bit with speed in SP as well.
Probably should add handling for events. Oh, and need to check garrisons

dydzio0614:
there are also keymaster gates, and in hota there are quest gates...

IvanSavenko:
Slightly changed / expanded approach.
Now:
- non-instant movement speeds work as before
- instant movement speed will use batched / instant movement requests IF this is possible, and switch to 'standard' movement if batched movement may not be possible.

So if hero path goes through events / garrisons / teleporter, game will use 1st batch to reach such object, wait for object visit (if any) to proceed, like non-batched movement does, and then submit 2nd batch to finish movement

dydzio0614:
Based on code only it looks OK

To get people to test this we will need to merge first

dydzio0614:
If movement stops permanently in singleplayer after reaching border guard, monolith etc. it may be unwanted though

Issue #3780, opened 2024-04-16
Min damage greater than max damage will cause crash when attack


kdmcser:
**Describe the bug**
if a creature min and max damage are same. and stackExperience value not well designed. This bug comes.

![1](https://github.com/vcmi/vcmi/assets/8289682/a6e6a53d-aa9c-44a5-ad9d-d3280b44a4e7)

**Game logs**

[VCMI_Client_log.txt](https://github.com/vcmi/vcmi/files/15000108/VCMI_Client_log.txt)

**To Reproduce**
1. enable refugee-town mod and all submodules and wog mod
2. give hero 26 enchanter and other 99999 strong creatures
3. kill 9999 * 2 Scorpicore make enchanter to level 7
4. the next battle enchanter damage will be 19 - 18，when he attack，game crashed.

**Expected behavior**
game not crash

**Actual behavior**
game crash

**Did it work earlier?**
I dont know, I only test on daily build

**Screenshots**
no

**Version**
 - OS: Windows
 - Version: daily build

**Additional context**
no

kdmcser:
> its modder responsibility to avoid situations like this one

yes, its modder responsibility, but it is difficult for modder to find this problem since there is no hint in console.

dydzio0614:
well - as Ivan said it should produce diagnostics on mod loading

IvanSavenko:
Actually while vcmi will detect misconfiguration (e.g.  `"min" : 19, "max" : 18`) vcmi will not detect cases if this occurs due to bonuses - e.g. if you give bonus only to minimal damage.

kdmcser:
and if this problem happened, make game crash maybe unfair to users. user often download lots of mods from repos. this problem may happened in AI turn. AI take 2 - 3 months or longer to make its enchanter to level 7. The user will find his game crash in one day. most of users are unable to find which mod cause it. he only know that his game is lost and have to start a new game and dont know whether it can be happened again.

the user becomes the victim of the modder's mistake. he dont care much about whether the damage is 18 or 19. he only care if his could finish his game.

kdmcser:
I fix it it #3787

PR #3787, opened 2024-04-17
Fix creature min damage greater than max will cause game crash. #3780

kdmcser:
in 1.5.0, RNG will throw a exception when min value greater than max value. My friend encounter this problem when he play 3 mouths. The game crashed in AI turn and he doesnt know which mod cause it. And    finally ask me to fixed. 

I spend a whole night to locate this problem, asI mentioned in #3780.  I told him that refugee-town mod modify the enchanter's stack bounse and lead to this problem, And help him to make the data right, but he is still worried about if there is another mod will cause it again.

So I do a fix problem. And I found that when min damage is less than max damage, both increasing 1 mini damage and 1 max damage can increase the average damage by 0.5. when min damage greater than max damage，if I exchange min damage and max damage，this formula is still established.

Athouth the display damage will be 19 -18, it read as 19 to 18, most people will treat it as 18 ~ 19.

bless and curse are also considered in this commit.

I also display the creature name to error log,  user can report it to modder. 

Now the game can be continued finally.

### DIFF
--- a/lib/battle/DamageCalculator.cpp
+++ b/lib/battle/DamageCalculator.cpp
@@ -33,9 +33,8 @@ DamageRange DamageCalculator::getBaseDamageSingle() const
 
     if(minDmg > maxDmg)
     {
-        auto creature = VLC->creatures()->getById(info.attacker->creatureId());
-        const auto & creatureName = creature->getNamePluralTranslated();
-        logGlobal->error("Creature %s min damage %lld exceed max damage %lld.", creatureName.c_str(), minDmg, maxDmg);
+		const auto & creatureName = info.attacker->creatureId().toEntity(VLC)->getNamePluralTranslated();
+		logGlobal->error("Creature %s min damage %lld exceed max damage %lld.", creatureName, minDmg, maxDmg);
         logGlobal->error("This problem may lead to unexpected result, please report it to modder.");
         // to avoid RNG crash. and make bless and curse spell correct
         std::swap(minDmg, maxDmg);
--- a/lib/battle/DamageCalculator.cpp
+++ b/lib/battle/DamageCalculator.cpp
@@ -18,6 +18,8 @@
 #include "../mapObjects/CGTownInstance.h"
 #include "../spells/CSpellHandler.h"
 #include "../GameSettings.h"
+#include "../VCMI_Lib.h"
+
 
 VCMI_LIB_NAMESPACE_BEGIN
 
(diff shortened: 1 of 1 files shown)


dydzio0614 (approved):

lib/battle/DamageCalculator.cpp
```diff
@@ -29,6 +31,16 @@ DamageRange DamageCalculator::getBaseDamageSingle() const
 	minDmg = info.attacker->getMinDamage(info.shooting);
 	maxDmg = info.attacker->getMaxDamage(info.shooting);
 
+    if(minDmg > maxDmg)
+    {
+        auto creature = VLC->creatures()->getById(info.attacker->creatureId());
+        const auto & creatureName = creature->getNamePluralTranslated();
+        logGlobal->error("Creature %s min damage %lld exceed max damage %lld.", creatureName.c_str(), minDmg, maxDmg);
```
IvanSavenko, line 38:
```suggestion
		const auto & creatureName = info.attacker->creatureId().toEntity(VLC)->getNamePluralTranslated();
		logGlobal->error("Creature %s min damage %lld exceed max damage %lld.", creatureName, minDmg, maxDmg);
```

lib/battle/DamageCalculator.cpp
```diff
@@ -29,6 +31,15 @@ DamageRange DamageCalculator::getBaseDamageSingle() const
 	minDmg = info.attacker->getMinDamage(info.shooting);
 	maxDmg = info.attacker->getMaxDamage(info.shooting);
 
+    if(minDmg > maxDmg)
+    {
+		const auto & creatureName = info.attacker->creatureId().toEntity(VLC)->getNamePluralTranslated();
+		logGlobal->error("Creature %s min damage %lld exceed max damage %lld.", creatureName, minDmg, maxDmg);
+        logGlobal->error("This problem may lead to unexpected result, please report it to modder.");
+        // to avoid RNG crash. and make bless and curse spell correct
+        std::swap(minDmg, maxDmg);
```
Alexander-Wilms, line 40:
```suggestion
	const auto & creatureName = info.attacker->creatureId().toEntity(VLC)->getNamePluralTranslated();
	logGlobal->error("Creature %s: min damage %lld exceeds max damage %lld.", creatureName, minDmg, maxDmg);
        logGlobal->error("This may lead to unexpected results, please report it to the mod's creator.");
        // to avoid an RNG crash and make bless and curse spells work as expected
        std::swap(minDmg, maxDmg);
```

nullkiller:
Agree, better to not crash mid-battle but instead log all bonus values caused this to happen so that modders can address it later.

PR #3793, opened 2024-04-19
(NKAI) Add option to ignore troops in on-map garrisons

IvanSavenko:
Up to discussion.

Adds option for NKAI to ignore troops located in on-map garrisons, and turns it off by default.

Reasoning:
In H3, AI apparently never takes such troops. As a result - design of all maps including H3 campaigns accounts for this. However in vcmi this is not true. Leading to situations where player forced to face AI with massive amount of troops collected from garrisons that were supposed to be end-game challenge for player.
One example of such behavior is "Dungeons and Devils" campaign, scenario 3, "Fall of Steadwick".

And while VCMI can not replicate H3 behavior 1:1, differences that significantly affect game balance IMO should be accounted for.

Change only affects NKAI. VCAI can still take such troops, but since VCAI is effectively deprecated and has no user-configurable settings, I decided to keep it as it.


dydzio0614:
We already denied feature to make AI never take pandora boxes "as in H3" - so IMO this setting makes sense to exist but I probably would not turn it on by default since the goal of our AI is to provide challenge rather than aim for closest H3 compatibility and simulating its weaknesses. I do not think it is high priority for VCMI to be playable for maps that rely on original AI quirks and behaviors - wayfarer, rise of the sun king, the empire of the world 2 etc. (I recommend playing such maps in original H3). Also garrisons is only drop in the ocean, for example H3 AI also does not take armies from patrolling heroes, probably does not retreat with heroes that have patrol, etc.

IvanSavenko:
And telling casual players to "git gud *rofl*" and defeat 125 champions, 100 zealots, 200 griffins on week 2 that AI collected when player has no armies other than starting hero armies is no better than telling them to fuck off and uninstall vcmi.

dydzio0614:
Well, cant comment if they will be reasonably playable since difficulty levels are flipped upside down in VCMI, especially the lower ones are generally easier in H3. I personally never liked artificial AI limitations in H3 as well. Many missions will be way off from original feeling due to difficulty level, and also due to unusual tactics that AI will use there and even with obvious AI benefits patched i still would not recommend playing campaigns "for the first time ever" in VCMI.

dydzio0614:
I dont think that in nearby future VCMI can provide good H3 replacement for people who just wanna do campaign run without mods and that's it. I am not sure how big that group is, but any audience that heavily depends on AI changes will be hard to cater to atm.

IvanSavenko:
I suspect that very few people play H3 campaigns for the first time ever in vcmi.
However a lot of players played H3 before and now trying to play in h3 on their phones. They are not newcomers to h3 but they are also not pro players. And they don't view vcmi as different game from h3. They even call vcmi 'emulator'. They likely don't even remember how h3 AI worked.

In any case, vcmi is a game. And end goal of game AI is not to 'defeat human player', but to make game fun. And currently vcmi in such h3 campaigns is anything but fun, unless you are into Heroes 3: Dark Souls Edition.

nullkiller (approved):

nullkiller:
Btw AI does not recognize garrison as the source of army in its routines so consistently we can set this setting to not take army from garrison by default and set it on in BoostAI or/and extreme AI

Povelitelh3:
Some campaigns may be hard, it's true, but still playeble. Player has save/load in tools. And open map. It's imba too.
But, as you wish.  Difference in campaigns between oh3 and vcmi very tangible in some maps. And I understand what do you want to say.

dydzio0614:
i guess i will try to play all campaigns in VCMI at some point (on version 1.5? 1.6?)

dydzio0614 (approved):

IvanSavenko:
After more investigation in how H3 works, turned out that AI can in fact grab troops from the garrison. However specifically in "Steadwick Fall" scenario there is clearly a hardcoded check for AI to ignore them. In fact, extracting the very same map and playing it as single scenario changes AI behavior in H3.

Based on that:
- Option to disable garrisons usage is now ON by default to match H3 and behavior before this PR
- Added hardcoded check for this particular scenario to match H3 for both VCAI and NKAI

PR #3796, opened 2024-04-19
possibility to disable dimming of small windows

Laserlicht:
Setting to disable dimming of small windows.

As requested per @dydzio0614

### DIFF
--- a/client/adventureMap/AdventureMapInterface.cpp
+++ b/client/adventureMap/AdventureMapInterface.cpp
@@ -171,10 +171,10 @@ void AdventureMapInterface::show(Canvas & to)
 
 void AdventureMapInterface::dim(Canvas & to)
 {
-	auto const isBigWindow = [&](std::shared_ptr<IShowActivatable> window) { return std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 800 && std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 600; }; // OH3 fullscreen
+	auto const isBigWindow = [&](std::shared_ptr<CIntObject> window) { return std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 800 && std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 600; }; // OH3 fullscreen
 
 	if(settings["adventure"]["hideBackground"].Bool())
-		for (auto window : GH.windows().findWindows<IShowActivatable>())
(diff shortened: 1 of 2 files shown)


client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -171,18 +171,20 @@ void AdventureMapInterface::show(Canvas & to)
 
 void AdventureMapInterface::dim(Canvas & to)
 {
+	auto const isBigWindow = [&](std::shared_ptr<IShowActivatable> window) { return std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 800 && std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 600; }; // OH3 fullscreen
+
 	if(settings["adventure"]["hideBackground"].Bool())
 		for (auto window : GH.windows().findWindows<IShowActivatable>())
 		{
-			if(!std::dynamic_pointer_cast<AdventureMapInterface>(window) && std::dynamic_pointer_cast<CIntObject>(window) && std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 800 && std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 600)
+			if(!std::dynamic_pointer_cast<AdventureMapInterface>(window) && std::dynamic_pointer_cast<CIntObject>(window) && isBigWindow(window))
```
IvanSavenko, line 179: Use CIntObject everywhere and drop those casts from IShowActivatable.
We should not have any IShowActivatable that is not CIntObject

client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -171,18 +171,20 @@ void AdventureMapInterface::show(Canvas & to)
 
 void AdventureMapInterface::dim(Canvas & to)
 {
+	auto const isBigWindow = [&](std::shared_ptr<CIntObject> window) { return std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 800 && std::dynamic_pointer_cast<CIntObject>(window)->pos.w >= 600; }; // OH3 fullscreen
```
IvanSavenko, line 174: Redundant cast.
And you have typo - check for `pos.w` in both cases
```suggestion
	auto const isBigWindow = [&](std::shared_ptr<CIntObject> window) { return window->pos.w >= 800 && window->pos.h >= 600; }; // OH3 fullscreen
```

PR #3814, opened 2024-04-21
NKAI: fix patrolling heroes never retreat and town danger evaluation

nullkiller:
Patrolling heroes should never retreat
Visiting town is also considered when evaluating hero strength


dydzio0614:
hmmmm what about heroes that stand in garrisons? also something to consider but not neceseeraily as part of this PR

nullkiller:
When hero stands in garrison and attacked will it merge armies as with town?

dydzio0614:
It is not exactly about merging - I wonder if AI can be fooled to attack hero by putting 1 creature to the garrison or something

nullkiller:
No, hero is always checked in this case. Town is only because it was merging garrison and hero. In case there is garrison hero merging will not happen and it will be possible to kill visiting one

PR #3817, opened 2024-04-22
Qt launcher for Android

kambala-decapitator:
- Qt launcher now builds for Android
- old launcher and most Java code removed
- gradle project adjustments to satisfy `androiddeployqt` assumptions
- SDL for Android updated to 2.26.5
- added launcher option to ignore SSL errors when downloading mods list - disabled by default, useful on old OSs that no longer update system certificates
- launcher's icons and translations are now packaged inside the binary (using Qt resource system) for all platforms
- Android additionally packages `config` and `Mods` dirs inside the binary and unpacks that at launcher start: this replaces packing them in a zip and unpacking with Java code
- Qt code to copy oh3 files from existing directory doesn't work on Android, so it was taken from the old Java implementation


kambala-decapitator:
dev docs to be updated

nullkiller:
Abort message: 'terminating with uncaught exception of type std::runtime_error: Resource with name CONFIG/FILESYSTEM and type JSON wasn't found.' samsung

android/vcmi-app/src/main/java/eu/vcmi/vcmi/VcmiSDLActivity.java
```diff
@@ -63,10 +63,10 @@ public void hackCallNewIntentDirectly(final Intent intent)
 
     public void displayProgress(final boolean show)
     {
-        if (mProgressBar != null)
+        /*if (mProgressBar != null)
```
nullkiller, line 66: Comment not needed?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/VcmiSDLActivity.java
```diff
@@ -100,7 +98,8 @@ protected void onCreate(Bundle savedInstanceState)
     {
         super.onCreate(savedInstanceState);
 
-        if(mBrokenLibraries)
+        // TODO
```
nullkiller, line 101: TODO

kambala-decapitator: thanks, it's no longer needed

android/vcmi-app/src/main/java/eu/vcmi/vcmi/VcmiSDLActivity.java
kambala-decapitator: will bring it back, thanks

launcher/main.cpp
```diff
@@ -25,17 +26,27 @@
...
+#define EXPORT
+#endif
+
+int EXPORT main(int argc, char * argv[])
 {
 	int result;
 #ifdef VCMI_IOS
 	{
 #endif
 	QApplication vcmilauncher(argc, argv);
```
Laserlicht, line 47:
```suggestion
	QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);

	QApplication vcmilauncher(argc, argv);
```

Can you try to set this attribute. On my device (Samsung S23) it looks much better afterwards. Maybe we also have to shorten some texts.

Laserlicht:
Also add the possibility to add scrolling to the elements: https://github.com/Laserlicht/vcmi/compare/0c236da...2ac85d9

After that the only remaining issue is that after changing the language in setup it's not possible to proceed (error with resizing). And some long texts.

kambala-decapitator:
> Also add the possibility to add scrolling to the elements: https://github.com/Laserlicht/vcmi/compare/0c236da...2ac85d9

Nice trick! Will definitely add both changes.

.github/workflows/github.yml
```diff
@@ -88,16 +89,16 @@ jobs:
             preset: windows-mingw-conan-linux
             conan_profile: mingw32-linux.jinja
           - platform: android-32
-            os: ubuntu-22.04
+            os: macos-14
```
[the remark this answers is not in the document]
kambala-decapitator: If you could provide Qt host tools  (all the stuff in bin folder, plus androiddeployqt) for Linux, then we can use Linux builder again.

nullkiller:
Was not able to pass initial wizard. First of all language selector is quite laggy. But main problem is file storage. Was not able to upload files there using standard controls. Says can not detect HOMM3 files when I select a folder which previous launcher was able to accept. Maybe copying them manually may work.

nullkiller:
#06 pc 0000000000e531a0  /data/app/~~geG1DA1WdRGmf9Cv0LgjYw==/is.xyz.vcmi.daily-2av4AcFJ8wkIA7hvKitSCg==/lib/arm64/libvcmiclient_arm64-v8a.so (VCMI::CObjectClassesHandler::getHandlerFor(std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> > const&, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char

nullkiller:
Daily build from develop works fine.

launcher/CMakeLists.txt
```diff
@@ -86,38 +119,63 @@ endif()
 # to always look for includes there:
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 
-if(TARGET Qt6::Core)
-	qt_wrap_ui(launcher_UI_HEADERS ${launcher_FORMS})
+if(ENABLE_SINGLE_APP_BUILD OR ANDROID)
```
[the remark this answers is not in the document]
kambala-decapitator: because the only possible way on Android is add_library regardless of anything, will keep

android/vcmi-app/src/main/res/values/strings.xml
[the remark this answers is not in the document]
kambala-decapitator: thanks. one line is still used in the manifest, removed everything else.

kambala-decapitator:
applied all suggestions

@nullkiller that was upstream issue

kambala-decapitator:
still need to update dev docs and rename apk to include commit hash / branch name

IvanSavenko:
Not sure if you've tested it, but there is also `bundle_release` job that generates .aab for Google Play. Not sure if it will work as it or if it needs adjustments

kambala-decapitator:
> Not sure if you've tested it, but there is also `bundle_release` job that generates .aab for Google Play. Not sure if it will work as it or if it needs adjustments

only signing fails locally because I don't have the key. I've made adjustments to CI, should work I believe. In the worst case let's fix it after upload attempt fails.

IvanSavenko:
> Was not able to pass initial wizard. First of all language selector is quite laggy. But main problem is file storage. Was not able to upload files there using standard controls. Says can not detect HOMM3 files when I select a folder which previous launcher was able to accept. Maybe copying them manually may work.

I also have this problem right now.
Selecting directory with data / maps / mp3 directories from H3 results in "data not found" message.
And gog.com installer files are not selectable if I try to pick innoextract option.
Latest beta build on PC works, so does not looks like an upstream regression.

Manual copy still works.
I have not tried data import with previous build, so might not be a new problem.

===

However if files are copied manually, everything seems to be working as intended! Yeah, some things are not very suitable for touchscreens, and I would tone down that "overshot" effect when you try to scrolling too far, especially on mod list.

But these things are quite minor and certainly something we can live with.

kambala-decapitator:
the gog fix made it to the beta branch, should I retarget PR?

IvanSavenko:
Sure. Don't see any problems with targeting beta, it is not like we have to merge it in 1.5.0 if you do so.

But probably won't fix not working "copy existing data" option.

launcher/main.cpp
```diff
@@ -51,6 +51,15 @@ int MAIN_EXPORT main(int argc, char * argv[])
 	MainWindow mainWindow;
 	mainWindow.show();
 
+#ifdef VCMI_ANDROID
+	// changing language causes window to increase height over the bounds, force the height back to a proper value
+	// TODO: check in 5.15.14+ / 6 if the hack is still needed
+	QObject::connect(vcmilauncher.focusWindow(), &QWindow::heightChanged, []{
+		auto appWindow = qApp->focusWindow();
+		appWindow->resize(appWindow->width(), appWindow->screen()->availableSize().height());
```
Laserlicht, line 62: Can you do the same thing on width?
Height is now okay. But with still have the problem. :D

IvanSavenko:
Tested new build:
- "Copy existing data" still fails for me with the same message
- "Install gog files" now works as expected. Takes a bit too long with no indication of progress, but that's unrelated to this PR.

launcher/main.cpp
kambala-decapitator: sure. I faced only height issue on modern android (just like everybody else), didn't think that width would also be a problem :)

dydzio0614:
BTW i think there should be a way to run device simulator where you can pick device and see how app looks on it?

So you can pick for example google nexus, motorola g7 (for testing notch), some samsung galaxy tablet etc.

launcher/main.cpp
kambala-decapitator: done

Laserlicht: fixed. :)

kambala-decapitator:
> BTW i think there should be a way to run device simulator where you can pick device and see how app looks on it?

yes, should work somehow, but first you need to build all dependencies for the emulator (AVD)

Laserlicht:
Tested it on 3 devices:
* Galaxy S23: Runs great!
* Galaxy Tab S6 lite (2020): Runs good. The Mod list is a little bit slow. But usable.
* Motorola Moto X (2013):
  * innoextract needs 3 min, but works!
  * QT launcher works (for the age of the device) well
  * only mod list is laggy, but fully usable
  * some big mods installations crashes app (I think RAM - only 2GB is full), probably the same problem with old launcher

Should we wait for 1.5.x or already add in 1.5.0? It's a huge improvement I think.

IvanSavenko:
> Should we wait for 1.5.x or already add in 1.5.0? It's a huge improvement I think.

Not sure. There is still that not working copying of existing files. And I'd rather avoid large last-minute changes to 1.5.0.
Personally I'd rather add it to 1.5.2:
1.5.1 - fixes for any critical issues we find in initial release, fixes for new crashes. Should be released in 1 week.
1.5.2 - any new features that won't make into 1.5.0, probably 1-2 weeks after 1.5.1. Probably most suitable option for launcher.

Any other opinions?

kambala-decapitator:
> There is still that not working copying of existing files

started investigating that. So far I saw this (on 4.4):
```
05-02 21:59:00.056 D/Instrumentation(18575): checkStartActivityResult  :Intent {  }
05-02 21:59:00.056 D/Instrumentation(18575): checkStartActivityResult  inent is instance of inent:
05-02 21:59:00.061 W/System.err(18575): android.content.ActivityNotFoundException: No Activity found to handle Intent {  }
05-02 21:59:00.066 W/System.err(18575): 	at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:1660)
05-02 21:59:00.066 W/System.err(18575): 	at android.app.Instrumentation.execStartActivity(Instrumentation.java:1430)
05-02 21:59:00.066 W/System.err(18575): 	at android.app.Activity.startActivityForResult(Activity.java:3532)
05-02 21:59:00.066 W/System.err(18575): 	at android.app.Activity.startActivityForResult(Activity.java:3493)
05-02 21:59:00.066 W/System.err(18575): 	at org.qtproject.qt5.android.QtNative.startQtApplication(Native Method)
05-02 21:59:00.066 W/System.err(18575): 	at org.qtproject.qt5.android.QtNative$7.run(QtNative.java:446)
05-02 21:59:00.066 W/System.err(18575): 	at org.qtproject.qt5.android.QtThread$1.run(QtThread.java:61)
05-02 21:59:00.066 W/System.err(18575): 	at java.lang.Thread.run(Thread.java:841)
05-02 21:59:00.086 W/libvcmiclient_armeabi-v7a.so(18575): Empty filename passed to function
```

Seems that there's some issue with folder selection (unlike file selection), will also check on a clean project.

> And I'd rather avoid large last-minute changes to 1.5.0

fully agree

> Personally I'd rather add it to 1.5.2

sounds good. Will also have time to check about the slow scroll of mods list (happens on all platforms basically).

IvanSavenko:
> Seems that there's some issue with folder selection (unlike file selection), will also check on a clean project.

That reminds me - there is one hack-ish logic in there.
If player selects "Data" folder, as in, one that contains .lod archives (after all, we did ask player to select "Heroes 3 data"), Launcher will try to go one level up to find correct directory in there so we can also grab Maps and Mp3's
Works fine on PC but Android permissions will likely interfere with this logic.

> sounds good. Will also have time to check about the slow scroll of mods list (happens on all platforms basically).

Personally I would consider getting rid of vertical scrolling in there - remove "version" column, possibly drop "mod type" column, resize name column to fit.
This is the only control with both vertical and horizontal scroll and the only slow control. Not sure if this is coincidence or related.

Could also be caused by having like 100 items in it. Not much can be done here though.

Laserlicht:
Okay for me.

But then we should also bump repo url for android app. One last time. :)

IvanSavenko:
Yeah, will do that alongside with the rest of my usual release prep

nullkiller:
Campaigns and maps work now.

nullkiller:
My Samsung still cannot copy homm3 data. Says can not detect.

kambala-decapitator:
added copying files from existing directory in Java

IvanSavenko:
@kambala-decapitator 
What's the status of this PR? 
Any remaining issues that needs to be fixed before merge?

If only minor stuff is left then we can merge it as it and work on minor stuff later.

kambala-decapitator:
Left to upload new version of Qt to pre-built deps and we're done

launcher/main.cpp
kambala-decapitator: FYI seems it's no longer an issue in 5.15.14, so I dropped this hack

kambala-decapitator:
new Qt not uploaded yet due to a small blocker with a Conan PR (https://github.com/conan-io/conan-center-index/pull/23944 has been abandoned, so not sure if I can use its changes), but I think I'll know how to proceed in a few days

kambala-decapitator:
everything's ready, Android builds should be green now

IvanSavenko:
Strange, I can't install 64-bit apk on my phone.
"Not possible to install application on device" (may not be exact translation)

Never had this issue before - either from apk's from this branch or from apk's from other branches. Beta build apk installs fine.

dydzio0614:
maybe app signing got broken

kambala-decapitator:
this issue was also reported on Discord

seems it wasn't wise to ignore [weird] lint errors, will investigate what's wrong

IvanSavenko:
In AndroidManifest.xml I see
```
	<uses-sdk
		android:minSdkVersion="21"
		android:targetSdkVersion="android-33"/>
```

Previously it was
```
	<uses-sdk
		android:minSdkVersion="19"
		android:targetSdkVersion="33"/>
```

kambala-decapitator:
targetSdkVersion is the reason most probably, thanks!

dydzio0614:
googling shows this: testOnly="true" means the APK can only be installed via adb (which Android Studio uses)

kambala-decapitator:
let's see if it's fixed, then will remove the temp commit

IvanSavenko:
Apk installs, but translations are broken? Changing language does nothing.

Everything else seems to be working so far

kambala-decapitator:
> Apk installs, but translations are broken? Changing language does nothing.
> 
> Everything else seems to be working so far

yeah, sorry, copied fix blindly from #4091 and didn't test it. Reverted it and removed the temp commit.

IvanSavenko:
Seems to be working.

Issues that I've noticed:
- using "Import existing files" leads to UI freeze, with no indication of progress
- Checkboxes are extremely tiny and almost impossible to hit
- Some popup windows (mod screenshots, updates check) look weird due to no window borders
- Buttons to open directories in "Help" tab do nothing
- It is no longer possible to configure "relative cursor mode" since Launcher has no entry in settings for it
- Mod list lags a lot when dragged

Any way to fix/alleviate 1st issue? Progress bar would be great, but even simple "copying data, please wait" message would be better than UI freeze. 

In any case don't see anything critical. Will try to look into these issues after merge. Perhaps post on Discord and if nothing critical is found - merge tomorrow?

kambala-decapitator:
> * using "Import existing files" leads to UI freeze, with no indication of progress

yes, I already mentioned it above

> Buttons to open directories in "Help" tab do nothing
> Mod list lags a lot when dragged

same on iOS

> Any way to fix/alleviate 1st issue? Progress bar would be great, but even simple "copying data, please wait" message would be better than UI freeze.

it's doable, but would need more Java code (and possibly notifying Qt app to show the progress bar / message if we want to display it from Qt)

IvanSavenko:
> it's doable, but would need more Java code (and possibly notifying Qt app to show the progress bar / message if we want to display it from Qt)

Would be great to fix that for 1.5.3. Progress bar would be amazing, but even simple "Copying data, please wait" message will do, whether via Java or via Qt.
This is going to be one of the first things new player is going to see after all.

Can look into rest of issues I've mentioned myself, but probably not this one unfortunately.

kambala-decapitator:
ok, will try to implement that

IvanSavenko:
Perhaps merge as it and fix UI freeze in separate PR? What do you think?

If yes - can you rebase PR to fix that merge conflict?

kambala-decapitator:
> Perhaps merge as it and fix UI freeze in separate PR? What do you think?
> 
> 
> 
> If yes - can you rebase PR to fix that merge conflict?

Fine for me. Will rebase in the evening.

launcher/main.cpp
Laserlicht: @kambala-decapitator We should introduce it again. The bug is not gone. :(
Change between some languages (e.g. french) after some constellations (e.g. german -> french) the button is partially out of screen. Even for the next pages.

kambala-decapitator: Ok, let's add back. Couldn't reproduce when switching even to French.

Laserlicht: @kambala-decapitator Maybe it has to do with the control bar of android...

https://github.com/vcmi/vcmi/assets/13953785/6d276734-89a9-4555-b5ac-fa49cd35a4db

kambala-decapitator: That's possible. Qt 5 made various changes to computing available screen area, maybe some are available only in 5.15.15+.

That's why I want to switch android 64-bit to Qt 6 to gain all the latest fixes to modern Android versions.

kambala-decapitator: #4114

PR #3833, opened 2024-04-26
Artifacts costumes

SoundSSGood:


lib/networkPacks/PacksForServer.h
```diff
@@ -429,6 +429,31 @@ struct DLL_LINKAGE ManageBackpackArtifacts : public CPackForServer
...
+	ManageEquippedArtifacts() = default;
+	ManageEquippedArtifacts(const ObjectInstanceID & artHolder, const size_t costumeIdx, bool saveCostume = false)
+		: artHolder(artHolder)
+		, costumeIdx(costumeIdx)
+		, saveCostume(saveCostume)
+	{
+	}
+
+	ObjectInstanceID artHolder;
+	size_t costumeIdx;
```
IvanSavenko, line 443: Please use only fixed-width types in serialization (e.g. uint8_t / uint32_t / etc, whatever is suitable).
This is required for cross-platform gameplay

lib/networkPacks/PacksForClient.h
```diff
@@ -1288,6 +1288,30 @@ struct DLL_LINKAGE ChangeObjectVisitors : public CPackForClient
 	}
 };
 
+struct DLL_LINKAGE ChangeArtifactsCostume : public CPackForClient
+{
+	std::map<ArtifactPosition, ArtifactID> costumeSet;
+	size_t costumeIdx = 0;
```
IvanSavenko, line 1294: Also here

client/gui/CIntObject.cpp
```diff
@@ -314,12 +314,20 @@ CKeyShortcut::CKeyShortcut(EShortcut key)
...
+}
+
 void CKeyShortcut::keyPressed(EShortcut key)
 {
 	if( assignedKey == key && assignedKey != EShortcut::NONE && !shortcutPressed)
 	{
 		shortcutPressed = true;
 		clickPressed(GH.getCursorPosition());
+		if(keyPressedCallback)
+			keyPressedCallback();
```
IvanSavenko, line 330: Why not override clickPressed instead of callback?

SoundSSGood: CArtifactsOfHeroMain does not inherit from CKeyShortcut. CArtifactsOfHeroMain contains a vector of handlers for each key. That's why a callback is needed.

SoundSSGood: But it is possible to do it through inheritance. Perhaps this option is better

IvanSavenko: Inheritance is probably more clear.
Callback can be set by any entity from anywhere (and pain to debug since debugger loves going into std::function internals). And in case of inheritance list of functions that can be called is way easier to find.

IvanSavenko:
I see that costumes are per-player, and not per-hero. Works fine for me either way, but is this HD Mod logic? Haven't seen how it works in HD mod actually.

Also - I assume that feature is complete / mostly complete? If yes - then probably safe to include in 1.5 release. Will create beta branch soon, release is planned for May, 10th.

SoundSSGood:
> but is this HD Mod logic?

Yeah

> Also - I assume that feature is complete / mostly complete?

I think the PR will be ready tomorrow

SoundSSGood:
Mostly ready. Costumes are also available to be read on client side. If anyone wants, they can implement a GUI for viewing.

lib/CPlayerState.h
```diff
@@ -63,6 +63,7 @@ struct DLL_LINKAGE PlayerState : public CBonusSystemNode, public Player
 	std::vector<ConstTransitivePtr<CGDwelling> > dwellings; //used for town growth
 	std::vector<QuestInfo> quests; //store info about all received quests
 	std::vector<Bonus> battleBonuses; //additional bonuses to be added during battle with neutrals
+	std::map<size_t, std::map<ArtifactPosition, ArtifactID>> costumesArtifacts;
```
IvanSavenko, line 66:
```suggestion
	std::map<uint32_t, std::map<ArtifactPosition, ArtifactID>> costumesArtifacts;
```

PR #3840, opened 2024-04-27
Allow limited interaction with map during enemy turn

IvanSavenko:
This PR allows limited interaction with map during enemy turn - common request from MP players.

- Server now blocks all actions made by inactive players, with few exceptions like chat messages.
- Game will not attempt to track actions of other human players even if enemy speed is below "instant"
- It is now possible to interact with map - change heroes paths, select heroes / towns, open town/hero screens, etc
- Some common actions, like moving heroes, arranging troops, recruiting troops are also blocked on client to avoid invalid requests from server. 
- More rare actions may result in "server error" message and need to be tracked down, probably via player reports

Need to think about other possible edge cases which may need handling


dydzio0614:
some actions I can think of: attempting to buy spellbook, attempting to buy building, attempting to use inferno town gate building, attempting to recruit / buy artifact / trade resources, attempting to interact with map object if hero stands on one, attempting to change troops/artifacts, attempting to give artifact to commander, attempt to move commander's artifact back to hero, attempt to cast spell with hero, attempt to rearrange town creatures

dydzio0614:
and about attempt to track actions... I think one of F buttons (F6?) does this for spectator move, so maybe I would use same button to allow locking camera to track AI actions, leaving unlocked as with this PR by default

IvanSavenko:
AI actions are still tracked as before (except for simturns). In total we have 3 different modes for handling turns of other players:
- during simturns, when our player also acts: actions of human players and AI are ignored (AI needs simturns AI on for this to happen). Existing logic, was not changed in PR.
- our player ended turn, another human player is acting (MP): other player actions are ignored, our player can scroll map / open windows / etc. This is new logic from this PR.
- our player ended turn, AI player is acting (SP): AI actions are tracked, interface of our player is locked. Existing H3 logic.

dydzio0614:
I mean toggle to lock/unlock tracking opponent player actions after we ended turn

IvanSavenko:
Blocked actions from your list that I've missed.

Depending on feedback we can unblock "safe" actions - for example, I don't see any downsides in allowing resource-resource trades, and several similar actions. Can be adjusted later at any point.

PR #3845, opened 2024-04-27
innoextract mobile optimization

Laserlicht:
- shows colored button also on iOS and Android
- mobile platforms can select any files now (fixes Android problem) -> checked by file ending afterwards
- show information which file to select before opening file manager (as there is no title bar: mobile player don't know what to select currently)

### DIFF
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -305,7 +305,7 @@ void FirstLaunchView::extractGogData()
 	QString fileExe = QFileDialog::getOpenFileName(this, tr("Select a GOG installer (exe) file..."), QDir::homePath(), filterExe);
 	if(fileExe.isEmpty())
 		return;
-	if(!fileExe.endsWith(".exe", Qt::CaseInsensitive)
+	if(!fileExe.endsWith(".exe", Qt::CaseInsensitive))
 	{
 		QMessageBox::critical(this, tr("Invalid file selected"), tr("You have to select a GOG installer (exe) file!"));
 		return;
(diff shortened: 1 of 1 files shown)


kambala-decapitator (commented):
looks good, but please extract the new code with dialog + extension check to a separate function, as exe/bin code is identical except the extension and dialog text

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -296,29 +296,46 @@ QString FirstLaunchView::getHeroesInstallDir()
 void FirstLaunchView::extractGogData()
 {
 #ifdef ENABLE_INNOEXTRACT
-	QString filterExe = tr("GOG executable") + " (*.exe)";
-	QString fileExe = QFileDialog::getOpenFileName(this, tr("Select a GOG installer (exe) file..."), QDir::homePath(), filterExe);
+
+	auto fileSelection = [this](QString type, QString filter, QString startPath = {}) {
+		QString titleSel = tr("Select a ") + filter + tr(" file...");
+		QString titleErr = tr("You have to select a ") + filter + tr(" file!");
```
kambala-decapitator, line 302: never do this for translatable strings, always use placeholders:
```suggestion
		QString titleSel = tr("Select %1 file...", "param is file extension").arg(filter);
		QString titleErr = tr("You have to select %1 file!", "param is file extension").arg(filter);
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -296,29 +296,46 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+		else if(!file.endsWith("." + type, Qt::CaseInsensitive))
+		{
+			QMessageBox::critical(this, tr("Invalid file selected"), titleErr);
+			return QString{};
+		}
+
+		return file;
+	};
+
+	QString fileExe = fileSelection("exe", tr("GOG installer") + " (*.exe)");
```
kambala-decapitator, line 319: FYI in rare cases like this placeholder can be omitted, but still it might be an issue e.g. for RTL languages

kambala-decapitator (changes requested):

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Thanks! Understood.

kambala-decapitator (approved):

PR #3846, opened 2024-04-28
Handle triple resources in creature recruitment window

DjWarmonger:
The best I could do:

image

Don't ask me what happens if there are more than 3 resources ¯\_(ツ)_/¯


dydzio0614:
good enough for me

PR #3861, opened 2024-04-30
Pvp options

Laserlicht:
Implementation like in HD mod except:
- Town area is scrollable (no town limit)
- no highlight of selected town (no green icons in OH3)
- No gold transfer yet (not part of this PR)

Also adds possibility to translate strings within messages

https://github.com/vcmi/vcmi/assets/13953785/a3c78365-4a69-4990-b3c8-8cb8d6c76bd6

### DIFF
--- a/client/GameChatHandler.cpp
+++ b/client/GameChatHandler.cpp
@@ -47,7 +47,9 @@ void GameChatHandler::sendMessageGameplay(const std::string & messageText)
 void GameChatHandler::sendMessageLobby(const std::string & senderName, const std::string & messageText)
 {
 	LobbyChatMessage lcm;
-	lcm.message = messageText;
+	MetaString txt;
+	txt.appendRawString(messageText);
+	lcm.message = txt;
 	lcm.playerName = senderName;
 	CSH->sendLobbyPack(lcm);
 	CSH->getGlobalLobby().sendMatchChatMessage(messageText);
@@ -63,8 +65,6 @@ void GameChatHandler::onNewLobbyMessageReceived(const std::string & senderName, 
 
 	auto * lobby = dynamic_cast<CLobbyScreen*>(SEL);
 
-	std::string messageTextReplaced = translationReplace(messageText);
-
 	// FIXME: when can this happen?
 	assert(lobby);
 	assert(lobby->card);
@@ -74,14 +74,14 @@ void GameChatHandler::onNewLobbyMessageReceived(const std::string & senderName, 
 		MetaString formatted = MetaString::createFromRawString("[%s] %s: %s");
 		formatted.replaceRawString(TextOperations::getCurrentFormattedTimeLocal());
 		formatted.replaceRawString(senderName);
-		formatted.replaceRawString(messageTextReplaced);
+		formatted.replaceRawString(messageText);
 
 		lobby->card->chat->addNewMessage(formatted.toString());
 		if (!lobby->card->showChat)
 				lobby->toggleChat();
 	}
 
-	chatHistory.push_back({senderName, messageTextReplaced, TextOperations::getCurrentFormattedTimeLocal()});
+	chatHistory.push_back({senderName, messageText, TextOperations::getCurrentFormattedTimeLocal()});
 }
 
 void GameChatHandler::onNewGameMessageReceived(PlayerColor sender, const std::string & messageText)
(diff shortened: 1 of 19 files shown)


dydzio0614:
Since I think access to this panel should not block chat window etc... probably optimal place for it would be here at the moment
obraz

dydzio0614:
Also if you do not want to deal with this feature then just skip it :P

Laserlicht:
> Since I think access to this panel should not block chat window etc...

Yes, but I also think that we should bot flood this ui with multiplayer stuff. It's also used by peoples which never play multiplayer. Because of this it would be great to limit it to one button

dydzio0614:
My initial idea was to disable feature completely unless game is multiplayer / hot seat and number of players on map is more than 4

We can try to go the way you show but i am almost sure the first time feature gets used in pvp it will get github report due to "not being as convenient to use as in hota"

Laserlicht:
@dydzio0614 What about this space. Do we need to display this, when chat is opened?

grafik

dydzio0614:
it doesn't serve much purpose so as long as difficulty was already picked it could be used for something else

Hard for me to judge final convenience of final solution you have in mind :P

Laserlicht:
@dydzio0614 :D

grafik

And without chat as OH3.

dydzio0614:
hmm but how are we going to fit 12+ towns this way :D

dydzio0614:
VCMI multiplayer tournament will have 13 towns... for a start :P

Laserlicht:
> hmm but how are we going to fit 12+ towns this way :D

Scrollbar for town list?

dydzio0614:
oh ok :P

Laserlicht:
Do you think something like this would be okay?

dydzio0614:
sounds good enough for me, in worst case we will have some logic of the panel itself :D

Laserlicht:
@dydzio0614 See first post

dydzio0614:
looks good

Laserlicht:
Ready now

dydzio0614 (approved):

server/processors/PlayerMessageProcessor.cpp
```diff
@@ -42,7 +42,7 @@ void PlayerMessageProcessor::playerMessage(PlayerColor player, const std::string
 	if (handleCheatCode(message, player, currObj))
 	{
 		if(!gameHandler->getPlayerSettings(player)->isControlledByAI())
-			broadcastSystemMessage(VLC->generaltexth->allTexts[260]);
+			broadcastSystemMessage("~~core.genrltxt.260~~");
```
IvanSavenko, line 45: If you need client-side translations you should send MetaString to client instead

client/lobby/CSelectionBase.cpp
```diff
@@ -373,6 +393,120 @@ void CChatBox::addNewMessage(const std::string & text)
...
+void TownSelector::updateListItems()
+{
+	OBJ_CONSTRUCTION;
+	int line = slider ? slider->getValue() : 0;
+	int x_offset = slider ? 0 : 8;
+	
+	towns.clear();
+	townsArea.clear();
+
+	int x = 0, y = 0;
```
IvanSavenko, line 475:
```suggestion
	int x = 0;
	int y = 0;
```

client/lobby/CSelectionBase.cpp
```diff
@@ -373,6 +393,120 @@ void CChatBox::addNewMessage(const std::string & text)
...
+}
+
+TownSelector::TownSelector(const Point & loc)
+{
+	OBJ_CONSTRUCTION;
+	pos += loc;
+	setRedrawParent(true);
+
+	int count = 0;
+	CGI->factions()->forEach([this, &count](const Faction *entity, bool &stop){
```
IvanSavenko, line 447: You can use `getDefaultAllowed()` and use standard for loop over returned vector. Tends to be more clear and easy to debug than methods with lambdas

client/GameChatHandler.cpp
```diff
@@ -86,22 +90,43 @@ void GameChatHandler::onNewGameMessageReceived(PlayerColor sender, const std::st
...
+	while(std::regex_search(searchStart, txt.cend(), match, expr) )
+	{
+		result += match.prefix();
+		result += VLC->generaltexth->translate(match.str(1));
+		
+		searchStart = match.suffix().first;
+		tmp_suffix = match.suffix();
+	}
+	return result.empty() ? txt : result + tmp_suffix;
+}
```
IvanSavenko, line 132: Replace with MetaString

server/NetPacksLobbyServer.cpp
```diff
@@ -366,3 +369,42 @@ void ApplyOnServerNetPackVisitor::visitLobbyForceSetPlayer(LobbyForceSetPlayer &
...
+{
+	std::vector<FactionID> allowedTowns;
+
+	VLC->townh->forEach([pack, &allowedTowns](const Faction *entity, bool &stop){
+		if(!entity->hasTown())
+			return;
+		if(std::find(pack.bannedTowns.begin(), pack.bannedTowns.end(), entity->getId()) == pack.bannedTowns.end()) {
+			allowedTowns.push_back(entity->getId());
+		}
+	});
```
IvanSavenko, line 389:
```suggestion
	for (auto const & factionID : VLC->townh->getDefaultAllowed())
		if(std::find(pack.bannedTowns.begin(), pack.bannedTowns.end(), factionID) == pack.bannedTowns.end()) {
			allowedTowns.push_back(factionID);
		}
	});
```

server/processors/PlayerMessageProcessor.cpp
Laserlicht: And change the messaging logic?

IvanSavenko: Why not?
Existing messages with player input can be added as raw strings to MetaString.
Better than these `~~` hacks

client/GameChatHandler.cpp
Laserlicht: Done :)

PR #3867, opened 2024-04-30
Simple implementation of heartbeat package

IvanSavenko:
Not yet sure about this change.

Open sockets will now send empty 'heartbeat' packet to show that connection is alive every 10 seconds. This should fix bug where due to vcmiserver crash lobby will still show crashed room for a rather long period. 

However since in vcmi network thread may wait for long periods (blocking dialogs on client, RMG generation on server) proper heartbeat tracking with timestamps of last received message is not possible - it is completely legal for network to be idle for long periods of time if player has blocking dialog open.

Before this PR, lobby used TCP keepalive signal to discover dead connections. However it is too slow (2+ hours). Version from this PR should work better, since interval between packets is now 10 seconds (and can be changed by us at any point), but it still relies on TCP mechanics - even if network thread on client is waiting for blocking dialog, OS would still notify server that TCP connection is alive and empty heartbeat package has been delivered. And if client actually crashed - then server will close connection by timeout.

### DIFF
--- a/lib/network/NetworkConnection.cpp
+++ b/lib/network/NetworkConnection.cpp
@@ -12,12 +12,12 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
-NetworkConnection::NetworkConnection(INetworkConnectionListener & listener, const std::shared_ptr<NetworkSocket> & socket)
+NetworkConnection::NetworkConnection(INetworkConnectionListener & listener, const std::shared_ptr<NetworkSocket> & socket, const std::shared_ptr<NetworkContext> & context)
 	: socket(socket)
+	, context(context)
 	, listener(listener)
 {
 	socket->set_option(boost::asio::ip::tcp::no_delay(true));
-	socket->set_option(boost::asio::socket_base::keep_alive(true));
 
 	// iOS throws exception on attempt to set buffer size
 	constexpr auto bufferSize = 4 * 1024 * 1024;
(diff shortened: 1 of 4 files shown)


dydzio0614:
What do you think about also handling case when player is trying to remake room while his previous room is still hanging? That would be additional fail-safe logic to allow creating new room that is visible on list regardless of this logic

IvanSavenko:
Need to think about it.
at the moment it is vcmiserver who notifies lobby that player left room (by sending message) or that room has been closed (by closing connection).

Need to reproduce it in some artificial way and see what's happening on both client & lobby.

kambala-decapitator:
package? you mean _packet_?

dydzio0614:
@IvanSavenko starting lobby game solo might do the trick, unsure though

After hanging room is already there, both host and joiner dont see new room on list after its created so it may be possible to check from host's perspective

Laserlicht:
@IvanSavenko If I open Lobby on android and login. I will kicked after 10 seconds out of the lobby.

IvanSavenko:
are you using new build made after merge of this PR?

Laserlicht:
Ah. No... Build 20fc2b6

IvanSavenko:
Yeah, had to make such breaking change. Should be okay-ish since 1.5 is not out, so breaking protocol is not that dangerous yet.

Laserlicht:
Ah. No problem. Tested again with actual build. Everything fine. :)

PR #3870, opened 2024-05-01
Launcher translations update

IvanSavenko:
- Fixed not applied translation for some strings in Launcher
- Added missing strings to translations (mostly mod management error messages)
- Reworded data import strings to account for innoextract
- Removed mentions of vcmibuilder tool since it is now almost completely replaced by innoextract
- Regenerated Qt translation files for Launcher
- Updated Ukrainian translation

### DIFF
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -97,12 +97,6 @@ void FirstLaunchView::on_pushButtonDataCopy_clicked()
 	copyHeroesData();
 }
 
-void FirstLaunchView::on_pushButtonDataHelp_clicked()
-{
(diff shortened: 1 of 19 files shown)


Laserlicht:
Maybe also shorten some of the long strings in launcher? (preparation for QT-launcher on android/optimization on iOS)

IvanSavenko:
Not this time probably.

Not yet sure what the final text size would be. I can check my device, but not every Android out there.
Not yet sure what we need to shorten and whether I can come up with short but clear version
And not sure if shortening is best approach. We can also switch to start using scrollable area similar to one in settings tab and keep using longer texts.

Yes, I'll probably take a look on Android UX once porting is over, including smaller tweaks like making sure that text makes sense for mobiles, but only after Android launcher is ready.

PR #3888, opened 2024-05-05
Deprecate fightValue

DjWarmonger:


dydzio0614:
AiValue is used by VCAI IIRC

DjWarmonger:
> AiValue is used by VCAI IIRC

Oh right. Still, both AI value and VCAI should be deprecated eventually.

PR #3895, opened 2024-05-06
CWindowWithArtifacts refactoring part2

SoundSSGood:
- [X] CWindowWithArtifacts - refactoring. Reducing code complexity.
- [X] ctrl+click; alt+click now works in the backpack window.
- [x] ctrl+click now works in the artifact altar window.
- [X] ArtifactPosition::TRANSITION_POS is now simple ArtSlotInfo. Not a vector.
- ~~Shift+click on the artifact slot should open the quick backpack window.~~

### DIFF
--- a/client/widgets/CArtifactsOfHeroMarket.cpp
+++ b/client/widgets/CArtifactsOfHeroMarket.cpp
@@ -26,7 +26,7 @@ CArtifactsOfHeroMarket::CArtifactsOfHeroMarket(const Point & position, const int
 		artPlace->setSelectionWidth(selectionWidth);
 };
 
-void CArtifactsOfHeroMarket::onClickPrassedArtPlace(CArtPlace & artPlace)
+void CArtifactsOfHeroMarket::onClickPressedArtPlace(CArtPlace & artPlace)
 {
 	if(const auto art = getArt(artPlace.slot))
 	{
--- a/client/widgets/CArtifactsOfHeroMarket.cpp
+++ b/client/widgets/CArtifactsOfHeroMarket.cpp
@@ -25,3 +25,21 @@ CArtifactsOfHeroMarket::CArtifactsOfHeroMarket(const Point & position, const int
 	for(auto artPlace : backpack)
 		artPlace->setSelectionWidth(selectionWidth);
 };
+
+void CArtifactsOfHeroMarket::onClickPrassedArtPlace(CArtPlace & artPlace)
+{
+	if(const auto art = getArt(artPlace.slot))
+	{
+		if(onSelectArtCallback && art->artType->isTradable())
+		{
+			unmarkSlots();
+			artPlace.selectSlot(true);
+			onSelectArtCallback(&artPlace);
+		}
+		else
+		{
+			if(onClickNotTradableCallback)
+				onClickNotTradableCallback();
+		}
+	}
+}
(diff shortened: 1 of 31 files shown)


dydzio0614:
why was shift + click idea abandoned? it would be good feature in addition to having this at mouse wheel

SoundSSGood:
> why was shift + click idea abandoned? it would be good feature in addition to having this at mouse wheel

There is no way to close the window without swapping artifact.

SoundSSGood:
Maybe yes. I'll deal with this a little later

client/widgets/CArtifactsOfHeroMarket.cpp
```diff
@@ -25,3 +25,21 @@ CArtifactsOfHeroMarket::CArtifactsOfHeroMarket(const Point & position, const int
 	for(auto artPlace : backpack)
 		artPlace->setSelectionWidth(selectionWidth);
 };
+
+void CArtifactsOfHeroMarket::onClickPrassedArtPlace(CArtPlace & artPlace)
```
IvanSavenko, line 29:
```suggestion
void CArtifactsOfHeroMarket::onClickPressedArtPlace(CArtPlace & artPlace)
```

client/widgets/CArtifactsOfHeroMarket.h
```diff
@@ -14,7 +14,9 @@
 class CArtifactsOfHeroMarket : public CArtifactsOfHeroBase
 {
 public:
-	std::function<void(const CArtPlace*)> selectArtCallback;
+	std::function<void(const CArtPlace*)> onSelectArtCallback;
+	std::function<void()> onClickNotTradableCallback;
 
 	CArtifactsOfHeroMarket(const Point & position, const int selectionWidth);
+	void onClickPrassedArtPlace(CArtPlace & artPlace);
```
IvanSavenko, line 21:
```suggestion
	void onClickPressedArtPlace(CArtPlace & artPlace);
```

client/windows/CWindowWithArtifacts.h
```diff
@@ -19,41 +19,44 @@
...
-		std::weak_ptr<CArtifactsOfHeroMain>,
-		std::weak_ptr<CArtifactsOfHeroBackpack>,
-		std::weak_ptr<CArtifactsOfHeroQuickBackpack>>;
+	using ArtifactsOfHeroVar = std::variant<
+		std::shared_ptr<CArtifactsOfHeroMarket>,
+		std::shared_ptr<CArtifactsOfHeroAltar>,
+		std::shared_ptr<CArtifactsOfHeroKingdom>,
+		std::shared_ptr<CArtifactsOfHeroMain>,
+		std::shared_ptr<CArtifactsOfHeroBackpack>,
+		std::shared_ptr<CArtifactsOfHeroQuickBackpack>>;
```
IvanSavenko, line 28: I know that this is not a new code, but why you opted to use variant here instead of using interface with different implementations with virtual methods?

IvanSavenko (approved):

client/windows/CWindowWithArtifacts.h
SoundSSGood: At a certain stage of working with the code, it was easier to distinguish which window we were working with at the moment. But for now, you may be right. Switching to virtual methods may make the code even simpler.

SoundSSGood: I'll try to redo it

PR #3909, opened 2024-05-08
Disable 'catch everything' block in RMG in favor of clean crash

IvanSavenko:
Disable try-catch block in RMG that hides any exceptions, leading to weird crashes down the line, for example on attempt to access not placed starting heroes, not existing starting town, etc.

Alternatively, we could try to regenerate maps until RMG succeed, but don't want to hide potential bugs or create infinite loop by accident.

### DIFF
--- a/lib/rmg/CMapGenerator.cpp
+++ b/lib/rmg/CMapGenerator.cpp
@@ -130,6 +130,7 @@ std::unique_ptr<CMap> CMapGenerator::generate()
 	catch (rmgException &e)
 	{
 		logGlobal->error("Random map generation received exception: %s", e.what());
+		throw;
 	}
 	Load::Progress::finish();
 	return std::move(map->mapInstance);


DjWarmonger:
What actually does this code "hide"? It only catches well-defined RMG exceptions and logs them.

And yes, "failed to generate random map" message should be handled, but it turned out to be quite complicated.

IvanSavenko:
> What actually does this code "hide"? 

Game crashes on attempt to set active object (hero or town) in client.
Reason: RMG failed to generate map and current player has 0 towns and heroes

Game crashes somewhere inside NewTurn applier in server
Reason: RMG failed to generate map

Debugger does not point you to actual crash location, when you get crash log from player you need to search through logs to find actual crash reason, call stacks that will be present in crash dumps won't point to crash reason

> And yes, "failed to generate random map" message should be handled, but it turned out to be quite complicated.

But at the moment it is not handled.

IvanSavenko:
Changed approach - re-throw received exception.
This would log exception as before and then - immediately crash since there are no try .. catch blocks in callers.
So log would still contain useful info, and any RMG failure would still result in instant crash instead of weird behavior later.

Can be removed if / when we get clean handling of RMG failure that won't cause issues later during map loading.

PR #3924, opened 2024-05-10
Route roads away from zone borders

DjWarmonger:


dydzio0614:
I would change target branch to beta

Alucard648:
I think, this should be optional in template properties.

DjWarmonger:
Guys complained about roads on stream, so everybody thinks it should be improved.

PR #3929, opened 2024-05-10
[1.5.1] Stabilization

IvanSavenko:
Fixes for crashes reported on Google Play:

- Fixes crash on attempt to load 1.4 save with removed quests (e.g. Quest Guards / Border Guards)
- Attempt to fix weird crash where player attempts to start the game but there is already ongoing game? Removed some code that should be unused, but not sure whether this would actually fix the crash.


IvanSavenko:
One more fix added;
- Use async_resolve() instead of (synchronous) resolve(). This serves two purposes:
- - remove possible freeze if resolving is not instant
- - fixes crash on resolving failure - we actually can't report errors to player from synchronous calls - due to already locked UI mutex

Will merge these changes directly into master in order to deploy them on Google Play. That crash on loading saves is a bit too common.

PR #3931, opened 2024-05-10
[1.5.1] Retaliation preview

### LINKED ISSUE #1760: Add retaliation information to attack tooltip in battle
**Describe your proposal**
In addition to damage & kills, we could show information about potential counter-attack in attack tooltip in battle:
- Whether creature is capable of retaliating (e.g. Will retaliate / No retaliations)
- Potential damage dealt to our creature by retaliation
- Potential kills dealt to our creature by retaliation

However this might not be trivial and should account for multiple potential cases:
- ranged units generally don't retaliate, however mods might use RANGED_RETALIATION bonus, allowing retaliation in ranged attacks
- blinded or othervice disabled unit will not be able to retaliate even if he has not spent his retaliations
- if our creature has "no retaliation" bonus, or has a way to guarantee disabling target (e.g. 100% blind)
- if entire enemy unit is destroyed by our attack, it won't be able to retaliate - we might display message like "May retaliate", to account for potential bad luck roll.

**Additional context**
Split from #857

IvanSavenko:
Show retaliation preview when hovering over units that can be attacked.

Does not accounts for luck / abilities for both sides.
If enemy will be killed or spent all retaliations text would change to "Will not retaliate".
In case if enemy may be killed (and retaliation might not happen) text would change to "May retaliate" instead of "Will retaliate"

- Fixes #1760 

зображення

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -252,6 +252,13 @@
 	"vcmi.battleWindow.damageEstimation.damage.1" : "%d damage",
 	"vcmi.battleWindow.damageEstimation.kills" : "%d will perish",
 	"vcmi.battleWindow.damageEstimation.kills.1" : "%d will perish",
+	
+	"vcmi.battleWindow.damageRetaliation.will" : "Will retaliate ",
+	"vcmi.battleWindow.damageRetaliation.may" : "May retaliate ",
+	"vcmi.battleWindow.damageRetaliation.never" : "Will not retaliate.",
+	"vcmi.battleWindow.damageRetaliation.damage" : "(%DAMAGE).",
+	"vcmi.battleWindow.damageRetaliation.damageKills" : "(%DAMAGE, %KILLS).",
+	
 	"vcmi.battleWindow.killed" : "Killed",
 	"vcmi.battleWindow.accurateShot.resultDescription.0" : "%d %s were killed by accurate shots!",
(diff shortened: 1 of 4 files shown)


IvanSavenko:
Side effect - this PR fixes computation of retaliation damage that is also used by AI.

This means that AI now would prefer attacking units that can not retaliate (e.g. spent all their retaliations). 

There also was bug that resulted in incorrect estimation of retaliation (only minimal damage was accounted for), but this one probably won't affect AI significantly

DjWarmonger:
Honestly I think that showing retaliation damage is an overkill, so far it's been a challenge to estimate the damage done by retaliating creature. Shouldn't this text be optional?

Nice job fixing AI, BTW.

IvanSavenko:
We actually have retaliation estimation already. And it should work quite well if there are no randomly triggered abilities, e.g. luck, deadly strike, etc.

- game computes damage range that would be dealt by attacker
- after that, game simulates how much damage would the enemy deal to attacker if he receives
- - minimal possible damage in attack
- - maximal possible damage in attack

In my tests this seems to be work reasonable well. Ranged retaliation is not supported (and does not happens in H3) so only used for melee attacks.

Can add option for that, but we already provide only custom version of damage tooltip - "X units would perish" was not present in H3 and we have no option for that either.

dydzio0614:
I would not make it optional - creature queue also does not exist in H3 :P

Issue #3923, opened 2024-05-10
v1.5.0 release tarball lacks innoextract


Gigas002:
[These files](https://github.com/vcmi/vcmi/tree/develop/launcher/lib) are lacking from 1.5.0 release tarball

kambala-decapitator:
I believe it's a GitHub issue that it doesn't pack submodules

you can build with `-D ENABLE_INNOEXTRACT=OFF`

IvanSavenko:
Unfortunately Github tarball don't contain submodules for some reason.
Same goes to our other submodules - fuzzylite & google test.

You can either:
- use ENABLE_INNOEXTRACT=OFF flag, as suggested by kambala
- get our innoextract sources from here: https://github.com/vcmi/innoextract (or I can create separate 'release' for innoextract that would match what was used for 1.5)
- use `git clone --recursive` instead of tarball. This would also grab submodules

Gigas002:
I see, no problem, will disable innoextract for a time being! :+1:

kambala-decapitator:
but you won't be able to build AIs because they depend on fuzzylite

better clone the repo and checkout tag, then you can populate submodules

Gigas002:
Yeah, I've been using `-DFORCE_BUNDLED_FL=OFF` to not trigger problems with fuzzylite, I've only encountered the issue with innoextract on this release and thought it'd be good to report this, but that seems like to be an already known limitation of github

PR #3933, opened 2024-05-11
Implement github action to produce source code archive with submodules for release

Gigas002:
After giving some thoughts at #3923, I realized that shouldn't be too hard and won't add much burden to support. Please, feel free to close the PR if you think this doesn't fit the project well.
I also had to add the `unzip -q release.zip -d ./src` part, because github upload action always wraps result in a `.zip` archive, and that would produce archive in archive.. ugh.

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -387,25 +387,31 @@ jobs:
 
 
   deploy-src:
+    if: always() && github.ref == 'refs/heads/master'
     runs-on: ubuntu-latest
     defaults:
       run:
-        shell: pwsh
+        shell: bash
     steps:
         - uses: actions/checkout@v4
           with:
             submodules: recursive
+
+        - name: Build Number
+          run: |
+            source '${{github.workspace}}/CI/get_package_name.sh'
+            echo VCMI_PACKAGE_FILE_NAME="$VCMI_PACKAGE_FILE_NAME" >> $GITHUB_ENV
             
         - name: Create source code archive (including submodules)
           run: |
-            git archive HEAD -o "release.zip" --worktree-attributes -v
+            git archive HEAD -o "release.tar" --worktree-attributes -v
             git submodule update --init --recursive
-            git submodule --quiet foreach 'cd "$toplevel"; zip -ru "release.zip" "$sm_path"'
-            unzip -q release.zip -d ./src
+            git submodule --quiet foreach 'cd "$toplevel"; tar -rvf "release.tar" "$sm_path"'
+            gzip release.tar
             
         - name: Upload source code archive
           uses: actions/upload-artifact@v4
           with:
-            name: src
+            name: ${{ env.VCMI_PACKAGE_FILE_NAME }}
             path: |
-              ./src
+              ./release.tar.gz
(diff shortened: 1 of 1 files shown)


.github/workflows/github.yml
```diff
@@ -384,3 +384,28 @@ jobs:
         name: ${{ env.VCMI_PACKAGE_FILE_NAME }}
         path: |
           ${{ env.ANDROID_APK_PATH }}
+
+
+  deploy-src:
```
kambala-decapitator, line 389: should be run only when making release, so something like

    if: always() && github.ref == 'refs/heads/master'

.github/workflows/github.yml
```diff
@@ -384,3 +384,28 @@ jobs:
...
+          run: |
+            git archive HEAD -o "release.zip" --worktree-attributes -v
+            git submodule update --init --recursive
+            git submodule --quiet foreach 'cd "$toplevel"; zip -ru "release.zip" "$sm_path"'
+            unzip -q release.zip -d ./src
+            
+        - name: Upload source code archive
+          uses: actions/upload-artifact@v4
+          with:
+            name: src
```
IvanSavenko, line 409: Can we use same naming scheme as other artifacts?, e.g.:
`9040593311-vcmi-PR-3933-844da66 - ios.zip`
Currently sources are named simply `src.zip`

.github/workflows/github.yml
```diff
@@ -384,3 +384,28 @@ jobs:
...
+    steps:
+        - uses: actions/checkout@v4
+          with:
+            submodules: recursive
+            
+        - name: Create source code archive (including submodules)
+          run: |
+            git archive HEAD -o "release.zip" --worktree-attributes -v
+            git submodule update --init --recursive
+            git submodule --quiet foreach 'cd "$toplevel"; zip -ru "release.zip" "$sm_path"'
```
IvanSavenko, line 403: Perhaps use tar.gz instead of zip?

.github/workflows/github.yml
```diff
@@ -384,3 +384,28 @@ jobs:
...
+            unzip -q release.zip -d ./src
```
IvanSavenko, line 404: Probably not needed?
Yes, archive in archive is weird, but first step when preparing release is "unpack all artifacts", so having uniform format is more convenient.

.github/workflows/github.yml
```diff
@@ -384,3 +384,28 @@ jobs:
         name: ${{ env.VCMI_PACKAGE_FILE_NAME }}
         path: |
           ${{ env.ANDROID_APK_PATH }}
+
+
+  deploy-src:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: pwsh
```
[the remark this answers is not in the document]
Gigas002: Sorry, forgot to remove this. Fixed

.github/workflows/github.yml
Gigas002: Should be fixed now

.github/workflows/github.yml
Gigas002: I've changed the naming scheme to `${{ env.VCMI_PACKAGE_FILE_NAME }}`, would this be ok?

.github/workflows/github.yml
Gigas002: Changed this behavior

.github/workflows/github.yml
Gigas002: Ah, didn't see that part. I think it's good not to unzip it here then :+1:

IvanSavenko: It is something I do locally on my system, so yeah, you won't find it in our CI scripts.

PR #3959, opened 2024-05-12
Better handling of Ctrl / Cmd modifier keys on Apple systems

IvanSavenko:
- Test only for Cmd modifier key when processing mouse clicks
- Test both Ctrl and Cmd modifier key when processing keyboard input
- This logic is now used for all Apple systems - macOS and iOS

### DIFF
--- a/client/eventsSDL/InputSourceKeyboard.cpp
+++ b/client/eventsSDL/InputSourceKeyboard.cpp
@@ -120,10 +120,20 @@ void InputSourceKeyboard::handleEventKeyUp(const SDL_KeyboardEvent & key)
 	GH.events().dispatchShortcutReleased(shortcutsVector);
 }
 
+bool InputSourceKeyboard::isKeyboardCmdDown() const
+{
+#ifdef VCMI_APPLE
+	return SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_LGUI] || SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_RGUI];
+#else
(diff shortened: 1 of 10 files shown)


client/eventsSDL/InputSourceKeyboard.cpp
```diff
@@ -120,15 +120,25 @@ void InputSourceKeyboard::handleEventKeyUp(const SDL_KeyboardEvent & key)
...
 #endif
 }
 
+bool InputSourceKeyboard::isKeyboardCtrlDown() const
+{
+#ifdef VCMI_APPLE
+	return SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_LCTRL] || SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_RCTRL] ||
+		   SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_LGUI] || SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_RGUI];
+#else
+	return SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_LCTRL] || SDL_GetKeyboardState(nullptr)[SDL_SCANCODE_RCTRL];
```
kambala-decapitator, line 138: You could extract it to a variable to reduce copy-paste

kambala-decapitator (approved):

client/eventsSDL/InputSourceKeyboard.cpp
IvanSavenko: I guess I could, but not sure if this improves readability in this case, with all these tiny methods and ifdef's

kambala-decapitator: At least you'd immediately see that both code paths handle normal Ctrl and apple just does something extra

PR #3970, opened 2024-05-13
ExtraOptions improvements

Laserlicht:
- allow to change in loadgame (not readonly anymore)
- default for multiplayer is cheat off
- saves last setting (multiplayer & singleplayer seperate)
- disable buttons for guests

### DIFF
--- a/client/lobby/CLobbyScreen.cpp
+++ b/client/lobby/CLobbyScreen.cpp
@@ -230,6 +230,16 @@ void CLobbyScreen::toggleChat()
 
 void CLobbyScreen::updateAfterStateChange()
 {
+	if(CSH->isHost() && screenType == ESelectionScreen::newGame)
+	{
+		bool isMultiplayer = CSH->loadMode == ELoadMode::MULTI;
+		ExtraOptionsInfo info = SEL->getStartInfo()->extraOptionsInfo;
+		info.cheatsAllowed = isMultiplayer ? persistentStorage["startExtraOptions"]["multiPlayer"]["cheatsAllowed"].Bool() : !persistentStorage["startExtraOptions"]["singlePlayer"]["cheatsNotAllowed"].Bool();
+		info.unlimitedReplay = persistentStorage["startExtraOptions"][isMultiplayer ? "multiPlayer" : "singlePlayer"]["unlimitedReplay"].Bool();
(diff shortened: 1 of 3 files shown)


AI/Nullkiller/AIGateway.cpp
```diff
@@ -828,7 +828,10 @@ void AIGateway::makeTurn()
 	boost::shared_lock<boost::shared_mutex> gsLock(CGameState::mutex);
 	setThreadName("AIGateway::makeTurn");
 
-	cb->sendMessage("vcmieagles");
+	if(cb->getStartInfo()->extraOptionsInfo.cheatsAllowed)
+		cb->sendMessage("vcmieagles");
+	else
+		cb->sendMessage("Nullkiller AI currently requires the ability to cheat in order to function correctly! Please enable!");
```
IvanSavenko, line 834: Will this result in multiple messages from each AI on every turn?

Mods/vcmi/config/vcmi/english.json
```diff
@@ -63,7 +63,11 @@
 	"vcmi.mainMenu.serverClosing" : "Closing...",
 	"vcmi.mainMenu.hostTCP" : "Host TCP/IP game",
 	"vcmi.mainMenu.joinTCP" : "Join TCP/IP game",
-	
+
+	"vcmi.multiplayerWelcomeMessage.additionalSettings" : "Additional settings",
+	"vcmi.multiplayerWelcomeMessage.cheatsAllowed" : "Cheats allowed",
+	"vcmi.multiplayerWelcomeMessage.unlimitedReplays" : "Unlimited battle replays",
```
IvanSavenko, line 69: Why not reuse existing strings?
- vcmi.optionsTab.extraOptions.hover
- vcmi.optionsTab.cheatAllowed.hover
- vcmi.optionsTab.unlimitedReplay.hover

AI/Nullkiller/AIGateway.cpp
Laserlicht: One message per turn per AI player. Yes. but it's still better than cryptic "vcmieagles" message (which also appears in the same interval).

And I think it's better, than completly silent it (as the AI may not work as intended).

IvanSavenko: Why not display it once per game?

PR #3988, opened 2024-05-16
[1.5.2] Use Ubuntu 24.04 runner for some presets

IvanSavenko:
Changes non-crosscompiled Ubuntu runners to use Ubuntu-24.04 preset with gcc-14. These runners don't produce any artifacts so as long as run succeeds there should be no unexpected side effects.

- mingw runners use 22.04 preset as before. Can be changed later, e.g. for 1.6. May require rebuilding package & testing, so will keep 22.04 for now
- android runners use 22.04 preset as before - since OS was changed in Android branch.

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -20,11 +20,11 @@ jobs:
       matrix:
         include:
           - platform: linux-qt6
-            os: ubuntu-22.04
+            os: ubuntu-24.04
             test: 0
             preset: linux-clang-test
           - platform: linux
-            os: ubuntu-22.04
+            os: ubuntu-24.04
             test: 1
             preset: linux-gcc-test
           - platform: linux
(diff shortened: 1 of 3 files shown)


.github/workflows/github.yml
```diff
@@ -201,8 +201,8 @@ jobs:
 
     - name: Configure
       run: |
-        if [[ ${{matrix.preset}} == linux-gcc-test ]]; then GCC12=1; fi
-        cmake -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }} ${GCC12:+-DCMAKE_C_COMPILER=gcc-12 -DCMAKE_CXX_COMPILER=g++-12}
+        if [[ ${{matrix.preset}} == linux-gcc-test ]]; then GCC14=1; fi
+        cmake -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }} ${GCC12:+-DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14}
```
Alexander-Wilms, line 205:
```suggestion
        cmake -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }} ${GCC14:+-DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14}
```

PR #3991, opened 2024-05-16
preselect last save in loading screen

### LINKED ISSUE #3987: pre-select the latest save in Load Game screen
Would be great to pre-select the latest save instead of the first one as quite frequently you simply want to continue from where you stopped

Laserlicht:
fixes #3987

### DIFF
--- a/client/lobby/SelectionTab.cpp
+++ b/client/lobby/SelectionTab.cpp
@@ -677,6 +677,9 @@ void SelectionTab::selectFileName(std::string fname)
 		}
 	}
 
+	filter(-1);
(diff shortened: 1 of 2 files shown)


client/lobby/SelectionTab.cpp
```diff
@@ -677,6 +677,9 @@ void SelectionTab::selectFileName(std::string fname)
 		}
 	}
 
+	filter(-1);
+	selectAbs(-1);
+
```
IvanSavenko, line 682: @Laserlicht Why was this change necessary?

PR #3998, opened 2024-05-17
[1.6] Reorganize SDL sound/music/video handling

IvanSavenko:
Our video playback has multiple issues and ancient code that leads to various bugs and one of the remaining places that requires direct access to SDL functions from UI system.

This PR reworks our video player and reorganizes sound/music players (without any major changes in these two).
Note that I kept "convert to wav and play audio as sound" approach we use for playback of audio from video files. This is unfortunately the only clear way to play audio that I found - SDL_Mixer does not provide any functionality for streaming or other forms of 'sideloading' audio into the system other than loading it from some form of file.

General changes:
- Media handlers were moved from root client directory to `media/`
- Reorganized massive files into smaller, 1 file per class form
- Extracted interface from implementation, CCS now only has pointers to base interfaces - ISoundPlayer / IMusicPlayer / IVideoPlayer

Video player changes:
- Now all windows use new VideoWidget class instead of direct access to videoHandler to hide all implementation details and to be in line with rest of our UI system.
- Video player would now prefer high-quality .bik video instead of low-quality .smk, if such video exists
- All audio formats should now be supported by our video player (provided that support is present in ffmpeg)
- Video player now uses RGB surface instead of YUV for some videos that use RGB palette to avoid artifacts from double-conversion
- Unified audio and video handling in video player to reduce code duplication
- It is now (technically) possible to play multiple video files at once, if such functionality is needed
- If popup window is opened on top of video playback, game will no longer replace video with black square
- after popup window has been closed video will now resume playback instead of starting from scratch


client/media/CVideoHandler.cpp
```diff
@@ -0,0 +1,611 @@
...
+		case AV_SAMPLE_FMT_FLTP: return { 4, 3, true};
+		case AV_SAMPLE_FMT_DBL:  return { 8, 3, false};
+		case AV_SAMPLE_FMT_DBLP: return { 8, 3, true};
+	}
+	throw std::runtime_error("Invalid audio format");
+}
+
+int FFMpegStream::findAudioStream() const
+{
+	for(int i = 0; i < formatContext->nb_streams; i++)
```
Laserlicht, line 422: As comment: As far as I know there is one special case. For the intro of the chronicles there are multiple audio streams. One per language. Even the english version contains the german audio stream for the intro.

IvanSavenko: Noted, thanks for info.

Which Chronicle should I look up? Or any will do?

IvanSavenko: Found it, 1st chronicle, Warlords of the Wasteland, Intro.bik and smk
5 different audio stream.

Unfortunately no information on language of each of these streams, will have to figure it out & hardcode it

IvanSavenko: So in total we have:
HC1, HC3: english, french, german, italian, spanish
HC2, HC4, HC5, HC6: english
HC7, HC8: english, german

Laserlicht: Order should be:
```
english
french
german
italian
spanish
```

Used: https://translatedlabs.com/spoken-language-identifier

Laserlicht: Ah, you already figured it out. GitHub has not shown your last post to me. :(

IvanSavenko: Still good to have confirmation from 2nd source :)

Didn't knew about such service so I transcribed English audio and then tried to compare auto-translated text to what I hear on other streams.

Laserlicht: I've checked it again. I think you mistaken HC3 and HC2?

HC1, HC2: english, french, german, italian, spanish
CD-Version (tested with german cd): HC3, HC4: english, french, german, italian, spanish
GOG-Version: HC3, HC4: english
HC5, HC6: english
HC7, HC8: english, german

IvanSavenko:
Yeah, HC2 has 5 streams, HC3 only English (gog version)

Wasn't aware of CD version. 
So, 5-stream version is always english/french/german/italian/spanish, 2-stream version is always english+german.

Laserlicht:
> Wasn't aware of CD version. So, 5-stream version is always english/french/german/italian/spanish, 2-stream version is always english+german.

Yeah, was skeptical because as far as I remember the only HC without german was the digital distributed (HC5+6).

PR #3999, opened 2024-05-17
update Chinese translation

kdmcser:


dydzio0614:
can you change target to beta branch instead of develop?

kdmcser:
ok

PR #4004, opened 2024-05-18
[1.5.2] background for multiple windows

Laserlicht:
Draw standard background for:
- campaign end video
- highscore input
- highscore screen

Looks more clean as these screens aren't part of the current game anymore.


client/mainmenu/CHighScoreScreen.cpp
```diff
@@ -289,6 +291,9 @@ int CHighScoreInputScreen::addEntry(std::string text) {
 
 void CHighScoreInputScreen::show(Canvas & to)
 {
+	if(background)
+		background->redraw();
```
IvanSavenko, line 295: Please don't use 'redraw' method  inside show/showAll calls.
showAll(to) is more suitable when simple show(to) is not enough

(same goes to other such cases)

PR #4009, opened 2024-05-19
NKAI: port exploration from VCAI

nullkiller:
Ported exploration logic from VCAI
Now NKAI can fallback to exploration if not allowed to cheat or allied humans.


AI/Nullkiller/AIGateway.cpp
```diff
@@ -829,13 +829,9 @@ void AIGateway::makeTurn()
...
 	setThreadName("AIGateway::makeTurn");
 
-	if(cb->getStartInfo()->extraOptionsInfo.cheatsAllowed)
-		cb->sendMessage("vcmieagles");
-	else
+	if(nullkiller->isOpenMap())
 	{
-		if(!announcedCheatingProblem)
-			cb->sendMessage("Nullkiller AI currently requires the ability to cheat in order to function correctly! Please enable!");
-		announcedCheatingProblem = true;
```
Laserlicht, line 838: Can be also removed from header.

PR #4018, opened 2024-05-19
[1.5.2?] Multiplayer voting

IvanSavenko:
This is just a quick draft, not yet sure whether this feature is actually needed, in which form should it be and what else could it be used for.

At the moment it allows any player to initiate voting and once all players agree to it this request will be applied:
- change duration of simturns (including restarting them if simturns ended)
- aborting simturns (primary intended for cases when players set too long period for non-breaking simturns)
- prolong base timer by specified duration

Initiation of voting is done via `!vote` command, e.g. `!vote simturns abort`
Once voting has been initiated, all players other than initiator can vote  -`!vote yes` or `!vote no`
All player must agree to change before it applied. Can also be used in single-player games or in hotseat if needed for some reason.


dydzio0614:
info about usage should be added to docs

PR #4019, opened 2024-05-19
[1.5.2] Simturn tweaks

IvanSavenko:
Two minor changes to simultaneous turns that were proposed on Discord:
- now contact between allied players will not break simturns.
- Having hero in range of object owned by another players will now be registered as contact

No UI for 1st option to avoid breaking MP between different versions of 1.5


dydzio0614:
what if both allies want to capture same building?

IvanSavenko:
> what if both allies want to capture same building?

Same as if forced simturns are active - first to come takes it.

PR #4020, opened 2024-05-19
[1.6] Quickspell support

### LINKED ISSUE #3252: Spell shortcut
**Describe your proposal**
Would make it easier to cast the same spell. Side bar should be possible to disable. Like in SOD_SP.

Preset should be editable by spellbook (press 0-9) or sidebar (maybe long press; short press should be select this spell -> touch compatibility)

**Additional context**
image

Laserlicht:
- adds possibility to use 10 slots for quickspell
- spells can be accessed over button in pop up or keyboard with number
- spell defines are stored on local pc and shared between all heros
- changing defines with right click (long tap) -> all spells are shown
- accessible via hovering (swipe up) over spellbook icon
- not usable spells are grayed out

Fixes  #3252

To be discussed: On Desktop -> Use hovering or new shortcut to open quickspell menu? Or hovering while pressing CTRL?

https://github.com/vcmi/vcmi/assets/13953785/9d5911f5-d329-4f50-9f13-fee9abc352ba


Laserlicht:
@IvanSavenko Should I target develop?

IvanSavenko:
> @IvanSavenko Should I target develop?

Probably? Don't want to experiment with 1.5 too much at this point.

IvanSavenko:
Perhaps instead of trying to abuse button (and rather hidden UI) we should instead display 10 slots vertically, to the left of the screen? Next to hero/creature info. Or alternatively - on the same side as ours, so left for attacker, right for defender.

Won't be available for 800x600, but that's rather niche use-case IMO which I think is fine for sacrifice if rest of players will be able to find this system without external help (or tutorial)

dydzio0614:
@IvanSavenko it would overlap with "beyond screen" hero/creature info and probably annoy some players on 1024x768 resolution or equivalent scaling etc.

IvanSavenko:
> @IvanSavenko it would overlap with "beyond screen" hero/creature info 

Not necessarily. We can move hero/creature info further to the left.

> and probably annoy some players on 1024x768 resolution or equivalent scaling etc.

And current logic would annoy players on any resolution

Laserlicht:
New approach:
grafik

grafik

Now:
- 19 slots (looks better -> same size as battle)
- slots > 9 are available with Shift (e.g. Shift + 2)
- Quick panel can be toggled with T (no other fitting key left...)
- Option in menu missing (should be added, but no more space... for another PR...)
- Every time on the left side (easier than left/right switch)

IvanSavenko:
- Perhaps instead of showing slot number we should show hotkey? Also since we have 16 slots - perhaps we should bind them to '1'...'8' + 'Shift+1' ... 'Shift+8'?
- Perhaps show spellbook instead of this popup list when editing spells? Locating spell in spellbook is easier since you likely know its exact position, spells are sorted by level and can be filtered by school. It takes me ~5 seconds to find specific spell on your screenshot.

Laserlicht:
Good idea!
Implemented.

Laserlicht:
Ready now.

client/battle/BattleWindow.cpp
```diff
@@ -64,6 +64,40 @@ BattleWindow::BattleWindow(BattleInterface & owner):
...
+		}
+		catch(const IdentifierResolutionException& e)
+		{
+			return;
+		}
+		if(id.hasValue() && owner.getBattle()->battleGetMyHero() && id.toSpell()->canBeCast(owner.getBattle().get(), spells::Mode::HERO, owner.getBattle()->battleGetMyHero()))
+		{
+			owner.castThisSpell(id);
+		}
+	};
```
IvanSavenko, line 82: Probably better to turn it into proper method of BattleWindow, similar to other shortcut handlers

Laserlicht: Done :)

godric3:
Not sure about 16 slots on mobile devices, IMO it can be hard to select correct slot there, given they are next to each other and rather small. Also is showing not only known spells but all of them in spellbook intentional?

Laserlicht:
> Not sure about 16 slots on mobile devices, IMO it can be hard to select correct slot there, given they are next to each other and rather small. Also is showing not only known spells but all of them in spellbook intentional?

Tested on Galaxy S23 (6,1") which is rather small today. For me it's absolutely okay.
We could decrease amount. And increse gaps between. But it doesn't look good I think.

Yeah, showing all spells is intentional. The shortcut's are global (independent from hero). It would be bad if you can set up all your shortcuts only if you have the spell.

IvanSavenko:
Have to agree. Haven't tested this PR on phone, but battle hexes do need some caution / precision on my phone. And these slots seems to be of a similar size.

Perhaps for mobile devices:
- reduce number of slots to 12? 9?
- remove hotkey name overlay - since there is no keyboard. Probably needs some way to detect presence of external keyboard (can do this in separate PR)

What do you think?

Laserlicht:
> Have to agree. Haven't tested this PR on phone, but battle hexes do need some caution / precision on my phone. And these slots seems to be of a similar size.
> 
> Perhaps for mobile devices:
> 
>     * reduce number of slots to 12? 9?
> 
>     * remove hotkey name overlay - since there is no keyboard. Probably needs some way to detect presence of external keyboard (can do this in separate PR)
> 
> 
> What do you think?

For me also okay. I'll test it with 12 (0-9 + two letters) and gap between the buttons (without difference for PC and mobile).

For overlay: Yes of course in a separate PR if we have support for this. But the greatest way would be if we have a on the fly detection of current input. Depending on last used input (touch vs. Keyboard+Mouse vs. Gamepad). New games usually works like this (HUD changes when switching form mouse to gamepad ingame).

Laserlicht:
It's okay, I think:

grafik

IvanSavenko:
Can you merge / rebase to solve conflict so I can merge this one?
(sorry that it took so long)

Laserlicht:
> Can you merge / rebase to solve conflict so I can merge this one? (sorry that it took so long)

Done :)

Issue #4015, opened 2024-05-19
incorrect counting of artefact set pieces


kambala-decapitator:
I have Angelic Alliance assembled and also other parts of AA. When AA is not equipped, then other parts are counted correctly:

![Screenshot 2024-05-19 at 13 12 10](https://github.com/vcmi/vcmi/assets/1557784/ca6a4ac4-fb84-481b-adcf-31b4c122dd0a)

But once I equip AA, then game thinks that other parts are full (although doesn't offer to assemble another AA):

![Screenshot 2024-05-19 at 13 11 59](https://github.com/vcmi/vcmi/assets/1557784/abb2ddd8-bbce-4538-aa5b-9a9a416fc649)

check the first hero:
[q.vsgm1.tgz](https://github.com/vcmi/vcmi/files/15369092/q.vsgm1.tgz)

1.5.1 release

Issue #2582, opened 2023-08-14
When I have some components from a combined artifact, highlight them in  GUI


krs0:
Something like:
![image](https://github.com/vcmi/vcmi/assets/15194321/631cf448-1b13-423e-b1d7-d24c9f83a241)

Or add a chekmark after artfact name ✓

krs0:
ERA has it like this:
![image](https://github.com/vcmi/vcmi/assets/15194321/e0319a01-6960-47a3-a1da-5eda8c28a76c)

PR #4028, opened 2024-05-21
Parts calculation for combined artifacts

SoundSSGood:
https://github.com/vcmi/vcmi/issues/4015 and https://github.com/vcmi/vcmi/issues/2582

image


lib/MetaString.h
```diff
@@ -44,6 +44,7 @@ class DLL_LINKAGE MetaString
 		APPEND_LOCAL_STRING,
 		APPEND_TEXTID_STRING,
 		APPEND_NUMBER,
+		APPEND_EOL,
```
IvanSavenko, line 47: Please move new value to the end of list - this enum can be serialized

PR #4032, opened 2024-05-22
NKAI: speedup exploration a bit

nullkiller:
Improve exploration speed by exploring a few tiles at once


nullkiller:
Sonar cloud complains about code duplication with VCAI which is actually true. No way to share code between them so far.

PR #4037, opened 2024-05-23
Fix end turn button not assigned to shortcut

IvanSavenko:


kambala-decapitator:
in the latest beta build I observe that pressing the button does nothing while E shortcut works => confusing PR title

PR #4038, opened 2024-05-23
[1.5.2] Pathfinder fixes

IvanSavenko:
- If original movement rules are on, it is not possible to attack guards from visitable object directly, only from free tile
- Fixed bug leading that allowed picking up objects while flying on top of water
- Hero can now land when flying from 'guarded' tile to accessible 'guarded' tile even without original movement rules
- Interface will now use same arrow for U-turns in path as H3


lib/pathfinder/CGPathNode.h
```diff
@@ -37,6 +37,7 @@ enum class EPathAccessibility : ui8
 	NOT_SET,
 	ACCESSIBLE, //tile can be entered and passed
 	VISITABLE, //tile can be entered as the last tile in path
+	GUARDED,  //visitable, but in zone of control of nearby monster
```
nullkiller, line 40: GUARDED_VISITABLE?

lib/pathfinder/PathfindingRules.cpp
```diff
@@ -377,7 +379,7 @@ void LayerTransitionRule::process(
 
 	case EPathfindingLayer::SAIL:
 		//tile must be accessible -> exception: unblocked blockvis tiles -> clear but guarded by nearby monster coast
-		if((destination.node->accessible != EPathAccessibility::ACCESSIBLE && (destination.node->accessible != EPathAccessibility::BLOCKVIS || destination.tile->blocked))
+		if((destination.node->accessible != EPathAccessibility::ACCESSIBLE && destination.node->accessible != EPathAccessibility::GUARDED)
 			|| destination.tile->visitable)  //TODO: passableness problem -> town says it's passable (thus accessible) but we obviously can't disembark onto town gate
```
nullkiller, line 383: Interesting comment here.

lib/pathfinder/PathfindingRules.cpp
```diff
@@ -391,16 +393,22 @@ void LayerTransitionRule::process(
...
 				}
-				else
+
+				if (destination.node->accessible == EPathAccessibility::FLYABLE)
+					destination.blocked = true;
+			}
+
+			if(destination.node->accessible == EPathAccessibility::VISITABLE)
+			{
+				if (destination.node->accessible != EPathAccessibility::VISITABLE)
```
nullkiller, line 411: Feels like two exclusive conditions?

lib/pathfinder/PathfindingRules.cpp
```diff
@@ -377,7 +379,7 @@ void LayerTransitionRule::process(
 
 	case EPathfindingLayer::SAIL:
 		//tile must be accessible -> exception: unblocked blockvis tiles -> clear but guarded by nearby monster coast
-		if((destination.node->accessible != EPathAccessibility::ACCESSIBLE && (destination.node->accessible != EPathAccessibility::BLOCKVIS || destination.tile->blocked))
+		if((destination.node->accessible != EPathAccessibility::ACCESSIBLE && destination.node->accessible != EPathAccessibility::GUARDED)
```
nullkiller, line 382: Om pathfinder it should be possible to express it with BLOCKVIS & destination.guarded or am I wrong?

nullkiller: Quite dangerous. Also I do not know how AI will react on it. And feels like we need to cover this with unit tests in order to refactor it any further

IvanSavenko: Double-checked this code.
1) `destination.tile->blocked` check is actually redundant - we can't move onto tiles with obstacles anyway
2) There is already a separate check for visitable objects 1 line below, so any visitable objects (blockvis or not) will be viewed as blocked

So looks correct. 

Also checked all possible cases I could think of - tile guarded, blocked, visitable, etc (including presence of town). All work as intended.

Will clarify comments in code.

lib/pathfinder/PathfindingRules.cpp
IvanSavenko: Forgot to remove this code, will fix.

lib/pathfinder/CGPathNode.h
IvanSavenko: Corrected comment.
GUARDED is used always when there is an active zone of control, whether there is a visitable object or not.

Issue #3957, opened 2024-05-12
amount of resurrected Vampires doesn't appear in battle log


kambala-decapitator:
Should be something like "...and X rise from the dead!"

![Screenshot 2024-05-12 at 13 20 15](https://github.com/vcmi/vcmi/assets/1557784/39545d00-dcbd-41ff-b19c-4ed7452f7fae)

1.5 release

MichalZr6:
Hi everyone!
I'm new to OpenSource contributions, but also a great fun of HOMM3.
I think this small issue can be fixed starting with this part of code in BattleActionProcessor.cpp:
![obraz](https://github.com/vcmi/vcmi/assets/50074545/45741b1b-0f58-49a7-b9c8-0ba2ab465dfc)

Can someone point me to the file that contains those string literals used in for ex. in the battle log ?
They are obtained from "core.genrltxt" but where is it located ?
![obraz](https://github.com/vcmi/vcmi/assets/50074545/c1620dca-24a9-4341-9ced-9ca5da616af6)

IvanSavenko:
Hi. You can look up these strings for example in English translation mod:
https://github.com/vcmi-mods/h3-for-vcmi-englisation/blob/vcmi-1.4/H3forVCMIenglisation/content/config/h3forvcmienglisation/game.json

Alternatively you can export this data using `/translate` command

kambala-decapitator:
please never post code as image, paste as text or give a link to specific line inside a file of repository

MichalZr6:
After digging into the topic, I think I can come up with a solution :) I built the repo yesterday and might do some tests later today. Side note: Cure also doesn't log info about restored HP.

kambala-decapitator:
FYI issue about restored hp: #1007

kambala-decapitator:
fixed by #4040

sanfanling:
I am using 1.7.0, when start the animation of resurrection vampire lord, the whole game is crashes, 100%

kambala-decapitator:
> I am using 1.7.0, when start the animation of resurrection vampire lord, the whole game crashes, 100%

please open new issue about that, it's not related to this one

PR #4040, opened 2024-05-24
Add amount of resurrected Vampires to the battle log

MichalZr6:
Fix issue #3957: amount of resurrected Vampires doesn't appear in battle log 
Fixed plurar / singular text choice for battle log - creatures after life drain action

### DIFF
--- a/lib/battle/Unit.h
+++ b/lib/battle/Unit.h
@@ -41,6 +41,27 @@ namespace BattlePhases
 	};
 }
 
+// Healed HP (also drained life) and resurrected units info
+struct HealInfo
+{
+	HealInfo()
+		: healedHealthPoints(0), resurrectedCount(0)
+	{ }
+	HealInfo(int64_t healedHP, int32_t resurrected)
+		: healedHealthPoints(healedHP), resurrectedCount(resurrected)
+	{ }
+
+	int64_t healedHealthPoints;
+	int32_t resurrectedCount;
+
+	HealInfo& operator +=(const HealInfo& other)
+	{
+		healedHealthPoints += other.healedHealthPoints;
+		resurrectedCount += other.resurrectedCount;
+		return *this;
+	}
+};
+
 class CUnitState;
 
 class DLL_LINKAGE Unit : public IUnitInfo, public spells::Caster, public virtual IBonusBearer, public ACreature
@@ -138,7 +159,7 @@ public:
 	virtual void load(const JsonNode & data) = 0;
 
 	virtual void damage(int64_t & amount) = 0;
-	virtual void heal(int64_t & amount, EHealLevel level, EHealPower power) = 0;
+	virtual HealInfo heal(int64_t & amount, EHealLevel level, EHealPower power) = 0;
 };
 
 class DLL_LINKAGE UnitInfo
--- a/lib/battle/Unit.h
+++ b/lib/battle/Unit.h
@@ -44,17 +44,15 @@ namespace BattlePhases
 // Healed HP (also drained life) and resurrected units info
 struct HealInfo
 {
-	HealInfo()
-		: healedHealthPoints(0), resurrectedCount(0)
-	{ }
+	HealInfo() = default;
 	HealInfo(int64_t healedHP, int32_t resurrected)
 		: healedHealthPoints(healedHP), resurrectedCount(resurrected)
 	{ }
 
-	int64_t healedHealthPoints;
-	int32_t resurrectedCount;
+	int64_t healedHealthPoints = 0;
+	int32_t resurrectedCount = 0;
 
-	HealInfo& operator +=(const HealInfo& other)
+	HealInfo & operator+=(const HealInfo & other)
 	{
 		healedHealthPoints += other.healedHealthPoints;
 		resurrectedCount += other.resurrectedCount;
--- a/lib/spells/effects/Summon.cpp
+++ b/lib/spells/effects/Summon.cpp
@@ -122,6 +122,7 @@ int32_t Summon::summonedCreatureAmount(const Mechanics * m) const
 
 void Summon::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const
 {
+	using battle::HealInfo;
 	BattleUnitsChanged pack;
 	pack.battleID = m->battle()->getBattle()->getBattleID();
 
--- a/lib/spells/effects/Summon.cpp
+++ b/lib/spells/effects/Summon.cpp
@@ -122,7 +122,6 @@ int32_t Summon::summonedCreatureAmount(const Mechanics * m) const
 
 void Summon::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const
 {
-	using battle::HealInfo;
 	BattleUnitsChanged pack;
 	pack.battleID = m->battle()->getBattle()->getBattleID();
 
--- a/server/battles/BattleActionProcessor.cpp
+++ b/server/battles/BattleActionProcessor.cpp
@@ -956,18 +956,18 @@ void BattleActionProcessor::makeAttack(const CBattleInfoCallback & battle, const
 			bat.flags |= BattleAttack::BALLISTA_DOUBLE_DMG;
 	}
 
-	int64_t drainedLife = 0;
+	battle::HealInfo healInfo;
 
 	// only primary target
 	if(defender->alive())
-		drainedLife += applyBattleEffects(battle, bat, attackerState, fireShield, defender, distance, false);
+		applyBattleEffects(battle, bat, attackerState, fireShield, defender, &healInfo, distance, false);
 
 	//multiple-hex normal attack
 	std::set<const CStack*> attackedCreatures = battle.getAttackedCreatures(attacker, targetHex, bat.shot()); //creatures other than primary target
 	for(const CStack * stack : attackedCreatures)
 	{
 		if(stack != defender && stack->alive()) //do not hit same stack twice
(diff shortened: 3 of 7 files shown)


IvanSavenko (approved):

server/battles/BattleActionProcessor.cpp
```diff
@@ -1039,26 +1039,16 @@ void BattleActionProcessor::makeAttack(const CBattleInfoCallback & battle, const
...
 	if(drainedLife > 0)
 	{
-		MetaString text;
-		attackerState->addText(text, EMetaText::GENERAL_TXT, 361);
-		attackerState->addNameReplacement(text, false);
-		text.replaceNumber(drainedLife);
-		defender->addNameReplacement(text, true);
-		blm.lines.push_back(std::move(text));
+		addGenericDrainedLifeLog(blm, attackerState, defender, drainedLife);
+		addGenericResurrectedLog(blm, attackerState, defender, static_cast<int64_t>(drainedLife / attackerState->health.getFirstHPleft()));
```
IvanSavenko, line 1051: Oh, missed this one.
I think it is incorrect?
`getFirstHPleft` is remaining hp of *first* unit. Not max HP of this unit type.
And even if replaced with correct method this still won't be correct.

If Vampire Lord with 30/40 HP drains 20 HP, 1 unit will be resurrected and be left with 10 HP. Even though drained HP was less than 40 (total HP of Vampire Lord).

Hmm.. Perhaps 'drainedLife' should not be int but rather simple struct `{ healthPoint, unitsCount }`?

MichalZr6: You are right... I used wrong method. Will fix.
Yet I left the mechanics unchanged. They heal by the amount of damage dealt resurrecting if its value covers all missing HP of wounded unit in the stack... or is this wrong ?

> LIFE_DRAIN
Affected unit will heal itself, resurrecting any dead units, by amount of dealt damage (Vampire Lord)

https://github.com/vcmi/vcmi/blob/5b7e355953d00bb863240a6d328bdfe9f97bc92f/docs/modders/Bonus/Bonus_Types.md#life_drain

IvanSavenko: Mechanics is correct.
However knowing heal amount is not enough to determine how many units have resurrected.

For example, If first unit in your stack had 35/40 HP before attack then life drain for 20 HP would heal 5 HP to last unit and would resurrect one unit with 15 HP
But if first unit had 10/40 HP then life drain for 20 HP would only heal this unit to 30/40 HP and won't resurrect units.

MichalZr6: Ahhh I know now what you mean and where is the problem. After heal the first unit HP value is useless and the same goes for damageDealt. I was thinking ealier about adding one property to CHealth and a function getSingleHealResurrected() but maybe just counting units before and after heal inside makeAttack is fine ? There is only one more place where getSingleHealResurrected() would be used:
[Heal.cpp#L119](https://github.com/vcmi/vcmi/blob/5b7e355953d00bb863240a6d328bdfe9f97bc92f/lib/spells/effects/Heal.cpp#L119)

IvanSavenko: > counting units before and after heal inside makeAttack is fine ?

If defender has Fire Shield he may be able to kill some of newly resurrected units. So we need to track actual number of resurrected units similar to how we track actual number of healed HP.

Perhaps:
- make applyBattleEffects return void instead of int64_t (which is not exactly logical / expected anyway)
- add structure similar to `FireShieldInfo & fireShield`, e.g. `LifeDrainInfo` and pass it as parameter to applyBattleEffects 
- this structure should contain int64_t that describes amount of drained health points and int32_t that describes number of resurrected units

dydzio0614: i think in H3 fire shield is applied after resurrection, so it is not accounted for

kambala-decapitator: It's easy to verify: just create a test map with respective units in oh3 editor, then you can test it in both oh3 and vcmi

kambala-decapitator:
When merging, please perform squash merge

MichalZr6:
Hi,
I have made some changes related to the issue. I know this is slightly different from what you suggested, but it might also be a good solution. Please review and let me know your thoughts :)
PS: should singleAttackDrainedLife and the map (or whole struct ?) be also serialized ?

server/battles/BattleActionProcessor.cpp
```diff
@@ -1456,7 +1446,8 @@ int64_t BattleActionProcessor::applyBattleEffects(const CBattleInfoCallback & ba
 		CStack::prepareAttacked(bsa, gameHandler->getRandomGenerator(), bai.defender->acquireState()); //calculate casualties
 	}
 
-	int64_t drainedLife = 0;
+	auto &drainedLife = attackerState->health.resurrectedInfo.singleAttackDrainedLife;
+	drainedLife = 0;
```
IvanSavenko, line 1450: Please avoid non-const references - they are often confusing.
Use singleAttackDrainedLife directly

IvanSavenko:
> I know this is slightly different from what you suggested, but it might also be a good solution. Please review and let me know your thoughts :) 

Will take a closer look a bit later.
This version is more complex and now also modifies gamestate and not just logging generation. Not sure if this is needed and whether it might have unintended side effects.

Also, I think it won't work if mod gives some creature both life drain and multi-attack. In this case creature would attack multiple enemies, but life drain would only report resurrected count for last attack.

> PS: should singleAttackDrainedLife and the map (or whole struct ?) be also serialized ?

If health is serialized then yes - add new properties as well

IvanSavenko:
Hmm... To be honest it feels that best approach would be to have `CUnitState::heal` return number of resurrected units & possibly - actual amount of healed health points (so if already fully healed vampires use life drain it would return 0).

MichalZr6:
The thing is CHealth so far stored in `resurrected` field the amount of oneBattle resurrected units (ex. using Resurrect spell with low level Earth Magic). I wanted to make its code more verbose, especially for `getResurrected()` and `takeResurrected()`. Overall however the logic should not differ from the original.

server/battles/BattleActionProcessor.cpp
```diff
@@ -1456,14 +1446,13 @@ int64_t BattleActionProcessor::applyBattleEffects(const CBattleInfoCallback & ba
 		CStack::prepareAttacked(bsa, gameHandler->getRandomGenerator(), bai.defender->acquireState()); //calculate casualties
 	}
 
-	int64_t drainedLife = 0;
+	battle::HealInfo tmpHealInfo;
```
kambala-decapitator, line 1449: why do you need this variable? You can write to the passed parameter directly

server/battles/BattleActionProcessor.cpp
```diff
@@ -1435,7 +1425,7 @@ void BattleActionProcessor::handleAfterAttackCasting(const CBattleInfoCallback &
 	}
 }
 
-int64_t BattleActionProcessor::applyBattleEffects(const CBattleInfoCallback & battle, BattleAttack & bat, std::shared_ptr<battle::CUnitState> attackerState, FireShieldInfo & fireShield, const CStack * def, int distance, bool secondary)
+void BattleActionProcessor::applyBattleEffects(const CBattleInfoCallback & battle, BattleAttack & bat, std::shared_ptr<battle::CUnitState> attackerState, FireShieldInfo & fireShield, const CStack * def, battle::HealInfo * healInfo, int distance, bool secondary) const
```
kambala-decapitator, line 1428: please change the new param to be a reference

lib/spells/effects/Summon.cpp
```diff
@@ -122,6 +122,7 @@ int32_t Summon::summonedCreatureAmount(const Mechanics * m) const
 
 void Summon::apply(ServerCallback * server, const Mechanics * m, const EffectTarget & target) const
 {
+	using battle::HealInfo;
```
kambala-decapitator, line 125: seems unused in this method?

lib/battle/Unit.h
```diff
@@ -41,6 +41,27 @@ namespace BattlePhases
...
+{
+	HealInfo()
+		: healedHealthPoints(0), resurrectedCount(0)
+	{ }
+	HealInfo(int64_t healedHP, int32_t resurrected)
+		: healedHealthPoints(healedHP), resurrectedCount(resurrected)
+	{ }
+
+	int64_t healedHealthPoints;
+	int32_t resurrectedCount;
```
kambala-decapitator, line 55:
```suggestion
	HealInfo() = default;
	HealInfo(int64_t healedHP, int32_t resurrected)
		: healedHealthPoints(healedHP), resurrectedCount(resurrected)
	{ }

	int64_t healedHealthPoints = 0;
	int32_t resurrectedCount = 0;
```

lib/battle/Unit.h
```diff
@@ -41,6 +41,27 @@ namespace BattlePhases
...
+
+	HealInfo& operator +=(const HealInfo& other)
```
kambala-decapitator, line 57: by our coding style it should be:
```suggestion
	HealInfo & operator+=(const HealInfo & other)
```

server/battles/BattleActionProcessor.cpp
MichalZr6: Yeah, it came to me.. just after the commit. Will fix.

IvanSavenko (commented):
Please fix issues mentioned by kambala.
Other than that - looks OK

IvanSavenko:
Can you also update test?
Check file `test/mock/mock_battle_Unit.h`
Please replace return type for `heal` method from `void` to `HealInfo `

Issue #3663, opened 2024-03-16
[HotA] Black Tower battle should be a regular layout instead of banks layout


thiscris:
**Describe the bug**
The Black Tower object added in HotA originally has a regular battlefield (enemies are on the right, attacking army on the left, with war machines if available). In the VCMI implementation the battlefield is similar to dragon utopia and other banks (enemies around the corners of the battlefield, attacking army in the middle, no war machines). As a result the Black Tower poses a greater challenge then the way it is implemented in the original HotA.

**Game logs**
not relevant

**To Reproduce**
1. Start a game
2. Visit a Black Tower map object

**Expected behavior**
The battlefield should be a regular one (standard battle)

**Actual behavior**
The battlefield is that used for banks battles

**Did it work earlier?**
I suppose not

**Screenshots**
![map object](https://github.com/vcmi/vcmi/assets/31136319/94bd9286-079b-4e61-aa79-48d096109356)

![battlefield](https://github.com/vcmi/vcmi/assets/31136319/46acda2f-8cac-44bc-8b02-3564e91f0cfb)

**Version**
 - OS: Linux
 - Version: 1.4.5

**Additional context**
Not sure if this should be treated as a feature request or a bug.

thiscris:
I think I should have opened this in https://github.com/vcmi-mods/horn-of-the-abyss/issues instead of here. Should I close this and open a new one there?

misiokles:
Developers seldom look for bugs in mods repositories. It's better to duplicate reports.

thiscris:
My guess is that all map objects battles in VCMI are treated as "banks" battles, that is why there is nothing mentioned in the json file for the black tower: https://github.com/vcmi-mods/horn-of-the-abyss/blob/vcmi-1.4/hota/Mods/mapObjects/content/config/hotaMapObjects/blackTower.json

DjWarmonger:
Probably we should add an extra option to bank config for battlefield layout.

DjWarmonger:
Implemented in https://github.com/vcmi-mods/horn-of-the-abyss/pull/108

PR #4052, opened 2024-05-29
Config option for regular unit placement in Black Tower

DjWarmonger:
Implements #3663

### DIFF
--- a/lib/mapObjects/CBank.h
+++ b/lib/mapObjects/CBank.h
@@ -22,6 +22,7 @@ class DLL_LINKAGE CBank : public CArmedInstance
 	ui32 daycounter;
 	ui32 resetDuration;
 	bool coastVisitable;
+	bool regularUnitPlacement;
 
 	void setPropertyDer(ObjProperty what, ObjPropertyID identifier) override;
 	void doVisit(const CGHeroInstance * hero) const;
--- a/lib/mapObjects/CBank.h
+++ b/lib/mapObjects/CBank.h
@@ -51,6 +51,10 @@ public:
 		h & bankConfig;
 		h & resetDuration;
 		h & coastVisitable;
+		if (h.version >= Handler::Version::BANK_UNIT_PLACEMENT)
+			h & regularUnitPlacement;
+		else
+			regularUnitPlacement = false;
 	}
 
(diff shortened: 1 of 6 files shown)


dydzio0614:
makes sense though i think we should make vastly different bank config and make unit placement configurable like siege moat hexes etc

IvanSavenko:
OK for a quick fix, but we need an option to configure guards placement precisely, possibly using some common presets to reduce copy-paste.

Hota banks like Churchyard use layout that is different from both h3 banks and regular battles

DjWarmonger:
I see, but where does this actually come from? I mean, numbers 6 and 7?

```
			/// Description of guards, stacks will be ordered
			/// on battlefield according to this scheme:
			/// 4    7    1
			/// 
			/// 6         5
			/// 
			/// 3         2
```

IvanSavenko:
> I see, but where does this actually come from? I mean, numbers 6 and 7?

You mean why such placement was chosen? No reason, simply to have well-defined behavior if there are banks with more than 5 stacks. Feel free to change.

If you mean where these positions are defined then this file:
https://github.com/vcmi/vcmi/blob/develop/config/battleStartpos.json


Also - you want to have this in 1.5.2?

DjWarmonger:
I think yes, quick fix for 1.5.2 would be nice with minimal chance of bugs.

lib/mapObjects/CBank.h
```diff
@@ -22,6 +22,7 @@ class DLL_LINKAGE CBank : public CArmedInstance
 	ui32 daycounter;
 	ui32 resetDuration;
 	bool coastVisitable;
+	bool regularUnitPlacement;
```
IvanSavenko, line 25: This might be a problem for a hotfix, not sure how to solve it nicely.

It needs to be serialized, as everything in CBank.
You can add check for save version and it would work for saves.

But in multiplayer we don't have save compatibility checks. Meaning, 1.5.1 client won't be able to deserialize gamestate from 1.5.2 server (and vice versa).

IvanSavenko: I think I have an idea how to handle multiplayer.

So this PR should be OK for 1.5.2

PR #4061, opened 2024-05-30
Support HotA thives guild

Laserlicht:
HotA mod also has to be changed:
`"special3":       { "type" : "thievesGuild", "id" : 22, "requires" : [ "tavern" ] },`

Missing (not part of PR) but looked also a little bit into it -> but found no solutions in reasonable time...:
- **Grotto** -> maybe something for @SoundSSGood ? The main problem here is that the sacrifice dialog is needed without hero "dependency" and always as "neutral".
- **Bank** -> Maybe we need a functionality to buy a town bonus in structure. Buying the bonus for -2500 gold and it gives you -500 gold for 5 days. New bonus can only buyed again is old bonus isn't exist anymore. I think this functionality would be really generic and also really interesting for other modding stuffs.


IvanSavenko:
> Bank -> Maybe we need a functionality to buy a town bonus in structure

Strictly speaking, this logic is doable as rewardable map object. And strictly speaking, town buildings can use rewardable objects logic.

However it is undocumented and untested neither by vcmi nor by any mods since... 1.3 I think? So it might still work. Or might not. Many thanks to author for not doing any of these 3 things...

lib/CGameInfoCallback.cpp
```diff
@@ -224,15 +224,17 @@ void CGameInfoCallback::getThievesGuildInfo(SThievesGuildInfo & thi, const CGObj
...
+	if(obj->ID == Obj::TOWN || obj->ID == Obj::TAVERN)
 	{
 		int taverns = 0;
 		for(auto town : gs->players[*getPlayerID()].towns)
 		{
 			if(town->hasBuilt(BuildingID::TAVERN))
 				taverns++;
+			
+			if(town->hasBuilt(BuildingSubID::THIEVES_GUILD))
+				taverns += 2;
```
IvanSavenko, line 236: Not possible to configure at the moment? Not sure if town buildings have configurable parameters...

IvanSavenko (approved):

lib/CGameInfoCallback.cpp
Laserlicht: Looked into the schema files. I don't think it's possible now. :(

SoundSSGood:
>  and always as "neutral"

As far as I remember, I did not enable the grotto in the cove because there were some problems with special buildings. If you can do it, enable it. I'll correct the neutral issue after.

Another problem is that in VCMI, the implementation of a store in a city involves buttons to switching the available trading modes. Grotto will add two buttons at once. And it is needed to find a place for these buttons, or not add them at all.

Laserlicht:
> As far as I remember, I did not enable the grotto in the cove because there were some problems with special buildings. If you can do it, enable it. I'll correct the neutral issue after.

See here: https://github.com/vcmi/vcmi/commit/873c4c0ec39b1acc0aa0444c8603b386d53aa27b
Basically it's work. I can do trades creatures to experience.
Buttons are still a problem:
grafik

And this: https://github.com/vcmi/vcmi/blob/0d99adc15af3b76ad0a9450e6f87e28126fd0c88/client/widgets/markets/CAltarArtifacts.cpp#L34
wont work with `town`.
CAltarCreatures is no problem here, but CAltarArtifacts.

Any idea to fix?

> Another problem is that in VCMI, the implementation of a market in a town involves buttons to switching the available trading modes. Grotto will add two buttons at once. And it is needed to find a place for these buttons, or not add them at all.

I don't understand the problem here!?
Grotto in official HotA itself:
grafik
grafik

Altar of Sacrifice in VCMI:
grafik
grafik

SoundSSGood:
> I don't understand the problem here!?

See your first screenshot. Mode switch button in the lower left corner. 
Buttons for the creatures altar and for the artifacts altar allso can be added to the resource market window (two buttons). This is if you follow the VCMI style. If not, then you can simply remove all the buttons and the altar will work separately from the market (Hota style).

> Any idea to fix?

Don't worry about it.

SoundSSGood:
> Buttons for the creatures altar and for the artifacts altar allso can be added to the resource market window (two buttons). This is if you follow the VCMI style.

The idea that was implemented in VCMI is the ability to directly switch from any market window to any other available one.

Laserlicht:
@SoundSSGood https://github.com/Laserlicht/vcmi/commit/fead50c1d77d7393d34bcb1811bcf67812a9aca4 fixes the frontend problems...

But I think it's better to keep grotto out of this PR.

Do you want to add these necessary changes (from the two commits) into market rework? And also change the remaining backend logic?

SoundSSGood:
> But I think it's better to keep grotto out of this PR.

OK

> Do you want to add these necessary changes (from the two commits) into market rework? And also change the remaining backend logic?

I think, yes

PR #4071, opened 2024-06-01
[1.6] Fix potential desync if client uses different stdlib with different random number generators

IvanSavenko:
While this is only potential problem, it may still lead to extremely hard to track issues if client and server end up with different states.

Standard has no guarantees that random numbers generators (or distributions) produce identical results on every platform. In addition we also serialize generators to string, which is very likely to have different results.

To solve this, all random number rolls that involve game mechanics must be performed only on server. This is already *mostly* the case, but there are some exceptions that are not immediately obvious. This PR should track down such cases & resolve them

Changes in this PR:
- Library and server now primary use vstd::RNG interface instead of direct access to CRandomGenerator class
- Simplified logic of CRandomGenerator and moved some its methods to vstd::RNG interface
- CGameState no longer has own random number generator, instead it uses generator provided by server. Attempt to use it on client will throw an exception.
- Removed mostly unused seeds from StartInfo. Game seed is now only known to server.

TODO:
- [x] Make rumors generation server-sided
- [x] Make rewardable objects re-randomization server-sided
- [x] Make creature banks reset server-sided
- [x] Look for any other cases of rng access from client

### DIFF
--- a/AI/BattleAI/StackWithBonuses.h
+++ b/AI/BattleAI/StackWithBonuses.h
@@ -24,20 +24,17 @@ class HypotheticBattle;
 class RNGStub : public vstd::RNG
 {
 public:
-	vstd::TRandI64 getInt64Range(int64_t lower, int64_t upper) override
+	int nextInt(int lower, int upper) override
 	{
-		return [=]()->int64_t
-		{
-			return (lower + upper)/2;
-		};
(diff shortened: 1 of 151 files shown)


IvanSavenko:
Seems to be ready

PR #4097, opened 2024-06-06
[1.6] Render refactor

IvanSavenko:
Another part of reworking of our rendering logic, hopefully approaching finish line. All image loading logic is now part of (previously placeholder) RenderHandler class. Should also speed up game loading thanks to lazy loading & reduce data duplication caused by loading of same image multiple times.

- All images (whether animation frames or single bitmaps) are now loaded & managed by RenderHandler
- RenderHandler now caches all image access. Multiple requests for the same image will no longer lead to loading of duplicated image data. Similarly, any images that use vertical or horizontal flips are now also done via RenderHandler and tracked
- IImage interface implementations no longer own image data. Instead they hold reference to potentially shared image data and only own instance-specific data, such as transparency factor or transformed palette
- CAnimation class now uses lazy loading and only loads additional frames on request. This speeds up loading to adventure map and to battle since game no longer needs to decode hundreds of frames at once.
- Removed multiple now redundant animation caching implementations
- University window now receives updated hero skill via call on netpack instead of refreshing state on every frame
- Unified naming of `setPlayerColor` functions (colorize, playerColor, playerColored, etc)
- Fixed excessive (~200 Mb) memory allocation on adventure map (should be probably backported to 1.5)

TODO:
- [ ] Make image caching less aggressive and set upper limit to cache size instead of current unlimited cache
- [ ] Reduce number of palette copies. Instead use reference counting provided by SDL. For example, when loading .def there is no need to create separate palette for every tiny frame - we can share single palette between all frames. Memory usage savings are small for a single image, but H3 has thousands of such images.
- [x] Review cases in which we use SDL blit and in which - our slow custom blit that can use alpha channel in palettes. (In SDL3 we will be able to drop our custom blit completely, but that won't happen soon)
- [x] Remove custom animations generation from obstacle rendering in battle interface
- [x] Remove custom animation generation from ranged attack highlight and move it to json config
- [ ] Consider whether to move existing image scaling handling to RenderHandler (but probably without caching)

Things to check for regressions:
- [x] Creatures in battles
- [x] Obstacles in battles
- [x] 'Absolute' obstacles in battles

### DIFF
--- a/Mods/vcmi/Sprites/battle/rangeHighlights/rangeHighlightsGreen.json
+++ b/Mods/vcmi/Sprites/battle/rangeHighlights/rangeHighlightsGreen.json
@@ -2,33 +2,33 @@
 	"basepath" : "battle/rangeHighlights/green/",
 	"images" :
 	[
-		{ "frame" :  0, "file" : "empty.png"},						// 000001 -> 00 empty frame
+		{ "frame" :  0, "file" : "empty.png"},                                                              // 000001 -> 00 empty frame
 			
 		// load single edges
-		{ "frame" :  1, "file" : "topLeft.png"},                    //000001 -> 01 topLeft
-		{ "frame" :  2, "file" : "topLeft.png"},                    //000010 -> 02 topRight
-		{ "frame" :  3, "file" : "left.png"},                       //000100 -> 04 right
-		{ "frame" :  4, "file" : "topLeft.png"},                    //001000 -> 08 bottomRight
-		{ "frame" :  5, "file" : "topLeft.png"},                    //010000 -> 16 bottomLeft
-		{ "frame" :  6, "file" : "left.png"},                       //100000 -> 32 left
+		{ "frame" :  1, "file" : "topLeft.png"},                                                            //000001 -> 01 topLeft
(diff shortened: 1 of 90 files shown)


dydzio0614:
I get spam in console with text "SDL_UpperBlit failed! Blit combination not supported", it goes away when i scroll into unexplored area on adventure map

obraz

dydzio0614:
also i noticed that roads look weird, also visible on screenshot from above and this one as well 
obraz

IvanSavenko:
Road - might be the reason behind those SDL errors. Will check.

Opened PR because wanted to focus on Launcher issues, even though some stuff is unfinished here.

dydzio0614:
well i only wanted to check ram memory improvements for now :P

dydzio0614:
disabled roads in random map generator - error gone :P

IvanSavenko:
Try playing in 4k (or with 50% scaling on FullHD). 
I think VCMI would allocate like 500 Mb for map view in there :)

Also right now new cache is unlimited in size. Meaning if you play long enough vcmi would load all h3 assets in memory. But hey, way lower loading times!

IvanSavenko:
Probably ready. Will finish remaining TODO's in separate PR to avoid this one from growing too much

PR #4098, opened 2024-06-06
Launcher settings update

IvanSavenko:
(to be merged after Android Launcher PR)

Since Android finally getting Qt launcher - update to settings window in Launcher, both to account for Android support and to add options that already existed before but not accessible to players.

Once Android launcher PR is merged - will rebase this PR and recheck that all options make sense, e.g. that we don't show desktop-only options on Android.

List of changes:
- Removed "Heroes 3 data files language" field from UI - confusing and auto-detection works well enough.
- Added option to select upscaling filter used by SDL
- Added option to select sound/music volume in Launcher
- (mobile only) Added option to use 'relative cursor mode' that was accessible via Android launcher
- Added option to toggle haptic feedback / vibration via settings
- Added option to configure long touch duration
- Added option to configure input tolerance (aka fatty finger mode from touchscreen) for any input type - mouse, controller, touch.
- Added option to configure online lobby address & port
- VSync now uses drop-down instead of hard to hit checkbox, in line with other settings
- On mobile platforms, cursor types now shows "Enabled" / "Disabled" text instead of unclear "hardware" / "software"
- Changed location of map editor & launcher Qt settings (e.g. window size & position) to be in the same location as rest of vcmi config files instead of separate "VCMI Team" directory)
- Remove no longer needed other_cp1250, other_cp1251 and other_cp1252 pseudo-languages

Changed added after initial PR:
- (mobile only) haptic feedback is now enabled by default
- (mobile only) added UI to toggle haptic feedback option
- Added option to configure controller sticks sensitivity & acceleration
- Replaced checkboxes with toggleable buttons for more convenient UI on mobile systems
- Replaced dropdowns with on/off parameters with toggleable buttons
- 

Screenshot:
settings


Laserlicht:
Notes/suggestions:
- Disable "Repository download on start" on mobile by default (or adding a third option only checking for updates of already existing mods) -> avoid downloading unnecessary files on mobile network
- Should we set haptic feedback by default "on"?
- Rename vibration to hapic feedback? (could be confusing because a controller also can "vibrate")
- replace checkboxes by "on"/"off" for the remaining things (at least autosave)

IvanSavenko:
> * Disable "Repository download on start" on mobile by default (or adding a third option only checking for updates of already existing mods) -> avoid downloading unnecessary files on mobile network

Yeah, not yet sure what to do with it. Such third option is hard to do. Wanted to check how large our repo checkout is. If it is small enough then we can keep it on. If it is large - then yeah, might be a good idea to have it off by default. Or cache between runs & use old checkout.

>  * Should we set haptic feedback by default "on"?

It is not? My screenshots don't show default state. Default values are same as in 1.5.2.
In any case - personally I am fine with either option.

> * Rename vibration to hapic feedback? (could be confusing because a controller also can "vibrate")

But this is phone-only option? Or you mean controllers attached to phone?

> * replace checkboxes by "on"/"off" for the remaining things (at least autosave)

Yeah, will do soon.
Although not sure whether to use dropdown with "on" / "off" options or checkable button like in https://github.com/vcmi/vcmi/pull/4106 (but without icon) - to avoid that drop-down with just 2 elements

Laserlicht:
> It is not? My screenshots don't show default state. Default values are same as in 1.5.2. In any case - personally I am fine with either option.

No currently not. Do you want to add it in this PR?

> But this is phone-only option? Or you mean controllers attached to phone?

Yes phone attached controller. Or one of this many android based tv boxes with bluetooth controller.

PR #4099, opened 2024-06-06
Fixes to help tab in Launcher

IvanSavenko:
Will merge after Android launcher PR

- Added "Configuration files directory" entry since it is located in different path at least on Linux and may be of interest to a player
- Removed non-functioning "Open directory" buttons on mobile systems and stretched text box into now-empty space

Screenshot (desktop):
help-desktop

Screenshot (mobile emulation):
help-mobile

### DIFF
--- a/launcher/aboutProject/aboutproject_moc.cpp
+++ b/launcher/aboutProject/aboutproject_moc.cpp
@@ -16,6 +16,20 @@
 #include "../../lib/GameConstants.h"
 #include "../../lib/VCMIDirs.h"
 
+void AboutProjectView::hideAndStretchWidget(QGridLayout * layout, QWidget * toHide, QWidget * toStretch)
+{
+	toHide->hide();
+
+	int index = layout->indexOf(toStretch);
+	int row;
+	int col;
+	int unused;
+	layout->getItemPosition(index, &row, &col, &unused, &unused);
+	layout->removeWidget(toHide);
+	layout->removeWidget(toStretch);
+	layout->addWidget(toStretch, row, col, 1, -1);
+}
+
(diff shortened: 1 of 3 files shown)


launcher/aboutProject/aboutproject_moc.cpp
```diff
@@ -16,6 +16,20 @@
...
+	toHide->hide();
+
+	int index = layout->indexOf(toStretch);
+	int row;
+	int col;
+	int unused;
+	layout->getItemPosition(index, &row, &col, &unused, &unused);
+	layout->removeWidget(toHide);
+	layout->removeWidget(toStretch);
+	layout->addWidget(toStretch, row, col, 1, -1);
```
kambala-decapitator, line 30: doesn't it stretch automatically if you simply remove & delete button from the row?

kambala-decapitator (commented):
.ui file has changed a lot, but I think you only added new row?

IvanSavenko:
> .ui file has changed a lot, but I think you only added new row?

And renamed some widgets that had auto-generated names. No other changes.
I actually have newer Qt Creator than before, so this might be the main reason.

launcher/aboutProject/aboutproject_moc.cpp
IvanSavenko: If entire column is empty Qt will resize it to ~0px
But this column in layout still has one element - "Check for updates" button.

Not sure how to deal with that other than splitting layouts. And splitting layouts would make columns in "Build information" and "data directories have different width which does not looks nice.

kambala-decapitator: Yeah, the only way would be to replace grid with rows

PR #4100, opened 2024-06-06
Fix lag on dragging mod list in Launcher on Android

IvanSavenko:
Tested by enabling dragging on desktop. According to profiler, most of time was spent on generating mod information. Cheap to do once, but it looks like Qt does this A LOT.

Added simple caching & reduced copying. Seems to be working fine on PC, but not yet confirmed whether same goes for mobile.
Cache resets may be to excessive, but better to play it safe. Seems to be fast enough anyway. Method that previously took ~90% of time before is now completely gone from profiler view.


Laserlicht:
Much better now. For my devices it's good.

PR #4108, opened 2024-06-08
Fix .ui files not visible in generated project (QtCreator/CMake/Ninja)

IvanSavenko:
Seems to be caused by #3817
Qt .ui files are no longer visible in generated project. Project builds successfully, and you can still open .ui files manually, but it is less convenient than before.

Probably caused by enabling AUTO_UIC
Not sure if there is proper/better fix, but this change seems to be working on my system


kambala-decapitator (approved):
Thanks, missed to add ui to sources

Issue #3584, opened 2024-01-31
NKAI support for new map object


DjWarmonger:
**Describe your proposal**
Right now NKAI has hardcoded values for vanilla objects, but does not visit new objects added in mods. This includes HoTA buildings (Warehouses or Colosseum of the Magi), Antagarich Burning assets, WoG objects (these are weird tho) and more.

It already does support banks that give standard rewards (but not spells IIRC).

TODO:

- [ ] Make NKAI actually read object config instead of using hardcoded values
- [ ] Pre-calculate values of all objects once on AI launch, rather than every time it checks an object
- [ ] Handle objects casting spells (Town Gate and Mirror of the Home-Way)
- [ ] Handle WoG Mushrooms
- [ ] Handle Elemental Nodes
- [ ] Allow AI to deliberately search object types by particular property (resources, monsters, stats, other?)
- [ ] Possibly support other reward types

nullkiller:
Town Gate - if I understand it correctly - this should be a part of hero chaining implemented in form of special action allowing to visit towns from that point where it stands. What Mirror of the Home-Way is?

nullkiller:
Spells are now converted into skill reward mostly so same approach as with CBank can be used

DjWarmonger:
>  What Mirror of the Home-Way is?

It's a WoG object that also casts Town Portal, but costs 1000 gold to use.

PR #4110, opened 2024-06-08
NKAI: rewardables

nullkiller:
Basic support of rewardables. Now AI visits gold warehouse.
Partially address #3584

### DIFF
--- a/AI/Nullkiller/AIUtility.cpp
+++ b/AI/Nullkiller/AIUtility.cpp
@@ -423,8 +423,15 @@ bool shouldVisit(const Nullkiller * ai, const CGHeroInstance * h, const CGObject
 		return false;
 	}
 
-	if(obj->wasVisited(h)) //it must pointer to hero instance, heroPtr calls function wasVisited(ui8 player);
+	if(obj->wasVisited(h))
 		return false;
+
+	auto rewardable = dynamic_cast<const Rewardable::Interface *>(obj);
+
+	if(rewardable && rewardable->getAvailableRewards(h, Rewardable::EEventType::EVENT_FIRST_VISIT).empty())
+	{
+		return false;
+	}
 
 	return true;
 }
--- a/AI/Nullkiller/Behaviors/StartupBehavior.cpp
+++ b/AI/Nullkiller/Behaviors/StartupBehavior.cpp
@@ -79,6 +79,15 @@ bool needToRecruitHero(const Nullkiller * ai, const CGTownInstance * startupTown
 		bool isGoldPile = dynamic_cast<const CGResource *>(obj)
 			&& dynamic_cast<const CGResource *>(obj)->resourceID() == EGameResID::GOLD;
 
+		auto rewardable = dynamic_cast<const Rewardable::Interface *>(obj);
+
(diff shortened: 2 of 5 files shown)


DjWarmonger:
Will look into it

DjWarmonger:
Okay, at the very least AI picks Vases on Dunes terrain, that's easy to observe.

How about objects which cost gold (negative gold alue), such as Gazebo (Highlands), Genie's Lamp (HoTA) or Gypsy Caravan (Antagarich Burning assets:?

nullkiller:
Negative rewards? Will take a look.

dydzio0614:
i think complex logic would be needed to make AI judge fairly objects like shop that sells implosion spell for 30 gems

nullkiller:
I was surprised but it works! It handles gold cost through limiters. I only fixed it not producing negative gold reward for evaluation but it actually does not matter much

nullkiller:
[bring-genies.zip](https://github.com/user-attachments/files/15944431/bring-genies.zip)
A sample map with ancient lamps

AI/Nullkiller/AIUtility.cpp
```diff
@@ -421,9 +421,16 @@ bool shouldVisit(const Nullkiller * ai, const CGHeroInstance * h, const CGObject
 	case Obj::BOAT:
 	case Obj::SIGN:
 		return false;
+	case Obj::SHRINE_OF_MAGIC_GESTURE:
+	case Obj::SHRINE_OF_MAGIC_INCANTATION:
+		return h->hasSpellbook();
+	case Obj::SHRINE_OF_MAGIC_THOUGHT:
+		return h->hasSpellbook() && h->getSecSkillLevel(SecondarySkill::WISDOM) >= MasteryLevel::BASIC;
+	case Obj::PYRAMID:
+		return h->hasSpellbook() && h->getSecSkillLevel(SecondarySkill::WISDOM) == MasteryLevel::EXPERT;
```
IvanSavenko, line 430: Kind of wish to have this somewhere inside rewardable object...

IvanSavenko (approved):

AI/Nullkiller/AIUtility.cpp
IvanSavenko: I think calling `getAvailableRewards(h, Rewardable::EEventType::EVENT_FIRST_VISIT);` should tell whether it makes sense for AI to visit this object, but needs tests

nullkiller: Ok will take a look onto your suggestion

nullkiller: getAvailableRewards is protected, but exactly what I did in other places except EEventType. Have not tested it, just hero limiter. So getAvailableRewards can be used if we make it public

nullkiller: So can I make it public?

IvanSavenko: Yeah, sure. 
I usually mark stuff as private / protected to indicate that it is not used outside of the class, so generally safe to make public.

PR #4132, opened 2024-06-12
Try to enable AUTO_RCC on msvc 

IvanSavenko:
Added debug logging to cmake to try to figure out reason behind CI failure on msvc

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -66,7 +66,7 @@ jobs:
             pack: 1
             pack_type: RelWithDebInfo
             extension: exe
-            preset: windows-msvc-release-ccache
+            preset: windows-msvc-release
           - platform: mingw
             os: ubuntu-22.04
             test: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/github.yml
```diff
@@ -208,11 +208,11 @@ jobs:
     - name: Configure
       run: |
         if [[ ${{matrix.preset}} == linux-gcc-test ]]; then GCC14=1; fi
-        cmake -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }} ${GCC14:+-DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14}
+        cmake -CMAKE_VERBOSE_MAKEFILE=ON -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }} ${GCC14:+-DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14}
```
kambala-decapitator, line 211: you missed leading `D` in both places :)

IvanSavenko:
So, looks to be caused by combination of msvc, ccache and autorcc.

Not exactly sure what caused this bug, but disabling ccache for msvc seems to be fixing this bug. And it does not looks to be functioning anyway - msvc CI run always takes 30+ minutes.

What seems to be happening is that cmake generates msvc projects that have custom steps to:
- compile each of translations (ts -> qm) 
- step that compiles these translations into a single resource file (qm -> cpp)
- After which this generated file is (presumably) used as input to final executable.

But if ccache is enabled, msbuild would run qm ->cpp step without running ts -> qm step at all. Maybe ccache attempted to cache qm's somehow? Or something is wrong with msbuild in this scenario?

In any case unless somebody is willing to investigate this further I'd rather just disable ccache for msvc - if only to fix broken icons & translations in msvc launcher build

kambala-decapitator:
ccache explains why AUTO_RCC worked on my local machine :) let's just disable ccache then.

.github/workflows/github.yml
```diff
@@ -205,11 +205,18 @@ jobs:
...
         PULL_REQUEST: ${{ github.event.pull_request.number }}
 
-    - name: Configure
+    - name: Configure (ccache)
+      if: ${{ matrix.platform != 'msvc' }}
       run: |
         if [[ ${{matrix.preset}} == linux-gcc-test ]]; then GCC14=1; fi
         cmake -DENABLE_CCACHE:BOOL=ON --preset ${{ matrix.preset }} ${GCC14:+-DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14}
 
+    - name: Configure (no ccache)
```
kambala-decapitator, line 214: instead of creating new step you could simply check if `preset` ends with `ccache` and add `-DENABLE_CCACHE:BOOL=ON` based on that

PR #4152, opened 2024-06-17
[1.5.3] Fix rmg generation if player list is not continuous

### LINKED ISSUE #1233: Player doesn't participate in game
1) Open random map menu
3) Keep all map settings random
4) Open additional settings
5) Select pink flag (the last one) for human player
6) Start game

Expected: game is started - player has pink flag
Actual: game is started between AI players only - they are playing in background

Note: if RMG will generate map for 8 players issue will not be reproduced, so it's better to generate M+U or L map

IvanSavenko:
Since most common crashes are now fixed, this crash became one of the most common crashes on Google Play stats.

Main source of the crash is that when player selects 'random' number of players and then selects pink color to play as, game can roll 2 for random number of players and generate map with red & blue players, without pink human player.

With this change, game will always keep colors that were allocated to human players when 'random' number of players is chosen, so:
a) if player selects random template, random number of players and pink human player, and game rolls for 4 players, then map would have red, blue, brown, and pink player.
b) if player selects template with 3+ players (e.g. Jebus), picks green human player, and game rolls for 2 players, then map would have red & green players on map.

Also, fixes bug where changing player color would mark new color as 'human-controlled' (correct), but would not erase such mark from old color.

- Fixes #1233

### DIFF
--- a/lib/rmg/CMapGenOptions.cpp
+++ b/lib/rmg/CMapGenOptions.cpp
@@ -284,9 +284,12 @@ void CMapGenOptions::resetPlayersMap()
 
 	while (players.size() > realPlayersCnt)
 	{
-		while (eraseLastPlayer(EPlayerType::AI));
-		while (eraseLastPlayer(EPlayerType::COMP_ONLY));
-		while (eraseLastPlayer(EPlayerType::HUMAN));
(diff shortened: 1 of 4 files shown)


DjWarmonger (approved):
Works ok, looks like I made a mistake in player erasing logic.

PR #4156, opened 2024-06-17
[1.5.3] Crashfixes

IvanSavenko:
Fixes for few more crashes from Google Play:
- Attempt to fix crash on 'emergency save' when Android requests complete app shutdown to free resources
- Add better crash message for creatures with missing battle animation
- Fix crash on receiving high score value outside of 0...500 range.
- Avoid crash if town building selection highlight area image is missing
- Replace boost with MetaString in tavern window to avoid crash in some localizations (no '%d' placeholder in string)

### DIFF
--- a/config/highscoreCreatures.json
+++ b/config/highscoreCreatures.json
@@ -1,6 +1,6 @@
 {
     "creatures": [
-        { "min" : 1, "max" : 4, "creature": "imp" },
+        {            "max" : 4, "creature": "imp" },
         { "min" : 5, "max" : 8, "creature": "gremlin" },
         { "min" : 9, "max" : 12, "creature": "gnoll" },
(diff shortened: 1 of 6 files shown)


config/highscoreCreatures.json
```diff
@@ -1,6 +1,6 @@
 {
     "creatures": [
-        { "min" : 1, "max" : 4, "creature": "imp" },
```
Laserlicht, line 3: That's also a functional change I think... Currently cheater (0 points) has no creature. I think that's better than "imp" for cheater.

IvanSavenko: That's probably a separate logic. 
If creature is actually not found then it will result in a crash.
Besides - score of 0 is legal, if you play long enough.

PR #4169, opened 2024-06-20
Attempt of fix #1294 stables and mana vortex greetings

### LINKED ISSUE #1294: the "stables" message appears on new week start
Steps to reproduce:
1. build Stables (Castle)
2. place a hero in this town (in garrison)
3. wait till next week
4. observe the message after closing the "new week" dialog

Correct behavior: message appears only when you open the town.

MichalZr6:
just a proposal for the final fix. Why pendingGreeting is empty in OpenTownWindow ?


IvanSavenko:
VCMI uses client-server model (even though they are now in a single process). 

This means that 
1. Client and server operate using their own, separate instances of gamestate (but shared & permanent VLC)
2. All changes to gamestate (e.g. to town instance / town buildings) must be done via netpack appliers
3. Client should only read from gamestate, and never write.

In your case, `onVisit` is executed on server, while `openTownWindow` is client-side call. 
Hmm... another potential problem with your approach (even if server->client propagation is fixed) is that effect is still applied on day 1, and not when client opens window.

MichalZr6:
Thanks for clarification. I'll try to come up with some solution.

PR #4171, opened 2024-06-20
reset tutorial button

Laserlicht:
possibility to reset tutorial in launcher

QT-Tool reordered ui file. Sorry, no idea how to avoid...


Alexander-Wilms:
I found https://github.com/davidosterberg/Qt-UI-file-sorter

One could add a git hook that runs it before each commit that touches a .ui file.

PR #4172, opened 2024-06-20
Sort QGridLayout children in *.ui files

Alexander-Wilms:
These get put in a random order when editing a .ui files with Qt Designer, resulting in messy Git diffs.

As part of subsequent PRs making changes to the files, they should be sorted as well.

Script used:

https://github.com/rvjr/Qt-UI-file-sorter/tree/UTF-8

which is a fork of

https://github.com/davidosterberg/Qt-UI-file-sorter


Alexander-Wilms:
Output for unordered .ui files:

image

After running `find . -type f -name "*.ui" -print | xargs python3 CI/linux-qt6/sort_ui/sort_ui.py --no-remove-stdset --no-remove-native`:

image

kambala-decapitator:
is this really needed? As I understand, ui file gets "broken" only once, all subsequent saves don't change file structure any more.

kambala-decapitator:
oh, it's only about QGridLayout, then it should be useful. Strange though, I couldn't find any bug report about that at https://bugreports.qt.io/.

CI/linux-qt6/sort_ui/sort_ui.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python3
```
kambala-decapitator, line 1: you could add this as a submodule instead of copying

kambala-decapitator:
ideally it should run as pre-commit hook

CI/linux-qt6/sort_ui/sort_ui.py
Alexander-Wilms: I was considering it, but thought a copy would be better since the original repo has 3 stars and the fork none. So the repos might just get deleted at some point.

IvanSavenko:
Since this is (mostly) an automated change let's merge it after final 1.5.X release is out & beta branch is gone to avoid painful conflicts.

Alexander-Wilms:
Qt Designer 6.7.2 doesn't seem to put elements in a random order anymore.

PR #4177, opened 2024-06-22
handle more error cases with innoextract

Laserlicht:
Changes suggested by @IvanSavenko 
- check for Gog Galaxy installer
- ask for bin file first
- expose innoextract exceptions to user error messagebox

Selected branch is `master`. Hotfix?

### DIFF
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -337,10 +337,21 @@ void FirstLaunchView::extractGogData()
 		if (!is)
 			return false;
 		is >> std::noskipws;
+		is.seekg(0, std::ios::end);
+		std::streampos fileSize = is.tellg();
+		is.seekg(0, std::ios::beg);
+
+		if(fileSize > 10 * 1024 * 1024)
+			return false; // avoid to load big files; galaxy exe is smaller...
+
+		std::vector<char> buffer;
+		buffer.reserve(fileSize);
+		buffer.insert(buffer.begin(), std::istream_iterator<char>(is), std::istream_iterator<char>());
+
 		std::array<char, 20> magic_id{ 0x47, 0x00, 0x4F, 0x00, 0x47, 0x00, 0x20, 0x00, 0x47, 0x00, 0x61, 0x00, 0x6C, 0x00, 0x61, 0x00, 0x78, 0x00, 0x79, 0x00 }; //GOG Galaxy
-		auto eos = std::istream_iterator<char>();
-		auto res = std::search(std::istream_iterator<char>(is), eos, magic_id.begin(), magic_id.end());
-		return res != eos;
+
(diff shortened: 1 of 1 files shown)


Laserlicht:
@IvanSavenko Any idea why it fails on android NDK?

IvanSavenko:
> @IvanSavenko Any idea why it fails on android NDK?

Looks like `istream_iterator` can't be used with `std;:search`. Perhaps read this data to buffer and compare arrays?

Laserlicht:
> > @IvanSavenko Any idea why it fails on android NDK?
> 
> Looks like `istream_iterator` can't be used with `std;:search`. Perhaps read this data to buffer and compare arrays?

Strange. On linux this wasn't a problem with gcc and clang...
Tried another approach. Hopefully it will compile.

Laserlicht:
@IvanSavenko  Seems to compile now. But there are some other problems. Any idea?

IvanSavenko:
Ah, CI is a bit broken in master due to urgent Google Play upload. Rebase to beta for now.

Laserlicht:
Contains also your both `master` commits... Hopefully it's ok.

Laserlicht:
Seems to work now (tested on android and linux). :)

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -352,6 +358,33 @@ void FirstLaunchView::extractGogData()
...
+			if(fileSize > 10 * 1024 * 1024)
+				return false; // avoid to load big files; galaxy exe is smaller...
+
+			std::vector<char> buffer;
+			buffer.reserve(fileSize);
+			buffer.insert(buffer.begin(), std::istream_iterator<char>(is), std::istream_iterator<char>());
+
+			std::array<char, 20> magic_id{ 0x47, 0x00, 0x4F, 0x00, 0x47, 0x00, 0x20, 0x00, 0x47, 0x00, 0x61, 0x00, 0x6C, 0x00, 0x61, 0x00, 0x78, 0x00, 0x79, 0x00 }; //GOG Galaxy
+
+			auto res = std::search(buffer.begin(), buffer.end(), magic_id.begin(), magic_id.end());
```
kambala-decapitator, line 381: is the header located at a random place and not at 0x0?

kambala-decapitator:
QFile + QByteArray would probably be easier to use instead of std classes to search for the header

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Yes. At a "random" position. It isn't really a header. It's a String containing "GOG Galaxy". So basically this way it should work with newer versions too.

kambala-decapitator: > It's a String containing "GOG Galaxy"

then in code you should use something like `u"GOG Galaxy"` to get utf-16 string and then view it as a byte array

PR #4185, opened 2024-06-23
Fix typos

Alexander-Wilms:


Alexander-Wilms:
I went through all the changes, so it should be mergeable without someone reviewing everything again.

lib/CGameInfoCallback.cpp
```diff
@@ -781,12 +781,12 @@ int CPlayerSpecificInfoCallback::getHeroSerial(const CGHeroInstance * hero, bool
...
 
-	for (auto & heroe : heroes)
+	for (auto & hero : heroes)
 	{
-		if (includeGarrisoned || !(heroe)->inTownGarrison)
+		if (includeGarrisoned || !(hero)->inTownGarrison)
 			index++;
 
-		if (heroe == hero)
+		if (hero == hero)
```
IvanSavenko, line 789: Something went wrong here.

lib/CSoundBase.h
```diff
@@ -259,7 +259,7 @@ VCMI_SOUND_NAME(Dig) VCMI_SOUND_FILE(DIGSOUND.wav) \
 VCMI_SOUND_NAME(DIPMAGK) VCMI_SOUND_FILE(DIPMAGK.wav) \
 VCMI_SOUND_NAME(DISEASE) VCMI_SOUND_FILE(DISEASE.wav) \
 VCMI_SOUND_NAME(DISGUISE) VCMI_SOUND_FILE(DISGUISE.wav) \
-VCMI_SOUND_NAME(DISPELL) VCMI_SOUND_FILE(DISPELL.wav) \
+VCMI_SOUND_NAME(DISPEL) VCMI_SOUND_FILE(DISPEL.wav) \
```
dydzio0614, line 262: @IvanSavenko should file name change stay as it was?

lib/CSoundBase.h
```diff
@@ -842,7 +842,7 @@ VCMI_SOUND_NAME(RDDRMove) VCMI_SOUND_FILE(RDDRMOVE.wav) \
 VCMI_SOUND_NAME(RDDRWNCE) VCMI_SOUND_FILE(RDDRWNCE.wav) \
 VCMI_SOUND_NAME(REGENER) VCMI_SOUND_FILE(REGENER.wav) \
 VCMI_SOUND_NAME(REMoveOB) VCMI_SOUND_FILE(REMOVEOB.wav) \
-VCMI_SOUND_NAME(RESURECT) VCMI_SOUND_FILE(RESURECT.wav) \
+VCMI_SOUND_NAME(RESURRECT) VCMI_SOUND_FILE(RESURRECT.wav) \
```
dydzio0614, line 845: file name changed here as well

mapeditor/mapsettings/eventsettings.cpp
```diff
@@ -41,8 +41,8 @@ QVariant toVariant(const CMapEvent & event)
 	result["players"] = QVariant::fromValue(event.players);
 	result["humanAffected"] = QVariant::fromValue(event.humanAffected);
 	result["computerAffected"] = QVariant::fromValue(event.computerAffected);
-	result["firstOccurence"] = QVariant::fromValue(event.firstOccurence);
-	result["nextOccurence"] = QVariant::fromValue(event.nextOccurence);
+	result["firstOccurrence"] = QVariant::fromValue(event.firstOccurrence);
```
dydzio0614, line 44: does this have chance to break compatibility with existing maps?

launcher/modManager/cmodmanager.cpp
```diff
@@ -112,7 +112,7 @@ void CModManager::loadMods()
 			auto json = JsonUtils::toJson(mod);
 			json["localSizeBytes"].Float() = total;
 			if(!name.is_absolute())
-				json["storedLocaly"].Bool() = true;
+				json["storedLocally"].Bool() = true;
```
Alexander-Wilms, line 115: Would this cause issues with existing JSON files?

lib/CSoundBase.h
Alexander-Wilms: Does the change to `VCMI_SOUND_NAME` have to be reverted as well?

mapeditor/inspector/inspector.cpp
```diff
@@ -464,7 +464,7 @@ void Inspector::updateProperties()
 		return;
 	table->setRowCount(0); //cleanup table
 	
-	addProperty("Indentifier", obj);
+	addProperty("Identifier", obj);
```
Alexander-Wilms, line 467: Would this be problematic?

IvanSavenko: This should be safe. Seems to be only used inside of map editor

launcher/modManager/cmodmanager.cpp
IvanSavenko: Safe. Unused code actually. Property is never read.

mapeditor/mapsettings/eventsettings.cpp
IvanSavenko: Somewhat. Maps would still load, but events may not be configured correctly (assuming that we have maps in vcmi format with daily events)

lib/CSoundBase.h
IvanSavenko: Yes. 
These values refer to names of files inside H3 sound archive.
To be honest we can remove most of these values - they are kind of deprecated, with most of them now migrated to config.

lib/CGameInfoCallback.cpp
Alexander-Wilms: Fixed in https://github.com/vcmi/vcmi/pull/4195

PR #4186, opened 2024-06-24
End of battle artifacts transfer

SoundSSGood:
- The order of the infowindows after the battle has been corrected. First Level Up. Then window with spells. Then window with loot (artifacts). Artifacts assembly requests will be at the end, which ensures that the request queue will not be interrupted by other windows. (Multiple levels up still doesn't work as expected.)
- Transferring artifacts after battle now uses BulkMoveArtifacts.  MoveArtifact has been removed.
- After battle, as before, artifacts that end up in the backpack are not asked about the possibility of assembly. For equipped ones, the request to assemble an artifact will not be repeated for the same type of artifact.
- Created the ArtifactsUIController class as part of the CPlayerInterface. All code related to artifact UI has been moved. It might make sense to put ArtifactsUIController in separate folder, but, I’m not sure about plans to split and structurize CPlayerInterface.
- Fixed minor bugs, such as a warning in the console after swapping artifacts.


SoundSSGood:
Is sonarcloud no more works?

IvanSavenko:
> Is sonarcloud no more works?

Looks like it. Possibly some issue on Sonar side. Let's wait for now I guess? Might come back online on its own.

SoundSSGood:
> Looks like it.

Looks like develop branch only. It works on beta.

lib/registerTypes/RegisterTypesClientPacks.h
```diff
@@ -112,7 +112,6 @@ void registerTypesClientPacks(Serializer &s)
 	s.template registerType<CPackForClient, CArtifactOperationPack>();
 	s.template registerType<CArtifactOperationPack, PutArtifact>();
 	s.template registerType<CArtifactOperationPack, EraseArtifact>();
-	s.template registerType<CArtifactOperationPack, MoveArtifact>();
```
IvanSavenko, line 115: This might break saves. Can you check whether saves made before this PR can still be loaded in this PR?

Or just add dummy pack with comment / static_assert, e.g. 
```cpp
static_assert(ESerializationVersion::RELEASE_150 < ESerializationVersion::CURRENT, "Please remove this compatibility check once it no longer needed");
```

client/CPlayerInterface.h
```diff
@@ -56,11 +56,9 @@ namespace boost
 }
 
 /// Central class for managing user interface logic
-class CPlayerInterface : public CGameInterface, public IUpdateable
+class CPlayerInterface : public CGameInterface, public IUpdateable, public ArtifactsUIController
```
IvanSavenko, line 59: Composition is generally better than inheritance.

Personally I would have used `std::unique_ptr<ArtifactsUIController> artifactController` member instead.

client/ArtifactsUIController.h
```diff
@@ -0,0 +1,41 @@
...
+VCMI_LIB_NAMESPACE_END
+
+class ArtifactsUIController
+{
+public:
+	size_t numOfMovedArts;
+	size_t numOfArtsAskAssembleSession;
+	std::set<ArtifactID> ignoredArtifacts;
+
+	boost::mutex askAssembleArtifactMutex;
```
IvanSavenko, line 28: Do we need mutex? All UI-related events are already protected by UI mutex

client/ArtifactsUIController.h
```diff
@@ -0,0 +1,41 @@
...
+class CGHeroInstance;
+
+VCMI_LIB_NAMESPACE_END
+
+class ArtifactsUIController
+{
+public:
+	size_t numOfMovedArts;
+	size_t numOfArtsAskAssembleSession;
+	std::set<ArtifactID> ignoredArtifacts;
```
IvanSavenko, line 26: Make private?

SoundSSGood:
GH.interfaceMutex is unlocked after each dialogue. Which means that the dialogue of other artifacts may be shown mixed. Using a mutex was the simplest solution (maybe not the best). Because there is no need to implement custom waiting procedures.

client/ArtifactsUIController.h
SoundSSGood: Using a mutex was the simplest solution (maybe not the best). Because there is no need to implement custom waiting procedures.

lib/registerTypes/RegisterTypesClientPacks.h
SoundSSGood: > Can you check whether saves made before this PR can still be loaded in this PR?

Yep Can be loaded.
As I see MoveArtifact was used only for network packets

Issue #4206, opened 2024-06-29
Graphics improvement


kdmcser:
**Describe your proposal**

There are some graphics problems need to be improved.

1. Town select and hero select

![1](https://github.com/vcmi/vcmi/assets/8289682/5cda0be6-9c76-46b4-98fb-f7dce29b0c3d)
![2](https://github.com/vcmi/vcmi/assets/8289682/6343f71a-2e72-41bd-8cae-2982ab7b1e81)

2. Hero invitation in tavern

![3](https://github.com/vcmi/vcmi/assets/8289682/31880ec5-dd00-4d8f-89f7-0251e0089f33)

3. Can not scroll long texts in right click (artifact, command skills)

![4](https://github.com/vcmi/vcmi/assets/8289682/91178e64-13d8-44e3-b884-16c24991212c)
![5](https://github.com/vcmi/vcmi/assets/8289682/f797045f-f4c4-4566-afe7-43781924d42b)

Laserlicht:
>     1. Town select and hero select

hero select is kind of intended, that there are many empty fields -> if your host change town while window is open (guest), windows size shouldn't changed. But yeah, more squarish would be good...

by003:
I don't think player should use such many mod towns in the game.
If you are Chinese, you can play my Expansion mod, just find it in bilibili.

kdmcser:
> I don't think player should use such many mod towns in the game. If you are Chinese, you can play my Expansion mod, just find it in bilibili.

your Expansion mod is a quite high quality mod. All towns are well designed, and art designing is quite amazing. I think all heroes3 player should have a try.  But I am a content designer more than a player, I want to build my own vcmi version. So I have to play more mods. ^_^

Laserlicht:
Should be fixed

PR #4210, opened 2024-06-29
[1.6] Graphics improvements

Laserlicht:
partitially fixing #4206

introduce scrollbars for (if necessary):
- Town select and hero select
- hero invite

remove unused lines form window

fix multiplayer (guest has opened hero selection; host change to random town) -> now closes window

### DIFF
--- a/client/lobby/OptionsTab.cpp
+++ b/client/lobby/OptionsTab.cpp
@@ -452,7 +452,7 @@ int OptionsTab::SelectionWindow::calcLines(FactionID faction)
 			count++;
 	}
 
-	return std::ceil(std::max((double)count + additionalItems, (double)allowedFactions.size() + additionalItems) / (double)elementsPerLine);
+	return std::ceil(((double)count + additionalItems) / (double)elementsPerLine);
 }
 
 void OptionsTab::SelectionWindow::apply()
@@ -492,6 +492,8 @@ void OptionsTab::SelectionWindow::recreate()
 {
 	OBJ_CONSTRUCTION_CAPTURING_ALL_NO_DISPOSE;
 
+	sliderLine = slider ? slider->getValue() : 0;
+
 	int amountLines = 1;
 	if(type == SelType::BONUS)
 		elementsPerLine = allowedBonus.size();
@@ -518,9 +520,9 @@ void OptionsTab::SelectionWindow::recreate()
 	}
 
 	int x = (elementsPerLine) * (ICON_BIG_WIDTH-1);
-	int y = (amountLines) * (ICON_BIG_HEIGHT-1);
+	int y = (std::min(amountLines, MAX_LINES)) * (ICON_BIG_HEIGHT-1);
 
-	pos = Rect(0, 0, x, y);
+	pos = Rect(0, 0, x + ((amountLines > MAX_LINES) ? 16 : 0), y);
 
 	backgroundTexture = std::make_shared<FilledTexturePlayerColored>(ImagePath::builtin("DiBoxBck"), pos);
 	backgroundTexture->playerColored(PlayerColor(1));
@@ -532,7 +534,15 @@ void OptionsTab::SelectionWindow::recreate()
 		genContentHeroes();
 	if(type == SelType::BONUS)
 		genContentBonus();
-	genContentGrid(amountLines);
+	genContentGrid(std::min(amountLines, MAX_LINES));
+
+	if(amountLines > MAX_LINES)
+	{
+		slider = std::make_shared<CSlider>(Point(x, 0), y, std::bind(&OptionsTab::SelectionWindow::sliderMove, this, _1), MAX_LINES, amountLines, 0, Orientation::VERTICAL, CSlider::BLUE);
+		slider->setPanningStep(ICON_BIG_HEIGHT);
+		slider->setScrollBounds(Rect(0, 0, x, y));
+		slider->scrollTo(sliderLine);
+	}
 
 	center();
 }
@@ -741,6 +751,14 @@ void OptionsTab::SelectionWindow::setElement(int elem, bool doApply)
 		apply();
 }
 
+void OptionsTab::SelectionWindow::sliderMove(int slidPos)
+{
+	if(!slider)
+		return; // ignore spurious call when slider is being created
+	recreate();
+	redraw();
+}
+
 bool OptionsTab::SelectionWindow::receiveEvent(const Point & position, int eventType) const
 {
 	return true;  // capture click also outside of window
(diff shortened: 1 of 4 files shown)


client/lobby/OptionsTab.h
```diff
@@ -105,8 +106,14 @@ class OptionsTab : public OptionsTabBase
 		const int TEXT_POS_X = 29;
 		const int TEXT_POS_Y = 56;
 
+		const int MAX_LINES = 6;
+		const int MAX_ELEM_PER_LINES = 4;
+
 		int elementsPerLine;
 
+		std::shared_ptr<CSlider> slider;
+		int sliderLine;
```
IvanSavenko, line 115: Probably needs better name?
Also - why can't you acquire this information from slider?

If this is because slider may not exists then add method that would either return value from slider or 0 if there is no slider and remove this member entirely

client/lobby/OptionsTab.cpp
```diff
@@ -740,13 +746,25 @@ void OptionsTab::SelectionWindow::setElement(int elem, bool doApply)
 		apply();
 }
 
+void OptionsTab::SelectionWindow::sliderMove(int slidPos)
+{
+	if(!slider)
+		return; // ignore spurious call when slider is being created
+	if(slidPos != sliderLine)
+		recreate();
+	redraw();
```
IvanSavenko, line 755: Slider can emit calls even if position is still the same?

client/lobby/OptionsTab.cpp
```diff
@@ -658,7 +664,7 @@ void OptionsTab::SelectionWindow::genContentBonus()
 int OptionsTab::SelectionWindow::getElement(const Point & cursorPosition)
 {
 	int x = (cursorPosition.x - pos.x) / (ICON_BIG_WIDTH-1);
-	int y = (cursorPosition.y - pos.y) / (ICON_BIG_HEIGHT-1);
+	int y = ((cursorPosition.y - pos.y) / (ICON_BIG_HEIGHT-1)) + sliderLine;
```
IvanSavenko, line 667: Unnecessary
```suggestion
	int y = (cursorPosition.y - pos.y) / (ICON_BIG_HEIGHT-1) + sliderLine;
```

client/lobby/OptionsTab.cpp
```diff
@@ -609,21 +612,24 @@ void OptionsTab::SelectionWindow::genContentHeroes()
...
 
-			components.push_back(std::make_shared<CAnimImage>(helper.getImageName(true), helper.getImageIndex(true), 0, x * (ICON_BIG_WIDTH-1), y * (ICON_BIG_HEIGHT-1)));
-			drawOutlinedText(x * (ICON_BIG_WIDTH-1) + TEXT_POS_X, y * (ICON_BIG_HEIGHT-1) + TEXT_POS_Y, (selectedHero == elem) ? Colors::YELLOW : Colors::WHITE, helper.getName());
-			ImagePath image = ImagePath::builtin("lobby/townBorderBig");
-			if(selectedHero == elem)
-				image = ImagePath::builtin("lobby/townBorderBigActivated");
-			if(unusableHeroes.count(elem))
-				image = ImagePath::builtin("lobby/townBorderBigGrayedOut");
-			components.push_back(std::make_shared<CPicture>(image, x * (ICON_BIG_WIDTH-1), y * (ICON_BIG_HEIGHT-1)));
+			if(y >= 0 && y <= MAX_LINES - 1)
```
IvanSavenko, line 622: Perhaps use early return (or `continue` since this is a loop) instead of nesting multiple if's?

Same for pre-existing `type->heroClass->faction == selectedFaction` check above

client/lobby/OptionsTab.cpp
```diff
@@ -569,19 +567,23 @@ void OptionsTab::SelectionWindow::genContentFactions()
...
 
 		PlayerSettings set = PlayerSettings();
 		set.castle = elem;
 
 		CPlayerSettingsHelper helper = CPlayerSettingsHelper(set, SelType::TOWN);
 
-		components.push_back(std::make_shared<CAnimImage>(helper.getImageName(true), helper.getImageIndex(true), 0, x * (ICON_BIG_WIDTH-1), y * (ICON_BIG_HEIGHT-1)));
-		components.push_back(std::make_shared<CPicture>(ImagePath::builtin(selectedFaction == elem ? "lobby/townBorderBigActivated" : "lobby/townBorderBig"), x * (ICON_BIG_WIDTH-1), y * (ICON_BIG_HEIGHT-1)));
-		drawOutlinedText(x * (ICON_BIG_WIDTH-1) + TEXT_POS_X, y * (ICON_BIG_HEIGHT-1) + TEXT_POS_Y, (selectedFaction == elem) ? Colors::YELLOW : Colors::WHITE, helper.getName());
+		if(y >= 0 && y <= MAX_LINES - 1)
```
IvanSavenko, line 581: Use continue instead of deep nesting?

IvanSavenko (commented):
Probably target 1.6/develop to be safe.

IvanSavenko:
> fix multiplayer (guest has opened hero selection; host change to random town) -> now closes window

Huh, interesting find.

Can you:
- check whether server correctly checks player input for validness? So even if client sends incorrect data (e.g. without fixed UI) server would reject it and not crash or anything like it?
- check same for faction selection? What if host moves player to another color while that player has town selection dialog?

Also - these fixes can target beta. We actually have some rare crashes related to faction/hero selection, could be caused by this.

Laserlicht:
> > fix multiplayer (guest has opened hero selection; host change to random town) -> now closes window
> 
> Huh, interesting find.

Can make a separate PR with this patch for beta.

But it's not crash the application. Only corrupted graphics (window with 0px size, because there are no heroes for random).

I don't think the crash is caused by this. Do you have more informations about the crash?

IvanSavenko:
> I don't think the crash is caused by this. Do you have more informations about the crash?

Sorry, missed this comment.

Not much information unfortunately. Sometimes I can find a crash that happened because server attempted to change starting town for non-existing player (due to request from client). However no idea about actual crash conditions.

My best guess based on this PR was that (for example) host moves player to another color (or even to observer) while guest has hero/town selector open. And if guest attempts to change starting town for a unexpected player this may lead to some weird cases.

Laserlicht:
Have not found any way to trigger a crash. If I change the slot with opened menu, than the menu will be closed...

IvanSavenko:
Understood. 
Well, could be unrelated. No idea what's causing it. 

In any case - not a big deal, this is rather rare crash, perhaps somebody will report it with steps to reproduce.

PR #4212, opened 2024-06-30
Save all human player names in the configuration

alkazandzhiev:
Add up to 8 human player names to the configuration schema and save/load them to avoid entering them each time.

### DIFF
--- a/client/mainmenu/CMainMenu.cpp
+++ b/client/mainmenu/CMainMenu.cpp
@@ -485,9 +485,9 @@ const std::vector<std::string> CMultiMode::getPlayersNames()
 		playerNameStr = CGI->generaltexth->translate("core.genrltxt.434");
 	playerNames.push_back(playerNameStr);
 
-	for (JsonNode playerName : settings["general"]["multiPlayerNames"].Vector())
+	for (const auto & playerName : settings["general"]["multiPlayerNames"].Vector())
 	{
-		std::string &nameStr = playerName.String();
+		const std::string &nameStr = playerName.String();
 		if (!nameStr.empty())
 		{
 			playerNames.push_back(nameStr);
@@ -549,10 +549,10 @@ void CMultiPlayers::enterSelectionScreen()
 	Settings playerName = settings.write["general"]["playerName"];
 	Settings multiPlayerNames = settings.write["general"]["multiPlayerNames"];
 	multiPlayerNames->Vector().clear();
-	if (playerNames.size() > 0)
+	if (!playerNames.empty())
 	{
 		playerName->String() = playerNames.front();
-		for (std::vector<std::string>::iterator playerNameIt = playerNames.begin()+1; playerNameIt != playerNames.end(); playerNameIt++)
+		for (auto playerNameIt = playerNames.begin()+1; playerNameIt != playerNames.end(); playerNameIt++)
 		{
 			multiPlayerNames->Vector().push_back(JsonNode(*playerNameIt));
 		}
--- a/client/mainmenu/CMainMenu.cpp
+++ b/client/mainmenu/CMainMenu.cpp
@@ -444,7 +444,7 @@ CMultiMode::CMultiMode(ESelectionScreen ScreenType)
 
 	statusBar = CGStatusBar::create(std::make_shared<CPicture>(background->getSurface(), Rect(7, 465, 440, 18), 7, 465));
 	playerName = std::make_shared<CTextInput>(Rect(19, 436, 334, 16), background->getSurface());
-	playerName->setText(getPlayerName());
+	playerName->setText(getPlayersNames()[0]);
 	playerName->setCallback(std::bind(&CMultiMode::onNameChange, this, _1));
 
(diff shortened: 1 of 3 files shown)


config/schemas/settings.json
```diff
@@ -48,6 +55,34 @@
...
+					"default" : ""
+				},
+				"playerName7" : {
+					"type" : "string",
+					"default" : ""
+				},
+				"playerName8" : {
+					"type" : "string",
+					"default" : ""
+				},
```
IvanSavenko, line 85: Why not use array instead?

client/mainmenu/CMainMenu.cpp
```diff
@@ -530,8 +550,17 @@ void CMultiPlayers::enterSelectionScreen()
 			names.push_back(name->getText());
 	}
 
-	Settings name = settings.write["general"]["playerName"];
-	name->String() = names[0];
+	for (int i = 0; i < 8; i++)
```
IvanSavenko, line 553: There is PlayerColor::PLAYER_LIMIT_I instead of hardcoded '8'

client/mainmenu/CMainMenu.cpp
```diff
@@ -466,22 +468,37 @@ void CMultiMode::hostTCP()
...
 {
+	playerNames.clear();
+
 	std::string name = settings["general"]["playerName"].String();
 	if(name == "Player")
 		name = CGI->generaltexth->translate("core.genrltxt.434");
-	return name;
+	playerNames.push_back(name);
+
+	for (int i = 2; i < 9; i++)
```
IvanSavenko, line 494: From 2 to 9? Why?

client/mainmenu/CMainMenu.cpp
```diff
@@ -466,22 +468,37 @@ void CMultiMode::hostTCP()
...
 	auto savedScreenType = screenType;
 	close();
-	GH.windows().createAndPushWindow<CMultiPlayers>(getPlayerName(), savedScreenType, false, ELoadMode::MULTI);
+	std::vector<std::string> playerNames;
+	getPlayersNames(playerNames);
+	GH.windows().createAndPushWindow<CMultiPlayers>(playerNames, savedScreenType, false, ELoadMode::MULTI);
 }
 
-std::string CMultiMode::getPlayerName()
+void CMultiMode::getPlayersNames(std::vector<std::string> & playerNames)
```
IvanSavenko, line 485: Why pass vector as parameter when it can be returned?
```suggestion
std::vector<std::string> CMultiMode::getPlayersNames()
```

config/schemas/settings.json
alkazandzhiev: Could be an array for all the names including the first player's one? The drawback would be that configs would become incompatible with older versions.

IvanSavenko: Option a - use "hotseatNames" that would include only names for non-first players. Probably easiest one.
Option b - import old data in some way. For example, add property with different name, e.g. "playerNames" and if first name empty - set it to current value of old "playerName" property

alkazandzhiev: Yup I was thinking about the second one. I'll play with it, thanks!

client/mainmenu/CMainMenu.cpp
```diff
@@ -466,22 +466,35 @@ void CMultiMode::hostTCP()
...
-		name = CGI->generaltexth->translate("core.genrltxt.434");
-	return name;
+	std::vector<std::string> playerNames;
+
+	std::string playerNameStr = settings["general"]["playerName"].String();
+	if (playerNameStr == "Player")
+		playerNameStr = CGI->generaltexth->translate("core.genrltxt.434");
+	playerNames.push_back(playerNameStr);
+
+	for (JsonNode playerName : settings["general"]["multiPlayerNames"].Vector())
```
IvanSavenko, line 488:
```suggestion
	for (const auto & playerName : settings["general"]["multiPlayerNames"].Vector())
```

client/mainmenu/CMainMenu.cpp
```diff
@@ -523,17 +539,34 @@ void CMultiPlayers::onChange(std::string newText)
...
+		if (playerName->getText().length())
+			playerNames.push_back(playerName->getText());
 	}
 
-	Settings name = settings.write["general"]["playerName"];
-	name->String() = names[0];
+	Settings playerName = settings.write["general"]["playerName"];
+	Settings multiPlayerNames = settings.write["general"]["multiPlayerNames"];
+	multiPlayerNames->Vector().clear();
+	if (playerNames.size() > 0)
```
IvanSavenko, line 552:
```suggestion
	if (!playerNames.empty())
```

client/mainmenu/CMainMenu.cpp
```diff
@@ -523,17 +539,34 @@ void CMultiPlayers::onChange(std::string newText)
...
+	{
+		playerName->String() = playerNames.front();
+		for (std::vector<std::string>::iterator playerNameIt = playerNames.begin()+1; playerNameIt != playerNames.end(); playerNameIt++)
```
IvanSavenko, line 555:
```suggestion
		for (auto playerNameIt = playerNames.begin()+1; playerNameIt != playerNames.end(); playerNameIt++)
```

alkazandzhiev:
Thanks for the pointers!

PR #4219, opened 2024-07-03
Add TBB in conanfile and CMakeLists

smanolloff:
Fixes cmake error:

```
CMake Error at AI/BattleAI/CMakeLists.txt:40 (target_link_libraries):
  Target "BattleAI" links to:

    TBB::tbb

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```


AI/BattleAI/CMakeLists.txt
```diff
@@ -24,6 +24,8 @@ set(battleAI_HEADERS
 		BattleExchangeVariant.h
 )
 
+find_package(TBB REQUIRED)
```
kambala-decapitator, line 27: it must be fixed here instead: https://github.com/vcmi/vcmi/blob/1b10aaa493e54947d6829eefd1673d462f5ec406/CMakeLists.txt#L522-L523

conanfile.py
```diff
@@ -19,9 +19,9 @@ class VCMI(ConanFile):
         "sdl_image/[~2.0.5]",
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
+        "onetbb/[^2021.3]",
     ]
     _nullkillerAIRequires = [
-        "onetbb/[^2021.3]",
```
kambala-decapitator, line 24: _nullkillerAIRequires can now be removed completely

kambala-decapitator (changes requested):
please target `beta` branch instead

kambala-decapitator (approved):

Issue #4222, opened 2024-07-04
Map Editor: Need to set original artifacts for hero, and spells for town


by003:
**Describe the bug**
Now I cannot set any original artifact for hero, and cannot set allowed spells in mage guild.
Town time event is also unavailable.

**Version**
 - OS: Windows
 - Version: 1.5.3 (editor version 0.2)

PR #4225, opened 2024-07-04
Map editor town configuration improvements (buildings, spells, events)

### LINKED ISSUE #3631: Automatically check/uncheck previous/next building in Building Dialog in Map Editor
**Describe your proposal**
In Town Building Dialog,  in Map Editor,
- When you **check** the Magic Guild Level 3 as built, it should **check** also Magic Guild Level 1 and Magic Guild Level 2.
- When you **uncheck** the Magic Guild Level 3 as built, it should **uncheck** also Magic Guild Level 5 and Magic Guild Level 4.
- When you check the Magic Guild Level 3 as **buildable**, it should check also as buildable Magic Guild Level 1 and Magic Guild Level 2.
- When you uncheck the Magic Guild Level 3 as **buildable**, it should uncheck also as buildable Magic Guild Level 5 and Magic Guild Level 4.

**How to reproduce**

1. Open _Map Editor_
2. Add a town
3. Go on _Inspector_
4. Click on _Building..._

godric3:
~Started some time ago, I want to add town events before merging. PR created for visibility(and maybe better motivation to finish...)~
Improvements for town customization to have same possibilities as original editor. Unfortunately quite a lot of code duplication from map events and other town widgets as they have almost the same content
- [x]  auto update base/upgrade building fixes #3631 :
  - if building is built it have to be enabled
  - if building is built/enabled base building must be built/enabled
  - if building is disabled it cannot be build
  - if building is disabled/unbuilt upgrade building must be disabled/unbuilt
- [x]  build/demolish/enable/disable all building option
- [x] set town spells (possible and required) part of #4222
- [x] set town timed events part of #4222 
- [x] translations

image

image

image

### DIFF
--- a/lib/gameState/CGameState.cpp
+++ b/lib/gameState/CGameState.cpp
@@ -871,7 +871,7 @@ void CGameState::initTowns()
 		}
 		//init spells
 		vti->spells.resize(GameConstants::SPELL_LEVELS);
-
+		vti->possibleSpells -= SpellID::PRESET;
 		for(ui32 z=0; z<vti->obligatorySpells.size();z++)
 		{
 			const auto * s = vti->obligatorySpells[z].toSpell();
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -1207,22 +1207,12 @@ void CGTownInstance::serializeJsonOptions(JsonSerializeFormat & handler)
 	{
 		handler.serializeIdArray( "possibleSpells", possibleSpells);
 		handler.serializeIdArray( "obligatorySpells", obligatorySpells);
+	}
 
-		if (handler.saving) 
-		{
-			auto eventsHandler = handler.enterArray("events");
-			std::vector<CCastleEvent> temp(events.begin(), events.end());
-			eventsHandler.serializeStruct(temp);
-		}
-		else
-		{
-			auto eventsHandler = handler.enterArray("events");
-			std::vector<CCastleEvent> temp;
-			eventsHandler.serializeStruct(temp);
-			events.clear();
-			events.insert(events.begin(), temp.begin(), temp.end());
-		}
-		
+	{
+		auto eventsHandler = handler.enterArray("events");
+		eventsHandler.syncSize(events, JsonNode::JsonType::DATA_VECTOR);
+		eventsHandler.serializeStruct(events);
 	}
 }
 
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -1207,6 +1207,22 @@ void CGTownInstance::serializeJsonOptions(JsonSerializeFormat & handler)
 	{
 		handler.serializeIdArray( "possibleSpells", possibleSpells);
 		handler.serializeIdArray( "obligatorySpells", obligatorySpells);
+
+		if (handler.saving) 
+		{
+			auto eventsHandler = handler.enterArray("events");
+			std::vector<CCastleEvent> temp(events.begin(), events.end());
+			eventsHandler.serializeStruct(temp);
+		}
+		else
+		{
+			auto eventsHandler = handler.enterArray("events");
+			std::vector<CCastleEvent> temp;
+			eventsHandler.serializeStruct(temp);
+			events.clear();
+			events.insert(events.begin(), temp.begin(), temp.end());
+		}
+		
 	}
 }
 
--- a/mapeditor/inspector/townbuildingswidget.cpp
+++ b/mapeditor/inspector/townbuildingswidget.cpp
@@ -93,9 +93,9 @@ QStandardItem * getBuildingParentFromTreeModel(const CBuilding * building, QStan
 	return parent;
 }
 
-std::set<QVariant> getBuildingVariantsFromModel(QStandardItemModel & model, int modelColumn, Qt::CheckState checkState)
+std::list<QVariant> getBuildingVariantsFromModel(QStandardItemModel & model, int modelColumn, Qt::CheckState checkState)
 {
-	std::set<QVariant> result;
+	std::list<QVariant> result;
 	std::vector<QModelIndex> stack(1);
 	while (!stack.empty())
 	{
@@ -107,7 +107,7 @@ std::set<QVariant> getBuildingVariantsFromModel(QStandardItemModel & model, int 
 			QModelIndex index = model.index(i, modelColumn, pindex);
 			if (auto * item = model.itemFromIndex(index))
 				if (item->checkState() == checkState)
-					result.emplace(item->data(MapEditorRoles::BuildingIDRole));
+					result.push_back(item->data(MapEditorRoles::BuildingIDRole));
 			index = model.index(i, 0, pindex);
 			if (model.hasChildren(index))
 				stack.push_back(index);
--- a/mapeditor/inspector/townbuildingswidget.cpp
+++ b/mapeditor/inspector/townbuildingswidget.cpp
@@ -73,7 +73,7 @@ QStandardItem * getBuildingParentFromTreeModel(const CBuilding * building, QStan
 {
 	QStandardItem * parent = nullptr;
 	std::vector<QModelIndex> stack(1);
-	while (!parent && !stack.empty())
+	do
 	{
 		auto pindex = stack.back();
 		stack.pop_back();
@@ -89,15 +89,15 @@ QStandardItem * getBuildingParentFromTreeModel(const CBuilding * building, QStan
 			if (model.hasChildren(index))
 				stack.push_back(index);
 		}
-	}
+	} while(!parent && !stack.empty());
 	return parent;
 }
 
-std::list<QVariant> getBuildingVariantsFromModel(QStandardItemModel & model, int modelColumn, Qt::CheckState checkState)
+QVariantList getBuildingVariantsFromModel(QStandardItemModel & model, int modelColumn, Qt::CheckState checkState)
 {
-	std::list<QVariant> result;
+	QVariantList result;
 	std::vector<QModelIndex> stack(1);
-	while (!stack.empty())
+	do
 	{
 		auto pindex = stack.back();
 		stack.pop_back();
@@ -105,14 +105,14 @@ std::list<QVariant> getBuildingVariantsFromModel(QStandardItemModel & model, int
 		for (int i = 0; i < rowCount; ++i)
 		{
 			QModelIndex index = model.index(i, modelColumn, pindex);
-			if (auto * item = model.itemFromIndex(index))
-				if (item->checkState() == checkState)
-					result.push_back(item->data(MapEditorRoles::BuildingIDRole));
+			auto * item = model.itemFromIndex(index);
+			if(item && item->checkState() == checkState)
+				result.push_back(item->data(MapEditorRoles::BuildingIDRole));
 			index = model.index(i, 0, pindex);
 			if (model.hasChildren(index))
 				stack.push_back(index);
 		}
-	}
+	} while(!stack.empty());
 
 	return result;
 }
@@ -263,7 +263,7 @@ void TownBuildingsWidget::setRowColumnCheckState(QStandardItem * item, Column co
 void TownBuildingsWidget::setAllRowsColumnCheckState(Column column, Qt::CheckState checkState)
 {
 	std::vector<QModelIndex> stack(1);
-	while (!stack.empty())
+	do
 	{
 		auto parentIndex = stack.back();
 		stack.pop_back();
@@ -277,7 +277,7 @@ void TownBuildingsWidget::setAllRowsColumnCheckState(Column column, Qt::CheckSta
 			if (model.hasChildren(index))
 				stack.push_back(index);
 		}
-	}
+	} while(!stack.empty());
 }
 
 void TownBuildingsWidget::onItemChanged(QStandardItem * item) {
@@ -297,7 +297,8 @@ void TownBuildingsWidget::onItemChanged(QStandardItem * item) {
 	else if (item->checkState() == Qt::Unchecked) {
 		std::vector<QStandardItem*> stack;
 		stack.push_back(nextRow);
-		while (!stack.empty()) {
+		do
+		{
 			nextRow = stack.back();
 			stack.pop_back();
 			setRowColumnCheckState(nextRow, Column(item->column()), Qt::Unchecked);
@@ -310,7 +311,7 @@ void TownBuildingsWidget::onItemChanged(QStandardItem * item) {
 				}
 			}
 			
-		}
+		} while(!stack.empty());
 	}
 	connect(&model, &QStandardItemModel::itemChanged, this, &TownBuildingsWidget::onItemChanged);
 }
--- a/mapeditor/inspector/townbuildingswidget.cpp
+++ b/mapeditor/inspector/townbuildingswidget.cpp
@@ -76,7 +76,7 @@ TownBuildingsWidget::TownBuildingsWidget(CGTownInstance & t, QWidget *parent) :
 	ui->setupUi(this);
 	ui->treeView->setModel(&model);
 	//ui->treeView->setColumnCount(3);
-	model.setHorizontalHeaderLabels(QStringList() << QStringLiteral("Type") << QStringLiteral("Enabled") << QStringLiteral("Built"));
+	model.setHorizontalHeaderLabels(QStringList() << tr("Type") << tr("Enabled") << tr("Built"));
 	connect(&model, &QStandardItemModel::itemChanged, this, &TownBuildingsWidget::onItemChanged);
 	//setAttribute(Qt::WA_DeleteOnClose);
 }
--- a/mapeditor/inspector/townbuildingswidget.cpp
+++ b/mapeditor/inspector/townbuildingswidget.cpp
@@ -77,7 +77,7 @@ TownBuildingsWidget::TownBuildingsWidget(CGTownInstance & t, QWidget *parent) :
 	ui->treeView->setModel(&model);
 	//ui->treeView->setColumnCount(3);
 	model.setHorizontalHeaderLabels(QStringList() << QStringLiteral("Type") << QStringLiteral("Enabled") << QStringLiteral("Built"));
-	
+	connect(&model, &QStandardItemModel::itemChanged, this, &TownBuildingsWidget::onItemChanged);
 	//setAttribute(Qt::WA_DeleteOnClose);
 }
 
@@ -196,12 +196,12 @@ std::set<BuildingID> TownBuildingsWidget::getBuildingsFromModel(int modelColumn,
 
 std::set<BuildingID> TownBuildingsWidget::getForbiddenBuildings()
 {
-	return getBuildingsFromModel(1, Qt::Unchecked);
+	return getBuildingsFromModel(Column::ENABLED, Qt::Unchecked);
 }
 
 std::set<BuildingID> TownBuildingsWidget::getBuiltBuildings()
 {
-	return getBuildingsFromModel(2, Qt::Checked);
+	return getBuildingsFromModel(Column::BUILT, Qt::Checked);
 }
 
 void TownBuildingsWidget::on_treeView_expanded(const QModelIndex &index)
@@ -214,6 +214,37 @@ void TownBuildingsWidget::on_treeView_collapsed(const QModelIndex &index)
 	ui->treeView->resizeColumnToContents(0);
 }
 
+
+void TownBuildingsWidget::setRowColumnCheckState(QStandardItem * item, Column column, Qt::CheckState checkState) {
+	auto sibling = item->model()->sibling(item->row(), column, item->index());
+	model.itemFromIndex(sibling)->setCheckState(checkState);
+}
+
+void TownBuildingsWidget::onItemChanged(QStandardItem * item) {
+	disconnect(&model, &QStandardItemModel::itemChanged, this, &TownBuildingsWidget::onItemChanged);
+	auto rowFirstColumnIndex = item->model()->sibling(item->row(), Column::TYPE, item->index());
+	QStandardItem * nextRow = model.itemFromIndex(rowFirstColumnIndex);
+	if (item->checkState() == Qt::Checked) {
+		while (nextRow) {
+			setRowColumnCheckState(nextRow, Column(item->column()), Qt::Checked);
+			if (item->column() == Column::BUILT) {
+				setRowColumnCheckState(nextRow, Column::ENABLED, Qt::Checked);
+			}
+			nextRow = nextRow->parent();
+
+		}
+	}
+	else if (item->checkState() == Qt::Unchecked) {
+		while (nextRow) {
+			setRowColumnCheckState(nextRow, Column(item->column()), Qt::Unchecked);
+			if (item->column() == Column::ENABLED) {
+				setRowColumnCheckState(nextRow, Column::BUILT, Qt::Unchecked);
+			}
+			nextRow = nextRow->child(0, Column::TYPE);
+		}
+	}
+	connect(&model, &QStandardItemModel::itemChanged, this, &TownBuildingsWidget::onItemChanged);
+}
 
 TownBuildingsDelegate::TownBuildingsDelegate(CGTownInstance & t): town(t), QStyledItemDelegate()
 {
--- a/mapeditor/inspector/townbuildingswidget.cpp
+++ b/mapeditor/inspector/townbuildingswidget.cpp
@@ -68,6 +68,56 @@ std::string defaultBuildingIdConversion(BuildingID bId)
 	}
 }
 
+QStandardItem * getBuildingParentFromTreeModel(const CBuilding * building, QStandardItemModel & model)
+{
+	QStandardItem * parent = nullptr;
+	std::vector<QModelIndex> stack(1);
+	while (!parent && !stack.empty())
+	{
+		auto pindex = stack.back();
+		stack.pop_back();
+		auto rowCount = model.rowCount(pindex);
+		for (int i = 0; i < rowCount; ++i)
+		{
+			QModelIndex index = model.index(i, 0, pindex);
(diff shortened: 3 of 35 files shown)


lib/mapObjects/CGTownInstance.cpp
```diff
@@ -1206,6 +1206,22 @@ void CGTownInstance::serializeJsonOptions(JsonSerializeFormat & handler)
...
+			eventsHandler.serializeStruct(temp);
+		}
+		else
+		{
+			auto eventsHandler = handler.enterArray("events");
+			std::vector<CCastleEvent> temp;
+			eventsHandler.serializeStruct(temp);
+			events.clear();
+			events.insert(events.begin(), temp.begin(), temp.end());
+		}
```
IvanSavenko, line 1223: Completely untested, but will something like this work?

```suggestion
	{
		auto eventsHandler = handler.enterArray("customSize");
		eventsHandler.syncSize(events, JsonNode::JsonType::DATA_VECTOR);
		eventsHandler.serializeStruct(events);
	}
```

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+	auto resourcesMap = params.value("resources").toMap();
+	for (int i = 0; i < GameConstants::RESOURCE_QUANTITY; ++i)
+	{
+		auto name = QString::fromStdString(GameConstants::RESOURCE_NAMES[i]);
+		int val = resourcesMap.value(name).toInt();
+		ui->resourcesTable->setItem(i, 0, new QTableWidgetItem(name));
+
+		QSpinBox * edit = new QSpinBox(ui->resourcesTable);
+		edit->setMaximum(i == GameResID::GOLD ? 999999 : 999);
+		edit->setMinimum(i == GameResID::GOLD ? -999999 : -999);
```
IvanSavenko, line 76: Why limit to 999?

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+}
+
+void TownEvent::setRowColumnCheckState(QStandardItem * item, int column, Qt::CheckState checkState) {
+	auto sibling = item->model()->sibling(item->row(), column, item->index());
+	buildingsModel.itemFromIndex(sibling)->setCheckState(checkState);
+}
+
+void TownEvent::onItemChanged(QStandardItem * item)
+{
+	disconnect(&buildingsModel, &QStandardItemModel::itemChanged, this, &TownEvent::onItemChanged);
```
IvanSavenko, line 294: I think you might want to use QSignalBlocker class here instead

mapeditor/inspector/townspellswidget.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void TownSpellsWidget::resetSpells()
+{
+	town.possibleSpells.clear();
+	town.obligatorySpells.clear();
+	for (auto spell : VLC->spellh->objects)
+	{
+		if (!spell->isSpecial() && !spell->isCreatureAbility())
+			town.possibleSpells.push_back(spell->id);
+	}
```
IvanSavenko, line 57:
```suggestion
	for (auto spellID : VLC->spellh->getDefaultAllowed())
		town.possibleSpells.push_back(spellID);
```

mapeditor/inspector/townspellswidget.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+void TownSpellsWidget::initSpellLists()
+{
+	QListWidget * possibleSpellLists[] = { ui->possibleSpellList1, ui->possibleSpellList2, ui->possibleSpellList3, ui->possibleSpellList4, ui->possibleSpellList5 };
+	QListWidget * requiredSpellLists[] = { ui->requiredSpellList1, ui->requiredSpellList2, ui->requiredSpellList3, ui->requiredSpellList4, ui->requiredSpellList5 };
+	auto spells = VLC->spellh->objects;
```
IvanSavenko, line 64: `VLC->spellh->getDefaultAllowed()` is preferred to direct access to `objects` field.

And to convert SpellID to Spell you can use (preferred) `spellID.toEntity(VLC)` or `spellID.toSpell()` (if you need info not present in entity)

lib/mapObjects/CGTownInstance.cpp
godric3:
```
 no instance of overloaded function matches the argument list
argument types are: (std::list<CCastleEvent, std::allocator<CCastleEvent>>) 
object type is: JsonArraySerializer
 ```
 
 events are list, and serializeStruct expect vector

File: mapeditor/inspector/townevent.cpp (hunk 2 above)
Comment (godric3) [reply mapeditor/inspector/townevent.cpp:76]: couldn't find any RESOURCE_LIMIT const so I used some numbers that other rewards widget is using

File: lib/mapObjects/CGTownInstance.cpp (hunk 1 above)
Comment (IvanSavenko) [reply lib/mapObjects/CGTownInstance.cpp:1223]: Personally I would change list to vector.
Performance hit likely would be negligible (or even would have an improvement)

File: mapeditor/inspector/townevent.cpp (hunk 3 above)
Comment (godric3) [reply mapeditor/inspector/townevent.cpp:294]: tried and works, but have side effect of not updating checkboxes until user hovers over them or move mouse outside of buildings treeview, so I guess I will live it as is. A little bit uglier code for better user experience

**Review by IvanSavenko — APPROVED** [review]
Looks OK for me.
@kambala-decapitator anything to add?

**Comment by kambala-decapitator** [comment]
will have a look today or tomorrow

File: mapeditor/inspector/townevent.h [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+
+#include "../StdInc.h"
+#include <QDialog>
+#include "../lib/mapObjects/CGTownInstance.h"
+
+namespace Ui {
+	class TownEvent;
+}
+
+class TownEvent : public QDialog
```
kambala-decapitator, line 20: TownEventDialog

mapeditor/inspector/towneventswidget.h
```diff
@@ -0,0 +1,59 @@
...
+	TownEventsDelegate(CGTownInstance &, MapController &);
+
+	QWidget* createEditor(QWidget * parent, const QStyleOptionViewItem & option, const QModelIndex & index) const override;
+	void setEditorData(QWidget * editor, const QModelIndex & index) const override;
+	void setModelData(QWidget * editor, QAbstractItemModel * model, const QModelIndex & index) const override;
+
+private:
+	CGTownInstance & town;
+	MapController & controller;
+};
```
kambala-decapitator, line 59: please add empty line at EOF

mapeditor/inspector/towneventswidget.cpp
```diff
@@ -0,0 +1,179 @@
...
+	result.creatures = creaturesFromVariant(v.value("creatures"));
+	return result;
+}
+
+void TownEventsWidget::obtainData()
+{
+	for (const auto & event : town.events)
+	{
+		auto eventName = QString::fromStdString(event.name);
+		auto itemText = QString::fromStdString("Day %1 - %2").arg(event.firstOccurrence+1, 3).arg(eventName);
```
kambala-decapitator: `QString{"Day %1 - %2"}.arg...`

mapeditor/inspector/towneventswidget.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+void TownEventsWidget::on_timedEventRemove_clicked()
+{
+	if (auto* item = ui->eventsList->currentItem())
+		ui->eventsList->takeItem(ui->eventsList->row(item));
+}
+
+void TownEventsWidget::on_eventsList_itemActivated(QListWidgetItem* item)
+{
+	new TownEvent(town, item, parentWidget());
```
kambala-decapitator, line 141: is it supposed to show modal dialog?

mapeditor/inspector/towneventswidget.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+void TownEventsWidget::onItemChanged(QStandardItem * item)
+{
+}
```
kambala-decapitator, line 146: what is this for?

mapeditor/inspector/towneventswidget.cpp
```diff
@@ -0,0 +1,179 @@
...
+	auto* item = new QListWidgetItem(QString::fromStdString(event.name));
+	item->setData(Qt::UserRole, toVariant(event));
+	ui->eventsList->addItem(item);
+	on_eventsList_itemActivated(item);
+}
+
+void TownEventsWidget::on_timedEventRemove_clicked()
+{
+	if (auto* item = ui->eventsList->currentItem())
+		ui->eventsList->takeItem(ui->eventsList->row(item));
```
kambala-decapitator, line 136: `delete ui->eventsList->takeItem(ui->eventsList->currentRow());`

mapeditor/inspector/townevent.h
```diff
@@ -0,0 +1,53 @@
...
+	QVariantList creaturesToVariant();
+
+	QStandardItem * addBuilding(const CTown & ctown, BuildingID bId, std::set<si32> & remaining);
+
+	Ui::TownEvent * ui;
+	CGTownInstance & town;
+	QListWidgetItem * item;
+	QMap<QString, QVariant> params;
+	mutable QStandardItemModel buildingsModel;
+};
```
kambala-decapitator, line 53: please add empty line at EOF

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+	ui->eventAffectsHuman->setChecked(params.value("humanAffected").toBool());
+	ui->eventFirstOccurrence->setValue(params.value("firstOccurrence").toInt()+1);
+	ui->eventRepeatAfter->setValue(params.value("nextOccurrence").toInt());
+
+	initPlayers();
+	initResources();
+	initBuildings();
+	initCreatures();
+
+	show();
```
kambala-decapitator, line 44: this is a wrong approach to work with windows, the calling code is responsible to call show/exec

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+	params = item->data(Qt::UserRole).toMap();
+	ui->eventFirstOccurrence->setMinimum(1);
+	ui->eventFirstOccurrence->setMaximum(999);
+	ui->eventRepeatAfter->setMaximum(999);
+	ui->eventNameText->setText(params.value("name").toString());
+	ui->eventMessageText->setPlainText(params.value("message").toString());
+	ui->eventAffectsCpu->setChecked(params.value("computerAffected").toBool());
+	ui->eventAffectsHuman->setChecked(params.value("humanAffected").toBool());
+	ui->eventFirstOccurrence->setValue(params.value("firstOccurrence").toInt()+1);
```
kambala-decapitator, line 36: why do you need to use +/- 1? can't you save as is?

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+TownEvent::~TownEvent()
+{
+	delete ui;
+}
+
+void TownEvent::initPlayers()
+{
+	for (int i = 0; i < PlayerColor::PLAYER_LIMIT_I; ++i)
+	{
+		bool isAffected = (1 << i) & params.value("players").toInt();
```
kambala-decapitator, line 56: what is this bit magic? can't you store an array?

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+	item(item)
+{
+	ui->setupUi(this);
+
+	ui->buildingsTree->setModel(&buildingsModel);
+
+	params = item->data(Qt::UserRole).toMap();
+	ui->eventFirstOccurrence->setMinimum(1);
+	ui->eventFirstOccurrence->setMaximum(999);
+	ui->eventRepeatAfter->setMaximum(999);
```
kambala-decapitator, line 31: magic numbers should be named constants

mapeditor/inspector/townevent.cpp
kambala-decapitator: magic numbers should be named constants

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+	}
+}
+
+void TownEvent::initBuildings()
+{
+	auto * ctown = town.town;
+	if (!ctown)
+		ctown = VLC->townh->randomTown;
+	if (!ctown)
+		throw std::runtime_error("No Town defined for type selected");
```
kambala-decapitator, line 90: do you really want to crash editor? Haven't seen exception handling where dialog is created

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+	auto bId = buildingId.num;
+	const CBuilding * building = ctown.buildings.at(buildingId);
+	if (!building)
+	{
+		remaining.erase(bId);
+		return nullptr;
+	}
+
+	QString name = tr(building->getNameTranslated().c_str());
```
kambala-decapitator, line 111: `tr` doesn't work with runtime strings

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+				QModelIndex index = buildingsModel.index(i, 0, pindex);
+				if (building->upgrade.getNum() == buildingsModel.itemFromIndex(index)->data(Qt::UserRole).toInt())
+				{
+					parent = buildingsModel.itemFromIndex(index);
+					break;
+				}
+				if (buildingsModel.hasChildren(index))
+					stack.push_back(index);
+			}
+		}
```
kambala-decapitator, line 150: looks quite similar to the other place, can you extract it to a function?

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+	remaining.erase(bId);
+	return checks.front();
+}
+
+void TownEvent::initCreatures()
+{
+	auto creatures = params.value("creatures").toList();
+	auto * ctown = town.town;
+	for (int i = 0; i < 7; ++i)
```
kambala-decapitator, line 172: magic number

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+	{
+		QString creatureNames;
+		if (!ctown)
+		{
+			creatureNames.append(QString("Creature %1 / Creature %1 Upgrade").arg(i + 1));
```
kambala-decapitator: why don't you want to make this string translatable?

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+				auto creatureName = QString::fromStdString(cre->getNameSingularTranslated());
+				creatureNames.append(creatureNames.isEmpty() ? creatureName : " / " + creatureName);
+			}
+		}
+		auto * item = new QTableWidgetItem();
+		item->setFlags(item->flags() & ~Qt::ItemIsEditable);
+		item->setText(creatureNames);
+		ui->creaturesTable->setItem(i, 0, item);
+
+		auto creatureNumber = creatures.size() > i ? creatures.at(i).toInt() : 0;
```
kambala-decapitator, line 194: does it make sense to add spinbox in the false case?

mapeditor/inspector/townevent.cpp
```diff
@@ -0,0 +1,320 @@
...
+		QSpinBox* edit = new QSpinBox(ui->creaturesTable);
+		edit->setValue(creatureNumber);
+		edit->setMaximum(999999);
```
kambala-decapitator, line 197: same magic number as above

kambala-decapitator, line 197: same magic number as above

lib/gameState/CGameState.cpp
```diff
@@ -904,7 +904,7 @@ void CGameState::initTowns()
 		}
 		//init spells
 		vti->spells.resize(GameConstants::SPELL_LEVELS);
-
+		vti->possibleSpells -= SpellID::PRESET;
```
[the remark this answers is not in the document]
IvanSavenko: It is not harmful, but not sure how SpellID::PRESET can even appear in towns.
It is used to indicate starting spells of a hero.

kambala-decapitator (changes requested):
please avoid creating such huge PRs that contain separate unrelated changes

lib/gameState/CGameState.cpp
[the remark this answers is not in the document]
kambala-decapitator: I saw that editor uses it as some kind of placeholder, probably related to that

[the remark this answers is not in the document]
godric3: Yes, I used it also for marking that town possible spells had been customized, so editor can detect that, otherwise I would have to compare town possible spells with allowed spells for differences to know if spells were customized. 
May create separate const specific for towns if that one was meant bo be hero only

mapeditor/inspector/towneventswidget.cpp
godric3: Yes

lib/gameState/CGameState.cpp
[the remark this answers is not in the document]
IvanSavenko: Should be fine. At the moment it is only used for heroes, but using it for towns if needed should also be fine.

Personally I don't like this approach with fake spells, but don't see any clean solution either. And it is not like this weird approach was introduced in this PR.

mapeditor/inspector/townevent.cpp
godric3: Code copied from timedevent.cpp, looks like players in events are stored in bit field
https://github.com/vcmi/vcmi/blob/3c611ffa5b76686156d89aded68e87fc4e34c2b6/lib/mapping/CMapDefines.h#L39

IvanSavenko: Might be a good idea to replace with std::set<PlayerColor>, but not insisting.

mapeditor/inspector/townevent.cpp
godric3: don't really see this triggering without corrupted townh, but anyway I used same approach as in townbuildingswidget

mapeditor/inspector/townevent.cpp
godric3: Already changed, somehow you reviewed older commit?

mapeditor/inspector/towneventswidget.cpp
kambala-decapitator: then the right way is:
```cpp
TownEventDialog dlg{town, item, parentWidget()};
dlg.exec();
```

and drop that `show()` from constructor

mapeditor/inspector/townevent.cpp
kambala-decapitator: ah, since it's coming not from editor settings, then it's fine of course

mapeditor/inspector/townevent.cpp
kambala-decapitator: yeah, was looking commit by commit because it's easier to understand separate changes like that, not all changes at once

mapeditor/inspector/townevent.cpp
godric3: To show user 1 based day numbering, not 0 based as in code, probably could be done better...

mapeditor/inspector/towneventswidget.cpp
godric3: leftover, will delete

mapeditor/inspector/townevent.cpp
kambala-decapitator: are these params read/written from/to map object? Probably I missed that, then it's fine to apply +- 1 for UI. I thought that's some kind of editor-only data.

mapeditor/inspector/townevent.h
```diff
@@ -0,0 +1,53 @@
...
+	QVariantList buildingsToVariant();
+	QVariantList creaturesToVariant();
+
+	QStandardItem * addBuilding(const CTown & ctown, BuildingID bId, std::set<si32> & remaining);
+
+	Ui::TownEvent * ui;
+	CGTownInstance & town;
+	QListWidgetItem * item;
+	QMap<QString, QVariant> params;
+	mutable QStandardItemModel buildingsModel;
```
kambala-decapitator, line 52: why is it marked as mutable?

mapeditor/inspector/townevent.h
```diff
@@ -0,0 +1,53 @@
...
+	QVariant playersToVariant();
+	QVariantMap resourcesToVariant();
+	QVariantList buildingsToVariant();
+	QVariantList creaturesToVariant();
+
+	QStandardItem * addBuilding(const CTown & ctown, BuildingID bId, std::set<si32> & remaining);
+
+	Ui::TownEvent * ui;
+	CGTownInstance & town;
+	QListWidgetItem * item;
```
kambala-decapitator, line 50: please give the variable a better name to make it clear what kind of item it is

godric3:
ok, pipelines are finally happy, ready to review

mapeditor/inspector/townspellswidget.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+	if (auto * ed = qobject_cast<TownSpellsWidget *>(editor))
+	{
+		ed->commitChanges();
+	}
+	else
+	{
+		QStyledItemDelegate::setModelData(editor, model, index);
+	}
+}
```
kambala-decapitator, line 175: please add line break at EOF

mapeditor/mapeditorroles.h
```diff
@@ -0,0 +1,20 @@
...
+#pragma once
+
+#include "StdInc.h"
+
+enum MapEditorRoles
+{
+    TownEventRole = Qt::UserRole +1,
+    PlayerIDRole,
+    BuildingIDRole,
+	SpellIDRole
```
kambala-decapitator, line 19: something's wrong with indentation

mapeditor/mapeditorroles.h
```diff
@@ -0,0 +1,20 @@
...
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#pragma once
+
+#include "StdInc.h"
+
+enum MapEditorRoles
+{
+    TownEventRole = Qt::UserRole +1,
```
kambala-decapitator, line 16:
```suggestion
    TownEventRole = Qt::UserRole + 1,
```

mapeditor/inspector/townbuildingswidget.cpp
```diff
@@ -68,6 +69,56 @@ std::string defaultBuildingIdConversion(BuildingID bId)
...
+			if (model.hasChildren(index))
+				stack.push_back(index);
+		}
+	}
+	return parent;
+}
+
+std::list<QVariant> getBuildingVariantsFromModel(QStandardItemModel & model, int modelColumn, Qt::CheckState checkState)
+{
+	std::list<QVariant> result;
```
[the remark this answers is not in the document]
godric3: list is just for easier creation of QVariantList, and QVariant is because of the same reason. QVariantList is then saved into QVariantMap that is stored as townEventListItem data

IvanSavenko:
And by the way - take a look at stuff found by SonarCloud in this PR:
https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&sinceLeakPeriod=true&pullRequest=4225&id=vcmi_vcmi

Not asking you to fix everything from it, just items that you agree with and makes sense to fix.

mapeditor/inspector/townbuildingswidget.cpp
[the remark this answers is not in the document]
kambala-decapitator: - QVariantList is **not** `std::list` under the hood
- you can create it from any container that has begin/end

use vector if you don't need special behavior of some other container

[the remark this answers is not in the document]
godric3: I know it's not std::list but it have convenient api for conversion from it. I had problems with using set (and I think the same with vector) for construction of QVaraintList. I already cleaned git history but I may double check that in the evening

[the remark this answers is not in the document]
kambala-decapitator: this should work:
```cpp
std::vector<QVariant> vec;
QVariantList list{vec.cbegin(), vec.cend()};
```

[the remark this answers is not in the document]
godric3: even before qt 5.14?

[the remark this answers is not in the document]
IvanSavenko: Why not simply use QVariantList? Mixing std and Qt containers is a bit weird approach IMO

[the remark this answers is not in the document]
godric3: Perhaps that will be better solution, will check it

mapeditor/inspector/towneventdialog.cpp
```diff
@@ -0,0 +1,299 @@
...
+	ui->buildingsTree->resizeColumnToContents(0);
+
+	connect(&buildingsModel, &QStandardItemModel::itemChanged, this, &TownEventDialog::onItemChanged);
+}
+
+QStandardItem * TownEventDialog::addBuilding(const CTown& ctown, BuildingID buildingId, std::set<si32>& remaining)
+{
+	auto bId = buildingId.num;
+	const CBuilding * building = ctown.buildings.at(buildingId);
+	if (!building)
```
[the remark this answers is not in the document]
godric3: `buildingId` here may come only from two places
1.  element from `ctown->getAllBuildings()`
2.  `building->upgrade`

So it looks like `at` will always return valid pointer, will remove null checks

IvanSavenko:
@kambala-decapitator Anything else that you want to address in this PR?

kambala-decapitator (approved):

Fabrice-TIERCELIN:
@godric3, if you are on Windows, you can screenshot a dialog directly by pressing `Alt+Print Screen`.

Issue #3599, opened 2024-02-07
NKAI doesn't move at all on Barren Lands map


DjWarmonger:
**Describe the bug**
AI just passes turns and all enemies lose at day 8.
Could be because the only way out of starting position in Barren Lands is through one-way portal.

**To Reproduce**

- Launch HoTA map "Barren Lands"
- Use `vcmigod` and capture closest town behind the portal
- Use autoskip and just pass turns

**Expected behavior**
AI would play normally and at least try to move outisde starting spot

**Actual behavior**
AI just passes turn and does nothing, losing without town after a week

**Did it work earlier?**
No idea

**Version**
1.4.5

nullkiller:
Yes, one-way monoliths is the thing which is kind of problem. I do not know exact conditions when it is ok to pass through one-way monolith. Sometimes it can act as a trap for a weak hero. So it should be a main hero and either we should be able to go back or the thing on the other side should be very important.

Issue #4216, opened 2024-07-02
AI doesn't work thorough one way monolits


Povelitelh3:
Some time ago I'v seen as AI entry thorough one way monolits.
Now they just ignore them.

behaviour:

vcmi https://youtu.be/yvuBWFpD8h4 
SoD https://youtu.be/egMs6QveUCY
map [HeroMonolits.zip](https://github.com/user-attachments/files/16070014/HeroMonolits.zip)

![изображение](https://github.com/vcmi/vcmi/assets/36762538/9446992a-4f5c-4ce4-a510-06fa3f121111)
![изображение](https://github.com/vcmi/vcmi/assets/36762538/44bdd77c-ee30-4238-b307-8681b623cae6)

1.5.3

DjWarmonger:
Same as #3599

Issue #4104, opened 2024-06-07
NKAI doesn't scouting thorough whirlpool


Povelitelh3:
![изображение](https://github.com/vcmi/vcmi/assets/36762538/5ce68703-3a52-4c6d-921f-4636f7d7c94f)
![изображение](https://github.com/vcmi/vcmi/assets/36762538/047d9bb6-1ca2-42dd-a720-0cde0d5263e7)
Start map and press end turn, or use autoskip
Nothing happens. AI doesn't scouting thorough whirlpool. Even on 200% (open map?? for AI)
It do nothing

But if you type cheat code `vcmieagles blue`, after that it going to me and kill. Works as intended.

test map [Whirlpool.zip](https://github.com/user-attachments/files/15721902/Whirlpool.zip)

mods
![изображение](https://github.com/vcmi/vcmi/assets/36762538/1684c62d-1298-4cad-a3b7-8e3ae1aff005)

PR #4226, opened 2024-07-05
Fix - NKAI not using monoliths and whirlpools

MichalZr6:
Probably fixes bugs: #3599 #4216 #4104


MichalZr6:
Sorry, I might have done this commit too soon... Have to test it.

nullkiller:
Actually this logic is for a different branch. Monolith itself is useless for AI. oneways are blocked at pathfinder as we do not know if we can get out of it. But I can suggest easy hack: hero is main and army strength like 20 000 or more. This one is a bit more reliable but still AI can get into a trap. The most reliable is to try figure out if a hero can get back from there but in case of FoW it can not be accomplished anyway and complex to implement.
Wirlpools, are you sure they do not work? Might be because they cause army losses. Btw theoretically this can be reflected by pathfinder. It has own column for army loss so adding there will allow tracking this fact and make decisions accordingly.

nullkiller:
AIPathfinderConfig is responsible for selecting allowed teleports. But it is still tricky as additional changes are required

nullkiller:
If you feel like enabling them is useful you can introduce configuration settings instead and mod. One day I will fix it but not now.

DjWarmonger:
IIRC my logic in VCAI was to use Monoliths for exploration. Just by entering it you reveal part of the map. Plus, this is typically done by scouts, so risk is lower.

nullkiller:
That may be ported to NKAI. At least with twoway it works exactly like this.

AI/Nullkiller/AIUtility.cpp
```diff
@@ -393,11 +393,12 @@ bool shouldVisit(const Nullkiller * ai, const CGHeroInstance * h, const CGObject
 		}
 		return false;
 	}
-	case Obj::MONOLITH_ONE_WAY_ENTRANCE:
 	case Obj::MONOLITH_ONE_WAY_EXIT:
+		return false;
+	case Obj::MONOLITH_ONE_WAY_ENTRANCE:
 	case Obj::MONOLITH_TWO_WAY:
 	case Obj::WHIRLPOOL:
-		return false;
```
nullkiller, line 400: This should be left as is. The mechanics of entering one-way monolith is a bit different and should relay on pathfinder instead. You can check my PR regarding whirlpools #4313 .

PR #4235, opened 2024-07-07
reorder fields to shrink struct TerrainTile from 96 to 80 bytes

kaja47:
When I was playing a very large map, computer's turn took up to 2 minutes and that give me time to fire up perf and profile the code while waiting.

I noticed that 1.7% of time was spent in function CPathfinderHelper::getNeighbors and half of that on a single load from an array backing multi_array of TerrainTiles. That signals the CMap::terrain is too big to fit in CPU caches.

This patch reorders fields in TerrainTile struct, shrinking it from 96 bytes to 80 bytes and hopefully helping to keep more tiles in the cache and speeding things up a little bit.

### DIFF
--- a/lib/mapping/CMapDefines.h
+++ b/lib/mapping/CMapDefines.h
@@ -97,10 +97,10 @@ struct DLL_LINKAGE TerrainTile
 	bool hasFavorableWinds() const;
 
 	const TerrainType * terType;
-	ui8 terView;
 	const RiverType * riverType;
-	ui8 riverDir;
 	const RoadType * roadType;
+	ui8 terView;
+	ui8 riverDir;
 	ui8 roadDir;
 	/// first two bits - how to rotate terrain graphic (next two - river graphic, next two - road);
 	///	7th bit - whether tile is coastal (allows disembarking if land or block movement if water); 8th bit - Favorable Winds effect


DjWarmonger:
Good spot, however this will impact serialization and makes savegames incompatible. This needs serialization version bump, and I'm not sure if you can even make it backwards-compatible at all.

kaja47:
I don't know how exactly it works, but order of fields, how they are serialized in the serialize handler, is kept the same as before.

IvanSavenko:
No, serialization is fine. 
Only order of field in structure has been changed. Serialization still uses old order.

DjWarmonger:
Oh, if that's the case, such optimization could be applied in other places. Ie. Bonus System.

IvanSavenko:
True, but it only has effect is some edge cases.
For example this particular case is actually a bug - vcmi would copy TerrainTile instead of using it by reference (fixed in my AI PR in beta, but not yet in develop).

Clang-tidy actually have warning for this case: https://releases.llvm.org/11.1.0/tools/clang/tools/extra/docs/clang-tidy/checks/clang-analyzer-optin.performance.Padding.html
(no real docs on that page, but I saw such warnings on my system in vcmi code)

BTW - @kaja47 I actually made some optimization targeting AI performance recently. These changes will be in develop in a week or two, you can either wait or check beta branch instead of develop.

MichalZr6:
I'm not very experienced in programming and still learning. This is very interesting that reordering fields can result in smaller object. Must read more about this.

dydzio0614:
@MichalZr6 you can read about memory layout / data alignment of c++ classes to find more info about this

kaja47:
> Oh, if that's the case, such optimization could be applied in other places. Ie. Bonus System.

@DjWarmonger I have spent yesterday and today gazing into the abyss of the  bonus system and I concluded some improvements can be made. With them on my testing save I can move from 28% time spent in bonus system during AI turns to 13% of the time according to perf. Maybe I can push it little bit more.

IvanSavenko:
@kaja47 
Recent develop I hope? 
I've made some improvements to AI performance here: https://github.com/vcmi/vcmi/pull/4220

However from what I see most of time spend in bonus system was spent on mutex locks. And haven't noticed any easy way to remove / reduce them.

If you've managed to improve performance on top of my changes - that would be great though.

PR #4236, opened 2024-07-07
2 small commits to fix inconsistent AI-behavior

Xilmi:
Description is in the description of the individual commits.

### DIFF
--- a/AI/Nullkiller/Pathfinding/AINodeStorage.h
+++ b/AI/Nullkiller/Pathfinding/AINodeStorage.h
@@ -27,8 +27,8 @@ namespace NKAI
 {
 namespace AIPathfinding
 {
-	const int BUCKET_COUNT = 3;
-	const int BUCKET_SIZE = 7;
+	const int BUCKET_COUNT = 1;
+	const int BUCKET_SIZE = 32;
 	const int NUM_CHAINS = BUCKET_COUNT * BUCKET_SIZE;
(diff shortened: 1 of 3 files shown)


AI/Nullkiller/Pathfinding/AINodeStorage.h
```diff
@@ -27,8 +27,8 @@ namespace NKAI
 {
 namespace AIPathfinding
 {
-	const int BUCKET_COUNT = 3;
-	const int BUCKET_SIZE = 7;
+	const int BUCKET_COUNT = 1;
```
nullkiller, line 30: This should be fetched from settings if compiler can allow it or should be left as is for now. Or you can change it according to AI_TRACE_LEVEL >= 1 for debug purposes like I did for parallelism.

IvanSavenko: @nullkiller should we merge with this change?

nullkiller: No I do not think we should include it in this form. This configuration turn hashmap into plain vector. This reduce certain random behavior as it no longer depends on Actor address but may hurt performance a bit. When I introduced it the impact of searching through plain vector seamed to be significant according to profiler.
The best option is to make it configurable, Reducing number will reduce memory footprint. Some 1-10 configuraiton for instance

PR #4239, opened 2024-07-08
changed to signature based file validation (gog installer)

Laserlicht:
use signature based validation instead of file ending

### DIFF
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -313,7 +313,7 @@ QString FirstLaunchView::getHeroesInstallDir()
 void FirstLaunchView::extractGogData()
 {
 #ifdef ENABLE_INNOEXTRACT
-	auto fileSelection = [this](QString type, QString filter, QString startPath = {}) {
+	auto fileSelection = [this](QString magic, QString filter, QString startPath = {}) {
 		QString titleSel = tr("Select %1 file...", "param is file extension").arg(filter);
 		QString titleErr = tr("You have to select %1 file!", "param is file extension").arg(filter);
 #if defined(VCMI_MOBILE)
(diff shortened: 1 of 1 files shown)


launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -332,10 +340,10 @@ void FirstLaunchView::extractGogData()
 		return file;
 	};
 
-	QString fileBin = fileSelection("bin", tr("GOG data") + " (*.bin)");
+	QString fileBin = fileSelection("idska32", tr("GOG data") + " (*.bin)");
 	if(fileBin.isEmpty())
 		return;
-	QString fileExe = fileSelection("exe", tr("GOG installer") + " (*.exe)", QFileInfo(fileBin).absolutePath());
+	QString fileExe = fileSelection("MZ", tr("GOG installer") + " (*.exe)", QFileInfo(fileBin).absolutePath());
```
kambala-decapitator, line 346: technically speaking, we should also check for PE format: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#signature-image-only

although I think that even this solution is already good

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -323,7 +323,15 @@ void FirstLaunchView::extractGogData()
...
-		else if(!file.endsWith("." + type, Qt::CaseInsensitive))
+		
+		QFile tmpFile(file);
+		if(!tmpFile.open(QIODevice::ReadOnly))
+		{
+			QMessageBox::critical(this, tr("File cannot opened"), titleErr);
+			return QString{};
+		}
+		QByteArray magicFile = tmpFile.read(magic.length());
+		if(!magicFile.startsWith(QByteArray{magic.toStdString().c_str()}))
```
kambala-decapitator, line 334: better pass `magic` to the lambda as QByteArray

Laserlicht: Thanks!

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Thanks. Yes, was aware of this.
But I think in our case it's good enough like it is. To keep it simple.

kambala-decapitator (approved):

PR #4246, opened 2024-07-10
remove slack references

Laserlicht:


docs/maintainers/Project_Infrastructure.md
IvanSavenko, line 1: I would keep this info from this file just in case

PR #4251, opened 2024-07-11
Fix RNG seeding

smanolloff:
### Overview

This PR implements RNG seeding and makes the application's behaviour deterministic when desired.

### Problem statement

The concept of RNG seeding seems to have been floating around VCMI's codebase for a long time, but unfortunately it was defunct in its current state. RNGs throughout the codebase were duplicated and/or re-seeded in an unpredictable manner. The lack of logging made it even harder to understand what's going on.

I saw this as an area of improvement, mainly due to its important role in troubleshooting. 

### Changes summary

* Added `settings["server"]["seed"]` (int, defaults to `0`) to use as the root seed. When `0`, a random, unpredictable seed will be used.
* Added `logRng` with a new `rng` logger domain for observability over the generated values.
* Replaced RNGs stored by value with references to the (thread-local) RNG, as copies get seeded unpredictably.

### Future improvements

The changes do not affect client-side RNGs, as I could not find an easy way to seed them. I believe client-side RNGs are much less important (to be honest, I have no idea what are they really used for), but that can be addressed in a future PR.

### Example

Using a seed of `5`, the below rng logs are produced at game start: (rng loglevel is `DEBUG`, comments are added by me):

```
// unpredictable seeds during RNG creation in each thread (these should get re-seeded later)
[runNetwork][rng] DEBUG setSeed(816555936)
[MainGUI][rng] DEBUG setSeed(1353855255)
[runServer][rng] DEBUG setSeed(-1567101344)

// re-seeding with the user-configured seed (5):
[runServer][rng] DEBUG setSeed(5)

// RNG sequence is now reproducible
// (re-launching with the same seed produces the exact same values)
[runServer][rng] DEBUG nextInt(8) -> 0
[runServer][rng] DEBUG nextInt(8) -> 4
[runServer][rng] DEBUG nextInt(0, 6) -> 0
[runServer][rng] DEBUG nextInt(0, 6) -> 5
[runServer][rng] DEBUG nextInt(0, 6) -> 5
[runServer][rng] DEBUG nextInt(0, 6) -> 6
[runServer][rng] DEBUG nextInt(0, 6) -> 4
[runServer][rng] DEBUG nextInt(0, 6) -> 6
[runServer][rng] DEBUG nextInt(0, 6) -> 5
[runServer][rng] DEBUG nextInt(0, 6) -> 0
[runServer][rng] DEBUG nextInt(0, 6) -> 0
[runServer][rng] DEBUG nextInt(0, 6) -> 5
[runServer][rng] DEBUG nextInt(0, 6) -> 6
[runServer][rng] DEBUG nextInt(0, 6) -> 2
```

### DIFF
--- a/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
+++ b/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
@@ -165,7 +165,7 @@ void DangerHitMapAnalyzer::calculateTileOwners()
 	auto addTownHero = [&](const CGTownInstance * town)
 	{
 			auto townHero = temporaryHeroes.emplace_back(std::make_unique<CGHeroInstance>(town->cb)).get();
-			CRandomGenerator rng;
+			CRandomGenerator & rng = CRandomGenerator::getDefault();
 			auto visitablePos = town->visitablePos();
 			
 			townHero->setOwner(ai->playerID); // lets avoid having multiple colors
--- a/lib/CRandomGenerator.cpp
+++ b/lib/CRandomGenerator.cpp
@@ -15,6 +15,7 @@ VCMI_LIB_NAMESPACE_BEGIN
 
 CRandomGenerator::CRandomGenerator()
 {
+	logRng->trace("(constructor)");
 	resetSeed();
 }
 
@@ -25,11 +26,13 @@ CRandomGenerator::CRandomGenerator(int seed)
 
 void CRandomGenerator::setSeed(int seed)
 {
+	logRng->debug("setSeed(%d)", seed);
 	rand.seed(seed);
 }
 
 void CRandomGenerator::resetSeed()
 {
+	logRng->trace("resetSeed()");
 	boost::hash<std::string> stringHash;
 	auto threadIdHash = stringHash(boost::lexical_cast<std::string>(boost::this_thread::get_id()));
 	setSeed(static_cast<int>(threadIdHash * std::time(nullptr)));
@@ -37,6 +40,7 @@ void CRandomGenerator::resetSeed()
 
 TRandI CRandomGenerator::getIntRange(int lower, int upper)
 {
+	logRng->trace("getIntRange(%d, %d)", lower, upper);
 	if (lower <= upper)
 		return std::bind(TIntDist(lower, upper), std::ref(rand));
 	throw std::runtime_error("Invalid range provided: " + std::to_string(lower) + " ... " + std::to_string(upper));
@@ -44,6 +48,7 @@ TRandI CRandomGenerator::getIntRange(int lower, int upper)
 
 vstd::TRandI64 CRandomGenerator::getInt64Range(int64_t lower, int64_t upper)
 {
+	logRng->trace("getInt64Range(%d, %d)", lower, upper);
 	if(lower <= upper)
 		return std::bind(TInt64Dist(lower, upper), std::ref(rand));
 	throw std::runtime_error("Invalid range provided: " + std::to_string(lower) + " ... " + std::to_string(upper));
@@ -51,21 +56,28 @@ vstd::TRandI64 CRandomGenerator::getInt64Range(int64_t lower, int64_t upper)
 
 int CRandomGenerator::nextInt(int upper)
 {
-	return getIntRange(0, upper)();
+	auto res = getIntRange(0, upper)();
+	logRng->debug("nextInt(%d) -> %d", upper, res);
+	return res;
 }
 
 int CRandomGenerator::nextInt(int lower, int upper)
 {
-	return getIntRange(lower, upper)();
+	auto res = getIntRange(lower, upper)();
+	logRng->debug("nextInt(%d, %d) -> %d", lower, upper, res);
+	return res;
 }
 
 int CRandomGenerator::nextInt()
 {
-	return TIntDist()(rand);
+	auto res = TIntDist()(rand);
+	logRng->debug("nextInt() -> %d", res);
+	return res;
 }
 
 vstd::TRand CRandomGenerator::getDoubleRange(double lower, double upper)
 {
+	logRng->trace("getDoubleRange(%f, %f)", lower, upper);
 	if(lower <= upper)
 		return std::bind(TRealDist(lower, upper), std::ref(rand));
 	throw std::runtime_error("Invalid range provided: " + std::to_string(lower) + " ... " + std::to_string(upper));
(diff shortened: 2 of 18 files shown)


dydzio0614:
I would target develop branch instead of beta for these changes

smanolloff:
> I would target develop branch instead of beta for these changes

Thanks. I am still learning the git flows in this project. I will target `develop` and rebase to get rid of the unrelated commits that appeared.

IvanSavenko:
> Thanks. I am still learning the git flows in this project. I will target `develop` and rebase to get rid of the unrelated commits that appeared.

`develop` - used for general development. Right now this is where 1.6 development is happening
`beta` - used for current release patches, 1.5.X at the moment. Limited to safe changes or those that we don't want to postpone to next major release.
`master` - latest published version (or soon-to-be-published version as of right now). Not used for feature PR targeting

So yeah, PR's like this one should probably target develop. If rebase is an issue, you can wait a day for 1.5.4 to be out (and for beta to be merged into develop)

smanolloff:
> So yeah, PR's like this one should probably target develop. If rebase is an issue, you can wait a day for 1.5.4 to be out (and for beta to be merged into develop)

No problem, develop is OK. The rebase got messy pretty fast, so I cherry-picked instead.

AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
```diff
@@ -165,7 +165,7 @@ void DangerHitMapAnalyzer::calculateTileOwners()
 	auto addTownHero = [&](const CGTownInstance * town)
 	{
 			auto townHero = temporaryHeroes.emplace_back(std::make_unique<CGHeroInstance>(town->cb)).get();
-			CRandomGenerator rng;
+			CRandomGenerator & rng = CRandomGenerator::getDefault();
```
IvanSavenko, line 168: To be honest this is not much better. Not worse, sure, but not really an improvement.
`getDefault` returns thread-local generator and AI shuffles through threads constantly.

lib/CRandomGenerator.cpp
```diff
@@ -25,47 +26,58 @@ CRandomGenerator::CRandomGenerator(int seed)
...
 {
+	logRng->trace("resetSeed()");
 	boost::hash<std::string> stringHash;
 	auto threadIdHash = stringHash(boost::lexical_cast<std::string>(boost::this_thread::get_id()));
 	setSeed(static_cast<int>(threadIdHash * std::time(nullptr)));
 }
 
 TRandI CRandomGenerator::getIntRange(int lower, int upper)
 {
+	logRng->trace("getIntRange(%d, %d)", lower, upper);
```
IvanSavenko, line 43: This logging would likely become very spammy. And we already have a lot of logging.

IvanSavenko (commented):
A lot of it actually duplicates my open (but not yet finished) PR: https://github.com/vcmi/vcmi/pull/4071

So not sure what the best approach here would be. Perhaps best option would be for me to integrate any changes from here that my PR is still missing.

lib/CRandomGenerator.cpp
smanolloff: Would you please clarify what are you suggesting here - should I remove the `trace` logging (getIntRange, getDoubleRange, etc.) and leave only the `debug` logging (nextInt, nextDouble, etc.)? Or should the entire RNG log be removed?

IvanSavenko: Not sure. 
At the moment even `trace` level is written to log file. And it is not uncommon for log file size to be measured in hundreds of Mb's.

So we probably should either remove it or change logging level of logger to `debug`. Although don't remember where this is configured.

AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
smanolloff: Now that you mention it. It is _worse_, I believe. `CRandomGenerator` uses `std::minstd_rand` under the hood and passing it by reference between threads is not safe (please correct me if I am wrong). Concurrently generating numbers might lead to UB due to concurrent RNG state modification. With that said, I believe there's some more work to be done here :/

smanolloff:
> A lot of it actually duplicates my open (but not yet finished) PR: #4071
> 
> So not sure what the best approach here would be. Perhaps best option would be for me to integrate any changes from here that my PR is still missing.

As you wish. If you prefer to roll it out all at once - we can close this one and you can cherry-pick the relevant changes.

It may turn out this PR is far from ready anyway, given my raised concern regarding storing references to a thread-local RNG.

lib/CRandomGenerator.cpp
smanolloff: > At the moment even trace level is written to log file

That sounds incorrect to me -- at least it does not correlate to what I am observing locally. Please tell me where am I wrong -- here is what I know:

The log levels are set per-domain in `settings.json` (platform-dependant, on macOS is `~/Library/Application Support/vcmi/config/settings.json`), and here is what I have in mine:

```json
    "loggers" : [
      {"domain" : "global", "level" : "error"},
      {"domain" : "ai", "level" : "error"},
      {"domain" : "rng", "level" : "error"},
      {"domain" : "network", "level" : "trace"},
      {"domain" : "mod", "level" : "error"},
      {"domain" : "animation", "level" : "error"},
      {"domain" : "bonus", "level" : "error"}
    ]
```

If a domain is missing from this file, it will fallback to the default value (`trace`) as defined in [config/schemas/settings.json](https://github.com/vcmi/vcmi/blob/1.5.3/config/schemas/settings.json#L540).

One reason I can think of is: new installations come with no `config.json` so they always use the default level (`trace`) for _all_ loggers. Is that the issue?

IvanSavenko: > One reason I can think of is: new installations come with no config.json so they always use the default level (trace) for all loggers. Is that the issue?

Yep. You can configure it that way. But at the moment all players that have not deliberately configured log system (so >99% of them) have all logging domains at 'trace' level

nullkiller: @IvanSavenko Should we change it? Trace is definitely not that level needed by default?

nullkiller: @smanolloff regarding very extensive logging you can go the way of #ifdef. It is better than log level switching if will be used in very high-load place because even preparing log string may take some CPU.

smanolloff: Ah, true, I also thought it initially, then I did some benchmarks and it turned out the performance impact is around 2%, which I would consider negligible (and comparable to the measurement  accuracy error itself). But I agree that  using a preprocessor macro is fine, if it wouldn't degrade readability.
We took the discussion off the GH comment section, but in summary decided to leave this PR open until Ivan decides which changes are worth integrating into his own work branch.

IvanSavenko: > @IvanSavenko Should we change it? Trace is definitely not that level needed by default?

Don't know. It is certainly useful since quite a lot of player reports are simple "game crashed". So detailed info from logs help a lot with tracking down last in-game events.

In theory we could try to disable trace level only for release builds, but release builds is where most player reports come from. If we do so, we should review what we log and decide which trace-level info is useful and should be promoted to 'debug' and what can be left in 'trace' and disabled.

> @smanolloff regarding very extensive logging you can go the way of #ifdef. It is better than log level switching if will be used in very high-load place because even preparing log string may take some CPU.

Actually log string is only generated if log level check passes. Formatting arguments are computed unconditionally though. So log entries like these ones will only generate overhead in form of a single check for log level.

smanolloff: If the amount of logged trace messages is a concern, **and** there is no way to turn it off for players by default without affecting all other logging, I would suggest to either:
* think about options for shipping a `settings.json` with pre-configured loglevels (as part of the game installation).
* leave it on for debug builds only, and use it to ensure it's working correctly. For release builds we would only care about the log line `setSeed(), as everything else can be inferred from those. Maybe leaving the logs in the constructor and `resetSeed()` as well for a bit more context, as they are not called often as well.

IvanSavenko:
Closing, integrated into PR #4071

PR #4253, opened 2024-07-11
Bind VCMI server to a random TCP port

smanolloff:
### Overview

This PR allows to set `settings["server"]["localPort"]=0` in order to bind VCMI server to a random TCP port.

### Problem statement

For the purposes of ML, being able to run multiple VCMI instances in parallel is essential in order to efficiently use the available hardware resources by training multiple AI agents simultaneously. Since there can only be one process bound to a given port, the workaround is to manually select a free TCP port prior to launching the application, but there are many issues with this approach.

In addition, when running VCMI server+client locally, there is a (small) chance of a port conflict with third party software using the same port.

### Changes summary

* `settings["server"]["localPort"]` can now be set to `0` (i.e. randomly assigned port)
* A new function `CVCMIServer::prepare` (naming suggestions are welcome) must now be called in order to bind to a network socket (this used to be part of the server constructor). This function returns the assigned port.
* Several functions (`IServerRunner::start`, `ServerProcessRunner::start`, `ServerThreadRunner::start`, `NetworkInterface::start`, etc.) return type was changed from `void` to `uint16_t` (the assigned port).

### Example

Using a `localPort` of `0`, we can see that a random port is assigned (63615 in this case):

```
[0x1eaadfac0][network] INFO Randomly assigned port will be used
[0x1eaadfac0][network] INFO Listening for connections at port 63615
```

Subsequently started VCMI server processes will bind to a different port without conflict.

### DIFF
--- a/client/ServerRunner.cpp
+++ b/client/ServerRunner.cpp
@@ -39,7 +39,6 @@ uint16_t ServerThreadRunner::start(uint16_t cfgport, bool connectToLobby, std::s
 
 	threadRunLocalServer = boost::thread([this, connectToLobby, &promise]{
 		setThreadName("runServer");
-		std::this_thread::sleep_for(std::chrono::seconds(5));
 		uint16_t port = server->prepare(connectToLobby);
 		promise.set_value(port);
 		server->run();
--- a/client/ServerRunner.cpp
+++ b/client/ServerRunner.cpp
@@ -20,6 +20,8 @@
 #include <boost/process/io.hpp>
 #endif
 
+#include <future>
+
 ServerThreadRunner::ServerThreadRunner() = default;
 ServerThreadRunner::~ServerThreadRunner() = default;
 
--- a/client/ServerRunner.cpp
+++ b/client/ServerRunner.cpp
@@ -23,28 +23,29 @@
 ServerThreadRunner::ServerThreadRunner() = default;
 ServerThreadRunner::~ServerThreadRunner() = default;
 
-uint16_t ServerThreadRunner::start(uint16_t port, bool connectToLobby, std::shared_ptr<StartInfo> startingInfo)
+uint16_t ServerThreadRunner::start(uint16_t cfgport, bool connectToLobby, std::shared_ptr<StartInfo> startingInfo)
 {
-	server = std::make_unique<CVCMIServer>(port, true);
+	// cfgport may be 0 -- the real port is returned after calling prepare()
+	server = std::make_unique<CVCMIServer>(cfgport, true);
 
 	if (startingInfo)
 	{
 		server->si = startingInfo; //Else use default
 	}
 
-	uint16_t srvport = port;
+	std::promise<uint16_t> promise;
 
-	threadRunLocalServer = boost::thread([this, connectToLobby, &srvport]{
+	threadRunLocalServer = boost::thread([this, connectToLobby, &promise]{
 		setThreadName("runServer");
-		srvport = server->prepare(connectToLobby);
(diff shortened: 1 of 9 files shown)


client/ServerRunner.cpp
```diff
@@ -23,19 +23,30 @@
...
 		setThreadName("runServer");
+		srvport = server->prepare(connectToLobby);
 		server->run();
 	});
+
+	while(srvport == 0) {
+		logNetwork->trace("Waiting for server port...");
+		boost::this_thread::sleep(boost::posix_time::milliseconds(100));
+		logNetwork->debug("Server port: %d", srvport);
+	}
```
IvanSavenko, line 47: Can we do something better than sleep-based thread synchronization? std::future for example? Or anything else that is not sleep?

lib/network/NetworkServer.cpp
```diff
@@ -19,16 +19,17 @@ NetworkServer::NetworkServer(INetworkServerListener & listener, const std::share
 {
 }
 
-void NetworkServer::start(uint16_t port)
+uint16_t NetworkServer::start(uint16_t port)
 {
-	acceptor = std::make_shared<NetworkAcceptor>(*io, boost::asio::ip::tcp::endpoint(boost::asio::ip::tcp::v4(), port));
-	startAsyncAccept();
+	acceptor = std::make_shared<NetworkAcceptor>(*io, boost::asio::ip::tcp::endpoint(boost::asio::ip::address::from_string("127.0.0.1"), port));
```
IvanSavenko, line 24: Why was this change necessary?

client/ServerRunner.cpp
smanolloff: Sure, I will look into that.

lib/network/NetworkServer.cpp
smanolloff: oops. Not only is it not necessary, it is a mistake to include it in the PR. I will remove it, thanks for being vigilant.

Context: VCMI binds to 0.0.0.0 by default, which produces an annoying OS firewall warning and that's why I had changed it to localhost. That would, of course, prevent external connections to the server, so, yeah, I will change it back.

Laserlicht:
@smanolloff Thanks for your contribution! Do you plan to backport more of your gym environment stuff in upstream? Would be great if we had an easy to use environment to train models (that will be keep in sync with current VCMI version). And maybe in the future a good neural network based AI.

Feel free to also join discord. There are some discussion about weak points of current AI.

smanolloff:
> @smanolloff Thanks for your contribution! Do you plan to backport more of your gym environment stuff in upstream? Would be great if we had an easy to use environment to train models (that will be keep in sync with current VCMI version). And maybe in the future a good neural network based AI.
> 
> Feel free to also join discord. There are some discussion about weak points of current AI.

Oh, yes, actually I am finalizing some of the changes there and was wondering what would be the proper approach for such a contribution. I even have some decent AI models trained already which seem to outperform the BattleAI bot. I am currently working on abstracting away the changes in order to make them as less intrusive as possible, but, yeah, a discussion in the Discord channel is surely a good idea, I am planning to approach you (and the other VCMI devs) there soon to work out the possible rollout strategies for such a feature :D And thank you 🍺 Cheers

Laserlicht:
Would it be possible to make this default. Any drawbacks?

For testing I've often multiple instances open. Which crashes with bind. Even on android if you have develop version in background and try to start release version.

Laserlicht:
Maybe only if default port fails.

PR #4261, opened 2024-07-14
Reduce size of Bonus struct from 320 bytes to 296 bytes.

kaja47:
- Internal enums were resized to occupy single byte.
- Duration bitmask uses 16 bit integer directly instead of std::bitset<11> which consumed 8 bytes.
- Fields shuffled to minimise padding and keep the most useful data on first 2 cache lines.

### DIFF
--- a/lib/bonuses/BonusEnum.cpp
+++ b/lib/bonuses/BonusEnum.cpp
@@ -60,10 +60,11 @@ namespace BonusDuration
 	JsonNode toJson(const Type & duration)
 	{
 		std::vector<std::string> durationNames;
-		for(auto durBit = 0; durBit < duration.size(); durBit++)
+		for(size_t durBit = 0; durBit < Size; durBit++)
 		{
-			if(duration[durBit])
-				durationNames.push_back(vstd::findKey(bonusDurationMap, duration & Type().set(durBit)));
+			Type value = duration & (1 << durBit);
+			if(value)
+				durationNames.push_back(vstd::findKey(bonusDurationMap, value));
 		}
 		if(durationNames.size() == 1)
 		{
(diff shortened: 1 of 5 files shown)


lib/bonuses/BonusEnum.cpp
```diff
@@ -60,10 +60,10 @@ namespace BonusDuration
 	JsonNode toJson(const Type & duration)
 	{
 		std::vector<std::string> durationNames;
-		for(auto durBit = 0; durBit < duration.size(); durBit++)
+		for(auto durBit = 0; durBit < BonusDuration::Size; durBit++)
```
kambala-decapitator, line 63:
```suggestion
		for(size_t durBit = 0; durBit < BonusDuration::Size; durBit++)
```

lib/bonuses/BonusEnum.cpp
```diff
@@ -60,10 +60,10 @@ namespace BonusDuration
...
 		{
-			if(duration[durBit])
-				durationNames.push_back(vstd::findKey(bonusDurationMap, duration & Type().set(durBit)));
+			if(duration & (1 << durBit))
+				durationNames.push_back(vstd::findKey(bonusDurationMap, (unsigned short)(duration & (1 << durBit))));
```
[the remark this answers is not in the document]
IvanSavenko: vstd::to_underlying is more suitable here instead of static cast - will also handle different type (if we were to change it in future)

lib/bonuses/BonusEnum.h
```diff
@@ -212,15 +212,18 @@ class JsonNode;
 	BONUS_VALUE(INDEPENDENT_MIN) //used for SECONDARY_SKILL_PREMY bonus
 
 
-enum class BonusType
+enum class BonusType : ui8
```
IvanSavenko, line 215: These type names are kind of deprecated, use standard ones:
```suggestion
enum class BonusType : uint8_t
```
Same for rest of changes in this file

kaja47:
I fixed what you pointed out. Do you prefer the first commit to be amended to include all the changes or a series of incremental changes?

IvanSavenko:
> I fixed what you pointed out. Do you prefer the first commit to be amended to include all the changes or a series of incremental changes?

Either is fine.

PR #4263, opened 2024-07-14
JsonParser::parse(): Print file name if there were errors while parsing

Alexander-Wilms:
In cases where the file name was not specified, the warning messages were not very useful.

Example:

```
File <unknown> is not a valid JSON file!
At line 33, position 1 warning: Comma expected!
```
```json
{
        "name" : "New Old Heroes",
        "description" : "New heroes based on old 3DO artwork and other game appearances. Requires Horn of the Abyss.",
        "modType" : "Heroes",
        "version" : "1.2.0",
        "author" : "Aphra",
        "contact" : "",

        "heroes" :
        [
                "config/gwenneth.json",
                "config/balindar.json",
                "config/nicolas.json",
                "config/kastore.json",
                "config/kydoimos.json",
                "config/athe.json",
                "config/miseria.json",
                "config/areshrak.json",
                "config/pactal.json",
                "config/zog.json"
        ],

        "changelog" :
    {
        "1.0.0"   : ["Initial release"],
        "1.1.0"   : ["Added Nicolas Gryphonheart and Kastore"],
        "1.1.1"   : ["Bug fixes"],
        "1.2.0"   : ["Added some HotA portrait-only campaign heroes"]
    },

        "depends" :
                [ "hota.neutralCreatures" ]
        "keepDisabled" : true
}
```

### DIFF
--- a/lib/json/JsonNode.h
+++ b/lib/json/JsonNode.h
@@ -72,8 +72,8 @@ public:
 	explicit JsonNode(const std::string & string);
 
 	/// Create tree from Json-formatted input
-	explicit JsonNode(const std::byte * data, size_t datasize);
-	explicit JsonNode(const std::byte * data, size_t datasize, const JsonParsingSettings & parserSettings);
+	explicit JsonNode(const std::byte * data, size_t datasize, const std::string fileName);
+	explicit JsonNode(const std::byte * data, size_t datasize, const JsonParsingSettings & parserSettings, const std::string fileName);
 
 	/// Create tree from JSON file
 	explicit JsonNode(const JsonPath & fileURI);
--- a/lib/json/JsonNode.h
+++ b/lib/json/JsonNode.h
@@ -72,8 +72,8 @@ public:
 	explicit JsonNode(const std::string & string);
 
 	/// Create tree from Json-formatted input
-	explicit JsonNode(const std::byte * data, size_t datasize, const std::string fileName);
-	explicit JsonNode(const std::byte * data, size_t datasize, const JsonParsingSettings & parserSettings, const std::string fileName);
+	explicit JsonNode(const std::byte * data, size_t datasize, const std::string & fileName);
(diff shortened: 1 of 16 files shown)


IvanSavenko:
To be honest I would rather find all cases that generate 'unknown' file and provide better name for it.

Alexander-Wilms:
That'd be all calls to the JsonNode constructor with three arguments, but I don't know how to find them.

```
JsonNode::JsonNode(const std::byte * data, size_t datasize, const JsonParsingSettings & parserSettings)
{
	JsonParser parser(data, datasize, parserSettings);
	*this = parser.parse("<unknown>");
}
```

IvanSavenko:
Edit this constructor to also require string parameter and check for compile errors. Or remove it and again, check for compile errors.

Alexander-Wilms:
Done. In case of network packets with invalid JSON, one still does not have much to go on.

One could only print the JSON in `JsonParser::parse` in case of errors if the fileName matches `<.*>`. Or one could create a more elegant solution or just leave it as is if these issues are rare.

lib/json/JsonNode.h
```diff
@@ -72,8 +72,8 @@ class DLL_LINKAGE JsonNode
 	explicit JsonNode(const std::string & string);
 
 	/// Create tree from Json-formatted input
-	explicit JsonNode(const std::byte * data, size_t datasize);
-	explicit JsonNode(const std::byte * data, size_t datasize, const JsonParsingSettings & parserSettings);
+	explicit JsonNode(const std::byte * data, size_t datasize, const std::string fileName);
+	explicit JsonNode(const std::byte * data, size_t datasize, const JsonParsingSettings & parserSettings, const std::string fileName);
```
IvanSavenko, line 76:
```suggestion
	explicit JsonNode(const std::byte * data, size_t datasize, const std::string & fileName);
	explicit JsonNode(const std::byte * data, size_t datasize, const JsonParsingSettings & parserSettings, const std::string & fileName);
```

PR #4272, opened 2024-07-15
Allow opening single hero when two heroes in town via shift+click

dydzio0614:
As in title


IvanSavenko:
BTW:
- Shift+H will open garrisoned hero (if there is one)
- Ctrl+H will open visiting hero (if any)

dydzio0614:
yeah, though it is potentially way faster to prepare shift & doubleclick for timer-constrained games

kambala-decapitator:
> Ctrl+H will open visiting hero (if any)

Might not work on macos because cmd+h is a standard shortcut to hide current window, need to test

PR #4274, opened 2024-07-15
Rule based AI

Xilmi:
This is the first draft of my experimental AI-changes that I deem somewhat presentable.

I basically replaced the fuzzy-logic-approach of evaluating the AI's options by my own evaluating-heuristics that are supposed to make it play in more or less the same way that I would.

Instead of evaluating all options in a single pass, the AI now does three passes. The latter passes are only attempted, when their predecessor didn't provide a valid result.

The first pass is all about conquest. The AI will consider only options that allow it to kill an enemy hero or conquer a town. It will, however also be preservative and not just throw large its army away to achieve this goal as quickly as possible. Earlier test-versions did this and conquered maps in a very low amount of turns. But they bled so many troops clearing the path that I had to make it more conservative.

The second pass is: "Do whatever is the most opportunistic thing that yields any kind of advantage without losing troops." If the AI isn't in a position to conquer anything, it will look for tasks that their heros can accomplish easily. It will try to maximize movement-efficiency for this purpose rather than being overly strategic of the value of what it picks up. Note that hero-safety is also an important part of that. So it won't loot around stronger enemy heros.

If nothing is available that fulfills the above requirements, the AI looks for good cost:benefit fights against neutrals it can take. It evaluates safety, distance and profit to find the best compromise.

There are some exceptions in these general rules. For example if something doesn't really yield any safe option but it can go and defend a town, it will do so in order to prevent easy recaptures.


dydzio0614:
I would change pull request type to "draft"

AI/Nullkiller/Pathfinding/AINodeStorage.cpp
```diff
@@ -1536,7 +1536,7 @@ bool AIPath::containsHero(const CGHeroInstance * hero) const
 
 uint64_t AIPath::getTotalArmyLoss() const
 {
-	return armyLoss + targetObjectArmyLoss;
+	return armyLoss > targetObjectArmyLoss ? armyLoss : targetObjectArmyLoss;
```
nullkiller, line 1539: Army loss should be cumulative. The issue of double loss counting may be because we simulate battle with same monster twice. You can check how it will work with guarded dragon utopia for instance when guard strength is close to utopia strength.

Xilmi: How do I avoid checking the battle with the same enemy twice? I've looked at logs and usually it was either armyLoss or targetObjectArmyLoss that contained a value. In rare occassions it was both and in some of those both had the exact same value, which indicated the double-checking of the same fight.

nullkiller: Maybe I can take a look onto it if you provide me a small simple map where it is visible so I can debug.

nullkiller: Or I can take a look myself. Most likely it is fighting monster by cluster unlock

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -647,7 +647,20 @@ double CGHeroInstance::getFightingStrength() const
...
+	{
+		if (spellbookContainsSpell(spell) && spell.toSpell()->isCombat())
+		{
+			atLeastOneCombatSpell = true;
+			break;
+		}
+	}
+	if (!atLeastOneCombatSpell)
+		return 1;
+	return sqrt((1.0 + 0.05*getPrimSkillLevel(PrimarySkill::KNOWLEDGE) * mana / manaLimit()) * (1.0 + 0.05*getPrimSkillLevel(PrimarySkill::SPELL_POWER) * mana / manaLimit()));
```
IvanSavenko, line 663: This will also affect selection of hero to transfer in campaigns

lib/ResourceSet.h
```diff
@@ -171,14 +191,14 @@ class ResourceSet
...
-// 			return true;
-// 		}
+ 	bool operator<(const ResourceSet &rhs)
+ 	{
+ 		for(int i = 0; i < size(); i++)
+			if (this->container.at(i) < rhs[i])
+ 				return true;
+
+ 		return false;
+ 	}
```
IvanSavenko, line 201: Is this used for sorting? In that case it is not suitable as sorting function - a < b does not means that b > a.

If this is used for things like "can I afford this" then might be better to turn it into a method with clear name instead of operator overloading

lib/ResourceSet.h
```diff
@@ -148,6 +148,26 @@ class ResourceSet
...
+		for (size_t i = 0; i < container.size(); ++i) {
+			if (container.at(i) > 0) { // We only care about fulfilling positive needs
+				if (income[i] == 0) {
+					// If income is 0 and we need a positive amount, it's impossible to fulfill
+					return INT_MAX;
+				}
+				else {
+					// Calculate the number of times we need to accumulate income to fulfill the need
+					float divisionResult = static_cast<float>(container.at(i)) / static_cast<float>(income[i]);
+					int ceiledResult = static_cast<int>(std::ceil(divisionResult));
```
IvanSavenko, line 163:
```suggestion
					int ceiledResult = vstd::divideAndCeil(container.at(i), income[i]);
```

lib/ResourceSet.h
```diff
@@ -148,6 +148,26 @@ class ResourceSet
 		return ret;
 	}
 
+	int div(const ResourceSet& income) {
```
IvanSavenko, line 151: Method probably needs better name, or at least a comment to explain what it does.

Returns how many items of "this" we can afford with provided income?

lib/CCreatureSet.h
```diff
@@ -109,6 +109,7 @@ class DLL_LINKAGE CStackInstance : public CBonusSystemNode, public CStackBasicDe
 	FactionID getFaction() const override;
 
 	virtual ui64 getPower() const;
+	virtual ui64 getCost() const;
```
IvanSavenko, line 112: - use better name
- add comment to explain what it does
```suggestion
	/// Returns total market value of resources needed to recruit this unit
	virtual ui64 getMarketValue() const;
```

AI/Nullkiller/Behaviors/RecruitHeroBehavior.cpp
```diff
@@ -45,50 +46,68 @@ Goals::TGoalVec RecruitHeroBehavior::decompose(const Nullkiller * ai) const
...
-			{
-				if((obj->ID == Obj::RESOURCE)
-					|| obj->ID == Obj::TREASURE_CHEST
-					|| obj->ID == Obj::CAMPFIRE
-					|| isWeeklyRevisitable(ai, obj)
-					|| obj->ID ==Obj::ARTIFACT)
+				score *= hero->getArmyCost();
+				if (hero->type->heroClass->faction == town->getFaction())
+					score *= 1.5;
+				if (visitability == 0)
```
IvanSavenko, line 90: visitability is floating point, and equality comparison with floating points might break in platform-specific way in release mode.

Either replace with integer types or use vstd::isAlmostZero
```suggestion
				if (vstd::isAlmostZero(visitability))
```

AI/Nullkiller/Engine/PriorityEvaluator.cpp
```diff
@@ -1259,15 +1358,203 @@ float PriorityEvaluator::evaluate(Goals::TSubgoal task)
...
+
+		switch (priorityTier)
+		{
+			case PriorityTier::INSTAKILL: //Take towns / kill heroes in immediate reach
+			{
+				if (evaluationContext.turn > 0)
+					return 0;
+				if(evaluationContext.conquestValue > 0)
+					score = 1000;
+				if (score == 0 || (evaluationContext.enemyHeroDangerRatio > 1 && (evaluationContext.turn > 0 || evaluationContext.isExchange) && !ai->cb->getTownsInfo().empty()))
```
IvanSavenko, line 1407: More floating point comparisons here (and below)
Either use integer type for score or vstd::isAlmostZero for comparing with zero

AI/BattleAI/BattleEvaluator.cpp
```diff
@@ -729,6 +729,60 @@ bool BattleEvaluator::attemptCastingSpell(const CStack * activeStack)
...
+								ps.dest.at(0).hexValue.hex,
+								unit->creatureId().toCreature()->getNameSingularTranslated(),
+								unit->getCount(),
+								dpsReduce,
+								oldHealth,
+								newHealth);
+#endif
+						}
+					}
+				}
```
IvanSavenko, line 785: Copy-pasted block of code? Identical / almost identical code immediately below

AI/Nullkiller/Behaviors/BuildingBehavior.cpp
```diff
@@ -49,26 +49,52 @@ Goals::TGoalVec BuildingBehavior::decompose(const Nullkiller * ai) const
 	auto & developmentInfos = ai->buildAnalyzer->getDevelopmentInfo();
 	auto isGoldPressureLow = !ai->buildAnalyzer->isGoldPressureHigh();
 
+	ai->dangerHitMap->updateHitMap();
+
 	for(auto & developmentInfo : developmentInfos)
 	{
-		for(auto & buildingInfo : developmentInfo.toBuild)
+		bool emergencyDefense = false;
+		uint8_t closestThreat = UINT8_MAX;
```
IvanSavenko, line 57:
```suggestion
		uint8_t closestThreat = std::numeric_limits<uint8_t>::max();
```

AI/Nullkiller/Behaviors/BuyArmyBehavior.cpp
```diff
@@ -28,29 +28,45 @@ Goals::TGoalVec BuyArmyBehavior::decompose(const Nullkiller * ai) const
...
 	if(heroes.empty())
 	{
 		return tasks;
 	}
 
+	ai->dangerHitMap->updateHitMap();
+
 	for(auto town : cb->getTownsInfo())
 	{
+		//If we can recruit a hero that comes with more army than he costs, we are better off spending our gold on them
```
IvanSavenko, line 42: You might also want to consider hero faction. Sure, necromancer might provide more troops than buying troops directly, but its undead troops would kill morale and require free slots in army to actually benefit from new troops

lib/ResourceSet.h
Xilmi: I checked, it's just for a check whether something can be afforded. For which a method "canAfford" already existed. Which I probably didn't know at the time. I'll just remove it and use the existing method instead.

lib/ResourceSet.h
Xilmi: Method got renamed and got a comment describing what it does.

AI/Nullkiller/Engine/PriorityEvaluator.cpp
Xilmi: Replaced these with vstd::isAlmostZero aswell.

AI/Nullkiller/Behaviors/BuyArmyBehavior.cpp
Xilmi: I will remove this check entirely because it is pointless. I debugged at some point and found out that the army heroes come with usually is around 1k and thus never fulfills the criteria of having more value than if the army was bought directly.

AI/BattleAI/BattleEvaluator.cpp
Xilmi: The block is copy-pasted and modified but it has a different meaning.
The old part tracked the health before and after of units that survived a spell and added that to the value of the spell.

The new part checks for units that existed before but no longer exist after the spell.
Without it the spell-score-evaluation comes to wrong results and would basically never get a positive value for a spell that kills the whole stack.

Maybe @nullkiller should chime in here and see if he can come up with a solution that is more elegant than doing the almost same thing twice.

lib/mapObjects/CGHeroInstance.cpp
Xilmi: Can we add a default-parameter like "bool useMaxMana = false", set it to true for the selection of heroes to transfer in campaigns and if it's true use it as if mana is manaLimit?

Xilmi: Made a separate version of getMagicStrengthForCampaign and getHeroStrengthForCampaign, so that heroes' magic-power is calculated the traditional way for the purpose of whether the hero should be brought for the next campaign-mission.

themaster567:
So this more or less boils down to an evolution of your AI work for Brutal OpenXcom Extended?

Will the AI here too prioritize hiding in far off corners of the map? (joke)

AI/BattleAI/BattleEvaluator.cpp
nullkiller: Possibly I can take a look on it. Are you sure it is not already fixed just in case?

Xilmi: Okay, then I just remove that block

AI/BattleAI/BattleEvaluator.cpp
```diff
@@ -769,10 +823,14 @@ bool BattleEvaluator::attemptCastingSpell(const CStack * activeStack)
 
 #if BATTLE_TRACE_LEVEL >= 1
 						logAi->trace(
-							"Spell affects %s (%d), dps: %2f",
+							"Spell %s to %d affects %s (%d), dps: %2f oldHealth: %d newHealth: %d", 
+							ps.spell->getNameTranslated(),
+							ps.dest.at(0).hexValue.hex,
```
nullkiller, line 828: Will Armageddon have ps.dest.at(0).hexValue.hex?

AI/Nullkiller/Behaviors/BuildingBehavior.cpp
```diff
@@ -49,26 +49,49 @@ Goals::TGoalVec BuildingBehavior::decompose(const Nullkiller * ai) const
 	auto & developmentInfos = ai->buildAnalyzer->getDevelopmentInfo();
 	auto isGoldPressureLow = !ai->buildAnalyzer->isGoldPressureHigh();
 
+	ai->dangerHitMap->updateHitMap();
```
nullkiller, line 52: May be quite expensive. Treats should not change significantly during our turn. Only in case some enemy hero was killed.

AI/Nullkiller/Behaviors/BuildingBehavior.cpp
```diff
@@ -49,26 +49,49 @@ Goals::TGoalVec BuildingBehavior::decompose(const Nullkiller * ai) const
...
 				{
-					if(ai->getLockedResources().canAfford(buildingInfo.buildCost))
-						continue;
-
-					Composition composition;
+					tasks.push_back(sptr(BuildThis(buildingInfo, developmentInfo)));
+					emergencyDefense = true;
+				}
+			}
+		}
```
nullkiller, line 72: Possibly you can make it a part of DefenceBehavior. Also if you want to emphasize importance of it you may want to build a composition of DefendTown + BuildThis. DefendTown will act as priority booster here.
So all this makes sense only if there is hero or army defending town. Otherwise you just help enemy. Another option in terms of defense is building a dwelling and buying troops depending on enemy strength.
Btw it is also possible to build a composition of DefendTown + ExecuteHeroChain + BuildThis. So that we build Citadel or Castle only when we see a way to actually protect our town.
FORT is not a protection structure for AI now. Only a waste of money. Later possibly if BattleAI can use fort as an obstacle for attacker but this is unlikely in any nearest future.

AI/Nullkiller/Behaviors/BuyArmyBehavior.cpp
```diff
@@ -28,29 +28,34 @@ Goals::TGoalVec BuyArmyBehavior::decompose(const Nullkiller * ai) const
...
 	if(heroes.empty())
 	{
 		return tasks;
 	}
 
+	ai->dangerHitMap->updateHitMap();
+
 	for(auto town : cb->getTownsInfo())
 	{
+		uint8_t closestThreat = UINT8_MAX;
```
nullkiller, line 42: closestTreatTurn

nullkiller: dangerHitMap also can give treat node for town->visitablePos() and it has fastestDanger. For towns only it also gives all possible treats. For all the rest tiles only fastest and strongest

AI/Nullkiller/Behaviors/RecruitHeroBehavior.cpp
```diff
@@ -45,50 +46,68 @@ Goals::TGoalVec RecruitHeroBehavior::decompose(const Nullkiller * ai) const
...
-			int treasureSourcesCount = 0;
-
-			for(auto obj : ai->objectClusterizer->getNearbyObjects())
-			{
-				if((obj->ID == Obj::RESOURCE)
-					|| obj->ID == Obj::TREASURE_CHEST
-					|| obj->ID == Obj::CAMPFIRE
-					|| isWeeklyRevisitable(ai, obj)
-					|| obj->ID ==Obj::ARTIFACT)
+				score *= hero->getArmyCost();
```
nullkiller, line 87: This may make severe difference between heroes with cheap and expensive 1-unit stack while hero with cheaper stack may have better skills. Moreover it may prefer buying hero with army over retreated main with arts and high level

AI/Nullkiller/Behaviors/RecruitHeroBehavior.cpp
```diff
@@ -45,50 +46,68 @@ Goals::TGoalVec RecruitHeroBehavior::decompose(const Nullkiller * ai) const
...
+		}
+		if (town->hasCapitol())
+			haveCapitol = true;
+	}
+	if (bestHeroToHire && bestTownToHireFrom)
+	{
+		if (ai->cb->getHeroesInfo().size() < ai->cb->getTownsInfo().size() + 1
+			|| (ai->getFreeResources()[EGameResID::GOLD] > 10000 && !ai->buildAnalyzer->isGoldPressureHigh() && haveCapitol))
+		{
+			tasks.push_back(Goals::sptr(Goals::RecruitHero(bestTownToHireFrom, bestHeroToHire).setpriority((float)3 / ourHeroes.size())));
```
nullkiller, line 110: 0-division possible

AI/Nullkiller/Behaviors/StayAtTownBehavior.cpp
```diff
@@ -39,24 +39,15 @@ Goals::TGoalVec StayAtTownBehavior::decompose(const Nullkiller * ai) const
...
-			continue;
-
 		ai->pathfinder->calculatePathInfo(paths, town->visitablePos());
 
 		for(auto & path : paths)
 		{
 			if(town->visitingHero && town->visitingHero.get() != path.targetHero)
 				continue;
 
-			if(!path.targetHero->hasSpellbook() || path.targetHero->mana >= 0.75f * path.targetHero->manaLimit())
```
nullkiller, line 52: This was about mana regen using mages guild in town. Sure it may be worth to somehow estimate magic power in order to do this at all.

AI/Nullkiller/Engine/Nullkiller.cpp
```diff
@@ -62,7 +62,7 @@ bool canUseOpenMap(std::shared_ptr<CCallback> cb, PlayerColor playerID)
 		return false;
 	}
 
-	return cb->getStartInfo()->difficulty >= 3;
```
nullkiller, line 65: Can we get it back?

AI/Nullkiller/Engine/Nullkiller.cpp
```diff
@@ -554,4 +599,102 @@ void Nullkiller::lockResources(const TResources & res)
...
+				{
+					float ratio = available[i];
+					if (required[i] > 0)
+						ratio = static_cast<float>(available[i]) / required[i];
+					else
+						ratio = available[i];
+
+					bool okToSell = false;
+
+					if (i == 6)
```
nullkiller, line 658: Res::GOLD or whatever?

AI/Nullkiller/Behaviors/BuildingBehavior.cpp
Xilmi: void DangerHitMapAnalyzer::updateHitMap()
{
	if(hitMapUpToDate)
		return;

has this at right the beginning. So it won't really do anything if it ran already. it only gets updated here because it wasn't updated before. This then also means it doesn't need to be updated again later.

AI/Nullkiller/Behaviors/StayAtTownBehavior.cpp
Xilmi: I know and it will still be used for that. However, I also wanted a fallback for heros to go to towns when they don't have anything better to do. So I made it create tasks for that here for everything.

AI/Nullkiller/Engine/Nullkiller.cpp
Xilmi: Using GameResId::GOLD now.

AI/Nullkiller/Engine/Nullkiller.cpp
Xilmi: Ok, put it back.

AI/Nullkiller/Behaviors/RecruitHeroBehavior.cpp
Xilmi: Now dividing by (ourHeroes.size() + 1).

AI/Nullkiller/Behaviors/RecruitHeroBehavior.cpp
Xilmi: How about I do the following:
I count the currentArmyCost while iterating over our existing heros and then do this instead:

score *= (hero->getArmyCost() + currentArmyValue);

So at the beginning of the game the heros army cost will be significant.
For example if he has 1000 armycost and we only have 2000 current army, then it will have 50% higher score than comparable hero without army.

However, when we have like 40k army then the score-difference due to bringing a little bit of army himself won't really have a huge impact on the score and it should prefer hero with higher stats.

AI/BattleAI/BattleEvaluator.cpp
Xilmi: No idea. :o

nullkiller: Just to ensure .at(0) will not crash

AI/Nullkiller/Behaviors/RecruitHeroBehavior.cpp
nullkiller: Sounds better

AI/BattleAI/BattleEvaluator.cpp
Xilmi: Okay, I check ps.dest.empty() and in this case give another message without a location.

AI/Nullkiller/Behaviors/BuildingBehavior.cpp
Xilmi: I removed the fort from valid emergencyDefenses. I agree that it doesn't make enough of a difference to warrant building it.
The problem of making it a composition of that is that currently building is processed before everything else. So if the BuildingBehavior doesn't know about the defending also wanting to get a building, it will still fill the tasks with other building-projects.
So the difficult part would be: How to make sure it doesn't build something else instead when the BuildingBehavior doesn't know about it?
I mean it's possible to find a better solution but it's not trivial.

nullkiller: Defense will work only in slow evaluation loop while building behavior acts earlier in fast tasks loop. So yes you are right, without removing this fast task loop all together it will not be possible easily

AI/Nullkiller/Behaviors/BuyArmyBehavior.cpp
nullkiller: Should work, is not it?

AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
```diff
@@ -73,11 +72,18 @@ void BuildAnalyzer::updateOtherBuildings(TownDevelopmentInfo & developmentInfo)
 
 	if(developmentInfo.existingDwellings.size() >= 2 && ai->cb->getDate(Date::DAY_OF_WEEK) > boost::date_time::Friday)
 	{
-		otherBuildings.push_back({BuildingID::CITADEL, BuildingID::CASTLE});
 		otherBuildings.push_back({BuildingID::HORDE_1});
 		otherBuildings.push_back({BuildingID::HORDE_2});
```
nullkiller, line 76: Maybe for hordes these conditions have even less sense so you can remove them and add hordes unconditionally

Laserlicht:
Another thing that would be great to test:
We have the possibility for 8 creatures in town now. Not sure if AI can use 8th creature yet.
Can be tested with HotA factory town.

PR #4292, opened 2024-07-18
NKAI: reduce double army loss cases

nullkiller:
Known issue plaguing AI for long time. Sometimes AI can count some dangerous object twice doubling its army loss. Fixed a few such cases.

### DIFF
--- a/client/ClientCommandManager.cpp
+++ b/client/ClientCommandManager.cpp
@@ -89,7 +89,7 @@ void ClientCommandManager::handleGoSoloCommand()
 		// unlikely it will work but just in case to be consistent
 		for(auto & color : CSH->getAllClientPlayers(CSH->logicConnection->connectionID))
 		{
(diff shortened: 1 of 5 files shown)


client/ClientCommandManager.cpp
```diff
@@ -89,7 +89,7 @@ void ClientCommandManager::handleGoSoloCommand()
 		// unlikely it will work but just in case to be consistent
 		for(auto & color : CSH->getAllClientPlayers(CSH->logicConnection->connectionID))
 		{
-			if(CSH->client->getStartInfo()->playerInfos.at(color).isControlledByHuman())
+			if(color.isValidPlayer() && CSH->client->getStartInfo()->playerInfos.at(color).isControlledByHuman())
```
nullkiller, line 92: Regression from previous PR #4289

PR #4294, opened 2024-07-18
vcmilauncher: Use consistent margins

Alexander-Wilms:
Before:

image

After:

image


DjWarmonger:
To be honest, it took me a while to even spot the difference and I'm usually sensitive to any asymmetry :P

Install Icon is a nice touch, tho.

kambala-decapitator:
will test on macOS

Alexander-Wilms:
Hm, it breaks the layout on Windows:

Before (73ea0ce7bc999640e1fccb51b8adda4f4cb3c208):

image


After:

image

kambala-decapitator:
> Hm, it breaks the layout on Windows:

that's exactly why I wanted to test on macOS :)

kambala-decapitator:
yeah, not nice on macOS either

Screenshot 2024-07-18 at 20 24 13

PR #4299, opened 2024-07-19
input mode detection

Laserlicht:
- detects which method is used for last action (mouse+keyboard, touch, gamepad)
- show spell shortcut label only when mouse+keyboard is used
- disable haptic feedback when playing with other input than touch
- show touch tutorial only if touch used

### DIFF
--- a/client/battle/BattleInterfaceClasses.cpp
+++ b/client/battle/BattleInterfaceClasses.cpp
@@ -499,6 +499,7 @@ void QuickSpellPanel::show(Canvas & to)
 void QuickSpellPanel::inputModiChanged(InputModi modi)
 {
 	create();
+	redraw();
 }
 
 HeroInfoBasicPanel::HeroInfoBasicPanel(const InfoAboutHero & hero, Point * position, bool initializeBackground)
--- a/client/battle/BattleInterfaceClasses.cpp
+++ b/client/battle/BattleInterfaceClasses.cpp
@@ -483,7 +483,7 @@ void QuickSpellPanel::create()
 		{
 			buttonsDisabled.push_back(std::make_shared<TransparentFilledRectangle>(Rect(2, 7 + 50 * i, 48, 36), ColorRGBA(0, 0, 0, 172)));
 		}
-		if(GH.input().getCurrentInputMode() == InputMode::MOUSE)
+		if(GH.input().getCurrentInputMode() == InputMode::KEYBOARD_AND_MOUSE)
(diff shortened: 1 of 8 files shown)


client/eventsSDL/InputHandler.h
```diff
@@ -23,6 +23,13 @@ class InputSourceTouch;
 class InputSourceText;
 class InputSourceGameController;
 
+enum class InputModi
```
Alexander-Wilms, line 26:
```suggestion
enum class InputModus
```

I think it's more common to use the singular for enum names.

Laserlicht: thanks :)

Alexander-Wilms: Turns out 'modus' isn't used in English...

Laserlicht: Mode?

IvanSavenko: 'Mode' is probably better. This is also what we use in code

Laserlicht: Done :)

client/battle/BattleInterfaceClasses.cpp
```diff
@@ -483,7 +483,8 @@ void QuickSpellPanel::create()
 		{
 			buttonsDisabled.push_back(std::make_shared<TransparentFilledRectangle>(Rect(2, 7 + 50 * i, 48, 36), ColorRGBA(0, 0, 0, 172)));
 		}
-		labels.push_back(std::make_shared<CLabel>(7, 10 + 50 * i, EFonts::FONT_TINY, ETextAlignment::TOPLEFT, Colors::WHITE, config["keyboard"]["battleSpellShortcut" + std::to_string(i)].String()));
+		if(GH.input().getCurrentInputMode() == InputMode::MOUSE)
```
IvanSavenko, line 486: You're actually checking for keyboard here, not for mouse.
What if player connects external mouse to his phone? Or an opposite case - external keyboard?

Although not sure what's the proper solution here. Might be better to leave as it, at least until we actually run into such issue.

Laserlicht: Yeah, I know. Currently I assume Keyboard and Mouse are combined. Should be 99.9% of the playerbase.

Theoretically it would be also possible to play with touch and keyboard (or gameplay and mouse 🤣).

Only disadvantage currently is using mouse and touch combined showing the numbers.

But no idea how to improve. Would leave as it is for now.

IvanSavenko: In that case - let's just rename this input mode to KEYBOARD_AND_MOUSE or something like that.

Current code looks a bit weird even without accounting for these edge cases - we check for MOUSE input mode and proceed to show keyboard hotkeys.

Laserlicht: Yes, sure. Done. :)

PR #4306, opened 2024-07-19
BattleAI: fix bait for archers when need to go long way

nullkiller:
Attempt to fix bait when stack goes around long obstacle while it is faster to kill shielding creature


dydzio0614:
I would say to wait patiently with merging this until collecting some practical cases and then being able to compare results, because we have too much "anecdotal evidence" without saves/videos

Especially since i dont think this PR is prone to running into merge conflicts

nullkiller:
Yes, this one needs more testing.

PR #4308, opened 2024-07-19
MSVC CI script: Don't run vswhere to find dumpbin.exe, since it is no…

Alexander-Wilms:
…t used

Just the search took 3 to 15 min

Or is it used somewhere and I missed it?


nullkiller:
If nsis installer (msvc alternative build) is ok I believe it is not used

kambala-decapitator:
yes, seems no longer used. You can also check blame to find when and why it was introduced.

Issue #3109, opened 2023-10-26
fails to build on armel with undefined reference to `__atomic_fetch_add_8'


josch:
When compiling latest vcmi under Debian unstable on armel it fails with:

```
[ 65%] Linking CXX executable ../bin/vcmiserver
cd /<<PKGBUILDDIR>>/obj-arm-linux-gnueabi/server && /usr/bin/cmake -E cmake_link_script CMakeFiles/vcmiserver.dir/link.txt --verbose=1
[...]

PR #4309, opened 2024-07-20
On armel link with -latomic if necessary

josch:
Otherwise, the build will fail with:

   /usr/bin/ld: ../bin/libvcmi.so: undefined reference to `__atomic_fetch_add_8'
   /usr/bin/ld: ../bin/libvcmi.so: undefined reference to `__atomic_load_8'

This problem only happens on architectures without native atomics, so this might also happen on parisc or sh architectures.

Other projects are taking very similar approaches: https://github.com/google/highway/pull/1008

Closes: https://github.com/vcmi/vcmi/issues/3109

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -632,6 +632,23 @@ if(ENABLE_LAUNCHER OR ENABLE_EDITOR)
 	set(qtBinDir "${qtDir}/bin")
 endif()
 
+# link with -latomic if necessary
+# on armel the build will otherwise fail with
+# /usr/bin/ld: ../bin/libvcmi.so: undefined reference to `__atomic_fetch_add_8'
+# /usr/bin/ld: ../bin/libvcmi.so: undefined reference to `__atomic_load_8'
+if(CMAKE_LIBRARY_ARCHITECTURE STREQUAL "arm-linux-gnueabi")
+	file(WRITE ${CMAKE_BINARY_DIR}/test_atomics.cpp
+	"#include <stdint.h>\nint main(void){uint64_t x=0,y=0;return (int)__atomic_fetch_add_8(&x,y,0);}\n")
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
the right way is to pass such system-specific flags from outside (e.g. from preset) instead of polluting project's CML. I know that VCMI's CML already contains some hardcoded flags, but would be great to stop doing that. Please try to pass the flags you need externally: if it works, then the PR can be closed.

IvanSavenko:
Have to agree with kambala.
While I am not against accepting this change, I don't really understand why this is our issue.

Why should VCMI be the one to specify dependencies of standard libraries? Shouldn't environment or CMake be the one to specify this?

We don't specify dependencies of SDL or ffmpeg for example (which are many)

josch:
Thank you for your feedback!

Yes, flags can obviously passed from outside as well. But only a few cmake targets seem to require `-latomic`. If I pass flags from the outside, they will be applied to the whole project. Since only libvcmi and vcmiserver need this, it seemed like this is something that the vcmi build system has to take care of.

Undoubtedly, it would be better if vcmi would not have to carry the knowledge that libvcmi and vcmiserver need `-latomic` on armel. But I also struggle to find the right place in the stack which should carry this fix. If CMake itself should carry it, then which component would need it? Is it a library used by vcmi which needs adjustments?

Maybe you have some ideas, thank you! :)

kambala-decapitator:
VCMI itself doesn't use libatomic directly. iirc it's common for Linux-based distros to require it when you use atomics from C or C++ standard library. Normally it should come from toolchain or dependencies.

you could try using https://cmake.org/cmake/help/latest/envvar/LDFLAGS.html (or more specific variables described there), maybe specifying it globally would work. Otherwise you'd need to expose per-target cache variables to pass custom flags from outside.

PR #4314, opened 2024-07-20
Implemented simple target selection logic for arrow towers

IvanSavenko:
Apparently when hero can't control arrow towers their target selection logic would be completely empty - it would simply select first alive unit. End of logic.

Now automatic arrow towers will use following priorities for target selection:
1. Enemies inside fort walls
2. Enemy shooters
3. Melee units outside walls (everyone else actually)

If there are multiple potential targets within same group, such as multiple shooters, towers would target most dangerous target (which is ai value * stack size)


IvanSavenko:
In that case I can simply remove it from considerations. Don't want too complex logic in there.

Issue #1070, opened 2022-10-15
[FEATURE] Possibility to use HD textures


Laserlicht:
It would be great to be able to use HD textures. Which are already available through the official HD version. Or through mods then.

In my opinion, this would be a great advantage that vcmi would bring over other possibilities.

One possibility would be if you could specify to render the game for example in 2x. And then upscale each texture before output (different modes: linear, cubic, xbrz, etc.).
Except of course the HD textures. Those would be scaled up / down with a different scaling factor (we could specify that in config.json).

But the better way would be complete prescaled textures (for slow upscaling algorithm).

See also here for similar approch in fheroes2: https://github.com/ihhub/fheroes2/discussions/5016 / https://github.com/ihhub/fheroes2/pull/6438 (also hardcoded coordinates)

Some example with ESRGAN (Textures must be upscaled before, regardless of the game -> ESRGAN is slow, and for good results every texture should upscaled seperatly). Please see in Fullscreen, not in the tiny Github Preview.

Left part of the picture is scaled, right the original:
![4x_BSRGAN_town](https://user-images.githubusercontent.com/13953785/211052258-8c50eafb-5307-4942-8db6-49e5fec42b97.jpg)
![4x_BSRGAN_town2](https://user-images.githubusercontent.com/13953785/211054986-e18bfe82-3bd4-4de8-84c1-4924e08ed19a.jpg)

angel in different variants:
![comp_angel](https://user-images.githubusercontent.com/13953785/212484403-8c8571cf-e6e2-4cfe-a772-44fa85fce177.png)

town image original and ESRGAN 
![town_image](https://user-images.githubusercontent.com/13953785/211059775-588904b7-5293-488f-98fb-488ba0ab6bc0.png)

misiokles:
HD graphics are minor problem in this case. What about new values for coordinates and cursor positions almost in every aspects in the game? They are hardcoded and vary when resolution is changed.

Laserlicht:
Yes, this is probably not an easy task. The biggest problem is not even the hardcoded coordinates, I think, but that there is no abstraction layer to sdl.

Otherwise you could extend the render functions: For Scale factor 2 multiply all coordinates (and non hd-textures with filter) by 2. Of course, for hd textures it must be specified by which factor they are larger than the original textures.

But it would be great if heroes iii would get a decent hd version.

Laserlicht:
Here's an idea of how this could be implemented:

[...]

PR #4318, opened 2024-07-21
Support for xBRZ upscaling filter

### LINKED ISSUE #3108: [HDmod] Screen filter
**Describe your proposal**
Add screen filters like in HD mod:
image

Especially xBRZ.

**Additional context**
Veroks implementation isn't lost. Here is the solution for OH3: https://github.com/Nemirtingas/HeroesGL/tree/master (but not the latest 1.65).
He solved it with GLSL Shaders.

Another similar thing: [zelda3](https://github.com/snesrev/zelda3/tree/master/src) uses also SDL and can use xBRZ as GLSL shader.

I've also made some tests: https://github.com/Laserlicht/vcmi/tree/xbrz_test and https://github.com/Laserlicht/vcmi/tree/xbrz_hack but software based solution is to slow for real time. I'm not able to do a hardware based implementation.

Maybe a step before #1070 because it may much easier to implement.

Maybe we can use this example? https://github.com/AugustoRuiz/sdl2glsl/blob/master/src/main.cpp
Together with this shader? https://github.com/libretro/glsl-shaders/blob/master/xbrz/shaders/xbrz-freescale.glsl

IvanSavenko:
WIP, mostly to indicate that I am working on it. Somewhat playable, but still a lot to do.
- Resolves #3108

With this change vcmi will:
- upscale all loaded images by a given factor using xBRZ library
- entire game will now be rendered at specified factor (e.g. x2)
- SDL can still perform up or downscaling, so it is possible to play with x2 scaling even if your display is less than 1600x1200 (e.g. typical FullHD display)

Since all images are upscaled on load, size of pixel data in memory will grow by A LOT - 8-bit h3 images will be converted to 32-bit at double resolution - so could be up to 16 times compared to before. However VCMI uses memory not just for images, so total memory usage won't grow as much.

#### Done
- [x] Implement input remapping according to scaling factor
- [x] Scale bitmap h3 fonts 
- [x] Render ttf fonts at specified scaling, bypassing xBRZ
- [x] Fix broken dialog border rendering
- [x] Fix adventure map rendering
- [x] Check adventure map scaling
- [x] Fix minimap on map preview window
- [x] Fix minimap in game
- [x] Fix minimap radar border
- [x] Fix video playback
- [x] Check combat animations
- [x] Implement player-coloring for UI
- [x] Implement glue selection border for battles
- [x] Implement player-coloring for map objects
- [x] Investigate not working movement range / cursor highlight in battles

#### Known bugs
- [x] Fix palette artifacts on 'grass' battlefield background
- [ ] Fix computation of TTF string length (likely rounding artifact since chars width can now be half of a pixel)
- [x] Clean up CSDL_Ext::scaleSurface method, use CSDL_Ext::scaleSurfaceIntegerFactor directly
- [ ] Fix FPS indicator
- [x] Fix spellbook page flip animation
- [ ] Fix large spellbook scaling
- [x] Fix cursor size calculation
- [x] Canvas::drawLine should produce wider lines depending on scaling
- [ ] On wide-screen resolution section to the left from resource panel is not player-colored

#### Missing features
- [ ] Implement berserk & clone effect
- [ ] Implement petrification grayscale effect
- [ ] Implement terrain animations
- [ ] Check whether fade-in effect for summons in combat works
- [ ] Fix colorization of stack amount box in combat in upscaled mode

#### Configuration 
- [x] Allow disabling xBRZ
- [x] Allow configuring base scale factor (2, 3, 4?, 5? ,6?)
- [ ] Allow configuring filter for adventure map zoom
- [ ] Allow configuring cursor scale factor
- [ ] Allow configuring xBRZ parameters
- [ ] Add 'interesting' parameters from this list to Launcher

#### Optimizations
- [x] use TBB for multithreading upscaling of large images, e.g. dialog backgrounds
- [ ] ~use TBB for multithreading upscaling of video frames using sws_scale method~. Not possible or requires investigation. zBRZ upscaling for video might be better option
- [x] Reduce image dimensions before upscaling to avoid high memory usage by empty areas
- [ ] Investigate memory & CPU usage changes

#### Optional features
- [ ] review scaling filters available in sws_getContext and check if there are algorithms better than bicubic
- [ ] use xBRZ when zooming in on adventure map
- [ ] use xBRZ for video upscaling. May be too CPU expensive though.
- [ ] Review places where we use image scaling and see if they can benefit from adjustments
- [ ] Consider whether it makes sense / not too complex to implement scaling of minimap from map size to UI size using xBRZ
- [ ] Provide a way to load pre-scaled images in place of xBRZ
- [ ] Consider integration of TTF fonts as default for upscaled mode


Laserlicht:
related:  #3108

IvanSavenko:
Current very WIP screenshots:
(adventure map is completely broken, so no screenshots with it)
xbrz-battle
xbrz-pregame
xbrz-town

IvanSavenko:
Core functionality is done - fonts, adventure map, battles, video, input. So somewhat playable now. There is noticeable lag in debug build when game upscales new images, will check release mode later.

Now need to look into various palette-based effects - player coloring, terrain animation, etc.

Laserlicht:
Nice!

The great thing with this approach is, that we can combine the really clear True type fonts with upscaled graphics. That's something you cannot do with the HD mod.

IvanSavenko:
Yeah. Frankly not sure whether we should keep using H3 fonts for upscaled mode or switch to ttf. For SD mode h3 fonts are better, but for upscaled - ttf is clearly an improvement over upscaled h3 fonts

Another improvement is minimap - in HD mod it is a bit blurry, and there is some loss of details in maps larger than XL. While in vcmi minimap is fully scaling-aware.

Laserlicht:
> Yeah. Frankly not sure whether we should keep using H3 fonts for upscaled mode or switch to ttf. For SD mode h3 fonts are better, but for upscaled - ttf is clearly an improvement over upscaled h3 fonts

Yes, I think it would be okay for scaled mode as default. Maybe a config to change back?

> Another improvement is minimap - in HD mod it is a bit blurry, and there is some loss of details in maps larger than XL. While in vcmi minimap is fully scaling-aware.

Looks good. :)

I think especially asian fonts benefit from this PR. Some asian character are very detailed. Too much for our current resolutions.

Laserlicht:
Tested it also on smartphone. Performance is good. :)
Still a lot of work (many graphical bugs, I think e.g. spellbook is missing in your list), but great approach!

`Provide a way to load pre-scaled images in place of xBRZ`

This would be a killer feature. Especially if they can loaded via mod. :)
I've also already scaled "HD" images on disk for test in the past. See #1070. [Upscayl](https://upscayl.org/) does a great job here.

IvanSavenko:
Will recheck spellbook. Was there anything else? I am not aware of any bugs other than palette effects at the moment, which are:
- terrain/river animaitions
- player-colored UI
- flags for owned objects
- selection highlight in battles
- clone/berserk effect

Loading upscaled images - yeah, that's pretty much the goal. We have our own assets (e.g. icons) that we can easily provide in higher definitions. Modders might be able to provide their own mods with high-definition assets.
Per-scaling using some external tool might also be doable, although at least in my experiments with stable diffusion results were not exactly amazing, even with ESRGAN - too many scaling artifacts.

dydzio0614:
users may want to pre-scale with AI on their own and load these assets into game

IvanSavenko:
It was mostly about whether it makes sense to make such a tool.
Played a bit with stable diffusion myself - town screens are the only decent result.
Large adventure map objects are... bearable, but probably still behind xBRZ
But for creatures in battle and smaller adventure map objects - AI-upscaled version is very poor. To a point where I don't see need to bother with AI upscale support (even if user-provided tool)

IvanSavenko:
For content provided by mods - sure

But stuff like this:
> further photoshopped by some modder if needed etc.

is definitely no for H3 assets.

Laserlicht:
> Was there anything else? 

Will look into it again.

> It was mostly about whether it makes sense to make such a tool.

If it's not too complicated it would great if we still able to use hires replacements for existing (non mod content) graphics e.g. with a mod.
Such a script can use different methods for different types of graphics. Or maybe scale only town images? Or make an converter that converts official HD graphics to vcmi mod. Or maybe we got better AI to upscale in future?

But yes. A HD pack (with official graphics) we shouldn't provide by our self because of copyright.

altiereslima:
Could you also add Integer Scaling?

IvanSavenko:
> Could you also add Integer Scaling?

Should be already possible in 1.5:
- set resolution scale to 200%
- set SDL screen filter option (already present in Launcher) to 'nearest'

However you need at least 1200px-high screen for it. So 1080p / FullHD is not enough. You need 1440p or 2160p monitor.

IvanSavenko:
Will probably fix any regressions, merge and finish all remaining tasks separately

PR #4323, opened 2024-07-22
Battle ai fixes

nullkiller:
Feels like it is ready for review.
Managed to fix retaliation issue in dragon breath mechanics when dragon hits dragon back from top. Compared also behavior with original HoMM3. See unit tests.
Now AI also correctly calculates dragon breath and dragon breath retaliation.
Waited attack fixed.


lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1303,8 +1303,8 @@ AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes(const battle:
 
 				// if targeted double-wide creature is attacked from above or below ( -> second hex is also adjacent to attack origin)
 				// then dragon breath should target tile on the opposite side of targeted creature
-				if (BattleHex::mutualPosition(attackOriginHex, secondHex) != BattleHex::NONE)
-					nextHex = secondHex.cloneInDirection(direction, false);
+				/*if(BattleHex::mutualPosition(attackOriginHex, secondHex) != BattleHex::NONE)
```
nullkiller, line 1306: The most interesting part is this logic. If I uncomment it the client sees dragon breath hitting creatures which server live untouched.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1235,19 +1235,40 @@ ReachabilityInfo CBattleInfoCallback::getFlyingReachability(const ReachabilityIn
...
+		destinationTile,
+		attackerPos,
+		defender->getPosition());
+}
+
+AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes(
+	const battle::Unit* attacker,
+	const battle::Unit * defender,
+	BattleHex destinationTile,
+	BattleHex attackerPos,
```
nullkiller, line 1260: Both attackerPos and defenderPos can be hypothetic now. Attacker position is always hypothetical when AI calculates possibilities but also when attacker receives retaliation it becomes defender. In this case we need hypothetic defender position

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1291,19 +1312,26 @@ AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes(const battle:
 	else if(attacker->hasBonusOfType(BonusType::TWO_HEX_ATTACK_BREATH))
 	{
 		auto direction = BattleHex::mutualPosition(attackOriginHex, destinationTile);
+		
+		if(direction == BattleHex::NONE
```
nullkiller, line 1316: Bug in mechanics.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1248,19 +1248,40 @@ ReachabilityInfo CBattleInfoCallback::getFlyingReachability(const ReachabilityIn
...
-AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes(const battle::Unit* attacker, BattleHex destinationTile, BattleHex attackerPos) const
+AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes(
+	const battle::Unit * attacker,
+	BattleHex destinationTile,
+	BattleHex attackerPos) const
+{
+	const auto * defender = battleGetUnitByPos(destinationTile, true);
+
+	if(!defender)
+		return AttackableTiles(); // can't attack thin air
```
IvanSavenko, line 1259: Why we even have this check? Might hide incorrect calls, e.g. potential bug in caller method.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1304,19 +1325,26 @@ AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes(const battle:
...
 			{
-				auto secondHex = destinationTile == defender->getPosition() ?
-					defender->occupiedHex():
-					defender->getPosition();
+				auto secondHex = destinationTile == defenderPos ? defender->occupiedHex(defenderPos) : defenderPos;
 
 				// if targeted double-wide creature is attacked from above or below ( -> second hex is also adjacent to attack origin)
 				// then dragon breath should target tile on the opposite side of targeted creature
-				if (BattleHex::mutualPosition(attackOriginHex, secondHex) != BattleHex::NONE)
+				if(BattleHex::mutualPosition(attackOriginHex, secondHex) != BattleHex::NONE)
```
IvanSavenko, line 1347: I think we talked on Discord that this might not be correct? Or was it confirmed to be correct?

lib/battle/CBattleInfoCallback.cpp
nullkiller: Might be for case when defender position is hypothetic. Most likely not needed after latest change. It is not new but moved from below.

lib/battle/CBattleInfoCallback.cpp
nullkiller: You can see unit tests. Verified according to original h3 as well.

PR #4328, opened 2024-07-24
NKAI: some changes and fixes to AI logic using one-way monoliths

MichalZr6:
- Allow AI to use one-way teleports (monoliths) when it has no town. The risk of being trapped is irrelevant in this case.
- Fix chat error appearing when AI was stepping off the exit for the case of one-way entrance having multiple exits:

obraz

- Replacing `ui32` with `int32_t` for answers in ...DialogsAnswered() functions.


MichalZr6:
@IvanSavenko @nullkiller Could you please review my commits when you have time? Thanks!

PS: For the changes regarding the use of one-way teleports when the AI is townless, I implemented it differently than you proposed on Discord. The approach via PathfinderSettings required many changes, so I opted for a simpler solution, which hopefully is also good enough as no changes had to be done with CPathfinderHelper const qualifiers.

lib/pathfinder/PathfinderOptions.cpp
```diff
@@ -44,6 +45,18 @@ PathfinderConfig::PathfinderConfig(std::shared_ptr<INodeStorage> nodeStorage, st
...
+	// FIXME: should be enabled by default with some bigger AI refactor so that main heroes are strong enough to
+	// go through and/or enable scouts (at some point ?) to visit such monoliths
+	// see first TODO in the top of PathfinderOptions.h
+	if (hero->cb->howManyTowns(hero->getOwner()) == 0)
+	{
+		options.useTeleportOneWay = true;
+		options.useTeleportOneWayRandom = true;
+	}
+}
+
```
IvanSavenko, line 59: 1) Should be in AI code. I believe AI configures its pathfinder parameters somewhere. Should be somewhere in there
2) This would not detect case if this particular hero is in small room with one-way teleport BUT has town someplace else

lib/pathfinder/PathfinderOptions.cpp
```diff
@@ -60,6 +73,7 @@ SingleHeroPathfinderConfig::~SingleHeroPathfinderConfig() = default;
 SingleHeroPathfinderConfig::SingleHeroPathfinderConfig(CPathsInfo & out, CGameState * gs, const CGHeroInstance * hero)
 	: PathfinderConfig(std::make_shared<NodeStorage>(out, hero), buildRuleSet())
 {
+	checkIfTownlessAndAllowOneWayTeleports(hero);
```
IvanSavenko, line 76: I think this would also affect pathfinder for human player

MichalZr6: I tried implementing the changes via `PathfinderSettings`, as nullkiller suggested on Discord, but after making numerous changes in various places, it didn't work. This is because `SingleHeroPathfindingConfig` inherits from `PathfinderConfig`. Should I try another round on this ? I'm not an experienced C++ developer (well... not even a developer by trade, I code because I enjoy it), so if you could give me some hints, I will work in that direction to do it the right way ;)

lib/mapObjects/MiscObjects.cpp
```diff
@@ -490,7 +490,7 @@ void CGMonolith::onHeroVisit( const CGHeroInstance * h ) const
 	TeleportDialog td(h->id, channel);
 	if(isEntrance())
 	{
-		if(cb->isTeleportChannelBidirectional(channel) && 1 < cb->getTeleportChannelExits(channel).size())
+		if(cb->getTeleportChannelExits(channel).size() > 1)
```
nullkiller, line 493: Unlikely we should change server code to allow AI going through one-way monoliths. Do we have monolith in AI path? If not you may want to make changes similar to Whirlpool. Most likely the one in AIPathNode::commit which sets theNodeBefore

lib/pathfinder/PathfinderOptions.cpp
nullkiller: Yes because AI needs to know that it needs to visit monolith first to continue path. This can be achieved by linking exit node to entrance one using theNodeBefore property. See AINodeStorage::commit for Whirlpool.

nullkiller: AIPath at the end should contain 2 nodes: monolith entrance, target tile

nullkiller: You can use some very simple map with oneway and town behind it for testing and temporarily change
#define AI_TRACE_LEVEL 2 in AINodeStorage.h to see more logs. There in logs you should see paths printed. Path to town should contain monolith entrance tile.

nullkiller (changes requested):

nullkiller:
Feel free to ask questions on discord ai-development

lib/mapObjects/MiscObjects.cpp
MichalZr6: This is more of a global and not only AI related issue. I saw the error message (screenshot in the top) some time ago already, while playing. oneWayTeleports (UniDirectional) can have multiple exits with one entrance. So the check with isTeleportChannelBidirectional failed and exit was taken randomly. Then client also used RNG and sometimes it hit the right one, and sometimes not. Am I understanding it correctly or the problem is somewhere else ?

nullkiller: Strange, I still do not understand how it works for you. First of all these settings should be false by default I beleave

> , useTeleportOneWay(VLC->settings()->getBoolean(EGameSettings::PATHFINDER_USE_MONOLITH_ONE_WAY_UNIQUE))
> , useTeleportOneWayRandom(VLC->settings()->getBoolean(EGameSettings::PATHFINDER_USE_MONOLITH_ONE_WAY_RANDOM))

MichalZr6: I'm totally surprised... Have to look into your commits because I reverted changes to initial code in terms of the above Bidirectional check and **it works without errors now**! Probably should've check before committing

nullkiller:
Regarding my PR for whirlpool and AI passing through it when it has multiple exits. Pathfinder instructs AI to first visit Whirlpool entrance. It expects that hero will be teleported to required exit and attempts to move further to the target destination. If random behavior choose wrong exit it will see that it can not rich target tile and discard path. Next pass it will build a new path through whirlpool and will attempt again and again until success. But if AI has whirlpool protection it will go directly to required exit because regular pathfinder allows it in this case.

PR #4330, opened 2024-07-25
check spellbook

### LINKED ISSUE #4326: Some bugs found in pvp tournament so far
Some bugs found in pvp tournament so far:

- [x] spellbook icon during battle should be inactive if spellbook not purchased (cannot cast anyway, the only way to access it in h3 in this case to get "you cannot cast spell" popup is to click hero)
- [x] thieves guild from adventure map does not work during opponent turn (G shortcut)
- [x] damage prediction does not update when swapping action from shoot to melee (half damage related)
- [ ] #4352
- [x] pandora box with spells does not have "nothing found" message if it contained spells with level too high for hero to learn
- [x] on receiving new turn notification, all windows other than adventure map must be closed
- [ ] System no longer reminds of timer when DENABLE_GOLDMASTER is enabled for vcmi release build

Since these bugs got found during pvp games, they are likely to improve pvp gameplay when fixed

Laserlicht:
partially fix for #4326

### DIFF
--- a/client/battle/BattleWindow.cpp
+++ b/client/battle/BattleWindow.cpp
@@ -218,7 +218,7 @@ void BattleWindow::showStickyQuickSpellWindow()
 	Settings showStickyQuickSpellWindow = settings.write["battle"]["enableQuickSpellPanel"];
 	showStickyQuickSpellWindow->Bool() = true;
 
-	if(GH.screenDimensions().x >= 1050)
+	if(GH.screenDimensions().x >= 1050 && owner.getBattle()->battleGetMyHero()->hasSpellbook())
 	{
 		quickSpellWindow->enable();
 		quickSpellWindow->isEnabled = true;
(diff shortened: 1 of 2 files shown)


client/battle/BattleWindow.cpp
```diff
@@ -838,7 +838,7 @@ void BattleWindow::blockUI(bool on)
 		ESpellCastProblem spellcastingProblem = owner.getBattle()->battleCanCastSpell(hero, spells::Mode::HERO);
 
 		//if magic is blocked, we leave button active, so the message can be displayed after button click
-		canCastSpells = spellcastingProblem == ESpellCastProblem::OK || spellcastingProblem == ESpellCastProblem::MAGIC_IS_BLOCKED;
+		canCastSpells = (spellcastingProblem == ESpellCastProblem::OK || spellcastingProblem == ESpellCastProblem::MAGIC_IS_BLOCKED) && hero->hasSpellbook();
```
IvanSavenko, line 841: Strange. We have ESpellCastProblem::NO_SPELLBOOK, but for some reason it was not handled in here?

PR #4332, opened 2024-07-26
fix campaign highscores

### LINKED ISSUE #4180: Campaign score should be the sum of each level
While finish the campaign, only one level score will be shown on High Score menu.

For example, from campaign Hack and Slash, I can get 2000 score for campaign in OH3, but only 483 in VCMI.

**Version**
 - OS: Windows, Android
 - Version: 1.5.3

Laserlicht:
fixes #4180

### DIFF
--- a/lib/gameState/HighScore.h
+++ b/lib/gameState/HighScore.h
@@ -19,7 +19,7 @@ public:
 	int townAmount;
 	bool usedCheat;
 	bool hasGrail;
-	bool allDefeated;
+	bool allEnemiesDefeated;
 	std::string campaignName;
 	std::string scenarioName;
 	std::string playerName;
@@ -31,7 +31,7 @@ public:
 		h & townAmount;
 		h & usedCheat;
 		h & hasGrail;
-		h & allDefeated;
+		h & allEnemiesDefeated;
 		h & campaignName;
 		h & scenarioName;
 		h & playerName;
--- a/lib/gameState/HighScore.h
+++ b/lib/gameState/HighScore.h
@@ -1,0 +1,41 @@
+/*
+ * HighScore.h, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#pragma once
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+class DLL_LINKAGE HighScoreParameter
+{
+public:
+	int difficulty;
+	int day;
+	int townAmount;
+	bool usedCheat;
+	bool hasGrail;
+	bool allDefeated;
+	std::string campaignName;
+	std::string scenarioName;
+	std::string playerName;
+
+	template <typename Handler> void serialize(Handler &h)
+	{
+		h & difficulty;
+		h & day;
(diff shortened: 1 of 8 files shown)


Laserlicht:
@IvanSavenko Can you look over it? I do not really understand, when "Forward declaration" works and when not.
It's also compiler specific. It compiles locally but not in CI especially with android compiler.

IvanSavenko:
> @IvanSavenko Can you look over it? I do not really understand, when "Forward declaration" works and when not. 

- Function parameter - forward declaration is enough
- Direct member of a class/struct (not pointer / reference) - requires include
- Pointer/reference member - forward declaration
- Various containers - this one is rather compiler-specific, so easier to use include
- shared_ptr - forward declaration
- unique_ptr - forward declaration works, but requires instantiation of `= default`-ed destructor in cpp

So in your case - you should be able to use forward-declaration in CServerHandler, but needs direct include for `std::vector<HighScoreParameter>` in CampaignState

Laserlicht:
> So in your case - you should be able to use forward-declaration in CServerHandler, but needs direct include for `std::vector<HighScoreParameter>` in CampaignState

Yes. Was my first test like this. Compiles under Linux.

But Android has an problem: `arithmetic on a pointer to an incomplete type 'HighScoreParameter'`

I don't know, how to fix.

Laserlicht:
`error: reference to 'HighScoreParameter' is ambiguous` - a nightmare. :D

IvanSavenko:
Library declaration must be wrapped in macro:
```
VCMI_LIB_NAMESPACE_BEGIN
struct HighScoreParameter;
VCMI_LIB_NAMESPACE_END
```

lib/gameState/HighScore.h
```diff
@@ -0,0 +1,41 @@
...
+
+class DLL_LINKAGE HighScoreParameter
+{
+public:
+	int difficulty;
+	int day;
+	int townAmount;
+	bool usedCheat;
+	bool hasGrail;
+	bool allDefeated;
```
IvanSavenko, line 22:
```suggestion
	bool allEnemiesDefeated;
```

lib/gameState/HighScore.h
```diff
@@ -0,0 +1,41 @@
...
+VCMI_LIB_NAMESPACE_BEGIN
+
+class DLL_LINKAGE HighScoreParameter
+{
+public:
+	int difficulty;
+	int day;
+	int townAmount;
+	bool usedCheat;
+	bool hasGrail;
```
IvanSavenko, line 21: Just in case - can you double-check that this flag works? There was report somewhere that it was broken (unconfirmed)

Laserlicht: Checked following:
- no grail
- grail in hero inventar
- grail in hero inventar + hero in town
- grail as building in town

For the three last you will get the correct 25 points. Seems correct. Cannot replicate the reported error.

Just as reference, report was:

> Now for my question: the last game I played was a random map generated with a Spider template. Free for all, me and 7 npc:s.
> First time the game ended I did not dig the grail and got some amount of points. Then I decided to try and find the grail from the map so I loaded an autosave, dug around and found it pretty fast and ended the game. The score, however, did not go up as I think it should have gone. Instead, I lost 3 points from the final score which is due to the couple of days of digging.

PR #4334, opened 2024-07-26
quickspell improvements

Laserlicht:
- autofill with spells available by current hero (algorithm can improved by separate PR; current approach: use first spells)
- custom spells selected from spellbook (as before) are marked with orange square now -> these are fixed for all heroes
  -> can be removed with opening spellbook and canceling
- checkbox in spellbook selection to switch between all spells and spells available to hero (default)

grafik
grafik


client/battle/BattleInterfaceClasses.cpp
```diff
@@ -433,6 +433,49 @@ QuickSpellPanel::QuickSpellPanel(BattleInterface & owner)
...
+
+	// autofill empty slots with spells if possible
+	auto hero = owner.getBattle()->battleGetMyHero();
+	for(int i = 0; i < QUICKSPELL_SLOTS; i++)
+	{
+		if(spellIds[i] != SpellID::NONE)
+			continue;
+
+		for(auto & availableSpell : CGI->spellh->objects)
+		{
```
IvanSavenko, line 464: Please avoid direct access to 'objects' member of our handlers.
This should work, if I haven't missed something:
```suggestion
		for(const auto & availableSpellID : CGI->spellh->getDefaultAllowed())
		{
			const auto * availableSpell = availableSpellID->toSpell();
```

Laserlicht:
@IvanSavenko merge?

PR #4340, opened 2024-07-29
[1.5.6?] Fixes for issues with multiplayer

IvanSavenko:
Not sure whether to release this as 1.5.6, or whether to release 1.5.6 at all. But at the very least it should be possible for tournament players to use this build in place of 1.5.5

Fixed issues reported in #4326:
- Do not hide spells that hero can't learn on receiving them from rewardable object, e.g. Pandora.
  - Also, spells that can not be learned will now be displayed with gray text in name
- Fixed some shortcuts not active during enemy turn, such as thieves guild
- Fixed ranged mode always used for damage preview, even when forcing melee attack
- Close all open dialogs on start of our turn, to prevent weird bugs like locked right-click popups

Other safe fixes:
- Fixed typo where incorrect game settings was used for dwelling recruitment to full garrison
- Fixed possible crash on invalid SPELL_LIKE_ATTACK bonus
- Added compatibility check for loading maps with old names for boats

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -169,40 +169,44 @@ void CPlayerInterface::initGameInterface(std::shared_ptr<Environment> ENV, std::
 	adventureInt.reset(new AdventureMapInterface());
 }
 
+void CPlayerInterface::closeAllDialogs()
+{
+	// remove all active dialogs that do not expect query answer
+	for (;;)
+	{
+		auto adventureWindow = GH.windows().topWindow<AdventureMapInterface>();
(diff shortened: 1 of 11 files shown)


dydzio0614:
one more thing that would be good to have: unlock UI for player to allow quitting game while not making turn etc. - for example if during simultaneous turns host ended turn and joined did not, then joiner losing connection makes host unable to quit game without alt+f4 etc.

DjWarmonger:
Looks like mod conflicts are puzzling players: https://github.com/vcmi/vcmi/issues/4342

It's a quite simple GUI feature as far as I see.

PR #4346, opened 2024-07-31
[1.5.6?] online lobby - mod compatibility improvements

### LINKED ISSUE #4342: Small QOL suggestion regarding mod conflicts in online lobby
I think, mod conflicts are very easy to miss, since all messages are in the same, yellow color.
I suggest to change font color or add red-ish background color for mods that are not compatible with the lobby.

I also suggest, making improvement where mod conflicts are at the top of the mod list.
Zrzut ekranu_20240730_130120

Laserlicht:
Fixes #4342

- compatible -> gray
- version conflict -> yellow
- other error -> red
- order for status: -> errors first -> then for name

grafik


IvanSavenko:
> other than compatible -> red

Perhaps yellow for version mismatch?

Laserlicht:
> > other than compatible -> red
> 
> Perhaps yellow for version mismatch?

Version mismatch isn't a problem?

IvanSavenko:
Mod version mismatch *may* be a problem. Mods use different versioning schemes, and are very inconsistent with savegame / network breakages.

So indicating it is probably a good idea.

Oh, but we already use yellow in there. So: 
- red - error, can't join (already done in your PR)
- yellow - warning, version mismatch, may work, may not work
- gray - OK, compatible

Laserlicht:
Aye! :)

PR #4348, opened 2024-08-01
Hill Fort unavailable upgrade and status bar new messages

MichalZr6:
Some players get confused when they can't upgrade their units in Hota's (aka wooden or mini) Hill Fort. They get rather misleading message that the creature is already upgraded, when in fact it is in its basic level.
This PR adds new message for that case and also shows some additional info on the status bar about the building.

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -517,7 +517,7 @@
 	"core.seerhut.quest.reachDate.visit.4" : "Closed till %s.",
 	"core.seerhut.quest.reachDate.visit.5" : "Closed till %s.",
 	
-	"mapObject.core.hillFort.object.description" : "Upgrades creatures. Levels 1 - 4 are less expensive than in associated town",
+	"mapObject.core.hillFort.object.description" : "Upgrades creatures. Levels 1 - 4 are less expensive than in associated town.",
 	
 	"core.bonus.ADDITIONAL_ATTACK.name": "Double Strike",
 	"core.bonus.ADDITIONAL_ATTACK.description": "Attacks twice",
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -283,9 +283,6 @@
 
 	"vcmi.adventureMap.revisitObject.hover" : "Revisit Object",
 	"vcmi.adventureMap.revisitObject.help" : "{Revisit Object}\n\nIf a hero currently stands on a Map Object, he can revisit the location.",
-	"vcmi.adventureMap.miniHillFort.notAvailableUpgrade.message" : "This building cannot upgrade creatures of level 5 and higher.",
-	"vcmi.adventureMap.miniHillFort.statusBar.info" : "Upgrade creatures levels 1 - 5 at double the normal price",
-	"vcmi.adventureMap.HillFort.statusBar.info" : "Upgrades creatures. Levels 1 - 4 are less expensive than in associated town",
 
 	"vcmi.battleWindow.pressKeyToSkipIntro" : "Press any key to start battle immediately",
 	"vcmi.battleWindow.damageEstimation.melee" : "Attack %CREATURE (%DAMAGE).",
@@ -519,7 +516,9 @@
 	"core.seerhut.quest.reachDate.visit.3" : "Closed till %s.",
 	"core.seerhut.quest.reachDate.visit.4" : "Closed till %s.",
 	"core.seerhut.quest.reachDate.visit.5" : "Closed till %s.",
-
+	
+	"mapObject.core.hillFort.object.description" : "Upgrades creatures. Levels 1 - 4 are less expensive than in associated town",
+	
 	"core.bonus.ADDITIONAL_ATTACK.name": "Double Strike",
 	"core.bonus.ADDITIONAL_ATTACK.description": "Attacks twice",
 	"core.bonus.ADDITIONAL_RETALIATION.name": "Additional retaliations",
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -283,6 +283,9 @@
 
 	"vcmi.adventureMap.revisitObject.hover" : "Revisit Object",
 	"vcmi.adventureMap.revisitObject.help" : "{Revisit Object}\n\nIf a hero currently stands on a Map Object, he can revisit the location.",
+	"vcmi.adventureMap.miniHillFort.notAvailableUpgrade.message" : "This building cannot upgrade creatures of level 5 and higher.",
+	"vcmi.adventureMap.miniHillFort.statusBar.info" : "Upgrade creatures levels 1 - 5 at double the normal price",
+	"vcmi.adventureMap.HillFort.statusBar.info" : "Upgrades creatures. Levels 1 - 4 are less expensive than in associated town",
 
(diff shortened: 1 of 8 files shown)


client/windows/GUIClasses.h
```diff
@@ -443,9 +443,11 @@ class CGarrisonWindow : public CWindowObject, public IGarrisonHolder
 class CHillFortWindow : public CStatusbarWindow, public IGarrisonHolder
 {
 private:
-	static const int slotsCount = 7;
+
+	enum class State { UNAFORDABLE, ALREADY_UPGRADED, MAKE_UPGRADE, EMPTY, UNAVAILABLE };
```
dydzio0614, line 447: UNAFORDABLE SHOULD BE UNAFFORDABLE (typo)

client/windows/GUIClasses.cpp
```diff
@@ -1129,6 +1129,12 @@ CHillFortWindow::CHillFortWindow(const CGHeroInstance * visitor, const CGObjectI
 	statusbar = CGStatusBar::create(std::make_shared<CPicture>(background->getSurface(), Rect(8, pos.h - 26, pos.w - 16, 19), 8, pos.h - 26));
 
 	garr = std::make_shared<CGarrisonInt>(Point(108, 60), 18, Point(), hero, nullptr);
+
+	if(object->typeName == "miniHillFort")
+		statusbar->write(VLC->generaltexth->translate("vcmi.adventureMap.miniHillFort.statusBar.info"));
+	else if(object->typeName == "hillFort")
+		statusbar->write(VLC->generaltexth->translate("vcmi.adventureMap.HillFort.statusBar.info"));
+
```
IvanSavenko, line 1137: Should be done in config:
- for H3 fort - in `config/objects/generic.json`
- for hota fort - in hota mod.

Game should:
- register this string in HillFortInstanceConstructor::initTypeData
- set this string to HillFort instance HillFortInstanceConstructor::initializeObject

Check CRewardableConstructor for similar code.

client/windows/GUIClasses.cpp
```diff
@@ -1281,22 +1305,28 @@ std::string CHillFortWindow::getTextForSlot(SlotID slot)
...
-	if(hero->slotEmpty(slot))//no creature here
-		return -1;
+	if(hero->slotEmpty(slot))
+		return State::EMPTY;
 
 	UpgradeInfo info;
 	LOCPLINT->cb->fillUpgradeInfo(hero, slot, info);
-	if(!info.newID.size())//already upgraded
-		return 1;
+	if (!info.newID.size())
```
IvanSavenko, line 1317:
```suggestion
	if (info.newID.empty())
```

client/windows/GUIClasses.cpp
```diff
@@ -1281,22 +1305,28 @@ std::string CHillFortWindow::getTextForSlot(SlotID slot)
...
+	{
+		// new Hill Fort allows upgrades level 5 and below
+		if (hero->getStack(slot).type->getLevel() >= 5 && hero->getCreature(slot)->hasUpgrades())
```
IvanSavenko, line 1320: Why do you need check for level here?
If list of potential upgrades is empty, but creature has upgrades then upgrade is unavailable.
Also, hill fort level limits (and costs) are configurable, nothing stops modders from making hill fort with configuration different from hota
```suggestion
		// Hill Fort may limit level of upgradeable creatures, e.g. mini hill fort from hota
		if (hero->getCreature(slot)->hasUpgrades())
```

Mods/vcmi/config/vcmi/english.json
```diff
@@ -243,6 +243,9 @@
 
 	"vcmi.adventureMap.revisitObject.hover" : "Revisit Object",
 	"vcmi.adventureMap.revisitObject.help" : "{Revisit Object}\n\nIf a hero currently stands on a Map Object, he can revisit the location.",
+	"vcmi.adventureMap.miniHillFort.notAvailableUpgrade.message" : "This building cannot upgrade creatures of level 5 and higher.",
+	"vcmi.adventureMap.miniHillFort.statusBar.info" : "Upgrade creatures levels 1 - 5 at double the normal price",
```
IvanSavenko, line 247: These two strings should be in hota mod, not in our source code.

MichalZr6:
Thanks for the remarks. Will fix :)

MichalZr6:
BTW: found a bug in miniHillFort.json: it allows to upgrade lvl 5 creatures while it shouldn't (checked with original hota).

MichalZr6:
@IvanSavenko @dydzio0614 
Had to do some cleanup with commits and now I think this PR is ready for review.
Dydzio, please check if polish strings are ok.
I will also make separate PR for HOTA if this one gets accepted.
PS: It would be super nice if the description on status bar is not flushed right after onHover text was there. But is it worth the effort ?

Mods/vcmi/config/vcmi/polish.json
```diff
@@ -481,7 +481,7 @@
 	"core.seerhut.quest.reachDate.visit.3" : "Zamknięte do %s.",
 	"core.seerhut.quest.reachDate.visit.4" : "Zamknięte do %s.",
 	"core.seerhut.quest.reachDate.visit.5" : "Zamknięte do %s.",
-
+	"mapObject.core.hillFort.object.description" : "Ulepsza jednostki. Koszt ulepszenia dla poziomów 1 - 4 są bardziej korzystne niż w mieście",
```
dydzio0614, line 484: should be "Koszt ulepszenia dla poziomów 1 - 4 jest bardziej korzystny"

test/googletest
IvanSavenko, line 1: Revert this submodule change please

IvanSavenko (approved):

test/googletest
MichalZr6: Done :)

lib/mapObjectConstructors/HillFortInstanceConstructor.cpp
```diff
@@ -11,17 +11,27 @@
...
 {
 	parameters = config;
+	VLC->generaltexth->registerString(parameters.getModScope(), TextIdentifier(getBaseTextID(), "unavailableUpgradeMessage"), parameters["unavailableUpgradeMessage"].String());
+	VLC->generaltexth->registerString(parameters.getModScope(), TextIdentifier(getBaseTextID(), "description"), parameters["description"].String());
 }
 
 void HillFortInstanceConstructor::initializeObject(HillFort * fort) const
 {
 	fort->upgradeCostPercentage = parameters["upgradeCostFactor"].convertTo<std::vector<int>>();
+	fort->descriptionToolTip = VLC->generaltexth->translate(TextIdentifier(getBaseTextID(), "description").get());
```
IvanSavenko, line 28: Translation should be done on client side, not on gamestate initialization.

lib/mapObjects/MiscObjects.h
```diff
@@ -459,10 +461,22 @@ class DLL_LINKAGE HillFort : public CGObjectInstance, public ICreatureUpgrader
...
+	{
+		return unavailableUpgradeMessage;
+	}
+
 	template <typename Handler> void serialize(Handler &h)
 	{
 		h & static_cast<CGObjectInstance&>(*this);
 		h & upgradeCostPercentage;
+		h & descriptionToolTip;
+		h & unavailableUpgradeMessage;
```
IvanSavenko, line 479: This would break loading of 1.5 saves.

Perhaps instead of storing strings, HillFort should return string ID generated in runtime? 
Something like this:
`return TextIdentifier(getHandler()->getBaseTextID(), "description").get();`

IvanSavenko:
Merged, sorry for delay.

PR #4350, opened 2024-08-01
make armageddon can effect to allied creatures in ENCHANTER bonus

kdmcser:
in ENCHANTER bonus, when spell is armageddon, only empty will be effected. So I not setting massive and smart mark when spell ID is armageddon.

### DIFF
--- a/server/battles/BattleFlowProcessor.cpp
+++ b/server/battles/BattleFlowProcessor.cpp
@@ -326,8 +326,10 @@ void BattleFlowProcessor::activateNextStack(const CBattleInfoCallback & battle)
 
 		if (!tryMakeAutomaticAction(battle, next))
 		{
-			setActiveStack(battle, next);
-			break;
+			if(next->alive()) {
+				setActiveStack(battle, next);
+				break;
+			}
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


server/battles/BattleFlowProcessor.cpp
```diff
@@ -757,8 +759,11 @@ void BattleFlowProcessor::stackTurnTrigger(const CBattleInfoCallback & battle, c
 				});
 				spells::BattleCast parameters(&battle, st, spells::Mode::ENCHANTER, spell);
 				parameters.setSpellLevel(bonus->val);
-				parameters.massive = true;
-				parameters.smart = true;
+				//todo: not hardcode
+				if (spellID != SpellID::ARMAGEDDON) {
+					parameters.massive = true;
+					parameters.smart = true;
+				}
```
IvanSavenko, line 766: This is not a fix but ugly workaround.
What if mod adds another spell similar to Armageddon? 
Are you sure that Armageddon is the only such spell, what about others? Like Death Ripple or possible some non-damaging spells.
What's so special in Armageddon that requires hardcoding it?

kdmcser: so check the the range of the spell, if a damage spell which range is X, and smart is false, not set its smart and massive mark to true?

kdmcser:
move to #4399

PR #4358, opened 2024-08-01
Add helper functions for integer division rounding

IvanSavenko:
Added set of functions that perform integer division with different rounding modes:
- divideAndCeil - rounds up to next integer
- divideAndRound - rounds to nearest integer
- divideAndFloor - rounds to previous integer (equivalent to default division)

Intended for use in library, where usage of floating point might lead to desync in multiplayer games.

Replaced some cases that I knew of, including recent handicap PR

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -700,6 +700,33 @@ namespace vstd
 		return a + (b - a) * f;
 	}
 
+	/// Divides dividend by divisor and rounds result up
+	/// For use with integer-only arithmetic
+	template<typename Integer1, typename Integer2>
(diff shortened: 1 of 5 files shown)


Global.h
```diff
@@ -700,6 +700,33 @@ namespace vstd
 		return a + (b - a) * f;
 	}
 
+	/// Divides divident by divisor and rounds result up
```
Alexander-Wilms, line 703:
```suggestion
	/// Divides dividend by divisor and rounds result up
```

nullkiller (approved):

PR #4359, opened 2024-08-01
Writing some statistics and make it possible to export via csv

Laserlicht:
Current table columns:

```
Map
Timestamp
Day
Player
Team
IsHuman
Status
NumberHeroes
NumberTowns
NumberArtifacts
ArmyStrength
Income
MapVisitedRatio
ObeliskVisited
MightMagicRatio
Score
MaxHeroLevel
NumBattlesNeutral
NumBattlesPlayer
NumWinBattlesNeutral
NumWinBattlesPlayer 
```
and Resources and num of mines for resources.

Export functionality to CSV. For further analysis e.g. with Excel or Python. Want to add more values. What would be interesting?

Export can be triggered by host with: `!statistic`

@IvanSavenko Would this approach be okay?


IvanSavenko:
Looks OK so far.

One suggestion - completely remove statsHLP and move all remaining methods from it to statistics.

Laserlicht:
I think ready for now. May add some interesting values in seperate PR.

Don't want to make it too big (caused by refactoring).

Laserlicht:
Example of with a few clicks generated diagrams in libreoffice from this csv export.
Auto battle with AI (using autoskip and strong hero in own town).

One player:

grafik
grafik
grafik
grafik
grafik
grafik

comparing player:
grafik
grafik

lib/gameState/GameStatistics.h
```diff
@@ -0,0 +1,134 @@
...
+	TeamID team;
+	bool isHuman;
+	EPlayerStatus status;
+	TResources resources;
+	int numberHeroes;
+	int numberTowns;
+	int numberArtifacts;
+	si64 armyStrength;
+	int income;
+	double mapVisitedRatio;
```
IvanSavenko, line 37: 1. Explored, not visited.
2. Float is precise enough even for G+U maps
```suggestion
	float mapExploredRatio;
```

lib/gameState/GameStatistics.h
```diff
@@ -0,0 +1,134 @@
...
+    std::string toCsv();
+
+	struct ValueStorage
+	{
+		std::map<PlayerColor, int> numBattlesNeutral;
+		std::map<PlayerColor, int> numBattlesPlayer;
+		std::map<PlayerColor, int> numWinBattlesNeutral;
+		std::map<PlayerColor, int> numWinBattlesPlayer;
+		std::map<PlayerColor, int> numHeroSurrendered;
+		std::map<PlayerColor, int> numHeroEscaped;
```
IvanSavenko, line 96: How are these field different from fields in StatisticDataSetEntry?

server/CGameHandler.cpp
```diff
@@ -669,6 +669,19 @@ void CGameHandler::onPlayerTurnEnded(PlayerColor which)
 		heroPool->onNewWeek(which);
 }
 
+void CGameHandler::addStatistics()
+{
+	for (auto & elem : gs->players)
+	{
+		if (elem.first == PlayerColor::NEUTRAL || !elem.first.isValidPlayer())
+			continue;
```
IvanSavenko, line 677: I think neither case is possible here

lib/gameState/GameStatistics.h
```diff
@@ -0,0 +1,134 @@
...
+		h & data;
+		h & values;
+	}
+};
+
+class DLL_LINKAGE Statistic
+{
+	static std::vector<const CGMine *> getMines(const CGameState * gs, const PlayerState * ps);
+public:
+	using TStat = std::pair<PlayerColor, si64>;
```
IvanSavenko, line 121: Find better name? Or add comment what 'stat' is this?

lib/gameState/GameStatistics.cpp
```diff
@@ -0,0 +1,328 @@
...
+double Statistic::getMightMagicRatio(const PlayerState * ps)
+{
+	double numMight = 0;
+
+	for(auto h : ps->heroes)
+		if(h->type->heroClass->affinity == CHeroClass::EClassAffinity::MIGHT)
+			numMight++;
+
+	return numMight / ps->heroes.size();
+}
```
IvanSavenko, line 313: Not sure if this stat is 'interesting' in any way. Especially since it does not accounts for hero strength.
If I am playing with Crag Hack who gets all army & exp and bunch of lvl 1 magic heroes-scouts my 'might magic ratio' would lean to magic.

Might be easier to remove it probably.
Asking pvp players for what stats they might be interested will likely give you better ideas than this one.

lib/gameState/GameStatistics.h
```diff
@@ -0,0 +1,134 @@
...
+	bool isHuman;
+	EPlayerStatus status;
+	TResources resources;
+	int numberHeroes;
+	int numberTowns;
+	int numberArtifacts;
+	si64 armyStrength;
+	int income;
+	double mapVisitedRatio;
+	int obeliskVisited;
```
IvanSavenko, line 38: Perhaps save as ratio? 10 visited obelisks on map with 48 obelisks is way different from map with 10 obelisks
```suggestion
	float obeliskVisitedRatio;
```

server/CGameHandler.cpp
Laserlicht: In the parent function both there are checked on other place: https://github.com/vcmi/vcmi/blob/c8b48318b7a511c5a9a26812d4b9c26494a18f9b/server/CGameHandler.cpp#L759

lib/gameState/GameStatistics.h
Laserlicht: Removed and used `std::pair<PlayerColor, si64>` directly. To make it consistent (function calling wasn't using TStat)

DjWarmonger:
Oh, this could be helpful for machine learning. Finally some feedback on AI performance :)

lib/gameState/GameStatistics.cpp
Laserlicht: Yes, I've noticed that too. :)
Removed

lib/gameState/GameStatistics.h
Laserlicht: The difference is the interval/point in time.

The data from these fields can no longer be retrieved retrospectively, as it has not been saved since then. The data set represents the current values.

Whereas StatisticDataSetEntry contains the complete data set at the end of a turn.

Perhaps this can also be put together somehow. But I think it's much cleaner this way.

Laserlicht:
> Oh, this could be helpful for machine learning. Finally some feedback on AI performance :)

Yes. And some data for data analysts/scientists. :)

Do other interesting values come to mind?

DjWarmonger:
- Artifact count / total artifact value (notice that in lategame artifacts can be traded for resources)
- Total experience on all heroes

Laserlicht:
> * Artifact count / total artifact value (notice that in lategame artifacts can be traded for resources)

Not sure about this.

> * Total experience on all heroes
> 
>  * Flagged dwellings

added

>  * Flagged mines

Already saved (separate for each resource)

Laserlicht:
added also:
- TownBuiltRatio
- HasGrail
- SpentResourcesForArmy (each resource)
- SpentResourcesForBuildings (each resource)
- TradeVolume (each resource)

lib/gameState/GameStatistics.cpp
```diff
@@ -0,0 +1,383 @@
...
+	
+	return tmp;
+}
+
+// get total gold income
+int Statistic::getIncome(const CGameState * gs, const PlayerState * ps)
+{
+	int percentIncome = gs->getStartInfo()->getIthPlayersSettings(ps->color).handicap.percentIncome;
+	int totalIncome = 0;
+	const CGObjectInstance * heroOrTown = nullptr;
```
IvanSavenko, line 239: No longer used? Only assigned twice

lib/gameState/GameStatistics.h
```diff
@@ -0,0 +1,156 @@
...
+		std::map<PlayerColor, int> numBattlesNeutral;
+		std::map<PlayerColor, int> numBattlesPlayer;
+		std::map<PlayerColor, int> numWinBattlesNeutral;
+		std::map<PlayerColor, int> numWinBattlesPlayer;
+		std::map<PlayerColor, int> numHeroSurrendered;
+		std::map<PlayerColor, int> numHeroEscaped;
+		std::map<PlayerColor, TResources> spentResourcesForArmy;
+		std::map<PlayerColor, TResources> spentResourcesForBuildings;
+		std::map<PlayerColor, TResources> tradeVolume;
+		std::map<PlayerColor, si64> movementPointsUsed;
```
IvanSavenko, line 114: - Perhaps add "Acumulated" into name?
- Perhaps replace with a single map?
```cpp
struct PlayerAccumulatedValueStorage
{
    int numBattlesNeutral;
    int numBattlesPlayer;
    ...
}
std::map<PlayerColor, PlayerAccumulatedValueStorage> accumulatedValues;
```

PR #4384, opened 2024-08-04
add shortcut for list item ordering

Laserlicht:
for hero and town list:
- add possibility to move with keyboard and mouse
  - hover item with mouse and press PageUp, PageDown, End, Pos1 to order
- make it possible to dismiss hero with Del in list
- fix bug of not updating list when ordering in town screen


client/adventureMap/CList.cpp
```diff
@@ -301,6 +305,72 @@ void CHeroList::CHeroItem::gesture(bool on, const Point & initialPosition, const
...
+		ony += [=](){ LOCPLINT->cb->dismissHero(hero); };
+		LOCPLINT->showYesNoDialog(CGI->generaltexth->allTexts[22], ony, nullptr);
+		return;
+	}
+
+	if(heroes.size() < 2)
+		return;
+
+	size_t heroPos = vstd::find_pos(heroes, hero);
+	const CGHeroInstance * heroUpper = (heroPos < 1) ? nullptr : heroes.at(heroPos - 1);
```
nullkiller, line 345: Can we use boolean isHeroUpper here instead? Or can heroes.at(index) contain nullptr itself?

client/adventureMap/CList.cpp
```diff
@@ -301,6 +305,72 @@ void CHeroList::CHeroItem::gesture(bool on, const Point & initialPosition, const
...
+		return;
+
+	if(!pos.isInside(GH.getCursorPosition()))
+		return;
+
+	auto & heroes = LOCPLINT->localState->getWanderingHeroes();
+
+	if(key == EShortcut::LIST_HERO_DELETE)
+	{
+		CFunctionList<void()> ony = [=](){ };
```
nullkiller, line 335: What does ony mean?

Laserlicht:
> Perhaps instead of targeting hovered hero/town instead apply shortcuts on selected object? Should simplify code a bit and will make it possible to use these shortcuts without mouse

The problem here is that this shortcuts overlaps with adventure map shortcuts. With the current approach no problem.

client/adventureMap/CList.cpp
Laserlicht: Not sure. Used the same line as for the gesture here. To be safe.

client/adventureMap/CList.cpp
Laserlicht: I think `on yes`. Also used in another place in VCMI code base like this. Removed. As it doesn't make sense to store lambda in var.

IvanSavenko:
> The problem here is that this shortcuts overlaps with adventure map shortcuts. With the current approach no problem.

If player already moved mouse he might just as well use mouse wheel. 

Perhaps assign to free hotkeys, e.g. Ctrl+PageUp / Ctrl+Up / etc? For me it is weird to have keyboard hotkeys that require mouse positioning.

Laserlicht:
Changed to Ctrl+...

client/gui/Shortcut.h
```diff
@@ -293,5 +293,15 @@ enum class EShortcut
 	SPELLBOOK_TAB_ADVENTURE,
 	SPELLBOOK_TAB_COMBAT,
 
+	LIST_HERO_UP,
+	LIST_HERO_DOWN,
+	LIST_HERO_TOP,
+	LIST_HERO_BOTTOM,
+	LIST_HERO_DELETE,
```
IvanSavenko, line 300:
```suggestion
	LIST_HERO_DISMISS,
```

IvanSavenko (approved):

PR #4391, opened 2024-08-04
[HotA] Bank support

Laserlicht:
Bank support for HotA 1.7

add in config: `"type" : "bank",` to `special1`


CCallback.cpp
```diff
@@ -217,6 +217,16 @@ bool CCallback::buildBuilding(const CGTownInstance *town, BuildingID buildingID)
 	return true;
 }
 
+bool CCallback::triggerTownSpecialBuildingAction(const CGTownInstance *town, BuildingSubID::EBuildingSubID subBuildingID)
+{
+	if(town->tempOwner!=player)
+		return false;
+
```
IvanSavenko, line 224: Should probably be handled in UI?

CCallback.cpp
```diff
@@ -217,6 +217,16 @@ bool CCallback::buildBuilding(const CGTownInstance *town, BuildingID buildingID)
 	return true;
 }
 
+bool CCallback::triggerTownSpecialBuildingAction(const CGTownInstance *town, BuildingSubID::EBuildingSubID subBuildingID)
```
IvanSavenko, line 220: We might also want to extend this logic to mana whirlpool, similar to what hota did with it.
Or even allow it for any building, e.g. flag in building config, so you can make explicitly visitable stables if you want to

client/windows/CCastleInterface.cpp
```diff
@@ -1054,6 +1058,21 @@ void CCastleBuildings::enterAnyThievesGuild()
 	LOCPLINT->showInfoDialog(CGI->generaltexth->translate("vcmi.adventureMap.noTownWithTavern"));
 }
 
+void CCastleBuildings::enterBank()
+{
+	std::vector<std::shared_ptr<CComponent>> components;
+	if(town->bonusValue.second > 0)
+	{
+		components.push_back(std::make_shared<CComponent>(ComponentType::RESOURCE, GameResID(GameResID::GOLD), "-500/" + CGI->generaltexth->translate("core.genrltxt.64")));
```
IvanSavenko, line 1066: There is `ComponentType::RESOURCE_PER_DAY` for this purpose if I understand correctly

server/CGameHandler.cpp
```diff
@@ -2469,6 +2469,30 @@ bool CGameHandler::buildStructure(ObjectInstanceID tid, BuildingID requestedID,
...
 }
 
+bool CGameHandler::triggerTownSpecialBuildingAction(ObjectInstanceID tid, BuildingSubID::EBuildingSubID sid)
+{
+	const CGTownInstance * t = getTown(tid);
+
+	bool hasBuilding = false;
+	for (auto building : t->town->buildings)
+		if(vstd::contains(t->builtBuildings, building.first) && building.second->subId == sid)
+			hasBuilding = true;
```
IvanSavenko, line 2479: use `CTown::getBuildingType` instead?

CCallback.cpp
Laserlicht: adapted this logic from `CCallback::buildBuilding`

server/CGameHandler.cpp
```diff
@@ -2469,6 +2469,30 @@ bool CGameHandler::buildStructure(ObjectInstanceID tid, BuildingID requestedID,
...
+		return false;
+
+	if(sid == BuildingSubID::EBuildingSubID::BANK)
+	{
+		TResources res;
+		res[EGameResID::GOLD] = 2500;
+		giveResources(t->getOwner(), res);
+
+		setObjPropertyValue(t->id, ObjProperty::BONUS_VALUE_SECOND, 2500);
+	}
```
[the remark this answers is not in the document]
Laserlicht: Maybe rather an task for later refactoring?
It already contains handling of several BuildingSubID's and BuildingID's.

CCallback.cpp
Laserlicht: Yes, there's nothing wrong with extending it?

server/CGameHandler.cpp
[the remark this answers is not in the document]
IvanSavenko: Will probably do this myself in separate PR

CCallback.cpp
IvanSavenko: Yeah, should be fine allowing this for all buildings.

Explicit visit can also be used to fix our old feature-bug where stables & such are visited on new week automatically, without even opening town screen

dydzio0614: well there are 3 types of visit that make sense:
1. automatic (like now) - it is kinda QoL that buildings like stables work without opening town screen
2. on opening town screen (original h3 mana vortex)
3. hota 1.7 has its new manual handling of mana vortex

IvanSavenko: > automatic (like now) - it is kinda QoL 

Yeah, but sometimes result in unwanted message spam on new week, so kind of mixed feelings. Perhaps alternative approach would be to hide these messages.

PR #4393, opened 2024-08-04
[HotA] Grotto support

Laserlicht:
Support for grotto

Only one problem left. Assert triggered here: https://github.com/Laserlicht/vcmi/blob/8832a51071e80466bbc427870c9060a3b5dbb675/client/widgets/markets/CAltarArtifacts.cpp#L34

Creature works already correct.

Don't know how to fix - @SoundSSGood Can you give me a hint how to fix?


lib/constants/Enumerations.h
```diff
@@ -59,7 +60,7 @@ namespace BuildingSubID
 enum class EMarketMode : int8_t
 {
 	RESOURCE_RESOURCE, RESOURCE_PLAYER, CREATURE_RESOURCE, RESOURCE_ARTIFACT,
-	ARTIFACT_RESOURCE, ARTIFACT_EXP, CREATURE_EXP, CREATURE_UNDEAD, RESOURCE_SKILL,
+	ARTIFACT_RESOURCE, ARTIFACT_EXP, CREATURE_EXP, ARTIFACT_CREATURE_EXP, CREATURE_UNDEAD, RESOURCE_SKILL,
```
SoundSSGood, line 63: Do we really need once more market mode?

Laserlicht: No, but should be the cleanest way, I think?

Alternatives?

client/windows/CMarketWindow.cpp
```diff
@@ -151,13 +154,13 @@ void CMarketWindow::createChangeModeButtons(EMarketMode currentMode, const IMark
 	if(isButtonVisible(EMarketMode::CREATURE_EXP))
 	{
 		addButton(AnimationPath::builtin("ALTSACC.DEF"), CGI->generaltexth->zelp[572], std::bind(&CMarketWindow::createAltarCreatures, this, market, hero), EShortcut::MARKET_CREATURE_EXPERIENCE);
-		if(marketWidget->hero->getAlignment() == EAlignment::GOOD)
+		if(marketWidget->hero->getAlignment() == EAlignment::GOOD && mode != EMarketMode::ARTIFACT_CREATURE_EXP)
```
SoundSSGood, line 157: I just checked, the grotto in the city and the altar on the map work the same way. I believe that this change is necessary only for debugging?

Laserlicht: Yeah, you're right.

I assumed that all hero's alignment can do everything. But the grotto behaves like a normal altar of sacrifice on map. So I can greatly simplify this...

Laserlicht:
Only remaining question: How do we get the artifact thing to work.

SoundSSGood:
> Only remaining question: How do we get the artifact thing to work.

Just comment assert line. For now.

client/windows/CMarketWindow.h
```diff
@@ -43,6 +43,8 @@ class CMarketWindow final : public CStatusbarWindow, public CWindowWithArtifacts
 	const Point quitButtonPos = Point(516, 520);
 	std::shared_ptr<CMarketBase> marketWidget;
 
+	EMarketMode mode;
```
SoundSSGood, line 46:
```suggestion
```

SoundSSGood: No need at the moment. I think.

Laserlicht:
> > Only remaining question: How do we get the artifact thing to work.
> 
> Just comment assert line. For now.

No that wouldn't work...
We do a cast, that is needed: `auto altarObj = dynamic_cast<const CGArtifactsAltar*>(market);`. So we get an nullptr for this.

SoundSSGood:
> No that wouldn't work...

Ahh, I see. But unfortunately there is no quick fix. The town instance must be inherited from CArtifactSet. Like CGArtifactsAltar.

Laserlicht:
@SoundSSGood I tried it, but it doesn't work. :(

Would be great to support the building (last not supported form HotA yet).

Issue #4398, opened 2024-08-05
Allow setting hero spellpower and knowledge to 0


kdmcser:
**Describe your proposal**
I want to desire a powerful magic hero for dungeon. For balance, I want to set start spellpower and knowledge to 0 and he can not get any spellpowers and knowledges since level grows.

But VCMI limits minial spellpower and knowledge to 1, so can zero value be supported?

PR #4400, opened 2024-08-05
support configure for minimal primary skill values

kdmcser:
Since I request in #4398 , it is better for modders to design minimal primary skills, so I add 4 config in game_config.json to make modder can modify them.

- "minimalAttack": 0,
- "minimalDefence": 0,
- "minimalKnowledge": 1,
- "minimalSpellPower": 1


config/gameConfig.json
```diff
@@ -296,7 +296,12 @@
 			// number of artifacts that can fit in a backpack. -1 is unlimited.
 			"backpackSize"		: -1,
 			// if heroes are invitable in tavern
-			"tavernInvite"            : false
+			"tavernInvite"            : false,
+			// minimai primary skills for heroes
+			"minimalAttack": 0,
+			"minimalDefence": 0,
+			"minimalKnowledge": 1,
+			"minimalSpellPower": 1
```
IvanSavenko, line 304: No need for 4 separate values, especially if they are handled only as array
```suggestion
			"minimalPrimarySkills": [ 0, 0, 1, 1]
```

kdmcser:
@IvanSavenko I use array and commit my code

lib/MimialPrimarySkill.cpp
```diff
@@ -0,0 +1,34 @@
...
+};
+
+int getPrimarySkillMinimum(PrimarySkill pSkill)
+{
+    auto minialPSkills = VLC->settings()->getVector(EGameSettings::HEROES_MINIMAL_PRIMARY_SKILLS);
+    if(minialPSkills.size() != DEFAULT_MINIMAL_PSKILLS.size())
+        logGlobal->error("gameConfig.json: heroes/minimalPrimarySkills format error. need a vector with 4 elements.");
+    int index = PSKILL_INDEX_MAP.at(pSkill);
+    return minialPSkills.size() > index ? minialPSkills[index] : DEFAULT_MINIMAL_PSKILLS[index];
+}
```
IvanSavenko, line 32: All code from this file can be replaced with a single line, without any new free functions:
```cpp
VLC->settings()->getVector(EGameSettings::HEROES_MINIMAL_PRIMARY_SKILLS)[pSkill.getNum()]
```

1. PrimarySkill class has method getNum() that converts it to integer
2. We generally don't validate json via checks in code, only via schema validation (not present for game settings though, but that's separate issue)
3. We generally don't use free functions

kdmcser: I replace all getPrimarySkillMinimum to your suggestion, not using  free functions

Issue #4376, opened 2024-08-03
Carryover doesn't work in VCMI Campaign format


madaosoul:
**Describe the bug**
Finally got to test possibilities of vcmi campaign and discovered that heroes and creatures don't carry over to next scenario. The first one should be possible with the use of hero placeholders, but VCMI doesn't seem to register them. And the campaign format page on github states that line in the 00.json:
"keepCreatures": ["angel"],
should make the carryover of units possible, but it just doesn't work.

I attach a set of maps I made to test a bunch of stuff (packed into a campaign with all of the components separately). Orrin in Test 2 is a placeholder and should receive experience from Test 1 or Test 3.

**To Reproduce**
1. Play the custom campaign I attached.
2. Start from Test 1 or Test 3 and level up Orrin a bit (vcmiglorfindel).
3. Play Test 2 and notice that Orrin is missing.
4. Notice that you don't have more than 5 Angels in each scenario despite picking them as a bonus every time.

**Expected behavior**
Orrin's placeholder should receive all things from one of previous scenarios and be a playable hero. You should start your 2nd and 3rd scenario with 10 and 15 angels if you pick them each time.

**Actual behavior**
Orrin doesn't appear in Test 2 and you start each scenario with the same amount of 5 angels.

**Version**
 - OS: Windows 11
 - Version: VCMI 1.5.5, 1.6.0
[custom campaign test.zip](https://github.com/user-attachments/files/16480773/custom.campaign.test.zip)

by003:
I have tested campaign and sent it to Godric. Seems he has already found the conditions of crossover hero.

PR #4415, opened 2024-08-07
map editor: check hero and town spells during mod assessment

godric3:
While looking at some test campaigns (like one from #4376) I noticed that town possibleSpells list have all spells from all mods active while adding town. Map creator wasn't aware that there is some reference to mod in his map so it wasn't marked as required. This PR should both add error on validation and improve marking required mods based on map objects

### DIFF
--- a/lib/CPlayerState.cpp
+++ b/lib/CPlayerState.cpp
@@ -50,6 +50,11 @@ std::string PlayerState::getJsonKey() const
 	return color.toString();
 }
 
+std::string PlayerState::getModScope() const
+{
+	return "core";
+}
+
 std::string PlayerState::getNameTranslated() const
 {
 	return VLC->generaltexth->translate(getNameTextID());
--- a/lib/ObstacleHandler.cpp
+++ b/lib/ObstacleHandler.cpp
@@ -31,6 +31,11 @@ std::string ObstacleInfo::getJsonKey() const
 	return identifier;
 }
 
+std::string ObstacleInfo::getModScope() const
+{
+	return "core";
+}
+
 std::string ObstacleInfo::getNameTranslated() const
 {
 	return identifier;
--- a/lib/ObstacleHandler.cpp
+++ b/lib/ObstacleHandler.cpp
@@ -28,12 +28,12 @@ int32_t ObstacleInfo::getIconIndex() const
 
 std::string ObstacleInfo::getJsonKey() const
 {
-	return identifier;
+	return modScope + ':' + identifier;
 }
 
 std::string ObstacleInfo::getModScope() const
 {
-	return "core";
+	return modScope;
 }
 
 std::string ObstacleInfo::getNameTranslated() const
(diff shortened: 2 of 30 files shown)


mapeditor/mapcontroller.cpp
```diff
@@ -613,16 +613,49 @@ ModCompatibilityInfo MapController::modAssessmentAll()
...
 {
 	ModCompatibilityInfo result;
+
+	auto extractEntityMod = [&result](const auto & entity) 
+	{
+		auto modName = QString::fromStdString(entity->getJsonKey()).split(":").at(0).toStdString();
+		if(modName != "core")
+			result[modName] = VLC->modh->getModInfo(modName).getVerificationInfo();
+	};
+
```
IvanSavenko, line 623: Perhaps instead we should add `getModScope` method to entity class?
More clear and `getJsonKey()` is often implemented as `return modScope + ':' + identifier;`

lib/CPlayerState.cpp
```diff
@@ -50,6 +50,11 @@ std::string PlayerState::getJsonKey() const
 	return color.toString();
 }
 
+std::string PlayerState::getModScope() const
+{
+	return "core";
```
godric3, line 55: I guess core make sense here as it's not something provided by mods

lib/ObstacleHandler.cpp
```diff
@@ -31,6 +31,11 @@ std::string ObstacleInfo::getJsonKey() const
 	return identifier;
 }
 
+std::string ObstacleInfo::getModScope() const
+{
+	return "core";
```
godric3, line 36: `ObstacleInfo` is missing `modScope` but I think obstacles can be provided by mods. Is it intentional or should I introduce it and initialize in `loadFromJson`?

mapeditor/mapcontroller.cpp
```diff
@@ -613,16 +613,49 @@ ModCompatibilityInfo MapController::modAssessmentAll()
...
 	for(auto obj : map.objects)
 	{
-		if(obj->ID == Obj::HERO)
-			continue; //stub! 
-		
 		auto handler = obj->getObjectHandler();
-		auto modName = QString::fromStdString(handler->getJsonKey()).split(":").at(0).toStdString();
-		if(modName != "core")
-			result[modName] = VLC->modh->getModInfo(modName).getVerificationInfo();
+		auto modScope = QString::fromStdString(handler->getJsonKey()).split(":").at(0).toStdString();
```
godric3, line 627: does it make sense to introduce `getModScope` to `AObjectTypeHandler`?

lib/ObstacleHandler.cpp
IvanSavenko: Yeah, sounds like a bug.
While not 100% safe - fixing this one might reveal bugs someplace else (if any), but still a good idea to do.

PR #4425, opened 2024-08-09
Remove old code from UI system

IvanSavenko:
Cleanup of old parts of our UI system.

1. Removed `defActions` member from CIntObject. It was used to limit propagation of event (e.g. activation / redraws) to children. However by now all objects use such propagation and remaining places where this member is used is old, unclear code. Have not found not a single place where `defActions` were actually used for some non-zero effect.
2. Replaced 5 of various `OBJ_CONSTRUCTION_XXX` macros with two - one for use in self-construction (e.g. inside object constructor) and second one - for initiating same logic for specific UI element
3. Removed magical `255-DISPOSE` parameter from construction macro. Seems to be part of UI refactoring that was never cleaned up. Also completely remove DISPOSE flag since UI now always uses smart pointers with auto-deletion, have not found a single case where this flag was actually active

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -137,7 +137,6 @@ CPlayerInterface::CPlayerInterface(PlayerColor Player):
 	
 {
 	logGlobal->trace("\tHuman player interface for player %s being constructed", Player.toString());
-	GH.defActionsDef = 0;
 	LOCPLINT = this;
 	playerID=Player;
 	human=true;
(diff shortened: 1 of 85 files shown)


SoundSSGood:
May we somehow get rid of the macro at all?

IvanSavenko:
> May we somehow get rid of the macro at all?

Not really.
```c++
buttonQuit = std::make_shared<CButton>(...);
```
We have code like this. How CButton should know that it should be deactivated alongside with class that holds this button?
Right now this awareness comes from `parent->addChild(newElement)` call. And in order to find this parent we need some way to provide it.

Qt for example requires providing it as 1st parameter of every widget construction. If you're willing to make such change to all our UI code - sure.

Laserlicht:
Some regression of this PR?

grafik

PR #4428, opened 2024-08-10
VCMP format -> change to zip

Laserlicht:
As some modders want to use it it's time to get rid of this painful GZIP stuff.

Now VCMP is a simple zip file with an header.json (like in VMAP). And map path defined in this file.

Loading of old VCMP files should still work...

### DIFF
--- a/lib/campaign/CampaignHandler.cpp
+++ b/lib/campaign/CampaignHandler.cpp
@@ -156,7 +156,7 @@ void CampaignHandler::readHeaderFromJson(CampaignHeader & ret, JsonNode & reader
 	ret.campaignRegions = CampaignRegions::fromJson(reader["regions"]);
 	ret.numberOfScenarios = reader["scenarios"].Vector().size();
 	ret.name.appendTextID(readLocalizedString(ret, reader["name"].String(), filename, modName, "name"));
-	ret.description.appendTextID(readLocalizedString(ret, reader["description"].String(), filename, modName, "name"));
+	ret.description.appendTextID(readLocalizedString(ret, reader["description"].String(), filename, modName, "description"));
 	ret.author.appendRawString(reader["author"].String());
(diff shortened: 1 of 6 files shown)


lib/campaign/CampaignHandler.cpp
```diff
@@ -588,32 +589,68 @@ CampaignTravel CampaignHandler::readScenarioTravelFromMemory(CBinaryReader & rea
...
 {
-	CCompressedStream stream(std::move(file), true);
+	std::vector<ui8> magic(2);
+	file->read(magic.data(), magic.size());
+	file->seek(0);
 
 	std::vector< std::vector<ui8> > ret;
 
-	try
+	if (magic == std::vector<ui8>{0x50, 0x4B}) // VCMP (ZIP)
```
IvanSavenko, line 598: 1. Create named variable to explain what these magic numbers are
2. Use statically-allocated array instead of vector

```suggestion
	static const std::array<ui8, 2> zipHeaderMagic{0x50, 0x4B};
	if (magic == zipHeaderMagic ) // ZIP archive - assume VCMP format
```

lib/campaign/CampaignHandler.cpp
```diff
@@ -588,32 +589,68 @@ CampaignTravel CampaignHandler::readScenarioTravelFromMemory(CBinaryReader & rea
...
+				throw std::runtime_error(resource.getName() + " not found in " + filename);
+			auto data = loader.load(resource)->readAll();
+			ret.push_back(std::vector<ui8>(data.first.get(), data.first.get() + data.second));
 		}
-		while (!headerOnly && stream.getNextBlock());
+
+		return ret;
 	}
-	catch (const DecompressionException & e)
+	else // H3M
```
IvanSavenko, line 627:
```suggestion
	else // H3C
```

PR #4447, opened 2024-08-14
Split vcmiclient in two

smanolloff:
Similar to vcmiserver (app) and vcmiservercommon (lib), now there is vcmiclient (app) and vcmiclientcommon (lib).

This will allow to compile other clients (such as an ML client) which can now link against `vcmiclientcommon` to avoid the need to duplicate all of the code related to client initialization, networking, etc.

(I have not tried to cross-compile it, so I will rely on the CI results for iOS, Android and Windows builds)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -677,6 +677,7 @@ endif()
 
 if (ENABLE_CLIENT)
 	add_subdirectory(client)
+	add_subdirectory(clientapp)
 endif()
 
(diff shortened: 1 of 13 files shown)


DjWarmonger:
Actually this should target the develop branch. Beta is only for small patches that do not add new features.

smanolloff:
https://github.com/vcmi/vcmi/pull/4448

PR #4448, opened 2024-08-14
Split vcmiclient in two

smanolloff:
Similar to vcmiserver (app) and vcmiservercommon (lib), now there is vcmiclient (app) and vcmiclientcommon (lib).

This will allow to compile other clients (such as an ML client) which can now link against `vcmiclientcommon` to avoid the need to duplicate all of the code related to client initialization, networking, etc.

(I have not tried to cross-compile it, so I will rely on the CI results for iOS, Android and Windows builds)


clientapp/EntryPoint.cpp
IvanSavenko, line 1: A bit weird to have app startup in this file in clientapp project, but app shutdown - in CMT in client project.

I understand why it ended up that way, but still weird. Need to think how to make it nicer.

client/CMakeLists.txt
```diff
@@ -396,76 +396,58 @@ set(client_HEADERS
...
 	)
 endif()
 
-assign_source_group(${client_SRCS} ${client_HEADERS} VCMI_client.rc)
+assign_source_group(${vcmiclientcommon_SRCS} ${vcmiclientcommon_HEADERS})
 
 if(ANDROID)
-	add_library(vcmiclient SHARED ${client_SRCS} ${client_HEADERS})
-	set_target_properties(vcmiclient PROPERTIES
+	add_library(vcmiclientcommon SHARED ${vcmiclientcommon_SRCS} ${vcmiclientcommon_HEADERS})
```
IvanSavenko, line 410: How `clientapp/EntryPoint.cpp` is included on Android now if it is no longer part of vcmiclient shared library?

IvanSavenko (commented):
Considering changes - might be better to double-check that iOS and Android artifacts still work

smanolloff:
> Considering changes - might be better to double-check that iOS and Android artifacts still work

I will try to set up build environments for ios/android locally and check.

kambala-decapitator:
> > Considering changes - might be better to double-check that iOS and Android artifacts still work
> 
> I will try to set up build environments for ios/android locally and check.

it's much easier to grab artifacts from CI and install on your device(s). I have both iOS and Android if you need help in testing.

smanolloff:
> > > Considering changes - might be better to double-check that iOS and Android artifacts still work
> > 
> > 
> > I will try to set up build environments for ios/android locally and check.
> 
> it's much easier to grab artifacts from CI and install on your device(s). I have both iOS and Android if you need help in testing.

That's really kind of you, thanks. I may ask you to try later, but for now I want to make the runs complete successfully -- they did once, but are constantly being cancelled now, do you know how can I restart the builds without a git push?

IvanSavenko:
Generally, you need a git push.
You can make a fake one:
`git commit --amend`
`git push -f`
Empty amend would still change commit date, and as result - hash.

Repository owners (e.g. me) can also restart CI without push. Let me do that...

client/CMakeLists.txt
smanolloff: correct - it was not. I have changed clientapp to be built as a shared library if ANDROID is set

kambala-decapitator:
> but are constantly being cancelled now

if they're cancelled because one of the builds fails, you can disable that by setting [this variable](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) to false

smanolloff:
fixed & verified Android artifact as well -- it works now \o/

client/CMT.h
```diff
@@ -20,8 +20,6 @@ extern SDL_Surface *screen;      // main screen surface
 extern SDL_Surface *screen2;     // and hlp surface (used to store not-active interfaces layer)
 extern SDL_Surface *screenBuf; // points to screen (if only advmapint is present) or screen2 (else) - should be used when updating controls which are not regularly redrawed
 
+// defined in clientapp EntryPoint
 void handleQuit(bool ask = true);
-
-/// Notify user about encountered fatal error and terminate the game
-/// TODO: decide on better location for this method
-[[noreturn]] void handleFatalError(const std::string & message, bool terminate);
+void handleFatalError(const std::string & message, bool terminate);
```
IvanSavenko, line 25: Why was comment & `[[noreturn]]` attribute removed?

clientapp/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+      PROJECT_LABEL "VCMI_client"
+  )
+  set_property(DIRECTORY ${CMAKE_SOURCE_DIR} PROPERTY VS_STARTUP_PROJECT vcmiclient)
+  if(NOT ENABLE_DEBUG_CONSOLE)
+    set_target_properties(vcmiclient PROPERTIES WIN32_EXECUTABLE)
+    target_link_libraries(vcmiclient SDL2::SDL2main)
+  endif()
+  target_compile_definitions(vcmiclient PRIVATE WINDOWS_IGNORE_PACKING_MISMATCH)
+
+# TODO: very hacky, find proper solution to copy AI dlls into bin dir
```
kambala-decapitator, line 59: please fix indentation

kambala-decapitator (commented):
please change indentation in the new CML file to use tabs instead of spaces

in general looks fine to me

smanolloff:
> please change indentation in the new CML file to use tabs instead of spaces
> 
> in general looks fine to me

👍 done

kambala-decapitator (approved):

pppbb:
After this commit linux version don't build anymore.

IvanSavenko:
Not related to this PR, there were issues with boost 1.86. Can't find issue about it for some reason though...

Gigas002:
@pppbb Could you try out removing the build caches and rebuilding the package, or rebuilding in a clean chroot? The package builds fine for me without old caches

AlexFolland:
Deleting the issue ticket must have been a mistake, since it doesn't seem like it could be intended.  Can github staff restore it?

Crandel:
I've created a new issue #4561

PR #4449, opened 2024-08-14
Updated changelog with new changes

IvanSavenko:


DjWarmonger:
I think it's a good idea to separate battle AI and adventure AI. Many people still think it's one and the same.

IvanSavenko:
> I think it's a good idea to separate battle AI and adventure AI. Many people still think it's one and the same.

As far as changelog goes - we rarely get large number of AI changes to warrant separate blocks. Hopefully will be different in 1.6

PR #4461, opened 2024-08-17
Generates prebuilt libraries package in CI

IvanSavenko:
Very WIP, just to test whether it is doable.

Currently our prebuilt packages are compiled by one of our developers that has corresponding environment set up on his system. This works, but requires a lot of coordination whenever we need to update a package and if such developer unavailable or left the project, updating the package would be a pain.

One option (and one is recommended by conan) is to use persistent cache (artifactory), but it requires setup by somebody who known how to set it up. 

However another option is to build such packages as part of CI that we already have, especially since Conan can be used to build missing packages, e.g. after configuration change.

This PR:
- add separate workflow that instead of building vcmi will instead build missing packages for Conan.
- on success, this workflow will upload prebuilt libraries as artifacts, that can be used as input for our main CI.
- changes ffmpeg configuration to minimize its size by disabling most of its codecs which we don't use 

TODO:
- [x] find out what causes Android builds to fail
- [x] consider moving workflow to a separate repository, e.g. vcmi-dependencies
- [ ] try to create prebuilt package for msvc (with Win7 & 32-bit support), add CI for msvc+conan in addition to msvc+vcpkg
- [ ] move ffmpeg changes into separate PR and reupload prebuilt packages
- [x] consider upgrading ffmpeg from 4.4.3 to 7.0.1

Opinions? Especially @kambala-decapitator ?

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -100,6 +100,9 @@ class VCMI(ConanFile):
         self.options["ffmpeg"].postproc = False
         self.options["ffmpeg"].swresample = True
         self.options["ffmpeg"].with_asm = False
+        self.options["ffmpeg"].with_libaom = False
+        self.options["ffmpeg"].with_libsvtav1 = False
+        self.options["ffmpeg"].with_libdav1d = False
         self.options["ffmpeg"].with_zlib = False
         self.options["ffmpeg"].with_bzip2 = False
         self.options["ffmpeg"].with_lzma = False
@@ -124,7 +127,7 @@ class VCMI(ConanFile):
         # H3:SoD - .bik and .smk
         # H3:HD  -  ogg container / theora video / vorbis sound
         # and for mods - webm container / vp8 or vp9 video / opus sound
-        # TODO: add av1 support for mods (likely require newer ffmpeg than 4.4.3
+        # TODO: add av1 support for mods (requires enabling libdav1d which currently fails to build via Conan)
         self.options["ffmpeg"].enable_protocols = "file"
         self.options["ffmpeg"].enable_demuxers = "bink,binka,ogg,smacker,webm_dash_manifest"
         self.options["ffmpeg"].enable_parsers = "opus,vorbis,vp8,vp9,webp"
@@ -135,10 +138,11 @@ class VCMI(ConanFile):
         self.options["ffmpeg"].with_sdl = False
 
         #optionally, for testing - enable ffplay/ffprobe binaries in conan package:
-        #self.options["ffmpeg"].with_programs = True
-        #self.options["ffmpeg"].avfilter = True
-        #self.options["ffmpeg"].with_sdl = True
-        #self.options["ffmpeg"].enable_filters = "aresample,scale"
+        #if self.settings.os == "Windows":
+        #    self.options["ffmpeg"].with_programs = True
+        #    self.options["ffmpeg"].avfilter = True
+        #    self.options["ffmpeg"].with_sdl = True
+        #    self.options["ffmpeg"].enable_filters = "aresample,scale"
 
         self.options["sdl"].sdl2main = self.settings.os != "iOS"
         self.options["sdl"].vulkan = False
@@ -232,7 +236,7 @@ class VCMI(ConanFile):
 
         # client
         if self.options.with_ffmpeg:
-            self.requires("ffmpeg/[^4.4]")
+            self.requires("ffmpeg/[^7.0]")
 
         # launcher
         if self.settings.os == "Android":
--- a/conanfile.py
+++ b/conanfile.py
@@ -87,24 +87,58 @@ class VCMI(ConanFile):
         self.options["boost"].without_type_erasure = True
         self.options["boost"].without_wave = True
 
+        self.options["ffmpeg"].disable_all_bitstream_filters = True
+        self.options["ffmpeg"].disable_all_filters = True
+        self.options["ffmpeg"].disable_all_muxers = True
+        self.options["ffmpeg"].disable_all_hardware_accelerators = True
+        self.options["ffmpeg"].disable_all_protocols = True
+        self.options["ffmpeg"].disable_all_encoders = True
+        self.options["ffmpeg"].disable_all_decoders = True
+        self.options["ffmpeg"].disable_all_parsers = True
+        self.options["ffmpeg"].disable_all_demuxers = True
         self.options["ffmpeg"].avdevice = False
-        self.options["ffmpeg"].avfilter = False
         self.options["ffmpeg"].postproc = False
-        self.options["ffmpeg"].swresample = False
-        self.options["ffmpeg"].with_asm = self.settings.os != "Android"
+        self.options["ffmpeg"].with_asm = False
+        self.options["ffmpeg"].with_zlib = False
+        self.options["ffmpeg"].with_bzip2 = False
+        self.options["ffmpeg"].with_lzma = False
+        self.options["ffmpeg"].with_libiconv = False
         self.options["ffmpeg"].with_freetype = False
-        self.options["ffmpeg"].with_libfdk_aac = False
-        self.options["ffmpeg"].with_libmp3lame = False
-        self.options["ffmpeg"].with_libvpx = False
-        self.options["ffmpeg"].with_libwebp = False
+        self.options["ffmpeg"].with_openjpeg = False
+        self.options["ffmpeg"].with_openh264 = False
+        self.options["ffmpeg"].with_opus = False
+        self.options["ffmpeg"].with_vorbis = False
         self.options["ffmpeg"].with_libx264 = False
         self.options["ffmpeg"].with_libx265 = False
-        self.options["ffmpeg"].with_openh264 = False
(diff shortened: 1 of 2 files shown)


kambala-decapitator:
> One option (and one is recommended by conan) is to use persistent cache (artifactory), but it requires setup by somebody who known how to set it up.

The problem is, our servers aren't powerful enough to run Artifactory CE (I've already tried).

I also have transparent building of deps on CI on my list.

IvanSavenko:
> The problem is, our servers aren't powerful enough to run Artifactory CE (I've already tried).

Artifactory actually builds libraries locally on server? I though it is mostly for caching externally provided libraries. Not an option then, at least for now.


Also, @kambala-decapitator any idea what causes Android CI to fail here?
https://github.com/IvanSavenko/vcmi/actions/runs/10432518876/job/28893290880

Both 32-bit and 64-bit androids failed to rebuild ffmpeg. Something wrong with autotools? There is something else that needs to be set up for cross-compile?

conanfile.py
```diff
@@ -87,24 +87,58 @@ def configure(self):
...
+        self.options["ffmpeg"].swscale = True
+        # We want following options supported:
+        # H3:SoD - .bik and .smk
+        # H3:HD  -  ogg container / theora video / vorbis sound
+        # and for mods - webm container / vp9 video / opus sound
+        # TODO: add av1 support for mods (likely require newer ffmpeg than 4.4.3
+        self.options["ffmpeg"].enable_protocols = "file"
+        self.options["ffmpeg"].enable_demuxers = "bink,binka,ogg,smacker,webm_dash_manifest"
+        self.options["ffmpeg"].enable_parsers = "opus,vorbis,vp9,webp"
+        self.options["ffmpeg"].enable_decoders = "bink,binkaudio_dct,binkaudio_rdft,smackaud,smacker,theora,vorbis,vp9,opus"
```
Laserlicht, line 131: Please also add VP8. Tutorial videos are currently VP8 (don't have the original anymore to re-encode...).
And also if possible AV1 (for supporting all WEBM codecs).

IvanSavenko: vp8 - sure
av1 - need investigation. Either I need to enable something else or our ffmpeg 4.4 does not have full support for av1.

kambala-decapitator:
> Artifactory actually builds libraries locally on server? I though it is mostly for caching externally provided libraries. Not an option then, at least for now.

yes, it's for caching. But Artifactory CE is a java thing and requires a considerable amount of RAM - simply doesn't start.

>any idea what causes Android CI to fail here? 

difficult to say just like that, need to see the build logs as "C compiler test failed" - maybe wrong compiler was chosen

conanfile.py
kambala-decapitator: iirc 4.4 supports av1, but we can easily upgrade ffmpeg version if needed. Also I believe you'd need dav1d library to support av1 in ffmpeg.

kambala-decapitator:
> consider upgrading ffmpeg from 4.4.3 to 7.0.1

not sure we need to jump that far, probably 5.x as minimum would be enough. However, it's probably better to configure Conan to enable any later FFmpeg version, e.g. `>=5.0`

kambala-decapitator:
I'm starting to remember now that `--conf tools.android:ndk_path=$ANDROID_NDK_ROOT` isn't sufficient when you want to build (although dunno why, maybe fixed in Conan v2) - you should add NDK as a Conan dependency, so basically simply use `android-64-ndk` / `android-32-ndk` profiles instead.

IvanSavenko:
@kambala-decapitator 
That was it, thanks! Android CI seems to be working now:
https://github.com/IvanSavenko/vcmi/actions/runs/10439970233/job/28909278934
However something is wrong with output size - prebuilt package is 1.5 Gb, and not ~100 Mb like other platforms. Investigating.

Now need to think how to test & deploy these changes in a somewhat automated way.

> not sure we need to jump that far, probably 5.x as minimum would be enough

Yeah. Not planning to actually remove support for old versions, so minimum supported would stay at what we have now, just bump version in our conan packages to latest one.

===

Also - tried to generate msvc package. Unfortunately latest msvc is not supported by conan 1, needs conan 2: https://github.com/conan-io/conan/issues/16321

kambala-decapitator:
> However something is wrong with output size - prebuilt package is 1.5 Gb, and not ~100 Mb like other platforms. Investigating.

Right, sorry, forgot to say about that: Conan dir also contains source and build dirs which we don't need, as well as the ndk dependcy isn't needed for end user. For the latter just delete data/android-r* dir, for the former the command is `conan remove --src --build --force '*'` (writing out of my head, please check in conan remove -h)

kambala-decapitator:
> Unfortunately latest msvc is not supported by conan 1

Just use msvc2019

IvanSavenko:
I already have this one: `conan remove --builds --src --force '*'`
And seems to be working for platforms other than Android. Will try removing NDK, thanks.

> Just use msvc2019

Not sure if github runners have it preinstalled. And kind of don't like downgrading. In any case will try, still better than nothing.

kambala-decapitator:
https://github.com/actions/runner-images/blob/main/images/windows/Windows2019-Readme.md this one should have I believe, not sure about the 2022 one

kambala-decapitator:
BTW you can build android packages on Linux, currently we have to use macos builder in vcmi only because I built Qt for Android on macos

IvanSavenko:
Yep. Removal of Android NDK from conan reduces packages to below 100 Mb, similar to current ones. 

Will now try to generate package for older msvc via conan

Building Android on Linux - might look into it later. Rather low priority for me. The only downside is that it sometimes overloads our CI since we don't have many mac OS runners, but well, it is already configured and works.

kambala-decapitator:
> Building Android on Linux - might look into it later. Rather low priority for me.

I believe the only change required would be changing the runner image

IvanSavenko:
Wasn't able to build msvc package unfortunately.
ffmpeg - fails to build due to not found dependency. Reported as fixed on Conan, but still happens for me (or was it only fixed in conan2?)
Qt - also failure, due to some weird error. Found open ticket on Conan, with 0 discussions/solutions

Rest of packages seems to have been built though.

32-bit fails even before building. Probably fixable issue with our conanfile, but likely to run into same issues as 64-bit later

https://github.com/IvanSavenko/vcmi/actions/runs/10441871261

IvanSavenko:
Managed to build ffmpeg 7.0.1, with vp8 & vp9 support. 

av1 requires libdav1d which right now fails to build on Android and on mingw (ignore iOS ci failure, not related)
https://github.com/IvanSavenko/vcmi/actions/runs/10442254065

So current plan is:
- try to switch android runners to ubuntu (if there are no hard to fix issues)
- close this PR
- move workflow to vcmi-dependencies repository since this is where our mingw & android deps are
- once workflow succeeds, upload generated binaries to vcmi-dependencies release (in future we could make automatic release creation via CI)
- open new PR here with new ffmpeg settings, make sure that library works, adjust video-related code & docs.

kambala-decapitator:
> Wasn't able to build msvc package unfortunately.
> 
> ffmpeg - fails to build due to not found dependency. Reported as fixed on Conan, but still happens for me (or was it only fixed in conan2?)
> 
> Qt - also failure, due to some weird error. Found open ticket on Conan, with 0 discussions/solutions
> 
> 
> 
> Rest of packages seems to have been built though.
> 
> 
> 
> 32-bit fails even before building. Probably fixable issue with our conanfile, but likely to run into same issues as 64-bit later
> 
> 
> 
> https://github.com/IvanSavenko/vcmi/actions/runs/10441871261

The 32 bit error is easy to fix; the 64 bit one is weird.

kambala-decapitator:
libdav1d: most probably Conan recipe needs some fixes for that. To understand what needs to be done, check docs/issues in upstream first.

IvanSavenko:
ffmpeg error is marked as resolved here:
https://github.com/conan-io/conan-center-index/issues/23309
but not sure why I keep getting it.

Might no longer be a problem though - we don't need this library for ffmpeg, dav1d is better for us

There also was Qt error:
https://github.com/IvanSavenko/vcmi/actions/runs/10440924464/job/28911427192
Looks to be reported here, but no useful info:
https://github.com/conan-io/conan-center-index/issues/24686

kambala-decapitator:
looks like this is the real failure reason in Qt:
```
sh: line 1: C:.conan47b0911qt5qtbasesrccorelibmimetypesmimepackagesfreedesktop.org.xml: No such file or directory
qt_pch.h
Failed to run zstd -cq19 -T1 < C:\.conan\47b091\1\qt5\qtbase\src\corelib\mimetypes\mime\packages\freedesktop.org.xml at C:\.conan\47b091\1\qt5\qtbase\src\corelib\mimetypes\mime\generate.pl line 86.
```

btw 5.15.15 will be available on August 31

IvanSavenko:
Closing, superseded by https://github.com/vcmi/vcmi-dependencies/pull/1 and https://github.com/vcmi/vcmi/pull/4480

Things like building Android on Linux and msvc package - might look into that in future in separate PR's

PR #4462, opened 2024-08-17
Town buildings refactoring

IvanSavenko:
Removes a lot of unnecessarily hardcoded logic from town buildings system.

Changes:
- removed hardcoded bonusing buildings, e.g. brotherhood of sword, lighthouse, etc. Clarified docs to describe how to make similar town buildings in mods.
- add compatibility config for mods that use such buildings. At the moment, these buildings should work, but will lead to validation error.
- renamed `CGTownBuilding` to more clear `TownBuildingInstance`, to differentiate from global `CBuilding`
- simplified configuration for common buildings, e.g. mages guild. Now mods no longer need to specify (for example) dependencies of Capitol in their mod - this definition is now shared and loaded from vcmi config. Custom configuration in mod will still be respected.


IvanSavenko:
Why not check how lookout tower in Tower faction is done?
`"special2":       { "height" : "high", "requires" : [ "fort" ] },`
Although I am considering replacing these weird "heights" with either "scoutingRange" : (number) or with a bonus. Current solution is quite weird

PR #4463, opened 2024-08-17
IMarket refactoring

SoundSSGood:
IMarket now directly owns the CArtifactSet, which is required for the artifact altar. 
To save memory, IMarket itself allocates/deallocates the internal CArtifactSet if necessary. 
The CGArtifactsAltar class has been removed as no longer needed. 
IMarket is also now a serializable class. (This PR breaks save compatibility)

@Laserlicht  Now you can enable grotto at the cove.

### DIFF
--- a/client/windows/CMarketWindow.cpp
+++ b/client/windows/CMarketWindow.cpp
@@ -239,10 +239,10 @@ void CMarketWindow::createAltarArtifacts(const IMarket * market, const CGHeroIns
 	OBJECT_CONSTRUCTION;
 
 	background = createBg(ImagePath::builtin("ALTRART2.bmp"), PLAYER_COLORED);
-	auto altarArtifacts = std::make_shared<CAltarArtifacts>(market, hero, marketId);
-	marketWidget = altarArtifacts;
+	auto altarArtifactsStorage = std::make_shared<CAltarArtifacts>(market, hero, marketId);
+	marketWidget = altarArtifactsStorage;
 	artSets.clear();
-	const auto heroArts = altarArtifacts->getAOHset();
+	const auto heroArts = altarArtifactsStorage->getAOHset();
 	heroArts->clickPressedCallback = [this, heroArts](const CArtPlace & artPlace, const Point & cursorPosition)
 	{
 		clickPressedOnArtPlace(heroArts->getHero(), artPlace.slot, true, true, false);
@@ -258,7 +258,7 @@ void CMarketWindow::createAltarArtifacts(const IMarket * market, const CGHeroIns
 	addSet(heroArts);
 	initWidgetInternals(EMarketMode::ARTIFACT_EXP, CGI->generaltexth->zelp[568]);
 	updateExperience();
-	quitButton->addCallback([altarArtifacts](){altarArtifacts->putBackArtifacts();});
+	quitButton->addCallback([altarArtifactsStorage](){altarArtifactsStorage->putBackArtifacts();});
 }
 
 void CMarketWindow::createAltarCreatures(const IMarket * market, const CGHeroInstance * hero)
--- a/client/windows/CMarketWindow.cpp
+++ b/client/windows/CMarketWindow.cpp
@@ -32,9 +32,13 @@
 #include "../../lib/mapObjects/CGMarket.h"
 #include "../../lib/mapObjects/CGHeroInstance.h"
 
-CMarketWindow::CMarketWindow(const IMarket * market, const CGHeroInstance * hero, const std::function<void()> & onWindowClosed, EMarketMode mode)
+#include "../../CCallback.h"
+
+CMarketWindow::CMarketWindow(const IMarket * market, const CGHeroInstance * hero, const ObjectInstanceID & marketId,
+	const std::function<void()> & onWindowClosed, EMarketMode mode)
 	: CWindowObject(PLAYER_COLORED)
 	, windowClosedCallback(onWindowClosed)
+	, marketId(marketId)
 {
 	assert(mode == EMarketMode::RESOURCE_RESOURCE || mode == EMarketMode::RESOURCE_PLAYER || mode == EMarketMode::CREATURE_RESOURCE ||
 		mode == EMarketMode::RESOURCE_ARTIFACT || mode == EMarketMode::ARTIFACT_RESOURCE || mode == EMarketMode::ARTIFACT_EXP ||
@@ -180,7 +184,8 @@ void CMarketWindow::createArtifactsBuying(const IMarket * market, const CGHeroIn
 	OBJECT_CONSTRUCTION;
 
 	background = createBg(ImagePath::builtin("TPMRKABS.bmp"), PLAYER_COLORED);
-	marketWidget = std::make_shared<CArtifactsBuying>(market, hero);
+	marketWidget = std::make_shared<CArtifactsBuying>(market, hero, LOCPLINT->cb->getTown(marketId) ?
+		VLC->generaltexth->translate("building.core.conflux.special1.name") : CGI->generaltexth->allTexts[349]);
 	initWidgetInternals(EMarketMode::RESOURCE_ARTIFACT, CGI->generaltexth->zelp[600]);
 }
 
@@ -192,13 +197,14 @@ void CMarketWindow::createArtifactsSelling(const IMarket * market, const CGHeroI
 	// Create image that copies part of background containing slot MISC_1 into position of slot MISC_5
 	artSlotBack = std::make_shared<CPicture>(background->getSurface(), Rect(20, 187, 47, 47), 0, 0);
 	artSlotBack->moveTo(pos.topLeft() + Point(18, 339));
-	auto artsSellingMarket = std::make_shared<CArtifactsSelling>(market, hero);
+	auto artsSellingMarket = std::make_shared<CArtifactsSelling>(market, hero, LOCPLINT->cb->getTown(marketId) ?
+		VLC->generaltexth->translate("building.core.conflux.special1.name") : LOCPLINT->cb->getObj(marketId)->getObjectName());
 	artSets.clear();
 	const auto heroArts = artsSellingMarket->getAOHset();
 	heroArts->showPopupCallback = [this, heroArts](CArtPlace & artPlace, const Point & cursorPosition){showArifactInfo(*heroArts, artPlace, cursorPosition);};
 	addSet(heroArts);
 	marketWidget = artsSellingMarket;
-	initWidgetInternals(EMarketMode::ARTIFACT_RESOURCE, CGI->generaltexth->zelp[600]);	
+	initWidgetInternals(EMarketMode::ARTIFACT_RESOURCE, CGI->generaltexth->zelp[600]);
 }
 
 void CMarketWindow::createMarketResources(const IMarket * market, const CGHeroInstance * hero)
@@ -233,7 +239,7 @@ void CMarketWindow::createAltarArtifacts(const IMarket * market, const CGHeroIns
 	OBJECT_CONSTRUCTION;
 
 	background = createBg(ImagePath::builtin("ALTRART2.bmp"), PLAYER_COLORED);
-	auto altarArtifacts = std::make_shared<CAltarArtifacts>(market, hero);
+	auto altarArtifacts = std::make_shared<CAltarArtifacts>(market, hero, marketId);
 	marketWidget = altarArtifacts;
 	artSets.clear();
 	const auto heroArts = altarArtifacts->getAOHset();
--- a/client/windows/CMarketWindow.cpp
+++ b/client/windows/CMarketWindow.cpp
@@ -34,11 +34,9 @@
 
 #include "../../CCallback.h"
 
-CMarketWindow::CMarketWindow(const IMarket * market, const CGHeroInstance * hero, const ObjectInstanceID & marketId,
-	const std::function<void()> & onWindowClosed, EMarketMode mode)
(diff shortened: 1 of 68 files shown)


lib/mapping/MapReaderH3M.h
```diff
@@ -48,7 +48,7 @@ class MapReaderH3M
 	PlayerColor readPlayer();
 	PlayerColor readPlayer32();
 
-	void readBitmaskBuildings(std::set<BuildingID> & dest, std::optional<FactionID> faction);
+	std::set<BuildingID> readBitmaskBuildings(std::optional<FactionID> faction);
```
IvanSavenko, line 51: What's the reason behind this change?

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -1023,6 +994,55 @@ bool CGTownInstance::hasBuilt(const BuildingID & buildingID, FactionID townID) c
...
+	marketBuildingModeMapper(buildingID, [this](const EMarketMode mode) {removeMarketMode(mode);});
+}
+
+void CGTownInstance::marketBuildingModeMapper(const BuildingID & buildingID, const std::function<void(const EMarketMode)> & func)
+{
+	const auto townType = (*VLC->townh)[getFaction()]->town;
+	if(townType->buildings.find(buildingID) == townType->buildings.end())
+		return;
+
+	// TODO how to remove hardcoded buildings?
```
IvanSavenko, line 1021: 1. Add field `marketModes` to `CBuilding` class
2. Load this value from Json in `CTownHandler::loadBuilding`. See `MarketInstanceConstructor::initTypeData` for similar code.
3. Check for this field instead of checking for ID's in here
4. Update town building docs

Might be better to wait for my PR #4462 first though. I edited a lot of town buildings definitions & moved town building docs into a separate file

lib/mapObjects/IMarket.h
```diff
@@ -11,24 +11,41 @@
...
+	std::shared_ptr<CArtifactSet> getArtifactsStorage() const;
 	bool getOffer(int id1, int id2, int &val1, int &val2, EMarketMode mode) const; //val1 - how many units of id1 player has to give to receive val2 units
-	std::vector<EMarketMode> availableModes() const;
+
+	template <typename Handler> void serialize(Handler & h)
+	{
+		h & marketModes;
+
+		if(h.loadingGamestate && vstd::contains(marketModes, EMarketMode::ARTIFACT_EXP))
+			altarArtifactsStorage = std::make_shared<CArtifactSetAltar>();
```
IvanSavenko, line 43: Needs save compatibility check

server/CGameHandler.cpp
```diff
@@ -3920,15 +3920,15 @@ bool CGameHandler::sacrificeCreatures(const IMarket * market, const CGHeroInstan
 	return true;
 }
 
-bool CGameHandler::sacrificeArtifact(const IMarket * m, const CGHeroInstance * hero, const std::vector<ArtifactInstanceID> & arts)
+bool CGameHandler::sacrificeArtifact(const IMarket * market, const CGHeroInstance * hero, const ObjectInstanceID & marketId, const std::vector<ArtifactInstanceID> & arts)
```
IvanSavenko, line 3923: Why both pointer to IMarket and ObjectInstanceID are needed? It should be possible to obtain either one from another

lib/registerTypes/RegisterTypesMapObjects.h
```diff
@@ -51,7 +51,7 @@ void registerTypesMapObjects(Serializer &s)
 	s.template registerType<CGObjectInstance, CGLighthouse>();
 	s.template registerType<CGObjectInstance, CGTerrainPatch>();
 	s.template registerType<CGObjectInstance, HillFort>();
-	s.template registerType<CGObjectInstance, CGMarket>();
+	s.template registerType<CGObjectInstance, CGMarket>(); s.template registerType<IMarket, CGMarket>();
```
IvanSavenko, line 54: Please move this new type registration to end of the list

lib/mapObjects/CGMarket.h
```diff
@@ -78,20 +71,8 @@ class DLL_LINKAGE CGUniversity : public CGMarket
...
-{
-public:
-	using CGMarket::CGMarket;
-
-	ArtBearer::ArtBearer bearerType() const override;
-
-	template <typename Handler> void serialize(Handler & h)
-	{
-		h & static_cast<CGMarket&>(*this);
-		h & static_cast<CArtifactSet&>(*this);
```
IvanSavenko, line 94: This would break saves... Let me think...

lib/mapObjectConstructors/CommonConstructors.cpp
```diff
@@ -238,25 +238,23 @@ CGMarket * MarketInstanceConstructor::createObject(IGameCallback * cb) const
...
+	market->addMarketMode(marketModes);
 	market->marketEfficiency = marketEfficiency;
 	
-	market->title = market->getObjectName();
-	if(!title.empty())
-		market->title = VLC->generaltexth->translate(title);
-	
-	if (!speech.empty())
-		market->speech = VLC->generaltexth->translate(speech);
+	if(auto univercity = dynamic_cast<CGUniversity*>(market))
```
IvanSavenko, line 249:
```suggestion
	if(auto university = dynamic_cast<CGUniversity*>(market))
```

lib/mapObjects/CGTownInstance.h
```diff
@@ -172,6 +172,8 @@ class DLL_LINKAGE CGTownInstance : public CGDwelling, public IShipyard, public I
 	//checks if building is constructed and town has same subID
 	bool hasBuilt(const BuildingID & buildingID) const;
 	bool hasBuilt(const BuildingID & buildingID, FactionID townID) const;
+	void addBuilding(const BuildingID & buildingID);
+	void removeBuilding(const BuildingID & buildingID);
```
IvanSavenko, line 176: Perhaps make builtBuildings private and only allow modification via these methods?

Laserlicht:
@SoundSSGood Thanks!

Tested it with the grotto. Two things:

Still not possible to place artifacts onto altar in town. Got an exception:
grafik

`marketBuildingModeMapper` doesn't fits the needs here. The market mode depends on which building I open, not on the buildings existing in town. With grotto enabled I've some overlapping buttons in the window. But with grotto I should only see altar of sacrifice buttons and with market only market buttons.

lib/mapping/MapReaderH3M.h
SoundSSGood: These are not mandatory changes. Rolled back.

server/CGameHandler.cpp
SoundSSGood: I thought that passing ids directly is better than a certain number of dynamic_casts. Perhaps even IMarket should be ObjectInstance. Because IMarket is always an object on the map.

IvanSavenko: Perhaps instead add method to CGObjectInstance or to IObjectInterface directly:
```cpp
const IMarket * CGObjectInstance::getMarket() const
{
    return dynamic_cast<const IMarket*>(this);
}
```
Market and shipyard already act as optional extensions to IObjectInterface, so makes sense to integrate them better. And I would like to have few more such extensions in future.

Alternatively, if we want to minimize overhead of dynamic_cast we can turn it into virtual method - CGObjectInstance will always return nullptr, and CMarketInstance (or what's its name) will return `this`, removing need of dynamic cast

SoundSSGood: > so makes sense to integrate them better

Do you mean composition?

IvanSavenko: Not necessarily.
If we go virtual function route, then `CGMarkeInstance::getMarket()` can be implemented as either `return this` in case of inheritance or `return market` in case of composition. This would be implementation detail of CGMarketInstance.

Outside code would always use CGObjectInstance::getMarket(). Ideally - outside code should not even access CGMarketInstance, and always use getMarket() method.

SoundSSGood: In this case, we will still need to additionally pass the object id for methods that require a market.

IvanSavenko: We'll need to pass *either* object ID or CGObjectInstance pointer. Not both. `getMarket` would be a virtual method of IObjectInterface.

If we go for object ID then you can retrieve market via something like `cb->getObject(objectID)->getMarket()`, if you have CGObjectInstance - then simply via `object->getMarket()`

SoundSSGood: If we inherit CMarketInstance from IObjectInterface, and then CGTownInstance inherits from CMarketInstance. CGTownInstance must virtually inherit from IObjectInterface. Is it ok?

SoundSSGood: > Or, alternatively, we could also add `getObject` method to IMarket that would return either CGObjectInstance or IObjectInterface.

But the market does not own information about CGObjectInstance or IObjectInterface. Should the market inherit from CGObjectInstance? Or include it?

IvanSavenko: `getObject` method will have to be virtual and implemented by CGMarket / CGTownInstance as simple `return this`

SoundSSGood: This solution is cleaner than the previous one. I like it.

lib/mapObjects/CGTownInstance.cpp
SoundSSGood: Should the market repeat the same structure
`"types" : { "object" : { "modes": [] } }`  as [in](https://github.com/vcmi/vcmi/blob/develop/docs/modders/Map_Objects/Market.md) ?

IvanSavenko: No need. Only inner-most "modes". Although would be better to name it "marketModes" or "tradeModes" or something similar since town building might have other functionality than market.

So something like this will do:
```json
"marketplace" : {
    "tradeModes" : [ "resource-resource" ]
}
```

lib/entities/faction/CTownHandler.cpp
```diff
@@ -342,6 +342,11 @@ void CTownHandler::loadBuilding(CTown * town, const std::string & stringID, cons
 		ret->upgrade = BuildingID::NONE;
 
 	ret->town->buildings[ret->bid] = ret;
+	for(const auto & element : source["marketModes"].Vector())
+	{
+		if(MappedKeys::MARKET_NAMES_TO_TYPES.count(element.String()))
+			ret->marketModes.insert(MappedKeys::MARKET_NAMES_TO_TYPES.at(element.String()));
+	}
```
IvanSavenko, line 349: Please also add this to our json validator at `config/schemas/townBuilding.json`, probably with "enum" check

lib/registerTypes/RegisterTypesMapObjects.h
```diff
@@ -55,6 +55,7 @@ void registerTypesMapObjects(Serializer &s)
 		s.template registerType<CGMarket, CGBlackMarket>();
 		s.template registerType<CGMarket, CGUniversity>();
 	s.template registerType<CGObjectInstance, CGHeroPlaceholder>();
+	s.template registerType<IMarket, CGMarket>();
```
IvanSavenko, line 58: Can you move this line to the end - to line 135? So that all objects inbetween won't change their order?

docs/modders/Entities_Format/Town_Building_Format.md
```diff
@@ -248,3 +247,17 @@ In addition to above, it is possible to use same format as [Rewardable](../Map_O
...
+
+##### Marketplace
+```jsonc
+	"marketplace":    { "marketModes" : ["resource-resource", "resource-player"] },
+```
+
+##### Artifact merchant
+```jsonc
+	"special1":       { "type" : "artifactMerchant", "requires" : [ "marketplace" ], "marketModes" : ["resource-artifact", "artifact-resource"] },
+```
```
IvanSavenko, line 263: Should also be mentioned in "Town building node" section (lines 113-175).

client/windows/CMarketWindow.cpp
```diff
@@ -192,13 +200,15 @@ void CMarketWindow::createArtifactsSelling(const IMarket * market, const CGHeroI
 	// Create image that copies part of background containing slot MISC_1 into position of slot MISC_5
 	artSlotBack = std::make_shared<CPicture>(background->getSurface(), Rect(20, 187, 47, 47), 0, 0);
 	artSlotBack->moveTo(pos.topLeft() + Point(18, 339));
-	auto artsSellingMarket = std::make_shared<CArtifactsSelling>(market, hero);
+	const auto mapObj = LOCPLINT->cb->getObj(market->getObjInstanceID());
+	auto artsSellingMarket = std::make_shared<CArtifactsSelling>(market, hero, LOCPLINT->cb->getTown(mapObj->id) ?
+		VLC->generaltexth->translate("building.core.conflux.special1.name") : mapObj->getObjectName());
```
IvanSavenko, line 205: Perhaps we should instead provide this string from outside somehow? Or provide its building ID?

SoundSSGood: Added `CMarketWindow::getMarketTitle` func

docs/modders/Entities_Format/Town_Building_Format.md
SoundSSGood: Done. Looks like "type" : "artifactMerchant" is not needed. "marketModes" is enough.

IvanSavenko: Somewhat.
Yeah, mechanics no longer need that type but I think it is still used for UI - to open correct mode when clicking on window. Same goes to other similar buildings - freelancers guild, marketplace, possibly - university of magic, etc.

Will take a look, already working on removing no longer necessary building types.

PR #4473, opened 2024-08-18
Small mapeditor fixes

MichalZr6:
The case when Player is defined but has no heroes or towns assigned will lead to crash when choosing this map in VCMI client.


mapeditor/validator.cpp
```diff
@@ -58,12 +58,14 @@ std::list<Validator::Issue> Validator::validate(const CMap * map)
 		//check player settings
 		int hplayers = 0;
 		int cplayers = 0;
-		std::map<int, int> amountOfCastles;
+		std::map<int, int> amountOfTowns;				// std::map<PlayerID, TownCounter>
+		std::map<int, int> amountOfHeroes;				// std::map<PlayerID, HeroCounter>
```
IvanSavenko, line 62:
```suggestion
		std::map<PlayerColor, int> amountOfTowns;
		std::map<PlayerColor, int> amountOfHeroes;
```

mapeditor/mapcontroller.cpp
```diff
@@ -623,7 +628,7 @@ ModCompatibilityInfo MapController::modAssessmentMap(const CMap & map)
 
 	for(auto obj : map.objects)
 	{
-		auto handler = obj->getObjectHandler();
+		auto handler = obj->getObjectHandler();
```
IvanSavenko, line 631:
```suggestion
		auto handler = obj->getObjectHandler();
```

mapeditor/graphics.cpp
```diff
@@ -48,9 +48,10 @@ void Graphics::loadPaletteAndColors()
 	for(int i = 0; i < 256; ++i)
 	{
 		QColor col;
-		col.setRed(pals[startPoint++]);
-		col.setGreen(pals[startPoint++]);
-		col.setBlue(pals[startPoint++]);
+		// Cast to unsigned char to ensure values are in the range 0-255
```
IvanSavenko, line 51: If you want to ensure that there is no overflow then you need to use std::clamp.
Casting will not handle overflows for you

mapeditor/validator.cpp
```diff
@@ -159,9 +156,18 @@ std::list<Validator::Issue> Validator::validate(const CMap * map)
...
+		for (auto & playerTCounter : amountOfTowns)
+		{
+			if (playerTCounter.second == 0)
+			{
+				// FIXME: heroesNames are empty even though heroes are on the map
+				// if(map->players[playerTCounter.first].heroesNames.empty())
+				if(amountOfHeroes[playerTCounter.first] == 0)
+					issues.emplace_back(QString(tr("Player %1 has no towns and heroes assigned")).arg(playerTCounter.first + 1), true);
+				else
+					issues.emplace_back(QString(tr("Player %1 doesn't have any starting town")).arg(playerTCounter.first + 1), false);
```
kambala-decapitator, line 168:
```suggestion
					issues.emplace_back(tr("Player %1 has no towns and heroes assigned").arg(playerTCounter.first + 1), true);
				else
					issues.emplace_back(tr("Player %1 doesn't have any starting town").arg(playerTCounter.first + 1), false);
```

lib/spells/CSpellHandler.cpp
```diff
@@ -687,7 +687,7 @@ std::vector<int> CSpellHandler::spellRangeInHexes(std::string input) const
 	std::set<BattleHex> ret;
 	std::string rng = input + ','; //copy + artificial comma for easier handling
 
-	if(rng.size() >= 2 && rng[0] != 'X') //there is at least one hex in range (+artificial comma)
+	if(rng.size() >= 2 && rng[0] != 'X' && rng[0] != 'x') //there is at least one hex in range (+artificial comma)
```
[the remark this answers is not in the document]
MichalZr6: Yes, I thought about it but... didn't apply. Moreover this change was made via another PR which has been already merged. I just added this code in this branch for the reason not to get exception while testing and haven't unchecked for staging. Can I fix this - as you suggest - in this PR ?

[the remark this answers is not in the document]
kambala-decapitator: > Yes, I thought about it but... didn't apply. Moreover this change was made via another PR which has been already merged. I just added this code in this branch for the reason not to get exception while testing and haven't unchecked for staging. Can I fix this - as you suggest - in this PR ?

yes, sure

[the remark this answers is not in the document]
MichalZr6: > you could also do this: `toLowerCase(rng[0]) != 'x'`

I think it has to be `std::tolower()` in this case

mapeditor/mapcontroller.cpp
```diff
@@ -384,6 +384,11 @@ void MapController::pasteFromClipboard(int level)
 	for(auto & objUniquePtr : _clipboard)
 	{
 		auto * obj = CMemorySerializer::deepCopy(*objUniquePtr).release();
+		QString errorMsg;
+		if (!canPlaceObject(level, obj, errorMsg))
+		{
+			QMessageBox::information(main, QObject::tr("Warning"), errorMsg);
```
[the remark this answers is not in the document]
kambala-decapitator: > perhaps I should put error messages into vector 

exactly. I don't think that trimming is necessary, at least now.

lib/spells/CSpellHandler.cpp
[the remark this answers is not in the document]
kambala-decapitator: > > you could also do this: `toLowerCase(rng[0]) != 'x'`
> 
> I think it has to be `std::tolower()` in this case

yeah, that was sort of pseudocode :)

mapeditor/validator.cpp
MichalZr6: I changed `std::list` to `std::set` in case there would be some duplicated messages.

mapeditor/mapcontroller.cpp
```diff
@@ -381,13 +381,14 @@ void MapController::pasteFromClipboard(int level)
 	if(_clipboardShiftIndex == int3::getDirs().size())
 		_clipboardShiftIndex = 0;
 	
+	QStringList errors;
+	QString errorMsg;
```
kambala-decapitator, line 385: not sure why declaring outside the loop

mapeditor/validator.h
```diff
@@ -30,13 +31,18 @@ class Validator : public QDialog
 		bool critical;
 		
 		Issue(const QString & m, bool c): message(m), critical(c) {}
+
+		bool operator <(const Issue & other) const
+		{
+			return message < other.message;
```
kambala-decapitator, line 37: technically you should also take the second field into account

kambala-decapitator (commented):
why was google test submodule updated?

MichalZr6:
> why was google test submodule updated?

I have no idea why (from time to time) this google file is showing up as _changed_. I saw once on discord someone was also asking why it happens.

> and please resolve conflicts
Wanted to do so and used merge but this messed up the PR. Should I do `rebase --onto` ? I still have to learn a lot about this github staff...

mapeditor/mapcontroller.cpp
MichalZr6: Wanted to use it again with `std::vector` approach, but then changed to `QStringList` which has this nice `join()` function. Should be back in loop or dosn't matter ?

mapeditor/validator.h
MichalZr6: Thought it doesn't matter for this struct. Will fix.

kambala-decapitator: > Thought it doesn't matter for this struct. Will fix.

I think it doesn't indeed - guess you don't have same messages that can be true and false

mapeditor/mapcontroller.cpp
kambala-decapitator: I mean, only errorMsg should be inside the loop, makes no sense to have it outside

kambala-decapitator:
> I have no idea why (from time to time) this google file is showing up as _changed_. I saw once on discord someone was also asking why it happens.

depends on how you pull changes. E.g. GUI client could be updating submodules automatically.

> Wanted to do so and used merge but this messed up the PR. Should I do `rebase --onto` ? I still have to learn a lot about this github staff...

yeah, currently your branch looks quite messed up. I suggest to create new branch (on develop) and cherry-pick only your changes into it, and open new PR.

normally, when you're in your branch, you'd do either `git rebase origin/develop` or `git merge origin/develop`.

MichalZr6:
>depends on how you pull changes. E.g. GUI client could be updating submodules automatically.

saw that in Github Desktop as well. If it appears again what should be done preferably ?

kambala-decapitator:
> If it appears again what should be done preferably ?

simply don't commit the submodule change

> Should I change this PR to draft or close it ?

just close

kambala-decapitator:
But why are you still pushing to this broken branch?

MichalZr6:
One last commit before cherry-picking to new branch. However, probably stashing should be the way...

PR #4478, opened 2024-08-19
generate assets to file

Laserlicht:
- allows modding of small and big spellbook
  - always unmodded file is loaded now for generation
- speed up (?) resource loading -> generation is only done once

### DIFF
--- a/client/gui/CGuiHandler.cpp
+++ b/client/gui/CGuiHandler.cpp
@@ -24,6 +24,7 @@
 #include "../render/Graphics.h"
 #include "../render/IFont.h"
 #include "../render/EFont.h"
+#include "../render/AssetGenerator.h"
 #include "../renderSDL/ScreenHandler.h"
 #include "../renderSDL/RenderHandler.h"
 #include "../CMT.h"
@@ -63,6 +64,8 @@ void CGuiHandler::init()
 	shortcutsHandlerInstance = std::make_unique<ShortcutHandler>();
 	inputHandlerInstance = std::make_unique<InputHandler>(); // Must be after windowHandlerInstance and shortcutsHandlerInstance
 	framerateManagerInstance = std::make_unique<FramerateManager>(settings["video"]["targetfps"].Integer());
+
+	AssetGenerator::generate();
 }
 
 void CGuiHandler::handleEvents()
--- a/client/render/AssetGenerator.cpp
+++ b/client/render/AssetGenerator.cpp
@@ -1,0 +1,86 @@
+/*
+ * AssetGenerator.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#include "StdInc.h"
+#include "AssetGenerator.h"
+
+#include "../gui/CGuiHandler.h"
+#include "../render/IImage.h"
+#include "../render/IImageLoader.h"
+#include "../render/Canvas.h"
+#include "../render/IRenderHandler.h"
+
+#include "../lib/filesystem/Filesystem.h"
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+void AssetGenerator::generate()
+{
+    createBigSpellBook("data/SpelBk2.bmp");
+}
+
+void AssetGenerator::createBigSpellBook(std::string filename)
+{
+	if(!CResourceHandler::get("local")->createResource(filename))
+		return;
+
+    auto res = ImagePath::builtin("SpelBack");
+    res.setOriginalResource(true);
+
+	std::shared_ptr<IImage> img = GH.renderHandler().loadImage(res, EImageBlitMode::OPAQUE);
+	Canvas canvas = Canvas(Point(800, 600));
+	// edges
+	canvas.draw(img, Point(0, 0), Rect(15, 38, 90, 45));
+	canvas.draw(img, Point(0, 460), Rect(15, 400, 90, 141));
+	canvas.draw(img, Point(705, 0), Rect(509, 38, 95, 45));
+	canvas.draw(img, Point(705, 460), Rect(509, 400, 95, 141));
+	// left / right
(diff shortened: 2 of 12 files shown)


Laserlicht:
@IvanSavenko Keep in mind, that this PR may hide the remaining issue with upscaling and the spellbook.

If you start this PR with 1x first, then 2x, 3x... are working too afterwards. But if you start with > 1x first then there is a corrupted spellbook.

IvanSavenko:
Perhaps we should generate resource only once per run AND reset/delete it on next start?

So if I were to install mod that changes original spellbook game would upscale it instead of loading cached upscaled version from previous runs?

Can be done either as caching on disk (and deleting temporary directory on start) or as caching in memory (by keeping image in memory even after it is no longer needed)

Laserlicht:
> Perhaps we should generate resource only once per run AND reset/delete it on next start?

The idea of PR is that new resources are created which we cannot supply, but which consist of other, already existing resources.

The generated resources must therefore be generated unscaled (always from unmodded files). They are then handled in the same way as OH3 resources. They can therefore also be overwritten by mods.

It therefore makes sense to generate them only once. But yes. we still have the case of vcmi updates... So it's still better to generate it once per start, I think.

Laserlicht:
Tested a little bit:
- Generates resources at start
- Generated folder will be deleted every start
- if modded file with same name exists: generation is skipped for that resource
- generation will always use unmodded resource

client/render/AssetGenerator.cpp
```diff
@@ -0,0 +1,118 @@
...
+	canvas.draw(img, Point(53, 520), Rect(53, 264, 339, 47));
+
+	std::shared_ptr<IImage> image = GH.renderHandler().createImage(canvas.getInternalSurface());
+
+	image->exportBitmap(*CResourceHandler::get("local")->getResourceName(savePath));
+}
+
+void AssetGenerator::createBigSpellBook()
+{
+	std::string filename = "data/SpelBk2.bmp";
```
IvanSavenko, line 59: We're not limited to 8.3 filenames
```suggestion
	std::string filename = "data/SpellBookLarge.png";
```

client/render/AssetGenerator.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+void AssetGenerator::generate()
+{
+	createBigSpellBook();
+	createAdventureOptionsCleanBackground();
+}
+
+void AssetGenerator::createAdventureOptionsCleanBackground()
+{
+	std::string filename = "data/ADVOPTBC.bmp";
```
IvanSavenko, line 29: Please use human-readable names. "ADVOPTBC" does not tells much about what this image is.

client/render/AssetGenerator.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+	if(CResourceHandler::get()->existsResource(ResourcePath(filename))) // overridden by mod, no generation
+		return;
+
+	if(!CResourceHandler::get("local")->createResource(filename))
+		return;
+	ResourcePath savePath(filename, EResType::IMAGE);
+
+	auto res = ImagePath::builtin("SpelBack");
+	res = res.setOriginalResource(true);
```
IvanSavenko, line 69: Perhaps instead of this flag we could create separate directory for autogenerated resources?
So if mod want to provide own spellbook image, vcmi will look for it in `data/SpellBookLarge.png`. But autogenerated image will be placed into `(local fs)/generated/data/SpellBookLarge.png`

Don't see why we need such flag as part of FS system.

client/gui/CGuiHandler.cpp
```diff
@@ -63,6 +64,8 @@ void CGuiHandler::init()
 	shortcutsHandlerInstance = std::make_unique<ShortcutHandler>();
 	inputHandlerInstance = std::make_unique<InputHandler>(); // Must be after windowHandlerInstance and shortcutsHandlerInstance
 	framerateManagerInstance = std::make_unique<FramerateManager>(settings["video"]["targetfps"].Integer());
+
+	AssetGenerator::generate();
```
IvanSavenko, line 68: Should be done on first access to corresponding image, to avoid increasing startup time

client/render/AssetGenerator.cpp
Laserlicht: Yes, this is currently the case. assets are already generated in `(local fs)/generated/data/xyz.bmp`.

This flag is to force load the original, unmodded resource, even if there is a modded one. If someone will do a modded (**small**) spellbook: Without this flag the generation is done on the modded file. This may create a corrupted big spellbook (if there is no modded **big** spellbook).

client/gui/CGuiHandler.cpp
Laserlicht: Yes. But not sure, how to improve this without changes in the resource system.

At the moment for the two resources it's not a big problem as resource generation is quite fast.

client/render/AssetGenerator.cpp
Laserlicht: good point!

client/gui/CGuiHandler.cpp
Laserlicht: On my laptop-cpu inside of a VM it needs 10ms currently.

client/render/AssetGenerator.cpp
IvanSavenko: That sounds like a problem with mod. Mod should either provide valid small spellbook image or also provide large spellbook.

Sounds like an overkill solution for a problem that likely does not exists in the first place

client/gui/CGuiHandler.cpp
IvanSavenko: 1 ms here, 1 ms there...
And on low-end phone this would be not 10 ms but 50 ms or even more.

client/render/AssetGenerator.cpp
Laserlicht: Hmm. Yes you're right. This could be a little bit overengineered.
Will remove.

client/gui/CGuiHandler.cpp
Laserlicht: now they will only generated when needed...

PR #4480, opened 2024-08-20
Update ffmpeg

IvanSavenko:
Continuation of https://github.com/vcmi/vcmi/pull/4461

Part 1 - update conan ffmpeg configuration:
- updated options to include new options added in upstream
- enabled swresample to allow audio resampling via ffmpeg instead of custom methods
- disabled all components other than those that are needed for vcmi
- changed required version to allow newer ffmpeg versions

Part 2 - use newer dependencies package from shared repository (instead of repository-per-platform)

Part 3 - update code to reflect changes
- [x] make sure that code compiles with new ffmpeg
- [x] test new package on all affected platforms
- [x] update documentation for modders on what video formats are supported
- [x] check our virtual filesystem to make sure that list of extensions recognized as video files matches with actually supported formats

### DIFF
--- a/CI/ios/before_install.sh
+++ b/CI/ios/before_install.sh
@@ -2,6 +2,4 @@
 
 echo DEVELOPER_DIR=/Applications/Xcode_14.2.app >> $GITHUB_ENV
 
-mkdir ~/.conan ; cd ~/.conan
-curl -L 'https://github.com/vcmi/vcmi-dependencies/releases/download/1.2/dependencies-ios.tgz' \
-	| tar -xzf -
+. CI/install_conan_dependencies.sh "dependencies-ios"
--- a/CI/ios/before_install.sh
+++ b/CI/ios/before_install.sh
@@ -3,5 +3,5 @@
 echo DEVELOPER_DIR=/Applications/Xcode_14.2.app >> $GITHUB_ENV
 
 mkdir ~/.conan ; cd ~/.conan
-curl -L 'https://github.com/vcmi/vcmi-ios-deps/releases/download/1.2.1/ios-arm64.txz' \
-	| tar -xf -
+curl -L 'https://github.com/vcmi/vcmi-dependencies/releases/download/1.2/dependencies-ios.tgz' \
+	| tar -xzf -
--- a/client/media/CVideoHandler.cpp
+++ b/client/media/CVideoHandler.cpp
@@ -339,10 +339,11 @@ FFMpegStream::~FFMpegStream()
 {
 	av_frame_free(&frame);
 
+#if (LIBAVCODEC_VERSION_MAJOR < 61 )
+	// deprecated, apparently no longer necessary - avcodec_free_context should suffice
 	avcodec_close(codecContext);
-	avcodec_free_context(&codecContext);
+#endif
 
-	avcodec_close(codecContext);
 	avcodec_free_context(&codecContext);
 
 	avformat_close_input(&formatContext);
--- a/conanfile.py
+++ b/conanfile.py
@@ -20,6 +20,7 @@ class VCMI(ConanFile):
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
         "onetbb/[^2021.3]",
+        "xz_utils/[>=5.2.5]", # Required for innoextract
     ]
 
     requires = _libRequires + _clientRequires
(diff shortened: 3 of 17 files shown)


IvanSavenko:
Strange. Now CMake can no longer find lzma package.
Even though that package is present in both old & new packages and in fact is completely identical.

Comparing mingw package:
- ffmpeg/4.4.3 removed (expected)
- ffmpeg/7.0.1 added (also expected)
- nasm/2.16.01 added (looks to be new dependency of ffmpeg)
- pkgconfig/2.1.0 added (in addition to 1.7.4 and 1.9.3)

Could be related to newer pkgconfig?

IvanSavenko:
Ah, apparently we used lzma without actually having it as dependency in conan.

In old package, lzma was provided by conan as ffmpeg dependency. However we also use it (almost) directly - as dependency for innoextract. Now I have disabled lzma in ffmpeg, so conan no longer provides it in generated profile.

Will restore it for ffmpeg (since we need it anyway) and add explicit dependency

IvanSavenko:
Seems to be working, at least mingw build. Will recheck android and mingw-32 a bit later.
Install size also went down by ~6-8 Mb, depending on platform.

CI/ios/before_install.sh
```diff
@@ -3,5 +3,5 @@
 echo DEVELOPER_DIR=/Applications/Xcode_14.2.app >> $GITHUB_ENV
 
 mkdir ~/.conan ; cd ~/.conan
-curl -L 'https://github.com/vcmi/vcmi-ios-deps/releases/download/1.2.1/ios-arm64.txz' \
-	| tar -xf -
+curl -L 'https://github.com/vcmi/vcmi-dependencies/releases/download/1.2/dependencies-ios.tgz' \
+	| tar -xzf -
```
kambala-decapitator, line 7: -z not needed on macOS in x mode

client/media/CVideoHandler.cpp
```diff
@@ -339,10 +339,13 @@ FFMpegStream::~FFMpegStream()
...
+#if (LIBAVCODEC_VERSION_MAJOR < 61 )
+	// deprecated, apparently no longer necessary - avcodec_free_context should suffice
 	avcodec_close(codecContext);
-	avcodec_free_context(&codecContext);
-
 	avcodec_close(codecContext);
+#endif
+
+	avcodec_free_context(&codecContext);
 	avcodec_free_context(&codecContext);
```
kambala-decapitator, line 349: why doing double close+free?

conanfile.py
```diff
@@ -87,24 +88,63 @@ def configure(self):
...
         self.options["ffmpeg"].with_opus = False
-        self.options["ffmpeg"].with_programs = False
-        self.options["ffmpeg"].with_ssl = False
         self.options["ffmpeg"].with_vorbis = False
+        self.options["ffmpeg"].with_libx264 = False
+        self.options["ffmpeg"].with_libx265 = False
+        self.options["ffmpeg"].with_libvpx = False
+        self.options["ffmpeg"].with_libmp3lame = False
+        self.options["ffmpeg"].with_libwebp = False
+        self.options["ffmpeg"].with_ssl = False
```
kambala-decapitator, line 121: settings should be either in alphabetical order or split into logical groups

conanfile.py
```diff
@@ -87,24 +88,63 @@ def configure(self):
...
+        self.options["ffmpeg"].with_programs = False
+        self.options["ffmpeg"].avfilter = False
+        self.options["ffmpeg"].with_sdl = False
+
+        #optionally, for testing - enable ffplay/ffprobe binaries in conan package:
+        #if self.settings.os == "Windows":
+        #    self.options["ffmpeg"].with_programs = True
+        #    self.options["ffmpeg"].avfilter = True
+        #    self.options["ffmpeg"].with_sdl = True
+        #    self.options["ffmpeg"].enable_filters = "aresample,scale"
```
kambala-decapitator, line 147: FYI you can pass it on command line

client/media/CVideoHandler.cpp
IvanSavenko: Technically old code, but yeah, looks wrong. Either copy-paste error or some broken merge. Will take a look.

CI/ios/before_install.sh
IvanSavenko: I actually wanted to unify this code in a single .sh file.

Right now we have like 7 `before_install.sh`, all of which execute mkdir, cd, curl & tar. And you need to change url in each of them when updating dependencies, which is a pain.

So I would rather create shared `install_dependencies.sh` file that will be executed for all conan CI's

conanfile.py
IvanSavenko: I used it on CI, so no command line.

Another problem was to find correct set of switches. Simple 'with_programs' would silently fail, had to dig into ffmpeg's configure to figure out that it also needs avfilter & with_sdl. So I'd rather keep it documented somewhere.

CI/ios/before_install.sh
kambala-decapitator: single script makes complete sense, I fully agree

IvanSavenko:
Addressed all comments.

Still need rebuild packages as xz archives & test android build, but other than that - should be ready.

conanfile.py
```diff
@@ -87,24 +88,64 @@ def configure(self):
...
         self.options["ffmpeg"].with_openh264 = False
         self.options["ffmpeg"].with_openjpeg = False
         self.options["ffmpeg"].with_opus = False
         self.options["ffmpeg"].with_programs = False
+        self.options["ffmpeg"].with_sdl = False
         self.options["ffmpeg"].with_ssl = False
         self.options["ffmpeg"].with_vorbis = False
+        self.options["ffmpeg"].with_zlib = False
+        if self.settings.os != "Android":
+            self.options["ffmpeg"].with_libfdk_aac = False
```
kambala-decapitator, line 123: @IvanSavenko could you remind the reason of this?

IvanSavenko: I think this option simply does not exists on Android.
And in conan disabling non-existing option is illegal, so I had to add Android check.

PR #4481, opened 2024-08-20
fix for 8th creature

Laserlicht:


client/windows/CCastleInterface.cpp
```diff
@@ -1906,12 +1906,10 @@ const CBuilding * CFortScreen::RecruitArea::getMyBuilding()
...
-	while (town->town->buildings.count(myID))
-	{
-		if (town->hasBuilt(myID))
-			build = town->town->buildings.at(myID);
-		myID.advance(town->town->creatures.size());
-	}
+	myID = BuildingID::getDwellingFromLevel(level, 1);
+	if (town->hasBuilt(myID))
+		build = town->town->buildings.at(myID);
+
```
IvanSavenko, line 1912: I think it would show incorrect dwellings for pirates (3rd upgrade) in cove and for alternative creatures from tow

Laserlicht: Thanks. Will recheck later.

Laserlicht:
Not nice. But works.

Hardcoded Enum logic is a pain... 
Why didn't the OH3 developers make a placeholder for an 8th town? :)

PR #4483, opened 2024-08-21
Customizable objects in RMG zones

DjWarmonger:
This PR is ready now. Configuration for special objects will come in separate PR.

- [x] Banning objects per zone
- [x] Copying object settings between zones
- [x] Customized object frequency, value and count

- Additionally, fixed rmg values of Pandoras with creatures - they were 2/3 of SoD values, making game too easy.


lib/rmg/modificators/TreasurePlacer.cpp
```diff
@@ -1180,4 +1197,89 @@ void TreasurePlacer::ObjectPool::discardObjectsAboveValue(ui32 value)
 	});
 }
 
+ObjectConfig::EObjectCategory TreasurePlacer::ObjectPool::getObjectCategory(CompoundMapObjectID id)
```
krs0, line 1200: Use a map for trivial cases? It will get rid of most of If / else if

DjWarmonger: I'm afraid it will get non-trivial very soon.

config/schemas/template.json
```diff
@@ -49,6 +50,9 @@
 						},
 						"additionalProperties" : false
 					}
+				},
+				"customObjects" : {
+					"type" : "object"
```
IvanSavenko, line 55: Would be nice to actually validate this entry

PR #4485, opened 2024-08-22
Optimize CMap::isInTheMap

kaja47:
replace x >= 0 && x < size by (unsigned)x < size

By converting signed coordinate to unsigned number, negative values became very large positive ones, larger than every positive signed number and therefore also bigger than the map size. As a result check against size also implicitly checks if coordinate is negative.

Compiler cannot do this transformation automatically because it doesn't know that map dimensions are always positive.

The change shrinks isInTheMap from 19 instructions to 11 on x86.

### DIFF
--- a/lib/mapping/CMap.cpp
+++ b/lib/mapping/CMap.cpp
@@ -332,15 +332,6 @@ bool CMap::isCoastalTile(const int3 & pos) const
 	return false;
 }
 
-bool CMap::isInTheMap(const int3 & pos) const
-{
-	// Check whether coord < 0 is done implicitly. Negative signed int overflows to unsigned number larger than all signed ints.
-	return
-		static_cast<uint32_t>(pos.x) < static_cast<uint32_t>(width) &&
-		static_cast<uint32_t>(pos.y) < static_cast<uint32_t>(height) &&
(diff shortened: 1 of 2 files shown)


DjWarmonger:
Nice one. Did you do any profiling on how much ths method is actually used?

IvanSavenko:
A lot actually. Like for every fow check, for every tile access check, etc.
Don't expect noticeable performance improvement, but still a nice change.

IvanSavenko:
Perhaps also move it into header as inlined function? Because right now compiler is unlikely to inline it, especially for calls from client or AI (which are in different dll/so).

kaja47:
During computer players' turn on G sized map, the game spends 0.99% time in isInTheMap function and 0.28% in plt stub. If the function is moved to the header file and inlined, we can get rid of most of this overhead.

Funnily enough, after inlining code of CPathfinderHelper::getNeighbours which calls isInTheMap gets very slightly smaller. This is somewhat relevant because topdown profiling by perf shown that 16 - 25% of issue slots are wasted because of frontend bottlenecks indicating that the code might be bit too big and/or too indirect for my puny machines to smoothly run. Still the main bottleneck is memory because AI allocates about 300MB of state for G map with underground and there's no locality when we have ~2 kB of state per tile.

PR #4496, opened 2024-08-22
Statistics improve

Laserlicht:
- add grid for weeks/months
- fix touchscreen indicator on mobile

grafik

### DIFF
--- a/Mods/vcmi/config/vcmi/german.json
+++ b/Mods/vcmi/config/vcmi/german.json
@@ -163,7 +163,7 @@
 	"vcmi.systemOptions.townsGroup" : "Stadt-Bildschirm",
 
 	"vcmi.statisticWindow.statistics" : "Statistik",
-	"vcmi.statisticWindow.tsvCopy" : "Daten in Zwischenabl.",
+	"vcmi.statisticWindow.tsvCopy" : "In Zwischenablage",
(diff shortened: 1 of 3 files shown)


DjWarmonger:
How about thousands separator for large numbers?

IvanSavenko:
Perhaps also add horizontal grid?
You'll need to come with some way to generate number of steps. For values in 1-10 range we might want line per value, but for values that can reach millions we want smaller number of lines.
Probably something like this?
```cpp
int computeGridStep(int maxAmount)
{
	constexpr int linesLimit = 10; 

	for (int lineInterval = 1;;lineInterval *= 10)
	{
		for (int factor : { 1, 2, 5 } )
		{
			int lineIntervalToTest = lineInterval * factor;
			if (maxAmount / lineIntervalToTest <= linesLimit)
				return lineIntervalToTest;
		}
	}
}
```
(untested)
Basically, use set of human-readable numbers (1, 2, 5, 10, 20, 50, 100...) until we find one that would result in nice-looking plot

Laserlicht:
> (untested) Basically, use set of human-readable numbers (1, 2, 5, 10, 20, 50, 100...) until we find one that would result in nice-looking plot

Thanks! Implemented. Also adapted maxValue to the grid.

> How about thousands separator for large numbers?

Use `formatMetric` for big numbers

grafik

Issue #4446, opened 2024-08-14
Crash in AI with Orb of Inhibition


by003:
**Describe the bug**
There is a crucial bug in 1.5.6, For the reason of blocked spells.

Actually it only occurs in AI turn.

**To Reproduce**
1. Make an AI hero, who has Orb of Inhibition in his/her misc blanks.
![image](https://github.com/user-attachments/assets/da468707-54e7-4db3-841f-1bebb5d514b2)
2. Make a player hero, who can use any spells.
3. Let AI hero attacks player hero
4. Click the spell button during the battle.
![image](https://github.com/user-attachments/assets/e337a675-4c85-4e15-b8da-c10458d534a5)
5. Expected behavior: the message shows the hero cannot open the spell book.
6. Actual behavior: The hero still cannot open spell book, but no message.
7. Click escape button, and replay the battle
8. See error, the game crashes

Actually if player hero defeats the AI hero, in next turn, the message will show several times, which depends on how many times player has clicked the button. 

The same condition in Recanter's Cloak.
![image](https://github.com/user-attachments/assets/62812dce-8a6c-49b9-b4fa-11884d4b2fe6)

**Did it work earlier?**
In 1.5.5, there is no problem like this.

**Version**
 - OS: Android. Windows
 - Version: 1.5.6 formal version & 1.6.0 test version

IvanSavenko:
Not reproducible for me for some reason.
I was attacked by AI with orb of inhibition, pressing spellbook icon correctly shows error message. 1.5.6 release.

by003:
Hello @IvanSavenko 
I show a test map here.
[2.zip](https://github.com/user-attachments/files/16706619/2.zip)

The video is here:

https://github.com/user-attachments/assets/02d54c9e-2105-4712-a234-5b14401fb7f2

You can try it with 1.5.6 release or 1.6.0 test version. 
![image](https://github.com/user-attachments/assets/f69611ea-5ec6-433d-b3ca-747322702f10)

by003:
I have tested from this version.
[VCMI-WindowsAlternate.exe](https://github.com/vcmi/vcmi/releases/download/1.5.6/VCMI-WindowsAlternate.exe)

[...]

PR #4500, opened 2024-08-23
Allow dialogs when in battle interface but not on the players turn

MichalZr6:
This change may allow cases like those in issue #4446 to be playable, but it does not fully resolve the underlying problem that leads to crashes.

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -1013,7 +1013,7 @@ void CPlayerInterface::showInfoDialog(const std::string &text, const std::vector
 	}
 	std::shared_ptr<CInfoWindow> temp = CInfoWindow::create(text, playerID, components);
 
-	if (makingTurn && GH.windows().count() > 0 && LOCPLINT == this)
+	if ((makingTurn || battleInt) && GH.windows().count() > 0 && LOCPLINT == this)
 	{
 		CCS->soundh->playSound(static_cast<soundBase::soundID>(soundID));
 		showingDialog->setBusy();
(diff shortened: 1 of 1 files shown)


client/CPlayerInterface.cpp
```diff
@@ -1013,7 +1013,7 @@ void CPlayerInterface::showInfoDialog(const std::string &text, const std::vector
 	}
 	std::shared_ptr<CInfoWindow> temp = CInfoWindow::create(text, playerID, components);
 
-	if (makingTurn && GH.windows().count() > 0 && LOCPLINT == this)
+	if ((makingTurn || battleInt) && GH.windows().count() > 0 && LOCPLINT == this)
```
IvanSavenko, line 1016: Can you check this version?
```suggestion
	if ((makingTurn || battleInt->curInt.get() == this) && GH.windows().count() > 0 && LOCPLINT == this)
```

MichalZr6: Sure. Currently AFK. Will fix in about 1 hr

MichalZr6: It works with some additional checks that `battleInt` and `battleInt->curInt` are not null. Otherwise exceptions are thrown.

dydzio0614:
@IvanSavenko what do you think about targetting beta with this fix and adding it to 1.5.7?

IvanSavenko:
I guess that's an option. In that case will move 1.5.7 to Monday.

This piece of code needs more detailed investigation - the way we handle dialogs queue is very weird and overcomplicated, but for 1.5.7 simple fix will do.

IvanSavenko:
Will cherry-pick to beta

PR #4504, opened 2024-08-23
disintegrate bonus

Laserlicht:
...after death no corpse remains...

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -660,5 +660,7 @@
 	"core.bonus.WATER_IMMUNITY.name": "Water immunity",
 	"core.bonus.WATER_IMMUNITY.description": "Immune to all spells from the school of Water magic",
 	"core.bonus.WIDE_BREATH.name": "Wide breath",
-	"core.bonus.WIDE_BREATH.description": "Wide breath attack (multiple hexes)"
+	"core.bonus.WIDE_BREATH.description": "Wide breath attack (multiple hexes)",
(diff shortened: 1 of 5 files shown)


dydzio0614:
does it correctly count these creatures to battle casualties after battle?

Laserlicht:
> does it correctly count these creatures to battle casualties after battle?

Yes. Tested this one. Other things to test?

IvanSavenko:
Necromancy perhaps? Can't think of any other cases.

by003:
Disintegrate bonus might be mainly used in Ring of Oblivion.
Actually it blocks any 
resurrection in battle, which cannot be resurrected and raised after battle by necromancy.

Laserlicht:
Yes this PR based on  @by003 ideas for bonuses as it was easy to integrate.

Alucard648:
Will it be possible to add this attribute to creature and spell ability? Like H4 didintegration spell.

by003:
Send my suggestions for subtype here:
disintegrateAbsolute: used for Ring of Oblivion, it will block any resurrection ability and cannot be drained blood. Then, it blocks necromancy after battle.
disintegrateNormal: used for automaton in detonation ability, it can be resurrected unless be eliminated.

If disintegrate spell needs to be implemented, a new bonus 'DISINTEGRATE_ATTACK' is needed.It can only affect the part which is eliminated by the ability owner. Like a spell or a creature.If all eliminated quantities are caused by ability owner, no body remains.

by003:
Disintegrate bonus icon is finished.
image

Anyone can add it into vcmi extra.

PR #4519, opened 2024-08-27
Fix server crash on mutual paralysis

smanolloff:
### Overview

This PR fixes a server-side crash when all units on the battlefield are inactive.

Error logs:
```
[runServer][global] ERROR Disaster happened.
[runServer][global] ERROR Reason: Failed to find valid stack to act!
```

### Steps to reproduce

Not too easy to reproduce due to low cast chance of paralyze/petrify:

1. Engage into a battle with only 2 stacks: scorpicore and medusa.
1. Attack with scorpicore, paralyzing the medusa. Medusa retaliates, petrifying the scorpicore.

### Changes summary

Added a loop which skips up to 3 rounds (duration of petrify/blind/paralyze) before throwing the exception.

The client does not properly update the UI - scorpicore color is gray as if petrified, but is otherwise active as normal and restores its color on any action. I don't know how to fix that, but it's a minor issue (does not break the game). The message bar shows "Next round begins" several times in a row, which is expected.

image

### DIFF
--- a/server/battles/BattleFlowProcessor.cpp
+++ b/server/battles/BattleFlowProcessor.cpp
@@ -304,10 +304,20 @@ void BattleFlowProcessor::activateNextStack(const CBattleInfoCallback & battle)
 		if (!next)
 		{
 			// No stacks to move - start next round
-			startNextRound(battle, false);
-			next = getNextStack(battle);
-			if (!next)
-				throw std::runtime_error("Failed to find valid stack to act!");
+			// If no units are able to act, wait up to 3 rounds
+			// Example: scorpicore + medusa
+			// Scorpicore paralyzes medusa, medusa retaliates
+			// and petrifies scopricore => no units can act for 3 rounds
+			int skippedRounds = 0;
+			while(true)
+			{
+				startNextRound(battle, false);
+				next = getNextStack(battle);
+				if (next) break;
+				if (++skippedRounds == 3)
+					throw std::runtime_error("Failed to find valid stack to act for 3 consecutive rounds!");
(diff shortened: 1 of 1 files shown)


DjWarmonger:
Alright, but paralyzed unit should not retaliate.

IvanSavenko:
1. As Warmonger said, the bug is someplace else - paralyzed unit should not retaliate
2. In theory, it is possible to achieve such situation for any number of rounds:
- Have enemy unit with Magic Mirror on it, like Faerie Dragons
- Cast Blind on it once. Assuming that reflect roll failed and unit was blinded / inactive
- Cast Blind on it again. Assume that reflect roll succeed and our own unit was blinded instead
Now we have both our unit and enemy unit blinded. If they are the only units in combat then nobody will act for N rounds, where N = spellpower of our hero.

smanolloff:
> Alright, but paralyzed unit should not retaliate.

At first I thought exactly the same.
Then I read this [wiki](https://heroes.thelazy.net/index.php/Manticore_and_Scorpicore#Special_Ability:_Paralyzing_Venom) and decided I was wrong. Was I?

kambala-decapitator:
paralyzed units retaliate, petrified units retaliate only after being attacked the next time

smanolloff:
> paralyzed units retaliate, petrified units retaliate only after being attacked the next time

So... is this _how it should be_, or _how it currently is_ ? It certainly is not both, so please clarify further.

kambala-decapitator:
how it should be

IvanSavenko:
I would rather double-check H3 behavior to be safe

PR #4524, opened 2024-08-27
grotto support

Laserlicht:
Support for grotto.

Many thanks to @SoundSSGood to refactor markets!

@misiokles @IvanSavenko  Can you create hota `vcmi-1.6` branch?


misiokles:
Already created!

IvanSavenko:
Kind of superseded by #4502 actually (sorry). 

Although I went different route and instead removed all remaining hardcoded checks for things like artifact merchant. Now UI will be automatically selected based on market mode provided by building

Haven't checked grotto specifically, but it should work provided that "marketModes" field is set in building config.

Laserlicht:
Okay, no problem :)

What change is needed to make it work?
https://github.com/vcmi-mods/horn-of-the-abyss/commit/2affb5bc7a2e4154fabd9d191571797a846adaf9

Just remove `type`?

IvanSavenko:
Yep. Remove type.
Should work as it as well, but would result in error from json validator.

PR #4528, opened 2024-08-28
Fix crashes on AI turn

MichalZr6:
This PR fixes:
1. Crash when AI attacks player's town who has turret control (crashes on evaluation of `reachabilityMap.at(hex)` when hex is -2).
2. Crash when `HeroManager::update()` specifically on line `heroRoles.clear()`.


IvanSavenko:
> Crash when HeroManager::update() specifically on line heroRoles.clear().

How this could've lead to a crash?

MichalZr6:
There are nullptr as keys.

nullkiller:
This might be caused by this HeroPtr magic. So if hero is killed HeroPtr becomes invalid while const CGHeroInstance * will still be ok. Never liked it and removes from codebase time to time.

nullkiller:
Maybe we can either use hero ObjectInstanceID instead.

nullkiller:
Still it might be that we just need to lock in RecruitHero::accept when we do heroManager->update()
std::unique_lock lockGuard(aiStateMutex);

IvanSavenko:
Yeah, this sounds more like multithreading issue than std::map / nullptr one.
@MichalZr6 I think that change needs to be replaced by a lock, like nullkiller suggested.

MichalZr6:
I felt like this HeroPtr (or related code's) refactor is going to be a better solution. And better solution... is better solution 🙂

AI/BattleAI/BattleExchangeVariant.cpp
```diff
@@ -484,15 +484,18 @@ ReachabilityData BattleExchangeEvaluator::getExchangeUnits(
 		vstd::concatenate(allReachableUnits, turn == 0 ? reachabilityMap.at(hex) : getOneTurnReachableUnits(turn, hex));
 	}
 
-	for(auto hex : ap.attack.attacker->getHexes())
+	if(!ap.attack.attacker->isTurret())
```
nullkiller, line 487: I integrated this fix and added lock in RecruitHero in my branch.

IvanSavenko:
Closing, included in #4551

PR #4530, opened 2024-08-28
Support for configurable town fortifications

IvanSavenko:
Removed most of hardcoded checks for fort level or for presence of fort/ citadel/castle buildings.

It is now possible to define which parts of town fortifications are provided by town buildings

Configuration for H3-like fortifications is provided in buildingsLibrary.json and will be used automatically by mods as long as mods have buidings named "fort", "citadel" and "castle".

Alternatively, mods can separately define:
- hitpoints of walls (shared value for all sections)
- hitpoints of central, upper and lower towers (separate values)
- presence of moat
- shooters for each tower (separate values)

### DIFF
--- a/AI/BattleAI/BattleAI.cpp
+++ b/AI/BattleAI/BattleAI.cpp
@@ -229,7 +229,7 @@ BattleAction CBattleAI::useCatapult(const BattleID & battleID, const CStack * st
 		{
 			auto wallState = cb->getBattle(battleID)->battleGetWallState(wallPart);
 
-			if(wallState == EWallState::REINFORCED || wallState == EWallState::INTACT || wallState == EWallState::DAMAGED)
+			if(wallState != EWallState::NONE && wallState != EWallState::DESTROYED)
 			{
 				targetHex = cb->getBattle(battleID)->wallPartToBattleHex(wallPart);
 				break;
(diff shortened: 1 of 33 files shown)


Alucard648:
Will it be feastible to rework structure damage in a way similar to H5?

IvanSavenko:
Can you describe it? Don't remember such details.

Alucard648:
I mean give walls actual HP, instead of just structural states. So walls had more variety in durability between towns. So dwarven rune-enchanted Fortress walls can easily withstand 5-6 direct catapult hits before giving away a breach. And flimsy barbarian castle walls can be 1-thit killed with single catapult shot, even if fully upgraded.

dydzio0614:
i would say such features may start the trend of going too far into mixing H3 mechanics with H5 mechanics for now - there are bunch of more wanted stuff to deliver in VCMI before going for such fluff features

IvanSavenko:
Should be already possible.
Walls/towers can have up to... 127 hp I think? 
However there is no visual indication other than intact / damaged / destroyed. So if 10-hp wall appears as damaged in combat it can have 1-9 HP.

1-hp walls should also be possible.

dydzio0614:
he means stuff like catapult dealing 50-200 damage while wall has 1000 so it is "more random" etc

PR #4531, opened 2024-08-28
map drag with right mouse

Laserlicht:
grafik


Laserlicht:
@IvanSavenko Any open points here?

PR #4533, opened 2024-08-29
Change logic of neutral towns garrisons to match H3

### LINKED ISSUE #2000: Neutral towns use different logic for creature growth in H3
Wasn't actually aware that this is a thing, might need verification in case if I misunderstood something

Neutral towns need separate logic for creature growth. After quick test in H3 this is what I found:
- In neutral towns, creature growth is multiplied by 50% (TODO: check how this stacks with other modifiers, e.g. Castle or prebuilt hordes)
- If town garrison on new week grows and dwelling for new creature is build, recruited creatures should be substracted from accumulated growth

**To Reproduce**
1. Start on Arrogance
2. Go to town neutral to south-east from starting location
3. Wait for new week

**Expected behavior**
Exact logic should be checked in H3

**Actual behavior**
Creatures in neutral towns accumulate using their unmodified base growth.

**Did it work earlier?**
Very unlikely

**Version**
 - VCMI 1.2

IvanSavenko:
- Resolves #2000 

Now both initial guards for neutral towns as well as growth of garrison on new week should (mostly) match H3 logic. There still might be small differences, since a lot of this logic is random-based, so observing H3 only allows to determine approximate changes / logic, but should be close enough to original game.

Also fixed two visions-related bugs that I discovered while testing these changes:
- Visions was not revealing town information
- Right-click on neutral town would lead to a crash (probably only with visions) - regression from previous PR

### DIFF
--- a/client/windows/InfoWindows.cpp
+++ b/client/windows/InfoWindows.cpp
@@ -271,7 +271,7 @@ CInfoBoxPopup::CInfoBoxPopup(Point position, const CGTownInstance * town)
 	: CWindowObject(RCLICK_POPUP | PLAYER_COLORED, ImagePath::builtin("TOWNQVBK"), toScreen(position))
 {
 	InfoAboutTown iah;
-	LOCPLINT->cb->getTownInfo(town, iah, LOCPLINT->localState->getCurrentTown()); //todo: should this be nearest hero?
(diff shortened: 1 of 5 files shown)


dydzio0614:
Can you explain the actual logic you implemented?

IvanSavenko:
> Can you explain the actual logic you implemented in words?

As per comment in linked issue:
https://github.com/vcmi/vcmi/issues/2000#issuecomment-1655419799

PR #4535, opened 2024-08-29
Removed save compatibility with 1.4

IvanSavenko:
All save compatibility checks targeting 1.4 saves have now been removed. Saves from 1.5 can still be loaded in 1.6

Implemeted few TODO's in serialization that were postponed to avoid breaking save compatibility in MP for 1.5.X releases.

Fixed missed case for loading black market object from 1.5 saves

### DIFF
--- a/lib/mapObjects/IMarket.h
+++ b/lib/mapObjects/IMarket.h
@@ -36,6 +36,11 @@ public:
 	CArtifactSet * getArtifactsStorage() const;
 	bool getOffer(int id1, int id2, int &val1, int &val2, EMarketMode mode) const; //val1 - how many units of id1 player has to give to receive val2 units
 
+	template <typename Handler> void serializeArtifactsAltar(Handler &h)
+	{
+		h & *altarArtifactsStorage;
+	}
+
 private:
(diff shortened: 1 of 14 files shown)


lib/mapObjects/IMarket.h
```diff
@@ -36,6 +36,11 @@ class DLL_LINKAGE IMarket : public virtual Serializeable, boost::noncopyable
 	CArtifactSet * getArtifactsStorage() const;
 	bool getOffer(int id1, int id2, int &val1, int &val2, EMarketMode mode) const; //val1 - how many units of id1 player has to give to receive val2 units
 
+	template <typename Handler> void serializeArtifactsAltar(Handler &h)
+	{
+		h & *altarArtifactsStorage;
```
SoundSSGood, line 41: why should we save artifacts storage?

IvanSavenko: Because we did so in 1.5 saves.

This code will never be called when saving new games in 1.6 or when loading 1.6 save and will be removed once we drop support for 1.5 saves.

PR #4540, opened 2024-08-30
Convert creature banks to rewardable objects

IvanSavenko:
Rewardable objects can now define guards as part of reward. If rewardable object has guards set, player must defeat guards before receiving rewards. 

This allow to define creature banks as a rewardable object, and give any reward that is currently possible in rewardable objects as reward for bank clearing (such as terrain reveal for that hota bank).

Since rewardable objects have quite different config format and to avoid mod breakage, old functionality for banks will stay for now, but will be removed later, e.g. in 1.6

Configurable obstacles for combats in banks are not planned (at least not by me / not now). And should be done as battlefield type, not as part of bank/rewardable format

Progress:
- [x] Allow specifying guards in rewardable configuration
- [x] Show message before confirming battle
- [ ] Show guards in battle confirmation message using same logic as bank
- [ ] Show guards in right-click popup using same logic as bank
- [ ] Check whether upgrade chance parameters works for guards in a rewardable object
- [ ] Allow specifying guards layout type in config
- [ ] Allow defining new guards layout by mods
- [ ] Implement 'coast visitable' mode for rewardable objects
- [ ] Convert all H3 banks to rewardable objects
- [ ] Update schemas, make sure that old bank handler is reported by validator
- [ ] Update documentation, add example of bank in new format and guide on converting old banks


dydzio0614:
I think it would be worth it to spend even more time to allow customizing obstacles, like here: 
obraz

IvanSavenko:
Obstacles - not this time. Completely separate functionality, and not one I am familiar with (although probably nothing complex in there).

dydzio0614:
Will configurable bank guardian positions support more than 7 stacks?

dydzio0614:
already got question if there will be support for bank that will allow 100 onestacks of hota sea dogs :rofl:

IvanSavenko:
No, 7 stacks limit for now.
I guess you can get more, e.g. by having creature with GUARDIAN bonus, but armies have limit of 7 slots.

We could try to remove it, but not sure how difficult it will be so low priority for me.

PR #2247, opened 2023-06-21
Fix accumulated issues with pathfinding

IvanSavenko:
Fixes several pathfinding-related issues accumulated over time.
 Progress:
- [x] #1402 
- [x] #1723
- [x] #1747
- [x] #1975
- [x] #2065
- [x] #1320
- [x] #2192
- [x] Move pathfinder settings to GameSettings

Issue #3286, opened 2023-12-10
Path is not updated after giving Angel Wings for quest


DjWarmonger:
**Describe the bug**
After returning Angel WIngs to Seer Hut player is no longer able to fly. However, flight path is not pdated.

**To Reproduce**
Campaign scenario

**Expected behavior**
Invalid path should be removed

**Screenshots**
https://github.com/vcmi/vcmi/assets/2566990/85f3e9f3-ed85-4e58-b5bf-3da0b85b9e5c

kambala-decapitator:
it should have been fixed with #2247 , but seems not

IvanSavenko:
That fix should work if you've unequipped artifact manually, but apparently does not covers removal via quest

PR #4542, opened 2024-08-30
Update displayed hero path when hiring/dismissing creatures that give move penalty

### LINKED ISSUE #4466: Buying creature that gives terrain penalty from mercenary camp (probably dwellings as well) doesn't invalidate pathfinder cache
As in title - if all your troops are terrain-native and you buy creature giving terrain penalty then pathfinder data does not refresh

I guess witch hut with pathfinding might also be worth checking...

MichalZr6:
~This PR extends this fix https://github.com/vcmi/vcmi/pull/2247/commits/6546242c0360f0e37f6da26355f7bf913f864be2~

Displayed hero path on AdventureMapInterface will now be updated when:
- buying creatures in dwellings (level 2 and above) that give movement penalty
- buying creatures in refugee camps that give movement penalty
- dismissing creatures which should take off movement penalty
- giving artifacts like Angel Wings to Seer's Huts

Edit: Resolves #4466, #3286


client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -168,7 +166,6 @@ void AdventureMapInterface::show(Canvas & to)
 {
 	CIntObject::show(to);
 	dim(to);
-	LOCPLINT->cingconsole->show(to);
```
IvanSavenko, line 171: Why was this removed?

client/adventureMap/AdventureMapInterface.cpp
```diff
@@ -159,6 +155,8 @@ void AdventureMapInterface::deactivate()
 
 void AdventureMapInterface::showAll(Canvas & to)
 {
+	refreshDisplayedPath();
+
```
IvanSavenko, line 159: Might be better to place it in required netpack appliers / CPlayerInterface.
showAll() should be used only to display stuff.

MichalZr6: Hi Ivan,

I have some doubts if this is the proper way to deal with that. After digging into the code got some observations:

- Shutting down InfoWindow updates the displayed route properly. That's why it works for level 1 dwellings.

- Interface not updated after closing CWindowObject is getting the update on closing (any) popup window.

It looks like the issue might be related to how the display is refreshed when windows are closed. In the destructor of CWindowObject, I see that show() is called to change the cursor visibility:

```
CWindowObject::~CWindowObject()
{
    if (options & RCLICK_POPUP)
        CCS->curh->show();
}
```
Inside show(), the SDL_ShowCursor() function is called, which seems to be triggering some kind of screen refresh:

```
void CursorHandler::show()
{
    if (showing)
        return;

    showing = true;
    cursor->setVisible(true);
}
```

And:

```
void CursorHardware::setVisible(bool on)
{
    GH.dispatchMainThread([on]()
    {
        if (on)
            SDL_ShowCursor(SDL_ENABLE);
        else
            SDL_ShowCursor(SDL_DISABLE);
    });
}
```

Perhaps there should be some function that calls SDL to be updated just before leaving destructors body ? Or this is due to some threads code ?

I also tried to get it done how `InfoWindow::close()` does it but with no success :(

Let me know which approach we should take to resolve this issue — the current approach or the one related to above.

dydzio0614: ok so... it should be fixed like here https://github.com/vcmi/vcmi/pull/3685/commits/96b18b1b8e841bd4f3697e494be68e48a5684ccb

IvanSavenko: Why do you need to do this on infowindow closing?

When hero accepts creatures from level 1 dwelling game sends RebalanceStacks pack.
This will in turn call ApplyClientNetPackVisitor::visitRebalanceStacks on client
Which calls CPlayerInterface::garrisonsChanged

Updating path must be done in one of these methods. And `visitRebalanceStacks` does have cl.invalidatePaths()

When hero purchases creatures from a dwelling, game sends InsertNewStack pack, which also results in the same call chain including cl.invalidatePaths()

Judging from your changes, I suspect what needs to be changed is that cl.invalidatePaths should also call that verifyPath method

dydzio0614 (approved):

MichalZr6:
Perhaps `updatePath` instead `verifyPath` would be more suitable naming ?

client/NetPacksClient.cpp
```diff
@@ -230,30 +231,66 @@ void ApplyClientNetPackVisitor::visitEraseStack(EraseStack & pack)
...
 	cl.invalidatePaths(); //it is possible to remove last non-native unit for current terrain and lose movement penalty
+	
+	if(LOCPLINT)
+	{
+		for(const auto hero : LOCPLINT->localState->getWanderingHeroes())
+		{
+			if(hero->id == pack.army)
+				LOCPLINT->localState->verifyPath(hero);
+		}
+	}
```
IvanSavenko, line 242: Perhaps move this copy-pasted code into CClient::invalidatePaths?

PR #4544, opened 2024-08-31
Per-map game settings

IvanSavenko:
WIP
Game settings that are currently defined in `config/gameConfig.json` can now be defined on per-map basis and not only in mod. This allows to change these settings in map, map template, or perhaps even in UI (for those allow cheats / infinite replay options)

Progress:
- [x] Store and access game settings via gamestate instead of VLC
- [x] Allow overriding game settings with custom values 
- [x] Serialize only overriden settings,
- [x] Use string ID's for serialization for better save format stability (or serialize as json)
- [x] Implement 'no special weeks' option from hota map format as an override of game setting
- [x] Allow random map templates to define own game setting overrides, like max number of heroes.
- [ ] Integrate allow cheats / infinite replay options into game settings
- [ ] Try to remove unnecessary access to VLC game settings, e.g. from DamageCalculator which only has access to battle callback
- [ ] Clearly define (separate enum? Just comment in config?) which settings can be defined in map and which are not

Not planned by me / outside PR scope
- Allow defining map settings in map editor for vcmi maps


IvanSavenko:
Game settings can now be part of rmg template:
```json
	"Jebus Cross":
	{
		"settings" : {
			"heroes" : {
				"perPlayerOnMapCap" : 1
			}
		}
```

PR #4545, opened 2024-08-31
Update conan prebuilt package

IvanSavenko:
Another update of conan package.

Changes:
- mingw runners now use ubuntu-24.04 image with newer mingw and no workarounds
- android CI now runs on linux image (ubuntu-24.04) instead of macos
- removed separate before_install.sh for platforms that now have identical setup, removed it from android runners since it is no longer needed.
- android aab's are now generated only on master branch (TODO for future - unify checks for master branch in CI)
- updated SDL library & its component to latest version from conan
- removed some old & unused leftovers from coverity scripts

TODO:
- [x] Test resulting builds
- [x] Cleanup branch, replace WIP commit with meaningful ones
- [x] Update docs on supported audio and image formats


.github/workflows/github.yml
```diff
@@ -343,11 +364,6 @@ jobs:
     steps:
         - uses: actions/checkout@v4
 
-        - uses: actions/setup-python@v5
-          if: "${{ matrix.conan_profile != '' }}"
-          with:
-            python-version: '3.10'
```
kambala-decapitator, line 349: does Conan work fine now without this?

CI/before_install/linux_qt5.sh
```diff
@@ -9,6 +8,6 @@ sudo apt-get update
 # - debian build settings at debian/control
 sudo apt-get install libboost-dev libboost-filesystem-dev libboost-system-dev libboost-thread-dev libboost-program-options-dev libboost-locale-dev libboost-iostreams-dev \
 libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev \
-qtbase5-dev \
+qtbase5-dev qttools5-dev \
 ninja-build zlib1g-dev libavformat-dev libswscale-dev libtbb-dev libluajit-5.1-dev \
-libminizip-dev libfuzzylite-dev qttools5-dev libsqlite3-dev # Optional dependencies
+libminizip-dev libfuzzylite-dev libsqlite3-dev # Optional dependencies
```
kambala-decapitator, line 13: Linux Qt 5/6 scripts should also be merged at some point to avoid copy-paste

CI/before_install/mingw_x86_64.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+sudo apt-get update
+sudo apt-get install ninja-build mingw-w64 nsis
+sudo update-alternatives --set x86_64-w64-mingw32-g++ /usr/bin/x86_64-w64-mingw32-g++-posix
```
kambala-decapitator, line 5: mingw scripts can also be merged

conanfile.py
```diff
@@ -15,10 +15,10 @@ class VCMI(ConanFile):
...
     ]
     _clientRequires = [
-        "sdl/[~2.26.1 || >=2.0.20 <=2.22.0]", # versions in between have broken sound
-        "sdl_image/[~2.0.5]",
-        "sdl_mixer/[~2.0.4]",
-        "sdl_ttf/[~2.0.18]",
+        "sdl/[>=2.26.1 || >=2.0.20 <=2.22.0]", # versions in between have broken sound
+        "sdl_image/[>=2.0.5]",
+        "sdl_mixer/[>=2.0.4]",
+        "sdl_ttf/[>=2.0.18]",
```
kambala-decapitator, line 21: `>=` is dangerous because it allows to use completely untested v3. IIRC `^` locks major version,

.github/workflows/github.yml
IvanSavenko: Almost. Had to replace pip3 with pipx - requirement for newer Ubuntu, but both mac & ubuntu hosts work. Windows/msvc does not needs python.

Haven't checked mac os image, but ubuntu image comes with preinstalled python of roughly the same version.

conanfile.py
IvanSavenko: @kambala-decapitator 
I get what you're saying but not sure which approach would be better.

Examples of various 'bad' scenarios that we had over time:
- boost 1.86 - our conanfile permits 1.86, but it was not tested, so would have caused breakage if somebody were to use conan with no prebuilts package and conan downloaded 1.86
- sdl - for Android must be exact match, since its version is hardcoded in SDL wrapper for java code. So must be changed in sync. 
- sdl - 2.22 had broken sound bug we had some time ago, that first appeared on Debian
- sdl_image - there was hard to figure out regression on Flatpak due to broken png loading because of minor version update.
- ffmpeg seems to have rather stable API now, only one minor change despite 4.4 -> 7.0 jump, but was quite bad before.

So, approach that you're proposing is to set restrictions that are extremely unlikely to cause regressions.
Pros: less chance of regressions
Cons: if somebody were to try to build vcmi with newer libs he will run into bugs - 1.86 update on Arch, sdl_image update to 2.6 on Flatpak some time ago, sdl update to 2.22 on Debian before that. 

Alternatively, we could try to exclude versions that are known to cause problems and permit everything else. Combined with daily/weekly/monthly automated rebuild test job it should at the very least detect any new compilation / linking issues.
So in case of weird bug from some bleeding-edge system (Arch, Flatpak, unstable Debian) we could grab package from this scheduled rebuild and try to reproduce it.

Might change with artifactory / drop of prebuilts package, but at least right now it won't cause regressions as long as package *has* been tested - since we have `build=never` in CI, so conan will never grab newer version unless we added this newer version into the package.

Opinions?

CI/before_install/linux_qt5.sh
IvanSavenko: Or perhaps drop qt5 and always use qt6 for these CI's? They don't produce any artifacts and exists only to run tests & check different compilers. All conan-based CI's use qt5 so any qt5-specific issues will still be found

kambala-decapitator: don't think it would hurt to keep Linux Qt 5 as a sanity check

CI/before_install/mingw_x86_64.sh
kambala-decapitator: yes, as I see the only different is 32/64 parameter

conanfile.py
```diff
@@ -47,21 +48,39 @@ def configure(self):
         self.options["freetype"].shared = self.settings.os == "Android"
 
         # SDL_image and Qt depend on it, in iOS both are static
-        # Enable static libpng due to https://github.com/conan-io/conan-center-index/issues/15440,
-        # which leads to VCMI crashes of MinGW
-        self.options["libpng"].shared = not (self.settings.os == "Windows" and cross_building(self)) and self.settings.os != "iOS"
+        self.options["libpng"].shared = self.settings.os != "iOS"
```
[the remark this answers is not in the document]
kambala-decapitator: not sure which limitations you mean. Currently the only real limitations are:
- Qt for iOS can only built as static
- oneTBB must be shared (their heavy recommendation)

I follow a simple rule: if a dependency is needed by more than 2 "separate" targets (e.g. an app and a shared library), then the dependency should be built as shared, otherwise static should be fine.

conanfile.py
kambala-decapitator: We can't be responsible for bugs in dependencies. Once we identify a buggy version, we can exclude it (like we already do with SDL) - I'm totally fine with that.

What I don't like is giving zero restrictions on future versions. Considering that most libs follow semver in one way or another, I'd rather lock major version of all dependencies, so that testing a new major version (e.g. SDL 3) is an explicit user's choice.

- Boost doesn't seem to follow semver at all as we see, but I don't think it's an often case that they break API like that
- SDL for android - well, if one updates only C++ part without the android one, then it's a problem of the user who didn't follow SDL manual

In the future would be useful to run Conan with newer libs versions to check if dependencies and project at least build, testing should also be done before updating binaries of course.

IvanSavenko: Actually in case of SDL 3 it is not that we have not tested it and it may not work with vcmi.
We know *for sure* that it won't work with vcmi as it: https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md
Sure, we don't use 100% of SDL functions, but amount of breaking changes in this list is quite high.

So yeah, SDL 3 fits the definition of "exclude versions that don't work with vcmi", at least until we adapt code to SDL 3.

> SDL for android - well, if one updates only C++ part without the android one, then it's a problem of the user who didn't follow SDL manual

Kind of. Right now on develop if you simply generate conan package for Android from scratch (which our docs do allow) you will end up with different version of SDL in conan.

So enforcing correct version via conan makes sense. If you want to use different one - then update both conanfile and that java wrapper.

> Considering that most libs follow semver in one way or another

- sdl - major version breaks API. A lot.
- ffmpeg - no API breakage despite 4.4 -> 7.0 upgrade. Only one minor deprecation warning.
- boost - chance of API breakage in any version
- libtbb - uses year.month for releases, not semver
- qt - Not 100% sure, but I think at least with 5 -> 6 transition there was only one minor change that affected vcmi, so not really API breakage

So as far as our dependencies go, only SDL follow semver in terms of "change in major version number breaks API". Possibly also Qt, but we only use basic stuff from it.

kambala-decapitator: Qt 5->6 requires cmake changes at least for android. For others I agree.

CI/before_install/linux_qt5.sh
IvanSavenko: Probably keeping as it for now.
One option is to install both qt5 and qt6, but then we'll need to specify which one to use in CMake
Another option is to split this script into common and qt-specific parts, but I'd rather keep things as simple as possible, without scripts calling other scripts.

If somebody else wants to change it - go ahead.

conanfile.py
IvanSavenko: Changed. 
For SDL on PC we now request specifically 2.X. On mobiles - we request specific version.
Also added a comments to explain why.

kambala-decapitator:
will check until EOW

also something to look at: https://github.com/conan-io/docs/pull/3799

conanfile.py
```diff
@@ -46,21 +47,40 @@ def configure(self):
...
-        if self.settings.os == "Android":
-            self.options["qt"].android_sdk = tools.get_env("ANDROID_HOME", default="")
-
         # TODO: enable for all platforms
         if self.settings.os == "Android":
             self.options["bzip2"].shared = True
             self.options["libiconv"].shared = True
             self.options["zlib"].shared = True
 
+        # TODO: enable for all platforms?
```
kambala-decapitator, line 60: for "single process" static should be sufficient, but could be enabled

conanfile.py
```diff
@@ -46,21 +47,40 @@ def configure(self):
...
+        if self.settings.os == "Windows":
+            self.options["sdl"].shared = True
+            self.options["sdl_image"].shared = True
+            self.options["sdl_mixer"].shared = True
+            self.options["sdl_ttf"].shared = True
+
+        if self.settings.os == "iOS": 
+            # TODO: ios - newer sdl fails to link
```
kambala-decapitator, line 68: could you point to the error?

conanfile.py
```diff
@@ -46,21 +47,40 @@ def configure(self):
...
+            self.requires("sdl/2.26.1")
+            self.requires("sdl_mixer/2.0.4")
+        elif self.settings.os == "Android":
+            # On Android SDL version must be same as version of Java wrapper for SDL in VCMI source code
```
kambala-decapitator, line 72: let's also provide relative path to the wrapper

conanfile.py
```diff
@@ -46,21 +47,40 @@ def configure(self):
...
+            self.requires("sdl/2.26.5")
+            self.requires("sdl_mixer/2.0.4")
+        else:
+            # upcoming SDL version 3.0+ is not supported at the moment due to API breakage
+            # SDL versions between 2.22-2.26.1 have broken sound
+            self.requires("sdl/[^2.26 || >=2.0.20 <=2.22.0]")
+            self.requires("sdl_mixer/[>=2.0.4]")
```
kambala-decapitator, line 79: why are we restricting mixer versions for iOS and Android?

became clear further. Better to update mixer to a later version in the future to remove that condition/options completely.

kambala-decapitator (approved):
all good in general, only minor stuff to adjust

CI/before_install/linux_qt6.sh
```diff
@@ -1,9 +1,11 @@
 #!/bin/sh
 
-sudo apt remove needrestart
 sudo apt-get update
 
 # Dependencies
+# In case of change in dependencies list please also update:
+# - developer docs at docs/developer/Building_Linux.md
```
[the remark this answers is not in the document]
IvanSavenko: This place should actually be easier to find.
Since if you were to add new library CI will fail and you'll have to update this file and (hopefully) see this message.

Can mention it to docs for completeness sake.

.github/workflows/github.yml
```diff
@@ -177,7 +195,13 @@ jobs:
       env:
         GENERATE_ONLY_BUILT_CONFIG: 1
 
-    - uses: actions/setup-java@v4
+    # Workaround for gradle not discovering SDK that was installed via conan
```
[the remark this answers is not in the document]
IvanSavenko: Using env variable in place of symlink might work, but was hoping there is better way to retrieve path to package from conan.

IvanSavenko:
CI passed, merging to fix broken CI in develop.

Will fix remaining review comments in separate PR.

conanfile.py
```diff
@@ -15,10 +15,11 @@ class VCMI(ConanFile):
         "minizip/[~1.2.12]",
     ]
     _clientRequires = [
-        "sdl/[~2.26.1 || >=2.0.20 <=2.22.0]", # versions in between have broken sound
-        "sdl_image/[~2.0.5]",
-        "sdl_mixer/[~2.0.4]",
-        "sdl_ttf/[~2.0.18]",
+        # Versions between 2.5-2.8 have broken loading of palette sdl images which a lot of mods use
+        # there is workaround that require disabling cmake flag which is not available in conan recipes.
```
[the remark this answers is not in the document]
kambala-decapitator: stb-image is an external lib, but I see that SDL can vendor it if you ask so: https://github.com/libsdl-org/SDL_image/blob/release-2.6.3/stb_image.h

Conan uses its own CML to build SDL_image: https://github.com/conan-io/conan-center-index/blob/master/recipes/sdl_image/all/CMakeLists.txt As you can see, there's no mention of STB. So I'm not sure what the issue is, Linux package?

If you want v2.8, the we can submit a PR to CCI to add it.

PR #4553, opened 2024-09-02
fixes build on clang

Laserlicht:
fix for (clang 18.1.3 build)
`error: 'result_of<void (*&())()>' is deprecated: use 'std::invoke_result' instead [-Werror,-Wdeprecated-declarations]`


IvanSavenko:
gcc bug actually:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114394

In any case, don't see anything wrong with removing std::bind usage

PR #4554, opened 2024-09-03
fixes bug for 8th creature

Laserlicht:
reported by @by003


by003:
Is this for fixing the upgrade problem for Couatls?

Laserlicht:
> Is this for fixing the upgrade problem for Couatls?

Yes

by003:
Thanks I will check it.

PR #4555, opened 2024-09-03
Improvements for fonts when xbrz is in use

IvanSavenko:
- Native H3 bitmap fonts now use nearest-neighbor upscaling instead of xbrz. Added option to config (with no UI) to select preferred upscaler for h3 fonts.
- Added options to manually define point size of true type fonts for xbrz mode, overriding internal multiplication, e.g. `"size" : [ 10, 16, 22, 26 ]` will have font at 10px in unscaled mode and 16px in xbrz2 mode instead of 10*2 = 20px
- Fixed crash on attempt to upscale nonexisting image - reported in hota mod
- Removed duplicated code in image upscaling logic
- Fixed string width computation errors sometimes leading to text overflow, e.g. in scenario description.
  - Fixed rounding error caused by xbrz scaling factor - 9px-wide symbol would be interpreted as 4px-wide instead of 4.5px wide if xbrz 2x is in use
  - String width is now computed for entire string at once using SDL_ttf instead of being sum of widths of all characters (may not match due to ttf features like hinting / fractional widths)
  - Last character in string is now properly accounted for and won't cause text overflow

### DIFF
--- a/client/renderSDL/CBitmapFont.cpp
+++ b/client/renderSDL/CBitmapFont.cpp
@@ -16,13 +16,14 @@
 #include "../render/Colors.h"
 #include "../render/IScreenHandler.h"
 
+#include "../../lib/CConfigHandler.h"
 #include "../../lib/Rect.h"
+#include "../../lib/VCMI_Lib.h"
 #include "../../lib/filesystem/Filesystem.h"
 #include "../../lib/modding/CModHandler.h"
 #include "../../lib/texts/Languages.h"
 #include "../../lib/texts/TextOperations.h"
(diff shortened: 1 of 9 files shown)


IvanSavenko:
Comparion

xbrz2 images + integer scaled H3 font:
зображення

xbrz2 images + xbrz H3 fonts:
зображення

kdmcser:
Is this bug fixed?In battle the background is still wrong.
mmexport1725358592846

IvanSavenko:
Not yet, will take a look a bit later.

PR #4556, opened 2024-09-03
Implemented explicitly visitable town buildings, e.g. hota mana vortex

IvanSavenko:
Added flag `manualHeroVisit` flag to town building. If this flag is set, then building will only be activated on click and will not give its effect on hero recrutiment, hero visit, or new day.

This allows implementing changes to Mana Vortex from HotA


dydzio0614:
there should be also mode for visiting via opening town, like in original h3

IvanSavenko:
Maybe later. 

Already found a bunch of edge cases, including some critical ones that needs to be fixed for 1.6. For example multiple rewardable buildings in town are broken if at least one of them have selectable rewards / canRefuse flag set.

Likely was broken from the very beginning - never discovered since h3 buildings were not converted to this form and there were no docs for modders to use.

PR #4563, opened 2024-09-04
Improvements to the artifact creation

SoundSSGood:
- CMap removed from CArtifactSet level
- ArtifactUtils::createArtifact ArtifactUtils::createNewArtifactInstance united. 
- Introduced cheat code "vcmiscrolls". Puts scrolls with all available spells on the selected hero.

### DIFF
--- a/lib/mapping/MapFormatH3M.cpp
+++ b/lib/mapping/MapFormatH3M.cpp
@@ -954,15 +954,15 @@ bool CMapLoaderH3M::loadArtifactToSlot(CGHeroInstance * hero, int slot)
 	// H3 bug workaround - Enemy hero on 3rd scenario of Good1.h3c campaign ("Long Live The Queen")
 	// He has Shackles of War (normally - MISC slot artifact) in LEFT_HAND slot set in editor
 	// Artifact seems to be missing in game, so skip artifacts that don't fit target slot
-	auto * artifact = ArtifactUtils::createArtifact(artifactID);
-	if(artifact->canBePutAt(hero, ArtifactPosition(slot)))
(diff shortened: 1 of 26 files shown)


lib/mapping/MapFormatH3M.cpp
```diff
@@ -954,10 +954,11 @@ bool CMapLoaderH3M::loadArtifactToSlot(CGHeroInstance * hero, int slot)
 	// H3 bug workaround - Enemy hero on 3rd scenario of Good1.h3c campaign ("Long Live The Queen")
 	// He has Shackles of War (normally - MISC slot artifact) in LEFT_HAND slot set in editor
 	// Artifact seems to be missing in game, so skip artifacts that don't fit target slot
-	auto * artifact = ArtifactUtils::createArtifact(map, artifactID);
+	auto * artifact = ArtifactUtils::createArtifact(artifactID);
```
IvanSavenko, line 957: I think this will now result in memory leak?
Since such artifact will not be added to map.
Probably needs delete in `else` block

Switching storage of artifacts in CMap to shared_ptr / unique_ptr might be better option, but is also way more complex

IvanSavenko: No, I mean in this particular piece of code. Check several lines below.

Here `addNewArtifactInstance` is only called if condition on next line is true, but artifact itself is created unconditionally

SoundSSGood: > Switching storage of artifacts in CMap to shared_ptr / unique_ptr might be better option, but is also way more complex

In todo list.

PR #4566, opened 2024-09-05
Campaign outro

Laserlicht:
Support for official video after RoE-Campaign "Song for the Father".

Also usable for own campaigns.

Also implemented AB intro video.


Laserlicht:
@IvanSavenko Not sure about last commit.

Advantage: RoE video is handled the same as AB and Chronicles - starting video with the first campaign
Disadvantage: differs from OH3 - only the two logo videos are played at start

IvanSavenko:
Probably better to keep intro as it is now. See no reason to change.

PR #4569, opened 2024-09-07
appimage support

### LINKED ISSUE #2801: AppImage support
**Describe your proposal**
Support AppImage distribution for Linux in CI.

**Additional context**
I have some test branch: https://github.com/Laserlicht/vcmi/tree/appimage
But no idea, why it's not working: https://github.com/Laserlicht/vcmi/actions/runs/6140058275/job/16658546576 (locally I was at least able to build appimage).

If someone want to try to get it ready feel free.

@Alexander-Wilms Do you have some experience in AppImage?

Laserlicht:
Fixes #2801

Basically seems to work.

Only one SSL error has to be fixed...


kambala-decapitator:
why do we need it if there's flatpak already?

Laserlicht:
> why do we need it if there's flatpak already?

The disadvantage in my opinion is that we don't have a Linux artifact. To test new PRs I always have to compile it myself.

kambala-decapitator:
Couldn't we simply publish the binary without its dependcies?

Laserlicht:
> Couldn't we simply publish the binary without its dependcies?

Would also ok, I think. But not sure how distribution independent it would be.

IvanSavenko:
Another option would be to build flatpak package as part of CI.

1. It will be (hopefully) smaller than this one - 100 Mb packed and 300 Mb unpacked image
2. It will also be useful for testing our official flatpak from flathub
3. With this approach our dependencies will be managed by flatpak (or at least I think so?)

No idea how difficult it would be to set it up compared to appimage

Laserlicht:
> Another option would be to build flatpak package as part of CI.

Good idea

kenderipa:
As a user, I would like to see an AppImage. It's also better than a Flatpak, since it doesn't create a full copy od my OS and drivers and thus allows a lightweight system tovstay lightweight.

Laserlicht:
If someone want to maintain it feel free.

PR #4585, opened 2024-09-12
Fix being stuck in blocked tiles while flying towards hero

dydzio0614:
Treat visitable tiles as if promoted to blockvis while occupied by ally / enemy hero so pathfinder does not suggest wrong path in such cases while flying with original rules


IvanSavenko:
But why tiles with hero on them are "visitable" and not "blocked visitable"?

dydzio0614:
no idea, i assumed this could be by design where tile info contains "absolute state" without factors that may change tile state temporarily

IvanSavenko:
Checked once more - I think bug is someplace else. Hero must be blockvis:
- `evaluateAccessibility` is the only method that can return EPathAccessibility::BLOCKVIS / EPathAccessibility::VISITABLE and it checks `isBlockedVisitable` for that
- `CGHeroInstance::isBlockedVisitable` returns true unconditionally

PR #4587, opened 2024-09-12
Fix crash on accessing hero preview in map setup

IvanSavenko:


pppbb:
The patch worked for me.

PR #4588, opened 2024-09-12
Intro video rework

Laserlicht:
- Moved Videos to Main Window
- Added rim to intro video
- Moved everything to 800x640 windows (in line with HD mod) instead of full screen stretched

grafik
grafik
grafik

### DIFF
--- a/clientapp/EntryPoint.cpp
+++ b/clientapp/EntryPoint.cpp
@@ -30,6 +30,7 @@
 #include "../client/render/Graphics.h"
 #include "../client/render/IRenderHandler.h"
 #include "../client/render/IScreenHandler.h"
+#include "../client/lobby/CBonusSelection.h"
 #include "../client/windows/CMessage.h"
 #include "../client/windows/InfoWindows.h"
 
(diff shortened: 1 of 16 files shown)


clientapp/EntryPoint.cpp
```diff
@@ -379,7 +372,8 @@ int main(int argc, char * argv[])
 	}
 	else
 	{
-		auto mmenu = CMainMenu::create();
+		bool playIntroVideo = !settings["session"]["headless"].Bool() && !vm.count("battle") && !vm.count("nointro") && settings["video"]["showIntro"].Bool();
+		auto mmenu = CMainMenu::create(playIntroVideo);
```
IvanSavenko, line 376: Perhaps call intro playback separately and remove that bool parameter? Should make code more clear:
```suggestion
		bool playIntroVideo = !settings["session"]["headless"].Bool() && !vm.count("battle") && !vm.count("nointro") && settings["video"]["showIntro"].Bool();
		auto mmenu = CMainMenu::create();
		if (playIntroVideo)
			mmenu->playIntro();
```

PR #4591, opened 2024-09-13
Placeholders fix in map validation

wb180:


wb180:
Also updated translation files

PR #4595, opened 2024-09-14
Bonus for spell: cast without skip the turn

### LINKED ISSUE #4088: Need a bonus for spell: cast without skip the turn
**Describe your proposal**
Similar as the setting: "canCastOnSelf" : true,

It should be an extra setting in spell format,
like "canCastWithoutSkip", if in true, the creature will not skip the turn.
default for false. 

For mod use, can make the label for modding.

Laserlicht:
Fixes #4088

### DIFF
--- a/lib/battle/CUnitState.h
+++ b/lib/battle/CUnitState.h
@@ -141,6 +141,7 @@ public:
 	bool drainedMana;
 	bool fear;
 	bool hadMorale;
+	SpellID usedSpell;
 	bool ghost;
 	bool ghostPending;
 	bool movedThisRound;
--- a/lib/battle/CUnitState.h
+++ b/lib/battle/CUnitState.h
@@ -141,7 +141,7 @@ public:
 	bool drainedMana;
 	bool fear;
 	bool hadMorale;
-	SpellID usedSpell;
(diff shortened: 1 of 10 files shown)


lib/battle/CUnitState.h
```diff
@@ -141,6 +141,7 @@ class DLL_LINKAGE CUnitState : public Unit
 	bool drainedMana;
 	bool fear;
 	bool hadMorale;
+	SpellID usedSpell;
```
IvanSavenko, line 144: Perhaps simple `castSpellThisTurn` will do? From what I see it is only used as a boolean.

Laserlicht: It's needed as SpellID in BattleFlowProcessor because functionality based on spell config.

Laserlicht: It's not. This was my first approach and not working. I've forgotten, why, but it's not the same (it's NONE at time of check).

IvanSavenko: Don't see how this could be possible to be honest...

Laserlicht: Yes. You're right. My mistake.

PR #4597, opened 2024-09-14
simplify usage of ttf font

Laserlicht:
New Settings entry in launcher (enables/disables mod):

grafik

### DIFF
--- a/launcher/modManager/cmodlistview_moc.cpp
+++ b/launcher/modManager/cmodlistview_moc.cpp
@@ -1080,6 +1080,12 @@ bool CModListView::isModEnabled(const QString & modName)
 	return mod.isEnabled();
 }
 
+bool CModListView::isModInstalled(const QString & modName)
+{
+	auto mod = modModel->getMod(modName);
+	return mod.isInstalled();
+}
+
 QString CModListView::getTranslationModName(const QString & language)
 {
 	for(const auto & modName : modModel->getModList())
--- a/launcher/modManager/cmodlistview_moc.h
+++ b/launcher/modManager/cmodlistview_moc.h
@@ -95,6 +95,9 @@ public:
 	/// returns true if mod is currently enabled
 	bool isModEnabled(const QString & modName);
 
+	/// returns true if mod is currently installed
+	bool isModInstalled(const QString & modName);
+
 public slots:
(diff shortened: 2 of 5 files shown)


IvanSavenko:
Perhaps call it "Use original game fonts"? And invert it obviously.
Term 'True Type font' is probably not exactly clear to most players.

However I think we need way more changes to ttf handling:
- integrate ttf fonts into main vcmi install instead of extras. This would be useful for multiplayer - if one player enters symbol that is not present in H3 font of another player.
- consider whether we should integrate existing (Liberation family) fonts from vcmi extras or replace it with another fonts, e.g. Noto which has better unicode coverage. Would be great to also integrate Chinese / Korean support, but that would increase install size too much IMO.
- allow scaling of all ttf fonts by specific factor, similar to resolution scale. Some players with small phones find current fonts to be too small. Implementation would be rather simple - multiply point size of loaded font by factor specified by player. Obviously, this option should only be used if ttf fonts are in use
- allow using multiple fonts at once. There are basically no good fonts that cover entire unicode and have decent typography for all of them. Fonts that are good for European languages are no good for Chinese fonts and vice versa.

Laserlicht:
> Perhaps call it "Use original game fonts"? And invert it obviously. Term 'True Type font' is probably not exactly clear to most players.

Not sure about inverting. I think most players would ask themself "Why shouldn't I use the original fonts?" They know nothing of the benefits.

Maybe "Use high resolution fonts"?

Should we finish this PR anyway? I think the TTF rework does involve a certain amount of effort. Not sure if we'll have that by 1.6?

IvanSavenko:
> Not sure about inverting. I think most players would ask themself "Why shouldn't I use the original fonts?" They know nothing of the benefits.
> 
> Maybe "Use high resolution fonts"?

Or "Use scalable fonts"? 

> Should we finish this PR anyway? I think the TTF rework does involve a certain amount of effort. Not sure if we'll have that by 1.6?

Actually considering doing this myself for 1.6. Wanted to do that even before, for 1.5/lobby, with xbrz this gets even more useful.

PR is fine, just feels a bit of a waste since some of these changes will have to be dropped / replaces, since for example, integrating fonts into base game means there is no mod to enable.

PR #4599, opened 2024-09-15
Fix error "hero access check" flooding console during quick-battle

MichalZr6:
This PR fixes this 

obraz

### DIFF
--- a/AI/BattleAI/AttackPossibility.cpp
+++ b/AI/BattleAI/AttackPossibility.cpp
@@ -44,24 +44,30 @@ void DamageCache::buildObstacleDamageCache(std::shared_ptr<HypotheticBattle> hb,
 		if(!triggerIsNegative)
 			continue;
 
-		const auto * hero = hb->battleGetFightingHero(spellObstacle->casterSide);
-		auto caster = spells::ObstacleCasterProxy(hb->getSidePlayer(spellObstacle->casterSide), hero, *spellObstacle);
+		std::unique_ptr<spells::BattleCast> cast = nullptr;
+		if(spellObstacle->obstacleType == SpellCreatedObstacle::EObstacleType::SPELL_CREATED)
+		{
+			const auto * hero = hb->battleGetFightingHero(spellObstacle->casterSide);
+			auto caster = spells::ObstacleCasterProxy(hb->getSidePlayer(spellObstacle->casterSide), hero, *spellObstacle);
+			cast = std::make_unique<spells::BattleCast>(spells::BattleCast(hb.get(), &caster, spells::Mode::PASSIVE, obst->getTrigger().toSpell()));
+		}
 
 		auto affectedHexes = obst->getAffectedTiles();
 		auto stacks = hb->battleGetUnitsIf([](const battle::Unit * u) -> bool {
 			return u->alive() && !u->isTurret() && u->getPosition().isValid();
 		});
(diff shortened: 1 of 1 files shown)


AI/BattleAI/AttackPossibility.cpp
```diff
@@ -44,24 +44,30 @@ void DamageCache::buildObstacleDamageCache(std::shared_ptr<HypotheticBattle> hb,
 		if(!triggerIsNegative)
 			continue;
 
-		const auto * hero = hb->battleGetFightingHero(spellObstacle->casterSide);
-		auto caster = spells::ObstacleCasterProxy(hb->getSidePlayer(spellObstacle->casterSide), hero, *spellObstacle);
+		std::unique_ptr<spells::BattleCast> cast = nullptr;
+		if(spellObstacle->obstacleType == SpellCreatedObstacle::EObstacleType::SPELL_CREATED)
```
nullkiller, line 48: Probably we can skip whole iteration in case we have no cast there

nullkiller: Maybe we can just replace battleGetFightingHero with getSideHero.

MichalZr6: `getSideHero` uses it anyway:
```
const CGHeroInstance * BattleProxy::getSideHero(BattleSide side) const
{
	return subject->battleGetFightingHero(side);
}
```

MichalZr6: > Probably we can skip whole iteration in case we have no cast there. Just continue. But this should act on moats as well. I do not know if SPELL_CREATED will work for moats.

Moats have separate type:
```
	enum EObstacleType : ui8
	{
		//ABSOLUTE needs an underscore because it's a Win
		USUAL, ABSOLUTE_OBSTACLE, SPELL_CREATED, MOAT
	};
```

AI/BattleAI/AttackPossibility.cpp
```diff
@@ -44,24 +44,30 @@ void DamageCache::buildObstacleDamageCache(std::shared_ptr<HypotheticBattle> hb,
...
 			auto updated = inner->getForUpdate(stack->unitId());
 
 			spells::Target target;
 			target.push_back(spells::Destination(updated.get()));
 
-			cast.castEval(inner->getServerCallback(), target);
+			if(cast)
+				cast->castEval(inner->getServerCallback(), target);
 
 			auto damageDealt = stack->getAvailableHealth() - updated->getAvailableHealth();
```
MichalZr6, line 72:
```suggestion
			int64_t damageDealt = 0;
			if(spellObstacle->obstacleType == SpellCreatedObstacle::EObstacleType::MOAT)
				damageDealt = spellObstacle->minimalDamage;
			else
				damageDealt = stack->getAvailableHealth() - updated->getAvailableHealth();
```

MichalZr6: Never mind. It was like that and changed last year oO

MichalZr6:
@IvanSavenko this still needs @nullkiller to look at. 
At this point it doesn't build damage cache for moats. My suggested workaround in the loop would do that but I think nullkiller should review as it is his code from about 3 weeks ago.

IvanSavenko:
Will take a look

PR #4602, opened 2024-09-15
fix error during ios build

smanolloff:
`onetbb/2021.10.0` adds a transitive dependency for `hwloc/2.9.1`

The problem is that `hwloc` does not support mobile OS.

```
checking which OS support to include... Unsupported! (aarch64-apple-ios)
configure: WARNING: ***********************************************************
configure: WARNING: *** hwloc does not support this system.
configure: WARNING: *** hwloc will *attempt* to build (but it may not work).
configure: WARNING: *** hwloc run-time results may be reduced to showing just one processor,
configure: WARNING: *** and binding will not be supported.
configure: WARNING: *** You have been warned.
configure: WARNING: *** Pausing to give you time to read this message...
configure: WARNING: ***********************************************************
checking which CPU support to include... unknown
```

...which eventually leads to:

```
ninja: error: '/Users/simo/.conan/data/hwloc/2.9.3/_/_/package/81c684be1f19b5b5aefe294dd826377be1235f07/lib/libhwloc.so', needed by 'appleclang_15.0_cxx17_64_release/libtbbbind_2_5.3.12.dylib', missing and no known rule to make it
onetbb/2021.12.0:
onetbb/2021.12.0: ERROR: Package '9fcc0a90f088a5fd6ea9bf93b3534713b5635296' build failed
```

Additionally, hwloc recipe apparently has dropped support for the `shared` option at some point, so it was removed from conanfile as well.

```
ERROR: onetbb/2021.7.0: option 'shared' doesn't exist
Possible options are ['tbbmalloc', 'tbbproxy', 'tbbbind', 'interprocedural_optimization']
```

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,7 +19,7 @@ class VCMI(ConanFile):
         "sdl_image/[~2.0.5]",
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
-        "onetbb/[^2021.3]",
+        "onetbb/[^2021.7 <2021.10]",  # 2021.10+ breaks mobile builds due to added hwloc dependency
         "xz_utils/[>=5.2.5]", # Required for innoextract
     ]
 
@@ -39,7 +39,6 @@ class VCMI(ConanFile):
 
         "boost/*:shared": True,
         "minizip/*:shared": True,
-        "onetbb/*:shared": True,
     }
 
     def configure(self):


smanolloff:
The plot thickens.

Onetbb versions 2021.3+ **all have dependency to hwloc now**. This explains my confusion earlier -- it came from some older cached recipes that I had downloaded. 

I believe the guys in the onetbb team had either screwed up, or intentionally made the poor decision to backport the `hwloc` dependency into all 2021.x versions. The information in [conan.io is wrong](https://conan.io/center/recipes/onetbb) - it says version 2021.8.0 has no hwloc dependency, for example, but it does:

```bash
# obtain download url for 2021.8.0
curl 'https://center.conan.io/v1/conans/onetbb/2021.8.0/_/_/download_urls'

# download recipe for 2021.8.0
curl 'https://center.conan.io/artifactory/api/conan/conan-center/v1/files/_/onetbb/2021.8.0/_/fa6894d676f63db5f9dad05298a8eb0e/export/conanfile.py'

# inspect the recipe and confirm hwloc dependency
# (same for all 2021.x versions)
```

I also confirmed the issue with another PC which had contained a cached for 2021.10.0 from March 6 -- it looks very different from the one being downloaded now.

The last `onetbb` recipe that has no hwloc dependency is `2020.3.3`. I believe we are forced to downgrade to it, as all conan builds for mobile will fail unless the user is lucky to have an old recipe already cached.

IvanSavenko:
From what I see, hwloc is only needed for tbbbind component. So I guess another option is to disable tbbbind on platforms that don't support it

smanolloff:
I come back with more information.

The recipe has changed and that apparently is a normal thing with conan center - besides _package_ versions, we have _recipe_ revisions. 

What we need here is to pin the recipe revision -- for a specific package version, it can be obtained like this:

```bash
$ export CONAN_REVISIONS_ENABLED=1
$ conan search onetbb/2021.9.0@ --revisions -r=conancenter
```

```bash
c646b08937225a183d8c59fa898a8617 (2024-04-19 16:21:43 UTC)
267e946d0523422b77828743a737a9c5 (2024-01-17 08:42:46 UTC)
059b7663b508dbacb2930fdeb781dde5 (2023-12-14 13:39:24 UTC)
a616cfe692d20ac24e9b26f199658dae (2023-12-01 20:37:10 UTC)
187aec39a7f7e65201473a3decb2d491 (2023-11-22 20:51:23 UTC)
e2784151a7f3987d66ca9cc7ab99089c (2023-11-02 14:34:48 UTC)
c3fa9033fc321cefb278fbc6cab6c916 (2023-10-27 07:11:34 UTC)
b0be0a5a91eb894cbdc9e59275ca4fc5 (2023-10-20 11:20:10 UTC)
81f921a75e782df14f4d9528facef986 (2023-09-19 23:01:19 UTC)
fdbd02e09e207e76a5cac2c2a80d5bcf (2023-08-31 07:40:19 UTC)
dbe36c31c136b486a4018bf0d52984af (2023-08-29 08:54:28 UTC)
4496abd4b5c9f222234e8ecb0007c359 (2023-08-28 21:12:04 UTC)
7fcd98c0308db7c5635dc83ad6fe9d2a (2023-08-13 08:58:20 UTC)
02440635872b7ac79f7e8c50df1a5359 (2023-08-09 11:43:46 UTC)
1105c7540a10e720abb08d6eaf4bd9b3 (2023-08-03 13:15:19 UTC)
938109fcfbb9aa4568cc85a971191883 (2023-05-08 18:45:11 UTC)
061f1da702c13a380100aeb32a871116 (2023-05-07 12:56:07 UTC)
```

So, for `2021.9.0`, the latest recipe revision without dependency is `1105c7540a10e720abb08d6eaf4bd9b3`.
For `2021.10.0`, there is no such revision (all have the dependency).

So, we can go no higher than 2021.9.0, and we must use the exact revision mentioned above. In conanfile, we can pin like so:

```python
"onetbb/2021.9.0@#1105c7540a10e720abb08d6eaf4bd9b3"
```

The only side effect I can think of is the prebuilt binaries from conan center -- do you thing that pinning the version like this would somehow imact them?

IvanSavenko:
From what I see, hwloc is optional dependency that is only used if tbbbind is enabled.
Perhaps disabling tbbbind for ios is better idea?

@kambala-decapitator your opinion?

kambala-decapitator:
we simply need to disable the following options of tbb: tbbmalloc, tbbproxy, tbbbind (hwloc is enabled only for the last one)

I see there were some changes related to hwloc: https://github.com/conan-io/conan-center-index/commits/master/recipes/onetbb/all/conanfile.py Maybe another change is needed for iOS.

also, revisions in Conan v1 are broken and shouldn't be used to pin recipe versions.

smanolloff:
> also, revisions in Conan v1 are broken and shouldn't be used to pin recipe versions.

yeah, I just came back here to post that something wasn't right after all, there was a boot error on iOS: `Library not loaded: @rpath/libtbb.12.dylib` and was about to suggest to abort mission and not use pinned revisions at all, and downgrade onetbb to `2020.3.3` (which works, but is old).
I thought I was going crazy already, but your statement certainly made me feel better - let's forget about recipe pinning. 

> we simply need to disable the following options of tbb: tbbmalloc, tbbproxy, tbbbind (hwloc is enabled only for the last one)

Disabling `tbbbind` is certainly seems the most sensible approach here, thanks.
Just tried it - compiles and runs on iOS fine. I will submit a follow-up PR with the change.

smanolloff:
https://github.com/vcmi/vcmi/pull/4630

PR #4606, opened 2024-09-16
Added ability to choose road types in editor RMG options

wb180:
Screenshot_20240916_173820

### DIFF
--- a/mapeditor/jsonutils.cpp
+++ b/mapeditor/jsonutils.cpp
@@ -17,7 +17,7 @@ static QVariantMap JsonToMap(const JsonMap & json)
 	QVariantMap map;
 	for(auto & entry : json)
 	{
-		map.insert(QString::fromUtf8(entry.first.c_str()), JsonUtils::toVariant(entry.second));
+		map.insert(QString::fromStdString(entry.first), JsonUtils::toVariant(entry.second));
 	}
 	return map;
 }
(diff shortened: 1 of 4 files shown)


mapeditor/jsonutils.cpp
```diff
@@ -108,10 +108,11 @@ JsonNode toJson(QVariant object)
 
 	if(object.canConvert<QVariantMap>())
 		ret.Struct() = VariantToMap(object.toMap());
-	else if(object.canConvert<QVariantList>())
-		ret.Vector() = VariantToList(object.toList());
 	else if(object.userType() == QMetaType::QString)
 		ret.String() = object.toString().toUtf8().data();
+	//QStirng can be converted to QVariantList, need to check first
+	else if(object.canConvert<QVariantList>())
+		ret.Vector() = VariantToList(object.toList());
```
IvanSavenko, line 115: I remember there was some problem with Qt5/Qt6 having different behavior. 
Are you testing with Qt5 or Qt6?

kambala-decapitator: you probably mean deprecated toSet / toList methods on QSet / QList, but here it's QVariant

IvanSavenko: No, I mean this commit:
https://github.com/vcmi/vcmi/commit/01af231084d5ed23dfad5571eb633624d56f0cf2

Yay for more copy-pasted code in editor, which now diverges even more for no reason...

In any case, I think it effectively the same change, albeit done a little bit differently - starting from Qt6, QVariant that contains QString can be converted into QList<char>.

I see 3 options:
- keep change from this PR as it
- copy-paste entire launcher/jsonutils into editor/jsonutils to bring them back in sync
- remove copy-pasted files. Add some sort of "vcmiqt" static library?

wb180: > No, I mean this commit: [01af231](https://github.com/vcmi/vcmi/commit/01af231084d5ed23dfad5571eb633624d56f0cf2)
> 
> Yay for more copy-pasted code in editor, which now diverges even more for no reason...
> 
> In any case, I think this is effectively the same change, albeit done a little bit differently - starting from Qt6, QVariant that contains QString can be converted into QList of chars
> 
> I see 3 options:
> 
>     * keep change from this PR as it
> 
>     * copy-paste entire launcher/jsonutils into editor/jsonutils to bring them back in sync

Updated editor/jsonutils with launcher/jsonutils

> 
>     * remove copy-pasted files. Add some sort of "vcmiqt" static library?

Think that requires separate PR.

PR #4608, opened 2024-09-16
Map editor fix event regressions

godric3:
Fixed two regressions from recent PRs:
 - map editor event from didn't supported 8th creature in town (resulted in crash of editor)
 - events didn't appeared because of broken saving of affected players (was recently changed to std::set)

There is still problem that creatures from town event are not added to town... works in 1.5.7

### DIFF
--- a/mapeditor/inspector/towneventdialog.cpp
+++ b/mapeditor/inspector/towneventdialog.cpp
@@ -13,6 +13,7 @@
 #include "towneventdialog.h"
 #include "ui_towneventdialog.h"
 #include "mapeditorroles.h"
+#include "../mapsettings/eventsettings.h"
 #include "../../lib/entities/building/CBuilding.h"
 #include "../../lib/entities/faction/CTownHandler.h"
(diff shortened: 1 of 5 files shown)


mapeditor/inspector/towneventdialog.cpp
```diff
@@ -213,12 +214,12 @@ void TownEventDialog::on_TownEventDialog_finished(int result)
...
 QVariant TownEventDialog::playersToVariant()
 {
-	int players = 0;
+	QVariantList players;
 	for (int i = 0; i < ui->playersAffected->count(); ++i)
 	{
 		auto * item = ui->playersAffected->item(i);
 		if (item->checkState() == Qt::Checked)
-			players |= 1 << i;
+			players.push_back(QString::fromStdString(PlayerColor(i).toString()));
```
kambala-decapitator, line 222: QString construction is copy-pasted multiple times, deserves a helper function

kambala-decapitator (approved):

PR #4615, opened 2024-09-17
Block ability to move on adventure map when holding shift

dydzio0614:
When creating path before holding shift key, and then clicking path end with shift it was possible to move hero, which is undesired. This PR fixes that behavior.


kambala-decapitator:
why was it merged into beta? Will we have 1.5.8?

dydzio0614:
Maybe, so merging to beta does not hurt - this fix is wanted by pvp players

IvanSavenko:
Ah, haven't noticed beta.
Not sure about 1.5.8. Most likely - not. But pvp players can use daily beta build with this fix

PR #4616, opened 2024-09-17
show creature level

### LINKED ISSUE #4600: QOL UI feature - Show tier of the unit
**Describe your proposal**
VCMI is cool for casuals like me and my friends because it shows some more information that new players don't know. I suggest to print stars to determine tier of the unit when right clicking on them on adventure map, so it's a little bit easier to determine the power and value of the creature.
I know there is threat level implemented, but it's not always accurate.

Mockup
tiers

Laserlicht:
Fixes #4600

Make sense for me, when playing many mod towns.

grafik

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -12,7 +12,7 @@
 	"vcmi.adventureMap.monsterThreat.levels.9"  : "Overpowering",
 	"vcmi.adventureMap.monsterThreat.levels.10" : "Deadly",
 	"vcmi.adventureMap.monsterThreat.levels.11" : "Impossible",
-	"vcmi.adventureMap.monsterLevel.title"     : "\n\nLevel: ",
+	"vcmi.adventureMap.monsterLevel"            : "\n\n%TOWN unit of level %LEVEL",
 
 	"vcmi.adventureMap.confirmRestartGame"               : "Are you sure you want to restart the game?",
 	"vcmi.adventureMap.noTownWithMarket"                 : "There are no available marketplaces!",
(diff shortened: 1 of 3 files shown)


dydzio0614:
did you check if this does not destroy layout when using vision spell?

Laserlicht:
> did you check if this does not destroy layout when using vision spell?

grafik

misiokles:
...and it is possible to add creatures' rank with stack experience module on?
For example
Few Firebirds
Level: 1 (Rank: 3)

krs0:
Another important information I miss...is it ranged or not. Or is it a spellcaster, or unit speed/initiative.

All this info could be covered if we could have a modal with all creature info, including abilities.

DjWarmonger:
Please rename it to "tier", which is more precise.

Laserlicht:
> Please rename it to "tier", which is more precise.

Can do it. But will be a little bit inconsistent to codebase.

IvanSavenko:
> Please rename it to "tier", which is more precise.

H3 always uses term 'level' here. For example:
> "{Head of Legion} This item, when equipped by a hero in a town, increases the growth of your 6th level units (in that town) by 1 per week.",

Perhaps instead we can use something like "Fortress unit of level 1"? Should be more clear that this is unit tier and will also tell its faction (might be useful info for new units)

kdmcser:
in Chinese, a better translation for Fortress unit of level 1 is 要塞(Fortress)1级(level)生物(unit)，so it is better to make translator can determine the positions of faction string and level number.

Laserlicht:
grafik

krs0:
Stronghold unit of Level 1...

Level 1 Stronghold unit, to me would sound more natural.

IvanSavenko:
> Level 1 Stronghold unit

Yeah, that probably sounds better.

Laserlicht:
> > Level 1 Stronghold unit
> 
> Yeah, that probably sounds better.

Done

PR #4627, opened 2024-09-19
avoid warnings if not installed chronicles

Laserlicht:


launcher/modManager/chroniclesextractor.cpp
```diff
@@ -106,6 +106,18 @@ void ChroniclesExtractor::createBaseMod() const
 		{ "author", "3DO" },
 		{ "version", "1.0" },
 		{ "contact", "vcmi.eu" },
+		{ "heroes", QJsonArray({"config/heroes/portraitsChronicles.json"}) },
```
IvanSavenko, line 109: Might still result in warnings if player installs only some of 8 chronicles

Laserlicht: It doesn't. Every chronicle campaign contains all tarnum portraits.

PR #4629, opened 2024-09-20
Update swedish.ts

XCOM-HUB:
@Fabrice-TIERCELIN

### DIFF
--- a/launcher/translation/swedish.ts
+++ b/launcher/translation/swedish.ts
@@ -11,7 +11,7 @@
     <message>
         <location filename="../aboutProject/aboutproject_moc.ui" line="29"/>
         <source>Have a question? Found a bug? Want to help? Join us!</source>
-        <translation type="unfinished">Har du något som är svår att förstå? Har du hittat ett fel eller ett annat problem? Behöver du hjälp? Förena oss i samma grupp!</translation>
+        <translation type="unfinished">Har du en fråga? Hittat en bugg? Vill du hjälpa till? Anslut dig till oss!</translation>
     </message>
     <message>
         <location filename="../aboutProject/aboutproject_moc.ui" line="36"/>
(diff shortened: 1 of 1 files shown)


launcher/translation/swedish.ts
```diff
@@ -1241,13 +1240,13 @@ Heroes® of Might and Magic® III HD stöds för närvarande inte!</translation>
     </message>
     <message>
         <location filename="../firstLaunch/firstlaunch_moc.cpp" line="330"/>
-        <source>File cannot opened</source>
+        <source>File cannot be opened</source>
```
Fabrice-TIERCELIN, line 1243: @XCOM-HUB, thank you for fixing the original mistakes but it won't work doing so. This text must perfectly match the source code so it will just not be translated (and it's also the wrong typo that will appear). It should be fixed by the developers in the source code and then be fixed in all the translation files. Have you done this at other locations?

Fabrice-TIERCELIN (changes requested):
@XCOM-HUB, tell us all the source captions you have fixed

launcher/translation/swedish.ts
XCOM-HUB: I have only done it here. Can you give me the URL for the other location so I can fix it?

XCOM-HUB:
> @XCOM-HUB, tell us all the source captions you have fixed

I also capitalized the P in row number 353: <source>please upgrade mod</source>

launcher/translation/swedish.ts
XCOM-HUB: @Fabrice-TIERCELIN 

I also made a slight change on row/line 353 (capitalized the first letter 'p' - <source>Please upgrade mod</source>

Fabrice-TIERCELIN:
OK, @XCOM-HUB, I have those locations in memory. I will fix it in a further PR.

@IvanSavenko, it's OK for me.

PR #4630, opened 2024-09-20
onetbb options in conanfile

smanolloff:
Follow-up PR from https://github.com/vcmi/vcmi/pull/4602 

Reverts upper-version limit for onetbb and adds appropriate options to prevent the `hwloc` dependency to prevent errors during ios builds.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,7 +19,7 @@ class VCMI(ConanFile):
         "sdl_image/[~2.0.5]",
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
-        "onetbb/[^2021.3]",
+        "onetbb/2021.10.0",
         "xz_utils/[>=5.2.5]", # Required for innoextract
     ]
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,7 +19,7 @@ class VCMI(ConanFile):
         "sdl_image/[~2.0.5]",
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
-        "onetbb/[^2021.12]",
+        "onetbb/[^2021.3]",
         "xz_utils/[>=5.2.5]", # Required for innoextract
     ]
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,7 +19,7 @@ class VCMI(ConanFile):
         "sdl_image/[~2.0.5]",
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
-        "onetbb/2021.10.0",
+        "onetbb/[>=2021.12.0]",
         "xz_utils/[>=5.2.5]", # Required for innoextract
     ]
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,7 +19,7 @@ class VCMI(ConanFile):
         "sdl_image/[~2.0.5]",
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
-        "onetbb/[^2021.3]",
+        "onetbb/[^2021.12]",
         "xz_utils/[>=5.2.5]", # Required for innoextract
     ]
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,7 +19,7 @@ class VCMI(ConanFile):
         "sdl_image/[~2.0.5]",
         "sdl_mixer/[~2.0.4]",
         "sdl_ttf/[~2.0.18]",
-        "onetbb/[^2021.7 <2021.10]",  # 2021.10+ breaks mobile builds due to added hwloc dependency
+        "onetbb/[^2021.3]",
         "xz_utils/[>=5.2.5]", # Required for innoextract
     ]
 
@@ -145,6 +145,10 @@ class VCMI(ConanFile):
         #    self.options["ffmpeg"].avfilter = True
         #    self.options["ffmpeg"].with_sdl = True
         #    self.options["ffmpeg"].enable_filters = "aresample,scale"
+
+        self.options["onetbb"].tbbmalloc = False
+        self.options["onetbb"].tbbproxy = False
+        self.options["onetbb"].tbbbind = False
 
         self.options["sdl"].sdl2main = self.settings.os != "iOS"
         self.options["sdl"].vulkan = False


conanfile.py
```diff
@@ -232,6 +232,11 @@ def _disableQtOptions(disableFlag, options):
         if self.settings.os == "Android":
             self.options["sqlite3"].build_executable = False
 
+        # Needed for mobile builds (see https://github.com/vcmi/vcmi/pull/4602)
+        self.options["onetbb"].tbbmalloc = False
+        self.options["onetbb"].tbbproxy = False
+        self.options["onetbb"].tbbbind = False
```
kambala-decapitator, line 238: please place them before `self.options["sdl"]`

also the comment isn't needed as we don't use those parts anyway

kambala-decapitator:
use something like "if tbb version >= that_new_one then set tbbproxy=false"

kambala-decapitator:
tbbind is available since version 2021.11.0

smanolloff:
I don't think such a check is possible. During .configure we only have the specs, which is `>=2012.3`. 

The only way I see from here is to use `>=2012.12` which has this option.  The problem is that CI builds fail because no pre-built package could be found for that version.

kambala-decapitator:
then just limit max version for now `< 2021.11.0`, we'll adjust it later

smanolloff:
You don't understand -- the problem is that regardless of the version, **`hwloc` is as a dependency** because the recipes have been updated, and ios cannot be built with that dependency.

* Version 2021.12.0 has `hwloc` dependency but allows to set the `onetbb` option to False, which removes the dependency. This fixes the ios built, but breaks CI builds with "missing pre-built package".

* All other 2021.* versions also have `hwloc` dependency, but do not allow to set `onetbb` option to False.
This means CI builds will still work (they use pre-built versions from long ago without that dependency), but breaks dev builds for ios with the error mentioned in https://github.com/vcmi/vcmi/pull/4602.

I believe there are only two possible ways to go from here:

1. Downgrade to `onetbb/2020.3.3` which has no `hwloc` dependency and no `onetbb` option. Its recipes are no longer updated and it's kind of "stable". It an OK option, if it's not too outdated for you.
2. Upgrade to `onetbb/[>=2021.12] and fix CI builts by providing a pre-built package, if this is possible (I don't know how pre-built packages are shipped to conan center).

IvanSavenko:
But from what I see, that dependency is only active if tbbbind is enabled?
https://github.com/conan-io/conan-center-index/blob/master/recipes/onetbb/all/conanfile.py#L88

kambala-decapitator:
> But from what I see, that dependency is only active if tbbbind is enabled? https://github.com/conan-io/conan-center-index/blob/master/recipes/onetbb/all/conanfile.py#L88

exactly. Not sure why you're not trying to read the tbb recipe file @smanolloff. Your issue is that the latest tbb version is fetched, but you can easily fix that by providing a cap on max version like I wrote above; alternatively, pass `-o 'onetbb/*:tbbbind=False' on the command line to be able to build the latest tbb version.

> and then fix CI builts by providing a pre-built package, if this is possible (I don't know how pre-built packages are shipped to conan center).

FYI Conan center doesn't provide binaries for iOS / android at all. We build everything for all platforms ourselves.

smanolloff:
It's not about reading the recipe. I've read it many times, debugged it with python debugger. Lower version - tried that. IOS builds do not work locally. It just doesn't behave as you would expect. I challenge you to try your own simple/easy solutions yourself 😉

kambala-decapitator:
> Lower version - tried that. IOS builds do not work locally. It just doesn't behave as you would expect

please provide details. So far you only reported that the latest tbb version (and recipe) can't be used in VCMI out of the box.

smanolloff:
### Case: `onetbb/2021.12.0` (default options)

Output of `grep onetbb conanfile.py`
```
        "onetbb/2021.12.0",
```

* Result of `conan install ...` (CI): error "missing package" (see [this build](https://github.com/vcmi/vcmi/actions/runs/10963490826/job/30445076718?pr=4630))
* Result of `conan install ...` (local):
    * for `profile:build=default profile:host=default`: OK (hwloc builds successfully for mac)
    * for `profile:build=ios profile:host=CI/conan/ios-arm64`: error "hwloc does not support his system" (originally reported in https://github.com/vcmi/vcmi/pull/4602)

### Case: `onetbb/2021.12.0` (option `tbbbind` set to `False`)

Output of `grep onetbb conanfile.py`
```
        "onetbb/2021.12.0",
        self.options["onetbb"].tbbbind = False
```

* Result of `conan install ...` (CI): error "missing package"
* Result of `conan install ...` (local):
    * for `profile:build=default profile:host=default`:  OK (no hwloc dependency)
    * for `profile:build=ios profile:host=CI/conan/ios-arm64`: OK (no hwloc dependency)

### Case: `onetbb/[<2021.10.0]` (default options)

Output of `grep onetbb conanfile.py`
```
        "onetbb/[<2021.10.0]",
```

* Result of `conan install ...` (CI): OK "onetbb/2021.7.0: Already installed!" (see [this build](https://github.com/vcmi/vcmi/actions/runs/10975254339/job/30474696967)) - CI uses pre-built package.
* Result of `conan install ...` (local):
    * for `profile:build=default profile:host=default`: OK (no hwloc dependency)
    * for `profile:build=ios profile:host=CI/conan/ios-arm64`: error "hwloc does not support his system" (originally reported in https://github.com/vcmi/vcmi/pull/4602)

### Case: `onetbb/[<2021.10.0]` (option `tbbbind` set to `False`)

Output of `grep onetbb conanfile.py`:
```bash
        "onetbb/[<2021.10.0]",
        self.options["onetbb"].tbbbind = False
```

* Result of `conan install ...` (CI): error "option 'tbbbind' doesn't exist" (see [this build](https://github.com/vcmi/vcmi/actions/runs/10962267014/job/30441104900))
* Result of `conan install ...` (local):
    * for `profile:build=default profile:host=default`: error "option 'tbbbind' doesn't exist"
    * for `profile:build=ios profile:host=CI/conan/ios-arm64`: OK (no hwloc dependency)

# Conclusion

* Using `onetbb/>2021.11.0` is a possible solution, but requires pre-building the package for CI.
* Using `onetbb/<2021.11.0` is not an option
* Using `onetbb/<2021` is another possible oslution (2020.* work ok), but would also require pre-building for CI and makes no sense given we can use the newer version for the same effort.

kambala-decapitator:
> self.options["onetbb"].tbbbind = False

not sure why you're changing the conanfile.py instead of providing this option from command line like I wrote above: -o 'onetbb/*:tbbbind=False' - this would allow you to build the latest version for iOS and not touch VCMI's file.

and, like I said, tbb recipe must be fixed to skip hwloc for all Apple non-macOS platforms because hwloc can't be built for them.

kambala-decapitator:
> and, like I said, tbb recipe must be fixed to skip hwloc for all Apple non-macOS platforms because hwloc can't be built for them.

opened https://github.com/conan-io/conan-center-index/pull/25361

smanolloff:
Alright, my understanding was that dealing with the issue at hand is a top priority -- that's why I suggested a workaround that does not involve changes into a third-party repo we don't have direct control of. If you prefer to wait for the conan recipe fix instead, then I am closing this PR.
Thank you for your help.

kambala-decapitator:
> does not involve changes into a third-party repo we don't have direct control of

it's not a third-party repo - it's the primary source of all recipes, upstream. Recipes that CI uses were also taken from it, but earlier - since then some recipes evolved, but incompatibilities may happen like you have encountered here. From time to time we update library versions in the pre-built packages as well as the recipes used.

you could also have taken the onetbb recipe that CI uses - while it doesn't contain the latest tbb versions, it would have worked for you out of the box.

PR #4640, opened 2024-09-21
Add all artifacts/spells/skills to quest widget lists and disable banned

godric3:
Previous code added at most as many items as were in allowed lists, which caused to skip last items if some were banned.

### DIFF
--- a/mapeditor/inspector/questwidget.cpp
+++ b/mapeditor/inspector/questwidget.cpp
@@ -48,11 +48,10 @@ QuestWidget::QuestWidget(MapController & _controller, CQuest & _sh, QWidget *par
 	}
 	
 	//fill artifacts
-	for(auto const & artifactPtr : VLC->arth->objects)
+	for(const auto artifactPtr : VLC->arth->objects)
 	{
 		auto artifactIndex = artifactPtr->getIndex();
 		auto * item = new QListWidgetItem(QString::fromStdString(artifactPtr->getNameTranslated()));
-		item->setData(Qt::UserRole, QVariant::fromValue(artifactIndex));
 		item->setFlags(item->flags() | Qt::ItemIsUserCheckable);
 		item->setCheckState(Qt::Unchecked);
 		if(controller.map()->allowedArtifact.count(artifactIndex) == 0)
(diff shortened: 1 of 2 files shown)


mapeditor/inspector/questwidget.cpp
```diff
@@ -48,48 +48,51 @@ QuestWidget::QuestWidget(MapController & _controller, CQuest & _sh, QWidget *par
 	}
 	
 	//fill artifacts
-	for(int i = 0; i < controller.map()->allowedArtifact.size(); ++i)
+	for(auto const & artifactPtr : VLC->arth->objects)
```
kambala-decapitator, line 51: we use `const auto` style

reference to a pointer isn't needed here

mapeditor/inspector/questwidget.cpp
```diff
@@ -48,48 +48,51 @@ QuestWidget::QuestWidget(MapController & _controller, CQuest & _sh, QWidget *par
...
 	{
-		auto * item = new QListWidgetItem(QString::fromStdString(VLC->artifacts()->getByIndex(i)->getNameTranslated()));
-		item->setData(Qt::UserRole, QVariant::fromValue(i));
+		auto artifactIndex = artifactPtr->getIndex();
+		auto * item = new QListWidgetItem(QString::fromStdString(artifactPtr->getNameTranslated()));
+		item->setData(Qt::UserRole, QVariant::fromValue(artifactIndex));
```
kambala-decapitator, line 55: please create enum for the custom role

mapeditor/inspector/questwidget.cpp
```diff
@@ -48,48 +48,51 @@ QuestWidget::QuestWidget(MapController & _controller, CQuest & _sh, QWidget *par
...
+	for(auto const & skillPtr : VLC->skillh->objects)
 	{
-		auto * item = new QTableWidgetItem(QString::fromStdString(VLC->skills()->getByIndex(i)->getNameTranslated()));
-		item->setData(Qt::UserRole, QVariant::fromValue(i));
+		auto skillIndex = skillPtr->getIndex();
+		auto * item = new QTableWidgetItem(QString::fromStdString(skillPtr->getNameTranslated()));
+		item->setData(Qt::UserRole, QVariant::fromValue(skillIndex));
 		
 		auto * widget = new QComboBox;
 		for(auto & s : NSecondarySkill::levels)
```
kambala-decapitator, line 85: please also change here to `const auto &`

mapeditor/inspector/questwidget.cpp
godric3: hmm... data from those items isn't event retrieved, so I will remove it

godric3:
also added missing serialization of spells in Limiter

PR #4664, opened 2024-09-24
Integrate true type fonts into VCMI

IvanSavenko:
VCMI now has true type fonts as part of its base install. Unlike similar submod in vcmi-extras, vcmi uses Noto font family (instead of Liberation) - due to slightly better unicode coverage and better match for h3 fonts, especially at low UI scaling mode.

Changes: 
- Selection of fonts to use is possible in Launcher - automatic, prefer original, or prefer true type fonts, without need to install vcmi extras.
- Added option to scale fonts (only for true type fonts) - this was requested several times for mobile platforms with small screens. If enabled, game will use true type fonts and they will be scaled by specified percentage. Due to space limitation this may lead to overflowing text in some UI elements, so option is marked as 'experimental'.
- Game will now load true type fonts from all mods. So if player uses our integrated fonts and installs mod that provides additional fonts (e.g. Chinese localization), game will use both our and fonts from mod to extend unicode coverage.
- By default, game will use original fonts as before for unscaled mode,  but for upscaled / xbrz mode game will switch to scalable TTF fonts due to better visuals. It is possible to override this by opting to always use original fonts in Launcher

This fixes following issues:
- If player selects language that is not supported by his H3 fonts (e.g. cyrillic languages in English H3), VCMI will now be able to display fonts instead of displaying empty space 
- In multiplayer, player will now be able to see chat messages from another players, even if they contain symbols not present in their H3 font.

TODO - remove true type fonts submod from vcmi-extras


dydzio0614:
I though I replaced Noto with WenQuanYi Micro Hei font back in the day, but if not - wouldn't that font be better?

IvanSavenko:
Not sure. My primary goal was to find font that is very similar to H3 looks & feels, with better coverage as nice bonus.

WenQuanYi  pros: 
- coverage of Chinese (and possibly Korean?)
Cons: 
- very poor quality of European languages
- likely worse coverage of European languages
- larger size (to be expected considering Chinese presence)
- very different from H3 fonts

dydzio0614:
I tested WenQuanYi with european fonts and following text:

Flygande bäckasiner söka strax hwila på mjuka tuvor. Любя, съешь щипцы, — вздохнёт мэр, — кайф жгуч. Nechť již hříšné saxofony ďáblů rozzvučí síň úděsnými tóny waltzu, tanga a quickstepu. Ξεσκεπάζω την ψυχοφθόρα βδελυγμία. Įlinkusi fechtuotojo špaga blykčiodama gręžė apvalų arbūzą. Nne, nna, wepụ he'l'ụjọ dum n'ime ọzụzụ ụmụ, vufesi obi nye Chukwu, ṅụrịanụ, gbakọọnụ kpaa, kwee ya ka o guzoshie ike. Ìwò̩fà ń yò̩ séji tó gbojúmó̩, ó hàn pákànpò̩ gan-an nis̩é̩ rè̩ bó dò̩la.

only few characters in last 2 sentences were shown wrong

IvanSavenko:
Chinese & Korean translations already includes true type fonts for their language.

So the only case that won't be covered is Chinese text in chat for non-Chinese players. Which right now is not automatic anyway - since ttf mod in vcmi-extras is disabled by default.

In theory we could include Chinese fonts in vcmi, but that means larger install size for everyone and not sure whether that font also covers Korean.
Alternatively, we can replace current submod in extras with Chinese fonts submod, that will be active by default.

Laserlicht:
Tested on smartphone. Looks quite good with Noto.

Is it still possible to override Noto in a mod? (if I prefer other fonts; or for our troll mod `Erathian font` 😅)

Do we have a way to fix y offset for small fonts:
grafik

Our coordinates are optimized for bitmap font. Can we somewhere define some y offsets per font (only ttf) to correct this issue?

IvanSavenko:
> Is it still possible to override Noto in a mod? (if I prefer other fonts; or for our troll mod Erathian font 😅)

Yeah, possible - enabling true type fonts from vcmi-extras replaces all Noto fonts from vcmi

> Do we have a way to fix y offset for small fonts:

Found strange logic for text positioning in there. Fixed.

Laserlicht:
> > Is it still possible to override Noto in a mod? (if I prefer other fonts; or for our troll mod Erathian font 😅)
> 
> Yeah, possible - enabling true type fonts from vcmi-extras replaces all Noto fonts from vcmi
> 
> > Do we have a way to fix y offset for small fonts:
> 
> Found strange logic for text positioning in there. Fixed.

Yes, that strange logic was because of the polish version: See here: https://github.com/vcmi/vcmi/pull/3467/commits/d309a0002554e94e2e431a2c89c69815e52ba246 and #3533

The polish font have a different height (I think one pixel). With rounding and the strange logic it was working for both (polish/english). But yeah - now it's a problem for ttf. 😅
Maybe we need a way to fix this for polish font.

IvanSavenko:
Checked polish version from gog - don't see anything wrong with that stack amount box

PR #4668, opened 2024-09-24
Chronicles improvements

### LINKED ISSUE #4663: Add Scenario name in the Campaign of Heroes Chronicles
**Describe your proposal**
@Laserlicht Add Scenario name in the Campaign of Heroes Chronicles
**Additional context**

You can see that the Scenario name with the red line is in this picture.

Laserlicht:
- fixes #4663
- move chronicles campaigns to subfolder (to group them if there are more custom campaigns) ----- @lishaoxia1985 : Breaking change; you have to recreate or move files to subdir in your mod
- playback main menu theme, when using back button instead of keep playing campaign music


lishaoxia1985:
Please give the instruction about how to edit the created mod adjust to the lastest commit about Heroes Chronicles.
Maybe we can upload the Heroes Chronicles as a mod.
Of course, it is maybe illegal.

Laserlicht:
If this PR will merged:
Just open your Chronicles Mod and do following for each of the 8 submods:
- open the maps folder
- create folder "Chronicles"
- and move the h3c file from maps folder there

> Maybe we can upload the Heroes Chronicles as a mod.

This approach with installing with launcher is exactly to avoid this. That the players can use own (legally acquired) copies from GOG. If we distribute chronicles as mod problems with UBI could be possible (they still sell it; no abandonware) and a risk for the project itself. Yes, UBI is not Big N, but they still want to earn money.

lib/campaign/CampaignState.h
```diff
@@ -42,12 +42,19 @@ class DLL_LINKAGE CampaignRegions
 		std::string infix;
 		int xpos;
 		int ypos;
+		std::optional<int> xLabelpos;
+		std::optional<int> yLabelpos;
```
IvanSavenko, line 46: Not sure if Point has serialize() method. If not - then easy to add anyway.
```suggestion
		Point pos;
		std::optional<Point> labelPos;
```

PR #4680, opened 2024-09-27
Basic spell research

Laserlicht:
Basic "stateless" spell research. Algorithm can be improved later...

Enable with mod:
```
{
	"name" : "Spell research",
	"description" : "Spell research",
	"version" : "1.0",
	"author" : "",
	"modType" : "Expansion",
	"compatibility":
	{
		"min" : "1.6.0"
	},
	"changelog":
	{
		"1.0": ["Initial version"]
	},
	"settings" : {
		"towns" : {
			"spellResearch" : true
		}
	}
}
```


server/CGameHandler.cpp
```diff
@@ -2233,6 +2242,53 @@ bool CGameHandler::razeStructure (ObjectInstanceID tid, BuildingID bid)
...
+	if(!getPlayerState(t->getOwner())->resources.canAfford(cost) && complain("Spell replacement cannot be afforded!"))
+		return false;
+
+	giveResources(t->getOwner(), -cost);
+
+	auto spells = t->spells.at(level);
+
+	std::swap(spells.at(t->spellsAtLevel(level, false)), spells.at(vstd::find_pos(spells, spellAtSlot)));
+	auto it = spells.begin() + t->spellsAtLevel(level, false);
+	std::rotate(it, it + 1, spells.end()); // move to end
```
IvanSavenko, line 2280: This means that once you roll through all spells you can learn *all* of them.
AFAIK hota spell research allows you to only learn spell once you've "finished" spell research.
So if you're researching lvl 5 spell your hero can only learn spell that you had originally and final spell that you've confirmed.

IMO if we want simple version of spell research something like this is better option:
> when player requests a spell reroll, game will propose replacement spell that player may accept or deny
> - if player accepts, spell will be added to list of obligatory spells
> - if player denies, spell will be added to list of banned spell (yes, permanently)
> - if player denies, he can reroll spell again (unless there is daily limit)
> - if all spells are either obligatory or banned, spell reroll is not possible

server/CGameHandler.cpp
```diff
@@ -2233,6 +2242,53 @@ bool CGameHandler::razeStructure (ObjectInstanceID tid, BuildingID bid)
...
+	int daysSinceLastResearch = gs->getDate(Date::DAY) - t->lastSpellResearchDay;
+	if(!daysSinceLastResearch && complain("Already researched today!"))
+		return false;
+
+	TResources cost;
+	cost[EGameResID::GOLD] = 1000;
+	cost[EGameResID::MERCURY] = (level + 1) * 2;
+	cost[EGameResID::SULFUR] = (level + 1) * 2;
+	cost[EGameResID::CRYSTAL] = (level + 1) * 2;
+	cost[EGameResID::GEMS] = (level + 1) * 2;
```
IvanSavenko, line 2269: Should be in GameSettings. For example as two values: base cost and cost / level. 
I think ResourceSet already has constructor from json

lib/mapObjects/CGTownInstance.h
```diff
@@ -73,6 +73,7 @@ class DLL_LINKAGE CGTownInstance : public CGDwelling, public IShipyard, public I
 	std::vector<std::vector<SpellID> > spells; //spells[level] -> vector of spells, first will be available in guild
 	std::vector<CCastleEvent> events;
 	std::pair<si32, si32> bonusValue;//var to store town bonuses (rampart = resources from mystic pond, factory = save debts);
+	int lastSpellResearchDay;
```
IvanSavenko, line 76: Would also be nice to have "spell research allowed in town" boolean property - part of hota map format. Check `CMapLoaderH3M::readTown`

client/windows/CCastleInterface.h
```diff
@@ -393,8 +394,11 @@ class CMageGuildScreen : public CStatusbarWindow
 
 	std::shared_ptr<CMinorResDataBar> resdatabar;
 
+	ObjectInstanceID townId;
+
 public:
 	CMageGuildScreen(CCastleInterface * owner, const ImagePath & image);
+	void update();
```
IvanSavenko, line 401: 1) More clear name as to what changed
2) Method needs to check whether it is our town changed or some other one, e.g. another player in simturns
```suggestion
	void townSpellsChanged(ObjectInstanceID townID);
```

server/CGameHandler.cpp
dydzio0614: seems to be better, but i see abuse potential because if it works the way that I think then all spell slots of same tier share banned list then player may use first spell slots as "always accept if spell is at least mediocre" kind of thing and only resort to use last spell slot for getting the spell he wants

client/windows/CCastleInterface.h
Laserlicht: Should be irrelevant. This is for opened window. And all opened mage guild windows will be updated, when there was an spell change.

IvanSavenko: No. 
If player A updates spell in his town X, player B does not needs an update if he is looking at his mages guild in town Y.

server/CGameHandler.cpp
Laserlicht: changed the algorithm a little bit:
- it's now possible to skip spell (can do a new decision next day)
- each accepted research the costs increases exponential. Adds some strategic decisions to make. Avoids also to learn all (at least without unlimited resources).

client/windows/CCastleInterface.h
Laserlicht: Changed

client/widgets/CComponent.h
```diff
@@ -52,6 +52,7 @@ class CComponent : public virtual CIntObject
 	std::string customSubtitle;
 	ESize size; //component size.
 	EFonts font; //Font size of label
+	bool newLine; //Line break after component
```
IvanSavenko, line 55: Why was this necessary?

IvanSavenko:
Code-wise OK, mechanically-wise - I would rather have version I described here in the comment, it should be more balanced, but don't care much about this feature so OK I guess.

Would like confirmation from pvp players that they are also OK with this approach before merging.

client/widgets/CComponent.h
Laserlicht: To make the spell (and only this one) in first line and ressources in following lines.

lib/mapObjects/CGTownInstance.h
```diff
@@ -73,6 +73,9 @@ class DLL_LINKAGE CGTownInstance : public CGDwelling, public IShipyard, public I
 	std::vector<std::vector<SpellID> > spells; //spells[level] -> vector of spells, first will be available in guild
 	std::vector<CCastleEvent> events;
 	std::pair<si32, si32> bonusValue;//var to store town bonuses (rampart = resources from mystic pond, factory = save debts);
+	std::map<int, int> spellResearchActionsPerDay;
+	int spellResearchCounter;
```
IvanSavenko, line 77: Why do you need `spellResearchActionsPerDay`?
If you need "how many times spell research was done today" then you can use simple counter and reset it in `NewTurn::applyGs`, similar to built buildings counter

Mods/vcmi/config/vcmi/english.json
```diff
@@ -59,6 +59,13 @@
 
 	"vcmi.spellBook.search" : "search...",
 
+	"vcmi.spellResearch.canNotAfford" : "You can't afford to replace {%SPELL1} with {%SPELL2}. But you can still discard this spell and continue spell research.",
```
IvanSavenko, line 62: I though that discarding / continuing spell research also costs resources?

server/CGameHandler.cpp
```diff
@@ -2233,6 +2243,59 @@ bool CGameHandler::razeStructure (ObjectInstanceID tid, BuildingID bid)
 	return true;
 }
 
+bool CGameHandler::spellResearch(ObjectInstanceID tid, SpellID spellAtSlot, bool accepted)
+{
+	CGTownInstance *t = gs->getTown(tid);
+
+	if(!(getSettings().getBoolean(EGameSettings::TOWNS_SPELL_RESEARCH) && t->spellResearchAllowed) && complain("Spell research not allowed!"))
+		return false;
```
IvanSavenko, line 2251:
```suggestion
	if(!getSettings().getBoolean(EGameSettings::TOWNS_SPELL_RESEARCH) && complain("Spell research not allowed!"))
		return false;

	if (!t->spellResearchAllowed && complain("Spell research not allowed in this town!"))
		return false;
```

lib/mapObjects/CGTownInstance.h
Laserlicht: Ah thanks. This was the missing puzzle piece. :)

Mods/vcmi/config/vcmi/english.json
Laserlicht: Currently not.

IvanSavenko:
Attempted to make better / more clear icons for this menu.
Can you try using them instead? Feel free to also ask modders for better version. Can't say that I am entirely happy with my version, but still should be more clear than UI from this PR.

You can use our pseudo-button `settingsWindow\button80` in place of .def (will result in empty button with no images / text)
And these icons: [researchIcons.zip](https://github.com/user-attachments/files/17268978/researchIcons.zip) - you can add them to button via setOverlay() method of button (may not be the exact name)
I've also added 3rd icon for cancel / close dialog - strictly speaking not necessary, but it prevents mixing slightly different art styles.

pppbb:
@Laserlicht please rebase - i get errors applying patch.

Laserlicht:
> Attempted to make better / more clear icons for this menu. Can you try using them instead? Feel free to also ask modders for better version.

Thanks!

grafik

PR #4690, opened 2024-09-28
Tweaks for scalable / true type fonts

### LINKED ISSUE #4681: incorrect position of all labels
**Describe the bug**
It seems that with ttf introduction all bitmap fonts are re positioned a little bit.

In almost every field the text is now vertically decentered (in 1.6 they were perfectly centered).

It seems they moved 1px down.

**Did it work earlier?**
Before #4664 it was ok

**Screenshots**
Before:
grafik
grafik

After:
grafik
grafik

IvanSavenko:
- Reduced point size of some fonts to better match H3 (including localized versions). I originally use upper case symbols as reference to match, but upper/lower case size ratio is different in H3 & Noto, so some fonts ended up a little too big.
- Likely fixed #4681. Can confirm that cases described in there are now fine, but haven't made detailed comparison vs before or vs H3
- Tweaked autoselection of ttf/h3 fonts. Now, if player uses languages without H3 fonts (e.g. Chinese) game will always prefer ttf fonts - to avoid excessive mixing of ttf/h3 fonts. Specifically - if font type is set to default (autoselection), game will now prefer ttf fonts if encoding of H3 fonts is different from encoding used by language selected by player.

### DIFF
--- a/client/lobby/OptionsTab.cpp
+++ b/client/lobby/OptionsTab.cpp
@@ -1037,11 +1037,11 @@ OptionsTab::PlayerOptionsEntry::PlayerOptionsEntry(const PlayerSettings & S, con
 	}
 	const auto & font = GH.renderHandler().loadFont(FONT_SMALL);
 
-	labelWhoCanPlay = std::make_shared<CMultiLineLabel>(Rect(6, 23, 45, font->getLineHeight()*2), EFonts::FONT_TINY, ETextAlignment::CENTER, Colors::WHITE, CGI->generaltexth->arraytxt[206 + whoCanPlay]);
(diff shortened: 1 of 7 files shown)


Laserlicht:
Looks much better!

**Whats the reason for this different font?**
This PR (no font scaling):
grafik
grafik

V1.5.x
grafik
grafik

IvanSavenko:
It uses tiny font both in 1.5 and 1.6.
You're using same game data in both cases, right?

Do you have any active mod that adds H3 fonts? Could be the cause. In 1.5 vcmi would attempt to load multiple h3 fonts to maximize unicode coverage. Now it only loads one and uses ttf to extend coverage.

So if you have localization mod that provides h3 font vcmi will always use this h3 font (and only this font), while in 1.5 vcmi would also load font from h3 data.

Laserlicht:
Yes the german translation has fnt files. But I think these are the same than english (only needed when coming from another language).

Even if I delete them it doesn't change.

Otherwise only vcmi-extras are installed.

PR #4694, opened 2024-09-29
cast only on self

Laserlicht:
Requested by @by003

Untested yet. Can you test it?


Laserlicht:
@IvanSavenko From logic side okay?

@by003 wants to test it as part of develop...

PR #4696, opened 2024-09-30
color shift parameter to config

Laserlicht:
possibility to fix this color issue
grafik

in https://github.com/vcmi-mods/small-era-mods/tree/vcmi-1.4/small-era-mods/Mods/Darkened%20players


Laserlicht:
I'm currently not sure what's the best way to fix. Basically it would be also possible to replace generated assets by mod.

But I think adding these parameters to config isn't a bad idea?

IvanSavenko:
> But I think adding these parameters to config isn't a bad idea?

It has one downside - multiplayer. Mod that modifies game settings would be (or at least should be) considered as gameplay altering mod and is required to load saves & to join multiplayer games that use this mod.

However same can be said for mods that change portraits / animations / etc. So I'd say that this is vcmi bug that we need to fix.

config/gameConfig.json
```diff
@@ -555,6 +555,22 @@
 					"valueType" : "BASE_NUMBER"
 				}
 			}
+		},
+
+		"general" :
```
IvanSavenko, line 560: Let's call new section 'interface'. Might add more stuff to it later. But probably only after figuring out multiplayer / save compatibility.

```suggestion
		"interface" :
```

Issue #3426, opened 2024-01-03
Improve UI of a heroes' backpack


sushi-shi:
**Describe your proposal**
Current backpack implementation does not support sorting artifacts in any manner, which makes it hard to understand which artifacts are available and in what quantity.

The idea mostly stems from Darkest Dungeon game, where UI provides a few buttons for sorting:
![image](https://github.com/vcmi/vcmi/assets/47691267/ccfa7001-f6f1-48be-9fc3-c7328aa6b938)
 
As I see it, four buttons can be added to a backpack (from left to right):
* unequipping all artifacts
* sorting by slots the artifacts can be placed in
* sorting by artifact's class (relic, minor, etc)
* sorting by artifact's name

If artifacts are sorted by slots and two artifacts can be placed in the same slot, they will be subsorted by class, which in turn will be subsorted by name. (fallthrough switch statements, yay!)

This is how I see the interface:
![image](https://github.com/vcmi/vcmi/assets/47691267/bc4fe438-5bac-47ad-ada2-3cca132a10a5)

**Additional context**
After implementing this feature I want to build on it, to be able to use backpack in Artifact Merchants to improve visibility so that people selling artifacts could see how many of each artifact they have, instead of using a small 5-slot window:
![image](https://github.com/vcmi/vcmi/assets/47691267/ef39b996-e0c6-47a2-90cd-d8f47de0f464)

The idea is to click on a resource in which artifacts will be sold, open the backpack, sort, and then use LMB for quickly selling non-needed artifacts.

I am willing to implement this feature.

dydzio0614:
Sounds good, I may have more comments when I see actual UI changes

IvanSavenko:
@SoundSSGood you might be interested in this.

> If artifacts are sorted by slots and two artifacts can be placed in the same slot, they will be subsorted by class, which in turn will be subsorted by name. (fallthrough switch statements, yay!)

Stable sorting might be more preferrable. In this case I can
a) sort first by rarity, second - by slot (and get artifacts sorted by slots, and within each group - sorted by rarity)
b) sort first by name, second - by slot (and get artifacts sorted by slots, and within each group - sorted by name).

luk3ZPL:
Scrolls on the end I think. If you add a weights to arts then could be easier to sort them.

SoundSSGood:
> @SoundSSGood you might be interested in this.
> 
> >

[...]

PR #4699, opened 2024-09-30
Backpack artifacts sorting

SoundSSGood:
https://github.com/vcmi/vcmi/issues/3426 Added algorithm for sorting artifacts in the backpack. 
Supports three sorting modes: by cost, by equipped slot, by artifact class.
image
Unfortunately I don't have icons for the buttons 🙁.
I also removed the quit button. Now it works the same as the quick backpack window. And as hd mod.

### DIFF
--- a/CCallback.cpp
+++ b/CCallback.cpp
@@ -192,6 +192,24 @@ void CCallback::scrollBackpackArtifacts(ObjectInstanceID hero, bool left)
 	sendRequest(&mba);
 }
 
+void CCallback::sortBackpackArtifactsBySlot(const ObjectInstanceID hero)
+{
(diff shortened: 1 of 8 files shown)


krs0:
Very useful feature!

Would it make sense for one change?

Scrolls are a complete separate category from the rest of artifacts.

So it would help to separate them and keep them always at the end, regardless of other sorting?

Like a button or checkbox that toggles, showing the scrolls at the end?

SoundSSGood:
Scrolls are also sorted depending on the sorting type. For example, if the sorting type is by cost, the scrolls will be sorted by cost along with the rest.

IvanSavenko:
Looks OK, merging. Will try to do something with icons for 1.6.

If somebody wants to make them - that would be great (only icon itself is needed, without button)

PR #4705, opened 2024-10-01
map object search

### LINKED ISSUE #4007: search on adventuremap
**Describe your proposal**
Add feature to find objects on adventuremap with shortcut CTRL+F (and button?). Often I search one specific object but I don't know anymore where exactly it was...

**Additional context**
Feature in SOD_SP looks like this:
grafik

Not optimal I think, to much elements. Maybe combine with filter (like in spellbook).

Laserlicht:
grafik

`CTRL+F` is used to open menu.
`ALT+F` to find next (alternatively menu can be reopened and select the same entry)

fixes #4007


misiokles:
How about searching under Fog of war? Should be not possible without vcmieagles

Laserlicht:
> How about searching under Fog of war? Should be not possible without vcmieagles

Only visible objects are searchable and in list.

Laserlicht:
[Peek 2024-10-01 23-27.webm](https://github.com/user-attachments/assets/f595b585-4169-4dab-bc21-c56df8e8ced4)

Laserlicht:
@IvanSavenko Anything to do here?

client/adventureMap/AdventureMapShortcuts.cpp
```diff
@@ -457,6 +463,72 @@ void AdventureMapShortcuts::zoom( int distance)
...
+{
+	// get all relevant objects
+	std::vector<ObjectInstanceID> visitableObjInstances;
+	int3 mapSizes = LOCPLINT->cb->getMapSize();
+	for(int x = 0; x < mapSizes.x; x++)
+		for(int y = 0; y < mapSizes.y; y++)
+			for(int z = 0; z < mapSizes.z; z++)
+				for(auto & obj : LOCPLINT->cb->getVisitableObjs(int3(x, y, z), false))
+					if(obj->ID != MapObjectID::MONSTER && obj->ID != MapObjectID::EVENT && obj->ID != MapObjectID::HERO)
+						visitableObjInstances.push_back(obj->id);
```
IvanSavenko, line 476: Sounds like rather overcomplicated solution.
Perhaps add method `getAllVisitableObjects` to `CGameInfoCallback` that will do the same via a single pass over map->objects?

client/adventureMap/AdventureMapShortcuts.cpp
```diff
@@ -457,6 +463,72 @@ void AdventureMapShortcuts::zoom( int distance)
...
+
+	// count of elements for each group
+	std::map<std::string, int> mapObjCount;
+	for(auto & obj : visitableObjInstances)
+		mapObjCount[{ VLC->objtypeh->getObjectName(LOCPLINT->cb->getObjInstance(obj)->getObjGroupIndex(), LOCPLINT->cb->getObjInstance(obj)->getObjTypeIndex()) }]++;
```
IvanSavenko, line 481: Avoid long and complex lines
```suggestion
	// count of elements for each group
	std::map<std::string, int> mapObjCount;
	for(auto & obj : visitableObjInstances)
	{
		const auto * objectPtr = LOCPLINT->cb->getObjInstance(obj);
		const std::string & objectName = VLC->objtypeh->getObjectName(objectPtr->getObjGroupIndex(), objectPtr->getObjTypeIndex());
		mapObjCount[objectName]++;
	}
```

IvanSavenko, line 481: And by the way - why you're using `VLC->objtypeh->getObjectName` instead of `CGObjectInstance::getObjectName`?

client/adventureMap/AdventureMapShortcuts.cpp
```diff
@@ -457,6 +463,72 @@ void AdventureMapShortcuts::zoom( int distance)
...
+	// sort by name
+	std::vector<std::pair<std::string, int>> textCountList;
+	for (auto itr = mapObjCount.begin(); itr != mapObjCount.end(); ++itr)
+		textCountList.push_back(*itr);
+	std::sort(textCountList.begin(), textCountList.end(),
+		[=](std::pair<std::string, int>& a, std::pair<std::string, int>& b)
+		{
+			return a.first < b.first;
+		}
+	);
```
IvanSavenko, line 492: Objects in std::map are already sorted. No need to sort it again.

client/adventureMap/AdventureMapShortcuts.cpp
```diff
@@ -457,6 +463,72 @@ void AdventureMapShortcuts::zoom( int distance)
...
+
+	// function to center element from list on map
+	auto selectObjOnMap = [this, textCountList, visitableObjInstances](int index)
+		{
+			auto selObj = textCountList[index].first;
+
+			// filter for matching objects
+			std::vector<ObjectInstanceID> selVisitableObjInstances;
+			for(auto & obj : visitableObjInstances)
+				if(selObj == VLC->objtypeh->getObjectName(LOCPLINT->cb->getObjInstance(obj)->getObjGroupIndex(), LOCPLINT->cb->getObjInstance(obj)->getObjTypeIndex()))
```
IvanSavenko, line 513: Same - why don't use `LOCPLINT->cb->getObjInstance(obj)->getObjectName()`?

Laserlicht:
Done :)

lib/CGameInfoCallback.cpp
```diff
@@ -479,6 +479,17 @@ std::vector <const CGObjectInstance *> CGameInfoCallback::getVisitableObjs(int3
 
 	return ret;
 }
+
+std::vector<ConstTransitivePtr<CGObjectInstance>> CGameInfoCallback::getAllVisitableObjs() const
+{
+	std::vector<ConstTransitivePtr<CGObjectInstance>> ret;
+	for(auto & obj : gs->map->objects)
+		if(obj->isVisitable() && obj->ID != Obj::EVENT && getTile(obj->pos, false))
```
IvanSavenko, line 487: `getTile` is needed to filter visible objects? There should be `isVisible(int3)` and `isVisible(CGObjectInstance*)` methods specifically for this.
```suggestion
		if(obj->isVisitable() && obj->ID != Obj::EVENT && isVisible(obj))
```

IvanSavenko (approved):

lib/CGameInfoCallback.cpp
Laserlicht: Yes, for visible objects. Thanks. That's a better fitting function. :)

misiokles:
It is possible to search for creatures?

Laserlicht:
> It is possible to search for creatures?

Creatures are ignored currently. To avoid blow up object list with many creature names.

Currently heroes, creatures and towns are ignored.

IvanSavenko:
While creature/heroes names will add quite a lot to noise, but might not be a bad idea to add:
- generally, you know name of object you want to add. So you can enter entire name of object and filter out everything else
- searching for specific hero or creature may also be useful, especially in campaigns and similar maps which may have victory conditions / quests targeting them

misiokles:
It seems that CTRL+F crashes the game when hero is on boat.

Laserlicht:
Also with this?
https://github.com/vcmi/vcmi/pull/4802

PR #4726, opened 2024-10-05
Separate vcmiqt static lib

wb180:
Attempt to share code between launcher and editor through the static lib.


launcher/modManager/cdownloadmanager_moc.cpp
```diff
@@ -10,7 +10,7 @@
 #include "StdInc.h"
 #include "cdownloadmanager_moc.h"
 
-#include "../launcherdirs.h"
+#include "vcmiqt/launcherdirs.h"
```
kambala-decapitator, line 13: external includes should use angle brackets
```suggestion
#include <vcmiqt/launcherdirs.h>
```

vcmiqt/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+set(vcmiqt_HEADERS
+		StdInc.h
+
+		jsonutils.h
+		launcherdirs.h
+		convpathqstring.h
+)
+
+assign_source_group(${vcmiqt_SRCS} ${vcmiqt_HEADERS})
+add_library(vcmiqt STATIC ${vcmiqt_SRCS} ${vcmiqt_HEADERS})
```
kambala-decapitator, line 17: it should be static only for "single app" builds. "Normal" desktop builds have launcher and editor as separate processes, so it's better to share code via dynamic (shared) library.

launcher/modManager/cdownloadmanager_moc.cpp
IvanSavenko: For vcmi library we use form like `"../lib/header.h"` - relative path & double-quotes

PR #4731, opened 2024-10-06
Reduce usage of pointers to VLC entities

IvanSavenko:
Final goal (of multiple PR's) is to remove all remaining pointers from serializable game state, and replace them with either identifiers ID's (which are simple integer internally) or with shared/unique pointers.

CGTownInstance::town and CGHeroInstance::type members have been removed. Now this data is computed dynamically using subID member.

VLC entity of a town can now be accessed via following methods:
- getFactionID() returns ID of a faction
- getFaction() returns pointer to a faction
- getTown() returns pointer to a town

VLC entity of a hero can now be accessed via following methods:
- getHeroTypeID() returns ID of a hero
- getHeroClassID() returns ID of a hero class
- getHeroType() returns pointer to a hero
- getHeroClass() returns pointer to a hero class

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -1454,7 +1454,7 @@ bool AIGateway::moveHeroToTile(int3 dst, HeroPtr h)
 
 void AIGateway::buildStructure(const CGTownInstance * t, BuildingID building)
 {
-	auto name = t->town->buildings.at(building)->getNameTranslated();
+	auto name = t->getTown()->buildings.at(building)->getNameTranslated();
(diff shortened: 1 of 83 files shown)


kambala-decapitator:
are pointers still sent between client and server? is this only about save file serialization?

IvanSavenko:
> are pointers still sent between client and server? is this only about save file serialization?

VLC entities are always serialized as integer / index in entities vector. Both for server-client communication and for saves.

However serializer still needs logic for "if this is a pointer to a VLC entity, then instead retrieve identifier and serialize it instead", which I would like to remove in the future (alongside with entire logic for serializing raw pointers)

And with town specifically there is edge case where game serializes CTown* member of CGTownInstance which is NOT a VLC entity (CFaction* is, and CFaction contains CTown), requiring another workaround in CGTownInstance::serialize

PR #4748, opened 2024-10-10
Quick workaround for failing Android CI

IvanSavenko:
Till proper fix in #4545 is in

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -17,6 +17,7 @@ env:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         include:
           - platform: linux-qt6


kambala-decapitator:
but what is the reason of the failure?

IvanSavenko:
https://github.com/actions/runner-images/pull/10741

Specifically, this part: (we're using `25.2.9519653`)
зображення

kambala-decapitator:
thanks. You could just switch to macos-12 image - it doesn't have a deprecation date yet.

IvanSavenko:
Might work, but I'd rather finish that PR. We can live without Android for a few days.

kambala-decapitator (approved):
ok, not insisting

PR #4766, opened 2024-10-12
Fused artifacts

SoundSSGood:
Fused artifacts have been added. Like combined ones, they are assembled from parts. Equipped in one slot. Cannot be disassembled.

To create a fused artifact add the following to json:
```
	"components": 
	[
		"artifact1",
		"artifact2",
		"artifact3"
	],
	"fusedComponents" : true,
```

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -356,6 +356,7 @@
 	"vcmi.heroWindow.sortBackpackBySlot.help"  : "Sort artifacts in backpack by equipped slot.",
 	"vcmi.heroWindow.sortBackpackByClass.hover"  : "Sort by class",
 	"vcmi.heroWindow.sortBackpackByClass.help"  : "Sort artifacts in backpack by artifact class. Treasure, Minor, Major, Relic",
+	"vcmi.heroWindow.fusingArtifact.fusing" : "You possess all of the components needed for the fusion of the %s. Do you wish to perform the fusion? {All components will be consumed upon fusion.}",
 
 	"vcmi.tavernWindow.inviteHero"  : "Invite hero",
 
--- a/client/ArtifactsUIController.cpp
+++ b/client/ArtifactsUIController.cpp
@@ -74,7 +74,10 @@ bool ArtifactsUIController::askToAssemble(const CGHeroInstance * hero, const Art
 					MetaString message = MetaString::createFromTextID(art->artType->getDescriptionTextID());
 					message.appendEOL();
 					message.appendEOL();
-					message.appendRawString(CGI->generaltexth->allTexts[732]); // You possess all of the components needed to assemble the
+					if(combinedArt->isFused())
+						message.appendRawString(CGI->generaltexth->translate("vcmi.heroWindow.fusingArtifact.fusing"));
+					else
+						message.appendRawString(CGI->generaltexth->allTexts[732]); // You possess all of the components needed to assemble the
 					message.replaceName(ArtifactID(combinedArt->getId()));
 					LOCPLINT->showYesNoDialog(message.toString(), [&assembleConfirmed, hero, slot, combinedArt]()
 						{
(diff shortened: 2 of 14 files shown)


Olegmods:
So fused artifact doesn't occupy the slots that its components? Or it's just cannot be disassembled?

SoundSSGood:
> So fused artifact doesn't occupy the slots that its components? Or it's just cannot be disassembled?

Both

IvanSavenko:
By the way - is it possible to fuse (or combine, H3 style) two identical artifacts?
In other words, will something like this work?
```json
"endlessBagOfGold":
{
	"components": 
	[
		"endlessPurseOfGold",
		"endlessPurseOfGold"
	]
}
```

(will take a look at PR a bit later)

by003:
Good for modders...I hope Rune Words can be designed in HoMM3.

SoundSSGood:
> By the way - is it possible to fuse (or combine, H3 style) two identical artifacts?

Now is possible

Alucard648:
H2 used fused artifacts, instead of combined ones. Example is Battle Garb of Anduran.

by003:
I have a suggestion:
Due to the irreversible of fusion and some potential clashes from other combined artifacts, 
You could: 
1. Make fused components DO NOT show reminding message while obtain all of them.
2. Create an icon to active it. The icon can be named as 'Fusion Mode'. If fusion mode is actived, click the components, they will prioritize fusion over combination.

SoundSSGood:
> 2\. The button can be named as 'Fusion Mode'

Where do you propose to place this button?
I think this will cause confusion among players. And if you don't know about this button, you might not even know that there are fused artifacts. 
Question for modders: What about a visitable object for crafting fused artifacts?
CC: @IvanSavenko @dydzio0614  @krs0  @misiokles @Olegmods

edeksumo:
> > 2. The button can be named as 'Fusion Mode'
> 
> Where do you propose to place this button? I think this will cause confusion among players. And if you don't know about this button, you might not even know that there are fused artifacts. Question for modders: What about a visitable object for crafting fused artifacts? CC: @IvanSavenko @dydzio0614 @krs0 @misiokles @Olegmods

I think that this is good idea but only, if you still preserve option for fuze in backpack. This will allow us to make powerful artifacts only in special object.

Olegmods:
> 
 > Where do you propose to place this button? I think this will cause confusion among players. And if you don't know about this button, you might not even know that there are fused artifacts. Question for modders: What about a visitable object for crafting fused artifacts? CC: @IvanSavenko @dydzio0614 @krs0 @misiokles @Olegmods

> 
> I think that this is good idea but only, if you still preserve option for fuze in backpack. This will allow us to make powerful artifacts only in special object.

I agree with @edeksumo 's opinion. Some artifacts (Armageddon's Blade, for example) could be fused in the special object. But some (the weaker ones) should be fuseable in the backpack.

misiokles:
Actually WoG has similar object (Artificer - script 26).
https://www.celestialheavens.com/h3maparchives/wog/wogfeatures.html

IvanSavenko:
Not sure what would be typical usages of this features, but:
- don't feel need of different handling compared to combined artifacts. At most - add warning that fusion is irreversible to assembly dialog (that dialog should have new message btw, in place of H3 "you possess all components...")
- new button - against it unless it is really necessary AND we have space for it (and for fusion I think my anwer is "no" to both points)
- custom map object - not sure. I guess we can add support for taking away artifacts in rewardable objects, so modders will be able to implement such object on their own - currently it only allows giving new artifacts. However it won't be connected in any way to this fusion logic - just seer hut-like object that asks for a set of artifacts and gives you another one(s).

SoundSSGood:
> * that dialog should have new message btw, in place of H3 "you possess all components...

Right. Will add.

by003:
I think the button can be added here. 
image

IvanSavenko:
> I think some players will ignore the warning information. It still can be considered.

And even more will not notice the button. 
And even more will be playing with no fusable artifacts but with unnecessary (for them) button.

krs0:
2 combined artifacts could collide, same with 2 fused ones, same with 1 fused and one combined.

Do not see the need of an extra button. Simple warning should be ok.

PR #4772, opened 2024-10-13
fix: use sdl mixer to get audio duration

Laserlicht:
Support getting audio length for other audio formats by using sdl_mixer.

Fixes error getting audio length when using .ogg as in @heroesiiifan new audio translation for german and @Fabrice-TIERCELIN for french.

Notable at incorrect scrolling speed of text.

Used this logic: https://discourse.libsdl.org/t/time-length-of-sdl-mixer-chunks/12852/2

### DIFF
--- a/client/media/CSoundHandler.cpp
+++ b/client/media/CSoundHandler.cpp
@@ -153,14 +153,15 @@ uint32_t CSoundHandler::getSoundDurationMilliseconds(const AudioPath & sound)
 
 	int freq = 0;
 	Uint16 fmt = 0;
-	int chans = 0;
-	if(!Mix_QuerySpec(&freq, &fmt, &chans))
+	int channels = 0;
+	if(!Mix_QuerySpec(&freq, &fmt, &channels))
 		return 0;
 
 	if(chunk != nullptr)
 	{
-		Uint32 points = (chunk->alen / ((fmt & 0xFF) / 8));
-		Uint32 frames = (points / chans);
+		Uint32 sampleSizeBytes = (fmt & 0xFF) / 8;
+		Uint32 samples = (chunk->alen / sampleSizeBytes);
+		Uint32 frames = (samples / channels);
 		milliseconds = ((frames * 1000) / freq);
 
 		Mix_FreeChunk(chunk);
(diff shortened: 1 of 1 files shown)


Fabrice-TIERCELIN:
FYI, I have used `.ogg` because the sound had problems with `.wav`.

Laserlicht:
> FYI, I have used `.ogg` because the sound had problems with `.wav`.

Yes, no problem. Ogg Vorbis is officially supported by VCMI. Because of that we should fix this problem anyway.

Wav should be work too. Could be another bug.

client/media/CSoundHandler.cpp
```diff
@@ -147,18 +147,23 @@ uint32_t CSoundHandler::getSoundDurationMilliseconds(const AudioPath & sound)
...
+	if(chunk != nullptr)
 	{
-		SDL_FreeWAV(audioBuf);
-		uint32_t sampleSize = SDL_AUDIO_BITSIZE(spec.format) / 8;
-		uint32_t sampleCount = audioLen / sampleSize;
-		uint32_t sampleLen = sampleCount / spec.channels;
-		milliseconds = 1000 * sampleLen / spec.freq;
+		Uint32 points = (chunk->alen / ((fmt & 0xFF) / 8));
+		Uint32 frames = (points / chans);
+		milliseconds = ((frames * 1000) / freq);
```
IvanSavenko, line 164:
```suggestion
		Uint32 sampleSizeBytes = (fmt & 0xFF) / 8;
		Uint32 samples = (chunk->alen / sampleSizeBytes);
		Uint32 frames = (points / chans);
		milliseconds = ((frames * 1000) / freq);
```

client/media/CSoundHandler.cpp
```diff
@@ -147,18 +147,23 @@ uint32_t CSoundHandler::getSoundDurationMilliseconds(const AudioPath & sound)
...
-	uint32_t audioLen;
-	uint8_t * audioBuf;
 	uint32_t milliseconds = 0;
 
-	if(SDL_LoadWAV_RW(SDL_RWFromMem(data.first.get(), data.second), 1, &spec, &audioBuf, &audioLen) != nullptr)
+	Mix_Chunk * chunk = Mix_LoadWAV_RW(SDL_RWFromMem(data.first.get(), data.second), 1);
+
+	int freq = 0;
+	Uint16 fmt = 0;
+	int chans = 0;
```
IvanSavenko, line 156:
```suggestion
	int channels = 0;
```

PR #4782, opened 2024-10-15
Ship obstacles fix

### LINKED ISSUE #4781: Ship-to-ship battle obstacles
**Describe the bug**
The battlefield layout in ship-to-ship battles may prevent land units from reaching the other ship.

**Expected behavior**
At least 1 possible route to the enemy ship.

**Actual behavior**
[This obstacle](https://github.com/vcmi/vcmi/blob/94a0de716a9b652193abdfb164b020b7969d1886/config/obstacles.json#L539-L547) gets placed in ship-to-ship battles and reachability between ships is not checked, so it may end up blocking all possible paths.

**Screenshots**
Screenshot 2024-10-15 at 9 20 53

**Additional context**
In a Discord discussion, @IvanSavenko suggested that maybe this obstacle should be for sand-shore battlefields (i.e. land-to-ship) instead, but this must be verified in original H3 first.

Olegmods:
Should fix #4781

### DIFF
--- a/config/battlefields.json
+++ b/config/battlefields.json
@@ -162,5 +162,5 @@
 			159, 160, 161, 162, 163,
 			176, 177, 178, 179, 180]
 	},
-	"ship": { "graphics" : "CMBKDECK.BMP" }
+	"ship": { "graphics" : "CMBKDECK.BMP", "isSpecial" : true}
 }
--- a/config/obstacles.json
+++ b/config/obstacles.json
@@ -538,7 +538,7 @@
 	},
 	"55":
 	{
-		"allowedTerrains" : ["water"],
+		"specialBattlefields" : ["ship"],
 		"width" : 3,
 		"height" : 3,
 		"blockedTiles" :  [-15, -16, -33],
--- a/config/obstacles.json
+++ b/config/obstacles.json
@@ -538,6 +538,7 @@
 	},
 	"55":
 	{
+		"allowedTerrains" : ["water"],
 		"specialBattlefields" : ["ship"],
 		"width" : 3,
 		"height" : 3,
--- a/config/obstacles.json
+++ b/config/obstacles.json
@@ -538,7 +538,6 @@
 	},
 	"55":
 	{
-		"allowedTerrains" : ["water"],
 		"specialBattlefields" : ["ship"],
 		"width" : 3,
 		"height" : 3,


Olegmods:
> This change removes obstacles in ship-to-ship battles. However, it also removes obstacles in land-to-ship battles (not sure if this is desired).

Hmm, that's strange. It should appear on a single ship layout. The ship doesn't have obstacles every single time?

smanolloff:
Yeah, tried it 5-6 times, always an empty battlefield. If I revert the config back, the "rope" obstacles appear (both in ship-to-ship and single-ship)

Olegmods:
Will try to test myself.

smanolloff:
I think it will work if you set `isSpecial` in [battlefields.json](https://github.com/vcmi/vcmi/blob/94a0de716a9b652193abdfb164b020b7969d1886/config/battlefields.json):
Marking both `ship` and `ship_to_ship` as special means that [this](https://github.com/vcmi/vcmi/blob/94a0de716a9b652193abdfb164b020b7969d1886/lib/ObstacleHandler.cpp#L87-L88) code branch will be triggered in both cases, but obstacles will be placed only for `ship` (due to the obstacle's `specialBattlefields` config).

Olegmods:
> I think it will work if you set `isSpecial` in [battlefields.json](https://github.com/vcmi/vcmi/blob/94a0de716a9b652193abdfb164b020b7969d1886/config/battlefields.json): Marking both `ship` and `ship_to_ship` as special means that [this](https://github.com/vcmi/vcmi/blob/94a0de716a9b652193abdfb164b020b7969d1886/lib/ObstacleHandler.cpp#L87-L88) code branch will be triggered in both cases, but obstacles will be placed only for `ship` (due to the obstacle's `specialBattlefields` config).
> 
> I tried that and seems to do the job. Not sure what are the possible side effects of marking those two battlefields as special, though.

Thanks for clarification! I won't change it in the PR for now. I'll try to test it on my PC when I'll have access to it. If I won't encounter any crashes due to the changes, I'll add these to the PR. Or I may add the changes if someone more knowledgeable than me will say that it would work without side effects.

Olegmods:
From my testing, it works correctly.

In the docs, it stated that `isSpecial` works only on obstacles. So I think the PR could be merged after the all workflows will be completed.

PR #4785, opened 2024-10-16
Subtitles for videos / sync

### LINKED ISSUE #4783: Support for Subtitles in Videos via .srt Files
Hi,

Could you please add subtitle support via .srt files for the H3INTRO and ENDGAME videos in Heroes of Might and Magic III? This would greatly improve accessibility for players.

Thanks!

Laserlicht:
Fixes #4783

Todo:
- [x] Parse subtitles only once (not every tick)

### DIFF
--- a/client/media/CVideoHandler.h
+++ b/client/media/CVideoHandler.h
@@ -77,10 +77,12 @@ class CVideoInstance final : public IVideoInstance, public FFMpegStream
 	SDL_Surface * surface = nullptr;
 	Point dimensions;
 
-	/// video playback current progress, in seconds
-	double frameTime = 0.0;
+	/// video playback start time point
+	std::chrono::high_resolution_clock::time_point startTime;
 
 	void prepareOutput(float scaleFactor, bool useTextureOutput);
+	
+	const int MAX_FRAMESKIP = 5;
 
 public:
 	~CVideoInstance();
--- a/client/media/CVideoHandler.h
+++ b/client/media/CVideoHandler.h
@@ -88,6 +88,7 @@ public:
 	void openVideo();
 	bool loadNextFrame();
 
+	double timeStamp() final;
 	bool videoEnded() final;
 	Point size() final;
 
--- a/client/media/CVideoHandler.h
+++ b/client/media/CVideoHandler.h
@@ -78,7 +78,8 @@ class CVideoInstance final : public IVideoInstance, public FFMpegStream
 	Point dimensions;
 
 	/// video playback start time point
-	std::chrono::high_resolution_clock::time_point startTime;
+	std::chrono::steady_clock::time_point startTime;
+	std::chrono::steady_clock::time_point deactivationStartTime;
 
 	void prepareOutput(float scaleFactor, bool useTextureOutput);
 	
@@ -96,6 +97,8 @@ public:
 
 	void show(const Point & position, Canvas & canvas) final;
 	void tick(uint32_t msPassed) final;
+	void activate() final;
+	void deactivate() final;
 };
 
 class CVideoPlayer final : public IVideoPlayer
--- a/client/widgets/VideoWidget.cpp
+++ b/client/widgets/VideoWidget.cpp
@@ -38,26 +38,19 @@ void VideoWidgetBase::playVideo(const VideoPath & fileToPlay)
 {
 	OBJECT_CONSTRUCTION;
 
-	using SubTitlePath = ResourcePathTempl<EResType::SUBTITLE>;
-	SubTitlePath subTitlePath = fileToPlay.toType<EResType::SUBTITLE>();
-	SubTitlePath subTitlePathVideoDir = subTitlePath.addPrefix("VIDEO/");
+	JsonPath subTitlePath = fileToPlay.toType<EResType::JSON>();
+	JsonPath subTitlePathVideoDir = subTitlePath.addPrefix("VIDEO/");
 	if(CResourceHandler::get()->existsResource(subTitlePath))
-	{
-		auto rawData = CResourceHandler::get()->load(subTitlePath)->readAll();
-		srtContent = std::string(reinterpret_cast<char *>(rawData.first.get()), rawData.second);
-	}
-	if(CResourceHandler::get()->existsResource(subTitlePathVideoDir))
-	{
-		auto rawData = CResourceHandler::get()->load(subTitlePathVideoDir)->readAll();
-		srtContent = std::string(reinterpret_cast<char *>(rawData.first.get()), rawData.second);
-	}
+		subTitleData = JsonNode(subTitlePath);
+	else if(CResourceHandler::get()->existsResource(subTitlePathVideoDir))
+		subTitleData = JsonNode(subTitlePathVideoDir);
 
 	videoInstance = CCS->videoh->open(fileToPlay, scaleFactor);
 	if (videoInstance)
 	{
 		pos.w = videoInstance->size().x;
 		pos.h = videoInstance->size().y;
-		subTitle = std::make_unique<CMultiLineLabel>(Rect(0, (pos.h / 5) * 4, pos.w, pos.h / 5), EFonts::FONT_HIGH_SCORE, ETextAlignment::CENTER, Colors::WHITE, "");
+		subTitle = std::make_unique<CMultiLineLabel>(Rect(0, (pos.h / 5) * 4, pos.w, pos.h / 5), EFonts::FONT_HIGH_SCORE, ETextAlignment::CENTER, Colors::WHITE);
 	}
 
 	if (playAudio)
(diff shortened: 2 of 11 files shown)


client/widgets/VideoWidget.cpp
```diff
@@ -33,11 +36,28 @@ VideoWidgetBase::~VideoWidgetBase() = default;
 
 void VideoWidgetBase::playVideo(const VideoPath & fileToPlay)
 {
+	OBJECT_CONSTRUCTION;
+
+	using SubTitlePath = ResourcePathTempl<EResType::SUBTITLE>;
```
kambala-decapitator, line 41: not sure why you need this alias if you can use auto

client/widgets/VideoWidget.cpp
```diff
@@ -33,11 +36,28 @@ VideoWidgetBase::~VideoWidgetBase() = default;
...
+	SubTitlePath subTitlePath = fileToPlay.toType<EResType::SUBTITLE>();
+	SubTitlePath subTitlePathVideoDir = subTitlePath.addPrefix("VIDEO/");
+	if(CResourceHandler::get()->existsResource(subTitlePath))
+	{
+		auto rawData = CResourceHandler::get()->load(subTitlePath)->readAll();
+		srtContent = std::string(reinterpret_cast<char *>(rawData.first.get()), rawData.second);
```
kambala-decapitator, line 47: what about encoding?

client/widgets/VideoWidget.cpp
```diff
@@ -91,6 +113,34 @@ void VideoWidgetBase::stopAudio()
 	}
 }
 
+std::string VideoWidgetBase::getSubTitleLine(double timestamp)
+{
+	if(srtContent.empty())
+		return "";
```
kambala-decapitator, line 119:
```suggestion
		return {};
```

client/widgets/VideoWidget.cpp
```diff
@@ -91,6 +113,34 @@ void VideoWidgetBase::stopAudio()
...
+	
+    std::regex exp("^\\s*(\\d+:\\d+:\\d+,\\d+)[^\\S\\n]+-->[^\\S\\n]+(\\d+:\\d+:\\d+,\\d+)((?:\\n(?!\\d+:\\d+:\\d+,\\d+\\b|\\n+\\d+$).*)*)", std::regex::multiline);
```
kambala-decapitator, line 121: wow that looks terrible... A parser would be better, simply istream might even work. Please also add a comment with line format, then we'll see how to improve it.

client/widgets/VideoWidget.cpp
```diff
@@ -91,6 +113,34 @@ void VideoWidgetBase::stopAudio()
...
+		double timestamp2 = std::stoi(timestamp2Str[0]) * 3600 + std::stoi(timestamp2Str[1]) * 60 + std::stoi(timestamp2Str[2]) + (std::stoi(timestamp2Str[3]) / 1000.0);
+        std::string text = res[3];
+		text.erase(0, text.find_first_not_of("\r\n\t "));
+
+		if(timestamp > timestamp1 && timestamp < timestamp2)
+			return text;
+		
+        searchStart = res.suffix().first;
+    }
+	return "";
```
kambala-decapitator, line 141:
```suggestion
	return {};
```

client/widgets/VideoWidget.cpp
```diff
@@ -91,6 +113,34 @@ void VideoWidgetBase::stopAudio()
 	}
 }
 
+std::string VideoWidgetBase::getSubTitleLine(double timestamp)
```
kambala-decapitator, line 116: please fix indentation in this method

client/widgets/VideoWidget.cpp
```diff
@@ -91,6 +113,34 @@ void VideoWidgetBase::stopAudio()
...
+		std::vector<std::string> timestamp1Str;
+		boost::split(timestamp1Str, static_cast<std::string>(res[1]), boost::is_any_of(":,"));
+		std::vector<std::string> timestamp2Str;
+		boost::split(timestamp2Str, static_cast<std::string>(res[2]), boost::is_any_of(":,"));
+		double timestamp1 = std::stoi(timestamp1Str[0]) * 3600 + std::stoi(timestamp1Str[1]) * 60 + std::stoi(timestamp1Str[2]) + (std::stoi(timestamp1Str[3]) / 1000.0);
+		double timestamp2 = std::stoi(timestamp2Str[0]) * 3600 + std::stoi(timestamp2Str[1]) * 60 + std::stoi(timestamp2Str[2]) + (std::stoi(timestamp2Str[3]) / 1000.0);
+        std::string text = res[3];
+		text.erase(0, text.find_first_not_of("\r\n\t "));
+
+		if(timestamp > timestamp1 && timestamp < timestamp2)
```
kambala-decapitator, line 136: would be much better to parse string content only once and then search your data structure

client/widgets/VideoWidget.h
```diff
@@ -14,19 +14,23 @@
...
 class VideoWidgetBase : public CIntObject
 {
 	std::unique_ptr<IVideoInstance> videoInstance;
+	std::unique_ptr<CMultiLineLabel> subTitle;
 
 	std::pair<std::unique_ptr<ui8[]>, si64> audioData = {nullptr, 0};
 	int audioHandle = -1;
 	bool playAudio = false;
 	float scaleFactor = 1.0;
+	std::string srtContent = "";
```
kambala-decapitator, line 28:
```suggestion
	std::string srtContent;
```

kambala-decapitator (dismissed):

client/widgets/VideoWidget.cpp
kambala-decapitator: oh, didn't notice your TODO

kambala-decapitator:
std::regex::multiline is available only since clang 15 it seems, but android and Apple platforms use clang 14. Just use boost's regex for now, should be a simple std -> boost replacement.

client/widgets/VideoWidget.cpp
Laserlicht: Yes. Also a non noticed todo. Was the fastest way to get it basically working. Just copy an existing regex.

I've only published it as draft to show that I'm currently working on it. 😅

client/widgets/VideoWidget.cpp
Laserlicht: Haven't tested it yet with non ASCII stuff. How to handle it correctly as utf-8 in C++?

IvanSavenko:
Do we actually need srt?
Might be easier to add some json-based format in there, e.g:
```json
[
    {
        "timeStart" : 5.640, // start time, seconds
        "timeStart" : 8.120, // end time, seconds
        "text" : " ... " // text to show during this period
    },
    ...
]
```
I doubt that we can provide non-primitive support for srt.

Base spec might be simple, but modern video players have massive number of extensions which we won't support (e.g. text styling). Meanwhile our text renderer uses different format for styling, e.g. `{ ... }` to create yellow text, so standard srt editors likely won't work with our srt's

So different format might not be a bad idea.
And it would avoid painful parsing of srt like that dark magic level regex

Laserlicht:
Hmm... yes. Could be a good idea to change it to json.

Only advantage of srt I see: You can use external tools to create and test it. For example you can load it via VLC player.

But to be honest: For the few subtitles we have full srt support could be overengeneered.

IvanSavenko:
Yeah. There are only two video files with audio after all.
And if we really need, it should be possible to make srt->json converter, with those amazing regexp's

dydzio0614:
I think our video support level is already overkill for release 1.6 considering its low impact on gameplay and lacking stuff in other VCMI areas (map editor, launcher mod dependencies management, mechanics / visual related bugs in battles etc.)

IvanSavenko:
@dydzio0614 and there is also that... Will likely look into launcher/dependencies stuff for 1.6, but everything else - probably not.

Laserlicht:
Changed to JSON format.

altiereslima:
The video audio is significantly out of sync, making it very difficult to get the subtitles to work properly.

Laserlicht:
> The video audio is significantly out of sync, making it very difficult to get the subtitles to work properly.

Can you test again? Video and Audio should be much better synced now (especially on low end hardware).

altiereslima:
Now it's working perfectly.

Laserlicht:
Also fixes slow playback with xbrz (due frameskipping).

Any edge case to test?

client/media/CVideoHandler.h
```diff
@@ -77,17 +77,20 @@ class CVideoInstance final : public IVideoInstance, public FFMpegStream
 	SDL_Surface * surface = nullptr;
 	Point dimensions;
 
-	/// video playback current progress, in seconds
-	double frameTime = 0.0;
+	/// video playback start time point
+	std::chrono::high_resolution_clock::time_point startTime;
```
IvanSavenko, line 81: steady_clock might be more suitable.
We don't need max resolution, but we don't want clock to suddently return smaller time stamp than before

client/widgets/VideoWidget.cpp
```diff
@@ -33,11 +36,21 @@ VideoWidgetBase::~VideoWidgetBase() = default;
...
+		subTitleData = JsonNode(subTitlePath);
+	else if(CResourceHandler::get()->existsResource(subTitlePathVideoDir))
+		subTitleData = JsonNode(subTitlePathVideoDir);
+
 	videoInstance = CCS->videoh->open(fileToPlay, scaleFactor);
 	if (videoInstance)
 	{
 		pos.w = videoInstance->size().x;
 		pos.h = videoInstance->size().y;
+		subTitle = std::make_unique<CMultiLineLabel>(Rect(0, (pos.h / 5) * 4, pos.w, pos.h / 5), EFonts::FONT_HIGH_SCORE, ETextAlignment::CENTER, Colors::WHITE);
```
IvanSavenko, line 53: Perhaps only create if subtitle data is not empty?

IvanSavenko:
Also - @kambala-decapitator any further comments?

PR #4787, opened 2024-10-16
Support for mod presets (part 1)

IvanSavenko:
Initial part of support for mod presets. Main goal for this PR is to make mod installation and mod activation completely separate actions internally and resolve various issues related to mod dependencies.

Actual switchable mod presets (and any UI for those) will go into a separate PR later.

Progress:
- [x] Store list of active mods in a named preset
- [x] Interpret submods as 'settings' of a mod internally, and store them as part of preset. This should clarify various edge cases when submod is considered to be 'active' even though parent mod is inactive
- [x] Automatic migration - create default preset using 1.5 active mods list if no presets exist
- [x] Handle mod installation (or migration) - automatically add all submods to settings of a preset, taking into account 'keepDisabled' property in submods. 
- [x] Prevent cases where player can end up with missing dependencies, such as disabling main mod that has submod on which another mod depends.
- [x] More clear display of submod active/inactive status in Launcher - for example, use grayed-out icons for submods of inactive mods
- [x] Allow updates of a mod if another active mod depends on it
- [x] Better handling of dependencies/conflicts list in UI - show name of parent mod for submods in addition to submod name
- [x] Better handling of dependencies/conflicts list in UI - if mod depends on a submod of another, not installed mod, then show name of parent mod instead (due to Launcher having no knowledge of submods of not installed mods)

NOT part of this PR (to avoid large PR and potential painful conflicts in Qt UI files)
- UI to toggle presets
- Add virtual and always available "no mods" preset as a quick option to disable all mods
- Add exporting / importing / copying of a preset. Consider format of an exported preset. Json? gzip + base64?
- Install of missing mods from newly imported preset
- Consider handling of translation mods in presets - probably should be excluded from exported preset.
- Add 'update all mods' button
- (not for 1.6) consider integration of presets with MP - show preset used in MP game and a way to apply it on local system.


dydzio0614:
we also want "update all mods", "disable all mods" buttons somewhere

IvanSavenko:
Disable all mods - I plan to add virtual "No mods" preset that player can always enable
Update all mods - noted, but all changes that require changing UI files will go into separate PR

dydzio0614:
i also remember there is issue if you want to disable / uninstall mod that depends on another mod

GeorgeK1ng:
Nice, this is what I'm currently looking for when I have tons of mods installed. Create config and save that preset to be able to switch between wanted scenarios in second instead of manualy searching for mods / submods

GeorgeK1ng:
This feature could be useful also with online lobby. Because now you can see only list of mods that must be disable / enabled to allow you connect to game. When presets exists then can could be added some kind of new button to lobby to allow export mod list settings. Like "RoomName_ModSettings". Of course chaning these mods settings on the fly in lobby will be cool, but currently are loaded on game launch. So maybe considering about "Reload" button too. So user can choose their wanted game room, generate preset, hit "Reload" and then will be current game killed and loaded again using preset into same room window. Something like this. It's just idea: 
image

IvanSavenko:
Finally got some progress with this PR. 
Initially ended up with a lot of almost identical code in lib and in launcher - one that operates on JsonNode's and std::string's and another one that operates on QVariantMap's and QString's. And this code was getting progressively more complex.

So decided to use more lengthy approach and to minimize amount of such duplicated code.

Now most of mod management (other than downloading / installing) is implemented in library. Qt launcher code now mostly contains wrapper to convert between std<->Qt classes.

### Current progress
- Launcher now always allow disabling of a mod. All mods that depend on this mod (directly or indirectly) will be automatically disabled

- Similarly, Launcher (almost) always allows enabling of a mod. All conflicting mods (and mods that depends on them) will be automatically disabled. However enabling mod is not possible if one of its dependencies (direct or indirect) are not installed - since downloading mod on attempt to enable different mod is not exactly expected.

- Added repository caching. Now when launcher starts, it will immediately load cached repository copy. This eliminates jumpy behavior on launcher starts, when launcher finishes checkout and adds mods from repository to mod list.

- screenshot/changelog tabs are now only accessible when mod provides them. As result, you can now see which mods have them just by looking at tab list.

### TODO
- restore & recheck and remove restrictions on:
  - mod uninstalling
  - mod installing
  - mod updates
- do not display submods of not installed mods as mod ID's but rather with names of parent mod
- update "Notes" block in Launcher to account for less restrictive mod management
- implement remaining FIXME's and TODO's in code
- reimplement language-specific options support - show translated fields in mod description
- consider allowing other language-specific options e.g screenshots
- show 'reverse conflicts' in mod description - when another mod marks this mod as a conflicting.
- consider hinting which mods are active in list of mod dependencies / conflicts in UI.
- regenerate Qt translations
- update VCMI translations

krs0:
_However enabling mod is not possible if one of its dependencies (direct or indirect) are not installed - since downloading mod on attempt to enable different mod is not exactly expected_

If I want to install a mod, I expect that all missing dependencies are installed as well. So if this logic is in place, could this not be used for enabling missing dependencies as well? So missing mods are installed and enabled?

When 1 mod depends on WoG graphic fix, and wants the rest of WoG submods disabled, at the moment those submods need to be set as conflicts! So if player would like to also enable one of those (level8 creatures for example), needs to modify localy the mod.json and delete that artificial conflict.
Upon dependency install, could only submods that are specified as dependency be enabled? and the rest set disabled by default?

I think client now performs some checks on disabled mods. Could those be skipped with the new changes? Would make warning detection easier in mod development, becausecyou will see errors/warnings only for the enabled mods.

IvanSavenko:
> If I want to install a mod, I expect that all missing dependencies are installed as well. So if this logic is in place, could this not be used for enabling missing dependencies as well? So missing mods are installed and enabled?

Possible. But having an "Enable" button that may download something is not exactly expected from player point of view. So for now keeping as it, to avoid more complex logic, can be reviewed & changed later based on usability feedback.

> When 1 mod depends on WoG graphic fix, and wants the rest of WoG submods disabled, at the moment those submods need to be set as conflicts!  So if player would like to also enable one of those (level8 creatures for example), needs to modify localy the mod.json and delete that artificial conflict. Upon dependency install, could only submods that are specified as dependency be enabled? and the rest set disabled by default?

For me it feels like a job for presets, not for mods themselves. I am considering how to allow providing preconfigured mod presets via our mod system / repository, so player will automatically get preset defined in installed mods.
As well as how to use this feature in multiplayer - so when player creates room when using specific preset, it will be shown in room description and other player does not have to check entire mod list to confirm that this is expected configuration and is what he wants to play with. 

However this is likely not for 1.6, and definitely not for 1.6.0.

> I think client now performs some checks on disabled mods. Could those be skipped with the new changes? Would make warning detection easier in mod development, becausecyou will see errors/warnings only for the enabled mods.

Only mod.json is loaded for inactive mods. Probably makes sense to disable validation for inactive mods. But loading them is easier, especially since I'd like to eventually support hot-loading for mods (at least for installed ones)

IvanSavenko:
Mostly done, few small points remaining:
- [ ] check that submods are correctly re-enabled when enabling parent mod indirectly, due to dependency
- [ ] check that submod configuration remains the same if main mod is disabled by enabling conflicting mod
- [ ] check that submods of submods enable/disable correctly
- [ ] check that compatiblity submods are always enabled when needed
- [ ] consider automatic re-enabling of translation mods (and disabling them explicitly when language changes)
- [ ] investigate why download size is not set in progress bar when installing mods
- [ ] consider displaying *total* download size in mod description, accounting for any missing dependencies

krs0:
If a mod is completely disabled, it does not make sense to appear in preset. It only complicates the json.

EG: I had HotA enabled then disabled it, but it remained in presets with everything set to false.

image

So if all submods are disabled, remove from preset.json

IvanSavenko:
> If a mod is completely disabled, it does not make sense to appear in preset. It only complicates the json.

Not exactly. If player later re-enabled mod, he generally would want to keep same submod configuration as he had before.

For exporting presets - yeah, should be trimmed. But that's not part of this PR

Everything being set to false is a bug, have yet to fix it. Goal is to remember submod configuration.

IvanSavenko:
Presumably ready. Final list of player-visible changes (to be copy-pasted to changelog):
- It is now always possible to disabled or uninstall a mod. Any mods that depend on this mod will be automatically disabled
- It is now always possible to update a mod, even if there are mods that depend on this mod.
- It is now possible to enable mod that conflicts with already active mod. Conflicting mods will be automatically disabled
- If main mod is disabled, all its submods will have their active or inactive status shown as greyed-out for clarity
- If mod depends or conflicts with a submod, Launcher will now also show name of parent mod in list of dependencies / conflicts
- Game will now cache result of mod repository checkout and restore it immediately on next start. This removes flickering when game fills list of available mods.
- Screenshot and Changelog tabs in mod description are now disabled for mods that do not have them.
- Launcher will now correctly show conflicts on both sides - if mod A is marked as conflicting with B, then information on this conflict will be shown in description of both mod A and mod B (instead of only in mod B)

(no references of presets for now, since at the moment they are internal detail and are not expose to UI yet)

PR #4788, opened 2024-10-16
MMAI: an ML-powered combat AI

smanolloff:
# MMAI: an ML-powered combat AI

This PR adds a new battle AI powered by machine learning.

## Description

`MMAI` is a new type of AI which controls units in battle based on the output of an ML model pre-trained in the in the [`vcmi-gym`](https://github.com/smanolloff/vcmi-gym) project using Reinforcement Learning algorithms. It aims to improve the HOMM3 experience by making AIs non-scripted, less-predictive and more challenging for the player.

Being the very first MMAI release, it is marked as **experimental**. It was tested only by me and I expect there will be issues that need fixing. The models are not very smart (yet), but training new ones is an ongoing effort which will not stop here -- with the gates now open, _the sky is the limit_ ;)

#### New AI lib: `MMAI`

A new git submodule under `AI/MMAI` is added. It is compiled only if the `ENABLE_MMAI` compile option is `ON` (default: `OFF`). When enabled, MMAI can be selected as the neutral AI from the launcher:

launcher-combobox
<br><br>

Whenever MMAI is engaged in combat, it loads a pre-trained ML model (from disk or cache) and uses it for predicting the actions throughought the battle.

#### New mod: `MMAI`

A new git submodule under `Mods/MMAI` is added. It contains the models loaded by the MMAI lib.

launcher-mod
<br><br>

The MMAI lib cannot function without the MMAI mod. The reason the two are made separate is to allow new model versions to be delivered through mod updates instead.

mmai-components

#### New CI builds

Several new CI builds are added which produce MMAI-enabled VCMI packages. These packages require no additional setup after installation and can be used to test this new feature. The builds can be identified by the "-mmai" suffix, e.g. `11296438272-vcmi-PR-4274-420935d-mmai`.

I am not sure if separate builds for MMAI will be needed in the future, as it could be included in the default builds instead. I am keeping them separate for now to allow to easily switch between VCMI built with and without MMAI.

#### AvA testing (AI vs AI 😀)

Measuring how "good" the AI is, is not an easy task. I did a simple test comparing MMAI to the two existing AIs: BattleAI and StupidAI. The following ground rules were set for the test:
* each battle uses a random pair of heroes
* each battle uses a randomized terrain+obstacle layout
* armies contain HOMM3 SOD units only
* no hero spellcasting (hero mana is 0)
* no morale and luck (set to 0 through a combination of skills and artifacts)

| Attacker | Defender | MMAI winrate |
| - | - | - |
| BattleAI (simulated turns=10) | MMAI | 46% |
| StupidAI | MMAI | 75% |

Each result is based on 500 battles played with randomly selected armies. A total of 512 different armies are used in the test, grouped into 8 pools based on their total AI value: 25k, 50k, 100k, 200k, 400k, 800k and 1.6M This map is similar to the one used during training, but contains different armies. (see the ML-related section for more details).

#### Speed

Using a modified VCMI binary, I measured the average time needed to act for each of the 3 AIs. The test was performed on an Intel Core i7-7700K - a 4-core/8-thread desktop CPU produced 7-8 years ago. The "Action time" includes server processing time and packet roundtrips, so StupidAI is likely faster than shown here:

<table>
  <thead>
    <tr>
      <th>AI</th>
      <th>Action time<br>(3-stack army)</th>
      <th>Action time<br>(7-stack army)</th>
      <th>Action time<br>(10-stack* army)</th>
      <th>CPU utilization</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>StupidAI</td>
      <td>0.002s</td> 
      <td>0.003s</td>
      <td>0.003s</td>
      <td>100% (1 core)</td>
    </tr>
    <tr>
      <td>BattleAI</td>
      <td>0.01s</td>
      <td>0.07s</td>
      <td>0.14s</td> 
      <td>100% (1 core)</td> 
    </tr>
    <tr>
      <td>MMAI</td>
      <td>0.03s</td>
      <td>0.04s</td>
      <td>0.05s</td>
      
      <td>800% (8 cores)</td> 
    </tr>
  </tbody>
</table>

\* 10 stacks = 7 stacks + 3 war machines

Although MMAI uses more CPU, it is faster than BattleAI on "crowded" battlefields as it utilizes all available cores, making it seem instant during gameplay.

#### Memory usage

MMAI's memory footprint is minor compared to modern standards:

| Component | Memory used |
| - | - |
| MMAI lib | 40 MB |
| MMAI model | 30 MB |
| **Total** | **100 MB** |

The important thing to note is that once loaded, an MMAI model is reused throughought the lifetime of the VCMI process and multiple MMAIs (e.g. for multiple opponents) incur no additional memory overhead.

These results are consistent across the 3 different devices tested: Mac, iPhone and Windows.

#### Artifact Size

The size of the produced CI artifact with MMAI enabled varies depending on the platform:

| Platform | Artifact size (no MMAI) | Artifact size (with MMAI) |
| - | - | - |
| ios | 19.3 MB | 39.5 MB |
| mac-arm | 25.5 MB | 54.6 MB |
| mac-intel | 30.5 MB | 97.9 MB |
| msvc | 32.9 MB | 57.2 MB |

There's a ~7MB fixed increase due to the bundled MMAI mod (because of the model contained within) and rest is due to the bundled `libtorch` library for the NN.

#### Training-related information

The model was trained for ~15 days in a hardware-constrained environment (my personal MacBook M2). The training session consisted of ~1.2M battles played vs. StupidAI, each battle with randomly chosen armies among a total of 4096 armies available on the map, pre-balanced armies of similar strength.

The "rebalancing" of the armies is done before the training starts and consists of repeated modifications made to the armies based on statistical data collected from StupidAI vs StupidAI battles (~1M battles total), in order to minimize the "spread" of the distribution of winrates across the armies. Below is a screenshot of the map used, as well as histograms with the winrate distributions before and after rebalancing (more spread = less balanced):

testmap
testmap-winrates
<br><br>

The model was trained using PPO-DNA (Proximal Policy Optimization with Dual Network Architecture) RL algorithm with added invalid action masking and hyperparam optimization through PBT (Population Based Training).

The neural network used here is a 4-layer deep neural network with 2.7M parameters. This will likely change in the future, as the NN architecture plays a vital part of a model's performance and usually frequently changed when training new models.

The input layer of the network has 12685 units, i.e. an array of 12685 floats representing the battlefield's current state. The output is a single integer in the \[0, 2312\] range, representing the encoded action to perform.

The trained model is a single file with a size of ~7.4MB. As it is loaded into the VCMI memory, the UI may freeze for a moment, which is why a form of caching was implemented to re-use the already loaded model instead of reading it from disk the next time an instance of MMAI would be created (the model was made stateless and thread-safe, so multiple MMAI instances can share it).

mmai-activity

I could talk for days, but I'd like to keep it short (it's a PR after all), but you can find more info on [this](https://smanolloff.github.io/projects/vcmi-gym/) page.

## Limitations

MMAI has its limitations, some of which will be addressed in the future. For this reason, the initial versions of MMAI will be restricted to Neutral AI only.

#### No spellcasting

Neither heroes, nor creatures commanded by MMAI will cast spells. This limitation will be removed in a future release.

#### No retreating

MMAI does not retreat. This limitation will be removed in a future release.

#### Not supported on some platforms

MinGW and 32-bit Android MMAI builds are not available, because `libtorch` (the NN lib used by MMAI) does not support these platforms. They will most likely never be supported (VCMI can still be compiled, but without MMAI).

#### Depends on AI Value

The "AI Value" creature property plays a central role in model training and inference. I heard there are plans to deprecate it, so this dependency will have to be removed in a future release.

#### Behaviour correction

Sometimes MMAI will make questionable decisions. Patterns I have noticed while experimenting with the current model:
* weak to shooters: sometimes it's too afraid to get in full shooting range
* weak to AoE spells: it tends to re-group in a very tight formation (it was never trained against spellcaster heroes)
* unaware of blinded/paralyzed enemies (it has no exposure to that property), so it will perceive them as a regular (non-blinded) unit.
* afraid of strong armies (rare): if the enemy is too strong, it stops attacking (_don't we do the same, though?_)
* confused in less-typical situations: e.g. if troops have switched sides, it makes weird actions as such situations rarely occur during training. But it happens, especially in long battles where both sides have moved to the opposide side of the battlefield.

Unwanted behaviour can be corrected through targeted training, but great care must be taken to avoid [catastrophic forgetting](https://en.wikipedia.org/wiki/Catastrophic_interference). This will be an ongoing effort for the forseeable future. Hopefully new, better models will eventually replace the existing ones.

## Future plans

Here's _some_ of the stuff I plan to work on next (not necessarily in this order):

* Train an attacker model (up next).
* Fix any VCMI issues caused by MMAI.
* Add spellcasting support:
  * phase 1: add scripted spellcasting (e.g. use BattleAI's spellcasting logic).
  * phase 2: train models to cast spells.
* Confidence report: MMAI should be able to tell what are the estimated odds of winning the current battle. This can be used for retreat/surrender decisions.
* "AI Value" alternative:  Maybe the creature's resource cost can be used to measure its "importance"?
* Migrate training from local to cloud/rental hardware for faster results.
* Submit PR with changes used for the MMAI training (vcmi-ml library, ML mod).

A long-term vision concerning the future of MMAI involves introducing imitation-based learning algorithms to train from a database of recorded VCMI battles played by humans. Even further in the future - an ML-powered adventure AI. Those are just concepts and nothing particular is planned at this point.

## Similar projects

I could not find any open-source projects (any projects at all, rather) about an ML-driven HOMM3 AI. However, there have been discussions in the VCMI Discord channels about other AIs, so more ML-powered AIs could be coming to VCMI soon.

I am truly curious to see how the future will unfold. Many different AIs would mean VCMI is gaining momentum and the ecosystem is growing. One AI may be better at tasks others are weak at and I wouldn't rule out the option for a composite AI with an ensemble of "heads", each tasked with a job it does well. Hell, we could even set up an AI league ;) Still, MMAI seems to be alone in it for now.

## Conclusion

Despite some shortcomings in its current state, MMAI may prove to be a viable neutral army commander. It enhances gameplay experience by adding engaging and unpredictable behaviour of the enemy troops during battle. Through the MMAI mod, new model versions can be easily released when they finish training. I would say there's a lot of future potential for MMAI (and ML-driven AIs in gereral), this is just the beginning.

Give it a try, have some fun with it. Keep in mind MMAI has a lot to learn, so don't go too hard on it ;) Cheers 🍻


IvanSavenko (commented):
After a quick look:
- Don't see need for separate MMAI-enabled builds
- Not sure about submodules, might be better to include those parts directly into vcmi git
- Memory/CPU usage changes are probably fine, but artifact size increase is massive. 3x time compared to non-MMAI build is A LOT. Is there a way to reduce size of libtorch to more reasonable size?
- also, libtorch somehow needs to be built as part of our dependencies / our sources - to make sure that builds are reproducible and are made in clean environment.

Also, it looks like libtorch is not available in Conan. @kambala-decapitator how should we handle this?

smanolloff:
I did try to minimize the size of libtorch through compile flags. The version used here is prebuilt using [these](https://github.com/smanolloff/vcmi-libtorch-builds) scripts. I am not sure I am capable of reducing it any further, maybe someone else may have more success. In the end, I didn't believe it would be such a big deal, but I guess I was wrong 🙁

Olegmods:
Android build (arm one) is broken. There is no MMAI in the options.

Olegmods:
Screenshot_2024-10-17-11-22-22-710_is xyz vcmi daily

misiokles:
Yes, it is mentioned.

Olegmods:
> Yes, it is mentioned.

The 32-bit ones are mentioned. The 64-bit ones are not.

psyhlo:
may be you didnt download the right apk?
image

Olegmods:
> may be you didnt download the right apk? image image

Nope, it was the right one.

Olegmods:
> may be you didnt download the right apk? image image

Seems like that. Why there would be two .apk-s for arm?

psyhlo:
image

Olegmods:
Yeah downloaded another .apk and it works now.

DjWarmonger:
Can MMAI utilize CUDA at inference time?

smanolloff:
@DjWarmonger unfortunately, no - `libtorch` for VCMI is (intentionally) built without CUDA support.

CUDA could be added, but that would be impractical for a couple of reasons:
* Compiled size - libtorch with CUDA is huge (uncompressed it is over 2GB)
* We would have to make separate CUDA and CPU-only VCMI builds to support devices without NVIDIA drivers (I have not confirmed this, though - it's possible cuda-libtorch has a cpu fallback)
* The performance gains at runtime (inference) would be negligible. In the end, I don't believe it's gonna make a difference for the user experience.

CUDA makes sense for training purposes only. What you could do if you need CUDA for inference in VCMI (e.g. if you have some special dev use-cases using on a modified VCMI binary where the CUDA performance benefits would make a difference) is to compile VCMI from this branch with `-DENABLE_MMAI=1 -DLIBTORCH_PATH=/path/to/libtorch-with-cuda` where _libtorch-with-cuda_ can be compiled from source (note: it takes a *lot* of time to compile it and depends on CUDA SDK), or downloaded from e.g. https://download.pytorch.org/libtorch/cu124. However, if you decide to go that path, let me know, as I will need to make a small modification to the pre-trained model as well (it is hard-coded for cpu for now).

kambala-decapitator:
> * also, libtorch somehow needs to be built as part of our dependencies / our sources - to make sure that builds are reproducible and are made in clean environment.
> 
> Also, it looks like libtorch is not available in Conan. @kambala-decapitator how should we handle this?

for a quick solution we can incorporate those custom build scripts into our pipeline - at least they build everything from source. They'd need some love though.

Yes, currently there's no libtorch recipe, but I see there's already a working attempt: https://github.com/conan-io/conan/issues/11273#issuecomment-2063232517 With enough desire/motivation we could create proper recipe and submit it to CCI. There's even a WIP: https://github.com/conan-io/conan-center-index/pull/24759

DjWarmonger:
> CUDA makes sense for training purposes only. What you could do if you need CUDA for inference in VCMI (e.g. if you have some special dev use-cases using on a modified VCMI binary where the CUDA performance benefits would make a difference) is to compile VCMI from this branch with -DENABLE_MMAI=1 -DMMAI_LIBTORCH_PATH=/path/to/libtorch-with-cuda where libtorch-with-cuda can be compiled from source (note: it takes a lot of time to compile it and depends on CUDA SDK), or downloaded from e.g. https://download.pytorch.org/libtorch/cu124. However, if you decide to go that path, let me know, as I will need to make a small modification to the pre-trained model as well (it is hard-coded for cpu right now).

Actually I do have RTX 4080 and have been doing ML experiments on it last year.

DjWarmonger:
With recent optimizations in battle AI (bonus system, but not only) it should be possible to run battles **much** faster.

Not sure if you need more training data at this point, but this might help with extra AI features in future scenarios.

GeorgeK1ng:
Any chance to target this into develop and get compiled artifacts for testing please?

smanolloff:
@GeorgeK1ng I am currently preparing the builds, targeting develop. I will hopefully have them ready in a few days.

smanolloff:
# MMAI: 2025 update

I present to you an improved version of MMAI. It addresses issues outlined in the initial version and brings some new functionality. 

**Is it better than the 2024 version?** Yes.

**Is it stronger than BattleAI now?** Yes. MMAI defeats BattleAI in ~65% of battles now.

**Does it still wait for players to attack?** No, such behaviour was rated _annoying_ and _not fun_ to play against. The new models are more aggressive and will rarely stay back in battle.

**What else is new in MMAI 2025?**

* delegates spell casting to BattleAI.
* respects surrender/retreat decisions made by the adventure interface.
* includes an attacker model (instead of only defender).
* can now be chosen as any battle AI (instead of just Neutral).
* no longer relies on the "AI Value" creature property.
* trained against BattleAI  (thanks to the VCMI optimizations pointed out by @DjWarmonger).
* no longer greedy and may opt for different actions in identical states (see temperature below)
* has a `temperature` parameter for altering the stochastic sampling of actions:
  * `0.0` = greedy (deterministic). 
  * `1.0` = baseline value used during training
  * `2.0` = uncertain/hesitant. Occasionally chooses less-meaningful actions.
  * `3+` = garbage (setting it too high essentially makes it an RNG).
  *  value is set in `Mods/MMAI/config/mmai-settings.json`.
* has a `seed` parameter which makes the model decisions reproducible even when temperature > 0.
 
**What limitations remain?**
  * Still trained on H3 SOD units only. However, I believe model has generalised well and should perform adequately with units it has never seen (e.g. from HoTA or other expansions).
  * Castle siege - the models know how to play siege, but are not particularly good at it. Siege battles constituted only 10% of the total training they had and will occasionally act weird. This is not a big issue, siege-trained models will follow soon.
  * Artifact size - the increase in artifact size remains, and is in fact bigger now (the new models are bigger, plus there's two of them bundled in each artifact). I believe this can be partly addressed by excluding the MMAI models from the installation package and asking the user to download the mod via the launcher. However, a certain fixed increase in size _cannot be avoided_ due to the libtorch dependency (<30MB for most platforms, 70MB for mac-intel).
  * Not available on some platforms - 32-bit platforms are not supported (MMAI is not visible in the launcher's drop-down menu for those builds).
  * Possible incompatibility with older OS versions - e.g. older Androids. It does work on Android 13 (where I tested it), but that is a 3-year old phone. If VCMI aims to support old devices (5-10 years old), that might be a problem. Same goes for other platforms. Introducing separate builds for older OS-es (without MMAI) is one potential workaround to this.

**Model information**
The new models use a neural network with 3-layer heterogenous graph convolutional neural network with a total of ~11M parameters. The network architecture is tailored to the VCMI battlefield with 165 nodes representing the hexes and features 7 different edge types representing various relations between them such as adjacency, reachability, turn order and more. Everything is kept strictly within the observational capabilities of the player interface, no cheating. The observable state has been expanded to a size of 120KB, consisting of floating-point numbers representing various global, hex and stack attributes, absolute and relative, encoded as normalized scalars or one-hot categorical values. The number of graph node edges is not included here and adds an additional 20-50KB based on how "crowded" the battlefield is. That's only the tip of the iceberg, as I went through a variety of frameworks, RL algorithms, NN architectures in the process, went from local to rental GPU hardware (which cost me more than a "dime"...), began experimenting with self-training, and there's much more to talk about, but I haven't had the time to update my blog yet :) So there is no "for more info click here" link, not yet.

**Conclusion**

Grab a [build](https://github.com/vcmi/vcmi/actions?query=actor%3Asmanolloff), enable the MMAI mod, select MMAI as the Neutral AI in the launcher and give it a try.
Tweak the temperature parameter, see how it feels.
As usual, have fun and know that MMAI _will_ make mistakes, so don't set your expectations too high ;)
Cheers 🍻

thehonestcto:
Bravo!!! Exciting!!! 2026 is going to be a very interesting year for VCMI

Laserlicht:
Wouldn't it better to use onnx and onnxruntime instead of libtorch for inferencing and sharing models?

Onnx feels more universal than torch stuff. And we could also use conan package for onnxruntime.

smanolloff:
Thank you for the suggestion, it does make sense. Not sure if it will work, but will try it.

smanolloff:
@Laserlicht I have implemented the changes you suggested. The MMAI models are now exported as ONNX graphs and the VCMI code is updated to handle the new format. I have also opened a PR in the vcmi-dependencies repo, please see the corresponding PR: https://github.com/vcmi/vcmi-dependencies/pull/19 

I have tested with VCMI (develop branch) on MacOS 15 (arm), Windows 10 (x64), iOS 18 and Android 13 without noticing any issue. In fact, the  ONNX model performance exceeds my expectations as the runtime predictions is 30% faster. The size footprint of the VCMI build artifact is also smaller.

PS. the android-64 build here has failed with error "No space left on device" and I believe a restart of the build would resolve that, but I have no permissions to restart builds. It does otherwise build successfully.

Laserlicht:
Nice. 30% faster is great.

Some things I've noticed:
- CI does not create installer for Win7
- MMAI mod is disabled by default (leading to crash on battle -> maybe show info window and use fallback ai in case onnx is missing)

In general I would suggest:
- Adding mod to our mod repository, launcher should enforce download, when selecting MMAI -> Reduces size and you can update models independently from VCMI build
- Using our normal file handler for onnx (registering new type), see: lib/filesystem/ResourcePath.cpp
- (before merge) your mmai submodule should forked under vcmi org (i think we could give you write permissions to that)

Not sure what else is missing to bring this into a mergable state. @IvanSavenko what do you think?

kambala-decapitator:
I haven't checked this PR, but will there an option to build VCMI without ONNX library?

> I have also opened a PR in the vcmi-dependencies repo, please see the corresponding PR: [vcmi/vcmi-dependencies#19](https://github.com/vcmi/vcmi-dependencies/pull/19)

will review

IvanSavenko:
- Not sure about having new AI as submodule. We generally use them only for 3rd-party stuff. Right now making any change to AI interface or to our library classes used by new AI will require a lot of busywork - possibly clone submodule, create new branch in submodule, create separate commit in submodule, push submodule, update submodule in main repo, push main repo, finally open 2 PR's instead of 1. 

- from what I see, AI is still unable to use non-standard starting positions, like fighting in creature banks (whether as defender or as attacker). And some banks especially in mods like hota have slightly different starting unit layout. If AI performs poorly in such combats - should be fixed, at least via fallback to BattleAI.

- Mod should be downloadable via launcher from separate repo, not a submodule. Selecting new AI should propose download of your mod, and keep old AI if player declines. Also needs runtime check for enabled mod (or more correctly - for existing model file via our virtual filesystem) with fallback to BattleAI - if player starts game with new AI but without mod active.

- Still not a huge fan of large increase in download size, but considering positive reaction from GrafDeVan, not against it.

> I haven't checked this PR, but will there an option to build VCMI without ONNX library?

Both CMake and conan have option to enable or disable new AI

> PS. the android-64 build here has failed with error "No space left on device" and I believe a restart of the build would resolve that, but I have no permissions to restart builds. It does otherwise build successfully.

You can restart *all* builds by modifying last commit & force-push:
```
git commit --amend
git push --force repo branch
```

smanolloff:
Thanks everyone for the prompt feedback.

> @Laserlicht
> CI does not create installer for Win7

👍 will add that

> @Laserlicht
> MMAI mod is disabled by default (leading to crash on battle

Yes, there is the problem that the launcher allows to select MMAI even when the mod is disabled. The fallback fails in this case as the mod's settings are unavailable (where "fallback" is looked up). I will hard-code a BattleAI fallback there.

> @Laserlicht
> Adding mod to our mod repository, launcher should enforce download, when selecting MMAI -> Reduces size and you can update models independently from VCMI build

I baked the models into VCMI for easier testing, but now I am working on making them downloadable. Indeed, it would greatly reduce the build artifact size.

> @Laserlicht
> Your mmai submodule should forked under vcmi org

Given what @IvanSavenko commented, I am not sure how to proceed here.

> @Laserlicht
> Using our normal file handler for onnx (registering new type), see: lib/filesystem/ResourcePath.cpp

👍 will add that

> @kambala-decapitator
> will there an option to build VCMI without ONNX library?

Yes, you can either:
-  add `-DENABLE_MMAI=OFF` compile flag, which excludes MMAI and, by extension - onnxruntime from VCMI
- also speed up conan install by excluding onnxruntime: `conan install .... -o "&:with_onnxruntime=False"`

> @IvanSavenko
> Not sure about having new AI as submodule. We generally use them only for 3rd-party stuff

I can make it part of main VCMI source tree. But given what @Lasersicht suggested, I am not sure how to proceed here.

> @IvanSavenko
> from what I see, AI is still unable to use non-standard starting positions

Did you test it? I believe MMAI should be able to handle those battles similarly to BattleAI or better. Adding a fallback for such layouts is easy, but let's first see if it's needed at all.

> @IvanSavenko
> Mod should be downloadable via launcher from separate repo

👍 I will work on that.

> @IvanSavenko
Selecting new AI should propose download of your mod, and keep old AI if player declines

That might take more time to implement. Can we agree on a simpler version for the initial rollout, e.g. via informative text (e.g. label can be "MMAI (requires MMAI mod)") -- selectable, but will always fallback to BattleAI if mod is not enabled?

> @IvanSavenko
Also needs runtime check for enabled mod (or more correctly - for existing model file via our virtual filesystem) with fallback to BattleAI

There is already such a check and a (configurable) fallback. I will add a hard-coded BattleAI fallback in case MMAI mod is disabled (currently it's broken).

> @IvanSavenko
> Still not a huge fan of large increase in download size

I have good news for you - the model files account for the huge increase. They are part of the MMAI mod and once we make it downloadable, the size increase becomes relatively small:

| Build                 | Old size | New size  |
|-----------------------|----------|-----------|
| ios                   | 19.4 MB  | 25.3 MB   |
| mac-arm               | 25.9 MB  | 32.1 MB   |
| msvc-arm64            | 32 MB    | 41 MB     |
| android-64            | 26.5 MB  | 34.1 MB   |
| android-32            | 23.9 MB  | 30.6 MB   |
| mac-intel             | 27.1 MB  | 34 MB     |
| android-64-intel      | 28 MB    | 35.6 MB   |
| msvc-x64              | 34.6 MB  | 44.5 MB   |
| msvc-x86              | 30.1 MB  | 38.9 MB   |


---
As a final comment, I would like to add that migrating MMAI to ONNX improved [platform compatibility](https://onnxruntime.ai/docs/reference/compatibility.html#environment-compatibility), eliminating earlier concerns regarding support on older systems.

IvanSavenko:
> I can make it part of main VCMI source tree. But given what @Lasersicht suggested, I am not sure how to proceed here.

Let's go with part of main vcmi source tree, without submodule

> That might take more time to implement. Can we agree on a simpler version for the initial rollout, e.g. via informative text (e.g. label can be "MMAI (requires MMAI mod)") -- selectable, but will always fallback to BattleAI if mod is not enabled?

Not against it, but we already have similar logic (but unconditional) in settings menu - translations. Check `CSettingsView::on_pushButtonTranslation_clicked`:
```cpp
	if (mainWindow->getModView()->isModAvailable(modName))
	{
		mainWindow->switchToModsTab();
		mainWindow->getModView()->doInstallMod(modName);
	}
```
(modName here is name of directory with your mod, `MMAI`, or possibly - lower-case version, so `mmai`, and obviously needs mod to be present in our mod repo)

> I have good news for you - the model files account for the huge increase. They are part of the MMAI mod and once we make it downloadable, the size increase becomes relatively small:

Great :)

kambala-decapitator:
> migrating MMAI to ONNX improved platform compatibility, eliminating earlier concerns regarding support on older systems

That page says that the lib is tested with ios 12 / macos 10.14, and may be compatible with 10.12+, but your pr raises these requirements to 13 and 10.15 respectively. Are the docs outdated perhaps? Is the raise related to std::filesystem usage?

smanolloff:
> > migrating MMAI to ONNX improved platform compatibility, eliminating earlier concerns regarding support on older systems
> 
> That page says that the lib is tested with ios 12 / macos 10.14, and may be compatible with 10.12+, but your pr raises these requirements to 13 and 10.15 respectively. Are the docs outdated perhaps? Is the raise related to std::filesystem usage?

Seems so, yes. The compile error I encountered is about std::filesystem calls -- the relevant code is introduced in a recent release of onnx 1.16.0 (1.16.1 being the latest). I will try to patch the change to see if it will allow to keep the OS requirements lower.

EDIT: well, not quite the _latest_ release - the latest onnx is 1.18. I guess the discrepancy comes from the fact that onnxruntime (which hosts this documentation) and onnx are separate projects.

kambala-decapitator:
> I will try to patch the change to see if it will allow to keep the OS requirements lower.

would be great, but only if the patch is relatively small, otherwise I don't think it's worth it

smanolloff:
> > I will try to patch the change to see if it will allow to keep the OS requirements lower.
> 
> would be great, but only if the patch is relatively small, otherwise I don't think it's worth it

Please see my [comment](https://github.com/vcmi/vcmi-dependencies/pull/19#issuecomment-3573308548) in the other PR. _tl; dr_ for macos/ios it worked, for win7 it did not

smanolloff:
Functionality-wise, the implementation is done. I will now spend some time on addressing the codestyle checks.

AI/MMAI/BAI/model/NNModel.cpp
```diff
@@ -0,0 +1,868 @@
...
+ * NNModel.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#include <array>
```
Laserlicht: Any reasons for adding these includes here?
Most (or maybe all) of them should come from `Global.h` (-> included by `StdInc.h` which you already added)

AI/MMAI/BAI/base.cpp
```diff
@@ -0,0 +1,264 @@
+/*
```
Laserlicht, line 1: Especially for all new files:
Would be nice to follow our [Coding guidelines](https://vcmi.eu/developers/Coding_Guidelines/).

Especially noticed:
- we use tab as indent
- opening braces (e.g. after `if(...)`) should be in a new line

Should be possible to auto correct... We have some clang-tidy stuff for formatting in here. But I don't have experience with it. :(

AI/MMAI/CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+    set_target_properties(MMAI PROPERTIES INSTALL_RPATH "@loader_path")
+  elseif(UNIX)
+    set_target_properties(MMAI PROPERTIES INSTALL_RPATH "$ORIGIN")
+  endif()
+endif()
+
+target_link_libraries(MMAI PRIVATE vcmi)
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+  set(ONNXRUNTIME_ROOT "/opt/onnxruntime" CACHE PATH "Path to onnxruntime containing lib/onnxruntime.so and include/onnxruntime_cxx_api.h")
```
Laserlicht, line 80: Maybe add check if onnxruntime is available in system and only if not use it from downloaded file.

There are already some distributions with packages, e.g.:
- [Debian 13](https://packages.debian.org/trixie/libonnxruntime-dev)
- [Ubuntu 24.10](https://launchpad.net/ubuntu/oracular/+package/libonnxruntime-dev) (should be in next [LTS 26.04](https://launchpad.net/ubuntu/resolute/+package/libonnxruntime-dev) as well)
- [Arch](https://archlinux.org/packages/extra/x86_64/onnxruntime/)
- [Fedora](https://packages.fedoraproject.org/pkgs/onnxruntime/onnxruntime-devel/)

AI/MMAI/BAI/model/NNModel.cpp
```diff
@@ -0,0 +1,868 @@
...
+        seed = std::chrono::high_resolution_clock::now().time_since_epoch().count();
+        logAi->info("Seed is 0, using %1%", seed);
+    }
+    rng = std::mt19937(seed);
+
+    auto opts = Ort::SessionOptions();
+    opts.SetIntraOpNumThreads(4);
+    opts.SetGraphOptimizationLevel(GraphOptimizationLevel::ORT_ENABLE_BASIC);
+
+    model = std::make_unique<Ort::Session>(ort_env(), ToOrtPath(path).c_str(), opts);
```
Laserlicht: As already mentioned you should use VCMI Resource Manager to load ONNX file:
- `lib/filesystem/ResourcePath.cpp` add e.g. `{".ONNX",   EResType::AI_MODEL}` (header needs also to be edited to add `EResType`)
- load file to byte array: `auto data = CResourceHandler::get()->load(ResourcePath("DATA/ABCDEF.ONNX", EResType::AI_MODEL))->readAll();`, you can use `CResourceHandler::get()->existsResource` to check if res exists
- Use overload of Ort::Session to load from byte array: https://onnxruntime.ai/docs/api/c/struct_ort_1_1_session.html#ac657d96c0ea760e0f8712f23a28aaa08

smanolloff: Thanks, will do that.

AI/MMAI/BAI/base.cpp
smanolloff: Sure, I already ran it on all new files (in newer commits)

AI/MMAI/BAI/model/NNModel.cpp
smanolloff: These get auto-added by my LSP-clangd plugin, I will remove them

AI/MMAI/BAI/model/NNModel.cpp
smanolloff: If I may ask you one additional question: is it possible to use a JSON schema validator for validating a settings file that comes from a mod like MMAI? Currently I have this resource `MMAI/CONFIG/mmai-settings.json` and I am manually validating all configuration values, because I couldn't figure out how to use the already existing JsonValidator.

IvanSavenko: For example: https://github.com/vcmi/vcmi/blob/develop/lobby%2FLobbyServer.cpp#L370

There is `JsonUtils::validate` convenience method. In this example, data that failed to pass verification is discarded, but generally vcmi attempts to continue anyway - validator always prints all errors to log & console for modders/devs to see

Edit: more straightforward example - validate using schema from file & continue. 
https://github.com/vcmi/vcmi/blob/develop/lib%2FGameSettings.cpp#L135

GeorgeK1ng: @MichalZr6 Could probably help with validator invrstigation

MichalZr6: You may give a try to the script I wrote some time ago: https://github.com/MichalZr6/vcmi-mod-validator

smanolloff:
Summary of changes:

- MMAI is supported on all builds, no new installers or excluded platforms.
- The `onnxruntime` dependency is added with [this PR](https://github.com/vcmi/vcmi-dependencies/pull/19).
  When it gets merged, URLs in [`install_conan_dependencies.sh`](https://github.com/vcmi/vcmi/blob/5b35006370/CI/install_conan_dependencies.sh#L5-L7) and [`.gitmodules`](https://github.com/vcmi/vcmi/blob/5b35006370/.gitmodules#L15) can be updated accordingly.
- build artifacts are ~21% larger in size (+ 5..10MB)
- MMAI git submodules are removed. All code is added to the main VCMI source tree with VCMI licensing headers.
- the `mmai` [mod](https://github.com/vcmi-mods/mmai) is now part of VCMI org and is downloadable via the launcher. Selecting the "MMAI (experimental)" option triggers automatic download.
- MMAI loading errors (e.g. mod inactive; or models are not found) handled with fallback to `BattleAI` with error written to logs (no UI notification)
- `.onnx` file extension (models downloaded via the mod) registered in the VCMI filesystem
- mod configuration is validaded via JsonValidator [schema](https://github.com/vcmi/vcmi/blob/5b35006370/config/schemas/mmaiSettings.json)
- all code is formatted with `clang-format` and (almost all) sonarcloud issues are addressed (the remaining I don't know how to fix or are false positives)
- Installation [docs](https://github.com/vcmi/vcmi/blob/5b35006370/docs/developers/Building_Linux.md#on-debian-based-systems-eg-ubuntu) are updated

Looking forward to your feedback.

lib/callback/CDynLibHandler.cpp
```diff
@@ -59,7 +62,7 @@ VCMI_LIB_NAMESPACE_BEGIN
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wcast-function-type"
 #endif
-	HMODULE dll = LoadLibraryW(libpath.c_str());
+	HMODULE dll = LoadLibraryExW(libpath.c_str(), NULL, LOAD_WITH_ALTERED_SEARCH_PATH);
```
IvanSavenko, line 65: Not against it, but why was this change necessary?

smanolloff: To allow MMAI to link with BattleAI (needed for spellcasting).
Without this change, there is an error on windows because BattleAI.lib is in the `AI/` subdirectory, which is not searched by LoadLibraryW. The LOAD_WITH_ALTERED_SEARCH_PATH says "also look in the current dir" (which for MMAI is `AI/` dir). On other OS this can be handled in the [CMakeLists.txt](https://github.com/vcmi/vcmi/blob/5b3500637015a1d40ae08da0cb332f05b7eee00c/AI/MMAI/CMakeLists.txt#L69-L75). If you think there's a better way to do this, please let me know.

CI/before_install/linux_qt5.sh
```diff
@@ -1,8 +1,15 @@
...
-set -euo pipefail
+set -euxo pipefail
 export DEBIAN_FRONTEND=noninteractive
 
+# https://github.com/microsoft/onnxruntime/discussions/6489
+ONNXRUNTIME_URL=https://github.com/microsoft/onnxruntime/releases/download/v1.18.1/onnxruntime-linux-x64-1.18.1.tgz
+ONNXRUNTIME_ROOT=/opt/onnxruntime
+sudo mkdir -p "$ONNXRUNTIME_ROOT"
+curl -fsSL "$ONNXRUNTIME_URL" | sudo tar -xzv --strip-components=1 -C "$ONNXRUNTIME_ROOT"
+find "$ONNXRUNTIME_ROOT"  # print file tree
```
kambala-decapitator: maybe extract this part to a shared file? also this comment is different in the two files

smanolloff: Sure, I will extract these 4 lines in a separate file, please let me know what is the specific filename you would like to see.

kambala-decapitator: `linux_common.sh` should be good I think

smanolloff:
URLs in `install_conan_dependencies.sh` and `.gitmodules` [updated](https://github.com/vcmi/vcmi/commit/909f86e345e22de633b66f0e0413d77d2a138a82) back to vcmi's dependencies as `onnxruntime` is now included there. Please let me know if there's anything else you would like me to change.

CI/install_conan_dependencies.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-RELEASE_TAG="2025-11-06"
+set -x
```
kambala-decapitator: don't think this is needed

AI/MMAI/CMakeLists.txt
```diff
@@ -0,0 +1,137 @@
...
+  if(APPLE)
+    set_target_properties(MMAI PROPERTIES INSTALL_RPATH "@loader_path")
+  elseif(UNIX)
+    set_target_properties(MMAI PROPERTIES INSTALL_RPATH "$ORIGIN")
+  endif()
+endif()
+
+target_link_libraries(MMAI PRIVATE vcmi)
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
kambala-decapitator, line 79: you could also use `WIN32 OR APPLE` condition and swap if/else branches

smanolloff: It would break Android builds I believe.

kambala-decapitator: `OR ANDROID` :)

smanolloff:
@kambala-decapitator I added a "Free up disk space" step in CI in order to overcome the flaky Android builds (they often fail with "No space left on device" error). This step frees up ~10GB of space by removing unused tools installed on the CI runner, such as .NET, Haskell compiler, etc.

AI/MMAI/BAI/v13/stack.cpp
```diff
@@ -0,0 +1,543 @@
...
+
+	int cid = cstack->creatureId().num;
+	if(cid > Schema::V13::CREATURE_ID_MAX)
+	{
+		// This would trigger every time a non-vanilla H3 creature becomes active for MMAI
+		// logAi->debug("MMAI was not trained with this creature: id=%d (%s)", cid, cstack->getDescription());
+#ifdef ENABLE_ML
+		throw std::runtime_error("unknown creature id: " + std::to_string(cid));
+#endif
+		cid = 122; // this is a "NOT USED (1)" creature
```
IvanSavenko: FYI - this would trigger for large number of games, since a lot of players use some mods, at the very least - wog or hota.

Ideally AI should operate solely on traits (stats/bonuses) of units and not on unit ID's

AI/MMAI/BAI/v13/stack.cpp
```diff
@@ -0,0 +1,543 @@
...
+			case BonusType::SPELL_LIKE_ATTACK:
+				switch(bonus->subtype.as<SpellID>())
+				{
+					case SpellID::DEATH_CLOUD:
+						d += 0.2;
+				}
+				break;
+			case BonusType::SPELL_AFTER_ATTACK:
+				switch(bonus->subtype.as<SpellID>())
+				{
```
IvanSavenko: Would have been great to somehow cover units from mods, at least with some default value

AI/MMAI/BAI/v13/stack.cpp
```diff
@@ -0,0 +1,543 @@
...
+			return slot;
+		}
+		else if(slot == SlotID::WAR_MACHINES_SLOT)
+		{
+			return S13::STACK_SLOT_WARMACHINES;
+		}
+		else
+		{
+			// "special" slot, e.g. summoned, commander, etc.
+			return S13::STACK_SLOT_SPECIAL;
```
IvanSavenko, line 48: By the way - it might be a good idea for AI to be aware of summons, and deliberately use them as meat shields (e.g. not including in received damage/losses) - we don't care what happens to summoned elementals, but we want to keep our permanent army alive for next battle

AI/MMAI/BAI/base.cpp
```diff
@@ -0,0 +1,292 @@
...
+}
+
+/*
+ * These methods MUST be overridden by derived BAI (e.g. BAI::V1)
+ * Their base implementation is is for logging purposes only.
+ */
+
+void Base::activeStack(const BattleID & bid, const CStack * astack)
+{
+	debug("*** activeStack ***");
```
IvanSavenko, line 60: Will need to check how it works in practice, but a lot of logging can significantly slow down AI turn times

AI/MMAI/BAI/model/NNModel.cpp
```diff
@@ -0,0 +1,585 @@
...
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#include <onnxruntime_c_api.h>
+#include <onnxruntime_cxx_api.h>
+
+#include <algorithm>
+
+#include "StdInc.h"
```
IvanSavenko, line 11: Please make sure that StdInc.h is the very first include in file

AI/MMAI/BAI/v13/stack.cpp
```diff
@@ -0,0 +1,543 @@
...
+// static
+int Stack::CalcValue(const CCreature * cr)
+{
+	static const bool verbose = []()
+	{
+		const char * envvar = std::getenv("MMAI_VERBOSE");
+		return envvar != nullptr && std::strcmp(envvar, "1") == 0;
+	}();
+
+	static std::map<const CCreature *, float> cache;
```
IvanSavenko: Any way to avoid non-const static variable? 

Perhaps compute score for each in-game unit on AI initialization and store it as part of AI itself?
This would ensure that value is guaranteed to exists and will also allow dropping mutex guards

AI/MMAI/BAI/v13/stack.cpp
smanolloff: This is for informational purposes only, AI does not use it. ENABLE_ML flag is always disabled here, so this entire code block can be removed.

AI/MMAI/BAI/v13/stack.cpp
smanolloff: You are correct, this list is far from exhaustive. It only targets vanilla H3 effects, and many of those are omitted as well (ones I considered less impactful). All the rest will be ignored, which is sub-optimal, but not critical for now. 

Regarding the default value you suggested - I am not sure if it would add any value to the model. An "unknown bonus" flag is one experiment I might try at some point.

My future improvement idea is to add most available bonuses from vanilla H3 + popular mods (the ones you mentioned - hota, wog),  then train a new models (v14 and later) with those mods enabled. There are also two ways to go here - 1/ using the current graph topology where each hex contains stack information (filled with zeros if not occupied), effectively expanding the observation with 165*X additional fields with X being the number of added flags, but requiring less effort; and 2/ using a bipartite graph where nodes and stacks are separate nodes, which would significantly reduce the observation size at the cost of more NN complexity and implementation effort. I plan to try both in the future, depending on my free time.

AI/MMAI/BAI/v13/stack.cpp
smanolloff: Fully agree. A "value dampening" factor for summons might help here and might even work for the existing models.

AI/MMAI/BAI/base.cpp
smanolloff: I would like to have those around when needed, so I can suggest hiding them behind an envvar flag (`MMAI_VERBOSE`, set once at battle start or even at library load). What do you think?

AI/MMAI/BAI/v13/stack.cpp
smanolloff: If there's a way to precompute score for each unit once and store it in a const static var, that would be great.
Can you help me with an approach to iterate through all in-game units? I was unable figure it out :(

Laserlicht: See here for working with creatures: https://github.com/vcmi/vcmi/blob/b6c19f4895392e0440ead08386a34c04dea43ae1/client/lobby/BattleOnlyModeTab.cpp#L561

If you want all (not only DefaultAllowed): Iterate over `LIBRARY->creh->objects`.

Does this help?

smanolloff: I will try this, thanks

AI/MMAI/BAI/v13/stack.cpp
smanolloff: Removed

AI/MMAI/BAI/v13/stack.cpp
smanolloff: Added a multiply factor of `0.2` for summons to make them much less valuable for the AI

AI/MMAI/BAI/model/NNModel.cpp
smanolloff: Done

AI/MMAI/BAI/v13/stack.cpp
smanolloff: Done 👍

AI/MMAI/BAI/base.cpp
IvanSavenko: NKAI solves this by having compile-time define. If envvar is more convenient to you - sure

IvanSavenko (approved):

AI/MMAI/BAI/base.cpp
smanolloff: On second thought, I would like to keep them for now - they would be helpful for investigating reported errors. Let's remove them later, after we have confirmed that MMAI is stable. Is that OK?

PS. They currently result in 3...10 DEBUG logs per stack turn, I believe that shouldn't be a problem anyway.

IvanSavenko: 3-10 logs per stack turn is not much if we're talking about human-vs-ai

But during AI turns there may be multiple battles between AI's or between AI and neutrals. And we want all AI turns to resolve in like 1-2 seconds. 

In any case - OK, let's merge MMAI and see how it goes.

PR #4808, opened 2024-10-20
Bonus: prism breath

Laserlicht:
Requested by @by003 for modding

[Peek 2024-10-20 21-57.webm](https://github.com/user-attachments/assets/f296e2c9-6e7d-4369-832a-d546113bbe52)

### DIFF
--- a/Mods/vcmi/config/vcmi/english.json
+++ b/Mods/vcmi/config/vcmi/english.json
@@ -687,5 +687,7 @@
 	"core.bonus.DISINTEGRATE.name": "Disintegrate",
 	"core.bonus.DISINTEGRATE.description": "No corpse remains after death",
 	"core.bonus.INVINCIBLE.name": "Invincible",
-	"core.bonus.INVINCIBLE.description": "Cannot be affected by anything"
+	"core.bonus.INVINCIBLE.description": "Cannot be affected by anything",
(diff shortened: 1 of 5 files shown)


Alucard648:
What about expanding this bonus to allow arbitrary attack patterns?

Laserlicht:
> What about expanding this bonus to allow arbitrary attack patterns?

Not sure how we could configure this... Of course we could some binary translation to a single number. But that wouldn't be self explaining...

Alucard648:
Apparently, a table consisting of hex positions (relative to attack spot) and damage modifiers for each spot, ala H5.

IvanSavenko:
Not sure if we need yet another type of multi-tile attack. We have what, 3 of them already? Cerberi, Dragons, and one from wog.

Perhaps we can encode pattern in bonus addinfo? And convert existing bonuses to such form.

We have EDir enum which specifies one of 6 possible direction. So we can specify addInfo as list of trasformations.
For example, Dragon Breath = clone primary attack hex in right / east direction (with obvious adjustment if attack comes from another direction). Cerberi multi-head attack = clone primary hex to top-left, clone primary hex to bottom-left, etc.

Laserlicht:
Yes. Should be possible to combine:
- TWO_HEX_ATTACK_BREATH
- WIDE_BREATH
- PRISM_HEX_ATTACK_BREATH

However `THREE_HEADED_ATTACK` needs another config. It doesn't attack own units...

PR #4812, opened 2024-10-21
[Hota] SeafaringAcademy. Initial PR

SoundSSGood:
While working on the implementation of the seafaringAcademy from Hota, I decided to split PR into several. This is the first one.

- Fixed assert
- CArtPlace now contains ArtifactID instead of artifact instance.
- CSecSkillPlace added. Similar to CArtPlace. For now both are placed in CComponentHolder. I'll decide later where it would be best to place it. I also plan to add other places here.
- various minor fixes

### DIFF
--- a/client/widgets/CArtifactsOfHeroAltar.cpp
+++ b/client/widgets/CArtifactsOfHeroAltar.cpp
@@ -22,6 +22,8 @@
 CArtifactsOfHeroAltar::CArtifactsOfHeroAltar(const Point & position)
 {
 	init(position, std::bind(&CArtifactsOfHeroBase::scrollBackpack, this, _1));
+	setClickPrassedArtPlacesCallback(std::bind(&CArtifactsOfHeroBase::clickPrassedArtPlace, this, _1, _2));
+	setShowPopupArtPlacesCallback(std::bind(&CArtifactsOfHeroBase::showPopupArtPlace, this, _1, _2));
 	enableGesture();
 	// The backpack is in the altar window above and to the right
 	for(auto & slot : backpack)
--- a/client/widgets/CArtifactsOfHeroAltar.cpp
+++ b/client/widgets/CArtifactsOfHeroAltar.cpp
@@ -22,7 +22,7 @@
 CArtifactsOfHeroAltar::CArtifactsOfHeroAltar(const Point & position)
 {
 	init(position, std::bind(&CArtifactsOfHeroBase::scrollBackpack, this, _1));
-	setClickPrassedArtPlacesCallback(std::bind(&CArtifactsOfHeroBase::clickPrassedArtPlace, this, _1, _2));
+	setClickPressedArtPlacesCallback(std::bind(&CArtifactsOfHeroBase::clickPressedArtPlace, this, _1, _2));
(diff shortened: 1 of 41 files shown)


lib/CSkillHandler.cpp
```diff
@@ -45,7 +45,7 @@ int32_t CSkill::getIndex() const
 
 int32_t CSkill::getIconIndex() const
 {
-	return getIndex(); //TODO: actual value with skill level
+	return getIndex() * 3 + 3; //TODO: actual value with skill level
```
SoundSSGood, line 48: Tried to fix, but still doesn't return index depending on level.

client/widgets/CArtifactsOfHeroAltar.cpp
```diff
@@ -22,6 +22,8 @@
 CArtifactsOfHeroAltar::CArtifactsOfHeroAltar(const Point & position)
 {
 	init(position, std::bind(&CArtifactsOfHeroBase::scrollBackpack, this, _1));
+	setClickPrassedArtPlacesCallback(std::bind(&CArtifactsOfHeroBase::clickPrassedArtPlace, this, _1, _2));
```
IvanSavenko, line 25: You might want to do search and replace Prassed -> Pressed
```suggestion
	setClickPressedArtPlacesCallback(std::bind(&CArtifactsOfHeroBase::clickPressedArtPlace, this, _1, _2));
```

lib/CSkillHandler.cpp
IvanSavenko: One option would be to add an override. 
Check getNameSingular / getNamePlural in CCreature.

So, CSkill will have private `getIconIndex()` override that returns... well any arbitrary-selected level (basic?)
And public getIconIndex(int skillMasterLevel) that returns icon for specified mastery.

So all code that calls Entity::getIconIndex() will still work and get some icon, and all code that is aware that this is a skill will need to specify mastery.

client/widgets/markets/TradePanels.h
```diff
@@ -28,19 +28,19 @@ class CTradeableItem : public SelectableSlot, public std::enable_shared_from_thi
 	using ClickPressedFunctor = std::function<void(const std::shared_ptr<CTradeableItem>&)>;
 
 	EType type;
-	int id;
-	const int serial;
+	int32_t id;
+	const int32_t serial;
```
IvanSavenko, line 32: Not sure if it worth the effort, but you could use `VariantIdentifier` here.
Check TradeItemSell / TradeItemBuy defines for example.

IvanSavenko:
BTW - you are aware of current (incomplete) implementation of Seafaring Academy, right? 
Although it is mostly "university with navigation + 2 skills" right now.

SoundSSGood:
> BTW - you are aware of current (incomplete) implementation of Seafaring Academy, right? Although it is mostly "university with navigation + 2 skills" right now.

Right

client/widgets/markets/TradePanels.h
SoundSSGood: I assume that CTradeableItem will not have an id field at all. It will be in CComponentHolder. But that will be later.

Issue #4816, opened 2024-10-23
save before each battle like HOMM3 launcher


nickhuang99:
**Describe your proposal**
In "HOMM3 HD launcher", there is a cool feature to allow save game before each battle so that player can load game when a battle is lost. For example, opening up a paradox box is always risky and searching ruins also often resulting in unexpected loss etc.
![launcher-with-save-before-battle](https://github.com/user-attachments/assets/0f941c27-91a9-415c-942a-0782a9ce02e4)

**Additional context**
This feature can be easily implemented with a call to "save" before each "startBattle" call in "GameHandler" class. I am already working on this feature and about to PR the patch.

GorgonUK:
@nickhuang99 Wrong project buddy

kambala-decapitator:
> @nickhuang99 Wrong project buddy

what is wrong here in your opinion?

Whalios:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

GorgonUK:
> > @nickhuang99 Wrong project buddy
> 
> what is wrong here in your opinion?

My bad, I skim read the issue description. Disregard my previous comment.

IvanSavenko:
Hmm... I think this feature can become way easier to implement if we were to rebrand it into "save before visit".

In this case it might be rather trivial to add since we won't need to determine whether there will be battle ahead of time

mrhaandi:
> "save before visit"

Some logic to save before committing to a full movement chain (which ultimately leads to a visit) would still be required.

For specifically the use case underlying this issue a quicksave adventure button would be an OK solution (would reduce the number of clicks to at most 1/3).

dydzio0614:
i think save before battle should work this way:
on double clicking path that will trigger fight - before moving create save in "pending" mode - if path is cancelled it gets deleted
if battle actually happens then "pending" save replaces last battle save on fight start

Laserlicht:
> i think save before battle should work this way: on double clicking path that will trigger fight - before moving create save in "pending" mode - if path is cancelled it gets deleted if battle actually happens then "pending" save replaces last battle save on fight start

How does HD handle edge cases like events and buildings like dragon utopia? Do they also create save before fighting?

dydzio0614:
[...]

PR #4817, opened 2024-10-23
save before each battle. The launcher UI adds 'Save before battle' by…

nickhuang99:
… replacing original 'Network port' button because insertion new button meaning to move all row index by 1. Also 'network port' seems fitting to be category of 'Network' which is near end of all buttons and it saves a lot modifications.

This is a new feature for this [save before battle](https://github.com/vcmi/vcmi/issues/4816#issue-2607282983).

### DIFF
--- a/config/schemas/settings.json
+++ b/config/schemas/settings.json
@@ -42,7 +42,8 @@
 				"savePrefix",
 				"startTurnAutosave",
 				"enableUiEnhancements",
-				"audioMuteFocus"
+				"audioMuteFocus",
+				"saveBeforeBattle"
 			],
 			"properties" : {
 				"playerName" : {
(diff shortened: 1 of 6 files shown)


IvanSavenko:
From what I see, save is done immediately before starting the battle.
This means that:
- if player attacks wandering monsters and loads save, hero will be inside monsters zone of control
- if player attacks enemy garrison, hero will be on top of garrison (and can move forwards without defeating it)
- if combat is triggered by event, then unless hero moves onto this tile again battle won't be re-triggered

And there could be other edge cases that are not accounted for. 
Such save should be made *before* moving hero.

nickhuang99:
very insightful observation and really appreciate this invaluable feedback. Now I recall original HOMM3 launcher does have trick to save before the move. i.e. hero with fly spell across the sea must be saved before crossing the sea if battle with monsters happens exactly on landing. These are complex cases. I have to dive deeper into code.

thanks again.

Sent via the Samsung Galaxy S22 5G, an AT&T 5G smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #4819, opened 2024-10-24
added MECHANICAL bonus

Laserlicht:
Requested by @by003 e.g. for factory:

> Just make a MECHANICAL bonus for creatures, separate from NON-LIVING.
> There is not any original version creatures has this bonus, only for some mods, like Factory and ToW, or other forge mods.
> MECHANICAL has the same property with NON-LIVING, but due to the repairing ability from engineer, and my new plan to set a dependency from courtyard 'reconstruction' spell, I need a MECHANICAL flag to mark them.
> In courtyard, reconstruction spell can fix creatures with 'GARGOYLE' or 'MECHANICAL'. But in engineer(lvl 2 factory units), they can only fix 'MECHANICAL', but there are no restrictions on use, except the Ring of Oblivion.

Basically just a copy of `NON-LIVING` to separate it in mods.


misiokles:
Maybe is better to aggregiate all those types in one?
For example
"type" : "NON_LVING"
"subtype" : "golem",
"subtype" : "gorgoyle",
"subtype" : "mechanical" ?

edeksumo:
> Maybe is better to aggregiate all those types in one? For example "type" : "NON_LVING" "subtype" : "golem", "subtype" : "gorgoyle", "subtype" : "mechanical" ?

I agree with misiokles. Make subtype by string. This would allow to make different types of creatures like for example: mechanical, non_living, plant, etc on the fly.

Laserlicht:
Should be possible. Is it ok to break compatibility? (mods needs to be changed for 1.6)

misiokles:
Many mods must be rewritten for 1.6 format ;)
So another change makes no diffirence.

IvanSavenko:
Perhaps this should be creature property instead?
And also introduce "living" creature type. Internally can still be a bonus, to minimize changes

> Should be possible. Is it ok to break compatibility? (mods needs to be changed for 1.6)

Some compatibility is needed, at least to avoid crashes.

by003:
Actually add a MECHANICAL bonus will not contribute to other changes.
And I think target condition can easily linked to bonus type instead of subtype.
There is not any mechanical creature in SOD, golem does not belong this category.

Alucard648:
> There is not any mechanical creature in SOD, golem does not belong this category.
War machines.

Alucard648:
> Maybe is better to aggregiate all those types in one? For example "type" : "NON_LVING" "subtype" : "golem", "subtype" : "gorgoyle", "subtype" : "mechanical" ?

I don`t think so. Gargoyles and golems are, AFAIK, can have morale, unlike HotA mechanical units. Also i think, mechanical units should be vulnerable to Aging spell from Ghost Dragons. Old machine is easier to destroy, than newly built one of the same complexity. Also Stone Skin cannot be cast on Gargoyle.

Olegmods:
> > Maybe is better to aggregiate all those types in one? For example "type" : "NON_LVING" "subtype" : "golem", "subtype" : "gorgoyle", "subtype" : "mechanical" ?
> 
> I don`t think so. Gargoyles and golems are, AFAIK, can have morale, unlike HotA mechanical units. Also i think, mechanical units should be vulnerable to Aging spell from Ghost Dragons. Old machine is easier to destroy, than newly built one of the same complexity. Also Stone Skin cannot be cast on Gargoyle.

Golems don't have morale.

DjWarmonger:
> Maybe is better to aggregiate all those types in one? For example "type" : "NON_LVING" "subtype" : "golem", "subtype" : "gorgoyle", "subtype" : "mechanical" ?

Good idea, but requires a lot of mod fixing and will likely lead to some bug reports. It's on you ;)

DjWarmonger:
> > Maybe is better to aggregiate all those types in one? For example "type" : "NON_LVING" "subtype" : "golem", "subtype" : "gorgoyle", "subtype" : "mechanical" ?
> 
> I don`t think so. Gargoyles and golems are, AFAIK, can have morale, unlike HotA mechanical units.

Gargoyles do, so morale mechanics will require update as well.

Alucard648:
> Golems don't have morale.

Good point. But does Aging work on Automatons(HotA) and Golems?

by003:
War machines are not mechanical creatures.

DjWarmonger:
> War machines are not mechanical creatures.

In H5 they are mechanical, and it's possible to repair them with Gremlin ability. I miss it in H3 :)

by003:
Mechanical creatures only exist in several mods.

HotA: Automaton, Dreadnought
ToW: Dragon Golem
and several forge relevant mods.

There isn't any difference from non-living and mechanical, I just make them can be fixed by engineer, and reconstruction spells from mod courtyard (my next plan to balance and fix)

Olegmods:
> > Golems don't have morale.
> 
> Good point. But does Aging work on Automatons(HotA) and Golems?

It doesn't, at least that's what the wiki says.

Alucard648:
> BTW: If anyone really want elementals can be resurrected, I can use the name 'reflow' to make a new spell to realize. Reflow might be a water spell.

And Repair will be definitely a fire spell.

Olegmods:
> > War machines are not mechanical creatures.
> 
> In H5 they are mechanical, and it's possible to repair them with Gremlin ability. I miss it in H3 :)

Perhaps it will be possible with one commander from one mod that is in development...

Olegmods:
> > Maybe is better to aggregiate all those types in one? For example "type" : "NON_LVING" "subtype" : "golem", "subtype" : "gorgoyle", "subtype" : "mechanical" ?
> 
> I don`t think so. Gargoyles and golems are, AFAIK, can have morale, unlike HotA mechanical units. Also i think, mechanical units should be vulnerable to Aging spell from Ghost Dragons. Old machine is easier to destroy, than newly built one of the same complexity. Also Stone Skin cannot be cast on Gargoyle.

But stone golems, stone gargoyles stoned. So I think there is no reason to find logic in a computer game. Plus some people will find these changes bad, I guess.

Alucard648:
> Perhaps it will be possible to repair war machines (and not only them) with one commander from one mod that is in development...

Kremlin?

Olegmods:
> > Perhaps it will be possible to repair war machines (and not only them) with one commander from one mod that is in development...
> 
> Kremlin?

Yeah.

config/spells/timed.json
```diff
@@ -859,6 +860,7 @@
 			"noneOf" : {
 				"bonus.MIND_IMMUNITY" : "normal",
 				"bonus.NON_LIVING" : "normal",
+				"bonus.MECHANICAL" : "normal",
 				"bonus.SIEGE_WEAPON" : "absolute",
```
IvanSavenko, line 864: Why siege weapon and mechanical get different immunity type?

Alucard648: In original H3 Implosion spell cannot be cast on catapult. Otherwise defending hero could just wipe attacker`s catapult on turn 1 (10 SP and Expert Earth are enough) and leave an opponent with heavily limited options to continue castle siege

PR #4821, opened 2024-10-25
Basic support for configurable flaggable objects

IvanSavenko:
Converted CGLighthouse class into FlaggableMapObject

Bonus provided by Lighthouse/Flaggable is now defined in config and is accessible to mods.

Additionally, it is possible to define daily income of resources, similar to mines


Alucard648:
Will it be possible to have various time intervals between resource generations (like +3000 gold once per week), like capturable weekly resource generators in H4?

IvanSavenko:
No, only daily income at the moment.

dydzio0614:
For now hota banks need to be enough, and i personally like them more (i also like dwellings not accumulating creatures - active building visit every week is more engaging).

Also hota banks got established as liked solution in the community.

Alucard648:
Some players, especially Outcast players, will strongly disagree with having to visit all those generators every week.

Issue #4833, opened 2024-10-27
Map filename with repeated extension will cause crash due to "No such file or directory"


nickhuang99:
**Describe the bug**
Placing a map file with such name pattern(i.e. "**someName.h3m.h3m**") will cause crash due to **No such file or directory**

The root cause is class "**ResourcePath**" ctor initialize "originalName" by calling "readName" during initialization which chop off extension. If later, a "resource name" is given "**ResourcePath**" again, the "**originalName**" will continue to be chopped off "extension". Then, "**originalName**" won't match "original" file path. 

In sense of variable name of "**originalName**", do we really need to chop off "extension"?  The other variable "**name**" which represent "resource" name without directory and extension already fulfill this purpose. And we should leave "**originalName**" as it is without chopping off extension. If we still want to remove directory part for uniformity, that should be ok.

**Game logs**
For a single map file name 
<code>
$ ll Maps/
total 28
drwxrwxr-x  2 nick nick  4096 Oct 28 05:43 ./
drwxrwxr-x 11 nick nick  4096 Oct 28 06:18 ../
-rw-rw-r--  1 nick nick 18120 Oct 27 05:47 M_xandallies.h3m.h3m
</code>

This is relevant backtrace of crash in gdb:

<code>
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140736093750848) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140736093750848) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140736093750848, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff4442476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff44287f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff48d836f in  () at /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff6de28c5 in std::_Optional_base_impl<boost::filesystem::path, std::_Optional_base<boost::filesystem::path, false, false> >::_M_get() (this=0x7fffacdfe360)
    at /usr/include/c++/12/optional:477
#7  0x00007ffff72ed920 in std::optional<boost::filesystem::path>::operator*() && (this=0x7fffacdfe360) at /usr/include/c++/12/optional:978
[...]

PR #4834, opened 2024-10-28
ResourcePath ctor chopoff extension of originalName when filename has…

nickhuang99:
… multiple dots.
This is to fix [bug](https://github.com/vcmi/vcmi/issues/4833) when map file name has multiple dots which causes **ResourcePath** ctor modify **originalName**. As **ResourcePath** is used whenever necessary to reconstruct from "**originalName**", this would lead to "file not found".

Keeping **originalName** unchanged seems to be original design because it allows reconstruction of **ResourcePath** from it without issue. This is the rational of this fix.

### DIFF
--- a/lib/filesystem/ResourcePath.cpp
+++ b/lib/filesystem/ResourcePath.cpp
@@ -51,19 +51,19 @@ static inline std::string readName(std::string name, bool uppercase)
 ResourcePath::ResourcePath(const std::string & name_):
 	type(readType(name_)),
 	name(readName(name_, true)),
-	originalName(readName(name_, false))
+	originalName(name_)
 {}
 
 ResourcePath::ResourcePath(const std::string & name_, EResType type_):
 	type(type_),
 	name(readName(name_, true)),
-	originalName(readName(name_, false))
+	originalName(name_)
 {}
 
 ResourcePath::ResourcePath(const JsonNode & name, EResType type):
 	type(type),
 	name(readName(name.String(), true)),
-	originalName(readName(name.String(), false))
+	originalName(name.String())
 {
 }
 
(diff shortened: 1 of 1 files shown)


IvanSavenko:
This would break at the very least `AssetGenerator::createCombatUnitNumberWindow` (would result in double extension). Other usage of this method should also be checked.

nickhuang99:
> AssetGenerator::createCombatUnitNumberWindow

Thank you for pointing out this issue. I will do more code searching and appreciate your feedback!

After some code searching, I begin to have some doubt about this ctor of **ResourcePath**:

<code>
	/// Constructs resource path based on filename and selected type. File extension is ignored
	ResourcePath(const std::string & name, EResType type);
</code>

If indeed **extension is ignored**, then ctor should **NOT** be using **readName** to chop off extension.  At least this comment would mislead developers. 

<code>
ResourcePath::ResourcePath(const std::string & name_, EResType type_):
	type(type_),
	name(readName(name_, true)),
	originalName(name_)
{}
</code>

The purpose of this ctor is , I guess, to "reconstruct" from "resource name". Since there exists a **ResourcePath** with such "resource name", can't we design a method to retrieve such **ResourcePath** by its name? i.e. _GetResourcePathByName_ which searches filesystem by matching its resource name.

nickhuang99:
I put

> This would break at the very least `AssetGenerator::createCombatUnitNumberWindow` (would result in double extension). Other usage of this method should also be checked.
> 
> `CMapInfo::mapInit` also suspicious but might be fine - it receives original name, and then computes original name from it for some reason.

I submit in another [PR](https://github.com/vcmi/vcmi/pull/4845) and close this one.

PR #4837, opened 2024-10-28
Possibility to set hero patrol radius in map editor

godric3:
No specific reason to use 60 as max, but original editor allowed to set up to 10 and I thought someone may find useful allowing higher values

### DIFF
--- a/lib/mapObjects/CGHeroInstance.h
+++ b/lib/mapObjects/CGHeroInstance.h
@@ -92,7 +92,7 @@ public:
 	static constexpr si32 UNINITIALIZED_MANA = -1;
 	static constexpr ui32 UNINITIALIZED_MOVEMENT = -1;
 	static constexpr auto UNINITIALIZED_EXPERIENCE = std::numeric_limits<TExpType>::max();
-	static constexpr int NO_PATROLLING = -1;
+	static inline constexpr int NO_PATROLLING = -1;
 
 	//std::vector<const CArtifact*> artifacts; //hero's artifacts from bag
 	//std::map<ui16, const CArtifact*> artifWorn; //map<position,artifact_id>; positions: 0 - head; 1 - shoulders; 2 - neck; 3 - right hand; 4 - left hand; 5 - torso; 6 - right ring; 7 - left ring; 8 - feet; 9 - misc1; 10 - misc2; 11 - misc3; 12 - misc4; 13 - mach1; 14 - mach2; 15 - mach3; 16 - mach4; 17 - spellbook; 18 - misc5
--- a/lib/mapObjects/CGHeroInstance.h
+++ b/lib/mapObjects/CGHeroInstance.h
@@ -100,10 +100,9 @@ public:
 
 	struct DLL_LINKAGE Patrol
 	{
-		Patrol(){patrolling=false;initialPos=int3();patrolRadius=NO_PATROLLING;};
-		bool patrolling;
+		bool patrolling{false};
 		int3 initialPos;
-		ui32 patrolRadius;
+		ui32 patrolRadius{NO_PATROLLING};
 		template <typename Handler> void serialize(Handler &h)
 		{
 			h & patrolling;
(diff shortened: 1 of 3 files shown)


lib/mapObjects/CGHeroInstance.h
```diff
@@ -92,14 +92,15 @@ class DLL_LINKAGE CGHeroInstance : public CArmedInstance, public IBoatGenerator,
...
 	//std::map<ui16, const CArtifact*> artifWorn; //map<position,artifact_id>; positions: 0 - head; 1 - shoulders; 2 - neck; 3 - right hand; 4 - left hand; 5 - torso; 6 - right ring; 7 - left ring; 8 - feet; 9 - misc1; 10 - misc2; 11 - misc3; 12 - misc4; 13 - mach1; 14 - mach2; 15 - mach3; 16 - mach4; 17 - spellbook; 18 - misc5
 	std::set<ObjectInstanceID> visitedObjects;
 
 	struct DLL_LINKAGE Patrol
 	{
-		Patrol(){patrolling=false;initialPos=int3();patrolRadius=-1;};
+		Patrol(){patrolling=false;initialPos=int3();patrolRadius=NO_PATROLLING;};
 		bool patrolling;
 		int3 initialPos;
 		ui32 patrolRadius;
```
kambala-decapitator, line 106:
```suggestion
		bool patrolling{false};
		int3 initialPos;
		ui32 patrolRadius{NO_PATROLLING};
```

kambala-decapitator: also, setting -1 to an **unsigned** integer smells

godric3: Yeah, will recheck logic and probably change to max ui32 value

mapeditor/inspector/inspector.cpp
```diff
@@ -711,6 +720,13 @@ void Inspector::setProperty(CGHeroInstance * o, const QString & key, const QVari
 		o->randomizeArmy(o->getHeroType()->heroClass->faction);
 		updateProperties(); //updating other properties after change
 	}
+
+	if(key == "Patrol radius")
+	{
+		auto radius = value.toInt();
+		o->patrol.patrolRadius = radius;
+		o->patrol.patrolling = radius != std::numeric_limits<ui32>::max();
```
kambala-decapitator, line 728: this doesn't look nice of course.

if mingw is giving a hard time for w/e reason, simply define the value of the static variable in the cpp file:
```cpp
// h
static constexpr ui32 NO_PATROLLING; // maybe you'll have to use const instead of constexpr

// cpp
CGHeroInstance::NO_PATROLLING = std::numeric_limits<ui32>::max();
```

mapeditor/inspector/inspector.cpp
```diff
@@ -335,6 +335,15 @@ void Inspector::updateProperties(CGHeroInstance * o)
 		}
 		addProperty("Hero type", o->getHeroTypeID().hasValue() ? o->getHeroType()->getNameTranslated() : "", delegate, false);
 	}
+	{
+		const int maxRadius = 60;
+		auto * patrolDelegate = new InspectorDelegate;
+		patrolDelegate->options = { {QObject::tr("No patrol"), QVariant::fromValue(std::numeric_limits<ui32>::max())} };
+		for(int i = 0; i <= maxRadius; ++i)
+			patrolDelegate->options.push_back({ QObject::tr("%1 tile(s)").arg(i), QVariant::fromValue(i) });
+		auto patrolRadiusText = o->patrol.patrolling ? QObject::tr("%1 tile(s)").arg(o->patrol.patrolRadius) : QObject::tr("No patrol");
```
kambala-decapitator, line 344: this is an incorrect way to localise plurals. You should pass the number as third parameter to `tr()`, see here: https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

kambala-decapitator (approved):

PR #4840, opened 2024-10-29
Various translation fixes

GeorgeK1ng:
1) Updated Czech translation
2) Fixed coalesting text space in Pyramid, same as last fix for Shrine
3) Fixed ImpCache definition name typo
4) Updated Russian translation with corrected ImpCache instead of InpCache


dydzio0614:
unless @IvanSavenko recommends otherwise I would revert imp cache typo fix for now as it has severe consequences

GeorgeK1ng:
> unless @IvanSavenko recommends otherwise I would revert imp cache typo fix for now as it has severe consequences

Let's see what Ivan found then, I have uploaded fixes for all dependends mods aswell.

GeorgeK1ng:
I have searched across both VCMI and Mods repositories and it should be fine, when applied in main bank definition, then in dependend mods. InpCache is not hardcoded anywhere from my search in GitHub, just 3 mods depency.

misiokles:
There is also one bug in fortress faction. Special1 and special2 buildings add +2 defense to garrosoned heroes, but one should add +2 defense, second +2 attack.

Olegmods:
> There is also one bug in fortress faction. Special1 and special2 buildings add +2 defense to garrosoned heroes, but one should add +2 defense, second +2 attack.

Fixed in #4842

config/objects/creatureBanks.json
```diff
@@ -261,7 +261,7 @@
 					}
 				]
 			},
-			"inpCache" :  {
+			"impCache" :  {
```
IvanSavenko, line 264:
```suggestion
			"impCache" :  {
				"compatibilityIdentifiers" : [ "inpCache" ],
```

IvanSavenko:
Should be safe I think.
The only thing that may actually break would be maps with, ahem, inpCache strings in them. Can be fixed via compatibilityIdentifiers field

config/objects/creatureBanks.json
Laserlicht: But would be better anyway to also fix it in any translation mod?

GeorgeK1ng: It exist as InpCache only in Russian translation.

Laserlicht: In most cases these strings are in the translation mod.

Examples:
- https://github.com/vcmi-mods/german-translation/blob/vcmi-1.6/content/config/vcmi-german/game.json
- https://github.com/vcmi-mods/french-translation/blob/vcmi-1.6/french-translation/content/config/vcmi-french/game.json

GeorgeK1ng: There is something I don't understand. That's why I told it's not in translations. Will check it manually instead.
image

misiokles: github search options can't read string from files, only in filenames

IvanSavenko: Looks like it

> Beyond that, we also don't include all files in the search index:
    Vendored and generated code is excluded
    Empty files and files over 350 kiB are excluded
    Only UTF-8 encoded files are indexed
    Very large repositories may not be indexed

Laserlicht: I think `mega-pack-rus` and `h3-for-vcmi-englisation` are missing (for historical reasons there isn't `translation` in mod name)

GeorgeK1ng: Created PRs for them aswell now. Hope thats all

misiokles: I polish there's as well inpCache

GeorgeK1ng: > I polish there's as well inpCache. I guess all translations have exported inpCache string, not impCache

Polish PR is already done too

PR #4854, opened 2024-10-30
Autosave name

### LINKED ISSUE #4846: Autosave naming of files/directiry
**Describe your proposal**
It is quite confusing to navigate and find last saved/autosaved games in in-game interface. Sorting by date doesn't seems to effect the directory names. When the amount of saved maps grows finding a proper directory for recently played games is very annoying. The game is hiding somewhere in the middle of the alpha-sorted list, which requires to read all unaligned timstamps, which is quite painful.

Will it be nicer to use already existed sorting and place the timestamp in front of the map name, and align them to the left. All timestamp has same width, so that will ensure that each timestamp will be placed one below another, which makes ease to compare specific timestamp part (eg specific date, time, ... ) by eye.

Will it be possible to change the naming format of the directory of saved game to :
- "%Y-%m-%d-%H-%M \<name of the map\>" 
or
- "%Y-%m-%d %H-%M \<name of the map>\" 
or
- "%Y%m%d%H%M \<name of the map>\" 

It should be very easy to do and will ensure all saves are sorted by date ASC, and it make a lot easier to navigate to the latest saves regardless of map name: 

- Original
image
- Timestam in front without spaces
image
- Timestam in front with spaces
image

Laserlicht:
Fixes #4846

- DateTime first
- left align folder names (only for Autosave folders)
- Folders follows order direction now
- fix unwanted selection of element by clicking on empty slot

image


IvanSavenko:
What about making it a bit more readable while you're at it?
- drop creation time, instead use suffix, e.g. `_1`, `_2`, if player starts game on the same map in the same day
- format creation date in more readable form, e.g. `2024_10_30`

IvanSavenko:
> Do not have a smart idea for this. Only parsing of all saves/folders... Personally I would prefer with time... But...

Yeah, will need searching for first non-existing directory. 

I understand that this date+time format was chosen for simplicity, but unfortunately it is not very readable and time does not provides really useful information

krs0:
I think this format with 'complicated' date times, will be a little unfriendly to not so technical players.

Laserlicht:
Changed datetime format to `%Y-%m-%d_%H-%M` (without seconds).

Will not implement counter atm. Feel free if someone want to implement on top of this.
But counter has to be also serialized into savegame (if you load save from older folder it should continue old folder -> with timestamp it's working atm).

PR #4857, opened 2024-11-01
Delete saves / random maps

### LINKED ISSUE #3389: Delete save from saved games list
**Describe your proposal**
Heroes 3 actually allows deletion of saves - via "Delete" key.
However this option
1) not obvious
2) not available on mobile platforms which vcmi needs to account for

VCMI currently does not provides this option in any form however. We need to:
1) Implement such option to begin with. Ideally - should be done as request to server since saves *should* be loaded by server, not client.
2) Assign this option to "delete" key AND add button in UI. For example - in place of random map dialog (not available from save/load game UI)
3) Consider how to handle saves for old versions of the game. We should either display all saves to allow their deletion or add option to delete all outdated saves

Laserlicht:
Fixes #3389

- Button to toggle to delete mode (load game & newgame for random maps)
- In delete mode: clicking on element asks to delete it (folder & files possible)
- Asks for deleting incompatible files if found
- Deleting with Del button is not part of this PR because of buggy shortcut system (Can't use same shortcut combination twice if at least one of them use it in keyPress)

grafik
grafik
grafik


Olegmods:
Just pure interest. Does "Unsupported games" include ones that were made with previous versions of mods and aren't working anymore now?

Laserlicht:
> Just pure interest. Does "Unsupported saves" include ones that were made with previous versions of mods and aren't working anymore now?

No. only for these errors:
- cannot open to read
- not a VCMI file
- too old file format
- too new file format

client/lobby/SelectionTab.cpp
```diff
@@ -868,7 +927,28 @@ void SelectionTab::parseSaves(const std::unordered_set<ResourcePath> & files)
...
+					CInfoWindow::showYesNoDialog(CGI->generaltexth->translate("vcmi.lobby.deleteUnsupportedSave"), std::vector<std::shared_ptr<CComponent>>(), [&deleteUnsupported, doDeleteUnsupported, file](){
+						doDeleteUnsupported(file.getName());
+						deleteUnsupported = DELETE;
+					}, [&deleteUnsupported](){ deleteUnsupported = REMAIN; });
+
+					while(deleteUnsupported == NONE)
+					{
+						auto unlockInterface = vstd::makeUnlockGuard(GH.interfaceMutex);
+						boost::this_thread::sleep_for(boost::chrono::milliseconds(5));
+					}
```
IvanSavenko, line 946: Please avoid such busy-waits. Better to generate list of saves & process it later

client/lobby/SelectionTab.cpp
```diff
@@ -868,7 +927,28 @@ void SelectionTab::parseSaves(const std::unordered_set<ResourcePath> & files)
 		}
 		catch(const std::exception & e)
 		{
-			logGlobal->error("Error: Failed to process %s: %s", file.getName(), e.what());
+			// asking for deletion of unsupported saves
```
IvanSavenko, line 930: Also, this would generate this warning on every run.
Perhaps we should remember VCMI version on which this option was proposed and only ask again when player ugrades / downgrades VCMI?

Mods/vcmi/config/english.json
```diff
@@ -106,6 +106,12 @@
 	"vcmi.lobby.handicap.resource" : "Gives players appropriate resources to start with in addition to the normal starting resources. Negative values are allowed, but are limited to 0 in total (the player never starts with negative resources).",
 	"vcmi.lobby.handicap.income" : "Changes the player's various incomes by the percentage. Is rounded up.",
 	"vcmi.lobby.handicap.growth" : "Changes the growth rate of creatures in the towns owned by the player. Is rounded up.",
+	"vcmi.lobby.deleteUnsupportedSave" : "Unsupported saves found (e.g. from previous versions).\n\nDelete them?",
```
IvanSavenko, line 109: Slightly better phrasing, added title, added placeholder for number of such saves.
```suggestion
	"vcmi.lobby.deleteUnsupportedSave" : "{Unsupported saves found}\n\nVCMI has found %d saved games which are no longer supported, such as from different versions of VCMI. \n\nDo you wish to delete them?",
```

MichalZr6: Please consider also: 
`"{Unsupported saves found}\n\nVCMI has found %d saved games that are no longer supported, possibly due to differences in VCMI versions.\n\nDo you want to delete them?"`

client/lobby/SelectionTab.cpp
Laserlicht: Still prefer asking every time. It's unnecessary to keep them a long time on HDD as them never will work again.

Only to give the player the possibility to finish a savegame by downgrading VCMI.

Laserlicht:
@GeorgeK1ng You need to update translation. String has changed...

client/lobby/SelectionTab.cpp
```diff
@@ -866,11 +921,36 @@ void SelectionTab::parseSaves(const std::unordered_set<ResourcePath> & files)
...
 		{
+			unsupported.push_back(file); // IdentifierResolutionException is not relevant -> not ask to delete, when mods are disabled
 			logGlobal->error("Error: Failed to process %s: %s", file.getName(), e.what());
 		}
 	}
+
+	return unsupported;
+}
+
+void SelectionTab::handleUnsupportedSavegames(const std::vector<ResourcePath> files)
```
IvanSavenko, line 938:
```suggestion
void SelectionTab::handleUnsupportedSavegames(const std::vector<ResourcePath> & files)
```

lib/networkPacks/PacksForLobby.h
```diff
@@ -381,4 +383,22 @@ struct DLL_LINKAGE LobbyPvPAction : public CLobbyPackToServer
 	}
 };
 
+struct DLL_LINKAGE LobbyDelete : public CLobbyPackToServer
+{
+	enum EType : ui8 {
```
IvanSavenko, line 388: To prevent accidental int<->enum conversions
```suggestion
	enum class EType : ui8 {
```

IvanSavenko (approved):

PR #4858, opened 2024-11-01
Gamepad qt launcher

### LINKED ISSUE #4111: Gamepad support for QT
**Describe your proposal**
As the old launcher on android is usable over gamepad (e.g. on Android TV) it would be good to have at least simple support.
At least a simple press on "Start" to start the game should work.

Or implement pointer emulation: https://doc.qt.io/qt-5/qtgamepad-mouseitem-example.html

**Additional context**
Implement for example using: https://doc.qt.io/qt-5/qtgamepad-index.html
Examples: https://code.qt.io/cgit/qt/qtgamepad.git/tree/examples/gamepad/simple?h=5.15

Laserlicht:
fixes #4111

- `start` button starts client
- other keys are mapped to keyboard-keys (almost everything is possible with gamepad)


kambala-decapitator:
Not sure why you need SDL https://doc.qt.io/qt-5/qtgamepad-index.html

Laserlicht:
> Not sure why you need SDL https://doc.qt.io/qt-5/qtgamepad-index.html

Would also prefer this one. But it's seems to be dead (not supported by qt6 anymore).

launcher/sdlHandler.cpp
```diff
@@ -0,0 +1,141 @@
...
+			break;
+		case SDL_CONTROLLER_BUTTON_X:
+			sendKey(Qt::Key_Tab);
+			break;
+		case SDL_CONTROLLER_BUTTON_Y:
+			sendKey(Qt::Key_Enter);
+			break;
+		case SDL_CONTROLLER_BUTTON_A:
+			sendKey(Qt::Key_Space);
+			break;
```
IvanSavenko, line 139: What for we need space?

Perhaps:
LB / RB -> shift-tab / tab. These buttons are usually used to switch between controls
A is cancel so Esc. Do we have controls where Esc makes sense? If not then can drop it.
B is confirm, so Enter.

Laserlicht: Space is needed to open dropdown (not possible with enter)

Laserlicht:
@kambala-decapitator Any idea to bring it up working on android? It crashes on android.

I think the reason is that SDL is not correctly initialized (Java stuff). But we need `#define SDL_MAIN_HANDLED` because SDL_main black magic isn't working with single binary builds (on desktop that's not a problem).

Tried to fix it with:
```
	QAndroidJniObject::callStaticObjectMethod("org/libsdl/app/SDL", "setupJNI", "(V)Lorg/libsdl/app/SDL;");
	QAndroidJniObject::callStaticObjectMethod("org/libsdl/app/SDL", "initialize", "(V)Lorg/libsdl/app/SDL;");
```

Generally it's working on Linux without problems.

kambala-decapitator:
> > Not sure why you need SDL https://doc.qt.io/qt-5/qtgamepad-index.html
> 
> Would also prefer this one. But it's seems to be dead (not supported by qt6 anymore).

Indeed. But there's a port: https://github.com/pumphaus/qtgamepadlegacy

as for android, that would probably be difficult. Currently SDL activity is started once you leave the launcher, and the actual gamepad support is inside SDL Java code iirc.

Laserlicht:
-> Main reason for PR was android, but this approach doesn't work on android...

kambala-decapitator:
You could use Qt's gamepad support as we're currently on Qt 5 on android - would be better than nothing

PR #4864, opened 2024-11-02
Creature window rework

Laserlicht:
- changed text size (and limit field for long texts)
- fix title Y-pos
- add colored border and labels for bonus source
- remove scrollbar for < 4 lines

Also added sort with following order:
1. By Source (Ability forced as first)
2. By Name

(outdated) Screenshot:
grafik


krs0:
Could abilities be sorted by json order?
Alphabetical order makes no logical sense when creature has many abilities. Even more evident when scrolling is involved.

This way, moders will be able to show to players, a more logical order of abilities.

{48735C6B-E74F-47C8-BD87-295D7E586ECA}

{DD7A37EA-B236-4C23-BF91-3FE9328CF363}

IvanSavenko:
There is no such thing as 'json order'.
We don't preserve it on parser stage.

The only option is to sort by our internal bonus ID enumeration

client/windows/CCreatureWindow.cpp
```diff
@@ -786,6 +829,9 @@ void CStackWindow::initBonusesList()
 	BonusList output;
 	BonusList input;
 	input = *(info->stackNode->getBonuses(CSelector(Bonus::Permanent), Selector::all));
+	std::sort(input.begin(), input.end(), [this](std::shared_ptr<Bonus> v1, std::shared_ptr<Bonus> & v2){
+		return v1->source == v2->source ? info->stackNode->bonusToString(v1, false) < info->stackNode->bonusToString(v2, false) : v1->source == BonusSource::CREATURE_ABILITY || (v1->source < v2->source);
```
IvanSavenko, line 833:
```suggestion
		if (v1->source != v2->source)
			return v1->source == BonusSource::CREATURE_ABILITY || (v1->source < v2->source);
		else
			return  info->stackNode->bonusToString(v1, false) < info->stackNode->bonusToString(v2, false);
```

misiokles:
BTW. Creature ability with "propagator" : "HERO" command doesn't show in creature window (for example Courtyard's Homunculi, ToW's Mermaids).

krs0:
I found this artifact... the V in **V**ulnerable is cut short. Is this related? Or something different?

image

PR #4874, opened 2024-11-03
prescaled image support

### LINKED ISSUE #1070: [FEATURE] Possibility to use HD textures

Laserlicht:
Support for prescaled images. Allows hybrid usage of xbrz and prescaled.

Original size is still required. But additional hires can be loaded:
E.g. for `ADOPFLGB.bmp` -> `ADOPFLGB$3.png`. It searches images with the same name but `$SCALEFACTOR`-Suffix. If it is found it uses it instead doing xbrz. Multiple scale factors can be provided (2-4). It prefer the one with the current xbrz factor. If it's not found it uses the biggest found image and downscale it. Works also with def (but strongly suggest json for animations; def has some size limitations).

Can be used to provide HQ mods and maybe also for a conversion script converting official HD files.

Fixes #1070

Here's VCMI with some HD graphics from official HD version:
grafik

### DIFF
--- a/client/render/ImageLocator.h
+++ b/client/render/ImageLocator.h
@@ -33,6 +33,7 @@ struct ImageLocator
 	bool verticalFlip = false;
 	bool horizontalFlip = false;
 	int8_t scalingFactor = 0; // 0 = auto / use default scaling
+	int8_t preScaledFactor = 1;
 	EImageLayer layer = EImageLayer::ALL;
 
 	ImageLocator() = default;
--- a/client/renderSDL/RenderHandler.cpp
+++ b/client/renderSDL/RenderHandler.cpp
@@ -148,7 +148,9 @@ void RenderHandler::initFromJson(AnimationLayoutMap & source, const JsonNode & c
 
 RenderHandler::AnimationLayoutMap & RenderHandler::getAnimationLayout(const AnimationPath & path)
 {
-	AnimationPath actualPath = boost::starts_with(path.getName(), "SPRITES") ? path : path.addPrefix("SPRITES/");
+	auto tmp = getScalePath(path);
+	auto animPath = AnimationPath::builtin(tmp.first.getName());
+	AnimationPath actualPath = boost::starts_with(animPath.getName(), "SPRITES") ? animPath : animPath.addPrefix("SPRITES/");
 
 	auto it = animationLayouts.find(actualPath);
 
@@ -175,10 +177,14 @@ RenderHandler::AnimationLayoutMap & RenderHandler::getAnimationLayout(const Anim
 		std::unique_ptr<ui8[]> textData(new ui8[stream->getSize()]);
 		stream->read(textData.get(), stream->getSize());
 
-		const JsonNode config(reinterpret_cast<const std::byte*>(textData.get()), stream->getSize(), path.getOriginalName());
+		const JsonNode config(reinterpret_cast<const std::byte*>(textData.get()), stream->getSize(), animPath.getOriginalName());
 
 		initFromJson(result, config);
 	}
+
+	for(auto & g : result)
+		for(auto & i : g.second)
+			i.preScaledFactor = tmp.second;
 
 	animationLayouts[actualPath] = result;
 	return animationLayouts[actualPath];
@@ -235,7 +241,17 @@ std::shared_ptr<ISharedImage> RenderHandler::loadImageFromFileUncached(const Ima
 	if (locator.defFile)
 	{
 		auto defFile = getAnimationFile(*locator.defFile);
-		return std::make_shared<SDLImageShared>(defFile.get(), locator.defFrame, locator.defGroup, locator.preScaledFactor);
+		int preScaledFactor = locator.preScaledFactor;
+		if(!defFile) // no presscale for this frame
+		{
+			auto tmpPath = (*locator.defFile).getName();
+			boost::algorithm::replace_all(tmpPath, "$2", "");
+			boost::algorithm::replace_all(tmpPath, "$3", "");
+			boost::algorithm::replace_all(tmpPath, "$4", "");
+			preScaledFactor = 1;
+			defFile = getAnimationFile(AnimationPath::builtin(tmpPath));
+		}
+		return std::make_shared<SDLImageShared>(defFile.get(), locator.defFrame, locator.defGroup, preScaledFactor);
 	}
 
 	throw std::runtime_error("Invalid image locator received!");
@@ -369,14 +385,7 @@ std::shared_ptr<IImage> RenderHandler::createImage(SDL_Surface * source)
 
 std::shared_ptr<CAnimation> RenderHandler::loadAnimation(const AnimationPath & path, EImageBlitMode mode)
 {
-	auto tmp = getScalePath(path);
-	auto animPath = AnimationPath::builtin(tmp.first.getName());
-	auto layout = getAnimationLayout(animPath);
-	for(auto & g : layout)
-		for(auto & i : g.second)
-			i.preScaledFactor = tmp.second;
-
-	return std::make_shared<CAnimation>(animPath, layout, mode);
+	return std::make_shared<CAnimation>(path, getAnimationLayout(path), mode);
 }
 
 void RenderHandler::addImageListEntries(const EntityService * service)
--- a/client/renderSDL/RenderHandler.cpp
+++ b/client/renderSDL/RenderHandler.cpp
@@ -246,9 +246,9 @@ std::shared_ptr<ISharedImage> RenderHandler::loadImageFromFileUncached(const Ima
 		if(!defFile) // no prescale for this frame
 		{
 			auto tmpPath = (*locator.defFile).getName();
-			boost::algorithm::replace_all(tmpPath, "2X/", "/");
-			boost::algorithm::replace_all(tmpPath, "3X/", "/");
-			boost::algorithm::replace_all(tmpPath, "4X/", "/");
+			boost::algorithm::replace_all(tmpPath, "SPRITES2X/", "SPRITES/");
+			boost::algorithm::replace_all(tmpPath, "SPRITES3X/", "SPRITES/");
+			boost::algorithm::replace_all(tmpPath, "SPRITES4X/", "SPRITES/");
 			preScaledFactor = 1;
 			defFile = getAnimationFile(AnimationPath::builtin(tmpPath));
 		}
--- a/client/renderSDL/RenderHandler.cpp
+++ b/client/renderSDL/RenderHandler.cpp
@@ -361,6 +361,18 @@ std::shared_ptr<IImage> RenderHandler::loadImage(const ImageLocator & locator, E
 	}
 	else
 	{
+		if(loc.image)
+		{
+			std::string imgPath = (*loc.image).getName();
+			if(loc.layer == EImageLayer::OVERLAY)
+				imgPath += "-overlay";
+			if(loc.layer == EImageLayer::SHADOW)
+				imgPath += "-shadow";
+
+			if(CResourceHandler::get()->existsResource(ImagePath::builtin(imgPath)))
+				loc.image = ImagePath::builtin(imgPath);
+		}
+
 		return loadImageImpl(loc)->createImageReference(mode);
 	}
 }
--- a/client/renderSDL/RenderHandler.cpp
+++ b/client/renderSDL/RenderHandler.cpp
@@ -299,7 +299,7 @@ std::shared_ptr<ISharedImage> RenderHandler::scaleImage(const ImageLocator & loc
 	if (locator.layer == EImageLayer::ALL && locator.playerColored != PlayerColor::CANNOT_DETERMINE)
 		handle->playerColored(locator.playerColored);
 
-	handle->scaleInteger(locator.scalingFactor);	
+	handle->scaleInteger(locator.scalingFactor);
 
 	// TODO: try to optimize image size (possibly even before scaling?) - trim image borders if they are completely transparent
 	auto result = handle->getSharedImage();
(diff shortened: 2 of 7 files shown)


misiokles:
One guy wrote years ago Software for splitting H3 HD Edition graphics for frames. He used some scripts.
Link to forum: http://heroescommunity.com/viewthread.php3?TID=42984&pagenumber=1
Maybe you'll find there sth usefull

Laserlicht:
Yes, also know this. Adapted this last year to a python script, which is much easier to handle: https://github.com/Laserlicht/heroes3_image_extractor/blob/main/hd_extract.py

Also used it for this test. But it still needs some work to build automatically a mod from this. There are a few things different between HD and complete which needs to handled.

Laserlicht:
Marked as draft.

There seems to be a error for fallback logic. @IvanSavenko Any idea?

Game currently crashes at `MapRendererRiver::renderTile`. Maybe not all frames exists...

https://github.com/Laserlicht/vcmi_hd_mod.git can be used to generate mod from HD install.

Many things are working already. But still a lot of issues (mainly offsets - which have to be handled in mod generation script)...

See screenshots in full screen:
grafik
grafik
grafik
grafik
grafik

misiokles:
Rivers don't have animations. It's some kind of palette shifting, like in vanilla version.

IvanSavenko:
Can you show generated json's for rivers? Do you see how hd edition implemented river animations btw?

Laserlicht:
> Can you show generated json's for rivers? Do you see how hd edition implemented river animations btw?

Understood the problem. HD mod has multiple frames instead of paletteshifting...

Laserlicht:
Main problem is currently some offsets of some animations. Don't know why!?

And a bugging adventure map rendering

See here (sorry for low fps video):
[Peek 2024-11-07 02-42.webm](https://github.com/user-attachments/assets/85bfce77-1a1f-41ef-aad7-d2ba5aaf66aa)

IvanSavenko:
> Main problem is currently some offsets of some animations

Haven't checked your code, but h3 def files have support for offsets. For example, actual image for combat animation is often way smaller than 450x400. Internally, def's store smaller image, ~150x200px and also store information on borders width.

Check `SDLImageShared`, members `margins` and `fullSize`. I suspect HD Edition also uses similar scheme, however not sure if values are identical to h3 ones. From video it looks like you're loading images without any offset information.

> And a bugging adventure map rendering

Not 100% sure, but looks like only flipped terrain/road/river tiles are broken. You can try checking this by changing several `rotationIndex` in MapRenderer.cpp to 0. If it (somewhat) helps - then something is broken with vertical/horizontal flipping of tiles.

Laserlicht:
Thanks @IvanSavenko :)

Huge progress. Fixed adventure map issue (error in PR here). Also added offsets from sd version.

Remaining problems:
- [ ] River animation currently disabled (HD uses separate images instead of `paletteAnimation`)
- [ ] Palette effects are not working (e.g. flags, colored menu... HD uses seperate images instead of palette stuff)
- [ ] Offsets/Margins from SD are not 100% perfect fitting -> not sure how we get the real values (may be hardcoded!?) - Has to be fixed in convert process
- [ ] Some graphics still sd (should have hd variant) -> need to investigate

IvanSavenko:
> River animation currently disabled (HD uses separate images instead of paletteAnimation)

How exactly are those organized? Similarly to other animations, e.g. hero? So animation file has large number of 'groups', one for each animated tile?

> Palette effects are not working (e.g. flags, colored menu... -> setOverlayColor HD uses seperate images instead of palette stuff)

Actually same goes for vcmi in xbrz mode. Flags are extracted from main image and then colorized & upscaled separately. 

> Offsets/Margins from SD are not 100% perfect fitting -> not sure how we get the real values (may be hardcoded!?) 

Not hardcoded. From what I remember about hd edition, there should be text description for each .def replacement which contains table of numbers. This is most likely what you need.

client/renderSDL/RenderHandler.cpp
```diff
@@ -55,6 +55,58 @@ std::shared_ptr<CDefFile> RenderHandler::getAnimationFile(const AnimationPath &
...
+}
+
+std::pair<ResourcePath, int> RenderHandler::getScalePath(ResourcePath p)
+{
+	auto path = p;
+	auto name = p.getName();
+	int scaleFactor = 1;
+	if(getScalingFactor() > 1)
+	{
+		std::vector<int> factorsToCheck = {getScalingFactor(), 4, 3, 2};
```
IvanSavenko, line 92: Why we need to check multiple scaling factors? We should only try to load one that matches desired one.

client/renderSDL/RenderHandler.cpp
```diff
@@ -55,6 +55,58 @@ std::shared_ptr<CDefFile> RenderHandler::getAnimationFile(const AnimationPath &
 	return result;
 }
 
+std::optional<ResourcePath> RenderHandler::getPath(ResourcePath path)
```
IvanSavenko, line 58: Perhaps we should clean up this logic? So:
- path MUST be specified without `Data` or `Sprites` prefix
- for unscaled images `Sprites` is preferred, `Data` is only used as fallback location
- for animations, always load them from `Sprites`

client/renderSDL/RenderHandler.cpp
```diff
@@ -55,6 +55,58 @@ std::shared_ptr<CDefFile> RenderHandler::getAnimationFile(const AnimationPath &
...
+	auto name = p.getName();
+	int scaleFactor = 1;
+	if(getScalingFactor() > 1)
+	{
+		std::vector<int> factorsToCheck = {getScalingFactor(), 4, 3, 2};
+		for(auto factorToCheck : factorsToCheck)
+		{
+			ResourcePath scaledPath = ImagePath::builtin(name + "$" + std::to_string(factorToCheck));
+			if(p.getType() != EResType::IMAGE)
+				scaledPath = AnimationPath::builtin(name + "$" + std::to_string(factorToCheck));
```
IvanSavenko, line 97: What about instead of suffix use separate directory, e.g. `Sprites2x` / `Sprites3x` / `Sprites4x`? 
Might be more convenient for batch-generation of other scaling factors

client/renderSDL/SDLImage.cpp
```diff
@@ -274,9 +277,15 @@ std::shared_ptr<ISharedImage> SDLImageShared::scaleInteger(int factor, SDL_Palet
 	if (palette && surf && surf->format->palette)
 		SDL_SetSurfacePalette(surf, palette);
 
-	SDL_Surface * scaled = CSDL_Ext::scaleSurfaceIntegerFactor(surf, factor, EScalingAlgorithm::XBRZ);
+	SDL_Surface * scaled = nullptr;
+	if(preScaleFactor == factor)
+		scaled = CSDL_Ext::scaleSurfaceIntegerFactor(surf, 1, EScalingAlgorithm::NEAREST); // keep size
```
IvanSavenko, line 282: I think in this case we can safely return this (or well, shared_from_this) and skip copy creation

Laserlicht:
Thanks for reviewing. Will address them later.

> How exactly are those organized? Similarly to other animations, e.g. hero? So animation file has large number of 'groups', one for each animated tile?

Def file is the same as e.g. in complete. But for x2 and x3 there are 11 frames per "def frame" (separated with "_"):

grafik

> Actually same goes for vcmi in xbrz mode. Flags are extracted from main image and then colorized & upscaled separately.

They removed the flag from sprites:

grafik

And there are seperate flags per color as png in gamefolder:
grafik

> Not hardcoded. From what I remember about hd edition, there should be text description for each .def replacement which contains table of numbers. This is most likely what you need.

Would very happy if I had a file like this one. But I've only found `spriteFlagsInfo.txt` (position of flag for each object with flag).

In the pak files are informations that looks similar. But they are for correctly extracting the single images from dds sheet.

IvanSavenko:
> Def file is the same as e.g. in complete. But for x2 and x3 there are 11 frames per "def frame" (separated with "_"):

Is this different from (for example) animation of heroes / creatures in battle?

> They removed the flag from sprites:

Ah, yeah, it is a bit different. VCMI also generates such base image with no flag (but since we've cut it out from image, base image now has hole in place of flag).

However offsets are not needed since that data is taken from cut out part of the image.
Not sure what would be best approach here...
You could try to generate animation with offsets that always points to same image. But at the moment we don't support loading of such additional overlays from json animation - needs to be implemented.

> Would very happy if I had a file like this one. But I've only found spriteFlagsInfo.txt (position of flag for each object with flag).

Perhaps such information is unnecessary? For example, what if HD Edition always centers images? So if hd edition stores 150x200 px image for creatures in battles, which h3 expects to be 450x400, then actual offset will be:
`offset = (expected size - stored size) / 2 = (450x400 - 150x200) / 2 = 300x200 / 2 = 150x100`

IvanSavenko:
BTW - what about those unused values from your importer script?
https://github.com/Laserlicht/heroes3_image_extractor/blob/main/hd_extract.py#L104
Values 2-5 are not used. Perhaps these are the offsets you're looking for?

Laserlicht:
> BTW - what about those unused values from your importer script? https://github.com/Laserlicht/heroes3_image_extractor/blob/main/hd_extract.py#L104 Values 2-5 are not used. Perhaps these are the offsets you're looking for?

Good idea. :)
In every source of the file format description they are marked as unknown. And basically really low values.

Laserlicht:
Fixed offset problem. Value 2 was x, value 4 y :)

client/renderSDL/SDLImage.cpp
Laserlicht: `error: could not convert ‘std::enable_shared_from_this<_Tp>::shared_from_this() const [with _Tp = SDLImageShared]()’ from ‘shared_ptr<const SDLImageShared>’ to ‘shared_ptr<ISharedImage>’`

client/renderSDL/RenderHandler.cpp
Laserlicht: Because also a downscaling is applied, example:
We only have x4 images. User selected x2. Then the images are downscaled to x2.
If there are also x2 images they are used without scaling.

Laserlicht:
Ready for now.

Official HD is working now.
Some small things needed to fix, but only in conversation script, not in VCMI implementation.

client/renderSDL/SDLImage.cpp
IvanSavenko: Ah, this is const method that returns non-const value, so can't return non-const this.
A bit of a hack, but should work:
```suggestion
		scaled = std::const_pointer_cast<const SDLImageShared>(shared_from_this());
```

Although we should probably replaces all `std::shared_ptr<ISharedImage>` with `std::shared_ptr<const ISharedImage>`

client/render/ImageLocator.h
```diff
@@ -33,6 +33,7 @@ struct ImageLocator
 	bool verticalFlip = false;
 	bool horizontalFlip = false;
 	int8_t scalingFactor = 0; // 0 = auto / use default scaling
+	int8_t preScaledFactor = 1;
```
IvanSavenko, line 36: Why `scalingFactor` is not enough? How does `preScaledFactor` differs from it?

client/renderSDL/RenderHandler.cpp
```diff
@@ -177,13 +236,23 @@ std::shared_ptr<ISharedImage> RenderHandler::loadImageFromFileUncached(const Ima
...
 	{
 		auto defFile = getAnimationFile(*locator.defFile);
-		return std::make_shared<SDLImageShared>(defFile.get(), locator.defFrame, locator.defGroup);
+		int preScaledFactor = locator.preScaledFactor;
+		if(!defFile) // no prescale for this frame
+		{
+			auto tmpPath = (*locator.defFile).getName();
+			boost::algorithm::replace_all(tmpPath, "2X/", "/");
+			boost::algorithm::replace_all(tmpPath, "3X/", "/");
+			boost::algorithm::replace_all(tmpPath, "4X/", "/");
```
IvanSavenko, line 251: This would break mods that use '2x' as part of directory name (not necessarily top-level one)
We should only try this if path starts with 'sprites2x'

client/render/ImageLocator.h
Laserlicht: Tried to unify them but failed... to complicated...

`scalingFactor` - describes the desired factor to scale
`preScaledFactor` - describes the factor a image is already scaled before loading (e.g. there is no auto like in `preScaledFactor`)

IvanSavenko:
Ah, also 
1) can you add documentation for modders - how one can provide pre-scaled animations and images for their mod? At least for cases where such data exists.
2) does this PR supports (for example) creatures in battle? They require selection highlight for some images.

client/renderSDL/SDLImage.cpp
Laserlicht: Also not possible...

It's also not possible to re-use surf pointer...

Replaced it with simple copy...

Laserlicht:
> Ah, also
> 
>     1. can you add documentation for modders - how one can provide pre-scaled animations and images for their mod? At least for cases where such data exists.
> 
>     2. does this PR supports (for example) creatures in battle? They require selection highlight for some images.

Done

Laserlicht:
Anything open here before merging?

Working on porting the python HD extractor to C++. Switching branches is painful. :D

kambala-decapitator:
> Switching branches is painful. :D

git worktree

client/renderSDL/RenderHandler.cpp
```diff
@@ -257,10 +326,24 @@ std::shared_ptr<ISharedImage> RenderHandler::scaleImage(const ImageLocator & loc
 
 std::shared_ptr<IImage> RenderHandler::loadImage(const ImageLocator & locator, EImageBlitMode mode)
 {
-	if (locator.scalingFactor == 0 && getScalingFactor() != 1 )
+	ImageLocator loc = locator;
```
IvanSavenko, line 329:
```suggestion
	ImageLocator adjustedLocator = locator;
```

IvanSavenko (approved):

IvanSavenko:
> Working on porting the python HD extractor to C++. Switching branches is painful. :D

Would be great to have, hopefully with usable performance / import time than python one.

Although ideally would've been great to use HD edition pack's as it.

IvanSavenko:
Thinking about it more - for HD edition using data as it makes more sense than import with conversion.
- faster load times (single archive with fast-loading dds vs massive number of slow-loading png's)
- ease of install for player, can be installed even manually if needed
- faster install for player, since there is no reencoding

Should also be fairly easy to implement I think

Issue #4689, opened 2024-09-28
Hero models on map from HotA are not used


Olegmods:
**Describe the bug**
In HotA mod there is a config that changes heroes' models according to gender (also configs from other mods doesn't seem to work correctly either). But it doesn't work at all and seems to still use the "Default" template. Only heroes on adventure map do not work, in battle they are displayed correctly. The bug was confirmed.

**Game logs**
I don't think it's needed.

**To Reproduce**
1. Download Horn of the Abyss mod
2. Choose Dungeon as starting faction
3. Choose Lorelei as starting hero.
4. See its model being male on the adventure map

**Expected behavior**
Lorelei having this model:
![image](https://github.com/user-attachments/assets/23ed7c43-e36e-4e08-968a-952f27554230)

**Actual behavior**
Lorelei (right) having the same model as the male hero (left):
![image](https://github.com/user-attachments/assets/3165ac70-c09e-416b-b5c7-b070d5f068b4)

**Did it work earlier?**
I think it worked in 1.1.1, and also confirmed to break on some version

**Screenshots**
![image](https://github.com/user-attachments/assets/3165ac70-c09e-416b-b5c7-b070d5f068b4)
![image](https://github.com/user-attachments/assets/23ed7c43-e36e-4e08-968a-952f27554230)

**Version**
 - OS: Windows, Android
 - Version: 1.5.6 and 1.5.7

Issue #4882, opened 2024-11-07
The latest branch will crash on Android


lishaoxia1985:
The crash occurs when you choose the scenario and touch the start button.

by003:
same as windows
![image](https://github.com/user-attachments/assets/15bef88b-f08d-4632-94ba-1ac00859e2e0)
Shows invalid creature id

pppbb:
The same problem here.

PR #4883, opened 2024-11-07
Fixes for recently reported issues

IvanSavenko:
- Fixes crash on starting a new map (#4882)
- Fixes save compatibility with 1.5
- Reworked and fixed selection of gendered sprites for heroes on adventure map. Will also update hota mod to use new logic.
- Updated schema to account for lighthouse -> flaggable change
- Added schema for validation of flaggable objects
- Removed broken addition of mod dependency for wog maps


Olegmods:
@IvanSavenko should I close #4689 ?

PR #4889, opened 2024-11-09
crash fix: artifact sort by slot

kdmcser:
Fix a crash: when creature artifacts and commander artifacts in backpack, sort by slot will lead a crash due to `inf.getArt()->getType()->getPossibleSlots().at(ArtBearer::HERO)` return a empty vector.


SoundSSGood:
Please try if there is a grail in the backpack

kdmcser:
image
seems OK, go to the last else branch

kdmcser:
the error message have been deleted.

SoundSSGood (approved):

Issue #3245, opened 2023-11-29
When you start a campaign, you see a folder with maps and other campaigns that are located in the “others” folder


Povelitelh3:
Not critical of course, but distracting
I made a video [OtherMaps.zip](https://github.com/vcmi/vcmi/files/13500297/OtherMaps.zip)

PR #4900, opened 2024-11-11
Chronicles menu

Laserlicht:
Solution for organizing chronicles in campaign set.

Changes:
- Custom Campaigns organized in campaign set and shown on mainmenu will not be displayed in Custom campaign list
- Campaigns can be marked as optional -> if not exists creation of preview-button is skipped
- Creates campaign screen with 8 slots from existing resources
- Crop out preview-buttons from chronicle (if exists)

Maybe we can add button to the theme mods. If no chronicle installed, then it displays an empty screen. But nothing bad happens.

Further PR in future: Find solution for custom campaign button in main menu, that will not be overridden by another theme...

[Peek 2024-11-11 23-53.webm](https://github.com/user-attachments/assets/6f39d7e4-c565-4add-b430-925749b18a45)


GeorgeK1ng:
This is great, but seems like you have same issue with displaying campaings folder on your video like in https://github.com/vcmi/vcmi/issues/3245

Laserlicht:
> This is great, but seems like you have same issue with displaying campaings folder on your video like in #3245

Yes. But this isn't easy to fix. Tried some approaches. But gave up. To time consuming to find a solution.

It's a problem caused by the actual implementation of server-client architecture of VCMI.

client/lobby/SelectionTab.cpp
```diff
@@ -875,15 +875,38 @@ void SelectionTab::parseSaves(const std::unordered_set<ResourcePath> & files)
...
+					if(file.getName() == ResourcePath(item["file"].String()).getName())
+						foundInSet = set.first;
+			
+			// set has to be used in main menu
+			bool setInMainmenu = false;
+			if(!foundInSet.empty())
+				for (auto const & item : mainmenu["window"]["items"].Vector())
+					if(item["name"].String() == "campaign")
+						for (auto const & button : item["buttons"].Vector())
+							if(boost::algorithm::contains(boost::algorithm::to_lower_copy(button["command"].String()), boost::algorithm::to_lower_copy(foundInSet)))
```
IvanSavenko, line 904: `boost::iequals`?

docs/players/Heroes Chronicles.md
IvanSavenko, line 1: Perhaps also add link to this file in main Readme.md?

client/mainmenu/CCampaignScreen.cpp
```diff
@@ -67,7 +67,8 @@ CCampaignScreen::CCampaignScreen(const JsonNode & config, std::string name)
 	}
 
 	for(const JsonNode & node : config[name]["items"].Vector())
-		campButtons.push_back(std::make_shared<CCampaignButton>(node, config, campaignSet));
+		if(CResourceHandler::get()->existsResource(ResourcePath(node["file"].String(), EResType::CAMPAIGN)))
+			campButtons.push_back(std::make_shared<CCampaignButton>(node, config, campaignSet));
```
IvanSavenko, line 71: We have logic that if some of references h3c's are missing then corresponding campaign screen won't be available.
Check references to `vcmi.client.errors.missingCampaigns` string ID.

I think right now this check is pointless / unreachable - player can't open such campaign screen.

Laserlicht: Yes. Know this. It's already checked before entering the screen:
https://github.com/vcmi/vcmi/blob/719859cc62bbcd54dc3f5541bca697f5df8706b3/client/mainmenu/CMainMenu.cpp#L428

"optional" parameter is exactly to avoid to make check useless.

by003:
I think we can make a snapshot from for chronicles open CG, which can more consistent with the original look.
Like this to become Sword of Frost 
image

Laserlicht:
> I think we can make a snapshot from for chronicles opening CG, which can more consistent with the original look. Like this to become Sword of Frost

Yeah. But decoding video frame isn't that easy as it sounds like. And possibly can't be done "on the fly". I think it's not worth the effort.

Issue #4904, opened 2024-11-13
Crash on Hill fort right click on random map


GeorgeK1ng:
When you try to right click on Hill fort on randomly generated map, game crash in latest daily build.

![image](https://github.com/user-attachments/assets/01a6f036-f89d-4a54-9487-03429591e89d)
[VCMI_Client_log.zip](https://github.com/user-attachments/files/17735474/VCMI_Client_log.zip)
[VCMI_client.exe_crashinfo.zip](https://github.com/user-attachments/files/17735567/VCMI_client.exe_crashinfo.zip)

PR #4906, opened 2024-11-14
Omit unnecessary string registration for HillFort and fix popup window crash for miniHillFort

MichalZr6:


MichalZr6:
No, although I tried to do some investigation on that one. Will see more into the issue a bit later today.

lib/mapObjectConstructors/HillFortInstanceConstructor.cpp
```diff
@@ -18,7 +18,9 @@ VCMI_LIB_NAMESPACE_BEGIN
 void HillFortInstanceConstructor::initTypeData(const JsonNode & config)
 {
 	parameters = config;
-	VLC->generaltexth->registerString(parameters.getModScope(), TextIdentifier(getBaseTextID(), "unavailableUpgradeMessage"), parameters["unavailableUpgradeMessage"].String());
+	if(getModScope() != "core")
```
IvanSavenko, line 21:
```suggestion
	if(!parameters["unavailableUpgradeMessage"].isNull())
```

MichalZr6: Done :)

MichalZr6:
This PR resolves #4904.

PR #4914, opened 2024-11-16
Launcher: HD-Mod Creator

Laserlicht:
Creates HD-Mod from Steam data.

Accessible with "Install file" button. Conversion needs some time (up to 1 hour).

Depends on #4874 -> alternative for https://github.com/Laserlicht/vcmi_hd_mod

Maybe we get a full HD integration in future. But I think as mod it's a starting point.


launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+			QString imageConfig;
+			QByteArray data;
+			QByteArray dataCompressed;
+			std::vector<QByteArray> dataCompressedArray;
+
+			file.seek(nameOffset);
+			char tmpName[8];
+			in.readRawData(tmpName, 8);
+			name = QString(tmpName);
+			file.seek(file.pos() + 12); // dummy
```
kambala-decapitator, line 241:
```suggestion
			in.skipRawData(12); // dummy
```

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+				in >> chunkZsize;
+				chunkZsizeArray.push_back(chunkZsize);
+			}
+			std::vector<int32_t> chunkSizeArray;
+			for(int j = 0; j < chunks; j++)
+			{
+				int32_t chunkSize;
+				in >> chunkSize;
+				chunkSizeArray.push_back(chunkSize);
+			}
```
kambala-decapitator, line 256: extract to a function / lambda

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+			nameOffset = file.pos();
+
+			file.seek(offset);
+			char tmpImageConfig[100000] = {  };
```
kambala-decapitator, line 260: why such number?

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+				if(dataCompressed.length() == 0 || dataCompressed.at(0) != 0x78) // check for zlib magic
+					break;
+			}
+
+			if(data.length() < dataCompressedLength)
+				outData = dataCompressedArray;
+			else
+				outData.push_back(data);
+
+			if(cb)
```
kambala-decapitator, line 302: does it even make sense to call this function without callback?

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+				cb(name, imageConfig, outData);
+		}
+	}
+}
+
+QImage HdExtractor::loadDds(QByteArray & data)
+{
+	dds::Image image;
+	if(dds::readImage((uint8_t*)data.data(), data.length(), &image) == dds::ReadResult::Success)
```
kambala-decapitator, line 311: don't use C-style casts

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+	{
+		assert(image.format == DXGI_FORMAT_BC1_UNORM || image.format == DXGI_FORMAT_BC3_UNORM); // Heroes HD should use only this variant
+		assert(image.mipmaps.size() == 1);
```
kambala-decapitator, line 314: use Q_ASSERT_X instead

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+
+		auto src = image.mipmaps[0].data();
+		uchar *uncompData = new uchar[image.width * image.height * 4];
```
kambala-decapitator, line 317: what is 4? number of color components? Create a constant.

also, use unique_ptr or the same thing from Qt (QScopedPointer IIRC)

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+				}
+			}
+		}
+
+		QImage img = QImage(uncompData, image.width, image.height, image.width * 4, QImage::Format::Format_RGBA8888).copy();
+		delete[] uncompData;
+		return img;
+	}
+
+	return QImage{};
```
kambala-decapitator, line 340:
```suggestion
	return {};
```

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+			else
+				outData.push_back(data);
+
+			if(cb)
+				cb(name, imageConfig, outData);
+		}
+	}
+}
+
+QImage HdExtractor::loadDds(QByteArray & data)
```
kambala-decapitator, line 308: do you really need non-const param here?

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+	bool isSprite = vstd::contains({ArchiveType::SPRITE_X2, ArchiveType::SPRITE_X3, ArchiveType::SPRITE_LOC_X2, ArchiveType::SPRITE_LOC_X3}, type);
+	std::string folder = isSprite ? "sprites" : "data";
+	folder += vstd::contains({SubModType::X2, SubModType::LOC_X2}, subModType) ? "2x/" : "3x/";
+	if(isSprite)
+		folder += groupName.toStdString() + "/";
+
+	QByteArray imgData;
+	QBuffer buffer(&imgData);
+	buffer.open(QIODevice::WriteOnly);
+	img.save(&buffer, "png", 85);
```
kambala-decapitator, line 395: what is 85? Quality? Why not 90?

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+{
+	if (auto result = submodType) {
+		std::string scale = (*result == SubModType::X2 || *result == SubModType::LOC_X2) ? "2" : "3";
+		bool isTranslation = (*result == SubModType::LOC_X2 || *result == SubModType::LOC_X3);
+		QJsonObject mod;
+		if(isTranslation)
+		{
+			mod = QJsonObject({
+				{ "modType", "Translation" },
+				{ "name", QString::fromStdString("HD Localisation (" + language + ") (x" + scale + ")") },
```
kambala-decapitator, line 411: use QString formatting instead

launcher/modManager/extractor/hdextractor.cpp
```diff
@@ -0,0 +1,648 @@
...
+			}
+		}
+	}
+	return QImage{};
+}
+
+bool HdExtractor::ModGenerator::filter(QString groupName, ArchiveType type)
+{
+	// skip mainmenu (RoE)
+	if(vstd::contains({"MAINMENU", "GAMSELBK", "GSELPOP1", "SCSELBCK", "LOADGAME", "NEWGAME", "LOADBAR"}, boost::to_upper_copy(groupName.toStdString())))
```
kambala-decapitator, line 536: another candidate to be moved to a constant

launcher/modManager/extractor/hdextractor.h
```diff
@@ -0,0 +1,110 @@
...
+		{"ES", "spanish"},
+		{"FR", "french"},
+		{"IT", "italian"},
+		{"PL", "polish"},
+		{"RU", "russian"}
+	};
+
+	const int FILES = 71110;
+
+	Q_OBJECT
```
kambala-decapitator, line 63: move to the top

launcher/modManager/extractor/hdextractor.h
```diff
@@ -0,0 +1,110 @@
...
+
+	QWidget *parent;
+	std::function<void(float percent)> cb;
+	QDir tempDir;
+	QDir baseDir;
+	std::mutex mutex;
```
kambala-decapitator, line 69: split ivars into a separate `private:` section

launcher/modManager/extractor/hdextractor.h
```diff
@@ -0,0 +1,110 @@
...
+		std::string language;
+		std::map<std::string, std::vector<std::vector<std::string>>> sdData;
+
+		std::map<std::string, std::vector<std::string>> flagData;
+		QImage flag2x;
+		QImage flag3x;
+
+		std::shared_ptr<CZipSaver> saver;
+	public:
+		ModGenerator(QDir outputDir, std::string language, SubModType subModType);
```
kambala-decapitator, line 88: the method is public, but the class is private - this doesn't make sense. If the method doesn't depend on internal state, move the class to cpp and "convert" method to a function.

launcher/StdInc.h
```diff
@@ -11,6 +11,11 @@
 
 #include "../Global.h"
 
+#include "tbb/blocked_range.h"
+#include "tbb/parallel_for.h"
+#include "tbb/parallel_for_each.h"
```
kambala-decapitator, line 16: do you really need tbb? Check if QtConcurrent suits your needs.

kambala-decapitator (changes requested):
it's quite difficult to understand what's going on here. Please add some documentation, like description of the algorithm.

Ideally, external libraries should be proper CMake dependencies. They should also be added via Conan or at least as submodules.

Also check if you can replace tbb with QtConcurrent.

dydzio0614:
i think "hd mod" is kinda bad name that can cause confusion with hd mod for h3

Olegmods:
> i think "hd mod" is kinda bad name that can cause confusion with hd mod for h3

Just a suggestion: maybe HD-Edition-Mod will be better?

dydzio0614:
> > i think "hd mod" is kinda bad name that can cause confusion with hd mod for h3
> 
> Just a suggestion: maybe HD-Edition-Mod will be better?

or "hd edition graphics mod" or something similar

IvanSavenko:
Really not sure about this one. Sorry...

Idea is nice, sure, but:
- This is something that we should use as it, without conversion, similarly to how we use H3:Complete data without any conversion process.
- Original format (two archives with dds) is way better than massive number of extracted png's proposed here. Both because of use of archive and because of using fast to load dds instead of slow to decode png's
- Unless something has changed, conversion is extremely long, about a hour on PC, right? On mobiles this likely will be even longer. IMO this is too long to be acceptable

So not sure if this should be part of vcmi proper. Also if we were to add proper support later, we'll have some players with this version of HD edition, and some players with proper version, which would be pain to maintain.

Perhaps keep it as open PR so power users can use it on their own risk? And meanwhile, focus on proper support of HD Edition for vcmi 1.7? We're quite close to being able to use it as it.

IvanSavenko:
Also - with exporter approach we won't be able to easily fix mod later if needed. Unless we'll make another "hd edition mod upgrader" tool.

For example I believe we don't support terrain animations from HD Edition and player-colored images at the moment? If such support were to be added, how can we upgrade generated mod for players who already have it? Ask them to re-export and wait for 1 hour more? Make tool for every such upgrade?

Laserlicht:
> This is something that we should use as it, without conversion, similarly to how we use H3:Complete data without any conversion process.

Yes. But this is IMO much more complicated.
If you want to give it a try: Feel free.

> Original format (two archives with dds) is way better than massive number of extracted png's proposed here. Both because of use of archive and because of using fast to load dds instead of slow to decode png's

Also tested BMP. Make not much difference in loading times. But also using ZIP (to avoid thousands of files in filesystem. Not sure if decoding DDS is faster than PNG. DDS is usually used in combination with GPU's and optimized for them.

Process to get one sprite as image:
- Extract from PAK (decoding, decompressing)
- Decode DDS -> Decompress with BC1/BC3 -> Crop desired sprite from sheet (one sheet contains many bitmaps/sprites) -> rotate sprite to correct alignment (usually at least partially done by GPU)
- Extend sprites to match H3 sprites
- calculate overlays (e.g. flags, creature outlines)

> Unless something has changed, conversion is extremely long, about a hour on PC, right? On mobiles this likely will be even longer. IMO this is too long to be acceptable

Not sure if we need extraction process on mobile. Android/iOS version isn't in store anymore. And because of sandbox restriction not possible to access data from it. Because of this you'll need to copy data from Steam version from PC anyway. Just do the process on PC and copy files.

> Perhaps keep it as open PR so power users can use it on their own risk? And meanwhile, focus on proper support of HD Edition for vcmi 1.7? We're quite close to being able to use it as it.

For me also ok. But then I think I'll close this PR -> And prefer the Python script for Power users...

> Also - with exporter approach we won't be able to easily fix mod later if needed. Unless we'll make another "hd edition mod upgrader" tool.

Yes. Easy to detect (mod have version). But a lot of work to make an upgrader.

> For example I believe we don't support terrain animations from HD Edition and player-colored images at the moment? If such support were to be added, how can we upgrade generated mod for players who already have it? Ask them to re-export and wait for 1 hour more? Make tool for every such upgrade?

Yes terrain animations are also not supported by xbrz for now. Currently they are ignored (use xbrz fallback -> if supported in future). But to be honest: I think prescaled terrain animations doesn't look better than xbrz ones.

Player colored images works (overlay images are generated automatically (with flags / borders).

Laserlicht:
As described. Closing this. Hoping for native support in 1.7. :)

For power users in meantime: https://github.com/Laserlicht/vcmi_hd_mod

IvanSavenko:
> Player colored images works 

What about player-colored UI?

Laserlicht:
> decoding is not needed at all - pixel data is copied as it.

Yes, but at least decompressing BC1/BC3 used in Heroes III DDS. I think this is usually done by GPU. But may also fast on CPU?

> What about player-colored UI?

Skipped. HD graphics aren't better here than xbrz IMO.

IvanSavenko:
> Yes, but at least decompressing BC1/BC3 used in Heroes III DDS. I think this is usually done by GPU. But may also fast on CPU?

Yes. Its advantage is that *decompression* is extremely trivial and is implemented in hardware by GPU and can be decoded on the fly during rendering, so GPU keeps data in compressed form reducing memory usage. Optimal encoding is way more complex though.

CPU can do the same. For example 4x4 BC1 block can be decoded as:
```
for each 4x4 block:
    read 16-bit color 1
    read 16-bit color 2
    compute color 3 from colors 1&2
    compute color 4 from colors 1&2
    for each pixel in block:
        read 2-bit sized lookup index
        pick color based on lookup index
```

SoundSSGood:
> Also - with exporter approach we won't be able to easily fix mod later if needed. Unless we'll make another "hd edition mod upgrader" tool.

Maybe it's not a bad idea to have a separate tool for converting data for all hero editions. SOD, HD Edition, Chronicles... This way we can get rid of legacy and switch only to the native VCMI format. In any case, these are not plans for the near future.

Laserlicht:
> Add support for compressed mods?

Already possible and used here... instead providing a folder named "content" you can provide a zip named "content.zip". But normally not suggested (a nightmare to manage on git repositorys as "blob").

IvanSavenko:
> Maybe it's not a bad idea to have a separate tool for converting data for all hero editions. SOD, HD Edition, Chronicles... This way we can get rid of legacy and switch only to the native VCMI format. In any case, these are not plans for the near future.

Don't see any advantages of doing that. What would this offer to us or to players?

I see only disadvantages, since we'll need to maintain output of such tool. For example assume that we've exported all h3 files and applied better naming scheme, e.g. rename hero portraits filenames to include hero ID. And then we discover another `'torosar '`. Now we need another tool that would fix that and rename every such file to exclude space.

Initial import also won't be instant.

> Add support for compressed mods?

Technically - already in, mods can use "Content.zip" in place of Content folder. However at the moment it is not convenient due to how we distribute mods from repositories on Github. 

Warmonger had an idea to put everything from Content directory into archive on installing mod via Launcher, which could work, but obviously needs developer time.

And this does not eliminates png decoding & loading time. DDS textures also have memory advantage - they are used in compressed form. For example, DXT1/BC1 texture loaded into video memory needs just 4 bits / pixel. Compared to 24 bits / pixel for same RGB texture (x6 times decrease in memory use!). Similarly, DXT5 / BC3 texture needs only 8 bits / pixel, compared to 32 bits / pixel for same RGBA texture (4x decrease in memory use!).

Sure, not really relevant with modern hardware that offers like 16 GB of VRAM, but vcmi also runs on phones and on ancient PC's.

Gagert:
> As described. Closing this. Hoping for native support in 1.7. :)
> 
> For power users meanwhile: https://github.com/Laserlicht/vcmi_hd_mod

Does native support in VCMI 1.7 mean having a user friendly way of extracting the HD assets from Heroes 3 HD Edition from VCMI (auto-extracter)? Like all the 2D sprites from Heroes 3 HD Edition being extracted in say the png image format into the right directories inside the VCMI folder? In that way, users could alter every sprite in-game and then HD-ifying them (expansion 2D art)?

This just sounds way too cool, that's the reason for my enthusiasm right now.

Laserlicht:
> Does native support in VCMI 1.7 mean having a user friendly way of extracting the HD assets from Heroes 3 HD Edition to VCMI (auto-extracter)? Like all the 2D sprites from Heroes 3 HD Edition being extracted in say the png image format into the right directories inside the VCMI folder? In that way, users could alter every sprite in-game and then HD-ifying them (expansion 2D art)?

No. That's the canceled approach here.
We would prefer directly using the HD files without converting. As it should deliver better performance. But overriding them should be still possible.

But as HD functionality has merged for 1.6: You can use linked converter do make a mod from official HD. There each image could be altered easily (PNG files).
And any mod (e.g. new creatures, towns) can deliver now also HD textures.

PR #4919, opened 2024-11-16
Remove '-Og' from MINGW debug builds

Kuxe:
Remove `-Og` from MINGW debug builds. It interferes with debugging. See https://stackoverflow.com/a/63386263/2323908. Instead, the default O0 is used.

This change makes it possible to use whatever optimization levels you want debug builds, which was previously impossible due to the passing of `-Og` in CMakeLists.txt.

Without this PR, breakpoints won't trigger on exact lines, several variables get optimized out, and generally makes it very difficult to debug with mingw builds.

The comments that motivates the addition of `-Og`
```cmake
# Prevent compiler issues when building Debug
# Assembler might fail with "too many sections"
# With big-obj or 64-bit build will take hours
```
does not hold any longer.
* It is possible to compile, without issues, without -Og
* I do not observe "too many sections" problems when compiling
* Compilation does not take hours

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -355,13 +355,6 @@ if(MINGW OR MSVC)
 		if(ICONV_FOUND)
 			set(SYSTEM_LIBS ${SYSTEM_LIBS} iconv)
 		endif()
-
-		# Prevent compiler issues when building Debug
-		# Assembler might fail with "too many sections"
-		# With big-obj or 64-bit build will take hours
-		if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-			set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Og")
-		endif()
 	endif(MINGW)
 endif(MINGW OR MSVC)
 


kambala-decapitator (approved):

Kuxe:
I had not built all targets when I concluded that the "too many sections" problem didn't appear - it does surface. Investigating if possible to get rid of this problem *without* `-Og`

Kuxe:
@kambala-decapitator We should rollback this PR until I sort it out

IvanSavenko:
Quick search suggests `-Wa,-mbig-obj` flag
Can you try it? You're probably the only one who uses mingw on Windows.
Some also suggest `-flto` but no idea how it can have any effect.

IvanSavenko:
What causes it btw? Some specific cpp files in our code?

Kuxe:
> Quick search suggests `-Wa,-mbig-obj` flag Can you try it? You're probably the only one who uses mingw on Windows. Some also suggest `-flto` but no idea how it can have any effect.

Yep, the original comments on why `-Og` was present to begin with explicitly mentions that `big-obj` add hours to compile time. May still be valid, haven't tested myself.

There's at least one other person running mingw on Windows, some timestamp bug was reported for mingw builds on windows, but yeah - might not need to rollback :)

> What causes it btw? Some specific cpp files in our code?

`SerializerReflection.cpp`. A (preferable?) solution would be to shrink the object file size for said file. I'm gonna try attempt that approach later this evening

IvanSavenko:
`SerializerReflection.cpp`? Might be easier to try that big-obj flag.
This is where most of our template-based serializer is instantiated.
For me it compiles with no issues, however object size is 30 Mb - about 10 times more than rest of object files.

Not sure how can it be split easily though. Easiest option is probably try to split registerTypes() method in several chunks. Was actually split before, around 1.3, but got merged alongsize with simplification of serializer. Apparently was not simplified enough.

Another option is to reduce number of methods instantiation in that template magic. But not sure how easy it will be.

Kuxe:
Also `IGameCallback.cpp` give issues under MingW without `-Og`.

So, adding `-Og` to these files makes them sufficiently small for MinGW builds to compile again.

* `SerializerReflection.cpp` (25.9 M B)
* `IGameCallback.cpp` (17.3 MB)

PR inc...

IvanSavenko:
Perhaps `-Wa,-mbig-obj` is better option for these two files?

`-Og` working as a fix sounds like just a coincidence, since those optimizations reduced object file size. And if that option breaks debugging, then debugging anything from these two files might be impossible.

Kuxe:
https://github.com/vcmi/vcmi/pull/4923

Ill update that PR

PR #4922, opened 2024-11-16
[FIX] angels are not doubleWide

abd-msyukyu-odoo:


krs0:
That one is on Chat-gpt :)
Ty for the fix.

PR #4927, opened 2024-11-17
Prelauncher

Laserlicht:
- Optional (enable setting in launcher options)
- Mainly to fix issue that android gamepad user needs to use mouse just for launching VCMI every time
  - now mouse is only needed for first launch and settings/mod-managment, but not for starting (if prelauncher is enabled)

Related: #4111 

grafik

### DIFF
--- a/launcher/mainwindow_moc.cpp
+++ b/launcher/mainwindow_moc.cpp
@@ -125,6 +125,7 @@ MainWindow::MainWindow(QWidget * parent)
 
 void MainWindow::skipLauncher()
 {
+#ifdef ENABLE_SKIP_LAUNCHER
 	QMessageBox msgBox;
 	int secondsRemaining = 10;
 	auto getText = [&] { return tr("How to start?\n\nVCMI starts automatically in %n second(s)", nullptr, secondsRemaining); };
@@ -154,6 +155,7 @@ void MainWindow::skipLauncher()
 		hide();
 		startGame({});
 	}
+#endif
 }
 
 void MainWindow::detectPreferredLanguage()
--- a/launcher/mainwindow_moc.cpp
+++ b/launcher/mainwindow_moc.cpp
@@ -127,14 +127,14 @@ void MainWindow::preLauncher()
 {
 	QMessageBox msgBox;
 	int secondsRemaining = 10;
-	auto setText = [&](){ msgBox.setText(tr("How to start?\n\nVCMI starts automatically in %n second(s)", "", secondsRemaining)); };
-	setText();
+	auto getText = [&] { return tr("How to start?\n\nVCMI starts automatically in %n second(s)", nullptr, secondsRemaining); };
+	msgBox.setText(getText());
 	msgBox.setWindowTitle("VCMI");
 	msgBox.setIcon(QMessageBox::Question);
 	msgBox.addButton(tr("Start Launcher"), QMessageBox::YesRole);
-	QPushButton * startClient = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
-	QTimer *timer = new QTimer(this);
-	connect(timer, &QTimer::timeout, this, [&, this](){
+	auto startClient = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
+	auto timer = new QTimer(this);
+	connect(timer, &QTimer::timeout, this, [&, this, timer](){
 		if(secondsRemaining == 0)
 		{
 			timer->stop();
@@ -143,7 +143,7 @@ void MainWindow::preLauncher()
 			startGame({});
 			return;
 		}
-		setText();
+		msgBox.setText(getText());
 		secondsRemaining--;
 	});
 	timer->start(1000);
--- a/launcher/mainwindow_moc.cpp
+++ b/launcher/mainwindow_moc.cpp
@@ -116,8 +116,40 @@ MainWindow::MainWindow(QWidget * parent)
 
 	ui->settingsView->setDisplayList();
 	
+	if(settings["launcher"]["preLauncher"].Bool())
+		preLauncher();
+
 	if(settings["launcher"]["updateOnStartup"].Bool())
 		UpdateDialog::showUpdateDialog(false);
+}
+
+void MainWindow::preLauncher()
+{
+	QMessageBox msgBox;
+	int secondsRemaining = 10;
+	QString text = tr("How to start?\n\nVCMI starts automatically in ");
+	msgBox.setText(text + QString::number(secondsRemaining) + tr(" Seconds"));
+	msgBox.setWindowTitle(tr("VCMI"));
+	msgBox.setIcon(QMessageBox::Information);
+	msgBox.addButton(tr("Start Launcher"), QMessageBox::YesRole);
+	QPushButton * startClient = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
+	QTimer *timer = new QTimer(this);
+    connect(timer, &QTimer::timeout, this, [this, &timer, &text, &msgBox, &secondsRemaining](){
+		if(secondsRemaining == 0)
+		{
+			timer->stop();
+			msgBox.close();
+			ui->startGameButton->click();
+			return;
+		}
+		msgBox.setText(text + QString::number(secondsRemaining) + tr(" Seconds"));
+		secondsRemaining--;
+	});
+    timer->start(1000);
+	msgBox.exec();
+	timer->stop();
+	if(msgBox.clickedButton() == startClient && secondsRemaining)
+		ui->startGameButton->click();
 }
 
 void MainWindow::detectPreferredLanguage()
--- a/launcher/mainwindow_moc.cpp
+++ b/launcher/mainwindow_moc.cpp
@@ -116,7 +116,7 @@ MainWindow::MainWindow(QWidget * parent)
 
 	ui->settingsView->setDisplayList();
 	
-	if(settings["launcher"]["preLauncher"].Bool())
+	if(settings["launcher"]["preLauncher"].Bool()) // MessageBox before launcher asking how to start (automatically starts client after 10 seconds)
 		preLauncher();
 
 	if(settings["launcher"]["updateOnStartup"].Bool())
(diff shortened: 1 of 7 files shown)


launcher/settingsView/csettingsview_moc.ui
```diff
@@ -1449,6 +1449,35 @@ Fullscreen Exclusive Mode - game will cover entirety of your screen and will use
          </attribute>
         </widget>
        </item>
+       <item row="61" column="0">
+        <widget class="QLabel" name="labelPreLauncher">
+         <property name="text">
+          <string>Prelauncher</string>
```
IvanSavenko, line 1455: Prelauncher is very unclear IMO.
Not sure what to call it though. If somebody has better suggestions than this - go ahead.
```suggestion
          <string>Skip Launcher on start</string>
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,43 @@ MainWindow::MainWindow(QWidget * parent)
...
 		UpdateDialog::showUpdateDialog(false);
 }
 
+void MainWindow::preLauncher()
+{
+	QMessageBox msgBox;
+	int secondsRemaining = 10;
+	QString text = tr("How to start?\n\nVCMI starts automatically in ");
+	QString textSeconds = tr(" Seconds");
+	msgBox.setText(text + QString::number(secondsRemaining) + textSeconds);
```
IvanSavenko, line 132:
```suggestion
    QString textTemplate = tr("How to start?\n\nVCMI starts automatically in %1 seconds");
	msgBox.setText(textTemplate.arg(secondsRemaining));
```

Can be improved further - to allow plural translations (1 second, 2 seconds) but not sure about Qr syntax for that

IvanSavenko (commented):
@kambala-decapitator any comments?

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,42 @@ MainWindow::MainWindow(QWidget * parent)
...
 	if(settings["launcher"]["updateOnStartup"].Bool())
 		UpdateDialog::showUpdateDialog(false);
 }
 
+void MainWindow::preLauncher()
+{
+	QMessageBox msgBox;
+	int secondsRemaining = 10;
+	auto text = "How to start?\n\nVCMI starts automatically in %n second(s)";
+	msgBox.setText(tr(text, "", secondsRemaining));
```
kambala-decapitator, line 131: `tr` doesn't work with variables

```suggestion
	msgBox.setText(tr("How to start?\n\nVCMI starts automatically in %n second(s)", nullptr, secondsRemaining));
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,42 @@ MainWindow::MainWindow(QWidget * parent)
...
+	msgBox.setWindowTitle(tr("VCMI"));
```
kambala-decapitator, line 132: why do you want to translate it?

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,42 @@ MainWindow::MainWindow(QWidget * parent)
...
+	msgBox.setIcon(QMessageBox::Information);
```
kambala-decapitator, line 133:
```suggestion
	msgBox.setIcon(QMessageBox::Question);
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,42 @@ MainWindow::MainWindow(QWidget * parent)
...
+	msgBox.addButton(tr("Start Launcher"), QMessageBox::YesRole);
+	QPushButton * startClient = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
+	QTimer *timer = new QTimer(this);
+	connect(timer, &QTimer::timeout, this, [this, &timer, &text, &msgBox, &secondsRemaining](){
```
kambala-decapitator, line 137:
```suggestion
	connect(timer, &QTimer::timeout, this, [this, timer, &] {
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,42 @@ MainWindow::MainWindow(QWidget * parent)
...
+		if(secondsRemaining == 0)
+		{
+			timer->stop();
+			msgBox.close();
+			ui->startGameButton->click();
```
kambala-decapitator, line 142: call the actual handler instead of simulating click

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,42 @@ MainWindow::MainWindow(QWidget * parent)
...
+			return;
+		}
+		msgBox.setText(tr(text, "", secondsRemaining));
+		secondsRemaining--;
+	});
+	timer->start(1000);
+	msgBox.exec();
+	timer->stop();
+	if(msgBox.clickedButton() == startClient && secondsRemaining)
+		ui->startGameButton->click();
```
kambala-decapitator, line 152: same

kambala-decapitator (changes requested):
- preLauncher is not a good name indeed
- this is needed only for Android, please enclose new functionality in ifdef

I don't really like the current implementation. Makes no sense to create MainWindow just to show the message box. As a quick solution it would do, but better make the message box a "starting" widget that would then create MainWindow if needed. For example, you could subclass QApplication and implement this logic there.

Ideally, the dialog should be implemented in Java and would never even start Qt activity if not needed. But that might not be easy to implement.

launcher/mainwindow_moc.cpp
Laserlicht: Then adding `tr("How to start?\n\nVCMI starts automatically in %n second(s)", nullptr, secondsRemaining)` twice in source?

kambala-decapitator: better use lambda to create the string

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,46 @@ MainWindow::MainWindow(QWidget * parent)
...
+
 	if(settings["launcher"]["updateOnStartup"].Bool())
 		UpdateDialog::showUpdateDialog(false);
 }
 
+void MainWindow::preLauncher()
+{
+	QMessageBox msgBox;
+	int secondsRemaining = 10;
+	auto setText = [&](){ msgBox.setText(tr("How to start?\n\nVCMI starts automatically in %n second(s)", "", secondsRemaining)); };
```
kambala-decapitator, line 130: it's better to just return a string from lambda and set it at the calling site:

```suggestion
	auto getText = [&] { return tr("How to start?\n\nVCMI starts automatically in %n second(s)", nullptr, secondsRemaining); };
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,46 @@ MainWindow::MainWindow(QWidget * parent)
...
+	setText();
+	msgBox.setWindowTitle("VCMI");
+	msgBox.setIcon(QMessageBox::Question);
+	msgBox.addButton(tr("Start Launcher"), QMessageBox::YesRole);
+	QPushButton * startClient = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
```
kambala-decapitator, line 135:
```suggestion
	auto startClientButton = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,46 @@ MainWindow::MainWindow(QWidget * parent)
...
+	QTimer *timer = new QTimer(this);
```
kambala-decapitator, line 136:
```suggestion
	auto timer = new QTimer(this);
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,46 @@ MainWindow::MainWindow(QWidget * parent)
...
+	connect(timer, &QTimer::timeout, this, [&, this](){
```
kambala-decapitator, line 137:
```suggestion
	connect(timer, &QTimer::timeout, this, [this, timer, &] {
```

launcher/mainwindow_moc.cpp
```diff
@@ -116,10 +116,46 @@ MainWindow::MainWindow(QWidget * parent)
...
+		if(secondsRemaining == 0)
+		{
+			timer->stop();
+			msgBox.close();
+			hide();
```
kambala-decapitator, line 142: does hiding improve something?

Laserlicht: Hiding mouse for intro :)

launcher/mainwindow_moc.cpp
```diff
@@ -127,14 +127,14 @@ void MainWindow::preLauncher()
...
-	setText();
+	auto getText = [&] { return tr("How to start?\n\nVCMI starts automatically in %n second(s)", nullptr, secondsRemaining); };
+	msgBox.setText(getText());
 	msgBox.setWindowTitle("VCMI");
 	msgBox.setIcon(QMessageBox::Question);
 	msgBox.addButton(tr("Start Launcher"), QMessageBox::YesRole);
-	QPushButton * startClient = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
-	QTimer *timer = new QTimer(this);
-	connect(timer, &QTimer::timeout, this, [&, this](){
+	auto startClient = msgBox.addButton(tr("Start VCMI"), QMessageBox::NoRole);
```
kambala-decapitator, line 135: please give proper names to variables: startClientButton

kambala-decapitator (commented):
so, what about this?

> preLauncher is not a good name indeed
> this is needed only for Android, please enclose new functionality in ifdef

IvanSavenko:
> portable gaming consoles (linux/windows -> there are some without touchscreen/touchpad out there)
pc (all OS) connected to TV

- Such systems most likely have mapping of controller input, e.g. sticks or D-pad to mouse input.
- You still need a way to launch VCMI itself, e.g. double-click on VCMI_Launcher.exe or any equivalent. So you should also be able to change whatever you need in Launcher and start client from it.
- On all PC OS's you can always start VCMI_Client directly for the same effect

What about hiding all this functionality behind a separate define, e.g. ENABLE_SKIP_LAUNCHER, and by default - enable it only on Android?

Laserlicht:
>     * On all PC OS's you can always start VCMI_Client directly for the same effect

Yes you're right. Would be easier to add this one in player owned launcher.
Only our flatpak doesn't expose it seperatly atm (but should be addable over command line args).

> What about hiding all this functionality behind a separate define, e.g. ENABLE_SKIP_LAUNCHER, and by default - enable it only on Android?

What's the advantage over current solution? I think even most of android users are fine with starting in launcher.

IvanSavenko:
> What's the advantage over current solution? I think even most of android users are fine with starting in launcher.

I meant in addition to current solution.

So this option will be available only on Android (where there is known use-case for it), and toggled to false by default. While for all other platforms it will be disabled / not compiled in. 

And if we decide that another platform needs it or we want to enable it everywhere we can adjust that single define instead of searching for all `#ifdef VCMI_ANDROID` related to this feature.

Laserlicht:
> I meant in addition to current solution.

Yeah, but still don't understand the point. :D

Adding code to save maybe 100 Bytes in binary and to disallow usage of an already optional feature for non android users?

I mean. This feature is also working on PC. Maybe also someone prefer it like this. And if not - there is also no problem.

kambala-decapitator:
> Yes we could do a separate window/view. But I think it would be overengineered for this little functionality.

writing clean code may often feel like overengineering, but in the end you get benefits :)

> What about hiding all this functionality behind a separate define, e.g. ENABLE_SKIP_LAUNCHER, and by default - enable it only on Android?

sounds good. IMO this feature is useless everywhere (for reasons already given) except Android.

IvanSavenko:
> Yeah, but still don't understand the point. :D

The way I see it, this is a workaround for an Android-specific issue that (hopefully) will be unnecessary and safe to remove once somebody adds a way to use controller in Launcher on Android. And not an independent feature.

So from this point of view, having it Android-specific makes sense.

PR #4934, opened 2024-11-17
Fix for missing creature abilities 

krs0:
Some abilities were missing for siege weapons and fly for Fairy Dragon.

### DIFF
--- a/config/creatures/castle.json
+++ b/config/creatures/castle.json
@@ -346,11 +346,6 @@
 		"faction": "castle",
 		"abilities":
 		{
-			"KING_2" : // Will be affected by Advanced Slayer or better
-			{
-				"type" : "KING",
-				"val" : 2
-			},
 			"canFly" :
 			{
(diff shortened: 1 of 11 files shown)


config/creatures/castle.json
```diff
@@ -346,11 +346,6 @@
 		"faction": "castle",
 		"abilities":
 		{
-			"KING_2" : // Will be affected by Advanced Slayer or better
```
SoundSSGood, line 349: Does this mean that the fields in json now have to follow a certain order?

IvanSavenko: Just for reading convenience.
Game does not retains this information at all, it is all lost during json parsins

krs0: _it is all lost during json parsing_

*unfortunately it is all lost during json parsing. :)

Hopefully sometime in the future the order could be preserved. The information would be presented more player friendly.

PR #4939, opened 2024-11-18
Add 'Open Recent' to mapeditor

Kuxe:
Add 'Open Recent' to mapeditor

https://github.com/user-attachments/assets/519351cb-9795-46fd-950d-5d290d5d563e

**Implementation**

Works identically to the "Open" action but with CTRL+R as hotkey. The callback itself retrieves a list

```c++
	QSettings s(Ui::teamName, Ui::appName);
	QStringList recentFiles = s.value(recentlyOpenedFilesSetting).toStringList();
```

and then uses that list to open a dialog. Double-clicking an item in the dialog, or just "activating" an item via <kbd>Enter</kbd> also triggers the `openMap` function. The `openMap` function is updated to update the `recentlyOpenedFilesSetting`-setting.


kambala-decapitator:
I'd rather go with a more traditional approach by placing Recents in the File menu

IvanSavenko:
Agree with kambala.
Feature is nice, but why not make it both in line with H3 editor and with most of software that implements such feature? Namely - show recent files in "File" menu. Either directly in this menu, or in "Recent Files" submenu.

Kuxe:
> I'd rather go with a more traditional approach by placing Recents in the File menu

It's a mistake with the recording, seems like there is a slight delay from "Start Recording" to it actually starting... The first occurrence of the "Recently Opened Files"-dialog was opened by `File` - `Open Recent` which is placed just under `Open` in the dropdown menu.

Kuxe:
This shows the `File` -> `Open Recent` too

https://github.com/user-attachments/assets/3dec3eb9-b06f-4dd0-9a39-c3cf9b74da40

Kuxe:
Oh, I see what you mean. That was my first attempt but I couldn't figure out how to get the submenu working in Qt. In VSCode, CTRL+R maps to "Open Recent" action which creates a popup dropdown menu.

Maybe it's best to align with H3 Map Ed here, which also supports keyboard driven workflow via <kbd>ALT</kbd>+<kbd>F</kbd> + "arrow keys to select recent file" + <kbd>Enter</kbd>. A bit more cumbersome than CTRL+R, which many editors support.

How about this: I can take a stab at getting the submenu working in addition to existing functionality, highlight what I mean by "submenu" in your pic just to be clear.
image

In the end this means that CTRL+R would work as-is, but opening `File` and selecting "Open Recent" would *not* open the dialog, but instead expand a submenu? This is essentially what VSCode does.

Kuxe:
Submenu is added. It is dynamically populated with recently opened maps. There's a "More..."-button with CTRL+R shortcut that opens up a "Recently Opened File"-dialog. This approach is inspired from VSCode.

https://github.com/user-attachments/assets/6d64aa33-f94b-4f67-82ba-d0237afa1610

IvanSavenko:
Looks OK for me.
Not sure if button on toolbar is needed for this feature, but not against it.

Kuxe:
Anything else to be done here?

mapeditor/mainwindow.cpp
```diff
@@ -16,6 +16,9 @@
 #include <QFile>
 #include <QMessageBox>
 #include <QFileInfo>
+#include <qdialog.h>
+#include <qlistwidget.h>
+#include <qnamespace.h>
```
kambala-decapitator, line 21: please use the extension-less includes

mapeditor/mainwindow.cpp
```diff
@@ -412,9 +417,21 @@ bool MainWindow::openMap(const QString & filenameSelect)
 	
 	filename = filenameSelect;
 	initializeMap(controller.map()->version != EMapFormat::VCMI);
+
+	updateRecentMenu(filenameSelect);
+
 	return true;
 }
 
+void MainWindow::updateRecentMenu(const QString & filenameSelect) {
```
kambala-decapitator, line 426: Most Qt core classes implement COW, no need to pass them via const ref
```suggestion
void MainWindow::updateRecentMenu(QString filenameSelect)
{
```

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
 	openMap(filenameSelect);
 }
 
+void MainWindow::on_actionOpenRecent_triggered()
+{
+	QSettings s(Ui::teamName, Ui::appName);
+	QStringList recentFiles = s.value(recentlyOpenedFilesSetting).toStringList();
+
+	struct RecentFileDialog : public QDialog {
```
kambala-decapitator, line 454: don't declare such things inside a function, rather place it in anonymous namespace.

also, it should be a class as QDialog is class.

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+		
+		RecentFileDialog(QWidget *parent, const QStringList& recentFiles) : QDialog(parent), layout(this) {
```
kambala-decapitator, line 456: parent traditionally goes last
```suggestion
		RecentFileDialog(QStringList recentFiles, QWidget *parent = nullptr) : QDialog(parent), layout(this)
		{
```

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+
+			setWindowTitle("Recently Opened Files");
+			setMinimumWidth(600);
+
+			for (const QString &file : recentFiles) {
```
kambala-decapitator, line 461:
```suggestion
			for (QString file : recentFiles)
			{
```

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+			if(listWidget.count() > 0) {
+				listWidget.item(0)->setSelected(true);
+			}
+
+			layout.setSizeConstraint(QLayout::SetMaximumSize);
+			layout.addWidget(&listWidget);
+		}
+
+		QVBoxLayout layout;
+		QListWidget listWidget;
```
kambala-decapitator, line 477: everything related to widgets should normally be created on heap and with proper parents
```suggestion
		QVBoxLayout * layout;
		QListWidget * listWidget;
```

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+void MainWindow::on_actionOpenRecent_triggered()
+{
+	QSettings s(Ui::teamName, Ui::appName);
+	QStringList recentFiles = s.value(recentlyOpenedFilesSetting).toStringList();
+
+	struct RecentFileDialog : public QDialog {
+		
+		RecentFileDialog(QWidget *parent, const QStringList& recentFiles) : QDialog(parent), layout(this) {
+
+			setWindowTitle("Recently Opened Files");
```
kambala-decapitator, line 458: all user-visible strings should be localized
```suggestion
			setWindowTitle(tr("Recently Opened Files"));
```

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+
+	auto onSelect = [this, &d](QListWidgetItem *item) {
+		if(getAnswerAboutUnsavedChanges()) {
+			QString filename = item->text();
+			openMap(filename);
+		}
+		d.close();
+	};
+
+	connect(&d.listWidget, &QListWidget::itemActivated, this, onSelect);
```
kambala-decapitator, line 490: rather expose a signal from RecentFileDialog that sends file path, no need to leak internal implementation outside

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+		QVBoxLayout layout;
+		QListWidget listWidget;
+	};
+
+	RecentFileDialog d(this, recentFiles);
+
+	auto onSelect = [this, &d](QListWidgetItem *item) {
+		if(getAnswerAboutUnsavedChanges()) {
+			QString filename = item->text();
+			openMap(filename);
```
kambala-decapitator, line 485:
```suggestion
			QString filePath = item->text();
			openMap(filePath);
```

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+
+void MainWindow::on_menuOpenRecent_aboutToShow()
+{
+	// Clear all actions except "More...", lest the list will grow with each
+	// showing of the list
+	for (QAction* action : ui->menuOpenRecent->actions()) {
+		if (action != ui->actionOpenRecentMore) {
+			ui->menuOpenRecent->removeAction(action);
+		}
+	}
```
kambala-decapitator, line 503: alternatively, you could simply assemble the menu from scratch

mapeditor/mainwindow.cpp
```diff
@@ -429,6 +446,87 @@ void MainWindow::on_actionOpen_triggered()
...
+	// Finally add a separator between recent entries and "More..."
+	if(recentFiles.size() > 0) {
+		ui->menuOpenRecent->insertSeparator(ui->actionOpenRecentMore);
+	}
+}
+
+void MainWindow::on_actionOpenRecentMore_triggered()
+{
+	on_actionOpenRecent_triggered();
+}
```
kambala-decapitator, line 528: just connect to the called slot directly

kambala-decapitator (changes requested):

mapeditor/mainwindow.cpp
kambala-decapitator: also not sure what qnamespace.h is for

mapeditor/mainwindow.cpp
kambala-decapitator: actually, a cleaner way would be to store selection internally and exposing it via a "getter", closing the dialog on its own. And the caller would use it like:
```qt
RecentFileDialog d{recentFiles, this};
if (d.exec())
  openMap(d.selectedFilePath());
```

mapeditor/mainwindow.cpp
Kuxe: All functions in MainWindow.cpp pass QString by const reference. Might be out of scope of this PR to break consistency with the rest of the code. The intent is also to say that the QString is expected not to be modified, so it should be const.

mapeditor/mainwindow.cpp
Kuxe: Same comment as previous post: https://github.com/vcmi/vcmi/pull/4939#discussion_r1855459520

mapeditor/mainwindow.cpp
Kuxe: struct -> class fixed with 2678bf1, will change from local function to either anonymous namespace or just make it a reusable recent-file dialog with own header in a follow-up commit

Kuxe: > will change from local function to either anonymous namespace or just make it a reusable recent-file dialog with own header in a follow-up commit

Before I do this, what's the reason for making it an anonymous class in favor of a local class? The class is not meant to be used anywhere else outside of `on_actionOpenRecent_triggered`.

mapeditor/mainwindow.cpp
Kuxe: Good suggestions! Fixed with [5747781](https://github.com/vcmi/vcmi/pull/4939/commits/5747781dda52d97f17765603b96c15ab5597e350).

mapeditor/mainwindow.cpp
Kuxe: Fixed with [e3b520a](https://github.com/vcmi/vcmi/pull/4939/commits/e3b520a1f10a844488ba9e8250b3d70c3d866d52)

mapeditor/mainwindow.cpp
Kuxe: Committed suggestion with [5bff1ec](https://github.com/vcmi/vcmi/pull/4939/commits/5bff1ec2c1b1ef045d92795152c0d088b1a1699f), accedentially undid it with [2678bf1](https://github.com/vcmi/vcmi/pull/4939/commits/2678bf1e05a89a694c39ca8f3f8f78ee3a5d260e) while rebasing, re-introduced with [9b0e676](https://github.com/vcmi/vcmi/pull/4939/commits/9b0e6763e76c54b817c0854d2b7eeec48a8bdf59).

mapeditor/mainwindow.cpp
Kuxe: Fixed with [d1e4041](https://github.com/vcmi/vcmi/pull/4939/commits/d1e40411dda31c6aa73bae84fefd81fb6b28920c)

mapeditor/mainwindow.cpp
Kuxe: Fixed with [bdaaa0b](https://github.com/vcmi/vcmi/pull/4939/commits/bdaaa0b58425fe42ce92c016ac0c90c2f9d20aca)

mapeditor/mainwindow.cpp
Kuxe: Resolving discussion, please re-open it if you have further remarks.

IvanSavenko:
@kambala-decapitator any further comments?

kambala-decapitator:
> @kambala-decapitator any further comments?

Will check today

mapeditor/mainwindow.cpp
kambala-decapitator: a good policy is to write new code properly, even if it "violates" style of the old code. Not insisting on changing it in this PR then, but let's prefer this approach in the future.

if you pass by value, original object isn't modified as well.

mapeditor/mainwindow.cpp
kambala-decapitator: local class makes it more difficult to read a function IMO

kambala-decapitator (approved):

mapeditor/mainwindow.cpp
Kuxe: I'll fix it for this PR

mapeditor/mainwindow.cpp
Kuxe: fair enough, I'll move it out to an anonymous namespace

PR #4954, opened 2024-11-21
Curved roads

DjWarmonger:
Real roads have curves, you know.

### DIFF
--- a/lib/rmg/RmgPath.h
+++ b/lib/rmg/RmgPath.h
@@ -21,7 +21,7 @@ namespace rmg
 class Path
 {
 public:
-	typedef std::function<float(const int3 &, const int3 &)> MoveCostFunction;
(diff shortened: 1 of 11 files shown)


lib/rmg/RmgPath.h
```diff
@@ -21,7 +21,8 @@ namespace rmg
 class Path
 {
 public:
-	const static std::function<float(const int3 &, const int3 &)> DEFAULT_MOVEMENT_FUNCTION;
+	typedef std::function<float(const int3 &, const int3 &)> MoveCostFunction;
```
IvanSavenko, line 24:
```suggestion
	using MoveCostFunction = std::function<float(const int3 &, const int3 &)>;
```

IvanSavenko (approved):

PR #4958, opened 2024-11-22
test macOS ARM runners

kambala-decapitator:


kambala-decapitator:
to start using ARM runner for macOS ARM build, we should build dependencies on ARM runner first:
```
CMake Error at conan-generated/cmakedeps_macros.cmake:39 (message):
  Library 'Qt5Designer' not found in package.  If 'Qt5Designer' is a system
  library, declare it with 'cpp_info.system_libs' property
  conan-generated/Qt5-Target-release.cmake:24 (conan_package_library_targets)
  conan-generated/Qt5Targets.cmake:26 (include)
  conan-generated/Qt5Config.cmake:16 (include)
  CMakeLists.txt:508 (find_package)
```

the issue comes from the fact that currently Qt is cross-compiled from Intel runner and doesn't create this target.

kambala-decapitator:
TODO: Xcode 15 throws warning

> /Users/runner/work/vcmi/vcmi/lib/texts/TextOperations.cpp:202:28: error: 'codecvt_utf8<char32_t>' is deprecated [-Werror,-Wdeprecated-declarations]

which fails build

kambala-decapitator:
Xcode 16 build works fine on iOS though

IvanSavenko:
> /Users/runner/work/vcmi/vcmi/lib/texts/TextOperations.cpp:202:28: error: 'codecvt_utf8<char32_t>' is deprecated [-Werror,-Wdeprecated-declarations]

This looks like a mess...
```
class codecvt_utf8
(since C++11)
(deprecated in C++17)
(removed in C++26)
```
There is also this one, but:
```
std::codecvt<char32_t, char, std::mbstate_t>
(since C++11)
(deprecated in C++20)
```
Probably better to remove its usage and replace with something like this:
```cpp
for (size_t i=0; i<text.size(); i += getUnicodeCharacterSize(text[i]))
    charactersCount += 1;
```

kambala-decapitator:
I remember reading recommendation to stop using codecvt in user code and rely on external libs like ICU

> getUnicodeCharacterSize

if we already have it, then it'd probably work

IvanSavenko:
> if we already have it, then it'd probably work

Yep. Defined in same file / same namespace.

ICU is probably overkill. Sure, it would support some edge cases that we may not support, but I doubt that (for example) our ttf render / SDL_ttf can handle all those details anyway. And I never heard about any of languages we use hitting those edge cases.

kambala-decapitator:
btw clang-18 from NDK r27 also throws this warning

> ICU is probably overkill

I agree. And we need only UTF encodings, right?

IvanSavenko:
> I agree. And we need only UTF encodings, right?

Not exactly.
For all text processing we use only UTF-8.
However we also need to import strings from H3 text files or from h3m's that come in variety of encodings - win125X + chinese one (GBK was it?). Those strings are converted to utf-8 immediately upon loading, but we still need to perform conversion.

At the moment we use boost::locale for encoding conversions. Not sure which backend our boost::locale uses though. iconv I think?

kambala-decapitator:
yes, should be libiconv on most (or all) platforms. Also just thought about using libiconv directly

Issue #4733, opened 2024-10-07
Modding: Add times limit for a spell


by003:
**Describe your proposal**
In HotA, Eovacius's specialty is creating two clones in the first time, not always.
And in ToW mod, there are some magic staffs which can allow hero to cast free spells.

My suggestion is making a limitation for some special spells.

For example, set a parameter:
`"spellLimitForOneBattle": 1`

Alucard648:
Imagine 7-day cooldown for Town Portal spell.

by003:
Actually Dimension Door is still not configurable for limit times.
Limit times should be used in some modding adventure spells as well, but for one day instead of one battle.
For example, if someone create a spell to gain extra move points, it needs to be used only for one or two times per day.

val-gaav:
DD number of spellcasts  is configurable with this parameter 			"power" : 1  

It is just not documented, but you can get it with /get config  from vcmi in game chat. This will generate all the jsons with current game configs.

by003:
> `number of spell casts per day is configurable with this parameter in json : "power" : 1`

Is this line configured here? However, I'd like to have the function to cast spell per battle. 
Like Eovacius's specialty.
```
// Mandatory. Spell level, value in range 1-5, or 0 for abilities
		"level": 1,

		// Mandatory. Base power of the spell
		"power": 10,
	
		// Mandatory. Default chance for this spell to appear in Mage Guilds
		// Used only if chance for a faction is not set in gainChance field
		"defaultGainChance": 0, 
```

GeorgeK1ng:
Yes, this feature is really needed for [Adventure Spells Pack](https://github.com/vcmi-mods/adventure-spells-pack), because you can cast them multiple times for now, but if you got bonus for week, spell should have defined it's cooldown to prevent multicasting. For example Empathy - Basic needs 1 day colldown, Advanced and Expert 7 days cooldown. According to documentation, this kind of configuration is not possible at the moment. But there is already defined duration ONE_DAY, ONE_WEEK. Now just needs to limit them to cast it only once in casted period.

![image](https://github.com/user-attachments/assets/5559b770-331c-45aa-900b-9babb326eebb)

PR #4966, opened 2024-11-23
Add "special" property for secondary skills

DjWarmonger:
They will be banned by default


by003:
Will you try to make a spell limit time (#4733 ) for mods?
I plan to use it in HotA mod, Eovacius's specialty,

DjWarmonger:
No, I just did this one as it was very quick feature.

PR #4975, opened 2024-11-24
third upgrade for 8th dwelling

Laserlicht:
As discussed on discord.

Possibility to use also 3th to 5th upgrade for 8th creature.

ID 150 - 154 are reserved for 8th creature.

@by003 please test it.


by003:
thanks I will test if merged

PR #4989, opened 2024-11-26
Fixes for unfinished items from AI pull request review

IvanSavenko:
Technically should have been done as part of AI pull request review, but wanted to avoid postponing PR.
List of changes:
- Removed copy-pasted code in BattleAI spellcast evaluation - no longer needed due to changes in develop branch
- Reverted leftover hotfix that is now properly fixed in develop
- Unhardcoded weekly bonus income for AI, can now be configured in game settings.
- Moved pathfinder parameters to nkai config. According to nullkiller, this may lead to worse performance, so would like to make it easy to configure / change
- Reverted change to submodule, likely cause by git not updating submodules automatically.


dydzio0614:
I would unhardcode this as well... probably as true/false rather than making it tied to difficulty level https://github.com/vcmi/vcmi/blob/develop/AI/Nullkiller/Engine/Nullkiller.cpp#L65

IvanSavenko:
Not in this PR.
Would like to add per-difficulty configuration for nkai first. 
Yes, I know that you want map reveal to be gone completely, but this config has bunch of values that might be useful for difficulty/performance configuration, so from my point of view:
1. Allow per-difficulty properties in nkai-config, with no changes to AI behavior
2. Decide what to do with map reveal
3. Consider making other options different depending on difficulty, like number of heroes.

dydzio0614:
there is already difficulty.json for AI boosts per difficulty, not sure if anything more is needed here because tying actual mechanics to difficulty can be confusing to players

IvanSavenko:
Those are just starting resources. And not AI behavior configuration. For example H3 AI behavior differs greatly based on difficulty (at least for first 3 levels).

Parameters like how many heroes AI uses, how far AI builds pathfinder data for these heroes makes sense to have different per difficulty.

PR #4994, opened 2024-11-26
[WIP] Support ENABLE_STRICT_COMPILATION for Clang on Windows.

Kuxe:
Support `ENABLE_STRICT_COMPILATION` for Clang on Windows.

Previously `-Werror` was only passed to Clang if `NOT WIN32`.

WIP until I can compile all targets with `ENABLE_STRICT_COMPILATION` on Windows with `clang 18.1.8`.


kambala-decapitator:
rather use https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html instead of injecting flags manually. This variable should be set in preset.

Kuxe:
> rather use https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html instead of injecting flags manually. This variable should be set in preset.

I agree, but `CMAKE_COMPILE_WARNING_AS_ERROR` is added in version 3.24 and would require a bump from `cmake_minimum_required(VERSION 3.16.0)`. Not sure if we want to bump minimum CMake version? 3.24 is pretty recent too! Not available on Ubuntu 22.04 for instance.

kambala-decapitator:
it's not required to raise cmake version. If an Ubuntu user absolutely wants to treat warnings as errors, he would simply pass `-D CMAKE_CXX_FLAGS=-Werror` on the command line or from his own preset.

Kuxe:
> it's not required to raise cmake version. If an Ubuntu user absolutely wants to treat warnings as errors, he would simply pass `-D CMAKE_CXX_FLAGS=-Werror` on the command line or from his own preset.

Sure, but then we've wound up in a situation where a user sets `ENABLE_STRICT_COMPILATION` to `ON` and it doesn't work as documented, only to find out that (s)he is meant to pass `-D CMAKE_CXX_FLAGS=-Werror` on CMake versions lower than 3.24.

kambala-decapitator:
`ENABLE_STRICT_COMPILATION` should be removed as there's a cmake variable for that. VCMI is not a place to teach cmake to users, one should find how to treat warnings as errors on their own.

Kuxe:
> `ENABLE_STRICT_COMPILATION` should be removed as there's a cmake variable for that. VCMI is not a place to teach cmake to users, one should find how to treat warnings as errors on their own.

`ENABLE_STRICT_COMPILATION` is used by the `base-release`-preset and hence used on most (all?) release builds. To me it makes perfect sense to have one flag determine what VCMI thinks is a strict compilation, instead of leaving that up to each user. Is common in CI/CD farms to have a "strict" build.

Should we allow ENABLE_STRICT_COMPILATION for Clang+Windows in the interim until someone have removed ENABLE_STRICT_COMPILATION?

PR #4996, opened 2024-11-27
Enable AI to use all Hill Fort upgrades, including alternative ones

### LINKED ISSUE #4749: Nullkiller AI's visit of Hill Fort crashes game
**Describe the bug**

**Game logs**
[2024-Oct-10 16:23:24.550626] TRACE [runNetwork] ai - Entering virtual void NKAI::AIGateway::requestRealized(PackageApplied*).
[2024-Oct-10 16:23:24.550657] TRACE [runNetwork] ai - Leaving virtual void NKAI::AIGateway::requestRealized(PackageApplied*).
[2024-Oct-10 16:23:24.550697] TRACE [runNetwork] network - 	Made second apply on cl: 14PackageApplied
[2024-Oct-10 16:23:24.550811] TRACE [AIGateway::makeTurn] ai - Entering void NKAI::AIGateway::performObjectInteraction(const CGObjectInstance*, NKAI::HeroPtr): Hero Alex and object Hill Fort at (96 238 0).
[2024-Oct-10 16:23:24.551116] ERROR [AIGateway::makeTurn] global - Disaster happened.
[2024-Oct-10 16:23:24.553044] ERROR [AIGateway::makeTurn] global - Reason: 0xc0000005 - EXCEPTION_ACCESS_VIOLATION at 0033:0x7ff,aea,905,fda
[2024-Oct-10 16:23:24.554780] ERROR [AIGateway::makeTurn] global - Attempt to read from 0x0xf,fff,fff,fff,fff,ffc
[2024-Oct-10 16:23:24.556284] ERROR [AIGateway::makeTurn] global - Thread ID: 8968
[2024-Oct-10 16:23:24.564881] ERROR [AIGateway::makeTurn] global - Crash info will be put in VCMI_client.exe_crashinfo.dmp

**To Reproduce**

**Expected behavior**
NKAI doesn't crash game

**Actual behavior**
NKAI  crashes game

**Did it work earlier?**
If this something which worked well some time ago, please let us know about version where it works or at date when it worked.

**Screenshots**

**Version**
 - OS: Windows 11
 - Version: VCMI-branch-develop-f6f6f98

**Additional context**

MichalZr6:
This PR fixes #4749 issue.

AI will pick now the most valuable upgrade if creature has alternative ones.

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -19,6 +19,7 @@
 #include "../../lib/CConfigHandler.h"
 #include "../../lib/IGameSettings.h"
 #include "../../lib/gameState/CGameState.h"
+#include "../../lib/gameState/UpgradeInfo.h"
 #include "../../lib/serializer/CTypeList.h"
 #include "../../lib/networkPacks/PacksForClient.h"
 #include "../../lib/networkPacks/PacksForClientBattle.h"
@@ -805,7 +806,7 @@ bool AIGateway::makePossibleUpgrades(const CArmedInstance * obj)
 						myCb->upgradeCreature(obj, SlotID(i), upgID);
 						upgraded = true;
 						logAi->debug("Upgraded %d %s to %s", s->count, upgradeInfo.oldID.toCreature()->getNamePluralTranslated(), 
-							upgradeInfo.getNextUpgrade().toCreature()->getNamePluralTranslated());
+							upgradeInfo.getUpgrade().toCreature()->getNamePluralTranslated());
 					}
 					else
 						break;
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -788,14 +788,29 @@ bool AIGateway::makePossibleUpgrades(const CArmedInstance * obj)
 	{
 		if(const CStackInstance * s = obj->getStackPtr(SlotID(i)))
 		{
-			UpgradeInfo ui;
-			myCb->fillUpgradeInfo(obj, SlotID(i), ui);
-			if(ui.oldID != CreatureID::NONE && nullkiller->getFreeResources().canAfford(ui.cost[0] * s->count))
+			UpgradeInfo ui(s->getId());
+			do
 			{
-				myCb->upgradeCreature(obj, SlotID(i), ui.newID[0]);
-				upgraded = true;
-				logAi->debug("Upgraded %d %s to %s", s->count, ui.oldID.toCreature()->getNamePluralTranslated(), ui.newID[0].toCreature()->getNamePluralTranslated());
+				myCb->fillUpgradeInfo(obj, SlotID(i), ui);
+
+				if(ui.hasUpgrades())
+				{
+					// creature at given slot might have alternative upgrades, pick best one
+					CreatureID upgID = *vstd::maxElementByFun(ui.getAvailableUpgrades(), [](const CreatureID & id)
+						{
+							return id.toCreature()->getAIValue();
+						});
+					if(nullkiller->getFreeResources().canAfford(ui.getUpgradeCostsFor(upgID) * s->count))
+					{
+						myCb->upgradeCreature(obj, SlotID(i), upgID);
+						upgraded = true;
+						logAi->debug("Upgraded %d %s to %s", s->count, ui.oldID.toCreature()->getNamePluralTranslated(), ui.getNextUpgrade().toCreature()->getNamePluralTranslated());
+					}
+					else
+						break;
+				}
 			}
+			while(ui.hasUpgrades());
 		}
 	}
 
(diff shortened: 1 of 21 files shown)


client/windows/CCreatureWindow.cpp
```diff
@@ -755,17 +759,16 @@ CStackWindow::CStackWindow(const CStackInstance * stack, bool popup)
...
+CStackWindow::CStackWindow(const CStackInstance * stack, std::function<void()> dismiss, UpgradeInfo && upgradeInfo, std::function<void(CreatureID)> callback)
 	: CWindowObject(BORDERED),
 	info(new UnitView())
 {
 	info->stackNode = stack;
 	info->creature = stack->getCreature();
 	info->creatureCount = stack->count;
 
-	info->upgradeInfo = std::make_optional(UnitView::StackUpgradeInfo());
+	info->upgradeInfo = std::make_optional(UnitView::StackUpgradeInfo(std::move(upgradeInfo)));
```
IvanSavenko, line 770: To be honest previous approach might have been better.

Sure, your change might result in a little bit faster code, depending on compiler behavior. But this is not performance-critical code, and it makes code a little more complex for no reason (due to now required std::move). And most of our codebase does not use this approach.

lib/gameState/CGameState.h
```diff
@@ -39,12 +39,67 @@ struct SThievesGuildInfo;
 class CRandomGenerator;
 class GameSettings;
 
-struct UpgradeInfo
+class UpgradeInfo
```
IvanSavenko, line 42: Since this is no longer tiny struct, can you move it to a separate file in same directory?

lib/gameState/CGameState.h
```diff
@@ -39,12 +39,67 @@ struct SThievesGuildInfo;
...
+	UpgradeInfo(CreatureID base)
+		: oldID(base), isAvailable(true)
+	{ }
+
 	CreatureID oldID; //creature to be upgraded
-	std::vector<CreatureID> newID; //possible upgrades
-	std::vector<ResourceSet> cost; // cost[upgrade_serial] -> set of pairs<resource_ID,resource_amount>; cost is for single unit (not entire stack)
-	UpgradeInfo(){oldID = CreatureID::NONE;};
+
+	const auto & getAvailableUpgrades() const
```
IvanSavenko, line 52: Not sure about auto as return type since it reduces readability a bit.
Previously I only needed to make a quick look on method signature to determine its return type. Now I need to also lookup what `upgradesIDs` is. Trivial in this method, but not so trivial in some of the other methods in this class.

Sure, IDE will detect return type, and show it in a tooltip, but when looking through list of methods it is way easier to locate method you're looking for with explicit return types.

So proposition: replace all deduced / auto return types with concrete values

lib/gameState/CGameState.h
```diff
@@ -39,12 +39,67 @@ struct SThievesGuildInfo;
...
+
+	const auto & getUpgradeCosts() const
+	{
+		return upgradesCosts;
+	}
+
+	const auto & getNextUpgradeCosts() const
+	{
+		return upgradesCosts.back();
+	}
```
IvanSavenko, line 79: What exactly this does? What is "next upgrade" and why it always picks last element from list?

AI/Nullkiller/AIGateway.cpp
```diff
@@ -790,14 +790,29 @@ bool AIGateway::makePossibleUpgrades(const CArmedInstance * obj)
 	{
 		if(const CStackInstance * s = obj->getStackPtr(SlotID(i)))
 		{
-			UpgradeInfo ui;
-			myCb->fillUpgradeInfo(obj, SlotID(i), ui);
-			if(ui.oldID != CreatureID::NONE && nullkiller->getFreeResources().canAfford(ui.cost[0] * s->count))
+			UpgradeInfo ui(s->getId());
```
IvanSavenko, line 793: No need to save on variable name length. UI is also very unclear, since it often stands for "user interface"
```suggestion
			UpgradeInfo upgradeInfo(s->getId());
```

lib/gameState/CGameState.h
MichalZr6: fields haven't changed, just moved 2 of them to private and added few functions. Should I make it back a struct or stay with class and move it to its own header and source files ?

IvanSavenko: Move to own header (and source file if necessary).

By "tiny struct" I mean not size in memory and such, but rather size of code, which is now rather large due to having ~10 new methods.

client/windows/CCreatureWindow.cpp
MichalZr6: Got rid of std::move

lib/gameState/CGameState.h
MichalZr6: It used to pick first element in code previous to [this PR](https://github.com/vcmi/vcmi/pull/4348/commits/6cf423d97760553ff7d32f3cea814598ce8d6331), but works correctly when it takes the last one up untill there are no upgrades. UpgradeInfo is refilled on every upgrade attempt. 
I tested this code to work with alternative upgrades and with third-upgrades-mod that allows to do upgrades two times from basic. This `getNextUpgradeCosts` will pick the upgrade costs of upgrade which should be done at that specific moment. 
Hmmm... I have to admit: "Next" is rather misleading and the naming should be just `getUpgradeCosts` and maybe the other function with that name should be `getAvailableUpgradesCosts`.

IvanSavenko: Yeah, `getUpgradeCosts` sounds more clear

PR #4998, opened 2024-11-27
Allow per-difficulty parameters for NKAI

IvanSavenko:
NKAI now selects its settings based on game difficulty.

Fow reveal cheat usage now defined in config and no longer hardcoded.
No changes to AI behavior.

Actual selection of parameters for different levels will go to a separate PR (including deciding on how to handle map reveal)


dydzio0614:
it may also make sense to expose safe attack constant - could be useful to go in pair with battle bonuses

Issue #4547, opened 2024-09-01
Pictures as frames from def files stopped showing.


Macron1Robot:
**Describe the bug**
Functionality of using def frames as pictures stopped working.

**Game logs**

**To Reproduce**
`"icons" :
 {
 "village" : 
{
"normal" :
{
"large":  "newfaction/itpt.def:14",
"small" : "newfaction/ITPA.def:16"
}, 
"built" : 
{
"large" : "newfaction/itpt.def:15",
"small" : "newfaction/ITPA.def:17"
} 
},
"fort": 
{
"normal" :
{
"large":  "newfaction/itpt.def:14",
"small" : "newfaction/ITPA.def:16"
}, 
"built" : 
{
"large":  "newfaction/itpt.def:15",
"small" : "newfaction/ITPA.def:17"
} 
}
 }`

**Expected behavior**
Frames will be taken from def file and shown in-game

**Actual behavior**
Frames are not taken from def file and not shown in-game

**Did it work earlier?**
In Beta it works

**Screenshots**

**Version**
VCMI-branch-develop-39e5ba3

**Additional context**

Macron1Robot:
Please someone fix it. It crashes game when i open hero window and there are artifacts which use frames from DEF files.

mikeiit:
Does this have anything to do with it?
![Screenshot 2024-09-15 at 8 11 43 PM](https://github.com/user-attachments/assets/78e6a2cb-7670-4718-9ea7-78c81c8e661d)

mikeiit:
Does this have anything to do with it? 
![Screenshot 2024-09-15 at 8 13 58 PM](https://github.com/user-attachments/assets/8914a30c-7162-4861-8dc2-390fd58c7df0)

Macron1Robot:
I think its some other.

Olegmods:
Some citation from Ivan from Discord:
"FYI: unlikely to be fixed soon, and when it will - it likely will have different form. 
Had to remove that logic when implementing support for xbrz / upscaled assets."

MichalZr6:
Can you check if there are still crashes using daily build ?

Macron1Robot:
> Can you check if there are still crashes using daily build ?

doesnt crash and doesnt show

Macron1Robot:
Is there a slightest chance that using frames from defs as images will be fixed any time soon?

Macron1Robot:
> Can you check if there are still crashes using daily build ?

[...]

PR #4999, opened 2024-11-27
Fix crash on missing DEF frames for in-game assets

MichalZr6:
This PR addresses the issue #4547 where the game crashes when attempting to load DEF frames for in-game assets, particularly when the specified frames are missing or invalid. While the underlying functionality of using DEF frames as pictures is currently disabled (due to changes to xbrz/upscaled support), this fix ensures the game handles missing frames without crashing. You will only see log message, for example:

> Frame index 50 is out of bounds for group 0 in file: SPRITES/TOWN-SCREEN/CASTLE/HALLCSTL

Please note: I have not tested this extensively, and the fix might not work in all cases.


IvanSavenko:
I think way easier and simple logic is to check whether DefFile contains specified frame, and if not - return "DEFAULT" image in `RenderHandler::loadImageFromFileUncached`

MichalZr6:
What do you mean by default image to return ? Something like this ?
```cpp
if(defFile->hasFrame(locator.defFrame, locator.defGroup))
	return std::make_shared<SDLImageShared>(defFile.get(), locator.defFrame, locator.defGroup, preScaledFactor);
else
{
	logGlobal->error("Frame %d in group %d not found in file: %s", 
		locator.defFrame, locator.defGroup, locator.defFile->getName().c_str());
	return std::make_shared<SDLEmptyImageShared>();
}
```

```cpp
class SDLEmptyImageShared : public SDLImageShared
{
public:
	SDLEmptyImageShared() : SDLImageShared(createDefaultSurface().get()) {}

private:
	std::shared_ptr<SDL_Surface> createDefaultSurface()
	{
		auto surface = SDL_CreateRGBSurface(0, 32, 32, 32, 0, 0, 0, 0);
		SDL_FillRect(surface, nullptr, SDL_MapRGB(surface->format, 255, 0, 0));
		return std::shared_ptr<SDL_Surface>(surface, SDL_FreeSurface);
	}
};
```
the size would have to be probably acquired from defFile with some additional code in this class for retrieving it, but not a problem.

IvanSavenko:
What I mean is add something like this to `RenderHandler::loadImageFromFileUncached`:
```cpp
		if (defFile->containsImage(locator.defFrame, locator.defGroup))
			return std::make_shared<SDLImageShared>(defFile.get(), locator.defFrame, locator.defGroup, preScaledFactor);
		else
			return std::make_shared<SDLImageShared>(ImagePath::builtin("DEFAULT"), locator.preScaledFactor);
```
(`containsImage` method does not exists yet, and needs to be implemented)

Also - method `RenderHandler::getLocatorForAnimationFrame` already has check for non-existing frame in def. Why that check is not triggered in this scenario?

MichalZr6:
Thanks. I have already implemented `containsImage` on my local branch (named it `hasFrame` instead) but didn't know what the default image should be. With `ImagePath::builtin("DEFAULT")` the town hall screen (with missing img) looks like this:

obraz

As for `getLocatorForAnimationFrame` - it never triggers. Both in my test case with missing frame and also while clicking on some creatures, dwellings, running some battles, etc. Looked through call stack. Another function is used (`CAnimation::getImageLocator`) in [CAnimation::loadFrame](https://github.com/vcmi/vcmi/blob/150f4904aba88b9d1a9896adc0a863fb2c8a82ac/client/render/CAnimation.cpp#L21)

client/renderSDL/SDLImage.h
```diff
@@ -11,6 +11,7 @@
 
 #include "../render/IImage.h"
 #include "../../lib/Point.h"
+#include <SDL_surface.h>
```
IvanSavenko, line 14: Please avoid includes in headers, especially 3rd party libraries

client/renderSDL/SDLImage.h
```diff
@@ -66,6 +67,20 @@ class SDLImageShared final : public ISharedImage, public std::enable_shared_from
...
+
+private:
+	std::shared_ptr<SDL_Surface> createDefaultSurface()
+	{
+		auto surface = SDL_CreateRGBSurface(0, 32, 32, 32, 0, 0, 0, 0);
+		SDL_FillRect(surface, nullptr, SDL_MapRGB(surface->format, 255, 0, 0));
+		return std::shared_ptr<SDL_Surface>(surface, SDL_FreeSurface);
+	}
+};
+
```
IvanSavenko, line 83: Unused? 
Also - what's exact point of this class?

If you're implementing my TODO, then its goal was somewhat different - it was to create dummy implementation of ISharedImage to replace checks for non-null surface in every method of SDLImageShared

client/render/CDefFile.cpp
```diff
@@ -38,7 +38,8 @@ enum class DefType : uint32_t
 
 CDefFile::CDefFile(const AnimationPath & Name):
 	data(nullptr),
-	palette(nullptr)
+	palette(nullptr),
+	fileName(Name.getName())
```
IvanSavenko, line 42: Unused?

IvanSavenko (commented):
I think all changes in SDLImage.h can be safely reverted?

client/renderSDL/SDLImage.h
MichalZr6: Just tried to implement some "default" as I didn't realise that there is sth like that already.

MichalZr6:
> I think all changes in SDLImage.h can be safely reverted?

Reverted unnecessary changes.

PR #5000, opened 2024-11-28
Add Start Game page to launcher with preset management

IvanSavenko:
With this PR, "Start Game" button in Launcher will no longer immediately start the game, but will instead open new tab. This tab is now also the default tab in launcher and is shown on start instead of mods tab.

Goals of this changes are:
- provide a way for player to manage mod presets. Existing Mods tab does not have enough free space for preset management.
- provide more logical place for "Import files" functionality, since it can now be used for a lot of things other than importing mods.
- show various self-diagnostic information - VCMI updates, issues with H3 files such as no music (which sometimes fails to copy for some reason on Android), broken town hall screen (due to usage of pirated game), and hints on VCMI functionality - import of Chronicles, and hopefully - import of HD Edition in future.
- slightly reduce focus on mods, since vcmi no longer shows them on start
- add "Resume game" / "load latest save" option to quickly continue game (extremely likely to be postponed to later release)

Current quick draft of UI:
(some options on the screenshot are mutually exclusive, or shown only in case of issues, so it will be less overloaded in actual use)
launcher_start

### DIFF
--- a/launcher/CMakeLists.txt
+++ b/launcher/CMakeLists.txt
@@ -15,6 +15,7 @@ set(launcher_SRCS
 		modManager/imageviewer_moc.cpp
 		modManager/chroniclesextractor.cpp
 		settingsView/csettingsview_moc.cpp
+		startGame/StartGameTab.cpp
 		firstLaunch/firstlaunch_moc.cpp
 		main.cpp
 		helper.cpp
@@ -46,6 +47,7 @@ set(launcher_HEADERS
 		modManager/imageviewer_moc.h
 		modManager/chroniclesextractor.h
 		settingsView/csettingsview_moc.h
+		startGame/StartGameTab.h
 		firstLaunch/firstlaunch_moc.h
 		mainwindow_moc.h
 		languages.h
@@ -63,6 +65,7 @@ set(launcher_FORMS
 		settingsView/csettingsview_moc.ui
 		firstLaunch/firstlaunch_moc.ui
 		mainwindow_moc.ui
+		startGame/StartGameTab.ui
 		updatedialog_moc.ui
 )
 
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -549,15 +549,13 @@ void FirstLaunchView::modPresetUpdate()
 
 QString FirstLaunchView::findTranslationModName()
 {
-	if (!getModView())
+	auto * mainWindow = dynamic_cast<MainWindow *>(QApplication::activeWindow());
+	auto status = mainWindow->getTranslationStatus();
+
+	if (status == ETranslationStatus::ACTIVE || status == ETranslationStatus::NOT_AVAILABLE)
 		return QString();
 
 	QString preferredlanguage = QString::fromStdString(settings["general"]["language"].String());
-	QString installedlanguage = QString::fromStdString(settings["session"]["language"].String());
-
-	if (preferredlanguage == installedlanguage)
-		return QString();
-
 	return getModView()->getTranslationModName(preferredlanguage);
 }
 
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -389,7 +389,7 @@ void FirstLaunchView::extractGogData()
 		};
 
 		if(isGogGalaxyExe(tmpFileExe))
-			errorText = tr("You've provided GOG Galaxy installer! This file doesn't contain the game. Please download the offline backup game installer!");
+			errorText = tr("You've provided a GOG Galaxy installer! This file doesn't contain the game. Please download the offline backup game installer!");
 
 		if(errorText.isEmpty())
 			errorText = Innoextract::extract(tmpFileExe, tempDir.path(), [this](float progress) {
@@ -459,7 +459,7 @@ void FirstLaunchView::copyHeroesData(const QString & path, bool move)
 	QStringList dirMaps = sourceRoot.entryList({"maps"}, QDir::Filter::Dirs);
 	QStringList dirMp3 = sourceRoot.entryList({"mp3"}, QDir::Filter::Dirs);
 
-	const auto noDataMessage = tr("Failed to detect valid Heroes III data in chosen directory.\nPlease select directory with installed Heroes III data.");
+	const auto noDataMessage = tr("Failed to detect valid Heroes III data in chosen directory.\nPlease select the directory with installed Heroes III data.");
 	if(dirData.empty())
 	{
 		QMessageBox::critical(this, tr("Heroes III data not found!"), noDataMessage);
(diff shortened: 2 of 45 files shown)


DjWarmonger:
...I can't find that XKCD comic with proper roast for this design...

Why not just add this as a 4th panel (icon on the right) and do not make mod manager default? Preset manager should be the first panel player sees, with few recommended presets already there. Let's make mod manager "advanced" feature for curious players only.

IvanSavenko:
What exactly do you think is wrong in this design? "can't find proper roast" does not tells me much.

> Preset manager should be the first panel player sees with few recommended presets already there. Let's make mod manager "advanced" feature for curious players only.

Why? Even in MP 1/3 of players don't use mods at all. Number is likely way larger for casual singleplayer. For them VCMI is about playing H3 on non-Windows OS, not a way to mod H3.

I also would like to have some logical place for "resume latest save" and to report on missing H3 files, e.g. soundtrack - which seems to be rather common issue for some reason. Preset manager is not the best place for these.

Also this is not what presets (at the moment) are about. At the moment this is just a way to quickly switch between mod configurations without manual reenabling of mods. We also don't have way to provide recommended presets other than embed them into vcmi. 

May be an interesting idea to consider for 1.7+, depending on feedback on presets.

krs0:
I fully agree twith chainging the first page from the mod manager to something more simple and useful.

First proposal looks a little crowded.

In no particular order...

Translations - Would only a dropdown with supported languages work?
When you select another language, translation file presence is cheched for core and all mods in active preset, and selected? Defaulting to EN when not present?

As to installing tranlations...I think now for mods they are installed by default, so only core needs download? but these are small files so maybe ok to have them installed by default, and not shown to player?


Presets
Copy to clipboard, if the intended usage is for support, would make sense to have it under help menu? toghether with some other stuff like detailed version info and logs?
Paste from cliboard - what is it good for? I would expect presets are comming bundled in mods or from files via file manager or from another player and here a button in lobby to import preset would be nice.

I think if presets would be individual files, and not bundled with other stuff, they would be easier to manage.

Campaigns... why would I see them here? Arent they just a special kind of mod? I never play them so why would I see them on every launch?

Missing instalation files... here no ideas, but seems much to see it all the time. Maybe just show an indicator when something is missing and redirect player to instalation page, where player can fix various stuff?

krs0:
Create empty preset -> could be an entry called new or empty under preset selection dropdown?

Save preset is missing, allowind the olayer to save locally a custom preset.
Then we will need a way to delete local preset as well.

DjWarmonger:
> Copy to clipboard, if the intended usage is for support, would make sense to have it under help menu? toghether with some other stuff like detailed version info and logs?

This. We don't want to place any optional tools for modders (or developers) in front of a user who tries to launch the game for the first time.

Middle panel should be either separate tab, or a part of initial setup wizard. Left panel can get its own tab, too. It can be the default panel if no auto-update is available.

This leads to conclusion that presets need their own panel, too. I envisioned as a part of current mod list view, just add it above so players instantly know that **some** preset is selected and they are ready to go. That's better than displaying endless list of mods in no particular order, as the latter invites players to install all of them at once.

IvanSavenko:
> First proposal looks a little crowded.

Irrelevant information would be hidden (e.g. no info on translations if you already have it or if your H3 copy already has your language). For example in central list only "install files" and possibly Chronicles is expected to be visible for most players.

Also, this is about Heroes III translations, not about mod translations.

> Copy to clipboard, if the intended usage is for support, would make sense to have it under help menu? toghether with some other stuff like detailed version info and logs?

Intended use is to share presets between players. Player uses "copy to clipboard", sends it to another player, another player uses "paste from clipboard" to get same mod config as first player.

Perhaps rename to "export to clipboard" / "import from clipboard"? Or you have other suggestions?

> Campaigns... why would I see them here? Arent they just a special kind of mod? 

Which ones? For Chronicles goals is to indicate possibility of import (via gog installer). AB campaigns - this is one more check for invalid data, some players somehow ended up without those files.
We definitely had reports about those from player who use that "Heroes Launcher" or what's its name. And I saw similar crashes on Android in 1.5

> And move icon for this Start Game tab to the top

You mean one on the left? Not sure... You might be right, but existing players are definitely used to having it at the bottom. Might be better to keep as it.

> Then we will need a way to delete local preset as well.

Good point. Missed that one.

krs0:
For campains or top content selection (Chronicles, HoTA, SoD, DoR?)
A nice graphical section with buttons or large image would be nice. And those buttons just select the right preset. with instalation and everything.

I also like the Heroes launcher approach cause its graphical.



But uses too much space and lists all mods all the time. Only one large image with left and rright cursors, like in web pages.

IvanSavenko:
> This. We don't want to place any optional tools for modders (or developers) in front of a user who tries to launch the game for the first time.

Probably bad choice of name. Intent is to provide for a player a way to share his list of mods with another player.

> Middle panel should be either separate tab, or a part of initial setup wizard. 

Makes sense to show it during setup. Will keep that in mind if/when I'll try to update it. But for existing players or those who ignored / skipped that check might be better to display it somewhere where they can see it.

> Left panel can get its own tab, too. It can be the default panel if no auto-update is available.

We are already low on space in there. Currently we don't even have space for Map Editor tab/button on small mobile devices (luckily, editor is PC-only now). Also goal is to make sure that player can see it / bring it to his attention, not hide it somewhere in yet another tab.

> This leads to conclusion that presets need their own panel, too. I envisioned as a part of current mod list view, just add it above so players instantly know that some preset is selected and they are ready to go. That's better than displaying endless list of mods in no particular order, as the latter invites players to install all of them at once.

Hmm... I guess we can show drop-down to quickly switch preset in mod list (even though it would eat some space from the list). 

Separate tab for presets might be useful, especially if this feature turns out to be popular (and yeah, it likely will be). But most common operation(s) like switching between presets should also be available from main screen as well.

I also don't want to spend too much time on presets at the moment. So it is either cut it completely from 1.6 or add some basic functionality (player-managed presets) and improve it later, e.g. in 1.7 (community presets similar to mod list, multiplayer integration, etc).

IvanSavenko:
> For campains or top content selection (Chronicles, HoTA, SoD, DoR?) A nice graphical section with buttons or large image would be nice. And those buttons just select the right preset. with instalation and everything.

To be honest this looks more like upgraded version for our mod downloader / manager. For example, we also provide screenshots, but they are hidden in submenus.

Definitely better than looking through 100+ mods but not yet sure how to integrate it into our existing approach.

Something to think about for 1.7 probably.

dydzio0614:
i think design should assume hardcoded presets coming at some point, like "pvp ranked preset"

IvanSavenko:
> i think design should assume hardcoded presets coming at some point, like "pvp ranked preset"

Not sure if I would call them "hardcoded". I get their need & usecase, but don't see need to hardcode it in any way.

For example, what if we were to add "pvp ranked preset" as a mod and show its name somewhere in lobby room UI instead of (potentially massive) list of used mods? So any players interested in joining only need to activate such mod to play in this room AND they know that this is not some custom mod configuration, but one that is somewhat balanced and one they are likely already familiar with.

Of course in UI this mod must be integrated with presets in one way or another, but point is - information on what "pvp ranked preset" is can come from mod repository and does not have to be hardcoded.

> the thing is some people consider VCMI moddability (unlike hota) "disadvantage" in certain contexts, so one feature that will be needed at some point is built-in presets or some other idiotproof abstraction layer over mod system that should make VCMI usage for such use cases easier

Yeah, I agree that is hard to find something "interesting" in our mod list without recommendations. On the other hand, such additional abstraction layer basically duplicates mods. I mean, to attract players, such preset would need to have description, screenshots, list of included mods, etc. Basically - everything that mods already have!

Presets would definitely need 2nd iteration to account for these use-cases, but it is not something that I can do in time for 1.6.

Laserlicht:
Looks good.

But I'm not sure if we need everything on that page. Seems already too much for small smartphones. Maybe add diagnostics tab for some informations from this page?

Would also add icon to play button to make it more prominent.

IvanSavenko:
> But I'm not sure if we need everything on that page. Seems already too much for small smartphones. Maybe add diagnostics tab for some informations from this page?

Most of it would be hidden and only shown when needed. For example on left column "Update available", "You're using latest version" and "Check for updates" should all be mutually exclusive. And on central column you need to really know how to break the game for all those diagnostics to trigger. Othervice you'll only see "Install files" and Chronicles.

> Would also add icon to play button to make it more prominent.

Of course. Icons will be added where applicable.

DjWarmonger:
Looks like a lot of unused space left from my perspective: 
image

How a vertical bar became horizontal? With duplicated VCMI icon?

IvanSavenko:
First version ready, there are still few things I'd like to add, but those won't be in time for 1.6.0:
- Resume game - may be postponed to 1.7
- Export / Import presets - will try to add in 1.6.x
- Better automatic update check without window jumping in front of Launcher.

All Qt translation files have been regenerated to add new strings

Here is current look (may be different depending on game data):
зображення

PR #5003, opened 2024-11-29
Change NKAI configuration for different difficulty levels

IvanSavenko:
First attempt, feedback welcomed.

Changes compared to existing logic:
- Fow reveal is now on for every difficulty. AI behavior right now is just too poor othervice.
- Heroes limit is now 3/3/4/6/8 depending on difficulty
- Easy/Pawn AI will never attempt to retreat from combat, Normal / knight will be more risky with retreats
- Easy/Pawn and normal/knight are now more risky with combats, should decrease their capabilities to assemble an army.

And there is also already existing weekly resource bonuses / maluses, similar to H3. Can't say that I extensively tested this on every possible map, but in my tests AI now plays fairly differently on different difficulty levels.

### DIFF
--- a/AI/Nullkiller/Engine/PriorityEvaluator.cpp
+++ b/AI/Nullkiller/Engine/PriorityEvaluator.cpp
@@ -1349,7 +1349,8 @@ float PriorityEvaluator::evaluate(Goals::TSubgoal task, int priorityTier)
 	else
 	{
 		float score = 0;
-		float maxWillingToLose = ai->cb->getTownsInfo().empty() || (evaluationContext.isDefend && evaluationContext.threatTurns == 0) ? 1 : 0.25;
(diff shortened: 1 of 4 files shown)


edeksumo:
In OH at easy AI won't build 7 tier in towns

dydzio0614:
Needs to be written
# like this

in patch notes so people will maybe not complain after release that "ai bad" because most does not read patch notes and will assume million stuff based on previous VCMI versions like "ai plays same on every difficulty"

Xilmi:
Limiting the AI's capabilities on anything above Knight-difficulty seems wrong to me.
When Rook is supposed to be the fair difficulty-level it seems wrong to limit it's heroes and how smart it is about retreating.

I'd say limitations should only hit pawn- and knight-difficulty.

PR #5018, opened 2024-12-01
Added missing translation strings

Laserlicht:
@GeorgeK1ng Have fun. :D

Anything else missing?


GeorgeK1ng:
Does this mean whole lobby is now translatable? :)

GeorgeK1ng:
Last time I found only lobby related messages, so probably it's all, but who knows :D

lib/texts/TextOperations.cpp
```diff
@@ -221,7 +221,7 @@ std::string TextOperations::getFormattedDateTimeLocal(std::time_t dt)
 
 std::string TextOperations::getFormattedTimeLocal(std::time_t dt)
 {
-	return vstd::getFormattedDateTime(dt, "%H:%M");
+	return vstd::getFormattedDateTime(dt, "%H:%M:%S");
```
IvanSavenko, line 224: Why was this changed?
Right now this is only used for chat in lobby where seconds are generally not needed

Laserlicht: To harmonize time format with datetime format defined for each language (every language uses %H:%M:%S)

Laserlicht: And in chats IMO sometimes a useful information.

Laserlicht: Yes, but why hide it?

At least here timestamps with seconds are much more usual as without.

IvanSavenko: And why show it?

Normally you're not making saves every second. For like 99% of cases this is irrelevant information. Usually you only need to find latest save, so having less text is certainly better.
As far as saves go, I'd even consider replacing recent dates with "Today, 12:34" / "Yesterday, 12:34"

Similarly in game chat - generally you only need to know whether this message is recent or old one.

Laserlicht: As said, the additional information is that you know if two people written the same time (and one message can't be a respond of the other).

Personally I don't like different formats everywhere.

Microsoft teams for example uses 15:11 for today and 13.11. two days before. Misleading...

But part of this PR is only to remove differences within VCMI itself. If we want both in hh:mm we can change both afterwards.

IvanSavenko: If unification is the goal then I'd suggest using hours+minutes everywhere. And check if this would result in any complains / inconveniences / etc

> two people written the same time 

But that can depend a lot on typing speed. Or simply not paying attention to last messages.

PR #5021, opened 2024-12-01
Fix for AI not defending in some cases

Xilmi:
Instead of skipping the defense of other towns entirely when hiring a hero was considered as a way of defending, now only hiring the same hero twice will be avoided.


Xilmi:
Converted back to draft since disbanding of units seems to disband a bit more than intended.

Xilmi:
Reopened PR since it now all seems to be working quite fine. The inability to disband small stacks really hurt the AI's army-growth-potential on maps where they had many different kinds of towns. And now it looks really quite decent in that regard.
Army-Strength

PR #5022, opened 2024-12-01
generate file hashes if innoextract fails

### LINKED ISSUE #4879: Gog installer "bad chunk error"
**Describe the bug**
`bad chunk error` appears after selecting bin and exe file.
This error happens, if exe is from another language than bin file (e.g. polish exe, english bin file)

**Expected behavior**
Easy to understand error description

**Actual behavior**
For common user cryptic error.

**Screenshots**
grafik

**Version**
 - OS: e.g. Android
 - Version: 1.5.7

**Additional context**
Currently no idea to fix. Only idea is validating over file hashes... not sure about that. Maybe there are also some infos about language in bin/exe to extract...

Laserlicht:
Second error window after (already existing) innoextract error:
grafik
grafik
grafik
grafik

fixes #4879

### DIFF
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -391,7 +391,7 @@ void FirstLaunchView::extractGogData()
 		
 		QString hashError;
 		if(!errorText.isEmpty())
-			hashError = Innoextract::getHashError(tmpFileExe, tmpFileBin);
+			hashError = Innoextract::getHashError(tmpFileExe, tmpFileBin, fileExe, fileBin);
 
 		ui->progressBarGog->setVisible(false);
 		ui->pushButtonGogInstall->setVisible(true);
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -359,8 +359,9 @@ void FirstLaunchView::extractGogData()
 			return; // should not happen - but avoid deleting wrong folder in any case
 
 		QString tmpFileExe = tempDir.filePath("h3_gog.exe");
+		QString tmpFileBin = tempDir.filePath("h3_gog-1.bin");
 		QFile(fileExe).copy(tmpFileExe);
-		QFile(fileBin).copy(tempDir.filePath("h3_gog-1.bin"));
+		QFile(fileBin).copy(tmpFileBin);
 
 		QString errorText{};
 
@@ -387,6 +388,10 @@ void FirstLaunchView::extractGogData()
 				ui->progressBarGog->setValue(progress * 100);
 				qApp->processEvents();
 			});
+		
+		QString hashError;
+		if(!errorText.isEmpty())
+			hashError = Innoextract::getHashError(tmpFileExe, tmpFileBin);
 
 		ui->progressBarGog->setVisible(false);
 		ui->pushButtonGogInstall->setVisible(true);
@@ -395,8 +400,12 @@ void FirstLaunchView::extractGogData()
 		QStringList dirData = tempDir.entryList({"data"}, QDir::Filter::Dirs);
 		if(!errorText.isEmpty() || dirData.empty() || QDir(tempDir.filePath(dirData.front())).entryList({"*.lod"}, QDir::Filter::Files).empty())
 		{
-			if(!errorText.isEmpty()) //
+			if(!errorText.isEmpty())
+			{
 				QMessageBox::critical(this, tr("Extracting error!"), errorText, QMessageBox::Ok, QMessageBox::Ok);
+				if(!hashError.isEmpty())
+					QMessageBox::critical(this, tr("Hash error!"), hashError, QMessageBox::Ok, QMessageBox::Ok);
+			}
 			else
 				QMessageBox::critical(this, tr("No Heroes III data!"), tr("Selected files do not contain Heroes III data!"), QMessageBox::Ok, QMessageBox::Ok);
 			tempDir.removeRecursively();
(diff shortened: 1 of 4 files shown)


GeorgeK1ng:
That’s interesting approach, but what about older installers? This adds support only for latest ones. Any chance to turn it to config so we can add more known variants in future without needs to change hardcoded values?

Laserlicht:
> That’s interesting approach, but what about older installers? This adds support only for latest ones. Any chance to turn it to config so we can add more known variants in future without needs to change hardcoded values?

This new error dialog is only opened if there is any error in extraction process. If everything is fine there is no hashing and no new window.
IMO it doesn't make sense to add it to config, because it's nothing to change by user or by mod.

If you have hashes of older installers (before 2019?): It would be great if you could share them.

GeorgeK1ng:
I'm sure I have them somewhere, just need to find them on some externals disks.

launcher/innoextract.cpp
```diff
@@ -60,3 +60,99 @@ QString Innoextract::extract(QString installer, QString outDir, std::function<vo
...
+				return QString{}; // error with hashing
+			binSize = bin.size();
+		}
+		else
+			return QString{}; // reading problem
+	}
+	
+	QString hashOutput = tr("SHA1 hash of provided files:\n") + tr("Exe") + " (" + QString::number(exeSize) + " " + tr("bytes") + ")" + ":\n" + QString::fromStdString(exeHash);
+	if(!binHash.empty())
+		hashOutput += "\n" + tr("Bin") + " (" + QString::number(binSize) + " " + tr("bytes") + ")" + ":\n" + QString::fromStdString(binHash);
```
IvanSavenko, line 132: Please use strings with placeholders (e.g. `%1`) instead of appending to string. It is both more easy to read in code and easier to translate.

launcher/innoextract.cpp
```diff
@@ -60,3 +60,99 @@ QString Innoextract::extract(QString installer, QString outDir, std::function<vo
...
+				binHash = hash.result().toHex().toLower().toStdString();
+			else
+				return QString{}; // error with hashing
+			binSize = bin.size();
+		}
+		else
+			return QString{}; // reading problem
+	}
+	
+	QString hashOutput = tr("SHA1 hash of provided files:\n") + tr("Exe") + " (" + QString::number(exeSize) + " " + tr("bytes") + ")" + ":\n" + QString::fromStdString(exeHash);
```
IvanSavenko, line 130: Perhaps instead of "exe" and "bin" show actual names as provided by player? Before copy I mean.

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -359,8 +359,9 @@ void FirstLaunchView::extractGogData()
 			return; // should not happen - but avoid deleting wrong folder in any case
 
 		QString tmpFileExe = tempDir.filePath("h3_gog.exe");
+		QString tmpFileBin = tempDir.filePath("h3_gog-1.bin");
 		QFile(fileExe).copy(tmpFileExe);
-		QFile(fileBin).copy(tempDir.filePath("h3_gog-1.bin"));
+		QFile(fileBin).copy(tmpFileBin);
```
IvanSavenko, line 364: Unrelated to this PR, but could be this copy somehow causing those "file corruption" errors?

Never heard about corrupted downloads via browsers for like 10+ years. It is only gog installers that are corrupted for some reason.

What if original installer in Downloads or where user has it is fine, but Qt fails to copy it for some reason?

What if we were to change this code to only copy in case of "bad" filenames, as in - bin name is not `stem(exe) + "-1.bin`?

launcher/innoextract.cpp
Laserlicht: Do not prefer it, because the filename can be really long (even with the standard names).

And the space is already very limited.

IvanSavenko: Can we show it in a separate line then? Or still too long?
> setup_heroes_of_might_and_magic_3_complete_4.0_(28740).exe
> SHA1: 66646a353b06417fa12c6384405688c84a315cc1
> File size: 12345667 bytes
>
> setup_heroes_of_might_and_magic_3_complete_4.0_(28740)-1.bin
> SHA1: c624e2071f4e35386765ab044ad5860ac245b7f4
> File size: 1234566789 bytes

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Not sure. But if it would be a 'random' copy error: It wouldn't be reproducible multiple times.

The idea was to copy the files from outside the sandbox to inside to avoid any access problems when working further with the files.

Maybe the new information (hashes, filesize) will help us to find errors.

But I think most problems caused by the sandboxing (remember one case we've got hashes calculated on phone which are correct).

But in general I don't not have any experience here. But it seems to be a android only issue.

launcher/innoextract.cpp
Laserlicht: Still too long. The currently displayed values (screenshots are not up to date) fits exactly in message box without scrolling on my phone.
No scrolling would be good for screenshots.

IvanSavenko: okay then

launcher/firstLaunch/firstlaunch_moc.cpp
IvanSavenko: If goal is to debug this case then we could also try to get file size / checksum of original files. If they don't match with copied files - then it is copy that is the problem.

Laserlicht: Good idea. Can also add this case.

PR #5030, opened 2024-12-03
Timed events objects removal

dydzio0614:
Feature that was ready for some time - did small cleanup and now it looks to be ready


mapeditor/mapsettings/timedevent.h
```diff
@@ -20,18 +22,27 @@ class TimedEvent : public QDialog
...
 	void on_TimedEvent_finished(int result);
 
+	void on_addObjectToDelete_clicked();
+
+	void on_removeObjectToDelete_clicked();
+
+	void onObjectPicked(const CGObjectInstance * obj);
+
+	void insertObjectToDelete(const CGObjectInstance * obj);
+
```
IvanSavenko, line 39:
```suggestion
	void on_addObjectToDelete_clicked();
	void on_removeObjectToDelete_clicked();
	void onObjectPicked(const CGObjectInstance * obj);
	void insertObjectToDelete(const CGObjectInstance * obj);
```

mapeditor/mapsettings/timedevent.cpp
```diff
@@ -89,10 +98,63 @@ void TimedEvent::on_TimedEvent_finished(int result)
...
+
+void TimedEvent::on_removeObjectToDelete_clicked()
+{
+	delete ui->deletedObjects->takeItem(ui->deletedObjects->currentRow());
+}
+
+void TimedEvent::onObjectPicked(const CGObjectInstance * obj)
+{
+	show();
+	dynamic_cast<QWidget *>(parent()->parent()->parent()->parent()->parent()->parent()->parent())->show();
```
IvanSavenko, line 136: wtf wtf wtf wtf wtf wtf wtf is this?

lib/mapping/CMapDefines.h
```diff
@@ -64,6 +69,14 @@ class DLL_LINKAGE CMapEvent
...
 		h & firstOccurrence;
 		h & nextOccurrence;
+		if(h.version >= Handler::Version::EVENT_OBJECTS_DELETION)
+		{
+			h & deletedObjectsInstances;
+		}
+		else
+		{
+			h & unused;
+		}
```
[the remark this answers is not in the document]
dydzio0614: IIRC some padding from the time when this feature was first created - related to removing another variable (and adjusting datatype to match deletedObjectsInstance is my mistake)

indeed, should not be here now, will change

server/processors/NewTurnProcessor.cpp
```diff
@@ -41,6 +41,8 @@ NewTurnProcessor::NewTurnProcessor(CGameHandler * gameHandler)
 
 void NewTurnProcessor::handleTimeEvents(PlayerColor color)
 {
+	std::vector<const CGObjectInstance*> removedObjects;
+
```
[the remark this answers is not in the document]
dydzio0614: will recheck

mapeditor/mapsettings/timedevent.cpp
dydzio0614: not my code :rofl:

IvanSavenko: @godric3 then?
In any case - can this be fixed or at least explained?

godric3: ah yes, guilty, copied from victory conditions (https://github.com/vcmi/vcmi/blob/develop/mapeditor/mapsettings/victoryconditions.cpp#L536)
If I remember correctly it was required to hide/show map setting dialog, to properly show map with selection layer/return to event settings. 
I don't know qt well enough to have it fixed quickly so I used same approach as found in codebase, I may look more into in on the weekend but can't promise anything

lib/networkPacks/NetPacksLib.cpp
Xilmi, line 1: Removal of 

PlayerState * p = gs->getPlayerState(beatenHero->tempOwner);
p->removeOwnedObject(beatenHero);

Breaks the game!

PR #5035, opened 2024-12-04
Fixes for mod management in Launcher

IvanSavenko:
- Fixes mod updates
- Fixes update/install/enable/etc buttons showing up on launcher startup
- Fixes mod list initialization if automatic repository checkout is off


Laserlicht:
Installation of translation from settings page also seems to doesn't work.

PR #5044, opened 2024-12-07
New container for BattleHexes. Refactors aiming for quick-battle better performance.

MichalZr6:
### Purpose of the PR

This is something I've been working on for some time. The PR introduces two major changes:  
- A new container for battlefield hexes (currently stored in `std::vector` or `std::set`)  
- Refactored battle AI code to improve performance  

---

### Rationale

The `BattleHexArray` is a container designed to efficiently store [battlefield hexes](https://github.com/vcmi/vcmi/blob/develop/docs/developers/BattleField.png) ranging from 0 to 186. The underlying data structure combines:  
- `std::vector` for fast access to functions like `front()`, `back()`, `pop()`, etc., while also maintaining the order of insertion.  
- `std::array` (used as `presenceFlags`) for O(1) lookups to maximize efficiency in that matter.

This hybrid approach was chosen after testing alternative containers, and it provided the best performance. Notably, `presenceFlags` does not use boolean values to avoid potential overhead from bitwise operations.  

Additionally, static caches have been introduced in cases where values can be precomputed for further optimization.

---

### Testing and Profiling

To test these changes, I prepared a custom test map featuring two heroes with large armies.  
obraz
obraz

Although I conducted earlier tests with other armies, but rebasing the branch at one point caused compatibility issues, and the initial test map could no longer be opened. For the current tests, Gelu, commanding a large Inferno army, attacks a town with a Fort (one arrow tower) and a garrisoned hero with a large Castle army. Both heroes have no spellbooks.  

While time measurements might be better placed elsewhere, the overhead of printing multiple output messages is negligible, especially when testing by restarting the build for each run.

---

##### **Develop Build (05.12.2024):**
```
Battle elapsed for: 7256 ms, 8015 ms, 7557 ms, 4748 ms, 5371 ms, 6265 ms, 8586 ms, 6581 ms, 7226 ms, 6589 ms, 7019 ms, 4840 ms, 5394 ms  

Avg: 6550 ms  
```

##### **BattleHexArray + Refactors:**
```
Battle elapsed for: 6476 ms, 4963 ms, 4801 ms, 6522 ms, 5589 ms, 6245 ms, 4712 ms, 7013 ms, 4523 ms, 5666 ms, 5369 ms, 4470 ms, 6704 ms  

Avg: 5680 ms  
Improvement: 13.3%  
```

##### **Static Caches at Class Scope:**
```
Battle elapsed for: 5570 ms, 5066 ms, 4542 ms, 4674 ms, 6117 ms, 5411 ms, 6042 ms, 8350 ms, 5915 ms, 6484 ms, 5668 ms, 7334 ms, 4044 ms, 5307 ms  

Avg: 5732 ms  
```

---

##### **Develop Build (05.12.2024):**
```
Battle elapsed for: 3897 ms, 3886 ms, 5440 ms, 5939 ms, 4560 ms, 6534 ms, 7246 ms, 5952 ms, 6453 ms, 5829 ms, 7150 ms, 5752 ms  

Avg: 5663 ms  
```

##### **Whole BattleHexArray Branch:**
```
Battle elapsed for: 4182 ms, 4255 ms, 4879 ms, 4662 ms, 5053 ms, 4856 ms  

Avg: 4648 ms  
Improvement: 17.9%  
```

---

### Summary

The overall results show some performance improvements, but there is undoubtedly room for further optimization. 

PS:
Note I’m not a professional programmer - just a self-taught, passionate enthusiast. Since such large-scale changes are not something I do daily, I greatly appreciate any insights or recommendations to enhance this work further.


lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1448,11 +1435,10 @@ AttackableTiles CBattleInfoCallback::getPotentiallyShootableHexes(const battle::
 	AttackableTiles at;
 	RETURN_IF_NOT_BATTLE(at);
 
-	if(attacker->hasBonusOfType(BonusType::SHOOTS_ALL_ADJACENT) && !vstd::contains(attackerPos.neighbouringTiles(), destinationTile))
+	if(attacker->hasBonusOfType(BonusType::SHOOTS_ALL_ADJACENT) && !BattleHexArray::neighbouringTilesCache[attackerPos].contains(destinationTile))
```
IvanSavenko, line 1438: Perhaps attackerPos.neighbouringTiles() should return `BattleHexArray::neighbouringTilesCache[pos]` instead? 

Old version of this code is more clear - user code should not be aware of implementation details.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1199,7 +1186,7 @@ std::pair<const battle::Unit *, BattleHex> CBattleInfoCallback::getNearestStack(
 
 	for(const battle::Unit * st : possible)
 	{
-		for(BattleHex hex : avHexes)
+		for(BattleHex hex : battleGetAvailableHexes(reachability, closest, false))
```
IvanSavenko, line 1189: Shouldn't old code be a little bit faster since this method is called only once?

lib/battle/BattleHex.h
```diff
@@ -83,29 +91,31 @@ struct DLL_LINKAGE BattleHex //TODO: decide if this should be changed to class f
 	BattleHex cloneInDirection(EDir dir, bool hasToBeValid = true) const;
 	BattleHex operator+(EDir dir) const;
 
-	/// returns all valid neighbouring tiles
-	std::vector<BattleHex> neighbouringTiles() const;
```
IvanSavenko, line 87: Already mentioned this, but I think this method should be restored and it should return cached / precomputed values - more easy to read code than what you have now

lib/battle/AccessibilityInfo.h
```diff
@@ -27,15 +27,15 @@ enum class EAccessibility
...
+	SIDE_COLUMN //used for first and last columns of hexes that are unavailable but war machines can stand there
 };
 
 
 using TAccessibilityArray = std::array<EAccessibility, GameConstants::BFIELD_SIZE>;
 
 struct DLL_LINKAGE AccessibilityInfo : TAccessibilityArray
 {
-	std::map<BattleHex, ui8> destructibleEnemyTurns;
+	const std::array<int8_t, GameConstants::BFIELD_SIZE> * destructibleEnemyTurns = nullptr;
```
IvanSavenko, line 38: Another reminder that this may break checks on invalid hexes, e.g. towers.

Also - why is this a pointer now?

client/NetPacksClient.cpp
```diff
@@ -813,11 +813,21 @@ void ApplyClientNetPackVisitor::visitBattleNextRound(BattleNextRound & pack)
 	callBattleInterfaceIfPresentForBothSides(cl, pack.battleID, &IBattleEventsReceiver::battleNewRound, pack.battleID);
 }
 
+uint64_t timeElapsed(std::chrono::time_point<std::chrono::high_resolution_clock> start)
+{
+	auto end = std::chrono::high_resolution_clock::now();
+
+	return std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
+}
+
```
IvanSavenko, line 822: Debug code?

If you want to keep it - then probably better to reuse `CStopWatch`. And potentially - upgrade it to use `steady_clock`.

Also - `high_resolution_clock` is problematic since it is not guaranteed to be steady...

AI/StupidAI/StupidAI.cpp
```diff
@@ -310,9 +310,9 @@ BattleAction CStupidAI::goTowards(const BattleID & battleID, const CStack * stac
 		}
 	}
 
-	BattleHex bestNeighbor = hexes.front();
+	BattleHex bestneighbour = hexes.front();
```
IvanSavenko, line 313: Why was this one renamed?
```suggestion
	BattleHex bestNeighbour = hexes.front();
```

lib/battle/BattleHexArray.h
```diff
@@ -0,0 +1,321 @@
...
+		}
+
+		if(contains(hex))
+			return;
+
+		presenceFlags[hex] = 1;
+		internalStorage[index] = hex;
+	}
+
+	iterator BattleHexArray::insert(iterator pos, BattleHex hex) noexcept
```
IvanSavenko, line 128: There is no need to specify class name for methods that are defined in place
```suggestion
	iterator insert(iterator pos, BattleHex hex) noexcept
```

lib/spells/effects/Summon.cpp
```diff
@@ -207,4 +207,4 @@ EffectTarget Summon::transformTarget(const Mechanics * m, const Target & aimPoin
 }
 }
 
-VCMI_LIB_NAMESPACE_END
+VCMI_LIB_NAMESPACE_END
```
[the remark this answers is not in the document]
MichalZr6: This might have been some leftover after resolving conflicts.

lib/battle/BattleHexArray.h
```diff
@@ -0,0 +1,321 @@
...
+
+	template <typename Container, typename = std::enable_if_t<
+		std::is_convertible_v<typename Container::value_type, BattleHex>>>
+		void merge(const Container & container) noexcept
+	{
+		for(auto value : container)
+		{
+			insert(value);
+		}
+	}
```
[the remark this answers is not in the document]
MichalZr6: Ok. Will rename.

lib/spells/effects/Moat.cpp
```diff
@@ -43,8 +43,12 @@ static void serializeMoatHexes(JsonSerializeFormat & handler, const std::string
 			JsonArraySerializer inner = outer.enterArray(outerIndex);
 			inner.syncSize(moatHexes.at(outerIndex), JsonNode::JsonType::DATA_INTEGER);
 
+			BattleHex hex;
 			for(size_t innerIndex = 0; innerIndex < inner.size(); innerIndex++)
-				inner.serializeInt(innerIndex, moatHexes.at(outerIndex).at(innerIndex));
+			{
+				inner.serializeInt(innerIndex, hex);
+				moatHexes.at(outerIndex).set(innerIndex, hex);
```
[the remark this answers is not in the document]
MichalZr6: I tried to implement serialization for BattleHexArray in BinarySerializer.h the way ex. std::vector is done, but it gave me strange exceptions on length. Is this the way or just fix the current code to work as expected ?

lib/battle/BattleHexArray.h
```diff
@@ -0,0 +1,321 @@
...
+	using difference_type = typename StorageType::difference_type;
+	using iterator = typename StorageType::iterator;
+	using const_iterator = typename StorageType::const_iterator;
+	using reverse_iterator = typename StorageType::reverse_iterator;
+	using const_reverse_iterator = typename StorageType::const_reverse_iterator;
+
+	using ArrayOfBattleHexArrays = std::array<BattleHexArray, GameConstants::BFIELD_SIZE>;
+
+	static const ArrayOfBattleHexArrays neighbouringTilesCache;
+	static const std::map<BattleSide, ArrayOfBattleHexArrays> neighbouringTilesDblWide;
```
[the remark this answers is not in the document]
MichalZr6: well... `neighbouringTilesCache` was its original naming. But of course you are right. These are precomputed values. I firstly done them in class scope, then in function scope and currently they are in class scope again (wanted to see if there is performance gain in using those statics in class scope), however I think there might be the risk of ["static initialization order fiasco"](https://isocpp.org/wiki/faq/ctors#static-init-order-on-first-use-members) and probably it would be best to initialize them in function scope as I did before.

lib/battle/BattleHexArray.h
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	[[nodiscard]] inline BattleHexArray::const_reverse_iterator BattleHexArray::rend() const noexcept
+	{
+		return const_reverse_iterator(begin());
+	}
+
+private:
+	StorageType internalStorage;
+	std::array<uint8_t, totalSize> presenceFlags = {};
```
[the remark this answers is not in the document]
MichalZr6: Just tested with steady_clock and bitset (internalStorage remained std::vector for now). It shows some improvement, but will have to run new measurments on develop branch with steady_clock to be sure.

lib/spells/effects/Moat.cpp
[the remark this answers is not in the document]
IvanSavenko: This is used not by binary seralizer, but by json serializer, which is indeed pain to use... Having 50 lines of code to serialize single field with it is nothing new.

Will take a look as to how your case can be fixed.

lib/battle/ReachabilityInfo.h
```diff
@@ -30,13 +32,16 @@ struct DLL_LINKAGE ReachabilityInfo
 		bool flying = false;
 		bool ignoreKnownAccessible = false; //Ignore obstacles if it is in accessible hexes
 		bool bypassEnemyStacks = false; // in case of true will count amount of turns needed to kill enemy and thus move forward
-		std::vector<BattleHex> knownAccessible; //hexes that will be treated as accessible, even if they're occupied by stack (by default - tiles occupied by stack we do reachability for, so it doesn't block itself)
-		std::map<BattleHex, ui8> destructibleEnemyTurns; // hom many turns it is needed to kill enemy on specific hex
+		const BattleHexArray * knownAccessible; //hexes that will be treated as accessible, even if they're occupied by stack (by default - tiles occupied by stack we do reachability for, so it doesn't block itself)
+		std::array<int8_t, GameConstants::BFIELD_SIZE> destructibleEnemyTurns; // how many turns it is needed to kill enemy on specific hex (index <=> hex)
```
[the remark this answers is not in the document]
MichalZr6: You mean knownAccessible ? The only place it is initialized is in `ReachabilityInfo::Parameters` constructor via `battle::Unit::getHexes` which returns precomputed, occupied hexes by units. It might have negative values, but then `AccessibilityInfo CBattleInfoCallback::getAccessibility` does check to filter those.

lib/battle/AccessibilityInfo.h
MichalZr6: `destructibleEnemyTurns` in `AccessibilityInfo` is just a view on `destructibleEnemyTurns` from `ReachabilityInfo::Parameters`
```cpp
ReachabilityInfo CBattleInfoCallback::getReachability(const ReachabilityInfo::Parameters & params) const
{
	if(params.flying)
		return getFlyingReachability(params);
	else
	{
		auto accessibility = getAccessibility(params.knownAccessible);

		accessibility.destructibleEnemyTurns = & params.destructibleEnemyTurns;

		return makeBFS(accessibility, params);
	}
}
```

lib/battle/ReachabilityInfo.h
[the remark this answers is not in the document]
IvanSavenko: Before `destructibleEnemyTurns[-1] = X` was legal. Now it is not since arrays don't have negative indices. And arrow towers during sieges are located on such hexes.

If you are sure that this case works - then OK. Just making sure since this was a problem in the past.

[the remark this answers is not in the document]
MichalZr6: Re-checked. All places that access this container are done with pre-check for hex validity.

lib/battle/BattleHex.h
MichalZr6: Done but not the way I thought it would be. Tried to implement it all (precomputed static containers) into `BattleHex` but failed due to circular dependency with `ArrayOfBattleHexArrays`.

MichalZr6:
Some updated breakdown for this PR:
_RelWithDebInfo Mode (/O2 /Ob1)_

#### **Beta Branch - 31.12.2024**
```
Battle elapsed for: 4201 ms, 6052 ms, 3199 ms, 3363 ms, 4731 ms, 3208 ms, 2719 ms, 3472 ms, 3298 ms, 3421 ms
Avg: 3766 ms
```

#### **BattleHexArray on branch beta from 31.12.2024 with std::bitset for BattleHexArray presenceFlags**
```
Battle elapsed for: 3316 ms, 2871 ms, 3137 ms, 3019 ms, 3278 ms, 2669 ms, 3213 ms, 2952 ms, 3706 ms, 3743 ms, 3161 ms
Avg: 3179 ms
```

#### **BattleHexArray based on branch beta from 31.12.2024 with boost::small_vector as BattleHexArray's internal storage**
```
Battle elapsed for: 3091 ms, 3026 ms, 2553 ms, 2863 ms, 2891 ms, 2642 ms, 3402 ms, 3582 ms, 3607 ms, 3991 ms, 4155 ms, 2837 ms
Avg: 3123 ms
```

#### **As above but without reserve(totalSize) in default constructor as small_vector<BattleHex, 8> is used now**
``` 
Battle elapsed for: 3234 ms, 2371 ms, 2672 ms, 3399 ms, 3180 ms, 3511 ms, 3658 ms, 3736 ms, 2692 ms, 2997 ms, 4255 ms
Avg: 3145 ms
```

#### **As above but refactor: pass BattleHexArray by-reference instead of by-copy in some functions within codebase**
```
Battle elapsed for: 2645 ms, 2572 ms, 2565 ms, 2248 ms, 3312 ms, 4034 ms, 3697 ms, 2715 ms, 2993 ms, 3604 ms, 2805 ms
Avg: 2980 ms
```

#### **Refactor BattleHex Class:**
- Move short functions to header for inline mechanism
- Make BattleHex a class with a private hex value
- Move getClosestTile implementation back to BattleHex
- Access static precalculated data from BattleHexArray via BattleHex (direct static containers in BattleHex were not possible due to circular dependencies)

```
Battle elapsed for: 2753 ms, 2764 ms, 3163 ms, 2849 ms, 2878 ms, 3424 ms, 3029 ms, 2851 ms, 2549 ms, 2690 ms, 3500 ms
Avg: 2945 ms
```

lib/battle/AccessibilityInfo.h
IvanSavenko, line 38: To be honest I would rather use either direct member (without pointer) or shared_ptr to guarantee that this pointer is always valid

lib/battle/BattleHexArray.h
```diff
@@ -0,0 +1,317 @@
...
+
+	template <typename Serializer>
+	void serialize(Serializer & s)
+	{
+		s & internalStorage;
+		if(!internalStorage.empty() && presenceFlags[internalStorage.front()] == 0)
+		{
+			for(auto hex : internalStorage)
+				presenceFlags[hex] = 1;
+		}
```
IvanSavenko, line 205:
```suggestion
		if(!s.saving)
		{
			for(auto hex : internalStorage)
				presenceFlags[hex] = 1;
		}
```

lib/battle/BattleHexArray.cpp
```diff
@@ -0,0 +1,132 @@
...
+		ret[hex].resize(6);
+
+		for(auto dir : BattleHex::hexagonalDirections())
+			ret[hex].set(dir, BattleHex(hex).cloneInDirection(dir, false));
+	}
+
+	return ret;
+}
+
+BattleHexArray::ArrayOfBattleHexArrays BattleHexArray::precalculateNeighbouringTilesDblWide(BattleSide side)
```
IvanSavenko, line 86:
```suggestion
BattleHexArray::ArrayOfBattleHexArrays BattleHexArray::precalculateNeighbouringTilesDoubleWide(BattleSide side)
```

lib/battle/BattleHex.h
```diff
@@ -64,52 +68,177 @@ struct DLL_LINKAGE BattleHex //TODO: decide if this should be changed to class f
...
+		setXY(x, y);
+	}
+	BattleHex(std::pair<si16, si16> xy)
+	{
+		setXY(xy);
+	}
+	operator si16() const
+	{
+		return hex;
+	}
```
IvanSavenko, line 88: Do we need it? Implicit conversion often lead to accidental type conversions

lib/spells/effects/Moat.cpp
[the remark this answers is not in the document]
IvanSavenko: I think in this case it is rather simple:
```suggestion
				BattleHex hex = moatHexes.at(outerIndex).at(innerIndex);
				inner.serializeInt(innerIndex, hex);
				moatHexes.at(outerIndex).set(innerIndex, hex);
```

All you need is to initialize `hex` correctly to handle serialization.

Same change in another place that uses json serializer

client/battle/BattleAnimationClasses.h
```diff
@@ -143,7 +143,7 @@ class MovementAnimation : public StackMoveAnimation
 private:
 	int moveSoundHandler; // sound handler used when moving a unit
 
-	std::vector<BattleHex> destTiles; //full path, includes already passed hexes
+	const BattleHexArray & destTiles; //full path, includes already passed hexes
```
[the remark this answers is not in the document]
MichalZr6: I would never have done it if I wasn't sure. Just found it by-the-way and made that change. Perhaps it should be a const pointer ? Or doesn't matter ?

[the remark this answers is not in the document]
IvanSavenko: Problem is, it is very easy to write code like this:
```cpp
BattleHexArray hexes;
hexes.add(someHex);
return new MovementAnimation(hexes);
```
Compiler won't warn you about anything in here.
Code does not looks off in any way
It might even work sometimes.

Only to cause strange bugs later.

So for cases like this I'd rather eat potential overhead of like 1 mks but ensure that something like this would never happen.

Another option is to use shared_ptr and ALWAYS pass it as shared_ptr. Meaning, member is shared_ptr, constructor receives shared_ptr, etc.

lib/battle/AccessibilityInfo.h
MichalZr6: It is only used as a view for `destructibleEnemyTurns` that is in `ReachabilityInfo::Parameters`. That's why `const *`. It also gave surprisingly nice performance gain. What do you mean about "this pointer is always valid" ? So the original object will not get destroyed somewhere externally ? Shared pointer will decrease performance (perhaps not much, though - will test) as it needs reference counting.

IvanSavenko: I just don't want to track down some weird issues if this assumption change (that this is only a view on entity with longer lifetime).

shared_ptr will at the very least ensure that this object will never become invalid no matter how code was changed outside of this class.

client/battle/BattleAnimationClasses.h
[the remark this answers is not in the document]
MichalZr6: > So for cases like this I'd rather eat potential overhead of like 1 mks but ensure that something like this would never happen.
Yeah, looking at the class its in it's probably not worth it. Reverting.

lib/battle/BattleHex.h
MichalZr6: I think it can be removed and replaced with `toInt()` for example.
And what about this code:
```cpp
void CUnitState::serializeJson(JsonSerializeFormat & handler)
{
	// ...
	si16 posValue = position.toInt();
	handler.serializeInt("position", posValue);
	position = posValue;
}
```
is it correct ?

IvanSavenko: `toInt()` - yes, sounds good.

Code looks correct.

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -1048,20 +1040,20 @@ AccessibilityInfo CBattleInfoCallback::getAccessibility() const
 
 AccessibilityInfo CBattleInfoCallback::getAccessibility(const battle::Unit * stack) const
 {
-	return getAccessibility(battle::Unit::getHexes(stack->getPosition(), stack->doubleWide(), stack->unitSide()));
+	return getAccessibility(& battle::Unit::getHexes(stack->getPosition(), stack->doubleWide(), stack->unitSide()));
 }
 
-AccessibilityInfo CBattleInfoCallback::getAccessibility(const std::vector<BattleHex> & accessibleHexes) const
+AccessibilityInfo CBattleInfoCallback::getAccessibility(const BattleHexArray * accessibleHexes) const
```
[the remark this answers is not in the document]
MichalZr6: Should be reference. Fixing.

IvanSavenko:
Looks OK, so let's merge and see how it goes.

Although this PR already approaches limits of what can be reviewed, especially considering its relatively high complexity, so not sure if I have checked everything...

PR #5048, opened 2024-12-07
remove chains of `parent()->parent()` in map editor

godric3:


IvanSavenko (approved):
Thanks for fixing those. Looks OK, will merge a bit later into develop.

PR #5049, opened 2024-12-07
fix registry check for systems with WOW6432Node

Laserlicht:
... for installing with installed gog version...

### DIFF
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -304,11 +304,13 @@ QString FirstLaunchView::getHeroesInstallDir()
 		{ "HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\New World Computing\\Heroes of Might and Magic III\\1.0",  "AppPath" }, // some localized H3 on x64 system
 	};
 
-	for(auto & regKey : regKeys)
+	QHash<QString, QString>::const_iterator i = regKeys.constBegin();
+	while (i != regKeys.constEnd())
 	{
-		QString path = QSettings(regKey, QSettings::NativeFormat).value(regKeys[regKey]).toString();
+		QString path = QSettings(i.key(), QSettings::NativeFormat).value(i.value()).toString();
 		if(!path.isEmpty())
 			return path;
+		++i;
 	}
 #endif
 	return QString{};
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -298,8 +298,14 @@ QString FirstLaunchView::getHeroesInstallDir()
 	QString gogPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\GOG.com\\Games\\1207658787", QSettings::NativeFormat).value("path").toString();
 	if(!gogPath.isEmpty())
 		return gogPath;
+	gogPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\GOG.com\\Games\\1207658787", QSettings::NativeFormat).value("path").toString();
+	if(!gogPath.isEmpty())
+		return gogPath;
 
 	QString cdPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\New World Computing\\Heroes of Might and Magic® III\\1.0", QSettings::NativeFormat).value("AppPath").toString();
+	if(!cdPath.isEmpty())
+		return cdPath;
+	cdPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\New World Computing\\Heroes of Might and Magic® III\\1.0", QSettings::NativeFormat).value("AppPath").toString();
 	if(!cdPath.isEmpty())
 		return cdPath;
 #endif
--- a/launcher/firstLaunch/firstlaunch_moc.cpp
+++ b/launcher/firstLaunch/firstlaunch_moc.cpp
@@ -295,7 +295,7 @@ bool FirstLaunchView::heroesDataDetect()
 QString FirstLaunchView::getHeroesInstallDir()
 {
 #ifdef VCMI_WINDOWS
-	std::map<QString, QString> regKeys = {
+	QHash<QString, QString> regKeys = {
 		{ "HKEY_LOCAL_MACHINE\\SOFTWARE\\GOG.com\\Games\\1207658787",                                            "path"    }, // Gog on x86 system
 		{ "HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\GOG.com\\Games\\1207658787",                               "path"    }, // Gog on x64 system
 		{ "HKEY_LOCAL_MACHINE\\SOFTWARE\\New World Computing\\Heroes of Might and Magic® III\\1.0",              "AppPath" }, // H3 Complete on x86 system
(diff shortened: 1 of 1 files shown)


GeorgeK1ng:
Please add also strings without ® in registry. I have just tested original Czech distro setup from 2007 and it creates keys without ®

```
	QString cdPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\New World Computing\\Heroes of Might and Magic III\\1.0", QSettings::NativeFormat).value("AppPath").toString();
	if(!cdPath.isEmpty())
		return cdPath;
	cdPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\New World Computing\\Heroes of Might and Magic III\\1.0", QSettings::NativeFormat).value("AppPath").toString();
	if(!cdPath.isEmpty())
		return cdPath;
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -296,12 +296,18 @@ QString FirstLaunchView::getHeroesInstallDir()
 {
 #ifdef VCMI_WINDOWS
 	QString gogPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\GOG.com\\Games\\1207658787", QSettings::NativeFormat).value("path").toString();
+	if(!gogPath.isEmpty())
+		return gogPath;
+	gogPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\GOG.com\\Games\\1207658787", QSettings::NativeFormat).value("path").toString();
```
kambala-decapitator, line 301: please don't copy-paste identical suffix

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -295,13 +295,21 @@ bool FirstLaunchView::heroesDataDetect()
 QString FirstLaunchView::getHeroesInstallDir()
 {
 #ifdef VCMI_WINDOWS
-	QString gogPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\GOG.com\\Games\\1207658787", QSettings::NativeFormat).value("path").toString();
-	if(!gogPath.isEmpty())
-		return gogPath;
+	std::map<QString, QString> regKeys = {
```
kambala-decapitator, line 298: Mixing std with Qt types is usually not a good idea. Use QHash instead.

launcher/firstLaunch/firstlaunch_moc.cpp
kambala-decapitator: @Laserlicht this is not done yet

Laserlicht: Yes. I've used another approach.

I prefer a data table of exact registry keys over string manipulation / prefix here. IMO it's better, because its the full key you can read out from registry:
grafik

kambala-decapitator: With the current approach you still need to replace double backslashes with a single one to put it in regedit, I don't see any advantage.

The proper way here is to use Cartesian product of arrays: you have some prefixes and some suffixes and you want to get all possible combinations. The intent will be clear to anyone reading the code.

We're writing production code here, please avoid any form of copy-paste.

Laserlicht: > With the current approach you still need to replace double backslashes with a single one to put it in regedit, I don't see any advantage.

But even that's much faster and error resistant than clicking trough the tree. Clicking trough the registry is sometimes painful. :D

I've still another opinion in this specific case (not in general) - I do not want to invest time to implement suggested solution.

IvanSavenko: While kambala is not wrong, such approach would probably result in more complex and less readable code in this particular case.

Unless number of potential locations growth, current approach is fine IMO

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -295,13 +295,23 @@ bool FirstLaunchView::heroesDataDetect()
 QString FirstLaunchView::getHeroesInstallDir()
 {
 #ifdef VCMI_WINDOWS
-	QString gogPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\GOG.com\\Games\\1207658787", QSettings::NativeFormat).value("path").toString();
-	if(!gogPath.isEmpty())
-		return gogPath;
+	QHash<QString, QString> regKeys = {
```
IvanSavenko, line 298: Perhaps QVector<QPair<>> is more suitable here?
It is not actually a map / dictionary, but a collection of pairs of strings - path in registry and name of registry key.

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -295,13 +295,23 @@ bool FirstLaunchView::heroesDataDetect()
...
+		{ "HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\New World Computing\\Heroes of Might and Magic® III\\1.0", "AppPath" }, // H3 Complete on x64 system
+		{ "HKEY_LOCAL_MACHINE\\SOFTWARE\\New World Computing\\Heroes of Might and Magic III\\1.0",               "AppPath" }, // some localized H3 on x86 system
+		{ "HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\New World Computing\\Heroes of Might and Magic III\\1.0",  "AppPath" }, // some localized H3 on x64 system
+	};
 
-	QString cdPath = QSettings("HKEY_LOCAL_MACHINE\\SOFTWARE\\New World Computing\\Heroes of Might and Magic® III\\1.0", QSettings::NativeFormat).value("AppPath").toString();
-	if(!cdPath.isEmpty())
-		return cdPath;
+	QHash<QString, QString>::const_iterator i = regKeys.constBegin();
+	while (i != regKeys.constEnd())
```
IvanSavenko, line 308: Why not use for loop?

Laserlicht: Was First approach (with std::map). But not working with QHash.

kambala-decapitator: Iirc for-in with QHash yields only values. But you could write

    for (auto i = hash.constBegin(); it != hash.constEnd(); ++it)

launcher/firstLaunch/firstlaunch_moc.cpp
kambala-decapitator: The suggested solution is indeed complex 😄 

for p in prefixes
  for s in suffixes
    keys += p+s

IvanSavenko: and here is the rest of the owl:
```cpp
QStringList prefixes = {
 "HKEY_LOCAL_MACHINE\\SOFTWARE",
 "HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node"
};

QVector<QPair<QString, QString>> suffixes = {
	{ "\\GOG.com\\Games\\1207658787",                               "path"    }, // Gog on x86 system
	{ "\\New World Computing\\Heroes of Might and Magic® III\\1.0", "AppPath" }, // H3 Complete on x86 system
	{ "\\New World Computing\\Heroes of Might and Magic III\\1.0",  "AppPath" }, // some localized H3 on x86 system
}

for (const auto & p : prefixes)
{
	for (const auto & s : suffixes)
	{
		QString actualKey = p + s;
		QString path = QSettings(actualKey, QSettings::NativeFormat).value(s.value()).toString();
		if(!path.isEmpty())
			return path;
	}
}
```

Does not looks to be more readable than current version

launcher/firstLaunch/firstlaunch_moc.cpp
Laserlicht: Yes `QVector<QPair<...>>` is much better here. :)

PR #5050, opened 2024-12-08
Graphic fixes

Laserlicht:
- fix cell shades with xbrz
- center round indicator

### DIFF
--- a/client/battle/BattleFieldController.cpp
+++ b/client/battle/BattleFieldController.cpp
@@ -114,7 +114,7 @@ BattleFieldController::BattleFieldController(BattleInterface & owner):
 
 	//preparing cells and hexes
 	cellBorder = GH.renderHandler().loadImage(ImagePath::builtin("CCELLGRD.BMP"), EImageBlitMode::COLORKEY);
-	cellShade = GH.renderHandler().loadImage(ImagePath::builtin("CCELLSHD.BMP"), EImageBlitMode::SIMPLE);
+	cellShade = GH.renderHandler().loadImage(ImagePath::builtin("CCELLSHD.BMP"), EImageBlitMode::ONLY_SHADOW);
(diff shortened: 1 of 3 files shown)


client/battle/BattleFieldController.cpp
```diff
@@ -114,7 +114,7 @@ BattleFieldController::BattleFieldController(BattleInterface & owner):
 
 	//preparing cells and hexes
 	cellBorder = GH.renderHandler().loadImage(ImagePath::builtin("CCELLGRD.BMP"), EImageBlitMode::COLORKEY);
-	cellShade = GH.renderHandler().loadImage(ImagePath::builtin("CCELLSHD.BMP"), EImageBlitMode::SIMPLE);
+	cellShade = GH.renderHandler().loadImage(ImagePath::builtin("CCELLSHD.BMP"), EImageBlitMode::ONLY_SHADOW);
```
IvanSavenko, line 117: Thanks for investigation, but this is not where problem comes from initially. Proper fix would be this:
```diff
diff --git a/client/renderSDL/ImageScaled.cpp b/client/renderSDL/ImageScaled.cpp
index 219efb999..c28d73025 100644
--- a/client/renderSDL/ImageScaled.cpp
+++ b/client/renderSDL/ImageScaled.cpp
@@ -137,7 +137,6 @@ void ImageScaled::prepareImages()

        switch(blitMode)
        {
-               case EImageBlitMode::SIMPLE:
                case EImageBlitMode::WITH_SHADOW:
                case EImageBlitMode::ONLY_SHADOW:
                case EImageBlitMode::WITH_SHADOW_AND_OVERLAY:
```
basically, game would upscale image as a whole (including shadow), and then - also upscale shadow separately, resulting in two part of image, with both having shadow.

Laserlicht: Thanks for fixing the fix. :)

PR #5060, opened 2024-12-09
translate strings for mapeditor

Laserlicht:
Only `inspector.cpp` is missing.

No interest in doing this as it is too much work...

@GeorgeK1ng


GeorgeK1ng:
But where should I add translation? I don't see any source in commit for them?

IvanSavenko:
Will merge beta PR's today and regenerate translations

Laserlicht:
> But where should I add translation? I don't see any source in commit for them?

Yes. Don't want to generate a conflict with Ivan's PR.

Issue #5059, opened 2024-12-09
left or right click map drag option bugged, causing options menu to lag


ToRRent1812:
latest develop build.
Tested on linux/opensuse tumbleweed with xbrz 2x
https://www.youtube.com/watch?v=KB2gt9HKlBU

Issue #5040, opened 2024-12-05
victory condition: transport a specific artifact


Povelitelh3:
The enemy AI fulfills this condition, but I get the victory!
![изображение](https://github.com/user-attachments/assets/cf407196-a281-4318-8eab-28a4e543dd41)
![изображение](https://github.com/user-attachments/assets/7f0623ca-1a61-4bfe-9125-0ef81886140e)

[Victory loss condition.zip](https://github.com/user-attachments/files/18024940/Victory.loss.condition.zip)

Issue #5023, opened 2024-12-02
Game hangs after winning 


edeksumo:
**Describe the bug**
After winning random map on hot seat (human players are allies) game skips into next turn and hangs. Window with point didn't show up.

**Version**
 Daily build

Issue #5015, opened 2024-12-01
xbrz - artifacts


Laserlicht:
**Describe the bug**
Some artifacts are visible if xbrz is enabled:
![grafik](https://github.com/user-attachments/assets/201bb407-6988-4df4-804b-90fb7561ebfb)

Is there a way to reduce them?

IvanSavenko:
Yes, possible, but needs minor tweak in xbrz code.

For RGBA images xbrz assumes image border to be transparent. 
For RGB xbrz instead uses different policy, where it assumes image border to have same color as nearest pixel inside image.

We need to expose that to library API, and use their rgb mode for such images.

PR #5061, opened 2024-12-10
Fixes for issues found in current beta

IvanSavenko:
- Added simple fix for xbrz upscaling artifacts, most notable on window border. Not sure if heuristics check would correctly identify all cases, but looks OK so far.
- Fix installing mods from outside of mod view (like from settings or from setup) only installing mod dependencies, and not actual requested mod
- Launcher will now prefer displaying local data for mods, unless there is newer version of mod in repository
- Block install / update buttons for mods that are already being downloaded, to prevent double-downloads
- Mod enable button is now correctly blocked for mods with missing dependencies
- fixes "transport artifact" victory condition triggering as victory for human players even when artifact was transported by enemy AI if "standard victory" is also enabled on this map
- Quick workaround for constantly switching input mode in settings window
- Fix game not ending on victory of allied human players in hotseat

Fixes:
- #3275
- #5015
- #5023
- #5040
- #5059

### DIFF
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -430,9 +430,10 @@ void ApplyClientNetPackVisitor::visitPlayerEndsGame(PlayerEndsGame & pack)
 {
 	callAllInterfaces(cl, &IGameEventsReceiver::gameOver, pack.player, pack.victoryLossCheckResult);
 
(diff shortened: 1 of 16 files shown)


Laserlicht:
Another issue detected:
- directly after chronicles installation: it will be disabled by default - has to be enabled first by hand

ToRRent1812:
Maybe change unlimited to something less intrusive like 100.

PR #5066, opened 2024-12-11
Change saves default settings to unlimited

GeorgeK1ng:
I have noticed 2 complains with "VCMI is broken and doesn't work correctly" from new players.

1) Saves are broken, only five are there.
       - change 5 to 0 to do not confuse player at beginning. 

2) I can't do anything before music stops playing in battle
       - even you can skip it by click, would be better to have this enabled by default so music play and player can start gaming


Laserlicht:
Not sure about the changes:

> Saves are broken, only five are there.

This means unlimited, right? Will generate a lot of files. Keep in mind, that every file(-header) is loaded when opening "load files" menu. This could lead in long loading times for players which played a lot of games.

> I can't do anything before music stops playing in battle

Personally I prefer H3 way by default.

GeorgeK1ng:
Yes unlimited, same as H3 have. What machine could have issue with that? My old CPU K7 600MHz have never issue with original game with tons of day saves.

That battle music is questionable of course.

dydzio0614:
i think unlimited saves make sense, since there is save wrapping so first or last on the list is not necessarily newest... this is not very player friendly

i would probably keep music as it currently is personally

Laserlicht:
> Yes unlimited, same as H3 have. What machine could have issue with that? My old CPU K7 600MHz have never issue with original game with tons of day saves.

To be honest: I've never tested it. But I've some discussions about 6000+ Mappacks in mind which causes slow load.
And a save is basically a map + additional infos.

Maybe someone is interested in testing? Just enable unlimited save and let game play itself over night on a map that no one can win. Should generate a lot of saves. And then compare loading menu times.

GeorgeK1ng:
I have not encountered bigger issues with 6000+ map pack and all VCMI map packs enabled together. maybe 2-3s delay when loading them. But sorting was not issue at all.

I will try VCMI on friday on 2007 laptop, which is not much powerfull. Mostly android phones from 2012 are quicker.

And you can always change that settings. To any wanted number. This is mainly to no confuse newjoiners. Experienced players won't kill game and load previous save.

GeorgeK1ng:
I have reverted that battle music default settings. Just to have unlimited saves

GeorgeK1ng:
Closing for now

PR #5078, opened 2024-12-13
Update swedish.ts

XCOM-HUB:


XCOM-HUB:
@heroesiiifan 

Could you change the locale code to sv_SE instead of se_SE? The correct locale code for the Swedish language is sv_SE. I would appreciate it much.

IvanSavenko:
> Could you change the locale code to sv_SE instead of se_SE? The correct locale code for the Swedish language is sv_SE. I would appreciate it much.

Where we use this code? You can change it in `swedish.ts` on your own, which you did in this pull request.

However we don't actually use it anywhere in vcmi. I think this code is only used to hint Qt Linguist that this is Swedish translation. VCMI loads file named `swedish.ts` irregardless of what code is set in there.

PR #5087, opened 2024-12-14
fix boost 1.87

Laserlicht:


Laserlicht:
@IvanSavenko can we Merge?

To test newest beta with flatpak.

PR #5101, opened 2024-12-16
Attempt to fix Chronicles installation on Android

IvanSavenko:
Seems to be working now, at least on my phone.

- Simplified detection of which Chronicles was provided by player (by examining extracted content instead of searching in binary file)
- Fixed enabling of 1st imported Chronicle (mod was enabled before mod manager reloaded its state)
- Added logging to help with tracking down similar issues in the future

### DIFF
--- a/launcher/modManager/chroniclesextractor.cpp
+++ b/launcher/modManager/chroniclesextractor.cpp
@@ -41,37 +41,23 @@ void ChroniclesExtractor::removeTempDir()
 	tempDir.removeRecursively();
 }
 
-int ChroniclesExtractor::getChronicleNo(QFile & file)
+int ChroniclesExtractor::getChronicleNo()
 {
-	if(!file.open(QIODevice::ReadOnly))
+	for (size_t i = 1; i < chronicles.size(); ++i)
 	{
-		QMessageBox::critical(parent, tr("The file cannot be opened"), file.errorString());
-		return 0;
+		QString chronicleName = chronicles.at(i);
+
+		QStringList appDirCandidates = tempDir.entryList({"app"}, QDir::Filter::Dirs);
+		QDir appDir = tempDir.filePath(appDirCandidates.front());
+
+		QStringList chroniclesDirCandidates = appDir.entryList({chronicleName}, QDir::Filter::Dirs);
+
+		if (!chroniclesDirCandidates.empty())
+			return i;
 	}
 
-	QByteArray magic{"MZ"};
-	QByteArray magicFile = file.read(magic.length());
-	if(!magicFile.startsWith(magic))
-	{
-		QMessageBox::critical(parent, tr("Invalid file selected"), tr("You have to select a gog installer file!"));
-		return 0;
-	}
-
-	QByteArray dataBegin = file.read(1'000'000);
-	int chronicle = 0;
-	for (const auto& kv : chronicles) {
-		if(dataBegin.contains(kv.second))
-		{
-			chronicle = kv.first;
-			break;
-		}
-	}
-	if(!chronicle)
-	{
-		QMessageBox::critical(parent, tr("Invalid file selected"), tr("You have to select a Heroes Chronicles installer file!"));
-		return 0;
-	}
-	return chronicle;
+	QMessageBox::critical(parent, tr("Invalid file selected"), tr("You have to select a Heroes Chronicles installer file!"));
+	return 0;
 }
 
 bool ChroniclesExtractor::extractGogInstaller(QString file)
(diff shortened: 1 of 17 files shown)


IvanSavenko:
Seems to be working now.
Changes:
- Simplified detection of which Chronicles was provided by player (by examining extracted content instead of searching in binary file)
- Fixed enabling of 1st imported Chronicle (mod was enabled before mod manager reloaded its state)
- Added logging to help with tracking down similar issues in the future

launcher/modManager/chroniclesextractor.cpp
```diff
@@ -41,37 +41,23 @@ void ChroniclesExtractor::removeTempDir()
...
 
-	QByteArray magic{"MZ"};
-	QByteArray magicFile = file.read(magic.length());
-	if(!magicFile.startsWith(magic))
-	{
-		QMessageBox::critical(parent, tr("Invalid file selected"), tr("You have to select a gog installer file!"));
-		return 0;
-	}
+		QStringList appDirCandidates = tempDir.entryList({"app"}, QDir::Filter::Dirs);
+		QDir appDir = tempDir.filePath(appDirCandidates.front());
```
kambala-decapitator, line 51: Can the list be empty?

launcher/modManager/chroniclesextractor.cpp
```diff
@@ -147,14 +133,13 @@ void ChroniclesExtractor::createChronicleMod(int no)
...
-	tmpChronicles.replace('\0', "");
+	QString tmpChronicles = chronicles.at(no);
 
 	QJsonObject mod
 	{
 		{ "modType", "Expansion" },
-		{ "name", QString::number(no) + " - " + QString(tmpChronicles) },
-		{ "description", tr("Heroes Chronicles") + " - " + QString::number(no) + " - " + QString(tmpChronicles) },
+		{ "name", QString::number(no) + " - " + tmpChronicles },
+		{ "description", tr("Heroes Chronicles") + " - " + QString::number(no) + " - " + tmpChronicles },
```
kambala-decapitator, line 142: Please use QString{}.arg()

launcher/modManager/chroniclesextractor.cpp
```diff
@@ -228,29 +212,46 @@ void ChroniclesExtractor::extractFiles(int no) const
...
 
+		logGlobal->info("Creating temporary directory");
 		if(!createTempDir())
 			continue;
 		
+		logGlobal->info("Copying offline installer");
+		// FIXME: this is required at the moment for Android (and possibly iOS)
+		// Incoming file names are in content URI form, e.g. content://media/internal/chronicles.exe
+		// Qt can handle those like it does regular files
+		// however, innoextract fails to open such files
```
kambala-decapitator, line 231: Iirc there's a way to get path on disk that's compatible with standard file read functions

launcher/modManager/chroniclesextractor.h
```diff
@@ -24,21 +24,22 @@ class ChroniclesExtractor : public QObject
...
-		{1, QByteArray{reinterpret_cast<const char*>(u"Warlords of the Wasteland"), 50}},
-		{2, QByteArray{reinterpret_cast<const char*>(u"Conquest of the Underworld"), 52}},
-		{3, QByteArray{reinterpret_cast<const char*>(u"Masters of the Elements"), 46}},
-		{4, QByteArray{reinterpret_cast<const char*>(u"Clash of the Dragons"), 40}},
-		{5, QByteArray{reinterpret_cast<const char*>(u"The World Tree"), 28}},
-		{6, QByteArray{reinterpret_cast<const char*>(u"The Fiery Moon"), 28}},
-		{7, QByteArray{reinterpret_cast<const char*>(u"Revolt of the Beastmasters"), 52}},
-		{8, QByteArray{reinterpret_cast<const char*>(u"The Sword of Frost"), 36}}
+	const QStringList chronicles = {
+		"", // fake	0th "chronicle", to create 1-based list
```
kambala-decapitator, line 34: {}

launcher/modManager/chroniclesextractor.cpp
IvanSavenko: Wanted to avoid localization changes... I guess I can also adjust .ts files manually.

launcher/modManager/chroniclesextractor.cpp
IvanSavenko: Should never happen.
We could try raising some error, in case of some unsupported installer or some other unknown case.

Need to check code though, this was based on some already existing code that also made no checks for emptyness.

launcher/modManager/chroniclesextractor.cpp
IvanSavenko: Possible, but haven't found it. Found one reference to Qt 6.X (6.6 I think?), but we're at Qt 5.

In any case this code was before, and it worked on all tested OS's. Added comment mostly to explain why this strange copy can't be simply removed.

launcher/modManager/chroniclesextractor.cpp
kambala-decapitator: It's also an incorrect way to write localized strings (didn't notice them first). The right way is tr("something %1 etc", "comment about parameters might also be useful").arg()

launcher/modManager/chroniclesextractor.cpp
kambala-decapitator: Oh, then it's probably possible only via jumping into android sdk

PR #5110, opened 2024-12-18
[1.6.1] AI optimization

IvanSavenko:
See commits list for more details.

Optimized various library methods used by AI. Actual AI code is mostly untouched (haven't seen anything problematic in profiler)

Tested on "Free for all" map (L, 7 AI), queen difficulty, running with autoskip for 2 weeks. AI run time went from ~240 seconds to ~150 seconds

Bonus system seems to be a little bit faster now, but still extremely slow - 15% of total AI turn time, second only to pathfinder (which also uses bonus system inside). Needs to be looked into, but not yet sure how to solve this problem.


DjWarmonger:
Looks good to me :)

Didn't investigate bonus caching though, never really understood how does it work.

IvanSavenko:
> Didn't investigate bonus caching though, never really understood how does it work.

Idea seems to be rather simple:
- for each node game caches entire list of bonuses that are affecting this node
- for each request to node (e.g. "all bonuses of type X") game caches list of bonuses filtered according to this request.
Nothing complex.

All of this is done via lazy evaluation - list of bonuses is computed only when a new request is done for outdated node, same for request caching.

But problem is that AI does HUGE number of requests and it does this in multiple threads. And this is where everything crashes.
1. Full list of bonuses needs to be regenerated on every bonus tree change. And must be done in threadrace-safe form.
2. Server operates on own set of bonus nodes, but uses shared tree change counter, leading to false-positives - when server changes counter, it invalidates entire cache on client
3. Since per-request cache is stored in node itself in form of map, every request is also query into map. Fast when done once, but slow when you do it thousands of times per ms.
4. Similarly, per-request cache must be updated in threadsafe form.

Can't think how to improve what we have though, other than completely redo it - I think we've reached the limits of what can be done with current approach...

IvanSavenko:
Regarding bonus system. Not a simple solution, but should work:

- instead of global tree update / invalidation counter, this should be done per node. Node is up to date if:
  - node itself is up-to-date
  - all its parents are up-to-date

- invalidation should be done not globally, but only on:
  - node itself
  - all its children
  - all parents, to which this node has bonuses it propagates to.

This should greatly reduce amount of times node gets invalidated - if hero equips an artifact, we only need to recompute bonuses on hero himself, his troops, and possibly - for player & visited town (if hero propagates any bonuses to them). And not for the entire world.

Nodes should probably keep their cache of local bonuses. Not sure what can be changed/improved here.

Per-request cache should be replaced with fixed list of cached requests, probably using enum + array. Less flexible than current solution that uses arbitrary strings as dictionary keys, but should be way faster than what we have now and won't require locking entire node. Besides - number of problematic requests is in single digits. 

===

Will think about it later, not planning for 1.6.1 / this PR.

IvanSavenko:
> If AI needs to access tree in multiple threads, will it be possible to have a lock per node to only lock part of a tree at a time? And potentially have parallel access?

That's how it works currently. Tree is effectively constant, so it is actually accessed (read-only) without any locks. Tree (or its content) can only change during netpack apply, and there is lock to prevent any access to gamestate when this happens.

In beta, entire node is locked for every access. But apparently even this level is too much for AI. For reference, I ran 7 AI's for 1 turn and it generated over several millions of accesses to bonus system (in total, for all nodes). With most popular queries hitting 500k calls.

In my PR, I instead use concurrent hash map from TBB which even allows concurrent access to values in hashmap. Most of accesses to bonus system now go through it, and bonus node itself is locked only in case of cache miss. And even this is not enough...

CMakeLists.txt
```diff
@@ -255,6 +255,10 @@ if(ENABLE_GOLDMASTER)
 	add_definitions(-DENABLE_GOLDMASTER)
 endif()
 
+# Qt defines 'emit' as macros, which conflicts with TBB definition of method with same name
+add_definitions(-DQT_NO_EMIT)
```
kambala-decapitator, line 259: it should be added to launcher's CML

kambala-decapitator: this workaround is also good:
```
#ifndef Q_MOC_RUN
#include <tbb/tbb.h>
#endif
```

IvanSavenko: Moved to map editor CMakeList.

Still workaround, but should be more reliable than correct include order which can easily break in future if we were to use something else from tbb.

Povelitelh3:
Tested, and now it's much better! Took map 'Supremacy of the Lich King' and turn was ~ 4-5 min on my PC

PR #5116, opened 2024-12-20
Polish translation for .ts files to reach 100%

MichalZr6:


mapeditor/translation/polish.ts
```diff
@@ -1096,10 +1096,10 @@
...
         <location filename="../inspector/inspector.cpp" line="360"/>
         <source>%n tile(s)</source>
-        <translation type="unfinished">
-            <numerusform></numerusform>
-            <numerusform></numerusform>
-            <numerusform></numerusform>
+        <translation>
+            <numerusform>%n heks</numerusform>
+            <numerusform>%n heksy</numerusform>
+            <numerusform>%n heksów</numerusform>
```
IvanSavenko, line 1102: Is this correct for Polish when referring to adventure map tiles? They are not hexagons, but squares / tiles.

MichalZr6: Heks might relate to a single, occupiable field in a wider context, however - as for adventure's map tiles - maybe "pole" is better. Does anyone have additional thoughts on that ?

dydzio0614: use "pole"

MichalZr6: done

PR #5118, opened 2024-12-20
Try to fix failure to copy file on Android

IvanSavenko:
Some people have been reporting that vcmi fails to copy gog installer on Android during Chronicles import.

And docs for QFile::copy mention that it does not works with content URI on Android. This PR replaces it with manual file copy.

Needs testing & confirmation that this approach works

### DIFF
--- a/launcher/modManager/chroniclesextractor.cpp
+++ b/launcher/modManager/chroniclesextractor.cpp
@@ -72,10 +72,15 @@ bool ChroniclesExtractor::extractGogInstaller(QString file)
 
 	if(!errorText.isEmpty())
 	{
+		logGlobal->error("Gog installer extractio failure! Reason: %s", errorText.toStdString());
+
 		QString hashError = Innoextract::getHashError(file, {}, {}, {});
 		QMessageBox::critical(parent, tr("Extracting error!"), errorText);
 		if(!hashError.isEmpty())
+		{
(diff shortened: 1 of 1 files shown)


Laserlicht:
> And docs for QFile::copy mention that it does not works with content URI on Android.

The interesting thing is that it works on most devices. :D

If this solution works: Maybe also add this method to Complete installation (we're also using `QFile::copy` there).
Would be great if it fixes the problems.

And I'm not sure if removing binary search a good idea as there are the first reports for using incorrect installers. :D

kambala-decapitator:
use java code to copy then, there's already a function that copies directories

IvanSavenko:
Closing. Does not helps, and https://github.com/vcmi/vcmi/pull/5198 offers hopefully better solution.

Laserlicht:
I think I've found one of the reasons why it doesn't help.

In last logfile in #5193 you can see that the file seems to be renamed inside content uri: `content://com.android.providers.downloads.documents/document/msf%3A1000089012`. And all our filename based approaches (e.g. file extension failed here).

IvanSavenko:
Ah, so in this case we don't even reach innoextract stage, since file does not have .exe extension. Great...

I guess we'll have to either use java to figure out original filename with extension, or open file and check for magic numbers in header.

PR #5128, opened 2024-12-21
seperate android icon for client

Laserlicht:
Another approach for fixing problem that it isn't possible to launch VCMI with gamepad on android. Also allows to skip launcher if not needed.

Almost same approach as in fheroes2.

grafik

@kambala-decapitator How can the client also linked to the main (launcher) app, before H3 data is installed?

related #4927


kambala-decapitator (commented):
sorry, I don't fully understand what this PR is doing. Will VCMI now have 2 icons instead of 1? Sounds like it will confuse players. Rather add a "context menu" option on the main icon to open client directly (I was lazy to do it for iOS).

> How can the client also linked to the main (launcher) app, before H3 data is installed?

not sure what for. Like, if no data, then open launcher instead of client? Sounds quite complicated for a niche fix, player should do it manually.

IvanSavenko:
> Rather add a "context menu" option on the main icon to open client directly 

This sounds like way better option to be honest. And looks like quite a lot of apps use it.

While I am not against the idea of having separate icon for direct start of client, but pushing it to everyone is bad idea IMO:
- this option is needed only for tiny number of players, so we should not force such confusing solution for all of them
- this would lead to return of long-gone questions on how to setup game, since client does not have proper logic for when H3 data is not found

Laserlicht:
Yes second icon. As you don't need launcher normally (it's also slow to load). Only if you want to change something.
Normally you just want to play -> Client is way faster. 

fheroes2 also do this, but reversed:
grafik

Laserlicht:
> Rather add a "context menu" option on the main icon to open client directly

Also a good idea! Will try

kambala-decapitator:
> As you don't need launcher normally (it's also slow to load)

like I said in #4927:

> Ideally, the dialog should be implemented in Java and would never even start Qt activity if not needed. But that might not be easy to implement.

or do what iOS is doing now: it has a system preference to choose how to start the game - via launcher or client directly.

although icon context menu is the best IMO

PR #5130, opened 2024-12-21
[1.6.1] add shortcut for android

Laserlicht:
Start shortcut for skipping launcher.

Shortcut:
grafik

Also possible to manually add to homescreen by user:
grafik

Also accessable via gamepad (hold A for opening shortcut menu, then arrow down and A again).

Sadly I had to add `shortcuts.xml` three times. That's a workaround as there are no variables allowed (we need different appids for release/debug/daily) in it. After searching for possible solutions it seems that this is the best solution here.

replacement for: #5128

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -248,7 +248,7 @@ jobs:
             cmake -DENABLE_CCACHE:BOOL=ON -DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14 --preset ${{ matrix.preset }}
         elif [[ (${{matrix.preset}} == android-conan-ninja-release) && (${{github.ref}} != 'refs/heads/master') ]]
         then
-            cmake -DENABLE_CCACHE:BOOL=ON -DANDROID_GRADLE_PROPERTIES="applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily" --preset ${{ matrix.preset }}
+            cmake -DENABLE_CCACHE:BOOL=ON -DANDROID_GRADLE_PROPERTIES="applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily;applicationVariant=daily" --preset ${{ matrix.preset }}
         elif [[ ${{startsWith(matrix.platform, 'msvc') }} ]]
         then
             cmake --preset ${{ matrix.preset }}
--- a/android/vcmi-app/src/main/res/values/strings.xml
+++ b/android/vcmi-app/src/main/res/values/strings.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="server_name">VCMI Server</string>
+    <string name="shortcut_play">Play VCMI</string>
 </resources>
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -317,7 +317,7 @@
             "description": "VCMI Android daily build",
             "inherits": "android-conan-ninja-release",
             "cacheVariables": {
-                "ANDROID_GRADLE_PROPERTIES": "applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily"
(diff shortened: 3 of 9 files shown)


kambala-decapitator:
you should be able to avoid copies of shortcuts.xml by starting activity through code:

```java
package eu.vcmi.vcmi;

import android.app.Activity;
import android.content.Intent;
import android.os.Bundle;

import androidx.annotation.*;

public class ShortcutActivity extends Activity
{
	@Override
	protected void onCreate(@Nullable Bundle savedInstanceState)
	{

		final var shortcutIntent = getIntent();
		final var mainIntent = VcmiSDLActivity.createIntent(shortcutIntent.getAction(), this);
		startActivity(mainIntent);
		finish();
	}
}

```

Laserlicht:
And how to start `.ShortcutActivity` from `shortcuts.xml`?

kambala-decapitator:
sorry, forgot to post shortcuts.xml. It's from another project, but should give you the idea:

```
<shortcuts xmlns:android="http://schemas.android.com/apk/res/android">
	<shortcut
		android:shortcutId="OpenRandomProblemID"
		android:enabled="true"
		android:icon="@drawable/ic_baseline_shuffle_24"
		android:shortcutShortLabel="@string/OpenRandomProblem"
		android:shortcutLongLabel="@string/OpenRandomProblemLong">
		<intent
			android:action="ACTION_OpenRandomProblem"
			android:targetPackage="ru.etudes.iSchet"
			android:targetClass="ru.etudes.iSchet.ShortcutActivity" />
	</shortcut>
</shortcuts>
```

kambala-decapitator:
hmmm, or does `targetPackage` have to match the app ID?

Laserlicht:
> hmmm, or does targetPackage have to match the app ID?

Yes. That's the problem. :D
Sorry if I was unclear here.

.github/workflows/github.yml
```diff
@@ -248,7 +248,7 @@ jobs:
             cmake -DENABLE_CCACHE:BOOL=ON -DCMAKE_C_COMPILER=gcc-14 -DCMAKE_CXX_COMPILER=g++-14 --preset ${{ matrix.preset }}
         elif [[ (${{matrix.preset}} == android-conan-ninja-release) && (${{github.ref}} != 'refs/heads/master') ]]
         then
-            cmake -DENABLE_CCACHE:BOOL=ON -DANDROID_GRADLE_PROPERTIES="applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily" --preset ${{ matrix.preset }}
+            cmake -DENABLE_CCACHE:BOOL=ON -DANDROID_GRADLE_PROPERTIES="applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily;applicationVariant=daily" --preset ${{ matrix.preset }}
```
kambala-decapitator, line 251: why is this variable copy-pasted here? it's already a part of preset, so it'll be taken from there

kambala-decapitator:
> The current approach is working. But not that nice because we need the copies for the different targetPackage.

then let's go with it and search for a better solution later

.github/workflows/github.yml
Laserlicht: Thanks

Laserlicht: Reverted. Removing breaks functionality.

kambala-decapitator: wat

ok, I'll check what's going on there

android/vcmi-app/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="server_name">VCMI Server</string>
+    <string name="shortcut_play">Play VCMI</string>
```
IvanSavenko, line 4: Can you add mention of this file in translation docs?

Laserlicht: Is already there: https://github.com/vcmi/vcmi/blob/d0ee1c89fbb1ace1f3d14ee616bdc89cb401d565/docs/translators/Translations.md?plain=1#L130

PR #5135, opened 2024-12-22
[1.7]Lua expressions in damage  formula

kdmcser:
make modder can define damage formula in gameConfig.json. 
the performance is 3~4 μs per call, so it will not make AI turns very long.


kdmcser:
unable to solve android dependencies

kambala-decapitator:
disable android temporarily on CI or set `fail-fast: false` that other platforms don't stop building, we'll merge without it and provide LuaJIT later. IIRC same issue will be with Apple and mingw.

kambala-decapitator:
you can also hide new feature under "feature flag" (a define basically) that other platforms don't fail

kdmcser:
use #ifdef in the code？

kambala-decapitator:
> use #ifdef in the code？

Yep

lib/battle/DamageCalculator.h
```diff
@@ -25,8 +26,10 @@ class DLL_LINKAGE DamageCalculator
 {
 	const CBattleInfoCallback & callback;
 	const BattleAttackInfo & info;
+    static std::unique_ptr<LuaExpressionEvaluator> attackSkillEvaluator;
+    static std::unique_ptr<LuaExpressionEvaluator> defenseSkillEvaluator;
```
IvanSavenko, line 30: Static lua states will not work when accessed from multiple threads. And BattleAI does uses multithreading. And creating lua state every time would be too costly...

Need to investigate what lua offers for concurrency.

kdmcser: how many threads will game use for damage calculator？

IvanSavenko: Why you're asking? 
As many as player has cores in CPU, possibly more.

IvanSavenko: Might actually be interesting idea - allocate N states, and then search for a free one. But probably overkill, at least for now.

So let's start with simple and straightforward solution - protect this class with mutex

kdmcser: in that solution, should I merge setVariable() and evaluate() into one function?

IvanSavenko: Yeah, makes sense

kdmcser: ok

IvanSavenko:
@kambala-decapitator 
What should we do with lua library in conan?

Should we wait for conan2 or might be better to rebuild existing conan1 package and add luajit to it?

IvanSavenko:
In that case - I'd rather keep this PR open until we can use lua on all platforms, to avoid mods that only work on some systems.

And if luajit can't be used on some platforms - consider how to handle those, like switching to pure lua as fallback

kdmcser:
If we cannot import the LuaJIT library, can we manually implement a simple Lua expression parser? I used DeepSeek to generate the parsing code:

`#include <string>
#include <memory>
#include <cmath>
#include <stdexcept>
#include <cctype>
#include <unordered_set>
#include <unordered_map>

using namespace std;

enum class TokenType {
    Number, Variable, Plus, Minus, Multiply, Divide, Modulo, Power, LeftParen, RightParen, End, Invalid
};

struct Token {
    TokenType type;
    string lexeme;
    double numberValue;

    Token(TokenType t, const string& l, double num) 
        : type(t), lexeme(l), numberValue(num) {}
};

class Lexer {
    string input;
    size_t position = 0;

    void skipWhitespace() {
        while (position < input.size() && isspace(input[position])) ++position;
    }

    Token parseNumber() {
        size_t start = position;
        bool hasDot = false;
        while (position < input.size()) {
            char c = input[position];
            if (isdigit(c)) ++position;
            else if (c == '.' && !hasDot) { hasDot = true; ++position; }
            else break;
        }
        string numStr = input.substr(start, position - start);
        try {
            return {TokenType::Number, numStr, stod(numStr)};
        } catch (...) {
            return {TokenType::Invalid, numStr, 0};
        }
    }

    Token parseIdentifier() {
        size_t start = position;
        while (position < input.size() && isalnum(input[position])) ++position;
        string id = input.substr(start, position - start);
        return {TokenType::Variable, id, 0}; // 接受所有合法标识符作为变量
    }

public:
    Lexer(const string& input) : input(input) {}

    Token getNextToken() {
        skipWhitespace();
        if (position >= input.size()) return {TokenType::End, "", 0};

        char c = input[position];
        if (isdigit(c) || c == '.') return parseNumber();
        if (isalpha(c)) return parseIdentifier();

        switch (c) {
            case '+': ++position; return {TokenType::Plus, "+", 0};
            case '-': ++position; return {TokenType::Minus, "-", 0};
            case '*': ++position; return {TokenType::Multiply, "*", 0};
            case '/': ++position; return {TokenType::Divide, "/", 0};
            case '%': ++position; return {TokenType::Modulo, "%", 0};
            case '^': ++position; return {TokenType::Power, "^", 0};
            case '(': ++position; return {TokenType::LeftParen, "(", 0};
            case ')': ++position; return {TokenType::RightParen, ")", 0};
            default: ++position; return {TokenType::Invalid, string(1, c), 0};
        }
    }
};

class ExprNode {
public:
    virtual ~ExprNode() = default;
    virtual double evaluate(const unordered_map<string, double>& vars) const = 0;
};

class BinaryOpNode : public ExprNode {
    char op;
    unique_ptr<ExprNode> left, right;

    static double lua_mod(double a, double b) {
        return a - floor(a / b) * b;
    }

public:
    BinaryOpNode(char op, unique_ptr<ExprNode> l, unique_ptr<ExprNode> r)
        : op(op), left(move(l)), right(move(r)) {}

    double evaluate(const unordered_map<string, double>& vars) const override {
        double lVal = left->evaluate(vars);
        double rVal = right->evaluate(vars);
        switch(op) {
            case '+': return lVal + rVal;
            case '-': return lVal - rVal;
            case '*': return lVal * rVal;
            case '/': 
                if (rVal == 0) throw runtime_error("Division by zero");
                return lVal / rVal;
            case '^': return pow(lVal, rVal);
            case '%': 
                if (rVal == 0) throw runtime_error("Modulo by zero");
                return lua_mod(lVal, rVal);
            default: throw runtime_error("Unknown operator");
        }
    }
};

class UnaryOpNode : public ExprNode {
    char op;
    unique_ptr<ExprNode> operand;

public:
    UnaryOpNode(char op, unique_ptr<ExprNode> node) 
        : op(op), operand(move(node)) {}

    double evaluate(const unordered_map<string, double>& vars) const override {
        double val = operand->evaluate(vars);
        return op == '-' ? -val : val;
    }
};

class VariableNode : public ExprNode {
    string name;

public:
    VariableNode(const string& name) : name(name) {}

    double evaluate(const unordered_map<string, double>& vars) const override {
        auto it = vars.find(name);
        if (it == vars.end()) {
            throw runtime_error("Undefined variable: " + name);
        }
        return it->second;
    }
};

class NumberNode : public ExprNode {
    double value;

public:
    NumberNode(double value) : value(value) {}

    double evaluate(const unordered_map<string, double>&) const override {
        return value;
    }
};

class Parser {
    Lexer& lexer;
    Token currentToken;
    unordered_set<string> allowedVariables;

    void eat(TokenType expected) {
        if (currentToken.type == expected) {
            currentToken = lexer.getNextToken();
        } else {
            throw runtime_error("Syntax error");
        }
    }

    unique_ptr<ExprNode> parsePrimary() {
        if (currentToken.type == TokenType::Number) {
            double val = currentToken.numberValue;
            eat(TokenType::Number);
            return make_unique<NumberNode>(val);
        }

        if (currentToken.type == TokenType::Variable) {
            string name = currentToken.lexeme;
            if (allowedVariables.find(name) == allowedVariables.end()) {
                throw runtime_error("Undefined variable: " + name);
            }
            eat(TokenType::Variable);
            return make_unique<VariableNode>(name);
        }

        if (currentToken.type == TokenType::LeftParen) {
            eat(TokenType::LeftParen);
            auto node = parseExpression();
            eat(TokenType::RightParen);
            return node;
        }

        if (currentToken.type == TokenType::Minus) {
            eat(TokenType::Minus);
            return make_unique<UnaryOpNode>('-', parsePrimary());
        }

        throw runtime_error("Unexpected token");
    }

    unique_ptr<ExprNode> parseFactor() {
        auto node = parsePrimary();
        while (currentToken.type == TokenType::Power) {
            eat(TokenType::Power);
            node = make_unique<BinaryOpNode>('^', move(node), parseFactor());
        }
        return node;
    }

    unique_ptr<ExprNode> parseTerm() {
        auto node = parseFactor();
        while (currentToken.type == TokenType::Multiply || 
               currentToken.type == TokenType::Divide ||
               currentToken.type == TokenType::Modulo) {
            char op = currentToken.lexeme[0];
            eat(currentToken.type);
            node = make_unique<BinaryOpNode>(op, move(node), parseFactor());
        }
        return node;
    }

    unique_ptr<ExprNode> parseExpression() {
        auto node = parseTerm();
        while (currentToken.type == TokenType::Plus || 
               currentToken.type == TokenType::Minus) {
            char op = currentToken.lexeme[0];
            eat(currentToken.type);
            node = make_unique<BinaryOpNode>(op, move(node), parseTerm());
        }
        return node;
    }

public:
    Parser(Lexer& lexer, const unordered_set<string>& vars) 
        : lexer(lexer), allowedVariables(vars) {
        currentToken = lexer.getNextToken();
    }

    unique_ptr<ExprNode> parse() {
        auto expr = parseExpression();
        if (currentToken.type != TokenType::End) {
            throw runtime_error("Unexpected end of expression");
        }
        return expr;
    }
};

class LuaExpressionParser {
    unique_ptr<ExprNode> ast;
    unordered_set<string> allowedVariables;

    string preprocess(const string& expr) {
        const string prefix = "return ";
        if (expr.substr(0, prefix.size()) == prefix) {
            return expr.substr(prefix.size());
        }
        return expr;
    }

public:
    LuaExpressionParser(const string& expr, const vector<string>& variables) {
        for (const auto& var : variables) {
            allowedVariables.insert(var);
        }
        string processed = preprocess(expr);
        Lexer lexer(processed);
        Parser parser(lexer, allowedVariables);
        ast = parser.parse();
    }

    double evaluate(const unordered_map<string, double>& variables) {
        return ast->evaluate(variables);
    }
};
`

kambala-decapitator:
we can import it, just wait a bit please

Issue #5056, opened 2024-12-09
Bonus immuneToFire don't display the info in Creature info panel


lishaoxia1985:
@Laserlicht 
![YuneR.png](https://github.com/user-attachments/assets/37685e4e-9f87-4bb2-8a4f-a3e6bedfdd78)

![YueR.png](https://github.com/user-attachments/assets/fa7bc4b4-acf0-4ef8-9bc5-6ffe2b038036)

![XueR.png](https://github.com/user-attachments/assets/0712563f-60f6-4754-b649-a54d7daf42bc)

lishaoxia1985:
@Laserlicht It seems that we should add these code below in json:
```
"core.bonus.SPELL_SCHOOL_IMMUNITY.name": "Immune to ${subtype.spellSchool}",    "core.bonus.SPELL_SCHOOL_IMMUNITY.description": "Immune to ${subtype.spellSchool}",
```

Issue #4682, opened 2024-09-28
Bonuses display bug


kdmcser:
I only install a mod: vcmi-extra, then this bug appears. In both 1.6 and 1.5.7
![屏幕截图 2024-09-28 124809](https://github.com/user-attachments/assets/6aadd974-12f7-44f5-bee0-f594a63b50ac)

**Non-living**
the vcmi-extra mod modify bonus NON-LIVING icon to zvs/Lib1.res/E_MORAL, but it is still using zvs/Lib1.res/NonLiving

```
"core:NON_LIVING":
    {
        "graphics":
        {
            "icon":  "zvs/Lib1.res/E_MORAL"
        }
    }
```
I find some log in console:
![屏幕截图 2024-09-28 125112](https://github.com/user-attachments/assets/f27b266d-da35-4fa3-a2a1-50813f552a05)

**Spell School Immunity**

this bonus is not in config/bounses.json, but `bonusToGraphics` still gives icon for them

![屏幕截图 2024-09-28 125455](https://github.com/user-attachments/assets/4d50184c-5a2c-4cc9-a65b-6a73c7d89085)

but `bonusToString` method seem not deal with it，bt.hidden is 1 and return empty string.

![屏幕截图 2024-09-28 125522](https://github.com/user-attachments/assets/c74b016d-5719-44ee-80bd-482615ce86ba)

I dont know which is the best way to solve it， so make a report.

Issue #3543, opened 2024-01-22
Missing description for spell school immunity


by003:
**Describe the bug**
Missing description for all spell school immunities. 

It is still linked to FIRE_IMMUNITY
![image](https://github.com/vcmi/vcmi/assets/127420617/d8f935dd-3cd1-43a1-a8df-949de592bd6e)

But now this bonus name is:
![image](https://github.com/vcmi/vcmi/assets/127420617/a584de35-f7e3-4228-b7af-81a36629bf81)
 

![image](https://github.com/vcmi/vcmi/assets/127420617/2058d097-8e31-43a3-9ef1-e25d28b78fc7)

**Version**
 - OS: Windows
 - Version: 1.44

PR #5137, opened 2024-12-22
[1.6.1] Bonus display

Laserlicht:
fixes
- Spell immunity #3543 #5056 #4682
- OPENING_BATTLE_SPELL (reported on discord)

### DIFF
--- a/lib/CBonusTypeHandler.cpp
+++ b/lib/CBonusTypeHandler.cpp
@@ -75,6 +75,23 @@ std::string CBonusTypeHandler::bonusToString(const std::shared_ptr<Bonus> & bonu
 	std::string textID = description ? bt.getDescriptionTextID() : bt.getNameTextID();
 	std::string text = VLC->generaltexth->translate(textID);
 
+	if (bonus->subtype.as<SpellSchool>().hasValue())
+	{
+		auto school = bonus->subtype.as<SpellSchool>();
+		if(school != SpellSchool::ANY)
+		{
+			auto specificTextID = description ? bt.getDescriptionTextID() + ".specific" : bt.getNameTextID() + ".specific";
+			auto specificText = VLC->generaltexth->translate(specificTextID);
+
+			if(specificText.find("${subtype.spellSchool}") != std::string::npos)
+			{
+				auto schoolName = VLC->generaltexth->translate(TextIdentifier("core.spellSchools", school).get());
+				boost::algorithm::replace_all(specificText, "${subtype.spellSchool}", schoolName);
+				text = specificText;
+			}
+		}
+	}
+
 	if (text.find("${val}") != std::string::npos)
 		boost::algorithm::replace_all(text, "${val}", std::to_string(bearer->valOfBonuses(Selector::typeSubtype(bonus->type, bonus->subtype))));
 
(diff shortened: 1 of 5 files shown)


lib/CBonusTypeHandler.cpp
```diff
@@ -84,6 +84,13 @@ std::string CBonusTypeHandler::bonusToString(const std::shared_ptr<Bonus> & bonu
 	if (text.find("${subtype.spell}") != std::string::npos && bonus->subtype.as<SpellID>().hasValue())
 		boost::algorithm::replace_all(text, "${subtype.spell}", bonus->subtype.as<SpellID>().toSpell()->getNameTranslated());
 
+	if (text.find("${subtype.spellSchool}") != std::string::npos && bonus->subtype.as<SpellSchool>().hasValue())
+	{
+		auto school = bonus->subtype.as<SpellSchool>();
+		auto schoolName = VLC->generaltexth->zelp[school == SpellSchool::ANY ? 458 : 454 + school].first;
+		boost::algorithm::replace_all(text, "${subtype.spellSchool}", schoolName);
+	}
+
```
IvanSavenko, line 93: This won't work in Ukrainian and probably rest of Slavic languages due to more complex grammatical cases:
Water Spells: Магія Води / Закляття Води
Immunity to Water spells: Іммунітет до магії Води / Іммунітет до заклять Води
"Magiya" became "Magiyi", "Zaklyattya" became "Zaklyat`"

And even in English this would result in somewhat awkward "Immunity to Water Spells" - why "Spells" are capitalized?

Proposition: add custom descriptions (not sure about names) for all bonuses that have spell school as subtype.

For this you would need:
- In `CBonusTypeHandler::loadItem`, for these bonuses instead of registering description string, you will need to register 4 additional string for all 4 spell schools
- In `CBonusTypeHandler::bonusToString`, if spell has spell school that is not "all" as subtype, use specific translation instead of generic one.

And please check also bonuses `SPELL_DAMAGE_REDUCTION` and `NEGATIVE_EFFECTS_IMMUNITY`. I think they also may need per-school translations

Laserlicht: Added `.specific`-Identifier.

Should also work with `SPELL_DAMAGE_REDUCTION` and `NEGATIVE_EFFECTS_IMMUNITY`. If translations will be added (not part of this PR).

GeorgeK1ng:
I can't properly translate these newly added strings. ${subtype.spellSchool} is supposed to be in core.bonus.SPELL_SCHOOL_IMMUNITY.description.specific + core.bonus.SPELL_SCHOOL_IMMUNITY.name.specific. 
And in Czech is needed in both sentences different wording. For example to Immunity to Fire magic.

"core.bonus.SPELL_SCHOOL_IMMUNITY.name.specific": "Imunní vůči ohn**i**",
"core.bonus.SPELL_SCHOOL_IMMUNITY.description.specific": "Imunní vůči magii ohn**ě**",

Laserlicht:
Wasn't aware of that. But in worst case: Use the same name as description.
I've no good solution other for that for this in mind.

IvanSavenko:
Let me take a look

PR #5139, opened 2024-12-22
Fix crash on mapobjects with description popup window

MichalZr6:
This PR fixes crashes for non-core mapobjects with descriptions when popup window is being indicated.

I started to investigate when Junkman popup window crashed recently the game for me (and also it was reported on Discord today).

Non-core mapobjects don't use `json["lastReservedIndex"]` so after 
[newObject->objectTypeHandlers.resize(json["lastReservedIndex"].Float() + 1);](https://github.com/vcmi/vcmi/blob/367e1f192599e9788e60be5ce8af0095d8682541/lib/mapObjectConstructors/CObjectClassesHandler.cpp#L292) and then [push_back](https://github.com/vcmi/vcmi/blob/367e1f192599e9788e60be5ce8af0095d8682541/lib/mapObjectConstructors/CObjectClassesHandler.cpp#L176) there is valid handler at second index whereas the first one is empty.
After that, when popup window is called, code tries to get handler at [index = 0](https://github.com/vcmi/vcmi/blob/367e1f192599e9788e60be5ce8af0095d8682541/lib/texts/MetaString.cpp#L398) and this function [CObjectClassesHandler::getHandlerFor](https://github.com/vcmi/vcmi/blob/367e1f192599e9788e60be5ce8af0095d8682541/lib/mapObjectConstructors/CObjectClassesHandler.cpp#L357) crashes.

@IvanSavenko maybe you can figure out where this should be fixed during the loading stage or perhaps we leave it as it is.

That way or another: the fix is neutral in its core.

The similar crash was with miniHillFort popup window a while ago. I reverted the commit I made back then. It generated unnecessary log message and additional logic for MetaString is unneeded.

### DIFF
--- a/lib/mapObjects/CGMarket.cpp
+++ b/lib/mapObjects/CGMarket.cpp
@@ -45,7 +45,7 @@ std::string CGMarket::getPopupText(PlayerColor player) const
 		return getHoverText(player);
 
 	MetaString message = MetaString::createFromRawString("{%s}\r\n\r\n%s");
-	message.replaceName(ID);
+	message.replaceTextID(TextIdentifier("object", getObjectHandler()->getModScope(), getObjectHandler()->getTypeName(), "name").get());
 	message.replaceTextID(TextIdentifier(getObjectHandler()->getBaseTextID(), "description").get());
 	return message.toString();
 }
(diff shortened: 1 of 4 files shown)


IvanSavenko:
Actual reason was hardcoded 0 in this method:
```cpp
void MetaString::replaceName(const MapObjectID & id)
{
	replaceTextID(VLC->objtypeh->getObjectName(id, 0));
}
```
Instead it should have used object subID. See #5143

IvanSavenko:
Closing, superseded by https://github.com/vcmi/vcmi/pull/5143

PR #5141, opened 2024-12-23
Fixes excessive logging causing RMG slowdown

IvanSavenko:
- Do not apply formatting on log entries that will be discarded due to high log level
- Do not log incorrect access to visitablePos due to its numerous use by RMG

### DIFF
--- a/lib/mapObjects/CGObjectInstance.cpp
+++ b/lib/mapObjects/CGObjectInstance.cpp
@@ -217,8 +217,8 @@ int CGObjectInstance::getSightRadius() const
 
 int3 CGObjectInstance::getVisitableOffset() const
 {
-	if (!isVisitable())
-		logGlobal->debug("Attempt to access visitable offset on a non-visitable object!");
+//	if (!isVisitable())
+//		logGlobal->debug("Attempt to access visitable offset on a non-visitable object!");
 	return appearance->getVisitableOffset();
 }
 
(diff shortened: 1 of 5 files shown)


DjWarmonger:
Oh man:
`Generated random map in 7257 ms.`

That's over 4x speedup on my gear o.O

lib/mapObjects/CGObjectInstance.cpp
DjWarmonger, line 1: These don't seem to have any impact on RMG performance. RMG shouldn't even check if ObjectInstance is visitable or not, as this is known in advance.

`Generated random map in 6744 ms.`

DjWarmonger (changes requested):

lib/mapObjects/CGObjectInstance.cpp
IvanSavenko: Are you sure?
```
1 CGObjectInstance::visitablePos              CGObjectInstance.cpp 321 0x7ffff7464aa5 
2 rmg::Object::Instance::getVisitablePosition RmgObject.cpp        67  0x7ffff7689bb3 
3 rmg::Object::getVisitablePosition           RmgObject.cpp        267 0x7ffff768b71a 
4 ObjectManager::placeObject                  ObjectManager.cpp    590 0x7ffff7749f3e 
5 ObstaclePlacer::placeObject                 ObstaclePlacer.cpp   142 0x7ffff7796531 
6 ObstacleProxy::createObstacles              ObstacleProxy.cpp    337 0x7ffff75b9854 
7 ObstaclePlacer::process                     ObstaclePlacer.cpp   115 0x7ffff779611d 
8 Modificator::run                            Modificator.cpp      87  0x7ffff7743d1b 
```
This happens for Obj::REEF, Obj::MOUNTAIN Obj::CRATER, and multiple others

I have no active mods, so this is not issue with mods.

IvanSavenko: Effect on RMG performance from this logging is rather low, 5-10%, at most, so not very noticeable. Unlike that formatter change.

DjWarmonger:
Yep, they were called only for each static object actually placed, and that part was quick: https://github.com/vcmi/vcmi/pull/5149

PR #5146, opened 2024-12-24
[1.7] Added Inno Setup Installer files

GeorgeK1ng:
Added all needed files for future building new custom installer using Inno Setup


kambala-decapitator:
not sure why you converted Windows-only files to LF - CRLF is ok for them

CI/wininstaller/installer.iss
```diff
@@ -0,0 +1,733 @@
...
+Name: "turkish"; MessagesFile: "{#LangPath}\Turkish.isl"
+Name: "ukrainian"; MessagesFile: "{#LangPath}\Ukrainian.isl"
+Name: "vietnamese"; MessagesFile: "{#LangPath}\Vietnamese.isl"
+
+[Files]
+Source: "{#SourceFilesPath}\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsubdirs; BeforeInstall: PerformHeroes3FileCopy
+
+[Icons]
+Name: "{group}\{cm:ShortcutLauncher}"; Filename: "{app}\VCMI_launcher.exe"; Comment: "{cm:ShortcutLauncherComment}"
+Name: "{group}\{cm:ShortcutMapEditor}"; Filename: "{app}\VCMI_mapeditor.exe"; Comment: "{cm:ShortcutMapEditorComment}"
```
kambala-decapitator, line 103: maybe also add Client, but only to start menu? The editor could also go only there, most users don't need to have it on desktop.

CI/wininstaller/build_installer.cmd
```diff
@@ -0,0 +1,73 @@
...
+set "LangPath=%BaseDir%CI\wininstaller\lang"
+set "LicenseFile=%BaseDir%license.txt"
+set "IconFile=%BaseDir%clientapp\icons\vcmi.ico"
+set "SmallLogo=%BaseDir%CI\wininstaller\vcmismalllogo.bmp"
+set "WizardLogo=%BaseDir%CI\wininstaller\vcmilogo.bmp"
+set "InstallerScript=%BaseDir%CI\wininstaller\installer.iss"
+
+REM Determine Program Files directory based on system architecture
+if exist "%WinDir%\SysWow64" (
+    set "ProgFiles=%programfiles(x86)%"
```
kambala-decapitator, line 34: is inno setup always 32-bit?

CI/wininstaller/installer.iss
```diff
@@ -0,0 +1,733 @@
...
+
+
+function GetDefaultDir(Default: String): String;
+begin
+  if IsAdmin then
+    // Default to Program Files for admins
+    Result := GetCommonProgramFilesDir + '\VCMI'
+  else
+    // Default to User AppData for non-admin users
+    Result := GlobalUserAppdataFolder + '\VCMI';
```
kambala-decapitator, line 312: would `C:\Games\` be a better choice perhaps?

CI/wininstaller/installer.iss
```diff
@@ -0,0 +1,733 @@
...
+  OriginalUserName := '\' + GetUserNameString + '\';
+
+  // Expand the {userdesktop} constant
+  UserDesktopPath := ExpandConstant('{userdesktop}');
+
+  // Replace the original username with the current session username
+  StringChangeEx(UserDesktopPath, OriginalUserName, CurrentSessionUserName, True);
+
+  // Return the modified user desktop folder path
+  Result := UserDesktopPath;
```
kambala-decapitator, line 381: this code is used 3 times and differs only by the constant used :)

CI/wininstaller/installer.iss
```diff
@@ -0,0 +1,733 @@
...
+      // Disable the first option ("Install for All Users") for non-admins
+      InstallModePage.CheckListBox.ItemEnabled[0] := False;
+
+      // Force a redraw of the CheckListBox to fix appearance
+      InstallModePage.CheckListBox.Invalidate();
+    end;
+  end;
+ 
+  // Check for Heroes 3 installation paths
+  Heroes3Path := '';
```
kambala-decapitator, line 448: this line can be removed

CI/wininstaller/installer.iss
```diff
@@ -0,0 +1,733 @@
...
+  if Heroes3Path = '' then
+    Heroes3Path := RegistryQueryPath('SOFTWARE\WOW6432Node\GOG.com\Games\1207658787', 'path');
+  if Heroes3Path = '' then
+    Heroes3Path := RegistryQueryPath('SOFTWARE\New World Computing\Heroes of Might and Magic® III\1.0', 'AppPath');
+  if Heroes3Path = '' then
+    Heroes3Path := RegistryQueryPath('SOFTWARE\WOW6432Node\New World Computing\Heroes of Might and Magic® III\1.0', 'AppPath');
+  if Heroes3Path = '' then
+    Heroes3Path := RegistryQueryPath('SOFTWARE\New World Computing\Heroes of Might and Magic III\1.0', 'AppPath');
+  if Heroes3Path = '' then
+    Heroes3Path := RegistryQueryPath('SOFTWARE\WOW6432Node\New World Computing\Heroes of Might and Magic III\1.0', 'AppPath');
```
kambala-decapitator, line 459: in what form can inno consume constants from external files? C++ code contains the same paths in the launcher, I'd like to share them instead of copy-pasting.

CI/wininstaller/installer.iss
```diff
@@ -0,0 +1,733 @@
...
+      InstallModePage.CheckListBox.ItemEnabled[0] := False;
+
+      // Force a redraw of the CheckListBox to fix appearance
+      InstallModePage.CheckListBox.Invalidate();
+    end;
+  end;
+ 
+  // Check for Heroes 3 installation paths
+  Heroes3Path := '';
+  Heroes3Path := RegistryQueryPath('SOFTWARE\GOG.com\Games\1207658787', 'path');
```
kambala-decapitator, line 449: does this variable become global? I see it's used in another function.

CI/wininstaller/installer.iss
```diff
@@ -0,0 +1,733 @@
...
+      WizardForm.DirEdit.Text := GetCommonProgramFilesDir + '\VCMI'
+    else
+      WizardForm.DirEdit.Text := GlobalUserAppdataFolder + '\VCMI';
+  end;
+
+  Result := True;
+end;
+
+
+procedure PerformHeroes3FileCopy();
```
kambala-decapitator, line 533: we already have this logic in the launcher, not sure it should be copied to installer

CI/wininstaller/installer.iss
GeorgeK1ng: I did, but removed it as client runs only with console window. That could be confusing for normal users.

CI/wininstaller/build_installer.cmd
GeorgeK1ng: For now seems it's only 32-bit

CI/wininstaller/installer.iss
kambala-decapitator: normal users will just run launcher from desktop :)

CI/wininstaller/installer.iss
GeorgeK1ng: Will change it to this:

```
function GetUserFolderPath(Constant: String): String;
var
  FolderPath: String;
  OriginalUserName: String;
  CurrentSessionUserName: String;
begin
  // Retrieve the current username from the session
  CurrentSessionUserName := '\' + GlobalUserName + '\';

  // Retrieve the original username
  OriginalUserName := '\' + GetUserNameString + '\';

  // Expand the specified constant
  FolderPath := ExpandConstant(Constant);

  // Replace the original username with the current session username in the path
  StringChangeEx(FolderPath, OriginalUserName, CurrentSessionUserName, True);

  // Return the modified folder path
  Result := FolderPath;
end;

// Specific functions for user folders
function GetUserDocsFolder: String;
begin
  Result := GetUserFolderPath('{userdocs}');
end;

function GetUserAppdataFolder: String;
begin
  Result := GetUserFolderPath('{userappdata}');
end;

function GetUserDesktopFolder(Default: String): String;
begin
  Result := GetUserFolderPath('{userdesktop}');
end;
```

CI/wininstaller/installer.iss
GeorgeK1ng: Yes, it becomes global

CI/wininstaller/installer.iss
GeorgeK1ng: Inno Setup uses RegQueryStringValue in it's pascal scripting.

CI/wininstaller/installer.iss
GeorgeK1ng: I have added it to simplify setup process. Not sure why this behavior is 2-3 times quicker on testing machine instead of copying using launcher.

CI/wininstaller/installer.iss
GeorgeK1ng: Removed

CI/wininstaller/installer.iss
GeorgeK1ng: User can always check it. Also not sure if in non-admin environment will be this allowed without addition system checks or prompts. Also installing to system drive root becomes not only user profile installation. Normally will users install in same way to Program Files as usual

GeorgeK1ng:
> not sure why you converted Windows-only files to LF - CRLF is ok for them

I just did it because GitHub check failed on that. So it now match standard with other files in repository.

CI/wininstaller/installer.iss
GeorgeK1ng: And don't think I can link external function from non installed app. For example if we have this function in launcher I can use it from installer. Probably just system apis like I did for logged user name detection

CI/wininstaller/installer.iss
kambala-decapitator: @IvanSavenko @Laserlicht what do you think?

CI/wininstaller/installer.iss
kambala-decapitator: is it possible to "pre-extract" a file (say, to `%TEMP%`) and then reference it inside installer? E.g. a json or just something csv-like, then parse it to get the list of paths and keys.

CI/wininstaller/installer.iss
GeorgeK1ng: I have added it according to Discrord discussions few days back. Also it matches logic uninstall wizard.
And is useful only for players with already installed H3, if not installed user still must do this in launcher.

kambala-decapitator:
> > not sure why you converted Windows-only files to LF - CRLF is ok for them
> 
> I just did it because GitHub check failed on that. So it now match standard with other files in repository.

that's probably our custom script, it should be configured with exclusions

kambala-decapitator:
please drop the CRLF->LF commit and add `-and -not -name '*.cmd' -and -not -name '*.iss' -and -not -name '*.isl'` before `-print0` in https://github.com/vcmi/vcmi/blob/develop/.github/workflows/github.yml#L398

CI/wininstaller/installer.iss
GeorgeK1ng: Question is why? Launcher have it hardcoded aswell.

kambala-decapitator: that's exactly the point: to get rid of hardcoding them in 2 places

CI/wininstaller/installer.iss
GeorgeK1ng: Yes, and then non power users who run search in start menu / taskbar will be confused with just another icon which is likely for modders only.

CI/wininstaller/installer.iss
GeorgeK1ng: These strings will probably never changes. And even with another json, it means to distribute useless file with VCMI. Same as I wanted with GOG installers informations instead of hardcoding. And even with json, its likely impossible to deal it like this in Installer.

kambala-decapitator: ok, let's not touch it then

GeorgeK1ng:
> please drop the CRLF->LF commit and add `-and -not -name '*.cmd' -and -not -name '*.iss' -and -not -name '*.isl'` before `-print0` in https://github.com/vcmi/vcmi/blob/develop/.github/workflows/github.yml#L398

Ok, will do it soon.

CI/wininstaller/installer.iss
kambala-decapitator: not insisting on that

CI/wininstaller/installer.iss
IvanSavenko: Not against it I guess? 
Not sure how it is done currently, but personally I would have added it as selected by default checkbox, alongside with other options - if only to indicate what exactly installer is doing.

IvanSavenko:
Not sure if we need it in beta / ASAP, could go to develop / 1.7, unless there is strong reason othervice?

Will take a look later, after 1.6.2 probably.

CI/wininstaller/installer.iss
GeorgeK1ng: It's invisible feature. If installed path is present in registry and files not already exist in My games\VCMI, then they are automatically copied to proper place. It's basic check if folders Data, Maps, Mp3 folder is bigger than 1 MB. That copy process is done during "Extracting files" - at the begining of installation. It's too quick in my tests so I can't screenshot it.
But it's done on this screen at the beginning.
image

I have tested this info screen notification about file copy.
image


Adding another checkbox section when H3 installation is possible. But it will require installer resizing, bacause it becomes invisible and user need to scroll.
image

image

Also don't know what could be reason to uncheck that checkbox. 

For me looks more clear just to inform user that it will be done automatically in "Ready to Install" page

IvanSavenko: What about reducing number of groups?
System integration:
- Associate h3m
- Associate vmap 
- Created desktop shortcuts
- Create start menu shortcuts
VCMI configuration
- Add firewall rules
- Copy H3 data

GeorgeK1ng:
I see various reasons to do it. This commit will just add required files to repository, so it can be set later.

## NSIS
- Requires admin rights, can't be installed without them.
- Always installs into `Program Files` instead of `Program Files (x86)` on x64 systems when installing x86 build.
- Always allows x64 setup on x86 systems, which makes the installed program unusable.

Public v1.6 x64 screenshots
image
image
image
image
image
image
image
image
image


## Inno Setup
- Supports both elevated and non-elevated installations.
- Offers custom options to select file associations.
- Offers custom options to create shortcuts.
- Automatically installs into the correct `Program Files` folder based on the included binaries.
- Supports automatic detection and copying of installed Heroes III files to the required folder during setup.
- Includes an uninstall wizard with an option to remove user files from "My Games\VCMI."
- Supports silent installation with language selection.
- Blocks x64 setup on x86 systems.
- Fully translatable and customizable text.
- Features a fresh and modern look.

Testing installer screenshots
image
image
image
image
image
image
image
image
image
image
image
image
image
image

CI/wininstaller/installer.iss
GeorgeK1ng: Great idea! Checkbox will appear only if H3 installation is detected
image

I have also added warning to unchecking firewall rule
image

Will redo all languages strings and add all modded files to PR

.github/workflows/github.yml
```diff
@@ -395,7 +395,7 @@ jobs:
           run: |
             find . -path ./.git -prune -o -path ./AI/FuzzyLite -prune -o -path ./test/googletest \
             -o -path ./osx  -prune -o -type f \
-            -not -name '*.png' -and -not -name '*.ttf' -and -not -name '*.wav' -and -not -name '*.webm' -and -not -name '*.ico' -and -not -name '*.bat' -print0 | \
+            -not -name '*.png' -and -not -name '*.ttf' -and -not -name '*.wav' -and -not -name '*.webm' -and -not -name '*.ico' -and -not -name '*.bat' -and -not -name '*.cmd' -and -not -name '*.iss' -and -not -name '*.isl -print0 | \
```
kambala-decapitator, line 398: you have also not dropped the CRLF conversion commit yet

```suggestion
            -not -name '*.png' -and -not -name '*.ttf' -and -not -name '*.wav' -and -not -name '*.webm' -and -not -name '*.ico' -and -not -name '*.bat' -and -not -name '*.cmd' -and -not -name '*.iss' -and -not -name '*.isl' -print0 | \
```

GeorgeK1ng: Thanks for fixing that typo. New commit in few minutes

GeorgeK1ng:
I think there is just one question now. As VCMI require Microsoft Visual C++ 2022 installed, installer should probably do this check and have ability to download and install them during setup when launched elevated setup. But not sure if this is really needed to be as part of this PR.

IvanSavenko:
Should not be needed. We package msvc redist as part of our installer.

GeorgeK1ng:
I don't think they are part of current installer. Installed clean **en_windows_7_ultimate_with_sp1_x64_dvd_u_677332.iso** or **en_windows_7_ultimate_with_sp1_x86_dvd_u_677460.iso** + GOG H3 + VCMI v1.6 and VCMI can't be launched. Because od missing imports from not present libraries. For example.
image

IvanSavenko:
In that case either we're using it wrong, or CMake does not do what we need:
https://github.com/vcmi/vcmi/blob/f9e9b275a7e46d7714012aa01bdf006f8d5d4180/CMakeLists.txt#L782

Laserlicht:
> I don't think they are part of current installer.

Also remember this thing on my testing VM (also on older versions before 1.6).

Other points:
- Maybe also add hint for translation to translations doc?
- Does innosetup also support png instead of bmp (bmp isn't well supported anymore - e.g. Github cannot preview them)

GeorgeK1ng:
Seems like it's compiled as dynamic instead of static build.

GeorgeK1ng:
> > I don't think they are part of current installer.
> 
> Also remember this thing on my testing VM (also on older versions before 1.6).
> 
> Other points:
> 
> * Maybe also add hint for translation to translations doc?
> * Does innosetup also support png instead of bmp (bmp isn't well supported anymore - e.g. Github cannot preview them)

Will add translation hint to Translation page. Currently is that hint only in ISS script itself
Only BMPs are supported - https://jrsoftware.org/ishelp/index.php?topic=setup_wizardsmallimagefile

kambala-decapitator:
> Seems like it's compiled as dynamic instead of static build.

Dynamic sounds correct.

Anyway, standard approach is to include vcredist installer instead of having those dlls inside the program directory.

GeorgeK1ng:
> In that case either we're using it wrong, or CMake does not do what we need:
> 
> https://github.com/vcmi/vcmi/blob/f9e9b275a7e46d7714012aa01bdf006f8d5d4180/CMakeLists.txt#L782

I have just checked generated Visual Studio 2019 project and it's dynamic.
image

It should be static to not require additional ms-api* files
image

IvanSavenko:
I believe all our libraries are compiled with dynamic runtime, so we can't simply change that - will need full rebuild of dependencies package.

So:
option 1 - find a way for CMake to include dynamic runtime libraries
option 2 - replace those libraries with vc_redist installation (which should be taken from system on which vcmi was built)

GeorgeK1ng:
> > Seems like it's compiled as dynamic instead of static build.
> 
> Dynamic sounds correct.
> 
> Anyway, standard approach is to include vcredist installer instead of having those dlls inside the program directory.

Dynamic creates additional depencies to ms-api-* files. Static build must be used for independent build to not require these additional libraries. This is how I did my ListDisk or ResolutionChanger possible to run in WinPE 6.1+ without any additional libraries. Should be same here to get rid of needs of installed VC++ 2022.

kambala-decapitator:
> to get rid of needs of installed VC++ 2022

personally I don't consider it right (every exe will grow in size). Besides, like Ivan said, dependencies were also built with dynamic, and it makes zero sense to build them with static runtime.

GeorgeK1ng:
Sadly these missing libraries are OS dependent. They are all part of KB2999226 for 6.0 (Vista), 6.1 (7), 6.2 (8.0)

IvanSavenko:
> Sadly these missing libraries are OS dependent. They are all part of KB2999226 for 6.0 (Vista), 6.1 (7), 6.2 (8.0)

redist ones? No, I don't think that they are system-dependent. 

However they are COMPILER-dependent. Microsoft can (and does) silently update redist that is shipped with Visual Studio. Such update can even happen in minor msvc update. In order for app to run you need to:
a) Install redist that was shipped with same version of Visual Studio (or newer). 
b) Install latest redist, available somewhere on MS website.

GeorgeK1ng:
Just KB2999226 is needed to be installed, tested it right now on Windows 7 x86 and x64 untouched ISO.
GOG H3 + VCMI 1.6 =  error, missing libraries. Installed KB2999226, VCMI runs.

This update is part of standard Windows Update, so it should be fine for now.

GeorgeK1ng:
Maybe I can add something like this. If every check will be passed, then this page won't be visible. 

image

GeorgeK1ng:
Still thinking how to provide fully working independent package. Whart about including needed updates in installer?
image

x64 installer size will grown with +-2 MB on size that's not too much for including update file for Windows 7, 8.0 and 8.1.
I can then add simple check if needed files are present in system32 / syswow64 and if not, just extract needed files from update CAB into VCMI installation directory instead. 

I have tested it now on non-admin account on Windows 7 x86 and it worked fine.

IvanSavenko:
How are these updates different from vc_redist from here? Because this is what we actually need.
https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170

Considering that your update was released in 2015 (!), way before VS 2022 it is very likely that those dll's are way older than those that are provided in redist and are very likely to cause strange and obscure bugs.

GeorgeK1ng:
No you don't need that redist. Current build mechanims already have it's own msvcp140.dll. Part of redist it this update - UCRT libraries. Redist itself do not contains any updated ms-api*. Compiled mechanism is set to UCRT, which is delivered to pre Windows 10 system only as KB2999226.

GeorgeK1ng:
I have just installed Visual Studio 2022 on testing machine and found updated UCRT for redist are there: C:\Program Files (x86)\Windows Kits\10\Redist\10.0.22621.0\ucrt\DLLs. Nothing here is OS specific just Windows 7 and newer when needed. So adding them will be easier. When certain UCRT not exist in system32 / syswow64 then these files will be expanded to VCMI install directory to make it universal on any Windows 7 and newer system. This will work with modded windows images aswell, when users have not present these packages in system.

GeorgeK1ng:
I think thats currently all we need for building new installer with Inno Setup.

New installers builded from v1.6.1 release for testing.
https://www.mediafire.com/folder/za10rbwveqk3j/VCMI

kambala-decapitator (changes requested):
not sure what dlls are doing in the repo - they should be copied from the local machine at the install time. This would also ensure that proper dlls are copied when ucrt we build against changes. Otherwise all good (not an Inno guy, so haven't really checked the script).

This PR should be merged with squash.

GeorgeK1ng:
These dlls are UCRT redist from Visual Studio, they are extracted to install folder at install time only when they are not preset in machine. This is here to be able to run VCMI on pure Windows 7 / 8.0 / 8.1. As when they are not present, VCMI can’t be launched.

kambala-decapitator:
> These dlls are UCRT redist from Visual Studio, they are extracted to install folder at install time only when they are not preset in machine. This is here to be able to run VCMI on pure Windows 7 / 8.0 / 8.1. As when they are not present, VCMI can’t be launched.

yes, I understand. The point is that they don't have to be stored in the repo as they're present on any development machine (like CI) that has msvc (note that it's not related to Visual Studio installation). For example, I have:

    c:\Program Files (x86)\Windows Kits\10\Redist\10.0.20348.0\ucrt\DLLs\x64\

so when CI (or a dev) builds installer, dlls should be copied from such a local path - it ensures that dlls are of the same version that was used to produce VCMI binaries.

GeorgeK1ng:
Does they really exist in CI? Can you check it? If they are there, then we don’t need them. For building script its just needed to point path. If they exist in CI, then we don’t need to store them.

kambala-decapitator:
> Does they really exist in CI? Can you check it? If they are there, then we don’t need them. For building script its just needed to point path. If they exist in CI, then we don’t need to store them.
> 
> Also note that build number changes with WDK updates

of course they exist as without a WDK installed you can't really build anything beyond basic console programs :) it contains WinAPI headers, libs etc. You can check here what CI has: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

yes, WDK version can't be hardcoded, probably CMake already has knowledge about it

CI/wininstaller/build_installer.cmd
```diff
@@ -38,6 +37,15 @@ if exist "%WinDir%\SysWow64" (
     set "ProgFiles=%programfiles%"
 )
 
+REM Dynamically locate the UCRT path
+set "UCRTBasePath=%ProgFiles%\Windows Kits\10\Redist"
+set "UCRTFilesPath="
+for /d %%d in ("%UCRTBasePath%\*") do (
+    if exist "%%d\ucrt\DLLs" (
```
kambala-decapitator, line 44: it should use user-provided directory name instead of selecting on its own. CMake has [CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION](https://cmake.org/cmake/help/latest/variable/CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION.html) with this info.

GeorgeK1ng:
For CMD script it's easy. So I have removed UCRT from repo. For setting CI must be defined same paramaters as are in build_installer.cmd for iscc.exe

CI/wininstaller/build_installer.cmd
GeorgeK1ng: I don't think Visual Studio users use CMake instead of "hit play" button. But I have modded script to have easy ability to provide custom folder. For example I have installed 10.0.19041.0 on my testing machine.

kambala-decapitator: visual studio project is generated by cmake, so... :)

later we'll adjust the installer script to pull info from cmake (probably just read some prepared file with directory name)

kambala-decapitator: still, your script pulls a random folder from WDK which is incorrect, as one can have multiple WDK installations

GeorgeK1ng: ISS script takes params correctly from ISCC.exe, can be used directly as fheroes2 have or if is needed to use this CMD script in CI it should be modded them to accept parameters itsefl.
But that CI is out of my knowledge.. :)

And if you wanna to implement some add_custom_command to CMakeList then it will be needed if wanna to build installer automatically.

GeorgeK1ng:
So I quess its ready, I don’t have for now any other ideas for installer. Just to tell users when we switch to new installer “Its recommended to uninstall previous VCMI if you upgrading from 1.6.x (2) or older release”. I have not added anything like auto removal of old NSIS setup.

CI/wininstaller/build_installer.cmd
GeorgeK1ng: > still, your script pulls a random folder from WDK which is incorrect, as one can have multiple WDK installations. 
Not random, latest available under Windows 10 SDK.

GeorgeK1ng: Now it correctly detect latest Windows 10 SDK in any cases. Changed that dir loop for detection to be sure.

kambala-decapitator (approved):
still not sure that this should go to beta though

GeorgeK1ng:
Yeah, depends how @IvanSavenko decide. Or in which upcoming version want to switch to modern installer.

It would be great if we can sign that installer with code signing certificate "VCMI Community" to avoid "Unknow publisher" on setup or download, but seems all certs starts from about 100$ per year.

CI/wininstaller/build_installer.cmd
```diff
@@ -0,0 +1,95 @@
+@echo off
+title VCMI Installer Builder
+setlocal enabledelayedexpansion
+cls
+
+REM Define variables dynamically relative to the normalized base directory
+set "AppVersion=1.6.1"
+set "AppBuild=1122334455A"
+set "InstallerArch=x64"
```
IvanSavenko, line 9: We also package 32-build for ancient systems

CI/wininstaller/build_installer.cmd
```diff
@@ -0,0 +1,95 @@
+@echo off
+title VCMI Installer Builder
+setlocal enabledelayedexpansion
+cls
+
+REM Define variables dynamically relative to the normalized base directory
+set "AppVersion=1.6.1"
```
IvanSavenko, line 7: Can this be updated automatically in some way?

CI/wininstaller/build_installer.cmd
IvanSavenko, line 1: Should be run as part of CI, as new, Windows-only step

IvanSavenko (commented):
Let's postpone it to 1.7, so I don't have to make new update release process for hotfixes.

Also - this needs to run as part of CI. Don't want to run installer generation locally for every hotfix. Also needed to detect any issues (e.g. broken script / translations) immediately, and not when I run installer just before release.

CI/wininstaller/build_installer.cmd
IvanSavenko, line 1: Also - perhaps move to `win/installer` directory?

CI/wininstaller/build_installer.cmd
GeorgeK1ng: If you want you use this CMD script for building I can't turn these variables into CMD script params aswell. 
I was thinking you can use similar method as is here https://github.com/ihhub/fheroes2/blob/2498a89c6f55222032f651812c512da22a45add9/.github/workflows/msvc.yml#L86

Whole installer is written to support both x86 and x64 packaging using iscc.exe params.

IvanSavenko: If this script is just for testing / as example - then fine.

Still needs integration into workflow, similar to fheroes2.

Current installer can be removed & that workflow step should be replaced with simple .zip packaging instead

CI/wininstaller/build_installer.cmd
GeorgeK1ng: Let me know where you want to have it. I can adjust then CMD script which is currently written to detect script path -2 directories levels to grab needed files.

CI/wininstaller/build_installer.cmd
GeorgeK1ng: Updated script to parse these 4 needed variables as params.

`Build_Installer.cmd AppVersion AppBuild InstallerArch VCMIFolder`

GeorgeK1ng:
It's fine to have it in v1.7 as part of release news. This PR just adds needed files which must be set / used later when times come. They don't change anything in current installer build process. But it updates translation docs page with informations for Inno Setup installer translations.

GeorgeK1ng:
> @GeorgeK1ng We should get rid of this batch file and use CPack: https://cmake.org/cmake/help/latest/cpack_gen/innosetup.html
> 
> Newest version of cmake also supports wine. So we could build it on linux with binfmt (and of course windows). Have done some tests for CI, but will not finish it because it should converted to cmake first: [develop...Laserlicht:vcmi:test_iss](https://github.com/vcmi/vcmi/compare/develop...Laserlicht:vcmi:test_iss)

Are you sure about custom ISS script usage with CPACK? From what I remember it was ignored and not supported with CPACK. Final builded installer was not builded with custom script.

kambala-decapitator:
> minimal test setup with ubuntu, wine and newest cmake

don't waste time on Linux -> Windows builds, they will be removed soon

GeorgeK1ng:
> > Are you sure about custom ISS script usage with CPACK? From what I remember it was ignored and not supported with CPACK. Final builded installer was not builded with custom script.
> 
> Custom script is loaded. It's basically imported at the top of script generated by cpack. But some variables has to removed from custom script, because they are setted by cpack and setting them twice cause an error.
> 

That's problem, you dropped for example default app dir, which is driven by ISS pascal code. Without that we loose defaul handling for appdata instead of program files for non admin users.

PR #5148, opened 2024-12-25
[1.6.x] h3c to vcmp converter

Laserlicht:
Adds converter to convert H3C to VCMP (also converts H3M inside it to VMAP).

Should be safe for 1.6.1 as there are almost no change of existing code.

Implementing a campaign editor should be much easier now. But not on my todo.

grafik

### DIFF
--- a/lib/campaign/CampaignHandler.cpp
+++ b/lib/campaign/CampaignHandler.cpp
@@ -50,12 +50,14 @@ void CampaignHandler::readCampaign(Campaign * ret, const std::vector<ui8> & inpu
 	{
 		JsonNode jsonCampaign(reinterpret_cast<const std::byte*>(input.data()), input.size(), filename);
 		readHeaderFromJson(*ret, jsonCampaign, filename, modName, encoding);
+		ret->overrideCampaign();
 
 		for(auto & scenario : jsonCampaign["scenarios"].Vector())
 		{
 			auto scenarioID = static_cast<CampaignScenarioID>(ret->scenarios.size());
 			ret->scenarios[scenarioID] = readScenarioFromJson(scenario);
 		}
+		ret->overrideCampaignScenarios();
 	}
 }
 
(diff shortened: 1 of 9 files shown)


Olegmods:
> Adds converter to convert H3C to VCMP (also converts H3M inside it to VMAP).
> 
> Should be safe for 1.6.1 as there are almost no change of existing code.
> 
> Implementing a campaign editor should be much easier now. But not on my todo.
> 
> grafik

Does it work with HotA campaigns? The old ones, I mean.

Laserlicht:
> Does it work with HotA campaigns? The old ones, I mean.

For all maps / campains that are loadable by vcmi. So I think no!?

Olegmods:
> > Does it work with HotA campaigns? The old ones, I mean.
> 
> For all maps / campains that are loadable by vcmi. So I think no!?

I meant the ones that can be played in VCMI (tried these with HotA mod, they are definitely playable, just some HotA-related bonuses are broken.

Laserlicht:
Then it could work.

misiokles:
Missing hota related bonuses can be added by editing jsons then?

Olegmods:
> Missing hota related bonuses can be added by editing jsons then?

Well, that's what I want to do.

Laserlicht:
Tested it. HotA campaigns are possible to extract and to launch from vcmp.

@misiokles With this tool you also may able fix Chronicles Enhanced. Export it and change portraits icons manually in jsons. In `campaignOverrides.json` you can also see the additional campaign parameters needed for a perfect integration with menu, videos and voices from chronicles.

lib/campaign/CampaignHandler.cpp
```diff
@@ -203,56 +225,86 @@ CampaignScenario CampaignHandler::readScenarioFromJson(JsonNode & reader)
...
+	node["regionText"].String() = scenario.regionText.toString();
+	node["prolog"] = prologEpilogWriter(scenario.prolog);
+	node["epilog"] = prologEpilogWriter(scenario.epilog);
+
+	writeScenarioTravelToJson(node, scenario.travelOptions);
+
+	return node;
+}
+
+std::map<std::string, CampaignStartOptions> startOptionsMap = {
```
IvanSavenko, line 258: Please make theses globals `static const`

Laserlicht:
@IvanSavenko Anything left to do here?

lib/campaign/CampaignHandler.cpp
```diff
@@ -50,12 +50,14 @@ void CampaignHandler::readCampaign(Campaign * ret, const std::vector<ui8> & inpu
 	{
 		JsonNode jsonCampaign(reinterpret_cast<const std::byte*>(input.data()), input.size(), filename);
 		readHeaderFromJson(*ret, jsonCampaign, filename, modName, encoding);
+		ret->overrideCampaign();
```
IvanSavenko, line 53: Shouldn't campaigns in json format already contain complete data? Reason why we have these overrides is due to some of hardcoded logic in h3 campaigns

Laserlicht: Yes, you're right.

PR #5151, opened 2024-12-25
Cherrypick of translations from develop branch

IvanSavenko:
- Cherry-picked Swedish and Portuguese translations from develop
- Regenerated .ts files, all obsolete entries have been removed
- Fixed typo in Launcher string: many -> may


XCOM-HUB:
If I understand correctly, I should only translate the "unfinished" parts and ignore the "vanished" and "obsolete" ones!?

IvanSavenko:
@XCOM-HUB 
Correct. Obsolete and vanished are old strings that are no longer used.

You can also use Qt Linguist (if you not using it already), it should show you only entries that need to be translated. You can get it for example here: https://download.qt.io/linguist_releases/

XCOM-HUB:
Thanks @IvanSavenko for the pro tip. I was not using "Qt Linguist". I will check it out.

PR #5158, opened 2024-12-26
[1.6.3] Bonus caching improvements

IvanSavenko:
This PR adds faster version of bonus system caching for parts of bonus system that are the most actively used by AI (both NKAI and BattleAI). Now cache hit event is basically instant (atomic access) and is completely lock-free. Cache misses have been optimized to minimize overhead and should also be faster than before

Unlike existing caching, this system is not very generic, and instead has parts made for a specific purpose - cache for all hero primary skills (since they are usually used & updated together), cache for per-turn values in battles (since AI estimates result of his actions), etc. So as a downside, it can't be easily reused to replace existing caching system, so old system will stay as it - it is used in large number of places and does not interferes with new system.

Total performance improvement is smaller than I hoped unfortunately - about 10%-20%, but better than nothing.

TODO (separate PR's):
- Implement more efficient bonus system node invalidation logic - there is no need to drop all cache values for entire gamestate if a hero equips an artifact.
- Try to optimize pathfinder since it now has even bigger chunk of AI turn
- Review slow routines in battle callback code that have several very inefficient methods. Will do after #5044 to avoid conficts and to see what still needs improvements after that PR


IvanSavenko:
Added more improvements:
- better bonus caching for pathfinder
- moved some trivial methods used by pathfinder to headers - too much overhead from calls to simple getters
- per-node bonus system invalidation
- removed overengineered system of hero movement speed updater - with bonus updater that depends on value of other bonuses.

Now need to test everything. Seems to be stable, but need to test and see whether there are any side effects, especially in bonus system.

Also found A LOT OF poorly designed code in both BattleAI and in BattleCallback that is *very* suboptimal, like having complexity of O(n*m) where O(n+m) would suffice. But definitely not for this PR. And only after battlehex PR

PR #5160, opened 2024-12-26
[1.6.2] show building names with ALT

Laserlicht:
same as adventure map:

grafik

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -36,6 +36,7 @@
 #include "../render/IRenderHandler.h"
 #include "../render/CAnimation.h"
 #include "../render/ColorFilter.h"
+#include "../render/IFont.h"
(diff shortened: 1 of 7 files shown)


client/windows/CCastleInterface.cpp
```diff
@@ -646,6 +649,54 @@ void CCastleBuildings::recreate()
...
+		auto building = buildingRect->getBuilding();
+		if (!building)
+			continue;
+		int bid = building->bid;
+
+		auto overlay = town->getTown()->buildings.at(bid)->getNameTranslated();
+		const auto & font = GH.renderHandler().loadFont(FONT_TINY);
+
+		auto backColor = Colors::GREEN; // Other
+		if(bid >= 30)
```
IvanSavenko, line 669:
```suggestion
		if(bid.IsDwelling())
```

PR #5180, opened 2025-01-01
CMessage::breakText fixed

### LINKED ISSUE #5016: Color hex values are displayed verbatim on artifacts
**Describe the bug**
Color hex values are displayed verbatim on artifacts

**Game logs**
-

**To Reproduce**
1. Start a game
2. Obtain an artifact that is part of a set

**Expected behavior**
Set members have colored text

**Actual behavior**
Text is not colored, instead the hex value is displayed directly

**Did it work earlier?**
Not sure

**Screenshots**
image

**Version**
 - Windows
 - `develop`, 22e820b5995ab20fa802b3e7c823613ab85e7ecb

SoundSSGood:
Fixed https://github.com/vcmi/vcmi/issues/5016
Added tool for сcolored text validation


Laserlicht:
There may many edge cases to check:
- does shortening of text work even with colored text?
- does automatic linebreak keep color?
- we've allowed multiline colors in the past - it seems this PR remove this functionality (e.g. `{test\ntest2}`)

I've also tested one of your builds. It broke the credits.

client/windows/CMessage.cpp
```diff
@@ -60,125 +60,222 @@ void CMessage::dispose()
 		item.reset();
 }
 
-std::vector<std::string> CMessage::breakText(std::string text, size_t maxLineWidth, EFonts font)
+bool CMessage::validateTags(
+	const std::vector<std::string_view::const_iterator> & openingTags,
+	const std::vector<std::string_view::const_iterator> & closingTags)
```
IvanSavenko, line 65: Not sure about using string view to be honest - it is too fragile, even more fragile than pointers.
Example:
```cpp
void foo(std::string_view value);

int main()
{
    std::string someString = "helloworld";
    foo(std::string("abcd")); // Illegal! std::string is destroyed after constructing its string view
    foo(someString.substr(1,2)); // Illegal! substr returns temporary string that will be destroyed
}
```

Your code works only because `std::string_view::const_iterator` is basically `const char*`, but this is clearly undefined behavior.

IMO we should use std::strings (they usually have short string optimization anyway) and only consider other approaches for performance-critical parts.

SoundSSGood: > IMO we should use std::strings (they usually have short string optimization anyway) and only consider other approaches for performance-critical parts.

Ok, I'll try to do it based on string

SoundSSGood:
> Rather complicated change. There may many edge cases to check:
> 
> * does shortening of text work even with colored text?
> * does automatic linebreak keep color?
> * we've allowed multiline colors in the past - it seems this PR remove this functionality (e.g. `{#ff0000|test\ntest2}`)
> 
> I've also tested one of your builds. It broke the credits.

All of the above should work. 
Will check credits.

Laserlicht:
Noticed one thing that are broken (worked in the past):
- `{green|test}`

Laserlicht:
It's working now. :)

client/windows/CMessage.cpp
```diff
@@ -60,125 +60,227 @@ void CMessage::dispose()
...
-			/* We don't count braces in string length. */
-			if(text[currPos] == '{')
+std::vector<CMessage::coloredline> CMessage::getPossibleLines(
+	const std::string & line, size_t maxLineWidth, const EFonts font, const char & splitSymbol)
+{
+	const auto & fontPtr = GH.renderHandler().loadFont(font);
+	std::vector<coloredline> result;
+	std::string_view lastFoundColor;
+	const std::function<size_t(const std::string_view&)> findColorTags = [&](const std::string_view & subLine) -> size_t
+		{
```
IvanSavenko, line 114: Might be better to have them as separate method. Large lambdas inside method are hard to read

client/windows/CMessage.cpp
```diff
@@ -60,125 +60,227 @@ void CMessage::dispose()
...
-				printableString.append(text.data() + currPos, symbolSize);
-			currPos += symbolSize;
-		}
+			return foundWidth;
+		};
 
-		// not all line has been processed - it turned out to be too long, so erase everything after last word break
-		// if string consists from a single word (or this is Chinese/Korean) - erase only last symbol to bring line back to allowed length
-		if(currPos < text.length() && (text[currPos] != 0x0a))
+	const std::function<void(const std::string_view&)> splitBySymbol = [&](const std::string_view & subLine)
```
IvanSavenko, line 147: Same here

client/windows/CMessage.cpp
```diff
@@ -60,125 +60,227 @@ void CMessage::dispose()
...
-		{
-			// trim only if line does not starts with LF
-			// FIXME: necessary? All lines will be trimmed before returning anyway
-			boost::algorithm::trim_left_if(text, boost::algorithm::is_any_of(std::string(" ")));
-		}
+	// Split text by lines
+	auto possibleLines = getPossibleLines(*preparedTextPtr, 0, font, breakSymbol);
 
-		if(opened)
+	// Add additional empty lines if needed. In Case "....\n\n...."
```
IvanSavenko, line 240: Perhaps this could be avoided by splitting input into sections to begin with? E.g. something like:
```cpp
std::vector<std::string> result;
std::vector<std::string> inputLines;
boost::split(inputLines, inputText, boost::is_any_of("\n"));
for (const auto & inputLine)
    result += getPossibleLines(inputLine)
```

With this approach there won't any need to handle line breaks in input text

client/windows/CMessage.h
```diff
@@ -26,15 +26,32 @@ class CMessage
 {
 	/// Draw simple dialog box (borders and background only)
 	static SDL_Surface * drawDialogBox(int w, int h, PlayerColor playerColor = PlayerColor(1));
+	static std::vector<size_t> getAllPositionsInStr(const char & targetChar, const std::string_view & str);
 
 public:
+	struct coloredline
```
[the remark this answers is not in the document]
SoundSSGood: > Also - `TextSegment` / `TextChunk` / `TextSection` might be better names

I couldn't find these names in the code.

client/windows/CMessage.cpp
SoundSSGood: This approach may break the case where the colored string needs to be split "{..../n /n ....}"

client/windows/CMessage.h
```diff
@@ -26,15 +26,32 @@ class CMessage
...
 	/// Draw border on exiting surface
 	static void drawBorder(PlayerColor playerColor, Canvas & to, int w, int h, int x, int y);
 
 	static void drawIWindow(CInfoWindow * ret, std::string text, PlayerColor player);
+	static bool validateColorBraces(const std::string & text);
+	static std::vector<coloredline> getPossibleLines(const std::string & line, size_t maxLineWidth, const EFonts font, const char & splitSymbol);
 
 	/// split text in lines
-	static std::vector<std::string> breakText(std::string text, size_t maxLineWidth, EFonts font);
+	static std::vector<std::string> breakText(const std::string & text, size_t maxLineWidth, const EFonts font);
```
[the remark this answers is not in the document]
SoundSSGood: Moved to lib. Made unittest.

client/windows/CMessage.cpp
```diff
@@ -60,125 +60,227 @@ void CMessage::dispose()
...
+	if(text.find("\r\n") != std::string::npos)
+	{
+		preparedText = text;
+		preparedTextPtr = &preparedText;
+		preparedText.erase(std::remove(preparedText.begin(), preparedText.end(), '\r'), preparedText.end());
+	}
+	else if(text.find('\r') != std::string::npos)
+	{
+		breakSymbol = '\r';
+	}
```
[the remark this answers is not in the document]
SoundSSGood: I didn't see any like that in the VCMI. Removed.

client/windows/CMessage.cpp
SoundSSGood: Done

client/windows/CMessage.cpp
SoundSSGood: This lambda doesn't make sense outside of getPossibleSublines. I think adding an extra entity will make it harder to understand the code logic.

GeorgeK1ng:
@SoundSSGood Can you please resolve conficts for this PR and let actions build artifacts? I would like to check it, if this fixes these issues https://github.com/vcmi/vcmi/issues/5016 + https://github.com/vcmi/vcmi/issues/5505

SoundSSGood:
@GeorgeK1ng

PR #5185, opened 2025-01-02
Fix blocked paths

DjWarmonger:
Fixed regression from 1.6.0 where curved path could lead to blocked Monolith

### DIFF
--- a/lib/rmg/RmgObject.cpp
+++ b/lib/rmg/RmgObject.cpp
@@ -271,6 +271,11 @@ const int3 & Object::getPosition() const
 int3 Object::getVisitablePosition() const
 {
 	assert(!dInstances.empty());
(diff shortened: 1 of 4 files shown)


lib/rmg/RmgObject.cpp
```diff
@@ -271,6 +271,11 @@ const int3 & Object::getPosition() const
 int3 Object::getVisitablePosition() const
 {
 	assert(!dInstances.empty());
+
+	// FIXME: This doesn't take into account Hota monster offset
+	if (guarded)
+		return getGuardPos();
+
```
IvanSavenko, line 278: It does, from what I see. Since `getGuardPos` returns visitable position, it should work with hota guards

PR #5187, opened 2025-01-02
Improved "Create new map" dialog for map editor

GeorgeK1ng:
This PR was for just rearrange, but it involved into review of whole dialog.

1) Redesigned
2) Added ability to choose between Standard size vs Custom size
3) Replaced lineEdits with spinBoxes
4) Fixed disabling of whole Random map group instead of checkbox only
5) Fixed Humans vs Computer dropdown selection. Now selecting 8 Humans correctly set 0 Computers when more than 0 Computer were selected before changing Humans to 8
6) Removed help button in all dialogs accross map editor
7) Hide "Create new map" dialog when generating map using template starts
8) Changed texts to make it clear
     "Two level map"        -> "Underground"
     "Human/Computer"  -> "Humans"
     "Computer only"       -> "Computers"

From this
image

To this:
image
image

### DIFF
--- a/mapeditor/generatorprogress.cpp
+++ b/mapeditor/generatorprogress.cpp
@@ -19,10 +19,10 @@ GeneratorProgress::GeneratorProgress(Load::Progress & source, QWidget *parent) :
 	source(source)
 {
 	ui->setupUi(this);
+	
+	setWindowFlags(Qt::Dialog | Qt::WindowTitleHint | Qt::CustomizeWindowHint);
 
 	setAttribute(Qt::WA_DeleteOnClose);
-
-	setWindowFlags(Qt::Window);
 
 	show();
 }
--- a/mapeditor/windownewmap.cpp
+++ b/mapeditor/windownewmap.cpp
@@ -367,7 +367,7 @@ void WindowNewMap::on_cpuCombo_activated(int index)
 void WindowNewMap::on_randomMapCheck_stateChanged(int arg1)
 {
 	randomMap = ui->randomMapCheck->isChecked();
-	ui->templateCombo->setEnabled(randomMap);
+	ui->randomOptions->setEnabled(randomMap);
 	updateTemplateList();
 }
 
--- a/mapeditor/windownewmap.cpp
+++ b/mapeditor/windownewmap.cpp
@@ -63,7 +63,8 @@ WindowNewMap::WindowNewMap(QWidget *parent) :
 	connect(ui->mapOpt2, &QRadioButton::toggled, this, &WindowNewMap::on_mapOpt2_toggled);
 
 	on_mapOpt1_toggled(true);
-	
+	on_checkSeed_toggled(false);
+
 	// size combo
 	connect(ui->sizeCombo, QOverload<int>::of(&QComboBox::activated), this, &WindowNewMap::on_sizeCombo_activated);
 	
--- a/mapeditor/windownewmap.cpp
+++ b/mapeditor/windownewmap.cpp
@@ -295,6 +295,8 @@ void WindowNewMap::on_okButton_clicked()
 			return;
 		}
 		
+		hide();
+
 		int seed = std::time(nullptr);
 		if(ui->checkSeed->isChecked() && ui->lineSeed->value() != 0)
 			seed = ui->lineSeed->value();
--- a/mapeditor/windownewmap.cpp
+++ b/mapeditor/windownewmap.cpp
@@ -221,16 +221,15 @@ std::unique_ptr<CMap> generateEmptyMap(CMapGenOptions & options)
 	return map;
 }
 
-int getSelectedMapSize(QComboBox* comboBox, int dimension) {
-	QString selectedText = comboBox->currentText();
-	QRegularExpression regex(R"(\((\d+)x(\d+)\))"); // Regex for format (width x height)
-	QRegularExpressionMatch match = regex.match(selectedText);
+std::pair<int, int> getSelectedMapSize(QComboBox* comboBox, const std::map<int, std::pair<int, int>>& mapSizes) {
+	int selectedIndex = comboBox->currentIndex();
 
-	if (match.hasMatch() && (dimension == 1 || dimension == 2)) {
-		return match.captured(dimension).toInt(); // Return width (1) or height (2)
+	auto it = mapSizes.find(selectedIndex);
+	if (it != mapSizes.end()) {
+		return it->second; // Return the width and height pair
 	}
 
-	return -1; // Return -1 if no match or invalid dimension
+	return { 0, 0 };
 }
 
 void WindowNewMap::on_okButton_clicked()
@@ -263,8 +262,9 @@ void WindowNewMap::on_okButton_clicked()
 	
 	if(ui->sizeStandardRadio->isChecked())
 	{
-		mapGenOptions.setWidth(getSelectedMapSize(ui->sizeCombo, 1));
-		mapGenOptions.setHeight(getSelectedMapSize(ui->sizeCombo, 2));
+		auto size = getSelectedMapSize(ui->sizeCombo, mapSizes);
+		mapGenOptions.setWidth(size.first);
+		mapGenOptions.setHeight(size.second);
 	}
 	else
 	{
@@ -319,8 +319,9 @@ void WindowNewMap::on_okButton_clicked()
 
 void WindowNewMap::on_sizeCombo_activated(int index)
 {
-	mapGenOptions.setWidth(getSelectedMapSize(ui->sizeCombo, 1));
-	mapGenOptions.setHeight(getSelectedMapSize(ui->sizeCombo, 2));
+	auto size = getSelectedMapSize(ui->sizeCombo, mapSizes);
+	mapGenOptions.setWidth(size.first);
+	mapGenOptions.setHeight(size.second);
 	updateTemplateList();
 }
 
--- a/mapeditor/windownewmap.cpp
+++ b/mapeditor/windownewmap.cpp
@@ -57,20 +57,9 @@ WindowNewMap::WindowNewMap(QWidget *parent) :
 		ui->cpuTeamsCombo->addItem(!i ? randomString : QString::number(cpuPlayers.at(i)));
 		ui->cpuTeamsCombo->setItemData(i, QVariant(cpuPlayers.at(i)));
 	}
-	
-	// map option radios
-	connect(ui->mapOpt1, &QRadioButton::toggled, this, &WindowNewMap::on_mapOpt1_toggled);
-	connect(ui->mapOpt2, &QRadioButton::toggled, this, &WindowNewMap::on_mapOpt2_toggled);
 
-	on_mapOpt1_toggled(true);
+	on_sizeStandardRadio_toggled(true);
 	on_checkSeed_toggled(false);
-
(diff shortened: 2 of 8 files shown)


GeorgeK1ng:
I have removed help button from dialog, not sure if is there way to do it for all dialogs at once instead of doing it in each dialog file code.

kambala-decapitator:
the last 3 group boxes look rather strange to me. Why not simply fit content in them instead of stretching to width?

GeorgeK1ng:
I just tweaked current state. There is of course space for more tweaking, for example radio for custom size map vs dropdown with values. Adding +/- to size line edits, checkbox to increase size on both edits same. 

Last 3 group were adjusted with spacer from both side, to not have radio at group line at left side. 
That was main goal, and set same size for each items. 

More improvements are welcome, but that’s something what’s must do someone more experienced. Last 3 group boxes I probably can rearrange.

GeorgeK1ng:
@kambala-decapitator I have fixed last 3 groups. Looks now much better.

GeorgeK1ng:
Seems I will try to add radios to handle selected map size from combo vs custom size and needed functionalities aswell.

GeorgeK1ng:
Ok, I have impelemented remaining ideas. 
Now is possible to choose  between Standard or Custom size map. 
Replaced lineEdits with spinBoxes

Everything seems to work.

GeorgeK1ng:
Added also full help button removal for all dialogs.

Improved generator UI, by hiding map settings when generating starts
From this:
image

To this:
image

GeorgeK1ng:
No more ideas or issues detected. Ready from my side now.

mapeditor/windownewmap.cpp
```diff
@@ -56,6 +58,19 @@ WindowNewMap::WindowNewMap(QWidget *parent) :
...
+
+	on_mapOpt1_toggled(true);
+	on_checkSeed_toggled(false);
+
+	// size combo
+	connect(ui->sizeCombo, QOverload<int>::of(&QComboBox::activated), this, &WindowNewMap::on_sizeCombo_activated);
+	
+	// spin boxes
+	connect(ui->widthTxt, QOverload<int>::of(&QSpinBox::valueChanged), this, &WindowNewMap::on_widthTxt_valueChanged);
+	connect(ui->heightTxt, QOverload<int>::of(&QSpinBox::valueChanged), this, &WindowNewMap::on_heightTxt_valueChanged);
```
IvanSavenko, line 73: I think Qt will automatically connect such signals with slots (since you're using `on_XXX_YYYChanged` form)

mapeditor/windownewmap.cpp
```diff
@@ -464,3 +503,23 @@ void WindowNewMap::on_cpuTeamsCombo_activated(int index)
 	updateTemplateList();
 }
 
+
+
+void WindowNewMap::on_mapOpt1_toggled(bool checked)
```
IvanSavenko, line 508: What is `mapOpt1` widget? Perhaps give it more clear name?

mapeditor/windownewmap.cpp
```diff
@@ -464,3 +503,23 @@ void WindowNewMap::on_cpuTeamsCombo_activated(int index)
...
+{
+	if (checked) {
+		ui->sizeGroup1->setEnabled(true);
+		ui->sizeGroup2->setEnabled(false);
+	}
+	updateTemplateList();
+}
+
+
+void WindowNewMap::on_mapOpt2_toggled(bool checked)
```
IvanSavenko, line 518: Same here

mapeditor/windownewmap.cpp
GeorgeK1ng: Yes, seems it connected it automatically. I have removed all useless connects.

mapeditor/windownewmap.cpp
GeorgeK1ng: Changed to sizeStandardRadio

mapeditor/windownewmap.cpp
GeorgeK1ng: Changed to sizeCustomRadio

mapeditor/generatorprogress.cpp
```diff
@@ -19,11 +19,11 @@ GeneratorProgress::GeneratorProgress(Load::Progress & source, QWidget *parent) :
 	source(source)
 {
 	ui->setupUi(this);
+	
+	setWindowFlags(Qt::Dialog | Qt::WindowTitleHint | Qt::CustomizeWindowHint);
```
kambala-decapitator, line 23: are you trying to remove the "question mark" button on Windows for all dialogs? There's an easier way:
```cpp
setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
```

mapeditor/windownewmap.cpp
```diff
@@ -217,6 +221,18 @@ std::unique_ptr<CMap> generateEmptyMap(CMapGenOptions & options)
 	return map;
 }
 
+int getSelectedMapSize(QComboBox* comboBox, int dimension) {
+	QString selectedText = comboBox->currentText(); // Get the selected text
```
kambala-decapitator, line 225:
```suggestion
	QString selectedText = comboBox->currentText();
```

mapeditor/windownewmap.cpp
```diff
@@ -217,6 +221,18 @@ std::unique_ptr<CMap> generateEmptyMap(CMapGenOptions & options)
...
+	QRegularExpression regex(R"(\((\d+)x(\d+)\))"); // Regex for format (width x height)
```
kambala-decapitator, line 226: that's a questionable approach. I believe all available combobox options have format WxH, so you could avoid regex and simply split on `x`:
```cpp
return selectedText.split(QLatin1Char{'x'})[dimension - 1].toInt();
```
dimension being 1 or 2 is also a strange choice: you could return `QSize` object and then read whatever component you need. As I see, you use both components immediately, so returning QSize would be much better - making a single call, and you won't need the `dimension` param.
```cpp
const auto components = selectedText.split(QLatin1Char{'x'});
return QSize{components[0].toInt(), components[1].toInt()};
```

mapeditor/windownewmap.h
```diff
@@ -73,7 +74,7 @@ class WindowNewMap : public QDialog
 		{3, {CMapHeader::MAP_SIZE_XLARGE,	CMapHeader::MAP_SIZE_XLARGE}},
 		{4, {CMapHeader::MAP_SIZE_HUGE,		CMapHeader::MAP_SIZE_HUGE}},
 		{5, {CMapHeader::MAP_SIZE_XHUGE,	CMapHeader::MAP_SIZE_XHUGE}},
-		{6, {CMapHeader::MAP_SIZE_GIANT,	CMapHeader::MAP_SIZE_GIANT}},
+		{6, {CMapHeader::MAP_SIZE_GIANT,	CMapHeader::MAP_SIZE_GIANT}}
```
kambala-decapitator, line 77: why is this change needed?

mapeditor/windownewmap.cpp
GeorgeK1ng: Combo text is not simply "36x36". It's "S (36x36)". So it's can't be simply splited by "x".
image.

mapeditor/generatorprogress.cpp
GeorgeK1ng: Yes, that works too. But I have choosed full flags control instead. So instead of just removing "?" help hutton. I'm setting what exactly should be there - Title + Close button and overwrite all flags there. There are probably no flags set.

mapeditor/windownewmap.h
GeorgeK1ng: Probably not needed, I have tried to add new map size "I (999x999)" - MAP_SIZE_INSANE accross game and in CRmgTemplate.cpp is not "," at the end so I did it same after reverting my changes.
https://github.com/vcmi/vcmi/blob/543e241769fa8986080031d59da2061f1788dfd9/lib/rmg/CRmgTemplate.cpp#L903

mapeditor/windownewmap.cpp
GeorgeK1ng: I don't think there si better approach to get that size from text.

mapeditor/windownewmap.h
GeorgeK1ng: And here I think is valid approach without traling comma at the end.

mapeditor/windownewmap.cpp
IvanSavenko: Potential problem - if somebody were to translate these text, this could easily break. For example by using cyrillic 'Х' symbol in place of latin 'X'

Better solutions:
a) Fill combobox in code, on UI initialization AND set user data to actual size, e.g:
```
ui->sizeCombo->addItem(tr("S (36x36)"), QVariant(36));
```
And then retrieve said index, via `itemData` if I remember correctly.

b) Since combobox elements are known in advance, add fixed mapping for item index -> map size, e.g:
```cpp
static constexpr std::array indexToMapSize = { 36, 72, 108 ... }
int index = ui->sizeCombo->selectedIndex();
int mapSize = indexToMapSize[index];
```

mapeditor/generatorprogress.cpp
GeorgeK1ng: And for this generatorprogress is not only about "?" help button removal. It's also about "x" close button removal to never allow cancelling map generating which will lead into crash.

mapeditor/windownewmap.cpp
GeorgeK1ng: Got different idea right now. I forgot about that mapSizes vs Combo index.

std::pair<int, int> getSelectedMapSize(QComboBox* comboBox, const std::map<int, std::pair<int, int>>& mapSizes) {
	int selectedIndex = comboBox->currentIndex();

	auto it = mapSizes.find(selectedIndex);
	if (it != mapSizes.end()) {
		return it->second; // Return the width and height pair
	}

	return { 0, 0 };
}

mapeditor/windownewmap.h
kambala-decapitator: anything that accepts initializer list allows trailing comma since C++11 (enum cases as well btw). It's better to add it because when you add new entries in the future, you won't have to touch the last existing line.

mapeditor/windownewmap.cpp
kambala-decapitator: yeah, now it's way better

mapeditor/windownewmap.h
GeorgeK1ng: Well, it was only place where it was. All three sections in that file have now same standard. And I don't think there is currently any plan in adding new map sizes / more players.

image

kambala-decapitator: I suggest to revert this comma change. And you could also remove extra tab in MAP_SIZE_HUGE.

GeorgeK1ng: This is how it looks after extra tab removal.
image

kambala-decapitator: different tab width settings :) then it should be separated with spaces to avoid misalignment

GeorgeK1ng: OK. 
image

kambala-decapitator (approved):
Pr should be merged with squash

PR #5191, opened 2025-01-02
[1.6.x] mapeditor translations

### LINKED ISSUE #5188: Translation issues in Map Editor
I have found various issues with translations in Map Editor.

In inspector window are Owner, Army etc - not translatable
image

Also colors are not translatable in all places in editor
image

Window title is not translatable
image

Whole validator
image

Laserlicht:
fixes #5188


Laserlicht:
Should be ready.

Fixed also existing translations, that never worked...

@IvanSavenko Can you merge soon? As I needed to regenerate translations and I want to avoid conflicts. And also to give peoples time to also translate missing strings.

IvanSavenko:
Will check & merge this and other ready PR's after 1.6.2.
Either later today or tomorrow.

mapeditor/inspector/inspector.cpp
```diff
@@ -913,6 +909,49 @@ QTableWidgetItem * Inspector::addProperty(CGCreature::Character value)
 
 Inspector::Inspector(MapController & c, CGObjectInstance * o, QTableWidget * t): obj(o), table(t), controller(c)
 {
+	reusedTr = {
```
IvanSavenko, line 912: There is generally no need to perform manual deduplication - lupdates automatically groups identical strings

Laserlicht: Wasn't aware of that. Thanks!

Changed it! :)

IvanSavenko: Yeah. Saw this multiple times when resolving conflicts in .ts files :)

Works only within same 'context' (e.g. in the same class). For same translation in different files lupdate uses different mechanism - existing translation will be copied to a new location, but will be marked as 'unfinished' for validation by translator.

PR #5208, opened 2025-01-05
map editor: show edited values in inspector

godric3:
Improved inspector to show real values instead of `...`
As description of most values can get quite long I added also tooltip. Might be overkill for quests and rewards, but better than nothing

image
image
image

### DIFF
--- a/mapeditor/inspector/armywidget.cpp
+++ b/mapeditor/inspector/armywidget.cpp
@@ -158,9 +158,5 @@ void ArmyDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex
 			textList += QString("%1 %2").arg(stack->count).arg(QString::fromStdString(stack->getCreature()->getNamePluralTranslated()));
 	}
 
-	QString text = textList.join("\n");
-	model->setItemData(index, {
-		{Qt::DisplayRole, QVariant{text}},
-		{Qt::ToolTipRole, QVariant{text}}
-	});
+	setModelTextData(model, index, textList);
 }
--- a/mapeditor/inspector/armywidget.cpp
+++ b/mapeditor/inspector/armywidget.cpp
@@ -115,7 +115,7 @@ ArmyWidget::~ArmyWidget()
 
 
 
-ArmyDelegate::ArmyDelegate(CArmedInstance & t): army(t), QStyledItemDelegate()
+ArmyDelegate::ArmyDelegate(CArmedInstance & t): army(t), BaseInspectorItemDelegate()
 {
 }
 
@@ -141,9 +141,27 @@ void ArmyDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, cons
 	if(auto * ed = qobject_cast<ArmyWidget *>(editor))
 	{
 		ed->commitChanges();
+		updateModelData(model, index);
 	}
 	else
 	{
 		QStyledItemDelegate::setModelData(editor, model, index);
 	}
+}
+
+void ArmyDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex & index) const
+{
+	QStringList textList;
+	for(auto i = army.stacks.begin(); i != army.stacks.end(); i++)
+	{
+		auto stack = i->second;
+		if(stack->count != 0 && stack->getCreature() != nullptr)
+			textList += QString::number(stack->count) + " " + QString::fromStdString(stack->getCreature()->getNamePluralTranslated());
+	}
+
+	QString text = textList.join("\n");
+	QMap<int, QVariant> data;
+	data[Qt::DisplayRole] = QVariant(text);
+	data[Qt::ToolTipRole] = QVariant(text);
+	model->setItemData(index, data);
 }
--- a/mapeditor/inspector/armywidget.cpp
+++ b/mapeditor/inspector/armywidget.cpp
@@ -152,16 +152,15 @@ void ArmyDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, cons
 void ArmyDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex & index) const
 {
 	QStringList textList;
-	for(auto i = army.stacks.begin(); i != army.stacks.end(); i++)
+	for(const auto & [_, stack] : army.stacks)
 	{
-		auto stack = i->second;
 		if(stack->count != 0 && stack->getCreature() != nullptr)
-			textList += QString::number(stack->count) + " " + QString::fromStdString(stack->getCreature()->getNamePluralTranslated());
+			textList += QString("%1 %2").arg(stack->count).arg(QString::fromStdString(stack->getCreature()->getNamePluralTranslated()));
 	}
 
 	QString text = textList.join("\n");
-	QMap<int, QVariant> data;
-	data[Qt::DisplayRole] = QVariant(text);
-	data[Qt::ToolTipRole] = QVariant(text);
-	model->setItemData(index, data);
+	model->setItemData(index, {
+		{Qt::DisplayRole, QVariant{text}},
+		{Qt::ToolTipRole, QVariant{text}}
+	});
 }
--- a/mapeditor/inspector/baseinspectoritemdelegate.h
+++ b/mapeditor/inspector/baseinspectoritemdelegate.h
@@ -1,0 +1,22 @@
+/*
+ * baseinspectoritemdelegate.h, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#pragma once
+
+#include <QStyledItemDelegate>
+#include <QAbstractItemModel>
+
+class BaseInspectorItemDelegate : public QStyledItemDelegate
+{
+    Q_OBJECT
+public:
+    using QStyledItemDelegate::QStyledItemDelegate;
+
+    virtual void updateModelData(QAbstractItemModel * model, const QModelIndex & index) const {};
+};
--- a/mapeditor/inspector/baseinspectoritemdelegate.h
+++ b/mapeditor/inspector/baseinspectoritemdelegate.h
@@ -19,4 +19,12 @@ public:
     using QStyledItemDelegate::QStyledItemDelegate;
 
     virtual void updateModelData(QAbstractItemModel * model, const QModelIndex & index) const {};
+    void setModelTextData(QAbstractItemModel * model, const QModelIndex & index, const QStringList & list) const
+    {
+        QString text = list.join("\n");
+        model->setItemData(index, {
+            {Qt::DisplayRole, QVariant{text}},
+            {Qt::ToolTipRole, QVariant{text}}
+        });
+    };
 };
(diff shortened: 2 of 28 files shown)


mapeditor/inspector/armywidget.cpp
```diff
@@ -141,9 +141,27 @@ void ArmyDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, cons
...
 	else
 	{
 		QStyledItemDelegate::setModelData(editor, model, index);
 	}
 }
+
+void ArmyDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex & index) const
+{
+	QStringList textList;
+	for(auto i = army.stacks.begin(); i != army.stacks.end(); i++)
```
kambala-decapitator, line 155: is it possible to use for-in?
```cpp
for(const auto & stack : army.stacks) // or [_, stack] instead of stack
```

mapeditor/inspector/armywidget.cpp
```diff
@@ -141,9 +141,27 @@ void ArmyDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, cons
...
+	{
+		auto stack = i->second;
+		if(stack->count != 0 && stack->getCreature() != nullptr)
+			textList += QString::number(stack->count) + " " + QString::fromStdString(stack->getCreature()->getNamePluralTranslated());
```
kambala-decapitator, line 159:
```suggestion
			textList += QString{"%1 %2"}.arg(stack->count).arg(QString::fromStdString(stack->getCreature()->getNamePluralTranslated()));
```

mapeditor/inspector/armywidget.cpp
```diff
@@ -141,9 +141,27 @@ void ArmyDelegate::setModelData(QWidget *editor, QAbstractItemModel *model, cons
...
+	}
+
+	QString text = textList.join("\n");
+	QMap<int, QVariant> data;
+	data[Qt::DisplayRole] = QVariant(text);
+	data[Qt::ToolTipRole] = QVariant(text);
+	model->setItemData(index, data);
```
kambala-decapitator, line 166:
```suggestion
	model->setItemData(index, {
		{Qt::DisplayRole, QVariant{text}},
		{Qt::ToolTipRole, QVariant{text}}
	});
```

mapeditor/inspector/baseinspectoritemdelegate.h
```diff
@@ -0,0 +1,22 @@
...
+#include <QStyledItemDelegate>
+#include <QAbstractItemModel>
+
+class BaseInspectorItemDelegate : public QStyledItemDelegate
+{
+    Q_OBJECT
+public:
+    using QStyledItemDelegate::QStyledItemDelegate;
+
+    virtual void updateModelData(QAbstractItemModel * model, const QModelIndex & index) const {};
```
kambala-decapitator, line 21: if it's not intended to create objects of type BaseInspectorItemDelegate, then this method can be made pure virtual:
```suggestion
    virtual void updateModelData(QAbstractItemModel * model, const QModelIndex & index) const = 0;
```

mapeditor/inspector/heroartifactswidget.cpp
```diff
@@ -136,9 +136,30 @@ void HeroArtifactsDelegate::setModelData(QWidget * editor, QAbstractItemModel *
...
 	}
 }
+
+void HeroArtifactsDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex & index) const
+{
+	QStringList textList;
+	for(const auto & [artPosition, artSlotInfo] : hero.artifactsWorn)
+	{
+		auto slotText = NArtifactPosition::namesHero[artPosition.num];
+		textList += QString::fromStdString(slotText) + ": " + QString::fromStdString(artSlotInfo.artifact->getType()->getNameTranslated());
```
kambala-decapitator, line 153: use QString{""}.arg() like above in all the other places where it makes sense

mapeditor/inspector/heroartifactswidget.cpp
```diff
@@ -136,9 +136,30 @@ void HeroArtifactsDelegate::setModelData(QWidget * editor, QAbstractItemModel *
...
+	textList += QString::fromStdString(NArtifactPosition::backpack) + ":";
+	for(const auto & art : hero.artifactsInBackpack)
+	{
+		textList += QString::fromStdString(art.artifact->getType()->getNameTranslated());
+	}
+	QString text = textList.join("\n");
+	QMap<int, QVariant> data;
+	data[Qt::DisplayRole] = QVariant(text);
+	data[Qt::ToolTipRole] = QVariant(text);
+	model->setItemData(index, data);
```
kambala-decapitator, line 164: create map in-place like above in all the other places

mapeditor/inspector/herospellwidget.cpp
```diff
@@ -127,9 +127,34 @@ void HeroSpellDelegate::setModelData(QWidget * editor, QAbstractItemModel * mode
...
+				textList += QString::fromStdString(spellID.toSpell()->getNameTranslated());
+		}
+		QString text = textList.join("\n");
+		data[Qt::DisplayRole] = QVariant(text);
+		data[Qt::ToolTipRole] = QVariant(text);
+	}
+	else
+	{
+		data[Qt::DisplayRole] = QObject::tr("Default Spells");
+		data[Qt::ToolTipRole] = QObject::tr("Default Spells");
```
kambala-decapitator, line 157:
```suggestion
		data[Qt::DisplayRole] = data[Qt::ToolTipRole] = QObject::tr("Default Spells");
```

mapeditor/inspector/inspector.cpp
```diff
@@ -905,6 +907,17 @@ QTableWidgetItem * Inspector::addProperty(CGCreature::Character value)
 	return item;
 }
 
+QTableWidgetItem * Inspector::addProperty(const std::optional<CGDwellingRandomizationInfo> & value)
+{
+	QString text = QObject::tr("Select town");
+	if(value.has_value() && !value->instanceId.empty())
```
kambala-decapitator, line 913: is it even possible that string exists but it's empty?

mapeditor/inspector/questwidget.cpp
```diff
@@ -439,3 +440,86 @@ bool QuestDelegate::eventFilter(QObject * object, QEvent * event)
...
+	{
+		if(resource == GameResID::MITHRIL)
+			continue;
+		if(quest.mission.resources[resource] == 0)
+			continue;
+		MetaString str;
+		str.appendName(resource);
+		resourcesList += QString::fromStdString(str.toString()) + ": " + QString::number(quest.mission.resources[resource]);
+	}
+	textList += QObject::tr("Resources: ") + resourcesList.join(", ");
```
kambala-decapitator, line 467: is it ok to have a single long string? Also better to make it a param of the user-visible string. (applies to all other similar places)

mapeditor/inspector/questwidget.cpp
```diff
@@ -439,3 +440,86 @@ bool QuestDelegate::eventFilter(QObject * object, QEvent * event)
...
+		str.appendName(player);
+		playersList += QString::fromStdString(str.toString());
+	}
+	textList += QObject::tr("Players: ") + playersList.join(", ");
+
+	QString text = textList.join("\n");
+	QMap<int, QVariant> data;
+	data[Qt::DisplayRole] = QVariant(text);
+	data[Qt::ToolTipRole] = QVariant(text);
+	model->setItemData(index, data);
```
kambala-decapitator, line 524: this code is repeated multiple times, better extract it to a helper function

mapeditor/inspector/townbuildingswidget.cpp
```diff
@@ -353,11 +353,41 @@ void TownBuildingsDelegate::setModelData(QWidget *editor, QAbstractItemModel *mo
...
 }
 
-
+void TownBuildingsDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex & index) const
+{
+	QStringList textList;
+	textList += QObject::tr("Built buildings:");
+	auto * ctown = town.getTown();
+	if(!ctown)
+		ctown = VLC->townh->randomTown;
```
kambala-decapitator, line 370: is it an often used pattern? If yes, better extract to a function

mapeditor/inspector/towneventswidget.cpp
```diff
@@ -169,9 +169,25 @@ void TownEventsDelegate::setModelData(QWidget * editor, QAbstractItemModel * mod
...
 	}
 }
+
+void TownEventsDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex & index) const
+{
+	QStringList eventLists;
+	for (const auto & event : town.events)
+	{
+		auto eventName = QString::fromStdString(event.name);
+		eventLists += tr("Day %1 - %2").arg(event.firstOccurrence + 1, 3).arg(eventName);
```
kambala-decapitator, line 186: what is 3? max 3 digits?

mapeditor/inspector/baseinspectoritemdelegate.h
godric3: Objects of type BaseInspectorItemDelegate are not created, but not all derived classes implements this method. Would be it better to add empty implementation in those classes or leave this method as is?

mapeditor/inspector/inspector.cpp
godric3: Yes, editor initialize it with empty string, so it could exist like that on some maps

mapeditor/inspector/towneventswidget.cpp
godric3: Day number will take minimum 3 characters and be right aligned, used same way as on list of events, here in fact it may be not needed

mapeditor/inspector/baseinspectoritemdelegate.h
kambala-decapitator: In such case empty implementation is fine. Unless you want to require subclasses to have explicit implementation.

mapeditor/inspector/questwidget.cpp
godric3: I think that in most cases those lists will have one or two entities, and then it's easier to see that some name is related to artifact or creature. Example how it look if all resources are required
image

kambala-decapitator: OK, let's leave like that now, can always be adjusted later if necessary

mapeditor/inspector/townbuildingswidget.cpp
IvanSavenko: Should be no longer necessary. Simple getTown should suffice:
```suggestion
	auto * ctown = town.getTown();
```

`getTown()` method now returns random town for well, random towns.

mapeditor/inspector/towneventswidget.cpp
```diff
@@ -179,15 +179,16 @@ void TownEventsDelegate::setModelData(QWidget * editor, QAbstractItemModel * mod
 
 void TownEventsDelegate::updateModelData(QAbstractItemModel * model, const QModelIndex & index) const
 {
-	QStringList eventLists;
+	QStringList eventList;
+	eventList += QObject::tr("Town Events:");
```
kambala-decapitator, line 183:
```suggestion
	QStringList eventList = QObject::tr("Town Events:");
```

kambala-decapitator (approved):

mapeditor/inspector/questwidget.cpp
kambala-decapitator: will this not be adjusted?

mapeditor/inspector/towneventswidget.cpp
godric3: Tried that already and it didn't work, but now I found `QStringList eventList(QObject::tr("Town Events:"))` ;)

mapeditor/inspector/questwidget.cpp
godric3: Sorry, missed that due to first making inline map change, will extract to base delegate class

kambala-decapitator (approved):

PR #5211, opened 2025-01-05
Map editor: random town same as player faction

godric3:
Feature mentioned at least twice on discord, and already present, just not in map editor.
based on beta but if not fit into it I can change to develop

### DIFF
--- a/lib/mapObjects/CGTownInstance.cpp
+++ b/lib/mapObjects/CGTownInstance.cpp
@@ -1134,6 +1134,7 @@ void CGTownInstance::serializeJsonOptions(JsonSerializeFormat & handler)
 		eventsHandler.syncSize(events, JsonNode::JsonType::DATA_VECTOR);
 		eventsHandler.serializeStruct(events);
 	}
+	handler.serializeId("alignmentToPlayer", alignmentToPlayer, PlayerColor::NEUTRAL);
 }
 
 const CFaction * CGTownInstance::getFaction() const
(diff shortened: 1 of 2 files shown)


mapeditor/inspector/inspector.cpp
```diff
@@ -612,6 +614,9 @@ void Inspector::setProperty(CGTownInstance * o, const QString & key, const QVari
 	if(key == QObject::tr("Town name"))
 		o->setNameTextId(mapRegisterLocalizedString("map", *controller.map(),
 			TextIdentifier("town", o->instanceName, "name"), value.toString().toStdString()));
+
+	if(key == QObject::tr("Same as player"))
```
kambala-decapitator, line 618: - comparing with a user-visible string is not a great idea, it should rather be some enum or internal string constant
- `else if`

lib/mapObjects/CGTownInstance.cpp
```diff
@@ -1134,6 +1134,7 @@ void CGTownInstance::serializeJsonOptions(JsonSerializeFormat & handler)
 		eventsHandler.syncSize(events, JsonNode::JsonType::DATA_VECTOR);
 		eventsHandler.serializeStruct(events);
 	}
+	handler.serializeId("alignmentToPlayer", alignmentToPlayer, PlayerColor::NEUTRAL);
```
kambala-decapitator, line 1137: @IvanSavenko please verify that this is ok

mapeditor/inspector/inspector.cpp
godric3: Would require fixing whole file, possibly in separate PR

kambala-decapitator: Ok, that's fine

lib/mapObjects/CGTownInstance.cpp
IvanSavenko: Looks OK. And since this method is only used for editor / vmap's it can't break base game.

PR #5212, opened 2025-01-05
Added popup for keymasters/border guards similar to teleporters

IvanSavenko:
Same logic as recently added popup for teleporter objects.

Also fixed display of popup on single-level maps - now it will only show surface minimap, without black rectangle for non-existing underground

зображення


Laserlicht:
Nice. :)

Maybe also for obelisks? Drawing all visible obelisks, and that missing ones in different color? Have to search all the time which one I've missed. :D

IvanSavenko:
Will think about obelisks tomorrow. Should be easy enough. 

Probably gold icon for non-visited ones, and grey icon for already visited.

PR #5219, opened 2025-01-06
Fixes for AI town building logic

IvanSavenko:
- Fixes AI not considering non-built dwellings
- AI will now skip evaluation of towns in which construction today is no longer possible

### DIFF
--- a/AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
+++ b/AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
@@ -11,6 +11,7 @@
 #include "../Engine/Nullkiller.h"
 #include "../Engine/Nullkiller.h"
 #include "../../../lib/entities/building/CBuilding.h"
+#include "../../../lib/IGameSettings.h"
 
 namespace NKAI
 {
@@ -30,30 +31,22 @@ void BuildAnalyzer::updateTownDwellings(TownDevelopmentInfo & developmentInfo)
 	for(int level = 0; level < developmentInfo.town->getTown()->creatures.size(); level++)
 	{
 		logAi->trace("Checking dwelling level %d", level);
-		BuildingInfo nextToBuild = BuildingInfo();
-
-		BuildingID buildID = BuildingID(BuildingID::getDwellingFromLevel(level, 0));
-
-		for(; developmentInfo.town->getBuildings().count(buildID); BuildingID::advanceDwelling(buildID))
+		for(BuildingID buildID = BuildingID::getDwellingFromLevel(level, 0); buildID.hasValue(); BuildingID::advanceDwelling(buildID))
 		{
-			if(!developmentInfo.town->hasBuilt(buildID))
+			if(developmentInfo.town->getTown()->buildings.count(buildID) == 0)
 				continue; // no such building in town
 
 			auto info = getBuildingOrPrerequisite(developmentInfo.town, buildID);
 
+			if (!info.exists && !info.canBuild && !info.notEnoughRes)
+				continue;
+
 			if(info.exists)
-			{
 				developmentInfo.addExistingDwelling(info);
+			else
+				developmentInfo.addBuildingToBuild(info);
 
-				break;
-			}
-
-			nextToBuild = info;
-		}
-
-		if(nextToBuild.id != BuildingID::NONE)
-		{
-			developmentInfo.addBuildingToBuild(nextToBuild);
+			break;
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
```diff
@@ -30,30 +31,22 @@ void BuildAnalyzer::updateTownDwellings(TownDevelopmentInfo & developmentInfo)
...
 		logAi->trace("Checking dwelling level %d", level);
-		BuildingInfo nextToBuild = BuildingInfo();
-
-		BuildingID buildID = BuildingID(BuildingID::getDwellingFromLevel(level, 0));
-
-		for(; developmentInfo.town->getBuildings().count(buildID); BuildingID::advanceDwelling(buildID))
+		for(BuildingID buildID = BuildingID::getDwellingFromLevel(level, 0); buildID.hasValue(); BuildingID::advanceDwelling(buildID))
 		{
-			if(!developmentInfo.town->hasBuilt(buildID))
+			if(developmentInfo.town->getTown()->buildings.count(buildID) == 0)
```
dydzio0614, line 36: why does "hasBuilt" not suffice?

AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
```diff
@@ -30,30 +31,22 @@ void BuildAnalyzer::updateTownDwellings(TownDevelopmentInfo & developmentInfo)
...
 			auto info = getBuildingOrPrerequisite(developmentInfo.town, buildID);
 
+			if (!info.exists && !info.canBuild && !info.notEnoughRes)
+				continue;
+
 			if(info.exists)
-			{
 				developmentInfo.addExistingDwelling(info);
+			else
+				developmentInfo.addBuildingToBuild(info);
```
dydzio0614, line 47: this will stop loop at first occurrence of !info.exists, rather than iterating until last one with info.exists is found so value in addBuildingToBuild will be different - what will this change?

AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
IvanSavenko: Because "has built" is for "was this building constructed in town".
But this code actually needed to check for "does building of such type exists in this town"

Cove might have 2nd upgrade of 3rd dwelling built (or not built), but Castle does not have such dwelling to begin with.

AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
```diff
@@ -30,30 +31,22 @@ void BuildAnalyzer::updateTownDwellings(TownDevelopmentInfo & developmentInfo)
...
-		for(; developmentInfo.town->getBuildings().count(buildID); BuildingID::advanceDwelling(buildID))
+		for(BuildingID buildID = BuildingID::getDwellingFromLevel(level, 0); buildID.hasValue(); BuildingID::advanceDwelling(buildID))
 		{
-			if(!developmentInfo.town->hasBuilt(buildID))
+			if(developmentInfo.town->getTown()->buildings.count(buildID) == 0)
 				continue; // no such building in town
 
 			auto info = getBuildingOrPrerequisite(developmentInfo.town, buildID);
 
+			if (!info.exists && !info.canBuild && !info.notEnoughRes)
```
dydzio0614, line 41: shouldn't be `!info.canBuild || !info.notEnoughRes` ?

AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
IvanSavenko: Should not change anything. This is actually part of the fix.
Compare this with code that was here before it was broken:
https://github.com/vcmi/vcmi/blob/60435a70e2a5615d0f1e49039e78a86d01286f87/AI/Nullkiller/Analyzers/BuildAnalyzer.cpp

It had reversed loop (upgrades 2/1/0), and then - last-most building was picked.
After PR #5174, loop direction has been changed, but it would still pick info from last-most iteration (so AI would attempt to build upgraded dwelling before base).

AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
```diff
@@ -18,35 +18,35 @@ namespace NKAI
...
+		for (const auto & buildID : dwellingsInTown)
+		{
+			if (developmentInfo.town->hasBuilt(buildID))
 				continue;
 
-			if(info.exists)
-				developmentInfo.addExistingDwelling(info);
-			else
+			const auto & info = getBuildingOrPrerequisite(developmentInfo.town, buildID);
+			if (info.canBuild || info.notEnoughRes)
```
dydzio0614, line 48: if not enough resources building is added to queue?

PR #5225, opened 2025-01-07
Various main menu improvements

GeorgeK1ng:
1) Added ability to use more random images in multiplayer selection (MuMap)
2) Added missing header text to multiplayer selection
3) Corrected header text for HotSeat / HostTCP / JoinTCP
4) Corrected position of "Connecting..." text
5) Added ability to use random loading bar squares
6) Added campaign buttons validation

No new translations needed, uses already present text

image

### DIFF
--- a/client/mainmenu/CMainMenu.cpp
+++ b/client/mainmenu/CMainMenu.cpp
@@ -469,18 +469,26 @@ CMultiMode::CMultiMode(ESelectionScreen ScreenType)
 	background = std::make_shared<CPicture>(ImagePath::builtin("MUPOPUP.bmp"));
 	pos = background->center(); //center, window has size of bg graphic
 
-	picture = std::make_shared<CPicture>(ImagePath::builtin("MUMAP.bmp"), 16, 77);
+	const auto& multiplayerConfig = CMainMenuConfig::get().getConfig()["multiplayer"];
+	if (multiplayerConfig.isVector() && !multiplayerConfig.Vector().empty())
+		picture = std::make_shared<CPicture>(ImagePath::fromJson(*RandomGeneratorUtil::nextItem(multiplayerConfig.Vector(), CRandomGenerator::getDefault())), 16, 77);
+
+	if (multiplayerConfig.isString())
+		picture = std::make_shared<CPicture>(ImagePath::fromJson(multiplayerConfig), 16, 77);
+
+	textTitle = std::make_shared<CTextBox>("", Rect(7, 18, 440, 50), 0, FONT_BIG, ETextAlignment::CENTER, Colors::WHITE);
+	textTitle->setText(CGI->generaltexth->zelp[263].second);
 
 	statusBar = CGStatusBar::create(std::make_shared<CPicture>(background->getSurface(), Rect(7, 465, 440, 18), 7, 465));
 	playerName = std::make_shared<CTextInput>(Rect(19, 436, 334, 16), background->getSurface());
 	playerName->setText(getPlayersNames()[0]);
 	playerName->setCallback(std::bind(&CMultiMode::onNameChange, this, _1));
 
-	buttonHotseat = std::make_shared<CButton>(Point(373, 78 + 57 * 0), AnimationPath::builtin("MUBHOT.DEF"), CGI->generaltexth->zelp[266], std::bind(&CMultiMode::hostTCP, this), EShortcut::MAIN_MENU_HOTSEAT);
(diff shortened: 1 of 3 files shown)


client/mainmenu/CMainMenu.cpp
```diff
@@ -469,18 +469,26 @@ CMultiMode::CMultiMode(ESelectionScreen ScreenType)
 	background = std::make_shared<CPicture>(ImagePath::builtin("MUPOPUP.bmp"));
 	pos = background->center(); //center, window has size of bg graphic
 
-	picture = std::make_shared<CPicture>(ImagePath::builtin("MUMAP.bmp"), 16, 77);
+	const auto& multiplayerConfig = CMainMenuConfig::get().getConfig()["multiplayer"];
+	if (multiplayerConfig.isVector() && !multiplayerConfig.Vector().empty())
+		picture = std::make_shared<CPicture>(ImagePath::fromJson(*RandomGeneratorUtil::nextItem(multiplayerConfig.Vector(), CRandomGenerator::getDefault())), 16, 77);
+
+	if (multiplayerConfig.isString())
+		picture = std::make_shared<CPicture>(ImagePath::fromJson(multiplayerConfig), 16, 77);
```
IvanSavenko, line 477: Probably needs fallback to `MuMap` if this key is not present - not everyone will update every mod on new release

client/mainmenu/CMainMenu.cpp
```diff
@@ -532,16 +540,29 @@ void CMultiMode::onNameChange(std::string newText)
...
+		text = CGI->generaltexth->allTexts[446];
+		boost::replace_all(text, "\t", "\n");
+		break;
+	case EShortcut::MAIN_MENU_HOST_GAME:
+		text = CGI->generaltexth->translate("vcmi.mainMenu.hostTCP");
+		break;
+	case EShortcut::MAIN_MENU_JOIN_GAME:
+		text = CGI->generaltexth->translate("vcmi.mainMenu.joinTCP");
+		break;
+	}
```
IvanSavenko, line 563: Shouldn't existing `ESelectionScreen` or `ELoadMode` enums suffice?

client/mainmenu/CMainMenu.cpp
GeorgeK1ng: I was thinking about that fallback too. But I have removed it and prepared update for HotA, WoG and H3 themes pack instead, as all of these mods have broken loading bar.

IvanSavenko: Not everyone updates mods ASAP, so game should work even if player have not update mod, especially for hotfixes.

You can mark it with `//mods compatibilty for 1.6.3` comment, and remove it in 1.7+

GeorgeK1ng: True, we’ll re-add it. Maybe fallbacks for backgrounds should be added and keep for case when certain element doesn’t exist in json?

IvanSavenko: Which backgrounds? This "mumap"?
IMO fallback is only needed to give some time for mods to update & for players to install update. Error message to log would also be a good idea btw.

Mods should generally provide all elements in cases like this one, if not - then this should be viewed as bug in a mod. Keeping fallback for long time may be useful, but only to avoid crashes.

client/mainmenu/CMainMenu.cpp
GeorgeK1ng: They are not enough as we need "sub multiplayer entry". Have not found better way to get it, only from EShortcut.

```
ESelectionScreen = newGame, loadGame, saveGame, scenarioInfo, campaignList
ELoadMode = NONE, SINGLE, MULTI, CAMPAIGN, TUTORIAL
```

client/mainmenu/CMainMenu.cpp
GeorgeK1ng: Mumps fallback added. Will remove it in MainMenu PR for 1.7

GeorgeK1ng:
Campaign buttons validation example when Armageddon's Blade h3c files are missing in LOD.

Before
image

After
image

And log

by003:
@GeorgeK1ng 
Last time I try to make vcmp format HotA campaigns, althought they are still not finished, but I tried to make a button to save them. All they are vcmp, and it is blocked by 1.6.4, but it can show in 1.6.3.
Now I found the problem, if the campaign set only have all vcmp campaigns, it cannot be read for an independent button(like'Horn of the Abyss' or 'WOG')
However, if I put some old format h3c campaigns with them, it can be read.
For example, there are seven postitions, 
If: 1. vcmp， 2. h3c, 3. h3c, 4. vcmp, it can be read.
If 1. vcmp, 2. vcmp, 3. vcmp, 4. vcmp, it cannot be read.

GeorgeK1ng:
Yeah, that check is for button is based on h3c extension, will check it to keep both formats.

GeorgeK1ng:
Should be fixed by https://github.com/vcmi/vcmi/pull/5361

PR #5227, opened 2025-01-08
CMake improvements

kambala-decapitator:


AI/Nullkiller/CMakeLists.txt
```diff
@@ -157,7 +157,7 @@ else()
 endif()
 
 target_include_directories(Nullkiller PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-target_link_libraries(Nullkiller PUBLIC vcmi fuzzylite::fuzzylite)
+target_link_libraries(Nullkiller PUBLIC vcmi fuzzylite::fuzzylite TBB::tbb)
```
kambala-decapitator, line 160: @IvanSavenko so you're saying that this is wrong? But doesn't NKAI use tbb?

PR #5229, opened 2025-01-08
[1.6.3] AI performance-related changes 

IvanSavenko:
- AI now operates in non-graph mode - since with all previous optimizations, this mode is now either on par with graph mode or slightly faster
- Tile owners data (which town is nearest to a tile) is now updated only on tile reveal and not on every turn
- Reduced range for calculation of enemy heroes threats since information on whether enemy hero can reach this tile in 10 turns is generally irrelevant (this will cover entire map even on XL), and AI generally only wants to know threats in 2 turns range

With these changes, pathfinder now accounts for 28% percents of AI turn, down from 36%


dydzio0614:
I will ask @Xilmi  if he knows any potential side effects that last change might bring

Xilmi:
"AI generally only wants to know threats in 2 turns range"
That used to be the case without some of my recent changes on my fork.
But I wouldn't insist on it as I'm not even sure of how helpful the changes are or if they are even working as intended. (maybe there's another place where everything > 2 turns is filtered out anyways.

IvanSavenko:
They are not filtered out, but problem is, in 10 turns enemy heroes can be everywhere on map (15000-20000 movement points -> 150+ ~tiles distance). Even for 5 turns (7500-1000 points -> 75+ tiles radius) this prediction is very weak. 

So for now I've set it to 5 turns, with lower values for easy difficulties.

Also, AI currently iteratively evaluates every enemy hero one by one. On maps with large number of players and large number of heroes this part is very slow. In theory we could run pathfinder for all enemy heroes at once in parallel, but not familiar with this code enough to implement it in reasonable time.

dydzio0614:
@Xilmi I will keep updating "AI - disable speedups" mod to match "certified" AI behavior from 1.6.0 so you will be able to easily toggle changes and see if there is any notable difference

PR #5230, opened 2025-01-08
Vietnamese translation update by Hunter8x from Discord

dydzio0614:
Not sure why diff is so large, but content is OK


GeorgeK1ng:
Change line endings to Unix (LF), then it will be smaller diff with real changes.

PR #5231, opened 2025-01-08
Replace random hero info in scenario info window with generated hero data

dydzio0614:
Random heroes in scenario info get properly replaced with their identities after map starts
image


DjWarmonger:
There was a bug that allows player to change his starting hero during multiplayer game load. Does this PR affect it?

dydzio0614:
i have no idea. Saves made before this pull request have starting random heroes marked as "random" which is wrong, only towns are properly unrandomized

Laserlicht:
Randomization at scenario restart still works correctly?

dydzio0614:
yep

PR #5245, opened 2025-01-09
[1.7] More deterministic AI

IvanSavenko:
Several changes to make AI more deterministic. These changes are not complete - AI can still diverge over time, but should be a bit more stable than before.

Changes so far:
- Added fake random generator that produces deterministic output for cases where AI needs fake objects
- Added custom comparator function to std::map's that use CGHeroInstance pointer as key
- Replaced bucket index selection logic - AI now uses object position on map instead of object address in memory

TODO:
- investigate remaining sources of non-determinism (BattleAI?)
- decide whether to include this in 1.6 or in 1.7

### DIFF
--- a/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
+++ b/AI/Nullkiller/Analyzers/DangerHitMapAnalyzer.cpp
@@ -79,7 +79,7 @@ void DangerHitMapAnalyzer::updateHitMap()
 	enemyHeroAccessibleObjects.clear();
 	townThreats.clear();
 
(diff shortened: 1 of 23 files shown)


thehonestcto:
"but should be a bit more stable than before."
-> Just curious, why would this be an improvement and not a minus? To my mind, the more deterministic, the more boring it is.

dydzio0614:
1. this does not have too much to do with player point of view - inability to reproduce ai bugs from saves is annoying
2. randomness based on sorting order will be minor improvement for players at best, and IMO it reveals flaw where ai is unable to decide on absolutely best action, but it will pick what is "relatively best" based on some random factors.
3. i am sure better kinds of "intended non deterministic behavior" can be coded as potential feature

PR #5259, opened 2025-01-12
Translation update

IvanSavenko:
- backported Swedish translations update to beta
- regenerated map editor translations (launcher is up-to-date)

### DIFF
--- a/Mods/vcmi/Content/config/swedish.json
+++ b/Mods/vcmi/Content/config/swedish.json
@@ -21,7 +21,7 @@
 	"vcmi.adventureMap.confirmRestartGame"               : "Är du säker på att du vill starta om spelet?",
 	"vcmi.adventureMap.noTownWithMarket"                 : "Det finns inga tillgängliga marknadsplatser!",
 	"vcmi.adventureMap.noTownWithTavern"                 : "Det finns inga tillgängliga städer med värdshus!",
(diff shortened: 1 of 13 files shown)


XCOM-HUB:
> backported Swedish translations update to beta

What does this mean? I'm not savvy enough to understand this lingo, sorry for that.

PS: I have one question about a string in the TS file, and that is "Fullscreen", shouldn't it say "Display Mode"?
Full Screen - Display Mode

IvanSavenko:
> What does this mean? I'm not savvy enough to understand this lingo, sorry for that.

Currently we have two "branches" in active use on Github:
- `develop` - anything that is added to this branch is assumed to be for next major release (so 1.7.0 at the moment)
- `beta` - anything that is added to this branch will go to next hotfix (so 1.6.4)

Since your translation updates were made for `develop` branch, they would not have been added to 1.6.4 release. So I have moved (backported) your changes from future 1.7.0 release to earlier 1.6.4 release. 

Not sure how familiar with Git you are, but you can submit changes directly to `beta` branch:
- switch to `beta` branch (instead of default `develop`)
- make your changes, as usual
- open new pull request, but change targeted branch from `develop` to `beta`. 
 
If you did everything correctly, Github should show a preview of changes, which should only consist from your commits.


> PS: I have one question about a string in the TS file, and that is "Fullscreen", shouldn't it say "Display Mode"?

Good point. Probably leftover from times when this was a simple on/off switch

PR #5266, opened 2025-01-13
Minor fixes to BattleHexArray. BattleHex'es as const reference where possible.

MichalZr6:
These are some final changes to this PR:
- #5044

I managed to change code and run test battle (mentioned in the PR above) in a loop. Every time it finished the results were reset to initial state and quick-battle was re-run. That gave more reliable values of average elapsed times that each battle took.

Here are some results:
_(RelWithDebInfo (/O2 /Ob1)_
- before BattleHexArray (beta branch from 09.01.2025 - 1.6.2):

obraz
- with BattleHexArray **and Ivan's optimisations** (1.6.3):

obraz
- with this PR changes:

obraz

One of this PR's commit is a fix for the above issue with stack ammo overuse.

There is also small TODO: code from `namespace SRSLPraserHelpers` in `BattleSpellMechanics.cpp` can be refactored to use BattleHex and BattleHexArray classes.


lib/battle/CUnitState.cpp
```diff
@@ -44,7 +44,7 @@ int32_t CAmmo::available() const
 
 bool CAmmo::canUse(int32_t amount) const
 {
-	return (available() - amount >= 0) || !isLimited();
+	return (available() - amount > 0) || !isLimited();
```
IvanSavenko, line 47: Are you sure?
If we have 2 shots, and want to use 2 of them, `>=` is correct

lib/battle/BattleHex.h
```diff
@@ -240,25 +249,20 @@ class DLL_LINKAGE BattleHex
...
 
-	// Postfix increment
-	BattleHex operator++(int) noexcept
-	{
-		BattleHex temp = *this;
-		++hex;
-		return temp;
-	}
+	// Postfix increment is deleted; use prefix increment as it has better performance
+	BattleHex operator++(int) = delete;
```
IvanSavenko, line 253: Not against it, but if this change makes any difference - I have a very serious question to compiler developers.

lib/CStack.cpp
```diff
@@ -167,7 +167,7 @@ std::string CStack::nodeName() const
 	oss << owner.toString();
 	oss << " battle stack [" << ID << "]: " << getCount() << " of ";
 	if(typeID.hasValue())
-		oss << typeID.toEntity(VLC)->getNamePluralTextID();
+		oss << typeID.toCreature()->getNamePluralTranslated();
```
IvanSavenko, line 170: Access to translated name is relatively slow, and some of these methods are used A LOT. Json key should be faster and still be readable (e.g. `core:archer`)
```suggestion
		oss << typeID.toEntity(VLC).getJsonKey();
```

client/battle/BattleOverlayLogVisualizer.h
IvanSavenko, line 1: Entire file changed? Line endings?

lib/battle/BattleHexArray.h
```diff
@@ -180,23 +205,19 @@ class DLL_LINKAGE BattleHexArray
...
-	static const BattleHexArray & getNeighbouringTilesDoubleWide(BattleHex hex, BattleSide side) noexcept
+	static const BattleHexArray & getNeighbouringTilesDoubleWide(const BattleHex & hex, BattleSide side) noexcept
 	{
 		assert(hex.isValid() && (side == BattleSide::ATTACKER || side == BattleSide::DEFENDER));
 
 		return neighbouringTilesDoubleWide.at(side)[hex.toInt()];
 	}
 
-	[[nodiscard]] inline bool contains(BattleHex hex) const noexcept
+	/// note: returns true when param is ivalid BattleHex
```
IvanSavenko, line 215: Any reason for this btw? false might be more logical for such cases, but hard to say what can be affected by such change...

lib/battle/BattleHex.h
MichalZr6: There are 2 or 3 places where this is used. From what I measured battles with prefix increment were faster by about 5% vs postfix increment. Maybe it is because postfix has to make another copy and an additional cast ?

client/battle/BattleOverlayLogVisualizer.h
MichalZr6: Yeah, one or two files I did fix but didn't catch that one. I have to check VS settings for those silent CRLF changes...

lib/battle/BattleHexArray.h
MichalZr6: I tried false (that was at the beginning of implementing BattleHexArray) and also tried some different function `isNotValidForInsertion`. I think this is the fastest way and works correctly with the rest of code. In this PR I just wanted to make a note of using it with invalid BattleHexes as it really isn't very... straightforward ;)

lib/CStack.cpp
MichalZr6: Ok, it's just... those were super long. I will check more on that

lib/battle/CUnitState.cpp
MichalZr6: Made that fix assumption too thoughtlessly and rashly :) I've found what was probably wrong, however. Marksman with his double attack was not updated if its ammo stack went empty.

lib/CStack.cpp
MichalZr6: Done as suggested. Also to one more BattleAI logging place.

MichalZr6:
Nope, apparently that small change didn't satisfy CI. I'm pretty sure it's not dangling reference as getHexes returns statics but I don't know how to make workaround for this.

IvanSavenko:
Sounds like false-positive actually.

Also, don't see any need to use static `Unit::getHexes( ... )` method in most places - regular `unit->getHexes()` would suffice:

```diff
diff --git a/lib/battle/CBattleInfoCallback.cpp b/lib/battle/CBattleInfoCallback.cpp
index be47abe5a..f76edca19 100644
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -1024,7 +1024,7 @@ AccessibilityInfo CBattleInfoCallback::getAccessibility() const
 
 AccessibilityInfo CBattleInfoCallback::getAccessibility(const battle::Unit * stack) const
 {
-	return getAccessibility(battle::Unit::getHexes(stack->getPosition(), stack->doubleWide(), stack->unitSide()));
+	return getAccessibility(stack->getHexes());
 }
 
 AccessibilityInfo CBattleInfoCallback::getAccessibility(const BattleHexArray & accessibleHexes) const
@@ -1458,7 +1458,7 @@ battle::Units CBattleInfoCallback::getAttackedBattleUnits(
 		if (unit->isGhost() || !unit->alive())
 			return false;
 
-		const auto & hexes = battle::Unit::getHexes(unit->getPosition(), unit->doubleWide(), unit->unitSide());
+		const auto & hexes = unit->getHexes();
 		for (const BattleHex & hex : hexes)
 		{
 			if (at.hostileCreaturePositions.contains(hex))
diff --git a/lib/battle/Unit.cpp b/lib/battle/Unit.cpp
index 0317fe282..3395f7634 100644
--- a/lib/battle/Unit.cpp
+++ b/lib/battle/Unit.cpp
@@ -68,19 +68,13 @@ const BattleHexArray & Unit::getSurroundingHexes(const BattleHex & position, boo
 
 BattleHexArray Unit::getAttackableHexes(const Unit * attacker) const
 {
-	const BattleHexArray & defenderHexes = battle::Unit::getHexes(
-		getPosition(),
-		doubleWide(),
-		unitSide());
+	const BattleHexArray & defenderHexes = getHexes();
 	
 	BattleHexArray targetableHexes;
 
 	for(const auto & defenderHex : defenderHexes)
 	{
-		auto hexes = battle::Unit::getHexes(
-			defenderHex,
-			attacker->doubleWide(),
-			unitSide());
+		auto hexes = attacker->getHexes(defenderHex);
 
 		if(hexes.size() == 2 && BattleHex::getDistance(hexes.front(), hexes.back()) != 1)
 			hexes.pop_back();
diff --git a/lib/spells/effects/DemonSummon.cpp b/lib/spells/effects/DemonSummon.cpp
index 60d728f34..a2d974e70 100644
--- a/lib/spells/effects/DemonSummon.cpp
+++ b/lib/spells/effects/DemonSummon.cpp
@@ -107,7 +107,7 @@ bool DemonSummon::isValidTarget(const Mechanics * m, const battle::Unit * unit)
 		return false;
 
 	//check if alive unit blocks rising
-	const auto & hexes = battle::Unit::getHexes(unit->getPosition(), unit->doubleWide(), unit->unitSide());
+	const auto & hexes = unit->getHexes();
 	for(const BattleHex & hex : hexes)
 	{
 		auto blocking = m->battle()->battleGetUnitsIf([& hex, unit](const battle::Unit * other)
diff --git a/lib/spells/effects/Heal.cpp b/lib/spells/effects/Heal.cpp
index f09eae04e..a9c9a965a 100644
--- a/lib/spells/effects/Heal.cpp
+++ b/lib/spells/effects/Heal.cpp
@@ -70,7 +70,7 @@ bool Heal::isValidTarget(const Mechanics * m, const battle::Unit * unit) const
 	if(unit->isDead())
 	{
 		//check if alive unit blocks resurrection
-		const auto & hexes = battle::Unit::getHexes(unit->getPosition(), unit->doubleWide(), unit->unitSide());
+		const auto & hexes = unit->getHexes();
 		for(const BattleHex & hex : hexes)
 		{
 			auto blocking = m->battle()->battleGetUnitsIf([& hex, unit](const battle::Unit * other)
diff --git a/lib/spells/effects/UnitEffect.cpp b/lib/spells/effects/UnitEffect.cpp
index 47620ed0a..a87729aad 100644
--- a/lib/spells/effects/UnitEffect.cpp
+++ b/lib/spells/effects/UnitEffect.cpp
@@ -202,7 +202,7 @@ EffectTarget UnitEffect::transformTargetByChain(const Mechanics * m, const Targe
 
 	for(const auto *unit : possibleTargets)
 	{
-		possibleHexes.insert(battle::Unit::getHexes(unit->getPosition(), unit->doubleWide(), unit->unitSide()));
+		possibleHexes.insert(unit->getHexes());
 	}
 
 	BattleHex destHex = mainDestination.hexValue;
@@ -221,7 +221,7 @@ EffectTarget UnitEffect::transformTargetByChain(const Mechanics * m, const Targe
 		else
 			effectTarget.emplace_back();
 
-		const auto & hexes = battle::Unit::getHexes(unit->getPosition(), unit->doubleWide(), unit->unitSide());
+		const auto & hexes = unit->getHexes();
 		for(const auto & hex : hexes)
 			if (possibleHexes.contains(hex))
 				possibleHexes.erase(hex);

```

lib/battle/CUnitState.cpp
MichalZr6: I drop those changes and deal with it in a separate PR.

PR #5292, opened 2025-01-18
Per-hero artifact scoring for NKAI

IvanSavenko:
AI now has better (but not perfect) system for scoring artifacts.

All artifacts now have their base score, based on bonuses provided by the artifact. For composite artifacts, AI will also consider bonuses from its components. When exploring, AI will now weight artifacts as min(base score, price / 5), meaning even artifacts that AI can't score will still have some priority, if only to sell them or deny them to enemy

AI will now also consider what to equip when trading with heroes or on the end of AI pass. When considering what to equip, hero will use base score of an artifact, adjusted by how relevant particular bonus / artifact for a hero. Some examples:
- Morale-bonusing artifacts have their score reduced based on percentage of undeads in the army, all the way to 0 for undead-only troops
- Artifacts that give spells (books, hat) will have score reduced based on how many of these spells are already known by hero
- Land movement artifacts have zero score while on water and vice versa
- Necromancy artifacts will have zero score if hero does not have necromancy
- Archery artifacts are scaled by amount of ranged troops, and only used if hero has archery skill

AI may still equip 'bad' artifact, however this should only happen if hero has no other artifacts to equip into said slot.

---

TODO's for future PR's (although not sure if / when I will work on these):
- avoid equipping duplicates of the same artifact. Most notably with scrolls, but may happen with other misc artifacts or rings.

- consideration for various spell-immunity neclaces

- consideration for birectional artifacts, like Shackles of War, or some Spheres - since these artifacts need consideration on what our expected enemy is to correctly estimate whether to equip it

- equipping artifacts based on immediate need - for example, equipping recovery artifacts before end of day, equipping legion pieces only in town on day 7, preparing for strong / weak enemies (or even preparing for specific enemy in advance)

- transferring resource-generating artifacts and such to scout heroes, allowing main hero to focus on combat

- ensure that AI can equip combined artifacts even if non-primary slots are in use - by considering whether score of combined artifact is higher than score of all used slots

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -1092,19 +1092,24 @@ void AIGateway::pickBestArtifacts(const CGHeroInstance * h, const CGHeroInstance
 				}
 				if(!emptySlotFound) //try to put that atifact in already occupied slot
 				{
+					int64_t artifactScore = getArtifactScoreForHero(target, artifact);
+
 					for(auto slot : artifact->getType()->getPossibleSlots().at(target->bearerType()))
 					{
 						auto otherSlot = target->getSlot(slot);
 						if(otherSlot && otherSlot->artifact) //we need to exchange artifact for better one
 						{
+							int64_t otherArtifactScore = getArtifactScoreForHero(target, otherSlot->artifact);
+							logAi->trace( "Comparing artifacts of %s: %s vs %s. Score: %d vs %d", target->getHeroTypeName(), artifact->getType()->getJsonKey(), otherSlot->artifact->getType()->getJsonKey(), artifactScore, otherArtifactScore);
+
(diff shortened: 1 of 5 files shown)


IvanSavenko:
Current base score of H3 artifacts:
https://docs.google.com/spreadsheets/d/1WRjZ8VvNico3aw41qvVnAIxw9Pqx3NLiHY4QFBiNifQ/edit?gid=0#gid=0

"AI cost" in this table is base weight that is used for picking up artifacts.
For equpping, this score is adjusted. For example, spellbinder hat has 30k score, but this score is reduced for every level 5 spell known by hero, all the way to 0 if hero knowns all level 5 spells. Deadman boots have 6k score, however that score is multiplied by 2 for heroes with necromancy, and set to 0 for heroes without one.

PR #5293, opened 2025-01-18
Fix loading vcmp campaign with specific hero used in scenario bonuses

godric3:
`heroIdentifier` "can be specified as explicit hero name and as one of keywords: strongest, generated", but `heroSpecialMap` contains "strongest", "generated" and "random", so `at()` was throwing when explicit hero name was used


Laserlicht:
Have you also checked the opposite direction (writing json)? `writeScenarioTravelFromJson`

godric3:
I converted some custom sod campaign using map editor converter so I guess 🤔 
anyway it uses `vstd::contains(vstd::reverseMap(heroSpecialMap), bonus.info1))` so should be safe

PR #5327, opened 2025-01-25
GameSettings unit test

dydzio0614:
I expected to make bigger unit test PR but found some issues that notably slowed me down and decreased temporarily my interest in pursuing more test coverage, so i guess better this than nothing

### DIFF
--- a/test/testdata/gameSettingsOverride.json
+++ b/test/testdata/gameSettingsOverride.json
@@ -1,0 +1,6 @@
+{
+	"creatures" :
+	{
+		"weeklyGrowthPercent": 50
+	}
+}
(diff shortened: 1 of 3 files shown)


test/testdata/gameSettingsOverride.json
```diff
@@ -0,0 +1,6 @@
+{
+	"creatures" :
+	{
+		"weeklyGrowthPercent": 50
+	}
+}
```
IvanSavenko, line 6: You can create same json in runtime if you wish to:
```cpp
JsonNode node;
node["creatures"]["weeklyGrowthPercent"] = JsonNode(50);
```

PR #5331, opened 2025-01-26
Fix Castle Overlapping

Olegmods:
Fixes a small graphic bug.


dydzio0614:
i would target beta rather than develop

Olegmods:
> i would target beta rather than develop

My bad. Will close that one and reopen for beta.

PR #5337, opened 2025-01-27
[1.6?] Explicit toggle for ffmpeg video player compilation

IvanSavenko:
Right now our ffmpeg-based video player is optional (which is not a bad thing), however its state depends on presence of ffmpeg. And if CMake fails to find ffmpeg - video support will be silently disabled, with just 1 line in CMake log which is easy to miss.

This PR adds explicit `ENABLE_VIDEO` toggle to CMake, which is set to on by default. If toggle is set to on, then ffmpeg is required for build, and if set to off, then video player will be always disabled and CMake won't search for ffmpeg.
Also:
- flipped our preprocessor macro from `DISABLE_VIDEO` to `ENABLE_VIDEO` - to be in line with CMake toggle, and in line with other similar defines, e.g. `ENABLE_EDITOR`
- added logging to make it easier to determine reason for missing video in the future


clientapp/CMakeLists.txt
```diff
@@ -42,6 +42,11 @@ target_include_directories(vcmiclient
 	PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
+if(ENABLE_VIDEO)
+	target_compile_definitions(vcmiclient PRIVATE ENABLE_VIDEO)
+endif()
```
kambala-decapitator, line 47: no longer needed in this file I guess?

IvanSavenko: Still needed. See changes in `EntryPoint.cpp` - used to construct video handler if ffmpeg support is on.

kambala-decapitator: I wonder why it was only inside WIN32 then...

kambala-decapitator (approved):

PR #5350, opened 2025-01-30
Fix creating buttons in Stack Window when upgrade is unavailable

MichalZr6:
Clicking these buttons caused a crash (reported on Discord).

This PR fixes the issue and includes minor changes to the UpgradeInfo class.

### DIFF
--- a/client/windows/CCreatureWindow.cpp
+++ b/client/windows/CCreatureWindow.cpp
@@ -769,9 +769,13 @@ CStackWindow::CStackWindow(const CStackInstance * stack, std::function<void()> d
 	info->creature = stack->getCreature();
 	info->creatureCount = stack->count;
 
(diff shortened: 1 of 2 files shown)


dydzio0614:
My personal opinion is that readability after these performance optimizations is questionable for me (if whole code looked like that i would have issues with navigating back and forth)

IvanSavenko:
@dydzio0614 are you talking about UpgradesVector / UpgradesCostsVector defines?

dydzio0614:
yep

MichalZr6:
Removed those aliases.

IvanSavenko:
While I am not against it, I don't see much point in using boost containers in places like this.

Not a bad change, sure, but it will make code a bit more complex to understand in return for... what? 0.1 ms speedup of dialog that takes 1 ms to show up?

MichalZr6:
Tiny optimization it is ;) These vectors never grow big and I rather thought about AI that uses this code, not dialog speedup.

PR #5355, opened 2025-01-31
Hotfix for 1.6.4

IvanSavenko:
Apparently this is legal code:
```cpp
presenceFlags.set(hex.toInt()) = true;
```
... and will result in bitset that has only 1st bit set to true due to multiple implicit conversions. Thank you, ISO.

And thanks to Manfred for noticing it.


MichalZr6:
> Apparently this is legal code

Wow, I thought it is a no-op. Good catch, then! 🙃

IvanSavenko:
That was my first thought too.
But no-op would have likely triggered compiler error or at least warning.

And apparently, that method returns bitset reference for some reason...

PR #5385, opened 2025-02-07
[1.6.6] Clicking on blocked tile of a visitable object now builds a path to it

IvanSavenko:
Suggestion from Discord.

If player has hero selected, clicking on a blocked tile of a visitable, reachable object will now build a path to its visitable position (or move hero to it, in case of second click / tap).

Objects that have interaction on left click (allied town and shipyards) are excluded from this change and work as before


dydzio0614:
I would probably make it default only on mobiles for now

PR #5388, opened 2025-02-07
Separate icons for artifact sorting screen

IvanSavenko:
Artifact sorting icons from backpack screen now use separate set of icons instead of reusing images from altar of sacrifice.

Also makes it possible to provide different icons in mod without breaking altar. Can't say that I am extremely happy with these versions, so better icons are welcomed - either in mods, or in case of small icons with clear copyrights - in vcmi itself

зображення


SoundSSGood:
What does the down arrow mean?

IvanSavenko:
Replaced with text description for clarity, as suggested on Discord:
зображення

Issue #5330, opened 2025-01-26
If having First Aid Skill, make it possible to auto-skip First Aid Tent if all units have full health


XCOM-HUB:
This will make the battles a little quicker when you don't have to manually skip the turn of the First Aid Tent when all of your units already have full HP.

IvanSavenko:
First aid tent turn is not actually useless in such scenario.

You can use it to cast a spell or retreat from combat for example

XCOM-HUB:
> First aid tent turn is not actually useless in such scenario.
> 
> You can use it to cast a spell or retreat from combat for example

Didn't think of that. Thanks for clarifying that, and sorry for the late response.

PR #5399, opened 2025-02-08
[1.6.6] Combat Health Bar & Calculation fix

### LINKED ISSUE #5396: Combat Health Bar
**Describe your proposal**
Another cool feature intruduced in ERA. 
https://youtu.be/7Tk1XXDPdWE?si=m2UKgJDMqcIeQgGq

What about pressing ALT during battle show stack healt status for quick viewing?

Laserlicht:
Shows health bar (indicating remaining health until one unit dies).

Currently permanently enabled. Is it better to use alt to show (but then this feature isn't accessible on non PC)?

grafik

fixes #5396


Olegmods:
> Shows health bar (indicating remaining health until one unit dies).
> 
> Currently permanently enabled. Is it better to use alt to show (but then this feature isn't accessible on non PC)?

Well, since that's an idea from ERA, I can say that they added a button in in-game settings that leads to a HP bar configuration menu IIRC.

IvanSavenko:
Yeah, might be better as off by default, with toggle somewhere in config.

Sure, might be useful for dragons, but for creatures like pixies this is only confusing, since health of single unit is irrelevant compared to stack size.

And might be confusing since it is not clear whether this is single unit hp or entire stack size

Olegmods:
Also, just an as idea to consider. Maybe making some sort of gradient as color can look better.

Olegmods:
> And might be confusing since it is not clear whether this is single unit hp or entire stack size

Should be one unit HP I guess.

IvanSavenko:
> Should be one unit HP I guess.

And how is this indicated to player?

Alucard648:
> > Should be one unit HP I guess.
> 
> And how is this indicated to player?

Mentioned in tooltip for option flag to turn on/off this setting.

XCOM-HUB:
This is great! This could be a good opportunity also to implement https://github.com/vcmi/vcmi/issues/5330

PR #5417, opened 2025-02-12
[1.7] Support for loading HotA 1.7.X maps and campaigns

IvanSavenko:
Potentially useful reference with partially decode 1.7.2 h3m format: https://github.com/Andarch/H3_HotA_Map_Editor_X

Progress:
- [x] parsing of 1.7.0 h3m's
- [ ] parsing of 1.7.0 h3c's
- [ ] parsing of 1.7.1 h3m's
- [ ] parsing of 1.7.1 h3c's
- [ ] parsing of 1.7.2 h3m's
- [ ] parsing of 1.7.2 h3c's
- [ ] identify purpose of all new values
- [ ] add bounds checking for all new values
- [ ] load new properties where corresponding vcmi functionality already exists
- [ ] remove debug code

Not part of this PR:
- mapping of objects/def's in HotA mod
- implementation of new h3m functionality


IvanSavenko:
First part done - all standard hota maps can be parsed, and all missing functionality should be detected & logged.

However if there are user-made maps / campaigns that use functionality that was not used by hota - they would need additional adjustments.

Next steps:
- Add format mapping to hota mod. If somebody from modders would look into it - that would be great. In theory this is all that is needed to make 1.7 maps work in vcmi after this PR.

- take a closer look at campaigns. Hota finally gave their campaigns distinct version marker, so we can actually distinguish them from sod, and with this PR vcmi can load h3m's from 1.7 campaigns and parse entire header structure, but it still needs some work for campaigns to actually be playable in game. Will probably do this a bit later.
  -  it looks like campaign regions layout (e.g. background map / region segments) is still hardcoded in hota - we'll need to add a way to modify it via mods or add it to vcmi configs
  - campaign media (music / voiceover) is also hardcoded, and extends hardcoded h3 lists. Needs extension on our side.
  - hota campaigns need to go through our ID remapper, like hota maps do - so hota ID's to creatures / artifacts / etc are properly converted to vcmi mod ID's
  - looks like they added support for different prologs / epilogs for a scenario. Probably depends on selected starting bonus? Need handling on vcmi side.

- Start looking into newly added h3m functionality. Looks like quite a lot this time. Rather low in my priorities, so free for taking:
  - List of allowed difficulties for a map
  - Option to block hiring of defeated heroes (well, technically to unblock it, since it is what hota has as default)
  - Option to ban creation of specific combined artifacts
  - Per-player option to ban hero recruitment
  - Option to prevent hero from auto-gaining skills if he starts with level above 1
  - Option to block hero from gaining experience
  - Option to set hero level to specific value (yes, including level 20 000 heroes, for whatever reason)
  - Support for receiving scroll in event/pandora, seer hut (both reward and quest)
  - Support for giving movement points via pandora / event (slightly different from what is supported by vcmi)
  - Pandora / event that only appear on some map difficulties
  - Option to set size of wandering monster to be of specific AI value
  - Pickup modes for artifacts. Also needs investigation at what these modes are
  - Custom guards for Abandoned Mine
  - Starting troops and starting artifacts for hero placeholders
  - Reward configuration for a lot of H3 objects
  - Black market artifact configuration, including spell scrolls trading
  - University skills configuration
  - Game difficulty seer hut "quest" (yes, that's the thing now)
  - Support for constructing / banning special buildings per faction type (mostly for random town I guess). Both on start and in events 

- And possibly, add remaining 1.6 and below hota h3m functionality:
  - arena and mirror maps (if they need any implementation)
  - support for round limit in combat
  - configuration of wandering monsters - aggression, diplomacy join percentage, etc
  - creature banks configuration
  - multiple quests in seer hut, fixes for apparently broken repeatable quests
  - random town alignment "not as player X"

PR #5423, opened 2025-02-14
[mapeditor] Map validation results window changes

MichalZr6:
This PR introduces the following changes:
- the results window now dynamically adjusts its size based on message lengths
- a valid map will now properly display an appropriate message in the validation results
- word wrap for item text activates only when the window reaches a defined width limit

Additionally, one of the issue messages has been updated to make the error clearer.

If this PR gets accepted, I’ll also regenerate and commit the updated .ts files.

Old validation results window with multiple issues:
old1

After manual resize:
old2

Empty issue list (valid map):
old3

New validation results window with multiple issues:
new1
new2

Empty issue list (valid map):
new3

### DIFF
--- a/mapeditor/PlayerSelectionDialog.cpp
+++ b/mapeditor/PlayerSelectionDialog.cpp
@@ -9,8 +9,14 @@
  */
 
 #include "PlayerSelectionDialog.h"
-#include "mainwindow.h"
 #include "../lib/mapping/CMap.h"
+#include "mainwindow.h"
+
+#include <QCheckBox>
+#include <QDialogButtonBox>
+#include <QAction>
+#include <QLabel>
+
 
 PlayerSelectionDialog::PlayerSelectionDialog(MainWindow * mainWindow)
 	: QDialog(mainWindow), selectedPlayer(PlayerColor::NEUTRAL)
--- a/mapeditor/PlayerSelectionDialog.cpp
+++ b/mapeditor/PlayerSelectionDialog.cpp
@@ -72,24 +72,21 @@ PlayerColor PlayerSelectionDialog::getSelectedPlayer() const
 void PlayerSelectionDialog::setupDialogComponents()
 {
 	setWindowTitle(tr("Select Player"));
-	setFixedSize(sizeHint());
+	setFixedWidth(dialogWidth);
 	setWindowFlags(Qt::Dialog | Qt::MSWindowsFixedSizeDialogHint);
 	font.setPointSize(10);
 	setFont(font);
 
-	mainLayout.setSpacing(10);
-	mainLayout.setContentsMargins(20, 20, 80, 20);
-	checkboxLayout.setContentsMargins(0, 10, 0, 20);
-	mainLayout.addSpacing(4);
-
 	QLabel * errorLabel = new QLabel(tr("Hero cannot be created as NEUTRAL"), this);
 	font.setBold(true);
 	errorLabel->setFont(font);
+	errorLabel->setWordWrap(true);
 	mainLayout.addWidget(errorLabel);
 
 	QLabel * instructionLabel = new QLabel(tr("Switch to one of the available players:"), this);
 	font.setBold(false);
 	instructionLabel->setFont(font);
+	instructionLabel->setWordWrap(true);
 	mainLayout.addWidget(instructionLabel);
 
 	QWidget * checkboxContainer = new QWidget(this);
--- a/mapeditor/PlayerSelectionDialog.cpp
+++ b/mapeditor/PlayerSelectionDialog.cpp
@@ -12,26 +12,24 @@
 #include "mainwindow.h"
 #include "../lib/mapping/CMap.h"
 
-PlayerSelectionDialog::PlayerSelectionDialog(QWidget * parent)
-	: QDialog(parent), selectedPlayer(PlayerColor::NEUTRAL)
+PlayerSelectionDialog::PlayerSelectionDialog(MainWindow * mainWindow)
+	: QDialog(mainWindow), selectedPlayer(PlayerColor::NEUTRAL)
 {
-	auto main = qobject_cast<MainWindow *>(parent);
-	assert(main);
+	assert(mainWindow);
 
 	setupDialogComponents();
 
 	int maxPlayers = 0;
-	if(main && main->controller.map())
-		maxPlayers = main->controller.map()->players.size();
-	
-	bool defaultIsChecked = false;
+	if(mainWindow && mainWindow->controller.map())
+		maxPlayers = mainWindow->controller.map()->players.size();
 
 	for(int i = 0; i < maxPlayers; ++i)
 	{
 		PlayerColor player(i);
-		QAction * action = main->getActionPlayer(player);
+		QAction * action = mainWindow->getActionPlayer(player);
+		bool isEnabled = mainWindow->controller.map()->players.at(i).canAnyonePlay();
 
-		addCheckbox(action, player, main->controller.map()->players.at(i).canAnyonePlay(), & defaultIsChecked);
+		addCheckbox(action, player, isEnabled);
 	}
 }
 
@@ -53,9 +51,8 @@ void PlayerSelectionDialog::onCheckboxToggled(bool checked)
 		}
 		else
 		{
-			cb->blockSignals(true);
+			QSignalBlocker blocker(cb);
 			cb->setChecked(false);
-			cb->blockSignals(false);
 		}
 	}
 }
@@ -101,7 +98,7 @@ void PlayerSelectionDialog::setupDialogComponents()
 	setLayout(& mainLayout);
 }
 
-void PlayerSelectionDialog::addCheckbox(QAction * checkboxAction, PlayerColor player, bool isEnabled, bool * isDefault)
+void PlayerSelectionDialog::addCheckbox(QAction * checkboxAction, PlayerColor player, bool isEnabled)
 {
 	QHBoxLayout * rowLayout = new QHBoxLayout();
 	auto * checkbox = new QCheckBox(checkboxAction->text(), this);
@@ -122,16 +119,17 @@ void PlayerSelectionDialog::addCheckbox(QAction * checkboxAction, PlayerColor pl
 	rowLayout->addWidget(shortcutLabel, 1);
 
 	checkbox->setEnabled(isEnabled);
-	if(checkbox->isEnabled() && !*isDefault)
+
+	if(isEnabled && !defaultCheckedSet)
 	{
 		checkbox->setChecked(true);
 		selectedPlayer = player;
-		*isDefault = true;
+		defaultCheckedSet = true;
 	}
 
 	checkboxLayout.addLayout(rowLayout);
 
-	connect(checkbox, &QCheckBox::clicked, this, [=]()
+	connect(checkbox, &QCheckBox::clicked, this, [this, checkbox, player]()
 		{
 			selectedPlayer = player;
 
@@ -148,7 +146,7 @@ void PlayerSelectionDialog::addCheckbox(QAction * checkboxAction, PlayerColor pl
 	addAction(checkboxAction);
 
 	// Connect action trigger to simulate checkbox click
-	connect(checkboxAction, &QAction::triggered, this, [=]()
+	connect(checkboxAction, &QAction::triggered, this, [checkbox]()
 		{
 			if(checkbox->isEnabled())
 				checkbox->click();
--- a/mapeditor/PlayerSelectionDialog.cpp
+++ b/mapeditor/PlayerSelectionDialog.cpp
@@ -12,7 +12,8 @@
 #include "../lib/mapping/CMap.h"
 #include "mainwindow.h"
 
-#include <QCheckBox>
+#include <QRadioButton>
+#include <QButtonGroup>
 #include <QDialogButtonBox>
 #include <QAction>
 #include <QLabel>
@@ -32,37 +33,9 @@ PlayerSelectionDialog::PlayerSelectionDialog(MainWindow * mainWindow)
 	for(int i = 0; i < maxPlayers; ++i)
 	{
 		PlayerColor player(i);
-		QAction * action = mainWindow->getActionPlayer(player);
-		bool isEnabled = mainWindow->controller.map()->players.at(i).canAnyonePlay();
-
-		addCheckbox(action, player, isEnabled);
+		addRadioButton(mainWindow->getActionPlayer(player), player);
 	}
 }
-
-void PlayerSelectionDialog::onCheckboxToggled(bool checked)
-{
-	if(!checked)
-		return;
-
-	QCheckBox * senderCheckBox = qobject_cast<QCheckBox *>(sender());
-	if(!senderCheckBox)
-		return;
-
-	for(int i = 0; i < checkboxes.size(); ++i)
-	{
-		QCheckBox * cb = checkboxes[i];
-		if(cb == senderCheckBox)
-		{
-			selectedPlayer = PlayerColor(i);
-		}
-		else
-		{
-			QSignalBlocker blocker(cb);
-			cb->setChecked(false);
-		}
-	}
-}
-
 
 PlayerColor PlayerSelectionDialog::getSelectedPlayer() const
 {
@@ -77,6 +50,9 @@ void PlayerSelectionDialog::setupDialogComponents()
 	font.setPointSize(10);
 	setFont(font);
 
+	buttonGroup = new QButtonGroup(this);
+	buttonGroup->setExclusive(true);
+
 	QLabel * errorLabel = new QLabel(tr("Hero cannot be created as NEUTRAL"), this);
 	font.setBold(true);
 	errorLabel->setFont(font);
@@ -89,9 +65,9 @@ void PlayerSelectionDialog::setupDialogComponents()
 	instructionLabel->setWordWrap(true);
 	mainLayout.addWidget(instructionLabel);
 
-	QWidget * checkboxContainer = new QWidget(this);
-	checkboxContainer->setLayout(& checkboxLayout);
-	mainLayout.addWidget(checkboxContainer);
+	QWidget * radioContainer = new QWidget(this);
+	radioContainer->setLayout(& radioButtonsLayout);
+	mainLayout.addWidget(radioContainer);
 
 	QDialogButtonBox * box = new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel, this);
 	connect(box, &QDialogButtonBox::accepted, this, &QDialog::accept);
@@ -101,57 +77,43 @@ void PlayerSelectionDialog::setupDialogComponents()
 	setLayout(& mainLayout);
 }
 
-void PlayerSelectionDialog::addCheckbox(QAction * checkboxAction, PlayerColor player, bool isEnabled)
+void PlayerSelectionDialog::addRadioButton(QAction * action, PlayerColor player)
 {
-	QHBoxLayout * rowLayout = new QHBoxLayout();
-	auto * checkbox = new QCheckBox(checkboxAction->text(), this);
-	
-	QLabel * shortcutLabel = new QLabel(checkboxAction->shortcut().toString(), this);
+	auto * radioButton = new QRadioButton(action->text(), this);
+	radioButton->setEnabled(action->isEnabled());
+	// Select first available player by default
+	if(buttonGroup->buttons().isEmpty() && radioButton->isEnabled())
+	{
+		radioButton->setChecked(true);
+		selectedPlayer = player;
+	}
+
+	buttonGroup->addButton(radioButton, player.getNum());
+
+	auto * shortcutLabel = new QLabel(action->shortcut().toString(), this);
 	QFont shortcutFont = font;
 	shortcutFont.setPointSize(9);
 	shortcutFont.setItalic(true);
 	shortcutLabel->setFont(shortcutFont);
 	shortcutLabel->setAlignment(Qt::AlignCenter | Qt::AlignVCenter);
+	shortcutLabel->setContentsMargins(0, 0, 6, 0); // Italic text was trimmed at the end
 
-	QWidget * checkboxContainer = new QWidget(this);
-	QHBoxLayout * cbLayout = new QHBoxLayout(checkboxContainer);
-	cbLayout->setContentsMargins(0, 0, 0, 0);
-	cbLayout->addWidget(checkbox, 0, Qt::AlignCenter);
+	auto * rowWidget = new QWidget(this);
+	auto * rowLayout = new QGridLayout(rowWidget);
+	rowLayout->setContentsMargins(0, 0, 0, 0);
+	rowLayout->addWidget(radioButton, 0, 0, Qt::AlignCenter | Qt::AlignVCenter);
+	rowLayout->addWidget(shortcutLabel, 0, 1, Qt::AlignCenter | Qt::AlignVCenter);
+	radioButtonsLayout.addWidget(rowWidget);
 
-	rowLayout->addWidget(checkboxContainer, 1);
-	rowLayout->addWidget(shortcutLabel, 1);
-
-	checkbox->setEnabled(isEnabled);
-
-	if(isEnabled && !defaultCheckedSet)
-	{
-		checkbox->setChecked(true);
-		selectedPlayer = player;
-		defaultCheckedSet = true;
-	}
-
-	checkboxLayout.addLayout(rowLayout);
-
-	connect(checkbox, &QCheckBox::clicked, this, [this, checkbox, player]()
+	connect(radioButton, &QRadioButton::clicked, this, [this, player]()
 		{
 			selectedPlayer = player;
-
-			// Radio-style logic: uncheck other boxes
-			for(auto * box : findChildren<QCheckBox *>())
-			{
-				if(box != checkbox)
-					box->setChecked(false);
-			}
 		});
 
-
-	// Add action to the dialog for shortcut support
-	addAction(checkboxAction);
-
-	// Connect action trigger to simulate checkbox click
-	connect(checkboxAction, &QAction::triggered, this, [checkbox]()
+	addAction(action);
+	connect(action, &QAction::triggered, this, [radioButton]()
 		{
-			if(checkbox->isEnabled())
-				checkbox->click();
+			if(radioButton->isEnabled())
+				radioButton->click();
 		});
 }
--- a/mapeditor/PlayerSelectionDialog.cpp
+++ b/mapeditor/PlayerSelectionDialog.cpp
@@ -1,0 +1,156 @@
+/*
+ * PlayerSelectionDialog.cpp, part of VCMI engine
+ *
+ * Authors: listed in file AUTHORS in main folder
+ *
+ * License: GNU General Public License v2.0 or later
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#include "PlayerSelectionDialog.h"
+#include "mainwindow.h"
+#include "../lib/mapping/CMap.h"
+
+PlayerSelectionDialog::PlayerSelectionDialog(QWidget * parent)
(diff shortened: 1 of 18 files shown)


GeorgeK1ng:
There is one more question. When is map randomly generated, why are not are used mods automatically added to map depencies?

Otherwise I like this detailed info about "issues".

MichalZr6:
I will take a look if I can add that. Such action should show some info-message about it to the user.

MichalZr6:
Will commit some more soon, within few days I guess.

MichalZr6:
> There is one more question. When is map randomly generated, why are not are used mods automatically added to map depencies?

Are you sure ? Random map creation currently enables all mods that could be used by available objects:
```cpp
ModCompatibilityInfo MapController::modAssessmentAll()
{
	ModCompatibilityInfo result;
	for(auto primaryID : LIBRARY->objtypeh->knownObjects())
	{
		for(auto secondaryID : LIBRARY->objtypeh->knownSubObjects(primaryID))
		{
			auto handler = LIBRARY->objtypeh->getHandlerFor(primaryID, secondaryID);
			auto modName = QString::fromStdString(handler->getJsonKey()).split(":").at(0).toStdString();
			if(modName != "core")
				result[modName] = LIBRARY->modh->getModInfo(modName).getVerificationInfo();
		}
	}
	return result;
}
```
so currently:

obraz

I can tune that so the code will use `ModCompatibilityInfo MapController::modAssessmentMap(const CMap & map)` and new randomly generated map will require only those mods which objects were placed on the map:

obraz

Also I think some mods should be filtered out of this list. I already did that to "core" (it was displayed as blank entry).

GeorgeK1ng:
Sounds logical to keep required mods only for placed objects.

Last time I generated maps in editor, validator gave me issues about missing mods and I must check them manually to get rid of that issue.

MichalZr6:
So this is the way ?
```cpp
auto modScope = handler->getModScope();
auto modName = LIBRARY->modh->getModInfo(modScope).getVerificationInfo();
```

And... any ideas on this one:
>Also I think some mods should be filtered out of this list. I already did that to "core" (it was displayed as blank entry).

?

IvanSavenko:
Yes.
"modScope" would be mod ID to write to map file as dependency.
"modName" is human-readable name to show to user

If you want to exclude internal mod ID's - add check for `ModScope::isScopeReserved`

MichalZr6:
Thanks for clarifying this! Speaking of filtering out mods I mean to exclude mods that have nothing to do with maps like ex. `VCMI extras`.

IvanSavenko:
What's wrong with vcmi extras? Does editor adds it to map dependency? Why?
"Scan all objects and add mods they come from as dependency" should not trigger for extras.

MichalZr6:
Currently code adds as checkboxes all active mods to ModSettings window:

`for(const auto & modName : LIBRARY->modh->getActiveMods())`

IvanSavenko:
Sort of makes sense I guess? As long as such mods are not auto-enabled as dependency.

I mean, portrait pack is not really needed to load a map. But if map is centered on reskinned heroes then map maker should be able to designate such mod as required.

MichalZr6:
> Sort of makes sense I guess? As long as such mods are not auto-enabled as dependency.
> 
> I mean, portrait pack is not really needed to load a map. But if map is centered on reskinned heroes then map maker should be able to designate such mod as required.

I left that as is. Map creator might want to set some mods as required even though they might be usually discarded (like music-theme mods).

I implemented two features that might be useful in those recent commits:

1. Prompt user to add mod to map's required mods list right at the moment the object is placed on the map:



2. Allow user to select proper player (color) when the default is set to NEAUTRAL and hero is being placed on the map:

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+	rowLayout->addWidget(checkboxContainer, 1);
+	rowLayout->addWidget(shortcutLabel, 1);
+
+	checkbox->setEnabled(isEnabled);
+	if(checkbox->isEnabled() && !*isDefault)
+	{
+		checkbox->setChecked(true);
+		selectedPlayer = player;
+		*isDefault = true;
+	}
```
IvanSavenko, line 130: Perhaps move this code to caller method? Instead of this sneaky modification of `isDefault` that was received from caller method?

Why this `isDefault` is needed to begin with? It should be possible to set `isEnabled` to true for 1st such call, and false for all subsequent calls

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+PlayerColor PlayerSelectionDialog::getSelectedPlayer() const
+{
+	return selectedPlayer;
+}
+
+void PlayerSelectionDialog::setupDialogComponents()
+{
+	setWindowTitle(tr("Select Player"));
+	setFixedSize(sizeHint());
+	setWindowFlags(Qt::Dialog | Qt::MSWindowsFixedSizeDialogHint);
```
kambala-decapitator, line 73: is there any serious reason to prevent dialog resizing?

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+	font.setPointSize(10);
+	setFont(font);
+
+	mainLayout.setSpacing(10);
+	mainLayout.setContentsMargins(20, 20, 80, 20);
+	checkboxLayout.setContentsMargins(0, 10, 0, 20);
```
kambala-decapitator, line 79: are there any issues with default margins?

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+#include "PlayerSelectionDialog.h"
+#include "mainwindow.h"
+#include "../lib/mapping/CMap.h"
+
+PlayerSelectionDialog::PlayerSelectionDialog(QWidget * parent)
+	: QDialog(parent), selectedPlayer(PlayerColor::NEUTRAL)
+{
+	auto main = qobject_cast<MainWindow *>(parent);
+	assert(main);
```
kambala-decapitator, line 19: just pass parent with appropriate type instead

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+		QCheckBox * cb = checkboxes[i];
+		if(cb == senderCheckBox)
+		{
+			selectedPlayer = PlayerColor(i);
+		}
+		else
+		{
+			cb->blockSignals(true);
+			cb->setChecked(false);
+			cb->blockSignals(false);
```
kambala-decapitator, line 58: use QSignalBlocker instead

kambala-decapitator, line 58: this logic sounds like you want an exclusive QButtonGroup. And maybe radiobuttons will suit better here?

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+	QWidget * checkboxContainer = new QWidget(this);
+	checkboxContainer->setLayout(& checkboxLayout);
+	mainLayout.addWidget(checkboxContainer);
+
+	QDialogButtonBox * box = new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel, this);
+	connect(box, &QDialogButtonBox::accepted, this, &QDialog::accept);
+	connect(box, &QDialogButtonBox::rejected, this, &QDialog::reject);
+	mainLayout.addWidget(box);
+
+	setLayout(& mainLayout);
```
kambala-decapitator, line 101: don't see any reason why you want to store both layouts as ivars

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+	QLabel * shortcutLabel = new QLabel(checkboxAction->shortcut().toString(), this);
+	QFont shortcutFont = font;
+	shortcutFont.setPointSize(9);
+	shortcutFont.setItalic(true);
+	shortcutLabel->setFont(shortcutFont);
+	shortcutLabel->setAlignment(Qt::AlignCenter | Qt::AlignVCenter);
+
+	QWidget * checkboxContainer = new QWidget(this);
+	QHBoxLayout * cbLayout = new QHBoxLayout(checkboxContainer);
+	cbLayout->setContentsMargins(0, 0, 0, 0);
```
kambala-decapitator, line 118: why?

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+	if(checkbox->isEnabled() && !*isDefault)
+	{
+		checkbox->setChecked(true);
+		selectedPlayer = player;
+		*isDefault = true;
+	}
+
+	checkboxLayout.addLayout(rowLayout);
+
+	connect(checkbox, &QCheckBox::clicked, this, [=]()
```
kambala-decapitator, line 134: prefer explicit capture when there're only a few variables to capture

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,156 @@
...
+		{
+			selectedPlayer = player;
+
+			// Radio-style logic: uncheck other boxes
+			for(auto * box : findChildren<QCheckBox *>())
+			{
+				if(box != checkbox)
+					box->setChecked(false);
+			}
```
kambala-decapitator, line 143: definitely QButtonGroup

mapeditor/PlayerSelectionDialog.h
```diff
@@ -0,0 +1,44 @@
...
+	PlayerColor getSelectedPlayer() const;
+
+private slots:
+	void onCheckboxToggled(bool checked);
+
+private:
+	std::vector<QCheckBox *> checkboxes;
+	PlayerColor selectedPlayer;
+
+	QFont font;
```
kambala-decapitator, line 37: don't see a need to store font either

mapeditor/PlayerSelectionDialog.h
```diff
@@ -0,0 +1,44 @@
...
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+
+#pragma once
+
+#include <QDialog>
+#include <QComboBox>
+#include <QCheckBox>
+#include <QDialogButtonBox>
```
kambala-decapitator, line 16: don't include things in a header that can be forward-declared

mapeditor/mainwindow.cpp
```diff
@@ -1147,9 +1159,20 @@ void MainWindow::on_inspectorWidget_itemChanged(QTableWidgetItem *item)
...
+
+	if(mapSettings->isVisible())
+	{
+		mapSettings->raise();
+		mapSettings->activateWindow();
+	}
+	else
+	{
+		mapSettings->setWindowModality(Qt::WindowModal);
+		mapSettings->setModal(true);
```
kambala-decapitator, line 1173: these properties should be set on widget creation

mapeditor/mapcontroller.cpp
```diff
@@ -29,12 +29,19 @@
...
 #include "maphandler.h"
 #include "mainwindow.h"
 #include "inspector/inspector.h"
 #include "GameLibrary.h"
+#include "PlayerSelectionDialog.h"
+
+MapController::MapController(QObject * parent)
+	: QObject(parent)
+{
+}
```
kambala-decapitator, line 44:
```suggestion
MapController::MapController(QObject * parent) = default;
```

mapeditor/mapcontroller.cpp
```diff
@@ -538,30 +545,85 @@ void MapController::commitObjectCreate(int level)
...
+	PlayerSelectionDialog dialog(main);
+	if(dialog.exec() == QDialog::Accepted)
 	{
-		error = QObject::tr("Hero %1 cannot be created as NEUTRAL.").arg(QString::fromStdString(newObj->instanceName));
-		return false;
+		main->switchDefaultPlayer(dialog.getSelectedPlayer());
+		return true;
 	}
 	
+	error = QObject::tr("Hero %1 cannot be created as NEUTRAL.").arg(QString::fromStdString(heroObj->instanceName));
```
kambala-decapitator, line 593:
```suggestion
	error = tr("Hero %1 cannot be created as NEUTRAL.").arg(QString::fromStdString(heroObj->instanceName));
```

mapeditor/mapcontroller.cpp
```diff
@@ -538,30 +545,85 @@ void MapController::commitObjectCreate(int level)
...
+	ModCompatibilityInfo modsInfo;
+	modAssessmentObject(obj, modsInfo);
+
+	for(auto & mod : modsInfo)
+	{
+		if(!_map->mods.count(mod.first))
+		{
+			QString submod;
+			if(!mod.second.parent.empty())
+				submod = " (" + tr("submod of") + " " + QString::fromStdString(mod.second.parent) + ")";
```
kambala-decapitator, line 608:
```suggestion
				submod = tr(" (submod of %1)").arg(QString::fromStdString(mod.second.parent));
```

mapeditor/mapcontroller.cpp
```diff
@@ -538,30 +545,85 @@ void MapController::commitObjectCreate(int level)
...
+
+			auto reply = QMessageBox::question(main,
+				tr("Missing Required Mod"),
+				tr("This object is from the mod '%1'%2.\n\n"
+					"The mod is currently not in the map's required modifications list.\n\n"
+					"Do you want to add this mod to the required modifications ?\n")
```
kambala-decapitator, line 614:
```suggestion
					"Do you want to add this mod to the required modifications?\n")
```

mapeditor/mapcontroller.cpp
```diff
@@ -538,30 +545,85 @@ void MapController::commitObjectCreate(int level)
...
+				.arg(QString::fromStdString(LIBRARY->modh->getModInfo(mod.first).getVerificationInfo().name), submod),
+				QMessageBox::Yes | QMessageBox::No);
+
+			if(reply == QMessageBox::Yes)
+				/* emit */ requestModsUpdate(modsInfo, true); // signal for MapSettings
```
kambala-decapitator, line 619: why is `emit` commented?

mapeditor/validator.cpp
```diff
@@ -179,12 +171,19 @@ std::set<Validator::Issue> Validator::validate(const CMap * map)
...
 			{
-				issues.insert({ tr("Map contains object from mod \"%1\", but doesn't require it").arg(QString::fromStdString(LIBRARY->modh->getModInfo(mod.first).getVerificationInfo().name)), true });
+				QString submod;
+				if(!mod.second.parent.empty())
+					submod = " ("+ tr("submod of") +" "+ QString::fromStdString(mod.second.parent) + ")";
+				
+				issues.insert({
+						tr("Map contains object(s) from mod '%1'%2, but the mod is missing from the map's required mods list."
+						" Add it to the map's required mods in Map->General settings.", "should be consistent with Map->General menu entry translation")
+						.arg(QString::fromStdString(LIBRARY->modh->getModInfo(mod.first).getVerificationInfo().name), submod), true });
```
kambala-decapitator, line 186: seems copy-pasted?

mapeditor/validator.cpp
```diff
@@ -199,3 +198,151 @@ std::set<Validator::Issue> Validator::validate(const CMap * map)
...
+	return pixmap;
+}
+
+void Validator::showValidationResults(const CMap * map)
+{
+	show();
+	setAttribute(Qt::WA_DeleteOnClose);
+	ui->listWidget->setItemDelegate(new ValidatorItemDelegate(ui->listWidget));
+
+	const std::array<QString, 2> icons{ ":/icons/mod-update.png", ":/icons/mod-delete.png" };
```
kambala-decapitator, line 232: you use it only once, no need to have an array

mapeditor/validator.cpp
```diff
@@ -199,3 +198,151 @@ std::set<Validator::Issue> Validator::validate(const CMap * map)
...
+		auto * item = new QListWidgetItem(QIcon(greenTick), validMessage, ui->listWidget);
+		ui->listWidget->addItem(item);
+	}
+
+	ui->listWidget->updateGeometry();
+
+	adjustWindowSize();
+}
+
+void Validator::adjustWindowSize()
```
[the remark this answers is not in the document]
kambala-decapitator: whe is this even needed? Just set a bigger default window size, user can resize the window.

mapeditor/validator.h
```diff
@@ -46,4 +46,34 @@ class Validator : public QDialog
...
+    int minItemWidth = 350;
+    int screenMargin = 350;     // some reserved space from screenWidth; used if text.width > screenWidth - screenMargin 
+	
+	const int offsetForIcon = iconPadding + textOffsetForIcon;
+
+    void paint(QPainter * painter, const QStyleOptionViewItem & option, const QModelIndex & index) const override;
+	QSize sizeHint(const QStyleOptionViewItem & option, const QModelIndex & index) const override;
+
+private:
+    QRect screenGeometry;
```
kambala-decapitator, line 78: broken indentation

kambala-decapitator (changes requested):

mapeditor/mainwindow.cpp
```diff
@@ -1147,9 +1159,20 @@ void MainWindow::on_inspectorWidget_itemChanged(QTableWidgetItem *item)
 
 void MainWindow::on_actionMapSettings_triggered()
 {
-	auto settingsDialog = new MapSettings(controller, this);
-	settingsDialog->setWindowModality(Qt::WindowModal);
-	settingsDialog->setModal(true);
+	if(!mapSettings)
+		return;
```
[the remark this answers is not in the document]
MichalZr6: Checked. Should be safe to delete that code.

mapeditor/validator.cpp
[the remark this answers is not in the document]
GeorgeK1ng: > why is this even needed? Just set a bigger default window size, user can resize the window.

Question is, why should map maker resize validator window each time when wanna to see results?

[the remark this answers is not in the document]
IvanSavenko: kambala makes a good point - why not have reasonable default size?

That ~100x100px window we had before is definitely not reasonable and inconvenient.

Instead of such complex logic we could:
1. Have it at more convenient size, like around 1000x600 (not insisting on particular number
2. Remember position / size of window between editor runs. We already do this for launcher window size/position, and I think - for editor main window.

Such approach would definitely work on any monitor, and even if default size will be inconvenient for somebody - he can resize it once and forget about this problem.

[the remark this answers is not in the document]
MichalZr6: > Can you confirm that this code works if desktop scaling is active?

If you mean the Windows desktop scaling then yes, confirmed with that being set to 100%, 125%, and 175% — the layout and sizing behave correctly across all cases.


> why is this even needed? Just set a bigger default window size, user can resize the window.

Worked in a company that had that policy: "user can resize". Users were super-annoyed.

mapeditor/PlayerSelectionDialog.cpp
MichalZr6: Serious? Maybe not. Reasonable? Absolutely.

mapeditor/validator.cpp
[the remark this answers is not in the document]
MichalZr6: > We already do this for launcher window size/position, and I think - for editor main window.

Yes, saw that in editor's code. It doesn't work.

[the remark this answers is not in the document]
kambala-decapitator: setting an adequate default size would cover the needs of 95% users

mapeditor/PlayerSelectionDialog.cpp
kambala-decapitator: I already foresee issues with localizations.

Also, haven't you seen those shitty non-resizable dialogs in Windows?

mapeditor/validator.cpp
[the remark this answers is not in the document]
IvanSavenko: > If you mean the Windows desktop scaling then yes

Ah, great.

> Worked in a company that had that policy: "user can resize".

Sounds like actual policy was "user has to resize". 

Sane defaults that work for most users + saving dimensions on resize (and fixing that logic if it is broken as you say) sounds more logical IMO

mapeditor/PlayerSelectionDialog.cpp
MichalZr6: Layouts are members because they're initialized in one method and modified in another

mapeditor/PlayerSelectionDialog.h
MichalZr6: If I do sth like this:

`checkbox->setFont(font);`

or with any label I have in dialog and the font is set as `QFont font;`

The text looks good (I have to only set its sizePoint which is done in PlayerSelectionDialog's constructor).

Otherwise it looks bad...

kambala-decapitator: I mean, no need to save the font object in class, just like with the layouts

mapeditor/mapcontroller.cpp
MichalZr6: But this needs `explicit` to be dropped from definition - no problem with that ?

mapeditor/mapcontroller.cpp
MichalZr6: Visual Studio marks it as `undefined identifier`. requestModsUpdate still works as signal without it but I wanted to mark it somehow like it is done in Qt. Perhaps the comment at the end of that line is sufficient, though...

mapeditor/validator.cpp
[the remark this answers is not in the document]
MichalZr6: I'm changing `setFixedWidth` and `setFixedHeight` to `setMinimumWidth` and `setMinimumHeight`.
This way they are resizable now and text is also wrapping correctly.

mapeditor/mapcontroller.cpp
kambala-decapitator: sure, `explicit` isn't that required

mapeditor/mapcontroller.cpp
IvanSavenko: `emit` macro from Qt is deliberately disabled, so yeah, won't compile / can be removed.

There was conflict since libtbb defines method with same name, and having that Qt macro defined breaks it.

kambala-decapitator: just use Q_EMIT

mapeditor/PlayerSelectionDialog.cpp
MichalZr6: I get the reasoning, but in this case I prefer to keep the layouts and font as private members — it keeps function signatures clean, especially since `addCheckbox` would otherwise take 6 parameters.

The dialog is small, case-specific, and rather not intended for reuse, so IMHO that encapsulation doesn’t bring any value here.

mapeditor/mapcontroller.cpp
MichalZr6: > But this needs `explicit` to be dropped from definition - no problem with that ?

... and the parameter as well

mapeditor/PlayerSelectionDialog.cpp
MichalZr6: I was probably making some layout adjustments, but in the end, deleting these lines is negligible.

mapeditor/PlayerSelectionDialog.cpp
MichalZr6: I didn't like how they are spread across height. Without that `setContentsMargins(0, 0, 0, 0);` it looks like this:




and with it:

mapeditor/validator.cpp
```diff
@@ -199,3 +198,151 @@ std::set<Validator::Issue> Validator::validate(const CMap * map)
...
+
+	painter.end();
+	return pixmap;
+}
+
+void Validator::showValidationResults(const CMap * map)
+{
+	show();
+	setAttribute(Qt::WA_DeleteOnClose);
+	ui->listWidget->setItemDelegate(new ValidatorItemDelegate(ui->listWidget));
```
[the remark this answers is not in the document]
MichalZr6: I wanted to customize how items are rendered in the list — specifically to control vertical spacing via `sizeHint()` so the entries don’t appear clamped. And also wanted to draw an icon **at the start** of each item, rather than centered in the multi-line cell.



vs

MichalZr6:
Fixed some conflicts with main branch and dropped commit with regenerated ts files (will do that in another PR).

Anything else to do here ?

IvanSavenko:
> Anything else to do here ?

@kambala-decapitator ?

mapeditor/validator.cpp
```diff
@@ -195,3 +194,125 @@ std::set<Validator::Issue> Validator::validate(const CMap * map)
...
+	{
+		auto * item = new QListWidgetItem(QIcon(issue.critical ? ":/icons/mod-delete.png" : ":/icons/mod-update.png"),
+			issue.message, ui->listWidget);
+
+		ui->listWidget->addItem(item);
+	}
+
+	if(ui->listWidget->count() == 0)
+	{
+		QPixmap greenTick = QPixmap(":/icons/mod-enabled.png");
```
kambala-decapitator, line 214:
```suggestion
		QPixmap greenTick(":/icons/mod-enabled.png");
```

mapeditor/validator.h
```diff
@@ -46,4 +46,33 @@ class Validator : public QDialog
 
 private:
 	Ui::Validator *ui;
+
+    QRect screenGeometry;
+
+	void showValidationResults(const CMap * map);
+    void adjustWindowSize();
```
kambala-decapitator, line 53: something's wrong with indentation in this file

mapeditor/mapcontroller.cpp
kambala-decapitator: it's very strange that `emit` is marked as undefined for you. But does the code compile? I'd be surprised if it doesn't, since we use Qt keywords heavily, like `slots` and `signals` in headers.

kambala-decapitator: Ah, just emit is disabled. Thanks!

mapeditor/PlayerSelectionDialog.cpp
```diff
@@ -0,0 +1,119 @@
...
+	// Select first available player by default
+	if(buttonGroup->buttons().isEmpty() && radioButton->isEnabled())
+	{
+		radioButton->setChecked(true);
+		selectedPlayer = player;
+	}
+
+	buttonGroup->addButton(radioButton, player.getNum());
+
+	auto * shortcutLabel = new QLabel(action->shortcut().toString(), this);
```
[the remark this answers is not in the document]
MichalZr6: Good idea, will do. This is going to fix at least one other... _issue_

mapeditor/mapcontroller.cpp
```diff
@@ -536,30 +543,85 @@ void MapController::commitObjectCreate(int level)
...
+	{
+		if(!_map->mods.count(mod.first))
+		{
+			QString submod;
+			if(!mod.second.parent.empty())
+				submod = tr(" (submod of %1)").arg(QString::fromStdString(mod.second.parent));
+
+			auto reply = QMessageBox::question(main,
+				tr("Missing Required Mod"),
+				tr("This object is from the mod '%1'%2.\n\n"
```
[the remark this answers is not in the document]
MichalZr6: Space is not missing. %2 has it and if it's empty then no trailing one is added to main message.
These strings are similar but not same. I made that helper function, but the message had to be rephrased, so it could be used in validator's warning list and user's yes/no dialog.

mapeditor/validator.h
MichalZr6: It had mixed space and tabs, on my side it looked good, but should be fixed in now.

GeorgeK1ng:
Any updates for merging this map editor improvements?

MichalZr6:
From my side, it's ready 🙂 
All review comments have been addressed, and I believe it's good to go.

kambala-decapitator (approved):

PR #5436, opened 2025-02-16
Small improvements to map sorting and search mapobjects feature

MichalZr6:
This PR enables sorting maps by file modification timestamp and changes sorting for translated map names that start with locale-specific character like "Żywa krew" in the example below.

After fixes (descending):
obraz

Also added to this PR small refinements to search mapobject feature so the matches that start with typed-in string will be placed at the top of the list:
obraz

Before fixes "Starożytna lampa" is somewhere down at the bottom of the list:
obraz

Wanted also to fix bad display of too long names in search window (they went out of the window's borders in beta branch). In pure develop branch it has been fixed somewhere along the way, however the quantity is not displayed as on image above.


MichalZr6:
The last ("calender") icon should allow sorting maps by last modified datetime. It was broken because those two variables `lastWrite` and `date` were not set among others in `mapInit()`

client/lobby/SelectionTab.cpp
```diff
@@ -114,13 +114,13 @@ bool mapSorter::operator()(const std::shared_ptr<ElementInfo> aaa, const std::sh
...
-			return boost::ilexicographical_compare(a->name.toString(), b->name.toString());
+			return TextOperations::compareLocalizedStrings(aaa->name, bbb->name);
 		case _fileName: //by filename
-			return boost::ilexicographical_compare(aaa->fileURI, bbb->fileURI);
+			return TextOperations::compareLocalizedStrings(aaa->fileURI, bbb->fileURI);
 		case _changeDate: //by changedate
 			return aaa->lastWrite < bbb->lastWrite;
 		default:
-			return boost::ilexicographical_compare(a->name.toString(), b->name.toString());
+			return aaa->name < bbb->name;
```
IvanSavenko, line 123: Why this place needs needs simple string comparison and not locale-aware like others?

lib/texts/TextOperations.h
```diff
@@ -74,10 +74,25 @@ namespace TextOperations
...
+		static const std::locale loc(getLocaleName());
+		static const std::collate<char> & col = std::use_facet<std::collate<char>>(loc);
+
+		return col.compare(str1.data(), str1.data() + str1.size(),
+			str2.data(), str2.data() + str2.size()) < 0;
+	}
 
 	/// Check if texts have similarity when typing into search boxes
-	DLL_LINKAGE bool textSearchSimilar(const std::string & s, const std::string & t);
+	/// 0 -> Exact match or substring match, 1 - 2 -> Close match based on Levenshtein distance, >100 -> Unrelated word(bad match).
```
IvanSavenko, line 93: Not a huge fan of magic constants...

Perhaps instead return std::optional?

client/windows/GUIClasses.cpp
```diff
@@ -1590,21 +1598,76 @@ void CObjectListWindow::init(std::shared_ptr<CIntObject> titleWidget_, std::stri
...
+	std::regex pattern(R"(.*(\(\d+\))$)");
+	std::smatch match;
 
-		selected = 0;
-		list->resize(itemsVisible.size());
-		list->scrollTo(0);
-		ok->block(!itemsVisible.size());
+	std::string quantity;
+	if(std::regex_match(text, match, pattern) && match.size() > 1)
+		quantity = match[1]; // Extract the quantity
```
IvanSavenko, line 1617: Why you need regex here?

client/windows/GUIClasses.cpp
```diff
@@ -1590,21 +1598,76 @@ void CObjectListWindow::init(std::shared_ptr<CIntObject> titleWidget_, std::stri
...
-		searchBoxDescription->setEnabled(text.empty());
+	searchBox->setCallback(std::bind(&CObjectListWindow::itemsSearchCallback, this, std::placeholders::_1));
+}
+
+void CObjectListWindow::trimTextIfTooWide(std::string & text) const
+{
+	int maxWidth = pos.w - 60;	// 60 px for scrollbar and borders
+
+	// Create a temporary label to measure text width
+	auto label = std::make_shared<CLabel>(0, 0, FONT_SMALL, ETextAlignment::CENTER, Colors::WHITE, text);
```
IvanSavenko, line 1609: Use `IFont::getStringWidth`

lib/texts/TextOperations.cpp
```diff
@@ -300,30 +301,62 @@ int TextOperations::getLevenshteinDistance(const std::string & s, const std::str
...
+
+		return localeName;
+	}
+	catch(const std::exception & e)
+	{
+		logGlobal->warn("Failed to retrieve locale. %s", e.what());
+	}
+
+	return "en_US.UTF-8";
+}
```
IvanSavenko, line 321: 1. Why you need exception throwing? Just warning + return would work just as well
2. In which conditions this might happen? From what I understand, it can never happen.

I think this entire method can be replaced with simple 1-liner
```suggestion
DLL_LINKAGE std::string TextOperations::getLocaleName()
{
	return Languages::getLanguageOptions(LIBRARY->generaltexth->getPreferredLanguage()).localeName;
}
```

lib/texts/TextOperations.h
MichalZr6: This function only enhances scoring which Levenshtein algorithm function produces. `std::optional` will not help here. The comment is outdated here, though.

lib/texts/TextOperations.cpp
MichalZr6: Totally right. It's leftover when I was trying to set up `std::locale` here. Should it be moved to header and marked with `inline` ?

lib/texts/TextOperations.h
IvanSavenko: If it is outdated, why there are multiple checks for magic 100 in code?

lib/texts/TextOperations.cpp
IvanSavenko: Can be kept in .cpp

Or alternatively - remove entirely (if it is only used in 1 place)

client/windows/GUIClasses.cpp
MichalZr6: To extract given mapobject's quantity and add it to new, trimmed text. Can be dropped but then another parameter has to be there or `item &` has to be passed not item's text.

IvanSavenko: Passing map name & quantity separately makes more sense (whether in form of `item` class or simply as separate parameters).

Concatenating strings only to use regex later to split them back makes no sense. And will break if map object uses `(1)` in its name.

lib/texts/TextOperations.h
MichalZr6: Ah, this magic. Ok, sure, will use `std::optional`

client/windows/GUIClasses.cpp
```diff
@@ -1590,21 +1598,76 @@ void CObjectListWindow::init(std::shared_ptr<CIntObject> titleWidget_, std::stri
...
+		// Trim text until it fits
+		while(!text.empty())
+		{
+			std::string trimmedText = text + suffix;
+			label->setText(trimmedText);
+
+			if(label->getWidth() < maxWidth)
+				break;
+
+			text.resize(text.size() - 1);
```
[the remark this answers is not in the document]
MichalZr6: Good point. Will use `TextOperations::trimRightUnicode()`

MichalZr6:
> > Also added to this PR small refinements to search mapobject feature so the matches that start with typed-in string will be placed at the top of the list
> 
> Is that logic also applied on spellbook search (which use same logic in general)?

I left the sorting logic unchanged as `SpellbookSpellSorter` class does that by spell level and then by name. Only added `compareLocalizedStrings()` for name comparisons there.
After typing just few letters the spell count drops drastically so I don't think it is worth to implement similarity score for additional sorting layer.

client/windows/GUIClasses.cpp
```diff
@@ -1590,21 +1598,76 @@ void CObjectListWindow::init(std::shared_ptr<CIntObject> titleWidget_, std::stri
 	searchBoxDescription = std::make_shared<CLabel>(r.center().x, r.center().y, FONT_SMALL, ETextAlignment::CENTER, grayedColor, LIBRARY->generaltexth->translate("vcmi.spellBook.search"));
 
 	searchBox = std::make_shared<CTextInput>(r, FONT_SMALL, ETextAlignment::CENTER, true);
-	searchBox->setCallback([this](const std::string & text){
-		searchBoxDescription->setEnabled(text.empty());
+	searchBox->setCallback(std::bind(&CObjectListWindow::itemsSearchCallback, this, std::placeholders::_1));
+}
+
+void CObjectListWindow::trimTextIfTooWide(std::string & text) const
```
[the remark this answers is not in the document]
MichalZr6: But you can see right away how many are there without touching the mouse.
Also one polish contributor - updating translations - has added to some mapobjects' names their descriptions making those names overly long. So I wanted to log this to console so someone might get the clue and fix it.
This can be also done while loading mapobjects, but dunno which length should be treated as 'too long'.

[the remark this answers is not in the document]
IvanSavenko: > Also one polish contributor - updating translations - has added to some mapobjects' names their descriptions making those names overly long. 

This should be fixed / removed, and not supported. Most object now support separate "description" entry in json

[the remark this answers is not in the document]
MichalZr6: oh my god... this is not quantity but id.

PR #5438, opened 2025-02-16
[1.6.6] portmaster support

Laserlicht:
@IvanSavenko Do you think we should merge this one?

It make it easier to build VCMI for portmaster. As portmaster builds are quite complicated...

With this we would avoid patching VCMI files and provide a docker file that could run on ARM64 (or [ARM64 qemu](https://portmaster.games/docker.html)) which makes this easier.

Even if we would not merge it's for documenting how it can build.

Video is working (related https://github.com/PortsMaster/PortMaster-New/issues/497).

@ddrsoul


IvanSavenko:
Not against it.

kambala-decapitator (approved):

PR #5460, opened 2025-02-21
[1.6.7?] Add settings

Laserlicht:
- setting for disabling overlay
- setting for disabling async upscaling (config only)
- save map zoom

### DIFF
--- a/client/mapView/MapView.cpp
+++ b/client/mapView/MapView.cpp
@@ -211,7 +211,7 @@ void MapView::onViewMapActivated()
 {
 	controller->activateAdventureContext();
 
-	int zoom = persistentStorage["tileZoom"].Integer();
+	int zoom = settings["adventure"]["tileZoom"].Integer();
 	if(zoom)
 		controller->setTileSize(Point(zoom, zoom));
 	else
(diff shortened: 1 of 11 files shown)


client/mapView/MapView.cpp
```diff
@@ -210,7 +210,12 @@ void MapView::onMapZoomLevelChanged(int stepsChange, bool useDeadZone)
 void MapView::onViewMapActivated()
 {
 	controller->activateAdventureContext();
-	controller->setTileSize(Point(32, 32));
+
+	int zoom = persistentStorage["tileZoom"].Integer();
```
IvanSavenko, line 214: Why not in settings, as usual?
```suggestion
	int zoom = persistentStorage["tileZoom"].Integer();
```

misiokles:
But editing this lines are hard for mobile devices, doesn't it?

client/mapView/MapView.cpp
IvanSavenko: tutorial makes sense. But kingdom overview & this one - not really. Should be in settings

IvanSavenko: Hmm...

- Is this related to game progress? Like completed campaigns? Or completed tutorial? To persistent.

- Is this something that can be changed by player explicitly at least in theory? To settings. Like last kingdom interface - there is nothing wrong with having player select this in UI. Sure, basically nobody would need it, and it would make things harder to find, but there is nothing wrong with making this option configurable

PR #5468, opened 2025-02-23
[1.6.7] make lobby 4:3 aspect ratio & fixes

Laserlicht:
Change aspect ratio of lobby to 4:3 to avoid broken window on 4:3 game.
After change it's in line with other big windows (800x600) in H3/VCMI.
If necessary it could extended with mod to 16:9...

Other fixes:
- fix graphical glitch when scrolling user (invite) list
- fix long text for (dynamic) label
- fix incorrect help for lobby create button
- fix incorrect help for lobby send message button
- fix stuck (pressed) slider when clicking on empty slider area

grafik

### DIFF
--- a/Mods/vcmi/Content/config/chinese.json
+++ b/Mods/vcmi/Content/config/chinese.json
@@ -175,7 +175,7 @@
 	"vcmi.lobby.match.solo" : "单人游戏",
 	"vcmi.lobby.match.duel" : "与 %s 的游戏", // %s -> nickname of another player
 	"vcmi.lobby.match.multi" : "%d 个玩家",
-	"vcmi.lobby.room.create" : "创建房间",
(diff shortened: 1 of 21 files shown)


IvanSavenko:
Perhaps provide both layouts and select in runtime? So if player has 16x9 or more -> pick wide layout, othervice - pick narrow version?

dydzio0614:
IMO wide version doesn't have real downsides so we could use resolution width check - similar one that determines positioning of hero info in battle to be inside window for resolution being 800x600 etc.

IvanSavenko:
No such functionality. So something like this should be OK:
```
if (screenWidth < 960) // or whatever is width of wide window
    config = JsonPath("...4x3");
else
    config = JsonPath("...16x9");
```

PR #5478, opened 2025-02-25
develop - add greek, japanese, norwegian; harmonize

heroesiiifan:


IvanSavenko:
Do we actually have volunteers to translate VCMI into these languages?

IvanSavenko:
However until we actually get translations to those languages we will have language selection in Launcher that does not actually translates anything. Which is not the best UX.

BTW - judging from Google Play stats, Romanian & Bulgarian are also candidates for translation support. However same question stands - will we find somebody to do actual translation?

IvanSavenko:
> So for now just deactivate selection in Launcher? Then it can still selected via config for the translator?

Yeah, that's one option.

> Should I add them too?

Don't know. Maybe? To be honest I am not sure about translations from your PR either. If they will be disabled/hidden - OK I guess, but don't see much point unless somebody is working / planning to work on them.

PR #5488, opened 2025-02-28
replace some boost functions with std

Laserlicht:
Replace some boost function with std (cpp11 - cpp17).

Thread isn't replaceable because `boost::this_thread::interruption_point()` is not supported.
Filesystem would need some manual rework (no 100% 1 to 1 replacement).

@IvanSavenko Any reasons that speak against it?

### DIFF
--- a/AI/BattleAI/PotentialTargets.cpp
+++ b/AI/BattleAI/PotentialTargets.cpp
@@ -84,7 +84,7 @@ PotentialTargets::PotentialTargets(
 		}
 	}
 
-	std::sort(possibleAttacks.begin(), possibleAttacks.end(), [](const AttackPossibility & lhs, const AttackPossibility & rhs) -> bool
+	boost::sort(possibleAttacks, [](const AttackPossibility & lhs, const AttackPossibility & rhs) -> bool
 	{
 		return lhs.damageDiff() > rhs.damageDiff();
 	});
--- a/AI/BattleAI/PotentialTargets.cpp
+++ b/AI/BattleAI/PotentialTargets.cpp
@@ -84,7 +84,7 @@ PotentialTargets::PotentialTargets(
 		}
 	}
 
-	boost::sort(possibleAttacks, [](const AttackPossibility & lhs, const AttackPossibility & rhs) -> bool
+	std::sort(possibleAttacks.begin(), possibleAttacks.end(), [](const AttackPossibility & lhs, const AttackPossibility & rhs) -> bool
 	{
 		return lhs.damageDiff() > rhs.damageDiff();
 	});
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -1582,7 +1582,7 @@ void AIGateway::buildArmyIn(const CGTownInstance * t)
 void AIGateway::finish()
 {
 	//we want to lock to avoid multiple threads from calling makingTurn->join() at same time
-	boost::lock_guard<std::mutex> multipleCleanupGuard(turnInterruptionMutex);
+	std::lock_guard<std::mutex> multipleCleanupGuard(turnInterruptionMutex);
 
 	if(makingTurn)
 	{
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -836,7 +836,7 @@ void AIGateway::makeTurn()
 	auto day = cb->getDate(Date::DAY);
 	logAi->info("Player %d (%s) starting turn, day %d", playerID, playerID.toString(), day);
 
-	boost::shared_lock gsLock(CGameState::mutex);
+	std::shared_lock gsLock(CGameState::mutex);
 	setThreadName("AIGateway::makeTurn");
 
 	if(nullkiller->isOpenMap())
@@ -1582,7 +1582,7 @@ void AIGateway::buildArmyIn(const CGTownInstance * t)
 void AIGateway::finish()
 {
 	//we want to lock to avoid multiple threads from calling makingTurn->join() at same time
-	boost::lock_guard<boost::mutex> multipleCleanupGuard(turnInterruptionMutex);
+	boost::lock_guard<std::mutex> multipleCleanupGuard(turnInterruptionMutex);
 
 	if(makingTurn)
 	{
@@ -1598,7 +1598,7 @@ void AIGateway::requestActionASAP(std::function<void()> whatToDo)
 	{
 		setThreadName("AIGateway::requestActionASAP::whatToDo");
 		SET_GLOBAL_STATE(this);
-		boost::shared_lock gsLock(CGameState::mutex);
+		std::shared_lock gsLock(CGameState::mutex);
 		whatToDo();
 	});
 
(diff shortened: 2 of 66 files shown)


IvanSavenko:
Generally no objections
I would also consider:

- Removing this from global.h:
```
// Lock typedefs
using TLockGuard = std::lock_guard<std::mutex>;
using TLockGuardRec = std::lock_guard<std::recursive_mutex>;
```
- Replacing `std::lock_guard<std::recursive_mutex>` with auto-deduction `std::lock_guard` (and similar)
- Use minimalistic `std::lock_guard` in place of `std::unique_lock` & `std::scoped_lock` where possible


Regarding filesystem - AFAIK we can't use it at the moment since this would require bumping minimal supported version of macOS and/or iOS, and kambala is against that.

As for thread - will probably take a look later, when I'll be working on implementing guaranteed shutdown order for AI.

IvanSavenko:
> std::sort instead of boost::sort

Not sure about this one to be honest, since this is not a direct replacement. Direct replacement would be ranges, but they are not in C++17.

However I am not against bumping our standard requirement from C++17 to newer version (C++20?). Need to check support though - I think there are a couple of features we can't use without bumping iOS/macOS version.

Worth discussing IMO, but might be better to do this on Discord or in separate issue here, to make sure that everyone is OK with that.

dydzio0614:
if I remember correctly, c++20's jthread is needed to replace boost threads

IvanSavenko:
> adding C++20 support would drop compatibility with some old iphones 

Not exactly. Even C++17 is not fully supported on apple devices we target. As long as we don't use unavailable features we should be fine even if we bump C++ standard

> + potentially older releases of stable linux distros

https://packages.debian.org/bullseye/g++

This is Debian's old stable and it already has gcc-10 which supports like 90% of C++20
Current Debian stable has gcc-12 which has basically full support of C++20

IvanSavenko:
> c++20's jthread would be needed to replace boost threads

jthread might actually be something that is not supported by apple clang. In any case I'd rather clean up AI code to remove need of jthread to begin with.

dydzio0614:
> Not exactly. Even C++17 is not fully supported on apple devices we target. As long as we don't use unavailable features we should be fine even if we bump C++ standard

I assume we would need to avoid notable amount of C++20 features - even bigger % of features than in C++17 would be unavailable for these devices

IvanSavenko:
https://developer.apple.com/xcode/cpp/#c++20

From my understanding:
- Concepts: not supported on any version
- barrier/latch/semaphore - requires newer OS 
- bunch of other features require newer Xcode, which I think we should be able to use?

So not that many of unusable features.

dydzio0614:
I see that apple deprecated some more iphone devices - it needs to be decided what we actually wanna support

IvanSavenko:
@kambala-decapitator 

How macOS/Xcode compatibility works? Using this page:
https://developer.apple.com/xcode/cpp/#c++20

It has 2 different comments on some newer c++ features:
> Minimum deployment target: macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0

This one is clear - we can't use it without bumping OS requirement. However other than std::filesystem such features are very rare, so not a problem, we can just exclude / blacklist them.

> Requires Xcode 15 or later

This one raises some questions:

- will using newer Xcode require raising minimal target OS?
- right now CI uses Xcode 14.2. Can we raise it to 15.3 for example? Will this require new prebuilts?
- any other potential problems from requiring newer Xcode?

kambala-decapitator:
regarding Apple OSs

C++17: currently we can use everything except filesystem, to_chars/from_chars and memory_resource (never heard about the latter :) ). I also vaguely remember there was some issue with one of `optional`'s constructors or methods.

> will using newer Xcode require raising minimal target OS?

it depends on Xcode version. For example, last raise was in Xcode 14 that started requiring iOS 12 (previously it was iOS 9, and VCMI+deps used to build fine for iOS 10), but for macOS that's still 10.13 (if not earlier). Hence, Xcode 16 (current) also requires minimum of iOS 12.

> right now CI uses Xcode 14.2. Can we raise it to 15.3 for example? Will this require new prebuilts?

**usually** it's fine, I've already tested building with Xcode 16 in #4958. Sometimes there can be sudden runtime issues, so it's better to test such change first. Prebuilts can stay, but I plan to switch them (as well as VCMI build) to Xcode 16 with Conan v2 transition anyway.

> any other potential problems from requiring newer Xcode?

things can break randomly :)

<hr>

I don't see much reason to drop older Apple devices. Do you see C++20 library features that are highly desired to be used? Btw next jump in deployment target would be iOS 15 (all devices capable of iOS 13 are upgradeable to 15).

regarding the PR: provide a way to use boost fs instead of std fs, that would be a new cmake option as well as a wrapper for things that aren't replaceable 1-to-1. And in code you'd have something like
```cpp
#if USE_STD_FILESYSTEM
using fs = std::filesystem;
#else
using fs = boost::filesystem;
#endif

// usage:
fs::path
```

IvanSavenko:
> Hence, Xcode 16 (current) also requires minimum of iOS 12.

Great.

> Prebuilts can stay, but I plan to switch them (as well as VCMI build) to Xcode 16 with Conan v2 transition anyway.

Then let's keep C++17 for now and consider C++20 after Xcode/Conan switch.

> Do you see C++20 library features that are highly desired to be used? 

"Highly" is debatable, but yes. However none of them are marked as requiring newer deployment target, only newer Xcode. So I assume we can use them without bumping OS requirements?
- Text formatting (std::format) -> Xcode 15.3
- Ranges -> Xcode 14.3
- Library support for the spaceship (comparison) operator `<=>`
- `<span>`

kambala-decapitator:
> However none of them are marked as requiring newer deployment target, only newer Xcode. So I assume we can use them without bumping OS requirements?

yes, correct. Using newer _language_ features is easier than _library_ features.

dydzio0614:
I am not sure about value of iOS 12 support these days though - can iphone 6 run VCMI at all with its RAM memory? Also it got already deprecated in previous deprecation pass prior to the latest one

IvanSavenko:
> I am not sure about value of iOS 12 support these days though - can iphone 6 run VCMI at all with its RAM memory? Also it got already deprecated in previous deprecation pass prior to the latest one

On the other hand - what we will gain from dropping iOS 12? According to that page:
- iOS 13: Elementary string conversions std::to_chars, std::from_chars - nice, but plenty of alternatives
- iOS 13: Filesystem library - can help with reducing boost usage, but not really needed.
- iOS 15: Synchronization library (<barrier>, <latch>, <semaphore> and notification functions on std::atomic) - not very useful IMO. And I'd rather try async approaches rather than waits
- iOS 17: memory_resource - not needed for VCMI at all

Everything else only requires newer Xcode which won't require bumping OS requirements

kambala-decapitator:
@dydzio0614 android 5.0 is way older than ios 12, but you're not questioning that ;)

IMO heroes 3 should be available to the biggest amount of devices as possible.

IvanSavenko:
Made some investigation into filesystem, and it is going to be fun.
- boost::filesystem is also used by innoextract. In theory we could try to switch our fork, but might not be trivial & hard to maintain in future. So we can't completely remove dependency on boost::fs.
- boost::filesystem have dropped support of Win7 in 1.87. So I guess we should keep Windows targets at boost 1.86. Switching to std would've helped, but we'll need to decide how to handle innoextract

kambala-decapitator:
> But I didn't experience the development of innoextract as really open

not sure what you mean. Some PRs were merged as is, some were used as a base in maintainer's commits (according to his words), stupid PRs were closed, some PRs were closed by their authors.

PR #5491, opened 2025-02-28
Use tbb thread pool instead of custom one for RMG

IvanSavenko:
Should have no/minimal effect on performance, but less code to maintain, especially since we already use TBB.

Also added basic docs on all threads created by VCMI.


DjWarmonger:
Clash of Dragons G + U
Ryzen 7900X3D, 12 cores:

- Current:
    - 9755 ms
    - 8134 ms
    - 14464 ms
    - 9197 ms
    - 7851 ms
- TBB:
    - 13658 ms
    - 9477 ms
    - 10781 ms
    - 8496 ms
    - 7808 ms

Looks like there's no difference between both.

PR #5495, opened 2025-03-02
ROE Demo support

Laserlicht:
Missing:
- Auto enable / disable demo mod
- good launcher support
- ARTRAITS, ARTEVENT, CRTRAITS loading (actually these there has to manually copied in data folder)

Otherwise txt mapping is functional and demo is playable. Some warnings about missing audio.

Demo is downloaded and extracted from Loki in launcher.


IvanSavenko:
Some thoughts about demo:
- would require a way to "upgrade" game data in Launcher, e.g. reopening first launch dialog. Might be actually easy now - can be added as extra button below import files.

- mods should probably be disabled / blocked. Reasoning: a lot of mods use modified H3 assets (so we probably shouldn't allow them to be downloaded freely), or rely on them (sounds for creatures, town hall backgrounds, etc, siege screens)

- not sure about having demo support as mod - since players with full game will also have it. Might be better to have it as alternative game config - load `gameConfig.json` for full game, and more limited `demoConfig.json` for demo (shared parts can be extracted into separate file to reduce duplications)

- multiplayer should also be blocked, if it is not already - because this is demo, because there are likely missing assets, and because other players generally have full game.

IvanSavenko:
> The problem with this PR is that it would need a hybrid thing for mod loading. For the textfiles you have to load 8 towns. But later there are only 3 towns enabled.

Actually we already sort of do. What makes VCMI associate 1st entry in h3 files with Castle is this:
```json
"castle" : { "id" : 0 }
```
In other words, if you were to completely remove this block, VCMI will parse, but won't load 1st entry from H3 files.

However this is not really tested, and would create null pointer in `VLC->factions` vector, which game might try to access in some places. So fixable in theory, but needs investigation

IvanSavenko:
Was afraid of that... Fixing those should be rather easy, but rather time-consuming

GeorgeK1ng:
> @GeorgeK1ng
> 
> Here is a way to download and extract RoE demo in C++. The Loki version is the easiest, because:
> 
> * It is legally downloadable
>   
>   * Loki is official distributor for linux release
>   * Still downloadable from official source (after 25 years!!!)
>   * Other demos from archive.org have a questionable right situation (they are copyrighted and unlike Heroes ii, the readme does not explicitly allow redistribution)
> * easy to extract (tar.gz with each file gzipped again)
> 
> http://updates.lokigames.com/loki_demos/heroes3-demo.run
> 
> https://github.com/Laserlicht/vcmi/blob/cf8ab42b322b2cb394f44634728873b86d03c23a/launcher/demo.cpp#L25

Sharing untouched demo should be fine, thats the purpose of games demo version. And it's real demo, it doesn't contain full game which can be unlocked somehow. 

There are more hostings where we can still obtain both Windows and MacOS demo. Softpedia, Fileplanet etc. Until you don't share touched demo files, then it's fine. That's why I have created script builder to merge both demos. Merging LOD, VID, SND from MacOS Demo over Windows demo give better results. And from Windows are grabbed MP3. 

I don't think we need support this approach directly from launcher, but just mention it's possible to get files for using VCMI with 3 towns legaly and free. I have already updated Demo mod support to clear almost all errors from console and added various compatibility submods. And Demo mod have disable multiplayer due missing resources, but could be possible on LAN for example when I create needed resources from non-copyrighted materials.

In my opinion is enought to add automatic Demo mod enabling when user uses Launcher option "Use existing files" so VCMI will copy just small amount of files. That should be all, nothing more. And on main Launcher screen just show red message "You are using Heroes 3 Demo files, your gameplay experience will be limited. To enjoy full game please import Heroes 3 Complete installer from GOG using **Import button**". GOG word should be [direct link ](https://www.gog.com/en/game/heroes_of_might_and_magic_3_complete_edition)to show players directly where to get full game at all.

Not sure if everything in VCMI works as expected with merged demos, but it's playable and can be combined with many mods - I have tested just some of them like HotA, WoG, ToW, New creatures pack, New terrains pack, New objects pack, 10+ towns.

And only document for players how to use Demo files with VCMI, nothing more. And of course to build ultimate demo you need to do that on Windows host. Have not found adequate source codes for needed apps, and don't want burn time with that.

So currently VCMI looks like it loads and is playable with Demo files. On Windows can user build required demo, on first launch can select folder where is merged demo. Then VCMI should detect it's demo selection and automatically download Demo support mod, same as we do for launguages and just show mentioned message on main launcher page.

GeorgeK1ng:
I think info about VCMI and Demo files is now clear in [Heroes III DEMO Builder for VCMI](https://github.com/GeorgeK1ng/heroes3demo)

PR #5496, opened 2025-03-02
Replace boost::thread with std::thread

IvanSavenko:
Removes all usages of boost::thread from vcmi in favor of std::thread. Also removed most of usages of boost::posix_time and boost::chrono - since std threading utilities expect only std::chrono. For thread interruptions used by adventure AI I've made replacement that should be identical to previous logic. Battle AI for some reason had check for interruption, but 0 cases where interruption could be generated (old code leftover?), so it was removed completely.

- [x] Remove CThreadHelper class, replace it with tbb
- [x] Switch non-AI code to std::thread
- [x] Implement replacement for thread interruption logic
- [x] Switch AI code to ~std::thread~ tbb
- [x] Remove workaround for potential crash on shutdown if game is being closed while AI is waiting for reply from server with proper fix
- [x] Consider replacing short-lived AI threads with tbb tasks or some other way to guarantee full AI shutdown


IvanSavenko:
Presumably done, lib/client now use std::thread, adventure AI's now use TBB to avoid spawning hundreds of threads. Will run some tests in release mode with profiler later, but don't expect any noticeable changes in performance or stability.

Now AI guarantees that all AI threads (or well, tbb tasks) are shut down when client requests termination, wasn't able to reproduce any crashes during AI turn anymore.

Dependency on boost thread still exists - apparently used by boost locale for some reason. In theory can be replaced with direct usage of iconv and remove the middleman, but not high in my priorities.

PR #5499, opened 2025-03-03
Remove most of usages of ConstTransitivePtr

IvanSavenko:
This removes most of remaining usages of ConstTransitivePtr:
- CTown now uses unique_ptr with const values to store buildings and structures
- StackLocation now uses ObjectInstanceID instead of pointer to army instance
- StartInfo and CMap in CGameState is now private (with accessors) and stored as unique_ptr's

This leaves only CMap using ConstTransitivePtr with unclear ownership. Will look into replacing those with shared_ptr in separate PR a bit later

### DIFF
--- a/AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
+++ b/AI/Nullkiller/Analyzers/BuildAnalyzer.cpp
@@ -202,7 +202,7 @@ BuildingInfo BuildAnalyzer::getBuildingOrPrerequisite(
 	BuildingID building = toBuild;
 	auto townInfo = town->getTown();
 
-	const CBuilding * buildPtr = townInfo->buildings.at(building);
+	const auto & buildPtr = townInfo->buildings.at(building);
 	const CCreature * creature = nullptr;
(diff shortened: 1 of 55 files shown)


IvanSavenko:
> again, what is/was ConstTransitivePtr used for?

To enforce const-correct access to pointers that are available as public members:

```cpp
class GameState {
ConstTransitivePtr<CMap> map;
}

GameState * gs;
gs->map->setSomething(); // OK, write access via non-const gamestate (e.g. server)

const GameState * gs;
gs->map->setSomething(); // FAIL, const gamestate allows only read access (e.g. client)
```

However:
- these are non-owning pointers. For example map object can be owned by net pack, CMap, or by some map editor entities, depending on context. And I don't want to play ownership juggling once I start trying to rollback netpacks.
- its implementation is buggy. For example, you can create a copy and get write access to stored pointer.

Don't see any easy way to fix these problems. Or more correctly - I do see them, but their complexity is basically same as switching to something better and well-tested, such as unique/shared pointers.

PR #5518, opened 2025-03-08
[develop] fix crash, when no cost for upgrade

### LINKED ISSUE #5517: If creature's upgrade fee is zero, it crashes when you use one-click upgrade
**Describe the bug**
If creature's upgrade fee is zero, it crashes while using one-click upgrade in Town.

This doesn't normally happen, but it can happen in some mods. If the creature's upgrade price is zero, one-click upgrades cannot be used and crash occurs.

Even if the gold cost is zero, it should be displayed normally.

**Version**

- OS: Windows，Android
- Version: 1.6.7

Laserlicht:
fixes #5517

@IvanSavenko Do we release 1.6.8? Or should fixes like this go into develop?

### DIFF
--- a/client/windows/CCastleInterface.cpp
+++ b/client/windows/CCastleInterface.cpp
@@ -399,6 +399,8 @@ void CHeroGSlot::gesture(bool on, const Point & initialPosition, const Point & f
 		std::vector<std::shared_ptr<CComponent>> resComps;
 		for(TResources::nziterator i(upgradableSlots.totalCosts); i.valid(); i++)
 			resComps.push_back(std::make_shared<CComponent>(ComponentType::RESOURCE, i->resType, i->resVal));
+		if(!resComps.size())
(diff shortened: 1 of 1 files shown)


client/windows/CCastleInterface.cpp
```diff
@@ -399,6 +399,8 @@ void CHeroGSlot::gesture(bool on, const Point & initialPosition, const Point & f
 		std::vector<std::shared_ptr<CComponent>> resComps;
 		for(TResources::nziterator i(upgradableSlots.totalCosts); i.valid(); i++)
 			resComps.push_back(std::make_shared<CComponent>(ComponentType::RESOURCE, i->resType, i->resVal));
+		if(!resComps.size())
```
IvanSavenko, line 402:
```suggestion
		if(resComps.empty())
```

IvanSavenko:
> Do we release 1.6.8? 

For now let's assume that we don't. And if we do - I can cherry-pick useful fixes from develop.

PR #5545, opened 2025-03-12
optimize skip video keys

### LINKED ISSUE #5528: Keyboard Volume roller incorrectly skips intro videos
**Describe the bug**
Keyboard Volume roller incorrectly skips intro videos.
It correctly does not skip campaign start/intermission videos.. But also no keyboard keys can skip them either, but  a left mouse click does.

**To Reproduce**
Use any keyboard with a volume roller/wheel/knob/button, spin/press it during intro video during startup, or post startup upon starting any chronicles campaign, it will play an intro video that also gets skipped like this.

**Expected behavior**
Really just dedicated keys should skip videos to prevent accidental skips (like my volumes too low i cant hear this and up the volume... ugh its gone). I'd guess its coded to ANY key, which can cause odd functions like this to trigger the skip. (and possibly bugs with other peripherals with unusual scan codes)
Suggested keys: Esc, Spacebar
And the behavior should be identical between intro video and campaign video.

**Actual behavior**
Strangely the campaign videos behave different than the intro videos. They are not skipped with volume wheel. 

**Did it work earlier?**
In the original game, this bug also exists. Tho worse as the volume wheel also skips campaign videos in the original game.

**Version**
- OS: Windows 11
- Version: 1.6.7

Laserlicht:
Campaign videos and main menu videos can skipped with ESC and ENTER now (mouse click still working).

Fixes #5528

### DIFF
--- a/client/mainmenu/CPrologEpilogVideo.cpp
+++ b/client/mainmenu/CPrologEpilogVideo.cpp
@@ -14,6 +14,7 @@
 #include "../media/IMusicPlayer.h"
 #include "../media/ISoundPlayer.h"
 #include "../GameEngine.h"
+#include "../gui/Shortcut.h"
 #include "../widgets/TextControls.h"
 #include "../widgets/VideoWidget.h"
 #include "../widgets/Images.h"
@@ -92,7 +93,7 @@ void CPrologEpilogVideo::show(Canvas & to)
 		text->showAll(to); // blit text over video, if needed
 }
 
-void CPrologEpilogVideo::clickPressed(const Point & cursorPosition)
+void CPrologEpilogVideo::exit()
 {
 	ENGINE->music().setVolume(ENGINE->music().getVolume() * 2); // restore background volume
 	close();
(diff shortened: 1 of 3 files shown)


client/mainmenu/CPrologEpilogVideo.cpp
```diff
@@ -102,3 +103,14 @@ void CPrologEpilogVideo::clickPressed(const Point & cursorPosition)
...
 }
+
+void CPrologEpilogVideo::clickPressed(const Point & cursorPosition)
+{
+	exit();
+}
+
+void CPrologEpilogVideo::keyPressed(EShortcut key)
+{
+	if(key == EShortcut::GLOBAL_ACCEPT || key == EShortcut::GLOBAL_CANCEL)
```
IvanSavenko, line 114: There is already key that is bound to both Enter and Escape.
Accept / Cancel are for Yes/No dialogs and similar.
```suggestion
	if(key == EShortcut::GLOBAL_RETURN)
```

PR #5548, opened 2025-03-13
Fix the problem that launcher doesn't close after starting game

### LINKED ISSUE #5129: launcher doesn't close after starting game
**Describe the bug**
Launcher window stays in background after starting the game and I have 2 VCMI icons in the dock. Also, launcher window is frozen (seems it's waiting for client to quit).

**To Reproduce**

1. Start launcher
2. Start game from it
3. Move game window to expose launcher window

**Expected behavior**
Launcher gets closed just like before 1.6

**Actual behavior**
Launcher window stays on screen (also eating RAM)

**Did it work earlier?**
yes, before 1.6

**Version**

- OS: macOS Intel,
- Version: 1.6

BBlaza:
Fixes #5129

The window of launcher will freeze after the game start becausethe "process" destoried after function startExecutable() return so it have to contain waitForFinished(-1) to ensure it exist before the game end.

By changing the "process" to unique pointer, I can use QObject::connect to make the lifetime of "process" long enough (destory when recieved &QProcess::finished), so there's no need for waitForFinished(-1). The window won't freeze in that case. (ChatGPT helped me in how to use QObject::connect to keep the lifetime long enough. The majority of the code is from myself)

### DIFF
--- a/launcher/main.cpp
+++ b/launcher/main.cpp
@@ -111,7 +111,7 @@ void startExecutable(QString name, const QStringList & args)
 	// Start vcmiclient and vcmieditor with QProcess::start() instead of QProcess::startDetached()
 	// since startDetached() results in a missing terminal prompt after quitting vcmiclient.
 	// QProcess::start() causes the launcher window to freeze while the child process is running. By changing
-	// the "process" to pointer, the launcher will be detached on MacOS while sticking with QProcess::start().
+	// the "process" to pointer, the "process" exist after startExecutable return.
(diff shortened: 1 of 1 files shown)


IvanSavenko:
- Please don't create duplicated pull requests: #5537
- Please address comment that was mentioned multiple times:
  - https://github.com/vcmi/vcmi/issues/5129#issuecomment-2709644916
  - https://github.com/vcmi/vcmi/pull/5537#pullrequestreview-2675507815

PR #5559, opened 2025-03-14
Randomize optional roads

DjWarmonger:


DjWarmonger:
Needs to be reworked to match original algorithm which deliberately routes roads between towns.

DjWarmonger:
Okay, now it's definitely ready

PR #5565, opened 2025-03-16
support setting dice molecule for rolling morale and luck

kdmcser:
In VCMI, morale and luck only support denominator-based configuration. This creates an issue: when modders want to set high morale or luck values, each additional point has an increasingly larger effect. Initially, the probability increases from 1/10 to 1/9, but at higher values, it jumps from 1/2 to 1. This makes fine-tuning quite inconvenient.

To address this, I made some modifications. In gameConfig.json, I changed the morale and luck-related arrays so that each element is now a two-dimensional array, representing both the numerator and the denominator separately. Additionally, I ensured compatibility with the old configuration format.


kdmcser:
When Morale or Luck is lower than -3 or higher than 3, the corresponding icons will not be displayed in the game if the probabilities are adjusted in gameConfig.json. Expanding the DEF file can resolve values higher than 3, but for values lower than -3, the index becomes negative, making it unworkable. Therefore, I have limited the DEF index range to 0–6. If modders require further extensions in the future, we can improve the image retrieval method accordingly.

IvanSavenko:
I am considering implementing slightly more fair / more human-like random distribution for RNG, and it likely won't work with your approach...

What about something like this in terms of config?
```jsonc
    // Number of sides in virtual dice for morale (fraction denominator)
    "moraleDiceSize" : 24,
    // Chance to roll good morale using morale dice (fraction nominator)
    "goodMoraleChance" : [ 1, 2, 3 ]
    // Chance to roll bad morale using morale dice (fraction nominator)
    "badMoraleChance" : [ 2, 4, 6 ]
```
(and identically for luck)
Since variable was renamed, mods that change it will need an update, but it won't lead to any unexpected effects, since old value is simply unused.

This would allow you to implement morale/luck in range 51%-99%, and won't interfere with my approach to more fair random which would require common denominator for all rolls.

As for icon - can be considered in future. Perhaps after lua support is in? I am somewhat reluctant to allow scripting for game mechanics at the moment, to avoid unpredictable effects like breaking AI, but it might be convenient for UI tasks

kdmcser:
Good idea. I have changed config to that.

PR #5571, opened 2025-03-18
Remove old boost includes and defines

IvanSavenko:
Remove no longer relevant references to boost from Global.h:
- Remove boost::thread leftover defines
- Remove no longer used boost headers
- Move some rarely used boost headers to .cpp's that use them

Should have no effect on supported libboost versions

### DIFF
--- a/lib/VCMIDirs.cpp
+++ b/lib/VCMIDirs.cpp
@@ -132,7 +132,7 @@ bool StartBatchCopyDataProgram(
 		(boost::format("start \"\" /D %1% %2%") % currentPath % (to / exeName));	// Start game in 'currentPath"
 
 	const bfs::path bathFilename = to / "_temp.bat";
-	bfs::ofstream bathFile(bathFilename, bfs::ofstream::trunc | bfs::ofstream::out);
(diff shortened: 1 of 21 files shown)


lib/VCMIDirs.cpp
```diff
@@ -132,7 +132,7 @@ bool StartBatchCopyDataProgram(
 		(boost::format("start \"\" /D %1% %2%") % currentPath % (to / exeName));	// Start game in 'currentPath"
 
 	const bfs::path bathFilename = to / "_temp.bat";
-	bfs::ofstream bathFile(bathFilename, bfs::ofstream::trunc | bfs::ofstream::out);
+	std::ofstream bathFile(bathFilename.c_str(), std::ofstream::trunc | std::ofstream::out);
```
kambala-decapitator, line 135: Are we having a bath here?😄

IvanSavenko: A perfect solution if you're not sure whether to call it `batFile` or `batchFile`!

GeorgeK1ng: Ïs that StartBatchCopyDataProgram really used somewhere these days?

Issue #3738, opened 2024-04-09
Support extending vectors in json without overriding original one,


DjWarmonger:
**Describe your proposal**
Currently, if a base mod adds a vector of objects and another mode overrides it, the contents of first vector are overriden and lost. This prevents us from apending any number of new content by new mods. Example uses:

- Loading screens
- Batllefield backgrounds
- (WIP) Biomes listing adventure map object templates

**Additional context**
Example:

I've got this core configuration:

```
	"swampPalms":{
		"templates" : ["avlswtr0", "avlswtr1", "avlswtr2", "avlswtr3", "avlswtr4", "avlswtr5", "avlswtr6", "avlswtr7", "avlswtr8", "avlswtr9"]
	},
```
In hota.hotaMapDecorations I'd like to add few more graphics to this list:

```
	"core:swampPalms":{
		"templates": ["AVLswn01", "AVLswn02", "AVLswn03"]
	},
```

However, this won't work as expected. Palms with OH3 will be disabled, and only new three can spawn.

Need to write all old and new explicitely:
```
	"core:swampPalms":{
		"templates" : ["avlswtr0", "avlswtr1", "avlswtr2", "avlswtr3", "avlswtr4", "avlswtr5", "avlswtr6", "avlswtr7", "avlswtr8", "avlswtr9", "AVLswn01", "AVLswn02", "AVLswn03"]
	},
```

But this is also not extendable if more mods (or submods) try to append objects to one vector.

thiscris:
Great idea!
Some mods may want to add content, while others may want to replace it. Maybe add a keyword to the parameter to distinguish the two cases. For example `"templates":` becomes `"templates:extend": ["AVLswn01", "AVLswn02", "AVLswn03"]`

IvanSavenko:
We actually already opposite option for structs (aka json objects):
`"someStruct#override" : { ... }`
This should completely replace entire json object when merging data with base mods, instead of merging field-by-field like we usually do.
Probably appending arrays can be implemented in a similar way, e.g. `"someArray#append" : [ ... ]` form

PR #5591, opened 2025-03-27
Support for replacement of individual entries in json lists

IvanSavenko:
- Implements #3738

It is now possible for mods to modify json lists precisely, without full replacement. Supported options:
- `append`: appends a single item to end of list
- `appendItems`: appends multiple items to end of list
- `insert@NUM`: inserts a single item *before* item NUM (item counting is 1-based)
- `modify@NUM`: allows editing of item NUM (item counting is 1-based)

Example - addition of a new item into town hall slots:
```json
"hallSlots":
{
	"modify@4" : {
		"append" : [ "dwellingLvl7B", "dwellingUpLvl7B" ]
	}
},
```
This would modify 4th element (last row) by appending new entry to the end of last row

```json
{
	"modify@4" : {
		"insert@1" : [ "dwellingLvl5B", "dwellingUpLvl5B" ]
	}
},
```
This would add new slot not in the end of last row, but before 1st item (between 5th and 6th dwellings)

Of course, can also be used in all other locations where vcmi uses json lists / vectors, such as allowed terrains in templates.

### DIFF
--- a/docs/modders/Readme.md
+++ b/docs/modders/Readme.md
@@ -134,6 +134,70 @@ Note that modification of existing objects does not requires a dependency on edi
 
 This allows using objects editing not just for rebalancing mods but also to provide compatibility between two different mods or to add interaction between two mods.
 
+### Modifying properties of existing objects
+
+As mentioned above, you can change any properties of existing objects using this form:
+
+```json
+"core:archer" : {
+	"hitPoints" : 10
+},
+```
+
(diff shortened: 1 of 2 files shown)


IvanSavenko:
Adjusted code & docs to use more user-friendly 1-based indexing. And to avoid weird phrasing like "modify@2 will modify third item"

PR #5593, opened 2025-03-28
make OwnerUpdater not hardcode OppositeSideLimiter

kdmcser:
OwnerUpdater is hardcode its limit to OppositeSideLimiter, now make more limiters can be used.

### DIFF
--- a/lib/bonuses/Updaters.h
+++ b/lib/bonuses/Updaters.h
@@ -125,7 +125,6 @@ public:
 class DLL_LINKAGE OwnerUpdater : public IUpdater
 {
 protected:
-	std::mutex mutex;
 	PlayerColor owner;
 
 public:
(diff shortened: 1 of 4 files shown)


lib/bonuses/Updaters.h
```diff
@@ -95,22 +117,27 @@ class DLL_LINKAGE ArmyMovementUpdater : public IUpdater
...
-	std::shared_ptr<Bonus> createUpdatedBonus(const std::shared_ptr<Bonus> & b, const CBonusSystemNode & context) const override;
+	std::shared_ptr<Bonus> createUpdatedBonus(const std::shared_ptr<Bonus> & b, const CBonusSystemNode & context) override;
 	std::string toString() const override;
 	JsonNode toJsonNode() const override;
 };
 
 class DLL_LINKAGE OwnerUpdater : public IUpdater
 {
+protected:
+	std::mutex mutex;
```
IvanSavenko, line 128: Why you need mutex here? 

This code should always be executed while gamestate mutex is locked on client. And server runs in single thread, so no need for multithreaded guards.

kdmcser: ok，the mutex is removed.

PR #5603, opened 2025-04-01
Fix crash when setting up locale on Android

MichalZr6:
Fixes Android crash during locale initialization, for ex. when entering Maps Selection Window.
This adjusts my previous PR: 
- #5436 

which was not tested on Android.


kambala-decapitator:
Would using boost locale instead have saved us from this problem?

IvanSavenko:
> Would using boost locale instead have saved us from this problem?

Maybe? Core problem (Android not having required locale) would remain since apparently library / OS does not provides such locales. Boost locale also relies on std::locale internally, but it might have own error handling instead of throwing exceptions.

kambala-decapitator:
not sure why ubuntu build failed: that error appears only in CMake v4, but the runner still has v3🤔

PR #5609, opened 2025-04-05
End of battle infowindows refactoring

SoundSSGood:
The idea of ​​PR is that information windows would be fully processed on the client side. The server knows nothing about these windows.
Only the end of battle artifact window is ready. If this approach is OK, then I will finish the eagle eye window and the necromancy window later.
Also, all new artifacts that get into the hero's backpack are now added to the end of the backpack (OH3).

### DIFF
--- a/lib/networkPacks/ArtifactLocation.h
+++ b/lib/networkPacks/ArtifactLocation.h
@@ -10,7 +10,6 @@
 #pragma once
 
 #include "../constants/EntityIdentifiers.h"
-#include <optional>
 
 VCMI_LIB_NAMESPACE_BEGIN
 
--- a/lib/networkPacks/ArtifactLocation.h
+++ b/lib/networkPacks/ArtifactLocation.h
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "../constants/EntityIdentifiers.h"
+#include <optional>
 
 VCMI_LIB_NAMESPACE_BEGIN
 
@@ -49,6 +50,33 @@ struct ArtifactLocation
 		h & artHolder;
 		h & slot;
 		h & creature;
+	}
+};
+
+struct MoveArtifactInfo
+{
+	ArtifactPosition srcPos;
+	ArtifactPosition dstPos;
+	bool askAssemble;
+
+	MoveArtifactInfo()
+		: srcPos(ArtifactPosition::PRE_FIRST)
+		, dstPos(ArtifactPosition::PRE_FIRST)
+		, askAssemble(false)
+	{
+	}
+	MoveArtifactInfo(const ArtifactPosition & srcPos, const ArtifactPosition & dstPos, bool askAssemble = false)
+		: srcPos(srcPos)
+		, dstPos(dstPos)
+		, askAssemble(askAssemble)
+	{
+	}
+
+	template <typename Handler> void serialize(Handler & h)
+	{
+		h & srcPos;
+		h & dstPos;
+		h & askAssemble;
 	}
 };
 
--- a/lib/networkPacks/PacksForClientBattle.h
+++ b/lib/networkPacks/PacksForClientBattle.h
@@ -96,8 +96,8 @@ struct DLL_LINKAGE BattleResultAccepted : public CPackForClient
 	struct HeroBattleResults
 	{
 		HeroBattleResults()
-			: armyId(ObjectInstanceID::NONE)
-			, heroId(ObjectInstanceID::NONE)
+			: heroId(ObjectInstanceID::NONE)
+			, armyId(ObjectInstanceID::NONE)
 			, exp(0) {}
 
 		ObjectInstanceID heroId;
@@ -426,7 +426,9 @@ struct DLL_LINKAGE BattleResultsApplied : public CPackForClient
 	BattleID battleID = BattleID::NONE;
 	PlayerColor victor;
 	PlayerColor loser;
+	ChangeSpells learnedSpells;
 	std::vector<BulkMoveArtifacts> artifacts;
+	CStackBasicDescriptor raisedStack;
 	void visitTyped(ICPackVisitor & visitor) override;
 	void applyGs(CGameState *gs) override;
 
@@ -435,7 +437,9 @@ struct DLL_LINKAGE BattleResultsApplied : public CPackForClient
 		h & battleID;
 		h & victor;
 		h & loser;
+		h & learnedSpells;
 		h & artifacts;
+		h & raisedStack;
 		assert(battleID != BattleID::NONE);
 	}
 };
--- a/lib/networkPacks/PacksForClientBattle.h
+++ b/lib/networkPacks/PacksForClientBattle.h
@@ -96,14 +96,18 @@ struct DLL_LINKAGE BattleResultAccepted : public CPackForClient
 	struct HeroBattleResults
 	{
 		HeroBattleResults()
-			: army(ObjectInstanceID::NONE), exp(0) {}
+			: armyId(ObjectInstanceID::NONE)
+			, heroId(ObjectInstanceID::NONE)
+			, exp(0) {}
 
-		ObjectInstanceID army;
+		ObjectInstanceID heroId;
+		ObjectInstanceID armyId;
 		TExpType exp;
 
 		template <typename Handler> void serialize(Handler & h)
 		{
-			h & army;
+			h & armyId;
+			h & heroId;
 			h & exp;
 		}
 	};
@@ -424,7 +428,7 @@ struct DLL_LINKAGE BattleResultsApplied : public CPackForClient
 	PlayerColor loser;
 	std::vector<BulkMoveArtifacts> artifacts;
 	void visitTyped(ICPackVisitor & visitor) override;
-	void applyGs(CGameState *gs) override {}
+	void applyGs(CGameState *gs) override;
 
 	template <typename Handler> void serialize(Handler & h)
 	{
--- a/lib/networkPacks/PacksForClientBattle.h
+++ b/lib/networkPacks/PacksForClientBattle.h
@@ -11,6 +11,7 @@
 
 #include "NetPacksBase.h"
 #include "BattleChanges.h"
+#include "PacksForClient.h"
 #include "../battle/BattleHexArray.h"
 #include "../battle/BattleAction.h"
 #include "../texts/MetaString.h"
@@ -95,15 +96,13 @@ struct DLL_LINKAGE BattleResultAccepted : public CPackForClient
 	struct HeroBattleResults
 	{
 		HeroBattleResults()
-			: hero(nullptr), army(nullptr), exp(0) {}
+			: army(ObjectInstanceID::NONE), exp(0) {}
 
-		CGHeroInstance * hero;
-		CArmedInstance * army;
+		ObjectInstanceID army;
 		TExpType exp;
 
 		template <typename Handler> void serialize(Handler & h)
 		{
-			h & hero;
 			h & army;
 			h & exp;
 		}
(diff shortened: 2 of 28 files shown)


lib/networkPacks/PacksForClientBattle.h
```diff
@@ -112,12 +111,14 @@ struct DLL_LINKAGE BattleResultAccepted : public CPackForClient
 	BattleID battleID = BattleID::NONE;
 	BattleSideArray<HeroBattleResults> heroResult;
 	BattleSide winnerSide;
+	std::vector<BulkMoveArtifacts> artifacts;
```
IvanSavenko, line 114: In which scenarios multiple `BulkMoveArtifacts` packs might be needed? Why single BulkMoveArtifacts pack can't be used?

lib/networkPacks/ArtifactLocation.h
```diff
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "../constants/EntityIdentifiers.h"
+#include <optional>
```
IvanSavenko, line 13: Move to global.h, if it is not there already

IvanSavenko:
> If this approach is OK, then I will finish the eagle eye window and the necromancy window later.

Probably OK? Don't see any significant need for such change though - it won't fix any existing problems / bugs for example.

lib/networkPacks/PacksForClientBattle.h
SoundSSGood: Duplication is necessary while there are two network packets BattleResultAccepted and BattleResultsApplied. On the server side, we transfer artifacts while both heroes exist. And on the client side, after the left-hand side. After finishing work with the necromancy window and the eagle eye, there will be only one packet left.

SoundSSGood:
This will simplify the code on both the client and server sides. It will also reduce the number of network packets after the battle to two. Which will reduce lags in the network game.
In addition, this will make the process of finishing the battle linear and remove loops during multi-level up:
level up query -> sending to client -> client response -> deleting the query -> level up query ...

IvanSavenko:
> It will also reduce the number of network packets after the battle to two. Which will reduce lags in the network game.

Lag in MP does not depends on number of packets (unless server waits for client to reply on 1st packet)

> In addition, this will make the process of finishing the battle linear and remove loops during multi-level up

How? We *have* to wait for result of 1st levelup since skills that are available for selection on 2nd levelup depend on what player picks in 1st levelup.

SoundSSGood:
> How? We _have_ to wait for result of 1st levelup since skills that are available for selection on 2nd levelup depend on what player picks in 1st levelup.

Yes, that will be a problem.

server/battles/BattleResultProcessor.h
```diff
@@ -46,19 +47,21 @@ struct FinishingBattleHelper
 
 	inline bool isDraw() const {return winnerSide == BattleSide::NONE;}
 
-	const CGHeroInstance *winnerHero, *loserHero;
+	ObjectInstanceID winnerId = ObjectInstanceID::NONE, loserId = ObjectInstanceID::NONE;
```
IvanSavenko, line 50: - 1 declaration per line
- identifiers have default constructor that would initialize them to NONE

```suggestion
	ObjectInstanceID winnerId;
	ObjectInstanceID loserId;
```

server/battles/BattleResultProcessor.cpp
```diff
@@ -188,15 +188,15 @@ FinishingBattleHelper::FinishingBattleHelper(const CBattleInfoCallback & info, c
 {
 	if (result.winner == BattleSide::ATTACKER)
 	{
-		winnerHero = info.getBattle()->getSideHero(BattleSide::ATTACKER);
-		loserHero = info.getBattle()->getSideHero(BattleSide::DEFENDER);
+		winnerId = info.getBattle()->getSideArmy(BattleSide::ATTACKER)->id;
+		loserId = info.getBattle()->getSideArmy(BattleSide::DEFENDER)->id;
```
IvanSavenko, line 192: I think there might be cases where winner army != winner hero - like town sieges. Or not?

server/battles/BattleResultProcessor.cpp
```diff
@@ -301,53 +301,45 @@ void BattleResultProcessor::endBattle(const CBattleInfoCallback & battle)
...
 {
-	auto battleQuery = std::dynamic_pointer_cast<CBattleQuery>(gameHandler->queries->topQuery(battle.sideToPlayer(BattleSide::ATTACKER)));
-	if(!battleQuery)
-		battleQuery = std::dynamic_pointer_cast<CBattleQuery>(gameHandler->queries->topQuery(battle.sideToPlayer(BattleSide::DEFENDER)));
-	if(!battleQuery)
-	{
-		logGlobal->trace("No battle query, battle end was confirmed by another player");
-		return;
-	}
-
```
IvanSavenko, line 312: Please check replaying autocombat against AI heroes. There were quite a lot of bugs related to those.

IvanSavenko:
In any case - not against these changes.

Don't think that this is something high-priority, or something that would help with existing issues (which we have quite a lot), but don't see these changes leading to any potential problems.

As far as hero levelups go - might be a good idea to investigate what's happening during that period, to see if there is way to reduce the lag.
For example, there is rather smelly `dialogs` member in `CPlayerInterface`, that is handled in similarly smelly `CPlayerInterface::update` method. Why we have it that way instead of handling it immediately when previous dialog is closed? Our logic where network thread waits for dialogs to be processed might also contribute to this delay.
Similar (and probably easier to investigate) happens when there are multiple messages on game start, for example on All for One map - you can clearly see that when you close first dialog on map start, game would first redraw map, and only after that 1 frame delay - show 2nd dialog. Why that delay even exists? Fixing it might also help with delay on multiple levelups.

SoundSSGood:
> As far as hero levelups go - might be a good idea to investigate what's happening during that period, to see if there is way to reduce the lag. For example, there is rather smelly `dialogs` member in `CPlayerInterface`

Thanx. Will look into.
Now I see that the PR is not complete. For now I have moved it to drafts. Until a better solution is found.

SoundSSGood:
Finished what I had planned, the necromancy and eagle eye windows. All convrsations have been resolved. Since this is no longer a small PR, I think CPlayerInterface::dialogs will be in the next one.

> Don't think that this is something high-priority, or something that would help with existing issues (which we have quite a lot), but don't see these changes leading to any potential problems.

Now it's just a refactoring PR.

SoundSSGood:
> As far as hero levelups go - might be a good idea to investigate what's happening during that period, to see if there is way to reduce the lag. For example, there is rather smelly `dialogs` member in `CPlayerInterface`, that is handled in similarly smelly `CPlayerInterface::update` method.

@IvanSavenko  `CPlayerInterface::update` is executed in the main loop, and I don't see any reason for it to cause lag. As for the implemented dialog queue mechanism, I would keep it. However, its implementation is not complete, as some dialog windows currently bypass the queue (e.g., `CInfoWindow::showYesNoDialog`). This shouldn't happen except in certain cases. I can take care of this. If you have any other ideas on this matter, feel free to share them.

IvanSavenko:
>  CPlayerInterface::update is executed in the main loop, and I don't see any reason for it to cause lag

Then why there is clearly 1 frame delay between displaying two dialogs on map start? 
Could be simply incorrect order of operations, for example, game checks for dialogs before processing input, so order is:
- check for new dialog
- process input and close current dialog
- redraw (with current dialog closed, but next one not opened yet)
- wait for next frame

If this theory is correct then moving input processing to occur before check for new dialog should be enough. But could also be something else.

> However, its implementation is not complete, as some dialog windows currently bypass the queue

What about levelup dialog, which initiated this investigation? 

In theory, for singleplayer games it should be like instant. 16 ms between 2 frames is more than enough for client to send confirmation pack, for server to receive it, send reply back to client, and for client to process incoming pack with next levelup. From what I've seen, single netpack client->server->client processing happens in like 1-2 ms

But there is clear delay, which feel to be way longer than 1 frame. What's the reason behind that delay?

SoundSSGood:
> But there is clear delay, which feel to be way longer than 1 frame. What's the reason behind that delay?

Although `heroGotLevel` is not added to the dialog queue, it does contain a check for the presence of dialogs in the queue — there's a cond.wait in there. 
Just for clarity: by "clear delay," do you mean the delay that occurs every time? In my case, the delay appears randomly and not consistently. 
So I assume there are only two possible reasons (my case):

- it's related to the loopback interface of localhost;
- the OS scheduler.

IvanSavenko:
Checked those 3 info window on start of "One for All" map.

This change have fixed flickering that I could consistently see between two dialogs. Feel free to try it yourself and compare before / after.
```diff
diff --git a/client/GameEngine.cpp b/client/GameEngine.cpp
index 30866a60a..ea2ed7120 100644
--- a/client/GameEngine.cpp
+++ b/client/GameEngine.cpp
@@ -120,9 +120,8 @@ void GameEngine::updateFrame()
 {
        std::scoped_lock interfaceLock(ENGINE->interfaceMutex);
 
-       engineUser->onUpdate();
-
        handleEvents();
+       engineUser->onUpdate();
        windows().simpleRedraw();
 
        if (settings["video"]["showfps"].Bool())
```

Looking into levelup dialogs

> Just for clarity: by "clear delay," do you mean the delay that occurs every time? In my case, the delay appears randomly and not consistently.

Yes. For me it is either guaranteed or almost guaranteed. Let's say 90%-100% chance. Sometimes delay is small-ish (1 frame / 16 ms probably?). But sometimes it can be longer, 2-3 frames (up to 50 ms?)

IvanSavenko:
For me there is huge 30 ms delay between closing previous levelup window and game opening new one.

Sure, thread juggling does have some overhead. But it is not 30 ms. For example:
```
[2025-Apr-13 23:52:59.087508] TRACE [MainGUI] network - Sending a pack of type 10QueryReply
[2025-Apr-13 23:52:59.087593] TRACE [runServer] network - Received CPack of type 10QueryReply
```
0.085 ms between GUI sending reply with selected skill and server receiving it. We can switch between threads 100 times and it would still take less time than we spend now

Where have we lost those 30 ms? Sorry, this can NOT be OS scheduler, unless system already at 100% CPU load. And it can't be loopback overhead. Even with loopback network delay would be way lower, and right now we use even faster intra-process pseudo-connection using boost::asio

IvanSavenko:
Looks like multiple problems that interact in a bad way. One of these on their own would not have been notable, but they stack on each other, resulting in delay.

1. `WindowHandler::totalRedraw` is called whenever window is closed. This is quite a slow call, and since it is done with locked UI mutex, network thread has to wait for processing.
2. Network thread locks & unlocks UI mutex for each received pack. And levelup generates 3 packs, and on each of those, main GUI thread can take over UI mutex increasing network processing delay
3. Loading background for levelup window is slow. 10 ms. Not sure why it is *that* long, file should be cached by OS. Probably worked better in 1.6.0 -> 1.6.7 when excessive image caching was present. Probably same problem with `AdventureMapInterface::onHeroChanged`, which takes 4 ms

So...
- Finishing hardware acceleration should help, since total redraw won't be that slow
- Fixing image caching to limit it to reasonable size would allow us to cache levelup window background and skip on window creation overhead.
- Merging those 3 packs into 1 would help a bit, but this is not the only place that needs to process multiple packs at once, so might require a lot of work
- Changing how network thread runs to avoid multiple locks/unlocks might help. Currently we have it like this:
  ```
  CServerHandler::threadRunNetwork()
  -> boost::asio::io_service::run()
    -> CServerHandler::onPacketReceived
      -> std::scoped_lock()
  ```
  In theory, if we were to somehow fetch any remaining packs inside `onPacketReceived` call without releasing mutex, delay should be way lower and overhead from sending multiple packs would be close to 0. But right now have no idea how to do this nicely...

SoundSSGood:
> * Merging those 3 packs into 1 would help a bit, but this is not the only place that needs to process multiple packs at once, so might require a lot of work

At the moment, I can help with this at most.
So, what do we decide with this PR? Delete it or leave it?

IvanSavenko:
> At the moment, I can help with this at most.

Don't see much point here. It would only fix this particular scenario and here it is less logical than in battles since most of packs are optional. Levelup may be followed by another levelup. Or it might be a single level. Hero may receive secondary skill or it might not, in case of high-level heroes.

Probably better to resolve it globally.

> So, what do we decide with this PR? Delete it or keep it?

Generally looks OK, will review rest of the code and merge.

PR #5613, opened 2025-04-08
Allow Expert skill removal

GeorgeK1ng:
Fix for Expert skill removal. Now will be correctly removed.

### DIFF
--- a/lib/rewardable/Interface.cpp
+++ b/lib/rewardable/Interface.cpp
@@ -108,7 +108,7 @@ void Rewardable::Interface::grantRewardBeforeLevelup(const Rewardable::VisitInfo
 	for(const auto & entry : info.reward.secondary)
 	{
 		auto currentLevel = static_cast<MasteryLevel::Type>(hero->getSecSkillLevel(entry.first));
-		if(currentLevel == MasteryLevel::EXPERT)
+		if(currentLevel == MasteryLevel::EXPERT && entry.second != 0)
 			continue;
 
 		if(currentLevel != MasteryLevel::NONE || hero->canLearnSkill())


lib/rewardable/Interface.cpp
```diff
@@ -108,7 +108,7 @@ void Rewardable::Interface::grantRewardBeforeLevelup(const Rewardable::VisitInfo
 	for(const auto & entry : info.reward.secondary)
 	{
 		auto currentLevel = static_cast<MasteryLevel::Type>(hero->getSecSkillLevel(entry.first));
-		if(currentLevel == MasteryLevel::EXPERT)
+		if(currentLevel == MasteryLevel::EXPERT && entry.second != 0)
```
IvanSavenko, line 111: Are you sure?
If you're trying to remove skill, value would be negative (so != 0). And this condition would still trigger.

I think you want to check whether it is above zero?
```suggestion
		if(currentLevel == MasteryLevel::EXPERT && entry.second >= 0)
```

GeorgeK1ng: I have tracked it exactly to this point. Skill removal is done by val 0. I was confused too, but when skill level is already Expert, it skipped whole Change skill / Set skill. When added this skill removal worked for Expert level.

IvanSavenko: Something is wrong.
```cb->changeSecSkill(hero, entry.first, entry.second, false);```

Last parameter (false) is named `abs` (for absolute).
Idea is that:
- if abs = false change is relative (so +1 would increase skill by 1 level, and 0 would do nothing)
- if abs = true, change is absolute (so 1 would set skill to basic, and 0 would remove skill)

GeorgeK1ng: Don't have any other ideas about that. In my initial test I have commented out that block and then was removal fine.. 
Problem was specially only on Expert level. When skill level was 1 or 2, removing using Market of Time worked. When it was 3 then new level 0 does nothing.

IvanSavenko: Sounds like bug is someplace else and this PR introduces different bug to compensate for existing bug

GeorgeK1ng:
I have no idea what else should be wrong. Just know skill level value is 0 when removing it. And it was not removed when skill level was already on Expert level.

PR #5615, opened 2025-04-10
Fixes to locale/unicode handling

### LINKED ISSUE #5519: when vampire kings are resurrecting, crash 100%
1. some vampire kings was killed
2. vampire kings attacked and some are resurrecting
3. crash

Console error report:
vcmiclient: /var/tmp/portage/games-strategy/vcmi-9999/work/vcmi-9999/lib/texts/TextOperations.cpp:121: uint32_t TextOperations::getUnicodeCodepoint(const char*, size_t): Assertion `isValidUnicodeCharacter(data, maxSize)' failed.

I use Gentoo amd64, with develop version. It happens every time and every builds. I disabled some mods, but nothing changed

[8xm8pp.txt](https://github.com/user-attachments/files/19144383/8xm8pp.txt)

[VCMI_Server_log.txt](https://github.com/user-attachments/files/19144384/VCMI_Server_log.txt)

IvanSavenko:
- Fixes crash on opening spellbook (recent regression)
- Fix crash if `haystack` string is shorter than `needle`
- Levenstein distance computation is now case-insensitive
- Text similarity computation now assumes utf-8 strings, not ascii
- Fix corruption of utf-8 string on life drain (fixes #5519)

### DIFF
--- a/lib/texts/TextOperations.cpp
+++ b/lib/texts/TextOperations.cpp
@@ -17,7 +17,7 @@
 
 #include <vstd/DateUtils.h>
 
-#include <boost/locale.hpp>
+#include <boost/locale/encoding.hpp>
 
 VCMI_LIB_NAMESPACE_BEGIN
 
@@ -97,7 +97,7 @@ bool TextOperations::isValidASCII(const char * data, size_t size)
 	return true;
 }
 
-bool TextOperations::isValidUnicodeString(const std::string & text)
+bool TextOperations::isValidUnicodeString(std::string_view text)
 {
 	for (size_t i=0; i<text.size(); i += getUnicodeCharacterSize(text[i]))
 	{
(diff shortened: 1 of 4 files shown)


lib/texts/TextOperations.cpp
```diff
@@ -253,10 +253,62 @@ std::string TextOperations::getCurrentFormattedDateTimeLocal(std::chrono::second
...
+
+	auto charactersEqual = [&s, &t](int sPoint, int tPoint)
+	{
+		uint32_t sUTF32 = getUnicodeCodepoint(s.data() + sPoint, s.size() - sPoint);
+		uint32_t tUTF32 = getUnicodeCodepoint(t.data() + tPoint, t.size() - tPoint);
+
+		if (sUTF32 == tUTF32)
+			return true;
+
+		// Windows - wchar_t represents UTF-16 symbol which does not covers entire unicode
```
kambala-decapitator, line 294:
```suggestion
		// Windows - wchar_t represents UTF-16 symbol which does not cover entire unicode
```

Issue #5534, opened 2025-03-10
Health bar graphical (?) bug.


misiokles:
**Describe the bug**
Ghosts Dragons can age attacked units (reduces HPs). Aged creatures have unappriopiate long health indicator.

![Image](https://github.com/user-attachments/assets/f4face4f-b83a-4c2f-b304-3f859459769c)

**Expected behavior**
Health bar should be scaled in this case.

**Did it work earlier?**
It's relatively new feature.

**Version**

- OS: Windows
- Version: 1.6.7

PR #5619, opened 2025-04-10
fix health bar for commander

Laserlicht:
Fixes bug reported in #5534

@misiokles Can you check if both issues are fixed? (commander & ghost dragon)

### DIFF
--- a/client/battle/BattleStacksController.cpp
+++ b/client/battle/BattleStacksController.cpp
@@ -321,7 +321,7 @@ void BattleStacksController::showStackAmountBox(Canvas & canvas, const CStack * 
 	{
 		double healthMaxType = stack->unitType()->getMaxHealth();
 		double healthMaxStack = stack->getMaxHealth();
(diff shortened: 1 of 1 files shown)


client/battle/BattleStacksController.cpp
```diff
@@ -321,7 +321,7 @@ void BattleStacksController::showStackAmountBox(Canvas & canvas, const CStack *
 	{
 		double healthMaxType = stack->unitType()->getMaxHealth();
 		double healthMaxStack = stack->getMaxHealth();
-		double healthMaxRatio = healthMaxStack / healthMaxType;
+		double healthMaxRatio = std::min(healthMaxStack / healthMaxType, 1.0);
```
IvanSavenko, line 324: Wouldn't such a fix lead to a similar kind of bug - where a half-dead unit still has a full health bar?

misiokles:
I'll test in next week.

misiokles:
Tested. All seem fine. Status health bar is updated correctly both for Ghost Dragons and Commanders.

PR #5623, opened 2025-04-13
do lupdate and update Chinese translation

kdmcser:


GeorgeK1ng:
Not sure if it's good idea right now. There are pending some map editor PRs which will add tons of new strings.

kdmcser:
OK, left it when things is ready.

PR #5611, opened 2025-04-06
Allow changing default directories in Windows

GeorgeK1ng:
Added support for overriding default hardcoded directories on Windows using a single `dirs.json` file, which must be placed in the standard `config` directory (alongside with other VCMI configuration files).  
If the file is missing, the engine falls back to the original hardcoded paths.

Environment variables like `%USERPROFILE%` are fully supported and automatically expanded.

 Example of full `dirs.json`

```json
{
    "userDataPath": "%USERPROFILE%\\Documents\\My Games\\vcmi",
    "userCachePath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\cache",
    "userConfigPath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\config",
    "userLogsPath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\logs",
    "userSavePath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\saves"
}
```

You can override any subset of paths – unspecified entries will automatically fall back to their default locations.

- Allows full redirection of previously hardcoded folders to any custom location.
- Makes VCMI fully portable on Windows – no longer depends on the user profile's "My Documents" folder.
- Enables storing the `Saves` folder (or any other) in a synchronized location (e.g., OneDrive, Dropbox) to sync between devices.
- Only affects behavior when `dirs.json` is present.

-  Windows is fully implemented and tested.
- Other platforms not yet implemented. Help is welcome from contributors familiar with macOS/Linux paths.

- Removed legacy migration logic for obsolete folder structures.
- Tested clean launch behavior – no issues observed, paths are correctly redirected.
- Currently supports only **manual creation** of `dirs.json` in the config folder.
- Future idea is to allow configuring paths directly via the VCMI launcher UI.

PR #5624, opened 2025-04-13
add predefine DocumentPath placeholder

kdmcser:
In #5611 ,it's suggested to set `dirs.json` as follows:

```
{
    "userDataPath": "%USERPROFILE%\\Documents\\My Games\\vcmi",
    "userCachePath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\cache",
    "userConfigPath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\config",
    "userLogsPath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\logs",
    "userSavePath": "%USERPROFILE%\\Documents\\My Games\\vcmi\\saves"
}
```

However, `%USERPROFILE%\\Documents\\` is not always the actual path to the user's My Documents folder. On Windows, users can change the location of My Documents, for example to `D:\My Document`.

Maybe we could introduce a predefined placeholder, such as `%MyDocuments%`, which VCMI can resolve at runtime to the correct path.

And function `SHGetSpecialFolderPathW` is Deprecated. Should replaced by `SHGetFolderPathW`.


GeorgeK1ng:
That's just example of dirs.json. VCMI itself don't use it until it's created. 

If you moved your directory, use hardcoded path. And if you did this move correctly, then it will be still loaded from CSIDL_PERSONAL without any change. It's connected with Shell Folders in registry.

And if is end user power user, then it's enough to add custom variable into PATH, then it will be correctly handled, no need to hardcode some kind of supported variables. That's why it have expand variables feature inside.

Btw mentioned function is also depricated and supported one is SHGetKnownFolderPath. But that doesn't matter at all. All 3 have same results.

I will try to add GUI that allows changing these folders into launcher soon.

kdmcser:
This feature is intended for mod pack integrators. They need to distribute their own mod packages to users and want those users to use a separate directory to avoid conflicts with other mod packs. However, they cannot assume the location of each user's "My Documents" folder—some users have changed it, while others have not. Therefore, they need VCMI to provide a unified way to retrieve the "My Documents" path, rather than relying on users to configure it themselves. Some users don't even know what an environment variable is. By using a built-in placeholder, their mod packages become much easier to use.

GeorgeK1ng:
What about OneDrive syncing? Keeping game data it in documents folder will still have same result in modern Windows. All files will goes online. That was one of case why was added support to change files location. 

Better is to use %AppData%, but then on server synced profile will payload there. 

So there will be always some kind of syncing issue. 

You can also use your portable VCMI directory, in dirs.json you can just add something like “vcmidata” into userDataPath and then will be used VCMI “application root\vcmidata”. Also if you don’t need to change other paths, you don’t need to specifiy them, userDataPath is enough.

kdmcser:
In China, no one uses OneDrive. Additionally, maintaining document storage remains necessary. If users encounter issues, paths like %APPDATA% and %LOCALAPPDATA% are completely inaccessible to novice users - they can't cooperate when troubleshooting is required. If configured in the game installation directory, this data faces accidental deletion risks. Users habitually delete entire game directories when reinstalling new versions, leading to save file loss.

I've changed the placeholder names to %USERDOCUMENTS% and %MyDocument% as recommended by DeepSeek, but upon reflection, these don't seem appropriate either.

IvanSavenko:
> In China, no one uses OneDrive. 

Everyone else seems to use it, even though a lot of people are probably not aware of that since it is forced by MS nowadays.

> Additionally, maintaining document storage remains necessary. If users encounter issues, paths like %APPDATA% and %LOCALAPPDATA% are completely inaccessible to novice users

Those locations can be easily opened via Launcher

kdmcser:
```
	std::ifstream in(pathToUtf8(configPath), std::ios::binary);
	if (!in)
		return;

	std::string buffer((std::istreambuf_iterator<char>(in)), {});
	dirsConfig = std::make_unique<JsonNode>(reinterpret_cast<const std::byte*>(buffer.data()), buffer.size(), pathToUtf8(configPath));
```
This code will lead a problem: some people installed vcmi in A Chinese path. Their Windows use GBK, converting to UTF8 will make VCMI unable to find the real GBK path, and dirsConfig will remain empty.

kdmcser:
This problem is fixed, but I found another encoding problem #5634,.

kambala-decapitator:
@kdmcser but what is the explanation regarding WideCharToMultibyte not working with GBK?

kambala-decapitator:
> When the resulting UTF-8 string is passed to std::ifstream (which expects ANSI-encoded paths on Windows)

Thanks, forgot about this

GeorgeK1ng:
This should be merged too to avoid issues in upcoming installer update with ability to select custom data folders locations. This PR should fix chinese characters in dirs.json.

Issue #3052, opened 2023-10-14
Text in InfoPanel not vertically centered


krs0:
![image](https://github.com/vcmi/vcmi/assets/15194321/b2209dca-99f5-47c9-b70a-7d4f5788380f)

PR #5628, opened 2025-04-13
Info window align fix

GeorgeK1ng:
Fixes align of text in both info windows

image

Fix for: https://github.com/vcmi/vcmi/issues/3052

### DIFF
--- a/client/widgets/MiscWidgets.cpp
+++ b/client/widgets/MiscWidgets.cpp
@@ -335,8 +335,8 @@ void CHeroTooltip::init(const InfoAboutHero & hero)
 
 		labels.push_back(std::make_shared<CLabel>(158, 100, FONT_TINY, ETextAlignment::CENTER, Colors::WHITE, std::to_string(hero.details->mana)));
 
-		morale = std::make_shared<CAnimImage>(AnimationPath::builtin("IMRL22"), hero.details->morale + 3, 0, 5, 74);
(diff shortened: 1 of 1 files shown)


client/widgets/MiscWidgets.cpp
```diff
@@ -261,7 +261,7 @@ void CArmyTooltip::init(const InfoAboutArmy &army)
 {
 	OBJECT_CONSTRUCTION;
 
-	title = std::make_shared<CLabel>(66, 2, FONT_SMALL, ETextAlignment::TOPLEFT, Colors::WHITE, army.name);
+	title = std::make_shared<CLabel>(66, 3, FONT_SMALL, ETextAlignment::TOPLEFT, Colors::WHITE, army.name);
```
IvanSavenko, line 264: Might be better to switch to CENTERLEFT alignment or to BOTTOMLEFT and adjust coordinates.

Your solution would fix fonts in your particular version of game with your particular fonts (English gog?), but might not work in other versions, including other languages from gog

GeorgeK1ng: How exactly works these alignments? I was thinking now with center set and max width it will be rendered from coordinates to both sides. It looks like it works like that from 7 vs 13 for example, both seems to be centered.

Also mentioned change from 2 to 3, I tried too, but it then overlap from bottom. These two windows looks same, but coordinates are moved from some reason. And probably by 1 exactly. No idea why.

I have tried English, Czech and Russian translation mods. Can check also others.

PR #5629, opened 2025-04-14
Fix Lord Haart vs Sir Munlich

GeorgeK1ng:
In original game from GOG is available for random maps Lord Haart not Sir Munlich. Their swap is directly HotA purpose.

Fix for: https://github.com/vcmi/vcmi/issues/2340

### DIFF
--- a/config/heroes/castle.json
+++ b/config/heroes/castle.json
@@ -78,7 +78,6 @@
 		"index": 4,
 		"class" : "knight",
 		"female": false,
-		"special" : true, // Lord Haart in his living form. Disabled in H3 expansions
 		"skills":
 		[
 			{ "skill" : "leadership", "level": "basic" },
--- a/config/heroes/special.json
+++ b/config/heroes/special.json
@@ -5,7 +5,7 @@
 		"index": 144,
 		"class" : "knight",
 		"female": false,
-		"special" : false, // available in single scenario, replacement for no longer available Lord Haart
+		"special" : true,
 		"skills":
 		[
 			{ "skill" : "leadership", "level": "advanced" }


Olegmods:
> In original game from GOG is available for random maps Lord Haart not Sir Munlich. Their swap is directly HotA purpose.
> 
> Fix for: #2340

Erm... In original game (AB, SoD) Sir Mullich actually replaces Lord Haart. He was available only in RoE (since he was dead during AB and was there as Haart Lich). See Appearance part here: https://heroes.thelazy.net/index.php/Lord_Haart . HotA replaced Mullich with Haart due to the specialty of the former being kinda OP.

And the issue you pointed is about Mullich being banned, which (according to the configs) shouldn't happen without HotA mod. In this PR you ban Mullich and unban Haart, which is the opposite of what should be done to fix that issue.

Olegmods:
> Install GOG H3 (No HD mod addition) and try to select random map - Castle, there is just Lord Haart. Main VCMI configs should respect original settings. Configured scenarios should not be affected with this change.

Well, again, Heroes 3 removed Haart from the main pool of heroes after AB. See Appearance part here: https://heroes.thelazy.net/index.php/Lord_Haart .
Also, citing another source I found:

> In Armageddon's Blade and The Shadow of Death, Lord Haart is withdrawn from the Knight hero pool and moved to the Necropolis faction as a campaign-only Death Knight. The Knight slot he leaves empty is filled by Sir Mullich.

Not sure what GoG has done, probably it's smth with the last patch (which they've made themselves), but Lord Haart is supposed to be banned. Currently didn't check, but all info I can find says that it should be banned. And Haart is unbanned in HotA only.

Citing H3 wiki:

> In Horn of the Abyss, Lord Haart is available as a knight and replaces Sir Mullich by default (Lord Haart the Death Knight is renamed to Haart Lich).

But nothing about Haart being available outside of RoE and HotA.

GeorgeK1ng:
Same is in 25 years old Heroes 3 Complete.

Random map -> Castle ->Lord Haart
Selecting maps -> Castle -> Sir Mullich

There is probably different issue, and it's about special flag, as it currently bans it completely for scenario selection too, not only for Random Maps.

IvanSavenko:
> Random map -> Castle ->Lord Haart
Selecting maps -> Castle -> Sir Mullich

Not exactly.

Random map -> Lord Haart
SoD map -> Sir Mullich
RoE map -> Lord Haart

So change looks correct:

- RoE map will always have Lord Haart (unless banned via map editor), and Sir Mullich availability would be based on special flag (so banned)
- SoD map will use settings set in .h3m for both Lord Haart and Sir Mullich (so in most cases - Lord Haart banned, Sir Mullich available)
- Random maps will use those heroes based on special flag (so Mullich banned, Haart available)

Olegmods:
> Same is in 25 years old Heroes 3 Complete.
> 
> Random map -> Castle ->Lord Haart Selecting maps -> Castle -> Sir Mullich
> 
> There is probably different issue, and it's about special flag, as it currently bans it completely for scenario selection too, not only for Random Maps.

Hmm... That's a bit strange. Can you check if the same happens in SoD?

Also, I'd suggest to remove mention of #2340 in the first post for the PR. The issue itself (Mullich being banned by default) is probably fixed as for now, and banning Mullich again certainly doesn't fix that issue.

And just FYI: swapping Haart with Mullich isn't a HotA change, they've done the opposite.

Olegmods:
> > Random map -> Castle ->Lord Haart
> > Selecting maps -> Castle -> Sir Mullich
> 
> Not exactly.
> 
> Random map -> Lord Haart SoD map -> Sir Mullich RoE map -> Lord Haart
> 
> So change looks correct:
> 
> * RoE map will always have Lord Haart (unless banned via map editor), and Sir Mullich availability would be based on special flag (so banned)
> * SoD map will use settings set in .h3m for both Lord Haart and Sir Mullich (so in most cases - Lord Haart banned, Sir Mullich available)
> * Random maps will use those heroes based on special flag (so Mullich banned, Haart available)

Hmmm... So Haart wasn't replaced entirely? I've seen some accounts about the Haart's id in H3 being the same as the Mullich one.
Also, does Mullich use the special flag in vanilla? Not sure about it tbh.

PR #5641, opened 2025-04-19
fix Chinese encoding problem in GBK

### LINKED ISSUE #5634: Chinese characters in map directory names and file paths are displayed as gibberish due to encoding issues
**Describe the bug**
On Chinese versions of Windows, two system encodings can be used: Unicode (UTF-8) and GBK. Unicode is currently a beta feature, and most users leave it disabled. When Unicode is enabled, everything displays correctly in-game. However, when it's disabled—which is the typical setting for most users—although maps load fine and the game runs normally, file and directory names appear as garbled text.

A similar issue also exists in ERA, but with a twist: ERA works fine under GBK and only shows problems when Unicode is enabled. If a Chinese player happens to be using both ERA and VCMI, it's bound to be a frustrating experience.

I've looked through the code several times, but due to the heavy nesting, I'm not sure where to start. Would anyone familiar with this part of the code be willing to take a look or offer some guidance? I’d really appreciate it—thanks in advance!

Image
Image

**Game logs**

[VCMI_Client_log.zip](https://github.com/user-attachments/files/19780509/VCMI_Client_log.zip)

**Version**

- OS: Windows
- Version: 1.7

kdmcser:
As mentioned in #5634, Chinese characters may display as garbled text in certain situations.
This commit addresses the issue following the approach suggested by @IvanSavenko , and the fix has been tested under both GBK and Unicode environments. No garbled output or read/write failures were observed.

Fixes #5634

### DIFF
--- a/lib/filesystem/CFilesystemLoader.cpp
+++ b/lib/filesystem/CFilesystemLoader.cpp
@@ -14,6 +14,8 @@
 
 #include "../ExceptionsCommon.h"
 
+#include <boost/locale/encoding_utf.hpp>
+
 VCMI_LIB_NAMESPACE_BEGIN
 
 CFilesystemLoader::CFilesystemLoader(std::string _mountPoint, boost::filesystem::path baseDirectory, size_t depth, bool initial):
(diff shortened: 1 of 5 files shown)


lib/filesystem/CFilesystemLoader.cpp
```diff
@@ -91,20 +93,21 @@ bool CFilesystemLoader::createResource(const std::string & requestedFilename, bo
 	}
 
 	filename = filename.substr(mountPoint.size());
+	std::wstring filePath = boost::locale::conv::utf_to_utf<wchar_t>(filename);
```
IvanSavenko, line 96:
```suggestion
#ifdef VCMI_WINDOWS
			boost::filesystem::path filePath(boost::locale::conv::utf_to_utf<wchar_t>(filename));
#else
			boost::filesystem::path filePath(filename.string());
#endif
```
```

File: lib/filesystem/CFilesystemLoader.cpp [hunk 2]
```diff
@@ -173,19 +176,26 @@ std::unordered_map<ResourcePath, boost::filesystem::path> CFilesystemLoader::lis
 				filename = it->path().filename();
 
 			std::string resName;
+
+#ifdef VCMI_WINDOWS
+			std::string filenameUtf8 = boost::locale::conv::utf_to_utf<char>(filename.native());
+#else
+			std::string filenameUtf8 = filename.string();
+#endif
+
```
IvanSavenko, line 185: Already copy-pasted once, and it's possible that there are other, not yet found cases.
Perhaps move to TextOperations file?
```cpp
std::string TextOperations::filesystemPathToUtf8(const boost::filesystem::path & path)
{
#ifdef VCMI_WINDOWS
	return boost::locale::conv::utf_to_utf<char>(filename.native());
#else
	return filename.string();
#endif
}
```

And possibly do the same for the opposite conversion - from utf8 string to filesystem path

PR #5651, opened 2025-04-26
fix crash in AI's turn

kdmcser:
In some cases, `ai->cb->getAvailableHeroes(town)` may return NULL, leading to a crash during the AI's turn. This fix addresses that issue. I'm not sure why this situation occurs, but defensive programming is necessary to handle potential NULL returns and prevent crashes.


IvanSavenko:
This is hiding bug someplace else.

From what I see, the only way for new element to be added to `currentTavern` is `TavernHeroesPool::setHeroForPlayer` method. 

That method does have suspicious checks for nullptr, however those checks are from like 10-years old code, and should be completely redundant since there is check for invalid hero ID at the very start of the method.

If this bug is reproducible for you - you can try placing checks/logging into that method: `CGHeroInstance * h = heroesPool[hero];` should never result in null pointer.

kdmcser:
But I have a problem: if only one hero or no hero can be hired, what will happend in tavern?

PR #5652, opened 2025-04-26
Prevent AI from repeatedly swapping the same artifacts

kdmcser:
This is the most basic solution to prevent the issue described in #5652, but it does not fully fix it. The root cause of the problem lies in how artifact scores are calculated. For example, two ring artifacts—one granting +1 Knowledge and +1 Mana regeneration, and the other granting +10 Mana regeneration—can cause the AI to loop endlessly exchanging them.

Moreover, even if we fix the current scoring logic, there is no guarantee that similar issues won't arise again in the future. Therefore, adding a safeguard is a more reliable approach.


kdmcser:
Compared to simply limiting the maximum number of swaps, tracking the history of trades and preventing previously swapped artifacts from being exchanged again seems like a more elegant solution.

If Artifact A replaces Artifact B, then A should not be allowed to replace B again, and B should not be allowed to replace A either.
This approach also effectively handles circular swaps, such as A replacing B, B replacing C, and C replacing A.

Therefore, I have removed the code that limited the maximum number of swaps and replaced it with a mechanism that disallows repeated exchanges.

PR #5659, opened 2025-04-29
Reorganized artifact-related classes

IvanSavenko:
- files now generally contain only 1 class (except for tightly coupled classes)
- files are now located in `lib/entities/artifact` directory
- removed excessive includes
- moved artifact class enum from CArtifact class to global scope

No changes to functionality


Laserlicht:
@IvanSavenko 
Seems to cause some regression. Open e.g. following campaign in editor and save as vcmp.

https://heroesportal.net/en/maps/download/1326

PR #5696, opened 2025-05-08
fix crash in map loading

kdmcser:
Some maps require placing artifacts during initialization, which involves accessing the map field in gameState. However, this field is still being constructed during init and hasn't been assigned yet, which leads to a crash. I’ve written a temporary fix for now, but ideally, the code should be refactored if time permits. Additionally, these bugs are catastrophic in the map editor—cb is completely unavailable, making it impossible to use gameState. This part might have to be scrapped altogether.

### DIFF
--- a/lib/gameState/CGameState.cpp
+++ b/lib/gameState/CGameState.cpp
@@ -624,6 +624,7 @@ void CGameState::initHeroes()
 
 			boat->setAnchorPos(hero->anchorPos());
 			boat->appearance = handler->getTemplates().front();
+			map->generateUniqueInstanceName(boat.get());
 			map->addNewObject(boat);
 			hero->setBoat(boat.get());
 		}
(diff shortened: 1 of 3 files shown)


IvanSavenko:
Most likely fixed as part of #5690, please recheck.

Yeah, callback needs some refactoring, but so far this was far from critical need. I think my current solution should work for now, but not 100% sure whether I checked everything

kdmcser:
still crash in some case：
image

MichalZr6:
I think this popup window crash could be fixed using new `EditorCallback` class:

```diff

diff --git a/client/windows/CMapOverview.cpp b/client/windows/CMapOverview.cpp
index 8163aa7da..317899ec8 100644
--- a/client/windows/CMapOverview.cpp
+++ b/client/windows/CMapOverview.cpp
@@ -34,6 +34,7 @@
 #include "../../lib/mapping/MapFormat.h"
 #include "../../lib/TerrainHandler.h"
 #include "../../lib/filesystem/Filesystem.h"
+#include "../../lib/callback/EditorCallback.h"
 
 #include "../../lib/StartInfo.h"
 #include "../../lib/mapObjects/CGHeroInstance.h"
@@ -105,7 +106,8 @@ std::vector<std::shared_ptr<CanvasImage>> CMapOverviewWidget::createMinimaps(con
        std::unique_ptr<CMap> map;
        try
        {
-               map = mapService.loadMap(resource, nullptr);
+               auto cb = std::make_unique<EditorCallback>(map.get());
+               map = mapService.loadMap(resource, cb.get());
        }
        catch (const std::exception & e)
        {
```
Crashes related to loading maps and mapeditor should be fixed in current build. Please check.

PR #5699, opened 2025-05-09
fix crash during loading map

kdmcser:
When loading certain maps, such as Hota's dead man's tales, the commander pointer may be nullptr, which will lead to a crash.


IvanSavenko:
Should be already fixed as part of #5690

PR #5713, opened 2025-05-15
Game callback rework (part 1?)

IvanSavenko:
- Completely removed `IGameCallback` class. Reasoning:
  - Despite its name, it is not interface, but rather class that combines two very different classes: `IGameEventCallback`, which defines events that can only be executed on server, and `CGameInfoCallback`, which (mostly) defines methods to access game information.
  - Both `CClient` and `CGameHandler` inherit from `IGameCallback`, but since `CClient` actually can't execute server events, half of `CClient` API are dummy no-op methods from `IGameEventCallback`, which was confusing for some of our new devs.

- `CPrivilegedInfoCallback` class is also completely removed, its method moved to class it inherited from, `CGameInfoCallback`, since none of such methods are actually "privileged" in any way, and can be used by non-privileged users.

As result of these changes:
- `CClient` now inherits solely from `CGameInfoCallback`, dummy implementation of IGameEventCallback has been removed
- `CGameHandler` now inherits from both `CGameInfoCallback` and `IGameEventCallback` directly, without dummy intermediate class
- `CGObjectInstance`-based classes now use more minimalistic `IGameInfoCallback` interface (which is actually an interface) for accessing game information, instead of `IGameCallback`
- `CGObjectInstance` methods that can only be called from server and can actually change gamestate, like `onHeroVisit` now receive reference to `IGameEventCallback` interface (implemented by `CGameHandler`). As result it is no longer possible to incorrectly call `onHeroVisit` and such methods on client, only to be surprised that they have no effect.
- `IGameInfoCallback` interface has been expanded with methods from CGameInfoCallback that were used by map object calls.
- Flaggable objects (Lighthouse) now contain bonus system mode instead of attaching bonuses to owning players, since this required multiple hacks, some of which are now blocked by API

At the moment not 100% sure as to what next changes would be - this PR should be self-contained, and any future changes would depend on what's needed for netpack rollback or to fix usage of callback by editor.


dydzio0614:
it would be extra benefit if rework design allows making stuff mockable in unit tests

IvanSavenko:
It should be *slightly* better.

For example map objects now operate through `IGameInfoCallback`, which is actually an interface, and it already had mocked implementation before this PR.

However at the moment that interface also has an amazing method to access entire `CGameState`, which some map objects require - not something you can easily mock. Good news is that only some map object calls require access to that method, and it should be possible to completely remove it after some (hopefully not too hard) refactoring.

Hmm... Once that part is done it actually might be possible to make unit tests for map objects if needed. Generally map objects can interact with outside world via:
- `IGameInfoCallback` - to access game information. Mock exists, but that method to access `CGameState` needs to removed for proper testing.
- `IGameEventCallback` - to modify gamestate, by generating network packs. Should be mockable after this PR
- and there is also small number of cases where map objects modify gamestate directly. Have yet to look as to how to resolve this, since this is basically connection of poorly designed edge cases. Possibly another small interface implemented by CGameState is needed for this scenario

PR #5717, opened 2025-05-16
Allow removal of expert-level skill using configurable map object

IvanSavenko:
- Replaced undescriptive true/false parameters in netpacks with more clear enum ChangeValueMode (ABSOLUTE/RELATIVE)
- Fixed changing of secondary skill values in both absolute and relative mode. This fixes removal of expert-level secondary skills via configurable map objects
- Supersedes #5613

### DIFF
--- a/Global.h
+++ b/Global.h
@@ -85,6 +85,9 @@ static_assert(sizeof(bool) == 1, "Bool needs to be 1 byte in size.");
 #  ifndef NOMINMAX
 #    define NOMINMAX				 // Exclude min/max macros from <Windows.h>. Use std::[min/max] from <algorithm> instead.
 #  endif
+#  ifndef NOGDI
+#    define NOGDI
+#  endif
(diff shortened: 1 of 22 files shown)


IvanSavenko:
Works for me with my test object:
```json
				"rewards" : [
					{
						"secondary": [
							{
								"type" : "wisdom",
								"amount" : -3
							}
						]
					}
				]
```

GeorgeK1ng:
I see now, different config type. Was using this:

```
					{
						"limiter": {
							"secondary" : { 
								"firstAid" : 1
							},
							"resources" : {
									"gold" : 5000
							}
						},
						"secondary": {
							  "firstAid"  : 0
						},
						"resources" : { 
							"gold" : -5000
						}
					}
```

Will update mod to mentioned schema

IvanSavenko:
Both versions are identical and both work for me...

Issue #4932, opened 2024-11-17
Fragile Artifacts 


edeksumo:
**Describe your proposal**
Fragile Artifacts after turn or battle would disappear from inventory. Maybe with chance to survive.
It would open new possiblity for artifact creation,  powerful one time use artifact. Potions etc. 

**Additional context**

krs0:
So basically Artifacts with Charges? or Charge-Based Artifacts?

I've encountered this proposal for scrolls.

The request to also have a chance not to consume a charge would be something extra?

edeksumo:
Yep, something like that. With ability to specify if the charge is used after battle or after player turn.

paddotk:
I think the inner workings of HoMM3 should not be changed. This might be nice for a mod, but not VCMI itself.

IvanSavenko:
@paddotk 

Correct. No idea whether such "fragile artifacts" would ever be implemented, but it would be added only for use in mods.

VCMI without mods uses and will keep using Heroes 3: SoD mechanics

IvanSavenko:
Implemented in #5720

PR #5720, opened 2025-05-16
Charged artifacts

SoundSSGood:
- Removed hardcoded binding of growing artifacts to commanders. Now the hero can have a growing artifact.

- Charged artifacts have been implemented. Three actions are available to choose from for discharging BATTLE, BUILDING (not yet implemented), SPELLCAST.

- Added new bonus type ARTIFACT_CHARGE. 

[Fragile](https://github.com/vcmi/vcmi/issues/4932) artifacts can now be implemented through "charged" mechanic.
```
{
	"charged" : {
		"usageType" : "BATTLE",
		"startingCharges" : 1,
		"removeOnDepletion" : true
	},
}
```
The Grail can also be implemented as a charged artifact with one charge.
Another example of a charged artifact with replenishing charges.
```
{
    "bonuses" : [
        {
            "type" : "SPELL",
            "subtype" : "bless"
        }
    ],
    "charged" : {
        "usageType" : "SPELLCAST",
        "startingCharges" : 2,
    },
    "growing" : {
        "bonusesPerLevel" :
        [
            {
                "level" : 1,
                "bonus" : 
                {
                    "type" : "ARTIFACT_CHARGE",
                    "val" : 1
                }
            }
        ]
    }
}
```

### DIFF
--- a/docs/modders/Entities_Format/Artifact_Format.md
+++ b/docs/modders/Entities_Format/Artifact_Format.md
@@ -98,7 +98,7 @@ In order to make functional artifact you also need:
     // Optional, by default is false. Remove when fully discharged
     "removeOnDepletion" : true,
     // Optional, by default is 0. Default starting charge amount.
-    "val" : 2,
+    "startingCharges" : 2,
 	}
 }
 ```
--- a/docs/modders/Entities_Format/Artifact_Format.md
+++ b/docs/modders/Entities_Format/Artifact_Format.md
@@ -87,5 +87,18 @@ In order to make functional artifact you also need:
 		"bonusesPerLevel" : {},
 		"thresholdBonuses" : {},
 	}
+
+	// Optional, used for artifacts with charges.
+	"charged" : {
+    // Artifact discharging action
+    // SPELLCAST - Consumes a charge for each spellcast. Applies to every spell added through the "bonuses" section.
+    // BATTLE - Consumes one charge per battle.
+    // BUILDING (not implemented)
+    "usageType": "BATTLE",
+    // Optional, by default is false. Remove when fully discharged
+    "removeOnDepletion" : true,
+    // Optional, by default is 0. Default starting charge amount.
+    "val" : 2,
+	}
 }
 ```
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -3931,6 +3931,9 @@ void CGameHandler::castSpell(const spells::Caster * caster, SpellID spellID, con
 
 	const CSpell * s = spellID.toSpell();
 	s->adventureCast(spellEnv, p);
+
+	if(const auto hero = caster->getHeroCaster())
+		verifyChargedArtifactUsed(hero->id, spellID);
 }
 
 bool CGameHandler::swapStacks(const StackLocation & sl1, const StackLocation & sl2)
@@ -4336,4 +4339,36 @@ void CGameHandler::startBattle(const CArmedInstance *army1, const CArmedInstance
 void CGameHandler::startBattle(const CArmedInstance *army1, const CArmedInstance *army2 )
 {
 	battles->startBattle(army1, army2);
+}
+
+void CGameHandler::verifyChargedArtifactUsed(const ObjectInstanceID & heroObjectID, const SpellID & spellID)
+{
+	if(const auto hero = getHero(heroObjectID))
+	{
+		assert(hero->canCastThisSpell(spellID.toSpell()));
+
+		if(vstd::contains(hero->getSpellsInSpellbook(), spellID))
+			return;
+
+		std::vector<std::pair<ArtifactPosition, ArtifactInstanceID>> chargedArts;
+		for(const auto & [slot, slotInfo] : hero->artifactsWorn)
+		{
+			const auto artInst = slotInfo.getArt();
+			const auto artType = artInst->getType();
+			if(artType->getDischargeCondition() == DischargeArtifactCondition::SPELLCAST)
+			{
+				chargedArts.emplace_back(slot, artInst->getId());
+			}
+			else
+			{
+				if(const auto bonuses = artInst->getBonusesOfType(BonusType::SPELL, spellID); !bonuses->empty())
+					return;
+			}
+		}
+
+		assert(!chargedArts.empty());
+		DischargeArtifact msg(chargedArts.front().second, 1);
+		msg.artLoc.emplace(hero->id, chargedArts.front().first);
+		sendAndApply(msg);
+	}
 }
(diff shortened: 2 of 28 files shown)


Alucard648:
Will it be possible to have more ways to recharge artifacts? In H4 wands have recharge rate at X per day, yet each usage expends Y charges. For instance, Wand of Implosion has 8 charges, each Implosion cast uses them all and the wand regenerates 2 charges per day. In H5 wands also allow casting specific spells and are recharged in castle mage guild upon visit.

edeksumo:
It would be cool to be able to determine whether artifact will disappear from inventory or be replaced with "broken" one, and then create bonuses that will be able to "recharge" artifact.

SoundSSGood:
> Will it be possible to have more ways to recharge artifacts?

Yes, it is possible. I think I will add conditions for the growth of the artifact. Siege victory, hero victory, monster victory,  per day. 

> Wand of Implosion has 8 charges, each Implosion cast uses them all and the wand regenerates 2 charges per day

This PR implements one charge per spellcast. Through the growth mechanics, it will be possible to adjust the number of days to get one charge.

> In H5 wands also allow casting specific spells and are recharged in castle mage guild upon visit.

Recharging by map object is possible. But it is beyond the scope of this PR. I do not plan to do this for now. But, in theory, it is not difficult to do.

> It would be cool to be able to determine whether artifact will disappear from inventory

See "removeOnDepletion" field.

misiokles:
"Removed hardcoded binding of growing artifacts to commanders. Now the hero can have a growing artifact." 
This part will break some WoG artifacts where bonuses grow with hero levels?

SoundSSGood:
> "Removed hardcoded binding of growing artifacts to commanders. Now the hero can have a growing artifact." This part will break some WoG artifacts where bonuses grow with hero levels?

Growing artifacts work for commanders and heroes.

docs/modders/Entities_Format/Artifact_Format.md
```diff
@@ -87,5 +87,18 @@ In order to make functional artifact you also need:
 		"bonusesPerLevel" : {},
 		"thresholdBonuses" : {},
 	}
+
+	// Optional, used for artifacts with charges.
+	"charged" : {
```
IvanSavenko, line 92: Not sure about this term.
When I hear term "charged" I first think about item that needs to be "charged" somehow, and then can be used to perform some powerful effect that uses accumulated charge.

But this seems more like consumable artifact? Not sure which one is better / more clear...

Same question to C++ class by the way.

server/CGameHandler.cpp
```diff
@@ -4337,3 +4340,35 @@ void CGameHandler::startBattle(const CArmedInstance *army1, const CArmedInstance
...
 }
+
+void CGameHandler::verifyChargedArtifactUsed(const ObjectInstanceID & heroObjectID, const SpellID & spellID)
+{
+	if(const auto hero = getHero(heroObjectID))
+	{
+		assert(hero->canCastThisSpell(spellID.toSpell()));
+
+		if(vstd::contains(hero->getSpellsInSpellbook(), spellID))
+			return;
```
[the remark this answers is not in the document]
SoundSSGood: Could you please clarify where exactly you propose to call this method from?

server/CGameHandler.cpp
```diff
@@ -3931,6 +3931,9 @@ void CGameHandler::castSpell(const spells::Caster * caster, SpellID spellID, con
 
 	const CSpell * s = spellID.toSpell();
 	s->adventureCast(spellEnv, p);
+
+	if(const auto hero = caster->getHeroCaster())
+		verifyChargedArtifactUsed(hero->id, spellID);
```
IvanSavenko, line 3936: @SoundSSGood 
I am talking about this place.
Perhaps that check for "hero has spell in spellbook" should be done here instead?
And only call `verifyChargedArtifactUsed` if hero does not have that spell in spellbook and has to expend charge from artifact

SoundSSGood: Theoretically, there could still be artifacts without charges, with a bonus in the form of a spell. Or temporary bonuses. I thought it would be more convenient to move the check to a separate method.

IvanSavenko: Like scrolls? In that case - yeah, probably makes sense.

docs/modders/Entities_Format/Artifact_Format.md
SoundSSGood: ChatGPT says that "consumable" is more suitable for single-use items that disappear after use. "charged" is for items that can be recharged. Since we have both cases at once, I would prefer to keep "charged".

GeorgeK1ng:
So we can now create real consumable artifacts? Like "You have found magic mushroom, do you want to eat it to make your attack +2 forever"?

SoundSSGood:
Well, at the moment, charged artifacts do not provide a permanent effect, only a temporary one, while there are charges and while the artifact is equipped. Your case is somewhat different. Instant use with the addition of a permanent bonus and the removal of the artifact. Equipping is not necessary at all and a separate dialogue is needed for consumption. Charges are not also needed in this case. This can be implemented, but it looks like a separate type of artifacts, "consumable" ?. If this really needed, fell free to create a proposal.

misiokles:
I don't think "consumable" artifacts are really needed. We have such artifacts in game - Star Axis, Marletto Tower, Arena...

IvanSavenko:
Code-wise looks OK. Still not sure about charged vs consumable, but obviously minor. Looks OK for merge for me.

Potential improvements from my point of view for future PR's, including some of already mentioned ones:
- Grail as consumable - not a bad idea, but rather low-priority compared to others IMO
- Option to explicitly "use" artifact, to get some effect, e.g. bonuses. For example, potion that provides bonuses for 1 battle that hero can drink before risky battle. Another option is to reuse configurable objects logic, but probably overkill for this mechanics.
- A way to configure number of charges per use. Yes, you've mentioned possibility to specify number of days to get 1 charge, however this would also need some way to present progress to player (how many days are left till recharge?). And existing charges can be used to support such logic - if object needs 8 charges to use and you have 5 it is clear that you'll be able to cast it in 3 days.

dydzio0614:
what about growing artifacts that gain experience with hero

SoundSSGood:
> what about growing artifacts that gain experience with hero

Did you mean GROWS_WITH_LEVEL ?  Should work as before.

dydzio0614:
no, I mean artifacts with own exp counter in description that can have list of new bonuses gained on specific levels etc. and artifact exprience stays when transferring between heroes

dydzio0614:
it was not question if it works, it was idea for new artifact feature :P

PR #5723, opened 2025-05-19
Cheats improvements

Laserlicht:
- adds all aliases (AB + RoE) from https://heroes.thelazy.net/index.php/Cheats
- implement `nwcphisherprice` and a grayscale cheat
- use tbb for grayscale conversation


client/renderSDL/SDL_Extensions.cpp
```diff
@@ -543,28 +543,42 @@ void CSDL_Ext::putPixelWithoutRefreshIfInSurf(SDL_Surface *ekran, const int & x,
 }
 
 template<int bpp>
-void CSDL_Ext::convertToGrayscaleBpp(SDL_Surface * surf, const Rect & rect )
+void loopOverPixel(SDL_Surface * surf, const Rect & rect, std::function<void(int &r, int &g, int &b)> func)
```
IvanSavenko, line 546:
```suggestion
void loopOverPixel(SDL_Surface * surf, const Rect & rect, Functor func)
```
Since this is template anyway, then it would be better to provide function not as std::function, but as template parameter, to avoid forced function call for every pixel

lib/serializer/ESerializationVersion.h
```diff
@@ -42,8 +42,9 @@ enum class ESerializationVersion : int32_t
 	REWARDABLE_EXTENSIONS, // new functionality for rewardable objects
 	FLAGGABLE_BONUS_SYSTEM_NODE, // flaggable objects now contain bonus system node
 	RANDOMIZATION_REWORK, // random rolls logic has been moved to server
+  COLOR_FILTER, // color filter cheat
```
IvanSavenko, line 45: 1. we only maintain serialization compatibility for saves, not for netpacks. Multiplayer between 1.X and 1.Y is not possible. Small exception that I usually try to avoid is hotfix MP compatibility,  e.g 1.X.Y and 1.X.Z should be generally compatible. But this is not the case here.
2. Spaces instead of tabs?

client/renderSDL/SDL_Extensions.h
```diff
@@ -71,6 +71,10 @@ SDL_Color toSDL(const ColorRGBA & color);
 	void convertToGrayscaleBpp(SDL_Surface * surf, const Rect & rect);
 	void convertToGrayscale(SDL_Surface * surf, const Rect & rect);
 
+	template<int bpp>
+	void convertToH2SchemeBpp(SDL_Surface * surf, const Rect & rect);
+	void convertToH2Scheme(SDL_Surface * surf, const Rect & rect);
```
IvanSavenko, line 76: Not critical, but screen surface is guaranteed to be 4 bpp (32-bit), so such templates are no longer necessary.

PR #5725, opened 2025-05-19
Custom descriptions and icons for creature abilities

IvanSavenko:
It is now possible to add custom icon and custom description  to list of creature abilities.

Expected format for descriptions is `{Ability Name}\nLong description text`. Note that space is still rather limited, with 4 lines in total (presumably - 1 line for name & 3 lines for description).

It is also possible to force-show normally hidden bonuses, such as luck & morale - if custom icon is used, such bonus will be visible in creature stats. 

For now only creature abilities support custom descriptions. If needed - functionality can be extended to other bonus sources, such as artifacts and hero secondary skills

Also:
- Game will no longer attempt to group similar abilities - this is generally irrelevant, since bonuses that can actually stack, like morale or stats are hidden anyway
- Game will now show abilities that are propagated "away" from creature, for example abilities that only affect enemy in combat (assuming that ability has icon, so Ghost Dragon's morale debuff is still hidden by default)

Example: 
image


DjWarmonger:
> Game will no longer attempt to group similar abilities - this is generally irrelevant, since bonuses that can actually stack, like morale or stats are hidden anyway

This was meant for stuff like accumulated bonuses from [stack experience](https://github.com/vcmi-mods/wake-of-gods/blob/f6329e70bd5078f501de4772dc5ef07fa371efd1/Mods/stackExperience/Content/config/confluxCreaturesStackExp.json#L394) and other bonuses granted by mods. For instance H4 artifacts mod grants premy to damage resistance. But event magic resistance from base OH3 stacks with resistance of dwarves. Player would rather wish to see total value than two separate bonuses which occupy extra space.

IvanSavenko:
Any examples other than magic resistance? 

For example pixies / sprites from your link don't have any bonuses that are shown in unit window and can stack with something else (other than magic resistance).

We also now display bonus source, meaning that we might end up with stuff like magic resistance with "hero skill" source, but actually that value is accumulated from hero skill, creature ability, and stack experience. Which is not clear either.

PR #5726, opened 2025-05-20
Configurable multi-hex attacks

IvanSavenko:
Breath-like attacks are now fully configurable, removing need of adding new hardcoded bonus whenever modders come up with new breath layout they find interesting. It is also possible to specify whether to use alternative attack animation (if provided in unit animations)

- Added bonus type `MULTIHEX_UNIT_ATTACK` - configurable version of Dragon Breath.
- Added bonus type `MULTIHEX_ENEMY_ATTACK` - configurable version of Cerberi multi-headed attack that only hits enemies
- Added bonus type `MULTIHEX_ANIMATION` - optional bonus that does not affects gameplay, but allows to define in which cases game should use alternative attack animation.
- All existing multi-hex attack bonuses other than `ATTACKS_ALL_ADJACENT` are presumable deprecated, but will be supported for now.
- It is now possible to precisely configure which hexes are targeted by `MULTIHEX_XXX` bonuses. See docs for details.
- Unified logic of all multi-hex attacks, all existing bonuses are now implemented as specific case of `MULTIHEX_XXX` bonus
- Added tests to cover Cerberi attack logic, and fixed incorrect edge case of Dragon Breath

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -943,6 +943,7 @@ void CPlayerInterface::battleAttack(const BattleID & battleID, const BattleAttac
 	info.unlucky = ba->unlucky();
 	info.deathBlow = ba->deathBlow();
 	info.lifeDrain = ba->lifeDrain();
+	info.playCustomAnimation = ba->playCustomAnimation();
 	info.tile = ba->tile;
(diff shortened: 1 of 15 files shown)


IvanSavenko:
Not a bad idea, will add to my TODO list, but at the moment not possible.

PR #5727, opened 2025-05-20
Remove hardcoded mass-casting of ENCHANTER bonus

IvanSavenko:
ENCHANTER bonus will now cast spells with range specified by their spell school mastery level, so spells are (generally) only cast on every target if they are at expert level.

Area-based spells are cast with unit at center of the spell. Single-target spells are cast on unit itself.

### DIFF
--- a/server/battles/BattleFlowProcessor.cpp
+++ b/server/battles/BattleFlowProcessor.cpp
@@ -768,7 +768,7 @@ void BattleFlowProcessor::stackTurnTrigger(const CBattleInfoCallback & battle, c
 		});
 
 		BattleSide side = battle.playerToSide(st->unitOwner());
-		if(st->canCast() && battle.battleGetEnchanterCounter(side) == 0)
+		if(st->canCast())
 		{
 			bool cast = false;
 			while(!bl.empty() && !cast)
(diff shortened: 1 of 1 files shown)


edeksumo:
🔥🔥🔥 and only one suggestion, if it won't be too much effort, parameter that will allow to turn of animation of casting a spell by creature. Something like:

"useSpellCastAnimation = true"

IvanSavenko:
> parameter that will allow to turn of animation of casting a spell by creature

Sorry, probably not this time. Doable if needed, but not trivial. 

I guess one option for now would be to give 1-frame long spellcasting animation.

madaosoul:
That change messed up one of my mods, hah. I don't really mind and would be ok with just removing it from the launcher, but I received a complaint that got my attention and realised, that maybe people really want to keep playing with my silly ideas and would be sad if one of them disappeared.

So how about introducing this little, yet elegant solution - value above 3 would apply mass range effect, similar to the ENCHANTED bonus (so "val" : 4 = basic with mass range; "val" : 5 = advanced with mass range; "val" : 6 = expert with mass range). If it could be added to the to-do list and realised at some later time, it would be splendid too. No pressure.

PR #5729, opened 2025-05-20
Gameplay improvements

dydzio0614:
Some minor improvements for VCMI 1.7

1. Allow "finish battle with autocombat" during tactics phase
2. Remove default 16 total heroes limit (probably some old H3 version thing)
3. Change scroll direction for horizontal slider (matches HD mod) - IMO not just comptibility change but also minor ergonomic improvement - I find scrolling down a bit easier than scrolling up
4. Allow fleeing while having shackles of war when other side got no hero - change made by HotA that allows fleeing in battles where shackles of war are useless, so you do not have to keep putting it in backpack when opponent does not have hero anyway etc.


IvanSavenko:
> Allow fleeing while having shackles of war when other side got no hero

Strictly speaking, there might be cases where mods might want old behavior, for example artifacts that gives stat boosts in addition to non-escaping. But no such mods so far AFAIK.

PR #5733, opened 2025-05-21
Mapeditor fixes for various crashes

MichalZr6:
This PR mainly implements `MapInfoCallback` and `EditorCallback` for the map editor. These changes resolve recent crashes caused by updates to the callback system used by the game client, which were not reflected in the editor.

### Fixes Included:
- Implemented `MapInfoCallback` and `EditorCallback` to align with updated callback expectations and prevent runtime crashes in the editor.
- Fixed a crash when closing the `HeroArtifactsWidget` dialog with no artifacts present.
- Fixed a crash that occurred when copy-pasting objects on the map in the editor.
- Fixed crash in `getAverageExperience` when creature stack had `count` = 0.


mapeditor/inspector/inspector.cpp
```diff
@@ -74,10 +75,10 @@ void Initializer::initialize(CGSignBottle * o)
 void Initializer::initialize(CGCreature * o)
 {
 	if(!o) return;
-	
+
 	o->character = CGCreature::Character::HOSTILE;
 	if(!o->hasStackAtSlot(SlotID(0)))
-		o->putStack(SlotID(0), std::make_unique<CStackInstance>(o->cb, CreatureID(o->subID), 0, false));
+		o->putStack(SlotID(0), std::make_unique<CStackInstance>(nullptr, CreatureID(o->subID), 1, false));
```
IvanSavenko: Passing null as callback - deliberate?

mapeditor/helper.cpp
```diff
@@ -23,7 +23,7 @@
 #include "../lib/mapping/MapFormatJson.h"
 #include "../lib/modding/ModIncompatibility.h"
 
-std::unique_ptr<CMap> Helper::openMapInternal(const QString & filenameSelect)
+std::unique_ptr<CMap> Helper::openMapInternal(const QString & filenameSelect, IGameInfoCallback * ecb)
```
IvanSavenko: `ecb`? Editor callback? But you're passing interface, not editor implementation of it.
call it `cb` or `gameInfo` perhaps?

lib/mapObjects/CGHeroInstance.cpp
```diff
@@ -1706,7 +1707,12 @@ void CGHeroInstance::serializeCommonOptions(JsonSerializeFormat & handler)
 	handler.serializeIdArray("spellBook", spells);
 
 	if(handler.saving)
-		CArtifactSet::serializeJsonArtifacts(handler, "artifacts", &cb->gameState().getMap());
+	{
+		if(auto * ecb = dynamic_cast<EditorCallback *>(cb))
+			CArtifactSet::serializeJsonArtifacts(handler, "artifacts", const_cast<CMap *>(ecb->getMapConstPtr()));
+		else
+			CArtifactSet::serializeJsonArtifacts(handler, "artifacts", &cb->gameState().getMap());
+	}
```
IvanSavenko: Ugly, but don't see any clear alternative either...
Can you add FIXME comment here that explains why such approach is needed, to help whoever will find this code in the future?

lib/entities/artifact/ArtSlotInfo.h
```diff
@@ -21,6 +21,7 @@ struct DLL_LINKAGE ArtSlotInfo : public GameCallbackHolder
 	ArtifactInstanceID artifactID;
 	bool locked = false; //if locked, then artifact points to the combined artifact
 
+	ArtSlotInfo() = delete;
```
IvanSavenko: Not against it, but why is it needed? Default constructor should be only created if there are no other constructrors

lib/callback/MapInfoCallback.h
```diff
@@ -0,0 +1,43 @@
...
+
+VCMI_LIB_NAMESPACE_BEGIN
+
+class CMap;
+
+class DLL_LINKAGE MapInfoCallback : public IGameInfoCallback
+{
+public:
+	virtual ~MapInfoCallback();
+	virtual const CMap * getMapConstPtr() const = 0;
```
IvanSavenko: Perhaps protected? We generally don't want to allow access to map via callbacks

IvanSavenko (approved):
Generally OK, don't see anything critical, especially since goal of PR is to fix crashes

lib/entities/artifact/ArtSlotInfo.h
MichalZr6: It took me some time to figure out why HeroArtifactsWidget is crashing and this is some leftover from my tries. Will remove.

lib/callback/MapInfoCallback.h
MichalZr6: Yes, right, should be protected in `MapInfoCallback`.

lib/mapObjects/CGHeroInstance.cpp
MichalZr6: Yeah, I know, not pretty at all. But apart from that: I was very, very surprised with existing code `&cb->gameState().getMap()` that compiles without needing of const_cast :open_mouth: 
Will add FIXME

Laserlicht:
FYI: This PR causes a few assert trigger on campaign editor.
Still work, but annoying when doing changes.

E.g. when importing H3M into region.

PR #5735, opened 2025-05-21
Per-instance bonuses for artifacts

IvanSavenko:
It is now possible to give artifacts per-instance bonuses, if needed.

Unlike shared bonuses, per-instance bonuses stack if multiple instances of same artifacts are equipped on hero.

This to implement resource-producing artifacts in line with H3 - equipping multiple such artifacts on a single hero will give bonus from each instance of such artifact.

Also, both existing bonuses and new instanceBonuses fields now use json object instead of json lists. This allows easier modification of individual bonuses of artifacts and potentially - custom icons / descriptions for artifact bonuses.

Old mods should work as before, but will produce warning during validation

### DIFF
--- a/config/artifacts.json
+++ b/config/artifacts.json
@@ -40,872 +40,872 @@
 	},
 	"centaurAxe":
 	{
-		"bonuses" : [
-			{
+		"bonuses" : {
+			"attack" : {
 				"subtype" : "primarySkill.attack",
 				"type" : "PRIMARY_SKILL",
 				"val" : 2,
 				"valueType" : "BASE_NUMBER"
 			}
-		],
+		},
 		"index" : 7,
 		"type" : ["HERO"]
 	},
 	"blackshardOfTheDeadKnight":
 	{
-		"bonuses" : [
-			{
+		"bonuses" : {
+			"attack" : {
 				"subtype" : "primarySkill.attack",
 				"type" : "PRIMARY_SKILL",
 				"val" : 3,
 				"valueType" : "BASE_NUMBER"
 			}
-		],
+		},
 		"index" : 8,
 		"type" : ["HERO"]
 	},
 	"greaterGnollsFlail":
 	{
-		"bonuses" : [
-			{
+		"bonuses" : {
+			"attack" : {
 				"subtype" : "primarySkill.attack",
 				"type" : "PRIMARY_SKILL",
 				"val" : 4,
 				"valueType" : "BASE_NUMBER"
 			}
-		],
+		},
 		"index" : 9,
 		"type" : ["HERO"]
 	},
 	"ogresClubOfHavoc":
 	{
-		"bonuses" : [
-			{
+		"bonuses" : {
+			"attack" : {
 				"subtype" : "primarySkill.attack",
 				"type" : "PRIMARY_SKILL",
 				"val" : 5,
 				"valueType" : "BASE_NUMBER"
 			}
-		],
+		},
 		"index" : 10,
 		"type" : ["HERO"]
 	},
 	"swordOfHellfire":
 	{
-		"bonuses" : [
-			{
+		"bonuses" : {
+			"attack" : {
 				"subtype" : "primarySkill.attack",
 				"type" : "PRIMARY_SKILL",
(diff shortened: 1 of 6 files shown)


config/artifacts.json
```diff
@@ -2148,14 +2148,14 @@
 	},
 	"titansThunder":
 	{
-		"bonuses" : [
-			{
+		"bonuses" : {
+			"titanBolt" : {
```
SoundSSGood, line 2152: Are these bonus names predefined somewhere?

IvanSavenko: No. Just arbitrary name, that can be used as key for overriding them in mods for example. It can be changed safely, the only thing that might break would be mods that change this particular bonus

Creature abilities and multiple other places also use this form, artifacts might be the only place that does not.

SoundSSGood:
It turns out that in order for the bonuses to stack, they need to be added to the instance. Charged artifacts use the instance to store bonuses. It turns out that this was a regression 🤔.

IvanSavenko:
No?
If it was a regression it was an extremely old one.

In current develop, all "generic" artifacts (not scrolls/growing/charged) store bonuses in CArtifact. As result, if hero has multiple copies of the same art, bonus system graph will be diamond-shaped - two CArtifactInstance's are attached to same CArtifact, so hero receives a single copy of bonus.

This logic was present for years, and works fine for most artifacts. What this PR changes is that it is now possible to effectively override this logic, and store selected bonuses in CArtifactInstance directly.

SoundSSGood:
I mean adding charged artifacts. The idea is that if an artifact contains other bonuses in addition to a spell, then instead of adding a bonus node, these bonuses are transferred to the instance. Thus, when completely discharged, the spell can be turned off, but leave the other bonuses.  And this is regression.

yche2990:
What about the old mod like fused artifact? Will it be compelled to read all other component artifacts bonuses? Will it still work in the new system?

SoundSSGood:
> What about the old mod like fused artifact? Will it be compelled to read all other component artifacts bonuses? Will it still work in the new system?

Fused ones do not include bonuses from components. Only those that are defined in the mod.

yche2990:
I understand that in the old system, the fused artefact mod must include all bonuses of the components as wells as the combined artifact bonuses in order to give effect of the mod. The bonus label cannot operate cumulatively. The new system will seperate each instance? Will the fused mod have to be remade?

SoundSSGood:
> Will the fused mod have to be remade?

Only possible if you take into account stacking/non-stacking bonuses.

SoundSSGood:
I'm talking about the issue that charged artifacts transfer bonuses to the instance. Thus, bonuses from non-stackable are transformed into stackable ones. This is the scenario for all charged artifacts. 
I don't quite see how own source ID can solve this. And I don't have any good ideas yet on how to fix it.

IvanSavenko:
> I don't quite see how own source ID can solve this.

Instead of selecting all local bonuses, you can select bonuses with ARTIFACT_CHARGED source ID to differentiate them. Or not?

SoundSSGood:
Still don't quite get it. I'll look at the code to see how it works. But will it allow to enable/disable spells without detach the general bonus node?

SoundSSGood:
> . This would allow you to exclude bonuses from spell scrolls, growing artifacts, or from instance bonuses from this PR.

But how does this make them non-stacking? (As if they were in a CArtifact)

IvanSavenko:
If you want to make bonuses from charged artifacts non-stacking then I guess the only option is to add new node, similar to CArtifact, that will contain bonuses from charged artifacts, and if charged artifact is depleted - detach it from that node.

To be honest not sure if charged artifacts should have stacking or non-stacking bonuses. It is rather rare case to begin with - only rings and misc artifacts can be equipped multiple times.

SoundSSGood:
Maybe stackedBonuses more clear instead of instanceBonuses? @IvanSavenko

IvanSavenko:
In H3 "stack" often refers to a unit in hero army. 
And there is also "stack experience".

Wanted to use term that is distinct from what we have

PR #5740, opened 2025-05-23
Improved bonus descriptions

IvanSavenko:
Attempt to improve description of creature abilities, mostly targeting H3 creatures.

- Bonuses no longer have separate name and description in localizations - not much point, since they are always used together anyway. All translations have been updated to new form.
- Added custom descriptions to Angels, Devils, and Bone Dragon battle-wide abilities (+/- to morale or luck)
- Added custom description to Marksmen & Grand Elves double-shot ability
- Added generic support for custom descriptions based on bonus subtype or value. Added custom descriptions to "king" abilities to fix awkward references to "Slayer level 0" (aka no-mastery Slayer)
- Added support for custom description of artifact abilities. Orb of Vulnerability now actually says what it does instead of showing up as strange "0% resistance" in creature window.
- Reviewed most of English bonus names & descriptions, and slightly expanded those where it makes sense, to make use of recently expanded space in UI.

TODO: check other translations & update changed keys, update json schemas


DjWarmonger:
> Bonuses no longer have separate name and description in localizations - not much point, since they are always used together anyway. All translations have been updated to new form.

Consider adding bonus info component to dialogs, just like icons for primary skills or luck / morale. Pandora Boxes can already give arbitrary bonus to hero, or his creatures. I had this idea to add bonus icon and name to dialog, but with full description it may look poor.

IvanSavenko:
> Pandora Boxes can already give arbitrary bonus to hero, or his creatures. I had this idea to add bonus icon and name to dialog, but with full name + description it may look poor.

Not a bad idea, but when messing with creature window I've discovered bunch of potential problems due to bonus mechanics, that would also apply to your suggestion:
- Only bonuses with BASE_VALUE or ADDITIVE_VALUE can be displayed in such way. For example specialty that gives +50% to base value of skill should be displayed not as +50%, but as +50%*skill value. Independent min/max bonuses also need custom handling
- Some bonuses have name with embedded value, but some are not. Example: "Mana Drain", but "Drain life (100%)".
- Icons only come in vcmi-extras, so players without it won't see any icons
- The only hidden bonus in H3 pandora's / map objects is movement points, so your suggestion would only apply to it, or to custom vcmi maps, which at the moment consist solely from Antagarich Burning.

If such feature has a demand, then it might be better to consider some way to customize visible icons, including arbitrary icons. This is already might be an issue in some cases. 

For example that "Market of Time" mod works as intended purely by chance - if hero has multiple skills, vcmi shows selection dialog with skill icons (and not with identical gold icon for cost) purely because skill icon comes before gold icon, and vcmi picks first icon for selection dialog.

Similar mod that (for example) allows you to forget skill for a resource of your choosing won't work because all icons would be skill (since resource is 2nd icon, and never picked to represent choice)

PR #5743, opened 2025-05-25
Charge cost for charge based spells 

SoundSSGood:


lib/json/JsonBonus.cpp
```diff
@@ -886,4 +886,19 @@ CSelector JsonUtils::parseSelector(const JsonNode & ability)
 	return ret;
 }
 
+std::shared_ptr<Bonus> JsonUtils::parseChargeBasedSpell(const JsonNode & ability)
+{
+	auto spellBonus = std::make_shared<Bonus>();
+	spellBonus->type = BonusType::SPELL;
+	LIBRARY->identifiers()->requestIdentifier( "spell", ability["spell"], [&spellBonus](int32_t identifier)
```
SoundSSGood, line 893: Every time I try to parse a spell I get "-1", could you take a look at what's wrong @IvanSavenko ?
I'm trying to parse this structure:
```
		"instanceBonuses" : {
			"chargeBasedSpell" : {
					"spell" : "spell.bless",
					"chargeCost" : 2,
					"level" : 3
			}
		},
```

SoundSSGood: I'm using a custom object "chargeBasedSpell" here, see config/schemas/artifact.json. In ability["spell"] there is a line exactly the same as in a regular bonus. But parsing works there with common binus spell.

SoundSSGood: Ok. But after loading is completed, I still don't get the spell. And I don't see any difference between parsing in JsonUtils::parseBonus -> loadBonusSubtype and here. But it works there, and not here.

SoundSSGood: The question is closed

SoundSSGood: > `requestIdentifier` is async method

Thanks for the tip

lib/json/JsonBonus.cpp
```diff
@@ -886,4 +886,19 @@ CSelector JsonUtils::parseSelector(const JsonNode & ability)
 	return ret;
 }
 
+std::shared_ptr<Bonus> JsonUtils::parseChargeBasedSpell(const JsonNode & ability, const std::function<void(const SpellID &)> & callback)
```
IvanSavenko: Might be better to keep in art handler - this is not some generic functionality that can be reused, but rather part of charged artifact loading

lib/gameState/GameStatePackVisitor.cpp
```diff
@@ -929,6 +929,9 @@ void GameStatePackVisitor::visitDischargeArtifact(DischargeArtifact & pack)
 		ePack.posPack.push_back(pack.artLoc.value().slot);
 		ePack.visit(*this);
 	}
+	// Workaround to inform hero bonus node about changes. Obviously this has to be done somehow differently.
+	if(pack.artLoc.has_value())
+		gs.getHero(pack.artLoc.value().artHolder)->nodeHasChanged();
```
IvanSavenko, line 951: Calling `nodeHasChanged` on artifact itself should be enough, for example as part of `discharge` call. 

Invalidating a node will also invalidate everything located "downwards" from the node, in this case - hero & his army.

server/CGameHandler.cpp
```diff
@@ -4338,24 +4338,32 @@ void CGameHandler::useChargedArtifactUsed(const ObjectInstanceID & heroObjectID,
...
+			{
 				return;
+			}
 		}
 	}
 
-	assert(!chargedArts.empty());
-	DischargeArtifact msg(chargedArts.front().second, 1);
-	msg.artLoc.emplace(hero->id, chargedArts.front().first);
+	COMPLAIN_RET_IF(!chargedArt.has_value(), "Hero does not have charged artifacts with the required spell");
```
IvanSavenko: Might be better to throw runtime error.

If this happens, this is not incorrect input, but rather incorrect logic somewhere in server code

server/CGameHandler.cpp
```diff
@@ -4338,24 +4338,32 @@ void CGameHandler::useChargedArtifactUsed(const ObjectInstanceID & heroObjectID,
 	if(vstd::contains(hero->getSpellsInSpellbook(), spellID))
 		return;
```
IvanSavenko: Found at least one case where this check is insufficient - use `vcmiistari` to get all spells and cast something.

config/schemas/artifact.json
```diff
@@ -17,6 +17,26 @@
 					"bonus" : { "$ref" : "bonusInstance.json" }
 				}
 			}
+		},
+		"chargeBasedSpell" : {
```
IvanSavenko: Docs reference `instanceBonuses` key code also checks `instanceBonuses` key, but schema checks `chargeBasedSpell`?

lib/entities/artifact/CArtHandler.cpp
```diff
@@ -204,7 +204,23 @@ std::shared_ptr<CArtifact> CArtHandler::loadFromJson(const std::string & scope,
...
+		{
+			// Parse "chargeBasedSpell"
+			bonus = JsonUtils::parseChargeBasedSpell(b.second, [b, art](const SpellID & spellId)
+			{
+				if(!b.second["chargeCost"].isNull())
+					art->setChargeCostForSpell(spellId, b.second["chargeCost"].Float());
+				else
+					art->setChargeCostForSpell(spellId, 1);
+			});
+		}
```
IvanSavenko: Might be better to use separate field.
Right now you have one field that is used for two rather different use-cases, even if they are somewhat similar internally. Most of this loop is "if" block for example.

lib/gameState/GameStatePackVisitor.cpp
SoundSSGood: In case of artifact node, `children` vector is empty. Even, if add hero node in it, It will work for a while but it will break after saving-loading.

server/CGameHandler.cpp
```diff
@@ -4341,33 +4341,43 @@ void CGameHandler::startBattle(const CArmedInstance *army1, const CArmedInstance
...
+	// Check if hero used charge based spell
+	// To do this, we create a local copy of the hero with the necessary magical bonuses, except for the bonuses of charged artifacts
+	CGHeroInstance caster(&gameInfo());
+	caster.setHypothetic(true);
+	for(const auto & b : *hero->getAllBonuses(Selector::type()(BonusType::SPELLS_OF_LEVEL), nullptr))
+		caster.addNewBonus(b);
+	for(const auto & b : *hero->getAllBonuses(Selector::type()(BonusType::SPELLS_OF_SCHOOL), nullptr))
+		caster.addNewBonus(b);
+	for(const auto & spell : hero->getSpellsInSpellbook())
+		caster.addSpellToSpellbook(spell);
```
IvanSavenko: Potentially easier solution:
make `bool CGHeroInstance::canCastThisSpell` method return some enumeration that represents how hero is able to cast this spell. Or even - make it return `BonusSourceID` of bonus that provided the spell, or own ObjectInstanceID (in case of spell in spellbook).

Current `canCastThisSpell` can be kept for convenience, and have it return true if this new/modified method returned non-empty result.

Oh, this might also make `SPELLCAST` discharge condition more flexible - since it would work with any bonus that provides spells, including SPELLS_OF_LEVEL/SPELLS_OF_SCHOOL

IvanSavenko:
Potentially more simple solution for charges: bonus limiter.

Instead of having `chargeBasedSpellList` what if we were to add "has charges" limiter?
Will probably only work with instanceBonuses - not yet sure how to make it work with non-stacking bonuses.
This limiter would be somewhat similar to existing HAS_ANOTHER_BONUS limiter and will:
- return "NOT_SURE" if undecided bonuses contain ARTIFACT_CHARGES bonus (or whats its name)
- return "ACCEPT" if already accepted bonuses contain ARTIFACT_CHARGES bonus with source ID of our artifact
- and return DISCARD othervice (e.g. if all charges were used).

This means that it will work with *any* bonus (and not just with SPELL bonus), and modder would have to add this new limiter to bonuses that are supposed to get disabled when artifact lost its charges.

With this approach it should be possible to completely drop any logic for removing / detaching bonuses on discharging - all of that will be handled by bonus system.

This is more advanced stuff, but in theory with bonus updaters it is also possible to:
- make non-instance bonuses work (by providing information on which artifact instance bonus came from, for example by changing source from ArtifactID to ArtifactInstanceID)
- make scaling charged artifacts - in which strength of specific or all bonuses scales based on charges (however this might need replacing ARTIFACT_CHARGES bonus with property of ChargedArtifact class - access to bonus system from updater is not possible)

SoundSSGood:
Good solution. Like it. But the implementation may take some time, since I still need to look into the limiters code.

SoundSSGood:
> With this approach it should be possible to completely drop any logic for removing / detaching bonuses on discharging - all of that will be handled by bonus system.

How exactly do you see this functionality? At the moment we don't have a unified `bonus.activate()` method that could handle charges.
Also, detaching bonuses on discharging , there is no unified method to disable the bonus. Only remove the bonus from the instance.

IvanSavenko:
> At the moment we don't have a unified bonus.activate() method that could handle charges.

We have. This is what limiter does - it allows activating bonus only if it passes a certain condition.

If limiter discard the bonus, it will not be added to bonus list, effectively deactivating the bonus

lib/bonuses/Limiters.cpp
```diff
@@ -591,4 +591,25 @@ ILimiter::EDecision NoneOfLimiter::limit(const BonusLimitationContext & context)
 	return wasntSure ? ILimiter::EDecision::NOT_SURE : ILimiter::EDecision::ACCEPT;
 }
 
+HasChargesLimiter::HasChargesLimiter(const uint16_t cost)
+	: chargeCost(cost)
+{
+}
+
+ILimiter::EDecision HasChargesLimiter::limit(const BonusLimitationContext & context, const BonusSourceID & sid) const
```
SoundSSGood: @IvanSavenko  Since IGameInfoCallback is not available in the limiter, is there a way to get `artifactID` from `ArtifactInstanceID` in the limiter?

IvanSavenko: Can't think of any clear way at the moment...

`context.node` contains pointer to entity from which this query is made. In most cases that would be hero, but could also be town/army/unit, possibly other sources.

So one option is to cast to GameCallbackHolder. Will work for queries from town/hero/army, but won't work for units in battle (since CStack does not inherits from GameCallbackHolder).

Another option would be to provide it during limiter construction, but that would likely require a lot of changes in code.

Perhaps:
- make `CStack` inherit from `GameCallbackHolder` - should be trivial change, since `CStack`'s should only be created by `BattleInfo` (which is `GameCallbackHolder`)
- use dynamic_cast to convert `context.node` into `GameCallbackHolder` or add method to `CBonusSystemNode` that returns one.

SoundSSGood:
Mostly ready. Can be placed in both "bonuses" and "instanceBonuses". 

And about the two bonus sections. Why can't it be an additional property in the bonus structure? Then the parser can simply distribute the bonuses across the required vectors. This way the internal structure of the code will not affect json files.

PS. Btw what is this field in Bonus.h `std::string stacking`  Maybe remove it?

by003:
Hello.
Can I ask a question?

How can I use it to make the artifact charged to 5 everyday?

Will it available now?

```
"growing" : {
        "bonusesPerDay" :
        [
            {
                "day" : 1,
                "bonus" : 
                {
                    "type" : "ARTIFACT_CHARGE",
                    "val" : 5
                },
               "limiters" : [
	        {
		      "type" : "HAS_CHARGES_LIMITER",
		       "cost" : 5
	         }
                           ]
            }
        ]
    }
```

SoundSSGood:
@by003 Charging from different sources is still not implemented. Currently, you can only get charges per battle. Or give the artifact a certain amount of charges at the start.

by003:
Thank you for replying!

If I cannot set an artifact to use five charges per day, how can I set it to use three charges per battle?

SoundSSGood:
I just checked if
"charged" : {"usageType" : "BATTLE"}
The code specifies one charge per battle. I think this needs to be fixed.

PR #5745, opened 2025-05-27
Icons for heroes on minimap

### LINKED ISSUE #5741: Feature Request: Configurable hero size/icon on minimap for visibility
**Describe your proposal**

Currently, hero icons on the minimap are rendered using a fixed-size 1-pixel logic, which makes them very hard to see — especially on large maps.

Please add support for one or more of the following:

- Configurable scale of hero icon rendering on minimap
- Ability to replace the minimap hero icon with a custom multi-pixel image
- Custom overlays per player/faction for minimap heroes

The goal is to make heroes much easier to identify in PvP or large-map play, where they currently blend into the background.

Thank you!

**Additional context**

Consider using the shield icon from the "View World" screen, even if this overlaps multiple tiles.

IvanSavenko:
VCMI will now show icons for all heroes on visible part of minimap to help with readability on large maps - since on maps larger than XL or on small screens sometimes it is impossible to notice that 1 little pixel that represents a hero.

Old behavior can be enable via toggle in game settings menu

- Fixes #5741

Example:

Image


IvanSavenko:
Clicking on minimap will already move camera to clicked position, which would be very close to position of hero - few tiles at most on XL map.

Is there any need to actually *center* on clicked hero?

misiokles:
I was about to write the same. Now clicking at dot in mimi-map almost center the view of objects. No need to work more with it.

kambala-decapitator:
> Hovering hero on hero list will change color or icon outline on minimap (and maybe selected hero have also different color)

great idea! In battle we already have a similar functionality. And hovering hero on the minimap would highlight the hero in the list.

IvanSavenko:
> Allow right click popup on icon and show hero info

Not a bad idea, but don't want to implement minimap interactivity at the moment, so low priority for me

> Hovering hero on hero list will change color or icon outline on minimap (and maybe selected hero have also different color)

Sounds good. Will take a look.

IvanSavenko:
Not sure if I like it...

зображення

Tried few other versions to show selected hero, but all look... less than perfect

PR #5751, opened 2025-05-28
Campaign code refactoring & Engine support for Hota campaings

IvanSavenko:
Game should now support loading of HotA h3c's. The only missing feature of h3c itself is different prolog/epilogs based on selected starting bonuses (used in one of factory campaigns).

Still requires some support from hota mod, like additional file mapping, so this PR alone won't make hota h3c's playable.

Changes:

- Replaced campaign bonuses from using 3 integers to store anything with type-safe version that uses std::variant that ensures that all bonuses are in correct state. Also removed "interesting" solutions like storing primary skills using bit shifts.
- Moved mapping of h3 video and music files to map format definition. All h3 files are mapped, hota files must be mapped in hota file format mod
- Moved mapping of h3 prolog / epilog audio mappings to campaign overrides, in line with Chronicles
- Moved mapping of h3 campaign regions to a separate handler. All ID's are now mapped to numeric indices in map format definition.
- Moved hardcoded checks for Yog, Gem, and for RoE to campaign overrides json
- Artifacts, secondary skills and spells now require additional image for campaign bonuses
- Implemented if check for json schemas

TODO: 
- [x] review CGHeroInstance::getValueForCampaign
- [x] test changes
- [x] fix loading of building starting bonus
- [ ] find a way to load or generate hota campaign region images
- [x] recheck schemas and update where necessary
- [ ] recheck docs and update where necessary


IvanSavenko:
PR upgraded to full support for Hota h3c's.

Requires update to hota map format mod for it to actually work

IvanSavenko:
Presumable complete, other than docs (will do soon)

Remaining problem - format of "map pieces" used by hota is different from h3. H3 format was easy to use, but needed like hundreds of images, so kind of makes sense, but hota format is not very suitable for vcmi.

Will think on how to import those files, worst-case scenario - will make small converter tool to generate h3-style images. In any case - this is for separate PR, and probably a little bit later.

lib/constants/EntityIdentifiers.cpp
```diff
@@ -55,11 +55,9 @@ const QueryID QueryID::NONE(-1);
 const QueryID QueryID::CLIENT(-2);
 const HeroTypeID HeroTypeID::NONE(-1);
 const HeroTypeID HeroTypeID::RANDOM(-2);
-const HeroTypeID HeroTypeID::GEM(27);
-const HeroTypeID HeroTypeID::SOLMYR(45);
-const HeroTypeID HeroTypeID::CAMP_STRONGEST(0xFFFD);
-const HeroTypeID HeroTypeID::CAMP_GENERATED(0xFFFE);
-const HeroTypeID HeroTypeID::CAMP_RANDOM(0xFFFF);
+const HeroTypeID HeroTypeID::CAMP_STRONGEST(-3);
```
Laserlicht, line 58: This change may causes https://github.com/vcmi/vcmi/issues/6347

PR #5758, opened 2025-05-30
Czech translation update + decreaseLuck spelling  fix

GeorgeK1ng:
- Updated Czech translation
- Fixed de**s**creaseLuck speliing to decreaseLuck


Mods/vcmi/Content/config/czech.json
XCOM-HUB, line 1: On line 632 - "descreaseLuck" should be "decreaseLuck"

Sometimes you have a space before % - ((${val} %) and 50 %), other times you have no space.

GeorgeK1ng: I don't see anything wrong on line 632 in key descreaseLuck. And good catch from Swedish about space before %, thanks!

XCOM-HUB: Compare "creatures.core.devil.bonus.**descrease**Luck" with the code string below it "creatures.core.boneDragon.bonus.**decrease**Morale" - do you notice the word "decrease" are different in these two strings!? The correct spelling is in the Bone Dragon string while the incorrect spelling is in the Devil string.

GeorgeK1ng: Still don’t see anything wrong against my translation? Matches English used key.
https://github.com/vcmi/vcmi/blob/45a29ac3c33d137179e823936a122e36d58bb84a/Mods/vcmi/Content/config/english.json#L632

XCOM-HUB: Ok, this misspelling is also present in the English version. Who is responsible for the keys? Can the keys be changed somehow, or is it too much hassle?

GeorgeK1ng: Should be easy to fix, will do that in this  PR too. Have not checked misspelling; just left (English) vs right (Czech) files in Beyond Compare.

GeorgeK1ng: Missspelling fix added to this PR. Thanks!

XCOM-HUB: Thank you for wanting to correct a small issue, my good sir. Much appreciated 🤠🎩*Takes off hat

XCOM-HUB: Not to be bothersome or anything, but I have noticed another keyword that could be trimmed/corrected. It's the Angel bonus on line 631. The issue is with the word "raisesMorale", it should be "raiseMorale". I have even double checked it with Chat GPT, and I got the following explanation: Because this is an inherent trait rather than an active action, creatures.core.angel.bonus.raiseMorale makes more sense—it describes a characteristic rather than a continuously triggered event.

PR #5563, opened 2025-03-15
migrate to Conan v2

kambala-decapitator:
- now everything is Conan v2 compatible
- the [dependencies repo](https://github.com/vcmi/vcmi-dependencies) is a submodule now, it contains the major part of `conanfile.py` as well as profiles and custom patches for packages. `conanfile.py` in VCMI repo only contains `generate()` method responsible for CMake integration.
- everything related to vcpkg and mingw cross-compiling was removed (as well as all related CML hacks): MSVC is fully compatible with Conan now.
- merged running (win|mac)deployqt into a single CMake function
- when using Conan on Windows, .bat files that configure `PATH` are created for each produced .exe to be able to run them without copying all the dlls and exes into a single directory
- SDL Java sources were removed from the repo as now they're taken directly from Conan package - no need to lock SDL dependency any more. They're injected into Gradle build via an environment variable as it's invoked from inside `androiddeployqt`.
- minizip source was removed from the repo as vcpkg was the only thing preventing us from this
- Apple platforms are built with Xcode 16.2 on macOS 14 runner now
- added Android x86_64 platform

closes #1608
closes #4571
closes #5965
supersedes & closes #1603

PR #5774, opened 2025-06-08
template editor

Laserlicht:
First version of template editor.

Remaining things for following PR:
- Implement dialog for several zone settings (treasure, custom objects) -> shows `Not implemented` message for now
- Find better, less cross free algorithm for auto positioning
- Improve SVG graphic (maybe increase icon size a little bit)
- Add support for conan platforms (See here https://github.com/vcmi/vcmi/pull/5563 - After merging only cmake presets have to be updated as there is a `ENABLE_TEMPLATE_EDITOR` flag)
- Better handling of linked zones
- validation?

Also existing templates should edited to provide nice visual.

But in general it's already working. But want to merge this first, to avoid too huge PR and maybe take a break.

grafik


config/schemas/template.json
```diff
@@ -43,6 +43,10 @@
 				"terrainTypeLikeZone" : { "type" : "number" },
 				"treasureLikeZone" : { "type" : "number" },
 				"customObjectsLikeZone" : { "type" : "number" },
+
+				"visPositionX" : { "type" : "number" },
+				"visPositionY" : { "type" : "number" },
+				"visSize" : { "type" : "number" },
```
IvanSavenko: Any need for abbreviations? Especially for files that are sort of expected to be edited by modders?
`visiblePositionX` is not that longer, especially compared to other keys in this schema

lib/rmg/CRmgTemplate.cpp
```diff
@@ -144,14 +144,17 @@ ZoneOptions::ZoneOptions():
 	type(ETemplateZoneType::PLAYER_START),
 	size(1),
 	maxTreasureValue(0),
-	owner(std::nullopt),
+	owner(PlayerColor(0)),
```
IvanSavenko, line 147: `PlayerColor(0)` is Red. Perhaps you wanted -1/NEUTRAL?

mapeditor/templateeditor/graphicelements.h
IvanSavenko, line 1: Might be better to split into CardItem.h/LineItem.h.

Clearer and saves some time on navigation. Same for corresponding .cpp's

mapeditor/templateeditor/graphicelements.cpp
```diff
@@ -0,0 +1,292 @@
...
+	std::map<PlayerColor, std::pair<QString, QString>> colors =
+	{
+		{ PlayerColor(0), { "#F80000", "#920000" } }, //red
+		{ PlayerColor(1), { "#0000F8", "#000092" } }, //blue
+		{ PlayerColor(2), { "#9B7251", "#35271C" } }, //tan
+		{ PlayerColor(3), { "#00FC00", "#009600" } }, //green
+		{ PlayerColor(4), { "#F88000", "#924B00" } }, //orange
+		{ PlayerColor(5), { "#F800F8", "#920092" } }, //purple
+		{ PlayerColor(6), { "#00FCF8", "#009694" } }, //teal
+		{ PlayerColor(7), { "#C07888", "#5A3840" } }, //pink
```
IvanSavenko: `class Graphics` already contains player colors. Or these are different?

mapeditor/templateeditor/graphicelements.cpp
```diff
@@ -0,0 +1,292 @@
...
+{
+	std::map<GameResID, QString> names =
+	{
+		{ GameResID::WOOD,    "Wood"    },
+		{ GameResID::ORE,     "Ore"     },
+		{ GameResID::MERCURY, "Mercury" },
+		{ GameResID::SULFUR,  "Sulfur"  },
+		{ GameResID::CRYSTAL, "Crystal" },
+		{ GameResID::GEMS,    "Gems"    },
+		{ GameResID::GOLD,    "Gold"    },
```
IvanSavenko: Why not get translatable name? Or if you need - string ID versions from `StringConstants` / `GameResID::encode`?

mapeditor/templateeditor/graphicelements.cpp
```diff
@@ -0,0 +1,292 @@
...
+	textElem.firstChild().setNodeValue(val ? QString::number(val) : "");
+
+	auto iconElem = getElementById(doc, "icon" + names[res]);
+	iconElem.setAttribute("opacity", val ? "1.0" : "0.1");
+}
+
+void CardItem::setChestValue(int val)
+{
+	auto textElem = getElementById(doc, "textChest");
+	textElem.setAttribute("style", textElem.attribute("style").replace(QRegularExpression("fill:.*?;"), "fill:" + QColor(useBlackText ? Qt::black : Qt::white).name() + ";"));
```
IvanSavenko: Why you need regexp here? For me it feels like unnecessary operation - you generate text someplace else, and then in here you want to transform it instead of regenerating.

mapeditor/templateeditor/graphicelements.cpp
```diff
@@ -0,0 +1,292 @@
...
+	textElem.firstChild().setNodeValue(val ? QString::number(val) : "");
+
+	auto iconElem = getElementById(doc, "iconChest");
+	iconElem.setAttribute("opacity", val ? "1.0" : "0.1");
+}
+
+void CardItem::setSword(EMonsterStrength::EMonsterStrength val)
```
IvanSavenko:
```suggestion
void CardItem::setMonsterStrength(EMonsterStrength::EMonsterStrength val)
```
More clear?

mapeditor/templateeditor/algorithm.h
IvanSavenko, line 1: Perhaps GeometryAlgorithm?

mapeditor/helper.cpp
```diff
@@ -77,6 +81,38 @@ std::shared_ptr<CampaignState> Helper::openCampaignInternal(const QString & file
 		throw std::runtime_error("Corrupted campaign");
 }
 
+std::map<std::string, std::shared_ptr<CRmgTemplate>> Helper::openTemplateInternal(const QString & filenameSelect)
```
IvanSavenko, line 80: This feels like copy-pasted from someplace else. Reuse code perhaps?

mapeditor/templateeditor/templateeditor.cpp
```diff
@@ -0,0 +1,1067 @@
...
+	if(templates.empty())
+		return;
+
+	auto & zones = templates[selectedTemplate]->getZones();
+	if(autoPosition || std::all_of(zones.begin(), zones.end(), [](auto & item){ return item.second->getVisPosition().x == 0 && item.second->getVisPosition().y == 0; }))
+		autoPositionZones();
+	
+	for(auto & zone : zones)
+		if(zone.second->getVisSize() < 0.01)
+			zone.second->setVisSize(1.0);
```
IvanSavenko: Perhaps handle it on loading of zone itself?

lib/rmg/CRmgTemplate.cpp
Laserlicht: The problem here is that the init state is not valid in itself.
For ETemplateZoneType::PLAYER_START we need a valid player.

mapeditor/templateeditor/graphicelements.cpp
Laserlicht: Yes, want to make it more robust. `style` contains many elements, not just `fill`. But I only want to set `fill` in c++ code. The other settings should be setted with svg editor (as they don't need to change in runtime).

Would be also possible with a placeholder and replace. But SVG itself should be valid.

mapeditor/templateeditor/graphicelements.cpp
Laserlicht: Don't think it make sense here?

Because I need two colors one brighter and one darker. Yes, could be done with some color conversations.
But I think it's better like this (less over engineered) and fine adjust it by hand.

Issue #5788, opened 2025-06-13
Commander can wear only 6 artifacts [WoG mod]


misiokles:
In commander window there's room for 9 commander's artifacts but it is possible to wear only six.

1. Launch game with WoG mod and Commanders submod enables)
2. Use vcmiartifacts cheat code to get all available artifacts
3. Try to put commanders artifacts into commander button
4. After putting sixth artifacts, game refuses to wear another.

**Expected behavior**
All nine slots for artifacts can be filled

**Actual behavior**
Only first six slot can be occupied.

**Did it work earlier?**
Yes, but I don't know when it was broken.

**Screenshots**

![Image](https://github.com/user-attachments/assets/ab5e3309-2561-4f19-b17a-da10e58c9fe8)

**Version**

- OS: Windows
- Version:  latest develops

PR #5791, opened 2025-06-14
9 commanders artifacts fix

GeorgeK1ng:
Fixes https://github.com/vcmi/vcmi/issues/5788

- Rectreated commanders artifacts background from VCMI asssets repository with adjusted artifacts grid
- Fixed 9 commanders artifacts, was able to place only 6 for 9 places

From this:
image

To this:
image


DjWarmonger:
Well, originally in WoG there were only 6 slots AFAIK. So it's not the problem with code, but with GUI which shows too many slots.

misiokles:
No, also six slots
image
But it's good to have nine :)

PR #5794, opened 2025-06-15
fix #5742

### LINKED ISSUE #5742: Game crash when casting Magic Arrow on enemy Commander during siege
**Describe the bug**
Casting the Magic Arrow spell on the enemy Commander during a siege battle may cause the game to crash.
The issue has been observed during a siege, but it is not yet confirmed whether it also affects non-siege battles or other types of targets.

**Game logs**

[log.zip](https://github.com/user-attachments/files/20430265/log.zip)

**To Reproduce**

1. Start a siege battle against an enemy with a Commander.

2. Cast Magic Arrow and target the enemy Commander.

3. The game may crash immediately after casting the spell.

**Expected behavior**
The spell should resolve normally and deal damage to the target.

**Actual behavior**
A crash may occur when the spell is cast on the Commander during a siege.

**Version**

- OS: Windows, Android
- Version: 1.7 develop

kdmcser:


dydzio0614:
Assert failure indicates situation that should not be possible to happen - can you explain circumstances that can lead to assert failure and why this situation should be allowed rather than other code fixed the way that chargedArts.empty() never happens?

kdmcser:
This is a temporary fix, and I know how to reproduce the issue: launch Good to Go, enter the cheat code vcmiartifacts, then use the vcminahar cheat to attack the bottom-right town, and cast Magic Arrow on the enemy commander. I've been sick recently and don't have much time to investigate the root cause, so I'm applying a temporary workaround to reduce crashes. I'd really appreciate it if you could help look into the issue. Thanks!

server/CGameHandler.cpp
```diff
@@ -4370,7 +4370,10 @@ void CGameHandler::useChargeBasedSpell(const ObjectInstanceID & heroObjectID, co
 		}
 	}
 
-	assert(chargedArt.has_value());
+
+	if(!chargedArt.has_value())
+		return;
```
thehonestcto, line 4375: To my mind, we should at least log the situation now that the assert is out, otherwise it will become very hard to trace and remember. Maybe with a one liner explanation of why this quick fix is added and how to actually look deeper.

IvanSavenko:
Closing, superseded by already merged PR by SoundsGood

Issue #5790, opened 2025-06-13
Heroes with neutral units don't receive movement penalty for rough terrain


Mich264:
Actual behavior
Heroes with neutral units don't receive movement penalty for rough terrain

To reproduce bug:
1. Hero without pathfinding should have army consisted only with neutral units (example: azure dragons, sharpshooters).
2. Move this hero thought swamp or other rough terrain.
3. After checking, movement point cost is standard 100 for straight move and 141 for diagonal move.

Expected behavior
Terrain penalty for this hero should be standard.
Only exception for this should be pathfinding skill and nomad special ability giving hero no penalty on sand.

Issue #5785, opened 2025-06-11
Familiar Magic Channel works infinitely


Mich264:
Describe the bug and actual behavior
Familiar Magic Channel works infinitely

To recreate bug you need:
Fight with hero with familiar(s)
Destroy all familiar stacks
Cast a spell and check hero mana points for hero who had familiar stacks.

Expected behavior
Magic Channel should stop working when all familiar stacks are destroyed

GeorgeK1ng:
What VCMI version do you use?

Mich264:
VCMI 1.6.8 OS Windows 11

Issue #5715, opened 2025-05-15
Victory condition not triggered after defeating last enemy hero


kdmcser:
**Describe the bug**
If the player captures all enemy towns and defeats the last enemy hero on the adventure map, the game does not immediately recognize the enemy faction as defeated, and no victory screen is shown. The game continues as if the enemy is still active. To trigger the victory condition, the player must engage in and win one additional battle, regardless of its relevance. This is likely due to the defeat condition check not being triggered correctly after the final enemy hero is defeated outside of town combat.

**To Reproduce**

1. Create a map with at least two factions: the player and an enemy.

2. Gradually capture all enemy towns so that the enemy has no towns remaining.

3. Locate and defeat the enemy’s last remaining hero on the adventure map.

4. Notice: the game does not trigger the victory screen at this point, and the enemy faction is still considered active.

5. Engage in any additional battle afterward (e.g., attacking neutral creatures or recapturing an empty town).

6. Only after winning this extra battle will the game recognize the enemy’s defeat and display the victory screen.

**Version**

- OS: Windows, Android
- Version: 1.7 develop

PR #5797, opened 2025-06-15
Fix for regressions from previous PR's

IvanSavenko:
- Fix crashes on loading some mods, e.g. hota
- Add docs on FEARFUL bonus


IvanSavenko:
Added:
- support for limiters in stack experience bonuses
- Fix for #5785
- Fix for #5790
- Fix for #5715

PR #5802, opened 2025-06-17
Remove aiValue from spells to clear warnings

GeorgeK1ng:


config/creatures/special.json
```diff
@@ -77,7 +77,7 @@
 	{
 		"special" : true,
 		"index": 146,
-		"level": 0,
+		"level": 5,
```
IvanSavenko, line 80: Might lead to side effects?
"Level" likely used for things other than specialty

GeorgeK1ng: If it's not needed, can be reverted. Was reported in console so I was thinking it should be default level 5.

IvanSavenko: Huh, interesting. Tested Deemer/Meteor Shower specialty in H3. And it looks like it treats Ballista as level 5 unit as well. So change is correct.

Haven't tested other spell specialists, and don't see any other mechanics that can be affected by Ballista level

PR #5847, opened 2025-06-24
Fix incorrect movement gain from objects like Rally Flag

### LINKED ISSUE #5845: Rally flag give too much movepoints
initially Image
after the visit Image

develop 1.7.0.de893a35752a5dbd8ac6ab545064a857004392c5, msvs
vcmi extras

Looks like regression. Other objects of movepoint is also broken. For example Fountain of youth

IvanSavenko:
Checked all possible places that can change move points - looks like all of them expect absolute values.

- Fixes #5845


GrafDeVan:
works as intended now

PR #5850, opened 2025-06-25
fix path not updated after cheat

### LINKED ISSUE #5846: Path not updated after cheat
**Describe the bug**
Path is not updated after cheat

**To Reproduce**

1. Start map
2. use all movement points with hero
3. select new destination (path is red)
4. use `vcmigod`
5. path is still red and will get only green if clicking on other point

**Expected behavior**
Path should be green immediately after using cheat

**Version**

- OS: Linux
- Version: develop

Laserlicht:
fix #5846

### DIFF
--- a/client/CPlayerInterface.cpp
+++ b/client/CPlayerInterface.cpp
@@ -498,7 +498,7 @@ void CPlayerInterface::heroMovePointsChanged(const CGHeroInstance * hero)
 	if (makingTurn && hero->tempOwner == playerID)
 		adventureInt->onHeroChanged(hero);
 	invalidatePaths();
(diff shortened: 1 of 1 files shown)


client/CPlayerInterface.cpp
```diff
@@ -498,6 +498,7 @@ void CPlayerInterface::heroMovePointsChanged(const CGHeroInstance * hero)
 	if (makingTurn && hero->tempOwner == playerID)
 		adventureInt->onHeroChanged(hero);
 	invalidatePaths();
+	GAME->interface()->localState->verifyPath(hero);
```
IvanSavenko:
```suggestion
	localState->verifyPath(hero);
```

PR #5851, opened 2025-06-25
max range for spell cast

### LINKED ISSUE #5796: Hand to hand spell cast.
Add New bonus or limiter that allow spellcast only when creature is neighbor to each other. It will be useful to convert mechanic ability from Factory.

Laserlicht:
Fixes #5796

can used like this:

```
			"spellcaster":
			{
				"type" : "SPELLCASTER",
				"subtype" : "spell.protectAir",
				"val" : 2
			},
			"spellcasterRange" : 
			{
				"type" : "SPECIFIC_SPELL_RANGE",
				"subtype" : "spell.protectAir",
				"val" : 1
			},
```


Alucard648:
Will it be possible to also do LINE_OF_SIGHT limiter for spells and creature abilities, ala H4?

Laserlicht:
> Will it be possible to also do LINE_OF_SIGHT limiter for spells and creature abilities, ala H4?

Not sure what you mean. With this bonus you can limit spell usage of creatures (SPELLCASTER) to a maximum range of hex tiles.

Alucard648:
LINE_OF_SIGHT means that the straight line between caster and spell target must mot intersect obstacles and other units.

Laserlicht:
> LINE_OF_SIGHT means that the straight line between caster and spell target must mot intersect obstacles and other units.

Ah, understood, No. not supported yet.

lib/battle/CUnitState.cpp
```diff
@@ -407,6 +407,11 @@ int32_t CUnitState::getCasterUnitId() const
 	return static_cast<int32_t>(unitId());
 }
 
+BattleHex CUnitState::getCasterPosition() const
+{
+	return getPosition();
+}
+
```
IvanSavenko: This won't works as intended for double-wide units. For example hex immediately behind unit would have distance of 1, even though it is adjacent to unit

Perhaps instead use interface like this?
```cpp
bool canCastAtTarget(const spells::Spell * spell, const BattleHex & hex)
```

For unit this would check distance, adjusted for double-wide units, and for heroes this would be empty method that always returns true

IvanSavenko: Oh, even more - what if *target* is double-wide?

So this method should either accept an overload with battle::Unit as 2nd argument, or accept spell::Target (or what's its name) class

edeksumo:
How it would work when creatures are far away from each other. Caster will come closer targeted creature?

Laserlicht:
> How it would work when creatures are far away from each other. Caster will come closer targeted creature?

Not exactly sure what you mean.

If the distance between caster and target is to big, no casting possible. If they get closer together casting will allowed.

edeksumo:
> > How it would work when creatures are far away from each other. Caster will come closer targeted creature?
> 
> Not exactly sure what you mean.
> 
> If the distance between caster and target is to big, no casting possible. If they get closer together casting will allowed.

In hota, when Mechanic isn't next to  Dreadnought but in range of movement, Mechanic will come next to Dreadnought and fix him. All takes one turn.

Laserlicht:
> > > How it would work when creatures are far away from each other. Caster will come closer targeted creature?
> > 
> > 
> > Not exactly sure what you mean.
> > If the distance between caster and target is to big, no casting possible. If they get closer together casting will allowed.
> 
> In hota, when Mechanic isn't next to Dreadnought but in range of movement, Mechanic will come next to Dreadnought and fix him. All takes one turn.

Ah. Not possible in one turn yet.

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -208,6 +209,45 @@ bool BattleSpellMechanics::canBeCast(Problem & problem) const
...
+		return true;
+
+	auto spell = getSpell();
+	int range = caster->getEffectRange(spell);
+
+	if(range <= 0)
+		return true;
+
+	std::vector<BattleHex> casterPos = { caster->getCasterPosition() };
+	BattleHex casterWidePos = battle()->battleGetStackByID(caster->getCasterUnitId(), false)->occupiedHex();
```
IvanSavenko: If we're already retrieving stack by unit ID then what's the need for `getCasterPosition` method? We can retrieve its position via unit just as well.

IvanSavenko (approved):

by003:
Like the spark spell in HoMM4, hero should move first then cast. But now I found the range is available, but need to cost one turn to move. Maybe to add a parameter: `'move':true` ? If true, caster can move to close the target, like spark in HoMM4 and mechanic in HotA. If false, caster can only cast for limit range without move, like dreadnought in HotA or red dragon in HoMM5.

PR #5859, opened 2025-06-30
Lag compensation for multiplayer

### LINKED ISSUE #1133: Hero movement is very slow when playing online

### LINKED ISSUE #5490: Anti-lag support for multiplayer using rollback mechanism
Currently multiplayer is not very convenient due to notable network lag even on small actions, like hero movement, army rearrangement, and many others.

### Considered solutions

- Emulate event in UI without changing underlying gamestate: this is what I attempted in #4745, however I quickly discovered that VCMI relies too much on current state of game and it would requires a lot of work for every possible action, including handling of cases when server denies our action, and will be very unstable and easy to break.

- Maintain two copies of gamestate: confirmed gamestate, as received from server, and unconfirmed gamestate that includes actions which were not yet confirmed by server, generated by local copy of GameHandler. For rollback, we'll need to clone confirmed gamestate and reapply all unconfirmed actions. Hard to implement for several reasons: 1. Our serializer is not that fast, for example saves take up to 100 ms on XL maps, and way more on G maps on mobiles. 2. Our UI & AI store a lot of pointers to various game objects, and different gamestates would have different set of pointers. So no easy way to swap in correct gamestate on failure 3. Not clear what to do with random events. If player receives a levelup, server WILL send pack with skills that WILL be different from rolled locally. And this would look rather bad in UI.

- We could also switch from TCP to UDP & implement direct connections between players (instead of using lobby as relay). However this will not help when players are far away. Dydzio mentioned attempts to play between Poland & Japan, which would have ~200 ms ping even in case of peer-to-peer connection. And even in case of (relatively) local games like Poland & Poland, there would still be some delay. Not a bad idea, but won't solve lag.

### Preferred solution

To minimize such issues & make implementation easier, most suitable solution for VCMI is probably this:
- Client maintains its own, slightly customized version of GameHandler that can not use random generator (e.g. throw & catch on access attempts)

- When client sends pack to server as guest in MP, in addition to sending pack, client would also attempt to apply it locally. However local GameHandler can deliberately fail applying it if this pack requires some random rolls or any other actions that GameHandler can't replicate.

- For each pack generated by such local pseudoserver, game must also generate "antipack" - a pack that when applied on client would completely revert all changes made by such pack. For example, pack that modifies movement points should generate antipack that would set movement points to amount hero had before applying.

- If local server failed to apply action (for example due to usage of RNG) or if some pack failed to prepare "antipack" (for example due to complex or not implemented action), then anti-lag for this action won't be applied and game would wait for correct reply from server, as usual.

- When client receives actual replies on its action from server, it must compare received netpacks with packs that were generated locally, and if any differences are found - rollback every local action and reapply correct packs from server

- All packs sent by client must have marker to indicate which request generated it - it is possible for client during simturns that is waiting for reply from server to receive netpacks related to action of another player. So client must be able to understand which packs are reply on its actions, and which packs were initiated by other players.

### Downsides of this solution

- Each netpack has to be adapted manually to support generation of "antipacks"
- Possibility of desync if "antipack" fails to correctly restore previous state and correct netpack have not overwrote all broken data
- Events like levelup would *always* have lag since its result (skill selection) depends on secret state of server RNG (not fixable by design if we want cheat-proof MP)
- GameHandler would need changes to support such approach. Hard to say how complex such changes would be.

### AI & autocombat

Currently AI is purely client-sided. For player AI this is probably OK - it runs on host with 0 ping, even though it might theoretically lead to a cheats where host player modifies AI to give human player advantage, e.g. send resources / focus on other human players. However this is purely theoretical, especially since MP is often played without AI opponents.

However for battle AI this is a bit different:
- Neutral AI for wandering monsters & such should run on server to prevent cheating using method similar to described above

- AI for quickcombat(and for "finish with quickcombat" feature) currently runs on player that initiated event. However this leads to long quickcombat resolution for non-host player since for each action AI waits for server reply (+ping delay). Such AI needs to run either on server, or on host.

- AI for "finish with quickcombat" needs to run only on server for proper implementation of game timers - we want server to actually run AI, since othervice once timer runs out we have to trust that client actually runs AI and not allowing player to keep controlling the combat.

So proper solution that would solve most of these problems would be to move quickcombat & neutral combat AI to server, but keep autocombat for both human & AI players on client - since player needs to be able to cancel it at any moment, and AI may want to either consider retreats, or potentially - add some interaction between adventure AI & combat AI (e.g. mana spending limits)

IvanSavenko:
VCMI is now able to predict result of applied netpack, and can apply it immediately, without waiting for server reply. This allows to remove visible lag when playing as guest in multiplayer.

Selected solution would require supporting prediction & rollback for each individual netpack, and not possible in some scenarios - when result depends on either random generators or on private state of server. Meaning that in some cases predictions are impossible by design. For example - gaining a level can't be predicted since primary skill & secondary skill rolls are decided by server.

Supporting rollback for every possible netpack is rather large amount of effort, so current goal is to provide compensation for most obvious sources of lags, such as hero movement, or in places where players often use rapid actions, like troops rearrangements in garrison (hero/town armies).

At the moment this is still WIP - only basic hero movement can be applied & verified, with no rollback support, but since most of groundwork is done, remaining parts should be relatively straightforward to implement.

### Changes

- Added `AntilagServer` class, that is owned by `CServerHandler`, and that is capable of predicting & generating replies to packs that are sent to remote server, as well as verifying, and rolling back if needed replies.
- `CVCMIServer` now implements `IGameServer` interface. `CGameHandler` now only needs access to this interface instead of entire `CVCMIServer`. This allows maintaining fake `CGameHandler` on client side
- Removed access to `CConnection`s  (renamed to `GameConnection`) from `CGameHandler`, which now only has access to connection ID.
- Replaced plain integers with enum's for connection ID's and connected player ID's
- Removed remaining parts of broken reconnection logic, due to discovering more bugs in those.
- List of fow tiles now uses std::set instead of unordered_set to force guaranteed ordering on any platform

### Fixed bugs

- Fixes #5490
- Fixes #1133

### TODO

- [x] decide what to do with unordered_set/map's in netpacks - since order of elements is unspecified, serialized buffers of identical unordered containers may be different, which breaks prediction verification.
- [x] test & fix or disable prediction for other cases of hero movement - embarking, teleportation, object visiting
- [x] implement prediction support for troops rearrangement
- [x] consider removing persistent CGameHandler and create temporary one for every prediction
- [x] implement pack rollback in case of prediction failure
- [ ] rollback & reapply predicted pack on top if client receives pack not related to predicted request - it is possible for our predicted pack to become invalid due to actions of another player
- [ ] block access to RNG for fake server
- [x] add chat command to enable/disable lag compensation
- [x] review code & add comments where necessary


dydzio0614:
Battle actions should also be subject to rollback - unit animations playing instantly after player makes move will hide the lag of waiting for AI unit response or getting morale info etc.

IvanSavenko:
> Battle actions should also be subject to rollback - unit animations playing instantly after player makes move will hide the lag of waiting for AI unit response or getting morale info etc.

Can try, but probably only after this PR.

Goal of this PR is to add support for lag compensation in general. Once it is done and works reliably - it should be relatively easy to extend it to other packs, like battle actions.

Battle actions are rather complex, since A LOT can happen once it start - obstacle (e.g. quick sands), various chance-based abilities, etc. Predicting damage also won't be possible in general (unless attacker under Bless/Curse) since it involves rolling server-only RNG.

But at the very least starting action (e.g. movement animation) should be possible, and while it plays server will send correct result.

Morale should be not a problem even in current build - since morale info is send basically instantly after your unit starts moving, but queued on client till animations are over.

IvanSavenko:
Hero movement should now work without lag.

Now investigating stack rearrangement. More complex than I expected since we also need to restore stack experience & stack artifacts during rollback to their initial values, but should be doable.

IvanSavenko:
For now assuming to be complete. Currently lag compensation is only supported for hero movement & for troop movement (but only when there is no stack artifacts or stack experience). Can be extended for more actions with relative ease, but in separate PR(s).

Lag compensation is active by default for guests in multiplayer, can be enabled or disabled with `antilag` command in game chat.

PR #5866, opened 2025-07-02
Updated changelog for 1.7

IvanSavenko:


dydzio0614:
I do not see stuff from my PR:

-Allow "finish battle with autocombat" during tactics phase
-Remove default 16 total heroes limit
-Change scroll direction for horizontal slider to match HD mod and hve potentially better ergonomics
-Allow fleeing while having shackles of war when other side got no hero - removes SoD limitation that has no benefits, so this is basically shortcut for putting away chains artifact before battle

IvanSavenko:
@dydzio0614 These changes are already in changelog - see #5756

Issue #5868, opened 2025-07-03
GlideX: Right-Click Not Working


przemsiedl:
## Describe the bug
When using GlideX to emulate a Bluetooth mouse on my PC, the right-click functionality does not work within VCMI. The connection is established via Driver Unification, and the PC successfully emulates a Bluetooth mouse.

## To Reproduce
1. Steps to reproduce the behavior:

2.  Connect GlideX via Driver Unification to your PC.

3. Ensure GlideX is set to emulate a Bluetooth mouse.

4. Open VCMI and start the game.

5. Attempt to use the right-click function of the emulated mouse within VCMI.

## Expected behavior
The right-click function of the GlideX emulated mouse should work correctly within VCMI, similar to how it operates in other applications.

## Actual behavior
The right-click function of the GlideX emulated mouse does not register or work within VCMI.

## Version

OS: Android 13 - Realme GT Master Edition

Version: 1.6.8

Additional context
The right-click functionality does work when using the GlideX emulated mouse in an Android browser and in Magic DosBox, suggesting the issue is specific to VCMI's interaction with the emulated mouse's right-click input.

kambala-decapitator:
Run any SDLv2 example that supports mouse right click on your android and see if it works there

IvanSavenko:
Please check if "Handle back as right mouse button" option in Launcher fixes this

przemsiedl:
@kambala-decapitator I'm not certain if Free Heroes 2 from the Google Store is an SDLv2 app, but I can confirm that both left and right mouse clicks work in it. The mouse movement isn't quite right, but that's beside the point.

It's no priority, just something that would be nice if it worked. You could do one thing on your computer and another on your phone using a single keyboard and mouse set. Heroes on the phone and work on the computer – sounds pretty good!


@IvanSavenko  No it doesn't. Checked it many times before.

IvanSavenko:
Both fheroes2 and SDL have `android.hardware.type.pc` feature enabled. And according to docs, it "disables input emulation for mouse and touchpad". Perhaps this is the reason?

Can try making such build and see if it fixes anything

https://developer.android.com/guide/topics/manifest/uses-feature-element#device-ui-hw-features
https://github.com/ihhub/fheroes2/blob/master/android/app/src/main/AndroidManifest.xml#L28

IvanSavenko:
@przemsiedl 

Can you check whether this build fixes mouse behavior for you?
[...]

PR #5869, opened 2025-07-03
Update Android app configuration

IvanSavenko:
- Added uses-feature flags in line with SDL app. Hopefully this should:
  - fix handling of right mouse button (#5868)
  - make VCMI available on TV's running Android - these devices don't have `touchscreen` feature, which is added as required by default
- Unified AndroidManifest formatting - was mix of tabs & spaces

NOTE: targeting master, but most likely will retarget to develop later unless we need to upload build to Google Play before 1.7


kambala-decapitator:
any help needed?

IvanSavenko:
@kambala-decapitator 

Yeah, probably... Changes that are right now in this PR seems to be OK (or at the very least - don't break anything). Can't check bluetooth mouse since I don't have one.

But using newer SDK is more problematic... New Gradle apparently requires Java 17, but switching to 17 break build of Qt5 bindings. And so far I was unable to correctly set up cross-compilation to use Java 17 for Gradle, but Java 11 as compiler. Any suggestions?

Here is my last attempt:
- Commit: https://github.com/IvanSavenko/vcmi/commit/f93f112dc831877c1d8beb1ef127ab67360ff9aa
- Github Actions: https://github.com/IvanSavenko/vcmi/actions/runs/16089048933/job/45403724382

There is also 1 hack: Qt checks for presence of `useLegacyPackaging` flag, and if not found - adds one more switch, that is no longer supported by Gradle. We do have that switch, but only in vcmi-app Gradle config, while Qt apparently only checks global config. Not sure how to move that flag to global scope. Would be nice to fix it as well, but not as critical as Java version fix.

Any ideas?

kambala-decapitator:
upgrading gradle and AGP: I've recently done that in another app that uses Qt 5, will do. A few hacks are required, will do.

haven't seen any issues with Java 17 on local machine, maybe something CI-related, will check.

probably upgrading to SDK 35 (or 36) wouldn't require any changes from our side, but will read release notes as well.

kambala-decapitator:
done, this commit applies on top of the current branch https://github.com/vcmi/vcmi/commit/dbd73ca21a11ce52cbdeec03cca98938eda87c6a (oops, forgot to bump JDK to 17 on CI)

one thing I noticed: launcher is now cut on edges, seems related to https://developer.android.com/about/versions/15/behavior-changes-15#window-insets

android 2025-07-07 22 42 48

kambala-decapitator:
@IvanSavenko so, how would you like to proceed here?

IvanSavenko:
Not sure. Would be great to find a way to fix those cutoff edges.

Main question I guess is whether we can find nice/proper way to fix it, or need to use some hacks.

From what I see, Qt supports this edge-to-edge mode only in 6.5.10 and later. But it looks like there is a way to opt-out? Not familiar with Android so not sure how to enable that flag: https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement

In any case we don't really have a choice - without API bump we won't be able to upload updates starting next month.

kambala-decapitator:
not sure why we need it in master right now

IvanSavenko:
Yeah, my mistake

PR #5877, opened 2025-07-04
remove border in battle

Laserlicht:
I think we should remove window border for battle window. It does not look that nice with the panels.

Also fine tunes turn timer layout.

Replaced with 1px border:
grafik

### DIFF
--- a/client/battle/BattleWindow.cpp
+++ b/client/battle/BattleWindow.cpp
@@ -218,8 +218,8 @@ void BattleWindow::hideStickyQuickSpellWindow()
 	quickSpellWindow->disable();
 	unitActionWindow->disable();
 
-	setPositionInfoWindow();
 	createTimerInfoWindows();
+	setPositionInfoWindow();
 	ENGINE->windows().totalRedraw();
 }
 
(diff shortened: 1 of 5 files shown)


client/battle/BattleWindow.cpp
```diff
@@ -256,15 +256,15 @@ void BattleWindow::createTimerInfoWindows()
 		if (attacker.isValidPlayer())
 		{
 			if (ENGINE->screenDimensions().x >= 1000)
```
IvanSavenko: I think we can now reduce this requirement a bit? Would have been great to squeeze it into 960px (which is screen width on 16x9 with maxed-out scaling)

client/battle/BattleWindow.cpp
```diff
@@ -191,10 +191,10 @@ void BattleWindow::createQuickSpellWindow()
 	OBJECT_CONSTRUCTION;
 
 	quickSpellWindow = std::make_shared<QuickSpellPanel>(owner);
-	quickSpellWindow->moveTo(Point(pos.x - 67, pos.y));
+	quickSpellWindow->moveTo(Point(pos.x - 55, pos.y));
 
 	unitActionWindow = std::make_shared<UnitActionPanel>(owner);
-	unitActionWindow->moveTo(Point(pos.x + pos.w + 15, pos.y));
+	unitActionWindow->moveTo(Point(pos.x + pos.w + 3, pos.y));
```
IvanSavenko: Perhaps place them even closer and completely remove gap between main battle window and these dialogs?

client/battle/BattleWindow.cpp
Laserlicht: Yes 955px are needed. Will work.

Laserlicht:
grafik

PR #5897, opened 2025-07-07
Fix magic_mirror spell

Opuszek:
Fixes most issues related to magic mirror spell mentioned in the issue 5786.

Bug 1 Magic Mirror don't defend affected unit against spells when trigger.
Fixed.

Bug 2 Redirected spell have fraction of his spell power
Fixed.

Bug 3 Magic Mirror and Hypnotize - In H3 Magic Mirror has feature that unit protected by Magic Mirror was immune to Hypnotize.
Fixed. Currently on dev there seems to be game-breaking bug when Hypnotize is used in a hot-seat battle with two players, though. I cleaned and rebuilded from dev to be sure that it is not caused by my changes.  

Bug 4 No information when redirected spell failed
Fixed.

Bug 5 Area of effect spells.
After the changes a magic mirror counters only direct spells, aoe and mass spells are ignored. This behavior is consistent with the original.

Bug 6 Two Magic Mirror.
Fixed. I could not reproduce a chain of magic mirrors in the original (maybe I was unlucky and/or not stubborn enough), but it seems to be logical behavior, so I implemented it.

Bug 7 Magic Mirror and Blind/Berserk Spell
Not fixed (yet).

Current differences from the original:

Animation: In a case when a missile-type spell (for example magic arrow) is reflected to a caster's stack, the animation looks identical to an attack perform on an enemy stack.
In the original the missile starting position is moved a little above the casting hero. I am not sure if it was intentional and if it should be imitated.

Animation: in a case when a missile-type spell is reflected to a caster's stack that is immune to magic (for example black dragons or a stack with Anti-Magic spell) four animation play out: hero casts spells, the first unit reflects, the second resist and the missile strikes the resisting unit. In the original missile animation was omitted.
It is a result of current logic for resisting spells so I consider it outside the scope of pr.

Logic: Bug 7: I plan to fix it, but I would like to do it in a different pr for clarity sake.

PS: I like how each time a caster deals damage to his/her own units a facepalm animation is triggered.
PS2: I am kinda new to C++. I was trying to be careful, but I hope I didn't do anything silly with memory allocation.

### DIFF
--- a/lib/spells/BattleSpellMechanics.cpp
+++ b/lib/spells/BattleSpellMechanics.cpp
@@ -19,7 +19,6 @@
 #include "../networkPacks/PacksForClientBattle.h"
 #include "../networkPacks/SetStackEffect.h"
 #include "../CStack.h"
-#include <boost/assign.hpp>
 
 #include <vstd/RNG.h>
 
--- a/lib/spells/BattleSpellMechanics.cpp
+++ b/lib/spells/BattleSpellMechanics.cpp
@@ -437,7 +437,7 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
 bool BattleSpellMechanics::isReflected(const battle::Unit * unit, vstd::RNG & rng)
 {
 	auto range = owner -> getLevelInfo(getRangeLevel()).range;
-	const auto directSpellRange = boost::assign::list_of(0);
+	const std::vector<int> directSpellRange = { 0 };
 	const std::string magicMirrorCacheStr = "type_MAGIC_MIRROR";
 	static const auto magicMirrorSelector = Selector::type()(BonusType::MAGIC_MIRROR);
 
--- a/lib/spells/BattleSpellMechanics.cpp
+++ b/lib/spells/BattleSpellMechanics.cpp
@@ -406,13 +406,19 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
 			affectedUnits.push_back(unit);
 	};
 
+	if (!target.empty())
+	{
+		const battle::Unit * targetedUnit = battle()->battleGetUnitByPos(target.front().hexValue, true);
+		if (isReflected(targetedUnit, rng)) {
+			reflect(sc, rng, targetedUnit);
+			return;
+			}
+	}
+
 	//prepare targets
 	effectsToApply = effects->prepare(this, target, spellTarget);
 
 	std::set<const battle::Unit *> unitTargets = collectTargets();
-
-	if (unitTargets.size()==1 && isReflected(*(unitTargets.begin()), rng))
-		return reflect(sc, rng, *(unitTargets.begin()));
 
 	//process them
 	for(const auto * unit : unitTargets)
@@ -435,13 +441,10 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
 
 bool BattleSpellMechanics::isReflected(const battle::Unit * unit, vstd::RNG & rng)
 {
-	auto range = owner -> getLevelInfo(getRangeLevel()).range;
-	const std::vector<int> directSpellRange = { 0 };
-	const std::string magicMirrorCacheStr = "type_MAGIC_MIRROR";
-	static const auto magicMirrorSelector = Selector::type()(BonusType::MAGIC_MIRROR);
-
-	bool spellIsDirect = !isMassive() && owner -> getLevelInfo(getRangeLevel()).range == directSpellRange;
-	bool spellIsReflectable = spellIsDirect && (mode == Mode::HERO || mode == Mode::MAGIC_MIRROR) && isNegativeSpell();
+	if (unit == nullptr)
+		return false;
+	bool isDirectSpell = owner->getTargetType() == AimType::CREATURE && !isMassive();
+	bool spellIsReflectable = isDirectSpell && (mode == Mode::HERO || mode == Mode::MAGIC_MIRROR) && isNegativeSpell();
 	bool targetCanReflectSpell = spellIsReflectable && unit->getAllBonuses(Selector::type()(BonusType::MAGIC_MIRROR))->size()>0;
 	return targetCanReflectSpell && rng.nextInt(0, 99) < unit->valOfBonuses(BonusType::MAGIC_MIRROR);
 }
@@ -450,6 +453,8 @@ void BattleSpellMechanics::reflect(BattleSpellCast & sc, vstd::RNG & rng, const 
 {
 	auto otherSide = battle()->otherSide(unit->unitSide());
 	auto newTarget = getRandomUnit(rng, otherSide);
+	if (newTarget == nullptr)
+		throw std::runtime_error("Failed to find random unit to reflect spell!");
 	auto reflectedTo = newTarget->getPosition();
 
 	mode = Mode::MAGIC_MIRROR;
@@ -459,7 +464,7 @@ void BattleSpellMechanics::reflect(BattleSpellCast & sc, vstd::RNG & rng, const 
 	if (!isReceptive(newTarget))
 		sc.resistedCres.insert(newTarget->unitId());    //A spell can be reflected to then resisted by an immune unit. Consistent with the original game.
 
-	beforeCast(sc, rng, { reflectedTo });
+	beforeCast(sc, rng, { Destination(reflectedTo) });
 }
 
 const battle::Unit * BattleSpellMechanics::getRandomUnit(vstd::RNG & rng, const BattleSide & side)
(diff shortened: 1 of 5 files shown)


dydzio0614:
thanks for trying, the code will be reviewed and you will get feedback

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -425,16 +432,50 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
...
 
-	for(const auto * unit : resisted)
-		sc.resistedCres.insert(unit->unitId());
+	beforeCast(sc, rng,
+		boost::assign::list_of(reflectedTo));
+}
+
+const battle::Unit* BattleSpellMechanics::getRandomUnit(vstd::RNG & rng, BattleSide & side)
+{
+	auto targets = battle()->getBattle()->getUnitsIf([this, & side](const battle::Unit * unit)
```
IvanSavenko: FYI: We have various automatic checks for PR's, mostly compiling on various platforms / various compilers and one of those has failed here - due to a warning that only present on clang:
https://github.com/vcmi/vcmi/actions/runs/16124607314/job/45498525834?pr=5897#step:16:1773
```suggestion
	auto targets = battle()->getBattle()->getUnitsIf([&side](const battle::Unit * unit)
```

Opuszek: Thanks. I was trying to figure out what went wrong.

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -425,18 +430,50 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
...
+
+	mode = Mode::MAGIC_MIRROR;
+	sc.reflectedCres.insert(unit->unitId());
+	sc.tile = reflectedTo;
+
+	if (!isReceptive(newTarget))
+		sc.resistedCres.insert(newTarget->unitId());    //A spell can be reflected to then resisted by an immune unit. Consistent with the original game.
+
+	beforeCast(sc, rng,
+		boost::assign::list_of(reflectedTo));
```
IvanSavenko: - No need to use boost when C++ features are enough. Brace-init will construct vector with 1 element
- We only use line breaks on *very* long lines
```suggestion
	beforeCast(sc, rng, { reflectedTo });
```

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -425,18 +430,50 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
...
+}
+
+const battle::Unit * BattleSpellMechanics::getRandomUnit(vstd::RNG & rng, BattleSide & side)
```
IvanSavenko: Prefer passing parameters by const reference when possible
```suggestion
const battle::Unit * BattleSpellMechanics::getRandomUnit(vstd::RNG & rng, const BattleSide & side)
```

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -19,6 +19,7 @@
 #include "../networkPacks/PacksForClientBattle.h"
 #include "../networkPacks/SetStackEffect.h"
 #include "../CStack.h"
+#include <boost/assign.hpp>
```
IvanSavenko: Looks to be used in 1 place where brace-init can be used instead
```suggestion
```

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -410,6 +412,9 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
 
 	std::set<const battle::Unit *> unitTargets = collectTargets();
 
+	if (unitTargets.size()==1 && isReflected(*(unitTargets.begin()), rng))
+		return reflect(sc, rng, *(unitTargets.begin()));
```
IvanSavenko: What about Chain Lightning? Not sure if it has single target at this point

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -425,18 +430,50 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
...
-	}
-
 	for(const auto * unit : resisted)
 		sc.resistedCres.insert(unit->unitId());
 }
 
+bool BattleSpellMechanics::isReflected(const battle::Unit * unit, vstd::RNG & rng)
+{
+	auto range = owner -> getLevelInfo(getRangeLevel()).range;
+	const auto directSpellRange = boost::assign::list_of(0);
```
IvanSavenko:
```suggestion
	const std::vector<int> directSpellRange = { 0 };
```

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -425,18 +430,50 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
...
+	const std::string magicMirrorCacheStr = "type_MAGIC_MIRROR";
+	static const auto magicMirrorSelector = Selector::type()(BonusType::MAGIC_MIRROR);
+
+	bool spellIsDirect = !isMassive() && owner -> getLevelInfo(getRangeLevel()).range == directSpellRange;
+	bool spellIsReflectable = spellIsDirect && (mode == Mode::HERO || mode == Mode::MAGIC_MIRROR) && isNegativeSpell();
+	bool targetCanReflectSpell = spellIsReflectable && unit->getAllBonuses(Selector::type()(BonusType::MAGIC_MIRROR))->size()>0;
+	return targetCanReflectSpell && rng.nextInt(0, 99) < unit->valOfBonuses(magicMirrorSelector, magicMirrorCacheStr);
```
IvanSavenko: And also remove `magicMirrorCacheStr` / `magicMirrorSelector`

This is basically old code. Now such bonus selection logic is only needed in some complex cases, for typical cases where you need to get value of a single bonus type, there is overload that should be fast enough as it is.
```suggestion
	return targetCanReflectSpell && rng.nextInt(0, 99) < unit->valOfBonuses(BonusType::MAGIC_MIRROR);
```

lib/spells/BattleSpellMechanics.cpp
```diff
@@ -425,18 +430,50 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
...
+}
+
+void BattleSpellMechanics::reflect(BattleSpellCast & sc, vstd::RNG & rng, const battle::Unit * unit)
+{
+	auto otherSide = battle()->otherSide(unit->unitSide());
+	auto newTarget = getRandomUnit(rng, otherSide);
```
IvanSavenko, line 455: What if `newTarget` is null? For example if there are only Black Dragons in army?

lib/spells/BattleSpellMechanics.cpp
Opuszek: Chain lighting can be reflected (consisted with HoMM 3 Complete on GOG).
It has single target.

Opuszek: Edit: I double checked and it turns out I was full of poo. Will look into this later when I have time.

lib/spells/BattleSpellMechanics.cpp
Opuszek: A spell can be reflected to an immune stack (like Black Dragons). In such a case a shield animation of unit resisting a spell plays out and unit receives no damage.
This code deals with such situation:
	if (!isReceptive(newTarget))
		sc.resistedCres.insert(newTarget->unitId());
I cannot think of a case when newTarget could be null, but it does not mean it cannot exist. I will add a null check later this day.

IvanSavenko: Ah, apologies, yeah, you're correct. `getRandomUnit` should never return null. It can return null if player has only war machines or turrets, but in this scenario player should not be able to cast spells.

Feel free to add null check, but generally I prefer crashing the app explicitly, for example via `throw std::runtime_error("Failed to find random unit to reflect spell!");`

Silently returning would prevent crash, but it would also hide potential bugs & make them harder to track down.

lib/spells/BattleSpellMechanics.cpp
Opuszek: I've moved checking for reflection earlier in the logic. I've also replaced that hacky range check with (owner->getTargetType() == AimType::CREATURE && !isMassive());
I needed to check for targeted unit calling battle()->battleGetUnitByPos as destination.unitValue returns null at this point.
Chain lighting works now.

lib/spells/BattleSpellMechanics.cpp
Opuszek: I've added a null check.

PR #5901, opened 2025-07-08
(somewhat) configurable adventure map spells

IvanSavenko:
- All adventure map spells have most of their parameters in json, 
- Parameters of adventure map spells can now be defined separately per each mastery level.
- It is now possible to add a new spell that will have effect similar to H3 adventure map spell

Will update docs later, as far as spell configuration goes, following parameters can now be defined for each spell for each master level:
- All spells: limit of casts per day (with alternative setting for XL+ maps for tournament rules)
- Summon boat: whether spell can move existing boats or create new ones
- Scuttle boat: range as well as which objects can be removed (yes, you can make spell that removes specific map objects)
- View Air/Earth: which objects are visible on each level (however no way to assign icons to completely new objects)
- Town portal: movement points cost, whether town can be selected via dialog
- Dimension Door: movement points cost, range, whether it is possible to DD into Fow (previously only existed as global setting)

TODO:
- [x] remove/reduce hardcoded spell ID checks in UI code
- [x] make UI aware of spell parameters, like spell range
- [x] remove/reduce hardcoded spell ID checks in AI code
- [x] update schemas
- [x] update docs


GeorgeK1ng:
Remove or build town spells sounds cool. Hero or creatures removal is bad. 

Town can be removed by timed event in map editor, so it sounds doable. 

Spell to remove mountain sounds cool for example, but it will ruin game play same as hero or creature removal.

Destroy / remove town is not bad. Same as build town from list. These spells should be like level 6 and maybe only on rare scrolls? Same as Summon troops from town.

IvanSavenko:
> can "removeObject" type remove enemy heroes, creatures on map or towns? 

It can remove any object, just like map event. However only visitable objects can be targeted by such spell (since this is how Scuttle Boat worked), so no for mountain removal and such

> can "addObject" type be implemented some time in future? (for example to build towns)

Not planning it myself at the moment. Such spell would also need to configure all parameters of such objects. I think currently only boat and wandering monsters can be created in runtime. Doable, but not a priority for me.

PR #5903, opened 2025-07-08
Hide bonus from creature window

### LINKED ISSUE #5899: Hidden parameter for bonuses
It will be awesome to be able to hide bonuses in creature window. With custom icons and description, it will open endless possibilities for bonuses!

Laserlicht:
fixes #5899

@edeksumo Is that what you wanted?


edeksumo:
🔥🔥🔥

PR #5904, opened 2025-07-08
fix graphical glitch

Laserlicht:
fix graphical glitch reported by @GeorgeK1ng 

Before:
grafik

After:
grafik


GeorgeK1ng:
Just reinstalled Extras on my side and no more glitch! Thanks!

PR #5905, opened 2025-07-09
Fix some sonar warnings and convert formatting from spaces to tabs

IvanSavenko:


mapeditor/templateeditor/GeometryAlgorithm.cpp
```diff
@@ -1,133 +1,135 @@
...
- */
+	/*
+	 * GeometryAlgorithm.cpp, part of VCMI engine
+	 *
+	 * Authors: listed in file AUTHORS in main folder
+	 *
+	 * License: GNU General Public License v2.0 or later
+	 * Full text of license available in license.txt file, in main folder
+	 *
+	 */
```
SoundSSGood: Why do we need to move file header?

IvanSavenko: Wil fix, not intended

PR #5910, opened 2025-07-10
Make a blind/berserk spells activate immediately if reflected on an active unit 

Opuszek:
Fixes the last bug mentioned in an issue 5786 - 
The first commit should not contain any changes to logic.
I needed to return to the old way of checking if a spell is direct, since berserk spell on the first level is classified as area spell with range 0. I tried to change timed.json, buit it seems the engine does not allow changing aim type across levels.
PS: reflecting chain lighting still works.

### DIFF
--- a/server/battles/BattleFlowProcessor.cpp
+++ b/server/battles/BattleFlowProcessor.cpp
@@ -342,6 +342,21 @@ void BattleFlowProcessor::activateNextStack(const CBattleInfoCallback & battle)
 
 bool BattleFlowProcessor::tryMakeAutomaticAction(const CBattleInfoCallback & battle, const CStack * next)
 {
+	bool actionPerformed = tryActivateMoralePenalty(battle, next) || tryActivateBerserkPenalty(battle, next) || tryAutomaticActionOfWarMachines(battle, next);
+
+	if (!actionPerformed) {
+		stackTurnTrigger(battle, next); //various effects
+
+		if(next->fear)
+			{
+				makeStackDoNothing(battle, next); //end immediately if stack was affected by fear
+				return true;
+			}
+		}
+	return actionPerformed;
+}
+
+bool BattleFlowProcessor::tryActivateMoralePenalty(const CBattleInfoCallback & battle, const CStack * next) {
 	// check for bad morale => freeze
 	int nextStackMorale = next->moraleVal();
 	if(!next->hadMorale && !next->waited() && nextStackMorale < 0)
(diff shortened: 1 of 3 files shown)


server/battles/BattleFlowProcessor.cpp
```diff
@@ -342,6 +342,21 @@ void BattleFlowProcessor::activateNextStack(const CBattleInfoCallback & battle)
...
+		stackTurnTrigger(battle, next); //various effects
+
+		if(next->fear)
+			{
+				makeStackDoNothing(battle, next); //end immediately if stack was affected by fear
+				return true;
+			}
+		}
+	return actionPerformed;
+}
```
IvanSavenko: Formatting is broken.
I've added clang-format configuration recently, you can try using it if your IDE supports it.

Also - perhaps simplify this method like this? Right now it is a bit hard to follow the flow
```suggestion
	if (tryActivateMoralePenalty(battle, next))
		return true;

	if (tryActivateBerserkPenalty(battle, next))
		return true;
	
	if (tryAutomaticActionOfWarMachines(battle, next))
		return true;
		
	stackTurnTrigger(battle, next); //various effects

	if(next->fear)
	{
		makeStackDoNothing(battle, next); //end immediately if stack was affected by fear
		return true;
	}

	return false;
}
```

server/battles/BattleFlowProcessor.cpp
```diff
@@ -359,36 +374,46 @@ bool BattleFlowProcessor::tryMakeAutomaticAction(const CBattleInfoCallback & bat
...
+			return true;
 		}
-		return true;
-	}
+	return false;
+}
 
+bool BattleFlowProcessor::tryAutomaticActionOfWarMachines(const CBattleInfoCallback & battle, const CStack * next) {
+	return tryMakeAutomaticActionOfBallistaOrTowers(battle, next) || tryMakeAutomaticActionOfCatapult(battle, next) || tryMakeAutomaticActionOfFirstAidTent(battle, next);
+}
```
IvanSavenko: Similar here.  Chain of 3 `if tryX return true` will be easier to read. We don't have line number limit, but screen width is limited.

And again - formatting - place bracket on next line, aligned to statement above.

```suggestion
bool BattleFlowProcessor::tryAutomaticActionOfWarMachines(const CBattleInfoCallback & battle, const CStack * next)
{
	if (tryMakeAutomaticActionOfBallistaOrTowers(battle, next))
		return true;
	 
	if (tryMakeAutomaticActionOfCatapult(battle, next))
		return true;
	
	if (tryMakeAutomaticActionOfFirstAidTent(battle, next)
		return true;
		
	return false;
}
```

server/battles/BattleFlowProcessor.cpp
```diff
@@ -359,36 +374,46 @@ bool BattleFlowProcessor::tryMakeAutomaticAction(const CBattleInfoCallback & bat
...
+
+					makeAutomaticAction(battle, next, attack);
+					logGlobal->trace("Attacked nearest target %s", attackInfo.first->getDescription());
+				}
+			else
+				{
+					makeStackDoNothing(battle, next);
+					logGlobal->trace("No target found");
+				}
+			return true;
```
IvanSavenko: Formatting - place bracket on next line, aligned to statement above.
```suggestion
			if (attackInfo.first != nullptr)
			{
				BattleAction attack;
				attack.actionType = EActionType::WALK_AND_ATTACK;
				attack.side = next->unitSide();
				attack.stackNumber = next->unitId();
				attack.aimToHex(attackInfo.second);
				attack.aimToUnit(attackInfo.first);

				makeAutomaticAction(battle, next, attack);
				logGlobal->trace("Attacked nearest target %s", attackInfo.first->getDescription());
			}
			else
			{
				makeStackDoNothing(battle, next);
				logGlobal->trace("No target found");
			}
			return true;
```

server/battles/BattleFlowProcessor.cpp
Opuszek: Fixed

server/battles/BattleFlowProcessor.cpp
Opuszek: Fixed

server/battles/BattleFlowProcessor.cpp
Opuszek: Fixed

PR #5916, opened 2025-07-11
Generate overlay & shadow

### LINKED ISSUE #5653: generation of shadow and overlay
**Describe your proposal**
If shadow and overlay (outline) images are missing they should generated by VCMI itself.

**Additional context**
Outline could be implemented with min and max filter.

In python:
```
import numpy as np
from scipy.ndimage import maximum_filter, minimum_filter
from PIL import Image

def get_stroke(origin_image, size):
    img = np.asarray(origin_image).copy()
    img_alpha = img[:, :, 3]
    img_max = maximum_filter(img_alpha, size=size, mode='mirror')
    img_min = minimum_filter(img_alpha, size=size, mode='mirror')
    img[:, :, 0] = 255
    img[:, :, 1] = 255
    img[:, :, 2] = 255
    img[:, :, 3] = img_max - img_min
    return Image.fromarray(np.uint8(img))
```

ChatGPT gives following c++ implementation for `maximum_filter` with promt `implement filter on sdlimage in c++ similar to maximum_filter from scipy.ndimage`.

```
SDL_Surface* maximum_filter(SDL_Surface* surface, int filter_size) {
    if (!surface) {
        std::cerr << "Invalid surface" << std::endl;
        return nullptr;
    }

    // Get pixel format
    SDL_PixelFormat* format = surface->format;
    
    // Create a copy of the surface to store the filtered image
    SDL_Surface* filtered_surface = SDL_ConvertSurface(surface, format, 0);
    if (!filtered_surface) {
        std::cerr << "Failed to create filtered surface: " << SDL_GetError() << std::endl;
        return nullptr;
    }

    int width = surface->w;
    int height = surface->h;

    // Iterate through the pixels of the image
    for (int y = 0; y < height; y++) {
        for (int x = 0; x < width; x++) {
            Uint32 max_pixel = 0;
            int half_filter = filter_size / 2;

            // Loop over the neighborhood around (x, y)
            for (int offset_y = -half_filter; offset_y <= half_filter; offset_y++) {
                for (int offset_x = -half_filter; offset_x <= half_filter; offset_x++) {
                    int neighbor_x = x + offset_x;
                    int neighbor_y = y + offset_y;

                    // Check bounds
                    if (neighbor_x >= 0 && neighbor_x < width && neighbor_y >= 0 && neighbor_y < height) {
                        // Get the pixel at the neighbor position
                        Uint32 pixel = *((Uint32*)surface->pixels + neighbor_y * width + neighbor_x);
                        
                        // Compare the pixel color value to find the maximum
                        if (pixel > max_pixel) {
                            max_pixel = pixel;
                        }
                    }
                }
            }

            // Set the maximum pixel value in the filtered image
            *((Uint32*)filtered_surface->pixels + y * width + x) = max_pixel;
        }
    }

    return filtered_surface;
}
```

Laserlicht:
It's possible to let VCMI create the overlays and shadows for json animations.

See doc changes.

Test creature (ignore transparent pixel of creature... convert issue unrelated to PR):
[axolotl-test.zip](https://github.com/user-attachments/files/21191227/axolotl-test.zip)

fixes #5653

Example of generated shadows and outline:
Peek 2025-07-12 12-59
Peek 2025-07-12 13-01

@IvanSavenko
Shadows and overlays are flickering the first time while loading. This is also the case before this PR by using `-OVERLAY` and `-SHADOW`. Seems a regression. Maybe #5306? Any idea for the reason?

### DIFF
--- a/client/renderSDL/SDL_Extensions.cpp
+++ b/client/renderSDL/SDL_Extensions.cpp
@@ -761,13 +761,41 @@ SDL_Surface* CSDL_Ext::drawOutline(SDL_Surface* sourceSurface, const SDL_Color& 
 
 void applyAffineTransform(SDL_Surface* src, SDL_Surface* dst, double a, double b, double c, double d, double tx, double ty)
 {
+	// Check if the transform is purely scaling (and optionally translation)
+	bool isPureScaling = vstd::isAlmostZero(b) && vstd::isAlmostZero(c);
+
+	if (isPureScaling)
+	{
+		// Calculate target dimensions
+		int scaledW = static_cast<int>(src->w * a);
+		int scaledH = static_cast<int>(src->h * d);
+
+		SDL_Rect srcRect = { 0, 0, src->w, src->h };
+		SDL_Rect dstRect = { static_cast<int>(tx), static_cast<int>(ty), scaledW, scaledH };
+
+		// Convert surfaces to same format if needed
+		if (src->format->format != dst->format->format)
+		{
+			SDL_Surface* converted = SDL_ConvertSurface(src, dst->format, 0);
+			if (!converted)
+				throw std::runtime_error("SDL_ConvertSurface failed!");
+
+			SDL_BlitScaled(converted, &srcRect, dst, &dstRect);
+			SDL_FreeSurface(converted);
+		}
+		else
+			SDL_BlitScaled(src, &srcRect, dst, &dstRect);
+
+		return;
+	}
+
 	// Lock surfaces for direct pixel access
 	if (SDL_MUSTLOCK(src)) SDL_LockSurface(src);
 	if (SDL_MUSTLOCK(dst)) SDL_LockSurface(dst);
 
 	// Calculate inverse matrix M_inv for mapping dst -> src
 	double det = a * d - b * c;
-	if (std::abs(det) < 1e-10)
+	if (vstd::isAlmostZero(det))
 		throw std::runtime_error("Singular transform matrix!");
 	double invDet = 1.0 / det;
 	double ia =  d * invDet;
@@ -781,7 +809,7 @@ void applyAffineTransform(SDL_Surface* src, SDL_Surface* dst, double a, double b
 	tbb::parallel_for(tbb::blocked_range<size_t>(0, dst->h), [&](const tbb::blocked_range<size_t>& r)
 	{
 		// For each pixel in the destination image
-		for(int y = r.begin(); y != r.end(); ++y)	
+		for(int y = r.begin(); y != r.end(); ++y)
 		{
 			for(int x = 0; x < dst->w; x++)
 			{
@@ -796,7 +824,6 @@ void applyAffineTransform(SDL_Surface* src, SDL_Surface* dst, double a, double b
 				// Check bounds
 				if (srcXi >= 0 && srcXi < src->w && srcYi >= 0 && srcYi < src->h)
 				{
-
 					Uint32 pixel = srcPixels[srcYi * src->w + srcXi];
 					dstPixels[y * dst->w + x] = pixel;
 				}


Laserlicht:
Outline looks now exactly like OH3 (all scalings):
grafik

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +685,335 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
+
+			// Nearest neighbor sampling (can be improved to bilinear)
+			auto srcXi = static_cast<int>(round(srcX));
+			auto srcYi = static_cast<int>(round(srcY));
+
+			// Check bounds
+			if (srcXi >= 0 && srcXi < src->w && srcYi >= 0 && srcYi < src->h)
+			{
+				auto srcPixels = (Uint32*)src->pixels;
+				auto dstPixels = (Uint32*)dst->pixels;
```
C3pa: Looks like these two lines can be moved outside of the loops.

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +685,335 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
+
+				Uint32 pixel = srcPixels[srcYi * src->w + srcXi];
+				dstPixels[y * dst->w + x] = pixel;
+			}
+			else
+			{
+				// Outside source bounds: set transparent or black
+				auto dstPixels = (Uint32*)dst->pixels;
```
C3pa: Also here.

Laserlicht:
Optimized.
Now needs ~1,5ms for battle shadow (before it needed ~15 ms).

Laserlicht:
Reverted use of algorithm on optimized image... It doesn't work correct (slightly jumping shadow in Y)... No idea why. Doesn't seem a rounding issue (added for testing the margin, but it doesn't help).

May a problem with already existing code...

But it's also working like this. But a little bit slower.

SoundSSGood:
> But it's also working like this. But a little bit slower.

If you are going to leave it as it is now, then add a comment in the code that optimization is possible.

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +686,283 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
+{
+	SDL_Surface *sourceSurface = SDL_ConvertSurfaceFormat(source, SDL_PIXELFORMAT_ARGB8888, 0);
+	SDL_Surface *destSurface = newSurface(Point(source->w, source->h));
+
+	assert(destSurface->format->format == SDL_PIXELFORMAT_ARGB8888);
+
+	double shearX = doSheer ? 0.5 : 0.0;
+	double scaleY = doSheer ? 0.5 : 0.25;
+
+	int lowestSource = getLowestNonTransparentY(sourceSurface);
```
IvanSavenko, line 968: Not 100% sure, but it should always be (height-1).

VCMI generally trims images on load (see `SDLImageOptimizer::optimizeSurface`).

In theory this can be bypassed by if images were packaged into .def without trimming, but apps like deftools usually trim image & add shadow anyway.

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +686,283 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
+
+	// Parameters for applyAffineTransform
+	double a = 1.0;
+	double b = shearX;
+	double c = 0.0;
+	double d = scaleY;
+	double tx = -shearX * lowestSource;
+	double ty = lowestSource - lowestTransformed;
+
+	applyAffineTransform(sourceSurface, destSurface, a, b, c, d, tx, ty);
```
IvanSavenko, line 979: I think this can be replaced with `SDL_BlitScaled`?

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +686,283 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
+			Uint32 pixel = pixels[i];
+			Uint8 pr;
+			Uint8 pg;
+			Uint8 pb;
+			Uint8 pa;
+			// Extract existing RGBA components using SDL_GetRGBA
+			SDL_GetRGBA(pixel, surface->format, &pr, &pg, &pb, &pa);
+
+			Uint32 newPixel = SDL_MapRGBA(surface->format, r, g, b, a);
+			if(pa == 0)
```
IvanSavenko: What if pixel is semi-transparent?

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +686,283 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
 	other = CSDL_Ext::fromSDL(rect);
 }
+
+SDL_Surface* CSDL_Ext::drawOutline(SDL_Surface* source, const SDL_Color& color, int thickness)
+{
+	if(thickness < 1)
+		return nullptr;
+
+	SDL_Surface* sourceSurface = SDL_ConvertSurfaceFormat(source, SDL_PIXELFORMAT_ARGB8888, 0);
+	SDL_Surface* destSurface = newSurface(Point(source->w, source->h));
```
IvanSavenko: Shouldn't destSurface be `thickness` pixels wider on each side?

This is *outline*, it should be around the image

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +686,283 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
 
 	other = CSDL_Ext::fromSDL(rect);
 }
+
+SDL_Surface* CSDL_Ext::drawOutline(SDL_Surface* source, const SDL_Color& color, int thickness)
+{
+	if(thickness < 1)
+		return nullptr;
+
+	SDL_Surface* sourceSurface = SDL_ConvertSurfaceFormat(source, SDL_PIXELFORMAT_ARGB8888, 0);
```
IvanSavenko: Why you need to create a copy? `SDL_GetRGBA` should be able to access data in any format

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +686,283 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
+	{
+		// Go through all transparent pixels and check if they border an opaque region
+		for(int y = r.begin(); y != r.end(); ++y)
+		{
+			for(int x = 0; x < width; x++)
+			{
+				if(getAlpha(x, y) != 0)
+					continue; // Skip opaque pixels
+
+				bool isOutline = false;
```
IvanSavenko: Perhaps instead of boolean yes / no we should locate most opaque pixel and set overlay opacity to `(most opaque nearby pixel alpha - this pixel alpha)`? 

Will probably look better if image has semi-transparency.

client/renderSDL/SDL_Extensions.cpp
```diff
@@ -685,3 +686,283 @@ void CSDL_Ext::getClipRect(SDL_Surface * src, Rect & other)
...
+			int sumR = 0;
+			int sumG = 0;
+			int sumB = 0;
+			int sumA = 0;
+			int count = 0;
+
+			for (int ky = -1; ky <= 1; ++ky)
+			{
+				int ny = std::clamp(y + ky, 0, height - 1);
+				for (int kx = -1; kx <= 1; ++kx)
```
IvanSavenko, line 935: Perhaps for upscaled images we should also increase blur? Similar to how outline has thickness equal to xbrz upscaling factor.

client/renderSDL/SDL_Extensions.cpp
Laserlicht: Yes. I know. But trimming is disabled here atm and is done after these algorithms.

If trimming is done before there we need to extend the surfaces or limiting trimming (see here: https://github.com/vcmi/vcmi/pull/5916/commits/2607767ab7da35258f2e0da0632ab2bdc612930c).

But it's not working correctly and searched some hours, why. Was not able to find the issue. Basically the issue is some "wobbeling" shadows. I think it comes from the different sizes of the animation images (some are bigger, some are smaller) after trimming. But was not able to find the exact problem.

Because of this I reverted back to full image where it works correctly.

client/renderSDL/SDL_Extensions.cpp
Laserlicht: Only the scaling/translating part. SDL2 is not capable of shearing images (needed for correct adventure map shadow).
Because of this I've used a affine transformation (which is capable of everything).

Laserlicht:
And @IvanSavenko any idea about this?

> Shadows and overlays are flickering the first time while loading. This is also the case before this PR by using -OVERLAY and - SHADOW. Seems a regression. Maybe https://github.com/vcmi/vcmi/pull/5306? Any idea for the reason?

Not directly related to this PR. But also a required step for a good replacement of PNG over DEF.

IvanSavenko:
Yeah, could be related to that PR. Feel free to open separate ticket.

client/renderSDL/SDL_Extensions.cpp
Laserlicht: Enough space as it is calculated on whole (unoptimized) image yet.

client/renderSDL/SDL_Extensions.cpp
Laserlicht: Asked ChatGPT:

`SDL_GetRGBA in SDL2 works with most standard pixel formats, but not all pixel formats are guaranteed to be compatible. It relies on the SDL_PixelFormat structure, which must describe the pixel format correctly for the function to extract RGBA components.`

IvanSavenko: Why you need to use Artificial Idiot when SDL has all its methods well-documented?
https://wiki.libsdl.org/SDL2/SDL_GetRGBA

Yes, in theory we can add custom pixel format that SDL_GetRGBA can't handle. But we don't do such thing. You're copying a massive image to account for scenario that will never happen

client/renderSDL/SDL_Extensions.cpp
Laserlicht: set threshold now to 128 alpha

client/renderSDL/SDL_Extensions.cpp
Laserlicht: > But we don't do such thing

Okay, good to know. Removed.

client/renderSDL/SDL_Extensions.cpp
Laserlicht: Done.

client/renderSDL/SDL_Extensions.cpp
Laserlicht: Not a bad idea. But not sure if it makes much visual difference. Blur is mainly to remove sharp edges from sprites. PNGs may already have smooth edges.

Needs some test, I think. May an improvement for future.

client/renderSDL/SDL_Extensions.cpp
IvanSavenko: Hmm... But can we even generate good enough shadow for map objects?

Can you check it on a few more examples? Our shadow generation algorithm basically assumes that sprite is sheet of paper - 0 depth, all parts are at the same distance from viewer.

This looks okay-ish for creatures, but for map objects this is often not true.

Sure, it will work for objects like Observation Tower. But will not work for large objects - typical forest / mountain obstacles, objects like Utopia. And will completely break for objects like water reefs that consist from several disconnected "patches". Might be better to require modders to provide shadows for those objects?

IvanSavenko: Or at the very least - can you try to use `SDL_BlitScaled` for creature shadows where such complexity is not needed? I suspect that this transformation is very slow, especially when compared to SDL blit routines.

Laserlicht: extended `applyAffineTransform` with an "speedup path".

PR #5909, opened 2025-07-09
Template editor - object & townHints selection

Laserlicht:
- add object selection to template editor
- add town hint selection to template editor
- add some missing translation strings

PR #5919, opened 2025-07-13
Support for banned game entities in random map templates

IvanSavenko:
The following entities can now be banned in a random map template definition:
- Hero
- Artifact
- Spell
- Secondary skill

The ban follows the same rules as banning via the map settings in the map editor.

It is also now possible to bypass dependencies and access identifiers from mods that are not dependencies when defining:
- Banned entities in random map templates
- the chance of a hero class appearing in a tavern of a specific faction
- the chance of a spell appearing in a mage guild of a specific faction
- the chance of a hero class receiving a secondary skill

For this to work, the identifier must be specified in full, e.g. `modName:objectName`. If the specified mod is not active, the game will silently ignore this entry.

This behaviour is not affected by mod load order - this format can be used to access a mod that has not yet been loaded.

### DIFF
--- a/client/lobby/RandomMapTab.cpp
+++ b/client/lobby/RandomMapTab.cpp
@@ -182,6 +182,9 @@ void RandomMapTab::updateMapInfoByHost()
 	mapInfo->mapHeader->name.appendLocalString(EMetaText::GENERAL_TXT, 740);
 	mapInfo->mapHeader->description.appendLocalString(EMetaText::GENERAL_TXT, 741);
 
+	if(mapGenOptions->getWaterContent() != EWaterContent::RANDOM)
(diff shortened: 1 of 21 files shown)


yche2990:
Can you extend this functionality to bypass dependency to more of the other items in rmg? For example, to include
Allowedmonsters, allowedtowns, bannedmonster, bannedtowns and custom objects?
To have RMG Template be confined with mod and dependency reduces the utility of such a template to be shared. And in fact if some of the items whose base mods are not active can be safely ignored will help to keep the template to still function in rmg.

IvanSavenko:
> Allowedmonsters, allowedtowns, bannedmonster, bannedtowns 

Those should actually work as well. Updated docs.

> and custom objects

Probably not this time

Laserlicht:
@IvanSavenko can you merge?
Prepared the changes for template Editor.

And maybe also #5909 to avoid conflicts.

Issue #5786, opened 2025-06-12
Multiple issues with magic mirror spell


Mich264:
Magic mirror spell have multiple bugs.

To Reproduce bugs:
Cast Magic Mirror spell
Trigger Magic Mirror

Bug 1 Magic Mirror don't defend affected unit against spells when trigger.
Expected behavior
After trigger magic mirror should defent unit agaists casted spell effects with 100% efficiency.
Actual behavior
When trigger magic mirror don't defend protected unit agaists casted spell effects.

Bug 2 Redirected spell have fraction of his spell power
Expected behavior
Redirected spell should attact with full efficiency.
Actual behavior
Redirected spell have laughable spell power.
When tested debuff with 47 turn duration spell was redirected with 3 turn duration.
Direct Damage spell with 1225 damage points was redirected with 10 points of damage.

Bug 3 Magic Mirror and Hypnotize
Expected behavior
In H3 Magic Mirror has feature that unit protected by Magic Mirror was immune to Hypnotize
Actual behavior
Magic Mirror don't give immune to Hypnotize

Bug 4 No information when redirected spell failed 
Expected behavior
In H3, when spell is redirected to unit which is immune to spell (example lightning against black dragon), is played spell resisted animation and battle log receive entry.
Actual behavior
Game don't give any informtion when spell was redirected to creature which is immune to spell.

Bug 5 Area of effect spells
Expected behavior
Area of effect spells should be redirected if unit protected by Magic Mirror is affected.
Actual behavior
Magic Mirror only redirect area of effect spells when center of spell is in a hex where unit protected by Magic Mirror stand.

Bug 6 Two Magic Mirror 
Actual behavior
Magic Mirror don't defent agaists spells redirected by another Magic Mirror
Expected behavior
Should be possible to rederict spell redirected by another Magic Mirror.
This should be disabled for mods when chance for redirecting spell is above 80% if all units during battle are protected by Magic Mirror.

Bug 7 Magic Mirror and Blind/Berserk Spell
Actual behavior
If blind is redirected to a unit that is currently moving then you can perform attack, wait and defend but you can't move without attacking.
If you attack another unit your attack cause 0 damage and this cause counterattack which remove blind.
If you wait or defend, blind will work properly in the next turn.
Similar issue is for berserk spell but in this case if you attack another unit berserk is automatically disabled.
If you perform another action berserk will trigger in the next turn.
Expected behavior
This spells should affect unit immediately.

OS: Win 11
[...]

PR #5920, opened 2025-07-14
Implement berserk behavior if no unit in melee range or unit can shoot.

Opuszek:
PR fixes following issues with Berserk spell:
- Units that can shoot (i. e. range units with ammunition) attack in melee.
- Melee units that cannot find a target within a single turn movement will pass their turn even if they are reachable units (they should move towards instead).

PS: I also renamed "getAttackableBattleHexes" because it is a misleading name and I lost time because of it.
PS2: If both this pr and 5910 got merged please do 5910 first, I think resolving conflicts will be easier this way.

### DIFF
--- a/server/battles/BattleFlowProcessor.cpp
+++ b/server/battles/BattleFlowProcessor.cpp
@@ -388,24 +388,36 @@ bool BattleFlowProcessor::tryActivateBerserkPenalty(const CBattleInfoCallback & 
 {
 	if (next->hasBonusOfType(BonusType::ATTACKS_NEAREST_CREATURE)) //while in berserk
 	{
-		logGlobal->trace("Handle Berserk effect");
-		std::pair<const battle::Unit *, BattleHex> attackInfo = battle.getNearestStack(next);
-		if (attackInfo.first != nullptr)
+		ForcedAction forcedAction = battle.getBerserkForcedAction(next);
+		if (forcedAction.type == EActionType::SHOOT)
 		{
-			BattleAction attack;
-			attack.actionType = EActionType::WALK_AND_ATTACK;
-			attack.side = next->unitSide();
-			attack.stackNumber = next->unitId();
-			attack.aimToHex(attackInfo.second);
-			attack.aimToUnit(attackInfo.first);
-
-			makeAutomaticAction(battle, next, attack);
-			logGlobal->trace("Attacked nearest target %s", attackInfo.first->getDescription());
+			BattleAction rangeAttack;
+			rangeAttack.actionType = EActionType::SHOOT;
+			rangeAttack.side = next->unitSide();
+			rangeAttack.stackNumber = next->unitId();
+			rangeAttack.aimToUnit(forcedAction.target);
+			makeAutomaticAction(battle, next, rangeAttack);
+		}
+		else if (forcedAction.type == EActionType::WALK_AND_ATTACK)
+		{
+			BattleAction meleeAttack;
+			meleeAttack.actionType = EActionType::WALK_AND_ATTACK;
+			meleeAttack.side = next->unitSide();
+			meleeAttack.stackNumber = next->unitId();
+			meleeAttack.aimToHex(forcedAction.position);
+			meleeAttack.aimToUnit(forcedAction.target);
+			makeAutomaticAction(battle, next, meleeAttack);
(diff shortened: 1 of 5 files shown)


dydzio0614:
i think shooters always melee with berserk in original h3, needs verification

Opuszek:
They perform a range attack unless there is a unit directly next to them, tested on Homm3 Complete GOG.

I will figure out what's wrong with the build, gimme a second.

Opuszek:
It seems I broke a mock of a unit test by adding a method to a class.
I will need to look into how to start unit tests later, for now I just removed those changes, they weren't important anyway.
Sorry, as I said, I am new to C++.

IvanSavenko:
Generally you'll need to set `ENABLE_TEST` CMake variable to `ON` to enable tests. 

This will allow you to compile tests alongside with solution & run them when needed.

For command line that would have been `-DENABLE_TEST=ON` switch. Not sure about Visual Studio though. Haven't used it with CMake much.

IvanSavenko:
Not related to this PR, but when referencing other issues/PR's, it is better to prepend `#` for auto-formatting: 5910 vs #5910

Using lists (e.g. `- #5910`) will also show issue title:
- #5910

And using keywords in PR description (e.g. "Fix" / "Fixes") will auto-close it on merge, .e.g:
- Fixes #5786

server/battles/BattleFlowProcessor.cpp
```diff
@@ -363,23 +364,93 @@ bool BattleFlowProcessor::tryMakeAutomaticAction(const CBattleInfoCallback & bat
...
-			attack.stackNumber = next->unitId();
-			attack.aimToHex(attackInfo.second);
-			attack.aimToUnit(attackInfo.first);
+			return u->isValidTarget(false) && u->unitId() != next->unitId();
+		});
+		auto cache = battle.getReachability(next);
 
-			makeAutomaticAction(battle, next, attack);
-			logGlobal->trace("Attacked nearest target %s", attackInfo.first->getDescription());
+		if (next->canShoot() && battle.battleAdjacentUnits(next).empty())
```
IvanSavenko: Shooting is blocked by both friendly and hostile units?

What if shooter can shoot while blocked - under Bow of Sharpshooter effect?

server/battles/BattleFlowProcessor.cpp
```diff
@@ -363,23 +364,93 @@ bool BattleFlowProcessor::tryMakeAutomaticAction(const CBattleInfoCallback & bat
 	if (next->hasBonusOfType(BonusType::ATTACKS_NEAREST_CREATURE)) //while in berserk
 	{
 		logGlobal->trace("Handle Berserk effect");
-		std::pair<const battle::Unit *, BattleHex> attackInfo = battle.getNearestStack(next);
```
IvanSavenko, line 392: `getNearestStack` is also used by AI for the same purpose - estimating Berserk. See `PotentialTargets::PotentialTargets` in BattleAI.

Perhaps `getNearestStack` should be renamed to something like `getBerserkTarget` and move all target selection logic into it?

Opuszek: I would expect BattleAI to run only after a new turn has been properly started. I wonder why this code in PotentialTargets is even needed. Will investigate.
As for renaming and refactoring, sounds like a good idea.

server/battles/BattleFlowProcessor.cpp
Opuszek: The stack is berserk, there are no friendly units. It will attack the closest, if the closest is right next to it, it will use a melee attack. This behavior is consistent with the original.
I forgot Bow of Sharpshooter even exists. Thanks.

server/battles/BattleFlowProcessor.cpp
Opuszek: Fixed.
There is a minor bug in PotentialTargets constructor: in case where there are multiple targets in equal distance from a berserk unit PT assumes that one specific stack will be targeted, though there is no guarantee that it will be so (neither there should be).
As the problem was existing before I consider it outside of the scope of pr.

server/battles/BattleFlowProcessor.cpp
Opuszek: Fixed.

Issue #5876, opened 2025-07-04
campaign: the hero was attacked in the sanctuary


GrafDeVan:
hero Clancy stands in the sanctuary, but on the next turn he is attacked (Straker)

<img width="858" height="375" alt="Image" src="https://github.com/user-attachments/assets/9bba22ab-3c48-42cf-b477-f30b26bccf0b" />

save  [gg2.zip](https://github.com/user-attachments/files/21064021/gg2.zip)

msvs, current develop

GrafDeVan:
The second case is even funnier. I lose Clancy in the sanctuary again, but the game lets me continue!

[ClansyLost!!.zip](https://github.com/user-attachments/files/21068533/ClansyLost.zip)

IvanSavenko:
Fixed in #5931

PR #5931, opened 2025-07-15
fixes sanctuary behavior

Opuszek:
Resolved problem:

Heroes inside a sanctuary can be attacked.

Solves #5876.

### DIFF
--- a/lib/pathfinder/PathfinderUtil.h
+++ b/lib/pathfinder/PathfinderUtil.h
@@ -33,12 +33,14 @@ namespace PathfinderUtil
 		case ELayer::SAIL:
 			if(tinfo.visitable())
 			{
-				auto frontVisitable = gameInfo.getObjInstance(tinfo.visitableObjects.front());
-				auto backVisitable = gameInfo.getObjInstance(tinfo.visitableObjects.front());
-
-				if(frontVisitable->ID == Obj::SANCTUARY && backVisitable->ID == Obj::HERO && backVisitable->getOwner() != player) //non-owned hero stands on Sanctuary
+				if (tinfo.visitableObjects.size() > 1)
 				{
-					return EPathAccessibility::BLOCKED;
+					auto frontVisitable = gameInfo.getObjInstance(tinfo.visitableObjects[0]);
+					auto backVisitable = gameInfo.getObjInstance(tinfo.visitableObjects[1]);
+					if(frontVisitable->ID == Obj::SANCTUARY && backVisitable->ID == Obj::HERO && backVisitable->getOwner() != player)
+					{
+						return EPathAccessibility::BLOCKED;
+					}
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


GrafDeVan:
I don't know what happened, but tried 2 saves and game now crashes. always
msvs 64
 
[VCMI_Client_logClancyLost.txt](https://github.com/user-attachments/files/21230150/VCMI_Client_logClancyLost.txt)
[VCMI_Client_logGG2.txt](https://github.com/user-attachments/files/21230151/VCMI_Client_logGG2.txt)

Opuszek:
Hmm... Maybe it is something predefined, like the procedure "attack mf on this hex" is already decided on and cached within the save, so when it is not possible the game panics and crashes? I tested changes with AI and AI did realize the heroes is not reachable.
I should have checked it on that save.
Will investigate.

Opuszek:
I was able to load saves running vcmi from QT Creator and they worked as expected (i.e AI ignored the hero hiding in a sanctuary).
Strangely enough, when I launch the application directly from the client in "out" directory loading both of those saves failed with vcmi displaying a pop-up about a failed iostream read. I clean and rebuild the project and the problem disappeared.
I tried both campaign AIs, both work.
So maybe it simply needs to be cleaned and rebuilt for reasons.
Other then that please double check used mods, maybe something went wrong there.
If that won't help we will go from there, though if the problem is environmental I doubt I will be able to help.

GrafDeVan:
> Other then that please double check used mods, maybe something went wrong there.

Of the active mods I only have vcmi extras

lib/pathfinder/PathfinderUtil.h
```diff
@@ -33,12 +33,14 @@ namespace PathfinderUtil
...
 			{
-				auto frontVisitable = gameInfo.getObjInstance(tinfo.visitableObjects.front());
-				auto backVisitable = gameInfo.getObjInstance(tinfo.visitableObjects.front());
-
-				if(frontVisitable->ID == Obj::SANCTUARY && backVisitable->ID == Obj::HERO && backVisitable->getOwner() != player) //non-owned hero stands on Sanctuary
+				if (tinfo.visitableObjects.size() > 1)
 				{
-					return EPathAccessibility::BLOCKED;
+					auto frontVisitable = gameInfo.getObjInstance(tinfo.visitableObjects[0]);
+					auto backVisitable = gameInfo.getObjInstance(tinfo.visitableObjects[1]);
```
IvanSavenko: Should not matter in this scenario, but might be better to access front/back (first/last) elements instead. 

Extreme edge case here would be modded Sanctuary that can be placed on water. In this scenario, visitable objects would be (sanctuary, boat, hero) - hero is no longer at index 1. 

```suggestion
					auto frontVisitable = gameInfo.getObjInstance(tinfo.visitableObjects.front());
					auto backVisitable = gameInfo.getObjInstance(tinfo.visitableObjects.back());
```

PR #5932, opened 2025-07-15
Replace NSIS with Inno Setup

GeorgeK1ng:
1) Updated Inno Setup ISS and CMD scripts
      - added ARM64 support
      - support for **/silent /launch** switches for future auto updating from Launcher
      
2) Updated ISL to latest ones
      - converted to UTF8
      - added new strings for Inno Setup 6.5
      
3) Added Windows ARM64 configuration and build

4) Updated Workflow to replace old NSIS installer with new Inno Setup installer
      - To make this alive, new Windows Depencies needs to be added see https://github.com/vcmi/vcmi-deps-windows/issues/6


GeorgeK1ng:
Current commit should produce correct artifacts. When everything builded I will change install_vcpkg_dependencies.sh repo id back to VCMI as new files doesn't exist in windows depencies release which will lead to fail aka red cross.

GeorgeK1ng:
Disabled Windows ARM64 build for now due reported Launcher crash in libcrypto-3-arm64.dll

`Faulting application name: VCMI_launcher.exe, Version: 0.0.0.0, Timestamp: 0x6877f51a  
Faulting module name: libcrypto-3-arm64.dll, Version: 3.5.0.0, Timestamp: 0x682ae0fc  
Exception code: 0xc0000005  
Fault offset: 0x000000000015a4c8  
Faulting process ID: 0x209C  
Faulting application start time: 0x1DBF698705C3790  
Faulting application path: C:\Program Files (x86)\VCMI (PR 5932)\VCMI_launcher.exe  
Faulting module path: C:\Program Files (x86)\VCMI (PR 5932)\libcrypto-3-arm64.dll  
Report ID: addf4027-d156-4dbf-be8d-d56f72581b8c  
Full name of the faulting package:  
Application ID related to the faulting package:  
`

Current commit will end in fail due missing ucrtRedist in Windows depencies repo.

EDIT: When added ucrtRedist packages to Windows depencies repo release, it can be probably merged, all naming logic are same as before. This should only replace NSIS with Inno Setup installer without any side effects. NSIS code removal can be done in separate PR later.

kambala-decapitator:
please don't concentrate on vcpkg specifics too much as vcpkg will soon be "sunset" for VCMI

GeorgeK1ng:
> please don't concentrate on vcpkg specifics too much as vcpkg will soon be "sunset" for VCMI

When times come, just one parameter needs to be omitted in installer building.

.github/workflows/github.yml
```diff
@@ -282,6 +295,43 @@ jobs:
...
         pgrep XProtect || true
 
+    - name: Install Inno Setup 6+
+      if: ${{ startsWith(matrix.platform, 'msvc') }}
+      run: |
+        $defaultPath = "$env:ProgramFiles`(x86`)\Inno Setup 6\ISCC.exe"
+        if (!(Test-Path $defaultPath)) {
+          Write-Host "Inno Setup 6 not found. Downloading and installing..."
+          $installArgs = @("/VERYSILENT", "/NORESTART")
+          Invoke-WebRequest -Uri "https://files.innosetup.nl/innosetup-6.4.3.exe" -OutFile "isetup.exe"
```
kambala-decapitator: IS is already installed on the runner: https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md

.github/workflows/github.yml
```diff
@@ -282,6 +295,43 @@ jobs:
...
+        } else {
+          Write-Host "Found Inno Setup 6 at: $defaultPath"
+        }
+        $innoBin = Split-Path $defaultPath
+        Add-Content -Path $env:GITHUB_PATH -Value $innoBin
+      shell: pwsh
+
+    - name: Build VCMI Installer
+      if: ${{ startsWith(matrix.platform, 'msvc') }}
+      run: >
```
kambala-decapitator, line 308: why not using `|` like in other places?

.github/workflows/github.yml
```diff
@@ -385,6 +435,42 @@ jobs:
...
+          $minorMatch = [regex]::Match($content, 'set\(VCMI_VERSION_MINOR\s+(\d+)\)')
+          $patchMatch = [regex]::Match($content, 'set\(VCMI_VERSION_PATCH\s+(\d+)\)')
+
+          if (!($majorMatch.Success -and $minorMatch.Success -and $patchMatch.Success)) {
+            Write-Host "Failed to extract version numbers"
+            exit 1
+          }
+
+          $shortVersion = "$($majorMatch.Groups[1].Value).$($minorMatch.Groups[1].Value).$($patchMatch.Groups[1].Value)"
+          $timestamp = Get-Date -Format "yyyyMMddHHmmss"
```
kambala-decapitator: simply using `grep` in bash seems like an easier alternative. Besides, I don't think we have PS speakers in the project, I wouldn't use it unless it greatly simplifies things.

.github/workflows/github.yml
```diff
@@ -385,6 +435,42 @@ jobs:
             name: ${{ env.VCMI_PACKAGE_FILE_NAME }}
             path: |
               ./release.tar.gz
+  
+  extract-version:
+    name: Extract Version
+    runs-on: ubuntu-24.04
```
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/github.yml
GeorgeK1ng: Yes I know, but I was never able to locate it in standard install path. So installing it solved it. If you know it's installed location, can be linked to that location.

.github/workflows/github.yml
GeorgeK1ng: When used | parameters to CMD were not passed correctly.

.github/workflows/github.yml
kambala-decapitator: was trivial to find out: https://github.com/kambala-decapitator/workflow-test/actions/runs/16346361412/workflow and https://github.com/kambala-decapitator/workflow-test/actions/runs/16346361412/job/46180999206

GeorgeK1ng: Hmm, I know why I have not found it I tested it in windows 11 arm runner. And it's in different than standard location. Will updated CMD script with workaround for this. CMD still will require some touch to make single click build outside Github and I will do that later when I can try it myself.

.github/workflows/github.yml
GeorgeK1ng: Replaced with bash equivalent

.github/workflows/github.yml
GeorgeK1ng: Done

GeorgeK1ng:
Simplified workflow as suggested. When ucrtRedist added to latest[ Windows Depencies release repo](https://github.com/vcmi/vcmi-deps-windows/releases/tag/v1.8), then will be everything compiled successfully.

CI/install_vcpkg_dependencies.sh
```diff
@@ -1,7 +1,17 @@
 #!/usr/bin/env bash
 
-RELEASE_TAG="v1.8"
-FILENAME="dependencies-$1"
-DOWNLOAD_URL="https://github.com/vcmi/vcmi-deps-windows/releases/download/$RELEASE_TAG/$FILENAME.txz"
+REPO="vcmi"
 
-curl -L "$DOWNLOAD_URL" | tar -xf - --xz
+# Fetch latest release tag from GitHub API
+RELEASE_TAG=$(curl -s "https://api.github.com/repos/$REPO/vcmi-deps-windows/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
```
IvanSavenko: This may break builds in future.

For example, let's assume that we removed library X, and created new release without that library.

When this happens, all CI in vcmi would fail (other than PR with removal of said library) - both daily builds and all PR's.

With current approach CI always points to *correct* release.

GeorgeK1ng: I have commented it out and added static v1.8 for now.

.github/workflows/github.yml
```diff
@@ -72,22 +73,33 @@ jobs:
             conan_prebuilts: dependencies-ios
             conan_options: --options with_apple_system_libs=True
           - platform: msvc-x64
+            arch: x64
             os: windows-latest
```
IvanSavenko, line 77: Potential problem in future: right now `windows-latest` points to Win-2022 image. But Win-2025 image comes without innosetup preinstalled.

Meaning, CI might break in future if Github were to switch `latest` image to 2025 or when 2022 image will be deprecated.

GeorgeK1ng: I have removed download and install Inno Setup step as @kambala-decapitator suggested. Can be added in future. I quess in 2025 runner will be Inno Setup preinstalled when right time commes. Otherwise, we can re-add that step now to be sure.

GeorgeK1ng: Maybe add just for sure Inno Setup step back?

    - name: Ensure Inno Setup is installed
      shell: bash
      run: |
        if [ ! -f "/c/Program Files (x86)/Inno Setup 6/ISCC.exe" ] && [ ! -f "/c/ProgramData/Chocolatey/bin/ISCC.exe" ]; then
          choco install innosetup --no-progress -y
        fi

IvanSavenko: Not against it. And we might also switch to Win-2025 runner explicitly. There is chance that when Github migrates `latest` to Win-2025 we'll get unexpected CI failures.

Laserlicht:
Related to SSL error on ARM64: https://github.com/actions/runner-images/issues/14073

GeorgeK1ng:
Until it's fixed in ARM64 Runner, we should simply have workaround for ARM64 users, after installation unpack these libcrypto-3-arm64.dll + libssl-3-arm64.dll into VCMI installation directory. Yes, it's workaround, but better than error reports. Or if same bug still exist when 1.7 release time come, we can manually repack AMR64 installer with these DLLs.

[ARM64_SSL_TemporaryFix.zip](https://github.com/user-attachments/files/21333448/AMR64_SSL_TemporaryFix.zip)

We should also add note for ZIP release users - VCRedist are required to run VCMI. But new installer can be run without administrator privilegies and supports setup "For me only", so anyone can use installer, even for creating "portable" VCMI.

.github/workflows/github.yml
```diff
@@ -72,22 +73,33 @@ jobs:
...
             test: 0
-            pack: 1
+            pack: 0
```
IvanSavenko, line 79: Might be better to keep packing, but instead switch to .zip format (should be supported by CPack).

Probably no need to upload such .zip's to vcmi.eu, but would be convenient to have in CI artifacts

CI/install_vcpkg_dependencies.sh
```diff
@@ -1,7 +1,19 @@
 #!/usr/bin/env bash
 
-RELEASE_TAG="v1.8"
-FILENAME="dependencies-$1"
-DOWNLOAD_URL="https://github.com/vcmi/vcmi-deps-windows/releases/download/$RELEASE_TAG/$FILENAME.txz"
+REPO="vcmi"
```
IvanSavenko: This is not a "repo" but "organization" (or account in general)

.github/workflows/github.yml
GeorgeK1ng: I have disabled it as currently it creates old NSIS one installer. Have not removed NSIS yet. Keeping this for future cleanup PR.
Adding ZIP release could be easy step in CI with same file ignore filters as are in ISS script. No idea about CPack ZIP.

CI/install_vcpkg_dependencies.sh
GeorgeK1ng: Renamed REPO -> ACCOUNT

GeorgeK1ng:
> Code-wise looks OK in general, will take a look at how new installer looks & works.

Hope it passes all your scenarios. Have not found anything wrong during various tests under Windows 7 and Windows 10.
For future improvement could be added optional default application path change wizard. But only when it will be allowed directly from launcher. So really in future.

IvanSavenko:
- Language is asked only after install for me / install for all. Should probably be the first option.

- On start it ask to pick install for me / install for all. After that I select language, press next. And it asks me again whether install for me or for all?

зображення
зображення

- First 2 transitions (install for me/all -> language select -> main install) don't remember window position / display. I dragged window to 2nd monitor, but it keeps resetting to 1st one. Might also be Wine bug though. (very minor bug obviously, feel free to ignore unless there is easy & clean fix)

- perhaps shortcuts creation & file association should be disabled by default for PR builds, and only enabled for release builds?

- I kept "copy H3 data automatically" option on (with no H3 installed in Wine)... and nothing happened? Checkbox should probably be disabled in my case, or at least there should be warning that H3 data was not found

GeorgeK1ng:
First dialog is there for UAC. When you hit first option installer is launched with admin rights. Then is there language prompt. This is currently Inno Setup behavior. There is no chance to add language selector first, they are built in features. Maybe fork and mod Inno Setup code - have not tried it.

We can modify text on UAC like prompt to make it more clear, because it's way how is installer launched. Will modify all languages files. Have nothing to do with installation mode. But if you run installer without admin right, you will always have there only one option - "Install for me only". If you run installer as admin, you can still choose mode "Install for me only" to install VCMI into profile instead of Program Files.

Shortucts and associations needs to be tested, if can be automatically unchecked on output installer variable name.
Even you install PR version

I don't see H3 checkbox when I dont' have installed H3
image

When I have installed H3 and VCMI is newly installed and H3 files are not copied checkbox is here.
image

Tested on fresh Windows 7 SP1 in Virtualbox.

IvanSavenko:
> We can modify text on UAC like prompt to make it more clear, because it's way how is installer launched

How it looks natively on Windows? If it shows that "uac prompt shield" icon then OK as it is for me.

> Shortucts and associations needs to be tested, if can be automatically unchecked on output installer variable name.

Would be nice to have. PR's are usually installed temporarily, so no need to overwrite paths to permanent / release installation

> I don't see H3 checkbox when I dont' have installed H3

Ah, apologies, false alarm. Looks like I've launched H3 from the same Wine environment, and it automatically created registry entries that installer checks for.

And noticed another nice-to-have: use language selected in Installer for Launcher & auto-skip language selection in Launcher (this part would require changes to Launcher)

GeorgeK1ng:
It should be now clear, I have modded required text for initial dialog. Inno Setup have it as some kind of prelauncher, so when selected "Admin setup" then it's launched with admin rights and they are requested when needed.

For language for launcher I have already tried it. It's doable, JSON with selected language can be created by installer and launcher is then launched in same language as installer. But it will require additional code for ISS. Will rather do this in later PR same as for Task page skipping for PR builds.

GeorgeK1ng:
Just noticed ARM64 runner should be updated and rebuilded dependencies should have now fixed mentioned SSL issue. So full ARM64 for Windows should be there.

.github/workflows/github.yml
IvanSavenko: Not insisting on particular method. In theory .zip via CPack should be as easy as switching CPack generator to zip.

I need to create .zip in one way or another for releases. Old installer could be extracted with 7z in command line, but not new installer.

Sure, I can always install build locally and repack it again into zip, but want to avoid extra steps if possible.

Laserlicht: > Old installer could be extracted with 7z in command line, but not new installer.

Should be possible to extract with innoextract command line tool.

IvanSavenko:
So, todo's for future PR(s):
- auto-select language for Launcher based on installer language
- do not create shortcuts / associations for PR builds
- add .zip as alternative format in place of windows installer

Before merging this PR will also reconfigure script on download.vcmi.eu to use following single-directory structure for builds:
```
https://download.vcmi.eu/branch/[branch name]/[os]-[platform]/
```
where `[os]-[platform]` would be:
```
windows-x64
windows-x86 (not present right now, but can be added)
windows-arm64 (new)
ios (no platform suffix)
android-arm64-v8a
android-armeabi-v7a
android-x64 (in future)
macos-intel
macos-arm

```

.github/workflows/github.yml
kambala-decapitator: > should be supported by CPack

Yes, this would be the preferred way

GeorgeK1ng:
Yes thats needed structure, when its prepared on VCMI.eu we can also add auto update soon. 

About ZIP and CPack, this Is something I have no idea how to do it.

Uncheck and skip Task page in PR builds will be next touch from my side.

kambala-decapitator:
zip & cpack: it should be as easy as either `cpack -G ZIP` or setting cpack generator in CML explicitly (like we do for other platforms) and just `cpack`

PR #5937, opened 2025-07-18
Add key binding menu

Laserlicht:
Shortcuts can now viewed ingame.

[Peek 2025-07-18 23-42.webm](https://github.com/user-attachments/assets/f389bddf-c2e4-43e7-be35-b18b69fed908)

### DIFF
--- a/Mods/vcmi/Content/config/german.json
+++ b/Mods/vcmi/Content/config/german.json
@@ -264,7 +264,7 @@
 	"vcmi.keyBindings.editButton.help" : "Tastenbelegung bearbeiten",
 	"vcmi.keyBindings.input" : "Tastenbelegung für {%s} ändern.\n\nBitte eine Taste oder Tastenkombination eingeben. Außerhalb klicken zum Abbrechen.",
 	"vcmi.keyBindings.inputSet" : "Tastenbelegung für {%s} wird zu {%s} geändert.\n\nZu den existierenten hinzufügen? Ansonsten wird ersetzt.",
-	"vcmi.keyBindings.popup" : "Für {%s} sind die folgenden Tastenbelegung konfiguriert:\n\n",
+	"vcmi.keyBindings.popup" : "Für {%s} ist die folgende Tastenbelegung konfiguriert:\n\n",
 	"vcmi.keyBindings.reset" : "Zurücks.",
 	"vcmi.keyBindings.reset.help" : "{Zurücksetzen}\n\nSetzt Tastenbelegungen auf Standardeinstellungen zurück",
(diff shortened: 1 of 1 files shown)


Laserlicht:
It's now possible to set keys:

[Peek 2025-07-19 16-08.webm](https://github.com/user-attachments/assets/5bb3c20d-8fc2-4146-adfd-7185786e5b8b)

Mods/vcmi/Content/config/english.json
IvanSavenko, line 1: +250 new strings...

We'll need to find an easier way to update localizations. Locating missing 1-2 strings is already inconvenient, with this PR this will become even worse.

(yeah, outside of this PR scope, but something to think about in future

GeorgeK1ng: Good is, translation files are easy to compare in Beyond compare for example, so it's not hard to track missing strings for example in English.json vs Czech.json but we can add some worklow to automatically fullfill missing strings into language files as comments into required places to see what require translation.

Laserlicht: Yeah,

I've a small script for me updating all missing strings into german. To track them.

I think the biggest problem are updated strings. Currently the best way is to delete every translation after updating english string. But not sure if every PR creator will do this...

IvanSavenko: There is actually `/translate missing` command that should export all missing strings.

Might be an interesting idea to set up workflow that would automatically export new strings to all language json's  on schedule (e.g. weekly) if somebody is willing to investigate it. We can also add `lupdate` run into it to regenerate Qt translations.

GeorgeK1ng: Will check it. I have already investigated it for mods translations export on PR to show real mods translations status.

Laserlicht: > We can also add lupdate run into it to regenerate Qt translations.

Good idea. I think actions should create PR (maximal once per week). To avoid conflicts with pending translation PR's when committing directly.

ToRRent1812:
Wouldn't be better to have key binding menu in general options? You can set hotkeys for both, adventure map and battle in there.

Laserlicht:
> Wouldn't be better to have key binding menu in general options? You can set hotkeys for both, adventure map and battle in there.

Yes. But no space :D

ToRRent1812:
Fullscreen preference could be a single button that opens new window like interface scaling/resolution change, then You get free space for shortcuts/input category

IvanSavenko:
Another option would be to switch to vertical tabs - move General/Adventure/Battle to left and add keybindings as 4th tab, instead of having it as separate window

PR #5940, opened 2025-07-20
case insensitive cheat identifier

Laserlicht:
Identifier in cheats are case insensitive now.

### DIFF
--- a/lib/modding/IdentifierStorage.cpp
+++ b/lib/modding/IdentifierStorage.cpp
@@ -157,22 +157,22 @@ CIdentifierStorage::ObjectCallback CIdentifierStorage::ObjectCallback::fromNameA
 
 void CIdentifierStorage::requestIdentifier(const std::string & scope, const std::string & type, const std::string & name, const std::function<void(si32)> & callback) const
 {
-	requestIdentifier(ObjectCallback::fromNameAndType(scope, type, name, callback, false, false));
+	requestIdentifier(ObjectCallback::fromNameAndType(scope, type, name, callback, false, false, true));
 }
 
 void CIdentifierStorage::requestIdentifier(const std::string & scope, const std::string & fullName, const std::function<void(si32)> & callback) const
 {
-	requestIdentifier(ObjectCallback::fromNameWithType(scope, fullName, callback, false));
+	requestIdentifier(ObjectCallback::fromNameWithType(scope, fullName, callback, false, true));
 }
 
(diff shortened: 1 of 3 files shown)


lib/modding/IdentifierStorage.cpp
```diff
@@ -343,6 +345,34 @@ void CIdentifierStorage::registerObject(const std::string & scope, const std::st
 	}
 }
 
+template <typename MultiMap>
+std::pair<typename MultiMap::const_iterator, typename MultiMap::const_iterator>
+caseInsensitiveEqualRange(const MultiMap& mmap, const std::string& key)
```
IvanSavenko: This won't work as expected if there are multiple items, since they can be stored non-continiously

Unlikely scenario, sure, but might happen eventually and would be hard to track down.

This method is also extremely slow since you need full lookup and that map is massive (thousands of elements, likely tens of thousands with mods).

Right now it is only used in this one scenario, but might change in future. Perhaps in addition to registeredObjects we can store separate map to lookup original case? E.g. map with `vampirelord -> vampireLord` pairs.

lib/modding/IdentifierStorage.h
```diff
@@ -97,10 +98,10 @@ class DLL_LINKAGE CIdentifierStorage
...
 
 	/// get identifier immediately. If identifier is not know and not silent call will result in error message
-	std::optional<si32> getIdentifier(const std::string & scope, const std::string & type, const std::string & name, bool silent = false) const;
-	std::optional<si32> getIdentifier(const std::string & type, const JsonNode & name, bool silent = false) const;
-	std::optional<si32> getIdentifier(const JsonNode & name, bool silent = false) const;
-	std::optional<si32> getIdentifier(const std::string & scope, const std::string & fullName, bool silent = false) const;
+	std::optional<si32> getIdentifier(const std::string & scope, const std::string & type, const std::string & name, bool silent = false, bool caseSensitive = true) const;
+	std::optional<si32> getIdentifier(const std::string & type, const JsonNode & name, bool silent = false, bool caseSensitive = true) const;
+	std::optional<si32> getIdentifier(const JsonNode & name, bool silent = false, bool caseSensitive = true) const;
+	std::optional<si32> getIdentifier(const std::string & scope, const std::string & fullName, bool silent = false, bool caseSensitive = true) const;
```
IvanSavenko: Might be better to add separate method, `getIdentifierCaseInsensitive`. Default parameters, especially bool's have problems with readability and with maintenance.

PR #5941, opened 2025-07-20
Fix regressions

IvanSavenko:
- Add workaround & logging for mods that use incorrect building ID's (e.g. Refugee Town mod)
- Fix crash on using JsonDeserializer on JsonNode without valid mod scope (e.g. on casting some spells in combat)
- Initialize mod scope of loaded rmg template in template editor


yche2990:
Will this fix the bug which I reported on the discord that the game crashed before entering a battle due to some creature contained in the mod which is not recognisable  by the game?

IvanSavenko:
Yep

PR #5942, opened 2025-07-20
Show skill that can be learned in university on right click

IvanSavenko:
As in title. 

Now if player has already visited university before, game will show list of skills that can be learned in this university on right click


GeorgeK1ng:
Maybe adding additional text to that popup window? Something like "_You may learn the following secondary skills at this university_" instead just "University".

IvanSavenko:
Actually there is already string for that:
```json
"core.xtrainfo.24" : "(Learn secondary skills for 2000 gold)",
```

Only used in editor in H3, but we can use it in game as well. Will do.

GeorgeK1ng:
What if we change skill price by mod? Not sure if we already can change that.

misiokles:
Modder should upgrade description in his mod too, like in hota balance submod.

GeorgeK1ng:
Yes, but this is tooltip for already visited object.

Issue #5943, opened 2025-07-21
Crash on AI turn after  Behavior StayAtTownBehavior


Macron1Robot:
**Describe the bug**
On AI turn game begins to crash after some turns.

**Game logs**
[2025-Jul-21 23:28:57.646235] DEBUG [AIGateway::makeTurn] ai - Player 3 will build Mage Guild Level 1 in town of Marshwall at (251 247 0)
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - AI state updated in 0 ms
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Checking behavior Recruit hero
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Behavior Recruit hero. Time taken 0
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Checking behavior Buy army
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Behavior Buy army. Time taken 0
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Checking behavior Build
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Behavior Build. Time taken 0
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Checking behavior Capture objects
[2025-Jul-21 23:28:57.675666] DEBUG [AIGateway::makeTurn] ai - Scanning objects, count 281
[2025-Jul-21 23:28:57.693290] DEBUG [AIGateway::makeTurn] ai - Scanning objects, count 187
[2025-Jul-21 23:28:57.716168] DEBUG [AIGateway::makeTurn] ai - Behavior Capture objects. Time taken 39
[2025-Jul-21 23:28:57.716168] DEBUG [AIGateway::makeTurn] ai - Checking behavior Unlock Clusters
[2025-Jul-21 23:28:57.722325] DEBUG [AIGateway::makeTurn] ai - Behavior Unlock Clusters. Time taken 6
[2025-Jul-21 23:28:57.722325] DEBUG [AIGateway::makeTurn] ai - Checking behavior Defend towns
[2025-Jul-21 23:28:57.723416] DEBUG [AIGateway::makeTurn] ai - Behavior Defend towns. Time taken 0
[...]

PR #5948, opened 2025-07-24
Fixes segmentation fault when getting magic strength of a hero

Opuszek:
Fixes issue #5943.
The problem was caused by dereferencing nullptr when getting adventure mechanics of a battle spell.


GrafDeVan:
my save now works as intended!
Now I can continue passing my campaign. Tnx!

PR #5949, opened 2025-07-25
Map Editor:Add filter to the herospell widget

Opuszek:
An improvement proposition.
I noticed that adding/removing spells is a little bit of a chore, so I've sorted the list and added a filter that hides all spells that do not contain the typed phrase (it also hides entire levels if they are empty).

### DIFF
--- a/mapeditor/inspector/herospellwidget.cpp
+++ b/mapeditor/inspector/herospellwidget.cpp
@@ -103,6 +103,70 @@ void HeroSpellWidget::on_customizeSpells_toggled(bool checked)
 	initSpellLists();
 }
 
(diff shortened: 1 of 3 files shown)


dydzio0614:
was this pull request closed by mistake?

Opuszek:
@dydzio0614 No, I've noticed that editor crashes when I uncheck "customize spells".
The same behavior happens on develop branch so **maybe** it is not related, but I want to make sure before I  request a merge.

kambala-decapitator:
You could mark the PR as draft

Opuszek:
It was unrelated. A bug that causes a crash is fixed in pr #5951.

Issue #5681, opened 2025-05-03
maps not importing


kambala-decapitator:
**Describe the bug**
Importing maps from launcher doesn't seem to work. Besides, after import attempt file name on the filesystem becomes "mangled" (see 2nd log entry).

In another case (user on discord) file doesn't have a readable name (e.g. `msf%3A1000007693`) and "domain" is `com.android.providers.downloads.documents`, while for me it's proper relative path and "domain" is `com.android.externalstorage.documents`. This difference may be related to the fact that I installed apk from releases and the discord user - from Google Play.

**Game logs**
```
// from Documents directory
Importing file content://com.android.externalstorage.documents/document/primary%3ADocuments%2Fgreece.h3m
Installing maps: started
Importing map 'content://com.android.externalstorage.documents/document/primary%3ADocuments%2Fgreece.h3m'
Installing maps: ended

// mangled from Download directory
Importing file content://com.android.externalstorage.documents/document/primary%3ADownload%2F_storage_emulated_0_Android_data_is.xyz.vcmi_files_vcmi-data_Maps_primary%253ADownload%252FInfested.h3m
Installing maps: started
Importing map 'content://com.android.externalstorage.documents/document/primary%3ADownload%2F_storage_emulated_0_Android_data_is.xyz.vcmi_files_vcmi-data_Maps_primary%253ADownload%252FInfested.h3m'
Installing maps: ended

// from Download directory
Importing file content://com.android.providers.downloads.documents/document/msf%3A1000007693
Installing maps: started
Importing map 'content://com.android.providers.downloads.documents/document/msf%3A1000007693'
Installing maps: ended
Importing file content://com.android.providers.downloads.documents/document/msf%3A1000007694
Installing maps: started
Importing map 'content://com.android.providers.downloads.documents/document/msf%3A1000007694'
Installing maps: ended
```

**To Reproduce**
1. open launcher
2. press Import files in the Game tab
3. select a h3m file

**Expected behavior**
selected maps appear in the game

**Actual behavior**
maps don't appear in the game, no error messages in the launcher or log

**Did it work earlier?**
never tried before

**Version**

- OS: Android 14, Samsung device (phone on my side, a discord user faced the same with a tablet)
- Version: 1.6.8

**Additional context**
files (including maps) import fine performing the initial data import from a directory

related: #5193

kambala-decapitator:
same issue on android 12 (I have Xiaomi)

kambala-decapitator:
[...]

PR #5959, opened 2025-07-28
Fix mod selection after installing + fix maps importing

GeorgeK1ng:
1) Fixed tree view selection, no more selection reset after installation
2) Installing phase now display correctly mod name instead of mod Id, same as in downloading process
3) Improved maps importing
    - Added support for importing ZIPs with maps without mod.json
    - Added import messages
    - Fixed maps imports on Android - Fixes: https://github.com/vcmi/vcmi/issues/5681

### DIFF
--- a/launcher/modManager/cmodlistview_moc.cpp
+++ b/launcher/modManager/cmodlistview_moc.cpp
@@ -883,27 +883,30 @@ void CModListView::installFiles(QStringList files)
 			ZipArchive archive(qstringToPath(realFilename));
 			auto fileList = archive.listFiles();
 
-			bool hasRootMaps = false;
+			bool hasModJson = false;
+			bool hasMaps = false;
 
 			for (const auto& file : fileList)
 			{
-				QString lower = QString::fromStdString(file);
+				QString lower = QString::fromStdString(file).toLower();
 
-				// Skip subfolders
-				if (lower.contains("/") || lower.contains("\\"))
-					continue;
-
-				if (lower.endsWith(".h3m", Qt::CaseInsensitive) || lower.endsWith(".h3c", Qt::CaseInsensitive) || lower.endsWith(".vmap", Qt::CaseInsensitive) || lower.endsWith(".vcmp", Qt::CaseInsensitive))
+				// Check for mod.json anywhere in archive
+				if (lower.endsWith("mod.json"))
 				{
-					hasRootMaps = true;
+					hasModJson = true;
 					break;
 				}
+
+				// Check for map files anywhere
+				if (lower.endsWith(".h3m") || lower.endsWith(".h3c") || lower.endsWith(".vmap") || lower.endsWith(".vcmp"))
+					hasMaps = true;
 			}
 
-			if (hasRootMaps)
-				maps.push_back(filename);
-			else
+			if (hasModJson)
 				mods.push_back(filename);
+
+			if (hasMaps)
+				maps.push_back(filename);
 		}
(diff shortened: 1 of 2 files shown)


launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -870,7 +879,35 @@ void CModListView::installFiles(QStringList files)
...
+				// Check for mod.json anywhere in archive
+				if (lower.endsWith("mod.json"))
+				{
+					hasModJson = true;
+					break;
+				}
+
+				// Check for map files anywhere
+				if (lower.endsWith(".h3m") || lower.endsWith(".h3c") || lower.endsWith(".vmap") || lower.endsWith(".vcmp"))
+					hasMaps = true;
```
IvanSavenko, line 899: What if we're importing *mod* with maps, e.g. Antagarich Burning or any map pack?

In this case it looks like we'll install 2 copies of maps - into Maps directory and as part of the mod

launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -1043,12 +1083,91 @@ void CModListView::installMods(QStringList archives)
...
+
+				if (!(name.endsWith(".h3m", Qt::CaseInsensitive) || name.endsWith(".h3c", Qt::CaseInsensitive) || name.endsWith(".vmap", Qt::CaseInsensitive) || name.endsWith(".vcmp", Qt::CaseInsensitive)))
+					continue;
+
+				QString destFile = destDir + name;
+
+				// prompt if exists
+				if (QFile::exists(destFile))
+				{
+					int ret = QMessageBox::question(this, tr("Map exists"), tr("Map '%1' already exists. Do you want to overwrite it?").arg(name), QMessageBox::Yes | QMessageBox::No, QMessageBox::No);
```
IvanSavenko: Can we also add "yes to all" / "no to all" to this dialog &  handle it?

Othervice importing large map pack for 2nd time by accident might be painful

launcher/modManager/cmodlistview_moc.cpp
GeorgeK1ng: When ZIP contains mod.json anywhere inside, it will be handled as mod, so it won't import maps.

launcher/modManager/cmodlistview_moc.cpp
GeorgeK1ng: This will require more investigating, as for each map file is called this map install function. It never passed list of maps from what I tried. Maybe unwanted behavior. So "Yes to all" never happens there.

launcher/modManager/cmodlistview_moc.cpp
IvanSavenko: Are you sure? When `hasModJson` is set, it would break out of the loop. But what if map was found before mod.json? In that case, both `hasModJson` and `hasMaps` will be set

launcher/modManager/cmodlistview_moc.cpp
IvanSavenko: Are you sure? This method receives list of maps. So local variable should work.

GeorgeK1ng: But it's called for each map standalone. So when you select 2 h3ms, you will never reach "Yes to all" as it seems when you select more maps they are not placed into list. It's probably bug, in import dialog processing. In any case I have added it. No idea how to fix it for now.

launcher/modManager/cmodlistview_moc.cpp
GeorgeK1ng: Fixed conditions, should be fine now.

GeorgeK1ng: Removed forgotten break.

Issue #5880, opened 2025-07-05
Ballista shoot in blind units, when exist non-blind


GrafDeVan:
Take from stream:  [https://www.youtube.com/live/oT-_l8Iq1OA?si=NFtuwW2LLoIPqPg0&t=4559](url)
blind [https://www.youtube.com/live/oT-_l8Iq1OA?si=lL09n8_z8QOrVzvX&t=4518](url)

Without the artillery skill, the ballista must first shoot at units that are not blinded, and only when there are no such units left does it begin to shoot at blinded ones.

PawelS-AoW:
Same with units turned to stone by Basilisks, and other similar effects. The Ballista seems to always target these first.

Alucard648:
Learn Artillery to avoid this issue.

PR #5961, opened 2025-07-31
Rework ballist and tower target acquisition

Opuszek:
Reworks target acquisition of ballista and towers to be similar to the original homm3.
Fixes - #5880

Tested on army:
1. 50 Archers
2. 50 Archers (close to ballista, no distance penalty)
3. 50 Archers (Air Shield)
4. 200 Sprites (Blind)
5. 200 Sprites
6. 2000  Dendroid Soldiers

ballista

Results were as following:
develop: 1-2-3, 6, 4-5
GOG Homm3: 2,5,1,3,6,4
feature branch: 5,2,1,3,6,4

Sprites barely edged out number 2 on the feature branch (~108 to ~106 calculated value).

### DIFF
--- a/server/battles/BattleFlowProcessor.cpp
+++ b/server/battles/BattleFlowProcessor.cpp
@@ -451,28 +451,27 @@ bool BattleFlowProcessor::tryMakeAutomaticActionOfBallistaOrTowers(const CBattle
 		attack.side = next->unitSide();
 		attack.stackNumber = next->unitId();
 
-		// TODO: unify logic with AI?
-		// Find best target using logic similar to H3 AI
-
-		const auto & isBetterTarget = [&battle](const battle::Unit * candidate, const battle::Unit * current)
+		const auto & getAttackValue = [&battle, &next] (const battle::Unit * unit)
 		{
-			bool candidateInsideWalls = battle.battleIsInsideWalls(candidate->getPosition());
-			bool currentInsideWalls = battle.battleIsInsideWalls(current->getPosition());
+			float singleHpValue = static_cast<float>(unit->unitType()->getAIValue()) / static_cast<float>(unit->getMaxHealth());
 
-			if (candidateInsideWalls != currentInsideWalls)
-				return candidateInsideWalls > currentInsideWalls;
+			int distance = BattleHex::getDistance(next->getPosition(), unit->getPosition());
+			BattleAttackInfo attackInfo(next, unit, distance, true);
+			DamageEstimation estimation = battle.calculateDmgRange(attackInfo);
+			float avgDmg = (static_cast<float>(estimation.damage.max) + static_cast<float>(estimation.damage.min)) / 2;
 
-			// also check for war machines - shooters are more dangerous than war machines, ballista or catapult
-			bool candidateCanShoot = candidate->canShoot() && candidate->unitType()->warMachine == ArtifactID::NONE;
-			bool currentCanShoot = current->canShoot() && current->unitType()->warMachine == ArtifactID::NONE;
+			return avgDmg * singleHpValue;
+		};
 
-			if (candidateCanShoot != currentCanShoot)
-				return candidateCanShoot > currentCanShoot;
+		const auto & isBetterTarget = [&getAttackValue](const battle::Unit * candidate, const battle::Unit * current)
+		{
+			bool candidateIsParalyzed = candidate->hasBonusOfType(BonusType::NOT_ACTIVE);
+			bool currentIsParalyzed = current->hasBonusOfType(BonusType::NOT_ACTIVE);
 
-			int64_t candidateTargetValue = static_cast<int64_t>(candidate->unitType()->getAIValue() * candidate->getCount());
-			int64_t currentTargetValue = static_cast<int64_t>(current->unitType()->getAIValue() * current->getCount());
+			if (candidateIsParalyzed != currentIsParalyzed)
+				return currentIsParalyzed > candidateIsParalyzed;
 
-			return candidateTargetValue > currentTargetValue;
+			return getAttackValue(candidate) > getAttackValue(current);
 		};
 
 		const battle::Unit * target = nullptr;


server/battles/BattleFlowProcessor.cpp
```diff
@@ -451,28 +451,27 @@ bool BattleFlowProcessor::tryMakeAutomaticActionOfBallistaOrTowers(const CBattle
...
+			float avgDmg = (static_cast<float>(estimation.damage.max) + static_cast<float>(estimation.damage.min)) / 2;
 
-			// also check for war machines - shooters are more dangerous than war machines, ballista or catapult
-			bool candidateCanShoot = candidate->canShoot() && candidate->unitType()->warMachine == ArtifactID::NONE;
-			bool currentCanShoot = current->canShoot() && current->unitType()->warMachine == ArtifactID::NONE;
+			return avgDmg * singleHpValue;
+		};
 
-			if (candidateCanShoot != currentCanShoot)
-				return candidateCanShoot > currentCanShoot;
```
IvanSavenko: Why were these parts removed?

Opuszek: As far as I've checked ballista and towers do not seem to prioritize units inside walls and do not prioritize archers per se.
Sprites
A big red arrow to the sprites that are getting wrecked.
 But retesting it I noticed that there is a bonus to the units that can attack this turn - if there are two stacks of the same unit the stack having a target within the reach will always be prioritized. I guess that's why in the original archers were attacked before sprites.
I found a wiki article with details:
https://heroes.thelazy.net/index.php/Arrow_tower#Arrow_towers_under_the_AI_control
I think I will change it to a draft and implement the logic as described there.

dydzio0614: current battle AI is not intended to be as close as possible to H3 but to be as strong as possible

Opuszek: Hmm... so maybe I will just add a 20% bonus for being able to attack this turn and call it a day?
The feature branch logic seems to me better than the existing one.
On develop the strongest stack is attacked (with range unit being prioritized before melee and units within walls before those without), when here the attack that can remove the biggest amount of  AIValue is chosen (as long as unit is not blinded or petrified). It takes such things like unit defense, defensive spells, distance penalty and value/hp ratio into consideration.

Opuszek: I don't think explicitly choosing range units over melee would be a good decision, those units have high value/hp ratio anyway, and we risk discarding other considerations.

IvanSavenko: This function was written mostly in regards with arrow towers in sieges, so arrow towers focus on units that can be a danger to defenders, namely - ranged units & units that are inside the walls (and likely - can reach defenders).

If you wish to adjust it - sure, but IMO target selection should remain reasonable. It makes sense to prioritize units that can attack ours units (so all ranged units that can shoot, and all melee units that can reach our units). While attacking Dendroids on left edge of battlefield while all our walls are intact is pointless.

Opuszek: I made a second attempt at it.
Currently it will never attack a paralyzed stack if there are not-paralyzed ones or machines, it will never attack a machine if there is a not-paralyzed stack and it will never attack units that cannot attack next turn if there are those that can.
Other than that, it uses average damage value (i.e. how much AI Value will be removed by an attack) to choose its target.
There is one snag - in case of melee units "can attack next turn" does not take into consideration a target can have its turn first and run away, but IMO it is a useful approximation anyway.

Opuszek: @IvanSavenko @dydzio0614  what do you think about it? Should I change to PR or cancel the draft?

dydzio0614: Was old behavior for arrow towers removed? I mean prioritizing targets inside walls. I am not familiar with this code so i can't tell by quickly looking at changes

Opuszek: Yes.
Instead it prioritizes enemy units that have an ally unit within range (can attack it during the next turn).
I think it is a better, more flexible behavior.

IvanSavenko: Not necessarily. It might be a good idea as an additional factor, but not as a replacement.

For example, if one of defenders leaves town (e.g. to destroy Catapult / kill enemy ranged units) this defender unit will be in range of most / all attackers.

However for towers it is still makes sense to prioritize enemies inside walls, since they may block/attack ranged defenders. If my Gold Dragons have took out a Catapult, I would still want towers to protect my Grand Elves.

Opuszek: Good point. 
What about such priorities: (behind the wall and can attack) -> (can attack) -> (behind the wall) -> (neither) -> machine -> paralyzed stack

dydzio0614: Seems good.

It would be also an improvement if towers / war machines won't hesistate to hit angel with 1 hitpoint instead of other targets, or 1 last pikeman to end battle - old code (before this pull request) leads to these targets having low priority due to attempt to maximize damage, so instead of killing last pikeman and ending battle towers will try to destroy catapult.

IvanSavenko: @Opuszek 
Sounds good to me too

@dydzio0614 
Not *against* it per se, as long as it won't result in towers *always* hitting pikeman - since towers are basically guaranteed to deal enough damage to kill pikeman even at full health. Not sure what's the best approach here...

I think BattleAI counts kills & taken HP separately. Formula is something like:
 `score = AI Value * kills + AI Value * total damage / max health`
As result BattleAI will prioritize finishing off units, but may also hit high AI Value targets in place of hitting pikeman if score is good enough. May be a good idea to reuse it or adjust it for autotargeting.

server/battles/BattleFlowProcessor.cpp
```diff
@@ -446,52 +446,85 @@ bool BattleFlowProcessor::tryMakeAutomaticActionOfBallistaOrTowers(const CBattle
...
+			float towerAttackValue;
+			const CStack * stack;
+		};
 
-			if (candidateInsideWalls != currentInsideWalls)
-				return candidateInsideWalls > currentInsideWalls;
+		const auto & getCanAttackNextTurn = [&battle] (const battle::Unit * unit)
+		{
+			if (unit->canShoot())
+				return true;
```
IvanSavenko: You might want to use `battle.battleCanShoot(unit)` instead - current code will return true for blocked shooters, which is probably undesired.

Issue #5928, opened 2025-07-14
Crash on starting single player game


fatal10110:
1. Choose map
2. Choose town / hero / bonus
3. Click play
4. Crash .. 

[VCMI_Client_log.txt](https://github.com/user-attachments/files/21221470/VCMI_Client_log.txt)

IvanSavenko:
Likely related to that map - "Talitha the Dragonmaster".
Will investigate.

Issue #5918, opened 2025-07-12
crash AI turn with a hole map (and No crash with vcmieagles)


GrafDeVan:
This is a save from the campaign. If I play with a face-down map, the game crashes. If I use a cheat code to open the map, then everything is fine.

[VCMI_Client_log.txt](https://github.com/user-attachments/files/21195448/VCMI_Client_log.txt)

[crash.zip](https://github.com/user-attachments/files/21195449/crash.zip)

[No_Crash_121.zip](https://github.com/user-attachments/files/21195450/No_Crash_121.zip)

Issue #5917, opened 2025-07-12
campaign: The victory condition is not met. An elixir has been assembled, but nothing happens


GrafDeVan:
<img width="1171" height="889" alt="Image" src="https://github.com/user-attachments/assets/5a32363c-46fe-4147-be02-c3fe307639df" />

<img width="453" height="455" alt="Image" src="https://github.com/user-attachments/assets/1289cb7b-67a9-4311-b7f4-e6fdfc245e04" />

Maybe I forgot how to complete this mission, but I have been to all the cities, nothing happens. Also, there is only one seer on the map.

save [Elixir.zip](https://github.com/user-attachments/files/21193066/Elixir.zip)

current develop, msvc, vcmi extras

Issue #5912, opened 2025-07-11
The "Buy all" button does not work further if the purchase queue fails


GrafDeVan:
https://youtu.be/FXkmbMr3ktk

misiokles:
Clean game or with some mods?

GrafDeVan:
> Clean game or with some mods?

develop 1.7.0.0d9fd09a6d539e093145bb3d02ed91b50d90db3b, clean game, only vcmi extras

misiokles:
Buy All option has always failed, but it was always because bad written mods. Question is why purchase queue fails in clean game.

GrafDeVan:
> Buy All option has always failed, but it was always because bad written mods. Question is why purchase queue fails in clean game.

The hero has no place for wolf riders

IvanSavenko:
Found it, will fix. Happens only if there is a garrisoned hero.

PR #5962, opened 2025-07-31
Fixes for recently reported issues

IvanSavenko:
Changes:

- Add debug logging for mods with invalid town building config
- Add check for victory condition on assembling artifact
- Fix crash on loading town with unit that has ability propagated to army
- Fix regression - crash on attempt to load saved game
- Fix Solmyr/Yog receiving spellbook on transferring to next scenario
- Add workaround for loading save when entity was moved to another mod
- Fix regresssion - crash on transferring hero to next scenario
- Fix quick recruitment failing when there are no free slots in army
- Fix crash on AI capturing artifacts in battle while covered by FoW

Fixed issues:

- #5872
- #5918
- #5906
- #5917
- #5928
- #5912
- #5719

### DIFF
--- a/client/NetPacksClient.cpp
+++ b/client/NetPacksClient.cpp
@@ -828,7 +828,7 @@ void ApplyClientNetPackVisitor::visitBattleResultsApplied(BattleResultsApplied &
 {
 	if(!pack.learnedSpells.spells.empty())
 	{
-		const auto hero = GAME->interface()->cb->getHero(pack.learnedSpells.hid);
+		const auto * hero = cl.gameInfo().getHero(pack.learnedSpells.hid);
 		assert(hero);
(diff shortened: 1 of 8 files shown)


GrafDeVan:
Hota maps: 
[Crash when try to start.zip](https://github.com/user-attachments/files/21537471/Crash.when.try.to.start.zip)
[Game closes when try to start.zip](https://github.com/user-attachments/files/21537473/Game.closes.when.try.to.start.zip)

IvanSavenko:
Checked those maps:
- some are now fixed
- some require fix to hota mod (map uses hero portraits added by HotA, but not in our hota mod)
- and few more are known bug, might also happen in h3/sod maps. Will fix later, but definitely before 1.7

Also added fix for #2536

Issue #4893, opened 2024-11-11
How to add next layer to game?


jaicmatija:
Heroes 3 has surface and underground layer
how to add next layer to game?
Let's call it "above" It could be next layer to the game maybe some clouds what you could access only by teleporters?
This would add third layer to the game and more space and interesting interactions to it, It could have uniq features like that there is no combat there...

Is is possible to add it to vcmi somehow? mod or feature?

jaicmatija:
Also if this is possible then we could also add 4th layer wich could be space (the final leyer)
There is no oxigen so only non living can go there

Olegmods:
Actually a proposal that came long ago before.

Certainly it's too much pain to make that one, since there are RMG, UI and more stuff needed to make that one.

Plus AFAIK devs already have a lot to do.

Macron1Robot:
At least just multiple subterra levels are easy doable, i think

Olegmods:
I guess it's still needs UI and a lot of code.

Alucard648:
Cloud/space layers are tecnically the same as extra subterranian levels. Many custom maps using above-ground terrains in undergaound layer for extra space on the map and underground terrain in upper layer to simulate caverns without using underground layer. As for high areas, most map builders use impassable water with Magic clouds on top of it to simulate high walkable mountains. HotA Conflux Highlands worked really well in such areas.

Laserlicht:
Open points:
- [x] adding types for layers (instead of directly addressing layer number) - done by  #6619
- [x] fix subterranean gate - done by #6661
- [x] terrains need support for `MapLayerId` instead of fixed "SUB" and "SURFACE" - done by  #6619
- [ ] rmg should handle based on `MapLayerId` (allow to have e.g. multiple undergrounds or changed orders & choosing terrains with `MapLayerId` - allows e.g. sky layer)
- [ ] rmg ui need an input for selecting `MapLayerId` for each layer (or template based?)
- [ ] ui button to change level should be based on `MapLayerId`, not on fixed levels (if e.g. first level is underground instead of second)
- [ ] use `MapLayerId` in some texts (subterranean gate?)
- [ ] layer as source of bonuses (e.g. creatures/hero in sky are slower than on ground) and maybe also limiter
  - but can also archived by terrains for now
- [ ] some minor hardcoded `isSurface` & `isUnderground` (rewards?)

@dydzio0614 @DjWarmonger You may interested in some tasks? Biggest tasks are already done...

PR #5964, opened 2025-08-01
multilevel support

Laserlicht:
Support for multiple levels (more than 2).
Layers > 2 are handled as surface layers (for rmg).

Missing stuff (known missing parts are marked with `// TODO: multilevel support` in code):
- Nicer handling for button in game ui (atm, layer 1 has underground icon, other layers normal icon)
- RMG menu ingame
- Map overview (right click on new/loading list)
- Info windows (map on popup on different map objects)
- RMG support itself (RMG generated maps with > 2 layers are trash atm -> possible to test in map editor)
- ... unknown stuff ...

Map editor shows warning `Expect issues.` when using more than two levels.

Related: #4893 #5966

### DIFF
--- a/mapeditor/mapcontroller.h
+++ b/mapeditor/mapcontroller.h
@@ -105,7 +105,7 @@ private:
 	std::vector<std::unique_ptr<CGObjectInstance>> _clipboard;
 	int _clipboardShiftIndex = 0;
 
-	const int MAX_LEVELS = 10;
+	const int MAX_LEVELS = 10; // TODO: multilevel support: remove this constant
 
 	std::map<TerrainId, std::unique_ptr<EditorObstaclePlacer>> _obstaclePainters;
 


Alucard648:
Speaking of layers in rmg: will it be possible to define which layers are underground layers in custom RMG templates?

Another suggestion (maybe in 1.8+) is "sky" layer with normally impassable sky terrain, except Fly spell, HotA skyships etc. HotA Conflux Highlands terrain will look best on this layer.

DjWarmonger:
> Speaking of layers in rmg: will it be possible to define which layers are underground layers in custom RMG templates?

Please create a feature request.

Laserlicht:
Missing stuff is for following PR's.

mapeditor/mapcontroller.cpp
```diff
@@ -45,7 +45,7 @@ MapController::MapController(QObject * parent)
 
 MapController::MapController(MainWindow * m): main(m)
 {
-	for(int i : {0, 1})
+	for(int i = 0; i < MAX_LEVELS; i++)
```
IvanSavenko, line 48: I might be wrong, and that would be great, but such loops might be destroying RAM on large maps.

Map Editor is *very* RAM-hungry and allocates massive memory for each map layer. Can you check that memory usage for XL+U (or G+U if you're brave enough) is the same as before?

mapeditor/mapcontroller.h
```diff
@@ -100,11 +100,13 @@ class MapController : public QObject
...
 	std::unique_ptr<MapHandler> _mapHandler;
 	MainWindow * main;
-	mutable std::array<std::unique_ptr<MapScene>, 2> _scenes;
-	mutable std::array<std::unique_ptr<MinimapScene>, 2> _miniscenes;
+	mutable std::map<int, std::unique_ptr<MapScene>> _scenes;
+	mutable std::map<int, std::unique_ptr<MinimapScene>> _miniscenes;
 	std::vector<std::unique_ptr<CGObjectInstance>> _clipboard;
 	int _clipboardShiftIndex = 0;
 
+	const int MAX_LEVELS = 10;
```
IvanSavenko:
```suggestion
	const int MAX_LEVELS = 10; // TODO: multilevel support: remove this constant
```

mapeditor/mapcontroller.cpp
Laserlicht: Checked with XL+U. Same as before.
~1,5GB for map editor after fresh loaded map.

Alucard648: And now try to load XL+10U... RAM usage needs to be optimized.

Laserlicht: Yes. But nothing to do with this PR.

PR #5971, opened 2025-08-02
Uncheck all tasks and disable tasks page for PR builds

GeorgeK1ng:
Lets see how will look installer in PRs compilation after this change.

### DIFF
--- a/CI/wininstaller/installer.iss
+++ b/CI/wininstaller/installer.iss
@@ -172,6 +172,7 @@ var
   InstallModePage: TInputOptionWizardPage;
   FooterLabel: TLabel;
   IsUpgrade: Boolean;
+  IsPR: Boolean;
   Heroes3Path: String;
   GlobalUserName: String;
   GlobalUserDocsFolder: String;
@@ -517,7 +518,20 @@ end;
 
 
 procedure InitializeWizard();
+var
+  i: Integer;
 begin
+
+  // Detect if installer name contains "-PR-"
+  IsPR := Pos('-PR-', ExpandConstant('{#InstallerName}')) > 0;
+ 
+  if IsPR then
+  begin
+    // Uncheck all tasks manually for PR builds
+    for i := 0 to WizardForm.TasksList.Items.Count - 1 do
+      WizardForm.TasksList.Checked[i] := False;
+  end;
+
   // Check if the application is already installed
   if not IsUpgrade then
   begin
@@ -578,7 +592,14 @@ end;
 function ShouldSkipPage(PageID: Integer): Boolean;
 begin
   Result := False; // Default is not to skip the page
-  
+
+  // Skip Tasks page if this is a PR build
+  if IsPR and (PageID = wpSelectTasks) then
+  begin
+    Result := True;
+    Exit;
+  end;
+
   if IsUpgrade then
   begin
     if (PageID = wpLicense) or (PageID = wpSelectTasks) or (PageID = wpReady) then
@@ -863,3 +884,4 @@ begin
       Abort;
   end;
 end;
+


GeorgeK1ng:
Task page is now correctly disabled for PR builds and all settings from that page are not applied.
Before:
image

After:
image

GeorgeK1ng:
@IvanSavenko Anything else to do here? Wanna to have fixed installer in develop builds soon as possible.

launcher/mainwindow_moc.cpp
```diff
@@ -81,7 +81,7 @@ MainWindow::MainWindow(QWidget * parent)
 	load(); // load FS before UI
 
 	bool setupCompleted = settings["launcher"]["setupCompleted"].Bool();
-
+
```
kambala-decapitator: please drop this change

GeorgeK1ng: Done, it was leftover from previous touches.

CI/wininstaller/installer.iss
```diff
@@ -863,3 +965,11 @@ begin
...
   end;
 end;
+
+
+
+
+
+
+
+
```
IvanSavenko: What is this?

launcher/mainwindow_moc.cpp
```diff
@@ -130,21 +130,42 @@ MainWindow::MainWindow(QWidget * parent)
 
 void MainWindow::detectPreferredLanguage()
 {
-	auto preferredLanguages = QLocale::system().uiLanguages();
+	// Skip autodetection if language is already set to a valid VCMI language
+	std::string currentLanguage = settings["general"]["language"].String();
+
+	if (!currentLanguage.empty())
+	{
```
IvanSavenko: This condition will always be true

launcher/mainwindow_moc.cpp
```diff
@@ -130,21 +130,42 @@ MainWindow::MainWindow(QWidget * parent)
...
+			{
+				logGlobal->info("Language '%s' is already valid, skipping autodetection", currentLanguage);
+				return;
+			}
+		}
+	}
+
+	// Proceed with autodetection
+	auto appLanguages = QLocale().uiLanguages();
+	auto sysLanguages = QLocale::system().uiLanguages();
```
IvanSavenko: Will this really result in different lists?

https://doc.qt.io/qt-6/qlocale.html#QLocale

> If no default locale was set using [setDefault](https://doc.qt.io/qt-6/qlocale.html#setDefault)(), this locale will be the same as the one returned by [system](https://doc.qt.io/qt-6/qlocale.html#system)().

launcher/mainwindow_moc.cpp
GeorgeK1ng: Will be not empty only when Launcher settings is not finished / settings created by installer if I understand that process correctly.

CI/wininstaller/installer.iss
GeorgeK1ng: Removed, somehow during online edit GitHub added empty end line after each touch.

launcher/mainwindow_moc.cpp
IvanSavenko: Nope. 

When settings are loaded, all fields that are missing from config (or if entire config is missing) will be set to their default value according to schema

launcher/mainwindow_moc.cpp
GeorgeK1ng: Only in some cases. In my test on English Windows 10 with Czech regional settings it correctly set it to Czech which is correct. Previously I got English based on default system language.

launcher/mainwindow_moc.cpp
GeorgeK1ng: I see default from here -https://github.com/vcmi/vcmi/blob/4129e218f7356f074bc32661dac3af14847f68a1/config/schemas/settings.json#L78 , forgot about that. 

So that condition can be completely removed I quess?

launcher/mainwindow_moc.cpp
IvanSavenko: Perhaps we should *always* use default-constructed locale then?

What OS / system language / regional settings you had that produced different results for these two methods?

Tested it on Linux - it returned same for both. HOWEVER when starting app through debugger, `uiLanguages` would always return English/US. Non-debugger sessions always had same (and correct) result

launcher/mainwindow_moc.cpp
IvanSavenko: Sort of. Probably remove entire block (and attempts by installer to set game language).

At the moment I don't see clean way to pass information from installer to launcher. One option would be to use command line, but probably not a quick fix & don't see it as critical enough - system language is probably sane default in most cases.

GeorgeK1ng:
I will drop required changes in few hours in this PR.

GeorgeK1ng:
Reverted launcher touches, disabled settings.json creation.

GeorgeK1ng:
@IvanSavenko This should be safe, only installer improvements are included.

For future PR I will create from scratch custom select target directory so it will became configurable for PR And Upgrade setup together.

PR #5976, opened 2025-08-03
Generate .zip installer archive via CPack on Windows

IvanSavenko:
(Untested, will test on CI)


kambala-decapitator (approved):
Looks correct

PR #5979, opened 2025-08-04
Fixes for bonus system and tests

### LINKED ISSUE #5975: Intimidation skill reduces morale of friendly creatures instead of enemies
**Describe the bug**
The Intimidation skill is not working as expected. Instead of reducing the morale of enemy creatures, it currently reduces the morale of friendly creatures.

Maybe there is something wrong with the OPPOSITE_SIDE limiter or the BONUS_OWNER_UPDATER. The bonus might be incorrectly applied due to a mistake in side-checking logic.

**To Reproduce**
1. Enable the Tide of War mod.

2. Let hero learn the Intimidation skill.

3. Start a battle and observe the morale of your own creatures.

4. Notice that the morale penalty is applied to your own units instead of the enemies.

**Expected behavior**
Creatures with the Intimidation skill should reduce the morale of enemy units, not allies.

**Actual behavior**
When a creature with the Intimidation skill is present on the battlefield, friendly creatures (on the same side) receive a morale penalty, while enemy creatures remain unaffected.

**Version**

- OS: Windows, macOS Intel, macOS ARM, Android, Linux, iOS
- Version: 1.7 develop

IvanSavenko:
- Fixed bonus propagation. Fixes #5975
- Added tests to cover recent issues with bonus system
- Fixed tests that apparently were failing for some time
- Re-enabled around half of previously disabled tests and fixed their failures
- Attempt to make failure of tests to fail workflow without terminating entire CI matrix - so daily builds will be generated even with failing tests

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -16,6 +16,7 @@ env:
 jobs:
   build:
     needs: extract-version
+    continue-on-error: ${{ matrix.test == 1 }}
     strategy:
       matrix:
         include:
@@ -301,7 +302,6 @@ jobs:
       env:
         HEROES_3_DATA_PASSWORD: ${{ secrets.HEROES_3_DATA_PASSWORD }}
       if: ${{ env.HEROES_3_DATA_PASSWORD != '' && matrix.test == 1 }}
-      continue-on-error: true
       run: |
         ctest --preset ${{matrix.preset}}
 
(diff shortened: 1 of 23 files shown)


IvanSavenko:
Seems to be working - failing tests are reported as failure, but don't terminate rest of workflow: https://github.com/vcmi/vcmi/actions/runs/16720562498/job/47323855112

There are still some disabled tests, but will keep those as it for now - haven't found any actual bugs, all test failures are due to environment setup / invalid access, like accessing non-existing test entities in game library.

Issue #4916, opened 2024-11-16
For Combined or Fused artifacts show the list of components


krs0:
**Describe your proposal**
When you right click a component of a combined artifact you are presented with 2 separate pieces of information:

- info on right-clicked artifact
- info that you could assemble a new artifact. 

This is a little confusing and does not provide info on components.

![image](https://github.com/user-attachments/assets/1e5d3ec5-9e96-433f-8520-db7062ecc316)

Would be nice if the full info regarding the selected artifact is shown on right click (like usual when you do not have all components)..

![image](https://github.com/user-attachments/assets/ab55962f-f752-480d-ac19-278bbc00a359)

And the question on assembling it would be separate popup! EG: new popup after rightmouse release. And this popup will have info only on combined artifact. Info, Icon, list of components.

Issue #828, opened 2022-09-05
tooltip of artefact set piece is shown as Empty


kambala-decapitator:
I think it's better to show set piece name there. Dunno the behaviour of original in this case.

![Screenshot 2022-09-05 at 16 16 48](https://user-images.githubusercontent.com/1557784/188462530-55c13741-e6c8-45bf-8cd9-e1a646c597c1.png)

kambala-decapitator:
same behavior in oh3

GeorgeK1ng:
Instead of empty "Occupied by .." will be better.

PR #5980, opened 2025-08-04
Artifact qol improvements

SoundSSGood:
Done https://github.com/vcmi/vcmi/issues/828
Done https://github.com/vcmi/vcmi/issues/4916

Now is possible to see the name of the artifact even in a locked slot.
image

Before asking about the artifact assembly, a window with information about the artifact will be shown.

### DIFF
--- a/client/ArtifactsUIController.cpp
+++ b/client/ArtifactsUIController.cpp
@@ -60,7 +60,7 @@ bool ArtifactsUIController::askToAssemble(const CGHeroInstance * hero, const Art
 	auto assemblyPossibilities = ArtifactUtils::assemblyPossibilities(hero, art->getTypeId(), onlyEquipped);
 	if(!assemblyPossibilities.empty())
 	{
-		auto askThread = new std::thread([this, hero, art, slot, assemblyPossibilities, checkIgnored]() -> void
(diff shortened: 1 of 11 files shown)


client/gui/WindowHandler.h
```diff
@@ -17,9 +17,6 @@ class WindowHandler
 	/// (includes adventure map, window windows, all kind of active dialogs, and so on)
 	std::vector<std::shared_ptr<IShowActivatable>> windowsStack;
 
-	/// Temporary list of recently popped windows
-	std::vector<std::shared_ptr<IShowActivatable>> disposed;
-
```
IvanSavenko: While I am not against it, this could be potentially dangerous change.

Please check scenario when:
- you build something in town. I think we recreate castle interface when this happens.
- you have multiple towns of the same type - switching between them should not break music playback (or crash obviously)
- you have multiple towns of different types - music should change correctly

Basically - make sure that `CCastleInterface::~CCastleInterface` is called at the right moment & has correct logic. Can't think of any other similar cases at the moment.

SoundSSGood: Yes, there were issues with the music. Removed unnecessary changes.

Issue #5195, opened 2025-01-03
fix ccache


Laserlicht:
**Describe the bug**
ccache is only working with iOS and Ubuntu.

E.g. Android is broken... Please fix if possible.

kambala-decapitator:
what makes you think so?

Laserlicht:
> what makes you think so?

If you build the almost same build again the ubuntu 24 & ios CI-jobs finished after 3 mins. All other builds needs 15-25 min.

Is a regression from changes in building/dependencies/whatever. A few month ago all jobs (except msvc) finished after a few mins.

Laserlicht:
E.g. here for example: https://github.com/vcmi/vcmi/actions/runs/11112140534 you can see the timings in the past.

kambala-decapitator:
I suggest to verify locally that ccache is broken, as it could be GhA issue for example. I see that it doesn't have anything to restore from cache, although cache seems created after each successful pr build.

Laserlicht:
> In the recent pr builds Linux and iOS also take full time

It's depends what's the build before because the cache is not unlimited. E.g. here (from yesterday) it works: https://github.com/vcmi/vcmi/actions/runs/12590072915

PR #5981, opened 2025-08-04
Try to fix non-functioning ccache in some CI runs

IvanSavenko:
- [x] Apparently some CI runs run without ccache at all, due to bug in bash script syntax 
- [x] When we have multiple active branches (beta/develop/master) ccache basically dies since cache between beta/develop branches is shared and is constantly invalidated
- [ ] Ccache *should* support MSVC, but we have it disabled, initially - here: #4132, and remaining bits - here: #5932. Would be great to figure out why that happens to get CI speedup, especially since we now have 3 msvc runs and not 1.
- [ ] For unknown reason, on Android platforms ccache runs with 100% misses

Fixed issues:
- Partial #5195

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -283,7 +283,7 @@ jobs:
         elif [[ (${{matrix.preset}} == android-conan-ninja-release) && (${{github.ref}} != 'refs/heads/master') ]]
         then
             cmake -DENABLE_CCACHE:BOOL=ON -DANDROID_GRADLE_PROPERTIES="applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily;applicationVariant=daily" --preset ${{ matrix.preset }}
-        elif [[ ${{startsWith(matrix.platform, 'msvc') }} ]]
+        elif ${{startsWith(matrix.platform, 'msvc') }}
         then
             cmake --preset ${{ matrix.preset }}
         else
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -25,13 +25,13 @@ jobs:
             before_install: linux_qt6.sh
             preset: linux-clang-test
 
-          - platform: linux
+          - platform: linux-qt5
             os: ubuntu-24.04
             test: 1
             before_install: linux_qt5.sh
             preset: linux-gcc-test
 
-          - platform: linux
+          - platform: linux-debug
             os: ubuntu-22.04
             test: 0
             before_install: linux_qt5.sh
@@ -189,21 +189,21 @@ jobs:
       uses: hendrikmuhs/ccache-action@v1.2
       if: ${{ github.event.number != '' && !startsWith(matrix.platform, 'msvc') }}
       with:
-        key: ${{ matrix.preset }}-PR-${{ github.event.number }}
+        key: ${{ matrix.platform }}-PR-${{ github.event.number }}
         restore-keys: |
-          ${{ matrix.preset }}-PR-${{ github.event.number }}
-          ${{ matrix.preset }}-no-PR
+          ${{ matrix.platform }}-PR-${{ github.event.number }}
+          ${{ matrix.platform }}-branch-${{ github.base_ref }}
         # actual cache takes up less space, at most ~1 GB
         max-size: "5G"
         verbose: 2
 
-    - name: ccache for everything but PRs
+    - name: ccache for branch builds
       uses: hendrikmuhs/ccache-action@v1.2
-      if: ${{ (!startsWith(matrix.platform, 'msvc')) && (github.repository == 'vcmi/vcmi' && github.event.number == '' && (github.ref == 'refs/heads/develop' || github.ref == 'refs/heads/beta' || github.ref == 'refs/heads/master')) || github.repository != 'vcmi/vcmi' }}
+      if: ${{ github.event.number == '' && !startsWith(matrix.platform, 'msvc')}}
       with:
-        key: ${{ matrix.preset }}-no-PR
+        key: ${{ matrix.platform }}-${{ github.ref_name }}
         restore-keys: |
-          ${{ matrix.preset }}-no-PR
+          ${{ matrix.platform }}-branch-${{ github.ref_name }}
         # actual cache takes up less space, at most ~1 GB
         max-size: "5G"
         verbose: 2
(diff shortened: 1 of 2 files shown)


GeorgeK1ng:
Seem it require creating sccache folder before "ccache for PRs (msvc)" step.

```
    - name: Ensure sccache folder exists
      shell: bash
      run: |
        mkdir -p "${GITHUB_WORKSPACE}/.sccache"
```
        
 This will just save it correcly, next run will restore it, but still not uses that cache.

And to get rid of mac/ios ninja warning macos.sh

```
#!/usr/bin/env bash
set -e

# Set specific Xcode version (ensure it exists on the runner)
echo "DEVELOPER_DIR=/Applications/Xcode_14.2.app" >> "$GITHUB_ENV"

# Check if ninja is installed, install only if missing
if ! command -v ninja >/dev/null 2>&1; then
	echo "Ninja not found, installing via Homebrew..."
	brew install ninja
else
	echo "Ninja already installed, skipping installation."
fi

```

kambala-decapitator:
simply drop `brew install ninja`, it's now preinstalled on the runners

CMakeLists.txt
```diff
@@ -303,17 +306,17 @@ if(MINGW OR MSVC)
...
+			# https://github.com/mozilla/sccache
+			if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+			  string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
+			  string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")
+			elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
+			  string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
+			  string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
+			elseif(CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
+			  string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
+			  string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO}")
```
kambala-decapitator: can't you inject something like "disable /Zi and enable /Z7" to preset?

also, use tabs
```suggestion
				string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG}")
				string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")
			elseif(CMAKE_BUILD_TYPE STREQUAL "Release")
				string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE}")
				string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
			elseif(CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
				string(REPLACE "/Zi" "/Z7" CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO}")
				string(REPLACE "/Zi" "/Z7" CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO}")
```

Laserlicht:
> For unknown reason, on Android platforms ccache runs with 100% misses

Seems like this PR broke it. https://github.com/vcmi/vcmi/pull/4545 (looked into run times of android jobs).
But there was a lot changes...

CMakeLists.txt
IvanSavenko: Newer cmake has option to do this automatically, but don't want to bump it.

Will fix, but only if this will actually help. So far I was unable to make msvc use ccache:
- ccache fails on running AUTO_RCC (old problem, unresolved)
- sccache seems to be doing nothing.

From what I found, switching from msbuild to ninja *should* fix ccache/sccache, but not idea how easy this change would be

IvanSavenko:
> Seems like this PR broke it. https://github.com/vcmi/vcmi/pull/4545 (looked into run times of android jobs).

But not many might have effect on Android:
- we now use NDK from conan, instead of preinstalled
- Android CI now runs on Ubuntu runners.

Still no idea how either of those could break ccache.

CMakeLists.txt
kambala-decapitator: switching generator should be trivial. In the worst case try `Ninja Multi-Config` in case msvc-specifics are tied to a multiconfig generator.

kambala-decapitator:
FYI NDK from Conan is the same as the official distribution just wrapped in a Conan recipe

CMakeLists.txt
GeorgeK1ng: I read on various sources, switching to Ninja is solution, but also something about broken PDBs. Probably needs to be tested. 

Maybe worth to try this before switching to Ninja? https://github.com/ccache/ccache/discussions/1140#discussioncomment-3611405

IvanSavenko: @kambala-decapitator 
Can try ninja I guess.

@GeorgeK1ng 
This is what we had before. Result it failure on AUTO_RCC which is needed for Launcher/Editor.

IvanSavenko: Broken pdb's *should* be fixed by this code block in question.

But to be honest not much of a difference. I no longer have msvc set up to use pdb's, and Nullkiller who also used pdb is inactive. In other words - at the moment we can't open crash dumps, so pdb's are unused.

kambala-decapitator: well, I have Windows with VS installed, can run it either normally or through a VM

IvanSavenko:
Sort of finished. Or at least I don't want to spend more time on this.

- Linux, macOS: ccache should now work as expected

- Windows: completely disabled. In theory, switching from msbuild to Ninja (or `Ninja Multi-Config` ) might fix ccache, and if not - sccache might work instead (supported by action that we use). However I was unable to switch to Ninja - it requires activation of msvc compiler that targets desired platform first, which I can't get right. Might have been able to figure out if I had msvc setup available, but not in blind.

- Android: enabled, but inactive (100% cache miss). Reason unknown, likely comes from Android changes in #4545

PR #5984, opened 2025-08-05
Setup workflow for signing VCMI binaries using SignPath

IvanSavenko:


IvanSavenko:
Seems to be blocked by https://github.com/SignPath/github-action-submit-signing-request/issues/10

PR #5989, opened 2025-08-06
Improvements for Github CI workflow

IvanSavenko:
Over time our main "build" workflow grow in size, with a lot of parameters & conditional steps, which makes it hard to navigate & locate part you're interested in.

Windows installer pushed this even further with multiple steps that only run on Windows.

This PR reorganizes workflow to make it more manageable:
- Linux runners now run in separate job. Reasoning: their workflow is very different from rest of jobs - no prebuilts, no daily builds, no installers. Even build step is different since it explicitly sets compilers.
-  Moved running tests to Linux runner jobs. For now test run on all Linux runners. Might be a bit excessive, but tests are rather fast & might detect some edge cases like undefined behavior.
- Replaced single linux-clang job with two jobs: clang-13/qt5/release and clang-18/qt6/debug to cover more configurations.
- Moved generation of Windows installer into separate job that uses artifact from main build job as input. Reasoning is simple - all steps in this job are Windows-only, so moving it to separate job allows to simplify main build job. Extract-version job was integrated into this job.
- Give more human-readable names to jobs. For example: `build (ios, macos-13, 0, 1, 1, R...` -> `Build (ios)` to make it easier to locate specific job
- Switched Windows runners to use windows-2025 image instead of windows-latest to make sure that we won't get random failures when Github switches windows-latest image next month
- Removed some ancient leftovers from other CI services, like Travis.
- Replaced multiple checks for develop/beta/master branch with simpler check for whether action was initiated as PR

Some relatively small bits of workflow were unfortunately copy-pasted, but overall this should help a bit with readability. There are still places that can be improved or clarified. Might revisit & fix those, but a bit later in a separate PR - don't want to make massive changes in a single PR at once.


.github/workflows/github.yml
```diff
@@ -270,18 +266,11 @@ jobs:
...
-        echo "Moving $builtAabPath to $ANDROID_AAB_PATH"
-        mv "$builtApkPath" "$ANDROID_APK_PATH"
-        mv "$builtAabPath" "$ANDROID_AAB_PATH"
-        echo "ANDROID_APK_PATH=$ANDROID_APK_PATH" >> $GITHUB_ENV
-        echo "ANDROID_AAB_PATH=$ANDROID_AAB_PATH" >> $GITHUB_ENV
-
-    - name: Artifacts
+        OUTPUT_DIRECTORY="${{ github.workspace }}/out/build/${{ matrix.preset }}/android-build/vcmi-app/build/outputs"
+        mv "$OUTPUT_DIRECTORY/apk/release/vcmi-release.apk" "${{ github.workspace }}/$VCMI_PACKAGE_FILE_NAME.apk"
+        mv "$OUTPUT_DIRECTORY/bundle/release/vcmi-release.aab" "${{ github.workspace }}/$VCMI_PACKAGE_FILE_NAME.aab"
```
kambala-decapitator, line 271: wouldn't `mv` fail when there's no input file thus failing the whole job?

.github/workflows/github.yml
```diff
@@ -453,35 +336,219 @@ jobs:
...
+
+        major=$(grep -m 1 "VCMI_VERSION_MAJOR" "$filePath" | tr -d -c 0-9)
+        minor=$(grep -m 1 "VCMI_VERSION_MINOR" "$filePath" | tr -d -c 0-9)
+        patch=$(grep -m 1 "VCMI_VERSION_PATCH" "$filePath" | tr -d -c 0-9)
+
+        short_version="${major}.${minor}.${patch}"
+        version_timestamp=$(date +"%Y%m%d%H%M%S")
+
+        echo "short_version=${short_version}" >> "$GITHUB_OUTPUT"
+        echo "version_timestamp=${version_timestamp}" >> "$GITHUB_OUTPUT"
```
kambala-decapitator, line 486: you could create simple env vars now

.github/workflows/github.yml
IvanSavenko: True, but:
- `mv` was unconditional before, so not really a change
- I'd rather have failing job (that would be easy to notice) than hidden error that would only be discovered months later, like we had with failing tests recently.

In this particular case - I am not sure why we were locating apk/aab via wildcards. I though that this is needed since name is unpredictable (e.g. has some random hash or timestamp), but name is fixed: `vcmi-release.apk` (or aab)

.github/workflows/github.yml
IvanSavenko: Yeah, good point.

And might be better to merge this step into `Build Number` step, which already does similar tasks.

.github/workflows/github.yml
kambala-decapitator: but why not doing `! [ -f <aab> ] || mv ...` ? AAB isn't generated on every run. (or a similar check when you know that AAB is expected)

wildcards: IIRC to avoid typing long path, don't really remember out of my head why I chose to do so.

IvanSavenko: AAB is generated on every run, but not uploaded as artifact on every run

PR #5996, opened 2025-08-08
Fix Chronicles extraction crash

Laserlicht:
There is a new version of the gog installers for chronicles available with a different folder structure than before.

### DIFF
--- a/docs/players/Heroes_Chronicles.md
+++ b/docs/players/Heroes_Chronicles.md
@@ -9,6 +9,8 @@ gog.com download page:
 
 For each chapter you will need a seperate exe file to install (if you want to install all, you'll need 8 exe files).
 
+There is also an "All in one" installer as alternative. Just login into your gog account and open this url: [All in one installer](https://www.gog.com/downloads/heroes_chronicles_all_chapters/en1installer0)
+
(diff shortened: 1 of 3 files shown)


kambala-decapitator:
Could you also prevent the crash and show an error dialog instead? What is the reason of the crash BTW?

Laserlicht:
The crash happened when trying to access a non existing directory.

Then `entryList` returns empty list, but we try to access `front` which causes the crash. We use this to access folder independent from case. We assumed, the files were correct (in this case the chronicle episode was already correctly detected).

Of course it could be nicer with error handling. Would require some refactoring. But testing is not great here... (There are differences between "All in one", the old installers, the new installers, the episodes itself - everything that should retested on changes).

Also we could change it to a "find file" approach, That it searches our files independent from structure.

So it's a quick fix for now. Hopefully they will not change the installer structure soon again.

GeorgeK1ng:
Good is they not bumped Inno Setup version for now.

Laserlicht:
> Good is they not bumped Inno Setup version for now.

Yes.
But in upstream commits and PR's there is also support for newer versions. So, we could bump our fork to support these.
In general the upstream innoextract is not really active maintained. :(

And in case of a new inno setup version it should throw an handled exception.

GeorgeK1ng:
Was also thinking about propagating required changes to our fork. Also adding workflow to make binaries for extracting our new installer easily when needed.

PR #6005, opened 2025-08-11
fix: monthly monsters aren't linked with other map objects at creatio…

thehonestcto:
…n, leaving Nullkiller unable to see the next danger after killing a monthly monster;
plus 2 minor log trace renaming to make them easier to spot in logs when debugging

Hi, everyone!

I've seen a funny yet hard to trace bug, but I traced it in the end. The random montly monsters which appear ever X months according to the new month logic are not linked with other monsters on the map when chains are calculated.
You'll see in the attached saved game (needs 1.6.8 to run it) that one of tan's players is dying after killing a monthly monster because it lands into the danger are of another neutral without realizing it. Since that neutral is a lot stringer, our tan hero has no chance. Nevertheless, it doesn't matter if it's stronger or not, this bug shouldn't happen.

The best solution would be to properly link the monthly spawned creatures with the other creatures on the map, but I'm not that familiar to do that without extra input, therefore I came up with a safe solution: to spawn monthly monsters only on coordinates which are not blocked around them. This way they never land next to another neutral, so problem doesn't happen, but the original problem is still there.

Saved game 1.6.8:
[game5 tan chain bug.vsgm1.zip](https://github.com/user-attachments/files/21721610/game5.tan.chain.bug.vsgm1.zip)

You'll see 2 other minor log renamings for helping with local debugging, because both logs were producing the same message, making it harder to spot where they came from. I hope you don't mind.

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -3996,7 +3996,7 @@ void CGameHandler::spawnWanderingMonsters(CreatureID creatureID)
 	std::vector<int3>::iterator tile;
 	std::vector<int3> tiles;
 	gameState().getFreeTiles(tiles, true);
-	ui32 amount = tiles.size() / 100; //Chance is 1% for each tile. Higher than the original 0.5% because there are fewer with nearby unguarded
+	ui32 amount = tiles.size() / 200; //Chance is 0.5% for each tile
 
 	RandomGeneratorUtil::randomShuffle(tiles, getRandomGenerator());
 	logGlobal->trace("Spawning wandering monsters. Found %d free tiles. Creature type: %d", tiles.size(), creatureID.num);


lib/callback/CGameInfoCallback.cpp
```diff
@@ -808,7 +808,38 @@ void CGameInfoCallback::getFreeTiles(std::vector<int3> & tiles) const
...
+							break;
+						}
+					}
+
+					// Ensure that CGameHandler::spawnWanderingMonsters won't set a random monster next to another monster
+					// because Nullkiller AI is not able to go to one monster without falling into the attack range of the nearby one
+					// See GraphPaths::addChainInfo if(node.linkDanger > 0) (no link between random monsters and map monsters)
+					if (!nearbyMonsterOrBlocked)
+						tiles.emplace_back(xd, yd, zd);
+				}
```
IvanSavenko: I think this entire block can be replaced with 
```cpp
	if (!guardingCreaturePosition(int3 (xd,yd,zd).isValid())
		tiles.emplace_back(xd, yd, zd);
```

thehonestcto: Thanks for the tip. A lot better since only guarding was the relevant part. Adjusted that and the percentage, only 0.5% more now since there are a ton more tiles unguarded than tiles without anything around.

server/CGameHandler.cpp
```diff
@@ -3995,8 +3995,8 @@ void CGameHandler::spawnWanderingMonsters(CreatureID creatureID)
 {
 	std::vector<int3>::iterator tile;
 	std::vector<int3> tiles;
-	gameState().getFreeTiles(tiles);
-	ui32 amount = tiles.size() / 200; //Chance is 0.5% for each tile
+	gameState().getFreeTiles(tiles, true);
+	ui32 amount = tiles.size() / 100; //Chance is 1% for each tile. Higher than the original 0.5% because there are fewer with nearby unguarded
```
dydzio0614: i do not think original game values should be changed

GeorgeK1ng: True, original game values should be changed by  configurable settings instead.

thehonestcto: Back to 0.5% then. Updated

dydzio0614:
Well - I am not a fan of temporary solutions, but from player point of view monthly spawns not working exactly as intended are less of a problem than AI losing armies, until there is proper fix done

thehonestcto:
> Well - I am not a fan of temporary solutions, but from player point of view monthly spawns not working exactly as intended are less of a problem than AI losing armies, until there is proper fix done

Me neither. 
For the record, this issue becomes even deadlier if you play with "weeklyGrowthPercent" : 25, instead of 10. I saw that with more than 25% Nullkiller becomes unable to fight them after a few months (without any AI unfair boosters). In my sample saved game, the main hero of a color died due to that, so "good bye opponent" due to the bug.

Issue #5993, opened 2025-08-07
Some issues with custom icons and descriptions


misiokles:
Possibility to add custom icons and descriptions for creature windows in vmci-1.7 is very cool idea, but I observed few issues with this system.

Creature bonuses with propagators don't use "icon" and "description" commands.
Example:

```
			"navigationBoost" :
			{
				"type" : "MOVEMENT",
				"subtype" : "heroMovementSea",
				"val" : 300, 
				"propagator" : "HERO",
				"description" : "{Navigator}\nInreases hero sea movement points by 300",
				"icon" : "haven/icons/bonuses/e_whirl.png"				
			},
```

The same with limiters:
Example:
```
			"attackInWater" : 
			{
				"type" : "PRIMARY_SKILL",
				"subtype" : "attack",
				"val" : 1,
				"limiters" : [ {
					"type" : "CREATURE_TERRAIN_LIMITER",
					"terrain" : "water"
				} ],
				"description" : "{Aquarian}\nDuring battles at water gains bonus to attack, defence and speed",
				"icon" : "haven/icons/bonuses/placeholder.png"
			},
```
BUT,
bonus with propagator and limiter is shown correctly.
Example:
```
			"desertRider" : 
			{
				"subtype" : "heroMovementLand",
				"type" : "MOVEMENT",
				"val" : 150,
				"valueType" : "ADDITIVE_VALUE",
				"limiters": [ 
					{
						"type":"CREATURE_TERRAIN_LIMITER",
						"terrain":  "sand"
					}
				],
				"propagator": "HERO",
				"description" : "{Sand Runner}\nBoosts movement by 150 points when staring turn on sandy terrains",
				"icon" : "bonusiconsPavilion/nomads.png"				
			} 
```

"stacking" command makes that every creature in army gain additional icon with stacked bonus.

Example:
```
			"tribute" :
			{
				"type" : "GENERATE_RESOURCE",
				"subtype" : "gold",
				"val" : 50,
				"propagator" : "HERO",
				"stacking" : "base",
				"description" : "{Generate Gold}\nThe presence of any Whipper in hero army produces 50 gold every day",
				"icon" : "bonusiconsPavilion/gold.png"				
			}
```

<img width="497" height="341" alt="Image" src="https://github.com/user-attachments/assets/1cc3a3e9-0966-4db5-a7c2-3a5011a180e7" />

[...]

PR #6010, opened 2025-08-12
Fix abilities propagation

Opuszek:
Solves an issue mentioned in #5993 with descriptions to unit-given hero abilities being propagated to all other units within an army.
It is the best solution I came up with, since, unless I am mistaken, there is no programmatic way to check what a bonus affects (hero, campaign or units). Also, I think an access to a list like this can be quite useful (though the list is arbitrary, unfortunately).
Don't bother checking if the list is correct, if the solution will be accepted I will do it more thoroughly.
PS: I found after I wrote this PR that the same effect can be achieved on develop with a JSON like this:
"sandWalker" :
{
    "type" : "NO_TERRAIN_PENALTY",
    "subtype" : "terrain.sand",
    "propagator" : "HERO",
},
"sandWalkerDescription" :
{
    "type" : "NONE",
    "description" : "Hero ignores terrain penalty on sand"
},
So maybe the PR is not needed at all?

### DIFF
--- a/client/windows/CCreatureWindow.cpp
+++ b/client/windows/CCreatureWindow.cpp
@@ -868,10 +868,10 @@ void CStackWindow::initBonusesList()
 			return  info->stackNode->bonusToString(v1) < info->stackNode->bonusToString(v2);
 	};
 
-	// these bonuses require special handling. For example they come with own descriptions, for use in morale/luck description
-	// also, this information is already available in creature window
-	receivedBonuses.remove_if(Selector::type()(BonusType::MORALE));
-	receivedBonuses.remove_if(Selector::type()(BonusType::LUCK));
+	receivedBonuses.remove_if([](const Bonus* b)
+	{
+		return !LIBRARY->bth->shouldPropagateDescription(b->type);
+	});
 
 	std::vector<BonusList> groupedBonuses;
 	while (!receivedBonuses.empty())
--- a/config/creatures/neutral.json
+++ b/config/creatures/neutral.json
@@ -512,7 +512,8 @@
 			{
 				"type" : "LUCK",
 				"val" : 1,
-				"valueType" : "INDEPENDENT_MAX"
+				"valueType" : "INDEPENDENT_MAX",
+				"description" : "PLACEHOLDER"
 			}
 		 },
 		"graphics" :
@@ -621,7 +622,8 @@
 			{
 				"type" : "NO_TERRAIN_PENALTY",
 				"subtype" : "terrain.sand",
-				"propagator" : "HERO"
+				"propagator" : "HERO",
+				"description" : "PLACEHOLDER"
 			}
 		},
 		"graphics" :
@@ -652,7 +654,8 @@
 				"subtype" : "visionsMonsters",
 				"val" : 3,
 				"valueType" : "INDEPENDENT_MAX",
-				"propagator" : "HERO"
+				"propagator" : "HERO",
+				"description" : "PLACEHOLDER"
 			},
 			"visionsHeroes" :
 			{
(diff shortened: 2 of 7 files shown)


misiokles:
Hm... Interesting walkaround!

client/windows/CCreatureWindow.cpp
```diff
@@ -870,6 +870,11 @@ void CStackWindow::initBonusesList()
 
 	// these bonuses require special handling. For example they come with own descriptions, for use in morale/luck description
 	// also, this information is already available in creature window
+
+	receivedBonuses.remove_if([](const Bonus* b)
+	{
+		return std::find(bonusAffectingCreature.begin(), bonusAffectingCreature.end(), b->type) == bonusAffectingCreature.end();
+	});
```
IvanSavenko: Instead of having list of such bonuses, better approach would be to add new boolean field to `CBonusType`, similar to `creatureNature`/`hidden` flags & add new flag to all such bonuses in `config/bonuses.json` (and if bonus is missing from that file - add bonus itself in there).

Name it `creatureAffecting`, or something like that

config/creatures/neutral.json
```diff
@@ -621,7 +621,8 @@
 			{
 				"type" : "NO_TERRAIN_PENALTY",
 				"subtype" : "terrain.sand",
-				"propagator" : "HERO"
+				"propagator" : "HERO",
+				"description" : "Hero ignores terrain penalty on sand"
```
IvanSavenko: Just in case - can you check if there are any other bonuses that would benefit from this treatment? Search for all creature abilities that have `propagator` set. Some already have custom descriptions (e.g. Ghost Dragons), but some might still be without custom description

config/creatures/neutral.json
```diff
@@ -652,23 +653,26 @@
 				"subtype" : "visionsMonsters",
 				"val" : 3,
 				"valueType" : "INDEPENDENT_MAX",
-				"propagator" : "HERO"
+				"propagator" : "HERO",
+				"description" : "Hero can see detailed informations about neutral monsters."
```
IvanSavenko: For translation convenience, please move actual text to `Mods/vcmi/config/english.json` and only keep some placeholder here.

You can check how Ghost Dragons / Archangels abilities are defined as reference

```suggestion
				"propagator" : "HERO",
				"description" : "PLACEHOLDER"
```

client/windows/CCreatureWindow.cpp
Opuszek: Thanks for the review.
I think I found a better (less messy) solution in the meantime and made a second draft, I will prepare a pr once it will be decided which one should be merged.

Opuszek: Done

config/creatures/neutral.json
Opuszek: Done.

config/creatures/neutral.json
Opuszek: Done.

lib/CBonusTypeHandler.cpp
```diff
@@ -142,6 +142,7 @@ void CBonusTypeHandler::loadItem(const JsonNode & source, CBonusType & dest, con
 	dest.identifier = name;
 	dest.hidden = source["hidden"].Bool(); //Null -> false
 	dest.creatureNature = source["creatureNature"].Bool(); //Null -> false
+	dest.propagateDescription = source["propagateDescription"].isNull() ? true : source["propagateDescription"].Bool(); //Null -> true
```
IvanSavenko: Not against it if you can't think of better name, but in VCMI, default value for boolean fields is usually `false`, partially for consistency, and partially to make code easier to read

Opuszek: I've changed it to blockDescriptionPropagation.
I like the name as it says exactly what the field is used for. Something like "affectsCreatures" can be misleading (e.g increased growth does affects creatures, but we don't want to propagate the description).

Opuszek: I've changed it to blockDescriptionPropagation.
I like the name as it says exactly what the field is used for. Something like "affectsCreatures" can be misleading (e.g increased growth does affects creatures, but we don't want to propagate the description).

PR #6014, opened 2025-08-14
Fixes abilities propagation the second (and better?) solution

Opuszek:
I know it is a second pr for the same issue, but I think I found a much better solution.
Instead of hardcoding everything I changed the code to remove description if it is propagated and targetSourceType was set to "COMMANDER" (e.g nomad's sandwalking) or "GLOBAL" (e.g whipper from new pavilion mod).
After that I deleted code that removes luck and morale bonuses and let json configuration take care of that. After that, bonuses like "lucky" of halfling and "positive luck" of mermaid from Haven mod display normally. 

With that two out of three problems mentioned in #5993 are resolved. There is still a problem with descriptions not displaying when limiters conditions are not fulfilled - for example the desertRider ability from new pavilion mod does not show unless hero stands on a desert tile.

### DIFF
--- a/Mods/vcmi/Content/config/english.json
+++ b/Mods/vcmi/Content/config/english.json
@@ -952,6 +952,12 @@
 	"creatures.core.marksman.bonus.extraAttack" : "{Shoots twice}\nThis unit can shoot twice",
 	"creatures.core.azureDragon.bonus.fearful" : "{Fear}\nEnemy units have a 10% chance of freezing in fear",
 	"creatures.core.azureDragon.bonus.fearless" : "{Fearless}\nImmune to Fear ability",
+	"creatures.core.azureDragon.bonus.fearful" : "{Fear}\nEnemy units have a 10% chance of freezing in fear",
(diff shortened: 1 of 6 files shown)


misiokles:
I've talked with edeksumo recently and he noticed that it's wise to add `"PROPAGATOR" : "HERO"` to the bonuses with `"CREATURE_TERRAIN_LIMITER"` because placing creature without hero propagator crashes map editor (funcionality stays the same).
So custom icons and descriptions should be revised once again!

IvanSavenko:
Personally I'd prefer this solution instead of hacky ones: https://github.com/vcmi/vcmi/pull/6010#discussion_r2275790307

For example, in case of this PR - what makes `COMMANDER` source type so special? Why this source type specifically and not another one? Why we're reusing target source type field for something that is completely unrelated to its original purpose?

IvanSavenko:
@misiokles 
Sounds like a bug. I've fixed a few cases like that one when loading h3m's/vmap's, but probably something similar still exists with editor.

PR #6017, opened 2025-08-15
Fix build with Boost 1.89.0

### LINKED ISSUE #6015: Not compatible with boost 1.89.0
**Describe the bug**
Does not compile with boost 1.89.0 stable.

**Additional context**
We need this patch:
https://github.com/dscharrer/innoextract/pull/199

But VCMI itself must be fixed in a similar way too.

See here: https://github.com/flathub/eu.vcmi.VCMI/issues/215

Laserlicht:
Fixes #6015

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -468,11 +468,15 @@ endif()
 #        Finding packages                  #
 ############################################
 
-set(BOOST_COMPONENTS date_time filesystem locale program_options system)
+set(BOOST_COMPONENTS date_time filesystem locale program_options)
 if(ENABLE_INNOEXTRACT)
 	list(APPEND BOOST_COMPONENTS iostreams)
 endif()
 find_package(Boost 1.48.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
+if(Boost_MAJOR_VERSION EQUAL 1 AND Boost_MINOR_VERSION LESS 69)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -468,11 +468,15 @@ endif()
 #        Finding packages                  #
 ############################################
 
-set(BOOST_COMPONENTS date_time filesystem locale program_options system)
+set(BOOST_COMPONENTS date_time filesystem locale program_options)
 if(ENABLE_INNOEXTRACT)
 	list(APPEND BOOST_COMPONENTS iostreams)
 endif()
 find_package(Boost 1.48.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
+if(Boost_MAJOR_VERSION EQUAL 1 AND Boost_MINOR_VERSION LESS 69) # variables setted after calling find_package with boost
```
kambala-decapitator: the comment looks useless.

but why 69 and not 89?

Laserlicht: good catch! :)

Comment was there because `find_package(Boost 1.48.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})` is copied line, only because the variables. I've removed comment.

kambala-decapitator (approved):

CMakeLists.txt
kambala-decapitator: 69 is correct, see upstream PR

PR #6022, opened 2025-08-16
Keep screen on

### LINKED ISSUE #5995: Do not turn off display during mods downloading / installing
Currently on Android (Android 15 - Galaxy Tab 9 FE) you can encounter broken downloads / mod installing when your display automatically turns off.

Would be great when will be display always on during these actions in Launcher.

Laserlicht:
Fixes #5995 for android.

Keeps screen on in launcher (**not** only for specific actions; should be okay, as game itself also keeps screen on).

### DIFF
--- a/android/vcmi-app/src/main/java/eu/vcmi/vcmi/ActivityLauncher.java
+++ b/android/vcmi-app/src/main/java/eu/vcmi/vcmi/ActivityLauncher.java
@@ -6,6 +6,7 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.os.Environment;
 import android.provider.DocumentsContract;
+import android.view.WindowManager;
 
 import androidx.annotation.Nullable;
 
@@ -31,6 +32,7 @@ public class ActivityLauncher extends org.qtproject.qt5.android.bindings.QtActiv
         super.onCreate(savedInstanceState);
         justLaunched = savedInstanceState == null;
         SDL.setContext(this);
+        getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
     }
 
     @Override


android/vcmi-app/src/main/java/eu/vcmi/vcmi/ActivityLauncher.java
```diff
@@ -31,6 +32,7 @@ public void onCreate(@Nullable final Bundle savedInstanceState)
         super.onCreate(savedInstanceState);
         justLaunched = savedInstanceState == null;
         SDL.setContext(this);
+        getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
```
kambala-decapitator, line 35: it'd make more sense to enable it only when download starts and disable once it finishes

also, doesn't Qt or SDL already have such a function?

Laserlicht: Agree, but have not found a functionality like this.

`FLAG_KEEP_SCREEN_ON` seems only possible on `onCreate`.

Laserlicht: But I think “always” is okay too, since that's how it is in the game itself.

IvanSavenko: SDL is not relevant, since we only need this in Launcher.

Qt might have something, but I presume this is only needed for Android, so this approach is probably OK? What about iOS?

And what if player turns screen off on his own?

Laserlicht: > And what if player turns screen off on his own?

Still possible as before.

IvanSavenko: No, I mean - what if player turns off screen during download? Won't it result in the same corruption?

Laserlicht: > What about iOS?

Not sure if it's an problem there too. Not an apple user.

Laserlicht: > No, I mean - what if player turns off screen during download? Won't it result in the same corruption?

Ah. I assume yes - same corruption. But at least this would be an active operation. Not like automatic screen off (you don't need to press the touch screen every few seconds anymore to keep it on).

But yes. Optimal way would be an background service doing such tasks. But no idea how and motivation to do this.

IvanSavenko: Can we at least detect such corruption & discard or restart download? Or this part is already working fine?

Laserlicht: > Can we at least detect such corruption & discard or restart download? Or this part is already working fine?

In my cases in the past: The mods was just not installed. Stopped while downloading. Just install it again and everything is fine. Not sure what happens if it aborts while extracting.

IvanSavenko: That part should be fine then - mods are extracted into temporary directory, and then directory is moved to correct location. Such operation should be close to instant, or even atomic / single transaction

kambala-decapitator: > SDL is not relevant, since we only need this in Launcher.

SDL is linked to the launcher exe/lib, so it could be used (unless it requires initializing video system first).

> What about iOS?

iOS has the same issue. Once you turn off the screen or minimize an app, all sockets are closed.

kambala-decapitator:
@Laserlicht please make a separate PR for the artifacts fix

PR #6038, opened 2025-08-20
Add working cache for Windows / Android builds

GeorgeK1ng:
This PR highly improve build time in CI.

1) Added working cache for Windows and Android builds
2) Added caching for install components for Ubuntu runners
3) Added initial report summary into actions

Average build time with generating report should be about 13 minutes

image

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -135,6 +135,7 @@ jobs:
             artifact_platform: arm64-v8a
 
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ !startsWith(matrix.platform, 'msvc-') }}
     defaults:
       run:
         shell: bash
@@ -381,17 +382,29 @@ jobs:
       run: |
         # Decide cache tool and family
         if [[ "${{ startsWith(matrix.platform, 'msvc') }}" == "true" ]]; then
-          TOOL="sccache";  CMD="sccache --show-stats"; FAMILY="windows-msvc"
+          TOOL="sccache";
+          CMD="sccache --show-stats";
+          FAMILY="windows-msvc"
         elif [[ "${{ startsWith(matrix.platform, 'mingw_') }}" == "true" ]]; then
-          TOOL="ccache";   CMD="ccache -s";            FAMILY="windows-mingw"
+          TOOL="ccache";
+          CMD="ccache -s";
+          FAMILY="windows-mingw"
         elif [[ "${{ startsWith(matrix.platform, 'mac') }}" == "true" ]]; then
-          TOOL="ccache";   CMD="ccache -s";            FAMILY="macos"
+          TOOL="ccache";
+          CMD="ccache -s";
+          FAMILY="macos"
         elif [[ "${{ matrix.platform }}" == "ios" ]]; then
-          TOOL="ccache";   CMD="ccache -s";            FAMILY="ios"
+          TOOL="ccache";
+          CMD="ccache -s";
+          FAMILY="ios"
         elif [[ "${{ startsWith(matrix.platform, 'android') }}" == "true" ]]; then
-          TOOL="ccache";   CMD="ccache -s";            FAMILY="android"
+          TOOL="ccache";
+          CMD="ccache -s";
+          FAMILY="android"
         else
-          TOOL="ccache";   CMD="ccache -s";            FAMILY="other"
+          TOOL="ccache";
+          CMD="ccache -s";
+          FAMILY="other"
         fi
         
         # Raw stats (never fail)
@@ -399,17 +412,30 @@ jobs:
         
         # Parse hits/misses
         if [[ "$TOOL" == "sccache" ]]; then
-          HITS="$(printf '%s\n' "$STATS" | awk '/^Cache hits/   {print $NF; exit}')"
-          MISS="$(printf '%s\n' "$STATS" | awk '/^Cache misses/ {print $NF; exit}')"
+          HITS=$(awk '/^Cache hits/   { print $NF; exit }'    <<<"$STATS")
+          MISS=$(awk '/^Cache misses/ { print $NF; exit }'    <<<"$STATS")
         else
-          HITS="$(printf '%s\n' "$STATS" | awk '/^[[:space:]]*Hits:/   {print $2; exit}')"
-          MISS="$(printf '%s\n' "$STATS" | awk '/^[[:space:]]*Misses:/ {print $2; exit}')"
+          HITS=$(awk '/^[[:space:]]*Hits:/   { print $2; exit }' <<<"$STATS")
+          MISS=$(awk '/^[[:space:]]*Misses:/ { print $2; exit }' <<<"$STATS")
         fi
-        [[ "$HITS" =~ ^[0-9]+$ ]] || HITS=0
-        [[ "$MISS" =~ ^[0-9]+$ ]] || MISS=0
+        
+        [[ $HITS =~ ^[0-9]+$ ]] || HITS=0
+        [[ $MISS =~ ^[0-9]+$ ]] || MISS=0
+        
         TOTAL=$((HITS+MISS))
+        
         # Two-decimal percent, e.g. 99.36%
-        RATE=$(awk -v h="$HITS" -v t="$TOTAL" 'BEGIN{ if(t>0) printf("%.2f%%",(100.0*h)/t); else print "n/a" }')
+        if (( $TOTAL > 0 )); then
+          RATE=$(
+            LC_ALL=C awk -v h="$HITS" -v t="$TOTAL" 'BEGIN {
+              if (t == 0) { print "n/a"; exit }
+              pct = 100.0 * h / t
+              printf "%.2f%%", pct
+            }'
+          )
+        else
+          RATE="n/a"
+        fi
         
         # Artifact URLs
         MAIN_URL='${{ steps.upload_artifact.outputs.artifact-url }}'
@@ -421,10 +447,10 @@ jobs:
         if [[ -z "$ARCH" ]]; then
           case "${{ matrix.platform }}" in
             mac-intel)     ARCH="Intel";;
-            mac-arm)       ARCH="ARM";;
+            mac-arm)       ARCH="ARM64";;
             ios)           ARCH="ARM64";;
             mingw_x86)     ARCH="x86";;
-            mingw_x86_64)  ARCH="x86_64";;
+            mingw_x86_64)  ARCH="x64";;
             android-32)    ARCH="ARMv7";;
             android-64)    ARCH="ARM64";;
             *)             ARCH="${{ matrix.platform }}";;
@@ -555,13 +581,11 @@ jobs:
       uses: actions/checkout@v5
       with:
         submodules: recursive
-    
-    # User-owned staging dir
+
     - name: Prepare APT staging dir
       if: contains(matrix.os, 'ubuntu')
       run: mkdir -p "$RUNNER_TEMP/apt-cache"
-    
-    # RESTORE (with unique key per matrix.platform)
+
     - name: APT cache restore
       if: contains(matrix.os, 'ubuntu')
       id: aptcache
@@ -574,8 +598,7 @@ jobs:
 
     - name: Prepare CI
       run: source '${{github.workspace}}/CI/before_install/${{matrix.before_install}}'
-    
-    # SAVE only if we didn't hit an existing cache
+
     - name: APT cache save
       if: contains(matrix.os, 'ubuntu') && steps.aptcache.outputs.cache-hit != 'true'
       uses: actions/cache/save@v4
@@ -635,6 +658,7 @@ jobs:
 
   windows-installer:
     needs: build
+    if: ${{ needs.build.result == 'success' }}
     name: Create Windows Installer (${{ matrix.arch }})
     strategy:
       matrix:
@@ -789,7 +813,7 @@ jobs:
     name: Build report
     if: always()
     needs: [build, windows-installer, validate-code, test, upload-source-package]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     permissions:
       contents: read
       actions: write
@@ -805,7 +829,13 @@ jobs:
           merge-multiple: true
           path: partials
   
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+  
       - name: Run final summary
         env:
-          GITHUB_TOKEN: ${{ github.token }}
-        run: python3 CI/final_summary.py
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python CI/final_summary.py


CMakeLists.txt
```diff
@@ -128,14 +128,31 @@ if(ENABLE_COLORIZED_COMPILER_OUTPUT)
...
-	set(CMAKE_CXX_COMPILER_LAUNCHER "ccache")
+    if(MSVC)
+        find_program(SCCACHE sccache)
+        if(SCCACHE)
+            message(STATUS "Enabling compiler launcher: sccache")
+            set(CMAKE_C_COMPILER_LAUNCHER "${SCCACHE}")
+            set(CMAKE_CXX_COMPILER_LAUNCHER "${SCCACHE}")
+        else()
+            # Optional fallback (only if ccache is present)
+            find_program(CCACHE ccache)
```
kambala-decapitator:
```suggestion
            find_program(CCACHE ccache REQUIRED)
```

CMakeLists.txt
```diff
@@ -128,14 +128,31 @@ if(ENABLE_COLORIZED_COMPILER_OUTPUT)
...
+            if(CCACHE)
+                message(STATUS "Enabling compiler launcher: ccache (fallback)")
+                set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE}")
+                set(CMAKE_CXX_COMPILER_LAUNCHER "${CCACHE}")
+            endif()
+        endif()
+    else()
+        find_program(CCACHE ccache)
```
kambala-decapitator:
```suggestion
        find_program(CCACHE ccache REQUIRED)
```

CMakePresets.json
```diff
@@ -203,9 +203,46 @@
             "inherits": "windows-msvc-release",
             "cacheVariables": {
                 "ENABLE_CCACHE": "ON"
-
             }
         },
+        {
+          "name": "windows-msvc-ninja-release",
```
kambala-decapitator, line 209: Why not inheriting it from the normal msvc?

.github/workflows/github.yml
```diff
@@ -151,6 +151,35 @@ jobs:
       if: ${{ startsWith(matrix.platform, 'msvc') }}
       run: source '${{github.workspace}}/CI/install_vcpkg_dependencies.sh' '${{matrix.platform}}'
 
+    - name: Setup Ninja (Windows)
+      if: ${{ startsWith(matrix.platform, 'msvc') }}
+      uses: seanmiddleditch/gha-setup-ninja@v5
```
kambala-decapitator: Runners already have ninja installed, why is this needed?

.github/workflows/github.yml
```diff
@@ -229,14 +258,26 @@ jobs:
...
+          cmake -DENABLE_CCACHE:BOOL=ON -DANDROID_GRADLE_PROPERTIES="applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily;applicationVariant=daily" --preset ${{ matrix.preset }}
+        elif ${{ startsWith(matrix.platform, 'msvc') }}; then
+          if [[ "${{ matrix.arch }}" == "x86" ]]; then
+            CC="$VCToolsInstallDir/bin/Hostx64/x86/cl.exe"
+            cmake \
+              -D CMAKE_C_COMPILER:FILEPATH="$CC" \
+              -D CMAKE_CXX_COMPILER:FILEPATH="$CC" \
+              --preset ${{ matrix.preset }}
+          elif [[ "${{ matrix.arch }}" == "arm64" ]]; then
+            CC="$VCToolsInstallDir/bin/HostARM64/ARM64/cl.exe"
```
IvanSavenko: Maybe move this path to job parameter for all 3 msvc runs? To avoid this if-else chain?

CMakeLists.txt
```diff
@@ -128,14 +128,31 @@ if(ENABLE_COLORIZED_COMPILER_OUTPUT)
...
-	set(CMAKE_C_COMPILER_LAUNCHER "ccache")
-	set(CMAKE_CXX_COMPILER_LAUNCHER "ccache")
+    if(MSVC)
+        find_program(SCCACHE sccache)
+        if(SCCACHE)
+            message(STATUS "Enabling compiler launcher: sccache")
+            set(CMAKE_C_COMPILER_LAUNCHER "${SCCACHE}")
+            set(CMAKE_CXX_COMPILER_LAUNCHER "${SCCACHE}")
+        else()
+            # Optional fallback (only if ccache is present)
```
IvanSavenko, line 140: Does this fallback actually works?
If not - why keep broken code?

.github/workflows/github.yml
```diff
@@ -151,6 +151,35 @@ jobs:
...
+        arch: ${{ matrix.arch }}
+
+    - name: Setup sccache (Windows/MSVC)
+      if: ${{ startsWith(matrix.platform, 'msvc') }}
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        variant: sccache
+        key: sccache-${{ matrix.platform }}-${{ github.ref_name }}
+        restore-keys: |
+          sccache-${{ matrix.platform }}-${{ github.ref_name }}
```
IvanSavenko: Perhaps use same naming scheme as non-sccache step? One that wont cause conflicts when we have both develop and beta branches?

GeorgeK1ng: Yes, I will PR more changes soon. Just few more touches needs to be done.

CMakePresets.json
GeorgeK1ng: It doesn't support architecture block which is present in normal msvc config.

.github/workflows/github.yml
GeorgeK1ng: Not needed, removed

.github/workflows/github.yml
GeorgeK1ng: Moved to matrix parameter

CMakeLists.txt
GeorgeK1ng: It should still work in local environment

CMakePresets.json
kambala-decapitator: you could refactor normal msvc by introducing a base config without architecture block

.github/workflows/github.yml
```diff
@@ -13,6 +13,7 @@ jobs:
   build:
     name: Build (${{ matrix.platform }})
     strategy:
+      fail-fast: false
```
kambala-decapitator, line 16: don't forget to remove

CMakePresets.json
GeorgeK1ng: We can polish it later in future PR. I'm currently happy everything is cached except downloading vcpkg, can save another 1 min in build time. But it's not crucial. A lot of other important things were touched right now.

.github/workflows/github.yml
```diff
@@ -229,14 +283,20 @@ jobs:
...
+          cmake -DENABLE_CCACHE:BOOL=ON \
+                -DCMAKE_C_COMPILER_LAUNCHER=ccache \
+                -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+                -DANDROID_GRADLE_PROPERTIES="applicationIdSuffix=.daily;signingConfig=dailySigning;applicationLabel=VCMI daily;applicationVariant=daily" \
+                --preset ${{ matrix.preset }}
+        elif ${{ startsWith(matrix.platform, 'msvc') }}; then
+            CL="$VCToolsInstallDir/bin/${{ matrix.cl }}"
+            cmake \
+              -D CMAKE_C_COMPILER:FILEPATH="$CL" \
+              -D CMAKE_CXX_COMPILER:FILEPATH="$CL" \
```
kambala-decapitator, line 297: why is this required? Can't CMake find proper compiler based on the architecture variant requested?

.github/workflows/github.yml
```diff
@@ -312,6 +375,92 @@ jobs:
         DEPLOY_RSA: ${{ secrets.DEPLOY_RSA }}
         PACKAGE_EXTENSION: ${{ matrix.extension }}
 
+    - name: Prepare partial JSON with build informations
+      id: make_partial_json
+      shell: bash
+      run: |
+        # Decide cache tool and family
+        if [[ "${{ startsWith(matrix.platform, 'msvc') }}" == "true" ]]; then
+          TOOL="sccache";  CMD="sccache --show-stats"; FAMILY="windows-msvc"
```
kambala-decapitator: why throwing everything onto a single line?

.github/workflows/github.yml
```diff
@@ -312,6 +375,92 @@ jobs:
...
+        # Raw stats (never fail)
+        STATS="$($CMD || true)"
+        
+        # Parse hits/misses
+        if [[ "$TOOL" == "sccache" ]]; then
+          HITS="$(printf '%s\n' "$STATS" | awk '/^Cache hits/   {print $NF; exit}')"
+          MISS="$(printf '%s\n' "$STATS" | awk '/^Cache misses/ {print $NF; exit}')"
+        else
+          HITS="$(printf '%s\n' "$STATS" | awk '/^[[:space:]]*Hits:/   {print $2; exit}')"
+          MISS="$(printf '%s\n' "$STATS" | awk '/^[[:space:]]*Misses:/ {print $2; exit}')"
```
kambala-decapitator: 1. why printf instead of echo? You could also use bash's `<<<` operator to pass string into stdin.
2. what is awk doing? is it just a grep?

.github/workflows/github.yml
```diff
@@ -312,6 +375,92 @@ jobs:
...
+        fi
+        [[ "$HITS" =~ ^[0-9]+$ ]] || HITS=0
+        [[ "$MISS" =~ ^[0-9]+$ ]] || MISS=0
```
kambala-decapitator: can these vars be anything besides digits and empty?

.github/workflows/github.yml
```diff
@@ -312,6 +375,92 @@ jobs:
...
+        TOTAL=$((HITS+MISS))
+        # Two-decimal percent, e.g. 99.36%
+        RATE=$(awk -v h="$HITS" -v t="$TOTAL" 'BEGIN{ if(t>0) printf("%.2f%%",(100.0*h)/t); else print "n/a" }')
```
kambala-decapitator: looks quite monstrous to me, can be replaced with a much easier to understand construct:
```suggestion
        if (( $TOTAL > 0 )) ; then
          # I think awk would also do here, might be even better
          percentage=$(bc -e "scale=2; (100 * $HITS) / $TOTAL")
          RATE="${percentage}%"
        else
          RATE="n/a"
        fi
```

.github/workflows/github.yml
```diff
@@ -312,6 +375,92 @@ jobs:
...
+        MAIN_URL='${{ steps.upload_artifact.outputs.artifact-url }}'
+        DEBUG_URL='${{ steps.upload_symbols.outputs.artifact-url }}'
+        AAB_URL='${{ steps.upload_aab.outputs.artifact-url }}'
+        
+        # Arch label
+        ARCH="${{ matrix.arch }}"
+        if [[ -z "$ARCH" ]]; then
+          case "${{ matrix.platform }}" in
+            mac-intel)     ARCH="Intel";;
+            mac-arm)       ARCH="ARM";;
```
```suggestion
            mac-arm)       ARCH="ARM64";;
```

.github/workflows/github.yml
```diff
@@ -312,6 +375,92 @@ jobs:
...
+            android-32)    ARCH="ARMv7";;
+            android-64)    ARCH="ARM64";;
+            *)             ARCH="${{ matrix.platform }}";;
+          esac
+        fi
+        
+        # Pick Python (works on all runners)
+        PY=python3; command -v "$PY" >/dev/null 2>&1 || PY=python
+        # JSON-escape the raw stats
+        STATS_ESC=$("$PY" -c 'import json,sys; print(json.dumps(sys.stdin.read()))' <<<"$STATS")
```
kambala-decapitator, line 463: is there any advantage of passing from stdin instead of simple argument (argv)?

.github/workflows/github.yml
```diff
@@ -312,6 +375,92 @@ jobs:
...
+        
+        mkdir -p .summary
+        cat > .summary/${{ matrix.platform }}.json <<EOF
```
kambala-decapitator, line 466: since end goal is to form a JSON, feels like writing this step in python would have been much easier and shorter

.github/workflows/github.yml
```diff
@@ -578,3 +784,340 @@ jobs:
           with:
             config: 'CI/example.markdownlint-cli2.jsonc'
             globs: '**/*.md'
+
+  final-summary:
+    name: Build report
```
kambala-decapitator, line 813: you must be joking.
- write it in python, not crazy bash with jq
- move to a separate file

.github/workflows/github.yml
```diff
@@ -578,3 +784,340 @@ jobs:
...
+      - name: Download all partial JSON artifacts
+        continue-on-error: true
+        uses: actions/download-artifact@v5
+        with:
+          pattern: partial-json-*
+          merge-multiple: true
+          path: partials
+
+      - name: Collect validation & test results
+        uses: actions/github-script@v7
```
kambala-decapitator: what is it doing in a nutshell? is it really required to introduce JS now? would python (or bash) + `gh` CLI simplify things?

CI/before_install/linux_qt5.sh
```diff
@@ -1,13 +1,27 @@
...
+APT_CACHE="${APT_CACHE:-${RUNNER_TEMP:-/tmp}/apt-cache}"
+sudo mkdir -p "$APT_CACHE"
+
+sudo apt-get -yq -o Acquire::Retries=3 update
+sudo apt-get -yq install eatmydata
+
+sudo eatmydata apt-get -yq --no-install-recommends \
+  -o Dir::Cache::archives="$APT_CACHE" \
+  -o APT::Keep-Downloaded-Packages=true \
+  -o Acquire::Retries=3 -o Dpkg::Use-Pty=0 \
```
kambala-decapitator, line 15: why adding all these options? Were there issues without them?

also, use `apt` instead of `apt-get`

CMakeLists.txt
```diff
@@ -302,18 +319,20 @@ if(MINGW OR MSVC)
...
-			set(CMAKE_VS_GLOBALS
-				"CLToolExe=cl.exe"
-				"CLToolPath=${CMAKE_BINARY_DIR}"
-				"TrackFileAccess=false"
-				"UseMultiToolTask=true"
-			)
+		if(ENABLE_CCACHE AND (CMAKE_GENERATOR MATCHES "Visual Studio"))
+		    # Wrapper hack only for VS generator; NOT used with Ninja+sccache
+		    if(CCACHE)
+		        file(COPY_FILE
```
kambala-decapitator, line 325: symlink would be better, IIRC it also has an option to fallback to copy

CMakeLists.txt
```diff
@@ -810,6 +829,7 @@ endif(WIN32)
 #       Packaging section             #
 #######################################
 
+# This needs to be removed in future
```
kambala-decapitator, line 832: at which point exactly? why is this no longer required?

CMakeLists.txt
```diff
@@ -838,6 +858,7 @@ endif()
 set(CPACK_PACKAGE_VENDOR "VCMI team")
 
 if(WIN32)
+	# This needs to be removed in future
```
kambala-decapitator, line 861: same question

kambala-decapitator (changes requested):

.github/workflows/github.yml
GeorgeK1ng: Exactly, without this you can't build it. Because mismatches x64 vs x86 etc. have not found better way to compile.

.github/workflows/github.yml
GeorgeK1ng: Currently it's not used in report. I have only keep it. it's single line in json with \n. sccache and ccache have different outputs. So easier was to put it into single line. Maybe replace \n with <br> will be better later when adding detailed output to Summayr build report.

kambala-decapitator:
I suggest to extract the beautiful stats page into a separate PR

CMakeLists.txt
GeorgeK1ng: When OpenSign can sign our libraries, whole CPACK must be dropped for Windows builds. It's not needed anymore. We don't need to ZIP anything as GitHub action will ZIP too. And be don't need to resign ms redist + ucrt with another certificate - it will lead to AV warnings.

kambala-decapitator: please update comment to reflect that as currently it's unclear

kambala-decapitator: or simply remove the comment

.github/workflows/github.yml
kambala-decapitator: no, I mean why not writing
```suggestion
          TOOL="sccache"
          CMD="sccache --show-stats"
          FAMILY="windows-msvc"
```

currently I don't see any advantage of having all 3 vars on a single line

.github/workflows/github.yml
GeorgeK1ng: I have added it to not cancel all jobs when single build fails.

CMakeLists.txt
GeorgeK1ng: Will update both comments, I have used it to mark required touches, to not forget them.

GeorgeK1ng:
> I suggest to extract the beautiful stats page into a separate PR

Good idea. Any tip, why mingw x86 build fails now?

CI/before_install/linux_qt5.sh
GeorgeK1ng: They are used to speed up whole process with cache. They are part of package caching steps.
Isn't recommended to use apt-get for scripts?

.github/workflows/github.yml
GeorgeK1ng: Done, forget to fix that typo

.github/workflows/github.yml
GeorgeK1ng: Done, have been already working on that.

.github/workflows/github.yml
GeorgeK1ng: Removed, not needed anymore as it's moved to PY.

.github/workflows/github.yml
GeorgeK1ng: Done

.github/workflows/github.yml
GeorgeK1ng: Probably not

.github/workflows/github.yml
GeorgeK1ng: Simplified as suggested

.github/workflows/github.yml
GeorgeK1ng: Simplified as suggested

.github/workflows/github.yml
kambala-decapitator: it was like that intentionally

CI/before_install/linux_qt5.sh
kambala-decapitator: apt-get is useful only when you want to parse its output

CMakeLists.txt
kambala-decapitator: if the OpenSign PR already removes those sections, no need to leave comments

kambala-decapitator:
> Any tip, why mingw x86 build fails now?

no idea. at least [here](https://github.com/vcmi/vcmi/actions/runs/17156823052/job/48675604992) it works, so it must be something about your changes.

.github/workflows/github.yml
kambala-decapitator: seems that because of this change Action run is shown as green, although there's a failing mingw job. In PR the failure is visible though.

GeorgeK1ng: Thats because another change to keep installer job running. In my branch it works just fine. Will investigate it in new branch and PR to see differencies.

GeorgeK1ng: Still thinking about platform workflow splits

CMakeLists.txt
GeorgeK1ng: Done

CI/before_install/linux_qt5.sh
GeorgeK1ng: Done

.github/workflows/github.yml
GeorgeK1ng: Done

PR #6040, opened 2025-08-20
move commander start position to 94 to avoid overlap

### LINKED ISSUE #6039: Crash Triggered by 2-Hex Creature with Attack-and-Return in Bank Battles
**Describe the bug**

This crash requires three conditions to be triggered:

    1. Attacking a bank, where the commander and a 2-hex creature overlap.
    2. The 2-hex creature has the "attack and return" ability.
    3. The creature attacks an enemy located on the left edge of its movement range.

Because the creature is unable to return to its original position, the crash occurs. See the figure below.

Image

**Version**

- OS: All
- Version: 1.7 develop branch

kdmcser:
fixed #6039

Placing the commander on hex 95 will cause it to overlap with a double-hex creature on hex 96, leading to a series of bugs. Moving it to hex 94 can avoid this issue.

### DIFF
--- a/config/gameConfig.json
+++ b/config/gameConfig.json
@@ -655,7 +655,7 @@
 				"creatureBankNarrow" : {
 					"tacticsAllowed" : false,
 					"obstaclesAllowed" : false,
-					"attackerCommander" : 95,
+					"attackerCommander" : 94,
 					"defenderCommander" : 8,
 					"attackerUnits": [ 57, 61, 90, 93, 96, 125, 129 ],
 					"defenderUnits": [ 15, 185, 172, 2, 100, 87, 8 ]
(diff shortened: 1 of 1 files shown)


dydzio0614 (approved):

config/gameConfig.json
```diff
@@ -655,7 +655,7 @@
 				"creatureBankNarrow" : {
 					"tacticsAllowed" : false,
 					"obstaclesAllowed" : false,
-					"attackerCommander" : 95,
+					"attackerCommander" : 94,
 					"defenderCommander" : 8,
 					"attackerUnits": [ 57, 61, 90, 93, 96, 125, 129 ],
```
IvanSavenko, line 660: I see that there is another unit at 93.
Wouldn't moving commander hex lead to overlap with that unit instead?

kdmcser: I tested it, and a double-hex creature placed on hex 93 actually occupies hexes 92 and 93.

PR #6047, opened 2025-08-22
[macOS] update app icon and generation script

kambala-decapitator:
the small icon version should no longer be "distorted" in Finder

Screenshot 2025-08-22 at 16 38 08

New script usage:
```sh
cd osx
bash generate_icns.sh
```


kambala-decapitator:
checked the artifact: the icon is correct now

PR #6055, opened 2025-08-23
Add working cache for Windows / Android

GeorgeK1ng:
This PR highly improve build time in CI from average 50 minutes to 14 minutes.

1) Added working cache for Windows and Android builds
2) Added caching for install components for Ubuntu runners
3) Added initial report summary into actions
4) Fixes: https://github.com/vcmi/vcmi/issues/5195

First run:
image

Second run:
image

### DIFF
--- a/.github/workflows/github.yml
+++ b/.github/workflows/github.yml
@@ -13,6 +13,7 @@ jobs:
   build:
     name: Build (${{ matrix.platform }})
     strategy:
+      fail-fast: false
       matrix:
         include:
           - platform: mac-intel
@@ -61,8 +62,9 @@ jobs:
             pack_type: RelWithDebInfo
             extension: zip
             before_install: msvc.sh
-            preset: windows-msvc-release
+            preset: windows-msvc-ninja-release
             artifact_platform: x64
+            cl: Hostx64/x64/cl.exe
 
           - platform: msvc-x86
             arch: x86
@@ -72,8 +74,9 @@ jobs:
             pack_type: RelWithDebInfo
             extension: zip
             before_install: msvc.sh
-            preset: windows-msvc-release-x86
+            preset: windows-msvc-ninja-release-x86
             artifact_platform: x86
+            cl: Hostx64/x86/cl.exe
 
           - platform: msvc-arm64
             arch: arm64
@@ -83,10 +86,12 @@ jobs:
             pack_type: RelWithDebInfo
             extension: zip
             before_install: msvc.sh
-            preset: windows-msvc-release-arm64
+            preset: windows-msvc-ninja-release-arm64
             artifact_platform: arm64
+            cl: HostARM64/ARM64/cl.exe
 
           - platform: mingw_x86_64
+            arch: x86_64
             os: ubuntu-24.04
             pack: 1
             pack_type: Release
@@ -98,6 +103,7 @@ jobs:
             conan_prebuilts: dependencies-mingw-x86-64
 
           - platform: mingw_x86
+            arch: x86
             os: ubuntu-24.04
             pack: 1
             pack_type: Release
@@ -129,6 +135,7 @@ jobs:
             artifact_platform: arm64-v8a
 
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ !startsWith(matrix.platform, 'msvc-') }}
     defaults:
       run:
         shell: bash
@@ -138,10 +145,32 @@ jobs:
       uses: actions/checkout@v5
       with:
         submodules: recursive
+    
+    - name: Prepare APT staging dir
+      if: contains(matrix.os, 'ubuntu')
+      run: mkdir -p "$RUNNER_TEMP/apt-cache"
+    
+    - name: APT cache restore
+      if: contains(matrix.os, 'ubuntu')
+      id: aptcache
+      uses: actions/cache/restore@v4
+      with:
+        path: ${{ runner.temp }}/apt-cache
+        key: ${{ matrix.platform }}-apt-${{ matrix.os }}
+        restore-keys: |
+          ${{ matrix.platform }}-apt-${{ matrix.os }}
 
     - name: Prepare CI
       if: "${{ matrix.before_install != '' }}"
-      run: source '${{github.workspace}}/CI/before_install/${{matrix.before_install}}'
+      run: source '${{github.workspace}}/CI/before_install/${{matrix.before_install}}' '${{matrix.arch}}'
+    
+    # SAVE only if we didn't hit an existing cache
+    - name: APT cache save
+      if: contains(matrix.os, 'ubuntu') && steps.aptcache.outputs.cache-hit != 'true'
+      uses: actions/cache/save@v4
+      with:
+        path: ${{ runner.temp }}/apt-cache
+        key: ${{ steps.aptcache.outputs.cache-primary-key }}
 
     - name: Install Conan Dependencies
       if: "${{ matrix.conan_prebuilts != '' }}"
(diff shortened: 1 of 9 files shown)


kambala-decapitator:
Replacement or an addition to that PR?

GeorgeK1ng:
Replacement, when second run success for all platforms. I hope it will now build everything. 

Then I will check Portmaster building. From some reason I was not able to build current version. But I tried it few weeks back

kambala-decapitator:
sorry, not sure why not simply updating the original PR

GeorgeK1ng:
Too much touches and MinGW build fails there. I think it have something to do with it's cache which was created during my updates. Probably it's broken for that PR. As it works fine in my testing branch, it must work here too.

GeorgeK1ng:
Works now as expected including Build summary report which is now PY based instead.

GeorgeK1ng:
@IvanSavenko Anything else to here? Also PRs should have their first compiling faster as it should use branch cache for hits when no PR cache exist.

.github/workflows/github.yml
```diff
@@ -13,6 +13,7 @@ jobs:
   build:
     name: Build (${{ matrix.platform }})
     strategy:
+      fail-fast: false
```
IvanSavenko: Leftovers from testing?

.github/workflows/github.yml
```diff
@@ -129,6 +135,7 @@ jobs:
             artifact_platform: arm64-v8a
 
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ !startsWith(matrix.platform, 'msvc-') }}
```
IvanSavenko: Leftovers from testing?
Or this is needed in some cases? If yes then comment would be useful

.github/workflows/github.yml
```diff
@@ -138,41 +145,87 @@ jobs:
...
+        key: ${{ matrix.platform }}-apt-${{ matrix.os }}
+        restore-keys: |
+          ${{ matrix.platform }}-apt-${{ matrix.os }}
 
     - name: Prepare CI
       if: "${{ matrix.before_install != '' }}"
-      run: source '${{github.workspace}}/CI/before_install/${{matrix.before_install}}'
+      run: source '${{github.workspace}}/CI/before_install/${{matrix.before_install}}' '${{matrix.arch}}'
+    
+    # SAVE only if we didn't hit an existing cache
```
IvanSavenko: Why? Shouldn't we refresh the cache?

.github/workflows/github.yml
```diff
@@ -276,23 +335,26 @@ jobs:
         mv "$OUTPUT_DIRECTORY/bundle/release/vcmi-release.aab" "${{github.workspace}}/out/build/${{matrix.preset}}/${{ env.VCMI_PACKAGE_FILE_NAME }}.aab"
 
     - name: Upload Artifact
+      id: upload_artifact
       uses: actions/upload-artifact@v4
       with:
         name: ${{ env.VCMI_PACKAGE_FILE_NAME }} - ${{ matrix.platform }}
-        compression-level: 0
+        compression-level: 9
```
IvanSavenko, line 345: Why? Idea was to avoid wasting time on compression of already compressed data. Or it can actually improve zipping/artifact upload time?

.github/workflows/github.yml
```diff
@@ -337,17 +416,35 @@ jobs:
             gzip release.tar
             
         - name: Upload source code archive
+          id: upload_source
           uses: actions/upload-artifact@v4
           with:
             name: ${{ env.VCMI_PACKAGE_FILE_NAME }}
-            compression-level: 0
+            compression-level: 9
```
IvanSavenko, line 426: Same here. Tar.gz is already compressed and with the same algorithm

.github/workflows/github.yml
```diff
@@ -560,7 +695,6 @@ jobs:
         shell: bash
     steps:
         - uses: actions/checkout@v5
-
```
IvanSavenko, line 563: Might be better to keep empty line before steps for readability

.github/workflows/github.yml
```diff
@@ -578,3 +712,29 @@ jobs:
           with:
             config: 'CI/example.markdownlint-cli2.jsonc'
             globs: '**/*.md'
+
+  final-summary:
+    name: Build report
+    if: always()
+    needs: [build, windows-installer, validate-code, test, upload-source-package]
```
IvanSavenko, line 723: Do we actually need to wait  test & validate steps ?

IvanSavenko (approved):
Generally looks ok for me,  just some minor tweaks/questions

.github/workflows/github.yml
GeorgeK1ng: All test are now done in few minutes. They were always delayed due installing components which are now cached too. Windows builds are ones which takes long 10-12 minutes. So no delays there.

.github/workflows/github.yml
GeorgeK1ng: I have added it to always compile other builds, currently it's quick, so doesn't matter if one build fails. Others can be completed.

.github/workflows/github.yml
GeorgeK1ng: It's here to always continue to building Windows Installer when other builds fails. It wont go to Windows Installer step only when MSVC builds fail. Can add this comment there at all.

.github/workflows/github.yml
GeorgeK1ng: I just wanted to keep compression same for all steps. Have not noticed any delay in compression.

.github/workflows/github.yml
GeorgeK1ng: Maybe remove tar.gz step there to avoid archive in archive? Result should be almost similar.

.github/workflows/github.yml
GeorgeK1ng: Fixed, accidentally removed when clearing empty spaces

.github/workflows/github.yml
GeorgeK1ng: I quess we don't need to refresh components cache each run. Will always same, not like cache for code.

.github/workflows/github.yml
IvanSavenko: No. Tar.gz (or tar.whatever) is default format for source tarballs. And .zip is enforced by Github.
When publishing release I extract .zip's of each artifact and add its content to release.
Heck, we now have zip in zip! Cmake generates .zip installer for Windows and Github packs it into another zip when uploading artifact.

But let's keep it as it for consistency.

GeorgeK1ng: Yes, ZIP in ZIP will needs to be solved in future for OpenSign. Already added corresponding comments to CMakeLists.
This can be done in future CI adjustments.

.github/workflows/github.yml
GeorgeK1ng: Added better comment

GeorgeK1ng:
Added required comments to make it more clear.
EDIT: Seems caches were removed, will rerun it again when finishes this run.

kambala-decapitator:
will review today

GeorgeK1ng:
Great, my next PR will require touching Windows Installer step with new compiling step for "Modern Folder Picker" x86 plugin for Inno Setup to use modern browse for folder dialog.

.github/workflows/github.yml
```diff
@@ -129,6 +135,10 @@ jobs:
             artifact_platform: arm64-v8a
 
     runs-on: ${{ matrix.os }}
+    # Allow non-MSVC builds to fail without failing whole job
+    # This keeps pipeline moving co Windows Installer job can still run
```
```suggestion
    # This keeps pipeline moving so Windows Installer job can still run
```

.github/workflows/github.yml
```diff
@@ -138,41 +148,87 @@ jobs:
...
           ${{ matrix.platform }}-branch-${{ github.ref_name }}
-        max-size: "5G"
+          ${{ matrix.platform }}-
+        create-symlink: ${{ !startsWith(matrix.platform, 'msvc') }}
+        max-size: '5G'
         verbose: 2
+        job-summary: ""           # <-- disable built-in summary to avoid duplicate block
+        
+    - name: CCache tuning (Android)
+      if: ${{ startsWith(matrix.platform, 'android') }}
```
kambala-decapitator, line 225: would it make sense to enable this step for anything non-msvc?

kambala-decapitator (approved):

.github/workflows/github.yml
GeorgeK1ng: Probably won't change anything, but have not tested it.

GeorgeK1ng:
Can we merge this cache fix / improvements please? @kambala-decapitator can then update conan v2 touches into quicker CI.

kambala-decapitator:
I'm not blocked by this, let's see if @IvanSavenko has any remarks

PR #6057, opened 2025-08-24
Fixes build warning in Potential Targets

Opuszek:
As per title.


AI/BattleAI/PotentialTargets.cpp
```diff
@@ -23,7 +23,7 @@ PotentialTargets::PotentialTargets(
 
 	//FIXME: this should part of battleGetAvailableHexes
 	bool isBerserk = attackerInfo->hasBonusOfType(BonusType::ATTACKS_NEAREST_CREATURE);
-	ForcedAction forcedAction;
+	ForcedAction forcedAction = {};
```
kambala-decapitator, line 26: FYI this change isn't necessary

Opuszek: I wanted to be explicit here just in case.

PR #6060, opened 2025-08-25
Add automated Qt translation workflow

GeorgeK1ng:
This PR adds worflow to automatically update Qt *.ts files for Launcher and Map editor. There will be no need to do it manually as repo will do it automatically every Sunday 00:00 UTC.

### DIFF
--- a/.github/workflows/qt_translation_update.yml
+++ b/.github/workflows/qt_translation_update.yml
@@ -28,12 +28,12 @@ jobs:
       - name: Update TS files (Launcher)
         working-directory: launcher
         run: |
-          lupdate -no-obsolete . -ts translation/*.ts
+          lupdate . -ts translation/*.ts
 
       - name: Update TS files (Map Editor)
         working-directory: mapeditor
         run: |
-          lupdate -no-obsolete . -ts translation/*.ts
+          lupdate . -ts translation/*.ts
 
       - name: Commit and push translation updates
         uses: EndBug/add-and-commit@v9


XCOM-HUB:
This reminds me that I have to correct the translation that was done by AI 2 months ago (Commit 7d017e3). I guess that is necessary...right!?

GeorgeK1ng:
Of course files must be translated. AI translated strings should be still marked as untranslated in Qt translation tool. This simply automatically adds new or remove obsolete strings.

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+on:
+  schedule:
+    - cron: '0 0 * * 0'   # Runs every Sunday at 00:00 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+env:
+  CURRENT_BRANCH: ${{ github.ref_name }}
```
kambala-decapitator: and which branch is that? the default one (develop)?

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+  refresh:
+    name: Qt translations refresh
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ env.CURRENT_BRANCH }}
+          fetch-depth: 0
```
kambala-decapitator, line 21: you could also disable submodules fetching

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Install Qt tools (lupdate)
+        run: |
+          sudo apt install -y --no-install-recommends qttools5-dev-tools
+
+        # Update all TS files; remove obsolete entries
```
kambala-decapitator: the comment looks useless

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: |
+          lupdate -no-obsolete . -ts translation/*.ts
+
+      - name: Show changed files
+        run: |
+          echo "Changed files (working tree):"
+          git status --porcelain || true
+          echo
+          echo "Diff names (unstaged):"
+          git diff --name-only || true
```
kambala-decapitator: not needed

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Stage translation changes explicitly
+        run: |
+          git add -A launcher/translation mapeditor/translation || true
```
kambala-decapitator: I think failing would be ok

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+          echo "Staged for commit:"
+          git diff --cached --name-only || true
```
kambala-decapitator: not needed

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: Detect staged changes
+        id: staged
+        run: |
+          COUNT=$(git diff --cached --name-only | wc -l)
+          echo "count=$COUNT" >> "$GITHUB_OUTPUT"
+          if [ "$COUNT" -eq 0 ]; then
+            echo "No staged translation changes."
+          else
+            echo "Staged $COUNT file(s)."
+          fi
```
kambala-decapitator: not needed

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,77 @@
...
+
+      - name: Commit and push changes
+        if: ${{ steps.staged.outputs.count != '0' }}
```
kambala-decapitator: I think that after omitting the condition, the step will be no-op when nothing changed, please check that

.github/workflows/qt_translation_update.yml
GeorgeK1ng: It's branch it which it runs. In this case Develop.

kambala-decapitator: I understand that :) But no explicit branch is specified in the workflow, that's why I wanted to clarify if it's the "default repo's branch".

GeorgeK1ng:
Updated with simplified workflow

kambala-decapitator (approved):
have you tested this approach in some test workflow to verify that it works as expected? I mean the scheduled run, commit+push some files and case when there's nothing to push.

GeorgeK1ng:
Not scheduled one, but manually launched correctly write modded files. Next run when nothing was changed bot wrote nothing.

1. run
image

2. run
image

.github/workflows/qt_translation_update.yml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Update TS files (Launcher)
+        working-directory: launcher
+        run: |
+          lupdate -no-obsolete . -ts translation/*.ts
+
+      - name: Update TS files (Map Editor)
+        working-directory: mapeditor
+        run: |
+          lupdate -no-obsolete . -ts translation/*.ts
```
IvanSavenko: Not sure about "no obsolete". Perhaps keep those strings for now?

For example, right now fixing a typo in English string would invalidate that string in all other languages.

But, since obsolete string still exists, qt linguist will suggest old (and likely correct) translation automatically.

I also suspect that obsolete strings are not part of final built artifact, so keeping them won't result in binary bloat

GeorgeK1ng: I have removed -no-obsolete. Should be safe now.

PR #6075, opened 2025-08-28
Fix mod update

GeorgeK1ng:


GeorgeK1ng:
Confirmed, mods updating now works as expected. Same for uninstall, no more selection reload.

Issue #6080, opened 2025-08-30
CBattleInfoCallback Atempt to attack from invalid position!


thehonestcto:
**Describe the bug**
There's an issue reported by GrafDeVan with AttackableTiles CBattleInfoCallback::getPotentiallyAttackableHexes triggering throw std::runtime_error("Atempt to attack from invalid position!");
After talking with Ivan and Dydzio, it was agreed to introduce a temporary workaround so we avoind game crashes, but I'm making this report so we can store the saved game with which we can reproduce the issue.

**Game logs**
[VCMI_Client_log attempt to attack from invalid position.txt](https://github.com/user-attachments/files/22058626/VCMI_Client_log.attempt.to.attack.from.invalid.position.txt)

**To Reproduce**
[20250930 NewCrash.vsgm1.zip](https://github.com/user-attachments/files/22058602/20250930.NewCrash.vsgm1.zip)

1. Download and start the saved game
2. End your turn so AI can start his.
3. AI goes into battle and triggers this error which crashes the game.

**Expected behavior**
Game shouldn't crash and battle should be valid.

**Actual behavior**
Game crashes

**Did it work earlier?**
Don't know when and if, didn't test it.

**Screenshots**
<img width="1104" height="1225" alt="Image" src="https://github.com/user-attachments/assets/8e05cb9e-6148-45c6-b038-42e3c8a045d0" />
<img width="1695" height="613" alt="Image" src="https://github.com/user-attachments/assets/05074c7d-ceac-469d-8ae8-9a3cb3e09aba" />

- OS: [Linux Mint @ Ubuntu 22.04 LTS]
- Version: [VCMI 1.7.0.3aedb442809f503dfc227b22a0c636f6a68dc18e (feature/nullkiller2 which has the latest develop as of today)]

**Additional context**
Only VCMI Extras mod is installed and active.

Opuszek:
I am pretty sure that https://github.com/vcmi/vcmi/pull/6016 fixes this issue, let me confirm.

thehonestcto:
Even better. 
@Opuszek, please try the attached saved game, that should confirm or not if it covers this scenario.

Opuszek:
@thehonestcto 
Sorry, I cannot confirm anything. I run the attached save multiple time on the develop branch, and it did not crash a single time, so I cannot check if anything changes.
I don't have right now time to look into this deeper.

thehonestcto:
[...]

PR #6081, opened 2025-08-30
Temporary fix for #6080 until the proper solution. Attempt to attack from invalid position!

thehonestcto:
See issue: https://github.com/vcmi/vcmi/issues/6080

We're returning an empty response until the proper solution will be implemented, as agreed with Ivan and Dydzio.


IvanSavenko:
Closing, proper fix is in

PR #6084, opened 2025-08-30
enable template editor on macos

Laserlicht:


kambala-decapitator (approved):
why was it disabled? :)

Laserlicht:
> why was it disabled? :)

Missing SVG libraries for QT. :)

PR #6087, opened 2025-08-31
fix android url encoding

### LINKED ISSUE #5945: Gog files extraction failure on Android
**Describe the bug**
When I try to install the Heroes data from a gog.com download, I get this error:

Image

**Game logs**

```
[2025-Jul-24 11:30:29.834185] INFO global [7bacccf500] - Extracting gog data from 'content://com.android.externalstorage.documents/document/primary%3ADownload%2FHeroes of Might and Magic 3 Complete%2Fdata.bin' and 'content://com.android.externalstorage.documents/document/primary%3ADownload%2FHeroes of Might and Magic 3 Complete%2Fsetup.exe'
[2025-Jul-24 11:30:29.845102] INFO global [7bacccf500] - Using '/storage/emulated/0/Android/data/is.xyz.vcmi/files/vcmi-data/tmp' as temporary directory
[2025-Jul-24 11:30:29.845427] INFO global [7bacccf500] - Performing native copy...
[2025-Jul-24 11:30:29.850267] INFO global [7bacccf500] - Native copy completed
[2025-Jul-24 11:30:29.850418] INFO global [7bacccf500] - Checking file /storage/emulated/0/Android/data/is.xyz.vcmi/files/vcmi-data/tmp/h3_gog-1.bin
[2025-Jul-24 11:30:29.850559] INFO global [7bacccf500] - File cannot be opened: No such file or directory
[2025-Jul-24 11:30:29.850598] INFO global [7bacccf500] - Installing exe '/storage/emulated/0/Android/data/is.xyz.vcmi/files/vcmi-data/tmp/h3_gog.exe' ('content://com.android.externalstorage.documents/document/primary%3ADownload%2FHeroes of Might and Magic 3 Complete%2Fsetup.exe')
[2025-Jul-24 11:30:29.850621] INFO global [7bacccf500] - Installing bin '/storage/emulated/0/Android/data/is.xyz.vcmi/files/vcmi-data/tmp/h3_gog-1.bin' ('content://com.android.externalstorage.documents/document/primary%3ADownload%2FHeroes of Might and Magic 3 Complete%2Fdata.bin')
[2025-Jul-24 11:30:29.867486] ERROR global [7bacccf500] - Gog installer extraction failure! Reason: Failed to open file: No such file or directory
```

**Version**

- OS: Android 16
- Version: 1.6.8

Laserlicht:
fixes #5945

Tested with space and german umlauts.

Not sure why https://github.com/qt/qtbase/commit/d5e03ad4e48c6e3229c88686cfc071ce57956dea doesn't work. We're using 5.15.16 which should have the fix.

Related: https://bugreports.qt.io/browse/QTBUG-114435

### DIFF
--- a/launcher/helper.cpp
+++ b/launcher/helper.cpp
@@ -93,11 +93,13 @@ void performNativeCopy(QString src, QString dst)
 		QByteArray utf8 = input.toUtf8();
 		QByteArray encoded;
 
-		for (char c : utf8) {
+		for (char c : utf8)
+		{
 			// If ASCII (0x00 to 0x7F), keep as is
-			if (static_cast<unsigned char>(c) < 0x80) {
+			if (static_cast<unsigned char>(c) < 0x80)
 				encoded.append(c);
-			} else {
+			else
+			{
 				// Non-ASCII: encode as %HH
 				encoded.append('%');
 				encoded.append(QByteArray::number(static_cast<unsigned char>(c), 16).toUpper().rightJustified(2, '0'));
--- a/launcher/helper.cpp
+++ b/launcher/helper.cpp
@@ -89,8 +89,34 @@ QString getRealPath(QString path)
 void performNativeCopy(QString src, QString dst)
 {
 #ifdef VCMI_ANDROID
-	auto srcStr = QAndroidJniObject::fromString(src);
-	auto dstStr = QAndroidJniObject::fromString(dst);
+	auto percentEncodeNonAscii = [](const QString &input) {
+		QByteArray utf8 = input.toUtf8();
+		QByteArray encoded;
+
+		for (char c : utf8) {
+			// If ASCII (0x00 to 0x7F), keep as is
+			if (static_cast<unsigned char>(c) < 0x80) {
+				encoded.append(c);
+			} else {
+				// Non-ASCII: encode as %HH
+				encoded.append('%');
+				encoded.append(QByteArray::number(static_cast<unsigned char>(c), 16).toUpper().rightJustified(2, '0'));
+			}
+		}
+
+		return QString::fromUtf8(encoded);
+	};
+
+	auto safeEncode = [&](QString uri) -> QString
+	{
+		if (!uri.startsWith("content://", Qt::CaseInsensitive))
+			return uri;
+		uri.replace(" ", "%20");
+		return percentEncodeNonAscii(uri);
+	};
+
+	auto srcStr = QAndroidJniObject::fromString(safeEncode(src));
+	auto dstStr = QAndroidJniObject::fromString(safeEncode(dst));
 	QAndroidJniObject::callStaticObjectMethod("eu/vcmi/vcmi/util/FileUtil", "copyFileFromUri", "(Ljava/lang/String;Ljava/lang/String;Landroid/content/Context;)V", srcStr.object<jstring>(), dstStr.object<jstring>(), QtAndroid::androidContext().object());
 #else
 	QFile::copy(src, dst);
--- a/launcher/helper.cpp
+++ b/launcher/helper.cpp
@@ -89,20 +89,21 @@ QString getRealPath(QString path)
 void performNativeCopy(QString src, QString dst)
 {
 #ifdef VCMI_ANDROID
-	auto percentEncodeNonAscii = [](const QString &input) {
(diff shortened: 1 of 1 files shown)


launcher/helper.cpp
```diff
@@ -89,8 +89,36 @@ QString getRealPath(QString path)
 void performNativeCopy(QString src, QString dst)
 {
 #ifdef VCMI_ANDROID
-	auto srcStr = QAndroidJniObject::fromString(src);
-	auto dstStr = QAndroidJniObject::fromString(dst);
+	auto percentEncodeNonAscii = [](const QString &input) {
+		QByteArray utf8 = input.toUtf8();
+		QByteArray encoded;
+
+		for (char c : utf8)
```
kambala-decapitator: seems easier to do this instead of casting later:
```suggestion
		for (unsigned char c : utf8)
```

launcher/helper.cpp
```diff
@@ -89,8 +89,36 @@ QString getRealPath(QString path)
...
+		{
+			// If ASCII (0x00 to 0x7F), keep as is
+			if (static_cast<unsigned char>(c) < 0x80)
+				encoded.append(c);
+			else
+			{
+				// Non-ASCII: encode as %HH
+				encoded.append('%');
+				encoded.append(QByteArray::number(static_cast<unsigned char>(c), 16).toUpper().rightJustified(2, '0'));
```
kambala-decapitator: there's no overload with `unsigned char`, better use `uint`
```suggestion
				encoded.append(QByteArray::number(static_cast<uint>(c), 16).toUpper().rightJustified(2, '0'));
```

kambala-decapitator (commented):
you could also build Qt with debug info and step into that method if you're interested why its code doesn't work

kambala-decapitator:
btw regarding code style: we use no space between if/for and opening bracket

Laserlicht:
> if you're interested why its code doesn't work

To be honest, I'm glad that it works this way and don't want to venture too deeply into QT. :)

IvanSavenko:
Do we need to reinvent this wheel?
https://doc.qt.io/qt-6/qurl.html#toPercentEncoding

Laserlicht:
> Do we need to reinvent this wheel? https://doc.qt.io/qt-6/qurl.html#toPercentEncoding

I know this function. And it doesn't work in our case.
It will double encode if there are already `%` present -> also breaking the path (already tested it).

Laserlicht:
> Can you add comment in code as to why this workaround is needed?

Sure. :)

Laserlicht:
Just noticed that `toPercentEncoding` has an exclude. So we can use this.
We have to exclude almost everything (except space), that we have the same behavior, than custom function before. If we don't do this e.g. `(` will be encoded, which doesn't work.

Only Space and UTF8 characters should encoded.

launcher/helper.cpp
```diff
@@ -89,8 +89,19 @@ QString getRealPath(QString path)
 void performNativeCopy(QString src, QString dst)
 {
 #ifdef VCMI_ANDROID
-	auto srcStr = QAndroidJniObject::fromString(src);
-	auto dstStr = QAndroidJniObject::fromString(dst);
+	// %-encode unencoded parts of string.
+	// This is needed because QT returns a mixed content url with %-encoded and unencoded parts. If Android >= 13 this causes problems reading this files, when using spaces and unicode characters in folder or filename.
+	// Only these should encoded (other typically %-encoded chars should not encoded because this leads to errors).
```
kambala-decapitator:
```suggestion
	// This is needed because Qt returns a mixed content url with %-encoded and unencoded parts. On Android >= 13 this causes problems reading these files, when using spaces and unicode characters in folder or filename.
	// Only these should be encoded (other typically %-encoded chars should not be encoded because this leads to errors).
```

GeorgeK1ng:
I have just tried to import ZIP with map where is space in file name. Nothing happends, no error, nothing. When I remove space from that file name, launcher import map from ZIP just fine. 
Tested on latest develop on Android 15 tablet.

Laserlicht:
Not sure if ZIP and map installing is using the same code path than exe installing. Because some stuff is going over the DownloadManager.

Laserlicht:
@GeorgeK1ng You've used in this PR: #5959

`QFile::copy` which is problematic. You should use `Helper::performNativeCopy`

Issue #6069, opened 2025-08-26
Hot seat never allows more players


GeorgeK1ng:
In 1.7 develop when launching Hot seat with 2 players is always launched as single player. Second player is not listed there.

<img width="369" height="414" alt="Image" src="https://github.com/user-attachments/assets/0b728353-c682-4fd8-8c3f-e2f12030f7a7" />

<img width="770" height="590" alt="Image" src="https://github.com/user-attachments/assets/aa9f5220-07f9-4168-9438-5bc792b6974d" />

And when hit back button game crash.

<img width="769" height="591" alt="Image" src="https://github.com/user-attachments/assets/215c359a-af44-44a5-b60f-9922eb1b9494" />

Olegmods:
Looks related to #5986 and #6045 .
Same stuff happens in lobby with develop.

Opuszek:
I remember finding this problem.
It was introduced by https://github.com/vcmi/vcmi/commit/44cc3a02148aa9267e776572d9c27eeb61b9a7b7
(I tested before the commit and after).
Sorry, I forgot to report it.

PR #6088, opened 2025-08-31
Fix PlayerConnectionID regression

MichalZr6:
@IvanSavenko, this should fix regression from [commit 52da332](https://github.com/vcmi/vcmi/commit/52da332640ed1ce0c70353739cf42d973038ec3b) that you mentioned on discord and - hopefully - also the issue with non-working hotseat reported by @edeksumo 

Other than that I don't see any more problems in code there.

### DIFF
--- a/server/CVCMIServer.cpp
+++ b/server/CVCMIServer.cpp
@@ -599,7 +599,7 @@ void CVCMIServer::setPlayer(PlayerColor clickedColor)
 	{
 		PlayerColor color;
 		PlayerConnectionID id;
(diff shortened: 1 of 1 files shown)


GeorgeK1ng:
Does this fix these ones too? 

https://github.com/vcmi/vcmi/issues/6069

edeksumo:
It seems that you still cannot play hot-seat

MichalZr6:
> Does this fix these ones too?
> 
> #6069

No, but I already know where the bug is. Will fix.

PR #6091, opened 2025-08-31
keep screen on in launcher while downloading

### LINKED ISSUE #5995: Do not turn off display during mods downloading / installing

Laserlicht:
Fixes #5995

Replacement for #6022

Only keep screen on while downloading/extracting.

iOS can be added in separate PR later.

Sadly there is no native function in QT for this.

### DIFF
--- a/launcher/helper.cpp
+++ b/launcher/helper.cpp
@@ -22,6 +22,7 @@
 #ifdef VCMI_ANDROID
 #include <QAndroidJniObject>
 #include <QtAndroid>
+#include <QAndroidJniEnvironment>
 #endif
 
 #ifdef VCMI_IOS
@@ -113,5 +114,33 @@ MainWindow * getMainWindow()
 		if(auto mainWin = qobject_cast<MainWindow*>(w))
 			return mainWin;
 	return nullptr;
+}
+
+
+void keepScreenOn(bool on)
+{
+#if defined(VCMI_ANDROID)
+	// based on https://stackoverflow.com/a/38846485
+	QtAndroid::runOnAndroidThread([on]
+	{
+		QAndroidJniObject activity = QtAndroid::androidActivity();
+		if(activity.isValid())
+		{
+			QAndroidJniObject window = activity.callObjectMethod("getWindow", "()Landroid/view/Window;");
+
+			if(window.isValid())
+			{
+				const int FLAG_KEEP_SCREEN_ON = 128;
+				if(on)
+					window.callMethod<void>("addFlags", "(I)V", FLAG_KEEP_SCREEN_ON);
+				else
+					window.callMethod<void>("clearFlags", "(I)V", FLAG_KEEP_SCREEN_ON);
+			}
+		}
+		QAndroidJniEnvironment env;
+		if (env->ExceptionCheck())
+			env->ExceptionClear();
+	});
+#endif
 }
 }
--- a/launcher/helper.cpp
+++ b/launcher/helper.cpp
@@ -22,11 +22,11 @@
 #ifdef VCMI_ANDROID
 #include <QAndroidJniObject>
 #include <QtAndroid>
-#include <QAndroidJniEnvironment>
 #endif
 
 #ifdef VCMI_IOS
 #include "ios/revealdirectoryinfiles.h"
+#include "iOS_utils.h"
 #endif
 
 #ifdef VCMI_MOBILE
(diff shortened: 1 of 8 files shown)


Laserlicht:
@kambala-decapitator Tested also SDL2 (`SDL_DisableScreenSaver`). Compiles fine. But calling these functions does nothing.

Seems to need an SDL activity to work. So we need this custom implementation.

kambala-decapitator:
the SDL function doesn't work because we don't configure its video subsystem in the launcher (we don't need it). But we can easily use SDL's actual implementation of the screensaver handling without reinventing the wheel:

https://github.com/libsdl-org/SDL/blob/98d1f3a45aae568ccd6ed5fec179330f47d4d356/src/core/android/SDL_android.c#L2198-L2209

basically what you need to is call this static method on the SDL activity class, please check if that'd work:

https://github.com/libsdl-org/SDL/blob/98d1f3a45aae568ccd6ed5fec179330f47d4d356/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L1123

or expose a new java method on the launcher activity that does the same thing instead of using that ugly C++ JNI code, implementation would be just a single line of code: https://developer.android.com/develop/background-work/background-tasks/awake/screen-on#java

on iOS we can't reuse SDL's implementation, but it's also a trivial one-liner of objc code, put it into `iOS_utils` (iirc it's already linked to launcher):
```objc
void setScreensaverEnabled(bool isEnabled)
{
	UIApplication.sharedApplication.idleTimerDisabled = isEnabled ? NO : YES;
}
```

launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -798,6 +798,7 @@ void CModListView::downloadFile(QString file, QUrl url, QString description, qin
 
 void CModListView::downloadProgress(qint64 current, qint64 max)
 {
+	Helper::keepScreenOn(true);
```
kambala-decapitator: better do it once on download start

Laserlicht:
Ugly, but it works. 🤣

Tested a lot. But testing here is a pain.

For example my first test:
https://github.com/vcmi/vcmi/commit/27ee1359a63fffdd25b26a437d4d215c0152f910

No idea why it doesn't work...

Will test SDL call.

kambala-decapitator:
your initial attempt looks correct. Were there any errors / crashes?

Maybe the difference is `QtAndroid::runOnAndroidThread`? Have you checked `logcat` for hints?

Laserlicht:
Works now! `QtAndroid::runOnAndroidThread` was the missing thing.

launcher/helper.cpp
```diff
@@ -114,4 +115,17 @@ MainWindow * getMainWindow()
...
+
+void keepScreenOn(bool on)
+{
+#if defined(VCMI_ANDROID)
+	QtAndroid::runOnAndroidThread([on]
+	{
+		QtAndroid::androidActivity().callMethod<void>("setScreensaverEnabled", "(Z)V", !on);
+	});
+#elif defined(VCMI_IOS)
+	iOS_utils::setScreensaverEnabled(!on);
```
kambala-decapitator: why not naming all the methods in the same way?

launcher/modManager/cmodlistview_moc.cpp
```diff
@@ -806,6 +807,7 @@ void CModListView::downloadProgress(qint64 current, qint64 max)
 
 void CModListView::extractionProgress(qint64 current, qint64 max)
 {
+	Helper::keepScreenOn(true);
```
kambala-decapitator: is the call really needed here? Currently it looks imbalanced: you call it with true twice, but with false - only once.

Laserlicht: Added it above `ui->progressBar->setVisible(true);` because it should be depend on progressbar.

launcher/helper.cpp
Laserlicht: Can do. But would prefer `keepScreenOn`. Because `setScreensaverEnabled` is a little bit confusing (inverted logic).

kambala-decapitator: yes please, keepScreenOn is fine

launcher/modManager/cmodlistview_moc.cpp
kambala-decapitator: it shouldn't depend on any UI element. Correct logic is:
1. long operation starts -> Helper::keepScreenOn(true)
2. long operation ends -> Helper::keepScreenOn(false)

Laserlicht: Adjusted positions to set.

It's still imbalanced because there is an abort possibility.

kambala-decapitator (approved):

Issue #6045, opened 2025-08-22
Broken MP game


ikazarf:
I was trying to play a multiplayer game and as soon as player 2 joined, the game room started glitching out. The map size and player number were randomly changing on their own, nobody (including me, the host) could change the town types of any of the players and player 2 was not listed as controlling one of the factions. It said "Computer" on all but red player, even though P2 had joined. We also couldn't get player 3 to join at all. 

[VCMI_Launcher_log.txt](https://github.com/user-attachments/files/21928049/VCMI_Launcher_log.txt)
[VCMI_Client_log.txt](https://github.com/user-attachments/files/21928050/VCMI_Client_log.txt)

1. Start Game,
2. Open a Multiplayer game, choose online server.
3. Have another player join.
4. Try to have a third player join.
5. The page glitches

- OS: Windows 10
- Version: VCMI 1.7

Context:
Tried with and without mods, no difference.

Olegmods:
Seems related to #5986

PR #6092, opened 2025-08-31
PlayerConnectionID rework: struct wrapper; Fix hotseat and placing players in lobby

### LINKED ISSUE #5986: Lobby doesn't work properly
**Describe the bug**
Lobby doesn't work properly and counts 2 players as one.

**Game logs**
No need

**To Reproduce**
1. Enter lobby
2. Host a game
3. Enter the room from another device
4. See issue

**Expected behavior**
2 players being in room

**Actual behavior**
Lobby counts two players as 1.

**Did it work earlier?**
I think it did.

**Screenshots**
No need, I assume.

**Version**

- OS: Windows
- Version: develop

### LINKED ISSUE #6069: Hot seat never allows more players

MichalZr6:
This PR aims to:

- fix #6069 
- fix #5986 

This one maybe:
- #6045

### DIFF
--- a/server/CVCMIServer.cpp
+++ b/server/CVCMIServer.cpp
@@ -419,18 +419,20 @@ void CVCMIServer::clientConnected(std::shared_ptr<GameConnection> c, std::vector
 		si->mode = mode;
 	}
 
-	logNetwork->info("Connection with client %d established. UUID: %s", static_cast<int>(c->connectionID), c->uuid);
+	auto connID = static_cast<int>(c->connectionID);
 
+	logNetwork->info("Connection with client %d established. UUID: %s", connID, c->uuid);
+
+	PlayerConnectionID id = currentPlayerId;
 	for(auto & name : names)
 	{
-		logNetwork->info("Client %d player: %s", static_cast<int>(c->connectionID), name);
-		PlayerConnectionID id = vstd::next(currentPlayerId, 1);
+		logNetwork->info("Client %d player: %s", connID, name);
 
 		ClientPlayer cp;
 		cp.connection = c->connectionID;
 		cp.name = name;
 		playerNames.try_emplace(id, cp);
-		announceTxt(boost::str(boost::format("%s (pid %d cid %d) joins the game") % name % static_cast<int>(id) % static_cast<int>(c->connectionID)));
+		announceTxt(boost::str(boost::format("%s (pid %d cid %d) joins the game") % name % static_cast<int>(id) % connID));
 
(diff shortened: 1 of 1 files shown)


GeorgeK1ng:
Hot Seat seems to work now!

lib/serializer/PlayerConnectionID.h
```diff
@@ -11,11 +11,48 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
-enum class PlayerConnectionID : int8_t
+struct DLL_LINKAGE PlayerConnectionID
```
IvanSavenko: Why not reuse existing Identifier class? Check for example classes like PrimarySkillID or SpellShoolID

server/CVCMIServer.cpp
```diff
@@ -419,18 +419,19 @@ void CVCMIServer::clientConnected(std::shared_ptr<GameConnection> c, std::vector
 		si->mode = mode;
 	}
 
-	logNetwork->info("Connection with client %d established. UUID: %s", static_cast<int>(c->connectionID), c->uuid);
+	auto connID = static_cast<int>(c->connectionID);
```
IvanSavenko, line 422: Again - identifier classes have explicit getNum method for such cases.

lib/serializer/PlayerConnectionID.h
MichalZr6: Sure, didn't think of that.

server/CVCMIServer.cpp
MichalZr6: I didn't touch `GameConnectionID` just used that `connID` as static_cast was done in multiple places there.
Should it be also made as struct instead of enum ?
Ok... just checked. It should: `c->connectionID = vstd::next(currentClientId, 1);`

IvanSavenko: If connectionID is enum - then code is ok. 
If connectionID is your new class -then switch to Identifier and use getNum(). Ideally - directly in logging call, without creating local.

IMO converting enum's or Identifiers to integers is necessary evil, so better to keep scope of this "evil" as small as possible

MichalZr6: This struct wrapper is unnecessary. Thought incrementing enum after defined fields is the bug. It is not. The fix is trivial will commit a bit later.

PR #6094, opened 2025-09-01
fix ASAN detected leaks and heap-use-after-free error

thehonestcto:
I have ASAN active all the time when working on Nullkiller2, therefore I saw that VCMI launcher has some issues when being ran with ASAN.
This small PR is fixing 2 leaks and one heap-use-after-free error, so you can open the editor or download a mod without ASAN being unhappy.

### DIFF
--- a/launcher/modManager/cdownloadmanager_moc.cpp
+++ b/launcher/modManager/cdownloadmanager_moc.cpp
@@ -73,17 +73,19 @@ void CDownloadManager::downloadFinished(QNetworkReply * reply)
 	if(possibleRedirectUrl.isValid())
 	{
 		QString filename;
+		qint64 totalSize = 0;
 
 		for(int i = 0; i< currentDownloads.size(); ++i)
 		{
 			if(currentDownloads[i].file == file.file)
 			{
 				filename = currentDownloads[i].filename;
+				totalSize = currentDownloads[i].totalSize;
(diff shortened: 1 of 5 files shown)


launcher/modManager/cdownloadmanager_moc.cpp
```diff
@@ -73,17 +73,19 @@ void CDownloadManager::downloadFinished(QNetworkReply * reply)
...
 			if(currentDownloads[i].file == file.file)
 			{
 				filename = currentDownloads[i].filename;
+				totalSize = currentDownloads[i].totalSize;
 				currentDownloads.removeAt(i);
 				break;
 			}
 		}
-		downloadFile(qurl, filename, file.totalSize);
+		downloadFile(qurl, filename, totalSize);
```
kambala-decapitator, line 88: What did the diagnostic say about this? Not quite clear why the change is needed.

kambala-decapitator (approved):

launcher/modManager/cdownloadmanager_moc.cpp
thehonestcto: @kambala-decapitator heap-use-after-free
image

kambala-decapitator: thanks, now I see it - it's due to `currentDownloads.removeAt(i)`

PR #6098, opened 2025-09-01
[docs] add minizip to Linux

Laserlicht:


GeorgeK1ng:
Btw, for PortMaster builds before migrating to Conan I must do more tweaks to be able to compile it.
https://github.com/GeorgeK1ng/vcmi/blob/portmaster/docker/BuildPortmaster-aarch64.dockerfile

And then only add required workflow steps to make it real for daily builds.

Laserlicht:
Yes. Maybe some libs from old ubuntu. Portmaster relies on old ubuntu 20.04.

kambala-decapitator:
> PortMaster builds before migrating to Conan

I thought PortMaster is supposed to use libs from the package manager just like any Linux?

using Conan is possible of course, but would require packaging all the dependencies together with the binaries.

GeorgeK1ng:
Good is, I reveived confirmation compiled it works on some testing device. But when we add it to workflow, will probably receive more confirmations.

kambala-decapitator:
@GeorgeK1ng please check my question above

GeorgeK1ng:
Can't tell, I only know it's linux and what must be done to compile 1.7 with "minimal" changes to original settings from @Laserlicht portmaster repo.

Issue #2499, opened 2023-08-06
Fragport graphics from mod objects are not present in Kingdom Overwiev window.


misiokles:
**Describe the bug**
Owned dwellings are displayed in Kingdom Overwiev window. Only OH3, dwelling from mods - not

**Game logs**
![image](https://github.com/vcmi/vcmi/assets/29593042/987c28c1-b6e5-448a-9c84-8c6e6f681acb)

**To Reproduce**
1. Enable HotA for example
2. Visit any hota dwelling on map.
3. Click Kingdom Overwiev button

**Version**
 - OS: Windows
 - Version: 1.3

GeorgeK1ng:
This issue is still there, dwellings from mods are no displayed in Kingdom Overview.

PR #6099, opened 2025-09-02
kingdom overview dwellings for mods

### LINKED ISSUE #6097: FlagPort support for mods
It's, one of the last missing features missing from OH according to mods (this little icons of dwelling in kingdom overview window). I think it should be implemented to attackFrame for map creatures, as not required to not break existing mods, but for those willing to provide the necessary graphics for it.

Laserlicht:
Fixes #6097

### DIFF
--- a/client/windows/CKingdomInterface.cpp
+++ b/client/windows/CKingdomInterface.cpp
@@ -515,6 +515,22 @@ void CKingdomInterface::generateObjectsList(const std::vector<const CGObjectInst
 				info.imageID = object->subID;
 			}
 		}
+		else if(auto * dwelling = dynamic_cast<const CGDwelling *>(object))
+		{
+			auto kingdomOverviewImage = dwelling->getKingdomOverviewImage();
+
+			if(!kingdomOverviewImage.empty())
+			{
+				OwnedObjectInfo & info = visibleObjects[object->subID];
+				if(info.count++ == 0)
+				{
+					info.hoverText = object->getObjectName();
+					info.imagePath = kingdomOverviewImage;
+					info.imageID = 0;
+				}
+			}
+		}
+
 		//Special objects from idToImage map that should be displayed in objects list
 		auto iter = idToImage.find(std::make_pair(object->ID, object->subID));
 		if(iter != idToImage.end())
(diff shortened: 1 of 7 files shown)


GeorgeK1ng:
This will require changes accross mods with additional settings. Or if not present this config in mod, will be used it's DEF and frame 0 instead if modder have not added custom image there?

Also related: https://github.com/vcmi/vcmi/issues/2499

misiokles:
It's similar issue like avwattak frames. If they are not set, nothing bad happens.

Laserlicht:
Exactly. If not set, there will be no dwelling.
So, no breaking change. Just optional possibility.

GeorgeK1ng:
Yes, but seems more compatible to automatically read default def frame 0 if not present optional settings for flagport.

Laserlicht:
> Yes, but seems more compatible to automatically read default def frame 0 if not present optional settings for flagport.

Yes. But will require more complexity because images needs to scaled. Just using frame 0 is not enough.

Would prefer OH3 way with enforcing extra image.

docs/modders/Map_Objects/Dwelling.md
```diff
@@ -22,7 +22,10 @@
 	"guards" : true,
 	"guards" : [
 		{ "amount" : 12, "type" : "earthElemental" }
-	]
+	],
+
+	/// Optional image showed on kingdom overview (animation; only frame 0 displayed)
```
IvanSavenko: Perhaps remove "optional"part, at least from docs?

This image is required for dwelling to function correctly after all.

Would be great to also enforce it via schema, but looks like dwellings don't have one at the moment & that would cause validation failure in multiple mods, so probably not for 1.7

client/windows/CKingdomInterface.cpp
```diff
@@ -515,6 +515,22 @@ void CKingdomInterface::generateObjectsList(const std::vector<const CGObjectInst
 				info.imageID = object->subID;
 			}
 		}
+		else if(auto * dwelling = dynamic_cast<const CGDwelling *>(object))
```
IvanSavenko: Perhaps instead:
If object is dwelling and has custom image - usr custom image.
Else- fall back to h3 logic.

This way, if mod were to define custom image for h3 dwelling in config, game will use it instead of h3 image

misiokles:
Is there sth should be done before merging? Because I could managed to make some frames for some popular mods...

Laserlicht:
From my side it's ready.

Laserlicht:
@IvanSavenko What about this one? :)

client/windows/CKingdomInterface.cpp
```diff
@@ -506,15 +506,31 @@ void CKingdomInterface::generateObjectsList(const std::vector<const CGObjectInst
 	for(const CGObjectInstance * object : ownedObjects)
 	{
 		//Dwellings
-		if(object->ID == Obj::CREATURE_GENERATOR1)
+		if(auto * dwelling = dynamic_cast<const CGDwelling *>(object))
```
IvanSavenko, line 509: Towns are also dwellings. No side effects?

IvanSavenko (approved):

client/windows/CKingdomInterface.cpp
Laserlicht: But they does not contain `kingdomOverviewImage`!? Then - old logic will be used.

PR #6104, opened 2025-09-03
Log error instead of throw when upgrade is same building ID

### LINKED ISSUE #6100: Add a check that  town building isn't upgrading to themselves
**Describe your proposal**
I had an issue that game was closing on start due to

"horde1Upgr" : 
{
...
"upgrades" : "horde1Upgr"
...
}

I propose to add check that "building" object don't have "upgrades" field set to itself

**Additional context**

MichalZr6:
- Fixes #6100

### DIFF
--- a/lib/entities/faction/CTownHandler.cpp
+++ b/lib/entities/faction/CTownHandler.cpp
@@ -361,14 +361,18 @@ void CTownHandler::loadBuilding(CTown * town, const std::string & stringID, cons
 		// building id and upgrades can't be the same
 		if(stringID == source["upgrades"].String())
 		{
-			throw std::runtime_error(boost::str(boost::format("Building with ID '%s' of town '%s' can't be an upgrade of the same building.") %
-												stringID % ret->town->faction->getNameTranslated()));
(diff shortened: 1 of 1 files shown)


lib/entities/faction/CTownHandler.cpp
```diff
@@ -361,14 +361,17 @@ void CTownHandler::loadBuilding(CTown * town, const std::string & stringID, cons
 		// building id and upgrades can't be the same
 		if(stringID == source["upgrades"].String())
 		{
-			throw std::runtime_error(boost::str(boost::format("Building with ID '%s' of town '%s' can't be an upgrade of the same building.") %
+			logMod->error(boost::str(boost::format("Building with ID '%s' of town '%s' can't be an upgrade of the same building.") %
```
IvanSavenko: Advantage of throwing is early detection. Most of throws were added after we had corrupted mod in the wild, often with rather unclear crash - not all modders check console

Perhaps keep throw but also add identical message to log?

MichalZr6: Sure, done.

Issue #6102, opened 2025-09-02
Generate new Garison background


GeorgeK1ng:
In HotA / HD+ is added special Garison background image, which fit's better for Garisons.

<img width="350" height="357" alt="Image" src="https://github.com/user-attachments/assets/03e7fe7b-27d5-400a-994b-38384998bd5d" />

<img width="371" height="331" alt="Image" src="https://github.com/user-attachments/assets/5f318c97-d08a-46cf-bd26-f002840a2e3f" />

This will require generate same background on the fly and use it for Garison.

GeorgeK1ng:
This is exact HD mod template and same is in original game data as GARRIPOP

<img width="292" height="124" alt="Image" src="https://github.com/user-attachments/assets/af4f5f58-6922-44f8-a004-bd74e41292f0" />

PR #6105, opened 2025-09-03
Use buildin Garrison background

GeorgeK1ng:
Original game data have special Garrison background - GARRIPOP.bmp which seems was never used. Now it's used correctly for Garrisons tooltip

Before:
image

After:
image

Fixes: https://github.com/vcmi/vcmi/issues/6102

### DIFF
--- a/client/windows/InfoWindows.cpp
+++ b/client/windows/InfoWindows.cpp
@@ -316,13 +316,13 @@ CInfoBoxPopup::CInfoBoxPopup(Point position, const CGHeroInstance * hero)
 }
 
 CInfoBoxPopup::CInfoBoxPopup(Point position, const CGGarrison * garr)
-	: AdventureMapPopup(RCLICK_POPUP | PLAYER_COLORED, ImagePath::builtin("TOWNQVBK"), position)
+	: AdventureMapPopup(RCLICK_POPUP | PLAYER_COLORED, ImagePath::builtin("GARRIPOP"), position)
 {
 	InfoAboutTown iah;
 	GAME->interface()->cb->getTownInfo(garr, iah);
 
 	OBJECT_CONSTRUCTION;
-	tooltip = std::make_shared<CArmyTooltip>(Point(9, 10), iah);
+	tooltip = std::make_shared<CGarrisonTooltip>(Point(9, 10), iah);
 	
 	addUsedEvents(DRAG_POPUP);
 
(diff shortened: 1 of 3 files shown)


dydzio0614:
Since this is not original game behavior, i think it may be better if new layout is only enabled if "enable UI enhancements" setting is toggled on

GeorgeK1ng:
Which settings key Is exactly? For me this sounds more like original game fix instead as it uses original game assets.

dydzio0614:
settings key is "enableUiEnhancements" and it should be available in game options - if you disable it you will not see "separate window backpack button" after opening hero window etc.

There are some oldschool h3 players who do not like hd mod's UI changes etc. and VCMI is mostly compatible with classic UI when that setting is turned off, no reason to abandon that compatibility

GeorgeK1ng:
Yeah, but creature extended window will always stay as is which is really more different against original game. So it's really questionable with small change in Garrison tooltip which uses buildin assets.

dydzio0614:
Removing original creature window was a mistake made long time ago

GeorgeK1ng:
Doing these fixes from phone Is real pain, but should be done. Someone need to test artifacts if possible.

GeorgeK1ng:
Anybody know what can cause crash when disable "Interface Enhancements" in ingame settings after my changes? When it's enabled, it works fine. But when when disabled, right click on Garrison crashes game..

Laserlicht:
> Anybody know what can cause crash when disable "Interface Enhancements" in ingame settings after my changes? When it's enabled, it works fine. But when when disabled, right click on Garrison crashes game..

Any stack trace?

client/windows/InfoWindows.cpp
```diff
@@ -316,14 +316,22 @@ CInfoBoxPopup::CInfoBoxPopup(Point position, const CGHeroInstance * hero)
...
-	tooltip = std::make_shared<CArmyTooltip>(Point(9, 10), iah);
-	
+
+	if(settings["general"]["enableUiEnhancements"].Bool())
+	{
+        tooltip = std::make_shared<CGarrisonTooltip>(Point(9, 10), iah);
+	}
+	else
+	{
+        tooltip = std::make_shared<CTownTooltip>(Point(9, 10), iah);
```
Laserlicht:
```suggestion
        tooltip = std::make_shared<CArmyTooltip>(Point(9, 10), iah);
```

Why `CTownTooltip`? Original is `CArmyTooltip`. This fixes crash...

GeorgeK1ng: Thanks! Great catch! I did this change from phone, have not noticed copy / paste issue.

PR #6107, opened 2025-09-03
Add possibility to define custom main menu around image

GeorgeK1ng:
Add possibility to define custom image around main menu.

So now can be used custom colored one or single color image there instead of hardcoded DIBOXBCK.

image


Laserlicht:
There are more usages of this graphics in a similar way. Not only in main menu.

For example with show background setting disabled.

GeorgeK1ng:
Yes, this is just possibility to add custom around image to main menu without breaking other usage of DIBOXBCK.

PR #6114, opened 2025-09-08
improve performance of Battle AI choosing spell

Opuszek:
The goal of this draft is to improve performance of method BattleEvaluator::attemptCastingSpell. It can take ridiculously long.
Two improvement have been introduced:
1) Removal of duplicate evaluation.
In the old code, evaluating fireball spell in such a situation
Screenshot from 2025-09-08 19-35-33

would produce 12 computations: one for a tile taken by a unit and one for each surrounding tile. In the new one, it takes only 3 computations: one in which fireball hits pikes, one in which it hits swordsmen and one in which it hits both of them.

2) Removal of evaluation for negative spells (e.g fireball) hitting only allied units.

So far the results seem encouraging. I run a particularly time-consuming battle both before and after and the results were exactly the same, only over two times quicker.
Screenshot from 2025-09-08 19-41-38
(The disparity increases later and in the end it was ~19 second vs ~40 (adding runtime of separate threads).

The code is a draft for now as I don't consider it merge-ready yet.
a) It cannot deal properly with "neutral spells" like firewall or force field. AI rarely knows what to do with them anyway, but I think It is something that needs to be addressed.
b) I need to check more thoroughly if BattleCast::findPotentialTargets is not used in different code and what it is expected to do. Maybe the change will need to be moved to a separate method.
c) It needs more testing.

I will leave it for now here to check if anybody has any comments/opinions.

### DIFF
--- a/lib/spells/BattleSpellMechanics.cpp
+++ b/lib/spells/BattleSpellMechanics.cpp
@@ -687,26 +687,38 @@ std::vector<Destination> BattleSpellMechanics::getPossibleDestinations(size_t in
 	case AimType::LOCATION:
 		if(fast)
 		{
-			auto stacks = battle()->battleGetAllStacks();
-			BattleHexArray hexesToCheck;
-
-			for(auto stack : stacks)
+			std::set<std::vector<const CStack *>> affectedUnitsSet;
+			for(int i = 0; i < GameConstants::BFIELD_SIZE; i++)
 			{
-				hexesToCheck.insert(stack->getPosition());
-				hexesToCheck.insert(stack->getPosition().getNeighbouringTiles());
-			}
+				BattleHex dest(i);
+				spells::Target target;
+				target.emplace_back(dest);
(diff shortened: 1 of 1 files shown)


lib/spells/BattleSpellMechanics.cpp
```diff
@@ -687,26 +687,38 @@ std::vector<Destination> BattleSpellMechanics::getPossibleDestinations(size_t in
...
+				BattleHex dest(i);
+				spells::Target target;
+				target.emplace_back(dest);
+				std::vector<const CStack *> affectedUnits = getAffectedStacks(target);
 
-			for(const auto & hex : hexesToCheck)
-			{
-				if(hex.isAvailable())
+				bool isAffectedAlly = false;
+				bool isAffectedEnemy = false;
```
C3pa, line 699: Looks like you can drop one of these variables and only use one.

Issue #6096, opened 2025-09-01
Antagarich Burning maps stop working


misiokles:
**Describe the bug**
In daily develops two of three Antagarich Burning maps (Catherine's Downfall and Return of the Wizard) stopped working. Game crashes during map opening and console spams red errors about missing mod dependencies. Which is strange, because dependencies are set and nothing was changed from working version from 1.6.8.

**Game logs**

[VCMI_Client_log.txt](https://github.com/user-attachments/files/22073256/VCMI_Client_log.txt)

**To Reproduce**

1. Download Antagarich Burning maps mod (Antagarich Burning assets and Forge2K should install automatically)
2. Go to Map selection panel, scroll down to the list to find three vcmaps.
3. Try to open them
4. See crash

**Did it work earlier?**
Yes, still work with 1.6.8 release

**Version**

- OS: [e.g. Windows]
- Version: [VCMI latest develops]

PR #6126, opened 2025-09-10
Rework handling maps that can't be launched via client or mapeditor

MichalZr6:
This PR aims to:
- follow-up for: #5919 -> fixes: #6096
- handle maps that were built with objects from particular mods, whereas at the same time these mods were not added to map's required mods list. On client this results in bypassing catching `ModIncompatibility` exceptions and crash in other place
- better error handling, distinguish `vcmi.server.errors.unknownEntity` message between saveFile and campaign/map file
- provide `modScope` field for each vmap's object on saving a map


Opuszek:
Just a question, since I was working on the same problem but there is still much I don't understand.
Remote scope is set in name of a JsonNode object (like this: "remoteScope:nameOfObject"), but local scope is set explicitly in "modScope" field. Now, "Game" local scope encompasses both builtin objects and mods. Assuming I am right, is there a reason for a map object to be serialized with a different modScope than "Game"?

I thought that maybe the problem can be resolved by moving the responsibility of setting local scope when looking for identifiers to JSONFormatter.h, like this:

+                               LIBRARY->identifiers()->requestIdentifier(ModScope::scopeGame(), E::entityType(), fieldValue[i].String(), [&value, i](int32_t index){

And indeed, at least on surface level it fixed the problems with Antagarich Burning maps, though it very probably introduced bugs when deserializing objects with a different local scope than Game. But what would be those objects?

Also, class JsonSerializer.cpp sets ModScope explicitly to Game in two cases, I think it was added as a quick fix to this problem. Maybe it should be removed now, when a more comprehensive fix has been introduced.

Screenshot from 2025-09-11 10-57-44

MichalZr6:
Thanks @Opuszek for your comment.

The "modScope" I added to each object in objects.json (part of vmap) is not used at runtime per-se right now. I added it only because maps used to carry it and I thought: this might help parsing maps. But I haven't explored it futher yet... :wink:
Probably cleaner approach might be to just default to `ModScope::scopeGame()`

I'll defer to @IvanSavenko  on this matter - could you share your thoughts ?

IvanSavenko:
When loading vmap:

Situation that happens all the time - mods get renamed, or objects are moved between submods.

As a more robust solution - try to resolve with intended mod scope and then - with game scope?

IvanSavenko:
When loading vmap:

`                configuration.setModScope(ModScope::scopeGame());`

This should be used for *all* cases.

Right now you may have situation where object from mod a (e.g. hero) contains objects from mod b (e.g. creatures). Changing mod scope to a means that id's from b will be inaccessible.

Correct solution is to use game scope or implement "map" scope that provides access to all map dependencies and only to them

IvanSavenko:
> Assuming I am right, is there a reason for a map object to be serialized with a different modScope than "Game"?

Not much. In theory would be great to yave actual "map" scope that actually consists from mods the map depends on. In which case vmaps would be saved & loaded using such scope

IvanSavenko:
> I thought that maybe the problem can be resolved by moving the responsibility of setting local scope when looking for identifiers to JSONFormatter.h, like this:

Mod scope of a json node must be set by whoever loaded/generated this json. Serializer should always receive json with mod scope set

MichalZr6:
Dropped setting modscope for objects in vmap.

MichalZr6:
Also, I’m splitting this PR into two. This one is much higher priority, as it fixes multiple issues currently in the develop branch. The other handles some edge-case errors.

lib/mapping/MapFormatJson.cpp
```diff
@@ -1044,7 +1046,22 @@ void CMapLoaderJson::MapObjectLoader::construct()
...
+		configuration.setModScope(ModScope::scopeGame());
+		modScope = ModScope::scopeGame();
+	}
+	else
+	{
+		static const auto & activeMods = LIBRARY->modh->getActiveMods();
+		if(vstd::find(activeMods, modScope) == activeMods.end())
+			throw ModIncompatibility({modScope});
+
+		configuration.setModScope(modScope);	// propagate it recursively
```
IvanSavenko: Same question as before - what if objects (e.g. hero) comes from mod a, but has creatures/artifacts from unrelated mod b?

In this case, identifiers wull be resolved using mod a as scope, but won't be found since mod a does not depends on mod b

lib/modding/IdentifierStorage.cpp
```diff
@@ -246,7 +246,7 @@ std::optional<si32> CIdentifierStorage::getIdentifierImpl(const ObjectCallback &
 
 	if (idList.size() == 1)
 		return idList.front().id;
-	if (!silent)
+	if (!silent && options.localScope != ModScope::scopeGame())
```
IvanSavenko, line 249: Workaround, but not sure of better solution...

lib/mapping/MapFormatJson.cpp
MichalZr6: Ok, so for now just `configuration.setModScope(ModScope::scopeGame());`

PR #6127, opened 2025-09-11
Creature window rework

Laserlicht:
Remove assets for creature window and create with AssetGenerator. Looks more like a native dialog now.

grafik
grafik


Laserlicht:
Also generate quest window

grafik

GeorgeK1ng:
Noticed bug on commander screen only.

image

PR #6136, opened 2025-09-14
Configurable resources

Laserlicht:
Only backend stuff...

Todo in future PRs:
- [ ] GUI stuff (not sure how to do yet...)
  - res bars
  - trading (`TPMrkAbS` `TPMrkAsS` `TPMrkCrS` `TPMrkPtS` `TPMrkReS`)
  - handicap
  - kingdom overview (mines)
  - ...
- [x] Remove hardcoded mithril (if possible)
- [ ] Add mine support
- [x] map objects (resources)
- [x] bonus


AI/Nullkiller/AIUtility.cpp
```diff
@@ -470,7 +471,7 @@ int32_t getArtifactBonusScoreImpl(const std::shared_ptr<Bonus> & bonus)
 		case BonusType::UNDEAD_RAISE_PERCENTAGE:
 			return bonus->val * 400;
 		case BonusType::GENERATE_RESOURCE:
-			return bonus->val * LIBRARY->objh->resVals.at(bonus->subtype.as<GameResID>().getNum()) * 10;
+			return bonus->val * LIBRARY->resourceTypeHandler->getById(bonus->subtype.as<GameResID>())->getPrice() * 10;
```
IvanSavenko: Instead of getById better to convert GamResID into corresponding entity using toEntity(LIBRARY) / toResource methods of GameResID class (don't exist in develop, not sure about your pr, check other library types for reference)

Alucard648:
GUI variant for custom resources:
1. Add multiple rows in main window for resource indication (HD), or pop-op indicator for smaller HD.
2. Building costs: add either pop-up, or scrollable cost listing, if 9 or more different types of resources are used as cost.
3. Creature recruitment: If >=4 different types of resources are used as creature cost, add special button that displays total cost in pop-up window.
4. Payment in external map objects, same as creature recruitment, but use small icons if it does not fit with larger icons.
5. Object rewards: just list rewards in batches of 8 resources per window.

Mods/vcmi/Content/config/english.json
```diff
@@ -1083,5 +1083,14 @@
...
+	"spellSchool.core.water.name" : "Water",
+
+	"resources.core.wood.name" : "Wood",
+	"resources.core.mercury.name" : "Mercury",
+	"resources.core.ore.name" : "Ore",
+	"resources.core.sulfur.name" : "Sulfur",
+	"resources.core.crystal.name" : "Crystal",
+	"resources.core.gems.name" : "Gems",
+	"resources.core.gold.name" : "Gold",
+	"resources.core.mithril.name" : "Mithril"
```
IvanSavenko: Other than Mithril, can we load this strings from h3 txt's? For example by returning already existing text ID's for built-in resources?

client/mainmenu/CStatisticScreen.cpp
```diff
@@ -79,7 +80,7 @@ void CStatisticScreen::onSelectButton()
 			auto possibleRes = std::vector<EGameResID>{EGameResID::GOLD, EGameResID::WOOD, EGameResID::MERCURY, EGameResID::ORE, EGameResID::SULFUR, EGameResID::CRYSTAL, EGameResID::GEMS};
 			std::vector<std::string> resourceText;
 			for(const auto & res : possibleRes)
-				resourceText.emplace_back(LIBRARY->generaltexth->translate(TextIdentifier("core.restypes", res.getNum()).get()));
+				resourceText.emplace_back(MetaString::createFromTextID(LIBRARY->resourceTypeHandler->getById(res)->getNameTextID()).toString());
```
IvanSavenko: Similar here - add toEntity/toResource method instead

config/resources.json
```diff
@@ -1,4 +1,49 @@
 {
-	// Price of each resource in gold, in usual resource order
-	"resources_prices": [ 250, 500, 250, 500, 500, 500, 1, 0 ]
+	"wood": {
+		"index" : 0,
```
IvanSavenko, line 3: Do we name this field index or id in other entities?

lib/mapObjects/CObjectHandler.h
```diff
@@ -19,8 +19,6 @@ class int3;
 class DLL_LINKAGE CObjectHandler
 {
 public:
-	std::vector<ui32> resVals; //default values of resources in gold
```
IvanSavenko, line 22: Remove this now-empty class entirely

config/resources.json
Laserlicht: index is "correct"

Laserlicht:
Ready.

@IvanSavenko Please check if refactoring is okay. And merge if possible - to allow further PRs. Tested some things, looks good.

GeorgeK1ng:
UI will be probably most problematic.

For mines, when more than 7 types exist can be probably added left / right button (mspaint design). Maybe require smaller gold coins image in that case.

For bottom "status" bar I don't see good solution, maybe generate background with 2 lines for resources?
image

Here it should be no problem, even on phone there is enough space to add few new resources.

image

Laserlicht:
> Here it should be no problem, even on phone there is enough space to add few new resources.

It is a problem... We still support 800 resolution...

For statusbar: maybe a small button to open popup with other resources. Maybe the easiest way.

Laserlicht:
Another option: (maybe the easiest to implement, until we have a better solution)
If pressing ALT button (or two finger touch): replace status bar with alternative one.

GeorgeK1ng:
True, forgot to open original 800x600 size. Small button sound fine, if it will just open new popup with all resources.

But on bigger resolution it should be displayed on main adventure screen as other resources.

IvanSavenko:
While I agree that we need to support 800x600, for extended functionality I think we're fine to assume wide resolution as base, e.g. 960x600, in this case meaning + 160px for new resources.

Or we could try displaying only non-zero amounts to save on space at least in some cases

Laserlicht:
Test mod
[bitcoin.zip](https://github.com/user-attachments/files/22359198/bitcoin.zip)

grafik

GeorgeK1ng:
Really cool! Seems we can get Mithiril mine soon.

Laserlicht:
Game seems to work as before.
Removed Mithril (can added as mod). Refactored some stuff.

Currently it's possible:
- to define resources
- add them in other json configs (as costs)
- they are shown in some windows (e.g. where components are used)
- get resource (e.g. via cheat) / use it (buying stuff)

Other stuff will added later. This PR is already big enough.

Laserlicht:
grafik
grafik


resources on map is also already possible

```
	"objects" : {
	    "core:resource" : {
		    "types" : {
			    "bitcoin" :    { "aiValue" : 1400, "resource" : "bitcoin", "amounts" : [ 6,7,8,9,10], "rmg" : { "value" : 1400, "rarity" : 300 }, "templates" : { "res" : { "animation" : "bitcoin/BitcoinAdventureRes.def" } } },
		    }
	    }
	}
```

Laserlicht:
For mines: I'm giving up. May needs an own MineHandler. But takes too much time to figure out how to refactor. Needs someone who is more familiar with this codebase. Can be done later.

If someone want to see my attempt: https://github.com/vcmi/vcmi/compare/develop...Laserlicht:vcmi:resource_ext

Problem is rmg: abandoned mine is hardcoded index 7 -> but we need index 7 for new resources. But even with a fix for this it does not load correct images for mines from RMG.

------------------

Also tested resource generate bonus. Works as intended.

------------------

Also it's possible to see custom resources by right clicking on resource bar.

------------------

So... ready to review and merge I guess.

GeorgeK1ng:
Hope mines will be added soon.

Regarding to trading interface. I think there can be supported max 12 resources. But will required to generate few new backgrounds on the fly. For 8 and 9 it should be easy, just adding 2 boxes on both sides. Generating background for 9-12 resources will require bigger touch. Something like this.

image

Laserlicht:
> Hope mines will be added soon.

Would be great. :)

> Regarding to trading interface.

Yes, could be done by AssetGenerator.

GeorgeK1ng:
Yes, I think it require whole "diboxbck" part "generating". And have 8, 9, 10, 11, 12 background variants.

Laserlicht:
Should be doable. But in later PR after merging this one. This is already big enough.

misiokles:
Don't forget about handicap window!
image

GeorgeK1ng:
Handicap should be easier to achieve. Players flags on columns, resources in rows.

mapeditor/inspector/rewardswidget.cpp
```diff
@@ -779,8 +780,6 @@ void RewardsDelegate::updateModelData(QAbstractItemModel * model, const QModelIn
 		QStringList resourcesList;
 		for(GameResID resource = GameResID::WOOD; resource < GameResID::COUNT ; resource++)
```
IvanSavenko, line 781: Only h3 resources?

client/adventureMap/AdventureMapWidget.cpp
```diff
@@ -284,9 +286,9 @@ std::shared_ptr<CIntObject> AdventureMapWidget::buildResourceDateBar(const JsonN
 
 	auto result = std::make_shared<CResDataBar>(image, area.topLeft());
 
-	for(auto i = 0; i < GameConstants::RESOURCE_QUANTITY; i++)
+	for (auto & i : LIBRARY->resourceTypeHandler->getAllObjects())
 	{
-		const auto & node = input[GameConstants::RESOURCE_NAMES[i]];
+		const auto & node = input[GameResID(i).toResource()->getJsonKey()];
```
IvanSavenko: `i` is already GaneResID, so `GameResID(i)` is unnecessary - just `i` would suffice

client/mainmenu/CStatisticScreen.cpp
```diff
@@ -79,7 +80,7 @@ void CStatisticScreen::onSelectButton()
 			auto possibleRes = std::vector<EGameResID>{EGameResID::GOLD, EGameResID::WOOD, EGameResID::MERCURY, EGameResID::ORE, EGameResID::SULFUR, EGameResID::CRYSTAL, EGameResID::GEMS};
```
IvanSavenko: Not including mod resources?

lib/ResourceSet.cpp
```diff
@@ -13,17 +13,26 @@
...
+ResourceSet::ResourceSet()
+{
+	container.resize(LIBRARY->resourceTypeHandler->getAllObjects().size());
+};
 
 ResourceSet::ResourceSet(const JsonNode & node)
 {
-	for(auto i = 0; i < GameConstants::RESOURCE_QUANTITY; i++)
-		container[i] = static_cast<int>(node[GameConstants::RESOURCE_NAMES[i]].Float());
+	container.resize(LIBRARY->resourceTypeHandler->getAllObjects().size());
```
IvanSavenko: Potential problem - this code might be called before all game resources were loaded.

Tricky to solve... Normally, you can fix it via requestIdentifier calls. But that method is async, and ResourceSet can be copied & old reference with scheduled async calls - destroyed.

Not yet sure how to handle this nicely...

And another problem - if you make a typo, like "gem" instead of "gems" it won't be detected. Similarly, you can reference resource from mod without dependency, and it won't be detected.

Using requestIdentifier here would fix both problems

lib/rmg/CRmgTemplate.cpp
```diff
@@ -534,10 +535,8 @@ void ZoneOptions::serializeJson(JsonSerializeFormat & handler)
 	{
 		auto minesData = handler.enterStruct("mines");
 
-		for(TResource idx = 0; idx < (GameConstants::RESOURCE_QUANTITY - 1); idx++)
-		{
-			handler.serializeInt(GameConstants::RESOURCE_NAMES[idx], mines[idx], 0);
-		}
+		for(auto & idx : LIBRARY->resourceTypeHandler->getAllObjects())
```
IvanSavenko: Needs serializeIdArray method or something similar. Current code will hide typos and ignores mid dependencies

lib/modding/ModManager.cpp
```diff
@@ -807,7 +807,10 @@ void ModDependenciesResolver::tryAddMods(TModList modsToResolve, const ModsStora
 			{
 				resolvedOnCurrentTreeLevel.insert(*it); // Not to the resolvedModIDs, so current node children will be resolved on the next iteration
 				assert(!vstd::contains(sortedValidMods, *it));
-				sortedValidMods.push_back(*it);
+				if(storage.getMod(*it).getValue("modType").String() == "Resources") // Resources needs to load before core to make it possible to override core elements with new resources and for correct init of ResourceSet
```
IvanSavenko: Please avoid such hacks

lib/modding/ContentTypeHandler.h
```diff
@@ -50,7 +50,7 @@ class DLL_LINKAGE ContentTypeHandler
 /// class used to load all game data into handlers. Used only during loading
 class DLL_LINKAGE CContentHandler
 {
-	std::map<std::string, ContentTypeHandler> handlers;
+	std::vector<std::pair<std::string, ContentTypeHandler>> handlers;
```
IvanSavenko: Why do you need such change?

lib/json/JsonRandom.cpp
```diff
@@ -298,9 +299,9 @@ JsonRandom::JsonRandom(IGameInfoCallback * cb, IGameRandomizer & gameRandomizer)
 			return ret;
 		}
 
-		for (size_t i=0; i<GameConstants::RESOURCE_QUANTITY; i++)
+		for(auto & i : LIBRARY->resourceTypeHandler->getAllObjects())
```
IvanSavenko, line 302: loadResources method immediately below this one still contains fixed h3 resource list

lib/constants/NumericConstants.h
```diff
@@ -37,7 +37,7 @@ namespace GameConstants
 
 	constexpr int SKILL_QUANTITY=28;
 	constexpr int PRIMARY_SKILLS=4;
-	constexpr int RESOURCE_QUANTITY=8;
+	constexpr int RESOURCE_QUANTITY=7;
```
IvanSavenko, line 40: I assume you've checked all usages of this constant?

IvanSavenko, line 40: I assume you've checked all usages of this constant?

lib/entities/faction/CTownHandler.cpp
```diff
@@ -53,11 +53,10 @@ JsonNode readBuilding(CLegacyConfigParser & parser)
 	JsonNode ret;
 	JsonNode & cost = ret["cost"];
 
-	//note: this code will try to parse mithril as well but wil always return 0 for it
 	for(const std::string & resID : GameConstants::RESOURCE_NAMES)
 		cost[resID].Float() = parser.readNumber();
 
-	cost.Struct().erase("mithril"); // erase mithril to avoid confusing validator
+	parser.readNumber(); //note: will try to parse mithril -> needs mapping to resource from mithril mod
```
IvanSavenko: I think this line can be safely removed

lib/constants/NumericConstants.h
Laserlicht: Yes. Not much used anymore. Only legacy.

mapeditor/inspector/rewardswidget.cpp
Laserlicht: Mapeditor is completely out of scope of this PR. There are more things to do there. So yes. For now only native res.

lib/modding/ContentTypeHandler.h
Laserlicht: Because map is ordered. But resources needed to loaded first. Because other handlers needs full list of available resources.

See `// needs to loaded first because needed by other handlers`.

lib/json/JsonRandom.cpp
Laserlicht: For later PR. Rng stuff should keep as before (may conflicts with mine stuff, which is not supported yet).
Add comment...

lib/modding/ModManager.cpp
Laserlicht: Hack is needed -> See your `Not yet sure how to handle this nicely...`. With this it's working just fine.

Maybe we can merge and remove this hack later? If you have more time?
I'm not able to fix this properly.

lib/ResourceSet.cpp
Laserlicht: See other comment

lib/modding/ContentTypeHandler.h
Laserlicht: Checked again. ResourceHandler must loaded before Townhandler otherwise income of town is empty.

lib/modding/ModManager.cpp
IvanSavenko: Here is solution then:
Remove ResourceSet constructor from json and instead convert it into method, e.g. resolveFromJson or something.

Optionally for debugging delete copying constructor of ResourceSet - to let compiler show you where copy might take place using not yet resolved ResourceSet as source

We already had such problems for even more core system - bonus type. All resolved afaik.


Your current hack is very unstable and can easily break

lib/modding/ContentTypeHandler.h
IvanSavenko: Change no longer necessary

Laserlicht: Still required. Crashes on some places.

E.g. in CTownHandler (which is loaded before, if we change back):

```
				ret->produce[EGameResID::WOOD] = 1;
				ret->produce[EGameResID::ORE] = 1;
```

Crashes because there isn't any resource at this time. Everything from the handlers accessing the resources directly (no json).

IvanSavenko: Either remove that mod compatibility code or switch it to use identifier resolving

IvanSavenko: Oh. Or have initial size of resourceset to be either number of resources (if they were loaded) or 7 othervice (for h3 logic to work)

Laserlicht: Ah... found a solution...

Issue #2166, opened 2023-05-18
Can't start these maps (SoD, AB, RoE) 


GrafDeVan:
Basically, all maps that didn't run under SoD run in HotA. Here they are [Looks like maps from SoD but launched only in HOTA.zip](https://github.com/user-attachments/files/28949809/Looks.like.maps.from.SoD.but.launched.only.in.HOTA.zip) Even though they're marketed as SoD maps. I don't know what's wrong; maybe it's some WoG objects, HotA objects, or something else.

Only this SoD map won't start in Hota [SoD_Corbus_Realm_2.zip](https://github.com/user-attachments/files/28949964/SoD_Corbus_Realm_2.zip) Game just closes

> Old description
> 
> Hi. Some of these maps were launched a long time ago (SoDTheLastCrusade), some of them were never launched (C-_Shadows_of_the_Forest).
> The rest don't really work. And I don't know their behavior.
> [Maps.zip](https://github.com/vcmi/vcmi/files/11504939/Maps.zip)

Povelitelh3:
version is current develop 1.3.0.7a54273b7c22147b8898ed87d634a2f89387c09b mingw-ubuntu

IvanSavenko:
Checked all provided maps:
- Incorrect handling of Troll in army. Will be fixed.
- Heroes have war machines in the army. Will be removed on load.
- Heroes have "Ironfist of the Ogre" - unfinished artifact from Heroes 3. Will be removed on load.
- Instant loss since hero specified in loss condition exists only as campaign placeholder. H3 replaces loss condition with generic one.
- Instant loss since hero specified in loss condition is in prison. H3 applies condition only once imprisoned hero is freed. 
- Instant loss since AI owns target town on day 1 and victory condition applies to AI. H3 ignores AI who already owns town?

However, some maps are unlikely to be fixed. Specifically:
- Attainment - File is corrupted
- Battle - Contains WoG objects
- Crisis of Destiny - Contains WoG objects
- Shadows of the Forest - Object outside map bounds. Not supported at the moment.
- Rise of the Sun King - Game hangs on start in VCMI due to unknown reason. Crashes in Heroes 3.

rilian-la-te:
@IvanSavenko we should have an ability to replace OH3's Ironfist of the Ogre with Hota one, at least in editor.

Povelitelh3:
@IvanSavenko Added map which doesn't work in 1.6.3

[...]

PR #6148, opened 2025-09-18
Minor fixes for artifact removal during map load and handling null map objects

MichalZr6:
These changes should be the fix for:
- #2166

### DIFF
--- a/lib/mapping/MapFormatH3M.cpp
+++ b/lib/mapping/MapFormatH3M.cpp
@@ -925,8 +925,7 @@ void CMapLoaderH3M::loadArtifactsOfHero(CGHeroInstance * hero)
 		logGlobal->debug("Hero %d at %s has set artifacts twice (in map properties and on adventure map instance). Using the latter set...", hero->getHeroTypeID().getNum(), hero->anchorPos().toString());
 
 		hero->artifactsInBackpack.clear();
-		while(!hero->artifactsWorn.empty())
-			hero->removeArtifact(hero->artifactsWorn.begin()->first);
+		hero->artifactsWorn.clear();
 	}
 
 	for(int i = 0; i < features.artifactSlotsCount; i++)
--- a/mapeditor/mainwindow.cpp
+++ b/mapeditor/mainwindow.cpp
@@ -618,6 +618,9 @@ void MainWindow::saveMap()
 
 	for(auto obj : controller.map()->objects)
 	{
+		if(!obj)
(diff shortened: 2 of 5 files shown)


lib/mapping/MapFormatH3M.cpp
```diff
@@ -925,8 +925,7 @@ void CMapLoaderH3M::loadArtifactsOfHero(CGHeroInstance * hero)
 		logGlobal->debug("Hero %d at %s has set artifacts twice (in map properties and on adventure map instance). Using the latter set...", hero->getHeroTypeID().getNum(), hero->anchorPos().toString());
 
 		hero->artifactsInBackpack.clear();
-		while(!hero->artifactsWorn.empty())
-			hero->removeArtifact(hero->artifactsWorn.begin()->first);
```
IvanSavenko, line 929: This is wrong.

Heroes can have their artifacts configured in two places:
- globally in scenario options
- in hero instance on map

If hero has artifacts configured in both places, game should only use 2nd set, and discard 1st one

GeorgeK1ng:
This PR also fixes opening LOTR4.h3m from https://heroes3wog.net/lord-of-the-rings-scenario-map-for-heroes-3-complete-or-sod/ 
Current develop can't open it.

MichalZr6:
So you mean this code loads the wrong set?

I reverted that change and debugged. The issue was a spellbook artifact that was in the backpack. Converting it to `SPELLBOOK_PRESET` during load fixes the map/playability without touching `removeArtifact` but it is still unsafe to use because callbacks (gamestate or EditorCallback) don’t have a fully initialized map yet and `removeArtifact` calls `getMapConstPtr()` along its call stack. I will look into this again later today when I have more time.

```diff
@@ -966,6 +960,13 @@ bool CMapLoaderH3M::loadArtifactToSlot(CGHeroInstance * hero, int slot)
                return false;
        }
 
+       if(artifactID == ArtifactID::SPELLBOOK)
+       {
+               hero->spells.insert(SpellID::SPELLBOOK_PRESET);
+               logGlobal->warn("Map '%s': Converted spellbook from artifact for hero %s.", mapName, hero->getNameTranslated().c_str());
+               return false;
+       }
+
        if(art->isBig() && slot >= ArtifactPosition::BACKPACK_START)
        {
                logGlobal->warn("Map '%s': A big artifact (war machine) in hero's backpack, ignoring...", mapName);

PR #6156, opened 2025-09-19
adventure map config button generation

Laserlicht:
PR to avoid this mess: https://github.com/vcmi-mods/vcmi-extras/tree/vcmi-1.7/Mods/adventureMap/Content/sprites/buttons

It creates icons in similar style than original game.

Just let the AssetGenerator generate icons (in all colors/states for desired size) from one overlay image (32x32):
https://github.com/vcmi/vcmi/blob/develop/Mods/vcmi/Content/Sprites/adventureLayers.png

```
				{
					"type" : "adventureMapButton",
					"name" : "buttonABC",
					"image" : "buttons/iam-newbut.def",
					"help" : "vcmi.adventureMap.revisitObject",
					"hotkey" : "adventureVisitObject",
					"playerColored" : true,
					"generateFromBaseImage" : "adventureLayers",
					"generateSmall" : true,
					"area" : {
						"top" : 128,
						"left" : 32,
						"width" : 32,
						"height" : 32
					}
				},
```

`image` is the new virtual button file (no real files needed).
`generateFromBaseImage` is the filename of the overlay image (if not set, then no generation process -> same behavior than before)
`generateSmall` is to choose between 64x32 and 32x32

@GeorgeK1ng

### DIFF
--- a/client/adventureMap/AdventureMapWidget.cpp
+++ b/client/adventureMap/AdventureMapWidget.cpp
@@ -160,6 +160,7 @@ std::shared_ptr<CIntObject> AdventureMapWidget::buildMapButton(const JsonNode & 
 		auto assetGenerator = ENGINE->renderHandler().getAssetGenerator();
 		auto layout = assetGenerator->createAdventureMapButton(ImagePath::fromJson(input["generateFromBaseImage"]), small);
 		assetGenerator->addAnimationFile(AnimationPath::builtin("SPRITES/" + input["image"].String()), layout);
+		ENGINE->renderHandler().updateGeneratedAssets();
 	}
 
 	auto button = std::make_shared<CButton>(position.topLeft(), image, help, 0, EShortcut::NONE, playerColored);
--- a/client/adventureMap/AdventureMapWidget.cpp
+++ b/client/adventureMap/AdventureMapWidget.cpp
@@ -21,6 +21,7 @@
 #include "../GameInstance.h"
 #include "../gui/Shortcut.h"
 #include "../mapView/MapView.h"
+#include "../render/AssetGenerator.h"
 #include "../render/IImage.h"
 #include "../render/IRenderHandler.h"
 #include "../widgets/Buttons.h"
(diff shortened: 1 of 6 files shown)


GeorgeK1ng:
Can we use it somehow to generate custom buttons in same style? Like Marketplace, Dig, Boat, AirShip etc? Like from only icons for buttons? 

Or this simply can load basic button icon and convert it to proper buttons during load time?

Laserlicht:
> Can we use it somehow to generate custom buttons in same style? Like Marketplace, Dig, Boat, AirShip etc? Like from only icons for buttons?
> 
> Or this simply can load basic button icon and convert it to proper buttons during load time?

Not sure if I understand correctly...

With this PR you only need one picture for a new button.

You only need this image in sprites folder and the config above.
newbut
Everything else is done by magic. No messing with palette, multiple frames, multiple sizes...

GeorgeK1ng:
We will also need better configuration for adventureMap.json config. Because we now have ability to resize whole VCMI window, without needs to change resolution. I mean, we need now specify resolution and then buttons config. But would be better to have one calculated block which will show small / big based on defined buttons. For example when we have defined 15 buttons it means 15x32 height and when window is resized to smaller and now there can fit only 13 use small ones instead. But same must be done for Hero / Towns list. Dynamically adjust buttons and Hero / Towns list. Based on remaining space between map and bottom window.

GeorgeK1ng:
> > Can we use it somehow to generate custom buttons in same style? Like Marketplace, Dig, Boat, AirShip etc? Like from only icons for buttons?
> > Or this simply can load basic button icon and convert it to proper buttons during load time?
> 
> Not sure if I understand correctly...
> 
> With this PR you only need one picture for a new button.
> 
> You only need this image in sprites folder and the config above. newbut Everything else is done by magic. No messing with palette, multiple frames, multiple sizes...

Yes! Thats it! Awesome work!

GeorgeK1ng:
Anything else here @IvanSavenko? This will bring new small / big buttons for adventure map mostly to look like original ones.

PR #6173, opened 2025-09-25
Improve editor performance

Opuszek:
The code still needs cleaning and it needs to undergo more rigorous tests, but it is working, so far no new bugs found.
The editor works smoothly on a low-end PC on huge maps sizes. 
I will take a break from it for a couple of days before making it pr-worthy. For now, if anyone interested, here it is.


mapeditor/scenelayer.cpp
```diff
@@ -55,69 +67,280 @@ void AbstractLayer::redraw()
...
+		QPainter painter(&pixmap);
+		for(int j = 0; j <= pixmap.height(); j += tileSize)
+		{
+			for(int i = 0; i < pixmap.width(); i += tileSize)
 			{
-				painter.fillRect(i * 32, j * 32, 31, 31, tl.visitable() ? QColor(200, 200, 0, 64) : QColor(255, 0, 0, 64));
+				auto tl = map->getTile(int3(section.x()/tileSize + i/tileSize, section.y()/tileSize + j/tileSize, scene->level));
+				if(tl.blocked() || tl.visitable())
+				{
+					painter.fillRect(i, j, 31, 31, tl.visitable() ? QColor(200, 200, 0, 64) : QColor(255, 0, 0, 64));
```
DjWarmonger, line 336: Please define these colors in one place instead of using ad-hoc magic numbers.

Opuszek: I would do it, if I had a smallest idea what those colors are.
This is (and all QColor as well, since the draft is concerned with performance) already existing line that has been moved from one place to another.
I will take a look whether it can be easily systematized, but I don't want to fall into a problem singularity. The commit is big as it is.

mapeditor/inspector/inspector.cpp
```diff
@@ -376,19 +376,22 @@ void Inspector::updateProperties(CGArtifact * o)
...
-				if(controller.map()->allowedSpells.count(spell->id) != 0)
-					delegate->options.push_back({QObject::tr(spell->getNameTranslated().c_str()), QVariant::fromValue(int(spell->getId()))});
+				auto * delegate = new InspectorDelegate;
+				for(auto const & spell : LIBRARY->spellh->objects)
+				{
+					if(controller.map()->allowedSpells.count(spell->id) != 0)
+						delegate->options.push_back({QObject::tr(spell->getNameTranslated().c_str()), QVariant::fromValue(int(spell->getId()))});
+				}
+				addProperty(QObject::tr("Spell"), LIBRARY->spellh->getById(spellId)->getNameTranslated(), delegate, false);
 			}
```
C3pa: It would be nice if this didn't indent as much. For example:
```c++
if(o->ID != Obj::SPELL_SCROLL) return;

const CArtifactInstance * instance = o->getArtifactInstance();
if (!instance) return;

SpellID spellId = instance->getScrollSpellID();
if(spellId == SpellID::NONE) return;

auto * delegate = new InspectorDelegate;
for(auto const & spell : LIBRARY->spellh->objects)
{
	if(controller.map()->allowedSpells.count(spell->id) != 0)
		delegate->options.push_back({QObject::tr(spell->getNameTranslated().c_str()), QVariant::fromValue(int(spell->getId()))});
}
addProperty(QObject::tr("Spell"), LIBRARY->spellh->getById(spellId)->getNameTranslated(), delegate, false);
```

mapeditor/scenelayer.h
```diff
@@ -28,56 +28,101 @@ class AbstractLayer : public QObject
...
 
-class GridLayer: public AbstractLayer
+	void update();
+	void redraw();
+protected:
+	void redraw(const std::vector<int3> & tiles);
+	void redrawWithSurroundingTiles(const std::vector<int3> & tiles);
+	void redraw(const std::set<CGObjectInstance *> & objects);
+	void redraw(const std::vector<QRectF> & areas);
+	QRectF getObjectArea(const CGObjectInstance * object);
```
C3pa: Can this method be marked as const?

mapeditor/scenelayer.h
```diff
@@ -28,56 +28,101 @@ class AbstractLayer : public QObject
...
+private:
+	void addSector(QGraphicsItem * item);
+	void removeSector(QGraphicsItem * item);
+	void redrawSectors(std::set<QGraphicsItem *> & items);
+	std::set<QGraphicsItem *> getContainingSectors(const std::vector<int3> & tiles);
+	std::set<QGraphicsItem *> getIntersectingSectors(const std::vector<QRectF> & areas);
```
C3pa: Same here.

mapeditor/inspector/inspector.cpp
Opuszek: I've removed this code completely, as it was only ad hoc fix to an unrelated bug that was making testing difficult. Bug will need to be fixed anyway, but one thing at a time.

mapeditor/scenelayer.h
Opuszek: I set methods:
- getObjectArea
- getAllSectors
- getContainingSectors
- getIntersectingSectors
to const. All the "redraw" methods change the object.
Thanks for the advice.

mapeditor/scenelayer.cpp
```diff
@@ -55,65 +72,284 @@ void AbstractLayer::redraw()
...
 	if(!map)
 		return;
-	
-	pixmap.reset(new QPixmap(map->width * 32, map->height * 32));
-	pixmap->fill(Qt::transparent);
-	
-	QPainter painter(pixmap.get());
-	for(int j = 0; j < map->height; ++j)
+
+	pixmap = std::make_unique<QPixmap>(map->width * 32, map->height * 32);
```
IvanSavenko, line 283: Would also be great to remove these massive pixmap's which cause absurdly high memory usage when loading large maps.

Not insisting, and not for this PR, so up to you

IvanSavenko (approved):

mapeditor/scenelayer.cpp
Opuszek: It is already done.
This is the only map-sized pixmap left, and it is EmptyLayer - it is not filled and does not take much memory. It only exist so the user have something to move the screen on, when all the other layers are generated dynamically to fit it.
A 252x252 randomly generated map took about ~700mb of my RAM.

kambala-decapitator:
> Here is the list

please try to split commits accordingly, will be easier to review

mapeditor/mapview.cpp
```diff
@@ -655,23 +665,25 @@ void MapView::dragLeaveEvent(QDragLeaveEvent * event)
...
+
 	controller->discardObject(sc->level);
 }
 
-
-bool MapView::viewportEvent(QEvent *event)
+void MapView::setViewports()
 {
-	if(auto * sc = static_cast<MapScene*>(scene()))
+	if(auto * sc = dynamic_cast<MapScene*>(scene()))
```
kambala-decapitator, line 678: can it really have another type?

mapeditor/scenelayer.cpp
```diff
@@ -55,65 +72,284 @@ void AbstractLayer::redraw()
...
+	QRectF bounds((minX - 2) * tileSize, (minY - 2) * tileSize, (maxX - minX + 4) * tileSize, (maxY - minY + 4) * tileSize);	//tiles start with 1, QRectF from 0
+	redraw({bounds});
+}
+
+void AbstractViewportLayer::redraw(const std::set<CGObjectInstance *> & objects)
 {
+	std::vector<QRectF> areas(objects.size());
+	for (const CGObjectInstance * object : objects)
+	{
+		areas.push_back(getObjectArea(object));
```
kambala-decapitator, line 176: `areas` already is of size `objects.size()`, and now it becomes of double size. I think you wanted to create vector like
```cpp
std::vector<QRectF> areas;
areas.reserve(objects.size());
```

mapeditor/scenelayer.cpp
```diff
@@ -55,65 +72,284 @@ void AbstractLayer::redraw()
...
+
+void AbstractViewportLayer::addSector(QGraphicsItem * sector)
+{
+	items->addToGroup(sector);
+}
+
+void AbstractViewportLayer::removeSector(QGraphicsItem * sector)
+{
+	items->removeFromGroup(sector);
+	delete sector;
```
kambala-decapitator, line 214: feels very easy to forget in other place and get a memory leak

mapeditor/scenelayer.cpp
```diff
@@ -55,65 +72,284 @@ void AbstractLayer::redraw()
...
+		}
+	}
+	for (QGraphicsItem * sectorToRemove : sectorsToRemove)
+	{
+		addSector(draw(sectorToRemove->sceneBoundingRect()));
+		removeSector(sectorToRemove);
+	}
+}
+
+const QList<QGraphicsItem *> AbstractViewportLayer::getAllSectors() const	//returning const is necessary to avoid "range-loop might detach Qt container" problem
```
kambala-decapitator, line 236: it's easier to use `std::as_const` or `qAsConst` at the calling site

mapeditor/scenelayer.cpp
kambala-decapitator: > The commit is big as it is

and this is the problem. Making 20 small commits even together with colors refactoring would be easy to review.

mapeditor/mapview.cpp
Opuszek: You have a point, no.
I changed it because the compiler did not like it and displayed a warning.

mapeditor/scenelayer.cpp
Opuszek: Yep, you are right, I misunderstood how the vector's constructor works.
I will add fixes in a separate commit.

mapeditor/scenelayer.cpp
Opuszek: That's why I created a separate method for it and I do not call removeFromGroup anywhere else in the code.
QGraphicsItemGroup uses raw pointers and I am not sure if I can do anything with it.

mapeditor/scenelayer.cpp
Opuszek: Maybe, but then you need to do it at all calling sides.
IMO from a reader perspective getAllSectors() is a better read than qAsConst(items->childItems()).
If somebody is interested about the implementation they can always look into the method.

PR #6177, opened 2025-09-26
fix list in creature window

Laserlicht:
fix list issue reported on discord

grafik


GeorgeK1ng:
Still exist for Commanders window
image

Also artifacts places have bad background color when used "left/right scroll" on ability bonuses.

Here is correct artifacts background.
image

PR #6180, opened 2025-09-28
[macOS] fix deploying package

kambala-decapitator:
- Editor depends on additional Qt libs/modules now, ensure that `macdeployqt` fetches them too
- *deployqt failure should fail cpack/install process (and whole build)


kambala-decapitator:
verified: editor can be launched on macOS again

PR #6185, opened 2025-09-29
Use performNativeCopy instead of copy

GeorgeK1ng:
This should fix all issues with spaces in filenames on Android.

1) Changed performNativeCopy to bool
2) Narroved getRealPath to same logic with path encodig using safeEncode
3) Replaced QFile::copy with Helper::performNativeCopy where possible. Now remains only in prepare_android.cpp - when I tried to replace it here too, Launcher crashed on load without any error. So keeping this file as is.

Now EXE / ZIP with spaces can be safely imported on Android devices using Launcher.

### DIFF
--- a/launcher/helper.cpp
+++ b/launcher/helper.cpp
@@ -42,6 +42,19 @@ static QScrollerProperties generateScrollerProperties()
 }
 #endif
 
+#ifdef VCMI_ANDROID
+static QString safeEncode(QString uri)
+{
+	// %-encode unencoded parts of string.
+	// This is needed because Qt returns a mixed content url with %-encoded and unencoded parts. On Android >= 13 this causes problems reading these files, when using spaces and unicode characters in folder or filename.
+	// Only these should be encoded (other typically %-encoded chars should not be encoded because this leads to errors).
+	// Related, but seems not completly fixed (at least in our setup): https://bugreports.qt.io/browse/QTBUG-114435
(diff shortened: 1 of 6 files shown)


GeorgeK1ng:
@Laserlicht I have tried to use your advice from here: https://github.com/vcmi/vcmi/pull/6087#issuecomment-3343063471.
So only in prepare_android.cpp remains QFile::copy (when used helped there, launcher crashed without any info on Android), all other instances were replaced with performNativeCopy to be sure.

But I still can't import ZIP named "the_mysterious_island_2809 (1).zip" space issue is still there. So problem seems is not only in QFile::copy.

Laserlicht:
> @Laserlicht 
> But I still can't import ZIP named "the_mysterious_island_2809 (1).zip" space issue is still there. So problem seems is not only in QFile::copy.

Possible. You could use logcat and search for error. Last time there were displayed some errors while opening file.

GeorgeK1ng:
Fixed, issue was in getRealPath, required same touch as is done in performNativeCopy.

I have tried to import "Heroes Chronicles AIO.exe" which was just renamed from original filename setup_heroes_chronicles_2.0.0.38.exe. And got same issue, nothing was done after selection from import. Now works just fine.

So any ZIP / EXE with spaces should work now for manual import. Both were installed now correctly on my Android tablet.

PR #6204, opened 2025-10-03
fix deserializion of ResourceSet for compatibility

Laserlicht:


lib/ResourceSet.h
```diff
@@ -214,7 +214,25 @@ class ResourceSet
 
 	template <typename Handler> void serialize(Handler &h)
 	{
-		h & container;
+		if (h.version >= Handler::Version::CONFIGURABLE_RESOURCES)
+			h & container;
+		else
+		{
+			if (h.saving)
```
IvanSavenko, line 221: Ok to merge, but there is no need to implement saving of old save format. We only need to load old save

IvanSavenko (approved):

PR #6209, opened 2025-10-04
Kingdomoverview additional mines

Laserlicht:
Support additional mines in kingdom overview.

Using `kingdomOverviewImage` parameter.

Scrollbar only visible if there are more resources than standard.

[Peek 2025-10-04 17-52.webm](https://github.com/user-attachments/assets/5e29e531-72fd-4a79-bdfa-f98dadb52dab)

Example:
```
{
	"core:mine" : {
		"types" : {
			"mithrilMine" : {
				"name" : "Mithril mine",
				"description" : "{Mithril mine}\n\nYou gain control of a Mithrilmine. It will provide you with 2 mithril per day.",
				"kingdomOverviewImage" : "mithril/OVMINES_Mithril",
				"aiValue" : 1500,
				"rmg" : {
					"value" : 1500
				},
				"sounds" : {
					"ambient" : ["LOOPLUMB"]
				},
				"resource" : "mithril",
				"defaultQuantity": 2,
				"templates" : {
					"avwrhsmt" : {
						"animation" : "mithril/AVMGOSW1",
						"mask" : [
							"VVVV",
							"VBAB"
						],
						"visitableFrom" : [
							"---",
							"+-+",
							"+++"
						]
					}
				}
			}
		}
	}
}
```
[OVMINES_Mithril.def.zip](https://github.com/user-attachments/files/22702655/OVMINES_Mithril.def.zip)

### DIFF
--- a/client/windows/CKingdomInterface.cpp
+++ b/client/windows/CKingdomInterface.cpp
@@ -648,15 +648,16 @@ void CKingdomInterface::generateMinesList(const std::vector<const CGObjectInstan
 
 	for(int i=0; i<GameConstants::RESOURCE_QUANTITY; i++)
 	{
-		std::string value = std::to_string(minesCount[i]);
+		int resID = line * GameConstants::RESOURCE_QUANTITY + i;
+		if(resID >= LIBRARY->resourceTypeHandler->getAllObjects().size())
+			break;
+
+		std::string value = std::to_string(minesCount[resID]);
 		std::shared_ptr<InfoBoxCustom> data;
 		if(line == 0)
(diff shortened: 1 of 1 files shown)


GeorgeK1ng:
I have just added MIthril to WoG mod, but seems here needs little but move numbers down?
image

First line looks normal, but 1-2 px moving down would be good too.
image

Laserlicht:
> I have just added MIthril to WoG mod, but seems here needs little but move numbers down? image
> 
> First line looks normal, but 1-2 px moving down would be good too. image

Maybe incorrect image size?

See def file above which is working correctly.

GeorgeK1ng:
Oh,yes it's 68x54. I have resized it by mistake to 68x45. Fixed image size and it's better now. Thanks!

image

Issue #5471, opened 2025-02-24
Possibility to search random map template by typing name template


misiokles:
When we have many templates installed and we want to quickly find given template, we must scroll whole template list to search it. 
Maybe make some space 'template search-able'?

Proposal.

![Image](https://github.com/user-attachments/assets/fa8ca870-6cf6-4658-8064-29ff671aa24b)

GeorgeK1ng:
There should be also working templates filter based on selection. So on G+U will be visible in list only compatible templates.

GeorgeK1ng:
Done in 1.7

PR #6211, opened 2025-10-05
Allowing editing town name

Laserlicht:
Allows to edit town name ingame in town screen. Similar to player name in lobby. For easier remembering. Coloring also possible with color syntax.

Contains also refactoring of player name edit.

grafik

### DIFF
--- a/server/NetPacksServer.cpp
+++ b/server/NetPacksServer.cpp
@@ -365,7 +365,6 @@ void ApplyGhNetPackVisitor::visitSetFormation(SetFormation & pack)
 void ApplyGhNetPackVisitor::visitSetTownName(SetTownName & pack)
 {
 	gh.throwIfWrongOwner(connection, &pack, pack.tid);
-	gh.throwIfPlayerNotActive(connection, &pack);
 
 	result = gh.setTownName(pack.tid, pack.name);
 }


Laserlicht:
Heroes would be an idea for later, but not sure about biographies!?

SoundSSGood:
I previously heard that a feature for changing town names was needed in the editor. But is this really needed in runtime in game?

IvanSavenko:
Yeah, definitely not high-priority feature, especially considering growing list of open issues

Laserlicht:
> Yeah, definitely not high-priority feature, especially considering growing list of open issues

Yes. Of course. Is basically a small function to support you remember your town name. Which is useful for lists, e.g. teleport. 

Also i wanted to refactor the code for the texbox I introduced two years ago. With a little bit more expirience. :)

Problem on the issue list is that most things are way more complicated. For example game mechanic changes. Or bugs that are hard to find. Not just a simple UI thing.

misiokles:
@Laserlicht how about this?
https://github.com/vcmi/vcmi/issues/5471

server/NetPacksServer.cpp
```diff
@@ -362,6 +362,14 @@ void ApplyGhNetPackVisitor::visitSetFormation(SetFormation & pack)
 	result = gh.setFormation(pack.hid, pack.formation);
 }
 
+void ApplyGhNetPackVisitor::visitSetTownName(SetTownName & pack)
+{
+	gh.throwIfWrongOwner(connection, &pack, pack.tid);
+	gh.throwIfPlayerNotActive(connection, &pack);
```
IvanSavenko: Can be allowed I guess? Don't think what could go wrong if we allow player to rename town while the player is waiting for his turn.

client/windows/CCastleInterface.cpp
```diff
@@ -1435,7 +1436,15 @@ CCastleInterface::CCastleInterface(const CGTownInstance * Town, const CGTownInst
 	garr->setRedrawParent(true);
 
 	heroes = std::make_shared<HeroSlots>(town, Point(241, 387), Point(241, 483), garr, true);
-	title = std::make_shared<CLabel>(85, 387, FONT_MEDIUM, ETextAlignment::TOPLEFT, Colors::WHITE, town->getNameTranslated());
+	title = std::make_shared<CTextInputWithConfirm>(Rect(83, 386, 140, 20), FONT_MEDIUM, ETextAlignment::TOPLEFT, town->getNameTranslated(), true, [this](){ 
+		std::string name = title->getText();
+		std::string originalName = LIBRARY->generaltexth->translate(town->getNameTextID());
+		if(name == originalName)
+			name = ""; // use textID again
+		GAME->interface()->cb->setTownName(town, name);
```
IvanSavenko, line 1444: Perhaps also handle empty string as request to reset to original name?

Laserlicht: Should already the case. `name` is coming from textbox. If it is empty then it will be similar than `if(name == originalName)`. And empty means it's loading it from translator.

PR #6216, opened 2025-10-07
template search

Laserlicht:
Just quick test draft for @misiokles

Please check if you like it that way (reusing existing stuff...)

Related #5471

### DIFF
--- a/client/lobby/RandomMapTab.cpp
+++ b/client/lobby/RandomMapTab.cpp
@@ -141,16 +141,18 @@ RandomMapTab::RandomMapTab():
 	//set combo box callbacks
 	if(auto w = widget<ComboBox>("templateList"))
 	{
-		w->onConstructItems = [](std::vector<const void *> & curItems){
+		auto getTemplates = [](){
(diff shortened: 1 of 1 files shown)


misiokles:
It wasn't easy to check this artifact for Windows because of this.
image
...and yes, I like this style
image
Of course, blue background would be better, as consistent with other windows, but it's handy!

Laserlicht:
> It wasn't easy to check this artifact for Windows because of this.

Strange. On every develop build?

> Of course, blue background would be better, as consistent with other windows, but it's handy!

Blue background isn't easy. Because background of this window is a static image.

misiokles:
No, developer builds install fine. Just this artifact build from your commit woke up mi windows defender ;)
As brown background - let it stay as it is. Functionality is most important.

IvanSavenko:
How this interacts with previously existing drop-down box we had for template selection?

Laserlicht:
> How this interacts with previously existing drop-down box we had for template selection?

It basically replacing the popup of drop down box (list will open instead of drop down). But keep the box itself visually the same.

PR #6217, opened 2025-10-08
PROPOSAL: Add Patch Category for Mods

Zaratuir:
I would like to suggest adding "Patch" as a category for modType. This would differ from "Compatibility" in that it would not be hidden by the UI. This would be for optional feature patches between mods. For example, typically Fused Upgrades converts combination artifacts into fused artifacts. Some mods add Combination artifacts such as Third Upgrades Mod adding the Magic Compendium. For balance reasons, players may not want the combination artifacts from certain mods to become fused artifacts. By using the category "Compatibility", this choice is taken away from players and it is forcibly applied under the hood if Fused Artifacts adds a compatibility mod for Third Upgrades. This proposal allows patches that can be added either as submods or completely independently that can be toggled and managed by the user in the mod manager, while still allowing other critical compatibility fixes to remain transparent to the player.

### DIFF
--- a/config/schemas/mod.json
+++ b/config/schemas/mod.json
@@ -65,7 +65,7 @@
 		},
 		"modType" : {
 			"type" : "string",
-			"enum" : [ "Translation", "Town", "Test", "Templates", "Spells", "Music", "Maps", "Sounds", "Skills", "Other", "Objects", "Mechanics", "Interface", "Heroes", "Graphical", "Expansion", "Creatures", "Compatibility", "Campaigns", "Artifacts", "AI", "Resources" ],
+			"enum" : [ "Translation", "Town", "Test", "Templates", "Spells", "Music", "Maps", "Sounds", "Skills", "Other", "Objects", "Mechanics", "Interface", "Heroes", "Graphical", "Expansion", "Creatures", "Compatibility", "Campaigns", "Artifacts", "AI", "Resources", "Patch" ],
 			"description" : "Type of mod, e.g. Town, Artifacts, Graphical."
 		},
 		"author" : {
--- a/docs/modders/Mod_File_Format.md
+++ b/docs/modders/Mod_File_Format.md
@@ -31,7 +31,7 @@
 
 	// Type of mod, list of all possible values:
 	// "Translation", "Town", "Test", "Templates", "Spells", "Music", "Maps", "Sounds", "Skills", "Other", "Objects", 
-	// "Mechanics", "Interface", "Heroes", "Graphical", "Expansion", "Creatures", "Compatibility", "Campaigns", "Artifacts", "AI", "Resources"
+	// "Mechanics", "Interface", "Heroes", "Graphical", "Expansion", "Creatures", "Compatibility", "Campaigns", "Artifacts", "AI", "Resources", "Patch"
 	//
 	// Some mod types have additional effects on your mod:
 	// Translation: mod of this type is only active if player uses base language of this mod. See "language" property. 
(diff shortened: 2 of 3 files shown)


misiokles:
I don't understand the problem. In such cases such submods can be labelled as Artifacts (for example). It leads for other problems, but what you propose is available for years.

Zaratuir:
You are correct that this is available by utilizing the other categories as you mentioned. The purpose of this would be to provide clear delineation for end users between a mod that adds new content vs an optional mod that does not add any new content, but provides some overlap for existing mods. I agree that this is not a strictly necessary change as it's not an unsolvable problem at the moment, which is why I labeled this as a proposal. Feel free to deny it if you feel that it's not a useful distinction.

GeorgeK1ng:
Don't we have names and descriptions for that purpose? Compatibility is there to force activate it and keep it hidden. You can use any mod type with "depends" to make mod toogable when needed if mod in "depends" is installed. Fused artifacts / Market of Time should always work for all possible artifacts / skills across mods which make it as real feature. There is no need to add something like mod type "Patch" in my eyes.

Zaratuir:
I disagree. The purpose of modding as an end user is to tweak the game in various ways into the ideal version that you enjoy playing, not to play someone else's vision of the game. Continuing with the fused artifacts example, someone may release a mod that has a combination artifact that, for balance purposes, should take up multiple slots even with the existence of fused artifacts. Some players may prefer to keep the authors vision of the components and some players may not care and want the new modded artifact to be fused as well, but that's a choice that should be left to the end user, not forced under the scenes by the mod author.

Another example where a patch would be a good category would be a third party mod that resolves overlap between two mods. For example, there are two different mods that add functionally almost identical Vampirism spells. A mod that disables one of them would be a patch that doesn't add any additional content and as such wouldn't really make sense in the spells category. While that's not as necessary in the current state of VCMI with a fairly controlled mod repository, if the team ever intends to extend support for other modding platforms like Nexus, it would become critical.

All of this said, these are my philosophies as a modder and mod author. As I'm new to this particular modding community, I will happily defer to the senior modders here.

misiokles:
We understand these issues. But all of them are all the time doable. Still modder can create a mod called "Vampirism patch" with label "Mechanical" or "Other" (not "Compatibility"). So any player who launches launcher and browses mods in blink of the eye has hint what this mod does. For example New Magic Balance works this way. Or pvp-balance.

GeorgeK1ng:
If you need to disable other mod by your one, there is "conflicts" section which do it, of course it must be correctly set in boths mod sides. Otherwise create "depends" or "softDepends" and touch and modify other mod. In similar way is done in Fused artifacts. 

"Internal" GitHub mods repo is required otherwise VCMI can easily goes to tons of crashes. Platform is under development and mod compatibility is removed in time from code. And another point is modders publish their mods and mostly never update them to latest required standards, which means they are not usable in new VCMI versions. And from I saw in past bad user experience commes from not updated / outdated mods.

Current mods categories covers probably all possible things. And many mods in repository have correctly configured "depends" / "softDepends" / "conflicts".

So if you wanna to create some other kind of for example Fused artifacts which targets to Game balance, it's good to check how it's done huge Game balance mod.

Which 2 mods adds Vampirism spell? It should be fixed in both mods itself to deliver best game experience in build-in launcher to don't duplicate them.

Zaratuir:
I do understand this perspective. To me, modding is inherently unstable and it's up to the end user to understand what they're putting in their game so as to handle conflicts, not the mod platform to enforce compatibility. That said, I again defer to the senior community members in this regard. As far as what mods add Vampirism, both Tides of War and New Old Spells add a Vampirism spell with extremely similar functionality.

GeorgeK1ng:
From what I see, we have already solved this duplicate in 1.7. In any case we will turn it to different ban based solution.

PR #6221, opened 2025-10-11
Fix buffered stream growth & decode river type from low 3 bits (compat for “A War” AB map)

MichalZr6:
- CBufferedStream::ensureSize: grow by the needed bytes (need = target - size), min step still 512, but no overshoot; avoids huge reallocs when only 1 byte is required.

- H3M reader (rivers): interpret river type as the low 3 bits (raw & 0x07; valid 0..4). Upper bits are editor flags (e.g., HotA).

- Prevents misreads/over-allocation and stops invalid river index crashes; with these changes the community map “A War” now opens and runs.

### DIFF
--- a/lib/mapping/MapReaderH3M.cpp
+++ b/lib/mapping/MapReaderH3M.cpp
@@ -199,9 +199,11 @@ RoadId MapReaderH3M::readRoad()
 
 RiverId MapReaderH3M::readRiver()
 {
-	RiverId result(readInt8());
-	assert(result.getNum() <= features.riversCount);
-	return result;
+	const uint8_t raw = readInt8();
+	// Keep low 3 bits as river type (0..4); discard high-bit flags set by some editors (HotA ?)
+	const uint8_t type = raw & 0x07;
+	assert(type <= features.riversCount);
+	return RiverId(type);
 }
 
 PrimarySkill MapReaderH3M::readPrimary()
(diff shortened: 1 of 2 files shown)


lib/mapping/MapReaderH3M.cpp
```diff
@@ -199,9 +199,11 @@ RoadId MapReaderH3M::readRoad()
 
 RiverId MapReaderH3M::readRiver()
 {
-	RiverId result(readInt8());
-	assert(result.getNum() <= features.riversCount);
-	return result;
+	const uint8_t raw = readInt8();
+	// Keep low 3 bits as river type (0..4); discard high-bit flags set by some editors (HotA ?)
```
IvanSavenko, line 203: Any idea what these flags are for? How many bits are used for them?

MichalZr6: I used AI to solve this. I asked how did it know that:

> Short answer: it’s not from an official spec. It’s an inference from hands-on hex diffs + behavior across tools.
> 
> How I got there
> 
> Take a clean AB map → hex dump tile bytes.
> 
> Open/save in HotA editor (no edits) → hex dump again.
> 
> Diff shows some river/road bytes get high bits set while low 3 bits stay identical. Re-saving in the original editor tends to clear those highs. That pattern repeats across multiple maps.

Haven't changed any roads-related-code yet, but it might be I will come across this again when testing other community maps.

PR #6226, opened 2025-10-12
show invisible

Laserlicht:
`showInvisible` command to show events and grail (only single player) for debugging help.

grafik

### DIFF
--- a/client/mapView/MapRenderer.cpp
+++ b/client/mapView/MapRenderer.cpp
@@ -27,6 +27,7 @@
 #include "../render/Graphics.h"
 
 #include "../../lib/CConfigHandler.h"
+#include "../../lib/callback/CCallback.h"
 #include "../../lib/gameState/CGameState.h"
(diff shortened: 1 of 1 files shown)


IvanSavenko:
You can already see events via showVisitable & client-side anticheat attempts are rather useless, especially in open-source.

Perhaps drop it? Or better - check for that "cheats enabled" flag we have already.

I would also extend it to all show xxx commands

Laserlicht:
> You can already see events via showVisitable & client-side anticheat attempts are rather useless, especially in open-source.

What do you mean exactly? `showInvisible` is not for anti cheat. It's a different command for events and grail.

IvanSavenko:
Why showInvisible is blocked in multiplayer? I guess to prevent cheating, right?

But players can easily circumvent that check in multiple ways

So I suggest to:
a) remove this check entirely or
b) use existing 'allow cheats' flag

And probably use such logic for all showXXX commands

PR #6227, opened 2025-10-12
[windows] use msvc v142 toolset for Intel builds

kambala-decapitator:
- msvc Intel builds are made with v142 toolset (v14.29) now for proper Windows 7 support, Conan changes for that: https://github.com/vcmi/vcmi-dependencies/pull/15
- x86 build is made using the x64 tools (amd64_x86)
- removed explicit CL setting that's not required, compiler cache still works
- reduced copy-paste of `conanrun.ps1` script invocation in the CI flow
- small improvements to the user-friendly platform names in the CI summary script

### DIFF
--- a/docs/developers/Conan.md
+++ b/docs/developers/Conan.md
@@ -31,7 +31,7 @@ The following platforms are supported and known to work, others might require ch
 1. Check if your build environment can use the prebuilt binaries: basically, that your compiler version (or Xcode major version) matches the information below. If you're unsure, simply advance to the next step.
     - *macOS*: libraries are built with Apple clang 16 (Xcode 16.2), should be consumable by Xcode / Xcode CLT 16.x and later
     - *iOS*: libraries are built with Apple clang 16 (Xcode 16.2), should be consumable by Xcode 16.x and later
-    - *Windows*: libraries are built with MSVC 19.4x (v143 toolset)
+    - *Windows*: libraries are built with MSVC 19.29 (v142 toolset, but can be consumed by v143) for Intel and with MSVC 19.4x (v143 toolset) for ARM64
     - *Android*: libraries are built with NDK r25c (25.2.9519653)
 
 2. Download the binaries archive from <https://github.com/vcmi/vcmi-dependencies/releases> (pre-release is for development version and the latest release is for respective VCMI release) and use `conan cache restore <path to archive>` command to unpack them.


.github/workflows/github.yml
```diff
@@ -292,11 +292,7 @@ jobs:
       run: |
         & conan-generated\conanrun.ps1
 
-        $CL = "$($env:VCToolsInstallDir)/bin/${{ matrix.cl }}"
-        cmake `
-          -D "CMAKE_C_COMPILER:FILEPATH=$CL" `
-          -D "CMAKE_CXX_COMPILER:FILEPATH=$CL" `
```
kambala-decapitator, line 298: @GeorgeK1ng do you remember the reason of setting compilers explicitly in #6055? Seems to work just fine without it.

kambala-decapitator:
@GeorgeK1ng windows builds were made with v142 toolset, please check if they work on w7 now

.github/workflows/github.yml
GeorgeK1ng: Cache work too without them? I remember it was needed to add them, otherwise it was not used. If works now without them, then it's fine to remove. But I think there were issue with them, because we are on x64 runner, we need x86 CL for x86, otherwise sccache was not working. Not sure if same issue is still there.

.github/workflows/github.yml
```diff
@@ -66,10 +67,10 @@ jobs:
             conan_prebuilts: dependencies-windows-x64
             conan_options: -s "&:build_type=RelWithDebInfo" -c tools.env.virtualenv:powershell=pwsh -o "&:target_pre_windows10=True"
             artifact_platform: x64
-            cl: Hostx64/x64/cl.exe
 
           - platform: msvc-x86
-            arch: x86
+            arch: amd64_x86
```
GeorgeK1ng, line 72: This will probably break installer building and it's working. 

EDIT: Oh, will work as we have now installer step independent on compile matrix.

Just PY for report needs to be updated with exact names for table headers to have there correctly x86. Same must be done for Android x64. But not sure, why it was not used there: emit_partial.py - https://github.com/vcmi/vcmi/blob/e81d5bc9730ed599c72152408ceada94c05130c0/CI/emit_partial.py#L89

Can be solved later, nothing crucial.

image

.github/workflows/github.yml
GeorgeK1ng: Can you please to rerun to see if there will be hits on PR cache?

.github/workflows/github.yml
kambala-decapitator: thanks, will update the script

.github/workflows/github.yml
kambala-decapitator: cache works

GeorgeK1ng: Great, will check artifacts imports later today to see how they looks on pure W7 SP1. Hope none will be missing.
142 usage should be safe for all newer.

GeorgeK1ng:
@Laserlicht Can you please test ARM64 build from this PR?

kambala-decapitator:
arm64 is built with v143, no change here

GeorgeK1ng:
I have just performed now check under Windows 7 SP1 x64 on all compiled files. 

No missing imports except **vccorlib140.dll** which uses 3 Window 8+ APIs imports from:
_api-ms-win-core-winrt-error-l1-1-0.dll
api-ms-win-core-winrt-l1-1-0.dll
api-ms-win-core-winrt-string-l1-1-0.dll_

This is probably caused by incorrectly configured NTDDI_WIN7 (0x06010000) for this file somewhere?

I have checked if they can be found somewhere for Windows 7, but these 3 are only available in Windows 8.0 setup media, have not found them in any redists. And even with that they are for 6.2 in header, but that probably doesn't matter as latest Windows 7 UCRT Redists have Windows 10 headers lock - from what I know, this is known issue in drivers or excutables, DLLs should not be affected by these locks. These 3 missing imports can be also just some forgotten leftover, because I was unable to locate them in Windows Kits 8.0, 8.1, and 10.19041.

In any case I can tell VCMI looks stable and have not faced any errors or crashes.

kambala-decapitator:
not really a windows dev, so I may be asking stupid questions below :)

> **vccorlib140.dll** which uses 3 Window 8+ APIs imports from

so is it a problem for win7? or it can require those dlls without actually using win8 APIs? can you find out which exe/dll it's coming from exactly?

> This is probably caused by incorrectly configured NTDDI_WIN7 (0x06010000) for this file somewhere?

we inject win7 macros to everything that's built via [Conan profile](https://github.com/vcmi/vcmi-dependencies/blob/main/conan_profiles/base/msvc-intel), but it may be possible that lib header(s) redefine that

> latest Windows 7 UCRT Redists have Windows 10 headers lock

you mean that they were built with win10 SDK? but is it a problem?

GeorgeK1ng:
I can only tell imports from vccorlib140.dll want to load these 3 files. But I don't think it's important same as UCRT builded with older Windows 10 SDK. Only in latest v143 MS did some stupid changes.

That macro can be also bugged somewhere else. I have also checked missing API files grabbed from W8 under W7 and they have no missing imports as they are UCRT ones, we can add them to our UCRT redist package to cover these gaps to be safe.

But no need to hastle with them for now from my point of view.

Lets merge :)

IvanSavenko:
No objections, but why was Android platform name changed from x86_64 to Intel x64?

kambala-decapitator:
> No objections, but why was Android platform name changed from x86_64 to Intel x64?

@GeorgeK1ng wanted a name that's similar to other Intel platforms. Keep it or revert?

GeorgeK1ng:
Just x64 should be enough, we can't quess if it's Intel, AMD or any other vendor.

IvanSavenko:
Google calls it x86_64, like we had before: https://developer.android.com/ndk/guides/abis

Either form is fine for me though

kambala-decapitator:
let's stick to the traditional x86_64 then

PR #6231, opened 2025-10-16
[windows] generate bat files that use conanrun.ps1

kambala-decapitator:
- now it's no longer required to run `conan install` twice when using Powershell
- paths to the `conanrun` script and to the exe are quoted to avoid issues with spaces


kambala-decapitator:
@GeorgeK1ng any idea why there were no cache hits on this run?

GeorgeK1ng:
Second run will hits fine. Cache tries to get our main branch cache "develop" one, but our caches takes too much space and github delete them automatically. So if we develop more and merge to branch often it will then hits normally.

Issue #2976, opened 2023-09-28
The program freezes at the stage of copying game data.


JoeD4ddy:
**Describe the bug**
iPad tablet 12.9 m2
At the second step of launching the program after selecting the language, you need to specify the path to the game files; when you click the copy data button, the application freezes. If an app's screen is minimized and maximized, the maximized app will be at a distorted screen resolution and inactive. Thus, it is impossible to specify the path to the game files.

**Game logs**

[VCMI_Launcher_log.txt](https://github.com/vcmi/vcmi/files/12753055/VCMI_Launcher_log.txt)

**Version**
 - OS: iOS 16,7
 - Version: VCMI-branch-develop-47a17cc.ipa (stable too)

Issue #1845, opened 2023-04-02
Launcher "freezes" while copying H3 data


kambala-decapitator:
UI/UX improvement is needed: if showing proper copy progress is impossible, at least "indefinite progress bar" is needed. Would also be good to show which file is being copied.

IvanSavenko:
Kind of related to #1787

rilian-la-te:
@IvanSavenko we can download in another thread and just show non-determinable progressbar for downloading.

IvanSavenko:
Downloading is already done in async mode (uses some Qt routines from what I remember)
However mod archive extraction and H3 data copying (apparently) happens in main thread.

Issue #6186, opened 2025-09-29
Rework launcher first launch


GeorgeK1ng:
I think we should spend some time on rebuilding first launch.

1) Language page

- Remove "Have a question" section, this should not be on first page, nobody will ever click on it.
- Keep it simple - "Thank you for installing VCMI, please choose your launguage"
- Replace language names with flags or turn them into buttons Flag + Language name.
It should look like welcome message with language buttons at bottom and next button

2) Heroes III Data
- Rename to "Game data"
- Completely redesign this page with "How would you like to set up VCMI?" with 3 radio buttons
     1.  Defaul GOG.com installer selected - simplify selection just one file exe / bin. Both must be and they have same name when downloaded. Maybe on mobiles try to scan default downloads folder for "setup_heroes_of_might_and_magic_3*.exe" without needs to select files when they are found. In auto scan should be also option for "setup_heroes_chronicles*.exe" and allow their import during first launch. When not found, only EXE / BIN files should be allowed to select and also visible in "file picker"
     2.  Copy existing data (Be sure it works on Android with spaces in path, ZIPed game should be allowed in this step too) - Display requirement SoD / Complete game there
     3.  Manual installation      - each radio buttons should reveal their info after selecting option, there is no need tell user all information at once. So path to app is visible only when clicked on 3. radio
- Display warning about not supported HD version there on bottom instead in any selection

3) Mods Preset
- Rename to "Recommended mods"
- Remove "Heroes III translation" button, this should be already done by option from first page, download automatically translation from initial selection. English should be downloaded to, fixes are applied here from time to time
- Add Tides of War button, for near future add Fallen Of the Depth button. Improve current Horn of the Abyss / In The Wake of Gods descriptions.
- Add Icons for each mod for selection to button

4) Finish
- "Done, everything is configured. Wait for mods installation. Recommend user to check settings"
- Show "Have a question" from first page there instead

This is only initial concept, needs to be designed in Qt Creator. Hope this will make initial setup easier for everyone. Second page is harder to quickly design and show initial idea.

<img width="762" height="404" alt="Image" src="https://github.com/user-attachments/assets/8d801803-31d9-449c-a88f-a97f0c6bfd5f" />

<img width="764" height="402" alt="Image" src="https://github.com/user-attachments/assets/ee2cb7dc-586a-447b-8feb-657593208eac" />

[...]

Issue #2304, opened 2023-07-08
iOS - launcher layout on small devices


Laserlicht:
**Describe the bug**
The launcher layout is not good on small devices. Font is too big and destroys the layout. Dark mode is not correct.

The most difficult is the setup (unfortunately I don't have a screenshot). Since there is relatively much text.

Is there a way to reduce font size on such devices?

**Screenshots**
Example on Phone SE 2020:
![IMG_1707](https://github.com/vcmi/vcmi/assets/13953785/ccdbcc18-520a-40b9-a62b-648390500465)
![aa](https://github.com/vcmi/vcmi/assets/13953785/47ea5c24-a02f-4e83-aac6-8210e2950ce9)

**Version**
 - OS: iOS
 - Version: develop

IvanSavenko:
Looks like we need to decrease not just text size but rather entirely scale down the app, including icons on left side...

I actually created setup layout based on Launcher screenshot from iOS, but it was probably another, larger, device.
- mods list - we should probably switch layout to 2 tabs on small screens - tab with mod list & tab with mod info
- settings - hiding navigation should work
- setup - perhaps make it scrollable, similar to settings window? Can't think of any other solutions.

Laserlicht:
I don't know how many users would want to use VCMI on such a small device. But it is the worst case scenario.

- mod list: the main issue is, that you cannot scroll the right part with swiping (for the list itself: it's working)
- settings: yes, the navigation should be hidden on small devices
- setup: Yes: Scrollable. And I think we can remove Discord, Slack and Github here? We have an help button now.

And by the way: There is currently no way to get back to the setup from the help menu (except closing the app).

DjWarmonger:
Whole Qt GUI needs to be scalable, it does not take ito account GUI scaling on Windows either.

kambala-decapitator:
> Whole Qt GUI needs to be scalable, it does not take ito account GUI scaling on Windows either.

it does, but not out of the box, see https://doc.qt.io/qt-5/highdpi.html

PR #6233, opened 2025-10-17
[1.7.2] Improve files import in first launch

GeorgeK1ng:
1) Added unified import overlay with progress bar
2) Added Tides of Wars to recommended mods
3) Added Fallen of the Depth to recommended mods, when it's released newjoiners will automatically see that button in first launch
4) Hidden buttons "sidePanel" during first launch
5) Fixed "Copy existing files" option, should now work in all scenarios - spaces, scpecial character in folder name
6) Moved GOG option as first one
7) Added option to display "Demo support" mod in recommended mods when player wanna to use "merged Windows + MacOS demo" - https://github.com/GeorgeK1ng/heroes3demo. It's safe to use VCMI with them, but that mods must be enabled. Of course it's limited gameplay. Button is visible only, when detected demo.
8) Added Fail-Fast for EXE / BIN selection. EXE is now choosed first to perform required checks

This PR partialy solves https://github.com/vcmi/vcmi/issues/6186

For future PR:
1) Make initial screen only with language selection
2) Improve data import screen with radios and dynamic text info when selection various import options
3) Remove "Navigation tabs" and drive initial wizard only with next / back buttons
4) Address all other common ideas from #6186 

How it looks on [Redmi 6](https://www.gsmarena.com/xiaomi_redmi_6-9237.php) with Android 9
Screenshot_2025-10-17-11-04-40-877_is xyz vcmi daily
Screenshot_2025-10-17-11-04-59-378_is xyz vcmi daily
Screenshot_2025-10-17-11-05-23-355_is xyz vcmi daily

Demo + Language buttons visible
Screenshot_2025-10-17-11-06-09-587_is xyz vcmi daily

How it looks on [Galaxy Tab S9 FE](https://www.gsmarena.com/samsung_galaxy_tab_s9_fe-12517.php) with Android 15
https://github.com/user-attachments/assets/d5700507-cb3c-4ac9-9563-bdba19c1db05

Fixes: https://github.com/vcmi/vcmi/issues/2976
Fixes: https://github.com/vcmi/vcmi/issues/1845


IvanSavenko:
As I said before - perhaps avoid words like "exclusive"? Not a big fan of exclusivity & vcmi was created (among other things) to make h3 available not only on Windows.

For example

Install the underground, mushroom-themed expansion Fallen of the Depth, created for VCMI

GeorgeK1ng:
Yes, but Fallen of the Depth going to be VCMI exclusive expansion - available only in VCMI engine of course on all operating systems, but require VCMI. That what we want to tell.  But yeah, if you really want, we can drop "exclusive" word.

IvanSavenko:
ProgressOverlay - we generally create layout ib Qt Designer / .ui files, not directly in c++ code. Any reason why we need to have it directly in c++ here?

IvanSavenko:
Exclusive - not insisting, but IMO softer version would be better. Like created for / designed for. 

I would rather add few more words to explain what this mod is about other than underground themed

GeorgeK1ng:
> ProgressOverlay - we generally create layout ib Qt Designer / .ui files, not directly in c++ code. Any reason why we need to have it directly in c++ here?

I tried to avoid too much touches to UI files and additional elements show / hide. My touches to UI files were done manually, because my Qt creator always change too much in UI files. 

I know I will need to touch UI files when I will add standalone language selection page + finish one. Im open to move it to UI files with more UI changes. 

EDIT: Just tried to put Widget in Qt creator and seems it always want to move other elements positions..

GeorgeK1ng:
@kambala-decapitator Can you please test artifacts on iOS / MacOS?

kambala-decapitator:
will check today or tomorrow

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
 
         return fileName;
     }
+
+	// Build a flat list of copy tasks for a SAF tree selection.
```
kambala-decapitator: sorry, what is SAF tree?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	// Output rows are "srcUri\tTarget\tName" where Target is one of Data/Maps/Mp3 decided by file extension.
+	// Heuristic: if user picked "Data" and we found no Maps/Mp3, also scan ../Maps and ../Mp3 one level up.
+	public static String[] findFilesForCopy(String treeUriStr, android.content.Context ctx) {
+	    // Build a DocumentFile tree from the SAF tree URI
+	    final android.net.Uri treeUri = android.net.Uri.parse(treeUriStr);
```
kambala-decapitator: it's much better to `import` types rather than using their full name

You can also use `var` for type auto-deduction, like `auto` in C++

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	    final androidx.documentfile.provider.DocumentFile root =  androidx.documentfile.provider.DocumentFile.fromTreeUri(ctx, treeUri);
+	    if (root == null) return new String[0];
+	
+	    final java.util.ArrayList<String> out = new java.util.ArrayList<>();
+	    int foundMaps = 0, foundMp3 = 0;
+	
+	    // 1) Traverse the selected tree, collect files by extension
+	    {
+	        final java.util.ArrayDeque<androidx.documentfile.provider.DocumentFile> stack = new java.util.ArrayDeque<>();
```
kambala-decapitator: if you need a stack, use `java.util.Stack`, otherwise call the variable appropriately

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	        stack.push(root);
+	
+	        while (!stack.isEmpty()) {
+	            final androidx.documentfile.provider.DocumentFile d = stack.pop();
```
kambala-decapitator: what is `d`?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	            final androidx.documentfile.provider.DocumentFile[] kids = d.listFiles();
```
kambala-decapitator: files? children? "directory contents"?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	            if (kids == null) continue;
+	
+	            for (androidx.documentfile.provider.DocumentFile f : kids) {
+	                if (f.isDirectory()) { stack.push(f); continue; }
```
kambala-decapitator: please don't place statements on the same line. Condition/loop body should be on the new line.

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	                    target = "Data";
+	                else if (lower.endsWith(".h3m"))
+	                    target = "Maps";
+	                else if (lower.endsWith(".mp3"))
+	                    target = "Mp3";
+	
+	                if (target != null) {
+	                    out.add(f.getUri().toString() + "\t" + target + "\t" + name);
+	                    if ("Maps".equals(target)) foundMaps++;
+	                    if ("Mp3".equals(target))  foundMp3++;
```
kambala-decapitator: - copy-pasted string literals
- why operator `==` can't be used to compare strings?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	                }
+	            }
+	        }
+	    }
+	
+	    // 2) If user picked the "Data" folder itself and Maps/Mp3 were not found, also scan siblings ../Maps and ../Mp3 one level up (heuristic).
+	    try {
+	        final String display = root.getName(); // tree top display name
+	        if (display != null && display.equalsIgnoreCase("Data") && (foundMaps == 0 || foundMp3 == 0)) {
```
kambala-decapitator: is it actually possible to select `Data` directory and have maps or mp3 inside?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	            // Compute parent tree docId: trim last path segment of current treeDocId
+	            final String treeDocId = android.provider.DocumentsContract.getTreeDocumentId(treeUri);
+	            final int colon = (treeDocId != null) ? treeDocId.indexOf(':') : -1;
+	            final String type = (colon < 0) ? treeDocId : treeDocId.substring(0, colon);
+	            final String rel  = (colon < 0) ? ""         : treeDocId.substring(colon + 1);
```
kambala-decapitator: you've tried to make internal indent with spaces but failed :)

doesn't string have a `split` method?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	                String target = null;
+	                if (lower.endsWith(".lod") || lower.endsWith(".snd") || lower.endsWith(".vid") || lower.endsWith(".pak"))
+	                    target = "Data";
+	                else if (lower.endsWith(".h3m"))
+	                    target = "Maps";
+	                else if (lower.endsWith(".mp3"))
+	                    target = "Mp3";
+	
+	                if (target != null) {
+	                    out.add(f.getUri().toString() + "\t" + target + "\t" + name);
```
kambala-decapitator: it's better to use format strings:
```suggestion
	                    out.add("%s\t%s\t%s".formatted(f.getUri().toString(), target, name));
```

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	
+	            if (parentRoot != null) {
+	                // Helper to find a child folder by (case-insensitive) name under parentRoot
+	                java.util.function.Function<String, androidx.documentfile.provider.DocumentFile> childByName =
+	                        (folderName) -> {
+	                            final androidx.documentfile.provider.DocumentFile[] kids = parentRoot.listFiles();
+	                            if (kids == null) return null;
+	                            for (androidx.documentfile.provider.DocumentFile f : kids) {
+	                                if (f.isDirectory()) {
+	                                    final String nm = f.getName();
```
kambala-decapitator: what is `nm`?

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	            final int slash = rel.lastIndexOf('/');
+	            final String parentRel = (slash >= 0) ? rel.substring(0, slash) : "";
+	            final String parentDocId = type + ":" + parentRel;
+	
+	            final android.net.Uri parentTree = android.provider.DocumentsContract.buildTreeDocumentUri(treeUri.getAuthority(), parentDocId);
+	            final androidx.documentfile.provider.DocumentFile parentRoot = androidx.documentfile.provider.DocumentFile.fromTreeUri(ctx, parentTree);
+	
+	            if (parentRoot != null) {
+	                // Helper to find a child folder by (case-insensitive) name under parentRoot
+	                java.util.function.Function<String, androidx.documentfile.provider.DocumentFile> childByName =
```
kambala-decapitator: since the helper function isn't a couple of lines long, it's better to extract it to a dedicated function. The method is already quite big.

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	                                if (f.isDirectory()) {
+	                                    final String nm = f.getName();
+	                                    if (nm != null && nm.equalsIgnoreCase(folderName)) return f;
+	                                }
+	                            }
+	                            return null;
+	                        };
+	
+	                // Append files from ../Maps (*.h3m)
+	                if (foundMaps == 0) {
```
kambala-decapitator: useless check, you've already made it above

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	                            for (androidx.documentfile.provider.DocumentFile f : kids) {
+	                                if (f.isDirectory()) { st.push(f); continue; }
+	                                final String name = f.getName();
+	                                if (name == null) continue;
+	                                if (name.toLowerCase(java.util.Locale.ROOT).endsWith(".h3m")) {
+	                                    out.add(f.getUri().toString() + "\tMaps\t" + name);
+	                                    foundMaps++;
+	                                }
+	                            }
+	                        }
```
kambala-decapitator: looks almost identical to traversing Data directory

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	                                final String name = f.getName();
+	                                if (name == null) continue;
+	                                if (name.toLowerCase(java.util.Locale.ROOT).endsWith(".mp3")) {
+	                                    out.add(f.getUri().toString() + "\tMp3\t" + name);
+	                                    foundMp3++;
+	                                }
+	                            }
+	                        }
+	                    }
+	                }
```
kambala-decapitator: absolutely identical to the maps part

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -152,4 +152,136 @@ private static String getFilenameFromUri(String sourceFileUri, Context context)
...
+	            }
+	        }
+	    } catch (Throwable ignore) {
```
kambala-decapitator: is there actually anything that can throw?

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -25,19 +25,83 @@
...
+#endif
+
+#ifdef VCMI_ANDROID
 #include <QtAndroid>
+#endif
 
-static FirstLaunchView * thiz;
-extern "C" JNIEXPORT void JNICALL Java_eu_vcmi_vcmi_NativeMethods_heroesDataUpdate(JNIEnv * env, jclass cls)
+// Unified progress overlay
+class ProgressOverlay final : public QWidget
```
kambala-decapitator: the class is big, please extract it to separate files. Also don't keep all method definitions in the header.

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -25,19 +25,83 @@
...
+        : QWidget(parent)
+    {
+        // Match parent background to feel native
+        setAutoFillBackground(true);
+        QPalette pal = palette();
+        pal.setColor(QPalette::Window, parent->palette().color(QPalette::Window));
+        setPalette(pal);
+
+        // Cover whole view except top offset
+        setGeometry(parent->rect().adjusted(0, topOffsetPx, 0, 0));
```
kambala-decapitator: what does this top offset represent? safe area on mobiles?

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -25,19 +25,83 @@
...
+
+        auto *layout = new QVBoxLayout(this);
+        layout->setContentsMargins(24, 24, 24, 24);
+        layout->setSpacing(12);
+
+        titleLabel = new QLabel("", this);
```
kambala-decapitator:
```suggestion
        titleLabel = new QLabel(this);
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -25,19 +25,83 @@
...
+        titleLabel->setAlignment(Qt::AlignCenter);
+        titleLabel->setStyleSheet("font-size: 16px; font-weight: 600;");
+
+        fileLabel = new QLabel("", this);
+        fileLabel->setAlignment(Qt::AlignCenter);
+        fileLabel->setWordWrap(true);
+
+        progressBar = new QProgressBar(this);
+        progressBar->setMinimumHeight(18);
+        setIndeterminate(true); // start indeterminate
```
kambala-decapitator: useless comment

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -25,19 +25,83 @@
...
+    ~ProgressOverlay() override
+    {
+        Helper::keepScreenOn(false);
+    }
+
+    void setTitle(const QString &t)        { titleLabel->setText(t); }
+    void setFileName(const QString &name)  { fileLabel->setText(name); }
+
+    // Switch between indeterminate and determinate progress
+    void setIndeterminate(bool on)         { on ? progressBar->setRange(0,0) : progressBar->setRange(0,100); }
```
kambala-decapitator: the ternary operator could be moved inside the second parameter

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -25,19 +25,83 @@
...
+    void setRange(int max)                 { progressBar->setRange(0, qMax(1, max)); progressBar->setValue(0); }
+    void setValue(int v)                   { progressBar->setValue(v); }
+
+private:
+    QLabel *titleLabel = nullptr;
+    QLabel *fileLabel = nullptr;
+    QProgressBar *progressBar = nullptr;
```
kambala-decapitator: nullptr is useless in this case

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -25,19 +25,83 @@
...
+};
+
+// Create and show overlay immediately
+static inline ProgressOverlay* createOverlay(QWidget *parent, const QString &title, bool indeterminate = true)
```
kambala-decapitator: don't see any point in having it inline

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
 	});
 #endif
 }
 
-void FirstLaunchView::extractGogDataAsync(QString filePathBin, QString filePathExe)
+bool performCopyFlow(const QString &path, FirstLaunchView *self, ProgressOverlay *overlay, bool removeSource = false)
 {
-	logGlobal->info("Extracting gog data from '%s' and '%s'", filePathBin.toStdString(), filePathExe.toStdString());
+    // 1) Scan -> "src \t Target \t Name"
+    overlay->setTitle(QObject::tr("Scanning selected folder..."));
```
kambala-decapitator: if you make this function a method, then you won't need to use `QObject::` and when translating this string the context will be clear

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
+        return false;
+    }
+
+    // 2) Validate signature
+	// TODO: Find proper way for pure SoD check in import or way to block pure RoE / AB
+	//       Or prepare RoE / AB Ban mod and allow VCMI to with any H3 version
+    auto validate = [](const QStringList &items)->QString {
+        bool anyLOD=false;
+		bool anySOD=false;
+		bool anyHD=false;
```
kambala-decapitator: something's very wrong with your indentation in this function

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
-		return QString();
-	};
-
-	QString filterBin = tr("GOG data") + " (*.bin)";
-	QString filterExe = tr("GOG installer") + " (*.exe)";
-
-	QDir tempDir(pathToQString(VCMIDirs::get().userDataPath()));
-	if(tempDir.cd("tmp"))
+            const auto part = line.split('\t');
+            if(part.size() < 3 || part[1].compare("Data", Qt::CaseInsensitive) != 0)
```
kambala-decapitator: size check seems useless as you always form those strings properly

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
+    for(const QString &line : items)
 	{
-		if(isGogGalaxyExe(tmpFileExe))
-		{
-			logGlobal->info("Gog Galaxy detected! Aborting...");
-			errorText = tr("You've provided a GOG Galaxy installer! This file doesn't contain the game. Please download the offline backup game installer!");
-		}
-	}
+        const auto part = line.split('\t');
+        if(part.size() < 3)
```
kambala-decapitator: this check again

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
-		logGlobal->info("Extraction done!");
-	}
+        const QString &src  = part[0];
+        const QString &tgt  = part[1]; // Data / Maps / Mp3
+        const QString &file = part[2];
 
-	QString hashError;
-	if(!errorText.isEmpty())
-		hashError = Innoextract::getHashError(tmpFileExe, tmpFileBin, filePathExe, filePathBin);
+        if(tgt.compare("Data", Qt::CaseInsensitive)!=0 && tgt.compare("Maps", Qt::CaseInsensitive)!=0 && tgt.compare("Mp3",  Qt::CaseInsensitive)!=0)
```
kambala-decapitator: I don't think it will ever be true

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
-	QStringList dirData = sourceRoot.entryList({"data"}, QDir::Filter::Dirs);
-	QStringList dirMaps = sourceRoot.entryList({"maps"}, QDir::Filter::Dirs);
-	QStringList dirMp3 = sourceRoot.entryList({"mp3"}, QDir::Filter::Dirs);
+        const QDir dstDir = targetRoot.filePath(tgt);
+        plan.push_back({ src, dstDir.filePath(file) });
+    }
 
-	const auto noDataMessage = tr("Failed to detect valid Heroes III data in chosen directory.\nPlease select the directory with installed Heroes III data.");
-	if(dirData.empty())
+    if(plan.isEmpty())
```
kambala-decapitator: you have already detected that H3 installation is a valid SoD, this will never be true

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
-	copyDirectories += dirData.front();
-	if (!dirMaps.empty())
-		copyDirectories += dirMaps.front();
+    // 5) Optional cleanup
+    if(removeSource)
+        QDir(path).removeRecursively();
 
-	if (!dirMp3.empty())
-		copyDirectories += dirMp3.front();
+    overlay->deleteLater();
```
kambala-decapitator: you call this every time, rather let the caller destroy the dialog it passed (and most likely created)

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
-		QDir targetDir = targetRoot.filePath(dirName);
+#ifdef ENABLE_INNOEXTRACT
+    //  Show overlay immediately so the UI doesn't look frozen
+    auto *overlay = createOverlay(this, tr("Checking installer..."), true);
+    overlay->setFileName(QFileInfo(filePathExe).fileName());
+    overlay->raise();
+    qApp->processEvents();
+
+    // Defer heavy work to next event-loop tick to ensure overlay is painted
+    QTimer::singleShot(0, this, [this, overlay, filePathBin, filePathExe]() mutable
```
kambala-decapitator: what's the purpose of `mutable`?

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
+            {
+                logGlobal->info("Invalid file selected: %s", filter.toStdString());
+                return QObject::tr("You have to select %1 file!", "param is file extension").arg(filter);
+            }
+
+            logGlobal->info("Checking file %s", filename.toStdString());
+            return QString();
+        };
+
+        QString errorText{};
```
kambala-decapitator:
```suggestion
        QString errorText;
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
+            quint64 fileSize = file.size();
+
+            if(fileSize > 10 * 1024 * 1024)
+                return false; // avoid loading big files; Galaxy exe is smaller...
+
+            if(!file.open(QIODevice::ReadOnly))
+                return false;
+
+            QByteArray data = file.readAll();
+            const QByteArray magicId{reinterpret_cast<const char*>(u"GOG Galaxy"), 20};
```
kambala-decapitator:
```suggestion
            constexpr std::u16string_view galaxyID = u"GOG Galaxy";
            const auto galaxyIDBytes = reinterpret_cast<const char*>(galaxyID.data());
            const auto magicId = QByteArray::fromRawData(galaxyIDBytes, galaxyID.size() * sizeof(decltype(galaxyID)::value_type));
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
+{
+    logGlobal->info("Extracting gog data from '%s' and '%s'", filePathBin.toStdString(), filePathExe.toStdString());
 
-	for(const QString & dirName : copyDirectories)
-	{
-		QDir sourceDir = sourceRoot.filePath(dirName);
-		QDir targetDir = targetRoot.filePath(dirName);
+#ifdef ENABLE_INNOEXTRACT
+    //  Show overlay immediately so the UI doesn't look frozen
+    auto *overlay = createOverlay(this, tr("Checking installer..."), true);
```
kambala-decapitator: what's the point of creating the overlay outside the actual handler? If you move it there, then you can use `QScopedPointer` to avoid manual call to `deleteLater`, as the handler is synchronous.

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
+        }
+    });
+#endif
+}
 
-		if(!targetRoot.exists(dirName))
-			targetRoot.mkdir(dirName);
+void FirstLaunchView::copyHeroesData(const QString &path, bool removeSource)
+{
+    auto *overlay = createOverlay(this, tr("Scanning selected folder..."), true);
```
kambala-decapitator: same point about the handler and QScopedPointer

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -338,248 +402,317 @@ void FirstLaunchView::extractGogData()
...
-			if(move)
-				sourceFile.rename(targetDir.filePath(filename));
-			else
-				sourceFile.copy(targetDir.filePath(filename));
-		}
-	}
+    auto work = [this, path, overlay]() {
+        if(performCopyFlow(path, this, overlay, false))
+            if(heroesDataUpdate())
+                activateTabModPreset();
```
kambala-decapitator: the above function has exactly the same logic just with `true` parameter

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -592,16 +725,22 @@ void FirstLaunchView::modPresetUpdate()
...
 	ui->labelPresetExtrasDescr->setVisible(checkCanInstallExtras());
+	ui->labelPresetDemoDescr->setVisible(checkCanInstallDemo());
 	ui->labelPresetHotaDescr->setVisible(checkCanInstallHota());
 	ui->labelPresetWogDescr->setVisible(checkCanInstallWog());
+	ui->labelPresetTowDescr->setVisible(checkCanInstallTow());
+	ui->labelPresetFodDescr->setVisible(checkCanInstallFod());
 
 	// we can't install anything - either repository checkout is off or all recommended mods are already installed
-	if (!checkCanInstallTranslation() && !checkCanInstallExtras() && !checkCanInstallHota() && !checkCanInstallWog())
+	if (!checkCanInstallTranslation() && !checkCanInstallExtras() && !checkCanInstallDemo() && !checkCanInstallHota() && !checkCanInstallWog() && !checkCanInstallTow() && !checkCanInstallFod())
```
kambala-decapitator: please compute the "check" values once and store them in constants

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -627,19 +766,59 @@ bool FirstLaunchView::checkCanInstallTranslation()
...
+
+    QDir userRoot = pathToQString(VCMIDirs::get().userDataPath());
+    QDir dataDir(userRoot.filePath(QStringLiteral("Data")));
+	QDir mapsDir(userRoot.filePath(QStringLiteral("Maps")));
+
+    bool hasDemoMap = false;
+    QStringList mapFiles = mapsDir.entryList(QDir::Files | QDir::Readable);
+    for(const QString &name : mapFiles)
+        if(name.compare(QStringLiteral("h3demo.h3m"), Qt::CaseInsensitive) == 0)
+            hasDemoMap = true;
```
kambala-decapitator: add `break`

you could also use `std::any_of` to simplify the code

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -627,19 +766,59 @@ bool FirstLaunchView::checkCanInstallTranslation()
...
+	
+    QStringList files = dataDir.entryList(QDir::Files | QDir::Readable);
+    for(const QString &name : files)
+    {
+        if(name.compare(QStringLiteral("H3ab_spr.lod"), Qt::CaseInsensitive) == 0)
+        {
+            QFile lod(dataDir.filePath(name));
```
kambala-decapitator: since you only need file size, use QFileInfo

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -627,19 +766,59 @@ bool FirstLaunchView::checkCanInstallTranslation()
...
+            quint64 fileSize = lod.size();
+			logGlobal->error("H3ab_spr.lod size: %s", static_cast<unsigned long long>(fileSize));
```
kambala-decapitator: - `%s` is for strings, most likely you need `%llu`
- unclear why the cast is needed, it's already the same type

launcher/helper.cpp
```diff
@@ -102,16 +105,42 @@ QString getRealPath(QString path)
...
+            jSrc.object<jstring>(),
+            jDst.object<jstring>(),
+            QtAndroid::androidContext().object()
+        );
+        return QFileInfo(dst).exists();
+    }
+
+    // Pure filesystem -> use Qt copy
+    if (QFile::exists(dst))
+        QFile::remove(dst);
```
kambala-decapitator: you can safely remove without the check

launcher/helper.cpp
```diff
@@ -102,16 +105,42 @@ QString getRealPath(QString path)
...
+	
+    return QFile::copy(src, dst);
 #else
-	return QFile::copy(src, dst);
+    QFileInfo dstInfo(dst);
+    QDir().mkpath(dstInfo.absolutePath());
+    if (QFile::exists(dst))
+        QFile::remove(dst);
+	
+    return QFile::copy(src, dst);
```
kambala-decapitator: code in the second android part is identical, simply move it out of the `#if`

launcher/helper.cpp
```diff
@@ -145,4 +173,157 @@ void keepScreenOn(bool isEnabled)
...
 #endif
 }
+
+#ifdef VCMI_ANDROID
+static constexpr int  kFolderPickerReqCode = 4242;
+
+// Intent flags: READ | WRITE | PERSISTABLE | PREFIX
+static constexpr jint kIntentFlags        = 1 | 2 | 64 | 128;
+
+class FolderPickReceiver final : public QAndroidActivityResultReceiver
```
kambala-decapitator, line 196: I don't get why not doing it in Java and avoiding all this ugly JNI stuff and copy-pasted constants (btw you could also retrieve them by name via JNI call). Java will signal to native code when it's done.

launcher/helper.cpp
```diff
@@ -145,4 +173,157 @@ void keepScreenOn(bool isEnabled)
...
+{
+public:
+    std::function<void(QString)> onDone;
+
+    // One-shot result handler for ACTION_OPEN_DOCUMENT_TREE
+    void handleActivityResult(int req, int res, const QAndroidJniObject &data) override
+    {
+        auto cb = std::move(onDone); // guarantee single-use
```
kambala-decapitator: it doesn't really guarantee anything. If you call `onDone` once more from any place by mistake or not, you'll get UB.

launcher/helper.cpp
```diff
@@ -145,4 +173,157 @@ void keepScreenOn(bool isEnabled)
...
+
+        // Bounce back to Qt thread
+        QMetaObject::invokeMethod(qApp, [cb, pickedTree]{ if (cb) cb(pickedTree); }, Qt::QueuedConnection);
+    }
+};
+
+static FolderPickReceiver g_receiver;
+#endif // VCMI_ANDROID
+
+void nativeFolderPicker(QWidget *parent, std::function<void(QString)> cb)
```
kambala-decapitator:
```suggestion
void nativeFolderPicker(QWidget *parent, std::function<void(QString)>&& cb)
```

launcher/helper.cpp
```diff
@@ -145,4 +173,157 @@ void keepScreenOn(bool isEnabled)
...
+    QtAndroid::startActivity(intent, kFolderPickerReqCode, &g_receiver);
+
+#elif defined(VCMI_IOS)
+    SelectDirectory iosDirectorySelector;
+    const QString dir = iosDirectorySelector.getExistingDirectory();
+    if (cb) cb(dir);
+
+#else
+    const QString dir = QFileDialog::getExistingDirectory(parent, {}, {}, QFileDialog::ShowDirsOnly | QFileDialog::DontResolveSymlinks);
+    if (cb) cb(dir);
```
kambala-decapitator: if `cb` is "false", then it makes no sense to select any directory at all, otherwise you select a directory for nothing

launcher/helper.cpp
```diff
@@ -145,4 +173,157 @@ void keepScreenOn(bool isEnabled)
...
+            return out;
+
+        QAndroidJniEnvironment env;
+        const jobjectArray arr = static_cast<jobjectArray>(jArr.object<jobject>());
+        const jsize n = env->GetArrayLength(arr);
+        out.reserve(n);
+        for (jsize i = 0; i < n; ++i)
+        {
+            QAndroidJniObject s((jstring)env->GetObjectArrayElement(arr, i));
+            out.push_back(s.toString());                                                       // "src \t Target \t Name"
```
kambala-decapitator: not enough whitespace, add more xD

kambala-decapitator (changes requested):
FYI it took me more than 2 hours to rip through this huge wall of text.

In PR description you state 7 different changes, but everything is basically stuffed in a single commit. In any serious project such PR would be immediately blocked or rejected without even looking at the changes until the author makes it actually reviewable.

Please stop using web UI to upload files and learn git basics, it will help you structure your work as well. There're nice GUI clients for git out there. Web UI is ok when you need to change a couple of lines, but not for big changes.

Using LLM to help you write code is also absolutely fine. But in 99% of cases you can't just throw in what it generated you and leave it like that: you must read it thoroughly and refactor to make it readable and comprehendable.

To conclude, the current PR state is a PoC, but not something that can be merged.

Some general notes:
- code style remarks that I made apply to more than a single place, you should review everything and adjust accordingly
- the PR has a lot of duplicated raw literals (strings / numbers) that should be constants
- project's code style is ignored completely. Use at least an IDE / editor that understands EditorConfig file to indent properly, you can also use clang format via `git-clang-format` command in terminal to reformat only changed lines
- big functions should be divided into smaller ones

GeorgeK1ng:
@kambala-decapitator Thanks for 40+ things to review. I can't use anything like git for now. Only web UI. I understand you work in code, me neither. I'm just enthusiast, not professional coder. 

I don't see any PR like mine to address first time user experience, current code works and is well tested on various devices. Mostly we need to know if artifacts works on Mac and iOS too. Windows and Android are safe and works fine in tons of scenarios.

Do you think I will commit change UI which broke code? Or how do you expect whole copy / import files logic commited in 122 web UI tries? I did my best to solve some of current issues and bring much better user experience for first launch which this project require a lot currently. 

There are currently just few options - use what is already in PR and note it must be improved in future. And will be needed in any case as not all cases to make first launch more user friendly were not addressed. Can be done by someone else - more experienced, but I don't see legions of coders in VCMI who wanna to bring something in this area.

Or simply close this PR, I can do some changes, but not going to address your more than 40 comments. If you can do that better and improve code style, you always can do that. But I can't do it much better. I'm happy it works and brings expected results.

If anyone can do it currently better, lets do it. Otherwise would be better to go with merge, do rest of required user visible changes in another PR and then do code style tweaking to keep newly introduced features.

@IvanSavenko It's up to you, of course there are places to improve, they are always there. Sorry, but I can't do testing like 40+ required changes now. And when area to test is mostly Android it's even more complicated for me.

kambala-decapitator:
I'm not going to argue about this, but it's not a healthy approach in an OSS project

Laserlicht:
Just for interest: Do you also have tested an smaller device?

I think it could be problematic because there too much elements (e.g. mods: 6 elements with long texts).
We already have problems with 4 mods: https://github.com/vcmi/vcmi/issues/2304
Maybe make scroll able to avoid this problem?

GeorgeK1ng:
I don't have anything smaller than old [Redmi 6](https://www.gsmarena.com/xiaomi_redmi_6-9237.php) but it's little bit smaller than mentioned [iPhone SE 2020](https://www.gsmarena.com/apple_iphone_se_(2020)-10170.php). But I can't test it on iPhone.

And sidePanel is now hidden during first launch, so we got some space with that. If I remove "navigation" tabs in future and drive it only with next / back buttons, another space will be there too.

Make that area scrollable when needed sounds good, will note it for future improvements.

Laserlicht:
Tested it on old android 5 (4,7 inch) phone. Don't think layout is a big problem. Android scales much better than iOS.

But importing from folder doesn't work anymore (working in develop). And gog neither (already crashes in develop).

Not sure if we need to support such old device. But VCMI seems to run flawless once installed.

GeorgeK1ng:
Interesting, does folder picker appear? On Google TV with Android 11 it does nothing as used component is not probably present in system. Also don't see these issues on Android 9.

Laserlicht:
> Interesting, does folder picker appear? On Google TV with Android 11 it does nothing as used component is not probably present in system. Also don't see these issues on Android 9.

Yes. It appears. Also showing import load bar. Then crash.

kambala-decapitator:
I'd bet on an android api method that is unavailable on old androids

launcher/firstLaunch/firstlaunch_moc.cpp
GeorgeK1ng: Currently it's for not hiding tabs, when they are removed in future PR this will be adjusted.

launcher/helper.cpp
GeorgeK1ng: Already enough to have them both on same position.

image

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
GeorgeK1ng: No, same as before. Only supported to directly scan for Maps / Mp3 folder one level up. So if user select folder with LODs directly and one level up exist Maps / Mp3 folders, it will be collected for import correctly.

GeorgeK1ng:
I have adjusted code now. Tested in my Android tablet and works fine, have not found any visible issue.

But I will add to this PR one more change "fail-fast" for EXE scanning soon.

@Laserlicht Java code is now different, can you test artifact on your Android 5 with "copy files" option please?

GeorgeK1ng:
Added Fail-Fast for EXE / BIN. Now user select EXE first. And checks are performed immediatelly, so when user selecting wrong file, error is visible "real-time". Also when selected valid EXE, correct BIN file name is shown to select.
For mobiles platforms added something like filters logic to be sure to select EXE in select EXE step.
And added default open dir logic, on desktop will try to open Downloads / Home / Desktop

Tested on Android 15 and Windows.

launcher/firstLaunch/firstlaunch_moc.cpp
GeorgeK1ng: Yes, _true_ one is for GOG behavior, _false_ is for copy user provided data. They are for 2 different cases.

GeorgeK1ng:
Fixed spaces vs tabs in firstlaunch_moc.cpp. I have also added 2 untested fixes. Don't see any side effects right now. And if I not forget any reviews, all of them should be addressed

1) Check for possibility to use DocumentUI when Android API >= 21. Hope it will correctly hide "copy existing data" button on Google TV or any other "lite Android", will see next week, or if there is anyone with Google TV and can test it, it will be great.
2) Android 5 require test - hope it won't crash anymore

launcher/helper.cpp
```diff
@@ -145,4 +151,195 @@ void keepScreenOn(bool isEnabled)
 	iOS_utils::keepScreenOn(isEnabled);
 #endif
 }
+
+bool canUseFolderPicker()
+{
+#if defined(VCMI_ANDROID)
+	// selecting directory with ACTION_OPEN_DOCUMENT_TREE is available only since API level 21
+	if(QtAndroid::androidSdkVersion() < 21)
+		return false;
```
IvanSavenko: No longer relevant since API level 21 will soon become requirement

```suggestion
```

launcher/firstLaunch/progressoverlay.cpp
```diff
@@ -0,0 +1,79 @@
+/*
+ * firstlaunch_moc.cpp, part of VCMI engine
```
IvanSavenko:
```suggestion
 * progressoverlay.cpp, part of VCMI engine
```

launcher/firstLaunch/progressoverlay.h
```diff
@@ -0,0 +1,34 @@
+/*
+ * firstlaunch_moc.h, part of VCMI engine
```
IvanSavenko:
```suggestion
 * progressoverlay.h, part of VCMI engine
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -627,27 +709,70 @@ bool FirstLaunchView::checkCanInstallTranslation()
...
+		}
+	
+	QStringList files = dataDir.entryList(QDir::Files | QDir::Readable);
+	for(const QString &name : files)
+	{
+		if(name.compare(QStringLiteral("H3ab_spr.lod"), Qt::CaseInsensitive) == 0)
+		{
+			QFileInfo lodInfo(dataDir.filePath(name));
+			quint64 fileSize = static_cast<quint64>(lodInfo.size());
+			logGlobal->error("H3ab_spr.lod size: %llu", fileSize);
```
IvanSavenko:
```suggestion
			logGlobal->trace("H3ab_spr.lod size: %llu", fileSize);
```

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -316,270 +305,349 @@ QString FirstLaunchView::getHeroesInstallDir()
...
+
+	QFileInfo exeInfo(fileExe);
+	QString expectedBinName = exeInfo.completeBaseName() + "-1.bin";
 	QString filterBin = tr("GOG data") + " (*.bin)";
-	QString filterExe = tr("GOG installer") + " (*.exe)";
+	QString titleBin = tr("Select the offline GOG installer data file: %1", "param is file name").arg(expectedBinName);
+
+	QString fileBin = fileSelection(titleBin, filterBin, exeInfo.absolutePath());
+	if(fileBin.isEmpty())
+		return;
```
IvanSavenko: Would be great to skip this step if file with such name already exists and can be accessed.

Obviously required on mobiles and flatpak due to access permissions, but rest of PC systems can skip this manual step and auto-pick correct file.

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -316,270 +305,349 @@ QString FirstLaunchView::getHeroesInstallDir()
...
-	QStringList dirData = tempDir.entryList({"data"}, QDir::Filter::Dirs);
-	if(!errorText.isEmpty() || dirData.empty() || QDir(tempDir.filePath(dirData.front())).entryList({"*.lod"}, QDir::Filter::Files).empty())
+	for(const QString &line : items)
 	{
-		if(!errorText.isEmpty())
+		const auto part = line.split('\t');
+
+		const QString &src  = part[0];
+		const QString &tgt  = part[1]; // Data / Maps / Mp3
+		const QString &file = part[2];
```
IvanSavenko:
```suggestion
		const QString &source  = part[0];
		const QString &target  = part[1]; // Data / Maps / Mp3
		const QString &file = part[2];
```

GeorgeK1ng:
Addressed review comments and added option to skip BIN when it exist in selected EXE directory and can be accessed. Also fixed canUseFolderPicker to disable that option on TVs. All other solutions I tried have not worked correctly on various testing devices. For example button was hidden even I know it will work. So switched ti for now to TV detection. Hope nobody install VCMI in Watch :D

IvanSavenko:
@kambala-decapitator any further comments or good to go?

GeorgeK1ng:
I have some touches to add there after testing on Google TV.

kambala-decapitator:
will review on Saturday

launcher/firstLaunch/firstlaunch_moc.cpp
```diff
@@ -380,32 +305,136 @@ QString FirstLaunchView::getHeroesInstallDir()
 #endif
 	return QString{};
 }
+	
+static QString defaultStartDirForOpen()
+{
+#if defined(VCMI_MOBILE)
+	const QStandardPaths::StandardLocation mobilePrefs[] = {
+		QStandardPaths::DocumentsLocation,
+		QStandardPaths::HomeLocation
```
kambala-decapitator, line 314: why Downloads are ignored on mobile?

kambala-decapitator (commented):
java files are a mix of tabs and spaces now

GeorgeK1ng:
Ready I quess from my side for now.

launcher/firstLaunch/firstlaunch_moc.cpp
GeorgeK1ng: No idea, they are ignored from some reason.

GeorgeK1ng:
Seems target needs to be updated to beta branch.

GeorgeK1ng:
@kambala-decapitator

android/vcmi-app/src/main/java/eu/vcmi/vcmi/NativeMethods.java
```diff
@@ -26,6 +26,7 @@ public NativeMethods()
     }
 
     public static native void initClassloader();
+    public static native boolean tryToSaveTheGame();
```
IvanSavenko: What's this change about?

launcher/firstLaunch/progressoverlay.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+    progressBar = new QProgressBar(this);
+    progressBar->setMinimumHeight(18);
+    setIndeterminate(true);
+
+    layout->addStretch();
+    layout->addWidget(titleLabel);
+    layout->addWidget(fileLabel);
+    layout->addWidget(progressBar);
+    layout->addStretch();
```
IvanSavenko, line 45: Still not sure why this class has to be constructed via c++ code and not via .ui file / Qt Designer, like rest of the Launcher

launcher/firstLaunch/firstlaunch_moc.ui
```diff
@@ -776,7 +803,55 @@ Offline installer consists of two files: &quot;.exe&quot; and &quot;.bin&quot; -
...
+         <item row="6" column="1">
+          <widget class="QLabel" name="labelPresetTowDescr">
+           <property name="sizePolicy">
+            <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+             <horstretch>100</horstretch>
+             <verstretch>0</verstretch>
+            </sizepolicy>
+           </property>
+           <property name="text">
+            <string>Install the Tides of War expansion, which adds the ability to build alternative level 7 creature dwellings for all 9 standard towns</string>
```
IvanSavenko: I know that this is actual description of ToW that we have at the moment, but would be great to update it to better represent ToW

launcher/helper.cpp
```diff
@@ -145,4 +151,198 @@ void keepScreenOn(bool isEnabled)
...
+	// If user picked "Data", also scan ../Maps and ../Mp3 (if present)
+	if(root.dirName().compare(QLatin1String("Data"), Qt::CaseInsensitive) == 0)
+	{
+		QDir parent = root;
+		if(parent.cdUp())
+		{
+			addIfExists(scan, parent, "Maps");
+			addIfExists(scan, parent, "Mp3");
+		}
+	}
```
IvanSavenko, line 328: Old code, but not sure if it works on mobile systems with restricted filesystem access or on flatpak.

Hopefully, `cdUp` will correctly return false on such systems, but untested.

launcher/helper.cpp
```diff
@@ -145,4 +151,198 @@ void keepScreenOn(bool isEnabled)
...
+
+		while(it.hasNext())
+		{
+			const QString filePath = it.next();
+			const QFileInfo file(filePath);
+			const QString target = classifyTargetByExt(file.fileName());
+			if(target.isEmpty())
+				continue;
+
+			out.push_back(filePath + QLatin1Char('\t') + target + QLatin1Char('\t') + file.fileName());	// "src \t Target \t Name"
```
IvanSavenko, line 343: Looks like what you're generating is not a string list, but list/vector of some struct with 3 string fields. So why not actually use struct, e.g.
```cpp
struct FileToCopy
{
    QString filePath;
    QString target;
    QString name;
}
QVector<FileToCopy> findFilesForCopy(const QString &path)
```

launcher/helper.cpp
```diff
@@ -145,4 +151,198 @@ void keepScreenOn(bool isEnabled)
...
+#elif defined(VCMI_IOS)
+	// selecting directory through UIDocumentPickerViewController is available only since iOS 13
+	return iOS_utils::isOsVersionAtLeast(13);
+#else
+	return true;
+#endif
+}
+
+#ifdef VCMI_ANDROID
+static constexpr int  kFolderPickerReqCode = 4242;
```
IvanSavenko, line 184: What's this magic number? If this is from Android docs comment with link would be helpful

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
```diff
@@ -27,129 +34,326 @@
...
+			target = TARGET_MP3;
+			foundMp3++;
+		}
+
+		if (target != null)
+			out.add(String.format(Locale.ROOT, "%s\t%s\t%s", entry.getUri().toString(), target, name));
+
+	}
+
+	// 2) If user picked "Data" and Maps/Mp3 were not found, also scan ../Maps and ../Mp3
```
IvanSavenko, line 244: I think this branch would never hit on Android due to restricted filesystem access?

launcher/firstLaunch/firstlaunch_moc.ui
IvanSavenko: Suggested version from krs on Discord:
> A feature-rich expansion for VCMI that introduces one alternative unit upgrade for every faction, adds powerful new neutral creatures, unique creature banks, new skills and spells to diversify the classic Heroes III gameplay.

launcher/firstLaunch/progressoverlay.cpp
GeorgeK1ng: I was not able to create it without breaking current elements layouts. If it can be somehow done in UI file it would be great of course. As we talked about this at PR openening, can be improved in next first launch things which are not adressed in this one.

launcher/helper.cpp
GeorgeK1ng: Initialy I have removed it, but later I restored it as it really works. 4 Android devices with different versions and can be choosen directly "Data" folder and both maps / mp3 folders are then found and used for file copy list.

android/vcmi-app/src/main/java/eu/vcmi/vcmi/util/FileUtil.java
GeorgeK1ng: Works too, tested it multiple times with success on Android devices.

launcher/helper.cpp
GeorgeK1ng: Can be probably done this way too. It's too long from days when I choose "tabbed" list. So I don't really remember why I choosed current working approach.

launcher/helper.cpp
IvanSavenko: Strange... But well, if it works then OK.

launcher/helper.cpp
GeorgeK1ng: Added comment about this magic number.

launcher/firstLaunch/firstlaunch_moc.ui
GeorgeK1ng: Thats too long text for current fields. But can be easily adapted to:

"A feature-rich expansion that expands gameplay with one alternative unit for each of the 9 standard towns, plus new neutral creatures, creature banks, skills, and spells"

android/vcmi-app/src/main/java/eu/vcmi/vcmi/NativeMethods.java
GeorgeK1ng: Removed, I was keep it there by accident when prepared files for upload.

GeorgeK1ng:
Was able to test iOS artifact now on iPad mini 2 with iOS 12.5.7. 

Just small visible glitch on top of overlay - currently is there 50px to keep visible navigation, which will be removed in future and only next / back buttons will be there.

IMG_0001
IMG_0002
IMG_0003
IMG_0004
IMG_0006
IMG_0008
IMG_0009

Will consider in few days if I can split language selection to first standalone page and remove completely that header.

Laserlicht:
Maybe also change the text about HD edition? We have some kind of support now. That's maybe a little bit irritating.

GeorgeK1ng:
I tried to avoid touching existing text to not invalidate these strings for translations. And HD Edition Is not supported as base files. So maybe completely remove that part of text?

Laserlicht:
> I tried to avoid touching existing text to not invalidate these strings for translations. And HD Edition Is not supported as base files. So maybe completely remove that part of text?

Would not remove it, just adjust. E.g. `...not supported as base`?. But also ok to do it at a later point outside of PR to avoid invalidation of text.

Issue #5303, opened 2025-01-20
Battle only mode


Laserlicht:
**Describe your proposal**
A great feature that exists in fheroes2 is missing in VCMI.

The battle only mode. This allows you to create battles without having an active game map. You can create a battle in the main menu (units, artifacts, stats of your own and the opponent's side) and start it directly.

Todo:
- [x] Basic implementation: #6243
- [x] Skill support #6301
- [x] Artifact support #6301
- [x] War machines #6301
- [x] Configurable spells #6301
- [x] Multiplayer support
- [x] Add shortcut possibility for mainmenu to directly start into battle mode #6301
- [ ] Presets
- [ ] Fights against neutrals
- [x] Seperate menu (in line with OH3) #6301
- [ ] Points mode (to "buy" by player)
- [ ] Town configurable (level, buildings)
- [ ] Fight against buildings (e.g. dragon utopia)

IvanSavenko:
For presets it would be great to have both premade presets (for example 1 preset for each hero class, so 2 per faction) - for quick battles as well as option to save/load own presets. Another option is to have an option (e.g. checkbox) to force preset usage for MP

As for UI - another option is completely separate menu (left panel, like random maps), to have UI more in line with H3

Few more potential improvements:
- points mode. Check H5 and I think some of Total War games. Both players have set number of points that they can use to buy army with or to give hero artifacts / skills / etc, forcing players to have roughly equal armies.
- possibly support for configuring town level (fort/citadel/castle) for siege battles. Perhaps some specific buildings as well, e.g. Grail. Perhaps even presence of hero during siege.

Laserlicht:
Added it to todo. Some points I already had in mind, but not written down. Thanks.

If we can merge soon I may can implement something from the list before beta. At least the panel stuff.

krs0:
+1 for creature bank support.
+1 for fights against neutrals.
+1 for preset loading saving.
Mana should be modifiable.
Obstacle presets should be possible.

krs0:
Points mode is a nice minigame in itself. To be able to min/max skills / spells / artifacts / creatures.

But for that, wouldn't a map template be easier to implement and have a lot more features?

Laserlicht:
Point mode is in general not hard to implement.

Just add
- functionality to modify max points per player
- functionality to check current points to max points and avoid game start if above
- calculate points from start `BattleOnlyModeStartInfo` -> This is the tricky point. Needs to be well balanced.

[...]

PR #6243, opened 2025-10-23
Battleonly mode

Laserlicht:
Currently supported:
- Battle against heroes
- Battle against hero in town
- Battle against town without hero
- Selecting hero
- Selecting creature
- Selecting primary skill level
- Selecting battleground

Selectable in single player at map selection (small icon on top left).

Progress is tracked here: #5303

### DIFF
--- a/client/lobby/BattleOnlyMode.cpp
+++ b/client/lobby/BattleOnlyMode.cpp
@@ -95,6 +95,9 @@ BattleOnlyModeWindow::BattleOnlyModeWindow()
 		auto & terrains = LIBRARY->terrainTypeHandler->objects;
 		for (const auto & terrain : terrains)
 		{
+			if(!terrain->isPassable())
+				continue;
+
 			texts.push_back(terrain->getNameTranslated());
 
 			const auto & patterns = LIBRARY->terviewh->getTerrainViewPatterns(terrain->getId());
@@ -142,12 +145,12 @@ BattleOnlyModeWindow::BattleOnlyModeWindow()
 			startInfo->selectedHero[0] = std::nullopt;
 			startInfo->selectedArmy[0].fill(CStackBasicDescriptor(CreatureID::NONE, 1));
 			for(size_t i=0; i<GameConstants::ARMY_SIZE; i++)
-				heroSelector1->selectedArmyInput.at(i)->setText("1");
+				heroSelector1->selectedArmyInput.at(i)->disable();
 		}
 		startInfo->selectedHero[1] = std::nullopt;
 		startInfo->selectedArmy[1].fill(CStackBasicDescriptor(CreatureID::NONE, 1));
 		for(size_t i=0; i<GameConstants::ARMY_SIZE; i++)
-			heroSelector2->selectedArmyInput.at(i)->setText("1");
+			heroSelector2->selectedArmyInput.at(i)->disable();
 		onChange();
 	});
 	buttonReset->setTextOverlay(LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeReset"), EFonts::FONT_SMALL, Colors::WHITE);
@@ -260,9 +263,10 @@ BattleOnlyModeHeroSelector::BattleOnlyModeHeroSelector(int id, BattleOnlyModeWin
 			{
 				parent.startInfo->selectedArmy[id][i].setCount(TextOperations::parseMetric<int>(text));
 				parent.onChange();
+				selectedArmyInput[i]->enable();
 			}
 			else
-				selectedArmyInput[i]->setText("1");
+				selectedArmyInput[i]->disable();
 		});
 	}
 
@@ -295,11 +299,11 @@ void BattleOnlyModeHeroSelector::setHeroIcon()
 		std::sort(heroes.begin(), heroes.end(), [](auto a, auto b) {
 			auto heroA = a.toHeroType();
 			auto heroB = b.toHeroType();
-			if (heroA->heroClass->getId() == heroB->heroClass->getId())
-				return heroA->getNameTranslated() < heroB->getNameTranslated();
-			if (heroA->heroClass->faction == heroB->heroClass->faction)
+			if(heroA->heroClass->faction != heroB->heroClass->faction)
+				return heroA->heroClass->faction < heroB->heroClass->faction;
+			if(heroA->heroClass->getId() != heroB->heroClass->getId())
 				return heroA->heroClass->getId() < heroB->heroClass->getId();
-			return heroA->heroClass->faction < heroB->heroClass->faction;
+			return heroA->getNameTranslated() < heroB->getNameTranslated();
 		});
 
 		int selectedIndex = !parent.startInfo->selectedHero[id] ? 0 : (1 + std::distance(heroes.begin(), std::find_if(heroes.begin(), heroes.end(), [this](auto heroID) {
@@ -319,7 +323,7 @@ void BattleOnlyModeHeroSelector::setHeroIcon()
 			image->scaleTo(Point(35, 23), EScalingAlgorithm::NEAREST);
 			images.push_back(image);
 		}
-		auto window = std::make_shared<CObjectListWindow>(texts, nullptr, LIBRARY->generaltexth->translate("object.core.hero.name"), LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeHeroSelect"), [this, heroes](int index){
+		auto window = std::make_shared<CObjectListWindow>(texts, nullptr, LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeHeroSelect"), LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeHeroSelect"), [this, heroes](int index){
 			if(index == 0)
 			{
 				parent.startInfo->selectedHero[id] = std::nullopt;
@@ -361,7 +365,7 @@ void BattleOnlyModeHeroSelector::setCreatureIcons()
 		if(parent.startInfo->selectedArmy[id][i].getId() == CreatureID::NONE)
 		{
 			creatureImage[i] = std::make_shared<CPicture>(drawBlackBox(Point(32, 32), LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeSelect"), id == 1 ? Colors::WHITE : parent.disabledColor), Point(6 + i * 36, 78));
-			selectedArmyInput[i]->setText("1");
+			selectedArmyInput[i]->disable();
 		}
 		else
 		{
@@ -369,6 +373,7 @@ void BattleOnlyModeHeroSelector::setCreatureIcons()
 			auto creatureID = unit.getId();
 			creatureImage[i] = std::make_shared<CPicture>(ENGINE->renderHandler().loadAnimation(AnimationPath::builtin("CPRSMALL"), EImageBlitMode::COLORKEY)->getImage(LIBRARY->creh->objects.at(creatureID)->getIconIndex()), Point(6 + i * 36, 78));
 			selectedArmyInput[i]->setText(TextOperations::formatMetric(unit.getCount(), 3));
+			selectedArmyInput[i]->enable();
 		}
 
 		creatureImage[i]->addLClickCallback([this, i](){
@@ -377,11 +382,13 @@ void BattleOnlyModeHeroSelector::setCreatureIcons()
 			std::sort(creatures.begin(), creatures.end(), [](auto a, auto b) {
 				auto creatureA = a.toCreature();
 				auto creatureB = b.toCreature();
-				if (creatureA->getLevel() == creatureB->getLevel())
-					return creatureA->getNameSingularTranslated() < creatureB->getNameSingularTranslated();
-				if (creatureA->getFactionID() == creatureB->getFactionID())
+				if(creatureA->getFactionID() != creatureB->getFactionID())
+					return creatureA->getFactionID() < creatureB->getFactionID();
+				if(creatureA->getLevel() != creatureB->getLevel())
 					return creatureA->getLevel() < creatureB->getLevel();
-				return creatureA->getFactionID() < creatureB->getFactionID();
+				if(creatureA->upgrades.size() != creatureB->upgrades.size())
+					return creatureA->upgrades.size() > creatureB->upgrades.size();
+				return creatureA->getNameSingularTranslated() < creatureB->getNameSingularTranslated();
 			});
 
 			int selectedIndex = parent.startInfo->selectedArmy[id][i].getId() == CreatureID::NONE ? 0 : (1 + std::distance(creatures.begin(), std::find_if(creatures.begin(), creatures.end(), [this, i](auto creatureID) {
@@ -401,11 +408,10 @@ void BattleOnlyModeHeroSelector::setCreatureIcons()
 				image->scaleTo(Point(23, 23), EScalingAlgorithm::NEAREST);
 				images.push_back(image);
 			}
-			auto window = std::make_shared<CObjectListWindow>(texts, nullptr, LIBRARY->generaltexth->translate("core.genrltxt.42"), LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeCreatureSelect"), [this, creatures, i](int index){
+			auto window = std::make_shared<CObjectListWindow>(texts, nullptr, LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeCreatureSelect"), LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeCreatureSelect"), [this, creatures, i](int index){
 				if(index == 0)
 				{
-					if(parent.startInfo->selectedArmy[id][i].getId() == CreatureID::NONE)
-						parent.startInfo->selectedArmy[id][i] = CStackBasicDescriptor(CreatureID::NONE, 1);
+					parent.startInfo->selectedArmy[id][i] = CStackBasicDescriptor(CreatureID::NONE, 1);
 					parent.onChange();
 					return;
 				}
--- a/lib/mapObjects/CGHeroInstance.cpp
+++ b/lib/mapObjects/CGHeroInstance.cpp
@@ -1345,28 +1345,6 @@ void CGHeroInstance::restoreBonusSystem(CGameState & gs)
 	}
 }
 
-void CGHeroInstance::attachToBonusSystem(CGameState & gs)
-{
-	CArmedInstance::attachToBonusSystem(gs);
-	if (boardedBoat.hasValue())
-	{
-		auto boat = gs.getObjInstance(boardedBoat);
-		if (boat)
-			attachTo(dynamic_cast<CGBoat&>(*boat));
-	}
-}
-
-void CGHeroInstance::detachFromBonusSystem(CGameState & gs)
-{
-	CArmedInstance::detachFromBonusSystem(gs);
-	if (boardedBoat.hasValue())
-	{
-		auto boat = gs.getObjInstance(boardedBoat);
-		if (boat)
-			detachFrom(dynamic_cast<CGBoat&>(*boat));
-	}
-}
-
 CBonusSystemNode & CGHeroInstance::whereShouldBeAttached(CGameState & gs)
 {
 	if(visitedTown.hasValue())
--- a/lib/mapObjects/CGHeroInstance.h
+++ b/lib/mapObjects/CGHeroInstance.h
@@ -311,8 +311,6 @@ public:
 
 	void afterAddToMap(CMap * map) override;
 	void afterRemoveFromMap(CMap * map) override;
-	void attachToBonusSystem(CGameState & gs) override;
-	void detachFromBonusSystem(CGameState & gs) override;
 	void restoreBonusSystem(CGameState & gs) override;
 
 	void updateFrom(const JsonNode & data) override;


Mods/vcmi/Content/config/english.json
```diff
@@ -138,6 +138,12 @@
 	"vcmi.lobby.deleteFile" : "Do you want to delete following file?",
 	"vcmi.lobby.deleteFolder" : "Do you want to delete following folder?",
 	"vcmi.lobby.deleteMode" : "Switch to delete mode and back",
+	"vcmi.lobby.battleOnlyMode" : "Battle only mode",
+	"vcmi.lobby.battleOnlyModeBattleground" : "Battleground",
+	"vcmi.lobby.battleOnlyModeBattlegroundSelect" : "Select Battleground",
```
IvanSavenko: Should be called "battlefield", not "battleground"

Can be replaced everywhere - e.g. via search-and-replace

client/CPlayerInterface.cpp
```diff
@@ -658,7 +659,7 @@ void CPlayerInterface::battleStart(const BattleID & battleID, const CCreatureSet
 {
 	EVENT_HANDLER_CALLED_BY_CLIENT;
 
-	bool useQuickCombat = settings["adventure"]["quickCombat"].Bool();
+	bool useQuickCombat = settings["adventure"]["quickCombat"].Bool() || GAME->map().getMap()->battleOnly;
```
IvanSavenko, line 662: Shouldn't it be the opposite? 

If game is in battle-only mode, never use quick combat and always manual combat?

lib/modding/IdentifierStorage.cpp
```diff
@@ -126,7 +126,7 @@ CIdentifierStorage::ObjectCallback CIdentifierStorage::ObjectCallback::fromNameW
 
 CIdentifierStorage::ObjectCallback CIdentifierStorage::ObjectCallback::fromNameAndType(const std::string & scope, const std::string & type, const std::string & fullName, const std::function<void(si32)> & callback, bool optional, bool bypassDependenciesCheck, bool caseSensitive)
 {
-	assert(!scope.empty());
+	//assert(!scope.empty());
```
IvanSavenko: Leftover from debug? This generally means corrupted request

Same for another edit in this file

lib/bonuses/CBonusSystemNode.cpp
```diff
@@ -184,7 +184,6 @@ CBonusSystemNode::~CBonusSystemNode()
 
 void CBonusSystemNode::attachTo(CBonusSystemNode & parent)
 {
-	assert(!vstd::contains(parentsToPropagate, &parent));
```
IvanSavenko, line 187: Why was this removed?

Same for other changes in this file

client/lobby/SelectionTab.cpp
```diff
@@ -241,6 +242,15 @@ SelectionTab::SelectionTab(ESelectionScreen Type)
 		sortByDate->setOverlay(std::make_shared<CPicture>(ImagePath::builtin("lobby/selectionTabSortDate")));
 		buttonsSortBy.push_back(sortByDate);
 
+		bool isMultiplayer = GAME->server().loadMode == ELoadMode::MULTI;
```
IvanSavenko: Perhaps we should check for *single* player instead? I think this check might trigger for game load / campaigns. Or not?

client/lobby/SelectionTab.cpp
```diff
@@ -325,6 +335,7 @@ void SelectionTab::toggleMode()
 				inputName->disable();
 				auto files = getFiles("Maps/", EResType::MAP);
 				files.erase(ResourcePath("Maps/Tutorial.tut", EResType::MAP));
+				files.erase(ResourcePath("Maps/BattleOnlyMode.vmap", EResType::MAP));
```
IvanSavenko, line 338: What is this map?

client/lobby/BattleOnlyMode.cpp
```diff
@@ -0,0 +1,461 @@
...
+			TerrainViewPattern pattern;
+			for(auto & p : patterns)
+				if(p[0].id == "n1")
+					pattern = p[0];
+			auto image = ENGINE->renderHandler().loadImage(terrain->tilesFilename, pattern.mapping[0].first, 0, EImageBlitMode::OPAQUE);
+			image->scaleTo(Point(23, 23), EScalingAlgorithm::NEAREST);
+			images.push_back(image);
+		}
+
+		auto factions = LIBRARY->townh->objects;
```
IvanSavenko: Copy of entire list of factions?

Better to use getDefaultAllowed or iterate over list directly without copy creation

client/lobby/BattleOnlyMode.cpp
```diff
@@ -0,0 +1,461 @@
...
+	buttonOk = std::make_shared<CButton>(Point(281, 288), AnimationPath::builtin("MuBchck"), CButton::tooltip(), [this](){ startBattle(); }, EShortcut::GLOBAL_ACCEPT);
+	buttonAbort = std::make_shared<CButton>(Point(355, 288), AnimationPath::builtin("MuBcanc"), CButton::tooltip(), [this](){ close(); }, EShortcut::GLOBAL_CANCEL);
+	title = std::make_shared<CLabel>(350, 20, FONT_BIG, ETextAlignment::CENTER, Colors::YELLOW, LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyMode"));
+
+	battlegroundSelector = std::make_shared<CButton>(Point(10, 294), AnimationPath::builtin("GSPButtonClear"), CButton::tooltip(), [this](){
+		std::vector<std::string> texts;
+		std::vector<std::shared_ptr<IImage>> images;
+
+		auto terrains = LIBRARY->terrainTypeHandler->objects;
+		for (const auto & terrain : terrains)
```
IvanSavenko:
```suggestion
		for (const auto & terrain : LIBRARY->terrainTypeHandler->objects)
```

client/lobby/BattleOnlyMode.cpp
```diff
@@ -0,0 +1,461 @@
...
+				{
+					if(!selectedArmy->slotEmpty(SlotID(i)))
+						selectedArmy->eraseStack(SlotID(i));
+					setCreatureIcons();
+					return;
+				}
+				index--;
+				auto creature = creatures[index];
+				int amount = selectedArmy->slotEmpty(SlotID(i)) ? 1 : selectedArmy->Slots().at(SlotID(i))->getCount();
+				ENGINE->windows().createAndPushWindow<NumberInputWindow>(amount, [this, creature, i](int quantity){
```
IvanSavenko: Perhaps instead allow to input number directly below creature icon instead of creating new window?

client/lobby/BattleOnlyMode.cpp
```diff
@@ -0,0 +1,461 @@
...
+		}
+		else
+		{
+			auto creatureID = selectedArmy->Slots().at(SlotID(i))->getCreatureID();
+			creatureImage[i] = std::make_shared<CPicture>(ENGINE->renderHandler().loadAnimation(AnimationPath::builtin("CPRSMALL"), EImageBlitMode::COLORKEY)->getImage(LIBRARY->creh->objects.at(creatureID)->getIconIndex()), Point(6 + i * 36, 78));
+			creatureImageLabel[i] = std::make_shared<CLabel>(36 + i * 36, 112, FONT_SMALL, ETextAlignment::BOTTOMRIGHT, Colors::WHITE, TextOperations::formatMetric(selectedArmy->Slots().at(SlotID(i))->getCount(), 4));
+		}
+
+		creatureImage[i]->addLClickCallback([this, i](){
+			auto creatures = LIBRARY->creh->objects;
```
IvanSavenko: Same here. Better to use LIBRARY->creh->getDefaultAllowed call and iterate over creature ID's

client/lobby/BattleOnlyMode.cpp
```diff
@@ -0,0 +1,461 @@
...
+		heroLabel = std::make_shared<CLabel>(160, 16, FONT_SMALL, ETextAlignment::CENTER, Colors::WHITE, LIBRARY->generaltexth->translate("core.genrltxt.507"));
+	}
+	else
+	{
+		heroImage = std::make_shared<CPicture>(ENGINE->renderHandler().loadAnimation(AnimationPath::builtin("PortraitsLarge"), EImageBlitMode::COLORKEY)->getImage(selectedHero->getHeroType()->imageIndex), Point(6, 7));
+		heroLabel = std::make_shared<CLabel>(160, 16, FONT_SMALL, ETextAlignment::CENTER, Colors::WHITE, selectedHero->getNameTranslated());
+	}
+
+	heroImage->addLClickCallback([this](){
+		auto heroes = LIBRARY->heroh->objects;
```
IvanSavenko: And here

IvanSavenko:
Also - can you create issue with your plans for this feature (e.g. MP support) to have proper place to discuss final form / suggestions for this feature that are out of scope for this PR?

GeorgeK1ng:
For what is used right side? I have tried to set battle and it worked - must click on manual battle as it uses quick combat right now as default.

Also, will be possible to use something like "start battleonly" in [mainmenu.json](https://github.com/vcmi/vcmi/blob/6b4179af86fae2bb5a78db20a539aba05d42cb95/config/mainmenu.json#L78)? Seems like we can add it to custom main menu pack when we have proper button image - probably two titans swords crossed with "Quick battle" text?

image

client/CPlayerInterface.cpp
Laserlicht: I prefer it this way because you can check if you are better (with fewer losses) than AI.

lib/modding/IdentifierStorage.cpp
Laserlicht: Yes. You are right. Leftovers. But maybe we should keep it commented out because this assert is triggered in normal gameplay since a few month if you use `vmap`. So every time if you use debug and vmap you have to comment this stuff out.

lib/bonuses/CBonusSystemNode.cpp
Laserlicht: It seems the bonus system does not work with `EditorCallback` (maybe not implemented gamestate?).
It's needed for fighting on boat on water. without removing the asserts are triggered. No idea why.

But after removing - it's fully working.

client/lobby/SelectionTab.cpp
Laserlicht: Not triggering for loading / campaigns.

Testing for `ELoadMode::SINGLE` does not work correctly atm (tested it at first). Would need some refactoring.

client/lobby/SelectionTab.cpp
Laserlicht: Battlemode currently generates map on the fly, saves it and start it with the normal, already existing game start process. This is basically needing a filename, we cannot send a CMap for starting game yet.

And this generated map should not show up in list.

client/lobby/BattleOnlyMode.cpp
Laserlicht: Yes. Also considered this. But identified it as improvement area for later. Because needs to adjust assets for fitting boxes.

Laserlicht:
> For what is used right side? I have tried to set battle and it worked - must click on manual battle as it uses quick combat right now as default.

Right area is for setting artifacts, secondary skills, spell selection, additional buttons (like "reset") and maybe other ideas.
Not sure when it comes in action. And if it is maybe better to drop right area and instead add custom buttons (more flexible).

> Also, will be possible to use something like "start battleonly" in [mainmenu.json](https://github.com/vcmi/vcmi/blob/6b4179af86fae2bb5a78db20a539aba05d42cb95/config/mainmenu.json#L78)? Seems like we can add it to custom main menu pack when we have proper button image - probably two titans swords crossed with "Quick battle" text?

Also considered it. But I think also for later (not really on the top of priority).

Laserlicht:
> Also - can you create issue with your plans for this feature (e.g. MP support) to have proper place to discuss final form / suggestions for this feature that are out of scope for this PR?

Yes we can use #5303. Can add a todo list later.

Laserlicht:
Added todo here: #5303

Changed layout and added reset button.
grafik

Also improved numeric input.

Open review points will be fixed later.

Laserlicht:
It's now possible to play in multiplayer.

PR is already huge (hard to review) so I'll keep it as it is. And may implement more stuff & fixes in next PR.

client/CPlayerInterface.cpp
IvanSavenko: Might look confusing to players, especially those who are not aware of "manual replay" button - you start duel mode... and battle immediately ends

lib/texts/TextOperations.h
```diff
@@ -116,4 +121,67 @@ inline std::string TextOperations::formatMetric(Arithmetic number, int maxDigits
...
+	{
+		return !(std::isdigit(static_cast<unsigned char>(c)) || c == '-');
+	}), numberPart.end());
+
+	if (numberPart.empty() || (numberPart == "-"))
+		return 0;
+
+	try
+	{
+		Arithmetic value = boost::lexical_cast<Arithmetic>(numberPart);
```
IvanSavenko: I guess lexical_cast is soft-deprecated now, since c++ now provides similar tools

```suggestion
		Arithmetic value = std::stoll(numberPart);
```

lib/texts/TextOperations.h
```diff
@@ -116,4 +121,67 @@ inline std::string TextOperations::formatMetric(Arithmetic number, int maxDigits
 	return std::to_string(number) + *iter;
 }
 
+template<typename Arithmetic>
+inline Arithmetic TextOperations::parseMetric(const std::string &text)
+{
+	static const std::string symbols = " kMGTPE";
```
IvanSavenko: Unused?
```suggestion
```

lib/StartInfo.h
```diff
@@ -239,5 +239,27 @@ struct DLL_LINKAGE LobbyInfo : public LobbyState
...
 };
 
+class DLL_LINKAGE BattleOnlyModeStartInfo : public Serializeable
+{
+public:
+	std::optional<TerrainId> selectedTerrain;
+	std::optional<FactionID> selectedTown;
+
+	std::array<std::optional<HeroTypeID>, 2> selectedHero;
+	std::array<std::array<std::optional<std::pair<CreatureID, int>>, 7>, 2> selectedArmy;
```
IvanSavenko: Perhaps use existing class instead of pair?
Also - no need to use optional, CreatureID::NONE already covers such case

```suggestion
	std::array<std::array<CStackBasicDescriptor, 7>, 2> selectedArmy;
```

client/CPlayerInterface.cpp
Laserlicht: Yes. For first time. But I think it's easy to understand the second time.

Fheroes2 uses the same approach.

Laserlicht: May add a checkbox to disable/enable this in a later PR.

lib/bonuses/CBonusSystemNode.cpp
IvanSavenko: When exactly these asserts are triggered? Can you show call stack? Looks like double-attach.

Previously, we had weird bugs where bonus system ended up in invalid state, like entities that were attached to wrong parents. Removing these asserts means that such bugs can resurface again

Laserlicht:
> When exactly these asserts are triggered? Can you show call stack? Looks like double-attach.

```
libc.so.6!__pthread_kill_implementation(int no_tid, int signo, pthread_t threadid) (pthread_kill.c:44)
libc.so.6!__pthread_kill_internal(int signo, pthread_t threadid) (pthread_kill.c:78)
libc.so.6!__GI___pthread_kill(pthread_t threadid, int signo, int signo@entry) (pthread_kill.c:89)
libc.so.6!__GI_raise(int sig, int sig@entry) (raise.c:26)
libc.so.6!__GI_abort() (abort.c:79)
libc.so.6!__assert_fail_base(const char * fmt, const char * assertion, const char * assertion@entry, const char * file, const char * file@entry, unsigned int line, unsigned int line@entry, const char * function, const char * function@entry) (assert.c:96)
libc.so.6!__assert_fail(const char * assertion, const char * file, unsigned int line, const char * function) (assert.c:105)
libvcmi.so!CBonusSystemNode::attachTo(class CBonusSystemNode * const this, class CBonusSystemNode & parent) (/home/user/vcmi/lib/bonuses/CBonusSystemNode.cpp:187)
libvcmi.so!CGHeroInstance::attachToBonusSystem(class CGHeroInstance * const this, class CGameState & gs) (/home/user/vcmi/lib/mapObjects/CGHeroInstance.cpp:1355)
libvcmi.so!CGameState::buildBonusSystemTree(class CGameState * const this) (/home/user/vcmi/lib/gameState/CGameState.cpp:1516)
libvcmi.so!CGameState::init(class CGameState * const this, const class IMapService * mapService, struct StartInfo * si, class IGameRandomizer & gameRandomizer, class Load::ProgressAccumulator & progressTracking, bool allowSavingRandomMap) (/home/user/vcmi/lib/gameState/CGameState.cpp:220)
CGameHandler::init(class CGameHandler * const this, struct StartInfo * si, class Load::ProgressAccumulator & progressTracking) (/home/user/vcmi/server/CGameHandler.cpp:576)
CVCMIServer::prepareToStartGame(class CVCMIServer * const this) (/home/user/vcmi/server/CVCMIServer.cpp:271)
ApplyOnServerNetPackVisitor::visitLobbyStartGame(class ApplyOnServerNetPackVisitor * const this, struct LobbyStartGame & pack) (/home/user/vcmi/server/NetPacksLobbyServer.cpp:210)
libvcmi.so!LobbyStartGame::visitTyped(struct LobbyStartGame * const this, class ICPackVisitor & visitor) (/home/user/vcmi/lib/networkPacks/NetPacksLib.cpp:728)
libvcmi.so!CPack::visit(struct CPack * const this, class ICPackVisitor & visitor) (/home/user/vcmi/lib/networkPacks/NetPacksLib.cpp:29)
CVCMIServer::handleReceivedPack(class CVCMIServer * const this, class std::shared_ptr<GameConnection> connection, struct CPackForLobby & pack) (/home/user/vcmi/server/CVCMIServer.cpp:368)
CVCMIServerPackVisitor::visitForLobby(class CVCMIServerPackVisitor * const this, struct CPackForLobby & packForLobby) (/home/user/vcmi/server/CVCMIServer.cpp:58)
libvcmi.so!CPackForLobby::visitBasic(struct CPackForLobby * const this, class ICPackVisitor & visitor) (/home/user/vcmi/lib/networkPacks/NetPacksLib.cpp:54)
libvcmi.so!CPack::visit(struct CPack * const this, class ICPackVisitor & visitor) (/home/user/vcmi/lib/networkPacks/NetPacksLib.cpp:24)
```

Triggers on:
```
void CGHeroInstance::attachToBonusSystem(CGameState & gs)
{
	CArmedInstance::attachToBonusSystem(gs);
	if (boardedBoat.hasValue())
	{
		auto boat = gs.getObjInstance(boardedBoat);
		if (boat)
			attachTo(dynamic_cast<CGBoat&>(*boat));
	}
}
```

IvanSavenko:
Looks like hero was attached to boat twice. Which is definitely a bug. Please try to investigate. Your map is tiny with minimal number of non-constant bonus system nodes, so should be doable.

Because there might be other side effects that basic testing and asserts have not covered. For example if hero is added twice to map.

Laserlicht:
Bonus is added twice:

### No 1
https://github.com/vcmi/vcmi/blob/c7a1558c81f6f0c85454471d9c0fc0beb2d4e68f/lib/gameState/CGameState.cpp#L633
https://github.com/vcmi/vcmi/blob/c7a1558c81f6f0c85454471d9c0fc0beb2d4e68f/lib/mapObjects/CGHeroInstance.cpp#L1322

### No 2
https://github.com/vcmi/vcmi/blob/c7a1558c81f6f0c85454471d9c0fc0beb2d4e68f/lib/mapObjects/CGHeroInstance.cpp#L1344

Do you think fix is correct (at least there is no assert triggered)?

IvanSavenko:
Now game won't attach boat to hero when boarding boat during regular games 

1st case should happen on server when loading map (call initiated by CGameHandler), while 2nd should happen on client (this call goes through CClient class). However they both should operate on their own, separate copies of CGHeroInstance 

Can try to take a look into this on Tuesday

Laserlicht:
Okay. Removed it.

It is also working on non debug builds without issues. Only assert is a problem. So it can fixed seperatly.

IvanSavenko:
Something is off:
- Title bar is named simply "creature", should have been "Select Creatures"
- What's up with sorting? It is neither alphabetical, nor sorted by index. Ideally should be faction / level / upgrade
зображення

IvanSavenko:
Rock (and any other impassable terrain from mods) should be excluded from selection

IvanSavenko:
Boat - fix is simple. Remove `CGHeroInstance::attachToBonusSystem` and `CGHeroInstance::detachFromBonusSystem` methods.

Server will correctly attach boat to hero inside `CGHeroInstance::setBoat`, and client will do the same as part of `CGHeroInstance::restoreBonusSystem`. No need for custom actions in `attachToBonusSystem`/`detachFromBonusSystem`


And one more thing - no need to show creature count if no creature has been selected in this slot. Disable / don't create corresponding UI element

ToRRent1812:
Maybe rename battle only to simulate battle?

IvanSavenko:
Battle only (or something similar, like battle only mode) is fine IMO.

Personally would've called it "Duel mode", but in H3 "Duel" often refers to that RMG template

Laserlicht:
Fixed mentioned points.

PR #6244, opened 2025-10-24
Try to fix tests not running on some presets

IvanSavenko:


IvanSavenko:
Seems to be working now - all 4 test configurations now run all active tests instead of 2 configuration with tests & 2 no-op configurations

test/events/ApplyDamageTest.cpp
```diff
@@ -22,9 +22,11 @@ namespace test
 using namespace ::testing;
 using namespace ::events;
 
-class ListenerMock
+class ApplyDamageListenerMock
 {
 public:
+	virtual ~ApplyDamageListenerMock() = default;
```
kambala-decapitator, line 28: not quite clear what's the purpose of the virtual destructor if nobody inherits from this class

IvanSavenko: This particular class - no.

But if somebody were to use `StrictMock<ApplyDamageListenerMock>` without virtual destructor it would lead to strange crashes inside Google Test codebase, since StrictMock inherits from template parameter

PR #6254, opened 2025-10-27
innoextract update

Laserlicht:
Merge upstream of innoextract into our fork. Support inno setup up to 6.4.0 (if H3 setup version might change in future).


IvanSavenko:
Was import of h3 gog.com installer tested with this new innosetup? Just to be safe?

Laserlicht:
Yes. I've tested it on one platform (linux). Should be save.

Laserlicht:
Tested also on android and with chronicles. Works.

PR #6257, opened 2025-10-27
Fixes mapeditor crashing when user changes terrain beyond map

Opuszek:
As in title.

I also added fixes suggested by kambala-decapitor's post-merge code review of #6173.

### DIFF
--- a/mapeditor/scenelayer.cpp
+++ b/mapeditor/scenelayer.cpp
@@ -170,7 +170,8 @@ void AbstractViewportLayer::redrawWithSurroundingTiles(const std::vector<int3> &
 
 void AbstractViewportLayer::redraw(const std::set<CGObjectInstance *> & objects)
 {
-	std::vector<QRectF> areas(objects.size());
+	std::vector<QRectF> areas;
+	areas.reserve(objects.size());
 	for (const CGObjectInstance * object : objects)
 	{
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
does it fix #6260?

just tested the artifact: doesn't seem so

mapeditor/mapview.cpp
```diff
@@ -675,7 +675,7 @@ void MapView::dragLeaveEvent(QDragLeaveEvent * event)
 
 void MapView::setViewports()
 {
-	if(auto * sc = dynamic_cast<MapScene*>(scene()))
+	if(auto * sc = static_cast<MapScene*>(scene()))
```
kambala-decapitator: you don't need `if` here

IvanSavenko: Why was dynamic_cast even changed to static_cast?

Sure, might lead to 0.001ms performance improvement, but might also cause weird crashes if `scene()` were ever to return something that is not `MapScene*`

Opuszek: Because I have no spine and change it whenever it is suggested :P. I first changed it from static to dynamic because my complainer compiled... my compiler complained, then I changed it to static because kambala-decapitator asked about it. 
I'll remove this change. I understand the rationale for both casts, and I don't think it is worthy further deliberations.

PR #6261, opened 2025-10-29
Implements undo and redo operations for terrain/river/road changes in map editor

Opuszek:
As in the title.

Please take a better look at the second commit, it was the first time I use a template class in C++ and it's more complicated than I expected to be frank.

### DIFF
--- a/lib/mapping/CDrawRoadsOperation.cpp
+++ b/lib/mapping/CDrawRoadsOperation.cpp
@@ -20,7 +20,8 @@
 
 VCMI_LIB_NAMESPACE_BEGIN
 
-const std::vector<CDrawLinesOperation::LinePattern> CDrawLinesOperation::patterns =
+template <typename T>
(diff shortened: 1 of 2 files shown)


lib/mapping/CMapOperation.cpp
```diff
@@ -103,32 +103,48 @@ void CDrawTerrainOperation::execute()
...
+	{
+		auto & tile = map->getTile(pos);
+		tile.terrainType = terrainType;
 		invalidateTerrainViews(pos);
 	}
 
-	updateTerrainTypes();
+	updateTerrainTypes(selection);
 	updateTerrainViews();
+	invalidatedTerViews.clear();
```
C3pa, line 124: Why not call this as the last line in `updateTerrainViews`?

Opuszek: DrawTerrain method inserts the first objects into invalidatedTerViews, so IMO it makes sense to make it responsible for clearing it at the end.

PR #6262, opened 2025-10-29
Battle fixes: Sacrifice targeting. GUI: cursor init stall fix

MichalZr6:
This PR targets:

- Disallow Sacrifice on alive, undead or mechanical targets during first phase of the spell

- Disallow Sacrifice on undead or mechanical targets during second phase of the spell

- ~~First Aid Tent: lessen possible actions; forbid self-targeting (fixes crash on that action).~~ Fixed in #6296 

- Initialize cursor `currentFrame` to 0 to avoid first-hover animation stall. It delayed any GUI display with 2-3 seconds (depending on software) every time `CursorHandler` was initialized.

### DIFF
--- a/client/battle/BattleActionsController.cpp
+++ b/client/battle/BattleActionsController.cpp
@@ -264,6 +264,7 @@ void BattleActionsController::reorderPossibleActionsPriority(const CStack * stac
 			case PossiblePlayerBattleAction::NO_LOCATION:
 			case PossiblePlayerBattleAction::FREE_LOCATION:
 			case PossiblePlayerBattleAction::OBSTACLE:
+			case PossiblePlayerBattleAction::SACRIFICE:
 				if(!stack->hasBonusOfType(BonusType::NO_SPELLCAST_BY_DEFAULT) && targetStack != nullptr)
 				{
 					PlayerColor stackOwner = owner.getBattle()->battleGetOwner(targetStack);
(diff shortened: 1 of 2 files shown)


lib/spells/BattleSpellMechanics.cpp
```diff
@@ -267,6 +267,9 @@ bool BattleSpellMechanics::canBeCastAt(const Target & target, Problem & problem)
 		mainTarget = battle()->battleGetUnitByPos(target.front().hexValue, true);
 	}
 
+	if(mainTarget && getSpell()->getId() == SpellID::SACRIFICE && spellPhase == ESpellCastPhase::FIRST_PHASE && mainTarget->alive())
+		return adaptProblem(ESpellCastProblem::WRONG_SPELL_TARGET, problem);
+
```
IvanSavenko: This should be checked as part of `Sacrifice::applicable`, not on `Mechanics` level, and without spell ID check

Oh, perhaps instead of introducing `ESpellCastPhase` we should add new "problem", e.g. `ESpellCastProblem::TARGET_INCOMPLETE` and return it if 1st target is valid, but second target does not exists. This would allow client code to detect this particular scenario - only one target, but correctly selected, and react accordingly.

BTW - please also check Teleport spell, another one that consists from 2 different targets

MichalZr6: Done via `applicable`. The code handles now case when first target is valid but there is no valid second target.
Also not possible now to select mechanical and undead units.
After merge of 
- #6263 

I will implement some extended on-hover status bar info for Sacrifice.

Issue #6183, opened 2025-09-28
Show how many by „Animate Dead“ rise


Schmogg:
Hi, I know it from HotA that when you hover the spell over a creature it shows you at the bottom of the description how many rise again.

Can this be adjusted or implemented?

Device: iPad
Version: latest

![Image](https://github.com/user-attachments/assets/b54dd962-9919-4833-94c8-da507e118197)

Alucard648:
Same goes to Resurrection.

MichalZr6:
I started working on this issue yesterday. Tested with animate dead for now and it already works on my local branch. l will also try to implement battle log text with estimated damage/heal values for other spells.

PR #6263, opened 2025-10-29
Battle hover: show spell impact (damage, heal, raised/summoned), incl. AoE, spell-like attack and First Aid Tent

MichalZr6:
This PR adds real-time previews of what a spell will do when you hover a target in battle:

- Damage / kills for offensive spells and spell-like ranged effects (e.g., Lich Death Cloud), with totals across all affected stacks.
- Healed HP for positive single-target spells.
- Raised units for Resurrection / Animate Dead.
- Summoned units count for Summon Demons.

Under the hood:
- Mechanics now expose battle effects (`forEachEffect`, `findEffect<T>()`) so we can detect things like DemonSummon directly.
- ~~Added `Spell::isRising()` - missing "getter" for `bool rising` property.~~ (turned out unneeded)
- Other minor tweaks.

Fixes:
- #6183

### DIFF
--- a/lib/spells/effects/Damage.cpp
+++ b/lib/spells/effects/Damage.cpp
@@ -21,7 +21,7 @@
 #include "../../texts/CGeneralTextHandler.h"
 #include "../../texts/Languages.h"
 #include "../../serializer/JsonSerializeFormat.h"
-#include "../AI/BattleAI/StackWithBonuses.h"
+#include "../lib/CRandomGenerator.h"
 
 #include <vcmi/spells/Spell.h>
 
@@ -91,7 +91,7 @@ SpellEffectValue Damage::getHealthChange(const Mechanics * m, const EffectTarget
 	SpellEffectValue result {};
 
 	size_t targetIndex = 0;
-	RNGStub fakeRng;
+	CRandomGenerator fakeRng;
 
 	for(auto const & t : spellTarget)
 	{


lib/battle/IBattleInfoCallback.h
```diff
@@ -43,6 +43,12 @@ struct DamageEstimation
 	DamageRange kills;
 };
 
+struct SpellEffectValue
+{
+	int64_t hpDelta = 0;    // +heal / -damage
+	int32_t unitsDelta = 0; // +resurrected / -killed
+};
+
```
IvanSavenko: To be honest this does not sounds to be different from existing `DamageEstimation`, but rather one that represents fixed damage.

Sure, this is not necessarily *damage*, but also includes heal. But other than that? Sounds like the same entity. So perhaps reuse existing `DamageEstimation` struct?

lib/battle/CBattleInfoCallback.h
```diff
@@ -111,6 +111,15 @@ class DLL_LINKAGE CBattleInfoCallback : public virtual CBattleInfoEssentials
...
 	DamageEstimation battleEstimateDamage(const battle::Unit * attacker, const battle::Unit * defender, int getMovementRange, DamageEstimation * retaliationDmg = nullptr) const;
 
+	/// preview of damage / restored HP and units killed or raised/summoned for given parameters
+	/// returns hpDelta and unitsDelta
+	SpellEffectValue getSpellEffectValue(const CSpell * spell, const spells::Caster * caster, const spells::Mode spellMode, const BattleHex & targetHex) const;
+
+	/// damage estimation for spell-like-attack case, eg. Death Cloud
+	DamageEstimation estimateSpellLikeAttackDamage(const battle::Unit * shooter, const CSpell * spell,const BattleHex & aimHex) const;
+
+	int64_t getFirstAidHealValue(const CGHeroInstance * owner, const battle::Unit * target) const;
```
IvanSavenko, line 141: Where this method is used? Can't see any usages of this method

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -808,6 +809,134 @@ DamageEstimation CBattleInfoCallback::battleEstimateDamage(const battle::Unit *
...
+
+	if(spell->isDamage())
+	{
+		for(auto const * target : targets)
+		{
+			result.hpDelta += std::max(-target->getAvailableHealth(), -effectValue);
+			result.unitsDelta += std::max(-target->getCount(), int32_t(-effectValue / target->getMaxHealth()));
+		}
+		return result;
+	}
```
IvanSavenko: This should call `Damage::damageForTarget` method

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -808,6 +809,134 @@ DamageEstimation CBattleInfoCallback::battleEstimateDamage(const battle::Unit *
...
+		{
+			// TODO: handle AOE heal spell ? Is there anything like that ?
+			auto state = targets.front()->acquireState();
+
+			HealInfo healInfo;
+			if(spell->isRising())
+				healInfo = state->heal(effectValue, EHealLevel::RESURRECT, EHealPower::ONE_BATTLE);
+
+			if(spell->getId() == SpellID::CURE)
+				healInfo = state->heal(effectValue, EHealLevel::HEAL, EHealPower::PERMANENT);
```
IvanSavenko: This should not rely on spell flags, but rather call some method of `Heal` effect. And please avoid spell ID checks.

What about this:
- add method `UnitEffect::targetHealthChange`, with no-op implementation, that accepts Target and returns `DamageEstimation` struct (probably negative for heal, to indicate heal?)
- implement this method in Damage/Heal/DemonSummon classes
- use something like `mech->findEffect<spells::effects::UnitEffect>()->targetHealthChange` to get effect power in identical way instead of multiple if branches

lib/battle/IBattleInfoCallback.h
MichalZr6: Yes, I was also considering the reuse of DamageEstimation but thought it will be not clear enough when used with heal / resurrected count code.

IvanSavenko: One option would be to rename `DamageEstimation` to something like `UnitHealthDelta` / `HealthDeltaEstimation`

MichalZr6: What about something like `SpellOrAttackResult` with `unitsDelta` and `hpDelta` which will fulfill the purpose of the struct?

IvanSavenko: Not sure about "SpellOrAttack" part. Should we add one more 'or' whenever struct is used for something else?

Name of class should tell what this class contains/implements. Not how it is used. `HealthChangeEstimation` or `UnitHealthChangeEstimation` feels to be the best option for me at the moment.

MichalZr6: Implemented that unified struct and - TBH - I don't like it.
Keeping them separate is much cleaner. Reasons:
- they describe a bit different semantics. One type trying to do both raises: “is this fixed or a range?”
- no sign gymnastics required
- making a single `HealthChangeEstimation` safe / clear to use would need some extra code - feels like overengineering

lib/battle/CBattleInfoCallback.h
MichalZr6: I must have probably reverted that code. Now it's in use, but I need to test it.

MichalZr6:
I fixed `estimateSpellLikeAttackDamage` calculations and also small bug in `DamageCalculator` to narrow damage by available health of defender.
I suppose it's ready. After it gets approved I will squash `!fixup` commit.

lib/spells/effects/DemonSummon.cpp
```diff
@@ -101,6 +102,22 @@ void DemonSummon::apply(ServerCallback * server, const Mechanics * m, const Effe
...
+	SpellEffectValue result;
+
+	auto targets = m->getAffectedStacks(spellTarget);
+
+	if(targets.empty())
+		return result;
+
+	auto unit = targets.front();
+	if(unit)
+		result.unitsDelta = raisedCreatureAmount(m, unit);
```
IvanSavenko, line 117: Perhaps also fill health delta using type of raised unit?

lib/spells/effects/Damage.cpp
```diff
@@ -85,6 +85,24 @@ void Damage::apply(ServerCallback * server, const Mechanics * m, const EffectTar
...
+	SpellEffectValue result {};
+
+	auto targets = m->getAffectedStacks(spellTarget);
+
+	size_t targetIndex = 0;
+	for(auto const * target : targets)
+	{
+		auto dmgForTarget = damageForTarget(targetIndex, m, target);
+		result.hpDelta += -dmgForTarget;
+		result.unitsDelta += std::max(-target->getCount(), int32_t(-dmgForTarget / target->getMaxHealth()));
```
IvanSavenko: What if our target has 1 hp out of 200 left and spell deals 100 damage?

Your code would result in 0 kills, while actually 1 unit will be killed

lib/battle/DamageCalculator.cpp
```diff
@@ -572,6 +572,12 @@ DamageEstimation DamageCalculator::calculateDmgRange() const
 
 	DamageRange killsDealt = getCasualties(damageDealt);
 
+	if(killsDealt.min > 0)
+	{
+		damageDealt.min = std::min(killsDealt.min * info.defender->getMaxHealth(), damageDealt.min);
+		damageDealt.max = std::min(killsDealt.max * info.defender->getMaxHealth(), damageDealt.max);
+	}
+
```
IvanSavenko: Sounds like workaround. Why code that have set killsDealt have not set damageDealt correctly? 

Will also result in incorrect damage info if target is not at full health - you don't need to deal 200 damage to kill unit with 1/200 hp

lib/spells/effects/DemonSummon.cpp
MichalZr6: Sure, done.

lib/spells/effects/Damage.cpp
MichalZr6: I fixed that, but then I've found out that chain lightning damage is incorrect. I went for a new approach on that code.

lib/battle/DamageCalculator.cpp
MichalZr6: :+1: Done.

MichalZr6:
- squashed approved feature commits
- dropped commit with unused `isRising()` code
- addressed review

I consider it ready :wink:

PR #6270, opened 2025-11-01
Add ucrt64 build instructions

mrhaandi:
Also link to VCMI_lib: WinSock2 (ws2_32) + socket extensions (mswsock) + Debug Help API (dbghelp) in MinGW64 and UCRT64. 

Why UCRT? see [msys2.org/docs/environments](https://www.msys2.org/docs/environments/)
- UCRT links against Microsoft’s Universal CRT (UCRT), which has much better C99/C11 compliance and fewer long-standing quirks than the old MSVCRT used by mingw64.
- UCRT is the same CRT used by Visual Studio 2015–2022. That makes cross-toolchain behavior more consistent and reduces surprises when sharing code/artifacts with MSVC builds.

### DIFF
--- a/docs/developers/Building_Windows.md
+++ b/docs/developers/Building_Windows.md
@@ -106,7 +106,7 @@ call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\devenv
 2. Open the correct shell
    - For MinGW64 (MSVCRT): start `MSYS2 MinGW x64`
    - For UCRT64: start `MSYS2 UCRT64`
-   
+
    (Sanity check: `echo $MSYSTEM` should be MINGW64 or UCRT64; don’t mix them.)
 3. Update MSYS2 packages: `pacman -Syu`
 4. Install dependencies


lib/CMakeLists.txt
```diff
@@ -868,6 +868,10 @@ if(WIN32)
 			OUTPUT_NAME "VCMI_lib"
 			PROJECT_LABEL "VCMI_lib"
 	)
+  if (MINGW)
+    # Link Windows system libs: WinSock2 (ws2_32) + socket extensions (mswsock) + Debug Help API (dbghelp)
+    target_link_libraries(vcmi PRIVATE ws2_32 mswsock dbghelp)
+  endif()
```
kambala-decapitator: please format with tabs just like the rest of the file

docs/developers/Building_Windows.md
```diff
@@ -100,12 +100,19 @@ call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\devenv
...
+1. Install MSYS2 from <https://www.msys2.org/>
+2. Open the correct shell
+   - For MinGW64 (MSVCRT): start `MSYS2 MinGW x64`
+   - For UCRT64: start `MSYS2 UCRT64`
+   
+   (Sanity check: `echo $MSYSTEM` should be MINGW64 or UCRT64; don’t mix them.)
+3. Update MSYS2 packages: `pacman -Syu`
+4. Install dependencies
+   - For MinGW64 (MSVCRT): `pacman -S mingw-w64-x86_64-SDL2 mingw-w64-x86_64-SDL2_image mingw-w64-x86_64-SDL2_mixer mingw-w64-x86_64-SDL2_ttf mingw-w64-x86_64-boost mingw-w64-x86_64-gcc mingw-w64-x86_64-ninja mingw-w64-x86_64-qt5-static mingw-w64-x86_64-qt5-tools mingw-w64-x86_64-tbb`
+   - For UCRT64: `pacman -S --needed mingw-w64-ucrt-x86_64-cmake mingw-w64-ucrt-x86_64-gcc mingw-w64-ucrt-x86_64-ninja mingw-w64-ucrt-x86_64-boost mingw-w64-ucrt-x86_64-minizip mingw-w64-ucrt-x86_64-ffmpeg mingw-w64-ucrt-x86_64-SDL2_image mingw-w64-ucrt-x86_64-SDL2_mixer mingw-w64-ucrt-x86_64-SDL2_ttf mingw-w64-ucrt-x86_64-qt5-static mingw-w64-ucrt-x86_64-tbb`
```
kambala-decapitator, line 114: perhaps it's easier to write a single list of packages using a variable? like
```
ucrt64Suffix=-ucrt # set only for UCRT64
pacman -S --needed mingw-w64$ucrt64Suffix-x86_64-cmake ...
```

mrhaandi: I think it would be overengineering for just a single line. Also, I did not change the MinGW line (which already diverges somewhat from ucrt).

lib/CMakeLists.txt
mrhaandi: good catch, done

PR #6272, opened 2025-11-01
Fix preserveCountSuffix if no parentheses are present

mrhaandi:
This PR fixes list of resolution scaling options display (before it just crashed when trying to show the options).
Basically, `GeneralOptionsTab.cpp` calls `ENGINE->windows().createAndPushWindow<CObjectListWindow>` with entries like `50%`, `60%`, ...
When those are passed to `trimTextIfTooWide` with the option `preserveCountSuffix`, then one searches for parentheses (which are missing) and crashes upon using `std::string::npos` as bounds.

### DIFF
--- a/client/windows/GUIClasses.cpp
+++ b/client/windows/GUIClasses.cpp
@@ -1676,9 +1676,12 @@ void CObjectListWindow::trimTextIfTooWide(std::string & text, bool preserveCount
 	{
 		auto posBrace = text.find_last_of("(");
 		auto posClosing = text.find_last_of(")");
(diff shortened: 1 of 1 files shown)


IvanSavenko (approved):
OK as far as workaround goes, but may lead to unexpected behavior.

Goal was to shorten names of objects while preserving count, e.g. "Library of Enlignment (10)" -> "Library of ... (10)".

But if brackets are used for something other than number, this logic might be undesired. Proper solution is to correctly propagate flag `preserveCountSuffix` into this method

PR #6278, opened 2025-11-02
Migrate codebase to C++20

IvanSavenko:
- VCMI can now be compiled in C++20 mode
- Replaced all references to C++17 with C++20
- Boost 1.74 is now set as minimal version (older version might work but untested)
- Updated documentation to reflect required versions of compilers / libraries
- Removed excessive fail-fast / continue-on-error from CI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,8 +223,9 @@ include(UseDoxygen OPTIONAL)
 #        Compile and linking options       #
 ############################################
 
-#Enable C++17 Globally
-set (CMAKE_CXX_STANDARD 17)
+set (CMAKE_CXX_STANDARD 20)
+set (CMAKE_CXX_STANDARD_REQUIRED ON)
+
 #General visibility options
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
@@ -423,18 +424,8 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR NOT WIN32)
 endif()
 
 # Check if some platform-specific libraries are needed for linking
-if(NOT WIN32 AND NOT IOS)
-	include(CheckLibraryExists)
-
-	# Shared memory functions used by Boost.Interprocess
-	# FindBoost handle linking with pthreads, but doesn't handle this
-	CHECK_LIBRARY_EXISTS(rt shm_open "" HAVE_RT_LIB)
-	if(HAVE_RT_LIB)
-		set(SYSTEM_LIBS ${SYSTEM_LIBS} rt)
-	endif()
-	if(HAIKU)
-		set(SYSTEM_LIBS ${SYSTEM_LIBS} network)
-	endif()
+if(HAIKU)
+	set(SYSTEM_LIBS ${SYSTEM_LIBS} network)
 endif()
 
 if(ENABLE_LUA)
@@ -449,10 +440,10 @@ set(BOOST_COMPONENTS date_time filesystem locale program_options)
 if(ENABLE_INNOEXTRACT)
 	list(APPEND BOOST_COMPONENTS iostreams)
 endif()
-find_package(Boost 1.48.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
+find_package(Boost 1.74.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
 if(Boost_MAJOR_VERSION EQUAL 1 AND Boost_MINOR_VERSION LESS 69)
 	list(APPEND BOOST_COMPONENTS system)
-	find_package(Boost 1.48.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
+	find_package(Boost 1.74.0 REQUIRED COMPONENTS ${BOOST_COMPONENTS})
 endif()
 
 find_package(ZLIB REQUIRED)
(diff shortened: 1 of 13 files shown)


client/renderSDL/FontChain.cpp
```diff
@@ -116,7 +116,7 @@ size_t FontChain::getGlyphWidthScaled(const char * data) const
 std::vector<FontChain::TextChunk> FontChain::splitTextToChunks(const std::string & data) const
 {
 	// U+FFFD - replacement character (question mark in rhombus)
-	static const std::string replacementCharacter = u8"�";
+	static const std::string replacementCharacter = reinterpret_cast<const char *>(u8"�");
```
kambala-decapitator: this character looks as if rendering is incorrect until you read the comment. I suggest to use `\u` literal to make everything clear

looks like we still can't use `u8string` here?

docs/developers/Building_Linux.md
```diff
@@ -15,8 +15,8 @@ To compile, the following packages (and their development counterparts) are need
 - SDL2 with devel packages: mixer, image, ttf
 - minizip or minizip-ng
 - zlib and zlib-devel
-- Boost C++ libraries v1.48+: program-options, filesystem, system, thread, locale
-- Recommended, if you want to build launcher or map editor: Qt 5, widget and network modules
+- Boost C++ libraries: program-options, filesystem, system, thread, locale
```
kambala-decapitator, line 18: the list of modules is a bit trickier, see CML

docs/developers/Coding_Guidelines.md
```diff
@@ -2,9 +2,9 @@
 
 ## C++ Standard
 
-VCMI implementation bases on C++17 standard. Any feature is acceptable as long as it's will pass build on our CI, but there is list below on what is already being used.
+VCMI implementation bases on C++20 standard. Any feature is acceptable as long as it's will pass build on our CI, but there is list below on what is already being used.
```
kambala-decapitator: you may want to add that C++20 library features (as opposed to the language features) will most likely fail on Apple platforms

> there is list below

where? :)

CMakeLists.txt
```diff
@@ -223,8 +223,8 @@ include(UseDoxygen OPTIONAL)
 #        Compile and linking options       #
 ############################################
 
-#Enable C++17 Globally
-set (CMAKE_CXX_STANDARD 17)
+#Enable C++20 Globally
```
kambala-decapitator: is the comment really needed?

CMakeLists.txt
```diff
@@ -223,8 +223,8 @@ include(UseDoxygen OPTIONAL)
...
+set (CMAKE_CXX_STANDARD 20)
```
kambala-decapitator, line 226: I think we should also set `CMAKE_CXX_STANDARD_REQUIRED`

docs/developers/Building_Linux.md
IvanSavenko: True, but as far as building on Linux goes, this is irrelevant. 

Most/all Linux distros only package boost libraries with runtime component as separate package. And all header-only boost libraries are part of the base package, e.g. `libboost-dev` on Ubuntu.

docs/developers/Coding_Guidelines.md
IvanSavenko: You mean C++20 dynamic/shared library features? As in - those that are marked as having "Minimum deployment target" in this list, correct?

 https://developer.apple.com/xcode/cpp/#c++20

client/renderSDL/FontChain.cpp
IvanSavenko: > looks like we still can't use u8string here?

Problem is, char8_t* is not conversible to char* (and same for corresponding string types). Which kind of makes sense, but as a result you can't use `u8string` unless you use it everywhere. Even worse - since char8_t can't be converted to char, any 3rd-party library that accepts either std::string's or char*'s will now require explicit conversions.

So at least so far looks like good idea but poor implementation

docs/developers/Coding_Guidelines.md
kambala-decapitator: yes exactly

kambala-decapitator:
btw were there build issues with the old NDK (clang 14)?

IvanSavenko:
> btw were there build issues with the old NDK (clang 14)?

Sort of. Build itself went fine, but pretty much any c++20 feature I attempted to use (std ranges, <=> operator) resulted in compilation failure on Android for one reason or another.

client/renderSDL/FontChain.cpp
kambala-decapitator: so will you use `\u` literal?

CMakeLists.txt
```diff
@@ -223,8 +223,9 @@ include(UseDoxygen OPTIONAL)
 #        Compile and linking options       #
 ############################################
 
-#Enable C++17 Globally
-set (CMAKE_CXX_STANDARD 17)
+set (CMAKE_CXX_STANDARD 20)
+set (CMAKE_CXX_STANDARD_REQUIRED 20)
```
kambala-decapitator:
```suggestion
set (CMAKE_CXX_STANDARD_REQUIRED ON)
```

client/renderSDL/FontChain.cpp
IvanSavenko: Yeah, will do it tomorrow

GeorgeK1ng:
Why is needed to remove fail-fast and continue-on-error "cheats"?

IvanSavenko:
It considerably slows down CI since now it will only fail if (generally slow) MSVC build fails.

You have compilation error that would fail on all compilers? Okay, ios would fail in 1-2 minutes, but msvc would take 10-20 minutes of CI time (x3 for 32bit/64bit/arm)

PR #6279, opened 2025-11-02
Fix for ALWAYS_MAXIMUM_DAMAGE bonus

GeorgeK1ng:
Before
image

After
image


misiokles:
I guess there's should be description of this bonus in english.json as well (visible in creature window).

GeorgeK1ng:
Just tried it, when I select that bonus, it's not visible in Commanders bonus list. Probably needs another touch somewhere. Seems it currently fixes UI glitch only.

IvanSavenko:
> Just tried it, when I select that bonus, it's not visible in Commanders bonus list

Only bonuses that have description are visible. Adding one to english.json in vcmi mod should make bonus visible

GeorgeK1ng:
Yes, just noticed it. Adding description make it visible.

MichalZr6:
Any chance it needs some console log message on such issue ?

GeorgeK1ng:
There was console error, about that, but was harder to track why.
image

Issue #5607, opened 2025-04-05
Artifacts sorting in backpack


GeorgeK1ng:
Currently there is artifacts sorting, which work only in one way. Sorting by value do it only once. Would be better to allow sort in opposite order with another click so low-high, high-low. This logic should be applied to all sort buttons.

Corresponding graphics will be great aswel. 

![Image](https://github.com/user-attachments/assets/93f815b2-f2b6-4a2a-ae37-c8eed9d8cf10)

PR #6286, opened 2025-11-03
Enable backpack sorting in both ways

GeorgeK1ng:
It's now possible to sort in both orders in backpack. So you can now click on sort button to see artifacts in both directions.

Fixes https://github.com/vcmi/vcmi/issues/5607

Sadly scrolling using stylus doesn't work in this dialog. Unrelated to my change, but would be great to fix it too.

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -2729,75 +2729,88 @@ bool CGameHandler::manageBackpackArtifacts(const PlayerColor & player, const Obj
 	const auto * artSet = gameState().getArtSet(heroID);
 	COMPLAIN_RET_FALSE_IF(artSet == nullptr, "manageBackpackArtifacts: wrong hero's ID");
 
-	auto isSortableCmd = [](const ManageBackpackArtifacts::ManageCmd & cmd)
-	{
-		using MC = ManageBackpackArtifacts::ManageCmd;
-		return cmd == MC::SORT_BY_SLOT || cmd == MC::SORT_BY_COST || cmd == MC::SORT_BY_CLASS;
-	};
-
-	static std::map<ObjectInstanceID, std::pair<ManageBackpackArtifacts::ManageCmd, bool>> lastSort;
-	bool descending = false;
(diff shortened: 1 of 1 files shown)


server/CGameHandler.cpp
```diff
@@ -2729,27 +2729,56 @@ bool CGameHandler::manageBackpackArtifacts(const PlayerColor & player, const Obj
 	const auto * artSet = gameState().getArtSet(heroID);
 	COMPLAIN_RET_FALSE_IF(artSet == nullptr, "manageBackpackArtifacts: wrong hero's ID");
 
+	auto isSortableCmd = [](const ManageBackpackArtifacts::ManageCmd & cmd)
+	{
+		using MC = ManageBackpackArtifacts::ManageCmd;
+		return cmd == MC::SORT_BY_SLOT || cmd == MC::SORT_BY_COST || cmd == MC::SORT_BY_CLASS;
+	};
+
+	static std::map<ObjectInstanceID, std::pair<ManageBackpackArtifacts::ManageCmd, bool>> lastSort;
```
IvanSavenko: Please don't use static, non-const variables unless absolutely necessary.

Also - do we even need memory?

When player requests sorting, always try sorting in ascending order. If list is already sorted - reverse it.

GeorgeK1ng: Changed it, also sonar is now clear. What about touch issue in that window? Have tried to add _addUsedEvents(GESTURE);_ into _CHeroBackpackWindow.cpp_ to _CHeroBackpackWindow_ but seems there is no difference. I can use "long touch" to display artifacts info, but scrolling doesn't work there by touch in artifacts or slider.
On PC with mouse wheel it works fine..

IvanSavenko: Not sire what's wrong with touch. You can try using define in this file on PC to debug it without real touch input source https://github.com/vcmi/vcmi/blob/develop/client%2FeventsSDL%2FInputSourceTouch.h

PR #6291, opened 2025-11-04
Fix artifact->artifactID.hasValue() nullptr

### LINKED ISSUE #6289: Crash on unit split in multyplayer
**Describe the bug**
When playing in multiplayer(TCP/IP) non host player crashes the game when trying to split/combine units.

Not crashing only when non host splits/combines units with "split" button
For the host everything is working. 
In hotseat everything works for everyone

**Game logs**

[unit_crash.zip](https://github.com/user-attachments/files/23329170/unit_crash.zip)

**To Reproduce**

1. Host multiplayer game 
2. Join other player through TCP/IP
3. Try to split unit
4. Crash

**Did it work earlier?**
Yes (could not test last cca 2 months as there was another bug preventing MP games, but before that worked)

**Version**
daily build:
VCMI 1.7.0.200ca013347f3b6d00068404bc648e66dac56e25

- OS: Windows

mrhaandi:
Fixes #6289
Basically `CArtifactSet::getSlot` returns `nullptr` and then `artifactID.hasValue()` crashes.

Note: is only happens for non-host clients (while the logic should be identical for hosts as well).
Possibly, this is not the correct place to have `// TODO: rollback creature artifacts & stack experience`.

### DIFF
--- a/client/netlag/PackRollbackGeneratorVisitor.cpp
+++ b/client/netlag/PackRollbackGeneratorVisitor.cpp
@@ -52,9 +52,11 @@ void PackRollbackGeneratorVisitor::visitRebalanceStacks(RebalanceStacks & pack)
 	const auto * srcArmy = dynamic_cast<const CArmedInstance *>(srcObject);
 	const auto * dstArmy = dynamic_cast<const CArmedInstance *>(dstObject);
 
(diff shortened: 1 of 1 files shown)


mrhaandi:
This also possibly fixes #6282 (the unit buying part).

PR #6300, opened 2025-11-06
Fix terrain patterns

Opuszek:
I am trying to carefully fix problems with terrain patterns.
The first commit resolves those problems:

Pattern s1:

https://github.com/user-attachments/assets/e326af92-217e-431b-9a84-dc90138e5136

Pattern s4:

https://github.com/user-attachments/assets/93c49e15-5dd3-49ca-adaa-f1e714e5d5b3

Pattern m4:

https://github.com/user-attachments/assets/804649f4-96bf-42e3-8146-42c6fa98b2a4

The second commit resolves problems with dirt pattern matching like this:

https://github.com/user-attachments/assets/28aa3efb-c94d-4ea0-9895-0176a67ef7f2

The pattern that did not work is s4. 
The problem lies in the fact that non-standard rules apply only when the center tile is of the same type as a tile with non-standard rules. Normally it is correct behavior, but in the case of dirt terrain neighboring a tile of terrain with dirt transition (like lava) is identical to neighboring another dirt tile (since a dirt tile never display transitions). This behavior was considered for standard rules:

Screenshot from 2025-11-06 21-15-39

but it was never included in non-standard ones, creating this bug.

To be honest, patterns feel arbitrary, but I don't know how to better organize them, at least yet. For now all that vouches for there not being any more bugs is an extended doodling session I did:

Screenshot from 2025-11-09 00-05-13

What isn't much, but it is something.
It is better than it was, at least.

### DIFF
--- a/lib/mapping/CMapOperation.cpp
+++ b/lib/mapping/CMapOperation.cpp
@@ -421,9 +421,9 @@ CDrawTerrainOperation::ValidationResult CDrawTerrainOperation::validateTerrainVi
 			{
 				if(recDepth == 0 && map->isInTheMap(currentPos))
 				{
-					if(terType->getId() == centerTerType->getId())
+					if(centerTerType->getId() == terType->getId() || (centerTerType->getId() == ETerrainId::DIRT && !terType->isTransitionRequired()))
 					{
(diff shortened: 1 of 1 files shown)


Opuszek:
I found a bug created by the changes, pr changes to a draft for now.

Opuszek:
Bugs fixed, draft changes back to a pr.

PR #6301, opened 2025-11-06
tab support / add artifacts, skills, warmachines & spellbook for battle mode

Laserlicht:
Layout will adjusted later, when adding more features from #5303

grafik

### DIFF
--- a/client/lobby/BattleOnlyModeTab.cpp
+++ b/client/lobby/BattleOnlyModeTab.cpp
@@ -97,81 +97,9 @@ BattleOnlyModeTab::BattleOnlyModeTab()
 	title = std::make_shared<CLabel>(220, 35, FONT_BIG, ETextAlignment::CENTER, Colors::YELLOW, LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyMode"));
 	subTitle = std::make_shared<CMultiLineLabel>(Rect(55, 40, 333, 40), FONT_SMALL, ETextAlignment::BOTTOMCENTER, Colors::WHITE, LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeSubTitle"));
 
-	battlefieldSelector = std::make_shared<CButton>(Point(57, 552), AnimationPath::builtin("GSPButtonClear"), CButton::tooltip(), [this](){
-		std::vector<std::string> texts;
-		std::vector<std::shared_ptr<IImage>> images;
-
-		std::vector<std::shared_ptr<TerrainType>> terrains;
-		std::copy_if(LIBRARY->terrainTypeHandler->objects.begin(), LIBRARY->terrainTypeHandler->objects.end(), std::back_inserter(terrains), [](auto terrain) { return terrain->isPassable(); });
-		for (const auto & terrain : terrains)
-		{
-			texts.push_back(terrain->getNameTranslated());
-
-			const auto & patterns = LIBRARY->terviewh->getTerrainViewPatterns(terrain->getId());
-			TerrainViewPattern pattern;
-			for(auto & p : patterns)
-				if(p[0].id == "n1")
-					pattern = p[0];
-			auto image = ENGINE->renderHandler().loadImage(terrain->tilesFilename, pattern.mapping[0].first, 0, EImageBlitMode::OPAQUE);
-			image->scaleTo(Point(23, 23), EScalingAlgorithm::NEAREST);
-			images.push_back(image);
-		}
-
-		auto factions = LIBRARY->townh->getDefaultAllowed();
-		for (const auto & faction : factions)
-		{
-			texts.push_back(faction.toFaction()->getNameTranslated());
-
-			auto image = ENGINE->renderHandler().loadImage(AnimationPath::builtin("ITPA"), faction.toFaction()->town->clientInfo.icons[true][false] + 2, 0, EImageBlitMode::OPAQUE);
-			image->scaleTo(Point(35, 23), EScalingAlgorithm::NEAREST);
-			images.push_back(image);
-		}
-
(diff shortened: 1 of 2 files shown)


Laserlicht:
It's possible to add artifacts and skills now.

And enable warmachines and spellbook.

grafik

We've more features than battle mode in fheroes2 now. :)

Laserlicht:
Actual screenshots from discord

grafik

grafik

Laserlicht:
Settings are saved now. So if you open menu again it will be the same configuration as before. Only deleted after using reset button.

Last part of this PR. It's ready to review and merge @IvanSavenko . Further improvements in seperate PR if necessary.

Mods/vcmi/Content/config/english.json
```diff
@@ -139,12 +139,36 @@
 	"vcmi.lobby.deleteFolder" : "Do you want to delete following folder?",
 	"vcmi.lobby.deleteMode" : "Switch to delete mode and back",
 	"vcmi.lobby.battleOnlyMode" : "Battle only mode",
+	"vcmi.lobby.battleOnlyModeSubTitle" : "Select heroes and army for simple battle without adventure map",
```
IvanSavenko: No longer just heroes & armies?

client/lobby/BattleOnlyModeTab.cpp
```diff
@@ -0,0 +1,950 @@
...
+	{
+		JsonNode node = persistentStorage["battleModeSettings"];
+		if(!node.isNull())
+		{
+			node.setModScope(ModScope::scopeGame());
+			JsonDeserializer handler(nullptr, node);
+			startInfo->serializeJson(handler);
+		}
+	}
+	catch(std::exception & e)
```
IvanSavenko, line 89: I think only Identifier resolution exception can be thrown here. In that case - might be better to catch it specifically instead of generic exception?

client/mainmenu/CMainMenu.h
```diff
@@ -154,7 +154,7 @@ class CMainMenu final : public CIntObject, public std::enable_shared_from_this<C
 	void activate() override;
 	void onScreenResize() override;
 	void makeActiveInterface();
-	static void openLobby(ESelectionScreen screenType, bool host, const std::vector<std::string> & names, ELoadMode loadMode);
+	static void openLobby(ESelectionScreen screenType, bool host, const std::vector<std::string> & names, ELoadMode loadMode, bool battleMode);
```
IvanSavenko, line 157: Perhaps we should move battle only flag to ESelectionScreen enum?

client/lobby/BattleOnlyModeTab.cpp
Laserlicht: Theoretically yes. But there were also some serializer issues under some edge cases. Not sure if every edge case solved.
This is to be on the save side: Better to ignore the settings (and write new) than crash.

client/mainmenu/CMainMenu.h
Laserlicht: Also considered that, but `ESelectionScreen` is in heavy use and many `ESelectionScreen::newGame` would need an "OR" for `ESelectionScreen::battle`.

And from technical side we're still `ESelectionScreen::newGame` but just selecting another tab by default.

That's why I've given up this idea.

C3pa:
I'd consider changing the window name to be in title case to be consistent with the other parts of the HoMM3's menus.

image

Laserlicht:
> I'd consider changing the window name to be in title case to be consistent with the other parts of the HoMM3's menus.

Do you mean "Battle Only Mode"?

Changed.

C3pa (approved):

client/lobby/BattleOnlyModeTab.cpp
```diff
@@ -0,0 +1,952 @@
...
+		logGlobal->error("Error loading saved battleModeSettings, received exception: %s", e.what());
+	}
+
+	init();
+
+	backgroundImage = std::make_shared<CPicture>(ImagePath::builtin("AdventureOptionsBackgroundClear"), 0, 6);
+	title = std::make_shared<CLabel>(220, 35, FONT_BIG, ETextAlignment::CENTER, Colors::YELLOW, LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyMode"));
+	subTitle = std::make_shared<CMultiLineLabel>(Rect(55, 40, 333, 40), FONT_SMALL, ETextAlignment::BOTTOMCENTER, Colors::WHITE, LIBRARY->generaltexth->translate("vcmi.lobby.battleOnlyModeSubTitle"));
+
+	battlefieldSelector = std::make_shared<CButton>(Point(57, 552), AnimationPath::builtin("GSPButtonClear"), CButton::tooltip(), [this](){
```
IvanSavenko: Might be better to extract this (and similar) lambdas into separate method(s) for readability, to separate ui construction code from input feedback code

client/lobby/BattleOnlyModeTab.cpp
```diff
@@ -0,0 +1,952 @@
...
+	allSpells.erase(std::remove_if(allSpells.begin(), allSpells.end(), [](const SpellID& spell) {
+		return !spell.toSpell()->isCombat();
+	}), allSpells.end());
+	std::sort(allSpells.begin(), allSpells.end(), [](auto a, auto b) {
+		auto A = a.toSpell();
+		auto B = b.toSpell();
+		if(A->getLevel() < B->getLevel())
+			return true;
+		if(A->getLevel() > B->getLevel())
+			return false;
```
IvanSavenko: You can use form 
```
if a.level != b.level
    return a level < b.level
```

client/lobby/BattleOnlyModeTab.cpp
Laserlicht: Make sense.

Placed these big lambdas in separate functions.

PR #6296, opened 2025-11-06
Fix violated assert(direction != BattleHex::NONE) 

mrhaandi:
Fixes #6282 (the other issue mentioned there is already fixed).

Also this fixes the situation in which a healing tent considers attacking itself, which crashes for the same reason.
In general, only units with `isMeleeAttacker() == true` consider attacking in melee. This should be faster and more stable.

What happens: 
if the active unit X is surrounded (by fried or foe) and you hover over X, then deciding if `PossiblePlayerBattleAction::ATTACK` sees that X can reach itself and decides that it is attackable.
However, `owner.fieldController->fromWhichHexAttack(targetHex)` cannot determine from which direction X is supposed to attack itself and the `assert(direction != BattleHex::NONE)` is violated.

PR #6307, opened 2025-11-11
Refactor batte move / attack logic

### LINKED ISSUE #6302: getHighlightedHexesForMovementTarget flawed in moat
**Describe the bug**
When the target stands in a moat, then hovering for attack fails to display attacker position (now shadowed).

**To Reproduce**
Put defender in moat.

**Expected behavior**
Position of attacker during attack is shadowed.

**Actual behavior**
No shadow.

**Screenshots**
Image

**Additional context**
This exact issue is due to wrong use `owner.getBattle()->battleCanAttack(stack, hoveredStack, hoveredHex)` in `BattleFieldController::getHighlightedHexesForMovementTarget()`.
In particular, `battleCanAttack` should get as third argument the destination from which the attack happens, not the `hoveredHex` (where the target stands).

mrhaandi:
There are two goals.
1. Improve separation of game logic (CBattleInfoCallback) from UI logic (BattleFieldController, BattleActionsController).
2. Deduplicate logic code (written ad-hoc across several files, sometimes slightly differently).

Hopefully, this improves maintainability and removes / prevents bugs.
Some changes are prior to #6296 (to be updated after merge).
Comments are welcome.

List of changes:
- move fromWhichHexAttack to CBattleInfoCallback
- add toWhichHexMove (unifying incoherent duplicates)
- add battleGetOccupiableHexes
- add battleCanAttackHex for spatial attack check
- add battleCanAttackUnit for non-spatial attack check
- add headDirection to Unit (removing destShiftDir from CStack)
- remove redundant game logic from selectAttackDirection
- remove redundant game logic from BattleFieldController and BattleActionsController
- fix no consideration for double-wide tail attack in BattleFlowProcessor
- fix #6302 wrong moat stopping condition

### DIFF
--- a/client/battle/BattleFieldController.cpp
+++ b/client/battle/BattleFieldController.cpp
@@ -263,9 +263,10 @@ void BattleFieldController::showBackgroundImageWithHexes(Canvas & canvas)
 void BattleFieldController::redrawBackgroundWithHexes()
 {
 	const CStack *activeStack = owner.stacksController->getActiveStack();
-	BattleHexArray attackableHexes;
 	if(activeStack)
-		occupiableHexes = owner.getBattle()->battleGetAvailableHexes(activeStack, false, true, &attackableHexes);
+		availableHexes = owner.getBattle()->battleGetAvailableHexes(activeStack, false);
+	else
+		availableHexes.clear();
 
 	// prepare background graphic with hexes and shaded hexes
 	backgroundWithHexes->draw(background, Point(0,0));
@@ -274,13 +275,16 @@ void BattleFieldController::redrawBackgroundWithHexes()
 		owner.siegeController->showAbsoluteObstacles(*backgroundWithHexes);
 
 	// show shaded hexes for active's stack valid movement and the hexes that it can attack
-	if(settings["battle"]["stackRange"].Bool())
+	if(activeStack && settings["battle"]["stackRange"].Bool())
 	{
-		BattleHexArray hexesToShade = occupiableHexes;
-		hexesToShade.insert(attackableHexes);
-		for(const BattleHex & hex : hexesToShade)
+		auto occupiableHexes = owner.getBattle()->battleGetOccupiableHexes(availableHexes, activeStack);
+		for(si16 hex = 0; hex < GameConstants::BFIELD_SIZE; hex++)
 		{
-			showHighlightedHex(*backgroundWithHexes, cellShade, hex, false);
+			//shade occupiable and attackable hexes
+			if (occupiableHexes.contains(hex) ||
+				(owner.getBattle()->battleCanAttackUnit(activeStack, owner.getBattle()->battleGetStackByPos(hex, true)) &&
+					owner.getBattle()->battleCanAttackHex(availableHexes, activeStack, hex)))
+				showHighlightedHex(*backgroundWithHexes, cellShade, hex, false);
 		}
 	}
 
@@ -330,10 +334,7 @@ BattleHexArray BattleFieldController::getMovementRangeForHoveredStack()
 		return BattleHexArray();
 
 	auto hoveredStack = getHoveredStack();
-	if(hoveredStack)
-		return owner.getBattle()->battleGetAvailableHexes(hoveredStack, true, true, nullptr);
-	else
-		return BattleHexArray();
+	return hoveredStack ? owner.getBattle()->battleGetOccupiableHexes(hoveredStack, true) : BattleHexArray();
 }
 
 BattleHexArray BattleFieldController::getHighlightedHexesForSpellRange()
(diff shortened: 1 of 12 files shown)


lib/battle/Unit.h
```diff
@@ -139,6 +139,8 @@ class DLL_LINKAGE Unit : public IUnitInfo, public spells::Caster, public virtual
 
 	bool coversPos(const BattleHex & position) const; //checks also if unit is double-wide
 
+	BattleHex::EDir headDirection() const;
```
IvanSavenko, line 144: While this is rather simple method, please add comment to explain what it does, e.g. returns direction in which double-wide unit is facing, or EDir::NONE for single-wide units

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -613,89 +613,158 @@ BattleHexArray CBattleInfoCallback::battleGetAvailableHexes(const ReachabilityIn
...
+
+bool CBattleInfoCallback::battleCanAttackUnit(const battle::Unit * attacker, const battle::Unit * target) const
+{
+	RETURN_IF_NOT_BATTLE(false);
+
+	if(battleTacticDist())
+		return false;
+
+	if (!attacker || !target)
+		return false;
```
IvanSavenko: Perhaps throw an exception? Or at least add assert in addition to return? Because this looks like invalid call, and potentially broken state of caller. Especially null attacker

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -613,89 +613,158 @@ BattleHexArray CBattleInfoCallback::battleGetAvailableHexes(const ReachabilityIn
...
+
+bool CBattleInfoCallback::battleCanAttackHex(const battle::Unit * attacker, const BattleHex & position) const
+{
+	return battleCanAttackHex(battleGetAvailableHexes(attacker, false), attacker, position);
+}
+
+bool CBattleInfoCallback::battleCanAttackHex(const BattleHexArray & availableHexes, const battle::Unit * attacker, const BattleHex & position) const
+{
+	RETURN_IF_NOT_BATTLE(false);
+	if (!attacker || !position.isValid())
```
IvanSavenko: Similarly - this sounds like invalid call, and may reserve assert or even exception throwing

lib/battle/CBattleInfoCallback.cpp
```diff
@@ -613,89 +613,158 @@ BattleHexArray CBattleInfoCallback::battleGetAvailableHexes(const ReachabilityIn
...
+BattleHex CBattleInfoCallback::toWhichHexMove(const BattleHexArray & availableHexes, const battle::Unit * unit, const BattleHex & position) const
 {
 	RETURN_IF_NOT_BATTLE(false);
 
-	if(battleTacticDist())
+	if (!unit || !position.isValid())
+		return BattleHex::INVALID;
+
+	return 0;
+}
```
IvanSavenko: Not implemented method? Or I am missing something here?

client/battle/BattleFieldController.cpp
```diff
@@ -667,164 +652,36 @@ BattleHex BattleFieldController::getHexAtPosition(Point hoverPos)
...
+		testPoint[i] = hexPositionAbsolute(neighbours[i]).center();
 
 	// For bottom/top directions select central point, but move it a bit away from true center to reduce zones allocated to them
-	if (attackAvailability[6])
-		testPoint[6] = (hexPositionAbsolute(neighbours[0]).center() + hexPositionAbsolute(neighbours[1]).center()) / 2 + Point(0, -5);
-
-	if (attackAvailability[7])
-		testPoint[7] = (hexPositionAbsolute(neighbours[3]).center() + hexPositionAbsolute(neighbours[4]).center()) / 2 + Point(0,  5);
+	testPoint[6] = (hexPositionAbsolute(neighbours[0]).center() + hexPositionAbsolute(neighbours[1]).center()) / 2 + Point(0, -5);
+	testPoint[7] = (hexPositionAbsolute(neighbours[3]).center() + hexPositionAbsolute(neighbours[4]).center()) / 2 + Point(0,  5);
```
IvanSavenko: From what I remember, `neighbours` may contain invalid hexes, for which it is illegal to compute position. Because of that, this code was conditional before

mrhaandi:
For some reason, part of the modified code is missing from he PR. I'll restore that and test the modifications again (currently switched to draft).

lib/battle/Unit.h
mrhaandi: Yes, done.

lib/battle/CBattleInfoCallback.cpp
mrhaandi: I agree. I added throw std::runtime_error if attacker is null. Target is null is still a valid questions, which is answered negatively.

lib/battle/CBattleInfoCallback.cpp
mrhaandi: Similar to above. For `!position.isValid()` I log a warning for now. This is a legitimate question with the corresponding negative answer.

lib/battle/CBattleInfoCallback.cpp
mrhaandi: Sorry, somehow I dropped a commit since I tested the code. I reimplemented the method, but need to run tests again to make sure.

client/battle/BattleFieldController.cpp
mrhaandi: Good catch! Now, I added checks whether individual attack directions are valid and do not compute / test points which result from invalid directions.

lib/battle/CBattleInfoCallback.cpp
mrhaandi: Now it is implemented.

mrhaandi:
> Looks ok. Please resolve merge conflicts, but other than that, changes are ok from what I see

Rebased.

PR #6311, opened 2025-11-12
Prevent mid-word wrapping in component captions (artifact names)

### LINKED ISSUE #6305: Some artifacts names have wrong word breakings
Image

Other cases: Ambassaror's sash, Inexhaustible Cart of Wood, Inexhaustible Cart of Lumber

**Version**
- OS: Windows
- Version: pre 1.7 daily builds

MichalZr6:
In `CComponent::init()` set max to at least the width of the first word (via `getStringWidth(firstWord) + 8)` before calling `breakText()`. This avoids splitting long artifact names mid-word while keeping existing wrapping when genuinely needed.

- fixes #6305

### DIFF
--- a/client/widgets/CComponent.cpp
+++ b/client/widgets/CComponent.cpp
@@ -96,14 +96,29 @@ void CComponent::init(ComponentType Type, ComponentSubType Subtype, std::optiona
 	if(Type == ComponentType::RESOURCE && !ValText.empty())
 		max = 80;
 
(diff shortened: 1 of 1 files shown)


IvanSavenko:
Why only first word ? What about artifacts where 2nd word is extremely long? Perhaps pick length of longest word?

MichalZr6:
> Why only first word ? What about artifacts where 2nd word is extremely long? Perhaps pick length of longest word?

:+1: 
Done.

PR #6313, opened 2025-11-13
Custom map size in RMG

Laserlicht:
Last button "G" allows to set custom map size.

Shows also layers on the top right now to fully see RMG map size.

VCMI-extras should replace "G" button with "C" after this PR.

image

Warning for more than 2 layers (RMG support is not ready yet...):
image

### DIFF
--- a/client/lobby/RandomMapTab.cpp
+++ b/client/lobby/RandomMapTab.cpp
@@ -54,37 +54,7 @@ RandomMapTab::RandomMapTab():
 	
 	addCallback("toggleMapSize", [this](int btnId)
 	{
-		if(btnId == -1)
-			return;
-
-		auto mapSizeVal = getPossibleMapSizes();
-
-		auto setTemplateForSize = [this](){
-			if(mapGenOptions->getMapTemplate())
-				if(!mapGenOptions->getMapTemplate()->matchesSize(int3{mapGenOptions->getWidth(), mapGenOptions->getHeight(), mapGenOptions->getLevels()}))
-					setTemplate(nullptr);
-			updateMapInfoByHost();
-		};
-
-		if(btnId == mapSizeVal.size() - 1)
-		{
-			ENGINE->windows().createAndPushWindow<SetSizeWindow>(int3(mapGenOptions->getWidth(), mapGenOptions->getHeight(), mapGenOptions->getLevels()), [this, setTemplateForSize](int3 ret){
-				if(ret.z > 2)
-				{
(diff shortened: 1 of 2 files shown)


Laserlicht:
> Please, show an example how is to use this

Peek 2025-11-17 22-31

client/lobby/RandomMapTab.cpp
```diff
@@ -49,19 +52,43 @@ RandomMapTab::RandomMapTab():
 	recActions = 0;
 	mapGenOptions = std::make_shared<CMapGenOptions>();
 	
-	addCallback("toggleMapSize", [&](int btnId)
+	addCallback("toggleMapSize", [this](int btnId)
 	{
+		if(btnId == -1)
+			return;
```
IvanSavenko: In which cases button might have such ID?

client/lobby/RandomMapTab.cpp
```diff
@@ -49,19 +52,43 @@ RandomMapTab::RandomMapTab():
 	recActions = 0;
 	mapGenOptions = std::make_shared<CMapGenOptions>();
 	
-	addCallback("toggleMapSize", [&](int btnId)
+	addCallback("toggleMapSize", [this](int btnId)
```
IvanSavenko, line 55: Perhaps extract this entire lambda or parts of it into a separate method?

client/lobby/RandomMapTab.cpp
```diff
@@ -345,18 +372,19 @@ void RandomMapTab::setMapGenOptions(std::shared_ptr<CMapGenOptions> opts)
 	
 	if(auto w = widget<CToggleGroup>("groupMapSize"))
 	{
+		const auto & mapSizes = getPossibleMapSizes();
```
IvanSavenko: Perhaps rename this method to get standard map sizes or whatever is more suitable? Because now any map size is possible

client/lobby/RandomMapTab.cpp
Laserlicht: Also in this PR:
`w->setSelected(position == mapSizes.size() - 1 || opts->getWidth() != opts->getHeight() ? -1 : position);`

It's used to unselect every button. Otherwise if "Custom" button is selected we cannot select it again (e.g. for changing custom values).

Laserlicht:
Created button. To be replaced in `vcmi-extras` after merging.

[RanSizC.zip](https://github.com/user-attachments/files/23661440/RanSizC.zip)

scbt_cn

Issue #6309, opened 2025-11-11
NullKiller AI - SegFault


ti-dic:
**Describe the bug**
When AI attack a castle and win a segfault appear

[nullkiller_segfault.patch](https://github.com/user-attachments/files/23485356/nullkiller_segfault.patch)

mrhaandi:
The proposed patch may not fix the segfault - maybe it even contributes to a decoherent game state (which is more difficult to debug).
Could you elaborate on why the server may have not accepted the action?
Does this appear in single player or multi player? Maybe a save file can be helpful.
By "and win", you mean just the battle? Or the game?
More details are welcome.

ti-dic:
Hi, this appear in single player mode, i have the savefile.
on a no modified game the sigseg raise after "System message: Server encountered a problem: Got false in applying 16GarrisonHeroSwap"

The AI just win the battle not the game. The log of the game :


```
Starting... 
Starting client of 'VCMI 1.7.0'
Creating console and configuring logger: 0 ms
The log file will be saved to "/home/ti_dic/.cache/vcmi/VCMI_Client_log.txt"
        Initialization: 56 ms
        Data loading: 56 ms
        Mod handler: 24 ms
Potential confict detected between 'hota.heroes3datapatch' and 'reworked-commanders': both mods add file 'sounds/OORCWNCE'
Potential confict detected between 'hota.heroes3datapatch' and 'reworked-commanders': both mods add file 'sounds/OORCKILL'
Potential confict detected between 'reworked-commanders' and 'hota.heroes3datapatch': both mods add file 'sounds/OORCKILL'
Potential confict detected between 'reworked-commanders' and 'hota.heroes3datapatch': both mods add file 'sounds/OORCWNCE'
        Mod filesystems: 1039 ms
Initialized logging system based on settings successfully.
[log level] ai => not set
[log level] animation => not set
[log level] bonus => not set
[log level] global => trace
[log level] mod => not set
[log level] network => not set
[log level] rng => info
Found 3 render drivers
        opengl (active)
        opengles2
        software
Created renderer opengl
Starting network thread
        Initializing content handler
                Parsing mod: OK (core)
                Parsing mod: OK (hota)
                Parsing mod: OK (reworked-commanders)
                Parsing mod: OK (vcmi)
                Parsing mod: OK (vcmi-extras)
                Parsing mod: OK (wake-of-gods)
                Parsing mod: OK (hota.airship)
                Parsing mod: OK (hota.artifacts)
[...]

PR #6316, opened 2025-11-14
Add BattleEnded pack

mrhaandi:
Proposed fix for #6309 (up for discussion).

What currently happens when nullkiller ai defeats a garrisoned hero in a town:
1. BattleResultProcessor sends the signal that the battle has ended
https://github.com/vcmi/vcmi/blob/81db5c7cf4746dc61cc2d38f03f99fcd3916939d/server/battles/BattleResultProcessor.cpp#L561
2. the ai sets `NO_BATTLE`
https://github.com/vcmi/vcmi/blob/81db5c7cf4746dc61cc2d38f03f99fcd3916939d/AI/Nullkiller/AIGateway.cpp#L488
3. the ai understand that it can continue (there is no ongoing battle, movement, or other tasks)
https://github.com/vcmi/vcmi/blob/81db5c7cf4746dc61cc2d38f03f99fcd3916939d/AI/Nullkiller/AIGateway.cpp#L1770
4. the ai obtains a incoherent gamestate, because the losing hero is not yet removed from the town
https://github.com/vcmi/vcmi/blob/81db5c7cf4746dc61cc2d38f03f99fcd3916939d/AI/Nullkiller/Analyzers/ArmyManager.cpp#L482
5. at a later point, losing hero removal is initiated in BattleResultProcessor (which is too late!)
https://github.com/vcmi/vcmi/blob/81db5c7cf4746dc61cc2d38f03f99fcd3916939d/server/battles/BattleResultProcessor.cpp#L567

The resulting behavior due to 4. is arbitrary. Mostly it is fine if the ai does not interact with the foreign dead player in town. Sometimes there are invalid instructions and sometimes a segfault. This also happens with no mods.

The proposed fix delays the signal which truly ends the battle until all battle effects (including hero removal) are processed.

### DIFF
--- a/AI/Nullkiller/AIGateway.cpp
+++ b/AI/Nullkiller/AIGateway.cpp
@@ -485,6 +485,13 @@ void AIGateway::battleResultsApplied()
 	LOG_TRACE(logAi);
 	NET_EVENT_HANDLER;
 	assert(status.getBattle() == ENDING_BATTLE);
+}
+
+void AIGateway::battleEnded()
+{
+	LOG_TRACE(logAi);
+	NET_EVENT_HANDLER;
+	assert(status.getBattle() == ENDING_BATTLE);
 	status.setBattle(NO_BATTLE);
 }
 
--- a/AI/Nullkiller/AIGateway.h
+++ b/AI/Nullkiller/AIGateway.h
@@ -141,6 +141,7 @@ public:
 	void heroManaPointsChanged(const CGHeroInstance * hero) override;
 	void heroSecondarySkillChanged(const CGHeroInstance * hero, int which, int val) override;
 	void battleResultsApplied() override;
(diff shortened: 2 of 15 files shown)


IvanSavenko:
Can you confirm that everything works as expected if quick combat is enabled, AI player attacks human-controlled hero, and human player uses manual combat replay?

mrhaandi:
> Can you confirm that everything works as expected if quick combat is enabled, AI player attacks human-controlled hero, and human player uses manual combat replay?

Yes, this works as well.
Specifically, quick combat when it displays the choice for manual replay does not call
`void BattleResultProcessor::battleFinalize(const BattleID & battleID, const BattleResult & result)` where the new code is situated. Only clicking ok (or playing out the combat manually and then clicking ok) finalizes the battle result.

Issue #6314, opened 2025-11-13
Crash to Desktop when try to town traveling.


zcchen:
**Describe the bug**
Crash to Desktop when try to town traveling, including __Castle Gate__ (building in Infero Towm), __Town Portal__ (Earth Magic level 5).
Usually it happens when at least one possible town is unavailable, like a hero is visiting at Town A.

**Game logs**
[VCMI_Client_log.txt](https://github.com/user-attachments/files/23525218/VCMI_Client_log.txt)
[VCMI_Launcher_log.txt](https://github.com/user-attachments/files/23525219/VCMI_Launcher_log.txt)

**To Reproduce**

1. Enable Mods as my attachment config file, including "HoTA". (I am not sure whethere its related.)
    [modSettings.json](https://github.com/user-attachments/files/23525414/modSettings.json)
3. Let hero A & B visit the town A and town B, and let both towns have __Castle Gate__.
4. Let hero C master spell __Town Portal__ and learn Expert EarthMagic.
5. Open the __Castle Gate__ in town A or town B, Crash to Desktop.
6. Cast the spell __Town Portal__ from hero C, Crash to Desktop as well

**Expected behavior**
1. Never crash to desktop.
2. Draw the town list, if those towns are unavailable, skip them

**Actual behavior**
Crash to desktop.

**Did it work earlier?**
On version 1.6.8, this issue looks not happened.
On latest `master` branch build, this issue happens.

**Screenshots**

**Version**

- OS: Linux
- Version: 1.7.0 (built by command `dpkg-buildpackage -uc -us -b`)

**Additional context**
[settings.json](https://github.com/user-attachments/files/23525420/settings.json)

kambala-decapitator:
> On latest master branch build, this issue happens.

Are you sure you're using master and not develop? Master is basically the same as 1.6.8

kambala-decapitator:
Please post git revision you're using. Unfortunately, your logs don't have it. Recently a number of town portal bugs have been fixed.

zcchen:
> > On latest master branch build, this issue happens.
> 
> Are you sure you're using master and not develop? Master is basically the same as 1.6.8

Sorry. I made a mistake. It is from `develop` branch. 
git revision is `81db5c7cf4746dc61cc2d38f03f99fcd3916939d`.

mrhaandi:
I tried to reproduce the above steps on `develop` with hota installed and I get no crash. For example on town portal it says "No available town.  Spell Failed!".
[...]

PR #6317, opened 2025-11-14
Fix a segment fault issue due to array index out of bounds in CObjectListWindow::CItem::CItem.

zcchen:
Proposed fix for https://github.com/vcmi/vcmi/issues/6314 (up for discussion).

`client/windows/GUIClasses.cpp` in `develop` branch (git revision: `81db5c7cf4746dc61cc2d38f03f99fcd3916939d`) 
```
	// Line 1537 - 1542
	OBJECT_CONSTRUCTION;
	auto imgIndex = parent->itemsVisible[index].first;
	if(parent->images.size() > index && parent->images[imgIndex])
		icon = std::make_shared<CPicture>(parent->images[imgIndex], Point(1, 1));
	border = std::make_shared<CPicture>(ImagePath::builtin("TPGATES"));
	pos = border->pos;
	// ....
```
but in `master` branch (git tag: 1.6.8)
``` 
	// Line 1474 - 1478
	OBJECT_CONSTRUCTION;
	if(parent->images.size() > index)
		icon = std::make_shared<CPicture>(parent->images[index], Point(1, 1));
	border = std::make_shared<CPicture>(ImagePath::builtin("TPGATES"));
	pos = border->pos;
	// ....
```

In the code from `develop` branch, the `imgIndex` looks not line the index for array. It may be bigger than the range of the array.
If you use `gdb` to debug my save file, the `imgIndex` returns `4399`, but the `parent->images.size()` is `6`. So `parent->images[imgIndex]` may cause a Segmentation fault.

### DIFF
--- a/client/windows/GUIClasses.cpp
+++ b/client/windows/GUIClasses.cpp
@@ -680,7 +680,7 @@ void CTavernWindow::HeroPortrait::clickPressed(const Point & cursorPosition)
 void CTavernWindow::HeroPortrait::clickDouble(const Point & cursorPosition)
 {
 	clickPressed(cursorPosition);
-	
+
 	if(onChoose)
 		onChoose();
 }
@@ -924,7 +924,7 @@ CTransformerWindow::CTransformerWindow(const IMarket * _market, const CGHeroInst
 		army = hero;
 	else
 		army = dynamic_cast<const CArmedInstance *>(market); //for town only
-	
+
 	if(army)
 	{
 		for(int i = 0; i < GameConstants::ARMY_SIZE; i++)
@@ -1535,9 +1535,8 @@ CObjectListWindow::CItem::CItem(CObjectListWindow * _parent, size_t _id, std::st
 	index(_id)
 {
 	OBJECT_CONSTRUCTION;
-	auto imgIndex = parent->itemsVisible[index].first;
-	if(parent->images.size() > index && parent->images[imgIndex])
-		icon = std::make_shared<CPicture>(parent->images[imgIndex], Point(1, 1));
+	if(parent->images.size() > index && parent->images[index])
+		icon = std::make_shared<CPicture>(parent->images[index], Point(1, 1));
 	border = std::make_shared<CPicture>(ImagePath::builtin("TPGATES"));
 	pos = border->pos;
 
@@ -1723,7 +1722,7 @@ void CObjectListWindow::itemsSearchCallback(const std::string & text)
 		std::string name = parts.back();
 		boost::erase_all(name, "{");
     	boost::erase_all(name, "}");
-		
+
 		if(auto score = TextOperations::textSearchSimilarityScore(text, name)) // Keep only relevant items
 			rankedItems.emplace_back(score.value(), item);
 	}


Laserlicht:
I assume this do not work correctly because you removed the itemsVisible.

Try it with many different towns (list needs to be scrollable, or use filter functionality).

I assume this PR showing incorrect town image when scrolling down (or filtering).

mrhaandi:
While the particular fix could be questioned, @zcchen is correct that (also without any mods installed) `parent->itemsVisible[index].first` is the id (which can be large) and not an index (leading to a segfault).

mrhaandi:
The proposed solution is insufficient wrt. images, as @Laserlicht says. Below clearly not a Rampart town.

filter

Possibly something like find the global index which corresponds to the local index of the visible item should be implemented?

zcchen:
I am new to vcmi project. If not the issue I found, I won't dive into the code to checkout why segfault.

Base on my understanding, `develop` branch adding code line `auto imgIndex = parent->itemsVisible[index].first;` is trying to ensure the town image is correctly. And this incorrect town image is an issue on current `master` branch (version `1.6.8`). 
Am I correct? @Laserlicht @mrhaandi

zcchen:
> The proposed solution is insufficient wrt. images, as @Laserlicht says. Below clearly not a Rampart town.
> 
> filter
> 
> Possibly something like find the global index which corresponds to the local index of the visible item should be implemented?

Could you show me how to reproduce this incorrect image steps?
I can assume, if I lost the first town and recapture another town (different from my first town), that may see this incorrent image.
Is there any cheat code to achieve this incorrect image issue?

mrhaandi:
While the following is somewhat crude, it solves the issue without introducing a regression:
```c++
CObjectListWindow::CItem::CItem(CObjectListWindow * _parent, size_t _id, std::string _text)
	: CIntObject(LCLICK | DOUBLECLICK | RCLICK_POPUP),
	parent(_parent),
	index(_id)
{
	OBJECT_CONSTRUCTION;
	for (auto imgIndex = 0; imgIndex < parent->items.size(); imgIndex++)
	{
		//find the image index corresponding to the item itemsVisible[index]
		if (parent->itemsVisible[index] == parent->items[imgIndex])
			if(parent->images.size() > imgIndex && parent->images[imgIndex])
			{
				icon = std::make_shared<CPicture>(parent->images[imgIndex], Point(1, 1));
				break;
			}
	}
	border = std::make_shared<CPicture>(ImagePath::builtin("TPGATES"));
	pos = border->pos;

	setRedrawParent(true);

	text = std::make_shared<CLabel>(pos.w/2, pos.h/2, FONT_SMALL, ETextAlignment::CENTER, Colors::WHITE, _text, 256);
	select(index == parent->selected);
}
```
correct

client/windows/GUIClasses.cpp
```diff
@@ -1535,9 +1535,8 @@ CObjectListWindow::CItem::CItem(CObjectListWindow * _parent, size_t _id, std::st
 	index(_id)
 {
 	OBJECT_CONSTRUCTION;
-	auto imgIndex = parent->itemsVisible[index].first;
-	if(parent->images.size() > index && parent->images[imgIndex])
-		icon = std::make_shared<CPicture>(parent->images[imgIndex], Point(1, 1));
+	if(parent->images.size() > index && parent->images[index])
+		icon = std::make_shared<CPicture>(parent->images[index], Point(1, 1));
```
Laserlicht, line 1539: You could try this:
```suggestion
	auto it = std::find(parent->items.begin(), parent->items.end(), parent->itemsVisible[index]);
	int imgIndex = (it != parent->items.end()) ? std::distance(parent->items.begin(), it) : -1;
	if(imgIndex >= 0 && imgIndex < parent->images.size() && parent->images[imgIndex])
		icon = std::make_shared<CPicture>(parent->images[imgIndex], Point(1,1));
```

Laserlicht:
Ah. @mrhaandi was a little bit faster with another solution. :)

> I am new to vcmi project.

No problem. Thank you for supporting. :)

client/windows/GUIClasses.cpp
mrhaandi: This is the c++ version of my crude c code :)

mrhaandi:
I tested both proposals and both work. By the way, a hacky way to enable the search feature for testing is changing the lines `init(titleWidget_, _title, _descr, searchBoxEnabled, blue);` to `init(titleWidget_, _title, _descr, true, blue);`.

zcchen:
Good. This segfault issue looks soluted.
@Laserlicht @mrhaandi Your solutions are both better than mine.
If only one solution must be select, I would like to pick those from @Laserlicht. Because it is a C++ project, C++ solution is more elegent for this project.

As I said I am new to vcmi project, I don't know how to continue (aka, skip my commit, use the solution from @Laserlicht or @mrhaandi  ).
Could you just go ahead and merge your fix to `develop` branch?

Thanks very much for your supporting.

Laserlicht:
> Could you just go ahead and merge your fix to develop branch?

Will merged soon (probably in the next few days).

See here: #6319

PR #6322, opened 2025-11-16
teleport cheat

Laserlicht:
cheat to teleport player to map coordinates


misiokles:
btw - is there a way to see map coordines during play?

Laserlicht:
> btw - is there a way to see map coordines during play?

Yes in status bar. Pressing Ctrl while hovering tiles.

misiokles:
Every day I learn sth new :)

PR #6325, opened 2025-11-16
image from def also for ImageLocator

### LINKED ISSUE #6324: In a mod icons from def's cannot be loaded anymore
**Describe the bug**
In a mod loading icons for creatures like: 
"graphics": {
			"iconSmall": "cprsmall.def:98",
			"iconLarge": "TWCRPORT.def:98",

does not work anymore.

**Game logs**

**To Reproduce**

1. In HotA change iconSmall for Nymph to cprsmall.def:10
2. No icon will be loaded in game

**Did it work earlier?**
Was working 1 year ago

**Version**

- OS: [e.g. Windows, 
- Version: [latest develop]

Laserlicht:
Fixes #6324

Extends already existing functionality also for overload of `loadImage` with `ImageLocator`.

Files can used with `deffile.def:0:5` (`0` is group, `5` is frame in example)

### DIFF
--- a/client/renderSDL/RenderHandler.cpp
+++ b/client/renderSDL/RenderHandler.cpp
@@ -349,6 +349,20 @@ std::shared_ptr<IImage> RenderHandler::loadImage(const ImageLocator & locator)
 {
 	ImageLocator adjustedLocator = locator;
 
+	if(locator.image)
+	{
+		std::vector<std::string> splitted;
+		boost::split(splitted, (*locator.image).getOriginalName(), boost::is_any_of(":"));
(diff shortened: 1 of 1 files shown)


krs0:
Needed fix! Thank you.

Usually there is only 1 group (0), could that be implicit if not specified?

Would make it easier/cleaner to enter the image in mods

Laserlicht:
> Usually there is only 1 group (0), could that be implicit if not specified?

Would prefer one syntax to avoid confusing. And a image from an animation is always defined with group and frame.

Issue #5554, opened 2025-03-14
RMG cannot handle object customization properly without additional steps


wnukos:
**Describe the bug**
An object customized in the "commonObjects" field does not follow the set values of "value", "rarity" and "zoneLimit" properties unless the object is banned in the "bannedObjects" or "bannedCategories" field at the same time. Only objects that are not available by default (for example Hut of the Magi) can be customized reliably without banning them.

**To Reproduce**

1. Enable mods VCMI Extras and "Templates with custom objects", available for download at: https://drive.google.com/file/d/1lstSv8-S8ta7YvUzzJTNj2gI87BNT82h/view?usp=sharing
2. Launch VCMI and choose "Test template". It is a Jebus Cross modification with Griffin Conservatory customized in the "commonObjects" field - rarity = 9999, value = 10000, zoneLimit=5. 
3. Choose "Test template 2" which is similar to "Test template" but Griffin Conservatory is modified in "commonObjects" and is listed in "bannedObjects" field at the same time.
4. Analyze the guards defending Griffin Conservatory on both templates.

**Expected behavior**
On both templates Griifin Conservatory should have 10 000 value and generate in amount 5 per zone.

**Actual behavior**
Griffin Conservatory is customized properly only if it is banned at the same time. If the object is only modified by "commonObjects" field, then both regular Conservatory (2000 value, rarity 100, no zone limit) and modified Conservatory (10000 vaue, rarity 9999, zone limit 5) are generated.

**Did it work earlier?**
From my tests problem persisted since object customization had been introduced, however it was hard to identify the root cause for strange behavior so I am not sure if it was the same issue as described here.

**Version**

- OS: Windows
- Version: 1.6.7

**Additional context**
Most probably issue #5265 describes the same bug.

IvanSavenko:
Will take a look for 1.7, probably querying for identifiers is done incorrectly.

PR #6331, opened 2025-11-19
Fix rmg not respecting dwelling info

mrhaandi:
Fixes https://github.com/vcmi-mods/horn-of-the-abyss/issues/244

What currently happens:
1) If a mod sets `rmg` info on a dwelling, then
https://github.com/vcmi/vcmi/blob/2db925347e82297d8d98bece9dbb254227002961/lib/rmg/modificators/TreasurePlacer.cpp#L109
may put it on the map, disregarding specific dwelling code (e.g. not setting owner, with leads to assertion errors and random owners).
2) Random dwelling placement
https://github.com/vcmi/vcmi/blob/2db925347e82297d8d98bece9dbb254227002961/lib/rmg/modificators/TreasurePlacer.cpp#L210
disregards `rmg` info and spawns too many rare dwellings.

Also fixed:
- a dwelling having mapLimit set will now be excluded from ObjectDistributor::distributeLimitedObjects(), which would otherwise produce unwanted behavior
- if in TreasurePlacer::ObjectPool::patchWithZoneConfig a dwelling is initialized via setBasicProperties it does set a neutral owner (and not produce a bug)

Edit: more context:
`lib/rmg/modificators/TreasurePlacer.cpp` has different routines to place random objects on the map:
- addCommonObjects();
- addDwellings();
- addPandoraBoxes();
- addSeerHuts();
- addPrisons();
- addScrolls();

Common objects in particular are supposed to be unowned and are just plaed based on rmg properties.
In current placement code dwellings are not supposed to have rmg properties. If they do, they are mistaken for common objects and placed by the wrong routine, which messes up e.g. ownership.

The PR is supposed to explicitly exclude objects for which dedicated routines exist:
```
std::set<MapObjectID> excludedIDs = {
		Obj::PRISON,
		Obj::CREATURE_GENERATOR1,
		Obj::CREATURE_GENERATOR4,
		Obj::SPELL_SCROLL,
		Obj::PANDORAS_BOX,
		Obj::SEER_HUT
	};
```

### DIFF
--- a/lib/rmg/modificators/TreasurePlacer.cpp
+++ b/lib/rmg/modificators/TreasurePlacer.cpp
@@ -150,7 +150,11 @@ void TreasurePlacer::setBasicProperties(ObjectInfo & oi, CompoundMapObjectID obj
 {
 	oi.generateObject = [this, objid]() -> std::shared_ptr<CGObjectInstance>
 	{
-		return LIBRARY->objtypeh->getHandlerFor(objid)->create(map.mapInstance->cb, nullptr);
+		auto obj = LIBRARY->objtypeh->getHandlerFor(objid)->create(map.mapInstance->cb, nullptr);
+		// adjust ownership for ownable objects (such as dwellings)
+		if (obj->asOwnable() && obj->tempOwner == PlayerColor::UNFLAGGABLE)
+			obj->setOwner(PlayerColor::NEUTRAL);		
+		return obj;
 	};
 	oi.setTemplates(objid.primaryID, objid.secondaryID, zone.getTerrainType());
 }
@@ -271,12 +275,6 @@ void TreasurePlacer::addDwellings()
 					oi.value = static_cast<ui32>(cre->getAIValue() * cre->getGrowth() * valueModifier);
 					oi.probability = 40;
 				}
-				oi.generateObject = [this, secondaryID, dwellingType]() -> std::shared_ptr<CGObjectInstance>
-				{
-					auto obj = LIBRARY->objtypeh->getHandlerFor(dwellingType, secondaryID)->create(map.mapInstance->cb, nullptr);
-					obj->tempOwner = PlayerColor::NEUTRAL;
-					return obj;
-				};
 				if(!oi.templates.empty())
 					addObjectToRandomPool(oi);
 			}


DjWarmonger:
Not sure about this change. It will allow to enable specific dwellings in RMG, but not to change their values or rarity. Need to think about it, I'll get back to RMG next week.

mrhaandi:
> It will allow to enable specific dwellings in RMG, but not to change their values or rarity.

Could you elaborate, why the proposed change (see also below) is not sufficient to take user-defined rarity, value, and  zone limit into account?
```c++
if(rmgInfo.value)
{
	if (rmgInfo.value > zone.getMaxTreasureValue())
		continue;
	oi.value = rmgInfo.value;
	oi.probability = rmgInfo.rarity;
	if (rmgInfo.zoneLimit != std::numeric_limits<ui32>::max())
		oi.maxPerZone = rmgInfo.zoneLimit;
}
```

mrhaandi:
Related [#5554](https://github.com/vcmi/vcmi/issues/5554)

mrhaandi:
@DjWarmonger would you consider taking a closer look?

For the HotA expansion, the underlying issue induces assertion errors and makes debugging cumbersome.

DjWarmonger:
Another issue: The current formula for dwelling value is supposed to scale their value by the number of matching towns on the map (as dwelling is more valuable, and more common if there are multiple matching towns). Setting fixed value for dwelling removes this feature.

HoTA has some extra modificators, like Vampire Dwelling having value from the above formular additionally multiplied by 3 and Peasant dwelling having similar change IIRC. Recreating these multipliers would match existing use cases better than single fixed value.

mrhaandi:
> Another issue: The current formula for dwelling value is supposed to scale their value by the number of matching towns on the map.

Now the scaling is also applied to the rmg value setting by the same factor.
Vampire dwellings and Peasant dwellings now could set a base value (which can be 3 times the normal value) and will also be subject to the multiplicity formula.

I also fixed two issues:
- a dwelling having `mapLimit` set will now be excluded from `ObjectDistributor::distributeLimitedObjects()`, which would otherwise produce unwanted behavior
- if in `TreasurePlacer::ObjectPool::patchWithZoneConfig` a dwelling is initialized via `setBasicProperties` it does set a neutral owner (and not produce a bug)

DjWarmonger:
Was this supposed to fix this issue?
https://github.com/vcmi-mods/horn-of-the-abyss/issues/244
It doesn't seem so, but In guess HOTA mod itself needs to be updated as it is the real culprit.

Otherwise - looks fine to me.

mrhaandi:
> Was this supposed to fix this issue?
> [vcmi-mods/horn-of-the-abyss#244](https://github.com/vcmi-mods/horn-of-the-abyss/issues/244)

Yes.

> It doesn't seem so, but In guess HOTA mod itself needs to be updated as it is the real culprit.

I'm not sure which part you say is not fixed. When applying this PR I get no errors for unassigned owner (yellow flags) and also the actual rmg values are taken into account on generation (they do not spawn all over the place).

PR #6333, opened 2025-11-19
Fixed golems' level

Olegmods:
Gold and Diamond golems' level changed from 4 and 5 to 5 and 6. https://heroes.thelazy.net/index.php/Diamond_Golem and some other source I found state that the ones from that PR are correct.


Olegmods:
@IvanSavenko can that PR be merged? Everything seems to be correct, plus it's a small config fix after all.

Issue #5803, opened 2025-06-17
App crashes on exit


Olegmods:
**Describe the bug**
As stated in title.

**Game logs**
Not necessary.

**To Reproduce**

1. Open VCMI
2. Click "exit".
3. See error

**Expected behavior**
No crash, just return to OS/launcher

**Actual behavior**
Crash

**Did it work earlier?**
Worked in 1.6.

**Screenshots**
Not necessary.

**Version**

- OS: Android
- Version: develop

**Additional context**
Probably OS-related.

Olegmods:
Still is a thing btw.

kambala-decapitator:
please upload client log and crash log, the latter can be obtained with logcat

Olegmods:
> please upload client log and crash log, the latter can be obtained with logcat

Will upload the client log later.
Probably won't upload the crash one tho - I'm not familiar with logcat.

kambala-decapitator:
> I'm not familiar with logcat

get familiar then :) IIRC there're android apps to capture logcat that you don't have to install adb on your computer

Olegmods:
@kambala-decapitator is that the crashlogs you need?

      #00 pc 0000000000070114  /apex/com.android.runtime/lib64/bionic/libc.so (pthread_mutex_lock+16) (BuildId: 2f5cbfb3b95d8901d213f7d663a48778)
      #01 pc 00000000000c5f08  /data/app/~~8PdaBvxVpabLAAHl_HP9cw==/is.xyz.vcmi.daily-PDuxIjZdsT-IaDRrumH0Yg==/lib/arm64/libc++_shared.so (std::__ndk1::mutex::lock()+12) (BuildId: df3f83af4e54f316c793c3655c4753a8851a234c)
      #02 pc 00000000010b11f4  /data/app/~~8PdaBvxVpabLAAHl_HP9cw==/is.xyz.vcmi.daily-PDuxIjZdsT-IaDRrumH0Yg==/lib/arm64/libvcmiclient_arm64-v8a.so (Java_eu_vcmi_vcmi_NativeMethods_tryToSaveTheGame+52) (BuildId: 768ff3c44ee53bb5802266c6b224bfc546a6c465)
      #03 pc 0000000000d5d128  /data/misc/apexdata/com.android.art/dalvik-cache/arm64/boot.oat (art_jni_trampoline+104)
[...]

PR #6334, opened 2025-11-20
Fix undefined unlocks on unlocked mutex

mrhaandi:
`makeUnlockGuard` sometimes tried to unlock a mutex which is not locked, leading to unspecified behavior (can be different on different platforms).
Let's control the situation better and also protocol it for debugging.

Edit: currently this is for debugging (in my build there is also undefined behavior). A better solution is needed.

Possibly related to https://github.com/vcmi/vcmi/issues/5803

### DIFF
--- a/lib/UnlockGuard.h
+++ b/lib/UnlockGuard.h
@@ -37,7 +37,7 @@ namespace vstd
 		protected:
 			void unlock(Mutex &m)
 			{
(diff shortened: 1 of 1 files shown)


mrhaandi:
Superseded by #6337 (which fixes one discovered issue).

PR #6348, opened 2025-11-24
QuickSave / QuickLoad

### LINKED ISSUE #6346: Quicksave / Quickload
**Describe your proposal**
Have an easily accessible from the adventure map quicksave button (maybe on the same panel as next turn) and an easily accessible quickload button (both from adventure map and combat).
- Quicksave could be akin to autosave with X last quicksaves.
- Quickload is supposed to try and restart the server with the exact current options (e.g. multiplayer -> hotseat -> two players) and then load the last quicksave. 

**Additional context**
When playing with young or novice players in hotseat, there tends to be a lot of manual save / manual load at many interactions (not only fights). Loading requires going back to the main menu and restarting the hotseat session, this unnecessarily interrupts the flow of the game.

mrhaandi:
Fixes #6346.

Developing two additional buttons for QuickSave / QuickLoad for singleplayer and hotseat scenarios.
The intention is to not restart the server, but to reset the gamestate / gamehandler to a prior state.

intended_buttons

**DONE**
- add two buttons to the adventure panel
- call quicksave, quickload functionality on button interaction
- add `LobbyQuickLoadGame` pack for lobby for direct server access
- add loading screen
- forbid quick save / quick load on multiplayer (hotseat is allowed)
- proper button images
- add question dialog for quickload
- add notification for saving
- add localization for the question whether to quickload
- forbid save if there are queued actions
- add keyboard hotkeys `F8`, `F9` to load from adventure map and battle
- externalize buttons to vcmi-extras
- testing
   - single player (with ai)
   - hotseat (with and without ai)
   - campaign

**DESIGN DECISIONS**
- unclear in how far the quicksave needs to be verified wrt. current map / player settings
- do not add user interface for quickload from battle (it is easy to redo a battle)
- disallow qickload in battle while animations are ongoing, or while the opponent is making a turn (otherwise deadlock with `ENGINE->interfaceMutex`)

**TODO**
- non-english localization (?)


config/widgets/adventureMap.json
```diff
@@ -237,14 +237,32 @@
 					"playerColored" : true,
 					"area": { "top" : 96, "left": 32, "width" : 32, "height" : 32 }
 				},
+				{
```
Laserlicht: `adventureMap.json` shouldn't modified here. In VCMI repo it reflects OH3 behavior.

Additional buttons can be placed in [VCMI-extras](https://github.com/vcmi-mods/vcmi-extras).
- https://github.com/vcmi-mods/vcmi-extras/blob/vcmi-1.7/Mods/adventureMap/Content/config/widgets/adventureMap.json (attention: every button is placed multiple times here, because of layout breakpoints - not nice, that's also a thing needs improvement: dynamic layout)
- and maybe: https://github.com/vcmi-mods/vcmi-extras/blob/vcmi-1.7/Mods/adventureMapVertical/Content/config/widgets/adventureMap.json

But not sure about adding in small layout breakpoints: We've another good buttons that aren't placed yet.

mrhaandi: I understand. I'll first try to integrate the functionality, and then externalize the buttons to vcmi-extras (so the buttons won't be part of this PR in the end).

Laserlicht: Yes. No problem - just remove before finishing feature. :)

mrhaandi:
@IvanSavenko I followed your advice and based the code on the `LobbyRestartGame` functionality. After testing individual configurations it appears to work well. Could you provide a preliminary review of this PR?

I understand that before completion the `adventureMap.json` needs to be restored and the newly designed buttons can be added to `vcmi-extras`.
Currently I use the vanilla `adventureMap.json` for ad-hoc testing.

mrhaandi:
I do want to add a quickload button to the battle screen (a common case for a quickload).
I'm not sure where to.
- There is usually space at 1. 
- 2 is already cluttered and I don't want to limit battle log further.
- There is space at 3, but it is currently reserved for (up to 12) creature actions.
- 4 (setting in battle). Requires one additional click. Don't know about exact spacing.
- 5 in battle replace the "Save" button by "Quick Load" button. Saving in battle is deactivated anyway. This would require some thought how to deal with `settingsMainContainer.json` ina  dynamic way. However, it already tests `parentBattleUi`, so it should be easy.
where

Laserlicht:
Do we really need a physical button for this?

Because:
- if you want to replay battle: just surrender and use replay button 
- if you want the state before battle I think shortcut for quickload and regular loading (quicksave) is enough

Laserlicht:
Btw. perhaps you also interested in: https://github.com/vcmi/vcmi/issues/4816

Requested multiple times. But attention: It's may tricky to solve. Because the save should created before action (e.g. before starting movement to creatures for battle). And there may many edge cases (e.g. events, buildings with garrision...).
It's not enough to save immediately before battle.

mrhaandi:
> * if you want to replay battle: just surrender and use replay button
> * if you want the state before battle I think shortcut for quickload and regular loading (quicksave) is enough

I agree on the first point. The second point relies on a keyboard (which sometimes, and also in my use-case is not present).
Unfortunately regular loading requires quite some menu interaction. Especially playing with children turns into a lot of clicking.

I also know of https://github.com/vcmi/vcmi/issues/4816, which is mainly about saving (here loading quickly is a concern as well).

Edit: Maybe for this particular PR I will keep it simple and add the basic functionality without additional quick load menu buttons. In edge cases key presses can be simulated externally.

Laserlicht:
Another idea would be a small check before regular loading. Check if quick save is done in last n minutes, then ask if it should loaded or open regular interface.

config/widgets/adventureMap.json
mrhaandi: Now the interface modifications are part of https://github.com/vcmi-mods/vcmi-extras/pull/80 (for mid and high resolution).

mrhaandi:
As discussed, now the hotkeys (default `F8` and `F9`) are set for adventure map and battle mode.

I do consider this PR as feature complete (no need to overengineer simple mechanics).

client/CServerHandler.cpp
```diff
@@ -691,6 +691,20 @@ void CServerHandler::endGameplay()
 	}
 }
 
+void CServerHandler::quickLoadGame(std::string path)
```
IvanSavenko:
```suggestion
void CServerHandler::quickLoadGame(const std::string & path)
```

client/CPlayerInterface.cpp
```diff
@@ -1800,6 +1802,30 @@ void CPlayerInterface::proposeLoadingGame()
...
+	// notify player about saving
+	GAME->server().getGameChat().sendMessageGameplay("Saving game to " + path);
+	GAME->interface()->cb->save(path);
+}
+
+void CPlayerInterface::proposeQuickLoadingGame()
+{
+	std::string path = "Saves/Quicksave";
+
+	if(!CResourceHandler::get("local")->existsResource(ResourcePath(path, EResType::SAVEGAME)))
```
IvanSavenko, line 1817: What would happen if quick save exists, but not from this game / map? Or even - if player is playing as different color, possibly one that does not exists in save, e.g.:
- start map with 2 players - red & blue, make quick save
- start new map, playing as pink, try to quick load

Expected behavior is either to block this completely, or ensure that player will correctly switch to playing as red

server/NetPacksServer.cpp
```diff
@@ -29,6 +29,12 @@
 
 void ApplyGhNetPackVisitor::visitSaveGame(SaveGame & pack)
 {
+	//if (!gh.queries->allQueries().empty())
+	//{
+	//	logGlobal->error("Cannot save game while queries are pending");
+	//	result = false;
+	//	return;
+	//}
```
IvanSavenko: Unused code?

client/NetPacksLobbyClient.cpp
```diff
@@ -41,6 +41,17 @@
...
 #include "../lib/texts/CGeneralTextHandler.h"
 
+void ApplyOnLobbyHandlerNetPackVisitor::visitLobbyQuickLoadGame(LobbyQuickLoadGame & pack)
+{
+	assert(handler.getState() == EClientState::GAMEPLAY);
+	if(auto w = ENGINE->windows().topWindow<CLoadingScreen>())
+	{
+		w->finish();
+		w->tick(0);
+		w->redraw();
```
IvanSavenko: Why exactly is this needed?

server/NetPacksServer.cpp
mrhaandi: Ah, yes, will be removed. This was a test, but turned out too restrictive and left as food for thought.

client/NetPacksLobbyClient.cpp
mrhaandi: Indeed, not needed anymore and I'll remove it. Previously the final tick of the loading screen was not sent and for quickload it looked as if no loading happens. Now it happens in `bool CVCMIServer::prepareToStartGame()` at `//send final progress`. So this code is indeed not necessary.

client/CPlayerInterface.cpp
mrhaandi: Now here some design decisions need to happen. Current design: try force load game with current single player / hotseat settings.
- If player setup is incompatible: undefined behavior
- If player setup compatible, then use it no matter what it was on save. So one can in fact change color or chance ai to human and vice versa.
- Currently the map is loaded from save and the current map is not taken into account.

I understand that one would not want undefined behavior, so I'll look into this.
Do you think that we want to enforce same map / player settings for successful loading?

mrhaandi: I added a method `CServerHandler::canQuickLoadGame` which checks compatibility with the initial server settings:
- check same exact initial start time (catches basically all cases)
- check same map name (for non rmg ones)
- check same difficulty
- check number of players, same player colors, and same ai / human settings

If any of the above does not match, an error is written to logs and no quick load happens.

mrhaandi:
> Looks OK. Please resolve merge conflict so that PR can be merged

Rebased, resolved merge conflicts, and squashed.

PR #6349, opened 2025-11-24
Fix issues with object customization in rmg

### LINKED ISSUE #5554: RMG cannot handle object customization properly without additional steps

### LINKED ISSUE #5556: Core templates cannot contain customized objects from mods
**Describe the bug**
VCMI allows to modify core templates available in the game, for example Jebus Cross or Nostalgia and introduce customized objects there. However, if the objects come from a mod, the game crashes upon launching. As a consequence a new template needs to be created instead of modifying the existing one (see for example "Jebus Cross Plus" template in PvP Balance mod) which is not always desired.

**To Reproduce**

1. Install mods VCMI Extras and "Modified Core Templates" (available for download at: https://drive.google.com/file/d/1HiknSu0H7qEpacJf1NZLZVKwgr_XeOmC/view?usp=sharing). The mod modifies the core template Nostalgia introducing some customized core objects.
2. Launch the game and choose Nostalgia template.
3. Install mods VCMI Extras, HotA and "Modified Core Templates with HotA Objects" (available for download at: https://drive.google.com/file/d/1ngImLSZ0atfqdqgs_mWqib9K5KhIFcWX/view?usp=sharing). It introduces customized objects from Hota mod in Nostalgia template.
4. Launch VCMI - the game crashes.
5. Modify the template introduced in the "Modified Core Templates with HotA Objects mod" giving it a new name, without the "vcmi:" reference. The game works properly.

**Expected behavior**
It should be possible to modify core templates including customized objects coming from mods, given that proper mod dependencies are set in the mod that modifies the core template. The game should work properly in both descirbed cases.

**Actual behavior**
Only core objects can be customized in core templates. In order to introduce customized objects from mods a new template needs to be created.

**Version**

- OS: Windows
- Version: 1.6.7

**Additional context**
The described issue is problematic for mods that try to modify core templates, for example PvP Balance mod.

### LINKED ISSUE #6354: Allow "Compatibility" type mods to add random map templates
**Describe your proposal**
Currently it is not possible to add new random map templates with "Compatibility" type mods. Such feature would be useful because it allows to adjust the templates to the pool of enabled mods. Probably this could be extended to other things that cannot be added by Compatibility mods that I currently can't think of.

IvanSavenko:
Fixed several issues with resolving identifiers of map objects when banning or overriding rmg properties of map objects.

Syntax should now be way simple and more readable, e.g:
```json
"bannedObjects" : {
	"randomArtifactRelic" : true,
	"hota.mapobjects:townGate" : true,
	"cartographer" : {
		"cartographerLand" : true
	}
}
```

Objects with customized rmg properties are now implicitly banned - to prevent default rmg properties from having an effect.

In addition, it is now possible to ban or customize objects even without explicit mod dependency, similar to how object bans work in 1.7. If corresponding mod is not active, object ban/customization will be ignored.

See changes in docs in this PR for docs.

Resolved issues:

- Fixes #5554
- Fixes #5556
- Fixes #6354

NOTE: old (1.6) rmg templates will work as before, however changed format will be reported by validation - please update mods either before or after 1.7 release

### DIFF
--- a/config/schemas/template.json
+++ b/config/schemas/template.json
@@ -86,9 +86,20 @@
 							}
 						},
 						"bannedObjects": {
-							"type": "array",
-							"items": {
-								"type": "string"
+							"type": "object",
+							"additionalProperties": {
+								"anyOf" : [
+									{
+										"type": "boolean"
+									},
+									{
+										"type": "object",
+										"additionalProperties": {
+											"type": "boolean"
+										}
+									}
+								]
+								
 							}
 						},
 						"commonObjects": {
@@ -96,7 +107,10 @@
 							"items": {
 								"type": "object",
 								"properties": {
-									"id": {
+									"type": {
+										"type": "string"
+									},
+									"subtype": {
 										"type": "string"
 									},
 									"rmg": {
@@ -115,7 +129,7 @@
 										"required": ["value", "rarity"]
 									}
 								},
-								"required": ["id", "rmg"]
+								"required": ["type", "rmg"]
 							}
 						}
 					}
--- a/docs/modders/Random_Map_Template.md
+++ b/docs/modders/Random_Map_Template.md
@@ -210,16 +210,51 @@
 		// All of objects of this kind will be removed from zone
 		// Possible values: "all", "none", "creatureBank", "bonus", "dwelling", "resource", "resourceGenerator", "spellScroll", "randomArtifact", "pandorasBox", "questArtifact", "seerHut", "other
 		"bannedCategories" : ["all", "dwelling", "creatureBank", "other"],
+
 		// Specify object types and subtypes
-		"bannedObjects" :["core:object.randomArtifactRelic"],
-		// Configure individual common objects - overrides banned objects
+		"bannedObjects" : {
+			// ban a specific object from base game or from any dependent mod. Object with such ID must exist
+			"randomArtifactRelic" : true,
+			
+			// ban object named townGate from mod 'hota.mapobjects'. Mod can be used without explicit dependency
+			// If mod with such name is not loaded, this entry will be ignored and will have no effect
+			"hota.mapobjects:townGate" : true,
+			
+			// ban only land version of Cartographer. Other versions, such as water and subterra cartographers may still appear on map
+			"cartographer" : {
+				"cartographerLand" : true
+			}
+		},
+
+		// Configure individual common objects. 
+		// Any object in this list will be excluded from regular placement rules, similarly to bannedObjects
 		"commonObjects":
 		[
 			{
-				"id" : "core:object.creatureBank.dragonFlyHive",
+				// configure water cartographer properties
+				"type" : "cartographer",
+				"subtype" : "cartographerWater",
 				"rmg" : {
 					"value"		: 9000,
 					"rarity"	: 500,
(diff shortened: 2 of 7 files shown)


mrhaandi:
What is the reasoning to not let rmg use the values correctly? This is the approach in #6331 to fix this issue.

IvanSavenko:
But this is completely different issue? My PR has no effect on dwellings.

mrhaandi:
> But this is completely different issue? My PR has no effect on dwellings.

Some problems stem from the fact that the TreasurePlacer tries to place common objects which have rmg properties.
https://github.com/vcmi/vcmi/blob/045251bdc2bfda59dc6351b4c4ab35b448f16296/lib/rmg/modificators/TreasurePlacer.cpp#L109
If a dwelling is modified to have rmg propeties (by default it does not), then the above method is insufficient to place the dwelling with neutral owner properly. This problem is not limited to dwellings, but those are most apparent.
I did assume that is what you mean by `Objects with customized rmg properties are now implicitly banned `.

IvanSavenko:
Same would happen in current build if you were to add dwelling to banned objects, so my PR won't change much here

mrhaandi:
> Same would happen in current build if you were to add dwelling to banned objects, so my PR won't change much here

I see, thanks for clarification.

wnukos:
Do I understand correctly that only rmg properties of objects can be customized in the template? So if I wanted to change for example guard numbers in a creature bank, it would have to be done by a mod?

IvanSavenko:
Correct. Only rmg parameters can be modified this way, just like in 1.6

This pr only fixes bugs & makes the feature more convenient to use

PR #6350, opened 2025-11-24
Prune roads that don't end it towns

DjWarmonger:


mrhaandi:
Could roads still go to/between navigational points of interest such as shipyards, portals, border guards, or border gates?

DjWarmonger:
Connected objects did not change. Portals and Shipyard get roads. Border Gates do not spawn on random maps at all.
(There was a feature to spawn border guard in SoD, but no one ever used it or request it as far as I know).

lib/rmg/CRoadRandomizer.cpp
```diff
@@ -197,6 +197,43 @@ void CRoadRandomizer::dropRandomRoads(vstd::RNG * rand)
 		}
 	}
 
+	// Trim roads ending in zones without towns (iteratively until no loose ends remain)
+	bool changed = true;
+	while(changed)
+	{
+		changed = false;
+		
+		for(auto & zonePtr : zones)
```
C3pa, line 206:
```suggestion
		for(const auto & zonePtr : zones)
```

C3pa:
What was the initial reason to prune such roads? Was it for consistency with the original? I always thought these roads leading nowhere were a nice thing.

DjWarmonger:
> What was the initial reason to prune such roads? Was it for consistency with the original? I always thought these roads leading nowhere were a nice thing.

Consistency with original, aimed at PvP gameplay. In particular roads on 6lm10a must behave in the same way to not confuse players.

PR #6353, opened 2025-11-24
Fix vcmiskill cheat code not working due to missing comma

ahammond:
## Description

Fixed a bug where the `vcmiskill` cheat code was not recognized due to a missing comma in the heroTargetedCheats vector initialization.

## Root Cause

When adjacent string literals in C++ lack a comma separator, they are concatenated into a single string. This caused `"nwcfleshwound"` and `"vcmiglaurung"` to be merged into `"nwcfleshwoundvcmiglaurung"`, which corrupted the entire cheat code list.

## Changes

- Added missing comma after `"nwcfleshwound"` in `server/processors/PlayerMessageProcessor.cpp`
- Added `PlayerMessageProcessorTest` to prevent regression

## Testing

The new unit test validates that:
- `vcmiskill` is properly recognized as a separate cheat code
- No accidental string concatenation occurs (e.g., `nwcfleshwoundvcmiglaurung` doesn't exist)

## User Impact

Before: Typing 'vcmiskill wisdom 3' in chat would send the message instead of applying the cheat
After: The cheat code works as expected

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -391,6 +391,11 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR NOT WIN32)
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wmismatched-tags")
 	endif()
 
+	# Enable string concatenation warning for Clang to catch missing commas in initializer lists
+	if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wstring-concatenation")
+	endif()
+
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-unused-parameter")   # low chance of valid reports, a lot of emitted warnings
 	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-switch")             # large number of false-positives, disabled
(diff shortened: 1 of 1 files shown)


ahammond:
Wow, pretty hard fail, SonarCloud. I'm guessing it's detecting duplicate code because of the UT? The fix is literally a character.

test/game/PlayerMessageProcessorTest.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * Full text of license available in license.txt file, in main folder
+ *
+ */
+#include "StdInc.h"
+
+#include <gtest/gtest.h>
+
+// This test verifies that cheat codes are properly recognized
+// It's a simple syntax validation test that doesn't require game state
+TEST(PlayerMessageProcessorTest, CheatCodeVectorSyntax)
```
IvanSavenko: Test feel to be largely pointless.
If somebody were to make similar typo in PlayerMessageProcessor again, it will not be picked up by this test.

Similarly, if somebody were to add or remove cheat, this test won't automatically pick such change

Laserlicht: Is it possible to use this for throwing an error?

https://clang.llvm.org/docs/DiagnosticsReference.html#wstring-concatenation

ahammond:
@Laserlicht good call on that warning. Added!

IvanSavenko:
Closing in favor of #6382 since that PR does the same thing but (IMO) better.

Feel free to submit new warning as separate PR to avoid such cases in future. Thanks for noticing it by the way.

Laserlicht:
Yes, thanks a lot for finding the issue! :)

PR #6356, opened 2025-11-25
Reverse seer hut and reward values

DjWarmonger:
Now Seer Huts will have low value while quest artifacts will be guarded.


IvanSavenko:
@DjWarmonger good to merge?

PR #6361, opened 2025-11-25
add template description as popup if available

Laserlicht:
grafik


IvanSavenko:
Perhaps add name of template as title for this popup? I mean the usual `{Template Name}\n\nTemplate desription`? Most of popups have title, even if it is duplicates underlying UI

Laserlicht:
> Perhaps add name of template as title for this popup? I mean the usual `{Template Name}\n\nTemplate desription`? Most of popups have title, even if it is duplicates underlying UI

Done

PR #6392, opened 2025-12-05
Postpone next turn logic after victory / loss status is settled

### LINKED ISSUE #6366: [Develop] Freeze on endgame
**Describe the bug**
Game freeze on final turn, when the enemy does not have any castle for the 7th day and looses, then ally turn begins and the game freezes

**To Reproduce**

1. Play with allies on a map with a victory condition to eliminate all enemies
2. Enemy turns should be before allies turns
3. The last enemy should loos all towns
4. The last enemy should fail to reclaim a town in a week
5. End your turn  and  the enemy looses on its turn
6. Now its ally turn, but game freezes instead of end as victory

**Expected behavior**
Game end immediately with victory as victory condition reached

**Actual behavior**
Game freeze

**Did it work earlier?**
it worked before in 1.7

**Version**

- OS: macOS
- Version: VCMI 1.7.0.5356511a93a279008b61894be498a395d955547c

mrhaandi:
Fixes #6366.

Only run next turn code after victory / loss status is settled and the server is not shutting down.

### DIFF
--- a/server/CGameHandler.cpp
+++ b/server/CGameHandler.cpp
@@ -3654,9 +3654,9 @@ void CGameHandler::checkVictoryLossConditionsForPlayer(PlayerColor player)
 			}
 			checkVictoryLossConditions(playerColors);
 			// give turn to next player(s)
-			// FIXME: this may cause multiple calls to onPlayerEndsGame if multiple players lose in chain reaction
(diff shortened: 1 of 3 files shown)


fatal10110:
Not sure its related to this PR, but happened on a build from this commit 
https://github.com/vcmi/vcmi/issues/6396

Looks like the AI entered to some endless loop

PR #6394, opened 2025-12-05
rework battle mode button

Laserlicht:
grafik
grafik


IvanSavenko:
Pros:
- Battle mode is clearly visible
Cons:
- Commonly used settings like turn timer & simturns are hard to find, while settings that rarely need change (cheats & battle replay) are in main UI

Perhaps do the opposite in that menu  - by default show turn options menu & add button to go to cheats/battle replay menu from there?

Laserlicht:
> Perhaps do the opposite in that menu - by default show turn options menu & add button to go to cheats/battle replay menu from there?

Also not that nice. Extra options has nothing to do with turn timers, while turn timers could seen as extra option.

Do we really need preset twice? We could for example replace the two preset comboboxes with the turn timer box? Yes. More clicks. But I think many users don't use turn timers (as we have many single player user).

Or if we reduce their sizes:
grafik

Then we would have this option only at one place.

Laserlicht:
Changed to this. IMO the cleanest variant.

Advantages:
- turn options and presets are at one place
- we got space for future button if required

Disadvantages:
- the field needed to resize (needed to trim one title text, but increased size on open popup to avoid shorten german texts) 

grafik

PR #6395, opened 2025-12-05
Exclude invincible units from multihex attacks

mrhaandi:
While an `INVINCIBLE` unit is not targetable, it also should not take damage in a multihex attack.

### DIFF
--- a/lib/battle/CBattleInfoCallback.cpp
+++ b/lib/battle/CBattleInfoCallback.cpp
@@ -329,14 +329,15 @@ BattleHexArray CBattleInfoCallback::battleGetAttackedHexes(const battle::Unit * 
 	for (const BattleHex & tile : at.hostileCreaturePositions)
 	{
 		const auto * st = battleGetUnitByPos(tile, true);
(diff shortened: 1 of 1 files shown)


by003:
My friend said invincible units can be casted by mass slow, can you verify it?

mrhaandi:
> My friend said invincible units can be casted by mass slow, can you verify it?

I tested the current develop (easy using battle mode).
If mass slow was casted, then the invincible unit was not hit.

by003:
> I tested the current develop (easy using battle mode). If mass slow was casted, then the invincible unit was not hit.

Thanks.

by003:
Hello @mrhaandi 
If you're still in charge of fixing this, I stumbled upon a problem where INVINCIBLE can be attacked by SPELL_LIKE_ATTACK, perhaps because you fixed a similar breath attack issue last time.

[#7046](https://github.com/vcmi/vcmi/issues/7046)

Karyoplasma:
> My friend said invincible units can be casted by mass slow, can you verify it?

That has been the case, but it got fixed shortly before 1.7.2 was released.

PR #6399, opened 2025-12-06
Quickload improvements

Laserlicht:
Small improvements:
- load button is now disabled when it's not possible to load
- translate player visible (chat) string

@mrhaandi

### DIFF
--- a/client/CServerHandler.cpp
+++ b/client/CServerHandler.cpp
@@ -695,7 +695,18 @@ void CServerHandler::endGameplay()
 std::optional<std::string> CServerHandler::canQuickLoadGame(const std::string & path) const
 {
 	auto mapInfo = std::make_shared<CMapInfo>();
-	mapInfo->saveInit(ResourcePath(path, EResType::SAVEGAME));
+	try
+	{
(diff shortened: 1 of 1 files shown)


mrhaandi:
I also though about a flag like `hasQuickSave` and in the initial commit decided against it in order not to introduce another potential decoherence. However, if computational resources are limited or disc access is slow, `hasQuickSave` works well.

Is there any imaginable interaction which can set `hasQuickSave` from `true` back to `false`?
Should we consider setting it to false, if `canQuickLoadGame` returns an error? (Both decisions are reasonable.)

Laserlicht:
I've also found some unhandled errors crashing the game (with new approach they are easier to see). Added try-catch - like in other places we load a map in similar way.

> Should we consider setting it to false, if `canQuickLoadGame` returns an error? (Both decisions are reasonable.)

Added an fallback to false.

> However, if computational resources are limited or disc access is slow

From what I've measured `checkQuickLoadingGame` takes a few ms (even on 200x200 and more maps). For initial loading this is okay.

PR #6403, opened 2025-12-06
fix NodeStorage::initialize for removed player

### LINKED ISSUE #1962: Game end condition happens multiple time in Hotseat + game is not ending
**Describe the bug**
When a player attacks another and retreats, multiple win messages appear (2) and game does not end.

**To Reproduce**
1. Start hotseat
2. Select attached map
3. Attack blue player
4. Retreat
5. Observe

**Expected behavior**
Victory condition only once and game should end.

[3069_vcmiarchangelresurrecttest.zip](https://github.com/vcmi/vcmi/files/11214854/3069_vcmiarchangelresurrecttest.zip)

Mantis: https://bugs.vcmi.eu/view.php?id=3069

mrhaandi:
Fixes #1962.

What happend: on victory / loss the server starts a shutdown, but object removal code is still executed leading to reinitialization of NodeStorage.

With the proposed fix the situation in the underlying issue is resolved as expected (one message for red loss, one message for blue win, end game screen).

I added a warning to global log, if the changed code would silence an otherwise illegal call to `NodeStorage::initialize` in a different setting.

### DIFF
--- a/lib/pathfinder/NodeStorage.cpp
+++ b/lib/pathfinder/NodeStorage.cpp
@@ -28,7 +28,14 @@ void NodeStorage::initialize(const PathfinderOptions & options, const IGameInfoC
 	int3 pos;
 	const PlayerColor player = out.hero->tempOwner;
 	const int3 sizes = gameInfo.getMapSize();
-	const auto & fow = gameInfo.getPlayerTeam(player)->fogOfWarMap;
+	const TeamState * team = gameInfo.getPlayerTeam(player);
+	if (!team)
+	{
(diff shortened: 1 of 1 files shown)


mrhaandi:
This PR is a rather minimalistic ad-hoc fix. It would be cleaner to communicate the servers `CGameHandler` to clients (immediately?) that the server is shutting down for a cleaner client-side unwind. Would `IGameInfoCallback.h` be the right place?

mrhaandi:
Superseded by #6433.

PR #6405, opened 2025-12-07
Reduce AI logging to avoid performance hit

IvanSavenko:
Not sure how much of AI performance drop was caused by excessive logging, and how much - by new AI, but should be safe change in any case.


fatal10110:
I can confirm this PR improves the AI times, from 2-3 miute or even more, to ~1 minute (all 7 AIs)

PR #6412, opened 2025-12-08
Extends creature terrain limiter to hero and towns nodes

Opuszek:
As in title.
Currently bonuses that are limited by terrain and propagated to hero/town do not work (e.g. Desert Warrior bonus from new Pavilion mod).
PS: if somebody is curious (I was) native terrain for a hero depends on the class (e.g. grass for a knight, snow for an alchemist) and current terrain for a town depends on the "entrance to town" tile.

### DIFF
--- a/config/schemas/bonusInstance.json
+++ b/config/schemas/bonusInstance.json
@@ -112,13 +112,13 @@
 						"type" : "object",
 						"required" : [ "type" ],
 						"properties" : {
-							"type" : { "type" : "string", "const" : "CREATURE_TERRAIN_LIMITER" }
+							"type" : { "type" : "string", "enum" : [ "CREATURE_TERRAIN_LIMITER", "TERRAIN_LIMITER" ] }
 						}
 					},
 					"then" : {
 						"additionalProperties" : false,
 						"properties" : {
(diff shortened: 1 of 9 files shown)


IvanSavenko:
Please also update docs and schemas - both limiter name & which objects are affected. Looks ok othervice

Potential edge case - for example, hero with bonus that has such limiter without specified terrain (native terrain mode) will affect creatures on native terrain of creature (and not on native terrain of hero) since limiters are checked after propagation

Opuszek:
I do realize it is a problem, unfortunately, I don't know how to resolve it. Bonuses are stateless - what is nice and tidy, but has its limitations. I played for some time with an idea of creating two types of limitators: one would work as they currently work, propagated freely and applied on each node, the second would propagate only if not blocked by limitator, but would propagate without limitator, applying freely to each descendant. The changes went too deep though and I dropped the idea.

Opuszek:
As of changing the limiter name: what about backward compatibility?

IvanSavenko:
```
                {"CREATURE_TERRAIN_LIMITER",   parseTerrainLimiter  },
```

For compatibility - copy- paste this line and rename one of them to TERRAIN _LIMITER

This way old name would still work for mods compatibility

PR #6424, opened 2025-12-11
Adhere to Heroes 3 chain lightning mechanics

### LINKED ISSUE #6419: Chain lightning reacts differently to spell resistance in VCMI
**Describe the bug**
in original H3 spell chain lightning spell works in pretty specific way - when spell resistance is going to trigger for non-first target then target gets omitted and next one gets picked instead as "replacement target" (this is pretty subtle mechanic as in original H3 resistance battle animation that usually plays is skipped in this case). So basically in VCMI chain lightning instance will have less total targets per each spell resistance trigger, which is not the case in original game.

**To Reproduce**
Fight a battle with few stacks of dwarves, cast chain lightning bunch of times and see that sometimes there are less than 4/5 total targets (on basic spell level there are 4 targets by default) when there are 4+ creatures on the battlefield

**Expected behavior**
Chain lightning behaving as in original game - pending target for chain lightning gets changed to next one rather than skipped when spell resistance triggers

**Actual behavior**
Each spell resistance trigger for non-first chain lightning target results in chain lightning having less total targets

**Did it work earlier?**
No

**Version**
Any

mrhaandi:
Fixes #6419.

- if first units triggers magic resistance : no chain
- if a later unit triggers magic resistance : skip unit, no resist animation, do not shorten chain length

This requires the communication of rng info on magic resistance to spell effects. This is done by `wouldResist`, which does not invoke rng, but encapsulates prior rng info on magic resistance.

### DIFF
--- a/test/mock/mock_spells_Mechanics.h
+++ b/test/mock/mock_spells_Mechanics.h
@@ -34,6 +34,7 @@ public:
 	MOCK_METHOD2(castEval, void(ServerCallback *, const Target &));
 
 	MOCK_CONST_METHOD1(isReceptive, bool(const battle::Unit * ));
+	MOCK_CONST_METHOD1(wouldResist, bool(const battle::Unit * ));
 	MOCK_CONST_METHOD0(getTargetTypes, std::vector<AimType>());
 	MOCK_CONST_METHOD4(getPossibleDestinations, std::vector<Destination>(size_t, AimType, const Target &, bool));
 


lib/spells/BattleSpellMechanics.cpp
```diff
@@ -445,16 +445,21 @@ void BattleSpellMechanics::beforeCast(BattleSpellCast & sc, vstd::RNG & rng, con
 
 	std::vector <const battle::Unit *> resisted;
 
-	auto filterResisted = [&, this](const battle::Unit * unit) -> bool
+	resistantUnitIds.clear();
+	if(isNegativeSpell() && isMagicalEffect())
 	{
-		if(isNegativeSpell() && isMagicalEffect())
+		//magic resistance
+		for (const auto * unit : battle()->battleGetAllUnits(false))
```
IvanSavenko, line 452: Would've been great to avoid testing resistance for every unit on battlefield for every cast of every spell, and only perform it if unit is actually hit by the spell...

Don't see any clear way to do this though. Any ideas?

IvanSavenko (approved):

lib/spells/BattleSpellMechanics.cpp
mrhaandi: > Don't see any clear way to do this though. Any ideas?

One can instead of an a priori set for resistant units have a map (which is basically a cache) from unit ids to bool, which is populated on demand in `BattleSpellMechanics::wouldResist`. This way only the requested units would be checked.

I usually prefer eager evaluation to lazy evaluation due to maintainability. If you are worried about performance, I could implement the map solution.

IvanSavenko: Let's keep as it for now.

Performance is very unlikely to be a factor here, it just feels illogical to test for resistance when resistance likely won't even happen.

PR #6428, opened 2025-12-11
Add bonus FORCE_NEUTRAL_ENCOUNTER_STACK_COUNT for neutral stack splitting

kdmcser:
This PR adds a new bonus named `FORCE_NEUTRAL_ENCOUNTER_STACK_COUNT`, which controls how neutral creatures are split into stacks when a hero engages them.

The bonus supports two modes:

Fixed mode (val > 0): the neutral army is always split into exactly val stacks.

Weighted mode (val = 0 + addInfo): addInfo is a list of up to 7 integers, representing the weights for generating 1–7 stacks.

I plan to use a distribution such as·`[20, 60, 20]`, which simulates the stack-splitting behavior in Heroes III when the player’s army is very strong.
This means the neutral army may be split into 1, 2, or 3 stacks with weights 20, 60, and 20 respectively.
Different stack counts lead to different tactical decisions, making encounters less repetitive than always using a fixed split, and ultimately making the mod more interesting.

Both val and addInfo are introduced in this PR:
val provides a simple deterministic option, while addInfo offers the flexible weighted configuration needed for mods that want controlled randomness.

Additionally, if VCMI supports Lua scripting in the future, the val = 0 case could also be implemented via script, allowing even more advanced and customizable rules for neutral stack generation.

However, Lua support still seems far off, and I plan to release my mod around VCMI 1.7. Therefore, I am providing a simple solution that does not rely on Lua for now.


docs/modders/Bonus/Bonus_Types.md
```diff
@@ -413,6 +413,14 @@ Heroes affected by this bonus will not lose all movement points when embarking o
 
 Heroes affected by this bonus won't lose army when moving through whirlpool
 
+### FORCE_NEUTRAL_ENCOUNTER_STACK_COUNT
+
+When the hero engages neutral creatures, this bonus forces the number of stacks that the neutral army will be split into.
+
+- val: the neutral army will be split into (0 – enable weighted mode using addInfo)
+- addInfo – [W1, W2, W3, W4, W5, W6, W7]
```
IvanSavenko, line 421: Is such complex configuration really necessary? Why not just stick to `val`? Considering hota map format might make more sense to reserve addInfo for "presence of random stack" flag

lib/mapObjects/CGCreature.cpp
```diff
@@ -582,7 +582,54 @@ bool CGCreature::containsUpgradedStack() const
 	return ((val % 32768) % 100) < 50;
 }
 
-int CGCreature::getNumberOfStacks(const CGHeroInstance *hero) const
+int CGCreature::getNumberOfStacks(const CGHeroInstance * hero) const
+{
+	int split = 0;
+	// TODO: If this value is defined in a HotA map, use it first.
```
IvanSavenko: Please also implement this option in a way that won't confict with such bonus.

Game already loads that value (see `CMapLoaderH3M::readMonster`), but it is immediately dropped

lib/mapObjects/CGCreature.cpp
```diff
@@ -582,7 +582,54 @@ bool CGCreature::containsUpgradedStack() const
...
+	return split;
+}
+
+int CGCreature::getNumberOfStacksFromBonus(const CGHeroInstance * hero) const
+{
+	auto bonus = hero->getBonus(Selector::type()(BonusType::FORCE_NEUTRAL_ENCOUNTER_STACK_COUNT));
+	if(bonus->val > 0)
+		return bonus->val;
+
+	// TODO: When VCMI adds support for Lua scripting, this should provide a script-based solution.
```
IvanSavenko: Too far in future IMO to think about it. And I still don't see any solid approach on how to allow scripts to redefine such functions
```suggestion
```

docs/modders/Bonus/Bonus_Types.md
kdmcser: Yes, it’s absolutely necessary. My mod design can’t have all creatures limited to a single stack size; otherwise, the tactics in every battle would become extremely monotonous.

I think these are two separate bonuses: one is the forced probability for the opponent’s stack, and the other is the unit’s own stack size. Mixing them together feels a bit odd. Or should it be defined as a property of CGCreature instead?

lib/mapObjects/CGCreature.cpp
kdmcser: Okay, then I’ll remove this TODO comment and won’t consider this case.

lib/mapObjects/CGCreature.cpp
kdmcser: I’ve already implemented this functionality by adding a stacksCount property to CGCreature.

docs/modders/Bonus/Bonus_Types.md
kdmcser: I’ve already implemented this functionality by adding a stacksCount property to CGCreature.

PR #6430, opened 2025-12-11
Remove unused water wheel from the configs

Olegmods:
Currently, the Fortress water wheel building is built along with the Capitol, which isn't in line how it is in SoD/Complete, where it's completely unused. This PR fixes it by removing water wheel from the config of Fortress


misiokles:
Bad move. If we want to be purists we should also bring infamous Armorer turred bug to be in line with SoD/Complete.

IvanSavenko:
Armorer bug is clearly a bug. 

This one?  Hard to say. From what I remember, this water wheel even has selection mask / yellow border image, meaning it might have been intended as independent building, not as decoration

Olegmods:
> Bad move. If we want to be purists we should also bring infamous Armorer turred bug to be in line with SoD/Complete.

Did I do sth wrong? Idk about the turret bug, but this one is a minor thing that can be added via a mod.

> Armorer bug is clearly a bug.
> 
> This one? Hard to say. From what I remember, this water wheel even has selection mask / yellow border image, meaning it might have been intended as independent building, not as decoration

It was intended as an independent building indeed, to my knowledge.

Also, to elaborate why I did that PR: that thing might've confused some amount of people (like you don't have Fortress Water Wheel in SoD/Complete/HotA, even in ERA except like one mod, IIRC, and there is functionally to it).
I think it's either better to bring that one back with some sort of mod, either as a decoration or as a functional object (what would be more interesting, in my opinion).

PR #6433, opened 2025-12-12
Add ingame guard to `onTileHovered`

### LINKED ISSUE #1962: Game end condition happens multiple time in Hotseat + game is not ending

mrhaandi:
Fixes #1962.

If the current player is not ingame, then we are in a shutdown process, which results in crashes if we execute `onTileHovered` game logic (e.g. `NodeStorage` is reinitialized with incoherent data).

Now on hotseat win there is no crash, but message flow as expected (one message for red loss, one message for blue win, end game screen).

Tested in single player, hotseat, and multi player using the map from #1962.

### DIFF
--- a/client/adventureMap/AdventureMapInterface.cpp
+++ b/client/adventureMap/AdventureMapInterface.cpp
@@ -603,7 +603,9 @@ void AdventureMapInterface::onTileHovered(const int3 &targetPosition)
 {
 	if(!shortcuts->optionMapViewActive())
 		return;
-
+	//if the player is not ingame (loser, winner, wrong) we are in a shutdown process
+	if (!GAME->interface()->cb || GAME->interface()->cb->getPlayerStatus(GAME->interface()->playerID) != EPlayerStatus::INGAME)
+		return;
 	//may occur just at the start of game (fake move before full initialization)
 	if(!GAME->interface()->localState->getCurrentArmy())
 		return;


IvanSavenko:
Feels like workaround. Why adventure map even active in this scenario? Shouldn't game immediately switch to win/loss screen?

mrhaandi:
> Feels like workaround. Why adventure map even active in this scenario? Shouldn't game immediately switch to win/loss screen?

At the core of the problem is the fact that `GameEngine::fakeMouseMove()` is called from game logic code and that the server shuts down in `CGameHandler::checkVictoryLossConditionsForPlayer` while also giving control to the next player (as it should).

So when the server is already in `EServerState::SHUTDOWN` and the winning player gets control and initiates `GameEngine::fakeMouseMove()`, it triggers `AdventureMapInterface` code.

I did not find a clean way in `GameEngine` to prevent `fakeMouseMove`. There is already extra code for initialization in `AdventureMapInterface` (where the state is also not fully coherent). The proposed PR adds winddown code for basically the same reason.

Issue #5643, opened 2025-04-22
About war machines' promotion


by003:
**Describe your proposal**
 In HotA 1.7.2, war machines cannot be dealt more than 40% of their health as damage from any single damage source, by using relevant skills.

Although we can use some json files to achieve two times 100% damage reduction, but it can be easily consumed by some low-damage units. 

I suggest to use a bonus like 'MAX_DAMAGE_PERCENT' to restore it.

Show example here.
```
"effects" : {
				"damageReduce" : {
					"subtype" : "creature.ballista",
					"type" : "MAX_DAMAGE_PERCENT",
					"valueType" : "BASE_NUMBER",
                                        "val": 40
				}
			}
```

PR #6436, opened 2025-12-12
[HotA] Add DAMAGE_RECEIVED_CAP bonus

### LINKED ISSUE #6385: Damage Cap Bonus (as in HotA)
**Describe your proposal**
In HotA:
```
War machine skills (Ballistics for the Catapult, Artillery for the Ballista and Cannon, and First Aid for the First Aid Tent) now prevent their respective war machines from losing more than 40% of their hit points in a single strike.
The percentage is calculated based on the machine’s total Hit Points at the beginning of the combat, factoring in all modifiers.
Protection applies to all types of damage (physical, magical, etc.)
If the attack deals damage multiple times, then each instance of damage taken counts as a separate strike.
```

in vcmi on can only set the bonus
```
### GENERAL_DAMAGE_REDUCTION

Affected units will receive reduced damage from attacks by other units

- val: damage reduction, percentage
- subtype:
  - damageTypeMelee: only melee damage will be reduced
  - damageTypeRanged: only ranged damage will be reduced
  - damageTypeAll: all damage will be reduced
```

but there is no possibility to set a damage cap based on max hit points.
I like the HotA bonus, which makes war machines less squishy in mid / late game.

It is imaginable to either extend `GENERAL_DAMAGE_REDUCTION` to include damage cap values (I think this could be elegant) or add a new bonus.

mrhaandi:
Fixes #6385, #5643.

Add bonus `DAMAGE_RECEIVED_CAP`, which limits the maximal damage a unit can received based on max hp of the unit (HotA war machines).
`DAMAGE_RECEIVED_CAP` with `"val" : 40` means that a single source of attack cannot deal more damage than 40% of the units max hp.

**DONE**
- add `DAMAGE_RECEIVED_CAP` bonus and adapt damage calculation
- fix propagated bonuses are not displayed for war machines (see `fakeNode` does not have current bonus info)
- add new bonus graphics to vcmi-extras https://github.com/vcmi-mods/vcmi-extras/pull/81
- add bonus to war machines according to HotA documentation https://github.com/vcmi-mods/horn-of-the-abyss/pull/271
- testing (first aid tent, ballista, catapult, cannon) in single player, hotseat, multiplayer, battle only

Bonus graphics: helath-based damage cap
damage_cap

### DIFF
--- a/lib/battle/DamageCalculator.cpp
+++ b/lib/battle/DamageCalculator.cpp
@@ -546,11 +546,11 @@ int64_t DamageCalculator::getDamageCap() const
 	const std::string cachingStrDamageCap = "type_DAMAGE_RECEIVED_CAP";
 	static const auto selectorDamageCap = Selector::type()(BonusType::DAMAGE_RECEIVED_CAP);
 
-	double damageCapPercent = info.defender->valOfBonuses(selectorDamageCap, cachingStrDamageCap) / 100.0;
-	if (damageCapPercent <= 0.0)
+	int damageCapPercentage = info.defender->valOfBonuses(selectorDamageCap, cachingStrDamageCap);
+	if (damageCapPercentage <= 0)
 		return std::numeric_limits<int64_t>::max();
 
-	return std::floor(info.defender->getMaxHealth() * damageCapPercent);
+	return info.defender->getMaxHealth() * damageCapPercentage / 100;
 }
 
 DamageEstimation DamageCalculator::calculateDmgRange() const


mrhaandi:
I adapted the HotA and vcmi-extras mods accordingly and tested the PR in different configurations (first aid tent, ballista, catapult, cannon) in single player, hotseat, multiplayer, battle only.

Everything seems to be working as expected and I consider this PR ready for review.

lib/battle/DamageCalculator.cpp
```diff
@@ -566,12 +578,13 @@ DamageEstimation DamageCalculator::calculateDmgRange() const
...
 	int64_t avail = info.defender->getAvailableHealth();
+	int64_t cap = getDamageCap();
 
 	auto dmin = std::max<int64_t>(1.0, std::floor(damageBase.min * resultingFactor));
 	auto dmax = std::max<int64_t>(1.0, std::floor(damageBase.max * resultingFactor));
 
-	dmin = std::min(dmin, avail);
-	dmax = std::min(dmax, avail);
+	dmin = std::min(std::min(dmin, avail), cap);
+	dmax = std::min(std::min(dmax, avail), cap);
```
IvanSavenko:
```suggestion
	dmin = std::min({dmin, avail, cap});
	dmax = std::min({dmax, avail, cap});
```

lib/battle/DamageCalculator.cpp
```diff
@@ -541,6 +541,18 @@ int DamageCalculator::battleBonusValue(const IBonusBearer * bearer, const CSelec
...
+int64_t DamageCalculator::getDamageCap() const
+{
+	const std::string cachingStrDamageCap = "type_DAMAGE_RECEIVED_CAP";
+	static const auto selectorDamageCap = Selector::type()(BonusType::DAMAGE_RECEIVED_CAP);
+
+	double damageCapPercent = info.defender->valOfBonuses(selectorDamageCap, cachingStrDamageCap) / 100.0;
+	if (damageCapPercent <= 0.0)
+		return std::numeric_limits<int64_t>::max();
+
+	return std::floor(info.defender->getMaxHealth() * damageCapPercent);
```
IvanSavenko: No need to use floating points here:
```suggestion
	int damageCapPercentage = info.defender->valOfBonuses(selectorDamageCap, cachingStrDamageCap);
	if (damageCapPercentage <= 0)
		return std::numeric_limits<int64_t>::max();

	return info.defender->getMaxHealth() * damageCapPercentage / 100;
```

mrhaandi:
@IvanSavenko I included both of the proposed changes and retested, to be on the safe side.

PR #6437, opened 2025-12-12
mapeditor - fix special terrain display

### LINKED ISSUE #6380: [Map editor] Map objects are obscured by terrain
**Describe the bug**
Map objects are obscured by special terrian in map editor, from a very classic map All for One.

Image

Actually it should be (in OH3's map editor):

Image

**Version**

- OS: Windows
- Version: 1.70 daily build

Opuszek:
Fixes #6380.
Currently there seems to be a bug in the client in case of multiple overlaying special terrains, when one terrain is displayed and a different one is active.

https://github.com/user-attachments/assets/7b91a538-3ee5-483d-8323-67b224abba41

I will take a look into this.

### DIFF
--- a/mapeditor/maphandler.cpp
+++ b/mapeditor/maphandler.cpp
@@ -371,29 +371,30 @@ void MapHandler::drawObjects(QPainter & painter, const QRectF & section, int z, 
 {
 	painter.setRenderHint(QPainter::Antialiasing, false);
 	painter.setRenderHint(QPainter::SmoothPixmapTransform, false);
-	std::map<int3, std::set<const CGObjectInstance *>> objectMap;	//following the natural order of int3 we draw from north-west to south-east, in accordance with H3's perspective
+	auto blitOrder = [](const CGObjectInstance * a, const CGObjectInstance * b) { return MapHandler::compareObjectBlitOrder(a, b);};
+	std::set<const CGObjectInstance *, decltype(blitOrder)> objects;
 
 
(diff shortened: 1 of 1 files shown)


mapeditor/maphandler.cpp
```diff
@@ -384,17 +385,27 @@ void MapHandler::drawObjects(QPainter & painter, const QRectF & section, int z,
 		for(int y = top; y < bottom; ++y)
 		{
 			for(auto & object : getObjects(x, y, z))
-				objectMap[object.obj->pos].insert(object.obj);
+				if (object.obj->isTile2Terrain())
+					specialTerrains[object.obj->pos].insert(object.obj);
+				else
+					standardObjects[object.obj->pos].insert(object.obj);
```
IvanSavenko: Can you check that (for example) Ore Mine does not have this problem as well? It should appear below all other map objects, potentially - even below special terrains 

Test case - map Deluge: https://heroes.thelazy.net/images/c/c5/Deluge_underground_map_fullauto.png
There is Ore Mine to the right from town that is below special terrains.

To clarify - what you need to check for is not `obj->isTile2Terrain()`, but rather `appearance->printPriority != 0`

Opuszek: Call me a donkey and make me jump through hoops in a circus, I was inventing circle again.
It turned out the correct comparator was already provided.

IvanSavenko (approved):

